PR #657, opened 2012-05-06
Auto-completion for bash

juzna:
**Work in progress!**

This is just a first experiment with *bash completion*, it can complete *commands* and *options*. It would need more work until it's ready. But I just **wanted your feedback** - do you consider it useful?

Thanks for comments

### DIFF
--- a/composer_completion.sh
+++ b/composer_completion.sh
@@ -1,0 +1,88 @@
+# Copy to /etc/bash_completion.d/composer and restart your bash
+
+_composer()
+{
+	local cur prev opts commands
+	COMPREPLY=()
+	cur="${COMP_WORDS[COMP_CWORD]}"
+	prev="${COMP_WORDS[COMP_CWORD-1]}"
+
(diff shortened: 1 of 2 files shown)


bamarni:
I made some improvements to my tool lately, it can work out of the box for Composer and Symfony based tools in general. While there is also a way to generate a static completion script for a given tool, see https://github.com/bamarni/symfony-console-autocomplete/blob/master/README.md

About the script in this PR, is it the correct place for it to live in Composer sources? Most people don't have them as they just use the phar. If we want to have a more advanced script with package completion, what about making it available at an endpoint in https://getcomposer.org/ for example? And have an example in the docs showing how to download / enable it.

Seldaek:
@bamarni I think the best would be to add a link to your repo somewhere in the doc/* maybe in 03-cli.md or something, just to have a pointer so people looking for it are more likely to find it. That way the instructions and so on stay on your end and up to date.

I'll close this PR now, please open a PR if you like to have us link to your repo.

PR #1013, opened 2012-08-19
Add "extends" property to JSON format allowing for config file inheritance

1stvamp:
Recently we've had the requirement to have multiple JSON files per deployment environment, e.g. `testing.json` and `production.json` configs for our CI and production environments that define different `repositories` to the development `composer.json` in order to use SSH aliases to get around the one-deploy-key-per-repo limitation for Github private repos.

Currently I'm using a shell script that is ran during CI and production deploys to generate a copy of the `composer.json` with hostnames replaced, in order to avoid duplicating lots of JSON properties.

The Python build tool buildout defines an `extends` property that allows you to inherit properties from another config file, e.g. have a main config file and then just override the properties you need to in your environment files.

I've added an `extends` property like this to the Composer JSON format that recursively adds parent properties during `JsonFile::read()`.
Unittest too, 'natch.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -8,5 +8,10 @@ use Composer\Compiler;
 error_reporting(-1);
 ini_set('display_errors', 1);
 
-$compiler = new Compiler();
-$compiler->compile();
+try {
+    $compiler = new Compiler();
+    $compiler->compile();
+} catch (\Exception $e) {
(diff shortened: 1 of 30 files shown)


firegate666:
Desperately waiting for this one. Please solve your merge conflicts and submit it again,

1stvamp:
This PR has been superseded by #4210 (in which @aripringle has refactored everything so it'll merge and work with latest composer HEAD).

Issue #515, opened 2012-03-31
Support for "git archive" downloader


GromNaN:
To deploy a private Git repository, there currently no dist option.
https://github.com/composer/composer/blob/master/src/Composer/Repository/Vcs/GitDriver.php#L112

To avoid downloading the whole Git repository to get only a snapshot, the command `git archive` can be used.
http://stackoverflow.com/a/353830

As `git archive --remote` is not supported by every server it may not be the standard dist option.

I can work on this feature, but I need some directions:
- Should I create a the class `GitArchiveDownloader` ?
- Should add the feature into the `GitDriver::getDist` method ?

Some documentation:
- http://ftp.sunet.se/pub/Linux/kernel.org/software/scm/git/docs/git-archive.html
- http://schacon.github.com/git/git-upload-archive.html

PR #1024, opened 2012-08-24
Feature dist (bis)

MattKetmo:
This is a refacto of #804 from @till.

It adds the ability to create zip/tar archives from sources.

The main differences with the initial PR:
- It uses a separate directory, named `Archiver` instead of `Dumper`, because as @Seldaek [said](https://github.com/composer/composer/pull/804#discussion_r1049991), it has nothing to do with `ArrayDumper`
- It uses a manager which have the main function `archive($package, $format)`, which will use its own archivers, so no code is duplicated
- We first try to use VCS archivers (git & hg archive commands). If the VCS is not able to archive a package (for instance svn or any future VCS which could be added to Composer), then it uses standard tar/zip archiver (which archive a directory no matter the source type).

I preferred to start from `master` because the initial PR started to have to much dummy commits, but I credited @till in the sources ;)

### DIFF
--- a/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
+++ b/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
@@ -33,7 +33,7 @@ class ArchiveManagerTest extends ArchiverTest
 
         $factory = new Factory();
         $this->manager = $factory->createArchiveManager(null, $factory->createConfig());
-        $this->targetDir = sys_get_temp_dir().'/composer_archiver_tests';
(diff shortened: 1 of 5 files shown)


kentr:
Got here by following https://github.com/composer/composer/issues/515 to https://github.com/composer/satis/issues/12 to here.

Getting back to "Support for "git archive" downloader", does the change referenced here allow for using `git archive` to pull `dist` versions of private BitBucket repos?

The use-case is for private repos that are accessed via public key (not via Oauth).

Issue #1074, opened 2012-09-06
Remote Code Execution Vulnerability via MITM.


StormTide:
Composer runs code from HTTP sources without validating the source of the download or the code downloaded. As such, trivial man-in-the-middle attacks through any number of vectors (dns, networking, local server exploit, etc) will result in execution of code of an attackers choosing at the userlevel of the user running composer. (Typically a developer account)

Example command to reproduce. (User)

> php composer.phar self-update

src/Composer/Command/SelfUpdateCommand.php lines 45, 50 and 59 are key. 

Exploitation:

Replace getcomposer.org for a given network perspective by replacing it with a malicious http instance (eg by changing the DNS locally, at the lan, at an isp or hosting provider dns resolver, or globally or equally easily by replacing a route to the legitimate server (eg arpspoof)) . The http server instance is configured to serve a malicious /composer.phar and a /version url that produces random data. When users run self-update, the malicious code will be downloaded and run as the user that is executing the self-update command.

The same technique can be used for any package configured over http.

Fix:

Disable HTTP scheme's in composer and implement TLS or redesign composer to support digital signatures. (See AptSecure)

mikedugan:
Sorry for the necrobump, but as a not-security-guru, could someone help me understand what the core problem is here and what has prevented it from being fixed after well over 2 years? Conversely, is there a reason this is no longer as much of an issue as it was 2 years ago?

Not trying to ruffle feathers, just want to understand what's going on here.

StormTide:
@mikedugan core problem is that by using composer and http sources anyone near you can install code on your server through a man-in-the-middle type attack. There is no code-signing mechanism in composer, as would be found in just about any other modern package manager.

As for what has stopped them from even updating the instructions on the website or fixing this bug for years.... I'll leave that as an exercise to the reader.

sybarite:
@StormTide Maybe working on a community PR would be better than expecting the original devs to fix this anytime soon? I wouldn't mind contributing in my off time, but my progress would be very slow as I have lot of other commitments. Has anyone ever worked like this? Like a one big community powered pull request?

StormTide:
@sybarite I dont have the time/energy to fork composer... The fix is a
one liner... disable http schemes. The fix to the website is also just
as easily patched. The more complex question of codesigning (which
really is the solution to this whole nonsense (and i'd add was invented
well over a decade ago and lives happily as apt-secure))... would take
some community effort and has been discussed over n over to no
[...]

PR #1092, opened 2012-09-10
[work-in-progress] Fix MITM vulnerability for issue #1074 - Needs review

EvanDotPro:
### Disclaimer: I am not a security expert. This PR is mostly common sense, but should still be reviewed by the community before being merged, especially those well-versed in security.

This PR contains the following security fixes, and attempts to address [issue #1074](https://github.com/composer/composer/issues/1074):
- Self-update and packagist are now forced to use SSL.
- This introduces a hard dependency on the openssl extension when using
  self-update or packagist. Users should not be using either without
  utilizing a proper SSL connection anyway.
- Keeps a trusted CA bundle file in the composer home directory.
- Attempts to copy over the operating system CA bundle if available.
- Upon encountering an untrusted certificate, the user is prompted if
  they would like to trust the certificate and the CA in the future. If
  they choose yes, the root CA for that cert is added to the CA bundle
  in the composer home directory.
- Expired certificates generate an extra warning and prompt.
- Invalid common names generate an extra warning and prompt.

(I'll try to add unit tests where possible.)

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -39,7 +39,7 @@ class ComposerRepository extends ArrayRepository implements NotifiableRepository
     {
         if (!preg_match('{^\w+://}', $repoConfig['url'])) {
             // assume http as the default protocol
-            $repoConfig['url'] = 'http://'.$repoConfig['url'];
+            $repoConfig['url'] = 'https://'.$repoConfig['url'];
         }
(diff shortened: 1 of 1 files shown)


rubensayshi:
@EvanDotPro so to go back on an old issue, but could you reference the stuff you're talking about here?

bd808:
@rubensayshi wrote:
> could you reference the stuff you're talking about here?

See PR #2745 (which also seems to have stalled out)

PR #1120, opened 2012-09-18
Downloader Dependencies

ukautz:
Check for dependencies of downloaders (eg whether git, hg, svn.. binary installed, if package requires so) on upload/install and dies with runtime error, if requirements not satisfied.


GromNaN:
Running these check commands will have a bad impact on composer performance. Having such check sounds interesting to unserstand issues, but useless once the system is configured.

To avoid performance impacts, but get debug information, it would be better to run the check only after an error occured to check if it is due to a missing system requirement of a runtime error.

Seldaek:
Well we already do that if a git clone fails and you don't have git for example, to be honest nobody complained about this in years so I think it can safely be closed.

Issue #867, opened 2012-07-03
Use require_once instead of require with 'autoload' option.


tmartin:
Hi!

I'm facing an issue with a project that depends on [Swiftmailer](https://github.com/swiftmailer/swiftmailer). This library doesn't implement PSR-0, so it's using the `autoload: files:` option.

The problem is that the code generated in `vendor/autoload.php` uses `require` instead of `require_once`. This causes vicious autoloading errors in some weird cases.

See [this issue](https://github.com/Behat/Symfony2Extension/issues/13) for more details.

Let me know if you need more information :)

PR #1176, opened 2012-10-02
Support a 'files-once' autoload option.

claylo:
For the edge cases where the most streamlined path is a "files" include with require_once, allow it with a "files-once" autoload node.

I've added discouraging language around this choice to the documentation, but as #867 illustrates, PHPUnit isn't, hasn't been, and won't be the only package seeking Composer compatibility that would benefit from this solution.

Package maintainers who want to get their packages into Composer-readiness as quickly as possible, and haven't yet had an opportunity to resolve their older approaches to autoloading could use this feature as a stopgap measure that reduces issues with their end users, which in turn frees them up to do the refactoring necessary to no longer need `files-once`. Everybody wins!


Wilkins:
+1 for @claylo patch

Why I need it : 
I created a composer package from a legacy library. In that library we have : 
- file A.php
- file B.php (which require_once A.php)

My `composer.json` file looks like : 
```
    "autoload": {
        "psr-0": { },
        "files": [
            "src/A.php",
            "src/B.php"
        ]
    }
```
I know B.php should not require_once A.php but if I remove it, projects who use my library without composer or without using autoload will break.
Having a require_once will fix this case.

Seldaek:
You can fix this by doing a conditional require in B.php IMO, so it only gets required if it's not been required yet? Or check for the presence of composer's autoload.php and if it's missing then require yourself.. There are many ways to go around this in legacy code bases.

PR #1184, opened 2012-10-05
translate to chinese 

wyixin:
I want to fork this and translates for chinese readers.

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -1,37 +1,30 @@
-# Introduction
+# 说明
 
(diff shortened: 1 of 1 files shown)


GromNaN:
Apparently, the [composer documentation has been translated into chinese](http://www.phpcomposer.com/).

@Seldaek This issue can be closed.

Seldaek:
I guess so yeah, thanks. If I knew who did this we could add a link to it on getcomposer.org - I mean a link in Chinese would be good.

PR #1188, opened 2012-10-08
Allow interactive resets or stash/apply cycles when updating dirty packages instead of failing hard

### LINKED ISSUE #466: Ask user if they want to revert local changes when an update is blocked
Currently Vcs-downloaders throw an exception if the dir to be updated isn't in a clean status, but asking (if in interactive mode) and then reverting for the user if he answers yes would be nicer.

Seldaek:
Fixes #466

Hg is missing but I don't know it enough so I'll let someone add that later.

Sample output:

```
$ echo "foo" > vendor/composer/composer/README.md
$ composer update composer/composer
Updating dependencies
  - Updating composer/composer dev-master (b3077b => c49247)
    The package has modified files:
    M README.md
    Discard changes [y,n,v,s,?]?
    y - discard changes and apply the update
    n - abort the update and let you manually clean things up
    v - view modified files
    s - stash changes and try to reapply them after the update
    ? - print help
    Discard changes [y,n,v,s,?]?
```

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -122,7 +122,6 @@ class GitDownloader extends VcsDownloader
 
                 case 'n':
                     throw new \RuntimeException('Update aborted');
-                    break;
 
                 case 'v':
                     $this->io->write($changes);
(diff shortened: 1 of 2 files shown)


MinsonLee:
Thanks~This solved a big problem for me.

PR #1407, opened 2012-12-11
Conformation to XDG Base Directory Specification

ntoniazzi:
A minor change to conform to the "XDG Base Directory Specification" (http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html - https://developer.gnome.org/basedir-spec/).


jrobeson:
there's an external library that already implements the xdg spec, is it possible to use that? https://github.com/dnoegel/php-xdg-base-dir/

ntoniazzi:
@alcohol It was not intended, and should be fixed now.

Seldaek:
It seems detecting XDG support is a clusterfuck. http://i.imgur.com/9B7qW1J.png / https://i.imgur.com/gk0FJab.png are on Fedora and Debian, on ubuntu server I only get XDG_SESSION_ID which doesn't even relate to the XDG Base Dir spec.

I would love to move this forward (at last..) but I am not sure how to best proceed.. Just assume XDG on non-windows? Check for `~/.config` and assume XDG then? Check all env vars for anything matching XDG and assume XDG?

@jrobeson I'd rather not add a dependency especially as it doesn't quite seem on par with what we need (cross platform and fallback to existing composer dirs, etc).

jrobeson:
XDG_SESSION_ID is unrelated. It's for systemd  user sessions. 

The environment variables XDG_DATA_HOME,XDG_DATA_DIRS, etc are used for overrides of the default configuration and thus not set by default and then XDG library basically checks for these environment variables and returns the appropriate fallbacks if they aren't set. It handles the differences to XDG_DATA_HOME, XDG_DATA_DIRS, etc automatically and prefers the appropriate one if set.

On Linux, we should assume .config , .cache , etc by default and create them if they don't exist. The OSX folks might have better defaults for their own directories, so I'm not sure what to do for them.

Many programs that adopt XDG after the fact tend to keep existing configurations/directories and use them if they exist, but use .config,.cache,etc on new installs.

jrobeson:
Here's how psysh initially implemented it: https://github.com/bobthecow/psysh/commit/52bce7ca23e7f2e4d7e02e663bb37a8b5e84b7c4

ntoniazzi:
On Windows and OSX, we don't change anything. On Linux, we forget about migrating the conf :
 * if a .composer directory exists, we keep the current behavior
 * if not, we just assume XDG and use [dnoegel/php-xdg-base-dir](https://github.com/dnoegel/php-xdg-base-dir/) to get the correct location.

marcinant:
Bump. Any progress on this issue?

johndrinkwater:
https://github.com/composer/composer/pull/1407#issuecomment-98746841 this seems the correct, though missing one caveat for people not using XDG, creating ~/.config/composer by default would be frustrating.
* if ~/.composer exists, use that
* if getHomeConfigDir() exists:
   * append "composer/" and use that
* else create ~/.composer & use that

https://github.com/composer/composer/pull/1407#issuecomment-97987321
> On Linux, we should assume .config , .cache , etc by default and create them if they don't exist.

```
john@joran ~/code/ ±:master❯ env | grep XDG_.*HOME
XDG_DATA_HOME=/home/john/settings/data
XDG_STATE_HOME=/home/john/settings/state
XDG_CONFIG_HOME=/home/john/settings/config
XDG_CACHE_HOME=/home/john/settings/cache
```

ntoniazzi:
On windows : no change
On other systems : 
* if ~/.composer exists, use it
* if XDG is active(1)
 * if XDG_CONFIG_HOME is set, append "composer" and use it
 * else create ".config/composer" and use it
* else create "~/.composer" and use it

(1) @johndrinkwater about the caveat: exact XDG detection does not seem possible. For now, any environment variable starting with "XDG_" will trigger it. Seems like libreoffice doesn't care and creates a ".config/libreoffice" directory as a fallback to XDG_CONFIG_HOME. Maybe we could do the same?

jrobeson:
I don't know the composer codebase.  Can someone tell me how this PR handles the storage for the composer rollback files. It's important that they don't end up in .config/composer

alcohol:
Rollback files end up in the Composer home directory currently.

ntoniazzi:
`data-dir` points to one of the following locations:
* `$XDG_DATA_HOME/composer`
* `~/.local/share/composer`
* `~/.composer`

Seldaek:
@ntoniazzi many thanks for the extreme patience here :) It's finally in!

I tried it out on ubuntu doing `mv ~/.composer ~/.config/composer && rm -rf ~/.config/composer/cache` is enough to trigger a switch over to XDG mode (assuming you have one of the XDG_ env vars set at least..).

steffenweber:
Seems to be working fine on Gentoo Linux, thank you!

EvanPurkhiser:
@ntoniazzi I'd like to extend a massive thank you for this. You're fighting the good fight to stop software from littering it's dotfiles in `$HOME` :-)

COLABORATI:
I just want to say thank you, too. One more piece of extra fiddling with backups gone away, thanks!

marcinant:
:+1:  Thank you!

stefanotorresi:
can anyone report if this change breaks existing `.travis.yml` cache directives?

i.e.
```yml
cache:
  directories:
    - $HOME/.composer/cache
```

Seldaek:
It might indeed, to be safe you should define COMPOSER_HOME as being `$HOME/.composer` - I am not sure if travis VMs pass our checks for enabling XDG..

stefanotorresi:
@Seldaek apparently it doesn't, so stuff still ends up in `$HOME/.composer` :sweat_smile:

alcohol:
XDG environment variables are not set on travis as they run no desktop environment by default (especially not on the container jobs). So there it should still default to the "old" convention/directories.

PR #1449, opened 2013-01-02
Dependency Graphing Command

felixalias:
Hi,
This pull request adds the functionality to composer to graph dependencies. It adds a new composer command, 'graph', which generates an HTML page at the specified path with a D3js powered adjacency matrix style graph.

Screenshot

### DIFF
--- a/src/Composer/Command/GraphCommand.php
+++ b/src/Composer/Command/GraphCommand.php
@@ -13,7 +13,6 @@
 namespace Composer\Command;
 
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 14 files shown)


curry684:
While I like the concept of the graphs I don't think this functionality belongs in the core composer project. Instead it can be generated by any external tool from a `composer.lock` file, or a plugin. It should not burden the core project. I would recommend building an add-on or companion project with this.

Seldaek:
Aye, closing this, would be lovely to have as an external tool or as a plugin once #3377 is merged.

PR #1607, opened 2013-02-21
added exclude function for composer autoload generation

rk3rn3r:
* added exclude function to composer / composer.json
* added exclude documentation to composer-schema.json
* added unit tests for the feature
* updated to latest master version

example:

    "autoload":
    {
        "classmap" : ["src", "vendor"],
        "exclude-from-classmap": ["/test/", "/tests/", "/Tests/", "/test-suite/"]
    }


klapifoch:
What happened with this feature?

andygrunwald:
@klapifoch nothing yet. We (@trivago / @rk3rn3r / me) are now responsible.
We will try to take care of this feature, rebase it against the current master and push the new version.

Just give us a few more days to solve this. Ok?
Thanks and you will get new information soon :)

andygrunwald:
Hey @stof, @Seldaek,
can you review it again? @msiebeneicher has updated this PR against the current master.
It would be cool if we can (finally) merge this feature. If you need anything else, please let us know.

msiebeneicher:
The state was tested against the unit tests and the test case https://gist.github.com/andygrunwald/4994295  by @andygrunwald with the following composer-json:

```
{
  "require": {
    "doctrine/doctrine-bundle": "v1.0.0"
  },
  "autoload": {
    "exclude-from-classmap": ["/Tests/"]
  }
}
```

Result: 
* 790 classes in classmap with current composer version
* 738 classes in classmap after this change

klapifoch:
Not merged yet?

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -182,14 +189,9 @@ public static function autoload(\$class)
                         if (!is_dir($dir)) {
                             continue;
                         }
-                        $whitelist = sprintf(
-                            '{%s/%s.+(?<!(?<!/)Test\.php)$}',
-                            preg_quote($dir),
-                            ($psrType === 'psr-0' && strpos($namespace, '_') === false) ? preg_quote(strtr($namespace, '\\', '/')) : ''
-                        );
```
Seldaek: As far as I can see the removal of this block still has not been accounted for..

The problem is such, if you use psr-0, right now the dir is usually src/ and the namespace prefix `Foo\Bar`. So it will build up a classmap for all of src/ but exclude files not matching the whitelist (i.e. anything not in `src/Foo/Bar`).

Removing the whitelist is ok but then I think the $dir should be modified to have the namespace prefix appended if `$psrType === 'psr-0' && strpos($namespace, '_') === false` is true, that way we only scan the relevant files.

Seldaek:
@andygrunwald @rk3rn3r Another thing that's missing for an easy merge is including it in the `doc/04-schema.md` in the autoload docs.

Unrelated: @klapifoch can you be more of an obnoxious and ungrateful person please?

msiebeneicher:
@Seldaek If have updated the doc/04-schema.md and integrate a desciption.

I checked the whitelist handling again and i'm not sure if the logic make also sense for the blacklist handling: 

With the "exclude-from-blacklist" you are free to define pattern to avoid classes in the classmap. If we extend the "user pattern" with the directories (and namespaces for psr-0), like the whitelist functionality, it's possbile to produce wrong or invalid patterns (if somebody use "^ - Start of string or line" for example).

klapifoch:
@Seldaek It wasn't my intention to be annoying, it is just a feature i want to use in a project.
Sorry.

msiebeneicher:
@Seldaek: We reintegrated the current master again to our fork.

andygrunwald:
Any update here?

Seldaek:
I guess 2.5y is a pretty long time, sorry about that, but we got there :) Finally took the time to merge this and try it out, tweaked things a bit because it was matching stuff a bit too greedily for my taste. Now you have to specify exact paths from package root to things you want to exclude, I see that it's not always the easiest think it's better that way rather than accidentally matching too much. We can always add support for wildcards or such later on if really needed. Thanks @rk3rn3r

Seldaek:
So I added support for wildcards after all, see https://github.com/composer/composer/commit/6c165107433b7b6c1f177a36fc5fcfe1b198f6fd

Tobion:
So when exluding test files from the autoload optimization, it means test classes might not be available anymore. So it is similar to using exclusion via gitattributes. But IMO using gitattributes is the better approach then as this will actually also make the package considerably smaller. And developers still have a choice with --prefer-dist vs --prefer-source.
See https://github.com/symfony/symfony/pull/16397#issuecomment-152684271

Seldaek:
No, that is only true if you rely on classmap autoloading. It you use psr0
or 4 plus optimized classmap dumping then what is not found in the classmap
is still found using psr rules. Unless you use the classmap authoritative
flag but that's something you really just should do in prod where tests
don't matter IMO.

Issue #1610, opened 2013-02-22
EC2 Possible SSL Errors


mtotheikle:
Running a update from an EC2 box I get the following:

The "https://packagist.org/packages.json" file could not be downloaded: SSL: crypto enabling timeout  
  Failed to enable crypto  
  failed to open stream: operation failed   

  [Composer\Downloader\TransportException]  
  The "https://packagist.org/packages.json" file could not be downloaded: SSL: crypto enabling timeout  
  Failed to enable crypto  
  failed to open stream: operation failed  

Anyone else experiencing this? I have seen some bug reports about a possible OpenSSL bug, but nothing I have found so far has led to a fix. Could there be a config option to not use SSL for the package files?

PR #1628, opened 2013-02-27
Added cURL support as a Driver

flavioheleno:
The work is pretty rough, but is working.
It still lack of options to choose between file_get_contents driver (aka. stream driver) or curl (aka. curl driver).

### DIFF
--- a/src/Composer/Util/CurlDriver.php
+++ b/src/Composer/Util/CurlDriver.php
@@ -57,9 +57,18 @@ class CurlDriver
             return $fd->get($originUrl, $fileUrl, $additionalOptions, $progress);
         }
         $options = $this->getOptionsForUrl($originUrl, $additionalOptions);
+        if (isset($options['github-token'])) {
+            $fileUrl .= (false === strpos($fileUrl, '?') ? '?' : '&') . 'access_token='.$options['github-token'];
+            unset($options['github-token']);
+        }
         $options[CURLOPT_URL] = $fileUrl;
+        echo "{$fileUrl}\n";
         curl_setopt_array($this->curl, $options);
-        return curl_exec($this->curl);
+        $result = curl_exec($this->curl);
+        if ($result === false) {
+        	throw new TransportException(curl_error($this->curl), curl_errno($this->curl));
+        }
+        return $result;
     }
 
     protected function headerCallback()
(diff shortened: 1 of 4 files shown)


XedinUnknown:
Is it true that these changes could resolve the problem with pre-5.5 PHP versions and SSL, like it says [here](https://github.com/composer/composer/issues/1610#issuecomment-15032964)? If so, what is the status on this 4 year old PR? Perhaps this is done elsewhere? The issue with SSL is horrible, many PHP 5.3 or 5.4 environment can build without hacks, it appears.

alcohol:
PHP 5.3 and 5.4 are dead.

XedinUnknown:
PHP 5.3 - 5.5 are all dead. But that doesn't mean they're not used. Composer supports PHP 5.3. I try to push all WP developers to use Composer, and with Bedrock there's limited success. But if it's impossible to use Composer normally with pre-PHP5 versions, the whole WP community will be left behind. And believe me, I know that the WP community is full of mediocrity and really bad code and practices. But I really hope that this will change. Anyhow, usually it doesn't take much effort to keep software PHP 5.4+ compatible, and I would really like to be able to test on legacy versions too. Alas, with the SSH problems, building on PHP 5.4 is problematic.

alcohol:
> Is it true that these changes could resolve the problem with pre-5.5 PHP versions and SSL, like it says here?

There is no actual data that suggests this to be true.

> If so, what is the status on this 4 year old PR?

It is out of sync with the repository. Also, adding cURL support has always been a controversial topic. It would be yet another PHP extension that users would have to install (though admittedly, it very often already is). Alternatively, it could be implemented in such a way that it would only be used if possible (based on presence of extension).

> But that doesn't mean they're not used.

We are not here to encourage bad behaviour. 5.3+ support is only available because that commitment was made when 1.x was originally developed. We are not building new features or functionality just to add additional support for PHP versions that are already EOL.

> But if it's impossible to use Composer normally with pre-PHP7 versions, the whole WP community will be left behind.

Drifting off-topic here, but it is perfectly doable to develop on a modern PHP version but still write code that is compatible with older PHP versions. Merely specify in your `composer.json` that your target platform is version `5.3` for example, and Composer will install dependencies as if you are running that version, even if your development machine would be running PHP `7.1` in actuality.

jduartedj:
Curl support is a must, please add it!

XedinUnknown:
> Merely specify in your composer.json that your target platform is version 5.3 for example, and Composer will install dependencies as if you are running that version, even if your development machine would be running PHP 7.1 in actuality. No more HTTPS issues.

That's what we've been doing. However, having to add and remove that config all the time is tedious and error prone.

alcohol:
> having to add and remove that config all the time is tedious and error prone

Why remove it?

jduartedj:
cUrl would solve a lot of problems, HTTPS issues (and optional config), proxy support, issues with  allow_url_fopen = false (where you can't change it) and overall security improvement. Downside is that your are trading a native solution for a plugin but in all fairness it is widely used anyway, I'm amazed it wasn't integrated into PHP 7+ as a native function!

stof:
@jduartedj almost all builtin parts of PHP are actually built using PHP extensions. So having curl being a PHP extension makes perfect sense. Note that this extension is part of the PHP codebase directly, not a pecl extension.

Seldaek:
Closed by https://github.com/composer/composer/pull/7904 which adds curl-multi support for the upcoming 2.0

PR #1734, opened 2013-03-26
Added support for additional Script parameters to install and update commands

ramondelafuente:
This pull request is an attempt at allowing scripts to have parameters defined at runtime. 
An example would be to use  >composer install --script-tag="env:--env=prod" on production.

Scripts would then be configured in composer.json with "myscript {env}".

PHP Callbacks would have access to the parameters through the Event object:
$event->getScriptParams();

There are some decisions that need review in my opinion:
-- Is it desirable to add the scriptParams to the Event constructor?
-- Would the InstallCommand and UpdateCommand need more explanation?
-- Is the help page correctly altered?
-- Are there tests needed? I could not find a good place to add tests for this...

### DIFF
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -48,6 +48,11 @@ For any given event:
 and command-line executables commands.
 - PHP classes containing defined callbacks must be autoloadable via Composer's
 autoload functionality.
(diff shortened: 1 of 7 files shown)


ramondelafuente:
I think we can safely put this one to rest :-)

PR #1857, opened 2013-05-03
Modify classmap

superdweebie:
EDITED to reflect recent commits

This pr does two things:

* It modifies the `ClassLoader` so that it keeps record of changes to it's `classmap` and can report if the classmap has been modified. It also allows a callback to be set which fires on `__destruct`.

* It extends the `autoload_real` generated by `autoload-dump` if a `--dynamic-optimize` flag is passed. `autoload_real` will then set a destruct callback on the Loader which will rewrite the classmap if the classmap has been modified, ready for the next time the loader is used (eg next request).

Why? The optimize flag for `autoload-dump` is good, but when using a framework such as zf2 or symfony, the dumped classmap can easily have 3000+ items. Most of these are not used, it is very rare for an app to use _all_ of a framework. This http://patrickallaert.blogspot.com.au/2013/03/benchmarking-composer-autoloading.html showed that loading a huge classmap can actually be a hindrance to performance. This pr allows the classmap to grow dynamically to only include those classes that are actually used by an app.

In my own testing on a zf2 app with doctrine, this is clearly faster than using the `optimize` flag.


Seldaek:
Closing as https://github.com/composer/composer/pull/5174 basically solves the classmap cache/loading performance.

PR #2232, opened 2013-09-05
Fix for custom install directory can not remove when uninstall the package

breath-co2:
Fix for custom install directory can not remove when uninstall the package

### DIFF
--- a/src/Composer/Installer/LibraryInstaller.php
+++ b/src/Composer/Installer/LibraryInstaller.php
@@ -130,6 +130,21 @@ class LibraryInstaller implements InstallerInterface
                 @rmdir($packageVendorDir);
             }
         }
+
+        $downloadPath = $this->getInstallPath($package);
+        while (true) {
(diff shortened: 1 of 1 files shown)


bilhackmac:
This PR (as #2637) need to be merged.

Without these changes all uninstalled packages that depend on `composer/installer` or any custom installer was not removed.

This bug was introduce by 661df121d9687fcebe76897e8efc4bc279613502 that make, I think, BC !

Seldaek:
Finally fixed this as getPackageBasePath now uses getInstallPath so installer plugins only overriding that should work fine now.

bilhackmac:
It's work like a charm. Thanks. :+1:

redthor:
I think I have an problem due to this change...

https://github.com/composer/composer/commit/d98b134dc3ef7d5bd4b6e37c9621a811f8f1599d#commitcomment-14518456

PR #2311, opened 2013-10-11
Add flag/option to skip autoload generation

mpdude:
I'd like to suggest to add a flag/switch to skip autoload code generation during the "install". When using this switch, one would have to "dump-autoload" later on, of course.

Reason:

I am using composer in a multistage build process. At the time "install" is run, not all directories to be included in the autoload scanning process have been generated or are available (some of them are the result of code generation tools that will be run later on).

Would you accept a patch for that?


mpdude:
This has been implemented lately in #3453

PR #2404, opened 2013-11-07
Status command

chEbba:
```
Usage:
 dep-status [--dev] [-p|--problem] [package]

Arguments:
 package               Show concrete package status.

Options:
 --dev                 Show dev requirements.
 --problem (-p)        Show only problem (missed or outdated) dependencies.

Help:
 The dep-status command display status of root package requirements:
 what packages are missed, changed or outdated; what package versions are installed
```

Example output:
```
chebba@chembp:~/Projects/example$ composer dep-status -p
foo/bar
status     : missed
dev        : no
required   : ~1.3.5

symfony/symfony
status     : outdated
dev        : no
required   : 2.3.*
installed  : v2.1.4
path       : /Users/chebba/Projects/example/vendor/symfony/symfony
```

Notes:
* For now this command does not return error statuses.
* I think we can change curernt "status" command for this one because it is more relative for this name, and rename current one with more logical name like "changes", "local-changes", "modify-status", etc.

Any comments and additions are welcome

### DIFF
--- a/src/Composer/Command/DependencyStatusCommand.php
+++ b/src/Composer/Command/DependencyStatusCommand.php
@@ -1,0 +1,222 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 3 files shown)


Seldaek:
We now have `composer show` showing current versions, `composer status` showing local changes and `composer outdated` showing available updates.

PR #2424, opened 2013-11-14
Added API for an install path in installable repository

### LINKED ISSUE #2174: "API" for install path
Disclaimer: This issue is not intended for an end-users. Instead it focuses on tooling support for composer (such as the [Composer-Eclipse-Plugin](https://github.com/pulse00/Composer-Eclipse-Plugin))

When packages installed, they are typically installed in the vendor directory. Though especially custom-installers change the path. One of the main features of the Eclipse-Composer-Plugin is to automatically manage the build-path, so that code-assist works seamlessly without any further user-interaction. For now, packages installed with a custom install path can't be automatically managed.

For these cases an install path "API" (maybe be another property in one of the json-files) would be handy and welcome to provide tooling support.

Thanks
gossi

hason:
Fixes #2174

### DIFF
--- a/src/Composer/Repository/InstalledFilesystemRepository.php
+++ b/src/Composer/Repository/InstalledFilesystemRepository.php
@@ -12,6 +12,12 @@
 
 namespace Composer\Repository;
 
+use Composer\Json\JsonFile;
+use Composer\Package\PackageInterface;
+use Composer\Package\Loader\ArrayLoader;
+use Composer\Package\Dumper\ArrayDumper;
+use Composer\Util\Filesystem;
+
 /**
  * Installed filesystem repository.
  *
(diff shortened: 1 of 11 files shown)


gossi:
At which place to find the `install-path` for each package (I'm sorry for asking I'm not into the composer api).

Seldaek:
This is the key part of how to access the install path for a package instance, I am not sure at which level you are working, but I don't think they persist to the installed.json in this PR.

```
+                            $repository = $composer->getRepositoryManager()->getLocalRepository();
+                            if ($repository instanceof InstalledRepositoryInterface && $repository->hasPackage($package)) {
+                                $path = $repository->getInstallPath($package);
```

gossi:
I'm consuming `installed.json` if it's in there it's the perfect place for composer and for me to read the desired information :smiley:

hason:
This PR is updated.

src/Composer/Repository/InstalledFilesystemRepository.php
```diff
@@ -19,4 +25,106 @@
...
+    {
+        $data = array();
+        $dumper = new ArrayDumper();
+        $fileDir = dirname($this->fs->normalizePath($this->file->getPath()));
+
+        foreach ($this->getCanonicalPackages() as $package) {
+            $path = $this->getInstallPath($package);
+            $data[] = array(
+                'package' => $dumper->dump($package),
+                'install-path' => null !== $path ? $this->fs->findShortestPath($fileDir, $path, true) : null,
```
gossi: This would break the functionality of tools that rely on the current structure of the dump. It would be better to put the install-path into the `$package` and dump it the way it currently is.

```php
// some pseudo-code:
foreach ($this->getCanonicalPackages() as $package) {
    $path = $this->getInstallPath($package);
    $package['installPath'] = null !== $path ? $this->fs->findShortestPath($fileDir, $path, true) : null;
    $data[] = $dumper->dump($package);
}
```

And then let the dumper transform `installPath` into `install-path`.

Seldaek:
Finally had a good look at this (about time? :P) anyway decided to take the key bits of it and do a minimalistic patch to avoid spreading this too much, thanks @hason though for the inspiration (and patience..) - see https://github.com/composer/composer/commit/63da7c6b2d164cf264cfe456780009ea5119d4f9

Issue #2038, opened 2013-06-25
more commands to modify composer.json from the CLI


lsmith77:
When I run the tests for https://github.com/jackalope/jackalope/blob/master/composer.json I would prefer if I could make either of the following adjustments via the CLI:
 * remove the require on "jackalope/jackalope-transport"
 * add a provide on "jackalope/jackalope-transport"

For the first I envision we could add a flag to the require command to remove rather than add. For provide we could add an analogous command as we have for require. We might also want to have something like that for suggest just for good measure?

dantleech:
As #2479 is merged can this be closed now?

Seldaek:
Yup I don't think provide/replace/conflict/suggest are hardly common enough to have their own commands to set/unset, but require/remove does the job.

Issue #1622, opened 2013-02-27
Unrequire command in cli.


ahilles107:
I need unrequire command on composer cli, or info how to extort package removing from vendor/ directory and fire uninstall composer event.

Seldaek:
See `composer remove`

PR #2479, opened 2013-12-04
Add remove command to remove a package from the list of current installed packages

pierredup:
Hi All

I have added a new command to be able to remove a package from the current list of installed packaged.
The command will remove the name from either the 'require' or 'require-dev' section of the composer.json, then run an update that will remove the installed package (removing the actual files and updating composer.lock).

Related: #1622, #2038


salarmehr:
the `remove` command does not works for the global packages.

Seldaek:
@salarmehr works for me with `composer global remove ...`

PR #2493, opened 2013-12-10
[alias] RootAliasPackage manages the requirements correctly

vatson:
The data integrity of requirements in RootAliasPackage is broken. 

Let's assume, we have an `aliased` package:

```json
{
    "name": "my/package",
    "description": "My best package",
    "require": {
        "php": ">=5.3.3",
        "symfony/symfony": "2.3.6"
    },
    "extra": {
        "branch-alias": {
            "dev-master": "0.1-dev"
        }
    }
}
```

If we want to dynamically install a new package

```php
use Composer\Factory;
use Composer\IO\BufferIO;
use Composer\Installer;

$newPackage = array('my/new-package' => '1.0');
$io = new BufferIO;

$composer = Factory::create($io);
$composer->getRootPackage()->setRequires(array_merge(
  $composer->getRootPackage()->getRequires(),
  $newPackage
));

$installer = Installer::create($io, $composer);
$installer
    ->setDryRun(false)
    ->setVerbose(true)
    ->setPreferSource(false)
    ->setPreferDist(true)
    ->setDevMode(false)
    ->setRunScripts(true)
    ->setUpdate(true)
    ->setUpdateWhitelist(array('my/new-package'))
    ->setOptimizeAutoloader(true)
    ->run()
;
```
we get an unexpected result. Composer will install nothing. The issue is that `Installer` uses `RootPackage` to compute the difference between already installed and new packages. But in the case of  `aliased` package, installer deals with a `RootAliasPackage`. RootAliasPackage package is able to set new requirements for original RootPackage, but this changes won't affect itself (an immutable property). So Installer works with outdated set of requirements.

This PR fixes the issue described above.

I didn't add the tests cuz couldn't find a test which should be edited. But all other passed successfully

### DIFF
--- a/src/Composer/Package/AliasPackage.php
+++ b/src/Composer/Package/AliasPackage.php
@@ -28,6 +28,7 @@ class AliasPackage extends BasePackage implements CompletePackageInterface
     protected $stability;
 
     protected $requires;
+    protected $devRequires;
     protected $conflicts;
     protected $provides;
     protected $replaces;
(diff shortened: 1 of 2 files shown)


Seldaek:
I fixed the lost change cited above and merged.. I had to rebase though so github is not picking up the merge. Better late than never I guess so thanks ;)

PR #2509, opened 2013-12-16
Ensure detected extensions have valid package names

### LINKED ISSUE #2489: Unable to depend on opcache
```
> php -v
PHP 5.5.6 (cli) (built: Nov 22 2013 22:44:00)
Copyright (c) 1997-2013 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2013 Zend Technologies
    with Zend OPcache v7.0.2, Copyright (c) 1999-2013, by Zend Technologies
    with Xdebug v2.2.3, Copyright (c) 2002-2013, by Derick Rethans
```

```
> ﻿﻿composer show --platform
# ...
ext-Zend OPcache    0         The Zend OPcache PHP extension
# ...
```

```
> cat composer.json
# ...
    "require": {
        "ext-Zend OPcache": "*"
    },
# ...
```

```
> composer validate
./composer.json is valid, but with a few warnings
See http://getcomposer.org/doc/04-schema.md for details on the schema
require.ext-Zend OPcache : invalid key, package names must be strings containing only [A-Za-z0-9_./-]
```

This then causes packagist to reject the package.

Please let me know if this is something relatively easy to fix, and if so which bit of the code - I'm happy to submit a PR for this.

glenjamin:
Fixes #2489

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -70,7 +70,8 @@ class PlatformRepository extends ArrayRepository
                 $version = $versionParser->normalize($prettyVersion);
             }
 
-            $ext = new CompletePackage('ext-'.$name, $version, $prettyVersion);
+            $packageName = $this->buildPackageName($name);
+            $ext = new CompletePackage($packageName, $version, $prettyVersion);
             $ext->setDescription('The '.$name.' PHP extension');
             parent::addPackage($ext);
         }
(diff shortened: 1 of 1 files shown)


Seldaek:
If you want to know what extension to require you can use `composer show --platform` to see which names it picks up. That's probably the best way to avoid guessing.

szepeviktor:
Thank you.

I was referring to the string replacement: " " -> "-" in the linked PR.
It may work in some parts of composer but not in others that is why it says "wrong version" instead of missing dependency.

szepeviktor:
Thanks.
So this commit https://github.com/composer/composer/commit/0e9a9a3875c317d610d6a995d94dd598b7bececd in this PR may not be necessary.

Seldaek:
Yes it is as it forces normalization to a given format.. and we can't change this anymore.

PR #2585, opened 2014-01-13
added ZipArchiver to actually compress zip files

jaypea:
the PharArchiver used to create zip files does not compress anything currently. 
when using satis with an archive directory, this new ZipArchiver saves a ton of hard disk space when the zip extension is enabled. 
fallback is the current behaviour using PharArchiver.

### DIFF
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -386,6 +386,7 @@ class Factory
         }
 
         $am = new Archiver\ArchiveManager($dm);
+        $am->addArchiver(new Archiver\ZipArchiver);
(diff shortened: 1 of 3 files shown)


andrerom:
@glye Regarding missing executable flags on archives made with PharArchiver *(with compression)*, maybe this would help?

staabm:
//cc @alcohol

alcohol:
@Seldaek was there anything at the time blocking this from being merged? Needs more review?

Seldaek:
Seems fine I think.. mostly it just got forgotten. Thanks @jaypea!

PR #2604, opened 2014-01-16
Add option --repository-options to create-project command

### LINKED ISSUE #2920: How to use create-project with private repositories?
I've implemented a Satis repository and I'm trying to use `create-project` to startup a new system based on a skeleton package from Satis.

I've added the repository to my `~/.composer/composer.json` file, and thus `composer global search myself/skeleton` works fine and it can find the given package, even require and install it as well. That repository currently has two tags (1.0 and 1.0.1), those are showing fine in the Satis page.

However, `composer global create-project -s dev myself/skeleton` does not work. I've tried using different stability and version flags, and all those return a similar error: "Could not find package myself/skeleton with version 1.0.1", or stability etc.

How is the best way to use `create-project` with private repositories? Is there any way to use the command without issuing the `global` one first? Finally, is it a bad behaviour adding the repository in the user's `composer.json`?

max-voloshin:
With ``--repository-url`` in some cases I want specify also custom options.

I have tested this with #2189, but this PR can be merged separately.

I have left some BC code, that can be safely removed if Composer maintainers allow to do this.

Docs was updated.

Fixes #2920

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -359,6 +359,7 @@ By default the command checks for the packages on packagist.org.
 * **--repository-url:** Provide a custom repository to search for the package,
   which will be used instead of packagist. Can be either an HTTP URL pointing
   to a `composer` repository, or a path to a local `packages.json` file.
+* **--repository-options:** Provide options(JSON format is used) for custom repository. This option does not make sense without **--repository-url**.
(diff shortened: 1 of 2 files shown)


lainosantos:
:+1:

max-voloshin:
WOOHOO! :)

Issue #704, opened 2012-05-17
Replace remote file_get_contents() calls with cURL


gergoerdosi:
I found two places in the code where file_get_contents() is used with remote URLs. These are:
- Composer\Repository\ComposerRepository:: notifyInstall()
- Composer\Util\RemoteFilesystem::get()

On many servers allow_url_fopen is disabled for security reasons. A solution could be to replace these file_get_contents() calls with cURL. This would make cURL as a requirement, but most of the servers have it installed.

karser:
I encountered the same issue.
The problem was in my CLI php.ini, option default_socket_timeout = -1.
I changed it to default value (60) and it works now.

digitalkaoz:
i would vote for `egeloen/http-adapter` which handles all that and can guess the most suitable driver

khaefeli:
+1

passioncoder:
3 years (and 2 PRs) later, allow_url_fopen is still a bad practise. 

+1 for a cURL option.

snetty:
yup, still can't use it on my live environment, no choice but to commit everything to repo just because composer devs cba to merge in a PR.

pmlt:
+1 for this fix. We are using a CI tool to package our dependencies and we are not willing to enable allow_url_fopen on that server. It is indeed bad practice. Is there an ETA for this fix? It seems to be a rather trivial change no?

Seldaek:
It's not quite trivial no as we do support quite a lot of things on the download level, and file_get_contents works everywhere (unless you disable it) while curl is hardly enabled everywhere. We'll get there eventually I am sure but in the meantime I'd suggest you run composer with allow_url_fopen enabled, you don't have to enable it for the whole server mind you you can just invoke `php -d allow_url_fopen=1 composer install` or such.

vBZachery:
cURL is considerably more ubiquitous than allow_url_fopen being enabled in webhosting environments, especially when it comes to shared hosting. This is based on nearly 11 years of supporting a major PHP software package (vBulletin).

There isn't a good reason not to have cURL as an option, and still fall back to file_get_contents as a backup.

maxiwheat:
+1 on this one, `allow_url_fopen`is disabled here for security reasons. Would be great for composer to support `curl` and `file_get_contents`

DIOHz0r:
is this still on a pr?. I'm in a hosting server with allow_url_fopen disabled and I can't edit it 
:(

siteleas:
[...]

PR #2696, opened 2014-02-12
Curl transfer

peteraba:
Added curl fallback for data transfer. Moved hard coded stream context calls into separate transfer class. Created interface for transfer classes. Added tests.

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -60,19 +60,25 @@ EOT
         $output->write('Checking http connectivity: ');
         $this->outputResult($output, $this->checkHttp());
 
-        $opts = stream_context_get_options(StreamContextFactory::getContext('http://example.org'));
-        if (!empty($opts['http']['proxy'])) {
-            $output->write('Checking HTTP proxy: ');
-            $this->outputResult($output, $this->checkHttpProxy());
-            $output->write('Checking HTTP proxy support for request_fulluri: ');
-            $this->outputResult($output, $this->checkHttpProxyFullUriRequestParam());
-            $output->write('Checking HTTPS proxy support for request_fulluri: ');
-            $this->outputResult($output, $this->checkHttpsProxyFullUriRequestParam());
+        if (ini_get('allow_url_fopen') || !extension_loaded('curl')) {
+            $opts = stream_context_get_options(StreamContextFactory::getContext('http://example.org'));
+            if (!empty($opts['http']['proxy'])) {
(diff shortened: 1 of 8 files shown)


wzrdtales:
+1

curl **MUST** be preferred over file_get_contents and not the other way around. In this current implementation of this PR it is the other way around. When curl is used over file_get_contents instead this PR is going to solve (if not all proxy isses) the following issues:

#4332
#704
#4701 

Thus to add: @Seldaek This PR is not a `Nice To Have`, it is a must and is also a bug fix of the buggy file_get_contents method. 

I needed to fork composer and we will use only our fork now instead and replace the transfer with curl as default, otherwise composer won't be able to download at all and falls back always to the slower git clone method.

staabm:
@wzrdtales feel free to discuss and provide a PR for a possible curl integration..

wzrdtales:
@staabm I was about to do that, but as there is already a PR for a curl implementation (this one here) I walked in here first instead of directly a new one.

peteraba:
@staabm @wzrdtales I'm no longer part of the organization under which this PR was created, although we did use it intensively within the organization a while back. Feel free to pick it up, it's mostly abandoned anyway.

wzrdtales:
I just saw the work of https://github.com/hirak/prestissimo/issues/67, glad I do not have to spend any time on this. Going to use his work, as it is also faster with the prefetching. Would recommend anyone to do the same even those who do not have problems with the fopen methods.

avindra:
Thanks to @hirak and everyone else working on the curl replacement / killing off fopen calls to urls. I am waiting to test to see if it fixes two of our problems:

1) This error thanks to a new security requirement at our company:

```
The "https://packagist.org/packages.json" file could not be downloaded: allow_url_fopen must be enabled in php.ini (https:// wrapper is disabled in the server configuration by allow_url_fopen=0
failed to open stream: no suitable wrapper could be found)
```

2) Issue #4332 

If there are builds I can use to test, please let me know, I'd be glad to.

avindra:
As a follow up, after installing `hirak/prestissimo`, not only are downloads faster, but they work with the proxy setup going on at my job. (I.e., it fixes Issue #4332). However, since prestissimo seems to only override how packages are fetched, commands like `composer outdated` still don't work if `allow_url_fopen` is disabled

jduartedj:
Just did a fresh install and my server allows **curl** but not **allow_url_fopen**.

I got the **allow_url_fopen** error and was not able to proceed.

How can I install and use composer?

peteraba:
Not even working with PHP much anymore. PR should have been closed a long time ago.

PR #2706, opened 2014-02-15
Support tar.gz and tar.bz2 compressed archiver

hfcorriez:
Sometimes someone need tar.gz package to deploy the project or host the package.

### DIFF
--- a/src/Composer/Package/Archiver/PharArchiver.php
+++ b/src/Composer/Package/Archiver/PharArchiver.php
@@ -20,8 +20,15 @@ namespace Composer\Package\Archiver;
 class PharArchiver implements ArchiverInterface
 {
     protected static $formats = array(
-        'zip' => \Phar::ZIP,
-        'tar' => \Phar::TAR,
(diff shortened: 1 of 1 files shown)


andrerom:
+1 would be nice addition, and as it it is different suffix then uncompressed files this does not have the ambiguity as the recently closed proposal for zip compression support had. /cc @Seldaek

andrerom:
Related: zip compression is now possible when zip extension is enabled as of #2585

Seldaek:
Yup fair enough, sorry for the delay looking at this @hfcorriez, and thanks for the PR :)

hfcorriez:
It's OK, thanks for your job. I am using node as my main language now. I follow the PHP too.



On Mon, Feb 8, 2016 at 12:35 PM -0800, "Jordi Boggiano" <notifications@github.com> wrote:












Yup fair enough, sorry for the delay looking at this @hfcorriez, and thanks for the PR :)

PR #2745, opened 2014-02-23
[WIP] SSL/TLS Support

padraic:
This is a work in progress! Please do review and comment.

This PR is designed to do a few things:
1. The RemoteFilesystem class configures itself with default TLS options in line with PHP 5.6's adopted values if TLS protection has NOT been explicitly disabled by the user. This means Composer does everything possible to do HTTPS right - and lets the user shoulder the risk of disabling it. All instances are now created using Composer\Factory::createRemoteFilesystem().
2. Where any TLS option needs user input, e.g. a CA certificate file, we give them an error. They can now set a cafile option in config.json or pass a cafile option with value on CLI. If they can't/won't do that, they can resort to setting a disable-tls option in config.json or via CLI at their own risk.
3. Composer will distribute a cacert.pem file. This is used as a last resort option wherever possible and will be updated alongside Composer self-updates. There will need to be a process introduced to regularly check/update this file in the git repository.

This should resolve the TLS portion of https://github.com/composer/composer/issues/1074.

The Composer Installer is also being updated: https://github.com/composer/getcomposer.org/pull/61

TODO:
1. Most of the TLS configuration is unnecessary under PHP 5.6 since it does all this automatically from that version.
2. Documentation updates.

I'm hoping to complete the PR over the next week - but feel free to PR my fork (the tls-config branch).


padraic:
I expect it's broken in Travis, but I did an initial update to current master. Will need to fix issues and review comments.

Seldaek:
Closing in favor of #4759 - I updated the branch to work with master, reviewed everything, updated things so it works well and also backported the changes we did in the installer to limit breakage.

If you can give it a last review that'd be great, but anyway I think I'll merge next few days and hope it doesn't wreak too much havoc. /cc @StormTide @padraic the most critical to look at really is https://github.com/composer/composer/compare/master...Seldaek:tls-config?expand=1#diff-628bd40c5519cc043eb3cbb81aa18641R33 - note that I shelved the --disable-tls/--cafile args, IMO if people want to do this it's best to do it in the config file than every time they run the command, and it keeps the code a bit leaner. It is still possible to disable TLS if you really want to, but I tried to also tweak wording to encourage people to enable openssl rather than just saying "openssl is missing, if you want just do this to disable the error" because this sort of thing leads to people just disabling to move on with their day.

StormTide:
I made one note where you're retrying the CN with what appears to be the CN from the certificate (which the attacker controls)... Otherwise I dont see anything at issue here.

PR #2814, opened 2014-03-16
[WIP] Developer Package Signing

padraic:
This PR is a work in progress. Expect changes and check the TODO list. Review/testing is welcome.

This pull request is intended to add support for developer-level package signing to Composer using openssl. It includes a "create-keys" command to generate an RSA private and public keys (the private key should be encrypted using a passphrase), an "add-dev-key" command which imports a private key and extracts the public key for distribution in a package within keys.json, and a "sign" command which assembles a manifest of files, their checksums, and lengths, and signs the manifest using a private key within manifest.json. This PR does not address any form of package or metadata signing by Packagist which is also required for repository-level security. This PR does not depend on SSL/TLS support.

TODO:
1. ADD support for Composer to cache keys.json data since it should only be populated on first encounter so that attacks to alter keys.json have a chance to be detected.
2. ADD support for Composer to verify package signatures as part of its install/update routines.
3. ADD verification status to install/update output.
4. ADD composer command options to reject unverified packages (I'll make it optional but it's strongly suggested that this become enabled by default in time).

You can currently test sign packages using the following commands (the first three are only required to setup keys):

```
php composer.phar create-keys --prefix=myUniquePrefix

php composer.phar add-dev-key ./path/to/myUniquePrefix-private.pem

php composer.phar sign-dev-keys ./path/to/myUniquePrefix-private.pem
```

```
php composer.phar sign ./path/to/myUniquePrefix-private.pem
```

This should result in two new files: keys.json (used to distribute authorised public keys) and manifest.json (a signed listing of all package files, their SHA256 checksums and lengths).

The theory is to sign packages and commit keys/manifest changes as the FINAL step before tagging a release. If you make any other changes, or altered the tag after signing, then the signatures could never be verified by users.

### DIFF
--- a/src/Composer/Command/AddDevKeyCommand.php
+++ b/src/Composer/Command/AddDevKeyCommand.php
@@ -37,13 +37,16 @@ class AddDevKeyCommand extends Command
             ->setDefinition(array(
                 new InputOption('threshold', 't', InputOption::VALUE_REQUIRED, 'Sets the threshold number of signatures required to verify this file (Default: 1)', 1),
                 new InputOption('expires', 'e', InputOption::VALUE_REQUIRED, 'Set an expiry date beyond which the keys.json file will be deemed untrusted', null),
+                new InputOption('root', 'r', InputOption::VALUE_NONE, 'Include the given key as a root signer, i.e. it may sign off on future key changes.'),
                 new InputArgument('private-key', InputArgument::REQUIRED, 'Path to the private key which will be used to sign the package'),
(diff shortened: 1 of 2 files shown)


mbrodala:
@padraic What would be the usual workflow here? I see no change to the existing code, thus signing and verification would be optional and require manual work? If this is integrated, it should be possible to enforce signature verification by default.

GrahamCampbell:
Is this PR dead?

PR #1882, opened 2013-05-08
[New Command] List of suggests

nkt:
For composer project this command has output:

``` sh
$ bin/composer suggests -D
Suggests:
  Nothing!
Dev suggests:
  phpunit/php-code-coverage:
    "ext-dom": "*"
    "ext-xdebug": ">=2.0.5"
  phpunit/phpunit:
    "ext-json": "*"
    "ext-simplexml": "*"
    "ext-tokenizer": "*"
    "phpunit/php-invoker": ">=1.1.0,<1.2.0"
  phpunit/phpunit-mock-objects:
    "ext-soap": "*"
```

PR #2819, opened 2014-03-17
[Command] Add suggests command

nkt:
Rebased #1882

### DIFF
--- a/src/Composer/Command/SuggestsCommand.php
+++ b/src/Composer/Command/SuggestsCommand.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Command;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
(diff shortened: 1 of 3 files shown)


nkt:
@Seldaek any updates on this?

src/Composer/Command/SuggestsCommand.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @author Gusakov Nikita <dev@nkt.me>
+ */
+class SuggestsCommand extends Command
+{
+    protected function configure()
+    {
+        $this
+            ->setName('suggests')
+            ->setDescription('Show packages suggests')
```
stof: I would say ``Show package suggestions``

src/Composer/Command/SuggestsCommand.php
```diff
@@ -0,0 +1,59 @@
...
+            ->setDefinition(array(
+                new InputOption('dev', null, InputOption::VALUE_NONE, 'Show dev suggests'),
+            ))
+            ->setHelp(<<<EOT
+
+The <info>suggests</info> command show packages that suggesting to install other packages.
```
stof: use ``%command.name%`` for the command name rather than using ``suggests``. This way, the help does not need to be updated in case the command gets renamed

src/Composer/Command/SuggestsCommand.php
```diff
@@ -0,0 +1,59 @@
...
+
+EOT
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $lockData = $this->getComposer()->getLocker()->getLockData();
```
stof: what if there is no lock data ? You will get a notice currently because this case is not handled

Seldaek:
Merged as part of #4177

Issue #1522, opened 2013-01-24
Feature: Install only packages that work up to a certain PHP version


finishingmove:
So, the problem I encountered is the following:

I have a development environment where I use PHP 5.4.7 and a staging environment which only supports PHP 5.3.10 for now. When I deploy my project to the staging environment, the composer hook automatically executes, but sometimes for "*" it resolves to a new version of a package that implements some 5.4.7 features, hence the install fails...

It would be good to be able to specify a top-limit php version and install only packages that work under that version, rather than going through all the problematic packages and downgrading their version manually.

So, instead of "vendorX/packageX": "version-that-works-with-php-5.3.10", i want to do just "vendorX/packageX": "*" and have a general setting like "php-version-available" : "5.3.10".

Regards,

Arsen.

Tobion:
:+1:

postalservice14:
:+1:

postalservice14:
Do you think a composer plugin could accomplish this?  I'm not sure where in the composer event chain this could happen.  Maybe on a `pre-dependencies-solving` event it could remove package versions that are above the specified PHP platform or something?

slbmeh:
@postalservice14 you could make sort of a decorator repository that wraps a composite repository and filters the results in the what provides phase. Then your plugin can replace the installed repositories with your composite containing the original.

Seldaek:
Done https://github.com/composer/composer/pull/2990#issuecomment-97597919

emillosanti:
:+1:

PR #2990, opened 2014-05-14
Ignore platform packages

HuffAndPuff:
Here's my attempt at #1522.

If a package is present in the 'platform'  config option, that package will override whatever PlatformRepository finds.
The "1.0.0" value is a version and not a constraint, so you can't do `"ext-pdo":"*"`. I don't know if that's a good idea though.

```
"config":{
  "platform": {
    "ext-pdo": "1.0.0"
  }
}
```

It also assumes that only a single version of each platform package will be found.
Finally, no tests in there. I wouldn't know where to start...
I have tested it locally though with success.

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -39,6 +39,7 @@ class Config
         'optimize-autoloader' => false,
         'prepend-autoloader' => true,
         'github-domains' => array('github.com'),
+        'platform' => array(),
     );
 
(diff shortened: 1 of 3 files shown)


Seldaek:
@HuffAndPuff thanks! Finally found the time to look at this, merged it and wrapped it up in a57c51e8d78156612e49dec1c54d3184f260f144

PR #3025, opened 2014-05-29
New `repositories` Command

astorm:
This pull request is a feature request + initial implementation.  It contains a new top level `repositories` command for managing packagist and internal/external repository configuration.  Before I get to the justification (i.e. why duplicate functionality that's possible with `config`?), two other areas I could use some feedback on are

1. Arguments vs. Options:  I opted for a pure argument approach which might not fit into an existing philosophy when it comes to arguments vs. options.  Feedback here is super welcome. (`./bin/composer help repositories` can get you started)

2. Tests: Other symfony console projects I've worked on rely heavily on integration tests and  [CommandTester](http://api.symfony.com/2.0/Symfony/Component/Console/Tester/CommandTester.html) objects.  I wasn't able to find enough information about composer's test harness to know how I should set this up.  Any pointers here are also super welcome. 

Why a `repositories` Command
--------------------------------------------------
When I first started with Composer, I had a hard time with the concept of a "repository".  On the surface it was simple enough, but 3 things

1. The idea that a "composer" repository was a list of VCS packages, but that there was also a repository type called "vcs" that listed a single package

2. That the main packagist repository was outside my control as a client user of composer 

3. That the "packagist" configuration in composer.json isn't consistant — it can be both an array or an object, and that the configuration options available for a "packagist" repository (i.e. "disabled":true) are different from the configuration options available to the "composer" and "vcs" types

all combined to create friction whenever I wanted to get up and running with composer on a non-packagist repository (required for platforms like Magento, Wordpress, etc).  Once you master these concepts it's pretty easy to work with composer, but until you do working with external repositories is a little voodoo/cookbook, and it make make debugging performance problems and dependency chains difficult.

If there had been an actual `repositories` command that helped solidify these concepts, this learning curve would have been much shorter.  I know the `config` command has options for adding and removing repositories, but `config` was equally opaque for me when I was coming to the platform new.  That is, the config command is great for adding configuration, but it does little to help a new user understand **why** they're configuring composer the way they are. 

I can understand why this isn't/hasn't been a priority for the project, but as composer continues to make inroads into new corners of the PHP world, making things smooth and frictionless when it comes to learning platform concepts will be increasingly important.  A `repositories` command seems like a win here.

### DIFF
--- a/src/Composer/Command/RepositoriesCommand.php
+++ b/src/Composer/Command/RepositoriesCommand.php
@@ -1,0 +1,273 @@
+<?php
+namespace Composer\Command;
+
+
+use Composer\Factory;
+use Composer\Json\JsonFile;
+use Composer\Config\JsonConfigSource;
+use Composer\Json\JsonManipulator;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class RepositoriesCommand extends Command
+{
+    protected $config;
+    protected $configFile;
+    protected $configSource;
+    
+    protected function configure()
+    {
+        $this->setName('repositories')
+        ->setDescription('Manage external repositories in project composer.json')
+        ->setDefinition(array(
+            new InputArgument('action', InputArgument::OPTIONAL, '[add], [remove], [list], or [packagist].'),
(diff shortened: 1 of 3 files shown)


Seldaek:
Closing as #3749 fixes the packagist false use case and I just added more docs to config in 0e977e3fde3d84b98824753fde0079f2d182a7b9. IMO it is preferable to keep one consistent way to do things than having multiple commands acting similarly.

PR #3041, opened 2014-06-08
define COMPOSER_BIN_DIR constant in autoloader.php

renanbr:
it allows root project and __dependencies__ to find binaries


Seldaek:
Closing as I don't think it's really necessary, and it adds global scope clutter for every Composer user.

Issue #2972, opened 2014-05-06
Installation order


itscaro:
Hello, I could not find any information on this issue, so sorry if this information already exists somewhrere.

In my package, I require a plugin @ stable version (first entry in require), other requires are @ dev-trunk. With this configuration, composer installs requirements in dev states before the plugin.

If I change the plugin to dev version, composer installs in the declared order in "require"

Is that what composer should do?

Is it possible to force composer-plugin packages to be installed before other non composer-plugin packages?

Issue #2879, opened 2014-04-04
force a restart of the current operation, if a plugin gets installed/update/removed


backbone87:
since you can manipulate the root package, with the new plugin API, you are able to inject additional provide into the root package for example. doin this may result in a different solver solution, than this of the current run.

therefore it would be nice, if there were a possibility to restart the current operation, each time a plugin gets installed/updated/removed.

for the concrete implementation there may be used flags checked or new methods on the  plugin API to be checked, or something like a "interrupted exception" to be thrown.

aschempp:
+1

backbone87:
@Seldaek closed because implemented already or because its in "wont fix" state?

Seldaek:
Yeah it's rather won't fix sorry.

Issue #2878, opened 2014-04-04
usage of custom installers/plugins in packages is inconvenient


backbone87:
with the current installer/plugin implementation, in order to use a custom installer/plugin, you have to do the following things:
- use a package type that the custom installer/plugin supports
- require the installer/plugin

i think this implementation is a little bit redundant and API wise, a little bit inflexible, because the installer has to decide solely on the package type token, whether it can install the package or not. (correctly spoken, the decision is done by the caller of the installer interface, the installer just says "i support types".)

to clearify and also improve the API i propose the following change:
- add a "canDeploy" method to the installer interface, receiving the package to be deployed as an argument
- add a new package property "to be named" (or other composer.json definition construct), defining the preferred/required installer/plugin package constraint for deployment of the package
- deprecate the "supports" method of the installer interface
- deprecate the package "type" property (or convert its canonical content to match this of the new property, but still supporting "old types")
- discourage the require of installers/plugins needed to install the package

installer selection process for deployment of a certain package:
1) ask all installed installers via the "canDeploy" method, if they can deploy the package in question
2) if none is found resolve the constraint and install the plugin/installer, if its missing
3) if the installer/plugin could not be found, exception out
4) if the installer/plugin got installed, restart on 1)
5) if the installer/plugin was already installed, exception out

if there are multiple installers that can deploy a package, which one gets preferred?
- the one being asked first
- the return of the canDeploy can be more grained insteadof just true/false to indicate a "supportiveness"
- if the installers source package matches the installer/plugin constraint of the package to be deployed, then prefer this installer
- this point is up for discussion

if i want to force another installer to be used with a package linking a another installer?
- require the forced installer in your root composer.json

BC breaks?
@tristanlins mentioned that he has an idea how to implement this without BC breaks, but this ticket is at first a more conceptual look at the issue

Issue #1121, opened 2012-09-18
Bower Repository


RobLoach:
[Bower](http://twitter.github.com/bower/) is a package management system for JavaScript and CSS assets. A list of Bower packages is available here: http://sindresorhus.com/bower-components/ . The source and repositories for these packages is fed from here: http://bower.herokuapp.com/packages

[JamJS](http://jamjs.org/) is another package management system for web assets. Worth mentioning.

Leveraging these repositories from Composer would expose a huge amount of new packages to be consumed.

mzgol:
FWIW, jQuery has been published to npm since 1.11/2.1; it's a regular jQuery, without any Node-specific wrapping.

Seldaek:
Closing as it's very unlikely and not sure it's desirable tbh.

lkraav:
PS in the JS ecosystem, numerous arguments have lately been made towards dropping `bower` in favor of `npm`.

https://github.com/necolas/normalize.css/pull/455
https://gofore.com/stop-using-bower/

etc. For more, google the relevant keywords and notice pull requests all around.

PR #3082, opened 2014-06-29
Installs plugins and dependencies before installation of other packages

francoispluchino:
| Q             | A
| ------------- | -------------------------------
| Bug fix?      | yes (installation plugin order)
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2972, #2879, #2878, #1121, #3078
| License       | MIT

This feature allows installing plugins and their dependencies before restarting the installation of other packages in the project. It allows the plugins to be added directly in the project and must not be installed in global.

With this feature it is for example possible to create a plugin adding automatically  the repositories, without that Composer throws an error when the first installation.

It does not change the original behavior of the install, but at the same time, fixes a "strange" behavior of the order of the installation of plugins and their dependencies.

The dependencies plugins must be installed before the plugin, then the plugin must be installed, and then the rest of depedencies can be installed.

Test example:

```json
{
    "repositories": [
        {
            "type": "package",
            "package": [
                { "name": "pkg", "version": "1.0.0" },
                { "name": "pkg2", "version": "1.0.0" },
                { "name": "inst", "version": "1.0.0", "type": "composer-plugin" },
                { "name": "inst2", "version": "1.0.0", "type": "composer-plugin", "require": { "pkg2": "*" } }
            ]
        }
    ],
    "require": {
        "pkg": "1.0.0",
        "inst": "1.0.0",
        "inst2": "1.0.0"
    }
}
```

The order should be:

```
Installing inst (1.0.0)
Installing pkg2 (1.0.0)
Installing inst2 (1.0.0)
Installing pkg (1.0.0)
```

and not (actually):

```
Installing inst (1.0.0)
Installing pkg (1.0.0)
Installing pkg2 (1.0.0)
Installing inst2 (1.0.0)
```

### DIFF
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -41,12 +41,15 @@ class Solver
     protected $problems = array();
     protected $learnedPool = array();
 
-    public function __construct(PolicyInterface $policy, Pool $pool, RepositoryInterface $installed)
+    protected $skipException;
+
+    public function __construct(PolicyInterface $policy, Pool $pool, RepositoryInterface $installed, $skipException = false)
     {
         $this->policy = $policy;
         $this->pool = $pool;
         $this->installed = $installed;
         $this->ruleSetGenerator = new RuleSetGenerator($policy, $pool);
+        $this->skipException = $skipException;
     }
 
     // aka solver_makeruledecisions
@@ -89,6 +92,13 @@ class Solver
                 $problem->addRule($conflict);
                 $this->disableProblem($rule);
                 $this->problems[] = $problem;
+
+                if ($this->skipException) {
+                    $cLiterals = $conflict->getLiterals();
+                    $cLiteral = abs($cLiterals[0]);
+                    $this->decisions->revertLast();
+                    $this->decisions->decide($cLiteral, 1, $rule);
+                }
                 continue;
             }
 
@@ -189,13 +199,35 @@ class Solver
             }
         }
 
-        if ($this->problems) {
-            throw new SolverProblemsException($this->problems, $this->installedMap);
+        if (!$this->skipException) {
+            $this->validate();
         }
 
         $transaction = new Transaction($this->policy, $this->pool, $this->installedMap, $this->decisions);
 
         return $transaction->getOperations();
+    }
+
+    /**
+     * Validate the solver, a thrown exception if an problem is present
+     *
+     * @throws SolverProblemsException
+     */
+    public function validate()
+    {
+        if ($this->problems) {
+            throw new SolverProblemsException($this->problems, $this->installedMap);
+        }
+    }
+
+    /**
+     * Check if an problem is present
+     *
+     * @return bool
+     */
+    public function hasProblems()
+    {
+        return (bool) $this->problems;
     }
 
     protected function literalFromId($id)
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -106,7 +106,7 @@ class Installer
     protected $update = false;
     protected $runScripts = true;
     protected $ignorePlatformReqs = false;
-    protected $runNewAnalysis = false;
+    protected $pluginsForNewAnalysis = array();
     /**
      * Array of package names/globs flagged for update
      *
@@ -217,8 +217,14 @@ class Installer
         try {
             $this->suggestedPackages = array();
             $res = $this->doInstall($localRepo, $installedRepo, $platformRepo, $aliases, $this->devMode, true);
-            if ($res === 0 && $this->runNewAnalysis) {
-                $this->io->write('<info>A new analysis of dependencies is required</info>');
+            if ($res === 0 && !empty($this->pluginsForNewAnalysis)) {
+                if ($this->io->isVerbose()) {
+                    $this->io->write('<info>A new analysis of dependencies is required by:</info>');
+                    foreach ($this->pluginsForNewAnalysis as $pluginName => $status) {
+                        $this->io->write('  - Plugin <info>' . $pluginName . '</info> (<comment>' . (is_string($status) ? $status : 'always') . '</comment>)');
+                    }
+                    $this->io->write('');
+                }
                 $res = $this->doInstall($localRepo, $installedRepo, $platformRepo, $aliases, $this->devMode, false);
             }
             if ($res !== 0) {
@@ -387,7 +393,7 @@ class Installer
         }
 
         if ($this->update) {
-            if ($delayValidation) {
+            if ($delayValidation || (!$delayValidation && $this->io->isVerbose())) {
                 $this->io->write('<info>Updating dependencies'.($withDevReqs ? ' (including require-dev)' : '').'</info>');
             }
 
@@ -440,7 +446,7 @@ class Installer
                 }
             }
         } elseif ($installFromLock) {
-            if ($delayValidation) {
+            if ($delayValidation || (!$delayValidation && $this->io->isVerbose())) {
                 $this->io->write('<info>Installing dependencies'.($withDevReqs ? ' (including require-dev)' : '').' from lock file</info>');
             }
 
@@ -462,7 +468,7 @@ class Installer
                 $request->install($link->getTarget(), $link->getConstraint());
             }
         } else {
-            if ($delayValidation) {
+            if ($delayValidation || (!$delayValidation && $this->io->isVerbose())) {
                 $this->io->write('<info>Installing dependencies'.($withDevReqs ? ' (including require-dev)' : '').'</info>');
             }
 
@@ -485,7 +491,11 @@ class Installer
         $solver = new Solver($policy, $pool, $installedRepo, $delayValidation);
         try {
             $operations = $solver->solve($request, $this->ignorePlatformReqs);
-            $this->validateOperations($solver, $operations, $delayValidation);
+
+            if ($delayValidation) {
+                $this->validateOperations($solver, $operations);
+            }
+
             $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::POST_DEPENDENCIES_SOLVING, $policy, $pool, $installedRepo, $request, $operations);
         } catch (SolverProblemsException $e) {
             $this->io->write('<error>Your requirements could not be resolved to an installable set of packages.</error>');
@@ -504,7 +514,10 @@ class Installer
 
         $operations = $this->movePluginsToFront($operations);
         $operations = $this->moveUninstallsToFront($operations);
-        $operations = $this->keepOnlyPluginsAndDependencies($operations, $delayValidation);
+
+        if ($delayValidation) {
+            $operations = $this->keepOnlyPluginsAndDependencies($operations);
+        }
 
         foreach ($operations as $operation) {
             // collect suggestions
@@ -664,42 +677,56 @@ class Installer
      *
      * @param Solver               $solver
      * @param OperationInterface[] $operations
-     * @param bool                 $delayValidation
      */
-    private function validateOperations(Solver $solver, array $operations, $delayValidation = false)
+    private function validateOperations(Solver $solver, array $operations)
     {
-        if (!$delayValidation) {
-            return;
-        }
-
-        $hasPlugin = false;
-
         foreach ($operations as $op) {
             $package = $this->getOperationPackage($op);
 
             if ($this->isValidOperation($op) && $this->isPlugin($package)) {
-                $hasPlugin = true;
-                $this->runNewAnalysis = $solver->hasProblems();
-                break;
+                if (false !== $status = $this->pluginHasNeedRestart($solver, $package)) {
+                    $this->pluginsForNewAnalysis[$package->getName()] = $status;
+                }
             }
         }
 
-        if (!$hasPlugin) {
+        if (empty($this->pluginsForNewAnalysis)) {
             $solver->validate();
         }
+    }
+
+    /**
+     * Check if the plugin need restart a new analysis.
+     *
+     * @param Solver           $solver
+     * @param PackageInterface $package
+     *
+     * @return bool|string
+     */
+    private function pluginHasNeedRestart(Solver $solver, PackageInterface $package)
+    {
+        $extra = $package->getExtra();
+        $restart = array_key_exists('installer-restart', $extra)
+            ? $extra['installer-restart']
+            : false;
+
+        if (true === $restart || ('on-exception' === $restart && $solver->hasProblems())) {
+            return $restart;
+        }
+
+        return false;
     }
 
     /**
      * Keeps only the plugins and their dependencies, only for the plugins installation process
      *
      * @param OperationInterface[] $operations
-     * @param bool                 $delayValidation
      *
      * @return OperationInterface[]
      */
-    private function keepOnlyPluginsAndDependencies(array $operations, $delayValidation = false)
+    private function keepOnlyPluginsAndDependencies(array $operations)
     {
-        if ($delayValidation && $this->runNewAnalysis) {
+        if (!empty($this->pluginsForNewAnalysis)) {
             for ($i = count($operations)-1; $i >= 0; $i--) {
                 $op = $operations[$i];
                 $package = $this->getOperationPackage($op);
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -106,6 +106,7 @@ class Installer
     protected $update = false;
     protected $runScripts = true;
     protected $ignorePlatformReqs = false;
+    protected $runNewAnalysis = false;
     /**
      * Array of package names/globs flagged for update
      *
@@ -215,7 +216,11 @@ class Installer
 
         try {
             $this->suggestedPackages = array();
-            $res = $this->doInstall($localRepo, $installedRepo, $platformRepo, $aliases, $this->devMode);
+            $res = $this->doInstall($localRepo, $installedRepo, $platformRepo, $aliases, $this->devMode, true);
+            if ($res === 0 && $this->runNewAnalysis) {
+                $this->io->write('<info>A new analysis of dependencies is required</info>');
+                $res = $this->doInstall($localRepo, $installedRepo, $platformRepo, $aliases, $this->devMode, false);
+            }
             if ($res !== 0) {
                 return $res;
             }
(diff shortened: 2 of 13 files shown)


bishopb:
Neither npm nor bower know anything about `vendor/`.  Only composer does.  Thus, some composer-based tool needs to stitch the `vendor` knowledge together in such a way that npm and bower can digest.  One such tool is [composer-npm-bridge](https://github.com/eloquent/composer-npm-bridge).  I've been using it for a while to resolve `package.json`, which includes bower, that then resolves `bower.json`.  It works well.

But composer-npm-bridge requires `npm`.  I like that this package doesn't require external programs.  Since composer, npm, and bower are all doing essentially the same thing, it seems reasonable that one package could emulate the others and that the dependency list could be specified in one place.

**But with all respect, I think this plugin's design of `npm-asset` and `bower-asset` as "virtual" vendors is misguided.**  That design decision is at the root of this issue.  And other future issues: when someone creates an "npm-asset" (or "bower-asset") packagist category, then this plugin will essentially stop working, without some patching to map around the new category.

I don't think we should go about changing fundamental composer behavior to suit a plugin, no matter how awesome that plugin would be.  So with a :cry: I must :-1:.

cebe:
> In this specific scenario, composer-asset-plugin could scan a separate file (eg composer-assets.json) that lists asset dependencies, then doing its own work after install has completed.

This is against what you claim before: "'work with the grain' of composer". If the plugin installs packages, these should be installed in the resolving process of composer and not afterwards and outside. You could use a completely separated tool for that. With the plugin we are able to specify dependencies on bower/npm assets and they get installed if matching constraints are found or we get an error if it is not possible to resolve the situation. A separate process can not fulfill the requirements here.

bishopb:
@cebe: In the use cases I've encountered (which I recognize does not encompass *all* use cases), front-end assets rarely conflict.  Either (a) I deliberately choose packages that have compatible front-end assets or (b) handle the front-end entirely in the root package.  For me, the bootstrapping requirement to "install npm and/or bower installers" trumps the requirement "install no dependencies if any of their assets conflict".

Holistic assertions like "This PHP package requires Laravel 5, jQuery UI 1, Bootstrap 3 SASS, and grunt-sass" are a noble goal.  I fully support that goal.  But the approach in this PR concerns me.  Not because of the implementation, *per se*, but because it's doing something composer was never designed for: "virtual" packages.

If there's a plug-in to be made, it seems to me satis and/or packagist would be the place to house that plug-in, not composer.  `virtual/npm-*` and `virtual/bower-*` would proxy through the existing plugin code to resolve/map npm and bower to dependency specs that composer already consumes.  In this approach, everyone using composer immediately benefits, rather than just those who happen to stumble upon the plugin and add it to their composer.json.  And, this approach avoids the squatting problem (someone taking the npm-asset or bower-asset vendor).

Bottom line: I would prefer to see this handled universally via packagist virtual "proxying".  Alternatively, I'm happy managing back- and front-end dependencies separately, so long as I don't have to manually manage npm/bower installers.  But not this PR.  I don't see the value.

samdark:
The use case we're talking about are "builders". PHP code that generated JavaScript, HTML and CSS code. For example, based on validation rules in your PHP code it can generate validation rules for a given version of JavaScript form validation library. The thing with dependencies is that JavaScript library can be updated independently from PHP one so everything will be broken. In this case an ability to have a dependency between PHP package and JS package is vital.

cebe:
> I would prefer to see this handled universally via packagist virtual "proxying"

This is an option that would be technically possible but this means that we would have to integrate the bower/npm packages into composer for everyone which was not accepted as a solution by many people in discussions we had before we started using this plugin.

I have made a proof of concept for this some time ago:
https://github.com/cebe/composer-bower
some comparison/discussion:
https://github.com/cebe/composer-bower/issues/5

bishopb:
@samdark Yes, I follow.  Both approaches I'm advocating handle that, and neither require this PR.

Suppose PHP code builds Javascript that requires jQuery 2.1.2 and jQuery UI 1.10.4.

In the first approach, the one that works today but is not ideal, `composer.json` contains:
```json
"require": { "eloquent/composer-npm-bridge": "~2" }
```
The `package.json` file contains:
```json
"dependencies": { "jquery": "2.1.2", "jquery-ui": "1.10.4" }
```
Running `composer install` pulls down composer-npm-bridge 2.* into vendor, which runs post install the npm installer to get jQuery 2.1.2 and jQuery UI 1.10.4.  One can independently change the versions by editing these separate files.  And upstream versions can happily release new versions without impacting the code.  If one later add a PHP package that uses npm to require jQuery 1.11.2 (which is incompatible with 2.1.2) the post install step will complain and can be resolved appropriately.

If packagist specifically or satis generally were to add the features from this plugin, which the second approach I advocate, then one only needs a single `composer.json` to specify all dependencies:
```json
"require": {
    "eloquent/composer-npm-bridge": "~2",
    "virtual/npm-jquery": "2.1.2",
    "virtual/npm-jquery-ui": "1.10.4"
}
```

The plugin wants to specify the syntax like above.  That syntax is fantastic!  I love it.  I want it.   But I do not feel the composer plugin approach is working.  It's asking composer to do things it wasn't designed to do.  It's reserving "npm-asset" and "bower-asset" without any teeth to enforce the reservation.  It only works for those who happen upon the plugin.

The PHP world needs what this plugin offers.  But not this way.  Let packagist handle the proxying.  Let composer continue to operate as it does.

---- 

@cebe I believe a blending of your experiment and this plugin is the best road to follow.  Packagist would expose npm and bower packages via a reserved `virtual` vendor.  Upon a request for a specific package within that vendor (eg, `virtual/npm-jquery`), packagist (via the plugin code) would scan the corresponding npm (or bower) repository for matching versions.  That dependency graph would be packaged in composer-expected format and returned (as the plugin does now), and cached for subsequent fetches.

This approach is one of on-demand proxying in Packagist proper, rather than a separate Packagist-like repository that bootstraps npm and bower registries.

Analysis: Existing packages continue to work unaffected.  If a package author decides he needs an npm- or bower-held asset, he can look them up in Packagist and add them directly to his package's composer.json.  He doesn't have to install a special plugin.  He doesn't have to do a composer update to get the revised version with this new restart feature.  We're throwing away none of the plugin's great code: just moving where the work is done.  True, the Packagist farm is doing more work (because there are now lots more packages), but the work done is just "more of the same" and done "just in time".

damiandennis:
This plugin works great but I see the point being made, It sounds like a program needs to be made that can integrate composer, npm and bower into one package and check dependencies before running each of the package managers. I am not sure that however is even possible as it depends on how flexible each of the package managers. Also it does sound like a lot of work...

schmunk42:
@bishopb

> Either (a) I deliberately choose packages that have compatible front-end assets or (b) handle the front-end entirely in the root package. For me, the bootstrapping requirement to "install npm and/or bower installers" trumps the requirement "install no dependencies if any of their assets conflict".

(c) I install a bundle of PHP, CSS and JS files like [described here](https://github.com/composer/composer/pull/3082#issuecomment-66279466) and without any manual installation or setup.

That's my everyday use-case and I see no better solution than using the asset plugin, which does a wonderful job here.

> It's asking composer to do things it wasn't designed to do.

That's why we have the possiblity to create plugins, you can never foresee all possible use-cases or new ideas.

> It's reserving "npm-asset" and "bower-asset" without any teeth to enforce the reservation. It only works for those who happen upon the plugin.

I agree that this could be improved, ie. these paths could be made configurable like the paths, where the npm or bower packages are installed, [see here for an example](https://github.com/yiisoft/yii2/blob/master/composer.json#L140).

If you use the virtual packages, you also have to require the plugin, the problem is, that you have to install it globally at the moment, that's what this PR is trying to solve.

As a sidenote: You can create packages under any namespace, there's no reservation like the user- or org-name on GitHub.

> True, the Packagist farm is doing more work (because there are now lots more packages), but the work done is just "more of the same" and done "just in time".

This looks problematic (kindly speaking) to me, see [Module Counts](http://www.modulecounts.com), for a comparision of the numbers:

- packagist 46480
- bower 24855
- npm 115298

You're suggesting to proxy over the threefold number of packages than currently hosted on packagist - while we are already experiencing a lag of several minutes when creating a new version - and moreover a very large percentage of npm and bower packages are not of interest for PHP projects.

This would totally clutter the packagist search and is no option IMHO.

francoispluchino:
@bishopb

> But with all respect, I think the plugin's design of npm-asset and bower-asset as "virtual" vendors is misguided. That design decision is at the root of this issue. And other future issues: when someone creates an "npm-asset" (or "bower-asset") packagist category, then this plugin will essentially stop working, without some patching to map around the new category.

I agree with you, and that's why I chose to prefix the asset dependency with `TYPE-asset`. Nobody will use this prefix, but it is true that it can be problematic.

Regarding the design problem: thank you for your constructive comments, and it's always interesting to have external advice. I will not repeat all of my approach to my plugin (because it is not the subject), but the main idea was to use the maximum the native components of Composer (Installer, Sovler, Locker, Commands, etc...) and simply transformed the package information for Composer. 
Personally I would have liked the possibility to prefix the asset dependency  with "@bower" and "@npm" for example. In this way, Composer knows that dependency is a virtual dependency and does not take into account in its installation (by the Solver).

I also had the possibility to put the asset dependencies in the "extra" section, but in this case, we do not use the native components of Composer.

Your proposal to use "virtual/npm-*" and "virtual/bower-*" has the same problem as my implementation.

Your proposal to use a proxy, was my initial research: technically, it is possible, but unfortunately, the analysis is extremely heavy, and therefore, the limitation of 5000 request/hours to Github API would soon exploded just NPM has over 80 000 packages).

@bishopb; @cebe, @samdark The debate around my plugin is very interesting, but it not concern this PR directly. Indeed, this PR is not specific for my plugibut it allows to solve the problem of installation order of dependencies of plugins, and allows to the plugins to do the same action whatever the chosen mode of installation (projet or global mode).

What makes me happy, is to see the interest aroused by this PR when there are more than 6 months, and that I was ready, for the new year, to close this PR to choose another solution (but significantly less native).

I think about the management of assets (javascript/stylesheet) for servers languages other than JS, is a very interesting and important topic, because with any Framework used, management of assets is simply bypassed. For example I use a lot Symfony2, and the recommendation is to not put the assets in Bundles, but it no offers a technical solution to easily handle this case. And for other Frameworks (ZF2, Laravel4, etc...), from my initial tests, it's the same problem, except for Yii2 (but that may have changed since).

Another point, it is not because a project is not initially designed to manage something, that this tool can not evolve, if that something is a recurring problem. The management of assets is a recurring and pervasive problem for any project that manage the back-end and front-end. Given that Composer manages the PHP dependencies, I think it would be interesting it can manage the assets (without for all that use my solution).

For finish, you find that the principle is cool, but not the implementation, because for you, the plugin asks Composer to do something it does not. In this regard I do not agree with you. Did you see the Doc and FAQs? (see [How does work the plugin?](https://github.com/francoispluchino/composer-asset-plugin/blob/master/Resources/doc/faqs.md#how-does-work-the-plugin)).

bishopb:
@francoispluchino @schmunk42 I'm glad we're continuing to have this discussion, as I feel a comprehensive asset management solution would help the whole community.  But, yes, we're off topic with this PR.  Is there a mailing list, IRC channel, or other place we can move this larger conversation?

For the PR.  I did read the plugin docs, and I did read the entire PR, and I know a little bit about the Composer code base.  I am not an expert, though.

I am with @stof when he said "define a repository providing these packages to the solver, rather than trying to make the solver behave in weird ways".  I would prefer that repository be up at Packagist.  But, I am ok with the dependencies being in "extra", as I think you meant [in your recent #3602 comment](https://github.com/composer/composer/issues/3607#issuecomment-68532805).

I am with @naderman when he said ["This is really just not what the solver is made for at all"](https://github.com/composer/composer/pull/3082#issuecomment-66270017) regarding the `Solver.php` change.  I am following the logic, and it seems to work.  But again I'm not an expert. I personally want to see more test coverage, because I do not want a "Composer debacle" (like the silly [npm debacle](http://www.letscodejavascript.com/v3/blog/2014/03/the_npm_debacle)).  Some tests that I think are needed:

* plugin-restart-always.test but with --no-dev
* plugin-restart-always.test but with a/a in conflicting versions
* plugin-restart-always.test but with a/a depending on b/b
* plugin-restart-always.test but with a/a depending on b/b in conflicting versions
* timing tests with real-world (ie "non-trivial") dependency graphs

francoispluchino:
@bishopb Ok for the new tests, but can you be more precise for the last test ("timing tests with real-world (ie "non-trivial") dependency graphs")?

bishopb:
@francoispluchino The test cases have a maximum of three dependencies.  The packages and applications I regularly work on have hundreds of dependencies, specified in cross-cutting fashion: A depends upon B and C; B and C both depend on D; and B and D depend upon E.  Network overhead aside, the composer solver takes some time to process that dependency graph (call that time T_i).  I would like to have a performance test that illustrates how much average additional time, if any, the restarts and other changes add to T_i.

francoispluchino:
@bishopb For the cost of time, you can see [this comment](https://github.com/composer/composer/pull/3082#issuecomment-56683610).

bishopb:
@francoispluchino Yep, I saw the stats.  [I *trust* them, but a test case will *verify* them](http://en.wikipedia.org/wiki/Trust,_but_verify) :smile:

schmunk42:
@francoispluchino Could your plugin detect if it was installed globally and if not, fail with a human friendly error mesage?

francoispluchino:
@schmunk42 I think this is feasible with the property HOME directory. But for example, you added the plugin such as dependency of your project, and you ask your users to install the plugin in global mode.

At the first installation, the plugin in global mode is used, on the other hand, once the plugin is installed in project mode, it is the project plugin which is used (project plugins has a priority to global plugin).

So if I add a exception, I will break compatibility.

cebe:
So basically what this fixes is the first install, right?

francoispluchino:
@cebe Right.

bishopb:
@cebe @francoispluchino Reaffirming that "first" appears to be the root issue, as mentioned in the [require-global](https://github.com/composer/composer/issues/3607#issuecomment-68533771) conversation.  And also reaffirming what @schmunk42 [said]((https://github.com/composer/composer/issues/3607#issuecomment-69312112)) in that same conversation, paraphrased: *`require-global` will not prevent fixing 3082, but will prevent first-time related errors`*.

francoispluchino:
@bishopb My proposal of `require-global` (see this [comment](https://github.com/composer/composer/pull/3082#issuecomment-56285632)) does not prevent an error if a global plugin is not installed. This could be possible only for the root Composer package (read the proposal).

Indeed, the sub dependencies indicating that they need a global plugin (via require-global) cannot be analyzed, because the problem is the same that this PR: that is, that Composer cannot know the require-global of the dependence, because, in this dependency, there will be a sub-dependency that Composer will not know, and therefore an error will thrown.

To correct the problem, we should in this case, performing a first analysis dedicated to `require-global`, and launch a second analysis to resolve dependencies.

For reminder, it was the first version of this PR, and performance is poor compared to the last version proposed (see [this comment](https://github.com/composer/composer/pull/3082#issuecomment-56280328) vs [this comment](https://github.com/composer/composer/pull/3082#issuecomment-56683610)).

It is true, that the require-global will allow for a complete project to avoid unnecessary error, but this PR is much more relevant, especially that this latest version does not change the performance for all current and future projects, but only for projects that require a plugin must be installed prior to analysis (and only for the first installation, not updates, il the plugin is not installed in global mode).

Marlinc:
Any news on this?

schmunk42:
@Seldaek Do you have more feedback why this is closed now? Is it just closed or solved by a commit?

Seldaek:
It's just not mergeable so closing to clean up. It's not solved and I doubt it is really solveable without relying on global pre-installed plugins as people have been doing until now.

schmunk42:
Thank you for the info. Let me reference https://github.com/composer/composer/issues/3607 one more time.

It would be really great if we could get something like that, because I've seen so many bug reports where people were missing a global installation of the mentioned plugin.

kirkmadera:
Just a note, we have been using this at the project level for a long while now without issues. It's only the initial installation that is an issue. We run `composer require fxp/composer-asset-plugin`, then require our npm/bower assets. Future `composer install` runs work because of the composer.lock file. Then, `composer update` and `composer require` commands work at that point because the plugin is installed.

schmunk42:
@kirkmadera Sorry, but I must state the exact opposite, i.e. recently we had an issue with a global installation of the plugin in version `1.2.1`, while [one package in the project was requiring the plugin](https://github.com/romka-chev/yii2-swiper/pull/4) - which lead to installation of `1.3.0` (which introduces changes in config options).

> It's only the initial installation that is an issue.

Regarding the above statement, the outcome is that on the first installation/update composer uses `1.2.1` with no warnings. On the second update `1.3.0` is used, which issues warnings about deprecated config options. If you change `composer.json` settings to the new config options to get rid of the warnings, your third installation runs fine ... but the next first installation on another machine (with global 1.2.1 installation) will fail (or run without the correct config), since `1.2.1` does not understand the new options and it also does not issue any warnings about that.

CC: @marc7000 @handcode

kirkmadera:
In our setup, we do not use a global installation. Here is a test.

Remove the global plugin if you have it:
```
composer global remove fxp/composer-asset-plugin;
```

Prep a test dir with jquery installed.
```
mkdir test;
cd test;
composer require fxp/composer-asset-plugin;
composer require npm-asset/jquery;
```

Remove the `vendor` dir so that you are in a state like you just cloned the repo.

Attemp to use the asset plugin here and it will fail because the plugin is not installed into vendor:
```
composer require npm-asset/bootstrap
```

Run `composer install` to get the plugin installed again.

Now, the bootstrap include will work:
```
composer require npm-asset/bootstrap
```

To make this work you have to:
1. Commit your lock file
2. Run `composer install` after a git clone before running `composer require` or `composer update`
3. Avoid manually editting your composer.json file with additional npm assets or bower assets before installing fxp/composer-asset-plugin

Committing the lock file is good practice anyways to make sure everyone is working with the same versions of everything. Running `composer install` on a checkout should be pretty standard. The third one is the gotcha when you start a project and are trying to build a reusable composer.json file to start with.

schmunk42:
Just for reference...

> Project scope installation
> The installation in the project scope is not supported (see the issue https://github.com/fxpio/composer-asset-plugin/issues/7).

https://github.com/fxpio/composer-asset-plugin/blob/master/Resources/doc/index.md#project-scope-installation

Also related: https://github.com/wikimedia/composer-merge-plugin/issues/138 but the other way around

Issue #3125, opened 2014-07-15
Always Recommend a Global Installation


weaverryan:
Hi guys!

This is an RFC to always propose that Composer is installed in some global way so that you can just run commands like `composer install`.

This would also be consistent with other package managers, like npm.

#### Current Problem

Currently, some people install Composer globally and some just download the phar. So, depending on how you install, your commands look different:

```bash
composer install
php composer.phar install
```

The problem is that this little inconsistency makes documentation on libraries harder to write/understand. For example, if I have a library, I would love to just say:

"Run the following command to install"

```bash
composer require cool-name/cool-lib
```

I've seen beginners see this command, copy it, get a command not found, and get confused :).

What do you think? And can we do this consistently over all OS's? Can we use `~/bin/composer` on all systems? This would avoid needing to sudo some command to move the phar file there.

Btw, @stof proposed this:

```bash
$ curl -sS https://getcomposer.org/installer | php -- --bin-dir ~/bin/composer
```

Seldaek:
Closing as I don't think we can really do so much here. Best would be to get composer pre-installed with php or at least available as distribution packages. I think it's getting there (distros) slowly and the 1.0 stable should help.

PR #3130, opened 2014-07-17
New command to install a global composer

weaverryan:
Hi guys!

See #3125. This introduces a new command - `add-global-cmd` - which attempts to move the PHAR file that's currently being executed into the bin directory where the php executable lives. The plan is to then (in a different PR, on the proper repository) modify the `https://getcomposer.org/installer` script to call this command on the brand new `composer.phar` file it just downloaded.

With this, the original `composer.phar` file remains in-tact, meaning that people can easily use "local" composer.phar installs just like before.

To summarize, after this is merged and the `installer` script is modified (not done here), the flow would be:

1. Run `curl -sS https://getcomposer.org/installer | php` (probably with a new option, like `--global` that triggers this new behavior, to protect BC)

2. The `composer.phar` file is downloaded into the current directory, like always

3. This new command asks the user if they want a global `composer` command also installed

4. If yes, the `composer.phar` file is copied into a global path. This requires sudo - so most users will be asked for their password

5. After finishing, the installer script will summarize what happened and how to run composer (e.g. "Run `composer` or `php composer.phar` to use Composer").

Two questions are:

A) Do we see any issues or improvements with this "flow"?

B) Do we see any technical/security issues, especially with running sudo and trying to figure out *where* to place the `composer` executable.

Thanks!

### DIFF
--- a/src/Composer/Command/AddGlobalCmdCommand.php
+++ b/src/Composer/Command/AddGlobalCmdCommand.php
@@ -38,13 +38,7 @@ EOT
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        // todo - try to get this to work on Windows too
-        if (defined('PHP_WINDOWS_VERSION_BUILD')) {
-            $output->writeln('This command cannot be run on Windows. Use the Windows installer.');
-
-            return 1;
-        }
-
+        $isWindows = defined('PHP_WINDOWS_VERSION_BUILD');
         $installMode = $input->getOption('install-mode');
 
         // if we're being called by the installer, ask for confirmation first
@@ -83,6 +77,11 @@ EOT
 
         // try to copy the composer file
         $targetPath = $targetDir.'/composer';
+        if ($isWindows) {
+            // windows will have a composer.phar AND a composer.bat that will call it
+            $targetPath .= '.phar';
+        }
+
         if (!$this->copyComposerExec($sourcePhar, $targetPath)) {
             if (!$installMode) {
                 $output->writeln(sprintf('<error>Failed copying composer into %s</error>', $targetDir));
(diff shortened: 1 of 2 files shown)


Seldaek:
Thinking about this some more I actually don't think this belongs in composer at all. It should be in the installer directly. That just means Process/ProcessFinder can't be relied on, but with php5.5 having PHP_BINARY that solves the latter, and hopefully we can trust shell_exec() with a simple sudo command on most linux boxes these days for the former. On windows it's a simple copy so no worries.

I think we could live with that --global flag being 5.5+ only and be done with it. @weaverryan would you like to try and migrate this to the composer/getcomposer.org repo?

weaverryan:
@Seldaek I was thinking again about this PR lately, so your timing is awesome :).

Yes, I will try to work this into the installer itself as soon as I can! I put it into Composer only originally so I had access to a few more tools, but no worries.

Cheers!

src/Composer/Command/AddGlobalCmdCommand.php
```diff
@@ -0,0 +1,233 @@
...
+     * @return bool
+     */
+    protected function copyComposerExec($sourcePhar, $targetPath, $isWindows)
+    {
+        if ($isWindows) {
+            // use a normal copy for Windows
+            return copy($sourcePhar, $targetPath);
+        } else {
+            // use a command with "sudo", which is likely needed
+            $cmd = sprintf('sudo cp %s %s', $sourcePhar, $targetPath);
```
weaverryan: What would you recommend instead? We could have the user run something like `curl -sS https://getcomposer.org/installer | sudo php` in the first place... but downloading a script and giving it sudo access immediately feels like a security problem, though that's basically what pip does ("download this file, then execute it - you may need admin access" - https://pip.pypa.io/en/latest/installing.html#install-pip)

Seldaek:
Ok awesome, so I close this and I would say do a best effort approach, if it fails we can link to the docs instructing how to do it manually. I'd prefer that than having a gigantic block of code handling every eventuality :)

src/Composer/Command/AddGlobalCmdCommand.php
Seldaek: If you demand admin access at any point, the users essentially have to trust you, so I don't think it's more or less of a problem no matter where you do it. On linux though if we can't move the file directly it would IMO be ok to error out with "run `mv composer.phar /usr/bin/composer` as root to finalize the installation". You can already mark it executable and so on so it's all ready but just has to be moved, and then you clearly see what you are doing as root, there is no ambiguity nor much trust required.

PR #3184, opened 2014-08-03
generateUnixyProxyCode() properly supporting Cygwin & Git Bash

Wirone:
Based on Composer's proxy: https://github.com/composer/windows-setup/blob/master/src/shims/composer

Previous `generateUnixyProxyCode()`'s version was generating unixy-proxy file which was causing problems:

```
$ phpunit
Could not open input file: /cygdrive/c/Users/<user>/AppData/Roaming/Composer/vendor/phpunit/phpunit/phpunit
```


Seldaek:
Thanks, finally merged this, and I applied the change also from #3229 so it should all be good now I hope.

Wirone:
I've updated my Composer, installed some vendors and yes - proxies are generated properly and even work! ;-) Thanks, glad I could contribute to this extremely useful project.

PR #3194, opened 2014-08-08
Add --no-scripts argument for dump-autoload command

cw-ozaki:
Script is executed, if post-autoload-dump specified and ````composer install --no-scripts```` is executed.

````
"scripts": {
    "post-autoload-dump": "echo 1"
},
````

This PR do not execute scripts.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -40,6 +40,8 @@ class AutoloadGenerator
 
     private $devMode = false;
 
+    private $scriptMode = false;
+
     public function __construct(EventDispatcher $eventDispatcher, IOInterface $io = null)
     {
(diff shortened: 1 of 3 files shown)


michalcarson:
@cw-ozaki I'd like to see this PR brought in. How can I help?

Also need a change to src/Composer/Installer.php around line 328 to pass $this->runScripts to the Autoload Generator instance.

Seldaek:
Thanks, and sorry for the long delay merging this, kinda got lost..

craigh:
the language is a little confusing for me here. Please let me clarify:

before this PR, the `--no-scripts` flag was not respected and scripts ran anyway.
after this PR, the `--no-scripts` flag operates as should be expected (no scripts are run).

correct?

alcohol:
Correct.

PR #3199, opened 2014-08-11
Add support of 'select' from option list in ConsoleIO and NullIO

fduch:
Sometimes it is useful in scripts (script handlers, plugins) to have ability to choose values from list of predefined options during console interactions. This pull request makes it possible due to support of 'select' method in IOInterface

### DIFF
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -237,4 +237,12 @@ class ConsoleIO extends BaseIO
         // not able to hide the answer, proceed with normal question handling
         return $this->ask($question);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function select($question, $choices, $default = null, $attempts = false, $errorMessage = 'Value "%s" is invalid', $multiselect = false)
+    {
+        return $this->helperSet->get('dialog')->select($this->output, $question, $choices, $default, $attempts, $errorMessage, $multiselect);
+    }
 }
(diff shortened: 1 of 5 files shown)


src/Composer/IO/ConsoleIO.php
```diff
@@ -237,4 +237,12 @@ public function askAndHideAnswer($question)
         // not able to hide the answer, proceed with normal question handling
         return $this->ask($question);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function select($question, $choices, $default = null, $attempts = false, $errorMessage = 'Value "%s" is invalid', $multiselect = false)
+    {
+        return $this->helperSet->get('dialog')->select($this->output, $question, $choices, $default, $attempts, $errorMessage, $multiselect);
```
nevvermind: You'll need to output to STDERR:
```php
if ($output instanceof ConsoleOutputInterface) {
    $output = $output->getErrorOutput();
}
return $this->helperSet->get('dialog')->select($output /* rest of params ... /*);
```

nevvermind: Also, you'll need to take into consideration not blocking the input when it's non-interactive (by checking `\Composer\IO\ConsoleIO::isInteractive()`). Because of this, I think _every_ caller should provide a default, just to make them aware - aka `$default` should not be optional.

Seldaek:
Thanks.. I finally took the time to wrap this up

Issue #1968, opened 2013-06-05
Add option to compare working state with lock file in composer status.


slbmeh:
Currently `composer status` shows me all of the dirty working trees of repositories that composer has installed.  An option to check the ref of the head for each package and compare it to that in the lock would be a nice feature to have.

PR #3265, opened 2014-09-08
Check for VCS reference difference in command 'status'

dol:
Composer status detects only uncommited changes for VCS packages. This patch
adds informations about different reference versions.
If a package has no uncommited changes, but differs from the
revision defined in composer.lock, a change message is triggered.

This commit is related to #1968

Missing pieces:
* Perforce support
* It might be better to add a separate abstract function to VcsDownloader for this check (reuse)
* Test case for new behavior


Seldaek:
Closing in favor of #3788 which is a cleaner patch IMO doing a similar thing.

PR #3272, opened 2014-09-11
Added cacheCredentials config flag for saved SVN credentials

bashofmann:
Added cacheCredentials config flag to svn repository configuration to control the --no-auth-cache flag

Example config:

``` json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "http://svn.example.org/projectA/",
            "svn-cache-credentials": false
        }
    ]
}
```


marcovtwout:
Is there a way to set this behavior in the global configuration?

zzanzare:
is there a way to set this for CLI command "create-project"?

Issue #3337, opened 2014-10-08
Change visibility of ClassLoader::findFileWithExtension to public


sndsgd:
I'd like to be able to use ```Composer\Autoload\ClassLoader::findFileWithExtension``` for locating files that are NOT ```.php``` or ```.hh``` files. Any reason this method is currently private? Any objections to making it public?

PR #3344, opened 2014-10-10
Changed visibility of ClassLoader::findFileWithExtension to public, issu...

jakoch:
...e #3337


naderman:
There is a question on why this is needed on the issue, so should not be merged before that is answered.

PR #3365, opened 2014-10-18
Updated validator

WouterJ:
It's really strange to say that `zendframework zend-eventmanager` is a full package name, since Composer always uses `zendframework/zend-eventmanager`.

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -328,6 +328,7 @@ EOT
             return $result;
         }
 
+        $versionParser = new VersionParser();
         while (null !== $package = $dialog->ask($output, $prompt)) {
             $matches = $this->findPackages($package);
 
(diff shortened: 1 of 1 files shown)


WouterJ:
Sorry for the delay, @naderman and @Seldaek.

I've now updated the PR. It now accepts both `acme/example ~2.3` and `acme/example` (and also `php >=5.6` and even `php`).

There were no tests for the command, so I've tested it locally.

barryvdh:
I also came across this. Was a bit confusing what the expected input was, especially because the proposed packages are just `vendor/name`. I would think that if you leave the version constraint out, it should find the best suitable package, just like the current `composer require` behaviour.

WouterJ:
@barryvdh that's exactly what this PR does. Not sure why it isn't merged yet

barryvdh:
Ok cool, :+1:  for this then ;)

WouterJ:
Any reason to not merge this PR? Got hit by this once again this morning :)

Seldaek:
Just time folks, just time, but thanks @WouterJ :)

PR #3369, opened 2014-10-21
Initial GitLab Driver

### LINKED ISSUE #3241: GitLab Support
It would be great if GitLab was supported in the same was that GitHub and BitBucket is - with credentials, api and dist.

henrikbjorn:
This is a proof of concept, and mostly done to gather feedback on the
structure of the driver and to see if this is something that Composer
should include in core.

Tests are coming. I find the VcsDriver interface confusion, or rather the GitHub implementation, but this is a first stab at it.

Any feedback?

fixes #3241

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -94,16 +94,25 @@ class GitLabDriver extends VcsDriver
         return false;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function getRepositoryUrl()
     {
         return 'https://'.$this->originUrl.'/'.$this->owner.'/'.$this->repository;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function getUrl()
     {
         return $this->getRepositoryUrl() . '.git';
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function getDist($identifier)
     {
         $url = $this->getApiUrl().'/repository/archive?sha='.$identifier;
@@ -111,17 +120,15 @@ class GitLabDriver extends VcsDriver
         return array('type' => 'zip', 'url' => $url, 'reference' => $identifier, 'shasum' => '');
     }
 
+    /**
+     * {@inheritDoc}
+     */
     public function getSource($identifier)
     {
-    }
-
-    public function cleanup()
-    {
+        return array('type' => 'git', 'url' => $this->getUrl(), 'reference' => $identifier);
     }
 
     /**
-     * Default branch
-     *
      * {@inheritDoc}
      */
     public function getRootIdentifier()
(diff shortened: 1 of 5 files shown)


naderman:
@henrikbjorn do you plan on finishing this up?

src/Composer/Repository/Vcs/GitLabDriver.php
```diff
@@ -0,0 +1,236 @@
...
+
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getRepositoryUrl()
+    {
+        return 'https://'.$this->originUrl.'/'.$this->owner.'/'.$this->repository;
```
GromNaN: Gitlab can be under **http**.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -149,10 +149,17 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
...
             $this->io->write((substr($fileUrl, 0, 4) === 'http' ? 'Downloading ' : 'Reading ') . $fileUrl);
         }
+
         if (isset($options['github-token'])) {
             $fileUrl .= (false === strpos($fileUrl, '?') ? '?' : '&') . 'access_token='.$options['github-token'];
             unset($options['github-token']);
         }
+
+        if (isset($options['gitlab-token'])) {
+            $fileUrl .= (false === strpos($fileUrl, '?') ? '?' : '&') . 'private_token='.$options['gitlab-token'];
```
GromNaN: For people using many Gitlab servers, the token should defined by domain name.

GromNaN:
I've discussed with @Seldaek: he agrees on the principle of including a Gitlab driver into the core of composer.

I have a great interest in this feature and I'll test this PR on my Gitlab server.

henrikbjorn:
Yes i intend to do so, but havent had the time yet. Since we started using Statis with GitLab and got it working that way.

GromNaN:
@henrikbjorn I'm using satis too, but I expect 2 improvements with the GitLab support in composer:
1. Provide "dist" packages downloaded from Gitlab (to speedup deployments)
2. Speedup satis build by avoiding calls to Git command.

henrikbjorn:
@GromNaN we have set Satis up to dump the dist packages in our install. So we wont benefit that much in deploys.

The gitlab download is correct, that will speed satis build up (a lot)

src/Composer/Repository/Vcs/GitLabDriver.php
henrikbjorn: Yup, ours is current. But we have to default to something. Otherwise we have to make it a required parameter for the domains thing.

henrikbjorn: We could try and guess it from the url, but that wont work over the `git` protocol

henrikbjorn: We could require the gitlab-domain option to include if it is secure or not, like `https://domain` and call it hosts instead?

henrikbjorn:
#3765 superseeds this one.

PR #3275, opened 2014-09-14
[WIP] Progress Tracker

dantleech:
As discussed in https://github.com/composer/composer/issues/2809

This WIP PR adds a `WorkTracker` set of classes which allow progress to be registered and subsequently evaluated by a work tracker formatter.

Currently progress is only tracked for the dependency resolution process (i.e. all the time that passes before it starts installing packages).

- [x] Review the code
- [ ] Implement progress tracking for the rest of the installation process (?)
- [ ] Improve output formatters

## Work Trackers

Work trackers have the methods `complete` and `ping` (to register a progression) in addition to factory methods for creating *nested* work trackers. 

There are three work tracker classes. One is a contextual tracker, which is used to access the *current* tracker the other two are `BoundedWorktracker` for tracking progressions with a known number of steps and `UnboundedWorkTracker` for handling progressions of an undefined length.

Library classes that need to track their progress are always injected with the `ContextWorkTracker`.

## Output

The output is managed by a class implementing the `FormatterInterface`.

The output formatter that will be used is determined by the `--pretty` option.

Because the work trackers are nested it is possible to produce hierchical output.

````
$ ./bin/composer install --pretty=multi                                                                                                             ⏎
[5/5] Installing dependencies (including require-dev)                                                             
[81/160] Getting rules for "70" jobs with "80" installed packages                                                 
[1/-] Add rules for package symfony/yaml
````

There is also a verbose debug formatter:

````
$ ./bin/composer install --pretty=debug | less                                                                                                                 ⏎
  BEGIN: Dispatching pre update event
  DONE: Dispatching pre update event
  BEGIN: Adding repsoitories
  PING: [#000001] [2.0231s] Adding repsoitories
  DONE: Adding repsoitories
  BEGIN: Removing unstable packages from the local repository (if they don't match the current stablitity settings)
  PING: [#000001] [0.0002s] Removing unstable packages from the local repository (if they don't match the current stablitity settings)
  PING: [#000002] [0.0001s] Removing unstable packages from the local repository (if they don't match the current stablitity settings)
  PING: [#000003] [0.0001s] Removing unstable packages from the local repository (if they don't match the current stablitity settings)
  PING: [#000004] [0.0001s] Removing unstable packages from the local repository (if they don't match the current stablitity settings)
  PING: [#000005] [0.0001s] Removing unstable packages from the local repository (if they don't match the current stablitity settings)
  PING: [#000006] [0.0001s] Removing unstable packages from the local repository (if they don't match the current stablitity settings)
[...]

PR #3376, opened 2014-10-26
Added Progress Output

neon64:
Often Composer can be quite slow, so I though it would be useful to have a system to output progress information.

I added a `ProgressInterface` interface that defines methods for writing progress information.
I also looked through the `Installer` source, and found slow parts of it that should be given the 'progress' treatment.

If you don't specifiy a `ProgressInterface`, it will just default to `VoidProgress` which will output nothing.

At the moment, the only proper `ProgressInterface` implementation is `FileProgress` writes to a JSON file like this:

    {
      "progress": {
        "section": {
          "message": "Processing Dev Packages (Task: 'force-links')",
          "count": 2,
          "total": 0
        },
        "item": {
          "message": "symfony\/translation",
          "count": 0,
          "total": 66
        }
      }
    }

However I was thinking that in the future you could add a `ConsoleProgress` implementation that would use Symfony's [ProgressBar](http://symfony.com/doc/current/components/console/helpers/progressbar.html) to output progress.

I understand that this probably won't fit in with your plans for Composer, but I thought I'd send a pull request anyway just in case you think it is useful.

EDIT: 

A quick search of Pull Requests showed that people are already doing/done this, however this implementation might add something new, perhaps. :)

I realised it's failed tests. I'm working on a fix now.


salarmehr:
+1

neon64:
Closed in favor of #3275

Issue #3312, opened 2014-09-29
Feature Idea - Require-style script


alexandresalome:
I often create small scripts to show a sample, to POC or to automate some boring things. For example, by now, I need to create quickly a script that deals with Symfony Finder. I have to:
- Create a project directory (oh wait, it was just a script)
- Create a **composer.json**
- Run **composer install**
- Create my script and vendor

What if... I just create a script:

``` php
$composer->require('symfony/finder', function () {
     $finder = new Symfony\Component\Finder\Finder();
     # ...
});
```

And then run it using composer:

``` bash
composer run-script test.php
```

What's your advice on such feature?

Issue #2761, opened 2014-02-25
Make it possible for plugins to add commands


mpdude:
As the title reads.

Spun off from #2610.

Seldaek:
Fixed by https://github.com/composer/composer/pull/3377#issuecomment-215548870

mpdude:
Thanks 🍻

PR #3377, opened 2014-10-26
WIP - Allow plugins to register commands

### LINKED ISSUE #1234: custom commands
Just wanted to start this here – let's say I want to add a `till` command to composer. Now I realize no one else has a use for this command, but at the same time, I wouldn't want to maintain a fork where I backport changes from upstream when they start to collide.

Do you guys have any idea how we could add a new type `composer-command` which I can declare as a vendor and then it just registers that command?

Till

jderusse:
This PR will move plugins to a next level by exposing the application to the plugins (and also loads the plugins BEFORE the application starts).

It might be a start to solve few tickets like #3312 or https://twitter.com/lsmith/status/526053583834603520
ping @lsmith77

It should also fixes #1234, #2761

Sample of use
```
composer global require FriendsOfPHP/security-advisories
cd my/project
composer security-check
```

Want to simply test  it see a sample plugin (https://github.com/jeremy-derusse/composer-plugin)
```
composer global require jderuse/plugin
composer demo:greet
```

Todo
* [x] Add tests
* [x] Refactor the "plugins warmup" thing
* [ ] Avoid fatal errors when loading the plugin in an old instance of composer
* [ ] Add documentation about plugin creation

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -27,6 +27,7 @@ use Composer\IO\IOInterface;
 use Composer\IO\ConsoleIO;
 use Composer\Json\JsonValidationException;
 use Composer\Util\ErrorHandler;
+use Composer\Plugin\CommandsProviderInterface;
 
 /**
  * The console application that handles the commands
@@ -142,6 +143,10 @@ class Application extends BaseApplication
         if ($input->hasParameterOption('--profile')) {
             $startTime = microtime(true);
             $this->io->enableDebugging($startTime);
+        }
+
+        if (!$input->hasParameterOption('--no-plugins')) {
+            $this->addCommands($this->getPluginCommands());
         }
 
         $result = parent::doRun($input, $output);
@@ -317,6 +322,7 @@ class Application extends BaseApplication
     {
         $definition = parent::getDefaultInputDefinition();
         $definition->addOption(new InputOption('--profile', null, InputOption::VALUE_NONE, 'Display timing and memory usage information'));
+        $definition->addOption(new InputOption('--no-plugins', null, InputOption::VALUE_NONE, 'Whether to disable plugins.'));
         $definition->addOption(new InputOption('--working-dir', '-d', InputOption::VALUE_REQUIRED, 'If specified, use the given directory as working directory.'));
 
         return $definition;
@@ -331,5 +337,25 @@ class Application extends BaseApplication
         $helperSet->set(new DialogHelper());
 
         return $helperSet;
+    }
+
+    private function getPluginCommands()
+    {
+        $commands = array();
+
+        $composer = $this->getComposer(false, false);
+        if (null === $composer) {
+            $composer = Factory::createGlobal($this->io, false);
+        }
+
+        if (null !== $composer) {
+            foreach ($composer->getPluginManager()->getPlugins() as $plugin) {
+                if ($plugin instanceof CommandsProviderInterface) {
+                    $commands = array_merge($commands, $plugin->getCommands());
+                }
+            }
+        }
+
+        return $commands;
     }
 }
--- a/src/Composer/Plugin/CommandsProviderInterface.php
+++ b/src/Composer/Plugin/CommandsProviderInterface.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Plugin;
+
+use Composer\Composer;
+use Composer\IO\IOInterface;
+
+/**
+ * Commands Provider Interface
+ *
+ * @author Jérémy Derussé <jeremy@derusse.com>
+ */
+interface CommandsProviderInterface
+{
+
+    /**
+     * Retreives list of commands
+     *
+     * @return array
+     */
+    public function getCommands();
+}
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -289,7 +289,11 @@ class Factory
         $this->createDefaultInstallers($im, $composer, $io);
 
         if ($fullLoad) {
-            $globalComposer = $this->createGlobalComposer($io, $config, $disablePlugins);
+            $globalComposer = null;
+            if (realpath($config->get('home')) !== $cwd) {
+                $globalComposer = $this->createGlobalComposer($io, $config, $disablePlugins);
+            }
+
             $pm = $this->createPluginManager($io, $composer, $globalComposer);
             $composer->setPluginManager($pm);
 
@@ -351,15 +355,11 @@ class Factory
      * @param  Config        $config
      * @return Composer|null
      */
-    protected function createGlobalComposer(IOInterface $io, Config $config, $disablePlugins)
+    protected function createGlobalComposer(IOInterface $io, Config $config, $disablePlugins, $fullLoad = false)
     {
-        if (realpath($config->get('home')) === getcwd()) {
-            return;
-        }
-
         $composer = null;
         try {
-            $composer = self::createComposer($io, $config->get('home') . '/composer.json', $disablePlugins, $config->get('home'), false);
+            $composer = self::createComposer($io, $config->get('home') . '/composer.json', $disablePlugins, $config->get('home'), $fullLoad);
         } catch (\Exception $e) {
             if ($io->isDebug()) {
                 $io->writeError('Failed to initialize global composer: '.$e->getMessage());
@@ -487,5 +487,17 @@ class Factory
         $factory = new static();
 
         return $factory->createComposer($io, $config, $disablePlugins);
+    }
+
+    /**
+     * @param  IOInterface $io             IO instance
+     * @param  bool        $disablePlugins Whether plugins should not be loaded
+     * @return Composer
+     */
+    public static function createGlobal(IOInterface $io, $disablePlugins = false)
+    {
+        $factory = new static();
+
+        return $factory->createGlobalComposer($io, static::createConfig($io), $disablePlugins, true);
     }
 }
--- a/tests/Composer/Test/Plugin/Fixtures/plugin-v5/Installer/Plugin.php
+++ b/tests/Composer/Test/Plugin/Fixtures/plugin-v5/Installer/Plugin.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Installer;
+
+use Composer\Composer;
+use Composer\IO\IOInterface;
+use Composer\Plugin\PluginInterface;
+use Composer\Plugin\CommandsProviderInterface;
+
+class Plugin implements PluginInterface, CommandsProviderInterface
+{
+    public $version = 'installer-v5';
+
+    public function activate(Composer $composer, IOInterface $io)
+    {
+    }
+
+    public function getCommands()
+    {
+        return null;
+    }
+}
--- a/tests/Composer/Test/Plugin/Fixtures/plugin-v5/composer.json
+++ b/tests/Composer/Test/Plugin/Fixtures/plugin-v5/composer.json
@@ -1,0 +1,14 @@
+{
+    "name": "plugin-v5",
+    "version": "5.0.0",
+    "type": "composer-plugin",
+    "autoload": { "psr-0": { "Installer": "" } },
+    "extra": {
+        "class": [
+            "Installer\\Plugin"
+        ]
+    },
+    "require": {
+        "composer-plugin-api": "1.0.0"
+    }
+}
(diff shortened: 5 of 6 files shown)


src/Composer/Console/Application.php
```diff
@@ -328,4 +332,24 @@ protected function getDefaultHelperSet()
 
         return $helperSet;
     }
+
+    private function getPluginCommands()
+    {
+        $commands = array();
+        foreach ($this->getComposer()->getPluginManager()->getPackagedPlugins() as $packagedPlugin) {
+            if ($packagedPlugin->getPlugin() instanceof PluginCommandInterface) {
+                foreach ( $packagedPlugin->getPlugin()->getCommands() as $command) {
```
GromNaN: There is an unecessary space after `(`

src/Composer/Package/BasePackage.php
```diff
@@ -88,6 +90,14 @@ public function getPrettyName()
     /**
      * {@inheritDoc}
      */
+    public function getCompanyName()
```
GromNaN: It's usualy called the "vendor" name, not the "company" name.
https://getcomposer.org/doc/04-schema.md#name
> The name of the package. It consists of vendor name and project name, separated by `/`.

jeremy-derusse:
Take last remark into accounts

src/Composer/Package/PackageInterface.php
```diff
@@ -29,6 +29,13 @@
     public function getName();
 
     /**
+     * Returns the vendor's name extract from the first part of the package's name.
+     *
+     * @return string vendor name
+     */
+    public function getVendorName();
```
slbmeh: Not all types of packages make sense to have a vendor name.

src/Composer/Package/BasePackage.php
```diff
@@ -88,6 +90,14 @@ public function getPrettyName()
     /**
      * {@inheritDoc}
      */
+    public function getVendorName()
+    {
+        return current(explode(self::PACKAGE_NAME_SEPARATOR, $this->name));
```
slbmeh: IRT my comment on PackageInterface it may be better to return null if the count is 1. It would feel natural to be able to get a package short name by chopping the return value of this from the beginning of the return value on `getName()`. If there is no vendor prefix that would leave a null package name.

Seldaek:
@jeremy-derusse sorry for the delay reviewing this, but I don't think it should automatically namespace things.. This leads to a lot of complications in how plugins are managed, this PackagedPlugin stuff etc.. Lots of unnecessary code IMO.

I think all we need is a CommandProviderInterface so plugins that want to provide commands can do so and we can detect it.. We don't need to bump the composer-plugin-api version, or add any other weird handling.

Then for initializing plugins and taking care of directories without composer.json, you can attempt to call `Factory::create` and if that fails call `Factory->createGlobalComposer` (needs to add a `createGlobal` static public method to go with `create` so it's accessible). That way you get a fully loaded local composer + global, or simply a global one.

And additionally one thing to take care of is that if `--no-plugins` is passed as arg, this behavior should be disabled entirely.

Thanks for the effort though, and if you update taking this into consideration I'll try to merge ASAP :)

jeremy-derusse:
@Seldaek Thank you for taking times to review this PR.

By namespace, did you mean prefixing the commands with the vendor's name?

I think bumping the version is a requirement. Because if someone with an old version of composer install globaly a plugin wich implements `CommandProviderInterface` the installation will not failed, but each time, his composer will loads plugins (require, install, update, ...), he'll have an error "Fatal error:  Interface 'Composer\Plugin\CommandProviderInterface".
Wheras with bumping the version, the installation of the plugin will faill.

Seldaek:
Namespace yes I mean the vendor prefix in the command name, I think we should let plugins decide, it's more powerful.

Regarding the version that is indeed a problem but I think it's one we need to fix in a different way, we probably need to provide the API in version `<=1.1.0` and then load all plugins that require those versions, because technically you could install a new plugin and then run it with an old composer and it should just skip plugins that are installed but not loadable instead of crashing.. Anyway that's kind of a different issue that should probably be fixed before merging this PR :) I'd rather keep it separate.

jeremy-derusse:
I Take remarks into account with small modifications
- allow `createGlobalComposer` to be fully loaded (because we need a composer with pluginManager)
- add a method `Application::getLoadedComposer` wich try to load a localComposer then a globalComposer. I did not re-use `getComposer` to keep BC, because this method is used by many commands.

Seldaek:
Doing a full load with the global one means you'll load plugins twice if you have global plugins.. not sure how to avoid this cleanly though, I don't quite have time to investigate now.

jeremy-derusse:
Thanks to https://github.com/jeremy-derusse/composer/blob/plugin-with-commands/src/Composer/Plugin/PluginManager.php#L201, the plugin will registered only once.

jeremy-derusse:
BTW, I change `createComposer` to not merge the Composer instance with a Global one if they are on the same folder.

I'm waiting for a PR wich solve the api-version thing, then I'll finish this one.

slbmeh:
@Seldaek I was just looking at the double load issue for something unrelated, but my thoughts were to have an SplObjectStorage hash of packages updated and checked in `registerPackage` and bail early if it was already registered.

src/Composer/Plugin/CommandsProviderInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @author Jérémy Derussé <jeremy@derusse.com>
+ */
+interface CommandsProviderInterface
+{
+
+    /**
+     * Retreives list of commands
+     *
+     * @return array
```
nevvermind: Let's use `\Composer\Command\Command[]` here, instead of `array`.

src/Composer/Console/Application.php
```diff
@@ -332,4 +338,24 @@ protected function getDefaultHelperSet()
...
+
+        $composer = $this->getComposer(false, false);
+        if (null === $composer) {
+            $composer = Factory::createGlobal($this->io, false);
+        }
+
+        if (null !== $composer) {
+            foreach ($composer->getPluginManager()->getPlugins() as $plugin) {
+                if ($plugin instanceof CommandsProviderInterface) {
+                    $commands = array_merge($commands, $plugin->getCommands());
```
nevvermind: Should we check that `getCommands` actually returns an array of Command objects?

nevvermind:
@jderusse 
Do you think you can continue working on this? I can help if you want.

src/Composer/Console/Application.php
nevvermind: What about overlapping commands? Composer should _not_ do this without at least notifying the user, if not breaking.

jderusse:
@nevvermind 
regarding the comment of @Seldaek, I'm waiting for a new plugin loader who skip unloadable plugins.

nevvermind:
@jderusse - yeah, I' waiting for that, too :stuck_out_tongue: : https://github.com/composer/composer/pull/4089 and https://github.com/composer/composer/issues/4085

src/Composer/Console/Application.php
nevvermind: What about custom command added via scripts: https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands?

nevvermind:
@jderusse - you're in business: https://github.com/composer/composer/pull/4089 got merged.
Now you can bump the Plugin API version. 

But I'm not sure why this is not a _major_ bump ( :smile: ) to `2.0.0`: plugin implementing this new interface will break on an old Composer. This is not about Composer being BC-friendly by _adding_ an interface and not chaging its API, but breaking a new plugin who wishes to work with an older - and presumably compatible - version of Composer. 

A plugin which needs a `1.*` API is expected to work with `1.0.0`, too.

Taluu:
The thing is, composer is still [in alpha stage](https://github.com/composer/composer/releases/tag/1.0.0-alpha10), and usually alphas can break BC...

src/Composer/Console/Application.php
Seldaek: Just to be clear here, custom commands added via scripts are loaded at line 127 so they would already be in the list of active commands if you do conflict detection when adding the plugin commands.

Seldaek:
@nevvermind This highlights an interesting problem, that maybe we should have a plugin method like "getFeatures" that returns the features this plugin is capable of/implements. If we had this already, we could call getCommands() without matching on an interface but just checking getFeatures. That would mean that new plugins implementing new features would still load fine in older composer versions, they just wouldn't run with full capabilities.

That said, IMO bumping the interface to 1.1 is enough, as older composer releases that provide a 1.0 will not activate a plugin that requires 1.1. And the 1.1 interface doesn't break 1.0 plugins, so I don't see a BC break here.

jderusse:
@Seldaek Older version of composer dowloaded before #4089 will try to load the plugin regardless the version of the interface.

IMO, i'm not sure that replacing interfaces by a 'getFeatures' pattern is a good things :(

Seldaek:
Well yes the older versions of composer will break, but upgrades happen fairly quickly overall and once we release a 1.0 stable I think it's fine to tell people to upgrade if they don't want breakage and then forget about pre-1.0 versions.

As for getFeatures, yeah it's not as clean as interfaces, but it would allow building more flexible plugins.. I don't really care either way though, it's just a versioning problem for the plugin authors.

jderusse:
So, this solve the issue: Recents versions of composer will take care of the plugin version.
Each time a new plugin's feature is implemented in composer, we can create a new pretty interface and bumping the plugin's version.
Everything is fine right now ?

nevvermind:
@Taluu - I know what you're saying, but try adding a BC break in this without @Seldaek going medieval on your PR. :stuck_out_tongue_winking_eye: 

@Seldaek - Ha! Interesting that your brought that up. I was thinking something similar to your idea for a couple of days now, but I didn't know if you'd agree with it: http://java.dzone.com/print/4771

```php
// ----------- the API -------------------------
foreach ($pluginManager->getPlugins() as $plugin) {
    /**
     * Compatible with future dev, when going full PHP 5.5:
     * $plugin->hasCapability(Composer\Plugin\CommandProviderInterface::class)
     */
    if ($plugin->hasCapability('Composer\Plugin\Capability\CommandProvider')) {
        var_dump($plugin->getCommands());
    }

    // example of Composer offloading to plugins; not the greatests of examples...
    if ($plugin->hasCapability('Composer\Plugin\Capability\Downloader')) {
        $plugin->downloadPackage('some_package', '3.0.0');
    }
}

// ---------------------- the Plugin implementation -----------------
// MyPlugin.php
class MyPlugin implements Composer\Plugin\PluginInterface
{
    public function activete() {}

    public function getCapabilities()
    {
        return array(
            'Composer\Plugin\Capability\CommandProvider' => 'MyPlugin\\Commands',
            'Composer\Plugin\Capability\Downloader' => 'MyPlugin\\Downloader',
        );
    }
}

// Commands.php
class Commands implements Composer\Plugin\Capability\CommandProvider
{}

// Downloader.php
class Commands implements Composer\Plugin\Capability\Downloader
{}
```
If `hasCapability()` would presume a change in the Plugin interface API, this would need a bump to API 2.0.0. 
If that's to be avoided - and I think it should, for the `1.*` plugin API lifetime - you could use a helper/manager for to check the capability: `$pluginManager->pluginHasCapability($plugin, 'Composer\Plugin\CommandProviderInterface')`.

This is just a suggestion, mind you. The main advantage of this is that plugins will work with any API version, as the interfaces will not be loaded unless they're supported by the plugins, separating the Plugin declaration from the actual implementations it provides.

nevvermind:
@Seldaek @jderusse - did a small prototype/show-off: a branch which implements the above Capabilities design in Composer ([here](https://github.com/nevvermind/composer/commit/e33ed4753cccb4c4b0607ad6c4a320860b2cb6af)) and a plugin which adds a new custom command via the said Capabilities api ([here](https://github.com/nevvermind/composer-version-util-plugin)).

The whole thing is more like a glorified strategy pattern, but works pretty good and it's well decoupled (even in the plugin's own namespace), if I may say so myself :stuck_out_tongue: :
```
$ bin/composer
... standard stuff ...
 validate         Validates a composer.json
 version-util     You can finally find out what "^1.0" means!

$ bin/composer version-util "1.5.*"
[>= 1.5.0.0-dev < 1.6.0.0-dev]
```
What do you think?

nevvermind:
@stof , @alcohol - Can I get your opinion about this? (I need to lobby to avoid a stalled PR; hope it's ok)

alcohol:
I honestly don't have an opinion to offer at this time, as I pretty much never used plugins so far. Would have to dive into the changes to figure out what it is your are proposing.

nevvermind:
Come to think of it, this PR brings even more design work. It needed a proper Plugin API constraint. That's done. Now it would need a more robust Plugin _API_. Because this PR just provides an implementation of the CommandProvider interface, but it needs a good API to do this in the first place.

I think I'm going to move the needed API changes in another PR. 
@jderusse - don't go away! We still need this, ok? Brb.

Seldaek:
This should be rebased now and brought in line to use the new plugin capability feature. I can't really merge it until the capability stuff is in a composer release though because otherwise plugins implementing the Capable interface would break older composer.

Seldaek:
Merged this and fixed things up in https://github.com/composer/composer/commit/090295dbcb783dcbcc2dadcfd3a8539bb4ef8e71 & https://github.com/composer/composer/commit/ba909d8795c442307f81f244e31168b946f1159a. Still needs docs though if someone wants to pick that up (see https://github.com/composer/composer/issues/5262).

PR #3380, opened 2014-10-29
Escape delimiter characters in BaseExcludeFilter.

esenti:
When parsing .gitignore files, it's possible that the rule will contain a '#' character (eg. https://github.com/yiisoft/yii/blob/master/.gitignore#L37). As it's the same as the delimiter character, this will cause the built regexp to be invalid. To fix this issue, the '#' characters should be properly escaped.

I've found this issue when trying to setup Satis with the Yii package, which resulted in:

    [ErrorException]                    
        preg_match(): Unknown modifier '['

### DIFF
--- a/src/Composer/Package/Archiver/BaseExcludeFilter.php
+++ b/src/Composer/Package/Archiver/BaseExcludeFilter.php
@@ -140,8 +140,9 @@ abstract class BaseExcludeFilter
             $pattern .= '/';
         }
 
-        // remove delimiters as well as caret (^) and dollar sign ($) from the regex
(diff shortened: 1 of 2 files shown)


stof:
@reenl your suggestion will not work: it will escape the pattern, thus breaking the regex

denis-sokolov:
The proper fix for this belongs in [symfony/Glob::toRegex](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Finder/Glob.php#L47).

Seldaek:
Pretty sure this has since been fixed. At least the `\#*\#` now works fine.

PR #3390, opened 2014-11-03
Add support for archive compression (gz/bz2)

andrerom:
Adds support for file extensions tar.gz and tar.bz2 where these
implies use of compression as supported by `PharData` natively.

Compression support is not added to zip file as Phar doc explicitly
says it is not supported on file level, and compression would then
have to be exposed differently as it uses same file name.

File size was in my limited testing reduced to 1/8th, but this depends
on content of course.

Todo (if approach is ok):
- [ ] update tests

### DIFF
--- a/src/Composer/Package/Archiver/PharArchiver.php
+++ b/src/Composer/Package/Archiver/PharArchiver.php
@@ -24,22 +24,49 @@ class PharArchiver implements ArchiverInterface
         'tar' => \Phar::TAR,
     );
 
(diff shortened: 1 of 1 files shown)


GromNaN:
This is a "Nice to Have" feature. Duplicates PR #2706.

andrerom:
Close in favor of #3890 (zip files) and #2706 (tar.gz and tar.bz2).

Issue #3417, opened 2014-11-14
Failure installing Git package where branch/tag reference is also a versioned file


colinodell:
Git thinks the reference to the `0.12` tag is ambiguous and thus the `git reset` operations fails:

![image](https://cloud.githubusercontent.com/assets/202034/5047609/35191ec0-6be5-11e4-8aef-bcf7fcdb6ea8.png)

You can replicate this issue by using this `composer.json` file and running `composer install`:

    {
        "require": {
            "jgm/CommonMark": "0.12"
        },
        "repositories": [
            {
                "type": "package",
                "package": {
                    "name": "jgm/CommonMark",
                    "version": "0.12",
                    "source": {
                        "url": "https://github.com/jgm/CommonMark.git",
                        "type": "git",
                        "reference": "0.12"
                    }
                }
            }
        ]
    }

I believe this could probably be solved by adding ` --` to the end of the command like so:

    git checkout '0.12' && git reset --hard '0.12' --

Instead of:

    git checkout '0.12' && git reset --hard '0.12'

I currently have to reference commit hashes instead which aren't as user-friendly.

colinodell:
Any further thoughts on this issue or the pull request (#3418)?

PR #3418, opened 2014-11-14
Fix Git failures due to ambiguous argument (issue #3417)

colinodell:
This should fix issue #3417.

I've tested it against the `composer.json` that replicates the issue and confirmed the download works properly.


nevvermind:
Can we get this merged? Seems like an easy pick.

Seldaek:
Yup, sorry for the delay and thanks for the fix!

Issue #1663, opened 2013-03-06
Option for disabling autoload generating


shoomyth:
I'm trying to use composer as assets package manager.

I have 2 composer.json files: one is default for Symfony Standard Edition for php libraries and another is for managing assets only. Php's composer.json writes to "vendor/" directory and assets' one - to "vendor-assets/" dir. 

Assets libraries have only client-side files and no php, and in this case we don't need autoload data, but composer writes this data in any case into files vendor-assets/autoload.php and vendor-assets/composer/*.

It would be useful to have an option for disabling this

PR #3453, opened 2014-11-21
Skip autoloader generation during install and update

mattsah:
Added no-autoloader option for install and update which skips the autoloder generation.  This has been requested by multiple users and is useful for deployment situations where one might be running composer in two stages.

See: https://github.com/composer/composer/issues/1663

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -88,6 +88,7 @@ resolution.
   installation and show you what would happen.
 * **--dev:** Install packages listed in `require-dev` (this is the default behavior).
 * **--no-dev:** Skip installing packages listed in `require-dev`.
+* **--no-autoloader:** Skips autoloader generation.
 * **--no-scripts:** Skips execution of scripts defined in `composer.json`.
 * **--no-plugins:** Disables plugins.
 * **--no-progress:** Removes the progress display that can mess with some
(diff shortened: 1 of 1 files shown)


nifr:
awesome, thanks Jordi :+1:

Jany-M:
Is there any way this could be integrated on a per library basis, in `composer.json`?

eg.
"require-dev": {
       "dontneed/autoloader": {
             version: "1.2.3",
            "autoloader": false
       }
}

stof:
@Jany-M this does not make sense, as there is a single generated autoloader, not one per library

mvl22:
Is there any equivalent of the CLI `no-autoloader` as a directive within `composer.json` itself?

(I would have thought it should be the package maintainer who determines whether their package uses autoloading, rather than the deployment.)

Seldaek:
@mvl22 that flag's point is that you can run the dump-autoload at a later step because $reasons, but the point is not to have no autoloader at all, because IMO not using the Composer autoloader for the dependencies is kinda nonsensical overhead.

So no there is no config to persist that..

mvl22:
> not using the Composer autoloader for the dependencies is kinda nonsensical overhead

I'm thinking of the scenario where you have a CMS that has its own autoloader definition, but there are embedded applications within the site that use Composer. In this scenario, you don't want the CMS to have an autoload.php file generated because that will take precedence if an include_path has been set in the VirtualHost. Maybe this too specific a scenario for generic support ;)

Seldaek:
rm -rf vendor/composer is a bad idea™️. This contains installed.json which will make any further composer install go poorly if you wipe it. It may also contain packages from the composer vendor which would go even worse if they're needed.

If you really must delete files then delete `vendor/composer/autoload_*.php` and `vendor/composer/ClassLoader.php`, but really I don't see why you'd need to do that. Just don't include the vendor/autoload.php if you don't want to use it.

And include_path shenanigans yeahhh I wish that was not used anymore.. but I guess it still is :)

PR #3475, opened 2014-11-28
Converted to draft v4 schema

madskristensen:
Editors such as Visual Studio leverages JSON Schema draft v4, but this schema was authored in the old draft v3. With just a few minor changes, the schema is now draft v4.

* Added `$schema` property to state it is using schema draft v4
* Changed `required` properties to arrays instead of strings. This is the only real difference from v3 to v4 in this schema


Seldaek:
Ok cool, thanks @madskristensen

PR #3477, opened 2014-11-29
Fixes to binary installer for windows proxies

universalsoftware:
* generateWindowsProxyCode() and generateUnixyProxyCode() now generate valid proxy code on ANY platform
* do not create extra unixy proxy code on windows if it's not needed (cygwin not found)
* now on windows exe installation works like: program.exe => program.bat, program.exe.bat
* windows proxy code now do not use "call" command to execute exe files, it solves problem with wrong arguments processing with "%" and "^" chars
* use windows path delimiter in bat proxy files instead of unix path delimiter

Notes:
* extra shell wrappers for cygwin on windows break pattern: add ./vendor/bin in PATH and run application by short name like "application". The problem is by default windows try to run "application.exe" when you type "application" and windows find shell wrapper which didn't work. Possible solutions: change PATEXT to search for ".bat" before ".exe" or just fix that in composer to do not create shell wrappers on windows if cygwin not found and it was done in this update. 
* Also, on windows, when we try to execute binary we can type "program" (which will run "program.exe") or program.exe (which will run "program.exe". The same effect can be reached if we will create two batch wrappers for each executable: when we run "program" system will run "program.bat" (if  "program.exe" not found) and when we run "program.exe" system will run "program.exe.bat" (if  "program.exe" not found). Currently composer creates only "program.exe.bat" batch wrapper and in this commit implemented also creation of "program.bat" wrapper for ".exe" files.
* Also current implementation of windows batch wrapper contains defect: "call application.exe %*" is not applicable because application.exe will get all arguments escaped for passing to BATCH file! What side effects? Can't pass "%" and "^" in normal way. So the fix is to use "call" only for batch files and just run for all other cases.
* generateWindowsProxyCode() - fixed to use correct directory delimiter on windows
* generateUnixyProxyCode() - removed useless processing through ProcessExecutor::escape() which differs from escapeshellarg() only on windows and extra escaping capabilities required only for arguments (who will use & or % for binary filenames?)


universalsoftware:
Is there any feedbacks regarding pull request?

src/Composer/Installer/LibraryInstaller.php
```diff
@@ -297,7 +331,7 @@ protected function generateWindowsProxyCode($bin, $link)
         }
 
         return "@ECHO OFF\r\n".
-            "SET BIN_TARGET=%~dp0/".trim(ProcessExecutor::escape($binPath), '"')."\r\n".
+            "SET \"BIN_TARGET=%~dp0".str_replace("/", "\\", $binPath)."\"\r\n".
```
Seldaek: I don't really get this change.. that code was working just fine AFAIK, and now you don't escape anymore.

src/Composer/Installer/LibraryInstaller.php
```diff
@@ -220,16 +229,31 @@ protected function installBinaries(PackageInterface $package)
...
-                        $this->io->writeError('    Skipped installation of bin '.$bin.'.bat proxy for package '.$package->getName().': a .bat proxy was already installed');
+                    if ($this->isWindowsUnixyEnvironment()) {
+                        // add unixy support for cygwin and similar environments
+                        file_put_contents($link, $this->generateUnixyProxyCode($binPath, $link));
+                        @chmod($link, 0777 & ~umask());
+                    } else {
+                        $savedLink = $link;
+                        $link = $savedLink.'.bat';
+                        if (file_exists($link)) {
+                            $this->io->write('    Skipped installation of bin '.$bin.'.bat proxy for package '.$package->getName().': a .bat proxy was already installed');
```
Seldaek: All the io->write call here should be io->writeError

PR #3478, opened 2014-11-30
Suppress exceptions for deprecation notices, and simply output them to console instead

driskell:
Hi,

We're running composer in a test environment with minimum-stability set to dev and seeing failures since Symfony started adding deprecation messages a day or so ago. Composer currently handles these deprecation messages as errors and throws an ErrorException, causing the composer run to fail.

This patch causes E_DEPRECATED and E_USER_DEPRECATED (which Symfony is using) to simply output to standard out, or if console is initialised as a "<warning>text</warning>" message, and to not end the composer run with a failure.

Let me know if you believe a different approach is necessary - but I think deprecation notices are likely able to be ignored in dev stability. It might make moving forward from the deprecation notices easier too as it will output them all at once. It should be easy to extend this patch too to have an option to hide deprecation messages, even by default, if it's felt necessary.

Thanks,

Jason

### DIFF
--- a/tests/Composer/Test/Util/ErrorHandlerTest.php
+++ b/tests/Composer/Test/Util/ErrorHandlerTest.php
@@ -54,4 +54,16 @@ class ErrorHandlerTest extends TestCase
 
         @trigger_error('test', E_USER_NOTICE);
     }
+
+    /**
+     * Test ErrorHandler ignored deprecation notices
+     */
+    public function testErrorHandlerIgnoresDeprecationNotices()
+    {
+        ErrorHandler::register();
+
+        ob_start();
+        trigger_error('A user deprecation notice', E_USER_DEPRECATED);
+        ob_end_clean();
+    }
 }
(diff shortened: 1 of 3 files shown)


tests/Composer/Test/Util/ErrorHandlerTest.php
```diff
@@ -54,4 +54,16 @@ public function testErrorHandlerRespectsAtOperator()
...
+    /**
+     * Test ErrorHandler ignored deprecation notices
+     */
+    public function testErrorHandlerIgnoresDeprecationNotices()
+    {
+        ErrorHandler::register();
+
+        ob_start();
+        trigger_error('A user deprecation notice', E_USER_DEPRECATED);
+        ob_end_clean();
```
driskell: Hi @stof 

I did the initial changes. Just let me know when you get around to the above question and if the bit about the restoring of the error handler is OK.

Thanks

driskell:
Hi @stof - I appreciate you're probably busy. If you do find time to review again it'll be great to put this to bed. Thanks!

driskell:
I've now rebased on master.

Let me know if there's anything I can do to help with this PR.

driskell:
Is this still on the radar or best closing for now?
Admittedly it's more a non issue and just a code improvement that might help in future at this point.

Seldaek:
This was brought up somewhere else and I ended up implementing it a while back without seeing this PR sorry, but the good news is it's done, so I guess we can close this. (see https://github.com/composer/composer/commit/1753c275ffbfd4c96045218e1d1ada17282a0e0a)

PR #3480, opened 2014-12-01
Fix broken aliases for numeric branches. Fixes #3461

### LINKED ISSUE #3461: Problem with alias of numeric branches
We had recently attempted to develop a new versioning and git-management system with the intention of using composer branch-alias to point contributors to the correct branches. However, it doesn't seem that the branch-alias directive is working as intended.

We have an installer with a few dependent packages, each using the same branching scheme. The below alias code is used in both the installer, as well as cms and framework modules.

```json
"branch-alias": {
    "3.x-dev": "3.2.x-dev"
}
```

* silverstripe/installer: https://github.com/silverstripe/silverstripe-framework/blob/2aa1c1e11e8475ff6fc99f33e933143a193fc57e/composer.json
* silverstripe/framework: https://github.com/silverstripe/silverstripe-installer/blob/133e9d03328cc3082c5d5a42ebe0f2d2917875a1/composer.json
* silverstripe/cms: https://github.com/silverstripe/silverstripe-cms/blob/e9e95b9dbab5a4ff4a7774b8b7c1eff9fcd1576a/composer.json

The idea is that we would use `3` as a feature development branch, to be later forked into minor versions once the feature set is to be fixed (e.g. `3.2`). That way users could use `3.2.x-dev` as a version directive and it would work both prior to the feature stabilisation, as well as immediately afterwards. Likewise we had aliased `master` as the next major version (`4.0.x-dev`) for incompatible api changes. (A trial git workflow following [semantic versioning](http://semver.org/)).

When this code exists on the `3` branch [packagist](https://packagist.org/packages/silverstripe/installer) seems to correctly detect this alias, but attempts to either `composer create-project ... 3.2.x-dev` or `composer require 3.2.x-dev` either the installer or any of the necessary modules seems to fail with the below.

```
$ composer create-project silverstripe/installer ~/Sites/ss32test 3.2.x-dev
  [InvalidArgumentException]                                             
  Could not find package silverstripe/installer with version 3.2.x-dev.                       
```

```
$ composer require silverstripe/framework 3.2.x-dev

  Problem 1
    - The requested package silverstripe/cms 3.2.x-dev could not be found.
  Problem 2
    - The requested package silverstripe/framework 3.2.x-dev could not be found.

Potential causes:
 - A typo in the package name
   see <https://groups.google.com/d/topic/composer-dev/_g3ASeIFlrc/discussion> for more details.

Read <http://getcomposer.org/doc/articles/troubleshooting.md> for further common problems.
```

So I suppose one of the following is correct:

- Branch alias ONLY works for non-numeric branches (e.g. dev-testing)
- Branch aliasing for numeric branches is broken.
- I am silly for another unknown reason.

I'd appreciate any feedback. :)

tractorcow:
The issue seems to be that aliased branches were checked for `dev-` prefix, where numeric branches are named with `-dev` as a suffix instead.

Fixes https://github.com/composer/composer/issues/3461

### DIFF
--- a/doc/articles/aliases.md
+++ b/doc/articles/aliases.md
@@ -38,10 +38,14 @@ specifying a `branch-alias` field under `extra` in `composer.json`:
 }
 ```
 
-The branch version must begin with `dev-` (non-comparable version), the alias
-must be a comparable dev version (i.e. start with numbers, and end with
-`.x-dev`). The `branch-alias` must be present on the branch that it references.
-For `dev-master`, you need to commit it on the `master` branch.
+If you alias a non-comparible version (such as dev-develop) `dev-` must prefix the
+branch name. You may also alias a comparible version (i.e. start with numbers,
+and end with `.x-dev`), but only as a more specific version.
+For example, 1.x-dev could be aliased as 1.2.x-dev.
+
+The alias must be a comparable dev version, and the `branch-alias` must be present on
+the branch that it references. For `dev-master`, you need to commit it on the
+`master` branch.
 
 As a result, anyone can now require `1.0.*` and it will happily install
 `dev-master`.
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -224,7 +224,7 @@ class ArrayLoader implements LoaderInterface
      */
     public function getBranchAlias(array $config)
     {
-        if ('dev-' !== substr($config['version'], 0, 4)
+        if (('dev-' !== substr($config['version'], 0, 4) && '-dev' !== substr($config['version'], -4))
             || !isset($config['extra']['branch-alias'])
             || !is_array($config['extra']['branch-alias'])
         ) {
@@ -245,6 +245,14 @@ class ArrayLoader implements LoaderInterface
 
             // ensure that it is the current branch aliasing itself
             if (strtolower($config['version']) !== strtolower($sourceBranch)) {
+                continue;
+            }
+
+            // If using numeric aliases ensure the alias is a valid subversion
+            if(($sourcePrefix = $this->versionParser->parseNumericAliasPrefix($sourceBranch))
+                && ($targetPrefix = $this->versionParser->parseNumericAliasPrefix($targetBranch))
+                && (stripos($targetPrefix, $sourcePrefix) !== 0)
+            ) {
                 continue;
             }
 
--- a/src/Composer/Package/Loader/ValidatingArrayLoader.php
+++ b/src/Composer/Package/Loader/ValidatingArrayLoader.php
@@ -251,6 +251,17 @@ class ValidatingArrayLoader implements LoaderInterface
                     if ('-dev' !== substr($validatedTargetBranch, -4)) {
                         $this->warnings[] = 'extra.branch-alias.'.$sourceBranch.' : the target branch ('.$targetBranch.') must be a parseable number like 2.0-dev';
                         unset($this->config['extra']['branch-alias'][$sourceBranch]);
+
+                        continue;
+                    }
+
+                    // If using numeric aliases ensure the alias is a valid subversion
+                    if(($sourcePrefix = $this->versionParser->parseNumericAliasPrefix($sourceBranch))
+                        && ($targetPrefix = $this->versionParser->parseNumericAliasPrefix($targetBranch))
+                        && (stripos($targetPrefix, $sourcePrefix) !== 0)
+                    ) {
+                        $this->warnings[] = 'extra.branch-alias.'.$sourceBranch.' : the target branch ('.$targetBranch.') is not a valid numeric alias for this version';
+                        unset($this->config['extra']['branch-alias'][$sourceBranch]);
                     }
                 }
             }
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -170,6 +170,21 @@ class VersionParser
     }
 
     /**
+     * Extract numeric prefix from alias, if it is in numeric format, suitable for
+     * version comparison
+     *
+     * @param string $branch Branch name (e.g. 2.1.x-dev)
+     * @return string|false Numeric prefix if present (e.g. 2.1.) or false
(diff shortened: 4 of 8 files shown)


tractorcow:
Excuse me, @Seldaek, would you be able to review this for me please? At least let me know if there's anything else I should do in order to get this merged. :)

Seldaek:
@tractorcow sorry for the wait, but I don't actually see this as a valid use case. IMO if you have a numeric branch you should just use the right number and not need an alias. The alias thing was mostly made for the `master` branch.

Why not call your branch 3.0.x and when 3.1.x comes about add a new 3.1.x branch? Or just use master with a 3.0.x-dev alias and then the master branch can evolve over time?

I find it rather misleading to allow a numeric branch to have another actual numeric value.

camfindlay:
@tractorcow perhaps go into some context of the real world use case you have here.

tractorcow:
@Seldaek that's a perfectly valid question and one I'd love to answer.

We are introducing a new contribution practise and custom git workflow, based on semantic versioning, and the problem this is attempting to solve is to always make it perfectly clear which branch certain changes should be pulled against, and to better managed contributions during release cycles.

Our ideal workflow is for breaking changes to always go against master, api introductions to always go against major version branches (3, 4, etc) and bugfixes to always go against minor versions (3.1, 3.2, 4.1, etc).

We could (and will have to, if this PR is rejected) revert to using only minor branch names (e.g. 3.3, instead of 3 when 3.3.0 is yet unreleased), but it adds an additional layer of complexity:

* What happens to pull requests against 3.3 that aren't merged by the time 3.3.0 is released. We would  need to require the PR re-submitted against 3.4 (created on 3.3 release).
* When new PR are introduced we would need some other way of flagging a minor release as allowing new API. Saying "only the highest minor version" would potentially work, but would likely increase the rate of PR targeting the wrong version, and requires more explanation to contributors. We want to help our contributors as much as possible. :D
* We would like to provide support for major-only specified dependencies. E.g., some users would like to select 3.x-dev as a framework requirement, such as core developers and contributors, testers, and those brave enough who like to work on the bleeding edge (and I've personally see a few work this way). That way they wouldn't have to update their minor version every release, if that's the level of stability they have chosen. Their only other option is to use dev-master, and that's not always feasible. Using only master and minor version branches doesn't support those people well.

As a note, packagist seems to report that the alias works, but the given command fails (see 3.x-dev) https://packagist.org/packages/silverstripe/framework. Aliasing also doesn't seem to be documented as non-numeric only, so it would be great to clarify the rules there. :)

I don't think it's misleading to have one numeric branch an alias of another, as long as the alias is a subversion of the aliased. (e.g. aliasing 5 as 5.5 should be fine, but maybe aliasing 6 as 5.5 should be an error?), wouldn't you agree? I'd be happy to update the PR to enforce this constraint.

Thanks greatly for your feedback, and for taking the time to read this through.

Seldaek:
First of all regarding the PR.. *maybe* yes allowing subset of branch name would be ok, but let's first get to the bottom of the use case.

- PRs against 3.3 not merged yet, if they're bugfixes they still belong to the 3.3 branch, if they're features well yes maybe they need to be reopened, but probably they need a rebase and some amount of work anyway. Or the maintainers can just merge it locally against 3.4 or whatever. It's not such a huge issue IMO, very large projects like symfony pull it off very well with master (aliased to 3.0.x) + 2.3/2.6 branches (at time of writing) for bugfix releases + a 2.7 branch for the upcoming 2.7.0.
- why not do like (almost?) everyone else and use master as *the next minor OR major release* depending on context? I think that consistency across projects is also an help for contributors. If I have to understanding your branching model before sending a PR it's weird, if I can just send new features to master it's easy. BC breaking releases don't happen very often, what I usually do is store those PRs in a 2.0 milestone until I am ready to start work on the 2.0 at which point master becomes 2.0 and those PRs can be merged.
- This is already possible, you just require `~3.4@dev` for example and that will take 3.4.0 up to 3.999.999, including dev versions, so when the master branch's alias changes from 3.4.x-dev to 3.5.x-dev, you will keep installing from master, but when you decide to start work on a 4.0 people with the right requires will stay stuck on the (newly created) 3.6 branch for example, as it will be the highest version matching `~3.4`. The master branch won't be upgraded to, and you should avoid requiring `dev-master` for exactly this reason.

As for packagist that seems like a bug because it just looks at the branch alias data and doesn't do any checks.

I can see that you thought this through but IMO these problems are already solved by many other projects and maybe falling in line would have more value than introducing a new model. If you don't well see first line of this comment :)

tractorcow:
Well I think you've answered my questions pretty well.

* We could, and we are getting better at supporting contributors a lot more in regards to rebasing and moving pull requests for them, but at the same time we want to support the community managers and core-devs to make it easy for them to help others. You're right, it's not a huge issue, but it is a quality of life improvement. We have a lot of pull requests to manage, and we are all very lazy and like things to be simple. :)
* The problem is that we often want to work on big breaking changes at the same time as normal features. There are always two places to send new features, depending on (1) if it's breaking, or (2) if it's non-breaking, and (2) could potentially be multiple places depending on which minor version it's against. :) We've discussed this openly in the community, and the community itself thought this would be the best approach; In effect it's something they had a part in making, in a way that made sense to them.
* Good point on the third one, I should have thought about that problem a bit more.

I'm happy for you to make the call. If you'd like me to close this pull request, I'd be happy to clarify the decision in the documentation (to prevent my mistake being made by others in the future).

If we have to live with only master and major-dot-minor branches then we would probably survive ok. :)

tractorcow:
I've implemented the version comparison check as a warning, meaning it'll be picked up by `composer validate`, and have documented the rules in the documentation. Invalid aliases should be skipped.

Hopefully I've narrowed the possible use of this feature to only the valid use case I've given (aliasing a version as a more specific one).

Seldaek:
Thanks, I think that's fine :) I also patched it to allow 1.0-dev as well as 1.0.x-dev since we support both.

tractorcow:
Oh thank you @Seldaek! Really appreciate this.

I really learned a lot more about composer through working on this, especially on the second iteration. I hope I can get more opportunities to contribute in the future. :)

PR #3503, opened 2014-12-03
Comment GC disable and enable GC after run

### LINKED ISSUE #3488: Consider re-enabling GC after solving
It would be great to enable GC again after solving dependencies, that way package installation and post-cmd hooks etc can run with GC and hopefully less memory usage so less chance of peaking into swap-land.

Doing this properly though would require that we make sure the entire solver data, pool, etc except for the packages contained in the operations are freed and contain as little cycles as possible, otherwise we risk hitting huge GC loads again. /refs #3482

Z1ni:
Fixes #3488

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -160,6 +160,7 @@ class Installer
      */
     public function run()
     {
+        // Disable garbage collection to save memory and time
         gc_collect_cycles();
         gc_disable();
 
@@ -335,6 +336,8 @@ class Installer
                 touch($vendorDir);
             }
         }
+
+        gc_enable();
 
         return 0;
     }


src/Composer/Installer.php
```diff
@@ -336,6 +337,8 @@ public function run()
             }
         }
 
+        gc_enable();
```
allan-simon: which kind of warning ? is it something that can be handled by wrapping it in a "if not hhvm" ?

stof: yes it can

allan-simon: ok, i've quickly check and it can be done with 

```
if (defined('HHVM_VERSION')) {
    // Code
}
```
(dunno if it helps)

marc-mabe:
` gc_enabled` should be used to only disable GC before run and re-enable after only if it was enabled before.

src/Composer/Installer.php
```diff
@@ -160,6 +160,7 @@ public function __construct(IOInterface $io, Config $config, RootPackageInterfac
      */
     public function run()
     {
+        // Disable garbage collection to save memory and time
```
stof: it is not saving memory. It is about saving time because the GC will be triggered many times (because of the amount of Zval used in the solver for the package graph) for cases where it cannot collect anything (the refcount will take care of all the package graph objects btw)

PR #3517, opened 2014-12-08
Added deprecated warning for the dev option

WouterJ:
The option isn't used for a long time already, but there was no warning about the usage of it. This could lead to confusion, as people could be thinking that no-dev is the default.

### DIFF
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -74,6 +74,10 @@ EOT
             $input->setOption('no-plugins', true);
         }
 
+        if ($input->getOption('dev')) {
+            $output->writeln('<warning>You are using the deprecated option "dev". Dev packages are installed by default now.</warning>');
+        }
+
         $composer = $this->getComposer(true, $input->getOption('no-plugins'));
         $composer->getDownloadManager()->setOutputProgress(!$input->getOption('no-progress'));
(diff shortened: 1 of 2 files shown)


WouterJ:
Thanks @stof, I've updated it now

Seldaek:
How about doing the same to the update command?

naderman:
@WouterJ bump

WouterJ:
I'm sorry for the delay. I've now added it for the update command as well, squashed commits and fixed the build error.

src/Composer/Command/InstallCommand.php
```diff
@@ -74,6 +74,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $input->setOption('no-plugins', true);
         }
 
+        if ($input->getOption('dev')) {
```
stloyd: Why not `hasOption()`?

WouterJ: `hasOption()` only checks if the InputDefinition contains the option you specify. Since we define the option in the definition, we are sure the definition has the option. What you want is check if it's used. If it's used, it'll be set to `true`. Otherwise, it's value will be `null`.

stof:
:+1:

Seldaek:
Thanks

PR #3522, opened 2014-12-08
Use caret operator for 0.x releases when guessing versions, fixes #3518

### LINKED ISSUE #3518: Require - Guessing versions for 0.x releases
Since recently, it is possible to just run `composer require vendor/package` and let Composer decide what is best. So this is currently in the tests:

```
// real version, is dev package, stability, expected recommendation, [branch-alias]
array('1.2.1', false, 'stable', '~1.2'),
array('1.2', false, 'stable', '~1.2'),
array('v1.2.1', false, 'stable', '~1.2'),
array('3.1.2-pl2', false, 'stable', '~3.1'),
array('3.1.2-patch', false, 'stable', '~3.1'),
```

That is fine according to SemVer, but 0.x is probably different. http://semver.org/ FAQ:

> **How should I deal with revisions in the 0.y.z initial development phase?**
> The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.

What I usually see is that packages start at 0.1.x and when the API changes signifcantly, go to 0.2.x. So would it for 0.x release be better to be more restrictive? For example:

`0.1.2` -> `0.1.x` or `~0.1.0` instead of `~0.1`.

Seldaek:
Should not be merged until the ^ op is in a release IMO.

### DIFF
--- a/src/Composer/Package/Version/VersionSelector.php
+++ b/src/Composer/Package/Version/VersionSelector.php
@@ -109,13 +109,8 @@ class VersionSelector
         if (count($semanticVersionParts) == 4 && preg_match('{^0\D?}', $semanticVersionParts[3])) {
             // remove the last parts (i.e. the patch version number and any extra)
             if ($semanticVersionParts[0] === '0') {
-                if ($semanticVersionParts[1] === '0') {
-                    $semanticVersionParts[3] = '*';
-                } else {
-                    $semanticVersionParts[2] = '*';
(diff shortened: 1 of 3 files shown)


hairmare:
Please don't merge this without tagging ``1.0.0-alpha10``. Even without automatically using the caret operator for 0.x releases it has been cropping up all over the place in issues (ie. #3877 and #3779).

PR #3531, opened 2014-12-10
allow package type to be set with init command

aaemnnosttv:
This PR adds the ability to set the package's type from the `composer init` command.

Adds support for `help init` command output, `init --type` flag, and interactive input.

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -65,6 +65,7 @@ class InitCommand extends Command
                 new InputOption('description', null, InputOption::VALUE_REQUIRED, 'Description of package'),
                 new InputOption('author', null, InputOption::VALUE_REQUIRED, 'Author name of package'),
                 // new InputOption('version', null, InputOption::VALUE_NONE, 'Version of package'),
+                new InputOption('type', null, InputOption::VALUE_OPTIONAL, 'Type of package'),
                 new InputOption('homepage', null, InputOption::VALUE_REQUIRED, 'Homepage of package'),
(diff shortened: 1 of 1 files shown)


GromNaN:
:+1:

src/Composer/Command/InitCommand.php
```diff
@@ -258,6 +259,14 @@ function ($value) use ($self, $minimumStability) {
         );
         $input->setOption('stability', $minimumStability);
 
+        $type = $input->getOption('type') ?: false;
+        $type = $dialog->ask(
+            $output,
+            $dialog->getQuestion('Type', $type),
```
GromNaN: The question should be "Package Type".
We need to choose a default type between "project" and "library".

src/Composer/Command/InitCommand.php
```diff
@@ -65,6 +65,7 @@ protected function configure()
                 new InputOption('description', null, InputOption::VALUE_REQUIRED, 'Description of package'),
                 new InputOption('author', null, InputOption::VALUE_REQUIRED, 'Author name of package'),
                 // new InputOption('version', null, InputOption::VALUE_NONE, 'Version of package'),
+                new InputOption('type', null, InputOption::VALUE_OPTIONAL, 'Type of package'),
```
GromNaN: You can define the default value as 5th argument.

src/Composer/Command/InitCommand.php
aaemnnosttv: Thanks for looking over my PR.  I've updated the question wording.

Composer will treat the package as a library if no type is set, is that correct?
If so, then library would seem like the easy choice there.

Does it really need a default value though since Composer will already assume that if it's missing?

GromNaN: We just need to be sure that if the question is skipped, then the line "type" doesn't appear in the file.

stof:
The list of commits in your PR looks really weird.

Seldaek:
Thanks

PR #3548, opened 2014-12-12
Add the ability to show and confirm changes before doing them.

ChadSikorra:
This adds the ability to do a `--confirm` on an update and be shown what the changes will be and confirm them. I implemented this because I like to see what will happen prior to everything happening. I also implemented a collection class around the different operation type classes, as I thought this made things easier to understand  as I implemented this and extracted some of the code from the Installer class. Something I wasn't too sure about: Is the order I'm returnining the operation types in  `sortOperationsToArray` in the OperationCollection class correct?

### DIFF
--- a/src/Composer/DependencyResolver/Operation/InstallOperation.php
+++ b/src/Composer/DependencyResolver/Operation/InstallOperation.php
@@ -61,6 +61,14 @@ class InstallOperation extends SolverOperation
      */
     public function __toString()
     {
-        return 'Installing '.$this->package->getPrettyName().' ('.$this->formatVersion($this->package).')';
+        return 'Installing '.$this->getChangeAsString();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getChangeAsString()
+    {
+        return $this->package->getPrettyName().' ('.$this->formatVersion($this->package).')';
     }
 }
--- a/src/Composer/DependencyResolver/Operation/InstallOperation.php
+++ b/src/Composer/DependencyResolver/Operation/InstallOperation.php
@@ -21,6 +21,11 @@ use Composer\Package\PackageInterface;
  */
 class InstallOperation extends SolverOperation
 {
+    /**
+     * The operation type.
+     */
+    const TYPE = 'install';
+
     protected $package;
 
     /**
@@ -47,13 +52,11 @@ class InstallOperation extends SolverOperation
     }
 
     /**
-     * Returns job type.
-     *
-     * @return string
+     * {@inheritDoc}
      */
     public function getJobType()
     {
-        return 'install';
+        return self::TYPE;
     }
 
     /**
--- a/src/Composer/DependencyResolver/Operation/OperationCollection.php
+++ b/src/Composer/DependencyResolver/Operation/OperationCollection.php
@@ -22,15 +22,25 @@ use Composer\Repository\PlatformRepository;
 class OperationCollection implements \IteratorAggregate
 {
     /**
-     * @var array The operations sorted by type
+     * The 'plugin' operation type that must be executed before all others.
      */
-    private $operations = array(
-        'install' => array(),
-        'update' => array(),
-        'uninstall' => array(),
-        'plugin' => array(),
-        'markAliasInstalled' => array(),
-        'markAliasUninstalled' => array(),
+    const TYPE_PLUGIN = 'plugin';
+
+    /**
+     * @var SolverOperation[] The operations, minus the uninstalls and plugins.
+     */
+    private $operations = array();
+
+    /**
+     * @var array The operations by type so they can be selectively retrieved.
+     */
+    private $operationTypes = array(
+        self::TYPE_PLUGIN => array(),
+        UninstallOperation::TYPE => array(),
+        UpdateOperation::TYPE => array(),
+        InstallOperation::TYPE => array(),
+        MarkAliasUninstalledOperation::TYPE => array(),
+        MarkAliasInstalledOperation::TYPE => array(),
     );
 
     /**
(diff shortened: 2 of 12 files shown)


src/Composer/DependencyResolver/Operation/InstallOperation.php
```diff
@@ -61,6 +61,14 @@ public function getJobType()
...
     public function __toString()
     {
-        return 'Installing '.$this->package->getPrettyName().' ('.$this->formatVersion($this->package).')';
+        return 'Installing '.$this->getChangeAsString();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getChangeAsString()
```
GromNaN: Why not just use `__toString` ?

ChadSikorra: If I recall correctly, the reason I did that was because the default output for `__toString()` made it difficult to actually tell what was happening in the context of seeing what was actually changing. I felt like there was probably a better way to handle it but couldn't come up with anything at the time I wrote all this. Or maybe it looks fine with the normal output and I'm just weird? hah. Always a possibility.

stof: @GromNaN we already have a string representation (see above).

src/Composer/DependencyResolver/Operation/InstallOperation.php
```diff
@@ -61,6 +61,14 @@ public function getJobType()
...
+    {
+        return $this->package->getPrettyName().' ('.$this->formatVersion($this->package).')';
```
stof: this is not consistent with other operations as the verb ``Installing`` is not there

src/Composer/DependencyResolver/Operation/UninstallOperation.php
```diff
@@ -61,6 +61,14 @@ public function getJobType()
...
-        return 'Uninstalling '.$this->package->getPrettyName().' ('.$this->formatVersion($this->package).')';
+        return 'Uninstalling '.$this->getChangeAsString();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getChangeAsString()
+    {
+        return $this->package->getPrettyName().' ('.$this->formatVersion($this->package).')';
```
stof: this has the same representation than an InstallInformation, which makes the configurmation ambiguous. Why not just using the string representation of the operation ?

src/Composer/DependencyResolver/Operation/OperationCollection.php
```diff
@@ -0,0 +1,191 @@
...
+    /**
+     * @return array The plugin operations.
+     */
+    public function getPlugins()
+    {
+        return $this->operations['plugin'];
+    }
+
+    /**
+     * @return array The markAliasInstalled operations.
```
stof: you should make the type more explicit for all these methods (``@return OperationInterface[]``, or even better the appropriate implementation each time)

src/Composer/Installer.php
```diff
@@ -602,67 +606,39 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
...
-        }
-
-        return array_merge($installerOps, $operations);
+        $this->io->write('');
+        $this->showChanges('The following plugins will be installed/updated:', $operations->getPlugins());
+        $this->showChanges('The following packages will be uninstalled:', $operations->getUninstalls());
+        $this->showChanges('The following packages will be installed:', $operations->getInstalls());
+        $this->showChanges('The following packages will be updated:', $operations->getUpdates());
+        $this->showChanges(
+            'The following alias changes will occur:',
```
stof: this does not need to be displayed IMO. Alias marking is quite internal (most people won't understand what it is about)

src/Composer/DependencyResolver/Operation/UninstallOperation.php
ChadSikorra: Same reason I previously commented on for @GromNaN . The uninstalls, installs, etc are grouped together in the confirmation output. When there are many of them grouped together the verb prefix becomes repititive and made it harder to read and just weird to look at in general, in my opinion anyway.

ChadSikorra:
Ugh. I may have to re-do this haha. I attempted to rebase after adding tests and making some suggested changes but apparently I did it incorrectly. My appoligies, I'm not too familiar with updating a pull request with git yet.

src/Composer/DependencyResolver/Operation/OperationCollection.php
ChadSikorra: Very cool, I did not even realize that was a supported syntax. I've updated the return types.

ChadSikorra:
Ok, I think I fixed my PhpStorm rebase snafu...hah. And learned a little bit in the process. Still some things to resolve, such as not showing/confirming alias operations suggested by @stof. And it seems like tests are failing now, though I'm not sure if the failures are directly related to these changes. I'll have to take a closer look.

ChadSikorra:
Ok, I corrected most of the failures. They were due to the new `OperationCollection` class not returning an array of operations based solely on the order they were added (something I was unsure of in my comment for this PR actually). As I'm unsure how much that really matters, I decided to keep the old behavior of how the operations were sorted in an array and only move the uninstall/plugin types to the front (that's how it's currently being done anyway). However, the collection class still maintains an array of operations by type for easy retrieval.

I also felt that sprinkling strings around that define the operation type was looking a little messy, hence the constant `TYPE` definition on the operation classes. Still a few things left to do, but if anyone has feedback on the recent changes let me know.

Seldaek:
Closing as I still don't think this is really needed, now that we have composer outdated on top of dry-run updates.

PR #3552, opened 2014-12-14
Updated satis archive documentation

CommandZ-io:
Added clarification on how archive downloads work when using ssh.


naderman:
Sorry I'm not sure this clarifies anything what do you mean by "if you use ssh"?

GromNaN:
Simplier:

> Whatever is your source repository, the archive files generated by Satis will be accessible over HTTP or HTTPS at the URL you specified as `prefix-url`. 
> In order to restrict access to these files, you can configure basic HTTP authentication on your webserver. Composer will then ask for username/password when installing a package from its dist archive.

PR #3571, opened 2014-12-18
Search command option -R / --repo to restrict search to single repository

gohanman:
I find it useful when dealing with a number of VCS repositories. More often than not each VCS repo is providing a single dependency. Redownloading package lists for all the repos can create a noticeable hang when all I really want to search is packagist.

### DIFF
--- a/src/Composer/Command/SearchCommand.php
+++ b/src/Composer/Command/SearchCommand.php
@@ -41,6 +41,7 @@ class SearchCommand extends Command
             ->setDescription('Search for packages')
             ->setDefinition(array(
                 new InputOption('only-name', 'N', InputOption::VALUE_NONE, 'Search only in name'),
+                new InputOption('repo', 'R', InputOption::VALUE_OPTIONAL, 'Search just this repo'),
                 new InputArgument('tokens', InputArgument::IS_ARRAY | InputArgument::REQUIRED, 'tokens to search for'),
             ))
             ->setHelp(<<<EOT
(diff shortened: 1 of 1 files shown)


naderman:
You have to specify the integer offset? This seems very unintuitive and difficult to understand?

src/Composer/Command/SearchCommand.php
```diff
@@ -60,6 +61,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $localRepo = $composer->getRepositoryManager()->getLocalRepository();
             $installedRepo = new CompositeRepository(array($localRepo, $platformRepo));
             $repos = new CompositeRepository(array_merge(array($installedRepo), $composer->getRepositoryManager()->getRepositories()));
+            if ($input->getOption('repo')) {
```
stloyd: `hasOption()`.

src/Composer/Command/SearchCommand.php
```diff
@@ -60,6 +61,18 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                $search_repo = $input->getOption('repo');
```
stloyd: You should follow CS and use camelCased names for variables.

gohanman:
RE: Integer offsets: it depends how your composer.json is written. If repos are defined like this, then yes it would be integers.
```json
"repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/igorw/monolog"
        }
    ],
```

Where as if repos are defined like this you can use names:
```
"repositories": {
       "igorw":  {
            "type": "vcs",
            "url": "https://github.com/igorw/monolog"
        }
    },
```

Composer itself accepts either format (or at least did last I checked)

PR #3580, opened 2014-12-21
Delegate problem message formatting to a dedicated class

Pherserk:
Exception was tought to know the installed map, now installed map is a dependency of the problem formatter.
More than one potential failure cause can be noticed now.
Sorry, surely it needs some evolution or refactoring, just a little step over the goal.

### DIFF
--- a/src/Composer/DependencyResolver/ProblemFormatter.php
+++ b/src/Composer/DependencyResolver/ProblemFormatter.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Composer\DependencyResolver;
+
+class ProblemFormatter
+{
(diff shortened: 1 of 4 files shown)


naderman:
Please stick to max 120 char line lengths, you may want to use heredoc syntax for long strings.

On naming: The word "advice" does not have a plural, so please rename that.

naderman:
I would go with "explanation" instead of "advice".

Pherserk:
Right

Pherserk:
@Naderman I did another p.r. addressing the problem in a better integrated way: https://github.com/composer/composer/pull/3693
This means that with a ProblemFormatter we can UnitTest the ProblemFormatter instead of being forced to make integration test verifying error messages.
I wuold be glad to help.

PR #3584, opened 2014-12-22
Simplify the way to assert parse error in json tests

stof:
The advantage is that we now have a meaningful error message when no parse error is detected instead of having an empty message.

Note that one of the test about parse errors is failing for me locally (no exception thrown), which is how I detected the empty failure message


Seldaek:
Fixed by d7f741824339890d2a903852ae2914c92f379cb7 in a better way since it keeps checking $text

stof:
My PR was checking ``$text`` too. See the second argument of ``setExpectedException``

Seldaek:
Oh right, well anyway it's been fixed and this conflicted so not worth upgrading I guess.

PR #3588, opened 2014-12-23
Fix self-update permissions

jderusse:
This PR fixes a wrong assertions on permission in the `self-update` command.
When composer rename the temp file to original file, it actualy check if the file is writable. Wich is partialy wrong:
- If the temp file is in the same partition than the original file, then the directory containg the file have to be writable.
- If the temp file is in the different partition (filesystem) than the original file, the rename function will in fact use a copy + delete. In this case, the file have to be writable and the directory have not to be writable.

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -72,8 +72,12 @@ EOT
         if (!is_writable($tmpDir)) {
             throw new FilesystemException('Composer update failed: the "'.$tmpDir.'" directory used to download the temp file could not be written');
         }
-        if (!is_writable($localFilename)) {
-            throw new FilesystemException('Composer update failed: the "'.$localFilename.'" file could not be written');
+
+        $tempFilename = $tmpDir . '/' . basename($localFilename, '.phar').'-temp.phar';
+
+        // check if the the user has permission for renaming file
+        if (!$this->hasRenamePermission($tempFilename, $localFilename)) {
+            throw new FilesystemException('Composer update failed: the "'.$localFilename.'" file could not be overwritten');
         }
 
         if ($input->getOption('rollback')) {
@@ -95,7 +99,6 @@ EOT
             return 0;
         }
 
-        $tempFilename = $tmpDir . '/' . basename($localFilename, '.phar').'-temp.phar';
         $backupFile = sprintf(
             '%s/%s-%s%s',
             $rollbackDir,
@@ -158,10 +161,12 @@ EOT
         if (!is_readable($old)) {
             throw new FilesystemException('Composer rollback failed: "'.$old.'" could not be read');
         }
+        if (!$this->hasRenamePermission($old, $localFilename)) {
+            throw new FilesystemException('Composer rollback failed: the "'.$newFilename.'" file could not be overwritten');
+        }
 
-        $oldFile = $rollbackDir . "/{$rollbackVersion}" . self::OLD_INSTALL_EXT;
         $output->writeln(sprintf("Rolling back to version <info>%s</info>.", $rollbackVersion));
-        if ($err = $this->setLocalPhar($localFilename, $oldFile)) {
+        if ($err = $this->setLocalPhar($localFilename, $old)) {
             $output->writeln('<error>The backup file was corrupted ('.$err->getMessage().') and has been removed.</error>');
 
             return 1;
@@ -221,5 +226,58 @@ EOT
             ->in($rollbackDir);
 
         return $finder;
+    }
+
+    /**
+     * Determine if renaming source to target is allowed
+     *
+     * Permissions needed for renaming a file is not the same when the source and the targe are on the same partion
+     * not. When files are on different paritions, PHP will copy + delete the the source
+     * see https://bugs.php.net/bug.php?id=54097
+     *  - To rename a file you need write permission on the directory
+     *  - To copy a file you need write permission on the file
+     *
+     * @param string $source
+     * @param string $target
+     *
+     * @return bool
+     */
+    private function hasRenamePermission($source, $target)
+    {
+        if ($this->isOnSamePartition($source, $target)) {
+            if (is_writable(dirname($target))) {
+                return true;
+            }
+        } elseif (is_writable($target)) {
+            return true;
+        }
+
+        return null;
+    }
+
+    private function isOnSamePartition($file1, $file2)
+    {
+        while (!is_dir($file1)) {
+            $file1 = dirname($file1);
+        }
+        while (!is_dir($file2)) {
+            $file2 = dirname($file2);
+        }
+
+        if (realpath($file1) === realpath($file2)) {
+            return true;
+        }
+
+        // File are on the same partition if the total and free space are equals.
+        if (disk_total_space($file1) !== disk_total_space($file2)) {
+            return false;
+        }
+
+        // For free space, we have to pike multiple samples, given the value can change between 2 picks
+        $free1 = disk_free_space($file1);
+        $free2 = disk_free_space($file2);
+        $free3 = disk_free_space($file1);
+
+        return $free2 <= max($free1, $free3) && $free2 >= min($free1, $free3);
     }
 }


naderman:
The only thing I'm concerned about is not writing a backup file, preventing rollback if the directory is unwritable, shouldn't we rather fail in this case as a precaution?

jeremy-derusse:
This PR still use the `$rollbackdir` (default location in `$config->get('home')`)  to store a copy of the current version.

Do you mean that the "replacement strategy" is not atomic? and, if something wrong append during the replacement, the binary will unusable?

stof:
@jeremy-derusse ``file_put_contents`` is indeed not atomic, which might cause issues

jeremy-derusse:
I rewrite this PR to take the last comments into account.

- remove the "replace content" thing
- just keep the `is_writable` in the target directory instead of the file

avindra:
If this works (restoring the original permissions) that would be awesome. I was just looking into it myself. It would be nice to not have to run chmod 755 (or whatever permissions you use) manually after doing composer self-update.

jeremy-derusse:
@avindra see #3688 for restoring the original permissions.

src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -72,8 +72,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if (!is_writable($tmpDir)) {
             throw new FilesystemException('Composer update failed: the "'.$tmpDir.'" directory used to download the temp file could not be written');
         }
-        if (!is_writable($localFilename)) {
-            throw new FilesystemException('Composer update failed: the "'.$localFilename.'" file could not be written');
+        if (!is_writable(dirname($localFilename))) {
```
stof: this is wrong. It does not need to write here all the time (it already chose the tmpDir according based on the fact it can write to this folder or no). This is removing a feature: currently, composer can self-update in ``/usr/local/bin/composer`` for me without sudoing, because ``/usr/local/bin/composer`` is owned by my user. but writing to ``/usr/local/bin`` itself is not possible (which means that Composer will use ``$HOME/.composer/cache`` as temp dir for the download)

stof:
and the backup file is already stored in a writable location, because it uses ``$tmpDir`` to build the path.

src/Composer/Command/SelfUpdateCommand.php
jeremy-derusse: Actualy, the current implementation does NOT work.
You can not run self-update without "write privilege" on the folder containing composer. Because renaming a file, is, in fact, writing on directories.

But you make me doubt, then I just retry it.
```
$ sudo chown $USER  /usr/local/bin/composer
$ ls -al /usr/local/bin/composer
-rwxr-xr-x 1 jderusse staff 1049776 Jan 27 09:05 /usr/local/bin/composer
$ composer selfupdate
Updating to version 2fbce61f6afa7151202a92a3804e6f06d769427e.
    Downloading: 100%         


                                                                                                        
  [ErrorException]                                                                                      
  rename(/home/jderusse/.composer/cache/composer-temp.phar,/usr/local/bin/composer): Permission denied  
                                                                                                        


self-update [-r|--rollback] [--clean-backups] [--no-progress] [version]
```

I'm realy surprised that it works for you. Can you please, @stof, send me the permissions and owner on both file and directory ?
```
$ ls -ald /usr/local/bin/
drwxrwsr-x 2 root staff 4096 Jan 27 09:05 /usr/local/bin/
$ ls -al /usr/local/bin/composer
-rwxr-xr-x 1 jderusse staff 1049776 Jan 27 09:05 /usr/local/bin/composer
```

stof:
```console
$ ls -ald /usr/local/bin 
drwxr-xr-x 2 root root 4096 janv.  9 23:44 /usr/local/bin
$ ls -al /usr/local/bin/composer 
-rwxrwxr-x 1 stof stof 1048439 janv.  7 10:27 /usr/local/bin/composer
$ composer self-update
Updating to version 2fbce61f6afa7151202a92a3804e6f06d769427e.
    Downloading: 100%         
Use composer self-update --rollback to return to version c58b7d917c65692eeb00c22b2fbaaa251cb390dc
```

jeremy-derusse: Oh... I got it. That's a tricky "issue".
* `copy` need permission on the file
* `rename` need permission on the directory

See the result of this test... http://3v4l.org/E6civ

In composer we use the `rename` function, but if the source and the target directory are on different files system (that's probably your case, /home is mounted separably), the rename function will use a copy + unlink.

How can we solve that ?.
* Cheking the permission on the file will works for user with a $HOME in a separt FS and not the others
* Cheking the permission on the directory will works for user with a $HOME in the same FS and not the others

We can use `disk_total_space`, on both source and target directory, to estimate (better than nothing) if we are on the same FS or not, and then running the correct test.
What do you think @stof?

jeremy-derusse: Btw, what's about this algorithm ?
```
if ($isSameFileSystem) {
  if (is_writable($dir)) => rename
  elseif (is_writable($file)) => copy // not atomic
  else => throw error
} else {
  if (is_writable($file)) => rename // neither not atomic because use copy under the hood
  else => throw error
}
```

jeremy-derusse:
Update the PR to take into account the partition of the temp file and the original file to determine if the write permission should be controlled on the directory or the file.

Seldaek:
@jderusse is this really necessary? I haven't seen anyone complain about it and I'm a bit scared that the free space check hacks end up causing more issues than anything.

If it is then it should be rebased.

jderusse:
This PR is not absolutely necessary, given I'm able to use `sudo` as workaround. 
Thank youfor the review and sorry for the noise.

I secretly hope, that, this PR will help people to learn things about Linux permissions on `file` vs `directory` for `copy` vs `move` on `same` vs `different` filesystem partition.

Seldaek:
Well I am not against the PR but as far as I can tell it doesn't really *fix* the problem it just detects more errors before trying to do the rename right?

jderusse:
In fact this PR fix the bug by avoiding to throw an exception when you don't have write permission on the file, BUT you have on the directory (and both files are on the same partition).

But you're right, it's over complicated for a small benefit (need permission on directory but not the file, which is not the major cases)

In summary here is the different cases:
1) same partition 
write on file: yes
write on dir: no
Rename possible: no
Before the PR: fail to rename
After the PR: warned the user

2) different partition 
write on file: yes
write on dir: no
Rename possible: yes
Before the PR: ok
After the PR: ok

3) same partition 
write on file: no
write on dir: yes
Rename possible: yes
Before the PR: warned wrongly the user
After the PR: ok

4) different partition 
write on file: no
write on dir: yes
Rename possible: no
Before the PR: warned the user
After the PR: warned the user

Seldaek:
Ok I see, thanks for the detailed explanation, it's interesting for sure! But indeed maybe not worth fixing given the code to do so is a bit odd.

mikedld:
Just had this same error. My use-case is something different though: I have Windows machine where I was running `php composer.phar self-update`, but the composer.phar file itself is on a network share with OS X on the other end. Despite me being able to create and modify files on that share, self-update is failing.

I'm not sure why is this check even there, except to maybe give user a better error description. Other than that, it's useless as usually to see if you could write to a file you would just open a file and attempt writing to it. Otherwise there's a race condition between the check and actual file opening/writing, and maybe even a difference in behavior as in check fails but write succeeds (here I presume that it would work without the "is writable" check, but don't know how to build a patched composer.phar to make sure).

Seldaek:
@mikedld I removed some checks for writability so it tries before failing. Can you check if it works? You can get the latest code using self-update (or rather whatever way you updated before) and then run it again with `self-update 1.0.0-alpha11` to force it to update to another version, if that second one passes then it should be good.

mikedld:
@Seldaek, it didn't go as anticipated (note C:\test is a symlink to a network share):

```
C:\test>php composer.phar self-update 1.0.0-alpha11
Updating to version 1.0.0-alpha11.
    Downloading: 100%


  [ErrorException]
  rename(C:/Users/mike/AppData/Local/Composer/composer-temp.phar,C:\test\composer.phar): Access is denied. (code: 5)


self-update [-r|--rollback] [--clean-backups] [--no-progress] [--update-keys] [--] [<version>]
```

Seldaek:
@mikedld then.. maybe it's not writable for real?

mikedld:
It's writable for sure. As I said before, I'm able to create and modify (and remove) files manually. Maybe the fact that files are on network share has something to do with it (like, they are being locked or opened in incompatible share mode) which makes it impossible to replace the file which is currently being executed, cannot say for sure. I ended up copying everything to local drive and it works now, but if you're up to more experiments I'm at your service.

Seldaek:
Can you try to do a rename() of some random file from outside the share into the share? Maybe that's not allowed, like renaming across partitions on linux.. Maybe if we do a write instead it'd work better? Not sure what other problems would occur then, but we could try to write in case rename fails for maximum ugly workarounds :/

kdambekalns:
FYI: Ran into this today, I think… `composer.phar` is writable for the user running it, but in a folder that isn't writable for that user. Using `cp …` on the shell works, but `mv …` does not. That's what blocks selfupdate, essentially.

Seldaek:
@kdambekalns if you can repro please create a new issue for this.

PR #3592, opened 2014-12-25
Only show instead of open the repository or homepage URL

SofHad:
Small but useful option to show instead of launching the repository or homepage URL.

use:  with -s or --show option
$composer home monolog/monolog -Hs
Homepage URL: http://github.com/Seldaek/monolog

$composer home monolog/monolog --show  (or -s)
Repository URL: http://github.com/Seldaek/monolog

### DIFF
--- a/src/Composer/Command/HomeCommand.php
+++ b/src/Composer/Command/HomeCommand.php
@@ -84,7 +84,7 @@ EOT
             }
 
             if ($input->getOption('show')) {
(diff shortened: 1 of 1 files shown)


naderman:
Can we get rid of the prefix then and really only print the URL? More convenient for automated usage by other programs, and you're already specifying what you want, so you'll know what URL it is?

SofHad:
@naderman thanks for your feedback.
This is done.

PR #3593, opened 2014-12-26
[Commande licenses] Remove deprecated TableHelper and minor update of the table style

SofHad:
What do you think?
screenshot- domain date time


Seldaek:
Thanks, I took the headers but removed the borders because I don't think they add much except for taking more horizontal space.

Issue #2202, opened 2013-08-23
GitDownloader fails on composer.json update, when there is packages added as git repositories


shivas:
```
[exec] [0m[31m*** [err :: internal.xxx.com] [RuntimeException]
     [exec] [0m[31m*** [err :: internal.xxx.com] Failed to execute git remote set-url composer 'ssh://xxx@xxx.xxx.com/home/repositories/xxx.git' && git fetch composer && git fetch --tags composer
     [exec] [0m[31m*** [err :: internal.xxx.com] 
     [exec] [0m[31m*** [err :: internal.xxx.com] Error reading response length from authentication socket.
     [exec] [0m[31m*** [err :: internal.xxx.com] Permission denied, please try again.
     [exec] [0m[31m*** [err :: internal.xxx.com] Permission denied, please try again.
     [exec] [0m[31m*** [err :: internal.xxx.com] Permission denied (publickey,password).
     [exec] [0m[31m*** [err :: internal.xxx.com] fatal: The remote end hung up unexpectedly
```

Looking at code i see only one option that there is no composer remote setup yet, and he trying to set url for that...?

its in composer.phar/src/Composer/Downloader/GitDownloader.php:74

frederikbosch:
@jkobus @shivas Could you clone the code in this PR (https://github.com/composer/composer/pull/3608) and see if the bug is still there? I have investigated quite some time now in GitDownloader and helper classes. If the problem still occurs, I might be able to help fixing the problem.

EagleEyeJohn:
i'm getting this issue.  my token got revoked, after months of use. using 

```composer config -g github-oauth.github.com myoathtoken```

'worked', but when running ```composer update``` again, it still tried to use the old token and fails.  As I've provided a token, why doesn't composer prompt me to update it if it's not working?   And as it doesn't, where is the config file stored on Windows so I can change the token manually?

frederikbosch:
What happends if you remove your vendor folder and then composer update again?

EagleEyeJohn:
Now you tell me!...   I've just returned to post that that is indeed the solution to the problem.  There must be a way for Composer to deal with this automatically?  I've now got to delete vendor directories on 20+ projects.

frederikbosch:
I already submitted a pull request for this issue a while ago. It has not been merged yet. From what I understand, there are other problems that need more attention at this moment.

frederikbosch:
[...]

PR #3608, opened 2015-01-02
Remove authentication credentials from git config file

frederikbosch:
When using private repositories, composer leaves authentication credentials in .git/config. This is undesired behaviour because credentials are potentially leaked. This PR removes credentials after fetching. It contains tests for download and update.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -13,7 +13,6 @@
 namespace Composer\Downloader;
 
 use Composer\Package\PackageInterface;
-use Composer\Util\GitHub;
 use Composer\Util\Git as GitUtil;
 use Composer\Util\ProcessExecutor;
 use Composer\IO\IOInterface;
@@ -44,18 +43,28 @@ class GitDownloader extends VcsDownloader
 
         $ref = $package->getSourceReference();
         $flag = defined('PHP_WINDOWS_VERSION_MAJOR') ? '/D ' : '';
-        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
+
+        $commands = array(
+            'git init %2$s',
+            'cd ' . $flag . '%2$s',
+            'git fetch %1$s +refs/heads/*:refs/remotes/origin/* --tags'
+        );
+        $command = implode(' && ', $commands);
+
         $this->io->write("    Cloning ".$ref);
 
         $commandCallable = function ($url) use ($ref, $path, $command) {
-            return sprintf($command, ProcessExecutor::escape($url), ProcessExecutor::escape($path), ProcessExecutor::escape($ref));
+            return sprintf(
+                $command,
+                ProcessExecutor::escape($url),
+                ProcessExecutor::escape($path)
+            );
         };
-
         $this->gitUtil->runCommand($commandCallable, $url, $path, true);
-        if ($url !== $package->getSourceUrl()) {
-            $url = $package->getSourceUrl();
-            $this->process->execute(sprintf('git remote set-url origin %s', ProcessExecutor::escape($url)), $output, $path);
-        }
+
+        $url = $package->getSourceUrl();
+        $this->process->execute(sprintf('git remote add origin %s', ProcessExecutor::escape($url)), $output, $path);
+
         $this->setPushUrl($path, $url);
 
         if ($newRef = $this->updateToCommit($path, $ref, $package->getPrettyVersion(), $package->getReleaseDate())) {
@@ -79,7 +88,7 @@ class GitDownloader extends VcsDownloader
 
         $ref = $target->getSourceReference();
         $this->io->write("    Checking out ".$ref);
-        $command = 'git remote set-url composer %s && git fetch composer && git fetch --tags composer';
+        $command = 'git fetch %s +refs/heads/*:refs/remotes/origin/* --tags';
 
         $commandCallable = function ($url) use ($command) {
             return sprintf($command, ProcessExecutor::escape ($url));
@@ -227,9 +236,9 @@ class GitDownloader extends VcsDownloader
         $gitRef = $reference;
         if (!preg_match('{^[a-f0-9]{40}$}', $reference)
             && $branches
-            && preg_match('{^\s+composer/'.preg_quote($reference).'$}m', $branches)
+            && preg_match('{^\s+origin/'.preg_quote($reference).'$}m', $branches)
         ) {
-            $command = sprintf('git checkout -B %s %s && git reset --hard %2$s', ProcessExecutor::escape($branch), ProcessExecutor::escape('composer/'.$reference));
+            $command = sprintf('git checkout -B %s %s && git reset --hard %2$s', ProcessExecutor::escape($branch), ProcessExecutor::escape('origin/'.$reference));
             if (0 === $this->process->execute($command, $output, $path)) {
                 return;
             }
@@ -238,12 +247,12 @@ class GitDownloader extends VcsDownloader
         // try to checkout branch by name and then reset it so it's on the proper branch name
         if (preg_match('{^[a-f0-9]{40}$}', $reference)) {
             // add 'v' in front of the branch if it was stripped when generating the pretty name
-            if (!preg_match('{^\s+composer/'.preg_quote($branch).'$}m', $branches) && preg_match('{^\s+composer/v'.preg_quote($branch).'$}m', $branches)) {
+            if (!preg_match('{^\s+origin/'.preg_quote($branch).'$}m', $branches) && preg_match('{^\s+origin/v'.preg_quote($branch).'$}m', $branches)) {
                 $branch = 'v' . $branch;
             }
 
             $command = sprintf('git checkout %s', ProcessExecutor::escape($branch));
-            $fallbackCommand = sprintf('git checkout -B %s %s', ProcessExecutor::escape($branch), ProcessExecutor::escape('composer/'.$branch));
+            $fallbackCommand = sprintf('git checkout -B %s %s', ProcessExecutor::escape($branch), ProcessExecutor::escape('origin/'.$branch));
             if (0 === $this->process->execute($command, $output, $path)
                 || 0 === $this->process->execute($fallbackCommand, $output, $path)
             ) {
--- a/src/Composer/Util/AuthHelper.php
+++ b/src/Composer/Util/AuthHelper.php
@@ -37,7 +37,7 @@ class AuthHelper
             $store = $configSource;
         } elseif ($storeAuth === 'prompt') {
             $answer = $this->io->askAndValidate(
-                'Do you want to store credentials for '.$originUrl.' in '.$configSource->getName().' ? [Yn] ',
+                'Do you want to store credentials unencrypted for '.$originUrl.' in '.$configSource->getName().' ? [Yn] ',
                 function ($value) {
                     $input = strtolower(substr(trim($value), 0, 1));
                     if (in_array($input, array('y','n'))) {
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -44,14 +44,6 @@ class Git
             throw new \InvalidArgumentException('The source URL '.$url.' is invalid, ssh URLs should have a port number after ":".'."\n".'Use ssh://git@example.com:22/path or just git@example.com:path if you do not want to provide a password or custom port.');
         }
 
-        if (!$initialClone) {
-            // capture username/password from URL if there is one
-            $this->process->execute('git remote -v', $output, $cwd);
-            if (preg_match('{^(?:composer|origin)\s+https?://(.+):(.+)@([^/]+)}im', $output, $match)) {
-                $this->io->setAuthentication($match[3], urldecode($match[1]), urldecode($match[2]));
-            }
-        }
-
         // public github, autoswitch protocols
         if (preg_match('{^(?:https?|git)://'.self::getGitHubDomainsRegex($this->config).'/(.*)}', $url, $match)) {
             $protocols = $this->config->get('github-protocols');
--- a/tests/Composer/Test/Downloader/GitDownloaderTest.php
+++ b/tests/Composer/Test/Downloader/GitDownloaderTest.php
@@ -61,23 +61,29 @@ class GitDownloaderTest extends \PHPUnit_Framework_TestCase
             ->will($this->returnValue('dev-master'));
         $processExecutor = $this->getMock('Composer\Util\ProcessExecutor');
 
-        $expectedGitCommand = $this->winCompat("git clone --no-checkout 'https://example.com/composer/composer' 'composerPath' && cd 'composerPath' && git remote add composer 'https://example.com/composer/composer' && git fetch composer");
+        $expectedGitCommand = $this->winCompat("git init 'composerPath' && cd 'composerPath' && git fetch 'https://example.com/composer/composer' +refs/heads/*:refs/remotes/origin/* --tags");
         $processExecutor->expects($this->at(0))
             ->method('execute')
             ->with($this->equalTo($expectedGitCommand))
             ->will($this->returnValue(0));
 
+        $expectedGitCommand = $this->winCompat("git remote add origin 'https://example.com/composer/composer'");
         $processExecutor->expects($this->at(1))
             ->method('execute')
-            ->with($this->equalTo($this->winCompat("git branch -r")), $this->equalTo(null), $this->equalTo($this->winCompat('composerPath')))
+            ->with($this->equalTo($expectedGitCommand), $this->equalTo(null), $this->equalTo($this->winCompat('composerPath')))
             ->will($this->returnValue(0));
 
         $processExecutor->expects($this->at(2))
             ->method('execute')
-            ->with($this->equalTo($this->winCompat("git checkout 'master'")), $this->equalTo(null), $this->equalTo($this->winCompat('composerPath')))
+            ->with($this->equalTo($this->winCompat("git branch -r")), $this->equalTo(null), $this->equalTo($this->winCompat('composerPath')))
             ->will($this->returnValue(0));
 
         $processExecutor->expects($this->at(3))
+            ->method('execute')
+            ->with($this->equalTo($this->winCompat("git checkout 'master'")), $this->equalTo(null), $this->equalTo($this->winCompat('composerPath')))
+            ->will($this->returnValue(0));
+
+        $processExecutor->expects($this->at(4))
             ->method('execute')
             ->with($this->equalTo($this->winCompat("git reset --hard '1234567890123456789012345678901234567890'")), $this->equalTo(null), $this->equalTo($this->winCompat('composerPath')))
             ->will($this->returnValue(0));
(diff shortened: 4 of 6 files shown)


frederikbosch:
Damn you PHP 5.3. It should be OK now.

frederikbosch:
Due to PHP 5.3 compatibility ($this context is unavailable inside closures in PHP 5.3) I had to choose: either make removeCredentialsFromUrl public inside GitDownloader.php or move it to public static context. I choose the latter. If you have other ideas regarding this solution, I am willing to reimplement according to your needs.

src/Composer/Downloader/GitDownloader.php
```diff
@@ -44,18 +45,27 @@ public function doDownload(PackageInterface $package, $path, $url)
 
         $ref = $package->getSourceReference();
         $flag = defined('PHP_WINDOWS_VERSION_MAJOR') ? '/D ' : '';
-        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
+        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %4$s && git fetch %1$s';
```
stof: why are you changing the last part ? It should still fetch the ``composer`` remote, not the URL. Otherwise things might not work properly

src/Composer/Util/AuthHelper.php
```diff
@@ -20,15 +20,29 @@
...
+     */
     protected $io;
+    /**
+     * @var Config
+     */
     protected $config;
 
+    /**
+     * @param IOInterface $io
+     * @param Config $config
```
stof: this adds strictly no value as the typehints are already giving all this information

src/Composer/Util/AuthHelper.php
```diff
@@ -20,15 +20,29 @@
...
+     */
     public function __construct(IOInterface $io, Config $config)
     {
         $this->io = $io;
         $this->config = $config;
     }
 
+    /**
+     * @param $originUrl
+     * @param $storeAuth
```
stof: This adds strictly no value given that you don't define the types (and it is invalid phpdoc because the types are missing)

src/Composer/Util/AuthHelper.php
```diff
@@ -60,4 +74,28 @@ function ($value) {
...
         }
     }
+
+    /**
+     * @param $url
+     * @return string
+     */
+    public static function removeCredentialsFromUrl($url)
+    {
+        $parsed_url = parse_url($url);
```
stof: the variable name should use camelCase.

And what if PHP cannot parse the URL ?

frederikbosch:
@stof, thanks for the feedback. I will revert the docblock, fetch the composer remote and use camelcase for $parsed_url variable.

Regarding what if PHP cannot parse the url, I will throw an exception. This can be catched by GitDownloader.php so the original unmodified url will be used.

frederikbosch:
@Seldeak, could you comment on the composer remote. What is the exact purpose? Why not fetch by URL? Is it needed elsewhere?

frederikbosch:
@Seldaek, could you comment on the composer remote. What is the exact purpose? Why not fetch by URL? Is it needed elsewhere?

frederikbosch:
@stof Could you elaborate why fetching by URL could break things? The git-fetch documentation does not indicate that there are issues with fetching by URL.

Let me clarify why I want to fetch by URL. I do not want want to add a remote with user credentials and remove credentials afterwards. When the clone or fetch command gets interupted (e.g. ctrl-c), the credentials are left in the remote url and composer is not able to restore this. That's why I would prefer fetching by url. Then nothing is stored at all.

While typing this message I also realize that cloning should be replaced by init, add remote without credentials and fetch with credentials.

stof:
@frederikbosch does fetching by URL fetches all branches of the remote repo or only the main branch ? If it fetches only the main branch only, it will break lots of things.
And the other issue is when people are referencing branches rather than commits in custom package definitions, because I don't think Composer will be able to checkout the branches if you fetched an URL rather than a remote (because it will not have references for all remote branches)

src/Composer/Util/AuthHelper.php
```diff
@@ -60,4 +74,28 @@ function ($value) {
...
+        if (isset($parsed_url['user'])) {
+            unset($parsed_url['user']);
```
GromNaN: You can unset safely even if the key is not set.

frederikbosch:
Thanks for the comments. I am rewriting the PR.

@stof You were right that fetching by URL does fetch the main branch only. However, the following command ```git fetch %1$s +refs/heads/*:refs/remotes/composer/* --no-tags``` fixes this.

Seldaek:
@frederikbosch the main reason to have the composer remote right now is to fetch everything easily and to store the credentials. Perhaps we could just do like in RemoteFilesystem now and ask the user if they want to store the credentials or not, and possibly take your solution of removing creds from the url and storing them in auth.json instead so it's all in one place. Anyway mostly historical reasons but I think whatever we do there needs to be a way to store credentials and not prompt all the time for people who don't use ssh keys.

frederikbosch:
@Seldaek Thanks for this feedback. I understand your mentioned concerns and will make sure to comply to them in the revision I am working on.

frederikbosch:
@Seldaek @stof Could you have a look at the new revision? Instead of fetching by remote name, I have chosen to fetch by URL. This makes sure authentication credentials never get into .git/config. So if the user aborts ```composer install``` or ```composer update```, the config file remains clean.

I tried to make sure that the commands do exactly the same as before and only mitigate the credentials issue. @stof thank again for pointing out that my initial fetch was not correct. Do you both think it is correct now? I am also not cloning the source anymore. It is now ```git init`` and ```git fetch```. The reason for this is that if there are credentials in the ```git clone``` command, then these credentials ended up in the origin remote config.

If you feel this is the right thing, I will make sure to repair my credential tests (both are skipped now) and moreover, looking at them again, I will improve both tests.

@Seldaek, I believe that both your concerns at met with this commit. The user is still able to store the credentials in auth.json. In the matter of fact: auth.json should now be the only place that contains the credentials. This logic is taken care of in Util/Git.php (lines 133-143) and I didn't touch that part.

src/Composer/Downloader/GitDownloader.php
```diff
@@ -44,18 +43,33 @@ public function doDownload(PackageInterface $package, $path, $url)
 
         $ref = $package->getSourceReference();
         $flag = defined('PHP_WINDOWS_VERSION_MAJOR') ? '/D ' : '';
-        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
+
+        $commands = array(
+            'git init %2$s',
+            'cd ' . $flag . '%2$s',
+            'git fetch %1$s +refs/heads/*:refs/remotes/composer/* --no-tags'
```
stof: we do want to fetch tags

frederikbosch: @stof Removing --no-tags would mean that the result of the command would change compared to the original command. The original command was ```git fetch composer``` which does not fetch tags. If you have a look at the original ```doUpdate``` method: there the tags are explicitly requested ```git fetch --tags```. And that is what I still do in the PR code.

After your comment, my intent was to replicate the result of the original commands without getting the credentials in the .git/config. But I also think that tags are to be fetched when downloading. So I will remove the --no-tags option.

frederikbosch:
@stof, @Seldaek I think all required tests are included. Also tested AuthHelper. The confirmation message when saving now includes the word *unencrypted*. If you do not like that, you/I can remove it.

frederikbosch:
Squashed the commits, hope this makes it easier to review.

stof:
@frederikbosch when squashing the commits, you need to force the push to github instead of merging with the old history of the remote branch. Otherwise it makes the history *more* complex than before the squash. Can you clean your branch again ?

frederikbosch:
@stof Thanks again. Hopefully it is alright now.

src/Composer/Util/Git.php
```diff
@@ -44,14 +44,6 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
             throw new \InvalidArgumentException('The source URL '.$url.' is invalid, ssh URLs should have a port number after ":".'."\n".'Use ssh://git@example.com:22/path or just git@example.com:path if you do not want to provide a password or custom port.');
         }
 
-        if (!$initialClone) {
```
alcohol: Any particular reason you want to ignore existing credentials?

src/Composer/Downloader/GitDownloader.php
```diff
@@ -44,18 +43,28 @@ public function doDownload(PackageInterface $package, $path, $url)
...
         };
-
         $this->gitUtil->runCommand($commandCallable, $url, $path, true);
-        if ($url !== $package->getSourceUrl()) {
-            $url = $package->getSourceUrl();
-            $this->process->execute(sprintf('git remote set-url origin %s', ProcessExecutor::escape($url)), $output, $path);
-        }
+
+        $url = $package->getSourceUrl();
+        $this->process->execute(sprintf('git remote add origin %s', ProcessExecutor::escape($url)), $output, $path);
```
alcohol: Considered using the `--fetch` and `--tags` options for `remote add`? Could make the next command redundant.

src/Composer/Util/Git.php
frederikbosch: They are already in auth.json. If you decided not to store them, you are prompted to enter the username and password again. And you can choose again to store them in auth.json.

To be really clear about this: there should be one place only for the credentials. That is auth.json, not the git remote.

alcohol: While I am all in favor of what you are trying to achieve, I think this is a change that could break existing automated flows. Removing this part of the code seems excessive and unnecessary.

src/Composer/Downloader/GitDownloader.php
frederikbosch: You do not want to to run ```composer remote add```. This is because the command is rewritten when the command requires authentication, and happends (here)[https://github.com/frederikbosch/composer/blob/master/src/Composer/Util/Git.php#L126] in combination with line closure that is passed in (here)[https://github.com/frederikbosch/composer/blob/master/src/Composer/Downloader/GitDownloader.php#L57].

So if you use ```composer remote add``` then authentication credentials end up in the git remote. And you do not want that. You do not want it ever to get there. What if composer install fails? Or the user aborts the command? Then credentails are there.

alcohol: But... this is your change?

src/Composer/Util/Git.php
frederikbosch: It could break thing. The scenario where it breaks is when someone decided not to store in auth.json but is still counting his ```composer update``` to execute without prompt.

However, do you think this user is even aware of the fact that his/her credentails are in .git/config? Did he not decided *not* to store the credentials in first place?

alcohol: I don't think this is a decision **you** can make for the **user** retroactively. In my opinion, it makes perfect sense doing this for any future setups. But I strongly discourage changing the behaviour of existing setups. There is no **need** for that.

frederikbosch: This method could be deprecated. Together with a warning, we could create a timeframe that the user has to upgrade. Yet again, composer is not 1.0 yet. It could also be a very good timing to drop something that causes user credentials leaks.

frederikbosch: I agree with your comment. It is not my decision. I trust in the wisdom of @Seldaek to take a wise decision in this matter. Hopefully our conversation helps him with that.

src/Composer/Downloader/GitDownloader.php
frederikbosch: Yes. The first command is passed to ```$this->gitUtil```. The util claas rewrites commands.

```
$this->gitUtil->runCommand($commandCallable, $url, $path, true);
```

Where as the ```git remote add``` is passed to the process executor, which executes directly.

```
$this->process->execute(sprintf('git remote add origin %s', ProcessExecutor::escape($url)), $output, $path);
```
For your info: this was the setup of the original code anyway. See (lines 54-58 in the current GitDownloader)[https://github.com/composer/composer/blob/master/src/Composer/Downloader/GitDownloader.php#L54].

src/Composer/Util/Git.php
alcohol: Well, I've shared my 2 cents regarding the topic. I personally don't see a requirement for doing this.

Yes, there is a risk involved. 
But no, it does not outweigh breaking backwards compatibility in my humble opinion.

Also, it doesn't **cause** credential leaks. It merely exposes the possibility. Quite a big difference.

src/Composer/Downloader/GitDownloader.php
alcohol: Ok then I don't see how your response is relevant to my suggestion. If it is still going to run the `git remote add`, then what is wrong with considering using the flags I suggested?

frederikbosch: After reading again, I misunderstood your suggestion. Sorry for that. The only thing I tried to do is to replicate the original behaviour, but without user credentails in .git/config. The original command was ```git clone --no-checkout``` followed by ```git remote add composer``` and finally ```git fetch```. The commands in this PR should replicate that.

I should look into your suggestion. If that gets this PR merged, I have no problem looking into that. It was just out my scope before.

alcohol:
Any reason you dropped the composer remote name almost everywhere by the way? If you're going to go down that path, I think you need to consistently change it everywhere in that case.

See [src/Composer/Downloader/GitDownloader.php](https://github.com/composer/composer/blob/master/src/Composer/Downloader/GitDownloader.php) for example

frederikbosch:
Because @Seldaek confirmed it exists mostly for historical reasons (see his comment above): to fetch things easily and save credentials. These are not valid anymore with this PR, so I dropped it. And I did change it consistently. Please see [GitDownloader.php in the PR](https://github.com/frederikbosch/composer/blob/b3c9dd1e05e8e6a5f750446d4239bbad0a093252/src/Composer/Downloader/GitDownloader.php).

If you see a link to the composer remote anyhow, please let me know, and I will remove it.

alcohol:
I'm sorry, I completely overlooked this. My head is really not in it this morning it seems.

frederikbosch:
Hehe, no problem, this is not a PR that you merge in a second. Maybe @Seldaek could confirm whether he likes the proposed solutions. If not, I am happy to work on enhancements. I also understand that he has got other things on his mind!

EagleEyeJohn:
@frederikbosch after your help today, it only seems fair to add that this PR would help with the situation I had today. 
My github token was revoked and my new one provided to and stored by Composer. This left me unable to ```composer update``` on any project as it was still trying to use the old revoked token.  The only (simple) solution was to delete /vendor directory in ~25 projects.  We use hooks within code included via Composer and as the methods are not callable until the subsequent update, I had to run composer update twice on each project. This will no doubt take several days to complete.

frederikbosch:
@EagleEyeJohn Could you actually confirm that this PR fixes issue #2202 ? Did you try replicate the issue with the composer binary from the PR?

masi:
I'm a bit lost in all these comments, so I only make my voide heard for the "no credentials in .git/config". At least not when I ask composer not to store my cerdentials.

I am quite unhappy with the current behaviour as I have now to go to all shared servers an remove my credentials manually - after each "composer install".

frederikbosch:
The following list represents the issued bugs that seem to be related to credentials in .git/config

1. User is not able to modify username and password, one has to delete vendor folder
2. When github gets revoked, one has to remove vendor folder completely before composer install/update works
3. Credentials are stored .git/config even when the requested not to store credentials
4. Spreading around of login credentials
5. Composer require asks for authentication whereas composer update does not

frederikbosch:
@Seldaek You added a label to this bug 8 days ago, and just now there was also someone in #3644 that was hit by the bug. Would it be helpful if I update this PR to current code? I would love to do that, if it gets merged afterwards.

alcohol:
If you can rebase it and bring it up to date, I can give it another thorough review. @Seldaek is currently travelling for the next two weeks or so at the very least, and is only online sporadically. So I cannot guarantee it will be merged asap, but it will definitely help :-)

frederikbosch:
@alcohol Sounds great. If I have some extra time during next few days I will do that.

curry684:
I've tried to rebase this PR but it's not really doable for an 'outsider' anymore with all other changes in the Git layers since. Would be great if @frederikbosch had a chance.

frederikbosch:
@curry684 Thanks for the reminder. I think about it sometimes. But I also see a gigantic mountain to climb over. The PR took quite an amount of time then and I guess it will take the same amount again.

Now I am still willing to make a commitment. But I would like to hear from @seldaek that if I update the PR to current master, that this PR is a priority. Of course I understand that there will be reviewing phase. But if this PR does not have priority, I am not going spend my time on it (now).

@seldaek Could you tell me how much priority this PR will have if it gets updated? Would you like to see it going into 1.0?

Seldaek:
Yeah I am really sorry for dropping the ball on this one, I know you 
spent time on it and it sucks that it's conflicting now.

I can't promise it'll get in 1.0, might prefer to merge it right after 
for safety as it touches sensitive code, but yes I think it can be 
pretty high on the list at this point.

curry684:
I was picking it up because it's one of the last 2 items in the "needed before alpha" list and we're currently in beta hehe.

frederikbosch:
@seldaek No need to be sorry. Fully understand it. I will do my best to update soon! Thanks for your quick reply!

alcohol:
@frederikbosch any chance you could update this? Or otherwise enable us to modify it/submit changes to it?

frederikbosch:
@alcohol I will have a look at this really soon.

alcohol:
Been trying to rebase this, but it is quite a nightmare now that we also have clones in the vcs cache to make use of the dissociate and reference flags.

frederikbosch:
@alcohol I will not try to rebase this. I am simply going to start over and copy parts from the initial PR when possible. Rebasing would be quite hard I guess after all these changes.

alcohol:
I was also wondering, since this removes credentials, anyone that was used to having a workflow whereby they simply modified packages in their vendor directory and committed + pushed changes from there, said flow will no longer work, right?

frederikbosch:
@alcohol Sometimes, I use this workflow. It would still work afterwards. You only have to type your username/password when pushing.

thammana:
can someone merge this as it is one of our blocker due to security reasons, waiting since one month for this fix. help is appreciated :-)

phixom:
Why isn't the request merged? This framework isn't useable in some companies for secure compliance issues. This fix could throwing some of them away.

alcohol:
Various reasons probably. Out of sync, contains conflicts that first need to be resolved. Low priority. Etc.

Also most security compliance issues is just people not understanding security or being too lazy to clean up their production build artifacts themselves. Most concerns raised in this thread can be solved in other ways as well, hence the low priority.

rgagnon24:
Just adding that we have the same problem at our company as well.  After composer prompts for the username and password to our private enterprise git repository, it same the credentials in the URL of the file:

It asks this question:
```Do you want to store credentials for git.viasat.com in ~/.config/composer/auth.json ? [Yn] ```

to which we reply "n", but the credentials are stored in a completely different file:
```~/.cache/composer/vcs/https---git.companyname.com-repository-name.git/config```
inside the section labeled   ```[remote "origin"]```, value for "```url = ```"

This results is a problem trying to update from composer when the user's password changes on the system, but the file has the old credentials.  It is also a problem for leaking security credentials.

alcohol:
@rgagnon24 I believe that file is managed by git though, not by Composer.

rgagnon24:
> 
> 
> @rgagnon24 I believe that file is managed by git though, not by Composer.

Wouldn't all files under ~/.cache/composer be "composer" files, and "~/.cache/git" for git?

alcohol:
No, `https---git.companyname.com-repository-name.git` is the result of a `git clone` using `--no-checkout` and possibly also `--dissociate` depending on your git version.

PR #3610, opened 2015-01-04
Add classmap-authoritative config setting

### LINKED ISSUE #3603: Optimized autoloader is slow (compared to MediaWiki's autoloader)
Hi,

We've recently started using composer on Wikimedia sites, and have noticed that the optimized composer autoloader is significantly slower than [MediaWiki's autoloader](https://github.com/wikimedia/mediawiki/blob/master/includes/AutoLoader.php#L38).

@tstarling observed that composer takes about 63µs per class, and MediaWiki takes about 7µs per class. This was causing an overhead of about 13% of CPU time on short requests like API calls.

For now we've set prepend-autoloader: false, since most of our classes are in MediaWiki's autoloader, and are more likely to be found. However this will continue to cause issues as we start migrating more and more code to use composer's autoloader.

Some data:

On my local MediaWiki install running in a Ubuntu VM:
* Using prepend-autoloader: true: http://p.defau.lt/?WLSEpunDkWl_5BmCaW5KxQ
* Using prepend-autoloader: false: http://p.defau.lt/?924vZrq5d1XDMiQskeOGwg

On Wikimedia production:
* https://performance.wikimedia.org/xenon/svgs/daily/2014-12-31.svgz

bd808:
Add a "classmap-authoritative" configuration setting that can be used to
disable searching the various prefix and fallback directories for
classes that have not been registered with the
Composer\Autoload\ClassLoader class map. This setting can be used to
optimize performance by avoiding a potentially large number of
`file_exists` calls when Composer is being used in a program with
additional autoloader facilities. Use of the setting implies
"optimize-autoloader" to ensure that the most complete class map
possible is generated.

Closes #3603

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -54,7 +54,7 @@ class ClassLoader
     private $useIncludePath = false;
     private $classMap = array();
 
-    private $classMapAuthoratative = false;
+    private $classMapAuthoritative = false;
 
     public function getPrefixes()
     {
@@ -254,11 +254,11 @@ class ClassLoader
      * Turns off searching the prefix and fallback directories for classes
      * that have not been registered with the class map.
      *
-     * @param bool $classMapAuthoratative
+     * @param bool $classMapAuthoritative
      */
-    public function setClassMapAuthoritative($classMapAuthoratative)
+    public function setClassMapAuthoritative($classMapAuthoritative)
     {
-        $this->classMapAuthoratative = $classMapAuthoratative;
+        $this->classMapAuthoritative = $classMapAuthoritative;
     }
 
     /**
@@ -266,9 +266,9 @@ class ClassLoader
      *
      * @return bool
      */
-    public function getClassMapAuthoratative()
+    public function isClassMapAuthoritative()
     {
-        return $this->classMapAuthoratative;
+        return $this->classMapAuthoritative;
     }
 
     /**
@@ -321,7 +321,8 @@ class ClassLoader
         // class map lookup
         if (isset($this->classMap[$class])) {
             return $this->classMap[$class];
-        } elseif ($this->classMapAuthoratative) {
+        }
+        if ($this->classMapAuthoritative) {
             return false;
         }
 
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -54,6 +54,8 @@ class ClassLoader
     private $useIncludePath = false;
     private $classMap = array();
 
+    private $classMapAuthoratative = false;
+
     public function getPrefixes()
     {
         if (!empty($this->prefixesPsr0)) {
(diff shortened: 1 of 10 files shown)


bd808:
It's probably worth discussing whether or not this should also have a related cli argument for `install`, `update` and `dump-autoload`.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -249,6 +251,27 @@ public function getUseIncludePath()
...
+    {
+        $this->classMapAuthoratative = $classMapAuthoratative;
+    }
+
+    /**
+     * Should class lookup fail if not found in the current class map?
+     *
+     * @return bool
+     */
+    public function getClassMapAuthoratative()
```
stof: ``is*`` would be better than ``get*``

src/Composer/Autoload/ClassLoader.php
```diff
@@ -298,6 +321,8 @@ public function findFile($class)
         // class map lookup
         if (isset($this->classMap[$class])) {
             return $this->classMap[$class];
+        } elseif ($this->classMapAuthoratative) {
```
stof: just use ``if`` rather than ``elseif`` as the ``if`` returns.

tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
```diff
@@ -483,6 +499,49 @@ public function testClassMapAutoloadingEmptyDirAndExactFile()
...
+                'ClassMapBar' => $this->vendorDir.'/b/b/test.php',
+                'ClassMapBaz' => $this->vendorDir.'/c/c/foo/test.php',
+                'ClassMapFoo' => $this->vendorDir.'/a/a/src/a.php',
+            ),
+            include $this->vendorDir.'/composer/autoload_classmap.php'
+        );
+        $this->assertAutoloadFiles('classmap5', $this->vendorDir.'/composer', 'classmap');
+
+        $this->assertRegExp('/\$loader->setClassMapAuthoritative\(true\);/', file_get_contents($this->vendorDir.'/composer/autoload_real.php'));
+        // FIXME: how can we actually test the ClassLoader implementation?
```
stof: not in this test at all

tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
```diff
@@ -483,6 +499,49 @@ public function testClassMapAutoloadingEmptyDirAndExactFile()
...
+            array(
+                'ClassMapBar' => $this->vendorDir.'/b/b/test.php',
+                'ClassMapBaz' => $this->vendorDir.'/c/c/foo/test.php',
+                'ClassMapFoo' => $this->vendorDir.'/a/a/src/a.php',
+            ),
+            include $this->vendorDir.'/composer/autoload_classmap.php'
+        );
+        $this->assertAutoloadFiles('classmap5', $this->vendorDir.'/composer', 'classmap');
+
+        $this->assertRegExp('/\$loader->setClassMapAuthoritative\(true\);/', file_get_contents($this->vendorDir.'/composer/autoload_real.php'));
```
stof: you could use ``assertContains`` for a better report on failure and a more readable assertion (no need for escaping regex special chars)

src/Composer/Autoload/ClassLoader.php
bd808: done

src/Composer/Autoload/ClassLoader.php
bd808: done

tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
bd808: done

tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
bd808: Removed the FIXME note. The desire to test this still remains, but I can't think of how it can legitimately be done as a ClassLoader class will almost certainly already be loaded to get to the tests themselves.

legoktm:
Bump?

legoktm:
Thank you!

hkdobrev:
> It's probably worth discussing whether or not this should also have a related cli argument for install, update and dump-autoload.

I think CLI interface for this would be useful for people who are not using this during development, but are doing `composer install --no-dev --optimize-autoloader` on production. Which I believe is quite common.

PR #3614, opened 2015-01-05
force symfony/console 2.3.* on composer.json

endel:
For some reason, `composer` was installing `symfony/console` version 2.7 in my project. 

I've changed my project's `composer.json` to use `"symfony/console": "2.3.*"` and now it's installing the correct version.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php": ">=5.3.2",
         "justinrainbow/json-schema": "~1.3",
         "seld/jsonlint": "~1.0",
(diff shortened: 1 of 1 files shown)


stof:
This is wrong. Composer is perfectly valid with newer stable versions of the Console component (and the lock file gets Console 2.6).

the reason why it gets 2.7 for you is probably because you changed your minimum stability to allow dev versions (and Symfony 2.7-dev is not ready to be used yet)

endel:
You're right. Thanks for the explanation.

PR #3620, opened 2015-01-06
Update 05-repositories.md

kiak:
it's require to use url that ending with .git to make things works


Seldaek:
Not really no, with bitbucket it is for example but not with github. You just have to use a working git repo url.

PR #3627, opened 2015-01-07
require ext-zip when developing

greg0ire:
It is necessary for the test suite to pass.


stof:
I would rather skip the tests requiring the zip extension when it is not there rather than forbidding people to contribute if they don't have it

greg0ire:
@stof : but it isn't that hard to install, is it ? Anyway, if you really think it could stop some people from contributing, here is #3631 .

PR #3631, opened 2015-01-08
skip some tests when the zip extension misses

greg0ire:
It is required for some tests to run.

### DIFF
--- a/tests/Composer/Test/Repository/ArtifactRepositoryTest.php
+++ b/tests/Composer/Test/Repository/ArtifactRepositoryTest.php
@@ -19,6 +19,14 @@ use Composer\Package\BasePackage;
 
 class ArtifactRepositoryTest extends TestCase
 {
+    public function setUp()
+    {
+        parent::setUp();
+        if (!extension_loaded('zip')) {
+            $this->markTestSkipped('You need the zip extension to run this test.');
+        }
+    }
+
     public function testExtractsConfigsFromZipArchives()
     {
         $expectedPackages = array(


greg0ire:
The referencing PR is an alternate solution.

tests/Composer/Test/Repository/ArtifactRepositoryTest.php
```diff
@@ -33,7 +33,11 @@ public function testExtractsConfigsFromZipArchives()
         );
 
         $coordinates = array('type' => 'artifact', 'url' => __DIR__ . '/Fixtures/artifacts');
-        $repo = new ArtifactRepository($coordinates, new NullIO(), new Config());
+        try {
+            $repo = new ArtifactRepository($coordinates, new NullIO(), new Config());
+        } catch (\RuntimeException $e) {
```
stof: instead of catching all RuntimeException (which can hide other bugs and does not make the test readable), I would prefer doing an ``extension_loaded`` check at the beginning of the test (which will make it clear, and it seems understandable that testing the zip extraction requires the zip extension). Your current code does not let uss understand why this catching is there

greg0ire: Right. It makes the potentially wrong assumption that `ArtifactRepository::__construct()` will always contain one `RuntimeException` . I'll fix that.

greg0ire: @stof : fixed. And it is more DRY, because I used `setUp()`, which makes sense to me since there won't be any test without an `ArtifactRepository` instanciation.

tests/Composer/Test/Repository/ArtifactRepositoryTest.php
```diff
@@ -19,6 +19,13 @@
 
 class ArtifactRepositoryTest extends TestCase
 {
+    public function setUp()
+    {
+        if (!extension_loaded('zip')) {
+            $this->markTestSkipped('You need the zip extension to run this test.');
+        }
```
stof: please call the parent method to be sure we don't get weird things if we had some logic in ``Composer\TestCase::setUp`` in the future

greg0ire: done.

PR #2085, opened 2013-07-15
`composer status` improvement for unpushed changes

asgrim:
`composer status` now prevents overwriting Git package if changes have not been pushed. This change, if accepted, would resolve #2046.

I'm happy to make further modifications to this - please be nice, it's my first Composer PR :)

Issue #2553, opened 2014-01-06
Call to undefined method Composer\Autoload\ClassLoader::setPsr4()


Anahkiasen:
I was trying to do a test PSR4 package for myself, I have the latest version of Composer. However after running `composer install` with the following autoload directive :

``` json
"autoload": {
    "psr-4": {
        "Warkham\\": "src"
    }
},
```

When including `vendor/autoload.php` I'm getting the following error :

```
PHP Fatal error:  Call to undefined method Composer\Autoload\ClassLoader::setPsr4() in /Users/anahkiasen/Sites/autopergamene/warkham/vendor/composer/autoload_real.php on line 36
PHP Stack trace:
PHP   1. {main}() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/composer/bin/phpunit:0
PHP   2. PHPUnit_TextUI_Command::main() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/composer/bin/phpunit:63
PHP   3. PHPUnit_TextUI_Command->run() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/PHPUnit/TextUI/Command.php:126
PHP   4. PHPUnit_TextUI_Command->handleArguments() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/PHPUnit/TextUI/Command.php:135
PHP   5. PHPUnit_TextUI_Command->handleBootstrap() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/PHPUnit/TextUI/Command.php:596
PHP   6. PHPUnit_Util_Fileloader::checkAndLoad() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/PHPUnit/TextUI/Command.php:756
PHP   7. PHPUnit_Util_Fileloader::load() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/PHPUnit/Util/Fileloader.php:77
PHP   8. include_once() /Users/anahkiasen/.composer/vendor/phpunit/phpunit/PHPUnit/Util/Fileloader.php:93
PHP   9. require() /Users/anahkiasen/Sites/autopergamene/warkham/tests/WarkhamTestCase.php:4
PHP  10. ComposerAutoloaderInit1715e0360e80b7f7434edb0d52b414a4::getLoader() /Users/anahkiasen/Sites/autopergamene/warkham/vendor/autoload.php:7
```

Tried doing `composer self-update` but nothing :

```
[...]

Issue #2253, opened 2013-09-11
Prevent Git update if head has diverged from remote


Burgov:
My colleague lost some work today (not too much luckily), after committing, but not pushing changes to a vendor library. To reproduce it, follow these steps:

1. have two copies of the vendor library in different projects
2. in copy A do a commit and push it
3. in copy B do a commit, but do not push it
4. in project B do a composer update

composer then executes a reset --hard in the library directory, effectively removing the commit from the face of the earth.

Suggested fix: just like when there are pending changes found in the library, and composer asks if you want to squash, reset or whatever, I think composer should detect diverged branches and ask you what to do with it (reset, rebase, etc.)

Seldaek:
Closing as #3633 is merged.

PR #3633, opened 2015-01-08
Improve status for unpushed branches

greg0ire:
This is a cleaned up, rebased, resurrection of #2085

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -22,7 +22,7 @@ use Composer\Config;
 /**
  * @author Jordi Boggiano <j.boggiano@seld.be>
  */
-class GitDownloader extends VcsDownloader
+class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
 {
     private $hasStashedChanges = false;
     private $gitUtil;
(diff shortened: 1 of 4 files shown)


src/Composer/Downloader/GitDownloader.php
```diff
@@ -120,6 +143,13 @@ protected function cleanChanges(PackageInterface $package, $path, $update)
     {
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
+
+        if ($this instanceof DvcsDownloaderInterface) {
```
stof: this makes no sense. It will always be true

greg0ire: @stof: indeed!
@asgrim : any rationale behind this ? Can I safely remove it ?

asgrim: No idea, I guess so! :)

greg0ire: @stof : fixed

naderman:
What's the upside of this over https://github.com/composer/composer/issues/2253#issuecomment-39662835 ?

greg0ire:
From the description of #2553 this does not only work for branches that have diverged, but for changes branches that are behind their remote counterpart (but maybe the description is bad). Plus, it gives a better feedback with what exactly is wrong with status codes : there are different status codes for different situations (changes commited but not pushed, changes not commited, (changes commited and not pushed in one vendor && changes not commited in another)).

greg0ire:
Also, it is more generic : there is an interface that is introduced so that the same functionality can be implemented for mercurial, bazaar and your favorite DCVS.

greg0ire:
I just rebased, the tests are green.

greg0ire:
ping @naderman are there any plans to merge this (or #2553) ?

Seldaek:
Thanks @greg0ire, sorry for the delay. I had to do quite a few tweaks though (see https://github.com/composer/composer/commit/be4d385942cce9b0d1116aba985a99380c7a53d1) to avoid false positives for me.. Hopefully it'll go well in the wild and I didn't miss anything.

greg0ire:
Better late than never I guess :) Thanks for merging that!

Issue #3282, opened 2014-09-18
Increase performance of the remote file system


francoispluchino:
It would be interesting to improve the performance of the remote file system. Currently only the method `file_get_contents` is used, but using `CURL`, we could significantly improve the speed of Composer, especially for VCS Repository that uses a lot the remote file system.

I know that `CURL` is not present by default, but we could do a test to detect its presence, and use `file_get_contents` when it is not active.

Of course, I've seen pass the issues about using `CURL`, but I would like to make a comparative test, showing that it would be really cool to add support when it is available.

Indeed, I did performance tests with `file_get_contents` and `CURL`, by enabling and disabling URL redirection, while performing 10 tests for each of the 4 types (representing 40 tests per attempt).

**Types:**
- `file_get_contents` with follow redirect
- `curl` with follow redirect
- `file_get_contents` without follow redirect
- `curl` without follow redirect

**Results of 10 attempts with PHP 5.5 on Windows 8:**

min increase | med increase | max increase | **report** (number of time faster)
-------------|--------------|--------------|-----------------
-20%         | -310%        | -513%        | 6.1x
-22%         | -270%        | -469%        | 5.7x
-21%         | -278%        | -460%        | 5.6x
-23%         | -273%        | -475%        | 5.7x
-24%         | -287%        | -486%        | 5.9x
-20%         | -272%        | -463%        | 5.6x
-21%         | -272%        | -470%        | 5.7x
-22%         | -270%        | -465%        | 5.7x
-22%         | -279%        | -469%        | 5.7x
-22%         | -280%        | -459%        | 5.6x
**avg**      | **avg**      | **avg**      | **avg**
-22%         | -279%        | -473%        | **5.7x**

Legend:
- `min increase`: if using CURL with follow redirect
- `med increase`: if using file_get_contens without follow redirect
- `max increase`: if using CURL without follow redirect
- `report`: number of time faster

Given that the vast difference in performance is when the redirection is not enabled, maybe, it would be interesting to perform redirection manually.

Furthermore, the redirects are rarely used for API, therefore, we could significantly improve the performance of Composer using `CURL` when it is available and managing the redirects manually.

**To conclude:**

In the worst case we increase the performance of 22% and at best 473% (that is to say up to 5.7 times faster), and that is no small thing, I think.

**The script to do the tests:**

```php
<?php

class TestPerf
{
    private $followUrl;
    private $useCurl;
    private $url = 'http://www.google.com';
    private $max = 10;
    private $values = array();

[...]

PR #3636, opened 2015-01-09
cUrl support for files download

a-goryachev:
Added CurlDriver class to make http(s) requests using cUrl library. All current logic of RemoteFilesystem was implemented in CurlDriver:
* Authentication support (github.com, private repos, etc)
* Proxy support
* Download progress indicator
* User-agent header

For https requests, curl option SSL_VERIFY_PEER setted to false. This is not good, but it neсessary for connections to local toran proxy with selfsigned certificate. We can add composer option (for example, --curl-verify-ssl) to manage verify peer behavior.

Current logic of RemoteFilesystem moved to StreamDriver. RemoteFilesystem now only invokes driver's method get() to getContents of requested file. For http(s) connections RemoteFilesystem uses CurlDriver (if curl extension loaded). If curl not loaded, StreamDriver used. StreamDriver also used for local files.

Related issues:
https://github.com/composer/composer/issues/704
https://github.com/composer/composer/issues/3282


a-goryachev:
My solution breaks BC and RemoteFilesystem tests fails. I will develop another solution for cUrl support.

PR #3639, opened 2015-01-09
Interactive interface with autocompletion that can help to select the packages to update

SofHad:
Here is an illustration
en2

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -80,11 +80,10 @@ EOT
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $composer = $this->getComposer(true, $input->getOption('no-plugins'));
-        $packages = $input->getOption('interactive') ? $this->getPackagesInteractively(
-            $input,
-            $output,
-            $composer
-        ) : $input->getArgument('packages');
+        $packages = $input->getArgument('packages');
+        if ($input->getOption('interactive')) {
+            $packages = $this->getPackagesInteractively($input, $output, $composer, $packages);
+        }
 
         if ($input->getOption('no-custom-installers')) {
             $output->writeln('<warning>You are using the deprecated option "no-custom-installers". Use "no-plugins" instead.</warning>');
@@ -146,20 +145,10 @@ EOT
         return $install->run();
     }
 
-    private function getPackagesInteractively(
-        InputInterface $input,
-        OutputInterface $output,
-        Composer $composer
-    ) {
-        if ($input->getArgument('packages')) {
-            throw new \InvalidArgumentException(
-                'The option --interactive does not require an argument'
-            );
-        }
-
+    private function getPackagesInteractively(InputInterface $input, OutputInterface $output, Composer $composer, $packages)
+    {
         $packagesMap = $composer->getRepositoryManager()
             ->getLocalRepository()->getPackages();
-        $config      = $composer->getConfig();
 
         $requiredPackageNames = array();
         foreach (
@@ -174,18 +163,10 @@ EOT
         foreach ($packagesMap as $package) {
             $InstalledPackageNames[] = $package->getPrettyName();
         }
-        $names = array_unique(
+
+        $autocompleterValues = array_unique(
             array_merge($InstalledPackageNames, $requiredPackageNames)
         );
-
-        $vendorDir      = $config->get('vendor-dir', true);
-        $vendorWildcard = sprintf('%s/*', $vendorDir);
-
-        $autocompleterValues = array($vendorWildcard);
-        foreach ($names as $name) {
-            $autocompleterValues[] = $name;
-            $autocompleterValues[] = sprintf('%s/%s', $vendorDir, $name);
-        }
 
         $helper       = $this->getHelper('question');
         $question     = new Question(
@@ -193,7 +174,7 @@ EOT
             null
         );
 
-        $packages = array();
+        $packages = is_array($packages) ? $packages : array();
         $output->writeln('<info>NB: Empty package ends submission.</info>'); // I couldn't find any better for now!
 
         $continue = true;
@@ -204,11 +185,7 @@ EOT
             if (is_string($addedPackage)) {
                 if (!in_array($addedPackage, $packages)) {
                     $packages[] = $addedPackage;
-                    if ($addedPackage === $vendorWildcard) {
-                        $autocompleterValues = array();
-                    } else {
-                        $autocompleterValues = array_diff($autocompleterValues, array($addedPackage));
-                    }
+                    $autocompleterValues = array_diff($autocompleterValues, array($addedPackage));
                 } else {
                     $output->writeln(sprintf('<error>The package "%s" was already added.</error>', $package));
                 }
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Command;
 
+use Composer\Composer;
 use Composer\Installer;
 use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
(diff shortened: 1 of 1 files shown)


hkdobrev:
@SofHad 

1. You can assume by default the answer to "Add more packages" is "yes" and then just have a sentence on top which says "Empty package ends submission" (or something better). When the user submits empty package name, Composer would assume all packages are entered.

2. This should actually be a flag to the `composer require` command. `composer update` could have interaction to specify just a subset of packages to update interactively, but not add new packages.

3. I don't see myself using this though. The `composer require` command is quite sufficient for my use. Could you provide some examples where a similar feature exists in another dependency/package manager?

SofHad:
Hi @hkdobrev and thank you for your feedbacks.
The idea is to provide the highest degree of user comfort, I think it is suited to composer update.
We can adapt it to the composer require command too.
I try to make the adjustments that you have suggested as soon as i can.

SofHad:
This is done

en3

src/Composer/Command/UpdateCommand.php
```diff
@@ -135,4 +145,102 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         return $install->run();
     }
+
+    private function getPackagesInteractively(
+        InputInterface $input,
+        OutputInterface $output,
+        Composer $composer
```
stof: you should keep this on 1 line

src/Composer/Command/UpdateCommand.php
```diff
@@ -135,4 +145,102 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+    ) {
+        if ($input->getArgument('packages')) {
+            throw new \InvalidArgumentException(
+                'The option --interactive does not require an argument'
```
stof: why forbidding to use them together ? Just treat them as already included

src/Composer/Command/UpdateCommand.php
```diff
@@ -135,4 +145,102 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            array_merge($InstalledPackageNames, $requiredPackageNames)
+        );
+
+        $vendorDir      = $config->get('vendor-dir', true);
+        $vendorWildcard = sprintf('%s/*', $vendorDir);
+
+        $autocompleterValues = array($vendorWildcard);
+        foreach ($names as $name) {
+            $autocompleterValues[] = $name;
+            $autocompleterValues[] = sprintf('%s/%s', $vendorDir, $name);
```
stof: I don't understand the reason to pass a vendor directory to an autocompleter for package names

src/Composer/Command/UpdateCommand.php
SofHad: I will do it.

src/Composer/Command/UpdateCommand.php
SofHad: I think it would be a good idea.
Ok, I will do it too.

SofHad:
To have the autocomplete suggestions for the words starting with vendor:
* vendor/*
* vendor/doctrine/orm
* doctrine/orm
* why not for vendor/doctrine/*

what do you think?

stof:
``vendor/doctrine/*`` does not make sense. The package whitelist is about package names. ``vendor/doctrine/orm`` and ``vendor/doctrine/*`` are invalid choices

SofHad:
vendor/doctrine/*  was not yet implemented contrary to vendor/doctrine/orm.
 I will update it. I leave only:
* vendor/*
* doctrine/orm

SofHad:
@stof 
I fixed these elements and removed the vendor for the autocomplete values.

src/Composer/Command/UpdateCommand.php
```diff
@@ -74,13 +79,17 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $composer = $this->getComposer(true, $input->getOption('no-plugins'));
```
stof: This is wrong. You cannot move it before the handling of the ``--no-custom-installers`` option

SofHad: Yes, of course. 
This is done :)

src/Composer/Command/UpdateCommand.php
```diff
@@ -78,9 +83,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $output->writeln('<warning>You are using the deprecated option "no-custom-installers". Use "no-plugins" instead.</warning>');
             $input->setOption('no-plugins', true);
         }
-
         $composer = $this->getComposer(true, $input->getOption('no-plugins'));
-        $composer->getDownloadManager()->setOutputProgress(!$input->getOption('no-progress'));
```
stof: removing this line is wrong

src/Composer/Command/UpdateCommand.php
```diff
@@ -78,9 +83,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+        $packages = $input->getArgument('packages');
+        if ($input->getOption('interactive')) {
```
stof: use ``$input->isInteractive()`` instead for the proper way to check for interactive input

src/Composer/Command/UpdateCommand.php
```diff
@@ -78,9 +83,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $output->writeln('<warning>You are using the deprecated option "no-custom-installers". Use "no-plugins" instead.</warning>');
             $input->setOption('no-plugins', true);
         }
-
```
stof: please keep this line

src/Composer/Command/UpdateCommand.php
```diff
@@ -135,4 +144,80 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+            if (is_string($addedPackage)) {
+                if (!in_array($addedPackage, $packages)) {
+                    $packages[] = $addedPackage;
+                    $autocompleterValues = array_diff($autocompleterValues, array($addedPackage));
+                } else {
+                    $output->writeln(sprintf('<error>The package "%s" was already added.</error>', $package));
+                }
+            } else {
+                $continue = false;
```
stof: use ``break;`` to break the loop instead of this variable

SofHad: Done

src/Composer/Command/UpdateCommand.php
SofHad: I don't know how it happened. :grimacing:

src/Composer/Command/UpdateCommand.php
SofHad: I added

```php
private function isInteractive($input)
{
     return $input->getOption('interactive');
}
```

stof: hmm, sorry, I confused the meaning of the ``interactive`` option here. No need to add a getter for that

src/Composer/Command/UpdateCommand.php
```diff
@@ -135,4 +147,84 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         return $install->run();
     }
+
+    private function getPackagesInteractively(InputInterface $input, OutputInterface $output, Composer $composer, $packages)
+    {
```
stof: you should do a check on ``$input->isInteractive()`` here and throw an exception saying that ``--interactive`` cannot be used in non-interactive terminals

SofHad: Done

SofHad:
ping @naderman @Seldaek

Seldaek:
Thanks @SofHad and sorry for the delay looking at this

SofHad:
@stof thank you for your cooperation ;-)

src/Composer/Command/UpdateCommand.php
```diff
@@ -135,4 +147,83 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $requiredPackageNames = array();
+        foreach (
+            array_merge(
+                $composer->getPackage()->getRequires(),
+                $composer->getPackage()->getDevRequires()
+            ) as $require) {
+            $requiredPackageNames[] = $require->getTarget();
+        }
+
+        $InstalledPackageNames = array();
```
SofHad: I will fix this PSR error in another PR

Seldaek:
@SofHad I already fixed the PSR stuff in 633e9f91a1a93bec5a22843cf1bba7d2bf97260a don't worry about it.

PR #3641, opened 2015-01-10
Updated the documentation of the require command

cebe:
It did not mention that it is possible to leave out the version constraint.


Seldaek:
Thanks

PR #3643, opened 2015-01-12
Added missing $listeners variable into EventDispatcher

vkartaviy:
Missing variable triggers ```PHP Notice: Undefined property: Composer\EventDispatcher\EventDispatcher::$listeners``` which throws ```ErrorException```


Seldaek:
Thanks

PR #3651, opened 2015-01-19
fix-token-retrying-itself

### LINKED ISSUE #3609: auth.json refers to "github.com" instead of "api.github.com" for authenticated api calls
I recently ran into a rate limit issue when I was running a composer update on a laravel project.  

I logged in via my credentials when prompted and had no luck.  I made sure the OAuth token in auth.json was correct, still no luck.

It finally worked when I noticed that the auth.json was mapping the OAuth token to github.com requests rather than api.github.com... so I created a new auth.json entry with the same OAuth token instead pointing to api.github.com, and voila, everything works as expected.

alcohol:
should close #3609


webdevilopers:
Great work, @Seldaek . Any idea when the next composer version will be available?
How can I install your version with the fix on my server?

alcohol:
Anything merged is immediately available. Just run `composer self-update`.

webdevilopers:
Indeed, thanks! :+1:

Issue #3654, opened 2015-01-20
[ErrorException] current() expects parameter 1 to be array, string given


theravel:
Hi,

I played with "repositories" in composer.json and I found an error case, when composer fails with just PHP error.

The correct config would be:
```
"repositories": [
    {
        "type": "vcs",
        "url": "https://github.com/whatever.git"
    }
]
```

But I did a mistake and wrote:
```
"repositories": {
    "type": "vcs",
    "url": "https://github.com/whatever.git"
}
```

Take a look at the screenshot how composer handles this case currently:
![screenshot from 2015-01-20 13 14 18](https://cloud.githubusercontent.com/assets/1029434/5815404/8322fabe-a0a6-11e4-8cbb-0f1865fd9392.png)

I guess this should be handled more graceful.

theravel:
Please review my pull request with suggested fix: https://github.com/composer/composer/pull/3655

PR #3655, opened 2015-01-20
Fix error handling for incorrect "repositories" array

theravel:
This is the fix for this issue:
https://github.com/composer/composer/issues/3654

Take a look at the screenshots:

Before fix:
screenshot from 2015-01-20 13 14 18

After fix:
screenshot from 2015-01-20 13 14 33

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -124,7 +124,7 @@ class Config
                 }
 
                 // disable a repository with an anonymous {"name": false} repo
-                if (1 === count($repository) && false === current($repository)) {
+                if (is_array($repository) && 1 === count($repository) && false === current($repository)) {
                     unset($this->repositories[key($repository)]);
                     continue;
                 }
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

alcohol:
Don't mean to be a wanker, but the error message you provided is misleading.

Type should not be an array. Repositories instead expects an array, not an object. Example:

```
"repositories": [
        { "type": "vcs", "url": "http://github.com/mycompany/privaterepo" },
        { "type": "vcs", "url": "http://svn.example.org/private/repo" },
        { "type": "vcs", "url": "http://github.com/mycompany/privaterepo2" }
    ]
```

Can I suggest adjusting and resubmitting the PR?

theravel:
@alcohol "type" here is considered as repository name, not repository type. But I agree message looks misleading. Probably should we change it to:
```
Repository "type" ("vcs") should be an array, string given
```

Seldaek:
I'll adjust it.

alcohol:
I would just change it to, but feel free to improvise

```
Repositories should be an array of repository defintions, repository given
```

PR #3661, opened 2015-01-21
Altered output of diagnose

alcohol:
- added some docblocks
- added rate limit check for users that have no oauth token configured
- made it 'scream' less (yellow text instead of yellow background)
- reformatted the php ini checks slightly to be more condensed

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -271,7 +271,7 @@ EOT
 
         try {
             $url = $domain === 'github.com' ? 'https://api.'.$domain.'/rate_limit' : 'https://'.$domain.'/api/rate_limit';
-            $json = $this->rfs->getContents($domain, $url, array('retry-auth-failure' => false));
+            $json = $this->rfs->getContents($domain, $url, false, array('retry-auth-failure' => false));
             $data = json_decode($json, true);
 
             return $data['rate'];
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -388,7 +388,7 @@ EOT
             foreach ($errors as $error => $current) {
                 switch ($error) {
                     case 'php':
-                        $text = "- Your PHP ({$current}) is too old, you must upgrade to PHP 5.3.2 or higher.";
+                        $text = "Your PHP ({$current}) is too old, you must upgrade to PHP 5.3.2 or higher.";
                         break;
 
                     case 'allow_url_fopen':
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -29,8 +29,13 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 class DiagnoseCommand extends Command
 {
+    /** @var RemoteFileSystem */
     protected $rfs;
+
+    /** @var ProcessExecutor */
     protected $process;
+
+    /** @var int */
     protected $failures = 0;
 
     protected function configure()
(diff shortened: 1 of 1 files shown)


alcohol:
with oauth token:
```
rob@emac019 /srv/git/github/composer (modify-diagnose-output) $ bin/composer diagnose
Checking composer.json: OK
Checking platform settings: FAIL
The xdebug extension is loaded, this can slow down Composer a little.
 Disabling it when using Composer is recommended.
Checking git settings: OK
Checking http connectivity: OK
Checking disk free space: OK
Checking composer version: OK
```
without oauth token:
```
rob@emac019 /srv/git/github/composer (modify-diagnose-output) $ bin/composer diagnose
Checking composer.json: OK
Checking platform settings: FAIL
The xdebug extension is loaded, this can slow down Composer a little.
 Disabling it when using Composer is recommended.
Checking git settings: OK
Checking http connectivity: OK
Checking disk free space: OK
Checking composer version: OK
```
without oauth token and nearing rate limit:
```
rob@emac019 /srv/git/github/composer (modify-diagnose-output) $ bin/composer diagnose
Checking composer.json: OK
Checking platform settings: FAIL
The xdebug extension is loaded, this can slow down Composer a little.
 Disabling it when using Composer is recommended.
Checking git settings: OK
Checking http connectivity: OK
Checking github.com rate limit: WARNING
Github has a rate limit on their API. You currently have 8 out of 60 requests left.
See https://developer.github.com/v3/#rate-limiting and also
    https://getcomposer.org/doc/articles/troubleshooting.md#api-rate-limit-and-oauth-tokens
Checking disk free space: OK
Checking composer version: OK
```

src/Composer/Command/DiagnoseCommand.php
```diff
@@ -227,10 +251,33 @@ private function checkGithubOauth($domain, $token)
...
-            return $this->rfs->getContents($domain, $url, false) ? true : 'Unexpected error';
+            return $this->rfs->getContents($domain, $url, false, array(
+                'retry-auth-failure' => false
+            )) ? true : 'Unexpected error';
+        } catch (\Exception $e) {
+            if ($e instanceof TransportException && $e->getCode() === 401) {
+                return '<comment>The oauth token for '.$domain.' seems invalid, run "composer config --global --unset github-oauth.'.$domain.'" to remove it</comment>';
+            }
+
+            return $e;
```
stof: returning an exception looks wrong to me. Shouldn't it throw instead ?

alcohol: To be honest I was not sure myself, since all other methods return them instead of throw them. I think that is because `outputResult()` handles those.

stof: the are returning strings, not Exception instances

stof: the other possibility is that you should return something else

alcohol: Oh? `checkHttp()`, `checkHttpProxy()` and the original `checkGithubOauth()` would disagree with that assessment. But if you prefer I will refactor them all to return strings instead. No problem, just let me know.

src/Composer/Command/DiagnoseCommand.php
```diff
@@ -271,7 +318,7 @@ private function outputResult(OutputInterface $output, $result)
             if ($result instanceof \Exception) {
```
alcohol: This specifically handles the returned exceptions. Should I remove it as well?

src/Composer/Command/DiagnoseCommand.php
stof: ah sorry, I forgot this. So that's fine

Seldaek:
Thanks :) also damn.. `g fetch alcohol`

PR #3662, opened 2015-01-21
Added the ability for packages to specify a release-branch-prefix, inste...

rawkode:
...ad of assuming "1.0" or "v1.0"

### DIFF
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -28,6 +28,8 @@ class VersionParser
 {
     private static $modifierRegex = '[._-]?(?:(stable|beta|b|RC|alpha|a|patch|pl|p)(?:[.-]?(\d+))?)?([.-]?dev)?';
 
+    protected $releaseBranchPrefix;
+
     /**
      * Returns the stability of a version
      *
@@ -197,6 +199,10 @@ class VersionParser
 
         if (in_array($name, array('master', 'trunk', 'default'))) {
             return $this->normalize($name);
+        }
+
+        if (!is_null($this->releaseBranchPrefix)) {
+            $name = str_replace($this->releaseBranchPrefix, '', $name);
         }
 
         if (preg_match('#^v?(\d+)(\.(?:\d+|[xX*]))?(\.(?:\d+|[xX*]))?(\.(?:\d+|[xX*]))?$#i', $name, $matches)) {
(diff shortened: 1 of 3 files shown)


stof:
Well, the issue is that this won't help if you intend your package to be registered on Packagist, because packagist does not allow custom repo configuration anyway.

rawkode:
It doesn't break any current functionality and will allow people running their own satis repositories to use their own release branch schema (Stash for instance defaults to release/)

rawkode:
I do not know how Packagist works, but if repository owners use release-1.0 for release branches and add release-branch-prefix to their composer.json, surely it would still parse correctly with this fix and work on packagist?

stof:
The issue is that ``release-branch-prefix`` does not go in the composer.json of the package. It goes in the repository configuration

src/Composer/Package/Version/VersionParser.php
```diff
@@ -28,6 +28,8 @@ class VersionParser
 {
     private static $modifierRegex = '[._-]?(?:(stable|beta|b|RC|alpha|a|patch|pl|p)(?:[.-]?(\d+))?)?([.-]?dev)?';
 
+    protected $releaseBranchPrefix;
```
stof: should not be protected

src/Composer/Repository/VcsRepository.php
```diff
@@ -280,8 +286,12 @@ protected function preProcess(VcsDriverInterface $driver, array $data, $identifi
         return $data;
     }
 
-    private function validateBranch($branch)
+    private function validateBranch($branch, $releaseBranchPrefix = null)
     {
+        if (!is_null($releaseBranchPrefix)) {
+            $this->versionParser->setReleaseBranchPrefix($releaseBranchPrefix);
```
stof: changing the state of the version parser can create wrong side effects (it will affect the following repos too)

src/Composer/Repository/VcsRepository.php
```diff
@@ -218,6 +219,11 @@ protected function initialize()
                     continue;
                 }
 
+                // Update parsedBranch if this package's composer.json specifies a prefix for their release branches
+                if (array_key_exists('config', $data) && array_key_exists('release-branch-prefix', $data['config'])) {
```
rawkode: It does get loaded from the packages composer.json. My tests used a package with:

```
{
  "name": "my-package",
  "config": {
    "bin-dir": "bin",
    "vendor-dir": "vendor",
    "release-branch-prefix": "release/"
  }
```

stof: this is totally inconsistent with the way other repository-level options are provided.

rawkode: What would you suggest as a better configuration and I'll modify to meet?

Seldaek:
I don't see this as a workable thing, given it would only work for private packages and it really does not provide much benefit from a user point of view. It complicates things to avoid making a minor tweak to your process.

Issue #3663, opened 2015-01-22
composer config --file shows incorrect error message


cjunge-work:
when trying to use composer to edit a different file name, I get the following error:

```[RuntimeException]                               
  No composer.json found in the current directory```

Even though the name of the requested file is not composer.json, and there is a composer.json in the directory, the error message seems to be hard-coded to composer.json.

Example call:
```composer config --file=auth.json -e```

PR #3665, opened 2015-01-22
#3663 fix error message for config command

theravel:
Minor fix for https://github.com/composer/composer/issues/3663

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -137,7 +137,7 @@ EOT
         }
 
         if (!$this->configFile->exists()) {
-            throw new \RuntimeException("File '$configFile' cannot be found in the current directory");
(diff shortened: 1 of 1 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -137,7 +137,7 @@ protected function initialize(InputInterface $input, OutputInterface $output)
         }
 
         if (!$this->configFile->exists()) {
-            throw new \RuntimeException('No composer.json found in the current directory');
+            throw new \RuntimeException("File '$configFile' cannot be found in the current directory");
```
Nek-: You should use `sprintf` (and simple quotes).

```php
<?php
throw new \RuntimeException(sprintf('File "%s" cannot be found in the current directory', $configFile);
```

theravel: Agree, thanks

PR #3675, opened 2015-01-25
Add note about /usr not existing in OSX Yosemite by default.

Shelob9:
BTW Seriously Apple?


naderman:
Pretty sure you would have to create `/usr/local/bin` as composer is the executable file? And receive is also misspelled?

Shelob9:
@naderman You are right, I updated the PR.

PR #3677, opened 2015-01-26
Install plugin dependencies before plugin installation

francoispluchino:
| Q             | A
| ------------- | -------------------------------
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2972
| License       | MIT

> **Note:** This PR is an extraction of the bug fix of 'order of installation of plugin dependencies' fixed by the PR #3082

Fixes a "strange" behavior of the order of installation of plugins and their dependencies.

The plugin dependencies must be installed before the plugin, then the plugin must be installed, and then the rest of depedencies can be installed.

Test example:

```json
{
    "repositories": [
        {
            "type": "package",
            "package": [
                { "name": "pkg", "version": "1.0.0" },
                { "name": "pkg2", "version": "1.0.0" },
                { "name": "inst", "version": "1.0.0", "type": "composer-plugin" },
                { "name": "inst-with-req", "version": "1.0.0", "type": "composer-plugin", "require": { "php": ">=5", "ext-json": "*", "composer-plugin-api": "*" } },
                { "name": "inst-with-req2", "version": "1.0.0", "type": "composer-plugin", "require": { "pkg2": "*" } }
            ]
        }
    ],
    "require": {
        "pkg": "1.0.0",
        "inst": "1.0.0",
        "inst-with-req2": "1.0.0",
        "inst-with-req": "1.0.0"
    }
}
```

The order should be:

```
Installing inst (1.0.0)
Installing pkg2 (1.0.0)
Installing inst-with-req2 (1.0.0)
Installing inst-with-req (1.0.0)
Installing pkg (1.0.0)
```

and not (actually):

```
Installing inst (1.0.0)
Installing inst-with-req (1.0.0)
Installing pkg (1.0.0)
Installing pkg2 (1.0.0)
Installing inst-with-req2 (1.0.0)
```

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -622,17 +622,13 @@ class Installer
      */
     private function movePluginsToFront(array $operations)
     {
-        $installerOps = array();
-        foreach ($operations as $idx => $op) {
-            if ($op instanceof InstallOperation) {
-                $package = $op->getPackage();
-            } elseif ($op instanceof UpdateOperation) {
-                $package = $op->getTargetPackage();
-            } else {
+        $pluginDependencies = array();
+        foreach ($operations as $op) {
+            if (null === $package = $this->getOperationPackage($op)) {
                 continue;
             }
 
-            if ($package->getType() === 'composer-plugin' || $package->getType() === 'composer-installer') {
+            if ($this->isPlugin($package)) {
                 // ignore requirements to platform or composer-plugin-api
                 $requires = array_keys($package->getRequires());
                 foreach ($requires as $index => $req) {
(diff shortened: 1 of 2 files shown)


naderman:
This entirely ignores multiple levels of dependencies. Add some tests, where the dependencies themselves have dependencies. Also verify that installation order remains correct, if there are plugins depending on plugins, and dependencies using different plugins depending on each other. That's why currently only plugins without dependencies are moved to the front.

francoispluchino:
Just for reference: finally implemented by the PR #6371.

PR #3678, opened 2015-01-26
Add the P character to the regex pattern

arcturial:
According to http://php.net/manual/en/function.preg-match.php and some other sources named groups should contain a 'P' character after the '?'

Without this, I receive the following error when running an update:

[ErrorException]                                                                
  preg_match(): Compilation failed: unrecognized character after (?< at offset 4  
                                                                                  
Exception trace:
 () at phar:///directory/composer.phar/src/Composer/Package/Version/VersionParser.php:181

### DIFF
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -178,7 +178,7 @@ class VersionParser
      */
     public function parseNumericAliasPrefix($branch)
     {
-        if (preg_match('/^(?<version>(\d+\\.)*\d+)(?:\.x)?-dev$/i', $branch, $matches)) {
+        if (preg_match('/^(?P<version>(\d+\\.)*\d+)(?:\.x)?-dev$/i', $branch, $matches)) {
             return $matches['version'].".";
         }
 


pborreli:
@arcturial what is the version of your PHP ?

stof:
@pborreli the actual question is against which version of PCRE it was built (which can be different than the one shipped with PHP as most distributions are building PHP against the system PCRE)

This is not the right place to add it anyway. You are not switching to the older syntax here. You are renaming the placeholder. This is wrong (as shown by the testsuite failure)

arcturial:
I am using PHP 5.3.3 and the PCRE version is 6.6 06 Feb 2006. I think it's safe to assume that it's probably an different/outdated version of PCRE like @stof suggsted. 

Before I fix the position of the 'P' in the regex (I apologize for this...trying to commit before leaving work instead of doing it properly), are there any better ways of dealing with this issue? Unfortunately the software versions are slightly out of my control at this point.

arcturial:
It seems that both (?<group>) and (?P<group>) works, however...the (?<group>) syntax is not supported in the older versions. Adding the P makes it backwards compatible.

naderman:
@arcturial can you squash this into one commit please?

arcturial:
I squashed it into the original commit.

PR #3681, opened 2015-01-27
Don't use a regex to parse installer tests to allow for longer tests

naderman:


tests/Composer/Test/InstallerTest.php
```diff
@@ -280,48 +266,44 @@ public function getIntegrationTests()
...
+                }
+                $run = $testData['RUN'];
+                if (!empty($testData['EXPECT-LOCK'])) {
+                    $expectLock = JsonFile::parseJson($testData['EXPECT-LOCK']);
+                }
+                $expectOutput = isset($testData['EXPECT-OUTPUT']) ? $testData['EXPECT-OUTPUT'] : null;
+                $expect = $testData['EXPECT'];
+                $expectExitCode = isset($testData['EXPECT-EXIT-CODE']) ? (int) $testData['EXPECT-EXIT-CODE'] : 0;
+            } catch (\Exception $e) {
+                die(sprintf('Test "%s" is not valid: '.$e->getMessage(), str_replace($fixturesDir.'/', '', $file)));
```
stof: Please don't use die. Killing the phpunit process is not a valid option

Issue #3647, opened 2015-01-15
Verbose (-vv) output with profiling (--profile) results in formatting misalignment


danemacmillan:
Running `composer update -vv --profile` produces verbose output with memory/time profiling output as well. This works fine. The formatting issue occurs when a dependency lists the changes it is pulling in. View the below excerpt for an example:

```
[145.2MB/55.86s] Updating dependencies
[480.6MB/80.87s]   - Updating linusshops/web-magento-theme-linus-creation dev-stage (37c1012 => 97d843f)
[480.7MB/81.49s]     Checking out 97d843f45fac826305124d36fae9fb76b79bdd94
[480.7MB/86.60s]     Pulling in changes:
[480.7MB/86.61s]       97d843f - Dane MacMillan: Merge branch 'develop' into stage
      3e57ea8 - Dane MacMillan: WEBBUG-115 #resolve Ensure cart with admin-deleted items notify user.
[480.7MB/86.61s]
[250.9MB/90.59s] Writing lock file
[250.9MB/90.59s] Generating autoload files
Memory usage: 250.99MB (peak: 481.31MB), time: 90.7s
```
Note that the line beginning with `3e57ea8` does not display any profiling data, which results in the misalignment.

Seldaek:
Then again --profile is a debug thing so it's not the most urgent issue but yes it could be fixed fairly easily in the ConsoleIO :)

danemacmillan:
Yeah, the urgency of this is almost bordering on WONTFIX, but I figured why not. Extra polish never hurt nobody.

PR #3685, opened 2015-01-28
fix for issue #3647

xfornesa:
ConsoleIO writing a message which contains end of lines characters

#3647 

Edited: wrong issue number, changed it

### DIFF
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -98,7 +98,7 @@ class ConsoleIO extends BaseIO
         if (null !== $this->startTime) {
             $messages = (array) $messages;
             $startTime = $this->startTime;
-            $messages = array_map(function (&$message) use ($startTime) {
+            $messages = array_map(function ($message) use ($startTime) {
                 return sprintf(
                     '[%.1fMB/%.2fs] %s',
                     memory_get_usage() / 1024 / 1024,
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -96,16 +96,11 @@ class ConsoleIO extends BaseIO
     public function write($messages, $newline = true)
     {
         if (null !== $this->startTime) {
-            $messages = (array) $messages;
-            $startTime = $this->startTime;
-            $messages = array_map(function ($message) use ($startTime) {
-                return sprintf(
-                    '[%.1fMB/%.2fs] %s',
-                    memory_get_usage() / 1024 / 1024,
-                    microtime(true) - $startTime,
-                    $message
-                );
-            }, $messages);
+            $memoryUsage = memory_get_usage() / 1024 / 1024;
+            $timeSpent = microtime(true) - $this->startTime;
+            $messages = array_map(function ($message) use ($memoryUsage, $timeSpent) {
+                return sprintf('[%.1fMB/%.2fs] %s', $memoryUsage, $timeSpent, $message);
+            }, (array) $messages);
         }
         $this->output->write($messages, $newline);
         $this->lastMessage = join($newline ? "\n" : '', (array) $messages);
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -97,11 +97,12 @@ class ConsoleIO extends BaseIO
     {
         if (null !== $this->startTime) {
             $messages = (array) $messages;
-            $messages = array_map(function (&$message) {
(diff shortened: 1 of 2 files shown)


naderman:
The issue you referenced seems unrelated, which one specifically did you mean to fix with this? I restarted the two failed builds on travis so hopefully fine now.

naderman:
Ah it actually fails on PHP 5.3, because you used $this in a closure, please keep in mind composer should remain compatible with PHP 5.3.

xfornesa:
Some kind of dislexia :P I put the wrong issue number, I was trying to solve an easy pick for "Verbose (-vv) output with profiling (--profile) results in formatting misalignment"
I'll try to check with php 5.3

xfornesa:
yep, 5.3 working :)
Sorry for not noticing before travis fail for php 5.3

src/Composer/IO/ConsoleIO.php
```diff
@@ -97,12 +97,15 @@ public function write($messages, $newline = true)
...
         if (null !== $this->startTime) {
             $messages = (array) $messages;
-            $messages[0] = sprintf(
-                '[%.1fMB/%.2fs] %s',
-                memory_get_usage() / 1024 / 1024,
-                microtime(true) - $this->startTime,
-                $messages[0]
-            );
+            $startTime = $this->startTime;
+            $messages = array_map(function (&$message) use ($startTime) {
```
naderman: why do you pass $message by reference here? Would be better if you don't.

xfornesa: Agree with you.
At the begging I wrote an array_walk (by reference), but immediately change to array_map to avoid modify original message parameter, I forgot to change from by reference to by value.

src/Composer/IO/ConsoleIO.php
```diff
@@ -97,12 +97,15 @@ public function write($messages, $newline = true)
...
         if (null !== $this->startTime) {
             $messages = (array) $messages;
-            $messages[0] = sprintf(
-                '[%.1fMB/%.2fs] %s',
-                memory_get_usage() / 1024 / 1024,
-                microtime(true) - $this->startTime,
-                $messages[0]
-            );
+            $startTime = $this->startTime;
+            $messages = array_map(function ($message) use ($startTime) {
```
GromNaN: A simple `foreach` would have avoided 2 method calls.

xfornesa: Yep
two options with some improvements:
with array_map:

```php
$memoryUsage = memory_get_usage() / 1024 / 1024
$timeSpent = microtime(true) - $this->startTime;
$messages = array_map(function ($message) use ($memoryUsage, $timeSpent) {
    return sprintf('[%.1fMB/%.2fs] %s', $memoryUsage, $timeSpent, $message);
}, (array) $messages);
```

versus using foreach:

```php
$memoryUsage = memory_get_usage() / 1024 / 1024;
$timeSpent = microtime(true) - $this->startTime;
$originalMessages = $messages;
$messages = array();
foreach ($originalMessages as $message) {
    $messages[] = sprintf('[%.1fMB/%.2fs] %s', $memoryUsage, $timeSpent, $message);
}
```

For the foreach option I do not like the initialisations, it seems a weird hack just to make things work:

```php
$originalMessages = $messages;
$messages = array();
```

I'd like the first option, I found easier to understand and more clear on the intention: transform the array of messages. With a foreach we simply iterate and then generate a new array messages that we will use instead of the original ones.

I will push this last refactor but open to change to foreach also.
What do you think @GromNaN ?

naderman:
This seems fine to me.

src/Composer/IO/ConsoleIO.php
GromNaN: You can work on the same array and replace values one by one.
```php
$memoryUsage = memory_get_usage() / 1024 / 1024;
$timeSpent = microtime(true) - $this->startTime;
foreach ($messages as $key => $message) {
    $messages[$key] = sprintf('[%.1fMB/%.2fs] %s', $memoryUsage, $timeSpent, $message);
}
```

xfornesa:
Yuju!!

src/Composer/IO/ConsoleIO.php
xfornesa: Yep, it's cleaner.
I cannot defend any below the other, so I need someone from composer to decide which implementation wants to merge.

naderman: think it's fine as is.

xfornesa:
cool

PR #3688, opened 2015-01-28
Reuse current file permissions in selfupdate

jeremy-derusse:
In selfupdate command, reuse original file's permission instead of using `umask`


avindra:
+1. Would like to see this integrated. (Comment for reference).

Seldaek:
Thanks

PR #3693, opened 2015-01-29
Delegate problem formatting to a reserved class

Pherserk:
In this way problems become more similar to a plain/storae object for reasons.
It is possible to work on problem formatter without change Problem behaviour.

### DIFF
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -31,14 +31,7 @@ class Problem
      */
     protected $reasons = array();
 
-    protected $section = 0;
-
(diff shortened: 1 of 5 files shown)


naderman:
I think it would make more sense to have one problem independent formatter and passing the problem into the getPrettyString function?

Pherserk:
Yes should be a static method of Formatter, but it also depends on Pool that is context dependent.

Pherserk:
Ok I Will refactor avoidib Construction dependencies

Pherserk:
mmm.... need to change the Solver... I'll get some time :)

PR #3699, opened 2015-02-02
added non-feature-branches

rk3rn3r:
Hello!

We had a problem on handling our release and testing branches with composer correctly, when using "self.version" on dependencies, when the release branch has a non-numeric / non-version name, like "latest" or "latest-testing" or "latest-production".

If the root package branch name does not look like a version number, then composer handles the branch as feature branch, which means, it checks all ancestors until it finds a parent branch or tag that looks like a version number or at least ends on master/develop/trunk.

This is a problem if during a testing or release phase you are working with (and also releasing) branches that are named "latest-testing" "current-release" or whatever, when you have dependencies defined with self.version. Since latest-testing is a branch created from master some time ago, self.version actually becomes "master". then latest-testing from our application depends on master of its self.version dependencies, instead of also depend on latest-testing branch from these dependencies.

I added a configuration, where you can specify, what patterns should not be handled like feature branches.
in your composer.json, you can now add something like:
```
"non-feature-branches": ["latest-.*", "release-.*"]
```

all branches matching given patterns will not be handled as feature branches, but as normal "dev" branches, so "latest-testing" becomes version "dev-latest-testing", "latest-production" is version "dev-latest-production".
Then self.version will reference the correct branches of these dependencies, that make use of self.version.

What do you think?

Best regards,
René.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -785,8 +785,6 @@ The example will include `/dir/foo/bar/file`, `/foo/bar/baz`, `/file.php`,
 
 Optional.
 
-&larr; [Command-line interface](03-cli.md)  |  [Repositories](05-repositories.md) &rarr;
-
 ### non-feature-branches
 
 A list of regex patterns of branch names that are non-numeric (e.g. "latest" or something), that will NOT be handled as feature branches. This is an array of string.
@@ -815,3 +813,5 @@ Then "composer show -s" will give you "versions : * dev-latest-testing".
 
 
 Optional.
+
+&larr; [Command-line interface](03-cli.md)  |  [Repositories](05-repositories.md) &rarr;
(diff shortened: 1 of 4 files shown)


markuspoerschke:
+1 Like!

dsiebel:
+1 :smile:

janvt:
+1

rk3rn3r:
hey @naderman  @Seldaek what do you think?

rk3rn3r:
still no progress on this? any feedback?

Seldaek:
@rk3rn3r looks good to me, although I am not a huge fan of adding a million options for rare use cases, but given the amount of code it's reasonable here.

Could you add it to the list of config options in `doc/04-schema.md` though?

rk3rn3r:
@Seldaek I totally agree on this. Too much options don't make sense, especially when they are for rare cases and the changeset is big.
I will do the change in 04-schema.md later today :-)

rk3rn3r:
hehe I'm not so trained in markdown :-D

Seldaek:
Thanks!

PR #3704, opened 2015-02-03
Adding bin-compat configuration option

### LINKED ISSUE #2771: Add flag to force .bat file creation in bin
When working with a shared folder in a VM I'd like to be able to run Composer on the host machine (Mac, Linux) while still building the .bat files for the VM (Windows).

kocsismate:
Hey,

This is a "minimum viable" pull request which closes #2771. I hope that you'll like it! Could you tell me what else I should implement? Do you want me to add test cases for this config option?

Anyway, I tested my composer.phar in various scenarios equally on Windows and on a Ubuntu Vagrant box (as mentioned in the issue) and everything proved to be OK.


kocsismate:
Oh, I am not sure why my build fails only on 5.5 with timeout error :( Any idea?

src/Composer/Config.php
```diff
@@ -215,6 +216,17 @@ public function get($key, $flags = 0)
             case 'home':
                 return rtrim($this->process($this->config[$key], $flags), '/\\');
 
+            case 'bin-compat':
+                $value= $this->getComposerEnv('COMPOSER_BIN_COMPAT') ?: $this->config[$key];
```
alcohol: Whitespace before assignment operator please :-)

kocsismate: Ah, sure :) It is so difficult to get rid of this bad habit! I'll fix it in some minutes!

res/composer-schema.json
```diff
@@ -181,6 +181,10 @@
                     "type": ["string", "integer"],
                     "description": "The cache max size for the files cache, defaults to \"300MiB\"."
                 },
+                "bin-compat": {
+                    "type": "string",
```
stof: rather than ``string``, it would be better to use ``enum`` given that only 3 strings are valid, not any string.

kocsismate: Good catch, thank you! I've fixed it.

kocsismate:
@Seldaek , I'd like to ask your opinion about this PR! Without this patch, Composer can't be used along with VirtualBox Vagrants boxes on Windows when one runs ``composer install`` inside the box. I know that it isn't the most frequest use-case, but e.g. if you want CI on your Windows, the bug right now is quite annoying.

alcohol:
As far as I can tell the changes look solid.

But you need to resolve some merge conflicts. Please do that, cause it will affect the change-set displayed right now :-).

Just one question; are you certain that moving the `getcwd()` and `chdir()` calls all within the try/catch of `installSymlinkBinaries()` doesn't break the flow in some way? Before it would `getcwd()`, try/catch, then outside of that scope `chdir()` back again. It's the only notable change I could discover.

kocsismate:
OK, I resolved the conflicts! They were really minor, fortunately.

I put ``getcwd()`` and ``chdir()`` into the ``Symlink::symlinkBin()`` method because i didn't want it to have a side-effect. But I think having these two system calls inside the scope of a try block shouldn't break the flow because they don't raise exceptions at all and only the following line can get performed without ``chdir(dirname($link));``: https://github.com/kocsismate/composer/blob/feature-bin-compat/src/Composer/Installer/LibraryInstaller.php#L263 (it is equivalent to https://github.com/composer/composer/blob/bd802407a8f3c6f221749af7d2dee795ae85b84a/src/Composer/Installer/LibraryInstaller.php#L246) which isn't concerned about the file generation-time ``CWD``. 

Tell me, if I got it wrong!

alcohol:
No no, I am not saying it is wrong. I was just pointing out it seems to be the only notable change (that was not directly related to the intent of this PR).

mmeuser:
+1 we would really appreciate a merge, is there anything to be done here?

res/composer-schema.json
alcohol: I don't think you have to keep the type: string definition now that you have added enum.

kocsismate: Yep, I was also unsure, but it's done now!

Seldaek:
Thanks, merged but I removed the nosymlink option because I think that isn't really necessary and I rather keep things simple (see commit above)

PR #3711, opened 2015-02-05
GitHub api v3 deprecations, closes #3566

### LINKED ISSUE #3566: Update to the new GitHub Authorizations API (breaking change)
The [GitHub Authorizations API](https://developer.github.com/v3/oauth_authorizations/) will no longer return `token` as of sometime next year (likely around March). You can read more on the [developer blog](https://developer.github.com/changes/2014-12-08-removing-authorizations-token/). The [login logic](https://github.com/composer/composer/blob/60ac971419ba4e8130b401757894b6b5636cf4e6/src/Composer/Util/GitHub.php#L77) will need to be updated to use something more like https://developer.github.com/v3/oauth_authorizations/#get-or-create-an-authorization-for-a-specific-app-and-fingerprint.

alcohol:
With the new changes to the GitHub authorizations API endpoint, it will no longer be possible to retrieve existing token values. As composer currently does not use an application token, but rather a personal token, I couldn't think of any solution other than simply creating a new token whenever one is not available in the auth configuration.

I also had to change the note to always include the timestamp. A 422 error regarding duplicate content creation would be thrown otherwise as this field has to be unique.

Alternatively, we could still retrieve a user's current authorizations. Then notify the user in case of an existing one with an identical note and instruct them to remove it from https://github.com/settings/applications before proceeding, or merely warn them and proceeding immediately.

If anyone has any suggestions for a better solution, please feel free to comment/reply with your feedback.

Related reading material:
https://developer.github.com/changes/2015-02-03-removing-authorizations-token-update/
https://developer.github.com/changes/2014-12-08-removing-authorizations-token/

Closes #3566

### DIFF
--- a/src/Composer/Util/GitHub.php
+++ b/src/Composer/Util/GitHub.php
@@ -76,124 +76,128 @@ class GitHub
      */
     public function authorizeOAuthInteractively($originUrl, $message = null)
     {
-        $attemptCounter = 0;
-
-        $apiUrl = ('github.com' === $originUrl) ? 'api.github.com' : $originUrl . '/api/v3';
-
         if ($message) {
             $this->io->write($message);
         }
-        $this->io->write('The credentials will be swapped for an OAuth token stored in '.$this->config->getAuthConfigSource()->getName().', your password will not be stored');
+
+        $this->io->write(sprintf('A token will be created and stored in "%s", your password will never be stored', $this->config->getAuthConfigSource()->getName()));
         $this->io->write('To revoke access to this token you can visit https://github.com/settings/applications');
+
+        $otp = null;
+        $attemptCounter = 0;
+
         while ($attemptCounter++ < 5) {
             try {
-                if (empty($otp) || !$this->io->hasAuthentication($originUrl)) {
-                    $username = $this->io->ask('Username: ');
-                    $password = $this->io->askAndHideAnswer('Password: ');
-                    $otp      = null;
-
-                    $this->io->setAuthentication($originUrl, $username, $password);
-                }
-
-                // build up OAuth app name
-                $appName = 'Composer';
-                if ($this->config->get('github-expose-hostname') === true && 0 === $this->process->execute('hostname', $output)) {
-                    $appName .= ' on ' . trim($output);
-                } else {
-                    $appName .= ' [' . date('YmdHis') . ']';
-                }
-
-                $headers = array();
-                if ($otp) {
-                    $headers = array('X-GitHub-OTP: ' . $otp);
-                }
-
-                // try retrieving an existing token with the same name
-                $contents = null;
-                $auths = JsonFile::parseJson($this->remoteFilesystem->getContents($originUrl, 'https://'. $apiUrl . '/authorizations', false, array(
-                    'retry-auth-failure' => false,
-                    'http' => array(
-                        'header' => $headers
-                    )
-                )));
-                foreach ($auths as $auth) {
-                    if (
-                        isset($auth['app']['name'])
-                        && 0 === strpos($auth['app']['name'], $appName)
-                        && $auth['app']['url'] === 'https://getcomposer.org/'
-                    ) {
-                        $this->io->write('An existing OAuth token for Composer is present and will be reused');
-
-                        $contents['token'] = $auth['token'];
-                        break;
-                    }
-                }
-
-                // no existing token, create one
-                if (empty($contents['token'])) {
-                    $headers[] = 'Content-Type: application/json';
-
-                    $contents = JsonFile::parseJson($this->remoteFilesystem->getContents($originUrl, 'https://'. $apiUrl . '/authorizations', false, array(
-                        'retry-auth-failure' => false,
-                        'http' => array(
-                            'method' => 'POST',
-                            'follow_location' => false,
-                            'header' => $headers,
-                            'content' => json_encode(array(
-                                'scopes' => array('repo'),
-                                'note' => $appName,
-                                'note_url' => 'https://getcomposer.org/',
-                            )),
-                        )
-                    )));
-                    $this->io->write('Token successfully created');
-                }
+                $response = $this->createToken($originUrl, $otp);
             } catch (TransportException $e) {
+                // https://developer.github.com/v3/#authentication && https://developer.github.com/v3/auth/#working-with-two-factor-authentication
+                // 401 is bad credentials, or missing otp code
+                // 403 is max login attempts exceeded
                 if (in_array($e->getCode(), array(403, 401))) {
-                    // 401 when authentication was supplied, handle 2FA if required.
-                    if ($this->io->hasAuthentication($originUrl)) {
-                        $headerNames = array_map(function ($header) {
-                            return strtolower(strstr($header, ':', true));
-                        }, $e->getHeaders());
-
-                        if ($key = array_search('x-github-otp', $headerNames)) {
-                            $headers = $e->getHeaders();
-                            list($required, $method) = array_map('trim', explode(';', substr(strstr($headers[$key], ':'), 1)));
-
-                            if ('required' === $required) {
-                                $this->io->write('Two-factor Authentication');
-
-                                if ('app' === $method) {
-                                    $this->io->write('Open the two-factor authentication app on your device to view your authentication code and verify your identity.');
-                                }
-
-                                if ('sms' === $method) {
-                                    $this->io->write('You have been sent an SMS message with an authentication code to verify your identity.');
-                                }
-
-                                $otp = $this->io->ask('Authentication Code: ');
-
-                                continue;
-                            }
+                    // in case of a 401, and authentication was previously provided
+                    if (401 === $e->getCode() && $this->io->hasAuthentication($originUrl)) {
+                        // check for the presence of otp headers and get otp code from user
+                        $otp = $this->checkTwoFactorAuthentication($e->getHeaders());
+                        // if given, retry creating a token using the user provided code
+                        if (null !== $otp) {
+                            continue;
                         }
                     }
 
-                    $this->io->write('Invalid credentials.');
+                    if (401 === $e->getCode()) {
+                        $this->io->write('Bad credentials.');
+                    } else {
+                        $this->io->write('Maximum number of login attempts exceeded. Please try again later.');
+                    }
+
+                    $this->io->write('You can also manually create a personal token at https://github.com/settings/applications');
+                    $this->io->write('Add it using "composer config github-oauth.github.com <token>"');
+
                     continue;
                 }
 
                 throw $e;
             }
 
-            $this->io->setAuthentication($originUrl, $contents['token'], 'x-oauth-basic');
-
-            // store value in user config
+            $this->io->setAuthentication($originUrl, $response['token'], 'x-oauth-basic');
             $this->config->getConfigSource()->removeConfigSetting('github-oauth.'.$originUrl);
-            $this->config->getAuthConfigSource()->addConfigSetting('github-oauth.'.$originUrl, $contents['token']);
+            // store value in user config
+            $this->config->getAuthConfigSource()->addConfigSetting('github-oauth.'.$originUrl, $response['token']);
 
             return true;
         }
 
         throw new \RuntimeException("Invalid GitHub credentials 5 times in a row, aborting.");
+    }
+
+    private function createToken($originUrl, $otp = null)
+    {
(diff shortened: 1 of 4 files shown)


src/Composer/Util/GitHub.php
```diff
@@ -185,15 +105,84 @@ public function authorizeOAuthInteractively($originUrl, $message = null)
...
+            $this->config->getConfigSource()->removeConfigSetting('github-oauth.' . $originUrl);
+            $this->config->getAuthConfigSource()->addConfigSetting('github-oauth.' . $originUrl, $response['token']);
 
             return true;
         }
 
         throw new \RuntimeException("Invalid GitHub credentials 5 times in a row, aborting.");
     }
+
+    protected function createToken($originUrl, $otp = null)
```
stof: I think these methods should be private

src/Composer/Util/GitHub.php
```diff
@@ -185,15 +105,84 @@ public function authorizeOAuthInteractively($originUrl, $message = null)
...
+    {
+        if (is_null($otp) || !$this->io->hasAuthentication($originUrl)) {
```
stof: please use ``null ===`` for consistency with the existing codebase

src/Composer/Util/GitHub.php
```diff
@@ -76,106 +76,26 @@ public function authorizeOAuth($originUrl)
...
-                                    $this->io->write('You have been sent an SMS message with an authentication code to verify your identity.');
-                                }
-
-                                $otp = $this->io->ask('Authentication Code: ');
-
-                                continue;
-                            }
-                        }
+                        $otp = $this->checkTwoFactorAuthentication($e->getHeaders());
+                        continue;
```
stof: this is wrong. You are now continuing even when there is no ``x-github-otp`` header, which changes the logic

src/Composer/Util/GitHub.php
```diff
@@ -185,15 +105,84 @@ public function authorizeOAuthInteractively($originUrl, $message = null)
...
             }
 
-            $this->io->setAuthentication($originUrl, $contents['token'], 'x-oauth-basic');
-
-            // store value in user config
-            $this->config->getConfigSource()->removeConfigSetting('github-oauth.'.$originUrl);
-            $this->config->getAuthConfigSource()->addConfigSetting('github-oauth.'.$originUrl, $contents['token']);
+            $this->io->setAuthentication($originUrl, $response['token'], 'x-oauth-basic');
+            $this->config->getConfigSource()->removeConfigSetting('github-oauth.' . $originUrl);
+            $this->config->getAuthConfigSource()->addConfigSetting('github-oauth.' . $originUrl, $response['token']);
```
stof: please don't change coding standards (you are adding spaces around the concatenation operator)

src/Composer/Util/GitHub.php
```diff
@@ -185,15 +105,84 @@ public function authorizeOAuthInteractively($originUrl, $message = null)
                 throw $e;
             }
 
-            $this->io->setAuthentication($originUrl, $contents['token'], 'x-oauth-basic');
-
-            // store value in user config
```
stof: why are you removing this comment ?

alcohol: It seemed like stating the obvious, but I'll put I guess.

staabm:
It this the reason for the issue I am running into in..
http://imgur.com/gr4vuss

Even after "Token successfully created" I am getting re-asked for authentication and cannot proceed with private reposes?

alcohol:
No, that issue was resolved in #3651. Have you run `composer self-update` recently?

staabm:
I am on
```
root@8754e3e995ce:/dvl/amp-chat# php7 ../composer.phar
Composer version 1.0-dev (07c644ac229a21df80180598d8bb9aaba232eecb) 2015-02-03 12:51:10
```

but running composer via php7

alcohol:
That is odd.. I'll investigate a bit. I see you are using two factor authentication? I have not been able to test this flow myself as I don't use it. But I don't recall it deviating from the normal flow. I'll get back to you in a bit.

alcohol:
Ok, I can't find anything that would prompt your scenario to occur. As far as I know, the old GitHub authorizations API endpoint is still valid. So these deprecations should not affect your flow either.

Can you check if the same issue occurs if you run composer through a stable version of PHP? You can reproduce the scenario where it asks for your authentication by removing `~/.composer/auth.json` (or make a backup) and running `while true ; do curl -v https://api.github.com/dummy ; done`. Once the header value of `X-RateLimit-Remaining` hits 0, try to run `composer install|update` in any of your projects.

staabm:
ok, created issue https://github.com/composer/composer/issues/3718 for further discussion.
sorry for hijacking this PR.

alcohol:
@naderman can you review my latest changes? I would appreciate any feedback

alcohol:
Sorry, that was supposed to be @stof ..

src/Composer/Util/GitHub.php
```diff
@@ -76,124 +76,125 @@ public function authorizeOAuth($originUrl)
...
+        $attemptCounter = 0;
+
         while ($attemptCounter++ < 5) {
             try {
-                if (empty($otp) || !$this->io->hasAuthentication($originUrl)) {
-                    $username = $this->io->ask('Username: ');
-                    $password = $this->io->askAndHideAnswer('Password: ');
-                    $otp      = null;
+                $response = $this->createToken($originUrl, $otp);
+            } catch (TransportException $e) {
```
Seldaek: Maybe here we should `->write('You can also create a token manually on github.com and add it using composer config github-oauth.github.com <token>')` if $attemptCounter>1?

Seldaek:
Apart from that comment it looks great IMO.

src/Composer/Util/GitHub.php
alcohol: We are counting up. So am I correct in understanding that you want to show that for every failed attempt after the first one? And would that be before or after "Bad credentials" message?

Seldaek: The idea is just that if we are failing the job it's nice to say "here 
is another way to do it that probably works" instead of people having to 
google for a way. I'd from the second failure  because the first might 
be expected due to OTP.

alcohol: Alright. Should we throw an exception then, or just `write()` and return false? Or do you also want to proceed with asking for authentication credentials? Just trying to figure out how the whole flow should progress then :-)

Seldaek: IMO it's ok to write() the hint and then proceed with the rest as it is 
now, eventually it'll fail hard but at least they see the info every time.

alcohol: How about this?

src/Composer/Util/GitHub.php
```diff
@@ -76,124 +76,128 @@ public function authorizeOAuth($originUrl)
...
 
-                // build up OAuth app name
-                $appName = 'Composer';
-                if ($this->config->get('github-expose-hostname') === true && 0 === $this->process->execute('hostname', $output)) {
-                    $appName .= ' on ' . trim($output);
-                } else {
-                    $appName .= ' [' . date('YmdHis') . ']';
-                }
+                    $this->io->write('You can also manually create a personal token at github.com/settings/applications,');
+                    $this->io->write('and then add it using "composer config github-oauth.github.com <token>"');
```
alcohol: How about this?

src/Composer/Util/GitHub.php
```diff
@@ -76,124 +76,128 @@ public function authorizeOAuth($originUrl)
...
+                    }
 
-                // build up OAuth app name
-                $appName = 'Composer';
-                if ($this->config->get('github-expose-hostname') === true && 0 === $this->process->execute('hostname', $output)) {
-                    $appName .= ' on ' . trim($output);
-                } else {
-                    $appName .= ' [' . date('YmdHis') . ']';
-                }
+                    $this->io->write('You can also manually create a personal token at github.com/settings/applications,');
```
cs278: Please add the `https://` so it will be hyperlinked by terminals. Possibly also drop the Oxford comma just to reduce the change of an invalid URL being opened.

alcohol:
Rephrased it a bit and squashed all the commits into one.

Seldaek:
Alright here goes nothing

ptoomey3:
> I couldn't think of any solution other than simply creating a new token whenever one is not available in the auth configuration.

@alcohol - The [new API](https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization) changes also introduce a `fingerprint` parameter that can be used to distinguish between authorizations. The `fingerprint` value is not user visible, so you can use whatever metadata would be useful to distinguish between multiple unique installs of an application for a user. For example, you could use something like `"Composer:{SHA256($users_mac_address)}"` and use that for the `fingerprint`. When Composer hits the Authorizations API it can take a [look through all of the authorizations](https://developer.github.com/v3/oauth_authorizations/#list-your-authorizations) and see if any of them match the `fingerprint` for that machine. If it does match you can likely safely [delete that authorization](https://developer.github.com/v3/oauth_authorizations/#delete-an-authorization) and [recreate it](https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization) with the same `fingerprint`.  That should solve the issue of requiring users to manually manage tokens.

P.S. Since you are using personal access tokens you will still need to generate a unique `description` for each token.

alcohol:
I didn't think of deleting tokens based on fingerprint. I can look into
that when I get back in two weeks.

src/Composer/Util/GitHub.php
```diff
@@ -76,124 +76,128 @@ public function authorizeOAuth($originUrl)
...
-                                    $this->io->write('You have been sent an SMS message with an authentication code to verify your identity.');
-                                }
+        $headers = array('Content-Type: application/json');
+        if ($otp) {
+            $headers[] = 'X-GitHub-OTP: ' . $otp;
+        }
 
-                                $otp = $this->io->ask('Authentication Code: ');
+        $note = 'Composer';
+        if ($this->config->get('github-expose-hostname') === true && 0 === $this->process->execute('hostname', $output)) {
```
Danack: $output appears to be an undefined variable here.

alcohol: Maybe have a look at the function that is being called...

stof: @Danack this is a reference argument here.

Seldaek:
@alcohol did you ever get a chance to check the fingerprint stuff above btw?

alcohol:
I only had a look at the documentation. It could help keep the amount of authorizations created by composer to a minimum. As I understand it though, the fingerprint stuff is only available when using the mirage-preview media type of the API at the moment. Do we want to wait till it becomes part of the stable API, or just go ahead and implement it and send the custom headers to enable it?

Danack:
For the record, I was looking at the code as part of investigating the new api calls, when I was confused by the reference.

Basically, it looks like neither the [Get-or-create an authorization for a specific app](https://developer.github.com/v3/oauth_authorizations/#get-or-create-an-authorization-for-a-specific-app) or [Get-or-create an authorization for a specific app and fingerprint](https://developer.github.com/v3/oauth_authorizations/#get-or-create-an-authorization-for-a-specific-app) API calls can be used by Composer.

Both of them require the client_secret parameter to be set, which the listAuthorizations call did not require previously. So unless composer was to distribute a client secret (which sounds like a bad idea) it won't be possible to recover a previously generated token.

This doesn't seem to be by accident. The whole point of the changes to the API was to make retrieval of previously generated tokens by more secure, by making it so that any application (or anonymous application) is incapable of retrieving the tokens for another application. 

I don't think there's going to be any easy way round this, and people will need to live with Composer generating more tokens that it used to, when for whatever reason the tokens are lost from their machine.

btw the changes are meant to be coming into effect at some point in the next 24 hours, so it's probably a bit late to do anything. Also it might be an idea to put a notice up in the readme or on packagist saying people needing to upgrade Composer. This PR wasn't listed prominently in the release notes, and for people who upgrade their software slowly, it might be a surprise that they need to.

alcohol:
Since we have the username and password at the time of checking, I think we can delete tokens. We don't use the specific app calls by the way. We use the normal create authorization call without a client id/secret.

ptoomey3:
Yeah, if you use an appropriate fingerprint you should be able to delete old tokens with a relatively high degree of confidence. Without a fingerprint it will be impossible to differentiate tokens between multiple installations for the same user. As a result, deleting any given token would risk deleting the wrong one.

Issue #3612, opened 2015-01-05
Downloader emitts a Backspace-character even when `--no-ansi` supplied.


staabm:
Shouldn't the downloader only emit this `BS` chars, when ansi is allowed?

![composer](https://cloud.githubusercontent.com/assets/120441/5612550/2c39aa8e-94d7-11e4-9298-ac0575c6764f.png)

PS: these `BS` chars are only visible with my editor, those are not rendered like that. Its a regular BS char.

stof:
hmm, maybe the ``--no-progress`` option should be impplied when using ``--no-ansi``. What do you think @Seldaek ?

staabm:
Is a non cli environment relevant for composer? If so, the progress indication should be aware of cli vs. non-cli environments - at least to not print the `BS` based progress-chars in non-cli sapis.

Seldaek:
I don't think SAPI based is good because even in a web context you might want fully formatted output for some reasons (there are js renderers for ansi codes etc).

Anyway I'll push a fix shortly.

staabm:
fix confirmed, thanks @Seldaek !

vlakoff:
With these changes, overwrites are disabled on Windows CLI… I tried `--ansi` but it produces garbage.

Although not a big deal, could it be corrected?

Seldaek:
@vlakoff to get full ANSI support on windows I'd recommend you use [ConEmu](https://conemu.codeplex.com/) as terminal or alternatively you can install [ansicon](https://github.com/adoxa/ansicon) which will add ANSI support to the cmd.exe or other terminals.

vlakoff:
Thank you for the tips :) Though, the less I have to install/store/maintain/memorize, the better I feel :blush:

I simply mentioned the issue because Windows CLI was working fine previously. But this "shell" is so shitty it doesn't deserve much care ;)

Seldaek:
Well I do care about windows as I use it myself but it makes things more 
complex for "nothing" if we need to have different flags for colors and 
overwrite. If you develop on windows IMO having ansicon or a proper 
terminal is a must.

dzuelke:
Wait, so c58b7d9 fixes this on its own, correct? Because the side effect of e0d36e1 is that it also disables  progress for piped output, see #3715.

Question is if ANSI == decoration and piping != decoration (done through `posix_isatty` on the stream) should really be treated the same?

And why not just use `\r` instead of backspaces; does that not do the trick?

dzuelke:
[...]

PR #3715, opened 2015-02-06
stderr output using writeError(), closes #1905

### LINKED ISSUE #1905: Error output printed to STDOUT instead of STDERR
I'm writing an installer that runs several Composer commands. I want to suppress informational messages, such as "composer.json has been updated", but not error messages.

Currently, Composer prints everything to STDOUT, so the following will suppress everything:

```
composer update > /dev/null
```

Same problem with the `--quiet` flag.

alcohol:
Since it was a low hanging fruit, voila. Some grunt work has been done.

/cc @dzuelke I was told by @Seldaek you might have some good insights into what should or should not go to stderr.

What qualifies as stderr I interpreted solely on feedback from @Seldaek and gut feeling so far. Any input is welcome.

Closes #1905
Also closes heroku/heroku-buildpack-php#100

### DIFF
--- a/src/Composer/Command/ArchiveCommand.php
+++ b/src/Composer/Command/ArchiveCommand.php
@@ -97,7 +97,7 @@ EOT
             $package = $this->getComposer()->getPackage();
         }
 
-        $io->write('<info>Creating the archive.</info>');
+        $io->writeError('<info>Creating the archive.</info>');
         $archiveManager->archive($package, $format, $dest);
 
         return 0;
@@ -105,14 +105,14 @@ EOT
 
     protected function selectPackage(IOInterface $io, $packageName, $version = null)
     {
-        $io->write('<info>Searching for the specified package.</info>');
+        $io->writeError('<info>Searching for the specified package.</info>');
 
         if ($composer = $this->getComposer(false)) {
             $localRepo = $composer->getRepositoryManager()->getLocalRepository();
             $repos = new CompositeRepository(array_merge(array($localRepo), $composer->getRepositoryManager()->getRepositories()));
         } else {
             $defaultRepos = Factory::createDefaultRepositories($this->getIO());
-            $io->write('No composer.json found in the current directory, searching packages from ' . implode(', ', array_keys($defaultRepos)));
+            $io->writeError('No composer.json found in the current directory, searching packages from ' . implode(', ', array_keys($defaultRepos)));
             $repos = new CompositeRepository($defaultRepos);
         }
 
@@ -125,14 +125,14 @@ EOT
 
         if (count($packages) > 1) {
             $package = reset($packages);
-            $io->write('<info>Found multiple matches, selected '.$package->getPrettyString().'.</info>');
-            $io->write('Alternatives were '.implode(', ', array_map(function ($p) { return $p->getPrettyString(); }, $packages)).'.');
-            $io->write('<comment>Please use a more specific constraint to pick a different package.</comment>');
+            $io->writeError('<info>Found multiple matches, selected '.$package->getPrettyString().'.</info>');
+            $io->writeError('Alternatives were '.implode(', ', array_map(function ($p) { return $p->getPrettyString(); }, $packages)).'.');
+            $io->writeError('<comment>Please use a more specific constraint to pick a different package.</comment>');
         } elseif ($packages) {
             $package = reset($packages);
-            $io->write('<info>Found an exact match '.$package->getPrettyString().'.</info>');
+            $io->writeError('<info>Found an exact match '.$package->getPrettyString().'.</info>');
         } else {
-            $io->write('<error>Could not find a package matching '.$packageName.'.</error>');
+            $io->writeError('<error>Could not find a package matching '.$packageName.'.</error>');
 
             return false;
         }
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -108,6 +108,11 @@ class ConsoleIO extends BaseIO
         $this->doWrite($messages, $newline, true);
     }
 
+    /**
+     * @param array $messages
+     * @param boolean $newline
+     * @param boolean $stderr
+     */
     private function doWrite($messages, $newline, $stderr)
     {
         if (null !== $this->startTime) {
@@ -133,12 +138,38 @@ class ConsoleIO extends BaseIO
      */
     public function overwrite($messages, $newline = true, $size = null)
     {
-        if (!$this->output->isDecorated()) {
+        $this->doOverwrite($messages, $newline, $size, false);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function overwriteError($messages, $newline = true, $size = null)
+    {
+        $this->doOverwrite($messages, $newline, $size, true);
+    }
+
+    /**
+     * @param array $messages
+     * @param boolean $newline
+     * @param integer $size
+     * @param boolean $stderr
+     */
+    private function doOverwrite($messages, $newline, $size, $stderr)
+    {
+        if (true === $stderr && $this->output instanceof ConsoleOutputInterface) {
+            $output = $this->output->getErrorOutput();
+        } else {
+            $output = $this->output;
+        }
+
+        if (!$output->isDecorated()) {
             if (!$messages) {
                 return;
             }
 
-            return $this->write($messages, count($messages) === 1 || $newline);
+            $this->doWrite($messages, count($messages) === 1 || $newline, $stderr);
+            return;
         }
 
         // messages can be an array, let's convert it to string anyway
@@ -147,24 +178,24 @@ class ConsoleIO extends BaseIO
         // since overwrite is supposed to overwrite last message...
         if (!isset($size)) {
             // removing possible formatting of lastMessage with strip_tags
-            $size = strlen(strip_tags($this->lastMessage));
+            $size = strlen(strip_tags($stderr ? $this->lastMessageErr : $this->lastMessage));
         }
         // ...let's fill its length with backspaces
-        $this->write(str_repeat("\x08", $size), false);
+        $this->doWrite(str_repeat("\x08", $size), false, $stderr);
 
         // write the new message
-        $this->write($messages, false);
+        $this->doWrite($messages, false, $stderr);
 
         $fill = $size - strlen(strip_tags($messages));
         if ($fill > 0) {
             // whitespace whatever has left
-            $this->write(str_repeat(' ', $fill), false);
+            $this->doWrite(str_repeat(' ', $fill), false, $stderr);
             // move the cursor back
-            $this->write(str_repeat("\x08", $fill), false);
+            $this->doWrite(str_repeat("\x08", $fill), false, $stderr);
         }
 
         if ($newline) {
-            $this->write('');
+            $this->doWrite('', true, $stderr);
         }
         $this->lastMessage = $messages;
     }
@@ -174,7 +205,16 @@ class ConsoleIO extends BaseIO
      */
     public function ask($question, $default = null)
     {
-        return $this->helperSet->get('dialog')->ask($this->output, $question, $default);
+        $output = $this->output;
+
+        if ($output instanceof ConsoleOutputInterface) {
+            $output = $output->getErrorOutput();
+        }
+
+        /** @var \Symfony\Component\Console\Helper\DialogHelper $dialog */
+        $dialog = $this->helperSet->get('dialog');
+
+        return $dialog->ask($output, $question, $default);
     }
 
     /**
@@ -182,7 +222,16 @@ class ConsoleIO extends BaseIO
      */
     public function askConfirmation($question, $default = true)
     {
-        return $this->helperSet->get('dialog')->askConfirmation($this->output, $question, $default);
+        $output = $this->output;
+
+        if ($output instanceof ConsoleOutputInterface) {
+            $output = $output->getErrorOutput();
+        }
+
+        /** @var \Symfony\Component\Console\Helper\DialogHelper $dialog */
+        $dialog = $this->helperSet->get('dialog');
+
+        return $dialog->askConfirmation($output, $question, $default);
     }
 
     /**
@@ -190,7 +239,16 @@ class ConsoleIO extends BaseIO
      */
     public function askAndValidate($question, $validator, $attempts = false, $default = null)
     {
-        return $this->helperSet->get('dialog')->askAndValidate($this->output, $question, $validator, $attempts, $default);
+        $output = $this->output;
+
+        if ($output instanceof ConsoleOutputInterface) {
+            $output = $output->getErrorOutput();
+        }
+
+        /** @var \Symfony\Component\Console\Helper\DialogHelper $dialog */
+        $dialog = $this->helperSet->get('dialog');
+
+        return $dialog->askAndValidate($output, $question, $validator, $attempts, $default);
     }
 
     /**
@@ -204,9 +262,9 @@ class ConsoleIO extends BaseIO
 
             // use bash if it's present
             if ($finder->find('bash') && $finder->find('stty')) {
-                $this->write($question, false);
+                $this->writeError($question, false);
                 $value = rtrim(shell_exec('bash -c "stty -echo; read -n0 discard; read -r mypassword; stty echo; echo $mypassword"'));
-                $this->write('');
+                $this->writeError('');
 
                 return $value;
             }
@@ -230,9 +288,9 @@ class ConsoleIO extends BaseIO
                 $exe = $tmpExe;
             }
 
-            $this->write($question, false);
+            $this->writeError($question, false);
             $value = rtrim(shell_exec($exe));
-            $this->write('');
+            $this->writeError('');
 
             // clean up
             if (isset($tmpExe)) {
@@ -252,11 +310,11 @@ class ConsoleIO extends BaseIO
                 }
             }
             if (isset($shell)) {
-                $this->write($question, false);
+                $this->writeError($question, false);
                 $readCmd = ($shell === 'csh') ? 'set mypassword = $<' : 'read -r mypassword';
                 $command = sprintf("/usr/bin/env %s -c 'stty -echo; %s; stty echo; echo \$mypassword'", $shell, $readCmd);
                 $value = rtrim(shell_exec($command));
-                $this->write('');
+                $this->writeError('');
 
                 return $value;
             }
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -13,6 +13,7 @@
 namespace Composer\IO;
 
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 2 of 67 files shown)


src/Composer/IO/ConsoleIO.php
```diff
@@ -93,7 +95,7 @@ public function isDebug()
     /**
      * {@inheritDoc}
      */
-    public function write($messages, $newline = true)
+    public function write($messages, $newline = true, $stderr = false)
```
Seldaek: I think it'd be better to move the write content into a private doWrite or such that is called by write/writeError instead of extending this method's signature as it's part of the public interface.

dzuelke:
Will take a look ASAP; thanks!

src/Composer/Command/ArchiveCommand.php
```diff
@@ -132,7 +132,7 @@ protected function selectPackage(IOInterface $io, $packageName, $version = null)
             $package = reset($packages);
             $io->write('<info>Found an exact match '.$package->getPrettyString().'.</info>');
```
dzuelke: This should also be writeError() I guess

alcohol: Ah, alright. Yeah I'll give it a second go-over with the more strict guidelines you mentioned in #1905 (I admit I did not completely read all of them before).

alcohol:
What to do with the Commands that use the Symfony ConsoleOutput? Something like the following?

```
$output->writeln($messages)
```

becomes

```
$output->getErrorOutput()->writeln($messages)
```

Seldaek:
Yup that sounds fine.

alcohol:
Should we then hint `ConsoleOutputInterface`? Or have a wrapper that does an instanceof check and delegates accordingly?

Seldaek:
Hinting sounds fine to me, if it causes issues with anyone we can just 
migrate to using the composer IO as output instead.

alcohol:
The typehint breaks a lot of tests. Not sure if I should clean up the tests, or change the approach. Feedback?

Seldaek:
Yeah I guess this has to be reverted.. I think it should work if you just replace `$output->writeln` by `$this->getIO()->write[Error]`

dzuelke:
I need some more time to review in detail, but here is a preliminary list that I think should output to `stdout`. Everything not on this list, including warnings, notices about outdated versions in the commands below etc should go to `stderr`:

* `composer config` when getting a setting or when using `--list`
* `composer depends` probably without the headers which should go to `stderr` (might need a more "condensed" more for grepping)
* `composer licenses` (might need a more "condensed" more for grepping)
* `composer search`
* `composer show` (could maybe default to `-N` when `stdout` is not a TTY but likely shouldn't)
* `composer status`

A few good rules for figuring out if something should go to `stdout` or `stderr` is:

* if everything is fine, don't output anything to `stdout` unless that's the point of the command (e.g. `composer show`)
* any `composer` command, in particular if it returns data, should be able to be in a pipeline like `composer config --list | grep foo`, and a sudden "this version of Composer is over 30 days old" should not break this pipeline, so it has to go to `stderr`.

In general, Composer will need quite a bit more Unix love. Run any command through `| more` and see how the color codes mess stuff up; I guess `--no-color` should be the default if stderr is not a TTY. Or run a command that outputs to `stderr`, exits with code 0, and then is treated as failed (I think) because it output to `stderr`, which really isn't an error condition.

dzuelke:
Also, `composer --version` should echo the version string to `stdout`, but no warnings. `composer --help` and any usage info should always go to `stderr`.

alcohol:
Any idea how to explain this?

```
$ bin/composer status
No local changes
```
^ the response is in green (color codes)

```
$ bin/composer status | more
ESC[32mNo local changesESC[39m
```

but...

```
$ bin/composer status 2>/dev/null | more
No local changes
```

Seldaek:
BTW "No local changes" should be stderr as far as I understand.

As for how color usage is determined by default, AFAIK it is done here:

https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Console/Output/StreamOutput.php#L101

It seems this is always true though here:

```
$ php -r 'var_dump(posix_isatty("php://stdout"));' 2>/dev/null | more
bool(true)
$ php -r 'var_dump(posix_isatty("php://stdout"));' | more
bool(true)
$ php -r 'var_dump(posix_isatty("php://stdout"));'
bool(true)
```

Not sure what's up on yours.

alcohol:
@dzuelke this should solve some color issues as well: symfony/symfony#13661

dzuelke:
Correct, "no local changes" from `composer status` should go to `stderr`; if there are changes listed, then those should to to `stdout` I guess.

Keep up the great work on this, @alcohol!

alcohol:
@dzuelke One thing I've come across frequently while doing some reading myself is the following, "stderr is also the place where you print stuff like help docs when an invalid argument is passed to a command" with the exception being "when asking explicitly for help docs they should be printed to stdout". Personally I also feel that makes sense. Do you agree?

On a side note, do you use IRC by chance? If so, drop me a line sometime on freenode (ethanol).

dzuelke:
Yes, I'd also print to stderr if the help screen for any command was explicitly requested. Seems most Unix-y programs do that.

alcohol:
But then wouldn't that stop you from doing for example `command --help | grep X` ?

alcohol:
If you do a checkout and run `grep -irPC 3 'io->write\(|getio\(\)->write\(' src` you should get a picture of what is left going to stdout.

I'm not sure on the following cases:
* AboutCommand - expected output, so stdout?
* DiagnoseCommand - mostly (even warnings and errors) expected output, so stdout?
* InitCommand - not sure about this.. lot of user interaction stuff, not likely to be used in a piped process
* Util/[Git|Svn] - the prompt for authentication, followed by user input, should that be stderr or not?

As for help going to stderr, I'd either have to overload the default command from SF2, or submit a PR and cross fingers that they accept it (which might not be the case).

dzuelke:
Sorry I meant stdout of course for explicit help!

dzuelke:
> - AboutCommand - expected output, so stdout?

:+1: 

> - DiagnoseCommand - mostly (even warnings and errors) expected output, so stdout?

Good question. I was going to say that e.g. the xdebug warning should go to stderr, but then the output of the command is mostly useless because there is no context.

> - InitCommand - not sure about this.. lot of user interaction stuff, not likely to be used in a piped process

Well that command is completely quiet (like it should be) when running using `-n`, so the prompts (and any associated status messages) should go to `stderr` I think.

> - Util/[Git|Svn] - the prompt for authentication, followed by user input, should that be stderr or not?

Absolutely!

alcohol:
Aside from InitCommand, pretty much everything should be in order now. I'm personally conflicted about InitCommand. I feel that the interactive session with the user should mostly belong on `stdout`. But if you guys prefer `stderr`, I'll replace it. Let me know.

/cc @dzuelke @Seldaek for feedback

Seldaek:
Alright @dzuelke if you can review one last time I'll merge end of the week I guess, shouldn't let this linger as it needs to be kept up to date otherwise.

dzuelke:
It needs to be `stderr`, @alcohol. Take curl as an example, where if you use a user on the command line, it'll prompt for a password via `stderr` unless you provide it.

The init command in the future could spit out a summary for subsequent processes to use, and various degrees of prompt text mixed into it depending on what flags you used would make piping impossible.

dzuelke:
Will do @seldaek. ConFoo is in full swing so give me a few days.

alcohol:
Alright @dzuelke, I'll get on it today. For the actual input prompts, we use the dialog helpers and such from the console component which take an OutputInterface as one of their arguments. Should we pass stderr there as well then (if available)?

alcohol:
So I implemented the last few changes. 

Right now I'm looking into 
* [x] having the `ask()` and related methods output to stderr if available
* [ ] moving dialog calls within `Command`s to `ConsoleIO` methods if applicable/possible
* [ ] switching to the [QuestionHelper](http://symfony.com/doc/current/components/console/helpers/questionhelper.html) which replaces the deprecated [DialogHelper](http://symfony.com/doc/current/components/console/helpers/dialoghelper.html).

alcohol:
I think I'll leave those last two for a next PR, otherwise this becomes too large.

Seldaek:
Alright merged and we'll see if any tweaking is needed in the future. Thanks a bunch @alcohol

josegonzalez:
I'm wayyy late to the game (seeing as this has been merged), but shouldn't status reports depend upon the verbosity of the command? If I run with no verbosity and everything is going well, why am I getting output written to stderr that isn't an error?

stof:
@josegonzalez writing to stderr does not mean that the command resulted in an error. The only thing which can indicate it is the exit code

josegonzalez:
I'm not saying it's resulted in error, I'm saying that the status output shouldn't be there if the command is running fine *and* I haven't specified a verbosity level. While I realize it's the unix way to write status to stderr, it's also nice to respect different logging levels - if anything, this should only be available if I set `logging => debug` in composer somehow.

Seldaek:
Not sure what status you mean but most of them do depend on the verbosity. And there is always -q although that's a bit carpet-bomb style :)

josegonzalez:
As used [here](https://github.com/cakephp/bake/issues/51), while we should definitely check the status code, it's common to assume that any output to stderr is error output and thus users using composer in tooling may be confused as to whether their command is failing or not. I know it's a bad example because of the message itself, but would it be possible to instead move to proper log levels?

The original issue pointed out the following:

> I'm writing an installer that runs several Composer commands. I want to suppress informational messages, such as "composer.json has been updated", but not error messages.

I think the pr went the opposite way, and just moved informational messages to error messages.

stof:
@josegonzalez assuming this means you are making non-standard assumption, because there is no convention about the way to use output to report failures. Detecting failures is based on the exit code only in CLI conventions (it is even something being common to both Unix and Windows systems).

stderr vs stdout is **not** about error report vs non error report (btw, I don't think that shell scripting calls them stdout and stderr at all). They are about message being part of the command output (i.e. something you would pipe to the next command in the chain) vs message not being part of it (informational messages for the user, and error reports are a specific kind of informational messages: they inform of an error)

dzuelke:
> As used here, while we should definitely check the status code, it's common to assume that any output to stderr is error output and thus users using composer in tooling may be confused as to whether their command is failing or not. I know it's a bad example because of the message itself, but would it be possible to instead move to proper log levels?

That's a wrong assumption. Try a `curl https://google.com/ > google.txt`. The standard error stream is not just for errors. Please use the exit code of a process to determine its success or failure. These are standard Unix conventions, and mixing diagnostics/progress/informational output with the "real" output of a program breaks the whole concept of Unix pipelines, unless you like spending your evenings messing around with `awk` and `sed` scripts to parse output from badly written applications.

josegonzalez:
I'm not saying that it's the right way to do this, and the example I give is a bad one. This is probably a philosophical issue, but here is a better example.

I'm the maintainer of multiple vagrant installs at my company, and we manage things through chef. I have it setup so that if chef executes a command which has error output, that is always output in red. Very useful for debugging, and you or I can easily see that the message above is not an error but status output. I would see that error and silence stderr because it's not actually an error, and is going to be confusing to users.

Now imagine I give vm access to someone else on the team, a frontend dev, who has no idea how any of the tooling below vagrant works. All they know is about npm and gulp. They bring up the vm, there is an issue with composer failing, but no error output because I silenced status output earlier. They come to me with their issue and now we have to turn on stderr, which happens to also output debug output.

What I'm asking for is a way to still have this output behind a `log-level` flag. You can still write it to stderr if you'd like, it's just not as useful when it's a status message.

While the above vagrant thing happens all the time, from time to time I get an email from my boss - who is by no means non-technical - as to why our deploy is erroring. Turns out npm likes to output status messages to stderr as well, resulting in hundreds of lines like this:

faux errors

It's not an error, npm is running fine, it's just letting us know that certain packages we depend on could use cleanup. And rather than place that behind a flag that I can control - we write deploy logs to files and selectively write to stdout depending upon verbosity of the deploy - it decided to just write to stderr. The deploy didn't fail - because of course i checked the exit code of the command - but it seems like there was an error to the end-user.

I guess what I'm saying is that writing these status messages to stderr by default without some notion of log-levels just isn't friendly for developers.

dzuelke:
Then use the `-q` flag, @josegonzalez.

josegonzalez:
Would that silence actual error output?

Seldaek:
Yup -q basically kills all output so that's a bit too much. I agree we 
could use a -Q or so that quiets all but errors, but then we need to be 
able to flag error output (vs stderr output, which isn't necessarily 
errors).

josegonzalez:
I'm not familiar with the logger you are using in composer, but if it supports the standard log-levels, would you all be opposed to me adding a `--log-level` argument which sorts the messages into different levels?

stof:
@josegonzalez the console output is not using a logger

josegonzalez:
Would you be opposed to a change that uses a logger (probably monolog)?

alcohol:
I don't feel a logger would be the right answer here, as tempting as it
might be. Instead, making more use of the various verbosity levels would
benefit composer more. During my scavenger hunt through the code I found
very little use of these, and more often than not it just checks for
verbose or it doesn't check anything at all, even though there are three
levels available. If anything, I would invest time in improving that. Just
my 2 cents :-)

Seldaek:
Yeah I think what we need is a way to indicate the minimum verbosity level in the write/writeError call, so we can make more use of it without annoying ifs everywhere. That wouldn't solve the whole issue though, we still need a new VERBOSITY_ERROR level IMO that's used for actual error messages. It would be the same as verbosity normal (i.e. show always by default), but with this -Q flag or so we could suppress NORMAL but not ERROR, so we get quiet-except-errors or full-shut-up mode :)

alcohol:
Would it be an option to consider any stderr output without a verbosity
flag to always be an error as opposed to stderr output with a verbosity
flag being informational?

src/Composer/IO/ConsoleIO.php
```diff
@@ -102,6 +122,13 @@ public function write($messages, $newline = true)
                 return sprintf('[%.1fMB/%.2fs] %s', $memoryUsage, $timeSpent, $message);
             }, (array) $messages);
         }
+
+        if (true === $stderr && $this->output instanceof ConsoleOutputInterface) {
```
dzuelke: Why are you taking this shortcut here @alcohol? It prevents self-updating progress, for instance, with `composer install 2>&1 | somepipethatdecoratesoutput`; you get many newlines instead (331425b and #3818 are symptoms of that).

Is there a particular reason why you don't want any overwriting behavior when piping? IMO it should still behave that way. Can/should I fix?

dzuelke: Just for the sake of testing, I removed the entire `if` block and it seems to all still work fine for a `composer install` (and now the progress works again as expected when piping), but I'm sure there's some other command where things would be broken now, or else you wouldn't have added that code? :) please advise.

alcohol: Not sure I follow? Doesn't self updating progress use overwrite? This if block is to differentiate between stdout and stderr? I'm missing a link in the context I think... been gone for two weeks.

alcohol: doWrite() is only used internally (private method). write() and writeError() pass the $stderr boolean. I merely added an extra check to make sure we do have an output interface that supports stderr. Not sure what you mean by shortcut here?

dzuelke:
Run this to reproduce:

    COMPOSER_CACHE_DIR=/dev/null composer install 2>&1 | tee /dev/null

The download progress will be written in many new lines instead of in one. The reason is this `if`, because it skips the entire overwriting stuff using backspaces.

src/Composer/IO/ConsoleIO.php
dzuelke: @alcohol run this to reproduce:

COMPOSER_CACHE_DIR=/dev/null composer install 2>&1 | tee /dev/null
The download progress will be written in many new lines instead of in one. The reason is this if, because it skips the entire overwriting stuff using backspaces.

alcohol: While I am able to reproduce the new lines, I really don't see how this if statement is related. Could it be you are commenting on the wrong line?

The calls go as:
https://github.com/composer/composer/blob/master/src/Composer/Util/RemoteFilesystem.php#L161
https://github.com/composer/composer/blob/master/src/Composer/Util/RemoteFilesystem.php#L332 (repeated)
https://github.com/composer/composer/blob/master/src/Composer/Util/RemoteFilesystem.php#L231

It seems more likely to be related to this line https://github.com/composer/composer/pull/3715/files#diff-bd1b00b10d2c20047a8732546579ce78R166

Before my change, overwrite only ever checked if stdout was decorated. Now it checks if stderr is decorated instead (if trying to overwrite stderr).

alcohol: I'm sorry, I'm going to sleep now. Have a flight of nearly 12 hours behind me and need to get back to work tomorrow morning. Will pick this up again then if not resolved/clear yet.

dzuelke: Yes, sorry, @alcohol, I mean lines 166+; picked the wrong spot in the code by accident, sorry!

src/Composer/IO/ConsoleIO.php
```diff
@@ -111,12 +138,38 @@ public function write($messages, $newline = true)
...
+     */
+    private function doOverwrite($messages, $newline, $size, $stderr)
+    {
+        if (true === $stderr && $this->output instanceof ConsoleOutputInterface) {
+            $output = $this->output->getErrorOutput();
+        } else {
+            $output = $this->output;
+        }
+
+        if (!$output->isDecorated()) {
```
alcohol: So this seems to be the culprit. But I'm not sure why the decorated check would cause different behaviour now. I was under the impression that this value for stderr is equal to that of stdout.

This check **did** exist before my changes, so hence why I kept it. Not sure I understand the purpose of it yet though. Will investigate further.

dzuelke: My guess would be that `$output->isDecorated()` behaves differently now for stdout and stderr?

dzuelke: Behavior is correct; the branch is executed if you pipe `stderr` using either `2>&1` or a separate `2> >(...)` command.

That means that this must have been broken before but it now works correctly. Apparently, before your changes, `if (!$this->output->isDecorated()) {` never evaluated to `true`, even if the output of e.g. a `composer install` was piped somewhere (because Composer never did that back then).

dzuelke: But I can't tell why. Weird weird weird.

dzuelke: Okay, figured it out. It's simply coincidence, as @Seldaek disabled progress for `--no-ansi` using a bit of a "carpet bomb" approach for #3612 (specifically, in e0d36e1, although I think c58b7d9 should be enough to address that issue).

Question is if ANSI == decoration and piping != decoration (done through `posix_isatty` on the stream) should really be treated the same?

alcohol: I'm fine with removing this check. I'm not entirely certain what the impact would be on Windows users though. Nor am I in a position to test this.

dzuelke: Alrighty, all sorted via #3612's reverts :)

PR #3724, opened 2015-02-10
Update 00-intro.md adding an install tutorial link for shared hosting

HongPong:
adding an install tutorial link for shared hosting finding the correct environment settings.

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -91,6 +91,11 @@ option and providing a target directory (it can be an absolute or relative path)
 curl -sS https://getcomposer.org/installer | php -- --install-dir=bin
 ```
 
+On shared hosting, the correct PHP CLI interpreter may not be the default php.
+For example, on a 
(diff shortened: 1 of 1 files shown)


alcohol:
Please no. If you're going to include something like this, make it generic and don't include commercial brands or businesses.

HongPong:
Makes sense, but it would be best if specific vendors install tutorials can
be listed on a separate page then. There is nothing for the shared install
edge cases in the current docs and it is difficult to get the software
working without searching elsewhere.

alcohol:
To be honest, the link you provided has little to do with composer, and much more with a hosting provider's php setup (also, holy moly 5.2). I believe composer should stay away from trying to document the entire landscape and keep the focus on managing dependencies.

Seldaek:
Sorry but hosts have to stop misconfiguring their shit. It makes no sense to have a php on the CLI that calls in a non-CLI binary. If they provide ssh access they should support it and configure the PATH correctly.

PR #3725, opened 2015-02-10
composer-schema strict check for minimum-stability, closes #3723

### LINKED ISSUE #3723: Set json with empty values as invalid
Unintentionally, I leave the "minimum-stability" blank so for all commands I've tried composer shown the following message:

screen shot 2015-02-09 at 8 24 22 pm

The JsonParser finds this json is valid, but it isn't in composer world.

What do you guys think is better to solve this?
I guess setting up some properties as required must solve this problem in an elegant way.

Here is my composer.json:

```json
{
    "name": "yourwebmaker/test",
    "description": "nono no no",
    "minimum-stability": "",
    "license": "proprietary",
    "authors": [
        {
            "name": "Daniel Lima",
            "email": "yourwebmaker@gmail.com"
        }
    ],
    "require": {}
}
```

alcohol:
Not sure if the testcase is in the right place, but it seemed fitting when I wrote it.

Closes #3723

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -284,7 +284,8 @@
         },
         "minimum-stability": {
             "type": ["string"],
-            "description": "The minimum stability the packages must have to be install-able. Possible values are: dev, alpha, beta, RC, stable."
+            "description": "The minimum stability the packages must have to be install-able. Possible values are: dev, alpha, beta, RC, stable.",
+            "pattern": "^dev|alpha|beta|rc|RC|stable$"
         },
         "prefer-stable": {
             "type": ["boolean"],
--- a/tests/Composer/Test/Json/ComposerSchemaTest.php
+++ b/tests/Composer/Test/Json/ComposerSchemaTest.php
@@ -1,0 +1,90 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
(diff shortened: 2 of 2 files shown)


stof:
why not using an ``enum`` field type instead in the schema ?

alcohol:
I tried that too. Same effect, except the error message becomes N+3 lines long, where N is the amount of values allowed, as it basically includes a `print_r()` of the allowed values arary. The regex is just a short oneliner.

stof:
Well, this may warrant a PR to the schema validator to make the error message nicer (maybe json-encoding the array of allowed values instead, as done in Symfony in several places). But using an enum would be more correct (and would help if someone uses a UI to edit the composer.json file based on its schema, as such tool already exists btw)

alcohol:
You mean a PR that modifies https://github.com/justinrainbow/json-schema/blob/master/src/JsonSchema/Constraints/Enum.php#L36 ?

stof:
yes. print_r is quite bad for building readable error messages

alcohol:
Well, I doubt a PR to [json-schema](https://github.com/justinrainbow/json-schema/) will get merged any time soon considering the activity levels there. Do you want me to just change it to enum and adjust the test suite according to how json-schema behaves now?

PR #3726, opened 2015-02-10
fix bug in GitDriver::supports for remote repo

davidverholen:
for some reason it does not work (in packagist) without the $output param. I don't get any error message here, maybe someone has an idea, why?

Anyway, need this ;)

### DIFF
--- a/src/Composer/Repository/Vcs/GitDriver.php
+++ b/src/Composer/Repository/Vcs/GitDriver.php
@@ -242,7 +242,7 @@ class GitDriver extends VcsDriver
         }
 
         $process = new ProcessExecutor($io);
-        if($process->execute('git ls-remote --heads ' . ProcessExecutor::escape($url)) === 0) {
+        if($process->execute('git ls-remote --heads ' . ProcessExecutor::escape($url), $output) === 0) {
             return true;
         }
 


alcohol:
Doesn't work is not a helpful statement. Was there an error? Unexpected results? Does it sit on the couch all day eating all your cheetos and ignoring the classifieds?

Details please :-)

davidverholen:
yes, I'm using it with packagist.

It just says "We're so sorry, something is wrong on our end." ;)

no log entries.

And since the Submit Package action is called by an ajax request its not quite easy to debug.
I also tried a try/catch block and log the result, but nothing.

I assume It's raising a php error, most likely here:
https://github.com/composer/composer/blob/master/src/Composer/Util/ProcessExecutor.php#L61

Maybe some Ideas?

alcohol:
Rather than assumptions, do you have any actual errors to share? A way to reproduce the problem? Anything tangible?

davidverholen:
as I said, no.

you can try to reproduce it setting up a packagist and adding a remote git repository (not github). It fails without the $output param but atm I'm not able to see any errors. (Except the very helpful message from packagist, that something went wrong)

Currently I'm trying to reproduce the error in a smaller, more controllable and debug-able environment, but still without success.

Seldaek:
I think the problem is that without $output it defaults to echoing out the result, and that probably breaks the json response when you do a force update. I guess the update goes through though.. Anyway outputing this does not make sense so let's merge. Thanks

Issue #3720, opened 2015-02-09
Added dependencies will be ignored by install command


bestform:
To reproduce:

1. install composer dependencies creating a composer.lock file
2. add a new dependency to composer.json
3. run composer install again

Expected:
Error indicating that the new dependency isn't available

Actual:
"Nothing to install or update"

Note:
Composer does complain about changed version constraints of already installed packages, that cannot be met by the versions saved in the composer.lock file.

I've looked into this issue and found out, that composer does indeed complain about the newly added dependency, but only if you specify an extra/branch-alias in the composer.json. This will result in the `RootPackage` being a `RootAliasPackage` instead of a `RootPackage` - which seems like a bug to me, as `RootPackageLoader::Load()` should return `$realPackage` instead of `$package` in https://github.com/composer/composer/blob/master/src/Composer/Package/Loader/RootPackageLoader.php#L109

If you patch this, the branch-alias will no longer cause composer install to complain about the newly added dependency - which in turn is not what I would expect from it at all - but at least the error is now happening constantly, no matter whether a branch-alias is being configured or not.

I've not yet had the time to look into it more deeply, but I guess, composer should complain about the new dependency like it does about changed and not met version constraints.

Seldaek:
require/branch-alias probably should be part of the lock file to prevent such issues, so that an install from lock file always succeeds no matter if you changed your json. On the other hand it's sort of a misuse and it's caught already by the warning about the lock file being out of date with composer.json, so that should indicate you must run an update instead of an install

bestform:
I see that the warning should be enough, but then again, composer does complain about changed version constraints. So whatever it does to warn the user about inconsistencies, it should handle it at least consistently - so, if you show an error should the version constraints change, an additional dependency should trigger an error, too. Currently it is just being ignored.

PR #3736, opened 2015-02-12
Ignore composer.json if installing from lock file

bestform:
This is my proposal to this issue: https://github.com/composer/composer/issues/3720

I've added a flag to the composer factory which allows to ignore all requirements from composer.json, should a lock file be present. This flag will be set by the installer to avoid evaluating the composer.json file and complaining about not met requirements if it just should install from the lock file. The warning, that the json file is not in sync with the lock file will still be shown.

### DIFF
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -79,7 +79,7 @@ EOT
             $output->writeln('<warning>You are using the deprecated option "dev". Dev packages are installed by default now.</warning>');
         }
 
-        $composer = $this->getComposer(true, $input->getOption('no-plugins'));
+        $composer = $this->getComposer(true, $input->getOption('no-plugins'), true);
         $composer->getDownloadManager()->setOutputProgress(!$input->getOption('no-progress'));
         $io = $this->getIO();
 
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -222,7 +222,7 @@ class Factory
             $localConfig = $file->read();
         }
 
-        if($ignoreRootReqsIfLockIsPresent && isset($composerFile) && file_exists($this->getLockFilename($composerFile))){
+        if ($ignoreRootReqsIfLockIsPresent && isset($composerFile) && file_exists($this->getLockFilename($composerFile))) {
             // ignore all requirements from the original composer.json file in case we are installing from a lock file.
             $localConfig["require"] = array();
             $localConfig["require-dev"] = array();
@@ -498,9 +498,8 @@ class Factory
      */
     protected function getLockFilename($composerFile)
     {
-        $lockFile = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
+        return "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
             ? substr($composerFile, 0, -4) . 'lock'
             : $composerFile . '.lock';
-        return $lockFile;
     }
 }
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -244,6 +244,7 @@ class Factory
                 $config->setAuthConfigSource(new JsonConfigSource($localAuthFile, true));
             }
         }
+
         $vendorDir = $config->get('vendor-dir');
         $binDir = $config->get('bin-dir');
 
@@ -362,7 +363,7 @@ class Factory
 
         $composer = null;
         try {
-            $composer = self::createComposer($io, $config->get('home') . '/composer.json', $disablePlugins, $config->get('home'), false, false);
+            $composer = self::createComposer($io, $config->get('home') . '/composer.json', $disablePlugins, $config->get('home'), false);
         } catch (\Exception $e) {
             if ($io->isDebug()) {
                 $io->write('Failed to initialize global composer: '.$e->getMessage());
@@ -496,7 +497,7 @@ class Factory
      * @param $composerFile
      * @return string
      */
-    protected function getLockFilename($composerFile)
+    private function getLockFilename($composerFile)
     {
         return "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
             ? substr($composerFile, 0, -4) . 'lock'
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -195,7 +195,7 @@ class Factory
      * @throws \UnexpectedValueException
      * @return Composer
      */
-    public function createComposer(IOInterface $io, $localConfig = null, $disablePlugins = false, $cwd = null, $fullLoad = true)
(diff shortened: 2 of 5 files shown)


src/Composer/Factory.php
```diff
@@ -482,10 +485,22 @@ protected function purgePackages(WritableRepositoryInterface $repo, Installer\In
...
+        return $factory->createComposer($io, $config, $disablePlugins, null, true, $ignoreRootReqsIfLockIsPresent);
+    }
+
+    /**
+     * @param $composerFile
+     * @return string
+     */
+    protected function getLockFilename($composerFile)
+    {
+        $lockFile = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
```
alcohol: Why not just return right away? Don't need a variable in the function context anymore.

src/Composer/Factory.php
```diff
@@ -222,6 +222,12 @@ public function createComposer(IOInterface $io, $localConfig = null, $disablePlu
             $localConfig = $file->read();
         }
 
+        if($ignoreRootReqsIfLockIsPresent && isset($composerFile) && file_exists($this->getLockFilename($composerFile))){
```
alcohol: PSR-2 `if (`

src/Composer/Command/InstallCommand.php
```diff
@@ -79,7 +79,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $output->writeln('<warning>You are using the deprecated option "dev". Dev packages are installed by default now.</warning>');
         }
 
-        $composer = $this->getComposer(true, $input->getOption('no-plugins'));
+        $composer = $this->getComposer(true, $input->getOption('no-plugins'), true);
```
alcohol: This should not be a hardcoded value. Add an inputOption please, and default to current behaviour (false).

src/Composer/Factory.php
bestform: True, will change.

src/Composer/Factory.php
bestform: Good find. Will change. too

src/Composer/Command/InstallCommand.php
bestform: I am not entirely sure about this. If i understand @Seldaek correctly, the new behavior is what really should have been done in the first place. But I see your point.

alcohol: No I misunderstood when I read your issue. `true` does seem the right value for the install command. Even though it does change behaviour, I think generally speaking it would be an improvement, so that's probably fine.

src/Composer/Factory.php
```diff
@@ -482,10 +485,21 @@ protected function purgePackages(WritableRepositoryInterface $repo, Installer\In
...
 
-        return $factory->createComposer($io, $config, $disablePlugins);
+        return $factory->createComposer($io, $config, $disablePlugins, null, true, $ignoreRootReqsIfLockIsPresent);
+    }
+
+    /**
+     * @param $composerFile
+     * @return string
+     */
+    protected function getLockFilename($composerFile)
```
stof: why creating this protected method ?

src/Composer/Factory.php
```diff
@@ -238,7 +244,6 @@ public function createComposer(IOInterface $io, $localConfig = null, $disablePlu
                 $config->setAuthConfigSource(new JsonConfigSource($localAuthFile, true));
             }
         }
-
```
stof: thi empty line should be preserved

src/Composer/Factory.php
```diff
@@ -359,7 +362,7 @@ protected function createGlobalComposer(IOInterface $io, Config $config, $disabl
 
         $composer = null;
         try {
-            $composer = self::createComposer($io, $config->get('home') . '/composer.json', $disablePlugins, $config->get('home'), false);
+            $composer = self::createComposer($io, $config->get('home') . '/composer.json', $disablePlugins, $config->get('home'), false, false);
```
stof: useless given it is the default already

stof:
A functional test must be added to avoid regressions

src/Composer/Factory.php
bestform: This of course should be private. It is an extra method, as we need to get the name later on again. So I moved the code outside to be able to reuse it. We could of course inline it and save the result for later use, but I don't really like the idea of generating data at the beginning of a function and preserving it without using it until the end.

bestform:
@stof: Thanks for the input. Concerning the functional test: There isn't a test that tests the current behavior. So this should be added first I guess?

stof:
@bestform you are adding a new feature (ability to install from lock even when you add new requirements). This must be covered by tests to ensure that we don't break this feature in the future

stof:
FYI, installer tests are defined in https://github.com/composer/composer/tree/master/tests/Composer/Test/Fixtures/installer

bestform:
@stof I see what you mean. I thought about it more like fixing a bug, but even then, a test would have been appropriate. I will write a test to reproduce the problem I am trying to fix and come back to you.

bestform:
@stof OK, I've added a test, which reproduces the bug. You can see it here (I haven't pushed the commit yet, as my code doesn't pass either): https://gist.github.com/bestform/ee0f866bbc958d197e4b

My code doesn't pass the test either, because the test suite will work with a mocked file system. My code depends on a file existing (the lock file) to decide, what to do. My code is not the only part, where this is the case. Another example is: https://github.com/composer/composer/blob/master/src/Composer/Factory.php#L309 . This also is code, that'll never be reached by the tests.

So what should we do about that? Is there an alternative way to check for the existence of a lock file?

alcohol:
Is it just me, or does your test not make any sense? Why is foo/baz in there twice?

bestform:
@alcohol it is there in two versions. I want to check, that composer does only upgrade foo/baz when installing, but not foo/bar. So both versions are installed in version 1.0.0, the lock file says to update baz to 2.0.0, the composer json wants to upgrade bar and baz. If my patch works, composer should only update baz.

bestform:
@alcohol ah, sorry, I didn't recognize that you edited your reply. I've added the test to my branch.

Seldaek:
TBH I am not so keen on the implementation here, it propagates stuff everywhere, and I believe this could instead be fixed very locally in the Installer class by just ignoring the root package requirements there or removing them if that's needed for things to work.

bestform:
@Seldaek I am not so keen on this solution, either, but I couldn't find a way to ignore the root requirements in the installer itself. I would absolutely prefer to do so instead of telling the factory what to ignore, but this isn''t possible at the moment, or is it? Am I missing something?

bestform:
@Seldaek OK, i might have found a way. Just a second. I'll add another PR for it.

bestform:
@Seldaek @alcohol @stof Thank you all for the healthy discussion about this topic. I've implemented an alternative, way less intrusive way here: https://github.com/composer/composer/pull/3740

It would be great if you could have a look at it. I am going to close this one for now.

PR #3740, opened 2015-02-13
Ignore root requirements in installer when installing from locked state

bestform:
Concerning https://github.com/composer/composer/issues/3720

Original PR was https://github.com/composer/composer/pull/3736

After a healthy discussion in the original PR this is my alternative proposal, that is way less intrusive than the original solution.
Now the installer itself removes the requirements from the root package, should it install from a locked state.


Seldaek:
Looking better, thanks!

PR #3746, opened 2015-02-16
Requiring conflicting, replaced packages

bestform:
If you require a package, that replaces another package in a specific version, an installation from the json file will install a conflicting version of the replaced package alongside the replacing package.
An install attempt from the resulting lock file will then fail as it recognizes the conflict.

This problem became apparent after this PR: https://github.com/composer/composer/pull/3740, as an install request with an existing lock file will now ignore the json completely. Before the patch it seems to have ignored the conflict even when installing from lock.

This PR currently just contains two tests, that reproduce the problem. No fix has been implemented, yet.

As requested by @Seldaek I hereby /cc @naderman for this PR.

### DIFF
--- a/tests/Composer/Test/Fixtures/installer/replaced-packages-should-not-be-installed-when-installing-from-lock.test
+++ b/tests/Composer/Test/Fixtures/installer/replaced-packages-should-not-be-installed-when-installing-from-lock.test
@@ -1,0 +1,39 @@
+--TEST--
+Requiring a replaced package in a version, that is not provided by the replacing package, should result in a conflict, when installing from lock
+--COMPOSER--
+{
+    "repositories": [
+        {
+            "type": "package",
+            "package": [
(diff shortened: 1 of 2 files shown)


bestform:
@alcohol just made the problem pretty clear on IRC, so i'll add his quote for further clarification:

<ethanol> if package A replaces package B, and you require both package A and B, a conflict should arise, no matter what versions you have them restrained to. because you simply should not be able to install the same package twice (even if different versions)
<ethanol> currently, a conflict does not arise if there is no lock file
<ethanol> a conflict does arise if there is a lock file
<ethanol> we're trying to figure out why a conflict does not arise when running install without a .lock file

bestform:
Thanks for the merge. Just as a reference, this issue is a follow up for this topic: https://github.com/composer/composer/issues/3754

PR #3747, opened 2015-02-16
Added missing $learnedWhy property

vkartaviy:
Missing property triggers ```PHP Notice: Undefined property: Composer\DependencyResolver\Solver::$learnedWhy``` which throws ```ErrorException```


fprochazka:
:clap:

Seldaek:
Thanks

PR #3749, opened 2015-02-16
composer config repositories.packagist false

### LINKED ISSUE #3713: Command to set packagist repository to false
Hi, I am able to add new repositories from the CLI without a problem, but the syntax to disable Packagist doesn't seem to be supported.

Can a command like:
```
composer config repositories.packagist false
```
be added?

### LINKED ISSUE #3728: Feature Request: "init --plain" that doesn't use the internet for anything (like git init)
I just recently started a composer project just so I could get access to an autoloader. I didn't define *any* dependencies, but still attempted to contact packagist anyway. The problem is, that packagist was having some serious performance problems, so what should have been an instantaneous setup of the composer vendor and autoloader, took several minutes.

I want the ability to setup a blank composer project like I can with `git init` - simply include the vendor/composer directory from local cache, as well as the autoloader. Then even if I don't have internet access (say, on a plane), I can easily get the autoloader and a blank project which I can then modify later.

alcohol:
closes #3728, closes #3713, closes #3290, not sure about #3025

I specifically left the original exception message as-is in order to not expose this behaviour deliberately.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -373,8 +373,9 @@ EOT
                 ));
             }
 
-            if (1 === count($values) && $booleanValidator($values[0])) {
-                if (false === $booleanNormalizer($values[0])) {
+            if (1 === count($values)) {
+                $bool = strtolower($values[0]);
+                if (true === $booleanValidator($bool) && false === $booleanNormalizer($bool)) {
                     return $this->configSource->addRepository($matches[1], false);
                 }
             }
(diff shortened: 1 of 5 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -402,6 +360,54 @@ function ($vals) {
...
+
+            if (2 === count($values)) {
+                return $this->configSource->addRepository($matches[1], array(
+                    'type' => $values[0],
+                    'url'  => $values[1],
+                ));
+            }
+
+            if (1 === count($values) && $booleanValidator($values[0])) {
+                if (false === $booleanNormalizer($values[0])) {
```
stloyd: This will never be reached as if it's `false` it will not enter the `if` above.

alcohol: I think you're not reading it right then.

Seldaek:
Thanks

PR #3750, opened 2015-02-16
Config Git repository user

DavidPrevot:
One can’t assume a Git user is already setup for the tests.

### DIFF
--- a/tests/Composer/Test/Package/Archiver/ArchivableFilesFinderTest.php
+++ b/tests/Composer/Test/Package/Archiver/ArchivableFilesFinderTest.php
@@ -187,6 +187,8 @@ class ArchivableFilesFinderTest extends \PHPUnit_Framework_TestCase
         $this->finder = new ArchivableFilesFinder($this->sources, array());
 
         $this->assertArchivableFiles($this->getArchivedFiles('git init && '.
+            'git config user.email "you@example.com" && '.
+            'git config user.name "Your Name" && '.
             'git add .git* && '.
             'git commit -m "ignore rules" && '.
             'git add . && '.
--- a/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
+++ b/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
@@ -78,6 +78,18 @@ class ArchiveManagerTest extends ArchiverTest
             throw new \RuntimeException('Could not init: '.$this->process->getErrorOutput());
         }
 
+        $result = $this->process->execute('git config user.email "you@example.com"', $output, $this->testDir);
+        if ($result > 0) {
+            chdir($currentWorkDir);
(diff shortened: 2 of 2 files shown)


alcohol:
How hard is it to get indentation right?

DavidPrevot:
> How hard is it to get indentation right?

Apologies, should be fixed now.

alcohol:
Doesn't Git just default to the local user though? Any context for this "fix"? What was broken? What errors were being thrown? I don't recall these tests causing any issues.

DavidPrevot:
Hi,

Le 17/02/2015 04:31, Rob a écrit :
> Doesn't Git just default to the local user though?

Yet, the local user name should have been defined first, it does output
an error message on the first commit anyway.

> Any context for this "fix"?

The goal is to be able to run the testsuite in a pristine machine (a
throwable chroot). I’m in the process of introducing Composer into the
Debian archive. Without this patch, the package fails to build from
source (there are other Debian specific patches, but they are less
likely to be interesting for upstream).

https://ftp-master.debian.org/new/composer_1.0.0~alpha9+dfsg-1.html

> What was broken?

Well, the testsuite (two errors).

> What errors were being thrown?

There were 2 errors:

1) Composer\Test\Package\Archiver\ArchivableFilesFinderTest::testGitExcludes
ErrorException:
unlink(/tmp/composer_archiver_test121910414154e349ba323f41.37738342/archive.zip):
No such file or directory

/tmp/buildd/composer-1.0.0~alpha9+dfsg/tests/Composer/Test/Package/Archiver/ArchivableFilesFinderTest.php:268
/tmp/buildd/composer-1.0.0~alpha9+dfsg/tests/Composer/Test/Package/Archiver/ArchivableFilesFinderTest.php:195

2) Composer\Test\Package\Archiver\ArchiveManagerTest::testArchiveTar
RuntimeException: Could not commit:
*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name (for <pbuilder@mikado.tilapin.org>) not allowed


/tmp/buildd/composer-1.0.0~alpha9+dfsg/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php:90
/tmp/buildd/composer-1.0.0~alpha9+dfsg/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php:43

FAILURES!
Tests: 1235, Assertions: 1950, Errors: 2.

> I don't recall these tests causing any issues.

You probably never ran the testsuite on a minimal chroot then. That’s
also probably the reason why travis has been explicitly set up in
https://github.com/composer/composer/commit/6f61d95829502d1558db57b59ca488aa2b5f9b29
(that commit can be reverted if this pull request is accepted by the way).

Regards

David

alcohol:
That would explain it indeed. Alright, thanks for the PR!

Seldaek:
Thanks

PR #3758, opened 2015-02-18
Conform with Symfony Coding Standards

qw3rtman:
According to Symfony Coding Standards, a single space must be added around binary operators excluding the concatenation operator.

"Add a single space around binary operators (==, &&, ...), with the exception of the concatenation (.) operator" (http://symfony.com/doc/current/contributing/code/standards.html)

### DIFF
--- a/src/Composer/Console/HtmlOutputFormatter.php
+++ b/src/Composer/Console/HtmlOutputFormatter.php
@@ -83,6 +83,6 @@ class HtmlOutputFormatter extends OutputFormatter
             }
         }
 
-        return $out . '">'.$matches[2].'</span>';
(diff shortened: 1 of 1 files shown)


GromNaN:
This CS error is everywhere in the code. Fixing it would require editing almost all the files. 
https://github.com/GromNaN/composer/commit/3818aa61621eb248461aa4c4375c72c43d7225c7

I think fixing CS of code already merged is pointless as long as we don't have a bot to check every pull request (like http://fabbot.io/ scanning symfony).

Seldaek:
Yeah this is kind of a useless nitpick IMO.. `the exception of the concatenation` I understand it as meaning for concat it's up to the circumstances..

PR #3759, opened 2015-02-19
symfony/console version fix in composer.json

rafalgalka:
According to [documentation](https://getcomposer.org/doc/01-basic-usage.md#package-versions) ~2.3 means >=2.3.0 < 3.0.0. This causes a problem for version >= 2.5

`Fatal error: Uncaught exception 'ErrorException' with message "Symfony\Component\Console\Helper\DialogHelper" is deprecated since version 2.5 and will be removed in 3.0. Use "Symfony\Component\Console\Helper\QuestionHelper" instead. in [...]/vendor/symfony/console/Symfony/Component/Console/Helper/DialogHelper.php:34`

This deprecation error is converted to Exception by `Composer\Util\ErrorHandler::handle`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php": ">=5.3.2",
         "justinrainbow/json-schema": "~1.3",
         "seld/jsonlint": "~1.0",
-        "symfony/console": "~2.3",
+        "symfony/console": "~2.3.0",
         "symfony/finder": "~2.2",
         "symfony/process": "~2.1"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "2bc9cc8aa706b68d611d7058e4eb8de7",
+    "hash": "4537c5ae28210483397d6cbe0f33378a",
     "packages": [
         {
             "name": "justinrainbow/json-schema",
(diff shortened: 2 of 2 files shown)


alcohol:
That's probably because you are enforcing E_STRICT in your php configuration. We are aware of the deprecations and they will be replaced eventually. This is no reason to lower the version requirements.

rafalgalka:
We have this issue when using "composer/composer" as a dependency: https://github.com/pimcore-extensions/manager/blob/master/composer.json
Composer is called via [\Composer\Console\Application](https://github.com/pimcore-extensions/manager/blob/master/lib/ExtensionManager/Composer.php#L48)

alcohol:
Is it viable for you to just set `error_reporting(E_ALL ^ E_STRICT)` for now?

```
$original = error_reporting(E_ALL ^ E_STRICT);
/* do composer stuff */
error_reporting($original);
```

Also, this particular issue you have was introduced in [2.7.x](https://github.com/symfony/Console/blob/2.7/Helper/DialogHelper.php#L34).

Composer currently ships with 2.6.x in the .lock file. It's likely your project is overruling that. Another solution would be to this component to your project's root composer.json with a stricter version requirement (one that does not allow 2.7.x for example).

Basically, this is something you can solve within your project. Composer is not affected by this :-)

rafalgalka:
I can't agree because situation is more complex:
1. [pimcore-extensions/manager](https://github.com/pimcore-extensions/manager) is a plugin for [pimcore/pimcore](https://github.com/pimcore/pimcore)
2. final developer is creating project using composer.json [similar to this](https://github.com/pimcore/pimcore/blob/master/composer.json) adding: `pimcore-extensions/manager` as dependency

So our composer.json is not authority in dependency solving. To solve this we should force on final developer to add `symfony/console: ~2.3.0` to his project's composer - which is hard to explain and maintain.

alcohol:
That is simply not true. Your composer.json **is** just as much an authority.

If anything, your PR would only be valid until #3760 is taken care of. 
Aside from that, `~2.3.0` is far too restrictive. It should be `~2.3,<2.7`.

rafalgalka:
I was trying adding this to our composer.json and it didn't worked. Maybe I missed something.

IMHO such powerful component like composer should give good example in dependency management which is intended to help maintain high stability. Allowing huge range of versions for dependencies is not a good idea at all.

alcohol:
Most likely you missed something. As long as your dependency graph has at least one component that does not allow 2.7.x, it should install the highest 2.6.x instead.

As for allowing a "huge" range of versions, I beg to differ. The range allowed falls within backwards compatibility, which is the primary concern. Also, the dependencies of Composer itself are currently locked to a version that does not implicitly trigger an error regarding deprecations. If you choose to run with a higher version, then that is entirely your own choice. As I explained before, you can add a constraint yourself to limit this.

alcohol:
If you show me your composer.json I could help you debug as to why it is not working out for you as expected, perhaps? :-)

rafalgalka:
I'm not a composer master but it looks that composer's lock file is not considered in dependency resolution. I will try this once again and return with reproducible environment if it will fail.

alcohol:
Correct, the composer.lock file is not considered. Only the composer.json is considered from each package. So your plugin should have a require on `symfony/console` with a constraint of `~2.3,<2.7`.

rafalgalka:
OK, I definitely missed something. Thank you for your time, and sorry for the confusion.

alcohol:
No problem at all. Glad we are able to resolve your issue :-)

Issue #3760, opened 2015-02-19
Replace deprecated TableHelper and DialogHelper before updating Composer dependencies


alcohol:
Components affected:
- [TableHelper](http://symfony.com/doc/current/components/console/helpers/tablehelper.html)
- [DialogHelper](http://symfony.com/doc/current/components/console/helpers/dialoghelper.html)

Replacements:
- [Table](http://symfony.com/doc/current/components/console/helpers/table.html)
- [QuestionHelper](http://symfony.com/doc/current/components/console/helpers/questionhelper.html)

These are deprecated as of 2.5.x. Composer is currently using 2.6.x. 
As of 2.7.x these will actually trigger errors.
- See TableHelper [here](https://github.com/symfony/Console/blob/2.7/Helper/TableHelper.php#L40)
- See DialogHelper [here](https://github.com/symfony/Console/blob/2.7/Helper/DialogHelper.php#L34)

The TableHelper is already replaced in the following PR's: #3593 and #3715
The QuestionHelper has no open PR's

stof:
the issue with these 2 PRs is that they are bumping the Console requirement to 2.5+ (without doing it properly), which would prevent Packagist from depending on Composer.

alcohol:
Good catch. I can remove the specific commit from my own PR (#3715) and leave a comment on #3593 regarding this, or create a new PR from my commit and include an updated composer.json requirement. What would you suggest?

stof:
@alcohol given that your own PR is about a totally different change, it should not include the replacement anyway (mixing different changes in the same PR is making the review harder)

alcohol:
Yeah, it's just something I ran into while updating all the files. I explicitly committed it separately so it would be easy to remove. I'll go do that right now.

Seldaek:
@alcohol if you have time to tackle the DialogHelper > QuestionHelper part of this that'd be great.

alcohol:
Yeah I got time. I think I had some questions regarding it though, but I can't recall at the moment. They will probably resurface as I dive into it again.

Issue #3754, opened 2015-02-17
replace behaviour


alcohol:
Discussion: Is the following desired behaviour? Should this conflict or not?
Priority: low

While resolving, `replace` doesn't seem to be taken into account, or at least not in a manner which one might expect. For example, I expect the following to either conflict or not install the separate component. But composer does so happily.

In other words, replace allows you to install the same library in different versions potentially.

```
$ composer require symfony/symfony:2.4.10 symfony/http-foundation:2.6.4
./composer.json has been created
Loading composer repositories with package information
Updating dependencies (including require-dev)
  - Installing psr/log (1.0.0)
    Loading from cache

  - Installing symfony/http-foundation (v2.6.4)
    Downloading: 100%

  - Installing twig/twig (v1.18.0)
    Loading from cache

  - Installing doctrine/lexer (v1.0.1)
    Loading from cache

  - Installing doctrine/annotations (v1.2.3)
    Loading from cache

  - Installing doctrine/collections (v1.2)
    Loading from cache

  - Installing doctrine/cache (v1.4.0)
    Loading from cache

  - Installing doctrine/inflector (v1.0.1)
    Loading from cache

  - Installing doctrine/common (v2.4.2)
    Loading from cache

  - Installing symfony/icu (v1.2.2)
    Downloading: 100%

  - Installing symfony/symfony (v2.4.10)
    Downloading: 100%

Writing lock file
Generating autoload files
```

Normally composer does not allow you to install different versions of the same library.

/cc @naderman @Seldaek

Seldaek:
BTW the second use case is a bug in require if anything. I mean if you require twice the same in one `require` command it should possibly error, and for sure it should overwrite in the json file and not end up with twice the same. But again that's a misuse thing which isn't so important.

Seldaek:
As for the first use case, I think either we conflict or we leave it as is now, but we can not possibly treat a replace of 2.4.10 as replacing 2.6.4 as if something required that they probably/should have a good reason.

alcohol:
It didn't end up in the composer.json twice. Those are two seperate examples. One where I used the require command, and one where I manually generated a composer.json. Sorry if that wasn't clear enough. I'll modify the post.

bestform:
As discussed on IRC, I strongly think, that should result in a conflict.
Having both versions installed will result in problems, that won't be apparent immediately. The autoloader might or might not chose the correct version for your case.
[...]

PR #3761, opened 2015-02-19
Update tablehelper to table

alcohol:
Refs #3760, pulled out of #3715

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php": ">=5.3.2",
         "justinrainbow/json-schema": "~1.3",
         "seld/jsonlint": "~1.0",
-        "symfony/console": "~2.3",
+        "symfony/console": "~2.5",
         "symfony/finder": "~2.2",
         "symfony/process": "~2.1"
     },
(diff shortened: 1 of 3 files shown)


alcohol:
Meh, silly timeout. The build actually succeeded.

stof:
@Seldaek dropping the support for symfony/console 2.3 will create issues for Packagist, because it still uses Symfony 2.3

alcohol:
It should be backwards compatible though, right?

stof:
@alcohol what do you mean ?

alcohol:
That it will not conflict. Since packagist will install symfony/symfony with Console at 2.3, and then because of the dependency on composer it will also install symfony/console 2.5.

alcohol:
See related #3754

stof:
@alcohol except that this is a bug in the solver (previous versions of the solver used to forbid this. it may be related to optimizations which have been done). If symfony/symfony replaces the symfony/console, installing both together is invalid.

alcohol:
@Seldaek well that answers that mystery ^

Seldaek:
Well either way I need to update packagist :)

PR #3764, opened 2015-02-20
retry slow test if timeout occurs

alcohol:
Hopefully fixes sporadic failure of test suite.

Alternatively, we could replace
```
$proc = new Process($cmd, __DIR__.'/Fixtures/functional');
```
with
```
$proc = new Process($cmd, __DIR__.'/Fixtures/functional', null, null, $timeout);
```
where `$timeout` is something higher than the default of 60.

/cc @Seldaek @stof suggestions / feedback?

### DIFF
--- a/tests/Composer/Test/AllFunctionalTest.php
+++ b/tests/Composer/Test/AllFunctionalTest.php
@@ -89,7 +89,7 @@ class AllFunctionalTest extends \PHPUnit_Framework_TestCase
         putenv('COMPOSER_HOME='.$this->testDir.'home');
 
         $cmd = 'php '.escapeshellarg(self::$pharPath).' --no-ansi '.$testData['RUN'];
-        $proc = new Process($cmd, __DIR__.'/Fixtures/functional');
+        $proc = new Process($cmd, __DIR__.'/Fixtures/functional', null, null, 300);
         $exitcode = $proc->run();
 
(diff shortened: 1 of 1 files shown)


alcohol:
Might have to take into account cleanup or something. Not sure.

GromNaN:
Do you have any idea what causes these timeouts?

Seldaek:
IMO it's just due to network slowness or such that's a bit random..

Seldaek:
BTW @alcohol the travis build failed because create-project needs an empty dir to work with and tearDown is not executed in between your retries.. Not sure what's best between calling tearDown/setUp in the loop or just extending the timeout instead of a retry loop.

alcohol:
@Seldaek yeah I was worried that would be the case. I can call `tearDown()` and `setUp()` manually but it seems a bit excessive perhaps. On the other hand, extending the timeout might not solve the problem. Not sure if the download is just really slow or if it is hanging on something, but in the latter case a longer timeout still would not help. Let me know how to proceed and I'll get right on it tomorrow.

Seldaek:
I'd say let's extend timeout to 5min and see how it goes.. if it doesn't 
fix it we can reconvene :)

alcohol:
Raising the timeout it is. :-) 
/done

GromNaN:
:clap:

PR #3775, opened 2015-02-25
VCS section: add requirements for the branch name

rantonmattei:
If you want to use a custom branch as a requirement in your `composer.json`, that branch must be prefixed with "dev-" or you'll get an error:
```sh
[UnexpectedValueException]
  Could not parse version constraint some-branch: Invalid version string "some-branch".
```

It took me a while to find that info on forums and blogs, so I figured I'd rather update the doc. :grin: 
Let me know.


Seldaek:
Thanks

PR #3784, opened 2015-02-26
Use data provider in RepositoryManager tests

slbmeh:
Updates RepositoryManager tests to use the data provider that was defined and adds an additional repository to assert the exception for unrecognized repository types.


GromNaN:
It's a good idea to add such test cases, but we shouldn't complexify the code that way.

What about creating separate test function and factorize the `RepositoryManager` initialization into a method `createRepositoryManager` ?

slbmeh:
I didn't put much thought into it. I might be needing changes to that exception handling and wanted to see it's coverage. I saw the data provider for the test case and it wasn't doing anything with the provided data.

PR #3788, opened 2015-02-26
Show VCS state when displaying local changes in StatusCommand.

slbmeh:
This adds the concept of current ref vs expected ref to the `VcsDownloader` to address #1968. I've only implemented support in the Git driver so far.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Downloader;
 
+use Composer\Package\Version\VersionParser;
 use Composer\Package\PackageInterface;
 use Composer\Util\Git as GitUtil;
 use Composer\Util\Platform;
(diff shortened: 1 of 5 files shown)


slbmeh:
I don't use svn, perforce, or mercurial so I don't have a project readily available to functionally test changes to the other drivers. So I'm making this PR now and maybe someone can pick up the pieces I'm missing in the future.

Seldaek:
@slbmeh would you mind taking in the Hg and Svn support from https://github.com/composer/composer/pull/3265/files to implement getVersionDifference on those drivers? I think the only other comment I have is that the output in case of differing version should maybe be a bit more clear that the version has changed from X to Y and not just listing the current version like it's an innocuous fact :)

Aside from that, looks good to me and I'd like to get this merged!

slbmeh:
It's been a while since I've looked at this.  Going to rebase on to master and apply the changes you suggested.

slbmeh:
After adding friendlier output that shows version changed from x to y, I feel a VCSVersionParser may be desirable because now the downloader and repository are parsing versions.

slbmeh:
I added support to compare references for mercurial and svn.  I don't have any sample repositories to ensure they work, but it gets a ref as done in #3265.  If the ref does not match it will show that the package changed from the installed version to the ref unless the vcs downloader has a `getCurrentVersion` implementation.  Currently the GitDownloader provides a richer output by parsing the current version that is checked out.

@Seldaek what are your thoughts on the PR now?

Seldaek:
Closing in favor of #5328 which indeed is a much nicer solution.

Issue #2896, opened 2014-04-10
Untracked working tree files blocking updates


pornel:
Composer, when told to throw away changes blocking an update, fails to checkout the update if there is any file or local change preventing git from doing so.

```
[RuntimeException]
  Failed to execute git checkout 'xxx' && git reset --hard 'xxx'

  error: The following untracked working tree files would be overwritten by checkout:
  	some_rubbish_temp_file
  Please move or remove them before you can switch branches.
  Aborting
```

This can be fixed in two ways:

* when stashing, add `--include-untracked` flag which will stash newly added files as well.
* when checking out, add `--force` flag which will overwrite files instead of aborting.

glennmschultz:
god damit stuck by this thing.

mcd-php:
@pornel , @CaptainQuirk I'm not from composer team and so can't talk in their name, but I believe what dirs/files not created by me and not tracked by VCS, be it dependencies, auto-generated code or documentation, should be capable to be completely purged by build automation scripts without any harm for the project. Nothing what you want to keep should be in that folders/files.

Seldaek:
Fixed by #3789

PR #3789, opened 2015-02-26
stash untracked and force checkout if discarding changes

slbmeh:
This addresses #2896


Seldaek:
Thanks, merged as 27d51642c8d4522ad65eef35b5a5b13ecd96b34e (had to rebase it) and I tweaked a bit in https://github.com/composer/composer/commit/690c7f572f5e16d832f15b923bb78960f400b4fe to make sure it works even when the user selects an option and the discard-changes config has no value set.

Issue #2885, opened 2014-04-06
local package --prefer-source  option


alexey-baranov:
Most time interesting install my own packages with --prefer-source option to be able to modify and test them and then update remote repository from main project.

But if I use --prefere-source command line option all other packages install the same way and lose very much time/ 

Is it possible to specify few packages with --prefer-source option? Thanks!

motin:
A workaround to achieve this is to install all packages using `--prefer-dist` first, then remove the directory of a specific few packages from under the vendor directory. Run install again but with `--prefer-source` and only the missing packages will be installed from source.

afraca:
I just bumped into this as well, and barryvdh got me here. (Shame on me for not finding this.... ) 

Any update on this? The reasons are pretty much: dist is nice to save bandwidth and is faster, but I've got some of my own packages where it would be nice to edit them in the /vendor folder and push.

RdeWilde:
Would like this too, can't be difiicult right, because all the functionality already exists.

Seldaek:
Fixed by https://github.com/composer/composer/pull/3790

Issue #2787, opened 2014-03-05
selectively git clone packages


etcook:
I'd love the ability to selectively git clone certain packages. Bonus if you are able to do it with packagist based package definitions as well.

We have a private packagist repo that's setup with both third party as well as internal packages. We'd love the ability to "git clone" the internal packages so that we can commit changes easily.

Seldaek:
Fixed by https://github.com/composer/composer/pull/3790

PR #3790, opened 2015-02-26
add support for package based install type preferences

slbmeh:
addresses #2787 #2885

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -830,9 +830,9 @@ preferred-install Example:
     "config": {
         "preferred-install": {
             "my-organization/stable-package": "dist",
-		    "my-organization/*": "source",
-			"partner-organization/*": "auto",
-			"*": "dist"
+            "my-organization/*": "source",
+            "partner-organization/*": "auto",
+            "*": "dist"
         }
     }
 }
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -743,7 +743,8 @@ The following options are supported:
   will also look for classes in the PHP include path.
 * **preferred-install:** Defaults to `auto` and can be any of `source`, `dist` or
   `auto`. This option allows you to set the install method Composer will prefer to
-  use.
+  use. Can optionally be a hash of patterns for more granular install preferences.
+  See example below.
 * **store-auths:** What to do after prompting for authentication, one of:
   `true` (always store), `false` (do not store) and `"prompt"` (ask every
   time), defaults to `"prompt"`.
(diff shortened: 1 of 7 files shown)


slbmeh:
I guess I need to update the tests. Will do if the PR looks good otherwise.

slbmeh:
And while I'm at it, update the docs.

splenetic:
:+1: This would be really useful
By the way, if I specify --prefer-source on the cli, will it override the setting in composer.json?

slbmeh:
@splenetic i believe i left the passed arguments `--prefer-source` and `--prefer-dist` as defacto overrides for every package.

src/Composer/Config.php
```diff
@@ -108,6 +108,20 @@ public function merge($config)
...
                 if (in_array($key, array('github-oauth', 'http-basic')) && isset($this->config[$key])) {
                     $this->config[$key] = array_merge($this->config[$key], $val);
+                } elseif ('preferred-install' === $key && isset($this->config[$key])) {
+                    if (is_string($val)) {
+                        $val = array('*' => $val);
+                    }
+                    if (is_string($this->config[$key])) {
+                        $this->config[$key] = array('*' => $this->config[$key]);
+                    }
+                    $this->config[$key] = array_merge($this->config[$key], $val);
```
Seldaek: I would reverse the two arguments here so that latter configs win over the earlier one. As we load the global config first the local composer.json rules need to come first to override global.

Seldaek:
@slbmeh looks good to me, and indeed updating docs and tests would be great :)

src/Composer/Config.php
slbmeh: wouldn't that give the existing values precedence? this method doc says that it merges configuration overriding the existing. the above for the auth places the merging of existing config first

slbmeh:
That broken test was a backwards compatibility breakage I overlooked. I addressed that so it works 100% as it used to with the new options available. I may have gone overboard with the tests, they could probably be simplified with a data provider. I think they verbosely express the way it is intended to work with all paths covered.

Seldaek:
@slbmeh thanks for adding docs and all, looking good now but thanks to long delays on my end it's now conflicting :/ Would you mind rebasing it and ping me when it's done then I try to merge it ASAP

doc/04-schema.md
```diff
@@ -743,7 +743,8 @@ The following options are supported:
   will also look for classes in the PHP include path.
 * **preferred-install:** Defaults to `auto` and can be any of `source`, `dist` or
   `auto`. This option allows you to set the install method Composer will prefer to
-  use.
+  use. Can optionally be a hash of patterns for more granular install preferences.
+  See example below.
```
alcohol: I would suggest you make the example a small chapter or paragraph with a heading that you can link to.

Seldaek:
Thanks, I ended up doing the rebase myself :)

PR #3794, opened 2015-02-27
Require only once `files` (defined in the schema)

Hywan:
Imagine the following scenario:

  * 2 projects: A and B.
  * Both have a `autoload.files` directive in there respective `composer.json` files: A.a and B.b.
  * A.a loads B.b,
  * B loads B.b.

Composer will load files in this order: A.a and B.b. But because A.a already did the load of B.b, when Composer does that, there is an error (“<entity> already exists” for instance).

While, in appareance, it seems a bug that A.a loads B.b, according to the context, it makes sense. For instance, we install dependencies of a project with Composer but we don't use its autoloader. In this case, we need a specific autoloader, the one from A for instance.

From the point of view of A, it is difficult to avoid the load of B.b because Composer did not load it yet.

So this patch does one simple think: It `require_once` the `autoload.files` instead of `require`.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -573,7 +573,7 @@ METHOD_FOOTER;
 
 function composerRequire$suffix(\$file)
 {
-    require \$file;
+    require_once \$file;
 }
 
 FOOTER;
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
@@ -51,5 +51,5 @@ class ComposerAutoloaderInitFilesAutoloadOrder
 
 function composerRequireFilesAutoloadOrder($file)
 {
-    require $file;
+    require_once $file;
 }
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_functions.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_functions.php
@@ -51,5 +51,5 @@ class ComposerAutoloaderInitFilesAutoload
 
 function composerRequireFilesAutoload($file)
 {
-    require $file;
+    require_once $file;
 }
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_include_path.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_include_path.php
@@ -67,5 +67,5 @@ class ComposerAutoloaderInitIncludePath
 
 function composerRequireIncludePath($file)
 {
-    require $file;
+    require_once $file;
 }
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_target_dir.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_target_dir.php
@@ -71,5 +71,5 @@ class ComposerAutoloaderInitTargetDir
 
 function composerRequireTargetDir($file)
 {
-    require $file;
+    require_once $file;
 }


stof:
Please do a bench for this change. This is likely to make the autoloader slower because of the extra stat call done for each file loading

Hywan:
@stof And how do I do a bench? Does Composer provide at least one?

stof:
@Hywan use a project using the composer autoloader and profile it with each version of the autoloader. It could be the symfony-standard repo for instance (or any other example project for a different framework), or any other project using the composer autoloader (though using your work project is probably not a good idea as the profile is meant to be shared). Fabien used the Drupal 8 codebase to show the difference between composer with and without classmap for instance.

What is important is the comparison of the profile before and after the change anyway. And if only the composer autoloader code changes between both profiles, other parts of the profiles should not impact the comparison much.

Hywan:
Do these projects use the `autoload.files` directives? If we have a project with, let's say, 100 dependencies, with only 10% of them having an `autoload.files` directories, we have only 10 `require_once`. I don't think we can measure it :-/.

stof:
ah, I missed the fact that the require call inside the ClassLoader has not been changed.

Hywan:
@stof What's next?

stof:
@Hywan waiting for @Seldaek or @naderman to review this. I'm not a collaborator on the repo.

Seldaek:
If A requires B composer should load the files from B first and then A can skip loading since it can check if B's stuff is defined before doing it's fallback-require... To be honest I don't really like doing this change because once we do it we can not go back as people will start depending on it, and require_once is usually a code smell IMO so I'd rather not go there unless we have a really good use case for it.

Hywan:
@Seldaek How to control the order? It is possible with an extension, yup. However, in my example, a user is requiring A and B but he does not necessary know how A and B work. He does not even know that B partly loads A. The result is: A bug in the autoloader, without understanding why.

I ensure you I understand your point of view. However, we don't have a lot of possibilities here:

  1. B stops loading A and we assume Composer's autoloader will be used everywhere, everytime, which is a huge and wrong assumption ;-),
  2. Composer allows to register, and thus, delegate some class loading to, another autoloader.
  3. We use this patch.

3 is a short-term solution. 1 is not a solution to consider at all. 2 is the best solution but more difficult technically, and even before starting to develop it, it is difficult to specify it (on the paper).

Thoughts?

PS: I didn't say it yet, but thank you for this great tool :-).

Seldaek:
@Hywan what I am saying is that if B requires A to be there, you should have a require in B's composer.json towards A. That way composer should load them in the right order (A first, then B), or the other way if that's what is needed. Are the two packages public by any chance? Because all this speculation and A/B talk isn't very helpful :)

Hywan:
I invite @jubianchi, he shares the issue with me.

jubianchi:
@Hywan @Seldaek I don't think I will be of much help here as everything was said.

The problem is the following: we have some project providing their own autoloaders (for use without composer). But this autoloader was also used as an `autloader.files`.

Because of its standalone nature, this autoloader was trying to load some libraries (which require some files to be statically loaded) by itself (again, for use without composer). 

But when using this same autoloader through composer this was causing errors.

@Seldaek here are the repositories where you will find the projetcs/libraries. One thing to note: today, they do not use `autloader.files` anymore due to this behavior. They still provides their autoloader but, when using them through composer, we let composer handle everything.

* https://github.com/atoum/ruler-extension
* https://github.com/hoaproject/Core

Seldaek:
Then it sounds to me like it's fixed no? I don't really see why you'd 
want to define your own autoloader instead of defining autoloader rules 
in composer when you install via composer.

jubianchi:
@Seldaek because atoum and its extensions can also be used in a standalone manner (without composer).

I wanted to have a single entry point for autoloader that's why I built my own and used it through `autoloader.files`. But as you said, it's like it's fixed today but I have two autoloaders.

Seldaek:
Well you don't really have two, you have one and composer has one.. and every user of your libs has less files included on every request so that's a win.

jubianchi:
@Seldaek kind of :)

Our autoloader handles much more: class and namespace aliases. But this is not a big deal. 
It works now for everyone. We just have towrite a clear documentation explaining why we did what we did on the extensions.

BTW, thanks for you help!

Hywan:
:-)

PR #3802, opened 2015-02-28
Allow Composer to use the Opcache function opcache_is_script_cached. 

Danack:
This eliminates almost all file-system access once the cache is warmed. In the case where the file for the class is in the first directory that the Composer autoloader looks this eliminates one stat call which is usually cached by PHP, which is only a tiny benefit.

The real benefit comes when a class is not in the first directory that Composer looks in. In that case there will be file stat calls which will fail and so are not cached by PHP, which results in an actual filesystem hit for each one.

I can make a benchmark if you really want - but the difference is so huge I'm not sure one is actually needed.

btw this change almost certainly removes any benefit from using the dumped/optimized classmap. OpCache has all the information available to it, and C code is far better at searching for strings than PHP code is. That eliminates the relatively large amount of memory used by PHP to load a large array.

## Questions remaining

1) This stuff isn't unit testable. It would be possible to make an integration test that mocks the functions `opcache_is_script_cached` and `file_exists`, but want to hear if that's the way you want it tested before doing do.

2) How should it be turned off/on? Enabling the OpCache optimisation is done by setting a define currently:

```
define('COMPOSER_OPCACHE_OPTIMIZE', true);

require(__DIR__.'/../vendor/autoload.php');
```

I've left it like that for now as it makes testing the performance easier. I can't see any reason why it shouldn't be enabled by default....but there might be. 

btw [opcache_is_script_cached](http://php.net/manual/en/function.opcache-is-script-cached.php) is available since PHP 5.6

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -439,7 +439,11 @@ EOF;
 
 require_once $vendorPathToTargetDirCode . '/autoload_real.php';
 
-return ComposerAutoloaderInit$suffix::getLoader();
+if (defined('COMPOSER_OPCACHE_OPTIMIZE') && COMPOSER_OPCACHE_OPTIMIZE == true) {
+    return ComposerAutoloaderInit$suffix::getLoader(true);
+}
+
+return ComposerAutoloaderInit$suffix::getLoader(false);
 
 AUTOLOAD;
     }
@@ -469,14 +473,14 @@ class ComposerAutoloaderInit$suffix
         }
     }
 
-    public static function getLoader()
+    public static function getLoader(\$enableOpcacheOptimize = false)
     {
         if (null !== self::\$loader) {
             return self::\$loader;
         }
 
         spl_autoload_register(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'), true, $prependAutoloader);
-        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader();
+        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader(\$enableOpcacheOptimize);
         spl_autoload_unregister(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'));
 
 
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -439,11 +439,7 @@ EOF;
 
 require_once $vendorPathToTargetDirCode . '/autoload_real.php';
 
-if (defined('COMPOSER_OPCACHE_OPTIMIZE') && COMPOSER_OPCACHE_OPTIMIZE == true) {
-    return ComposerAutoloaderInit$suffix::getLoader(true);
-}
-
-return ComposerAutoloaderInit$suffix::getLoader(false);
+return ComposerAutoloaderInit$suffix::getLoader(defined('COMPOSER_OPCACHE_OPTIMIZE') && COMPOSER_OPCACHE_OPTIMIZE);
 
 AUTOLOAD;
     }
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -56,8 +56,8 @@ class ClassLoader
 
     private $classMapAuthoritative = false;
 
-    const   SEARCHMODE_OPCACHE = 0;
-    const   SEARCHMODE_FILE = 1;
+    const SEARCHMODE_OPCACHE = 0;
+    const SEARCHMODE_FILE = 1;
     
     public function __construct($enableOpcacheOptimize = false)
     {
@@ -364,7 +364,8 @@ class ClassLoader
         return $file;
     }
 
-    private function findFileWithExtension($class, $ext) {
+    private function findFileWithExtension($class, $ext)
+    {
         foreach ($this->searchModes as $searchMode) {
 
             // PSR-4 lookup
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -378,7 +378,7 @@ class ClassLoader
                         foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
                             $file = $dir.DIRECTORY_SEPARATOR.substr($logicalPathPsr4, $length);
                             if ($searchMode == self::SEARCHMODE_OPCACHE) {
-                                if (opcache_is_script_cached($file) == true) {
+                                if (opcache_is_script_cached($file)) {
                                     return $file;
                                 }
                             }
@@ -396,7 +396,7 @@ class ClassLoader
             foreach ($this->fallbackDirsPsr4 as $dir) {
                 $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr4;
                 if ($searchMode == self::SEARCHMODE_OPCACHE) {
-                    if (opcache_is_script_cached($file) == true) {
+                    if (opcache_is_script_cached($file)) {
                         return $file;
                     }
                 }
@@ -424,7 +424,7 @@ class ClassLoader
                         foreach ($dirs as $dir) {
                             $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr0;
                             if ($searchMode == self::SEARCHMODE_OPCACHE) {
-                                if (opcache_is_script_cached($file) == true) {
+                                if (opcache_is_script_cached($file)) {
                                     return $file;
                                 }
                             }
@@ -442,7 +442,7 @@ class ClassLoader
             foreach ($this->fallbackDirsPsr0 as $dir) {
                 $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr0;
                 if ($searchMode == self::SEARCHMODE_OPCACHE) {
-                    if (opcache_is_script_cached($file) == true) {
+                    if (opcache_is_script_cached($file)) {
                         return $file;
                     }
                 }
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -377,12 +377,12 @@ class ClassLoader
                     if (0 === strpos($class, $prefix)) {
                         foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
                             $file = $dir.DIRECTORY_SEPARATOR.substr($logicalPathPsr4, $length);
-                            if ($searchMode == self::SEARCHMODE_OPCACHE) {
+                            if ($searchMode === self::SEARCHMODE_OPCACHE) {
                                 if (opcache_is_script_cached($file)) {
                                     return $file;
                                 }
                             }
-                            else if ($searchMode == self::SEARCHMODE_FILE) {
+                            else if ($searchMode === self::SEARCHMODE_FILE) {
                                 if (file_exists($file)) {
                                     return $file;
                                 }
@@ -395,12 +395,12 @@ class ClassLoader
             // PSR-4 fallback dirs
             foreach ($this->fallbackDirsPsr4 as $dir) {
                 $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr4;
-                if ($searchMode == self::SEARCHMODE_OPCACHE) {
+                if ($searchMode === self::SEARCHMODE_OPCACHE) {
                     if (opcache_is_script_cached($file)) {
                         return $file;
                     }
                 }
-                else if ($searchMode == self::SEARCHMODE_FILE) {
+                else if ($searchMode === self::SEARCHMODE_FILE) {
                     if (file_exists($file)) {
                         return $file;
                     }
@@ -423,12 +423,12 @@ class ClassLoader
                     if (0 === strpos($class, $prefix)) {
                         foreach ($dirs as $dir) {
                             $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr0;
-                            if ($searchMode == self::SEARCHMODE_OPCACHE) {
+                            if ($searchMode === self::SEARCHMODE_OPCACHE) {
                                 if (opcache_is_script_cached($file)) {
                                     return $file;
                                 }
                             }
-                            else if ($searchMode == self::SEARCHMODE_FILE) {
+                            else if ($searchMode === self::SEARCHMODE_FILE) {
                                 if (file_exists($file)) {
                                     return $file;
                                 }
@@ -441,12 +441,12 @@ class ClassLoader
             // PSR-0 fallback dirs
             foreach ($this->fallbackDirsPsr0 as $dir) {
                 $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr0;
-                if ($searchMode == self::SEARCHMODE_OPCACHE) {
+                if ($searchMode === self::SEARCHMODE_OPCACHE) {
                     if (opcache_is_script_cached($file)) {
                         return $file;
                     }
                 }
-                else if ($searchMode == self::SEARCHMODE_FILE) {
+                else if ($searchMode === self::SEARCHMODE_FILE) {
                     if (file_exists($file)) {
                         return $file;
                     }
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -56,6 +56,29 @@ class ClassLoader
 
     private $classMapAuthoritative = false;
 
+    const   SEARCHMODE_OPCACHE = 0;
+    const   SEARCHMODE_FILE = 1;
+    
+    public function __construct($enableOpcacheOptimize = false)
+    {
+        $this->searchModes = array(
+            self::SEARCHMODE_FILE
+        );
+
+        if ($enableOpcacheOptimize == false) {
+            return;
+        }
+        if (function_exists('opcache_is_script_cached') == false) {
+            //todo - should a warning be generated?
+            return;
+        }
+
+        $this->searchModes = array(
+            self::SEARCHMODE_OPCACHE,
+            self::SEARCHMODE_FILE,
+        );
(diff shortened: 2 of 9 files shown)


Danack:
The test fails because it's hard-coded against what should be generated....it seems ok apart from that, but does still require an integration test.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -56,6 +56,29 @@ class ClassLoader
...
+    {
+        $this->searchModes = array(
+            self::SEARCHMODE_FILE
+        );
+
+        if ($enableOpcacheOptimize == false) {
+            return;
+        }
+        if (function_exists('opcache_is_script_cached') == false) {
+            //todo - should a warning be generated?
```
GromNaN: This function has been introduced un PHP 5.6.

Danack: Which is noted in the PR, and also why the code doesn't call `opcache_is_script_cached` unless it is available as a function.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -56,6 +56,29 @@ class ClassLoader
 
     private $classMapAuthoritative = false;
 
+    const   SEARCHMODE_OPCACHE = 0;
+    const   SEARCHMODE_FILE = 1;
```
stof: constants should be before properties, and you have extra spaces between the keyword and the name

src/Composer/Autoload/ClassLoader.php
```diff
@@ -341,63 +364,98 @@ public function findFile($class)
...
-        $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR) . $ext;
-
-        $first = $class[0];
-        if (isset($this->prefixLengthsPsr4[$first])) {
-            foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
-                if (0 === strpos($class, $prefix)) {
-                    foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
-                        if (file_exists($file = $dir . DIRECTORY_SEPARATOR . substr($logicalPathPsr4, $length))) {
-                            return $file;
+    private function findFileWithExtension($class, $ext) {
```
stof: please don't alter coding standards

stof:
I think it would be great to show the impact of this change. A great way for this would be to use [Blackfire](http://blackfire.io/) to profile several cases: existing code with the classmap optimization, new cod with and without the classmap optimization, and sharing the profile comparisons for them.

src/Composer/Autoload/ClassLoader.php
Danack: Fixed.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -439,7 +439,11 @@ protected function getAutoloadFile($vendorPathToTargetDirCode, $suffix)
 
 require_once $vendorPathToTargetDirCode . '/autoload_real.php';
 
-return ComposerAutoloaderInit$suffix::getLoader();
+if (defined('COMPOSER_OPCACHE_OPTIMIZE') && COMPOSER_OPCACHE_OPTIMIZE == true) {
+    return ComposerAutoloaderInit$suffix::getLoader(true);
+}
+
+return ComposerAutoloaderInit$suffix::getLoader(false);
```
stof: all this could be simplified:

```php
return ComposerAutoloaderInit$suffix::getLoader(defined('COMPOSER_OPCACHE_OPTIMIZE') && COMPOSER_OPCACHE_OPTIMIZE);
```

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
+            $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR).$ext;
+
+            $first = $class[0];
+            if (isset($this->prefixLengthsPsr4[$first])) {
+                foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
+                    if (0 === strpos($class, $prefix)) {
+                        foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
+                            $file = $dir.DIRECTORY_SEPARATOR.substr($logicalPathPsr4, $length);
+                            if ($searchMode == self::SEARCHMODE_OPCACHE) {
+                                if (opcache_is_script_cached($file) == true) {
```
stof: ``== true`` is not needed

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
 
-        // PSR-4 fallback dirs
-        foreach ($this->fallbackDirsPsr4 as $dir) {
-            if (file_exists($file = $dir . DIRECTORY_SEPARATOR . $logicalPathPsr4)) {
-                return $file;
+            // PSR-4 fallback dirs
+            foreach ($this->fallbackDirsPsr4 as $dir) {
+                $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr4;
+                if ($searchMode == self::SEARCHMODE_OPCACHE) {
+                    if (opcache_is_script_cached($file) == true) {
```
stof: same here

Danack:
I tried to use Blackfire, but it seems to require PHP5.3, and clashes with the self packaged version of PHP 5.6 I'm using. So instead I made a benchmark using [siege 3.0.9](http://www.joedog.org/siege-home/)

The test was done on a webserver running in a vagrant box, which has incredibly slow IO and not great memory throughput either. So the benchmark I ran may be very different from a benchmark on actual hardware, but it's not that dissimilar to the cheap end of Amazons EC2 instances.

The test was run with `siege -b -t20S http://basereality.test/` and the results are:

Opcache optimisation disabled + standard classmap
Transaction rate:	        7 trans/sec

Opcache optimisation disabled + Composer dumpautoload -o
Transaction rate:	       22 trans/sec

OPcache optimisation enabled + Composer dumpautoload -o  
Transaction rate:	       81 trans/sec

OPcache optimisation enabled + standard classmap  
Transaction rate:	       84 trans/sec

The results show more than anything else that vagrant is very slow on disk IO. But 


The numbers jumped around a little, but it's reasonable for the 'optimized classmap + opcache optimisation' to be slower than just the opcache optimisation as:

i) the lookup in OPCache is doing the same thing as the optimised classmap is doing, except it's doing it in C rather than PHP.

ii) The memory used on the page decreases with the classmap the peak memory used is 2,752,832 and about 1,911,632 without.

iii) Array access is slow in PHP. There are some ramblings about where the classmap values are copied internally by PHP are here http://pastebin.com/NB8G1zCj It's not guaranteed to be correct.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
+            // PSR-4 lookup
+            $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR).$ext;
+
+            $first = $class[0];
+            if (isset($this->prefixLengthsPsr4[$first])) {
+                foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
+                    if (0 === strpos($class, $prefix)) {
+                        foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
+                            $file = $dir.DIRECTORY_SEPARATOR.substr($logicalPathPsr4, $length);
+                            if ($searchMode == self::SEARCHMODE_OPCACHE) {
```
staabm: Minor: Tripple `=`

stof:
@Danack I'm using Blackfire with PHP 5.6 without issue

stof:
If the optimization is that efficient, I'm not sure we should require the user to define a constant (before requiring the autoloader) to benefit from it.

what do you think @Seldaek ?

staabm:
> I can't see any reason why it shouldn't be enabled by default....but there might be.

Danack:
As the author of the PR - please don't enable it by default straight away. I am not convinced that OPCache is totally bug free, and enabling by default could break a lot of people's stuff.

If the functionality was added, but disabled by default, it would allow many more people to test the behaviour across different platforms, which should reveal any issues.

Then in the next release of Composer it could be enabled by default, with a much better chance that it's not going to cause issues for huge numbers of people.


Also to note - I don't believe this PR would replace the need for the dumped classmap optimization for all people. For some patterns of access the large classmap optimization should still give better performance.

Danack:
So.... I realise that being hassled about PR's is annoying, but so is thinking that I've made a nice addition to a project and not having it be merged.

Seldaek:
@Danack yup. I took a look and beyond CS issues and that todo line that probably should go, it looks alright to me. Regarding how to enable this, typically we'd do this with a composer option but given that ideally this should just be enabled by default once proven, and then the constant would vanish, it's cleaner/easier to remove like this.

I'll happily merge it after the next release is in so it has time to stabilize and be tested further and hopefully become default before the next-next release.

Thanks for the idea and effort, and regarding benchmarks I have one question: Opcache optimization disabled/enabled means that opcache was disabled entirely or just the COMPOSER_OPCACHE_OPTIMIZE flag?

Danack:
> Thanks for the idea and effort, and regarding benchmarks I have one question: Opcache optimization disabled/enabled means that opcache was disabled entirely or just the COMPOSER_OPCACHE_OPTIMIZE flag?

Just the COMPOSER_OPCACHE_OPTIMIZE flag . As I said, those benchmarks are not going to be a typical result for a real non-virtualised server - they difference is so dramatic because the IO is so slow in vagrant.

> I'll happily merge it after the next release is in so it has time to stabilize and be tested further and hopefully become default before the next-next release.

Thanks - I will leave it in your hands then. Obviously even if it's enabled by default, being able to disable it would be awesome if any bugs are revealed in OPCache.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -56,6 +56,29 @@ class ClassLoader
...
+    const SEARCHMODE_OPCACHE = 0;
+    const SEARCHMODE_FILE = 1;
+    
+    public function __construct($enableOpcacheOptimize = false)
+    {
+        $this->searchModes = array(
+            self::SEARCHMODE_FILE
+        );
+
+        if ($enableOpcacheOptimize == false) {
```
alcohol: Please, yoda + strict. :-)

src/Composer/Autoload/ClassLoader.php
```diff
@@ -56,6 +56,29 @@ class ClassLoader
...
+            return;
+        }
+        if (function_exists('opcache_is_script_cached') == false) {
```
alcohol: Same as above, yoda + strict.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
-            if (file_exists($file = $dir . DIRECTORY_SEPARATOR . $logicalPathPsr4)) {
-                return $file;
+            // PSR-4 fallback dirs
+            foreach ($this->fallbackDirsPsr4 as $dir) {
+                $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr4;
+                if ($searchMode === self::SEARCHMODE_OPCACHE) {
+                    if (opcache_is_script_cached($file)) {
+                        return $file;
+                    }
+                }
```
alcohol: This is not PSR-2.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
+                foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
+                    if (0 === strpos($class, $prefix)) {
+                        foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
+                            $file = $dir.DIRECTORY_SEPARATOR.substr($logicalPathPsr4, $length);
+                            if ($searchMode === self::SEARCHMODE_OPCACHE) {
+                                if (opcache_is_script_cached($file)) {
+                                    return $file;
+                                }
+                            }
+                            else if ($searchMode === self::SEARCHMODE_FILE) {
```
alcohol: This is not PSR-2.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
+                foreach ($this->prefixesPsr0[$first] as $prefix => $dirs) {
+                    if (0 === strpos($class, $prefix)) {
+                        foreach ($dirs as $dir) {
+                            $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr0;
+                            if ($searchMode === self::SEARCHMODE_OPCACHE) {
+                                if (opcache_is_script_cached($file)) {
+                                    return $file;
+                                }
+                            }
+                            else if ($searchMode === self::SEARCHMODE_FILE) {
```
alcohol: This is not PSR-2.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
-            // PEAR-like class name
-            $logicalPathPsr0 = strtr($class, '_', DIRECTORY_SEPARATOR) . $ext;
-        }
+            // PSR-0 lookup
+            if (false !== $pos = strrpos($class, '\\')) {
+                // namespaced class name
+                $logicalPathPsr0 = substr($logicalPathPsr4, 0, $pos + 1)
+                    .strtr(substr($logicalPathPsr4, $pos + 1), '_', DIRECTORY_SEPARATOR);
+            }
+            else {
```
alcohol: PSR-2.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
-                return $file;
+            // PSR-0 fallback dirs
+            foreach ($this->fallbackDirsPsr0 as $dir) {
+                $file = $dir.DIRECTORY_SEPARATOR.$logicalPathPsr0;
+                if ($searchMode === self::SEARCHMODE_OPCACHE) {
+                    if (opcache_is_script_cached($file)) {
+                        return $file;
+                    }
+                }
+                else if ($searchMode === self::SEARCHMODE_FILE) {
```
alcohol: PSR-2.

lisachenko:
Hello, guys!

I want to clarify some information about the real need for this trick with `opcache_is_script_cached()` which is available only for the modern PHP>=5.6.

Instead of adding this check, it will be better just to enable [opcache.enable-file-override](http://php.net/manual/en/opcache.configuration.php#ini.opcache.enable-file-override) option.

In that case, all positive `file_exists()` check will be cached for production and performance will be nice.

Danack:
> Instead of adding this check, it will be better just to enable opcache.enable-file-override option.

As discussed elsewhere, PHP does not support negative caches for file_exists. 

This is the difference between opcache_is_script_cached and using the caching of file_exists info. If you look at the PR, the way it works is to look in all possible places that a file could exist, and check it with the function opcache_is_script_cached which never incurs a filesystem hit.

It then checks again using file_exists, which will incur file system hit for each wrong location checked, even if caching is enabled. Using opcache.enable-file-override would only replace this PR if we could look in the correct place first.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -343,61 +366,97 @@ public function findFile($class)
...
+            // PSR-4 lookup
+            $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR).$ext;
+
+            $first = $class[0];
+            if (isset($this->prefixLengthsPsr4[$first])) {
+                foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
+                    if (0 === strpos($class, $prefix)) {
+                        foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
+                            $file = $dir.DIRECTORY_SEPARATOR.substr($logicalPathPsr4, $length);
+                            if ($searchMode === self::SEARCHMODE_OPCACHE) {
```
lisachenko: Suggestion: it can be better to refactor this code to use inheritance or something like strategies to avoid extra level of code nesting. Complexity of the method will be lower. Just a thought :)

acasademont:
Question: if i have this setting enabled http://php.net/manual/en/opcache.configuration.php#ini.opcache.enable-file-override would this patch provide any performance improvement?

acasademont:
Seems like the question was already solved!

src/Composer/Autoload/ClassLoader.php
slbmeh: a monad would work well here

Danack:
Apparently there is a bug in PHP that stops the opcache_is_script_cached function working correctly if validate_timestamps is set to false https://bugs.php.net/bug.php?id=69281 This affect PHP 5.6.6 and 5.6.7 . It does not seem to be present on 5.6.2 which is where I did my testing.

This should be fixed in the next version of 5.6.x It shouldn't be an issue for this PR except that people won't see any performance benefit if they have validate_timestamps set to 0 and they're using one of the affected PHP versions.

Danack:
btw I don't intend to update the pull-request to 'fix' the alleged code style issues. The pull-request is consistent with the code style that is already there. The PR is intentionally a minimal one that is a pure functional change without changing the code style, so that it is easy to review the functional change.

If someone with commit access want to apply a code-style fixer before this is merged, or do other code style refactorings that would be 100% fine with me.

Seldaek:
@nicolas-grekas as you're in the autoloading benchmark business atm, would you want to check if this is a change still worth doing? :)

nicolas-grekas:
Not sure it's still worth either...  Benchmark would be mandatory to decide on which use case it would be faster, but it can't be faster than a dumped classmap...

Seldaek:
Yeah I doubt it as well, it saves some memory and gets rid of the file_exists() in the psr-x case but still lots of computing vs just a classmap isset(). I guess let's close then, but if you have time to give it a shot it would be interesting to know for sure. Should be easy enough to patch the ClassLoader in a simple way just to benchmark.

PR #3804, opened 2015-02-28
Add --no-check-publish option to "composer validate"

legoktm:
This is useful when you want to use composer to manage dependencies, but
don't actually want your project to be installable as a composer package.

Any issues that would prevent publishing are still shown, but as a warning
instead of an error.


Seldaek:
Thanks

PR #3806, opened 2015-03-01
Added php7 to travis

fonsecas72:


stof:
it would be great to debug why the JSON layer does not work well on PHP 7

fonsecas72:
@stof  It'll take time but'll look into that in a few hours. I already have PHP7 in a box but I'm working right now. I'm also curious about it, amazing speed.

fonsecas72:
@stof This is caused by some weird behavior of PHP7.
It happens when using the ```[]``` like ```$var[] = 'something';``` the incremental index is not being correctly made.
This happens with seld/jsonlint in these lines
```php
$this->stack[] = $this->productions_[$action[1]][0];    // push nonterminal (reduce)
$this->vstack[] = $yyval->token;
$this->lstack[] = $yyval->store;
$newState = $this->table[$this->stack[count($this->stack)-2]][$this->stack[count($this->stack)-1]];
$this->stack[] = $newState;
```
And if I change those to:
```php
$this->stack[count($this->stack)] = $this->productions_[$action[1]][0];    // push nonterminal (reduce)
$this->vstack[count($this->vstack)] = $yyval->token;
$this->lstack[count($this->lstack)] = $yyval->store;
$newState = $this->table[$this->stack[count($this->stack)-2]][$this->stack[count($this->stack)-1]];
$this->stack[count($this->stack)] = $newState;
```
The problem is "fixed".
The first ```stack``` parameter equals:
```php
Array
(
    [0] => 0
    [1] => 17
    [2] => 14
    [4] => 3
)
```
and the second (correct):
```php
Array
(
    [0] => 0
    [1] => 17
    [2] => 14
    [3] => 3
)
```

fonsecas72:
BTW I only tested with "testParseErrorDetectExtraCommaInArray". I'm guessing all other failures are relative the same thing.

stof:
@fonsecas72 this is a major regression in PHP. Have you reported it ? If yes, please post the link to the PHP bug report so that we can follow it

fonsecas72:
@stof I couldn't yet reproduce it in a simpler env. And I'm still not sure that this is a problem with PHP7.
I see a ```array_slice``` there and maybe it is related (?)
I'm still getting in touch with the parser and trying to create a simple script to replicate the misbehave.

fonsecas72:
Yup, I managed to replicate with the following
```php
$stack[] = 'wew';
$stack[] = 'wew';
$stack[] = 'wew';
$stack[] = 'wew';
$stack = array_slice($stack, 0, - 2);
$stack[] = 'wew';

print_r($stack);
```
Output:
```
Array
(
    [0] => wew
    [1] => wew
    [3] => wew
)
```
With php 5.5.9
```
Array
(
    [0] => wew
    [1] => wew
    [2] => wew
)
```

fonsecas72:
preserve_keys default must have changed in php7.
I've PR to fix this https://github.com/Seldaek/jsonlint/pull/29

Danack:
Apparently this you reported this already - bugs.php.net/bug.php?id=69172 which means it should be fixed automatically in the next nightly of 7 on Travis.

The change to JsonParser.php shouldn't be required as the existing code should work on 7.

fonsecas72:
@Danack yup, so it seems :)

stof:
@Seldaek there is still a failure on PHP 7

Seldaek:
I saw but it's just a network fuckup it seems.

Issue #3791, opened 2015-02-26
Plugins that has dependencies does not sort properly (uninstall)


vancoz:
The workaround introduced here (https://github.com/composer/composer/commit/62b506214640ae66ad88a07240a0bbb0201bd1e3) does not cover two use-cases:
*  composer plugin has dependency (other then compose-plugin-api)
*  composer plugin is dependency for module and we are trying to uninstall module (composer plugin subscribed for post-package-uninstall and cleanups module related code, folders, etc...) and uninstall plugin too.

For example:

1. Clone Magento2 (https://github.com/magento/magento2), run _composer install_
2. Add _"texanhogman/m2-minimal": "dev-master"_ to "require" in composer.json 
3. Run _composer update_  (*result:* new module added to _../app/code/_ (/))
4. Remove _"texanhogman/m2-minimal": "dev-master"_  from composer.json 
5. Run _composer update_ 

This happens due to wrong operations sorting, as for me would be good to implement the next behavior:

1. uninstall operations - move all packages uninstalls to the top, then plugin's dependencies and plugins uninsntall operations.

2. install and update operations - move all plugin's dependencies to the top (but after uninstall stuff) then plugins install operations, then install of all other packages.

slbmeh:
I'm not sure I understand your issue fully. For updating and uninstalling plugins it is advised in the [plugin documentation](https://getcomposer.org/doc/articles/plugins.md) to run composer with `--no-plugins`.  There is no way to reverse actions done by a plugin activation so it's a bit more complicated than moving it's install order to the end.

vancoz:
We are not going to install plugins independently. 

Plugin is the part (dependency) of our package and plugin should deploy package properly on post install or update and it should cleanup package  on post-package-uninstall.

We do not want to have installer plugin as part of our application composer requirements, it should be requirement of package.

So it should be available before install and should be uninstalled after the package.

Seldaek:
Closing as duplicate of #4520

PR #3807, opened 2015-03-02
Fix for #3791

vancoz:
This coed change does fix issue with operations order for plugins.
After that fix will be introduced following order:
1. uninstall packages
2. uninstall plugins and plugin's dependencies
3. install plugin's dependencies and plugin
4. install packages

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -624,7 +624,7 @@ class Installer
      */
     private function movePlugins(array $operations)
     {
-        $packagesIndex = array();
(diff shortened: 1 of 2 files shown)


vancoz:
Hello @Seldaek, may you please review this PR? This changes logic introduced by you.
Thank you.

Seldaek:
Closing as I am not exactly sure what it fixes and it is so old it doesn't seem that necessary. Does not seem worth the risk given the lack of info and tests to clarify what is being changed sorry.

PR #3813, opened 2015-03-03
Run processes with predictable LANGUAGE=C

nicolas-grekas:
For some reason, I ran composer with LANGUAGE=fr_FR, and that breaks parsing git's output.

### DIFF
--- a/src/Composer/Util/ProcessExecutor.php
+++ b/src/Composer/Util/ProcessExecutor.php
@@ -56,7 +56,7 @@ class ProcessExecutor
 
         $this->captureOutput = count(func_get_args()) > 1;
         $this->errorOutput = null;
-        $process = new Process($command, $cwd, null, null, static::getTimeout());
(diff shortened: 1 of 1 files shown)


src/Composer/Util/ProcessExecutor.php
```diff
@@ -56,7 +56,7 @@ public function execute($command, &$output = null, $cwd = null)
 
         $this->captureOutput = count(func_get_args()) > 1;
         $this->errorOutput = null;
-        $process = new Process($command, $cwd, null, null, static::getTimeout());
+        $process = new Process($command, $cwd, array('LANGUAGE' => 'C'), null, static::getTimeout());
```
stof: this will not inherit the current env anymore, which is a BC break. If you want to pass extra env vars, you should merge the existing env (see the Symfony ProcessBuilder which has some logic doing that)

nicolas-grekas: Fixed hopefully

Seldaek:
Thanks

PR #3816, opened 2015-03-04
Switch to the docker-based infrastructure on Travis

stof:
The docker-based infrastructure is faster and allows to use the build cache, which can make things even faster by caching deps downloads between builds for dists.

Given that the only usage of sudo is about installing an APT package and this package is in the whitelist for the travis addon (because I requested adding it a while ago), the build is compatible with this infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,15 @@
 language: php
 
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.composer/cache
+
+addons:
+  apt_packages:
+    - parallel
+
 php:
   - 5.3.3
   - 5.3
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks.. About the failure I am wondering if it's really network or not after all, could be another php7 regression.

fonsecas72:
@Seldaek @stof this is outputting ```4``` in php master
http://3v4l.org/6gh25

Seldaek:
And what do you get null or false as return value? http://3v4l.org/kACp7

fonsecas72:
NULL

stof:
First question, how can the ``$json`` string be ``null`` in this code ? the method document that it expects a string as argument.

but anyway, if the behavior changed in PHP, it should be reported as a bug, so that they either fix it or document it as an intentional BC break.

fonsecas72:
https://bugs.php.net/bug.php?id=69187

fonsecas72:
@stof @Seldaek They document it as an intentional BC break.
https://github.com/php/php-src/commit/9d037d574cc359405c7a818c2235644633705999

PR #3818, opened 2015-03-05
Duplicate download progress

Hellov:
When package has been downloaded, composer duplicate progress message.
https://drive.google.com/file/d/0B9XRnt4GsyJrZV9MTmZoSFc0OFk/


Seldaek:
Thanks, I also fixed the first line with 0% that was on the wrong line in 331425bcb3cd13a5d5d807c13c6a651229b56ac8

PR #3819, opened 2015-03-05
Add/Fix DocBlock

fonsecas72:
Related https://github.com/composer/composer/pull/3816#issuecomment-77207670

* Fixed $json parameter type from parseJson() and validateSyntax()
* Added docBlock to throwEncodeError()

### DIFF
--- a/src/Composer/Json/JsonFile.php
+++ b/src/Composer/Json/JsonFile.php
@@ -215,21 +215,27 @@ class JsonFile
         return $result;
     }
 
+    /**
+     * Throws an exception according to a given code with a customized message
+     *
+     * @param int $code return code of json_last_error function
+     * @throws \RuntimeException
+     */
     private static function throwEncodeError($code)
     {
         switch ($code) {
             case JSON_ERROR_DEPTH:
                 $msg = 'Maximum stack depth exceeded';
-            break;
+                break;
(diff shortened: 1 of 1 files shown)


src/Composer/Json/JsonFile.php
```diff
@@ -258,7 +264,7 @@ public static function parseJson($json, $file = null)
     /**
      * Validates the syntax of a JSON string
      *
-     * @param  string                    $json
+     * @param  mixed                     $json
```
Seldaek: Why mixed? These kinda expect a json string..

fonsecas72: Because it receives a ```getContents()``` in some places which can return ```null```, ```false``` or ```string```

Seldaek: Yeah but the function still expects a string, and same for the mixed above. The fact that garbage is sometimes given to it in case of failure doesn't make the argument mixed.

fonsecas72: mm! Then I'm learning something, thank you! :)

fonsecas72:
I've removed my mistake (sorry) and squashed the commits. But if you're not interested in this PR you can close it, no problem.

Seldaek:
No problem and thanks for the docblock :)

PR #3825, opened 2015-03-07
Use get home from Config instead of factory

fonsecas72:
If I understood this correctly, we have no guarantee that the COMPOSER_HOME will not end with ```/``` or ```\``` so we should rely on the Config instead.


Seldaek:
Thanks, although just FYI, PHP deals with multiple / and \ in paths just fine, e.g.:

```
$ php -r 'var_dump(file_exists("../////composer////\\\\////src"));'
bool(true)
```

fonsecas72:
Thank you. I realized this latter on but I don't know, It just feels better even if the reason I gave doesn't make sense :)

PR #3828, opened 2015-03-08
Test on HHVM nightly releases. Allow to fail.

hkdobrev:
Nightly releases could change in unpredictable way.  So they should not fail the whole build.

However we could catch problems and breaking changes on next versions of HHVM early.

I suspect Composer have some problems on HHVM nightly, because of this build: https://travis-ci.org/erusev/parsedown/jobs/53556585#L94

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ php:
   - 5.6
   - 7.0
   - hhvm
+  - hhvm-nightly
+
+matrix:
+  allow_failures:
+    - php: hhvm-nightly
 
 before_script:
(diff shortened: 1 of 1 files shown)


hkdobrev:
@Seldaek Do you have any idea why the HHVM nightly build fails like that?

Seldaek:
It seems to fail already at `composer --version` with some phar related errors so I guess they have a critical bug in there but I am not really used to reading their debug output so I might be wrong.. In any case I'd report it over there :)

fonsecas72:
@hkdobrev   I got here a little late but since we have builds in the allowed_failures section I think we should add the fast_finish option.
```
matrix:
  fast_finish: true
```

This will make the travis to not wait for the builds that are allowed to fail.

Issue #3823, opened 2015-03-06
Support autoloading enums in Hack code


fredemmott:
eg:

```Hack
<?hh

enum Foo: int {
  BAR = 1;
  BAZ = 2;
}
```

Hack Enums are syntactic sugars around classes, and can be autoloaded like classes via spl_autoload_register() (or HH\autoload_set_paths)

Tokens are T_ENUM (388) followed by T_STRING (class name)

stof:
is it about the classmap generator or the classloader itself (i.e. does it work already for PSR-0 or PSR-4 autoloading) ?

fredemmott:
Classmap generator. When it's being loaded, it's identical to a class, so should already work for PSR0/4.

stof:
This would be a matter of updating the regex on https://github.com/composer/composer/blob/b80038804ff791a9d4608d737ad937b6033b0bae/src/Composer/Autoload/ClassMapGenerator.php#L157 to add ``|enum`` on HHVM (similar to the way traits are added on PHP 5.4+)

PR #3831, opened 2015-03-09
Add support for using classmap to autoload Hack enums

fredemmott:
fixes composer/composer#3823

Ran tests with both PHP5.5.9-1ubuntu4.5 and HHVM 3.6. Test fails on HHVM only
if I back out the ClassMapGenerator.php change.

### DIFF
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -112,10 +112,9 @@ class ClassMapGenerator
      */
     private static function findClasses($path)
     {
-        $traits = version_compare(PHP_VERSION, '5.4', '<') ? '' : '|trait';
-        $enums = '';
+        $extraTypes = version_compare(PHP_VERSION, '5.4', '<') ? '' : '|trait';
         if (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.3', '>=')) {
-            $enums = '|enum';
+            $extraTypes .= '|enum';
         }
 
         try {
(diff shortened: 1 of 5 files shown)


src/Composer/Autoload/ClassMapGenerator.php
```diff
@@ -113,6 +113,10 @@ public static function createMap($path, $whitelist = null, IOInterface $io = nul
     private static function findClasses($path)
     {
         $traits = version_compare(PHP_VERSION, '5.4', '<') ? '' : '|trait';
+        $enums = '';
+        if (defined('HPHP_VERSION') && version_compare(HPHP_VERSION, '3.3', '>=')) {
+            $enums = '|enum';
```
stof: instead of 2 variables for traints and enums, I would use a single variables (``$extraTypes`` or something like that)

src/Composer/Autoload/ClassMapGenerator.php
```diff
@@ -113,6 +113,10 @@ public static function createMap($path, $whitelist = null, IOInterface $io = nul
     private static function findClasses($path)
     {
         $traits = version_compare(PHP_VERSION, '5.4', '<') ? '' : '|trait';
+        $enums = '';
+        if (defined('HPHP_VERSION') && version_compare(HPHP_VERSION, '3.3', '>=')) {
```
stof: use ``HHVM_VERSION`` ratheer than the old ``HPHP_VERSION``

fredemmott:
Both changed.

stof:
:+1:

Seldaek:
Thanks!

PR #3845, opened 2015-03-14
Avoid parser error if composer.json existing but empty

vlakoff:
Using the require command, if there is a blank composer.json a parser error is thrown:

    [Seld\JsonLint\ParsingException]
    "./composer.json" does not contain valid JSON
    Parse error on line 1:
    ^

### DIFF
--- a/src/Composer/Command/RequireCommand.php
+++ b/src/Composer/Command/RequireCommand.php
@@ -82,6 +82,10 @@ EOT
             return 1;
         }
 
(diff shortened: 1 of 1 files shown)


alcohol:
The PR title and actual context don't quite match up. This doesn't "fix" the parser error (not sure there is anything to fix), it just circumvents the issue (and only during the use of `require`).

Seldaek:
Thanks

PR #3853, opened 2015-03-19
Git Checkout Branch / File Differentiation

hakre:
This fixes a minor inaccuracy issue when creating git checkout commands
for branches.

The git checkout command used within `GitDownloader::updateToCommit()`
does  not use the "`--`" sequence to separate branch from file parameters.

This leads to an inaccuary as git tries as well the branch name as file
name. If the non-existent branch is actually the name of a file, the file
is checked out. If the non-existent branch is not the name of a file, an
error message is given:

> error: pathspec 'non-existent-branch' did not match any file(s) known to
git.

Both cases are not expected for the program flow in
`GitDownloader::updateToCommit()`.

The only thing that is expected is a non-existent branch to fail to
checkout - but with a different error message:

> // reference was not found (prints "fatal: reference is not a tree:
 $ref")

This can be easily fixed by adding the missing separator when constructing
the command which is applied with this commit.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -215,7 +215,7 @@ class GitDownloader extends VcsDownloader
      */
     protected function updateToCommit($path, $reference, $branch, $date)
     {
-        $template = 'git checkout %s && git reset --hard %1$s';
+        $template = 'git checkout %s -- && git reset --hard %1$s';
         $branch = preg_replace('{(?:^dev-|(?:\.x)?-dev$)}i', '', $branch);
 
         $branches = null;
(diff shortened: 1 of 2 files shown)


stof:
please update tests as well

hakre:
All tests run smooth within windows cmd shell, but on cygwin some fail because the composer path is expected in the tests with drive letter "C:" in lowercase (as per `getcwd()`) but on cygwin the actual one is an uppercase "C". Should I take a look for that, too?

hakre:
this removes the test-issue under cygwin, please review so it's OK to squash here if wanted: https://github.com/hakre/composer/commit/a5c63aa485a1974ac1b1d2c98510c6e337405245

Seldaek:
Thanks. I'll merge this as is. If you like the cygwin fix (msys works fine for me) please send a PR for it.

PR #3854, opened 2015-03-19
Prevent fatal error on missing symlink() function in tests

hakre:
The testsuite didn't run through for me because the php symlink() function
was missing.

It is only available on Windows Visa/Server 2008 or higher.

This commit fixes the issue by checking if the method exists, and if not,
marks the test as skipped because of a non-matching precondition.

### DIFF
--- a/tests/Composer/Test/Util/FilesystemTest.php
+++ b/tests/Composer/Test/Util/FilesystemTest.php
@@ -188,6 +188,7 @@ class FilesystemTest extends TestCase
         @mkdir($basepath . "/real", 0777, true);
         touch($basepath . "/real/FILE");
 
+        $this->skipTestIfSymlinkPhpFunctionIsMissing();
         $result = @symlink($basepath . "/real", $symlinked);
 
         if (!$result) {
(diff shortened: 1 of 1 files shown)


tests/Composer/Test/Util/FilesystemTest.php
```diff
@@ -237,4 +239,11 @@ public function testRemoveSymlinkedDirectoryWithTrailingSlash()
         $this->assertFalse(file_exists($symlinkedTrailingSlash));
         $this->assertFalse(file_exists($symlinked));
     }
+
+    private function skipTestIfSymlinkPhpFunctionIsMissing()
+    {
+        if (!function_exists('symlink')) {
```
staabm: maybe you should add a comment when this function is not available

hakre: That's documented on the manual page of that function: http://php.net/symlink - standard behavior. When I first wrote the tests it was an oversight of me to not check for the availability in the first place.

Seldaek:
Using windows xp still? :| I'll merge because it's kinda harmless but I'd say it's time to update :)

PR #3857, opened 2015-03-20
Fix misspellings

isoroku:


src/Composer/DependencyResolver/PolicyInterface.php
```diff
@@ -21,5 +21,5 @@
 {
     public function versionCompare(PackageInterface $a, PackageInterface $b, $operator);
     public function findUpdatePackages(Pool $pool, array $installedMap, PackageInterface $package);
-    public function selectPreferedPackages(Pool $pool, array $installedMap, array $literals);
+    public function selectPreferredPackages(Pool $pool, array $installedMap, array $literals);
```
xabbuh: This becomes an issue for people who rely on the interface.

Seldaek: I added a BC method in the DefaultPolicy for people using that. I doubt many people have custom implementation of that interface. At least http://packanalyst.com/class?q=Composer%5CDependencyResolver%5CPolicyInterface finds nothing.

xabbuh: :+1: thanks

PR #3866, opened 2015-03-23
Improved wording

khromov:
All other verbs are in the form of "doing something", ie Installing, Downloading. "connection" is the odd one out.


Seldaek:
Thanks

PR #3868, opened 2015-03-24
Support comments in json files.

trent-waddington:
People often complain that they can't comment their json files. This is especially true for configuration files.


Seldaek:
This has been discussed a few times already.. it would make composer.json a non-standard json file which will break third parties using the file (IDEs and others). I don't think it's worth it.

PR #3871, opened 2015-03-24
Fix Travis install for PHP 7

stof:


Seldaek:
Seems the problem fixed itself now.

stof:
@Seldaek Travis deployed an update of the VMs, and so a newer composer. But it has been reverted today because of errors with MySQL: http://docs.travis-ci.com/user/build-environment-updates/2015-04-09/

Seldaek:
Well the last few builds passed on PHP7 so I'd guess we're good no?

stof:
yeah, seems so.

PR #3872, opened 2015-03-25
Enhancement: Sort packages by importance, then alphabetically

localheinz:
This PR

* [x] sorts packages by importance (platform packages first, then PHP dependencies) and alphabetically

Follows #3549.

:bulb: Not sure if anyone but me uses that feature, but it would be a lot nicer if packages were sorted by importance (platform packages first), then followed by PHP dependencies, both sorted alphabetically.

See https://getcomposer.org/doc/02-libraries.md#platform-packages.
> 
* `php` represents the PHP version of the user, allowing you to apply constraints, e.g. `>=5.4.0`. To require a 64bit version of php, you can require the php-64bit package.
>* `hhvm` represents the version of the HHVM runtime (aka HipHop Virtual Machine) and allows you to apply a constraint, e.g., `>=2.3.3`.
> * `ext-<name>` allows you to require PHP extensions (includes core extensions). Versioning can be quite inconsistent here, so it's often a good idea to just set the constraint to *. An example of an extension package name is `ext-gd`.
>* `lib-<name>` allows constraints to be made on versions of libraries used by PHP. The following are available: curl, iconv, icu, libxml, openssl, pcre, uuid, xsl.

### DIFF
--- a/src/Composer/Json/JsonManipulator.php
+++ b/src/Composer/Json/JsonManipulator.php
@@ -92,14 +92,47 @@ class JsonManipulator
 
         if (true === $sortPackages) {
             $requirements = json_decode($links, true);
-
-            ksort($requirements);
+            $this->sortPackages($requirements);
             $links = $this->format($requirements);
         }
 
         $this->contents = $matches[1] . $matches[2] . $links . $matches[4];
 
         return true;
+    }
+
+    /**
+     * Sorts packages by importance (platform packages first, then PHP dependencies) and alphabetically.
+     *
+     * @link https://getcomposer.org/doc/02-libraries.md#platform-packages
+     *
+     * @param array $packages
+     */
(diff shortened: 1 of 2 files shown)


localheinz:
@Seldaek 

Any idea if you think this is cool for merging?

hannesvdvreken:
Ofcourse. I would always expect this with `--sort-packages`.

kayladnls:
:+1:

Seldaek:
Thanks, made a few adjustments though:

https://github.com/composer/composer/commit/06fb00bfe501fb9f1e7d375d241e821ea3fa3e27?w=1

https://github.com/composer/composer/commit/ab02867015786c742e2a57d3be0eb50648945088

localheinz:
@Seldaek 

Awesome, thanks a lot!

PR #3878, opened 2015-03-27
added "example using a custom HTTP Header field" to docu, issue https://...

jakoch:
...github.com/composer/composer/issues/3756


Seldaek:
Thanks

PR #3882, opened 2015-03-28
Added retries to GitDownloader when code is 128

fonsecas72:
I'm having some trouble downloading (git cloning) packages due to connection problems.
I don't known if I'm missing anything but this seems to fix my problem.


fonsecas72:
I guess this is not needed

PR #3884, opened 2015-03-28
Fix #3813,  #3830

### LINKED ISSUE #3830: composer install prompts for Basic Auth Username/Password with configured credentials
I'm installing a private package and I'm authenticating with basic auth. I'm storing the credentials in a global `auth.json` in the COMPOSER_HOME.

When running `php composer.phar install --no-interaction --no-dev --no-ansi --optimize-autoloader` I get the strange error in my deploy script:

```
[RuntimeException]                                                           
  Failed to execute git clone --no-checkout 'REDACTED' 'REDACTED' && cd 'REDACTED' && git remot  
  e add composer 'http://git.redacted.com' && git fetch composer                                                     
  fatal: could not read Username for 'http://git.redacted.com': No such   
  device or address
```

That error message obviously happens because the deploy script hasn't access to `/dev/tty` and thus cannot prompt for the credentials.

The real issue though is, that it's prompting for credentials in the first place, although `store-auths` is set to `false` and `--no-interaction` is passed on the CLI.

The regression must have happened after version `2015-02-27_12-27-56-c7b0628`. That version doesn't exhibit that behavior.

frederikbosch:
Fix #3830

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -177,6 +177,11 @@ class Git
             unset($_SERVER['GIT_WORK_TREE']);
         }
 
+        // Run processes with predictable LANGUAGE
+        if (getenv('LANGUAGE') !== 'C') {
+            putenv('LANGUAGE=C');
+        }
+
         // clean up env for OSX, see https://github.com/composer/composer/issues/2146#issuecomment-35478940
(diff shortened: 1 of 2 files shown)


Seldaek:
Ah damn it another issue caused by this env override.. I guess putenv is definitely a better idea then, thanks :)

PR #3890, opened 2015-03-30
Add gz compression of zip files

andrerom:
PharData does not compress zip files unless asked for it, so check if zlib is installed and always compress if it is. Reduced files size by order of magnitude if package is mainly php scripts and other text files.

( this was originally part of #3390 )

### DIFF
--- a/src/Composer/Package/Archiver/PharArchiver.php
+++ b/src/Composer/Package/Archiver/PharArchiver.php
@@ -41,6 +41,12 @@ class PharArchiver implements ArchiverInterface
             $files = new ArchivableFilesFinder($sources, $excludes);
             $phar->buildFromIterator($files, $sources);
 
+            // Compress zip file if possible (not implemented yet as of HHVM 3.5.0)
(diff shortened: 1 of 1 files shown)


Seldaek:
I am not sure if that's a good idea because the phar archives might be used on other systems that don't have gzip capability and then they can't be extracted. That's why we don't use compression in composer.phar for example.

andrerom:
Ok, one one side I would argue that gzip should be a requirement for composer, but I get the argument of course.

What about #2706 (same + this in #3390 ) for adding support for .gz and .bz2 files? Given they have different file names it is easier to handle case where users don't have the needed extension installed at least.

stof:
@andrerom HHVM does not support compressed phars IIRC. So this makes it a no-go

andrerom:
True, I agree that is a blocker, added ping back on relevant hhvm issue.

Seldaek:
Closing as I don't think this is practical nor really worth it in the phar format.

PR #3894, opened 2015-03-31
No need to update prefered options twice

ceeram:


src/Composer/Command/CreateProjectCommand.php
```diff
@@ -154,9 +152,6 @@ public function installProject(IOInterface $io, Config $config, $packageName, $d
             $composer->getEventDispatcher()->dispatchScript(ScriptEvents::POST_ROOT_PACKAGE_INSTALL, $installDevPackages);
         }
 
-        $rootPackageConfig = $composer->getConfig();
-        $this->updatePreferredOptions($rootPackageConfig, $input, $preferSource, $preferDist);
```
stof: This is not equivalent. The installation of deps uses a **different** config instance, because it takes the project-level config into option once it has the code of the project.

PR #3895, opened 2015-03-31
Add a setup() method to generated autoloaders.

deviantintegral:
I've run into a static caching issue when using Composer with Drupal 7 projects.

[Composer Manager](https://www.drupal.org/project/composer_manager) handles composer.json management and updating vendors. After Drupal modules are enabled or composer libraries are updated, other code can respond to those events.

The issue arises when:

1. You enable a new module with something like `drush pm-enable <modulename>`.
1. During the drush bootstrap, we pull in composer's autoloader.
1. Composer manager will update composer.json and update dependencies on disk.
1. The new module tries to use new libraries it's required, but can't because the autoloader is already registered. The old `autoload_classmap.php` and so on are being used.
1. We can't force a rescan due to the static cache of $loader in `ComposerAutoloaderInitComposerManager::getLoader()`.

Rather than adding a reset flag, I've pulled out a new setup() function so other code can call getLoader() followed by setup().

Tests pass and code coverage stats are the same.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -487,7 +487,13 @@ class ComposerAutoloaderInit$suffix
         spl_autoload_register(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'), true, $prependAutoloader);
         self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader();
         spl_autoload_unregister(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'));
+        self::setup(\$loader);
 
+        return \$loader;
+    }
+
+    public static function setup(\$loader)
+    {
 
 HEADER;
 
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
@@ -22,7 +22,13 @@ class ComposerAutoloaderInitFilesAutoloadOrder
         spl_autoload_register(array('ComposerAutoloaderInitFilesAutoloadOrder', 'loadClassLoader'), true, true);
         self::$loader = $loader = new \Composer\Autoload\ClassLoader();
         spl_autoload_unregister(array('ComposerAutoloaderInitFilesAutoloadOrder', 'loadClassLoader'));
+        self::setup($loader);
 
+        return $loader;
+    }
+
+    public static function setup($loader)
+    {
         $map = require __DIR__ . '/autoload_namespaces.php';
         foreach ($map as $namespace => $path) {
             $loader->set($namespace, $path);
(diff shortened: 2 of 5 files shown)


stof:
Your implementation is broken when using autoload_files: it will require again the files. This is because the whole autoloading initialization is currently designed to be called exactly once.

deviantintegral:
Good catch. I've got that fixed locally, but I'm running into real issues creating a test for it.

The problem is that each call to setUp() is creating a new working directory for the generated autoloader. This makes sense, but it means if we try to test `ComposerAutoloaderInitFilesAutoload` in more than one test we either get duplicate function declarations *or* file not found fatals.

For example, putting this in `AutoloadGeneratorTest`:

```php
public function testNewAutoloader()
    {
        $package = new Package('a', '1.0', '1.0');

        $this->repository->expects($this->once())
          ->method('getCanonicalPackages')
          ->will($this->returnValue(array()));
       $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', false, 'FilesAutoload');

        $loader = include_once $this->vendorDir.'/autoload.php';
        $loader = \ComposerAutoloaderInitFilesAutoload::getLoader();
    }
```

If you run just this test by itself, it will run fine. If you try to run all of the tests in the class, it will fatal on `Cannot redeclare composerRequireFilesAutoload()`. If I wrap that with a class_exists() test, and unregister the class, __DIR__ is still set to the test directory from whatever test was run first. That directory was removed, so we end up with an unresolvable require.

Any ideas on how to test this? Or, I'm open to any suggestions for solving the original issue if they will sidestep this testing problem.

stof:
@deviantintegral your case covering the fact that it can call setup twice should be a single test, using its own autoloader suffix. this way, it won't create duplicate function names with other tests. the issue is that you are reusing the same suffix than an existing test here

stof:
and given that the only use case for doing an autoloader setup again is adding new packages in it when your PHP process installed more dependencies, the proper solution IMO would be that your tool running the composer installation would get the autoloading info for the newly installed packages only and would add them in the composer autoloader instance (available through ``require __DIR__.'/vendor/autoload.php';``, with an adjusted path of course), rather than trying to reload the full autoloading config

On a side note, your current implementation will duplicate all mapping in the ClassLoader instance for PSR-0 and PSR-4 prefixes, making the autoloading slower.

Thus, your proposal expects the ``ComposerAutoloaderInit*`` class to become a public API of composer, which is not an option given that it has a generated name by default (and forcing a specific suffix should be done very carefully as it may lead to duplicate class definitions). This generated class is currently considered as an internal implementation details.
Note that the supported way to access the autoloader instance is not ``$loader = ComposerAutoloaderInit*``::getLoader()`` but ``$loader = require __DIR__.'/vendor/autoload.php';`` (which is why the autoload.php file is specifically designed to be loadable as many times as you want)

deviantintegral:
Thanks @stof for your feedback, it's appreciated! I ended up solving this <a href="https://www.drupal.org/node/2466871">entirely within the Drupal composer_manager module</a>, so this PR isn't needed.

Issue #3874, opened 2015-03-25
Composer-update keeps old remotes when git repository URL changes


DHager:
Composer-update should re-clone the temporary repositories in the `vendor` folder when their original source URL changes, or at least update the git remotes.  

Because it does not, you can end up with a workflow that pushes and pulls your vendor-folder changes against the wrong remote repository.

# Bug repro example

## Initial setup

1. We have two composer packages, `example/bug` and `example/dep`. 
2. Each contains only a minimal `composer.json` file.
3. There's an original git repository for the `example/dep` project in `/tmp/cbug/dep1` and a clone in `/tmp/cbug/dep2`. 

In `/tmp/cbug/mainproj/composer.json` : 

    {
        "name": "example/bug",    
        "require": {            "example/dep": "*@dev"        },    
        "repositories":[
            {"packagist": false},
            {"type" : "vcs", "url" : "/tmp/cbug/dep1" }
        ]
    }

In `/tmp/cbug/dep/composer.json` :

    { "name": "example/dep" }

## First composer update

Running commands:

    user@host:/tmp/cbug/mainproj$ ls
    composer.json
    user@host:/tmp/cbug/mainproj$ composer update
    Loading composer repositories with package information
    Updating dependencies (including require-dev)         
      - Installing example/dep (dev-master 7b14991)
        Cloning 7b149918414b8954e3d86db52201c22569dfc7a4
    
    Writing lock file
    Generating autoload files

Checking origin of the vendor folder's `example/dep`:

    user@host:/tmp/cbug/mainproj/vendor/example/dep$ git remote -v
    composer	/tmp/cbug/dep1 (fetch)
    composer	/tmp/cbug/dep1 (push)
    origin	/tmp/cbug/dep1 (fetch)
    origin	/tmp/cbug/dep1 (push)

## Changing the repository URL and seeing issues

So far so good, but now my remote repo has moved, perhaps because IT migrated it to a different server or something. Easy enough, all I should need to do is update `composer.json`. This new version points to the `dep2` folder:

    {
        "name": "example/bug",    
        "require": {            "example/dep": "*@dev"        },    
        "repositories":[
            {"packagist": false},
            {"type" : "vcs", "url" : "/tmp/cbug/dep2" }
        ]
    }

Running update with new repo-url:

    user@host:/tmp/cbug/mainproj$ composer update
    Loading composer repositories with package information
    Updating dependencies (including require-dev)         
    Nothing to install or update
    Writing lock file
    Generating autoload files

[...]

Issue #3703, opened 2015-02-03
Composer keeps remote if you have the correct version in vendor folder.


fabiocarneiro:
I guess it's a bug. When you update your packages and then you want to use a new remote and you have both remotes in the same version, although it properly connects to the new remote to get the last version. If you have that same version currently installed (in your vendor folder), it considers this package to be up to date and doesn't properly update the composer.lock with the new remote address, instead keeping the old remote address, which you may have access locally but removed this remove permission in a deployment environment (which was exactly what happened to me).

To fix this, once composer connects to a remote to retrieve the last package version, it must update the remote address in composer.lock no matter if the package is updated or not.

stof:
Duplicate of #2416

PR #3896, opened 2015-04-01
Bugfix for #3874

DHager:
This attempts to fix issues #3874 and #3703.

In summary, these changes will:

1. When updating a git-repository within the vendor folder, always update the "origin" remote. (As opposed to only during the initial cloning, which git does automatically.)
2. Consider a git-repository in the vendor folder in need of updating whenever the source repo-URL changes. (Without this change, it will not notice if the revision-hash is the same.)

### DIFF
--- a/src/Composer/Downloader/DownloadManager.php
+++ b/src/Composer/Downloader/DownloadManager.php
@@ -250,6 +250,14 @@ class DownloadManager
             return;
         }
 
+        // Totally different repositories mean it is not safe to update-in-place
+        if($initial->getSourceUrl() !== $target->getSourceUrl()){
+            $this->io->writeError('    <warning>Repository URL changed, re-downloading.</warning>');
+            $downloader->remove($initial, $targetDir);
+            $this->download($target, $targetDir);
+            return;
+        }
+
         if ($initialType === $targetType) {
             $target->setInstallationSource($installationSource);
             $downloader->update($initial, $target, $targetDir);
--- a/src/Composer/Downloader/DownloadManager.php
+++ b/src/Composer/Downloader/DownloadManager.php
@@ -250,14 +250,6 @@ class DownloadManager
             return;
         }
 
-        // Totally different repositories mean it is not safe to update-in-place
-        if($initial->getSourceUrl() !== $target->getSourceUrl()){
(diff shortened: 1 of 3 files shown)


stof:
Composer never use the ``origin`` remote for its operation. and there is no reason to force an UpdateOperation in case the source url changed if we already have the right commit (we may need to update the lock file and the installed.json metadata, but this is a totally different topic)

DHager:
> there is no reason to force an UpdateOperation in case the source url changed if we already have the right commit

I agree that `UpdateOperation` may be too heavy-weight, but *something* needs to happen, because the commit-hash is not enough to uniquely/accurately identify the state of a vendor-folder, dev-mode repository.

The same hash means you have the same files in your repo, but it does not mean you have the same remotes for when you start editing those files.

> Composer never use the  `origin`  remote for its operation.

True, but *developers* do. It seems editing inside `vendor/` and pushing the changes to a remote is fairly common advice when it comes to complex projects and co-developed packages.

Composer might not care, but having an `origin` unexpectedly pointing to a wrong or obsolete location is potentially very bad, because developers' work will be getting based-on/changing the wrong repository as they pull and push.

stof:
> The same hash means you have the same files in your repo, but it does not mean you have the same remotes for when you start editing those files.

Composer does not care about the remote at all. and it does not set it at all

> True, but developers do. It seems editing inside vendor/ and pushing the changes to a remote is fairly common advice when it comes to complex projects and co-developed packages.

but half of the time, you won't be allowed to push to the repo used by Composer when installing anyway, and would need to push to your own fork on Github instead.
In this case, resetting the remote url on each composer operation actually makes things harder for the developer

DHager:
> Composer does not care about the remote at all. and it does not set it at all

I'm confused, then what's the source/significance of the remote labeled `composer`? (The other one being `origin`.)

> In this case, resetting the remote url on each composer operation actually makes things harder for the developer

It only resets the remotes once, when the repository-location (not the version) changes. If that happens, it usually implies a significant shift in development, such as a developer editing their local `composer.json` to point to their own version of the repo. 

When that happens, I'd argue it's safer to refresh the remotes than to leave them at historical values.

Seldaek:
Closing this as I don't think it's an appropriate fix, let's discuss again in your issue.

PR #2898, opened 2014-04-10
[WIP] Extension installer

naderman:
Supports PHP & HHVM.

Todo:
- [ ] Need to come up with a solution for ext-\* packages actually being installable now, so extensions can be compiled automatically to satisfy dependencies on extensions.
- [ ] Documentation
- [ ] Use a different downloader to download binary builds from github releases for windows (or pecl for PHP)
- [ ] Output information on configuring the server to include extension.so

Replaces the earlier PR #498

PR #3897, opened 2015-04-01
draft support for Pickle in Composer

pierrejoye:
Draft support for Pickle (https://github.com/FriendsOfPHP/pickle) in Composer.

Spent more time to dig composer code than actually implement that but at least it works to install exts from within a project.

Example project and extension composer.json:
https://gist.github.com/pierrejoye/1bbc2bdcec5887e60490

If you like to test with more extensions:
git  clone myext
cd myext
pickle convert

Add the dep to your project. If you do not want to fork an ext, you can create a dep vs a zip file and use the artifact repository, it should work.

Still many things to solve but at least there is something to play with now

Update:

According to our discussions, it is now a two steps process to install extensions, to make sure users understand that installing extensions may affect many projects:

First, the usual command:
composer install

composer will tell you to run
composer install --install-extensions

if the project has extensions dependencies

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -345,6 +345,7 @@ class ClassLoader
     {
         // PSR-4 lookup
         $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR) . $ext;
+
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {
             foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -345,7 +345,6 @@ class ClassLoader
     {
         // PSR-4 lookup
         $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR) . $ext;
-
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {
             foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -125,8 +125,8 @@ EOT
             ->setIgnorePlatformRequirements($input->getOption('ignore-platform-reqs'))
 			->setInstallExtensions($input->getOption('install-extensions'));
         ;
-
-        if ($input->getOption('no-plugins')) {
+		
+		if ($input->getOption('no-plugins')) {
             $install->disablePlugins();
         }
 
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -47,6 +47,7 @@ class InstallCommand extends Command
                 new InputOption('verbose', 'v|vv|vvv', InputOption::VALUE_NONE, 'Shows more details including new commits pulled in when updating packages.'),
                 new InputOption('optimize-autoloader', 'o', InputOption::VALUE_NONE, 'Optimize autoloader during autoloader dump'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore platform requirements (php & ext- packages).'),
+				new InputOption('install-extensions', null, InputOption::VALUE_NONE, 'install PHP extensions'),
                 new InputArgument('packages', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, 'Should not be provided, use composer require instead to add a given package to composer.json.'),
             ))
             ->setHelp(<<<EOT
@@ -122,6 +123,7 @@ EOT
             ->setRunScripts(!$input->getOption('no-scripts'))
             ->setOptimizeAutoloader($optimize)
             ->setIgnorePlatformRequirements($input->getOption('ignore-platform-reqs'))
+			->setInstallExtensions($input->getOption('install-extensions'));
         ;
 
         if ($input->getOption('no-plugins')) {
--- a/src/Composer/DependencyResolver/RuleSetGenerator.php
+++ b/src/Composer/DependencyResolver/RuleSetGenerator.php
@@ -179,7 +179,7 @@ class RuleSetGenerator
         }
     }
 
-    protected function addRulesForPackage(PackageInterface $package, $ignorePlatformReqs, $installExtensions = false)
+    protected function addRulesForPackage(PackageInterface $package, $ignorePlatformReqs)
     {
         $workQueue = new \SplQueue;
         $workQueue->enqueue($package);
@@ -189,10 +189,6 @@ class RuleSetGenerator
             if (isset($this->addedMap[$package->id])) {
                 continue;
             }
-
-			if (!$installExtensions && $package->getType() == 'extension') {
-				continue;
-			}
 
             $this->addedMap[$package->id] = true;
 
--- a/src/Composer/DependencyResolver/RuleSetGenerator.php
+++ b/src/Composer/DependencyResolver/RuleSetGenerator.php
@@ -179,7 +179,7 @@ class RuleSetGenerator
         }
     }
 
-    protected function addRulesForPackage(PackageInterface $package, $ignorePlatformReqs)
+    protected function addRulesForPackage(PackageInterface $package, $ignorePlatformReqs, $installExtensions = false)
     {
         $workQueue = new \SplQueue;
         $workQueue->enqueue($package);
@@ -189,6 +189,10 @@ class RuleSetGenerator
             if (isset($this->addedMap[$package->id])) {
                 continue;
             }
+
+			if (!$installExtensions && $package->getType() == 'extension') {
+				continue;
+			}
 
             $this->addedMap[$package->id] = true;
 
@@ -311,7 +315,7 @@ class RuleSetGenerator
         }
     }
 
-    public function getRulesFor($jobs, $installedMap, $ignorePlatformReqs = false)
+    public function getRulesFor($jobs, $installedMap, $ignorePlatformReqs = false, $installExtensions = false)
     {
         $this->jobs = $jobs;
         $this->rules = new RuleSet;
@@ -327,7 +331,7 @@ class RuleSetGenerator
 
         $this->addedMap = array();
         foreach ($this->installedMap as $package) {
-            $this->addRulesForPackage($package, $ignorePlatformReqs);
+            $this->addRulesForPackage($package, $ignorePlatformReqs, $installExtensions);
         }
 
         $this->addRulesForJobs($ignorePlatformReqs);
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -108,6 +108,7 @@ class Installer
     protected $dumpAutoloader = true;
     protected $runScripts = true;
     protected $ignorePlatformReqs = false;
+    protected $installExtensions = false;
     protected $preferStable = false;
     protected $preferLowest = false;
     /**
@@ -500,7 +501,7 @@ class Installer
         $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
         $solver = new Solver($policy, $pool, $installedRepo);
         try {
-            $operations = $solver->solve($request, $this->ignorePlatformReqs);
+            $operations = $solver->solve($request, $this->ignorePlatformReqs, $this->installExtensions);
             $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::POST_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request, $operations);
         } catch (SolverProblemsException $e) {
             $this->io->writeError('<error>Your requirements could not be resolved to an installable set of packages.</error>');
@@ -1249,6 +1250,19 @@ class Installer
     public function setIgnorePlatformRequirements($ignorePlatformReqs = false)
     {
         $this->ignorePlatformReqs = (boolean) $ignorePlatformReqs;
+
+        return $this;
+    }
+
+    /**
+     * set install Extensions flag
+     *
+     * @param  boolean   $installExtensions
+     * @return Installer
+     */
+    public function setInstallExtensions($installExtensions = false)
+    {
+        $this->installExtensions = (boolean) $installExtensions;
 
         return $this;
     }
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -527,6 +527,7 @@ class Installer
             if ('install' === $operation->getJobType()) {
 				if (!$this->installExtensions && $operation->getPackage()->getType() ==  'extension') {
 					$this->io->writeError('<warning>This package has extensions dependencies, run composer install --install-extensons</warning>');
+					continue;
 				}
                 foreach ($operation->getPackage()->getSuggests() as $target => $reason) {
                     $this->suggestedPackages[] = array(
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -489,7 +489,7 @@ class Installer
                 $links = $this->package->getRequires();
             }
 
-            foreach ($links as $link) {
+            foreach ($links as $link) {				
                 $request->install($link->getTarget(), $link->getConstraint());
             }
         }
@@ -500,8 +500,9 @@ class Installer
         // solve dependencies
         $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
         $solver = new Solver($policy, $pool, $installedRepo);
+
         try {
-            $operations = $solver->solve($request, $this->ignorePlatformReqs, $this->installExtensions);
+            $operations = $solver->solve($request, $this->ignorePlatformReqs);
             $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::POST_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request, $operations);
         } catch (SolverProblemsException $e) {
             $this->io->writeError('<error>Your requirements could not be resolved to an installable set of packages.</error>');
@@ -524,6 +525,9 @@ class Installer
         foreach ($operations as $operation) {
             // collect suggestions
             if ('install' === $operation->getJobType()) {
+				if (!$this->installExtensions && $operation->getPackage()->getType() ==  'extension') {
+					$this->io->writeError('<warning>This package has extensions dependencies, run composer install --install-extensons</warning>');
+				}
                 foreach ($operation->getPackage()->getSuggests() as $target => $reason) {
                     $this->suggestedPackages[] = array(
                         'source' => $operation->getPackage()->getPrettyName(),
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -489,7 +489,7 @@ class Installer
                 $links = $this->package->getRequires();
             }
 
-            foreach ($links as $link) {				
+            foreach ($links as $link) {             
                 $request->install($link->getTarget(), $link->getConstraint());
             }
         }
@@ -525,10 +525,10 @@ class Installer
         foreach ($operations as $operation) {
             // collect suggestions
             if ('install' === $operation->getJobType()) {
-				if (!$this->installExtensions && $operation->getPackage()->getType() ==  'extension') {
-					$this->io->writeError('<warning>This package has extensions dependencies, run composer install --install-extensons</warning>');
-					continue;
-				}
+                if (!$this->installExtensions && $operation->getPackage()->getType() ==  'extension') {
+                    $this->io->writeError('<warning>This package has extensions dependencies, run composer install --install-extensons</warning>');
+                    continue;
+                }
                 foreach ($operation->getPackage()->getSuggests() as $target => $reason) {
                     $this->suggestedPackages[] = array(
                         'source' => $operation->getPackage()->getPrettyName(),
--- a/src/Composer/Installer/ExtensionInstaller.php
+++ b/src/Composer/Installer/ExtensionInstaller.php
@@ -1,0 +1,163 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Pierre Joye <pierre.php@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Installer;
+
+use Composer\Composer;
+use Composer\IO\IOInterface;
+use Composer\Repository\InstalledRepositoryInterface;
+use Composer\Package\PackageInterface;
+use Composer\Util\Filesystem;
+use Composer\Util\ProcessExecutor;
+
+/**
+ * Extension installation manager.
+ *
+ * @author Pierre Joye <pierre.php@gmail.com>
+ */
+class ExtensionInstaller implements InstallerInterface
+{
+    protected $composer;
+    protected $vendorDir;
+    protected $binDir;
+    protected $downloadManager;
+    protected $io;
+    protected $type;
+    protected $filesystem;
+	protected $pickle;
+
+    /**
+     * Initializes library installer.
+     *
+     * @param IOInterface $io
+     * @param Composer    $composer
+     * @param string      $type
+     * @param Filesystem  $filesystem
+     */
+    public function __construct(IOInterface $io, Composer $composer, $type = 'extension', Filesystem $filesystem = null)
+    {
+        $this->composer = $composer;
+        $this->downloadManager = $composer->getDownloadManager();
+        $this->io = $io;
+        $this->type = $type;
+
+        $this->filesystem = $filesystem ?: new Filesystem();
+        $this->vendorDir = rtrim($composer->getConfig()->get('vendor-dir'), '/');
+        $this->binDir = rtrim($composer->getConfig()->get('bin-dir'), '/');
+        $this->pickle = "/home/pierre/projects/pickle/pickle/bin/pickle";
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function supports($packageType)
+    {
+        return $packageType === $this->type || null === $this->type;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function isInstalled(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {
+		return false;
+        //return $repo->hasPackage($package) && is_readable($this->getInstallPath($package));
+    }
+
+    /**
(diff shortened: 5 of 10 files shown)


src/Composer/Autoload/ClassLoader.php
```diff
@@ -345,7 +345,6 @@ private function findFileWithExtension($class, $ext)
     {
         // PSR-4 lookup
         $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR) . $ext;
-
```
stof: should be reverted

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return $packageType === $this->type || null === $this->type;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function isInstalled(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {
+		return false;
```
stof: please use spaces for the indentation everywhere (you can run PHP-CS-Fixer on your file to fix most things)

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        $this->composer = $composer;
+        $this->downloadManager = $composer->getDownloadManager();
+        $this->io = $io;
+        $this->type = $type;
+
+        $this->filesystem = $filesystem ?: new Filesystem();
+        $this->vendorDir = rtrim($composer->getConfig()->get('vendor-dir'), '/');
+        $this->binDir = rtrim($composer->getConfig()->get('bin-dir'), '/');
+        $this->pickle = "/home/pierre/projects/pickle/pickle/bin/pickle";
```
stof: a much saner default would be to expect pickle to be in the PATH

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+		/* Add interactions */
+		$cmd = sprintf('%s install -q -n --save-logs=%s %s', $this->pickle, ProcessExecutor::escape($pkg_dir . 'logs'), ProcessExecutor::escape($pkg_dir));
+		$process->execute($cmd);
+		return;
+    }
+
+	protected function createTempDir()
+	{
+		$tmpdir = sys_get_temp_dir();
+		if (!$tmpdir) {
```
stof: you should try creating a temp folder inside the composer cache in case the system temp dir cannot be used, before failing hard

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function install(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {		
+		$this->io->write("Pickle: fetching " . $package->getName());
+
+		$dist_url = $package->getDistURL();
+		if (strtolower(substr($dist_url, -3)) == "zip") {
```
stof: why not relying on the ZipDownloader to download the zip as well ? This would avoid the hard dependency on ext-zip (the ZipDownload is able to use the ``unzip`` command as well, and this is even faster than unzipping in PHP IIRC)

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function install(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {		
+		$this->io->write("Pickle: fetching " . $package->getName());
+
+		$dist_url = $package->getDistURL();
```
stof: variable names should be camelCased

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+		if (strtolower(substr($dist_url, -3)) == "zip") {
+			$extract_dir = $this->uncompress($package);
+			$json = $this->findComposerJson($extract_dir);
+		} else {
+			$pkg_dir = $this->createTempDir();
+			$this->downloadManager->download($package, $pkg_dir);
+		}
+		$process = new ProcessExecutor($this->io);
+		/* Add interactions */
+		$cmd = sprintf('%s install -q -n --save-logs=%s %s', $this->pickle, ProcessExecutor::escape($pkg_dir . 'logs'), ProcessExecutor::escape($pkg_dir));
```
stof: the path to pickle should be escaped too, in case there are spaces in the path

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+     * {@inheritDoc}
+     */
+    public function install(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {		
+		$this->io->write("Pickle: fetching " . $package->getName());
+
+		$dist_url = $package->getDistURL();
+		if (strtolower(substr($dist_url, -3)) == "zip") {
+			$extract_dir = $this->uncompress($package);
+			$json = $this->findComposerJson($extract_dir);
```
stof: ``$json`` is unused

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+        $this->binDir = rtrim($composer->getConfig()->get('bin-dir'), '/');
+        $this->pickle = "/home/pierre/projects/pickle/pickle/bin/pickle";
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function supports($packageType)
+    {
+        return $packageType === $this->type || null === $this->type;
```
stof: this is wrong. It should not support ``null`` here. We don't want it to handle every package but only extensions

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,163 @@
...
+     * @param Composer    $composer
+     * @param string      $type
+     * @param Filesystem  $filesystem
+     */
+    public function __construct(IOInterface $io, Composer $composer, $type = 'extension', Filesystem $filesystem = null)
+    {
+        $this->composer = $composer;
+        $this->downloadManager = $composer->getDownloadManager();
+        $this->io = $io;
+        $this->type = $type;
```
stof: does it actually make sense to have a configurable type here ? The class can only handle extensions anyway

src/Composer/Installer/ExtensionInstaller.php
stof: And the installer needs to handle the case where pickle is not available

src/Composer/Installer/ExtensionInstaller.php
pierrejoye: yes, just did not find (or look too much) the API to get it :) hint?

src/Composer/Installer/ExtensionInstaller.php
pierrejoye: good idea, will do it later

pierrejoye:
Made most of the changes you pointed at, except the tmp dir one and the zip downloader

naderman:
I'm really not sure this is a good idea. We discussed in https://github.com/composer/composer/pull/2898#issuecomment-46298594 why having an ExtensionInstaller is a bad idea, and outlined how this should be implemented instead.

The ExtensionInstaller installs dependencies globally which is very different from how the rest of composer behaves. It does not follow the per-project behavior of composer. If you use the ExtensionInstaller in one project it impacts all other projects on the same machine too, but you will never be informed that you broke the dependencies of the other projects. This needs to be handled much more explicitly and transparent for the user, as is outlined in the comment I linked to.

pierrejoye:
@naderman huh.
We discussed that intensively and you agreed on that. Composer integration was a must have. And one of the ground reason why I started pickle in the 1st place.

So what do you suggest now?

naderman:
Yes I do want it to be integrated, just in the way outlined in that other PR, not as an extension installer.

src/Composer/Installer/ExtensionInstaller.php
stof: well, the logic is already implemented for other formats than zip. See the ``else`` clause. I don't understand why this logic is not used all the time

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,168 @@
...
+ */
+class ExtensionInstaller implements InstallerInterface
+{
+    protected $composer;
+    protected $downloadManager;
+    protected $io;
+    protected $pickle = 'pickle';
+    protected $process;
+    /**
+     * Initializes library installer.
```
stof: this is not the library installer

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * @param IOInterface $io
+     * @param Composer    $composer
+     * @param string      $type
+     * @param Filesystem  $filesystem
```
stof: wrong phpdoc as you removed arguments

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,168 @@
...
+        $this->io = $io;
+
+        if (($pickle = getenv('COMPOSER_PICKLE_PATH'))) {
+            $this->pickle = escapeshellcmd($pickle);
+        }
+        $this->process = new ProcessExecutor($this->io);
+
+        $res = $this->process->execute($this->pickle . ' --version', $output);
+        if ($res != 0) {
+            Throw new \ErrorException("Error while calling pickle command: $res");
```
stof: this should be delayed until the first usage of the installer. Otherwise you break composer for anyone not having pickle installed, even if they never try to install an extension

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,168 @@
...
+        }
+        $verpos = strpos($output, 'version');
+        $version = trim(substr($output, $verpos + 7));
+        if (!version_compare('0.4.0', $version, '>=')) {
+            Throw new \ErrorException("pickle 0.4.0 required.");
```
stof: ``throw`` should be lowercased.

And the message should say ``pickle >= 0.4.0`` IMO

src/Composer/Installer/ExtensionInstaller.php
stof: it is available in the Config object, same than the vendor-dir

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,168 @@
...
+     * {@inheritDoc}
+     */
+    public function install(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {       
+        $this->io->write("Pickle: fetching " . $package->getName());
+
+        $distUrl = $package->getDistURL();
+        if (strtolower(substr($distUrl, -3)) == "zip") {
+            $extractDir = $this->uncompress($package);
+            $json = $this->findComposerJson($extractDir);
```
stof: ``$json`` is still unused (and so ``findComposerJson`` is useless)

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,168 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Pierre Joye <pierre.php@gmail.com>
```
romqin: Should it not be like others files?
```
```

stof: it should

pierrejoye:
@naderman Ok, after discussions on IRC, makes sense for now. I will try to post a summary :)

Will update the patch accordingly, this PR was a draft to start this exact discussion, so no worry. I am not sure where to start to skip extension during install step 1 and skip non extension during install step 2, somehow an ExtensionInstaller is still required, but do not have to be called if '--install-extensions' is not used. Or?

pierrejoye:
Some start for the solver:

https://gist.github.com/pierrejoye/44e8a641ae46b4826487

Is it the right place to begin with? From a logic point of view?

@naderman @Seldaek

pierrejoye:
Could not resist, 2nd attempt (latest commit) sounds clear, no API change, skip being done at the right place. Would be nice if I can put the warning message about having to call "install --install-extension" at the end, so the user can read it nicely, ideas?

src/Composer/DependencyResolver/RuleSetGenerator.php
```diff
@@ -179,7 +179,7 @@ protected function whitelistFromPackage(PackageInterface $package)
         }
     }
 
-    protected function addRulesForPackage(PackageInterface $package, $ignorePlatformReqs)
+    protected function addRulesForPackage(PackageInterface $package, $ignorePlatformReqs, $installExtensions = false)
```
naderman: This is the wrong place for this. You still need to resolve all dependencies to correctly figure out which dependencies you need. You should simply pick which parts to install later on based on the solver's generated transaction. There should not be a need to modify the solver.

pierrejoye: I removed that in the last change. See latest commits

pierrejoye: Ah no, I removed the ones in the solver.

Then I have no idea where you want this check. Please tell it. As it is mixed up already (like platform).

src/Composer/Installer.php
```diff
@@ -523,6 +525,9 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
         foreach ($operations as $operation) {
             // collect suggestions
             if ('install' === $operation->getJobType()) {
+				if (!$this->installExtensions && $operation->getPackage()->getType() ==  'extension') {
+					$this->io->writeError('<warning>This package has extensions dependencies, run composer install --install-extensons</warning>');
+				}
                 foreach ($operation->getPackage()->getSuggests() as $target => $reason) {
```
pierrejoye: @nadernan this is actually where it skips extension. I suppose the part in the other file is not required anymore, right? Can test again tomorrow but advice's welcome

pierrejoye: So no, it is still required.

pierrejoye:
Ok, my mistake, forgot to add continue; in InstallationManager. Now it works as expected.

What do you think? Right way?

pierrejoye:
Only thing I like to change for now is to display the "please run composer install --install-extensions" at the end of the 1st install call, to be sure users see it. Any APIs for that? or does it require some extra code in the Command/InstallerCommand directly (sounds double :)?

pierrejoye:
One thing I am not sure about is the ext-prefix.

What is the technical limitation to force it?

Also we have a huge problem with core extensions, even more for extensions available in core and pecl. Versioning there is just a big mess.

What do you think to make special cases for core extensions and handle the version based on the php release version, for example mysqli-5.6.7?

Seldaek:
The ext- prefix IMO is just for consistency and clear communication what package is an extension since those are a bit special things anyway.

And yes we could have special rules for core extensions that don't have proper versions in <7, but for php7+ it'd be nice to clean things up in php directly IMO. For pecl it'd be nice if they could be cleaned up at least get a new correct version for php7 or something. People just have to agree to get better at this if we want a workable ecosystem.

pierrejoye:
pickle enforces good version and so will do the web app, a must.

For the core, I think we should just use php versions instead, total mess
and I do not see us fix that any time soon, thoughts?

About the patch? Better?

pierrejoye:
Ping.

localheinz:
Interesting!

stof:
@pierrejoye the issue is that it would force Composer to hardcode a list of extensions provided by the core (and it could still be wrong in case of package managers bundling them separately, even though they might use strict deps to prevent this). It would be better to have core extensions being updated to expose the PHP version as their own version too.

src/Composer/Command/InstallCommand.php
```diff
@@ -47,6 +47,7 @@ protected function configure()
                 new InputOption('verbose', 'v|vv|vvv', InputOption::VALUE_NONE, 'Shows more details including new commits pulled in when updating packages.'),
                 new InputOption('optimize-autoloader', 'o', InputOption::VALUE_NONE, 'Optimize autoloader during autoloader dump'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore platform requirements (php & ext- packages).'),
+				new InputOption('install-extensions', null, InputOption::VALUE_NONE, 'install PHP extensions'),
```
stof: wrong indentation (in all lines changed in this file btw)

src/Composer/Installer.php
```diff
@@ -488,7 +489,7 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
                 $links = $this->package->getRequires();
             }
 
-            foreach ($links as $link) {
+            foreach ($links as $link) {
```
stof: you should remove these trailing whitespaces

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function install(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {
+
+        $this->checkPickleCmd();
+        $this->io->write("Pickle: fetching " . $package->getPrettyName());
```
stof: IMO, this should not be written at normal verbosity.

and it should be written to STDERR btw, not STDOUT

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,172 @@
...
+
+        $distUrl = $package->getDistURL();
+        if (strtolower(substr($distUrl, -3)) == "zip") {
+            $extractDir = $this->uncompress($package);
+            $json = $this->findComposerJson($extractDir);
```
stof: ``$json`` is unused. Why searching the composer.json here ?

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,172 @@
...
+     * {@inheritDoc}
+     */
+    public function install(InstalledRepositoryInterface $repo, PackageInterface $package)
+    {
+
+        $this->checkPickleCmd();
+        $this->io->write("Pickle: fetching " . $package->getPrettyName());
+
+        $distUrl = $package->getDistURL();
+        if (strtolower(substr($distUrl, -3)) == "zip") {
```
stof: same question than in the initial implementation: why doing a special logic for zip archives instead of letting the DownloadManager handle them like any other archives ?

src/bootstrap.php
```diff
@@ -21,5 +21,4 @@ function includeIfExists($file)
         'php composer.phar install'.PHP_EOL;
     exit(1);
 }
-
```
stof: should be reverted

stof:
how are available extensions known by Composer ? Does pecl.php.net support the Composer protocol to expose packages ? Does it require a special repository implementation ? Or does it need to wait for pickle-web ?

pierrejoye:
No, it won't. Or I don't see why composer should do it. Mind to enlighten
me?

pierrejoye:
@stof yes, see http://news.php.net/php.internals/85814

and we need a list anyway. Which one is bundled in which major version. They have to be handled differently.

pierrejoye:
Also it is not composer's job to keep that list nor to deal with but pickle, composer does nothing about extensions but check if they are loaded.

slbmeh:
How far off would a PECL repository be from the PEAR repository?

pierrejoye:
@slbmeh  I do not get the question but let me try to answer based on a guess :)

Pickle does not use pear, at all. It does not need package.xml but can handle package with only package.xml but no composer.json (live conversion).

pierrejoye:
Also, I would *love* feedback about the patch (not CS pls, I will fix them) and if it is how you like to have it.

I can go with a plugin instead for the meantime, as this is a blocker for now.

slbmeh:
@pierrejoye sorry I had the context in my head. That was a question building off of @stof's question about discovering extensions in regards to a special repository. I believe the PECL and PEAR repositories are rather similar and may be able to work with the existing PEAR stuff.

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,172 @@
...
+
+        /* Add interactions */
+        $cmd = sprintf('%s install -q -n --save-logs=%s %s', $this->pickle, ProcessExecutor::escape($pkgDir . 'logs'), ProcessExecutor::escape($pkgDir));
+        $this->process->execute($cmd);
+        return;
+    }
+
+    protected function createTempDir()
+    {
+        $lockfile = tempnam($this->cacheDir, 'pickle');
```
slbmeh: Wouldn't this create a file every time you generate a new directory name?

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,172 @@
...
+        return $lockfile . '_dir';
+    }
+
+    protected function uncompress(PackageInterface $package)
+    {
+        $extractDir = $this->createTempDir();
```
slbmeh: If this is intended to be temporary, should there be some attempt to clean it up?

src/Composer/Installer/ExtensionInstaller.php
```diff
@@ -0,0 +1,172 @@
...
+            $this->downloadManager->download($package, $pkgDir);
+        }
+
+        /* Add interactions */
+        $cmd = sprintf('%s install -q -n --save-logs=%s %s', $this->pickle, ProcessExecutor::escape($pkgDir . 'logs'), ProcessExecutor::escape($pkgDir));
+        $this->process->execute($cmd);
+        return;
+    }
+
+    protected function createTempDir()
```
slbmeh: This name is somewhat deceiving because it doesn't actually create a directory.

pierrejoye:
@slbmeh please check what Pickle is, it could be a good start.

And no, we are not going to use pear for that, that's exactly the goal, to do not use pear anymore as it is a pain to use as an extension developer.

src/Composer/Installer/ExtensionInstaller.php
pierrejoye: yes, it should :)

Once we agree on how and where it is done, I will spend more time on making this PR more than a draft. I sadly do not have the time/resources to do 20 attempts. I would rather agree on a concept and finalize it instead of finalizing something that does not match composer leads goals or ideas about that.

src/Composer/Installer/ExtensionInstaller.php
pierrejoye: it actually does, but it is a protected method anyway. :) Same comment as previous one tho'

slbmeh:
@pierrejoye I'm aware of Pickle. I'm not suggesting anything about Pickle itself. I thought the question was about how will composer know about extensions. My reference to pear was about https://github.com/composer/composer/blob/master/src/Composer/Repository/PearRepository.php, I don't think it's too far of a stretch to make that work with pecl.php.net so you don't have to duplicate metadata.

stof:
@pierrejoye the PlatformRepository providing the info about the platform itself needs to be able to know the version of installed extensions, otherwise you cannot use the version in the constraint.
so if PHP extensions don't expose their version properly, we would have to add such knowledge in Composer itself to reuse the PHP version

src/Composer/Installer/ExtensionInstaller.php
stof: see my previous comments. The ``uncompress`` method should probably disappear entirely

stof:
@slbmeh the PearRepository is a huge performance killer, because the PEAR protocol is just not adapter for Composer needs, required a huge number of requests to fetch the necessary data.
This is why Pirum introduced first-class composer support for the pear repositories it generates, to avoid having to use the PearRepository with them.
If we want to make extensions installable with Composer, the extension registry should expose its metadata in an efficient format for Composer (the best idea being to use the composer protocol, used by Packagist and Satis, as this means that Composer already has the support for loading the metadata without doing another implementation).

pierrejoye:
OK, sorry but still no :)

pecl is by far not the place anymore where ppl releases their extensions.
That is why we need pickle.

And pickleweb (packagist for extensions) instead. Bringing the simplicity
of composer to ext developers when it comes to releases, and a good ext
index for the users.

pierrejoye:
@slbmeh pickle works with pecl.php.net transparently and out of th box, even if a release has no composer.json.

Same for the web hooks on registration or release tag.

@stof pickle ensures that the version is set correctly. It cannot be released if not. I also started a discussion on internals to fix the core extensions versions.

pierrejoye:
@stof Will look at the PlaformInfo part, it could be possible to get it to know the installed extensions. I have to check to do not duplicate the installed version management. Not sure how yet.

Waiting on @Seldaek and @naderman to comment on this patch, where it is done and how, then I will investigate further about these things.

pierrejoye:
A couple of things we support in pickle and not sure how to make them work smoothly when called from composer.

Old releases of an extension may not have composer.json (or will most not have), composer will fail on these release. What pickle does is to check for package(2).xml and convert on the fly to generate a usable composer.json. Absence of package.xml will be supported as well, composer.json being created from the src (README, header, etc). It is already partially done for the convert command.

Ideas for the composer part? Or we may simply require it when used from composer, that will help extensions developers to get to add composer.json.

Seldaek:
Well composer doesn't really need the composer.json if the pickle repo provides the metadata. It is needed however if we want to load a github repo of an ext as a vcs repository in composer, so yes if people add it eventually it'd be great.

pierrejoye:
We will enforce it for any new release. But the question is for older
releases, they have none. If composer called pickle for every ext
dependency directly without checking anything, then it is easy.

If not, and the repo (source, VCS, whatever) does not have composer, it may
fail or I do not see how composer can get all the metas.

It also brings me back to the main point. What do you think about the draft
patch? Where it is hooked and so on? :/ about to open beta for the
pickleweb, would be great if we can get composer supporting it. If not, no
big deal but the gap will get bigger.

src/Composer/Installer.php
```diff
@@ -523,6 +525,10 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
         foreach ($operations as $operation) {
             // collect suggestions
             if ('install' === $operation->getJobType()) {
+                if (!$this->installExtensions && $operation->getPackage()->getType() ==  'extension') {
+                    $this->io->writeError('<warning>This package has extensions dependencies, run composer install --install-extensons</warning>');
+                    continue;
+                }
```
Seldaek: Should maybe support 'update' operations too

pierrejoye: If an extension is loaded it cannot be updated. That's the tricky part. It
is also why we somehow need to be able to run pickle against a different
php than the one running pickle. Or use -n but then... :)

Seldaek: Ah right, this kinda sucks though

flip111:
any news?

gabrielrcouto:
@Seldaek Do you need any help to test and close this PR?

faizanakram99:
bump

Seldaek:
Closing to clean up here.. We'll see if https://github.com/php/pie becomes more viable perhaps we can integrate it at some point.

Issue #3887, opened 2015-03-29
Feature request: composer --need-self-update


ross-p:
It would be nice to have a programmatic way to know if composer is outdated and in need of a self-update.

For example I'd like to be able to run some code like this:

```bash
if composer --need-self-update; then
  composer self-update
fi
```

Or maybe something like

```bash
if test $(composer --days-since-self-update) -gte 30; then
  composer self-update
fi
```

Basically I don't want to be forced to `composer self-update` all the time.  But when it gets time that composer itself starts complaining that it should be updated, I'd like to update it then.

jakoch:
Hey Ross,

Maybe the following script fits your intention? 

`getComposerWarningTime.php`
```
<?php
$p = new Phar('composer.phar', 0, 'composer.phar');
$stubString = $p->getStub();
preg_match("#COMPOSER_DEV_WARNING_TIME\'\,\s(\d+)#", $stubString, $matches);
echo $matches[1]; // timeout
```
The returned value is the compilation time of the Composer.phar plus 30 days.
Composer is outdated, when the timeout value is lower than current time (`if (time() > $timeout)...`).
You might retrieve the value in a bash script and do the comparison.

ross-p:
@jakoch Thanks for the suggestion.  I am doing something similar:

```bash
if ! test $(grep --text COMPOSER_DEV_WARNING_TIME /path/to/composer | egrep '^\s*define' | sed -e 's,[^[:digit:]],,g') \> $(php -r 'echo time();'); then
  composer self-update
fi
```

Source: https://github.com/saltstack-formulas/php-formula/blob/master/php/composer.sls#L31

While that is working, I fear that composer may change the formatting of their code or the name of the constant or who knows what, which will then break any/all attempts such as yours and mine to grep this feature out of the code.

An explicit export of the data would be very welcome.

jakoch:
I think we need to discuss the return value of the command.
It has to be consistent (also in regard to the install --needed request).

In your first script you are using the command for testing based on the exit code, where
Exit code 0 (success) means update needed and 1 (error) means no update needed, right?
Code 0 equals an empty string `""` and the condition evals to true if string is empty.

Should 1 indicate the need to update? (string comparison required)
Should this be a short or full sentence output? (string comparison or match/grep required)
What's better from your perspective?

ross-p:
It's an interesting question.

[...]

PR #3899, opened 2015-04-01
added "self-update --if-outdated", issue #3887

jakoch:
also added the missing doc for option "no-progress"


stof:
The name is weird. If there is a newer version, your version is outdated, even if it is not older than the threshold making Composer complain about it

jakoch:
Please suggest a better one.

PR #3902, opened 2015-04-02
Readme: 00-intro - use __DIR__

TomasVotruba:
It is best practice to use `__DIR__` constant. Also it's used in `autoload.php` itself.


fprochazka:
:+1:

TomasVotruba:
Thanks @Seldaek

Issue #3658, opened 2015-01-20
Directory repository type and local stability


yguedidi:
First of all, thank you for composer!

But currently it's a pain to work localy on packages that depend on each other. You must run `composer update` each time a dependency of a package you work on is updated.

That's why I propose to have a repository type `directory`, with a `path` option. And they have only the `local` stability. If the directory exists locally, then composer create a symlink to that directory in vendors.

#### Exemple 1:
Here the repo must be defined in `~/.composer/config.json`, but this allow the contributor to put it where he want on his machine (usage between windows and linux)
```json
// ~/.composer/config.json
{
    "repositories": [
        {
            "type": "directory",
            "path": "/path/to/my/local/repo"
        }
    ]
}
```
```json
// /myrepo/composer.json
{
    "require": "myother/repo@local"
}
```
An error is throwed when the local repository is not available.

#### Exemple 2:
Here the required repo **must** be at the same level of the current one, in a `repo2` directory.

```json
// /myrepo/composer.json
{
    "repositories": [
        {
            "type": "directory",
            "path": "../repo2"
        }
    ],
    "require": "myother/repo@local"
}
```
---
I hope I'm understandable... Any thoughts?

alcohol:
Composer already supports (local) directories as repository sources. These however are used as actual repositories.

Your issue is a common one and could easily be resolved by simply creating symlinks yourself for example.

See also related issues: #601, #1299, #2171, #3254 (and probably more)

torches:
@yguedidi I agree, this is a pain, on one hand composer allows for rapid development with what it provides, but if you want to continue that rapid development, while working in a packaged module structure (keeping your in house modules split out) it becomes a cumbersome effort.
There should be an easy way to define the local path to your code without overriding the autoloader inside your front controller, or using symlinks.
I think a drop in file as suggested above that would allow you to specify the local static path of your module namespace makes perfect sense.
A good option for some is to have an "app" which has all of your modules included and their namespaces defined, but this is only practical for some projects.

+1 on this feature.

jreinke:
Just gave a solution for this on another related issue: https://github.com/composer/composer/issues/1299#issuecomment-88861371

Seldaek:
The path repo helps with this so closing, could be we have a better solution eventually but right now that's it https://getcomposer.org/doc/05-repositories.md#path

mreschke:
[...]

Issue #3254, opened 2014-09-02
Skip symlinked vendor packages when updating


noccy80:
To avoid cluttering my git repositories with lots of small changes, I often symlink libraries into the vendors directory of the app that uses them while I'm working on both. This causes problems when doing a `composer update`. It would be great if composer was aware of this, and could take the appropriate action:

- Without `--force` just print out "ignoring symlinked package foo/bar..."
- With `--force`, unlink the symlink and install the package fresh

Seldaek:
The new path repository (https://getcomposer.org/doc/05-repositories.md#path) lets you do this cleaner now.

dantleech:
This would still be good imo. 

I often work on a project and its multiple dependencies simultaneously - and those dependencies are projects in their own right in their own workspaces. So duplicating them and working in `vendor` is just somewhat confusing/frustrating - as I am then working on the same project in two different places. Hence I prefer to symlink.

Seldaek:
@dantleech there is the path repo or https://github.com/franzliedke/studio that can do it cleanly, I'm not really willing to add dirt hacks like ignoring all symlinks.

Issue #2171, opened 2013-08-13
Read a package from a relative local path.


Vittel:
**Add a repository type: dir/file/local that accepts a local relative path as source**

**Let me point out the requirement context a little bit:**
We have a working deployment system up and running, deploying needed code to the exact right versions and positions in file system.
Now we try to enable composer on certain parts of the code that we identified as packages and to resolve further dependencies and autoload the contained classes.
Its not an option for us to break our deployment system, but it shouldn't be necessary when we could just read the prepared package from a local path (where it was deployed).
This is currently not possible. All existing repository configurations expect the repo to be remote or be accessible with a remote URL.
This is only the first step of our composer migration, and we are already stumbling.

jreinke:
Just gave a solution for this on another related issue: https://github.com/composer/composer/issues/1299#issuecomment-88861371

Vittel:
@jreinke even though your solution might work for most people, symlinking files instead of just reading them from their original location is not a solution for this problem.

slbmeh:
I work with a handful of 100+ dependency projects so I'll give my two cents.

I'm not quite sure how well a relative path repository declaration would work in the context of a project. Suppose someone checks out your project and doesn't have the same path structure. It generally works better for large teams and CI to have your repositories declared elsewhere. The file protocol works in some circumstances, but if you're working with a VCS repo you need to commit your work for it to be updated anyhow, why not just push it to a central location on some experimental branch.

With development and testing uncommitted changes we use an alternative composer file that excludes the codependencies from the require section and places their relative paths in `autoload-dev`. This also needs to have the child dependencies to be placed as a direct dependency of the root package.

@laurentj `Unfortunately, It seems there is no way to indicate to Composer: "include in my project this local package, which is in this directory (without copying it somewhere)".`: the approach in the second paragraph does exactly that.

Vittel:
The main advantage here is not the "composition" part (since everything is in the place to be), 
but letting composer generate one autoload-file for all projects that are required, without duplicating or symlinking any of the code into each other project.

just like: forget about the packages, i managed that. but please take care of the autoloading for me

timglabisch:
i am playing with a local repository type.
companies like google and facebook are working this way, so it should scale =)

https://github.com/timglabisch/composer_local_downloader_demo

far-blue:
[...]

Issue #1299, opened 2012-11-07
Symlink local development libraries 


Petah:
Is there any way to simply symlink a vendor library from another directory rather than keeping a local copy for each project?

This is primarily for speed of development. 

We have many in house modules (20-30) and are running a Satis server. 

When we require a change to a module for a project, you must follow this process at present.

 - Make change in module
 - Commit and push change to Git server
 - Update Satis server (currently on a 5 min cron schedule) 
 - Update the project using `composer update --dev`

The turn around on this is quite lengthy, and doing it multiple time for multiple modules/projects is impractical. 

If we could just symlink each of our in house libraries into the vendor directory, turn around time would be nil.

Jeremlicious:
+1

gilbitron:
+1

piwi:
Hello all,
I made a plugin for that (not so amazing but it works fine) : <https://github.com/piwi/composer-symlinker>.
If it helps ...

jreinke:
Hi there,

@piwi solution did not work for me because my local packages are not under VCS.

So I had to modify Composer itself to allow local packages installation, modification is quite simple https://github.com/jreinke/composer/commit/92ad4e4c631d5c611634aa64b4f44ce66cf26491. You have to modify your `composer.phar` like that:
```shell
$ git clone https://github.com/jreinke/composer.git && cd composer/
$ curl -sS https://getcomposer.org/installer | php && php composer.phar install
$ ./bin/compile
```
Now you can use custom `composer.phar` for local packages installation like below:
```json
"require": {
     "<vendor>/<local-package>": "*"
},
"repositories": [
    {
        "type": "folder",
        "url": "../path/to/package/"
    }
]
```
Local packages are installed with symlinks so that any modification made is automatically applied to projects depending on them.

Vittel:
Hey,

just a complain: having symlinks in anything related to source control is not a good idea. Why dont you just read the packages from where they are and not link them in?

jreinke:
I agree, but symlinks are also a good way for developers productivity. Hard copying packages would involve frequent `composer update` commands on all depending packages.
Maybe a good option would be having the choice between hard copy and symlink...

Vittel:
No, i mean do not copy or link them at all. just take the position of the sourcecode as is, and do include the files in their original position in the autoloader

Divi:
If someone can't wait for this feature, I made a plugin to automatically symlink only dev packages : https://github.com/Letudiant/composer-shared-package-plugin

[...]

PR #3904, opened 2015-04-02
Allow local package dependency that is not under VCS

jreinke:
Several issues/requests were reported for such a feature: #1299, #2171, #3254, #3658.

This PR adds a repository that allows installing local packages that are not under VCS.
Just add paths in repositories so the local packages can be found.
The local packages will be added as symlinks, it means that any modification
made in local packages will be automatically applied to other packages depending on it.
Both relative and absolute paths are handled.
```json
"require": {
    "<vendor>/<local-package>": "*"
},
"repositories": [
    {
        "type": "folder",
        "url": "../../relative/path/to/package/"
    },
    {
        "type": "folder",
        "url": "/absolute/path/to/package/"
    }
]
```

### DIFF
--- a/src/Composer/Downloader/FolderDownloader.php
+++ b/src/Composer/Downloader/FolderDownloader.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Downloader;
+
+use Composer\Package\PackageInterface;
+use Composer\Package\Version\VersionParser;
+
+/**
+ * Downloader for folders
+ *
+ * @author Johann Reinke <johann.reinke@gmail.com>
+ */
+class FolderDownloader extends FileDownloader
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function download(PackageInterface $package, $path)
+    {
+        if (is_dir($path)) {
+            rmdir($path);
+        }
+        if (false === symlink($package->getDistUrl(), $path)) {
+            throw new \ErrorException();
+        }
+        $this->io->writeError("  - Installing <info>" . $package->getName() . "</info> (<comment>" . VersionParser::formatVersion($package) . "</comment>)");
+        $this->io->writeError('');
+    }
+}
--- a/src/Composer/Repository/FolderRepository.php
+++ b/src/Composer/Repository/FolderRepository.php
@@ -25,7 +25,7 @@ use Composer\Package\Loader\ArrayLoader;
  * @code
  * "require": {
  *     "<vendor>/<local-package>": "*"
- * }
+ * },
  * "repositories": [
  *     {
  *         "type": "folder",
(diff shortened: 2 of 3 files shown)


oflannabhra:
:+1: This looks amazing for package development. It kind of blows my mind that this hasn't been available in Composer.

TomasVotruba:
:+1:_

smazurov:
any chance to add a fallback? if folder fails, try remote vcs/packagist?

oflannabhra:
Bundler (ruby's dependency manager) also allows for a [local override](http://bundler.io/v1.2/git.html#local) configuration, so that you don't have to touch your Gemfile to point to a local path. I believe it still requires the local gem to be under VCS, however.

Would it be worth adding something similar to Composer?

smazurov:
actually, there is a plugin called [merge](https://github.com/wikimedia/composer-merge-plugin) that handles the use case, so in hopes of getting this merged, I retract my earlier statement.

airways:
Can this get merged? This would be a massive productivity improvement.

far-blue:
+1

testprobe:
+1

sroze:
Definitely useful feature for monolith repositories :+1:

src/Composer/Repository/FolderRepository.php
```diff
@@ -0,0 +1,85 @@
...
+        $file = $this->path . 'composer.json';
+        if (is_file($file) && is_readable($file)) {
+            $json = file_get_contents($file);
+            $package = JsonFile::parseJson($json, $file);
+            $package['dist'] = array(
+                'type' => 'folder',
+                'url' => $this->path,
+            );
+            if (!isset($package['version'])) {
+                $package['version'] = 'dev-master';
```
Seldaek: This could use a version detection mechanism like we have in the RootPackageLoader to determine the version, assuming that directory has VCS data.

src/Composer/Downloader/FolderDownloader.php
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function download(PackageInterface $package, $path)
+    {
+        if (is_dir($path)) {
+            rmdir($path);
+        }
+        if (false === symlink($package->getDistUrl(), $path)) {
+            throw new \ErrorException();
```
Seldaek: It'd be nice to have a fallback that does a dir copy instead of a symlink here I think. You can use https://github.com/symfony/Filesystem/blob/master/Filesystem.php#L374 as a fallback.

mellowplace:
:+1: for this.  Would make development so much simpler.

vmikki:
+1 it would be great to have this

kylesome:
+1 A much needed feature. Would make local package development a lot simpler and faster!

nishsinghal:
+1

webmozart:
Cool! :) From a usage perspective, I'd rather use the term "path" instead of "folder".

carstenwindler:
+1 definitely waiting for this :-)

sroze:
@Seldaek I've created this PR (keeping @jreinke in the author loop) with the recommended changes: https://github.com/composer/composer/issues/4365

nibsirahsieu:
+1

aderuwe:
+1

Seldaek:
Closing as #4365 is now merged and provides a path repository.

apflieger:
For those who use Phpstorm, there's an open feature request to make phpstorm recognize symlinks so it would be no more "multiple definitions" of a same class/file that it symlinked through path repositories

Vote for it !! https://youtrack.jetbrains.com/issue/IDEA-85774

PR #3906, opened 2015-04-03
Update 00-intro.md

edmilsonefs:
Add chmod and sudo on installation steps

Why not add chmod and sudo on installation steps?


Seldaek:
I am pretty sure this was discussed already although I can't find it right now, but sudo is not always available so I don't think it's a good idea to add it by default, as for the chmod.. it is done already by the installer.

PR #3907, opened 2015-04-03
Update 00-intro.md - typo fix

manti:


stof:
See https://github.com/composer/composer/pull/3284, https://github.com/composer/composer/pull/2614 or https://github.com/composer/composer/pull/2030 for previous PRs doing the same change, and rejected each time for the same reason

PR #3917, opened 2015-04-08
Fixes #3900

### LINKED ISSUE #3900: Require command unable to detect latest stable
Composer seems to be unable to detect the latest stable release when running:

```sh
composer require beberlei/DoctrineExtensions
```

But the following does work:

```sh
composer require beberlei/DoctrineExtensions "~1.0"
```

Packagist and other services seems seem happy that `1.0.1` is the latest stable, but the require command isn't happy? Wondering if it's to do with the `v` tag prefix?

https://github.com/beberlei/DoctrineExtensions/issues/145

alcohol:
Also refs beberlei/DoctrineExtensions#145


alcohol:
@Seldaek updated it.

PR #3923, opened 2015-04-09
Silence symlink() error

mathielen:
If not silenced it gives "symlink(): Operation not supported" on my ubuntu system with mounted cifs share.


Seldaek:
Thanks

PR #3936, opened 2015-04-14
Ignore all .git dirs in the vendor dir recursively

ajpevers:
In accordance with http://git-scm.com/docs/gitignore


Seldaek:
Thanks

PR #3885, opened 2015-03-29
updated spdx-license handling to include metadata

jakoch:
issue https://github.com/composer/composer/issues/3744

summary of changes
- replaced bin script "fetch-spdx-identifiers" by "update-spdx-licenses"
- "update-spdx-licenses" uses "Composer/Util/SpdxLicensesUpdater" to update "spdx-licenses.json" file with license identifier, fullname and osi-approved-status
- dropped "spdx-identifiers.json" (identifiers only)
- dropped "SpdxLicenseIdentifier", it's just "SpdxLicense" now
- modified "ShowCommand" to output the license with metadata and removed some unused method arguments (cleanup)

ShowCommand Output Example (license line only):
`license  : MIT License (MIT) (OSI approved) http://spdx.org/licenses/MIT#licenseText`

PR #3939, opened 2015-04-14
Fix coding standards

stof:
This is the result of applying PHP-CS-Fixer at the Symfony level, except the ``concat_without_space`` and ``empty_return`` fixers (because the code currently does not apply these rules) and with the ``concat_with_space`` one (because the code follows this rule).

### DIFF
--- a/src/Composer/Util/SpdxLicense.php
+++ b/src/Composer/Util/SpdxLicense.php
@@ -34,7 +34,7 @@ class SpdxLicense
 
     private function loadLicenses()
     {
-        if(is_array($this->licenses)) {
+        if (is_array($this->licenses)) {
             return $this->licenses;
         }
 
@@ -71,7 +71,7 @@ class SpdxLicense
     public function getIdentifierByName($name)
     {
         foreach ($this->licenses as $identifier => $licenseData) {
-            if($licenseData[0] === $name) { // key 0 = fullname
+            if ($licenseData[0] === $name) { // key 0 = fullname
                 return $identifier;
             }
         }
@@ -89,7 +89,7 @@ class SpdxLicense
 
     /**
      * Check, if the identifier for a license is valid.
-     * 
+     *
      * @param string $identifier
      *
      * @return bool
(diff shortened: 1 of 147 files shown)


src/Composer/Util/SpdxLicense.php
```diff
@@ -1,229 +1,231 @@
-<?php
-
```
stof: the huge diff here is because #3885 used CRLF line endings

stof:
note that I checked all changes performed by the PHP-CS-Fixer tool to ensure they are OK.

Seldaek:
Thanks but this is a doing a lot of useless changes IMO - which might break open PRs or just pollute the history for no good reason. I'll commit the php_cs config I use to run the periodic CS fixes commits I do I guess.

Seldaek:
c.f. https://github.com/composer/composer/commit/fda4a594f904ff0032a4e4b4370b28e7685c31c0

stof:
@Seldaek I think some of the newer fixers might be useful to ensure consistency though.
For instance, aligning phpdoc params looks much nicer when separating them from other phpdoc tags which might be much longer than types.

and then, should we change the contributing.md to stop saying that the library uses the Symfony coding standards, as this is not the case (this PR was about that) ? I'm not even sure the codebase is following PSR-2 currently (I don't remember exactly which rules are from PSR-2 and which ones are from Symfony as I generally don't care about this distinction giiven I'm following the Symfony rules)

stof:
anyway, I suggest adding the ``multiline_array_tailing_comma`` fixer in the config. Having the trailing comma in arrays tends to limit useless diffs later when adding new elements.

Seldaek:
I added that fixer but it doesn't do anything right now so that's good. About the rest yeah I am not sure which to take in.. I think some of it is too greedy/noisy. Adding dots at the end of short descriptions is really not a huge deal. Similarly the phpdoc formatting I don't really care.. only IDEs read this mostly, or nobody at all I am not sure. 

As for contribuding.md, it's not like people follow things strictly anyway ;) But I updated it to say psr-2 as this represents the important parts IMO.

PR #3943, opened 2015-04-16
Skip JSON downloads when no dependencies exist in composer.json

ashnazg:
When composer.json lists zero dependencies, the many JSON files are still downloaded from packagist on each install (since no lockfile will exist).  This patch avoids those downloads in this scenario.

This proves to be a timesaver in apps where composer's sole job is to generate an autoloader (thus, no dependencies are listed).

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -496,10 +496,10 @@ class Installer
         // force dev packages to have the latest links if we update or install from a (potentially new) lock
         $this->processDevPackages($localRepo, $pool, $policy, $repositories, $lockedRepository, $installFromLock, 'force-links');
 
+        $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
         $operations = array();
         if ($this->package->hasDependencies()) {
             // solve dependencies
-            $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
             $solver = new Solver($policy, $pool, $installedRepo);
             try {
                 $operations = $solver->solve($request, $this->ignorePlatformReqs);
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -496,17 +496,20 @@ class Installer
         // force dev packages to have the latest links if we update or install from a (potentially new) lock
         $this->processDevPackages($localRepo, $pool, $policy, $repositories, $lockedRepository, $installFromLock, 'force-links');
 
-        // solve dependencies
-        $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
-        $solver = new Solver($policy, $pool, $installedRepo);
-        try {
-            $operations = $solver->solve($request, $this->ignorePlatformReqs);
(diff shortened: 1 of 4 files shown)


ashnazg:
(this might seem like a minor optimization, until you work on a network with habitually slow interaction with packagist)

src/Composer/Installer.php
```diff
@@ -496,17 +496,20 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
...
-        $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
-        $solver = new Solver($policy, $pool, $installedRepo);
-        try {
-            $operations = $solver->solve($request, $this->ignorePlatformReqs);
-            $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::POST_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request, $operations);
-        } catch (SolverProblemsException $e) {
-            $this->io->writeError('<error>Your requirements could not be resolved to an installable set of packages.</error>');
-            $this->io->writeError($e->getMessage());
+        $operations = array();
+        if ($this->package->hasDependencies()) {
```
stof: this breaks in case a plugin registers more dependencies on the ``PRE_DEPENDENCIES_SOLVING`` event

ashnazg: I'm not familiar with that hook behavior.  Do you think it's just a matter of moving that `eventDispatcher` line to just above my new `if`, thus leaving only the `$solver` activity being controlled by `hasDependencies()`?

src/Composer/Installer.php
```diff
@@ -384,7 +384,7 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
             $pool->addRepository($lockedRepository, $aliases);
         }
 
-        if (!$installFromLock) {
+        if (!$installFromLock && $this->package->hasDependencies()) {
```
stof: are you sure this is needed ?

ashnazg: Without this, a packages.json download from packagist still occurs... that's the one thing I'm trying to have it avoid here.

src/Composer/Installer.php
stof: The check should be changed for several reasons:

- event listeners will add more requirements in the solver request, not in the package (altering the root package could create weird issues, and it is too late for that anyway as the package requirements are added in the request earlier). So the check should be based on the solver request, not on the root package
- using a ``hasDependencies`` method on the package means that you considers there are dependencies even when we run in ``--no-dev`` mode and you only have dev requirements. The check should be based only on the requirements being used by the installer, which implies moving this logic outside the package (this also avoids the need to modify the AliasPackage). Note that if you change the logic to be based on the solver request, this point will be solved automatically as the solver request already takes the dev mode into account.

ashnazg: Are you saying it's feasible that a composer.json that has neither a require nor require-dev section could still pick up some dependencies due to event listener / plugin behavior?

stof: it could (the composer asset plugin for instance)

ashnazg: Ok, then the process sounds much more nuanced than I had discovered just from step-debugging to find where the downloads were occurring.  This also tells me that either the test suite is not covering such a possibility as that asset plugin example, or perhaps my patched code was just getting lucky.

I'll fall back to enforcing a `repositories: packagist: false` section in all composer.json files where zero dependencies are listed, and thus prevent the downloads that way.

Thanks for the review and info, @stof.

ashnazg:
Patch is too naive, in that only considering root package's lack of dependencies is not enough for all use cases.  `"repositories": [ "packagist": false ]` is an existing workaround.

Seldaek:
@ashnazg if all you're looking for is an autoloader you can simply use `composer dump-autoload` (`composer dump` works too).

PR #3951, opened 2015-04-19
Concurrent git clone and downloads with React

nicolas-grekas:
This adds an optional event loop argument and promises as return values to the GitDownloader.
The Installer takes advantage of it to trigger several `git clone` and manage them concurrently.
No change is made to the behavior of the code when no event loop is used, so that tests still pass.
This could be the first step towards parallel downloads in composer: the other downloaders would need to be updated following the same principle, the GitDownloader also needs to be finished (only public repos can be managed by the loop for now).
The console output also would need to be handled differently.
If someone else likes the approach, help is welcomed for porting the other downloaders to use and return promises! (a PR on this PR anyone?).
PHP 5.3 is not supported by React, but there is no blocker to change that.

### DIFF
--- a/src/Composer/Downloader/DownloadManager.php
+++ b/src/Composer/Downloader/DownloadManager.php
@@ -15,8 +15,6 @@ namespace Composer\Downloader;
 use Composer\Package\PackageInterface;
 use Composer\IO\IOInterface;
 use Composer\Util\Filesystem;
-use React\EventLoop\LoopInterface;
-use React\Promise\Deferred;
 use React\Promise\PromiseInterface;
 
 /**
@@ -169,7 +167,7 @@ class DownloadManager
      * @throws \InvalidArgumentException if package have no urls to download from
      * @throws \RuntimeException
      */
-    public function download(PackageInterface $package, $targetDir, $preferSource = null, LoopInterface $loop = null)
+    public function download(PackageInterface $package, $targetDir, $preferSource = null, $loop = null)
     {
         $preferSource = null !== $preferSource ? $preferSource : $this->preferSource;
         $sourceType   = $package->getSourceType();
@@ -193,51 +191,39 @@ class DownloadManager
 
         $this->filesystem->ensureDirectoryExists($targetDir);
 
-        $i = 0;
-        $deferred = new Deferred();
-        $retryLoop = function (\Exception $e = null) use ($sources, $package, $targetDir, $loop, &$i, $deferred, &$retryLoop) {
-            if (!isset($sources[$i])) {
-                $deferred->reject($e);
+        return $this->retryDownload(0, $sources, $package, $targetDir, $loop);
+    }
 
-                return;
-            }
-            $source = $sources[$i++];
-            try {
-                if (isset($e)) {
-                    $this->io->writeError(
-                        '    <warning>Failed to download '.
-                        $package->getPrettyName().
-                        ' from ' . $sources[$i - 1] . ': '.
-                        $e->getMessage().'</warning>'
-                    );
-                    $this->io->writeError('    <warning>Now trying to download from ' . $source . '</warning>');
-                }
-                $package->setInstallationSource($source);
-                $downloader = $this->getDownloaderForInstalledPackage($package);
-                $promise = $downloader ? $downloader->download($package, $targetDir, $loop) : null;
-                if ($promise instanceof PromiseInterface) {
-                    $promise->then(array($deferred, 'resolve'), $retryLoop);
-                } else {
-                    $deferred->resolve($promise);
-                }
-            } catch (\RuntimeException $e) {
-                $retryLoop($e);
-            }
-        };
-
-        $promise = $deferred->promise();
-        if (!$loop) {
-            $promise->done(function ($result) use (&$promise) {$promise = $result;});
-        }
-        try {
-            $retryLoop();
-            $retryLoop = null;
-        } catch (\Exception $e) {
-            $retryLoop = null;
+    private function retryDownload($i, $sources, $package, $targetDir, $loop, \Exception $e = null)
+    {
+        if (!isset($sources[$i])) {
             throw $e;
         }
+        $source = $sources[$i++];
+        if (isset($e)) {
+            $this->io->writeError(
+                '    <warning>Failed to download '.
+                $package->getPrettyName().
+                ' from ' . $sources[$i - 1] . ': '.
+                $e->getMessage().'</warning>'
+            );
+            $this->io->writeError('    <warning>Now trying to download from ' . $source . '</warning>');
+        }
+        $package->setInstallationSource($source);
+        $downloader = $this->getDownloaderForInstalledPackage($package);
 
-        return $promise;
+        if ($downloader) {
+            try {
+                $result = $downloader->download($package, $targetDir, $loop);
+                if ($result instanceof PromiseInterface) {
+                    return $result->then(null, function (\Exception $e) use ($i, $sources, $package, $targetDir, $loop) {
+                        return $this->retryDownload($i, $sources, $package, $targetDir, $loop, $e);
+                    });
+                }
+            } catch (\Exception $e) {
+                return $this->retryDownload($i, $sources, $package, $targetDir, $loop, $e);
+            }
+        }
     }
 
     /**
@@ -249,9 +235,8 @@ class DownloadManager
      *
      * @throws \InvalidArgumentException if initial package is not installed
      */
-    public function update(PackageInterface $initial, PackageInterface $target, $targetDir, LoopInterface $loop = null)
+    public function update(PackageInterface $initial, PackageInterface $target, $targetDir, $loop = null)
     {
-        $promise = null;
         $downloader = $this->getDownloaderForInstalledPackage($initial);
         if ($downloader) {
             $installationSource = $initial->getInstallationSource();
@@ -267,23 +252,15 @@ class DownloadManager
             // upgrading from a dist stable package to a dev package, force source reinstall
             if ($target->isDev() && 'dist' === $installationSource) {
                 $downloader->remove($initial, $targetDir);
-                $promise = $this->download($target, $targetDir, $loop);
-            } elseif ($initialType === $targetType) {
-                $target->setInstallationSource($installationSource);
-                $promise = $downloader->update($initial, $target, $targetDir, $loop);
-            } else {
-                $downloader->remove($initial, $targetDir);
-                $promise = $this->download($target, $targetDir, 'source' === $installationSource, $loop);
+                return $this->download($target, $targetDir, $loop);
             }
+            if ($initialType === $targetType) {
+                $target->setInstallationSource($installationSource);
+                return $downloader->update($initial, $target, $targetDir, $loop);
+            }
+            $downloader->remove($initial, $targetDir);
+            return $this->download($target, $targetDir, 'source' === $installationSource, $loop);
         }
-
-        if (!$promise) {
-            $deferred = new Deferred();
-            $deferred->resolve();
-            $promise = $deferred->promise();
-        }
-
-        return $promise;
     }
 
     /**
--- a/src/Composer/Downloader/DownloadManager.php
+++ b/src/Composer/Downloader/DownloadManager.php
@@ -15,6 +15,9 @@ namespace Composer\Downloader;
 use Composer\Package\PackageInterface;
 use Composer\IO\IOInterface;
 use Composer\Util\Filesystem;
+use React\EventLoop\LoopInterface;
+use React\Promise\Deferred;
+use React\Promise\PromiseInterface;
 
 /**
  * Downloaders manager.
@@ -166,7 +169,7 @@ class DownloadManager
      * @throws \InvalidArgumentException if package have no urls to download from
      * @throws \RuntimeException
      */
-    public function download(PackageInterface $package, $targetDir, $preferSource = null)
+    public function download(PackageInterface $package, $targetDir, $preferSource = null, LoopInterface $loop = null)
     {
         $preferSource = null !== $preferSource ? $preferSource : $this->preferSource;
         $sourceType   = $package->getSourceType();
@@ -190,30 +193,51 @@ class DownloadManager
 
         $this->filesystem->ensureDirectoryExists($targetDir);
 
-        foreach ($sources as $i => $source) {
-            if (isset($e)) {
-                $this->io->writeError('    <warning>Now trying to download from ' . $source . '</warning>');
-            }
-            $package->setInstallationSource($source);
-            try {
-                $downloader = $this->getDownloaderForInstalledPackage($package);
-                if ($downloader) {
-                    $downloader->download($package, $targetDir);
-                }
-                break;
-            } catch (\RuntimeException $e) {
-                if ($i === count($sources) - 1) {
-                    throw $e;
-                }
+        $i = 0;
+        $deferred = new Deferred();
+        $retryLoop = function (\Exception $e = null) use ($sources, $package, $targetDir, $loop, &$i, $deferred, &$retryLoop) {
+            if (!isset($sources[$i])) {
+                $deferred->reject($e);
 
-                $this->io->writeError(
-                    '    <warning>Failed to download '.
-                    $package->getPrettyName().
-                    ' from ' . $source . ': '.
-                    $e->getMessage().'</warning>'
-                );
+                return;
             }
+            $source = $sources[$i++];
+            try {
+                if (isset($e)) {
+                    $this->io->writeError(
+                        '    <warning>Failed to download '.
+                        $package->getPrettyName().
+                        ' from ' . $sources[$i - 1] . ': '.
+                        $e->getMessage().'</warning>'
+                    );
+                    $this->io->writeError('    <warning>Now trying to download from ' . $source . '</warning>');
+                }
+                $package->setInstallationSource($source);
+                $downloader = $this->getDownloaderForInstalledPackage($package);
+                $promise = $downloader ? $downloader->download($package, $targetDir, $loop) : null;
+                if ($promise instanceof PromiseInterface) {
+                    $promise->then(array($deferred, 'resolve'), $retryLoop);
+                } else {
+                    $deferred->resolve($promise);
+                }
+            } catch (\RuntimeException $e) {
+                $retryLoop($e);
+            }
+        };
+
+        $promise = $deferred->promise();
+        if (!$loop) {
+            $promise->done(function ($result) use (&$promise) {$promise = $result;});
         }
+        try {
+            $retryLoop();
+            $retryLoop = null;
+        } catch (\Exception $e) {
+            $retryLoop = null;
+            throw $e;
+        }
+
+        return $promise;
     }
 
     /**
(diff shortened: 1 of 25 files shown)


Soullivaneuh:
Big :+1: for this!

I will take a look for `FileDownloader` implement.

Soullivaneuh:
@nicolas-grekas Sorry for thoses questions but I'm quite new with reactphp.

Seems you implemented promise on `Git` class (`runCommand`). There is not such kind of thise clase for `FileDownloader`.

How would you implement it properly?

If you have some docs to advice me, I will be glad to read it! :+1:

src/Composer/Downloader/DownloadManager.php
```diff
@@ -166,7 +168,7 @@ public function getDownloaderForInstalledPackage(PackageInterface $package)
      * @throws \InvalidArgumentException if package have no urls to download from
      * @throws \RuntimeException
      */
-    public function download(PackageInterface $package, $targetDir, $preferSource = null)
+    public function download(PackageInterface $package, $targetDir, $preferSource = null, LoopInterface $loop = null)
```
staabm: why don't you pass in the loop at construction time instead of per method?

nicolas-grekas: because the download manager is used a a service, and this service sometimes can be called with a loop, and sometimes not

nicolas-grekas:
See README on react repositories (e.g. https://github.com/reactphp/promise/) and the code in this PR, I have nothing else...

nicolas-grekas:
Done for FileDownloader and ArchiveDownloader.
Next identified step: replace the main foreach loop in Installer by a callable hooked into the React loop to allow better concurrency management (limiting the number of parallel tasks)

barryvdh:
So is React being bumped down to php5.3 or Composer up to 5.4?

Soullivaneuh:
> Done for FileDownloader and ArchiveDownloader.

:+1: I'll learn a bit by reading this PR. ;-)

nicolas-grekas:
The InstallationManager now handles 8 concurrent install or update operations.
The result is significant when using `install --prefer-source`: from 6'30" to 1'50".
When using --prefer-dist, I go from 2'45" to 2'00, which is a much lower gain. This is due to the HTTP client (fopen) that blocks when sending the request. Unfortunately, I see no easy fix. The React http client would need a serious upgrade before replacing fopen.

stof:
@nicolas-grekas yeah, I think the React HTTP client would deserve being built on top of the async curl API instead of reimplementing HTTP from scratch on top of react-sockets

staabm:
For async HTTP Artax is a much better alternative 
https://github.com/amphp/artax

MarnuLombard:
@staabm : using Amp\Artax would move the min version requirement for Composer up to PHP 5.5 due to it using the Amp concurrency framework. 
https://github.com/amphp/amp

nicolas-grekas:
Artax looks better, so we could require Amphp instead of React. But Amphp lacks an async Process component. Any plan to have one @rdlowrey?
It looks like it would be much easier if we could at least require PHP 5.4 (which is the lowest req of both React and Amphp/Artax).
@Seldaek could we imagine a composer v1.1 or v2, requiring 5.4 or may be even 5.5?
I can imagine a process where self-update allows switching to v2 just by checking the local env first. Same for the installer.
The main point is using $this in closures. But that's also something we can workaround, in collaboration with @rdlowrey to make Artax 5.3 compatible, if he's OK considering that.
It's one way or an other (or none :) )
Finding a way to move composer forward looks the more sensible to me, but I may miss some context.

MattKetmo:
What about Guzzle 5, which also use react/promise?

staabm:
> It looks like it would be much easier if we could at least require PHP 5.4 (which is the lowest req of both React and Amphp/Artax).
> @Seldaek could we imagine a composer v1.1 or v2, requiring 5.4 or may be even 5.5?
> I can imagine a process where self-update allows switching to v2 just by checking the local env first. Same for the installer.

Alternatively we could also add this feature as only available when composer is running with php 5.4+ (without breaking it when used with an older php version)

bwoebi:
@nicolas-grekas I'm not sure, but what do you need a dedicated process library for?

It's pretty simple with Amp.

```PHP
$future = new Future;
$fds = [["pipe","r"],["pipe", "w"],["pipe", "w"]];
$proc = proc_open($cmd, $fds, $pipes, $cwd);
stream_set_blocking($pipes[1], false);
stream_set_blocking($pipes[2], false);
$timer = $reactor->once(function($reactor) use ($proc, $timeout, $future) {
    proc_terminate($proc);
    $future->fail(new \RuntimeException(sprintf('The process "%s" exceeded the timeout of %s seconds.', $command, $timeout)));
}, $timeout);
$reactor->onReadable($pipes[1], function($reactor, $watcher, $sock) use ($result, $callback, $timer, $future) {
    if ("" == $data = @fread($sock, 8192)) {
        $reactor->cancel($timer);
        $reactor->cancel($watcher);
        $reactor->immediately(function() use ($result, $future) {
            $future->resolve($result);
        });
    } else {
        $result->stdout .= $data;
        $callback("out", $data);
    }
});
$reactor->onReadable($pipes[2], function($reactor, $watcher, $sock) use ($result, $callback) {
    if ("" == $data = @fread($sock, 8192)) {
        $reactor->cancel($watcher);
    } else {
        $result->stderr .= $data;
        $callback("err", $data);
    }
});
return $future;
```

Then add a `proc_get_status()` for exit code at the end etc. … But don't return the _status_ in an execute() command like this. That's not async. That's blocking.

nicolas-grekas:
Thanks for the snippet @bwoebi!

stof:
@nicolas-grekas do it with the Process component rather than using ``proc_open`` directly though (``proc_open`` is very hard to make it portable, which is precisely why the Process component exists).

Using ``Process::start()`` instead of ``Process::run()`` allows to start a Process in a non-blocking way

bwoebi:
@nicolas-grekas And because it was so much fun… Quickly wrote up a nice class [with its own repository] ... https://github.com/amphp/process/tree/master

stof:
@bwoebi now make it work on windows :smile: I can tell you that your code won't work on Windows as is

@nicolas-grekas the cross-platform compatibility of react/child-process should be checked. Their code is based on an old version of the Symfony Process component, and lots of compatibility fixes have been done since then.

bwoebi:
@stof It's rather a myth that proc_open() has portability issues. There's only an issue with sigchild, but that's not something one can fix (because e.g. distros distribute PHP without configure command in phpinfo() etc.).

@stof Also, I was told by @rdlowrey it would work on Windows. Just check your react/child-process class there. It doesn't have _any_ extra Windows handling. Why should mine need some?

bwoebi:
@stof I need to correct myself. There's an issue with cwd on windows. I'll fix that. But I just looked at Symfony Process (WindowsPipes). It has some workarounds for bugs which were fixed a long time ago in php-src.

@stof Please tell me what won't work on Windows (except that what's explicitly documented in e.g. proc_get_status() that it won't work?)

stof:
@bwoebi it works on Windows for simple case easily. and then, start dealing with the different behavior of process streams in more complex cases and you will see why Symfony has a dedicated component

bwoebi:
@stof well… No. There aren't any issues. At least not telling from the comments in Symfony code.

https://bugs.php.net/bug.php?id=50524 — fixed.
https://bugs.php.net/bug.php?id=51800 — fixed.
https://bugs.php.net/bug.php?id=65650 — specific to file handles. Unrelated to my lib.

Please, find me something which is not yet fixed by today in 5.5+.

nicolas-grekas:
@staabm coding composer for 5.3 with code paths for 5.4+ is doable and could be the way to go. Doing this with code paths for 5.5+ could be harder: using yield is a parse error for 5.3. But it would be great.

@bwoebi thanks for the repos! You should register it on packagist, and be prepared to get contribs :) It would be great to have 5.4+ compat on this (see above)...

bwoebi:
@nicolas-grekas I would like to not support 5.4 because Windows. Some fixes are 5.5+ only. It'd create unnecessary complexity not really needed.

5.5+ IMO is a sane minimum requirement today...

nicolas-grekas:
OK, I'll try to make composer use the concurrency framework on 5.5+ only

staabm:
> Doing this with code paths for 5.5+ could be harder: using yield is a parse error for 5.3. But it would be great.

amphp provides some functional equivalents for `yield` etc. so your client code which does the phpversion checks can be written with php 5.4 compat; see e.g. https://github.com/amphp/amp/blob/master/lib/functions.php#L533

but as said above, doing this kind of optimized workflows on 5.5+ would be a good start and ease maintenance of the concurrency code.

stof:
@Seldaek could you give the stats about the PHP version usage when running composer ?

frederikbosch:
@stof He wrote an [article on this subject](http://seld.be/notes/my-view-of-php-version-adoption) a while ago.

bwoebi:
So, seems like PHP 5.3 _and_ 5.4 are slowly phasing out. So not really worth supporting 5.4 too, here. :-)

Seldaek:
I think if you make it work optionally only on 5.5 and above it could be merged, but otherwise we'll have to wait for a stable release so we can keep maintaining that for 5.3/5.4 for a while still, phasing out yes but hardly dead :/

bwoebi:
@Seldaek Well okay, I'll look into it (tomorrow or Saturday). Some version switches are needed, but then at least a big number of people will profit from it :-)

I just mean that it's not worth to do extra work and make everything supported with 5.4 [especially Windows] because it adds too much complexity for no real gain.

shouze:
:+1:

joshdifabio:
I've also forked Composer and worked on asyncing some of the functionality, although I worked on the Git and SVN drivers, not the downloaders.

To support this, I created [future process](https://github.com/joshdifabio/future-process) which provides mixed async/sync proc_open using futures, which makes it a bit easier to integrate async functionality into a synchronous application.

I'll be adding Symfony's sigchild and windows workarounds soon as well as support for output handler callables a la Symfony.

nicolas-grekas:
Tests are green. Could some of you confirm that it works for you? Only git + file download are supported. Other scm could be done later (I don't plan to do them since I don't use them).
The console output still needs some tweaks, and a few tests would be nice for the new "react" code paths. That's the only remaining things I'd like to do before merging. Or did I forgot something important?

src/Composer/Installer.php
```diff
@@ -519,90 +520,107 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
...
-                        'target' => $target,
-                        'reason' => $reason,
-                    );
-                }
-            }
+        $opsGroups = array(
+            $this->extractUninstalls($operations),
+            $this->extractPlugins($operations),
+            $operations,
+        );
```
Seldaek: IMO plugins should not be installed in parallel just in case they depend on one another. Uninstalls also execute pretty fast anyway so I would say having uninstalls and plugins first in a sequential manner then installs/updates in parallel would be the easiest.

Seldaek:
@nicolas-grekas what happened with artas/amp stuff btw? It sounded like it was better and then it got dropped, yet you still require php5.5 here to get parallelization. Just curious.

bwoebi:
@Seldaek I indeed have some branch (https://github.com/bwoebi/composer/commits/amp) (No doesn't work yet), but I abandoned after it just got too weird with 5.3/5.4. support. If we can use `yield` (5.5+ only), then I'm totally willing to continue my work. [Which will really mean a separate branch because T_YIELD will parse error on 5.3/5.4.]

But this code is just too weird with PHP 5.3 … A lot of callables chained each other through tons of when()'s … No, not really…

Seldaek:
Is it possible to use yield only in classes that would get loaded on 5.5 
and not below? Because then if we do it conditionally like in this PR it 
would still work.

The other question is which solution is likely to result in the least 
amount of new code? Because this PR is kind of a huge blob at the moment.

- react
- amp
- react or amp with no 5.3/5.4 support

If there is a drastic difference in amount of code I think we have to 
consider delaying this a bit and then branching off a new composer 
without 5.3/5.4 support. It will be a bit more pain to maintain for a 
while but then we end up with a cleaner solution in the end I hope.

I haven't had time to review in detail what part of the PR is due to 
conditionally doing it and what part is just regular react boilerplate. 
Curious to hear what you both think.

bwoebi:
It's half of classes which would need duplication in case because some Promise needs to be routed through...

When we use Amp with yield I mean. It'd only minimally affect code size (compared to now).

src/Composer/Installer.php
nicolas-grekas: Done

nicolas-grekas:
@Seldaek At some point, I tried replacing react's promises by amp's ones, but reverted because I'm not used to them. My feeling is that react's promises are more mature and should be kept. But that may be only because I don't know amp's ones well enough. The bonus of react's promises is that version ~1.0 is 5.3 compatible. That makes it possible to use here and some code paths with the async ones.

Then I focused on making the code 5.3 compatible, enabling the event loop only on 5.5 (no blocker to put 5.4 here). This is now done and the code is able to have both parallel and sequential behaviors depending on the PHP version.
The result is a very significant gain with --prefer-source with git, and a still significant one when downloading with --prefer-dist.

As said earlier, the bottleneck is now in --prefer-dist, with the fopen() blocking when sending the request (fetching the response is handled asynchronously).

My recommendation is to see this issue later, because changing the HTTP client is a much more involving process. At this time, we may have to consider changing for the amp event loop to use Artax, or React's HTTP client if it catches up fast enough. I see interest on both sides so both could happen.

For now, this PR looks huge, but it's mostly boilerplate related to react's promises, with a bit of event-loop related updates in the InstallationManager and in the RemoteFileSystem and Git utils.

zuhairnaqvi:
Any updates on this front? Composer is a great tool but the frustration of wasting considerable parts of my life every time i run composer update (in a project with private repos) is just too much!

frederikbosch:
@zuhairnaqvi Sorry dude, but I guess that is your own problem. In my experience, with a lot of private repos, is that composer handles pretty well. Especially if you compare with some other dependency/package managers.

zuhairnaqvi:
@frederikbosch No one's arguing that composer handles dependencies/versions well, it just takes minutes to even do a composer update --dry-run for projects using full stack frameworks. For instance, try a composer update --dry-run in https://github.com/sonata-project/sandbox (60+ packages) and tell me how long it takes just to resolve dependencies.

Compare this with npm (try installing something like https://github.com/punkave/apostrophe ) - it's blazingly fast.

If you add up all the wasted minutes across the whole php community using composer, you'll end up with several lifetimes.. 

Not sure if this is because Node is asynchronous while PHP (at least out of the box) is not and if there's some kind of blocking going on in dependency resolution?

PS I'm using the latest version of composer on PHP 5.5

frederikbosch:
@zuhairnaqvi Well, I guess this is not the place to start a discussion on this topic. So I leave it as is. If you want make composer faster: contribute. But leave those ```wasting considerable parts of my life``` comments at home. Nobody cares.

zuhairnaqvi:
@frederikbosch really nobody cares about developer UX? You're misrepresenting the composer community. People do care about performance of their libraries, as they should.

In any case, i'd be happy to contribute, in fact I just forked composer to see if it's possible to add some form of concurrency to dependency resolution.

stof:
@zuhairnaqvi we are all contributing to Composer on our free time only. Comments like ``wasting considerable parts of my life`` are not helpful.

anyway, any discussion about the dependency resolution should happen on dedicated issues, not here. This PR is about a totally different part of Composer: the download of the source code after the dependency resolution (which means it won't affect the time used by ``composer update --dry-run`` btw, as a dry-run run skips the package installation entirely)

frederikbosch:
@stof Thank you.

staabm:
This PR is ATM not managable because of its size. therefore we already discussed a few things which need to be refactored beforehand, so we can deliver a solid next version of composer.

After those refactors landed we will have a deeper look into the async topic again.

nicolas-grekas:
This PR is in a state where the async works, is disabled on lower supported PHP version.
The last thing I did was adding an env var to allow one to configure the concurrency level. When set to 1, this disables the async loop, the default is 8. A merge path could be changing the default to 1 so that the feature is disabled but widely testable.
The change is huge, but quite mechanical. I've been careful to change as least as possible (no refacto) and this is why tests still pass.
The one big missing step is a consul output manager that can catch concurrent console output and put them back in some semantic order for humans. Does anyone know if such a thing already exists btw?
If we default to no concurrency, we could merge this PR without the console output manager.
Happy to learn you're talking about a refacto to ease async (although it's not that here with the current code). Should I continue this PR or are you going to take over it?

staabm:
@nicolas-grekas I will do the required refactors after talking with Jordi what exactly is required.

This PR is nevertheless very helpfull because shows already which places need to be adjusted. Thanks for it.

trowski:
@staabm @nicolas-grekas You might be interested in the async framework I developed: [Icicle](https://github.com/icicleio/Icicle). It uses promises to build coroutines, so writing callbacks can be avoided. I'm nearing the first release of an [HTTP/1.1 server/client component](https://github.com/icicleio/Http) that could be used to concurrently download packages.

```php
use Icicle\Coroutine;
use Icicle\Http\Client\Client;
use Icicle\Loop;

$coroutine = Coroutine\create(function () {
    $client = new Client();

    $response = (yield $client->request('GET', 'https://www.google.com/teapot'));

    $stream = $response->getBody();
    while ($stream->isReadable()) {
        $data = (yield $stream->read());
        echo $data;
    }
    echo "\n";
});

Loop\run();
```

Obviously PHP 5.5 would be required, but if this is targeted at a future release, or something that can be selectively included, this could be a possibility.

Seldaek:
So we now have 4 concurrent implementations.. 

Guzzle v6 with its own event loop not sure if it has promises, ReactPHP's event loop/promise stuff, amphp's artax client and it has promises as well it seems, then Icicle also with an http client and promises.

What this tells me is:

- there is great competition going on, which is great.
- nobody talks to each other and people start from scratch instead, or perhaps there are crucial differences in implementation that mean we *need* new projects.
- half of those will die off or not be maintained enough in the future, so I would rather wait before merging any in composer.

/cc @mtdowling @rdlowrey @cboden

I don't want to start a huge debate about which is better in this thread, but it'd be interesting to hear from everyone how their implementation is unique compared to others, and if you by any chance realize that two or more can be combined and the other(s) can be killed off, that would be a win for everyone IMO.

mtdowling:
Guzzle doesn't really have an event loop. It does use promises. Guzzle is more of an abstraction over other HTTP transports (e.g., curl and PHP stream wrappers are just bindings into Guzzle), which makes it higher level than the other clients you've listed.

I think the most important thing to know for your purposes is that Guzzle is just an HTTP client. If you need sockets for something like the git protocol, then Guzzle isn't the answer.

One more option to consider might be conditionally checking if pthreads is installed, and if so, downloading through a thread pool.

nicolas-grekas:
@Seldaek this is something for @fig-standards :)

nicolas-grekas:
@Seldaek this is something for @php-fig :)

src/Composer/Downloader/FileDownloader.php
```diff
@@ -131,45 +150,71 @@ protected function doDownload(PackageInterface $package, $path, $url)
...
+                throw $e;
+            }
+            if ($this->io->isVerbose()) {
+                $this->io->writeError('    Download failed, retrying...');
+            }
+            usleep(500000);
+        }
+        try {
+            $result = $rfs->copy($hostname, $processedUrl, $fileName, $this->outputProgress, $package->getTransportOptions(), $loop);
+            if ($result instanceof PromiseInterface) {
```
staabm: @Seldaek I would propose refactor to use a new `Composer\Promise` for the "regular path" as a first step, so we dont need to do this `PromiseInterface` checks at each callsite.

beind this `Composer\Promise` we could later on hide whatever lib we gonna use.

wdyt?

//cc @nicolas-grekas

frederikbosch: @staabm In my opinion it does not really matter. If you create a ```Composer\Promise```, then you probably are going to inject an adapter into its constructor, e.g. Icicle or ReactPHP. And because you do not want lean on a concretion but on an interface something like ```PromiseAdapterInterface``` will be created. Hence, the result is quite the same. If there is going to be an interface, my vote would go to the current ```PromiseInterface```.

nicolas-grekas: This check for a promise allowed me to move to promises by parts:
- for switching the async/sync code paths,
- for handling downloader/installers/etc. that where not fully moved to promises (which is not required for purely local code paths)

frederikbosch: @nicolas-grekas But there is more to it: you hide the concretion. @Seldaek If Composer goes forward on that part, then the decision which library to use is not as important as it may seem.

trowski: I'm working toward a solution to this, providing a common interface to all promises: [Thenable](https://github.com/thenable/Thenable).

nevvermind: @trowski - why do you need a project for PHP FIG? Wouldn't it be better to just throw the idea in the FIG mailing list?

trowski: @nevvermind PHP-FIG is't going to standardize something unless there's a need. It would be nice to show that such a need exists. Same reason projects such as this exist: https://github.com/container-interop/container-interop

For now though I removed the FIG reference as I think this is a far off goal, and don't want to mis-represent it as being supported by the PHP-FIG.

nicolas-grekas:
Please note that I forgot to add concurrency do the downloading of packagist's packages.json files

Baachi:
Any news here?

digitalkaoz:
@nicolas-grekas i use https://packagist.org/packages/hirak/prestissimo for parallel installs, and it speeds up enough...

o5:
@digitalkaoz do you use latest composer with `hirak/prestissimo`? I think it does not work now, [discussion](https://github.com/hirak/prestissimo/issues/137#issuecomment-291805514)

LKDevelopment:
@Seldaek Any Progress on this? It is a shame, that speedups actually must be done on a "per-project" basis like Symfony did it (https://github.com/symfony/flex/pull/261).

akalongman:
Any update?

nicolas-grekas:
Time to close this one :)

PR #3953, opened 2015-04-20
Allow numeric author name

### LINKED ISSUE #3948: Composer init Author name not valid.
Hi,

Whenever I run composer init and want to accept the default suggestion for Author, composer says it's an invalid author string. I know this is because my Github username has numbers in it, but I always manually modify the composer.json to add those anyway. Could the validation be modified so that it also accepts numbers in the Author string?

alcohol:
Could also change it to just accept whatever. Do we really want to enforce strict rules regarding valid author names?

Closes #3948

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -40,7 +40,7 @@ class InitCommand extends Command
 
     public function parseAuthorString($author)
     {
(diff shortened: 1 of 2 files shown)


TomasVotruba:
:+1:

stof:
:+1:

Seldaek:
I don't think there should be too much garbage in there, and if it's the first problem reported until now I would say reality agrees :)

Seldaek:
But thanks for the fix, numbers seem like a legit use case.

PR #3960, opened 2015-04-22
Add argument to update lock without installing

andrewnicols:
The documentation for update --lock states:
`Only updates the lock file hash to suppress warning about the lock file being out of date.`

From this I understand that we are performing a `composer update`, which would normally update the dependency tree, update the lock file, and install the dependencies into the vendor directory; but without actually installing the dependencies.

However, if I update my composer.json from one fixed version to another, then call `composer update --lock`, I get a confusing message which suggests that the lock file is being updated based on the data in the composer.json *and* the data currently in composer.lock.

### Steps to reproduce:

    # Clone my sample repo and install the dependencies:
    git clone https://github.com/andrewnicols/composer-mtc-issue.git
    git checkout master
    composer install

    # Bump the version of the dependency, to 1.28.0, or checkout the branch I created for you:
    git checkout versionUpdated

    # Update the lock file:
    composer update --lock

### Expected Result
The lock file is updated without installing the dependencies into the vendor directory.

### Actual Result
Some confusing mismash of composer.lock + composer.json:

      Problem 1
        - The requested package moodlehq/behat-extension == 1.28.6.0 could not be found.
      Problem 2
        - Installation request for moodlehq/behat-extension 1.28.8 -> satisfiable by moodlehq/behat-extension[v1.28.8].
        - Conclusion: remove guzzle/common v3.9.2
        - moodlehq/behat-extension v1.28.8 requires guzzlehttp/guzzle ~3.1 -> satisfiable by guzzlehttp/guzzle[v3.1.0, v3.1.1, v3.1.2, v3.2.0, v3.3.0, v3.3.1, v3.4.0, v3.4.1, v3.4.2, v3.4.3, v3.5.0, v3.6.0, v3.7.0, v3.7.1, v3.7.2, v3.7.3, v3.7.4, v3.8.0, v3.8.1].
        - don't install guzzlehttp/guzzle v3.1.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.1.1|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.1.2|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.2.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.3.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.3.1|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.4.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.4.1|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.4.2|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.4.3|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.5.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.6.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.7.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.7.1|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.7.2|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.7.3|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.7.4|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.8.0|don't install guzzle/common v3.9.2
        - don't install guzzlehttp/guzzle v3.8.1|don't install guzzle/common v3.9.2
        - Installation request for guzzle/common == 3.9.2.0 -> satisfiable by guzzle/common[v3.9.2].

    Potential causes:
    - A typo in the package name
      see <https://groups.google.com/d/topic/composer-dev/_g3ASeIFlrc/discussion> for more details.

    Read <http://getcomposer.org/doc/articles/troubleshooting.md> for further common problems.

### Usecase
We wish to add a CI server task which runs composer update --lock to ensure that the lock file is up-to-date with the composer.json and will warn accordingly if someone attempts to push a change to one without a change to the other.
We do not necessarily need to install the dependencies for this task.

### Summary
This could simply be an issue in the documentation - e.g. that --lock does not actually do as we expect, or it could be doing completely the wrong thing.

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -39,6 +39,7 @@ class UpdateCommand extends Command
                 new InputOption('dev', null, InputOption::VALUE_NONE, 'Enables installation of require-dev packages (enabled by default, only present for BC).'),
                 new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables installation of require-dev packages.'),
                 new InputOption('lock', null, InputOption::VALUE_NONE, 'Only updates the lock file hash to suppress warning about the lock file being out of date.'),
+                new InputOption('no-install', null, InputOption::VALUE_NONE, 'Update the lock file, but do not install packages.'),
                 new InputOption('no-plugins', null, InputOption::VALUE_NONE, 'Disables all plugins.'),
                 new InputOption('no-custom-installers', null, InputOption::VALUE_NONE, 'DEPRECATED: Use no-plugins instead.'),
                 new InputOption('no-autoloader', null, InputOption::VALUE_NONE, 'Skips autoloader generation'),
@@ -126,6 +127,7 @@ EOT
             ->setRunScripts(!$input->getOption('no-scripts'))
             ->setOptimizeAutoloader($optimize)
             ->setUpdate(true)
+            ->setInstall(!$input->getOption('no-install'))
             ->setUpdateWhitelist($input->getOption('lock') ? array('lock') : $input->getArgument('packages'))
             ->setWhitelistDependencies($input->getOption('with-dependencies'))
             ->setIgnorePlatformRequirements($input->getOption('ignore-platform-reqs'))
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -110,6 +110,8 @@ class Installer
     protected $ignorePlatformReqs = false;
     protected $preferStable = false;
     protected $preferLowest = false;
+    protected $install = true;
+
     /**
      * Array of package names/globs flagged for update
      *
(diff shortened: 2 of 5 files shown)


andrewnicols:
I had more of a read of things, and it appears that the --lock option is only intended to update the lock file's hash. As a result, it *will* die horribly if versions between the lock and the json differ.

I've implemented a --lock-only flag which essentially runs the same operations as a dry-run but does update the lockfile. It is possible to run this *without* a vendor directory.

stof:
``--no-install`` or ``--no-download`` would be a better name than ``--lock-only`` IMO.

And this feature must be tested to ensure it works properly. Otherwise it cannot be accepted.

Regarding the documentation of ``--lock``, please open a separate issue asking to improve the doc (or even better a PR doing the improvement)

andrewnicols:
Thanks @stof,

I've renamed this to --no-install, and have written a pair of unit test fixtures to cover the changes.

Seldaek:
This is fixed by #4140 IMO as the lock file warning should not happen without good reason anymore, and ignoring it with a update --no-install wouldn't really be wise.

PR #3963, opened 2015-04-23
Fix test classes with invalid PHP.

deviantintegral:
This was causing our PHP linter to fail on builds. Even though these files aren't actually loaded by Composer as PHP code, to me it makes sense to at least always be shipping valid PHP in .php files.


deviantintegral:
Also, the small script I used to regenerate these files:

```php
<?php

$file = file_get_contents('php://stdin');

$count = 1;

foreach (explode("\n", $file) as $line) {
  $name = 'a' . $count;
  echo str_replace('1', $name, $line, $replaced) . "\n";
  if ($replaced) {
    $count++;
  }
}
```

xabbuh:
Technically, the last semicolon in a PHP file can be omitted. So maybe you may want to take this into account in your linter.

Seldaek:
Yeah this doesn't really make sense to me. These files are valid PHP according to `php -l` and they are not part of your project at all so I don't see why your linter would even check them nor why it'd fail.

Seldaek:
Ah well, obviously I checked the last one which was the only one that actually lints correctly. I guess I'll merge then as indeed the other ones don't lint correctly. Sorry about that.

That being said I still think your build shouldn't necessarily lint into dependencies (at least test dirs are kind of a waste of cpu cycles).

deviantintegral:
Thanks @Seldaek !

PR #3967, opened 2015-04-27
Fixed svn update command

xi-ao:
=> Fails before attempting to use authentication provided in auth.json because the svn command outputs to STDOUT *and* STDERR, which breaks the condition that checks for an authentication issue before throwing a RuntimeException.

### DIFF
--- a/src/Composer/Util/Svn.php
+++ b/src/Composer/Util/Svn.php
@@ -121,7 +121,7 @@ class Svn
         }
 
         $errorOutput = $this->process->getErrorOutput();
-        $fullOutput = join("\n", array($output, $errorOutput));
+        $fullOutput = implode("\n", array($output, $errorOutput));
 
         // the error is not auth-related
         if (false === stripos($fullOutput, 'Could not authenticate to server:')
--- a/src/Composer/Util/Svn.php
+++ b/src/Composer/Util/Svn.php
@@ -120,16 +120,18 @@ class Svn
             return $output;
         }
 
+        $errorOutput = $this->process->getErrorOutput();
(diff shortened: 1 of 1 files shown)


src/Composer/Util/Svn.php
```diff
@@ -120,16 +120,18 @@ public function execute($command, $url, $cwd = null, $path = null, $verbose = fa
             return $output;
         }
 
+        $errorOutput = $this->process->getErrorOutput();
         if (empty($output)) {
-            $output = $this->process->getErrorOutput();
+            $output = $errorOutput;
         }
+        $fullOutput = "$output\n$errorOutput";
```
stof: in case ``$output`` was empty, this now contains a duplicate of the error output because of the previous condition

xi-ao: You're absolutely right, and I came to the same observation after commiting this fix. I'll push a better one very soon.

src/Composer/Util/Svn.php
```diff
@@ -120,16 +120,15 @@ public function execute($command, $url, $cwd = null, $path = null, $verbose = fa
             return $output;
         }
 
-        if (empty($output)) {
-            $output = $this->process->getErrorOutput();
-        }
+        $errorOutput = $this->process->getErrorOutput();
+        $fullOutput = join("\n", array($output, $errorOutput));
```
stof: please use ``implode`` rather than the ``join`` alias

xi-ao:
Is the fix okay now? Let me know if I missed something.

Seldaek:
Thanks

PR #3970, opened 2015-04-28
[Docs] [Typo] Correct a misused article in the docs

fabschurt:
I'm not a native English speaker, but I think that we must use _a_ (and not _an_) in front of a _u_ when it is pronounced _you_ (like in _user_).

Source: https://www.englishclub.com/pronunciation/a-an.htm


TomasVotruba:
:+1:

PR #3977, opened 2015-04-29
added the actual phpversion into the error message when phpversion doesn...

staabm:
... satisfy the requirement

Sometimes it takes a few moments until you check that composer does not run on the composer binary you initally guessed.


kelunik:
:+1:

rdlowrey:
:+1:

Seldaek:
Thanks

PR #3978, opened 2015-04-29
fixes #3976: drop ext/ctype dependency

### LINKED ISSUE #3976: drop ext/ctype dependency
requiring less dependencies is a good thing for a fundamental tool like composer is.

what do you think about rewriting stuff to "simple" sanity checks instead of relying on ext/ctype?
could provide a PR if you agree with it.

marc-mabe:
#3976

### DIFF
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -147,7 +147,7 @@ class ArrayLoader implements LoaderInterface
         }
 
         if (!empty($config['time'])) {
-            $time = ctype_digit($config['time']) ? '@'.$config['time'] : $config['time'];
+            $time = preg_match('/^\d+$/D', $config['time']) ? '@'.$config['time'] : $config['time'];
 
             try {
                 $date = new \DateTime($time, new \DateTimeZone('UTC'));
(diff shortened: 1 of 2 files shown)


src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -147,7 +147,7 @@ public function load(array $config, $class = 'Composer\Package\CompletePackage')
         }
 
         if (!empty($config['time'])) {
-            $time = ctype_digit($config['time']) ? '@'.$config['time'] : $config['time'];
+            $time = preg_match('/^[0-9]$/m', $config['time']) ? '@'.$config['time'] : $config['time'];
```
Seldaek: Can you drop the `m` modifier here? I don't see how it makes sense

Seldaek:
I guess I am fine with it given it's one line, but can you also PR the installer file in https://github.com/composer/getcomposer.org then? It also needs to be updated.

src/Composer/Package/Loader/ArrayLoader.php
marc-mabe: You are right, the `m` modifier is wrong. What I mean was the `D` modifier.
-> see http://3v4l.org/chkLF

marc-mabe:
@Seldaek done composer/getcomposer.org#80

Seldaek:
Thanks!

PR #3982, opened 2015-04-30
Prioritize repository over package versions in the default policy

### LINKED ISSUE #3509: Replacing package from private repository with package from packagist
I think that it's common way that users use private and public repositories (packagist.org) at the same time. But villain could discover vendor and package name of your private package(I do not think that this information has to be private) and create new evil one on packagist with the same vendor and package name, but with higher version.  It will substitute private package without any notice (changing of version only) . I think that is security issue.

My proposals(easy):
Add note to the documentation: using private and public repositories at the same time could be dangerous.
Add source of installed to an output of composer install or update.
Attract attention when source of package changed.

(hard):
Change logic of composer. Add possibility to choose repository of package.

Seldaek:
In practice it is quite rare to have a package in multiple repos unless you intend to install your own fork of it, so I don't think this breaks any real use case, and it would improve security by preventing hijacks from packagist using a higher version of a private package.

@naderman I'd be happy to get your feedback in case you think of something I missed, otherwise I'll go ahead and merge.

Fixes #3509

### DIFF
--- a/doc/dev/DefaultPolicy.md
+++ b/doc/dev/DefaultPolicy.md
@@ -12,6 +12,13 @@ resulting order in which the solver will try to install them.
 
 The rules are to be applied in the order of these descriptions.
 
+### Repository priorities
+
+Packages Repo1.Av1, Repo2.Av1
+
+* priority(Repo1) >= priority(Repo2) => (Repo1.Av1, Repo2.Av1)
+* priority(Repo1) <  priority(Repo2) => (Repo2.Av1, Repo1.Av1)
+
 ### Package versions
 
 Packages: Av1, Av2, Av3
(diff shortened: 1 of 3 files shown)


enumag:
So the priority of repositories is determined by the order on "respositories" section and packagist behaves as if it was the last?

Seldaek:
Yes packagist if not specified explicitly is added last so lowest priority, and the others you define yourself work in top-to-bottom first wins priority

naderman:
Yeah I think this makes sense, use case is somewhat different from what I expected when I originally build that.

Seldaek:
Ok let's run with it then and see but it sounds safe enough to me.

PR #3983, opened 2015-04-30
Upgrade DialogHelper to QuestionHelper

### LINKED ISSUE #3760: Replace deprecated TableHelper and DialogHelper before updating Composer dependencies

alcohol:
Closes #3760


src/Composer/Command/InitCommand.php
```diff
@@ -255,44 +244,64 @@ function ($value) use ($self, $minimumStability) {
...
             $devRequirements = $this->determineRequirements($input, $output, $input->getOption('require-dev'));
         }
         $input->setOption('require-dev', $devRequirements);
     }
 
+    /**
+     * @param string $author
+     * @return array
+     */
+    public function parseAuthorString($author)
```
stof: public methods should stay before protected ones (and avoiding to move the code around like this also makes the review easier)

Seldaek: I agree on the review, but it's kinda private anyway, it's only public because it has to be used by a closure, so in that case maybe tagging it with `@private` and moving it down makes more sense.

stof:
This change currently requires bumping the Console requirement to 2.5+ (because it supports only the new API), and this is not possible yet because Packagist still runs on Symfony 2.3 and it needs Composer.
This means that the compatibility with Symfony Console 2.3 must be kept.

src/Composer/Command/InitCommand.php
alcohol: Oh, I'm sorry. I'm just so used to putting helper methods after the config and execute methods. I'll revert that change.

Seldaek:
We already require 2.5, and I can live with that for now..

alcohol:
I was about to say.. https://github.com/composer/composer/blob/master/composer.json#L28

Seldaek:
Thanks!

PR #3994, opened 2015-05-04
Load only explicitly named package data

naderman:
This will significantly reduce the amount of loaded metadata for the solver. This should improve network time as well as solver complexity.

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -132,76 +132,6 @@ class ComposerRepository extends ArrayRepository
         return $loadedPackages;
     }
 
-    protected function loadName($name, $acceptableCallback, $exactMatch = true)
-    {
-        // skip platform packages
-        if (preg_match(PlatformRepository::PLATFORM_PACKAGE_REGEX, $name) || '__root__' === $name) {
-            return array();
-        }
-
-        if (null === $this->providerListing) {
-            $this->loadProviderListings($this->loadRootServerFile());
-        }
-
-        if ($this->lazyProvidersUrl && !isset($this->providerListing[$name])) {
-            $hash = null;
-            $url = str_replace('%package%', $name, $this->lazyProvidersUrl);
-            $cacheKey = false;
-        } elseif ($this->providersUrl) {
-            // package does not exist in this repo
-            if (!isset($this->providerListing[$name])) {
-                return array();
-            }
-
-            $hash = $this->providerListing[$name]['sha256'];
-            $url = str_replace(array('%package%', '%hash%'), array($name, $hash), $this->providersUrl);
-            $cacheKey = 'provider-'.strtr($name, '/', '$').'.json';
-        } else {
-            // BC handling for old providers-includes
-            $url = 'p/'.$name.'.json';
-
-            // package does not exist in this repo
-            if (!isset($this->providerListing[$url])) {
-                return array();
-            }
-            $hash = $this->providerListing[$url]['sha256'];
-            $cacheKey = null;
-        }
-
-        if ($cacheKey && $this->cache->sha256($cacheKey) === $hash) {
-            $packages = json_decode($this->cache->read($cacheKey), true);
-        } else {
-            $packages = $this->fetchFile($url, $cacheKey, $hash);
-        }
-
-        $loadedPackages = array();
-        foreach ($packages['packages'] as $versions) {
-            foreach ($versions as $version) {
-                if ($exactMatch && $version['name'] !== $name) {
-                    continue;
-                }
-
-                if ($acceptableCallback && !call_user_func(
-                    $acceptableCallback, strtolower($version['name']), VersionParser::parseStability($version['version'])
-                )) {
-                    continue;
-                }
-
-                // load acceptable packages in the providers
-                $package = $this->createPackage($version, 'Composer\Package\Package');
-                $this->addPackage($package);
-
-                $loadedPackages[] = $package;
-
-                if ($package instanceof AliasPackage) {
-                    $loadedPackages[] = $package->getAliasOf();
-                }
-            }
-        }
-
-        return $loadedPackages;
-    }
-
     /**
      * {@inheritDoc}
      */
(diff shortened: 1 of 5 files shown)


src/Composer/Repository/ComposerRepository.php
```diff
@@ -96,6 +96,151 @@ public function setRootAliases(array $rootAliases)
         $this->rootAliases = $rootAliases;
     }
 
+    public function ensureLoaded($pool, array $constrainedNames)
```
stof: ``$pool`` should be typehinted

src/Composer/Repository/ComposerRepository.php
```diff
@@ -96,6 +96,151 @@ public function setRootAliases(array $rootAliases)
...
+                    $workQueue->enqueue(array(
+                        'name' => $link->getTarget(),
+                        'constraint' => $link->getConstraint()
+                    ));
+                }
+            }
+        }
+    }
+
+    protected function loadName($pool, $name)
```
stof: same here

src/Composer/Repository/ComposerRepository.php
naderman: I don't mean to leave any of this code as-is, this was a POC to see if this is doable, I'll rewrite all of this anyway.

src/Composer/Repository/ComposerRepository.php
naderman: staaahp.

src/Composer/Repository/ComposerRepository.php
```diff
@@ -96,6 +96,151 @@ public function setRootAliases(array $rootAliases)
...
+            if (isset($this->loadedMap[$packageSpec['name']])) {
+                continue;
+            }
+
+            $this->loadedMap[$packageSpec['name']] = true;
+
+            $packages = $this->loadName($pool, $packageSpec['name']);
+
+            foreach ($packages as $package) {
+                foreach ($package->getRequires() as $link) {
```
stof: shouldn't this be handled in the Pool to ensure that dependencies can be provided by other repositories ?

naderman: please read my reply :(

src/Composer/Repository/ComposerRepository.php
stof: and the method should be private

tests/Composer/Test/Repository/ComposerRepositoryTest.php
```diff
@@ -97,7 +97,7 @@ public function loadDataProvider()
     }
 
     public function testWhatProvides()
-    {
+    {/*
         $repo = $this->getMockBuilder('Composer\Repository\ComposerRepository')
```
stof: this code should be moved to test ``ensureLoaded``

src/Composer/Repository/ComposerRepository.php
stof: and what about dev requirements for the root package ?

stof: @naderman your replies don't appear in the diff of the PR until I reload the diff :smile:

stof:
Something looks broken. Tests are not passing

naderman:
Yes because the current list of loaded packages doesn't actually contain everything that is needed.

naderman:
Again this is really just a FYI PR that I was discussing with jordi which I didn't want to just live on my disk, no need to discuss this yet.

naderman:
current composer:

```
naderman@Montsoreau:~/projects/composer/test/ezpublish-community (master) $ ./composer.phar update -v --profile --dry-run --ignore-platform-reqs --no-dev                                                                                                                      
[8.6MB/0.11s] Loading composer repositories with package information
[8.9MB/0.37s] Updating dependencies
[312.7MB/29.30s] Memory usage: 312.66MB (peak: 899.13MB), time: 29.3s
```

with this patch which still loads full provider files, and only reduces the data after it is downloaded:

```
naderman@Montsoreau:~/projects/composer/test/ezpublish-community (master) $ ../../composer/bin/composer update -v --profile --dry-run --ignore-platform-reqs --no-dev
[8.9MB/0.11s] Loading composer repositories with package information
[9.2MB/0.35s] Updating dependencies
[259.6MB/22.85s] Memory usage: 259.6MB (peak: 759.64MB), time: 22.85s
```

So pretty confident this can make a significant performance impact if implemented properly.

naderman:
Alright I think this is ready to be reviewed as-is @Seldaek  @stof - as a next step we could look into changing the json data that is downloaded off of packagist, this is unchanged in the PR for now.

src/Composer/Repository/ComposerRepository.php
```diff
@@ -96,6 +95,43 @@ public function setRootAliases(array $rootAliases)
         $this->rootAliases = $rootAliases;
     }
 
+    public function getRootAliases()
+    {
+        return $this->rootAliases;
+    }
+
+    public function loadRecursively(array $packageNames, $acceptableCallback)
```
stof: please add some phpdoc to document that the second argument is a callable, as we cannot typehint it

src/Composer/Repository/ComposerRepository.php
```diff
@@ -222,40 +258,17 @@ public function getProviderNames()
...
-            $package->setId(-1);
-        }
-    }
-
-    public function whatProvides(Pool $pool, $name)
+    /**
+     * Loads package data for a given package name or provider name
+     *
+     * @param string $name
+     * @param callable $acceptableCallback A callback to check if a package should be loaded
```
stof: ``callable|null`` actually

src/Composer/Repository/ComposerRepository.php
```diff
@@ -296,69 +309,51 @@ public function whatProvides(Pool $pool, $name)
...
-                        // check for root aliases
-                        if (isset($this->providersByUid[$version['uid'].'-root'])) {
-                            $this->providers[$name][$version['uid'].'-root'] = $this->providersByUid[$version['uid'].'-root'];
-                        }
-                    }
-                } else {
-                    if (!$pool->isPackageAcceptable(strtolower($version['name']), VersionParser::parseStability($version['version']))) {
-                        continue;
-                    }
+                if ($exactMatch && $version['name'] !== $name) {
```
stof: couldn't this be checked for packages rather than for versions ? IIRC, ``$packages['packages']`` is indexed by package name in the composer repository format (we could even avoid looping over all packages in this case)

src/Composer/Repository/ComposerRepository.php
```diff
@@ -296,69 +309,51 @@ public function whatProvides(Pool $pool, $name)
...
-                        $this->providersByUid[$version['uid'].'-root'] = $alias;
-                    }
-                }
+                return;
             }
         }
+    }
 
-        return $this->providers[$name];
+    public function hasProviders()
```
stof: I suggest keeping these methods before ``loadName`` to keep the diff more readable

naderman:
Fixed everything @stof brought up - want to merge this then @Seldaek ?

Seldaek:
Alright looking good, it reduces memory footprint a little and runtime as well since we process less files so that means less network time too which is nice.

A good side effect though is that it moves all the loading ahead of the solver, so there won't be confusion anymore as to what it's doing between downloading two files, now they should download quick and then solving starts.

splenetic:
@Seldaek @naderman After this was merged I've started to get errors when using private (Satis) and public packages together.
Here's a slightly obfuscated error message:
```
Problem 1
  - Installation request for private/package1 dev-master -> satisfiable by private/package1[dev-master].
  - private/package1 dev-master requires anchu/ftp 0.0.1 -> no matching package found.
Problem 2
   - Installation request for private/package2 dev-master -> satisfiable by private/package2[dev-master].
   - private/package2 dev-master requires squizlabs/php_codesniffer 1.* -> no matching package found.
 Problem 3
   - Installation request for private/package3 1.*@dev -> satisfiable by private/package3[1.0.x-dev].
   - private/package3 1.0.x-dev requires aws/aws-sdk-php 2.6.* -> no matching package found.
```
Reverting composer to a version right before this change (b025d0982b78ad76d381d16f8303d6b0769fc5ea) fixes this.

stof:
Do you have capital letters in the name of packages in your Satis repo ?

splenetic:
Nope, no capital letters in package names. There are capital letters in some version names though (branch names) - not sure if that matters.

stof:
nope, it should not

naderman:
Well the issue seems to be that for some reason some packages are not being loaded. Since this is multiple repositories I figure there is some issue in that section of the code, that tries to load names it found as deps in one repo in another.

naderman:
Would be great if we had a test for this ^_^

naderman:
Going to revert this, while I debug, as this may take a while.

PR #4003, opened 2015-05-05
PHP version checks tweaking

SpacePossum:
`PHP_VERSION_ID` is available since PHP 5.2.7 and already used in Composer. This will speeds up PHP version checking. With `Composer requires PHP 5.3.2+ to run.` this should be safe to do.


Seldaek:
Thakns

SpacePossum:
Thank you :)

PR #4016, opened 2015-05-08
for multi lang

PheRum:


Seldaek:
As I have said a few times in other translation related PRs, I don't think this is the best approach. Forking the repository would allow you to merge in upstream changes while seeing what changed compared to your translations. If we duplicate all the files we lose this capability.

PR #4018, opened 2015-05-10
Fix link to Composer\Installer\InstallerInterface.

olavmrk:
The link was pointing to Composer\Plugin\PluginInterface since
commit 98e5eab. Fix the link by updating the reference from
`[3]` to `[4]`.


Seldaek:
Thanks

Issue #1358, opened 2012-11-24
Load "extra" from config.json and merge with composer.json


naderman:
Since we allow the "extra" section of composer.json to contain pretty much any data for customization, it'd be useful to allow setting it globally as well. Should work in the same way "config" is currently handled I think.

gggeek:
+1 for env vars being made available to script handlers.

My current gripe is that I have some script handlers which are not in my control (3rd party), and those only accept values set in the 'extra' section or composer.json, or ask the user for input.
It goes without saying that this makes it hard for me to automate calls to 'composer install' for those apps.

While I agree that I could ask to the maintainers of those app to change their script handler classes to look for env vars as well, I think it would be better if Composer itself made the env vars available via an API, so that developers (of script handler scripts) would be encouraged to use them.

Thanks
Gaetano

gggeek:
ps: sorry for posting OT on closed issues. Shall I open a new issue ?

Seldaek:
@gggeek pleas open a new issue with more details because I  don't quite get what you mean. As for this issue I think it's old enough that we can close it at this point. Doesn't seem all too important.

PR #4035, opened 2015-05-15
Mark "extra" as root only in docs

nevvermind:
I was so glad that "extra" wasn't set as root-only so that I could have other packages merge stuff in the root-package. 
Then I read this guy: https://tobymackenzie.wordpress.com/2014/07/11/composer-scripts-and-extra-from-non-root-package/

Dammit!

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -832,7 +832,7 @@ through the use of scripts.
 
 See [Scripts](articles/scripts.md) for events details and examples.
 
-### extra
+### extra <span>(root-only)</span>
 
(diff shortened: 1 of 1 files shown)


stof:
Well, this is not true.

Extras are loaded from other packages too. and installers could read extras of a package when installing it for instance.
The issue described in the blog post is that it was expecting that a script reading th extras of the root package would magically see the extras of other packages too merged in it. This does not happen. Extras are per package.

nevvermind:
I see. Then... never mind.
I'm guessing there's no similar way of doing things like https://github.com/composer/composer/issues/1358?

PR #4038, opened 2015-05-15
Remember that a class does exist

loicfevrier:
If a class is already loaded and we call
<pre>
$composer_loader->findFile($class)
</pre>
the composer will redo the full search for the class.

We can now remember that the class exist which improves performances.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -337,6 +337,7 @@ class ClassLoader
             // Remember that this class does not exist.
             return $this->classMap[$class] = false;
         }
+        
         return $this->classMap[$class] = $file;
     }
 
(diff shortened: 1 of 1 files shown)


GromNaN:
But the same class cannot be loaded twice.

loicfevrier:
I actually need to call 
<pre>
$composer_loader->findFile($class)
</pre>
after the class has been loaded.

I have a 
<pre>
$composer_loader->loadClass($new_class)
</pre>
and a few line after I need to inspect the path by calling
<pre>
$path = $composer_loader->findFile($class)
</pre>
(if the class is found in a specific directory, some post-process is needed)

alcohol:
Can you undo the removal of the newline?

alcohol:
Other than that, while there is no immediate benefit in general, it does benefit scenarios as described by @loicfevrier that use the internals of the autoloader.

If you could revert the newline change, I'd say :+1: .

loicfevrier:
It's done.

Seldaek:
I feel like this is adding overhead (memory and the actual storing operation, not much but still) for everyone while it only benefits your edge case. I am not sure why you need to do it, but you can do it easily from the outside by wrapping the composer ClassLoader, and either calling findFile instead of loadClass directly, or checking the return value of loadClass to see if it found it or not.

PR #4049, opened 2015-05-22
Mention that CRLFs can prevent finding "RUN" in test files.

forevermatt:


forevermatt:
This came up because I'm developing on a Windows machine, and the default Git setting for core.autocrlf (which is "true") converted the line endings from LF to CRLF... which actually ended up breaking around 25 of the tests.  Maybe this updated error message will help others catch this problem if they experience it as well.

alcohol:
How about instead you fix https://github.com/forevermatt/composer/blob/clarify-crlf-error/tests/Composer/Test/AllFunctionalTest.php#L139 so that it is more portable? :-)

forevermatt:
Since so many tests were broken by Git messing up the line endings, and since (in my semi-limited experience) it seems a better practice to tell Git not to mess with line endings (by setting core.autocrlf to false), I thought it would be better to alert people when they'll need fix that setting and re-checkout the files rather than trying to address all of the problems caused by "messed up" line endings.

Would you prefer I try to focus on resolving those CRLF issues, or do you feel it's an acceptable requirement that LF line endings be used, and that some sort of warning to people with CRLFs would be a better approach long-term?

hkdobrev:
Can't the problem with Git be solved with a `.gitattributes` file?

alcohol:
@hkdobrev I believe so, thanks for mentioning that.

alcohol:
@forevermatt can you give my branch a try on Windows?

alcohol:
I tried using a modified regex instead. It properly splits the sections and data for the tests, however the expected output and actual output still fail cause of "#Warning: Strings contain different line endings!".

forevermatt:
@alcohol I tried pulling a fresh copy of composer/composer (which had your new ```.gitattributes``` file), but many more files than just the "tests/Composer/Test/Fixtures/installer/*.test" files were being messed up by CRLFs.  I finally tried using the ```.gitattributes``` file to tell Git not to modify line endings at all, and that resolved all the 20+ failing tests (though on my system there were still 3 skipped tests and 1 error).  I'll submit a new pull request with that change.

Seldaek:
@forevermatt I just pushed a fix for the other tests that don't rely on .json/.test files.

forevermatt:
@Seldaek Okay, thanks.  Perhaps that's why GitHub is telling me https://github.com/composer/composer/pull/4078 can't be merged automatically.  I'll re-run the tests after pulling those changes to see if there are still any issues still on my machine.

Seldaek:
Worked for me on a clean checkout with autocrlf=true, but I'd still recommend you disable that option as it'll lead to many unnecessary problems.

forevermatt:
@Seldaek Your changes resolved all the failing tests for me, too. Thanks.

PR #4050, opened 2015-05-22
Add Vagrant (and provisioning) file for consistent testing environment.

forevermatt:
Trying to get set up to simply run all of the tests (in a way where they all pass, as they do on Travis-CI) is proving rather difficult.  Setting up a Ubuntu VM (via Vagrant), tweaking some settings in php.ini, and making sure I cloned the repo with Git's core.autocrlf setting to false has gotten me most of the way there.  (Oddly enough, the only remaining failing test is that json_decode is accepting single quotes for some strange reason).

This commit provides my (mostly) working Vagrantfile and vagrant provisioning script to enable others to get (at least nearly) up and running more quickly than I did.

Since my setup isn't fully working, I won't be surprised if this pull request is rejected or requires me to make some more changes before it's accepted.  It seemed better, though, to offer this pull request now (rather than waiting for it to get behind master) in case it does seem valuable to others as-is.

I also don't know why Vagrantfile was being ignored previously, so there may be a very good reason to continue to ignore it rather than including it now (as I've done).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,5 @@
 /nbproject
 phpunit.xml
 .vagrant
-Vagrantfile
 .idea
 .php_cs.cache
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,11 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure(2) do |config|
+
+  # Define which vagrant box to use.
+  config.vm.box = "ubuntu/trusty64"
+
+  # Configure the VM as necessary so that the tests can pass.
+  config.vm.provision :shell, :path => "vagrant-provision.sh"
+end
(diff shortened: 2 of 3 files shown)


alcohol:
Please no. There are many ways to setup an environment. Composer does not (or should not) care, nor does it want to set a convention (Vagrant in this case).

:-1:

phansys:
I agree with @alcohol. It should be kept in a "agnostic environment" basis.
:-1:

forevermatt:
Thanks for the feedback. There is certainly a lot of value in avoiding locking composer into a particular environment.

My offer of a Vagrantfile was to provide an easier way for people to get started running all the tests. When I first cloned the repo, around 25 of the tests were failing (primarily due to Git converting line endings to CRLFs, but at least one related to php.ini settings).

I was merely trying to reduce the effort required for someone to get to the point of being able to offer more meaningful contributions.  Since it sounds like the preference is to work out cross-platform issues rather than providing a consistent environment, I'll try to focus my efforts in that direction.

Again, thanks for the feedback.

PR #4056, opened 2015-05-22
Remove HTTP support from composer. (Solves HTTP-related part of #1074)

StormTide:
This PR resolves the http portion of issue https://github.com/composer/composer/issues/1074 (Remote Code Execution via MITM) by disabling plain http schemes entirely. It creates a hard dependency on openssl and a valid certificate bundle. 

This is likely BC-breaking for those using insecure http urls to install code or those operating without https support installed into PHP. This is an acceptable BC break given the security/privacy implications. 

This PR has seen limited testing so is submitted for discussion only at this time.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -90,7 +90,7 @@ To edit the global config.json file:
 
 To add a repository:
 
-    <comment>%command.full_name% repositories.foo vcs http://bar.com</comment>
+    <comment>%command.full_name% repositories.foo vcs https://bar.com</comment>
 
 To remove a repository (repo is a short alias for repositories):
 
@@ -412,7 +412,7 @@ EOT
                 }
             }
 
-            throw new \RuntimeException('You must pass the type and a url. Example: php composer.phar config repositories.foo vcs http://bar.com');
+            throw new \RuntimeException('You must pass the type and a url. Example: php composer.phar config repositories.foo vcs https://bar.com');
         }
 
         // handle platform
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -90,7 +90,7 @@ To setup a developer workable version you should create the project using the so
 controlled code by appending the <info>'--prefer-source'</info> flag.
 
 To install a package from another repository than the default one you
-can pass the <info>'--repository-url=http://myrepository.org'</info> flag.
+can pass the <info>'--repository-url=https://myrepository.org'</info> flag.
 
 EOT
             )
@@ -247,10 +247,10 @@ EOT
             } else {
                 $sourceRepo = new FilesystemRepository($json);
             }
-        } elseif (0 === strpos($repositoryUrl, 'http')) {
+        } elseif (0 === strpos($repositoryUrl, 'https')) {
             $sourceRepo = new ComposerRepository(array('url' => $repositoryUrl), $io, $config);
         } else {
-            throw new \InvalidArgumentException("Invalid repository url given. Has to be a .json file or an http url.");
+            throw new \InvalidArgumentException("Invalid repository url given. Has to be a .json file or an https url.");
         }
 
         $parser = new VersionParser();
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -81,15 +81,13 @@ EOT
         $this->getIO()->write('Checking git settings: ', false);
         $this->outputResult($this->checkGit());
 
-        $this->getIO()->write('Checking http connectivity: ', false);
-        $this->outputResult($this->checkHttp());
+        $this->getIO()->write('Checking https connectivity: ', false);
+        $this->outputResult($this->checkHttps());
 
-        $opts = stream_context_get_options(StreamContextFactory::getContext('http://example.org'));
+        $opts = stream_context_get_options(StreamContextFactory::getContext('https://example.org'));
         if (!empty($opts['http']['proxy'])) {
-            $this->getIO()->write('Checking HTTP proxy: ', false);
-            $this->outputResult($this->checkHttpProxy());
-            $this->getIO()->write('Checking HTTP proxy support for request_fulluri: ', false);
-            $this->outputResult($this->checkHttpProxyFullUriRequestParam());
+            $this->getIO()->write('Checking HTTPS proxy: ', false);
+            $this->outputResult($this->checkHttpsProxy());
             $this->getIO()->write('Checking HTTPS proxy support for request_fulluri: ', false);
             $this->outputResult($this->checkHttpsProxyFullUriRequestParam());
         }
@@ -168,11 +166,10 @@ EOT
         return true;
     }
 
-    private function checkHttp()
+    private function checkHttps()
     {
-        $protocol = extension_loaded('openssl') ? 'https' : 'http';
         try {
-            $this->rfs->getContents('packagist.org', $protocol . '://packagist.org/packages.json', false);
+            $this->rfs->getContents('packagist.org', 'https://packagist.org/packages.json', false);
         } catch (\Exception $e) {
             return $e;
         }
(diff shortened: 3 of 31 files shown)


fprochazka:
Is it really neccesary to do it this aggresively? Doesn't this just cut off users with custom repositories, that don't have https? Couldn't it be done more gradually?

fprochazka:
Imagine you have a private satis repository that is accesible only inside company infrastructure. Is it still security issue, if it doesn't have https?

paragonie-scott:
I vote :+1: on removing plaintext HTTP (or "barebacking" as @ioerror likes to call it).

> Doesn't this just cut off users with custom repositories, that don't have https?

* https://startssl.com
* https://letsencrypt.org

I just listed two free SSL certificate registrars (the latter is launching this year and provides a CLI tool to quickly get certificates). **There is no excuse to not deploy HTTPS anymore.**

StormTide:
@fprochazka yes that would still suffer the same mitm potential. Even internally you shouldn't be running code over http (wifi pineapples and all)... you can always setup a self-signed cert and install the ephemeral ca in your developer machines. IMO http really isnt a valid way to distribute code in a system without a codesigning mechanism. 

Is it aggressive? yes... but maybe thats a good thing...

fprochazka:
@paragonie-scott well, wrong. Startssl not really free, since they charge you for revocation and letsencrypt haven't started yet.

fprochazka:
Full disclosure: I advocate HTTPS wherever I go and all my websites are on HTTPS+HSTS+all the fancy stuff, but this just seems too much.

Having it try to work using HTTPS by default and showing warnings when fallbacking to http would be much better IMHO.

fprochazka:
Also, the programmer is writing the repositories list in composer json. Why can't I just tell it which one I wanna use using https and which one on http?

StormTide:
@fprochazka its a weakest-link-in-the-chain problem. If you have even one http resource, anyone with a mitm can run code on your system. The problem described in #1074 is extremely serious.

fprochazka:
I understand that. Packagist and github are already on HTTPS so it's basically a question of what other repositories I use. And if I wanna setup a private satis repository that is accessible only using VPN without having to worry about setting up https, why couldn't I?

paragonie-scott:
@fprochazka 

> Startssl is not really free, since they charge you for revocation and letsencrypt haven't started yet.

Revocation isn't a problem I've run into often and letsencrypt is launching very soon. You can use startssl for now and letsencrypt when it launches. If you need to revoke certs between now and letsencrypt launches, I'll eat my hat.

StormTide:
@fprochazka actually, if you look at the pr code, http is used frequently as a fallback in composer where the system doesnt have openssl setup. This fixes those fallback-to-insecurity pain points. It also means that to publish code, folks need to either publish to packagist/github or setup a https resource. That seems like a good thing, given the ecosystem to-date.

StormTide:
Also, to be clear, this doesnt close #1074 ... there's still the php prior to 5.6 ssl peer validation issues that need to be addressed.

paragonie-scott:
I think a good solution looks like this:

1. Remove silent HTTP fallbacks.
2. Enforce HTTPS on HTTPS sources.
3. Require explicit enabling for plaintext HTTP on a per-repository basis for people where they place all of their trust in the VPN rather than endpoint security. (I don't think this is *wise* but it will satisfy @fprochazka's requirements)

StormTide:
@paragonie-scott I'm against that approach as we can see from the ecosystem of composer.json files that http schemes are used entirely too often and usually out of security ignorance, not out of some vpn edge case.

paragonie-scott:
I don't mean just using the `http://` prefix, I mean adding a parameter (somehow) that says `accept_insecure=1`, for example.

StormTide:
@paragonie-scott we see that approach fail too with the CURLOPT_SSL_VERIFYPEER nightmare. Folks will 'make it work' and then stop there, never thinking again about the security implication.

paragonie-scott:
Fair point.

Honestly, self-signed TLS with HPKP would be a better solution for the corporate network situation. Too bad [PHP's cURL extension doesn't support public key pinning](https://bugs.php.net/bug.php?id=68575)

StormTide:
I think for most folks, the requirement to put their project on github, packagist, get a market ssl certificate or setup a local ca certificate, isn't going to be that unreasonable.

tomasfejfar:
> Folks will 'make it work' and then stop there, never thinking again about the security implication.

Just name it `i_dont_care_about_security_allow_http` :) That will make them think about it. As seen in https://facebook.github.io/react/tips/dangerously-set-inner-html.html

fprochazka:
To be honest, I don't need the http fallback, I'm really just worried about new composer users who would rather not use composer at all, that fiddle with https issues.
Take for example windows users with some old root certs... composer just won't work for them and most begginers won't be able to fix this.

StormTide:
> Take for example windows users with some old root certs... composer just won't work for them and most begginers [sic] won't be able to fix this.

Again this is like the mythical edge case. I think most users will be able to install openssl and a ca bundle and that this is just basic ante for playing in the PHP ecosystem.

padraic:
@fprochazka This is probably the only significant issue I ever see around HTTP - Windows issues. Most of these are easily fixed with a few steps, two ini options and a download usually does it. My opposing argument is just as simple - don't degrade everyone's security simply to serve a minority of edge cases with obviously broken systems. It's just not the correct approach.

tomasfejfar:
@padraic make it very hard and maybe even annoying, but possible ;)

StormTide:
What like a 'http-key' and require them to calculate the md5 of the system hostname to override? I mean, it should be of equal complexity to installing a ca bundle and actually fixing, rather than overriding, the problems.

fprochazka:
@padraic it's not only windows users, I've seen badly maintained hosting/vps solutions that also didn't work because of outdated ca bundle. I have seen quite a few people starting with php, and none of them were able to solve each of those 1000 little problems, that you have when you simple don't know anything about it. I think it's important to teach people to use composer and if they don't get it working under a hour or less, most of them will just give up if they're not forced to make it work.

But hey, if you guys really think it's an edge case and these can be easily fixed then I don't mind dropping http here. I just wanted to voice the concerns :)

padraic:
@fprochazka As I say a few times, document all the things ;). Most of these issues will crop up in the issues for Composer right now anyway since the Installer script does actually work to enforce HTTPS. So you can search the issues list for `ssl` and find a host of these things to assess and improve documentation for. Given that the Installer is actually the main public means of installing Composer to begin with...it's likely a decent snapshot even. I'm not sure of % of who uses the script vs the actual Windows installer file though.

slbmeh:
The ssl issue on windows is common even with the current implementation. I believe it's common enough of an issue here to be documented on how to install/update a ca bundle on windows if it isn't already.

With that in mind, they already have issues, I advocate being aggressive.  Have a check in composer diagnose, and maybe hint at some documentation if they don't have ssl support.  The people that will struggle are the beginners, and I'd rather teach them the proper way to address it than allow them to take the easy way by disregarding security.

buzzybear69:
king Kong1969

buzzybear69:
king Kong1969

Seldaek:
Closing as the PR itself really is breaking tons of stuff, I took the good parts in that commit, and I hope that #4759 including the warning it contains when running without php_openssl enabled will be enough deterrent for the rest of the problem.

The only issue left AFAICT, is that of people adding repositories (or similar..) that are http only. I don't think we can really disable those per se because not everyone can do SSL (although letsencrypt really helps and I hope that'll allow everyone to finally move over). What I would happily consider there is printing a warning when a repo is added with http protocol, or a package repo with a http dist/source download URL, etc. What would be great is for someone to identify these places and then we can add the warnings pretty easily I believe. I don't agree it should be a hard fail but hopefully big red lines will be enough for people to react.

paragonie-security:
> The only issue left AFAICT, is that of people adding repositories (or similar..) that are http only. I don't think we can really disable those per se because not everyone can do SSL (although letsencrypt really helps and I hope that'll allow everyone to finally move over).

Would you be opposed to adding a "https-only" option to the composer.json file, default it to true, and explicitly require `"https-only": false` for them to work?

Seldaek:
I guess we can try and see the damage it does. It would indeed be easier to enforce that way in the code, except for VCS URLs that go through external processes but I don't think those are the most likely to use http..

StormTide:
I'm not sure the ref'd PR closes this so long as http support exists.  HTTP should definitely be a hard-fail as its fully a RCE, every can do ssl (thanks letsencrypt) or use github et al to share their code. Not disabling HTTP will leave the #1074 active within typical composer deployments.

PR #4061, opened 2015-05-24
Added installation using npm

jadjoubran:
## Node module
[getcomposer](https://github.com/jadjoubran/getcomposer)

## Why?
Being able to install composer from `npm` makes it easier for people/organizations already using `npm` to install composer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,21 @@ Global installation of Composer (manual)
 
 Follow instructions [in the documentation](https://getcomposer.org/doc/00-intro.md#globally)
 
+
+Installation as a Node Module (requires npm)
+--------------------------------------------
+
+Download composer using npm
+
+````shell
+#global
+npm install -g getcomposer
+
+#local
+npm install getcomposer
+````
+
+
 Updating Composer
 -----------------
 


Nek-:
Proposal: installing npm through composer. Let's add it to the readme of npm.

fprochazka:
What is the benefit?

jadjoubran:
@fprochazka if you already have npm installed, it provides a much easier way to install composer locally and/or globally.
Installing it locally allows you to save this dependency inside your package.json.

Let me know what you think

nevvermind:
I don't think we shoud tie Composer with npm. Not like this anyway. Besides, composer is already easy to install. And why would someone use _npm_ to install composer?

jadjoubran:
@nevvermind Thank you for your comment. I actually rely on the community's feedback and if you all agree that it's not useful, then it makes sense. Although some of the teams that I've worked with found it very convenient since you can memorize the command.
But I just want to add that they're not **tied** together. `getcomposer` is just an installer for the original composer source.

nevvermind:
By "tied", I mean having a Composer _tool_ place some information on readme. There are lots of tools out there and they're part of the ecosystem - and I don't want to sound mean - but why should only yours change the readme?

jadjoubran:
@nevvermind it actually makes sense! Thank you for presenting it in a mature way!

PR #4063, opened 2015-05-25
Added missing information about --keep-vcs in create-project.

ctubio:
Hi,
i found a bit confusing that --keep-vcs only works under dev, and it silently-doest-nothing under stable versions.

This PR attempts to notice the users about this behavior.

btw, why it doesnt keep the vcs files for stable versions? i cant understand this limitation.
Thank you for your time,

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -466,7 +466,7 @@ By default the command checks for the packages on packagist.org.
   terminals or scripts which don't handle backspace characters.
 * **--keep-vcs:** Skip the deletion of the VCS metadata for the created
   project. This is mostly useful if you run the command in non-interactive
-  mode.
+  mode. (only works together with: -s dev|--stability=dev|dev-master)
 * **--ignore-platform-reqs:** ignore `php`, `hhvm`, `lib-*` and `ext-*`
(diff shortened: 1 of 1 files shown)


Seldaek:
I did an alternative fix which I think removes that confusion.

ctubio:
Yes, very nice fix, thank you alot!

Also, now is clear that vcs files are only possible to be keep if installing from sources, and is not based on the stability.

gennadiylitvinyuk:
Sorry, this may be fixes this particular problem, but this is obviously not correct.

For example I want to create new project with `composer create-project`. The main project should come from SVN, and dependencies from dists.

Earlier I could use
```
composer create-project namespace/project --keep-vcs --prefer-dist -s dev
```
After this update it is impossible to get dist versions of dependent packages.

Seldaek:
@gennadiylitvinyuk oh that's true, it should only prefer-dist for the root package, but not imply anything else for the rest. I'll update that.

Seldaek:
@gennadiylitvinyuk see https://github.com/composer/composer/commit/e727ed4d7305110b5f39689e47a637be19316483

PR #4064, opened 2015-05-25
[travis-ci] Enable PHPUnit color output

Maks3w:


Seldaek:
Thanks that's much better indeed :)

PR #4065, opened 2015-05-25
Remove incorrect deprecation documentation

JeroenDeDauw:
This comment seems to be misplaced, as several of the constants
below it are not explicitly deprecated and do not seem to have
any replacements. Perhaps they should actually be explicitly
deprecated, I cannot rule that out without knowing the code better.
In any case, currently the documentation is only confusing for
people that want to use these constants, so it should be changed
in some manner.

Comment added in 3efed220a6c9fd6e4a13ea806879e124941c9fc7

### DIFF
--- a/src/Composer/Script/ScriptEvents.php
+++ b/src/Composer/Script/ScriptEvents.php
@@ -74,8 +74,6 @@ class ScriptEvents
      */
     const POST_STATUS_CMD = 'post-status-cmd';
 
-    /** Deprecated constants below */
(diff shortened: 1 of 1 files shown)


Seldaek:
No you're right it's at the wrong spot in the file not sure why.

JeroenDeDauw:
Why have the comment to begin with though? There are already deprecation tags, and clearly this type of comment is prone to fail.

PR #4067, opened 2015-05-26
Add support for overriding the archive filename

JoakimLofgren:
Instead of running a shell command to rename the archive I want to be able to specify a option to do it straight away.

`composer archive --format=zip --filename=my-package` would create a archive `my-package.zip`

### DIFF
--- a/src/Composer/Package/Archiver/ArchiveManager.php
+++ b/src/Composer/Package/Archiver/ArchiveManager.php
@@ -96,12 +96,13 @@ class ArchiveManager
      *
      * @param  PackageInterface          $package   The package to archive
      * @param  string                    $format    The format of the archive (zip, tar, ...)
-     * @param  string                    $targetDir The diretory where to build the archive
+     * @param  string                    $targetDir The directory where to build the archive
+     * @param  string                    $filename  Overrides the archive's default filename
      * @throws \InvalidArgumentException
      * @throws \RuntimeException
      * @return string                    The path of the created archive
      */
-    public function archive(PackageInterface $package, $format, $targetDir)
+    public function archive(PackageInterface $package, $format, $targetDir, $filename = null)
     {
         if (empty($format)) {
             throw new \InvalidArgumentException('Format must be specified');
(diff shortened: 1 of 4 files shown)


tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
```diff
@@ -55,6 +55,23 @@ public function testArchiveTar()
...
+
+        $package = $this->setupPackage();
+
+        $this->manager->archive($package, 'tar', $this->targetDir, 'override');
+
+        $target = $this->targetDir.'/override.tar';
+        $this->assertFileExists($target);
+
+        $tmppath = sys_get_temp_dir().'/composer_archiver/'.$this->manager->getPackageFilename($package);
+        $this->assertFileNotExists($tmppath);
```
JoakimLofgren: I copied this from the `testArchiveTar` test.
Looking at the [ArchiveManager](https://github.com/composer/composer/blob/master/src/Composer/Package/Archiver/ArchiveManager.php#L157) wouldn't this always assert to true?
Considering that the $tempTarget is using `uniqid()` and the directories doesn't seem to match anymore either ?

Commenting out the clean up steps in the ArchiveManager seems to still give the same test results.

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -123,6 +124,9 @@ public function archive(PackageInterface $package, $format, $targetDir)
 
         $filesystem = new Filesystem();
         $packageName = $this->getPackageFilename($package);
+        if ($filename) {
+            $packageName = $filename;
```
stof: you could avoid building the default filename entirely in thi case

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
Seldaek: Yes that seems like a useless assertion indeed. It's kinda hard to test for given the uniqid so I'd just remove the assertion altogether.

Seldaek:
Ping @JoakimLofgren - any chance you wrap up those few things that were commented upon above?

JoakimLofgren:
Oh sorry, had kinda forgotten about this :)
I'll take a look at it tomorrow.

JoakimLofgren:
Wasn't this implemented by #4479 ?
So this PR can be closed ?

Seldaek:
Oh god yes sorry I missed that :)

JoakimLofgren:
No problem, at least someone fixed the filename feature ;)

PR #4069, opened 2015-05-27
Add archive configuration

tjamet:
Add some configuration values to be able to customize the archive command behaviour in `$COMPOSER_HOME/.composer/config.json`

Thus a content of 
```json
{
    "repositories": [
        {
            "type": "artifact",
            "url": "/var/lib/composer/artifacts"
        }
    ],
    "config": {
        "archive-format":"zip",
        "archive-dir":"/var/lib/composer/artifacts",
    }
}
```
would make  `composer archive` to build by default `zip` archives to be built into `/var/lib/composer/artifacts`
but `composer archive --dir=. --format=tar` would build archives locally in `tar` format

### DIFF
--- a/src/Composer/Command/ArchiveCommand.php
+++ b/src/Composer/Command/ArchiveCommand.php
@@ -33,16 +33,18 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 class ArchiveCommand extends Command
 {
+    private $config;
     protected function configure()
     {
+        $this->config = Factory::createConfig();
         $this
             ->setName('archive')
             ->setDescription('Create an archive of this composer package')
             ->setDefinition(array(
                 new InputArgument('package', InputArgument::OPTIONAL, 'The package to archive instead of the current project'),
                 new InputArgument('version', InputArgument::OPTIONAL, 'A version constraint to find the package to archive'),
-                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the resulting archive: tar or zip', 'tar'),
-                new InputOption('dir', false, InputOption::VALUE_REQUIRED, 'Write the archive to this directory', '.'),
+                new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Format of the resulting archive: tar or zip', null),
+                new InputOption('dir', false, InputOption::VALUE_OPTIONAL, 'Write the archive to this directory', null),
             ))
             ->setHelp(<<<EOT
 The <info>archive</info> command creates an archive of the specified format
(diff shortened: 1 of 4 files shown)


gam3runner:
Will be useful !

Seldaek:
I don't really like the idea of adding new config settings for everything. Archiving isn't really something you do on a regular basis I'd imagine, or if that's the case it can probably be part of a build script of some kind containing these --dir/--format values?

tjamet:
you can always do scripting on top of anything
The purpose is to handle libraries shared by several projects, inspired from the mavenLocal repository with maven
For instance, you are using 'sql-formatter' in your own project and need to add a feature
In such a case, it sounds easy to me to create the archive with local modifications and then update your project for integration tests.
Once everything looks good to you, you cleanup both repos and commits and submit PRs
Maybe there is another exciting way to solve shared components development issues i'd love to hear about?

Seldaek:
Ok I see, interesting. If you do that you'd quickly end up with a rather huge mess in that artifact repo no? 

As for alternative ways, right now the best way I guess is to install the sql-formatter package from source (rm -rf vendor/sql-formatter && composer install --prefer-source) and then you can develop it inline in the vendor dir, commit from there when you're done and then you can update the parent repo to bring its composer.lock in sync. The issue if you use the artifact repo approach is the composer.lock will contain paths to that instead of the original paths so you'd have to disable it to produce a composer.lock everyone can use.

tjamet:
Things becomes even more complicated when you considere updating project 'A' and checking integration  in 'B' and 'C'
Moreover, such a feature enables user to export archives by default in the format they want (tar,zip,...) and even to the subfolder they want ( ./archives for instance)

src/Composer/Command/ArchiveCommand.php
```diff
@@ -35,14 +35,15 @@ class ArchiveCommand extends Command
...
         $this
             ->setName('archive')
             ->setDescription('Create an archive of this composer package')
             ->setDefinition(array(
                 new InputArgument('package', InputArgument::OPTIONAL, 'The package to archive instead of the current project'),
                 new InputArgument('version', InputArgument::OPTIONAL, 'A version constraint to find the package to archive'),
-                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the resulting archive: tar or zip', 'tar'),
-                new InputOption('dir', false, InputOption::VALUE_REQUIRED, 'Write the archive to this directory', '.'),
+                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the resulting archive: tar or zip', $config->get('archive-format')),
+                new InputOption('dir', false, InputOption::VALUE_REQUIRED, 'Write the archive to this directory', $config->get('archive-dir')),
```
Seldaek: Could you use null defaults here, and instead create the config and set the option values in execute in case they are null? In some cases (like the global command) the config might change between configure() and execute() are called so this code here might induce weird hard to trace behaviors.

Seldaek:
Alright fair enough, if you can just fix the thing I commented on I think it's good to merge then.

tjamet:
i've updated the PR to considere your remarks

Seldaek:
Thanks

PR #4081, opened 2015-05-29
[travis-ci] Remove hhvm-nightly

Maks3w:
HHVM Nightly is not supported by the current OS used by Travis-CI


alcohol:
It doesn't break the build though. No need to remove it only to add it back again at some future point in time.

Issue #3545, opened 2014-12-12
Extract methods from VersionParser.php into it's own package


nickvergessen:
[Composer\Package\Version\VersionParser](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php) has some good methods dealing with versions which would be nice to have in a seperate project, so they can be reused instead of either copy pasting them or relying on the whole composer package.

This especially applies to the following methods:
* [parseStability()](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L37)
* [normalizeStability()](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L65)
* [normalize()](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L86)
* [normalizeBranch()](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L86)
* [manipulateVersionString()](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L445)
* [expandStability()](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L479)
* [parseNameVersionPairs()](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionParser.php#L498)

legoktm:
+1

I'm specifically interested in doing this for MediaWiki, where we would like to re-use composer's version parsing code instead of writing our own. I think we would want to include `VersionParser` and all the classes implementing `LinkConstraintInterface`.

VersionParser currently depends on other parts of composer outside of LinkConstraintInterface in:
* parseConstraints - BasePackage::$stabilities
* parseConstraint - BasePackage::$stabilities
* parseLinks - returns Link's
* formatVersion - uses PackageInterface

I don't think it would be too difficult to move those methods somewhere else, and I would be willing to work on a patch that did that.

legoktm:
- [ ] parseConstraints - BasePackage::$stabilities - #4213 
- [ ] parseConstraint - BasePackage::$stabilities - #4213 
- [ ] parseLinks - returns Link's - #4214 
- [x] formatVersion - uses PackageInterface #4086

legoktm:
I think we can close this now that https://github.com/composer/semver exists :)

Seldaek:
https://github.com/composer/semver

PR #4086, opened 2015-05-31
Move VersionParser::formatVersion() to BasePackage::getFullPrettyVersion()

legoktm:
Working towards #3545.

formatVersion() does not belong in VersionParser since it depends upon a
Package object, and is creating a more complete pretty formatted
version, not parsing anything.

The new getFullPrettyVersion() method can be seen as an extension to
getPrettyVersion(), and is located in BasePackage as a result.

I'm also not very good at naming things, and `getFullPrettyVersion` was the best I could come up with. Better suggestions welcome :)

### DIFF
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -69,8 +69,14 @@ class VersionParser
         return $stability === 'rc' ? 'RC' : $stability;
     }
 
+    /**
+     * @deprecated Use PackageInterface::getFullPrettyVersion instead
+     */
     public static function formatVersion(PackageInterface $package, $truncate = true)
     {
+        trigger_error(__METHOD__.' is deprecated. Use '.
+            '\Composer\Package\PackageInterface::getFullPrettyVersion() instead', E_USER_DEPRECATED);
+
         return $package->getFullPrettyVersion($truncate);
     }
 
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -71,16 +71,7 @@ class VersionParser
 
     public static function formatVersion(PackageInterface $package, $truncate = true)
     {
-        if (!$package->isDev() || !in_array($package->getSourceType(), array('hg', 'git'))) {
-            return $package->getPrettyVersion();
-        }
-
-        // if source reference is a sha1 hash -- truncate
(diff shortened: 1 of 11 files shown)


src/Composer/Package/Version/VersionParser.php
```diff
@@ -71,16 +71,7 @@ public static function normalizeStability($stability)
...
-            return $package->getPrettyVersion();
-        }
-
-        // if source reference is a sha1 hash -- truncate
-        if ($truncate && strlen($package->getSourceReference()) === 40) {
-            return $package->getPrettyVersion() . ' ' . substr($package->getSourceReference(), 0, 7);
-        }
-
-        return $package->getPrettyVersion() . ' ' . $package->getSourceReference();
+        return $package->getFullPrettyVersion($truncate);
```
Seldaek: It's ok for me to keep this around but I think it should do a `trigger_error('VersionParser::formatVersion is deprecated, you should use $package->getFullPrettyVersion() instead', E_USER_DEPRECATED);`

nevvermind: Indeed. Along with a `@deprecated` annotation. That way you could easily catch it in Composer 2.0.

legoktm: Done both.

src/Composer/Package/Version/VersionParser.php
```diff
@@ -69,18 +69,14 @@ public static function normalizeStability($stability)
...
-
-        // if source reference is a sha1 hash -- truncate
-        if ($truncate && strlen($package->getSourceReference()) === 40) {
-            return $package->getPrettyVersion() . ' ' . substr($package->getSourceReference(), 0, 7);
-        }
-
-        return $package->getPrettyVersion() . ' ' . $package->getSourceReference();
+        trigger_error(__METHOD__.' is deprecated. Use '.
+            '\Composer\Package\PackageInterface::getFullPrettyVersion() instead', E_USER_DEPRECATED);
+        return $package->getFullPrettyVersion($truncate);
```
alcohol: Empty line before return construct.

legoktm: Fixed.

legoktm:
Fixed code style, rebased against master, and then fixed failing tests due to a missing namespace import.

Seldaek:
Thanks :)

PR #4088, opened 2015-05-31
Fix "Invalid argument supplied for foreach()" on a new PlaformRepository obj

nevvermind:
This happens on `new PlatformRepository(array($somePackage))`.

The parent constructor calls `\Composer\Repository\ArrayRepository::addPackage()`, which, on a brand new repo object, further calls `\Composer\Repository\PlatformRepository::initialize()` and finally this iterates over a `NULL` `$this->overrides`, triggering the error.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -35,10 +35,10 @@ class PlatformRepository extends ArrayRepository
 
     public function __construct(array $packages = array(), array $overrides = array())
     {
-        parent::__construct($packages);
         foreach ($overrides as $name => $version) {
             $this->overrides[strtolower($name)] = array('name' => $name, 'version' => $version);
         }
+        parent::__construct($packages);
     }
 
     protected function initialize()
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -31,12 +31,11 @@ class PlatformRepository extends ArrayRepository
      *
      * @var array
      */
-    private $overrides;
+    private $overrides = array();
 
     public function __construct(array $packages = array(), array $overrides = array())
     {
         parent::__construct($packages);
-        $this->overrides = array();
         foreach ($overrides as $name => $version) {
             $this->overrides[strtolower($name)] = array('name' => $name, 'version' => $version);
         }


Seldaek:
IMO we should set up the overrides BEFORE calling the parent constructor in this case, otherwise it'll be initialized wrong. Or maybe we don't forward packages to the parent.. what's the use case for initializing it with some packages? Is this happening in composer itself?

nevvermind:
Yes, the alternative was to initialize the array just before the parent constructor call. But why not properly initialize it without depending on it's location? This happened while playing with unit testing, not while using composer.

stof:
I agree with the current change. It seems sensible to me (and btw, this is what we do in Symfony too)

Seldaek:
I don't disagree with the change, but initialize **needs** the overrides from the constructor to be set up for it to do its job correctly, so I do think that should be moved above the parent::__construct() call.

nevvermind:
I still don't get how a compile-time initialization isn't safer than a runtime one. It's not a dynamic value anyway, to justify a runtime init. It's just an array. The sooner it's set, the better. You still want me to change it?

src/Composer/Repository/PlatformRepository.php
```diff
@@ -31,12 +31,11 @@ class PlatformRepository extends ArrayRepository
...
-    private $overrides;
+    private $overrides = array();
 
     public function __construct(array $packages = array(), array $overrides = array())
     {
         parent::__construct($packages);
-        $this->overrides = array();
         foreach ($overrides as $name => $version) {
             $this->overrides[strtolower($name)] = array('name' => $name, 'version' => $version);
         }
```
Seldaek: I feel like I am not getting through today :p What I mean is that the three lines above here, are used below at line 52 inside initialize(). So these 3 lines need to happen before the parent constructor is called and therefore initialize is triggered. I don't care how $overrides is initialized to an array in the first place :)

nevvermind: > I don't care how $overrides is initialized to an array in the first place :)  

Well, that's one reason you didn't get through: you didn't talked about the subject at hand - the array init. I assumed you did.

I'll amend accordingly.

Seldaek: Yeah sorry I probably wasn't clear enough, anyway looks good now thanks :)

Issue #4085, opened 2015-05-30
Plugins can only declare one exact Plugin API version


nevvermind:
### The problem
Currently, plugin developers can only specify an **exact** version of the Plugin API on which their functionality depends:
```json
{ "require": { "composer-plugin-api": "1.0.0" } }
```
### Why is it a problem
Scenario: Plugin `PLUG1` was originally created with `"composer-plugin-api": "1.0.0"`. After a while, the Composer plugin API evolves, changing its version to `1.1.0`. Because `1.1.0` implies no BC break, `PLUG1` should still work with this version. But it cannot specify that.  
The developer can only set _either_ `1.0.0` (possibly breaking when used with a newer Composer), or `1.1.0` (possibly breaking with an older Composer). Ideally, the developer should be able to mention that `PLUG1` works with _both_ `1.0.0` and `1.1.0`. Or even up until `1.9.9`. Or, if generic enough, across major versions of the plugin API.

One other consequence is that, on the next Plugin API _minor_ version bump, _all_ plugins will trigger a warning (Composer still registers them).  

### Considerations
1. **BC!** :grimacing: Whatever we end up changing, the old style of declaring an API version (`"composer-plugin-api": "1.0.0"`) *must* continue to function as expected. And by "expected" I mean, up until - and including - `1.9.9.9-stable`.
2. From looking in the codebase, plugins are still registered, even if the version didn't match, triggering a warning. This behaviour should be changed, as the next major Plugin API version could, by definition, break old plugins if they're allowed to be loaded. Mismatches should _not_ be registered.
3. Some plugins _can_ be so generic that they would work even across _major_ API version changes. We should avoid automatically assuming that a plugin made initially for `1.0.0` will not work with `2.3.4` or `12.0.1`. This is the developer's say.

#### 1. API version as API version _interval_ (constrained)
@Seldaek said [this](https://github.com/composer/composer/pull/3377#issuecomment-76076235):
> Regarding the version that is indeed a problem but I think it's one we need to fix in a different way, we probably need to provide the API in version <=1.1.0 and then load all plugins that require those versions, because technically you could install a new plugin and then run it with an old composer and it should just skip plugins that are installed but not loadable instead of crashing.

[...]

PR #4089, opened 2015-05-31
Make plugins have actual constraints instead of fixed versions

nevvermind:
Related issue: https://github.com/composer/composer/issues/4085

Instead of developing plugins against a single, fixed Plugin API version - `"composer-plugin-api": "1.0.0"`, this change will allow plugin developers to use versions like `"composer-plugin-api": "~1.1"` or `"composer-plugin-api": ">=2.1 <3.0"`, aka actual Composer-compatible constraints.

To maintain BC, old requires like `"composer-plugin-api": "1.0.0"` will be internally changed to mean `"composer-plugin-api": "^1.0"`.

Because of this new constraint flexibility, plugin version mismatches will be skipped, which means those plugin will NOT be registered to the system. Previously, a mismatch triggered a warning, but plugins were still registered.

### DIFF
--- a/doc/articles/plugins.md
+++ b/doc/articles/plugins.md
@@ -16,7 +16,7 @@ specific logic.
 
 ## Creating a Plugin
 
-A plugin is a regular composer package which ships its code as part of the
+A plugin is a regular Composer package which ships its code as part of the
 package and may also depend on further packages.
 
 ### Plugin Package
@@ -24,23 +24,36 @@ package and may also depend on further packages.
 The package file is the same as any other package file but with the following
 requirements:
 
-1. the [type][1] attribute must be `composer-plugin`.
-2. the [extra][2] attribute must contain an element `class` defining the
+1. The [type][1] attribute must be `composer-plugin`.
+2. The [extra][2] attribute must contain an element `class` defining the
    class name of the plugin (including namespace). If a package contains
-   multiple plugins this can be array of class names.
+   multiple plugins, this can be array of class names.
+3. You must require the special package called `composer-plugin-api`
+   to define which Plugin API versions your plugin is compatible with.
 
-Additionally you must require the special package called `composer-plugin-api`
-to define which composer API versions your plugin is compatible with. The
-current composer plugin API version is 1.0.0.
+The required version of the `composer-plugin-api` follows the same [rules][7]
+as a normal package's, except for the `1.0`, `1.0.0` and `1.0.0.0` _exact_ 
+values. In only these three cases, Composer will assume your plugin 
+actually meant `^1.0` instead. This was introduced to maintain BC with 
+the old style of declaring the Plugin API version.  
+  
+In other words, `"require": { "composer-plugin-api": "1.0.0" }` means
+`"require": { "composer-plugin-api": "^1.0" }`.
 
-For example
+The current composer plugin API version is 1.0.0.
+
+An example of a valid plugin `composer.json` file (with the autoloading 
+part omitted):
 
 ```json
 {
     "name": "my/plugin-package",
     "type": "composer-plugin",
     "require": {
-        "composer-plugin-api": "1.0.0"
+        "composer-plugin-api": "~1.0"
+    },
+    "extra": {
+        "class": "My\\Plugin"
     }
 }
 ```
@@ -149,3 +162,4 @@ local project plugins are loaded.
 [4]: https://github.com/composer/composer/blob/master/src/Composer/Composer.php
 [5]: https://github.com/composer/composer/blob/master/src/Composer/IO/IOInterface.php
 [6]: https://github.com/composer/composer/blob/master/src/Composer/EventDispatcher/EventSubscriberInterface.php
+[7]: ../01-basic-usage.md#package-versions
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -229,7 +229,7 @@ class VersionParser
             }
 
             // if the required Plugin API version is exactly "1.0.0", convert it to "^1.0", to keep BC
-            if ('composer-plugin-api' === $target && $this->isOldStylePluginApiVersion($constraint)) {
+            if ('composer-plugin-api' === strtolower($target) && $this->isOldStylePluginApiVersion($constraint)) {
                 $parsedConstraint = $this->parseConstraints('^1.0');
             }
 
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -246,7 +246,7 @@ class VersionParser
     private function isOldStylePluginApiVersion($requiredPluginApiVersion)
     {
         // catch "1.0", "1.0.0", "1.0.0.0" etc.
-        return (bool) preg_match('#^1(\.0)+$#', trim($requiredPluginApiVersion));
+        return (bool) preg_match('#^1(\.0)++$#', trim($requiredPluginApiVersion));
     }
 
     /**
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -227,10 +227,26 @@ class VersionParser
             } else {
                 $parsedConstraint = $this->parseConstraints($constraint);
             }
+
+            // if the required Plugin API version is exactly "1.0.0", convert it to "^1.0", to keep BC
+            if ('composer-plugin-api' === $target && $this->isOldStylePluginApiVersion($constraint)) {
+                $parsedConstraint = $this->parseConstraints('^1.0');
+            }
+
             $res[strtolower($target)] = new Link($source, $target, $parsedConstraint, $description, $constraint);
         }
 
         return $res;
+    }
+
+    /**
+     * @param string $requiredPluginApiVersion
+     * @return bool
+     */
+    private function isOldStylePluginApiVersion($requiredPluginApiVersion)
+    {
+        // catch "1.0", "1.0.0", "1.0.0.0" etc.
+        return (bool) preg_match('#^1(\.0)+$#', trim($requiredPluginApiVersion));
     }
 
     /**
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -130,7 +130,7 @@ class PluginManager
                     throw new \RuntimeException("Plugin ".$package->getName()." is missing a require statement for a version of the composer-plugin-api package.");
                 }
 
-                $currentPluginApiVersion    = $this->getPluginApiVersion();
+                $currentPluginApiVersion = $this->getPluginApiVersion();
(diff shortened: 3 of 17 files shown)


nevvermind:
I'll leave this out in the open to reviewers for now. 
I'm trying to finish the unit tests, but I cound't find a way to simulate another Plugin API (in [here](https://github.com/composer/composer/blob/master/tests/Composer/Test/Plugin/PluginInstallerTest.php)).

I've tried things like
```php
/** @var CompletePackage $pluginApiInternalPackage */
$pluginApiInternalPackage = $this->getPackage(
    'composer-plugin-api',
   '6.6.6',
    '\Composer\Package\CompletePackage'
);

$this->repository
     ->expects($this->any())
     ->method('getPackages')
     ->will($this->returnValue(array($pluginApiInternalPackage)));
```
, but no dice.

Pretty close though. I need a break now. Too frustrating.

src/Composer/Package/Version/VersionParser.php
```diff
@@ -234,6 +240,16 @@ public function parseLinks($source, $sourceVersion, $description, $links)
     }
 
     /**
+     * @param string $requiredPluginApiVersion
+     * @return bool
+     */
+    private function isOldStylePluginApiVersion($requiredPluginApiVersion)
+    {
+        // catch "1.0", "1.0.0", "1.0.0.0" etc.
+        return (bool) preg_match('#^1\.0(\.0)*$#', trim($requiredPluginApiVersion));
```
nevvermind: Maybe use `^1(\.0)+$` instead.

nevvermind:
This is ready to be reviewed.

One concern (which I'll happily refactor, if needed): I needed a way to mock the Plugin API version. You can't do that with a class constant. So I took the liberty of adding [a new public method to the Config object](https://github.com/composer/composer/pull/4089/files#diff-577304d6b2175e0b5c3fe90ff0a78239R231). This is where I thought it would be best. The problem is that now you have to get through the Config object to get the API version. It's quite universal (the Config obj), it's easy to reach, so it shouldn't be _that_ big a problem.

When (If? :disappointed: ) we finish this review iteration, I'll continue with the next one: documentation.

Seldaek:
Looks good to me. Do you want to add docs to this PR before I merge? If so please don't overwrite the current commit so it can remained in a "reviewed" state in my head ;)

Seldaek:
As for the config hack, yeah it's not great and ideally we shouldn't need this.. I wonder if it wouldn't be less invasive to have a protected getPluginApiVersion in the PluginManager so that one can be mocked when testing it? It seems to be the only place where it matters.

src/Composer/Package/Version/VersionParser.php
```diff
@@ -234,6 +240,16 @@ public function parseLinks($source, $sourceVersion, $description, $links)
     }
 
     /**
+     * @param string $requiredPluginApiVersion
+     * @return bool
+     */
+    private function isOldStylePluginApiVersion($requiredPluginApiVersion)
+    {
+        // catch "1.0", "1.0.0", "1.0.0.0" etc.
+        return (bool) preg_match('#^1(\.0)+$#', trim($requiredPluginApiVersion));
```
stof: you could use ``++`` to make the quantifier possessive. It will match the same things in this case, but without trying backtracking.

src/Composer/Plugin/PluginManager.php
```diff
@@ -129,14 +130,17 @@ public function loadRepository(RepositoryInterface $repo)
                     throw new \RuntimeException("Plugin ".$package->getName()." is missing a require statement for a version of the composer-plugin-api package.");
                 }
 
-                if (!$requiresComposer->matches(new VersionConstraint('==', $this->versionParser->normalize(PluginInterface::PLUGIN_API_VERSION)))) {
-                    $this->io->writeError("<warning>The plugin ".$package->getName()." requires a version of composer-plugin-api that does not match your composer installation. You may need to run composer update with the '--no-plugins' option.</warning>");
+                $currPluginApiVersion = $this->composer->getConfig()->getPluginApiVersion();
+                $currPluginApiConstraint = new VersionConstraint('==', $this->versionParser->normalize($currPluginApiVersion));
```
stof: please don't abbreviate current to curr. It is less readable.

tests/Composer/Test/Plugin/Fixtures/plugin-v6/Installer/Plugin.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Installer;
+
+use Composer\Composer;
+use Composer\IO\IOInterface;
+use Composer\Plugin\PluginInterface;
+
+class Plugin implements PluginInterface
```
stof: IMO, you should use separate class names each time to avoid issues

src/Composer/Plugin/PluginManager.php
```diff
@@ -129,14 +130,17 @@ public function loadRepository(RepositoryInterface $repo)
...
+                if (!$requiresComposer->matches($currPluginApiConstraint)) {
+                    $this->io->writeError('<warning>The "' . $package->getName() . '" plugin was skipped because it requires a Plugin API version ("' . $requiresComposer->getPrettyString() . '") that does not match your Composer installation ("' . $currPluginApiVersion . '"). You may need to run composer update with the "--no-plugins" option.</warning>');
+                    continue;
```
stof: AFAICT, the only real change in this block (except the retrieval of the plugin API version in a mockable way) is to avoid installing non-matching plugins. am I right ?

stof: and this would mean that this is already supported

nevvermind: Yes, a) omitting plugins which didn't match and b) changing the warning message to be a little more informative.

stof: but the PR title says it adds support for constraints, and the code change shows me that this is already supported in Composer without the code change

stof:
couldn't we write the tests a bit differently, avoiding the need to mock the plugin api version, and so the need to have it in the Config object ?

nevvermind:
And keeping it a constant? Maybe. No idea how, though.

Seldaek:
How about the suggestion I wrote above? https://github.com/composer/composer/pull/4089#issuecomment-107984940 - if nobody reads it that'll teach me to review things ;)

nevvermind:
I saw that, @Seldaek :)
I was assuming @stof wanted to keep the constant, as it's the sexiest.

Seldaek:
Well my solution keeps the constant except there's a protected accessor for that constant in the PluginManager so it's mockable, but it doesn't leak to the public interface and the rest of composer.

nevvermind:
Ok, I'll do that. But you're not _keeping_ the constant, not in the functional sense. It's there for BC only. Otherwise, if people will use the constant instead of the method, there's no use for a method.

In v2, the constant should be removed.

Seldaek:
No, I say *keep* the constant, it's only *within* the PluginManager that 
you should read the constant via the accessor, and not read it directly. 
The rest of composer can hit the constant directly and work just fine.

nevvermind:
I see. Not _that_ encapsulated, but yeah, it'll work.

Let's see my tests behave, as the method is used in https://github.com/composer/composer/pull/4089/files#diff-577304d6b2175e0b5c3fe90ff0a78239R71 as well.

doc/articles/plugins.md
```diff
@@ -24,23 +24,36 @@ package and may also depend on further packages.
...
 ```json
 {
     "name": "my/plugin-package",
     "type": "composer-plugin",
     "require": {
-        "composer-plugin-api": "1.0.0"
+        "composer-plugin-api": "~1.0"
+    },
+    "extra": {
+        "class": "My\\Plugin"
```
nevvermind: I think the `extra` bit is important when showing an example.

doc/articles/plugins.md
```diff
@@ -24,23 +24,36 @@ package and may also depend on further packages.
...
+An example of a valid plugin `composer.json` file (with the autoloading 
+part omitted):
 
 ```json
 {
     "name": "my/plugin-package",
     "type": "composer-plugin",
     "require": {
-        "composer-plugin-api": "1.0.0"
+        "composer-plugin-api": "~1.0"
```
nevvermind: The example Plugin API I used tries to scare away people from using `1.0.0` anymore and express that it's safe to use normal version strings.

nevvermind:
@Seldaek and/or @stof - Refactored and updated the docs. I'd appreciate another gander at this.

src/Composer/Package/Version/VersionParser.php
```diff
@@ -227,6 +227,12 @@ public function parseLinks($source, $sourceVersion, $description, $links)
             } else {
                 $parsedConstraint = $this->parseConstraints($constraint);
             }
+
+            // if the required Plugin API version is exactly "1.0.0", convert it to "^1.0", to keep BC
+            if ('composer-plugin-api' === $target && $this->isOldStylePluginApiVersion($constraint)) {
```
stof: you should compare ``'composer-plugin-api'`` to the *lowercased* target

src/Composer/Plugin/PluginManager.php
```diff
@@ -129,14 +130,18 @@ public function loadRepository(RepositoryInterface $repo)
                     throw new \RuntimeException("Plugin ".$package->getName()." is missing a require statement for a version of the composer-plugin-api package.");
                 }
 
-                if (!$requiresComposer->matches(new VersionConstraint('==', $this->versionParser->normalize(PluginInterface::PLUGIN_API_VERSION)))) {
-                    $this->io->writeError("<warning>The plugin ".$package->getName()." requires a version of composer-plugin-api that does not match your composer installation. You may need to run composer update with the '--no-plugins' option.</warning>");
+                $currentPluginApiVersion    = $this->getPluginApiVersion();
```
stof: assignments should not be aligned

src/Composer/Plugin/PluginManager.php
```diff
@@ -129,14 +130,18 @@ public function loadRepository(RepositoryInterface $repo)
...
+                    $this->io->writeError('<warning>The "' . $package->getName() . '" plugin was skipped because it requires a Plugin API version ("' . $requiresComposer->getPrettyString() . '") that does not match your Composer installation ("' . $currentPluginApiVersion . '"). You may need to run composer update with the "--no-plugins" option.</warning>');
+                    continue;
                 }
 
                 $this->registerPackage($package);
-            }
+
             // Backward compatibility
-            if ('composer-installer' === $package->getType()) {
+            } elseif ('composer-installer' === $package->getType()) {
```
stof: this change is not needed, and it should be left out IMO

nevvermind: The change is to make the code look intuitive.
 
Two `if`s on the same entity express conjunction (two different things at the same time), but `$package->getType()` cannot be two things at the same time. Actually, this is textbook `case` here, but I reckoned an `elseif` should suffice. What do you say?

Seldaek:
Thanks for all the updates! This is a lot more contained change, much better.

PR #4091, opened 2015-06-01
write progress (of which composer.json is read) on stdout instead stderr

maik-burkert:
The output which composer.json is currently read should be on stdout instead of stderr.

Otherwise some systems behave unexpectedly because they assume an error occurred. For example the crontab sends an email.

The following test fails, but also fails on master, so it's not introduced by our change:

```
There was 1 failure:

1) Composer\Test\Util\RemoteFilesystemTest::testCaptureAuthenticationParamsFromUrl
Failed asserting that 0 matches expected 404.
```

Thanks.

### DIFF
--- a/src/Composer/Repository/VcsRepository.php
+++ b/src/Composer/Repository/VcsRepository.php
@@ -134,9 +134,9 @@ class VcsRepository extends ArrayRepository
         foreach ($driver->getTags() as $tag => $identifier) {
             $msg = 'Reading composer.json of <info>' . ($this->packageName ?: $this->url) . '</info> (<comment>' . $tag . '</comment>)';
             if ($verbose) {
-                $this->io->writeError($msg);
+                $this->io->write($msg);
             } else {
-                $this->io->overwriteError($msg, false);
+                $this->io->overwrite($msg, false);
             }
 
             // strip the release- prefix from tags if present
@@ -198,9 +198,9 @@ class VcsRepository extends ArrayRepository
         foreach ($driver->getBranches() as $branch => $identifier) {
             $msg = 'Reading composer.json of <info>' . ($this->packageName ?: $this->url) . '</info> (<comment>' . $branch . '</comment>)';
             if ($verbose) {
-                $this->io->writeError($msg);
+                $this->io->write($msg);
             } else {
-                $this->io->overwriteError($msg, false);
+                $this->io->overwrite($msg, false);
             }
 
             if (!$parsedBranch = $this->validateBranch($branch)) {


alcohol:
Negative, it should not be on `stdout`.

Fix your test instead.

BreiteSeite:
> Negative, it should not be on stdout.

Why? It's no error information, but progress updates.

> Fix your test instead.

Which one? We have not added or modified tests. The test which fails, fails also on master, so we won't fix it because we ain't broke it.

alcohol:
`stderr` is not only for errors.

Please see [#1905 (comment)](https://github.com/composer/composer/issues/1905#issuecomment-41631659) and [#3715 (comment)](https://github.com/composer/composer/pull/3715#issuecomment-73271923)

alcohol:
Also, your change and the test you mentioned are completely unrelated.

BreiteSeite:
> Also, your change and the test you mentioned are completely unrelated.

Yes, we did this to explicitly inform you there is a failing test, but that it's not failing because of this PR/change.

alcohol:
I don't understand what you are talking about. I have no failing tests.

```
rob@emac019 /srv/git/github/composer (master) $ vendor/bin/phpunit
PHPUnit 4.6.6 by Sebastian Bergmann and contributors.

Configuration read from /srv/git/github/composer/phpunit.xml.dist

.............................................................   61 / 1345 (  4%)
.............................................................  122 / 1345 (  9%)
.............................................................  183 / 1345 ( 13%)
.............................................................  244 / 1345 ( 18%)
.............................................................  305 / 1345 ( 22%)
.............................................................  366 / 1345 ( 27%)
.............................................................  427 / 1345 ( 31%)
......................S......................................  488 / 1345 ( 36%)
.............................................................  549 / 1345 ( 40%)
.............................................................  610 / 1345 ( 45%)
.............................................................  671 / 1345 ( 49%)
.............................................................  732 / 1345 ( 54%)
.............................................................  793 / 1345 ( 58%)
.............................................................  854 / 1345 ( 63%)
.............................................................  915 / 1345 ( 68%)
.............................................................  976 / 1345 ( 72%)
............................................................. 1037 / 1345 ( 77%)
............................................................. 1098 / 1345 ( 81%)
............................................................. 1159 / 1345 ( 86%)
............................................................. 1220 / 1345 ( 90%)
............................................................. 1281 / 1345 ( 95%)
............................................................. 1342 / 1345 ( 99%)
...

Time: 11.45 seconds, Memory: 106.25Mb

OK, but incomplete, skipped, or risky tests!
Tests: 1345, Assertions: 2092, Skipped: 1.
```

Seldaek:
Sorry but we are not going to change the output to stdout because broken error detection takes stderr output as an error. If you want to check for errors you should look at the process exit code. If "some systems" behave incorrectly then you need to look at those and fix them to follow unix conventions.

PR #4095, opened 2015-06-02
Added no-interaction option to the list of composer install options

mickaelandrieu:
Hi,
 this option is available through the ``composer install`` command but non visible in list and I always try ``--no-interactive`` (don't know why). So ... I think maybe some others have also some difficulties to remember this option, and I suggest to display it on the list.


mickaelandrieu:
my bad this won't fix anything, the issue is "how to display  ``no-interaction`` option when we have a mistake in command, but it's not composer related.

The option appears when doing ``composer install -h``.

Again, sorry.

PR #4096, opened 2015-06-02
Fix hyphen range constraint

alcohol:


hannesvdvreken:
Update semver.mwl.be to reflect this change.

http://semver.mwl.be/#?package=phpunit%2Fphpunit&version=3.7%20-%204.0&minimum-stability=stable

PR #4100, opened 2015-06-03
load xml as string from local file, fixes #3942

### LINKED ISSUE #3942: pear-pear.php.net packages install fails on Windows
When I'm trying to install this project with a PEAR dependancy:
```
...
"repositories": [
    {
        "type": "pear",
        "url": "http://pear.php.net"
    }    
  ],
  "require": {
    "php": ">=5.4",
    "pear-pear.php.net/Services_GeoNames":"*"    
  } 
...
```
 i'm getting the following error:
```
Initializing PEAR repository http://pear.php.net
  - Installing pear-pear.php.net/net_url2 (2.1.1)
    Loading from cache

[UnexpectedValueException]
Failed to extract PEAR package C:\Users\Username\project\vendor/pear-pear.php.net/Net_URL2/Net_URL2-2.1.1.tgz to C:\Users\Username\project\vendor/pear-pear.php.net/Net_URL2. 
Reason: simplexml_load_file(): I/O warning : failed to load external entity "C:\Users\Username\project\vendor/pear-pear.php.net/Net_URL2/tarball/package.xml"

[ErrorException]
simplexml_load_file(): I/O warning : failed to load external entity "C:\Users\Username\project\vendor/pear-pear.php.net/Net_URL2/tarball/package.xml"

update [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--lock] [--no-plugins] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--with-dependencies] [-v|vv|v
vv|--verbose] [-o|--optimize-autoloader] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [packages1] ... [packagesN]
```

I'm on a Windows 8.1 machine and I noticed that the path passed to simplexml_load_file(), which causes the error, has mixed backward and forward slashes. Could be that the problem?

alcohol:
If the entity loader is disabled on a system, loading files, even from the local file system, is considered as external to the running php process, and thus not allowed by the libxml extension. Reading the file contents and loading the xml as a string is a valid alternative however.

Fixes #3942

### DIFF
--- a/src/Composer/Downloader/PearPackageExtractor.php
+++ b/src/Composer/Downloader/PearPackageExtractor.php
@@ -137,7 +137,7 @@ class PearPackageExtractor
     private function buildCopyActions($source, array $roles, $vars)
     {
         /** @var $package \SimpleXmlElement */
-        $package = simplexml_load_file($this->combine($source, 'package.xml'));
+        $package = simplexml_load_string(file_get_contents($this->combine($source, 'package.xml')));
         if (false === $package) {
             throw new \RuntimeException('Package definition file is not valid.');
         }
--- a/tests/Composer/Test/Downloader/PearPackageExtractorTest.php
+++ b/tests/Composer/Test/Downloader/PearPackageExtractorTest.php
@@ -18,11 +18,15 @@ class PearPackageExtractorTest extends \PHPUnit_Framework_TestCase
 {
     public function testShouldExtractPackage_1_0()
     {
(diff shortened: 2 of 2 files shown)


stof:
:+1:

alcohol:
Modified the unit tests to simulate the scenario in which previously the `PearPackageExtractor` would fail. Tests failed with the fix reverted, pass with the fix.

PR #4105, opened 2015-06-03
Remove extra '/' when findShortestPath $from is a directory

### LINKED ISSUE #3024: Composer adding extra '../' in classmap based on $baseDir
In my production environment, when I deploy my app and run `composer install --no-dev`, every Class mapped in `$baseDir` has an extra '../' in path.

```php
<?php

return array(
    'AlterFaltasEcode' => $baseDir . '/../app/database/migrations/2014_05_27_171840_alter_faltas_ecode.php',
    'AlterFaltasOrigem' => $baseDir . '/../app/database/migrations/2014_05_27_171906_alter_faltas_origem.php',
    'BaseController' => $baseDir . '/../app/controllers/BaseController.php',
    'Carbon\\Carbon' => $vendorDir . '/nesbot/carbon/src/Carbon/Carbon.php',
    'ClassPreloader\\Application' => $vendorDir . '/classpreloader/classpreloader/src/ClassPreloader/Application.php',
    'ClassPreloader\\ClassList' => $vendorDir . '/classpreloader/classpreloader/src/ClassPreloader/ClassList.php',
    'ClassPreloader\\ClassLoader' => $vendorDir . '/classpreloader/classpreloader/src/ClassPreloader/ClassLoader.php',
'User' => $baseDir . '/app/models/User.php',
    'Whoops\\Exception\\ErrorException' => $vendorDir . '/filp/whoops/src/Whoops/Exception/ErrorException.php',
    'Whoops\\Exception\\Frame' => $vendorDir . '/filp/whoops/src/Whoops/Exception/Frame.php',
    'Whoops\\Exception\\FrameCollection' => $vendorDir . '/filp/whoops/src/Whoops/Exception/FrameCollection.php',
    'Whoops\\Exception\\Inspector' => $vendorDir . '/filp/whoops/src/Whoops/Exception/Inspector.php',
    'Whoops\\Handler\\CallbackHandler' => $vendorDir . '/filp/whoops/src/Whoops/Handler/CallbackHandler.php',
    'Whoops\\Handler\\Handler' => $vendorDir . '/filp/whoops/src/Whoops/Handler/Handler.php',
    'Whoops\\Handler\\HandlerInterface' => $vendorDir . '/filp/whoops/src/Whoops/Handler/HandlerInterface.php',
    'Whoops\\Handler\\JsonResponseHandler' => $vendorDir . '/filp/whoops/src/Whoops/Handler/JsonResponseHandler.php',
    'Whoops\\Handler\\PrettyPageHandler' => $vendorDir . '/filp/whoops/src/Whoops/Handler/PrettyPageHandler.php',
    'Whoops\\Handler\\XmlResponseHandler' => $vendorDir . '/filp/whoops/src/Whoops/Handler/XmlResponseHandler.php',
    'Whoops\\Module' => $vendorDir . '/filp/whoops/src/Whoops/Provider/Zend/Module.php',
    'Whoops\\Provider\\Phalcon\\WhoopsServiceProvider' => $vendorDir . '/filp/whoops/src/Whoops/Provider/Phalcon/WhoopsServiceProvider.php',
    'Whoops\\Provider\\Silex\\WhoopsServiceProvider' => $vendorDir . '/filp/whoops/src/Whoops/Provider/Silex/WhoopsServiceProvider.php',
    'Whoops\\Provider\\Zend\\ExceptionStrategy' => $vendorDir . '/filp/whoops/src/Whoops/Provider/Zend/ExceptionStrategy.php',
    'Whoops\\Provider\\Zend\\RouteNotFoundStrategy' => $vendorDir . '/filp/whoops/src/Whoops/Provider/Zend/RouteNotFoundStrategy.php',
    'Whoops\\Run' => $vendorDir . '/filp/whoops/src/Whoops/Run.php',
);
```
In dev environment the generated autoload_classmap,php is working fine.
Removing  '../' in the production generated file fix the problem, but this is not intend to happen. 

I'm using Laravel 4.1 in my application, lastest version of composer, php 5.3, apache 2.2 in a Windows server (Xampp 1.7).

My composer.json (ommited some parts):
```json
{
	"require": {
		"laravel/framework": "4.1.*",
		"rhumsaa/array_column": "1.*",
		"patricktalmadge/bootstrapper": "dev-develop@dev"
	},
	"require-dev": {
		"barryvdh/laravel-ide-helper": "1.*",
		"barryvdh/laravel-debugbar" : "1.*",
		"orangehill/iseed": "dev-master@dev",
		"way/generators": "1.*",
		"barryvdh/laravel-migration-generator": "dev-master"
	},
	"autoload": {
		"classmap": [
			"/app/commands",
			"/app/controllers",
			"/app/models",
			"/app/database/migrations",
			"/app/database/seeds",
			"/app/tests/TestCase.php"
		],
		"psr-0": {
			"Pcm\\Indicators\\": "/src/"
		}
	}
}
```

gmsantosxl:
Resolves #3024

Passing the third argument as `true` on `findShortestPath ` method, a dummy file is added to `$from`. 
If `$from` ends with `/` (e.g. C:/ , /foo/bar/ ), a second `/` is inserted resulting on a invalid path.

### DIFF
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -316,7 +316,7 @@ class Filesystem
         $to = lcfirst($this->normalizePath($to));
 
         if ($directories) {
-            $from .= $this->isDirectory($from) ? 'dummy_file' : '/dummy_file';
+            $from = rtrim($from, '/') . '/dummy_file';
         }
 
         if (dirname($from) === dirname($to)) {
@@ -459,17 +459,6 @@ class Filesystem
     public static function isLocalPath($path)
     {
         return (bool) preg_match('{^(file://|/|[a-z]:[\\\\/]|\.\.[\\\\/]|[a-z0-9_.-]+[\\\\/])}i', $path);
-    }
-    
-    /**
-     * Return if the given path is a directory
-     *
-     * @param  string $path
-     * @return bool
-     */
-    public static function isDirectory($path)
-    {
-        return substr($path, -1) === '/';
     }
 
     public static function getPlatformPath($path)
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -316,7 +316,7 @@ class Filesystem
         $to = lcfirst($this->normalizePath($to));
 
         if ($directories) {
-            $from .= '/dummy_file';
+            $from .= $this->isDirectory($from) ? 'dummy_file' : '/dummy_file';
         }
 
         if (dirname($from) === dirname($to)) {
@@ -459,6 +459,17 @@ class Filesystem
     public static function isLocalPath($path)
     {
         return (bool) preg_match('{^(file://|/|[a-z]:[\\\\/]|\.\.[\\\\/]|[a-z0-9_.-]+[\\\\/])}i', $path);
+    }
+    
+    /**
+     * Return if the given path is a directory
+     *
+     * @param  string $path
+     * @return bool
+     */
+    public static function isDirectory($path)
+    {
+        return substr($path, -1) === '/';
     }
 
     public static function getPlatformPath($path)
--- a/tests/Composer/Test/Util/FilesystemTest.php
+++ b/tests/Composer/Test/Util/FilesystemTest.php
@@ -112,6 +112,7 @@ class FilesystemTest extends TestCase
             array('/foo/bar_vendor', '/foo/bar', '../bar', true),
             array('/foo/bar_vendor', '/foo/bar/src', '../bar/src', true),
             array('/foo/bar_vendor/src2', '/foo/bar/src/lib', '../../bar/src/lib', true),
+            array('C:/', 'C:/foo/bar/', "foo/bar", true),
         );
     }
 


alcohol:
Please rename the function as it does not relate to the context. To check if you're dealing with a directory you would use http://php.net/is_dir. You are simply interested in knowing if the string ends with a `/` or not. That doesn't have anything to do with it actually being a directory.

The problem here is that after the `normalizePath()` call (which strips trailing `/` characters), root paths remain `/` (on *nix) and `<DRIVE>:/` (on windows). This then turns `$from` into `//dummy_file` on *nix and into `c://dummy_file` on windows. The function then correctly determines the path on *nix, but fails on windows. So the problem lies elsewhere in my humble opinion. Most likely due to on windows there being a prefix before the root `/`.

src/Composer/Util/Filesystem.php
```diff
@@ -460,6 +460,17 @@ public static function isLocalPath($path)
...
         return (bool) preg_match('{^(file://|/|[a-z]:[\\\\/]|\.\.[\\\\/]|[a-z0-9_.-]+[\\\\/])}i', $path);
     }
+    
+    /**
+     * Return if the given path is a directory
+     *
+     * @param  string $path
+     * @return bool
+     */
+    public static function isDirectory($path)
```
stof: why it is public ?

alcohol:
On *nix the `$commonPath` ends up being `/` which then results in simply returning `$to` as shortest path. On Windows the `$commonPath` value is `c:/` which does not equal `/` so it continues processing and ends up with the leading `../`.

src/Composer/Util/Filesystem.php
```diff
@@ -316,7 +316,7 @@ public function findShortestPath($from, $to, $directories = false)
         $to = lcfirst($this->normalizePath($to));
 
         if ($directories) {
-            $from .= '/dummy_file';
+            $from .= $this->isDirectory($from) ? 'dummy_file' : '/dummy_file';
```
alcohol: I would personally just change this line to the following, no need for intermediate function

``` php
$from = rtrim($from, '/') . '/dummy_file';
```

src/Composer/Util/Filesystem.php
Seldaek: Yeah that should definitely be private or made to work with \ paths as well, because as is it depends on the path being normalized first.

src/Composer/Util/Filesystem.php
alcohol: At this point `$from` is normalized so you can't go wrong with this solution pretty much.

gmsantosxl: Indeed, `rtrim($from, '/')` is enough

Seldaek:
Thanks

PR #4114, opened 2015-06-06
fix typo + better express order of events

nevvermind:
The fix here being more "occurs before a package has been uninstalled", which is incorrect.


alcohol:
Maybe phrase the pre- events as 'will be' and the post events as 'has been'.

Seldaek:
IMO will be sounds a bit strange (I tried but.. had to revert :P) but I updated other post-* to has been.

PR #4118, opened 2015-06-08
Make an invalid package name a publish error, not warning

legoktm:
Since invalid names will prevent publishing to packagist, they should
be considered publish errors. If people do not plan on submitting their
package to packagist, they can use the --no-check-publish flag to turn
it into a normal warning again.

### DIFF
--- a/src/Composer/Util/ConfigValidator.php
+++ b/src/Composer/Util/ConfigValidator.php
@@ -102,7 +102,7 @@ class ConfigValidator
             $suggestName = preg_replace('{(?:([a-z])([A-Z])|([A-Z])([A-Z][a-z]))}', '\\1\\3-\\2\\4', $manifest['name']);
             $suggestName = strtolower($suggestName);
 
-            $warnings[] = sprintf(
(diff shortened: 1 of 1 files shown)


alcohol:
Packagist already fails on warnings, so there is no need for this.

alcohol:
The warning also explicitly states they will not be able to publish to packagist.

legoktm:
My usecase here is to have CI fail when someone tries creating a package with an uppercase name in it. We currently use `composer validate` to do that. But since it's just a warning, it doesn't return a failure exit code, causing CI to pass.

This is kind of a follow up to #3804, which officially differentiated between validation errors and publish errors, and let people ignore publish ones.

alcohol:
Upon further review I agree, in the context of the validate command this should be a publish error. See also [ValidateCommand.php#L82-L86](https://github.com/composer/composer/blob/4f80e7ff68466cab970c22b425725b8058c32970/src/Composer/Command/ValidateCommand.php#L82-L86).

Seldaek:
Makes sense yes thanks

PR #4122, opened 2015-06-08
Revert the plugin versions converting from "1.0.0" to "^1.0" feature

nevvermind:
This PR relates to https://github.com/composer/composer/pull/4089, which adds an unnecessary feature - converting a plugin's constraint from `1.0.0` to `^1.0`. That was a mistake on my part. I wrongly assumed that Composer has no support for multiple Plugin versions. Poor knowledge of the VersionParser class and some misinterpreted errors lead me to that conclusion.

The tests were kept, but changed accordingly. The exclusion of plugins which didn't match was kept. The text mentioning that a plugin was not registered was kept.

Thanks, @stof, for letting me know, albeit in a very obscure and then-hidden line comment. I saw it just now.

### DIFF
--- a/doc/articles/plugins.md
+++ b/doc/articles/plugins.md
@@ -32,13 +32,7 @@ requirements:
    to define which Plugin API versions your plugin is compatible with.
 
 The required version of the `composer-plugin-api` follows the same [rules][7]
-as a normal package's, except for the `1.0`, `1.0.0` and `1.0.0.0` _exact_ 
-values. In only these three cases, Composer will assume your plugin 
-actually meant `^1.0` instead. This was introduced to maintain BC with 
-the old style of declaring the Plugin API version.  
-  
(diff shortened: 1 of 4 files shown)


Seldaek:
Yeah this felt weird, I guess the docs were the main problem, because plugins requiring 1.0.0 exactly will break once we bump the version in composer, but they should be fixed anyway.. Maybe I should check in the packagist db which are doing that.

nevvermind:
Yes, 5% of the reason I misinterpreted things was that I've never seen a Composer plugin requiring anything else than `"composer-plugin-api": "1.0.0"`. And that's, in part, about docs.

Seldaek:
Ok I sent a few PRs to people that had mildly broken requires, and then emailed the 150+ plugin authors that had `1.0.0` and `*` as requirement. Case closed I hope.

Issue #4120, opened 2015-06-08
Script loop not detected


theofidry:
Given a script in the `composer.json` which refer to itself, like the following:

```json
"scripts": {
    "loop": "composer loop"
}
```

if I were to run `composer loop`, nothing would happen, just the program running and running until it blows the memory. Of course it is not a common use case and should be avoided since it's an error. Still it feels wrong not to have an error thrown in this case.

alcohol:
Why? If a user feels the need to do this, that is their prerogative.

:-1:

theofidry:
What do you mean? If a user do that that's obviously a mistake. My point is that I find it wrong to launch a script, which because used to take times you don't follow the progress and switch to something else, and have to wait for you computer to freeze to see that something is wrong.

alcohol:
Stupid is as stupid does. No need to add sanity checks to everything that could possibly go wrong. If people want to fork bomb themselves, they should be free to do so.

nevvermind:
@theofidry - It would've been worth the fix if the fix were that easy. The problem though is that I don't think it's that easy to detect that a loop is happening. And being somewhat of an edge case as you said, maybe it's just better to let this go as a quirk and maybe document it in the plugins section.

theofidry:
It would be useless to try to prevent every mistake one may do, but I can't help but feel uneasy knowing that there is absolutely no control on a circular reference, be it on the maximum number of iteration or memory limit. But that's just my opinion and its up to you to decide if it's worth to do a sanity check for that. It's just to me it's the same kind of check as to check if the script does exitst (currently if the script does not exist an error is thrown).

alcohol:
Most languages allow you to implement a fork bomb, even though it might be harmful. While we aim to make composer convenient and easy to use, we're not going to make it completely stupid-proof. For starters, no matter how stupid-proof we will make it, there will always be people that are creative enough to find a way to make things go **BOOM** anyway.

theofidry:
I agree that it making it stupid-proof may be too unproductive for what it's worth. But there's still something wrong in this case. If you expect your script to run fast and see that it takes more than 10sec instead of usually 1 ou 2, you know it ok. But that's not always the case.

I've check on `npm` out of curiosity, without much surprise they don't handle this case too. However the difference is here: they print the scirpt name so you see right away that something is wrong:

```
$ npm run loop

> myPackage@0.0.1 loop /path/to/mypackage
> npm run loop


> myPackage@0.0.1 loop /path/to/mypackage
[...]

PR #4123, opened 2015-06-09
Output script command in verbose mode

alcohol:
Follow up on #4120

Not sure if we should only show it in verbose mode, or show by default and hide in quiet mode. Feedback/thoughts?

/cc @Seldaek @theofidry

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -197,6 +197,8 @@ class EventDispatcher
     {
         $event = $this->checkListenerExpectedEvent(array($className, $methodName), $event);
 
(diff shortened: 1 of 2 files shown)


theofidry:
IMO it should be shown even in quiet mode. It is not too verbose to begin with and otherwise it would be meaningless for the use case described in #4120.

stof:
what about the case of PHP scripts ?

alcohol:
What exactly do you mean? Like..

```
"scripts": {
  "command": "Some\\Class::method"
}
```

?

Seldaek:
Thanks, I added the event name in verbose mode too, I think this might become a pain for people with plugins having lots of listeners though so we might need to downgrade it to -v/-vv instead of */-v but let's see how it goes.

PR #4124, opened 2015-06-09
Added Capable plugins for a more future-proof Plugin API

nevvermind:
Plugins can now present their capabilities to the PluginManager, through which it can act accordingly, thus making Plugin API more flexible, BC-friendly and decoupled.

**The Plugin API was bumped to version `1.1.0`.** :boom: 

**a) Plugin API evolution**
Every time Composer decides that a new extension should be added, it will extend its Plugin API to support that extension point - through an interface, an abstract class or a concrete one.  
Some example of extension points would be: allowing users to add their own Composer commands, allowing users to add their own validation steps in the `validate` command, offloading certain operations to userland implementations, otherwise falling back to default etc.

Because the PluginManager and the plugins communicate via strings and arrays, the possibility of autoloading non-existent classes in case of a version mis-match between them is avoided - thus a BC break.

**b) Plugin evolution**
The plugin's job is simple: if Composer provides an API and the plugin wants to implement it, it is free to do so, provided it declares this in its capabilities method, which tells the PluginManager it can handle that particular API. 

Because plugins depend on the `Capable` interface to declare what they support, they *must* require `composer-plugin-api` versions starting from `1.1.0`, so that newer plugins will not be loaded in an older version of Composer, which doesn't have the `Capable` interface.

Examples of API capabilities:
```php
namespace Composer\Plugin\Capability;

interface CommandProvider extends Capability
{
    public function getCommands();
}

interface AdditionalValidator extends Capability
{
    public function validate();
}
```
A plugin implementation could look like this:
```php
namespace My;

use Composer\Plugin\PluginInterface;
use Composer\Plugin\Capable;
use Composer\Composer;
use Composer\IO\IOInterface;

class Plugin implements PluginInterface, Capable
{
    public function activate(Composer $composer, IOInterface $io)
    {}

    public function getCapabilities()
    {
        return array(
            'Composer\Plugin\Capability\CommandProvider' => 'My\CommandProviderImpl',
            'Composer\Plugin\Capability\AdditionalValidator' => 'My\ValidatorImpl',
        );
    }   
}
```
Note the additional `Capable` interface that can be implemented by the plugin - only if the plugin wants to provide implementations. It can very well only implement the `PluginInterface`. `Capable` and `PluginInterface` can merge in some other major version of Composer.

The PluginManager handles the instantiation and setup of the implementations:
```php
if ($commandProvider = $pm->getPluginCapability($plugin, 'Composer\Plugin\Capability\CommandProvider')) {
    var_dump($commandProvider->getCommands());
}
```
Values can be passed to implementation via a constructor-injected array. This will allow us to change what we pass to plugins, without having to change the method signature of the APIs, thus making _minor_ version bumps, instead of _major_:
```php
// args in Plugin API 1.1
$args = array('io' => $io, 'next' => $nextValidator);

// args in Plugin API 1.2
$args = array('io' => $io, 'next' => $nextValidator, 'previous' => $previousValidator);

$pm->getPluginCapability($plugin, 'Composer\Plugin\Capability\AdditionalValidator', $args);
```
PR related to https://github.com/composer/composer/pull/3377.

### DIFF
--- a/src/Composer/Plugin/Capability/Capability.php
+++ b/src/Composer/Plugin/Capability/Capability.php
@@ -17,7 +17,6 @@ namespace Composer\Plugin\Capability;
  * Every new Capability which is added to the Plugin API must implement this interface.
  *
  * @api
- * @since Plugin API 1.1
  */
 interface Capability
 {
--- a/src/Composer/Plugin/Capability/Capability.php
+++ b/src/Composer/Plugin/Capability/Capability.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Plugin\Capability;
+
+/**
+ * Marker interface for Plugin capabilities.
+ * Every new Capability which is added to the Plugin API must implement this interface.
+ *
+ * @api
+ * @since Plugin API 1.1
+ */
+interface Capability
+{
+}
--- a/src/Composer/Plugin/Capable.php
+++ b/src/Composer/Plugin/Capable.php
@@ -28,9 +28,7 @@ interface Capable
      * The key must be a string, representing a fully qualified class/interface name
      * which Composer Plugin API exposes.
      * The value must be a string as well, representing the fully qualified class name
-     * of the API class.
-     *
-     * Every implementation will be passed a single array parameter via their constructor.
+     * of the implementing class.
      *
      * @tutorial
      *
--- a/src/Composer/Plugin/Capable.php
+++ b/src/Composer/Plugin/Capable.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Plugin;
+
+/**
+ * Plugins which need to expose various implementations
+ * of the Composer Plugin Capabilities must have their
+ * declared Plugin class implementing this interface.
+ *
+ * @api
+ * @since Plugin API 1.1
+ */
+interface Capable
+{
+    /**
+     * Method by which a Plugin announces its API implementations, through an array
+     * with a special structure.
+     *
+     * The key must be a string, representing a fully qualified class/interface name
+     * which Composer Plugin API exposes - named "API class".
+     * The value must be a string as well, representing the fully qualified class name
+     * of the API class - named "SPI class".
+     *
+     * Every SPI must implement their API class.
+     *
+     * Every SPI will be passed a single array parameter via their constructor.
+     *
+     * Example:
+     * // API as key, SPI as value
+     * return array(
+     *      'Composer\Plugin\Capability\CommandProvider' => 'My\CommandProvider',
+     *      'Composer\Plugin\Capability\Validator'       => 'My\Validator',
+     * );
+     *
+     * @return string[]
+     */
+    public function getCapabilities();
+}
--- a/src/Composer/Plugin/Capable.php
+++ b/src/Composer/Plugin/Capable.php
@@ -18,7 +18,6 @@ namespace Composer\Plugin;
  * declared Plugin class implementing this interface.
  *
  * @api
- * @since Plugin API 1.1
  */
 interface Capable
 {
--- a/src/Composer/Plugin/Capable.php
+++ b/src/Composer/Plugin/Capable.php
@@ -26,19 +26,17 @@ interface Capable
      * with a special structure.
      *
      * The key must be a string, representing a fully qualified class/interface name
-     * which Composer Plugin API exposes - named "API class".
+     * which Composer Plugin API exposes.
      * The value must be a string as well, representing the fully qualified class name
-     * of the API class - named "SPI class".
+     * of the API class.
      *
-     * Every SPI must implement their API class.
+     * Every implementation will be passed a single array parameter via their constructor.
      *
-     * Every SPI will be passed a single array parameter via their constructor.
+     * @tutorial
      *
-     * Example:
-     * // API as key, SPI as value
      * return array(
-     *      'Composer\Plugin\Capability\CommandProvider' => 'My\CommandProvider',
-     *      'Composer\Plugin\Capability\Validator'       => 'My\Validator',
+     *     'Composer\Plugin\Capability\CommandProvider' => 'My\CommandProvider',
+     *     'Composer\Plugin\Capability\Validator'       => 'My\Validator',
      * );
      *
      * @return string[]
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -23,6 +23,7 @@ use Composer\Package\PackageInterface;
 use Composer\Package\Link;
 use Composer\Semver\Constraint\Constraint;
 use Composer\DependencyResolver\Pool;
+use Composer\Plugin\Capability\Capability;
 
 /**
  * Plugin manager
@@ -186,16 +187,6 @@ class PluginManager
     }
 
     /**
-     * Returns the version of the internal composer-plugin-api package.
-     *
-     * @return string
-     */
-    protected function getPluginApiVersion()
-    {
-        return PluginInterface::PLUGIN_API_VERSION;
-    }
-
-    /**
      * Adds a plugin, activates it and registers it with the event dispatcher
      *
      * @param PluginInterface $plugin plugin instance
@@ -298,5 +289,58 @@ class PluginManager
         }
 
         return $this->globalComposer->getInstallationManager()->getInstallPath($package);
+    }
+
+    /**
+     * Returns the version of the internal composer-plugin-api package.
+     *
+     * @return string
+     */
+    protected function getPluginApiVersion()
+    {
+        return PluginInterface::PLUGIN_API_VERSION;
+    }
+
+    /**
+     * @param PluginInterface $plugin
+     * @param string $capability
+     * @return bool|string The fully qualified class of the implementation or false if none was provided
+     */
+    protected function getCapabilityImplementationClassName(PluginInterface $plugin, $capability)
+    {
+        if (!($plugin instanceof Capable)) {
+            return false;
+        }
+
+        $capabilities = (array) $plugin->getCapabilities();
+
+        if (empty($capabilities[$capability]) || !is_string($capabilities[$capability])) {
+            return false;
+        }
+
+        return trim($capabilities[$capability]);
+    }
+
+    /**
+     * @param PluginInterface $plugin
+     * @param string          $capability The fully qualified name of the API interface which the plugin may provide
+     *                                    an implementation.
+     * @param array           $ctorArgs   Arguments passed to Capability's constructor.
+     *                                    Keeping it an array will allow future values to be passed w\o changing the signature.
+     * @return Capability|boolean         Bool false if the Plugin has no implementation of the requested Capability.
+     */
+    public function getPluginCapability(PluginInterface $plugin, $capability, array $ctorArgs = array())
+    {
+        if ($capabilityClass = $this->getCapabilityImplementationClassName($plugin, $capability)) {
+            if (class_exists($capabilityClass)) {
+                $capabilityObj = new $capabilityClass($ctorArgs);
+                if ($capabilityObj instanceof Capability &&
+                    $capabilityObj instanceof $capability
+                ) {
+                    return $capabilityObj;
+                }
+            }
+        }
+        return false;
     }
 }
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -187,6 +187,16 @@ class PluginManager
     }
 
     /**
+     * Returns the version of the internal composer-plugin-api package.
+     *
+     * @return string
+     */
+    protected function getPluginApiVersion()
+    {
+        return PluginInterface::PLUGIN_API_VERSION;
+    }
+
+    /**
      * Adds a plugin, activates it and registers it with the event dispatcher
      *
      * @param PluginInterface $plugin plugin instance
@@ -292,55 +302,57 @@ class PluginManager
     }
 
     /**
-     * Returns the version of the internal composer-plugin-api package.
-     *
-     * @return string
-     */
-    protected function getPluginApiVersion()
-    {
-        return PluginInterface::PLUGIN_API_VERSION;
-    }
-
-    /**
      * @param PluginInterface $plugin
      * @param string $capability
-     * @return bool|string The fully qualified class of the implementation or false if none was provided
+     * @return null|string The fully qualified class of the implementation or null if Plugin is not of Capable type
+     * @throws \RuntimeException On empty or non-string implementation class name value
      */
     protected function getCapabilityImplementationClassName(PluginInterface $plugin, $capability)
     {
         if (!($plugin instanceof Capable)) {
-            return false;
+            return null;
         }
 
         $capabilities = (array) $plugin->getCapabilities();
 
-        if (empty($capabilities[$capability]) || !is_string($capabilities[$capability])) {
-            return false;
+        if (!empty($capabilities[$capability]) && is_string($capabilities[$capability])) {
+            $capabilities[$capability] = trim($capabilities[$capability]);
         }
 
-        return trim($capabilities[$capability]);
+        if (empty($capabilities[$capability]) || !is_string($capabilities[$capability])) {
+            throw new \RuntimeException('Plugin provided invalid capability class name(s)');
+        }
+
+        return $capabilities[$capability];
     }
 
     /**
      * @param PluginInterface $plugin
-     * @param string          $capability The fully qualified name of the API interface which the plugin may provide
+     * @param string $capabilityClassName The fully qualified name of the API interface which the plugin may provide
      *                                    an implementation.
-     * @param array           $ctorArgs   Arguments passed to Capability's constructor.
-     *                                    Keeping it an array will allow future values to be passed w\o changing the signature.
-     * @return Capability|boolean         Bool false if the Plugin has no implementation of the requested Capability.
+     * @param array $ctorArgs Arguments passed to Capability's constructor.
+     *                        Keeping it an array will allow future values to be passed w\o changing the signature.
+     * @return null|Capability
      */
-    public function getPluginCapability(PluginInterface $plugin, $capability, array $ctorArgs = array())
+    public function getPluginCapability(PluginInterface $plugin, $capabilityClassName, array $ctorArgs = array())
     {
-        if ($capabilityClass = $this->getCapabilityImplementationClassName($plugin, $capability)) {
+        if ($capabilityClass = $this->getCapabilityImplementationClassName($plugin, $capabilityClassName)) {
             if (class_exists($capabilityClass)) {
                 $capabilityObj = new $capabilityClass($ctorArgs);
                 if ($capabilityObj instanceof Capability &&
-                    $capabilityObj instanceof $capability
+                    $capabilityObj instanceof $capabilityClassName
                 ) {
                     return $capabilityObj;
+                } else {
+                    throw new \RuntimeException(
+                        'Class ' . $capabilityClass . ' must be of both \Composer\Plugin\Capability\Capability and '.
+                        $capabilityClassName . ' types.'
+                    );
                 }
+            } else {
+                throw new \RuntimeException("Cannot instantiate Capability, as class $capabilityClass does not exist.");
             }
         }
-        return false;
+        return null;
     }
 }
--- a/tests/Composer/Test/Plugin/Fixtures/plugin-v1/composer.json
+++ b/tests/Composer/Test/Plugin/Fixtures/plugin-v1/composer.json
@@ -7,6 +7,6 @@
         "class": "Installer\\Plugin"
     },
     "require": {
-        "composer-plugin-api": "1.0.0"
+        "composer-plugin-api": "*"
     }
 }
--- a/tests/Composer/Test/Plugin/Fixtures/plugin-v1/composer.json
+++ b/tests/Composer/Test/Plugin/Fixtures/plugin-v1/composer.json
@@ -7,6 +7,6 @@
         "class": "Installer\\Plugin"
     },
     "require": {
-        "composer-plugin-api": "*"
+        "composer-plugin-api": "^1.0"
     }
 }
--- a/tests/Composer/Test/Plugin/PluginInstallerTest.php
+++ b/tests/Composer/Test/Plugin/PluginInstallerTest.php
@@ -249,24 +249,6 @@ class PluginInstallerTest extends TestCase
         $this->pm->loadInstalledPlugins();
     }
 
-    public function testExactPluginVersionStyleAreRegisteredCorrectly()
-    {
-        $pluginsWithFixedAPIVersions = array(
-            $this->packages[0],
-            $this->packages[1],
-            $this->packages[2],
-        );
-
-        $this->setPluginApiVersionWithPlugins('1.0.0', $pluginsWithFixedAPIVersions);
-        $this->assertCount(3, $this->pm->getPlugins());
-
-        $this->setPluginApiVersionWithPlugins('1.0.1', $pluginsWithFixedAPIVersions);
-        $this->assertCount(0, $this->pm->getPlugins());
-
-        $this->setPluginApiVersionWithPlugins('2.0.0-dev', $pluginsWithFixedAPIVersions);
-        $this->assertCount(0, $this->pm->getPlugins());
-    }
-
     public function testStarPluginVersionWorksWithAnyAPIVersion()
     {
         $starVersionPlugin = array($this->packages[4]);
--- a/tests/Composer/Test/Plugin/PluginInstallerTest.php
+++ b/tests/Composer/Test/Plugin/PluginInstallerTest.php
@@ -314,4 +314,67 @@ class PluginInstallerTest extends TestCase
         $this->setPluginApiVersionWithPlugins('5.5.0', $pluginWithApiConstraint);
         $this->assertCount(0, $this->pm->getPlugins());
     }
+
+    public function testIncapablePluginIsCorrectlyDetected()
+    {
+        $plugin = $this->getMockBuilder('Composer\Plugin\PluginInterface')
+                       ->getMock();
+
+        $this->assertFalse($this->pm->getPluginCapability($plugin, 'Fake\Ability'));
+    }
+
+    public function testCapabilityImplementsComposerPluginApiClassAndIsConstructedWithArgs()
+    {
+        $capabilityApi = 'Composer\Plugin\Capability\Capability';
+        $capabilitySpi = 'Composer\Test\Plugin\Mock\Capability';
+
+        $plugin = $this->getMockBuilder('Composer\Test\Plugin\Mock\CapablePluginInterface')
+                       ->getMock();
+
+        $plugin->expects($this->once())
+               ->method('getCapabilities')
+               ->will($this->returnCallback(function() use ($capabilitySpi, $capabilityApi) {
+                   return array($capabilityApi => $capabilitySpi);
+               }));
+
+        $capability = $this->pm->getPluginCapability($plugin, $capabilityApi, array('a' => 1, 'b' => 2));
+
+        $this->assertInstanceOf($capabilityApi, $capability);
+        $this->assertInstanceOf($capabilitySpi, $capability);
+        $this->assertSame(array('a' => 1, 'b' => 2), $capability->args);
+    }
+
+    public function invalidSpiValues()
+    {
+        return array(
+            array(null),
+            array(""),
+            array(0),
+            array(1000),
+            array("   "),
+            array(array(1)),
+            array(array()),
+            array(new \stdClass()),
+            array("NonExistentClassLikeMiddleClass"),
+        );
+    }
+
+    /**
+     * @dataProvider invalidSpiValues
+     */
+    public function testInvalidCapabilitySpiDeclarationsAreDisregarded($invalidSpi)
+    {
+        $capabilityApi = 'Composer\Plugin\Capability\Capability';
+
+        $plugin = $this->getMockBuilder('Composer\Test\Plugin\Mock\CapablePluginInterface')
+                       ->getMock();
+
+        $plugin->expects($this->once())
+               ->method('getCapabilities')
+               ->will($this->returnCallback(function() use ($invalidSpi, $capabilityApi) {
+                   return array($capabilityApi => $invalidSpi);
+               }));
+
+        $this->assertFalse($this->pm->getPluginCapability($plugin, $capabilityApi));
+    }
 }
--- a/tests/Composer/Test/Plugin/PluginInstallerTest.php
+++ b/tests/Composer/Test/Plugin/PluginInstallerTest.php
@@ -147,7 +147,7 @@ class PluginInstallerTest extends TestCase
         $this->repository
             ->expects($this->exactly(2))
             ->method('getPackages')
-            ->will($this->returnValue(array()));
+            ->will($this->returnValue(array($this->packages[3])));
         $installer = new PluginInstaller($this->io, $this->composer);
         $this->pm->loadInstalledPlugins();
 
--- a/tests/Composer/Test/Plugin/PluginInstallerTest.php
+++ b/tests/Composer/Test/Plugin/PluginInstallerTest.php
@@ -302,31 +302,31 @@ class PluginInstallerTest extends TestCase
         $plugin = $this->getMockBuilder('Composer\Plugin\PluginInterface')
                        ->getMock();
 
-        $this->assertFalse($this->pm->getPluginCapability($plugin, 'Fake\Ability'));
+        $this->assertNull($this->pm->getPluginCapability($plugin, 'Fake\Ability'));
     }
 
     public function testCapabilityImplementsComposerPluginApiClassAndIsConstructedWithArgs()
     {
         $capabilityApi = 'Composer\Plugin\Capability\Capability';
-        $capabilitySpi = 'Composer\Test\Plugin\Mock\Capability';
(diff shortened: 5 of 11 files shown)


slbmeh:
I think this kind of sidesteps the underlying problem. What happens with changes to the capabilities?

An adapter pattern might be better suited here to allow for multiple implementation versions to live in the same ecosystem.

I think if you have a baseline `PluginInterface` that gets the capabilities and the version, you can then pass the plugin to a versioned capability handler.

Suppose you have a plugin that implements the installer capability, down the road the concept of installers may for whatever reason be shifted to use the mediator pattern between the repositories and downloaders... The capability API would be able to differentiate a version 1 installer from a version 2 installer and pass it off to different installation managers.

nevvermind:
> What happens with changes to the capabilities?

That's actually a Plugin API change. Just like with any other package, and depending on the situation, you bump its major or minor versions.

> to allow for multiple implementation versions to live in the same ecosystem.

I could imagine scenarios in which that might be helpful, but, overall, and don't think it's worth having multiple implementations. The backward-compatibility is actually granted by the Plugin API version. If Composer acts nicely, it can remain in the `1.*` version until the end of days, but, if it does change the capability/implementation in a major way, that would mean Plugin API `2.0`. Why would it keep `1.0`-compatible API in its `2.0` version?

> I think if you have a baseline `PluginInterface` that gets the capabilities and the version

If you suggest to change the `PluginInterface`, that would push the Plugin API to 2.0 instantly.

> you can then pass the plugin to a versioned capability handler.

I think this is doable, sure (although I don't know exactly _how_), but then you'll have the Plugin API version **and** the Capabilities'. Wasn't the first covering the latter? What's the purpose of the Plugin API then?

nevvermind:
I can definitely think of component-specific (capability-specific) versioning (netbeans, phpstorm, most of the complex apps basically), but that's more suited if those were in their own package, not in the same codebase.

nevvermind:
This is ready to be reviewed, btw.

slbmeh:
> capability/implementation in a major way, that would mean Plugin API 2.0. Why would it keep 1.0-compatible API in its 2.0 version?

If you have a backwards compatibility breakage on one capability that forces 2.0, but the other capabilities are still compatible with 1.*, otherwise I don't see the gain in splitting out capabilities.

If a plugin requires `^1.7` because the status capability was added to that version, you should be able to assume that 1.8 didn't remove the status capability otherwise that would be 2.0.  On the other hand, if the validator capability caused the bump to 2.0 because you require a different interface now, then you are forcing every plugin author regardless of whether or not they use the validator capability to create a new release because the constraint is vaguely on the plugin implementation and not the capabilities.

src/Composer/Plugin/Capable.php
```diff
@@ -0,0 +1,48 @@
...
+     * of the API class - named "SPI class".
+     *
+     * Every SPI must implement their API class.
+     *
+     * Every SPI will be passed a single array parameter via their constructor.
+     *
+     * Example:
+     * // API as key, SPI as value
+     * return array(
+     *      'Composer\Plugin\Capability\CommandProvider' => 'My\CommandProvider',
```
slbmeh: Wouldn't simply implementing the corresponding interface suffice for determining what type it is?

src/Composer/Plugin/PluginManager.php
```diff
@@ -287,4 +288,47 @@ protected function getPluginApiVersion()
...
+
+    /**
+     * @param PluginInterface $plugin
+     * @param string          $capability The fully qualified name of the API interface which the plugin may provide
+     *                                    an implementation.
+     * @param array           $ctorArgs   Arguments passed to Capability's constructor.
+     *                                    Keeping it an array will allow future values to be passed w\o changing the signature.
+     * @return Capability|boolean         Bool false if the Plugin has no implementation of the requested Capability.
+     */
+    public function getPluginCapability(PluginInterface $plugin, $capability, array $ctorArgs = array())
```
slbmeh: I get the concept of the array here for constructor args, IMO it's a bit messy.  What about an implementation where there is no constructor?

Each interface can be responsible for defining the setters the implementation will utilize making the logic easier to follow.

src/Composer/Plugin/Capable.php
nevvermind: I don't understand exactly what you're saying. It's not about type resolution, it's about getting an implementation class.

src/Composer/Plugin/PluginManager.php
nevvermind: > Each interface can be responsible for defining the setters the implementation will utilize making the logic easier to follow.

What about when we want to inject something else? Wouldn't a new interface setter create a BC-break?
And why messy? Is it because we're using hose old-fashioned arrays? :stuck_out_tongue:

nevvermind:
Yeah, I get what you're saying. What do you suggest in this case?

src/Composer/Plugin/PluginManager.php
slbmeh: If you add a method to the interface it would cause a BC break, but if you add an interface like `InstallationManagerAware` then it can set the installation manager to those implementations that need it.

Just my opinion about the array. I feel an array should represent a set. It's definitely valid here. I tend to prefer having a signature and the ability to type hint.

src/Composer/Plugin/Capable.php
slbmeh: It might be useful to have a lookup by command interface with the key value pairs.

My initial thought was you could achieve the same thing with just listing your implementations. Some capabilities may be related and make sense to implement in the same class.
If your capabilities implement interfaces, you should be able to use introspection or reflection to determine what command it is implementing.
```php
['My\CommandProvider','My\Validator']
```

slbmeh:
That scenario was what I was thinking about solving with the adapter pattern.

Your `getCapabilities()` could return version information.

```php
[
    'My\CommandProvider' => '1.1',
    'My\Validator' => '2.0'
]
````

And the API implementation might take those versions into consideration and proxy to different implementation versions if it were desirable to maintain BC for older versions of popular used capabilities such as the case of installers.  Or, simply reject the plugin capability with a warning because the version is unsupported.

File: src/Composer/Plugin/PluginManager.php (hunk 2 above)
Comment (nevvermind) [inline src/Composer/Plugin/PluginManager.php]: > I tend to prefer having a signature and the ability to type hint.

Who doesn't?! I dislike having to use arrays as well, but the problem with the whole interface-with-setters paradigm is that you end up with public setters. Then it's bye-bye immutability. Whereas if you throw an array in the constructor, the plugin is free to be as mutable or immutable as it wants to be.

Immutability, for me, is slightly more important than type hinting. Especially when we're talking 3rd parties.

I would still love those type hints, though.

File: src/Composer/Plugin/Capable.php (hunk 1 above)
Comment (nevvermind) [inline src/Composer/Plugin/Capable.php]: > Some capabilities may be related and make sense to implement in the same class.

This PR supports that: just declare the same impl for two different interfaces.

> If your capabilities implement interfaces, you should be able to use introspection or reflection to determine what command it is implementing.

That defeats the purpose of having interfaces, don't you think?

The reason I'm using strings and not `instanceof` comparisons is that I don't want to wake up that monster called autoloader, so that, in case the plugin is installed on an _older_ version of Composer, it wouldn't trigger a fatal "no such class exists". And why would I use reflection when I can use keys?

You must give a little more credit to arrays.

PS: We're splitting our discussion here. Continue in the main thread I you will.

**Comment by nevvermind** [comment]
I don't like it. It's versions all the way down! :stuck_out_tongue:  
Again, that wouldn't be a problem if those interfaces were in separate packages.

But, if we are to continue this path, at least let's assume something like this (the keys were explained in a previous line comment):
```php
[
    'Composer\Plugin\Capability\CommandProvider' => ['My\CommandProvider', '1.0'],
    'Composer\Plugin\Capability\Validator'       => ['My\Validator',       '2.0'],
]
```

slbmeh:
I apologize, I missed some of your questions.

> If you suggest changing the PluginInterface, that would push the Plugin API to 2.0 instantly.
> > you can then pass the plugin to a versioned capability handler.  

The `PluginInterface` doesn't necessarily have to change.  I meant to say Plugin API, you can add additional interfaces to the Plugin API without a BC break.  Just as you're doing in the PR.

The plugin interface itself could ideally remain at whatever version it is now.  And each additional feature extending the Plugin API such as capabilities would just be an additional interface on the plugin implementation.

> I think this is doable, sure (although I don't know exactly how), but then you'll have the Plugin API version and the Capabilities'. Wasn't the first covering the latter? What's the purpose of the Plugin API then?

If capabilities is done well, the Plugin API version shouldn't ever have to change unless the fundamentals of plugins change entirely.  It should just satisfy that there is a plugin, and initialize it.
The capabilities of a plugin are more likely to change, and may be desirable to have two implementations. See comment above for more.

nevvermind:
You've quite convinced me the advantage of keeping capability BC, instead on focusing on the Plugin API, so no need to push on that part.

What I don't like at this per-interface versioning is the amount of attention one needs to give to each one's version, instead of just the Plugin API's version.

I guess one is better doing `require "composer-capability-validator:1.1"` instead of `require "composer-plugin-api:~1.0"`.

src/Composer/Plugin/PluginManager.php
slbmeh: If the instantiated object is wholly encapsulated in a capability manager then immutability is irrelevant.  I suppose it all comes down to how capabilities are interacted with.

I don't understand the need for immutability however, if someone really wants to change it, then there's reflection.  I do it to parts of composer in my plugins all the time.  The only way to prevent that is to present clones whenever you allow something to access it rather than the actual object.

I think enforcing immutability is bad for interoperability. How do you know that someone doesn't need to change something with what they're trying to do. I think an attempt at immutability if the paranoia arises, should be on the capability author and not on composer. They can always implement a single use setter.

slbmeh:
I don't think a plugin author really needs to take care in the versioning of their capabilities. Simply choose the version that has the features you require. Whatever uses the capabilities is where the version needs to come in to consideration.

I would imagine most of the time capabilities are unlikely to have version changes if they're relatively small in scope. My other assumption would be that a good number of these major version bumps probably won't care about older versions. The capabilities I could imagine leveraging side by side support would be the more critical capabilities, and the functionality of those consumers would dictate the BC breakages, and should know how to deal with it, and when there may be a scenario for the subset of a capability that will work `1.1` to `9999.9999.9999.9999`

I don't like the require line because that makes it hard on the plugin author juggling multiple versions of the same plugin with variations in capability handling. Providing the version information in the capability declaration would allow for the same plugin to have multiple implementations of the same capability to support older versions of composer with their plugin.

src/Composer/Plugin/PluginManager.php
nevvermind: Immutability is bad for interoperability? Well now, I thought it's the _exact_ opposite. Isn't that like  saying that the interoperability between threads is made _worse_ by the fact that objects are immutable/locked. It's not about paranoia - it's about state consistency (which granted, can be achieved by your read-only clones). I'm guessing a couple of more coments on this thread, right? :smiley: 

I plan to make those plugins as dumb as they need to be. I actually dislike having an architecture like the current activate/execute simple calls. Ironically, I don't plan for make plugin _do_ stuff, so I don't need them to be wild and free. Quite the opposite: I plan to make Composer _ask_ them _very specific_ things. This will a) make plugin developers have an easy job creating plugins: just implement stuff like `getName`, `getVersion`, `getCommandName` and b) will make the Plugin API flexible: it can change the logic at any moment, because the plugins don't dictate it: they just provide information. Of course, if needed, Composer could offload entire business logic to them, but that's not future-proof.

nevvermind:
Regarding your previous snippet:
```php
public function getInstaller($type)
{
    foreach ($this->composer->getCapabilities('InstallerInterface', '2.3') as $cap) {
        if ($cap->getInstaller()->supports($type)) {
            return $cap->getInstaller();
        }
    }

    foreach ($this->composer->getCapabilities('InstallerInterface', '1.19') as $cap) {
        if ($cap->handles($type)) {
            return $cap;
        }
    }
}
```
I just _have_ to ask: how would Composer 5's code look like?

src/Composer/Plugin/PluginManager.php
slbmeh: It isn't about interoperability with composer, I know what composer is going to do with my plugin, and composer has a defined way of handing it.

The activate call is perfect.  I can do pretty much anything I need.

It's the third parties that cause the interoperability headaches.  I've lost track of how many times I've run in to scenarios where there is another plugin causing issues for one of my plugins. Very common with installers, for example I often need to iterate over all of the installers in the installation manager and wrap all of the others in a blocking decorator.

Easy is a plus, but I'd rather be able to do what i'm trying to do without needing to do a hack like replacing the root package with a stub so I can use my own solver.

nevvermind: > The activate call is perfect. I can do pretty much anything I need.

Ofc is perfect: it's giving you freedom. :+1:

> It's the third parties that cause the interoperability headaches. I've lost track of how many times I've run in to scenarios where there is another plugin causing issues for one of my plugins.

Ofc it's _their_ fault: they're given too much freedom.   :-1: 

You see what I mean now? None of this would've happen with proper use of dumb plugins and _Composer_ doing the work. Both your plugins and the others' know and are allowed too much. This, as you say, is lack interoperability and I agree.

slbmeh:
That was exactly my thought, a capability is a completely different change path that the plugin api.

slbmeh:
> I just have to ask: how would Composer 5's code look like? (Or is this what you're trying to say - that a Capability can be 2.3 indefinitely and regardless of the Plugin API/Composer version?)

That was what I was thinking, capabilities are part of the plugin subsystem, but they should be versioned independently in order to remain flexible.  It's reasonable to require a plugin author to update to support the latest version of the capability they implement, but not to require an update because a capability they don't care about has changed.

> The reason I'm using strings and not instanceof comparisons is that I don't want to wake up that monster called autoloader, so that, in case the plugin is installed on an older version of Composer, it wouldn't trigger a fatal "no such class exists". And why would I use reflection when I can use keys?

There are other means of introspection than `instanceof`, the functions that provide the support can optionally support strings as well.  The reflection statement might not be the best option, but if you wanted to have a constructor signature and instantiate from an argument array it's a likely scenario to utilize `ReflectionClass::newInstanceArgs()`.

As for the fatal, I can't really think of a scenario that would try to use a capability and not actually have the capability. Wouldn't the capability implementations be defined in the composer versions that leverage those capabilities? And for third parties using the capabilities, IMO it should fail regardless if it requires a capability that does not exist. If they want something that can work without a capability it should be on the plugin author to handle the sanity checks. Composer can't forsee what plugins are capable of handling. If it tries, it will introduce a lifecycle of continuous changes to the Plugin API and/or Capability APIs whenever a plugin author wants to do something new.

#### Thread TLDR: Summary of my Perspective
I write a lot of composer plugins (~130 currently) for different scenarios for companies I support. I'm completely behind the introduction of a capability API. I'm concerned that with a `dumb plugin` approach is going to introduce unnecessary maintenance and support either on the composer side or the plugin author's side, likely both.

Plugins are optional extensions and should be given the flexibility they need. Restricting what I can do is going to cause me to open feature requests and PRs to support my edge cases that composer shouldn't have to worry about.

I may have been a little too critical on this PR trying to identify potential use cases in the future. I think we need to take a step back and discuss how capabilities are going to be used before we can discuss which implementation is the correct way to do things.

nevvermind:
Can we get some more opinions up in here?

nevvermind:
> I'm concerned that with a dumb plugin approach is going to introduce unnecessary maintenance and support either on the composer side or the plugin author's side, likely both.

> Plugins are optional extensions and should be given the flexibility they need. Restricting what I can do is going to cause me to open feature requests and PRs to support my edge cases that composer shouldn't have to worry about.

The first quote is just speculation on your part.

You're confusing "dumb" with "inflexible". No one said that plugins couldn't be given callback freedom. There would be no point in having plugins if they weren't given the power to make their own decisions, now, wouldn't it? But sometimes, this exact freedom is what makes plugins not work with each other. I'm just saying that the current state of affairs grants _too much freedom_. 

You mentioned having a hard time making plugins to work with each other? The only reason that's the case is because of one argument: `Composer $composer`. And guess what - that's the _exact_ argument you love the most, right? Don't you see you're asking for the exact thing that makes plugin development hard in the first place?!

The way to fix this is by having less powerful plugins (not just a single `execute()` method), but flexible ones (several _simple, dumb and specialised_ callbacks or methods) and some effort on Composer's part when using said plugins.

And, yes - sometimes, Composer can just call `execute($composer)` on a plugin. If the scenario justifies it, why not?

slbmeh:
After the discussion on IRC, versioning simply by interfaces is probably the most straightforward route.

nevvermind:
@Seldaek - can I have your opinion on this?

Seldaek:
@nevvermind @slbmeh I'd like to resume discussion here, sorry for dropping the ball back then, but it probably should be tackled before wrapping up a 1.0.

After a quick read, I'd tend to agree with @nevvermind's approach, but I think having capability versions would probably be good. The question is how to describe them both in the sense of how the plugin describes versions (I kinda like https://github.com/composer/composer/pull/4124#issuecomment-111634963) and how we document which versions of which capabilities do what and so on. It could get pretty hairy but then again I don't expect having so many changes to those capabilities.

nevvermind:
Ok, so you decided on
```php
[
    'Composer\Plugin\Capability\CommandProvider' => ['My\CommandProvider', '1.0'],
    'Composer\Plugin\Capability\Validator'       => ['My\Validator',       '2.0'],
]
```
No problem. I think it's doable. Question though: what would `1.0` or `2.0` will represent in this case? _Exact_ versions? 
Would it be worth plugins declaring their desired Capability using versioning numbers?
```php
[
    'Composer\Plugin\Capability\CommandProvider' => ['My\CommandProvider', '^1.0'],
    'Composer\Plugin\Capability\Validator'       => ['My\Validator',       '~2.0'],
]
```
> "Composer: hey, what plugins are _compatible_ with this Capability's 2.0 version?" as opposed to "Composer: hey, what plugins want version 2.0 of this Capability?".

Thus plugins could relax a little bit (or not):
> "Plugin: I want everything from 1.1 to 2.0, it's ok, chillax" as opposed to "I only want 2.0, even though I'm compatible with 2.1 by definition, so yeah, nothing I can do about it. Composer plugins suck".

Would you care for such a feature, @Seldaek? I'm thinking this one would help Composer core people avoid doing lots of `if`s to cater for multiple API Capability versions.

Seldaek:
Yeah having requirements there would probably be good.. so if say we add an argument at the end of a method then we can bump that capability to 1.1 and it shouldn't break anything as there is no interface to implement and thus no signature to conflict with (right?).

The Semver stuff we have now probably helps to achieve this with very little code.

nevvermind:
It's more about Plugins being compatible with minor versions instead of declaring stuff like:
```php
[
'Composer\Plugin\Capability\CommandProvider' => ['My\CommandProvider', '1.0'], // OK
'Composer\Plugin\Capability\Validator' => ['My\Validator', '2.0'], // OK
'Composer\Plugin\Capability\Validator' => ['My\Validator', '2.1'], // <-- not cool
'Composer\Plugin\Capability\Validator' => ['My\Validator', '2.2'], // <-- not cool
'Composer\Plugin\Capability\Validator' => ['My\Validator3', '3.0'], // OK
]
```
They would, instead, do:
```php
[
'Composer\Plugin\Capability\CommandProvider' => ['My\CommandProvider', '1.0'], // OK
'Composer\Plugin\Capability\Validator' => ['My\Validator', '^2.0'], // OK
'Composer\Plugin\Capability\Validator' => ['My\Validator3', '3.0'], // OK
]
```
, thus leveraging what people are already used to when it comes to Composer. The core Capability then could change (minor or otherwise) at will.

What changes _core_ decides to do (be it another param, another interface altogether etc.), it's up to you/them, as long as you/they keep semantic meaning to your Capability changes.

Seldaek:
BTW.. your examples have multiple times the same key, that's not gonna work, but I am not sure if it's really a problem or not. Would a given plugin need to implement the same capability in two different versions?

Seldaek:
The question really.. is whether we need any of that versioning or not.

I am thinking it would be easier to just keep it to a simple list of capabilities, and say that if we break anything in a breaking way we will bump the composer-plugin-api major version.

Because that way at least a plugin that requires composer-plugin-api v1 won't be installed on a composer v2.. and that's the key point we need to achieve.

With the versioning scheme above, people would have to track down which composer plugin version works with their composer version, and do the resolution themselves. That kinda sucks.

So I think it might be best to keep using the initial proposal after all.

nevvermind:
Sure.

```php
[
    'Composer\Plugin\Capability\CommandProvider' => ['My\CommandProvider', '1.0'],
    'Composer\Plugin\Capability\Validator'       => ['My\Validator',       '2.0'],
]
```
it is.

Major bumps' fine, but what happens when `Composer\Plugin\Capability\Validator` is bumped to `2.1`. I think it's easier just to assume major-version compatibility on core's part and be done with it. So a core capability at version `2.1` asking for plugins, could get ones which declared `'Composer\Plugin\Capability\Validator' => ['My\Validator', '2.0']`, too.

I'll need some time to dust off this PR though, getting it up to date with the latest.

nevvermind:
Wait, you meant going back to 
```php
[
    'Composer\Plugin\Capability\CommandProvider' => 'My\CommandProvider',
    'Composer\Plugin\Capability\Validator'       => 'My\Validator',
]
```
?

Seldaek:
Yes I mean that. Where everything essentially has one version. Because that means the solver can do its job. If we pass in new parameters to the capabilities that's fine it's BC, but if we want to break anything we need to break the plugin api for everyone. And yes it sucks as most plugins wouldn't need an update unless they rely on something that broke, but it is a small pain for plugin author vs a lot of pain for all their users. It just means we want to avoid bumping the plugin API too often, and rather collect BC breaks into rare ones that require plugin authors to check their plugins..

I don't really see the other scheme with capabilities being versionned as workable in the long run.

nevvermind:
Yeah, well, now you understand why the `I don't like it. It's versions all the way down!`. Having both the Composer Plugin API versioned and each of its own Capabilities versioned seems like an step too complicated. But I'm trying to remember why @slbmeh wanted them.

Seldaek:
Yes.. well @slbmeh wants it because of the pain of updating all those plugins, and I get it, but I don't see a way around it. The proposed solution just doesn't cut it as far as I can tell.

src/Composer/Plugin/Capability/Capability.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+namespace Composer\Plugin\Capability;
+
+/**
+ * Marker interface for Plugin capabilities.
+ * Every new Capability which is added to the Plugin API must implement this interface.
+ *
+ * @api
+ * @since Plugin API 1.1
```
Seldaek: While you overhaul this would you mind dropping the @since stuff? I don't really see the point.. The code is what it is at a given version of the plugin api and that's that no?

tests/Composer/Test/Plugin/Fixtures/plugin-v1/composer.json
```diff
@@ -7,6 +7,6 @@
         "class": "Installer\\Plugin"
     },
     "require": {
-        "composer-plugin-api": "1.0.0"
+        "composer-plugin-api": "*"
```
Seldaek: Shouldn't these require plugin api ^1.0 at least? * sounds like it'll break eventually. I know it's just tests but still.

src/Composer/Plugin/Capability/Capability.php
nevvermind: Sure thing.

tests/Composer/Test/Plugin/Fixtures/plugin-v1/composer.json
nevvermind: Ok.

adragus-inviqa: I left one to be `*` as I've got a test named `testStarPluginVersionWorksWithAnyAPIVersion`.

nevvermind: Actually, I've got a test named `testStarPluginVersionWorksWithAnyAPIVersion` so I hope you don't mind my leaving one `*` in there.

Seldaek: No that's fine

nevvermind:
Updated.

src/Composer/Plugin/PluginManager.php
```diff
@@ -299,4 +290,57 @@ private function getInstallPath(PackageInterface $package, $global = false)
...
+     */
+    protected function getPluginApiVersion()
+    {
+        return PluginInterface::PLUGIN_API_VERSION;
+    }
+
+    /**
+     * @param PluginInterface $plugin
+     * @param string $capability
+     * @return bool|string The fully qualified class of the implementation or false if none was provided
```
alcohol: Instead of returning false, why not return a `NoCapabilities` classname of sorts? Keeps the return value consistent.

tests/Composer/Test/Plugin/PluginInstallerTest.php
```diff
@@ -314,4 +296,67 @@ public function testPluginRangeConstraintsWorkOnlyWithCertainAPIVersion()
...
+        $plugin = $this->getMockBuilder('Composer\Plugin\PluginInterface')
+                       ->getMock();
+
+        $this->assertFalse($this->pm->getPluginCapability($plugin, 'Fake\Ability'));
+    }
+
+    public function testCapabilityImplementsComposerPluginApiClassAndIsConstructedWithArgs()
+    {
+        $capabilityApi = 'Composer\Plugin\Capability\Capability';
+        $capabilitySpi = 'Composer\Test\Plugin\Mock\Capability';
```
alcohol: Did you mean `spy`? Or is `spi` short for something cryptic?

alcohol:
I personally never cared much for the whole plugin ecosystem or concept. But here's my 2 cents on the subject, if you're willing to read it :d

---

As far as I see it, "plugins" usually do one of two things;

- they try to extend either existing functionality,
- they try to add new functionality.


Now, as far as I understand it;

- The former can partially be done right now through events. But most of that is very limited in both the amount of information it provides (e.g. the state of everything that one might need to know about), and simply the raw amount of "events" that Composer offers (e.g. moments to hook into / interrupt / adjust the process).

- The latter is pretty much a no-go. For example, there's no way to simply "drop-in" new console commands or anything like that at the moment.

This capability stuff sounds neat and all that. **But**, how exactly is it going to work? It feels like it is only one very small part of a solution that is still very much obscure and vague and non-descript. Who knows if this part is even going to fit once we start working out the rest?

src/Composer/Plugin/PluginManager.php
Seldaek: IMO this should rather be `null`. `false` is a value, `null` is the absence of one. foo|null makes sense but foo|bool usually leads to weird stuff :)

alcohol: But, but... `null` is weird. It's like, "I dunno wtf I am doing, so here, have some nothing".

tests/Composer/Test/Plugin/PluginInstallerTest.php
adragus-inviqa: I meant "Service Provide Interface" - a Java-speak for API concrete implementations. The API being what Composer exposes as interfaces and SPI being what userland provides as implementation (concrete classes). If it's too pendatic, I can change it, np.

adragus-inviqa:
> I personally never cared much for the whole plugin ecosystem or concept.

You should.

I really do think that a complex enough and popular enough application benefits quite a bit from some sort of extensibility. You can't possibly cater for all of your userbase's requests. But maybe you didn't care as there already was such a system in place - the event system.

Or maybe because of Composer's focused scope: dependency management and only that. With such a tight ship to run, extensibility is seldom useful, I agree. 

Additional commands I think could very well be added via events, too, why not. 

The reason I like the plugin approach - as opposed to events - is the fact that you don't use dumb callbacks. Not in this PR anyway. They offer more control to the application, instead of delegating to implementator and pray it works. Especially when using an OO approach.

```php
// plugin style
// more work for core people, but more control
$cmdProvider = $this->pluginManager
                    ->getPluginCapability(
                       $plugin, 
                       'Composer\Plugin\Capability\CommandProvider'
                    );

foreach ($cmdProvider->getCommands() as $command) {
    $this->composer->addCommand(
        $command->getTitle(), 
        $command->getActionCallback(), 
        $command->getHelp()
    );
}

// as opposed to
// less work for core people, but less control
$this->triggerEvent('command_manager_before_aggregate', [$composer]);
```
Either way, once an extension point is made, expect less fuss from the community; thus core maintainers have more time for updates and optimizations.

You're asking about how exactly it should work. Something along these lines:
a) core decides, based on people's number of requests or by themselves, that Composer needs an extension point whilst doing X - say - validating through the `validate` command;
b) core adds said extension point, calling all plugins which implement the `ValidatorCapability` or whatever;

I have a functional example here: https://github.com/composer/composer/pull/3377#issuecomment-109114527 (the two links in that comment).

src/Composer/Plugin/PluginManager.php
adragus-inviqa: Why not `I know all, and you get nothing`? 

I'd go with `null`, too.

alcohol:
@adragus-inviqa well, we can agree to disagree there ;-)

That being said, I am definitely not going to stand this in the way if that is what you are worried about. I will support any solution that the majority stands behind. And by standing behind I don't mean just voting for it, but also implementing and maintaining. This is an open source project after all. Contributions are more than welcome. I merely shared my opinion and questions and appreciate the answer you have provided.

Let us proceed :+1:

tests/Composer/Test/Plugin/PluginInstallerTest.php
Seldaek: I don't know about pedantic, but it's obscure to me, I never heard that term, and the Spi means nothing at all when I see it. How about a down to earth `$implementor`?

src/Composer/Plugin/PluginManager.php
```diff
@@ -299,4 +290,57 @@ private function getInstallPath(PackageInterface $package, $global = false)
...
+
+    /**
+     * Returns the version of the internal composer-plugin-api package.
+     *
+     * @return string
+     */
+    protected function getPluginApiVersion()
+    {
+        return PluginInterface::PLUGIN_API_VERSION;
+    }
```
Seldaek: Any reason to having moved that down here? It seems like an unnecessary change.

src/Composer/Plugin/PluginManager.php
```diff
@@ -299,4 +290,57 @@ private function getInstallPath(PackageInterface $package, $global = false)
...
+     */
+    protected function getCapabilityImplementationClassName(PluginInterface $plugin, $capability)
+    {
+        if (!($plugin instanceof Capable)) {
+            return false;
+        }
+
+        $capabilities = (array) $plugin->getCapabilities();
+
+        if (empty($capabilities[$capability]) || !is_string($capabilities[$capability])) {
```
Seldaek: Here I would argue if a plugin returns a capability map that is not a classname we should error out or at the very least print a warning. I'd tend to error though as it clearly means the plugin is broken and the author should know before publishing it.

src/Composer/Plugin/PluginManager.php
```diff
@@ -299,4 +290,57 @@ private function getInstallPath(PackageInterface $package, $global = false)
...
+     * @param string          $capability The fully qualified name of the API interface which the plugin may provide
+     *                                    an implementation.
+     * @param array           $ctorArgs   Arguments passed to Capability's constructor.
+     *                                    Keeping it an array will allow future values to be passed w\o changing the signature.
+     * @return Capability|boolean         Bool false if the Plugin has no implementation of the requested Capability.
+     */
+    public function getPluginCapability(PluginInterface $plugin, $capability, array $ctorArgs = array())
+    {
+        if ($capabilityClass = $this->getCapabilityImplementationClassName($plugin, $capability)) {
+            if (class_exists($capabilityClass)) {
```
Seldaek: Same here if the class name does not exist I would throw as the plugin is very broken.

Seldaek:
OK I think we have a plan, and aside from those few comments the code looks good too. When you find a minute to tackle these I think we can go forward and merge :)

src/Composer/Plugin/PluginManager.php
adragus-inviqa: No reason. I guess I moved when I solved the conflicts. Yeah, it not necessary.

src/Composer/Plugin/PluginManager.php
adragus-inviqa: No problem.

How do you error in Composer? You mean to throw an exception? Or `$this->io->writeError('<warning>`?

src/Composer/Plugin/PluginManager.php
adragus-inviqa: Will do.

tests/Composer/Test/Plugin/PluginInstallerTest.php
adragus-inviqa: Sure thing.

src/Composer/Plugin/PluginManager.php
Seldaek: throw a RuntimeException is fine for such things that should not be overlooked. We use io writes for simple warnings of even errors that don't need to interrupt the program flow. I think here we want to abort.

adragus-inviqa:
@Seldaek - updated. 

Just a reminder: the Composer Plugin API version _will_ be bumped to `1.1.0` by this PR, ok?

Seldaek:
Yeah, I'm ok with it, I just hope it won't break anything/too much :) I'll take a peek at packagist composer-plugin-api requires when I get a chance.

Seldaek:
I think it'll be fine though except that plugin implementing this API will not work anymore on older composer, but that's why we're still in alpha :p

adragus-inviqa:
Yeah, dunno what to say. Maybe give a shout to people on social media or something and give them some time to react? Or if you think you've waited enough, just merge it.

Seldaek:
So I checked packagist and it seems we have ~100 plugin packages that require 1.0.0 still which is quite horrible. *But* I figured that actually the plugin api didn't really change yet.. Until we start using those capabilities, this is just dead code right? So I put it back to 1.0.0 in https://github.com/composer/composer/commit/ddd140fd1cceece4841e2f0182d0c5e4fed5c10e and added a big warning so hopefully people start migrating to proper requires and if not well they've been warned.

Also applied a few changes (837fa805ec9f8dcb1e05e0fca4099f0dab4f1e04) to allow a plugin *not* to provide a capability because before it was assuming it as an error but it's really only an error to define it incorrectly, not to skip a given capability.

Anyway thanks for all the work here! Have a nice weekend.

adragus-inviqa:
Yeah, I thought of mentioning keeping `1.0.0`, but I assumed having 1.0 _and_ a new interface didn't sit too well with you, but if you're ok with it, it's fine by me, too.

Thanks for the suggestions and refactorings.

PR #4130, opened 2015-06-10
Include jar extension directory scanning

ammmze:
Include jar extensions when scanning for artifacts. A jar is more/less equivalent to a zip file and normal zip tools are able to interact with jar files in the same way as a zip file. 

See [JAR (file format)](http://en.wikipedia.org/wiki/JAR_%28file_format%29) for more information about JAR files.

We use a maven as a build tool to generate an api client for various languages, and these are placed in jar archives (maven provides an easy ability to create a jar archive vs creating zip is a little more involved in maven).

### DIFF
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -62,7 +62,6 @@ class Factory
 
     /**
      * @param string $home
-     *
      * @return string
      */
     protected static function getCacheDir($home)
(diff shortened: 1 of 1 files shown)


Seldaek:
The list of file extensions to scan for could be made into an option in $repoConfig but I am a bit reluctant to add jar by default as that means we'll have to keep it in forever for BC..

ammmze:
That sounds like an acceptable alternative

PR #4134, opened 2015-06-11
[docs] mention the `memory_limit` actually being set

craue:
The [current implementation](https://github.com/composer/composer/blob/c471edc488f49f254c77ab97ecc14ea50bf3d2ab/bin/composer#L37) sets `1G`. Also, suggesting to set the `memory_limit` to `2G` would be more useful.


Seldaek:
Thanks

PR #4136, opened 2015-06-12
Typo in Caret Operator example

rahul286:
I think Caret Operator example have wrong version.


rahul286:
Ignore. I got confused between tilde and caret

PR #4139, opened 2015-06-12
Resolves #4138: Removed *Test.php filtering from AutoloadGenerator->dump()

### LINKED ISSUE #4138: Classmaps generated with -o don't consistently include files from autoload-dev
According to the Options sections of the docs for the [`dump-autoload`](https://getcomposer.org/doc/03-cli.md#dump-autoload), [`install`](https://getcomposer.org/doc/03-cli.md#install), and [`update`](https://getcomposer.org/doc/03-cli.md#update) Composer CLI commands, their `-o` flag should "convert PSR-0/4 autoloading to classmap."

However, it appears that classes referenced in [`autoload-dev`](https://getcomposer.org/doc/04-schema.md#autoload-dev) aren't consistently included in this classmap, even if `--no-dev` is not specified.

[This gist](https://gist.github.com/elazar/29d42163cbbb116ca3b8) contains an example where code from `autoload-dev` (specifically `Dev\Foo` as defined in `dev/Foo.php` per the PSR-4 autoloader configured in `autoload-dev`) _does_ appear to be included in the classmap.

[This repo](https://github.com/elazar/phantestic) is an example where code from `autoload-dev` (specifically classes under `Phantestic\Tests` as defined within the `tests` directory per the PSR-4 autoloader configured in `autoload-dev`) _does not_ appear to be included in the classmap, but does contain classes from the `evenement/evenement` dependency and from the PSR-4 autoloader defined in `autoload`.

From a related discussion with @Seldaek:

```
I think it just was omitted when we added autoload-dev. shouldn't be too hard tbh.
down the line it's just something in AutoloadGenerator somewhere. if you search
for psr in that file you should stumble upon it.
```

elazar:


Seldaek:
Thanks

PR #4140, opened 2015-06-12
Add the hash of the parsed content of the composer.json to the lock file, and use it to verify the json is not changed

szeber:
Changing the formatting of the composer.json (ie. if there is a convention of vertically aligning the versions of the dependencies), the hash in the lock file will not match that of the JSON file. Instead if we use the hash of the parsed content, this is not going to be an issue.

This change adds a content-hash key to the lock file with the hash, which is the MD5 of the parsed and sorted content (decode to an associative array, sort it then re-encode and hash). 

This change should be backwards compatible.

Example use case. Given this composer.json:
```javascript
{
    "name": "composer/composer",
// SNIP
    "require": {
        "php":                       ">=5.3.2",
        "justinrainbow/json-schema": "~1.4",
        "seld/jsonlint":             "~1.0",
        "symfony/console":           "~2.5",
        "symfony/finder":            "~2.2",
        "symfony/process":           "~2.1",
        "seld/phar-utils":           "~1.0"
    },
// SNIP
}
```

running `composer.phar require "seld/cli-prompt ~1.0"` would result in the following composer.json:

```javascript
{
    "name": "composer/composer",
// SNIP
    "require": {
        "php":                       ">=5.3.2",
        "justinrainbow/json-schema": "~1.4",
        "seld/jsonlint":             "~1.0",
        "symfony/console":           "~2.5",
        "symfony/finder":            "~2.2",
        "symfony/process":           "~2.1",
        "seld/phar-utils":           "~1.0",
        "seld/cli-prompt": "~1.0"
    },
// SNIP
}
```

After fixing the alignment in the JSON file, the file's hash no longer matches. Having a hash of the contents solves the issue. Currently the other way to solve is to generate a new MD5 for the file and changing the hash value in the lock file.

### DIFF
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -496,7 +496,26 @@ class Factory
     private function getContentHash($composerFilePath)
     {
         $content = json_decode(file_get_contents($composerFilePath), true);
-        ksort($content);
-        return md5(json_encode($content));
+
+        $relevantKeys = array(
+            'require',
+            'require-dev',
+            'conflict',
+            'replace',
+            'provide',
+            'suggest',
+            'minimum-stability',
+            'prefer-stable',
+            'repositories',
+        );
+
+        $relevantContent = array();
+
+        foreach (array_intersect($relevantKeys, array_keys($content)) as $key) {
+            $relevantContent[$key] = $content[$key];
+        }
+
+        ksort($relevantContent);
+        return md5(json_encode($relevantContent));
     }
 }
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -306,7 +306,7 @@ class Factory
             $lockFile = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
                 ? substr($composerFile, 0, -4).'lock'
                 : $composerFile . '.lock';
-            $locker = new Package\Locker($io, new JsonFile($lockFile, new RemoteFilesystem($io, $config)), $rm, $im, md5_file($composerFile));
+            $locker = new Package\Locker($io, new JsonFile($lockFile, new RemoteFilesystem($io, $config)), $rm, $im, md5_file($composerFile), $this->getContentHash($composerFile));
             $composer->setLocker($locker);
         }
 
@@ -484,5 +484,19 @@ class Factory
         $factory = new static();
 
         return $factory->createComposer($io, $config, $disablePlugins);
+    }
+
+    /**
+     * Returns the md5 hash of the sorted content of the composer file.
+     *
+     * @param string $composerFilePath Path to the composer file.
+     *
+     * @return string
+     */
+    private function getContentHash($composerFilePath)
+    {
+        $content = json_decode(file_get_contents($composerFilePath), true);
+        ksort($content);
+        return md5(json_encode($content));
     }
 }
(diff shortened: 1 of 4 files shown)


stof:
There was a previous discussion (which I don't find anymore) about hashing only the parts of the composer.json which could have an impact on the lock file.
For instance, changing the root package autoloading or the root package description does not change anything to the dependency resolution.

I would rather implement such feature instead of adding a new hash built by ignoring only formatting but still taking everything into account, forcing to add yet another hash later when implementing the check in the optimal way.

szeber:
That's a good point. I've gone through the JSON schema, and so far the following values seemed to be relevant to the dependency resolution:
* require
* requie-dev
* conflict
* replace
* provide
* suggest
* minimum-stability
* prefer-stable
* repositories

I'll start the changes on this assumption, but let me know if I missed something.

slbmeh:
One problem this may introduce is BC breakage for plugins that introduce repositories. A plugin manufactured repository cannot be added to the repositories config in case the plugin is not yet installed or compose is called with the `--no-plugins` flag.

If you include the `extra` block, that's a lot of irrelevant information being hashed, but it's sure to take in to account the extra repositories.

It might be a better idea to figure out an ideal way to work with plugin repositories instead.

szeber:
I've added the extra attribute to the list of fields included in the hash. That should solve the BC breakage.

src/Composer/Factory.php
```diff
@@ -485,4 +485,38 @@ public static function create(IOInterface $io, $config = null, $disablePlugins =
...
+            'require',
+            'require-dev',
+            'conflict',
+            'replace',
+            'provide',
+            'suggest',
+            'minimum-stability',
+            'prefer-stable',
+            'repositories',
+            'extra',
```
stof: this is not enough. The name of the root package is also relevant

src/Composer/Factory.php
```diff
@@ -485,4 +485,38 @@ public static function create(IOInterface $io, $config = null, $disablePlugins =
...
+    {
+        $content = json_decode(file_get_contents($composerFilePath), true);
+
+        $relevantKeys = array(
+            'require',
+            'require-dev',
+            'conflict',
+            'replace',
+            'provide',
+            'suggest',
```
stof: sugget should not be there. It does not impact the resolution process at all so it is not relevant field for warning about outdated file (it is purely a display feature)

src/Composer/Factory.php
stof: and the ``version`` field should also be used in case the version is passed explicitly

src/Composer/Package/Locker.php
```diff
@@ -85,6 +88,11 @@ public function isFresh()
     {
         $lock = $this->lockFile->read();
 
+        if (!empty($lock['content-hash'])) {
+            // There is a content hash key, use that instead of the file hash
+            return $this->contentHash == $lock['content-hash'];
```
stof: this should use strict comparison

src/Composer/Factory.php
```diff
@@ -485,4 +485,38 @@ public static function create(IOInterface $io, $config = null, $disablePlugins =
...
+
+        $relevantContent = array();
+
+        foreach (array_intersect($relevantKeys, array_keys($content)) as $key) {
+            $relevantContent[$key] = $content[$key];
+        }
+
+        ksort($relevantContent);
+        return md5(json_encode($relevantContent));
+    }
```
Seldaek: IMO this method should be on the locker to keep locker matters there in a central place. Maybe it should just accept the whole json string as input so it can compute both hashes internally.

szeber:
I've added the mentioned fields and moved the checking code to the Locker.

Seldaek:
Thanks @szeber!

PR #4141, opened 2015-06-13
Try to symlinks on Windows systems, too

derula:
The symlink() function supports Windows since PHP 5.3. Most current Windows installations use NTFS, which supports symlinks.

The problem with these wrapper scripts is that they don't work as expected in some cases, for example if the bin file is a jar file.

In this case, on Windows, you will end up calling `java -jar vendor/bin/example.jar`, with example.jar really being an sh script.

So we should try to create symlinks on Windows, too. In case it doesn't work, there is still a fallback.

Or is there any reason why this isn't done?

### DIFF
--- a/src/Composer/Installer/LibraryInstaller.php
+++ b/src/Composer/Installer/LibraryInstaller.php
@@ -219,34 +219,34 @@ class LibraryInstaller implements InstallerInterface
                 $this->io->writeError('    Skipped installation of bin '.$bin.' for package '.$package->getName().': name conflicts with an existing file');
                 continue;
             }
-            if (defined('PHP_WINDOWS_VERSION_BUILD')) {
-                // add unixy support for cygwin and similar environments
-                if ('.bat' !== substr($binPath, -4)) {
-                    file_put_contents($link, $this->generateUnixyProxyCode($binPath, $link));
-                    @chmod($link, 0777 & ~umask());
-                    $link .= '.bat';
-                    if (file_exists($link)) {
-                        $this->io->writeError('    Skipped installation of bin '.$bin.'.bat proxy for package '.$package->getName().': a .bat proxy was already installed');
+            $cwd = getcwd();
+            try {
+                // symlinks are not always supported, for example on legacy
+                // windows-specific filesystems such as smb or fat
+                $relativeBin = $this->filesystem->findShortestPath($link, $binPath);
+                chdir(dirname($link));
+                if (false === @symlink($relativeBin, $link)) {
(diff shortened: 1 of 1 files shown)


slbmeh:
> Or is there any reason why this isn't done?

I believe it's because Windows XP was still pretty common at the time of implementation.

derula:
Hmm, but even in this case, would it not work with my proposed change as well? If executed on Windows XP (or on any file system that does not support symlinks, or on newer Windows versions when run without administrator privileges, etc.), it would create the scripts as a fallback, wouldn't it? I don't have a Windows XP box to test this on, though...

alcohol:
It looks good, but I'm not really in a position to test unfortunately. :-(

curry684:
Has become obsolete with #4845. As symlinks require elevation on Vista and later a symlink approach is not viable regrettably.

derula:
@curry684 can you please clarify how #4845 makes this obsolete?

1. Junctions can only link to directories. In this case, symlinks to files are being created.
2. If a symlink cannot be created, it will automatically fall back to wrapper scripts. So the current behavior remains intact, and a feature is added that symlinks are created _if_ the process is run with elevated privileges.

I would understand if this got rejected for weird behavior (certain feature that only works if elevated, which usually never happens), but to my understanding, the general idea of this would still provide an improvement in certain cases. (Yes, I saw that there are conflicts now... I could try to fix that if anyone is still interested.)

curry684:
Sorry didn't notice this was about files. I don't really think however one would want to always run Composer with elevation - that means you can no longer run it from an IDE for example.

In return for having to have an elevated command prompt open permanently you get a major arbitrary code execution loophole if an indirectly referenced plugin introduces post-install scripts that nest some keyloggers and DDoS bots deep in your registry, as it has unlimited privileges to do so. I think that would be irresponsible to add to Composer.

Seldaek:
I tend to agree with @curry684 that in the current state of affairs symlinks on windows with php are not really practical. As for using jar files as binaries.. I guess in this sort of rare edge case referencing the jar direclty in the vendor/foo/bar/bin/file.jar isn't *that* bad?

tored:
Just want to update on the symlink situation on Windows. Since Creators update for Windows 10 you have the possibility to avoid elevation if you have your system correctly configured by setting it to developer mode (or you can set the correct policy). Here is how Git for Windows handles symlinks.

https://github.com/git-for-windows/git/wiki/Symbolic-Links

Git for Windows requires however that you clone your repository with a flag to enable it for Windows.

PR #4143, opened 2015-06-14
Dropped obsolete condition

staabm:
The conditional a few lines above already checks for `bytesMax > 0`


staabm:
Any feedback regarding this trivial change @Seldaek  ?

PR #4145, opened 2015-06-14
Simplify gzip encoding response header check

staabm:
I am not 100% sure I understood why the code looked like it did before.

The code was dependent on the order of the actual headers, so when the content-encoding header wasnt the last, it was detected as unsupported.

Blaming the line showed that it is somehow related to redirect, see https://github.com/composer/composer/commit/22149d3a70c726c7709a221bfc62f65b18049bb9 .. This info wasnt helpfull for me though.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -207,9 +207,7 @@ class RemoteFilesystem
             foreach ($http_response_header as $header) {
                 if (preg_match('{^content-encoding: *gzip *$}i', $header)) {
                     $decode = true;
-                    continue;
-                } elseif (preg_match('{^HTTP/}i', $header)) {
-                    $decode = false;
+                    break;
                 }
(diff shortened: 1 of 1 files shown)


staabm:
Ps: sorry for all those small PRs but I feel it eases handling to have this in tiny pieces.

Actually I try to simplify where possible so next actual next refactoring step can be achieved more easily.

Seldaek:
So the reason it was doing this, is that if a request is redirect, `$http_response_header` will contain something like:

```php
[
    'HTTP/1.0 302',
    'Location: foo',
    'Content-Type: bla',
    'HTTP/1.0 200',
    'Content-Type: foo',
    'Content-Length: ...'
]
```

So basically you get the two responses' headers together in the array, that's why it resets the flag to false whenever it encounters a HTTP/.. header, because that one should always be first AFAIK. I believe the existing code is correct, but it could use a comment :)

PR #4146, opened 2015-06-14
Moved errorCode var closer to the lines where it is used

staabm:
The prev block can now be refactored in a separate method


staabm:
In case you consider this one too tiny, I added this and the next logical step together in https://github.com/composer/composer/pull/4150

Seldaek:
Yeah might as well close this, I'll try to review the other one.

PR #4163, opened 2015-06-18
[WIP] RepositorySet / Pool Split & Load only explicitly referenced package data

naderman:
This is a fixed version of https://github.com/composer/composer/pull/3994

Todo:
- [ ] RepositorySetTest
- [ ] don't use object state to construct pool in RepositorySet
- [x] move request name collection logic into request - solver should take repo set
- [ ] evaluate if RepositorySet and CompositeRepository should be merged
- [ ] refactor policy to work with repositoryset in addition to pool
- [ ] remove uses of small pools from installer
- [ ] this should be empty: `grep -ri "pool" src/ | grep -v "src/Composer/DependencyResolver"`

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -19,13 +19,13 @@ use Composer\Installer\ProjectInstaller;
 use Composer\Installer\InstallationManager;
 use Composer\IO\IOInterface;
 use Composer\Package\BasePackage;
-use Composer\DependencyResolver\Pool;
 use Composer\DependencyResolver\Operation\InstallOperation;
 use Composer\Package\Version\VersionSelector;
 use Composer\Repository\ComposerRepository;
 use Composer\Repository\CompositeRepository;
 use Composer\Repository\FilesystemRepository;
 use Composer\Repository\InstalledFilesystemRepository;
+use Composer\Repository\RepositorySet;
 use Composer\Script\ScriptEvents;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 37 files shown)


stof:
Can you add a test covering the broken case ?

naderman:
@stof yes I will, there are also more bugs.

naderman:
@Seldaek so I incremented the plugin api version to 2.0.0 for this, as the calls to plugins have changed to use repository sets rather than pools. This means all plugins will stop working when updating to this version - any idea how we would handle updating to this version? Afterall if we merge something like this into master it immediately breaks all installs using plugins. So we need some smoother way to upgrade, but I'm not sure what that would be.

staabm:
@naderman would it be possible to ship a Adapter class which can translate between API V1 & V2?
This could be used for plugins which rely on V1 automatically.

naderman:
No that's really not possible due to the nature of these changes, otherwise I would have suggested that already.

I imagine we'll have to do another release and announce that this breaking change is coming, and then anyone relying on old plugins will need to stick to the release, while master is updated with these changes.

Seldaek:
I think yes we need to do a soft migration process of some kind for this, so maybe we need to delay this a little while, which sucks. On the other hand, I guess we should take the opportunity to look at the plugin API in depth and break anything else that needs breaking (cc @nevvermind).

Seldaek:
@naderman what about what we discussed regarding having DefaultPolicy not dependent on a Pool btw? It would be nice if it could be used to sort packages standalone, but it doesn't seem to have change yet here.

naderman:
Should check if this resolves https://github.com/composer/composer/issues/6840

naderman:
Closing this in favour of the new PR which is based on current composer.

PR #4165, opened 2015-06-19
Minor optimization on the DependencyResolver

c88lopez:
The commit message contains the reason of the request.

### DIFF
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -683,8 +683,9 @@ class Solver
                 $systemLevel = $level;
             }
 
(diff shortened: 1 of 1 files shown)


stof:
This is wrong. You are now counting the rules only before starting the iteration instead of counting them at each iteration when evaluating the condition to stop the loop. Given that the code inside the loop can inject more rules, this changes the logic of the solver, potentially breaking things.

@naderman looks like we are missing some tests to cover such cases in the solver though

naderman:
Indeed, there should be a test that fails with this change.

naderman:
Closing this, created an issue to make a test that fails if this is applied.

PR #4169, opened 2015-06-21
Set this test as slow

remicollet:
This test requires internet connectivity.

I notice other test are set as "slow" (the pear ones).
Perhaps a group "online" will be better.

This will allow to run test suite from computer without network.

Reproducer:

    echo "# nothing" > /etc/resolv.conf
    phpunit -v
    1) Composer\Test\Util\RemoteFilesystemTest::testCaptureAuthenticationParamsFromUrl
Failed asserting that 0 matches expected 404.


remicollet:
@Seldaek can youplease check this really minor change which will make downstream life easier ?

Seldaek:
Thanks, sorry for the delay

remicollet:
Thanks

PR #4170, opened 2015-06-22
Attempting to add support for windows network absolute paths.

WillSkates:


alcohol:
Weird, when I want to directly open samba mounts in windows explorer, I use `\\1.2.3.4`, not `//`. Is this PR necessary?

alcohol:
Uhm, also, `substr($path, 0, 1) === '/'` matches that same condition already. Closing this as it literally adds nothing.

WillSkates:
Heyah, sorry for the long time to reply.

I was running into situations where the path is automatically determined for you. For example using the git shell on windows machines where z:\index.php would get mapped automatically to //z/index.php. z:\ is a drive on my machine that I've mapped some network storage onto.

Cheers,
Will

PR #4171, opened 2015-06-22
splitting up some docs, closes #4155

### LINKED ISSUE #4155: Split config docs from 04-schema into 07-config or such
That way we could do a proper section per config setting, and get deep linking to precise settings for free.

alcohol:
Work in progress, feedback please.

I feel like it doesn't really make sense to put config **after** community. However, because of the numbering of the filenames and this actually being reflected in the urls generated, this PR might break some external links. Should I revert that change and move config to the last page (07)?

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -64,18 +64,18 @@ If there is a `composer.lock` file in the current directory, it will use the
 exact versions from there instead of resolving them. This ensures that
 everyone using the library will get the same versions of the dependencies.
 
-If there is no `composer.lock` file, composer will create one after dependency
+If there is no `composer.lock` file, Composer will create one after dependency
 resolution.
 
 ### Options
 
 * **--prefer-source:** There are two ways of downloading a package: `source`
-  and `dist`. For stable versions composer will use the `dist` by default.
+  and `dist`. For stable versions Composer will use the `dist` by default.
   The `source` is a version control repository. If `--prefer-source` is
-  enabled, composer will install from `source` if there is one. This is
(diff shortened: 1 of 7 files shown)


doc/03-cli.md
```diff
@@ -424,7 +426,7 @@ php composer.phar config repositories.foo vcs https://github.com/foo/bar
 ## create-project
 
 You can use Composer to create new projects from an existing package. This is
-the equivalent of doing a git clone/svn checkout followed by a composer install
+the equivalent of doing a git clone/svn checkout followed by a "composer install"
```
nevvermind: Ticks instead of quotes?

alcohol: I'm merely trying to follow the styling used in general. I found that often "" are used when the subject doesn't directly relate to the paragraph or an inline code block might distract from the flow of reading.

/cc @Seldaek got any feedback on desired use of quotes or backticks here?

doc/06-config.md
```diff
@@ -0,0 +1,177 @@
+# Config
+
+This chapter will describe the `config` section of the `composer.json` 
+[schema](04-schema.md).
+
+## config <span>([root-only](04-schema.md#root-package))</span>
+
+A set of configuration options. It is only used for projects.
+
+## Properties
```
Seldaek: I think I'd remove lines 6 to 10 above and use `##` for the properties so they're linkable.

Seldaek: Well they might be linkable anyway but still, unnecessary nesting IMO

alcohol: Roger

Issue #2752, opened 2014-02-24
create-project should list suggested packages


hkdobrev:
When you have a `suggest` section in the `composer.json` file of the package you use for the the `create-project` command, suggestions are listed only for the dependencies of the package. The `suggest` section in the root `composer.json` is ignored.

Here is an example `composer.json`: https://github.com/Despark/kohana/blob/d387f30d55170875014bf2e6bd19b8499f851879/composer.json#L19-L42

and the suggestions it lists are:
```
kohana/core suggests installing ext-http (*)
kohana/core suggests installing ext-mcrypt (*)
openbuildings/jam suggests installing openbuildings/jam-monetary (Convert and format currencies and amounts with a Jam field)
openbuildings/jam suggests installing openbuildings/jam-tart (Jam admin builder)
openbuildings/jam suggests installing openbuildings/jam-closuretable (Closuretable implementation for Jam ORM)
symfony/console suggests installing symfony/event-dispatcher ()
```

PR #4177, opened 2015-06-23
Add suggests command

### LINKED ISSUE #4175: Get an aggregate list of suggestion from cli
Hello,
When you run ```composer show vendor/package``` You have the list of suggestions for the package. 

But, you can't have an aggregate view of all packages suggestion.

So I propose this command ```composer suggests``` to make a aggregation of suggestion like we have on composer install.

alcohol:
Closes #4175, #2819

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -296,14 +296,11 @@ in your browser.
 
 ## suggests
 
-To list all of vendors suggesting to install packages, you can use the `suggests` command.
-
-    $ php composer.phar suggests
-
+Lists all packages suggested by currently installed set of packages.
 
 ### Options
 
-* **--dev:** Show dev suggests.
+* **--no-dev:** Exludes suggestions from `require-dev` packages.
 
 ## depends
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -87,7 +87,7 @@ resolution.
   installing a package, you can use `--dry-run`. This will simulate the
   installation and show you what would happen.
 * **--dev:** Install packages listed in `require-dev` (this is the default behavior).
-* **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader 
+* **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader
   generation skips the `autoload-dev` rules.
 * **--no-autoloader:** Skips autoloader generation.
 * **--no-scripts:** Skips execution of scripts defined in `composer.json`.
@@ -302,7 +302,7 @@ to limit output to suggestions made by those packages only.
 
 ### Options
 
-* **--no-dev:** Exludes suggestions from `require-dev` packages.
+* **--no-dev:** Excludes suggestions from `require-dev` packages.
 * **-v[v]:** Increased verbosity adds suggesting package name (`-v`) and
   reason for suggestion (`-vv`).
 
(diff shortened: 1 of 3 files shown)


alcohol:
I added the `packages` argument but I am not sure if it holds any real value. Should I leave it in (and document it) or remove that commit?

alcohol:
Also, maybe the "reason" should require at least 1 level of verbosity (`-v`)? And maybe I could output it to `stderr` still then. That way `suggests` would only return the package names on `stdout`.

stof:
mixing the output to STDOUT and STDERR for different parts of the message would be confusing IMO

alcohol:
Alright, I'm open to suggestions. Got a decent format/approach in mind that is able to work well for both scenarios?

alcohol:
I guess the reason could go to stdout as well. Since people should be expecting it if they supply `-v`. So they probably want to parse/grep the output (assuming a pipe scenario).

stof:
IMO, everything should go to STDOUT in this case.

and the output is not really readable currently, as you have both the package suggesting things and then the list of suggested packages. If you want to keep the list grouped by suggesting package, you should indent probably it

alcohol:
Well the `show` command doesn't use any indentation, merely colors. And it is somewhat easy to read, even without colors. You sure it matters?

stof:
the show command does not group the list. It only has a suggestion per line saying ``XXX suggest YYY``. So there is no need to indeed groups as suggestions are not grouped

alcohol:
Not like `xxx suggests yyy`;

```
# composer show composer/composer

...cut...

suggests
ext-zip Enabling the zip extension allows you to unzip archives, and allows gzip compression of all internet traffic
ext-openssl Enabling the openssl extension allows you to access https URLs for repositories and packages
```

alcohol:
@stof by the way, not sure if this makes sense or not (to add), but how could I add the `suggests` of the root package? Where can I find that data?

alcohol:
I personally believe this is the cleanest solution (based on verbosity). And with or without the colors it is quite readable in this format.

```
rob@thuis /srv/git/composer (add-suggests-command) $ bin/composer suggests
psr/log
symfony/event-dispatcher
symfony/process
phpunit/php-invoker
ext-soap
ext-uopz
rob@thuis /srv/git/composer (add-suggests-command) $ bin/composer suggests -v
symfony/console suggests psr/log
symfony/console suggests symfony/event-dispatcher
symfony/console suggests symfony/process
phpunit/phpunit suggests phpunit/php-invoker
phpunit/phpunit-mock-objects suggests ext-soap
sebastian/global-state suggests ext-uopz
rob@thuis /srv/git/composer (add-suggests-command) $ bin/composer suggests -vv
symfony/console suggests psr/log: For using the console logger
symfony/console suggests symfony/event-dispatcher: *
symfony/console suggests symfony/process: *
phpunit/phpunit suggests phpunit/php-invoker: ~1.1
phpunit/phpunit-mock-objects suggests ext-soap: *
sebastian/global-state suggests ext-uopz: *
```

stof:
ah sorry. I thought about ``composer install``.

for composer show, there is a single list of suggestions, so it is also the reason why indentation for groups is not necessary.

for the root package, use ``$composer->getPackage()->getSuggests()``. but I'm not sure it makes sense. Using suggestions in the root package (when it is a project, not a library meant to be referenced as a dependency) is useless: either you use it and you should depend on it, or you don't use it and there is no point suggesting it

alcohol:
Regarding root package requires, it was just something I was wondering about, but it is more related to this issue: #2752 - but I just realised there already is a PR with a fix for it available :-).

Seldaek:
Thanks, I merged the -v and -vv together though because IMO at -v it was already not so useful for forwarding output so might as well show everything.

Issue #3003, opened 2014-05-21
Detect `files` autoload conflicts when a second autoloader gets called


Seldaek:
Not sure if possible but should be investigated, to solve cases like https://github.com/guzzle/guzzle/issues/676

itafroma:
What about just changing [L585 in AutoloadGenerator.php](https://github.com/composer/composer/blob/8e9659bd8317ad3f623c4d17940380443cf8772c/src/Composer/Autoload/AutoloadGenerator.php#L585) to `require_once` instead of require? We ran into this issue and making that change solved the problem.

stof:
@itafroma this won't help to solve the issue of @PommeVerte. A global install will have a **different** file containing the same code, as you will have it installed globally and locally.

The **only** solution for that is to have a safeguard in the file checking whether functions can be defined safely (until the day we have function autoloading in PHP to be able to load them on-demand like classes)

stof:
@samdark the content of ``autoload_files.php`` does not end up in the Composer autoloader, as the files strategy is actually not doing autoloading but preloading (and is used for cases where autoloading is not possible in PHP)

itafroma:
@stof Perhaps it wouldn't solve @PommeVerte's issue, but it would solve the issue as originally reported. `require_once` is a construct designed for this purpose: why shouldn't Composer only `require_once` in this case?

stof:
@itafroma the original issue is when having the same library installed globally and locally. So ``require_once`` will not solve anything here as you have 2 **different** files defining the same function.

Seldaek:
Fixed by #4186

PR #4186, opened 2015-06-27
Explicitly check if a file has already been required before requiring it

jeskew:
This would address #3003 by assigning an identifier to each included file during `AutoloadGenerator::dump`. Currently, this identifier is the value returned by `md5_file` (to address the concern raised in https://github.com/guzzle/guzzle/issues/676#issuecomment-43651280), but it could also be a logical file address such as `$packageName::$relativePath`, e.g. `guzzlehttp/guzzle::src/functions.php`. 

This is preferable to simple changing `require` to `require_once` because it prevents two copies of the same file from being loaded, as would happen if a globally installed package loads project's autoloader and both rely on the same package. A known example of this is using a globally installed copy of CodeCeption on a project requiring Guzzle.

Would also address:
https://github.com/guzzle/guzzle/issues/1146
https://github.com/guzzle/psr7/issues/22

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -410,8 +410,9 @@ EOF;
     protected function getIncludeFilesFile(array $files, Filesystem $filesystem, $basePath, $vendorPath, $vendorPathCode, $appBaseDirCode)
     {
         $filesCode = '';
-        foreach ($files as $functionFile) {
-            $filesCode .= '    '.$this->getPathCode($filesystem, $basePath, $vendorPath, $functionFile).",\n";
+        foreach ($files as $fileIdentifier => $functionFile) {
+            $filesCode .= '    ' . $fileIdentifier . ' => '
+                . $this->getPathCode($filesystem, $basePath, $vendorPath, $functionFile) . ",\n";
         }
 
         if (!$filesCode) {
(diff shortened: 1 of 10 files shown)


alcohol:
Wouldn't it be easier instead if perhaps the Composer autoloader defined a constant (e.g. `COMPOSER_AUTOLOAD`, and people could use that to check if the autoloader has been loaded already or not?

Basically approaching the situation from a different angle. Just trying to add more ideas. :-)

jeskew:
@alcohol The two autoloaders might load different files (e.g., if you install PHPUnit globally and Guzzle locally), so I'm not sure a simple flag would be useful.

alcohol:
What if the value of it was the resolved path to the autoloader? For
comparison values.

jeskew:
I'm not sure that would solve the issue reported in #3003, which was that the same file, existing in two locations on disk, could be included by two separate autoloaders.

alcohol:
> as would happen if a package is installed globally and in a project

I don't understand how this is relevant? Globally installed doesn't mean it is always loaded. You would still have to explicitly include `~/.composer/vendor/autoloader.php` first if I am not mistaken?

jeskew:
That's correct, and that is the issue that was reported. https://github.com/composer/composer/issues/3003#issuecomment-64849282

stof:
@alcohol the problem occurs when a tool installed globally (PHPUnit for instance) also loads the autoloading of your project after loading the global one

jeskew:
@stof Do you think this is a good way to solve the problem? I used file checksums instead of logical identifiers so that accidentally loading two version of the same package would continue to cause a failure, but using something like `$packageName::$pathRelativeToVendorDir => $absolutePath` would also be viable.

Seldaek:
Thanks!

Issue #4184, opened 2015-06-27
Files are not automatically autoloaded for triggered scripts


jeskew:
I'm using a static method as a `post-autoload-dump` script and found that files are not automatically loaded.  This can be resolved by explicitly requiring the Composer autoloader, but I wasn't sure if this was intentional. Essentially, in order to use a function in static method being used as a post-autoload-dump hook, I have to `require 'vendor/autoload.php'` even though the autoloader is the one calling the method.

The [AWS Symfony Bundle](https://github.com/aws/aws-sdk-php-symfony/blob/master/build/DocumentationUpdater.php#L22) has to explicitly require the autoloader in a triggered script.

alcohol:
>For any given event:
>
>* Scripts execute in the order defined when their corresponding event is fired.
>* An array of scripts wired to a single event can contain both PHP callbacks and command-line executable commands.
>* PHP classes containing defined callbacks must be autoloadable via Composer's autoload functionality.

It's because the `autoload.php` file is not used for the `scripts` part, but rather the autoloader functionality itself. This can resolve classes, but indeed it does not implicitly load the `files` autoload definitions.

I'm not even sure if `require 'vendor/autoload';` would work on an initial install. It depends on whether or not the autoload files are dumped before or after the event gets triggered. So your workaround might not even work in all scenarios. Did you try and see if it covers all event types?

jeskew:
That's a good point, but there's already a disclaimer on the documentation page. https://github.com/composer/composer/blame/master/doc/articles/scripts.md#L64

alcohol:
That disclaimer is about Composer managed dependencies though. So say your script requires a class from `symfony/console` which is in your `require` section. Composer cannot guarantee that that package is already installed when your event gets called. 

I guess it's sort of related to what I was hinting at. However, this does not depend on when the autoloader gets dumped, but purely on the event type you are using and whether or not it gets triggered before or after said package is installed.

jeskew:
Right; one can't be certain of the state of your dependencies before install or update has been called. Those dependencies might include autoloaded files. As the documentation is written, I believe it's clear that scripts that rely on external dependencies only make sense to run after install or update has run.

jeskew:
Solved by #4186 and #4187. Thanks!

PR #4187, opened 2015-06-27
Mention need to require autoloader in hooks using functions

jeskew:
Add a line and example to the documentation covering the need to require the autoloader file in order to use autoloaded files.

Addresses #4184

### DIFF
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -83,10 +83,8 @@ and command-line executable commands.
 - PHP classes containing defined callbacks must be autoloadable via Composer's
 autoload functionality.
 - If a defined callback relies on functions defined outside of a class, the 
-callback must explicitly require the composer autoloader. If used in a 
-context where`vendor/autoload.php` might not yet exist (such as during a 
-`pre-install` or `pre-update` command), the callback should explicitly require
-whatever files within your root package it needs to execute successfully.
+callback itself is responsible for loading the appropriate files, as no files 
+are autoloaded during Composer commands.
 
 Script definition example:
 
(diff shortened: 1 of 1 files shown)


alcohol:
This will fail if `autoload.php` does not exist yet.

jeskew:
@alcohol As we discussed in the issue related to this PR, I believe the existing disclaimer about the state of the autoloader being undetermined in pre-install and pre-update commands covers the possible non-existence of `vendor/autoload.php`. I could repeat part of that disclaimer in the bullet point added in this PR if you believe it's unclear.

alcohol:
Well at least set a good example :-(. 
Don't use string interpolation, `$vendorDir . '/autoload.php'` is cleaner.

Also I think the current wording is a bit vague as to what _kind_ of functions you are referring to. Remove the reiteration of the disclaimer (it's not necessary), and try to emphasise it's about functions that are normally only available via the `files` autoload attribute.

jeskew:
What phrasing would you suggest is more clear than 'functions defined outside of a class'?

I've always found interpolation to be cleaner, and since there's [no performance difference](http://nikic.github.io/2012/01/09/Disproving-the-Single-Quotes-Performance-Myth) between interpolation and concatenation, we'll have to agree to disagree on the aesthetics.

alcohol:
We can disagree on aesthetics, but we can agree on convention. So if you don't mind.

alcohol:
As for phrasing, maybe it's better to turn it around and simply make it more clear that "must be autoloadable via Composer's autoload functionality" does not mean that the autoloader is actually loaded, but merely that the same mechanism is used and that this only applies to the `psr-0`, `psr-4` and `classmap` autoload types, not `files`.

jeskew:
@alcohol Were there any other changes you wanted to see made?

alcohol:
Well,

> If a defined callback relies on functions defined outside of a class, the callback itself is responsible for loading the appropriate files, as no files beyond those required for `psr-0`, `psr-4`, and `classmap` autoloading are loaded during Composer commands.

could probably better be phrased as

> Callbacks can only autoload classes from psr-0, psr-4 and classmap definitions. If a defined callback relies on functions defined outside of a class, the callback itself is responsible for loading the file containing these functions.

PR #4192, opened 2015-06-29
Fix: Add missing imports

localheinz:
This PR

* [x] adds a bunch of missing imports

### DIFF
--- a/src/Composer/DependencyResolver/Operation/MarkAliasInstalledOperation.php
+++ b/src/Composer/DependencyResolver/Operation/MarkAliasInstalledOperation.php
@@ -13,6 +13,7 @@
 namespace Composer\DependencyResolver\Operation;
 
 use Composer\Package\AliasPackage;
+use Composer\Package\PackageInterface;
 
 /**
  * Solver install operation.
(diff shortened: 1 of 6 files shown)


xabbuh:
Wouldn't it be better to check why Composer didn't fail even when these imports where missing? Changes can be high that there is orphaned code that could safely be removed. Otherwise, it might be a good idea to add tests that cover these paths to make sure they actually work as intended.

alcohol:
@xabbuh as far as I can tell, most (or all) of those are only related to references in `docblock`s.

localheinz:
@xabbuh @alcohol 

Totally up for tests, if you'd prefer me to add them!

Here's what I could do:

* amend this PR to add imports only where missing for docblocks
* create separate PR with tests for those occasions where actual objects are created, but would currently fail because imports are missing

alcohol:
@localheinz No need, they're all `docblock` related only (I've checked). Nothing to test here.

localheinz:
@alcohol 

:+1: Thank you!

xabbuh:
> @localheinz No need, they're all docblock related only (I've checked). Nothing to test here.

Then please completely ignore my comment. :)

PR #4196, opened 2015-06-29
Fix: Import Symfony\Component\Finder\Finder

localheinz:
This PR

* [x] imports `Symfony\Component\Finder\Finder`, fixing a docblock

### DIFF
--- a/src/Composer/Package/Archiver/ArchivableFilesFinder.php
+++ b/src/Composer/Package/Archiver/ArchivableFilesFinder.php
@@ -13,8 +13,7 @@
 namespace Composer\Package\Archiver;
 
 use Composer\Util\Filesystem;
-
-use Symfony\Component\Finder;
+use Symfony\Component\Finder\Finder;
 
 /**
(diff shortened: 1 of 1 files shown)


alcohol:
I would..

- [ ] remove the whitespace (empty line) in between both `use` statements,
- [ ] and possibly even import `Symfony\Component\Finder\Finder` (not just the parent namespace).

It doesn't seem like anything else is used from the `Symfony\Component\Finder` namespace.

alcohol:
@localheinz do you think you could address those points? :-)

localheinz:
@alcohol 

Hang on a moment!

localheinz:
@alcohol 

Updated!

Seldaek:
At line 51 there is a `new Finder\Finder()` call that is now failing.

alcohol:
Uh yeah, don't forget to update the class instantiation after updating the import :p

localheinz:
@alcohol @Seldaek 

Amended!

PR #4199, opened 2015-06-30
>=2.3.0 now includes pre-releases, added docs for this and -stable suffix Fixes #4080.

### LINKED ISSUE #4080: Version constraint matching alpha,beta,rc releases does not work as expected
Consider the following version constraint:

`>=2.7.0-dev,<2.7.0`

this should include every development version and alpha,beta,rc release that comes before the 2.7.0 release. the match works for the first part:
http://semver.mwl.be/#?package=symfony%2Fsymfony&version=%3E%3D2.7.0-dev&minimum-stability=dev
But as soon as the second part is added the constraint does not match anything:
http://semver.mwl.be/#?package=symfony%2Fsymfony&version=%3E%3D2.7.0-dev,%3C2.7.0&minimum-stability=dev

I also tried this with the following composer.json to verify the bug is not only on `semver.mwl.be`:

```json
{
	"minimum-stability": "dev",
    "require": {
        "symfony/symfony": ">=2.7.0-dev,<2.7.0"
    }
}
```

```
# composer update

  Problem 1
    - The requested package symfony/symfony could not be found in any version, there may be a typo in the package name.
```

See https://github.com/FriendsOfPHP/security-advisories/pull/72#discussion_r31214680 for more details. /cc @stof

sminnee:
On the 4080 ticket we concluded that the feature works correctly but could use documentation.

However, I also noticed that >=2.3.0 was inconsistent with the general rule "when stabilities aren't specified, treat all pre-releases the same way as you treat the stable release". So I've made a functional change as well.

### DIFF
--- a/doc/01-basic-usage.md
+++ b/doc/01-basic-usage.md
@@ -94,6 +94,29 @@ so using [stability flags](04-schema.md#package-links). To change that for all
 packages instead of doing per dependency you can also use the
 [minimum-stability](04-schema.md#minimum-stability) setting.
 
+If you are using range comparisons when selecting non-stable packages, and you
+specify a numeric version number (that is, no suffix indicating alpha, beta,
+rc, or stable), then both non-stable and stable versions of a particular
+release number will be treated as equally valid.
+
+ * `>=`/`<=` will accept non-stable releases as well as the stable release.
+ * `<`/`>` will reject non-stable releasese as well as the stable release.
+
+If you wish to consider only the stable release in the comparison, add the
+suffix `-stable` to the version number.
+
+Here are some examples:
+
+ Example         | Interpretation
+ --------------- | --------------
+`>=1.0.0`        | Any release, stable or non-, of 1.0.0 will be allowed
+`>=1.0.0-stable` | Only the stable release of 1.0.0 will be allowed
+`<2.0.0`         | Neither release, stable or non-, of 2.0.0 will be allowed
+`<2.0.0-stable`  | Only the stable release of 2.0.0 will be disallowed; non-stable releases will be allowed
+
+Note that the packages matched by these constraints are still checked against
+the `minimum-stability` setting and each package's stability flags.
+
 ### Test version constraints
 
 You can test version constraints using [semver.mwl.be](http://semver.mwl.be). Fill in
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -443,9 +443,11 @@ class VersionParser
 
                 if (!empty($stabilityModifier) && $this->parseStability($version) === 'stable') {
                     $version .= '-' . $stabilityModifier;
-                } elseif ('<' === $matches[1]) {
+                } elseif ('<' === $matches[1] || '>=' === $matches[1]) {
                     if (!preg_match('/-' . self::$modifierRegex . '$/', strtolower($matches[2]))) {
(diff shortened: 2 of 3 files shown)


alcohol:
Your functional change is highly undesirable. > and >= by default should assume stable as that makes the most sense. They should **not** include pre-releases.

The only reason that < and <= assumes -dev is because otherwise the behaviour isn't what most people expect it to be.

sminnee:
OK, I've switched back to #4080 to continue the discussion.

sminnee:
OK. I've updated the docs and force-pushed a squashed branch.

doc/01-basic-usage.md
```diff
@@ -94,6 +94,29 @@ so using [stability flags](04-schema.md#package-links). To change that for all
...
+
+ Example         | Interpretation
+ --------------- | --------------
+`>=1.0.0`        | Any release, stable or non-, of 1.0.0 will be allowed
+`>=1.0.0-stable` | Only the stable release of 1.0.0 will be allowed
+`<2.0.0`         | Neither release, stable or non-, of 2.0.0 will be allowed
+`<2.0.0-stable`  | Only the stable release of 2.0.0 will be disallowed; non-stable releases will be allowed
+
+Note that the packages matched by these constraints are still checked against
+the `min-stability` setting and each package's stability flags.
```
alcohol: No such thing as `min-stability`, should be `minimum-stability`.

src/Composer/Package/Version/VersionParser.php
```diff
@@ -443,9 +443,11 @@ private function parseConstraint($constraint)
 
                 if (!empty($stabilityModifier) && $this->parseStability($version) === 'stable') {
                     $version .= '-' . $stabilityModifier;
-                } elseif ('<' === $matches[1]) {
+                } elseif ('<' === $matches[1] || '>=' === $matches[1]) {
                     if (!preg_match('/-' . self::$modifierRegex . '$/', strtolower($matches[2]))) {
-                        $version .= '-dev';
+                        if(substr($matches[2], 0,4) !== 'dev-') {
```
alcohol: Spaces please :-)

doc/01-basic-usage.md
sminnee: Fixed in rebased commit.

src/Composer/Package/Version/VersionParser.php
sminnee: Fixed in rebased commit.

alcohol:
Cheers.

Seldaek:
Thanks Sam

doc/01-basic-usage.md
```diff
@@ -94,6 +94,29 @@ so using [stability flags](04-schema.md#package-links). To change that for all
...
+
+If you wish to consider only the stable release in the comparison, add the
+suffix `-stable` to the version number.
+
+Here are some examples:
+
+ Example         | Interpretation
+ --------------- | --------------
+`>=1.0.0`        | Any release, stable or non-, of 1.0.0 will be allowed
+`>=1.0.0-stable` | Only the stable release of 1.0.0 will be allowed
```
stof: this looks weird. 1.0.1-beta will still match the constraint you give on the left. So the description is misleading.

alcohol: well.. my documentation PR kind of rewords this section already, maybe review that instead?

PR #4205, opened 2015-07-01
Changing config.json to satis.json

gammamatrix:
Updating docs to use `satis.json` instead of `config.json`


alcohol:
The filename can be anything though, so this change isn't really necessary?

/cc @Seldaek shouldn't satis have it's own documentation though? I imagine it's more convenient at the moment, but perhaps by moving it into the Satis repo it could also get some more (deserved) attention?

PR #4208, opened 2015-07-02
Updated SPDX related files in light of 2.0 release of SPDX specification

alcohol:
Not sure if desired or not, but figured it was time for an update.

* [x] make update binary executable
* [x] update spdx updater
* [x] update SpdxLicense class to support new specification.
* [x] add "with" exceptions list (http://spdx.org/licenses/exceptions-index.html)

### DIFF
--- a/src/Composer/Util/SpdxLicense.php
+++ b/src/Composer/Util/SpdxLicense.php
@@ -12,8 +12,6 @@
 
 namespace Composer\Util;
 
-use Composer\Json\JsonFile;
-
 /**
  * Supports composer array and SPDX tag notation for disjunctive/conjunctive
  * licenses.
@@ -22,26 +20,16 @@ use Composer\Json\JsonFile;
  */
 class SpdxLicense
 {
-    /**
-     * @var array
-     */
+    /** @var array */
     private $licenses;
+
+    /** @var array */
+    private $exceptions;
 
     public function __construct()
     {
         $this->loadLicenses();
-    }
-
-    private function loadLicenses()
-    {
-        if (is_array($this->licenses)) {
-            return $this->licenses;
-        }
-
-        $jsonFile = new JsonFile(__DIR__ . '/../../../res/spdx-licenses.json');
-        $this->licenses = $jsonFile->read();
-
-        return $this->licenses;
+        $this->loadExceptions();
     }
 
     /**
@@ -58,17 +46,34 @@ class SpdxLicense
         }
 
         $license = $this->licenses[$identifier];
-
-        // add URL for the license text (it's not included in the json)
-        $license[2] = 'http://spdx.org/licenses/' . $identifier . '#licenseText';
+        $license[] = 'http://spdx.org/licenses/' . $identifier . '.html#licenseText';
 
         return $license;
     }
 
     /**
-     * Returns the short identifier of a license by full name.
+     * Returns license exception metadata by license exception identifier.
      *
      * @param string $identifier
+     *
+     * @return array|null
+     */
+    public function getExceptionByIdentifier($identifier)
+    {
+        if (!isset($this->exceptions[$identifier])) {
+            return;
+        }
+
+        $license = $this->exceptions[$identifier];
+        $license[] = 'http://spdx.org/licenses/' . $identifier . '.html#licenseExceptionText';
+
+        return $license;
+    }
+
+    /**
+     * Returns the short identifier of a license (exception) by full name.
+     *
+     * @param string $name
      *
      * @return string
      */
@@ -79,10 +84,18 @@ class SpdxLicense
                 return $identifier;
             }
         }
+
+        foreach ($this->exceptions as $identifier => $licenseData) {
+            if ($licenseData[0] === $name) { // key 0 = fullname
+                return $identifier;
+            }
+        }
     }
 
     /**
      * Returns the OSI Approved status for a license by identifier.
+     *
+     * @param string $identifier
      *
      * @return bool
      */
@@ -106,6 +119,20 @@ class SpdxLicense
     }
 
     /**
+     * Check, if the identifier for a exception is valid.
+     *
+     * @param string $identifier
+     *
+     * @return bool
+     */
+    private function isValidExceptionIdentifier($identifier)
+    {
+        $identifiers = array_keys($this->exceptions);
+
+        return in_array($identifier, $identifiers);
+    }
+
+    /**
      * @param array|string $license
      *
      * @return bool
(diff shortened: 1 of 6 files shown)


alcohol:
@Seldaek or @stof ; where could I find some details regarding `isValidLicenseString()` from the `SpdxLicense` class? It's a bit.. cryptic (for me at least).

stof:
@alcohol this method is parsing strings like ``(LGPL-2.0 or GPL-3.0+)``. the cryptic thing is a lexer and a parser

stof:
and failing tests are because some of the licenses used by the tests are now removed from the SPDX list (see the diff on the JSON file)

alcohol:
Yeah I figured parts of it out.

I was thinking, do we still want to add the deprecated licenses to the current list and add a `deprecated` attribute?

I fixed the tests btw but it doesn't support the new specification fully yet, need to add additional tests for that.

stof:
I would say no. We should rather support the new SPDX way to provide such licenses.
Using ``GPL-3.0+`` as a project license is not deprecated. The way to consider it changed.

alcohol:
Btw the notation of `(A or|and B)`, is that defined by SPDX or something Composer adds on top?

stof:
I don't remember whether it was part of SPDX 1 or no. But it is part of SPDX 2 for sure (look at the spec for license expressions, and you will see it covers them)

src/Composer/Util/SpdxLicense.php
```diff
@@ -220,6 +221,7 @@ private function isValidLicenseString($license)
                     $require = 0;
                     break;
                 case 'ws':
+                case 'pl':
```
stof: pl should not be allowed anywhere. It can only follow a license identifier

alcohol:
Just a thought; would you guys be at all opposed to the idea of extracting the SPDX related stuff into package by itself?

src/Composer/Util/SpdxLicense.php
```diff
@@ -144,8 +144,9 @@ private function isValidLicenseString($license)
             'op' => '(?:or|and)',
             'lix' => '(?:NONE|NOASSERTION)',
             'lir' => 'LicenseRef-\d+',
-            'lic' => '[-+_.a-zA-Z0-9]{3,}',
+            'lic' => '[-_.a-zA-Z0-9]{3,}',
```
Seldaek: Maybe just having the `\+?` here at the end would be fine and then rtrim it in isValidLicenseIdentifier?

stof:
@alcohol I thought about doing exactly that when reviewing this PR

alcohol:
This covers it all and makes it relatively easy to extract. However, it's not incredibly elegant.. so yeah, feedback/thoughts/suggestions?

src/Composer/Util/SpdxLicense.php
```diff
@@ -131,20 +161,51 @@ public function validate($license)
...
-     * @throws \RuntimeException
+     * @throws RuntimeException
      */
     private function isValidLicenseString($license)
     {
         $tokens = array(
             'po' => '\(',
             'pc' => '\)',
-            'op' => '(?:or|and)',
+            'op' => '(?:OR|AND)',
```
stof: I suggest allowing the lower case versions too, to keep BC

src/Composer/Util/SpdxLicense.php
```diff
@@ -12,7 +12,7 @@
 
 namespace Composer\Util;
 
-use Composer\Json\JsonFile;
+use RuntimeException;
```
stof: adding use statements for classes in the global namespace is not consistent with the Composer codebase

src/Composer/Util/SpdxLicense.php
```diff
@@ -171,29 +232,32 @@ private function isValidLicenseString($license)
...
 
-            throw new \RuntimeException('At least the last pattern needs to match, but it did not (dot-match-all is missing?).');
+            throw new RuntimeException(
+                'At least the last pattern needs to match, but it did not (dot-match-all is missing?).'
+            );
         };
 
         $open = 0;
         $require = 1;
+        $with = 0;
```
stof: using a real boolean would be better than using ``0`` or ``1``

src/Composer/Util/SpdxLicense.php
alcohol: Well the tests don't allow mixed casing. So just using a case insensitive modifier wouldn't work. It would then be `or|OR|and|AND`. But the specs explicitly only use uppercase. You sure you want to keep BC for this?

src/Composer/Util/SpdxLicense.php
alcohol: Oh, my bad. I always like knowing what other classes are used, whether they're from the global namespace or not. I'll revert this.

Seldaek:
Looks good to me, are you done done @alcohol? :)

alcohol:
Yeah the regex is too complicated. I'll save it for a rainy day.

```
(?(DEFINE)
    # idstring: 1*( ALPHA / DIGIT / "-" / "." )
    (?<idstring>[\pL\pN\-\.]{1,})

    # license-id / license-exception-id: not really specified
    (?<licenseid>[a-zA-Z0-9\-\_\.]{3,})

    # license-ref: ["DocumentRef-"1*(idstring)":"]"LicenseRef-"1*(idstring)
    (?<licenseref>(?:DocumentRef-(?&idstring):)?LicenseRef-(?&idstring))

    # simple-expresssion: license-id / license-id”+” / license-ref
    (?<simple_expression>
        (?:(?&licenseid)\+?) | (?&licenseid) | (?&licenseref)
    )

    # compound expression: 1*(
    #   simple-expression /
    #   simple-expression "WITH" license-exception-id /
    #   compound-expression “AND" compound-expression /
    #   compound-expression “OR" compound-expression
    # ) / “(“ compound-expression “)” )
    (?<compound_expression>
        (?:
            (?&simple_expression) \s+ (?:with|WITH) \s+ (?&licenseid) |
            (?&compound_expression) \s+ (?:and|AND) \s+ (?&compound_expression) |
            (?&compound_expression) \s+ (?:or|OR) \s+ (?&compound_expression) |
            (?&simple_expression)
        ) | \( \s* (?&compound_expression) \s* \)
    )

) # end of define

# license-expression: 1*1(simple-expression / compound-expression)
^(?<license_expression>(?&compound_expression) | (?&simple_expression))$
```

tests/Composer/Test/Util/SpdxLicenseTest.php
```diff
@@ -51,19 +58,23 @@ public static function provideInvalidLicenses()
...
-            array("(MIT)"),
+            array("(MIT"),
+            array("MIT)"),
             array("MIT NONE"),
+            array("MIT AND NONE"),
             array("MIT (MIT and MIT)"),
             array("(MIT and MIT) MIT"),
             array(array("LGPL-2.0", "The system pwns you")),
             array("and GPL-3.0+"),
-            array("EUDatagrid and GPL-3.0+"),
```
stof: is this one actually valid now ?

alcohol: It always was valid actually. I don't know why it was not marked as valid.

alcohol: The regex does match a string such as "MIT AND MIT" though. So I guess it's not good enough :-/

alcohol:
Reverted it so it stays available in history, if someone ever wants to revisit it.

alcohol:
Not sure why "EUDatagrid and GPL-3.0+" is considered invalid now. Going to investigate that.

alcohol:
Rebased it for now.

/cc @Seldaek Ready to merge and then extract to self contained package.

Seldaek:
Cool thanks

Issue #183, opened 2012-01-05
Consider adding basic inheritance support for composer.json


schmittjoh:
Opening at Jordi's request as discussed on FOSRestBundle.

This would come in handy for testing a bundle against multiple different vendors without duplicating the entire file.

``` json
// composer.json
{
    require: { 
        "symfony/symfony": "2.*", 
        // ... more dependencies
    }
}
```

``` json
// tests/composer-2.0.json
{
    inherit: "../composer.json",
    require: { "symfony/symfony": "2.0.*" }
}
```

``` json
// tests/composer-2.1.json
{
    inherit: "../composer.json",
    require: { "symfony/symfony": "2.1.*" }
}
```

nadar:
+1

maximilianredt:
+1

bgallagher:
:+1:

RobinRadic:
+1 i would use this a lot during development

legoktm:
For MediaWiki we have developed a [composer-merge-plugin](https://github.com/wikimedia/composer-merge-plugin) to add this feature.

The core MediaWiki repo has a composer.json file which includes a composer.local.json file if it exists. Since some MediaWiki extensions also have dependencies installed via composer, a sysadmin would source those composer.json files in their composer.local.json file, and running "composer update" would fetch everything.

marcegarba:
+1

Project inheritance (similar to what Maven does) would be wonderful.

gpupo:
+1

Zoddo:
+1

onlab:
+1

maechler:
+1

cansik:
+1

johnitvn:
+1

masasron:
+1

secondtruth:
This is extremely useful for the use case that @gggeek described. Contrary to @schmittjoh's suggestion, for this use case a `include(s)` field should be preferred ([see here](https://github.com/composer/composer/issues/2480)). A structure like the one that the [composer-merge-plugin](https://github.com/wikimedia/composer-merge-plugin) uses, seems very good.

Another project that, I think, would like to use this feature is [Roundcube Webmail](https://github.com/roundcube/roundcubemail). They use Composer for their [plugin system](https://plugins.roundcube.net/). And I prefer this solution for [my own framework-like project](https://github.com/FlameCore/Infernum), too.

aequasi:
+1

haydenk:
+1

[...]

PR #4210, opened 2015-07-02
add support for config inheritance

aripringle:
Refs #183. This is a refactoring of #1013, with attempts to address the comments in that thread.

All credit goes to @1stvamp for the original code.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -730,7 +730,7 @@ override packages from it.
 
 ### config <span>([root-only](04-schema.md#root-package))</span>
 
-A set of configuration options. It is only used for projects. See 
+A set of configuration options. It is only used for projects. See
 [Config](06-config.md) for a description of each individual option.
 
 ### scripts <span>([root-only](04-schema.md#root-package))</span>
@@ -791,7 +791,7 @@ Optional.
 
 ### non-feature-branches
 
-A list of regex patterns of branch names that are non-numeric (e.g. "latest" or something), 
+A list of regex patterns of branch names that are non-numeric (e.g. "latest" or something),
 that will NOT be handled as feature branches. This is an array of strings.
 
 If you have non-numeric branch names, for example like "latest", "current", "latest-stable"
@@ -821,6 +821,19 @@ If you configure `latest-.*` as a pattern for non-feature-branches like this:
 ```
 
 Then "composer show -s" will give you `versions : * dev-latest-testing`.
+
+Optional.
+
+### extends
+
+Defines a parent JSON configuration filepath to a JSON file that this configuration extends.
+
+All properties from the parent JSON file are loaded and any property defined in the current
+file override (replace) those properties. This is done recursively so if the parent JSON file
+defines a parent that will be loaded (with any properties of it are also replaced) and so on.
+
+Useful for creating environment specific configurations, e.g. `testing.json`, `production.json`
+etc. which may need to override properties like `repositories`.
 
 Optional.
 
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -828,12 +828,18 @@ Optional.
 
 Defines a parent JSON configuration filepath to a JSON file that this configuration extends.
 
-All properties from the parent JSON file are loaded and any property defined in the current
-file override (replace) those properties. This is done recursively so if the parent JSON file
-defines a parent that will be loaded (with any properties of it are also replaced) and so on.
+All settings from the parent JSON file are loaded, and any settings defined in the current
+file will be merged into those settings.
+
+If the setting is defined as a JSON object, new properties will be appended to the object,
+whereas existing properties will be overwritten. In other cases (arrays and strings), the
+child value will simply overwrite the parent value.
+
+This is done recursively so if the parent JSON file defines a parent, that will be loaded
+first, merging the settings from its child, and so on.
 
 Useful for creating environment specific configurations, e.g. `testing.json`, `production.json`
-etc. which may need to override properties like `repositories`.
+etc. which may need to extend settings like `repositories`.
 
 Optional.
 
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -824,7 +824,7 @@ Then "composer show -s" will give you `versions : * dev-latest-testing`.
 
 Optional.
 
-### extends
+### extends <span>([root-only](04-schema.md#root-package))</span>
 
 Defines a parent JSON configuration filepath to a JSON file that this configuration extends.
 
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -434,6 +434,10 @@
             "items": {
                 "type": "string"
             }
+        },
+        "extends": {
+           "type": "string",
+            "description": "Relative or absolute path to the parent JSON file this file extends."
         }
     }
 }
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -219,7 +219,7 @@ class Factory
             }
 
             $file->validateSchema(JsonFile::LAX_SCHEMA);
-            $localConfig = $file->read();
+            $localConfig = static::readConfig($file);
         }
 
         // Load config and override with local config/auth config
@@ -311,6 +311,31 @@ class Factory
         }
 
         return $composer;
+    }
+
+    public static function readConfig($file, $mergeConfig = array())
+    {
+        $config = $file->read();
+
+        if (!empty($mergeConfig)) {
+            foreach ($mergeConfig as $key => $val) {
+                if (is_array($val) && array_key_exists($key, $config) && is_array($config[$key])) {
+                    // If dealing with arrays, attempt to array_replace into all first-level keys.
+                    $config[$key] = array_replace($config[$key], $val);
+                } else {
+                    // Otherwise, the values in $mergeConfig will take precedence
+                    $config[$key] = $val;
+                }
+            }
+        }
+
+        if (!empty($config['extends'])) {
+            $parent = new JsonFile($config['extends']);
+            unset($config['extends']);
+            $config = static::readConfig($parent, $config);
+        }
+
+        return $config;
     }
 
     /**
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -318,12 +318,17 @@ class Factory
         $config = $file->read();
 
         if (!empty($mergeConfig)) {
+            // If the first-level key is an associative array (JSON object), use
+            // array_replace to merge into it. Otherwise, the value in $mergeConfig
+            // will overwrite the parent values.
             foreach ($mergeConfig as $key => $val) {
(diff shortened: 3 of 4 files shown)


alcohol:
It should probably be a root-only attribute (marked as such in documentation).

far-blue:
+1

doc/04-schema.md
```diff
@@ -824,4 +824,17 @@ Then "composer show -s" will give you `versions : * dev-latest-testing`.
 
 Optional.
 
+### extends <span>([root-only](04-schema.md#root-package))</span>
+
+Defines a parent JSON configuration filepath to a JSON file that this configuration extends.
+
+All properties from the parent JSON file are loaded and any property defined in the current
+file override (replace) those properties. This is done recursively so if the parent JSON file
+defines a parent that will be loaded (with any properties of it are also replaced) and so on.
```
cmfcmf: Do I understand correctly? It is *not* possible to only add some dependencies inside the child `composer.json` file, right? I.e. the parent requires `symfony/symfony` and the child requires `everything from parent` + `a/dependency`.

aripringle: > I.e. the parent requires symfony/symfony and the child requires everything from parent + a/dependency.

This is exactly how it works. I can try to clarify this in the documentation some.

Every first-level key in the composer.json file will have `array_replace` performed against it, effectively meaning that  keys from the child's `require` block will be appended to the parent's `require` block (keys of the same name will be overwritten).

While the child can overwrite values in the parent, you can't currently remove a parent value (you could technically try to set `symfony/symfony: null` or similar, but I don't think composer will treat that well).

cmfcmf: I was confused by 
> override (replace) those properties

It doesn't sound like extending to me.

src/Composer/Factory.php
```diff
@@ -313,6 +313,31 @@ public function createComposer(IOInterface $io, $localConfig = null, $disablePlu
         return $composer;
     }
 
+    public static function readConfig($file, $mergeConfig = array())
```
alcohol: Why static by the way?

aripringle: It appeared that the calling method (createComposer) is sometimes called statically, and other Factory method calls are already being called statically from within that method.

doc/04-schema.md
1stvamp: Alternatively you could use `inherits` to describe this functionality (`extends` was borrowed from `zc.buildout` terminology when I wrote the original PR).

aripringle:
I wanted to explain my most recent commits, and open it up for discussion.

Anything defined as an object in composer.json will be "extended", meaning that new keys will be appended to the object, and existing keys will be overwritten (using `array_replace`). For example:

parent.json:
```json
{
  "require": {
    "php": ">=5.4.0",
    "monolog/monolog": "~1.0"
  }
}
```

child.json:
```json
{
  "require": {
    "monolog/monolog": "dev-master",
    "phpunit/phpunit": "dev-master"
  },
  "extends": "parent.json"
}
```

results in:
```json
{
  "require": {
    "php": ">=5.4.0",
    "monolog/monolog": "dev-master",
    "phpunit/phpunit": "dev-master"
  }
}
```

However, for numerically-indexed arrays, I opted to have the child values (if present) overwrite the parent values completely. I could opt to use `array_merge` here, always appending the child values to the parent, but this makes it impossible to override anything in the parent array (and could also allow for duplicate values in the array).

Any thoughts on the matter are welcome.

ccoley:
This feature would be very useful. Any timeline on when it will be merged?

czernitko:
An example of numerically indexed array is a list of repositories. We would like to use `extends` feature to add some customized bundles to Symfony project. Parent config file will be shared amongst all project installations, but for some installations we need to install additional customized bundles.
In such case, in child config file, we need to 
 - define new "require" item for custom bundle and 
 - add a new "repositories" record. 
If child values overwrote the parent values, it would mean that we have to copy parent values into child config file and add the new repositories. This basically means that we are not **extending**, but **replacing** the parent values. Therefore I would definitely vote for using `array_merge` to combine parent and child values in case of numerically indexed arrays.

I'm also very interested in an answer to ccoley's question :)

src/Composer/Factory.php
Taluu: I don't see the point of creating a new method, as it is only called once...

src/Composer/Factory.php
```diff
@@ -313,6 +313,36 @@ public function createComposer(IOInterface $io, $localConfig = null, $disablePlu
...
+                if (is_array($val)
+                    && array_key_exists($key, $config)
+                    && is_array($config[$key])
+                    && array_values($val) !== $val
+                    ) {
+                    $config[$key] = array_replace($config[$key], $val);
+                } else {
+                    $config[$key] = $val;
+                }
+            }
```
Taluu: isn't that basically the behaviour of `array_replace_recursive` ?

src/Composer/Factory.php
```diff
@@ -313,6 +313,36 @@ public function createComposer(IOInterface $io, $localConfig = null, $disablePlu
         return $composer;
     }
 
+    public static function readConfig($file, $mergeConfig = array())
+    {
+        $config = $file->read();
+
+        if (!empty($mergeConfig)) {
```
Taluu: no need to test if `mergeConfig` is empty, the foreach will not iterate over it (as it is already empty)

src/Composer/Factory.php
Taluu: nevermind, didn't see the recursive call

src/Composer/Factory.php
```diff
@@ -313,6 +313,36 @@ public function createComposer(IOInterface $io, $localConfig = null, $disablePlu
...
+                    ) {
+                    $config[$key] = array_replace($config[$key], $val);
+                } else {
+                    $config[$key] = $val;
+                }
+            }
+        }
+
+        if (!empty($config['extends'])) {
+            $parent = new JsonFile($config['extends']);
```
Taluu: I dont think this is the correct path... If you input a `../parent.json`, wouldn't it load the file from the current directory (which is `/.../src/Composer`) ?

alcohol: No this is not really the current directory. Remember everything ends up wrapped in a Phar. It would be relative to the working directory of where you are running `composer`.

stof: @alcohol it will be resolved related to the current working directory (not the directory of this file), as any relative path in PHP. And this is what would be really confusing. A relative path in a composer.json should be resolved relatively to the composer.json file.

Taluu: And what if the path is not relative, or even on a distant remote filesystem ?

src/Composer/Factory.php
aripringle: Almost, except the "else" here will also catch numerically-indexed arrays. If I do an array_replace in those, it would only replace matching keys, which could lead to strange-seeming behavior. For example:

```
"repositories": [
        { 
            // repository A...
        },
        { 
            // repository B...
        }
]
```

Extended with:

```
"repositories": [
        { 
            // repository C...
        }
]
```

If we use array_replace_recursive on this, we would end up with:

```
"repositories": [
        { 
            // repository C... (0th index was replaced)
        },
        { 
            // repository B...
        }
]
```

Instead, I opted to completely replace numerically-indexed arrays (this behavior is currently under discussion).

aripringle:
@czernitko I was originally thinking that there might be a use case might that would require repositories to be replaced instead of extended. Without a more verbose syntax (such as allowing the user to specify which properties get extended), I thought this approach was a little more flexible.

However, after considering it a little more, extending the arrays as you're suggesting may be fine. For the repositories example specifically, it would probably make sense to `array_merge($childArray, $parentArray)`. This would allow the child repositories to take precedence over the parent repositories (in the case that they contain the same packages, for development/testing repositories for example).

manfredjb:
Really useful, hope will be merged.

composerfan8:
Hello. I am relatively new to composer, but could someone explain the rationale behind the bottom-up approach to inheritance using the 'extend' keyword, as opposed to a top-down approach ('include' in root composer.json) such as that found in composer-merge-plugin?

Looking at this from the perspective of a user with multiple subprojects as a part of a larger project, I think I must be missing something. It seems to me that requiring the 'extend' clause in these subprojects would make them less portable, where 'including' subproject composer.json would allow inheritance in the same way. This type of situation might occur in a monolithic repository situation.

Is the intent that subprojects should instead be referenced as composer packages in the main composer.json?

I may be missing the entire point here, but it would be nice to see a semi-detailed explanation of the thought process that went into this decision. Either way, it's great that progress is being made on the inheritance issue. Long live composer!

res/composer-schema.json
```diff
@@ -434,6 +434,10 @@
             "items": {
                 "type": "string"
             }
+        },
+        "extends": {
+           "type": "string",
```
dashdanw: indentation?

andrerom:
Cool!

Important step towards fixing distribution problems with composer based applications  *(maker and user both modifying same files)*. However any thoughts on seperating lock file and vendor dir as well? And should all relevant composer commands accept a -f <composer.json> like docker compose does?

alcohol:
@andrerom please explain the purpose of a separate lock file and vendor directory? How do you see this working?

andrerom:
Well, for me composer inheritance is a step towards solving a generic problem, "distribution problems". Example some symfony application, where users will be forced to migrate config, kernel, composer.json, .. and deal with conflicts as part of upgrades. This is somewhat pain, and you can easily make mistake if you're not that experienced and/or not fully awake. This is fixed by being able to have a dist version of composer.json.

As for the lock and vendor dir, this is more related to issues with distribution via docker images, however it is a separate issue and one I'm not fully able to put words on yet :) But if someone wants to discuss it, ping me on twitter or something.

BernhardPosselt:
I may add to that that this is also important for anything that acts as a platform. For instance when writing an ownCloud app, you are defining your dependencies in a composer.json in your app folder. However ownCloud itself also ships multiple vendor libs via composer. I don't want to install guzzle in my app's vendor directory for instance when its already present in the main vendor directory.

fabsalito:
In my opinion, this is a very useful feature. In my case, I work with multiple SSH identities and some clients has references in his composer.json to his own bundles in the way xxxxxx@bitbucket.org:client/yyyyy. It is complicated resolve this use case without the proposed feature.

lvidal1:
Any updates on this?

homerjam:
Yeesh, 4 years? LGTM?

Using https://github.com/wikimedia/composer-merge-plugin for now

ruudboon:
Will this ever be part of composer?

GrahamCampbell:
Is this PR dead?

Pictor13:
Tok tok 🖖🏻, wondering the same . 
Was this stall for any reason?
Is its code still compatible with new Composer's versions?

Is it potentially merge-able or are there concerns from the core team about this feature?

Seldaek:
Closing as per https://github.com/composer/composer/issues/183#issuecomment-716218186

PR #4217, opened 2015-07-06
[Feature / Tests] Add config validator test

SpacePossum:
When debugging a project I run into something I thought was odd with the configuration validating.
Turns out it was all good, but here is the unit test I'd made.

### DIFF
--- a/tests/Composer/Test/Util/ConfigValidatorTest.php
+++ b/tests/Composer/Test/Util/ConfigValidatorTest.php
@@ -1,0 +1,66 @@
+<?php
+namespace Composer\Test\Util;
+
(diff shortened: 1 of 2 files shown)


alcohol:
Your test actually doesn't pass cause the composer.json of composer contains a "_" property (used as a placeholder for a comment, sort of a "hack" as json doesn't support comments).

SpacePossum:
hmm ironic that the `composer.json` of composer is not valid according its own validator ;)
I'll take that one out of the tests.

SpacePossum:
on second thought, validation of a validator seems a bit to meta to be really needed.

PR #4219, opened 2015-07-07
Enhancement: Also validate if lock file is up to date

localheinz:
This PR

* [x] adds validation of whether `composer.lock` is up to date (if it exists), emits an error and exits with non-zero if it's not
* [x] adds a `--no-lock` option which will emit a warning instead
* [x] updates the docs

:information_desk_person: Useful if you want to break the build because someone modified `composer.json` without updating `composer.lock`.

## Example

Fiddle with `composer.json` in this repository, for example, change

```json
{
    "autoload": {
        "psr-0": { "Composer": "src/" }
    }
}
```

to 

```json
{
    "autoload": {
        "psr-0": { "Composer": "src" }
    }
}
```

### Before

```
$ ./bin/composer validate
./composer.json is valid

$ echo $?
0
```

### After

```
$  ./bin/composer validate
./composer.json is valid
The lock file is not up to date with the latest changes in composer.json.

$ echo $?
1
```

### After  (with `--no-lock` option)

```
$  ./bin/composer validate
./composer.json is valid
The lock file is not up to date with the latest changes in composer.json.

$ echo $?
0
```

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -341,8 +341,8 @@ php composer.phar validate
 ### Options
 
 * **--no-check-all:** Do not emit a warning if requirements in `composer.json` use unbound version constraints.
+* **--no-check-lock:** Do not emit an error if `composer.lock` exists and is not up to date.
 * **--no-check-publish:** Do not emit an error if `composer.json` is unsuitable for publishing as a package on Packagist but is otherwise valid.
-
 
 ## status
 
--- a/src/Composer/Command/ValidateCommand.php
+++ b/src/Composer/Command/ValidateCommand.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Command;
 
+use Composer\Factory;
 use Composer\Package\Loader\ValidatingArrayLoader;
 use Composer\Util\ConfigValidator;
 use Symfony\Component\Console\Input\InputArgument;
@@ -34,14 +35,15 @@ class ValidateCommand extends Command
     {
         $this
             ->setName('validate')
-            ->setDescription('Validates a composer.json')
+            ->setDescription('Validates a composer.json and composer.lock')
             ->setDefinition(array(
                 new InputOption('no-check-all', null, InputOption::VALUE_NONE, 'Do not make a complete validation'),
+                new InputOption('no-check-lock', null, InputOption::VALUE_NONE, 'Do not check if lock file is up to date'),
                 new InputOption('no-check-publish', null, InputOption::VALUE_NONE, 'Do not check for publish errors'),
                 new InputArgument('file', InputArgument::OPTIONAL, 'path to composer.json file', './composer.json')
             ))
             ->setHelp(<<<EOT
-The validate command validates a given composer.json
+The validate command validates a given composer.json and composer.lock
 
 EOT
             );
@@ -73,6 +75,15 @@ EOT
         $checkPublish = !$input->getOption('no-check-publish');
         list($errors, $publishErrors, $warnings) = $validator->validate($file, $checkAll);
 
+        $checkLock = !$input->getOption('no-check-lock');
+
+        $lockErrors = array();
+        $composer = Factory::create($this->getIO(), $file);
+        $locker = $composer->getLocker();
+        if ($locker->isLocked() && !$locker->isFresh()) {
+            $lockErrors[] = 'The lock file is not up to date with the latest changes in composer.json.';
+        }
+
         // output errors/warnings
         if (!$errors && !$publishErrors && !$warnings) {
             $this->getIO()->write('<info>' . $file . ' is valid</info>');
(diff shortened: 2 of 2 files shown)


localheinz:
@Seldaek 

Not sure, would you like me to provide tests for this one?

alcohol:
I'm not sure if it makes more sense to move this a command by itself, or keep it as an option/flag just for install. What if you want your build process to check first, and do a `composer update --lock` if it isn't up to date (because, reasons)? A command to check/validate the lockfile could also simply be relied on by exit code, and possibly provide some feedback if it is not in sync.

localheinz:
@alcohol 

I'm totally open for suggestions! Do you have anything in mind that would make sense?

Maybe a `VerifyCommand`, resulting in

```
$ composer verify
```

or (to leave space for other things which would be useful to verify)

```
$ composer verify --lock
```

alcohol:
The first example is too ambiguous, `verify` doesn't really describe what it does. I would sooner go for something like `verify-lock` then.

`verify --lock` looks ok too. Maybe you could also add a `--json` while you are at it then? It's pretty easy to validate the json schema. See #4217 for some inspiration.

hkdobrev:
> It's pretty easy to validate the json schema. See #4217 for some inspiration.

There's the [`composer validate`](https://getcomposer.org/doc/03-cli.md#validate) command which validates `composer.json` against the JSON config schema.

alcohol:
Ah right I forgot about that. Maybe add the `--lock` flag to that command?

localheinz:
@alcohol @hkdobrev 

Sounds great! Just give me a moment, I'm on the road, traveling.

localheinz:
@alcohol @hkdobrev 

Do you want to take a look?

alcohol:
Looks good, but I have one concern that maybe @Seldaek might be able to offer some feedback on.

Should the lock file be checked by default as well? Then the option would become `--no-check-lock` which would be more in line with the other options.

hkdobrev:
:+1: I also support @alcohol's idea about `--no-check-lock`.

You should also update CLI docs and CLI command descriptions.

localheinz:
@alcohol @hkdobrev 

This is the route I was going down as well, but then figured it would break people's builds if they were using it already, what do you think?

hkdobrev:
> it would break people's builds if they were using it already

I think this is a good thing in this case. And Composer is still in alpha so going for consistency rather than BC is better. Also I think this is rarely used at the moment.

doc/03-cli.md
```diff
@@ -340,6 +340,7 @@ php composer.phar validate
 ### Options
 
 * **--no-check-all:** Whether or not Composer does a complete validation.
+* **--no-lock:** Do not check if lock file is up to date.
```
localheinz: @alcohol @hkdobrev 

See https://github.com/composer/composer/pull/4230.

doc/03-cli.md
```diff
@@ -340,6 +340,7 @@ php composer.phar validate
 ### Options
 
 * **--no-check-all:** Whether or not Composer does a complete validation.
+* **--no-check-lock:** Do not check if lock file is up to date.
```
localheinz: @alcohol @hkdobrev 

Also see https://github.com/composer/composer/pull/4230.

src/Composer/Command/ValidateCommand.php
```diff
@@ -73,6 +74,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $checkPublish = !$input->getOption('no-check-publish');
         list($errors, $publishErrors, $warnings) = $validator->validate($file, $checkAll);
 
+        $lockErrors = array();
+
+        $checkLock = !$input->getOption('no-check-lock');
+        if ($checkLock) {
+            $locker = $this->getComposer()->getLocker();
+            if ($locker->isLocked() && !$locker->isFresh()) {
```
localheinz: @alcohol @hkdobrev 

Updated the condition here so validation doesn't unexpectedly fail when there's no lock file around.

alcohol: Looks good to me (but on phone right now). Cheers.

alcohol:
@localheinz: You need to update it, @naderman introduced a merge conflict.

localheinz:
@alcohol 

Done!

src/Composer/Command/ValidateCommand.php
```diff
@@ -37,6 +37,7 @@ protected function configure()
             ->setDescription('Validates a composer.json')
```
hkdobrev: The `composer validate` command no longer validates `composer.json` only.

localheinz:
@hkdobrev 

Updated!

@alcohol @hkdobrev 

What do you think, should this also work when specifying the `file` argument, that is, attempt to validate a lock file in the same directory as where the specified file has been found?

alcohol:
Yeah that probably makes sense. If they just want to validate a json schema they can skip the lock check. Otherwise, check for a lock file in the same directory.

localheinz:
@alcohol 

Amended the PR, please have a look!

alcohol:
Looks good to me.

/cc @Seldaek do you feel this might be too much of a BC change or acceptable?

Seldaek:
Looks good to me, thanks! I doubt it will create issues but perhaps more people use this in build systems than I imagine.

localheinz:
@alcohol @hkdobrev @Seldaek 

Thank you for reviewing and merging, I appreciate it!

PR #4220, opened 2015-07-07
[WIP] Generate faster code

SpacePossum:
This changes makes composer generate slightly faster code;

* remove concat of two string where possible, for example: 
```
-require_once __DIR__ . '/composer' . '/autoload_real.php';		
+require_once __DIR__ . '/composer/autoload_real.php';
```
* use `__DIR__` over ```dirname(__FILE__)``` if >=PHP 5.3.0 is set as requirement
```
-$vendorDir = dirname(dirname(__FILE__));		
+$vendorDir = dirname(__DIR__);
```


staabm:
lots of changes for nearly no benefit. therefore :-1:

SpacePossum:
Besides the tests and PHPDoc fixes the amount of changes is only a few lines, so I don't see the problem.

naderman:
Composer requires 5.3.2+ so not sure what all of that 5.3.0+ stuff is about?

SpacePossum:
>Composer requires 5.3.2+ so not sure what all of that 5.3.0+ stuff is about?

It is about the requirement of the project composer is generating code for, not about composers requirement itself.

alcohol:
@SpacePossum please just assume you will always have 5.3.2 as a minimum available. So you can just remove all of those redundant checks.

SpacePossum:
@alcohol are you sure? This is not about the PHP version composer requires nor is it the PHP version this code is executed on. It is about the PHP version one has set as minimum requirement for the project composer is generation code for. If I would assume everyone using composer only does this for projects with a minimum PHP version of 5.3.2 then it would be a big BC break wouldn't it?

alcohol:
5.3.x is end of life.. you are worried about people running 5.2 or lower? I
think those poor souls have bigger issues to worry about.

SpacePossum:
>5.3.x is end of life.. you are worried about people running 5.2 or lower? I
think those poor souls have bigger issues to worry about. 

:+1:   :)   
I'll update this PR soon

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -464,12 +489,19 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
 
     protected function getAutoloadFile($vendorPathToTargetDirCode, $suffix)
     {
+        $lastChar = $vendorPathToTargetDirCode[strlen($vendorPathToTargetDirCode)-1];
+        if ("'" === $lastChar || '"' === $lastChar) {
+            $vendorPathToTargetDirCode = substr($vendorPathToTargetDirCode, 0, -1).DIRECTORY_SEPARATOR.'autoload_real.php'.$lastChar;
```
stof: you can just use ``/`` all the time (it will simplify testing btw as the test assertion will not have to change on Windows)

PR #4223, opened 2015-07-07
Add missing PHPDoc, change visibility

emulienfou:
2 things have been done here:

1. Add missing PHPDoc into `Composer/Installer` class
2. Properties and Methods visibilities as been change from `private` to `protected`.

If people are in my case and want to overwrite some classes methods, you will need to duplicate some codes because a lot's of methods and properties as `private` visibilities.

### DIFF
--- a/src/Composer/Compiler.php
+++ b/src/Composer/Compiler.php
@@ -152,7 +152,7 @@ class Compiler
         $util->save($pharFile, \Phar::SHA1);
     }
 
-    private function addFile($phar, $file, $strip = true)
+    protected function addFile($phar, $file, $strip = true)
     {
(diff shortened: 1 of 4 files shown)


alcohol:
I think these methods are marked private because Composer intentionally does not want to expose the internal API. By exposing it, we commit to maintaining backwards compatibility or we have to up major versions every time something does change.

stof:
I agree with @alcohol here. The visibility change should be reverted. There is a big difference between private and protected regarding the work for maintainers (especially when we don't expect the implementation to be finalized yet)

Seldaek:
I cherry picked the good commits

PR #4224, opened 2015-07-07
Replace class mapper with Automap

flaupretre:
This PR replaces the original classMapper with Automap. [Automap](http://automap.tekwire.net) is a highly-optimized map-based autoloader. It features specific optimizations to avoid issues typically met with other map-based autoloaders (such as the ridiculously-long time taken to load large symbol maps).

Even if the pure-PHP version of Automap is already highly optimized, an optional PECL extension still increases performance by caching symbol maps in persistent memory, providing near-zero map-loading overhead.

This PR is a proof-of-concept, implementation may be discussed. I am releasing it here so that people can validate performance gains on real applications.

You may use this branch or directly get the corresponding composer.phar [from here](http://automap.tekwire.net/files/composer.phar).

The optional PECL extension is named 'phk'. It is available [from the PECL repository](http://pecl.php.net/package/phk) and is not available for PHP 7 yet (migration in progress). Once installed and active, it is used automatically. When the extension is installed, you should see a dramatic increase in performance. Of course, performance tests must be performed on web requests, as map caching is useless and inactive in CLI mode.

The ideal test should contain 4 measures : 

* using the original 'PSR' autoloader (original 'dumpautoload'),
* using the original classMapper (original 'dumpautoload -o'),
* using Automap in PHP-only mode (new 'dumpautoload -o' without the phk extension),
* and using Automap with its optional extension active.

TODO: Modify unit tests

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -270,6 +270,13 @@ EOF;
         fclose($targetLoader);
         unset($sourceLoader, $targetLoader);
 
+        $sourceLoader = fopen(__DIR__.'/Automap/Map.php', 'r');
+        $targetLoader = fopen($targetDir.'/AutomapMap.php', 'w+');
+        stream_copy_to_stream($sourceLoader, $targetLoader);
+        fclose($sourceLoader);
+        fclose($targetLoader);
+        unset($sourceLoader, $targetLoader);
+
         $this->eventDispatcher->dispatchScript(ScriptEvents::POST_AUTOLOAD_DUMP, $this->devMode, array(), array(
             'optimize' => (bool) $scanPsr0Packages,
         ));
@@ -544,10 +551,8 @@ PSR4;
 
         if ($useClassMap) {
             $file .= <<<'CLASSMAP'
-        $classMap = require __DIR__ . '/autoload_classmap.php';
-        if ($classMap) {
-            $loader->addClassMap($classMap);
-        }
+        $classMapFile = __DIR__ . '/autoload_classmap.php';
+        $loader->addClassMapFile($classMapFile);
 
 
 CLASSMAP;
--- a/src/Composer/Autoload/Automap/Map.php
+++ b/src/Composer/Autoload/Automap/Map.php
@@ -1,0 +1,501 @@
+<?php
+//=============================================================================
+//
+// Copyright Francois Laupretre <automap@tekwire.net>
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//       http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+//
+//=============================================================================
+/**
+* @copyright Francois Laupretre <automap@tekwire.net>
+* @license http://www.apache.org/licenses/LICENSE-2.0 Apache License, V 2.0
+* @category Automap
+* @package Automap
+*///==========================================================================
+
+//=============================================================================
+/**
+* A map instance (created from an existing map file)
+*
+* When the PECL extension is not present, this class is instantiated when the
+* map is loaded, and it is used by the autoloader.
+*
+* When the extension is present, this class is instantiated only when explicitely
+* referenced and is not used by the autoloader.
+*
+* API status: Public
+* Included in the PHK PHP runtime: Yes
+* Implemented in the extension: No
+*///==========================================================================
+
+namespace Automap {
+
+if (!class_exists('Automap\Map',false)) {
+
+class Map
+{
+/** Runtime API version */
+
+const VERSION = '3.1.0';
+
+/** We cannot load maps older than this version */
+ 
+const MIN_MAP_VERSION = '3.1.0';
+
+/** Map files start with this string */
+
+const MAGIC = "AUTOMAP  M\024\x8\6\3";// Magic value for map files
+
+//--------------------------
+/** The absolute path of the map file */
+
+private $path;			
+
+/** @var array(<key> => <target>)	The symbol table */
+
+private $symbols;
+
+/** @var array(<name> => <value>)	The map options */
+
+private $options;
+
+/** @var string The version of \Automap\Build\Creator that created the map file */
+
+private $version;
+
+/** @var string The minimum runtime version needed to understand the map file */
+
+private $minVersion;
+
+/** @var integer Load flags */
+
+private $flags;
+
+/** @var string Absolute base path */
+
+private $basePath;
+
+//-----
+/**
+* Construct a map object from an existing map file (real or virtual)
+*
+* @param string $path Path of the map file to read
+* @param integer $flags Combination of Automap load flags (@see Automap)
+* @param string Reserved for internal use (PHK). Never set this.
+*/
+
+public function __construct($path,$flags=0,$_bp=null)
+{
+	//$start_time = microtime(true);//TRACE
+	$this->path = self::mkAbsolutePath($path);
+	$this->flags = $flags;
+
+	try {
+		$buf = self::getMapHeader($this->path);
+
+		//-- Check magic
+
+		if (substr($buf,9,14)!=self::MAGIC) {
+			if (strpos($buf,'$vendorDir')!==false) {
+				$this->loadComposerClassmap();
+				return;
+			} else {
+				echo "$buf\n";
+				throw new \Exception('Bad Magic');
+			}
+		}
+
+		//-- Check min runtime version required by map
+
+		$this->minVersion = trim(substr($buf,24,12));	
+		if (version_compare($this->minVersion,self::VERSION) > 0) {
+			throw new \Exception($this->path.': Cannot understand this map.'.
+				' Requires at least Automap version '.$this->minVersion);
+		}
+
+		//-- Check if the map format is not too old
+
+		$this->version = trim(substr($buf,37,12));
+		if (strlen($this->version)==0)
+			throw new \Exception('Invalid (empty) map version');
+		if (version_compare($this->version,self::MIN_MAP_VERSION) < 0)
+			throw new \Exception('Cannot understand this map. Format too old.');
+
+		//-- Check file size
+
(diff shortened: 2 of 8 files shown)


alcohol:
What does phk offer over phar? I don't see why we would want to use a custom packaging extension when php offers one natively?

flaupretre:
phk has nothing to do here. It's just that the Automap and phk accelerators are merged in the 'phk' extension for performance reasons. Here, we just use the 'automap' accelerator. We're talking about autoloading here, not packaging.

aaukt:
Couldn't you already use any autoloader by just not including `vendor/autoload.php` and instead roll your own?

As of now, composer uses the "composer-autoloader", why should it include Automap and not one of the several other autoloader implementations?

flaupretre:
@aaukt - composer provides two optional autoloaders : one is path-based (PSR-0/PSR-4), one is map-based. I just propose to replace the current map-based autoloader by a more efficient implementation.

Of course, we could consider that people use the autoloader they want and implement it externally. If there were a significant number of existing alternate autoloaders, it would be worth the pain to develop an interface so that an external autoloader can get the information it needs from a composer package, but there are almost no alternate map-based autoloading software available.

So, I preferred to integrate Automap as a transparent replacement for classMapper because it allows to keep the same user interface ('dumpautoload -o'). As Automap provides the same features as classMapper with better performance, I don't see why we should keep classMapper available.

staabm:
Do you have any numbers/benchmarks?

flaupretre:
@staabm - I don't have numbers comparing the four cases (PSR, classMapper, Automap with and without the extension). That's what I'm expecting with this PR.

staabm:
@flaupretre 

> I don't have numbers comparing the four cases (PSR, classMapper, Automap with and without the extension). That's what I'm expecting with this PR.

So how do you know that Automap classloader is the fastest classmapper then? You should have benchmarked it already in isolation of composer, shouldn't you?
Please dont take this as offence.

flaupretre:
@staabm - No problem, I understand that performance needs to be demonstrated with numbers. I have some numbers outside of composer but that's not very significant because numbers need to be compared with composer's original autoloaders.

I am currently measuring map load and symbol resolution times using a symfony application with more than 6,000 symbols ([https://github.com/stfalcon/portfolio](https://github.com/stfalcon/portfolio)). The first results I got show, as expected, that the map load overhead with the 'phk' extension is negligible (less than 0.1 ms). But results without the extension are not as good as I expected. This can be improved and I am working on it.

Seldaek:
There are quite a few implementation details that disturb me but I don't want to discuss everything in detail yet as we should first see if the PR makes sense at all.

A few thoughts:

- The automap code is kinda hard to read due to following a rather odd coding style (little indenting at all), and I am not so happy to depend on code I don't know/understand. From the little I looked though it seems to have quite a few dependencies which I don't consider needed for the job it is doing (i.e. hard-coded logging calls).
- Is it actually faster and by how much? Compared to a standard psr-4 and especially compared to an optimized build (`composer dump -o`).
- Can it be implemented as a plugin instead somehow? I would think yes, but maybe we need to add a bit more data to the events we fire to make it easier. That would allow testing the idea out a bit broader and then maybe include it if it proves to be better.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -22,6 +22,12 @@
 use Composer\Util\Filesystem;
 use Composer\Script\ScriptEvents;
 
+error_reporting(($errlevel=error_reporting()) & ~E_NOTICE);
```
stof: why disabling notices here ? If the automap runtime triggers notices when loading it, I suggest fixing it instead of forcing to disable them

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -130,11 +140,10 @@ public function dump(Config $config, InstalledRepositoryInterface $localRepo, Pa
 <?php
 
 // autoload_classmap.php @generated by Composer
+// Composer is now using Automap for map-based autoloading.
+// This file is here for BC only and always returns an empty array.
```
stof: if it is for BC, it should continue returning the map if possible, for other tools looking at this file to configure external autoloaders, even if the composer autoloader would not load the file anymore

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -261,20 +251,56 @@ public static function autoload(\$class)
...
+     *
+     * @param \Automap_Creator $map The map we are addaing symbols to
+     * @param string $mapPath The path of this map file
+     * @param \Iterator|string $paths      The path to search in or an iterator
+     * @param string           $whitelist Regex that matches against the file path
+     * @param string           $namespace Optional namespace prefix to filter by
+     *
+     * @return null
+     */
+    public static function mapScan($map, $mapPath, $paths, $whitelist = null, $namespace = null)
```
stof: why is it public ?

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -261,20 +251,56 @@ public static function autoload(\$class)
...
+
+    private static function copyFile($spath,$tpath)
+    {
+    // use stream_copy_to_stream instead of copy
+    // to work around https://bugs.php.net/bug.php?id=64634
+    $source = fopen($spath, 'r');
+    $target = fopen($tpath, 'w+');
+    stream_copy_to_stream($source, $target);
+    fclose($source);
+    fclose($target);
```
stof: please indent the code properly

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -62,11 +68,15 @@ public function dump(Config $config, InstalledRepositoryInterface $localRepo, Pa
         $basePath = $filesystem->normalizePath(realpath(getcwd()));
         $vendorPath = $filesystem->normalizePath(realpath($config->get('vendor-dir')));
         $useGlobalIncludePath = (bool) $config->get('use-include-path');
-        $prependAutoloader = $config->get('prepend-autoloader') === false ? 'false' : 'true';
+//        $prependAutoloader = $config->get('prepend-autoloader') === false ? 'false' : 'true';
+        $prependAutoloader = 'false'; // Required for Automap
```
stof: why ? This looks wrong to me. this config setting is here to give more flexibility about the way Composer registers its autoloader when it is not the only one

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -261,20 +251,56 @@ public static function autoload(\$class)
...
+            $map->registerPath($path,$rpath,$namespace,$whitelist);
+        }
+    }
+    
+    private static function getRelativePath($path,$refpath)
+    {
+        $path=rtrim(strtr($path, '\\', '/'),'/');
+        $refpath=rtrim(strtr($refpath, '\\', '/'),'/');
+        if ($path===$refpath) return '.';
+        if (strpos($path.'/',$refpath.'/')===0) return substr($path,strlen($refpath)+1);
```
stof: please respect the PSR-2 coding standards

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -261,20 +251,56 @@ public static function autoload(\$class)
...
+        return $path;
+    }
+
+    private static function copyFile($spath,$tpath)
+    {
+    // use stream_copy_to_stream instead of copy
+    // to work around https://bugs.php.net/bug.php?id=64634
```
stof: this does not seem needed. We don't have any stream wrapper involved here

stof:
what is this automap.phk binary file ?

alcohol:
@stof it's basically a userland implementation of phar, but different.

stof:
including a phar inside a phar looks weird to me. Couldn't automap be installed normally as a composer dependency of composer, like other dependencies ? It would be much simpler IMO.

stof:
hmm, no. The automap code is also copied for the runtime.

It would be great to distinguish the part needed at runtime from the part needed only to generate the map btw, to include only the actual runtime code at runtime rather than the whole Automap library.
I'm not a fan of the current state of this library (not compatible with Composer installations, not playing well with other autoloaders apparently according to the changes done to the composer registration, and not nice to output or small builds)

flaupretre:
@stof - Thanks for your comments and suggestions. I am currently working on a new version that will solve most, if not all, your concerns, including PSR coding style, generating a compatible autoload_classmap.php, etc. Installing Automap as a composer dependency is an option too.

flaupretre:
@seldaek - Here is a less intrusive version, using the Automap runtime only when the PHK extension is present. When the PHK extension is not present, it still uses the original ClassMapper because, in this case, it is slightly faster.

Actually, compared to [Patrick Allaert's benchmarks](http://patrickallaert.blogspot.fr/2013/03/benchmarking-composer-autoloading.html), measured times are greatly modified when using the opcache extension. In particular, loading a huge array using a PHP script becomes faster than unserializing it.

Here are some numbers I got using a composer application with ~ 7,000 classes. I measured the map loading time and the loading of 100 random classes (times in ms) : 

<table border="1">
  <tr>
    <th>Autoloader</th>
    <th>Init time<br>
      (incl. PSR)</th>
    <th>Init time<br>
      (excl. PSR)</th>
    <th>Time to load<br>
      100 classes</th>
    <th>Request (Init, incl<br>
      &nbsp;PSR + loading<br>
      &nbsp;100 classes)</th>
    <th>Time compared<br>
      to PSR-only </th>
  </tr>
  <tr>
    <td>PSR-only (dumpautoload)</td>
    <td>0.3</td>
    <td bgcolor="#C0C0C0">&nbsp;</td>
    <td>2</td>
    <td>2.3</td>
    <td>100 %</td>
  </tr>
  <tr>
    <td>ClassMapper ('dumpautoload-o'<br>
      without PHK extension)</td>
    <td>2.3</td>
    <td>2</td>
    <td>1.3</td>
    <td>3.6</td>
    <td>+ 56 %</td>
  </tr>
  <tr>
    <td>Automap ('dumpautoload -o'<br>
      with PHK extension)</td>
    <td>0.3</td>
    <td>0 (&lt; 0.1)</td>
    <td>1.3</td>
    <td>1.6</td>
    <td>- 31 %</td>
  </tr>
</table>

As expected, these results show a great dependency on the numbers of classes loaded during the request. Anyway, while the ClassMapper may be slower than PSR, we see that Automap is always faster. Its speed is about the same as ClassMapper without the overhead due to map load.

We can also see that results can be still better if we disable the PSR autoloader (I didn't check this), making Automap almost 3 times faster than ClassMapper (for this example of 100 classes).

These are only artificial, theoritical results. Every results measured on real-world applications are more than welcome, because we cannot state on the usefulness of this addition without a significant panel of real-world numbers.

staabm:
This should also be compared against https://github.com/composer/composer/pull/3802

PR #4228, opened 2015-07-08
Remove unnecessary version comparison cache variable and store operator as int

naderman:
We construct a huge amount of VersionConstraint objects, storing the operator as an int can make a difference of multiple MB of memory usage. The comparison of version constraints is very fast, while the memory used for the static function cache can use a significant amount of the peak, so it makes no sense to cache these values. Accessing that much memory may be slower than the actual computation.

Sample projects with composer cache filled `update -v --dry-run --profile` (line 1 before, line 2 after):

Composer:
```
[67.0MB/2.71s] Memory usage: 67.03MB (peak: 95.82MB), time: 2.71s
[65.6MB/1.81s] Memory usage: 65.6MB (peak: 94.33MB), time: 1.81s
```

Packagist:
```
[157.0MB/10.02s] Memory usage: 156.99MB (peak: 429.76MB), time: 10.02s
[150.2MB/10.26s] Memory usage: 150.24MB (peak: 423.07MB), time: 10.26s
```

Symfony Standard:
```
[148.2MB/7.13s] Memory usage: 148.24MB (peak: 328.82MB), time: 7.13s
[143.6MB/7.13s] Memory usage: 143.58MB (peak: 324.22MB), time: 7.13s
```

### DIFF
--- a/src/Composer/Package/LinkConstraint/VersionConstraint.php
+++ b/src/Composer/Package/LinkConstraint/VersionConstraint.php
@@ -28,7 +28,7 @@ class VersionConstraint extends SpecificConstraint
     const OP_GE = 4;
     const OP_NE = 5;
 
-    static $transOpStr = array(
+    private static $transOpStr = array(
         '='  => self::OP_EQ,
         '==' => self::OP_EQ,
         '<'  => self::OP_LT,
@@ -39,7 +39,7 @@ class VersionConstraint extends SpecificConstraint
         '!=' => self::OP_NE,
     );
 
-    static $transOpInt = array(
+    private static $transOpInt = array(
         self::OP_EQ => '==',
         self::OP_LT => '<',
         self::OP_LE => '<=',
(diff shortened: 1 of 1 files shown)


src/Composer/Package/LinkConstraint/VersionConstraint.php
```diff
@@ -21,6 +21,33 @@
...
 class VersionConstraint extends SpecificConstraint
 {
+    const OP_EQ = 0;
+    const OP_LT = 1;
+    const OP_LE = 2;
+    const OP_GT = 3;
+    const OP_GE = 4;
+    const OP_NE = 5;
+
+    static $transOpStr = array(
```
staabm: `private` visibility? (same below)

src/Composer/Package/LinkConstraint/VersionConstraint.php
```diff
@@ -100,14 +103,14 @@ private function doMatchSpecific(VersionConstraint $provider, $compareBranches =
 
         // an example for the condition is <= 2.0 & < 1.0
         // these kinds of comparisons always have a solution
-        if ($this->operator != '==' && $noEqualOp == $providerNoEqualOp) {
+        if ($this->operator != self::OP_EQ && $noEqualOp == $providerNoEqualOp) {
```
staabm: strict comparison?

Seldaek:
I think @staabm's comment make sense so no @naderman?

naderman:
Indeed, thanks @staabm - fixed

PR #4229, opened 2015-07-08
VersionParser: remove a few hundred/thousand array_merge calls

naderman:
This reduces array_merge calls on larger projects by thousands and saves a few hundred kilobytes through fewer more efficient memory allocations: So mostly a small time saver.

Packagist before/after:
```
[148.3MB/7.25s] Memory usage: 148.25MB (peak: 328.81MB), time: 7.25s
[148.2MB/7.01s] Memory usage: 148.24MB (peak: 328.72MB), time: 7.01s
```

### DIFF
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -257,11 +257,12 @@ class VersionParser
         foreach ($orConstraints as $constraints) {
             $andConstraints = preg_split('{(?<!^|as|[=>< ,]) *(?<!-)[, ](?!-) *(?!,|as|$)}', $constraints);
             if (count($andConstraints) > 1) {
-                $constraintObjects = array(array());
+                $constraintObjects = array();
                 foreach ($andConstraints as $constraint) {
-                    $constraintObjects[] = $this->parseConstraint($constraint);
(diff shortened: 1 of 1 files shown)


src/Composer/Package/Version/VersionParser.php
```diff
@@ -257,10 +257,11 @@ public function parseConstraints($constraints)
         foreach ($orConstraints as $constraints) {
             $andConstraints = preg_split('{(?<!^|as|[=>< ,]) *(?<!-)[, ](?!-) *(?!,|as|$)}', $constraints);
             if (count($andConstraints) > 1) {
-                $constraintObjects = array();
+                $constraintObjects = array(array());
                 foreach ($andConstraints as $constraint) {
```
staabm: did you also consider
```php
$constraintObjects = array();
foreach ($andConstraints as $constraint) {
  foreach($this->parseConstraint($constraint) as $parsedConstraint) {
    $constraintObjects[] = $parsedConstraint;
  }
}
```

as another alternative?

IIRC `foreach` is faster then `array_merge`. This would also mean less memory-shuffling.

naderman: Indeed that's probably better since this isn't usually a large number of elements.

PR #4230, opened 2015-07-08
Fix: Update docs for validate command

localheinz:
This PR

* [x] updates the documentation for the validate command

Related to #4219, where I spotted that it's not up to date.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -339,7 +339,8 @@ php composer.phar validate
 
 ### Options
 
-* **--no-check-all:** Whether or not Composer does a complete validation.
+* **--no-check-all:** Do not emit a warning if requirements in `composer.json` use unbound version constraints.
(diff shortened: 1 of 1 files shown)


naderman:
Since you're rewriting this already, mind actually explain what "complete" means? That explanation doesn't explain anything really.

localheinz:
@naderman

Sure, will have a look tonight.

localheinz:
@naderman 

Had a look and updated the docs - what do you think?

naderman:
```
* **--no-check-all:** Do not emit a warning if requirements in `composer.json` use unbound version constraints.
* **--no-check-publish:** Do not emit an error if `composer.json` is unsuitable for publishing as a package on Packagist but is otherwise valid.
```

non-zero on error is implied I think.

localheinz:
Thank you, @naderman!

PR #4239, opened 2015-07-09
Reduce memory usage of Rule objects

naderman:
Storing the Rule data in a bitfield and getting rid of their id values saves up to 10% of memory.

This PR includes the commits from the rule hash changes in https://github.com/composer/composer/pull/4234

Comparison before/after: 
- Composer:
```
[96.2MB/1.63s] Analyzed 15693 rules to resolve dependencies
[67.7MB/1.80s] Memory usage: 67.7MB (peak: 96.48MB), time: 1.8s

[90.9MB/1.72s] Analyzed 15693 rules to resolve dependencies
[67.1MB/1.87s] Memory usage: 67.09MB (peak: 91.17MB), time: 1.87s
```
- Packagist:
```
[429.2MB/9.40s] Analyzed 144469 rules to resolve dependencies
[157.1MB/9.97s] Memory usage: 157.1MB (peak: 429.89MB), time: 9.97s

[386.0MB/10.05s] Analyzed 144469 rules to resolve dependencies
[157.1MB/10.52s] Memory usage: 157.09MB (peak: 386.64MB), time: 10.52s
```
- Symfony Standard:
```
[328.9MB/7.58s] Analyzed 94972 rules to resolve dependencies
[148.9MB/7.99s] Memory usage: 148.9MB (peak: 329.47MB), time: 7.99s

[299.4MB/7.15s] Analyzed 94972 rules to resolve dependencies
[148.3MB/7.69s] Memory usage: 148.35MB (peak: 299.88MB), time: 7.69s
```

### DIFF
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -29,6 +29,10 @@ class Rule
     const RULE_LEARNED = 12;
     const RULE_PACKAGE_ALIAS = 13;
 
+    const BITFIELD_TYPE = 0;
+    const BITFIELD_REASON = 8;
+    const BITFIELD_DISABLED = 16;
+
     /**
      * READ-ONLY: The literals this rule consists of.
      * @var array
@@ -50,7 +54,9 @@ class Rule
             $this->job = $job;
         }
 
-        $this->bitfield = (0 << 16) | ($reason << 8) | (255 << 0);
+        $this->bitfield = (0 << self::BITFIELD_DISABLED) |
+            ($reason << self::BITFIELD_REASON) |
+            (255 << self::BITFIELD_TYPE);
     }
 
     public function getHash()
@@ -66,7 +72,7 @@ class Rule
 
     public function getReason()
     {
-        return $this->getBitfield(1);
+        return $this->getBitfield(self::BITFIELD_REASON);
     }
 
     public function getReasonData()
@@ -110,32 +116,32 @@ class Rule
 
     public function setType($type)
     {
-        return $this->setBitfield(0, $type);
+        return $this->setBitfield(self::BITFIELD_TYPE, $type);
     }
 
     public function getType()
     {
-        return $this->getBitfield(0);
+        return $this->getBitfield(self::BITFIELD_TYPE);
     }
 
     public function disable()
     {
-        return $this->setBitfield(2, 1);
+        return $this->setBitfield(self::BITFIELD_DISABLED, 1);
     }
 
     public function enable()
     {
-        return $this->setBitfield(2, 0);
+        return $this->setBitfield(self::BITFIELD_DISABLED, 0);
     }
 
     public function isDisabled()
     {
-        return (bool) $this->getBitfield(2);
+        return (bool) $this->getBitfield(self::BITFIELD_DISABLED);
     }
 
     public function isEnabled()
     {
-        return !$this->getBitfield(2);
+        return !$this->getBitfield(self::BITFIELD_DISABLED);
     }
 
     /**
@@ -253,15 +259,13 @@ class Rule
         return implode(', ', $prepared);
     }
 
-    private function getBitfield($var)
+    private function getBitfield($offset)
     {
-        $offset = 8*$var;
         return ($this->bitfield & (255 << $offset)) >> $offset;
     }
 
-    private function setBitfield($var, $value)
+    private function setBitfield($offset, $value)
     {
-        $offset = 8*$var;
         $this->bitfield = ($this->bitfield & ~(255 << $offset)) | ((255 & $value) << $offset);
     }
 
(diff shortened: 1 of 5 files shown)


src/Composer/DependencyResolver/Rule.php
```diff
@@ -29,63 +29,50 @@ class Rule
...
-        $this->reason = $reason;
         $this->reasonData = $reasonData;
 
-        $this->disabled = false;
-
-        $this->job = $job;
-
-        $this->type = -1;
+        if ($job) {
+            $this->job = $job;
```
staabm: This property is no longer declared

src/Composer/DependencyResolver/Rule.php
```diff
@@ -29,63 +29,50 @@ class Rule
...
         $this->literals = $literals;
-        $this->reason = $reason;
         $this->reasonData = $reasonData;
 
-        $this->disabled = false;
-
-        $this->job = $job;
-
-        $this->type = -1;
+        if ($job) {
```
staabm: This if is unnecessary

src/Composer/DependencyResolver/Rule.php
```diff
@@ -133,32 +116,32 @@ public function equals(Rule $rule)
...
     public function disable()
     {
-        $this->disabled = true;
+        return $this->setBitfield(self::BITFIELD_DISABLED, 1);
     }
 
     public function enable()
     {
-        $this->disabled = false;
+        return $this->setBitfield(self::BITFIELD_DISABLED, 0);
```
staabm: `setBitfield` doesnt return a value

src/Composer/DependencyResolver/Rule.php
```diff
@@ -133,32 +116,32 @@ public function equals(Rule $rule)
...
     public function getType()
     {
-        return $this->type;
+        return $this->getBitfield(self::BITFIELD_TYPE);
     }
 
     public function disable()
     {
-        $this->disabled = true;
+        return $this->setBitfield(self::BITFIELD_DISABLED, 1);
```
staabm: `setBitfield` doesnt return a value

src/Composer/DependencyResolver/Rule.php
```diff
@@ -133,32 +116,32 @@ public function equals(Rule $rule)
 
     public function setType($type)
     {
-        $this->type = $type;
+        return $this->setBitfield(self::BITFIELD_TYPE, $type);
```
staabm: `setBitfield` doesnt return a value

src/Composer/DependencyResolver/Rule.php
naderman: Read the description of the respecitve commit. That's the point, that's how the memory is saved.

src/Composer/DependencyResolver/Rule.php
naderman: No

src/Composer/DependencyResolver/Rule.php
staabm: Not declaring a property will increase memory usage under php7

naderman: That does not matter at all. We have a large number of objects that do not need this property at all, and a tiny fraction that need them. Whether or not those few objects are larger because of the lack of declaration is offset by far by the amount of memory saved by all the objects that do not need the property in the first place.

Declared properties are already stored more efficiently than undeclared ones as of PHP 5.4 btw ...

staabm: Ok, never thought about this kind of micro optimization. ;)

Just to make things clear
No prop: http://3v4l.org/galLP vs declared prop: http://3v4l.org/lEcKC

You are saving (e.g. In 5.4.0-5.6.10)
3407872 (~3MB) bytes per 100000 objects => so it has a tiny impact (or your use case produces millions of objects)

In php7 it saves 0bytes. 

Dont take this as a offence, I am just playing numbers :).

naderman: Maybe you noticed how I posted benchmarks on the PR description ... This micro optimization saves 1 to 3 % of memory consumption, since we produce multiple hundreds of thousands of these objects on larger projects.

naderman:
For the record I'm using https://github.com/arnaud-lb/php-memory-profiler to identify areas that use the largest amounts of memory in order to optimize them.

Seldaek:
Changes look good to me although one detail I wonder about is whether the getBitfield/setBitfield indirection is necessary. It adds an extra function call on each retrieval of those bitfield values, I didn't check if they are called a lot but if so it might add up. 

It's not clear in your benchmarks if cpu time went up or down at all, but do you wanna try that if you have a decent bench setup at hand? I have too much crap running on my machine at the moment to make a valid test :)

naderman:
The difference at least isn't large enough to make a reliably measurable difference if you run composer completely including all the networking etc. However I'll simply remove it, as it's definitely more function calls than necessary. Was using pack/unpack at first which would have been to much to copy/paste around, but with the simple integer I think it's fine to just duplicate the logic.

PR #4245, opened 2015-07-11
Always optimize autoloader when classmap is authoritative

### LINKED ISSUE #4244: composer require should have an --optimize-autoloader switch
`composer install` has an `--optimize-autoloader` switch which makes it include every class in the classmap, but when it is invoked internally via `composer require` there is no way to trigger it. This is especially problematic for installations using the `classmap-authoritative` configuration flag, which will be broken by any invocation of `composer require` as it will regenerate the classmap but exclude PSR4 and similar files.

So, there should either be an `--optimize-autoloader` switch for `composer require`. (Alternatively, composer could always optimize when `classmap-authoritative` is enabled.)

tgr:
When the classmap is authoritative, a non-optimized classmap
will mean the autoloader does not load some of the files it is
supposed to. composer require/remove have no way of generating
optimized classmaps and thus would break an installation using
composer in authoritative mode. Thus, there is reason not to use
optimization when classmap-authoritative is enabled.

Fixes #4244

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -53,6 +53,10 @@ class AutoloadGenerator
 
     public function dump(Config $config, InstalledRepositoryInterface $localRepo, PackageInterface $mainPackage, InstallationManager $installationManager, $targetDir, $scanPsr0Packages = false, $suffix = '')
     {
+        if ($config->get('classmap-authoritative')) {
+            $scanPsr0Packages = true;
+        }
+
         $this->eventDispatcher->dispatchScript(ScriptEvents::PRE_AUTOLOAD_DUMP, $this->devMode, array(), array(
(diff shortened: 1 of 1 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -53,6 +53,10 @@ public function setDevMode($devMode = true)
 
     public function dump(Config $config, InstalledRepositoryInterface $localRepo, PackageInterface $mainPackage, InstallationManager $installationManager, $targetDir, $scanPsr0Packages = false, $suffix = '')
     {
+		if ($config->get('classmap-authoritative')) {
```
alcohol: Indentation is off.

Seldaek:
Closing in favor of #4362 because it kinda includes the fix and merging this will just make the other one conflict

PR #4262, opened 2015-07-16
Fix for tests

remicollet:
* test suite start failing with justinrainbow/json-schema 1.4.4
* 1 test is failing on distro using "jsonc" (instead of non-free json ext.)

PR #4264, opened 2015-07-16
remove spdx files, introduce external library

alcohol:
Title explains it all.

* [x] remove spdx files
* [x] add spdx-licenses to composer.json
* [x] adjust compiler

Includes fix for tests from #4262 - but only with updated json-schema dependency, did not update any other dependencies.


src/Composer/Util/SpdxLicense.php
```diff
@@ -1,306 +0,0 @@
...
-
-namespace Composer\Util;
-
-/**
- * Supports composer array and SPDX tag notation for disjunctive/conjunctive
- * licenses.
- *
- * @author Tom Klingenberg <tklingenberg@lastflood.net>
- */
-class SpdxLicense
```
stof: Maybe this class should be kept for a while (but deprecated), proxying calls to the new implementation, in case someone was using it already.

alcohol: I could also extend the new class and only add the deprecated function? How does that sound?

alcohol: Sorry I phrased that wrong, I mean extend the new class and add a deprecation warning in the constructor?

stof: yeah, looks good.

No need to add it in the constructor (which requires overwriting it). you can add it at the top of the file, as done in Symfony.

alcohol: Not sure what you meant by "at the top" (cannot use `__CLASS__` then right?); how does this look instead?

src/Composer/Compiler.php
```diff
@@ -95,7 +95,8 @@ public function compile($pharFile = 'composer.phar')
         $finder = new Finder();
         $finder->files()
             ->name('*.json')
-            ->in(__DIR__ . '/../../res')
+            ->in(__DIR__.'/../../res')
+            ->in(__DIR__.'/../../vendor/composer/spdx-licenses/res/')
```
alcohol: Maybe I should replace this with `SpdxLicenses:: getResourcesDir()`?

/cc @stof what do you think?

stof: looks like a good idea as we have this method

src/Composer/Util/SpdxLicense.php
stof: see https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Validator/Validator/LegacyValidator.php

PR #4305, opened 2015-07-27
Fixes #4302 allow COMPOSER env with config statements (if --file not used)

### LINKED ISSUE #4302: Issue: Setting the COMPOSER Env has no effect on composer config repositories statement
Regarding to the documentation https://getcomposer.org/doc/03-cli.md#composer it should be possible to change the name of the configuration file from the default value composer.json to another filename (e.g. my-configuration.json). This works fine for command line statements like

COMPOSER=my-configuration.json php ./composer.phar require vendorname/my-package:dev-master

which will add the vendorname/my-package package to the require block in my my-configuration.json file.

But if I try to add a new repository for vendorname/my-package (because its a private package which is not available at packagist.org) with the following statement

COMPOSER=my-configuration.json php ./composer.phar config repositories.vendorname/my-package git https://url-to-my-private-git-repository.git

this information always will be written to the composer.json file which is located in the same directory as the my-configuration.json file.

Its seems that the COMPOSER Env setting variable has no effect on the [config] option.

Tehem:
I chose to make the env variable override the parameter, which seems logical ?

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -139,7 +139,7 @@ EOT
         // passed in a file to use
         $configFile = $input->getOption('global')
             ? ($this->config->get('home') . '/config.json')
-            : $input->getOption('file');
+            : (trim(getenv('COMPOSER')) ? trim(getenv('COMPOSER')) : $input->getOption('file'));
 
         // create global composer.json if this was invoked using `composer global config`
         if ($configFile === 'composer.json' && !file_exists($configFile) && realpath(getcwd()) === realpath($this->config->get('home'))) {
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -138,12 +138,8 @@ EOT
         // Get the local composer.json, global config.json, or if the user
         // passed in a file to use
         $configFile = $input->getOption('global')
-            ? ($this->config->get('home').'/config.json')
-            : (
-            $input->getOption('file')
-                ? $input->getOption('file')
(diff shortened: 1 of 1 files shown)


alcohol:
I am not sure if that is logical. Environment variables are more of a global scope, whereas `--file` is local to the function only. I think local should have priority over global.

Tehem:
I see your point, but the 'file' input option has composer.json as default value, so if I give priority to this parameter, it will always be set to composer.json even when not specified. So it would imply more complex modification to the source code to make this work. But I will comply if this is the chosen behavior.

alcohol:
I agree the default value makes it a bit inconvenient. I guess this will suffice. Though I am not sure if this change is even necessary at all.

Seldaek:
I think it should rather default to COMPOSER env OR `composer.json` if --file is not provided, but if it is then we should take that.

alcohol:
So,

* [ ] remove the default value from the `file` argument
* [ ] `--file` > `COMPOSER` > default (`composer.json`)

Tehem:
Ok I will make the necessary adjustments. Thank you!

src/Composer/Command/ConfigCommand.php
```diff
@@ -138,8 +138,12 @@ protected function initialize(InputInterface $input, OutputInterface $output)
         // Get the local composer.json, global config.json, or if the user
         // passed in a file to use
         $configFile = $input->getOption('global')
-            ? ($this->config->get('home') . '/config.json')
-            : $input->getOption('file');
+            ? ($this->config->get('home').'/config.json')
+            : (
+            $input->getOption('file')
+                ? $input->getOption('file')
```
stof: you can use the ``?:`` shorthand here

Seldaek:
Thanks!

PR #4324, opened 2015-08-03
Improve documentation for WritableRepositoryInterface::getCanonicalPackages

MalteWunsch:
If I understand the API of `\Composer\Repository\WritableRepositoryInterface::getCanonicalPackages`:

```php
/**
 * Get unique packages, with aliases resolved and removed
 * @return PackageInterface[]
 */
public function getCanonicalPackages();
```

and `\Composer\Package\PackageInterface`:

```php
/**
 * Defines the essential information a package has that is used during solving/installation
 */
interface PackageInterface
{
    // ...

    /**
     * Returns the version of this package
     * @return string version
     */
    public function getVersion();
```

correctly, `getCanonicalPackages` should return multiple packages with the same name in different versions. At least we were expecting that behaviour as `\Composer\Repository\FilesystemRepository::write` makes use of it. Or am I thinking in a wrong direction?


stof:
This change is breaking tests

MalteWunsch:
Ooops. Got a segmentation fault when running the whole test suite locally, and before fixing that, I lazily wanted Travis to have a shot.

I'm not really sure why the tests are breaking, but I'd be willing to dig into that If one can assure me the API is meant the way I think, i.e. the idea of this PR is right?

stof:
they are breaking because you are now writing too much stuff to the filesystem repository. so it looks like the current behavior was expected

MalteWunsch:
Makes sense. So documenting would be the better PR?

Seldaek:
Thanks

PR #3975, opened 2015-04-29
fixes #3974: Autoloader have to distinguish between files and directories

marc-mabe:
#3974

PR #4327, opened 2015-08-04
Revert "Merge pull request #3975 from marc-mabe/hotfix/3974"

nicolas-grekas:
This reverts commit bdb6ecb29e9624fb767aa690d09c780249050fc6, reversing
changes made to 8a12e50a16dbc4ddf808b752ad64b412dba272d5.

While working on the [perf of the DebugClassLoader](https://github.com/symfony/symfony/pull/15443), I noticed the change.

`is_file()` makes a stat syscall, which makes it MUCH slower that `file_exists`. Given that the autoloader is already one of the most perf hungry component of the PHP stack, this slows down everyone for almost no reason (but a very rare edge case that does not deserve it).

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -351,7 +351,7 @@ class ClassLoader
             foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
                 if (0 === strpos($class, $prefix)) {
                     foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
-                        if (is_file($file = $dir . DIRECTORY_SEPARATOR . substr($logicalPathPsr4, $length))) {
+                        if (file_exists($file = $dir . DIRECTORY_SEPARATOR . substr($logicalPathPsr4, $length))) {
                             return $file;
                         }
                     }
@@ -361,7 +361,7 @@ class ClassLoader
 
         // PSR-4 fallback dirs
         foreach ($this->fallbackDirsPsr4 as $dir) {
-            if (is_file($file = $dir . DIRECTORY_SEPARATOR . $logicalPathPsr4)) {
+            if (file_exists($file = $dir . DIRECTORY_SEPARATOR . $logicalPathPsr4)) {
                 return $file;
             }
         }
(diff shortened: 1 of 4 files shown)


javiereguiluz:
I've tested this patch using the Symfony Demo app with PHP 5.6 in Mac OS X and it produces a small but nice performance improvement:

patch_composer

Seldaek:
Fair enough. Thanks for the research and details. /cc @marc-mabe just FYI

stof:
@javiereguiluz when linking to a comparison, you should not use the private comparison link. You should use the shareable link instead.

javiereguiluz:
@stof you are right. I'm sorry. The link has been fixed. Thanks.

marc-mabe:
@nicolas-grekas @Seldaek Interesting. I thought `is_file` ist using the same stat cache as `file_exists` does.

Form http://php.net/en/is_file
>     Note: The results of this function are cached. See clearstatcache() for more details.

And from the PHP source code (master) I found both functions are using exactly the same underlying C function:
https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/ext/standard/filestat.c#L1168
https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/ext/standard/filestat.c#L1183
-> using the same Macro: https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/ext/standard/filestat.c#L1080

So a performance difference doesn't make any sense.

stof:
@marc-mabe they are not using the same underlying function. One is calling the macro with ``FS_IS_FILE`` and the other one with ``FS_EXISTS``. This changes what is executed in the function: 

https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/ext/standard/filestat.c#L990
https://github.com/php/php-src/blob/0437aa2abf1eb494e4555d7d4dfbdbf566a59bdc/ext/standard/filestat.c#L996

marc-mabe:
@stof ah I See. You are right. Sorry for my mistake.

Michal-sk:
This was two years ago. Do you still have some benchmarks? Because when i run my app with composer it's faster when i use `is_file`. PHP 7.1.5, benchmark done in blackfire.

PR #4342, opened 2015-08-09
Add a new option to view local modifications (git diff)

iamluc:
When a vendor package has been modified locally, this PR allows to view the modifications (the result of `git diff HEAD`).
That's helpful when you don't remember what was the modifications.

Before this PR, it was only possible to see file names.


Seldaek:
Thanks!

Issue #4268, opened 2015-07-18
PHPDoc problems in Repository Class


xjchengo:
In some Repository Class like `CompositeRepository`, there are some method using `{@inheritdoc}` but noting to inherit.
Example:
```
class CompositeRepository implements RepositoryInterface
{
    /**
     * {@inheritDoc}
     */
    public function filterPackages($callback, $class = 'Composer\Package\Package')
    {
        foreach ($this->repositories as $repository) {
            if (false === $repository->filterPackages($callback, $class)) {
                return false;
            }
        }

        return true;
    }
}
```
There is no doc to inherit in `RepositoryInterface`.

PR #4345, opened 2015-08-10
Changed comments

nicolasdewez:
Comments of CompositeRepository::filterPackages was false #4268 
It's a small PR waiting an other :-)


Seldaek:
Actually looking at it, I don't know what this function does, it doesn't get called ever and it can't possibly work as other repositories don't implement that method.. :)

nicolasdewez:
Yes I had the same opinion but ... I had a doubt :)

PR #4356, opened 2015-08-14
copy LICENSE file during autoload dumping ( refs #4288 )

### LINKED ISSUE #4288: composer generated ClassLoader.php references non-existent LICENSE file
After building a composer autoloader, the `vendor/composer/ClassLoader.php` file has the following heading:
```
/*
 * This file is part of Composer.
 *
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */
```

but there is no LICENSE distributed. It may make sense to just include a url to the license instead.

alcohol:
closes #4288

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -261,14 +261,8 @@ EOF;
         file_put_contents($vendorPath.'/autoload.php', $this->getAutoloadFile($vendorPathToTargetDirCode, $suffix));
         file_put_contents($targetDir.'/autoload_real.php', $this->getAutoloadRealFile(true, (bool) $includePathFile, $targetDirLoader, (bool) $includeFilesFile, $vendorPathCode, $appBaseDirCode, $suffix, $useGlobalIncludePath, $prependAutoloader, $classMapAuthoritative));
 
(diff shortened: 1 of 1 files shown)


Seldaek:
yay

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -750,4 +744,20 @@ protected function sortPackageMap(array $packageMap)
...
         return $sortedPackageMap;
     }
+
+    /**
+     * Copy file using stream_copy_to_stream to work around https://bugs.php.net/bug.php?id=6463
+     *
+     * @param string $source
+     * @param string $target
+     */
+    protected function safeCopy($source, $target)
```
stof: should be private IMO

alcohol: all of the helper methods use protected, so I went for consistency :-(

Seldaek: Oh right, I'll fix that.

Seldaek: Yeah just saw that, we probably should fix it eventually, but let's 
leave as is for now, I doubt anyone extends this anyway.

PR #4357, opened 2015-08-14
Warn if cache directory cannot be created, refs #1161

### LINKED ISSUE #1161: Composer does not check whether cache directory is writeable
It seems like currently composer does not specifically check if it can create the .composer cache folder in the home directory.

In case, you have a private repository that will make it think that the private ssh-key authentication failed, and fall back to asking for the password (in interactive mode).

alcohol:
Let me know if

* [ ] I should add a conditional verbosity check around this warning.

Closes #1161

### DIFF
--- a/src/Composer/Cache.php
+++ b/src/Composer/Cache.php
@@ -45,6 +45,7 @@ class Cache
 
         if (!is_dir($this->root)) {
             if (!@mkdir($this->root, 0777, true)) {
+                $this->io->writeError('<warning>Cannot create cache directory ' . $this->root . ', proceeding without cache</warning>');
                 $this->enabled = false;
             }
         }


src/Composer/Cache.php
```diff
@@ -45,6 +45,7 @@ public function __construct(IOInterface $io, $cacheDir, $whitelist = 'a-z0-9.',
 
         if (!is_dir($this->root)) {
             if (!@mkdir($this->root, 0777, true)) {
+                $this->io->writeError('Cache directory ' . $this->root . ' is not writeable, proceeding without cache');
```
stof: What happens here is that the cache directory cannot be created. This code is not checking whether it is writable or no.

alcohol: Semantics.. :\

Issue #4254, opened 2015-07-14
Composer does not remove autoload_files.php when related packages are removed


hlubek:
How to reproduce:

    composer require guzzlehttp/guzzle "~5.2"

This creates a `autoload_files.php` that requires `react/promise`.

    composer remove guzzlehttp/guzzle
    composer update

This will remove the `autoload_files.php` from `autoload_real.php` but does not remove the file itself. This  is of course correct if you just rely on `autoload_real.php`. In the Flow framework we use a custom class loader that includes `autoload_files.php` directly if it exists. For this case it would be very helpful if Composer would remove the missing package from `autoload_files.php`.

KingCrunch:
Hi @Seldaek 

This seems to be fixed in #4359 and can get closed.

Seldaek:
Yup thanks

PR #4359, opened 2015-08-15
Removing file autoload_files.php was added (#4254)

voroks:
in cases of removing package (#4254) or removing autoload flag of package: if list of files to load is empty, file autoload_files.php will be removed


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -255,9 +255,12 @@ public static function autoload(\$class)
         if ($includePathFile = $this->getIncludePathsFile($packageMap, $filesystem, $basePath, $vendorPath, $vendorPathCode52, $appBaseDirCode)) {
             file_put_contents($targetDir.'/include_paths.php', $includePathFile);
         }
```
Seldaek: How about doing the same for the include_paths.php file above? Might as well fix both at once.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -255,9 +255,12 @@ public static function autoload(\$class)
...
+	$includeFilesFilePath = $targetDir.'/autoload_files.php';
         if ($includeFilesFile = $this->getIncludeFilesFile($autoloads['files'], $filesystem, $basePath, $vendorPath, $vendorPathCode52, $appBaseDirCode)) {
-            file_put_contents($targetDir.'/autoload_files.php', $includeFilesFile);
-        }
+            file_put_contents($includeFilesFilePath, $includeFilesFile);
+        } else if (file_exists($includeFilesFilePath) === true) {
+	    unlink($includeFilesFilePath);
+	}
```
stof: wrong indentation here

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -255,9 +255,12 @@ public static function autoload(\$class)
         if ($includePathFile = $this->getIncludePathsFile($packageMap, $filesystem, $basePath, $vendorPath, $vendorPathCode52, $appBaseDirCode)) {
             file_put_contents($targetDir.'/include_paths.php', $includePathFile);
         }
+	$includeFilesFilePath = $targetDir.'/autoload_files.php';
         if ($includeFilesFile = $this->getIncludeFilesFile($autoloads['files'], $filesystem, $basePath, $vendorPath, $vendorPathCode52, $appBaseDirCode)) {
-            file_put_contents($targetDir.'/autoload_files.php', $includeFilesFile);
-        }
+            file_put_contents($includeFilesFilePath, $includeFilesFile);
+        } else if (file_exists($includeFilesFilePath) === true) {
```
stof: no need for ``=== true``. Turning a boolean into another boolean is not needed

voroks: Thanks for your feedback. Code style standard in our company requires explicit comparison and indentation with using tabulation. Sorry for my inattention :)

src/Composer/Autoload/AutoloadGenerator.php
voroks: On the way

Seldaek:
Thanks for the updates, looks good now!

PR #4362, opened 2015-08-16
Add cli argument for classmap-authoritative

### LINKED ISSUE #4361: Add cli argument to enable classmap-authoritative functionality
In https://github.com/composer/composer/pull/3610#issuecomment-71929902 @hkdobrev wrote:
> > It's probably worth discussing whether or not this should also have a related cli argument for install, update and dump-autoload.
>
> I think CLI interface for this would be useful for people who are not using this during development, but are doing `composer install --no-dev --optimize-autoloader` on production. Which I believe is quite common.

Adding cli argument so that `composer install --no-dev --optimize-autoloader --classmap-authoritative` will create an `autoload_real.php` with `$loader->setClassMapAuthoritative(true);` The same option should be added to `composer update` and `composer dumpautoload`. The existing `classmap-authoritative` configuration setting should still be respected as well.

bd808:
Add a `--classmap-authoritative (-a)` argument to `composer install`,
`composer update` and `composer dumpautoload`. This enables the same
authoritative classmap behavior as the existing `classmap-authoritative`
configuration setting. The option can be used for creating highly
optimized production autoloaders via `composer install --no-dev
--optimize-autoloader --classmap-authoritative` for projects where
multiple autoloaders are present and unnecessary `file_exists` calls
introduce performance issues.

Closes #4361


alcohol:
The code looks fine, but I find the argument description rather confusing. A double negative is hard to understand for non-native English speakers I believe.

doc/03-cli.md
```diff
@@ -97,6 +97,9 @@ resolution.
 * **--optimize-autoloader (-o):** Convert PSR-0/4 autoloading to classmap to get a faster
   autoloader. This is recommended especially for production, but can take
   a bit of time to run so it is currently not done by default.
+* **--classmap-authoritative (-a):** Create autoloader that will not scan the
+  filesystem for classes that are not found in the classmap (implicitly enables
```
Ocramius: This is indeed very confusing to read. Simply get rid of the double negation

doc/03-cli.md
```diff
@@ -97,6 +97,9 @@ resolution.
...
+  --optimize-autoloader)
```
Ocramius: Add backticks here

doc/03-cli.md
```diff
@@ -140,6 +143,9 @@ php composer.phar update vendor/*
...
```
Ocramius: Backticks

doc/03-cli.md
```diff
@@ -140,6 +143,9 @@ php composer.phar update vendor/*
 * **--optimize-autoloader (-o):** Convert PSR-0/4 autoloading to classmap to get a faster
   autoloader. This is recommended especially for production, but can take
   a bit of time to run so it is currently not done by default.
+* **--classmap-authoritative (-a):** Create autoloader that will not scan the
+  filesystem for classes that are not found in the classmap (implicitly enables
```
Ocramius: Same as above

doc/03-cli.md
```diff
@@ -505,6 +511,9 @@ performance.
 * **--optimize (-o):** Convert PSR-0/4 autoloading to classmap to get a faster
   autoloader. This is recommended especially for production, but can take
   a bit of time to run so it is currently not done by default.
+* **--classmap-authoritative (-a):** Create autoloader that will not scan the
+  filesystem for classes that are not found in the classmap (implicitly enables
+  --optimize)
```
Ocramius: Same

doc/03-cli.md
```diff
@@ -505,6 +511,9 @@ performance.
 * **--optimize (-o):** Convert PSR-0/4 autoloading to classmap to get a faster
   autoloader. This is recommended especially for production, but can take
   a bit of time to run so it is currently not done by default.
+* **--classmap-authoritative (-a):** Create autoloader that will not scan the
+  filesystem for classes that are not found in the classmap (implicitly enables
```
Ocramius: same

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -39,6 +39,7 @@ class AutoloadGenerator
     private $io;
 
     private $devMode = false;
+    private $classMapAuthoritative = false;
```
Ocramius: Spacing + docblock, I'd say

alcohol:
I guess it is in line with the current description of `classmap-authoritative` though.

> Autoloader will not scan the filesystem for classes that are not found in the class map. Implies 'optimize-autoloader'.

But I can't say I'm a big fan of that description either.


Maybe the following is easier to comprehend:

> If true, the Composer autoloader will only load classes from the classmap. Implies '`optimize-autoloader`'.

And then the description for this option would be:

> Autoload classes from the classmap only. Implies '`optimize-autoloader`'.

alcohol:
Also please end your descriptions with a `.`. Punctuation is a part of descriptions too :-)

src/Composer/Autoload/AutoloadGenerator.php
alcohol: Maybe not a docblock, but spacing for consistency.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -51,6 +52,17 @@ public function setDevMode($devMode = true)
         $this->devMode = (boolean) $devMode;
     }
 
+    /**
+     * Whether or not generated autoloader considers the class map
+     * authoritative.
+     *
+     * @param bool $classMapAuthoritative
+     */
+    public function setClassMapAuthoritative($classMapAuthoritative = false)
```
alcohol: I would not set a default value here. Better to make it explicitly required to pass true or false.

src/Composer/Command/InstallCommand.php
```diff
@@ -110,7 +111,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $preferDist = $input->getOption('prefer-dist');
         }
 
-        $optimize = $input->getOption('optimize-autoloader') || $config->get('optimize-autoloader') || $config->get('classmap-authoritative');
+        $authoritative = $input->getOption('classmap-authoritative') || $config->get('classmap-authoritative');
+        $optimize = $input->getOption('optimize-autoloader') || $config->get('optimize-autoloader') || $authoritative;
```
Seldaek: Instead of the `|| $authoritative` everywhere I think it'd be best to internally set optimize to true in the AutoloadGenerator when authoritative is enabled, this would cover non-optimize-aware commands like require, and fix #4244

Seldaek: Or maybe we do it the other way, and instead disable authoritative classmaps if the build is not optimized. Any thoughts on that @tgr?

bd808:
Changes in new commit vs previous:

* Updated documentation to eliminate double negative
* Added phpdoc and whitespace for new AutoloadGenerator member var
* Force classmap in AutoloadGenerator::dump() when classMapAuthoritative=true
* Force matching optimizeAutoloader and classMapAuthoritative in Installer

@Seldaek said on irc that amending was preferred to a second commit for a small changeset. (Which is totally the right answer by the way.)

alcohol:
You forgot to update the description of the Command options.. sorry :-(

bd808:
Changes in new commit vs previous:
* Updated cli help to eliminate double negative

alcohol:
This mostly looks good to me now. Some of the inline comments feel a bit like stating the obvious and unnecessary. Does this now cover non-optimize aware commands such as `require` as well? It would be nice if this addresses #4244 as well, as @Seldaek mentioned.

bd808:
>Does this now cover non-optimize aware commands such as require too? It would be nice if this addresses #4244 as well, as @Seldaek mentioned.

As it stands, this pull request does address #4244 but possibly not in the way that @tgr imagined. Control of the classmap-authoritative feature has been pushed to the cli entry points by removing the `$config->get('classmap-authoritative')` call from `AutoloadGenerator::dump()`. This means that `composer require` and `composer remove` will no longer generate an autoloader that enforces classmap-authoritative no matter what config settings are present. This fixes the case where a classmap-authoritative autoloader was generated without also generating an updated classmap which was half of the problem.

A more complete fix would be to add `--optimize-autoloader` and `--classmap-authoritative` (and related config setting) processing to both `RequireCommand` and `RemoveCommand`. These appear to be the only two commands that invoke `Installer::run()` without setting `Installer::setOptimizeAutoloader()` and `Installer::setClassMapAuthoritative()`.

bd808:
> A more complete fix would be to add --optimize-autoloader and --classmap-authoritative (and related config setting) processing to both RequireCommand and RemoveCommand.

Added in 600f9eb

alcohol:
Maybe add some tests/assertions that turning optimize off indeed forced authoritative off (and the other way around, auth on forces optimize on) on the AutoloadGeneratorTest class? :)

That way a change in that behaviour cannot be snuck in inadvertently.

bd808:
>Maybe add some tests/assertions

I changed  `testClassMapAutoloadingAuthoritative` so that it asserts that a classmap autoloader is generated with authoritative=true and optimize=false. The inverse scenario only happens via Installer right now so it is a little more difficult to test.

bd808:
Rebased on to current HEAD (d831c86) and fixed merge conflict.

Hikariii:
+1

Seldaek:
Thanks!

Seldaek:
Also copy pasting is evil, reminder #943802 :p https://github.com/composer/composer/commit/1e8d7255b7d50238b95eb41b86ac49ecd2f10aba#diff-7ce9b716dc8b2b85ea617bd14a1235b6

PR #4365, opened 2015-08-18
Add a path repository and downloader

sroze:
This PR aims to replace #3904 because it's getting slow.

In addition, I've extracted everything related to version guessing from `RootPackageLoader` to a version `VersionGuesser` class. This class is used in the `PathRepository` to guess the version of the target dependency.

If the symlink fail, then we use the mirror method of Symfony Filesystem component.

### DIFF
--- a/tests/Composer/Test/Package/Version/VersionGuesserTest.php
+++ b/tests/Composer/Test/Package/Version/VersionGuesserTest.php
@@ -13,24 +13,21 @@
 namespace Composer\Test\Package\Version;
 
 use Composer\Config;
-use Composer\Package\Loader\RootPackageLoader;
 use Composer\Package\Version\VersionGuesser;
 use Composer\Package\Version\VersionParser;
 
 class VersionGuesserTest extends \PHPUnit_Framework_TestCase
 {
-    public function testDetachedHeadBecomesDevHash()
+    public function setUp()
     {
         if (!function_exists('proc_open')) {
             $this->markTestSkipped('proc_open() is not available');
         }
+    }
 
+    public function testDetachedHeadBecomesDevHash()
+    {
         $commitHash = '03a15d220da53c52eddd5f32ffca64a7b3801bea';
-
-        $manager = $this->getMockBuilder('\\Composer\\Repository\\RepositoryManager')
-            ->disableOriginalConstructor()
-            ->getMock()
-        ;
 
         $executor = $this->getMockBuilder('\\Composer\\Util\\ProcessExecutor')
             ->setMethods(array('execute'))
@@ -69,15 +66,6 @@ class VersionGuesserTest extends \PHPUnit_Framework_TestCase
 
     public function testTagBecomesVersion()
     {
-        if (!function_exists('proc_open')) {
-            $this->markTestSkipped('proc_open() is not available');
-        }
-
-        $manager = $this->getMockBuilder('\\Composer\\Repository\\RepositoryManager')
-            ->disableOriginalConstructor()
-            ->getMock()
-        ;
-
         $executor = $this->getMockBuilder('\\Composer\\Util\\ProcessExecutor')
             ->setMethods(array('execute'))
             ->disableArgumentCloning()
@@ -108,15 +96,6 @@ class VersionGuesserTest extends \PHPUnit_Framework_TestCase
 
     public function testInvalidTagBecomesVersion()
     {
-        if (!function_exists('proc_open')) {
-            $this->markTestSkipped('proc_open() is not available');
-        }
-
-        $manager = $this->getMockBuilder('\\Composer\\Repository\\RepositoryManager')
-            ->disableOriginalConstructor()
-            ->getMock()
-        ;
-
         $executor = $this->getMockBuilder('\\Composer\\Util\\ProcessExecutor')
             ->setMethods(array('execute'))
             ->disableArgumentCloning()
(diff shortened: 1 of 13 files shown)


tests/Composer/Test/Package/Version/VersionGuesserTest.php
```diff
@@ -0,0 +1,158 @@
...
+        $config->merge(array('repositories' => array('packagist' => false)));
+        $guesser = new VersionGuesser($executor, new VersionParser());
+        $version = $guesser->guessVersion($config, array());
+
+        $this->assertEquals("2.0.5.0-alpha2", $version);
+    }
+
+    public function testInvalidTagBecomesVersion()
+    {
+        if (!function_exists('proc_open')) {
```
stof: this should be checked in the setUp method as it impacts all tests of the class

stof:
Tests for the new repository are missing

sroze:
@stof added repository tests.

sroze:
@stof added repository tests.

/cc @Seldaek

Seldaek:
Merged, thanks for the work, but I am a bit sad to see that this clearly was not even tested from a user perspective, as creating the PathRepository was failing with a fatal error. See https://github.com/composer/composer/commit/89c6a68a76a8beb9284ece77f65530a9da5fdb38 for my changes.

Seldaek:
Oh also it would be nice if someone could add it to the documentation.  I added #4412 to keep track of that.

sroze:
OMG, forgot to push this `PathRepository` construction problem \o/. Btw, why not using the `Filesystem` component anymore [here](https://github.com/composer/composer/commit/89c6a68a76a8beb9284ece77f65530a9da5fdb38#diff-620a11fbfc7f8f33cc6bc0b1ebc253b4R101) as it's now a dependency ?

Seldaek:
Because everyone knows what file_exists does and wrapping it in some class
that achieves nothing in this case only obfuscates the code IMO.

philbates35:
Thanks for everyone's work on this, it looks really promising. I've just been trying it out using the current latest Composer version (013a748) and I've come across an issue on Linux and a different issue on Windows.

I used the following folder structure in a Vagrant VM (using shared folders) on both Linux and Windows hosts, which is basically the same as the [documentation](https://getcomposer.org/doc/05-repositories.md#path):

    code
      /package
      /project

Relevant snippet from `code/package/composer.json`:

    ...
    "name": "my/package",
    "type": "package",
    ...

Relevant snippet from `code/project/composer.json`:

    ...
    "type": "project",
    "repositories": [
         {
             "type": "path",
             "url": "../package"
         }
     ],
    "require": {
         "my/package": "*@dev"
     },
    ...

### Linux Host Issue
When I run `composer install`, the symlink is created incorrectly:

    $ pwd
    code
    $ cd project
    $ composer install
    Installing dependencies (including require-dev)
      - Installing my/package (dev-master)
      Symlinked from ../package

    Writing lock file
    Generating autoload files
    $ ls -l vendor/my
    lrwxrwxrwx 1 vagrant www-data 10 Sep 14 13:17 package -> ../package

You can see it's using the URL relative to where the project's `composer.json` is, rather than where the symlink will eventually be created. So it seems when you enter  `"url": "../package"`, the symlink that is actually created should add an extra two levels, i.e. it should be:

    $ pwd
    code/project
    $ ls -l vendor/my
    lrwxrwxrwx 1 vagrant www-data 10 Sep 14 13:17 package -> ../../../package

### Windows Host Issue

When I run `composer install` from within the project, the package is installed successfully by being copied rather than symlinking as expected. The issue I have is that there is no nice way to update the package when I make changes to the package source.

I assumed that I running `composer update my/package` from the project would see that the package source files have changed and copied over the new version, but instead is says that there is nothing to install.

If I remove the `vendor` folder and re-run `composer install`, then the newest version of the package is installed.

I'm not sure if this PR was built with this in mind, but from reading comments relating to the previous PR (#3904), it seems like this PR will be most useful during package development. During package development on Windows using this new functionality currently, you would have to manually remove the package from the project's `vendor` folder and re-run `composer install my/package` every time you make a change to the package's source which isn't ideal.

Hope this made sense, if you need any more information let me know!

stof:
@philbates35 note that on Windows, you can also use the symlink PHP function in case you run your console with enough privileges (I don't remember how this permission is named, but Google should be able to answer)

sroze:
Thanks for your feedback @philbates35. I was able to reproduce your relative path problem and you can find a PR that introduces a fix here: https://github.com/composer/composer/pull/4419.

The Windows related problem is inconvenient but your filesystem should support symlinks in order to have a successful dev setup. I'm having a look if we can force to recopy again the directory when running `update` but not sure if I'll be possible without changing the version number in your `composer.json` file.

Seldaek:
On windows even with a copy, an update should work if you commit your changes, then it'll detect the git hash changed in ../package and update it. But if you only change files and don't commit it's pretty much impossible (or rather way too slow) for us to diff the whole directories.

sroze:
@philbates35 In that case I think you'll have to have a look at @stof's suggestion to be able to use symlinks instead of us adding it to composer.

philbates35:
Thanks for the explanation @Seldaek, that makes sense, I just tried making a commit and then running `composer update my/package` and it found the update for the package just fine.

If I make changes to the package files, commit, and then make further changes without committing, from what you said I was hoping that running `composer update my/package` in the project would update the package to the state of the package at the commit hash it finds. However, it instead copies the current state of the package's working directory (including all changes made since the last commit in the package). I don't suppose there's a workaround for that? Not a major issue, but I thought I'd bring it up anyway.

Seldaek:
Well we could technically do it I suppose using a git export or somesuch 
but it'd also probably surprise people. I don't think there is a really 
good way to do this except with symlink tbh :/

yched:
Thanks for this awesome feature !

There are cases, though, where the local package is committed in the "monolithic git repo" direclty at the final install destination (for example because it does not go into /vendor according to its composer-installer install path). In those cases, PathDownloader breaks because it first deletes the target, and then doesn't find the source (since they are the same)

Opened a PR for that : #4431

PR #4368, opened 2015-08-20
Update the config of the apt addon to the new syntax

stof:
``apt_packages`` was the syntax for the initial beta previews of the feature.


alcohol:
Confirmed @ http://docs.travis-ci.com/user/apt/#Adding-APT-Packages

Thanks :-)

PR #4379, opened 2015-08-26
Archive command shows created package path

javihgil:
Using archive command for package your project would be useful to know the path of created file, because of there are a lot of factors that affects to the file name (package name, version property, COMPOSER_ROOT_VERSION environment variable, branch alias configuration, etc.).

I propose this PR to improve this thing.


Seldaek:
Thanks @javihgil

Seldaek:
Note that I tweaked a bit so that the generated file name (and that alone) ends up in stdout, so you can do `file=$(composer archive)` for example and then use `$file` in a bash script.

javihgil:
Thanks @Seldaek, your change is even better!!

PR #4380, opened 2015-08-27
Improved CreateProject command

phpitler:
This is a quick fix but it should stop the 90%.

Interested to hear if anyone has a more long term solution?

Probably needs a unit test too.


PhillippOhlandt:
WTF, will report you on github and twitter as well.

TheSerenin:
Nah man, this is messed up.

kelunik:
You should really use the packages as keys and use `isset($banned[$package])` instead.

`/close`.

PR #4389, opened 2015-08-31
Switch autoload files to require_once rather than require

oligriffiths:
Not quire sure why this was ever `require` given that composer loads classes not side effect files.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -605,7 +605,7 @@ METHOD_FOOTER;
 
 function composerRequire$suffix(\$file)
 {
-    require \$file;
+    require_once \$file;
 }
 
 FOOTER;
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
@@ -51,5 +51,5 @@ class ComposerAutoloaderInitFilesAutoloadOrder
 
 function composerRequireFilesAutoloadOrder($file)
 {
-    require $file;
+    require_once $file;
 }
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_functions.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_functions.php
@@ -51,5 +51,5 @@ class ComposerAutoloaderInitFilesAutoload
 
 function composerRequireFilesAutoload($file)
 {
-    require $file;
+    require_once $file;
 }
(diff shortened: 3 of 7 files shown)


oligriffiths:
It should be noted that there is a minor performance advantage with `require`, but in the case of autoload files, of which there usually aren't going to be that many, as compared to autoloading class files, switching to `require_once` should have no visible performance impact.

This came about due to a need to manually require an autoload file from a package before requiring the composer autoload file.

Seldaek:
I'd really like to avoid doing this change unless you can convince me that there is no way for you to do it smarter. Because as you there is a small penalty, and especially once we do that change we can't go back because people will start relying on it to do all sorts of weird things.

Right now it's clear that you require the autoloader and then do things, because an autoloader should at most declare classes/functions, this should be able to be done as the very first thing. Can you try and explain why exactly you need to include files *before* the autoloader?

oligriffiths:
I want to use an alternative autoloader to the composer autoloader, however, I want the autoloader to be an installable composer package (to make use of the dependency management), that can be a dependency of other packages.

So here is the problem. I need other packages that intend to use my autoloader, be able to call that autoloader in order to register themselves (via autoload files). However, I need the options for the autoloader to be instantiated and configured PRIOR to any other packages registering themselves, but I don't want to force developers to HAVE to include my autoloader before composer, that should only happen if they NEED to provide configuration options. For example:

https://github.com/oligriffiths/nooku-class-loader/blob/master/composer.json

This defines an autoload file to pre-load the class. The class has a `getInstance($config)` method that allows the instance to be configured, this needs to be called BEFORE any dependencies call it IF one wants to supply configuration information.

```php
require 'vendor/oligriffiths/nooku-class-loader/src/loader.php';
Nooku\Library\ClassLoader::getInstance($options);
```

Then the dependencies autoload files would look like:

```php
<?php

Nooku\Library\ClassLoader::getInstance()
    ->getLocator('component')
    ->registerNamespace('Something','some/path');
```

I don't see any other way to solve this other than not putting the autoload definition of my package, and requiring it to be manually included before composer :cry:

alcohol:
Why not just prepend your autoloader when you are done bootstrapping it?

Third argument, http://php.net/manual/en/function.spl-autoload-register.php

oligriffiths:
Regardless, that doesn't get around the circularity issue. In order for me to be able to configure my autoloader, I need to include it and do some setup BEFORE composer is required. Once composer is required, the other autoload files are then included and things can be registered with my autoloader. I also don't want it to be a requirement to have to setup my autoloader before composer, it should work just fine on its own.

oligriffiths:
Any further thoughts on this?

alcohol:
Well it's virtually identical to #4093. So it seems there definitely is a demand for this change. As to whether or not it is justified (both scenarios are somewhat edge-case'ish), I cannot really comment. It would be nice perhaps to have some benchmarks to see if this change is insignificant.

Seldaek:
As far as I understand, #4186 is an alternative fix for this same issue, that actually fixes more of it, since require_once by itself if there are multiple paths involved will not solve the issue fully.

Seldaek:
Closing as this issue should be fixed by #4186

xwiz:
I may have run into some edge case scenario, for some reason once I rename a file to a particular name, I get Cannot declare class {className used in the file}, rename the file and the problem disappears. I have cleared the generated autoload severally but problem still persists, now I'm wondering if it has something to do with phpunit..

PR #4390, opened 2015-09-01
for github increasing the unauthenticated rate limit

jhyeon1010:
Some code for used github api
Section is [this](https://developer.github.com/v3/#increasing-the-unauthenticated-rate-limit-for-oauth-applications)

* Sample
```php
$io = new NullIO();
$config = Factory::createConfig();
$io->loadConfiguration($config);

$repository = new VcsRepository(['url' => $repoUrl, 'apiParam' => ['client_id' => 'Client ID', 'client_secret' => 'Client Secret']], $io, $config);
```

### DIFF
--- a/src/Composer/Repository/Vcs/VcsDriver.php
+++ b/src/Composer/Repository/Vcs/VcsDriver.php
@@ -28,6 +28,7 @@ abstract class VcsDriver implements VcsDriverInterface
 {
     protected $url;
     protected $originUrl;
+    protected $apiParam;
     protected $repoConfig;
     protected $io;
     protected $config;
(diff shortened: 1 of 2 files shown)


stof:
why not simply authenticating on github ?

jhyeon1010:
I'm making my own packagist using [composer/packagist](https://github.com/composer/packagist).
There is parameters for oauth in this package as "client_id" and "clien_secret".
I want to use these.

stof:
@jhyeon1010 the client id and client secret in the parameters.yml are of Packagist are here to handle the oauth login on the website.
for the composer authentication, it will rely on the auth.json file available on the server, and so you can pass it a token.

jhyeon1010:
@stof thank you for your answer.

alcohol:
I think this can be closed as it is not really relevant anymore?

PR #4391, opened 2015-09-01
less runs of substr in findFileWithExtension()

individual-it:
putting the substr outside of the foreach loop should save some substr runs

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -349,8 +349,8 @@ class ClassLoader
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {
             foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
-                $logicalPath=substr($logicalPathPsr4, $length);
                 if (0 === strpos($class, $prefix)) {
+                    $logicalPath=substr($logicalPathPsr4, $length);
                     foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
                         if (is_file($file = $dir . DIRECTORY_SEPARATOR . $logicalPath)) {
                             return $file;
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -349,9 +349,10 @@ class ClassLoader
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {
             foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
+                $logicalPath=substr($logicalPathPsr4, $length);
                 if (0 === strpos($class, $prefix)) {
                     foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
-                        if (file_exists($file = $dir . DIRECTORY_SEPARATOR . substr($logicalPathPsr4, $length))) {
+                        if (is_file($file = $dir . DIRECTORY_SEPARATOR . $logicalPath)) {
                             return $file;
                         }
                     }


src/Composer/Autoload/ClassLoader.php
```diff
@@ -349,9 +349,10 @@ private function findFileWithExtension($class, $ext)
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {
             foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
+                $logicalPath=substr($logicalPathPsr4, $length);
```
staabm: this could even be moved into the `if` below

individual-it: is $length not coming from the ````foreach```` line? ````as $prefix => $length````

staabm: It is. I am talking about the foreach 2 lines below

individual-it: you are right, I've changed it

src/Composer/Autoload/ClassLoader.php
```diff
@@ -350,8 +350,9 @@ private function findFileWithExtension($class, $ext)
         if (isset($this->prefixLengthsPsr4[$first])) {
             foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
                 if (0 === strpos($class, $prefix)) {
+                    $logicalPath=substr($logicalPathPsr4, $length);
                     foreach ($this->prefixDirsPsr4[$prefix] as $dir) {
-                        if (file_exists($file = $dir . DIRECTORY_SEPARATOR . substr($logicalPathPsr4, $length))) {
+                        if (is_file($file = $dir . DIRECTORY_SEPARATOR . $logicalPath)) {
```
staabm: why do you change `file_exists` with `is_file`?

stof: this change was already done in the past, and it was reverted because of the performance impact: https://github.com/composer/composer/pull/4327

Seldaek:
It's less substr, but it's one more variable allocated, and most of the time there is only one prefix=>dir mapping, so who knows if it's really beneficial. This sort of patch should come with serious benchmarks otherwise it's just guesswork and gut feelings.

alcohol:
Any intention of providing a benchmark, @individual-it ?

individual-it:
Could try on the basis of OwnCloud, that where I've found this.
Just didn't sound like there is a big change of this PR being taken so I haven't put in a lot of effort into it

alcohol:
Well it's a very minor change and the intent is purely on performance improvement. So the whole accept or deny aspect hinges on a benchmark. If you can prove that it improves performance, I see no reason not to accept your suggested change. Just be sure that your benchmark is not tied to a very specific use-case. Or if it is, at least also include a generic use-case for comparison (a complete Symfony setup for example).

Seldaek:
What is for sure is the is_file has to go back to file_exists, but the other change indeed depends on benchmark results.

stof:
Well, this would improve performance when having multiple paths for the same PSR-4 prefix (as the subpath would be computed only once).
It would not change anything for cases where you have a single path for the prefix (as the loop runs only once anyway). This means it won't improve the performance if you take care of using the most specific prefix possible for each of your packages.

The only case where the perf could degrade a little is when you have a prefix registered without any path, but this is really a weird edge case (and cannot happen if your ClassLoader is configured purely based on package metadata)

individual-it:
closing this, because I cannot find time to make a good performance analysis in the next time

PR #4414, opened 2015-09-13
Pre-validates composer packages before adding them to the pool

Nicofuma:
Pre validates the packages coming from a composer repo against the root package
constraints before adding them to the pool. In order to save memory and computation time.

`php /opt/composer.phar require 'symfony/translation:2.8.*@dev'`
[Blackfire comparison](https://blackfire.io/profiles/compare/13937aa9-540b-4aea-979d-107aa831d918/graph?callname=main()&selected=Composer%5CDependencyResolver%5CSolver%3A%3Asolve&settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes)

`A composer.json requiring mostly exact  versions, a more favourable case`
[Blackfire comparison](https://blackfire.io/profiles/compare/9b4b97a3-73bd-4d39-a741-f122d9b35787/graph?callname=Composer%5CDependencyResolver%5CSolver%3A%3Asolve&selected=&settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes)

Note: It might be useful in some other places but I don't know the composer source code well enough to be sure...

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -324,10 +324,11 @@ class ComposerRepository extends ArrayRepository
                         }
                     }
                 } else {
-                    if (!$pool->isPackageAcceptable(strtolower($version['name']), VersionParser::parseStability($version['version']))) {
+                    $loweredName = strtolower($version['name']);
+                    if (!$pool->isPackageAcceptable($loweredName, VersionParser::parseStability($version['version']))) {
                         continue;
                     }
-                    if (!$pool->doesPackageMatchRequires(strtolower($version['name']), $version['version'])) {
+                    if (!$pool->doesPackageMatchRequires($loweredName, $version['version'])) {
                         continue;
                     }
 
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -327,6 +327,9 @@ class ComposerRepository extends ArrayRepository
                     if (!$pool->isPackageAcceptable(strtolower($version['name']), VersionParser::parseStability($version['version']))) {
                         continue;
                     }
+                    if (!$pool->doesPackageMatchRequires(strtolower($version['name']), $version['version'])) {
+                        continue;
+                    }
 
                     // load acceptable packages in the providers
                     $package = $this->createPackage($version, 'Composer\Package\Package');
--- a/tests/Composer/Test/Repository/ComposerRepositoryTest.php
+++ b/tests/Composer/Test/Repository/ComposerRepositoryTest.php
@@ -148,6 +148,10 @@ class ComposerRepositoryTest extends TestCase
             ->method('isPackageAcceptable')
             ->will($this->returnValue(true));
 
+        $pool->expects($this->any())
+            ->method('doesPackageMatchRequires')
+            ->will($this->returnValue(true));
+
         $versionParser = new VersionParser();
         $repo->setRootAliases(array(
             'a' => array(
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -296,6 +296,16 @@ class Pool implements \Countable
         return $prefix.' '.$package->getPrettyString();
     }
 
+    public function doesPackageMatchRequires($name, $version)
+    {
+        if (isset($this->filterRequires[$name])) {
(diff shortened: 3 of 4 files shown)


Seldaek:
Thanks, that looks like a good optimization as far as I can see, I think it could also allow Pool::match() to be simplified, the `$requireFilter` definition if/else could be removed and then all the `                return $requireFilter->matches($pkgConstraint) ? self::MATCH : self::MATCH_FILTERED;
` can become simply `                return self::MATCH` if I'm not mistaken. Do you want to investigate that one as well?

Nicofuma:
I have tested that, but it breaks 2 functional tests: https://travis-ci.org/composer/composer/jobs/80186560#L353-L379

src/Composer/Repository/ComposerRepository.php
```diff
@@ -327,6 +327,9 @@ public function whatProvides(Pool $pool, $name)
                     if (!$pool->isPackageAcceptable(strtolower($version['name']), VersionParser::parseStability($version['version']))) {
                         continue;
                     }
+                    if (!$pool->doesPackageMatchRequires(strtolower($version['name']), $version['version'])) {
```
stloyd: You could move `strtolower($version['name']` too the another vartiable to save doing it twice.

Nicofuma: done

staabm: on line 327 there is another occurence of `strtolower($version['name'])`

Nicofuma: It's done but there is a bug with github and the PR isn't updated....

Nicofuma: I amended the commit and push again... should be good now...

tests/Composer/Test/Repository/ComposerRepositoryTest.php
```diff
@@ -148,6 +148,10 @@ public function testWhatProvides()
             ->method('isPackageAcceptable')
             ->will($this->returnValue(true));
 
+        $pool->expects($this->any())
+            ->method('doesPackageMatchRequires')
+            ->will($this->returnValue(true));
```
stof: you would need a test covering the behavior in case it does not return true

stof:
I see an issue here in the implementation (which already exists today because of the stability filtering but becomes much worse): the repository is caching providers for a given name. but the cached list now depends on the pool being passed as argument. This means that passing a different pool later, with different requirement filters would not produce the expected behavior anymore (if the first pool is more strict than the second one).

And using several pools can happen in several cases:

- the ``require`` command runs a first pool to find the package version, and then a second one to run the partial update
- the ``create-project`` uses similar things  (the first one does not have root filters set, as the root is not yet available)
- a better error reporter could also make use of this to look for other versions of the package which may have been filtered by the root requirement or stability filters, instead of saying that the package does not exist at all (this already does not work, because the second pool is less restrictive)

stof:
https://github.com/composer/composer/pull/4163 is a much better solution to this problem IMO

stof:
and your blackfire comparisons are flawed. You have not performed them with a clean composer cache, which can be seen from the list of HTTP requests being made. Your reference profile was reusing many files from the cache, while your compared profile had to redownload them (hence the 22% difference in IO time)

Nicofuma:
Actually the io doesn't matter in this case, only the memory and CPU (where you can see the improvement).

About your first comment, I see the issue... For me it's a bit weird to reuse the pool but I can see why. Just to say, about the last point, one of the main purpose here is to save memory by not loading a bunch of packages and a such error reporter would have the complete opposite purpose...

Seldaek:
Yes @stof has a good point, in this sense it's not a great idea, although we could also make sure we flush those caches when needed. The question is whether it will still be overall better in most cases when we blow up the cache or if that offsets the wins in amount of packages kept in memory.

stof:
@Nicofuma the pool is not reused. The *repository* is used with different pools, and the internal cache does not depend on the pool (this issue is solved in #4163 btw AFAICT)

naderman:
Indeed, I think we should move ahead with #4163 to solve this properly, rather than build more sort of workarounds that may have odd side effects.

Seldaek:
Closing in favor of https://github.com/composer/composer/pull/4163

PR #4417, opened 2015-09-14
Path repository store relative path

sroze:
This PR introduce two BC changes:
- [x] It updates the output of the `path` repository to be coherent with others
- [x] It stores the URL as it is in the `composer.json` in the `lock` file to be able to reuse the same `lock` file on different environments


Seldaek:
Makes sense thanks

PR #4418, opened 2015-09-14
Optimized this->getIO()

olvlvl:
Optimized the code by replacing a lot of `$this->getIO()` by `$io`.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -464,6 +464,7 @@ EOT
     protected function listConfiguration(array $contents, array $rawContents, OutputInterface $output, $k = null)
     {
         $origK = $k;
+        $io = $this->getIO();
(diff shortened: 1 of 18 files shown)


stof:
How much difference does it make ?

olvlvl:
I don't have metrics, it's more about code quality, and this is not good practice:

```php
foreach ($a as $b)
{
    $this->getIO()->doSomething();
}
```

Seldaek:
It's a CLI application, if you try to save nanoseconds in anything else than the solver, IMO it's a waste of time and has no real effect, even a getOption() in a loop is basically transparent, the command will take 1sec or .999sec to run, it is nothing noticeable, especially compared to the amount of i/o we typically do..

I'll merge because it's done and $io is used pretty much everywhere, but I'd rather not have a billion variables hanging around to save a few getFoo() calls.

PR #4419, opened 2015-09-14
Uses `realpath` to fix relative path issues

sroze:
Depending on the user configuration, the symlinked path may not be the good one, using `realpath` fixes the problem.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -39,18 +39,19 @@ class PathDownloader extends FileDownloader
         ));
 
         $url = $package->getDistUrl();
-        if (!file_exists($url) || !is_dir($url)) {
+        $realUrl = realpath($url);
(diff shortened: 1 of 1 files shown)


Seldaek:
Someone complained on IRC about the absolute paths breaking things in VMs. And I agree that relative paths would be better. I think the problem is just that the relative paths were created using the relative-to-root path, and not relative-to-vendor/foo/bar path.

Would you mind investigating whether https://github.com/composer/composer/blob/master/src/Composer/Util/Filesystem.php#L309 can help you (set $directories to true) to find the relative path from `vendor/foo/bar` to the `realpath($pathRepoPath)`? I think it should be possible with that to create proper relative paths that work everywhere.

sroze:
You know what @Seldaek? You just have to merge this one: https://github.com/composer/composer/pull/4426 :)

Seldaek:
Hah, I didn't see that, all good then ;)

PR #4422, opened 2015-09-15
Enable wildcards in path names for path repositories

dennisbirkholz:
I extended the path repositories feature to use glob to resolve wildcards. It would be nice if you consider it for inclusion.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -608,8 +608,8 @@ update to the latest version.
 
 ### Path
 
-In addition to the artifact repository, you can use the path one, which allow
-you to depends on a relative directory. This can be especially useful when dealing
+In addition to the artifact repository, you can use the path one, which allows
+you to depend on a relative directory. This can be especially useful when dealing
 with monolith repositories.
 
 For instance, if you have the following directory structure in your repository:
@@ -638,6 +638,9 @@ file, you can use the following configuration:
     }
 }
 ```
+
+Repository paths can also contain wildcards like ``*`` and ``?``.
+For details, see the [PHP glob function](http://php.net/glob).
 
 ## Disabling Packagist
 
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -639,8 +639,8 @@ file, you can use the following configuration:
 }
 ```
 
-Repository paths can also contain wildcards like ``*`` and ``?``.
-For details, see the [PHP glob function](http://php.net/glob).
+> **Note:** Repository paths can also contain wildcards like ``*`` and ``?``.
+> For details, see the [PHP glob function](http://php.net/glob).
 
 ## Disabling Packagist
 
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -101,8 +101,6 @@ class PathRepository extends ArrayRepository
     {
         parent::initialize();
 
-        $foundPackage = false;
-
         foreach ($this->getPaths() as $path) {
             $path = realpath($path) . '/';
 
@@ -111,7 +109,6 @@ class PathRepository extends ArrayRepository
                 continue;
             }
 
-            $foundPackage = true;
             $json = file_get_contents($composerFilePath);
             $package = JsonFile::parseJson($json, $composerFilePath);
             $package['dist'] = array(
@@ -131,7 +128,7 @@ class PathRepository extends ArrayRepository
             $this->addPackage($package);
         }
 
-        if (!$foundPackage) {
+        if (count($this) == 0) {
             throw new \RuntimeException(sprintf('No `composer.json` file found in any path repository in "%s"', $this->url));
         }
     }
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -38,6 +38,10 @@ use Composer\Util\ProcessExecutor;
  *     {
  *         "type": "path",
  *         "url": "/absolute/path/to/package/"
+ *     },
+ *     {
+ *         "type": "path",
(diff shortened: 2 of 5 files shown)


src/Composer/Repository/PathRepository.php
```diff
@@ -97,36 +101,39 @@ protected function initialize()
     {
         parent::initialize();
 
-        $path = $this->getPath();
-        $composerFilePath = $path.'composer.json';
-        if (!file_exists($composerFilePath)) {
-            throw new \RuntimeException(sprintf('No `composer.json` file found in path repository "%s"', $path));
```
stof: you should keep throwing an exception when none of the path contains a composer.json IMO. It makes it much easier to debug issues in the repository configuration rather than silently ignoring the path

dennisbirkholz:
@stof thx for the fast response. I updated the pull request accordingly.

barryvdh:
So this would allow just setting the path to: `modules/*/*` and find `modules/vendor/package1` and `modules/vendor2/package2`, right? Looks cool :)

dennisbirkholz:
@barryvdh That's how it is supposed to work. The used ``glob()`` function can use ``*`` and ``?`` as wildcards.

src/Composer/Repository/PathRepository.php
```diff
@@ -97,36 +101,46 @@ protected function initialize()
...
-     * @return string
+     * Get a list of all absolute path names matching the supplied urls
+     *
+     * @return string[]
      */
-    private function getPath()
+    private function getPaths()
     {
-        return realpath(rtrim($this->url, '/')) . '/';
+        return glob($this->url, GLOB_MARK|GLOB_ONLYDIR);
```
alcohol: I'm wondering if `glob()` is the way to go here or if perhaps an instance of `Finder` could be used. I think the latter could provide more flexibility when it comes to patterns?

dennisbirkholz:
@alcohol I didn't want to overcomplicate this. ``glob()`` just behaves like paths in a (linux) shell, so it would be very clear what is meant. My main intention here is to allow to use more than one folder as a repository. I am not sure if there is really a use case for using full blown regular expressions. But if that stands in the way to merge it, I can change it to use ``Finder``, no problem.

alcohol:
No it was just a question/suggestion, hence the `?`. It's definitely not blocking.

doc/05-repositories.md
```diff
@@ -639,6 +639,9 @@ file, you can use the following configuration:
 }
 ```
 
+Repository paths can also contain wildcards like ``*`` and ``?``.
+For details, see the [PHP glob function](http://php.net/glob).
```
alcohol: Maybe use the info markup box to make this stand out more?

src/Composer/Repository/PathRepository.php
```diff
@@ -97,36 +101,46 @@ protected function initialize()
...
-            $package['dist']['reference'] = trim($output);
-        }
-
-        $package = $this->loader->load($package);
-        $this->addPackage($package);
     }
 
     /**
-     * @return string
+     * Get a list of all absolute path names matching the supplied urls
```
alcohol: I'd go with something like...

> Get a list of all absolute path names matching given url (supports globbing).

src/Composer/Repository/PathRepository.php
alcohol: Another thing to point out here, you did change the behaviour from looking at an absolute path to relative paths. Not sure if intentional or overlooked?

dennisbirkholz:
@alcohol I intentionally avoided to walk all paths again and make them absolute. The ``PathDownloader`` class (which is the only consumer of path repositories imho) makes the paths absolute itself. So maybe there is no need to make the repository path absolute inside ``PathRepository``. But I can make the path absolute when storing it in the package data inside ``PathRepository::initialize()`` of course.

alcohol:
Personally I'm not really sure if absolute or relative is preferred. I know that absolute can lead to issues when dealing with VM's that have their code directory shared by the host system.

@seldaek any hint as to what is preferred (and why)?

alcohol:
Aside from the absolute vs relative note, this PR looks good to go to me.

stof:
@dennisbirkholz IIRC, paths were made absolute to fix a bug

src/Composer/Repository/PathRepository.php
```diff
@@ -97,36 +101,48 @@ protected function initialize()
...
-        $package = JsonFile::parseJson($json, $composerFilePath);
-        $package['dist'] = array(
-            'type' => 'path',
-            'url' => $this->url,
-            'reference' => '',
-        );
-
-        if (!isset($package['version'])) {
-            $package['version'] = $this->versionGuesser->guessVersion($package, $path) ?: 'dev-master';
+        if (!$foundPackage) {
```
Seldaek: Instead of keeping track of $foundPackages, this check could simply be `if (!$this->getPackages())`

dennisbirkholz:
@Seldaek I didn't dive too deep into the parent class's properties so I overlooked the obvious check whether packages where created or not. Personally I don't like to rely on the fact that empty array evaluate to false so I hope the fix is OK with you.

Seldaek:
Eh well, `count($this)` looks a bit confusing to me, I guess I'd prefer `count($this->getPackages())` because everyone gets what that means. The former looks like a bug unless you know it is Countable. But down the line all these solutions work, it's not that critical.

dennisbirkholz:
After I pushed I wondered if that would be the better solution but I didn't want to spam you with micro-commits. But here you go ;-)

Seldaek:
Thanks!

dennisbirkholz:
Thank you for building composer :-)

PR #4423, opened 2015-09-15
Fix check for flag in case RELATIVE_PATHS should be changed somehow

dennisbirkholz:
Just a minor fix in the flag check as "1" is the hardcoded value of RELATIVE_PATHS. Alternative the check could be ``$flags & self::RELATIVE_PATHS > 0``

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -180,7 +180,7 @@ class Config
                     return $val;
                 }
 
-                return ($flags & self::RELATIVE_PATHS == 1) ? $val : $this->realpath($val);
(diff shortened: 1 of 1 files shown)


alcohol:
I'm wondering, would simply `($flags & self::RELATIVE_PATHS)` be sufficient already?

dennisbirkholz:
@alcohol It would be, but personally I prefer checks that return a boolean explicitly, not implicitly. In any case, the ``= 1`` has to be changed.

alcohol:
Eh, I would prefer to go for legibility in this case. The > 0 just seems a bit odd. We don't care about the numerical value. We just care about the bit being set.

alcohol:
Sorry I derped. Your commit already did exactly what I was looking for. Got confused with your comment.

PR #3985, opened 2015-04-30
Check if package URLs are up to date after composer update is done

Seldaek:
Fixes #3214

PR #4428, opened 2015-09-18
Update source type and reference along with URLs

mpdude:
In #3985, the source URL for all repos is updated during the `update` task. However, the change did not include take the source type and reference.

So, when moving a repo from `hg` to `git`, only the new git URL was written to the .lock file but the repository type and commit ID were not updated.


Seldaek:
Thanks

mpdude:
Thank you @seldaek for your quick reaction

PR #4431, opened 2015-09-20
'path' repository : PathDownloader should be a no-op if source == target

yched:
If the local package is already in its final install location, PathDownloader should not remove it and re-symlink it.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -29,8 +29,13 @@ class PathDownloader extends FileDownloader
      */
     public function download(PackageInterface $package, $path)
     {
-        $fileSystem = new Filesystem();
-        $this->filesystem->removeDirectory($path);
+        $url = $package->getDistUrl();
+        $realUrl = realpath($url);
+        $realPath = realpath($path);
+
+        if ($realUrl !== $realPath) {
+            $this->filesystem->removeDirectory($path);
+        }
 
         $this->io->writeError(sprintf(
             '  - Installing <info>%s</info> (<comment>%s</comment>)',
@@ -38,8 +43,6 @@ class PathDownloader extends FileDownloader
             $package->getFullPrettyVersion()
         ));
 
-        $url = $package->getDistUrl();
-        $realUrl = realpath($url);
         if (false === $realUrl || !file_exists($realUrl) || !is_dir($realUrl)) {
             throw new \RuntimeException(sprintf(
                 'Path "%s" is not found',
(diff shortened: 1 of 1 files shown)


sroze:
Why do you need to install a dependency in the dependency path ? Can't you just use the autoload configuration ?

yched:
Because that local package has dependencies of its own in its local composer.json 'require', and you need those dependencies merged and resolved and downloaded along with the dependencies of the other packages.

Nothing to do in the "installation / download" phase for this package, the code is already in the right place, but you still need its composer.json processed like those of any other non-local package used by the project.

Seldaek:
Can you show an example of the root compoesr.json and the linked/path-repo composer.json? I don't quite get what you mean. It sounds to me like maybe what you need is just to use a provide or a replace in the root composer.json instead of defining a path repo, but maybe I'm mistaken.

alcohol:
It sounds like he has a local repository in `vendor/$vendor/$package`, and now wants to use a path repository to install that. But the path repository will overwrite that directory with a symlink to itself (how does that even work?)? Basically, I guess he is trying to develop inside his vendor directory  but now wants to install from his vendor directory to his vendor directory..

sroze:
Yes, that's what I understood too, and that's why I would like to know a bit more what you are doing @yched. Because I'm not sure that I agree with this kind of setup and I'm not sure at all that the `path` repository is the solution to your problem.

sroze:
ping @yched.

curry684:
@yched #4851 was just merged to throw an exception instead of a no-op if source and target are 'intertwined', which is a lot safer, can never cause recursive symlinks, or the worst scenario accidental deletion of new sources. Are you satisfied with that solution?

curry684:
I understand your use case now, and while that means this issue is definitely not resolved by my PR I think the requested feature was also never really a solution to your issue. The exception is needed to avoid many dangerous situations, like a path repo to `~/vendor/foo/bar/acme` containing the package `foo/bar`. An update would effectively cause an `rm -rf ~/vendor/foo/bar` followed by an install from... oops. The path repositories just were never intended for use cases like this, and I stand behind 'our' choice to now completely block it being used like that.

Your clarification boils down to a 'composer submodule' analogous to git's submodules - a child project seamlessly merged into its container. There are similar feature requests for Composer, some calling for `composer.json` inheritance, others for extension, and others for includes. To a certain extent, they all boil down to the same thing - multiple `composer.json` files being merged into one master definition leading to an installation and a single autoloader.

Nearly everything I read on the subject has been surprisingly quiet on the most dangerous aspect of them all - security. Unchecked inclusion of `composer.json` files from external contributors is essentially a wide open arbitrary code execution exploit, or worse a privilege escalation.

The Wikimedia merge plugin has tackled this to a certain extent by only merging relatively harmless fields of the composer file, but a malicious update far downstream the dependency tree could still do major damage. Their README states with good reason that it was intended to `to allow easier dependency management for applications which ship a composer.json file and expect some deployments to install additional Composer managed libraries`. In other words - to allow experienced developers a customized MediaWiki install with extra dependencies without having to modify the installed project package. Essentially they implemented inheritance, not submodules, although it can serve that purpose equally well in a fully trusted deployment. I would never in my right mind allow a Drupal end user randomly select Packagist packages for inclusion in the root composer.json though.

I think if someone were to actually write a good proposal tackling **all** those use cases and the issues stemming from them, it would definitely be considered developing on this end.

alcohol:
I think the biggest issue is so many frameworks still relying on fixed paths for installing "plugins", "modules" or other such "components". Configuration through filesystem layout is a very bad architectural choice in my personal opinion.

PR #4432, opened 2015-09-20
Added ability to locate classes/resources in various file extensions

eserozvataf:
I'm planning to use composer to resolve resource paths for my components / bundles.

`ClassLoader`'s `findFileWithExtension` method would be the best case for my scenario, but it's marked as private. So I decided to extend `findFile`s functionality to allow multiple filename extensions apart from/includes '.php'.

e.g.:
```php
$config = $classLoader->findFile('Scabbia\SampleBundle\bundle', ['.json', '.yaml']);
```

The second parameter is optional and it's default is still `['.php']` (or `['.php', '.hh']` under HHVM). That means this modification is backward compatible for existing API.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -53,11 +53,20 @@ class ClassLoader
 
     private $useIncludePath = false;
     private $classMap = array();
+    public $extensions = array('.php');
 
     private $classMapAuthoritative = false;
 
+    public function __construct()
+    {
+        if (defined('HHVM_VERSION')) {
+            $this->extensions[] = '.hh';
+        }
+    }
+
     public function getPrefixes()
     {
+        // Search for Hack files if we are running on HHVM
         if (!empty($this->prefixesPsr0)) {
             return call_user_func_array('array_merge', $this->prefixesPsr0);
         }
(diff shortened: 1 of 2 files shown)


stof:
This break things. If you call ``$classLoader->findFile('Scabbia\SampleBundle\bundle', ['.json', '.yaml']);``, it will then break the loading of a ``Scabbia\SampleBundle\bundle`` in subsequent calls due to the optimizations performed in the class loader (remembering whether the class was found or no). This will happen any time you call the method again with the same class but a different set of extensions.

And your use case is a clear abuse of the ClassLoader here (you are *not* passing it a class). IMO, you should rather look into using [Puli](http://puli.io) for your need

eserozvataf:
Yeah, my use case is not flawless and definitely this is not responsibility of ClassLoader but I'm still thinking / working on how I can use the power of composer for this. Let's say it's a practical research / prototype for now.

Composer is well-optimized and well-maintained component. My reasoning was duplication of its functionality will/would be poor decision.

However, this PR is completely apart from my use case, it's still makes sense for class loading if there are various file extensions.

Seldaek:
This definitely falls more in the puli use case, or at least you should just grab the autoloader configuration and do what you want with it, but making the autoloader more complex to support such exotic use cases is a no go for us sorry.

PR #4444, opened 2015-09-23
Fix #4439

### LINKED ISSUE #4439: Absolute instead of relative path in lock file
I updated Composer today and it seems like a recent commit changed the behaviour of the path repository. Previously the relative path to the package was stored in the lock file, this is reflected in the `composer update` output as well:

```
- Installing vendor/package (dev-develop)
    Symlinked from ./components/package
```

`composer.lock` looks like this:

```
"dist": {
    "type": "path",
    "url": "./components/package",
    "reference": "",
    "shasum": null
},
```
Now if I update Composer and remove an already installed "path package" (or add a new one) it'll use the absolute path instead, once again reflected in the output:

```
- Installing vendor/package (dev-develop)
    Symlinked from /absolute/path/to/components/package/
```

`composer.lock` looks like this:

```
"dist": {
    "type": "path",
    "url": "/absolute/path/to/components/package/",
    "reference": "",
    "shasum": null
},
```

This makes it impossible to do a `composer install` of the project on another machine, since Composer will complain that it can't find the absolute path. I've been trying to figure out the cause, but with Composer being a phar and all it's not so straightforward (for me at least).

Is this change of behaviour intended or did it accidentally make it into the code? Makes it kind of hard to use path repositories for anything that lives on more than one machine :)

Thanks!

dennisbirkholz:
See bug #4439

### DIFF
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -101,10 +101,10 @@ class PathRepository extends ArrayRepository
     {
         parent::initialize();
 
-        foreach ($this->getPaths() as $path) {
-            $path = realpath($path) . '/';
-
+        foreach ($this->getUrlMatches() as $url) {
+            $path = realpath($url) . '/';
             $composerFilePath = $path.'composer.json';
+            
             if (!file_exists($composerFilePath)) {
                 continue;
             }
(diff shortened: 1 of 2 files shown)


sroze:
@Seldaek looks :+1:

Seldaek:
Thanks

PR #4446, opened 2015-09-24
Switch to semver dep

alcohol:


src/Composer/Command/InitCommand.php
```diff
@@ -16,10 +16,10 @@
 use Composer\Json\JsonFile;
 use Composer\Factory;
 use Composer\Package\BasePackage;
+use Composer\Package\Version\VersionParser;
```
Nicofuma: shouldn't it be `Composer\Semver\VersionParser`?

src/Composer/Package/LinkConstraint/EmptyConstraint.php
```diff
@@ -12,36 +12,13 @@
...
+@trigger_error('The ' . __NAMESPACE__ . '\EmptyConstraint class is deprecated, use Composer\Semver\Constraint\EmptyConstraint instead.', E_USER_DEPRECATED);
+
 /**
- * Defines an absence of constraints
- *
- * @author Jordi Boggiano <j.boggiano@seld.be>
+ * @deprecated use Composer\Semver\Constraint\EmptyConstraint instead
  */
-class EmptyConstraint implements LinkConstraintInterface
+class EmptyConstraint extends BaseEmptyConstraint
```
stof: it still needs to implement the LinkConstraintInterface, otherwise it is not properly BC

src/Composer/Package/LinkConstraint/MultiConstraint.php
```diff
@@ -12,72 +12,13 @@
...
+
 /**
- * Defines a conjunctive or disjunctive set of constraints on the target of a package link
- *
- * @author Nils Adermann <naderman@naderman.de>
- * @author Jordi Boggiano <j.boggiano@seld.be>
+ * @deprecated use Composer\Semver\Constraint\MultiConstraint instead
  */
-class MultiConstraint implements LinkConstraintInterface
+class MultiConstraint extends BaseMultiConstraint
```
stof: same here. the old interface should be implemented

src/Composer/Package/Version/VersionGuesser.php
```diff
@@ -15,6 +15,7 @@
 use Composer\Config;
 use Composer\Repository\Vcs\HgDriver;
 use Composer\IO\NullIO;
+use Composer\Semver\VersionParser as BaseVersionParser;
```
stof: I would rather name it ``SemverVersionParser`` than ``BaseVersionParser``. It is not used as a base class here

src/Composer/Package/Version/VersionParser.php
```diff
@@ -12,513 +12,14 @@
...
-
     /**
-     * Parses a name/version pairs and returns an array of pairs + the
+     * @param array $pairs
      *
-     * @param  array   $pairs a set of package/version pairs separated by ":", "=" or " "
-     * @return array[] array of arrays containing a name and (if provided) a version
+     * @return array[]
      */
     public function parseNameVersionPairs(array $pairs)
```
stof: this should probably be moved to a dedicated class, to be able to deprecate this VersionParser class

src/Composer/Package/Version/VersionParser.php
```diff
@@ -12,513 +12,14 @@
...
-                return 'patch';
-            case 'rc':
-                return 'RC';
-            default:
-                return $stability;
-        }
-    }
-
     /**
-     * Parses a name/version pairs and returns an array of pairs + the
```
stof: the descriptiojn of the method should be kept

src/Composer/Package/Version/VersionParser.php
```diff
@@ -12,513 +12,14 @@
...
+     * @param array $pairs
      *
-     * @param  array   $pairs a set of package/version pairs separated by ":", "=" or " "
-     * @return array[] array of arrays containing a name and (if provided) a version
```
stof: the description of arguments and returned value should be kept

src/Composer/Package/Version/VersionParser.php
alcohol: This description makes no sense though. Did you ever actually read it? I think someone copied pasta'd 2 sentences together and then forgot to make a proper sentence out of it.

src/Composer/Package/Version/VersionParser.php
alcohol: I discussed that with @Seldaek but to minimize the amount of changes this PR introduces we felt it was best to keep it here for now.

PR #4455, opened 2015-09-28
Enable command line configuration for archive-dir and archive-format

samsaggace:
Those 2 parameters could be defined inside the config.json but were not accepted by the command 'composer config'.
So, just adding them to the known uniqueConfigValues

The doc on archive-dir is also improved with an example of configuration;


alcohol:
Thank you.

PR #4467, opened 2015-09-30
Put the obvious solution first

fulldecent:


alcohol:
I'm sorry but merely moving text around does not really provide any new or added value.

Issue #4397, opened 2015-09-03
Add --file parameter to the archive command.


tobiastom:
This is kind of related to #4379. As the filename is more or less hard to guess, I'd like to add a parameter `--file` that will take the name that should be used for the archive. 

With this it would be easier to automate the next step that could happen after an archive is created (e.g. by a build server).

I could also add a pull request, I just want to check if this would get accepted.

Seldaek:
Sounds good to me, if you want to PR it.

MakiCode:
I can make a PR for this.

tobiastom:
@MakiCode That would be awesome. I'm burried in tasks right now.

Seldaek:
Done in #4479

tobiastom:
Thank you @MakiCode. Seems to work fine for me!

MakiCode:
@tobiastom You're welcome!

PR #4479, opened 2015-10-05
Added --file flag: Feature#4397

MakiCode:
Add --file option to the archive command, as per issue #4397 . Added 2 tests for this in the Composer\Test\Package\Archiver\ArchiveManagerTest class: ::testArchiveCustomFileName, and ::testNonStringFileName.

### DIFF
--- a/src/Composer/Command/ArchiveCommand.php
+++ b/src/Composer/Command/ArchiveCommand.php
@@ -42,6 +42,7 @@ class ArchiveCommand extends Command
                 new InputArgument('version', InputArgument::OPTIONAL, 'A version constraint to find the package to archive'),
                 new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Format of the resulting archive: tar or zip'),
                 new InputOption('dir', false, InputOption::VALUE_OPTIONAL, 'Write the archive to this directory'),
+                new InputOption('file', false, InputOption::VALUE_OPTIONAL, 'Write the archive with the given file name.'),
             ))
             ->setHelp(<<<EOT
 The <info>archive</info> command creates an archive of the specified format
@@ -78,7 +79,8 @@ EOT
             $input->getArgument('package'),
             $input->getArgument('version'),
             $input->getOption('format'),
-            $input->getOption('dir')
+            $input->getOption('dir'),
+            $input->getOption("file")
         );
 
         if (0 === $returnCode && $composer) {
@@ -88,7 +90,7 @@ EOT
         return $returnCode;
     }
 
-    protected function archive(IOInterface $io, Config $config, $packageName = null, $version = null, $format = 'tar', $dest = '.')
+    protected function archive(IOInterface $io, Config $config, $packageName = null, $version = null, $format = 'tar', $dest = '.', $fileName = null)
     {
         $factory = new Factory;
         $downloadManager = $factory->createDownloadManager($io, $config);
@@ -105,7 +107,7 @@ EOT
         }
 
         $io->writeError('<info>Creating the archive into "'.$dest.'".</info>');
-        $packagePath = $archiveManager->archive($package, $format, $dest);
+        $packagePath = $archiveManager->archive($package, $format, $dest, $fileName);
         $fs = new Filesystem;
         $shortPath = $fs->findShortestPath(getcwd(), $packagePath, true);
 
--- a/src/Composer/Command/ArchiveCommand.php
+++ b/src/Composer/Command/ArchiveCommand.php
@@ -40,9 +40,9 @@ class ArchiveCommand extends Command
             ->setDefinition(array(
                 new InputArgument('package', InputArgument::OPTIONAL, 'The package to archive instead of the current project'),
                 new InputArgument('version', InputArgument::OPTIONAL, 'A version constraint to find the package to archive'),
-                new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Format of the resulting archive: tar or zip'),
-                new InputOption('dir', false, InputOption::VALUE_OPTIONAL, 'Write the archive to this directory'),
-                new InputOption('file', false, InputOption::VALUE_OPTIONAL, 'Write the archive with the given file name.'
+                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the resulting archive: tar or zip'),
+                new InputOption('dir', false, InputOption::VALUE_REQUIRED, 'Write the archive to this directory'),
+                new InputOption('file', false, InputOption::VALUE_REQUIRED, 'Write the archive with the given file name.'
                     .' Note that the format will be appended.'),
             ))
             ->setHelp(<<<EOT
@@ -81,7 +81,7 @@ EOT
             $input->getArgument('version'),
             $input->getOption('format'),
             $input->getOption('dir'),
-            $input->getOption("file")
+            $input->getOption('file')
         );
 
         if (0 === $returnCode && $composer) {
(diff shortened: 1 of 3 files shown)


src/Composer/Command/ArchiveCommand.php
```diff
@@ -78,7 +80,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $input->getArgument('package'),
             $input->getArgument('version'),
             $input->getOption('format'),
-            $input->getOption('dir')
+            $input->getOption('dir'),
+            $input->getOption("file")
```
alcohol: Single quotes please, consistency.

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -96,16 +96,21 @@ public function getPackageFilename(PackageInterface $package)
      *
      * @param  PackageInterface          $package   The package to archive
      * @param  string                    $format    The format of the archive (zip, tar, ...)
-     * @param  string                    $targetDir The diretory where to build the archive
+     * @param  string                    $targetDir The directory where to build the archive
+     * @param  string|null               $fileName  The relative file name to use for the archive, or null to generate
+     *                                   the package name. Note that the format will be appended to this name
```
alcohol: Continuing indentation is wrong here.

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -96,16 +96,21 @@ public function getPackageFilename(PackageInterface $package)
...
      * @throws \RuntimeException
      * @return string                    The path of the created archive
      */
-    public function archive(PackageInterface $package, $format, $targetDir)
+    public function archive(PackageInterface $package, $format, $targetDir, $fileName = null)
     {
         if (empty($format)) {
             throw new \InvalidArgumentException('Format must be specified');
         }
+        if(!is_null($fileName) && !is_string($fileName)) {
```
alcohol: Space after if.

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -122,7 +127,11 @@ public function archive(PackageInterface $package, $format, $targetDir)
         }
 
         $filesystem = new Filesystem();
-        $packageName = $this->getPackageFilename($package);
```
alcohol: Same

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -122,7 +127,11 @@ public function archive(PackageInterface $package, $format, $targetDir)
...
+        if(null === $fileName) {
```
alcohol: Same.

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
```diff
@@ -55,9 +60,42 @@ public function testArchiveTar()
...
+
+        $package = $this->setupPackage();
+
+        $this->manager->archive($package, 'tar', $this->targetDir, array());
+    }
+
+    protected function getTargetName(PackageInterface $package, $format, $fileName = null)
     {
-        $packageName = $this->manager->getPackageFilename($package);
+        if(null === $fileName) {
```
alcohol: Space after if.

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
```diff
@@ -55,9 +60,42 @@ public function testArchiveTar()
...
+
+        $this->assertFileExists($target);
+
+        $tmppath = sys_get_temp_dir().'/composer_archiver/'.$this->manager->getPackageFilename($package);
+        $this->assertFileNotExists($tmppath);
+
+        unlink($target);
+    }
+
+    public function testNonStringFileName() {
```
alcohol: Opening `{` should be on the next line.

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
```diff
@@ -55,9 +60,42 @@ public function testArchiveTar()
...
+    {
+        $this->setupGitRepo();
+
+        $package = $this->setupPackage();
+
+        $fileName =  "testArchiveName";
+
+        $this->manager->archive($package, 'tar', $this->targetDir, $fileName);
+
+        $target =  $this->targetDir . "/" . $fileName . ".tar";
```
alcohol: Double space after `=`?

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
```diff
@@ -55,9 +60,42 @@ public function testArchiveTar()
...
     }
 
-    protected function getTargetName(PackageInterface $package, $format)
+    public function testArchiveCustomFileName()
+    {
+        $this->setupGitRepo();
+
+        $package = $this->setupPackage();
+
+        $fileName =  "testArchiveName";
```
alcohol: Double space after `=`?

src/Composer/Command/ArchiveCommand.php
```diff
@@ -42,6 +42,8 @@ protected function configure()
                 new InputArgument('version', InputArgument::OPTIONAL, 'A version constraint to find the package to archive'),
                 new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Format of the resulting archive: tar or zip'),
                 new InputOption('dir', false, InputOption::VALUE_OPTIONAL, 'Write the archive to this directory'),
+                new InputOption('file', false, InputOption::VALUE_OPTIONAL, 'Write the archive with the given file name.'
```
alcohol: How can the value be optional?

src/Composer/Package/Archiver/ArchiveManager.php
stof: and it should use ``null !==`` for consistency too

src/Composer/Command/ArchiveCommand.php
stof: it cannot. Using ``--file`` does not make sense (btw, other options are also wrong currently)

src/Composer/Package/Archiver/ArchiveManager.php
MakiCode: What is the correct indentation?

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -96,16 +96,21 @@ public function getPackageFilename(PackageInterface $package)
...
      * @return string                    The path of the created archive
      */
-    public function archive(PackageInterface $package, $format, $targetDir)
+    public function archive(PackageInterface $package, $format, $targetDir, $fileName = null)
     {
         if (empty($format)) {
             throw new \InvalidArgumentException('Format must be specified');
         }
+        if (null !== $fileName && !is_string($fileName)) {
+            throw new \InvalidArgumentException('fileName must be a string');
```
alcohol: Can you make it just 'Filename' or 'File' in the error message here? There is no point in mimicking the internal variable name since it is never exposed to the user (they only know it as `--file`).

Also I personally think adding a string check is overkill. I don't see how anyone would pass anything else from the command line. And you don't have to worry about the internal API doing things wrong.

Personally I would just drop the whole if statement + exception.

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
```diff
@@ -55,9 +60,43 @@ public function testArchiveTar()
...
+
+        $fileName = 'testArchiveName';
+
+        $this->manager->archive($package, 'tar', $this->targetDir, $fileName);
+
+        $target = $this->targetDir . '/' . $fileName . '.tar';
+
+        $this->assertFileExists($target);
+
+        $tmppath = sys_get_temp_dir().'/composer_archiver/'.$this->manager->getPackageFilename($package);
```
alcohol: This reveals too much of the internal workings, not a good approach (cause if it changes this test will fail and it might not be immediately clear why it fails). I would recommend just string replacing the custom filename with the otherwise expected default name in the previous asserted full path string.

alcohol:
If you could rephrase the error message (or drop it all-together), I'll mark this as reviewed so it can be merged soon :-)

src/Composer/Package/Archiver/ArchiveManager.php
MakiCode: Dropped!

MakiCode:
@alcohol Done!

alcohol:
Thanks :-)

Seldaek:
Thanks

PR #4482, opened 2015-10-06
Prevent plugin classes to be loaded twice

Soullivaneuh:
Fixes https://github.com/Soullivaneuh/composer-versions-check/issues/9.
Fixes https://github.com/Soullivaneuh/composer-versions-check/issues/23.

Tells composer to not load an already loaded plugin class.

Concrete case with `composer/dummy-plugin`:

Require `composer/dummy-plugin` globally:

```bash
composer global require composer/dummy-plugin
```

Then require it again locally on a random composer project:

```bash
composer require composer/dummy-plugin
```

You will get a "Cannot redeclare class" php fatal error.

This PR prevent this issue.

See also my plugin issue for another case.

cc @OskarStark @alcohol

### DIFF
--- a/src/Composer/Command/Command.php
+++ b/src/Composer/Command/Command.php
@@ -41,16 +41,17 @@ abstract class Command extends BaseCommand
     /**
      * @param  bool              $required
      * @param  bool              $disablePlugins
+     * @param  array             $plugins        Plugin instances to be set before loading
      * @throws \RuntimeException
      * @return Composer
      */
-    public function getComposer($required = true, $disablePlugins = false)
+    public function getComposer($required = true, $disablePlugins = false, array $plugins = array())
     {
         if (null === $this->composer) {
             $application = $this->getApplication();
             if ($application instanceof Application) {
                 /* @var $application    Application */
-                $this->composer = $application->getComposer($required, $disablePlugins);
+                $this->composer = $application->getComposer($required, $disablePlugins, $plugins);
             } elseif ($required) {
                 throw new \RuntimeException(
                     'Could not create a Composer\Composer instance, you must inject '.
--- a/src/Composer/Command/RequireCommand.php
+++ b/src/Composer/Command/RequireCommand.php
@@ -142,8 +142,9 @@ EOT
         $authoritative = $input->getOption('classmap-authoritative') || $composer->getConfig()->get('classmap-authoritative');
 
         // Update packages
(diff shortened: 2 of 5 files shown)


alcohol:
No this will not work.

alcohol:
It's easier and safer to check if the `$plugin` is already in the array returned by `$this->getPlugins()` (or `$this->plugins` directly).

alcohol:
This should definitely trigger a warning also in the console output.

alcohol:
Actually it already has a check it seems...

```
        if (in_array($package->getName(), $this->registeredPlugins)) {
            return;
        }
```

Soullivaneuh:
> Actually it already has a check it seems...

Well, seems to not be functional.

Note: This issue appear only with require command, not update.

Soullivaneuh:
> This should definitely trigger a warning also in the console output.

When? If class is already loaded?

stof:
> Note: This issue appear only with require command, not update.

then it may be worth investigating the difference between both commands to know why only one of them is affected

alcohol:
`$this->resetComposer()` in require seems like a possible culprit..

Soullivaneuh:
> `$this->resetComposer()` in the `RequireCommand` seems like a possible culprit..

The question is: Why this method is called?

alcohol:
Well the question first should be, am I right? ;-)

Sorry, I haven't tested this and I'm pre-occupied for the time being. Maybe I'll have some time tonight.

Soullivaneuh:
Ok I'll ping here if I got time to investigate too. :+1:

Soullivaneuh:
BTW @alcohol:

> No this will not work.

You didn't explain me why. :-)

alcohol:
Well for starters, you broke the test-suite.

Soullivaneuh:
I found an another way.

In a nutshell, I get the activated plugin before reset, and I put it again on the new composer instance.

Seems to work.

alcohol:
Okay, but I'm leaving this one in the hands of @seldaek -- he knows more about the plugin system than me.

Soullivaneuh:
Okay, thanks for your time. :+1:

Seldaek:
Ok  after looking into this I see what happens.. the regex handling this case broke for you because you have `final class foo` and not just `class foo`. Should be fixed now.

Soullivaneuh:
So the problem was just because my class is final? Well, if I knew it...

Thanks for the bugfix. :+1:

PR #4502, opened 2015-10-13
php-cs-fixer

alcohol:
Added some rules we already follow but don't check for yet. Ran fixers.


src/Composer/DependencyResolver/Solver.php
```diff
@@ -283,6 +284,7 @@ private function revert($level)
      * returns the new solver level or 0 if unsolvable
      *
      */
+
```
stof: making the previous comment a recognized phpdoc would be better IMO

alcohol: Didn't catch that, sorry.

PR #4504, opened 2015-10-13
remove dead code/class: DebugSolver

alcohol:
This file has not been updated in a long time and seems to contain mostly dead or no longer compatible code.

/cc @naderman can you confirm this?


naderman:
Are these couple bytes really an issue? That thing was pretty useful when debugging in the solver before, and I would expect to use it again when making changes?

naderman:
Well i guess I an always just get it from git history :-)

alcohol:
Well hence I asked, but @Seldaek was a bit quick :d

It was referencing a lot of non-existant fields/properties though.

Seldaek:
Yeah it didn't seem to work at all so I figured it was probably not needed.. We can revert though and update it instead? Up to you :)

PR #4508, opened 2015-10-14
Improve PHPDoc for package requires

Soullivaneuh:


alcohol:
Please use `/** Link[] */` for the class properties.

Soullivaneuh:
> Please use `/** Link[] */` for the class properties.

I'm sorry but I don't think this is the right convention for this.

BTW, this will be not managed by cs-fixer tools (if you use it) AFAIK.

Do you have a link about that? I'm curious. :-)

alcohol:
It's fine.

staabm:
@Soullivaneuh I guess he meant write it on a single line and forgot the `@var`

alcohol:
I did not forget `@var`... :trollface:

staabm:
:-)

Soullivaneuh:
> I guess he meant write it on a single line and forgot the `@var`

No, bad copy paste apparently. :-)

> It's fine.

Fine what? :-)

alcohol:
It's fine to put it on one line.

src/Composer/Package/Package.php
```diff
@@ -43,10 +43,17 @@ class Package extends BasePackage
     protected $stability;
     protected $notificationUrl;
 
+    /**
+     * @var Link[]
+     */
```
stof: this should also be done for other properties (conflicts, provides and replaces are also package links)

Soullivaneuh: This PR was for requirement properties. But yeah I can do that.

@stof In the same time, could I please have your opinion about @alcohol comment? https://github.com/composer/composer/pull/4508#issuecomment-149172694

stof: Well, I don't really care whether it is single line or multiline here? It is a matter of preferences IMO

alcohol: I personally find it a bit ridiculous to use 30 lines of comments to document 10 lines of class properties, when you can achieve the same effect using just 10 lines.

Soullivaneuh:
Where are you with this PR? Should I do another things? Document each property as @stof asked?

alcohol:
If you could do that, as one-liners, I'll merge it for you, sure.

Soullivaneuh:
Will do this on a new PR. Lost the repository of this one.

Soullivaneuh:
@alcohol Please see #4676.

PR #4509, opened 2015-10-14
note about Bitbucket repository slug vs. name in docs

bedwards:
Adding a note to the docs about Bitbucket repository naming. A user was tripped up by this when adding a Bitbucket webhook pointing to https://packagist.org/.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -292,6 +292,10 @@ GitHub and BitBucket:
 
 The only requirement is the installation of SSH keys for a git client.
 
+> **Note:** Use the Bitbucket repository slug (aka full_name) and not the
+> repository name in your composer configuration. The repository slug is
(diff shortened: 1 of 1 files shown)


alcohol:
I'm sorry, but can you elaborate? I have a bitbucket account, but I have no clue what this note is referring to. I'm pretty sure you should just use the clone url just like with github repositories. I think this note might just confuse readers.

bedwards:
Yes, 'use the clone URL' is probably the best way to phrase this. The potential confusion is for repository names that contain uppercase characters or forward slashes. Those are converted to lowercase and dashes in the clone URL and URLs on the site. This tripped up one of our users.

Seldaek:
Yeah I think this needs to be clarified or closed because I feel like it adds more confusion than it removes as it is.

Issue #2384, opened 2013-10-30
Support additionally composer config setting keys


pschirch:
Hi,

we actually automate our devops processes. In this case we have too manipulate `composer.json` settings. Unfortunately

```bash
$ composer config [setting-key] [setting-value]
```

does not support the following needed setting-keys:

* prefer-stable
* minimum-stability
* extra.symfony-assets-install

```bash
[InvalidArgumentException]
Setting prefer-stable does not exist or is not supported by this command
```

It would be a pleasure, if the composer config command accepts all possible setting keys.

Any chance to implement this?

spengilley:
Is this still required? Very old issue so might be worth closing if not...

curry684:
I think the current stance should be "every scripting language in the world supports easy JSON modification so let's not complicate the Composer codebase with it".

alcohol:
I still think it would be desirable to add support for those keys through the `config` command.

Seldaek:
Yeah but there is already #2415 implementing some of it.. probably needs to be added to the config command somehow instead of a new command though. It'll need some work as it's so old. After 1.0 is out it'll be time to bring all these old PRs back to life or close them for good :)

alcohol:
So before 1.0 you'd rather leave this open? Or is this something that would be welcome in 1.0 also?

Seldaek:
No we're in beta we shouldn't make any big changes that aren't essential.

Seldaek:
Fixed by 135783299af0281db918c103cceb2b202ae154f2

PR #2415, opened 2013-11-11
added new command for changing property settings

frastel:
As of #1411 I have added a new Command for changing simple properties like
- name
- description
- homepage
- minimum-stability (**not** stability)
- license

Example calls:
```
composer property-set name "foo/bar"
composer property-set minimum-stability "beta"
```

Things to discuss:
- I re-used some logic from `RequireCommand` and moved it to `Command\Helper\FileHelper`: Should be `RequireCommand` adapted accordingly?
- I tried to test several tests for this new command. Currently not many test cases for commands are existing and so I want to go sure if the used workflow is OK for you

PR #4514, opened 2015-10-17
Issue #2384 allow 'prefer-stable' and 'minimum-stability' to be set 

sneako:
Now allows 'prefer-stable' and 'minimum-stability' to be set via composer config command

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -322,6 +322,11 @@ EOT
             'classmap-authoritative' => array($booleanValidator, $booleanNormalizer),
             'prepend-autoloader' => array($booleanValidator, $booleanNormalizer),
             'github-expose-hostname' => array($booleanValidator, $booleanNormalizer),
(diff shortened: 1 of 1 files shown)


Seldaek:
This does not work, as minimum-stability and prefer-stable are *not* part of the config key but top level keys. Reverting.

Seldaek:
The proper fix for this would be #2415 which is waaay overdue to be reviewed and merged :/

alcohol:
@seldaek got any notes/comments regarding that one, or would you just like to see it rebased and reviewed?

Seldaek:
A rebase/review would be a good start, I don't remember what I think of it to be honest, I think it's probably fine to have as simply `composer property x y`, I am wondering if it shouldn't be merged in config though because I guess most people don't know what property is where (as evidenced by this PR and the fact it was merged :P).

alcohol:
Yeah, well.... in my defense, it seemed like it should just work :p

Seldaek:
Sure did, but those are the sneakiest PRs, the ones you think are 
obviously fine you should watch especially for :p

Issue #4512, opened 2015-10-16
Repository type path do not update


joelwurtz:
I'm using the new repository type path and any update done to the composer.json of the dependency is not used when updating the root library, i.e : 

composer.json of root package

```
{
    "require": {
         "vendor/my-lib": "*@dev"
    }
    "repositories": [
        {
            "type": "path",
            "url": "../my-lib"
        }
    ],
}
```

composer of my-lib : 

```
{
     "name": "vendor/my-lib",
     "require": 

     ...
}
```

It work fine at installation, if i update the composer.json (adding a dependency, changing the autoload, ....) when i want to update the root package no changes are made.

The only workaround is to force a new version in both composer.json

root package :

```
require: { "vendor/my-lib": "new-version"}
```

my lib package :

```
"version": "new-version"
```

Then the dependancy is correctly updated.

After some searching, i found that the composer.lock of the root package always include the old version of my-lib dependency, is there some kind of cache ? (doing composer clearcache does not solve this)

alcohol:
I don't really understand the issue.

Can you give a step by step description of how to reproduce your problem?

joelwurtz:
I have a repository with multiples application / package linked together : 

 * projects/common : A common library
 * projects/application : A web application
 * projects/webapp : Another application

Each application and library has a composer.json. 
Composer.json of applications contain has a dependency on the common library by using the repository type "path"

```
{
    "name": "vendor/application",
    "license": "proprietary",
    "type": "project",
    "require": {
        "php": ">=5.5.9",
        "symfony/symfony": "~2.7.5",
         ...
        "vendor/common": "*@dev"
    },
    "repositories": [
        {
            "type": "path",
            "url": "../common"
        }
    ]
}
```

The library have also its own composer.json with dependencies : 

```
{
    "name": "vendor/common",
    "license": "proprietary",
    "authors": [
        {
            "name": "Joel Wurtz",
            "email": "jwurtz@jolicode.com"
        }
    ],
    "autoload": {
        "psr-4": {
            "Vendor\\Common\\": "src/",
        }
    },
    "require": {
        "egeloen/http-adapter": "dev-master"
    }
}
```

Everything work fine on the initial composer install or subsequent composer update. 

However when i change the composer.json of my common library, by adding a namespace to psr4, a new dependency, etc... : 

```
{
    "name": "vendor/common",
    "license": "proprietary",
    "authors": [
        {
            "name": "Joel Wurtz",
[...]

PR #4515, opened 2015-10-18
Issue #4512 Set reference in path repository

joelwurtz:
If there is no git folder, no reference is generated for the package, then subsequent update will never happen as for composer this will be the same version / reference as previous so it will be skipped.

In this case when the composer.json of the package is modified it is reinstalled.

### DIFF
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -121,6 +121,8 @@ class PathRepository extends ArrayRepository
             }
             if (is_dir($path.'/.git') && 0 === $this->process->execute('git log -n1 --pretty=%H', $output, $path)) {
                 $package['dist']['reference'] = trim($output);
+            } else {
+                $package['dist']['reference'] = sha1($json);
             }
 
             $package = $this->loader->load($package);


alcohol:
Why would there not be a `.git` folder?

joelwurtz:
Because, the git folder is on an upper level and not on the composer.json level (monolith repository with multiple applications like described in #4512)

alcohol:
I'm not sure if the path repository works properly for a monolith repository. Hence why there are projects like https://github.com/beberlei/fiddler and others.

This PR seems to address a specific issue though, which could be considered unrelated. It does need tests though.

joelwurtz:
If path repository is no meant to be used with monolith repository, then IMO this line should be changed : https://getcomposer.org/doc/05-repositories.md#path

alcohol:
I didn't say that the `path` repository approach is wrong. I'm simply not sure about it myself.

alcohol:
Can you confirm this fixes your issue (#4512) ?

joelwurtz:
Yes it's fix it, but don't know what test should be here, i would prefer to have a test that reproduce #4512 but don't know if there is a mechanisme for such "functional" test

alcohol:
Also, if possible, I think it would be better to use:
https://github.com/composer/composer/blob/7a9eb02190d334513e99a479510f87eed18cf958/src/Composer/Package/Locker.php#L401

Seldaek:
Thanks, merged and I switched it to use the locker because that makes more sense indeed.

Issue #3771, opened 2015-02-23
Proposal: adding a composer outdated command


jenssegers:
Hey guys,

When I'm using other package managers, I often to stuff like `bower list` or `npm outdated` to find out if packages have a newer release, and can be upgraded. However, composer is kind of lacking this functionality. There is the possibility to run the update with the `--dry-run` option, but that will not tell you if there is a newer version if you have strict versions in `composer.json`.

Is this something that could be on the roadmap?

hannesvdvreken:
@jenssegers What does `npm outdated` do exactly? Will it check if the max version constraint of your dependencies is lower than the newest `stable` or `dev` version of that package? Or does it check recursive dependencies as well?

jenssegers:
I think `npm outdated` shows both the max version based on version constraints (similar to --dry-run) and the lastest version of the package in general.

dantman:
```
Package             Current    Wanted  Latest  Location
lodash                3.3.0     3.3.1   3.3.1  lodash
ms                    0.6.2     0.6.2   0.7.0  debug > ms
chalk                 0.4.0     0.4.0   1.0.0  nomnom > chalk
...
```

Yes, it shows the current version, highest matching version according to constraints, the latest version you can update to, and it does it for the dependencies of your dependencies too.

danez:
I also would like to see something like this. I thought about starting a pull request, but this gets kinda complex, at least with no knowledge in composer internals.

nervo:
:+1:

ojhaujjwal:
+1

watilde:
+1

Addvilz:
+95

davidspiess:
i miss this feature a lot!

hannesvdvreken:
You can actually do `composer update --dry-run` and watch the output. It gives you a list of all packages (sub dependencies as well) which could be updated. All you gotta do is filter out your first-line dependencies.

Addvilz:
@hannesvdvreken --dry-run does not give information about packages, versions of which are locked at some version. For example, if package has version 2.5.0 and composer.json has ~2.4.4, we will not receive 2.5.0 update. That is what this proposal is about.

dnagirl:
yes please!

OzzyCzech:
+1

nevvermind:
This could be easily implemented as a plugin if this was in place: https://github.com/composer/composer/pull/3377

fedeisas:
+1

[...]

PR #4517, opened 2015-10-18
Added new command - outdated

freezy-sk:
`composer outdated` will check for outdated packages and show latest versions as requested in #3771 
Options:
`--prefer-stable` Overrides prefer-stable defined in composer.json
`--no-dev` Ignore require-dev packages

```
Package                                      Current      Wanted      Latest
composer/spdx-licenses                         1.1.1        ^1.0       1.1.2
symfony/console                              v2.6.11        ~2.5      v2.7.5
symfony/finder                               v2.6.11        ~2.2      v2.7.5
symfony/process                              v2.6.11        ~2.1      v2.7.5
symfony/filesystem                           v2.6.11        ~2.5      v2.7.5
seld/phar-utils                                1.0.0        ~1.0       1.0.1
phpunit/phpunit                                4.8.6 ~4.5|^5.0.5       5.0.6
phpunit/phpunit-mock-objects                   2.3.0  2.3.0|~3.0       3.0.3
```

### DIFF
--- a/src/Composer/Command/OutdatedCommand.php
+++ b/src/Composer/Command/OutdatedCommand.php
@@ -77,8 +77,16 @@ EOT
 
         // Process packages
         foreach ($requiredPackages as $packageName => $package) {
-            $installedVersion = $installedPackages[$packageName]->getPrettyVersion();
-            $latestVersion = $latestVersions[$packageName];
+            $installedVersion = '';
+            $latestVersion = '';
+
+            if (isset($installedPackages[$packageName])) {
+                $installedVersion = $installedPackages[$packageName]->getPrettyVersion();
+            }
+
+            if (isset($latestVersions[$packageName])) {
+                $latestVersion = $latestVersions[$packageName];
+            }
 
             if (Comparator::greaterThanOrEqualTo($installedVersion, $latestVersion)) {
                 continue;
--- a/src/Composer/Command/OutdatedCommand.php
+++ b/src/Composer/Command/OutdatedCommand.php
@@ -102,7 +102,7 @@ EOT
         }
 
         // Print package table if there is more than a header
-        if ($outputLines > 1) {
+        if (count($outputLines) > 1) {
             $this->getIO()->write($outputLines);
         }
 
--- a/src/Composer/Command/OutdatedCommand.php
+++ b/src/Composer/Command/OutdatedCommand.php
@@ -1,0 +1,228 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Command;
+
+use Composer\Installer;
+use Composer\Plugin\CommandEvent;
+use Composer\Plugin\PluginEvents;
+use Composer\Repository\PlatformRepository;
+use Composer\Semver\Comparator;
+use Composer\Semver\Semver;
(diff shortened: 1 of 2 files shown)


alcohol:
While this does work as advertised, it does introduce more tech debt, in the sense that we actually want to make all Command classes as dumb as possible. In the current state of this PR, this class contains all logic, meaning it cannot easily be re-used. Also it comes with 0 tests (another reason we want to strip the logic out of all Command classes).

freezy-sk:
@alcohol Could you, please, give me some more informations about that new way commands should be done? Because when I was making this one, at first I've checked CONTRIBUTING.md and there was nothing about this, so I checked some other commands (show, status etc) and they have also whole logic inside.
Some guidance about tests will be also helpful. I didn't found command tests in `tests/Composer/Test/Command/`.

alcohol:
I know. Those are the problems. None of the Command classes are well tested right now and they contain too much logic. Refactoring them is part of a to-do list. So I certainly don't blame you for not knowing this. I'm just pointing it out. I don't really know where to move logic at this point yet.

staabm:
for what its worth, there is a symfony doc chapter on how-to test console commands
http://symfony.com/doc/current/components/console/introduction.html#testing-commands

freezy-sk:
@alcohol In this case there isn't that much logic. Basically there are 3 functions as helpers to get data in format that can be used for comparison in the main `execute` function and 2 functions for output formatting.

freezy-sk:
@staabm thanks!

alcohol:
The problem is that the Command classes contain helpers to begin with. Helpers contain logic. This logic should be moved to reusable components. It's not desirable to have to inject instances of Command classes into other classes simply to get access to their helper methods.

freezy-sk:
@alcohol can I catch you on Gitter? I have few questions regarding design of tables and I don't want to spam it here :)

alcohol:
I'm not a Gitter user. You can find me on IRC, or you'll have to be more specific about this Gitter stuff (I did a quick login but no clue where to go from there).

freezy-sk:
@alcohol Gitter is like IRC for Github :) so you should join room `composer/composer`

enumag:
@freezy-sk Shouldn't the wanted column rather show a latest version matching the current constraints?

vinkla:
@enumag Yeah, that would be preferable +1

freezy-sk:
@enumag good point. I'll change that and add constraint as last column.

curry684:
@alcohol I fundamentally agree with your point about keeping logic out of the 'edge' classes, but right now the codebase isn't really at a point where that's realistic. A lot of the time the code is just juggling associative arrays and JSON files around, instead of real objects. I highly doubt that's going to be fixed before 2.0.x-dev is branched, having looked at the current state of the code over the past few days.

The question is whether the 'need' to reorganize that should block good new functionality right now with the project even still being in alpha formally, or that the reorganization of the code should be its own issue without blocking progress elsewhere. I would argue the second option.

naderman:
The issue actually asked for a tool that would check for the latest version beyond the composer.json constraints. I didn't review the PR, but based on your example, this PR does not achieve what https://github.com/composer/composer/issues/3771 asked for but instead implements a nicer output of what you can already get with `composer update --dry-run`?

curry684:
I did just review the PR and although the output and implementation appear correct I have to agree with @alcohol that we shouldn't want to implement this this way. It's more or less introducing about 200 lines of code duplication from core classes into a command class - 2 things we should avoid at all costs.

The paradox would be that this implementation would break once the core gets near the point where the internal classes are starting to support thinner command implementations, but not yet at the point where it could replace this command.

Seldaek:
Closing as it's been implemented in #5028 - thanks though but the integration in show command seemed to make more sense.

PR #4518, opened 2015-10-19
Add setIO() function to easy embedding composer in other projects

pilot911:
Hello. I just want to embed Composer in our framework, but i dont see function setIO().

Please, add it.


alcohol:
https://github.com/composer/composer/blob/master/src/Composer/Factory.php#L199

pilot911:
Thanks!

PR #4519, opened 2015-10-20
Centralized custom table styles

freezy-sk:
Centralized custom table styles for better consistency across application

### DIFF
--- a/src/Composer/Command/LicensesCommand.php
+++ b/src/Composer/Command/LicensesCommand.php
@@ -12,12 +12,12 @@
 
 namespace Composer\Command;
 
+use Composer\Console\Helper\Table;
 use Composer\Json\JsonFile;
 use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
 use Composer\Package\PackageInterface;
 use Composer\Repository\RepositoryInterface;
-use Symfony\Component\Console\Helper\Table;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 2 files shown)


alcohol:
Didn't I tell you you should use/create a `TableStyle()` instance instead? :-\

freezy-sk:
@alcohol yes, you told me that, but after analysis of that two classes and documentation, I know that this is the way it should be done. `TableStyle` has private properties because they don't want it to be extended and all styling should be done in `Table` like it's done with 3 default styles.

alcohol:
I think you misunderstood. Don't extend `TableStyle`. Simply create an instance of it in the Application and make it available.

freezy-sk:
Yes that's done in that custom style initialization like it's done in base Table class. It's not necessary to instanciate TableStyle on every run of Application

alcohol:
But having a whole extra class is necessary? Not sure I follow your logic :-\

freezy-sk:
My solution covers multiple custom styles if they should be used. It sets default style for basic Table instances so it's consistent through whole application and it's done the way it's done in base Symfony class.

alcohol:
I give up. Go wild.

Seldaek:
Don't really see the point of replacing 3 lines by 30lines.. sorry

freezy-sk:
@Seldaek point is that I prepared another command which uses table and they styles should be consistent. So instead of copy-paste of table style in licenses command there was a good way to create default styling of table.

Seldaek:
@freezy-sk ok then I would suggest adding that change together with the new command.

PR #4530, opened 2015-10-23
InstallationManager: notifyInstalls: append authorization header

kornrunner:
Notify installs should use repository auth info if it's set, otherwise request will be denied (unauthorized).

I'm not sure this is the most elegant way of doing it as I haven't checked much of composer internals. I'd be more than happy to rework it if there are any suggestions.

Thanks!

### DIFF
--- a/src/Composer/Installer/InstallationManager.php
+++ b/src/Composer/Installer/InstallationManager.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Installer;
 
+use Composer\IO\IOInterface;
 use Composer\Package\PackageInterface;
 use Composer\Package\AliasPackage;
 use Composer\Repository\RepositoryInterface;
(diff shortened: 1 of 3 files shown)


src/Composer/Installer/InstallationManager.php
```diff
@@ -230,9 +231,17 @@ public function getInstallPath(PackageInterface $package)
         return $installer->getInstallPath($package);
     }
 
-    public function notifyInstalls()
+    public function notifyInstalls(IOInterface $io)
     {
         foreach ($this->notifiablePackages as $repoUrl => $packages) {
+            $repositoryName = parse_url ($repoUrl, PHP_URL_HOST);
+            if ($io->hasAuthentication ($repositoryName))
+            {
```
alcohol: PSR2.

src/Composer/Installer/InstallationManager.php
```diff
@@ -230,9 +231,17 @@ public function getInstallPath(PackageInterface $package)
         return $installer->getInstallPath($package);
     }
 
-    public function notifyInstalls()
+    public function notifyInstalls(IOInterface $io)
     {
         foreach ($this->notifiablePackages as $repoUrl => $packages) {
+            $repositoryName = parse_url ($repoUrl, PHP_URL_HOST);
```
alcohol: PSR2.

PR #4531, opened 2015-10-24
Update composer.json

paladox:
Replace deprecated psr-0 with psr-4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,10 +49,10 @@
         "ext-openssl": "Enabling the openssl extension allows you to access https URLs for repositories and packages"
     },
     "autoload": {
-        "psr-0": { "Composer": "src/" }
+        "psr-4": { "Composer": "src/" }
     },
     "autoload-dev": {
-        "psr-0": { "Composer\\Test": "tests/" }
+        "psr-4": { "Composer\\Test": "tests/" }
     },
     "bin": ["bin/composer"],
     "extra": {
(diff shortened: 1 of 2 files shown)


hkdobrev:
@paladox You would also need to update the hash in the lock file. You could use `composer update --lock`.

hkdobrev:
@paladox You should also fix the PSR-4 syntax in `composer.json`. I believe you should add `\\` to the end of namespaces when using PSR-4. Like so:

``` json
{
     "autoload": {
         "psr-4": { "Composer": "src/" }
     },
     "autoload-dev": {
         "psr-4": { "Composer\\Test": "tests/" }
     }
}
```

Here is the error in Travis CI, because of that:

```
PHP Fatal error:  Uncaught exception 'InvalidArgumentException' with message 'A non-empty PSR-4 prefix must end with a namespace separator.'
```

Here is the relevant documentation: https://getcomposer.org/doc/04-schema.md#psr-4

alcohol:
Also PSR-4 does not work like that. You would have to point to the directory you are aliasing.

E.g.

```
// source
"Composer\\": "src/Composer"

// tests
"Composer\\Test\\": "tests/Composer/Test"
```

alcohol:
For what it is worth, if @Seldaek is ok with switching to `psr-4`, I would personally prefer to see `tests/Composer/Test` just become `tests`, and changing **autoload-dev** to `"Composer\\": "tests"`. There is no need for the extra `Test` namespace, as all test classes can just "live" in the same namespace as the target class they are testing (there should not be any name conflicts).

PR #4540, opened 2015-10-26
Update composer.json

paladox:
Bump required php version to php 5.3.3 since travis tests against 5.3.3.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "issues": "https://github.com/composer/composer/issues"
     },
     "require": {
-        "php": ">=5.3.2",
+        "php": ">=5.3.3",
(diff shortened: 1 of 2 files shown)


alcohol:
Unfortunately, 5.3.2 is the minimum requirement for --reasons--

As much as I would love to raise it to a version that isn't already declared end of life, this is something we simply cannot change without upsetting a lot of lost souls.

paladox:
Hi well platform already has it as 5.3.3 already just the required php version is being bumped.

alcohol:
Platform pretends that we are installing on 5.3.3 yes, also for --reasons--

That only works for this project though. This configuration is not inherited by other projects that depend on `composer/composer`.

paladox:
Oh ok.

PR #4541, opened 2015-10-26
Update composer.json

paladox:
Update required repos versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,20 +23,20 @@
     },
     "require": {
         "php": ">=5.3.2",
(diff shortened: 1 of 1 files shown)


alcohol:
Also not possible without upsetting a lot of people. Since Composer is actually included in some projects as a dependency, this is guaranteed to cause friction. Again, wishful thinking.

Seldaek:
There is absolutely no point to this.. The constraints are already allowing higher versions to be installed..

PR #4542, opened 2015-10-27
Allow easier manipulation of RootAliasPackage

bd808:
While implementing https://github.com/wikimedia/composer-merge-plugin/commit/f706e5ce863ee7f3cf61f165e120d547c64b1a78 I discovered that there was no way to modify the "conflict", "provide" and "replace" collections of a RootAliasPackage after construction. Since the "require" and "require-dev" collections already have support for post-construction changes I decided to propose this patch in the hope that the lack of setters for other cached collections was just a matter of them not having been needed before.

I also boldly added a commit to the pull request that extends RootPackageInterface and RootAliasPackage to expose methods of RootPackage that [composer-merge-plugin](https://github.com/wikimedia/composer-merge-plugin) uses internally. The Package interfaces generally stick to getters and omit setters, but RootPackageInterface did already specify some setter methods. If this part of the change is unwanted here I won't be too disappointed if I'm asked ot remove it before merging, but I though it would be worth proposing as well.

### DIFF
--- a/src/Composer/Package/AliasPackage.php
+++ b/src/Composer/Package/AliasPackage.php
@@ -32,8 +32,6 @@ class AliasPackage extends BasePackage implements CompletePackageInterface
     protected $conflicts;
     protected $provides;
     protected $replaces;
-    protected $recommends;
-    protected $suggests;
 
     /**
      * All descendants' constructors should call this parent constructor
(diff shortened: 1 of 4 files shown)


alcohol:
Is there a reason you ignored current coding styles? Too many violations to comment on individually, but please fix them :-)

bd808:
> Is there a reason you ignored current coding styles? 
Fixed I think. Sorry about that.

I added tests for RootAliasPackage as well to show that the new (and existing) setters actually effect their associated getters.

alcohol:
Thanks :)

Seldaek:
thanks

PR #4546, opened 2015-10-27
Allow fetching auth credentials from an envvar

### LINKED ISSUE #4285: Passing auth credentials to Composer in non-interactive mode?
We are using Composer to require self-hosted svn packages. Normally, when running a `composer install`, the interactive prompt asks me to supply a username and password for Satis, svn, possibly a github token.

Now I want to run a `composer install` in non-interactive mode. The only way I found to do this, is to pre-configure these credentials in `auth.json` (according to https://getcomposer.org/doc/articles/http-basic-authentication.md). But this requires this file to be stored to disk (with the passwords in plaintext), and I don't want that.

Is there a way to pass these credentials directly to a composer command, without storing them in auth.json? I could imagine something like:
```
composer install --no-interaction --config http-basic {"example.org": {"username": "alice", "password": "foo"}
```

firoxer:
When an environmental variable named "COMPOSER_AUTH" is set
as $USERNAME:$PASSWORD, it is automatically used for authentication
e.g. when fetching packages from Satis.

The envvar credentials are of lower priority than URL credentials.

Fixes #4285

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -207,6 +207,16 @@ class RemoteFilesystem
         $this->retryAuthFailure = true;
         $this->lastHeaders = array();
 
+        // Use COMPOSER_AUTH environment variable if set
+        if (getenv('COMPOSER_AUTH')) {
+            $credentials = [];
+            preg_match('/(.+):(.+)/', getenv('COMPOSER_AUTH'), $credentials);
+
+            if (count($credentials) === 2) {
+                $this->io->setAuthentication($originUrl, $credentials[0], $credentials[1]);
+            }
+        }
+
         // capture username/password from URL if there is one
         if (preg_match('{^https?://(.+):(.+)@([^/]+)}i', $fileUrl, $match)) {
             $this->io->setAuthentication($originUrl, urldecode($match[1]), urldecode($match[2]));
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -209,8 +209,7 @@ class RemoteFilesystem
 
         // Use COMPOSER_AUTH environment variable if set
         if (getenv('COMPOSER_AUTH')) {
-            $credentials = [];
-            preg_match('/(.+):(.+)/', getenv('COMPOSER_AUTH'), $credentials);
+            $credentials = explode(':', getenv('COMPOSER_AUTH'), 2);
 
             if (count($credentials) === 2) {
                 $this->io->setAuthentication($originUrl, $credentials[0], $credentials[1]);
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -207,15 +207,6 @@ class RemoteFilesystem
         $this->retryAuthFailure = true;
         $this->lastHeaders = array();
 
-        // Use COMPOSER_AUTH environment variable if set
-        if (getenv('COMPOSER_AUTH')) {
-            $credentials = explode(':', getenv('COMPOSER_AUTH'), 2);
-
-            if (count($credentials) === 2) {
-                $this->io->setAuthentication($originUrl, $credentials[0], $credentials[1]);
-            }
-        }
-
         // capture username/password from URL if there is one
         if (preg_match('{^https?://(.+):(.+)@([^/]+)}i', $fileUrl, $match)) {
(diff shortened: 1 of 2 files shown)


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -133,6 +133,16 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
         $this->retryAuthFailure = true;
         $this->lastHeaders = array();
 
+        // Use COMPOSER_AUTH environment variable if set
+        if (getenv('COMPOSER_AUTH')) {
+            $credentials = [];
+            preg_match('/(.+):(.+)/', getenv('COMPOSER_AUTH'), $credentials);
```
staabm: Sou could simplify by explode on `:` and a limit of 2

firoxer: Done in 13477bc

cs278:
I think your implementation needs to be revised, for 2/3 reasons:

1. Only works for http/https URLs and not things like git clones.
2. Just one credential can be specified in this way. As a consequence the same credential is sent to each different origin, e.g. my GitHub token will be sent to packagist.

firoxer:
Alright. I'm up for it. Just one question: isn't the implementation for fetching the same kind of credentials from the URL itself (e.g. http://user:secret@domain.tld) as naive as mine? It's in the same scope as my addition and has the same effect.

https://github.com/firoxer/composer/blob/13477bcc588be57ad4474f4f58c62e73ca746f14/src/Composer/Util/RemoteFilesystem.php#L145-L148

cs278:
When the credentials are in the URL they are only used for that origin, i.e. the credentials in `https://secret:password@example.com` are only ever sent to `example.com` in your PR credentials set in `COMPOSER_AUTH` are sent to every origin.

firoxer:
Good point! Is a JSON blob (as suggested in the original PR #4285) OK? Theme seemed to be differing opinions on this, but I can't come up with anything better.

alcohol:
Why can't people just use a configuration file?

Seldaek:
I think either a json blob or the bundle-style `export COMPOSER_EXAMPLE__ORG 'user:pass'` would be acceptable.. The latter is a bit more convenient to use but probably means we have to scan the whole environment vars to find them and that's a bit nasty.

So I would rather prefer COMPOSER_AUTH accepting the same as auth.json does. However it should be loaded in https://github.com/composer/composer/blob/master/src/Composer/IO/BaseIO.php#L63 and not in RemoteFilesystem, that way we'll get also support for git and so on and not just file downloads.

firoxer:
It seems that my original commits stray too far from these goals... I'm sorry, my current understanding of Composer won't get us far, and I can't promise anything :(

Seldaek:
@firoxer I don't think you're so far off, the code is just in the wrong place and it should json_decode the content and do some looping over stuff instead of treating it as a single user/pass. It's not extremely composer specific.

alexw23:
:+1:

firoxer:
Moved the handling to BaseIO.php and made it accept JSON blobs as in auth.json (`{"example.org": {"username": "alice", "password": "foo"}`).

Seldaek:
Thanks @firoxer, please note that I changed the behavior just so it works for other things than http-basic, so you'd now have to set the var to `{"http-basic":{"example.org": {"username": "alice", "password": "foo"}}`.

marcovtwout:
I just tried this new feature, but now Composer prompts me for Subversion credentials:

```
$ composer install --no-dev --prefer-dist --optimize-autoloader
Installing dependencies
  - Installing ........... (1.0.1)
    Checking out /tags/1.0.1/@584
The Subversion server (https://.............) requested credentials:
Username:
```

My env-var COMPOSER_AUTH contains a valid json-blob, I took it directly from a valid auth.json:

```
$ echo $COMPOSER_AUTH
{ "http-basic": { "xxxx": { "username": "xxx", "password": "xxx" }}}
```

LeonB:
Awesome feature! I have on issue:

If `$githubOauth` (https://github.com/composer/composer/blob/master/src/Composer/IO/BaseIO.php#L63) is not an array (when `$HOME/.composer/auth.json` doesn't exist) this throws an error:

``` sh
COMPOSER_AUTH='{"http-basic":{}}' composer create-project
  [ErrorException]                            
  array_merge(): Argument #1 is not an array
```

Seldaek:
@LeonB fixed

@marcovtwout I think I found the problem, can you try again after a self-update?

marcovtwout:
@Seldaek It works now, thanks for the quick fix!

PR #4554, opened 2015-10-28
Add "out of memory" tip

Minishlink:
Most of the "out of memory" errors that users encounter may come from XDebug being loaded/enabled.

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -100,7 +100,10 @@ If composer shows memory errors on some commands:
 
 `PHP Fatal error:  Allowed memory size of XXXXXX bytes exhausted <...>`
 
-Check first that XDebug is not loaded in your `php.ini` by running `composer diagnose`. If XDebug is loaded, you should disable it by commenting the line `zend_extension=path/to/xdebug` in your `php.ini`. Don't forget to enable XDebug again after using Composer, if you need it.
+Check first that XDebug is not loaded in your `php.ini` by runnin
+`composer diagnose`. If XDebug is loaded, you should disable it by
+commenting the line `zend_extension=path/to/xdebug` in your `php.ini`.
+Don't forget to enable XDebug again after using Composer, if you need it.
 
 If composer still raises the error, the PHP `memory_limit` should be increased.
 
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -100,7 +100,9 @@ If composer shows memory errors on some commands:
 
 `PHP Fatal error:  Allowed memory size of XXXXXX bytes exhausted <...>`
 
-The PHP `memory_limit` should be increased.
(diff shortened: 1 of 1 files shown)


alcohol:
Please wrap the lines at the correct limit.

alcohol:
Thanks

staabm:
@alcohol maybe we could be more agressive and tell the user on out-of-memory/max-excecution-time errors that x-debug is installed and might be the cause?

alcohol:
Well, when php runs out of memory, execution pretty much stops. So there isn't much "we" (Composer) can do at that point. As for max execution time, there are various factors to consider here. Xdebug rarely is the sole culprit there.

staabm:
@alcohol we can echo out information using a shutdown-function on out-of-memory, see https://3v4l.org/BiSm9

alcohol:
Interesting. @Seldaek how do you feel about that?

Seldaek:
Well we sure can, it's always good to give more helpful error messages. If someone wants to PR it :)

staabm:
@alcohol @Seldaek on the other hand, IMO we could always trigger a warning in case xdebug is enabled because it considerably slows down execution and lead to high memory usage.

Seldaek:
Even without profiling there's such a big impact?

alcohol:
@staabm I think the warning that `diagnose` throws is sufficient. I for one would not like to be bothered by a warning about xdebug every time I run `composer <something>` (I have xdebug enabled and experience virtually no impact in most of my day to day dealings with `composer`).

staabm:
@Seldaek just enabling it can make things very slow, see e.g. https://nikic.github.io/2012/01/19/Careful-XDebug-can-skew-your-performance-numbers.html

staabm:
@Seldaek @alcohol any opinions on this one? should we add a warning or only "some more information" on out-of-memory.

just quoting the main point of the above comment:
> Depending on what you use it can reduce speed by a factor of 4-5 - just because of the extension beeing loaded.

Seldaek:
The problem with the warning is that it won't just be us, it'll be a lot of people, and some won't be able to fix it themselves because running composer with specific php settings means tweaking things and a lot of people are very confused about CLI matters. We can try but I think it might generate a lot of support requests :/ It should be a short message (so not too noisy if you don't fix it) with a link to docs with more details then maybe it'll be fine.

alcohol:
I personally like the shutdown hook you demonstrated. I would prefer not to see a warning on every usage though. That's my opinion :)

Seldaek:
Shutdown hook only triggers if we have a crash though.. a quick test here without/with xdebug shows pretty drastic differences in speed, even though memory usage isn't affected:

`no xdebug => [208.3MB/17.92s] Memory usage: 208.26MB (peak: 528.58MB), time: 17.92s`
`xdebug => [210.3MB/80.05s] Memory usage: 210.35MB (peak: 530.71MB), time: 80.05s`

That's 4x slowdown, I guess it might be worth nagging people about it, I didn't realize it was that bad (I always run with it disabled).

alcohol:
How bad it is also depends on how you have it configured. But yes, it has a significant impact in most scenarios.

staabm:
@alcohol even with all features disabled its about 3-4x slower, see the linked artices above

alcohol:
I know, and I gave you my opinion already. Now go make a PR and I'll merge it, regardless of what you put in it :p

PR #4556, opened 2015-10-28
Stop fileloads on classMapAuthoritative 

bp1222:
When classMapAuthoritative is enabled autoload_real, will still look at and load the PSR0, PSR4, and include paths.  This is not necessary as in ClassLoader::findFile() if it's not in the classMap and we're authority, it will return false.  Only when classmap is not authority will it call findFileWithExtension, which goes and looks at PSR4/0 and IncludePaths.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -544,33 +544,35 @@ class ComposerAutoloaderInit$suffix
 
 HEADER;
 
-        if ($useIncludePath) {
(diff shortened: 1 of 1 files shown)


Seldaek:
I like the idea, it makes sense to not even configure it if we aren't going to use the config.

I think include paths should not be included here, as this is typically not used in combination with psr-0/4 but rather for legacy support of require_once/include_once statements in code. As you can see it calls set_include_path() and nothing is configured on the ClassLoader.

Seldaek:
Alright I did that fix I mentioned above, thanks!

Issue #4558, opened 2015-10-29
Escaping of output styles in commit messages


xelan:
When composer shows the commit messages for installed or updated dependencies, output style tags (e.g. `<info>`, `<error>`, ...) appearing in the text are not escaped properly.

Example: When doctrine/migrations is updated, the commit https://github.com/doctrine/migrations/commit/93a81ff0dcc858de4df5c17d97f2f24b3bfa3c36 causes the rest of the output to have the comment style.

![style-escaping-example](https://cloud.githubusercontent.com/assets/5080535/10812380/d2f91bca-7e15-11e5-9fb9-5a1650f6a9aa.png)

alcohol:
Please don't only share screenshots. Copy and paste the full output of the command you were running in `-vvv` mode, otherwise we cannot assist you.

xelan:
The text output doesn't make sense here, as it doesn't contain the color which is the problem!
You can reproduce the issue by creating a composer.json with a old version of doctrine/migrations as requirement, then updating it.

PR #4559, opened 2015-10-29
Escape angle brackets for commit message output

xelan:
Q | A
------------ | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Tests pass? | yes
Fixed tickets | #4558 

See https://github.com/symfony/symfony/pull/4832 for the PR where the escaping is described.

### DIFF
--- a/src/Composer/Downloader/VcsDownloader.php
+++ b/src/Composer/Downloader/VcsDownloader.php
@@ -145,6 +145,9 @@ abstract class VcsDownloader implements DownloaderInterface, ChangeReportInterfa
                     return '      ' . $line;
                 }, explode("\n", $logs)));
 
+                // escape angle brackets for proper output in the console
+                $logs = str_replace('<', '\<', $logs);
+
                 $this->io->writeError('    '.$message);
                 $this->io->writeError($logs);
             }


staabm:
@xelan you should add a unit test

xelan:
I could not find any tests for the IO output of the downloaders. For example, the tests of GitDownloader either use no IO, or a mock of IOInterface with no assertions.

src/Composer/Downloader/VcsDownloader.php
```diff
@@ -145,6 +145,9 @@ public function update(PackageInterface $initial, PackageInterface $target, $pat
                     return '      ' . $line;
                 }, explode("\n", $logs)));
 
+                // escape angle brackets for proper output in the console
+                $logs = str_replace('<', '\<', $logs);
```
staabm: I guess we can/could have this very same problem in all places where `$this->writeError` (or similar) is called with a value which is not contructed by composer but by user input.

Therefore I guess we need something in `io` which allows-/disallows-markup or escapes-markup of the message to log/write

xelan: There are a lot of messages with a mix of intentional output styles (like `<warning>`) mixed with external data. For example. try `grep -R '<warning>'` in the composer source directory.

How about the following? IO classes have an `escape()` method which can be used to escape external data. The ConsoleIO would add backslashes before opening angle brackets, while another IO type might just return the input when its `escape()` method is called.

This approach would be easy to understand, flexible enough and easy to unit-test. The disadvantage is that this would add an additional method to `IOInterface`, so custom implementations need to be changed. 

Example:
```php
$data = 'foo <bar>baz</bar>';
$this->io->write('<warning>Foo</warning>' . $this->io->escape($data));
```

src/Composer/Downloader/VcsDownloader.php
```diff
@@ -145,6 +145,9 @@ public function update(PackageInterface $initial, PackageInterface $target, $pat
                     return '      ' . $line;
```
alcohol: You can fix the problem here already.

src/Composer/Downloader/VcsDownloader.php
staabm: having something like `io->escape` is what I meant by by above comment.

Do we need this kind of escaping only for `ConsoleIO` or is this escaping required for all class which Implement `IOInterface`?

xelan: I don't know whether other IO classes need escaping - the concrete classes in `Composer\IO` are currently just `ConsoleIO`, `BufferIO` (subclass of `ConsoleIO`) and `NullIO`. While the first two need escaping, the latter doesn't. 

However, developers of custom IO classes would have to implement at least a noop implementation of `escape()` to avoid checking for the method's existence everywhere.

```php
public function escape($input)
{
    return $input;
}
```

As this pull request has already been merged and closed, I suggest you open a new issue and discuss the general escaping problem there.

staabm: here we go https://github.com/composer/composer/issues/4560

PR #4564, opened 2015-10-29
[Perf gain] Exclude tests folder in no-dev mode

ScullWM:
Hi,

Recently I've just saw on different performance tools how composer file "autoload_classmap.php" can impact memory usage on each request. Regardless the project size, this file is very important.

So I've look closer to this file content and I've been surprise to see the number of files used for the differents projects librarys self-tests onto a production server.

So I made a very small patch excluding thoses files on "no-dev" mode (best choice for production).

On a "out of the box" Symfony 2.7.6, it concern 1288 lines on a total of 4332 (still 3044 line).
So I've choice to exclude them on no-dev mode.

Here's a blackfire stacktrace on a optimized autoload 
https://blackfire.io/profiles/bf86642d-285c-481c-ae78-06447739b795/graph
And here a normal one to compare:
https://blackfire.io/profiles/4892e5be-07d2-480d-9249-95543599f136/graph

I've think that Test class hasn't utility on a server production mode, and reduce composer footprint is always welcome.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -48,6 +48,12 @@ class AutoloadGenerator
      */
     private $classMapAuthoritative = false;
 
+    /**
+     * Tests folder exclusion pattern
+     */
+    const EXCLUDE_TEST_PATTERN = '\\Tests\\';
+
+
     public function __construct(EventDispatcher $eventDispatcher, IOInterface $io = null)
     {
         $this->eventDispatcher = $eventDispatcher;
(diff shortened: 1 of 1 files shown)


xabbuh:
I see your need for this, but I think this is the wrong fix. What if I really wanted to use such a class (maybe the author just used a "weird" namespace)? The proper solution for this imho is to use `autoload-dev` in our packages for test classes (yeah we don't do that currently in Symfony as it's not really possible with the current directory structure).

ScullWM:
@xabbuh I never see this autoload-dev option, it would be great if it was used.

Concerning the use of a "weird" namespace, I've quickly imagine this case.
Seems that there is a "fallback" if a namespace is call and not previously dumped, so there is no blocking situation (just no dump performance gain). 
If the case of "weird" namespace is rare this should still represent gain for mostly project.

alcohol:
This isn't something Composer should be fixing. It already has the tools in place to solve it properly. It's up to projects/maintainers now to adopt and use these :-)

Seldaek:
Just merged #1607 which would allow fixing this more gracefully for package layouts like symfony. See https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps for usage.

PR #4571, opened 2015-10-31
Add test for RemoteFilesystem::getOptions

scotchfield:
First unit test for composer, I hope it's okay! Just trying to get a handle on some other unit testing suites. Thank you!


alcohol:
I appreciate the gesture, but I have some feedback;

- please don't add redundant whitespace to array definitions or function calls

Be consistent, look at how other code is formatted.

- add tests for logic/complexity

What you're testing is virtually identical to getter/setter functionality, not worth testing really.

Seldaek:
I agree with @alcohol and due to that I'll close but please don't take it the hard way, more tests for actual functionality beyond simple getter setters are very much welcome :)

Issue #2804, opened 2014-03-12
With *.tar.xz format not work ;(


samokspv:
Can't extract .tar.xz format
<code>
{
"type": "package",
"package": {
"name": "wkhtmltopdf2/wkhtmltopdf2",
"version": "0.12.0",
"dist": {
"url": "http://downloads.sourceforge.net/project/wkhtmltopdf/0.12.0/wkhtmltox-linux-amd64_0.12.0-03c001d.tar.xz",
"type": "tar"
}
}
}
</code>

<code>
composer update --no-dev
</code>
<code>
Loading composer repositories with package information
Updating dependencies
  Installing wkhtmltopdf2/wkhtmltopdf2 (0.12.0)
    Downloading: 100%
[UnexpectedValueException] 
internal corruption of phar "/var/www/test.dir.dev/vendor/wkhtmltopdf2/wkhtmltopdf2/6f9b4ecfd1a68d66942c056e93ea34f0.xz"  
(__HALT_COMPILER(); not found)
</code>

Seldaek:
There's now support for `"type": "xz"` to extract xz files.

PR #4573, opened 2015-10-31
.tar.xz downloader

Rudloff:
Hello,

As stated in https://github.com/composer/composer/issues/2804, it can be useful to be able to install from XZ tarballs. So here is an XzDownloader class based on GzipDownloader.

Regards,

### DIFF
--- a/src/Composer/Downloader/XzDownloader.php
+++ b/src/Composer/Downloader/XzDownloader.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Downloader;
+
+use Composer\Config;
+use Composer\Cache;
+use Composer\EventDispatcher\EventDispatcher;
+use Composer\Package\PackageInterface;
+use Composer\Util\ProcessExecutor;
+use Composer\IO\IOInterface;
+
+/**
+ * Xz archive downloader.
+ *
+ * @author Pavel Puchkin <i@neoascetic.me>
+ * @author Pierre Rudloff <contact@rudloff.pro>
+ */
+class XzDownloader extends ArchiveDownloader
+{
+    protected $process;
+
+    public function __construct(IOInterface $io, Config $config, EventDispatcher $eventDispatcher = null, Cache $cache = null, ProcessExecutor $process = null)
+    {
+        $this->process = $process ?: new ProcessExecutor($io);
+        parent::__construct($io, $config, $eventDispatcher, $cache);
+    }
+
+    protected function extract($file, $path)
+    {
+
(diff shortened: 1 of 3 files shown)


alcohol:
Would probably benefit from a test (mark as skipped on Windows perhaps?). Other than that, looks good enough. Not very portable though perhaps.

Rudloff:
I added a simple test based on ZipDownloaderTest. Is this the correct way to do it?

src/Composer/Downloader/XzDownloader.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function __construct(IOInterface $io, Config $config, EventDispatcher $eventDispatcher = null, Cache $cache = null, ProcessExecutor $process = null)
+    {
+        $this->process = $process ?: new ProcessExecutor($io);
+        parent::__construct($io, $config, $eventDispatcher, $cache);
+    }
+
+    protected function extract($file, $path)
+    {
+
```
alcohol: whitespace

src/Composer/Downloader/XzDownloader.php
```diff
@@ -0,0 +1,59 @@
...
+ * @author Pierre Rudloff <contact@rudloff.pro>
+ */
+class XzDownloader extends ArchiveDownloader
+{
+    protected $process;
+
+    public function __construct(IOInterface $io, Config $config, EventDispatcher $eventDispatcher = null, Cache $cache = null, ProcessExecutor $process = null)
+    {
+        $this->process = $process ?: new ProcessExecutor($io);
+        parent::__construct($io, $config, $eventDispatcher, $cache);
```
alcohol: should have a newline before the parent call

src/Composer/Downloader/XzDownloader.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+
+        $command = 'tar -xJf ' . ProcessExecutor::escape($file) . ' -C ' . ProcessExecutor::escape($path);
+
+        if (0 === $this->process->execute($command, $ignoredOutput)) {
+            return;
+        }
+
+        $processError = 'Failed to execute ' . $command . "\n\n" . $this->process->getErrorOutput();
+        throw new \RuntimeException($processError);
```
alcohol: should have a newline before the throw, not after

src/Composer/Downloader/XzDownloader.php
```diff
@@ -0,0 +1,59 @@
...
+
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getFileName(PackageInterface $package, $path)
+    {
+        return $path.'/'.pathinfo(parse_url($package->getDistUrl(), PHP_URL_PATH), PATHINFO_BASENAME);
```
alcohol: be consistent in using concatenation with or without whitespace, don't mix things up within the same file/class

Rudloff: Most of this code was copied from GzipDownloader.php which has the same inconsistencies :wink: 
But you're right, I'll edit my code.

alcohol: Yeah, copy paste isn't always the best solution ;-)

Seldaek:
Thanks!

PR #4580, opened 2015-11-04
Have xcopy overwrite existing files

maddhatter:
During a `composer install` on Windows, I had warning that xcopy was timing out. It ended up being because a file already existed in the target directory (non-composer issue, was a permissions issue on my end) and `xcopy` was waiting for a response to:
```
Overwrite [filename] (Yes/No/All)?
```
Adding the `/Y` switch will have `xcopy` overwrite without prompting, just to avoid this particular timeout error.


alcohol:
Thanks

PR #4583, opened 2015-11-05
fixes #4562

### LINKED ISSUE #4562: Autoload creation issues
Since the commit 1160b782c80b3243ab8fb8a1b02f6cb67741fd4b (Deduplicate classmap creation and add support for loading classmaps ...)

there is an issue with autoload generation.

It seems that autoload is trying to be generated now prior to any installation which might fail if you install packages (via tar etc.) and want to have the autoload be executed afterwards (could include autoload section in packages part but would be easier otherwise).

Futerhermore, --no-autoloader still tries to execute autoload at the beginning because I get the same error "could not scan".

alcohol:
Fixes #4562

Instead of failing on a `RuntimeException`, we can just skip the bad definition and proceed with the operation. I don't advocate this kind of configuration as it indicates an underlying problem in my opinion, but we can at least "fail" gracefully.

```
rob@macbookpro /tmp $ cat composer.json
{
    "require": { "alcohol/iso3166": "@stable" },
    "autoload": { "classmap": [ "yolo.php" ] },
    "scripts": {
        "pre-install-cmd": "touch yolo.php"
    }
}
rob@macbookpro /tmp $ /srv/git/github/composer/bin/composer install
Could not scan for classes inside "yolo.php" which does not appear to be a file nor a folder
> touch yolo.php
Installing dependencies (including require-dev)
  - Installing alcohol/iso3166 (3.0.0)
    Loading from cache

Writing lock file
Generating autoload files
```

/cc @seldaek how do you feel about this?

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -402,7 +402,11 @@ EOF;
 
         if (isset($autoloads['classmap'])) {
             foreach ($autoloads['classmap'] as $dir) {
-                $loader->addClassMap($this->generateClassMap($dir));
+                try {
+                    $loader->addClassMap($this->generateClassMap($dir));
+                } catch (\RuntimeException $e) {
+                    $this->io->writeError('<warning>'.$e->getMessage().'</warning>');
+                }
             }
         }
 


Seldaek:
Sounds fine to me, at least the warning is clear.

PR #4588, opened 2015-11-06
Update code sample to a working one

aurelienappere:
This update matches the documentation sample: https://getcomposer.org/doc/05-repositories.md#path
Moreover, it works while the previous one was not working with a `composer install`:
```
Problem 1
  - The requested package xxx/yyy could not be found in any version, there may be a typo in the package name.
[...]
```

### DIFF
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -27,7 +27,7 @@ use Composer\Util\ProcessExecutor;
  *
  * @code
  * "require": {
- *     "<vendor>/<local-package>": "*"
+ *     "<vendor>/<local-package>": "*@dev"
  * },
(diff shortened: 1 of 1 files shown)


aurelienappere:
@alcohol Could you please tell me why this PR was closed? The provided example in the code is not working and is different from the documentation (which has a working example).
Thanks :)

alcohol:
Because the documentation is wrong too. The version requirement has nothing to do with the functionality being described. The reason that `*@dev` works has nothing to do with the fact that it's a path repository and everything to do with the fact that you basically want to run `dev-master`. But that might not always be the case.

alcohol:
I'll accept a PR that fixes the documentation, gladly :-)

PR #4589, opened 2015-11-07
Update the autoloader PHPDoc: PSR-4 instead of PSR-0

dunglas:
Just a minor doc update.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -13,9 +13,7 @@
 namespace Composer\Autoload;
 
 /**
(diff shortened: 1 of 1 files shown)


Seldaek:
Well technically it implements both, and classmap support as well :)

TomasVotruba:
@Seldaek What would prefer to mention then. All 3?

Seldaek:
Well if we want to be complete yes

dunglas:
Changed ;-)

Seldaek:
Thanks

TomasVotruba:
Awesome!

Issue #4578, opened 2015-11-03
Improve script configuration by defining aliases or groups


giosh94mhz:
I think that the script section can be improved a bit, with some minor change, to allow DRY.

I often find myself writing composer.json files which are like:

```json
{
	"scripts" : {
		"post-deploy-cmd" : [
			"Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget",
			"bower install -q"
		],
		"post-install-cmd" : [
			"Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget",
			"bower install -q"
		],
		"post-update-cmd" : [
			"Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
			"Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::prepareDeploymentTarget",
			"bower update -q"
		]
	},
}
```

This is very very chatty and also prone to configuration error, since many command need to be in multiple script.

A simple solution to this is to write an alias like this:

```json
{
	"scripts" : {
[...]

PR #4592, opened 2015-11-09
Allow composer script call within other script group

giosh94mhz:
Since I got positive feedback in Issue #4578 here's a first implementation of scripts "group".

Script prefixed by `@` are now considered alias of other script group. I've also implemented an infinite recursion detection (in a separate commit), to avoid simple configuration errors.

I'm not aware of all composer internals, so this code will need an expert review. Here's my consideration:
* prefixing by `@` can cause BC? I can't think of any real situation...
* all command (including `post-install-cmd` and the like) are callable. Is this correct?
* is the minimal documentation I've added enough?

I've made some phpunit tests, and manually tested the following configurations:

```yml
{
    "scripts": {
        "root": "@group2",
        "group2": [
            "@group1",
            "echo 2",
            "@group3"
        ],
        "group1": [
            "echo 1"
        ],
        "group3": [
            "echo 3"
        ]
    }
}
```
which result in 1 2 3 echoes.

```yml
{
    "scripts": {
        "root": "@group",
        "group" : [
            "echo 1",
            "@root"
        ]
    }
}
```
which triggers a `\RuntimeException`

### DIFF
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -200,3 +200,18 @@ simply running `composer test`:
 
 > **Note:** Composer's bin-dir is pushed on top of the PATH so that binaries
 > of dependencies are easily accessible as CLI commands when writing scripts.
+
+Composer script can also called from other scripts, by prefixing the command name
+by `@`. For example the following syntax is valid:
+
+```json
+{
+    "scripts": {
+        "test": [
+            "@clearCache",
+            "phpunit"
+        ],
+        "clearCache": "rm -rf cache/*"
+    }
+}
+```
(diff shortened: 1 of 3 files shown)


giosh94mhz:
Just noticed that `CommandEvent` class is deprecated and fixed with `Script\Event`. I've not set constructor parameters `$args` and `$flags`: could this possibly be a bug?

giosh94mhz:
I've read carefully the Bash manual and the `@` is always an invalid character at the beginning of the command, so it is in POSIX.

In windows Batch commands on the other hand, the `@` is to suppress/reduce the verbosity of the command, so it can collide with this syntax; however, the composer options `--verbose` and `--quiet` may be enough to workaround this problem.

Seldaek:
Thanks!

PR #4594, opened 2015-11-09
Add "force-dist" option to avoid installation from sources

### LINKED ISSUE #4591: Add a `--force-dist` option to avoid cloning
As cloning from source is a bad practice for production deployment (time spent, dependencies on VCS, ...), i propose to add a "--force-dist" option to "install" / "update" commands who act as "--prefer-dist", but who will not fallback on "source" install, and raise an exception instead.

Today, the only way to detect source download is a weird trick who look for "Cloning from" in the command output :/

mikaelrandy:
close #4591 

This new "--force-dist" option allow to avoid using of downloading from sources.

### DIFF
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -34,8 +34,9 @@ class InstallCommand extends Command
             ->setName('install')
             ->setDescription('Installs the project dependencies from the composer.lock file if present, or falls back on the composer.json.')
             ->setDefinition(array(
-                new InputOption('prefer-source', null, InputOption::VALUE_NONE, 'Forces installation from package sources when possible, including VCS information.'),
-                new InputOption('prefer-dist', null, InputOption::VALUE_NONE, 'Forces installation from package dist even for dev versions.'),
+                new InputOption('prefer-source', null, InputOption::VALUE_NONE, 'Prefer installation from package sources when possible, including VCS information.'),
+                new InputOption('prefer-dist', null, InputOption::VALUE_NONE, 'Prefer installation from package dist even for dev versions.'),
+                new InputOption('force-dist', null, InputOption::VALUE_NONE, 'Forces installation from package dist, and avoid to use sources installation'),
                 new InputOption('dry-run', null, InputOption::VALUE_NONE, 'Outputs the operations but will not execute anything (implicitly enables --verbose).'),
                 new InputOption('dev', null, InputOption::VALUE_NONE, 'Enables installation of require-dev packages (enabled by default, only present for BC).'),
                 new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables installation of require-dev packages.'),
@@ -91,6 +92,7 @@ EOT
 
         $preferSource = false;
         $preferDist = false;
+        $forceDist = false;
 
         $config = $composer->getConfig();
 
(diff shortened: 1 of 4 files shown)


src/Composer/Command/InstallCommand.php
```diff
@@ -34,8 +34,9 @@ protected function configure()
             ->setName('install')
             ->setDescription('Installs the project dependencies from the composer.lock file if present, or falls back on the composer.json.')
             ->setDefinition(array(
-                new InputOption('prefer-source', null, InputOption::VALUE_NONE, 'Forces installation from package sources when possible, including VCS information.'),
-                new InputOption('prefer-dist', null, InputOption::VALUE_NONE, 'Forces installation from package dist even for dev versions.'),
+                new InputOption('prefer-source', null, InputOption::VALUE_NONE, 'Prefer installation from package sources when possible, including VCS information.'),
+                new InputOption('prefer-dist', null, InputOption::VALUE_NONE, 'Prefer installation from package dist even for dev versions.'),
+                new InputOption('force-dist', null, InputOption::VALUE_NONE, 'Forces installation from package dist, and avoid to '),
```
pborreli: Sentence doesn't look finished `and avoid to ` ...

src/Composer/Downloader/DownloadManager.php
```diff
@@ -171,17 +185,22 @@ public function download(PackageInterface $package, $targetDir, $preferSource =
...
         }
 
-        if (empty($sources)) {
+        if (!$sourceType && !$distType) {
             throw new \InvalidArgumentException('Package '.$package.' must have a source or dist specified');
         }
 
+        if (empty($sources)) {
+            $this->io->writeError('    <warning>No sources available for package '.$package.' for your request</warning>');
+            return;
```
stof: returning is wrong. The installation must fail if you force installing from dist and the package does not support it. Otherwise, you might mark the installation as successful although some packages were not downloaded

src/Composer/Downloader/DownloadManager.php
```diff
@@ -240,15 +259,19 @@ public function update(PackageInterface $initial, PackageInterface $target, $tar
...
+        if ($this->forceDist) {
+            $targetType = $target->getDistType();
+        }
+
         // upgrading from a dist stable package to a dev package, force source reinstall
         if ($target->isDev() && 'dist' === $installationSource) {
             $downloader->remove($initial, $targetDir);
             $this->download($target, $targetDir);
-
-            return;
```
stof: please don't remove the early returns

src/Composer/Command/InstallCommand.php
mikaelrandy: Good catch ! I changed the other sentences i forgot to finish mine :/

src/Composer/Downloader/DownloadManager.php
mikaelrandy: As you can see, i transform the "one if / one if else" to "if elseif else" structure, so there is no more need for early returns as there is no more code executed once entered in a structure control

src/Composer/Command/InstallCommand.php
```diff
@@ -110,6 +112,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $preferSource = $input->getOption('prefer-source');
             $preferDist = $input->getOption('prefer-dist');
         }
+        if ($input->getOption('force-dist')) {
```
jubianchi: the `if` here can be removed

mikaelrandy: It's just to be consistent with affectation of `$prefer{Source,Dist}` vars

jubianchi:
seems good to me.

Seldaek:
Closing, please discuss in #4591

PR #4599, opened 2015-11-10
Fix handling of http 400 errors

davidfuhr:
Before 400 errors were silently dropped. This caused composer to store
the returned body (usually some html) in cache. This resulted in later
errors when composer tried to extract this error response as zip.

With this exceptions thrown it handles the error gracefully and falls
back to the next URL provided by the package configuration.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -334,6 +334,12 @@ class RemoteFilesystem
     {
         switch ($notificationCode) {
             case STREAM_NOTIFY_FAILURE:
+                if (400 === $messageCode) {
+                    // This might happen if your host is secured by ssl client certificate authentication
+                    // but you do not send an appropriate certificate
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -334,6 +334,12 @@ protected function callbackGet($notificationCode, $severity, $message, $messageC
     {
         switch ($notificationCode) {
             case STREAM_NOTIFY_FAILURE:
+                if (400 === $messageCode) {
+                    // This might happen if your host is secured by ssl client certificate authentication
+                    // but you do not send an appropriate certificate
+                    throw new TransportException("The '" . $this->fileUrl . "' URL could not be accessed: " . $message, $messageCode);
+                }
+
```
stof: you should add a comment saying that the fallthrough to the next case is intended, to avoid mistake in the future

Seldaek: Fixed in 58a6d4b7d305a0ff98e8417a51b59b7d2cfa638c

PR #4607, opened 2015-11-12
Up-level relative paths in exclude-from-classmap

### LINKED ISSUE #4595: Cannot exclude for classmaps generated from a parent directory
composer.json path: ``/project-dir/sub/composer.json``

I am adding files to classmap from ``/dir``. This all works fine. But when I try to exclude some paths from the classmap it doesn't work at all

I have a directory ``/project-dir/a/b/c/unwanted-dir`` which I want to exclude from classmap

I added the following to `composer.json`:

```
"exclude-from-classmap": [
    "**/unwanted-dir"
]
```

After some debugging I discovered composer directory gets added into the regex that matches paths to exclude.

```
$blacklist:
string(35) "{(/dir/sub/.+?/skeleton)}"
```

I think this is a bug, but until it gets fixed, is there any other way of getting excludes working than moving ``composer.json`` to ``/project-dir``?

ockcyp:
Fixes #4595

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -697,8 +697,25 @@ FOOTER;
                         // add support for wildcards * and **
                         $path = str_replace('\\*\\*', '.+?', $path);
                         $path = str_replace('\\*', '[^/]+?', $path);
+                        // add support for up-level relative paths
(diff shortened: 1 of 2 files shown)


ockcyp:
does anyone know why it's failing on Windows?

Seldaek:
Thanks

Issue #4412, opened 2015-09-12
Document new path repository type


Seldaek:
C.f. #4365

sroze:
I'll do it today.

sroze:
Follows up in #4416

potherca:
There is no mention in the docs about the fact that Composer will create a symlink (when possible). Should I open a PR to add this information?

sroze:
Any PR is welcome :)

potherca:
I'll do that right now... Saves me from having to write a blogpost :-)

potherca:
Details added in PR #4612

PR #4612, opened 2015-11-15
Adds more details about Path Repository to the Documentation - #4412

potherca:
- Mentions symlinking
- Mentions absolute paths also being supported

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -639,9 +639,12 @@ file, you can use the following configuration:
 }
 ```
 
-The local package will be symlinked if possible, in which case the output in the console will read `Symlinked from ../../packages/my-package`. If symlinking is _not_ possible the package will be copied. In that case, the console will output `Mirrored from ../../packages/my-package`.
+The local package will be symlinked if possible, in which case the output in 
+the console will read `Symlinked from ../../packages/my-package`. If symlinking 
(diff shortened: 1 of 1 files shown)


alcohol:
Formatting please

potherca:
Explanation please.

alcohol:
If you look at the rest of the file you might notice that we try to stay within a certain line length.

doc/05-repositories.md
```diff
@@ -639,6 +639,10 @@ file, you can use the following configuration:
 }
 ```
 
+The local package will be symlinked if possible, in which case the output in the console will read `Symlinked from ../../packages/my-package`. If symlinking is _not_ possible the package will be copied. In that case, the console will output `Mirrored from ../../packages/my-package`.
+
+Instead of using a relative path, and absolute paths can also be used.
```
alcohol: s/and/an/

potherca:
Thanks for the review and feedback, the text has been amended.

alcohol:
I'm not sure if putting the exact console output text in there is wise, as it might change over time. But thank you for your contribution :)

potherca:
I was debating that with myself also. As part of the confusion surrounding this feature was in regards to when/whether symlinks or copies were used, I deemed it relevant enough to leave it in.

PR #4621, opened 2015-11-21
Fixes #4610

### LINKED ISSUE #4610: Add "abandoned" to composer.json schema.
I do like the fact that data structures seem to be the same in both composer.json and Packagist.

The feature to mark packages abandoned currently is a meta feature implemented in the Packagist UI. As soon as a package is marked, all published versions and branched inherit this flag.

To allow doing the same for my local Satis instance and private repos, I discovered that Satis will happily consume the "abandoned" flag from the composer.json file and integrate it into it's own package metadata. I know it's limited to the branch this has been committed and that you usually cannot change tags, but I like the idea that the source repository is able to communicate the package status directly instead of indirectly. My own afterburner script reads the created Satis metadata and propagates the "abandoned" flag to all tags.

Now the issue is: The schema for composer.json does not allow adding `"abandoned":true` or `"abandoned":"other/package"`, and I think it should be added. 

It would open a way for source repos to communicate the status of a package without the need for another channel. However, it's unclear on the consuming side how to deal with this info beyond "pass it through unaltered", and I'd think this would be out of scope for this issue.

thiagooak:


Seldaek:
Thanks, note that string is also a valid type for abandoned, but I added it :)

PR #4633, opened 2015-11-23
Adds support for gitlab installed in subfolder

jaapio:
Since gitlab can be hosted on any url also subdirectories should be
supported. (e.g https://mycompany.com/gitlab).
This supports only http and https protocols since the gitlab api url
is derived from the package repository url. And the ssh protocol doesn't
support folders this way.

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -62,6 +62,9 @@ class GitLabDriver extends VcsDriver
      */
     protected $gitDriver;
 
+    const URL_REGEX = '#^((https?)://(.*)/|git@([^:]+):)([^/]+)/(.+?)(?:\.git|/)?$#';
+
+
     /**
      * Extracts information from the repository url.
      * SSH urls uses https by default.
@@ -70,7 +73,7 @@ class GitLabDriver extends VcsDriver
      */
     public function initialize()
     {
-        if (!preg_match('#^((https?)://([0-9a-zA-Z\./]+)/|git@([^:]+):)([^/]+)/(.+?)(?:\.git|/)?$#', $this->url, $match)) {
+        if (!preg_match(static::URL_REGEX, $this->url, $match)) {
             throw new \InvalidArgumentException('The URL provided is invalid. It must be the HTTP URL of a GitLab project.');
         }
 
@@ -343,7 +346,7 @@ class GitLabDriver extends VcsDriver
      */
     public static function supports(IOInterface $io, Config $config, $url, $deep = false)
     {
-        if (!preg_match('#^((https?)://([^/]+)/|git@([^:]+):)([^/]+)/(.+?)(?:\.git|/)?$#', $url, $match)) {
+        if (!preg_match(static::URL_REGEX, $url, $match)) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


alcohol:
Why not simply add the URL to the dataset in `dataForTestSupports`?

stof:
I see an issue here: if the project name contains a ``/`` (which is allowed by the regex), the matching will be broken (the regex can also match it as a subfolder before the owner name)

src/Composer/Repository/Vcs/GitLabDriver.php
```diff
@@ -70,7 +70,7 @@ class GitLabDriver extends VcsDriver
      */
     public function initialize()
     {
-        if (!preg_match('#^((https?)://([^/]+)/|git@([^:]+):)([^/]+)/(.+?)(?:\.git|/)?$#', $this->url, $match)) {
+        if (!preg_match('#^((https?)://([0-9a-zA-Z\./]+)/|git@([^:]+):)([^/]+)/(.+?)(?:\.git|/)?$#', $this->url, $match)) {
```
stof: this is too restrictive. It forbids valid domain names

jaapio:
@stof The project name cannot contain a `/` this is not allowed by gitlab. I will have a look again at the domain part again.

jaapio:
I updated this pr. The regex is extracted to a const in the class. It was defined in 2 places.
Regex is more generic now so it should match all domains.

Thanks to @alcohol for pointing me at the  `dataForTestSupports` dataset. I missed one of the situations in the class.

Seldaek:
Seems fine to me, anyone else any comments?

tests/Composer/Test/Repository/Vcs/GitLabDriverTest.php
```diff
@@ -215,6 +216,17 @@ public function dataForTestSupports()
             array('git@gitlab.com:foo/bar.git', extension_loaded('openssl')),
             array('git@example.com:foo/bar.git', false),
             array('http://example.com/foo/bar', false),
+            array('https://mycompany.com/gitlab/mygroup/myproject', true),
         );
     }
+
+    public function testGitlabSubDirectory()
```
alcohol: This test is already covered by `testInitialize` honestly. If you want you can add the url/apiUrl combo to `getInitializeUrls`.

alcohol:
Other than the last comment I just provided, LGTM.

tests/Composer/Test/Repository/Vcs/GitLabDriverTest.php
jaapio: testInitialize is testing much more than only the extracted api url. I tried to reuse that test before but I had to modify the test so much that I preferred to leave it as is. Since I think that modifying an existing test is a bad practice. The existing test shows us that I didn't break any of the existing functionality. The new tests shows us that the new feature works as expected.

But if you want me to modify the existing test I will do that.

alcohol: Yes I can see how it might be a bit more work. I'm fine with the changes either way in that case.

jaapio: tnx!

Seldaek:
Merging then! Thanks.

Seldaek:
Actually.. this breaks the 5.3.3 build, not sure why but somehow the regex isn't matching it seems. I don't really have time to dig further right now.

jaapio:
It could be that openssl extension is not loaded on php 5.3.3? Since it is only the supports method?

alcohol:
That was my guess too. Seems fixed in #4656

PR #4637, opened 2015-11-25
Add alias trick on xdebug documentation part

Soullivaneuh:
Related @Seldaek comment: https://github.com/composer/composer/commit/7a06615bf06c5a391ab2d360907012529508e695#commitcomment-14599053

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -146,6 +146,27 @@ locating the related `zend_extension` directive and prepending it with `;` (semi
 ;zend_extension = "/path/to/my/xdebug.so"
 ```
 
+If you disable this extension and still want it to be added on `php` cli command, you can deal with aliases on *nix systems:
+
+```sh
+# Load xdebug Zend extension with php command
+alias php='php -dzend_extension=xdebug.so'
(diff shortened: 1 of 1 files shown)


doc/articles/troubleshooting.md
```diff
@@ -146,6 +146,27 @@ locating the related `zend_extension` directive and prepending it with `;` (semi
 ;zend_extension = "/path/to/my/xdebug.so"
 ```
 
+If you disable this extension and still want it to be added on `php` cli command, you can deal with aliases:
```
staabm: I would add that this "workaround" only works un *nix systems

Seldaek: Works fine for me on windows using bash but sure :p

Seldaek:
Thanks

doc/articles/troubleshooting.md
staabm: you need to provide your custom windows bash with composer then ;)

Seldaek: It's an optional but recommended download at 
http://www.mingw.org/wiki/msys :p

PR #4638, opened 2015-11-25
Allows repositories to implement ConfigurableRepositoryInterface

realshadow:
This PR attempts to solve an issue described here - https://github.com/composer/satis/pull/266#discussion_r45597333 by providing a new interface for all repository classes that use `$repoConfig` to implement. Which affects:

- `VcsRepository`
- `PearRepository`
- `ComposerRepository`
- `ArtifactRepository`
- `PathRepository`

PR #4646, opened 2015-11-27
Add missing getRepoConfig() method

mpdude:
Required at https://github.com/composer/satis/blob/85ed79ce87debb4ce6bdbe3d48821945657ccc50/src/Composer/Satis/PackageSelection/PackageSelection.php#L160, at least.

### DIFF
--- a/src/Composer/Repository/PackageRepository.php
+++ b/src/Composer/Repository/PackageRepository.php
@@ -39,6 +39,11 @@ class PackageRepository extends ArrayRepository
         }
     }
 
+    public function getRepoConfig()
+    {
+        return $this->config;
+    }
+    
     /**
      * Initializes repository (reads file, or remote address).
      */


mpdude:
Oh, sorry – added in #4638 already.

alcohol:
This isn't going to work as it's never stored on the instance. Also `PackageRepository` does not implement the `ConfigurableRepositoryInterface` which as introduced by #4638

mpdude:
> This isn't going to work as it's never stored on the instance.

Not sure what you mean.

Alternatively, we could add a check near https://github.com/composer/satis/blob/85ed79ce87debb4ce6bdbe3d48821945657ccc50/src/Composer/Satis/PackageSelection/PackageSelection.php#L160 to process only instances of `ConfigurableRepositoryInterface`.

alcohol:
@mpdude that was the idea already, please read https://github.com/composer/satis/pull/266#discussion_r46024402 and other comments in that discussion.

mpdude:
Sorry, missed that

Seldaek:
So this can be closed right? I am closing because I don't think it makes sense but if I misunderstood let me know.

mpdude:
Yes, sorry, could have closed that myself.

PR #4651, opened 2015-11-27
Add dry run update and install events

### LINKED ISSUE #4650: Update event with --dry-run option
Actually, if we launch `update` command with `--dry-run` option, absolutely [no script events is launched](https://github.com/composer/composer/blob/master/src/Composer/Installer.php#L193-L197).

I think those events should be dispatched in any case, but with the `--dry-run` option value inside or, at least, add another events like `PRE_DRY_UPDATE_CMD` / `POST_DRY_UPDATE_CMD`.

This will permit plugins to get information about dry run install and update commands.

Soullivaneuh:
Closes #4650.

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -194,6 +194,9 @@ class Installer
             // dispatch pre event
             $eventName = $this->update ? ScriptEvents::PRE_UPDATE_CMD : ScriptEvents::PRE_INSTALL_CMD;
             $this->eventDispatcher->dispatchScript($eventName, $this->devMode);
+        } elseif ($this->dryRun) {
+            $eventName = $this->update ? ScriptEvents::PRE_UPDATE_DRY_CMD : ScriptEvents::PRE_INSTALL_DRY_CMD;
+            $this->eventDispatcher->dispatchScript($eventName, $this->devMode);
         }
 
         $this->downloadManager->setPreferSource($this->preferSource);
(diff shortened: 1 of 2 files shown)


alcohol:
I personally don't think these should be different/new events. It makes more sense to me to fire the same events as we usually would; but the options that the command received should be passed along with it somehow (e.g. that it is a dry-run).

Seldaek:
Yeah I agree this fix doesn't make sense to me. But obviously if we start firing the events and plugins do crazy shit (tm) it's not gonna be good either. I am not sure what the best way is.. but down the line a dry run shouldn't really need to trigger the scripts IMO. What's the use case really?

Soullivaneuh:
> Yeah I agree this fix doesn't make sense to me.

I was thinking about that at start.

> But obviously if we start firing the events and plugins do crazy shit (tm) it's not gonna be good either.

Yeah. For example on Symfony projects, some scripts will be called on crash.

> but down the line a dry run shouldn't really need to trigger the scripts IMO. What's the use case really?

Maybe for project scripts, not for plugin. For the use case, I wrote a plugin that will show you outdated package after an update: https://github.com/Soullivaneuh/composer-versions-check

This will be useful if users can simulate an update just to see the plugin result.

Seldaek:
For that use case I think it'd be better to allow plugins to add new 
commands (there's a PR for it already I believe). Having to do a dry run 
or even an update at all to see the output doesn't make so much sense I 
think.

Soullivaneuh:
> For that use case I think it'd be better to allow plugins to add new
commands (there's a PR for it already I believe).

You got my curiosity. Can you tell me which PR?

> Having to do a dry run
or even an update at all to see the output doesn't make so much sense I
think.

If I can add a command through my plugin, in this case I agree.

But I think that dry-run events is still a good idea. Can be used by others.

Soullivaneuh:
I found indeed too PR like that:

* https://github.com/composer/composer/pull/3377
* https://github.com/composer/composer/pull/4124

Soullivaneuh:
I had to delete my fork to start a new one.

Tell me if I have to change something In this case, I'll open another PR.

Seldaek:
Yeah I think those PRs are more the way to go, I don't want to add a copy of every event for dry run mode.

PR #4656, opened 2015-11-30
fix a 5.3.3 unit test

alcohol:
it looks like php 5.3.3 might not have openssl extension?


Seldaek:
Yeah unfortunately it doesn't :/

PR #4672, opened 2015-12-04
[ClassLoader] performance improvement: use str_replace instead of strtr

Tobion:
str_replace is faster than strtr for this use-case. See symfony/symfony#15614

And we can even skip the str_replace on Windows completely.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -344,7 +344,7 @@ class ClassLoader
     private function findFileWithExtension($class, $ext)
     {
         // PSR-4 lookup
-        $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR) . $ext;
+        $logicalPathPsr4 = str_replace('\\', DIRECTORY_SEPARATOR, $class) . $ext;
 
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {
@@ -370,10 +370,10 @@ class ClassLoader
         if (false !== $pos = strrpos($class, '\\')) {
             // namespaced class name
             $logicalPathPsr0 = substr($logicalPathPsr4, 0, $pos + 1)
-                . strtr(substr($logicalPathPsr4, $pos + 1), '_', DIRECTORY_SEPARATOR);
+                . str_replace('_', DIRECTORY_SEPARATOR, substr($logicalPathPsr4, $pos + 1));
         } else {
             // PEAR-like class name
-            $logicalPathPsr0 = strtr($class, '_', DIRECTORY_SEPARATOR) . $ext;
+            $logicalPathPsr0 = str_replace('_', DIRECTORY_SEPARATOR, $class) . $ext;
         }
 
         if (isset($this->prefixesPsr0[$first])) {
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -344,7 +344,7 @@ class ClassLoader
     private function findFileWithExtension($class, $ext)
     {
         // PSR-4 lookup
-        $logicalPathPsr4 = str_replace('\\', DIRECTORY_SEPARATOR, $class) . $ext;
+        $logicalPathPsr4 = ('\\' !== DIRECTORY_SEPARATOR ? str_replace('\\', DIRECTORY_SEPARATOR, $class) : $class) . $ext;
 
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {


stof:
please provide a benchmark proving that the new code is indeed faster

Tobion:
See linked ticket.

alcohol:
I didn't notice the link in the referenced PR the first time.

It seems a minor increase could be won here. But I will have to defer to @seldaek to confirm if the change for Windows is valid or not.

stof:
@alcohol given that ``DIRECTORY_SEPARATOR`` is ``\`` on windows, the change avoids running a ``str_replace('\\', '\\', $string)``. This is indeed a valid change

alcohol:
Alright. Thanks @Tobion

vlakoff:
If I recall correctly, `strtr` is faster (for one-to-one character replacements, and only for this case).

Symfony made the change for code consistency, not performance.

vlakoff:
Concerning:

    (DIRECTORY_SEPARATOR !== ''\\'' ? strtr($class, '\\', DIRECTORY_SEPARATOR) : $class)

versus:

    strtr($class, '\\', DIRECTORY_SEPARATOR)

Is saving a function call worth the added complexity?

vlakoff:
To conclude on my side, all of this of this is of minor importance.
Just let @Seldaek pick its preferred style ;)

alcohol:
@vlakoff the "benchmark" provided by @Tobion also dealt with single character replacement, and it actually revealed `str_replace` to be faster.

Ideally, someone would provided a benchmark for both methods using a more real-world alike dataset rather than just single character replacement (e.g. a list of all possible classes in the `symfony/symfony` package).

Seldaek:
A few comments on benchmarking:

- I don't think 3v4l.org is a great platform for benchmarking.. I don't think those results mean much, and they're pretty hard to read anyway.
- benchmark results in the .00x second range are pretty  meaningless IMO as any kind of short CPU spike on the machine can skew them greatly.
- benchmarking a function in a context is one thing, but arguments matter. And adding a simple check for `\ === DIRECTORY_SEPARATOR` also changes the game entirely.
- environments matter, OSs matter, and so on.

So all in all I don't think this should have been merged blindly. But anyway let's have a look..

I created a gist with a benchmark script, please go have a look and tell me if I messed up: https://gist.github.com/Seldaek/07d8d899f72701bf549e

You can run it with this:

```
wget https://gist.githubusercontent.com/Seldaek/07d8d899f72701bf549e/raw/ad8da7630182f30d82cfd14f2c7d219f4cc9fb87/bench.php
php bench.php
```

Now for the results I have seen:

On windows obviously the new code is faster because it doesn't do any function call anymore, but aside from that it strtr seems faster for this use case.

```
PHP 7.0.0RC8 (windows)
new/str_replace/full :
    time: 1.1568069458008
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
old/strtr :
    time: 5.2230041027069
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
new/str_replace :
    time: 7.2021079063416
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
```

On linux I have 6 result sets, from 2 machines, and one has php5 + php7 results too just to make sure as I think we really rather want to check this only on php7 because that's the future and the place to be if you care about perf. In all of these, it seems like strtr is marginally faster than str_replace in php5 and quite a lot faster in php7. Not only that, the windows short-circuit check actually makes it even slower.

```
PHP 5.6.14-1+deb.sury.org~trusty+1
old/strtr :
    time: 7.8680801391602
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 8.0060811042786
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 9.0768179893494
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4

PHP 5.6.14-1+deb.sury.org~trusty+1
old/strtr :
    time: 7.8322699069977
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 7.9702620506287
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 8.9150371551514
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4

PHP 5.6.14-1+deb.sury.org~trusty+1
old/strtr :
    time: 8.2552790641785
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 8.2914118766785
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 8.556049823761
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4



PHP 5.5.9-1ubuntu4.13
old/strtr :
    time: 55.854964017868
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 56.090523004532
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 67.721940994263
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4

PHP 7.0.0-4+deb.sury.org~trusty+1
old/strtr :
    time: 12.563672065735
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 22.043679952621
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 22.849457979202
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4

PHP 7.0.0-4+deb.sury.org~trusty+1
old/strtr :
    time: 12.995949029922
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 22.223428964615
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 23.039249181747
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
```

According to these numbers, I would say this should be reverted entirely, but I would be happy to get results from other people as well.

Capenus:
```
PHP 5.5.28 (windows)
new/str_replace/full :
    time: 3.6592099666595
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
old/strtr :
    time: 10.77761721611
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
new/str_replace :
    time: 18.142036914825
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
PHP 5.5.28 (windows)
new/str_replace/full :
    time: 3.6262068748474
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
old/strtr :
    time: 10.673609972
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947
new/str_replace :
    time: 17.916025161743
    iterations: 5000
    retval: ad2e65f767c7e88b8f417d2706430947

PHP 5.4.45-1~he.1
old/strtr :
    time: 14.740041017532
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 20.53439617157
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 21.646430969238
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
PHP 5.4.45-1~he.1
old/strtr :
    time: 14.519618034363
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 20.330652952194
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 21.65713596344
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4


PHP 5.5.30-1~he.1
old/strtr :
    time: 15.6178150177
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 21.264933109283
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 22.114634990692
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
PHP 5.5.30-1~he.1
old/strtr :
    time: 15.333259105682
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 21.172958135605
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 22.155410051346
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4

PHP 5.6.15-1~he.0
old/strtr :
    time: 14.450623035431
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 20.315406084061
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 21.494570016861
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
PHP 5.6.15-1~he.0
old/strtr :
    time: 14.439452171326
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 20.340620040894
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 21.91211605072
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
```

alcohol:
**OSX 10.11.1**

*PHP 5.6.16 with Xdebug*
```
old/strtr :
    time: 18.955353975296
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 20.901481151581
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 21.974380970001
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
```
*PHP 5.6.16 without Xdebug*
```
old/strtr :
    time: 7.0231590270996
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 8.8984019756317
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 9.7666220664978
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
```

**Archlinux 4.2.5-1**
*PHP 5.6.15*
```
old/strtr :
    time: 7.7712788581848
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace :
    time: 8.0035860538483
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
new/str_replace/full :
    time: 8.3755970001221
    iterations: 5000
    retval: 1b8788cb29e73586277157ffdc989ff4
```

alcohol:
> According to these numbers, I would say this should be reverted entirely, but I would be happy to get results from other people as well.

I concur.

vlakoff:
Wow, that escalated quickly.

Totally approve:
> benchmark results in the .00x second range are pretty meaningless IMO as any kind of short CPU spike on the machine can skew them greatly.

You should have added also 'strtr/full' in your benchmark.

Previous code using `strtr()` was just fine.
But if performance is really, really crucial, you might opt for `DIRECTORY_SEPARATOR !== '\\' + strtr()`: adds a comparison on linux, saves a function call on windows.

alcohol:
Considering over 90% of development/production environments are probably Unixy (and that is being generous towards Windows), I would sooner opt for having one less comparison on Linux than one less function call on Windows.

stof:
Are you running this with or without OPCache ? This might make a difference as OPCache may be able to resolve the the constant check at compile time (maybe needed to use ``\DIRECTORY_SEPARATOR`` so that the compiler knows it does not need to check first in the current namespace)

vlakoff:
As alcohol said, almost all production servers are on Linux, so the best fit is simply the previous code.

Tobion:
It's the task of the compiler to optimize code. And in this case it's a constant comparision that should be removed by the compiler. We just write code that makes sense, and it does not make sense to replace a string with the same string.

stof:
@Tobion due to the fact that constants are checked first in the current namespace with a fallback to the global namespace, the compiler might not be able to perform the optimization. This is why I said to try using ``\DIRECTORY_SEPARATOR`` instead (and with OPCache enabled, as such optimizations are applied by OPCache)

Seldaek:
@stof given that the benchmark file I wrote has no namespace, I guess if opcache optimized this we wouldn't see a performance hit?

@Tobion I'd agree with you in general, but it's also very common to micro-optimize particularly tight loops when the compiler can't/won't.

Anyway.. I'll revert then, as it seems everyone's numbers say the same.

PR #4676, opened 2015-12-07
Improve PHPDoc for package requires

Soullivaneuh:
Replaces #4508.

### DIFF
--- a/src/Composer/Package/AliasPackage.php
+++ b/src/Composer/Package/AliasPackage.php
@@ -27,10 +27,15 @@ class AliasPackage extends BasePackage implements CompletePackageInterface
     protected $rootPackageAlias = false;
     protected $stability;
 
+    /** @var Link[] */
     protected $requires;
+    /** @var Link[] */
     protected $devRequires;
+    /** @var Link[] */
     protected $conflicts;
+    /** @var Link[] */
     protected $provides;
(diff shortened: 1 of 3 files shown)


src/Composer/Package/AliasPackage.php
```diff
@@ -27,7 +27,9 @@ class AliasPackage extends BasePackage implements CompletePackageInterface
     protected $rootPackageAlias = false;
     protected $stability;
 
+    /** @var Link[] */
     protected $requires;
+    /** @var Link[] */
```
stof: should be added for conflicts, provides and replaces too

src/Composer/Package/Package.php
```diff
@@ -441,7 +443,7 @@ public function getReplaces()
     /**
      * Set the recommended packages
      *
-     * @param array $devRequires A set of package links
+     * @param Link[] $devRequires A set of package links
```
stof: should be be done for other Link list properties too

alcohol:
I agree with @stof, if you're opening a PR for adding docblocks, it would be highly preferable if you add/fixup all relevant docblocks in the files you touch :-)

Soullivaneuh:
@stof @alcohol done. :+1:

alcohol:
Thanks

Issue #3542, opened 2014-12-12
Composer Update Fails due to Github Authorization


fiatux:
During composer update I am asked for my github credentials repeatedly. What may be causing this?

> An existing OAuth token for Composer is present and will be reused
> 
> Could not fetch https://api.github.com/authorizations, enter your GitHub credentials to go over the API >rate limit
> The credentials will be swapped for an OAuth token stored in /Users/[USER]/.composer/auth.json, your >password will not be stored
> To revoke access to this token you can visit https://github.com/settings/applications

Yes, I am entering my password correct, I have tried several different accounts and I can get authorization from Github if I use curl from cmdline.

yoanisgil:
@hugofonseca Thanks!

Commifreak:
The global Token config does not work (as mentiod in Post 3):

```
C:\Users\robin.kluth>composer diag
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity: OK
Checking HTTP proxy: OK
Checking HTTP proxy support for request_fulluri: OK
Checking HTTPS proxy support for request_fulluri:
Could not fetch https://api.github.com/repos/Seldaek/jsonlint/zipball/1.0.0, ent
er your GitHub credentials to go over the API rate limit
The credentials will be swapped for an OAuth token stored in C:/Users/robin.klut
h/AppData/Roaming/Composer/auth.json, your password will not be stored
To revoke access to this token you can visit https://github.com/settings/applica
tions
Username:
```

while in %appdata%\Composer\auth.json:

```
{
    "http-basic": {},
    "github-oauth": {
        "github.com": "my_token"}
}

```

If I add the token to projects composer.json, it is working as expected.

webdevilopers:
So the solution is to put the token into the `composer.json` directely instead of relying on the `auth.json`?

Will this be solved as suggested in the latest issue https://github.com/composer/composer/issues/3566 @Seldaek ?

Because as Malcolm Fell @emarref mentions:
`If you're working on a collaborative project, you should not commit your key in the project source control. The alternative is to set your key in your home directory...`
https://coderwall.com/p/kz4egw/composer-install-github-rate-limiting-work-around

Commifreak:
Hi,

I figured out, that the auto-generated token (via composer) ist not accepted. If I create the key via GitHub's WebUI and replace the old one in composer.json, it is working.

webdevilopers:
Have you tried that within the `auth.json` or `config.json` too?

[...]

Issue #1323, opened 2012-11-13
Improve git repo caching by cloning with `--reference` to a local `--mirror` clone


patcon:
As per #915, @Seldaek and @chEbba were interested in the idea of using `git clone --reference` and `--mirror` to cache git repositories and decrease build time.

More information is available on the git-clone man page:
http://git-scm.com/docs/git-clone

To be fair, the idea isn't my own and comes from how Drupal's drush CLI tool handles caching for Drupal modules pulled in from git repos:
https://github.com/drush-ops/drush/blob/master/commands/pm/package_handler/git_drupalorg.inc#L56-89

This appears to have been implemented by @msonnabaum, so perhaps he has input.

alexislefebvre:
I had the same idea while seeing that Travis CI clone git repositories for each build, which takes a lot of time. The Travis CI cache could be used to store these git mirrors and it would greatly speed up Composer.

Will this idea be implemented in the future? Thanks.

barryvdh:
In that case it would be better to download archives, without using the api as discussed in https://github.com/composer/composer/pull/4737 and cache that.

Seldaek:
Closing in favor of  #3722

Issue #3722, opened 2015-02-09
Faster git clones when the cache has a clone of the repo with Gith 2.3


stof:
when the composer cache has a clone of the repo already (when using the ``git`` driver), Composer could use the ``--dissociate --reference path/to/git/clone`` option to make the clone faster (this was not safe for older versions of Git because hard references require the referenced repo to keep existing but the new ``--dissociate`` option solves this)

AD7six:
:+1: 

That aught to greatly speed up cloning repeatedly-used dev dependencies.

burzum:
:+1:

netstyler:
:+1:

stof:
@AD7six for anything coming from github, they are not in the composer cache currently, so it would not help much in this case (we would either have to clone in the cache first to create it, or to use another local project as reference but we don't know where they are)

AD7six:
@stof oh? Would it not make sense for dev dependencies to:

 * Use the cache
 * Always update the cache first
 * create dev dependencies using `--dissociate --reference cache` (or even, just clone directly from the cache)

?

I can take a look at implementing that, would also open the door to allow composer to work offline... which may or may not be useful, depending on how often one develops whilst travelling.

hkdobrev:
I think either this or #1323 should be closed in favour of the other.

PhilLello:
I'm working on a patch for this at the moment, should be available at https://github.com/dunlop-lello/composer over the next few days. I haven't used --dissociate at the moment; tbh I'm not convinced there's a need, since people should be pushing their repos off-machine anyway if they care about error recovery, and could simply re-download. That said, I'm open to persuasion, but the horror stories I've read all seem to be from people who didn't have an off-machine backup of their repo.

PhilLello:
WIP Patch:

```
index 94a8d78..df2e200 100644
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -42,14 +42,28 @@ public function doDownload(PackageInterface $package, $path, $url)
     {
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
+        $cachePath = $this->config->get('cache-vcs-dir') . '/' . preg_replace('{[^a-z0-9.]}i', '-', $url);
 
         $ref = $package->getSourceReference();
         $flag = Platform::isWindows() ? '/D ' : '';
[...]

Issue #3060, opened 2014-06-18
Interpret environment variables in COMPOSER_BIN_DIR value


greg0ire:
I'm trying to install phpunit globally on my machine. I'd like to put my global composer.json under version control, and use it on other machines where my `$HOME` is not the same.

I've tried using `~` and `$HOME` as a value for the `bin-dir` config key, without success (the latter creates a directory named `$HOME`, litterally.

PR #4685, opened 2015-12-09
Speed git downloads by using cache dir as reference

luisfaceira:
Proposed solution for #1323

When downloading a git repository, it checks if that repository was already cached under cache-vcs-dir and if it was, use it as a "--reference".

By using this option, when downloading objects, git checks if those objects exist in that local mirror first, greatly reducing the network bandwidth usage, speeding up composer when reusing a package across multiple projects. See https://git-scm.com/docs/git-clone

The cache was already being created, when composer analyzes the dependencies of the source repo, so I only needed to check if the cache dir already existed.

Does anyone see any drawbacks of using this approach?

I tested manually on multiple situations (new repo, update, install, repo with new commits, etc.).

BTW, the original motivation to this was that trying to find a work-around to #3542 different than setting oauth tokens (which bring shared-security issues), I initially htought of using --prefer-source and was surprised that no cache was being used. Maybe with this, the performance penalty of using --prefer-source reduces enough for it to be an alternative to downloads from github api?

------------
> This pull request was made during the Open Source Day @ [beubi](http://beubi.com) :green_heart:

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -49,7 +49,6 @@ class Config
         'platform' => array(),
         'archive-format' => 'tar',
         'archive-dir' => '.',
-        'home' => '$HOME'
         // valid keys without defaults (auth config stuff):
         // github-oauth
         // gitlab-oauth
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -49,6 +49,7 @@ class Config
         'platform' => array(),
         'archive-format' => 'tar',
         'archive-dir' => '.',
+        'home' => '$HOME'
         // valid keys without defaults (auth config stuff):
         // github-oauth
         // gitlab-oauth
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -41,10 +41,12 @@ class GitDownloader extends VcsDownloader
     {
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
+        $cachePath = $this->config->get('cache-vcs-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $url).'/';
+        $cacheOptions = file_exists($cachePath) ? $cacheOptions = '--reference '.$cachePath.' ' : '';
 
         $ref = $package->getSourceReference();
         $flag = defined('PHP_WINDOWS_VERSION_MAJOR') ? '/D ' : '';
-        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
+        $command = 'git clone --no-checkout %s %s '.$cacheOptions.'&& cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
         $this->io->writeError("    Cloning ".$ref);
 
         $commandCallable = function ($url) use ($ref, $path, $command) {
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -42,7 +42,7 @@ class GitDownloader extends VcsDownloader
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
         $cachePath = $this->config->get('cache-vcs-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $url).'/';
-        $cacheOptions = file_exists($cachePath) ? $cacheOptions = '--reference '.$cachePath.' ' : '';
+        $cacheOptions = file_exists($cachePath) ? '--reference '.ProcessExecutor::escape($cachePath).' ' : '';
 
         $ref = $package->getSourceReference();
         $flag = defined('PHP_WINDOWS_VERSION_MAJOR') ? '/D ' : '';
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -42,7 +42,7 @@ class GitDownloader extends VcsDownloader
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
         $cachePath = $this->config->get('cache-vcs-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $url).'/';
-        $cacheOptions = file_exists($cachePath) ? '--reference '.ProcessExecutor::escape($cachePath).' ' : '';
+        $cacheOptions = file_exists($cachePath) ? '--dissociate --reference '.ProcessExecutor::escape($cachePath).' ' : '';
 
         $ref = $package->getSourceReference();
         $flag = defined('PHP_WINDOWS_VERSION_MAJOR') ? '/D ' : '';
--- a/tests/Composer/Test/Downloader/GitDownloaderTest.php
+++ b/tests/Composer/Test/Downloader/GitDownloaderTest.php
@@ -26,6 +26,10 @@ class GitDownloaderTest extends \PHPUnit_Framework_TestCase
         if (!$config) {
             $config = new Config();
         }
+        if (!$config->has('home')) {
+            $tmpDir = realpath(sys_get_temp_dir()).DIRECTORY_SEPARATOR.'cmptest-'.md5(uniqid('', true));
+            $config->merge(array('config' => array('home' => $tmpDir)));
+        }
 
         return new GitDownloader($io, $config, $executor, $filesystem);
     }


src/Composer/Downloader/GitDownloader.php
```diff
@@ -41,10 +41,12 @@ public function doDownload(PackageInterface $package, $path, $url)
     {
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
+        $cachePath = $this->config->get('cache-vcs-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $url).'/';
+        $cacheOptions = file_exists($cachePath) ? $cacheOptions = '--reference '.$cachePath.' ' : '';
```
staabm: Does it miss a `ProcessExecutor::escape()` to not error on some special chars?

staabm:
Do you have some number before/after the patch?

luisfaceira:
@staabm I do. I was developing with a single git repo included (~90M download) and got the following average times:
* first composer install: 14 to 18s
* following composer install (after removing lock and vendor): 1.5 to 3s
* update with a simple additional commit to be downloaded: 9 to 10s

staabm:
How long took it without your patch?

luisfaceira:
Without the patch it takes about the same time that the first composer install, but in every execution, not just the first one, because it was always downloading the 90M repo.

src/Composer/Downloader/GitDownloader.php
luisfaceira: Hmmm maybe it does, I'm not sure.

Special chars where, in the setting of "cache-vcs-dir" ?
The preg_replace applied was extracted from similar usages in the VcsDrivers, thought that would be enough.

Do you think I need to add that ProccessExecuter::escape() ?

staabm: If it works in other drivers I guess it is ok

src/Composer/Config.php
```diff
@@ -49,6 +49,7 @@ class Config
         'platform' => array(),
         'archive-format' => 'tar',
         'archive-dir' => '.',
+        'home' => '$HOME'
```
alcohol: Not sure how this change relates to your PR?

src/Composer/Downloader/GitDownloader.php
alcohol: If you check line 51 (now 53) you will note that all paths are escaped using `ProcessExecutor::escape()`. The same should be done for the `--reference` argument value (`$cachePath`).

src/Composer/Config.php
luisfaceira: When I added the usage of $this->config->get('cache-vcs-dir') , the GitDownloaderTest started failing with notices, because in that context, the config wasn't fully initialized, and the 'home' config was missing. Setting it in the default configs solved it.

Setting a default 'home' config that is automatically

src/Composer/Downloader/GitDownloader.php
luisfaceira: Ok. I haven't yet understood completely in what situation this might be necessary in this specific case where I'm getting a dir from conf, but better be safe than sorry :) will do the change right away!

src/Composer/Config.php
alcohol: @seldaek any idea if this might have adverse effects?

cs278:
I think you should use `--dissociate --reference /path/to/cache` otherwise the cloned repository is dependant on the cache directory existing, if you remove the cache directory the repo in vendor will break. The downside is the network I/O saving will only apply to initial clones.

luisfaceira:
Hmmm I was convinced that --reference would set an "alternate" source, and therefore, if it was not present, it would proceed normally. But after digging after your question, @cs278, it seems you're right, its safer to use --dissociate (also see #3722 ).
I will add it and try to see if I can reference it again temporarily on updates as well.

src/Composer/Config.php
stof: Well, this would not work. You would get ``$HOME/cache`` as cache dir, instead of resolving the env variable. IMO, the right fix is making sure the config you pass to the test is initialized properly

luisfaceira: Can you please examplify in what situations won't it work, @stof, I'm not understanding.
I haven't debugged it in such detail, but from looking at [Composer/Config.php#L181](https://github.com/luisfaceira/composer/blob/faster-git-clones-using-cache/src/Composer/Config.php#L181) I would expect you to get the full cache dir expanded from your environment such as `/home/myuser/.composer/cache` and not simply `$HOME/cache` as you said.

And if it does not work, what would you suggest to be a proper initialization in the test - explicitly setting in runtime the home config from the home environment, when initializing the Config class?

luisfaceira: Here's the issue that I think that made it possible to use `$HOME`: #3060

stof: but ``$config->get('home')`` would be broken in your case though

stof:
Note that this will only affect repositories added as git repositories in your composer.json. Any git repository for a package loaded from Satis or Packagist won't be in your vcs dir cache

src/Composer/Config.php
luisfaceira: Yes, I think you're right about the `$config->get('home')`, it won't expand properly the usage of $HOME inside, in the current implementation of that feature.

I'm not sure if I should improve (possibly in another PR?) what has been implemented in 449f8165ef64829dc34be836235ceb74d9227dd7 by @seldaek and make $HOME be interpreted also when used in the 'home' config (although it's not called "-dir" it is a dir config, after all, is it not?), or if I should instead force setting the home config value when creating the object for the GitDownloaderTest.

What do you think?

luisfaceira:
@stof I was aiming for that case, of git repositories added in composer.json. I'm not sure if I can squeeze the time in the weekend, but I'll try to expand this to cover more cases, and I'll also try to look into the option of using shallow cloning to speed things up even more (in separate branches).

src/Composer/Config.php
luisfaceira: After thinking this through and all the debate that followed this change, I decided that it would be a lot safer to revert it, and instead change the test to set a temp dir as home when creating the Donwloader Mock.

tests/Composer/Test/Downloader/GitDownloaderTest.php
```diff
@@ -26,6 +26,10 @@ protected function getDownloaderMock($io = null, $config = null, $executor = nul
         if (!$config) {
             $config = new Config();
         }
+        if (!$config->has('home')) {
```
staabm: Should we do a cleanup of this directory in the test teardown?

luisfaceira: My change only ensures that an home config is set, because there was some pre-existing code in [Config](https://github.com/composer/composer/blob/master/src/Composer/Config.php#L222) that was expecting it to exist.

I've only set it up as a dir inside /tmp/ to make sure that it would not break if something in the future actually needs a valid dir, and used the exact same technique to generate a random name that was being used by others in this test, for consistency. But, at least with the current tests, the home dir isn't actually created, so no teardown related to my change is necessary, at least for now. 

However, there are a bunch of dir and files that tests create (both in the GitDownloaderTest and other tests), at least in my /tmp/ after running phpunit I find a lot of leftovers:
> cmptest-2003111fcf4331c420d8c61ef429410f
> cmptest-3c7103f6ac29a4865b351466a4a730da
> cmptest-a71ac06271f7e35caa3c05f03206e459
> composer-test
> composer_testdir
> composer_test_file
> composer-xz-test-vendor
> composer-zip-test-vendor`

IMHO, files created inside /tmp/ do not absolutely require teardown, that would be a nice-to-have. But if it's considered to be relevant, maybe we should open an issue for it?

alcohol: Agreed, it's not relevant to this PR. Don't worry about it.

staabm: > But if it's considered to be relevant, maybe we should open an issue for it?

Having this information in a separate Issue to be fixed in a separate PR would be great IMO

luisfaceira: opened #4694 to do so in all tests that currently create dirs

hkdobrev:
We are using `config.preferred-install` to install our own organisation packages from source and this would be a very welcome change. Are there any updates on this?

Seldaek:
Nope, stlil have to give it a try

Seldaek:
Closing in favor of #5384 which includes these changes

PR #4690, opened 2015-12-10
Replaced relative path with absolute path in order to fix symlink on windows

DimaSoroka:
Fix symlink issue on windows
```
  [ErrorException]
  symlink(): Could not fetch file information(error 3)

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -48,8 +48,7 @@ class PathDownloader extends FileDownloader
         }
 
         try {
-            $shortestPath = $this->filesystem->findShortestPath($path, $realUrl);
-            $fileSystem->symlink($shortestPath, $path);
+            $fileSystem->symlink($realUrl, $path);
             $this->io->writeError(sprintf('    Symlinked from %s', $url));
         } catch (IOException $e) {
             $fileSystem->mirror($realUrl, $path);


alcohol:
Are you saying relative symlinks don't work on Windows? @Seldaek can you experiment with this?

Seldaek:
Hmm yeah I think this might well be the case.. In general relative symlinks kinda suck as they don't deal well with being moved around. I am not sure why they were relative to begin with though.

cs278:
Relative targets are used to allow for a file system being mounted at a different root, for example inside a VM. #4426 I suppose Composer could fallback to an absolute symlink and issue a warning.

DimaSoroka:
There is a comment in [php documentation](http://php.net/manual/en/function.symlink.php)

>Symlinks on windows  are created by Symlink() which accept only absolute paths  but not relative paths .relative paths on windows are not supported for symlinks

Unfortunately it is still valid. 

As an option I could add a configuration to "path" repository that will allow to configure symlink settings (absolute or relative path)

cs278:
Relative symlinks are much more useful from a developer perspective. It sounds like the PHP `symlink()` implementation is broken on Windows, what about calling out to [`mklink`](https://technet.microsoft.com/en-us/library/cc753194.aspx) on Windows?

alcohol:
From the documentation of `mklink`

> `<Target>` Specifies the path (relative or absolute) that the new symbolic link refers to.

Seems to suggest that relative symlinks should be possible.

xabbuh:
If `symlink()` does not work as expected on Windows we can maybe add a fallback using `mklink` in the Filesystem component.

By the way, this method should probably complete abstain from creating a new `Filesystem` instance and reuse the property from the parent class instead.

Seldaek:
@xabbuh we have two Filesystem classes here, one from Composer and one from Symfony. That's why the two instances.

Anyway I'd say @cs278's idea about falling back to absolute if relative fails sounds like the best approach. Or alternatively attempt using mklink on windows. But changing to always absolute sounds like a no-go.

DimaSoroka:
Thanks guys, will work on update.

curry684:
@DimaSoroka you still working on this? If not could you clarify what's left to be done?

DimaSoroka:
This is on hold for now, in order to resolve this issue i see a few options:
1. Resolve relative path support for Windows on Symfony\Component\Filesystem\Filesystem::symlink level
2. Collaborate with PHP team on symlink function implementation

curry684:
I've just did some extensive debugging and came to the conclusion this *cannot be fixed*, as the relative paths are not the root issue. The problem is that Windows considers symlinking an invasive insecure operation (not unjust), and therefore by default has a Local Security Policy in place to restrict it to Administrators ([link](http://stackoverflow.com/a/15692658/1729885)).

Now even if you lift this restriction, or are an Administrator, you'll still need elevation. I have confirmed this on a test machine where I am full local administrator, opening 2 command prompts of which one as administrator, and running `php -r "symlink('original', 'symlink');"`. The non-elevated one consistently fails (error 1314), the elevated one works fine and creates the link.

As noone is (in their right mind) going to run Composer exclusively from an elevated command prompt, and PHP does not expose UAC to elevate 'as required', it would seem this issue cannot be fixed, and current behavior is just fine.

Seldaek:
I think it might still be worth supporting it even if it requires an admin
shell. Typically you only need to run it once in a project to set up the
path repo then the symlink wouldn't need updating so it might be worth the
trouble to people.

You are right though that it can not be fixed for the general case. I do
wonder if abusing mklink as @cs278 suggested above we couldnt make it work
without admin.

curry684:
I'll be happy to look into it if I can find a practical way to debug, as I didn't manage to get my xdebug setup running with elevated privileges.

The `mklink` hack won't work, just tried in an unelevated shell:

    C:\Projects\Composer>mklink /D test src\Composer\Autoload
    You do not have sufficient privilege to perform this operation.

Seldaek:
Junctions using `mklink /D /J linkName targetPath` work for me without elevation. I don't know if that's enough for php and all to work, needs experimenting.

curry684:
Junctions may solve the issue but it should be noted that they can't be relative - they're essentially hard links on a directory within local filesystems ([reference](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365006%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396)).

Also, I have no clue what happens if you `rmdir \S vendor` while `vendor/myname/mypackage` is a junction to `C:\Projects\myDevPackage`. I think it could end up very painful and being exactly the case why you'd want symlinks.

Seldaek:
Yeah it definitely needs some testing to make sure it doesn't end in 
tears, but the absolute links restriction I don't think is such a big issue.

curry684:
The scenario mentioned above works fine - I was able to set it up with relative paths, and `rmdir /S` left the original files intact. Also, all of this worked without elevation, so it seems to be a valid solution. 

Main risky case to test is in the remove commands, and how to detect on that end that the folder should be unlinked instead of recursively deleted. Recursive deletion **will* be painful  :)

curry684:
Well that went to shit really fast. I have the junction working fine on install, working great.

Then I tried to remove it, and it failed to kill the junction, but killed the referenced folder, leaving it empty but without any ACLs whatsoever - can't even delete it anymore as administrator. Still not entirely sure this is a feature that should be shipped  ;)

curry684:
Think I've got it covered now, extended Filesystem with reliable junction detection and removal. Will review some more before making a PR though.

curry684:
@DimaSoroka can you test/review if #4845 properly implements this behaviour? In my limited test case both `require` and `remove` are working fine now on Win10 Enterprise.

**Disclaimer:** I would recommend making a thorough backup beforehand of the path repositories used in testing this hehe. I think the code is solid now but it's 1am here as well.

DimaSoroka:
works just fine, thanks @curry684!

Seldaek:
Closing as #4845 replaces it

Issue #3836, opened 2015-03-11
Show a list of all dependencies


evert:
It would be nice if I could ask composer to give me a list of all dependencies and their version constraint, including system packages.

This would allow me for example quickly find out what the minimum PHP version is based on what every package needs.

This can be either considered a feature request or a support question =)

jakoch:
You might be looking for the CLI command `show` - https://getcomposer.org/doc/03-cli.md#show

- `composer show --platform` - lists your platform/env = PHP + Extensions
- `composer show --installed` - lists your platform/environment on top of all available packages
- `composer show --self` - basically the output of composer.json of your project
   - see `require` section, `php`
- `composer show monolog/monolog` - package details
- `composer show monolog/monolog 1.2.3` - package details (specific version)

evert:
None of that does what I asked.

I would like to see a list of actual dependencies plus their version constraints, `composer show` just shows how those dependencies were resolved.

To reiterate an example of an actual use-case:

> This would allow me for example quickly find out what the minimum PHP version is based on what every package needs.

Seldaek:
The best I got is this.. it's not absolutely safe because it doesn't only look at lines in require, but it's pretty close:

```
$ for pkg in `composer show -i --name-only`; do composer show -i $pkg | grep 'php '; done
php >=5.3.2
php >=5.3,<8.0-DEV
php >=5.3.2
php >=5.3.3
...
```

evert:
That does help =) thnx

XedinUnknown:
I would like to see this functionality too. I'm using the [`composer-merge-plugin`](https://github.com/wikimedia/composer-merge-plugin) which is basically a critical part of my solution. I need to see a list of all dependencies, including those merged. But even without the plugin, it would be very useful to get a list of dependencies, and their version constraints. Also, this would help me resolve issue #4497.

XedinUnknown:
To be honest, I would like to see more options of getting structured output in a machine-readable format. Besides allowing other CLI utilities to use Composer output, it would also make the creation of a web interface possible.

Seldaek:
In any case, this isn't very hard to implement, the main question is how to define the interface, I'd rather not have a command that only lists dependencies otherwise we end up with a million commands, having a good generic interface to query this data would be nice. Someone has to spend the work to think and spec it out though.

XedinUnknown:
Sure.

[...]

PR #4696, opened 2015-12-11
Nice to have feature - Show a list of all dependencies #3836

gomboc:
https://github.com/composer/composer/issues/3836
Show a list of all dependencies #3836 - constraints option for show command - a list of all dependencies and their version constraint, including system packages


Seldaek:
Oh.. we kinda forgot to update this ticket I think, the tree command kinda does this. I'll write a comment over at #3836 then we see.

PR #4698, opened 2015-12-12
Improve error message about platform php version

WouterJ:
When seeing the error message, I though "config.platform.php" was some file/value configured for platform.sh instead of a setting value in composer.json. I propose to make it a bit more clear that this value is set in composer.json

### DIFF
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -207,7 +207,7 @@ class Rule
                         } elseif ($targetName === 'hhvm') {
                             $text .= ' -> you are running this with PHP and not HHVM.';
                         } else {
-                            $text .= ' -> your PHP version ('. phpversion() .') or "config.platform.php" value does not satisfy that requirement.';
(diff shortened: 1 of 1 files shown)


src/Composer/DependencyResolver/Rule.php
```diff
@@ -207,7 +207,7 @@ public function getPrettyString(Pool $pool, array $installedMap = array())
                         } elseif ($targetName === 'hhvm') {
                             $text .= ' -> you are running this with PHP and not HHVM.';
                         } else {
-                            $text .= ' -> your PHP version ('. phpversion() .') or "config.platform.php" value does not satisfy that requirement.';
+                            $text .= ' -> your PHP version ('. phpversion() .') or "config.platform.php" setting in composer.json does not satisfy that requirement.';
```
alcohol: I would phrase it as `value in composer.json` instead of `setting`, how does that sound to you? :)

WouterJ: I changed it to `value of "config.platform.php" in composer.json`.

alcohol:
Cheers

Issue #4694, opened 2015-12-11
Clean temporary dirs that are created on tests during teardown


luisfaceira:
On a quick run of phpunit, I've detected that the following dirs in /tmp/ are created and remain after the tests concluded successfully:
> cmptest-2003111fcf4331c420d8c61ef429410f
> cmptest-3c7103f6ac29a4865b351466a4a730da
> cmptest-a71ac06271f7e35caa3c05f03206e459
> composer-test
> composer_testdir
> composer_test_file
> composer-xz-test-vendor
> composer-zip-test-vendor

Some of them I've detected to be in GitDownloaderTest, but others are being created elsewhere (I haven't had the opportunity to look into it). Also, although it's unlikely I'm not sure if other directories are being created in other tests in other places other than /tmp/ which would be a lot worse.

As @staabm proposed in the context of [#4685](https://github.com/composer/composer/pull/4685/files#r47354731), it would be best to remove such directories during the teardown of the related tests. 

I detected a similar fix to one specific test that was done in the past: 8d3b40ff8ee5b8f550e187b141df4d00626aefed . I think the same approach could be easily replicated to other tests.

stof:
``composer-zip-test`` and ``composer-xz-test`` are related to the ZipDownloader and the XzDownloader (a simple search on the source code detects them).

``cmptest-...`` are for GitDownloaderTest, HgDownloaderTest and AutoloadGeneratorTest

and then, ``Composer\Util\Filesystem`` tests are also creating temp files.

And ``composer-test`` is used in many tests, but only some of them are removing it

chorry:
whoops, sorry for the mess - had no idea `--force push` will leave footprints here instead of replacing commit :(

stof:
@chorry force push creates a new commit. but this does not change the fact that the old commit references the PR.

chorry:
is it ok to close old pull request, and start a new one?

stof:
@chorry this is the issue, not the PR. the PR itself is clean (but not referenced here as you haven't referenced the issue in it), as it is about the branch, which is updated when doing a force push

Seldaek:
Fixed by #4702

PR #4702, opened 2015-12-14
clean garbage dirs in tmp after composer tests #4694

chorry:


tests/Composer/Test/CacheTest.php
```diff
@@ -12,15 +12,20 @@
...
 {
     private $files, $root, $finder, $cache;
+    /**
+     * @var Filesystem
+     */
+    private $fs;
 
     public function setUp()
     {
+        $this->fs = new Filesystem;
```
stof: no need to instantiate it on setUp and to store it in a property. Instantiate it only when necessary

tests/Composer/Test/Downloader/GitDownloaderTest.php
```diff
@@ -18,6 +18,25 @@
...
+    /** @var Filesystem */
+    private $fs;
+    /** @var string */
+    private $workingDir;
+
+    protected function setUp()
+    {
+        $this->fs = new Filesystem;
+        $this->workingDir = realpath(sys_get_temp_dir()).DIRECTORY_SEPARATOR.'cmptest-'.md5(uniqid('', true));
+        $this->fs->ensureDirectoryExists($this->workingDir);
```
stof: IMO, this line should be removed, as several tests don't need it (and tests needing it are actually already calling it to ensure that a subfolder exists)

tests/Composer/Test/Downloader/HgDownloaderTest.php
```diff
@@ -17,6 +17,25 @@
...
+    /** @var Filesystem */
+    private $fs;
+    /** @var string */
+    private $workingDir;
+
+    protected function setUp()
+    {
+        $this->fs = new Filesystem;
+        $this->workingDir = realpath(sys_get_temp_dir()).DIRECTORY_SEPARATOR.'cmptest-'.md5(uniqid('', true));
+        $this->fs->ensureDirectoryExists($this->workingDir);
```
stof: same here

tests/Composer/Test/Downloader/PerforceDownloaderTest.php
```diff
@@ -30,10 +31,13 @@ class PerforceDownloaderTest extends \PHPUnit_Framework_TestCase
     protected $repoConfig;
     protected $repository;
     protected $testPath;
+    protected $fs;
 
     protected function setUp()
     {
+        $this->fs = new Filesystem;
         $this->testPath        = sys_get_temp_dir() . '/composer-test';
+        $this->fs->ensureDirectoryExists($this->testPath);
```
stof: useless, as most tests don't need to have this folder. There is no reason to create it for nothing

tests/Composer/Test/Downloader/XzDownloaderTest.php
```diff
@@ -13,14 +13,32 @@
...
+     * @var string
+     */
+    private $testName;
+
     public function setUp()
     {
         if (defined('PHP_WINDOWS_VERSION_BUILD')) {
             $this->markTestSkipped('Skip test on Windows');
         }
+        $this->fs = new Filesystem;
```
stof: this should be instantiated only on teardown

tests/Composer/Test/Downloader/ZipDownloaderTest.php
```diff
@@ -13,14 +13,32 @@
...
+     * @var string
+     */
+    private $testName;
+
     public function setUp()
     {
         if (!class_exists('ZipArchive')) {
             $this->markTestSkipped('zip extension missing');
         }
+        $this->fs = new Filesystem;
```
stof: same here

tests/Composer/Test/Downloader/ZipDownloaderTest.php
```diff
@@ -13,14 +13,32 @@
...
 {
+    /**
+     * @var Filesystem
+     */
+    private $fs;
+
+    /**
+     * @var string
+     */
+    private $testName;
```
stof: ``testDir`` would be a better name than ``testName``

alcohol:
It seems to randomly fail the `InstallerTest` cause of tmp directory issues. I restarted the build but then it just failed on another PHP version. Going to see if anyone from travis can give feedback on that :-\

chorry:
I see travis is now marked as passed, but appveyour is failed due to failed provisioning of VM. 
Can somebody restart it?

Seldaek:
Seems fine on travis and on my machine so let's hope appveyor will go better in the future, thanks for the fix!

Issue #4715, opened 2015-12-18
sort-packages as config settings


hanovruslan:
Hi!

For many (our) projects we run 

```
composer require --sort-packages some/package:@some-version 
```

so it will be nice to have sort-packages as global settings with false by default according to the current RequireCommand options

I've already created some PR (not pushed to this repo for now) but i stacked at the test stuff

Did not find test for command RequireCommand

So what what should i do?

P.S.: Is there any existing PR with such behaviour ? did not found such

alcohol:
I'm personally not a fan for having each command line argument/flag also available as configuration option. But if anyone is willing to spend time on this, go ahead I guess.

hanovruslan:
@alcohol done

PR #4716, opened 2015-12-18
Added sort-packages into config

hanovruslan:
#4715 Added sort-packages into config

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -126,6 +126,10 @@ autoloader. When null a random one will be generated.
 
 Defaults to `false`. If `true`, always optimize when dumping the autoloader.
 
+## sort-packages
+
+Defaults to `false`. If `true`, Keep packages sorted in `composer.json`.
+
 ## classmap-authoritative
 
 Defaults to `false`. If `true`, the Composer autoloader will only load classes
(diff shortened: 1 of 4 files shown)


alcohol:
Please also add a way to programmatically set this new config flag while you are at it.

alcohol:
- Needs a test: just add something to the require scenarios, make sure it runs with the config set to true, verify the result is sorted.
- Needs documentation.

hanovruslan:
@alcohol 
```
Needs a test: just add something to the require scenarios
```

please tell in which (test?) file ?

alcohol:
I had a look at [tests/Composer/Test/Fixtures/installer](/composer/composer/tree/master/tests/Composer/Test/Fixtures/installer), but it actually does not seem possible to write a test for this particular scenario with the current tools available I'm afraid. Was hoping that we would be able to check that given a composer.json and running require x, we could check the resulting composer.json; but alas no such use-case is yet available among the installer tests.

alcohol:
Just some documentation will be sufficient. Check the CLI and config document.

hanovruslan:
(Y)

hanovruslan:
@alcohol [this](https://github.com/hanovruslan/composer/commit/38615181c64be6f4272df4a551e46a98c215f001) would be enough ?

Seldaek:
Thanks

PR #4719, opened 2015-12-19
more simple and better autoload_real code

cmmer:


alcohol:
I don't think breaking every single test qualifies as "better".

Issue #4484, opened 2015-10-06
composer_autoload.php can require files that do not exists.


daiglej:
If in composer.json of you requirement contains like this 
"autoload": {
"file":["src/somefile.php"]
}

In the HEAD of it's repository, but you have required a commit hash that is older than the HEAD (One that does not yet contain the file). The file will still be part of the composer_autoload.php.

then when in your code when you will "require vendor/autoload.php".

You will get a fatal error :  

Fatal error: require(): Failed opening required '.../vendor/cloudinary/cloudinary_php/src/Helpers.php' in .../vendor/composer/autoload_real.php on line 59

Note : you can reproduce this issue by doing : 
./composer.phar require cloudinary/cloudinary_php  dev-master#62c4da6

stof:
The ``dev:master#...`` syntax is an installer-level hack. Metadata downloaded from Packagist are not available for each commits (this would be insane, and would require running Composer with 1TB RAM to resolve dependencies). This is why the metadata contain the new autoload requirements.

My recommendation is to never use this syntax. If you want to use older versions, you should rely on tags instead.

alcohol:
It is known that using a #ref in your version constraint comes with various caveats. You simply have to learn to live with those. There is no intention on "fixing" any of this. Using a #ref in your version constraint is very bad practice and should be avoided at all times.

stof:
@alcohol I suggest adding a warning for this in ``composer validate`` (we already warn for unbound constraints). what do you think ?

alcohol:
That is definitely a good idea :-)

alcohol:
@stof #4485 if you don't pick it up I will, this weekend probably :-)

PR #4737, opened 2015-12-28
Use github.com instead of api.github.com to fetch zip files

nicolas-grekas:
This is just a patch that proves that fetching zip files on this github.com endpoint works the same as the one on api.github.com, but is not affected by rate limiting. This could speed up builds on travis by a significant factor, because we wouldn't be forced anymore to use `--prefer-source`

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -86,10 +86,17 @@ class FileDownloader implements DownloaderInterface
         $this->io->writeError("  - Installing <info>" . $package->getName() . "</info> (<comment>" . $package->getFullPrettyVersion() . "</comment>)");
 
         $urls = $package->getDistUrls();
+        $processUrl = true;
+
+        if (isset($urls[0]) && preg_match('#^https://api\.github\.com/repos/([^/]++/[^/]++)/zipball/(.++)$#', $urls[0], $match) && !$this->io->hasAuthentication('github.com')) {
+            array_unshift($urls, 'https://github.com/'.$match[1].'/archive/'.$match[2].'.zip');
+            $processUrl = false;
+        }
         while ($url = array_shift($urls)) {
             try {
-                return $this->doDownload($package, $path, $url);
+                return $this->doDownload($package, $path, $url, $processUrl);
             } catch (\Exception $e) {
+                $processUrl = true;
                 if ($this->io->isDebug()) {
                     $this->io->writeError('');
                     $this->io->writeError('Failed: ['.get_class($e).'] '.$e->getCode().': '.$e->getMessage());
@@ -107,13 +114,13 @@ class FileDownloader implements DownloaderInterface
         $this->io->writeError('');
     }
 
-    protected function doDownload(PackageInterface $package, $path, $url)
+    protected function doDownload(PackageInterface $package, $path, $url, $processUrl = true)
     {
         $this->filesystem->emptyDirectory($path);
 
         $fileName = $this->getFileName($package, $path);
 
-        $processedUrl = $this->processUrl($package, $url);
+        $processedUrl = $processUrl ? $this->processUrl($package, $url) : $url;
         $hostname = parse_url($processedUrl, PHP_URL_HOST);
 
         $preFileDownloadEvent = new PreFileDownloadEvent(PluginEvents::PRE_FILE_DOWNLOAD, $this->rfs, $processedUrl);


alcohol:
It seems to download everything from source:

```
Failed to download composer/semver from dist: You must enable the openssl extension to download files via https
```

The above is repeated for virtually every dependency.

nicolas-grekas:
Are you sure this is related to the attached patch?

stof:
@alcohol the PHP 5.3.3 job on Travis does not have openssl, so this is normal. And note that this output does not come from new composer codebase, but from the installation using the Travis composer bin first.

alcohol:
Right, I forgot about that. I'll try it out locally.

alcohol:
I've also emailed github support; would prefer to get some feedback from them before we switch from using the api to the more web based/frontend oriented download url.

alcohol:
Any particular reason you did not make this modification in the `GitHubDriver` though?

nicolas-grekas:
Quick proofing the idea :-)

nicolas-grekas:
And also because the githubdriver isn't used when the URLs are provided by the packages.json/composer.lock files, isn't it?

alcohol:
I am not sure to be honest, but you could be right about that.

cs278:
> And also because the githubdriver isn't used when the URLs are provided by the packages.json/composer.lock files, isn't it?

Correct they go through the `ArchiveDownloader`.

This change breaks downloads from private repositories.

nicolas-grekas:
how can it break them?

cs278:
> how can it break them?

Because they require authentication using OAuth tokens which is not supported by the website.

alcohol:
Ah, he has a point there I think. Only the api accepts oauth token credentials.

nicolas-grekas:
@cs278 good point, this is fixed now, and the behavior is also less intrusive: the api.github.com url is used as a fallback of the github.com one, which itself is used only when no github-token is configured.

Nicofuma:
Shouldn't this logic be left to packagist (or any other repo)? They are already responsible for setting the dist entries of the composer.json after-all.

nicolas-grekas:
@Nicofuma packagist doesn't know when a package is switched from private to public or vice versa. Also, the github-token exists only on composer's side.

cs278:
> Shouldn't this logic be left to packagist (or any other repo)? They are already responsible for setting the dist entries of the composer.json after-all.

@Nicofuma This is a bit of client side magic to ease the way in certain cases, no GitHub auth and public repo. The authoritative download link is still `https://api.github.com/...` this is just the client working around the rate limiting on the API.

@nicolas-grekas :+1: Works great. With credentials downloads occur following previous behaviour. Without credential downloads hit `github.com` and and falls back to `api.github.com` so Composer isn't brittle to GitHub changing their web interface again. Without credentials hitting a private download Composer falls back to source in the same way as before.

alcohol:
I would still like to hear back from Github staff though before I can give my thumbs up on this change. But aside from that, good catch and a welcome addition if they're ok with it :)

alcohol:
Speaking of, I just got a reply:

> Thanks for reaching out and asking about this.

> The api.github.com endpoint is an official endpoint of the GitHub API:

> https://developer.github.com/v3/repos/contents/#get-archive-link

> That means that you can rely on the behavior of that endpoint and all the other parts of the API, like rate limiting.

> The github.com endpoint is _not_ an official endpoint of the GitHub API -- it's an internal endpoint used to power the github.com website (the button for downloading a ZIP archive). That means that you shouldn't rely on the behavior of that endpoint, or that the endpoint will even exist tomorrow.

> In short, the GitHub API was designed and optimized for machines, so scripts and any other programmatic usage should go through that. GitHub.com was optimized for humans, and humans should be using that.

> While I understand that rate limiting sounds like a restriction in some cases, you should also think of it from the perspective of what we can support and guarantee. The GitHub API has a documented rate limit you can rely on, while the github.com endpoint doesn't and you shouldn't rely on _any_ rate limit (especially not an infinite one). For that reason, scripts using github.com endpoints programmatically might be blocked at any time and without warning if we notice any problems.

> In short, I'd discourage you from relying on any endpoint not covered by the GitHub API, but if you want to use that approach -- I suggest you implement rate limiting on your end.

> I hope this answers your question, but let me know if there's anything else.

It seems to me like they recommend against it, for valid reasons.

nicolas-grekas:
Well, personally I'd go against this recommendation for the following reasons:
- the benefit is huge on travis, where everyone is currently slowed down because git clone is the only viable approach.
- both travis and github have incentive to provide such an endpoint: git clone is way more costly for them than http fetch, which means that if they decide to break this URL on gthub.com, we have arguments for them.
- and last but not least, the implementation is safe from changes on the github side side it falls back to the api.

As a last resort, I'd like to know if enabling this behavior could be accepted if controlled by an env var (let's say `COMPOSER_USE_ALT_GITHUB=1`)

alcohol:
- I don't really agree. I use tokens on all my personal repositories (and they're not even private).

- Github provides a solid endpoint and it really isn't hard to setup a token for their api.

- The implementation actually has zero effect for anyone properly using tokens already.

I'll leave this up to @Seldaek, but personally I'd say :-1: .

nicolas-grekas:
OK, was worth to try...

cs278:
> I don't really agree. I use tokens on all my personal repositories (and they're not even private).

That works but if you are securing them in Travis they won't be used to build pull requests, which is sound precaution. I wonder what GitHub's take would be on registering an unprivileged user, issuing an API key to use on Travis and storing it plain in the repo.

nicolas-grekas:
@cs278 that's what I did here https://github.com/symfony/symfony/pull/17171

alcohol:
I use the travis cli tool to encrypt them and store them in my `.travis.yml`.

Nicofuma:
@alcohol But in this case they won't be used to test the PRs...

barryvdh:
Personally, I would be in favor of this. I know setting up OAuth tokens is not hard, but I do regularly run into this issue when setting up remote deployments on new servers (without interactive shell).
Also related on the Drupal issue; #4484 and speeding up Travis CI builds would be another big plus.

If there is a fallback to the API URL, it shouldn't be a big problem if Github changes. It can be fixed in a new Composer release, but wouldn't break anything.

Atleast Bower also users github.com urls directly: https://github.com/bower/bower/blob/78e443db0afb1889a9ec53c2ff94cce1b1ab017b/lib/core/resolvers/GitHubResolver.js#L59

Seldaek:
@nicolas-grekas could you restore your branch so I can reopen this please?

nicolas-grekas:
Reopened

PR #4756, opened 2016-01-08
Fix tests for justinrainbow/json-schema 1.6

remicollet:
Detected by Fedora QA, see https://apps.fedoraproject.org/koschei/package/composer

justinrainbow/json-schema version 1.6 introduce new properties in error output.

This fix cleanup the array to only keep and check used properties.

Another solution will be to add this new properties, but this will imply to raise minimal dep. on json-schame which is probably not needed / wanted.

### DIFF
--- a/tests/Composer/Test/Json/ComposerSchemaTest.php
+++ b/tests/Composer/Test/Json/ComposerSchemaTest.php
@@ -19,23 +19,44 @@ use JsonSchema\Validator;
  */
 class ComposerSchemaTest extends \PHPUnit_Framework_TestCase
 {
+    private function extractNeeded($in)
+    {
+        if (!is_array($in)) {
+            return $in;
+        }
+        // Cleanup JsonSchema output
+        // Keep only used / checked properties
+        // So drop new properties introduced in 1.6
+        $out = array();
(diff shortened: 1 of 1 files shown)


AdrienLucas:
IMO, instead of extracting the needed values, you should maybe replace the ``assertEquals`` by some ``assertContains``, it would be a much more simple way to deal with that issue.

remicollet:
Yes various way to fix...
But we also need to check for no additional error messages. So I think assertEquals still better.

AdrienLucas:
I still think that you should not add a new private method to the test. Maybe try to put your extraction code in the ``check`` method.
Something like this :
```php
         if (!$validator->isValid()) {
             $errors = $validator->getErrors();
             return array_map(function($e){ return array_intersect_key($e, array('property'=>'', 'message'=>'')); }, $errors);
         }
```

In any case, I don't know why the project doesn't have some travis builds with something like ``deps=high`` (or any other "composer update" directive) to detect those kind of issue. Is there a reason?

Seldaek:
I just bumped the requirement so we can update the tests and not argue about how to best support both versions :)

PR #4759, opened 2016-01-10
SSL support improvements

Seldaek:
Replaces #2745


src/Composer/Command/DiagnoseCommand.php
```diff
@@ -172,12 +173,32 @@ private function checkGit()
...
+        if ($proto === 'https' && !extension_loaded('openssl') && !$disableTls) {
+            $result[] = '<error>Composer is configured to use SSL/TLS protection but the openssl extension is not available.</error>';
+        }
+
         try {
             $this->rfs->getContents('packagist.org', $proto . '://packagist.org/packages.json', false);
-        } catch (\Exception $e) {
-            return $e;
+        } catch (TransportException $e) {
+            if (preg_match('|cafile|', $e->getMessage())) {
```
staabm: Could be simplified to `strpos`

Seldaek: Done

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -291,10 +317,24 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
             }
         }
 
+        // Check if the failure was due to a Common Name mismatch with remote SSL cert and retry once (excl normal retry)
```
StormTide: I'd have to do some more tracing, but at first glance it looks like you're retrying the match from the CN provided in the certificate (which comes from the attackers perspective)? If so, then this defeats the CN match entirely (attacker controls the CN provided by the server, can make it for an arbitrary (but otherwise valid) alternative domain) ..

Seldaek: That's interesting. It was already in @padraic's original PR and I guess it's only there to catch cases where guessing the CN based on hostname fails miserably. I don't know if we could limit it to something more conservative like.. only allow CN editing if it's on the same secondary level domain? i.e. if we tried api.github.com and that fails but sends a CN github.com we would use that? But if it's blub.com we abort.

Would that help? I am not sure what are the capabilities in terms of CN + MITM.

src/Composer/Factory.php
```diff
@@ -503,4 +512,48 @@ public static function create(IOInterface $io, $config = null, $disablePlugins =
...
+     * @param IOInterface   $io         IO instance
+     * @param Config        $config     Config instance
+     * @param array         $options    Array of options passed directly to RemoteFilesystem constructor
+     * @return RemoteFilesystem
+     */
+    public static function createRemoteFilesystem(IOInterface $io, Config $config = null, $options = array())
+    {
+        static $warned = false;
+        $disableTls = false;
+        if (isset($config) && $config->get('disable-tls') === true) {
```
stof: for consistency with the codebase, it should be ``null !== $config`` here (and in other places using ``isset($config)``

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -486,4 +552,205 @@ protected function getOptionsForUrl($originUrl, $additionalOptions)
 
         return $options;
     }
+
+    protected function getTlsDefaults()
```
stof: shouldn't this be private ?

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -486,4 +552,205 @@ protected function getOptionsForUrl($originUrl, $additionalOptions)
...
+                }
+            } else {
+                throw new TransportException('A valid cafile could not be located automatically.');
+            }
+        }
+
+        /**
+         * Disable TLS compression to prevent CRIME attacks where supported.
+         */
+        if (version_compare(PHP_VERSION, '5.4.13') >= 0) {
```
stof: this should use ``PHP_VERSION_ID``

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -486,4 +552,205 @@ protected function getOptionsForUrl($originUrl, $additionalOptions)
...
+                    $caBundle = dirname($caBundle);
+                    if (is_dir($caBundle) && glob($caBundle.'/*')) {
+                        $found = true;
+                        break;
+                    }
+                }
+            }
+        }
+        if ($found) {
+            $found = $caBundle;
```
stof: using ``$found`` for both the ca file path and the boolean looks confusing to me

stof:
@Seldaek I think this should also be ported to the installer script once it is merged here (as well as backporting previous changes which were forgotten)

Seldaek:
What do you mean by "this" should be backported? I just ported the installer code here so I don't know if there is much to be changed there.

stof:
oh, I missed that the work happened on the installer side first here

src/Composer/Util/RemoteFilesystem.php
cs278: > I guess it's only there to catch cases where guessing the CN based on hostname fails miserably

That's the whole point though if the CN doesn't match I'd expect an error, unless this is working around some PHP weirdness.

Seldaek: Any opinion on my proposal @StormTide?

StormTide: The RFC's are pretty clear on what a CN match is, and for the most part modern versions of PHP follow the CN matching algorithm. Old versions ignore subject-alt-name, so there's some area to try a SAN match on a CN match failure on specific old versions of PHP. That said, I cant ever see allowing a CN match from one domain (eg secure.google.com to work from say playgrounddomain.google.com) ... doing so would radically change the security model. 

The bigger issue here is that the attacker-controlled CN is being used in the current code for the retry. You can never get the desired CN from the certificate (as this is in the attackers control)... purpose of the CN match is to create 'authenticated encryption', that is, such that Mallory between Alice and Bob cannot substitute another certificate. The CN match creates a relationship between the server you think you're contacting and the CA that signed (vouched for) the domain control via the CN. 

I'd stick with the RFC's, and rely on PHP's internal matching on current versions (drop the retry), and on older ones, try a subject-alt-name match. I would never pull the CN from the certificate, and I would never expand a FQDN Certificate into a Wildcard Certificate, as it would allow a MITM (former case) or break the expectations of the TLS RFCs (latter case).

Seldaek: Alright thanks @StormTide, I dropped this whole CN stuff then. And I only set CN_match + SNI_server_name to the URL's host if we run on php < 5.6 - and trust php core for higher versions.

Seldaek:
Fixed all comments, I'll now merge and hopefully it goes well for all users.

StormTide:
@Seldaek changes look good, I don't see any obvious/outstanding issues here re the tls implementation. Not that I've done a thorough audit or anything, but nothing jumps out at me. 

There's still the matter of disabling http schemes however. Cant forget about the installed base of composer.json files pointing at http sources. Once plain http support is disabled and with this working TLS support, you can probably close out #1074 .. So long as HTTP is allowed, though, its going to continue to be a significant RCE exposure.

Seldaek:
@StormTide yup I moved that to #4777 so it doesn't get forgotten

PR #4660, opened 2015-12-01
Add --with-replaces option on depends command

Soullivaneuh:
Closes #4657.

Output sample:

```bash
$ composer depends --with-replaces symfony/symfony
a2lix/translation-form-bundle requires symfony/symfony (>=2.3)
behat/behat requires (dev) symfony/process (~2.1)
behat/behat requires symfony/class-loader (~2.1)
behat/behat requires symfony/config (~2.3)
behat/behat requires symfony/console (~2.1)
behat/behat requires symfony/dependency-injection (~2.1)
behat/behat requires symfony/event-dispatcher (~2.1)
behat/behat requires symfony/translation (~2.3)
behat/behat requires symfony/yaml (~2.1)
behat/gherkin requires (dev) symfony/yaml (~2.1)
behat/mink-browserkit-driver requires symfony/browser-kit (~2.3)
behat/mink-browserkit-driver requires symfony/dom-crawler (~2.3)
behat/mink-extension requires symfony/config (~2.2)
behat/mink requires symfony/css-selector (~2.1)
behat/symfony2-extension requires (dev) symfony/symfony (~2.1)
behat/symfony2-extension requires symfony/framework-bundle (~2.0)
behat/web-api-extension requires (dev) symfony/process (~2.1)
browscap/browscap-bundle requires (dev) symfony/symfony (>=2.1,<=2.5)
cocur/slugify requires (dev) symfony/dependency-injection (~2.4)
cocur/slugify requires (dev) symfony/http-kernel (~2.4)
doctrine/dbal requires (dev) symfony/console (2.*)
doctrine/doctrine-bundle requires (dev) symfony/validator (~2.2|~3.0)
doctrine/doctrine-bundle requires (dev) symfony/yaml (~2.2|~3.0)
doctrine/doctrine-bundle requires symfony/console (~2.3|~3.0)
doctrine/doctrine-bundle requires symfony/doctrine-bridge (~2.2|~3.0)
doctrine/doctrine-bundle requires symfony/framework-bundle (~2.3|~3.0)
doctrine/doctrine-cache-bundle requires (dev) symfony/console (~2.2|~3.0)
doctrine/doctrine-cache-bundle requires (dev) symfony/finder (~2.2|~3.0)
doctrine/doctrine-cache-bundle requires (dev) symfony/framework-bundle (~2.2|~3.0)
doctrine/doctrine-cache-bundle requires (dev) symfony/validator (~2.2|~3.0)
[...]

PR #4761, opened 2016-01-11
Add Symfony 3 special case on depends command doc

Soullivaneuh:
As requested on https://github.com/composer/composer/pull/4660#issuecomment-170267292.


sroze:
I like the idea :+1:

Seldaek:
Thanks

PR #4775, opened 2016-01-16
Add verification of signatures when running self-update

Seldaek:
Happy to get any feedback on this as it's quite sensitive.

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -334,7 +334,7 @@ EOT
     private function checkPubKeys($config)
     {
         $home = $config->get('home');
-        $errors = [];
+        $errors = array();
         $io = $this->getIO();
 
         if (file_exists($home.'/keys.tags.pub') && file_exists($home.'/keys.dev.pub')) {
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -22,6 +22,7 @@ use Composer\Util\ConfigValidator;
 use Composer\Util\ProcessExecutor;
 use Composer\Util\RemoteFilesystem;
 use Composer\Util\StreamContextFactory;
+use Composer\Util\Keys;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
@@ -132,6 +133,9 @@ EOT
 
         $io->write('Checking disk free space: ', false);
         $this->outputResult($this->checkDiskSpace($config));
+
+        $io->write('Checking pubkeys: ', false);
+        $this->outputResult($this->checkPubKeys($config));
 
         $io->write('Checking composer version: ', false);
         $this->outputResult($this->checkVersion());
@@ -325,6 +329,35 @@ EOT
         }
 
         return true;
+    }
+
+    private function checkPubKeys($config)
+    {
+        $home = $config->get('home');
+        $errors = [];
+        $io = $this->getIO();
+
+        if (file_exists($home.'/keys.tags.pub') && file_exists($home.'/keys.dev.pub')) {
+            $io->write('');
+        }
+
+        if (file_exists($home.'/keys.tags.pub')) {
+            $io->write('Tags Public Key Fingerprint: ' . Keys::fingerprint($home.'/keys.tags.pub'));
+        } else {
+            $errors[] = '<error>Missing pubkey for tags verification</error>';
+        }
+
+        if (file_exists($home.'/keys.dev.pub')) {
+            $io->write('Dev Public Key Fingerprint: ' . Keys::fingerprint($home.'/keys.dev.pub'));
+        } else {
+            $errors[] = '<error>Missing pubkey for dev verification</error>';
+        }
+
+        if ($errors) {
+            $errors[] = '<error>Run composer self-update --update-keys to set them up</error>';
+        }
+
+        return $errors ?: true;
     }
 
     private function checkVersion()
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -14,7 +14,9 @@ namespace Composer\Command;
 
 use Composer\Composer;
 use Composer\Factory;
+use Composer\Config;
 use Composer\Util\Filesystem;
+use Composer\IO\IOInterface;
 use Composer\Util\RemoteFilesystem;
 use Composer\Downloader\FilesystemException;
 use Symfony\Component\Console\Input\InputInterface;
@@ -44,6 +46,7 @@ class SelfUpdateCommand extends Command
                 new InputOption('clean-backups', null, InputOption::VALUE_NONE, 'Delete old backups during an update. This makes the current version of composer the only backup available after the update'),
                 new InputArgument('version', InputArgument::OPTIONAL, 'The version to update to'),
                 new InputOption('no-progress', null, InputOption::VALUE_NONE, 'Do not output download progress.'),
+                new InputOption('update-keys', null, InputOption::VALUE_NONE, 'Prompt user for a key update'),
             ))
             ->setHelp(<<<EOT
 The <info>self-update</info> command checks getcomposer.org for newer
@@ -71,7 +74,12 @@ EOT
 
         $cacheDir = $config->get('cache-dir');
         $rollbackDir = $config->get('data-dir');
+        $home = $config->get('home');
         $localFilename = realpath($_SERVER['argv'][0]) ?: $_SERVER['argv'][0];
+
+        if ($input->getOption('update-keys')) {
+            return $this->fetchKeys($io, $config);
+        }
 
         // check if current dir is writable and if not try the cache dir from settings
         $tmpDir = is_writable(dirname($localFilename)) ? dirname($localFilename) : $cacheDir;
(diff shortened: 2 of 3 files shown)


Seldaek:
cc @stormtide @paragonie-security @padraic

paragonie-scott:
This looks good to me, :+1:

One thing I'd suggest down the line for key management is make an option that will display a [fingerprint of the public key](https://twitter.com/FredericJacobs/status/688116076614307841) to deter and/or detect active attackers.

But for now this is great. :)

Seldaek:
So the way I did this it works with a `openssl genrsa` based pub/priv key pair. I'm not sure how to get a fingerprint of that public key except by generating it myself, which is obviously feasible but not as easy to reproduce then.

Am I missing something?

paragonie-scott:
A fingerprint is a concept, there is no "one true way" of doing that.

```php
function getFingerprint($publicKeyFile)
{
    $hash = hash('sha256', file_get_contents($publicKeyFile));
    return implode(' ', [
        substr($hash, 0, 8),
        substr($hash, 8, 8),
        substr($hash, 16, 8),
        substr($hash, 24, 8),
        '', // Extra space
        substr($hash, 32, 8),
        substr($hash, 40, 8),
        substr($hash, 48, 8),
        substr($hash, 56, 8),
    ]);
}
```

For example: https://3v4l.org/BNGgn

StormTide:
Feedback on this; 

1. Don't really see the need for a fingerprint (the user isnt going to confirm it anyway)

2. I'd suggest using x509 or GPG over direct rsa keys. RSA will work, but x509/gpg has a lot of other important bits like revocation, expiration and certificate status. It'd be a bit more work to implement, but in the long-run would help you when its time to rollover the keys or if someone hacks you. Since it doesnt appear you're using Phar's built-in signature model, its possible to validate the sigs at install time. See the [SecureApt]( https://wiki.debian.org/SecureApt) design architecture?

3. The choice of SHA384 is interesting. SHA256 might make more conformance with industry norms for code signing. (Note, they're totally different algorithms, its not just a key length thing, and have different traits and characteristics) 

4. In thinking about codesigning more generally for Composer, I might select PGP over X509, as it makes it easy to implement a HKP Keyserver for Packagist, and most people who package code are already familiar with GPG tooling for this stuff. HKP's actually a really simple protocol and wouldn't be all that difficult to implement for composer/packagist. Further, using the same tech for code-signing as debian/ubuntu/etc would make it so that if the package signing method were ever to fall, the more relevant vulnerability would be the base system rather than the PHP code on top. I think most folks have faith in the Apt design for this stuff, so why not follow their lead?

5. Some transparency in the key-management practices for Composer itself might be warranted. I'd suggest at a minimum, the codesigning key for composer be stored in some sort of dedicated crypto hardware (think Yubikey or HSM)... I continue to worry that the centralized nature of this distribution model leads to significant risk of persistent-threat-type actors. (Think Belgacom, "We hunt sysadmins", etc) 

6. Glad to see this stuff is finally getting some attention. Look forward to being able to close out 1074 some day ;)

paragonie-scott:
> Don't really see the need for a fingerprint (the user isnt going to confirm it anyway)

It's a value-add. Not essential, but having it available for power users means that attacking composer by replacing the public key risks detection.

> I'd suggest using x509 or GPG over direct rsa keys

That *would* be better, but that's not a feasible goal right now. And in this situation, I'd trust key pinning over PKI.

> The choice of SHA384 is interesting. SHA256 might make more conformance with industry norms for code signing. (Note, they're totally different algorithms, its not just a key length thing, and have different traits and characteristics) 

Including resistance to length-extension attacks that SHA-256 and SHA-512 lack. (This was my recommendation.) The actual industry choice, SHA-512/256, is not available in PHP.

> Some transparency in the key-management practices for Composer itself might be warranted. I'd suggest at a minimum, the codesigning key for composer be stored in some sort of dedicated crypto hardware (think Yubikey or HSM)... 

On that note, I'd totally be willing to donate a Yubikey (shipped directly from Yubico) towards this effort.

Seldaek:
1. I added fingerprints now if only because I think it'll make it easier to debug why someone has issues in case we need to change the keys some day. I added those in the output of `composer diagnose` as well.

2. GPG is just not practical for cross platform development as it can't be relied on on windows and I'm not even sure if it's default on osx for example, we already require openssl for https so using it as well for the signatures makes it easier.

The transparency of key management I am not sure what I could do about it really, nor if disclosing that does more good than harm. BTW there are two keys because one is needed for automated dev releases and the other is only held by me for signing tags, both are password-encrypted too. I'm not against the yubikey but not entirely sure how it'd work with an encrypted private key? Store half the pwd in the key or smth like that?

src/Composer/Command/DiagnoseCommand.php
```diff
@@ -327,6 +331,35 @@ private function checkDiskSpace($config)
         return true;
     }
 
+    private function checkPubKeys($config)
+    {
+        $home = $config->get('home');
+        $errors = [];
```
cs278: `array()`

StormTide:
@Seldaek .. 1. Whatever, cant hurt. 2. The default in the ecosystem is GPG, whether you're doing Git signing or .deb's or whatever. HKP's a basic http-based protocol and easy to setup. I don't think its beyond possibilities to support this model, and GPG4WIN et al, works just fine. 3. Think its about picking industry standards, both git and .deb went with GPG... so rolling your own system here is going to largely be re-inventing the wheel. In the openssl ecosystem, x509 is the tech you're looking for tho (see openssl x509 command)... The big thing missing is a way to rollover the keys, expire compromised ones, or replace it if you lose the key; at a minimum, theres really got to be some sort of CRL/OCSP mechanism so you can call back a broken key.

src/Composer/Command/DiagnoseCommand.php
```diff
@@ -327,6 +331,35 @@ private function checkDiskSpace($config)
...
+        $io = $this->getIO();
+
+        if (file_exists($home.'/keys.tags.pub') && file_exists($home.'/keys.dev.pub')) {
```
chr0n1x: `DIRECTORY_SEPARATOR`? Instead of `/keys.dev.pub`. Or does that not matter on windows anymore?

Seldaek:
@StormTide For signing the phars I feel like using openssl is better because it lets us check signatures for all users that have the openssl extension, and not just those that care enough to have gpg available. If we talk about package signing then gpg definitely sounds like a better option there, and there I think it'd be acceptable to say if you want signing you need gpg.

Regarding x509, I am not sure how this differs from a rsa keypair, nor how that helps me achieve any of the revocation stuff. What I have now seems to work well enough, except indeed we'd need a way to mark keys as being revoked. How about having self-update download a revocation list before checking signatures? The problem is though.. if you are being man-in-the-middle'd to the point that you can't trust what you download over ssl, how can you trust the revoked keys list? I'm not too clear on this.

src/Composer/Command/DiagnoseCommand.php
Seldaek: Thanks, fixed.

src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -106,24 +114,64 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
             $io->writeError('<error>The download of the new composer version failed for an unexpected reason</error>');
 
             return 1;
         }
 
+        // verify phar signature
+        if (!extension_loaded('openssl') && $config->get('disable-tls')) {
+            $io->writeError('<warning>Skipping phar signature verification as you have disabled OpenSSL via config.disable-tls</warning>');
+        } else {
+            if (!extension_loaded('openssl')) {
```
stloyd: This could be done before `if/else` as this is required to pass first anyway.

Seldaek: I don't think so as we need to check for the disable-tls flag, or maybe I don't get what you mean.

stloyd: Check: `!extension_loaded('openssl')` must always pass, so IMO it should be done first:
```php
if (!extension_loaded('openssl')) {
    // throw exception
}

if ($config->get('disable-tls')) {
   // add warning
} else {
   // do the magic (tm) :)
}

Comment (xabbuh) [inline src/Composer/Command/SelfUpdateCommand.php]: @stloyd Not really, if the `disable-tls` option is passed you do not want to throw an exception but only show a warning that the phar signature verification process will be skipped.

Comment (Seldaek) [inline src/Composer/Command/SelfUpdateCommand.php]: Yeah I don't see how I can add the warning once the exception has been thrown.

**Comment by StormTide** [comment]
@Seldaek GPG is pretty much pre-requisite on all modern linux distro's these days (eg its default bundled with ubuntu, debian, redhat, etc).. and its really not any harder to install than openssl on windows... Just given what it brings to the table, its probably worth the dependency.

x509 layers on certificate policy onto private keys. Things like ocsp, crl, are layered in via x509.

**Comment by Seldaek** [comment]
My point is openssl does not need to be installed, as I only use it from within php, and it's just *there*. 

x509.. sorry but you have to be a bit more verbose if you think it's worth convincing me, because as it stands I'll just go forward with what I have I think as it's *something* and most importantly it's *done*. I'm just not familiar enough with all that stuff to judge if it's really worth it or not for the simple use case I have right now.

**Comment by paragonie-scott** [comment]
> GPG is pretty much pre-requisite on all modern linux distro's these days

There is a world outside of GNU/Linux.

> and its really not any harder to install than openssl on windows

It's not included by default, which will cause breakage.

> x509 layers on certificate policy onto private keys. Things like ocsp, crl, are layered in via x509. 

There are some parts of OpenSSL I'd trust. The cipher implementations, the hash functions, and begrudgingly ECDSA. Fine.

Adding a layer of PKI to the mix just because that's what everyone else does seems reckless to me. This is what I'd do instead:

1. Use public-key pinning instead of X.509, which is what's already implemented.
2. Make sure there's a command line flag to disable the check (opt-out of security)
3. In the event of a compromise, tell the world your new public key (including fingerprint) and let people decide for themselves whether or not they trust it.

Don't try to automate trust. Don't add the nightmare complexity of a certificate authority system where it's not needed.

**Comment by StormTide** [comment]
@Seldaek you need openssl installed if you're going to generate signing certs, etc... Unless you're going to re-wrap that from composer (cant see why you'd want to..) ... x509 is what gives you the policy extensions re revocation checking, etc..  You wont know you need it until you run into the problems they've already solved. Just play around with openssl x509 command and look at some certs to see the bits it layers ontop of --BEGIN PRIVATE KEY--... CA's and Trust Chains aren't really needed here... but CRL's and OCSP definitely are.

@paragonie-scott 1. PK pinning means one key, and you cant change it without breaking the installed base (something this project doesnt have the will to do)... 2. Why? Whats the 80/20 use case for enabling an RCE. 3. Tell em how?

**Comment by paragonie-scott** [comment]
> 3. Tell em how?

"Hey guys, we got hacked really hard. Here's a video feed of me apologizing profusely with a new public key and fingerprint."

**Comment by Seldaek** [comment]
@StormTide are you talking about package signing? Because I am only talking about signing the composer.phar at the moment. And for that I am the only one that needs to generate a signing key. Users only need the pubkey and then the verification can be done with php's openssl extension (as in this PR's code) without the need for openssl on CLI.

**Comment by StormTide** [comment]
@Seldaek I'm talking about both really... imo should be same mechanism. Do it once, do it right. Bootstrapping trust from https is essentially just https, so no need to worry about signing composer.phar at the moment. (Can just publish the hash of it for manual verification shasum style)... 

@paragonie-scott So we're registering users on some sort of security-announce mailing list? I've found a lot of these ecosystem failures where something goes off the rails in the crypto setup and getting rid of the vulnerable installed base is always the unsolvable problem. x509 builds in those recall mechanisms into the validation mechanism, so its 100% coverage and automatic.

**Comment by paragonie-scott** [comment]
> x509 builds in those recall mechanisms into the validation mechanism, so its 100% coverage and automatic

In my opinion, key compromise shouldn't he handled automatically. It should be a painful shitstorm that breaks everyone's daily routine.

The second you make it automatic, you've just re-enabled APT style attacks.

**Comment by Seldaek** [comment]
Right now, if the keys get stolen, I'd re-issue keys, and even if you miss any kind of announcement the next time you run composer self-update you'll get a key validation error telling you to check your keys because they're invalid OR the file got hacked in transit. People go check their keys, run `composer self-update --update-keys` to load the new keys.

**Comment by StormTide** [comment]
@Seldaek that process boils out to https security in any event. (update keys i assume being protected by https)

**Comment by Seldaek** [comment]
`--update-keys` tells you to go https://composer.github.io/pubkeys.html and prompts you to paste in the keys in CLI. That page is static and in another repo + diff server than the one serving the phar files (just github hosting really).

**Comment by StormTide** [comment]
@Seldaek Right, so https. See what im saying, the signing key step is superfluous

**Comment by Seldaek** [comment]
Well right now if you gain control over the getcomposer.org server, the signing keys mean you can't ship a new hijacked tagged release to the whole user base as you don't have access to the private key. I think that's kinda valuable.

**Comment by StormTide** [comment]
@Seldaek meh. Do as you please, its not a security vulnerability, so I don't really care, but I dont think I see much value, and its packagist signing that i'm actually worried about. (due to the amplification of the threat surface)

**Comment by Seldaek** [comment]
It's not a vulnerability per se but it allows for a lot more damage if the server gets compromised. Packagist security also is an issue I agree but for that code signing is the only way AFAICT and that's a whole other level of complexity to solve. And to really solve it it'd need *every* package author to be on board which is another issue. It'd be nice but it's a problem for another day I'm afraid :)

**Comment by StormTide** [comment]
@Seldaek fair enough. Can roll it back into 'package signing' issue.

**Comment by paragonie-scott** [comment]
Yeah, agreed. Package signing is a good next step, but will require everyone to have an authentic copy of Composer to be meaningful. :)

**Comment by Seldaek** [comment]
Actually to facilitate the migration (and avoid everyone having to initialize keys manually next time they self-update) I think I should rather embed the pubkeys in the phar for now at least until they are widespread (the installer already sets them up but existing users won't have them). I don't think anything speaks against that? If not then I'll do that change and go ahead with the merge.

**Comment by StormTide** [comment]
@Seldaek should be fine to bootstrap the pubkeys.

**Comment by paragonie-scott** [comment]
No objections here.

File: src/Composer/Command/SelfUpdateCommand.php [hunk 4]
```diff
@@ -106,24 +114,64 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                throw new \RuntimeException('The openssl extension is required for phar signatures to be verified but it is not available. '
+                . 'If you can not enable the openssl extension, you can disable this error, at your own risk, by setting the \'disable-tls\' option to true.');
+            }
+
+            $sigFile = 'file://'.$home.'/' . ($updatingToTag ? 'keys.tags.pub' : 'keys.dev.pub');
+            if (!file_exists($sigFile)) {
+                $io->write('<warning>You are missing the public keys used to verify Composer phar file signatures</warning>');
+                if (!$io->isInteractive() || getenv('CI') || getenv('CONTINUOUS_INTEGRATION')) {
+                    $io->write('<warning>As this process is not interactive or you run on CI, it is allowed to run for now, but you should run "composer self-update --update-keys" to get them set up.</warning>');
```
naderman: Shouldn't it fail in this case?

paragonie-scott: If the keys are bootstrapped, yes. Fail closed.

Seldaek:
Alright keys are now bundled and I indeed removed the CI exception

Issue #4782, opened 2016-01-17
302 Redirect on SSL When Downloading Zip Archive Causes Issues


swimson:
Hey guys,
I saw some commits  related to openssl and TLS.  Dunno if this is related but wanted to pass along.
Declared WordPress as a custom repository of type zip.  This way I can use composer to run wordpress installs (which is awesome!)  This has been working great for about 6 months.  I tried doing a composer update to the newest release

URL:  https://github.com/WordPress/WordPress/archive/4.4.1.zip

This caused an exception I hadn't seen before,

    Composer through a Transport Exception:

    [Composer\Downloader\TransportException]                                                                                                                                
    The "https://github.com/WordPress/WordPress/archive/4.4.1.zip" file could not be downloaded: Peer certificate CN=`*.github.com' did not match expected CN=`github.com'  Failed to enable crypto                                                                                                                                              failed to open stream: operation failed using CN=  

Turns out, GitHub issues a 302 redirect, to a subdomain codeload.github.com.  Looks like there was some SSL error in the stack somewhere because the SSL Common Name was for the wildcard domain (*.github.com) but I was linking to the link provided by GitHub when you want to download a release.

The fix was simple,  I was able to update by changing the source URL in my composer config to the result of the 302 redirect.

https://codeload.github.com/WordPress/WordPress/zip/4.4.1

The fixed composer.json is ....

...

    "repositories": [
    {
      "type": "composer",
      "url": "http://wpackagist.org"
    },
    {
      "type": "package",
      "package": {
        "name": "wordpress",
        "type": "webroot",
        "version": "4.4.1",
        "dist": {
          "type": "zip",
          "url": "https://codeload.github.com/WordPress/WordPress/zip/4.4.1"
        },
        "require": {
          "fancyguy/webroot-installer": "1.0.0"
        }
      }
    }
  ],
...

Keep up the great work guys!

Seldaek:
Ok I see, we probably need to turn the github.com host match into a *.github.com CN_match if that's what they use and rely on. Given github constitutes most of our outbound traffic it probably warrants a special case. (cc @StormTide for quick sanity check here, let me know if you get tired of these..)

Can you confirm if you tried it with php5.5 or below btw? It should work with php5.6, would be nice if you could check.

nelsonko:
[...]

PR #4783, opened 2016-01-17
Follow redirects to properly validate certificates

cs278:
A general fix for #4782 rather than hard coded workaround for GitHub.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -327,16 +327,6 @@ class RemoteFilesystem
             if (!empty($targetUrl)) {
                 $this->redirects++;
 
-                // TODO: Disabled because this is (probably) different behaviour to PHP following for us.
-                // if ('http' === parse_url($targetUrl, PHP_URL_SCHEME) && 'https' === $this->scheme) {
-                //     // Do not allow protocol downgrade.
-                //     // TODO: Currently this will fail if a request goes http -> https -> http
-                //     $e = new TransportException('The "'.$this->fileUrl.'" file could not be downloaded, not permitting protocol downgrade');
-                //     $e->setHeaders($http_response_header);
-                //     $e->setResponse($result);
-                //     throw $e;
-                // }
-
                 if ($this->io->isDebug()) {
                     $this->io->writeError(sprintf('Following redirect (%u)', $this->redirects));
                 }
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -331,14 +331,15 @@ class RemoteFilesystem
                     throw $e;
                 }
 
-                if ('http' === parse_url($targetUrl, PHP_URL_SCHEME) && 'https' === $this->scheme) {
-                    // Do not allow protocol downgrade.
-                    // TODO: Currently this will fail if a request goes http -> https -> http
-                    $e = new TransportException('The "'.$this->fileUrl.'" file could not be downloaded, not permitting protocol downgrade');
-                    $e->setHeaders($http_response_header);
-                    $e->setResponse($result);
-                    throw $e;
-                }
+                // TODO: Disabled because this is (probably) different behaviour to PHP following for us.
+                // if ('http' === parse_url($targetUrl, PHP_URL_SCHEME) && 'https' === $this->scheme) {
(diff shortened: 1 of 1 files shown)


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -293,6 +303,59 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
...
+                    $targetUrl = $locationHeader;
+                } elseif (parse_url($locationHeader, PHP_URL_HOST)) {
+                    // Scheme relative; e.g. //example.com/foo
+                    $targetUrl = $this->scheme.':'.$locationHeader;
+                } elseif ('/' === $locationHeader[0]) {
+                    // Absolute path; e.g. /foo
+                    $urlHost = parse_url($this->fileUrl, PHP_URL_HOST);
+
+                    // Replace path using hostname as an anchor.
+                    $targetUrl = preg_replace('{^(.+(?://|@)'.preg_quote($urlHost).')(?:[/\?].*)?$}', '\1'.$locationHeader, $this->fileUrl);
```
Seldaek: I think you need to allow for an optional port number following the host as well.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -293,6 +303,59 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
...
+
+                // TODO: Disabled because this is (probably) different behaviour to PHP following for us.
+                // if ('http' === parse_url($targetUrl, PHP_URL_SCHEME) && 'https' === $this->scheme) {
+                //     // Do not allow protocol downgrade.
+                //     // TODO: Currently this will fail if a request goes http -> https -> http
+                //     $e = new TransportException('The "'.$this->fileUrl.'" file could not be downloaded, not permitting protocol downgrade');
+                //     $e->setHeaders($http_response_header);
+                //     $e->setResponse($result);
+                //     throw $e;
+                // }
```
Seldaek: Yeah I would just remove this entirely.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -293,6 +303,59 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
...
+
+                if ($this->io->isDebug()) {
+                    $this->io->writeError(sprintf('Following redirect (%u)', $this->redirects));
+                }
+
+                $additionalOptions['redirects'] = $this->redirects;
+
+                // TODO: Not so sure about preserving origin here...
+                return $this->get($this->originUrl, $targetUrl, $additionalOptions, $this->fileName, $this->progress);
```
Seldaek: I think it'd be best to take parse_url of PHP_URL_HOST for the originUrl here.

cs278: Yeah I think it depends how PHP behaves with the redirect, do all the same headers get sent? If so origin should probably stay the same. I'll look into it.

Seldaek: Ok, AFAIK it is wrong to forward the auth headers to the next host, because that'd be an information leak. So I would say we start with re-setting the originUrl to the host every time, and we can always revisit later if needed. It's still a big improvement over the current situation anyway.

cs278: My scepticism was warranted, it appears PHP sends all the user supplied headers when following redirects including the authorization header. Seems the HTTP spec doesn't specify a behaviour so PHP isn't doing anything wrong even if it is quite stupid. libcurl for example has an option to pass auth headers on to different hosts [CURLOPT_UNRESTRICTED_AUTH](http://curl.haxx.se/libcurl/c/CURLOPT_UNRESTRICTED_AUTH.html).

I suppose the question becomes does Composer replicate PHP's behaviour for PHP < 5.6 or handle all redirects to fix the information leak?

Seldaek: Oh well, then I guess if that was the current behavior and all was well let's leave it as is and replicate 5.6+ behavior as closely as we can. I'd rather not keep this code around forever if it can be dropped once we go for 5.6+ only.

Seldaek:
Thanks @cs278! I had to fix a couple things (see above) but overall it seems to work well :)

cs278:
@Seldaek Looks good to me, nice catch with 304. :+1:

PR #4784, opened 2016-01-17
Revert "Avoid loading plugins that do not match the current plugin api even during installation"

josegonzalez:
This reverts commit 21f5933e0170b9673108f05df262dbc7f98cce69.

Commit 21f5933e breaks fresh installations of cakephp applications, or
any project where a dependency uses a different composer plugin
installer than the upstream project.

Closes auraphp/Aura.Intl#20

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -106,26 +106,6 @@ class PluginManager
             return;
         }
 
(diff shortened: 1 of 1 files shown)


bravo-kernel:
:clap:

Seldaek:
Merging while we investigate and figure out the root cause, but this should not have broken like that.. very curious.

PR #4791, opened 2016-01-19
Clarify "path" repository type version requirement

balbuf:
A valid version specifier is required in the `composer.json` file of the package as it cannot be gleaned otherwise (as with a VCS repository). This was not immediately apparent to me.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -639,6 +639,11 @@ file, you can use the following configuration:
 }
 ```
 
+If the package is a local VCS repository, the version may be inferred by
+the branch or tag that is currently checked out. Otherwise, the version should
+be explicitly defined in the package's `composer.json` file. If the version
+cannot be resolved by these means, it is assumed to be `dev-master`.
+
 The local package will be symlinked if possible, in which case the output in
 the console will read `Symlinked from ../../packages/my-package`. If symlinking
 is _not_ possible the package will be copied. In that case, the console will


Seldaek:
Hmm as far as I know it *can* be inferred from VCS, but only if you're on a branch/tag.

balbuf:
Sorry&mdash;poor wording&mdash;the `path` repository type needs a version specified in `composer.json`, even if the requirement is `"*"` (any version). This is _unlike_ local VCS repos, which can infer a version from the branch(es). Perhaps it would warrant a logic change in which specifying `"*"` and matching only one available package, even if it does not specify a version, would satisfy the requirement. I'm not familiar enough with the solver to know if that would make sense, or if there are any implications about not knowing the stability in the absence of a version identifier.

Seldaek:
Sorry but I still don't follow :) Maybe because I never used it much myself, but AFAIK you should be able to require a local path repo package using `*`, it'll just install whatever version is currently checked out in that path?

alcohol:
Yes there is no need to have a `version` attribute in your composer.json. I can confirm this as I have used a path repository locally several times.

balbuf:
Let me try to reproduce what I experienced and get back to you guys.

alcohol:
The path repository uses the `VersionGuesser` which will use attempt to extract the version based on metadata from git/hg/svn. If the path you specified is neither of those, it will simply default to `dev-master` AFAIK.

balbuf:
Ok I was able to reproduce what I was talking about. So first set up a test package:
```sh
$ mkdir /var/tmp/foobar
$ echo '{"name": "foobar/foobar"}' > /var/tmp/foobar/composer.json
```
And then create another repo requiring said package as a path repo:
```
{
    "name": "test/path",
    "repositories": [
        {
            "type": "path",
            "url": "/var/tmp/foobar"
        }
    ],
    "require": {
         "foobar/foobar": "*"
    }
}
```
Then run `composer update` in the dependent repo; it fails with the following error:
```

  Problem 1
    - The requested package foobar/foobar could not be found in any version, there may be a typo in the package name.
```
Now add a version to the `composer.json` of the dependency:
```
$ echo '{"name": "foobar/foobar","version":"1.0.0"}' > /var/tmp/foobar/composer.json
```
Run `composer update` again; it succeeds with the following message:
```
  - Installing foobar/foobar (1.0.0)
    Symlinked from /var/tmp/foobar/

Writing lock file
Generating autoload files
```

balbuf:
See, I too expected the behavior that you described: if no version could be gleaned from VCS information, and no `version` property is set, the version would default to `dev-master`. So perhaps it would be more appropriate to introduce a change that alters the behavior to work on that expectation, instead of the documentation change in this PR. Or, am I doing something wrong?

Seldaek:
Ok that makes sense as /var/tmp/foobar isn't a git repo it can not infer a version from it. Maybe you can tweak the wording in your doc PR.

IMO if there is no version and none can be inferred we shouldn't go about guessing.. but maybe that's just me.

For the root package we *assume* 1.0.0 if we can't guess a version, but that has usually very little consequences on the dependency resolutions and forcing people to have a version there would be quite painful. For path repo though I believe it makes more sense to say if you want to use it define a version? I don't really see the path repo use case without a vcs repo I have to say.

balbuf:
Yeah I'm not sure how many people use that `path` repo type -- I just elected to use it over VCS for development purposes. I'm working on a plugin, and as I figure out how composer works under the hood, I don't want to commit every single change to test the plugin in the dependent repo. I think the point still stands about clarifying the docs for `path` repo, but would you have a better recommendation for testing a composer plugin while actively developing it?

Seldaek:
Well I think it's ok to do it with a path repo, but using a path repo 
doesn't mean the plugin repo has to be out of VCS. You can point a path 
repo at a git directory just fine, then you have a version so composer 
is happy, but you dont have to commit to update as it'll symlink it.

alcohol:
I still don't quite understand which part of the code is complaining. I guess it is the solver? Cause the path repo implementation itself should have no problem with the above scenario. I'll give your demo case a try later today maybe.

alcohol:
Wait, what if you lower the minimum stability or explicitly set your require to dev-master?

Seldaek:
Ah yeah that's true the path repo defaults to dev-master if no version 
can be guessed, I hadn't seen that. Weird indeed then.

balbuf:
Oh that makes sense! I didn't consider that a VCS repo would be symlinked just the same. Good to know! Now, I actually did a bit more exploring and here is what I found:

Turns out what @alcohol was saying is indeed correct: if no version is inferred, it defaults to `dev-master`. However, since the default for `minimum-stability` of the root repo is `stable`, a version of `dev-master` would obviously not satisfy that requirement.

balbuf:
So I think it is still worth pointing out (maybe in the composer.json schema section instead) that the default version, if not inferred from VCS or from the version property itself, will be considered `dev-master`.

balbuf:
What do you guys think? Maybe also a note in the troubleshooting would be helpful too? It's just that I was racking my brain trying to figure out why I was getting that error and couldn't really find any mention of it from googling, so I just wanted to clear that up for folks in the future.

Seldaek:
I think having some info about that in the path repo docs makes sense, I wouldn't put info related to that all over the place as it's quite a specific use case.

Basically if you can update your sentence to clarify things a bit more that it's only needed if you have no git and you don't want to use dev-master then you should define a version yourself?

balbuf:
Updated the note - let me know how that looks. :)

Seldaek:
Great, thanks!

PR #4793, opened 2016-01-19
make root server file url configurable

xjchengo:
Connection to `packagist.org` from my network is very poor but to `api.github.com` is good. So I want to put a CDN before packagist.org  and set packagist repo like this:
```
{
    "repositories": {
        "packagist": {
            "type": "composer",
            "url": "CDN_URL"
        }
    }
}
```
CDN is set to be a mirror of `packagist.org`. It works! After package provider files is cached in CDN, the speed of `composer install & update` is acceptable.

Because there isn't version-related info in root server file url, my CDN cannot update it immediately.

In this PR, I make root server file url configurable. So I can put root server file in a dynamic content server and other package provider files in CDN.

Now the CDN mirror is updated passively. Next step I will make it updated actively.
Any suggestion is welcome.

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -38,6 +38,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
     protected $config;
     protected $repoConfig;
     protected $options;
+    protected $rootServerFileUrl;
     protected $url;
     protected $baseUrl;
     protected $io;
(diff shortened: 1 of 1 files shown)


Seldaek:
What about doing it the other way? Point the repo url to the dynamic server where the main file is hosted, and then make that main file point all the providers-url/providers-includes to the CDN URL? I think there is nothing forcing these URLs to be domain-relative, I think they can point to a completely different server without problem, or if not then we could fix this.

PR #4798, opened 2016-01-20
Provide support for subjectAltName on PHP < 5.6

cs278:
This is an experimental (read crazy) PR to provide support for certificates using the [subjectAltName](https://en.wikipedia.org/wiki/SubjectAltName) (sAN) extension when Composer is being run on PHP < 5.6.

When running PHP < 5.6 and a CN error is thrown the following steps kick in:

1. Open socket to remote server and fetch certificate (disabling peer validation because PHP errors without giving up the certificate.)
2. Verifying the domain in the URL against the names in the sAN field. If there is a match record the authority [host/port], certificate common name, and certificate fingerprint.
3. Retry the original request but changing the CN_match parameter to the common name extracted from the certificate in step 2.
4. To prevent any attempt at being hoodwinked by switching the certificate between steps 2 and 3 the fingerprint of the certificate presented in step 3 is compared against the one recorded in step 2.

Basically I've aimed to move the sAN check into userland whilst delegating everything else back to PHP.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -33,6 +33,7 @@ class RemoteFilesystem
     private $progress;
     private $lastProgress;
     private $options = array();
+    private $peerCertificateMap = array();
     private $disableTls = false;
     private $retryAuthFailure;
     private $lastHeaders;
@@ -252,6 +253,18 @@ class RemoteFilesystem
         });
         try {
             $result = file_get_contents($fileUrl, false, $ctx);
+
+            if (PHP_VERSION_ID < 50600 && !empty($options['ssl']['peer_fingerprint'])) {
+                // Emulate fingerprint validation on PHP < 5.6
+                $params = stream_context_get_params($ctx);
+                $expectedPeerFingerprint = $options['ssl']['peer_fingerprint'];
+                $peerFingerprint = $this->getCertificateFingerprint($params['options']['ssl']['peer_certificate']);
+
+                // Constant time compare??!
+                if ($expectedPeerFingerprint !== $peerFingerprint) {
+                    throw new TransportException('Peer fingerprint did not match');
+                }
+            }
         } catch (\Exception $e) {
             if ($e instanceof TransportException && !empty($http_response_header[0])) {
                 $e->setHeaders($http_response_header);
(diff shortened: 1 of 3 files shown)


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -775,4 +823,113 @@ private function validateCaFile($filename)
...
+        // Close non authenticated connection without reading any content.
+        fclose($handle);
+
+        $params = stream_context_get_params($context);
+
+        if (!empty($params['options']['ssl']['peer_certificate'])) {
+            $peerCertificate = $params['options']['ssl']['peer_certificate'];
+
+            $fp = $this->getCertificateFingerprint($peerCertificate);
+            $cert = openssl_x509_parse($peerCertificate, false);
```
Seldaek: If you look at validateCaFile we avoid calling that as there is a vulnerability there. I'm not sure what's worse to be vulnerable to a bad cert or to be vulnerable to not checking certs.. Feels like the situation is pretty bleak for php <5.6 :/

cs278: RCE so bad and given there is a chance people foolishly run Composer as root this could be very bad.

I guess for older versions I could dump the certificate into a file and attempt to parse it by calling out to openssl. Another potential workaround is to run the test for that issue in an isolated PHP process and see if it fails, one of the problems with distros backporting bug fixes but not leaving any way of verifying if things have been fixed. https://github.com/php/php-src/commit/c1224573c773b6845e83505f717fbf820fc18415

Seldaek: Yeah the distro thing is a real issue. The easiest would be to assume the fix was applied and just call it.. If people run unpatched php they probably have other issues. It's not the best but it would at least keep well-behaving users safe. 

Calling into external processes to figure this out sounds really nasty but if you like to try and do it why not I guess. But please make sure it is cached and fails gracefully :)

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -775,4 +823,113 @@ private function validateCaFile($filename)
 
         return (bool) openssl_x509_parse($contents);
     }
+
+    private function getCertificateCnAndFp($url, $options)
```
Seldaek: Could you mark all those new methods with a big sign that says they're php<5.6 specific and should be deleted whenever possible?

src/Composer/Util/RemoteFilesystem.php
cs278: Yup it's going to be nasty, but I think it'll be worth while as the friction from sAN being unsupported is only going to get people to turn off TLS.

src/Composer/Util/RemoteFilesystem.php
cs278: Yeah, I can make them fail hard on PHP >= 5.6 just in case.

cs278:
@Seldaek Also I started prototyping functional testing of TLS support in the RFS: https://github.com/composer/composer/compare/master...cs278:test-rfs

src/Composer/Util/RemoteFilesystem.php
jakoch: Just an idea to ease maintenance: 
Would a shim or compat layer work out (to keep the compatibility code in a separate file and the original class clean form BC code)? 

It seems that the RemoteFilesystem object is instantiated via `Composer\Factory::createRemoteFilesystem()`.
That would allow to use this factory method to instantiate and deliver a class matching the PHP version constraint, e.g. `RemoteFilesystemPHP56Shim`.

Seldaek: That would be ideal but not sure if feasible, it'd need to catch 
exceptions and do smart things.. but it *might* work yes, and definitely 
would help keep that class sane (it already is on the edge..). I guess 
the redirection handling in the other PR also could be extracted into 
this layer.

cs278: I'd like to pull a lot of the Composer workarounds into a separate package so they can be used outside of Composer, that's one of the reason I started looking at functional testing.

cs278:
@Seldaek Updated.

Seldaek:
@cs278 Thanks! Quite the insane work you did there with isOpensslParseSafe :) Tested and it seems to work well. Hopefully that's the end of the ssl issues :s

cs278:
Thanks @Seldaek, I was expecting to be fixing a pile of comments this evening.

PR #4799, opened 2016-01-20
createPackage() method should use the $class argument it accepts

balbuf:
`createPackage()` accepts an argument for `$class` but doesn't actually use it. I added a default value of `'Composer\Package\CompletePackage'` which seems to be the preferred class to use, at least within `ComposerRepository`. Additionally, `'Composer\Package\Package'` was swapped for `'Composer\Package\CompletePackage'` on #L373, as this is the effective value that was being used having ignored the argument.


alcohol:
LGTM

Seldaek:
Thanks

PR #4805, opened 2016-01-20
Add capath configuration capability and refactor cafile resolving

### LINKED ISSUE #4804: TLS verification does not allow private CAs
I think I have a bug case here: We are using HTTPS for internal servers that host ZIP packages (the usual Satis installation case), the certificates are signed by an internal CA. The CA certificate is correctly placed in /etc/ssl/certs as a PEM file (according to the infrastructure admin).

However, `composer install` fails with a verification error:
```
  The "https://myUrl/package.tgz" file could not be downloaded: SSL operation failed with code 1. OpenSSL Error messages:  
  error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed                                                                                                                                                   
  Failed to enable crypto                                                                                                                                                                                                             
  failed to open stream: operation failed
```

On my way to investigate, I also issued `composer diag` with a very strange result:
```
 [ErrorException]
  fopen(/tmp/composer-cacert.pem): failed to open stream: Permission denied
```

Looking at the code, I realize that this case is triggered by the automatic CA search - the result of this search is the last entry in the array of possible file locations: The internal cacert.pem from composer.phar. However, this is the problem: I need to have Composer find the DIRECTORY that contains all the individual PEM files, but this case is never reached in the code because that last entry will always match.

I think you should reorder the evaluation of CA finding a bit to only come up with the Composer internal file if all other solutions fail, just before returning false.

alcohol:
Adds capath configuration capability.
Minor refactor of how cafile is resolved also (when not specified).

Closes #4804 #4812

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -55,9 +55,15 @@ php_openssl extension in php.ini.
 
 ## cafile
 
-A way to set the path to the openssl CA file. In PHP 5.6+ you should rather
-set this via openssl.cafile in php.ini, although PHP 5.6+ should be able to
-detect your system CA file automatically.
+Location of Certificate Authority file on local filesystem. In PHP 5.6+ you
+should rather set this via openssl.cafile in php.ini, although PHP 5.6+ should
+be able to detect your system CA file automatically.
+
+## capath
+
+If cafile is not specified or if the certificate is not found there, the
+directory pointed to by capath is searched for a suitable certificate.
+capath must be a correctly hashed certificate directory.
 
 ## http-basic
 
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -590,16 +590,10 @@ class Factory
         $remoteFilesystemOptions = array();
         if ($disableTls === false) {
             if ($config && $config->get('cafile')) {
-                $remoteFilesystemOptions = array_merge_recursive(
-                    $remoteFilesystemOptions,
-                    array('ssl' => array('cafile' => $config->get('cafile')))
-                );
+                $remoteFilesystemOptions['ssl']['cafile'] = $config->get('cafile');
             }
             if ($config && $config->get('capath')) {
-                $remoteFilesystemOptions = array_merge_recursive(
-                    $remoteFilesystemOptions,
-                    array('ssl' => array('capath' => $config->get('capath')))
-                );
+                $remoteFilesystemOptions['ssl']['capath'] = $config->get('capath');
             }
         }
         try {
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -595,6 +595,7 @@ class Factory
             if ($config && $config->get('capath')) {
                 $remoteFilesystemOptions['ssl']['capath'] = $config->get('capath');
             }
+            $remoteFilesystemOptions = array_replace_recursive($remoteFilesystemOptions, $options);
         }
         try {
             $remoteFilesystem = new RemoteFilesystem($io, $config, $remoteFilesystemOptions, $disableTls);
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -590,9 +590,17 @@ class Factory
         $remoteFilesystemOptions = array();
         if ($disableTls === false) {
             if ($config && $config->get('cafile')) {
-                $remoteFilesystemOptions = array('ssl' => array('cafile' => $config->get('cafile')));
+                $remoteFilesystemOptions = array_merge_recursive(
+                    $remoteFilesystemOptions,
+                    array('ssl' => array('cafile' => $config->get('cafile')))
+                );
             }
-            $remoteFilesystemOptions = array_merge_recursive($remoteFilesystemOptions, $options);
+            if ($config && $config->get('capath')) {
+                $remoteFilesystemOptions = array_merge_recursive(
+                    $remoteFilesystemOptions,
+                    array('ssl' => array('capath' => $config->get('capath')))
+                );
+            }
         }
         try {
             $remoteFilesystem = new RemoteFilesystem($io, $config, $remoteFilesystemOptions, $disableTls);
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -642,12 +642,12 @@ class RemoteFilesystem
                     fclose($target);
                     unset($source, $target);
 
-                    $options['ssl']['cafile'] = $targetPath;
+                    $defaults['ssl']['cafile'] = $targetPath;
                 } else {
                     if (is_dir($result)) {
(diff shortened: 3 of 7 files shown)


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -631,16 +623,26 @@ private function getTlsDefaults()
...
+        if (isset($options['ssl'])) {
+            $defaults['ssl'] = array_merge_recursive($defaults['ssl'], $options['ssl']);
+        }
+
         /**
          * Attempt to find a local cafile or throw an exception if none pre-set
          * The user may go download one if this occurs.
          */
-        if (!isset($this->options['ssl']['cafile'])) {
+        if (!isset($defaults['ssl']['cafile'], $defaults['ssl']['capath'])) {
```
Seldaek: I kinda hate isset() with multiple args because I'm never sure if people messed up or not.. It's so confusing.

> If multiple parameters are supplied then isset() will return TRUE only if all of the parameters are set.

So this is entered unless you pass both a cafile and a capath if I'm not mistaken, which sounds like.. always? Is that what you meant?

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -631,16 +623,26 @@ private function getTlsDefaults()
...
             $result = $this->getSystemCaRootBundlePath();
-            if ($result) {
-                if (preg_match('{^phar://}', $result)) {
-                    $targetPath = rtrim(sys_get_temp_dir(), '\\/') . '/composer-cacert.pem';
 
+            if (!$result) {
+                throw new TransportException('A valid cafile or capath could not be located automatically.');
+            }
```
Seldaek: This case isn't possible anymore I believe since you return the composer one as fallback always

Seldaek: Sorry I deleted this but after all I meant it, I don't think this is needed as getSystemCaRootBundlePath always returns something

doc/06-config.md
```diff
@@ -55,9 +55,15 @@ php_openssl extension in php.ini.
...
-detect your system CA file automatically.
+Location of Certificate Authority file on local filesystem. In PHP 5.6+ you
+should rather set this via openssl.cafile in php.ini, although PHP 5.6+ should
+be able to detect your system CA file automatically.
+
+## capath
+
+If cafile is not specified or if the certificate is not found there, the
+directory pointed to by capath is searched for a suitable certificate.
+capath must be a correctly hashed certificate directory.
```
Seldaek: capath is still missing in composer-schema.json

src/Composer/Factory.php
```diff
@@ -590,9 +590,17 @@ public static function createRemoteFilesystem(IOInterface $io, Config $config =
...
         if ($disableTls === false) {
             if ($config && $config->get('cafile')) {
-                $remoteFilesystemOptions = array('ssl' => array('cafile' => $config->get('cafile')));
+                $remoteFilesystemOptions = array_merge_recursive(
+                    $remoteFilesystemOptions,
+                    array('ssl' => array('cafile' => $config->get('cafile')))
+                );
+            }
+            if ($config && $config->get('capath')) {
+                $remoteFilesystemOptions = array_merge_recursive(
```
Seldaek: You want to use array_replace_recursive here and at line 595 I think, because merge does real weird things.

Seldaek: That said, I'd just simplify this to `$remoteFilesystemOptions['ssl']['capath'] = $config->get('capath');` which is a lot clearer and works fine.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -631,16 +623,26 @@ private function getTlsDefaults()
...
+
+            if (preg_match('{^phar://}', $result)) {
+                $hash = md5(file_get_contents($result));
```
cs278: I think it'd be best to change this to `sha256` and check if the hash is correct, otherwise Mallory can place a certificate file in your temp directory forcing you to trust their CA. You can probably use `hash_file()`  as well.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -648,29 +650,34 @@ private function getTlsDefaults()
                     stream_copy_to_stream($source, $target);
                     fclose($source);
                     fclose($target);
+                    chmod($targetPath, 0744);
```
cs278: `0644`

src/Composer/Util/RemoteFilesystem.php
alcohol: No. I actually confused myself there too. Because I want the inverse of `isset()`.

alcohol:
I think that addresses most/all concerns. Please let me know your opinion on the file permissions. Since it's in temp I felt that `0666` should be ok (also because we overwrite it if the content hash does not match expected hash). But on the other hand we might just want to fail with a fatal error if permissions are not correct. Not sure.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -631,80 +628,86 @@ private function getTlsDefaults()
             )
         );
 
+        if (isset($options['ssl'])) {
+            $defaults['ssl'] = array_merge_recursive($defaults['ssl'], $options['ssl']);
```
Seldaek: Should also be array_replace_recursive here I believe.

src/Composer/Factory.php
```diff
@@ -590,9 +590,11 @@ public static function createRemoteFilesystem(IOInterface $io, Config $config =
         $remoteFilesystemOptions = array();
         if ($disableTls === false) {
             if ($config && $config->get('cafile')) {
-                $remoteFilesystemOptions = array('ssl' => array('cafile' => $config->get('cafile')));
+                $remoteFilesystemOptions['ssl']['cafile'] = $config->get('cafile');
+            }
+            if ($config && $config->get('capath')) {
+                $remoteFilesystemOptions['ssl']['capath'] = $config->get('capath');
             }
-            $remoteFilesystemOptions = array_merge_recursive($remoteFilesystemOptions, $options);
```
Seldaek: thi sis still needed otherwise the options array is dropped, but it should use array_replace_recursive I think (that was probably a mistake).

src/Composer/Util/RemoteFilesystem.php
alcohol: Ah my bad

cs278:
Not sure 0666 is a good idea, even if the file is being validated it's not used immediately and could be tampered with. Perhaps storing it in the Composer cache-dir is a better/safer idea?

alcohol:
I like the composer cache-dir suggestion. Not sure why it wasn't put there in the first place, but I imagined there was a reason for that. That would enable use to use more strict permissions also. Maybe open an issue regarding this?

PR #4811, opened 2016-01-21
try to use unique test directories

alcohol:
Any tests that use the filesystem should have their own unique directory, as we run our test suite in parallel and cleanup of tests (removing directories) should not interfere with currently running tests.

Right now, this PR breaks testing locally on OSX due to the incredibly convoluted way the temporary directory is setup there. Need to find out why it breaks so many of our tests, cause I need to be able to run the test suite locally (have a hunch that it might be related to not enough use of `realpath()`).

### DIFF
--- a/tests/Composer/TestCase.php
+++ b/tests/Composer/TestCase.php
@@ -64,7 +64,7 @@ abstract class TestCase extends \PHPUnit_Framework_TestCase
         do {
             $unique = $root . DIRECTORY_SEPARATOR . uniqid('composer-test-');
             if (!file_exists($unique) && mkdir($unique, 0777)) {
-                return $unique;
+                return realpath($unique);
             }
         } while (--$attempts);
 
--- a/tests/Composer/TestCase.php
+++ b/tests/Composer/TestCase.php
@@ -56,12 +56,29 @@ abstract class TestCase extends \PHPUnit_Framework_TestCase
         return new AliasPackage($package, $normVersion, $version);
     }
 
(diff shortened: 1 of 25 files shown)


tests/Composer/TestCase.php
```diff
@@ -56,12 +56,27 @@ protected function getAliasPackage($package, $version)
...
-    protected function ensureDirectoryExistsAndClear($directory)
+    protected static function getUniqueTmpDirectory()
+    {
+        $root = sys_get_temp_dir();
+
+        do {
+            $path = $root . DIRECTORY_SEPARATOR . uniqid('composer-test-');
+        } while (file_exists($path));
+
+        static::ensureDirectoryExistsAndClear($path);
```
cs278: Making sure the directory is empty isn't needed if it's unique and if it's not unique it doesn't solve the clashes.

I'd use the following which avoids any race conditions at the expense of error suppression and ensures there isn't a runaway loop.

```php
$attempts = 5;
$root = sys_get_temp_dir();

do {
    $path = $root . DIRECTORY_SEPARATOR . uniqid('composer-test-');

    // umask is applied by PHP
    if (false !== @mkdir($path, 0700)) {
        return $path;
    }
} while (0 < --$attempts);

throw new \RuntimeException('Failed to create temporary directory');
```

alcohol: Fair enough. I simply re-used the existing method for creating a directory cause it does pretty much the same thing when the desired directory does not exist yet. But I can adjust this, no problem.

Seldaek:
Yay

PR #4813, opened 2016-01-21
Sometimes redirects behind proxys fails

Deltachaos:
Implementing a workaround if a redirect fails but works if you request it directly.
Some company proxies show this behavior, this is the reason I have implemented this workaround.


Seldaek:
Can you try and see if it works with latest composer version? We handle redirects manually now on php <5.6, 5.6+ is still handled at the php level but I assume if it fails for some reason we could use the fallback like we do on <5.6. 

I really think those proxies should be fixed though as we can only do so many workarounds.

PR #4814, opened 2016-01-22
Fix typo in repositories doc

svenluijten:
The folder is called "my-package" in the example above, but in the implementation it says "my/package". Fixed by replacing the `/` with a `-`.


Seldaek:
Thanks but my/package is the package name which has no relation to the path and as we kinda enforce the foo/bar convention for package names (at least on packagist) I think it's better to leave it as is actually.

PR #4815, opened 2016-01-22
[Config command] allow to pass options when adding a repo

bamarni:
It could help to allow an extra argument to the config command for repos, in my case it's to automate the creation of a satis repo, what do you think?

Related to #2920, also another alternative to #2604.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -466,6 +466,12 @@ changes to the repositories section by using it the following way:
 php composer.phar config repositories.foo vcs https://github.com/foo/bar
 ```
 
+If your repository requires more configuration options, you can instead pass its JSON representation :
+
+```sh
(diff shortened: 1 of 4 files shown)


bamarni:
@Seldaek : not in a hurry but just wondering if it's something you'd consider to merge? Otherwise I'll try to find another approach to solve this :)

Seldaek:
Given that some things like SvnDriver/VCS repo take options at the top level, or the package repo takes a `package` key as well, I don't think it's mergeable as is. Maybe if you update this to work with the `1` arg case that if it's a string passed we take it as the full repo block? That'd be more flexible.

bamarni:
Makes sense definitely, I've updated it. There is a weird failure in appveyor somehow.. Looks random.

Seldaek:
Cool thanks, don't worry about appveyor it fails fairly randomly :)

PR #4817, opened 2016-01-22
Bug #4795

### LINKED ISSUE #4795: "composer update vendor/package --with-dependencies" does not update root packages
Hi there,

I recently discovered that the --with-dependencies option for the update command does not seem to work properly.

The abstract scenario is that I have 3 packages installed:

    A[1.0.0]
    B[1.0.0] requires A[~1.0]
    C[1.0.0] requires A[~1.0]

in my composer.json I just have required B and C (not A):

    "require": {
        "B": "~1.0",
        "C": "~1.0"
    }

Now there is a new package B[1.1.0] which requires A[~1.1].

In my Application I just want to update package B so I executed:

    php composer.phar update B --with-dependencies

Output:

    Nothing to install or update

Expected Result:

As the dependencies from B should be updated as well, A and B should be updated, because A still matches the dependency ~1.0 from C

Weird behaviour:

If I run  ```php composer.phar update A B``` it simply updates both packages.

Even more weird:

if I update my composer json to:

     "require": {
        "B": "~1.1",
        "C": "~1.0"
    }

And after that I run 

    php composer.phar update B --with-dependencies

The output changes to:

    Problem 1
        - Can only install one of: A[1.1.0, 1.0.0].
        - Can only install one of: A[1.1.0, 1.0.0].
        - Can only install one of: A[1.0.0, 1.1.0].
        - B 1.1.0 requires A ~1.1.0 -> satisfiable by A[1.1.0].
        - Installation request for B ~1.1 -> satisfiable by B[1.1.0].
        - Installation request for A == 1.1.0.0 -> satisfiable by A[1.1.0].

Is this a bug or an indented behaviour?

*update*
Unlike said, this issue only appears if you require A[~1.0] in the root composer.json

alcohol:
Once resolved, should close #4795


alcohol:
It seems that `Composer\Installer::whitelistUpdateDependencies` intentionally ignores root requirements even if said root package is also a dependency of one the requirements that is whitelisted for update (an odd scenario I will admit, but given the fact that it was reported as a bug, not entirely unimaginable it seems).

/cc @naderman I heard from @Seldaek that he is under the impression that you wrote this. It seems like an inconvenience to me? Do you know if there is any particular reasoning behind this implementation?

naderman:
Yes I am, and it was done this way intentionally. The idea is that you want versions listed in your composer.json to be in specific versions unless you explicitly update them. `--with-dependencies` should only update packages which are not explicitly listed as dependencies themselves.

I'd be open to instead just outputting a warning that those packages are being skipped because they are listed in composer.json, and that they should be added as arguments to the update call if they should be updated, too.

alcohol:
> I'd be open to instead just outputting a warning that those packages are being skipped because they are listed in composer.json, and that they should be added as arguments to the update call if they should be updated, too.

That sounds like a very reasonable solution to me. /cc @Seldaek agreed as well?

Seldaek:
Yup for sure, it's definitely not a major pain point as it was only 
reported now, but it is indeed a waste of time for everyone if it's 
unclear why something does not update.

teohhanhui:
Even when the package is explicitly whitelisted for update, it's still not being updated (at least the warning still shows up).

```json
"require": {
    ...
    "symfony/symfony": "^3.0.3"
    ...
}
```

```shell
composer update symfony/symfony ... --with-dependencies -vvv
```

> Dependency "symfony/symfony" is also a root requirement, but is not explicitly whitelisted. Ignoring.

alcohol:
Are you using the latest version of Composer?

teohhanhui:
Composer version 1.0-dev (3933a7f404cfa197c5457f49b0665eae40ddedd9) 2016-03-07 01:50:41

alcohol:
Can you show your exact command and the output (`-vvv` please)?

teohhanhui:
https://gist.github.com/teohhanhui/ae159ce54fa8b4fe27f7

alcohol:
Huh, that is peculiar. I'll try to see if I can actually reproduce it first with a test scenario.

alcohol:
Could you share your composer.json btw? And possibly also the output of `composer show -i`?

teohhanhui:
All the `$rootRequires` are added to `$skipPackages` here: https://github.com/composer/composer/blob/8344c6d3d7c5036e6f9c475b0bd102f41134c35f/src/Composer/Installer.php#L1121-L1124

and then `$skipPackages` is checked here:
https://github.com/composer/composer/blob/8344c6d3d7c5036e6f9c475b0bd102f41134c35f/src/Composer/Installer.php#L1178-L1181

I don't see how we account for whitelisted packages in `$rootRequires`?

alcohol:
The fact that I can't remember how this worked just screams for a refactoring. Ugh this code. Let me dig a bit and get back to you ok?

alcohol:
I'd still like you to share those 2 files I asked for though, if that is possible.

teohhanhui:
composer.json - https://gist.github.com/teohhanhui/5a44085fedb9cc75dc97

teohhanhui:
composer show -i
https://gist.github.com/teohhanhui/bfa4fa8f5b6f91901b8e

alcohol:
Hrm.. I see nothing out of the ordinary :-(

Can you share your lock file also? I'll try reproducing the scenario locally with a debug version of Composer.

teohhanhui:
My lock file has already been changed. Let me try to recreate another one... :)

alcohol:
You don't have the previous version stored in git by chance?

teohhanhui:
composer.lock from before the `composer update` (not sure if this is right before that update, but you should be able to reproduce the warning easily...)

https://gist.github.com/teohhanhui/41a270e0877c5ca229e9

alcohol:
I managed to reproduce it using a test fixture actually. It seems it always outputs the warning, even if it does whitelist it for updating. It still attempts to update it though. So it seems in your case there simply was no new version available. But the warning should not be displayed in this scenario, that is a bug indeed.

Issue #4203, opened 2015-07-01
Using error suppression operator with @date_default_timezone_get() - hides rare, but fatal php error with no further output


yurakaras:
### No output while running the `php composer.phar`, with dozens of various settings, cause of using the @ operator with @date_default_timezone_get()

I had no success to get some output from composer.phar, on debian linux with php
5.5.21-1+deb.sury.org~precise+2

 
`curl -sS https://getcomposer.org/installer | php -- --check` 
output was "All settings correct for using Composer" 

*Contrary - i tested few phar files from popular CMS/CMF - they'd worked fine - with output and executing commands.*

While i can't at the moment update php version, and still addicted (in goodway) to Composer, so i stucked with old shared hosting. Recently i made more deep reversing of composer.phar on that debian machine, extracting the composer.phar and ran 'bin/composer' "by hand".

Appeared that i always had **php fatal error, which Composer was always hiding from me**.

*I know it's incorrect setup of php by server admin.*

**Fatal error: date_default_timezone_get(): Timezone database is corrupt - this sh
ould *never* happen! in /home/karas/tmp2/c/src/Composer/Console/Application.php
on line 66**

**Here is the code src/Composer/Console/Application.php**

```
  public function __construct()
    {
        if (function_exists('ini_set') && extension_loaded('xdebug')) {
            ini_set('xdebug.show_exception_trace', false);
            ini_set('xdebug.scream', false);
        }
        if (function_exists('date_default_timezone_set') && function_exists('date_default_timezone_get')) {
            date_default_timezone_set(@date_default_timezone_get());
        }
        parent::__construct('Composer', Composer::VERSION);
    }
```

**If someone with php timezone corrupted/wrong setup, he would no simple hint, just silence**

Proposal - remove the silence operator, and either stop on Fatal Error, or if the version of PHP is new, process it through some error handler - and give user the output.

p.s.

I know this is rare issue, with more then 5 years of PHP i'd never that this fatal error exists.

Thanks

alcohol:
I'm not sure what a decent alternative would be though. If you still have access to that environment, can you see if wrapping the `date_default_timezone_get()` call in a try/catch statement provides better results?

stof:
the issue is that without the silencing operator, any setup which does not define the timezone in the php.ini would get a warning here (and the default php.ini file does not define it)

yurakaras:
@alcohol Well i've tried few options - 
```error_get_last()```
Nothing

```ini_set('track_errors',1);
[...]

PR #4820, opened 2016-01-23
Integrated at-ease library to more gracefully handle warning suppression

curry684:
See #4203

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
         "symfony/process": "~2.1",
         "symfony/filesystem": "~2.5",
         "seld/phar-utils": "~1.0",
-        "seld/cli-prompt": "~1.0"
+        "seld/cli-prompt": "~1.0",
+        "mediawiki/at-ease": "^1.1"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.5|^5.0.5",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "af3956ae4c1a09e3e72cd66346a6176d",
-    "content-hash": "96817117d0ca449e7deff55c98eb7bdc",
+    "hash": "1b012c381e76f464eed9aff04d8b2b2c",
+    "content-hash": "fe9e5f4dff3c95a5e02a89ad0b627d3b",
     "packages": [
         {
             "name": "composer/semver",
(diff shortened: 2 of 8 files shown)


curry684:
The conflict is likely in composer.lock, wasn't sure how to handle that after changing composer.json

Seldaek:
AFAIK the date_ call is the only one that has the potential fatal error, so I'd rather not add another dependency just for this. The idea of [changing the error_reporting](https://github.com/composer/composer/issues/4203#issuecomment-122562635) sounds less intrusive than that and would fix the `@` issue. date_default_timezone_get() triggers a warning, so if we set the error_reporting to `E_ALL & ~E_STRICT & ~E_WARNING` and then reset it after the call I think we are fine. Would you like to do this?

curry684:
The dependency is a single file actually, for which the contents are trivial to duplicate inside Composer. I don't entirely agree that this specific call is the only one that needs to be fixed - I also fixed many of the other ones exactly because they could have similar issues, for example when a file was removed in a race condition. I think as a rule the `@` operator should be removed entirely in the end, as it always has the same issue - it hides fatal errors when its purpose is only to accept warnings silently. It should be noted that it also has severe performance issues.

Another problem I just noted is that at-ease is released under GPL2.0+ for those few lines of code, which can cause issues with Composer's MIT.

I will just replicate the functionality natively inside Composer and fix it that way.

Seldaek:
I mean that the other calls you changed have no way to cause a fatal error AFAIK, so I don't really see the benefit of changing them if the `@` works fine.

As for the performance issue, I don't think it is *severe* unless you use it in a tight loop, but also the overhead it has is due to changing the error_reporting + the fact that it tends to make you generate more notices than you should (and generating notices costs cpu time too). All this is still true with at-ease as it still changes the error_reporting and you still generate the notices. And you add some php code overhead instead of @ which is at least native code.

curry684:
Hmm issue got closed when I pushed a branch delete hehe, that wasn't intentional.

> I mean that the other calls you changed have no way to cause a fatal error AFAIK, so I don't really see the benefit of changing them if the @ works fine.

That is not true, see for example in `Composer/Factory.php`:

        foreach (array($home, $cacheDir) as $dir) {
            if (!file_exists($dir . '/.htaccess')) {
                if (!is_dir($dir)) {
                    @mkdir($dir, 0777, true);
                }
                @file_put_contents($dir . '/.htaccess', 'Deny from all');
            }
        }

The first one on `mkdir` suppresses warnings to prevent a nasty race condition, but also hides a fatal error if the disk is full or you're out of inodes. In both cases it will just die silently. The second one I don't even see the reason for the suppression, as you'll most certainly want to see everything that can go wrong there, but the solution is never to silently just create an insecure environment (perfectly illustrated by this example).

Seldaek:
I guess you're right about mkdir, the other one is arguably an unnecessary call anyway so if it fails we don't really care but yeah I see your point. As you said GPL is an issue though so if you can work it out with a Composer\Util\Something class instead that sounds good.

curry684:
Was already working on that  :)

Seldaek:
@curry684 as this got closed don't forget to re-submit a new PR when you're done :)

PR #4824, opened 2016-01-25
Update custom-installers.md

Bob4ever:


Seldaek:
Oh no we still had that in there, I had no idea, thanks :)

PR #4827, opened 2016-01-25
Added more graceful warning suppression utility

curry684:
Implementation without external dependencies of #4820

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -159,18 +159,16 @@ EOT
         $this->authConfigSource = new JsonConfigSource($this->authConfigFile, true);
 
         // Initialize the global file if it's not there, ignoring any warnings or notices
-        Silencer::suppress();
         if ($input->getOption('global') && !$this->configFile->exists()) {
             touch($this->configFile->getPath());
             $this->configFile->write(array('config' => new \ArrayObject));
-            chmod($this->configFile->getPath(), 0600);
+            Silencer::call('chmod', $this->configFile->getPath(), 0600);
         }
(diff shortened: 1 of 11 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -157,17 +158,19 @@ protected function initialize(InputInterface $input, OutputInterface $output)
...
-            @chmod($this->configFile->getPath(), 0600);
+            chmod($this->configFile->getPath(), 0600);
         }
         if ($input->getOption('global') && !$this->authConfigFile->exists()) {
             touch($this->authConfigFile->getPath());
             $this->authConfigFile->write(array('http-basic' => new \ArrayObject, 'github-oauth' => new \ArrayObject, 'gitlab-oauth' => new \ArrayObject));
-            @chmod($this->authConfigFile->getPath(), 0600);
+            chmod($this->authConfigFile->getPath(), 0600);
         }
+        Silencer::restore();
```
Seldaek: Just one thing I wonder here.. isn't it better to just use Silencer::call always? It's more overhead but these calls are far and few apart so I don't think that matters, and at least it would guarantee that if something is thrown between the suppress/restore it doesn't mean everything else in the program flow is suppressed because we failed to call restore somewhere.

curry684: As `finally` isn't supported in PHP 5.3 right now it would be most advisable indeed. In this case I thought it was safe but just noticed `write` can throw exceptions. I'll add an extra commit. Left one `::restore` in there that's inherently safe through try/catch.

Seldaek:
Thanks!

PR #4834, opened 2016-01-26
fix: missing parentheses

AppChecker:
Priority of "==" operation is higher then priority of "&" operation.

Possible defect was found by AppChecker static analyzer (http://cnpo.ru/en/solutions/appchecker.php).


Seldaek:
Thanks

PR #4838, opened 2016-01-26
Bail out if root package attempts to include itself

### LINKED ISSUE #4767: name
when the package name "packagenamexxxx"
the require list have "packagenamexxxx" package
the composer will not download the packagenamexxxx package

suggest add an tips that :

the require have the same name of current package ignore download

it's wast me two hour to find out:why the package was not download to vendor

curry684:
Fixes #4767

### DIFF
--- a/src/Composer/Package/Loader/RootPackageLoader.php
+++ b/src/Composer/Package/Loader/RootPackageLoader.php
@@ -113,8 +113,10 @@ class RootPackageLoader extends ArrayLoader
             }
         }
 
-        if (isset($links[$config['name']]))
-            throw new \InvalidArgumentException(sprintf('Root package \'%s\' cannot require itself in its composer.json', $config['name']));
+        if (isset($links[$config['name']])) {
+            throw new \InvalidArgumentException(sprintf('Root package \'%s\' cannot require itself in its composer.json' . PHP_EOL .
+                        'Did you accidentally name your root package after an external package?', $config['name']));
+        }
 
         $realPackage->setAliases($aliases);
         $realPackage->setStabilityFlags($stabilityFlags);
--- a/src/Composer/Package/Loader/RootPackageLoader.php
+++ b/src/Composer/Package/Loader/RootPackageLoader.php
@@ -113,6 +113,9 @@ class RootPackageLoader extends ArrayLoader
             }
         }
 
+        if (isset($links[$config['name']]))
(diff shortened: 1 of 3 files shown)


src/Composer/Package/Loader/RootPackageLoader.php
```diff
@@ -113,6 +113,9 @@ public function load(array $config, $class = 'Composer\Package\RootPackage', $cw
             }
         }
 
+        if (isset($links[$config['name']]))
```
alcohol: Missing some braces here `{}`.

alcohol:
Would be nice if you could include an installer test :-)

Simply drop a fixture in `tests/Composer/Test/Fixtures/installer`; look at `broken-deps-do-not-replace.test` on how to match against output text (as an example).

curry684:
That was actually far from trivial, as the fixture tests were not set up to expect Exceptions  :)

What I did now was allow for this by accepting the value `EXCEPTION` in the `--EXPECT--` field, and the specific message in the `--EXPECT-OUTPUT--` field.

Works fine now, but @Seldaek may want to review this change as well. I consider this approach better than adding an extra `--EXPECT-EXCEPTION--` field as it only introduces more obsolete parameters for edge cases, and `--EXPECT--` is already mandatory.

curry684:
Cleaned up the fixture testing for exception - you can now just set an Exception class in `--EXPECT-EXIT-CODE--` and its message in `--EXPECT--`, far cleaner than previous solution  :)

tests/Composer/Test/Fixtures/installer/install-self-from-root.test
```diff
@@ -0,0 +1,16 @@
...
+{
+    "name": "foo/bar",
+    "require": {
+        "foo/bar": "@dev"
+    }
+}
+--RUN--
+install
+--EXPECT-EXIT-CODE--
+InvalidArgumentException
```
stof: this looks weird to me as this is not an exit code

curry684: The alternative is introducing a 12th optional parameter to that function for `$expectException`... this seems less evil.

I could also globally rename `--EXPECT-EXIT-CODE--` to `--EXPECT-OUTCOME--` or `--EXPECT-EXIT-STATUS--` or something to more plainly convey it can contain an exit code or an exception.

Seldaek:
Thanks!

PR #4843, opened 2016-01-27
skip diag. which are only for phar mode

remicollet:
Version can be updated using "self-update" which is only available in phar mode

Pubkeys can be updated using "self-update --update-keys" which is also only available in phar mode

        if ('phar:' === substr(__FILE__, 0, 5)) {
            $commands[] = new Command\SelfUpdateCommand();
        }

So, this simple patch skip those diag when not in phar mode.

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -134,11 +134,13 @@ EOT
         $io->write('Checking disk free space: ', false);
         $this->outputResult($this->checkDiskSpace($config));
 
-        $io->write('Checking pubkeys: ', false);
-        $this->outputResult($this->checkPubKeys($config));
+        if ('phar:' === substr(__FILE__, 0, 5)) {
+            $io->write('Checking pubkeys: ', false);
+            $this->outputResult($this->checkPubKeys($config));
 
-        $io->write('Checking composer version: ', false);
-        $this->outputResult($this->checkVersion());
+            $io->write('Checking composer version: ', false);
+            $this->outputResult($this->checkVersion());
+        }
 
         return $this->failures;
     }


src/Composer/Command/DiagnoseCommand.php
```diff
@@ -134,11 +134,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
-        $io->write('Checking pubkeys: ', false);
-        $this->outputResult($this->checkPubKeys($config));
+        if ('phar:' === substr(__FILE__, 0, 5)) {
+            $io->write('Checking pubkeys: ', false);
+            $this->outputResult($this->checkPubKeys($config));
 
-        $io->write('Checking composer version: ', false);
-        $this->outputResult($this->checkVersion());
+            $io->write('Checking composer version: ', false);
```
stof: This should still be done when not using the phar, even though the way to fix it will depend of how you installed Composer. It is still important to warn the user that they are using an outdated version of Composer IMO

remicollet: What outdated means ?
1.0.0alpha11 is 80 days old ;)

remicollet:
Notice: this PR is mostly useful for downstream distribution which provides composer as a distro package (rpm, deb...) and will rely on "tagged" versions for update.

* https://packages.debian.org/search?keywords=composer => 1.0.0alpha11
* https://apps.fedoraproject.org/packages/composer => 1.0.0alpha11

src/Composer/Command/DiagnoseCommand.php
stof: Well, currently, using a tagged version is considered as outdated by the diagnose command. But once Composer reaches stable releases, the latest stable release will be considered as OK (and using 1.0.0 while 1.0.1 is available would be marked as outdated)

remicollet:
Thanks

PR #4845, opened 2016-01-27
Implement junctioning on Windows for path repositories

curry684:
Following up on #4690. Please note that this code was tested with a very limited test case, and you'll need to reboot if it corrupts the path repositories. Some test feedback is welcome before merging.

### DIFF
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -585,19 +585,23 @@ class Filesystem
     /**
      * Creates an NTFS junction.
      *
-     * @param string $originDir
-     * @param string $targetDir
+     * @param string $target
+     * @param string $junction
      */
-    public function junction($originDir, $targetDir)
+    public function junction($target, $junction)
     {
-        if (defined('PHP_WINDOWS_VERSION_BUILD')) {
-            $cmd = sprintf('mklink /J %s %s',
-                           ProcessExecutor::escape(str_replace('/', DIRECTORY_SEPARATOR, $targetDir)),
-                           ProcessExecutor::escape(realpath($originDir)));
-            if ($this->getProcess()->execute($cmd) === 0)
-                return;
+        if (!defined('PHP_WINDOWS_VERSION_BUILD')) {
+            throw new \LogicException(sprintf('Function %s is not available on non-Windows platform', __CLASS__));
         }
-        throw new IOException(sprintf('Failed to create junction from "%s" to "%s".', $originDir, $targetDir), 0, null, $targetDir);
+        if (!is_dir($target)) {
+            throw new IOException(sprintf('Cannot junction to "%s" as it is not a directory.', $target), 0, null, $target);
+        }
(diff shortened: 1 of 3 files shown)


src/Composer/Util/Filesystem.php
```diff
@@ -576,4 +581,54 @@ private function resolveSymlinkedDirectorySymlink($pathname)
...
+     * @param string $originDir
+     * @param string $targetDir
+     */
+    public function junction($originDir, $targetDir)
+    {
+        if (defined('PHP_WINDOWS_VERSION_BUILD')) {
+            $cmd = sprintf('mklink /J %s %s',
+                           ProcessExecutor::escape(str_replace('/', DIRECTORY_SEPARATOR, $targetDir)),
+                           ProcessExecutor::escape(realpath($originDir)));
+            if ($this->getProcess()->execute($cmd) === 0)
```
stof: missing curly braces

src/Composer/Util/Filesystem.php
```diff
@@ -576,4 +581,54 @@ private function resolveSymlinkedDirectorySymlink($pathname)
...
+                return;
+        }
+        throw new IOException(sprintf('Failed to create junction from "%s" to "%s".', $originDir, $targetDir), 0, null, $targetDir);
```
stof: IMO, the exception message should be different when you are not on windows or when it fails on windows (this will *never* work on other OS in this implementation)

src/Composer/Util/Filesystem.php
```diff
@@ -576,4 +581,54 @@ private function resolveSymlinkedDirectorySymlink($pathname)
...
+     *
+     * @param string $junction Path to check.
+     * @return bool
+     */
+    public function isJunction($junction)
+    {
+        if (!defined('PHP_WINDOWS_VERSION_BUILD')) {
+            return false;
+        }
+        $normalized = rtrim(str_replace('/', DIRECTORY_SEPARATOR, $junction), DIRECTORY_SEPARATOR);
```
stof: this does not normalize ``foo\..\bar``, making the implementation invalid (such path would be reported as being a junction)

stof:
And please write tests for the new code in the Filesystem class (skipping them when running tests on other OSes). This will ensure things work (There is a windows CI setup for Composer)

src/Composer/Util/Filesystem.php
curry684: Good point, did a bunch more testing and found a more reliable way of detection.

alcohol:
Can you update this to include benefits from changes made in #4873 ? :-)

curry684:
Of course I can   :)

alcohol:
Aside from some codestyle issues, looks good to me :-)

Seldaek:
Thanks @curry684 :)

david-garcia-garcia:
Please note that JUNCTIONING produces NON PORTABLE artifacts. Because junctions use absolute paths.

See this PR to properly support symlinks on windows:

https://github.com/composer/composer/pull/6213

PR #4847, opened 2016-01-29
Update Error Message In SolverProblemsException.php

jeffersoncarpenter:
The link might as well take you straight to the documentation, IMO.


alcohol:
Thanks

PR #4850, opened 2016-01-29
Trigger event in validate command

astehlik:
This allows plugins to run additional validations.

### DIFF
--- a/src/Composer/Command/ValidateCommand.php
+++ b/src/Composer/Command/ValidateCommand.php
@@ -14,6 +14,9 @@ namespace Composer\Command;
 
 use Composer\Factory;
 use Composer\Package\Loader\ValidatingArrayLoader;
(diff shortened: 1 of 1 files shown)


alcohol:
I think with #4124 being merged now, this will be handled quite differently in the future.

astehlik:
Thanks for your reply.

Can you give me a hint, how I should change my implementation? If I understand this new concept correctly, I need to create a new Capability and then call the plugins that provide this capability?

Seldaek:
This provides a simple hook which is fine by me I guess for this sort of simple functionality. The capability approach would be good if we want to provide more advanced integration with plugins. This doesn't extend the API available to the plugins though, if they listen to COMMAND events on older composer version they just won't be called for validate command but it doesn't break the API so it's ok as is.

PR #4851, opened 2016-01-30
Fixes in path repository handling of removed or nested sources

curry684:
If a path repository contained a package that was first successfully installed, and then something broke its source, the package would be removed before the issue was detected, leaving the dependency broken.

Along the way added a fatal check on installing a path download to (a subfolder of) its source. #4431  suggested this should be a no-op but it's so dangerous I consider it better to rely on the symlinking/junctioning behaviour instead for development scenarios.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -29,6 +29,21 @@ class PathDownloader extends FileDownloader
      */
     public function download(PackageInterface $package, $path)
     {
+        $url = $package->getDistUrl();
+        $realUrl = realpath($url);
+        if (false === $realUrl || !file_exists($realUrl) || !is_dir($realUrl)) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, would you recommend closing #4431 then? I am not familiar enough with the path repo tbh to have an informed opinion. Would appreciate a comment/review over there.

curry684:
What he suggested was a no-op, which is in my opinion far too dangerous. In the old implementation it was just a bug waiting to cause severe issues, because the original implementers just never considered and thus prepared for that scenario. Especially because it used to delete recursively before finding any problems it would likely delete the source packages (!) instead of installing anything.

So yeah, in my opinion the exception is just an awful lot saner solution to an insane scenario - better safe than sorrow. I'd say close #4431 if there's no reply in a few days, it's months old anyway.

PR #4863, opened 2016-02-02
bugfix for issue #4859 - path repo does not update when composer.json is adjusted

### LINKED ISSUE #4859: Take `autoload` into account for Locker::getContentHash()
Following #4515 and https://github.com/composer/composer/commit/6c9b161c43fe7e3f2094911ae69a9f850fd70dba, I wonder if the `Composer\Package\Locker::getContentHash()` method should take into account the `autoload` key (and maybe `autoload-dev`)? Because if I modify the autoload information of a package inside a monolyth repository (with the `path` type), then my package is not updated because composer think nothing changed.

Would you accept a PR for changing this? (not sure if changing this method has impact elsewhere tough)

alcohol:


pyrech:
:+1: It works :wink:

stof:
IMO, this change is wrong. The content-hash in the composer.lock has no reason to change when the autoload changes, as this does not impact the dependency resolution. Your change will mean that changing the root autoload will now mark the lock as outdated, while it make no sense.

The actual issue is that the path repository should not use ``Locker::getContentHash`` to generate a reference for the package. In a dependency, any change needs to be reflected by an update (as an install from lock does not read repositories), so you need to hash the whole metadata in the path repository. If you keep using the locker content hash, you will end up hashing everything in the future (due to similar reports for other property changes in the path repository), defeating the purpose of the content hash in the lock file.

So :-1: for this change.

alcohol:
@stof better?

Seldaek:
Looks fine now I think, it's a bit of a brute-force method but at least it won't miss any changes.

PR #4865, opened 2016-02-02
Keep empty folders after re-zip a module

mc388:
This is a fix to keep empty folders after re-zip the module. The solution
is to add a .keep file to each empty folder.

Known reported issues:

- composer/satis#289
- magento/magento2#2787
- magento/magento2#3248

### DIFF
--- a/src/Composer/Downloader/ArchiveDownloader.php
+++ b/src/Composer/Downloader/ArchiveDownloader.php
@@ -59,6 +59,7 @@ abstract class ArchiveDownloader extends FileDownloader
                 // move files back out of the temp dir
                 foreach ($contentDir as $file) {
                     $file = (string) $file;
+                    $this->findEmptyFolders($file);
                     $this->filesystem->rename($file, $path . '/' . basename($file));
                 }
 
@@ -153,5 +154,28 @@ abstract class ArchiveDownloader extends FileDownloader
             ->in($dir);
 
         return iterator_to_array($finder);
+    }
+
+    /**
+     * Check for empty folders and add a .keep file if it is empty
+     *
+     * @param string $file
+     */
+    private function findEmptyFolders($file) {
+        // Check if file is a directory
+        if(is_dir($file)) {
+            $directoryContent = scandir($file);
+
+            // Check if the folder is empty (there is more than . and ..)
+            if (count($directoryContent) <= 2) {
+                file_put_contents($file . "/.keep", "");
+            } else {
+                foreach ($directoryContent as $subPath) {
+                    if($subPath !== '.' && $subPath !== '..') {
+                        $this->findEmptyFolders($file.'/'.$subPath);
+                    }
+                }
+            }
+        }
     }
 }
--- a/src/Composer/Downloader/ArchiveDownloader.php
+++ b/src/Composer/Downloader/ArchiveDownloader.php
@@ -59,7 +59,7 @@ abstract class ArchiveDownloader extends FileDownloader
                 // move files back out of the temp dir
                 foreach ($contentDir as $file) {
                     $file = (string) $file;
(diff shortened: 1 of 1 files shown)


src/Composer/Downloader/ArchiveDownloader.php
```diff
@@ -154,4 +155,27 @@ private function getFolderContent($dir)
 
         return iterator_to_array($finder);
     }
+
+    /**
+     * Check for empty folders and add a .keep file if it is empty
+     *
+     * @param string $file
+     */
+    private function findEmptyFolders($file) {
```
stof: The method name is totally confusing. This method is not about finding empty folders but about adding ``.keep`` files in empty folders

ktomk:
Is there a technical reason not to use [`Phar::addEmptyDir`](http://php.net/manual/en/phar.addemptydir.php) ? If yes, I would love to hear about it!

mc388:
There are some reasons why I did not use `Phar::addEmptyDir`:

- The first reason is that git also can't handle empty folders. I know this is just an archive, but it is also a personal view.
- The next thing is that I did not know which folders are empty after archiving them.
- If I add empty folders after archiving I did not handle with the exclude files/folders.

ktomk:
I can see your points of which in case of myself I would have adopted them slightly differently (e.g. `.gitignore` instead of `.keep` in case of git but git is also not ubiquitous) and I was especially interested in technical reason not to use the `Phar::addEmptyDir`, like for let's say: It does not work.

I'd say it should be no show-stopper to actually know what to do before doing it and so therefore adding files and empty directories when it's needed in repackaging. No need to add files to the package IMHO.

mc388:
Currently it is an issue from magento 2 that they have empty folders in there packages and I don't know when they fix it. 

Moreover I found https://github.com/composer/satis/issues/289 with the same issue. Therfore I created this PR and maybe some other people need this to.

ktomk:
I can't say so either, but I'm not a fan to fix problems at the wrong end. However I think re-packaging in composer could done better by correctly taking over *empty* directories (but I have a problem seeing them filled up with new files).

It would be interesting to document if `Phar::buildFromIterator()` does process an iterator that *is* providing empty directories. I haven't checked that yet.

mc388:
`Phar::buildFromDirectory — Construct a phar archive from the files within a directory.` from [Phar::buildFromIterator()](http://www.php.net/manual/en/phar.buildfromdirectory.php) documentation.

I tried to skip the re-packaging process but how to handle different archive types? For example I download a `.tar` archive but want to offer `.zip` archives.

An other option was to replace `Phar::buildFromDirectory` with `exec(tar -c ...)` but this does not work for windows users. So this was from my point of view the easiest solution.

ktomk:
I wonder a bit for what composer needs re-packaging. I guess this is related to satis?

And still would be interesting to know if `ArchivableFilesFinder` does return entries for the empty directories or  not. Perhaps it already works if it does?

atishgoswami:
Adding a dot file to empty folders will also change the checksum of the package downloaded which might throw a man in the middle error.

Seldaek:
I think this is a definite no-go for me. We can't modify the archive in this way. Making sure we handle empty directories if present would be fine. Surely this can be achieved by adding all files/dirs one by one by going through the iterator manually instead of using buildFromIterator.

ktomk:
@Seldaek: I have not tried it but I can imagine the current iterator returns leafs only that is w/o the empty directories as entries. If they become part of the iteration it might be even work in one go. Pending a poc.

ktomk:
After playing with this locally I created a poc to include empty leaf directory paths into the zip archive. The PR is #5144

Seldaek:
Fixed by #5144

Issue #4866, opened 2016-02-02
sudo composer self-update and XDG


bamarni:
I noticed lately that composer home gets created in `~username/.config/composer`, because there is an XDG_ variable in my environment (looks related to https://github.com/composer/composer/pull/1407).

However, when I run `sudo composer self-update`, another composer directory gets created in `~username/.composer`, which is owned by root.

After this, when I run `composer install`, I get this warning at the beginning : 
```
Cannot create cache directory /home/vagrant/.composer/cache/repo/https---packagist.org/, or directory is not writable. Proceeding without cache
```

Happens on Ubuntu 14.04.3.

cs278:
This will be because sudo cleans up the environment removing the `XDG_` variables, but by default it keeps `HOME` set to your home directory use `sudo -H` to set it to `/root`. It's not optimal to run the `self-update` command using sudo without `-H` as it'll create files in your home directory owned by root.

bamarni:
I see.. The thing is that it's not commonly use I think, even composer doesn't document it (https://getcomposer.org/doc/03-cli.md#self-update), so more people will be affected.

Maybe it is fixable in composer itself somehow.

alcohol:
There is nothing to fix in Composer, as it isn't broken.

bamarni:
If there is nothing to fix in composer itself, we could at least improve the documentation, not really related to XDG but more the usage of sudo with composer commands.

Consider this other example, just following instructions from the "Getting started" page (https://getcomposer.org/doc/00-intro.md#globally) : 
``` shell
> curl -sS https://getcomposer.org/installer | sudo php -- --install-dir=/usr/local/bin --filename=composer
All settings correct for using Composer
Downloading...

Composer successfully installed to: /usr/local/bin/composer
Use it: php /usr/local/bin/composer
> composer global require phpspec/phpspec
Changed current directory to /home/vagrant/.composer


  [ErrorException]
  file_put_contents(./composer.json): failed to open stream: Permission denied


require [--dev] [--prefer-source] [--prefer-dist] [--no-plugins] [--no-progress] [--no-update] [--update-no-dev] [--update-with-dependencies] [--ignore-platform-reqs] [--sort-packages] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--] [<packages>]...
```

alcohol:
I guess we could add the `-H` switch to the documentation example.

PR #4868, opened 2016-02-03
[doc] add -H flag to sudo commands

bamarni:
cf. #4866, otherwise it causes issues since the latest security improvements and changes about XDG support.


cs278:
:+1:

Seldaek:
Thanks

PR #4869, opened 2016-02-03
Add 1.0.0-alpha12 to changelog

paladox:


Seldaek:
Thanks but we add those when doing releases.

PR #4873, opened 2016-02-03
Windows platform tests unified

curry684:
Scattered throughout the code there were 4 different ways of testing whether the code is running on Windows, each being used several to tens of times. Now there is one unified unit tested approach.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Command;
 
+use Composer\Util\Platform;
 use Composer\Util\Silencer;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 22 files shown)


stloyd:
How's this different to "tens of times" that platform is Windows or not? You just added class overhead with calling `defined()` "tens of times" anyway.

If we go this way this should be somehow cache (in-memory i.e.) that platform is or not Windows to not call same function "tens of times".

curry684:
> How's this different to "tens of times" that platform is Windows or not? You just added class overhead with calling defined() "tens of times" anyway.

Look at the diffs - there were several different tests. If one broke or changed, the whole codebase would collapse randomly. In the new situation, the unit tests will show it directly, and it can be fixed in one obvious location. As any professional programming course teaches: *"Code that would otherwise need to appear more than once in a program should be made into a function."* That's about maintainability and predictability, and it's what I've done now.

Class overhead is practically zero in these days of PHP7 and opcache, and 'inlining' is a compiler's responsibility anyway, not the programmer's. It only becomes relevant if a function is called millions of times. I've just tested `composer require symfony/symfony` on an empty repository on Windows, and `Platform::isWindows()` was invoked a grand total of 73 times. [Premature optimization is the root of all evil](http://c2.com/cgi/wiki?PrematureOptimization).

alcohol:
LGTM

PR #4875, opened 2016-02-04
Handle OpenSSL's many patch releases

cs278:
OpenSSL 0.9.8 reached a crazy `zh` patch level but Composer didn't handle this, previously it would have been detected as `0.9.8.26` now it is detected as `0.9.8.34`.

Also changed the package description to use `OPENSSL_VERSION_TEXT` so the original version number is available.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -146,8 +146,18 @@ class PlatformRepository extends ArrayRepository
                     break;
 
                 case 'openssl':
-                    $prettyVersion = preg_replace_callback('{^(?:OpenSSL\s*)?([0-9.]+)([a-z]?).*}', function ($match) {
-                        return $match[1] . (empty($match[2]) ? '' : '.'.(ord($match[2]) - 96));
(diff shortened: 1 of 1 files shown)


alcohol:
A unittest maybe? :>

Seldaek:
This isn't really testable without duplicating the whole logic in the test or mocking OPENSSL_VERSION_TEXT so I'd be ok to merge.. If you feel like refactoring it to allow testing fine by me but it's not the most critical piece of code anyway so I can live with this.

I do wonder what they'll do after zz though, becaues zh isn't so far off :)

alcohol:
Yeah, fair enough. Too much work for something not really that critical.

cs278:
> I do wonder what they'll do after zz though, because zh isn't so far off :)

Well I assumed they'd follow the same pattern so `zza`, but who knows they might switch to roman numerals.

> A unittest maybe? :>

I would, but it would have required quite a few changes.

Seldaek:
Oh right I thought you only added support for two chars at first glance but now I see it scales to infinite openssl versioning madness ;)

PR #4877, opened 2016-02-04
zlib suggest and fallback on non Windows

cs278:
* Added a suggest for `ext-zlib` which had got mangled in with `ext-zip`
* Fallback to zlib extension if available to extract gzip files


Seldaek:
Thanks

PR #4878, opened 2016-02-04
Suggest to reinstall package when failed updating from vcs metadata

bozerkins:
Good day, everyone!

Recently I have been using composer to create a production build for one of my projects.
Unfortunately for me, the plan was to store such build in VCS and deploy from VCS afterwards.
One of the easiest way of handling this (especially with git) is to delete .git folder and commit what is left.
Composer does not handle updates of source packages with missing .git folders (and any other VCS metadata folder) gracefully. Composer falls with exception, to be precise.

Thus, I have implemented an additional option (ignore-missing-metadata) into composer install command.
The option instructs the composer program, in case of failed , to remove the package folder and install it clean (with proper IO messages).

Fixed a little bug in process. 
Bug description: there was different exceptions thrown when running composer install with -vvv debug and with no / little debug, when VCS metadata folder is missing and the version of the package in composer.lock is changed from what is installed.

With regards,
Bogdans


alcohol:
While I understand the use-case, I don't personally believe that this should be solved in Composer. Instead, it is my opinion that you should fix your deployment process.

bozerkins:
Thank you for your response.

From the perspective of the problem I have faced, it seems, indeed, that the deployment process is wrong and changes should be made there.

From the perspective of composer behavior there are some points to consider (i might implemented them a little wrong, thought) as well.

When I'm running composer install (for VCS packages, in case of package version mismatch) composer works quite well for "dist" type - download a new distribution package. 
For "source" type the composer tries to update the repository. In case of failure - ends the process with exception.

But the thing is that during the "install" operation it should not really matter how exactly composer creates the packages folders. The pure goal of this operation is to install the packages into the required directory with versions that were defined previously (in composer.lock or compsoer.json). The "update" operation during "install" command is, as i see it, just a convenient way to download the next version of the source package and, as a side effect of this convenience, it fails sometimes (that may or may not be connected with missing .git folder).

What conceptually would be more accurate, is to allow users to skip this side effect, by telling the composer "that i don't really care how you download / sync / update my packages, just install what i need". With that in mind - we might change the "ignore-missing-metadata" to some more general and appropriate naming, something like "allow-vcs-reinstall".

With the following behavior:

command: composer install 
behavior: falls with exception, when failed to update VCS source package (as it is now)

command: composer install --allow-vcs-reinstall 
behavior: tries to update the package, fails, writes into output an error about that, deletes the package folder, downloads / clones / checkouts a new one

That probably would be conceptually more appropriate and general.

alcohol:
While I still not agree, I do have some feedback for the PR itself;

Rather than making it a flag/option, I would instead just make it the default behaviour. It is always in the best interest of the user to handle any kind of failure as gracefully as possible. If simply removing the existing directory and attempting to reinstall avoids any kind of failure, then that is even better.

cs278:
> If simply removing the existing directory and attempting to reinstall avoids any kind of failure, then that is even better.

Probably best to seek confirmation and continue to fail in non interactive mode.

bozerkins:
> Probably best to seek confirmation and continue to fail in non interactive mode.
Implemented this idea. 
The code seems clearer and the application behavior more user friendly.
Could implement some config value for non interactive mode later on.

bozerkins:
Good day!

Any chance this PR gets merged/reviewed?

extropic-engine:
This is also useful for situations when you don't want your deployment process to depend on external services. Being able to check in your `vendor` directory while ignoring `.git` subfolders is helpful in that you don't have to worry about whether you'll be able to download your dependencies from Github at deploy time.

alcohol:
> Being able to check in your vendor directory while ignoring .git subfolders is helpful in that you don't have to worry about whether you'll be able to download your dependencies from Github at deploy time.

That is almost identical to how we create release packages (`.tar.gz` for the win) at one of my employers. However, they are created by the build server (as part of our CI chain). So this has very little to do with composer and development environments in my humble opinion.

It is not really a reasonable expectation that you can take your development environment, bastardize and customize it to turn it into a release-able environment, and then have composer turn it back into a development environment again.

Seldaek:
Thanks, just please see https://github.com/composer/composer/commit/d347e1efca0439014ee8a31d292ff804b7e5dad5 for some changes I did, in case you wanna double check them.

PR #4881, opened 2016-02-05
"smarter" php version conflict message

alcohol:


src/Composer/DependencyResolver/Rule.php
```diff
@@ -201,13 +201,16 @@ public function getPrettyString(Pool $pool, array $installedMap = array())
                     $targetName = $this->reasonData->getTarget();
 
                     if ($targetName === 'php' || $targetName === 'php-64bit' || $targetName === 'hhvm') {
+                        $available = $pool->whatProvides($targetName);
+                        $version = count($available) ? $available[0]->getPrettyVersion() : phpversion();
```
cs278: I wonder if it might be better to display the PHP version Composer is interpreting, rather than just `phpversion()`.

curry684:
/me closes branch just forked for this very purpose   :P

src/Composer/DependencyResolver/Rule.php
```diff
@@ -201,13 +201,16 @@ public function getPrettyString(Pool $pool, array $installedMap = array())
...
+
                         // handle php/hhvm
                         if (defined('HHVM_VERSION')) {
                             $text .= ' -> your HHVM version does not satisfy that requirement.';
                         } elseif ($targetName === 'hhvm') {
                             $text .= ' -> you are running this with PHP and not HHVM.';
                         } else {
-                            $text .= ' -> your PHP version ('. phpversion() .') or value of "config.platform.php" in composer.json does not satisfy that requirement.';
+                            $text .= ' -> your PHP version or "config.platform.php" value ('.$version.') does not satisfy that requirement.';
```
curry684: Previous comment accidentally posted in wrong place.

Isn't this even more confusing? We've just seen that people cannot read an explicit pointer to `config.platform.php`, with this error message you're just giving them more places not to read.

I would build an extra else based on the constrainer, so something like in pseudocode `elseif ($platformOverridden) { $text .= "-> your config.platform.php setting $version blahblah "; } else { $text .= "your installed PHP version $version blahblah"; }`.

If people can't understand the previous error message I'd daresay you need to spell out to them in all caps that `config` is causing it, IF it is.

Seldaek:
I think if we really want to avoid confusion we should still show the actual php version *and* show the other one if present instead of resolving it to one version. That way you get max information and the discrepancy between the two would be a good hint as well.

alcohol:
You people =|

Seldaek:
Eh I can do it if you want but if you wanna improve things then they gotta be improved ;)

alcohol:
But what exactly do we want to show now?

Seldaek:
I'd say *without platform config*: `' -> your PHP version ('.phpversion().') does not satisfy that requirement.'` and *with* `' -> your PHP version ('.phpversion().') or "config.platform.php" value ('.$version.') does not satisfy that requirement.'`

alcohol:
Ok, I just need an answer to the following question in that case:

If we have no platform definition in config, does the pool still provide a php a package (but defaulting to the installed php version)?

If so, then how do we distinguish platform defined version from installed version? 
If not, then it's pretty straightforward I guess :d

Seldaek:
Well yes if you run with php at all I guess platform repo will provide one regardless.. One way to distinguish is to look at the package description.

`php                 5.6.17   Package overridden via config.platform (actual: 7.0.3)`

That's name, version and description.

Could check if description has overridden or not I guess. The default is just `The PHP interpreter`.

alcohol:
That sounds... nasty

Seldaek:
I agree, but sometimes we must do nasty things to provide a nice user experience :) We could maybe also write something in the package extra for overridden packages.. That'd be easier to programmatically check for.

alcohol:
Alright, adjusted. Had to disable the test though :-(

tests/Composer/Test/Fixtures/installer/github-issues-4319.test.skipped
```diff
@@ -0,0 +1,45 @@
+--TEST--
+
+See Github issue #4319 ( github.com/composer/composer/issues/4319 ).
+
+Present a clear error message when config.platform.php version results in a conflict rule.
+
+--CONDITION--
+!defined('HHVM_VERSION')
```
stloyd: Instead of skipping, add condition to run it only on PHP 5.3

alcohol:
Adjusted the matching of EXPECT-OUTPUT, enabled test again.

Seldaek:
Thanks!

Issue #4882, opened 2016-02-05
Json Schema 1.6.1 is breaking backwards compatibility for PHP < 5.3.29


7ochem:
Due to [this commit in justinrainbow/json-schema](https://github.com/justinrainbow/json-schema/commit/b47b52dca272a3f569f66c51ccc712c7abd2b9e5) (in release 1.6.1) backwards compatibility breaks for PHP < 5.3.29.

I was looking into rolling back the local composer.phar to a version which requires a lower version of justinrainbow/json-schema (1.6.0 or < 1.6), but looking at the commits on Composer's composer.json that affect the justinrainbow/json-schema require line, it was `~1.4` before [Jul 16, 2015](https://github.com/composer/composer/commit/9b9cbfe111b60afb8076d0a22d84107f7f6c642e), then it was `^1.4.4` up until Jan 9, 2016 when it was changed to `^1.6`. All of these come down to something in between >= 1.x.x and < 2.0.0.

So there's no way to use a certain version (commit hash) of composer that won't include justinrainbow/json-schema version 1.6.1. How can I achieve this?

alcohol:
I'm not sure I understand. It simply shouldn't install 1.6.1 for you since your environment does not support it. So what is the problem exactly?

curry684:
Does it also really break backwards compatibility? That commit in itself just raises the requirement, but it won't cause any PHP-level errors on its own. It's not really practical, but not fatal either.

If we decide to care about this on the Composer end easiest solutions seems to be to change `~1.6` to `>=1.6 <1.6.1`.

7ochem:
It is causing this:

    $ composer.phar install
    Your version of PHP, 5.3.2-1ubuntu4.21, is affected by CVE-2013-6420 and cannot safely perform certificate validation, we strongly suggest you upgrade.
    Loading composer repositories with package information
    Installing dependencies from lock file
    Your requirements could not be resolved to an installable set of packages.

    Problem 1
    - Installation request for justinrainbow/json-schema 1.6.x-dev -> satisfiable by justinrainbow/json-schema[1.6.x-dev].
    - justinrainbow/json-schema 1.6.x-dev requires php >=5.3.29 -> your PHP version (5.3.2-1ubuntu4.21) or value of "config.platform.php" in composer.json does not satisfy that requirement.
    Problem 2
    - Installation request for justinrainbow/json-schema dev-master -> satisfiable by justinrainbow/json-schema[dev-master].
[...]

PR #4883, opened 2016-02-05
Lock json-schema dependency at ~1.6.0.0

curry684:
See #4882

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     },
     "require": {
         "php": "^5.3.2 || ^7.0",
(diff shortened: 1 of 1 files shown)


stof:
This is useless. The php version checks already ensure 1.6.1 is not installed on incompatible platforms (this is the goal of having a complex dependency solver in composer and not a dumb one).

curry684:
@stof then why is https://github.com/composer/composer/issues/4882#issuecomment-180301902 apparently happening? I fully agree it ***should*** be useless.

stof:
See my reply in the discussion

Issue #4885, opened 2016-02-05
Getting PHP Warning Output from Composer/Util/Silencer.php


jaykelly450:
Since running self-update, am getting warning output from PHP when running install/update...

PHP Warning:  Unexpected character in input:  ' in phar://{..}/bin/composer/src/Composer/Util/Silencer.php on line 67

Can't locate the "'" character in the composer.json or any dependency's composer.json files.

Running PHP 5.6.8 on Debian 7.8 Wheezy

staabm:
Could you create a full backtrace?

jaykelly450:
Sure. Could you tell me how to turn on tracing from the CLI? I only see an option for verbosity as a possiblity.

jaykelly450:
In the mean time, here is the verbose output with logging on. I've included only the two unique warning, as both are output many times:

[5.8MB/0.04s] Reading ./composer.json
[6.0MB/0.04s] Loading config file ./composer.json
[6.2MB/0.05s] Checking CA file /etc/ssl/certs/ca-certificates.crt
[6.8MB/0.06s] Executing command (/home/{...}/workspaces/327/jki/php/applications/json/pagesAndFiles): git describe --exact-match --tags
[7.1MB/0.07s] Executing command (/home/{...}/workspaces/327/jki/php/applications/json/pagesAndFiles): git branch --no-color --no-abbrev -v
[7.1MB/0.08s] Executing command (/home/{...}/workspaces/327/jki/php/applications/json/pagesAndFiles): hg branch
[7.2MB/0.09s] Executing command (/home/{...}/workspaces/327/jki/php/applications/json/pagesAndFiles): svn info --xml
[9.2MB/0.14s] Reading /home/{...}/.composer/composer.json
[9.2MB/0.14s] Loading config file /home/{...}/.composer/composer.json
[9.3MB/0.15s] Reading /home/{...}/workspaces/327/jki/php/applications/json/pagesAndFiles/vendor/composer/installed.json
[9.6MB/0.16s] Reading /home/{...}/.composer/vendor/composer/installed.json
[10.8MB/0.19s] Loading composer repositories with package information
[11.3MB/0.20s] Downloading http://jkiciserver.local/satis/packages.json
[23.3MB/0.31s] Writing /home/{...}/.composer/cache/repo/http---jkiciserver.local-satis/packages.json into cache
[43.6MB/1.49s] Downloading https://packagist.org/packages.json
[...]

PR #4886, opened 2016-02-06
Ensure exception is thrown when classmaps are requested for corrupted files

curry684:
As indicated by #4885 the handling was incorrect for PHP files with invalid content.

The `Silencer` call was removed and replaced by `@` as this issue causes an `E_COMPILE_WARNING` level warning which cannot be silenced and actually points at the wrong location.

### DIFF
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -122,20 +122,20 @@ class ClassMapGenerator
             $extraTypes .= '|enum';
         }
 
-        try {
-            // Use @ here instead of Silencer to actively suppress 'unhelpful' output
-            // @link https://github.com/composer/composer/pull/4886
-            $contents = @php_strip_whitespace($path);
-            if (!$contents) {
-                if (!file_exists($path)) {
-                    throw new \RuntimeException(sprintf('File at "%s" does not exist, check your classmap definitions', $path));
-                } elseif (!is_readable($path)) {
-                    throw new \RuntimeException(sprintf('File at "%s" is not readable, check its permissions', $path));
-                }
-                throw new \RuntimeException(sprintf('File at "%s" could not be parsed as PHP - it may be binary or corrupted', $path));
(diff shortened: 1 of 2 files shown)


src/Composer/Autoload/ClassMapGenerator.php
```diff
@@ -123,14 +123,16 @@ private static function findClasses($path)
...
                 if (!file_exists($path)) {
-                    throw new \Exception('File does not exist');
-                }
-                if (!is_readable($path)) {
-                    throw new \Exception('File is not readable');
+                    throw new \RuntimeException(sprintf('File at "%s" does not exist, check your classmap definitions', $path));
+                } elseif (!is_readable($path)) {
+                    throw new \RuntimeException(sprintf('File at "%s" is not readable, check its permissions', $path));
                 }
+                throw new \RuntimeException(sprintf('File at "%s" could not be parsed as PHP - it may be binary or corrupted', $path));
```
cs278: An empty file would hit this point. What about using `error_get_last()` to pass through the original error message?

curry684: Correct, was just doublechecking it. Don't merge yet @Seldaek   ;)

curry684: I really hate PHP developers. A source file with only a comment will also fail now  :/

cs278: Shouldn't do becuase you need the `<?php`

curry684: Hmm good point, it will always keep that. I rebuilt it now with `error_get_last` though, using `error_clear_last` to check whether anything went to shit in the compiler. Should be pretty safe now, trying a huge install of all of Doctrine, Symfony and Sonata now to verify.

curry684:
@cs278 would appreciate your opinion on this last version.

Seldaek:
Looks fine I think, thanks for taking care of it. I'll merge and if @cs278 has more comments we can always address it later.

cs278:
:+1: Looks good.

PR #4887, opened 2016-02-06
Add `composer exec` command

dshafik:
This PR adds initial support for a `composer exec <script>` command. Passing in no arguments, or `--list|l` will show you the available scripts.

It allows you to pass additional arguments to the script by separating them with a `--`:

```sh
$ composer exec phpunit -- --filter ClassLoaderTest
```

**However**, if you try to pass in an argument that exists in composer (e.g. `--help` or `-V`) then it simply calls that — this is obviously sub-optimal. Preferably, anything after the `<script>` should be available as a string or an array of strings in the script, but I don't know how to make that happen.

### DIFF
--- a/src/Composer/Command/ExecCommand.php
+++ b/src/Composer/Command/ExecCommand.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Command;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * @author Davey Shafik <me@daveyshafik.com>
+ */
+class ExecCommand extends Command
+{
+    protected function configure()
+    {
+        $this
+            ->setName('exec')
+            ->setDescription('Execute a vendored binary/script')
+            ->setDefinition(array(
+                new InputOption('list', 'l', InputOption::VALUE_NONE),
+                new InputArgument('script', InputArgument::OPTIONAL, 'The script to run, e.g. phpunit'),
+                new InputArgument(
+                    'args',
+                    InputArgument::IS_ARRAY | InputArgument::OPTIONAL,
+                    'Arguments to pass to the script. Use <info>--</info> to separate from composer arguments'
+                ),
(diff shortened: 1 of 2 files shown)


weaverryan:
@dshafik Hmm. So, the help and version specifically are added early in the Application class: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Application.php#L161. It's so early, that the hook to do things like your saying (there's an event that's dispatched), is dispatched *after* this this. In fact, it's even before the Application knows *which* command is going to be run.

I can't think of a solution that works nicely here. Composer *does* have its own `Application` class, so in theory this behavior could be overridden. But the trick is: you won't know (yet) whether it's *this* command being run, or another (where you *do* want --help to be run). I suppose you'd need to do some parsing to figure that out... in the `Application` class, which isn't a great solution.

dzuelke:
So @dshafik... I can think of several scenarios where it might be interesting to not simply run a vendor binary, but instead to run any command, with `vendor/bin/` on the path (like `composer run-script anyscript` already does). Think cases like `composer exec 'phpunit || somethingelse'`.

In fact, you may be able to simply build upon the "composer run-script" functionality, and refactor it to accept "anonymous commands".

Some benefits of this would be that

1. anything you call may in turn invoke another vendored binary because it's on the path;
1. tooling can invoke `composer exec foobar`, and foobar will be taken from the path, whether fed from `$(composer config bin-dir)` or another location;
1. pipelines become possible (think the example above, or `cat tests.txt | phpunit` - PHPUnit doesn't accept stdin input, but you get the idea).

Also, please don't refer to these binaries as "scripts". Scripts are a different thing in Composer, and the docs refer to what you're doing here as "vendor binaries" (https://getcomposer.org/doc/articles/vendor-binaries.md).

/cc @Seldaek @naderman

dzuelke:
ping @dshafik

dzuelke:
Here is a way to do this out of the box via `composer.json` :D

```
"scripts": {
    "exec": "$(\"$@\")"
}
```

Then run `composer exec whoami` or `composer exec doctrine` :)

:rocket: :rocket: :rocket: :rocket: :rocket:

Seldaek:
Thanks @dshafik, I also implemented the PATH thing that @dzuelke suggested.

dzuelke:
I wonder if if would be better to *append* to `$PATH` (I know, that'll be complicated) for this functionality. That way, `$(composer bin-dir)/somebin` would be guaranteed to run whatever has been vendored, while `composer exec somebin` would give preference to an existing "somebin".

But maybe that makes no sense. Hmmh.

Or a new --append-path flag? :)

Either way, what about my concern to not call these things "scripts", but instead "vendor binaries", @Seldaek ? want me to do a PR to fix that?

Seldaek:
I fixed it as well, but if you still see some in latest feel free to PR.

As for the PATH. I don't really see how it makes sense no :)

PR #4894, opened 2016-02-09
Be sure that configuration file can be manipulated by actual user

stloyd:
Without this change if user runs, i.e.: `composer update` but global Composer was installed by other user, he can't store new token for Github authorization due too fact that he can't access this config file, which leads to "infinity loop" with generating new tokens.


alcohol:
Thanks

Issue #4893, opened 2016-02-08
Composer/Source User-Agent


chosroes:
Hi,

On the packages.json download (by file_get_contents), the user agent seems to be "Composer/Source" and this is a problem in my case since the server doesn't allow this user-agent and I don't know why and at which level...
My question is: Is it possible to specify another user agent on composer install/update ?

Thanks

alcohol:
This is currently not possible. Is there a particular reason this user agent is not allowed on your server?

chosroes:
@alcohol Hosting provider is responsible. I think for security reasons... But he doesn't want to change this behavior.

alcohol:
Do you mean, as an outgoing connection, it is not allowed? E.g. composer is not able to connect to packagist.org? Can you give us the output of a `composer -vvv update` on your project?

Seldaek:
Then please consider using another hosting provider.. seriously sorry but this is just a waste of time for everyone involved because of some bad actor. Not setting a proper user-agent would be bad practice for us.

curry684:
I agree it's bad on the hosting provider's part, and just breaking standards for fake security. But as an environment variable I think it may even be more used than stuff like `COMPOSER_DISABLE_XDEBUG_WARN` that's also in there. Not saying it should be any kind of priority, but if someone (@chosroes?) were to issue a PR to properly implement `COMPOSER_USER_AGENT` I think it would be fine to merge  :)

chosroes:
To be more specific, I have installed Satis on the server and I use composer on my dev environment with this Satis as composer repository.
But, between my computer and the Satis server, there is a "firewall" server where the problem is.

So, when I update, composer sends those headers: 
"Connection: close"
"Accept-Encoding: gzip"
"User-Agent: Composer/6eb50623e6e00bac4ed893802836e50b62c83434 (Linux; 4.2.0-27-generic; PHP 5.6.11)"

The composer file_get_contents on the Satis packages.json doesn't fail, but retrieve an html:
`"<HTML>
<HEAD><TITLE>Index of /</TITLE></HEAD>
<BODY BGCOLOR="#99cc99" TEXT="#000000" LINK="#2020ff" VLINK="#4040cc">
<H2>Index of /</H2>
<PRE>
mode  links  bytes  last-changed  name
<HR>dr-x   2        4096  Jan 31  2015  <A HREF="/./">.</A>/
dr-x   7        4096  Sep 16 11:26  <A HREF="/../">..</A>/
</PRE></BODY>
[...]

PR #4896, opened 2016-02-09
Disable PHP and system exposure in stream context factory

chosroes:
This can be needed for security reasons.
In some cases, user doesn't want or allow composer to send system information in the request. So comes COMPOSER_DISABLE_PHP_EXPOSE.
The php_uname function is often disabled for security reason and must not be mandatory.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -737,4 +737,8 @@ If set to 1, this env var will make Composer behave as if you passed the
 
 If set to 1, this env disables the warning about having xdebug enabled.
 
+### COMPOSER_DISABLE_PHP_EXPOSE
+
+If set to 1, this env disables the php version exposure in requests.
+
 &larr; [Libraries](02-libraries.md)  |  [Schema](04-schema.md) &rarr;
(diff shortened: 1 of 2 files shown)


staabm:
so this PR is motivated by the edge case in https://github.com/composer/composer/issues/4893 , right?

chosroes:
@staabm Yes, but I think it's not inappropriate in a global usecase

curry684:
:-1: I'd support `COMPOSER_USER_AGENT` as I can see its generic usability in rare edge cases, but as `COMPOSER_DISABLE_PHP_EXPOSE` you're limiting the applicability. What if a stupid firewall only allows Chrome browsers? Or iOS? Or needs an exact match on `Certified Company PHP Developer`?

Solve all of them or none imho.

chosroes:
@curry684 Are you satisfied like that ?

Seldaek:
Thanks for the PR, however given that AFAIK we have never had an issue reported because of php_uname, and the fact that #4893 is also a really weird one-off issue, I'd rather not add this.

Issue #4852, opened 2016-01-30
Running php scripts triggered by event causes Error Output: sh: $exec: not found


vaclavvanik:
I have problem when php scripts are triggered by events e.g. post-update-cmd. Event ends with error: `Error Output: sh: app-clear: not found`

When I run script manually `composer app-clear-cache` or `composer run-script post-update-cmd` everything works fine.

My composer.json looks:

```
{
    "scripts": {
        "app-clear-cache": [
            "app-clear data/cache/*"
        ],
        "post-update-cmd": [
            "@app-clear-cache"
        ]
    }
}
```

`app-clear` is in `vendor/bin`. and looks:

```
#!/usr/bin/env php
<?php

...

foreach ($patterns as $pattern) {
    array_map('unlink', glob($pattern));
}

```

Is this bug or expected behavior?

alcohol:
Hmm. I was under the impression that the bin dir was automatically added to the path of where Composer looks for executables in scripts. The above seems to contradict this however. When I have some time I will look further into this. Can you try running it as `vendor/bin/app-clear data/cache/*` for now?

PR #4898, opened 2016-02-09
bin directory detection fix for issue #4852

bozerkins:
the problem was that bin directory was not added to PATH
in update / install and other commands, except RunScript and ScriptAlias


Seldaek:
Thanks for the PR but the fix as you had it was duplicating code and removing all env vars except PATH from script handlers which is not really acceptable. I did an alternate fix which should do the job.

PR #4905, opened 2016-02-11
Added ignore-empty option for Path repositories, fixes #4903

### LINKED ISSUE #4903: Composer should ignore empty path repositories
Use case: when using Composer in local development, it's great to have a directory of local packages (so devs can work on them simultaneously), which can then be defined in `composer.json` as a Path repository. Great! (It can't be defined in the global `config.json`, because it wouldn't take priority over other repositories defined in `composer.json`.)

However, let's assume that on the production server, we don't want this directory to exist, because we're pulling the packages from the next available, lower-priority repository.

This almost works ... except that Composer gives an error when the path repository is empty (or doesn't exist):

> [RuntimeException]
>  No `composer.json` file found in any path repository in "../empty-path/*/*"

The solution could be to add an option to the Path repository to explicitly ignore errors when no `composer.json` could be found. What do you guys think?

cinamo:
Adds a new option to ignore empty Path repositories to the config. Fixes issues #4903.

### DIFF
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -71,7 +71,7 @@ class PathRepository extends ArrayRepository implements ConfigurableRepositoryIn
     /**
      * @var boolean
      */
-    private $ignoreEmpty;
+    private $ignoreEmpty = false;
 
     /**
      * @var array
@@ -108,11 +108,6 @@ class PathRepository extends ArrayRepository implements ConfigurableRepositoryIn
         $this->repoConfig = $repoConfig;
 
         parent::__construct();
-    }
-
-    function __call($name, $arguments)
-    {
-        // TODO: Implement __call() method.
     }
 
     public function getRepoConfig()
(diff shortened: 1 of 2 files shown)


src/Composer/Repository/PathRepository.php
```diff
@@ -98,6 +110,11 @@ public function __construct(array $repoConfig, IOInterface $io, Config $config)
         parent::__construct();
     }
 
+    function __call($name, $arguments)
```
stloyd: This should be removed.

src/Composer/Repository/PathRepository.php
```diff
@@ -141,7 +158,7 @@ protected function initialize()
             $this->addPackage($package);
         }
 
-        if (count($this->getPackages()) == 0) {
+        if (count($this->getPackages()) == 0 && $this->ignoreEmpty !== true) {
```
stloyd: That new check should be first as it's less costing than count.

src/Composer/Repository/PathRepository.php
```diff
@@ -67,6 +69,11 @@ class PathRepository extends ArrayRepository implements ConfigurableRepositoryIn
     private $url;
 
     /**
+     * @var boolean
+     */
+    private $ignoreEmpty;
```
stloyd: Should be `false` by default.

cinamo:
Thanks, updated those things!

PR #4906, opened 2016-02-11
Prefer 'editor' on a system that supports alternatives

curry684:
On Linux systems that support [alternatives](http://linux.about.com/library/cmd/blcmdl8_alternatives.htm), which would be pretty much all of them, the `editor` symlink refers to the preferred editor by the current user, so if it exists it should be preferred over all other editors.

Wouldn't want to start an [editor war](https://en.wikipedia.org/wiki/Editor_war) ;)


staabm:
thx for pointing to alternatives, wasnt aware of it. ;-)

sounds like a great change :+1:

Seldaek:
Makes sense thanks

Issue #4777, opened 2016-01-16
Move http support behind a secure-http flag


Seldaek:
As per discussion on #4056 - adding a flag that disables http unless the flag is set to false would be good. I'm thinking `secure-http: false` reads more dangerous than `https-only: false` so might be preferable. Anyway this should be fairly easy to add in Config/ConfigCommand and check for in the RemoteFilesystem file if someone wants to help.

Seldaek:
Implemented in #4907 /cc @stormtide ;)

davecarlson:
Please note, this breaks a lot of projects without config changes. 

For example, any project that uses wordpress packages, and calls wpackagist.org, needs changes to be able to be installed. This is pretty serious if your stack have auto-scaling,  resulting in failure.

Seldaek:
Yes I am sorry for the eventual breakage but it's a security fix and as such not fixing it would also be a bad situation.

davecarlson:
That's fine - although should there not have been some kind of announcement regarding this, seeing as though it's a breaking change?

Seldaek:
I did https://twitter.com/seldaek/status/705415089357266944 on top of the actual beta1 release with the changelog and warning on the first line.

davecarlson:
fair enough - you're now followed :)

PR #4907, opened 2016-02-11
Added secure-http flag, defaults to true

cinamo:
@Seldaek Something like this for issue #4777?

It seems somewhere there's still http://packagist.org links generated, so I added a little fix for those specifically. I guess that shouldn't be necessary, so maybe you can point me to the right spot to fix?

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -13,8 +13,8 @@
 namespace Composer\Util;
 
 use Composer\Config;
-use Composer\IO\IOInterface;
 use Composer\Downloader\TransportException;
+use Composer\IO\IOInterface;
 
 /**
  * @author François Pluchino <francois.pluchino@opendisplay.com>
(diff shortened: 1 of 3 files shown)


barryvdh:
You should usually not change multiple things in 1 PR, like unrelated CS fixes. You can do that in a different PR :)

cinamo:
Ouch, you're right of course. Meant to do that for just the method I added. What's the best way to proceed?

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -993,11 +1019,11 @@ private function getCertificateCnAndFp($url, $options)
     private function getUrlAuthority($url)
     {
         $defaultPorts = array(
-            'ftp' => 21,
-            'http' => 80,
-            'https' => 443,
+            'ftp'       => 21,
+            'http'      => 80,
+            'https'     => 443,
```
stof: -1 for this CS change. Aligning array arrows makes the maintenance a huge pain by generating extra useless diffs regularly (which means extra useless git conflicts potentially, and also breaks the usage of ``git blame``)

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -248,12 +250,28 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
             $fileUrl = 'http://' . gethostbyname('packagist.org') . substr($fileUrl, 20);
         }
 
-        $ctx = StreamContextFactory::getContext($fileUrl, $options, array('notification' => array($this, 'callbackGet')));
+        $ctx = StreamContextFactory::getContext($fileUrl, $options,
+            array('notification' => array($this, 'callbackGet')));
```
stof: such wrapping does not make sense (and is not PSR-2 compliant btw)

curry684:
Wtf is happening in Travis there, did Pear get nuked?

Seldaek:
Thanks!

PR #4908, opened 2016-02-11
Add recursive capabilities to  'composer depends'

curry684:
Add recursive capabilities and more details in verbose mode to `composer depends`.

Testcase: `composer require symfony/symfony doctrine/orm`

Output before:
```
$ composer depends doctrine/lexer
doctrine/annotations requires doctrine/lexer (1.*)
doctrine/common requires doctrine/lexer (1.*)
```
New output modes and combinations:
```
$ composer depends doctrine/lexer
doctrine/annotations requires doctrine/lexer (1.*)
doctrine/common requires doctrine/lexer (1.*)

$ composer depends doctrine/lexer -v
doctrine/annotations v1.2.7 requires doctrine/lexer (1.*)
doctrine/common v2.6.1 requires doctrine/lexer (1.*)

$ composer depends doctrine/lexer -r
doctrine/annotations requires doctrine/lexer (1.*)
doctrine/common requires doctrine/lexer (1.*)
doctrine/common requires doctrine/annotations (1.*)
doctrine/dbal requires doctrine/common (>=2.4,<2.7-dev)
doctrine/orm requires doctrine/common (>=2.5-dev,<2.7-dev)
symfony/symfony requires doctrine/common (~2.4)
 => root package requires symfony/symfony (^3.0)
doctrine/orm requires doctrine/dbal (>=2.5-dev,<2.6-dev)
symfony/symfony requires (dev) doctrine/dbal (~2.4)
 => root package requires doctrine/orm (^2.5)
symfony/symfony requires (dev) doctrine/orm (~2.4,>=2.4.5)

$ composer depends doctrine/lexer -r -v
doctrine/annotations v1.2.7 requires doctrine/lexer (1.*)
doctrine/common v2.6.1 requires doctrine/lexer (1.*)
doctrine/common v2.6.1 requires doctrine/annotations (1.*)
doctrine/dbal v2.5.4 requires doctrine/common (>=2.4,<2.7-dev)
doctrine/orm v2.5.4 requires doctrine/common (>=2.5-dev,<2.7-dev)
symfony/symfony v3.0.2 requires doctrine/common (~2.4)
 => root package requires symfony/symfony (^3.0)
doctrine/orm v2.5.4 requires doctrine/dbal (>=2.5-dev,<2.6-dev)
symfony/symfony v3.0.2 requires (dev) doctrine/dbal (~2.4)
 => root package requires doctrine/orm (^2.5)
symfony/symfony v3.0.2 requires (dev) doctrine/orm (~2.4,>=2.4.5)
```

### DIFF
--- a/src/Composer/Command/DependsCommand.php
+++ b/src/Composer/Command/DependsCommand.php
@@ -48,6 +48,7 @@ class DependsCommand extends Command
                 new InputOption('match-constraint', 'm', InputOption::VALUE_REQUIRED, 'Filters the dependencies shown using this constraint', '*'),
                 new InputOption('invert-match-constraint', 'i', InputOption::VALUE_NONE, 'Turns --match-constraint around into a blacklist instead of whitelist'),
                 new InputOption('with-replaces', '', InputOption::VALUE_NONE, 'Search for replaced packages as well'),
+                new InputOption('recursive', 'r', InputOption::VALUE_NONE, 'Recursively discover all dependencies'),
             ))
             ->setHelp(<<<EOT
 Displays detailed information about where a package is referenced.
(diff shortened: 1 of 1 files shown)


alcohol:
I'm a bit confused about the output. How does the root package requiring orm, relate to dbal?

curry684:
I did choose a pretty messy example hehe, as it has double diamond dependencies. Essentially it's a bottom up explosion of the requested package. I agree readability could be better, but doing so would mean also changing the default behavior.

Perhaps reversing the output would be better readable, and I could also add an optional tree view like `c show -it`

stof:
dev requirements should be taken into account only for the root package, as require-dev is root-only

curry684:
Hmm good point. I didn't consider that dev could be relevant, as it's a bottom up approach of an actually installed package, so there was always a valid reason for it to be installed anyway. But in rare cases a package could both be referenced without dev and in another path with, and it will now show both. It will never lead to really 'wrong' output though, just an extra dependency path which was never an original install reason.

Fixing that will mean rewriting the entire command though, not a bad idea as it's pretty messy, but a lot more work. It would also open up better rendering options though as I would decouple the resolution from the rendering then hehe.

stof:
@curry684 what you should do for a quick fix for that is to consider the requireDev link type only when the package is the root package

curry684:
@stof the problem is that that bug was already in there:
```
$ cat composer.json
{
    "require": {
        "symfony/symfony": "^3.0",
        "doctrine/orm": "^2.5"
    },
    "require-dev": {
        "phpunit/phpunit": "^5.2"
    }
}
$ composer depends phpunit/phpunit
__root__ requires (dev) phpunit/phpunit (^5.2)
doctrine/annotations requires (dev) phpunit/phpunit (4.*)
doctrine/cache requires (dev) phpunit/phpunit (~4.8|~5.0)
doctrine/collections requires (dev) phpunit/phpunit (~4.0)
doctrine/common requires (dev) phpunit/phpunit (~4.8|~5.0)
doctrine/dbal requires (dev) phpunit/phpunit (4.*)
doctrine/inflector requires (dev) phpunit/phpunit (4.*)
doctrine/instantiator requires (dev) phpunit/phpunit (~4.0)
doctrine/orm requires (dev) phpunit/phpunit (~4.0)
myclabs/deep-copy requires (dev) phpunit/phpunit (~4.1)
paragonie/random_compat requires (dev) phpunit/phpunit (4.*|5.*)
phpdocumentor/reflection-docblock requires (dev) phpunit/phpunit (~4.0)
phpunit/php-code-coverage requires (dev) phpunit/phpunit (~5)
phpunit/php-token-stream requires (dev) phpunit/phpunit (~4.2)
phpunit/phpunit-mock-objects requires (dev) phpunit/phpunit (~5)
sebastian/comparator requires (dev) phpunit/phpunit (~4.4)
sebastian/diff requires (dev) phpunit/phpunit (~4.8)
sebastian/environment requires (dev) phpunit/phpunit (~4.4)
sebastian/exporter requires (dev) phpunit/phpunit (~4.4)
sebastian/global-state requires (dev) phpunit/phpunit (~4.2)
sebastian/recursion-context requires (dev) phpunit/phpunit (~4.4)
```

curry684:
I'll close this PR for now, it needs a fundamental rewrite.

alcohol:
If you do go for a fundamental rewrite, it would be nice if you move as much logic as possible out of the Command class.

PR #4910, opened 2016-02-12
fix documentation which use a single pipe for a "logical OR"

Oliboy50:
As stated here: https://getcomposer.org/doc/articles/versions.md#range
A logical OR is done using a double pipe "||".

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -361,7 +361,7 @@ will not be allowed to be installed together with your package.
 Note that when specifying ranges like `<1.0 >=1.1` in a `conflict` link,
 this will state a conflict with all versions that are less than 1.0 *and* equal
 or newer than 1.1 at the same time, which is probably not what you want. You
-probably want to go for `<1.0 | >=1.1` in this case.
+probably want to go for `<1.0 || >=1.1` in this case.
 
 #### replace
 


alcohol:
Thanks

stof:
Actually, both are allowed (the single pipe is older). But it is indeed better to be consistent regarding the documented one

phansys:
@alcohol, is there any deprecation path at code for the old operator `|`?

Seldaek:
@phansys not sure I understand your sentence, but we can not effectively deprecate operators as that would break all old packages/tags that use them. We can only document that they are deprecated to nudge people towards using newer operators.

phansys:
Thank you for the response @Seldaek, you just understood my sentence as is. It comes from [this @fabpot's question](https://github.com/symfony/symfony/pull/18186#issuecomment-197299788).
The deprecations can not be handled here with `trigger_error()` alongside `E_USER_DEPRECATED` as in Symfony?

Seldaek:
No we are not going to trigger warnings as that would not make any sense since you can not fix the warnings you would see when using old packages. `|` and `||` will work the same forever..

phansys:
:+1: then, thank you for the clarification.

phansys:
@Seldaek, do you thing it could have a related validator at `ValidatingArrayLoader` in order to show a warning when the old operator is used?

curry684:
@phansys as there is no standard neither is old or deprecated.

phansys:
Thanks @curry684, I see at [mojombo/semver](https://github.com/mojombo/semver) that there aren't any documentation about the operators since the spec tends to be agnostic to these implementation details.

PR #4913, opened 2016-02-12
fix compat with symfony console < 2.8

remicollet:
See https://github.com/composer/composer/commit/49d7d65933c78e163354c2c580883926f1c8daee

As verbosity is tested in the start of doWrite function

        if ($sfVerbosity > $this->output->getVerbosity()) {
            return;
        }

There is no need to pass its value to write.
This restore compatibility with old version (tested with 2.7.9 and 2.8.2)

### DIFF
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -139,13 +139,13 @@ class ConsoleIO extends BaseIO
         }
 
         if (true === $stderr && $this->output instanceof ConsoleOutputInterface) {
-            $this->output->getErrorOutput()->write($messages, $newline, $sfVerbosity);
+            $this->output->getErrorOutput()->write($messages, $newline);
             $this->lastMessageErr = join($newline ? "\n" : '', (array) $messages);
 
             return;
         }
 
-        $this->output->write($messages, $newline, $sfVerbosity);
+        $this->output->write($messages, $newline);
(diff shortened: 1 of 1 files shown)


remicollet:
The other way is to bump requirement to symfony/console ^2.8

But this will prevent downstream distribution, e.g. in Fedora
(we are still sticked to symfony 2.7 for now... and when you notice such API break in a minor version...)

siwinski:
FYI (in anyone is interested): Fedora/EPEL is sticking to Symfony 2.7 for now because of [Drupal 8](https://www.drupal.org/8)'s [dependencies](http://cgit.drupalcode.org/drupal/tree/core/composer.json)

curry684:
As 2.7 is the current LTS it makes sense for both Composer and Fedora (and Drupal) to stay on there for a while. PR looks fine.

barryvdh:
Although 2.8 is also a LTS version; http://symfony.com/doc/current/contributing/community/releases.html#long-term-support-versions

Seldaek:
2.8 is also LTS yes. And the PR breaks the VERBOSITY_QUIET flag which, if not specified, will not let you write stuff through a --quiet as the symfony output will swallow it. I'll push a fix in a sec that I believe is more appropriate. Let me know if it goes well for you please.

remicollet:
Perfect, thanks.

PR #4917, opened 2016-02-15
Rewrote DependsCommand

### LINKED ISSUE #4911: Depends incorrectly handles dev dependencies
`composer depends` command should only consider `require-dev` dependencies from the root package, but it doesn't:
```
$ cat composer.json
{
    "require": {
        "symfony/symfony": "^3.0",
        "doctrine/orm": "^2.5"
    },
    "require-dev": {
        "phpunit/phpunit": "^5.2"
    }
}
$ composer depends phpunit/phpunit
__root__ requires (dev) phpunit/phpunit (^5.2)
doctrine/annotations requires (dev) phpunit/phpunit (4.*)
doctrine/cache requires (dev) phpunit/phpunit (~4.8|~5.0)
doctrine/collections requires (dev) phpunit/phpunit (~4.0)
doctrine/common requires (dev) phpunit/phpunit (~4.8|~5.0)
doctrine/dbal requires (dev) phpunit/phpunit (4.*)
doctrine/inflector requires (dev) phpunit/phpunit (4.*)
doctrine/instantiator requires (dev) phpunit/phpunit (~4.0)
doctrine/orm requires (dev) phpunit/phpunit (~4.0)
myclabs/deep-copy requires (dev) phpunit/phpunit (~4.1)
paragonie/random_compat requires (dev) phpunit/phpunit (4.*|5.*)
phpdocumentor/reflection-docblock requires (dev) phpunit/phpunit (~4.0)
phpunit/php-code-coverage requires (dev) phpunit/phpunit (~5)
phpunit/php-token-stream requires (dev) phpunit/phpunit (~4.2)
phpunit/phpunit-mock-objects requires (dev) phpunit/phpunit (~5)
sebastian/comparator requires (dev) phpunit/phpunit (~4.4)
sebastian/diff requires (dev) phpunit/phpunit (~4.8)
sebastian/environment requires (dev) phpunit/phpunit (~4.4)
sebastian/exporter requires (dev) phpunit/phpunit (~4.4)
sebastian/global-state requires (dev) phpunit/phpunit (~4.2)
sebastian/recursion-context requires (dev) phpunit/phpunit (~4.4)
```
It should only output the first line of course.

I'm in the process of rewriting the entire DependsCommand, but logging a separate issue for this bug in the current version in case I don't get that done for some reason.

curry684:
Rewrote `composer depends` completely.

- Dropped `--link-type` parameter as it implied wrong logic, fixes #4911 
- Dropped `--with-replaces` parameter as it caused [nonsensical indirect hits](https://github.com/composer/composer/issues/4911#issuecomment-183966204). Direct replaces are always shown though
- Added `--recursive` parameter to recursively find up to the root why a package is present
- Added `--tree` to render recursive output in a nice tree without undubbing

Options `-i` and `-m` were preserved.

Sample output:
```
$ composer depends doctrine/lexer
 doctrine/annotations v1.2.7 requires doctrine/lexer (1.*)
 doctrine/common      v2.6.1 requires doctrine/lexer (1.*)

$ composer depends doctrine/lexer -r
 __root__             -      requires doctrine/orm (^2.5)
 doctrine/orm         v2.5.4 requires doctrine/dbal (>=2.5-dev,<2.6-dev)
 __root__             -      requires symfony/symfony (^3.0)
 doctrine/common      v2.6.1 requires doctrine/annotations (1.*)
 doctrine/dbal        v2.5.4 requires doctrine/common (>=2.4,<2.7-dev)
 doctrine/orm         v2.5.4 requires doctrine/common (>=2.5-dev,<2.7-dev)
 symfony/symfony      v3.0.2 requires doctrine/common (~2.4)
 doctrine/annotations v1.2.7 requires doctrine/lexer (1.*)
 doctrine/common      v2.6.1 requires doctrine/lexer (1.*)

$ composer depends doctrine/lexer -t
doctrine/lexer v1.0.1 Base library for a lexer that can be used in Top-Down, Recursive Descent Parsers.
|- doctrine/annotations v1.2.7 (requires doctrine/lexer 1.*)
|  `- doctrine/common v2.6.1 (requires doctrine/annotations 1.*)
|     |- doctrine/dbal v2.5.4 (requires doctrine/common >=2.4,<2.7-dev)
|     |  `- doctrine/orm v2.5.4 (requires doctrine/dbal >=2.5-dev,<2.6-dev)
|     |     `- __root__ (requires doctrine/orm ^2.5)
|     |- doctrine/orm v2.5.4 (requires doctrine/common >=2.5-dev,<2.7-dev)
|     |  `- __root__ (requires doctrine/orm ^2.5)
|     `- symfony/symfony v3.0.2 (requires doctrine/common ~2.4)
|        `- __root__ (requires symfony/symfony ^3.0)
`- doctrine/common v2.6.1 (requires doctrine/lexer 1.*)
   |- doctrine/dbal v2.5.4 (requires doctrine/common >=2.4,<2.7-dev)
   |  `- doctrine/orm v2.5.4 (requires doctrine/dbal >=2.5-dev,<2.6-dev)
   |     `- __root__ (requires doctrine/orm ^2.5)
   |- doctrine/orm v2.5.4 (requires doctrine/common >=2.5-dev,<2.7-dev)
   |  `- __root__ (requires doctrine/orm ^2.5)
   `- symfony/symfony v3.0.2 (requires doctrine/common ~2.4)
      `- __root__ (requires symfony/symfony ^3.0)
```
BC parameters:
```
$ composer depends symfony/symfony -m ^4.0
There is no installed package depending on "symfony/symfony" in versions matching 4.0

$ composer depends symfony/symfony -m ^3.9
 __root__ - requires symfony/symfony (^3.0)

$ composer depends symfony/symfony -im ^3.9
There is no installed package depending on "symfony/symfony" in versions not matching 3.9

$ composer depends symfony/symfony -im ^2.7
 __root__ - requires symfony/symfony (^3.0)
```
Tests for #4911 and the new `replace` detection:
```
$ composer depends phpunit/phpunit
 __root__ - requires (for development) phpunit/phpunit (^5.2)

$ composer depends symfony/yaml
 phpunit/phpunit 5.2.4  requires symfony/yaml (~2.1|~3.0)
 symfony/symfony v3.0.2 replaces symfony/yaml (self.version)

$ composer depends symfony/yaml -t
symfony/symfony v3.0.2 The Symfony PHP framework
|- phpunit/phpunit 5.2.4 (requires symfony/yaml ~2.1|~3.0)
|  `- __root__ (requires (for development) phpunit/phpunit ^5.2)
`- symfony/symfony v3.0.2 (replaces symfony/yaml self.version)
   `- __root__ (requires symfony/symfony ^3.0)
```

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -350,12 +350,11 @@ php composer.phar depends symfony/symfony --with-replaces -im ^3.0
 
 ### Options
 
-* **--link-type:** The link types to match on, can be specified multiple
-  times.
+* **--recursive (-r):** Recursively resolves up to the root package.
+* **--tree (-t):** Prints the results as a nested tree, implies -r.
 * **--match-constraint (-m):** Filters the dependencies shown using this constraint.
 * **--invert-match-constraint (-i):** Turns --match-constraint around into a blacklist
   instead of a whitelist.
-* **--with-replaces:** Search for replaced packages as well.
 
 ## validate
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -337,16 +337,11 @@ php composer.phar depends doctrine/lexer
  doctrine/common      v2.6.1 requires doctrine/lexer (1.*)
 ```
 
-If you want, for example, to find any installed package that is **not**
-allowing Monolog to be upgraded to version 1.17 , try this:
+You can optionally specify a version constraint after the package to limit the
+search.
 
-```sh
-php composer.phar depends monolog/monolog -im ^1.17
-There is no installed package depending on "monolog/monolog" in versions not matching 1.17
-```
-
-Add the `--tree` or `-t` flag to show a recursive tree of why the package is depended
-upon, for example:
+Add the `--tree` or `-t` flag to show a recursive tree of why the package is
+depended upon, for example:
 
 ```sh
 php composer.phar depends psr/log -t
@@ -364,9 +359,36 @@ psr/log 1.0.0 Common interface for logging libraries
 
 * **--recursive (-r):** Recursively resolves up to the root package.
 * **--tree (-t):** Prints the results as a nested tree, implies -r.
-* **--match-constraint (-m):** Filters the dependencies shown using this constraint.
-* **--invert-match-constraint (-i):** Turns --match-constraint around into a blacklist
-  instead of a whitelist.
+
+## prohibits
+
+The `prohibits` command tells you which packages are blocking a given package
+from being installed. Specify a version constraint to verify whether upgrades
+can be performed in your project, and if not why not. See the following
+example:
+
+```sh
+php composer.phar prohibits symfony/symfony 3.1
+ laravel/framework v5.2.16 requires symfony/var-dumper (2.8.*|3.0.*)
+```
+
+Note that you can also specify platform requirements, for example to check
+whether you can upgrade your server to PHP 8.0:
+
+```sh
+php composer.phar prohibits php:8
+ doctrine/cache        v1.6.0 requires php (~5.5|~7.0)
+ doctrine/common       v2.6.1 requires php (~5.5|~7.0)
+ doctrine/instantiator 1.0.5  requires php (>=5.3,<8.0-DEV)
+```
+
+As with `depends` you can request a recursive lookup, which will list all
+packages depending on the packages that cause the conflict.
+
+### Options
+
+* **--recursive (-r):** Recursively resolves up to the root package.
+* **--tree (-t):** Prints the results as a nested tree, implies -r.
 
 ## validate
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -328,24 +328,36 @@ to limit output to suggestions made by those packages only.
 ## depends
 
 The `depends` command tells you which other packages depend on a certain
-package. You can specify which link types (`require`, `require-dev`)
-should be included in the listing. By default both are used.
+package. As with installation `require-dev` relationships are only considered
+for the root package.
 
 ```sh
-php composer.phar depends --link-type=require monolog/monolog
-
-nrk/monolog-fluent requires monolog/monolog (~1.8)
-poc/poc requires monolog/monolog (^1.6)
-propel/propel requires monolog/monolog (1.*)
-symfony/monolog-bridge requires monolog/monolog (>=1.2)
-symfony/symfony requires monolog/monolog (~1)
+php composer.phar depends doctrine/lexer
+ doctrine/annotations v1.2.7 requires doctrine/lexer (1.*)
+ doctrine/common      v2.6.1 requires doctrine/lexer (1.*)
 ```
 
-If you want, for example, find any installed package that is **not** allowing
-Symfony version 3 or one of its components, you can run the following command:
+If you want, for example, to find any installed package that is **not**
+allowing Monolog to be upgraded to version 1.17 , try this:
 
 ```sh
-php composer.phar depends symfony/symfony --with-replaces -im ^3.0
+php composer.phar depends monolog/monolog -im ^1.17
+There is no installed package depending on "monolog/monolog" in versions not matching 1.17
+```
+
+Add the `--tree` or `-t` flag to show a recursive tree of why the package is depended
+upon, for example:
+
+```sh
+php composer.phar depends psr/log -t
+psr/log 1.0.0 Common interface for logging libraries
+|- aboutyou/app-sdk 2.6.11 (requires psr/log 1.0.*)
+|  `- __root__ (requires aboutyou/app-sdk ^2.6)
+|- monolog/monolog 1.17.2 (requires psr/log ~1.0)
+|  `- laravel/framework v5.2.16 (requires monolog/monolog ~1.11)
+|     `- __root__ (requires laravel/framework ^5.2)
+`- symfony/symfony v3.0.2 (requires psr/log ~1.0)
+   `- __root__ (requires symfony/symfony ^3.0)
 ```
 
 ### Options
--- a/src/Composer/Command/BaseCommand.php
+++ b/src/Composer/Command/BaseCommand.php
@@ -18,7 +18,7 @@ use Composer\IO\IOInterface;
 use Composer\IO\NullIO;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Command\Command as BaseCommand;
+use Symfony\Component\Console\Command\Command;
 
 /**
  * Base class for Composer commands
@@ -26,7 +26,7 @@ use Symfony\Component\Console\Command\Command as BaseCommand;
  * @author Ryan Weaver <ryan@knplabs.com>
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-abstract class Command extends BaseCommand
+abstract class BaseCommand extends Command
 {
     /**
      * @var Composer
--- a/src/Composer/Command/BaseDependencyCommand.php
+++ b/src/Composer/Command/BaseDependencyCommand.php
@@ -1,0 +1,188 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Command;
+
+use Composer\DependencyResolver\Pool;
+use Composer\Package\Link;
+use Composer\Package\PackageInterface;
+use Composer\Repository\ArrayRepository;
+use Composer\Repository\CompositeRepository;
+use Composer\Repository\PlatformRepository;
+use Composer\Plugin\CommandEvent;
+use Composer\Plugin\PluginEvents;
+use Composer\Semver\VersionParser;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Base implementation for commands mapping dependency relationships.
+ *
+ * @author Niels Keurentjes <niels.keurentjes@omines.com>
+ */
+class BaseDependencyCommand extends BaseCommand
+{
+    const ARGUMENT_PACKAGE = 'package';
+    const ARGUMENT_CONSTRAINT = 'constraint';
+    const OPTION_RECURSIVE = 'recursive';
+    const OPTION_TREE = 'tree';
+
+    /**
+     * Set common options and arguments.
+     */
+    protected function configure()
+    {
+        $this->setDefinition(array(
+            new InputArgument(self::ARGUMENT_PACKAGE, InputArgument::REQUIRED, 'Package to inspect'),
+            new InputArgument(self::ARGUMENT_CONSTRAINT, InputArgument::OPTIONAL, 'Optional version constraint', '*'),
+            new InputOption(self::OPTION_RECURSIVE, 'r', InputOption::VALUE_NONE, 'Recursively resolves up to the root package'),
+            new InputOption(self::OPTION_TREE, 't', InputOption::VALUE_NONE, 'Prints the results as a nested tree'),
+        ));
+    }
+
+    /**
+     * Execute the command.
+     *
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     * @param bool @inverted Whether
+     */
+    protected function doExecute(InputInterface $input, OutputInterface $output, $inverted = false)
+    {
+        // Emit command event on startup
+        $composer = $this->getComposer();
+        $commandEvent = new CommandEvent(PluginEvents::COMMAND, $this->getName(), $input, $output);
+        $composer->getEventDispatcher()->dispatch($commandEvent->getName(), $commandEvent);
+
+        // Prepare repositories and set up a pool
+        $platformOverrides = $composer->getConfig()->get('platform') ?: array();
+        $repository = new CompositeRepository(array(
+            new ArrayRepository(array($composer->getPackage())),
+            $composer->getRepositoryManager()->getLocalRepository(),
+            new PlatformRepository(array(), $platformOverrides),
+        ));
+        $pool = new Pool();
+        $pool->addRepository($repository);
+
+        // Parse package name and constraint
+        list($needle, $textConstraint) = array_pad(explode(':', $input->getArgument(self::ARGUMENT_PACKAGE)),
+                    2, $input->getArgument(self::ARGUMENT_CONSTRAINT));
+
+        // Find packages that are or provide the requested package first
+        $packages = $pool->whatProvides($needle);
+        if (empty($packages)) {
+            throw new \InvalidArgumentException(sprintf('Could not find package "%s" in your project', $needle));
+        }
+
+        // Include replaced packages for inverted lookups as they are then the actual starting point to consider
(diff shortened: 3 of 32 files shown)


src/Composer/Command/DependsCommand.php
```diff
@@ -61,80 +63,153 @@ protected function configure()
...
-                    return $link->getTarget();
-                }, $package->getReplaces()));
+    /**
+     * @param string $needle The package to inspect.
+     * @param ConstraintInterface|null $constraint Optional constraint to filter by.
+     * @param bool $invert Whether to invert matches on the previous constraint.
+     * @param bool $recurse Whether to recursively expand the requirement tree.
+     * @return array An array with dependers as key, and as values an array containing the source package and the link respectively
+     */
+    private function getDependers($needle, $constraint = null, $invert = false, $recurse = true)
```
Seldaek: getDependents?

src/Composer/Command/DependsCommand.php
```diff
@@ -32,22 +35,21 @@
  */
 class DependsCommand extends Command
 {
-    protected $linkTypes = array(
-        'require' => array('requires', 'requires'),
-        'require-dev' => array('devRequires', 'requires (dev)'),
-    );
+    /** @var CompositeRepository  */
+    private $repository;
```
Seldaek: Not a fan of adding state like this, as it makes all methods depend on the instance while they could mostly be static methods, all you need is to pass $repo to getDependers AFAICT. It makes it easier to extract code into an external class (which would be good to make some of this testable eventually) when you have less object state to care about.

curry684: Agreed, it's kind of a weak solution to that it's currently not really better feasible. From a purist perspective it's currently correct, it's a property of the command, but that's only correct as the logic itself is as well. True issue is (but we all know that) that there should be a big cleanup session one day to remove all logic from the commands and only have a few oneliner calls and rendering.

doc/03-cli.md
```diff
@@ -350,12 +350,11 @@ php composer.phar depends symfony/symfony --with-replaces -im ^3.0
```
Seldaek: Could you also update the use case above with symfony lala? Because it uses stuff you removed, if you could update it to do the equivalent using the new flags that'd be good, because this was a good advanced use case IMO.

curry684: Oops didn't scroll back far enough hehe. Coming up.

curry684: Had to change the samples a bit as they were no longer reproducable, but think it's fine now.

Seldaek: No longer repro because you removed the --with-replaces? I'm just trying to understand the change as the PR is pretty hard to review. --with-replaces was pretty nice to see anything that blocks symfony/symfony from being upgraded due to a require on symfony/symfony *or* anything it replaces.

src/Composer/Command/DependsCommand.php
```diff
@@ -61,80 +60,153 @@ protected function configure()
...
+        $results = array();
+
+        /**
+         * Loop over all currently installed packages.
+         */
+        foreach ($packages as $package) {
+            // Retrieve all requirements, but dev only for the root package
+            $links = $package->getRequires();
+            $links += $package->getReplaces();
+            if ($package instanceof RootPackage) {
```
stof: you should use the interface actually

src/Composer/Command/DependsCommand.php
```diff
@@ -61,80 +60,153 @@ protected function configure()
...
+        $needles = is_array($needle) ? $needle : array($needle);
+        $results = array();
+
+        /**
+         * Loop over all currently installed packages.
+         */
+        foreach ($packages as $package) {
+            // Retrieve all requirements, but dev only for the root package
+            $links = $package->getRequires();
+            $links += $package->getReplaces();
```
stof: Why are you merging replacements and requirements ? These are totally different so it looks weird to treat them the same way

src/Composer/Command/DependsCommand.php
```diff
@@ -61,80 +60,153 @@ protected function configure()
...
+     * @param string $needle The package to inspect.
+     * @param PackageInterface[] $packages List of installed packages.
+     * @param ConstraintInterface|null $constraint Optional constraint to filter by.
+     * @param bool $invert Whether to invert matches on the previous constraint.
+     * @param bool $recurse Whether to recursively expand the requirement tree.
+     * @return array An array with dependers as key, and as values an array containing the source package and the link respectively
+     */
+    private function getDependents($needle, $packages, $constraint = null, $invert = false, $recurse = true)
+    {
+        $needles = is_array($needle) ? $needle : array($needle);
```
stof: this can never be an array according to the phpdoc, so something is wrong

src/Composer/Command/DependsCommand.php
```diff
@@ -61,80 +60,153 @@ protected function configure()
...
+            $constraint = null;
+        }
+        $matchInvert = $input->getOption('invert-match-constraint');
+        $renderTree = $input->getOption('tree');
+        $recursive = $renderTree || $input->getOption('recursive');
+
+        // Resolve dependencies
+        $results = $this->getDependents($needle, $repository->getPackages(), $constraint, $matchInvert, $recursive);
+        if (empty($results)) {
+            $extra = isset($constraint) ? sprintf(' in versions %smatching %s', $matchInvert ? 'not ' : '', $textConstraint) : '';
```
stof: please use ``null !== $constraint`` to check for ``null`` instead of ``isset``. We don't want to allow undefined variables here

src/Composer/Command/DependsCommand.php
```diff
@@ -61,80 +60,153 @@ protected function configure()
...
+            $versionText = (strpos($package->getPrettyVersion(), 'No version set') === 0) ? '' : $package->getPrettyVersion();
+            $packageText = rtrim(sprintf('%s %s', $package->getPrettyName(), $versionText));
+            $linkText = implode(' ', array($link->getDescription(), $link->getTarget(), $link->getPrettyConstraint()));
+            $output->write(sprintf("%s%s %s (%s)\n", $prefix, $isLast ? '`-' : '|-', $packageText, $linkText));
+            $this->printTree($output, $children, $prefix . ($isLast ? '   ' : '|  '));
+        }
+    }
 
-        $needles = array($needle);
-        if (true === $input->getOption('with-replaces')) {
```
stof: This logic adding packages replacing the needle in the list of needles too should be preserved to keep the existing features (and I suggest keeping the option)

stof: note that this array of needles is what you should pass to ``getDependants`` in your new system.

curry684:
I realize now what was bugging me about the `--with-replaces` option: the scenarios I tried only gave nonsensical results, but that was because it was overeager.

Say I have 2 packages `foo/bar` and `bar/foo` depending on `symfony/console' and `symfony/yaml` respectively, but I have installed `symfony/symfony` which replaces both. The old implementation would show the second dependency as well when querying the first, which is just plain wrong - it's an entirely different branch of the project's dependency true.

**However**, with the example in the docs:

    php composer.phar depends symfony/symfony --with-replaces -im ^3.0

Now it suddenly does make sense to expand into the replaced packages (so downwards, not up) as a dependency on `symfony/console:2.7` may very well be blocking installation of `symfony/symfony:3.0`.

I'll have to think a bit about whether that warrants having the option back, as I think it's well possible to implement default behaviour which fits all cases. I'll get to fixing that later on - most of the remarks here are related to the leftovers of this, like the array parameter :thumbsup:

src/Composer/Command/DependsCommand.php
curry684: From the dependency standpoint they're the same. It's an upward link causing a package to be present: whether `symfony/console` is installed because `symfony/symfony` replaces it, or because it requires it, is functionally moot - in both cases `symfony/symfony` is the reason that `symfony/console` is installed. So yes, they are equivalently merged, with differentiation in the output.

doc/03-cli.md
curry684: Only the second example. I had to replace the `monolog/monolog` sample because most packages have it in `require-dev`, so the old docs were actually a good example of that part not correctly functioning.

I can get into that it's hard to review, it's a complex bit of nastiness with nasty exceptions. I'm actually just realizing that the `-im --with-replaces` functionality was broken too to detect blockers as it ignores `conflicts`.

curry684:
Ok, had a lot more thinking time on this, and I came to the conclusion the underlying issue is that the previous implementation of this command was invalidly trying to combine 2 completely different things. Looking at `aptitude`, they have separated this behavior in 2 commands: `why` and `why-not`. The first checks for problems preventing **removal** of a package, the second checks for problems preventing **installation** of a package.

The current implementation of `depends` tries to haphazardly combine them. But on its own the `-m` flag is pretty much pointless - `c depends symfony/symfony -m 3.5` tells me nothing that isn't plain obvious from `c depends symfony/symfony`.

The `-i` flag is documented everywhere to invert the behavior of `-m`, but actually is valid standalone and useful, as `-m` defaults to `*` - it will just show why no version whatsoever of that package will ever be installed, for example because of conflicts. Combined with `-m` it can be ultra powerful. But only, and that's the real catch, if it also implies the old behavior of `--with-replaces`. Which is useless **without** `-i` as it resolves the wrong way then.

So yes, I think the `--with-replaces` behavior should go back in, but not as its own option, but implicit when `-i` is also specified.

The syntax and use for this powerful command would probably have been a lot clearer and cleaner if it had been split to begin with:
```
composer [depends|why] <package> (<constraint>)
composer [prohibits|why-not] <package> (<constraint>)
```

Seldaek:
I don't mind breaking BC on such small args, especially since they're pretty new (AFAIK not even in a tagged release). Let's do it clean, having two commands sounds good, and will force us to move most of this code away from Command classes :)

curry684:
Ok, I'll get on that then :smile:

curry684:
Well @Seldaek good luck, did my best to make it harder to review  :wink:

src/Composer/Command/DependsCommand.php
```diff
@@ -59,82 +41,15 @@ protected function configure()
...
-            sort($messages);
-            $io->write($messages);
-        } else {
-            $matchText = '';
-            if ($input->getOption('match-constraint') !== '*') {
-                $matchText = ' in versions '.($matchInvert ? 'not ' : '').'matching ' . $input->getOption('match-constraint');
-            }
-            $io->writeError('<info>There is no installed package depending on "'.$needle.'"'.$matchText.'.</info>');
-        }
+        parent::doExecute($input, $output, false);
```
staabm: Either phpdoc is wrong or this line should return

src/Composer/Command/ProhibitsCommand.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Execute the function.
+     *
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     * @return int|null
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        parent::doExecute($input, $output, true);
```
staabm: Same here: misses return?

src/Composer/Repository/BaseRepository.php
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param string|string[] $needle The package name(s) to inspect.
+     * @param ConstraintInterface|null $constraint Optional constraint to filter by.
+     * @param bool $invert Whether to invert matches to discover reasons for the package *NOT* to be installed.
+     * @param bool $recurse Whether to recursively expand the requirement tree up to the root package.
+     * @return array An associative array of arrays as described above.
+     */
+    public function getDependents($needle, $constraint = null, $invert = false, $recurse = true)
+    {
+        $needles = is_array($needle) ? $needle : array($needle);
```
staabm: Could be simplified to `$needles = (array) $needle;`

src/Composer/Command/BaseDependencyCommand.php
```diff
@@ -0,0 +1,188 @@
...
+            new InputOption(self::OPTION_TREE, 't', InputOption::VALUE_NONE, 'Prints the results as a nested tree'),
+        ));
+    }
+
+    /**
+     * Execute the command.
+     *
+     * @param InputInterface $input
+     * @param OutputInterface $output
+     * @param bool @inverted Whether
```
staabm: Doc incomplete

src/Composer/Command/DependsCommand.php
curry684: The ancestor implementation defines the return value, so IDE complains about invalid phpdoc if it isn't inherited, but returning null is implicit, and as that is allowed by the phpdoc this construct is valid. The other commands were already doing the same actually.

src/Composer/Repository/BaseRepository.php
curry684: Funny, had to look it up but that is indeed defined behaviour now, thought it was still semi undefined as automatic conversion still is :+1:

src/Composer/Console/Application.php
```diff
@@ -279,33 +279,35 @@ public function getHelp()
...
+            new Command\StatusCommand(),
+            new Command\ArchiveCommand(),
+            new Command\DiagnoseCommand(),
+            new Command\RunScriptCommand(),
+            new Command\LicensesCommand(),
+            new Command\GlobalCommand(),
+            new Command\ClearCacheCommand(),
+            new Command\RemoveCommand(),
+            new Command\HomeCommand(),
+        ));
```
Seldaek: Not sure what this change is good for? Also completely not related to the PR, which makes reviewing even harder.

src/Composer/Command/BaseCommand.php
```diff
@@ -26,7 +26,7 @@
  * @author Ryan Weaver <ryan@knplabs.com>
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-abstract class Command extends BaseCommand
+abstract class BaseCommand extends Command
```
Seldaek: Also don't see the point of renaming this. Lots of changes for nothing, and especially not relevant to the PR.

curry684: I needed a common ancestor for Depends and Prohibits. Could've just called it `DependencyCommand`, but that would've started obfuscation hell imo as there would be no visible distinction in the `Composer/Command` namespace between implementations and base classes. So as other namespace were already following the `BaseX` convention, like `BasePackage`, and we agreed here to start cleaning up the commands anyway, I refactored it into consistency, and now the Command namespace has a clear distinction between Base classes, both `abstract`, and other classes.

src/Composer/Console/Application.php
curry684: Unrelated, true, just something I hit while adding the `ProhibitsCommand`. Would appear this once started as 3 single array inserts and the next 20 commands were just copy/pasted without considering for that many a single bulk merge makes more sense than 23 single inserts.

Will happily revert it if you disagree  :)

Seldaek:
Thanks! Merged and did a couple minor fixes in d7855d03c7fd9a7afdc5f8d9c68cd74e8611cc60. If you do get a chance though once thing that'd be nice is to merge the ShowCommand tree display with this tree display so they get the same coloring treatment, because right now this one is kinda hard to read w/o color compared to the show -t one :)

Just one thing that trips me up, if you use `prohibits` without specifying a version number it basically acts like `depends`, right? I think it should error out because just listing what requires the package really doesn't tell you why the package can't be installed. If it's a meaningless use case I think we should at the very least warn.

curry684:
1. I'm considering submitting a decent tree renderer to symfony/console as it already has the table helper. That would allow us to replace both, as indeed they are different now here. I would probably dev it here to standardize behavior of both then submit a PR there. I looked into reusing the Show renderer but it's messily integrated and hard to extract right now.
2. Prohibits right now should require a version indeed. Thing is, without version it should do a full solution, and right now it naively acts on installed packages like before the -i option was pretty useless without -m. So it's not worse than before but has definite room for improvement. It's something we should discuss in a separate issue or elsewhere first as it has pretty complex consequences.

curry684:
Your fixes are :+1: btw

PR #4920, opened 2016-02-15
Add a --timeout parameter to run-script to override default timeout

### LINKED ISSUE #4918: Composer scripts are subject to the process-timeout config (default 300s)
Composer scripts are subject to the process-timeout config.

While it can be useful to have a process timeout on `composer install`. I think it shouldn't be applied to the custom scripts since they can be long-running scripts.

Error:
`COMPOSER_PROCESS_TIMEOUT=1 composer run api-server -vvv`
```
[Symfony\Component\Process\Exception\ProcessTimedOutException]                                           
  The process "#" exceeded the timeout of 1 seconds.
```

curry684:
Fixes #4918

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -562,6 +562,7 @@ Lists the name, version and license of every package installed. Use
 
 ### Options
 
+* **--timeout:** Set the script timeout in seconds, or 0 for no timeout.
 * **--no-dev:** Disable dev mode
 * **--list:** List user defined scripts
 
--- a/src/Composer/Command/RunScriptCommand.php
+++ b/src/Composer/Command/RunScriptCommand.php
@@ -14,6 +14,7 @@ namespace Composer\Command;
 
 use Composer\Script\CommandEvent;
 use Composer\Script\ScriptEvents;
+use Composer\Util\ProcessExecutor;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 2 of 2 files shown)


Seldaek:
We already have https://getcomposer.org/doc/03-cli.md#composer-process-timeout - isn't that enough? It works for all commands at least.. cc @grifx

curry684:
He wants to run an API server as part of this package, that implies an infinite timeout. I don't think the global timeout should have to be disabled for a single exception, be it for an infinite script or one that could run longer than the default 300 in special cases.

This PR adds the option to ignore the global default on a per instance basis (that's why it is set after the `new Composer` call.

Seldaek:
Aye fair enough. Doesn't hurt much and it's more self-documenting so I'll take it in, but the env var also lets you disable it for a single command you run which really is about the same.

curry684:
Good point actually that the documentation doesn't mention that `COMPOSER_PROCESS_TIMEOUT=0` disables it - I found that out by experiment too.

igorsantos07:
Wondering if it would be possible to pass that to script shortcuts? i.e. `composer server --timeout=0` instead of `composer run server --timeout=0`. Currently it spits out an "option does not exist" error.

Antnee:
I'm just looking at this myself. We run everything inside Docker, and our dev environments have two dependencies; Docker and Composer. For example, we can run `composer serve` and the containers are spun up etc and everything works great.

However, if we want to have a long-running process, such as Webpack watching the files, we have a problem. We literally just enter `composer webpack` at the moment, but the script will eventually terminate. In order to get this to work, I need to enter `composer run webpack --timeout=0`, but there are no cases where I want this process to not run indefinitely.

I want this single command to have a custom timeout. This section of our scripts looks like this:

```json
    "node":"docker-compose run --rm -w /app node",
    "npm-install": "@node npm install",
    "npm-rebuild-sass": "@node npm rebuild node-sass",
    "webpack": "@node webpack --watch",
    "webpack-compile": "@node webpack"
  }
```

Is it possible to have the timeout specified in the command, somehow? Rather than needing to pass in the longer command (which is less than ideal) or updating the entire app's timeout (even less ideal)?

PR #4921, opened 2016-02-15
BC cleanup

curry684:
- Checked code marked as 'backwards compatible' and removed any that were from early 2013 or even before.
- Corrected a missing deprecation warning.
- Fixed a unit test depending on deprecated implementation
- Removed long deprecated methods that were no longer referenced (2+ years old)

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -236,13 +236,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
             return array_keys($this->providerListing);
         }
 
-        // BC handling for old providers-includes
-        $providers = array();
-        foreach (array_keys($this->providerListing) as $provider) {
-            $providers[] = substr($provider, 2, -5);
-        }
-
-        return $providers;
+        return array();
     }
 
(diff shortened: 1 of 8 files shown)


Seldaek:
Thanks

src/Composer/Repository/ComposerRepository.php
```diff
@@ -304,15 +298,7 @@ public function whatProvides(Pool $pool, $name)
...
-            // BC handling for old providers-includes
-            $url = 'p/'.$name.'.json';
-
-            // package does not exist in this repo
-            if (!isset($this->providerListing[$url])) {
-                return array();
-            }
-            $hash = $this->providerListing[$url]['sha256'];
-            $cacheKey = null;
+            return array();
```
stof: should it return an empty array or throw an exception saying it does not support this kind of repositories ?

curry684: All repository parsing code appears to take the stance that unsupported properties are silently ignored, and repository data is extracted from properties it does know how to parse. Therefore I opted to revert these to the situation as if that property had never existed.

This specific code was BC'd in March 2013 I believe, so I think that's the best approach for this.

PR #4924, opened 2016-02-16
Wrap final attempt to touch a cache file in Silencer.

### LINKED ISSUE #4070: touch(): Utime failed: Permission denied
After commit 051b7bd44b52db04b4b5e54db99436b834a35c21 it is no longer possible to have a shared composer cache, as the user running composer has to own the file in order to run utime which is done by touch($this->root . $file, filemtime($this->root . $file), time());

It would be desirable to only try to run utime if the user owns the file.

curry684:
Permission issues in `touch` throw an E_WARNING error. Worst consequence I can see of ignoring it is extra cache misses from early GC, which is better than current 'immediate death', so wrapped it with Silencer.

Should fix #4070

### DIFF
--- a/src/Composer/Cache.php
+++ b/src/Composer/Cache.php
@@ -143,7 +143,7 @@ class Cache
             } catch (\ErrorException $e) {
                 // fallback in case the above failed due to incorrect ownership
                 // see https://github.com/composer/composer/issues/4070
-                touch($this->root . $file);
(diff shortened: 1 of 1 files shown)


Seldaek:
Was about to do it but thanks ;)

curry684:
Explains why you merged it within a minute hehe.

PR #4930, opened 2016-02-17
Regression when Git package has two or more URLs and only the last one works.

magnusnordlander:
Fixed (and added a test for) a regression introduced in a77e2fb, which causes package Git package updates to fail when the package has two or more URLs and the last URL is the only one that actually works.

Currently, this causes a RuntimeException to be thrown, even if, when trying the last URL, the update was successful.

Also, please look closely at the test. It passes, but as I am not familiar enough with which git commands composer *should* be executing, I'd like a second pair of eyeballs on it.


Seldaek:
Fix seems legit and the test passes so I'll say all good but if someone has time to review the test in more depth that'd be great!

Issue #4636, opened 2015-11-25
Path Repository Type: Add option / flag to force mirroring instead of symlinking


DanielSchwiperich:
There are scenarios where symlinking is not appropriate / best solution.
It would be nice to be able to set a flag to force mirroring:

```
{
    "repositories": [
        {
            "type": "path",
            "url": "../../packages/my-package",
            "symlink": false
        }
    ],
    "require": {
        "my/package": "*"
    }
}
```

If we take the following example project (vcs) structure: 
```
-  composer.json
- composer.lock
- build.xml
- Dockerfile
- ...
- packages
\_ my-package1
  \_ composer.json
\_ my-package2
  \_ composer.json
```

A project for a portal. Using a cms and a forum software.
The main composer.json has dependencies on Contao Standard Edition for example and phpbb

We now want to build an installable package:
- Run `composer install` to get contao and phpbb
- Create a build folder and put contao and phpbb from the vendor dir in it
- Now we add our custom packages / bundles to it and do the following in the build folder on the contao composer.json: 
 - composer config repositories.cms-bundles path ../packages/*
 - composer require  package1 package2
 - ...

At the end we want to deploy the build folder on a Docker Instance / Server or whatever.
The problem then is, that the symlinks are broken and the path packages missing.

We can get around if we zip the custom packages during build process and use the `artifcat` type. 
Sure we could also put all packages in their own git repos, but then we would have an almost empty main repos only containing meta data and multiple very small extension repositories also resulting in a lot of administration overhead regarding CI.

alcohol:
If you don't want to symlink, then just use the vcs or git type.

DanielSchwiperich:
vcs type does not work on sub directories. The above example is already a git repository containing the subpackages. 

From the `build` - dir perspective
```
 "repositories": {
        "0": {
            "type": "path",
            "url": "path/to/repos/packages/my-package1"
        },
    },
```
This works, but creates a symlink. Splitting of the build dir for deployment will break the symlink

```
 "repositories": {
        "0": {
            "type": "vcs",
            "url": "/path/to/repos/packages/my-package1"
        },
    },
```
does not work because `/path/to/repos` is the git repository and only the `/path/to/repos/composer.json` can be accessed but not  `/path/to/repos/packages/my-package-1/composer.json`

alcohol:
If they are already subdirs of your root package, why even use a path repository? Just add the proper autoload definitions into your root composer.json for those directories.

[...]

Issue #4895, opened 2016-02-09
cleanup inherited dependencies when removing a root dependency


djmattyg007:
If I install a package that requires various other packages, and some of those packages require other packages, etc etc, and then run 'composer remove' to remove the original package I installed, none of its dependencies (that I never specifically required) are removed. There doesn't appear to be any way of keeping track of which packages were installed automatically as dependencies, nor does there appear to be any way of automatically cleaning these up.

Various Linux distro package management tools such as [apt-get](https://help.ubuntu.com/community/AptGet/Howto#Removal_commands) and [pacman](https://wiki.archlinux.org/index.php/Pacman/Tips_and_tricks#Orphans) do support this kind of functionality if you're looking for example

djmattyg007:
Even worse, I just discovered that I can't use `composer remove` to remove these dependencies at all. It looks like I have to manually edit the composer.lock file or first add them to the `composer.json` file before I can remove them with `composer remove`.

curry684:
This is a known shortcoming indeed. Easiest way to fix it is `rm -rf vendor && composer install`.

alcohol:
Did you run `composer remove --update-with-dependencies`? Or just `composer remove`?

;-)

curry684:
Why is that not default actually?

alcohol:
Because then someone else will ask, "why is it default?", and then we get into an endless discussion about who has the biggest penis, and nobody wins in the end. Because, internet :d

curry684:
I would definitely win that discussion.

*runs* :grin:

Seldaek:
@curry684 it's not default because it's generally not what you meant to do, and in many cases ends up updating way more than what people expect. The option is there though because sometimes without it you are stuck.

curry684:
Not to get into the penis discussion, but I think that may have some unfortunate unintended by-effects. For example I include `composer/composer`. I use the `Symfony\Console

curry684:
Not to get into the penis discussion, but I think that may have some unfortunate unintended by-effects. For example I include `composer/composer`. I then use the `Symfony\Console` classes in my code. I remove `composer/composer`, but it's dependency on `symfony\console` is not removed, so my IDE continues to approve and autocomplete all my code. At runtime stuff starts to crash though because the autoloader is missing, confusion arises.

I don't really see the use case for leaving them around actually, as far as I can see it can only cause confusion like illustrated above. At the very least I think a message would be helpful like:

```
The following packages are no longer included because they are not referenced
[...]

PR #4933, opened 2016-02-19
Introduce option to force copy instead of symlinking in PathDownloader

burci:
This pull request implements the ability to force PathDownloader to use copy instead of symlink, refs:  #4636

New config option:
```json
{
    "repositories": [
        {
            "type": "path",
            "url": "../../packages/my-package",
            "options": {
                "symlink": false
            }
        }
    ]
}
```

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -649,6 +649,26 @@ the console will read `Symlinked from ../../packages/my-package`. If symlinking
 is _not_ possible the package will be copied. In that case, the console will
 output `Mirrored from ../../packages/my-package`.
 
+Instead of default fallback strategy you can force to use symlink with `"symlink": true` or
+mirroring with `"symlink": false` option.
+Forcing mirroring can be useful when deploying or generating package from a monolithic repository.
+
+```json
+{
+    "repositories": [
+        {
+            "type": "path",
+            "url": "../../packages/my-package",
+            "options": {
+                "symlink": false
+            }
+        }
+    ]
+}
+```
+
+
+
 Instead of using a relative path, an absolute path can also be used.
 
 > **Note:** Repository paths can also contain wildcards like ``*`` and ``?``.
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -25,6 +25,9 @@ use Symfony\Component\Filesystem\Filesystem;
  */
 class PathDownloader extends FileDownloader
 {
+    const STRATEGY_SYMLINK = 10;
+    const STRATEGY_MIRROR  = 20;
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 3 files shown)


doc/05-repositories.md
```diff
@@ -649,6 +649,22 @@ the console will read `Symlinked from ../../packages/my-package`. If symlinking
 is _not_ possible the package will be copied. In that case, the console will
 output `Mirrored from ../../packages/my-package`.
 
+You can force to use copy instead of the default symlinking logic, with:
```
staabm: you should mention a example why one needs that

src/Composer/Downloader/PathDownloader.php
```diff
@@ -54,18 +57,24 @@ public function download(PackageInterface $package, $path)
...
+                    // Implement symlinks as NTFS junctions on Windows
+                    $this->filesystem->junction($realUrl, $path);
+                    $this->io->writeError(sprintf('    Junctioned from %s', $url));
+                } else {
+                    $shortestPath = $this->filesystem->findShortestPath($path, $realUrl);
+                    $fileSystem->symlink($shortestPath, $path);
+                    $this->io->writeError(sprintf('    Symlinked from %s', $url));
+                }
+            } catch (IOException $e) {
+                $symlink = false;
```
staabm: please add a writeError status line so it gets obvious when things fallback because of symlink problems

burci:
Both example and writeError added.

doc/05-repositories.md
```diff
@@ -649,6 +649,25 @@ the console will read `Symlinked from ../../packages/my-package`. If symlinking
 is _not_ possible the package will be copied. In that case, the console will
 output `Mirrored from ../../packages/my-package`.
 
+You can force to use mirroring instead of the default symlinking logic with `"symlink": false` option.
+This can be useful when deploying or generating package for a monolithic repository.
```
curry684: This is grammatically messy. If you need help let me know.

burci: I would appreciate if you could.

curry684:
You are implementing a two-state option now, but isn't it more useful to make it tri-state?

- **true**: always symlink, otherwise fail
- **null (default) **: no preference, try symlink, otherwise mirror
- **false**: always mirror

Especially when using it for package development it makes sense to actually *require* symlinking/junctioning, like for the monoliths your use case is valid. The default should just stick to the old behavior: try to save disk space, but otherwise just deep copy.

dzuelke:
I LOVE IT SHIP IT WHEEEEE :+1: :rocket: /cc @Seldaek

burci:
I still have one problem to solve.

My scenario: I user monolithic repository. While develop using distributed package with symlink, on deploy i add "symlink": false option before composer. 

But:
```
{
    "repositories": [
        {
            "type": "path",
            "url": "../../packages/my-package"
            "options": {
                "symlink": false
            }
        }
    ],
    "require": {
        "my/package": "*"
    }
}
```

Because of "*" an Composer\Semver\Constraint\EmptyConstraint will be created and while using mirroring it is not touching the previous install even there was changes.

@Seldaek any idea, best practice what should i change to work with mirroring as expected?

It would be nice to check the symlink existence as well while using symlink.

curry684:
This is not feasible currently. The installer prefers existing installations if the constraint is still satisfied, so in this scenario it will always keep the old installed version. I don't see a way around this right now other than aliasing `rm -rf vendor composer.lock && composer install`.

burci:
But because of "*" and Composer\Semver\Constraint\EmptyConstraint even ```composer update``` will not update the package while using mirroring. (new version is parsed out right way, in ./vendor/composer/installed.json there is the old version)

curry684:
This is similar to what I described in https://github.com/composer/composer/issues/4895#issuecomment-184895601 - existing installations already matching constraints are preferred over potentially destructive actions. I'm not sure whether this is a side-effect of how the PathRepository works - it may just ignore the whole concept of updating at all because it expects symlinks, and does not have the 'branching' concept to detect version changes.

Seldaek:
@burci that problem of updates is unrelated to this PR though so please don't mingle that in. Let's open a new one if this needs to be addressed. Or a new issue for discussion.

I think https://github.com/composer/composer/pull/4933#issuecomment-186209850 makes sense so it would be good if you have time to implement that.

burci:
Implemented https://github.com/composer/composer/pull/4933#issuecomment-186209850

Seldaek:
Thanks @burci

dzuelke:
:rocket:

PR #4938, opened 2016-02-21
Add the --root-require option to the update command

### LINKED ISSUE #4729: update --root option for only updating root packages
It would be really helpful if there was a `--root` option for `composer update` that would allow you to just update the packages in the root composer.json file. Using `--root` with `--with-dependencies` is the same as using `composer update` without those options.

The reason I'm asking for this, is that it's difficult to determine which root packages need updating (and therefor need more testing as they are direct dependencies) vs. which are just dependencies of root packages. The only way you can do this right now is by manually comparing the output you get with `composer update` with what you have in your `composer.json` file.

davidbarratt:
This issue resolves #4729. I didn't see any tests for the update command, but if you'd like a test for this, let me know where to put it.

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -103,18 +103,17 @@ EOT
         }
 
         if ($input->getOption('root-require')) {
-          $require = array_keys($composer->getPackage()->getRequires());
-          if (!$input->getOption('no-dev')) {
-            $require_dev = array_keys($composer->getPackage()->getDevRequires());
-            $require = array_merge($require, $require_dev);
-          }
+            $require = array_keys($composer->getPackage()->getRequires());
+            if (!$input->getOption('no-dev')) {
+                $require_dev = array_keys($composer->getPackage()->getDevRequires());
+                $require = array_merge($require, $require_dev);
+            }
 
-          if (!empty($packages)) {
-            $packages = array_intersect($packages, $require);
-          }
-          else {
-            $packages = $require;
-          }
+            if (!empty($packages)) {
+                $packages = array_intersect($packages, $require);
+            } else {
+                $packages = $require;
+            }
         }
 
         $composer->getDownloadManager()->setOutputProgress(!$input->getOption('no-progress'));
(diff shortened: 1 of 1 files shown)


curry684:
Logic seems fine to me, so GTG :+1:

davidbarratt:
Do I need to worry about the Windows test failure? I it looks unrelated.

src/Composer/Command/UpdateCommand.php
```diff
@@ -101,6 +102,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $packages = $this->getPackagesInteractively($io, $input, $output, $composer, $packages);
         }
 
+        if ($input->getOption('root-require')) {
+            $require = array_keys($composer->getPackage()->getRequires());
+            if (!$input->getOption('no-dev')) {
+                $require_dev = array_keys($composer->getPackage()->getDevRequires());
```
stof: variable names should be camelCased

stof:
This would deserve a functional test IMO

davidbarratt:
@stof,

I don't see any functional tests for the `update` command, unless I'm missing something

src/Composer/Command/UpdateCommand.php
davidbarratt: Fixed.

curry684:
> Do I need to worry about the Windows test failure? I it looks unrelated.

They fail 95+% of the time  :wink:

stof:
> They fail 95+% of the time :wink:

it would still be great to determine why they fail, to fix this. Otherwise the CI server becomes useless as people will stop looking at failures and just assume it is OK for it to be red. But this is indeed out of the scope of this PR.

curry684:
I think it's just fundamentally broken, in https://github.com/composer/composer/pulls right now the top half is either red-crossed because of Appveyor failing without proper reason, or yellow as waiting for days already for Appveyor to complete. It would be helpful to have it removed until repaired on a fork.

stof:
functional tests are handled in https://github.com/composer/composer/tree/master/tests/Composer/Test/Fixtures/functional, but the setup indeed does not support putting a project in a non-empty starting state currently (unlike the installer tests which are handled in the sibling folder, but don't run the real command code)

Seldaek:
Thanks, I renamed it to --root-reqs for consistency with --ignore-platform-reqs though :)

Issue #4884, opened 2016-02-05
Drupal can't use composer install because it requires a Github Auth token on installation


klausi:
See also https://www.drupal.org/node/1475510

Problem: when running composer install the user is prompted to input a Github auth token. That is a serious burden for 3000+ Drupal core contributors that will need to run composer install and a significant waste of developer time. Some users do not want to create a Github account.

It seems that composer only prompts for a Github auth token in interactive shells? I have never seen that prompt on Travis CI for PHP projects for example.

Proposed solution: make composer install on interactive shells behave the same as in non interactive ones: always fall back to install from source (git clone) and never prompt the user for a Github token. Users should be able to opt-in for faster downloads with Github auth tokens, but this should never be an interactive usability WTF.

stof:
It only asks for it when reaching the API rate limit, as authenticated calls have a much higher rate limit.
Never asking for it would make the users unaware they should authenticate against github to be able to keep using downloads.

A better solution would be to improve the prompt to better explain why we ask for it, and to allow to skip the authentication by giving an empty token (i.e. just hitting enter), which would also be explained in the message.

neclimdul:
@stof I think you might be on the right track. It would by default leave some people with a odd mix of dist and source checkouts without an explanation. Could we carry that token through all additional requests during the run and/or expose this behavior as an option?

catch56:
"It only asks for it when reaching the API rate limit"
Yes, that happens with every composer install of Drupal core because in total there are more than 60 downloads (phpunit etc. with --require-dev is quite a lot by itself)

"and to allow to skip the authentication by giving an empty token (i.e. just hitting enter), which would also be explained in the message."

It does that already, but it prompts on every individual download, so you get into a few seconds for a package, same error message again, press enter, over and over.

For me I'd show a warning message but not actually interrupt the downloads. composer install takes long enough that I'm not usually sitting holding my breath in front of the terminal window for it to finish. So I'd rather it took longer and actually happened, than go back and see it got stuck half way through. If it's still running, there's the warning message explaining why it's taking so long.

stof:
well, once you give it a token, it won't ask anymore.
what we could do to improve things is remembering that you rejecting giving a github token so that we don't ask again in the same composer run.

klausi:
We do not want to force composer users to create a github account. The install process can throw warnings with bells and whistles and Github advertisements, but it should get the job done without prompting or bothering the user.

alcohol:
[...]

PR #4944, opened 2016-02-22
Add direct github.com downld as alternative to api

barryvdh:
Potential fix for #4884 and alternative for #4737

When the Download hits the API limits, it tries to download from github.com directly, using the regex @nicolas-grekas provided in his PR.

This approach still tries the API call first and warns the users, to promote github authentication and use the API as much as possible. When this fails, it will fall back to the source like before. It just doesn't ask for OAuth tokens anymore, but just downloads from alternative source.

This should provide faster downloads for unregistered users instead of forcing them to signup on Github. Would probably also create faster CI builds and/or deployments, without having to setup authentication (which isn't always easy in open source projects).

Potential downside to this approach:
 - When the URLs change, it will clone instead of asking for credentials (but can be fixed with a new release shortly and is unlikely to change (Bower also uses these urls).
- This always calls the api.github.com before each request. The direct link could be used first when unauthorized, but as api.github.com is the 'official' way, it's perhaps cleaner to try it first. Alternatively a flag could be set once the rate limit is reached, to skip the API calls altogether (but waiting for feedback before doing more drastic changes).

As @nicolas-grekas showed, there are more then 1 way to do this, so perhaps there is a better way for this :)

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -41,6 +41,7 @@ class RemoteFilesystem
     private $degradedMode = false;
     private $redirects;
     private $maxRedirects = 20;
+    private $githubRateLimited = false;
 
     /**
      * Constructor.
@@ -228,6 +229,10 @@ class RemoteFilesystem
         $userlandFollow = isset($options['http']['follow_location']) && !$options['http']['follow_location'];
 
         $this->io->writeError((substr($fileUrl, 0, 4) === 'http' ? 'Downloading ' : 'Reading ') . $fileUrl, true, IOInterface::DEBUG);
+
+        if ($originUrl === 'github.com' && $this->githubRateLimited && $directUrl = $this->getDirectGithubUrl($fileUrl)) {
+            $fileUrl = $this->fileUrl = $directUrl;
+        }
 
         if (isset($options['github-token'])) {
             $fileUrl .= (false === strpos($fileUrl, '?') ? '?' : '&') . 'access_token='.$options['github-token'];
@@ -530,6 +535,16 @@ class RemoteFilesystem
     {
         if ($this->config && in_array($this->originUrl, $this->config->get('github-domains'), true)) {
             $message = "\n".'Could not fetch '.$this->fileUrl.', please create a GitHub OAuth token '.($httpStatus === 404 ? 'to access private repos' : 'to go over the API rate limit');
+
+            // Try to download using an github.com instead of api.github.com
+            if ($httpStatus !== 404 && $this->getDirectGithubUrl($this->fileUrl)) {
+                $this->io->writeError($message);
+                $this->io->writeError('    <warning>Now trying to download from github.com directly.</warning>');
+                $this->retry = true;
+                $this->githubRateLimited = true;
+                throw new TransportException('RETRY');
+            }
+
             $gitHubUtil = new GitHub($this->io, $this->config, null);
             if (!$gitHubUtil->authorizeOAuth($this->originUrl)
                 && (!$this->io->isInteractive() || !$gitHubUtil->authorizeOAuthInteractively($this->originUrl, $message))
@@ -1013,5 +1028,19 @@ class RemoteFilesystem
         $port = parse_url($url, PHP_URL_PORT) ?: $defaultPort;
 
         return parse_url($url, PHP_URL_HOST).':'.$port;
+    }
+
+    /**
+     * Converts an API download url to a direct GitHub archive url.
+     *
+     * @param  string                        $url The original api.github.com url
+     * @return string|null                   null when not valid
+     */
+    private function getDirectGithubUrl($url)
+    {
+        if (preg_match('#^https://api\.github\.com/repos/([^/]++/[^/]++)/(tar|zip)ball/(.++)$#', $url, $match)) {
+            $extension = $match[2] === 'zip' ? 'zip' : 'tar.gz';
+            return 'https://github.com/'.$match[1].'/archive/'.$match[3].'.'.$extension;
+        }
     }
 }
--- a/tests/Composer/Test/Util/RemoteFilesystemTest.php
+++ b/tests/Composer/Test/Util/RemoteFilesystemTest.php
@@ -217,4 +217,30 @@ class RemoteFilesystemTest extends \PHPUnit_Framework_TestCase
         $attr->setAccessible(true);
         $attr->setValue($object, $value);
     }
+
+    /**
+     * @dataProvider provideGithubArchiveUrls
+     */
+    public function testDirectGithubUrls($url, $expected)
+    {
+        $fs = new RemoteFilesystem($this->getMock('Composer\IO\IOInterface'));
+        $ref = new \ReflectionMethod($fs, 'getDirectGithubUrl');
+        $ref->setAccessible(true);
+
+        $this->assertEquals($expected, $ref->invoke($fs, $url));
+    }
+
+    public function provideGithubArchiveUrls()
+    {
+        return array(
+            array('https://api.github.com/repos/composer/composer/zipball/master', 'https://github.com/composer/composer/archive/master.zip'),
+            array('https://api.github.com/repos/composer/composer/tarball/master', 'https://github.com/composer/composer/archive/master.tar.gz'),
+            array('https://api.github.com/repos/composer/packagist/zipball/SOMESHA', 'https://github.com/composer/packagist/archive/SOMESHA.zip'),
+            array('https://api.github.com/repos/composer/packagist/zipball', null),
+            array('https://github.com/composer/composer/zipball/master', null),
+            array('https://www.github.com/composer/composer/tarball/master', null),
+            array('https://github.com/composer/composer/archive/master.zip', null),
+            array('https://github.com/composer/composer/archive/master.tar.gz', null),
+        );
+    }
 }


barryvdh:
Changed the method name a bit, add tar.gz support and added some tests to check the conversion.

Tests on my dev server, installing Laravel, with cleared cache and deleted vendor dir:

Authenticated with token: 37s
Clone from source: 2m30s
Direct download fallback: 51s

So that would indicate a rough 3x improvement for CI/deployments without tokens. And when they're cacheable, it drops down to 1 sec for both dist cases..

nicolas-grekas:
Personally I'm in favor of pragmatism. The public non-api URLs look fine as fallback. If github changes them, we should make the code resilient. I'd suggest doing this only in non interactive mode thought, so that real people still have an incentive to get through the API.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -531,6 +531,16 @@ protected function promptAuthAndRetry($httpStatus, $reason = null)
         if ($this->config && in_array($this->originUrl, $this->config->get('github-domains'), true)) {
             $message = "\n".'Could not fetch '.$this->fileUrl.', please create a GitHub OAuth token '.($httpStatus === 404 ? 'to access private repos' : 'to go over the API rate limit');
             $gitHubUtil = new GitHub($this->io, $this->config, null);
+
+            // Try to download using an github.com instead of api.github.com
+            if ($httpStatus !== 404 && $directUrl = $gitHubUtil->getDirectArchiveUrl($this->fileUrl)) {
+                $this->io->writeError($message);
+                $this->io->writeError('    <warning>Now trying to download from github.com directly.</warning>');
```
nicolas-grekas: once we reach the api limit (60/h), composer will hit github twice per download, the first one failing then this second one. We could be nicer to them.

barryvdh: This is actually what happens now already with clones. It keeps trying the API, but fails everytime (in non-interactive mode). I've pushed a new commit to keep track of the API limit, once it fails one time, it tries the direct url completely. This gives the user 1 warning and 1 extra call, but after that it's only using it directly (for that composer run at least)

barryvdh:
I've added a commit that sets a flag once a download fails because of the rate limit. This will prevent calls that will fail anyway. This shows a warning to the users, but does not trigger a prompt.
I guess we could still ask for a prompt in interactive mode, but not sure if there is an actual benefit for the user (as long as the direct links work).

This brings the non-authenticated calls on par with the authenticated calls (like 37 vs 39 sec or so)

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -531,6 +536,16 @@ protected function promptAuthAndRetry($httpStatus, $reason = null)
         if ($this->config && in_array($this->originUrl, $this->config->get('github-domains'), true)) {
             $message = "\n".'Could not fetch '.$this->fileUrl.', please create a GitHub OAuth token '.($httpStatus === 404 ? 'to access private repos' : 'to go over the API rate limit');
             $gitHubUtil = new GitHub($this->io, $this->config, null);
+
+            // Try to download using an github.com instead of api.github.com
+            if ($httpStatus !== 404 && $this->getDirectGithubUrl($this->fileUrl)) {
+                $this->io->writeError($message);
+                $this->io->writeError('    <warning>Now trying to download from github.com directly.</warning>');
```
staabm: the message should also reflect that this is caused by the github rate limit (and promote using tokens to uplift the limit)

barryvdh: Thats already in $message, see a few lines aboves this.

staabm: you are right, missed that.

barryvdh:
FYI, this should work with the new caching option, for container based Travis runs. https://docs.travis-ci.com/user/caching

https://blog.wyrihaximus.net/2015/07/composer-cache-on-travis/
```
## Run on container environment
sudo: false

## Cache composer bits
cache:
  directories:
    - $HOME/.composer/cache
```

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -1014,4 +1029,18 @@ private function getUrlAuthority($url)
...
+    /**
+     * Converts an API download url to a direct GitHub archive url.
+     *
+     * @param  string                        $url The original api.github.com url
+     * @return string|null                   null when not valid
+     */
+    private function getDirectGithubUrl($url)
+    {
+        if (preg_match('#^https://api\.github\.com/repos/([^/]++/[^/]++)/(tar|zip)ball/(.++)$#', $url, $match)) {
+            $extension = $match[2] == 'zip' ? 'zip' : 'tar.gz';
```
stof: this should use a strict comparison to match coding standards

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -531,6 +536,16 @@ protected function promptAuthAndRetry($httpStatus, $reason = null)
         if ($this->config && in_array($this->originUrl, $this->config->get('github-domains'), true)) {
             $message = "\n".'Could not fetch '.$this->fileUrl.', please create a GitHub OAuth token '.($httpStatus === 404 ? 'to access private repos' : 'to go over the API rate limit');
             $gitHubUtil = new GitHub($this->io, $this->config, null);
```
stof: this instantiation should be done after your new condition, as it is not needed in some cases

src/Composer/Util/RemoteFilesystem.php
barryvdh: Fixed

src/Composer/Util/RemoteFilesystem.php
barryvdh: Moved it.

barryvdh:
> @alcohol:  @barryvdh does your PR include the functionality suggested regarding not re-trying the direct download url for consecutive packages if it fails on the first one?

No it does not. This is however also not the case in the old behaviour, where Composer keeps trying the API calls.

alcohol:
I know. I'm not saying it was there already for other functionality and should hence be included here, but it is definitely a nice-to-have. So if you think you could add this, I would much appreciate it :-)

barryvdh:
Yeah but what's is the best way to test this? Because it could fail because of a network glitch, or perhaps 1 package is missing a build because of an incorrect reference or .. Or do we assume that once a github.com build fails, it's because of the changed location? Cause that would be a big edge case imho.

We would need a flag like `$this->githubDirectDownloadFailed = true` and then match the fileUrl to see if it's a github.com direct download (not API call), and then just skip to the clones directly..

alcohol:
You raise a valid concern. Would it be much trouble instead to display a warning when a direct download does fail also?

barryvdh:
It already shows an error, like the other failed downloads.

screen shot 2016-02-25 at 10 26 05

alcohol:
Alright then. Nothing further to add. :)

Seldaek:
I think I prefer the approach in #4737 because it avoids adding even more chaos in RemoteFilesystem :) The idea of trying API first is good though, and I think we could do that too over in the other PR.

I'll give it some more thought and try it out some other day but thanks for the work either way everyone.

barryvdh:
Sure, no problem :)

I did have a hard time figuring out what the best place was to hook into this, because the logic is kinda all over the place. Using the API first does have the benefit that it's more like Github wants it to be, and incentive for people to register. But if we can do that there also, great :)

barryvdh:
@Seldaek Did you think about this? Do you want me to try some different things?

Seldaek:
Nah I think between the two PRs were good but I've been busy with other
things and still have a huge backlog.

barryvdh:
OK cool :)

barryvdh:
We could also think about linking directly to the codeload downloads directly, which would prevent a redirect, for more speed. Although not really sure if either one is more reliable.

https://api.github.com/repos/composer/composer/zipball/master --> https://codeload.github.com/composer/composer/legacy.zip/master

https://github.com/composer/composer/archive/master.zip --> https://codeload.github.com/composer/composer/zip/master

piotrekkaminski:
+1 to merging this from Magento project. We depend heavily on Composer and hitting the prompt in interactive mode is a major issue for our less technical users (and just a pain point for the others)

barryvdh:
@Seldaek The most used dist urls are currently generated server-side on Packagist, right? So instead of rewriting with regexes, we could also add the direct download as preferred mirror (or direct link as main url and the API as non-preferred mirror) here: https://github.com/composer/composer/blob/fe134d3359ff6533f5462743cada35eabaf2350f/src/Composer/Repository/Vcs/GitHubDriver.php#L132-L136
And make sure the dist mirrors are also included when building the with Packagist (which currently is not the case? https://github.com/composer/packagist/blob/a07b9f715b301a414f7a12f166432df7e9b956e1/src/Packagist/WebBundle/Package/Updater.php#L229-L234 )

This way, when the url changes, we still have the fallback, and you can regenerate the package info on packagist, so only cached package data is affected (in which case the archive is probably also cached anyways).

(Or we could just skip the API url entirely, as suggested in the Drupal thread.)

Seldaek:
@barryvdh I prefer to do this at the composer level because it's more flexible. Rebuilding the whole packagist index would take ages. Please relax about this, we have two PRs, it'll get in somehow, I just don't have time *right now* :)

barryvdh:
Ok sorry, I'll shut up about this now ;)

barryvdh:
So according to the other issue, no need for this anymore :)

PR #4946, opened 2016-02-22
Added the --no-install option to the composer udpate

Toflar:
Hi everybody,

I've added the option to pass `--no-install` to `composer update` so it does solve the dependencies, writes the `composer.lock` file but does not effectively download all the packages.

Why?
Composer gets more and more attention and for us php developers it's a must-have for many years already. However, there are a lot of content management systems still providing their own plugin system with their own package/extension/module managers and they slowly start to migrate or find solutions on how to make use of Composer and use it for their own plugin systems.
Now, the thing is, a `composer update` uses a lot of memory and for us developers it's not a real issue because we execute that command locally anyway. However, I guess the majority of all these plugin managers of the respective CMS (or be it whatever software you want to) have been executed on the web server itself, right? The old way of doing stuff was uploading the php files to your server and then call some `install.php` or `config.php` do setup the system and load or install extensions/plugins to it.

These times have long gone for us developers. We use deployment scripts, CI servers and software etc. But for a lot of smaller software and rather regular users of those systems, things haven't changed much. They don't want to use the command line and they certainly don't want to install locally and deploy the data using scripts they have to configure.
So we still have to provide some kind of GUI and web installer which we can build on top of Composer. It's not an easy task but it is certainly feasible. However, I think we all agree that a `composer update` command requires a lot more RAM than you would normally expect a regular shared hosting to provide and it's not needed quite frankly because a `composer update` !== delivery of a regular website.
However, `composer update` is the only thing that really needs a lot of RAM. If you already have a `composer.lock` and just call `composer install` you obviously won't have any memory issues normally.

This leads me to this PR. It's the foundation for any cloud based solution (`config/platform` already exists to mock a given platform) or anything else that resolves the dependencies using `composer update` and delivers the `composer.lock` so you can later install the dependencies using `composer install` yourself on the target server (like any deployment tool does as well). Whether it's a cloud based solution or you execute the command on your local computer and upload the `composer.lock` manually later doesn't really matter. In these cases it's just useless to download all the packages because you simply don't need them :smile: 

It's my first PR for Composer, so please bear with me if I completely forgot about a use case. And please, if somebody could give me a hint on how I can integrate a unit test for this in the existing test suite, please feel free. I'm a bit lost in the mass of fixtures :smile:

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -52,6 +52,7 @@ class UpdateCommand extends Command
                 new InputOption('verbose', 'v|vv|vvv', InputOption::VALUE_NONE, 'Shows more details including new commits pulled in when updating packages.'),
                 new InputOption('optimize-autoloader', 'o', InputOption::VALUE_NONE, 'Optimize autoloader during autoloader dump.'),
                 new InputOption('classmap-authoritative', 'a', InputOption::VALUE_NONE, 'Autoload classes from the classmap only. Implicitly enables `--optimize-autoloader`.'),
+                new InputOption('no-install', null, InputOption::VALUE_NONE, 'Whether to skip installation of the package dependencies (only writes the composer.lock file).'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore platform requirements (php & ext- packages).'),
                 new InputOption('prefer-stable', null, InputOption::VALUE_NONE, 'Prefer stable versions of dependencies.'),
                 new InputOption('prefer-lowest', null, InputOption::VALUE_NONE, 'Prefer lowest versions of dependencies.'),
@@ -149,6 +150,7 @@ EOT
             ->setIgnorePlatformRequirements($input->getOption('ignore-platform-reqs'))
             ->setPreferStable($input->getOption('prefer-stable'))
             ->setPreferLowest($input->getOption('prefer-lowest'))
+            ->setNoInstall($input->getOption('no-install'))
         ;
 
         if ($input->getOption('no-plugins')) {
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -113,6 +113,7 @@ class Installer
     protected $ignorePlatformReqs = false;
     protected $preferStable = false;
     protected $preferLowest = false;
+    protected $noInstall = false;
     /**
      * Array of package names/globs flagged for update
      *
@@ -168,8 +169,12 @@ class Installer
         gc_collect_cycles();
         gc_disable();
 
-        if ($this->dryRun) {
-            $this->verbose = true;
+        if ($this->dryRun || $this->noInstall) {
+            if ($this->dryRun) {
+                $this->verbose = true;
+            } else {
+                $this->dumpAutoloader = false;
+            }
             $this->runScripts = false;
             $this->installationManager->addInstaller(new NoopInstaller);
             $this->mockLocalRepositories($this->repositoryManager);
@@ -219,13 +224,13 @@ class Installer
                 return $res;
             }
         } catch (\Exception $e) {
-            if (!$this->dryRun) {
+            if (!($this->dryRun || $this->noInstall)) {
                 $this->installationManager->notifyInstalls($this->io);
             }
 
             throw $e;
         }
-        if (!$this->dryRun) {
+        if (!($this->dryRun || $this->noInstall)) {
             $this->installationManager->notifyInstalls($this->io);
         }
 
@@ -1482,6 +1487,24 @@ class Installer
     public function setPreferLowest($preferLowest = true)
     {
         $this->preferLowest = (boolean) $preferLowest;
+
+        return $this;
+    }
+
+    /**
+     * Should packages be installed?
+     *
+     * Call this if you want Composer to update the composer.lock but not
+     * effectively download the packages from the respective sources.
+     * Useful if you want to call "composer install" later.
+     *
+     * @param boolean $noInstall
+     *
+     * @return Installer
+     */
+    public function setNoInstall($noInstall = true)
+    {
+        $this->noInstall = $noInstall;
 
         return $this;
     }


alcohol:
I'm sorry, but I really don't see the use-case here. Why would you want to update your lock file locally but not update your vendor directory?

Toflar:
Because you want to outsource the heavy load of dependency resolving. So on my web server I still want to use `composer install` because I can use all `scripts` that e.g. create symlinks. However, a `composer update` does not work there because of memory limits. I would thus like to outsource the `composer update` command onto a more powerful computer (be it locally or wherever) so I can then fetch (in case of an API) or upload (in case you do it locally) the `composer.lock` and then do a regular `composer install` from my web server. In the case of "composer dependency resolving as a service" you possibly don't want to download the vendor directory. It's just a completely useless load for GitHub and all the other package providers. Do you understand? :)

curry684:
What @alcohol is hinting at is that `composer update` is usually followed by rigid testing, both unit tests and manual, to ensure nothing is broken by the newly updated packages. What you are suggesting here is breaking sane development workflow and promoting see-no-evil prayer-based deployment.

Besides that, you should be aware that this implicitly 'nukes' your local install as it implies you voluntarily request discrepancies between the lockfile and the physically installed packages. I'm not sure we should want to have commands and options that explicitly boil down to bringing your project into a incoherent state.

alcohol:
Exactly what @curry684 said. I don't understand why you would want to locally update your lock file but not your actual dependencies, and then subsequently blindly upload this new lock file and run a `composer install` on **production** without actually knowing if things will work or explode in your face.

The normal flow of doing a `composer update` locally, verifying and then uploading + running `composer install` on production (while also not recommended and personally I frown upon it) is more than sufficient.

Toflar:
I don't really think you all got the main point of this :-) You keep thinking the way we developers do but that's not what this situation is all about. Obviously when we are writing our libraries and unit tests and have our dependencies then we all need all our vendors in a matrix from highest to lowest dependency and stuff. This is all just fine (and besides: as this is a simple option it doesn't change anything in your workflow anyway. You don't have to use it ;-)).

So let me try to give you some more background again:
I'm talking about package managers for end users. The guys that don't code. The guys that don't know about Composer and don't care about unit tests. Those guys that use our extensions to systems like Wordpress, Drupal and Co. All of those systems provide some sort of plugin system and all of those thus need to resolve dependencies. But what these guys want to do is click on an "install" button and the plugin system does the rest.

The "verification process" you talk about is very often not done on a localhost. A lot of people just have a second web instance where they install and update dependencies. Then they click around and check if everything works before they do the same on the production system or move the vendor directory after the update ;-)

It's about everything else than "sane development workflow and promoting see-no-evil prayer-based deployment" :-) It's a different use case here as we're talking about some plugin system.

alcohol:
> The "verification process" you talk about is very often not done on a localhost. A lot of people just have a second web instance where they install and update dependencies. Then they click around and check if everything works before they do the same on the production system or move the vendor directory after the update ;-)

I don't see how a `--no-install` option has any use for the above case? It reads to me like you first do an upgrade on your "test" system, and then run an install on your "production" system?

Toflar:
No, you do this:

1. Client: Instead of calling `composer update`, call the API with your `composer.json`, `installed.json` and your platform information (`config/platform` in `composer.json`).
2. Server: Resolve using the information from the client **without** actually downloading the vendors (`composer update --no-install`) and responding new `composer.lock`
3. Client: `composer install` with the new `composer.lock`
4. Client: Testing installation
5. Client: If everything's fine, deploy to `prod`.

Again, the whole point about this is to outsource the resolving to a more powerful host.

curry684:
But how is that not what @alcohol and I are saying? This boils down to screwing the client, who is an incompetent developer in your scenario, over with an untested combination of packages.

What happens if 5 fails? He's going to call you, and you'll just restart at step 1 until the client reports success?

Toflar:
I don't really get what you are trying to tell me, to be honest. It's not about my clients at all and nobody's calling me for anything of this? It's about the millions of users that install plugins to a CMS, a CRM or be it whatever software written in PHP you might think of.

I try it again: Every software provides a plugin system. Normally, you have your back end where you manage your installation and install/uninstall plugins, right? So quite frankly, why would anyone not want to build their plugin management system on top of Composer given the fact that it provides access to all PHP libraries out there and can handle dependencies perfectly?

We built a plugin manager (a GUI) on top of Composer and everything works. It installs all the dependencies just fine. The only thing that does not work is the dependency resolving because of resource limits. That's why I want to outsource it :-) I don't get how this is in any case related to breaking your system.

alcohol:
I think I understand his use-case. But it seems a bit overly complicated due to the amount of systems involved;

1. some CMS submits an API call to "dependency solver system" (providing relevant data, e.g. composer.json/installed.json/etc - like he mentioned),
2. dependency solver does a dry-run install and generates the resulting .lock file, sending it back as a result to the API call,
3. CMS now uses said lock file to do an install on a targetted test environment,
4. user verifies test environment, and if all is well...
5. CMS now uses said lock file to do an install on a targetted production environment.

Along those lines, or am I mistaken @Toflar ?

Toflar:
@alcohol You got it perfectly right. It resolves based on settings and platform information from the client and just sends back what the client should install then. Composer still runs on the client, it just does not do any `composer update`. It only executes `composer install` which obviously doesn't use that much memory :-)

Everything is already possible today, the dependency resolver just wastes resources because it doesn't need to download the packages :-) Basically, with my PR a `composer update` can be split into `composer update --no-install && composer install` :-)

curry684:
Ok, but does that warrant a command line option for everyone in the world? Isn't it more practical to include the composer/composer package in the solver system and execute this code there?

Toflar:
Of course I will do this, if it's not accepted. No problem at all :-)

Seldaek:
Given that what you're doing there requires building up a whole server infrastructure for the resolving of dependencies, I think requiring composer/composer as dependency and doing things there how you want it makes more sense than adding a flag for anyone to shoot themselves in the foot with :)

Toflar:
I agree :-) Thanks for the constructive feedback everyone!

PR #4951, opened 2016-02-23
RepositoryManager::prependRepository()

franzliedke:
This method is useful for dynamically adding repositories with higher priority than Packagist, e.g. from a Composer plugin.

Use case:
This would allow me to finally finish the new version of my [Studio plugin](https://github.com/franzliedke/studio) for Composer. It essentially allows installing project dependencies from a local path repository, without editing the composer.json file (as that is checked into Git, and only project maintainers may want such a setup). We have found this to be very useful over at Flarum: In our working copy, the flarum/flarum project requires the flarum/core project as well as several bundled extensions (all installed with Composer), and developing them from within the vendor folder is no fun. Thus, this solution was born.

@alcohol You may remember me asking questions about this stuff at the #composer-dev channel some time before Christmas.

### DIFF
--- a/src/Composer/Repository/RepositoryManager.php
+++ b/src/Composer/Repository/RepositoryManager.php
@@ -90,6 +90,18 @@ class RepositoryManager
     }
 
     /**
+     * Adds a repository to the beginning of the chain
+     *
+     * This is useful when injecting additional repositories that should trump Packagist, e.g. from a plugin.
+     *
+     * @param RepositoryInterface $repository repository instance
+     */
+    public function prependRepository(RepositoryInterface $repository)
+    {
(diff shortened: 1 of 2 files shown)


RemiCollin:
:+1:

alcohol:
Would be great if you could add a unit test for this.

franzliedke:
@alcohol Added the test.

alcohol:
Thanks

franzliedke:
Awesome, thanks so much. Quick question: How are these rolled out? Are they available immediately with a `composer selfupdate`?

franzliedke:
Never mind. Just tried it. ;)

alcohol:
Yes. :-)

PR #4952, opened 2016-02-23
Fix for the unix bin proxy on cygwin if the directory contains spaces

Jos635:
Without the quotes cygpath will interpret the filename (containing spaces) as two or more different files. This results in the value of $dir being incorrect, which in turn causes the command to fail.


Seldaek:
Thanks

PR #4961, opened 2016-02-25
Attempt to use shallow clones, fixes #3449

### LINKED ISSUE #3449: Add support for git shallow clones
When working with a large repository it would be nice to opt for a clone depth of 1 to still leverage git to patch dependencies for a project

Seldaek:
This should do shallow clones but as we can not always know 100% the branch name (there's one edge case like both `2.0` and `2.0.x` branches will become `2.0.x-dev` so from that we can not know for sure which was the original name. In case it fails due to that or just an outdated git then there's a fallback to the complete fetch/clone.

In theory this makes initial installs a lot faster, but once it updates it has to fetch more data and I believe this will always result in a full fetch, so the first update will take the time hit vs the initial install. Further updates should still be fast as usual though.

This can be tried out using this phar file:

https://seld.be/composer-shallow.phar

To try it, use `composer install --prefer-source --profile` or similar to get timing info and force clones.

I'd love to get some more feedback from people before merging this because it has a high chance of unforeseen repercussions :)

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -331,9 +331,9 @@ for everyone.
 
 ## Composer hangs with SSH ControlMaster
 
-When you try to install packages from a Git repository and you use the ```ControlMaster```
-setting for you SSH connection  Composer might just hang endlessly and you see a ```sh```
-process in the ```defunct``` state in your process list
+When you try to install packages from a Git repository and you use the `ControlMaster`
+setting for you SSH connection  Composer might just hang endlessly and you see a `sh`
+process in the `defunct` state in your process list
 
 The reason for this is a SSH Bug: https://bugzilla.mindrot.org/show_bug.cgi?id=1988
 
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -238,6 +238,10 @@
                         "type": "string"
                     }
                 },
+                "git-clone-depth": {
+                    "type": "integer",
+                    "description": "Defaults to 1 (shallow clone). Set to the desired history size. If set to -1, the source will be cloned in full."
+                },
                 "github-expose-hostname": {
                     "type": "boolean",
                     "description": "Defaults to true. If set to false, the OAuth tokens created to access the github API will have a date instead of the machine hostname."
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -340,6 +340,7 @@ EOT
                 function ($val) { return is_dir($val) && is_readable($val); },
                 function ($val) { return $val === 'null' ? null : $val; },
             ),
+            'git-clone-depth' => array('is_numeric', 'intval'),
             'github-expose-hostname' => array($booleanValidator, $booleanNormalizer),
         );
         $multiConfigValues = array(
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -49,6 +49,7 @@ class Config
         'secure-http' => true,
         'cafile' => null,
         'capath' => null,
+        'git-clone-depth' => 1,
         'github-expose-hostname' => true,
         'gitlab-domains' => array('gitlab.com'),
         'store-auths' => 'prompt',
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -45,11 +45,13 @@ class GitDownloader extends VcsDownloader
 
         $ref = $package->getSourceReference();
         $flag = Platform::isWindows() ? '/D ' : '';
-        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
+        $command = '(git clone --no-checkout --depth 1 --single-branch %s %s --branch %s || git clone --no-checkout %1$s %2$s)'.
+            ' && cd '.$flag.'%2$s && git remote add composer %1$s && (git fetch composer %3$s || git fetch composer)';
         $this->io->writeError("    Cloning ".$ref);
 
-        $commandCallable = function ($url) use ($ref, $path, $command) {
-            return sprintf($command, ProcessExecutor::escape($url), ProcessExecutor::escape($path), ProcessExecutor::escape($ref));
+        $branch = $this->guessBranchName($package);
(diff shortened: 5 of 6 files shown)


stof:
Running this on a symfony-standard lock file (which contains tagged versions) made it took 84s for the install vs 194s for the uptodate composer phar. I will try it later on my work project, where I have a few dev versions locked (including ones not at the latest commit of the branch due to the time since the last ``composer update).

Seldaek:
Yup that'd be interesting, I only did basic testing here and it seems 
fine but indeed if you try to checkout/reset to an older reference it 
might fail.

bojanz:
I tested https://github.com/drupalcommerce/project-base, which contains Drupal 8 + Drush + 10 additional dependencies, resolving to about 107 packages (and git clones). Drupal itself is particularly problematic there, since it has a huge git history (and that clone alone sometimes breaks the 300s limitation).

I ran both the original and the shallow installs twice, stopped measuring at "Generating autoload files", since the project has a script that runs afterwords and isn't relevant here.

Original: 1315.57s (!), 733.85s
Shallow: 383s, 276.76s

The results obviously vary because of my internet connection, but the savings are obviously huge.

curry684:
> and that clone alone sometimes breaks the 300s limitation

If that is common we should consider raising it to 600s. You can disable it per command with `COMPOSER_PROCESS_TIMEOUT=0` env variable.

bojanz:
Most projects don't have a history as huge as Drupal. 

Even without "--depth 1", just doing "--single-branch" gives huge savings, since it won't fetch the many previous branches.

barryvdh:
Why not also fetch with `--depth=1`? Do you need the full history at all for the updates?

hkdobrev:
For the problem of exact commit references which are not the latest commit of a branch which @stof mentioned, I'd suggest to just not use the shallow cloning when the version is specified with a commit reference. The majority of version constraints do not use specific commits anyway, so cloning with depth set to 1 will still have a major impact on the majority of installing from source.

---

:+1: on what @bojanz suggested with `--single-branch` if `--depth=1` doesn't work.

stof:
> I'd suggest to just not use the shallow cloning when the version is specified with a commit reference. The majority of version constraints do not use specific commits anyway, so cloning with depth set to 1 will still have a major impact on the majority of installing from source.

That's not true. Anything coming from Packagist or a VCS repository will use a commit reference for the checkout, to allow proper locking (locking on a branch reference would not be locking anything).

hkdobrev:
@stof Sorry, I've misread your previous comment and missed you're talking about commit references in `composer.lock`.

bojanz:
Interesting comments:
https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193772935
https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193801376

Composer doesn't hold package metadata in a single GitHub repo, which means it's not directly applicable, but it's interesting to see which problems other dependency managers are having.

barryvdh:
Especially this part: https://github.com/CocoaPods/CocoaPods/issues/4989#issuecomment-193838934

> To reiterate: the major performance issue is not on doing an initial clone of the CocoaPods repository, but in performing a shallow clone and then repeatedly fetching into it, like the CocoaPods client is currently doing.

Seldaek:
Please be careful though, their repo is very special, so it doesn't necessarily mean the advice there applies to every use of shallow clones. Much like gc_disable helped us greatly but is not a good thing for the other 99% of php projects.

djmattyg007:
Is there a way to completely disable shallow cloning if desired? We have a lot of internal repositories where I work, and the nature of the development means we do a lot of work on these repositories while using them as dependencies. Having these be brought in as shallow clones to begin with would be very annoying for a lot of reasons.

Seldaek:
@djmattyg007 in this PR there is no flag to disable it no, you just need a git fetch in the repo to be able to start working with it though. Not such a hurdle.

djmattyg007:
Would you accept an amendment or a follow-up PR that added a flag to disable it?

Seldaek:
Yes for sure you can PR my branch if you like to add a config for it..

Tuurlijk:
I would love the depth to be configurable in ["config"]["vcs"]["git"]["clone-depth"]. My aim is to install a development box with a wide (about 4) range of TYPO3 installations. These repos have quit a large history, so I don't want all of it. But a history of the 100 most recent commits would be nice for a start. So you can see what changed recently.

If need be, the user can unshallow the repo with `git fetch --unshallow`.

Saphyel:
When this will be added to composer?

schmunk42:
@Seldaek Would be great if this could be added to composer. What's the blocker here?

We'd need this to further improve https://github.com/francoispluchino/composer-asset-plugin/issues/226

Seldaek:
As we have #5384 that caches the git clones I think it's preferable to abandon the shallow cloning approach. It's unlikely to be faster and introduces some edge cases + potential for user confusion so I'll close.

bojanz:
@Seldaek
Can we add --single-branch in a different PR? That one still feels essential, and brings big savings for large repositories (such as Drupal)

saada:
@Seldaek why would we want to download the entire history of a repo when we just need the raw files of the current revision? What do you mean by user confusion? What edge cases are you referring to?

Seldaek:
--single-branch still has the aforementioned downsides though.. I'm not sure what the impact would be, I'm not entirely opposed to it but it'd have to be experimented with. Then again if we have the downsides anyway we maybe want to use shallow clones as well. If you are interested in taking over this PR and playing around feel free.

Seldaek:
@saada I mean that the repo is then not usable unless you `git fetch` into it.. Right now the main use case for cloning instead of installing from dist is because you want to edit the vendors directly. Drupal is a bit of an edge case here because as the repo is at git.drupal.org it doesn't offer dist downloads. Maybe this will be fixed with the final D8 repo, not sure if it includes drupal/drupal packages or if those will still come from packagist. Either way it might be a good thing to use the github mirror of drupal on packagist instead of the original repo, that way you get zip downloads available.

Ping @webflo any clue about this? Or do you know who could tell us?

Tuurlijk:
In some cases the constraining factor is the disk space the clones are taking up. So the 'cache' does not help there. If I have a development box I want to test agains several versions of a cms for example. I don't need the full history for each 'tag' I want to test against.

Issue #4955, opened 2016-02-24
Content of junction directories gets deleted after running composer update


hediet:
In my scenario there are two composer projects: "lib" and "main". "main" has a dependency to "lib".
As I am developing both projects simultaneously I used the path type repository feature that instructs composer to create a junction directory rather than downloading the "lib" project from packagist.org.
However, when I now remove the dependency from "main" to "lib" and run `composer update` in the "main" directory, composer deletes the whole content of the "lib" directory instead of just removing the junction directory. This almost led to a devastating loss of really much work...

### Steps to reproduce
This is the `composer.json` file of the lib project in the directory "lib":
```json
{
    "name": "hediet/lib"
}
```
And of the main project in the directory "main":
```json
{
    "name": "hediet/main",
    "require": {
	    "hediet/lib": "*"
	},
	"repositories": [
        {
            "type": "path",
            "url": "C:\\Users\\Henning\\Desktop\\composer-bug\\lib"
        }
    ]
}
```
After running `composer update` in both the lib and the main directory, I removed the dependency from main to lib. Then I ran `composer update` again.

### Expected result
The vendor directory in the main directory is empty, the original lib directory including any files in it are not touched by the composer updates.

### Actual result
The vendor directory in the main directory is empty and the lib directory on root level is now empty too. All files (and subdirectories) of the lib directory were deleted with the last `composer update`.

curry684:
Please know I take this issue very seriously as it's my code and I feared it may have such consequences, as PHP does not provide any kind of API to reliably detect junctions itself. We did a few doublechecks before releasing, ie. https://github.com/composer/composer/pull/4690#issuecomment-175917892. I will try to reproduce and fix your issue when I have time.

curry684:
Thinking some more and I suspect this is a nasty fringe case where a recursive delete is still used when purging packages, while updates detect the junction correctly. As the Symfony filesystem class is oblivious of junctions, and NTFS junctions are close to undetectable anyway, a recursive delete does not notice the junction at all and just continues deleting files inside the linked folder.

Seldaek:
@curry684 seeing as we only use junctions in PathDownloader, it might be worth implementing PathDownloader::remove() and just doing a removeJunction always in there if we're on windows. I mean always *attempt* removing the junction using the proper command, and if that fails then run a regular remove?

[...]

PR #4962, opened 2016-02-25
Improve handling of removing junctions in Windows path repositories

curry684:
See #4955, added both extra detection reliability and lessened chances of unknown errors causing unwanted recursive deletions.


Seldaek:
Thanks, looks good

PR #4970, opened 2016-02-27
Change default behavior of show/list to --installed by default

barryvdh:
This might be just me, but I always feel that `--installed` is actually what I'm looking for, instead of the huge list from all the packages. This was probably more useful in the beginning with lesser packages, or with just custom repositories.

This removes the `--installed` flag and makes that the default behavior. A `--all` flag is added to show all packages (which was the old behavior).

I couldn't use the `-a` shortcut, as that is used by `--available`. Do you want to switch that, or have a better name?

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -53,6 +53,7 @@ class ShowCommand extends BaseCommand
                 new InputArgument('package', InputArgument::OPTIONAL, 'Package to inspect'),
                 new InputArgument('version', InputArgument::OPTIONAL, 'Version or version constraint to inspect'),
                 new InputOption('all', null, InputOption::VALUE_NONE, 'List all packages'),
+                new InputOption('installed', 'i', InputOption::VALUE_NONE, 'List installed packages only (enabled by default, only present for BC).'),
                 new InputOption('platform', 'p', InputOption::VALUE_NONE, 'List platform packages only'),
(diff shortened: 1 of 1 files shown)


curry684:
I agree with the change but you should at the very least keep supporting `-i` for BC scripts, marking it deprecated, as the current flag was part of tagged releases. See for example `--dev` in https://github.com/composer/composer/blob/master/src/Composer/Command/InstallCommand.php#L40

barryvdh:
Wel yeah, alpha tags. And --dev on the install command is a flag that was commonly used on CI scripts, so much bigger impact when failing. But sure, I can add it to keep BC for a while.

src/Composer/Command/ShowCommand.php
```diff
@@ -79,9 +80,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $composer = $this->getComposer(false);
         $io = $this->getIO();
 
-        if ($input->getOption('tree') && !$input->getOption('installed')) {
-            $io->writeError('The --tree (-t) option is only usable in combination with --installed (-i) or by passing a single package name to show, assuming -i');
-            $input->setOption('installed', true);
+        if ($input->getOption('installed')) {
+            $io->writeError('<warning>You are using the deprecated option "installed". Only installed packages are shown by default now. The --all option can be used to show all packages.</warning>');
```
barryvdh: @curry684 Better?

curry684: :+1:

Seldaek:
Thanks

PR #4972, opened 2016-02-27
Code quality

curry684:
Commits are separated into repaired groups, summarized:

- Fixed some config manipulation commands not working on PHP 5.3 as it didn't support use of `$this` in closures
- Added all missing parent constructor calls (bugs waiting to happen when base functionality does become meaningful)
- Stripped a lot of unused variables and imports
- Repaired some functions being called with more parameters than used
- Fixed a bunch of implicit member variable declarations

### DIFF
--- a/src/Composer/Config/JsonConfigSource.php
+++ b/src/Composer/Config/JsonConfigSource.php
@@ -79,10 +79,11 @@ class JsonConfigSource implements ConfigSourceInterface
      */
     public function addConfigSetting($name, $value)
     {
-        $this->manipulateJson('addConfigSetting', $name, $value, function (&$config, $key, $val) {
+        $that = $this;
+        $this->manipulateJson('addConfigSetting', $name, $value, function (&$config, $key, $val) use ($that) {
             if (preg_match('{^(github-oauth|gitlab-oauth|http-basic|platform)\.}', $key)) {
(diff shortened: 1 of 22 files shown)


src/Composer/Config/JsonConfigSource.php
```diff
@@ -98,10 +99,11 @@ public function addConfigSetting($name, $value)
      */
     public function removeConfigSetting($name)
     {
-        $this->manipulateJson('removeConfigSetting', $name, function (&$config, $key) {
+        $that = $this;
+        $this->manipulateJson('removeConfigSetting', $name, function (&$config, $key) use ($that) {
             if (preg_match('{^(github-oauth|gitlab-oauth|http-basic|platform)\.}', $key)) {
                 list($key, $host) = explode('.', $key, 2);
-                if ($this->authConfig) {
+                if ($that->authConfig) {
```
Seldaek: This actually fails as well on 5.3 because authConfig is private (as per https://3v4l.org/RTUFL). Fixed in 2ff19e4

curry684: Hmm weird, didn't turn up in code inspections, and I expected it to still have class visibility, just no `$this` context - sort of inline `static`.

Issue #4973, opened 2016-02-28
Unit test broken by uncommitted change detection


curry684:
Commit https://github.com/composer/composer/commit/be4d385942cce9b0d1116aba985a99380c7a53d1 broke a unit test in `GitDownloaderTest` called `testUpdateDoesntThrowsRuntimeExceptionIfGitCommandFailsAtFirstButIsAbleToRecover`.

The problem is that it sets up a mock `ProcessExecutor` expecting to see the same sequence of commands attempted twice. However since that commit the order of execution the first time may be different from the second time, depending on the test platform, as the diff detection inserts extra commands. Therefore Appveyor and at least Windows/phpStorm dev setups cannot complete this test anymore. I can't for the life of me figure out so far why it doesn't break on Travis though.

The original test was added a few days before in https://github.com/composer/composer/commit/4689d836fef76ede3238208ce5169b11f8cf1f53 but as far as I can see it did work initially.

Submitting a PR in a second to disable it until this has been more fundamentally researched. Ping @magnusnordlander.

PR #4974, opened 2016-02-28
Disabled unit test breaking on some platforms. Ref #4973

curry684:
#4973


curry684:
Thankfully the Appveyor tests completed correctly now hehe.

Issue #3990, opened 2015-05-01
Circular dependancy doesn't register


hackzilla:
After trying to push an update to a private bundle this morning, I noticed that Composer was completely ignoring.  I thought it was a corrupted version, but it wouldn't install any previous versions.

Some how my bundle had itself as a dependancy, which the last time I install it worked (March 2015).

```
git clone https://github.com/hackzilla/TestBundle.git test
cd test
composer install
```

Whilst agree that this shouldn't install itself, it should issue a warning/error, and not silently ignore it.

Seldaek:
Seems fine to me.. you require a package that is already there so it doesn't install it. It also doesn't tell you it's not installing php or any other package you already have when you do an install/update.

hackzilla:
in this example it is a bit pointless. 

However try and use that bundle as a dependancy.
You will not be able to install / update it.

hackzilla:
e.g.

```
cat composer.json 
{
    "name": "hackzilla/test-project",
    "description": "",
    "repositories": [
        {
          "type": "vcs",
          "url":  "git@github.com:hackzilla/TestBundle.git"
        }
    ],
    "require": {
        "hackzilla/test": "dev-master"
    },
    "authors": [
        {
            "name": "Daniel Platt",
            "email": "github@ofdan.co.uk"
        }
    ]
}

composer install

Loading composer repositories with package information
Installing dependencies (including require-dev)       
Nothing to install or update
Generating autoload files
```

Seldaek:
Ah I see.. That is weird indeed.

Not really a real use case though so not exactly an urgent problem right?

hackzilla:
Not urgent.

However I lost 3 hours of my morning, trying to debug this.
No idea how the dependancy got put into my bundle, but it potentially could help others.

The strange thing is only stopped working sometime in the last 2 months.

naderman:
@hackzilla can you test if this is fixed with the latest version of composer?

hackzilla:
I can confirm its working with composer version 6d1e8ebf90542a7c65ac6661f6c47b1d655641d9

Cheers.

PR #4977, opened 2016-02-29
Add failing test for #3990

Seldaek:
This reproduces the weird case in #3990 - it certainly isn't the smartest package definition that is requiring itself, but the outcome is weird nonetheless.

Basically:

- if root requires A, and A requires itself, A is skipped entirely
- if root requires A and B, A requires itself still, but B requires A as well, then A and B are installed.

You can see the second case by changing `__dummy__require` into `require`, which makes the test pass. All quite odd.

cc @naderman

### DIFF
--- a/tests/Composer/Test/Fixtures/installer/circular-dependency2.test
+++ b/tests/Composer/Test/Fixtures/installer/circular-dependency2.test
@@ -1,0 +1,39 @@
+--TEST--
+Circular dependencies are possible between packages
+--COMPOSER--
+{
+    "name": "root",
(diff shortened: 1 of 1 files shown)


naderman:
Heh that's funny, I imagine unless the B dependency is there it registers the package in two chains in the solver and just flips install flag back and forth. Curious if it skips it if you add another dependency C ^_^ Should be easy to fix, will take a look tomorrow.

naderman:
Closing in favour of https://github.com/composer/composer/pull/4989

PR #4982, opened 2016-03-01
Add changelog entry for #4970

barryvdh:
Add changelog entry for #4970


Seldaek:
Not really a major BC break but fair enough ;)

PR #4989, opened 2016-03-02
Ignore self references when anlaysing independent root requirements

naderman:
When creating a transaction we try to identify all requirements that are
not themselves required by any other package. If a package references
itself this should not mark it as being required by another package.

Includes test from https://github.com/composer/composer/pull/4977
Addresses the bug in https://github.com/composer/composer/issues/3990

### DIFF
--- a/src/Composer/DependencyResolver/Transaction.php
+++ b/src/Composer/DependencyResolver/Transaction.php
@@ -165,7 +165,9 @@ class Transaction
                 $possibleRequires = $this->pool->whatProvides($link->getTarget(), $link->getConstraint());
 
                 foreach ($possibleRequires as $require) {
-                    unset($roots[$require->id]);
+                    if ($require !== $package) {
+                        unset($roots[$require->id]);
+                    }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Composer/DependencyResolver/Transaction.php
```diff
@@ -165,7 +165,9 @@ protected function findRootPackages($installMap, $updateMap)
                 $possibleRequires = $this->pool->whatProvides($link->getTarget(), $link->getConstraint());
 
                 foreach ($possibleRequires as $require) {
-                    unset($roots[$require->id]);
+                    if ($require !== $package) {
```
stof: shouldn't you check the package name for equality rather than the Package object ? Are we sure we will always get only the same instance ?

naderman: Yes, because we delete `$roots[$require->id]` so the id would have to be identical anyway for this to cause a bug.

xuwenzhao:
76450cc

PR #4993, opened 2016-03-02
List project suggestions in create-project command

hkdobrev:
Resolves https://github.com/composer/composer/issues/2752

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -17,6 +17,7 @@ use Composer\Factory;
 use Composer\Installer;
 use Composer\Installer\ProjectInstaller;
 use Composer\Installer\InstallationManager;
+use Composer\Installer\SuggestedPackagesReporter;
 use Composer\IO\IOInterface;
 use Composer\Package\BasePackage;
 use Composer\DependencyResolver\Pool;
@@ -47,6 +48,11 @@ use Composer\Package\Version\VersionParser;
  */
 class CreateProjectCommand extends BaseCommand
 {
+    /**
+     * @var SuggestedPackagesReporter
+     */
+    protected $suggestedPackagesReporter;
+
     protected function configure()
     {
         $this
@@ -142,6 +148,8 @@ EOT
         // we need to manually load the configuration to pass the auth credentials to the io interface!
         $io->loadConfiguration($config);
 
+        $this->suggestedPackagesReporter = new SuggestedPackagesReporter($io);
+
         if ($packageName !== null) {
             $installedFromVcs = $this->installRootPackage($io, $config, $packageName, $directory, $packageVersion, $stability, $preferSource, $preferDist, $installDevPackages, $repository, $disablePlugins, $noScripts, $keepVcs, $noProgress);
         } else {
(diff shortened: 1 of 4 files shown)


curry684:
I understand *why* you implemented it like this but I don't like the implementation. What you did now was add a public utility function to the `Installer` for the sole purpose of injecting, to the installer at that point, arbitrary data into it to trick it into rendering more suggestions at the end. I would like this change not to touch the `Installer` class at all,  as it has no real relationship to it.

This also means that the implementation is flawed, as the suggestions are not rendered now if `--no-install` is supplied, which should have no relationship to whether the root package suggestions are listed. Secondly your PR introduces a member variable in a command class which we are trying to avoid as we want to remove logic and state from the commands some day.

Imo a better implementation would be to add an output parameter to `installRootPackage` returning the suggestions, and then plainly rendering them in the calling function after the `if ($noInstall === false)` section. This way they would always be shown regardless of `--no-install` and introduce a lot less invasive changes.

hkdobrev:
@curry684 Thanks for the pointers! I will try to refactor it the next few days.

Initially I've implemented it only in `CreateProjectCommand`, but the problem is it's not so easy to get all of the installed packages in that command, but even when I did so the output of the suggestions is different places. E.g. You first see the suggested packages from the installed dependencies, then the output of `post-install-cmd` hook, then suggestions from the root project package, then output from `post-create-project-cmd`. Also I've copy-pasted some logic from `Installer` in that implementation.

I think your suggested implementation would work, but at the moment I'm not sure the output would be structured in a nice way and suggestions would be presented together. Will look into it.

hkdobrev:
P.S. I was really amazed of the structure of Composer commands, lack of unit tests for major parts of the system which actually keeps the interface untestable and sometimes confusing.

curry684:
>  I'm not sure the output would be structured in a nice way and suggestions would be presented together

They wouldn't, but that's (again, my opinion) better. The depended upon libraries often suggest packages that just enhance their behavior or add additional interfacing, those suggestions are usually ignored. On the other hand a project that *wants* to be installed through `create-project` would suggest packages really improving the project as a whole. I think it makes sense to show those separately underneath the 'softer' suggestions.

hkdobrev:
@curry684 So you are saying it's better the more important suggestions (from the root package) to be shown below. Because it's more likely the developer would see them without scrolling?

curry684:
That would be the intended result yes. But before we get too far - I'm not the boss here, such details may be better left to @Seldaek to speak on hehe.

Seldaek:
I tend to agree with @curry684 about displaying them at the end. 

What would make sense though to clean things up and avoid duplication is to extract the suggest handling into a new class that would just take an IOInterface, InstalledRepositoryInterface, and have methods like addPackage() to gather suggestions then output() to write it out.

hkdobrev:
@Seldaek Updated it with a separate class taking care of suggestions. I hope this is not too much different than what you had in mind. I've started working on tests for the new class, but wanted to see if that's what you meant.

src/Composer/Command/CreateProjectCommand.php
```diff
@@ -318,6 +327,15 @@ protected function installRootPackage(IOInterface $io, Config $config, $packageN
         $im->install(new InstalledFilesystemRepository(new JsonFile('php://memory')), new InstallOperation($package));
         $im->notifyInstalls($io);
 
+        // collect suggestions
+        foreach ($package->getSuggests() as $target => $reason) {
+            $this->suggestedPackagesReporter->addPackage(
+                $package->getPrettyName(),
+                $target,
+                $reason
+            );
```
Seldaek: This block could be de-duplicated into the Reporter class if you just pass the $package

Seldaek:
Looking good yes :)

hkdobrev:
@Seldaek updated with unit tests and code coverage.

Seldaek:
Thanks!

PR #4995, opened 2016-03-03
Also clean backups if no Composer update actually performed

vlakoff:
I noticed `composer self-update --clean-backups` isn't removing the backups when no update is actually performed ("You are already using composer version...").

I had to do a little refactoring to implement the corrected behaviour. Split into several commits to make it easier to review.

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -105,6 +105,11 @@ EOT
         if (Composer::VERSION === $updateVersion) {
             $io->writeError('<info>You are already using composer version '.$updateVersion.'.</info>');
 
+            // remove all backups except for the most recent, if any
+            if ($input->getOption('clean-backups')) {
+                $this->cleanBackups($rollbackDir, $this->getLastBackupVersion());
+            }
+
             return 0;
         }
 
@@ -315,13 +320,16 @@ TAGSPUBKEY
         }
     }
 
-    protected function cleanBackups($rollbackDir)
+    protected function cleanBackups($rollbackDir, $except = null)
     {
         $finder = $this->getOldInstallationFinder($rollbackDir);
         $io = $this->getIO();
         $fs = new Filesystem;
 
         foreach ($finder as $file) {
+            if ($except && $file->getBasename(self::OLD_INSTALL_EXT) === $except) {
+                continue;
+            }
             $file = (string) $file;
             $io->writeError('<info>Removing: '.$file.'</info>');
             $fs->remove($file);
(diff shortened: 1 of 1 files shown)


src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -117,6 +110,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
             self::OLD_INSTALL_EXT
         );
 
+        if (Composer::VERSION === $updateVersion) {
+            $io->writeError('<info>You are already using composer version '.$updateVersion.'.</info>');
+
+            if ($input->getOption('clean-backups')) {
+                $this->cleanBackups($rollbackDir, $backupFile);
```
stof: this looks wrong to me, as ``$backupFile`` does not exist, so you will not keep any backup (while ``--clean-backups`` is meant to keep the latest backup).

You should instead call ``$this->cleanBackups`` with the last backup file (the logic to get it is at the beginning of ``getLastBackupVersion``)

src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -214,6 +206,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
         } else {
             $io->writeError('<warning>A backup of the current version could not be written to '.$backupFile.', no rollback possible</warning>');
         }
+
+        // remove saved installations of composer
+        if ($input->getOption('clean-backups')) {
```
stof: this should be kept at the previous location, otherwise it may not be done in case of corrupted file

src/Composer/Command/SelfUpdateCommand.php
vlakoff: I finally got it, thanks for pointing this out. This will be fixed.

src/Composer/Command/SelfUpdateCommand.php
vlakoff: In `setLocalPhar` the backup is written only after successful Phar validation.

vlakoff: Mmm, not sure about that. Still I'm fine moving this code back to original position.

vlakoff: @stof Looking at the code of `setLocalPhar` and the two calls to it, I think this:

    if ($backupTarget) {
        @unlink($newFilename);
    }

Should actually be only:

    @unlink($newFilename);

Can you please confirm this?

vlakoff:
Updated PR, thanks to @stof for his comments. As a bonus, changes are simpler than with the previous attempt.

Seldaek:
Thanks

PR #4998, opened 2016-03-03
Removed inefficiency in the autoloader generation

curry684:
This line was originally added for PHP 5.2 compatibility but that was dropped long ago, so no longer need to incur the handful of extra function calls on every autoloader init.


Seldaek:
It's actually not been dropped no, the point was that people rely on composer to generate autoloaders for project that they then run on old 5.2 or god knows maybe 5.1.6 servers. It's not really a huge inefficiency so I think we can keep this around a bit longer.. I can keep the PR open for now though, maybe merge it for composer 2.0 ;)

curry684:
I'm pretty sure nobody is doing that as the current release already breaks it in `vendor/autoload.php`:
```php
<?php

// autoload.php @generated by Composer

require_once __DIR__ . '/composer' . '/autoload_real.php';
```
I even think you broke that in december 2011 with https://github.com/composer/composer/commit/4517a2e5#diff-6df1954b3e65ffe3fa388171dca63088L93   :laughing:

Seldaek:
No I think they were probably using the autoload_* files to create their own 5.2 compatible autoloader then. Because the ClassLoader is namespaced so that can not be used either.

staabm:
@curry684 I used it (actually requested the feature) and it works. 

the use case is:
- generate/manage composer on a "current" php version
- but use 5.2 in production

curry684:
Ok, but PHP 5.2 itself went EOL 5 years and 2 months ago, and Ubuntu 10 LTS same 10 months ago. How long are we going to actively support exotic scenarios like this while the backfix is trivial for those that need it?

staabm:
As this few lines have nearly no vost I agree with Seldaek that it should be droppend with 2.0

curry684:
@Seldaek  A year has passed and since 2.0 isn't really happening soon - let's kick PHP 5.2 to the curb and win these 10 microseconds per pageview  ;)

stof:
@curry684 you won't win anything, given that these files are not used anymore on PHP 5.6+, in favor of `autoload_static.php`

Seldaek:
Closing in favor of #9433 which is rebased.

PR #5001, opened 2016-03-04
Show -t incorrectly expanded self.version constraints. Fixes #4999

### LINKED ISSUE #4999: composer show --tree fails when a package requires a dependency with self.version
Example:

```
$ composer require zf1/zend-pdf
$ composer show

zf1/zend-cache     1.12.11 Zend Framework 1 Cache package
zf1/zend-exception 1.12.11 Zend Framework 1 Exception package
zf1/zend-log       1.12.11 Zend Framework 1 Log package
zf1/zend-memory    1.12.11 Zend Framework 1 Memory package
zf1/zend-pdf       1.12.11 Zend Framework 1 Pdf package

$ composer show --tree

zf1/zend-pdf 1.12.11 Zend Framework 1 Pdf package
├──php >=5.2.11
├──zf1/zend-exception self.version

                                         
  [ErrorException]                       
  Illegal offset type in isset or empty  
                                         

show [--all] [-i|--installed] [-p|--platform] [-a|--available] [-s|--self] [-N|--name-only] [-P|--path] [-t|--tree] [--] [<package>] [<version>]
```

`composer show` without --tree works fine.

curry684:
Fixes #4999 - for self.version the constraint comes preparsed into `getPackage`.

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -19,6 +19,7 @@ use Composer\Package\Version\VersionParser;
 use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
 use Composer\Package\PackageInterface;
+use Composer\Semver\Constraint\ConstraintInterface;
 use Composer\Util\Platform;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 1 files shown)


barryvdh:
Alternative is #5000 
We can cast it to string, because it has a __toString() method, which gives the pretty version.

barryvdh:
Okay, so that is indeed a bit pointless. Better to check if it's already an instance of ConstraintInterface as we discussed, and check that also. I'll close the other PR.

curry684:
And I fixed the booboo with the other code path  :+1:

PR #5002, opened 2016-03-04
Make remove with dependencies default

### LINKED ISSUE #4895: cleanup inherited dependencies when removing a root dependency

barryvdh:
Fixes #4895

As discussed in #4895, the old behaviour was not intuitive. I deprecated the old option (`--update-with-dependencies`) and made that the default. Also added `--no-update-with-dependencies` if anyone would want to use the old behaviour.


curry684:
:+1:  on the change, just keep in mind that it doesn't *fix* #4895 and that issue shouldn't be closed yet.

Seldaek:
Thanks

PR #5006, opened 2016-03-04
Check for --ignore-platform-reqs flag during create-project

### LINKED ISSUE #5005: Running `create-project` does not honor `--ignore-platform-reqs` when installing root package
During `create-project`, the installed PHP version is used to determine the best root package install candidate, but fails to take into account the `--ignore-platform-reqs` flag. The check against PHP version was added at commit https://github.com/composer/composer/commit/ef2856ef55715027ce2275f3b3bf1e84bfc778a4 on October 29th, 2015

This can be tested by running `create-project` with an installed PHP version lower than the required version of the project you are installing and including the `--ignore-platform-reqs` flag. For example, with PHP 5.5 installed, run:

```
composer --ignore-platform-reqs create-project loadsys/skeleton ./test ~3.2
```

This project requires PHP v5.6 or greater, so the result is a version `3.2` or greater cannot be found even though version `3.3.1` of `loadsys/skeleton` is released. Version `3.0.0`, which required a lesser PHP version, is installed if the command is executed without the version `~3.2` parameter.

I can provide verbose command output if requested, but I'm not sure it is useful in this case. I'll provide a PR shortly that honors the `--ignore-platform-reqs` flag during the root package install.

ricog:
During `create-project`, the installed PHP version was used to determine
the best root package install candidate, but failed to take into account
the `--ignore-platform-reqs` flag.

Closes #5005

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -143,7 +143,7 @@ EOT
         $io->loadConfiguration($config);
 
         if ($packageName !== null) {
-            $installedFromVcs = $this->installRootPackage($io, $config, $packageName, $directory, $packageVersion, $stability, $preferSource, $preferDist, $installDevPackages, $repository, $disablePlugins, $noScripts, $keepVcs, $noProgress);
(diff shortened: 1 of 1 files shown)


ricog:
The AppVeyor build is failing when downloading from the repo. I'm not sure if I can do anything about this.

Seldaek:
Thanks, no worries about appveyor it's always a bit unstable.

PR #5010, opened 2016-03-05
Force @covers annotations

hkdobrev:
From https://phpunit.de/manual/current/en/appendixes.configuration.html:
> Code Coverage will only be recorded for tests that use the @covers annotation documented in the section called “@covers”.

This way if code coverage is being generated, tests without the @covers
annotation would cover nothing.

Code coverage does not guarantee everything, but with this option
you could be more sure the code coverage was intentional and explicit.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -6,6 +6,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
+         forceCoversAnnotation="true"
          processIsolation="false"
          stopOnFailure="false"
          syntaxCheck="false"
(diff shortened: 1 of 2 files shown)


curry684:
I'm pretty sure (about 101% hehe) we don't want this option set in phpUnit. What problem are you exactly trying to fix?

Seldaek:
I agree, covers annotations can be used regardless of this flag if they are deemed necessary. Carpet bombing isn't needed.

hkdobrev:
@curry684 :laughing: No, I'm not going to write that many unit tests, but I will think about adding the `@covers` annotations and see what would the coverage be then. Thank you for summarising what the option is doing and posting the result specific to Composer tests.

@curry684 @Seldaek My thoughts are that code coverage number is deceptive and without that option set, it is even more deceptive.

Cheers!

Seldaek:
I agree but I think coverage overall is a lie and just an indicator you can
use to know what is undertested. A line being covered never guarantees it
is well tested, and striving for a certain coverage percentage is pointless
IMO

PR #5026, opened 2016-03-08
Simulate dummy Composer when searching in uninitialized folder

### LINKED ISSUE #4392: search without composer.json does not use auth.json
so i run search, which uses repository using credentials, i let save the credentials, i search again, it does ask again credentials

```
$ composer --version
Composer version 1.0-dev (afd98b74f13b9def1b2447f16fd0bfd4d96dff94) 2015-08-31 09:58:50

$ composer search nada
No composer.json found in the current directory, showing packages from whatever
    Authentication required (whatever):
      Username: glen
      Password: 
Do you want to store credentials for whatever in /home/vagrant/.composer/auth.json ? [Yn] Y

$ composer search nada
No composer.json found in the current directory, showing packages from whatever
    Authentication required (whatever):
      Username: glen


$ ls -l composer.json ~/.composer/auth.json
ls: cannot access composer.json: No such file or directory
-rw------- 1 vagrant vagrant 155 sept   2 10:58 /home/vagrant/.composer/auth.json
```

curry684:
`SearchCommand` uses same `getComposer(false)` call as many other commands, but does not load any config obviously at all, including `auth.json`, when no `composer.json` is found. This PR injects a dummy `Composer` with empty config instead, thus including all global config.

Should fix #4392

### DIFF
--- a/src/Composer/Command/SearchCommand.php
+++ b/src/Composer/Command/SearchCommand.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Command;
 
+use Composer\Factory;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


Seldaek:
Yup seems fine, thanks

PR #5028, opened 2016-03-09
Add latest version to show command (option)

barryvdh:
This is a partial fix for #3771

Shows the recommend version, same as used in the require command. When used with the `--recommended` or `-r` flag, it shows an extra column in the table layout, or extra row for 1 package.
Multiple packages:
image

Single package:
image

Possible improvements (which I'm not really sure how to do easy):
 - Show the exact latest versions (eg. `3.1.2` instead of `^3.2`)
 - Show the current constraint? (`eg `3.x` instead of `3.1.2`)
 - Take PHP version into account?
 - Add coloring to see of the current constraint allows the latest version?

Update: Now using `--latest` / `-l` and showing the latest version, instead of best constraint.

image

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -16,6 +16,7 @@ use Composer\DependencyResolver\Pool;
 use Composer\DependencyResolver\DefaultPolicy;
 use Composer\Package\CompletePackageInterface;
 use Composer\Package\Version\VersionParser;
+use Composer\Package\Version\VersionSelector;
 use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
 use Composer\Package\PackageInterface;
@@ -45,6 +46,12 @@ class ShowCommand extends BaseCommand
     protected $versionParser;
     protected $colors;
 
+    /** @var CompositeRepository */
+    protected $repos;
+
+    /** @var Pool */
+    private $pool;
+
     protected function configure()
     {
         $this
@@ -62,6 +69,7 @@ class ShowCommand extends BaseCommand
                 new InputOption('name-only', 'N', InputOption::VALUE_NONE, 'List package names only'),
                 new InputOption('path', 'P', InputOption::VALUE_NONE, 'Show package paths'),
                 new InputOption('tree', 't', InputOption::VALUE_NONE, 'List the dependencies as a tree'),
+                new InputOption('latest', 'l', InputOption::VALUE_NONE, 'Show the latest version'),
             ))
             ->setHelp(<<<EOT
 The show command displays detailed information about a package, or
@@ -146,7 +154,11 @@ EOT
             if ($input->getOption('tree')) {
                 $this->displayPackageTree($package, $installedRepo, $repos);
             } else {
-                $this->printMeta($package, $versions, $installedRepo);
+                $latestVersion = null;
+                if ($input->getOption('latest')) {
+                    $latestVersion = $this->findBestVersionForPackage($package->getName(), null);
+                }
+                $this->printMeta($package, $versions, $installedRepo, $latestVersion);
                 $this->printLinks($package, 'requires');
                 $this->printLinks($package, 'devRequires', 'requires (dev)');
                 if ($package->getSuggests()) {
@@ -217,6 +229,7 @@ EOT
         }
 
         $showAllTypes = $input->getOption('all');
+        $showLatest = $input->getOption('latest');
         $indent = $showAllTypes ? '  ' : '';
         foreach (array('<info>platform</info>:' => true, '<comment>available</comment>:' => false, '<info>installed</info>:' => true) as $type => $showVersion) {
             if (isset($packages[$type])) {
@@ -225,7 +238,7 @@ EOT
                 }
                 ksort($packages[$type]);
 
-                $nameLength = $versionLength = 0;
+                $nameLength = $versionLength = $latestLength = 0;
                 foreach ($packages[$type] as $package) {
                     if (is_object($package)) {
                         $nameLength = max($nameLength, strlen($package->getPrettyName()));
@@ -249,9 +262,11 @@ EOT
                     $input->setOption('path', false);
                 }
 
+                $latestLength = $versionLength;
                 $writePath = !$input->getOption('name-only') && $input->getOption('path');
                 $writeVersion = !$input->getOption('name-only') && !$input->getOption('path') && $showVersion && ($nameLength + $versionLength + 3 <= $width);
-                $writeDescription = !$input->getOption('name-only') && !$input->getOption('path') && ($nameLength + ($showVersion ? $versionLength : 0) + 24 <= $width);
+                $writeLatest = !$input->getOption('name-only') && !$input->getOption('path') && $showLatest && ($nameLength + ($showVersion ? $versionLength : 0) + $latestLength + 3 <= $width);
+                $writeDescription = !$input->getOption('name-only') && !$input->getOption('path') && ($nameLength + ($showVersion ? $versionLength : 0) + ($showLatest ? $latestLength : 0) + 24 <= $width);
                 foreach ($packages[$type] as $package) {
                     if (is_object($package)) {
                         $io->write($indent . str_pad($package->getPrettyName(), $nameLength, ' '), false);
@@ -260,9 +275,15 @@ EOT
                             $io->write(' ' . str_pad($package->getFullPrettyVersion(), $versionLength, ' '), false);
                         }
 
+                        if ($writeLatest) {
+                            $latestVersion = $this->findBestVersionForPackage($package->getName());
+                            $type = $latestVersion == $package->getFullPrettyVersion() ? 'info' : 'comment';
+                            $io->write(' <'.$type.'>' . str_pad($latestVersion, $latestLength, ' ') . '</'.$type.'>', false);
+                        }
+
                         if ($writeDescription) {
                             $description = strtok($package->getDescription(), "\r\n");
-                            $remaining = $width - $nameLength - $versionLength - 4;
+                            $remaining = $width - $nameLength - $versionLength -  ($writeLatest ? $latestLength : 0) - 4;
                             if (strlen($description) > $remaining) {
                                 $description = substr($description, 0, $remaining - 3) . '...';
                             }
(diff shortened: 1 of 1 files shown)


vinkla:
:+1:

barryvdh:
If we want to make it more specific for versions, we could also make it a separate command, like the suggested `composer outdated`.

vinkla:
We could also add a `--latest` or `--outdated` to the `show` command.

barryvdh:
I've changed it to `--latest` / `-l` and used the exact version.

image

barryvdh:
+ some coloring when the version is a match (thus latest)

image

src/Composer/Command/ShowCommand.php
```diff
@@ -249,9 +262,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
                     $input->setOption('path', false);
                 }
 
+                $latestLength = $versionLength;
```
barryvdh: Yes I'm making the lazy assumption that they're equal. I should probably check the actual length, but that would require some caching of the best package version, so I'll leave that as improvement for now.

aminin:
:+1:

barryvdh:
@Seldaek Do you think making it a separate command would be better, regarding #5060 ?

Seldaek:
Not necessarily no. Although I guess for discoverability an outdated command that calls show --latest might be good.

barryvdh:
Okay I'll see if I can put a deprecation notice in here also then.

Seldaek:
Maybe best to keep it separated tbh.. Rather than having this PR grow more. At least start a new PR based on this one.

barryvdh:
I don't think it's too intrusive on the existing code, but created a new PR, based on this one: https://github.com/composer/composer/pull/5068

Seldaek:
Yayayay, merged :) This is very cool. I tweaked things a little (https://github.com/composer/composer/commit/e7069dd5e8ff3c2cc0fa63004f8ab0365a825037), added a `composer outdated` alias and some more coloring to follow npm style colors (orange if it needs an update but is not BC, red if it needs an update and you really should do it because it should not break). And well fixed a few things that I can't blame you for messing up because they needed pretty low level composer knowledge :) But now it handles stability flags and such properly.

vinkla:
@seldaek great job!

Question, is it possible to exclude the packages that already are green?

glen-84:
Maybe `list` should show all packages, and `outdated` should hide the up-to-date packages?

Seldaek:
I now added two filters, --outdated/-o (hide greens) and --direct/-D (show only your own direct dependencies).

rsdev000:
Hi 

Always get this error on Windows with latest php 7.
composer_error

Seldaek:
Can you share your composer.json?

rsdev000:
Sure. It's from a Laravel app.

```
{
    "name": "laravel/laravel",
    "description": "The Laravel Framework.",
    "keywords": ["framework", "laravel"],
    "license": "MIT",
    "type": "project",
    "require": {
        "php": ">=5.5.9",
        "laravel/framework": "5.2.*",
        "mcamara/laravel-localization": "^1.1",
        "barryvdh/laravel-httpcache": "^0.2.4",
        "radic/blade-extensions": "^6.2",
        "predis/predis": "^1.0",
        "aaronjan/housekeeper": "^2.3"
    },
    "require-dev": {
        "fzaninotto/faker": "~1.4",
        "mockery/mockery": "0.9.*",
        "phpunit/phpunit": "~4.0",
        "symfony/css-selector": "2.8.*|3.0.*",
        "symfony/dom-crawler": "2.8.*|3.0.*",
        "rap2hpoutre/laravel-log-viewer": "^0.6.0",
        "itsgoingd/clockwork": "^1.11",
        "barryvdh/laravel-ide-helper": "^2.1",
        "filp/whoops": "^2.1",
        "themsaid/laravel-langman": "^1.1",
        "sven/artisan-view": "^1.0"


    },
    "suggest": {
        "media24si/response-cache": "dev-master",
        "barryvdh/laravel-debugbar": "^2.2",
        "matthiasmullie/minify": "^1.3",
        "raveren/kint": "^1.0"
    },
    "autoload": {
        "classmap": [
            "database"
        ],
        "psr-4": {
              "App\\": "app/",
              "Larapack\\AssetCacheBuster\\": "packages/larapack/assetcachebuster/src",
              "Larapack\\DbTableGenerator\\": "packages/larapack/dbtablegenerator/src",
              "Larapack\\Logger\\": "packages/larapack/logger/src",
              "Larapack\\LoremIpsum\\": "packages/larapack/loremipsum/src",
              "Larapack\\Repository\\": "packages/larapack/repository/src"
        },
        "files": [
            "app/Lib/helpers.php"
        ]
    },
    "autoload-dev": {
        "classmap": [
            "tests/TestCase.php"
        ]
    },
    "scripts": {
        "post-root-package-install": [
            "php -r \"copy('.env.example', '.env');\""
        ],
        "post-create-project-cmd": [
            "php artisan key:generate"
        ],
        "post-install-cmd": [
            "Illuminate\\Foundation\\ComposerScripts::postInstall",
            "php artisan optimize --force"
        ],
        "post-update-cmd": [
            "Illuminate\\Foundation\\ComposerScripts::postUpdate",
            "php artisan ide-helper:generate",
            "php artisan ide-helper:meta",
            "php artisan larapack:dbtables",
            "php artisan optimize --force"
        ]
    },
    "config": {
        "preferred-install": "dist"
    }
}
```

Seldaek:
@rsdev000 can you also show the failure if you run the command with `-vvv`? That way we get a backtrace at least and other stuff that might help figure it out.

rsdev000:
@Seldaek yep

```
Exception trace:
 () at phar://C:/Windows/composer.phar/vendor/composer/semver/src/VersionParser.php:171
 Composer\Semver\VersionParser->normalize() at phar://C:/Windows/composer.phar/vendor/composer/semver/src/Semver.php:39
 Composer\Semver\Semver::satisfies() at C:\Users\Rui\AppData\Roaming\Composer\vendor\sllh\composer-versions-check\src\VersionsCheckPlugin.php:103
 SLLH\ComposerVersionsCheck\VersionsCheckPlugin::satisfiesComposerVersion() at C:\Users\Rui\AppData\Roaming\Composer\vendor\sllh\composer-versions-check\src\VersionsCheckPlugin.php:54
 SLLH\ComposerVersionsCheck\VersionsCheckPlugin->activate() at phar://C:/Windows/composer.phar/src/Composer/Plugin/PluginManager.php:230
 Composer\Plugin\PluginManager->addPlugin() at phar://C:/Windows/composer.phar/src/Composer/Plugin/PluginManager.php:203
 Composer\Plugin\PluginManager->registerPackage() at phar://C:/Windows/composer.phar/src/Composer/Plugin/PluginManager.php:255
 Composer\Plugin\PluginManager->loadRepository() at phar://C:/Windows/composer.phar/src/Composer/Plugin/PluginManager.php:79
 Composer\Plugin\PluginManager->loadInstalledPlugins() at phar://C:/Windows/composer.phar/src/Composer/Factory.php:357
 Composer\Factory->createComposer() at phar://C:/Windows/composer.phar/src/Composer/Factory.php:540
 Composer\Factory::create() at phar://C:/Windows/composer.phar/src/Composer/Console/Application.php:275
 Composer\Console\Application->getComposer() at phar://C:/Windows/composer.phar/src/Composer/Command/BaseCommand.php:53
 Composer\Command\BaseCommand->getComposer() at phar://C:/Windows/composer.phar/src/Composer/Command/UpdateCommand.php:97
 Composer\Command\UpdateCommand->execute() at phar://C:/Windows/composer.phar/vendor/symfony/console/Command/Command.php:259
 Symfony\Component\Console\Command\Command->run() at phar://C:/Windows/composer.phar/vendor/symfony/console/Application.php:844
 Symfony\Component\Console\Application->doRunCommand() at phar://C:/Windows/composer.phar/vendor/symfony/console/Application.php:192
 Symfony\Component\Console\Application->doRun() at phar://C:/Windows/composer.phar/src/Composer/Console/Application.php:197
 Composer\Console\Application->doRun() at phar://C:/Windows/composer.phar/vendor/symfony/console/Application.php:123
 Symfony\Component\Console\Application->run() at phar://C:/Windows/composer.phar/src/Composer/Console/Application.php:99
 Composer\Console\Application->run() at phar://C:/Windows/composer.phar/bin/composer:43
 require() at C:\Windows\composer.phar:25
```

Seldaek:
Ah ha, so it's failing in `SLLH\ComposerVersionsCheck\VersionsCheckPlugin`, it's a bug in the plugin and not composer, see https://github.com/Soullivaneuh/composer-versions-check/issues/40

Soullivaneuh:
@rsdev000 @Seldaek This PR should fix your issue: https://github.com/Soullivaneuh/composer-versions-check/pull/42

Soullivaneuh:
If you are already on `v2.0.0`, here is the way to upgrade to [`v2.0.1`](https://github.com/Soullivaneuh/composer-versions-check/releases/tag/v2.0.1) and get rid of the issue:

```bash
$ composer global update --no-plugins
Changed current directory to /home/sullivan/.composer
  - Removing sllh/composer-versions-check (v2.0.0)
  - Installing sllh/composer-versions-check (v2.0.1)
    Downloading: 100%         

Writing lock file
Generating autoload files
```

Soullivaneuh:
BTW @Seldaek, it would be great to implement this kind of system (version compatibility) directly on composer. What do you think?

rsdev000:
@Seldaek @Soullivaneuh  Thank you. Working as expected.

glen-84:
Shouldn't `--direct` be the default for `composer outdated`? Do people usually update sub-dependencies?

Seldaek:
@glen-84 IMO it's good to see all that's outdated because when you run update it'll typically update or if it doesn't you can then check why not and see if one of your deps needs fixing perhaps.

glen-84:
@Seldaek Hmm. It would only update sub-deps if they fell within the version range specified in the dependency, right? However, the outdated command would show updates outside that range if I'm not mistaken.

As an example, when I run `composer outdated`, I get:

```
nikic/fast-route            v0.8.0           v1.0.0           Fast request router for PHP
```

My dependency is on `zendframework/zend-expressive-fastroute`, which specifies a version of `^0.8.0` for `nikic/fast-route`. In other words, running `composer update` would not update `nikic/fast-route`, yet it's listed as `outdated`, and there isn't much that I can do about that.

I think in most cases users are interested in their own dependencies, and not sub dependencies (which they have less control over). For this reason I think that it might make sense for this to be the default, but I'd be interested to hear what others think.

stof:
@glen-84 depending on what the constraint is, it could or could not update it indeed. But this is true for root requirements too (except you can change those).
Knowing that a composer update could select a newer version of a dependency is still useful (to tell you to update your lock file).
And if you still get outdated packages just after an update, ``composer why-not`` is here to tell you what prevented the update (to ask the ZF team to update their requirement for instance)

Seldaek:
I prefer to see it if it's outdated so I can check out why it's not updating and if possible go nag ZF to allow 1.0, in some cases there might be security releases and whatnot that are important to update to and might still be blocked by one of your deps' requirements. I can see that it doesn't fit everyone though, and adding a `-d` isn't the end of the world I'd think :) Changing defaults now that it's in a stable release should be avoided really, and I don't think one default can please everyone anyway.

glen-84:
Okay, fair enough I guess.

Issue #5008, opened 2016-03-05
Thoughts on secure-http


mmelvin0:
Many of my organization's Composer dependencies are hosted on a GitLab instance behind our corporate firewall. It's plain HTTP and not externally accessible.

Since the update that changed `secure-http` to true by default, obviously that can't be accessed without changing that setting back to false. The problem is that many developers will set it to false globally and forget about it.

I think it could be more secure if one could white-list URLs (e.g. per-project) that can be accessed insecurely.

Thoughts/concerns/criticism?

curry684:
Have to say I agree. It's been ticking me off over the past few years that I've been 'forced' to secure internal applications that never leave the corporate network, like our vCenter console.

The way I see it there are 2 practical solutions:
 - We check the hostname of URLs, and if they resolve to known private IP ranges we ignore `secure-http`, or:
 - We add an 'http' flag to repositories to voluntarily override the requirement on a per-repository basis

Seldaek:
I don't really like this because it's easy to mess up even if you think you're safe in condition X, and because thanks to letsencrypt it's really easy to get free certs now.

We could for sure turn the secure-http bool into a hashmap of domain names or IPs that can be configured, similar to what we have for preferred-install.. But that's again complexity and all for something that ultimately I feel would be better fixed at the source.

curry684:
The problem is dat even Letsencrypt performs ownership validation as it's required for the SSL system as a whole to function. And thus do not issue certificates for `satis.company.intranet` and the like. So for smaller companies that means they have to maintain their own root CA, and distribute/install those with their employees, just to encrypt traffic that they're already privvy to and are the only ones with physical access to.

The current implementation means that the webdevelopment freelancer can't even run a local Satis as http://localhost/satis is also unacceptable. Good luck getting a certificate for `localhost`  ;)

mmelvin0:
What if we were to keep secure-http as-is (boolean, default true) and add an insecure-http-whitelist that defaults to empty, can be set globally or per-project, and is only consulted if secure-http is true?

The issue really is that in order to use ANY "insecure" resources (e.g. intranet or localhost as mentioned by curry684) you have to allow access to ALL insecure resources.

Seldaek:
Yes I can see the problem.. I guess I just wish it didn't exist ;)

mmelvin0:
I agree, it would be great if HTTP just didn't exist without HTTPS. Alas...

[...]

PR #5032, opened 2016-03-09
Centralize secure-http checking

curry684:
As discussed in #5008 this PR routes the `secure-http` check centrally, and checks whether the host is not on known private or reserved IP ranges.

Also reversed the protocol check - the previous implementations blacklisted `http://` hosts, while the new implementation whitelists `ssh:` and `https:` hosts. I'm not entirely sure this will work correctly with all Git use cases - review highly appreciated.

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -13,6 +13,7 @@
 namespace Composer;
 
 use Composer\Config\ConfigSourceInterface;
+use Composer\Downloader\TransportException;
 
 /**
  * @author Jordi Boggiano <j.boggiano@seld.be>
@@ -394,5 +395,29 @@ class Config
         }
 
         return false;
+    }
+
+    /**
+     * Validates that the passed URL is allowed to be used by current config, or throws an exception.
+     *
+     * @param string $url
+     */
+    public function prohibitUrlByConfig($url)
+    {
+        if (!$this->get('secure-http')) {
+            return;
+        }
+
+        // Parse the URL into its separate parts
+        $parsed = parse_url($url);
+        if (false === $parsed || !isset($parsed['scheme'])) {
+            // If the URL is malformed or does not contain a usable scheme it's not going to work anyway
+            return;
+        }
+
+        // Throw exception on known insecure protocols
+        if (in_array($parsed['scheme'], array('http', 'git', 'ftp', 'svn'))) {
+            throw new TransportException("Your configuration does not allow connections to $url. See https://getcomposer.org/doc/06-config.md#secure-http for details.");
+        }
     }
 }
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -260,20 +260,9 @@ class RemoteFilesystem
             $this->io->writeError("    Downloading: <comment>Connecting...</comment>", false);
         }
 
-        // Check for secure HTTP
-        if (
-            ($this->scheme === 'http' || substr($fileUrl, 0, 5) === 'http:')
-            && $this->config && $this->config->get('secure-http')
-        ) {
-            // Passthru unsecure Packagist calls to $hashed providers as file integrity is verified with sha256
-            if (substr($fileUrl, 0, 23) !== 'http://packagist.org/p/' || (false === strpos($fileUrl, '$') && false === strpos($fileUrl, '%24'))) {
-                // other URLs must fail hard
-                throw new TransportException(sprintf(
-                    'Your configuration does not allow connection to %s://%s. See https://getcomposer.org/doc/06-config.md#secure-http for details.',
-                    $this->scheme,
-                    $originUrl
-                ));
-            }
+        // Check for secure HTTP, but allow insecure Packagist calls to $hashed providers as file integrity is verified with sha256
+        if ((substr($fileUrl, 0, 23) !== 'http://packagist.org/p/' || (false === strpos($fileUrl, '$') && false === strpos($fileUrl, '%24'))) && $this->config) {
+            $this->config->prohibitUrlByConfig($fileUrl);
         }
 
         $errorMessage = '';
--- a/tests/Composer/Test/ConfigTest.php
+++ b/tests/Composer/Test/ConfigTest.php
@@ -13,6 +13,7 @@
 namespace Composer\Test;
 
 use Composer\Config;
+use Composer\Downloader\TransportException;
 
 class ConfigTest extends \PHPUnit_Framework_TestCase
 {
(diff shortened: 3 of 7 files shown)


Seldaek:
AFAIK the only relevant insecure protocols are http, git and svn. We could add ftp for safety. I feel like the whitelist will bite us in the ass because there are so many odd variants like svn+ssh, ssh2, git@ that lacks proto, .. it's just too easy to miss something.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -260,20 +260,9 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
...
-                // other URLs must fail hard
-                throw new TransportException(sprintf(
-                    'Your configuration does not allow connection to %s://%s. See https://getcomposer.org/doc/06-config.md#secure-http for details.',
-                    $this->scheme,
-                    $originUrl
-                ));
-            }
+        // Check for secure HTTP, but allow insecure Packagist calls to $hashed providers as file integrity is verified with sha256
+        if ((substr($fileUrl, 0, 23) !== 'http://packagist.org/p/' || (false === strpos($fileUrl, '$') && false === strpos($fileUrl, '%24'))) && $this->config) {
+            $this->config->prohibitUrlByConfig(sprintf('%s://%s', $this->scheme, $this->originUrl));
```
Seldaek: I think this should use $fileUrl here instead. originUrl isn't guaranteed to match the requested hostname, especially when handling redirects and such. It was ok to use in the error message but not for IP verifications.

tests/Composer/Test/ConfigTest.php
```diff
@@ -213,6 +214,43 @@ public function testGitDisabledByDefaultInGithubProtocols()
...
+            array('svn://192.168.0.1/trunk', true),
+            array('svn://1.2.3.4/trunk', false),
+            array('git@github.com:composer/composer.git', true),
+            array('git://5.6.7.8/git.git', false),
+            array('http://10.1.0.1/satis', true),
+            array('http://127.0.0.1/satis', true),
+            array('hg://user:pass@my.satis/satis', true),
+            array('\\myserver\myplace.git', true),
+            array('file://myserver.localhost/mygit.git', true),
+            array('file://example.org/mygit.git', false),
```
Seldaek: Would be nice to sort these in two pass/fail groups, makes it more readable what's valid and what not.

src/Composer/Util/RemoteFilesystem.php
curry684: Ah yeah that part was rather obscurely coded so I wasn't entirely sure I was patching it right. Will look into it  :+1:

hkdobrev:
I share the concerns of @Seldaek. I like the approach which @mmelvin0 suggested the most. Having a whitelist which is empty by default. The `secure-http` option could even be deprecated in favour of the whitelist. I guess it's up to you to decide whether the whitelist would support allowing all requests to be insecure with a wildcard like `"*"`.

curry684:
@hkdobrev the comment above is about a protocol whitelist, not a host whitelist. Right now this PR does not facilitate whitelisting publicly routed IPs and hosts in any way, and I'm very tempted to say it should remain that way. We're mainly concerned with private networks here.

src/Composer/Config.php
```diff
@@ -395,4 +396,35 @@ private function getComposerEnv($var)
...
+        }
+
+        // Ignore known secure schemes
+        if (in_array($parsed['scheme'], array('https', 'ssh'))) {
+            return;
+        }
+
+        // Resolve the host and validate it - unresolveable hosts are allowed to pass and fail on the network level
+        $ip = gethostbyname($parsed['host']);
+        if (filter_var($ip, FILTER_VALIDATE_IP, FILTER_FLAG_NO_PRIV_RANGE | FILTER_FLAG_NO_RES_RANGE)) {
```
Seldaek: So this allows private networks and reserved IPs by default.. I don't really like this. I think by default we should prevent everything. `secure-http: "except_private_network"` maybe? Meh. A whitelist would maybe be easier to handle and probably more secure tbh than just allowing all private IPs blindly.

src/Composer/Config.php
```diff
@@ -395,4 +396,35 @@ private function getComposerEnv($var)
...
+            return;
+        }
+
+        // Ignore known secure schemes
+        if (in_array($parsed['scheme'], array('https', 'ssh'))) {
+            return;
+        }
+
+        // Resolve the host and validate it - unresolveable hosts are allowed to pass and fail on the network level
+        $ip = gethostbyname($parsed['host']);
```
StormTide: I dont think this is a safe way to fall through here. gethostbyname is not immutable. There's probably a way to fuzz this such that sometimes it resolves one way and sometimes it resolves another. Eg one ns set to return 127.0.0.1 and another set to return a valid ip, and or set to servfail and resolve.. and so forth.

src/Composer/Config.php
Seldaek: BTW the symfony thing I meant https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/IpUtils.php#L62 which allows checking for netmasks, not just pre-defined private ranges. It uses filter as well though.

StormTide:
I'd vote against this PR. Enabling even a single HTTP host on an internet connected server, even if that host is a lan address, will allow an attacker with MITM privilege to run arbitrary code. This could lead to privilege escalation and the wider implications of configuring a permitted address are not immediately obvious to the average user. Enabling an override is /not/ limited to that server, but has the same net effect as secure-http = false does globally.

src/Composer/Config.php
curry684: `gethostbyname` was bugging me for 2 reasons mainly:
- it only works on ipv4
- it returns the original hostname if it could not be resolved for *any* reason

Especially the second one could leave loopholes admittedly. What do you suggest, retrieving the NS manually to resolve from there? That would still not mitigate an entirely compromised DNS infrastructure.

src/Composer/Config.php
StormTide: @Seldaek once you enable a single http source, it doesn't matter. The MITM attacker can impersonate the host name and ip address in a HTTP connection. The server might think its talking to 192.168. whatever, but its really talking to a MITM.

StormTide: @curry684 if you want a whitelist for ip, then that ip has to be resolved before the checks, and made sure that the ip is used for all future connections. gethostbyname at each stage is not a safe way to do this.

Seldaek:
Yeah I don't know if there is really a point to this whole thing. It adds complexity for no real gain. If people want security then they should run everything secure.

Otherwise they can turn secure-http off. It'll still use secure URLs by default as everyone running with it enabled will encourage all public repos to be configured securely. 

It's then only a problem for private repos, and those would end up being whitelisted anyway if people don't want to configure their servers securely. So simple flag or whitelist has kind of the same end result.

I think I'd vote close.

StormTide:
@Seldaek i agree, vote to close here too.

src/Composer/Config.php
curry684: But if we're being properly paranoid, shouldn't we also assume that an attacker capable of performing DNS poisoning and MitM on the LAN is quite likely also able to modify routing tables? And, taken to extremes, just modify data behind any SSL-protected vhost?

I think we should draw a line somewhere between 'user convenience' and 'security' that's still realistic. Once your entire LAN including DNS and routing is out of your control I think your Satis server is also already compromised and shouldn't be trusted with or without certificate.

StormTide: Package signing is being worked on. As for modifying routing tables, unless they can present a signed certificate for the hostname, wont do em any good. TLS is authenticated like that. Big jump from a local MITM (read any client on the same ARP network) to a filesystem level control on a webserver. Hard shell gooey center is not a 2016 technique.

curry684:
Well I didn't call it proposal for nothing hehe. I would still like to keep the core changes in place though as it deduplicates code from 6 spread out places and adds proper testing - the whitelisting part is only a few lines.

curry684:
I renamed the PR to illustrate what it does now - centralizing the check.

Seldaek:
@curry684 thanks, would you mind squashing commits though? Because it introduces really dodgy history in this case with stuff that got removed entirely.

curry684:
Good point, now to discover how to rebase pushed commits, never done that  :smile:

Seldaek:
Doesn't matter if pushed or not :) Just `rebase -i origin/master` to squash as usual and then `push --force`

curry684:
Worked fine, thanks  :+1:

PR #5035, opened 2016-03-10
[RFC] Add an update-all request when there is no lock file

Seldaek:
Ping @naderman

I tried to reproduce https://twitter.com/RomainMonceau/status/707992176517640192 (update and install have different results, no lock file in any case) and indeed the [composer.json](http://pastebin.com/7EifAtSb) installs correctly when using `composer update`, but it fails when doing `composer install` with the following:

```

  Problem 1
    - akeneo/pim-community-dev v1.5.1 requires doctrine/doctrine-bundle 1.2.0 -> satisfiable by doctrine/doctrine-bundle[v1.2.0].
    - akeneo/pim-community-dev v1.5.0 requires doctrine/doctrine-bundle 1.2.0 -> satisfiable by doctrine/doctrine-bundle[v1.2.0].
    - Conclusion: don't install doctrine/doctrine-bundle v1.2.0
    - Installation request for akeneo/pim-community-dev 1.5.* -> satisfiable by akeneo/pim-community-dev[v1.5.1, v1.5.0].
```

The only meaningful difference I found between update and install when there is no lock file is that update sends an update-all in the solver request. Effectively this creates update rules for all installed packages, which with a clean vendor dir means:

```
jml/foobar-9999999-dev      (this is the root)
composer-plugin-api-1.0.0.0
php-7.0.3.0
php-64bit-7.0.3.0
ext-bcmath-7.0.3.0
ext-calendar-7.0.3.0
ext-ctype-7.0.3.0
ext-date-7.0.3.0
ext-filter-7.0.3.0
ext-hash-1.0.0.0
ext-iconv-7.0.3.0
ext-json-1.4.0.0
ext-mcrypt-7.0.3.0
ext-spl-7.0.3.0
ext-pcre-7.0.3.0
ext-reflection-7.0.3.0
ext-session-7.0.3.0
ext-mysqlnd-0.0.0.0
ext-tokenizer-7.0.3.0
ext-zip-1.13.0.0
ext-zlib-7.0.3.0
ext-libxml-7.0.3.0
ext-dom-20031129
ext-pdo-7.0.3.0
ext-openssl-7.0.3.0
ext-....
```

I don't really see why this has any influence at all, but it seems to fix a problem.. so is there any objection? And also why the heck is this happening is a good question :)

----

As a sidenote, we should just simplify all the code and treat an install w/o lock file as an update and be done with it, which would also fix this. I'll update the PR later.

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -171,6 +171,11 @@ class Installer
         gc_collect_cycles();
         gc_disable();
 
+        // Force update if there is no lock file present
+        if (!$this->update && !$this->locker->isLocked()) {
+            $this->update = true;
+        }
+
         if ($this->dryRun) {
             $this->verbose = true;
             $this->runScripts = false;
@@ -196,10 +201,10 @@ class Installer
 
         // create installed repo, this contains all local packages + platform packages (php & extensions)
         $localRepo = $this->repositoryManager->getLocalRepository();
-        if (!$this->update && $this->locker->isLocked()) {
-            $platformOverrides = $this->locker->getPlatformOverrides();
-        } else {
+        if ($this->update) {
             $platformOverrides = $this->config->get('platform') ?: array();
+        } else {
+            $platformOverrides = $this->locker->getPlatformOverrides();
         }
         $platformRepo = new PlatformRepository(array(), $platformOverrides);
         $repos = array(
@@ -261,7 +266,7 @@ class Installer
 
         if (!$this->dryRun) {
             // write lock
-            if ($this->update || !$this->locker->isLocked()) {
+            if ($this->update) {
                 $localRepo->reload();
 
                 // if this is not run in dev mode and the root has dev requires, the lock must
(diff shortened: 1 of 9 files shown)


naderman:
Don't object to the updateAll, install without lock should do the same as update anyway. I really can't tell what the issue is either, there seems to be some piece of info missing from the problem description, too.

jmlrx:
> there seems to be some piece of info missing from the problem description

If you want more info, i'll be glad to help. :)

jmleroux:
> there seems to be some piece of info missing from the problem description

If you want more info, i'll be glad to help.

naderman:
@jmleroux I was refering to the problem description generated by composer ;-)

fitn:
@Seldaek So if I well understood the fix, it explains why composer install was working with vendor but not with a clean vendor directory! Right?

Seldaek:
*Maybe* :) I am not sure what's going on here to be honest, it's a kinda weird bug, but we'll clean up some code and then it should always do the same at least, and hopefully that's always the correct thing ;)

Seldaek:
Updated the PR to force updates when an install has no lock file. Looks good now I think? Any feedback?

naderman:
@Seldaek looks good to me :-)

PR #5037, opened 2016-03-11
Add missing "secure-http" config property

criolit:
Recently introduced "secure-http" configuration property is available in documentation [https://getcomposer.org/doc/06-config.md#secure-http](https://getcomposer.org/doc/06-config.md#secure-http) is missing in the schema.


curry684:
:+1:

Issue #2565, opened 2014-01-08
Failure if unzip is not installed (as happens on a clean Ubuntu 13.10 install)


roderik:
## Short situation:

I was trying to do a create-project using composer on a Ubuntu 13.10 clean vm. PHP 5.5 installed using Chef with https://github.com/Kunstmaan/chef-php/blob/master/recipes/package.rb (irrelevant for the issue) that mostly just does apt-get install php5-cgi php5 php5-dev php5-cli php-pear, ext-zip is built in.

## The issue:

```
  - Installing doctrine/common (2.3.0)
Downloading https://api.github.com/repos/doctrine/common/zipball/d1c7d4334e38cad603a5c863d4c7b91bb04ec6b2
    Downloading: 100%         
Writing /home/vagrant/.composer/cache/files/doctrine/common/d1c7d4334e38cad603a5c863d4c7b91bb04ec6b2.zip into cache
    Extracting archive
Executing command (CWD): unzip '/home/projects/test/data/test/vendor/doctrine/common/a259fc31d4f248f87439a5515c6aa490' -d 'vendor/composer/25117f89' && chmod -R u+w 'vendor/composer/25117f89'
Reading ./composer.json


                                                             
  [RuntimeException]                                         
  There was an error extracting the ZIP file. Corrupt file?  
                                                             


Exception trace:
 () at phar://composer.phar/bin/../src/../src/Composer/Downloader/ZipDownloader.php:76
 Composer\Downloader\ZipDownloader->extract() at phar://composer.phar/bin/../src/../src/Composer/Downloader/ArchiveDownloader.php:43
 Composer\Downloader\ArchiveDownloader->download() at phar://composer.phar/bin/../src/../src/Composer/Downloader/DownloadManager.php:182
 Composer\Downloader\DownloadManager->download() at phar://composer.phar/bin/../src/../src/Composer/Installer/LibraryInstaller.php:155
 Composer\Installer\LibraryInstaller->installCode() at phar://composer.phar/bin/../src/../src/Composer/Installer/LibraryInstaller.php:86
 Composer\Installer\LibraryInstaller->install() at phar://composer.phar/bin/../src/../src/Composer/Installer/InstallationManager.php:153
 Composer\Installer\InstallationManager->install() at phar://composer.phar/bin/../src/../src/Composer/Installer/InstallationManager.php:140
 Composer\Installer\InstallationManager->execute() at phar://composer.phar/bin/../src/../src/Composer/Installer.php:510
[...]

PR #5038, opened 2016-03-11
Prefer ZipArchive over unzip on Windows

nicolas-grekas:
b945fc4d83 introduced a performance regression on Windows: unzip is way slower than the native ZipArchive class. I confirmed this on Linux also.

~~This PR makes composer try unzipping with ZipArchive, then unzip, then 7z on Windows (bonus).~~

This PR reverts to the previous behavior on Windows (prefer ZipArchive).

### DIFF
--- a/src/Composer/Downloader/ZipDownloader.php
+++ b/src/Composer/Downloader/ZipDownloader.php
@@ -48,7 +48,18 @@ class ZipDownloader extends ArchiveDownloader
         }
 
         if (!class_exists('ZipArchive') && !self::$hasSystemUnzip) {
-            throw new \RuntimeException('The zip extension and unzip command are both missing, skipping');
+            // php.ini path is added to the error message to help users find the correct file
+            $iniPath = php_ini_loaded_file();
+
+            if ($iniPath) {
+                $iniMessage = 'The php.ini used by your command-line PHP is: ' . $iniPath;
+            } else {
+                $iniMessage = 'A php.ini file does not exist. You will have to create one.';
+            }
+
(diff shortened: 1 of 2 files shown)


src/Composer/Downloader/ZipDownloader.php
```diff
@@ -75,36 +115,7 @@ protected function extract($file, $path)
             }
         }
 
-        if (!class_exists('ZipArchive')) {
```
nicolas-grekas: this case is dead code currently, I moved it to the early skipping exception

curry684:
There were more reasons to switch to unzip as default: https://github.com/composer/composer/issues/2565. In short, using the PHP ZIP extension, file permissions were not (always) preserved correctly, and symlinks could screw up.

What kind of performance regression are we talking about?

curry684:
There's also some documentation related to this that would need to be changed: https://github.com/curry684/composer/commit/379d7691326304b91c250faabbf252f9e257c12f

nicolas-grekas:
@curry684 thanks for the link. Since none of these issues apply on Windows (ZIP that contains permissions/symlinks), I updated the PR to use ZipArchive on Windows.
The perf regression is visible when doing a composer install with a cache already filled. Installing eah package is way slower on Windows since end of February (b945fc4)

nicolas-grekas:
Doc updated

curry684:
Cheers, :+1: from me now, just curious how big the file corruption issue was, but as it has fallbacks now I don't see an issue with that.

Seldaek:
For me an install of symfony/symfony (which is quite large) went from 76s to 62s, not huge improvement but a bit better :)

expedientesxero-source (approved):

PR #5039, opened 2016-03-11
Default to dist for all versions

barryvdh:
Right now, when the version is dev, a `source` install is made. But `dist` installs are also available, with faster installs (and cacheable). This could be useful for making git changes, but the `--prefer-source` option is available for that.
Also, setting the minimum stability in composer.json doesn't actually guarantee that it will use a `dev` version. For example:

```
{
    "require": {
        "symfony/process": "^3.0",
        "paragonie/random_compat": "^1.2"
    },
    "minimum-stability": "dev"
}
```

This will install (at the time of writing) the `dev-master` source version for `process`, but a stable dist for `random_compat` (because he doesn't have a branch-alias setup to match a dev version).

This change will make the `auto` selector default to `--prefer-dist` behaviour.

### DIFF
--- a/src/Composer/Downloader/DownloadManager.php
+++ b/src/Composer/Downloader/DownloadManager.php
@@ -317,6 +317,6 @@ class DownloadManager
             }
         }
 
-        return $package->isDev() ? 'source' : 'dist';
+        return 'dist';
     }
 }


Seldaek:
I don't agree with this. The reason we prefer source for dev packages is that if you require dev stuff they tend to update more often as they update with every commit, and doing updates with a clone is very fast, usually faster than downloading a whole copy of the code again. Plus polluting your cache with a bunch of dist archives from random commits is kinda wasteful, better keep versions in there.

Obviously for CI cloning every time might not be good so if you do have dev deps you might wanna --prefer-dist, or just `composer config --global preferred-install dist` to make it default for you machine-wide.

barryvdh:
Okay sounds reasonable, thanks!

PR #5055, opened 2016-03-14
OAuth Support for Bitbucket

wenkepaul:
I used existing github-oauth code as a template to develop bitbucket-oauth functionality.  Users can add consumer-key and consumer-secret to auth.json.  Instructions on how to obtain the consumer keys have been added to the documentation.


src/Composer/Util/Bitbucket.php
```diff
@@ -0,0 +1,184 @@
...
+     */
+    public function authorizeOAuth($originUrl)
+    {
+        if (!in_array($originUrl, $this->config->get('bitbucket-domains'))) {
+            return false;
+        }
+
+        // if available use token from git config
+        if (0 === $this->process->execute('git config bitbucket.accesstoken', $output)) {
+            $this->io->setAuthentication($originUrl, trim($output), 'x-oauth-basic');
```
Seldaek: This seems like it won't work for bitbucket?

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -672,6 +672,11 @@ protected function getOptionsForUrl($originUrl, $additionalOptions)
                 $authStr = base64_encode($auth['username'] . ':' . $auth['password']);
                 $headers[] = 'Authorization: Basic '.$authStr;
             }
+
+            if ('bitbucket.org' === $originUrl) {
+                $options['http']['method'] = 'POST';
+                $options['http']['content']['grant_type'] = 'client_credentials';
+            }
```
Seldaek: This seems like it doesn't belong here, is it only needed for the initial retrieval of a token from the oauth secret?

Seldaek:
Is the bitbucket-domains config really needed? It seems to me like this stuff only works for "bitbucket cloud" which is bitbucket.org. I don't have a bitbucket server to check though so no idea.

wenkepaul:
I've never used Atlassian's Stash (now called Bitbucket Server), so I'm not sure if they allow different domains.  I can remove the config for now, and add it back later if it's requested.

src/Composer/Util/RemoteFilesystem.php
wenkepaul: Correct, it's only needed to retrieve the token.

src/Composer/Util/Bitbucket.php
wenkepaul: I'll double check this.

src/Composer/Util/RemoteFilesystem.php
Seldaek: OK then it would be better to pass it in the $options to the RemoteFilesystem->getContent call where it is relevant than having it there always as it might well have side effects to other requests done to bitbucket

src/Composer/Config.php
```diff
@@ -27,6 +27,7 @@ class Config
         'preferred-install' => 'auto',
         'notify-on-install' => true,
         'github-protocols' => array('https', 'ssh', 'git'),
+        'bitbucket-protocols' => array('https'),
```
stof: do we need it ? Are they allowing to clone with another protocol ?

src/Composer/Config/JsonConfigSource.php
```diff
@@ -81,7 +81,7 @@ public function addConfigSetting($name, $value)
     {
         $authConfig = $this->authConfig;
         $this->manipulateJson('addConfigSetting', $name, $value, function (&$config, $key, $val) use ($authConfig) {
-            if (preg_match('{^(github-oauth|gitlab-oauth|http-basic|platform)\.}', $key)) {
+            if (preg_match('{^(github-oauth|gitlab-oauth|http-basic|platform|bitbucket)\.}', $key)) {
```
stof: should be ``bitbucket-oauth`` actually

src/Composer/Util/Bitbucket.php
```diff
@@ -0,0 +1,184 @@
...
+
+namespace Composer\Util;
+
+use Composer\Factory;
+use Composer\IO\IOInterface;
+use Composer\Config;
+use Composer\Downloader\TransportException;
+
+/**
+ * I used GitHub.php as a template.
```
stof: This does not need to be here IMO

src/Composer/Util/Bitbucket.php
```diff
@@ -0,0 +1,184 @@
...
+ * @author Paul Wenke <wenke.paul@gmail.com>
+ */
+class Bitbucket
+{
+    protected $io;
+    protected $config;
+    protected $process;
+    protected $remoteFilesystem;
+    protected $token = array();
```
stof: I suggest using private visibility. It makes maintenance much easier in a semver context (as we don't need to preserve any BC for private stuff as it is purely internal)

curry684:
We're running a pretty out of date Stash 3.3 here (still need to plan an update to Bitbucket Server hehe) but I can't find OAuth anywhere in the configs. Don't think it's much use either behind the firewall...? All I can find on Google with a quick search refers to Bitbucket Cloud as well.

Let me know if there's anything I can test with our Stash instance though.

src/Composer/Util/Bitbucket.php
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * @param string $originUrl
+     * @return bool
+     */
+    private function requestAccessToken($originUrl)
+    {
+        try {
+            $apiUrl = 'bitbucket.org/site/oauth2/access_token';
+
+            $json = $this->remoteFilesystem->getContents($originUrl, 'https://'.$apiUrl, false, array(
```
stof: ``https://`` should be in the apiUrl variable directly IMO here

src/Composer/Util/Bitbucket.php
```diff
@@ -0,0 +1,184 @@
...
+    {
+        if ($message) {
+            $this->io->writeError($message);
+        }
+
+        $note = 'Composer';
+        if ($this->config->get('bitbucket-expose-hostname') === true && 0 === $this->process->execute('hostname', $output)) {
+            $note .= ' on ' . trim($output);
+        }
+        $note .= ' ' . date('Y-m-d Hi');
```
stof: The ``$note`` variable looks unused

tests/Composer/Test/Util/BitbucketTest.php
```diff
@@ -0,0 +1,137 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Test\Util;
+
+use Composer\Util\Bitbucket;
+
+/**
+ * I used GitHubTest.php as a template.
```
stof: This comment is not necessary

stof:
> Is the bitbucket-domains config really needed? It seems to me like this stuff only works for "bitbucket cloud" which is bitbucket.org. I don't have a bitbucket server to check though so no idea.

The OAuth logic always performs the OAuth calls on the bitbucket.org API in this API, so it does not support any other domain

wenkepaul:
I completed the updates per the conversations above.

src/Composer/Command/ConfigCommand.php
```diff
@@ -491,6 +491,17 @@ function ($vals) {
             return;
         }
 
+        // handle bitbucket-oauth
+        if (preg_match('/^(bitbucket-oauth)\.(.+)/', $settingKey, $matches)) {
+            if (2 !== count($values)) {
+                throw new \RuntimeException('Excepted two arguments (consumer-key, consumer-secret), got '.count($values));
```
smasty: There's a typo in the exception message. Should be _Expected_, not _Excepted_.

Seldaek:
alrighty, thanks @wenkepaul

PR #5059, opened 2016-03-15
simplified existing test, added a new test - related to #4795

alcohol:
Reproduces (and solves) an issue reported by @teohhanhui on the initial "fix" (#4817) for #4795.


alcohol:
@naderman can you let me know if this solution is good enough?

naderman:
Yeah this should be fine :-)

naderman:
@alcohol did you just remove c because it made no difference or was the test actually failing with c still there?

alcohol:
I removed `c` cause it made no difference for the test.

PR #5068, opened 2016-03-15
Add latest version to show + Deprecated message

barryvdh:
As request, a separate PR, following up on https://github.com/composer/composer/pull/5028 with the addition of a deprecation notice.

Screenshot:
screen shot 2016-03-15 at 20 48 01

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -243,11 +243,13 @@ EOT
                 foreach ($packages[$type] as $package) {
                     if (is_object($package)) {
                         $nameLength = max($nameLength, strlen($package->getPrettyName()));
-                        $versionLength = max($versionLength, strlen($package->getFullPrettyVersion()));
-                        if ($showLatest) {
-                            $latestPackage = $this->findLatestPackages($package->getName());
-                            $latestPackages[$package->getPrettyName()] = $latestPackage;
-                            $latestLength =  max($latestLength, strlen($latestPackage->getFullPrettyVersion()));
+                        if ($showVersion) {
+                            $versionLength = max($versionLength, strlen($package->getFullPrettyVersion()));
+                            if ($showLatest) {
+                                $latestPackage = $this->findLatestPackages($package->getName());
+                                $latestPackages[$package->getPrettyName()] = $latestPackage;
+                                $latestLength =  max($latestLength, strlen($latestPackage->getFullPrettyVersion()));
+                            }
                         }
                     } else {
                         $nameLength = max($nameLength, $package);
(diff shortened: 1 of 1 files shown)


barryvdh:
Although I'm not really sure I'd the message appears when the package if deprecated after the last update. Do I need to the the last package version, instead of the one from the installed pool?

barryvdh:
It now uses the latest package information to show deprecation messages. Also used an array to cache the latest package version, so it can be used to determine the actual version length without searching them twice.

src/Composer/Command/ShowCommand.php
```diff
@@ -334,22 +381,27 @@ protected function getPackage(RepositoryInterface $installedRepo, RepositoryInte
...
     {
         $io = $this->getIO();
         $io->write('<info>name</info>     : ' . $package->getPrettyName());
         $io->write('<info>descrip.</info> : ' . $package->getDescription());
         $io->write('<info>keywords</info> : ' . join(', ', $package->getKeywords() ?: array()));
         $this->printVersions($package, $versions, $installedRepo);
+        if ($latestPackage) {
+            $io->write('<info>latest</info>   : ' . $latestPackage->getPrettyVersion());
+        } else {
+            $latestPackage = $package;
```
barryvdh: I added this to check for deprecations on the latest package available, otherwise check the 'old' info.

src/Composer/Command/ShowCommand.php
```diff
@@ -249,9 +262,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
                     $input->setOption('path', false);
                 }
 
+                $latestLength = $versionLength;
                 $writePath = !$input->getOption('name-only') && $input->getOption('path');
                 $writeVersion = !$input->getOption('name-only') && !$input->getOption('path') && $showVersion && ($nameLength + $versionLength + 3 <= $width);
-                $writeDescription = !$input->getOption('name-only') && !$input->getOption('path') && ($nameLength + ($showVersion ? $versionLength : 0) + 24 <= $width);
+                $writeLatest = !$input->getOption('name-only') && !$input->getOption('path') && $showLatest && ($nameLength + ($showVersion ? $versionLength : 0) + $latestLength + 3 <= $width);
+                $writeDescription = !$input->getOption('name-only') && !$input->getOption('path') && ($nameLength + ($showVersion ? $versionLength : 0) + ($showLatest ? $latestLength : 0) + 24 <= $width);
```
greg0ire: These lines are a bit long.

barryvdh: Yes, but I just followed the coding style here.

src/Composer/Command/ShowCommand.php
```diff
@@ -260,9 +275,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                        if ($writeLatest) {
+                            $latestVersion = $this->findBestVersionForPackage($package->getName());
+                            $type = $latestVersion == $package->getFullPrettyVersion() ? 'info' : 'comment';
+                            $io->write(' <'.$type.'>' . str_pad($latestVersion, $latestLength, ' ') . '</'.$type.'>', false);
+                        }
+
                         if ($writeDescription) {
                             $description = strtok($package->getDescription(), "\r\n");
-                            $remaining = $width - $nameLength - $versionLength - 4;
+                            $remaining = $width - $nameLength - $versionLength -  ($writeLatest ? $latestLength : 0) - 4;
```
greg0ire: This line is too long, and (not your fault), there is a magic number. Why `4`? Should it be (strlen(' ...'))?

src/Composer/Command/ShowCommand.php
greg0ire: > To ensure a consistent code base, you should make sure the code follows the PSR-2 Coding Standards.

PSR-2 recommends shorter lines.

> Yes, but I just followed the coding style here.

By that logic, if you need to add, say 36 more variables, you will end up with a very long line. But surely there must be some kind of limit even the previous developer wouldn't break, isn't there?

barryvdh: I've reduced the line length by keeping the lengths 0 when not being written + checking if the version is writting, instead of the same input options.

src/Composer/Command/ShowCommand.php
barryvdh: Not sure. It was there before..

greg0ire: That's why I said "not your fault", but the fact that neither of us is sure shows how this is a bad thing :P The minus 3 below sure corresponds to strlen('...'), but this one is less obvious.

Seldaek:
And merged too, thanks :)

PR #4010, opened 2015-05-07
Add GitHub compare links to CHANGELOG.md

joshdifabio:
Compare links added to CHANGELOG.md, a la [keepachangelog.com](http://keepachangelog.com).

This is just an idea, and it only took a couple minutes, so feel free to reject it.

PR #5077, opened 2016-03-17
Add compare link to changelog for beta1

barryvdh:
Not entirely useful because the changes are too big to show, but this follows the line set in #4010 and keeps consistency with existing links.


Seldaek:
Oh yeah I always forget those, thanks

PR #5093, opened 2016-03-23
Silently correct _empty_ namespaces, fixes #5092

### LINKED ISSUE #5092: handle json_encode _empty_ gracefully please
So I just ran into this problem while making a satis repository script for composer. I don't see a good solution. The problem has to do with `""` keys getting turned into `"_empty_"` using json_encode when provided an object. 

There is a bug on php.net about this, but it sounds like DONTFIX-ACTUALLYAFEATURE: https://bugs.php.net/bug.php?id=46600

If I `$composerJson = json_decode($composerJson, true)` into associative arrays, manipulate it, then `json_encode($composerJson)` then my 
  `"require-dev": {}`
turns into
  `"require-dev": []`
which composer chokes on. So obviously I can't read the composer file as associative arrays! That isn't this bug, I just want to show upfront that using the associative array parameter on my json_decode cannot solve my problem.

So I do: `$composerJson = json_decode($composerJson, false)` and turn it into objects to manipulate it. Except the most common composer.json contains something like this:
```
  "autoload": {
    "psr-4": {
      "": "src/"
    }
  }
```

it reads it in fine, but when I do `json_encode($composerJson)` again... I get:
```
  "autoload": {
    "psr-4": {
      "_empty_": "src/"
    }
  }
```

which composer then chokes on because of course "_empty_" has a different meaning than "", and it thinks I mean the namespace `_empty_` AND it's missing the mandatory trailing `\\` slashes..

So I'm stuck doing a json_encode then using a hack, `$composerJson = str_replace('"_empty_":', '"":', $composerJson)` before writing it to file again.

The _empty_ key does not seem to me to be the right answer here. Should composer change the way composer files for virtually all projects be written? Is this a composer bug?

Or should json_encode and json_decode objects reproduce a blank key? Will this ever be fixed upstream or should composer just implement the hack for me?

**Composer could replace the "_empty_" into "" for me when it reads it in a composer.json**

I hope upstream would fix, but I doubt it. I read somewhere composer isn't getting much development, so may it will not get fixed here either. For now I have my hack.

curry684:
See #5092, bit of a hack to fix it on this end but it's an invalid namespace anyway.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -748,6 +748,10 @@ FOOTER;
                         continue;
                     }
 
+                    // Patch corrupted namespaces through external manipulation
+                    if ($namespace = '_empty_') {
+                        $namespace = '';
+                    }
                     $autoloads[$namespace][] = $relativePath;
                 }
             }
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -749,7 +749,7 @@ FOOTER;
                     }
 
                     // Patch corrupted namespaces through external manipulation
-                    if ($namespace = '_empty_') {
+                    if ($namespace === '_empty_') {
                         $namespace = '';
                     }
                     $autoloads[$namespace][] = $relativePath;


alcohol:
Eh.. this is the functional programmer side of me speaking I guess; but I personally would rather assume happy path always. If input is corrupt or bad, then just let things explode, like they should.

curry684:
I agree, the fix is ugly, but it serves a real world case regrettably as I've also run into this already hehe.

The bad thing is also: things don't explode until runtime. The '_empty_` autoloaders are just generated, causing confusion at runtime instead of install time. That's the only reason I think this oneliner warrants inclusion.

stof:
I would rather detect such case in ``composer validate`` to report the file as invalid, allowing you to know that you have to fix your composer.json file

PR #5100, opened 2016-03-23
Extract binary installation and removal to own class

helhum:
The functionality to install binaries might be useful
for other installers.
Create API for that by extracting this functionality from
the LibraryInstaller class.


helhum:
Please let me know if this is interesting and reasonable.

If so, I'm happy to rework this until it can be merged

Seldaek:
Seems sensible, thanks :) I just reorg'd the constructor arguments to have IO first as that is the case across most classes. Also renamed the class to be a bit shorter but no big deal.

Issue #5094, opened 2016-03-23
Composer install stuck at writing something into cache and here're something i found


HuNanZai:
With the following composer.json:

```json
{
        "require": {
		"php": ">=5.2.4",

		"rmccue/requests": "~1.6.1",
                "hunanzai/log": "~1.1.0",
                "laravelbook/ardent": "~2.4.2"
	},

	"repositories": [
		{
			"type": "vcs",
			"url":  "https://github.com/HuNanZai/log"
		}
	],

	"config": {
		"secure-http": false
	}
}
```

When I run this command:

```
composer install -vvv (please include -vvv!)
```

I get this output:
(the composer just stuck at some step like:)

```
[100.4MB/9.47s] Reading /home/ci123dev/.composer/cache/repo/http---packagist.phpcomposer.com/provider-symfony$config.json from cache

```

Then after google some related questions and trial, i found this:
```
Test A:
1. remove "hunanzai/log": "~1.1.0",
2. run composer install -vvv
3. OK
```
```
Test B:
1. remove "laravelbook/ardent": "~2.4.2"
2. run composer install -vvv
3. OK
````
So i look a little further, this problem may happened in the dependency version analysis "symfony/http-foundation"

```
hunanzai/log: ~1.1.0
        symfony/http-foundation: ~2.0.4

laravelbook/ardent: ~2.4.2
        illuminate/validation: 4.x
                symfony/http-foundation: 2.3.*
```
But i don't know exactly why this would happen and how to prevent from getting this error again.
Ps: the stucking installing step doesn't show anything related to the real problem.

curry684:
The described issue appears similar or even identical to https://github.com/composer/composer/issues/5084. What I found there is that it wasn't just so much infinitely looping as "incredibly freaking slow". The `Reading from` cache log line is misleading as it happens to be the last package being loaded, and nothing is logged before SAT solving is started after that.

Seldaek:
Confusing logging fixed by #5108.

edcarlos-92:
somehow enabling extension=zip in php.ini got this issue resolved

Issue #5084, opened 2016-03-20
Stuck at provider-symfony$cache when installing/updating


calinpristavu:
The scenario is that I want to deploy an oro/platform to an ec2 instance with 4gb of ram.
The app was previously deployed (not today) locally without any headaches.

Now, both locally and on the ec2, install/update both get stuck at 
`Reading /home/ubuntu/.cache/composer/repo/https---packagist.org/provider-symfony$cache.json from cache`

CPU stays at 100%, RAM ~2200 mb

With the following composer.json:

```json
{
    "name": "oro/platform-application",
    "description": "Oro Platform Empty Application",
    "homepage": "https://github.com/orocrm/platform-application.git",
    "license": "MIT",
    "autoload": {
        "psr-0": {
            "": "src/"
        }
    },
    "repositories": [
        {
            "type": "composer",
            "url": "https://packagist.orocrm.com"
        }
    ],
    "require": {
        "oro/platform": "1.9.0"
    },
    "require-dev": {
        "sensio/generator-bundle": "2.5.3"
    },
    "config": {
        "component-dir": "web/bundles/components"
    },
    "scripts": {
        "post-install-cmd": [
            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
            "Oro\\Bundle\\InstallerBundle\\Composer\\ScriptHandler::setPermissions",
            "Oro\\Bundle\\InstallerBundle\\Composer\\ScriptHandler::installAssets",
            "Oro\\Bundle\\InstallerBundle\\Composer\\ScriptHandler::setAssetsVersion"
        ],
        "post-update-cmd": [
            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
            "Oro\\Bundle\\InstallerBundle\\Composer\\ScriptHandler::setPermissions",
            "Oro\\Bundle\\InstallerBundle\\Composer\\ScriptHandler::installAssets",
            "Oro\\Bundle\\InstallerBundle\\Composer\\ScriptHandler::setAssetsVersion"
        ]
[...]

PR #5108, opened 2016-03-24
Add extra logging before and after SAT solving

curry684:
As both #5084 and #5094 are confusing issue reports to both users and developers misinterpreting the last line of logging before it 'hangs'.


Seldaek:
Thanks

Issue #5034, opened 2016-03-10
[question] Install fail, but update is successfull


jmleroux:
Hello,

I thought that `composer install` was the same thing as `composer update` when there is no composer.lock file.

But it seems i was wrong. :(

I struggled to test a composer.json with the Akeneo project : 

```
{
    "name": "jml/foobar",
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/akeneo/pim-community-dev.git",
            "branch": "master"
        }
    ],
    "require": {
        "akeneo/pim-community-dev": "1.5.*",
        "doctrine/migrations": "1.0.0-alpha3@alpha"
    }
}
```

I created a folder with just this file and tried to install it with `composer install` and i always get the same error : 

> Loading composer repositories with package information
> Installing dependencies (including require-dev)                                           
> Your requirements could not be resolved to an installable set of packages.
> 
>   Problem 1
>     - akeneo/pim-community-dev v1.5.1 requires doctrine/doctrine-bundle 1.2.0 -> satisfiable by doctrine/doctrine-bundle[v1.2.0].
>     - akeneo/pim-community-dev v1.5.0 requires doctrine/doctrine-bundle 1.2.0 -> satisfiable by doctrine/doctrine-bundle[v1.2.0].
>     - Conclusion: don't install doctrine/doctrine-bundle v1.2.0
>     - Installation request for akeneo/pim-community-dev 1.5.* -> satisfiable by akeneo/pim-community-dev[v1.5.1, v1.5.0].

I have searched for bad dependencies, but didn't find them.
And finally, i just tried a `composer update` and it worked !

So what can be the difference between these two commands and why would `install` fail ?

Thanks for reading,

JM

fitn:
Just to link it with the PR: https://github.com/composer/composer/issues/5034

PR #5113, opened 2016-03-25
Install / Update script event changed, now documented

Slamdunk:
Since this commit https://github.com/composer/composer/commit/55b0ed8c8be74af348d9fec6702da3efd50feb46 the command events related to the `install` command are no longer tightened to the raw command `install`.

In my humple opinion this is a weird change, and I hope you will be able to merge https://github.com/composer/composer/issues/5034 needs to old doc behaviour.

But if you are not going to revert the expected behaviour, at least let's document it :smile:

### DIFF
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -22,11 +22,10 @@ Composer fires the following named events during its execution process:
 
 ### Command Events
 
-- **pre-install-cmd**: occurs before the `install` command is executed.
-- **post-install-cmd**: occurs after the `install` command has been executed.
-- **pre-update-cmd**: occurs before the `update` command is executed.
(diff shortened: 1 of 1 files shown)


stof:
what needs to be explained clearly is that running ``composer install`` without a lock file actually runs an update internally (and this is why it has been updated to run the update hooks too, to run the proper update fully).

@Seldaek what do you think about display a warning message in the output saying that composer switched to an update command ?

Seldaek:
I think clarifying the docs doesn't hurt. When running install w/o lock it outputs `Updating dependencies (including require-dev)` which kinda indicates it's doing an install, but maybe we should show a yellow warning line or something yes.. Not sure if it really is such a big deal once people will have fixed their scripts.

PR #5122, opened 2016-03-27
Warn if running as root, and clobber sudo credentials

### LINKED ISSUE #5119: dont execute anything with active `sudo session`
its common that an update of composer happens with sudo, as composer is installed in a place which is only changeable with higher privileges.

Its not uncommon, that an self-update is made directly before an install/updaten command.
To avoid easy privilege escalation in case of an evil attack (like a worm spreading trough the composer ecosystem) we should add a check for an active sudo session, and block further execution if found.

Such malware could spread silently as sudo allows `sudo -n true` which allows for testing for an active session, without triggering the typical interactive password input.

curry684:
Should fix #5119

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -133,6 +133,15 @@ class Application extends BaseApplication
                 $input->setInteractive(false);
             }
 
+            if (!Platform::isWindows() && function_exists('posix_getuid') && posix_getuid() === 0) {
+                $io->writeError('<warning>Running composer as root is highly discouraged as packages, plugins and scripts cannot always be trusted</warning>');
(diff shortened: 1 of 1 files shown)


Flyingmana:
does this also prevent self-update to run via sudo?

curry684:
No, it clobbers the credentials for the outer user during the session. So the current executable will continue just fine as root, and warn for running as root. The warning is shown on purpose just to keep you aware of what you're doing and that are trusting Composer's update mechanism there.

curry684:
Travis fail on 5.5 is the old parallel testing issue, not related.

Seldaek:
Thanks, I sprinkled some extra Silencer on top just in case because this stuff should definitely not make composer fail.

PR #5125, opened 2016-03-28
Add RSS channel to support section in composer schema

### LINKED ISSUE #5073: [schema] [feature] Adding feed option to schema
**Why**
To allow package authors to attach feeds to their composer.json in order to inform users of new releases, issues, security vulnerabilities, BC breaks, important discussions etc...

**Use cases**
I think each package author needed at some point to communicate infos to their uses, but the users are not constantly looking at your github issues or your twitter feed, they're busy making great software.
This is a way of telling them: "Hey check out this feed, it has all you need"

This is also very useful for wrapper libraries and SDKs, with vendor APIs constantly changing or SaaS providers changing their terms randomly (like Mandrill recently) .

**How**
I propose that we add a **feed** option to composer schema (the option name doesn't matter).
The option will be exactly like the **support** option, an array of channels you can use to get the recent news about the package.

This is currently possible by using the **extra** option or even the **support** option
but by adding a new one we can standardize and automate, It will be trivial, for example, to create a tool that scans your composer.json and displays all important changes about all your dependencies in a nice UI.
 

**An example**

    "feeds": {
        "rss": "https://getcomposer.org/feed",
        "twitter": "@composer"
    }

**alternative**

Instead of adding an option we can simply add channels to  **support**

    "support": {
        ...
        "rss": "https://getcomposer.org/feed"
    }

I think **rss** is enough for now, package authors can propose other ones if needed.

markatom:
Should fix #5073.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -238,6 +238,7 @@ Support information includes the following:
 * **irc:** IRC channel for support, as irc://server/channel.
 * **source:** URL to browse or download the sources.
 * **docs:** URL to the documentation.
+* **rss:** URL to the RSS feed.
 
 An example:
 
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -465,6 +465,11 @@
                     "type": "string",
                     "description": "URL to the documentation.",
                     "format": "uri"
+                },
+                "rss": {
+                    "type": "string",
+                    "description": "URL to the RSS feed.",
+                    "format": "uri"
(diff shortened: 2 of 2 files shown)


markatom:
@sealeg Sorry, I didn't notice. I think they should be updated. Should I create a new pull request?

Seldaek:
Yes please @markatom :) No huge deal but would be good to get full coverage there.

Issue #1209, opened 2012-10-15
Recreate missing binary symlinks/proxy files


Sgoettschkes:
If a binary symlink or binary proxy file is missing after install/update, it should be recreated.

For me this would be expected behaviour. If a binary symlink is missing for some reason, it should be there after an install or update. 

The reason I'm asking is because I deleted one of the binary files (composer-php) and pdepend could not be executed anymore. Executing composer once more could not solve the problem.

ccharles:
:+1:

I suppose that the work [started in April, 2013](https://github.com/composer/composer/issues/1209#issuecomment-17150581) by @Sgoettschkes has stalled?

Sgoettschkes:
Yeah, the task was a bit to much for me. I started working through the code base but could not really figure it out. Maybe I'll give it another try but not anytime soon. If somebody would like to take over, please do!
I am sorry I did not get back in time, I really forgot about it tbh.

fcastillo-es:
:+1:

mauricios:
+1

pfuhrmann:
+1

edvinasme:
@pfuhrmann I have tested this, and it's enough to remove only `vendor/` directory inside your project to get symlinks in `bin/` folder regenerated.

geshan:
this will be a good feature to have.

ethanpooley:
+1. In my case, while testing Vagrant on Windows hosts, there was a problem with the file share that resulted in no symlinks being created on the first run of `composer install`. Yet on the second run, after the problem was fixed, Composer thought everything was fine.

JakeWorrell:
+1 here too. I want to set up my deployment scripts to copy the previous vendor directory and then let composer recreate the bin directory when running `composer install`. Copying the previous vendor directory makes a massive difference to deployment times, but this bin directory issue causes us serious problems

Hendra-Huang:
+1

christopher-hopper:
Once this new feature is completed it would be good to have some `composer update` and `composer install` parameters or switches that allowed:

- Overwrite existing vendor binaries if there is a name conflict
- Force recreate of vendor binaries

I ran into this problem where, during a `composer install` the vendor binaries were not created for the project I was in because composer reported:

    Skipped installation of bin foobar for package foo/bar: name conflicts with an existing file

While your new feature will help to resolve this, it would be nice to also just force the overwrite using a command-line parameter or switch.

KamilKopaczyk:
[...]

PR #5127, opened 2016-03-29
Re-install binaries on update/install

j0k3r:
Binaries are re-installed after an update/install (ie: removed and then installed).
This is a working draft that fix https://github.com/composer/composer/issues/1209.

For the moment it removes all binaries links and re-create them on every update/install.

I have few questions:

1. Re-installing binaries are performed every time on an update or installation. Should I run it only when user re-define the bin dir? If yes, how can I detect that the config is overridden in the `composer.json` with `"bin-dir": "bin"` ?
2. In `InstallationManager` I retrieve the installer but since I only updated the `LibraryInstaller`, I try/catch to avoid other installer to throw exception. Is there a better way?
3. I also display a message when installing binaries: _Installing binaries files_ Is that necessary? Or should be displayed only in verbose mode?
4. Should we just install binaries (and then displaying messages when binaries files exists, ie _Skipped installation of bin BIN_FILES for package PACKAGE_NAME: name conflicts with an existing file_)) instead of removing/installing? Or do we have create a new option to force that removal?

### DIFF
--- a/src/Composer/Installer/InstallationManager.php
+++ b/src/Composer/Installer/InstallationManager.php
@@ -128,6 +128,22 @@ class InstallationManager
     }
 
     /**
+     * Install binary for the given package.
+     * If the installer associated to this package doesn't handle that function, it'll do nothing.
+     *
+     * @param PackageInterface $package Package instance
+     */
+    public function installBinary(PackageInterface $package)
+    {
+        try {
+            $installer = $this->getInstaller($package->getType());
+            $installer->installBinary($package);
+        } catch (\InvalidArgumentException $e) {
+            // the given installer doesn't support installing binaries
+        }
+    }
+
+    /**
      * Executes solver operation.
      *
      * @param RepositoryInterface $repo      repository in which to check
--- a/src/Composer/Installer/InstallationManager.php
+++ b/src/Composer/Installer/InstallationManager.php
@@ -137,9 +137,14 @@ class InstallationManager
     {
         try {
             $installer = $this->getInstaller($package->getType());
-            $installer->installBinary($package);
         } catch (\InvalidArgumentException $e) {
-            // the given installer doesn't support installing binaries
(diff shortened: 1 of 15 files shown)


bamarni:
Nice one 👍 

Would it remove everything inside the bin-dir? I'm wondering if it's safe to drop executables not managed by composer inside the bin-dir.

j0k3r:
@bamarni it just remove binaries from installed package only.

Wilkins:
+1

src/Composer/Installer/InstallationManager.php
```diff
@@ -128,6 +128,22 @@ public function isPackageInstalled(InstalledRepositoryInterface $repo, PackageIn
...
+     *
+     * @param PackageInterface $package Package instance
+     */
+    public function installBinary(PackageInterface $package)
+    {
+        try {
+            $installer = $this->getInstaller($package->getType());
+            $installer->installBinary($package);
+        } catch (\InvalidArgumentException $e) {
+            // the given installer doesn't support installing binaries
```
stof: this is not good. There is no ``installBinary`` method in the installer interface, so an installer not supporting it could trigger a fatal error. It won't trigger an InvalidArgumentException

what you need to do is to introduce a new interface containing this method, and checking the installer with ``instanceof``

j0k3r: All right, I've pushed the interface `InstallerBinaryInterface`.
Let me know.

Seldaek:
Thanks @j0k3r, I modified the patch so it doesn't uninstall=>install them every time, but only tries to reinstall, because otherwies it causes potential problems where if you already have a file called like that for some reason it would wipe it and replace it with the composer one, which we don't do usually on install.

j0k3r:
Thanks Jordi, seems to be the best solution.

Issue #4494, opened 2015-10-11
Get extra options for plugin


Soullivaneuh:
Is there a way to get options from the `extra` key like we do for composer scripts but for plugins?

If yes, how? I have to admit I am a bit lost for this.

Concrete need: https://github.com/Soullivaneuh/composer-versions-check/issues/13#issuecomment-147082251

Thanks.

alcohol:
`$extra = $this->composer->getPackage()->getExtra();` ?

Soullivaneuh:
Oh, I was looking on `getConfig()`.

Will check that.

Soullivaneuh:
Ok this will give me the extra config of the root package.

But what if I want to get extra from `$HOME/.composer/composer.json` file?

alcohol:
The global composer config? Uh... I don't think you should ever put "extra" in there.

Soullivaneuh:
@alcohol Well, check my link on my first command. You will see that having a global config make more sense for this plugin.

What is your suggestion?

alcohol:
No, it doesn't.

Global composer.json is only for global packages. It has nothing to do with your local package (your project/root package). You cannot treat local as an extension of global. It does not work like that. It's nothing like npm or the sorts. This is why I personally feel like the whole "global" thing should just be deprecated and removed. It serves no real purpose and just confuses a lot of people.

Soullivaneuh:
The plugin **is** globally installed. It's a tools to check outdated packages.

> I personally feel like the whole "global" thing should just be deprecated and removed.

I hope not, I use global require for dev tools and composer plugins. This would be painful to require it on each projects...

But this is another debate. I ensure you my plugin aims to be installed globally and so have a global config. Nothing possible this way?

alcohol:
The plugin manager does get an instance of a "global" Composer (thats how global plugins can be loaded). However, it never passes it to the interface method `active()`. This only receives the local Composer instance. Hence you cannot get the "global" local package, and thus you cannot retrieve the "extra" values.

alcohol:
> I hope not, I use global require for dev tools and composer plugins. This would be painful to require it on each projects...

I certainly do not want to remove plugin capabilities. But there are many ways to globally install dev tools and such. I just find the entire *global* concept in its current state very misleading. People often get the wrong impression/idea (because of the name, perhaps).

Soullivaneuh:
> But there are many ways to globally install dev tools and such.

[...]

PR #5129, opened 2016-03-29
Added getter for global composer

hiqsol:
I need access to global composer from my plugin and didn't find other way to get it.

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -91,6 +91,16 @@ class PluginManager
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Seldaek:
Please see https://github.com/composer/composer/issues/4494#issuecomment-147336096 - I am still not sure what is best, to wait for 2.0 and do it clean in activate(), or add this getter.

curry684:
I'd say just add it now. The whole 1.0 Plugin API is going to break anyway, doesn't matter that it gets another breaking method for when it does. At least it solves an actual problem until 2.0 is done.

Seldaek:
Yeah I guess it's fairly harmless.

PR #5134, opened 2016-03-31
Keep empty directories

JBleijenberg:
When using satis for creating archives (may it be with tar files or zip), Satis ignores empty directories. 
Usually this won't be a too big of a problem because GIT for example, does the same.
But in a recent project (Magento2) who offers soms packages in zip format, those packages can contain empty directories which are required by it's installers.
Result. No installation, no magento.

This fix checks if a given directory is empty, and when it is it places an harmless .gitkeep file inside the directory so Satis will include it.

### DIFF
--- a/src/Composer/Downloader/ArchiveDownloader.php
+++ b/src/Composer/Downloader/ArchiveDownloader.php
@@ -59,6 +59,7 @@ abstract class ArchiveDownloader extends FileDownloader
                 // move files back out of the temp dir
                 foreach ($contentDir as $file) {
                     $file = (string) $file;
+                    $this->keepEmptyDirectories($file);
                     $this->filesystem->rename($file, $path . '/' . basename($file));
                 }
 
@@ -153,5 +154,27 @@ abstract class ArchiveDownloader extends FileDownloader
             ->in($dir);
 
         return iterator_to_array($finder);
+    }
+
+    /**
+     * Check for empty folders and add a .keep file if it is empty
+     *
+     * @param string $file
+     */
+    private function keepEmptyDirectories($file) {
+        // Check if file is a directory
+        if(is_dir($file)) {
+            $directoryContent = scandir($file);
+            // Check if the folder is empty (there is more than . and ..)
+            if (count($directoryContent) <= 2) {
+                file_put_contents($file . "/.gitkeep", "");
+            } else {
+                foreach ($directoryContent as $subPath) {
+                    if($subPath !== '.' && $subPath !== '..') {
+                        $this->keepEmptyDirectories($file.'/'.$subPath);
+                    }
+                }
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


JBleijenberg:
AppVeyor fails on an/the x64 test because of permissions

src/Composer/Downloader/ArchiveDownloader.php
```diff
@@ -154,4 +155,30 @@ private function getFolderContent($dir)
 
         return iterator_to_array($finder);
     }
+
+    /**
+     * Check for empty directories.
+     * If the directory is empty, add a .gitignore file.
```
j0k3r: You mean `.gitkeep`?

src/Composer/Downloader/ArchiveDownloader.php
```diff
@@ -154,4 +155,30 @@ private function getFolderContent($dir)
...
+     *
+     * This is mainly done because Satis doesn't add empty directories to it's
```
stof: Satis does not add anything to its archives. It asks Composer to generate the archive. So it is Composer not adding them

src/Composer/Downloader/ArchiveDownloader.php
```diff
@@ -59,6 +59,7 @@ public function download(PackageInterface $package, $path)
                 // move files back out of the temp dir
                 foreach ($contentDir as $file) {
                     $file = (string) $file;
+                    $this->keepEmptyDirectories($file);
```
stof: why adding gitkeep after *downloading* the archive ? this does not make sense to me. If the archive does not contain the empty directory, it won't be here either

JBleijenberg: @stof in some cases (like the magento/magento2-base package) the original archive contains empty directories. Composer extracts the archive and rebuilds it, thus leaving out the empty directories

src/Composer/Downloader/ArchiveDownloader.php
JBleijenberg: @j0k3r i do :) i'll change this

src/Composer/Downloader/ArchiveDownloader.php
JBleijenberg: @stof you're right. I've changed the description to make more sense.

src/Composer/Downloader/ArchiveDownloader.php
stof: but my point is that the issue is not here when downloading the archive. It is when rebuilding the archive (which lives in https://github.com/composer/composer/tree/master/src/Composer/Package/Archiver)

JBleijenberg: @stof ah, then i misunderstood your first comment :) it makes sense not to do this in the download process. I'll move this to the the part where it will actually create the new archive

Seldaek:
Please see https://github.com/composer/composer/pull/4865#issuecomment-183713827 for my opinion here.. I don't think this is the correct fix either as it's essentiallydoing the same in a different place

Seldaek:
On a side-note, https://toranproxy.com/ doesn't unbundle-then-rebundle the archives like satis does, so it probably doesn't have that problem. Maybe consider using it?

PR #5137, opened 2016-03-31
Fail download on content-length mismatch

dzuelke:
I host packages on S3, which has a nasty habit of just closing connections in the middle of a transfer. Later in the process, this will then cause a failure to unpack the tarball.

This change checks the length of the loaded content against the size determined by the stream wrapper, if known. As the exception bubbles up, `FileDownloader::doDownload()` will retry two more times before giving up.

I initially wanted to implement this in `RemoteFilesystem::callbackGet()` using `STREAM_NOTIFY_COMPLETE`, but it turns out that that event is plainly not implemented in any PHP version.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -277,6 +277,11 @@ class RemoteFilesystem
         try {
             $result = file_get_contents($fileUrl, false, $ctx);
 
+            if ($this->bytesMax && strlen($result) < $this->bytesMax) {
+                // alas, this is not possible via the stream callback because STREAM_NOTIFY_COMPLETED is documented, but not implemented anywhere in PHP
+                throw new TransportException('Content-Length mismatch');
+            }
+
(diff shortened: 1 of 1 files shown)


dzuelke:
Just to demonstrate that PHP silently returns less content than the server indicated:

```php
<?php
header('Content-Length: 1000');
echo "hi\n";
?>
```

```ShellSession
$ php -S 127.0.0.1:8080
PHP 7.0.3 Development Server started at Fri Apr  1 01:50:10 2016
Listening on http://127.0.0.1:8080
Press Ctrl-C to quit.
[Fri Apr  1 01:50:14 2016] 127.0.0.1:54199 [200]: /
[Fri Apr  1 01:50:46 2016] 127.0.0.1:54207 [200]: /
```

```ShellSession
$ curl localhost:8080
curl: (18) transfer closed with 997 bytes remaining to read
hi
$ php -r 'echo file_get_contents("http://localhost:8080");'
hi
```

PR #5139, opened 2016-04-01
Add support for empty CommonName in a certificate for php <5.6

### LINKED ISSUE #5075: Composer\Downloader\TransportException - file could not be downloaded: Unable to locate peer certificate CN
Hi,

I am using composer with the WordPress BedRock project, and there's a file that cannot be downloaded.

The error message is : 

```
  The "https://wpml.org//?download=1234&user_id=12345&subscription_key=my_sub_key" file could not be downloaded: Unable to locate peer certificate CN  
  Failed to enable crypto                                                                                                                                                    
  failed to open stream: operation failed  
```

This is a download for the WPML plugin, which allows composer. Here is my repo definition : 

```
    {
      "type": "package",
      "package": {
        "name": "wpml-multilingual-cms",
        "type": "wordpress-plugin",
        "version": "3.3.6",
        "dist": {
          "type": "zip",
          "url": "https://wpml.org//?download=1234&user_id=123456&subscription_key=my_sub_key"
        },
          "require" : {
          "composer/installers": "~1.0"
        }
      }
    },
```

What do you think? Can you help me?

I'm running PHP 5.5.30-1~dotdeb+7.1, ext-openssl 1.0.1.5 and OpenSSL 1.0.1e

Seldaek:
Just to explain what is going on here:

- on <5.6, we handle SSL ourselves as PHP is kinda useless.
- wpml.org has a certificate without  Common Name, only subjectAltNames present, one of which matches wpml.org, so the certificate is valid
- PHP fails to find/match the cert (`Unable to locate peer certificate CN`), so I trigger the legacy handling of SSL (first change in RemoteFilesystem)
- I then fixed TlsHelper so that it returns an empty CN as is, instead of failing
- Then in RemoteFilesystem the problem is if we use wpml.org as `CN_match`, PHP fails the request somehow, and does not retrieve the certificate, so [the check for certificate fingerprint](https://github.com/composer/composer/blob/49524bc4baa41ea7b7464cfd5820c4526300a020/src/Composer/Util/RemoteFilesystem.php#L280-L290) fails. Due to that, I have to set verify_peer to false so that the request passes in case we have an empty CN in the cert, and then I hope that the fingerprint verification of the cert is enough to trust the request is OK.

Ping @cs278 @stormtide I need people to review the above and the patch for sanity because I really have no clue if the assumptions are correct (especially the last one). It does fix the problem, but it might also poke a giant hole, so I don't want to push it out blindly.

Fixes #5075 (for more context)

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -401,7 +401,7 @@ class RemoteFilesystem
         }
 
         // Handle SSL cert match issues
-        if (false === $result && false !== strpos($errorMessage, 'Peer certificate') && PHP_VERSION_ID < 50600) {
+        if (false === $result && (false !== strpos($errorMessage, 'Peer certificate') || false !== strpos($errorMessage, 'Unable to locate peer certificate CN')) && PHP_VERSION_ID < 50600) {
             // Certificate name error, PHP doesn't support subjectAltName on PHP < 5.6
             // The procedure to handle sAN for older PHP's is:
             //
@@ -632,6 +632,13 @@ class RemoteFilesystem
 
                 $tlsOptions['ssl']['CN_match'] = $certMap['cn'];
                 $tlsOptions['ssl']['peer_fingerprint'] = $certMap['fp'];
+
+                // in case no CN is present in the certificate, we have to cancel peer verification
+                // so that PHP doesn't fail the request and we can inspect the certificate fingerprint
+                // after the request to check if the fingerprint matched the one we stored in $certMap
+                if ($certMap['cn'] === '') {
+                    $tlsOptions['ssl']['verify_peer'] = false;
+                }
             }
         }
 
--- a/src/Composer/Util/TlsHelper.php
+++ b/src/Composer/Util/TlsHelper.php
@@ -34,11 +34,7 @@ final class TlsHelper
     {
         $names = self::getCertificateNames($certificate);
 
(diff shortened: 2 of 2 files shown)


StormTide:
@Seldaek that is the most interesting certificate i've seen in a long time. Valid sign, no CN... I'm not even sure thats CAB-valid, but, then, there it is.. Quite the edge case.

The fix approach seems dangerous however, the peer verification is the only thing that chains the request back to the valid CA. I don't know what the implications would be cryptographically in relying on the fingerprint exclusively. I think it should be pre-image secure, but without the CA chain, is the surface expanded from the set of certs issued by a valid ca that match the fingerprint to the set of self-signed certificates generated by Mallory?.... the implications on preimage resistance... I'm not confident enough on this to say its safe or not safe. Send for the cryptographers ;)

The verifypeer false is going to trigger some code analysis suites that check for this configuration, so it will cause unrelated issues in that regard. 

TL;DR; not sure its good or bad but it makes me nervous.

StormTide:
Another thought; could this open up a timing attack? -- opening the connection to the NULL CN site, then mitming the subsequent verifypeer off connection.. I'm not sure on this either, but it'd be worth researching.

rmhrisk:
Not including a CN is totally legit. See - https://unmitigatedrisk.com/?p=381

Seldaek:
Closing as per https://github.com/composer/composer/issues/5075#issuecomment-208768084

PR #5142, opened 2016-04-01
Implementation of array for `repositories.tags-path` and `repositories.branches-path`

xylesoft:
I've implemented `array` awareness to the `repositories.tags-path` and `repositories.branches-path` configuration clauses, because in SVN you are able to create more complex tags and branches directories, where you may have multiple sub-directories (as my employer does). 

This relates to a issue I submitted to the "composer/satis": https://github.com/composer/satis/issues/312

I'll also be submitting a JSON schema update for the satis.json definition.

### DIFF
--- a/src/Composer/Package/Version/VersionGuesser.php
+++ b/src/Composer/Package/Version/VersionGuesser.php
@@ -219,8 +219,16 @@ class VersionGuesser
         // try to fetch current version from svn
         if (0 === $this->process->execute('svn info --xml', $output, $path)) {
             $trunkPath = isset($packageConfig['trunk-path']) ? preg_quote($packageConfig['trunk-path'], '#') : 'trunk';
-            $branchesPath = isset($packageConfig['branches-path']) ? preg_quote($packageConfig['branches-path'], '#') : 'branches';
-            $tagsPath = isset($packageConfig['tags-path']) ? preg_quote($packageConfig['tags-path'], '#') : 'tags';
+            $branchesPath = 'branches';
+            $tagsPath = 'tags';
+            if (isset($packageConfig['branches-path'])) {
+                $branchesPath = (is_array($packageConfig['branches-path'])) ? implode('|', $packageConfig['branches-path']) : $packageConfig['branches-path'];
+                $branchesPath = preg_quote($branchesPath, '#');
+            }
+            if (isset($packageConfig['tags-path'])) {
+                $tagsPath = (is_array($packageConfig['tags-path'])) ? implode('|', $packageConfig['tags-path']) : $packageConfig['tags-path'];
+                $tagsPath = preg_quote($tagsPath, '#');
+            }
 
             $urlPattern = '#<url>.*/(' . $trunkPath . '|(' . $branchesPath . '|' . $tagsPath . ')/(.*))</url>#';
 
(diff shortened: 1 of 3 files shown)


src/Composer/Package/Version/VersionGuesser.php
```diff
@@ -219,8 +219,16 @@ private function guessSvnVersion(array $packageConfig, $path)
         // try to fetch current version from svn
         if (0 === $this->process->execute('svn info --xml', $output, $path)) {
             $trunkPath = isset($packageConfig['trunk-path']) ? preg_quote($packageConfig['trunk-path'], '#') : 'trunk';
-            $branchesPath = isset($packageConfig['branches-path']) ? preg_quote($packageConfig['branches-path'], '#') : 'branches';
-            $tagsPath = isset($packageConfig['tags-path']) ? preg_quote($packageConfig['tags-path'], '#') : 'tags';
+            $branchesPath = 'branches';
+            $tagsPath = 'tags';
+            if (isset($packageConfig['branches-path'])) {
+                $branchesPath = (is_array($packageConfig['branches-path'])) ? implode('|', $packageConfig['branches-path']) : $packageConfig['branches-path'];
+                $branchesPath = preg_quote($branchesPath, '#');
```
stof: using preg_quote here will not allow supporting multiple paths, as it will escape the ``|`` used when imploding above. you need to quote each path separately instead.

src/Composer/Repository/Vcs/SvnDriver.php
```diff
@@ -342,4 +287,115 @@ protected function buildIdentifier($baseDir, $revision)
...
+        }
+    }
+
+    /**
+     * Process the configured branches path(s)
+     */
+    private function processBranches()
+    {
+        if ($this->branchesPath !== false) {
+            if (is_array($this->branchesPath)) {
```
stof: a better system would be to always make it an array here

Seldaek:
TBH, that's a lot of changes to review for adding more obscure features to the (somewhat legacy) SVN driver for an edge case that most likely only benefits you. I can't stay this perspective excites me a whole lot :/ Have you considered maybe switching to a more standard directory structure? This would be time better spent I feel.

src/Composer/Package/Version/VersionGuesser.php
xylesoft: @stof Very good point, I will change this.

src/Composer/Repository/Vcs/SvnDriver.php
xylesoft: What makes you come to this conclusion? Surely this would create a breaking change to many existing users of the system. The implementation I decide upon was to allowing a new feature without impeding existing user.

alcohol:
@Seldaek on the other hand, since it only affects the SVN driver and repos, there isn't much harm in accepting these changes. I can see merit in these changes.

As a team (@composer), we can simply state that SVN support is provided by the community/contributors, and as such we do not directly support it ourselves. Any issues arising from this would either result in a new patch from contributors, or we simply revert the merge if no solution is provided in a timely fashion.

src/Composer/Repository/Vcs/SvnDriver.php
alcohol: @xylesoft I think that @stof means that you could simply wrap it in an array if it is not an array yet (it would just be an array with 1 path then).

Seldaek:
Yeah, I don't really mind as it's svn-only, but there is some feedback above that still wasn't addressed, which is the main reason this wasn't merged yet.

Seldaek:
I also wonder if it's really that important if people manage to get by without such a patch for a few months..

PR #5144, opened 2016-04-02
Skip non-empty directories in zip generation

ktomk:
Empty dirs get archived (leafs).

Previously it seemed ok to skip all directories in zip generation.

References:

- Ref: 6066359944642e47d03c2dd367c9e43af787a9f7

- Issue: #4865 Keep empty folders after re-zip a module

### DIFF
--- a/src/Composer/Package/Archiver/ArchivableFilesFinder.php
+++ b/src/Composer/Package/Archiver/ArchivableFilesFinder.php
@@ -13,7 +13,9 @@
 namespace Composer\Package\Archiver;
 
 use Composer\Util\Filesystem;
+use FilesystemIterator;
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

astorm:
@Seldaek @ktomk Thanks for accepting this -- makes a Magento developer slightly less crazy :)  Quick naive question -- do tar archived need feature parity here? https://github.com/composer/composer/issues/5206

ktomk:
@astorm: Yes they need. Right now empty directories are added as zero-byte files due to these changes. Please see as well PR #5144 which fixes this flaw and I already added some comments in #5206.

PR #5150, opened 2016-04-04
Rollback installation of plugins if corrupted, fixes #5130

### LINKED ISSUE #5130: Failure rollback doesn't remove symlink for path type repositories
With a composer.json requiring a broken package in a path repository, a subsequent install attempt will fail with a runtime exception:

```
[RuntimeException]
Package package/name cannot install to "/some/path" inside its source at "/some/path"
```

* Previous failure rollback should unlink the symlink if it was newly created
* Path downloader shouldn't use realpath on the target location for the symlink (realpath for dirname(target) possibly?)

curry684:
Fixes #5130, sample output:
```
[64.8MB/4.70s] Analyzed 54 packages to resolve dependencies
[64.8MB/4.70s] Analyzed 55 rules to resolve dependencies
[64.9MB/4.71s]   - Installing slbmeh/gh-5130-bad-dependency (dev-master)
[64.9MB/4.73s]     Junctioned from ../libs/bad-dependency/
[64.9MB/4.73s] 
[64.9MB/4.73s] Plugin installation failed, rolling back
[64.9MB/4.73s]   - Removing junction for slbmeh/gh-5130-bad-dependency (dev-master)

                                                                                             
  [UnexpectedValueException]                                                                 
  Plugin slbmeh/gh-5130-bad-dependency could not be initialized, class not found: SomeClass  
```

### DIFF
--- a/src/Composer/Installer/PluginInstaller.php
+++ b/src/Composer/Installer/PluginInstaller.php
@@ -59,7 +59,14 @@ class PluginInstaller extends LibraryInstaller
         }
 
         parent::install($repo, $package);
-        $this->composer->getPluginManager()->registerPackage($package, true);
+        try {
+            $this->composer->getPluginManager()->registerPackage($package, true);
+        } catch(\Exception $e) {
+            // Rollback installation
+            $this->io->writeError('Plugin installation failed, rolling back');
+            parent::uninstall($repo, $package);
+            throw $e;
+        }
     }
 
     /**


src/Composer/Installer/PluginInstaller.php
```diff
@@ -59,7 +59,14 @@ public function install(InstalledRepositoryInterface $repo, PackageInterface $pa
         }
 
         parent::install($repo, $package);
-        $this->composer->getPluginManager()->registerPackage($package, true);
+        try {
+            $this->composer->getPluginManager()->registerPackage($package, true);
+        } catch(\Exception $e) {
+            // Rollback installation
+            $this->io->writeError('Plugin installation failed, rolling back');
```
stof: you should include the plugin package name in the output, to make debugging easier (knowing which plugin is broken)

curry684: Isn't that a bit overdoing it? That would make the output:
```
 - Installing bad/plugin (dev-master)
   Junctioned from ../libs/bad-dependency/

Plugin installation failed for bad/plugin, rolling back
 - Removing junction for bad/plugin (dev-master)

  [UnexpectedValueException]                                                                 
  Plugin bad/plugin could not be initialized, class not found: SomeClass  
```
Is a 4th mention really going to make it clearer?   :smile:

Seldaek:
Thanks, that's what I thought yeah :)

PR #5156, opened 2016-04-06
Fix config add/remove/unset/merge for Bitbucket OAuth

dzuelke:
Not sure why the code to handle `bitbucket-oauth` was in its own section, but it prevented `--unset` from working for that key. Moved it where it belongs.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -167,7 +167,7 @@ EOT
         }
         if ($input->getOption('global') && !$this->authConfigFile->exists()) {
             touch($this->authConfigFile->getPath());
-            $this->authConfigFile->write(array('http-basic' => new \ArrayObject, 'github-oauth' => new \ArrayObject, 'gitlab-oauth' => new \ArrayObject));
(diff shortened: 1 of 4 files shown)


dzuelke:
Before:

```ShellSession
$ composer config -g --unset bitbucket-oauth.bitbucket.org 

                                                                 
  [RuntimeException]                                             
  Expected two arguments (consumer-key, consumer-secret), got 0  
                                                                 

config [-g|--global] [-e|--editor] [-a|--auth] [--unset] [-l|--list] [-f|--file FILE] [--absolute] [--] [<setting-key>] [<setting-value>]...
```

After:

```ShellSession
$ bin/composer config -g --unset bitbucket-oauth.bitbucket.org
```

src/Composer/Command/ConfigCommand.php
```diff
@@ -466,7 +466,7 @@ function ($vals) {
         }
 
         // handle github-oauth
```
stof: this should be reworded

dzuelke: yeah I figured I'd rather not touch it. I can do it, together with a fix to the changelog entry possibly - @Seldaek do you want to do a beta or RC of 1.0.1 first?

dzuelke:
Fixed handling in a few more spots. @stof do you want to check this out? Also @wenkepaul please review this thoroughly and confirm that these are all correct since you did the original PR :)

PR #5162, opened 2016-04-06
Code quality

curry684:
Fixed a ton of invalid phpDoc sections and missing or incorrect `return` statements.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -18,7 +18,7 @@ use Composer\Installer\InstallationManager;
 use Composer\IO\IOInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\PackageInterface;
-use Composer\Repository\InstalledRepositoryInterface;
+use Composer\Repository\WritableRepositoryInterface;
 use Composer\Util\Filesystem;
 use Composer\Script\ScriptEvents;
 
@@ -85,7 +85,7 @@ class AutoloadGenerator
         $this->runScripts = (boolean) $runScripts;
     }
 
-    public function dump(Config $config, InstalledRepositoryInterface $localRepo, PackageInterface $mainPackage, InstallationManager $installationManager, $targetDir, $scanPsr0Packages = false, $suffix = '')
+    public function dump(Config $config, WritableRepositoryInterface $localRepo, PackageInterface $mainPackage, InstallationManager $installationManager, $targetDir, $scanPsr0Packages = false, $suffix = '')
     {
         if ($this->classMapAuthoritative) {
             // Force scanPsr0Packages when classmap is authoritative
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -305,6 +305,8 @@ TAGSPUBKEY
      * @param string $localFilename
      * @param string $newFilename
      * @param string $backupTarget
+     * @return \Exception|null
+     * @throws \Exception
      */
     protected function setLocalPhar($localFilename, $newFilename, $backupTarget = null)
     {
@@ -323,6 +325,7 @@ TAGSPUBKEY
             }
 
             rename($newFilename, $localFilename);
+            return null;
         } catch (\Exception $e) {
             if (!$e instanceof \UnexpectedValueException && !$e instanceof \PharException) {
                 throw $e;
--- a/src/Composer/Installer/SuggestedPackagesReporter.php
+++ b/src/Composer/Installer/SuggestedPackagesReporter.php
@@ -29,7 +29,7 @@ class SuggestedPackagesReporter
     protected $suggestedPackages = array();
 
     /**
-     * @var Composer\IO\IOInterface
+     * @var \Composer\IO\IOInterface
      */
     private $io;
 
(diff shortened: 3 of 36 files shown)


src/Composer/Package/Archiver/GitExcludeFilter.php
```diff
@@ -69,12 +69,9 @@ public function parseGitAttributesLine($line)
     {
         $parts = preg_split('#\s+#', $line);
 
-        if (count($parts) != 2) {
-            return null;
-        }
-
-        if ($parts[1] === 'export-ignore') {
+        if (count($parts) == 2 && $parts[1] === 'export-ignore') {
```
stloyd: Should be strict.

src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -305,6 +305,8 @@ protected function rollback(OutputInterface $output, $rollbackDir, $localFilenam
      * @param string $localFilename
      * @param string $newFilename
      * @param string $backupTarget
+     * @return \Exception|null
```
barryvdh: Return type could be more restricted: `\UnexpectedValueException|\PharException`

src/Composer/Installer/SuggestedPackagesReporter.php
```diff
@@ -29,7 +29,7 @@ class SuggestedPackagesReporter
     protected $suggestedPackages = array();
 
     /**
-     * @var Composer\IO\IOInterface
+     * @var \Composer\IO\IOInterface
```
barryvdh: Can be `IOInterface`, is already imported.

src/Composer/Package/Archiver/GitExcludeFilter.php
curry684: Which part? `count` is [very well defined](http://php.net/manual/en/countable.count.php) to return an `int`.

src/Composer/Command/SelfUpdateCommand.php
curry684: I think for this case the common ancestor more clearly conveys intentions.

alcohol:
Could you also run `php-cs-fixer`? That should fix the indentation in some of the docblocks you modified.

curry684:
Yeah still working on it. Still got a lot of parameters to reorganize but it's bitching complex here and there where newer stuff like `WriteableRepositoryInterface` is being passed as `RepositoryInterface` and the like. So the PR can remain open for a while and I'll run CS fixer at the end once.

src/Composer/Command/SelfUpdateCommand.php
barryvdh: Why? Other Exceptions are thrown, not returned.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -85,7 +85,7 @@ public function setRunScripts($runScripts = true)
         $this->runScripts = (boolean) $runScripts;
     }
 
-    public function dump(Config $config, InstalledRepositoryInterface $localRepo, PackageInterface $mainPackage, InstallationManager $installationManager, $targetDir, $scanPsr0Packages = false, $suffix = '')
+    public function dump(Config $config, WritableRepositoryInterface $localRepo, PackageInterface $mainPackage, InstallationManager $installationManager, $targetDir, $scanPsr0Packages = false, $suffix = '')
```
Seldaek: I reverted this change, InstalledRepositoryInterface is what's needed as far as intent is concerned. It doesn't matter if it's technically more strict than necessary IMO.

Seldaek:
Overall: thanks, good cleanups :)

I'll merge now because whatever is done already is reviewed and done.

A few comments though:

- the `@throws Exception` when we re-throw something I really don't see what value they bring, but I left  them.
- the explicit `return null;`  vs `return;` or just implicit return at end of the function, I also don't see what the point is. I leave them in as I am not gonna waste time going over and changing them back, but I find those changes to be silly IDE-pleasing wastes of time.
- please don't run php-cs-fixer, I rather do it myself every now and then than have to review large changes done by others. It's the job of the maintainer IMO.

Just for the record..

src/Composer/Package/Archiver/GitExcludeFilter.php
Seldaek: Not a huge deal I think, but I tend to agree that strict is better by default, because when you see strict you don't have to think twice whether count() can return something nasty or not, or whether the loose comparison was done intentionally or not (in the case count() returned something else). 

It's about reducing mental load when reading code.

curry684:
I don't agree with the `return null` changes being pointless. If there's a phpDoc stating a return type, it's confusing to use empty return statements. In my book `return;` is only valid inside a void-returning function, and using it to default a full type is a strictness violation. It's sort of a style preference though of course.

Overall my goal with this is to (greatly) reduce the amount of warnings in the IDE, thus causing more predictable and maintainable code for everyone. Thanks for merging, I'll get on more of them when I have a bored evening again   :wink:

Seldaek:
Yeah I realize a lot of this is IDE driven, like a lot of the silly 
inline /* @var */ which really have no reason to exist except to serve 
our robot overlords.

I use sublime text though so I don't see those warnings and it kinda 
bothers me to have all the added cruft comments sometimes. I wish IDEs 
were smarter at figuring things out (or at using what they figured out 
instead of just nagging humans to fix shit they know already).

PR #5164, opened 2016-04-07
trim trailing slash from symlink names

cherbst:
symlink names are not allowed to contain
a trailing slash, so trim it.


curry684:
Did this ever occur? I don't think it should be able to as the path is originally globbed with `GLOB_ONLYDIR`.

dzuelke:
Perfect fix, please merge. This is what causes https://github.com/composer/installers/issues/267 / https://github.com/composer/installers/issues/282 !

Seldaek:
Thanks. @curry684 I also fixed it at the source in the path repo, but better have multiple layers of safety IMO.

PR #5170, opened 2016-04-08
Update troubleshooting.md

lferro9000:
Alternative way to turn off composer on Mac OS, with PHP installed in user available folders by Homebrew.

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -202,6 +202,22 @@ echo 'function composer() { COMPOSER="$(which composer)" || { echo "Could not fi
 . ~/.bash_aliases
 ```
 
+On platforms without `php5enmod` and `php5dismod` you can run:
+
+```sh
+php -i | grep php.ini
+```
+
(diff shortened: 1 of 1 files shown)


doc/articles/troubleshooting.md
```diff
@@ -202,6 +202,22 @@ echo 'function composer() { COMPOSER="$(which composer)" || { echo "Could not fi
 . ~/.bash_aliases
 ```
 
+On platforms without `php5enmod` and `php5dismod` you can run:
+
+```sh
+php -i | grep php.ini
```
alcohol: Why not just recommend using `php --ini` here?

doc/articles/troubleshooting.md
```diff
@@ -202,6 +202,22 @@ echo 'function composer() { COMPOSER="$(which composer)" || { echo "Could not fi
...
+```
+
+To check where the PHP configuration is, and then run use a similar script:
```
alcohol: Either go with run, or use, but not both :-)

PR #5172, opened 2016-04-10
Speedup autoloading with APCu class loader

nicolas-grekas:
The fastest autoloader we can currently get with composer is when dumping the class map in authoritative mode: this reduces autoloading to an `isset()` check, one of the fastest in PHP.
But this autoloader has one drawback: it doesn't scale with the number of classes. Even with opcache enabled, the generated class map being a php array, it becomes slower when you add more classes to it. It also consumes more memory.

That made me wonder if using APCu could help build an even better autoloader.

This PR adds an ApcuClassLoader that wraps the regular ClassLoader by inheritance so that only APCu is used when the cache is warm, and the regular autoloader is lazily initialized when needed to warm up the APCu cache.

I did not yet wire any on/off way to configure which loader to use. Just change yourself the `$apcuLoader` property if you want to try the patch for now.

Still, I wanted to share with you the following Blackfire profile:
https://blackfire.io/profiles/compare/ace900a5-4936-4998-871d-680bbf744177/graph

This profile has been generated by taking the blackfire classmap (~8300 lines) and loading ~430 classes from it, this number being quite typical for rendering a single page/response in a Symfony app.

The results are quite interesting:
- loading the classmap goes from 2.66 ms and 2.56 MB to zero
- overall loading time is reduced by 1.59 ms and 2.54 MB!

On the memory side, we have a clear win.
On the time side, 2.66 ms removal from not-loading the classmap are turned into a lower 1.59 ms general perf increase. Logically, this could mean that *once the classmap is loaded*, using it is faster that calling apcu_fetch, but since we have to load this classmap anyway, the win is still a clear win in the typical web request use case where only a short excerpt of all existing classes are loaded at once.

The only remaining question to me now is: how should we plug this into composer configuration system.
Some ideas:
- add a command line option?
- `ext-apcu` must-be in the dependencies?
- both? Other way?

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -51,6 +51,11 @@ class AutoloadGenerator
     /**
      * @var bool
      */
+    private $apcuLoader = false;
+
+    /**
+     * @var bool
+     */
     private $runScripts = false;
 
     public function __construct(EventDispatcher $eventDispatcher, IOInterface $io = null)
@@ -73,6 +78,16 @@ class AutoloadGenerator
     public function setClassMapAuthoritative($classMapAuthoritative)
     {
         $this->classMapAuthoritative = (boolean) $classMapAuthoritative;
+    }
+
+    /**
+     * Whether or not generated autoloader uses the APCu loader.
+     *
+     * @param bool $apcuLoader
+     */
+    public function setApcuLoader($apcuLoader)
+    {
+        $this->apcuLoader = (boolean) $apcuLoader;
     }
 
     /**
@@ -287,6 +302,7 @@ EOF;
         file_put_contents($targetDir.'/autoload_real.php', $this->getAutoloadRealFile(true, (bool) $includePathFileContents, $targetDirLoader, (bool) $includeFilesFileContents, $vendorPathCode, $appBaseDirCode, $suffix, $useGlobalIncludePath, $prependAutoloader));
 
         $this->safeCopy(__DIR__.'/ClassLoader.php', $targetDir.'/ClassLoader.php');
+        $this->safeCopy(__DIR__.'/ApcuClassLoader.php', $targetDir.'/ApcuClassLoader.php');
         $this->safeCopy(__DIR__.'/../../../LICENSE', $targetDir.'/LICENSE');
 
         if ($this->runScripts) {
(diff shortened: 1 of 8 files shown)


beberlei:
@nicolas-grekas Can you make the same comparison on PHP 7? I would be interested how much of the measured benefit will go away, since hashtables were optimized so much.

omansour:
brilliant !!! 

:dancers: 

I think detecting ext-apcu and using the apcu loader automatically has my preference. But maybe it add some small overhead too. So ext-apcu will be suggested

jderusse:
Given the cache is shared, what happens when the same machine run 2 applications using different versions of the same class ?

jderusse:
Oh. I miss the prefix thing. 

Is the prefix regenerated after each install? For instance after that a dependency is upgrade/downgraded...

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -606,19 +622,42 @@ public static function getLoader()
...
+            $file .= <<<REGISTER_CLASS_LOADER
+        spl_autoload_register(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'), true, $prependAutoloader);
+        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader();
+        spl_autoload_unregister(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'));
+
+$initializer
+REGISTER_CLASS_LOADER;
+        } else {
+            $initializer = preg_replace('/^ +/m', '    $0', $initializer);
+            $apcuPrefix = substr(base64_encode(md5(uniqid('', true), true)), 0, -2);
```
stof: rather than using a random prefix here, you should reuse the ```$suffix`` variable, which can be configured to reuse the same value when regenerating the autoloader (default behavior is still to generate a random one each time)

stof:
@jderusse currently, it is regenerated randomly for each autoload dump. See my comment about reusing the existing configurable setting (which still defaults to randomly generating a value)

nicolas-grekas:
In reality, the suffix is not always regenerated, so that reusing it wouldn't always invalidate the cache when required, that's why I did it this way.

JanTvrdik:
> ApcuClassLoader that wraps the regular ClassLoader by inheritance

Why not use composition instead of inheritance? Why bother loading `ClassLoader` at all when APCu cache is ready?

nicolas-grekas:
@JanTvrdik @omansour because `$loader instanceof ClassLoader` must stay true, and also because we need lazy-initialization.

nicolas-grekas:
Closing in favor of #5174

PR #5174, opened 2016-04-11
Speedup autoloading on PHP 5.6 & 7.0+ using static arrays

nicolas-grekas:
In #5172 I tried an apcu lazy-loading proxy that raised [the following concern](https://twitter.com/lisachenko/status/719105126439497728) from @lisachenko:
> on a big project usage of APC can lead to slow response time due to the memory fragmentation issue

It also raised [this comment](https://twitter.com/nikita_ppv/status/719150439065751556) from @nikic, which correlates with an other of the same kind I've got from @jpauli saying also this should even work on 5.6:
> Assuming the class map is a fully constructed plain array, PHP 7 opcache should store it in SHM w/o copies

So here is the patch, and [the corresponding Blackfire profile comparison](https://blackfire.io/profiles/compare/991454f2-ecf4-4e44-b7f3-3c0d4def48e8/graph):
capture du 2016-04-11 12 59 22

The test script is exactly the same as in #5172: ~8300 lines in the class map amongst which ~420 are loaded.
As you can see on the above graph, PHP 7 this time, the performance increase is significant!
And this comes without any new configuration options!

Patch best viewed [ignoring white space diffs](https://github.com/composer/composer/pull/5174/files?w=1).


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +690,90 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+        $initializer = '';
+        $prefix = "\0Composer\Autoload\ClassLoader\0";
+        if (file_exists($targetDir . '/autoload_files.php')) {
+            $maps = array('files' => require $targetDir . '/autoload_files.php');
+        } else {
+            $maps = array();
+        }
+
+        foreach ((array) $loader as $prop => $value) {
+            if ($value && 0 === strpos($prop, $prefix)) {
```
lisachenko: It will be better not to use `strpos` here, because it scans the string, instead just use `substr`, or direct check via isset($prop[0]) && ("\0" === $prop[0])

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +690,90 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+                $maps[substr($prop, strlen($prefix))] = $value;
```
lisachenko: `strlen($prefix)` can be calculated outside foreach loop

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -578,33 +579,44 @@ public static function getLoader()
 INCLUDE_PATH;
         }
 
+        $file .= <<<STATIC_INIT
+        if (PHP_VERSION_ID >= 50600) {
+            require_once __DIR__ . '/autoload_static.php';
+
+            \$initializer = \Composer\Autoload\ComposerStaticInit$suffix::getInitializer(\$loader);
```
Ocramius: Can actually just `\call_user_func()` around it (`\` is important, as otherwise PHP can't optimize it away)

lisachenko:
I like this version of loader, because it uses native opcache optimization for static arrays. Nice job! :+1:

Seldaek:
Is it possible to get a test that checks the output of the autoload_static.php file? Because if this breaks.. everything goes boom.

Ocramius:
Absolutely :+1: to this! Didn't realize that non-constants were used for the classmap.

Just a note: need to:

 * document that anything built with this setting has to be deployed to the same directory path, or things will crash and burn (some projects build a zip/tar/gz package and upload it)
 * a way to turn this on conditionally, or else require a major BC break and a major version bump

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +690,90 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+namespace Composer\Autoload;
+
+class ComposerStaticInit$suffix
+{
+
+HEADER;
+
+        $filesystem = new Filesystem();
+
+        $vendorPathCode = ' ' . $filesystem->findShortestPathCode(realpath($targetDir), $vendorPath, true, true) . " . '";
```
Ocramius: This is where the BC break is: the last `true` parameter will cause systems where `build system != deployment system` to break.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +690,90 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+        $appBaseDirCode = ' ' . $filesystem->findShortestPathCode(realpath($targetDir), $basePath, true, true) . " . '";
```
Ocramius: This is where the BC break is: the last `true` parameter will cause systems where `build system != deployment system` to break.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +690,90 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+        foreach ((array) $loader as $prop => $value) {
+            if ($value && 0 === strpos($prop, $prefix)) {
+                $maps[substr($prop, strlen($prefix))] = $value;
+            }
+        }
+
+        foreach ($maps as $prop => $value) {
+            $value = var_export($value, true);
+            $value = str_replace($absoluteVendorPathCode, $vendorPathCode, $value);
+            $value = str_replace($absoluteAppBaseDirCode, $appBaseDirCode, $value);
```
lisachenko: Can be combined into the one `strtr()` here

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +690,90 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+
+            if (0 === strpos($value, 'array (')) {
```
Ocramius: This block seems extremely hacky. At this point, isn't it better to completely phase out `var_export()` instead?

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +690,90 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+                $value = preg_replace('/^ +/m', '    $0$0', $value);
+                $value = preg_replace('/=> $\r?\n +array \($/m', '=> [', $value);
+                $value = preg_replace('/\),$/m', '],', $value);
+                $value = preg_replace('/^array \(\)$/', '[$1]', $value);
+                $value = '[' . substr($value, 7, -1) . '    ]';
+            }
+
+            $file .= sprintf("    public static $%s = %s;\n\n", $prop, $value);
```
Ocramius: Can actually be a constant

Seldaek:
Aye see https://twitter.com/seldaek/status/719422049681338369 - if we can't make it relative, then it has to be optional, like `--optimize-autoloader=static` (and add `--optimize-autoloader=default` for the current model).

As far as I can tell right now it's building absolute paths but haven't reviewed in details.

lisachenko:
I have one more idea: what if try to make all autoload_xxx.php files containing static arrays with special prefixes for `baseDir` and `vendorDir` (or maybe only `baseDir`), like virtual schema:

```php
// autoload_psr4.php @generated by Composer
return array(
    'Symfony\\Polyfill\\Mbstring\\' => array('vendor://symfony/polyfill-mbstring'),
    'Symfony\\Component\\Console\\' => array('vendor://symfony/console'),
    'App\\' => array('base://src'),
);
```

And then define for `autoloader_real.php` our current values for `baseDir` and `vendorDir` as `__DIR__` and `dirname(__DIR__)`, so this will give as an ability to utilize opcache optimization for static map, as well, we won't use absolute paths everywhere. 

During initialization, we resolve all values with `strtr()` and use them. Profit!

src/Composer/Autoload/AutoloadGenerator.php
Ocramius: Hmm, no, it cannot. [HHVM doesn't accept it](https://3v4l.org/6TsKZ)

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: strpos is the fastest. I know that because we use it (and benched it heavily) on Symfony Router.
scanning is faster that duplicating a (sub)string.

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: changed

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: done

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: strtr is slower (algo is more complicated), and readability is already good IMHO.

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: Hacky but working. Rewritting means more code => more potential bugs. I'd prefer adding more tests. This is only code generating code...

Ocramius: True. Possible candidate for a rewrite then.

nicolas-grekas:
@lisachenko @Seldaek I don't get the issue that you're trying to solve: the paths are already relative, thanks to using `__DIR__`. @lisachenko your suggestion requires processing, which means changing the array, which means duplicating what's in shared memory. Which means almost canceling the benefit (because strtr time + memo copy).

Seldaek:
@nicolas-grekas for me on windows it built all paths absolute. Not sure if due to windows or I had an outdated version of the PR. I'll try again later.

lisachenko:
@nicolas-grekas from the PR it's not so clear what happens with variables, especially after several var_export and preg_matches, so code looks like it exporting absolute paths. So I should run your PR locally before future discussions. But if you use relative paths, then my suggestions are irrelevant.

Maybe you can publish a gist with new dumped files? So, anyone will be able to quickly look at the final result, without checking out a local version.

Seldaek:
Yup that's also why I'd like the `autoload_static.php` output to be part of the test suite (much like autoload_real and others are), so we can ensure it's generated properly on all platforms.

nicolas-grekas:
`autoload_static.php` files added to the test suite. Next step: Windows

tests/Composer/Test/Autoload/Fixtures/autoload_static_target_dir.php
```diff
@@ -0,0 +1,38 @@
...
+            ],
+            'Main\\Bar' => [
+                0 => __DIR__ . '/../..' . '/',
+            ],
+        ],
+    ];
+
+    public static $classMap = [
+        'ClassMapBar' => __DIR__ . '/../..' . '/lib/rootbar.php',
+        'ClassMapFoo' => __DIR__ . '/../..' . '/src/rootfoo.php',
```
Seldaek: Maybe an additional transformation to do on the arrays would be `' . '` => ``? Sounds safe but not sure, maybe not worth the risk though as I assume this gets resolved and then the result is cached in opcache.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +689,91 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+            if ('files' !== $prop) {
+                $initializer .= "            \$loader->$prop = ComposerStaticInit$suffix::\$$prop;\n";
+            }
+        }
+
+        return $file . <<<INITIALIZER
+    public static function getInitializer(ClassLoader \$loader)
+    {
+        return \Closure::bind(function () use (\$loader) {
```
Ocramius: Why is this needed, exactly?

src/Composer/Autoload/AutoloadGenerator.php
Ocramius: I must say that while the above statements are true, I am unable to understand what this code does: it should just go away, tbh.

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: As far as I understand that's used to inject values into the private properties of the ClassLoader instance by using the closure::bind as a way to gain access.

Ocramius: Eh, does it need to happen that way? Making the property public or having a public hook would be much better. Closure re-binding is one of those things that internals will unlikely be able to optimize soon anyway.

nicolas-grekas: There is nothing to optimize here about rebinding IMHO: this is a run once closure. Not changing the existing ClassLoader looks more important to me. Injecting into the private scope is mandatory to get the benefit of copy-on-write from shared memory.

tests/Composer/Test/Autoload/Fixtures/autoload_static_target_dir.php
nicolas-grekas: this will be done by the php compiler, no need to care on our side I think

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: this block of code is not strictly required: it just fixes indentation and turns `array (..)` code style to `[...]`

src/Composer/Autoload/AutoloadGenerator.php
Ocramius: I am aware of what it does (kinda), it just isn't possible to keep it as-is, as it is just a mess :P

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -674,6 +689,84 @@ function composerRequire$suffix(\$fileIdentifier, \$file)
...
+            if ($value && 0 === strpos($prop, $prefix)) {
+                $maps[substr($prop, $prefixLen)] = $value;
+            }
+        }
+
+        foreach ($maps as $prop => $value) {
+            $value = var_export($value, true);
+            $value = str_replace($absoluteVendorPathCode, $vendorPathCode, $value);
+            $value = str_replace($absoluteAppBaseDirCode, $appBaseDirCode, $value);
+            $value = ltrim(preg_replace('/^ */m', '    $0$0', $value));
```
nicolas-grekas: I removed the `var_export` cleaning and kept only this line that at least fixes the indentation.

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: @Seldaek please advise you preference: we can add a public method on ClassLoader. I had this "contract" in mind not to change the existing class, but maybe not a requirement to you?

nicolas-grekas:
Fixed on Windows, and green on Travis; only minor things should remain.
More comments welcomed if any.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -578,33 +579,43 @@ public static function getLoader()
 INCLUDE_PATH;
         }
 
+        $file .= <<<STATIC_INIT
+        if (PHP_VERSION_ID >= 50600) {
```
stof: Can OPCache optimize this condition, or do we need to access the constant with its fully qualified name (as done for ``call_user_func`` below) ?
This needs some input from the C devs here (or dumping the opcodes after optimization)

Seldaek: Given that this file/class is not namespaced, I don't think it's necessary, and the deref of call_user_func either actually.

nicolas-grekas: In fact, the backslash makes no difference here: we're in the global
namespace :-)

src/Composer/Autoload/AutoloadGenerator.php
naderman: I don't think adding a method there is a problem.

Seldaek: It's not a huge problem no, but there are two small points that bug me..

- It'd be a public method added just for an internal optimization, that has no real purpose in the wild.
- If someone has say phpunit installed globally (even though I don't think that's a great idea, some people do), then until they re-dump the phpunit autoloader they'll get a fatal error when their local  project tries to use the global ClassLoader instance (that was loaded first). Not a huge deal but a small migration pain which in my experience will lead to confused people posting new issues :)

So all in all I'd be rather for keeping the closure "hack".

Seldaek: Other alternative I mentioned to Nicolas earlier is to add the setter now, keep the closure hack for now, then in 1.2 we switch to use the setter and no more closure, which should mostly get rid of the upgrade pains.

Seldaek:
Alrighty, let's hope we didn't miss anything :) Thanks @nicolas-grekas and everyone for the review.

Ocramius:
Nice! :+1:

dmaicher:
Nice Job :+1:  Saves around 1.5 - 2ms per request using PHP 5.6 on one of my projects :)

The `autoload_classmap.php` file is around 11000 lines long in my case.

Seldaek:
Nice. I'd be curious to see numbers from other people if anyone has
production monitoring showing an impact on response times.

nicolas-grekas:
So do I, response time and memory usage also.

tests/Composer/Test/Autoload/Fixtures/autoload_static_files_by_dependency.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+// autoload_static.php @generated by Composer
+
+namespace Composer\Autoload;
+
+class ComposerStaticInitFilesAutoloadOrder
+{
+    public static $files = array (
```
Tobion: As autoloader is only used on PHP 5.6 anyway, shouldn't this also be a `const`, i.e. `const FILES = array()`?
Tested with https://3v4l.org/Q8s9s it seems that HHVM does not support const arrays yet, so would probably need a switch then.

mvrhov:
Sorry to break your bubble, but this won't work on 5.6 if there is more than 32k entries in array.
https://bugs.php.net/bug.php?id=68057

nicolas-grekas:
@mvrhov not sure what you mean by "bubble", but anyway, here is the workaround: #5183. Thanks for the remainder!

mvrhov:
It a figure of speech. And it seems that I remembered it wrong the correct one is "burst your bubble"
> When someone is feeling all happy and opptimistic, you take your asshole self and totally break his happyness, therefore bursting his bubble

backbone87:
@nicolas-grekas so we should **not** use the symfony ApcuClassLoader in PHP5.6/7 as hinted in https://github.com/symfony/symfony-standard/blob/master/web/app.php#L16 anymore, when we use `php composer.phar dump-autoload --classmap-authoritative`? Or is it still faster? If we shouldnt use it, then https://github.com/symfony/symfony-standard/blob/master/web/app.php#L16 should be removed.

nicolas-grekas:
@backbone87 good point, please submit a PR against 2.3!

backbone87:
@nicolas-grekas ok, i will do that, but is there still some gain in using the ApcuClassLoader, and if yes, is it still "worth it" considering the complications you mentioned in the OP.

nicolas-grekas:
To me, APCu classloader is dead now, espcially since @lisachenko also says:

>  on a big project usage of APC can lead to slow response time due to the memory fragmentation issue

pborreli:
it is still relevant for PHP5.5 users not covered by this PR

backbone87:
idk, what the strategy in terms of production settings hints is, but then this can be discussed in the PR to symfony-standard https://github.com/symfony/symfony-standard/pull/964

stof:
:+1: for that

src/Composer/Autoload/AutoloadGenerator.php
rvanvelzen: Isn't `0 === strncmp($prop, $prefix, $prefixLen)` even faster?

Ocramius: I wouldn't worry about performance here: it's build-time code with very tiny impact compared to other more complex operations (SAT, downloading)

iquito:
Great improvement! With PHP7 and my application I get about 1ms improvement with this per request (according to my estimates), and the memory requirements go down dramatically, here are three examples of three different pages, before and after (only change is the static class loader, and I have around 4000 classes in classmap):

2.5MB -> 1.8MB
3.025MB -> 2.4MB
2.856MB -> 2.24MB

So this saves around 0.6MB per request, or between 20% and 25%, which is huge! Especially with PHP7 already being very memory efficient.

src/Composer/Autoload/AutoloadGenerator.php
lisachenko: @nicolas-grekas confirmed that `strpos()` is the fastest in PHP7, but in PHP5.6 `isset()` with zero check is winner, see: https://3v4l.org/1EUVv.

Anyway, `substr()` is the slowest, and this check doesn't matter at all for runtime performance, as @Ocramius highlighted.

jewome62:
Finally, what we must do to enable this optimization on PHP7?
Just be on PHP7 ?

Thanks for your help

Ocramius:
@jewome62 you need PHP7 and you need to enable OpCache (which is provided with it, usually).

tests/Composer/Test/Autoload/Fixtures/autoload_static_files_by_dependency.php
dvikan: I also wonder about this.

PR #5177, opened 2016-04-11
Update bundled cacert.pem

ravage84:
Certificate data from Mozilla as of: Wed Jan 20 04:12:04 2016


Seldaek:
Thanks

ravage84:
Recently, the CakePHP  project also needed to update this file.

https://github.com/cakephp/cakephp/pull/8608

I assume quite a few others depend in one way or another.

So I was thinking, would it make sense to create a Packagist package out of it?
Like this:
https://github.com/ravage84/curl-cacert (draft)

Seldaek:
Hmm, yes that's a good question.. TBH for reasons of trust I'd rather maintain this myself than depending on a third party. We could publish it as an external repo though to let others re-use (as they most likely have to put trust into composer anyway..).

ravage84:
Sure, totally understand. I didn't plan to host it under my name anyway.
It makes sense that somebody or better said an organisatoin/team with a good reputation hosts it.

What do you think about automate the update process as I drafted it [here](https://github.com/ravage84/curl-cacert#todo)?

Seldaek:
Yes that sounds good but maybe not first order of business. I'll try and extract most of our CA logic to https://github.com/composer/ca-bundle then we can think about automating.

ravage84:
Cool. :bow: 

How about defining a constant for the cert path, like [this](https://github.com/ravage84/curl-cacert/blob/master/cacert.php) and including it via [Composer file autoloading](https://getcomposer.org/doc/04-schema.md#files)?

Certainly, not priority one, too, but it could ease the integration of the file for others.

Seldaek:
It's now available as https://packagist.org/packages/composer/ca-bundle and CaBundle::getBundledCaBundlePath() gets you the path if you need it. Or CaBundle::getSystemCaRootBundlePath() to autodetect.

PR #5183, opened 2016-04-12
Workaround https://bugs.php.net/68057

nicolas-grekas:
Static arrays are limited to 32767 values on PHP 5.6
See https://bugs.php.net/68057


Seldaek:
I dread the day I see a 32K item classmap, but eh well, thanks :)

PR #5184, opened 2016-04-12
Tilde and environment variable expansion for paths

curry684:
Allows using tildes in path repositories for easier sharing of path repositories between multiple team members. For example a path repository to `~/git/mysubproject/*` will automatically load all those packages from the developer's home folder.

### DIFF
--- a/src/Composer/Util/Platform.php
+++ b/src/Composer/Util/Platform.php
@@ -31,6 +31,10 @@ class Platform
             return self::getUserDirectory() . substr($path, 1);
         }
         return preg_replace_callback('#^([\\$%])(\\w+)\\1?(([/\\\\].*)?)#', function($matches) {
+            // Treat HOME as an alias for USERPROFILE on Windows for legacy reasons
+            if (Platform::isWindows() && $matches[2] == 'HOME') {
+                return (getenv('HOME') ?: getenv('USERPROFILE')) . $matches[3];
+            }
             return getenv($matches[2]) . $matches[3];
         }, $path);
     }
--- a/src/Composer/Util/Platform.php
+++ b/src/Composer/Util/Platform.php
@@ -20,10 +20,36 @@ namespace Composer\Util;
 class Platform
 {
     /**
+     * Parses magic constructs like tildes in paths. Right now only tildes are supported but we could add support for 
+     * environment variables on various platforms.
+     *
+     * @param string $path
+     * @return string
+     */
+    public static function expandPath($path)
+    {
+        // Tilde expansion for *nix
+        if (!self::isWindows() && 0 === strpos($path, '~/')) {
+            if (function_exists('posix_getuid') && function_exists('posix_getpwuid')) {
+                $info = posix_getpwuid(posix_getuid());
+                $home = $info['dir'];
+            } else {
+                $home = getenv('HOME');
+            }
+            // Cannot be empty or FALSE
+            if (!$home) {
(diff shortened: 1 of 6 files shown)


src/Composer/Util/Platform.php
```diff
@@ -20,10 +20,36 @@
...
+     * Parses magic constructs like tildes in paths. Right now only tildes are supported but we could add support for 
+     * environment variables on various platforms.
+     *
+     * @param string $path
+     * @return string
+     */
+    public static function expandPath($path)
+    {
+        // Tilde expansion for *nix
+        if (!self::isWindows() && 0 === strpos($path, '~/')) {
```
staabm: couldnt we also just use a ENV var on windows to make it also work on windows?

curry684: It's no convention on Windows, and `%appdata%` corresponds to a folder that isn't even easily findable. So yeah, of course we could make it work but I doubt there's a use case.

Seldaek: I'd rather use `getenv('HOME')` on windows if it's present, and default to `getenv('USERPROFILE')` otherwise. Some (like myself..) run in cygwin or similar setups where HOME is available, and I don't see why we shouldn't make use of it.

Basically it could implement the logic from https://github.com/composer/composer/blob/1.0/src/Composer/Config.php#L211 and replace that line by a Platform::expandPath call as well.

Seldaek: And FWIW regarding windows dev, I have everything under %USERPROFILE% because that makes it a lot easier to backup all my data, I don't need to escalate permissions, and it just makes sense these days IMO given the dir layout of windows. Dumping everything in C:\X is a bit old school ;)

src/Composer/Util/Platform.php
```diff
@@ -20,10 +20,36 @@
...
+
+    /**
      * @return bool Whether the host machine is running a Windows OS
      */
     public static function isWindows()
     {
         return defined('PHP_WINDOWS_VERSION_BUILD');
     }
-}
+ }
```
stof: this should be reverted

curry684: Hmm no clue how that came in there - just switched to phpStorm on Ubuntu and some key combos still kill me hehe. I'll pick it up alongside Seldaek's comments.

src/Composer/Util/Platform.php
curry684: Yeah I got al my personal crap there, but we have a sort of 'company convention' that all git checkouts go to `C:\Projects` because we've been doing that for 10 years and it just works hehe.

I'll add the Win support later. Putting the function in Platform was intentional indeed to make it usable in a (lot) more locations where it makes sense.

tests/Composer/Test/ConfigTest.php
```diff
@@ -151,7 +150,7 @@ public function testVarReplacement()
 
         $home = rtrim(getenv('HOME') ?: getenv('USERPROFILE'), '\\/');
         $this->assertEquals('b', $config->get('c'));
-        $this->assertEquals($home.'/', $config->get('bin-dir'));
```
curry684: @Seldaek I had to change this test to fit new behaviour, but it seems to me it was just plain wrong before in expecting the trailing slash while the next line doesn't?

Seldaek:
Cool thanks, I had to tweak the regex a bit because OCD but all good ;)

PR #5191, opened 2016-04-13
Observe COMPOSER environment variable in init command

micaherne:
The init command creates a file called "composer.json" even when the COMPOSER environment variable is set to a different filename. This means that subsequent commands like install will result in an error message saying that composer could not find the config file.

It would make sense for init to observe the COMPOSER environment variable so that this does not happen.

To test / reproduce: `COMPOSER=composer-other.json php composer.phar init`. Without patch, composer.json file created, with patch composer-other.json.

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -48,9 +48,11 @@ class InitCommand extends BaseCommand
      */
     protected function configure()
     {
+        $configFile = Factory::getComposerFile();
+
         $this
             ->setName('init')
-            ->setDescription('Creates a basic composer.json file in current directory.')
+            ->setDescription('Creates a basic ' . $configFile . ' file in current directory.')
             ->setDefinition(array(
                 new InputOption('name', null, InputOption::VALUE_REQUIRED, 'Name of the package'),
                 new InputOption('description', null, InputOption::VALUE_REQUIRED, 'Description of package'),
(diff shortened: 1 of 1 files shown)


src/Composer/Command/InitCommand.php
```diff
@@ -48,9 +48,11 @@ class InitCommand extends BaseCommand
      */
     protected function configure()
     {
+        $configFile = Factory::getComposerFile();
+
         $this
             ->setName('init')
-            ->setDescription('Creates a basic composer.json file in current directory.')
+            ->setDescription('Creates a basic ' . $configFile . ' file in current directory.')
```
alcohol: This should not be modified.

You are still creating a basic composer.json, you are merely specifying a different name.

src/Composer/Command/InitCommand.php
```diff
@@ -65,7 +67,7 @@ protected function configure()
                 new InputOption('repository', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Add custom repositories, either by URL or using JSON arrays'),
             ))
             ->setHelp(<<<EOT
-The <info>init</info> command creates a basic composer.json file
+The <info>init</info> command creates a basic $configFile file
```
alcohol: Same as previous comment.

src/Composer/Command/InitCommand.php
```diff
@@ -181,7 +183,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
         // namespace
         $io->writeError(array(
             '',
-            'This command will guide you through creating your composer.json config.',
+            'This command will guide you through creating your ' . Factory::getComposerFile() . ' config.',
```
alcohol: Same as previous comments.

Seldaek:
Merged as 830644f374fb1697c7e7230bac4a9ea9919d8102, thanks

PR #5193, opened 2016-04-14
Unix/Windows compatibility in binary file path

sjawhar:
When `bin-compat` is set to "full", running `composer install` from Windows generates a different .bat file than when running it in Linux. `BinaryInstaller::generateWindowsProxyCode()` currently trims double quotes from `$binPath` when creating the Windows .bat file in the binary directory. The problem is that, on Linux, $binPath is surrounded by single quotes, not double. So, running `composer install` on Linux would create a .bat file that did not work on Windows (which is a problem if you want to share the binaries, like on a volume mounted in a Docker container). See the screenshot below for an example, notice the quotes on the third line.

image

Both single and double quotes are being trimmed now, and the .bat file created is the same if created on Windows or on Linux.

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -173,7 +173,7 @@ class BinaryInstaller
 
         return "@ECHO OFF\r\n".
             "setlocal DISABLEDELAYEDEXPANSION\r\n".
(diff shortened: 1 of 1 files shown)


Seldaek:
Merged as 9d2db57f65c77839b38f172787af0edad0d0b6c2, thanks :)

Issue #5173, opened 2016-04-10
secure-http option bites with git-type urls git://


ktomk:
I'm using composer as  a dependency and just moved it to ^1.0.0. It must have been at an older state where the secure-http option was not yet available.

Now when downloading a git-type package with a git://... URL, I get the following error:

> Your configuration does not allow connections to git://github.com/OpenMage/magento-mirror.git. See https://getcomposer.org/doc/06-config.md#secure-http for details.

Which does not make much sense to me as it's not HTTP.

The error has been recorded: https://travis-ci.org/ktomk/n98-magerun/jobs/122019645

The error message however is correct when HTTP URLs are used (here with xdebug and more verbose): https://circleci.com/gh/ktomk/n98-magerun/253

Just wanted to report it as the error message does not make much sense to me. I understand this is non-standard composer use.

ktomk:
As I need to have HTTP in this case, I had set the secure-http option in config to false. Now the error reported does not appear any longer (before_script section in https://travis-ci.org/ktomk/n98-magerun/jobs/122022682) which is expected.

And just seeing that **git** is listed as a known insecure protocol:

~~~
        // Throw exception on known insecure protocols
        if (in_array($parsed['scheme'], array('http', 'git', 'ftp', 'svn'))) {
            throw new TransportException("Your configuration does not allow connections to $url. See https://getcomposer.org/doc/06-config.md#secure-http for details.");
        }
~~~

Therefore closing as bogus.

curry684:
I'll clarify as I typed that comment hehe. Please refer to https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols:

> Next is the Git protocol. This is a special daemon that comes packaged with Git; it listens on a dedicated port (9418) that provides a service similar to the SSH protocol, but with **absolutely no authentication**. [...]

> The downside of the Git protocol is the lack of authentication. It’s generally undesirable for the Git protocol to be the only access to your project. Generally, you’ll pair it with SSH or HTTPS access for the few developers who have push (write) access and have everyone else use git:// for read-only access. 

As Composer isn't just about *read* access, but actually *downloading code that you intend to put on production servers* using plain `git://` is just well.... pretty much terribly insecure, and therefore restricted.

ktomk:
[...]

Issue #3338, opened 2014-10-09
vcs (git) + file:// URIs makes composer throw different kind of exceptions


ktomk:
Exceptions in issue:

1. [RuntimeException] Directory does not exist: <path>
2. [ErrorException] proc_open(): CreateProcess failed, error code - 267

I don't know if this a regression (#3285 <del>#3294</del>).

> Composer version 73e9db5d9952d52a46ecbc20a269a8c5f9c5b011 2014-10-07 15:03:19

Taking the following composer.json as base:

    {
        "repositories": [
            {
                "type": "vcs",
                "url": "file:///c:/Users/DB/PhpstormProjects/Net_URL2/.git/"
            }
        ],
        "require": {
            "pear/net_url2": "dev-topics/composer-for-pear"
        }
    }

will report this exception:

    $ composer install
    Loading composer repositories with package information

      [RuntimeException]
      Directory does not exist: /c:/Users/DB/PhpstormProjects/Net_URL2/.git/

    [...]

at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Repository/Vcs/GitDriver.php:230

albeit:

    $ php -r 'var_dump(is_dir("file:///c:/Users/DB/PhpstormProjects/Net_URL2/.git/"));'
    bool(true)

This was the first problem I encountered. Removing the trailing slash:

    ...
    "repositories": [
        {
            "type": "vcs",
            "url": "file:///c:/Users/DB/PhpstormProjects/Net_URL2/.git"
        }
    ],
    ...

 makes composer wanting to do more:

    $ composer install
    Loading composer repositories with package information
    
    
    
      [ErrorException]
      proc_open(): CreateProcess failed, error code - 267
    
    [...]

This now is at

Executing command (/c:/Users/DB/PhpstormProjects/Net_URL2/.git): git show-ref --tags

    Exception trace:
     () at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/vendor/symfony/process/Symfony/Component/Process/Process.php:287
     Composer\Util\ErrorHandler::handle() at n/a:n/a
     proc_open() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/vendor/symfony/process/Symfony/Component/Process/Process.php:287
     Symfony\Component\Process\Process->start() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/vendor/symfony/process/Symfony/Component/Process/Process.php:20
    4
[...]

PR #5195, opened 2016-04-15
Support alternative scp-like syntax in http-secure check

ktomk:
Also support ssh:// URIs.

References:

- Report: #5173

- URL syntax: https://www.git-scm.com/docs/git-clone#_git_urls_a_id_urls_a

- Scheme syntax: http://tools.ietf.org/html/rfc3986#section-3.1

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -14,6 +14,7 @@ namespace Composer;
 
 use Composer\Config\ConfigSourceInterface;
 use Composer\Downloader\TransportException;
+use Composer\Package\Url;
 
 /**
  * @author Jordi Boggiano <j.boggiano@seld.be>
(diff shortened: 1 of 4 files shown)


ktomk:
@Seldaek: I think this PR is more interesting atm. ;)

Right now it's not possible to use `ssh:// ...` URLs or URLs like `git:Department/Repo.git` w/o disabling http-secure. The report is a bit hidden inside the (non-) issue I reported initially.

curry684:
Isn't this an awful lot of extra code just for the edgy case that someone decides to name their SSH alias after one of the four known insecure protocols? Surely we can solve this with less than 154 lines of code?  :wink:

ktomk:
Well it's perhaps because the original author wasn't able to create a one-liner PCRE expression that did the check firsthand ;)

If you think it's useful to inline at the current place in time, I'm happy to do so, it should reduce the non-test code to ca. 10-15 lines. I somewhat had the feeling that there might be other places (e.g. via. #3338) that in the end there is more benefit of a new type. Also a new type is easier to unit-test, but that's more showing my lazyness.

So just let me know if you would like to have this inlined (or as a PCRE one-liner even).

Seldaek:
Fixed in d5158d943f60c54e0964fc8febed37171f3e345a

PR #5204, opened 2016-04-16
Show channel even if there is no new version on the current channel

peterrehm:
At the moment using composer self-update you just get the message that you are on the current version:

````
$ composer self-update
You are already using composer version 1.0.0.
````

This PR adds the channel information so that a user can decide if he wants to select another channel.


Seldaek:
Thanks

Issue #5186, opened 2016-04-13
[ErrorException]   chmod(): No such file or directory


sneha2011:
Hi Everyone,

On Installing symfony via composer I am getting following error on command prompt : 

[ErrorException]
  chmod(): No such file or directory

What I have done :  Have downloaded and installed composer in subfolder name i.e project under root directory of wamp server but on running : 
composer create-project symfony/framework-standard-edition my_project_name 

I am getting  exception No such file or directory

Kindly help.

PS : I am new to symfony
Thanks in advance

Seldaek:
Please run it again with like `composer create-project symfony/framework-standard-edition my_project_name -vvv` (with -vvv!), that way you get a stack trace and maybe we can help, because that error alone is really not specific enough.

sneha2011:
Thanks Seldaek - Please find the Exception trace below :

Exception trace:

 () at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Util/RemoteFilesystem.php:811

 Composer\Util\ErrorHandler::handle() at n/a:n/a

 chmod() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Util/RemoteFilesystem.php:811

 Composer\Util\RemoteFilesystem->getTlsDefaults() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Util/RemoteFilesystem.php:63

 Composer\Util\RemoteFilesystem->__construct() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Factory.php:567

 Composer\Factory::createRemoteFilesystem() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Repository/RepositoryFactory.php:101

 Composer\Repository\RepositoryFactory::defaultRepos() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Command/CreateProjectCommand.php:255

 Composer\Command\CreateProjectCommand->installRootPackage() at 
phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Command/CreateProjectCommand.php:155

Composer\Command\CreateProjectCommand->installProject() at phar://C:/ProgramData/ComposerSetup/bin/composer.phar/src/Composer/Command/CreateProjectCommand.php:142

[...]

Issue #4797, opened 2016-01-19
composer asset plugin can't install


Leffka:
Hello all. 
I want to set yii2.
Install composer, no problem.
In yiiframework.com written : "install the Composer Asset Plugin"
In console I write : composer global require "fxp/composer-asset-plugin:~1.1" and get : 
scrin shot http://pixs.ru/showimage/1jpg_2622563_20311454.jpg
where problem? 
I user Windows 7, php 5.6, apache 2.4

alcohol:
Sorry, no clue. Looks like a CA error to me. Does your php have ssl support?

Leffka:
Yes, in php.ini : extension=php_openssl.dll and folder '/php/ext' have file php_openssl.dll

alcohol:
Can you retry with the latest Composer version?

Leffka:
I can only download new version composer from getcomposer.org
who owns the composer on Windows? You can find /tmp/composer-cacert.pem in his system? I will be happy if you just give a link to the file, if it exists.

alcohol:
https://getcomposer.org/download/ bottom of the page.

curry684:
Your screenshot shows the `composer-cacert.pem` file being accessed from the `/tmp/` folder, which of course doesn't exist on Windows. That path is constructed in https://github.com/composer/composer/blob/master/src/Composer/Util/RemoteFilesystem.php#L785 with:

    $targetPath = rtrim(sys_get_temp_dir(), '\\/') . '/composer-cacert-' . $hash . '.pem';

Therefore `sys_get_temp_dir` is returning `/tmp` on your system. On Windows this should be `C:\Windows\Temp` or something instead which is actually compatible with Windows filesystems.

*However*, this can be overridden in 2 ways:
 1. In your `php.ini` file with the `sys_temp_dir` option. You can check this in the output of `phpinfo()`
 2. In your environment with the `TMPDIR` variable. You can check this by entering `echo %TMPDIR%` in a command prompt

In both cases this is a system misconfiguration on your dev machine and not an issue on Composer's end.

Leffka:
Thank you! 
in php.ini : sys_temp_dir = "/tmp"
system variable $tmpdir don't exist

I did in php.ini : sys_temp_dir = "C:/Windows/temp"
in cmd : composer global require "fxp/composer-asset-plugin:~1.1"
and download is done.

curry684:
[...]

PR #5205, opened 2016-04-16
Check temp folder usability before runtime

curry684:
Extra startup check as #4797 and #5186 show this can cause exotic hard to find errors otherwise.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -150,9 +150,8 @@ class Application extends BaseApplication
 
             // Check system temp folder for usability as it can cause weird runtime issues otherwise
             Silencer::call(function() {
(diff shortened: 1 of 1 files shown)


alcohol:
Why not use [`tmpfile()`](http://php.net/manual/en/function.tmpfile.php) ?

curry684:
You'd still need to do the separate write/read/delete checks, and I'm not sure (it's not documented) that `tmpfile` is guaranteed to return a file within `sys_get_temp_dir()`. The mentioned issues depended on code explicitly using `sys_get_temp_dir` downloads and files.

Seldaek:
Merged but I switched it to just print an error instead of failing hard.

PR #5208, opened 2016-04-17
Support empty directories with tar-files, closes #5206

### LINKED ISSUE #5206: Tar Archives Need Parity with Zip Archives
In [this accepted pull request](https://github.com/composer/composer/pull/5144#issuecomment-205021044), the behavior of archiving has been changed such that zip files are now archiving empty folders. This is a welcome change, as it removes some barriers to using Satis with Magento 2's official Composer repository, as well as other repositories of projects that rely on empty directories.

Is a similar change needed for the tar archives as well? i.e. that pull request only changes the behavior of zip files.  This fixes the Magento problem since their repository is using ZIP archives, but it seems like the composer project would want its archiving systems behaving identically -- i.e. if zip files archives and unarchive empty directories, then tar archives should do the same things.  

If the Magento repository ever switched to tar archives, or a user (like me) attempted to create a satis mirror with tar archives instead of zip files, "weird things" might happen.  I did some initial tests and, when I created projects from a satis mirror using tar archives created from Magento's repository, the empty folders were represented as empty, 0 byte regular files.  I can work around this by mirroring with ZIPs -- but the fundamental difference between the archiving methods remain.

Apologies for not having a pull request on this, or a better test description/reproduction steps.  I'm not familiar enough with the composer and satis projects to tackle that without breaking things, and wanted to make sure I'm not making any derpy assumptions before I go to the trouble of creating a reproducible test case.

ktomk:
Compared with buildFromIterator this is much slower.

- Report: #5206

- Ref: 8389b4b

### DIFF
--- a/src/Composer/Package/Archiver/ArchivableFilesFilter.php
+++ b/src/Composer/Package/Archiver/ArchivableFilesFilter.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Package\Archiver;
+
+use FilterIterator;
+use PharData;
+
+class ArchivableFilesFilter extends FilterIterator
+{
+    private $dirs = array();
+
+    /**
+     * @return bool true if the current element is acceptable, otherwise false.
+     */
+    public function accept()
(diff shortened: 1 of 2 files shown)


staabm:
How much slower?

ktomk:
@staabm: Depends a lot on the amount of files. I have run a test with magento/magento2-base (2.0.4) and it's like one minute versus 20 minutes. I guess this also has to do with file-system and I/O speeds.

The good news is, you don't run into fopen limits *gg*.

I have another idea to filter out directories and add them later.

ktomk:
@staabm: Updated the PR, new approach is not slower any longer.

PR #5210, opened 2016-04-18
Fix partial updates from a dirty state and apply exactly the locked state

### LINKED ISSUE #3468: Partial update is broken
I have a composer.json and a composer.lock with a couple of packages and empty vendor directory.

When I run 
composer update sonata-project/notification-bundle sonata-project/media-bundle jms/serializer-bundle knplabs/knp-paginator-bundle oldsound/rabbitmq-bundle

I have all my bundles updated, not only the listed ones.

Behaviour is similar to https://github.com/composer/composer/issues/782

Seldaek:
Need a sanity check here (looking at the test files). I think it's better like this. The removed packages come back if they are still in the lock, which is maybe a bit WTF but is actually correct and I doubt it'll be much of an issue in practice. Other side-effects updates or packages that were in lock but not currently installed should not happen anymore.

Fixes #3468

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -450,7 +450,7 @@ class Installer
                     $candidates[$link->getTarget()] = true;
                     $rootRequires[$link->getTarget()] = $link;
                 }
-                foreach ($localRepo->getPackages() as $package) {
+                foreach ($currentPackages as $package) {
                     $candidates[$package->getName()] = true;
                 }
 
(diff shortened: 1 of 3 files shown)


stof:
:+1: for this

nikic:
Is the referenced issue #3468 still open intentionally?

Seldaek:
@nikic nope I think it can be closed now

PR #5220, opened 2016-04-19
Convert documentation to ordered list

MichaelFoss:
Switch the dependency management documentation
to an ordered list from a pseudo-ordered list. This
has the benefit of using more structured markdown,
and the added benefit of matching the "Suppose"
list items with corresponding "Composer" list items.

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -18,15 +18,13 @@ This idea is not new and Composer is strongly inspired by node's
 
 Suppose:
 
-a) You have a project that depends on a number of libraries.
-
-b) Some of those libraries depend on other libraries.
+1. You have a project that depends on a number of libraries.
+1. Some of those libraries depend on other libraries.
 
 Composer:
 
-c) Enables you to declare the libraries you depend on.
-
-d) Finds out which versions of which packages can and need to be installed, and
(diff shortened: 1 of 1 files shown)


doc/00-intro.md
```diff
@@ -18,15 +18,13 @@ This idea is not new and Composer is strongly inspired by node's
 
 Suppose:
 
-a) You have a project that depends on a number of libraries.
-
-b) Some of those libraries depend on other libraries.
+1. You have a project that depends on a number of libraries.
+1. Some of those libraries depend on other libraries.
```
Seldaek: Maybe best to use `2.` here (and below)? :) Otherwise yeah +1.

MichaelFoss: Yeah, I thought so too at first @Seldaek, and I used to use it that way too. However, if ever there were more lines to be added to the documentation, you'd have to renumber everything.

Markdown supports ordered lists simply by placing a `1.` in front of each line – doing this allows for easy addition, deletion, and reordering. Imagine if we had a list of 200 items!

Seldaek: Yeah except we won't ever have 200 here, and it makes it weird if you read the markdown directly, but fair enough, it's not a huge deal anyway so I'll merge ;)

PR #5221, opened 2016-04-20
Fix addMainKey method of JsonManipulator with content having $n

danieliancu:
This will allow a dollar sign followed by a digit to be used in the password stored in auth.json file.
If the content variable contains a dollar sign followed by a digit ("$n", where n consists of digits) it will match as a replacement reference for preg_replace and thus it will be replaced by the text captured by the n'th paranthesized pattern or blank if there's no such paranthesized pattern. A replacement reference can also be in the form of "\\\n", but "$n" is preferred (and also mentioned in the PHP documentation), being made available since PHP 4.0.4.

### DIFF
--- a/src/Composer/Json/JsonManipulator.php
+++ b/src/Composer/Json/JsonManipulator.php
@@ -353,7 +353,7 @@ class JsonManipulator
         if ($this->pregMatch('#[^{\s](\s*)\}$#', $this->contents, $match)) {
             $this->contents = preg_replace(
                 '#'.$match[1].'\}$#',
-                addcslashes(',' . $this->newline . $this->indent . JsonFile::encode($key). ': '. $content . $this->newline . '}', '\\'),
+                addcslashes(',' . $this->newline . $this->indent . JsonFile::encode($key). ': '. $content . $this->newline . '}', '\\$'),
                 $this->contents
             );
 
@@ -363,7 +363,7 @@ class JsonManipulator
         // append at the end of the file
         $this->contents = preg_replace(
             '#\}$#',
-            addcslashes($this->indent . JsonFile::encode($key). ': '.$content . $this->newline . '}', '\\'),
+            addcslashes($this->indent . JsonFile::encode($key). ': '.$content . $this->newline . '}', '\\$'),
             $this->contents
         );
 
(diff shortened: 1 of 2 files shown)


alcohol:
Please include a unit test for this. There are several tests available for the json manipulator already :-)

danieliancu:
@alcohol Done.

danieliancu:
Related #1588.

alcohol:
Oh god the escaping of `\` on both input and output is confusing the crap out of me haha. Damn, I think the tests are good. But honestly I'm not really sure.

Basically a string such as `\1bar`, needs to be set/passed in using `\\1bar`, and is stored as `\\\1bar`. Correct?

danieliancu:
I encountered this issue while trying to authenticate with a password having a replacement reference (in the form of `$n`) which got stored incorrectly in the `auth.json` file (`$n` being stripped because there's no such parenthesized pattern). This was a pretty easy fix, I just added the `$` character in the `charlist` parameter of the `addcslashes` call within the `addMainKey` method of the `JsonManipulator` class.

Now putting this one aside, I discovered that replacement references are not correctly handled in various places throughout the `JsonManipulator` class. But it may not be an issue because the use case is somewhat limited. I don't know all of the composer internals, but please correct me if I'm wrong.

So for the `$n` form which may be found in passwords (especially those created using password generators) the above fix will do. I don't think there's another use case, or not yet discovered. For `\\n` (and also `\n` because of the escaping) the current fix in place outlined below will do, because again it doesn't look like there are other use cases and regarding passwords, not even generators use the `\` character.

Due to escaping (`'\1'` being the same as `'\\1'`, both giving `'\1'` PHP string, note the use of `'`) and thus `preg_replace` is very permissive with replacement references in the `replacement` parameter, because it allows the use of both `\n` and `\\n`. When PR #1588 came around it was solved by 435cde0, so now all `\` will become `\\` while `\\` will remain the same. This means that if content is `'\1bar'` or `'\\1bar'` it will always get stored as `'\\1bar'` because of the `addcslashes` call.

Not having other use cases of replacement references within strings handled by the `JsonManipulator` and having solved the one that I encountered (which is very legit), I have gone ahed and only applied the fix for the issue I encountered and also updated the tests accordingly.

I hope that I have made it clear now.

Sorry for the length of the reply, I just love making things clear once and not just for myself but for others as well.

ping @alcohol

alcohol:
Don't be sorry, it's great to see such dedication :-)

Looks all good to me now. Thanks!

Seldaek:
Thanks! Merged :)

Note that it wasn't fixed in *all* addcslashes though so I did it more systematically, as it was surely a problem in other cases even though maybe less likely to have `$` signs.

PR #5224, opened 2016-04-20
Always resolve dev packages even when doing an update with --no-dev

### LINKED ISSUE #5016: Ignore --no-dev option for dependency resolution & lock file generation
As per https://github.com/composer/composer/issues/5013#issuecomment-193034939 I think the `--no-dev` option should only impact the installation part of the `update` command. The resolution and lock file generation should always include dev packages, to ensure consistent generation of the lock file, regardless of weather `--no-dev` is passed or not. The option should remain, to choose whether dev packages should really be downloaded and installed.

As a side effect this should simplify some of the code in `Composer\Installer`.

Seldaek:
Fixes #5016

Again, quite big changes in core code so could use a review. It's pretty well tested though so I don't expect major issues.

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -34,6 +34,8 @@ use Composer\IO\IOInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\CompletePackage;
 use Composer\Package\Link;
+use Composer\Package\Loader\ArrayLoader;
+use Composer\Package\Dumper\ArrayDumper;
 use Composer\Semver\Constraint\Constraint;
 use Composer\Package\Locker;
 use Composer\Package\PackageInterface;
@@ -192,13 +194,6 @@ class Installer
         $this->downloadManager->setPreferSource($this->preferSource);
         $this->downloadManager->setPreferDist($this->preferDist);
 
-        // clone root package to have one in the installed repo that does not require anything
-        // we don't want it to be uninstallable, but its requirements should not conflict
-        // with the lock file for example
-        $installedRootPackage = clone $this->package;
-        $installedRootPackage->setRequires(array());
-        $installedRootPackage->setDevRequires(array());
-
         // create installed repo, this contains all local packages + platform packages (php & extensions)
         $localRepo = $this->repositoryManager->getLocalRepository();
         if ($this->update) {
@@ -207,15 +202,7 @@ class Installer
             $platformOverrides = $this->locker->getPlatformOverrides();
         }
         $platformRepo = new PlatformRepository(array(), $platformOverrides);
-        $repos = array(
-            $localRepo,
-            new InstalledArrayRepository(array($installedRootPackage)),
-            $platformRepo,
-        );
-        $installedRepo = new CompositeRepository($repos);
-        if ($this->additionalInstalledRepository) {
-            $installedRepo->addRepository($this->additionalInstalledRepository);
-        }
+        $installedRepo = $this->createInstalledRepo($localRepo, $platformRepo);
 
         $aliases = $this->getRootAliases();
         $this->aliasPlatformPackages($platformRepo, $aliases);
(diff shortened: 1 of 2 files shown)


src/Composer/Installer.php
```diff
@@ -916,15 +1010,14 @@ private function processDevPackages($localRepo, $pool, $policy, $repositories, $
...
      * @param  RepositoryInterface $installedRepo
      * @return array
      */
-    private function getCurrentPackages($withDevReqs, $installedRepo)
+    private function getCurrentPackages($installedRepo)
     {
         if ($this->locker->isLocked()) {
             try {
-                return $this->locker->getLockedRepository($withDevReqs)->getPackages();
+                return $this->locker->getLockedRepository(true)->getPackages();
```
stof: do we still need the argument ?

Seldaek: Yeah because it defaults to false, and old lock files might not be up to speed with those changes and we shouldn't break on them (it's called 3lines below without arg which is false)

src/Composer/Installer.php
```diff
@@ -604,7 +568,117 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
...
+    }
+
+    /**
+     * Extracts the dev packages out of the localRepo
+     *
+     * This works by faking the operations so we can see what the dev packages
+     * would be at the end of the operation execution. This lets us then remove
+     * the dev packages from the list of operations accordingly if we are in a
+     * --no-dev install or update.
+     *
```
naderman: Isn't there an easier way to do this based on the lock file? I mean at this point you're guaranteed to know which packages are in the dev section and which are in the non-dev section of the lock file? So just see which operations relate to which section?

naderman: Sorry confused over the order, this is actually used to build the lock file :-)

Seldaek: Yup the only alternative I believe would be to walk the dev requires 
recursively, and then walk the requires recursively, and find packages 
that are only in the dev require tree, but with replaces and all that I 
am scared of introducing bugs, so rather do it the way it was done 
before, by leveraging the solver.

naderman:
Looks good to me @Seldaek

Seldaek:
Alright then let's go with it :)

jpuck:
I think [the documentation](https://getcomposer.org/doc/03-cli.md#install) deserves an update following this change.

> --no-dev: Skip installing packages listed in require-dev. The autoloader generation skips the autoload-dev rules.

This is somewhat misleading for me considering installation will fail in the absence of `composer.lock` if the dev dependency resolution exceeds the environment regardless of the runtime requirements.

naderman:
@jpuck I'm not really sure what you meant, that statement is still correct. If you have a suggestion, please open a new issue and describe the problem you see in detail, thanks.

jpuck:
Hi @naderman, I didn't mean to imply that the statement is incorrect, just that it might benefit from an amendment. I'm no communications expert, but I was imagining something as simple as this example:

> --no-dev: Skip installing packages listed in require-dev. The autoloader generation skips the autoload-dev rules. **Note: without a composer.lock, the require-dev section will still attempt to be resolved, even though the packages will not actually be downloaded.**

Not only would something like this encourage the use of a `composer.lock` file in VCS, but it should provide a hint to others with failed installations receiving the error message:

> Your requirements could not be resolved to an installable set of packages.
> 
>  Problem 1
>    - some package only listed in `require-dev` section

It seems to me that the first natural instinct in this situation is to think that the `--no-dev` flag isn't working.

Just my two cents for what it's worth. I'm a huge fan btw - not trying to complain - just trying to help.

Thanks,
Jeff P.

naderman:
@jpuck I didn't meant to say that you were wrong, just that it needed more explanation and that it should be posted in a new issue, so people other than myself will actually see it. I appreciate that you're trying to make this easier to understand for Composer users :-) It would really help if you could create a new issue for this and just post it there. Thank you!

naderman:
Or really just submit a PR with the documentation change you propose to https://github.com/composer/composer/tree/master/doc Cheers

PR #5243, opened 2016-04-25
Gitlab driver bugfix

msarca:
Fixed a bug that prevented interactive authentication to work properly


alcohol:
Thanks, small oversight on our part it seems :)

Issue #5244, opened 2016-04-25
Arbitrary command injection when run as root (but not via sudo)


derekmarcotte:
Hi there.  While looking for an unrelated issue, I see that SUDO_UID is passed un-sanitized from the environment when composer is run as root, which allows arbitrary command execution.

Granted, if you are root - that's your fault.  Probably not correct behaviour nonetheless.

With the following composer.json:

```json
{
    "name": "derek/composer-break",
    "authors": [
        {
            "name": "Derek Marcotte",
            "email": "derek.marcotte@experiencepoint.com"
        }
    ],
    "require": {},
    "scripts": {
      "dummy": [ "echo Hi dummy" ]
    }
}
```

When I run this command:

```
SUDO_UID="0 touch HITHERE;" composer dummy
```

I get this output:

```
ls -l:

ls -l
total 1668
-rw-r--r--  1 root   wheel        0 Apr 25 14:55 HITHERE
-rw-rw-r--  1 derek  wheel      257 Apr 25 14:21 composer.json
-rw-rw-r--  1 derek  wheel  1629436 Apr 25 14:12 composer.phar

```

And I expected this to happen:

Not to run any command I add to the environment

PR #5245, opened 2016-04-25
fix command injection from the environment when run as root, issue #5244

derekmarcotte:
A suggested fix for #5244


curry684:
Good catch :+1:

Seldaek:
Yup, nice catch indeed, thanks.

ktomk:
- Introduced: 557a55f
- Ref: dfd22c6 / #5122

PR #5246, opened 2016-04-26
doc - some rephrasing for global cli install

bamarni:
This is a follow up from https://github.com/bamarni/symfony-console-autocomplete/pull/16. I also added a command snippet to help locating that folder.

Should I submit this to `1.0` or master? I'm confused by all this stablility 😄

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -234,17 +234,21 @@ directory.
 This is merely a helper to manage a project stored in a central location that
 can hold CLI tools or Composer plugins that you want to have available everywhere.
 
-This can be used to install CLI utilities globally and if you add
-`$COMPOSER_HOME/vendor/bin` to your `$PATH` environment variable. Here is an
-example:
+This can be used to install CLI utilities globally. Here is an example:
 
 ```sh
-php composer.phar global require fabpot/php-cs-fixer:dev-master
+php composer.phar global require fabpot/php-cs-fixer
 ```
 
-Now the `php-cs-fixer` binary is available globally (assuming you adjusted
-your PATH). If you wish to update the binary later on you can just run a
-global update:
+Now the `php-cs-fixer` binary is available globally. Just make sure your global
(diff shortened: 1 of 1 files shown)


alcohol:
You broke the example. A lot of our users simply copy/paste. You modified the `$COMPOSER_HOME` environment variable to no longer include the `$`, thus making it a plain string.

bamarni:
This is intended. Here it shouldn't feel like there is something to be copy/pasted imo, as having this env var is an edge case.

stof:
the code snippet is meant to be ``$COMPOSER_HOME/vendor/bin``, not just ``/vendor/bin``. So IMO, this change is not good as is.

bamarni:
Why should we assume this variable is set? And if we do, why not `$COMPOSER_HOME/$COMPOSER_BIN_DIR `?

My suggestion is actually not to show it as a "code snippet", as I believe most people don't define this variable.

alcohol:
That's fine, but then this example in itself does not make any sense anymore. Cause what it is showing now is literally invalid information.

bamarni:
You're right that this PR doesn't make it something ready for copy/paste, my intention is instead to show the first part as a placeholder. Is there any change I can do so you'd stop calling this a "code snippet" or an "example"? Maybe putting curly brackets around the link?

Here I'm just trying to turn `$COMPOSER_HOME/vendor/bin` into something more meaningful for the majority of users.

Seldaek:
The way to retrieve the bin-dir reliably is to run `composer global config bin-dir --absolute` - maybe use this to introduce that directory instead of the COMPOSER_HOME dummy example?

bamarni:
Nice one, I've added this command instead.

As a side note, it seems that the documentation in getcomposer.org is built from the master branch and not 1.0, is it expected?

Seldaek:
Yeah master is more up to date and IMO most people should run from snapshots on dev machines (which is where you interact with composer). Thanks for the updates :)

PR #5249, opened 2016-04-26
Add plugin Compatible interface

Soullivaneuh:
The goal of this interface is to permit developers to specify
which minimum version of composer is compatible with the plugin.

Since composer has now stable version (Good job BTW!), I think this will be necessary.

The idea come from my own implementation on my plugin: https://github.com/Soullivaneuh/composer-versions-check/blob/4beadb69804f6d0fd7bdcd67b4b20bb8fa47296b/src/VersionsCheckPlugin.php#L54-L66

The best thing would be to add it directly on `PluginInterface` but this will be BC break.

@Seldaek I could work on it when you'll push a 2.x-dev branch on this repo.

See also: https://github.com/composer/composer/pull/5028#issuecomment-214698485

- [x] Compatible interface
- [ ] Add tests

### DIFF
--- a/src/Composer/Plugin/Compatible.php
+++ b/src/Composer/Plugin/Compatible.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
 
 namespace Composer\Plugin;
 
--- a/src/Composer/Plugin/Compatible.php
+++ b/src/Composer/Plugin/Compatible.php
@@ -1,0 +1,17 @@
+<?php
+
+
+namespace Composer\Plugin;
+
+/**
+ * Implement this interface on your plugin class to specify minimum composer version required to work with.
+ *
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+interface Compatible
+{
+    /**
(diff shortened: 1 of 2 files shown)


Seldaek:
You should just require composer-plugin-api in the version that you need, and composer won't install the plugin at all if it doesn't match the correct version. This has to be done at the solver level so that we can install the proper plugin version once we have let's say composer v1 and v2, and the plugin has multiple versions for each. Version checks should not be done as a hack like this, nor in your plugin IMO.

Soullivaneuh:
This is not totally right IMHO because I had by the past some error with older (and unstable) version of composer with the same api-plugin version.

So, how works the api-plugin version in this case? how to solve it?

Soullivaneuh:
BTW, in some case we could need some classes from `Composer` and not the `Plugin` part. So this could change not regarding the plugin-api version, right?

Seldaek:
Now that we are stable, we won't do BC breaks until 2.0, so the plugin api is reliable IMO. Doesn't matter which classes you use.

Soullivaneuh:
@Seldaek how about this? 

> BTW, in some case we could need some classes from `Composer` and not the `Plugin` part. So this could change not regarding the plugin-api version, right?

Maybe the plugin installer should check `composer-plugin-api` **and** `composer/composer` version constraint?

Seldaek:
No if we provide composer/composer then packages requiring it won't 
install it anymore. That's not possible.

Just require the plugin-api of the composer version you target and done 
IMO :)

Soullivaneuh:
Well, I think I'll just ignore the fact of before 1.0 version releases of composer could break my plugin. :-)

PR #5254, opened 2016-04-27
Added HTTP support to GitLab driver

msarca:
Added HTTP support to GitLab driver by introducing the `gitlab-unsecure-domains` config option. This option allows GitLab API to be accessed through HTTP for the specified domains.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -253,6 +253,13 @@
                         "type": "string"
                     }
                 },
+                "gitlab-unsecure-domains": {
+                    "type": "array",
+                    "description": "A list of domains used in gitlab mode that doesn't have support for HTTPS protocol. Defaults to [].",
+                    "items": {
+                        "type": "string"
+                    }
+                },
                 "archive-format": {
                     "type": "string",
                     "description": "The default archiving format when not provided on cli, defaults to \"tar\"."
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -385,6 +385,18 @@ EOT
                     return $vals;
                 },
             ),
+            'gitlab-unsecure-domains' => array(
+                function ($vals) {
+                    if (!is_array($vals)) {
+                        return 'array expected';
+                    }
+
+                    return true;
+                },
+                function ($vals) {
+                    return $vals;
+                },
+            ),
         );
 
         foreach ($uniqueConfigValues as $name => $callbacks) {
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -54,6 +54,7 @@ class Config
         'capath' => null,
         'github-expose-hostname' => true,
         'gitlab-domains' => array('gitlab.com'),
+        'gitlab-unsecure-domains' => array(),
         'store-auths' => 'prompt',
         'platform' => array(),
         'archive-format' => 'tar',
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -81,6 +81,10 @@ class GitLabDriver extends VcsDriver
         $this->owner = $match['owner'];
         $this->repository = preg_replace('#(\.git)$#', '', $match['repo']);
 
+        if (in_array($this->originUrl, (array) $this->config->get('gitlab-unsecure-domains'))) {
+            $this->scheme = 'http';
+        }
+        
         $this->cache = new Cache($this->io, $this->config->get('cache-repo-dir').'/'.$this->originUrl.'/'.$this->owner.'/'.$this->repository);
 
         $this->fetchProject();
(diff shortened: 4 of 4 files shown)


stof:
-1 for that. the scheme is already extracted from the repository URL. Just provide an HTTP url and it does not need extra configuration

msarca:
And how about the repositories that have SSH URLs? In this case the protocol used to access the API would be HTTPS and there's no option that allows you to change that.

GromNaN:
Every GitLab repository has 2 URLS: The HTTP and the SSH. If your instance runs without SSL, you have to specify the HTTP URL of the repository so that the [GitLabDriver can detect the scheme](https://github.com/composer/composer/blob/1.0/src/Composer/Repository/Vcs/GitLabDriver.php#L79).

```json
    "repositories": [
        { "type": "gitlab", "url": "http://gitlab.example.com/mirror/symfony_http_foundation.git" }
    ],
```

Is there something that doesn't work this way ?

sorinsarca:
Use case: if you use gitlab on your own server (which can or cannot be accessible from Internet) and you have private repos on it.

stof:
@sorinsarca this does not answer the question of @GromNaN. Is there anything which prevents you from registering the repository using an HTTP url ?

msarca:
@stof , @GromNaN the problem here is choice. By forcing me to call the API through HTTPS when using SSH URLs, you leave me without options. For example, If I have a custom gitlab setup that doesn't have support for SSL, but I'm planning to support it in the future, I could simply add a `github-unsecure-domains` config key to my Packagist instance and the API will be accessed through HTTP. That way I could add my packages by using SSH URLs and when I decide to add support for HTTPS, I would simply remove the `github-unsecure-domains` config key instead of edit each package separately.

stof:
@msarca IMO, it is better to reconfigure your packages to use a secure URL once your gitlab instance is secure rather than adding a new option in composer which will need to be maintained for the following years by the composer team and understood by anyone reading the doc. Any extra configuration option has a cost by asking composer users to understand the choice they have to make for the config setting (and saying they don't need to configure it is a bad argument, because they need to understand what the option is about to be able to know that they don't need it)

msarca:
@stof The feature is useful and I'm sure it will help not only me, but also other developers to stop wasting precious time explaining and trying to convince a customer that using SSL(from the beginning) would be the best option. Since composer is a tool that aims to make developers' life easier, I don't see any reason why this feature would not be available starting with version 1.1

stof:
@msarca if they don't use SSL, why not simply registering the repository with its HTTP URL, as this is what you want it to use ? This is my main point: it is **already** possible to tell the driver to use HTTP to access gitlab. And you still have not explained why the existing solution is not good for you.

msarca:
@stof Because the repositories were already registered using SSH and I should register again hundreds of repositories, which is just a waste of time. And when the customer will provide SSL support, I will be forced to register again hundreds of repositories. 
I agree, it is already possible to tell the driver to use HTTP, but in a very **limited** way.

stof:
@msarca if it is your own packagist instance, you could perform a single SQL query doing it in batch

sorinsarca:
Yes, or you can just set the proposed option to config, so the future repositories registered in packagist can still have SSH urls.

stof:
@sorinsarca but this requires adding a new config option in Composer (with the associated cost for the project, as explained above) just to solve your edge case

sorinsarca:
- composer is open-source and very popular, so there will always be someone to support it (if remains popular)
- developers using Gitlab and composer will support the gitlab driver for composer, or they will switch to something else like github, bitbucket
- the option is proposed to fix a composer limitation for gitlab driver, this option has minimal impact on composer because it only affects the gitlab driver (without side effects)
- now, probably this is my edge case, but if gitlab will gain popularity this will also be the case for many others

Honestly I don't care if this PR is merged or not, because so far it looks like "we are open-source but not open-minded".

msarca:
@stof  This is not an edge case, it's simply a composer limitation. I just want to be able to choose whether to use HTTP protocol when I'm dealing with repos that have SSH URLs or just stick with the default HTTPS.

stof:
> developers using Gitlab and composer will support the gitlab driver for composer

That's not true in practice. The composer core team is the one having to support all the code in practice, even when someone else contributed it initially. Your statement is only true in a dram world.

> this option has minimal impact on composer because it only affects the gitlab driver (without side effects)

This option then needs to be documented, which then means that someone reading about how to configure composer has one more option to understand.
Each new option adds more complexity, so it always has a cost. The question is whether the benefit of the new option outweight the cost. In this case, I'm arguing that it does not, as the case of a gitlab instance not supporting SSL can already be handled today without this option.

sorinsarca:
> This option then needs to be documented, which then means that someone reading about how to configure composer has one more option to understand.

That is not true because this option is specific for gitlab driver. So far there are two options: gitlab-domains and gitlab-oauth. I think that someone can read one more option for gitlab.

slbmeh:
Perhaps a solution using existing configuration would be to incorporate a fallback strategy when https fails if `secure-http` is set to false?

sorinsarca:
@slbmeh this sounds more like general solution. 
But I think it would be nicer to have the ability to specify the default scheme, because we'll avoid unnecessary steps like first checking for https, then for http and so on, and the whole process will speed up. The PR is just about this - the ability to specify the default scheme (for gitlab driver).

Seldaek:
Fixed by https://github.com/composer/composer/commit/40df5dddee7322ff12f05e15770f18fc425fde8d which lets you do `{"type":"vcs", "url": "git@insecure-gitlab.org:....", "secure-http": false}`.

GromNaN:
Thanks @Seldaek. That is the best way of solving this issue.

PR #5267, opened 2016-04-29
Ignore it for unfound packages

kamranahmedse:
In `outdated` command, when checking for the latest packages, it will fail in case it is unable to find the latest version.

In [Composer\Command\ShowCommand.php#L280](https://github.com/composer/composer/blob/master/src/Composer/Command/ShowCommand.php#L280), the method call `$latestPackage->getFullPrettyVersion()` will throw an exception as `$latestPackage` might be `false` when there is no "bestCandidate". See [VersionSelector.php#L64](https://github.com/composer/composer/blob/master/src/Composer/Package/Version/VersionSelector.php#L64).

Failing composer outdated

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -275,7 +275,12 @@ EOT
                         if ($showVersion) {
                             $versionLength = max($versionLength, strlen($package->getFullPrettyVersion()));
                             if ($showLatest) {
+
                                 $latestPackage = $this->findLatestPackage($package, $composer, $phpVersion);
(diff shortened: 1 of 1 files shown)


Seldaek:
Do you have an example of how to trigger this? Which package can't it find and why?

kamranahmedse:
It seemed to be failing when installing this https://github.com/symfony/yaml. But after I cleared the cache and updated packages `composer update`. It seems to be working fine now and I am no longer able to reproduce this.

Seldaek:
OK well I'll still merge as it's better to bulletproof the code, thanks.

Issue #5160, opened 2016-04-06
Cloning git repo on bitbucket and oauth2


vaclavvanik:
Hi,

I have problem with cloning git repo on bitbucket. I have set consumer key and secret in auth.json:

```
{
    "http-basic": {},
    "github-oauth": {},
    "gitlab-oauth": {},
    "bitbucket-oauth": {
        "bitbucket.org": {
            "consumer-key": "...",
            "consumer-secret": "..."
        }
    }
}
```

and I got

```
  - Installing vendor/repo (dev-master e408024)
    Cloning e408024dac5098371ae75f1de4acd6005a65a1bd
Executing command (CWD): git clone --no-checkout 'https://bitbucket.org/vendor/repo.git' '/home/venca/PhpstormProjects/my-project/vendor/vendor/repo' && cd '/home/venca/PhpstormProjects/my-project/vendor/vendor/repo' && git remote add composer 'https://bitbucket.org/vendor/repo.git' && git fetch composer
Downloading https://bitbucket.org/site/oauth2/access_token
Executing command (CWD): git clone --no-checkout 'https://x-token-auth:****@bitbucket.org/vendor/repo.git.git' '/home/venca/PhpstormProjects/my-project/vendor/vendor/repo' && cd '/home/venca/PhpstormProjects/my-project/vendor/vendor/repo' && git remote add composer 'https://x-token-auth:****@bitbucket.org/vendor/repo.git.git' && git fetch composer
Executing command (CWD): git --version
Failed: [RuntimeException] Failed to execute git clone --no-checkout 'https://x-token-auth:***@bitbucket.org/vendor/repo.git.git' '/home/venca/PhpstormProjects/my-project/vendor/vendor/repo' && cd '/home/venca/PhpstormProjects/my-project/vendor/vendor/repo' && git remote add composer 'https://x-token-auth:***@bitbucket.org/vendor/repo.git.git' && git fetch composer

Cloning into '/home/venca/PhpstormProjects/my-project/vendor/vendor/repo'...
remote: Not Found
fatal: repository 'https://x-token-auth:<token>@bitbucket.org/vendor/repo.git.git/' not found
```
It seems repo address has doubled .git extension. Is it bug or I have bad configuration anywhere?

I am using satis for handling private repos. Satis.json looks:

```
{
    "name": "my name",
    "homepage": "https://example.com",
    "repositories": [
        {"type": "vcs", "url": "https://bitbucket.org/vendor/repo.git"}
    ],
[...]

PR #5272, opened 2016-05-01
Bitbucket https fallback to ssh

stefangr:
This change provides a fallback to SSH when using https URLs for bitbucket.org.

A typical situations where this is needed include having a non-interactive install of dependencies in a location where oauth authentication (if it actually works) is not configured.

This also includes a fix for the problem with double .git extension metioned in #5160.

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -112,7 +112,7 @@ class Git
                         return;
                     }
                 }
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

dnetix:
This is not a fix, this is a fallback, if I could use SSH, well, I wouldn't be using https

PR #5275, opened 2016-05-02
document plugin capabilities (closes #5262)

### LINKED ISSUE #5262: Document CommandProvider capability
See https://github.com/composer/composer/tree/master/tests/Composer/Test/Plugin/Fixtures/plugin-v8/Installer for an example.

bamarni:


nevvermind:
Hey, @bamarni. I have a couple of suggestion here:

a) Maybe we should expand more on what exactly these plugin capabilities are. It's not exactly clear from that description;

b) I don't think we should provide a snippet for capabilities. That would mean to provide one for every future one, too; and that will considerably clutter the Plugins page. As they are interfaces, they should be their own documentation, so I would rather you augment their phpdocs [here](https://github.com/composer/composer/blob/761d6a549c229befce49863bf3d282b4a35bbbbe/src/Composer/Plugin/Capability/CommandProvider.php#L18-L21) and [here](https://github.com/composer/composer/blob/761d6a549c229befce49863bf3d282b4a35bbbbe/src/Composer/Plugin/Capable.php#L25-L31). This could also help in avoiding duplicating documentation;

c) your `My\Composer\Plugin` must implement `Composer\Plugin\Capable`, so that snippet's incorrect;

Conclusion: 
- I would add a more substantial description of what a Plugin Capability mechanism is;
- I would add a snippet showing how a Capable Plugin must declare the `getCapabilities` method in particular, as that's the tricky part
- I would just create a list like this:

`<example>`
Composer defines a standard set of capabilities which may be implemented by plugins through the [`Composer\Plugin\Capable`][8] interface:
- [CommandProvider](https://github.com/composer/composer/blob/master/src/Composer/Plugin/Capability/CommandProvider.php)  

`</example>`

bamarni:
Thanks for the feedback,

a) It's actually not even clear for me either :) as there is only one capability at the moment I could only come up with that generic sentence. What do you think is missing?

b) I think there should be an example here and not only a link to an interface, even though it'd contain phpdocs. There is also a full example above (for event listeners).

c) Indeed 👍  just added that interface

nevvermind:
a) No problem. Here's home history on this: https://github.com/composer/composer/pull/4124

b) Yes, we should have an example of a Capable plugin (we just have one such plugin capability atm), just like events. But imagine that Composer, in time, will have 5. Or 10. Would you document _every_ one of those capabilities in this page?

I'll try to help when I get a little bit of time.

bamarni:
a) I see, so how about extending the sentence like this? "The goal of plugin capabilities is to provide well-defined extension points and make plugins define their features explicitly, instead of manually modifying the internal state of `Composer\Composer`".

b) When other capabilities are added I'd indeed rather not provide full examples for all of them, but only snippets. And keep only one full example as a reference.

nevvermind:
How about this - https://github.com/nevvermind/composer/commit/dc3050af9e0ec31dc9b3d0973cedd30f6e305220? 
Pick stuff from there or not, but that's how I would've done it. Snippets showing how to implement a particular _Command_ should not be in that section, in my opinion. Keep it generic.

bamarni:
IMO your changes contain some low-level details which are not needed : 
- mentioning it is available since plugin api 1.1, is it even the case? From the discussion you linked it appears to be there in 1.0 already.
- the section about PHP 5.5 `::class`, especially given the above example doesn't use this.

> Besides, plugins are very much developer-oriented, so they don't need too many words

I don't really agree here. Composer documentation as a whole is developer-oriented, it doesn't mean it's fine to show incomplete examples with links to the source code so people can figure things out. In our case here it's just ~10 extra lines to have a working example.

But as you mentioned, my generic explanations about plugin capabilities could be improved. Maybe adding some notes about the benefits of using them (clean extension points, less likely to break)?

nevvermind:
> mentioning it is available since plugin api 1.1, is it even the case? From the discussion you linked it appears to be there in 1.0 already.

You're right, it's available in 1.0. Got confused there for a second or ten.

bamarni:
Just made an update, moving the plugin snippet one level up, with one of your paragraph.

Seldaek:
Thanks :)

PR #5277, opened 2016-05-02
Allow commands to tell that they're proxy commands

bamarni:
reopened #5276 against master


Seldaek:
Thanks, I merged it on 1.1 as it's just an adjustment to the new feature.. and fixed one issue in 591cbcee12546ec1b86efa36779b7da99295d285

PR #5278, opened 2016-05-03
Added @dev which is not documented anywhere

alper:
I found this possibility on Stack Overflow and found it curiously missing from the documentation.

### DIFF
--- a/doc/articles/aliases.md
+++ b/doc/articles/aliases.md
@@ -19,6 +19,9 @@ If anyone wants to require the latest `dev-master`, they have a problem: Other
 packages may require `1.0.*`, so requiring that dev version will lead to
 conflicts, since `dev-master` does not match the `1.0.*` constraint.
 
+You can also put `@dev` in your version requirement if you want to get the latest
+version from source control.
+
 Enter aliases.
 
 ## Branch alias


doc/articles/aliases.md
```diff
@@ -19,6 +19,9 @@ If anyone wants to require the latest `dev-master`, they have a problem: Other
 packages may require `1.0.*`, so requiring that dev version will lead to
 conflicts, since `dev-master` does not match the `1.0.*` constraint.
 
+You can also put `@dev` in your version requirement if you want to get the latest
+version from source control.
```
stof: This will not ensure that you get the latest version. It will tell composer that you want *any* version (including new major versions) and that dev versions are allowed (changing the minimum-stability policy for this package instead of using the global setting). ``@dev`` is actually parsed as ``*@dev``.

and in any case, this section is not an appropriate place to put this sentence, as this doc chapter is about documenting aliases

alper:
Where should it go then?

Seldaek:
Fixed in https://github.com/composer/composer/commit/d3eece0af2938b62129643ba8c9464770a0ffcf6 but for reference it's already documented in https://getcomposer.org/doc/04-schema.md#package-links

alper:
Excellent, thank you!

PR #5303, opened 2016-05-10
@composer not work, composer instead

Webysther:


alcohol:
You are mistaken.

staabm:
@Webysther in case `@composer` doesnt work for you, you should self-update your composer binary. this feature was added a few months ago.

Webysther:
I use the latest version 1.0.3 on docker + gitlab ci build system. If i use ```@composer``` not work with local ```phar``` version or installed version. ```composer``` work for both cases.

Webysther:
Or! ```@composer``` works for 1.1.* version only, like i read right now on release version:

* Added ability to call composer itself from scripts via ```@composer``` ...

Documentation not reallity for 1.0.* version.

alcohol:
Using `composer` creates a subprocess (shell) that runs the composer binary which is located in /usr/local/bin (most likely). The new `@composer` syntax simply calls the command internally in the current running process.

Yes, documentation is a bit of a mess right now.

PR #5317, opened 2016-05-11
IPv6 Platform Requirement

### LINKED ISSUE #5304: Require IPv6 Support
PHP can be built with the `--disable-ipv6` option and there is no way to ensure that a package that requires IPv6 support forces an installer to actually having it enabled (or rather not disabled). The only way one could perform this validation would be at runtime:

```php
if (defined('AF_INET6') === false) {
    // Houston, we have a problem.
}
```

It would be nice to have the ability to specify this requirement in a libraries `composer.json` file and have composer validate it before the package is being installed. Maybe something along the lines:

```json
{
    "require": {
        "ext-network-ipv6": "*"
    }
}
```

Note that `ext-network` is always enabled in PHP, it is only possible to disable IPv6 support individually.

Fleshgrinder:
I decided to add two additional PHP interpreter packages instead of an additional extension key. IPv6 support is part of PHP core (if enabled) and is not part of any optional extension.

Would close #5304.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -77,29 +77,21 @@ class PlatformRepository extends ArrayRepository
             $version = $versionParser->normalize($prettyVersion);
         }
 
-        // The AF_INET6 constant is only defined if ext-sockets is available but IPv6 support might still be available.
-        $ipv6Support = defined('AF_INET6') || @inet_pton('::') !== false;
-
         $php = new CompletePackage('php', $version, $prettyVersion);
         $php->setDescription('The PHP interpreter');
         $this->addPackage($php);
-
-        if ($ipv6Support) {
-            $phpIpv6 = new CompletePackage('php-ipv6', $version, $prettyVersion);
-            $phpIpv6->setDescription('The PHP interpreter with IPv6 support');
-            $this->addPackage($phpIpv6);
-        }
 
         if (PHP_INT_SIZE === 8) {
             $php64 = new CompletePackage('php-64bit', $version, $prettyVersion);
             $php64->setDescription('The PHP interpreter, 64bit');
             $this->addPackage($php64);
+        }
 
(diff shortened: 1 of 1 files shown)


Fleshgrinder:
Just found the [`Silencer` util](https://github.com/composer/composer/blob/1.1/src/Composer/Util/Silencer.php#L54-L76); should I use that instead of PHP's _normal_ silencer?

src/Composer/Repository/PlatformRepository.php
```diff
@@ -77,14 +77,29 @@ protected function initialize()
...
+            $this->addPackage($phpIpv6);
+        }
+
         if (PHP_INT_SIZE === 8) {
             $php64 = new CompletePackage('php-64bit', $version, $prettyVersion);
             $php64->setDescription('The PHP interpreter, 64bit');
             $this->addPackage($php64);
+
+            if ($ipv6Support) {
+                $php64Ipv6 = new CompletePackage('php-64bit-ipv6', $version, $prettyVersion);
```
staabm: Why do we need 64bit-ipv6?

Fleshgrinder: Because a x64 build does not ensure that you have IPv6 support enabled too.

```json
{
    "require": {
        "php-ipv6": "^5.6 || ^7.0",
        "php-64bit": "^5.6 || ^7.0"
    }
}
```

Too much duplicated configuration on user side for my taste.

alcohol:
Please make them `ext-` keys. It does not make sense to offer multiple variants of the PHP binary interpreter other than 32 and 64 bit.

alcohol:
LGTM

dzuelke:
IMO this should be called `php-ipv6` and not `ext-network-ipv6`, in line with `php-64bit`.

Future "core build" properties, such as debug or ZTS yes/no, could follow the same naming scheme.

PR #5318, opened 2016-05-11
fix `http-basic` json example

MrMitch:
The JSON example for the `http-basic` option was missing the closing curly brace.


Seldaek:
Thanks

PR #5322, opened 2016-05-12
Updated the version

ron4stoppable:
updated the version to the latest

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -142,7 +142,7 @@ Close your current terminal. Test usage with a new terminal:
 
 ```sh
 C:\Users\username>composer -V
-Composer version 1.0.0 2016-01-10 20:34:53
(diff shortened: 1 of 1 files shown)


alcohol:
It's going to be tedious to update this on every new release. Maybe the output should just be made more generic.

staabm:
@alcohol this looks like something which only makes sense when automated.

when composer would support some kind of "release" command and a corresponding scripts-hook this could be implemented as part of the release process.

alcohol:
Considering our documentation isn't even tied to a specific version yet, this is not very feasible at the moment.

Seldaek:
Yeah I don't really see the point anyway.. it's obviously an example in the docs.

ron4stoppable:
okay, it should be more atomic version than like 1.0, 2.0, etc; right?

PR #5325, opened 2016-05-12
[Archive] exclude option in composer.json accepts files

digitalkaoz:
with this PR the "archive.exclude" accepts files to (beside the legacy? array)

```json
{
    "archive" :  {
        "exclude" : ".myignore"
    }
}
```
```
# file .myignore
/test
/build
```

that would exclude `test` and `build` from the archive.

i needed the feature for creating Docker Contexts from Composer Archives:

* my Application in Git itself should contain `doc` and `tests` (so `.gitignore` wont help)
* `.gitattributes` is unhandy here, since it would exclude those folders in normal artifacts too
* `.dockerignore` wont help because i build the context/archive in a `build` folder, but this folder is in `.gitignore`
* providing an ignore pattern list in a json array felt ugly to me, and if you ask me, i would deprecate that feature completely and maybe create an own `.composerignore` file, that would clean up a lot code since we can handle it like a normal `.gitignore` file

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -325,7 +325,7 @@
             "description": "Options for creating package archives for distribution.",
             "properties": {
                 "exclude": {
-                    "type": "array",
+                    "type": ["array", "string"],
                     "description": "A list of patterns for paths to exclude or include if prefixed with an exclamation mark."
                 }
             }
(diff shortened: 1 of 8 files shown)


alcohol:
I find this feature very ambiguous. I would instead expect it to ignore this one file in particular. Not using the contents for ignore patterns.

:-1:

digitalkaoz:
as i said, i would drop this feature completly and introduce some `.composerignore` file for such use cases

stof:
As this is a totally different behavior than when passing an array, it should be a separate key in ``archive`` instead of reusing the same one

Seldaek:
I'll close because IMO it's not really bringing much value, it'll just add another way to do things confusing matters even more, and we can't easily deprecate/remove this as it'd break existing packages.

PR #5327, opened 2016-05-12
cs cleanup

digitalkaoz:
it seems nobody is checking the CS before creating a PR ;)

maybe add https://styleci.io/ webhook to PRs ?

### DIFF
--- a/src/Composer/IO/BaseIO.php
+++ b/src/Composer/IO/BaseIO.php
@@ -121,8 +121,8 @@ abstract class BaseIO implements IOInterface, LoggerInterface
     /**
      * System is unusable.
      *
-     * @param string $message
-     * @param array $context
+     * @param  string $message
+     * @param  array  $context
      * @return null
      */
     public function emergency($message, array $context = array())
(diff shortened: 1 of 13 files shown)


src/Composer/IO/BaseIO.php
```diff
@@ -121,8 +121,8 @@ public function loadConfiguration(Config $config)
     /**
      * System is unusable.
      *
-     * @param string $message
-     * @param array $context
+     * @param  string $message
```
kamilsk: here is different https://github.com/composer/composer/pull/5327/files#diff-b0c94b96dc86be178b130cd557412587R713

digitalkaoz: mh the docblocks should be aligned...whats wrong?

staabm:
@digitalkaoz Seldaek will update/fix the CS from time to time. PRs to the repos will land even if they not stick to the CS.

src/Composer/IO/BaseIO.php
kamilsk: `@param string` vs `@param  string`, one space vs two space. why?

digitalkaoz:
@staabm ok, ill close it then. thanks for the info! and yes its a good practise :)

PR #5328, opened 2016-05-13
Output version changes in StatusCommand when possible

slbmeh:
@Seldaek this is a cleaner solution to #1968 than my PR in #3788 and also considers PathRepositories as a potential candidate for a version change.

### DIFF
--- a/src/Composer/Command/StatusCommand.php
+++ b/src/Composer/Command/StatusCommand.php
@@ -116,7 +116,7 @@ EOT
                                 'ref'     => $previousRef
                             ),
                             'current'  => array(
-                                'version' => $currentVersion['version'],
+                                'version' => $parser->normalize($currentVersion['version']),
                                 'ref'     => $currentVersion['commit'],
                             ),
                         );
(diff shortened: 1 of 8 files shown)


src/Composer/Command/StatusCommand.php
```diff
@@ -63,46 +73,84 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                            $previousRef = null;
+                    }
 
-                if ($downloader instanceof DvcsDownloaderInterface) {
-                    if ($unpushed = $downloader->getUnpushedChanges($package, $targetDir)) {
-                        $unpushedChanges[$targetDir] = $unpushed;
+                    $currentVersion = $guesser->guessVersion($dumper->dump($package), $targetDir);
+
+                    if ($previousRef && $currentVersion['commit'] !== $previousRef) {
+                        var_dump($previousRef, $currentVersion['commit'], $package->getName(), $package->getInstallationSource());
```
mbrodala: This line should be removed I guess.

slbmeh: Oops, I had deleted that and suppose I never committed it.

Seldaek:
Thanks, looking good :)

Issue #5166, opened 2016-04-07
bitbucket oauth not working


mmoll:
When I run this command:
```
composer command -vvv:
```

I get this output:
```
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Reading ./composer.json
Loading config file /home/mmoll/.composer/config.json
Loading config file /home/mmoll/.composer/auth.json
Loading config file ./composer.json
Checking CA file /etc/ssl/certs/ca-certificates.crt
Executing command (/tmp/things): git branch --no-color --no-abbrev -v
Failed to initialize global composer: Composer could not find the config file: /home/mmoll/.composer/composer.json
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
Reading /tmp/things/vendor/composer/installed.json
Reading ./composer.lock
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Reading ./composer.lock
Resolving dependencies through SAT
Dependency resolution completed in 0.008 seconds
Analyzed 180 packages to resolve dependencies
Analyzed 481 rules to resolve dependencies
  - Installing zzz/yyy (1.3.1)
    Cloning 86ee5af60d7c8718fdfaaa07d53e44c2d4adea6c
Executing command (CWD): git clone --no-checkout 'https://bitbucket.org/xxx/yyy.git' '/tmp/things/vendor/zzz/yyy' && cd '/tmp/things/vendor/zzz/yyy' && git remote add composer 'https://bitbucket.org/xxx/yyy.git' && git fetch composer
Executing command (CWD): git config bitbucket.accesstoken
Enter your Bitbucket credentials to access private repos
Follow the instructions on https://confluence.atlassian.com/bitbucket/oauth-on-bitbucket-cloud-238027431.html
to create a consumer. It will be stored in "/home/mmoll/.composer/auth.json" for future use by Composer.
Consumer Key (hidden): 
Consumer Secret (hidden): 
Downloading https://bitbucket.org/site/oauth2/access_token
Failed: [Composer\Downloader\TransportException] The 'https://bitbucket.org/site/oauth2/access_token' URL could not be accessed: HTTP/1.1 400 BAD REQUEST


                                                                                                            
  [Composer\Downloader\TransportException]                                                                  
  The 'https://bitbucket.org/site/oauth2/access_token' URL could not be accessed: HTTP/1.1 400 BAD REQUEST  
                                                                                                            

Exception trace:
[...]

PR #5335, opened 2016-05-14
Update bitbucket oauth instructions

carlcraig:
I had similar issues described in #5166 and after investigation, I found out that Bitbucket needs you to set **Callback URL** in the **OAuth consumer** configuration to be able to generate access tokens.

If you leave this field blank the api `https://bitbucket.org/site/oauth2/access_token` will give the following error:

```
{
  "error_description": "No callback uri defined for the OAuth client.",
  "error": "invalid_request"
}
```

The grant type `client_credentials` will not actually use this callback URL, but it seems Bitbucket is enforcing that it is configured anyway.

This PR updates the log messages to inform the user that the `Callback URL` needs to be set to something. *Ensure you enter a "Callback URL" or it will not be possible to create an Access Token (this callback url will not be used by composer)*


alcohol:
Interesting, this should be further investigated.

peec:
I essentialy gave up on adding oauth authorization with bitbucket, but thanks to Carl i got this working by adding some random url to Callback URL in the github settings page! Now I just use:

+1

stof:
@peec the end of the comment seems to be missing

peec:
@stof updated with details on how its fixed now

Seldaek:
Thanks

PR #5342, opened 2016-05-15
Fix for a fatal error when system temp dir is not available

smottt:
When the temp dir is not available / writable, a fatal error occurs because a variable is not available in the function scope.

This is a simple fix accompanied by a unit test.

However, I'm not very comfortable with using `eval` for injecting `sys_get_temp_dir` function into the namespace. Comments on improving the test are welcome.

### DIFF
--- a/tests/Composer/Test/ApplicationTest.php
+++ b/tests/Composer/Test/ApplicationTest.php
@@ -113,6 +113,9 @@ class ApplicationTest extends TestCase
         $this->ensureNoDevWarning('self-up');
     }
 
+    /**
+     * @runInSeparateProcess
+     */
     public function testNoFatalErrorOnInvalidSysTempDir()
     {
         $inputMock = $this->getMock('\Symfony\Component\Console\Input\InputInterface');
--- a/tests/Composer/Test/ApplicationTest.php
+++ b/tests/Composer/Test/ApplicationTest.php
@@ -112,4 +112,49 @@ class ApplicationTest extends TestCase
     {
         $this->ensureNoDevWarning('self-up');
     }
+
+    public function testNoFatalErrorOnInvalidSysTempDir()
+    {
+        $inputMock = $this->getMock('\Symfony\Component\Console\Input\InputInterface');
+        $inputMock->expects($this->at(0))
+            ->method('getParameterOption')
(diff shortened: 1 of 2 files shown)


alcohol:
It's good that you spotted/found this bug. Not sure if the test is a good idea though as it relies on some awkward behaviour. I would suggest just fixing the bug only :-)

/cc @Seldaek your input on this?

Seldaek:
Yeah this test basically means the temp dir becomes invalid for all tests following it.. we don't run a lot of tests through the application so it might not impact tests right now but still sounds like not a great idea for the future. I'd rather have the bugfix alone as well.

tests/Composer/Test/ApplicationTest.php
```diff
@@ -112,4 +112,52 @@ public function testDevWarningPreventedAlias()
...
+
+                $errorMessage = $message;
+            })
+        ;
+
+        eval(<<<CODE
+namespace Composer\Console {
+    function sys_get_temp_dir()
+    {
+        return '/composer-tmp-dir';
```
stof: you should write the mock in a way allowing it to call the original function in most cases, to avoid breaking things when writing other tests in this namespace in the future. See https://github.com/symfony/symfony/blob/ea9d6e7235adb98cf6143cb89e6cdb2e86710944/src/Symfony/Bridge/PhpUnit/DnsMock.php for an example of doing that

curry684:
Heh was terribly confused by this as I was sure I'd tested this scenario, but the fatal bug is actually in the later change from an exception to a warning. :+1:  for the change itself.

As for the test I agree it's too fragile this way. Would probably be far easier to just `ini_set('sys_temp_dir', '/foo/bar');` during this test, ensuring teardown functions clean it up. Not really sure it warrants a real test though.

Seldaek:
Committed the one-liner fix as https://github.com/composer/composer/commit/360cb4ac2af350c8b96b2fd56b67ef9d5589836a

tests/Composer/Test/ApplicationTest.php
smottt: Thanks for the note.

smottt:
Thanks for merging this and for the comments. Cheers

Issue #5339, opened 2016-05-14
RFC: Do not suggest to avoid exact version constraints for unstable packages


localheinz:
With the following composer.json:

```json
{
    "name": "foo/bar",
    "description": "Foos all the bars",
    "type": "library",
    "license": "MIT",
    "require": {
        "php": "^5.5 || ^7.0",
        "sentry/sentry": "0.17.0",
    }
}
```

When I run this command:

```
$ composer validate
```

I get this output:

```
/composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
require.sentry/sentry : exact version constraints (0.17.0) should be avoided if the package follows semantic versioning
```

And I expected this to happen:

```
/composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
```

💁 Following #5338, `composer` should not suggest to use the `^` operator if I pin an unstable package to an exact version.

#### Version

```
$ composer --version
Composer version 1.1.0 2016-05-10 15:21:19
```

Seldaek:
This one makes sense I think. :+1: if you want to PR :)

localheinz:
@Seldaek 

Sure, will do, later today!

PR #5343, opened 2016-05-15
Fix: Do not suggest to avoid exact version constraints for unstable package

localheinz:
This PR

* [x] asserts that no warning is issued when using an exact version constraint for an unstable package
* [x] issues a warning only when the version for which an exact constraint is used isn't stable

Fixes https://github.com/composer/composer/issues/5339.

### DIFF
--- a/tests/Composer/Test/Package/Loader/ValidatingArrayLoaderTest.php
+++ b/tests/Composer/Test/Package/Loader/ValidatingArrayLoaderTest.php
@@ -329,6 +329,18 @@ class ValidatingArrayLoaderTest extends \PHPUnit_Framework_TestCase
             array(
                 array(
                     'name' => 'foo/bar',
+                    'require' => array(
+                        'bar/unstable' => '0.3.0',
+                    ),
+                ),
+                array(
+                    // using an exact version constraint for an unstable version should not trigger a warning
+                ),
+                false,
+            ),
+            array(
+                array(
+                    'name' => 'foo/bar',
                     'extra' => array(
                         'branch-alias' => array(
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Package/Loader/ValidatingArrayLoaderTest.php
```diff
@@ -315,6 +315,7 @@ public function warningProvider()
                         'bar/foo' => 'dev-master',
                         'bar/hacked' => '@stable',
                         'bar/woo' => '1.0.0',
+                        'bar/unstable' => '0.3.0',
```
alcohol: Considering the context of this test, this entry seems out of place. At least at a comment describing that this entry specifically should **not** trigger a warning (hence why I feel it does not belong in this test, since it explicitly tests each entry for the warning it should trigger).

alcohol:
LGTM

tests/Composer/Test/Package/Loader/ValidatingArrayLoaderTest.php
```diff
@@ -329,6 +329,18 @@ public function warningProvider()
...
                     'name' => 'foo/bar',
+                    'require' => array(
+                        'bar/unstable' => '0.3.0',
+                    ),
+                ),
+                array(
+                    // using an exact version constraint for an unstable version should not trigger a warning
+                ),
+                false,
+            ),
```
localheinz: @alcohol 

How do you like this now?

Seldaek:
Thanks

localheinz:
@Seldaek 

Thanks for having me!

PR #5360, opened 2016-05-19
ISSUE 3568 Show downloading package speed

stratdes:
* Added new class Download to represent the download progress
* Added the util class DownloadUnitsFormatter to format Download class speed and ETA values
* Some refactor on RemoteFileSystem

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -13,6 +13,7 @@
 namespace Composer\Util;
 
 use Composer\Config;
+use Composer\Downloader\Download;
 use Composer\IO\IOInterface;
 use Composer\Downloader\TransportException;
 use Composer\CaBundle\CaBundle;
@@ -28,13 +29,11 @@ class RemoteFilesystem
     private $io;
     private $config;
     private $scheme;
-    private $bytesMax;
     private $originUrl;
     private $fileUrl;
     private $fileName;
     private $retry;
     private $progress;
-    private $lastProgress;
     private $options = array();
     private $peerCertificateMap = array();
     private $disableTls = false;
@@ -44,6 +43,9 @@ class RemoteFilesystem
     private $degradedMode = false;
     private $redirects;
     private $maxRedirects = 20;
+
+    /** @var Download */
+    private $download;
 
     /**
      * Constructor.
@@ -196,12 +198,10 @@ class RemoteFilesystem
         }
 
         $this->scheme = parse_url($fileUrl, PHP_URL_SCHEME);
-        $this->bytesMax = 0;
         $this->originUrl = $originUrl;
         $this->fileUrl = $fileUrl;
         $this->fileName = $fileName;
         $this->progress = $progress;
-        $this->lastProgress = null;
         $this->retryAuthFailure = true;
         $this->lastHeaders = array();
         $this->redirects = 1; // The first request counts.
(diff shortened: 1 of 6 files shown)


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -529,17 +534,17 @@ protected function callbackGet($notificationCode, $severity, $message, $messageC
...
 
             case STREAM_NOTIFY_FILE_SIZE_IS:
-                $this->bytesMax = $bytesMax;
+                $this->getActiveDownload()->start($bytesMax);
                 break;
 
             case STREAM_NOTIFY_PROGRESS:
-                if ($this->bytesMax > 0 && $this->progress) {
-                    $progression = min(100, round($bytesTransferred / $this->bytesMax * 100));
+                if ($this->getActiveDownload()->isStarted() > 0 && $this->progress) {
```
staabm: IsStarted() > 0?

staabm: Assign the return obj of getActiveDownload would ease reading

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -529,17 +534,17 @@ protected function callbackGet($notificationCode, $severity, $message, $messageC
...
+                    $this->getActiveDownload()->progress($bytesTransferred);
 
-                    if ((0 === $progression % 5) && 100 !== $progression && $progression !== $this->lastProgress) {
-                        $this->lastProgress = $progression;
-                        $this->io->overwriteError("    Downloading: <comment>$progression%</comment>", false);
-                    }
+                    $progression = $this->getActiveDownload()->getProgressPercentage();
+                    $speed = $this->getActiveDownload()->getSpeedFormatted();
```
staabm: Feels a bit overengineered

staabm:
Thx for the PR. The feature itself is really cool :-)

stratdes:
> IsStarted() > 0?
You are right, brain disconnection.
> Assign the return obj of getActiveDownload would ease reading
If you take a look at the method, you'll see not only return the object, but also initialize it the first time. The problem is no DI in this project, so no proper testable way to operate with instances. In fact, there's not too much unit testing in this class, anyway, but you know...
> Feels a bit overengineered
Maybe. I think overengineer is a very slippery concept. In my opinion, this way to work let you the chance to add some features in a very easy way, like show the elapsed time for a download, for instance. Is scalable. Again, In my opinion :)
> Thx for the PR. The feature itself is really cool :-)
My pleasure to be able to contribute to this project, which is used by almost anyone everyday.

staabm:
> If you take a look at the method, you'll see not only return the object, but also initialize it the first time. The problem is no DI in this project, so no proper testable way to operate with instances. In fact, there's not too much unit testing in this class, anyway, but you know...

its a private method and after you called it once in line 541 you can assume that it wont go away until the end of the switch-case block is reached (I am not talking about removing `getActiveDownload` across this entire file)

stratdes:
> its a private method and after you called it once in line 541 you can assume that it wont go away until the end of the switch-case block is reached (I am not talking about removing getActiveDownload across this entire file)

The problem is for sure on RemoteFilesystemTest.php. I need to do some tricky shit to allow use of the method or put getActiveDownload public or protected, but I don't think this method has to be public or protected conceptually speaking. Also I could allow use of the attribute in the test and not the method, but using the method allows you to change the logic if needed. This entire class needs a refactor, but it was out of the scope.

stratdes:
Actually I have to fix the IsStarted thing so maybe I change the getActiveDownload method to protected and then mock the method. Any more ideas?

Seldaek:
Thanks for the work. However I'd rather not do this for the time being at least. A large overhaul of the RemoteFilesystem will be needed and adding more code to it will just make that harder.. I'll leave it open for now but please don't spend more time on it as chances are high it'll be wasted :/

PR #5363, opened 2016-05-21
All bitbucket (https) communication with OAuth

stefangr:
Currently downloading the source with https bitbucket urls configured works as intended.
However, downloading a zip file with the source is not.

This PR implements the OAuth way for all dist downloads.

### DIFF
--- a/src/Composer/Repository/Vcs/GitBitbucketDriver.php
+++ b/src/Composer/Repository/Vcs/GitBitbucketDriver.php
@@ -31,6 +31,10 @@ class GitBitbucketDriver extends VcsDriver implements VcsDriverInterface
     protected $branches;
     protected $rootIdentifier;
     protected $infoCache = array();
+    /**
+     * @var GitDriver
+     */
+    protected $gitDriver;
 
     /**
      * {@inheritDoc}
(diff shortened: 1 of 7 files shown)


dynasource:
great work. Would be great to be able to download zips to speed up deploys as Composer + Bitbucket + Git cloning is by far the biggest bottleneck

src/Composer/Repository/Vcs/GitBitbucketDriver.php
```diff
@@ -29,6 +31,10 @@ class GitBitbucketDriver extends VcsDriver implements VcsDriverInterface
     protected $branches;
     protected $rootIdentifier;
     protected $infoCache = array();
+    /**
+     * @var GitDriver
+     */
+    protected $gitDriver;
```
stof: I suggest making this new property private. Private things are easier to maintain as they are not covered by any backward compatibility concern and so can change anytime we need it (while protected stuff are harder to change outside of major versions)

src/Composer/Repository/Vcs/GitBitbucketDriver.php
```diff
@@ -167,4 +197,76 @@ public static function supports(IOInterface $io, Config $config, $url, $deep = f
...
+
+    /**
+     * Get the remote content.
+     *
+     * @param string $url The URL of content
+     * @param bool $fetchingRepoData
+     *
+     * @return mixed The result
+     */
+    protected function getContents($url, $fetchingRepoData = false)
```
stof: As this method does not have the same signature than the parent, I think it should actually be a different method rather than overwriting the signature

Seldaek:
Thanks @stefangr, this will hopefully help resolve a lot of bitbucket related issues :)

dynasource:
works great  @stefangr. Thanks! The time reduction is impressive.

PR #5380, opened 2016-05-27
Fixed indentation

yosmanyga:


Seldaek:
We print these without indent so they stand out even more, which IMO is good. And also because there might be contexts where a URL gets accessed out of the package installing process and thus the indent wouldn't make sense.

yosmanyga:
And what about using the correct indentation and some special characters at the beggining to stand out?

Seldaek:
See the second part of my reply..

PR #5383, opened 2016-05-28
Warn users more strongly when running as root, fixes #5382

### LINKED ISSUE #5382: Clarify in strong wording that Composer will run shell scripts
Composer has a feature that allows third-party packages to run arbitrary command lines on your system during installation. I do not complain here that the feature should be removed. However, I will show that such a feature is unintuitive and Composer should more clearly warn users of the security risks. **Patch will follow.**

I have used Composer for years as well as other dependency management / download tools. Only recently I learned that Composer can *execute* third-party code. Previously I thought that Composer simply *downloaded* code and that you executed it by running your web server (or however else you run it).

To illustrate, here is Composer's own installation instructions:

```bash
php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
php -r "if (hash_file('SHA384', 'composer-setup.php') === '92102166af5abdb03f49ce52a40591073a7b859a86e8ff13338cf7db58a19f7844fbc0bb79b2773bf30791e935dbd938') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"
php composer-setup.php
php -r "unlink('composer-setup.php');"
```

It is **perfectly safe** to run this as root, so long as you sudo to a non-privileged user while running the `php composer-setup.php` line.

However, basic usage of Composer as root, like this:

    $ composer require corepos/composer-installer

would cause your system to be rooted, if the author of that package, or its dependencies, so wished.

This is not true for other dependency management / download tools. You can safely run CocoaPods, Swift Package Manager, Git, cURL and Wget as root to fetch third party content and they are designed not to care about or trust the contents of what they are fetching.

---------------

Now you know. Composer warns users that run it as root with this line:

> Running composer as root/super user is highly discouraged as packages, plugins and scripts cannot always be trusted

This is not clear enough. It uses passive tone and it is not strong enough to explain why Composer is a bigger risk than the other tools I listed above. I recommend instead:

> Do not run composer as root/super user! When installing packages and doing other things with Composer, those packages have full access to your computer. See https://goo.gl/N7bjkY

fulldecent:


src/Composer/Console/Application.php
```diff
@@ -176,7 +176,7 @@ public function doRun(InputInterface $input, OutputInterface $output)
             if (!Platform::isWindows() && function_exists('exec') && !getenv('COMPOSER_ALLOW_SUPERUSER')) {
                 if (function_exists('posix_getuid') && posix_getuid() === 0) {
                     if ($commandName !== 'self-update' && $commandName !== 'selfupdate') {
-                        $io->writeError('<warning>Running composer as root/super user is highly discouraged as packages, plugins and scripts cannot always be trusted</warning>');
+                        $io->writeError('<warning>Do not run composer as root/super user! When installing packages and doing other things with Composer, those packages have full access to your computer. See https://goo.gl/N7bjkY</warning>');
```
staabm: Should this better link to a page from composer.org?

alcohol:
I disapprove of this PR.

PR #5384, opened 2016-05-29
Faster git clones using cache

marvil07:
I tried #4685 but it seems to not being creating the git mirror, so I extended it a little, on top of @luisfaceira changes; I have also added a version check for git, given that the `--dissociate` option is only available since 2.3.0-rc0, which is not yet the default on some *nix distributions.

PS: Please notice that this is my first attempt to read/modify composer, so I would be happy to rework this if needed.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -43,7 +43,7 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
         $cachePath = $this->config->get('cache-vcs-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $url).'/';
-        $cacheOptions = file_exists($cachePath) ? $cacheOptions = '--reference '.$cachePath.' ' : '';
+        $cacheOptions = file_exists($cachePath) ? '--reference '.ProcessExecutor::escape($cachePath).' ' : '';
 
         $ref = $package->getSourceReference();
         $flag = Platform::isWindows() ? '/D ' : '';
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -42,10 +42,12 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
     {
         GitUtil::cleanEnv();
         $path = $this->normalizePath($path);
+        $cachePath = $this->config->get('cache-vcs-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $url).'/';
+        $cacheOptions = file_exists($cachePath) ? $cacheOptions = '--reference '.$cachePath.' ' : '';
 
         $ref = $package->getSourceReference();
         $flag = Platform::isWindows() ? '/D ' : '';
-        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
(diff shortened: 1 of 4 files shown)


src/Composer/Util/Git.php
```diff
@@ -271,4 +271,20 @@ private function throwException($message, $url)
...
+     * Retrieves the current git version.
+     *
+     * @return string
+     *   The git version number.
+     */
+    public function getVersion() {
+        if (0 !== $this->process->execute('git --version', $output)) {
+            throw new \RuntimeException(self::sanitizeUrl('Failed retrieve git version, git was not found, check that it is installed and in your PATH env.' . "\n\n" . $this->process->getErrorOutput()));
+        }
+        if (strpos($output, 'git version ') === FALSE) {
```
alcohol: Maybe use a regex instead to capture the version from the string and return it?

alcohol:
I see that the test cases were adjusted to work as if this new feature is not available. I would really like to see a test case that does the opposite (assume the feature is available).

src/Composer/Downloader/GitDownloader.php
```diff
@@ -42,10 +42,23 @@ public function doDownload(PackageInterface $package, $path, $url)
...
+            if (!file_exists($cachePath)) {
+                $this->io->writeError(sprintf('    Cloning to cache at %s', ProcessExecutor::escape($cachePath)));
+                $mirrorCommand = 'git clone --mirror %s %s';
+                $mirrorCommandCallable = function ($url) use ($cachePath, $mirrorCommand) {
+                    return sprintf($mirrorCommand, ProcessExecutor::escape($url), ProcessExecutor::escape($cachePath));
+                };
+                $this->gitUtil->runCommand($mirrorCommandCallable, $url, $path, true);
+            }
+        }
+        $cacheOptions = file_exists($cachePath) ? '--dissociate --reference '.ProcessExecutor::escape($cachePath).' ' : '';
```
Taluu: imo this check should be done inside the `if(version_compare)` check. The ternary operator could even be gone.

src/Composer/Util/Git.php
```diff
@@ -271,4 +271,20 @@ private function throwException($message, $url)
...
+     * Retrieves the current git version.
+     *
+     * @return string
+     *   The git version number.
+     */
+    public function getVersion() {
+        if (0 !== $this->process->execute('git --version', $output)) {
+            throw new \RuntimeException(self::sanitizeUrl('Failed retrieve git version, git was not found, check that it is installed and in your PATH env.' . "\n\n" . $this->process->getErrorOutput()));
+        }
+        if (preg_match('/^git version (.*)/', $output, $matches) !== 1) {
```
alcohol: Not sure if this pattern is the best idea.. but it'll do for now.

src/Composer/Downloader/GitDownloader.php
```diff
@@ -42,10 +42,24 @@ public function doDownload(PackageInterface $package, $path, $url)
...
-
+        $cachePath = $this->config->get('cache-vcs-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $url).'/';
+        $cacheOptions = '';
         $ref = $package->getSourceReference();
         $flag = Platform::isWindows() ? '/D ' : '';
-        $command = 'git clone --no-checkout %s %s && cd '.$flag.'%2$s && git remote add composer %1$s && git fetch composer';
+
+        // --dissociate option is only available since git 2.3.0-rc0
+        if (version_compare($this->gitUtil->getVersion(), '2.3.0-rc0', '>=')) {
+            if (!file_exists($cachePath)) {
```
mvorisek, line 52: this check is not enought, we need to update the clone:
a) if the initial clone attempt failed (like on process timeout)
b) cache is out of date
In general, I think we need to always fetch the upstream to the cache if the directory already exists.

Issue #5379, opened 2016-05-27
Ability to Quiet Suggestions?


donatj:
I have an automated testing process that runs composer install with `--prefer-dist --no-interaction --no-progress`. Even with these set I still end up with about 40 lines of:

```
aws/aws-sdk-php suggests installing aws/aws-php-sns-message-validator (To validate incoming SNS notifications)
aws/aws-sdk-php suggests installing doctrine/cache (To use the DoctrineCacheAdapter)
jakub-onderka/php-parallel-lint suggests installing jakub-onderka/php-console-highlighter (Highlight syntax in code snippet)
phing/phing suggests installing pdepend/pdepend (PHP version of JDepend)
...
```

They're just taking up space in my logs.  It would be nice to be able to silence suggestions with a flag.

staabm:
You could use a custom `SuggestedPackagesReporter`

https://github.com/composer/composer/blob/master/src/Composer/Installer.php#L1623

Seldaek:
`composer install | grep -v 'suggests installing` should fix it.

donatj:
lol, that is the lamest solution I've ever heard.

PR #5392, opened 2016-06-01
Adds --no-suggest to install and update commands

donatj:
This allows suggestions to be disabled by the new flag `--no-suggest` 

I requested this in #5379 - it was declined, but I'm hoping as I did the footwork it might be more palatable. 

There is no reason to show suggested packages in an automated CI environment the same as the progress bar. This goes nicely with `--no-progress`.


henderjon:
👍 yes please.

hkdobrev:
I was going to suggest to include in `composer create-project` as well, but if this is targeted mostly towards automated environments, we could skip that as `create-project` normally would not be automated.

staabm:
if the target is automation environments, should it be the default on travis runs?

donatj:
@staabm I would think so. I assume that doesn't require me to change anything with this though, right?

staabm:
I dont know whether there is already some kind of config which is used only for travis

alcohol:
LGTM; a test would be nice to have though.

PR #5393, opened 2016-06-01
Strip out declare strict_types from plugins

mlocati:
If a plugin file starts with `declare(strict_types=1);`, Composer fails with this error:

```
Fatal error: strict_types declaration must be the very first statement in the script
in src/Composer/Plugin/PluginManager.php(194) : eval()'d code on line 3
```

So, let's strip out the strict types declaration before evaluating the file contents.

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -190,6 +190,7 @@ class PluginManager
                 $code = str_replace('__FILE__', var_export($path, true), $code);
                 $code = str_replace('__DIR__', var_export(dirname($path), true), $code);
                 $code = str_replace('__CLASS__', var_export($class, true), $code);
+                $code = preg_replace('/^(\s*<\?(?:php)?\s+)declare\s*\(\s*strict_types\s*=\s*\d+\s*\)\s*;/', '$1', $code);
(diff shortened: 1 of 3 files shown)


Seldaek:
Actually all we should have to strip is the `<?php` tag, and then we stop adding this `?>` in the eval to close the php context it starts with, then I think it should work?

mlocati:
> Actually all we should have to strip is the <?php tag, and then we stop adding this ?> in the eval to close the php context it starts with, then I think it should work?

Do you mean to replace
```php
$code = preg_replace('/^(\s*<\?(?:php)?\s+)declare\s*\(\s*strict_types\s*=\s*\d+\s*\)\s*;/', '$1', $code);
eval('?>'.$code);
```
with
```php
$code = preg_replace('/^\s*<\?(php)?/i', '', $code, 1);
eval($code);
```
?

Seldaek:
Yup, if tests still pass with that then I think that would be better :)

mlocati:
> Yup, if tests still pass with that then I think that would be better :)

Yes, they're passing on my machine

mlocati:
PR updated

Seldaek:
Thanks

PR #5394, opened 2016-06-01
--no-scripts option for remove and require command

svengerlach:
`--no-scripts` option wasn't available for require and remove command. 

Background: 

I'm working with a lot of small vagrant boxes (small in terms of low available memory). The memory consumption of composer sometimes makes uncomfortable to use composer within the vagrant boxes. Because of this, I tend to use composer outside of the vagrant boxes. This leads to thrown exceptions, if the scripts defined in composer.json try to access resources that are not accessible from outside of the boxes.


Seldaek:
Thanks

PR #5401, opened 2016-06-02
New getters & setters for RootPackage

CupOfTea696:
Adds `RootPackage::getConfig()` and `RootPackage::getScripts()`

### DIFF
--- a/src/Composer/Package/CompletePackage.php
+++ b/src/Composer/Package/CompletePackage.php
@@ -25,25 +25,8 @@ class CompletePackage extends Package implements CompletePackageInterface
     protected $authors;
     protected $description;
     protected $homepage;
-    protected $scripts = array();
     protected $support = array();
     protected $abandoned = false;
-
-    /**
-     * @param array $scripts
-     */
-    public function setScripts(array $scripts)
-    {
-        $this->scripts = $scripts;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function getScripts()
-    {
-        return $this->scripts;
-    }
 
     /**
      * Set the repositories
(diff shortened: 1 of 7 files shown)


src/Composer/Package/RootAliasPackage.php
```diff
@@ -65,6 +65,22 @@ public function getPreferStable()
     /**
      * {@inheritDoc}
      */
+    public function getConfig()
+    {
+        return $this->aliasOf->config;
```
Seldaek: Should be getConfig()

src/Composer/Package/RootPackage.php
```diff
@@ -22,6 +22,8 @@ class RootPackage extends CompletePackage implements RootPackageInterface
     protected $minimumStability = 'stable';
     protected $preferStable = false;
     protected $stabilityFlags = array();
+    protected $config = array();
+    protected $scripts = array();
```
stof: I suggest making new properties private to make maintenance easier regarding BC (this will not be an issue for usage anyway, as they have public getters and setters)

Seldaek: Don't think it matters a whole lot on these as they are kinda value objects and very unlikely to change.

src/Composer/Package/CompletePackage.php
```diff
@@ -25,27 +25,10 @@ class CompletePackage extends Package implements CompletePackageInterface
...
-        $this->scripts = $scripts;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    public function getScripts()
-    {
-        return $this->scripts;
-    }
```
CupOfTea696: Well that broke everything...

Seldaek:
Thanks, I removed the get/setScript stuff that were duplicates of the parent class, but otherwise looks good for the config

CupOfTea696:
Whoops! Must've missed those in the parent.

PR #5435, opened 2016-06-14
Check headers for missing scopes before asking for new OAuth token

PhilLello:
This patch stops the GitHub VCS driver prompting for a new access token when a repository is deleted/hidden.

Specifically, it checks the X-OAuth-Scopes and X-Accepted-OAuth-Scopes response headers to see if the scopes on the current request match those needed by the API call. If they do, the 404 means that the repo is deleted/hidden, and there's no point asking for a new OAuth token.

### DIFF
--- a/src/Composer/Repository/Vcs/GitHubDriver.php
+++ b/src/Composer/Repository/Vcs/GitHubDriver.php
@@ -325,7 +325,27 @@ class GitHubDriver extends VcsDriver
                         return $this->attemptCloneFallback();
                     }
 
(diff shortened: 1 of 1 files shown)


PhilLello:
Final (I hope) version of this patch pushed; local tests pass, will see what Travis thinks.

Seldaek:
Thanks! I just refactored a bit in https://github.com/composer/composer/commit/1c4acf358316689c8c5d215b5f37fdf38e78d0ac to re-use existing code for header parsing.

PR #5445, opened 2016-06-19
Fix typo

gitname:


GeeH:
@Seldaek this is correct, `into` is an incorrect construct there, but `in` is better.

PR #5467, opened 2016-06-24
Fossil VCS support for Composer

bohwaz:
This adds support for local and remote Fossil repositories as Composer packages.

Fossil is a popular DVCS that includes issues management, a wiki and more: https://www.fossil-scm.org/

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -37,7 +37,7 @@ flags are also required, but when using the installer you will be warned about
 any incompatibilities.
 
 To install packages from sources instead of simple zip archives, you will need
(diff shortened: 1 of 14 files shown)


alcohol:
Looks good, I think. Cannot say I ever heard of Fossil though. But then again, we still support SVN too.

staabm:
Misses at least some basic testcoverage

alcohol:
Agreed.

bohwaz:
Will add some tests asap

bohwaz:
Done.

Seldaek:
Thanks, looks like a fairly complete implementation.

PR #5473, opened 2016-06-27
Add new config autoload-static to disable generating autoload.static.php file

paladox:
Some projects that use composer need to support older php before php 5.6 such as php 5.5, but linting php is causing an issue which is what this pull is fixing by introducing a config, on a per project basis they can disable autoload static file if they want to without needing to fork composer or try to manually hack composer.


Seldaek:
I'd rather not add this.. PHP 5.5 is about to be EOL'd, so if you want to support dead software it'd be best to just configure the linter correctly.

PR #5478, opened 2016-06-29
error in string comparison

averagegeek:
The double equal (==) with single quote condition for string comparison gives a unexpected operator error.


alcohol:
Weird, what shell are you using?

```
==
is equal to

if [ "$a" == "$b" ]

This is a synonym for =.
```

```
[ $a == z* ]]   # True if $a starts with an "z" (pattern matching).
[[ $a == "z*" ]] # True if $a is equal to z* (literal matching).

[ $a == z* ]     # File globbing and word splitting take place.
[ "$a" == "z*" ] # True if $a is equal to z* (literal matching).
```

alcohol:
It seems http://mywiki.wooledge.org/BashGuide/TestsAndConditionals#Conditional_Blocks_.28if.2C_test_and_.5B.5B.29 agrees though. They only reference `==` inside double brackets (`[[ ]]`).

alcohol:
Yeah they have been doing that for a while. Dash is commonly used for init scripts and such because it has a slightly lower overhead than bash (or so I recall having read somewhere).

PR #5485, opened 2016-07-02
Fix minor bitbucket oauth issues

stefangr:
If authentication is required and composer downloads a zip from bitbucket, it's redirected to a login page asking for credentials.
With this fix, the code with apply existing (or ask for new oauth) credentials automatically.


Seldaek:
Thanks

PR #5490, opened 2016-07-05
Change the casting type of the argument of "memory limit".

ysdyy:
This change enables to increase "memory_limit" by "float" value.
Composer currently get the "memory_limit" in php.ini as a type of "integer".

### DIFF
--- a/bin/composer
+++ b/bin/composer
@@ -16,7 +16,7 @@ if (function_exists('ini_set')) {
 
     $memoryInBytes = function ($value) {
         $unit = strtolower(substr($value, -1, 1));
-        $value = (int) $value;
+        $value = (float) $value;
         switch($unit) {
             case 'g':
                 $value *= 1024;


staabm:
never seen someone uses a float value as memory limit in the wild. do people use it and why?

alcohol:
This change makes no sense to me. Even the manual states that `memory_limit` is an integer.

> `memory_limit` `integer`
This sets the maximum amount of memory in bytes that a script is allowed to allocate. This helps prevent poorly written scripts for eating up all available memory on a server. Note that to have no memory limit, set this directive to -1.
When an integer is used, the value is measured in bytes. Shorthand notation, as described in this FAQ, may also be used.

Can you explain the rationale behind your change?

ysdyy:
I set a value "memory_limit=1.5G" in php.ini, and confirmed "phpinfo()".
This shows "memory_limit => 1.5G => 1.5G", so I was confused as if this setting is effective.

But, php also cast this value as integer, so the result was as you say.
thanks.

alcohol:
Yeah if you want `1.5G`, I think you have to use the `1500M` shorthand notation instead.

PR #5491, opened 2016-07-05
Support gitlab private-token for easier access via Gitlab API

moolex:
Support gitlab private-token (or personal-access-token) for easier access via Gitlab API

Config example:

$HOME/.composer/auth.json

```json
{
	"gitlab-token": {
		"gitlab.com": "YOUR-TOKEN-HERE"
	}
}
```

Gitlab API authentication doc: http://doc.gitlab.com/ce/api/README.html

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -67,6 +67,12 @@ A list of domain names and oauth keys. For example using `{"gitlab.com":
 "oauthtoken"}` as the value of this option will use `oauthtoken` to access
 private repositories on gitlab.
 
+## gitlab-token
+
+A list of domain names and private tokens. For example using `{"gitlab.com":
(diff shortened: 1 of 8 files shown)


stof:
Too bad that they decided to use a custom header instead of putting this in the Authorization header...

You should also update the ConfigCommand to support configuring this auth setting (there are multiple places to update in the class)

stof:
and you also need to update the json schema

moolex:
@stof I have update docs and related config definitions, please help me to review it again. TKS

moolex:
@stof Is there any other problem in my commits ?

alcohol:
Aside from one PSR violation (but that's not for you to worry about), everything looks good to me.

alcohol:
Could do with a test though, imho.

Seldaek:
Thanks

PR #5495, opened 2016-07-05
Update troubleshooting.md

f1rf1n:
Added another cause and solution for the proc_open(): fork failed error.


alcohol:
This is rather generic though. What is "old" in this context? And are you sure that the version upgrade was really the reason? Where is the actual proof behind this? It is not really in our interest to add unsubstantiated claims into the documentation.

Seldaek:
Yeah I don't think we need to tell people to update php for every potential error they get, as that can be the cause of really any bug.

f1rf1n:
Good points. I was being a bit rash.

PR #5499, opened 2016-07-06
IPv6 Platform Requirement

Fleshgrinder:
Please see #5317 (this is the original proposal again) and #5304.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -16,6 +16,7 @@ use Composer\Package\CompletePackage;
 use Composer\Package\PackageInterface;
 use Composer\Package\Version\VersionParser;
 use Composer\Plugin\PluginInterface;
+use Composer\Util\Silencer;
 
 /**
  * @author Jordi Boggiano <j.boggiano@seld.be>
@@ -88,7 +89,7 @@ class PlatformRepository extends ArrayRepository
 
         // The AF_INET6 constant is only defined if ext-sockets is available but
         // IPv6 support might still be available.
-        if (defined('AF_INET6') || @inet_pton('::') !== false) {
+        if (defined('AF_INET6') || Silencer::call('inet_pton', '::') !== false) {
             $phpIpv6 = new CompletePackage('php-ipv6', $version, $prettyVersion);
             $phpIpv6->setDescription('The PHP interpreter with IPv6 support');
             $this->addPackage($phpIpv6);
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -12,9 +12,8 @@
 
 namespace Composer\Repository;
 
-use Composer\Config;
-use Composer\Package\PackageInterface;
 use Composer\Package\CompletePackage;
+use Composer\Package\PackageInterface;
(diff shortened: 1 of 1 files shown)


dzuelke:
Why the "combined" 64-bit IPv6 package? Someone can require both `php-64bit` and `php-ipv6`, there is no need for a `php-64bit-ipv6`.

I'd personally like to see `php-zts`, `php-nts` and `php-debug` virtual platform packages too, and then the possible permutations would quickly spiral out of control.

Fleshgrinder:
I am completely open to that. My initial thought (see original PR) was to avoid things like:

```
{
    "require": {
        "php-64bit": "^5.6 || ^7.0",
        "php-ipv6": "^5.6 || ^7.0"
    }
}
```

However, people should do it the same way as they do it with extensions:

```
{
    "require": {
        "php-64bit": "^5.6 || ^7.0",
        "php-ipv6": "*"
    }
}
```

In that case everything is fine.

src/Composer/Repository/PlatformRepository.php
```diff
@@ -88,10 +86,11 @@ protected function initialize()
             $this->addPackage($php64);
         }
 
-        // The AF_INET6 constant is only defined if ext-sockets is available but IPv6 support might still be available.
-        if (defined('AF_INET6') || (function_exists('inet_pton') && Silencer::call('inet_pton', '::') !== false)) {
-            $phpIpv6 = new CompletePackage('ext-network-ipv6', $version, $prettyVersion);
-            $phpIpv6->setDescription('PHP IPv6 support');
+        // The AF_INET6 constant is only defined if ext-sockets is available but
+        // IPv6 support might still be available.
+        if (defined('AF_INET6') || @inet_pton('::') !== false) {
```
Seldaek: Please kep the Silencer::call for inet_pton, that's the way we do @-calls :)

Seldaek: And tbh the function_exists check was safer as it was before, I don't see why that line was changed at all. Only the package name/description really needed changing I think

Fleshgrinder: The `function_exists` call is unnecessary because the function **always** exists.

Seldaek: Functions don't necessarily always exist, as they can be disabled via ini, and some people compile with --disable-all which sometimes disables unexpected stuff. I don't know for sure about this one, but anyway we can go like this and see if it causes any issues.

No biggie about the Silencer call, it's not common I guess but the advantage is it suppresses warnings/notices/.. but not errors. `@` is a bit too aggressive as it suppresses everything. Anyway thanks for updating.

stof: @Seldaek shouldn't you revert the removal of the ``function_exists`` call ?

Seldaek: I don't have time right now.. it probably doesn't matter, so we'll see.

Fleshgrinder: I highly doubt that someone disables `inet_pton` and definitely agree that it is safe to merge this without the check.

I understand the Silencer and I think it is a valid approach to the problem. I just wanted to lay out why I removed it and did not use it. All fine, thanks for the feature. :)

PR #5503, opened 2016-07-08
Urlencode Gitlab project names

calin-marian:
Url encode all non alphanumeric characters in project name for GitLabDriver.

If the project name has "." characters in it, which is supported in Gitlab, the Gitlab API will 404 when requesting the branches or tags of the repository. This commit urlencodes all non alphanumeric characters in the project name in requests to the Gitlab API.

Example:
A random repo I found on Gitlab.com: https://gitlab.com/davidnind/davidnind.gitlab.io

https://gitlab.com/api/v3/projects/davidnind%2Fdavidnind.gitlab.io/repository/branches?private_token=Your-Private-Girlab.com-Token
not working screenshot

https://gitlab.com/api/v3/projects/davidnind%2Fdavidnind%2Egitlab%2Eio/repository/branches?private_token= Your-Private-Girlab.com-Token
working screenshot

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -237,7 +237,7 @@ class GitLabDriver extends VcsDriver
      * @param string $string
      * @return string
      */
-    protected function urlEncodeAll($string)
+    private function urlEncodeAll($string)
     {
         $encoded = '';
         for ($i = 0; isset($string[$i]); $i++) {
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -228,7 +228,24 @@ class GitLabDriver extends VcsDriver
      */
     public function getApiUrl()
     {
-        return $this->scheme.'://'.$this->originUrl.'/api/v3/projects/'.$this->owner.'%2F'.$this->repository;
+        return $this->scheme.'://'.$this->originUrl.'/api/v3/projects/'.$this->urlEncodeAll($this->owner).'%2F'.$this->urlEncodeAll($this->repository);
+    }
+
+    /**
+     * Urlencode all non alphanumeric characters.
+     * 
+     * @param string $string
+     * @return string
+     */
+    protected function urlEncodeAll($string)
+    {
+        $encoded = '';
+        for ($i = 0; isset($string[$i]); $i++) {
+            $character = $string[$i];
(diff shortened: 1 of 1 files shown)


alcohol:
Any particular reason you decided not to use http://php.net/urlencode ?

calin-marian:
They don't encode "."

alcohol:
Ah, right. Bummer :(

src/Composer/Repository/Vcs/GitLabDriver.php
```diff
@@ -228,7 +228,24 @@ protected function fetchComposerFile($identifier)
...
+        return $this->scheme.'://'.$this->originUrl.'/api/v3/projects/'.$this->urlEncodeAll($this->owner).'%2F'.$this->urlEncodeAll($this->repository);
+    }
+
+    /**
+     * Urlencode all non alphanumeric characters.
+     * 
+     * @param string $string
+     * @return string
+     */
+    protected function urlEncodeAll($string)
```
stof: the method should be private IMO

src/Composer/Repository/Vcs/GitLabDriver.php
```diff
@@ -228,7 +228,24 @@ protected function fetchComposerFile($identifier)
      */
     public function getApiUrl()
     {
-        return $this->scheme.'://'.$this->originUrl.'/api/v3/projects/'.$this->owner.'%2F'.$this->repository;
+        return $this->scheme.'://'.$this->originUrl.'/api/v3/projects/'.$this->urlEncodeAll($this->owner).'%2F'.$this->urlEncodeAll($this->repository);
+    }
+
+    /**
+     * Urlencode all non alphanumeric characters.
```
staabm: I guess we should note in the docs that we dont use urlencode/rawurlencode because the `.` needs encoding

helhum:
> If the project name has "." characters in it, which is supported in Gitlab, the Gitlab API will 404 when requesting the branches or tags of the repository

It might make sense to create a workaround here, but shouldn't this be fixed on Gitlabs side (as well)?

As far as I understand the [RFC](https://www.ietf.org/rfc/rfc1738.txt), "." is explicitly allowed in a URL

alcohol:
> As far as I understand the RFC, "." is explicitly allowed in a URL

It is not about whether or not the character is illegal. Otherwise character encoding would not be a thing. It is about how it is interpreted. The urlencoding is necessary in this case.

Seldaek:
Thanks! Merged in 1.2 branch

Issue #5505, opened 2016-07-10
dev-master does not use branch-alias when compared against conflict rules (bug?)


core23:
With the following composer.json:

``` json
{
    "name": "acme/test-bundle",
    "type": "symfony-bundle",
    "require": {
        "php": "^5.3 || ^7.0",
        "sonata-project/classification-bundle": "^4.0@dev",
        "sonata-project/media-bundle": "^4.0@dev"
    }
}
```

When I run this command:

```
composer update
```

I get this output:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for sonata-project/classification-bundle ^4.0@dev -> satisfiable by sonata-project/classification-bundle[4.x-dev].
    - don't install sonata-project/media-bundle 4.x-dev|install sonata-project/media-bundle dev-master
    - sonata-project/media-bundle dev-master conflicts with sonata-project/classification-bundle[4.x-dev].
    - Installation request for sonata-project/media-bundle ^4.0@dev -> satisfiable by sonata-project/media-bundle[4.x-dev].
```

This should resolve in installing both libraries in the latest `master` branch versions, because both have a branch-alias `"dev-master": "4.x-dev"`. Both library have a cyclic dependency ([media](https://github.com/sonata-project/SonataMediaBundle/blob/master/composer.json) and [classification](https://github.com/sonata-project/SonataClassificationBundle/blob/master/composer.json)).

The classification bundle has also a `conflict` setting, which prevents installing a version greater or lower 4.x. The upper rule seams to be the problem, because when I removed the upper conflict rule (`>= 5.0`) the dependencies could be resolved. Refs https://github.com/sonata-project/SonataClassificationBundle/pull/224.

alcohol:
Cyclic dependencies are bad.

alcohol:
It is not clear to me what is triggering this conflict rule though. Theoretically it should not. If I manually create the version and constraint, they do not satisfy each other. So, it looks like there might be a bug somewhere else in the code.

alcohol:
I tried reproducing it with this fixture

```
--TEST--

See Github issue #5505 ( github.com/composer/composer/issues/5505 ).

--COMPOSER--
{
    "repositories": [
        {
            "type": "package",
            "package": [
                {
                    "name": "a",
                    "version": "4.x-dev",
                    "require-dev": { "b": "^3.0 || ^4.0@dev" },
                    "conflict": { "b": "<3.0 || >=5.0" }
                }, {
[...]

PR #5507, opened 2016-07-11
add a fixture for github issue #5505

alcohol:
not sure how to solve yet, or if it requires solving at all


Seldaek:
#5505 was fixed in 2.0 so closing this

Issue #3114, opened 2014-07-10
autoload-dev does not work


divsmith:
autoload-dev does not seem to be working. When my composer.json reads

``` js
"autoload": {
    "psr-4": {
        "Namespace\\": "src/"
    }
},
"autoload-dev": {
    "prs-4": {
        "Namespace\\Tests\\": "tests/"
    }
}
```

and I run

``` sh
composer dump-autoload
```

My /vendor/composer/autoload_prs4.php file appears as 

``` js
// autoload_psr4.php @generated by Composer

$vendorDir = dirname(dirname(__FILE__));
$baseDir = dirname($vendorDir);

return array(
    'Namespace\\' => array($baseDir . '/src'),
);
```

and autoloading the Namespace\Tests namespace does not work. But when I change my composer.json to

``` js
"autoload": {
    "psr-4": {
        "Namespace\\": "src/",
        "Namespace\\Tests\\": "tests/"
    }
}
```

and run

``` sh
composer dump-autoload
```

my autoload_psr4.php file reads

``` js
// autoload_psr4.php @generated by Composer

$vendorDir = dirname(dirname(__FILE__));
$baseDir = dirname($vendorDir);

return array(
    'Namespace\\Tests\\' => array($baseDir . '/tests'),
    'Namespace\\' => array($baseDir . '/src'),
);
```

and the testing namespace is autoloaded correctly. I am using composer version e77435cd0c984e2031d915a6b42648e7b284dd5c.

easy-system:
Same issue.
The file `autoload_psr4.php` not contains the namespace from `autoload-dev ` section. 
This is the latest version of `composer.phar`.

Josantonius:
Same issue.

`composer require vendor/namespace`

The file autoload_psr4.php not contains the namespace from autoload-dev section only from autoload.

`    
"autoload": {
        "psr-4": {
            "Namespace\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Namespace\\Tests\\": "tests/"
        }
    }
`
`
// autoload_psr4.php @generated by Composer

$vendorDir = dirname(dirname(__FILE__));
$baseDir = dirname($vendorDir);

return array(
    'Namespace\\' => array($baseDir . '/src'),
);
`

xabbuh:
@Josantonius Only the `autoload-dev` section from your project's `composer.json` file is taken into account: https://getcomposer.org/doc/04-schema.md#autoload-dev

[...]

PR #5537, opened 2016-07-21
Pass devMode from RunScriptCommand to Script\Event

Great-Antique:
The problem is described in #3114. `devMode` is not passed to `Script\Event`, so `AutoloadGenerator` doesn't parse `autoload-dev` section from `composer.json`

### DIFF
--- a/tests/Composer/Test/Command/RunScriptCommandTest.php
+++ b/tests/Composer/Test/Command/RunScriptCommandTest.php
@@ -1,0 +1,109 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Test\Command;
+
+use Composer\Composer;
+use Composer\Config;
+use Composer\Script\Event as ScriptEvent;
+use Composer\TestCase;
+
+class RunScriptCommandTest extends TestCase
+{
+
+    /**
+     * @dataProvider getDevOptions
+     * @param bool $dev
+     * @param bool $noDev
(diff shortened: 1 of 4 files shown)


alcohol:
Would love to see a test for this, but other than that, LGTM :-)

Great-Antique:
Add test (I'm not familiar with PHPUnit yet, so please check it)

Now with `composer.json`
```
{
    "autoload": {
        "psr-4": {
            "nodev\\": "nodev"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "dev\\": "dev"
        }
    },
    "scripts": {
        "nodev": "nodev\\nodev::test",
        "dev": "dev\\dev::test"
    }
}
```
you can run
```
$ composer run dev
> dev\dev::test
string(3) "dev"
```
But
```
$ composer run --no-dev dev
Class dev\dev is not autoloadable, can not call dev script
```

I know that the second one is weird (You will not run dev scripts in production, etc), but now you are able to move autoload rules of dev packages that have scripts to `autoload-dev` section.

tests/Composer/Test/Command/RunScriptCommandTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+        $ed = $this->getMockBuilder('Composer\EventDispatcher\EventDispatcher')
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $ed->expects($this->once())
+            ->method('hasEventListeners')
+            ->with($this->callback(function (ScriptEvent $event) use ($scriptName, $expectedDevMode) {
+                return $event->getName() === $scriptName
+                && $event->isDevMode() === $expectedDevMode;
```
slbmeh: This looks kind of risky to me... The test is relying on the result of the state of what you are testing in comparison to a test state.

Seldaek:
Thanks, this has been merged to 1.2 and master! You can use it now via `composer self-update --snapshot`.

PR #5549, opened 2016-07-25
Update Xdebug instructions in troubleshooting.md

morganestes:
In the code to add a `composer` function to disable Xdebug, add trailing slash to the paths we're copying ini files in and out of. Without that, bash thinks you want to rename the file itself, not just move it into another directory.

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -224,7 +224,7 @@ To check where the PHP configuration is, and then use a similar script:
 
 ```sh
 mkdir /usr/local/etc/php/7.0/conf.dis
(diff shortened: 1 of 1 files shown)


alcohol:
That is not entirely true. It will only rename the file if the target directory does not exist. If the target directory exists, it will simply move the file into it.

Appending a slash will not help to create the target directory if it does not exist, but it will throw an error, which makes sense in this situation.

So while the rationale behind the patch is flawed, the patch itself is valid.

On the other hand, we do state that a user should use a script "similar to" the example provided.

¯\\\_(ツ)\_/¯

slbmeh:
Agreed with @alcohol, this doesn't solve a problem, but raises an error for a problem.  IMO it should go somewhere with `mktemp` to try to avoid all potential of collision and moved back to the origin afterwards...

morganestes:
Hmm, so my fix isn't really the right fix, but uncovered another issue. Fantastic 😬!

> we do state that a user should use a script "similar to" the example provided.

I'm guessing that even though there's a disclaimer, I'm not the only one who does a copy/paste/see what's wrong when code samples are provided. :)

Seldaek:
Closing this as #5580 makes this whole documentation useless (hopefully).

PR #5552, opened 2016-07-26
Added minor-only option to show command to only show packages with minor updates

fvdb:
This adds a new option to the show command: --minor-only.

When used in combination with --outdated, it will only show packages that have minor semver-compatible updates.

The use case for this is that we want to use composer to check if our project has any direct dependencies with minor semver-compatible updates, and fail our CI build if that's the case. The alternative would've been either trying to parse the original output checking for minor versions, or write a separate tool that hooks into composer to do the check.

This seemed like the simplest option.

Looking forward to your feedback.

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -71,7 +71,7 @@ class ShowCommand extends BaseCommand
                 new InputOption('tree', 't', InputOption::VALUE_NONE, 'List the dependencies as a tree'),
                 new InputOption('latest', 'l', InputOption::VALUE_NONE, 'Show the latest version'),
                 new InputOption('outdated', 'o', InputOption::VALUE_NONE, 'Show the latest version but only for packages that are outdated'),
-                new InputOption('minor-only', 'm', InputOption::VALUE_NONE, 'Show the latest version but only for packages that have minor SemVer-compatible updates'),
+                new InputOption('minor-only', 'm', InputOption::VALUE_NONE, 'Show only packages that have minor SemVer-compatible updates. Use with the --outdated option.'),
                 new InputOption('direct', 'D', InputOption::VALUE_NONE, 'Shows only packages that are directly required by the root package'),
             ))
             ->setHelp(<<<EOT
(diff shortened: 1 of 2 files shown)


alcohol:
LGTM

src/Composer/Command/ShowCommand.php
```diff
@@ -314,6 +315,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
                         if ($showLatest && isset($latestPackages[$package->getPrettyName()])) {
                             $latestPackackage = $latestPackages[$package->getPrettyName()];
                         }
+
+                        if ($input->getOption('outdated') && $input->getOption('minor-only') && $latestPackackage && (!$this->isImmediateSemverCompliantUpgradeNeeded($package, $latestPackackage) || $latestPackackage->isAbandoned())) {
```
stof: This implementation is broken: even if the latest package is a new major version, it does not mean that there is no new versions of the packages for the major version you use.

You will need to change the way to fetch packages instead

alcohol: Ah yes, well spotted.

Assuming `1.6.1` is installed, and the latest is `1.7.2`, but the `1.6` branch also has a `1.6.2` tag, then `~1.6` and `1.7.2` would yield a no-upgrade-required state.

fvdb: Good point, thanks. Will look into this.

fvdb:
@alcohol @stof Instead of simply omitting major versions, I've changed how package versions are fetched when the minor-only option is enabled to ensure we get the latest semver-compatible versions regardless of whether there are major updates.

mvanbaak:
LGTM

Seldaek:
Thanks!

mbrodala:
Just from reading this change: doesn't this also include major changes, making the name `--minor-only` misleading?

If a package `foo` is installed in version `1.2.3`, internally the constraint `^1.2.3` is used which includes minor and major updates. Thus the flag should be renamed to `--without-breaking` or similar IMO.

PR #5558, opened 2016-07-28
Add ClassLoader::$missingClasses to not trigger a COW

nicolas-grekas:
Caching missing classes in `$this->classMap` breaks the benefit of #5174 by triggering a copy-on-write.
Let's track them in a dedicated array.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -53,8 +53,8 @@ class ClassLoader
 
     private $useIncludePath = false;
     private $classMap = array();
-
     private $classMapAuthoritative = false;
+    private $missingClasses = array();
 
     public function getPrefixes()
     {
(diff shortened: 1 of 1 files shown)


Seldaek:
Looks good to me but I'll let others chime in just in case as it touches critical code :)

alcohol:
How exactly does this avoid a COW?

src/Composer/Autoload/ClassLoader.php
```diff
@@ -322,20 +322,20 @@ public function findFile($class)
...
-        if ($this->classMapAuthoritative) {
+        if ($this->classMapAuthoritative || isset($this->missingClasses[$class])) {
             return false;
         }
 
         $file = $this->findFileWithExtension($class, '.php');
 
         // Search for Hack files if we are running on HHVM
-        if ($file === null && defined('HHVM_VERSION')) {
+        if (false === $file && defined('HHVM_VERSION')) {
```
curry684: This is a functional change unrelated to the PR?

src/Composer/Autoload/ClassLoader.php
```diff
@@ -322,20 +322,20 @@ public function findFile($class)
...
             $file = $this->findFileWithExtension($class, '.hh');
         }
 
-        if ($file === null) {
+        if (false === $file) {
             // Remember that this class does not exist.
-            return $this->classMap[$class] = false;
```
nicolas-grekas: Here was the copy-on-write: changing the `$this->classMap`, when this array comes from shared memory on PHP 7.

stof:
:+1:

src/Composer/Autoload/ClassLoader.php
alcohol: And `$this->missingClasses` does not?

alcohol: Ah wait, cause classmap gets injected (and thus only a reference, until modified)?

nicolas-grekas: exactly

stof: @alcohol this array is not coming from shared memory, as it is initialized when initializing the class (well, the initial empty array may be in shared memory, but copying the empty array is not a perf issue).

The COW on the classmap is breaking the optimization done in 1.1 where we precisely avoided copying the classmap when initializing the ClassLoader (because the classmap in it has thousands of entries in it for big projects)

Seldaek:
Alrighty, thanks

Issue #76, opened 2011-10-30
Keep "pretty" representation of versions


igorw:
Versions are converted to internal representation like 9999999-dev. May be a good idea to keep the original format too, in case we need to display it to the user.

PR #5569, opened 2016-08-02
Cleanup processes output handling

nicolas-grekas:
I was playing with a blackfire profile done on a composer install and figured out that `fread` was called 90k times. After investigating, it turns out that `unzip`'s output is ignored by composer, but it's not by the underlying Symfony Process object.

Let's pipe it's output to `/dev/null` (not on windows of course).

Meanwhile, some code needed a related cleanup, here it is :)

### DIFF
--- a/src/Composer/Downloader/ZipDownloader.php
+++ b/src/Composer/Downloader/ZipDownloader.php
@@ -70,7 +70,7 @@ class ZipDownloader extends ArchiveDownloader
         $processError = null;
 
         if (self::$hasSystemUnzip && !(class_exists('ZipArchive') && Platform::isWindows())) {
-            $command = 'unzip '.ProcessExecutor::escape($file).' -d '.ProcessExecutor::escape($path);
+            $command = 'unzip -qq '.ProcessExecutor::escape($file).' -d '.ProcessExecutor::escape($path);
             if (!Platform::isWindows()) {
                 $command .= ' && chmod -R u+w ' . ProcessExecutor::escape($path);
             }
--- a/src/Composer/Downloader/RarDownloader.php
+++ b/src/Composer/Downloader/RarDownloader.php
@@ -44,7 +44,7 @@ class RarDownloader extends ArchiveDownloader
 
         // Try to use unrar on *nix
         if (!Platform::isWindows()) {
-            $command = 'unrar x ' . ProcessExecutor::escape($file) . ' ' . ProcessExecutor::escape($path) . ' && chmod -R u+w ' . ProcessExecutor::escape($path);
+            $command = 'unrar x ' . ProcessExecutor::escape($file) . ' ' . ProcessExecutor::escape($path) . ' >/dev/null && chmod -R u+w ' . ProcessExecutor::escape($path);
 
             if (0 === $this->process->execute($command, $ignoredOutput)) {
                 return;


src/Composer/Downloader/ZipDownloader.php
```diff
@@ -72,11 +72,11 @@ protected function extract($file, $path)
         if (self::$hasSystemUnzip && !(class_exists('ZipArchive') && Platform::isWindows())) {
             $command = 'unzip '.ProcessExecutor::escape($file).' -d '.ProcessExecutor::escape($path);
             if (!Platform::isWindows()) {
-                $command .= ' && chmod -R u+w ' . ProcessExecutor::escape($path);
+                $command .= ' >/dev/null && chmod -R u+w ' . ProcessExecutor::escape($path);
```
nicolas-grekas: This is the main point of this patch.

src/Composer/Repository/Vcs/FossilDriver.php
```diff
@@ -63,7 +63,7 @@ public function initialize()
 
                 $fs->ensureDirectoryExists($this->checkoutDir);
 
-                if (0 !== $this->process->execute(sprintf('fossil clone %s %s', ProcessExecutor::escape($this->url), ProcessExecutor::escape($this->repoFile)), $output)) {
+                if (0 !== $this->process->execute(sprintf('fossil clone %s %s', ProcessExecutor::escape($this->url), ProcessExecutor::escape($this->repoFile)))) {
```
alcohol: What about line #76?

src/Composer/Downloader/ZipDownloader.php
alcohol: What about just passing `-qq` to `unzip`?

alcohol:
LGTM

src/Composer/Downloader/ZipDownloader.php
nicolas-grekas: is it portable?

alcohol: I would expect it to be. But I do not have any Windows machines in close proximity :-(
Let me go check some colleagues around here..

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -70,13 +70,13 @@ protected function extract($file, $path)
         $processError = null;
 
         if (self::$hasSystemUnzip && !(class_exists('ZipArchive') && Platform::isWindows())) {
-            $command = 'unzip '.ProcessExecutor::escape($file).' -d '.ProcessExecutor::escape($path);
+            $command = 'unzip -qq '.ProcessExecutor::escape($file).' -d '.ProcessExecutor::escape($path);
```
nicolas-grekas: `-qq` to discard outputs

src/Composer/Downloader/ZipDownloader.php
nicolas-grekas: I just looked on my Windows VM, looks OK

romainneutron:
What about using `Process::disableOutput`?

nicolas-grekas:
`Process::disableOutput` disables both stdout and stderr, but in this case we want to disable only stdout...

nicolas-grekas:
ping @Seldaek , this is an easy pick :)

Seldaek:
Am I missing something or is this enabling output on all these process calls? Because by default ProcessExecutor will `echo` all output unless you pass it a second $output argument in which case it captures it.

nicolas-grekas:
Nope, it won't echo. `proc_open` does that, but not ProcessExecutor... It
just ignores the output that had been buffered by Process.

Seldaek:
As per https://github.com/composer/composer/blob/master/src/Composer/Util/ProcessExecutor.php#L67 and https://github.com/composer/composer/blob/master/src/Composer/Util/ProcessExecutor.php#L102, no..

```php
$e = new Composer\Util\ProcessExecutor();
$e->execute('echo FOO');

$e = new Composer\Util\ProcessExecutor();
$e->execute('echo BAR', $output);
```

```
$ php test.php
FOO
```

This is basically potentially causing a huge mess of output as far as I can tell. I don't know if you tested it at all.. but anyway we can add the -qq on the zip and rar calls, that sounds fine, it should limit the amount of writes they do and therefore the amount of fread needed. The rest I'd rather not do it unless there is a very good reason, because it won't change the amount of fread the Process component does.

nicolas-grekas:
I did, not enough :) Fixed!

Seldaek:
Thanks :) Looks good now, and here's a good reminder to everyone that there is no such thing as an "easy pick" when it comes to reviewing :p

PR #5575, opened 2016-08-04
update help page on global for COMPOSER_HOME info

admonkey:
default installation directory for global packages
has changed since adding support for freedesktop.org standards
per confusion from http://stackoverflow.com/q/38703915/4233593

Now if your system uses freedesktop.org standards, then it will first check
`XDG_CONFIG_HOME` or default to `/home/<user>/.config/composer`

### DIFF
--- a/src/Composer/Command/GlobalCommand.php
+++ b/src/Composer/Command/GlobalCommand.php
@@ -42,6 +42,9 @@ is to add the COMPOSER_HOME/vendor/bin dir to your PATH env var.
 COMPOSER_HOME is c:\Users\<user>\AppData\Roaming\Composer on Windows
 and /home/<user>/.composer on unix systems.
 
+If your system uses freedesktop.org standards, then it will first check
+XDG_CONFIG_HOME or default to /home/<user>/.config/composer
+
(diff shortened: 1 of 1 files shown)


alcohol:
Only on fresh/clean installs that don't already have `~/.composer` though. It is going to be a bit awkward to phrase/write out all the different scenarios however.

admonkey:
@alcohol I think my one sentence added sums it up pretty good and covers all conditions. If it seems awkward, then at least the existing sentence should be updated to show the current default path since it's now outdated/incorrect in lieu of the freedesktop standard.

ccharles:
I can't believe it's preferable to have misleading documentation than to deal with the "awkwardness" of documenting how the tool actually works. If it's worth building that complexity into the tool it's worth documenting properly.

alcohol:
I was not hinting at any preferences or passing any judgement on this PR. I was merely commenting on a scenario that was not yet included in the current revised text, and that it is going to be a pain to document them all without making it rather confusing.

Seldaek:
I think it's fine as is really, at least it shows there might be two places to look into.

PR #5580, opened 2016-08-08
Restart with xdebug disabled

johnstevenson:
I was asked about implementing a solution regarding xdebug performance ([Windows-Setup: issue#58](https://github.com/composer/windows-setup/issues/58)) and wondered whether this functionality would be more useful inside Composer - if it can be made to work reliably. This PR is an initial stab at that.

To make testing easier, here is the compiled [composer.phar](https://github.com/johnstevenson/composer/releases/tag/v1.2-dev.0256f62)  (updated to 0256f62)

I have assumed that the preferred use-case is running Composer without `xdebug`, so this is the default behaviour. It can be controlled by setting an environment variable beforehand:

* `COMPOSER_ALLOW_XDEBUG=0` Composer will restart without xdebug if it has been enabled. This is the default behaviour and does not require any variable to be set.
* `COMPOSER_ALLOW_XDEBUG=1` Composer will continue and not care about xdebug.

### Downsides
So far:
* The APC warnings will be displayed twice if Composer is restarted.
* Now fixed ~~The ansi-colors disappear running linux distros via Vagrant on Windows. It can be overcome by adding an `--ansi` option to the restart arguments (not implemented as I have no idea yet why this happens).~~ I was thrown by Symfony `Console` always assuming that `STDOUT` is a terminal on current Win10 (even when it is a pipe or file) and overlooked the fact that colored output needed to be handled here. 
* ...

### DIFF
--- a/src/Composer/XdebugHandler.php
+++ b/src/Composer/XdebugHandler.php
@@ -248,9 +248,8 @@ class XdebugHandler
             $result = ($stat['mode'] & 0170000) === 0020000;
         }
 
-        if ($result && defined('PHP_WINDOWS_VERSION_BUILD')) {
-            $result = 0 >= version_compare('10.0.10586', PHP_WINDOWS_VERSION_MAJOR.'.'.PHP_WINDOWS_VERSION_MINOR.'.'.PHP_WINDOWS_VERSION_BUILD)
-            || false !== getenv('ANSICON')
+        if (defined('PHP_WINDOWS_VERSION_BUILD') && $result) {
+            $result = false !== getenv('ANSICON')
             || 'ON' === getenv('ConEmuANSI')
             || 'xterm' === getenv('TERM');
         }
--- a/src/Composer/XdebugHandler.php
+++ b/src/Composer/XdebugHandler.php
@@ -12,6 +12,8 @@
 
 namespace Composer;
 
+use Symfony\Component\Console\Output\OutputInterface;
+
 /**
  * @author John Stevenson <john-stevenson@blueyonder.co.uk>
  */
@@ -19,6 +21,7 @@ class XdebugHandler
 {
     const ENV_ALLOW = 'COMPOSER_ALLOW_XDEBUG';
 
+    private $output;
     private $loaded;
     private $tmpIni;
     private $scanDir;
@@ -26,8 +29,9 @@ class XdebugHandler
     /**
      * Constructor
      */
-    public function __construct()
+    public function __construct(OutputInterface $output)
     {
+        $this->output = $output;
         $this->loaded = extension_loaded('xdebug');
         $tmp = sys_get_temp_dir();
         $this->tmpIni = $tmp.'/composer-php.ini';
@@ -214,6 +218,9 @@ class XdebugHandler
     /**
      * Returns the restart script arguments, adding --ansi if required
      *
+     * If we are a terminal with color support we must ensure that the --ansi
+     * option is set, because the restarted output is piped.
+     *
      * @param array $args The argv array
      *
      * @return array
@@ -224,36 +231,12 @@ class XdebugHandler
             return $args;
         }
 
-        if ($this->isColorTerminal()) {
+        if ($this->output->isDecorated()) {
             $offset = count($args) > 1 ? 2: 1;
             array_splice($args, $offset, 0, '--ansi');
         }
 
         return $args;
-    }
-
-    /**
-     * Returns whether we are a terminal and have colour capabilities
-     *
-     * @return bool
-     */
-    private function isColorTerminal()
-    {
-        if (function_exists('posix_isatty')) {
-            $result = posix_isatty(STDOUT);
-        } else {
-            // See if STDOUT is a character device (S_IFCHR)
-            $stat = fstat(STDOUT);
-            $result = ($stat['mode'] & 0170000) === 0020000;
-        }
-
-        if (defined('PHP_WINDOWS_VERSION_BUILD') && $result) {
-            $result = false !== getenv('ANSICON')
-            || 'ON' === getenv('ConEmuANSI')
-            || 'xterm' === getenv('TERM');
-        }
-
-        return $result;
     }
 
     /**
--- a/src/Composer/XdebugHandler.php
+++ b/src/Composer/XdebugHandler.php
@@ -19,19 +19,16 @@ class XdebugHandler
 {
     const ENV_ALLOW = 'COMPOSER_ALLOW_XDEBUG';
 
-    private $argv;
     private $loaded;
     private $tmpIni;
     private $scanDir;
 
     /**
-     * @param array $argv The global argv passed to script
+     * Constructor
      */
-    public function __construct(array $argv)
+    public function __construct()
     {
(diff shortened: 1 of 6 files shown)


src/Composer/XdebugHandler.php
```diff
@@ -0,0 +1,321 @@
...
+
+    /**
+     * Constructor
+     */
+    public function __construct()
+    {
+        $this->loaded = extension_loaded('xdebug');
+        $tmp = sys_get_temp_dir();
+        $this->tmpIni = $tmp.DIRECTORY_SEPARATOR.'composer-php.ini';
+        $this->scanDir = $tmp.DIRECTORY_SEPARATOR.'composer-php-empty';
```
Seldaek: `'/'` is fine really instead of DIRECTORY_SEPARATOR

src/Composer/XdebugHandler.php
```diff
@@ -0,0 +1,321 @@
...
+     *
+     * @return bool
+     */
+    private function needsRestart()
+    {
+        if (PHP_SAPI !== 'cli' || !defined('PHP_BINARY')) {
+            return false;
+        }
+
+        return !getenv(self::ENV_ALLOW) && $this->loaded;
```
Seldaek: Maybe check for tmp_dir writability as well here, because if it's not writable I think it's best to skip the restart

src/Composer/XdebugHandler.php
```diff
@@ -0,0 +1,321 @@
...
+    private function getCommand($additional)
+    {
+        if ($additional) {
+            if (!file_exists($this->scanDir) && !@mkdir($this->scanDir, 0777)) {
+                return;
+            }
+            putenv('PHP_INI_SCAN_DIR='.$this->scanDir);
+        }
+
+        $phpArgs = array(PHP_BINARY, '-c', $this->tmpIni);
```
Seldaek: PHP_BINARY is only available on 5.5+ or something, anyway not available on all php versions we support, so I'd say also if it is missing just skip this restarting hack and we revert back to outputing a warning, I don't think this restart has to work for 100% of users.

src/Composer/XdebugHandler.php
```diff
@@ -0,0 +1,321 @@
...
+        if ($quote) {
+            $arg = preg_replace('/(\\\\*)$/', '$1$1', $arg);
+            $arg = '"'.$arg.'"';
+        }
+
+        if ($meta) {
+            $arg = preg_replace('/(["^&|<>()%])/', '^$1', $arg);
+        }
+
+        return $arg;
```
Seldaek: not entirely sure why this is done this way here, is it being of passthru being used? Or why can't you rely on ProcessExecutor::escape (which has a bunch of workarounds for escapeshellarg on windows already)?

src/Composer/XdebugHandler.php
```diff
@@ -0,0 +1,321 @@
...
+            $result = ($stat['mode'] & 0170000) === 0020000;
+        }
+
+        if (defined('PHP_WINDOWS_VERSION_BUILD') && $result) {
+            $result = false !== getenv('ANSICON')
+            || 'ON' === getenv('ConEmuANSI')
+            || 'xterm' === getenv('TERM');
+        }
+
+        return $result;
```
Seldaek: I don't really like the whole checking for color capability happens here, and essentially duplicates all of the symfony/console code. I think it might be better to extract https://github.com/composer/composer/blob/master/src/Composer/Console/Application.php#L99-L101 into a method so that we can create the output in bin/composer, check if it is decorated or not, and pass that along to the Xdebug handler, then inject $output in ->run().

Seldaek:
Aside from the above comments, :+1: for creativity :)

src/Composer/XdebugHandler.php
johnstevenson: Of course. I'll fix this.

src/Composer/XdebugHandler.php
johnstevenson: Tmp dir writability is covered in `prepareRestart()`, which handles any unexpected failure conditions at a later stage. `needsRestart()` simply deals with concrete conditions.

Seldaek: Right, I thought I saw that after.. anyway thanks for confirming, it's 
fine by me then

src/Composer/XdebugHandler.php
johnstevenson: PHP_BINARY was introduced in 5.4 (so it will be available for most). We already check for it in the `needsRestart()` method, so for 5.3 users running xdebug we would bail there (and they would get the standard warning)

src/Composer/XdebugHandler.php
johnstevenson: Yes, we need to escape the arguments because we are passing them to a new php process. We use `passthru` because:

- this function gives us all the control we need (it hooks up the standard handles and acts as a proxy between parent and child processes).
- it enables interactive composer commands to function correctly.

I've used it rather than ProcessExecutor::escape because it provides a more robust mechanism than the Symfony component, in that it follows Windows command-line parsing rules. The Symfony code will probably work fine in most cases, but I wanted to make this aspect as bullet-proof as possible, so as not to break stuff (by trying to be helpful!).

I've put the result in [https://github.com/johnstevenson/winbox-args](https://github.com/johnstevenson/winbox-args) because it turns out to be non-trivial and needs a fair bit of explanation. The wiki there describes the gory details.

src/Composer/XdebugHandler.php
johnstevenson: Hey thanks, that is a much better idea. I'll file a Symfony/console issue about Windows output always being seen as a terminal when color is supported.

Seldaek:
Thanks, let's see how it goes on the snapshot channel for a bit :)

bendavies:
hi @Seldaek @johnstevenson,
Something squiffy here, i get this:

```
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Generating autoload files
> Incenteev\ParameterHandler\ScriptHandler::buildParameters
Updating the "app/config/parameters.yml" file
> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::buildBootstrap
Cannot load Zend OPcache - it was already loaded
PHP Warning:  Module 'apcu' already loaded in Unknown on line 0
PHP Warning:  Module 'apc' already loaded in Unknown on line 0
PHP Warning:  Module 'blackfire' already loaded in Unknown on line 0
PHP Warning:  Module 'gmp' already loaded in Unknown on line 0
PHP Warning:  Module 'intl' already loaded in Unknown on line 0
PHP Warning:  Module 'pdo_pgsql' already loaded in Unknown on line 0
PHP Warning:  Module 'redis' already loaded in Unknown on line 0
PHP Warning:  Module 'tidy' already loaded in Unknown on line 0
etc...
```

This is on a symfony project.
Thanks

johnstevenson:
Something squiffy, indeed, but it is a Symfony issue that has been reported and needs to be fixed:
https://github.com/sensiolabs/SensioDistributionBundle/issues/299

bendavies:
ah, thanks

PR #5582, opened 2016-08-10
Modify the message for memory errors

### LINKED ISSUE #5565: composer shows error message "… not having swap configured" while swap is configured (but too low)
With a typical Symfony project (the packages are irrelevant here), When I run this command:

```
php ~/composer.phar update --no-progress --no-dev -vvv --profile
```

I get this output:

```
$ php ~/composer.phar update --no-progress --no-dev -vvv --profile
Reading ./composer.json
[…]
[884.6MB/35.87s]   - Removing symfony/symfony (v2.8.8)
[209.0MB/37.36s] The following exception is caused by a lack of memory and not having swap configured
[209.0MB/37.36s] Check https://getcomposer.org/doc/articles/troubleshooting.md#proc-open-fork-failed-errors for details
```

The message `not having swap configured` is confusing, because my swap is configured, as shown by `free`.

# Proposal

I think that the message should not pretend that swap is not configured.

It could be less precise:

> The following exception is caused by a lack of memory or swap, or not having swap configured

diegoholiveira:
Closes #5565

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -296,7 +296,7 @@ class Application extends BaseApplication
         }
 
         if (false !== strpos($exception->getMessage(), 'fork failed - Cannot allocate memory')) {
-            $io->writeError('<error>The following exception is caused by a lack of memory and not having swap configured</error>', true, IOInterface::QUIET);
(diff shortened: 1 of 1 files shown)


alcohol:
Thanks

PR #5591, opened 2016-08-12
Use https url of the repository instead of ssh given gitlab project is public

### LINKED ISSUE #4744: GitLab driver uses ssh instead of https when using --prefer-source
When the GitLab driver is used to install a module using `--prefer-source`, ssh is used instead https:
```
$ composer config repositories.foo-bar vcs https://gitlab.com/foo/bar
$ composer require foo/bar --prefer-source
```

When ssh is not set up, the last command gives the following result:

```
Using version ~1.0 for foo/bar
  - Installing foo/bar (v1.2.3)
    Cloning 9876543210fedcba9876543210fedcba
    Failed to download foo/bar from source: Failed to execute git clone --no-checkout 'git@gitlab.com:foo/bar.git' '/path/to/project/vendor/foo/bar' && cd '/path/to/project/vendor/foo/bar' && git remote add composer 'git@gitlab.com:foo/bar.git' && git fetch composer

Cloning into '/path/to/project/vendor/foo/bar'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

    Now trying to download from dist
```

I can understand that ssh used as a fallback, but not when the repository url is specified with https and a source download is explicitly requested.

ianlet:
With this change the `GitLabDriver` follows the same behavior as the `GitHubDriver` when the repository url is retrieved: if the project is public it returns the HTTPS url, otherwise it returns the SSH url.

Fixes #4744

### DIFF
--- a/tests/Composer/Test/Repository/Vcs/GitLabDriverTest.php
+++ b/tests/Composer/Test/Repository/Vcs/GitLabDriverTest.php
@@ -69,6 +69,7 @@ class GitLabDriverTest extends TestCase
 {
     "id": 17,
     "default_branch": "mymaster",
+    "public": false,
     "http_url_to_repo": "https://gitlab.com/mygroup/myproject.git",
     "ssh_url_to_repo": "git@gitlab.com:mygroup/myproject.git",
     "last_activity_at": "2014-12-01T09:17:51.000+01:00",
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Repository/Vcs/GitLabDriverTest.php
```diff
@@ -126,6 +166,20 @@ public function testGetSource()
         $this->assertEquals($expected, $driver->getSource($reference));
     }
 
+    public function testGetSource_GivenPublicProject()
```
alcohol: Underscore?

alcohol:
LGTM

tests/Composer/Test/Repository/Vcs/GitLabDriverTest.php
ianlet: For better readability I was used to separate the "when", "given", "then" by an underscore in a test name. But I can remove it if it doesn't follow the coding style of this project (I couldn't find an example of "given" in the tests).

PR #5598, opened 2016-08-15
Fix support for testing exit code

jadb:
The `$sectionData` was being overwritten (type mismatch), causing its assertion (`$this->assertSame()`) to fail.


alcohol:
Thanks

Issue #5600, opened 2016-08-15
require-dev package repository options do not get copied into composer.lock


john-whitley:
Firstly, thanks for all your hard work on the composer tool - it is a really valuable part of our environment.

**Summary**

I believe I have uncovered an edge case issue - `"require-dev"` packages do not get the required `"transport-options"` section in the `composer.lock` file, necessary when hosting an internal composer repository with specific ssl requirements.  I propose a choice of fixes for this, but humbly request review of the proposals from the composer maintainers.

**Background**

At work, we have a large (10<n<100) number of git repositories that have distinct responsibilities, allowing for correct reuse without having to have everything everywhere (composer is good at this! - thank you).  Without going into the detail of all of them, we have some we intend for production, and some we intend for development environments only (examples include our in-house coding standards via code-sniffer; internal test helpers; etc.).

To help us manage our dependencies, we host all our internal libraries on an internal satis server.  We use https with the satis server, and have configured this with an self-signed ssl certificate - we don't use letsencrypt as the satis server is protected by firewalls, so lets encrypt cannot verify the service; neither do we want to pay for a signed certificate.

**Problem discovery**

We have successfully configured the `composer.json` with the necessary ssl options in the `"options"` section of the `"repository"` section, and these transfer into the `composer.lock` file in the `"transport-options"` section associated with the correct packages; which is the correct and expected behaviour.

When we attempt the same with the `"require-dev"` packages, this transfer into the `composer.lock` `"transport-options"` section does not happen.

We have edited `composer.lock` to include the `"transport-options"` section, and this proof-of-concept implementation fixes our https error for `"require-dev"` packages, so we believe that it is acceptable to have the `"transport-options"` section within a `"require-dev"` `composer.lock` entry.

I investigated the composer code and believe that this is caused by the call to use a new `\Composer\Package\Loader\ArrayLoader` instance without setting the `loadOptions` argument to true: the call is from `src/Composer/Installer.php` line 631, which relies on a new `\Composer\Package\Loader\ArrayLoader` instance on line 628.  This new instance does not set the `loadOptions` setting in the constructor.

**Proposed solutions**
I have implemented and tested two solutions:

1. Set the loadConfig flag in the instance creation in `\Composer\Package\Loader\ArrayLoader` on line 628.
2. Alter the default value to loadConfig to true in `\Composer\Package\Loader\ArrayLoader` on line 32

I have implemented both choices to prove them before writing this: both solutions pass the (possibly naive) tests:
[...]

PR #5603, opened 2016-08-16
require-dev package repository options get copied into composer.lock (fix for #5600)

john-whitley:
commit 3994b556dcffcde7b1801c8bc712f3127e8f8e7c
Author: John Whitley 
Date:   Tue Aug 16 09:02:53 2016 +0100

https://github.com/composer/composer/issues/5600

This alters the default flag for loadOptions in \Composer\Package\Loader\ArrayLoader to true; and alters the assumption of the test to reflect this change.

**Rationale**

The `\Composer\Package\Loader\ArrayLoader` test (defined in tests/Composer/Test/Package/Loader/ArrayLoaderTest.php) assumed that a new `\Composer\Package\Loader\ArrayLoader` instance would be always created with the optional flag loadOptions set to true.

```php
$this->loader = new \Composer\Package\Loader\ArrayLoader(null, true);
```

This change alters the general case to reflect the default assumption as defined in the test.

commit b75fc4ad7238bc50f724bd29446ccbc33e82c34c
Author: John Whitley 
Date:   Mon Aug 15 16:55:27 2016 +0100

Altered the test for ArrayLoader to use the default loadConfig flag, and to test the true and false states for the loadConfig flag

### DIFF
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -29,7 +29,7 @@ class ArrayLoader implements LoaderInterface
     protected $versionParser;
     protected $loadOptions;
 
-    public function __construct(SemverVersionParser $parser = null, $loadOptions = false)
+    public function __construct(SemverVersionParser $parser = null, $loadOptions = true)
     {
         if (!$parser) {
             $parser = new VersionParser;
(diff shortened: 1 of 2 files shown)


alcohol:
LGTM;

@Seldaek any idea why this was specifically not implemented? Just a minor oversight? Cause it seems too obvious almost to not be intentional, but I could be mistaken of course.

Seldaek:
Thanks, I merged the PR to get the extra tests, but I took the other approach you mentioned (set the flag to true in Installer) instead of the changing the default value as that might introduce unexpected problems. The one in the Installer was definitely an oversight, it should have loaded the options if present.

john-whitley:
@Seldaek: Thank you for merging this in; for reviewing the original bug, and the two suggestions; and for altering this work to implement the alternative solution.

@alcohol: Thank you for the original reviews of this PR and the original bug I submitted.

I hope you both have great weekends.

PR #5604, opened 2016-08-16
Adds --no-suggestion flag to the docummentation

jcchavezs:
At the moment, the ``--no-suggest`` flag (implemented in https://github.com/composer/composer/pull/5392) is not being documented anywhere and in my opinion this is such an awesome feature that should be know everyone as soon as possible.

Ping @aeony @donatj @Seldaek


alcohol:
Thanks

PR #5608, opened 2016-08-18
Adds --no-suggest to RequireCommand

donatj:
`--no-suggest` was added on UpdateCommand and InstallCommand.  

This adds it to RequireCommand which can be seen quite visibly by running the following (which our CI system runs):

```bash
composer require phing/phing
```

vs

```bash
composer require phing/phing --no-suggest
```


alcohol:
Please update the documentation as well :-)

donatj:
@alcohol Done ;)

Seldaek:
Thanks

PR #5612, opened 2016-08-20
Add support for seamless execution of local binaries

davidolrik:
Projects that add binaries to `vendor-bin` can now execute those via the same command as projects that consume them without installing them first.

In list overview local commands have a `(local)` tag to distinguish them from commands installed in `vendor-bin`.

Local binaries take precedence over `vendor-bin` which takes precedence over binaries in path.


Seldaek:
Thanks - had to do a couple fixes for windows but otherwise it looks good :)

davidolrik:
Awesome!

PR #5613, opened 2016-08-21
Running sudo can brick your computer

pwolanin:
Trying to help a new Drupal contributor.  She had composer.phar in her $PATH, but was trying to follow instructions to put composer in /usr/local/bin.

I think teaching people to follow instructions to type "sudo" off the internet is a bad idea - better to have them put their own $HOME/bin dir imho.

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -102,6 +102,10 @@ mv composer.phar /usr/local/bin/composer
 > directory" then you must create the directory manually before proceeding:
 > `mkdir -p /usr/local/bin`.
 
+> **NOTE:** Running commands with sudo can brick your computer. So, a better
+> alternative could be to create the $HOME/bin directory, and mv composer to
+> that location and add $HOME/bin to your path.
+
 > **Note:** For information on changing your PATH, please read the
 > [Wikipedia article](https://en.wikipedia.org/wiki/PATH_(variable)) and/or use Google.
 


curry684:
The people you are trying to warn have no idea what "bricking" is.

alcohol:
If we go down this path, we can add warnings and notes everywhere throughout our documentation, and it would become a big clutter of nonsense. We provide and document a dependency manager. We are not the internet police.

pwolanin:
@alcohol ok, but this is the 1st page of the docs anyone will go to and you are guiding people in a direction that is both possibly harmful and not really necessary

alcohol:
That's a matter of personal opinion.

curry684:
@pwolanin Composer is a programmer's tool, not a toddler game. `sudo` exists because privileged actions are sometimes needed, especially in the field of software development. Putting overt warnings about that everywhere is just going to lull people in a false sense of security and cause more damage than good.

YesCT:
What is the disadvantage of giving instructions for people install in their home directory?

bojanz:
No need for warnings. A sane(r) default would do. Even "real programmers" benefit from those.

mlhess:
@curry684  Composer is a programmer, but there are newbie programers that are learning and thus could be called "toddler's"  no?

alcohol:
`/usr/local/bin` is a perfectly sane default. Depending on your environment, you won't even need `sudo`. The docs merely mention using `sudo` when required.

PR #5634, opened 2016-08-29
appveyor.yml: use one type of indentation

TomasVotruba:
2 spaces as in `travis.yml` and in the beginning of this file:

```
platform:
  - x86
  - x64
```

To make this united.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,30 +6,30 @@ platform:
 clone_folder: c:\projects\composer
 
 cache:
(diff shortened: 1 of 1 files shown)


alcohol:
Could you also add the following line to `.gitattributes` while you are at it?

```
*.yml   text eol=lf whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=2
```

PR #5636, opened 2016-08-29
[Doc] Enable curl extension without php.ini

sanpii:


alcohol:
I think you added it to the wrong section of the documentation (this is about xdebug, not cURL).

sanpii:
> I think you added it to the wrong section of the documentation

When you execute the command, you have an error:

```
$ php -n composer update

                                                                         
  [ErrorException]                                                       
  Use of undefined constant CURLOPT_HTTPGET - assumed 'CURLOPT_HTTPGET'  
```

I simply complete the alias to avoid this error.

alcohol:
Incorrect. Composer makes no use of `CURLOPT_HTTPGET`. Must be a plugin you installed.

Issue #5631, opened 2016-08-28
Composer init not accepting my name (similar to issue #3948)


emeraldinspirations:
Attempting to run composer.phar init produces the following error:

```
Author [Matthew "Juniper" Barlett <emeraldinspirations@gmail.com>, n to skip]: 
 Invalid author string.  Must be in the format: John Smith <john@example.com> 
Author [Matthew "Juniper" Barlett <emeraldinspirations@gmail.com>, n to skip]: Matthew (Juniper) Barlett <emeraldinspirations@gmail.com>
 Invalid author string.  Must be in the format: John Smith <john@example.com> 
Author [Matthew "Juniper" Barlett <emeraldinspirations@gmail.com>, n to skip]: 
```
[screenshot from 2016-08-28 04-13-03](https://cloud.githubusercontent.com/assets/20853049/18032768/3c16f514-6cd6-11e6-99ad-f63e79cab4a5.png)

curry684:
The regexp is at https://github.com/composer/composer/blob/master/src/Composer/Command/InitCommand.php#L323 if you want an easy PR :wink:

emeraldinspirations:
I am sorry, @curry684, I am new to gitHub.  What do you mean by PR?

curry684:
https://www.thinkful.com/learn/github-pull-request-tutorial/ :smile:

emeraldinspirations:
Status: Running into a circular dependency issue.
- To run the unit tests, composer is required
- Must fix the bug in order to install composer into the repo
- Can't verify the bug is fixed without the unit tests

Also, while I can simply adjust the filter, I have not yet finished tracing the code in order to verify that the quote character won't cause additional issues down the line.

PR #5638, opened 2016-08-31
Fix for issue #5631

emeraldinspirations:
- Add " ( and ) as valid characters in author name
- Add relevant unit tests
#5631

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -320,7 +320,7 @@ EOT
      */
     public function parseAuthorString($author)
     {
-        if (preg_match('/^(?P<name>[- \.,\p{L}\p{N}\'’\"\(\)]+) <(?P<email>.+?)>$/u', $author, $match)) {
+        if (preg_match('/^(?P<name>[- .,\p{L}\p{N}\'’"()]+) <(?P<email>.+?)>$/u', $author, $match)) {
             if ($this->isValidEmail($match['email'])) {
                 return array(
                     'name'  => trim($match['name']),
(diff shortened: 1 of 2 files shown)


alcohol:
LGTM

src/Composer/Command/InitCommand.php
```diff
@@ -320,7 +320,7 @@ function ($value) use ($self, $minimumStability) {
      */
     public function parseAuthorString($author)
     {
-        if (preg_match('/^(?P<name>[- \.,\p{L}\p{N}\'’]+) <(?P<email>.+?)>$/u', $author, $match)) {
+        if (preg_match('/^(?P<name>[- \.,\p{L}\p{N}\'’\"\(\)]+) <(?P<email>.+?)>$/u', $author, $match)) {
```
avant1: You are not required to escape double quote and parentheses in "character class" (characters sequence inside square brackets). The same is actual for existing dot.
The results probably will be the same, but pattern will look less complicated.

alcohol: That is correct. It could be simplified to:

```
^(?P<name>[- .,\p{L}\p{N}'’"()]+) <(?P<email>.+?)>$
```

emeraldinspirations:
@alcohol sorry, new to github.  What does LGTM mean?  P.S. is there some sort of abbreviation key somewhere?

alcohol:
Looks good to me. :-)

curry684:
@emeraldinspirations http://acronymfinder.com/Looks-Good-To-Me-(LGTM).html hehe

Seldaek:
Thanks, merged in 1.2 so it won't show merged here, but it's in the composer snapshots now and will be in the next 1.2.x release.

PR #5639, opened 2016-08-31
update composer/semver and other deps

alcohol:


Seldaek:
I already updated semver though :) Is the rest urgent?

alcohol:
Welp, I missed that. No the rest is not urgent.

Seldaek:
OK then let's close, I'll update anyway before releasing 1.2.x, rather avoid unnecessary merge conflicts..

PR #5645, opened 2016-09-03
Replace old deprecated packages with new one

yaman-jain:
1. replace package CommandEvent with \Composer\Script\Event.
2. replace package PackageEvent with \Composer\Installer\PackageEvent.

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -20,9 +20,8 @@ use Composer\IO\IOInterface;
 use Composer\Composer;
 use Composer\DependencyResolver\Operation\OperationInterface;
 use Composer\Repository\CompositeRepository;
-use Composer\Script;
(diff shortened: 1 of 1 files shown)


Seldaek:
This code is specifically for BC with outdated handlers, I guess you did this because  your IDE told you it's using deprecated code, but IDEs aren't very smart..

yaman-jain:
@Seldaek I think this line should be removed. By the way what does BC means? 
https://github.com/composer/composer/blob/master/src/Composer/Script/PackageEvent.php#L20

hkdobrev:
[BC as in Backwards compatibility](https://en.m.wikipedia.org/wiki/Backward_compatibility).

yaman-jain:
Thanks !

Seldaek:
Well no the line shouldn't be removed, it indicates to other people that they shouldn't use the class. But we still need to use it to provide backwards compatibility.

Issue #5407, opened 2016-06-05
Syntax errors in `autoload_static.php` (PHP < 5.6)


mgburns:
I'm using Composer to manage dependencies for a WordPress project for a client hosted on WP Engine.

After upgrading to Composer 1.1.0 our deployments started failing due to PHP syntax errors. We're deploying code using their [Git Push](https://wpengine.com/git/) method, which has a `post-receive` hook that runs all changed files through `php -l` to check for syntax errors. We have a custom deploy script that runs `composer install` locally (with `config.platform.php` set to `5.3.29` as WP Engine is very behind) and then commits and pushes the build to their remotes.

The file that is causing the failure is `autoload_static.php` (introduced in [1.1.0-RC](https://github.com/composer/composer/releases/tag/1.1.0-RC)). It looks like that file assumes PHP 5.6 (regardless of which version of PHP you're running `composer install` with) as it uses scalar expressions in all of the static property initializers:

```php
// autoload_static.php @generated by Composer

namespace Composer\Autoload;

class ComposerStaticInit55ee1fe9ded343dc8de3a805d154a4ba
{
    public static $files = array (
        'cb78221880aa21d756cc44a8539bb425' => __DIR__ . '/..' . '/timber/timber/init.php',
    );
```

I rolled Composer back to 1.0.3 as a workaround, but feel that Composer shouldn't be generating files with syntax errors for the platform it's installing on.

Taluu:
The autoload_static file is loaded only if the optimize option has been activated and if it is php >= 5.6. You shouldn't run `php -l` on vendors file (And thus autoload files generated by composer), as it is not used...

mgburns:
The `php -l` checks are out of my control (they're run by WP Engine), but I'm not sure why you say they shouldn't be run against vendor files -- I'd want to know if they have syntax errors just like any other file included in my application.

I do concede that a static syntax check may catch false-positives due to runtime conditionals but regardless I still believe a better approach here is to not generate a file that can't be loaded for the target platform. I'm open to suggestions for things that I could do on my end to more cleanly work around this, but sticking with an older version of Composer does not seem like a good solution.

staabm:
Duplicate of https://github.com/composer/composer/issues/5316 https://github.com/composer/composer/issues/5255

jb510:
[...]

Issue #5324, opened 2016-05-12
Composer v1.1.0 generated composer/autoload_static.php file syntactily invalid for PHP <5.6


bd808:
The [composer/autoload_static.php file](https://gerrit.wikimedia.org/r/#/c/288430/1/composer/autoload_static.php,unified) generated for the [REL1_25 release branch of mediawiki/vendor](https://github.com/wikimedia/mediawiki-vendor/tree/REL1_25) using Composer version 1.1.0 2016-05-10 15:21:19 fails a `php -l` lint check:

```
$ composer --version
Composer version 1.1.0 2016-05-10 15:21:19
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Nothing to install or update
Generating optimized autoload files
$ php -v
PHP 5.5.31 (cli) (built: Jan 11 2016 23:27:33)
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
$ php -l composer/autoload_static.php

Parse error: parse error, expecting `')'' in composer/autoload_static.php on line 23
Errors parsing composer/autoload_static.php
```

Using PHP 5.3 gives a slightly more useful lint failure message:
```
$ php -v
PHP 5.3.29 (cli) (built: Sep 15 2014 12:04:44)
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.3.0, Copyright (c) 1998-2014 Zend Technologies
    with Xdebug v2.2.5, Copyright (c) 2002-2014, by Derick Rethans
$ php -l composer/autoload_static.php
PHP Parse error:  syntax error, unexpected '.', expecting ')' in composer/autoload_static.php on line 23
Errors parsing composer/autoload_static.php
```

The underlying problem is that PHP versions before 5.6 cannot use expressions when defining static properties:
> Like any other PHP static variable, static properties may only be initialized using a literal or constant before PHP 5.6; expressions are not allowed.
-- https://secure.php.net/manual/en/language.oop5.static.php#language.oop5.static.properties

bd808:
I see that this use of this file is guarded in composer/autoload_real.php with a check for `PHP_VERSION_ID >= 50600`. The problem then is just one of the MediaWiki lint checks. Sorry for the noise.

Chealer:
For Eclipse, this can be worked around using "Use as Library Folder" on the vendor/ directory.

Issue #5316, opened 2016-05-11
Syntax error in autoload_static.php


ZachWatkins:
With the following composer.json:

```json
{
    "name": "agrilife/agriflex",
    "description": "Functionality for AgriLife AgriFlex theme",
    "license": "GPL2+",
    "authors": [
        {
            "name": "Zach Watkins",
            "email": "watkinza@gmail.com"
        }
    ],
    "minimum-stability": "stable",
    "require": {},
    "require-dev": {},
    "autoload": {
        "psr-4": {
            "AgriLife\\AgriFlex\\": "src"
        }
    }
}
```

When I run this command:

```
composer install
git push copy remote
```

I get this output:

```
remote: PHP Parse error: syntax error, unexpected '.', expecting ')' in - on line 19
```

And I expected this to happen:
No PHP errors

I am inexperienced with composer, and searching for a way to force composer.json to use a previous version of composer since this PHP error was not occurring last week.

Seldaek:
The file is fine, it is made for PHP5.6+ and only included on those versions of PHP. If you have something linting all files, you will have to white-list that file I'm afraid.

ZachWatkins:
The version of PHP running on our continuous integration service is not the same version of PHP running on our web host servers. The files are checked by our web host server.

Is there another solution? Because telling our web host not to lint files is not as easily done as providing a configuration file.

Seldaek:
One option would be to delete the file before deploying it so it isn't
there to be linted. It won't be included if it's less than 5.6 so it
doesn't need to be there really.

ZachWatkins:
How does composer check for the PHP version? Our CI service lets us set the PHP version using:
$ phpenv 5.4.45

Using this, we are still receiving the version of autoload for PHP 5.6+

stof:
@ZachWatkins look at the ``autoload_real.php`` file. It decides at runtime whether it should include the ``autoload_static.php`` file or other generated files

ZachWatkins:
By runtime, do you mean during the `$ composer install` operations? How do I define the PHP_VERSION_ID variable in bash in a way that autoload_real.php can access it and decide whether or not to add autoload_static.php to the package files?

Seldaek:
@ZachWatkins this isn't a fool-proof solution, as soon as you get new autoload rules it might end up being unable to load some classes.

[...]

Issue #5255, opened 2016-04-27
The new static loader prevents commits to WordPress SVN


daigo75:
WordPress repository has a pre-commit hook that parses the entire code base and reject a commit when they detect the static concatenation code introduce by the new static autoloader. The fact that such code may not even be loaded in PHP 5.5 and earlier is irrelevant, the parser considers the code invalid and any commit containing it is rejected.
For reference, see [Syntax error while updating SVN repository](https://wordpress.org/support/topic/syntax-error-while-updating-svn-repository).

This means that any plugin that should be published or updated in the WordPress repository will be rejected if it contains the static concatenation code. The only workaround found so far is to use an older version of Composer, but this is sub-optimal, as it doesn't allow to take advantage of the newer features.

It would be useful to have the possibility to exclude the *autoload_static.php* entirely, as such file cannot be committed to the SVN repository. Another solution would be to remove the pre-parsing of the committed code, but that won't be possible, as it's how the WordPress repository is configured.

Seldaek:
As I wrote in the commit, I don't think this is really something we have to solve. First of all because it's an external tool flipping on our code, not our code doing anything bad per se. Secondly because we recommend against committing the vendor dir in the first place, so the use case is not even something we'd rather support. If you do this you are on your own IMO, and fixing it in the commit hook sounds easy enough so it doesn't really warrant a new option.

stof:
> The only workaround found so far is to use an older version of Composer, but this is sub-optimal, as it doesn't allow to take advantage of the newer features

The workaround is to use the current stable version of composer instead of using the dev snapshot. It is not that bad.

daigo75:
@Seldaek actually, committing the vendor folder is necessary, if you want to publish a plugin or a theme. If you have ever used WordPress, you will know that installing plugins is a "click and activate" operation. The software is supposed to work straight away, and it should come with all its dependencies packaged and ready for use. Assuming that anyone who downloads a software based on Composer has to run *composer install* or *composer update* is definitely wrong, as that's not going to happen. Most likely, Composer is not even going to be installed in the target system, and the software users would not even know what it is.

As for the commit hooks, they are not up to me to fix. They are set by the repository Admins, who are set to leave them as they are. No matter how easy it would be to address them, I don't see them going away at any time soon.

daigo75:
@stof yes, that would work for the moment.

stof:
[...]

PR #5660, opened 2016-09-08
Allow optionally disabling the static autoloader

legoktm:
The static autoloader presents a problem for environments where pre-PHP
5.6 syntax compatibility is required. Many projects have automated CI
that runs a linter across all files in the repository to prevent
developers from accidentally introdcing incompatible syntax. And while
the static autoloader properly guards itself on lower PHP versions, it
will still trip up linters. Given the number of bugs filed about this,
it appears to be a common setup: #5255, #5316, #5324, and #5407.

This adds a --no-static option to the dump-autoload command, which will
disable the static autoloader by updating autoload_real.php to not look
for the file, and removes autoload_static.php if it exists.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -51,6 +51,11 @@ class AutoloadGenerator
     /**
      * @var bool
      */
+    private $staticAutoloader = true;
+
+    /**
+     * @var bool
+     */
     private $runScripts = false;
 
     public function __construct(EventDispatcher $eventDispatcher, IOInterface $io = null)
@@ -83,6 +88,15 @@ class AutoloadGenerator
     public function setRunScripts($runScripts = true)
     {
         $this->runScripts = (boolean) $runScripts;
+    }
+
+    /**
+     * Set whether to generate a static autoloader or not
+     *
+     * @param bool $staticAutoloader
+     */
+    public function setStaticAutoloader($staticAutoloader) {
+        $this->staticAutoloader = (boolean) $staticAutoloader;
     }
 
     public function dump(Config $config, InstalledRepositoryInterface $localRepo, PackageInterface $mainPackage, InstallationManager $installationManager, $targetDir, $scanPsr0Packages = false, $suffix = '')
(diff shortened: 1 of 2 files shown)


staabm:
Why does your linter lint external code?

legoktm:
Because regardless of the source of the code, it still has to work on our servers? Doing a generic syntax lint seems to a be a general practice in other projects too, based on the other bugs filed that I linked.

Seldaek:
I am really not so enthusiastic about this. You basically want to add a flag to add support for linting a file you shouldn't lint on PHP versions that are not even receiving security updates anymore. This also adds a bunch of cruft and complexity to a piece of code that's already quite lengthy and convoluted. 

Is it really not possible to exclude that file from the linter config instead? It has quite a specific file name.

legoktm:
Originally we looked into excluding that file from linters (and that was going to be our plan until I saw https://github.com/composer/composer/issues/5651#issuecomment-245191150 which suggested sending a PR), but so far we've found at least 3 different linters in our CI + deployment pipeline that need updating, and some of them are upstream projects (e.g. https://github.com/JakubOnderka/PHP-Parallel-Lint) that don't support excluding specific files (only directories). So we could update the linters, yeah, but after seeing other projects have similar issues (based on the issues filed), I think a solution in composer is warranted.

The code is kind of convoluted, but most of that was to make autoload_real.php look pretty to humans. If you were okay with the indentation of two blocks being wrong, then the code would be cleaner. Or even something like `if (false) {} else { ... }` would be better.

As far as supporting older PHP versions...MediaWiki's current LTS will be supporting 5.5 until 2019 roughly, but that's our choice for our users, and I understand if composer doesn't want to give us first class support.

In conclusion, I see our options as:
1. Fix all the linters
2. Accept this patch
3. Be stuck on composer 1.0.x
4. Make autoload_static.php 5.5 compatible (I kind of assumed someone had researched this and determined it not feasible...is that the case?)
5. ...?

paladox:
+1, and per @legoktm we supported php 5.5 due to the fact we have a lts and will receive updates until, 2019.

But without any configurations we will be stuck on an old composer version until 2019 unless we can approve this patch?

And nope we carnt exclude any files like that so we need something internaly in composer which can and this is what this patch does :)

Seldaek:
How about patching the CI/linter system to allow excluding files? Or just
exclude the whole vendor/composer/ dir if that's an option?

You're telling me changing composer is ok but not changing your other
stuff.. Not sure why composer has to bear the cost here :)

legoktm:
> How about patching the CI/linter system to allow excluding files?

As I stated earlier, that's what we'll have to do if this patch isn't accepted. But I would like to continue this discussion first.

> Or just exclude the whole vendor/composer/ dir if that's an option? 

Excluding vendor/composer doesn't work because we still want to lint vendor/composer/semver.

> You're telling me changing composer is ok but not changing your other stuff.. Not sure why composer has to bear the cost here :)

I submitted a pull request because that was what was suggested in the other ticket. I think the duplicate tickets show that linting everything, including upstream code is a common practice, and should be something that composer supports.

So, what are the actual objections against this? Is it that the code is too convoluted to support? Ideologically against improving support for unsupported PHP versions? Think that linting everything is bad and we should stop? ...

alcohol:
There is no need to lint check your vendor directory. As you said yourself, most projects already do lint checking. On top of that, in fact, one of the linters you linked even recommends explicitly excluding the `vendor` directory.

> https://github.com/JakubOnderka/PHP-Parallel-Lint
>
> Recommended setting for usage with Symfony framework
>
> For run from command line:
```
./bin/parallel-lint --exclude app --exclude vendor .
```

I do not see any justification for moving this into the Composer codebase. Lint checkers are nice, but they are not foolproof (as this case clearly demonstrates). If you choose to build your system on top of this, then dealing with the side effects should be your responsibility imho.

Seldaek:
Another aspect that you would benefit from if you fixed your CI, is that if you disable the static autoloader then essentially you're penalizing users of php 5.6+ which won't get the faster autoloader. Those users that upgrade "early" are usually more performance conscious and as such I don't think that's the right approach.

PR #5665, opened 2016-09-11
Allow setting default transport strategy for path repositories

radmiraal:
This change allows setting COMPOSER_DEFAULT_TRANSPORT_STRATEGY
in the environment to set the default strategy to 'mirror' if you like.

This allows using 'mirror' during deployments while still symlinking
on a development machine. The default is still overwritten by the
options on the repository configuration.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -868,4 +868,10 @@ If set to 1, this env disables the warning about running commands as root/super 
 It also disables automatic clearing of sudo sessions, so you should really only set this
 if you use Composer as super user at all times like in docker containers.
 
+### COMPOSER_MIRROR_PATH_REPOS
+
+If set to 1, this env changes the default path repository strategy to `mirror` instead
+of `symlink`. As it is the default strategy being set it can still be overwritten by
+repository options.
+
 &larr; [Libraries](02-libraries.md)  |  [Schema](04-schema.md) &rarr;
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -59,6 +59,11 @@ class PathDownloader extends FileDownloader implements VcsCapableDownloaderInter
         $currentStrategy = self::STRATEGY_SYMLINK;
         $allowedStrategies = array(self::STRATEGY_SYMLINK, self::STRATEGY_MIRROR);
 
+        $mirrorPathRepos = getenv('COMPOSER_MIRROR_PATH_REPOS');
+        if (true === (bool)$mirrorPathRepos) {
+            $currentStrategy = self::STRATEGY_MIRROR;
+        }
+
         if (true === $transportOptions['symlink']) {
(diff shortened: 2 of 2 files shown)


radmiraal:
Hey, this change is just a quick thought I had while playing around with composer path repositories. I now use the symlinking in my development environment, but would like to use 'mirror' during deployments so I can ignore my src folder (and thus not uploading unneeded code).

Please let me know if it makes any sense in your opinion, if so I'll see what else needs to be done to make this feature merged like tests / docs

Seldaek:
I'd say something like `COMPOSER_MIRROR_PATH_REPOS=1` is a bit more clear, but aside from that yes adding it to https://github.com/composer/composer/blob/master/doc/03-cli.md#environment-variables would be good.

joelwurtz:
Very interested to have this.

We actually use a workaround with the `composer config` command to set the correct mirroring option before each deploy, however having an environnement variable would definitely help.

radmiraal:
@Seldaek I updated the code to use the env var you suggested and added it to the environment docs. Let me know if it still needs tuning :-)

curry684:
Why not do the full implementation so you can also force symlinking on your dev machine?

radmiraal:
You mean adding COMPOSER_SYMLINK_PATH_REPOS? That wouldn't make a difference.

Symlink is already the default, if you'd change that by setting the option on the repository the option would still overwrite the env setting as happens with COMPOSER_MIRROR_PATH_REPOS.

Maybe parsing env variables in the composer options could be an option to have it work for both ways.

curry684:
No the default is "try symlink, if that fails mirror". The option for the path repos is a tri-state: mirror=true says always mirror, mirror=false says never mirror, mirror=null (default) says try whichever works, starting with the one taking least disk space.

If there's going to be an env variable to override it it should provide similar/identical configurability.

radmiraal:
@curry684 ok, didn't have a look at the fallback because of the IOException later on in the PathDownloader. But that still doesn't change that the setting only sets the default. The later exception handling just integrates an additional fallback if the default doesn't work.

So with this PR there's one thing added: a possibility to change the default strategy to 'mirror' while otherwise the default is 'symlink'. The fallback that happens later is still in place. I probably completely miss the point, but how would you see an implementation where the default value (which is either mirror or symlink) would reflect the tri-state? IMHO setting the default strategy only effects lines https://github.com/composer/composer/blob/master/src/Composer/Downloader/PathDownloader.php#L59-L60 Setting the default to symlink would not have any effect then as it is default already.

The env variable does btw not effect 2 of the options you mention: mirror=true and mirror=false would still behave exactly the same (because the option overwrites the env var). So the only case we talk about would be mirror=null right?

Seldaek:
OK looks fine to me so merging, @curry684 if you still had a gripe with it please comment, there's still time before 1.3 goes out :)

Thanks @radmiraal

Issue #5667, opened 2016-09-12
Composer overrides PHP configuration and then fails because of missing extensions


renan:
My `composer.json`:

```json
{
    "scripts": {
        "test-json": "php -r \"var_dump(function_exists('json_encode'));\""
    }
}
```

Having PHP with `xdebug` and `json` extensions installed, when running this script I get:

```shell
$ composer run test-json
> php -r "var_dump(function_exists('json_encode'));"
bool(false)
```

When Xdebug is enabled composer tries to restart itself and run without any extra .ini files being loaded, which causes `json` extension to not be present. So when you have scripts that depends on `json` extension the command fails.
This behavior is caused by [XdebugHandler](https://github.com/composer/composer/blob/master/src/Composer/XdebugHandler.php) and reproducible on latest master. Probably introduced on https://github.com/composer/composer/pull/5580

For example using `post-update-cmd` with `Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache`, which depends on `Monolog`, fails with:

```shell
[RuntimeException]
PHP's json extension is required to use Monolog's NormalizerFormatter
```

`XdebugHandler` re-runs the same command with the `-c` option (`-c <path>|<file> Look for php.ini file in this directory`) with the directory `/tmp/composer-php-empty/`, which is indeed empty.

Seldaek:
@renan could you show us what `php -i | grep 'Configuration|\.ini'` outputs? It shouldn't disable everything in theory it removes xdebug but reuses all other ini files that were loaded, so it seems it's not picking up whichever of your ini files loads the json extension.

cc @johnstevenson

renan:
I can see that `/tmp/composer-php.ini` is correctly created and does contain `extension=json`, but this is **never loaded**.

With `php -i` being run through `composer` script I get:
```
Configuration File (php.ini) Path => /etc/php5/cli
Loaded Configuration File => /etc/php5/cli/php.ini
Scan this dir for additional .ini files => /tmp/composer-php-empty
Additional .ini files parsed => (none)
```

`php -i` on the server:
```
Configuration File (php.ini) Path => /etc/php5/cli
Loaded Configuration File => /etc/php5/cli/php.ini
Scan this dir for additional .ini files => /etc/php5/cli/conf.d
Additional .ini files parsed => /etc/php5/cli/conf.d/05-opcache.ini,
/etc/php5/cli/conf.d/10-pdo.ini,
[...]

PR #5668, opened 2016-09-12
tentative fix for #5667

alcohol:


src/Composer/XdebugHandler.php
```diff
@@ -204,6 +212,12 @@ private function getCommand($additional)
             if (!file_exists($this->scanDir) && !@mkdir($this->scanDir, 0777)) {
                 return;
             }
+
+            $iniScanDirEnv = getenv('PHP_INI_SCAN_DIR');
+            if ($iniScanDirEnv) {
+                putenv('PHP_INI_SCAN_DIR_OLD='.$iniScanDirEnv);
```
Seldaek: How about a else block wiping PHP_INI_SCAN_DIR_OLD? And maybe also naming it something less clashy like COMPOSER_OLD_INI_SCAN_DIR?

alcohol: Doh, my bad.

Seldaek:
Thanks, see commit linked above for what I meant earlier, hopefully that's clear now :)

Issue #5596, opened 2016-08-14
Composer doesn't work with local filesystem subversion repos


glennschmidt:
When an svn repository is referenced with a `file://` URL, or a local pathname, Composer cannot download packages from the repository.
eg.
```json
    "repositories": [
        {
            "type": "svn",
            "url": "file:///srv/repos/mine",
```

This happens because Composer erroneously passes file:// URLs through `realpath()`, which will return `false` because it expects a path, not a URL. For example, in `Composer\Downloader\VcsDownloader::download()`
```php
if (Filesystem::isLocalPath($url)) {
    $url = realpath($url);
}
```

`Filesystem:isLocalPath()` returns `true` for file:// URLs, but `realpath` only accepts pathnames.

Also, even if you refer to the local repository in `composer.json` using a pathname (`/srv/repos/mine` instead of `file:///srv/repos/mine`), `SvnDriver::normalizeUrl()` turns it into a file:// URL, and you still get this problem.

alcohol:
Did you try using the VCS driver?

Seldaek:
As a workaround you can probably just use the path `/srv/repos/mine` without file://. Otherwise send a pull request to fix it so that file:// is stripped before doing a realpath.

Seldaek:
Fixed by #5688

PR #5688, opened 2016-09-18
cut out "file://" from VCS download() method

KryDos:
See issue #5596 
The bug is related to realpath() function that was used with file path starts with "file://". 

"file://" is valid filesystem url but not valid for realpath. So my change just checks if url starts with "file://" and cuts it off.


alcohol:
There is no need to use `file://` for local repositories. Just use `/path/to/repo`.

Seldaek:
Thanks, merged in 1.2 and master!

PR #5692, opened 2016-09-19
Allow to get/set repositories.0 in config command

joelwurtz:
Actually the following command:

```
composer config repositories.0
```

will return the same output as

```
composer config repositories
```

As PHP treat 0 as an empty value.

This modification will normally treat 0 as a valid value (i did not test it, change this from a "not super dev friendly environment")


Seldaek:
Thanks for the PR, but the fix isn't really working out as preg_match doesn't set the index 1 at all when there is no match (unless there's a `$match[2]` then it is set), unfortunately. So it leads to undefined index. Anyway fixed now :)

PR #5696, opened 2016-09-21
Fix process timeout when set to 0

fabpot:
This is used in silex/silex-skeleton for instance to run the PHP embed server (https://github.com/silexphp/Silex-Skeleton/blob/master/composer.json#L37).

### DIFF
--- a/tests/Composer/Test/ConfigTest.php
+++ b/tests/Composer/Test/ConfigTest.php
@@ -291,4 +291,11 @@ class ConfigTest extends \PHPUnit_Framework_TestCase
         );
         $this->assertTrue($config->get('disable-tls'));
     }
+
+    public function testProcessTimeout()
+    {
+        putenv('COMPOSER_PROCESS_TIMEOUT=0');
+        $config = new Config(true);
+        $this->assertEquals(0, $config->get('process-timeout'));
+    }
 }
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/ConfigTest.php
```diff
@@ -291,4 +291,11 @@ public function testDisableTlsCanBeOverridden()
         );
         $this->assertTrue($config->get('disable-tls'));
     }
+
+    public function testProcessTimeout()
+    {
+        putenv('COMPOSER_PROCESS_TIMEOUT=0');
```
stloyd: You should reset this value to previous one after test, no?

fabpot: It's not used anywhere else

staabm: agree with @stloyd. this might create global side effects in future tests when not cleared properly.

alcohol:
Thanks.

Seldaek:
It's great that the COMPOSER_PROCESS_TIMEOUT env var is fixed, but btw wouldn't it be better if silex in that file you linked had `config => process-timeout => 0` already so you don't need the env var?

fabpot:
@Seldaek That's a good question. I thought about that, but then, it means that the timeout is for all Composer operations, which might not be what we want... Or put another way, if that's good for Silex, why not set it to 0 at the Composer level? Why is there a 300seconds limit by default?

Seldaek:
By default just to stop random stuff from blocking for hours because some people will let it run waiting so long sometimes, and the few that need longer git clones etc can change it by hand. But on the other hand for silex users that'd be quite useful if it doesn't stop after 5min, and in theory they shouldn't do many slow git clones. It's a trade-off for sure though. Ease of use vs potential WTFs.

PR #5698, opened 2016-09-21
Simplify XDebugHandler restart process

johnstevenson:
To make testing easier, here is the compiled [composer.phar](https://github.com/johnstevenson/composer/releases/download/413a516/composer.phar) updated to 413a516
<hr>

There is no need to point `PHP_INI_SCAN_DIR` to an empty directory in order to bypass the compiled-in scan directory, because it can be overridden by providing an empty value.

The documentation is not entirely clear about this, but this was how it was designed. From [https://bugs.php.net/bug.php?id=45114](https://bugs.php.net/bug.php?id=45114):

> Now you can either use both -n and -c in command line options (-n disables every ini loading, -c enables loading the specified file / from specified path) or set PHP_INI_SCAN_DIR environment variable to empty (disables scanning) or to point to some other directory.

which references this [commit](https://github.com/php/php-src/commit/e5e6f553a25f972500b792170e03b5903ee924dd#diff-7cf540af7079d5a8052e0ea85871fab2R599) from August 2008.

This means that we don't have to create the tmp directory. This PR also removes the need for an extra environment variable, because we use `COMPOSER_ALLOW_XDEBUG` internally to flag that we are a restarted process and store the original value of `PHP_INI_SCAN_DIR`.

### DIFF
--- a/src/Composer/XdebugHandler.php
+++ b/src/Composer/XdebugHandler.php
@@ -20,13 +20,11 @@ use Symfony\Component\Console\Output\OutputInterface;
 class XdebugHandler
 {
     const ENV_ALLOW = 'COMPOSER_ALLOW_XDEBUG';
-    const ENV_INI_SCAN_DIR = 'PHP_INI_SCAN_DIR';
-    const ENV_INI_SCAN_DIR_OLD = 'COMPOSER_PHP_INI_SCAN_DIR_OLD';
+    const RESTART_ID = 'internal';
 
     private $output;
     private $loaded;
-    private $tmpIni;
-    private $scanDir;
+    private $envScanDir;
 
     /**
      * Constructor
@@ -35,9 +33,7 @@ class XdebugHandler
     {
         $this->output = $output;
         $this->loaded = extension_loaded('xdebug');
-        $tmp = sys_get_temp_dir();
-        $this->tmpIni = $tmp.'/composer-php.ini';
-        $this->scanDir = $tmp.'/composer-php-empty';
+        $this->envScanDir = getenv('PHP_INI_SCAN_DIR');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Composer/XdebugHandler.php
```diff
@@ -111,38 +119,38 @@ private function prepareRestart(&$command)
...
+    private function writeTmpIni(array $iniFiles, $replace, &$tmpIni)
     {
         if (empty($iniFiles)) {
             // Unlikely, maybe xdebug was loaded through the -d option.
-            $this->tmpIni = '';
+            $tmpIni = '';
             return true;
         }
 
+        $tmpIni = sys_get_temp_dir().'/composer-php.ini';
```
alcohol: Something struck me that I did not think of before; what if we have multiple users on a server making use of Composer in an environment where xdebug is enabled. Each of them would at some point try to create this file. Since we do not do any cleanup and always use the same filename, do you see potential here for permission or ownership conflicts?

src/Composer/XdebugHandler.php
```diff
@@ -35,33 +33,42 @@ public function __construct(OutputInterface $output)
...
-        if ($originalIniScanDir) {
-            putenv(self::ENV_INI_SCAN_DIR_OLD);
-            putenv(self::ENV_INI_SCAN_DIR . '=' . $originalIniScanDir);
-        } else {
-            putenv(self::ENV_INI_SCAN_DIR);
+            if (false !== $this->envScanDir){
+                // $args[1] contains the original value
+                if (isset($args[1])) {
+                    putenv('PHP_INI_SCAN_DIR='.$args[1]);
+                } else{
```
alcohol: Your editor doesn't point out CS violations? :P

Don't worry about addressing this though, we run `php-cs-fixer` from time to time.

src/Composer/XdebugHandler.php
```diff
@@ -111,38 +119,38 @@ private function prepareRestart(&$command)
...
+     * Writes the tmp ini file and sets it name
      *
      * @param array $iniFiles The php.ini locations
      * @param bool $replace Whether we need to modify the files
+     * @param string $tmpIni Set by method
      *
      * @return bool False if the tmp ini could not be created
      */
-    private function writeTmpIni(array $iniFiles, $replace)
+    private function writeTmpIni(array $iniFiles, $replace, &$tmpIni)
```
alcohol: I personally really do not like the amount of "pass by reference" usage in this class. It makes the code hard to follow. If you see an opportunity to refactor this, I would greatly appreciate it.

tests/Composer/Test/XdebugHandlerTest.php
```diff
@@ -16,18 +16,20 @@
 
 /**
  * @author John Stevenson <john-stevenson@blueyonder.co.uk>
+ * @requires PHP 5.4
```
alcohol: Which part of this requires `5.4`?

johnstevenson: We need `PHP_BINARY` which only became available in 5.4

alcohol: Ah, thanks for pointing that out :-)

src/Composer/XdebugHandler.php
johnstevenson: Oops! I normally run phpcs before pushing anything.

tests/Composer/Test/XdebugHandlerTest.php
johnstevenson: I'll add that snippet to the doc block

src/Composer/XdebugHandler.php
johnstevenson: No worries, I'll fix that.

src/Composer/XdebugHandler.php
johnstevenson: Good point. Currently I've left it to the semantics of `file_put_contents`, but this may cause problems, so I'll come up with something more robust.

We cannot do any cleanup because some commands display the ini location for the user to fix stuff (which also allows us to keep the same filename to keep things nice and tidy - or at least, that was the idea).

johnstevenson:
@alcohol Please give 413a516 a whirl when you have time. I've updated the compiled phar at the top of this page. If everything is acceptable, do you want me to rebase into a single commit?

Seldaek:
Looks fine to me, but yeah if you wanna squash into a single commit why not.

johnstevenson:
Ok, will do. But one more thing: when running a composer command from a script, [EventDispatcher](https://github.com/composer/composer/blob/master/src/Composer/EventDispatcher/EventDispatcher.php#L184) calls `realpath` on `$_SERVER['argv'][0]`. Is his necessary and should it be done here as well?

Seldaek:
I'm not entirely sure why this is there unfortunately. Probably doesn't hurt to replicate but not sure if necessary.

johnstevenson:
Rebased, but I haven't implemented realpath, based on the theory that if a script has been invoked from whatever was given on the command line, it can be invoked again using the same value.

Seldaek:
Thanks

Issue #5701, opened 2016-09-22
Classmap generation failed when using short opening tags in class files


tobias-trozowski:
My `composer.json`:
<details>
<summary>CLICK TO EXPAND</summary>
```json
{
  "license": "proprietary",
  "autoload": {
    "classmap": [
      "lib"
    ]
  }
}
```
</details>

Output of `composer diagnose`:
<details>
<summary>CLICK TO EXPAND</summary>
```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: WARNING
The OpenSSL library (0.9.8zc) used by PHP does not support TLSv1.2 or TLSv1.1.
If possible you should upgrade OpenSSL to version 1.0.1 or above.

Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK

```
</details>

When I run this command:
```
composer dump-autoload -vvv
```

I get the following output:
<details>
<summary>CLICK TO EXPAND</summary>
```
Reading ./composer.json
Loading config file /Users/user/.composer/config.json
Loading config file /Users/user/.composer/auth.json
Loading config file ./composer.json
Checked CA file /private/var/folders/0y/bg47wk2s0757t6cpjwqfj7wm3sdwhz/T/composer-cacert-12fdaece071ee9515fa28aabed5ab089876ae257833106e15a583e060eaff6b5.pem: valid
Executing command (/Users/user/composer-test): git branch --no-color --no-abbrev -v
Executing command (/Users/user/composer-test): git describe --exact-match --tags
Executing command (/Users/user/composer-test): git log --pretty="%H" -n1 HEAD
Failed to initialize global composer: Composer could not find the config file: /Users/user/.composer/composer.json
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
[...]

PR #5708, opened 2016-09-23
test cases for #5701

alcohol:


Seldaek:
Huge diff just to add spaces around dots.. makes it harder to review for nothing really :/ Anyway looks fine to me except it fails to build on hhvm because obviously they are not 100% compatible.

alcohol:
You can view diff without white-space changes ;-)

I'll look into the HHVM side of it in a moment.

Seldaek:
I did check the diff without white-space yes, but I also could not **have to** :p

alcohol:
Yeah yeah. I blame PhpStorm :p

PR #5714, opened 2016-09-25
Include static files on runtime autolad generation (fix #3578)

### LINKED ISSUE #3578: Add support for loading autoload.files when executing scripts
Hi,

At the moment, it seems that the `EventDispatcher::getScriptListeners` is only permitting psr-0/4 classes from being loaded when excuting scripts. This is great, but this causes problems if any of your scripts are using libraries which use the `autoload.files` (specified in `composer.json`) feature as those files will not be loaded.

From the little I know about composer, my guess would have been that once you attach the composer autoloader, you could attach the autoloader of the project you're in. But taking a look at the `EventDispatcher` code, it seems that if a `loader` was already assigned, this one would be unregistered and a new one would be created. I have no idea in what context that happens, but it tells me that we want the old loader gone and we want to generate a new one each time `getScriptListeners` is called.

Is it possible to have composer load the `autoload.files` when executing scripts? If not, what are the possible alternative to get a script that relies on these files to work (would that require manually `require`-ing the files)?

lainosantos:
This PR fixes runtime autoload generation.

Without it, composer fail to include static files on run-scripts and execute plugins (composer globally).

Only `EventDispatcher::getScriptListeners` and `PluginManager::registerPackage` calls this method (`AutoloadGenerator::createLoader`)

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -53,6 +53,11 @@ class AutoloadGenerator
      */
     private $runScripts = false;
 
+    /**
+     * @var array
+     */
+    private static $autoloadedPackages = array();
+
     public function __construct(EventDispatcher $eventDispatcher, IOInterface $io = null)
     {
         $this->eventDispatcher = $eventDispatcher;
@@ -139,7 +144,7 @@ EOF;
 
         // Collect information from all packages.
         $packageMap = $this->buildPackageMap($installationManager, $mainPackage, $localRepo->getCanonicalPackages());
-        $autoloads = $this->parseAutoloads($packageMap, $mainPackage);
+        $autoloads = $this->parseAutoloads($packageMap, $mainPackage, false);
 
         // Process the 'psr-0' base directories.
         foreach ($autoloads['psr-0'] as $namespace => $paths) {
@@ -366,9 +371,10 @@ EOF;
      *
      * @param  array            $packageMap  array of array(package, installDir-relative-to-composer.json)
      * @param  PackageInterface $mainPackage root package instance
+     * @param  bool             $requireFilesOneTime Should require autoload files only one time per package
      * @return array            array('psr-0' => array('Ns\\Foo' => array('installDir')))
      */
-    public function parseAutoloads(array $packageMap, PackageInterface $mainPackage)
+    public function parseAutoloads(array $packageMap, PackageInterface $mainPackage, $requireFilesOneTime = true)
     {
         $mainPackageMap = array_shift($packageMap);
         $sortedPackageMap = $this->sortPackageMap($packageMap);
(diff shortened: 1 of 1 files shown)


lainosantos:
#3578

Seldaek:
The problem with this is we are likely to end up with issues of classname or function already being defined, in case you update a package with files autoload that rename a file for example, it'll be included once, then again after the update with the new name, and we get a fatal error. I'm not sure if it's worth the risk really, IMO the constraint that you can't rely on files autoload in scripts isn't that stringent.

lainosantos:
It is a risk, but when executing plugins, composer requires these files only one time, ok? When run scripts (without update or require) it too requires one time. On update or require commands (with run script associated) composer will require two times, and may trigger an error. So i think that we must require only at first time when require or update and run scripts with a subprocess (it will load all new dependences).

Seldaek:
My point is you can not require only one time as if the file name changed you can not possibly know that it is still the same file. There is no way to do this safely.

lainosantos:
The pattern of file path is {prefix}/vendor/{package-vendor}/{package-name}/{file}. The `AutoloadGenerator::createLoader` only requires files for a package at first call.

lainosantos:
`
public function createLoader(array $autoloads)
    {
        static $packagesIncluded = [];
        
        //...

        if (isset($autoloads['files'])) {
            $listIncluded = [];
            foreach ($autoloads['files'] as $file) {
                preg_match("@^.*?/vendor/(\w+/\w+)/.*@", $file, $fileParts);

                if(in_array($fileParts[1], $packagesIncluded)) {
                    continue;
                }

                $listIncluded[] = $fileParts[1];

                require_once $file;
            }

            $packagesIncluded += array_unique($listIncluded);
        }

        return $loader;
    }
`

lainosantos:
```
public function createLoader(array $autoloads)
    {
        static $packagesIncluded = [];
        // ...

        if (isset($autoloads['files'])) {
            $listIncluded = [];
            foreach ($autoloads['files'] as $file) {
                preg_match("@^.*?/vendor/(\w+/\w+)/.*@", $file, $fileParts);

                if(in_array($fileParts[1], $packagesIncluded)) {
                    continue;
                }

                $listIncluded[] = $fileParts[1];

                require_once $file;
            }

            $packagesIncluded += array_unique($listIncluded);
        }

        return $loader;
    }
```

Seldaek:
It's a smart idea but that code isn't reliable though, /vendor/ can be changed, the packages are not necessarily installed in the vendor directory, etc. You can't just do this based on paths, it would have to be done earlier and would introduce quite a bit of complexity unnecessarily IMO. Just use a class with static methods if you need helpers in a script handler and that's it no? It's not that problematic in this limited scope.

lainosantos:
This is only to demostrate :). Do we can get in createLoader the package instalation dir trough the package name?

lainosantos:
I see that i can pass the packges objects to createLoader.

lainosantos:
Please, review the new comits.

Seldaek:
That's what I meant, it adds complexity in several places in that file, to do a half-way fix to an edge case that is not really necessary IMO, and that file is already too complex as it is. I don't think it's worth doing sorry.

lainosantos:
But plugins are limited to require packages with files autoload. If a puglin wants require e.g. guzzle/guzzle, it cant't do.

Seldaek:
Yes, that is a limitation.. but IMO it's one we can live with.

PR #5715, opened 2016-09-25
Set branch alias for 1.3-dev

ro0NL:
This is correct right compared to https://github.com/composer/composer/blob/1.2/composer.json#L62?


alcohol:
Assuming @Seldaek isn't already planning for 2.0, I would say this is correct yes :p

Seldaek:
Yup thanks I forgot about that :)

PR #5717, opened 2016-09-26
prevent (prompt for) bitbucket auth when it redirected

berlinger-rarents:
#5584

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -176,6 +176,27 @@ class RemoteFilesystem
     }
 
     /**
+     * @link https://github.com/composer/composer/issues/5584
+     *
+     * @param string $urlToBitBucketFile URL to a file at bitbucket.org.
+     *
+     * @return bool Whether the given URL is a public BitBucket download which requires no authentication.
+     */
+    public static function urlIsPublicBitBucketDownload($urlToBitBucketFile)
+    {
+        $path = parse_url($urlToBitBucketFile, PHP_URL_PATH);
+
+        // Path for a public download follows this pattern /{user}/{repo}/downloads/{whatever}
+        // {@link https://blog.bitbucket.org/2009/04/12/new-feature-downloads/}
+        $pathParts = explode('/', $path);
+        if (count($pathParts) >= 4 && $pathParts[2] != 'downloads') {
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
      * Get file content or copy action.
      *
      * @param string $originUrl         The origin URL
@@ -342,9 +363,9 @@ class RemoteFilesystem
 
         // check for bitbucket login page asking to authenticate
         if ($originUrl === 'bitbucket.org'
+            && !self::urlIsPublicBitBucketDownload($fileUrl)
             && substr($fileUrl, -4) === '.zip'
             && preg_match('{^text/html\b}i', $contentType)
-            && $statusCode != 302
         ) {
             $result = false;
             if ($this->retryAuthFailure) {
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -176,27 +176,6 @@ class RemoteFilesystem
     }
 
     /**
-     * @link https://github.com/composer/composer/issues/5584
-     *
-     * @param string $urlToBitBucketFile URL to a file at bitbucket.org.
-     *
-     * @return bool Whether the given URL is a public BitBucket download which requires no authentication.
-     */
-    public static function urlIsPublicBitBucketDownload($urlToBitBucketFile)
-    {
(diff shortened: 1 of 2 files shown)


alcohol:
Not sure this is the right approach. Perhaps instead the check for the login page should be more explicit. Does anyone know if the logic page has a status code such as 401 or 403? Or does it just return a 200?

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -246,7 +247,12 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
         }
 
         if (isset($options['bitbucket-token'])) {
-            $fileUrl .= (false === strpos($fileUrl, '?') ? '?' : '&') . 'access_token='.$options['bitbucket-token'];
+            // First time be optimistic and do not use the token for a BitBucket download.
+            if (isset($this->bitBucketUrlsTriedWithoutAuth[$origFileUrl]) && $this->bitBucketUrlsTriedWithoutAuth[$origFileUrl]) {
```
alcohol: Isn't the `isset()` check sufficient already?

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -342,8 +348,7 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
 
         // check for bitbucket login page asking to authenticate
         if ($originUrl === 'bitbucket.org'
-            && substr($fileUrl, -4) === '.zip'
-            && preg_match('{^text/html\b}i', $contentType)
+            && substr($fileUrl, 0, 37) === 'https://bitbucket.org/account/signin/'
```
alcohol: I was wondering how this could match if the `.zip` comparison also matched, but I am guessing they have a redirect query string at the end of the url?

Seldaek:
I forked the repo so you can use https://bitbucket.org/seldaek/composer-live-test-repo in the tests instead.

That said, the whole premise here is kinda flawed IMO. Keeping track of all URLs uses memory for nothing, and having to try public first is just what might work best for you but may result in a lot of failed requests for others.

Why can't bitbucket allow authenticated requests for public stuff? Have you reached out to them about this? That would be the real solution I think.

src/Composer/Util/RemoteFilesystem.php
berlinger-rarents: Correct. It's like `https://bitbucket.org/account/signin/?next=/berlinger-rarents/my-private-repo-with-public-downloads/get/master.tar.gz`. Note also that it's not only for zip files, that's why I removed that line.

berlinger-rarents:
Thx for forking. For this test it would be nice to have a public download for a small text file in it:
bitbucket downloads screenshot

alcohol:
We don't use the public downloads from Composer's perspective, so it makes no sense to test with that.

berlinger-rarents:
I agree, this solution is focused primarily on my own use case. Which is the [phantomjs-installer](https://github.com/jakoch/phantomjs-installer/blob/master/src/PhantomInstaller/Installer.php#L167) that uses the composer DownloadManager in its post-install script to download a file that ís a bitbucket public download.

Perhaps that is the whole problem. That now BitBucket api (using oAuth2) is mixed with BitBucket site (refusing oAuth2).

I would love some example of a "normal" case of files downloaded from bitbucket.org with composer. Where could I look for examples like that?

Seldaek:
Ah sorry I thought the download would come along when forking, I don't use bitbucket that much.. It's there now.

And yeah, maybe we should only add the bitbucket token when downloading from api.bitbucket.org? Except the zip files for dist installs are not taken from api. as per https://github.com/composer/composer/blob/master/src/Composer/Repository/Vcs/GitBitbucketDriver.php#L103 and for those if they are on private packages at least we need auth, I don't know if auth is problematic on public packages. Or maybe there is an api. URL we could use instead so that we can make sure the auth is used on that alone?

berlinger-rarents:
<table>
  <tr>
    <th>url type</th>
    <th>publicity</th>
    <th>use oauth?</th>
  </tr>
  <tr>
    <td rowspan="2">api.bitbucket.org</td>
    <td>public</td>
    <td>yes</td>
  </tr>
  <tr>
    <td>private</td>
    <td>yes</td>
  </tr>
  <tr>
    <td rowspan="2">bitbucket.org/user/repo/get</td>
    <td>public</td>
    <td>doesn't matter</td>
  </tr>
  <tr>
    <td>private</td>
    <td>yes</td>
  </tr>
  <tr>
    <td rowspan="2">bitbucket.org/user/repo/downloads</td>
    <td>public</td>
    <td>no</td>
  </tr>
  <tr>
    <td>private</td>
    <td>no</td>
  </tr>
</table>

stefangr:
Downloads are also available via api.bitbucket.org.
In the case of the phantomjs installer, prepending the filename with https://api.bitbucket.org/2.0/repositories/ariya/phantomjs/downloads/ should do the magic.

PR #5720, opened 2016-09-27
Added reference to logics/extcheck

vbashkirtsev:
logics/extcheck leveraging composer's functionality to require PHP extensions. Unfortunately in the way it is currently stands most packages on Packagist fail to require extensions used. Because of this installation of such packages on minimal PHP installations succeeds but then fails at run time. I think it is reasonable to mention existence of such tool on Composer site.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -328,7 +328,8 @@ a constraint that it otherwise would not. For more information [see the
 aliases article](articles/aliases.md).
 
 `require` and `require-dev` also support references to specific PHP versions
-and PHP extensions your project needs to run successfully.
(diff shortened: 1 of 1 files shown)


staabm:
php extension can already be declared as beeing required for a certain library/project to work. what exact benefit does your tool add on top of that?

alcohol:
@staabm as far as I understand it, this tool does code analysis and checks if it makes use of any PHP extensions (e.g. non-core functionality). You can then explicitly add those to your require section I guess.

However, we cannot accept this PR, because as soon as we do, anyone and their mother will want us to promote their plugin/etc.

PR #5724, opened 2016-09-27
VersionSelector package order bug

webflo:
``composer require`` does not select the latest stable release in some situations. It looks like, the order of the packages in the Pool is important for the selection.


webflo:
Related issue on Drupal.org https://www.drupal.org/node/2801413

alcohol:
Fix your PR please. Bug fixes should be based on 1.2

alcohol:
I'll see if I can debug the cause behind this in the meanwhile.

webflo:
Closed in favour of #5726

PR #5733, opened 2016-09-28
Elaborated on importance of PHP extension dependencies

vbashkirtsev:
@Seldaek: As discussed yesterday here addition to the docs re PHP extensions. I have put it as a note as it appears emphasized for users to notice easily.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -328,7 +328,9 @@ a constraint that it otherwise would not. For more information [see the
 aliases article](articles/aliases.md).
 
 `require` and `require-dev` also support references to specific PHP versions
-and PHP extensions your project needs to run successfully.
+and PHP extensions your project needs to run successfully. If PHP extension
+is not strictly required by your project but may improve its performance/functionality
+it is recommended to list such PHP extension in `suggest`.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


stof:
I don't think this message should be talking about extensions only. The same arguments apply to optional dependencies on userland packages too.

vbashkirtsev:
Under "userland packages" you mean RPM/APT packages?

stof:
no, I mean packages written in PHP (by opposition of PHP extension which are loaded in the engine itself). ``userland`` is the term used in the PHP ecosystem when needing to distinguish things implemented in core/extensions vs things implemented in PHP code.

vbashkirtsev:
In my view composer already manages PHP packages quite well. Non-composer packages: nothing should be done about them - they should convert :)

I believe any good package should have a test suite. Execution of one immediately demonstrates which dependencies are missing (provided good coverage). While such test suite is quite efficient to identify composer package dependencies it is utterly inefficient to identify PHP extensions used unless it is run on --disable-all system (and I don't know alot of developers doing so). Major issue that normally developer runs the tests on his own dev box where generally all used extensions are present or on Travis (and it also has good set of extensions) and in this case all tests pass without any issues. But then package lands on minimal install system: composer packages will get in, PHP extensions will not. Hence why there's need IMO to emphasize the need to add "ext-*" to composer.json and not mention optional userland packages in order not to distract away from the issue at hand.

PR #5738, opened 2016-09-30
Reduce calls on Rule::getHash()

Toflar:


stof:
this will still call ``getHash`` twice in ``RuleSetGenerator::addRule``, as it checks ``containsEqual`` and then calls ``add``.
You should add a new method in RuleSet ``addIfNotExists`` or something like that, which would do the ``containsEqual`` internally, allowing to compute the hash only once.

Toflar:
Before:
bildschirmfoto 2016-09-30 um 15 38 19

After:
bildschirmfoto 2016-09-30 um 15 38 37

naderman:
I don't think there is any need for an add function which does not check for existence. I'm pretty sure that the add() call in the Solver will never have a `$newRule` that is already in the set cause it wouldn't have generated it then. So I think you should condense the 3 functions in the Rulset all into one single function and get rid of the optional argument etc, which should simplify this code a bit.

Toflar:
Am I allowed to do this because of BC? People might use `containsEqual()` etc.?

naderman:
The solver is considered internal, and I really can't see why any plugin would want to deal with rules directly.

Toflar:
Okay, just verifying :) Will work on it 👍

Toflar:
@naderman better?

src/Composer/DependencyResolver/RuleSet.php
```diff
@@ -62,13 +62,24 @@ public function add(Rule $rule, $type)
             $this->rules[$type] = array();
         }
 
+        $hash = $rule->getHash();
```
naderman: think this should be moved up above the initialization of the per-type array.

Toflar: There you go :)

naderman: Sorry i meant the entire check, not just that line, so the type array won't be initialized if no rule gets added later

naderman:
Awesome, thanks a lot @Toflar

PR #5739, opened 2016-09-30
[wip] composer repository json schema

alcohol:
A stab at writing a more formal specification that covers the various metadata files that packagist hosts.

Ideally it would include refs to the composer-schema.json definitions for the package schemas contained in each version.

### DIFF
--- a/res/composer-repository-schema.json
+++ b/res/composer-repository-schema.json
@@ -1,0 +1,99 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "A representation of packages metadata.",
+    "type": "object",
+    "oneOf": [
+        { "required": [ "packages" ] },
+        { "required": [ "providers" ] },
+        { "required": [ "provider-includes", "providers-url" ] }
+    ],
+    "properties": {
+        "packages": {
+            "type": ["object", "array"],
+            "description": "A hashmap of package names in the form of <vendor>/<name>.",
+            "additionalProperties": { "$ref": "#/definitions/versions" }
+        },
+        "providers-url": {
+            "type": "string",
+            "description": "Endpoint to retrieve provider data from, e.g. '/p/%package%$%hash%.json'."
+        },
+        "provider-includes": {
+            "type": "object",
+            "description": "A hashmap of provider listings.",
+            "additionalProperties": { "$ref": "#/definitions/provider" }
+        },
+        "providers": {
+            "type": "object",
+            "description": "A hashmap of package names in the form of <vendor>/<name>.",
+            "additionalProperties": { "$ref": "#/definitions/provider" }
+        },
+        "notify-batch": {
+            "type": "string",
+            "description": "Endpoint to call after multiple packages have been installed, e.g. '/downloads/'."
+        },
+        "search": {
+            "type": "string",
+            "description": "Endpoint that provides search capabilities, e.g. '/search.json?q=%query%&type=%type%'."
+        },
+        "warning": {
+            "type": "string",
+            "description": "A message that will be output by Composer as a warning when this source is consulted."
+        }
+    },
+    "definitions": {
+        "versions": {
+            "type": "object",
+            "description": "A hashmap of versions and their metadata.",
+            "additionalProperties": { "$ref": "#/definitions/version" }
+        },
+        "version": {
+            "type": "object",
+            "oneOf": [
+                { "$ref": "#/definitions/package-generic" },
+                { "$ref": "#/definitions/package-metapackage" }
+            ]
+        },
+        "package-generic": {
+            "properties": {
+                "name": { "type": "string" },
+                "type": { "type": "string" },
+                "version": { "type": "string" },
+                "version_normalized": {
+                    "type": "string",
+                    "description": "Normalized version, optional but can save computational time on client side."
+                },
+                "source": { "type": "object" },
+                "dist": { "type": "object" },
+                "time": { "type": "string" }
+            },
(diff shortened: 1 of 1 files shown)


alcohol:
@stof I would appreciate it if you could have a look at this. You always have an eye for detail, would like to hear if you see anything that I missed or that could be improved. :-)

Also feedback on maybe including refs to the composer-schema.json regarding even deeper specification of package json data inside versions, would be appreciated.

res/composer-repository-schema.json
```diff
@@ -0,0 +1,105 @@
...
+            "properties": {
+                "name": { "type": "string" },
+                "type": { "type": "string", "enum": [ "metapackage" ] },
+                "version": { "type": "string" },
+                "version_normalized": {
+                    "type": "string",
+                    "description": "Normalized version, optional but can save computational time on client side."
+                },
+                "source": { "type": "object" },
+                "dist": { "type": "object" },
```
Seldaek: I would drop source/dist here as additionalProperties allow them but they are not strictly useful or needed

res/composer-repository-schema.json
```diff
@@ -0,0 +1,105 @@
...
+        },
+        "providers": {
+            "type": "object",
+            "description": "A hashmap of package names in the form of <vendor>/<name>.",
+            "additionalProperties": { "$ref": "#/definitions/provider" }
+        },
+        "notify": {
+            "type": "string",
+            "description": "Endpoint to call after a package has been installed, e.g. '/downloads/%package%'."
+        },
```
Seldaek: notify is kinda historical as well, notify-batch really should be preferred as it's a lot more efficient, I'd say remove.

res/composer-repository-schema.json
alcohol: I thought one of either was required (except if type is metapackage) ?

res/composer-repository-schema.json
alcohol: Will do.

res/composer-repository-schema.json
Seldaek: Yes.. and this is type metapackage so it isn't.

Seldaek: Ah no this got lost in the wrong place after I had to reload, well anyway I meant it for line 88/89 below

res/composer-repository-schema.json
```diff
@@ -0,0 +1,99 @@
...
+                "name": { "type": "string" },
+                "type": { "type": "string" },
+                "version": { "type": "string" },
+                "version_normalized": {
+                    "type": "string",
+                    "description": "Normalized version, optional but can save computational time on client side."
+                },
+                "source": { "type": "object" },
+                "dist": { "type": "object" },
+                "time": { "type": "string" }
```
stof: this is expected to be a date in ISO format. Isn't there a better tupe to represent it ?

res/composer-repository-schema.json
```diff
@@ -0,0 +1,99 @@
...
+            },
+            "oneOf": [
+                { "required": [ "name", "version", "source" ] },
+                { "required": [ "name", "version", "dist" ] }
+            ],
+            "additionalProperties": true
```
stof: ``autoload`` and links (require, conflict, replace, provide) should be listed explicitly IMO.

Links are also needed in metapackages (they are their main use case)

res/composer-repository-schema.json
```diff
@@ -0,0 +1,99 @@
...
+                "name": { "type": "string" },
+                "type": { "type": "string", "enum": [ "metapackage" ] },
+                "version": { "type": "string" },
+                "version_normalized": {
+                    "type": "string",
+                    "description": "Normalized version, optional but can save computational time on client side."
+                },
+                "time": { "type": "string" }
+            },
+            "required": [ "name", "version" ],
```
stof: type is required in this case actually

res/composer-repository-schema.json
alcohol: To my knowledge, there is no type more specifically suited for date/time.

res/composer-repository-schema.json
alcohol: Good point.

res/composer-repository-schema.json
stof: A search in the json-schema spec gave me ``format: "date-time"`` for that.

alcohol: Ah; found it http://json-schema.org/latest/json-schema-validation.html#anchor108

Sure, we can add that.

res/composer-repository-schema.json
alcohol: So should links then be part of `required` for metapackages in your opinion?

res/composer-repository-schema.json
stof: Packagist uses the expected format: https://github.com/composer/packagist/blob/42c8375b7dba899f41da60cd82c3983c8d9f4c9e/src/Packagist/WebBundle/Entity/Version.php#L242

It looks like the ArrayDumper in Composer (which is used by Satis) uses a different format (not breaking things for composer, as it will accept any format supported by the ``DateTime`` constructor).
I suggest updating the ArrayDumper to use the standard format

res/composer-repository-schema.json
stof: no, because not all of them are required (you can omit any of them), and having none would not be an issue for composer (the metapackage would be quite useless though, as it would not bring you any code). But IMO, they deserve being treated as first-class citizens (allowing to have more validation on them)

PR #5740, opened 2016-09-30
Add php-debug and php-zts virtual platform packages

dzuelke:


src/Composer/Repository/PlatformRepository.php
```diff
@@ -81,6 +81,18 @@ protected function initialize()
         $php->setDescription('The PHP interpreter');
         $this->addPackage($php);
 
+        if (PHP_DEBUG) {
+            $phpdebug = new CompletePackage('php-debug', $version, $prettyVersion);
+            $phpdebug->setDescription('The PHP interpreter, with debugging symbols');
+            $this->addPackage($phpdebug);
+        }
+
+        if (PHP_ZTS) {
```
Seldaek: Are those consts available since php 5.3? I guess so as it's nothing new but I didn't find any docs on them.

Seldaek: Ah well, it's not avail on HHVM it seems..

staabm: See https://3v4l.org/KLWNP

Seldaek: Right, seems both are available since 5.2.7 except for ZTS on HHVM https://3v4l.org/Hb4dI good idea with 3v4l :)

PR #5745, opened 2016-10-01
Warn about possible retag of version

xiian:
In the inadvisable event that a package has been retagged, it’s possible to be in a state where the versions are the same, but the underlying ref is different. We report that there has been a version variation, but the output is not helpful (`From 1.1.3 to 1.1.3`, which just causes more confusion).

Since this is an exceptional case, it makes sense to give some insight into why things may appear strange.

I'm not 100% that this message encompasses all possible reasons why we'd be in this state, so suggestions are welcome, but I think it is important to show some sort of messaging with possible explanations for the super confusing status.

### DIFF
--- a/src/Composer/Command/StatusCommand.php
+++ b/src/Composer/Command/StatusCommand.php
@@ -188,7 +188,7 @@ EOT
                     $io->write(sprintf('    From <comment>%s</comment> to <comment>%s</comment>', $previousVersion, $currentVersion));
 
                     if ($currentVersion === $previousVersion) {
-                        $io->write('    <warning>Version may have been updated since composer.lock was last updated</warning>');
+                        $io->writeError('    <warning>Version may have been updated since composer.lock was last updated</warning>');
                     }
                 } else {
                     $io->write($path);
--- a/src/Composer/Command/StatusCommand.php
+++ b/src/Composer/Command/StatusCommand.php
@@ -186,6 +186,10 @@ EOT
 
                     $io->write('<info>'.$path.'</info>:');
                     $io->write(sprintf('    From <comment>%s</comment> to <comment>%s</comment>', $previousVersion, $currentVersion));
+
+                    if ($currentVersion === $previousVersion) {
+                        $io->write('    <warning>Version may have been updated since composer.lock was last updated</warning>');
+                    }
                 } else {
                     $io->write($path);
                 }


src/Composer/Command/StatusCommand.php
```diff
@@ -186,6 +186,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
                     $io->write('<info>'.$path.'</info>:');
                     $io->write(sprintf('    From <comment>%s</comment> to <comment>%s</comment>', $previousVersion, $currentVersion));
+
+                    if ($currentVersion === $previousVersion) {
+                        $io->write('    <warning>Version may have been updated since composer.lock was last updated</warning>');
```
alcohol: This should be `writeError()`.

But perhaps the `isVeryVerbose()` check on IO should be removed instead (line 181). If we always include the ref in the output/message, it would be very clear when the versions do not point to the same ref anymore.

xiian: I had originally done `if ($io->isVeryVerbose() || $currentVersion === $previousVersion)` on line 181 so that the refs would be included instead, but figured it'd be better to call out the problem with more specific information so the user would know it's not necessarily their fault.

I've always liked that composer will *just* let you know about the version mismatch, since that's really the pertinent information. Including the refs is nice extra information, but I feel like having a friendly message helps people get to the root of the issue quicker without adding to the noise.

Regarding `writeError()`, I was just following the rest of the code block, but agree that this is more of an error state. Will be happy to update.

alcohol: Alright. If you want to stick to the extra message instead, then yeah, please use `writeError` as this is only informational output (thus belongs on `stderr`).

The message itself is a bit ambiguous for me. "Version may have been updated" does not make it explicitly clear that this is referring to a tag having been moved (versions are often based on tags, but **not** always).

xiian: If you have suggestions on clarifying the message, I'll gladly change it, but it sounds like we could have different ways to get to this state (since versions aren't necessarily based on tags) and all we know is that the version is the same, but the ref is different. So the ambiguity is a necessary evil.

In the cases that I'm dealing with, it appears that annotated tags were used (which have a different sha1 than the commit they point to, hence the problem), so it's not even that we'd get different code after checking out the different ref. Not entirely sure what order of events occurred to get us to this state, but it's been confusing to say the least.

I pushed up a commit to switch to using `writeError`, and will rebase that extra commit away before merge. Didn't want to tempt the Github PR discussion interface with a force push.

Seldaek:
The problem with the wording IMO is that it has nothing to do with composer.lock really, it just means you are updating to something different than what you have installed, despite it having the same version number. I don't know how to word this sanely, but I also don't think this happens frequently enough that it's such a big deal. I'm ok having a warning of some kind but the wording should be more accurate.

Seldaek:
Closing because it's not really all that important and if you are interested in details `composer status -vv` shows you the exact commit reference always.

PR #5752, opened 2016-10-04
changes based on feedback from @stof in #5739

alcohol:


res/composer-repository-schema.json
```diff
@@ -62,29 +62,34 @@
                     "type": "string",
                     "description": "Normalized version, optional but can save computational time on client side."
                 },
-                "source": { "type": "object" },
-                "dist": { "type": "object" },
+                "autoload": { "type": "object" },
+                "require": { "type": "object" },
```
stof: you should add more validation on them (links are maps with strings as values), and ``autoload`` has a specific schema too

res/composer-repository-schema.json
```diff
@@ -62,29 +62,34 @@
                     "type": "string",
                     "description": "Normalized version, optional but can save computational time on client side."
                 },
-                "source": { "type": "object" },
-                "dist": { "type": "object" },
```
stof: shouldn't these be listed explicitly for packages ?

alcohol: They are now in the package-base schema. Or do I misunderstand your comment?

res/composer-repository-schema.json
alcohol: Couldn't they also contain arrays as values? Not just strings?

PR #5765, opened 2016-10-06
removed obsolete code

fabpot:
Composer requires PHP 5.3.2 as a minimum.

The code I've removed in this PR is for a bug that was fixed in 5.3.2, so it should be safe (https://github.com/php/php-src/blob/PHP-5.3.29/NEWS#L2080).


Seldaek:
I'll merge in master because I agree it's probably time we drop this.. but I don't think it's nice to risk BC issues in a patch release.

PR #5775, opened 2016-10-09
Clarify allowed format for repositories key

### LINKED ISSUE #2802: Document "repositories" in key-value form
https://getcomposer.org/doc/04-schema.md#repositories and https://getcomposer.org/doc/05-repositories.md only give "repositories" section in a plain array form
```
{
    "repositories": [
        {
            "type": "composer",
            "url": "http://packages.example.com"
        }
    ]
}
```

Running cli command `php composer.phar config repositories.foo vcs https://github.com/foo/foo` I get it in a key-value form
```
{
    "repositories": {
        "foo": {
            "type": "vcs",
            "url": "https://github.com/foo/foo"
        }
    }
}
```
I could not find any reference that this form is a supported feature so I could rely my scripting on it.

tomasfejfar:
Using JSON object should be discouraged as there is no explicit order
of keys in JSON object. Even though it's deterministic in current PHP
version, it may change any time. As stated on http://json.org/:
"An object is an unordered set of name/value pairs. An object begins
with { (left brace) and ends with } (right brace). Each name is
followed by : (colon) and the name/value pairs are separated
by , (comma)."

Fixes #2802.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -775,14 +775,15 @@ will look from the first to the last repository, and pick the first match.
 By default Packagist is added last which means that custom repositories can
 override packages from it.
 
-Using JSON object is also allowed. However, it is discouraged because there is no explicit order of keys in JSON object.
+Using JSON object notation is also possible. However, JSON key/value pairs
+are to be considered unordered so consistent behaviour cannot be guaranteed.
 
  ```json
 {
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

tomasfejfar:
Happy to help 😎

PR #5781, opened 2016-10-11
Upgrade json-schema to allow v4

jyggen:


staabm:
these are the release notes for 4.x:

Return points, spelling and dead code
Add support for type coercion
performance improvements
Updated to check that $ref is a string
fix(NumberConstraint): force positive value on first arg
Add support for $ref on properties
chore(phpunit): ignore local phpunit.xml
#319 feat(demo): add simple demo script

PR #5785, opened 2016-10-14
Enhance performance by removing count() from Solver's main loop

nicolas-grekas:
Thanks to [blackfire.io](https://blackfire.io), I spotted that `count()` is hurting performance for no reason here. It was called 730023965 times in the case I was working on, not unusual.

Using `time` to "profile" the patch, `composer up` dropped from 290s to 140s (PHP7.1)!

### DIFF
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -758,8 +758,10 @@ class Solver
                 $systemLevel = $level;
             }
 
-            for ($i = 0, $n = 0; $n < count($this->rules); $i++, $n++) {
-                if ($i == count($this->rules)) {
+            $rulesCount = count($this->rules);
+
+            for ($i = 0, $n = 0; $n < $rulesCount; $i++, $n++) {
+                if ($i == $rulesCount) {
                     $i = 0;
                 }
 
@@ -805,6 +807,7 @@ class Solver
                 }
 
                 // something changed, so look at all rules again
+                $rulesCount = count($this->rules);
                 $n = -1;
             }
 


src/Composer/DependencyResolver/Solver.php
```diff
@@ -682,8 +682,9 @@ private function runSat($disableRules = true)
         $level = 1;
         $systemLevel = $level + 1;
         $installedPos = 0;
+        $rulesCount = count($this->rules);
 
-        while (true) {
+        for (;;) {
```
romainneutron: does it impact the performance?
I think it's really easier to read `while (true) {` than `for (;;) {`. Especially for newcomers

nicolas-grekas: When done 730M times, yes

stof: This looks like a potential optimization for OPCache though, as it should be able to recognize that they are the same.

nicolas-grekas: Well, in fact, measuring is the only way to tell, so I did. Result: no difference at all! Reverted to `while (true)`.

stof:
@nicolas-grekas could you share the link to the blackfire comparison ?

src/Composer/DependencyResolver/Solver.php
```diff
@@ -713,7 +714,7 @@ private function runSat($disableRules = true)
                             }
                         }
 
-                        if ($noneSatisfied && count($decisionQueue)) {
+                        if ($noneSatisfied && $decisionQueue) {
```
xabbuh: Does it make a difference in terms of performance when using `empty()` here instead? Imo that would be more readable than how it is now.

src/Composer/DependencyResolver/Solver.php
stof: looks like PHP already understands the optimization I was talking about :smile:

src/Composer/DependencyResolver/Solver.php
alterphp: Cast from array to boolean is indeed not obvious

nicolas-grekas:
Here is the blackfire comparison: https://blackfire.io/profiles/compare/01bc08d5-9907-497c-909a-8d2bf9e5d9d3/graph

src/Composer/DependencyResolver/Solver.php
nicolas-grekas: Put back `count()` here, since on arrays, it's fast (`$this->rules` is a countable object, which is why it's "slow" here)

Seldaek:
Oh nice catch, although here I only see a 17->15s improvement which isn't as impressive but I guess this depends a lot on the dependency graph. In any case any improvement is welcome :) We should make sure the patch is 100% safe though.

naderman:
@nicolas-grekas how certain are you that there isn't any other place in between there that changes the count of rules? Just want to avoid that this breaks some weird edge case that we don't have tests for. we really lack real life large project examples as tests.

nicolas-grekas:
@naderman I read the code, no better... Unless I read it wrong, `$this->rules` looks like a locally created object (local to the Solver instance), which eases a lot on this topic. The looking at all uses of the property.

naderman:
Well it's only modified in `setPropagateLearn` after the initial creation in `solve()`. `setPropagateLearn` in turn is called directly once (that's where you update the count currently), and twice through `selectAndInstall` on lines 737 and 802. So I think the count should be updated after those calls as well.

nicolas-grekas:
Indeed, should be better & safe now.

naderman:
@nicolas-grekas ok this version I am definitely fine with, did this slow it down much compared to the first version though?

nicolas-grekas:
Yes, just benched, same positive perf impact, the callgraph is the same.

staabm:
> Here is the blackfire comparison: https://blackfire.io/profiles/compare/01bc08d5-9907-497c-909a-8d2bf9e5d9d3/graph

@nicolas-grekas  why is Composer\DependencyResolver\RuleSet::add invoked ~300.000 more often on the diff.

naderman:
@staabm Hm indeed that looks odd, @nicolas-grekas can you post an updated comparison with the most recent version of the patch? Did you look at memory consumption too?

stof:
@naderman memory consumption is the same in the blackfire comparison ( ``+0.018%`` is not relevant IMO).

Regarding the calls to ``add``, this is because ``selectAndInstall`` (one of the way to alter rules) is called twice, but only one of the recomputes the ``$ruleCount``. Btw, ``setPropagateLearn`` is also called once outside ``selectAndInstall``, and this also misses the recomputation

src/Composer/DependencyResolver/Solver.php
```diff
@@ -758,8 +758,10 @@ private function runSat($disableRules = true)
                 $systemLevel = $level;
             }
 
-            for ($i = 0, $n = 0; $n < count($this->rules); $i++, $n++) {
-                if ($i == count($this->rules)) {
+            $rulesCount = count($this->rules);
+
+            for ($i = 0, $n = 0; $n < $rulesCount; $i++, $n++) {
+                if ($i == $rulesCount) {
```
dereuromark: should be ===

PR #5787, opened 2016-10-14
Improve installer extensibility

Toflar:
Hi everyone,

This is a follow up PR on https://github.com/composer/composer/pull/4946.
I'm quoting what @Seldaek wrote in the other PR:

> I think requiring composer/composer as dependency and doing things there how you want it makes more sense than adding a flag for anyone to shoot themselves in the foot with :)

This is what I set out to do now. Now the problem is that `Installer::run()` does a few things "hardcoded". For example, if the dry run mode is enabled, the output is always verbose an scripts are never run. Also, I cannot force composer to write my lock file even in dry run mode. It's all hard coded in the `run()` method and even if I'd extend the `Installer` and tried to override the `run()` method, it's not working because the method is so huge and does so many things that I cannot really call like `parent::run()` and then in addition to this do something else again, because all the variables are obviously missing (in my case e.g. `$aliases` etc.).

So I reworked the `Installer` just a little bit by moving the defaults for the dry run mode to the `setDryRun()` method instead. That way, everybody can be more flexible on working with the installer.

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -109,6 +109,7 @@ class Installer
     protected $dryRun = false;
     protected $verbose = false;
     protected $update = false;
+    protected $writeLock = true;
     protected $dumpAutoloader = true;
     protected $runScripts = true;
     protected $ignorePlatformReqs = false;
@@ -179,13 +180,6 @@ class Installer
             $this->update = true;
         }
 
-        if ($this->dryRun) {
-            $this->verbose = true;
-            $this->runScripts = false;
-            $this->installationManager->addInstaller(new NoopInstaller);
-            $this->mockLocalRepositories($this->repositoryManager);
-        }
-
         if ($this->runScripts) {
             // dispatch pre event
             $eventName = $this->update ? ScriptEvents::PRE_UPDATE_CMD : ScriptEvents::PRE_INSTALL_CMD;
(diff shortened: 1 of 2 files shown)


alcohol:
> Also, I cannot force composer to write my lock file even in dry run mode. 

Well that is a rather obvious one. Because it would not be a `dry-run` anymore then.

Toflar:
Yeah sure, I want to write the lock file without installing any libraries. This is not a needed feature for composer itself, which is why the other PR was closed. I do want to do this though and I cannot without overriding the whole `Installer::run()`. This PR doesn't change a thing for composer, it just makes the `Installer` more flexible to use for developers. That's all.

alcohol:
Fair enough. The changes look minimal.

Seldaek:
This is breaking BC really because if you call setDryRun and then other setters, the dry-run settings would now get overwritten by the other setters, while the current master branch just enforces the dry-run settings when run() is called.

For example if you look at InstallCommand setDryRun is called and then setRunScripts is called, etc. Of course we could adjust those usages, but it'd also break other people relying on the Installer class externally as you are trying to do.

IMO the best to keep it BC would be to add a couple flags like you did with writeLock and maybe add another executeOperations flag, so that if you set dry-run, it would set both of these to false and the other stuff as well as it is doing right now on top of run(), but if you want your own "dry-run-like" behavior you could set executeOperations false and runScripts false but keep writeLock true for example. Does that make sense?

Toflar:
First of all: the easiest way for me to achieve what I need is to add a simple `setForceWriteLock()` method and keep everything else as it is. I just found that it would cover just my use case and I wanted to find a more generic approach. That's why I changed other stuff. If you feel better with adding a simple additional flag, I'm totally okay with it because we have to make sure Composer's as stable as possible. So if you prefer this solution, just tell me and I'll adjust the PR accordingly and we leave the other discussion aside :-)

Coming back to my proposal: I don't really think it is a BC break to be honest. I obviously agree that it might change the behavior depending on the order you called the methods of the installer. Here's an example:

```php
$installer->setDryRun(true)->setDumpAutoload(true);
```

Behavior now: No autoload.php is dumped.
Behavior after: autoload.php is dumped.

Or this:

```php
$installer->setDryRun(true)->setRunScripts(true);
```

Behavior now: Scripts are not run.
Behavior after: Scripts are run.

But basically, if you did so until now, you were relying on a behavior of the installer that did not really represent its public api. I mean, if you call `setRunScripts(true)` you should be expecting it to happen, no matter what other methods you called before or called after ;-)

So to sum up: I agree it might change the behavior although I'm not sure if in this very case it would imply a BC break. But if we don't want this change of behavior - again - I'm very much okay with it but then I'd also leave everything else as is and just add a `forceWriteLock` flag.

Seldaek:
Well a change in behavior is a BC break, it's acceptable when doing bug fixes, but IMO changing a dry-run mode is a big WTF-inducing BC break, because you expect something to just do a dummy run and print output, and suddenly it starts writing stuff on disk, and we can not even warn about the BC break or a misuse of the code, so this is about as bad as it gets in terms of BC break.

I still think keeping the flag you added and adding a second one would let us clean this up nicely, but I can also do the changes myself if you aren't clear on what I explained above. It's not so much work anyway.

Toflar:
I don't think you should do the work as you're doing enough already anyway :-) However, I did not fully understand your `executeOperations` idea. Can you maybe provide a few lines of code to illustrate what you mean?

Seldaek:
What I mean is `$this->executeOperations` would replace `$this->dryRun` everywhere it's used right now in `if ([!]$this->dryRun) { ... }`, so that the dryRun flag controls only one thing: the suppression of other flags on top of the run() command, which would make it BC as if it is set it shadows other flags, no matter what they are set to after setDryRun has been called. I hope that's clearer :)

Toflar:
I see. I'll give it a try asap :)

Toflar:
@Seldaek did you mean something like this?

Seldaek:
Ahh well I meant more disableOperations as in installing/updating/removing packages are "operations" which means if we disable all these on dry-run the dry-run would have the same effect as it had before, but anyway I think your model works too, I'll have to double check later in more details though.

Seldaek:
Hi @Toflar I had a look and it seemed like dryRun was gonna call stuff like dumpAutoload still in your PR then I played with it from scratch locally and well I ended up with something very similar but not using your commit.. Sorry for that last bit, but on the upside it's merged :)

Toflar:
Hey @Seldaek, cheers! 

> was gonna call stuff like dumpAutoload still in your PR

Yeah, that was intentional so you can enable or disable it using `setDumpAutoload()` and I didn't want to change the current behaviour (output of "Generating autoload files" even though you did `--dry-run`) ;-) But anyway, looks good, will test it in my cloud resolver probably next week 👍

Seldaek:
But it didn't print the autoload stuff as it was inside the `if (!$this->dryRun)` block before, that's why your PR threw me off initially. Anyway doesn't matter :)

Toflar:
Hey Jordi, sorry for the late feedback, wanted to test your commits now. I use the installer as follows now:

```php
$installer = Installer::create($io, $composer)
    ->setUpdate(true) // Update
    ->setDryRun(false) // Disable dry run
    ->setWriteLock(true) // Still write the lock file
    ->setVerbose(true) // Always verbose for composer. Verbosity is managed on the IO
    ->setDevMode(true) // Default is true, use --no-dev to disable
    ->setDumpAutoloader(false) // Never dump the autoload file
    ->setExecuteOperations(false) // Disable the operations
;
```

So as you can see, I'm using your `setExecuteOperations(false)` in combination with `setWriteLock(true)` so that my `composer.lock` file is still written. That's what I'm expecting but it does not write a `composer.lock` file at all now. Just verifying that we had the same intention because that use case was the reason for the PR. I can debug why this is the case for me but just wanted to get your confirmation that what I'm expecting (`composer.lock` file still written) is what you would've expected from your code changes as well :-D

Seldaek:
That is definitely how I expected it to work :)

Toflar:
Okay, so I tracked this down and I have to admit that I don't understand enough of the resolving process with the different repositories yet which is why I cannot come up with a PR. I'll need your help to point me in the right direction. The reason why it does not work is this part: https://github.com/composer/composer/blob/master/src/Composer/Installer.php#L608

If `$localRepo->write();` is not called (which it is not if I set `setExecuteOperations()` to `false`), there won't be any packages the locker could write. Does that help?

Toflar:
@Seldaek want me to open a new issue for this?

Issue #5769, opened 2016-10-07
Set Environment Variable When `no-dev` option is passed


josephabrahams:
It would be very useful to be able to determine is the `no-dev` option has been set when running `post-install-cmd` scripts.

alcohol:
You will have to be more specific.

josephabrahams:
We use bash scripts defined in `post-install-cmd` to run administrative actions for setting up a site after a new deploy. Some scripts should/shouldn't run upon installation in a local development environment. Given that the --no-dev flag in passed for production installed, it would be nice if there an a way to test for this option from the Unix environment.

axel-rutz:
SO has some hints on this: https://stackoverflow.com/questions/13087088/composer-run-scripts-only-in-dev-environment

Seldaek:
There is now (as of 1.3.0) a COMPOSER_DEV_MODE env var which lets you react based on the --no-dev flag in scripts.

PR #5794, opened 2016-10-17
Issue 5769 dev mode varaible

andrewjt71:
**Issue addressed**
https://github.com/composer/composer/issues/5769

**Summary**
Sets a php environment variable during the composer install/update, to expose to any post-install/update scripts which mode the install/update was run in.

**Example Usage (In post install/update script)**

```
if (getenv("COMPOSER_DEV_MODE") ==1) {
    // do something
}
```

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -292,6 +292,9 @@ class Installer
             }
 
             if ($this->runScripts) {
+                $devMode = (int) $this->devMode;
+                putenv("COMPOSER_DEV_MODE=$devMode");
+
                 // dispatch post event
                 $eventName = $this->update ? ScriptEvents::POST_UPDATE_CMD : ScriptEvents::POST_INSTALL_CMD;
                 $this->eventDispatcher->dispatchScript($eventName, $this->devMode);
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -22,15 +22,19 @@ Composer fires the following named events during its execution process:
 
 ### Command Events
 
-- **pre-install-cmd**: occurs before the `install` command is executed with a lock file present.
-- **post-install-cmd**: occurs after the `install` command has been executed with a lock file present.
-- **pre-update-cmd**: occurs before the `update` command is executed, or before the `install` command is executed without a lock file present.
-- **post-update-cmd**: occurs after the `update` command has been executed, or after the `install` command has been executed without a lock file present.
+- **pre-install-cmd**: occurs before the `install` command is executed with a
+  lock file present.
+- **post-install-cmd**: occurs after the `install` command has been executed
+  with a lock file present.
+- **pre-update-cmd**: occurs before the `update` command is executed, or before
+  the `install` command is executed without a lock file present.
+- **post-update-cmd**: occurs after the `update` command has been executed, or
+  after the `install` command has been executed without a lock file present.
 - **post-status-cmd**: occurs after the `status` command has been executed.
 - **pre-archive-cmd**: occurs before the `archive` command is executed.
 - **post-archive-cmd**: occurs after the `archive` command has been executed.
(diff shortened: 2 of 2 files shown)


doc/03-cli.md
```diff
@@ -94,7 +94,9 @@ resolution.
   installation and show you what would happen.
 * **--dev:** Install packages listed in `require-dev` (this is the default behavior).
 * **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader
-  generation skips the `autoload-dev` rules.
+  generation skips the `autoload-dev` rules. This sets a `COMPOSER_DEV_MODE` php
+  environment variable to true, to expose to any post-install commands whether the
+  install was run in dev mode or not. that.
```
alcohol: "that." ?

doc/03-cli.md
```diff
@@ -139,7 +141,10 @@ php composer.phar update vendor/*
   fulfill these. See also the [`platform`](06-config.md#platform) config option.
 * **--dry-run:** Simulate the command without actually doing anything.
 * **--dev:** Install packages listed in `require-dev` (this is the default behavior).
-* **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader generation skips the `autoload-dev` rules.
+* **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader
+  generation skips the `autoload-dev` rules. This sets a `COMPOSER_DEV_MODE` php
+  environment variable to true, to expose to any post-update commands whether the
+  update was run in dev mode or not. that.
```
alcohol: Same as above.

src/Composer/Installer.php
```diff
@@ -291,7 +291,11 @@ public function run()
                 $this->autoloadGenerator->dump($this->config, $localRepo, $this->package, $this->installationManager, 'composer', $this->optimizeAutoloader);
             }
 
+            // set php environment variable so post install/update scripts can determine which mode install/update was run in
```
alcohol: I don't think this comment is necessary. The code speaks for itself.

src/Composer/Installer.php
```diff
@@ -291,7 +291,11 @@ public function run()
...
             if ($this->runScripts) {
+                $devMode = $this->devMode;
+                putenv("COMPOSER_DEV_MODE=$this->devMode");
```
alcohol: I would suggest casting it to an integer, 0 or 1 is easiest to deal with in most shell scripts.

andrewjt71: 👍  Good point, will change this

doc/03-cli.md
andrewjt71: I have no idea where `that.` came from! Long day!

andrewjt71:
Hi @alcohol, thanks for your feedback. I have addressed your points except the one regarding updating documentation in the `docs` directory. I assume you're referring to the `doc` directory, but this is the directory in which the documentation I initially modified existed (`doc/03-cli`). Have removed the docs I created, but happy to add some documentation around this feature if you give me some more guidance on where they should live. Thanks!

alcohol:
@andrewjt71 yes sorry, I should have phrased that better. It would be great if you could document this environment variable in the following section: https://github.com/composer/composer/blob/master/doc/03-cli.md#environment-variables

stof:
@alcohol it should not be documented in the cli.md chapter (which lists env variables allowing to control composer). It should be documented in the doc about scripts

andrewjt71:
Hi @alcohol, have added some docs as requested. Feel free to provide any additional feedback if required. Thanks!

alcohol:
I rephrased it slightly and applied some wordwrapping. Thanks!

andrewjt71:
Thanks for reviewing @alcohol!

Seldaek:
Thanks

PR #5797, opened 2016-10-18
fixed direct rename() usage

msiebeneicher:
We runned into an issue with composer/satis and the dist package creation, that we couldn't create the packages properly under FreeBSD:

```
[ErrorException]
  rename(/tmp//composer_archive58061e3f534fe.tar,/path/package-v1.0.7-20d198.tar): Operation not permitted
```

This small PR resolve that issue


alcohol:
Ah, I think I ran into that one too at some point. Thought it was a fluke though cause I was unable to reproduce it consistently.

msiebeneicher:
@alcohol : and you merge it than from the 1.2 branch back to the master branch?

alcohol:
Yes that is what we generally do for bugfixes. Apply to lowest version still supported, and then merge into the rest.

msiebeneicher:
ok - i have to do a new pull request for that i guess

alcohol:
Cheers.

PR #5803, opened 2016-10-20
Make a named repository definition leading/immutable

jmglsn:
I was looking for a way to redefine a repository source which was already named in a projects composer configuration. This is needed as the repository which is providing one of the required components is available on different networks, depending where the consumer is located. (Just imagine you need to change from ssh to https.)
First approach was to define that repository in global composer config, but as it turns out it don't make sense to define a named repository in that config, as it will always be overruled by any same named repository, if located in a project composer file.
So I added merge logic for a new attribute `immutable`. If any named repository has that attribute enabled, it will not be overruled by any other repository config for the same name.
- allows to define named repositories that will not be overruled by another same named repository
- allows to define required repository sources that can have a different source configuration

Expected usage / example:
global composer config

``` json
{
    "config": {
    },
    "repositories": {
        "a-company-component-repository": {
            "type": "git",
            "url": "https://1.2.3.4/scm/component.git",
            "immutable": true
        },
    }
}
```

project composer config

``` json
    "autoload-dev": {
    ...
    },
    "repositories": {
        "a-company-component-repository": {
            "type": "git",
            "url": "ssh://4.3.2.1:8150/intranet/whatever/component.git"
        },
    ...
    },
    ....
```

When it now comes to any composer action, `ssh://4.3.2.1:8150/intranet/whatever/component.git` should not be involved, as it's configuration is shadowed by the global composer entry for `a-company-component-repository`.

If there is already another way to solve this problem, please let me now - I spend already some time searching for this. (I don't want to host a proxy for this ;) )

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -264,6 +264,50 @@ GitHub and BitBucket:
 
 The only requirement is the installation of SSH keys for a git client.
 
+#### Using named private repositories
+
+When working with private repositories it's sometimes required to use different 
+ways to access a repository.
+While your CI system might use SSH to read from your repository a developer 
+needs to access the same repository using another resource or authentication.
+
+You'll need to define a named repository in your `composer.json` file:
+
+```json
+{
+    "require": {
+        "vendor/my-private-repo": "dev-master"
+    },
+    "repositories": {
+        "my-private-repo": {
+            "type": "vcs",
+            "url":  "git@bitbucket.org:vendor/my-private-repo.git"
+        }
+    }
+}
+```
+
+Any developer who cannot access the default repository source can now 
+define an own source for this named repository in his global Composer `config.json`.
+Adding `"immutable": true` to the repository config will override any same named
+repository configuration.
+
+```json
+{
+    "config": {
+    
+    },
+    "repositories": {
+        "my-private-repo": {
+            "type": "vcs",
+            "url":  "https://bitbucket.org:vendor/my-private-repo.git",
+            "immutable": true
+        }
+    }
+}
+```
+
+
 #### Git alternatives
 
 Git is not the only version control system supported by the VCS repository.
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -173,10 +173,12 @@ class Config
                 if (is_int($name)) {
                     $this->repositories[] = $repository;
                 } else {
-                    if ($name === 'packagist') { // BC support for default "packagist" named repo
-                        $this->repositories[$name . '.org'] = $repository;
-                    } else {
-                        $this->repositories[$name] = $repository;
+                    if (!isset($this->repositories[$name]) || !array_key_exists('immutable', $this->repositories[$name]) || !$this->repositories[$name]['immutable']) {
+                        if ($name === 'packagist') { // BC support for default "packagist" named repo
+                            $this->repositories[$name . '.org'] = $repository;
+                        } else {
+                            $this->repositories[$name] = $repository;
+                        }
                     }
                 }
             }
--- a/tests/Composer/Test/ConfigTest.php
+++ b/tests/Composer/Test/ConfigTest.php
@@ -219,6 +219,32 @@ class ConfigTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(array('https', 'git'), $config->get('github-protocols'));
     }
 
+    public function testNamedRepositorySourceWillBeOverruledByLast()
+    {
+        $config = new Config(false);
+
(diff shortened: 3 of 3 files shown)


alcohol:
Just don't use named repositories.

jmglsn:
This would end up with warnings, as some of the repositories are not reachable from all networks.

alcohol:
I am not sure how I feel about this. It seems like a weird edge case. Also, the naming of the flag is incredibly bad. It should instead be more in line with the intent of it, e.g. "priority", and give it a numerical value, so you have more control over it.

Regardless, still not sure how I feel about this feature in general. @Seldaek any comments?

Seldaek:
This is definitely not commonly needed, but also doesn't add a whole lot of code and it allows this somewhat complex situation to be resolved nicer so I'd say why not add it. The naming though I agree could be clearer, I get what you mean by shadowing but without the explanation I doubt I'd have guessed. Priority would be an option but then again not sure if really needed as we only have global and local repos.. it's not like you can have more than two repos in the race.

alcohol:
Suggestions;

- `immutable`
- `fixed`
- `leading`
- `overwrite` (inverse, would be set to `false` instead to achieve same result)

You guys have any other suggestions?

jmglsn:
I was close to name it `fixed`...

alcohol:
I feel like `immutable` is less ambiguous than `fixed` though.

src/Composer/Config.php
```diff
@@ -172,7 +172,9 @@ public function merge($config)
                 if (is_int($name)) {
                     $this->repositories[] = $repository;
                 } else {
-                    $this->repositories[$name] = $repository;
+                    if (!isset($this->repositories[$name]) || !isset($this->repositories[$name]['shadows']) || !$this->repositories[$name]['shadows']) {
```
alcohol: I would rather use `array_key_exists` for the second expression, not `isset`.

alcohol: It's not about benefit, it's about intent. You want to know if the array key exists.

jmglsn: Ok, I'll update this.

alcohol: Thanks :-)

jmglsn:
Sorry for the double commit, too early this day...

alcohol:
Didn't see anything about that. But would still like to see a commit for the documentation :-)

jmglsn:
Is the documentation update ok?

alcohol:
Good enough. Maybe I'll change the wording/phrasing a bit. But it's a good start, thanks!

doc/05-repositories.md
```diff
@@ -264,6 +264,50 @@ GitHub and BitBucket:
...
+    "repositories": {
+        "my-private-repo": {
+            "type": "vcs",
+            "url":  "git@bitbucket.org:vendor/my-private-repo.git"
+        }
+    }
+}
+```
+
+Any developer who cannot access the default repository source can now
```
Seldaek: Actually that's a pretty bad idea. If you define it on a dev machine, *any* project you run `composer update` in will have that global repository included. Depending on what you have in that private repo, it might end up being nasty. It'd be best to rather advice adding it on the CI server.

jmglsn: That's a good point. But what I've seen so far was not only a dependency between CI server and repository URL it was also a dependency in the way how you're connected to the companies network. So meanwhile it would help to add it directly to the repository `composer.json` and do some magic on the CI server, but in some situations it would still not work for the developer. This results in nasty workflows, leaving out the power of composer and ends with bad versioning, broken releases.
Do you think it would be ok to have something like a `composer.config` / `repositories.config`, which is located at repository level but not under SCM? This could act like a global config, but is restricted to the current repository? (It's a little bit like the `docker-compose.override` approach.)

alcohol: You could simple have 2 composer.json configurations in your repo. And then use 
```
COMPOSER=file.json composer <command>
```` 
to specify which one you want to use at run-time.

jmglsn:
@Seldaek rebased the PR in order to solve the merge conflict with master.

jmglsn:
Rebase on current master...

Seldaek:
@jmglsn looking at this again, I don't really think it's a great idea after all, as it breaks the lock file concept. I think a better way to do this would be to write a plugin overriding the URLs depending on some config value. That way it doesn't add more confusing options to composer for specific use cases.. and if you break the lock file you do it on your own terms in a plugin, where we don't have to offer guarantees.

jmglsn:
I agree with that. From my current point of view I also see that problem is generated by the stash toolchain.

PR #5805, opened 2016-10-21
Speed up git checkout by skipping fetch

RichardHeelin:
When composer needs to update a library which is installed as a git repository, it currently performs a fetch to make sure that the git repository is up to date and then issues the checkout. On slow networks this fetch can take 10-20 seconds to complete, slowing down the install of the new commit. This fetch, however, is not always needed, as the new reference commit may already be known by git. By checking if the reference commit exists before issuing the fetch, we can skip the fetch entirely when not needed, resulting in checkouts taking less than a second where the commit is know.

Testing with a simple project, switching between `composer/semver` version 1.4.1 and 1.4.2, has checkouts going from an average of 1.92s over 5 passes to 0.48s.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -105,10 +105,10 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
 
         $ref = $target->getSourceReference();
         $this->io->writeError("    Checking out ".$ref);
-        $command = 'git remote set-url composer %s && git fetch composer && git fetch --tags composer';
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks looks good although I didn't see a huge improvement here it shouldn't hurt anyway.

stof:
@Seldaek the improvement will depend on the size of the new data in the git repo (fetching on a small repo is fast) and whether you already had the commit (so updating to the latest release or latest dev version will generally not benefit from it)

Seldaek:
Yeah I realize that, but I tried it and did not see massive improvements 
in cases where a fetch is not needed vs doing a fetch. Simple statement 
of fact.

Issue #5357, opened 2016-05-19
Default composer init settings


CupOfTea696:
Hi,

It'd be useful to be able to configure default settings for the `composer init` command. For example, when I run `composer init` it always uses `cupoftea696` as a vendor name, but I use `cupoftea` as a vendor name for my composer packages. It'd be useful to set a global config option so that I don't have to customise the package name for each package I create. I imagine it'd be useful for all composer init settings, so that people can set a default license, minimum stability, ...

curry684:
If you have a full template of default settings, why not copy that to `composer.json` instead of using `composer init`? It doesn't seem like a really useful feature to me to build all of this into the init command as to be honest I don't think that many people actually use it - everyone I know just copies a stock file or creates it completely by hand.

CupOfTea696:
I actually use `composer init` quite a lot.

dbrumann:
I might be mistaken, but isn't this already supported?

```
mkdir my-package && cd my-package
composer init --name="acme/my-package"


  Welcome to the Composer config generator



This command will guide you through creating your composer.json config.

Package name (<vendor>/<name>) [acme/my-package]:
```

The option also has various levels of fallback if it's not used: https://github.com/composer/composer/blob/master/src/Composer/Command/InitCommand.php#L191-L204

CupOfTea696:
@dbrumann That still requires you to type it every time you run `composer init`.

jvasseur:
@CupOfTea696 you could create an alias that already contains the wanted default options.

naderman:
How often do you realistically use composer init that typing those few characters justifies adding 40 lines of code and a new config file to this project that will need to be maintained for all eternity?

curry684:
I'd be a lot more open-minded towards some optional env variables, like `COMPOSER_DEFAULT_VENDOR`. Hell we could just for the 3~4 properties involved also expose `COMPOSER_DEFAULT_AUTHOR`, `COMPOSER_DEFAULT_LICENSE` and `COMPOSER_DEFAULT_EMAIL`, and if set skip the question. Lot more practical than a separate config file for reasons given above.

CupOfTea696:
@curry684 that's actually not a bad idea, tho I don't necessarily think we should skip the question? Just change the default suggestion.

curry684:
Fine with me.

zachflower:
[...]

PR #5806, opened 2016-10-21
[WIP] Default composer init settings

CupOfTea696:
This allows setting default values for the `composer init` command, for example setting a default vendor prefix for packages, or a default author email that differs from your git email.

Defaults can be set via the `composer config` command with the `global` option set. 
The following defaults are supported:

``` js
[
   "vendor-prefix",
   "description",
   "author", // Using John Smith <john@example.com> format
   "author.name",
   "author.email",
   "minimum-stability",
   "type",
   "license"
]
```

Example `composer config` usage:

``` bash
composer config -g defaults.vendor-prefix vendor
composer config -g defaults.author John Smith <john@example.com>
composer config -g defaults.author.email john@example.com
```

TODO:
- Currently the `ConfigCommand` uses `InitCommand::parseAuthorString` to parse the author string. The `parseAuthorString` should be extracted to a `trait` or a `utility`, I'm not sure what the recommended approach is. However, creating an instance of `InitCommand` to run a single method is not the right approach, which is what I'm doing now.

Related issue: #5357

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -509,6 +509,66 @@ EOT
             return $this->configSource->addConfigSetting($settingKey, $values[0]);
         }
 
+        // handle defaults
+        $uniqueDefaults = array(
+            'vendor-prefix' => array('is_string', function ($val) { return $val; }),
+            'type' => array('is_string', function ($val) { return $val; }),
+            'description' => array('is_string', function ($val) { return $val; }),
+            'minimum-stability' => array(
+                function ($val) { return isset(BasePackage::$stabilities[VersionParser::normalizeStability($val)]); },
+                function ($val) { return VersionParser::normalizeStability($val); }
+            ),
+        );
+        $multiDefaults = array(
+            'license' => array(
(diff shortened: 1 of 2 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -509,6 +509,66 @@ function ($vals) {
...
+            }
+
+            if ($input->getOption('unset')) {
+                return $this->configSource->removeConfigSetting($settingKey);
+            }
+
+            if ('defaults.author' == $settingKey) {
+                // TODO: Extract parseAuthorString to Trait or Util. Not sure about best approach but this isn't it.
+                $initCmd = new InitCommand();
+                $author = $initCmd->parseAuthorString($values[0]);
```
CupOfTea696: Advice on how best to handle this requested.

CupOfTea696:
@alcohol well I think this would be really useful and it would speed up my process quite a bit.

Seldaek:
IMO this is better done using a bash alias or something if you really create so many packages that you need a template.. Not really worth adding to composer.

PR #5807, opened 2016-10-22
add check for non-strings in VersionParser

gregory-nisbet:
An invalid `composer.json` of the form 

``` json
{
    "name": "gregnisbet/thing",
    "require": {
        "a": ["a"]
    }
}
```

currently results in an error message that's rather difficult to interpret

```
 [ErrorException]
  Illegal offset type in isset or empty
```

I added an explicit check in `parseConstraints` to check that the parameter `$constraints` is a string before attempting to use it as a key in `$self::$constraints`, the error message now reads

```
  [UnexpectedValueException]
  Version constraint must be string.
```

Which is somewhat less than ideal because it doesn't tell you where in your `composer.json` the ill-formed constraint is, but hopefully is more informative.

### DIFF
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -23,6 +23,9 @@ class VersionParser extends SemverVersionParser
      */
     public function parseConstraints($constraints)
     {
+        if (!is_string($constraints)) {
+            throw new \UnexpectedValueException('Version constraint must be string.');
+        }
         if (!isset(self::$constraints[$constraints])) {
             self::$constraints[$constraints] = parent::parseConstraints($constraints);
         }
--- a/tests/Composer/Test/Package/Version/VersionParserTest.php
+++ b/tests/Composer/Test/Package/Version/VersionParserTest.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Test\Package\Version;
+
+
+class VersionParserTest extends \PHPUnit_Framework_TestCase
+{
+    public function testParseConstraintsRejectsArray()
+    {
+        $parser = new \Composer\Package\Version\VersionParser();
+        $invalid_constraint = array();
+        $did_throw_exception = false;
+        $exception_message = null;
+        try {
+            $parser->parseConstraints($invalid_constraint);
+        } catch (\UnexpectedValueException $e) {
+            $did_throw_exception = true;
+            $exception_message = $e->getMessage();
+        }
+        $this->assertEquals($did_throw_exception, true);
+        $this->assertEquals($exception_message, 'Version constraint must be string.');
+    }
+}


stof:
to get detailed info about what is invalid in your composer.json, use the ``composer validate`` command.

gregory-nisbet:
@stof 

In this case, just running `composer` (which I think should just print a usage string) or `composer validate` gives us the same exception.

``` [ErrorException]
  Illegal offset type in isset or empty```

Actually when you run `composer` with no arguments, does it need to parse the `composer.json` in order to figure out what the contents of the usage string should be?

**Comment by alcohol** [comment]
It always parses a local (cwd) `composer.json` if present. Adding a check for this might be ok, but the current implementation is not the right place for this.

We should probably also add a better definition for these properties (they all share the same format): https://github.com/composer/composer/blob/master/res/composer-schema.json#L77-L101

**Comment by alcohol** [comment]
@stof unfortunately running `validate` is not an option in this case because the `createComposer` command that gets called before most commands actually do their job already tries to parse the local configuration:

``` sh
rob ~/Git/composer > bin/composer validate -v

                                         
  [ErrorException]                       
  Illegal offset type in isset or empty  
                                         

Exception trace:
 () at /home/rob/Git/composer/src/Composer/Package/Version/VersionParser.php:26
 Composer\Util\ErrorHandler::handle() at /home/rob/Git/composer/src/Composer/Package/Version/VersionParser.php:26
 Composer\Package\Version\VersionParser->parseConstraints() at /home/rob/Git/composer/src/Composer/Package/Loader/ArrayLoader.php:238
 Composer\Package\Loader\ArrayLoader->parseLinks() at /home/rob/Git/composer/src/Composer/Package/Loader/ArrayLoader.php:124
 Composer\Package\Loader\ArrayLoader->load() at /home/rob/Git/composer/src/Composer/Package/Loader/RootPackageLoader.php:101
 Composer\Package\Loader\RootPackageLoader->load() at /home/rob/Git/composer/src/Composer/Factory.php:348
 Composer\Factory->createComposer() at /home/rob/Git/composer/src/Composer/Factory.php:569
 Composer\Factory::create() at /home/rob/Git/composer/src/Composer/Console/Application.php:314
 Composer\Console\Application->getComposer() at /home/rob/Git/composer/src/Composer/Console/Application.php:426
 Composer\Console\Application->getPluginCommands() at /home/rob/Git/composer/src/Composer/Console/Application.php:131
 Composer\Console\Application->doRun() at /home/rob/Git/composer/vendor/symfony/console/Application.php:123
 Symfony\Component\Console\Application->run() at /home/rob/Git/composer/src/Composer/Console/Application.php:100
 Composer\Console\Application->run() at /home/rob/Git/composer/bin/composer:53
```

Seldaek:
I just made sure it chokes on it in a more descriptive way, which is the best we can do here I'm afraid. I tried a clean fix but it got very nasty quickly.

staabm:
@Seldaek would it make sense to validate the composer.json against a json-schema in a first step?

stof:
@staabm not in the ArrayLoader, as this would slow down the loading of data from Packagist, which also uses the ArrayLoader. This is why we have ``composer validate`` (which performs such validation)

Seldaek:
@staabm we also can't validate everything with the json schema, for example the require key is defined only as an object with additional props allowed, because we can't list all keys. So this will never be good enough.. We just have to catch "common" mistakes explicitly IMO.

stof:
@Seldaek we can actually enhance the schema. I'm opening a PR

Issue #5808, opened 2016-10-23
Can't combine self and outdated options in show


curry684:
I expected `composer show -s --outdated` to show me outdated root packages, instead it gives:

```
PHP Fatal error:  Uncaught TypeError: Argument 4 passed to Composer\Command\ShowCommand::printMeta() must implement interface Composer\Package\PackageInterface, boolean given, called in phar:///opt/composer.phar/src/Composer/Command/ShowCommand.php on line 182 and defined in phar:///opt/composer.phar/src/Composer/Command/ShowCommand.php:461
Stack trace:
#0 phar:///opt/composer.phar/src/Composer/Command/ShowCommand.php(182): Composer\Command\ShowCommand->printMeta(Object(Composer\Package\RootPackage), Array, Object(Composer\Repository\ArrayRepository), false)
#1 phar:///opt/composer.phar/vendor/symfony/console/Command/Command.php(259): Composer\Command\ShowCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#2 phar:///opt/composer.phar/vendor/symfony/console/Application.php(849): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#3 phar:///opt/composer.ph in phar:///opt/composer.phar/src/Composer/Command/ShowCommand.php on line 461

Fatal error: Uncaught TypeError: Argument 4 passed to Composer\Command\ShowCommand::printMeta() must implement interface Composer\Package\PackageInterface, boolean given, called in phar:///opt/composer.phar/src/Composer/Command/ShowCommand.php on line 182 and defined in phar:///opt/composer.phar/src/Composer/Command/ShowCommand.php:461
Stack trace:
#0 phar:///opt/composer.phar/src/Composer/Command/ShowCommand.php(182): Composer\Command\ShowCommand->printMeta(Object(Composer\Package\RootPackage), Array, Object(Composer\Repository\ArrayRepository), false)
#1 phar:///opt/composer.phar/vendor/symfony/console/Command/Command.php(259): Composer\Command\ShowCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
[...]

PR #5812, opened 2016-10-24
Fix crash when listing root package without versions

curry684:
When using certain parameters together in 'composer show', specifically '-s --outdated', the root package itself is inspected for its latest version. If the git repository it belongs to does not have any tags yet this would return FALSE, while the next call to printMeta requires a PackageInterface or NULL, causing a crash.

Refs #5808

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -179,7 +179,7 @@ EOT
                 if ($input->getOption('latest')) {
                     $latestPackage = $this->findLatestPackage($package, $composer, $phpVersion);
                 }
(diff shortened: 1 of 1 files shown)


alcohol:
I don't want to be annoying, but bugfixes should be based off of the 1.2 branch. They will then be merged back from 1.2 into master. Do you think you could move/rebase?

curry684:
Oopsie, been out of the loop a bit   :P   I'll rebase tonight, it's on my home computer.

alcohol:
Probably yea.. :p

curry684:
@alcohol rebased!

PR #5813, opened 2016-10-25
Fix crash when listing root package without versions

curry684:
When using certain parameters together in 'composer show', specifically '-s --outdated', the root package itself is inspected for its latest version. If the git repository it belongs to does not have any tags yet this would return FALSE, while the next call to printMeta requires a PackageInterface or NULL, causing a crash.

Refs #5808


alcohol:
Cheers

PR #5824, opened 2016-10-28
Code simplification

fabpot:
Nothing fancy, but this avoids reading the JSON file twice and simplifies code when possible.


Seldaek:
Thanks

PR #5826, opened 2016-10-29
added more information in the output

fabpot:
This PR adds two pieces of output:
- One line of summary about what Composer is about to install, remove, update (always displayed, not on quiet mode of course);
- Up to 3 lines, in verbose mode only, to list the operations that are going to occur.

The goal is to give some more information upfront about what's going to happen next. A lot of package managers do that and I find it extremely useful for both informational purpose and debugging purpose as well.

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -492,6 +492,38 @@ class Installer
             $devPackages = null;
         }
 
+        if ($operations) {
+            $installs = $updates = $uninstalls = array();
+            foreach ($operations as $operation) {
+                if ($operation instanceof InstallOperation) {
+                    $installs[] = $operation->getPackage()->getPrettyName().':'.$operation->getPackage()->getFullPrettyVersion();
+                } elseif ($operation instanceof UpdateOperation) {
+                    $updates[] = $operation->getTargetPackage()->getPrettyName().':'.$operation->getTargetPackage()->getFullPrettyVersion();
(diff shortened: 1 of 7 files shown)


fabpot:
Which package manager does ask for confirmation? Anyway, the goal is to give the user some information about what's next in terms of wait. When updating a large application, you see package updates one after the other but you have no idea about how many are left. At least, with this, you know the amount of work scheduled by Composer.

alcohol:
Yum, Apt, Pacman (from the top of my head).

curry684:
For OS-level package managers it's more important to request for the confirmation as it may cause system instability, conflicts or downtimes. If you use Composer properly as a dev tool the worst that could happen is breaking your local copy of something temporarily. Also, on apt for example you can skip the confirmation with the `-y` flag, lot less typing than `--no-interaction`.

Middle road would be to only ask for confirmation for possibly unintended or harmful operations (removal/downgrade) and blindly continue if only installs and Semver-compliant upgrades are going to happen.

stof:
@curry684 ``--no-interaction`` can be replaced by its short version too: ``-n``

curry684:
True, always forget that because to me `-n` implies the reverse of `-y`: answer all simple yes/no questions with "no".

Seldaek:
Thanks, and FWIW I agree with @curry684, this doesn't need a confirmation step.. Run --dry-run if you have doubts.

PR #5827, opened 2016-10-29
fixed deprecated class usages

fabpot:
This PR mostly updates phpdoc to reference the non-deprecated event classes. It also fixes the event dispatcher which stills uses a deprecated event class.

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -21,8 +21,7 @@ use Composer\Composer;
 use Composer\DependencyResolver\Operation\OperationInterface;
 use Composer\Repository\CompositeRepository;
 use Composer\Script;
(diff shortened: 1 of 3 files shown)


src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -22,7 +22,7 @@
 use Composer\Repository\CompositeRepository;
 use Composer\Script;
 use Composer\Script\CommandEvent;
```
alcohol: As far as I can tell, you can also remove the `CommandEvent` use statement and its only reference on line 274 (inside a docblock). All deprecated events returned in said method extend from `Event` which is listed as the first return type.

fabpot: done, fixed the phpdoc as well

PR #5829, opened 2016-10-30
Fix POST_DEPENDENCIES_SOLVING trigger

fabpot:
When listening for `POST_DEPENDENCIES_SOLVING`, I was expecting to get all operations that will be executed by Composer, but that's not always the case.

Under some conditions, `processDevPackages` might add some `UpdateOperation` operations, which are not available during the event dispatching as it is dispatched too early.

This PR moves the event dispatch to after the call to `processDevPackages`. It means that it is not part of the `try/catch` block anymore, but throwing a `SolverProblemsException` from the event is not documented and I did not find any such usage on the "Internet".


fabpot:
PR on 1.2 now.

alcohol:
Thanks

Issue #5828, opened 2016-10-30
symfony/process constraint seems to be incorrect


AydinHassan:
`\Composer\Util\ProcessExecutor` uses `Symfony\Component\Process\ProcessUtils` from `symfony/process`. 

Composer's constraint on `symfony/process` is `"^2.1 || ^3.0"` but it seems `ProcessUtils` was only introduced in `2.3`.

Exists: https://github.com/symfony/process/blob/2.3/ProcessUtils.php
Doesn't exist: https://github.com/symfony/process/blob/2.2/ProcessUtils.php

Should we bump the constraint to `2.3` ?

I could be completely off here - noticed while debugging test failure for https://travis-ci.org/Ocramius/PackageVersions/jobs/171779275 (https://github.com/Ocramius/PackageVersions/pull/39) 

That particular test runs `composer update` with `--prefer-lowest --prefer-stable` before running the tests.

Composer stack trace is:

```
Fatal error: Uncaught Error: Class 'Symfony\Component\Process\ProcessUtils' not found in /Users/aydin/projects/PackageVersions/vendor/composer/composer/src/Composer/Util/ProcessExecutor.php on line 124
Error: Class 'Symfony\Component\Process\ProcessUtils' not found in /Users/aydin/projects/PackageVersions/vendor/composer/composer/src/Composer/Util/ProcessExecutor.php on line 124
Call Stack:
0.0004     360728   1. {main}() /Users/aydin/projects/PackageVersions/vendor/composer/composer/bin/composer:0
0.0221    2771032   2. Composer\Console\Application->run() /Users/aydin/projects/PackageVersions/vendor/composer/composer/bin/composer:44
0.0238    3086408   3. Symfony\Component\Console\Application->run() /Users/aydin/projects/PackageVersions/vendor/composer/composer/src/Composer/Console/Application.php:104
0.0244    3163704   4. Composer\Console\Application->doRun() /Users/aydin/projects/PackageVersions/vendor/symfony/console/Symfony/Component/Console/Application.php:124
0.0812    6230760   5. Symfony\Component\Console\Application->doRun() /Users/aydin/projects/PackageVersions/vendor/composer/composer/src/Composer/Console/Application.php:231
0.0813    6230760   6. Symfony\Component\Console\Application->doRunCommand() /Users/aydin/projects/PackageVersions/vendor/symfony/console/Symfony/Component/Console/Application.php:193
[...]

PR #5831, opened 2016-10-31
Bump symfony/process to 2.3 - and raise minimum php to 5.3.3

AydinHassan:
Refs #5828 

I've only updated `composer.json` - let me know if there is something else that should be added.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "issues": "https://github.com/composer/composer/issues"
     },
     "require": {
-        "php": "^5.3.2 || ^7.0",
+        "php": "^5.3.3 || ^7.0",
         "justinrainbow/json-schema": "^1.6 || ^2.0",
         "composer/ca-bundle": "^1.0",
         "composer/semver": "^1.0",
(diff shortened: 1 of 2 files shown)


alcohol:
Should probably update the lock file too (`composer update --lock`)?

alcohol:
Unable to rebase it? If it is too much work for you, we can probably do it. We just appreciate it if you could do it instead, cause it saves us the hassle.

AydinHassan:
@alcohol no it's fine - will do it after work :)

alcohol:
Ok. Was just checking, cause I saw you changed the target but then changed it back :-)

AydinHassan:
I thought it was as simple as that, but I have to rebase locally and fix some conflicts, will probably be quicker to just do it again 😄

alcohol:
Yeah, probably will be. You could use the patch. Simply take the commit diff (2 commits in this case) and apply it on a new branch:

```
git format-patch -2 HEAD --stdout > /tmp/composer.patch
git checkout -b new_branch <remotename>/1.2
git apply - < git.patch
```

curry684:
For 2 lines redoing the change on the proper branch is always faster  😆

AydinHassan:
@alcohol should be done now - let me know if there's anything else

alcohol:
Thanks

Seldaek:
Thanks but re 5.3.3 bump I don't think this matters much as of our users don't install composer via composer so this has little impact. But anyway this shouldn't reflect the minimum version required by our deps but by our own code. Composer takes care of enforcing the minimum version required by the deps.

Issue #5823, opened 2016-10-28
--prefer-dist doesn't respect commit refs


danepowell:
My `composer.json`:

``` json
{
  "minimum-stability": "dev",
  "require": {
    "drupal/simplesamlphp_auth": "8.3.x-dev#283994f",
    "drupal/acquia_lift": "8.3.x-dev#6640cab"
  },
  "repositories": [
    {
      "type": "composer",
      "url": "https://packagist.drupal-composer.org"
    }
  ]
}
```

When I run these commands:

```
composer update
rm -rf vendor
composer install --prefer-dist
```

I get the following output:

```
- Installing drupal/acquia_lift (dev-8.x-3.x 6640cab)
    Loading from cache
- Installing drupal/simplesamlphp_auth (dev-8.x-3.x 283994f)
    Cloning 283994f from cache
```

I would expect that both drupal/acquia_lift and drupal/simplesamlphp_auth are cloned from git, since they both specify a commit ref, and the packaged version for acquia_lift is much newer than the specified commit ref.

A couple notes:
- For simplesamlphp_auth, 283994f actually matches HEAD, and therefore the code in the packaged release. So it seems that the logic for --prefer-dist is actually exactly backwards: it's using the packaged version when it should be cloning from git and vice versa.
- The behavior is slightly different if a local copy of the package already exists. --prefer-dist is a bit "sticky" and will tend to use whatever is already in place, if you already have a source version of a package it will stick to the source version. That's why it's good to remove the vendor directory between tests, as in the testing steps above.

alcohol:
Commit refs are a hack and their behaviour is unspecified. Bugs cannot be filed against them since we do not officially support this "feature".

curry684:
@alcohol I agree with the statement but the [docs do not explicitly say the feature is unsupported](https://getcomposer.org/doc/04-schema.md#package-links), just not advised. I would recommend adding an explicit statement to this extent if policy is not to accept bugs on it.

alcohol:
I will happily accept a docs PR for that.

danepowell:
If commit refs are a hack, what's the proper solution to download a specific Git revision for a package? Is there one? This seems like pretty essential functionality.

stof:
@danepowell stick with versions available properly: branch heads and tags. Packagist does not provide metadata for each commit, and so you cannot require an arbitrary in a clean way (as the solver needs the metadata)

stof:
[...]

PR #5832, opened 2016-10-31
Clarify documentation that commit refs are unsupported

curry684:
Following the discussion in #5823 it has become clear that this feature is supplied at-your-own-risk, this should be more explicitly reflected in its documentation.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -316,12 +316,12 @@ Example:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


alcohol:
Cheers

curry684:
Mmmm are the docs generated from 1.2 or master actually?

Seldaek:
master.. I'll merge 1.2 into it eventually :) But generally doc fixes can go on master as we don't do per-version docs.

PR #5836, opened 2016-11-01
simplified code

fabpot:
The code used to overwrite on the console is needlessly complex. Why not using an ANSI sequence to do the job? Might be an issue on Windows?

Feel free to close if this indeed does not work on Windows.

### DIFF
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -33,10 +33,6 @@ class ConsoleIO extends BaseIO
     protected $output;
     /** @var HelperSet */
     protected $helperSet;
-    /** @var string */
-    protected $lastMessage;
-    /** @var string */
-    protected $lastMessageErr;
 
     /** @var float */
     private $startTime;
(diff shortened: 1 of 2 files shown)


Seldaek:
I think this was for windows compat indeed, but not sure anymore, will have to try it out..

curry684:
As mentioned elsewhere my Windows dev env is mucho borko right now, but I do know that ANSI sequences in the command prompt used to be hell on earth, especially those related to layout and cursor movements. It was [more or less rewritten from scratch](https://blogs.windows.com/buildingapps/2014/10/07/console-improvements-in-the-windows-10-technical-preview/#TGCKYIuDpQcdCBuH.97) for Windows 10 and its subsequent updates, 'fixing' a lot of fundamental architectural issues (Symfony Console is still having [its share of consequences today](https://github.com/symfony/symfony/issues/19520)).

So if you try this out, consider it likely that although this code may be obsolete in current Windows 10 builds that it may likely be required still for 7/8/8.1 or even vanilla Win10 compatibility. I'd therefore recommend marking this PR for 2.0 if we can't find someone able to confirm backwards compatibility.

Seldaek:
OK tested in cmd.exe on Win8.1 and indeed it fails to work. It goes back to beginning of the line, but does not erase it, so if the next line we (over)write is shorter than the previous, you have some leftover at the end. That's why we track the length and fill with spaces to make sure the line is cleared.

Issue #2036, opened 2013-06-25
Add ability use env variables in project composer.json


Koc:
We have smth like this

``` json
    "scripts": {
        "post-install-cmd": [
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile",
            "php app/console cache:clear",
            "php app/console cache:clear -e prod",
            "php app/console assetic:dump -e prod"
        ]
    },
```

but on production we should use `php54` instead of `php`. It would be nice add support something like `{%ENV:PHP_INTERPRETER%} app/console cache:clear`

fgilio:
+1

thiagomorato:
@Seldaek How can I use this? "Composer's config already supports vars in the form of {$name}" I have searched a lot but I couldn't find any reference of it, if you can point me in the right direction it would be much appreciated.

Seldaek:
See https://github.com/composer/composer/blob/1.1/src/Composer/Config.php#L34-L39 for some examples. We also have env var support as $VAR or %VAR% in https://github.com/composer/composer/blob/1.1/src/Composer/Util/Platform.php#L33 but we only do those for paths.

I'm not so sure if it's really useful though, in 3 years 2-3 people were interested, sounds like unneeded cruft to me.

phramz:
I'd like to see some kind of `{$env:FOO}` feature as well. I'm trying to pass a checksum hash to `autoloader-suffix` to get deterministic build-artifacts, which is kind of impossible. Setting a static suffix wouldn't help much, because i'd like to keep the cache-buster for opcache.

Another way (at least for the 'config' section) would be to have another CLI argument to overwrite the config like `-- config autoloader-suffix=FOO` or something`.

Seldaek:
@phramz sounds like you can just do `composer config autoloader-suffix $FOO` before your install?

phramz:
[...]

PR #5837, opened 2016-11-01
add support for environment variables in composer.json issue #2036

yodeler:


src/Composer/Json/JsonFile.php
```diff
@@ -295,4 +296,25 @@ protected static function validateSyntax($json, $file = null)
...
+     *
+     * @return string
+     */
+    public function replaceEnvironmentVariables($json)
+    {
+        preg_match_all('@\{\$env:([A-Za-z0-9_]*)\}@i', $json, $matches);
+        if (!empty($matches[1])) {
+            foreach ($matches[1] as $var) {
+                if (!getenv($var))
+                    throw new \RuntimeException( 'Environment variable "'.$var.'" does not exist.' );
```
Seldaek: I don't think this should throw, having empty env  vars is a valid use case for stuff to be disabled/not there.

src/Composer/Json/JsonFile.php
```diff
@@ -97,7 +97,8 @@ public function read()
         } catch (\Exception $e) {
             throw new \RuntimeException('Could not read '.$this->path."\n\n".$e->getMessage());
         }
-
+        
+        $json = $this->replaceEnvironmentVariables($json);
```
Seldaek: Doing this in JsonFile::read is a bit brute-forceish and possibly even dangerous as this is used to parse all json files composer loads, not only composer.json but even package metadata  from packagist, github api responses, etc.

IMO this should be done in a more granular level or at least enabled only when loading composer.json

alcohol:
Not really sure I understand the rationale behind this. I looked at the original issue, but it seems only one of the last comments hinted at a possible use-case for this. In general, I do not think this should be a concern that Composer has to deal with. My experience is that when you need to have environment variables in your config files, you simply template those config files and generate the static version on the environment with the correct values. I don't think we should add too many bells and whistles to our handling of configuration files / reading json. I'm in favor of keeping things simple; KISS.

stof:
Thus, such feature is useful only for a root package if you want to use env variables in the package metadata themselves as done in the tests. Using it in a shared package would break things, as you would not control the final metadata (as you don't know the env vars).

And this PR is incomplete. If the env variables are used in properties impacting the dependency resolution, the locker must store some data about the env var values in the lock file, to be able to warn about env var mismatches making the lock file outdated (as it would change the config).

This means that the working implementation would require much more complexity than the current PR. I don't think this complexity is worth it

Seldaek:
Yup, rethinking this further, as it was proposed like 3 years ago and is indeed adding a bunch of complexity and possible disasters when third party composer.json are loaded as dependencies, it doesn't seem worth doing. Sorry for wasting your time @ghost, thanks for the implementation nonetheless. Maybe it can be done as a plugin modifying the Config instance or something if you really want it.

PR #5840, opened 2016-11-03
Update min versions for Symfony dependencies

fabpot:
I propose to update the min versions for the Symfony dependencies to at least "modern" versions like 2.3/2.7. I would even suggest to bump everything to 2.7, WDTY?


alcohol:
I personally have no objections. But this PR brings the lock file out of sync, since you did not update it along with the json.

Seldaek:
Yeah I guess at this point we can push for 2.7.. we can always rollback if someone complains as it's not strictly required I guess. I'll update to 2.7 and lock file too.

PR #5843, opened 2016-11-03
Enhance the json schema with validation for map objects

stof:
``additionalProperties`` does not only accept a boolean, but also a schema object, in which case the value of the additional property must validate this schema (``true`` is equivalent to adding a catch-all schema).

I added this extra validation for all places using ``additionalProperties`` except 2 of them:

- ``extra`` because the content is really free-form
- ``repositories`` because the inner content is very complex, and I prefer keeping this in a separate PR (looking at a way to avoid duplicating things in Satis, where we already worked on adding validation for it, and probably submitted to master as it would make bigger changes to the schema)

Refs #5807

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -77,32 +77,44 @@
         "require": {
             "type": "object",
             "description": "This is a hash of package name (keys) and version constraints (values) that are required to run this package.",
-            "additionalProperties": true
+            "additionalProperties": {
+                "type": "string"
+            }
         },
         "replace": {
             "type": "object",
             "description": "This is a hash of package name (keys) and version constraints (values) that can be replaced by this package.",
-            "additionalProperties": true
+            "additionalProperties": {
+                "type": "string"
+            }
         },
         "conflict": {
             "type": "object",
             "description": "This is a hash of package name (keys) and version constraints (values) that conflict with this package.",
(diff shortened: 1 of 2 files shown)


stof:
it violates this schema (see the test I added, which is exactly the JSON given in the original issue). So it should be reported properly in ``composer validate``

alcohol:
Yes but as I described, before validate events get to the validating part, we try to read the JSON file, which results in the error reporting in that issue.

stof:
@alcohol not true. The validate commands runs soon enough to avoid reaching the bug.

Here is the output of ``composer validate`` when running it from my feature branch on the composer.json given in #5807:

```
■ ../composer/bin/composer validate

                                                              
   - require.a : Array value found, but a string is required  
                                                              

```

alcohol:
Weird, recent fixes must have solved that cause I tried it (comment is on the other issue) and it didn't work for me.

stof:
well, if you run ``composer validate`` without my patch, it indeed has the bug, because the JSON schema matches and so it proceeds to the next step of the command, which triggers the bug. But the schema validation happens earlier in ``composer validate``, so it allows to catch it

alcohol:
My bad.

Seldaek:
Thanks

PR #5851, opened 2016-11-04
Make download shell script PSR-2 compatible

szepeviktor:
Avoid `else`


Seldaek:
I can see that this simplifies it but I really don't see what this has to do with PSR-2 :) Anyway thanks.

PR #5852, opened 2016-11-06
Remove usage of echo when executing Composer script

fabpot:
All output go through the IO object, except the output of the script executions, which is output directly via `echo`. This PR fixes this bug.


Seldaek:
Thanks

PR #5857, opened 2016-11-06
Add a file_exists to make sure autoload_static.php exists

paladox:
This will help users who remove the file before uploading, per composer solution.

Makes it easyer just to remove the file.


Seldaek:
That adds an i/o check for everybody for nothing IMO. PHP 5.5 and below are all EOL. If you need to support them still then you should pay the cost, not the users that have upgraded.

PR #5860, opened 2016-11-08
Remove the dep to wget in the install script

dunglas:
This is especially useful when executing this script in a `Dockerfile` (`wget` is usually not installed by default).


alcohol:
LGTM

Seldaek:
The problem is you can't really trust php's SSL support unless you use 5.6+. wget or curl is probably more trustworthy..

dunglas:
You mean it's not secure or it doesn't work?

alcohol:
Not secure.

dunglas:
So we need to wait the end of life of older versions of PHP 😕

Seldaek:
Yeah, feel free to use your version if you have php5.6+, but under that, you shouldn't trust it IMO.

PR #5862, opened 2016-11-12
Feature: get file content

tflori:
To solve https://github.com/composer/packagist/issues/556 we need a function to get the content of a file. This function can also be used for getComposerInformation. And because this is the main difference between drivers (how to get the file) we can than create a common function `getComposerInformation()` to the absract class.

### DIFF
--- a/src/Composer/Repository/Vcs/BitbucketDriver.php
+++ b/src/Composer/Repository/Vcs/BitbucketDriver.php
@@ -54,7 +54,11 @@ abstract class BitbucketDriver extends VcsDriver
         }
 
         if (!isset($this->infoCache[$identifier])) {
-            $composer = parent::getComposerInformation($identifier);
+            if ($this->shouldCache($identifier) && $res = $this->cache->read($identifier)) {
+                return $this->infoCache[$identifier] = JsonFile::parseJson($res);
+            }
+
+            $composer = $this->getBaseComposerInformation($identifier);
 
             // specials for bitbucket
             if (!isset($composer['support']['source'])) {
@@ -100,6 +104,10 @@ abstract class BitbucketDriver extends VcsDriver
             }
 
             $this->infoCache[$identifier] = $composer;
+
+            if ($this->shouldCache($identifier)) {
+                $this->cache->write($identifier, json_encode($composer));
+            }
         }
 
         return $this->infoCache[$identifier];
@@ -114,19 +122,11 @@ abstract class BitbucketDriver extends VcsDriver
             return $this->fallbackDriver->getFileContent($file, $identifier);
         }
 
-        if (preg_match('{[a-f0-9]{40}}i', $identifier) && $res = $this->cache->read($identifier . ':' . $file)) {
-            return $res;
-        }
-
         $resource = $this->getScheme() . '://api.bitbucket.org/1.0/repositories/'
                     . $this->owner . '/' . $this->repository . '/src/' . $identifier . '/' . $file;
         $fileData = JsonFile::parseJson($this->getContents($resource), $resource);
         if (!is_array($fileData) || ! array_key_exists('data', $fileData)) {
             return null;
-        }
-
-        if (preg_match('{[a-f0-9]{40}}i', $identifier)) {
-            $this->cache->write($identifier . ':' . $file, $fileData['data']);
         }
 
         return $fileData['data'];
--- a/src/Composer/Repository/Vcs/BitbucketDriver.php
+++ b/src/Composer/Repository/Vcs/BitbucketDriver.php
@@ -1,8 +1,6 @@
 <?php
 
-
 namespace Composer\Repository\Vcs;
-
 
 use Composer\Cache;
 use Composer\Downloader\TransportException;
@@ -35,7 +33,15 @@ abstract class BitbucketDriver extends VcsDriver
         $this->owner = $match[1];
         $this->repository = $match[2];
         $this->originUrl = 'bitbucket.org';
-        $this->cache = new Cache($this->io, $this->config->get('cache-repo-dir').'/'.$this->originUrl.'/'.$this->owner.'/'.$this->repository);
+        $this->cache = new Cache(
+            $this->io,
+            implode('/', array(
+                $this->config->get('cache-repo-dir'),
+                $this->originUrl,
+                $this->owner,
+                $this->repository
+            ))
+        );
     }
 
     /**
@@ -48,12 +54,17 @@ abstract class BitbucketDriver extends VcsDriver
         }
 
         if (!isset($this->infoCache[$identifier])) {
-
             $composer = parent::getComposerInformation($identifier);
 
             // specials for bitbucket
             if (!isset($composer['support']['source'])) {
-                $label = array_search($identifier, $this->getTags()) ?: array_search($identifier, $this->getBranches()) ?: $identifier;
+                $label = array_search(
+                    $identifier,
+                    $this->getTags()
+                ) ?: array_search(
+                    $identifier,
+                    $this->getBranches()
+                ) ?: $identifier;
 
                 if (array_key_exists($label, $tags = $this->getTags())) {
                     $hash = $tags[$label];
@@ -62,7 +73,12 @@ abstract class BitbucketDriver extends VcsDriver
                 }
 
                 if (! isset($hash)) {
-                    $composer['support']['source'] = sprintf('https://%s/%s/%s/src', $this->originUrl, $this->owner, $this->repository);
+                    $composer['support']['source'] = sprintf(
+                        'https://%s/%s/%s/src',
+                        $this->originUrl,
+                        $this->owner,
+                        $this->repository
+                    );
                 } else {
                     $composer['support']['source'] = sprintf(
                         'https://%s/%s/%s/src/%s/?at=%s',
@@ -75,7 +91,12 @@ abstract class BitbucketDriver extends VcsDriver
                 }
             }
             if (!isset($composer['support']['issues']) && $this->hasIssues) {
-                $composer['support']['issues'] = sprintf('https://%s/%s/%s/issues', $this->originUrl, $this->owner, $this->repository);
+                $composer['support']['issues'] = sprintf(
+                    'https://%s/%s/%s/issues',
+                    $this->originUrl,
+                    $this->owner,
+                    $this->repository
+                );
             }
 
             $this->infoCache[$identifier] = $composer;
@@ -87,7 +108,8 @@ abstract class BitbucketDriver extends VcsDriver
     /**
      * {@inheritdoc}
      */
-    public function getFileContent($file, $identifier) {
+    public function getFileContent($file, $identifier)
+    {
         if ($this->sshDriver) {
             return $this->sshDriver->getFileContent($file, $identifier);
         }
@@ -96,7 +118,8 @@ abstract class BitbucketDriver extends VcsDriver
             return $res;
         }
 
-        $resource = $this->getScheme() . '://api.bitbucket.org/1.0/repositories/'.$this->owner.'/'.$this->repository.'/src/'.$identifier.'/' . $file;
+        $resource = $this->getScheme() . '://api.bitbucket.org/1.0/repositories/'
+                    . $this->owner . '/' . $this->repository . '/src/' . $identifier . '/' . $file;
         $fileData = JsonFile::parseJson($this->getContents($resource), $resource);
         if (!is_array($fileData) || ! array_key_exists('data', $fileData)) {
             return null;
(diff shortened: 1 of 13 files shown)


src/Composer/Repository/Vcs/GitBitbucketDriver.php
```diff
@@ -220,83 +136,30 @@ public static function supports(IOInterface $io, Config $config, $url, $deep = f
...
-    }
-
     /**
      * @param string $url
      */
-    private function setupGitDriver($url)
+    protected function setupSshDriver($url)
     {
-        $this->gitDriver = new GitDriver(
+        $this->sshDriver = new GitDriver(
```
stof: this is not a SSH driver. IT is a fallback driver.

src/Composer/Repository/Vcs/GitBitbucketDriver.php
```diff
@@ -194,12 +109,13 @@ public function getTags()
...
-        if ($this->gitDriver) {
-            return $this->gitDriver->getBranches();
+        if ($this->sshDriver) {
+            return $this->sshDriver->getBranches();
         }
 
         if (null === $this->branches) {
-            $resource =  $this->getScheme() . '://api.bitbucket.org/1.0/repositories/'.$this->owner.'/'.$this->repository.'/branches';
+            $resource =  $this->getScheme() . '://api.bitbucket.org/1.0/repositories/'
+                         . $this->owner . '/' . $this->repository . '/branches';
```
stof: these formatting changes should be reverted.

src/Composer/Repository/Vcs/BitbucketDriver.php
```diff
@@ -0,0 +1,213 @@
...
+    protected $owner;
+    protected $repository;
+    protected $hasIssues;
+    protected $rootIdentifier;
+    protected $tags;
+    protected $branches;
+    protected $infoCache = array();
+
+    /**
+     * @var GitDriver
```
stof: this is wrong. The Hg bitbucket driver will not have a GitDriver as fallback

src/Composer/Repository/Vcs/BitbucketDriver.php
```diff
@@ -0,0 +1,213 @@
...
+     */
+    protected $sshDriver;
+
+    /**
+     * {@inheritDoc}
+     */
+    public function initialize()
+    {
+        preg_match('#^https?://bitbucket\.org/([^/]+)/(.+?)(\.git|/?)$#', $this->url, $match);
```
stof: this regex is wrong if the class is meant to be used by the Hg driver too (and if it is not, the base class is useless)

src/Composer/Repository/Vcs/BitbucketDriver.php
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    /**
+     * Generate an SSH URL
+     *
+     * @return string
+     */
+    protected function generateSshUrl()
+    {
+        return 'git@' . $this->originUrl . ':' . $this->owner.'/'.$this->repository.'.git';
```
stof: does not make sense in the non-git driver

src/Composer/Repository/Vcs/HgBitbucketDriver.php
```diff
@@ -20,39 +20,25 @@
 /**
  * @author Per Bernhardt <plb@webfactory.de>
  */
-class HgBitbucketDriver extends VcsDriver
+class HgBitbucketDriver extends BitbucketDriver
```
stof: this breaks this driver entirely

src/Composer/Repository/Vcs/PerforceDriver.php
```diff
@@ -75,6 +76,22 @@ public function getComposerInformation($identifier)
...
+    {
+        return $this->perforce->getFileContent($file, $identifier);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getChangeDate($identifier)
+    {
+        return new \DateTime();
```
stof: this is broken

stof:
Please also revert all formatting changes you made. They make the review much harder (I skipped most places looking like formatting changes only, but I may have missed non-obvious changes in them; and the code does not match the coding standards used in the project anymore)

src/Composer/Repository/Vcs/BitbucketDriver.php
```diff
@@ -0,0 +1,213 @@
...
+     * @return mixed The result
+     */
+    protected function getContentsWithOAuthCredentials($url, $fetchingRepoData = false)
+    {
+        try {
+            return parent::getContents($url);
+        } catch (TransportException $e) {
+            $bitbucketUtil = new Bitbucket($this->io, $this->config, $this->process, $this->remoteFilesystem);
+
+            switch ($e->getCode()) {
```
stloyd: Why using switch with one statement instead of simple `if`?

src/Composer/Repository/Vcs/BitbucketDriver.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        return 'git@' . $this->originUrl . ':' . $this->owner.'/'.$this->repository.'.git';
+    }
+
+    protected function attemptCloneFallback()
+    {
+        try {
+            $this->setupSshDriver($this->generateSshUrl());
+
+            return;
```
stloyd: This is useless here.

src/Composer/Repository/Vcs/GitLabDriver.php
```diff
@@ -144,6 +158,54 @@ public function getComposerInformation($identifier)
...
+        try {
+            $content = $this->getContents($resource);
+        } catch (TransportException $e) {
+            if ($e->getCode() !== 404) {
+                throw $e;
+            }
+            return null;
+        }
+
+        if (preg_match('{[a-f0-9]{40}}i', $identifier)) {
```
stloyd: This check can be removed cause you done same above, just put first in variable & check value of it.

src/Composer/Repository/Vcs/SvnDriver.php
```diff
@@ -167,6 +170,66 @@ public function getComposerInformation($identifier)
...
+            $path = $identifier;
+            $rev = '';
+        }
+
+        $output = $this->execute('svn info', $this->baseUrl . $path . $rev);
+        foreach ($this->process->splitLines($output) as $line) {
+            if ($line && preg_match('{^Last Changed Date: ([^(]+)}', $line, $match)) {
+                return new \DateTime($match[1], new \DateTimeZone('UTC'));
+            }
+        }
```
stloyd: I guess that you missed `return new \DateTime()` after loop to match behavior of other classes here.

src/Composer/Util/Perforce.php
```diff
@@ -407,6 +409,50 @@ public function getComposerInformation($identifier)
...
+            return $path;
+        } else {
+            $path = substr($identifier, 0, $index) . '/' . $file . substr($identifier, $index);
+            $command = $this->generateP4Command(' files ' . $path, false);
+            $this->executeCommand($command);
+            $result = $this->commandResult;
+            $index2 = strpos($result, 'no such file(s).');
+            if ($index2 === false) {
+                $index3 = strpos($result, 'change');
+                if (!($index3 === false)) {
```
stloyd:
```php
if ($index3 !== false) {

File: src/Composer/Util/Perforce.php [hunk 13]
```diff
@@ -407,6 +409,50 @@ public function getComposerInformation($identifier)
...
+                    $phrase = trim(substr($result, $index3));
+                    $fields = explode(' ', $phrase);
+                    $id = $fields[1];
+                    $path = substr($identifier, 0, $index) . '/' . $file . '@' . $id;
```
stloyd: This variable is useless as you return value in next line, same for `$id`.

tflori:
I'm confused. The contribution guide for composer/composer says that it should be PSR-2. But there are a lot of violations to PSR-2 in the files.

src/Composer/Repository/Vcs/HgBitbucketDriver.php
tflori: I can't confirm. BitbucketDriver extends VcsDriver - so HgBitbucketDriver is still extending VcsDriver.

src/Composer/Repository/Vcs/SvnDriver.php
tflori: I will return null. `PerforceDriver` does not support change date. So it will just not fill `$composer['time']`.

src/Composer/Repository/Vcs/GitDriver.php
```diff
@@ -120,38 +120,38 @@ public function getDist($identifier)
...
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
-    public function getComposerInformation($identifier)
+    public function getFileContent($file, $identifier)
     {
-        if (preg_match('{[a-f0-9]{40}}i', $identifier) && $res = $this->cache->read($identifier)) {
-            $this->infoCache[$identifier] = JsonFile::parseJson($res);
+        if (preg_match('{[a-f0-9]{40}}i', $identifier) && $res = $this->cache->read($identifier . ':' . $file)) {
```
Seldaek: IMO the cache read/write should only be done if $file is composer.json, as the other stuff probably there's no point in caching it, as we won't re-read the README of the revision many times, and storing a bunch of READMEs in the cache is a waste of space. And also the cache key should just be $identifier to avoid wiping our entire cache on packagist.

src/Composer/Repository/Vcs/GitHubDriver.php
```diff
@@ -146,57 +146,79 @@ public function getComposerInformation($identifier)
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getFileContent($file, $identifier)
+    {
+        if ($this->gitDriver) {
+            return $this->gitDriver->getFileContent($file, $identifier);
+        }
+
+        if (preg_match('{[a-f0-9]{40}}i', $identifier) && $res = $this->cache->read($identifier . ':' . $file)) {
```
Seldaek: Same as above regarding cache

src/Composer/Repository/Vcs/GitLabDriver.php
```diff
@@ -144,6 +143,54 @@ public function getComposerInformation($identifier)
...
+    {
+        // Convert the root identifier to a cachable commit id
+        if (!preg_match('{[a-f0-9]{40}}i', $identifier)) {
+            $branches = $this->getBranches();
+            if (isset($branches[$identifier])) {
+                $identifier = $branches[$identifier];
+            }
+        }
+
+        if ($isHash = preg_match('{[a-f0-9]{40}}i', $identifier) && $res = $this->cache->read($identifier . ':' . $file)) {
```
Seldaek: Same here re cache

src/Composer/Util/Perforce.php
```diff
@@ -407,6 +409,47 @@ public function getComposerInformation($identifier)
...
+                if ($index3 !== false) {
+                    $phrase = trim(substr($result, $index3));
+                    $fields = explode(' ', $phrase);
+                    return substr($identifier, 0, $index) . '/' . $file . '@' . $fields[1];
+                }
+            }
+        }
+
+        return null;
+    }
```
Seldaek: Why is this not used by the getComposerInformation btw? Seems like getComposerInformation as well as getComposerInformationFromPath and getComposerInformationFromLabel could be removed and replaced by this implementation?

src/Composer/Repository/Vcs/SvnDriver.php
```diff
@@ -167,6 +167,68 @@ public function getComposerInformation($identifier)
     }
 
     /**
+     * @param string $file
+     * @param string $identifier
+     */
+    public function getFileContent($file, $identifier)
```
Seldaek: Same here curious why getComposerInformation wasn't removed on this driver

src/Composer/Repository/Vcs/PerforceDriver.php
```diff
@@ -75,6 +76,22 @@ public function getComposerInformation($identifier)
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function getFileContent($file, $identifier)
```
Seldaek: And getComposerInformation here again.

Seldaek:
PSR-2 for contributions means the code you add should ideally conform not that you should fix everything when submitting a PR :) Either way changing a bunch of CS stuff together with meaningful changes is never a good idea IMO.

Regarding the rest, I submitted a few comments in line. I am not entirely sure if the getComposerInformation removals were an oversight or just you didn't get to it yet. Anyway aside from that I think it seems to be on the right track, thanks.

src/Composer/Repository/Vcs/HgBitbucketDriver.php
tflori: first of all: it is still working (tested with own mercurial repo on bitbucket.org).

before there was no `fallbackDriver` for `HgBitbucketDriver` - I added this feature. The only other thing that is changing: the BitbucketDriver supports to fill `$composer['support']['issues']` and `$composer['support']['source']`. And this works in `HgBitbucketDriver` too because Bitbucket has the same features for git an mercurial.

If you think this breaks it: can you please provide a test for it? I don't see an error.

stof: The regex in ``initialize`` is *not* the same in both drivers, so you cannot move it to the parent class.

And at the time I commented, the fallback URL was generated in the parent class, using the git one all the time

tflori: the regex got corrected already, is valid for both and works (`/`are replaced by bitbucket to `-`) and the fallback url got corrected too. Is there still something where you see a bug?

tflori:
@Seldaek thanks for your feedback! I changed what you mentioned. Can you have a look again?

Seldaek:
Thanks! I think it looks good now.. let's just hope there's no regression given the large amount of changes/moves it's a bit hard to review in depth :) Anyway merged

PR #5876, opened 2016-11-18
Add documentation for providers option

copitz:
With composer/satis#295 an option for dumping provider includes was added which is missing here in the docs


alcohol:
Thanks

copitz:
Thank you

Issue #5783, opened 2016-10-12
Composer Install generates many php Module already loaded warnings php 7.1


red-smeg:
When I run this command:

composer Install

I get the following output:

> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::installAssets
> Cannot load Zend OPcache - it was already loaded
> PHP Warning:  Module 'mysqlnd' already loaded in Unknown on line 0
> PHP Warning:  Module 'PDO' already loaded in Unknown on line 0
> PHP Warning:  Module 'xml' already loaded in Unknown on line 0
> PHP Warning:  Module 'calendar' already loaded in Unknown on line 0
> PHP Warning:  Module 'ctype' already loaded in Unknown on line 0
> PHP Warning:  Module 'curl' already loaded in Unknown on line 0
> PHP Warning:  Module 'dom' already loaded in Unknown on line 0
> PHP Warning:  Module 'exif' already loaded in Unknown on line 0
> PHP Warning:  Module 'fileinfo' already loaded in Unknown on line 0
> PHP Warning:  Module 'ftp' already loaded in Unknown on line 0
> PHP Warning:  Module 'gd' already loaded in Unknown on line 0
> PHP Warning:  Module 'gettext' already loaded in Unknown on line 0
> PHP Warning:  Module 'iconv' already loaded in Unknown on line 0
> PHP Warning:  Module 'intl' already loaded in Unknown on line 0
> PHP Warning:  Module 'json' already loaded in Unknown on line 0
> PHP Warning:  Module 'mbstring' already loaded in Unknown on line 0
> PHP Warning:  Module 'mcrypt' already loaded in Unknown on line 0
> PHP Warning:  Module 'mysqli' already loaded in Unknown on line 0
> PHP Warning:  Module 'pdo_mysql' already loaded in Unknown on line 0
> PHP Warning:  Module 'Phar' already loaded in Unknown on line 0
> PHP Warning:  Module 'posix' already loaded in Unknown on line 0
> PHP Warning:  Module 'readline' already loaded in Unknown on line 0
> PHP Warning:  Module 'shmop' already loaded in Unknown on line 0
> PHP Warning:  Module 'SimpleXML' already loaded in Unknown on line 0
> PHP Warning:  Module 'sockets' already loaded in Unknown on line 0
> PHP Warning:  Module 'sysvmsg' already loaded in Unknown on line 0
> PHP Warning:  Module 'sysvsem' already loaded in Unknown on line 0
[...]

PR #5882, opened 2016-11-20
Use random name for tmp ini and delete after use

johnstevenson:
Question: Should `php_ini_scanned_files()` use a `function_exists` guard, since it was only introduced in HHVM in September 2014? I have left this out at the moment.

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -19,6 +19,7 @@ use Composer\Downloader\TransportException;
 use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
 use Composer\Util\ConfigValidator;
+use Composer\Util\IniHelper;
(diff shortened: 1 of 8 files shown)


Seldaek:
I wouldn't bother about 2years old HHVM too much.. except for the test failure it has right now on travis, if you can fix that the patch seems good to me otherwise :)

johnstevenson:
@Seldaek Oops, silly mistake. Incidentally, https://github.com/composer/composer/issues/5783 still needs fixing (it is your code!)

Seldaek:
Ah true thanks I'll add that to the 1.3 milestone too. I take [no responsibility](https://github.com/sensiolabs/SensioDistributionBundle/blame/master/Composer/ScriptHandler.php#L396-L410) for that code though :D 

Not sure how we can fix that though except by fixing the SensioDistributionBundle to act different depending on the Composer version? Any other idea? Well if so please post it in that other issue directly I guess.

PR #5888, opened 2016-11-21
use array as default value

alcohol:
Rather than assigning empty array in constructor, simply default to an array on the class attribute. Also add a guard clause in the dumper. This resolves failing tests on php 7.2.0 which no longer accepts non-countable arguments for `count()`.

### DIFF
--- a/src/Composer/Package/BasePackage.php
+++ b/src/Composer/Package/BasePackage.php
@@ -49,12 +49,14 @@ abstract class BasePackage implements PackageInterface
      * @var int
      */
     public $id;
-
+    /** @var string */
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -63,6 +63,7 @@
         }
     },
     "scripts": {
-        "test": "phpunit"
+        "test": "phpunit",
+        "dtest-72": "docker run -v $(pwd):"
```
Seldaek: seems like this doesn't quite belong here :)

alcohol: Eh indeed, not sure how that snuck in.

PR #5907, opened 2016-11-28
Non-zero status code when there are outdated packages (fixes #5901)

### LINKED ISSUE #5901: composer outdated always exits with a 0 status code
Hello,

It would be nice to have an option that treats outdated dependencies as an error and exits with a non-zero status code.
It would help check outdated dependencies with a CI tools.

Rudloff:


alcohol:
I would add it behind a flag, like validate does (either in positive or negative form), offering the user the ability to toggle it on and/orf off.

alcohol:
Please have a look at the scenarios inside https://github.com/composer/composer/tree/master/tests/Composer/Test/Fixtures/installer for ideas on how to write a test for this.

Rudloff:
I'm not sure I can write a test for this using fixtures as it seems it can only run install or update.

alcohol:
Ah, my bad. It has been a while since I wrote one of those. Forgot they hook directly into `run()`, thus indeed limiting their scope to install/update.

Rudloff:
Is there already a test for the outdated command I could expand? I can't find any.

alcohol:
Probably not. Commands are generally poorly tested :-/

src/Composer/Command/OutdatedCommand.php
```diff
@@ -66,6 +67,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if ($input->getArgument('package')) {
             $args['package'] = $input->getArgument('package');
         }
+        if ($input->getOption('strict')) {
```
alcohol: I don't understand the purpose of this conditional and the following assignment. This seems completely unnecessary. Are you sure you need this?

Rudloff: Without this, the `--strict` option is ignored when using `composer outdated --strict` as it is not passed to the show command.
You have a similar logic with `--direct` at line 64.

alcohol: Ah, I overlooked that. I forgot commands call other commands now a days :<

Seldaek:
Thanks, not sure I agree it needs to be behind a --strict flag, but not that big a deal..

PR #5910, opened 2016-11-29
disabling packagist.org did not work

debugteam:
While testing i found that 

"packagist.org": false 

didn't work to disable packagist.

I have read following issue 

https://github.com/composer/composer/issues/3713

afterwards i've testet 

"composer config repositories.packagist false"

did the review on the composer.json and found that 

"packagist": false 

actually disabled packagist.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -662,7 +662,7 @@ You can disable the default Packagist repository by adding this to your
 {
     "repositories": [
         {
-            "packagist.org": false
+            "packagist": false
         }
     ]
 }


alcohol:
Correct, oversight on our part, thank you for fixing :-)

Seldaek:
Actually just an issue of 1.3 not being released yet and our docs being build from master branch.. I did this change recently but yeah was a bit premature for the docs I guess :) Nvm.

alcohol:
Wait, what?

stof:
@Seldaek I suggest changing the website to build the doc from the 1.2 branch instead

Seldaek:
Yeah we could do that but then doc fixes sent to master won't be 
reflected which isn't great either.. Not sure what the best solution is 
(forcing docs to be rebased on 1.x instead of master kinda sucks for 
small contributions IMO). It's not like it's a common issue so maybe we 
can just live with it.

hkdobrev:
@Seldaek Isn't that a breaking change for existing projects which has disabled Packagist with the old configuration and they would need to change `composer.json` when upgrading to Composer 1.3 and would need to sync Composer minor version across the team in order to use either the new configuration or the old one?

Seldaek:
@hkdobrev of course both packagist and packagist.org still work. The docs don't contain both though to avoid confusion.

Issue #5917, opened 2016-11-30
Require lib-openssl fails when php is linked against LibreSSL


nocive:
It seems that composer is not able to detect lib openssl when using LibreSSL.

My `composer.json`:

```json
    "require": {
        "lib-openssl": "*"
    }
```

When running ```composer install``` this happens:
```
  Problem 1
    - namshi/jose 2.2.2 requires lib-openssl * -> the requested linked library openssl has the wrong version installed or is missing from your system, make sure to have the extension providing it.
    - namshi/jose 2.2.2 requires lib-openssl * -> the requested linked library openssl has the wrong version installed or is missing from your system, make sure to have the extension providing it.
    - Installation request for namshi/jose 2.2.2 -> satisfiable by namshi/jose[2.2.2].
```

More debug info
```
# php -v
PHP 7.0.13 (cli) (built: Nov 16 2016 17:49:30) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.13, Copyright (c) 1999-2016, by Zend Technologies
```
```
# composer show --platform
composer-plugin-api 1.1.0     The Composer Plugin API
ext-apcu            5.1.6     The apcu PHP extension
ext-bcmath          7.0.13    The bcmath PHP extension
ext-bz2             7.0.13    The bz2 PHP extension
ext-calendar        7.0.13    The calendar PHP extension
ext-ctype           7.0.13    The ctype PHP extension
ext-date            7.0.13    The date PHP extension
ext-dom             20031129  The dom PHP extension
ext-exif            0         The exif PHP extension (actual version: 1.4 $Id: 8bdc0c8f27c2c9dd1f7551f1f9fe3ab57a06a4b1 $)
ext-fileinfo        1.0.5     The fileinfo PHP extension
ext-filter          7.0.13    The filter PHP extension
ext-ftp             7.0.13    The ftp PHP extension
ext-hash            1.0       The hash PHP extension
ext-iconv           7.0.13    The iconv PHP extension
ext-intl            1.1.0     The intl PHP extension
ext-json            1.4.0     The json PHP extension
ext-libxml          7.0.13    The libxml PHP extension
ext-mbstring        7.0.13    The mbstring PHP extension
ext-mcrypt          7.0.13    The mcrypt PHP extension
ext-memcached       3.0.0b1   The memcached PHP extension
[...]

PR #5918, opened 2016-11-30
tentative fix for #5917

alcohol:


alcohol:
@Seldaek when you're done with SymfonyCon in Berlin, can you let me know what you think about this?

PR #5928, opened 2016-12-02
semver.mwl.be moved to HTTPS

Slamdunk:
http://semver.mwl.be/ produce a 404, it moved to https://semver.mwl.be/


alcohol:
Thanks

Issue #5869, opened 2016-11-16
documentation updates


alcohol:
The docs could use some updates; e.g. : 

- [x] The [xdebug](https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer) section will be rather irrelevant once 1.3 gets released.
- [x] The [memory issues](https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors) section could benefit from details about cPanel fork bomb protection being a possible root-cause as well.

Jean85:
I would like to tackle this! I need some basic info:

 * For the first point, you want to remove the xdebug section and the previous references too entirely?
 * for the cPanel protection, does suffice using your link and a small explanation?

Jean85:
I've created a PR, trying what I think is the right approach. Review & suggestions welcome!

staabm:
> The xdebug section will be rather irrelevant once 1.3 gets released.

may I ask how this issue was resolved with 1.3?

alcohol:
@staabm see #5580 and #5698, kudos to @johnstevenson :-)

staabm:
awesome, thx guys!

PR #5931, opened 2016-12-02
[Documentation] Update troubleshooting article

Jean85:
* remove the xdebug section, since it will be irrelevant from v1.3
 * add cPanel shell fork bomb protection as an other possible root cause of memory errors

This should solve #5869

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -858,10 +858,6 @@ This env var controls the [`discard-changes`](06-config.md#discard-changes) conf
 If set to 1, this env var will make Composer behave as if you passed the
 `--no-interaction` flag to every command. This can be set on build boxes/CI.
 
-### COMPOSER_DISABLE_XDEBUG_WARN
-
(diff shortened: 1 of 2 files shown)


Seldaek:
Looks fine thanks, but yeah I'll wait until post-1.3 release indeed.

Jean85:
Great!
I also want to thank @alcohol for this issue, it lead me to discover this really nice feature of 1.3!

alcohol:
Can you also remove the reference to `COMPOSER_IGNORE_XDEBUG_WARNING` in the docs for CLI (`03-cli.md`) and instead add `COMPOSER_ALLOW_XDEBUG` ?

Jean85:
Reference removed. For `COMPOSER_ALLOW_XDEBUG`, should we reference it? As far as I can see (#5698), it's used only internally, right?

Seldaek:
Thanks, 1.3 is out :)

PR #5932, opened 2016-12-02
Fix wrong Composer package "type" example

markwatney2016:
Details: https://wiki.typo3.org/Composer#TYPO3_Composer_installers


alcohol:
Thanks

Issue #5716, opened 2016-09-26
Create event before VCS clone


jamiel:
Currently there is a PRE_FILE_DOWNLOAD event which gets fired just before composer downloads a file. It would be great to have a similar event before it pulls a remote repository via the VCS downloaders.

PR #5935, opened 2016-12-03
 #5716 - Add new PluginEvents: for pre/post file/vcs downloads

mwheregroup:
As referenced in feature request #5716 this PR enables the FileDownloader to dispatch VCS Events.

- Also post-download events for regular file and VCS downloads were added.

- PreFileDownloadEvent was renamed to FileDownloadEvent for consistency.

Please note that VCS Downloads will **not** trigger file downloads anymore.

### DIFF
--- a/src/Composer/Plugin/FileDownloadEvent.php
+++ b/src/Composer/Plugin/FileDownloadEvent.php
@@ -19,8 +19,9 @@ use Composer\Util\RemoteFilesystem;
  * The pre file download event.
  *
  * @author Nils Adermann <naderman@naderman.de>
+ * @author Mohamed Tahrioui <mohamed.tahrioui@wheregroup.com>
  */
-class PreFileDownloadEvent extends Event
+class FileDownloadEvent extends Event
 {
     /**
      * @var RemoteFilesystem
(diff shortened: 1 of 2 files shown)


src/Composer/Plugin/FileDownloadEvent.php
```diff
@@ -19,8 +19,9 @@
  * The pre file download event.
  *
  * @author Nils Adermann <naderman@naderman.de>
+ * @author Mohamed Tahrioui <mohamed.tahrioui@wheregroup.com>
  */
-class PreFileDownloadEvent extends Event
+class FileDownloadEvent extends Event
```
staabm: This change breaks backwarde compat

mwheregroup: Shouldn't prefixing the Events with temporal information be considered a bad pattern ?
To ensure backwards compatibility it may coexist until its final deprecation.

staabm: I agree that the current filename is not that pretty, but we need to at least provide a "dummy" class or an class_alias to retain BC.

Seldaek:
The point of having the pre- event was to let the URL be modified or otherwise handled.. I don't quite see the point of the post- event. If there is no use case there is no point in renaming classes and deprecating them, etc.

Seldaek:
Closing as no valid use case was provided.

Issue #5937, opened 2016-12-03
Scripts output different than it was for an old composer (1.1.1)


AdrianSkierniewski:
Hi,

I recently updated composer to latest version (1.1.1 => 1.2.3), and I noticed one issue during running scripts. I'm using codeception to run my tests and I've put the command inside scripts to have a handy shortcut and unify it among all repositories.

```json
    "scripts": {
        "test": "./vendor/bin/codecept run"
    },
```

Before upgrading composer everything was displayed properly:

```
composer test
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
> ./vendor/bin/codecept run
Codeception PHP Testing Framework v2.2.6
Powered by PHPUnit 5.7.1 by Sebastian Bergmann and contributors.

Unit Tests (127) --------------------------------------------------------------------------------------------------------------------------------
✔ ContentRepositoryTest: Can_filter_content_files_list (0.10s)
✔ OptionRepositoryTest: It_checks_existence_of_category_when_deleting_an_option (0.04s)
✔ FileRepositoryTest: Can_create_file_with_unique_name_if_file_name_is_already_taken (0.07s)
✔ OptionsServiceTest: Can_get_categories (0.05s)
✔ ContentRepositoryTest: It_does_not_allow_to_delete_active_translation (0.04s)
✔ ContentRepositoryTest: It_checks_for_empty_array_of_files_to_remove (0.06s)
✔ BlockFinderTest: It_finds_correct_block_for_static_pages (0.05s)
✔ ContentRepositoryTest: It_checks_existence_of_parent (0.06s)
```

But after upgrade it looks like it ignores \r or something like that in output:

```
composer test
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
> ./vendor/bin/codecept run
Codeception PHP Testing Framework v2.2.6
Powered by PHPUnit 5.7.1 by Sebastian Bergmann and contributors.


Unit Tests (127) -------------------------------------------------------------

- FileRepositoryTest: Can_create_file_without_author

✔ 
FileRepositoryTest: Can_create_file_without_author
 
(0.10s)


- ContentRepositoryTest: Can_update_content_file

✔ ContentRepositoryTest: Can_update_content_file
 
(0.09s)

- BlockRepositoryTest: Can_create_block_without_author

✔ 
BlockRepositoryTest: Can_create_block_without_author
 
(0.06s)

[...]

Issue #5936, opened 2016-12-03
Composer 1.2.3 wraps scripts output


mhujer:
My `composer.json` (only the relevant part):
```json
    "scripts": {
        "lint": "parallel-lint -j 10 --exclude var --exclude vendor ."
    },
```

When I run this command:

```
composer run lint
```

I get the following output:

```shell
php c:\xampp\php\composer.phar run lint                            
> parallel-lint -j 10 --exclude var --exclude vendor .             
PHP 7.0.9 | 10 parallel jobs                                       
                                                                   
.............                                                      
.......                                                            
..........................                                         
...........                                                        
... 60/99 (60 %)                                                   
....................                                               
...................                      99/99 (100 %)             
                                                                   
                                                                   
Checked 99 files in 1 second                                       
No syntax error found                                              

```

And I expected this to happen:

```
> parallel-lint -j 10 --exclude var --exclude vendor .
PHP 7.0.9 | 10 parallel jobs
............................................................ 60/99 (60 %)
.......................................                      99/99 (100 %)


Checked 99 files in 1 second
No syntax error found
```
Works in 1.2.2 or when I run the command directly without Composer.

(W10 64bit, cmder shell)

Only change related to output between 1.2.2 and 1.2.3 is probably this one: https://github.com/composer/composer/compare/1.2.2...1.2.3#diff-425b09166cec48daca0a5f0dbbb634fd - in the PR https://github.com/composer/composer/pull/5852

AdrianSkierniewski:
It looks like duplicated to #5936

driesvints:
Seeing this as well in 1.2.3 when running composer scripts:

```bash
vagrant@laravelio:/vagrant$ composer test tests/Components/Users/
> phpunit 'tests/Components/Users/'
PHPUnit 5.6.4 by Sebastian Bergmann and contributors.



.
.
.
.
.
.

 6 / 6 (100%)


Time: 2.46 seconds, Memory: 18.00MB


OK (6 tests, 7 assertions)


vagrant@laravelio:/vagrant$
```

alcohol:
Yeah, duplicate of #5933 basically.

Issue #5933, opened 2016-12-03
Script execution fails with version 1.2.3


mwjames:
My `composer.json`:

```json
	},
	"config": {
		"process-timeout": 0
	},
	"scripts": {
		"phpunit": "php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist",
		"unit": [
			"composer dump-autoload",
			"php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist --testsuite=semantic-mediawiki-unit"
		],
		"cs-standalone": [
			"composer validate --no-interaction",
			"vendor/bin/phpcs src/* includes/* tests/* --standard=phpcs.xml --extensions=php -sp"
		],
		"cs": [
			"composer cs-standalone",
			"vendor/bin/phpmd src/,includes/,tests/ text phpmd.xml"
		],
		"ci": [
			"composer phpunit",
			"composer cs"
		]
	}
```

Running `composer phpunit -q` with `Composer version 1.2.2 2016-11-03 17:43:15` works as expected [0] but with `Composer version 1.2.3 2016-12-01 14:33:53` [1], the execution fails.

The use of the `-q` is expected to be applied to the Composer component and not to the script output but with `1.2.3` this behvaiour changed where `-q` is applied to the whole output.

For test purpose I removed `-q` from the caller [2] (`composer phpunit`) which returns an output but it is not formatted correctly (different to version 1.2.2 [0]).

[0] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/180316580
[1] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/180894326
[2] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/180897823

alcohol:
Try changing the order. I think some changes were applied in 1.2.3 that allow arguments to be passed to script commands, `composer --quiet phpunit` should still do the trick though.

alcohol:
Composer script behaviour was changed in https://github.com/composer/composer/commit/103624d4ed7da1719f8204f11f97147108e1e975

alcohol:
@Seldaek any idea if this was intentional? Side effect we did not take into account? Any suggestions or ideas?

Seldaek:
I think this was exactly the point that the inner output should not leak and instead go through the composer IO. Maybe @fabpot can confirm, but this seems to me like you just were abusing a bug and treating it as a feature :)

[...]

PR #5944, opened 2016-12-06
Do not add newlines to output

driesvints:
This should fix the newline bug reported in #5933, #5937 & #5936

I wasn't 100% sure if the `writeError` should also get this flag but I believe it doesn't.

Thanks to @hannesvdvreken for spotting the lines where it happened.

### DIFF
--- a/src/Composer/Util/ProcessExecutor.php
+++ b/src/Composer/Util/ProcessExecutor.php
@@ -108,7 +108,7 @@ class ProcessExecutor
         if (Process::ERR === $type) {
             $this->io->writeError($buffer);
         } else {
-            $this->io->write($buffer);
+            $this->io->write($buffer, false);
         }
     }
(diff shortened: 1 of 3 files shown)


stof:
> I wasn't 100% sure if the writeError should also get this flag bug I believe it doesn't.

it does. We should not add newlines when forwarding the process output, as we don't read it line by line.

driesvints:
@stof ok, added the flag to `writeError` as well.

Seldaek:
Thanks, needs to go on 1.2 though but I'll rebase your fix.

driesvints:
@Seldaek ok, let me know if you want me to update it.

Seldaek:
c.f. 58b27289d05be96fea13fd5443d269368fc06b23 and parent, closing, thanks!

PR #5946, opened 2016-12-06
Simplify composer output to take less lines

Seldaek:
As you can see below, this reformats the output slightly to save vertical space at a little horizontal space cost.

Less blank lines, and dist updates are shown as updates and not remove+install.

I'd be happy to get feedback here, from users perspective but also from people that may be doing wonky things with the output (/cc @dzuelke @fabpot ;) and might see breakage.

> Note that you should click the images to see them full size for a fair comparison..

Before:

image

After:

image

### DIFF
--- a/src/Composer/Downloader/ArchiveDownloader.php
+++ b/src/Composer/Downloader/ArchiveDownloader.php
@@ -28,14 +28,14 @@ abstract class ArchiveDownloader extends FileDownloader
     /**
      * {@inheritDoc}
      */
-    public function download(PackageInterface $package, $path)
+    public function download(PackageInterface $package, $path, $output = true)
     {
         $temporaryDir = $this->config->get('vendor-dir').'/composer/'.substr(md5(uniqid('', true)), 0, 8);
         $retries = 3;
         while ($retries--) {
-            $fileName = parent::download($package, $path);
+            $fileName = parent::download($package, $path, $output);
 
-            $this->io->writeError('    Extracting archive', true, IOInterface::VERBOSE);
+            $this->io->writeError(' Extracting archive', false, IOInterface::VERBOSE);
 
             try {
                 $this->filesystem->ensureDirectoryExists($temporaryDir);
@@ -76,6 +76,7 @@ abstract class ArchiveDownloader extends FileDownloader
 
                 // retry downloading if we have an invalid zip file
                 if ($retries && $e instanceof \UnexpectedValueException && class_exists('ZipArchive') && $e->getCode() === \ZipArchive::ER_NOZIP) {
+                    $this->io->writeError('');
                     if ($this->io->isDebug()) {
                         $this->io->writeError('    Invalid zip file ('.$e->getMessage().'), retrying...');
                     } else {
@@ -91,7 +92,9 @@ abstract class ArchiveDownloader extends FileDownloader
             break;
         }
 
-        $this->io->writeError('');
+        if ($output) {
+            $this->io->writeError('');
+        }
     }
 
     /**
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -77,13 +77,15 @@ class FileDownloader implements DownloaderInterface
     /**
      * {@inheritDoc}
      */
-    public function download(PackageInterface $package, $path)
+    public function download(PackageInterface $package, $path, $output = true)
     {
         if (!$package->getDistUrl()) {
             throw new \InvalidArgumentException('The given package is missing url information');
         }
 
-        $this->io->writeError("  - Installing <info>" . $package->getName() . "</info> (<comment>" . $package->getFullPrettyVersion() . "</comment>)");
+        if ($output) {
+            $this->io->writeError("  - Installing <info>" . $package->getName() . "</info> (<comment>" . $package->getFullPrettyVersion() . "</comment>)", false);
+        }
 
         $urls = $package->getDistUrls();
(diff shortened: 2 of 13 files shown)


hkdobrev:
@Seldaek I'm not very good on UI, but here are some suggestions, let me know what you think:

1. Could we have the information lines in plain white instead of green? I'm referring to the green lines at the top and bottom of the output.
2. We could then have `Updating`, `Removing` and `Installing` at the beginning of each line without an indentation and a hyphen, but just coloured in yellow.
3. Do you think we could shorten the repetitive word at the beginning, so instead of `Updating`, `Removing`, `Installing` we could have an icon or something else which is shorter?
4. What do you think about writing the text at the end of the line "Loading from cache", "Checking out ..." etc. just during execution and when the operation is done to overwrite with a checkmark or just remove it?

avant1:
@hkdobrev 
> 4. What do you think about writing the text at the end of the line "Loading from cache", "Checking out ..." etc. just during execution and when the operation is done to overwrite with a checkmark or just remove it?

IMO this information should be available after command finished too.
Consider scenario when you use composer on CI server, usually it uses composer cache during `install` and it is fast. But then something goes wrong with cache, built time increases, and you have no idea from build logs why `composer install` now takes 3 times longer.
Also there may be other useful information, like messages that `git` is not installed and instead of fetch and checkout dist archive will be downloaded.

And just to add my two cents, git commit hashes can be shorten to first 7-10 characters (7 is [default git value for short hashes](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1)).
This will allow to avoid line breaks on small terminals.

hkdobrev:
@avant1 Thanks for this! I completely agree and shortening Git hashes would reduce the problem I was trying to solve significantly there.

Could we use some of the functionality in Git 2.11 to get the shortest possible hash which guarantees uniqueness? https://github.com/blog/2288-git-2-11-has-been-released#abbreviated-sha-1-names

@avant1 Also what do you think of the other points?

@Seldaek Let me know if this is out of scope of this PR as I think the changes you've made are already a very welcome improvement and things could be improved further in separate PRs.

stof:
@Seldaek can you also show the output for the verbose mode, which displays more info ?

lunetics:
+1 this one looks good. Also helps a lot to keep it compact e.g. for c.i.

Seldaek:
@hkdobrev here answered:

1. that would change more than needed IMO, and when using higher verbosity modes it's quite useful to have the actual lines of output in green and debug white, plus they're easily visually ignored I feel. Anyway it wasn't really the focus of this PR :)
2. That's also kinda bikeshedding.. I don't think it's needed to change existing conventions as people are used to them.
3. Don't really like that because it's less obvious than words are, and doesn't save that much space. Beginners are already confused enough without us adding cryptic characters to the output :)
4. Agreed with what @avant1 said here. I will shorten the git hashes to 10chars if the output is not verbose, which is probably good enough to avoid clashes on most repos. You can usually find the full hashes in your composer.lock diff if really needed.

@stof verbose output is basically just like it used to be.. with multiline output for the additional info like commit logs when updating vcs repos etc. I don't have a screenshot at hand but you can try it out once merged.

emodric:
It's probably too late now, but personally, this compact output is much less readable than before and degraded UX considerably. I find myself strugling to read the output and understand what the changes in packages are.

Now you basically get one wide blob of text without any whitespace. Compare it to the book without paragraphs if you will, much more difficult to understand.

Is there a switch to get the old output back?

Seldaek:
@emodric I find that surprising because basically we mostly removed duplicate info and some whitespace indeed but it's not like blank lines adds much more context than a newline when every item is made of one line anyway. Do you have colors enabled?

emodric:
Sure, I have colors enabled.

Well, first of all, now I find myself scanning all the time from left to right to see details about a specific package (action, package name, version info, status), instead of just glancing at all the info at once to know exactly what happened. And when you try to read details of 5 or 10 packages in a row, things become messy quick. Removal of ALL whitespace doesn't help either.

I never had problems with duplicate info, it doesn't hurt to have some duplicate output if it helps readability and UX. It's not like it was lying on what's going on.

Seldaek:
I just wonder how much of this is your brain going WTF because what you 
were used to changed. Maybe give it some time and see if you learn to 
scan the new output as well as the old one?

Or do you see a way to improve the new one without returning it to its 
ultra-verbose glory? :)

emodric:
How about something in the middle between the old and new one?

An empty line after each package and moving the last status ("Loading from cache") to new line?

It would reduce the output a bit in comparison to the old one while maintaining readability, and would not have duplicate info.

In the meantime, I'll see if I can adapt :)

hkdobrev:
> An empty line after each package and moving the last status ("Loading from cache") to new line?

The message about using cache is useful when you are debugging a problem with caching (e.g. slow installation in CI due to wrong cache configuration). However, it's not something which you need so explicitly as to be on its own line when you are doing a regular install (let's say a co-worker has updated the lock file).

I could agree with the empty line (even though it's not that useful), but no need for multi-line message per package just to state it will use the cache or not.

emodric:
> but no need for multi-line message per package just to state it will use the cache or not.

I can't agree with that. That is the exact reason why it should go to a new line, because it's not related directly to information that is precending it. The main info is the action, package name and versions, status message is the additional info that can go on second line.

Seldaek:
@dzuelke valid points but are those real issues in practice though?

dzuelke:
I'll tell you soon ;)

PR #5954, opened 2016-12-08
Not actual parameters

Golodnyi:


src/Composer/Repository/CompositeRepository.php
```diff
@@ -102,7 +102,7 @@ public function search($query, $mode = 0, $type = null)
         $matches = array();
         foreach ($this->repositories as $repository) {
             /* @var $repository RepositoryInterface */
-            $matches[] = $repository->search($query, $mode, $type);
```
alcohol: While not specified in the `RepositoryInterface`, the `ComposerRepository` implementation does accept a type argument. So I would exclude this change from the PR.

Golodnyi: It is true, I made changes in my fork

Golodnyi:
You are satisfied with the PR or create a new, without one a change?

alcohol:
Aren't you forgetting something..? :)

Golodnyi:
I do not know...

Seldaek:
Thanks

alcohol:
@Seldaek But the tests were failing.. :-/

Golodnyi:
Time: 29.86 seconds, Memory: 57.00MB
OK (1051 tests, 2116 assertions)

¯\_(ツ)_/¯

alcohol:
It will not run the tests on your machine if you do not have perforce installed (they will be marked as skipped). Check the travis build results. Anyway, I think @Seldaek fixed them manually on master branch already.

PR #5962, opened 2016-12-13
Fix various path handling inconsistencies

AnrDaemon:
+ Normalize paths early.
+ Do not realpath() paths that are already normalized.
+ Correctly treat UNC paths.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -106,12 +106,12 @@ class AutoloadGenerator
         $vendorPath = $filesystem->normalizePath(realpath(realpath($config->get('vendor-dir'))));
         $useGlobalIncludePath = (bool) $config->get('use-include-path');
         $prependAutoloader = $config->get('prepend-autoloader') === false ? 'false' : 'true';
-        $targetDir = $vendorPath.'/'.$targetDir;
+        $targetDir = $filesystem->normalizePath(realpath(realpath($vendorPath.'/'.$targetDir)));
         $filesystem->ensureDirectoryExists($targetDir);
 
-        $vendorPathCode = $filesystem->findShortestPathCode(realpath($targetDir), $vendorPath, true);
+        $vendorPathCode = $filesystem->findShortestPathCode($targetDir, $vendorPath, true);
         $vendorPathCode52 = str_replace('__DIR__', 'dirname(__FILE__)', $vendorPathCode);
-        $vendorPathToTargetDirCode = $filesystem->findShortestPathCode($vendorPath, realpath($targetDir), true);
+        $vendorPathToTargetDirCode = $filesystem->findShortestPathCode($vendorPath, $targetDir, true);
 
         $appBaseDirCode = $filesystem->findShortestPathCode($vendorPath, $basePath, true);
         $appBaseDirCode = str_replace('__DIR__', '$vendorDir', $appBaseDirCode);
@@ -148,7 +148,7 @@ EOF;
         foreach ($autoloads['psr-0'] as $namespace => $paths) {
             $exportedPaths = array();
             foreach ($paths as $path) {
-                $exportedPaths[] = $this->getPathCode($filesystem, $basePath, $vendorPath, $path);
+                $exportedPaths[] = $this->getPathCode($filesystem, $basePath, $vendorPath, $filesystem->normalizePath($path));
             }
             $exportedPrefix = var_export($namespace, true);
             $namespacesFile .= "    $exportedPrefix => ";
@@ -160,7 +160,7 @@ EOF;
         foreach ($autoloads['psr-4'] as $namespace => $paths) {
             $exportedPaths = array();
             foreach ($paths as $path) {
-                $exportedPaths[] = $this->getPathCode($filesystem, $basePath, $vendorPath, $path);
+                $exportedPaths[] = $this->getPathCode($filesystem, $basePath, $vendorPath, $filesystem->normalizePath($path));
             }
             $exportedPrefix = var_export($namespace, true);
             $psr4File .= "    $exportedPrefix => ";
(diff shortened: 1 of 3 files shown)


alcohol:
...tests?

AnrDaemon:
I have a feeling, that the tests are not intended to success.
According to my limited inderstanding, the generated code is correct.
It's the tests are set up wrong.

The very first one,
Here's its setup:
```php
$this->vendorDir = $this->workingDir.DIRECTORY_SEPARATOR.'composer-test-autoload';
```
And here's generated code:
````php
<?php

// autoload_namespaces.php @generated by Composer

$vendorDir = dirname(__FILE__) . '/composer-test-autoload';

return array(
    'Main' => array($baseDir . '/src'),
    'Lala' => array($baseDir . '/src', $baseDir . '/lib'),
);
````

Is 100% correct. According to setup. But the fixture for it is…
```php
<?php

// autoload_namespaces.php @generated by Composer

return array(
    'Main' => array($baseDir . '/src'),
    'Lala' => array($baseDir . '/src', $baseDir . '/lib'),
);
```
…which is completely bogus.

Et cetera.

stof:
well, as the autoload namespace is generated inside ``<vendor dir>/Composer/autoload_namespaces.php``, the path to the vendor dir will **always** be ``dirname(dirname(__FILE__))``, whatever the name of the vendor dir is.

So the fixtures are not buggy at all. Your generated code is

src/Composer/Util/Filesystem.php
```diff
@@ -444,8 +444,8 @@ public function normalizePath($path)
         $prefix = '';
         $absolute = false;
 
-        // extract a prefix being a protocol://, protocol:, protocol://drive: or simply drive:
-        if (preg_match('{^( [0-9a-z]{2,}+: (?: // (?: [a-z]: )? )? | [a-z]: )}ix', $path, $match)) {
+        // extract a prefix being a protocol://host, protocol:, protocol://drive: or simply drive:
+        if (preg_match('{^( (?P<scheme> [a-z][0-9a-z]*: )? (?: // (?P<host>[^/]+) )? )}ix', $path, $match)) {
```
stof: this breaks things, as the code does not use named groups but positional groups when reading matches.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -106,12 +106,12 @@ public function dump(Config $config, InstalledRepositoryInterface $localRepo, Pa
         $vendorPath = $filesystem->normalizePath(realpath(realpath($config->get('vendor-dir'))));
         $useGlobalIncludePath = (bool) $config->get('use-include-path');
         $prependAutoloader = $config->get('prepend-autoloader') === false ? 'false' : 'true';
-        $targetDir = $vendorPath.'/'.$targetDir;
+        $targetDir = $filesystem->normalizePath(realpath(realpath($vendorPath.'/'.$targetDir)));
```
stof: realpath twice does not make sense

src/Composer/Util/Filesystem.php
AnrDaemon: It doesn't break anything, as the code doesn't use named groups at all.
But it fixes broken behavior when protocol-less URI's are treated as local absolute paths.

src/Composer/Autoload/AutoloadGenerator.php
AnrDaemon: Do read the code before commenting.  There's a link to the open PHP bug.

AnrDaemon:
> as the autoload namespace is generated inside  `<vendor dir>/Composer/`

Autoload namespaces are generated in `%TEMP%/composer-test-#somehash#`
If you had any different intention, you should have set up the test differently.

The code generated by tests adds up nicely, If I …
```php
<?php

require 'composer-test-autoload/autoload.php';

$cmm = new Lala\ClassMapMain();
$cmmt = new Lala\Test\ClassMapMainTest();
```
…it runs just fine and finding all the right classes.

src/Composer/Util/Filesystem.php
AnrDaemon: To clarify: Named captures are used only as a comment, a reminder of purpose for each part, the only capture the following code is using is the outer unnamed `\1` capture.

src/Composer/Autoload/AutoloadGenerator.php
alcohol: While you are correct, there is no need to be rude at the same time. Also if you read the code, you would notice that originally it did not use double `realpath`. Now unless you you wrote the code originally (which you did not), how can you be 100% certain that this was not intentionally left using a single `realpath` call? Changing behavior requires justification generally, which you have not really provided.

alcohol:
> I have a feeling, that the tests are not intended to success.
According to my limited inderstanding, the generated code is correct.
It's the tests are set up wrong.

Either you are being stubborn, or you are simply clueless when it comes to testing. The former I have no response to, but in the latter case, I can understand that and give you some clarification:

The tests are not wrong. The tests are simply telling us if we meet the expectations we have set or not. If that expectation has changed (a bug was found, desired outcome has changed, etc), then you should adjust the tests accordingly. After that, you modify the code to meet the new expectations.

You only changed the code, not the expectations.

AnrDaemon:
If you look closely, all I've changed is path formats, so that it always have same formatting where it is trying to compare them. All same variables are used in same order.
The amount of realpath() calls must not matter, and the 
```
realpath(/path) === realpath(realpath(/path))
```
invariant indeed holds true… under *NIX (this includes Cygwin).
Under Windows, due to questionable decisions of that function's developer, it doesn't expand all path components on a first try in certain cases.
Related PHP bugs:
 + https://bugs.php.net/bug.php?id=72642
 + https://bugs.php.net/bug.php?id=72738

AnrDaemon:
I will look into tests once I have a moment off.

Seldaek:
Added a commit that should fix the test suite and any regressions introduced. Please check if it still works for you with UNC paths. I don't have an easy way to test this here.

AnrDaemon:
In a `Composer\Test\Util\FilesystemTest::testFindShortestPathCode()`, the source data set
```
            array('C:/Temp/../..', 'd:\Temp\..\..\test', true, "'d:/test'"),
```
is highly questionable, IMO. If someone is trying to escape chroot via compowser, I don't want to be part of this attempt.

alcohol:
There is no such thing as a `chroot` for composer. They are merely relative paths, and they are allowed and supported.

AnrDaemon:
Relative path that escapes its root is a nonsense. Nonsenses aside, this is exactly how you escape chroot - force relative path to break above chroot's root directory.

AnrDaemon:
I have a principal question.
Should I fix Filesystem::normalizePath() The Right Way, or leave it in half-broken state it is in now?

alcohol:
It's not broken. We have no `chroot`. Relative paths outside of root are allowed and totally acceptable since there are valid use-cases for it. Some of those are better solved using the `path` repository type now a days. But regardless, it is not as black and white as you seem to perceive it. So no, the right way is the way things are right now. If it ain't broken, don't fix it :-)

Seldaek:
@AnrDaemon The question I still have is: Is this PR fixing UNC for you? Is it fixing anything at all for you? If yes I think we can merge it, otherwise I'd rather close it and keep things unchanged.

AnrDaemon:
I'll close this PR for now.
While diging for my issue, I stumbled upon some other that not directly related to it, are easily fixable and best fixed before it to not introduce too much instability to the system at once.

PR #5972, opened 2016-12-16
Add a public setter for package name

jreinke:
We have extended Satis in order to generate static virtual packages from real packages but we don't have a way to change virtual package name (after a clone) and this is quite dirty to use a class override for this. Maybe you could add this simple setter.

BTW, thanks for the great software.

### DIFF
--- a/src/Composer/Package/BasePackage.php
+++ b/src/Composer/Package/BasePackage.php
@@ -65,9 +65,17 @@ abstract class BasePackage implements PackageInterface
      */
     public function __construct($name)
     {
+        $this->setName($name);
+        $this->id = -1;
+    }
+
+    /**
+     * @param string $name
+     */
+    public function setName($name)
+    {
         $this->prettyName = $name;
         $this->name = strtolower($name);
-        $this->id = -1;
     }
 
     /**


alcohol:
> Maybe you could add this simple getter.

Typo?

But ehm, use-case is not clear to me. Elaborate please?

jreinke:
Yes this is setter of course, sorry.

I'll try to explain clearly the use case.

We have developed a classic PHP library that we need to distribute privately. This package cannot be distributed entirely to everyone. Some directories and classes have to be removed.

So we have extended Satis in order to allow extra config and to generate some virtual packages that we filter afterwards. For example `vendor/package-name` will generate some _virtual packages_ called `vendor/package-name-foo`, `vendor/package-name-bar` thanks to extra config below (from `satis.json` file):
```json
"extra": {
    "clone-packages": {
        "vendor/package-name": {
            "whitelist": [
                "README.md",
                "src/Vendor/Core/"
                "src/Vendor/Common/"
            ],
            "packages": {
                "vendor/package-name-foo": {
                    "whitelist": [
                        "src/Vendor/Foo/"
                    ]
                },
                "vendor/package-name-bar": {
                    "whitelist": [
                        "src/Vendor/Bar/"
                    ],
                    "blacklist": [
                        "src/Vendor/Bar/Baz.php"
                    ]
                }
            }
        }
    }
}
```
To achieve this, I had to override the Satis `build` command in order to clone the real package and rename it. This is what I did here: https://github.com/jreinke/satis/blob/master/src/Console/Command/BuildCommand.php#L336

I hope it's clearer now.

Seldaek:
I think we rather keep package names immutable because they really are not supposed to change.. You could do something like the AliasPackage does though and just wrap the package object, or use reflection to set the private property if you must, it's safe to assume the property name for the name won't change..

PR #5974, opened 2016-12-17
Allow use of parent directories in PathDownloader when not mirroring the path

cweagans:
I'm working on my composer plugin, and I want to be able to test the plugin with actual composer.json files. In order to do this, I want to use a composer.json like this: https://github.com/cweagans/composer-patches/blob/acceptance-tests/tests/acceptance/fixtures/_fixture-template/composer.json

Currently, `composer install` in that directory yields

```
                                                                                                                       
  [RuntimeException]                                                                                                   
  Package cweagans/composer-patches cannot install to "/home/cweagans/Code/composer-patches/tests/acceptance/fixtures  
  /_fixture-template/vendor/cweagans/composer-patches" inside its source at "/home/cweagans/Code/composer-patches"     
                                                                                                                      
```

I think this PR should do what I need. If you have a different way you want to handle this, I'm all ears. Let me know and I'll update the PR.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -45,13 +45,6 @@ class PathDownloader extends FileDownloader implements VcsCapableDownloaderInter
             ));
         }
 
-        if (strpos(realpath($path) . DIRECTORY_SEPARATOR, $realUrl . DIRECTORY_SEPARATOR) === 0) {
-            throw new \RuntimeException(sprintf(
-                'Package %s cannot install to "%s" inside its source at "%s"',
-                $package->getName(), realpath($path), $realUrl
-            ));
-        }
-
         // Get the transport options with default values
         $transportOptions = $package->getTransportOptions() + array('symlink' => null);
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Looks fine to me, as mirroring should be safe although it's a bit weird pattern. @curry684 anything to add here?

curry684:
The reason for the check is that I think it's a pretty incredibly bad idea to do this. If the package is uninstalled for any reason, like a bug during development, it will/may/could/should effectively `rm -rf /home/cweagans/Code/composer-patches`. I'm not sure that's what can/may/should happen right now, but it may one day be changed to that, without anyone thinking of the scenario you're enabling right here.

I would recommend a different test setup  😉

Seldaek:
Well in this patch it only does it for mirrored so it wouldn't result in wiping the parent dir accidentally through a symlink if I understand your concern correctly.

curry684:
Hmmm yeah I'm misreading something - never liked that change with the STRATEGY constants. Just reviewed the entire file and the change should be safe indeed. Still a bad idea methinks, but safe and no pressing reason not to allow it.

Seldaek:
OK just wanted to be sure I didn't miss anything, thanks :)

cweagans:
Thanks guys! Looking forward to 1.3 so I can start using this :)

curry684:
You know you can just use the snapshot channel right?  ;)

cweagans:
Definitely. I'm using that locally, but I don't want to tag a release that depends on this for testing until Composer 1.3 is out.

cweagans:
This should probably be reverted. :( Turns out that even when mirroring, there are issues with infinite recursion. I'll just have to find a different way to test my plugin.

Seldaek:
Hehe ok, I'll revert thanks for the update.

curry684:
That was another reason I built the check originally hehe. I'm paranoid on stuff like that, so never tested or verified whether it would actually happen, just prevented it before it could.

PR #5979, opened 2016-12-19
Extend travis matrix

alcohol:


Seldaek:
What's the benefit of this? I doubt HHVM acts very differently according to this flag, and until we support php7 only it is probably just additional build overhead no?

alcohol:
Mostly it was just to see if it would already reveal the error reported in the issue I linked. Unfortunately, it did not.

Seldaek:
Ok so rather close for now I think, and have some pity for those travis workers :p

stof:
@alcohol the issue is that HHVM on Precise is very old, and does not have the PHP 7 mode yet. You need to force this job to run on the Trusty environment (see the Symfony config file)

PR #5985, opened 2016-12-21
Cleanup tests

AnrDaemon:
+ Automate teardown for simple cases of "gimme a temp directory for work".
+ Consequently increase the speed of Test/Util/FilesystemTest tenfold.
+ Ensure git tests not run when git is not found.
+ Fix Test/Repository/PathRepositoryTest assumptions being wrong due to multiple PHP bugs.

### DIFF
--- a/tests/Composer/Test/Downloader/GitDownloaderTest.php
+++ b/tests/Composer/Test/Downloader/GitDownloaderTest.php
@@ -17,6 +17,7 @@ use Composer\Config;
 use Composer\TestCase;
 use Composer\Util\Filesystem;
 use Composer\Util\Platform;
+use Symfony\Component\Process\ExecutableFinder;
 
 class GitDownloaderTest extends TestCase
 {
@@ -80,6 +81,11 @@ class GitDownloaderTest extends TestCase
 
     public function testDownload()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
+        }
+
         $packageMock = $this->getMock('Composer\Package\PackageInterface');
         $packageMock->expects($this->any())
             ->method('getSourceReference')
@@ -130,6 +136,11 @@ class GitDownloaderTest extends TestCase
 
     public function testDownloadWithCache()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
+        }
+
         $packageMock = $this->getMock('Composer\Package\PackageInterface');
         $packageMock->expects($this->any())
             ->method('getSourceReference')
@@ -195,6 +206,11 @@ class GitDownloaderTest extends TestCase
 
     public function testDownloadUsesVariousProtocolsAndSetsPushUrlForGithub()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
+        }
+
         $packageMock = $this->getMock('Composer\Package\PackageInterface');
         $packageMock->expects($this->any())
             ->method('getSourceReference')
@@ -278,6 +294,11 @@ class GitDownloaderTest extends TestCase
      */
     public function testDownloadAndSetPushUrlUseCustomVariousProtocolsForGithub($protocols, $url, $pushUrl)
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
+        }
+
         $packageMock = $this->getMock('Composer\Package\PackageInterface');
         $packageMock->expects($this->any())
             ->method('getSourceReference')
@@ -329,6 +350,11 @@ class GitDownloaderTest extends TestCase
      */
     public function testDownloadThrowsRuntimeExceptionIfGitCommandFails()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
+        }
+
         $expectedGitCommand = $this->winCompat("git clone --no-checkout 'https://example.com/composer/composer' 'composerPath' && cd 'composerPath' && git remote add composer 'https://example.com/composer/composer' && git fetch composer");
         $packageMock = $this->getMock('Composer\Package\PackageInterface');
         $packageMock->expects($this->any())
@@ -623,5 +649,19 @@ composer https://github.com/old/url (push)
         }
 
         return $cmd;
+    }
+
+    /**
+     * Check whether or not the given process is available.
+     *
+     * @param string $process The name of the binary to test.
+     *
+     * @return bool True if the process is available, false otherwise.
+     */
+    protected function isProcessAvailable($process)
+    {
+        $finder = new ExecutableFinder();
+
+        return (bool) $finder->find($process);
     }
 }
--- a/tests/Composer/Test/Downloader/GitDownloaderTest.php
+++ b/tests/Composer/Test/Downloader/GitDownloaderTest.php
@@ -650,18 +650,4 @@ composer https://github.com/old/url (push)
 
         return $cmd;
     }
-
-    /**
-     * Check whether or not the given process is available.
-     *
-     * @param string $process The name of the binary to test.
-     *
-     * @return bool True if the process is available, false otherwise.
-     */
-    protected function isProcessAvailable($process)
-    {
-        $finder = new ExecutableFinder();
-
-        return (bool) $finder->find($process);
-    }
 }
--- a/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
+++ b/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
@@ -15,6 +15,7 @@ namespace Composer\Test\Package\Archiver;
 use Composer\Factory;
 use Composer\Package\Archiver\ArchiveManager;
 use Composer\Package\PackageInterface;
+use Symfony\Component\Process\ExecutableFinder;
 
 class ArchiveManagerTest extends ArchiverTest
 {
@@ -45,6 +46,11 @@ class ArchiveManagerTest extends ArchiverTest
 
     public function testArchiveTar()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
+        }
+
         $this->setupGitRepo();
 
         $package = $this->setupPackage();
@@ -62,6 +68,11 @@ class ArchiveManagerTest extends ArchiverTest
 
     public function testArchiveCustomFileName()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
+        }
+
         $this->setupGitRepo();
 
         $package = $this->setupPackage();
@@ -133,5 +144,19 @@ class ArchiveManagerTest extends ArchiverTest
         }
 
         chdir($currentWorkDir);
+    }
+
+    /**
+     * Check whether or not the given process is available.
+     *
+     * @param string $process The name of the binary to test.
+     *
+     * @return bool True if the process is available, false otherwise.
+     */
+    protected function isProcessAvailable($process)
+    {
+        $finder = new ExecutableFinder();
+
+        return (bool) $finder->find($process);
     }
 }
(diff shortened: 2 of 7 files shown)


AnrDaemon:
There's still 4 more cases, where tests don't cleanup after themselves, but it's getting increasingly harder to track them.
Still, 4 cases versus 90+… I think that's a good result.

tests/Composer/Test/Downloader/GitDownloaderTest.php
```diff
@@ -624,4 +650,18 @@ private function winCompat($cmd)
...
     }
+
+    /**
+     * Check whether or not the given process is available.
+     *
+     * @param string $process The name of the binary to test.
+     *
+     * @return bool True if the process is available, false otherwise.
+     */
+    protected function isProcessAvailable($process)
```
alcohol: Why not move this to the base `TestCase` class rather than implementing it over and over?

tests/Composer/TestCase.php
```diff
@@ -22,7 +22,37 @@
...
+    {
+        $fs = new Filesystem();
+        foreach ($this->tmpobjects as $object) {
+            if (is_dir($object)) {
+                $fs->removeDirectory($object);
+            }
+        }
+    }
+
+    public function __call($name, $arguments)
```
alcohol: What on earth do you need this magic for? You can just track the objects in the `getUniqueTmpDirectory()` method.

AnrDaemon: That method is static. I can't track anything with it without running into "using `$this` not in object context" every other time.
There's 3 places where that method is used statically. While I can fix `Composer\Test\Autoload\ClassMapGeneratorTest::testDump()` (it's used there statically in error, anyway), the two other places are `Composer\Test\AllFunctionalTest::setUpBeforeClass()` (itself static) and `Composer\Test\Mock\FactoryMock` (static, too).
If you have good idea, how to get around `$this` issue, I'm all ears.

tests/Composer/Test/Downloader/GitDownloaderTest.php
AnrDaemon: That was my next question. If you're okay with it, this can be done quick.

tests/Composer/TestCase.php
AnrDaemon: I could probably stuff a global static var in there. Supposedly it'll tear down when PHP shutdown.
BRB gonna test.

AnrDaemon: No… that won't do. If I make GC static, I'll won't have a way to call it unless I install shutdown handler.
Need something better.

AnrDaemon: I've got something working but it requires an additional proxy class to escape abstract TestCase class.
Short version is
```php
class TestCaseGC extends TestCase {}
```
```php
    private static $testCaseGc;

    public static function createConfig(IOInterface $io = null, $cwd = null)
    {
        if (!isset(self::$testCaseGc)) {
            self::$testCaseGc = new TestCaseGC();
        }

        $config = new Config(true, $cwd);

        $config->merge(array(
            'config' => array('home' => self::$testCaseGc->getUniqueTmpDirectory()),
            'repositories' => array('packagist' => false),
        ));

        return $config;
    }
```
On the upside, it entirely cleaned up the field. Not even a single leftover file or directory remains.

AnrDaemon: https://github.com/composer/composer/compare/master...AnrDaemon:fix-unclean-tests-1a

tests/Composer/Test/Downloader/GitDownloaderTest.php
alcohol: I think my suggestion implies that I am okay with it :-)

tests/Composer/TestCase.php
```diff
@@ -22,7 +22,37 @@
 {
     private static $parser;
 
-    public static function getUniqueTmpDirectory()
+    /** All temporary objects holding
+    * @var array
```
alcohol: I don't think this is how docblocks should be indented.. ;-)

tests/Composer/TestCase.php
alcohol: Why is this somewhere else and not in this branch?

tests/Composer/Test/Downloader/GitDownloaderTest.php
```diff
@@ -624,4 +650,18 @@ private function winCompat($cmd)
 
         return $cmd;
     }
+
+    /**
+     * Check whether or not the given process is available.
```
stof: ``git`` is not a process. It is an executable

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
```diff
@@ -45,6 +46,11 @@ public function testUnknownFormat()
 
     public function testArchiveTar()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
+            return $this->markTestSkipped('git is not available.');
```
stof: no need to return, as it throws

tests/Composer/TestCase.php
```diff
@@ -22,7 +22,37 @@
 {
     private static $parser;
 
-    public static function getUniqueTmpDirectory()
+    /** All temporary objects holding
```
stof: the short description should not be on the ``/**`` line

tests/Composer/TestCase.php
```diff
@@ -22,7 +22,37 @@
...
+    * @var array
+    */
+    private $tmpobjects = array();
+
+    public function __destruct()
```
stof: you should hook into ``tearDown`` rather than ``__destruct``, as PHPUnit has dedicated hooks

alcohol: That's a good observation. Also would mean you could do with a lot less magic.

tests/Composer/Test/Downloader/GitDownloaderTest.php
```diff
@@ -195,6 +206,11 @@ public function testDownloadWithCache()
 
     public function testDownloadUsesVariousProtocolsAndSetsPushUrlForGithub()
     {
+        // Ensure that git is available for testing.
+        if (!$this->isProcessAvailable('git')) {
```
povils: I see that code block as duplication

tests/Composer/Test/Downloader/GitDownloaderTest.php
```diff
@@ -624,4 +650,18 @@ private function winCompat($cmd)
 
         return $cmd;
     }
+
+    /**
+     * Check whether or not the given process is available.
+     *
+     * @param string $process The name of the binary to test.
+     *
+     * @return bool True if the process is available, false otherwise.
```
povils: this comment is redundant

tests/Composer/Test/Downloader/GitDownloaderTest.php
```diff
@@ -624,4 +650,18 @@ private function winCompat($cmd)
 
         return $cmd;
     }
+
+    /**
+     * Check whether or not the given process is available.
+     *
+     * @param string $process The name of the binary to test.
```
povils: $executableName

tests/Composer/TestCase.php
```diff
@@ -22,7 +22,37 @@
...
+            }
+        }
+    }
+
+    public function __call($name, $arguments)
+    {
+        if ($name === 'getUniqueTmpDirectory') {
+            $this->tmpobjects[] = $result = call_user_func_array(__CLASS__ . "::$name", $arguments);
+            return $result;
+        } else {
```
povils: else is unneccessary

tests/Composer/TestCase.php
AnrDaemon: @stof, using tearDown increase testing times tenfold for tests, that use setup but then don't do anything with it and run multiple times over a large dataset.
I.e. `tests\Composer\Test\Util\FilesystemTest.php `
setup+teardown = 45 sec
__destruct = 4 sec for PHP 5.3.29, 1.9 sec for 7.0.14.

stof: I'm not suggesting to create the temp dir in setUp. It should still be created by ``getUniqueTmpDirectory`` when needed.
I just suggested using ``tearDown`` for the tearDown logic.

AnrDaemon: How to use "tearDown logic" it in that FactoryMock ?

tests/Composer/TestCase.php
AnrDaemon: Why?

tests/Composer/TestCase.php
AnrDaemon: I wasn't sure it is an acceptable approach. Didn't want to push things back and forth.

tests/Composer/Test/Downloader/GitDownloaderTest.php
AnrDaemon: Duplication of what? Please finish your sentence.

tests/Composer/TestCase.php
AnrDaemon: In this specific case it is(was) necessary to avoid "the use of `$this` outside object context".

tests/Composer/Test/Downloader/GitDownloaderTest.php
AnrDaemon:
```php
    protected function isProcessAvailable($process)
```
The method named "isProcessAvailable", and its argument is `$process`.
You should appeal to the original creathor of this method, if you want to.

tests/Composer/Test/Downloader/GitDownloaderTest.php
povils:
```php
// Ensure that git is available for testing.
if (!$this->isProcessAvailable('git')) {
    return $this->markTestSkipped('git is not available.');
}
```
Even comment is duplicated :) try to extract into separate method  with meaningful name :)

tests/Composer/Test/Downloader/GitDownloaderTest.php
povils: Changing method name to more expressive and meaningful is one of the greatest refactoring :)

tests/Composer/TestCase.php
povils: I think you misundestood me. You don't need 'else' keyword to express condition
```php
if ($name === 'getUniqueTmpDirectory') {
    $this->tmpobjects[] = $result = call_user_func_array(__CLASS__ . "::$name", $arguments);

    return $result;
 }

 return call_user_func_array(__CLASS__ . "::$name", $arguments);
```

tests/Composer/Test/Downloader/GitDownloaderTest.php
stof: @AnrDaemon you are the one adding this method in Composer. and as I already said in my previous review, talking about ``process`` in the naming is wrong, as it is not about processes at all but about executables (``git`` is **not** a process)

tests/Composer/TestCase.php
AnrDaemon: Wasn't sure the approach is sensible. Or acceptable.
I've made a pull request against this one for review https://github.com/AnrDaemon/composer/pull/1

tests/Composer/Test/Downloader/GitDownloaderTest.php
AnrDaemon: I don't mind puzzles. When I have an idea how to solve them.
Now, if you have an idea, how to extract a method blocker code into a separate method, I'm all ears.

tests/Composer/Test/Downloader/GitDownloaderTest.php
AnrDaemon: @stof, blame @Seldaek https://github.com/composer/composer/commit/3aa78431464a695f0c707e33413a0f81a9347a0b

AnrDaemon: To clarify: I can change the method name, just not with this PR.
I could change many things here, but for now, I'm trying to keep it clear and concise for review, not changing code without a necessity.

tests/Composer/TestCase.php
AnrDaemon: I don't like this particular practice. It makes code less readable by concealing the intention.
The ONLY place where I use this is when I throw exceptions.

povils: sorry, but $x=$y=$z doesnt improve readiblity at all. I'm not saying that your code is bad, just advising you to make your code more readable, maintainable and clean :)

tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
AnrDaemon: Got it.
Please review https://github.com/composer/composer/pull/5990

tests/Composer/TestCase.php
AnrDaemon: `if ($name === 'getUniqueTmpDirectory') {` is a test to see if we should execute this call at all.
Your proposed change affects the logic of the code.

PR #5990, opened 2016-12-22
Skip git-related tests if no git found.

AnrDaemon:


AnrDaemon:
I'm not sure, if ArchiveManagerTest should be completely excluded, seeing as 2/3 of it is related to git.

PR #6001, opened 2016-12-28
Do not throw InvalidArgumentException with the 'show' command if the …

uberhacker:
…working-dir argument is set and composer.json exists

  - This prevents scripts from aborting when looping through a directory of projects

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -166,8 +166,19 @@ EOT
             if (empty($package)) {
                 list($package, $versions) = $this->getPackage($installedRepo, $repos, $input->getArgument('package'), $input->getArgument('version'));
 
-                if (!$package) {
-                    throw new \InvalidArgumentException('Package '.$input->getArgument('package').' not found');
+                if (empty($package)) {
+                    $options = $input->getOptions();
+                    if (isset($options['working-dir'])) {
+                        $composer_json = $options['working-dir'] . '/composer.json';
+                        if (file_exists($composer_json)) {
+                            $io->writeError('Package ' . $packageFilter . ' not found in ' . $composer_json);
+                            return;
+                        } else {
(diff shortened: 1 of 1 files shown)


Seldaek:
Merged into master/1.3, thanks

dzuelke:
This breaks `composer show package/nothere`, which no longer returns an exit status of 1. See #6092

dzuelke:
In fact, this whole change should be rolled back. The "fix" is actually what's broken. It intentionally no longer returns 1, just because someone wrote a bash script that doesn't handle errors properly.

dzuelke:
If you need to tell the cases apart, then please re-do a PR for separate exit codes, @uberhacker. In the meantime, this should be reverted ASAP.

uberhacker:
@dzuelke: Why does the exit code matter?  If the show command doesn't return results, isn't that a good indication the package doesn't exist?   Probably what happened in your case is the package was installed at one point and then removed later, leaving an empty composer.json file behind in the process.  Regarding scripts that don't handle errors properly, how do you recover from a throw new exception?

dzuelke:
The exit code matters because other programs may run `composer show --quiet some/package` to programmatically determine whether or not a package is there. Parsing output, which changes across versions, is not the proper way to do that. That's precisely what exit statuses are for - if the package exists, exit 0, otherwise, exit with another code.

If you run a script that invokes the command multiple times, then you need to handle the error from the `InvalidArgumentException` in the loop. If the difference between "package is not there" and "no composer.json is there" matters to you, then that can be addressed by defining several exit codes with different meanings.

uberhacker:
Please see PR #6093.

curry684:
In a properly set up application, performing a targeted operation with an invalid target should return no output on STDOUT, an error on STDERR, and return exit code 1 to faciliate scripting.

As opposed to this, a search without results should return 0 as that is a valid outcome for searches, which would allow scripters to check for the package existence before attempting the targeted operation.

PR #6003, opened 2016-12-28
Update xdebug troubleshooting, fixes #5997

### LINKED ISSUE #5997: Xdebug warning message links to non-existent documentation
When running composer with xdebug enabled, the following warning is emitted:

> You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug

But the URL redirects to a documentation resource that doesn't exist. The documentation apparently needs to be created or replaced, or the link needs to be updated or removed.

johnstevenson:


doc/articles/troubleshooting.md
```diff
@@ -149,6 +149,13 @@ php -d memory_limit=-1 composer.phar <...>
 
 This issue can also happen on cPanel instances, when the shell fork bomb protection is activated. For more information, see the [documentation](https://documentation.cpanel.net/display/ALD/Shell+Fork+Bomb+Protection) of the fork bomb feature on the cPanel site.
 
+## Xdebug impact on Composer
+
+To improve performance when the xdebug extension is enabled, Composer automatically restarts PHP without it.
+You can override this behavior by using an environment variable: `COMPOSER_ALLOW_XDEBUG=1`.
+Composer will always show a warning if xdebug is being used. If you see this unexpectedly, then
```
staabm: Is it no longer possible to prevent that warnung using a ENV var?

johnstevenson: Good point. `COMPOSER_DISABLE_XDEBUG_WARN` is still used, but I don't know if this is still required now that the default behavior is to disable xdebug. Perhaps @alcohol might have an idea here?

alcohol: Ugh. I don't know anymore. We used to warn about xdebug, now we disable xdebug by default, but then people can force xdebug to stay loaded and they get the warning again which we may or may not be able to disable and arghjfhgkjedhgfkjragfadnbkjanflkjadsnkjfmnal

johnstevenson: I think I'll document it as it still exists.

Issue #6004, opened 2016-12-28
Composer is losing memory_limit when starting process without xdebug


oleg-andreyev:
When I run this command:

```
php -dmemory_limit=3G composer.phar update symfony/symfony --prefer-stable --with-dependencies
// or
php -dmemory_limit=-1 composer.phar update symfony/symfony --prefer-stable --with-dependencies
```

I get the following output:

```
PHP Fatal error: Allowed memory size of XXXXXX bytes exhausted <...>
```

And I expected this to happen:
- composer update should pass

Reason:

Composer is starting new process without memory_limit and uses default one (1G):
```
'/usr/bin/php' '-c' '/tmp/kaoOdw' 'composer.phar' 'update' '--ansi' 'symfony/symfony' '--prefer-stable' '--with-dependencies' '-vvv'
```

alcohol:
@johnstevenson I think this is an edge case we did not entirely take into account..

johnstevenson:
@alcohol Not so much not taken into account, but more _didn't know what to do with_ I'm afraid!

We don't have direct access to user-supplied php params so it would be a case of second guessing what options might be used. There are various things you can do to compare a current ini setting to what is in an ini file but it is not going to be foolproof.

johnstevenson:
@alcohol I need to change the tmp ini file creation, as per #5995, so I'll incorporate some ini-setting comparison logic as well.

Seldaek:
I did a quick fix for the memory_limit for now. I don't know if we really need to support *all* ini settings or not.. I kinda doubt there are a lot in wide use, but memory_limit is kinda critical here.

PR #6005, opened 2016-12-28
persisting memory_limit when executing composer with xdebug enabled

oleg-andreyev:
My first PR to composer. Great achievementent 🥇 

This PR contains fix for #6004

### DIFF
--- a/src/Composer/XdebugHandler.php
+++ b/src/Composer/XdebugHandler.php
@@ -230,7 +230,8 @@ class XdebugHandler
      */
     private function getCommand()
     {
-        $phpArgs = array(PHP_BINARY, '-c', $this->tmpIni);
+        $memoryLimit = ini_get('memory_limit');
+        $phpArgs = array(PHP_BINARY, '-d', 'memory_limit=' . $memoryLimit, '-c', $this->tmpIni);
         $params = array_merge($phpArgs, $this->getScriptArgs($_SERVER['argv']));
 
         return implode(' ', array_map(array($this, 'escape'), $params));


staabm:
Should a fix to this problem pass all the `-d` params along?

johnstevenson:
How do we obtain all the -d params?

oleg-andreyev:
it's possible to use `ini_get_all` and iterate all values but I don't think it's a great solution.
Another solution to get all -d params passed from commands is to parse command string

staabm:
Not sure whether those are provided via $argv

johnstevenson:
Or you could just raise the memory-limit in the ini file?

oleg-andreyev:
@staabm $argv doesn't contain information which is passed before composer.
@johnstevenson just raising memory_limit isn't a solution

johnstevenson:
How about using the `COMPOSER_ALLOW_XDEBUG` environment variable then, so the process doesn't restart. Or not having xdebug enabled in the first place? 

Because we can never know what was passed on the command line as ini options, we will never be able to satisfy everyone by adding extra params. That's my view, any way.

oleg-andreyev:
Yes `COMPOSER_ALLOW_XDEBUG` will work. But I think this issue should be solved anyway, because before this improvement/workaround everything was working, so could be considered as bug.

staabm:
Underlying issue was fixed with https://github.com/composer/composer/commit/2c39c43a9e9c0ebf8b0477adcfdafc26b94116d7

Seldaek:
Yup sorry I missed this PR and did the fix in another way..

PR #6012, opened 2016-12-29
Fix the class name for fossil repositories

stof:
Fossil is a VCS driver. It does not have its own Repository class


Seldaek:
Thanks

PR #6013, opened 2016-12-29
Implement schema validation for repositories

stof:
This is the next step of the schema improvements, validating repositories. This should give a avoid mistakes putting the ``autoload`` config outside the ``package`` configuration for package repositories for instance.

For inline packages, I kept things simple by referencing the schema for the full composer.json, even though root-only properties are actually useless here as they would not be used.

The second commit changes the way we load the schema when validating so that references are resolved properly (otherwise they don't work). The previous code was working only because the schema did not have any references in it. To keep things simple, I dropped support for ``justinrainbow/json-schema`` 1.x and 2.x, as support for references was much harder for them (and maybe not working). These versions are totally outdated anyway (and Composer ships with 4.x anyway).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
     },
     "require": {
         "php": "^5.3.2 || ^7.0",
-        "justinrainbow/json-schema": "^1.6 || ^2.0 || ^3.0 || ^4.0",
+        "justinrainbow/json-schema": "^3.0 || ^4.0",
         "composer/ca-bundle": "^1.0",
         "composer/semver": "^1.0",
         "composer/spdx-licenses": "^1.0",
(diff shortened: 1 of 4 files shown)


staabm:
Those changes are awesome. Is this functionality somehow covered by tests already?

alcohol:
> For inline packages, I kept things simple by referencing the schema for the full composer.json, even though root-only properties are actually useless here as they would not be used.

Would the reverse not be possible? Start with non-root composer.json, and then reference this and append root-only attributes to it.

Seldaek:
Thanks, looks good!

stof:
@alcohol this is something I thought about. But I'm not sure it is possible.
And I wanted to move forward step by step to make changes less disruptive if something breaks

PR #6034, opened 2017-01-03
Refactor the validation of inline package definitions

### LINKED ISSUE #6028: Validation of inline package definition in repositories is broken in latest build
My `composer.json`:

```json
{
    "require": {
        "xxxx/yyyyy": "0.0.1"
    },
    "repositories": [
        {
          "type": "package",
          "package": {
            "name": "xxxx/yyyyy",
            "version": "0.0.1",
            "authors": [
              {
                "name": "xxxx xxxx",
                "email": "xxxxx@xxxx.xyz"
              }
            ],
            "dist": {
              "type": "zip",
              "url": "https://xxxxx/xyz.zip",
              "reference": "123abc"
            }
          }
        }
    ]
}
```

```
not possible, error occurs directly, see output
```

When I run this command:

```
php composer.phar install -o
```
or
```
php composer.phar update -o
```

I get the following output:

```

   - repositories[0] : Must contain no more than 1 properties
   - repositories[0].package : Object value found, but a boolean is required
   - repositories[0].package : Does not have a value in the enumeration [false]
   - repositories[0].package : Object value found, but an array is required
   - repositories[0].package : The property dist is not defined and the definition does not allow additional properties
   - repositories[0].package : Failed to match exactly one schema
   - repositories[0].url : The property url is required
   - repositories[0].type : Does not have a value in the enumeration ["composer"]
   - repositories[0].type : Does not have a value in the enumeration ["vcs","github","git","gitlab","git-bitbucket","hg","hg-bitbucket","fossil","perforce","svn"]
   - repositories[0].type : Does not have a value in the enumeration ["path"]
   - repositories[0].type : Does not have a value in the enumeration ["artifact"]
   - repositories[0].type : Does not have a value in the enumeration ["pear"]
   - repositories[0] : Failed to match exactly one schema

```

install package from given url

stof:
Reusing the root schema was a mistake, because the root schema does not have the dist and source properties (which are populated by repository objects).
And adding new properties in a referenced schema using ``additionalProperties: false`` is not possible in json-schema draft 4.

closes #6028 (tested with the files given by @skors and @Rudloff)

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -46,33 +46,7 @@
             "description": "License name. Or an array of license names."
         },
         "authors": {
-            "type": "array",
-            "description": "List of authors that contributed to the package. This is typically the main maintainers, not the full list.",
-            "items": {
-                "type": "object",
-                "additionalProperties": false,
-                "required": [ "name"],
-                "properties": {
-                    "name": {
-                        "type": "string",
-                        "description": "Full name of the author."
-                    },
-                    "email": {
-                        "type": "string",
-                        "description": "Email address of the author.",
-                        "format": "email"
-                    },
-                    "homepage": {
-                        "type": "string",
-                        "description": "Homepage URL for the author.",
-                        "format": "uri"
-                    },
-                    "role": {
-                        "type": "string",
-                        "description": "Author's role in the project."
-                    }
-                }
-            }
+            "$ref": "#/definitions/authors"
         },
         "require": {
             "type": "object",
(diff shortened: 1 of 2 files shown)


res/composer-schema.json
```diff
@@ -662,8 +671,156 @@
...
+                        "shasum": {
+                            "type": "string"
+                        },
+                        "mirrors": {
+                            "type": "array"
+                        }
+                    }
+                }
+            },
+            "additionalProperties": true
```
stof: I did this to be more kind to existing files for now, but the tested files actually don't need it.

Seldaek:
The tests are failing not entirely sure why (well first failure was an obvious json error, but now it got weirder, maybe you saw it already while working on it?)

staabm:
maybe you should include the known-to-be-failling composer-json files into the testsuite?

stof:
@Seldaek the weird failure is because this testcase does not resolve references when loading the schema (and I missed the fact that a test was using the json schema validator on its own when refactoring the composer main schema validation code)

Seldaek:
Thanks for the fix

PR #6035, opened 2017-01-03
Added possibility to include the VCS directories when creating dist package

meridius:
Works by adding the `ignore-vcs` bool option to the `archive` section. This option defaults to **true**, so VCS directories are ignored by default, hence original behavior is preserved.

This is extremely useful when working with bleeding edge of packages and/or in need to hot-switch the committed states of required package, whilst **preserving** the ability to cache the dist package and rapidly **speed-up** the install time.


src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -24,8 +24,15 @@
  */
 class ArchiveManager
 {
+    const IGNORE_VCS_KEY = 'ignoreVcs';
+
+    const IGNORE_VCS_DEFAULT = true;
```
stof: IMO, we don't need to create constants for them

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -163,7 +170,9 @@ public function archive(PackageInterface $package, $format, $targetDir, $fileNam
         $tempTarget = sys_get_temp_dir().'/composer_archive'.uniqid().'.'.$format;
         $filesystem->ensureDirectoryExists(dirname($tempTarget));
 
-        $archivePath = $usableArchiver->archive($sourcePath, $tempTarget, $format, $package->getArchiveExcludes());
+        $excludes = $package->getArchiveExcludes();
+        $excludes[self::IGNORE_VCS_KEY] = $package->getIgnoreVcs();
```
stof: mixing this inside the excludes is wrong. This has nothing to do with excluded paths. It is a separate config setting

stof:
This will be inconsistent, because archivers may also be used when the package is not installed with its VCS info (if the code was downloaded from dist from somewhere before that rather than installed from source)

stof:
And unless you change the preferences, any archive built for a non-dev version will be downloaded from source

src/Composer/Package/Archiver/ArchiveManager.php
meridius: I guess. But I thought it'd be better since they are later used across composer and in satis. This way we could maintain some kind of consistency.

- https://github.com/composer/composer/pull/6035/files#diff-40f2e6c42699610f9aa99aec74c1bad2R43
- https://github.com/composer/composer/pull/6035/files#diff-b57a8c02c0d8217f32f8ebc6871bf43cR62
- https://github.com/composer/satis/pull/387/files#diff-8d0e7cce9acecb86baa4eee824647adbR44

src/Composer/Package/Archiver/ArchiveManager.php
meridius: I know but I wanted to preserve the methods' signature and this seemed like the least painful way of how to pass the new flag.  
Do you have some better approach in mind?

stof: you are **not** keeping the method signature, as you change the meaning of the argument from a clear meaning to a mixed bag of stuff (and there might even be conflicts between your own key and the user-provided ones btw, so it is even broken)

meridius:
@stof I don't understand what you meant by https://github.com/composer/composer/pull/6035#issuecomment-271293920 and https://github.com/composer/composer/pull/6035#issuecomment-271294572

stof:
@meridius there is nothing guaranteeing you that you will have the ``.git`` folder inside your archive even when not ignoring it, because it may not be there already before calling the finder depending on the case.

Seldaek:
IMO it could be done as a flag to the archive command (`--include-vcs`?) like https://github.com/composer/composer/pull/6062 which should be sufficient I think. It would have less impact on the code base.

salarmehr:
http://stackoverflow.com/a/42707755/752603

Seldaek:
Closing as it doesn't seem to be going anywhere and the feature isn't clearly needed.

PR #6036, opened 2017-01-03
Preserve command line ini settings for restart

johnstevenson:
This is a more robust fix for #6004 as it handles any settings from the command line.

### DIFF
--- a/src/Composer/XdebugHandler.php
+++ b/src/Composer/XdebugHandler.php
@@ -136,7 +136,6 @@ class XdebugHandler
     {
         $this->tmpIni = '';
         $iniPaths = IniHelper::getAll();
-        $additional = count($iniPaths) > 1;
 
         if (empty($iniPaths[0])) {
             // There is no loaded ini
@@ -144,7 +143,7 @@ class XdebugHandler
         }
 
         if ($this->writeTmpIni($iniPaths)) {
-            return $this->setEnvironment($additional, $iniPaths);
+            return $this->setEnvironment($iniPaths);
         }
 
         return false;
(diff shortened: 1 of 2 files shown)


src/Composer/XdebugHandler.php
```diff
@@ -165,19 +166,56 @@ private function writeTmpIni(array $iniFiles)
...
+            if ($value === null || strpos($name, 'xdebug') === 0) {
+                continue;
+            }
+
+            if (!isset($iniConfig[$name]) || $iniConfig[$name] !== $value) {
+                // Based on main -d option handling in php-src/sapi/cli/php_cli.c
+                if ($value && !ctype_alnum($value)) {
+                    $value = '"'.str_replace('"', '\\"', $value).'"';
+                }
+                $content .= $name.'='.$value.PHP_EOL;
```
Seldaek: HHVM complains about an array to string conversion here. I guess some ini values can be arrays (maybe only in hhvm?) Anyway might be worth guarding against this..

Seldaek:
I do appreciate the work, but this PR adds 3-400ms of init time to composer for me.. if I compare before/after with `time composer -h`. I wonder if it is really worth it given it really just fixes the edge case of someone passing $randomFlagWeCantThinkOf to php.

Seldaek:
Given the impact I think I'd rather hold off on a merge until we see that this is needed (i.e. if people come up with new ini settings they use and we need to add every other week).

johnstevenson:
I'm surprised it adds that much time. I did worry about this but testing in a vm Ubuntu setup didn't seem to make much difference. On Windows it takes forever to create a new process anyway.

No worries about dropping this. The hhvm array thing is interesting because PHP's `ini_get_all()` specifically ignores array ini options. I'll fix this bit in case we should ever want to use it.

Seldaek:
Think we can close this for now :)

PR #6043, opened 2017-01-05
Don’t add github specials if there is no composer data

akinspe:
This can happen if an identifier doesn’t have a composer.json file (but other identifiers do)

VCSRepository->initialize() cycles through tags/branches. GitHubDriver->getComposerInformation() will always return data even if there is nothing returned by getBaseComposerInformation()

This request does not add the GitHub information if there is no composer file. Thus skipping the identifier.


Seldaek:
Thanks, merged in 1.3 and master branches.

PR #6053, opened 2017-01-09
use curl instead of wget for example script

euxneks:
curl seems to be more widely installed, especially in php docker images.
This sort of command is exactly what curl is for, really. See
https://daniel.haxx.se/docs/curl-vs-wget.html, particularly for the
analogy: "curl works more like the traditional unix cat command, it
sends more stuff to stdout, and reads more from stdin in a "everything
is a pipe" manner. Wget is more like cp, using the same analogue."

### DIFF
--- a/doc/faqs/how-to-install-composer-programmatically.md
+++ b/doc/faqs/how-to-install-composer-programmatically.md
@@ -9,7 +9,7 @@ An alternative is to use this script which only works with unix utils:
 ```bash
 #!/bin/sh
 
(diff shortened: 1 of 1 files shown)


Seldaek:
I don't think anyone can win this argument as we don't have enough numbers to know for sure.. wget is more common IMO, and replacing with curl is easier if you don't have wget as curl doesn't need any flags, so I'd rather keep the wget in the docs.

PR #6058, opened 2017-01-11
Fix Xdebug warning

laurisb:
Composer incorrectly shows Xdebug warning, even if extension is not loaded, but COMPOSER_ALLOW_XDEBUG is present.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -166,7 +166,7 @@ class Application extends BaseApplication
                 $io->writeError('<warning>Composer only officially supports PHP 5.3.2 and above, you will most likely encounter problems with your PHP '.PHP_VERSION.', upgrading is strongly recommended.</warning>');
             }
 
-            if (extension_loaded('xdebug') && !getenv('COMPOSER_DISABLE_XDEBUG_WARN') || getenv('COMPOSER_ALLOW_XDEBUG')) {
+            if (extension_loaded('xdebug') && (!getenv('COMPOSER_DISABLE_XDEBUG_WARN') || getenv('COMPOSER_ALLOW_XDEBUG'))) {
                 $io->writeError('<warning>You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug</warning>');
             }
 


alcohol:
@johnstevenson I think this makes sense. Would you agree?

Seldaek:
Thanks

johnstevenson:
@alcohol Yup, this makes sense. However, I don't think `COMPOSER_ALLOW_XDEBUG` is needed:

```php
if (extension_loaded('xdebug') && !getenv('COMPOSER_DISABLE_XDEBUG_WARN')) ...
```

vlakoff:
(introduced in https://github.com/composer/composer/commit/c5dcedd0db1b4f8576973b016ecc73be1c34492c)
The purpose is to disable warning if `COMPOSER_DISABLE_XDEBUG_WARN` or `COMPOSER_ALLOW_XDEBUG` is set.
So the correct logic is:
```php
if (extension_loaded('xdebug') && !getenv('COMPOSER_DISABLE_XDEBUG_WARN') && !getenv('COMPOSER_ALLOW_XDEBUG'))
```

alcohol:
@vlakoff you contradict yourself, the "correct logic" you present does not account for the "or" in your description of said scenario.

vlakoff:
Nope, I'm right.
Show warning iif both flags are not set, so don't show warning if either flag (or both) is set.
!A && !B = not (A || B)

But this is not relevant anymore, as in #6072 I removed that extra test.
Your statement proves it won't harm to make it simpler.

alcohol:
Either you are incredibly ignorant, or you are just trolling me now. Regardless, not really worth my time. So whatever.

vlakoff:
To summarize:
* The boolean expression in https://github.com/composer/composer/commit/c5dcedd0db1b4f8576973b016ecc73be1c34492c was incorrect.
* This PR intended to fix it but was incorrect too.
* In #6072 I fixed the boolean expression, to achieve the intent of https://github.com/composer/composer/commit/c5dcedd0db1b4f8576973b016ecc73be1c34492c.
* Then we all agreed to undo https://github.com/composer/composer/commit/c5dcedd0db1b4f8576973b016ecc73be1c34492c and I updated #6072 accordingly.

alcohol:
That seems correct.

PR #6061, opened 2017-01-11
Add full stop to all command descriptions

svenluijten:
Simple but effective, I simply added a full stop to each command's description if it was not already there.


alcohol:
I'm fine with either way (either full stop all descriptions, or full stop none). So whichever makes the current situation more consistent, :+1:.

svenluijten:
I'm fine with either as well. Reason I chose to add a full stop to descriptions that didn't have it is because I think it looks more natural. I'd be okay with changing it though 🙂

PR #6065, opened 2017-01-13
Trying to fix the new setExecuteOperations() on the installer

Toflar:
This is a follow up for https://github.com/composer/composer/pull/5787.
As mentioned there, @Seldaek's implementation unfortunately doesn't quite work for me.

The problem is that when you call

```php
$installer
    ->setWriteLock(true)
    ->setExecuteOperations(false);
```

a `composer.lock` file will never be written because the `$localRepo` has no packages.


Seldaek:

BTW note that this will update vendor/composer/installed.json too which will sort of corrupt the vendor dir, but as it is only used programmatically and you don't actually have vendors installed I guess it's an acceptable trade-off for you.

PR #6072, opened 2017-01-15
Fix Xdebug warning (proper)

vlakoff:
Proper fix, follow-up to #6058.

Alternatively, you could undo https://github.com/composer/composer/commit/c5dcedd0db1b4f8576973b016ecc73be1c34492c, as a result it would require to set both constants `COMPOSER_ALLOW_XDEBUG` and `COMPOSER_DISABLE_XDEBUG_WARN` to allow Xdebug without warning.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -166,7 +166,7 @@ class Application extends BaseApplication
                 $io->writeError('<warning>Composer only officially supports PHP 5.3.2 and above, you will most likely encounter problems with your PHP '.PHP_VERSION.', upgrading is strongly recommended.</warning>');
             }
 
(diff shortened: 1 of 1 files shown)


johnstevenson:
This is not the correct behavior. See the docs: https://getcomposer.org/doc/articles/troubleshooting.md#xdebug-impact-on-composer

Per my coment in #6058 It should be:

```php
if (extension_loaded('xdebug') && !getenv('COMPOSER_DISABLE_XDEBUG_WARN'))
```

vlakoff:
So, we agree on the alternative proposal I had written in my message above.

Updated PR accordingly.

alcohol:
This is the correct behaviour.

Seldaek:

Issue #6092, opened 2017-01-21
ShowCommand exits 0 even if given package is not installed


dzuelke:
Before afdae9c0e2094e8e6f0884167caa1022df7d99bf, `composer show something/notinstalled` would return 1 if the package was not installed.

Now the exit status is always 0, so code that calls the command can no longer determine if a package is installed or not.

dzuelke:
This is caused by #6001, which should be rolled back. The "fix" is actually what's broken. It intentionally no longer returns 1, just because someone wrote a bash script that doesn't handle errors properly.

alcohol:
I would agree that this should return `1` if the show command is called with a not installed package as argument.

Seldaek:
Fixed by #6093

PR #6093, opened 2017-01-21
Return 1 if composer.json exists, but the package doesn't.

uberhacker:
This should resolve issue #6092.

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -173,7 +173,7 @@ EOT
                     }
 
                     $io->writeError('Package ' . $packageFilter . ' not found in ' . $options['working-dir'] . '/composer.json');
-                    return;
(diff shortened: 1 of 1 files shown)


dzuelke:
That's the solution I also had at first, but then I saw the reason for your change... your use case will now break again, no? If a work dir is set, and `composer.json` exists, and the package does not exist, then you'll get an exit status of 1 again now, and your script that loops over directories will fail.

uberhacker:
I think this is fine.  The issue I was having earlier stemmed from using the `cgr` composer global require wrapper command: https://github.com/consolidation/cgr.  Specifically, the issue was with `cgr info`.  You are correct that scripts should handle errors properly.  This PR change does not affect the processing of that command.  If only composer didn't have so many shortcomings. :)

Seldaek:

PR #6094, opened 2017-01-22
Implement bitbucket api v2

stefangr:
This PR implements the following changes:

- implement v2 API calls as much as possible, this has the advantage of being able to request only the data needed and ordering the API results
- Bring the Mercurial implementation to the same level as the git implementation
- Store the access-token in the auth.json (with its expiration time) so it can be re-used while not expired
- Make sure that bitbucket access-token don't leak to other sites
- Set access-token in Authorization header instead of appending to the url


src/Composer/Repository/Vcs/BitbucketDriver.php
```diff
@@ -47,6 +64,49 @@ public function initialize()
...
+            $this->owner,
+            $this->repository,
+            http_build_query(
+                array('fields' => '-project,-owner')
+            )
+        );
+
+        $repoData = JsonFile::parseJson($this->getContentsWithOAuthCredentials($resource, true), $resource);
+        if ($this->fallbackDriver) {
+            throw new BitbucketFallbackException();
```
alcohol: I am a bit puzzled by this flow. Can you elaborate on your train of thought here?

src/Composer/Repository/Vcs/BitbucketFallbackException.php
```diff
@@ -0,0 +1,17 @@
...
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Repository\Vcs;
+
+class BitbucketFallbackException extends \RuntimeException
```
alcohol: What exactly is the purpose of this exception? In what case/scenario is it thrown?

src/Composer/Repository/Vcs/BitbucketDriver.php
stefangr: In the `getContentsWithOAuthCredentials` method, the code first tries to download the file without credentials.
If, however, the repo is private and no credentials are present in the auth.json and composer is used in a non-interactive way, the code tries to fall back to SSH (and assigns an instance of the GitDriver or HgDriver to the `fallbackDriver` property).

The `BitbucketFallbackException` is caught in the `getRootIdentifier` method in either GitBitbucketDriver of HgBitbucketDriver and from then on the fallback driver is used.

stefangr: I can change this to let the `getRepoData` method return true on success or false in case of the fallback and get rid of the exception.

src/Composer/Repository/Vcs/BitbucketDriver.php
```diff
@@ -47,6 +64,49 @@ public function initialize()
...
+     *
+     * @return void
+     */
+    protected function getRepoData()
+    {
+        $resource = sprintf(
+            'https://api.bitbucket.org/2.0/repositories/%s/%s?%s',
+            $this->owner,
+            $this->repository,
+            http_build_query(
```
stof: you should use ``http_build_query(array(...), null, '&')``. The default delimiter of ``http_build_query`` cannot be relied upon, as it is configurable through php.ini.

This also applies to other place

src/Composer/Repository/Vcs/GitBitbucketDriver.php
```diff
@@ -34,90 +30,24 @@ public function getRootIdentifier()
...
-            foreach ($branchData as $branch => $data) {
-                $this->branches[$branch] = $data['raw_node'];
+            if ($this->vcsType !== 'git') {
+                throw new \RuntimeException(
+                    $this->url.' does not appear to be a git repository, use '.
+                    $this->cloneHttpsUrl.' if this is a mercurial bitbucket repository'
+                );
             }
+
+            $main_branch_data = $this->getMainBranchData();
```
stof: please use camelCased variable names

src/Composer/Util/Bitbucket.php
```diff
@@ -175,17 +180,66 @@ public function authorizeOAuthInteractively($originUrl, $message = null)
...
+    /**
+     * @param string $originUrl
+     * @return bool
+     */
+    private function getTokenFromConfig($originUrl)
+    {
+        $authConfig = $this->config->get('bitbucket-oauth');
+
+        if (empty($authConfig) ||
+            ! isset($authConfig[$originUrl]) ||
```
stof: There 2 conditions are useless. They are already checked by ``isset($authConfig[$originUrl]['access-token'])``

src/Composer/Util/Git.php
```diff
@@ -122,17 +122,17 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
 
                     if (!$bitbucketUtil->authorizeOAuth($match[1]) && $this->io->isInteractive()) {
                         $bitbucketUtil->authorizeOAuthInteractively($match[1], $message);
-                        $token = $bitbucketUtil->getToken();
-                        $this->io->setAuthentication($match[1], 'x-token-auth', $token['access_token']);
+                        $access_token = $bitbucketUtil->getToken();
```
stof: camelCase please

src/Composer/Repository/Vcs/BitbucketDriver.php
stefangr: Thanks for the review.. Implemented the changes you mentioned.

Seldaek:
Thanks! However it's still quite painful to setup the consumer etc, I kinda wish bitbucket let us do this more easily, but I guess this is still an improvement.

One thing that might be easier is to tell people to use App passwords (https://bitbucket.org/account/user/**USER**/app-passwords/new), which can be used as basic auth with `$username`/`$appPassword`. That is a bit easier to setup and only requires one token be passed in. Then if that is present I think the requests to bitbucket would just work and not hit the oauth stuff at all? I think that might be worth looking into if you have time..

eiriksm:
>One thing that might be easier is to tell people to use App passwords (https://bitbucket.org/account/user/**USER**/app-passwords/new), which can be used as basic auth with $username/$appPassword. That is a bit easier to setup and only requires one token be passed in. Then if that is present I think the requests to bitbucket would just work and not hit the oauth stuff at all? I think that might be worth looking into if you have time..

Sorry for resurrecting this old thread, but this is more or less 1:1 with what I was thinking today, and found myself searching for ways to use that.

Is this still unsupported in composer? I can see myself needing it, so might take a stab at implementing it if interest is still there?

Issue #6088, opened 2017-01-19
Documentation for package versioning, version constraints, and stability flags is confusing


kael-shipman:
In general the docs are very good and complete, but I found these areas to be confusing. In general, the distinction between a VCS version (a "ref", i.e., commit or tag) and a Composer package version is unclear.

For example, on the [Libraries page](https://github.com/composer/composer/doc/02-Libraries.md#branches) under "Specifying the Version > Branches", it states "For every branch, a package development version will be created.... `master` results in a `dev-master` version." I took this to mean that Composer would look for a `dev-master` branch in my VCS, so I created that branch and then specified `dev-master` as the version constraint in my `require` clause. Upon trouble-shooting, I discovered that the `dev-` prefix simply meant that composer would clone the git repo at the branch specified after the prefix (`master` in this case), instead of just copying the files.

I've cloned the repo and am working on rewording this documentation. I'll submit a pull request when I've got something to look at.

PR #6097, opened 2017-01-23
Updated documentation to clarify versions and package resolution (refs #6088)

kael-shipman:
This is a first attempt at clarifying these points. I added a lot of information about the internals of Composer's process for resolving packages. I tried not to flood the user with unnecessary information, but the details I added were those that I myself found to be lacking and that, had they been present before, would have helped me better understand how to publish libraries and use Composer to manage them.

Having taken this foray into the docs, I remain convinced that they're very good, but could be better. I'd like to have a go at maybe a total rewrite of the Composer "book". There are just a few details that were hard to place in the current set of files, and I think this is because they're not structured in the best possible way.

Anyway, I know this pull request is a lot to read through, so please forgive that, but I hope the devs and users of Composer find it to be a valuable!

### DIFF
--- a/doc/01-basic-usage.md
+++ b/doc/01-basic-usage.md
@@ -33,18 +33,21 @@ As you can see, [`require`](04-schema.md#require) takes an object that maps
 **package names** (e.g. `monolog/monolog`) to **version constraints** (e.g.
 `1.0.*`).
 
+It uses this information to search for the right set of files in package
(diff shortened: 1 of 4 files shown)


naderman:
@kael-shipman I merged this with a couple more changes, see https://github.com/naderman/composer/compare/9a755fa7d475e4b5b082c37b605e5960bd9102b4...naderman:docs-improve if you want to review them. Send a new PR if you disagree with anything please.

Thank you!

PR #6105, opened 2017-01-26
Fix invalid calculation on non-numeric string in GitDownloader, errors in PHP 7.1

pfrenssen:
In `GitDownloader::cleanChanges()` the error message that is shown when more than 10 modified files are present in a dependency that is downloaded with git does an invalid calculation on a non-numeric string.

On older PHP versions this is ignored silently, but on PHP 7.1 this now throws the following error:

>  [ErrorException] A non-numeric value encountered

The fix is trivial.


Seldaek:
Merged in 1.3 and master, thanks :)

PR #6127, opened 2017-02-03
Add fix for incorrect encoding in getcwd() result with non-latin chars on Windows

wapmorgan:
Hello. I faced an issue when working with project, when working dir contains non-latic chars (e.g. `C:\Users\Сергей\Documents\symfony-test`). If root project has `autoload.classmap` directives, then composer generates invalid paths for files:

## Current behavior
Currently composer generates invalid classmap.

```php
<?php

return array(
    'AppCache' => $baseDir . 'pCache.php',
    'AppKernel' => $baseDir . 'pKernel.php',
...
);
```
It cuts few more bytes from file path due to incorrect encoding in file path string.

## Where is the problem
1. Composer generates elements for **autoload_classmap.php** in [AutoloadGenerator](https://github.com/composer/composer/blob/master/src/Composer/Autoload/AutoloadGenerator.php#L271).
2. It calls [`AutoloadGenerator::addClassMapCode`](https://github.com/composer/composer/blob/master/src/Composer/Autoload/AutoloadGenerator.php#L259) to generate list.
3. It calls [`AutoloadGenerator::generateClassMap`](https://github.com/composer/composer/blob/master/src/Composer/Autoload/AutoloadGenerator.php#L319) to scan for files.
4. It calls [`ClassMapGenerator::createMap`](https://github.com/composer/composer/blob/master/src/Composer/Autoload/AutoloadGenerator.php#L336) to scan directories and find classes.
5. In this method file path is [being compiled from `$cwd` and `$filePath`](https://github.com/composer/composer/blob/master/src/Composer/Autoload/ClassMapGenerator.php#L87). `$cwd` is result of `getcwd()`, that returns working dir with incorrect encoding.
6.  After 3 step it calls [`AutoloadGenerator::getPathCode`](https://github.com/composer/composer/blob/master/src/Composer/Autoload/AutoloadGenerator.php#L320) to generate valid php string for path.
7. It calls [`Filesystem::findShortestPath`](https://github.com/composer/composer/blob/master/src/Composer/Autoload/AutoloadGenerator.php#L530) to shorten file path.
8. It compares strings incorrectly in [`findShortestPath`](https://github.com/composer/composer/blob/master/src/Composer/Util/Filesystem.php#L322) that leads to cutting more bytes from `app/appKernel.php` than needed.

## After patch applying
Composer generates valid class map:

```php
<?php

return array(
    'AppCache' => $baseDir . '/app/AppCache.php',
    'AppKernel' => $baseDir . '/app/AppKernel.php',
...
);

```

### DIFF
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -75,7 +75,9 @@ class ClassMapGenerator
 
         $map = array();
         $filesystem = new Filesystem();
-        $cwd = getcwd();
+        // Do not remove realpath() call.
+        // Fixes incorrect encoding in working dir on Windows.
+        $cwd = realpath(getcwd());
 
         foreach ($path as $file) {
             $filePath = $file->getPathname();


Seldaek:
I can't reproduce this here.. if I dump stuff from within a directory called `Сергей` I get this:

```
"C:\??????" // getcwd()
bool(false) // realpath(getcwd())
```

So it doesn't seem like realpath really fixes it?

curry684:
Hmm to my knowledge `realpath` invokes OS-specific APIs directly to get the 'true' path to a folder, but it seems on Windows something is going weird.

It works here on my Ubuntu:
```
curry@ubuntu:/tmp$ mkdir Сергей
curry@ubuntu:/tmp$ php -r "echo realpath('./Сергей') . PHP_EOL;"
/tmp/Сергей
curry@ubuntu:/tmp$ 
```
And indeed fails on Win10:
```
C:\>mkdir Сергей
C:\>php -r "echo realpath('./Сергей') . PHP_EOL;"
C:\>
```
Could be a discrepancy between PHP and OS-level configuration?

staabm:
Php7 changed a lot on the win plattform regarding utf8 paths. I guess this might be highly php version dependent

Seldaek:
FWIW I tried this on php 7.1 / windows 10

wapmorgan:
On my win 10 x64, php 7.1 x64:
```
>php -r "echo getcwd().PHP_EOL.realpath(getcwd());"
C:\Users\������\Documents
C:\Users\Сергей\Documents
```

curry684:
It's not in the default_charset settings either, my Win10 is properly reporting UTF8:
```
C:\>php -r "var_dump(ini_get('default_charset'));"
Command line code:1:
string(5) "UTF-8"
```

wapmorgan:
Mine too:
``
>php -r "var_dump(ini_get('default_charset'));"
string(5) "UTF-8"
```

Seldaek:
Not sure what to do here.. It seems harmless to me, so if you say it helps for you I'm happy to merge but I'd also be happy to understand the root cause of this mess :/

curry684:
The merge itself is harmless by definition, in every conceivable case the added `realpath` call must provide a logically identical result to the `getcwd` call, with the only exception that `realpath` nulls out if the directory doesn't exist, which cannot happen for the cwd.

PR #6132, opened 2017-02-04
development dependency required

farhanwazir:
There is missing dependency for "require-dev", actually it is required for development of plugin.

### DIFF
--- a/doc/articles/custom-installers.md
+++ b/doc/articles/custom-installers.md
@@ -89,6 +89,30 @@ Example:
 }
 ```
 
+### Development Only
+
+When you are planning to develop a plugin then you require `composer/composer` package, it should not be in your published plugin. Published composer.json should looks same as above.
+
+```
+{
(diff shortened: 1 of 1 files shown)


curry684:
Why not a concise oneliner stating you should add "composer/composer" to your dev dependencies? Someone capable of writing a plugin should be able to take it from there...

farhanwazir:
Yes oneliner is a better idea, one more suggestion use SDK term for composer/composer package.
> You should add SDK `composer/composer` to your dev dependencies.

farhanwazir:
Oneliner is better and use SDK term for composer/composer package.

On Feb 5, 2017 4:36 PM, "Niels Keurentjes" <notifications@github.com> wrote:
>
> Why not a concise oneliner stating you should add "composer/composer" to
your dev dependencies? Someone capable of writing a plugin should be able
to take it from there...

Issue #6131, opened 2017-02-04
Inexplainable dependency conflict


Berdir:
Steps to reproduce:

https://github.com/Berdir/strict-dependency-bug

Output:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - berdir/strict-dependency-bug 1.0.0 requires webflo/drupal-core-strict 8.2.6 -> satisfiable by webflo/drupal-core-strict[8.2.6].
    - berdir/strict-dependency-bug 1.0.1 requires webflo/drupal-core-strict 8.2.6 -> satisfiable by webflo/drupal-core-strict[8.2.6].
    - Conclusion: don't install webflo/drupal-core-strict 8.2.6
    - Installation request for berdir/strict-dependency-bug ^1.0 -> satisfiable by berdir/strict-dependency-bug[1.0.0, 1.0.1].


Installation failed, reverting ./composer.json to its original content.
```

The problem is related to having more than one valid version in strict-dependency-bug, if you do "composer require berdir/strict-dependency-bug:1.0.1" instead, it installs fine.

The error doesn't really make sense, how can there be a conflict because of two identical versions?

drupal-core-strict is a package that aims to enforce the same dependencies as defined in Drupal's composer.lock. We've noticed that sometimes there are unexpected changes in untested dependencies versions, this is an idea to prevent that.

(That might or might not really be a good idea. I mostly sharing this in case there's a deeper depencency resolving problem here)

Any idea what's going on?

Seldaek:
Might be interesting for @naderman  to look at as it seems to happen with a minimal subset of packages, but can you please share the smallest possible composer.json you use to reproduce this?

Berdir:
Thanks for checking.

I can't try, but I suspect it only happens in combination with starting off with a drupal-project. I just tried the last two commands (adding the repo and requiring it) after initializing a new and empty composer.json and then it works fine.

I guess it is the combination of depending on drupal/core which has the same dependencies as drupal-core-strict, just not as strict. But testing that alone also doesn't really give the same. Whe I just run composer require drupal/core and then try to add mine then I do get conflicts, but they go away after a rm -rf composer.lock vendor/.

So as far as I see, it only happens when you actually start of with drupal-project.

webflo:
Thanks for helping. I did some testing a while ago over in https://gist.github.com/webflo/5d8a2310734a9089eb67ab5ec85ce1cd

[...]

PR #6139, opened 2017-02-06
Test case for #6131

webflo:
I think this is valid test case for #6131. Not 100% sure.

### DIFF
--- a/tests/Composer/Test/DependencyResolver/SolverTest.php
+++ b/tests/Composer/Test/DependencyResolver/SolverTest.php
@@ -836,6 +836,59 @@ class SolverTest extends TestCase
         ));
     }
 
+    public function testInstallMoreStrictVersion() {
+        $this->repo->addPackage($packageA = $this->getPackage('A', '2.0'));
+        $this->repo->addPackage($packageB = $this->getPackage('B', '2.0'));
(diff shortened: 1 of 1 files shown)


tests/Composer/Test/DependencyResolver/SolverTest.php
```diff
@@ -836,6 +836,59 @@ public function testInstallDevAlias()
...
+        $this->repoInstalled->addPackage($packageA);
+        $this->repoInstalled->addPackage($packageC1);
+
+        $this->reposComplete();
+
+        $this->request->install('B');
+
+        $this->checkSolverResult(array(
+            array('job' => 'update', 'from' => $packageC1, 'to' => $packageC2),
+            array('job' => 'install', 'package' => $packageB),
```
naderman: There is a third job generated here to remove packageA since you added it to repoInstalled but didn't request its installation. If you add a remove job here the test succeeds.

naderman:
The first test is correct and succeeds and does not test anything other tests don't already cover. The second test has a small bug which is why it fails. Otherwise it too would succeed and does not test anything not covered by other tests. The original issue this was for should be addressed in https://github.com/composer/composer/pull/7946

PR #6140, opened 2017-02-07
Add new reference for custom installer

farhanwazir:
This referenced plugin is much more flexible and customize able for any kind of packages. For detail you can see at [CPCInstaller](https://github.com/farhanwazir/cpcinstaller)


Seldaek:
Sorry but I don't see a good reason to promote custom installers there in the docs, especially as it's not really a best practice to have custom install paths and we already provide the composer/installers project for this.

Issue #604, opened 2012-04-20
Composer fails to install JMSSecurityExtraBundle because of a too long path length (ZipArchive Exception)


gnutix:
When trying to install vendors with composer.phar in the following directory (path length is **77 chars**) : 

_D:\Dorian\Dropbox\XAMPP Lite Portable PHP 5.4\htdocs\github\Symfony2Testing\_

the installation of the package "jms/security-extra-bundle" (1.1.0) fails with the following exception :

_ZipArchive::extractTo(): Full extraction path exceed MAXPATHLEN (260)_

Changing the path (renamed XAMPP folder) to :

_D:\Dorian\Dropbox\XAMPP PHP 5.4\htdocs\github\Symfony2Testing\_

fixed the problem.
Here is the complete log : https://gist.github.com/aa7e6e4207a63bc19f94

franz-josef-kaiser:
Same problem here

grisha87:
Occurs also on Windows. I saw that the packages are downloaded to the target vendor directory (so for example zendframework/zendframework1/[hash].zip . Prefixed with the working directory path, this exceeds the limit.

yellow1912:
I can confirm that I'm suffering from the same issue with phpdocumentor

vixriihi:
Also having same issue with phpdocumentor

RomanSyroeshko:
Same problem with phpdocumentor. Composer self-update didn't help.

extdevelopment:
@RomanSyroeshko
In my case `--prefer-source` helped, (MAXPATHLEN (260) error)

`php composer.phar global require "phpdocumentor/phpdocumentor:2.*" --prefer-source`

RomanSyroeshko:
The solution works for me too. Thanks, @extdevelopment!

said-jomedia:
I had the same issue with installing `phpdocumentor` but `--prefer-source` worked also for me.

kenylieou:
Same problem to me, is there any way to fix it without using --prefer-source

glen-84:
This is no longer an issue with PHP 7.1. 🎉

PR #6159, opened 2017-02-13
Fix unzip on win php56

gza:
Refactoring of unzip ZipArchive fallback.

* a solution to #604
* cleaner fallback mecanism between ZipArchive and SystemUnZip
* tests

### DIFF
--- a/src/Composer/Downloader/ZipDownloader.php
+++ b/src/Composer/Downloader/ZipDownloader.php
@@ -133,7 +133,7 @@ class ZipDownloader extends ArchiveDownloader
      * @param string $file File to extract
      * @param string $path Path where to extract file
      */
-    public function extract($file, $path)
+    protected function extract($file, $path)
     {
         $resultZipArchive = NULL;
         $resultUnzip = NULL;
--- a/src/Composer/Downloader/ZipDownloader.php
+++ b/src/Composer/Downloader/ZipDownloader.php
@@ -30,8 +30,10 @@ use ZipArchive;
 class ZipDownloader extends ArchiveDownloader
 {
     protected $process;
-    protected static $hasSystemUnzip;
-    protected static $hasZipArchive;
+    public static $hasSystemUnzip;
+    public static $hasZipArchive;
+    public static $isWindows;
+    private $zipArchiveObject;
 
     public function __construct(IOInterface $io, Config $config, EventDispatcher $eventDispatcher = null, Cache $cache = null, ProcessExecutor $process = null, RemoteFilesystem $rfs = null)
     {
@@ -53,6 +55,10 @@ class ZipDownloader extends ArchiveDownloader
             self::$hasZipArchive = class_exists('ZipArchive');
         }
 
+        if (null === self::$isWindows) {
+            self::$isWindows = Platform::isWindows();
+        }
+
         if (!self::$hasZipArchive && !self::$hasSystemUnzip) {
             // php.ini path is added to the error message to help users find the correct file
             $iniMessage = IniHelper::getMessage();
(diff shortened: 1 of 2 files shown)


src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,112 @@ public function download(PackageInterface $package, $path, $output = true)
...
         }
+        return new \RuntimeException($processError);
+    }
 
+    /**
+     * extract $file to $path with ZipArchive
+     *
+     * @param string $file File to extract
+     * @param string $path Path where to extract file
+     * @return bool True if succeed
```
stof: that's totally wrong. Your method does not return a boolean. It returns ``bool|\Exception`` (which is a very weird signature btw)

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,112 @@ public function download(PackageInterface $package, $path, $output = true)
...
     }
 
-    protected function extract($file, $path)
+    /**
+     * extract $file to $path with "unzip" command
+     *
+     * @param string $file      File to extract
+     * @param string $path      Path where to extract file
+     * @param bool $isFallback  If true it is called as a fallback and should not throw exception
+     * @return bool True if succeed
```
stof: wrong. It returns ``bool|\Exception|null``

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,112 @@ public function download(PackageInterface $package, $path, $output = true)
...
+     */
+    protected function extractWithSystemUnzip($file, $path, $isFallback)
```
stof: protected methods should go after public ones.

and btw, why are they protected ?

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+        return TRUE;
+    }
+
+    /**
+     * extract $file to $path
+     *
+     * @param string $file File to extract
+     * @param string $path Path where to extract file
+     */
+    public function extract($file, $path)
```
stof: why public ? This method should be protected, according to the parent requirement

src/Composer/Downloader/ZipDownloader.php
gza: I refactored a big "protected" function into 3 ones.
I didn't challenged that choice and created 3 protected functions.

src/Composer/Downloader/ZipDownloader.php
gza: ooops, protected was a problem during testing, but I worked around it.
I should not have commited this change.

thanks !

gza:
should I switch target branch to 1.3 ?

Seldaek:
I am not entirely sure what this fixes.. Right now on windows we use ZipArchive if present, and otherwise unzip if present. On linux/osx we use unzip if present, otherwise ZipArchive (the order is reversed because unzip is faster IIRC, but on windows it tends to create problems).

Your PR changes the order to always be ZipArchive if present then unzip. Why? How is that fixing anything?

gza:
Yes, when ZipArchive fails (like in #604), its falls back to unzip

Seldaek:
Yeah I can see that.. but I'm asking why / what you are trying to achieve with this PR? I don't see why I would merge it right now.

gza:
Ooops, excuse me, I didn't saw I was changing unzip/pecl priority on unix. I was focus on my windows problem.

I should restore original unix behavour by switching order priority depending on Platform::isWindows().
But, I need to implement fallback on ZipArchive errors.

Is it good for you ?

Seldaek:
Well no sorry but that still does not explain what your PR is fixing at all? I don't care about the code right now, I'm trying to understand the intent/goal of the PR.

gza:
this is workaround to an issue that appends with:
* PHP < 7.1
* Windows
* Composer packages with long pathes (like zendframework1)

Seldaek:
OK but where is the workaround? As far as I can tell, we currently try ZipArchive and then unzip on windows. And after your PR it does that too but for linux as well.. so what did you effectively change to fix that apart from refactoring the code a bit?

gza:
Original code does never fallback on windows, its just never does unzip on windows : 
https://github.com/composer/composer/blob/master/src/Composer/Downloader/ZipDownloader.php#L66

Seldaek:
Ah right ok yeah I guess I misread that the other day. So yes if you can update it to use unzip>ZipArchive on linux and the opposite on windows that would be great and then I think it's good to merge :)

gza:
thanks, I will do it tomorrow :)

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+    protected function extractWithZipArchive($file, $path)
+    {
         $zipArchive = new ZipArchive();
 
         if (true !== ($retval = $zipArchive->open($file))) {
-            throw new \UnexpectedValueException(rtrim($this->getErrorMessage($retval, $file)."\n".$processError), $retval);
+            return new \UnexpectedValueException(rtrim($this->getErrorMessage($retval, $file)."\n"), $retval);
+        }
+
+        $extractResult = FALSE;
```
povils: null, false, true.. is in lowercase

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
 
-        if (true !== $zipArchive->extractTo($path)) {
-            throw new \RuntimeException(rtrim("There was an error extracting the ZIP file, it is either corrupted or using an invalid format.\n".$processError));
+        if (true !== $extractResult) {
+            return new \RuntimeException(rtrim("There was an error extracting the ZIP file, it is either corrupted or using an invalid format.\n"));
         }
 
         $zipArchive->close();
+
+        return TRUE;
```
povils: null, false, true.. is in lowercase

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+
+    /**
+     * extract $file to $path
+     *
+     * @param string $file File to extract
+     * @param string $path Path where to extract file
+     */
+    protected function extract($file, $path)
+    {
+        $resultZipArchive = NULL;
```
povils: null, false, true.. is in lowercase

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+        $resultUnzip = NULL;
```
povils: null, false, true.. is in lowercase

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+
+        if ( self::$hasZipArchive ) {
+            // zip module is present
+            $resultZipArchive = $this->extractWithZipArchive($file, $path);
+            if ($resultZipArchive === TRUE) {
```
povils: null, false, true.. is in lowercase

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+                return;
+            }
+        }
+
+        if ( self::$hasSystemUnzip ) {
+            // we have unzip in the path
+            $isFallback=FALSE;
```
povils: null, false, true.. is in lowercase. ALso spaces

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+
+        if ( self::$hasZipArchive ) {
+            // zip module is present
+            $resultZipArchive = $this->extractWithZipArchive($file, $path);
+            if ($resultZipArchive === TRUE) {
+                return;
+            }
+        }
+
+        if ( self::$hasSystemUnzip ) {
```
povils: why spaces around?

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+     *
+     * @param string $file File to extract
+     * @param string $path Path where to extract file
+     */
+    protected function extract($file, $path)
+    {
+        $resultZipArchive = NULL;
+        $resultUnzip = NULL;
+
+        if ( self::$hasZipArchive ) {
```
povils: why these spaces?

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+
+        if ( self::$hasSystemUnzip ) {
+            // we have unzip in the path
+            $isFallback=FALSE;
+            if ( $resultZipArchive !== NULL) {
+                $this->io->writeError("\nUnzip using ZipArchive failed, trying with unzip");
+                $isFallback=TRUE;
+            };
+            $resultUnzip = $this->extractWithSystemUnzip($file, $path, $isFallback);
+            if ( $resultUnzip === TRUE ) {
```
povils: why these spaces?

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+            $resultZipArchive = $this->extractWithZipArchive($file, $path);
+            if ($resultZipArchive === TRUE) {
+                return;
+            }
+        }
+
+        if ( self::$hasSystemUnzip ) {
+            // we have unzip in the path
+            $isFallback=FALSE;
+            if ( $resultZipArchive !== NULL) {
```
povils: why these spaces?

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -59,42 +64,111 @@ public function download(PackageInterface $package, $path, $output = true)
...
+                $this->io->writeError("\nUnzip using ZipArchive failed, trying with unzip");
+                $isFallback=TRUE;
```
povils: null, false, true.. is in lowercase

src/Composer/Downloader/ZipDownloader.php
povils: why not just 
```php
if ($resultUnzip) {
```

gza:
@povils : I'm sorry I will not respond to each of your comments, but I have noted to run some phpcs-fixer on the final code. Thanks

btw: if ($resultUnzip) is not equivalent to if ( $resultUnzip === TRUE )

Seldaek:
Yeah the minor CS considerations are not a huge deal really, I run php-cs-fixer every now and then to clean things up myself. And commenting on every single one just spams people needlessly, please don't do this @povils.

povils:
@gza yes it's not equivalent if $resultUnzip is not strictly boolean
@Seldaek sorry bad habits :)

Regards

gza:
Well, I think it's ready ( and CS fixed ;-) )

Seldaek:
Thanks :)

PR #6174, opened 2017-02-17
Allow packages using PathDownloader to install to source directory.

david-garcia-garcia:
There's prefectly valid use cases where you might want a package to stay where it is when using repository type path.

I cannot imagine why this was limited in code.

I.e. hard forking remote repositories for custom fixes/improvements. With this, you could simply copy the package to a folder in your application, mess a little with installer paths and project types, and all dependency management will work as expected.

The symlink support is supposed to do this, but adds fragility to CI, so having this hard wired to the source directory might be a good option.

Proposed changes simply remove the hard coded constraint.


stof:
This will cause issues when asking to remove the dependency later, because it will remove the source of the path repo too. This is why it forbids this case.

david-garcia-garcia:
> This will cause issues when asking to remove the dependency later, because it will remove the source of the path repo too. This is why it forbids this case.

Could work on a fix for that.

> The symlink support is supposed to do this, but adds fragility to CI, so having this hard wired to the source directory might be a good option.

I'd rather use symlinks, the problem is that composer on windows uses junctions instead of symlinks, and junctions work with absolute paths. Meaning that I cannot produce an "artifact" out of whatever has been built by composer using junctions because if that is published to another environment the junctions will break.

Maybe efforts will be better spent trying to make composer also use symlinks on windows instead of junctions.

stof:
a solution could be to *try* using symlinks on Windows, and use junctions only as a fallback. The symlink support is available on windows in PHP, but it requires to run the process with a priviledge not granted by default to non-admin users on Windows (which is why we cannot always use symlinks).

david-garcia-garcia:
Added proper support for symlinks on windows. Yet, it would be very convenient to support a package destination folder be same as source. Will be working on that.

david-garcia-garcia:
Fails aside, having some feedback on the strategy used to "lock down" packages installed to their own source directory will be appreciated. Just to continue on the same line or work or look for another solution.

david-garcia-garcia:
Finally found a more simple approach.

Note that another commenly used composer based project (composer/installers) needs a fix so that it won't remove packages installed from their sources:

https://github.com/composer/installers/pull/348

src/Composer/Downloader/FileDownloader.php
```diff
@@ -266,4 +274,18 @@ private function getCacheKey(PackageInterface $package, $processedUrl)
...
+
+    /**
+     * Check if a package destination and source are the same.
+     *
+     * @param PackageInterface $package
+     * @param mixed $path
+     * @throws \RuntimeException
+     * @return boolean
+     */
+    public function destinationIsSource(PackageInterface $package, $path) {
```
stof: does it need to be public ?

src/Composer/Downloader/FileDownloader.php
```diff
@@ -266,4 +274,18 @@ private function getCacheKey(PackageInterface $package, $processedUrl)
 
         return $package->getName().'/'.$cacheKey.'.'.$package->getDistType();
     }
+
+    /**
+     * Check if a package destination and source are the same.
+     *
+     * @param PackageInterface $package
+     * @param mixed $path
```
stof: this is a string, not a mixed argument

src/Composer/Downloader/FileDownloader.php
```diff
@@ -266,4 +274,18 @@ private function getCacheKey(PackageInterface $package, $processedUrl)
...
+     * @throws \RuntimeException
```
stof: it does not throw

curry684:
> There's prefectly valid use cases where you might want a package to stay where it is when using repository type path.
> I cannot imagine why this was limited in code.

Because the Symfony Filesystem component, which is used all over Composer to manage directories and folders, is fully aware of symlinks, and fundamentally tested to treat them correctly. It is however blissfully unaware of the entire concept of junctions, and by default will happily recursively delete through them, wiping your entire system if need be.

By the time we added support for NTFS junctions there was already so much Composer code that we (or rather I) considered it impossible to ever test for every conceivable situation in which a package might be (partially) deleted, upgraded or overwritten, nor guarantee the safety of future changes in this regard, and concluded that it was infeasible to ever guarantee that in *no situation whatsoever would we accidentally delete your work in progress sources*. Hence it was decided to block the scenario completely, as it *has no added value whatsoever anyway*.

> I.e. hard forking remote repositories for custom fixes/improvements. With this, you could simply copy the package to a folder in your application, mess a little with installer paths and project types, and all dependency management will work as expected.

Just fork it to a different folder and let Composer symlink it from there. Everything will work as expected without breaking well considered limitations that impose no practical problems.

david-garcia-garcia:
> Just fork it to a different folder and let Composer symlink it from there.

Yes that was my first thought, so i fixed symlinks on windows (to support relative paths) here:

https://github.com/composer/composer/pull/6213.

But the moment I started using it it broke everywhere. Windows explorer itself is not properly aware of symlinks when copying, some zipping utils are aware of them - others not, AppVeyor's artifact builder messed up big time duplicating the files, my IDE was confused when running XDebug and trying to edit files, etc.. it turned out to be a total mess.

> has no added value whatsoever anyway.

The added value is that symlinks (at least on windows) are too "fragile" (see comment above). I believe that this might also be a problem on *nix system depending on what tools you use to manipulate the files and how "aware" they are about symlinks.

Having the package in it's source is way more robust and less fragile than using symlinks. I really really wished I could use symlinks for this, but it breaks elsewhere and they are too fragile to use them reliably.

> that in no situation whatsoever would we accidentally delete your work in progress sources.

Fair point. That is something that composer should "guarantee". But this can be introduced experimentally (because it was previously not possible) and only be allowed when using a flag or configuration setting (would have to expand the patch for that).

I tested different use cases with this patch (removing packages, changing the way they are pulled-in, ..) just to ensure that - precisely - sources where not destroyed. And this sources will - of course - be commited to svc so it would be easy to notice if composer messed up with them.

curry684:
Testing only gets you so far, future compatibility is also a huge concern here. Take for example https://github.com/composer/composer/issues/6223, a regularly recurring feature request to clean up certain folders inside installed packages. That specific request is going to get denied this time, but it may get added in 2018. Or 2019. And will then happily delete all your own test folders because its implementation accidentally didn't notice that, despite no symlink or junction being present, it was cleaning up the original source files.

Seldaek:
We tried merging https://github.com/composer/composer/pull/5974 a while back and had to revert it.. I don't know if this is exactly the same approach, but to be honest this is a bad idea that is too risky. It's not like it's a 100% required feature, there are other ways to do this, so I hope you will find a way to work without it.

david-garcia-garcia:
@Seldaek This patch (differently to the one in #5974) prevents package from being deleted during removes.

Would you consider this feature if I **add a required configuration flag (explicit) to be enabled before allowing packages to be installed from source**? Only and only if that setting has been explictly enabled, installation from source would be allowed.

I strongly believe there is a very valid use case for having this in composer, it would be useful to get this in, even if it is an experimental enable under you own risk feature.

Seldaek:
Sorry but nope.. I'd rather not have yet more options, especially if they enable dangerous stuff, and only are marginally useful to a few users. More code = more stuff to maintain for us..

david-garcia-garcia:
That's fine. We already have this setup using custom scripts. Just wanted to see if it could make it into core as it is getting more and more frequent to get inquiries from customers on how to locally fork crappy contributed packages/modules, and the way composer core solves this right now is simply not working/flawed. Also for packages you develop locally and do not plan to contribute at all, but still need to be managed through composer.

Issue #6178, opened 2017-02-17
Root package version not being calculated for SVN package


jbuncle:
When the root package version is being calculated for an SVN package, it is being `parsed as 1.0.0` even thought the repository uses standard /tags layout.

E.g. when running `show -s` from an SVN composer project:
`versions : * No version set (parsed as 1.0.0)`

This results in `Your requirements could not be resolved to an installable set of packages.` due to being unable to resolve a cyclic dependencies when using minimum version requirements (as the root package version is `parsed as 1.0.0`).

jbuncle:
Opened Pull Request with fix https://github.com/composer/composer/pull/6179

PR #6179, opened 2017-02-17
Root package version not being calculated for SVN package #6178

jbuncle:
Updated VersionGuesser to check if we actually have a version in the returned array (fixes bug originating from f6899e5 when array was first introduced) - as SVN Versions not being guessed as it would previously stop at Git check.


Seldaek:
Thanks! Merged in 1.3 and master

PR #6184, opened 2017-02-20
Disable static autoloader on xcache.cacher = On

EugeniyPetrov:
Hello,

Composer static autoloader works wrong when xcache cacher is enabled. This is due to bug in xcache https://xcache.lighttpd.net/ticket/348 which was fixed 2 years ago but still in trunk. Looks like xcache is not maintained more so there is no chance to get this fix released.

To reproduce the problem:
- install xcache with xcache.cacher = On
- create folder test1 and run `composer require symfony/console && composer update`. Then create index.php with `require_once 'vendor/autoload.php';`
- copy test1 to test2 (autoloader_static.php in test2 must have the same content as in test1 because xcache acts on file checksums)
- run test1/index.php under apache + mod_php
- remove test1/ and run test2/index.php. Fatal error raised:
`Warning: require(/var/www/html2/test1/vendor/composer/../symfony/polyfill-mbstring/bootstrap.php): failed to open stream: No such file or directory in /var/www/html2/test2/vendor/composer/autoload_real.php on line 66`

Usually `xcache_clear_cache(XC_TYPE_PHP, 0);` helps to fix the problem but in my production it does not. I suppose that due to high concurency old values cached again in xcache before the cache is cleared, and therefore there must not be any traffic before cache clearing.

This is not actually a composer's bug, but because of many users that may use composer together with xcache, this may affect them. I propose to disable static autoloader if xcache caching is enabled.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -604,7 +604,7 @@ INCLUDE_PATH;
         }
 
         $file .= <<<STATIC_INIT
-        \$useStaticLoader = PHP_VERSION_ID >= $staticPhpVersion && !defined('HHVM_VERSION') && (!function_exists('zend_loader_file_encoded') || !zend_loader_file_encoded());
+        \$useStaticLoader = PHP_VERSION_ID >= $staticPhpVersion && !defined('HHVM_VERSION') && (!function_exists('zend_loader_file_encoded') || !zend_loader_file_encoded()) && ini_get('xcache.cacher') !== '1';
         if (\$useStaticLoader) {
             require_once __DIR__ . '/autoload_static.php';
 
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -604,7 +604,7 @@ INCLUDE_PATH;
         }
 
         $file .= <<<STATIC_INIT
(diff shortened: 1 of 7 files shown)


staabm:
@EugeniyPetrov could you give any insight why you use xcache instead of opcache or apc or similar?

EugeniyPetrov:
@staabm no good reasons, just one mistake some time ago ) We are going to migrate to opcache.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -604,7 +604,7 @@ public static function getLoader()
         }
 
         $file .= <<<STATIC_INIT
-        \$useStaticLoader = PHP_VERSION_ID >= $staticPhpVersion && !defined('HHVM_VERSION') && (!function_exists('zend_loader_file_encoded') || !zend_loader_file_encoded());
+        \$useStaticLoader = PHP_VERSION_ID >= $staticPhpVersion && !defined('HHVM_VERSION') && (!function_exists('zend_loader_file_encoded') || !zend_loader_file_encoded()) && ini_get('xcache.cacher') !== '1';
```
stof: you should check for XCache being loaded before checking the php.ini IMO.

EugeniyPetrov: Make sense, thanks

Seldaek:
I have to say I'm not too keen on adding a check in there that we will need to keep indefinitely just for one company relying on outdated software.

I would rather offer the following workaround which I believe should work. In your bootstrap file before including composer's autoload.php you could define this zend-guard function which will disable the static autoloader as well:

```php
function zend_loader_file_encoded() {
    return true;
}

```

PR #6199, opened 2017-02-23
Changed the query string parameter `sha` to `ref`

### LINKED ISSUE #6198: GitLab dist URL is wrong for GitLab API v3
The dist URL for the GitLab driver generated by `Composer\Repository\Vcs\GitLabDriver::getDist()` uses `$url = $this->getApiUrl().'/repository/archive.zip?sha='.$identifier;` The correct query string parameter is `ref` as opposed to `sha`. Fixing this would enable GitLab repositories to be downloaded as zip packages  instead of being cloned.

This parameters is not documented correctly on the GitLab [docs](https://docs.gitlab.com/ce/api/repositories.html#get-file-archive) as well. A look into the source code for the responsible [controller](https://gitlab.com/gitlab-org/gitlab-ce/blob/v8.17.0/app/controllers/projects/repositories_controller.rb) lets us know is has to be `ref`. An [issue](https://gitlab.com/gitlab-com/gitlab-docs/issues/80) regarding this 'bug' has been submitted.

sjadema:
Updated the unit test to reflect the change
Fixes #6198

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -156,7 +156,7 @@ class GitLabDriver extends VcsDriver
      */
     public function getDist($identifier)
     {
-        $url = $this->getApiUrl().'/repository/archive.zip?sha='.$identifier;
+        $url = $this->getApiUrl().'/repository/archive.zip?ref='.$identifier;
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Merged in 1.3/master, thanks

PR #6168, opened 2017-02-16
Improve memory usage resolving dependencies 

rubenrua:
It is known that composer update takes a lot of memory: #5915, #5902, #5847, #5879

I am playing with a profiler (@blackfireio) to make a demo in my local PHP meetup (@phpvigo) and I found out a way to use less memory. These are my first tests:

* Private project using PHP 5.6: 
  * Memory: from 1.31GB to 1.07GB 
  * Wall Time: from 2min 8s to 1min 33s

* symfony-demo using PHP 7.1 in my old mac book:
  * Memory: from 667MB to 523MB 
  * Wall Time: from  5min 29s to 5min 28s

Not use an array inside conflict rules is this improvement main idea:

```php
<?php
//Memory 38MB
gc_collect_cycles();
gc_disable();


class Rule
{
    public $literals;

    public function __construct(array $literals)
    {
        $this->literals = $literals;
    }
}

$rules = array();

$i = 0;
while ($i<80000){ //
    $i++;

    $array = array(-$i, $i);
    $rule = new Rule($array);
    $rules[] = $rule;
}
```

```php
<?php
//Memory 11.1MB
gc_collect_cycles();
gc_disable();


class ConflictRule
{
    public $literal1;
    public $literal2;

    public function __construct($literal1, $literal2)
    {
        $this->literal1 = $literal1;
        $this->literal2 = $literal2;
    }
}

$rules = array();

$i = 0;
while ($i<80000){ //
    $i++;

    $rule = new ConflictRule(-$i, $i);
    $rules[] = $rule;
}
```

This is a WIP pull request to get feedback. I have to:

 - [x] Test with more public projects
 - [x] Check and add more unit test
 - [x] Doc

PR #6200, opened 2017-02-23
Fixes with comments from the community

rubenrua:
New commit with the comments of @vlakoff and @Pierstoval in pull request #6168.

### DIFF
--- a/src/Composer/DependencyResolver/RuleSet.php
+++ b/src/Composer/DependencyResolver/RuleSet.php
@@ -60,17 +60,19 @@ class RuleSet implements \IteratorAggregate, \Countable
 
         $hash = $rule->getHash();
 
+        $rulesByHash = isset($this->rulesByHash[$hash]) ? $this->rulesByHash[$hash] : null;
+        $isArray = $rulesByHash && is_array($rulesByHash);
+
         // Do not add if rule already exists
-        if (isset($this->rulesByHash[$hash])) {
-            $potentialDuplicates = $this->rulesByHash[$hash];
-            if (is_array($potentialDuplicates)) {
-                foreach ($potentialDuplicates as $potentialDuplicate) {
+        if (null !== $rulesByHash) {
+            if ($isArray) {
+                foreach ($rulesByHash as $potentialDuplicate) {
                     if ($rule->equals($potentialDuplicate)) {
                         return;
                     }
                 }
             } else {
-                if ($rule->equals($potentialDuplicates)) {
+                if ($rule->equals($rulesByHash)) {
                     return;
                 }
             }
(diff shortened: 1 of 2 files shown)


naderman:
I don't think this makes any practical difference, isset and is_array are not expensive. Please provide benchmarks for performance optimizations.

vlakoff:
Indeed, my code comment was just a thought. It would need proper benchmarking, and probably would be noticeable only with *many* method calls.

src/Composer/DependencyResolver/RuleSet.php
```diff
@@ -60,17 +60,19 @@ public function add(Rule $rule, $type)
...
 
+        $rulesByHash = isset($this->rulesByHash[$hash]) ? $this->rulesByHash[$hash] : null;
+        $isArray = $rulesByHash && is_array($rulesByHash);
+
         // Do not add if rule already exists
-        if (isset($this->rulesByHash[$hash])) {
-            $potentialDuplicates = $this->rulesByHash[$hash];
-            if (is_array($potentialDuplicates)) {
-                foreach ($potentialDuplicates as $potentialDuplicate) {
+        if ($rulesByHash) {
```
alcohol: Why not `null !== $rulesbyHash`? No need for type coercion then.

src/Composer/DependencyResolver/RuleSet.php
```diff
@@ -86,12 +88,12 @@ public function add(Rule $rule, $type)
 
         $this->nextRuleId++;
 
-        if (!isset($this->rulesByHash[$hash])) {
+        if (!$rulesByHash) {
```
alcohol: Same as previous.

src/Composer/DependencyResolver/RuleSet.php
rubenrua: Done

src/Composer/DependencyResolver/RuleSet.php
rubenrua: Done

rubenrua:
@naderman The goal of this pull request is not performance optimization. IMHO the changes are good and they try to show the composer code style (very optimized code).

> You can optimise for execution speed. You can optimise for space. But the most precious thing you should optimise for is understandability

Feel free to close it. :)

naderman:
That's really not the composer code style. Composer is optimized yes, in places where it is needed. But otherwise we try to make code as readable as possible, and this PR is not optimization like you said. There's nothing gained from any of these changes. (The triple equal is fine by itself, but has no practical effect either since both sides are guaranteed to be integers)

So thanks, but let's try and focus on actually improving the performance as in the original PR this was based on :-)

Issue #6193, opened 2017-02-23
PEAR installer doesn't work


oburk:
My `composer.json`:

```
	"require": {
		"pear-pear.php.net/Calendar": "0.5.5",
		"pear-pear.php.net/Date_Holidays_Germany": "0.1.2",
	}
	
	"repositories": [
		{
			"type": "pear",
			"url": "https://pear.php.net/"
		}
	]

```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
require.zendframework/zenddiagnostics : unbound version constraints (dev-master) should be avoided
require.pear-pear.php.net/XML_Util : exact version constraints (1.4.2) should be avoided if the package follows semantic versioning
require.adodb/adodb-php : exact version constraints (v5.19) should be avoided if the package follows semantic versioning
require.phpoffice/phpexcel : exact version constraints (1.7.9) should be avoided if the package follows semantic versioning
require.apache/log4php : exact version constraints (2.3.0) should be avoided if the package follows semantic versioning
require.ezyang/htmlpurifier : exact version constraints (4.7.0) should be avoided if the package follows semantic versioning
require.fpdf/fpdf : exact version constraints (1.81) should be avoided if the package follows semantic versioning
require.phpseclib/phpseclib : exact version constraints (2.0.3) should be avoided if the package follows semantic versioning
require.muz/mive : unbound version constraints (>=1.0.5) should be avoided
require.elasticsearch/elasticsearch : exact version constraints (1.0.x-dev) should be avoided if the package follows semantic versioning
require.pimple/pimple : unbound version constraints (>=v2.0.0) should be avoided
require.muz/docx-api : unbound version constraints (dev-trunk) should be avoided
require.simplesamlphp/simplesamlphp : exact version constraints (1.14.0) should be avoided if the package follows semantic versioning
require.bower-asset/bootstrap : exact version constraints (3.3.4) should be avoided if the package follows semantic versioning
require.bower-asset/respond-minmax : exact version constraints (1.4.2) should be avoided if the package follows semantic versioning
require.bower-asset/html5shiv : exact version constraints (3.7.2) should be avoided if the package follows semantic versioning
require.bower-asset/retina.js : exact version constraints (1.3.0) should be avoided if the package follows semantic versioning
require.bower-asset/mjolnic-bootstrap-colorpicker : exact version constraints (2.1.2) should be avoided if the package follows semantic versioning
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
[...]

PR #6202, opened 2017-02-24
Fix problem with pear.php.net

janczer:
If xml have not index `channel` use `uri`.
For issue #6193

### DIFF
--- a/src/Composer/Repository/Pear/PackageDependencyParser.php
+++ b/src/Composer/Repository/Pear/PackageDependencyParser.php
@@ -228,7 +228,11 @@ class PackageDependencyParser
         }
 
         $result = array();
-        foreach ($depItem as $subDepItem) {
+        foreach ($depItem as $key => $subDepItem) {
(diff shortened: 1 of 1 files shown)


src/Composer/Repository/Pear/PackageDependencyParser.php
```diff
@@ -228,7 +228,11 @@ private function buildDepPackageConstraints($depItem, $depType)
         }
 
         $result = array();
-        foreach ($depItem as $subDepItem) {
+        foreach ($depItem as $key => $subDepItem) {
+            if (!array_key_exists('channel', $subDepItem)) {
+                $depItem[$key]['channel'] = $depItem[$key]['uri'];
```
stof: there is no need to alter this one, as it is never used again

janczer: Thank for review. I change it.

Seldaek:
Merged in 1.3/master, thanks!

PR #6208, opened 2017-02-26
Improve performance on psr4 autoload file finding

Vedrillan:
This improves psr4 autoloading performance on projects that uses a lot of packages part of the the same vendor.

In my tests on a Magento2 project (~180 projects under Magento\\*) it is around 3 times faster with Composer\Autoload\ClassLoader::findFileWithExtension taking from ~140ms to ~30ms.

This method should in the end scale better on big projects.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -374,9 +374,13 @@ class ClassLoader
 
         $first = $class[0];
         if (isset($this->prefixLengthsPsr4[$first])) {
-            foreach ($this->prefixLengthsPsr4[$first] as $prefix => $length) {
-                if (0 === strpos($class, $prefix)) {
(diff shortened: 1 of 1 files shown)


Seldaek:
You really should use --optimize-autoloader and possibly --classmap-authoritative (if you can) though, instead of relying on psr4 autoload in production.. Could you check performance with that?

Anyway thanks for the patch. I'll have a more in-depth look later to see if there is no other case where it'd harm perf.

Vedrillan:
I was on my dev environment and findFileWithExtension was showing red in my blackfire graphs, so I decided to get it out of the way as I was distracted by it !

I did the test with --optimize-autoloader, which we use in production, and findFileWithExtension went down to 50ms, much better than the 150ms I had before, but it seems slightly slower than the change I am pushing here x)

I hope to hear soon about your additional tests.

PR #6213, opened 2017-02-27
Fix support for symlinks on Windows.

david-garcia-garcia:
Fix for discussion on:

https://github.com/composer/composer/pull/6174

The symlink function from PHP in Windows is totally broken when using relative paths. This pull requests uses several fallback mechanisms before using a junction.

The problems with junctions is that they are not portable, making artifacts built with composer unusable for deployments.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -113,13 +113,13 @@ class PathDownloader extends FileDownloader implements VcsCapableDownloaderInter
             if (!$linked && Platform::isWindows()) {
                 // Use command linke symlinking as a fallback.
                 try {
-                    $return_var = 0;
+                    $exitCode = 0;
                     $output = array();
                     // Without this cleanup for backslashes symlink will
                     // be created, but completely broken.
                     $shortestPath = str_replace('/', '\\', $shortestPath);
-                    exec(sprintf('mklink /d %s %s', escapeshellarg($absolutePath), escapeshellarg($shortestPath)), $ouptput, $return_var);
-                    $linked = $return_var == 0;
+                    exec(sprintf('mklink /d %s %s', escapeshellarg($absolutePath), escapeshellarg($shortestPath)), $ouptput, $exitCode);
+                    $linked = $exitCode == 0;
                 }
                 catch (\Exception $e) {
                     $this->io->writeError($e->getMessage(), false);
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -111,7 +111,7 @@ class PathDownloader extends FileDownloader implements VcsCapableDownloaderInter
             }
 
             if (!$linked && Platform::isWindows()) {
-                // Use command linke symlinking as a fallback.
+                // Use command-line symlinking as a fallback.
                 try {
                     $exitCode = 0;
                     $output = array();
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -37,6 +37,7 @@ class PathDownloader extends FileDownloader implements VcsCapableDownloaderInter
      */
     public function download(PackageInterface $package, $path, $output = true)
     {
+        // Make sure that the source path for the package exists.
         $url = $package->getDistUrl();
         $realUrl = realpath($url);
         if (false === $realUrl || !file_exists($realUrl) || !is_dir($realUrl)) {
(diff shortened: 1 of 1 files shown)


src/Composer/Downloader/PathDownloader.php
```diff
@@ -76,40 +79,79 @@ public function download(PackageInterface $package, $path, $output = true)
         $this->filesystem->removeDirectory($path);
 
         if ($output) {
-            $this->io->writeError(sprintf(
+            $this->io->write(sprintf(
```
stof: this is a no-go. Feedback given to humans goes to the second pipe (named STDERR in PHP), not the first one

src/Composer/Downloader/PathDownloader.php
```diff
@@ -76,40 +79,79 @@ public function download(PackageInterface $package, $path, $output = true)
...
-                if (Platform::isWindows()) {
+                // Creating symlinks on windows through PHP's symlink function
+                // is totally broken with relative paths.
+                // @see https://github.com/php/php-src/pull/1243
+                // @see https://bugs.php.net/bug.php?id=69473
+                $fileSystem->symlink($shortestPath, $path);
+                $this->io->writeError(sprintf(' Symlinked from %s', $url), false);
+                $linked = true;
+            }
+            catch (IOException $e) {
```
stof: should be on the previous line

src/Composer/Downloader/PathDownloader.php
```diff
@@ -76,40 +79,79 @@ public function download(PackageInterface $package, $path, $output = true)
...
+                $this->io->writeError($e->getMessage(), false);
+            }
+
+            if (!$linked && Platform::isWindows()) {
+                // Use command linke symlinking as a fallback.
+                try {
+                    $return_var = 0;
```
stof: Should use camelCase.

And you don't need to initialize it here, as it is a by-ref variable.

And ``$exitCode`` is a better name for what it does

src/Composer/Downloader/PathDownloader.php
david-garcia-garcia: Fixed. Had no clue... having "error" there was really confusing me.

src/Composer/Downloader/PathDownloader.php
david-garcia-garcia: Fixed. My IDE is using coding standards from other projects...

src/Composer/Downloader/PathDownloader.php
david-garcia-garcia: return_var is from PHP's exec documentation itself.

By ref does not guarantee that these variables are really set in the callee. I only do that to get the IDE to give them a type automatically. Should do no harm.

Seldaek: The name $return_var is up to you entirely, it's just an example in the docs. The function will set the value by ref no matter what the variable is called.

david-garcia-garcia:
Just as a final thought. Even though this patch fixes symlinks on Windows, I would NOT recommend at all their usage.

I.e. if you copy a symlink using the windows explorer, it will be ignored and copied as an empty folder. Or some tools (such as AppVeyor's artifact packages) will create Zip's where the files are duplicated and the symlinks removed.

All-in-all, the best approach is to have composer use the packages in their original local path:

https://github.com/composer/composer/pull/6174

src/Composer/Downloader/PathDownloader.php
david-garcia-garcia: > The function will set the value by ref no matter what the variable is called.

Sure. It was just a mater of consistency. Whatever, already changed it.

curry684:
> The problems with junctions is that they are not portable, making artifacts built with composer unusable for deployments.

I'm not really sure what issue you are describing with this, but the simple solution would just be to disable symlinking when building cross-platform artifacts.

david-garcia-garcia:
> I'm not really sure what issue you are describing with this, but the simple solution would just be to disable symlinking when building cross-platform artifacts.

With "portable" I am not talking about moving this from windows to linux (or the other way round). I mean that Junctions always use **absolute paths.** So if I build an artifact using composer on one machine, CI system, whatever, then I will only be able to use it on another system that has exactly the same folder structure. It cannot be moved around. Plus all the other issues with symlinks themselves.

> the simple solution would just be to disable symlinking when building cross-platform artifacts.

This means that these artifacts will have duplicate code, not a big deal, but can be an issue on big projects.

src/Composer/Downloader/PathDownloader.php
```diff
@@ -84,32 +87,70 @@ public function download(PackageInterface $package, $path, $output = true)
...
+                // @see https://bugs.php.net/bug.php?id=69473
+                $fileSystem->symlink($shortestPath, $path);
+                $this->io->writeError(sprintf(' Symlinked from %s', $url), false);
+                $linked = true;
+            } catch (IOException $e) {
+                $this->io->writeError($e->getMessage(), false);
+            }
+
+            if (!$linked && Platform::isWindows()) {
+                // Use command linke symlinking as a fallback.
```
alcohol: s/linke/line/

Seldaek:
As far as I understand, this doesn't fix much as symlinks or junctions are equally broken, and symlinks are generally not working anyway since most people don't run elevated shells, so I'd rather not merge it and risk a regression if it's not improving things much. If I'm mistaken let me know and we can revisit the issue.

PR #6224, opened 2017-03-03
Remove useless substr calls in the ClassLoader

stof:
When multiple directories are registered for the same PSR-4 prefix, the PSR-4 subpath does not need to be recomputed for each directory. It only depends on the prefix.


stof:
hmm, this could even be merged in 1.3.x IMO

Seldaek:
IIRC this was tried and it turned out it wasn't faster because of the extra allocation as most cases only have one iteration in the foreach().. Actual benchmarks would be valuable. Then again micro-optimizing this when we have the static classmap that's anyway faster isn't very important.

Seldaek:
Closing as it's been rewritten anyway by #6208 and that substr issue is gone AFAICT.

PR #6270, opened 2017-03-16
Fix Init command to accept author names with Unicode combining marks

twoixter:
The Init command was not accepting my author name. I have my default git `user.name` formed in a way that **Pérez** was not using the latin small "e" with acute (U+00E9), but instead using the combining diacritic mark for acute accent (U+0301).


Seldaek:
Thanks

Issue #6245, opened 2017-03-10
[Composer 1.4.0] Receiving JsonValidationException when disabling packagist in an object-based config


jkribeiro:
I have the following config:

```    
"repositories": [
        {"type": "composer", "url": "https://myprivatepackagist.com"},
        {"packagist": false}
    ]
```

My test build add other repo to the list, running something like:
`composer config repositories.foo vcs https://github.com/foo/bar`

After this command, my composer config is set to:
```
    "repositories": {
        "0": {
            "type": "composer",
            "url": "https://myprivatepackagist.com"
        },
        "1": {
            "packagist": false
        },
        "foo": {
            "type": "vcs",
            "url": "https://github.com/foo/bar"
        }
    }
```

Running any type of composer command like `composer config --list` I'm receiving the same error:

```
  [Composer\Json\JsonValidationException]                                
  "./composer.json" does not match the expected JSON schema:             
   - repositories[1] : Object value found, but a boolean is required     
   - repositories[1] : Does not have a value in the enumeration [false]  
   - repositories[1].type : The property type is required                
   - repositories[1].package : The property package is required          
   - repositories[1].url : The property url is required                  
   - repositories[1] : Failed to match exactly one schema     
```

alcohol:
Yeah, our tooling isn't exactly built to handle weird edge-cases where users mix named and unnamed repositories, and manual and automatic modification of their composer.json.

Feel free to provide a fix.

PR #6271, opened 2017-03-16
Fix bug #6245

swumao:


swumao:
I see the bug #6245 has the 'Easy Pick' label and try to fix it.

And the document [disabling-packagist-org](https://getcomposer.org/doc/05-repositories.md#disabling-packagist-org) sepcify key "packagist.org" to disable the packagist.org.

So I Add a special definition "packagist-repository" to fix the bug

Seldaek:
Thanks for the work but unfortunately that's not really the fix I had in mind.. This just makes an invalid json valid instead of avoiding the creation of the invalid json in the first place :) Will commit a fix in a minute.

Issue #6098, opened 2017-01-24
Invalid config.platform.php not noticed by diagnose


caiquecastro:
If a composer.json contains an invalid php version constraint:

```
{
    ...
    "config": {
        "platform": {
            "php": "^7.0"
        }
    },
    ...
}
```
It's not alerted by the diagnose command, but when i give another command, like `composer update`, it throwns an UnexpectedValueException with the message:  Invalid version string "^7.0".

Let me know if it's the desired behavior.

alcohol:
The validation is done according to the schema primarily, which simply expects a string.

https://github.com/composer/composer/blob/master/res/composer-schema.json#L177-L183

Not sure if additional checking is done anywhere else. The exception is the desired behavior because a version is expected, not a version constraint. However, this issue is detected at runtime, and not sooner.

stof:
@alcohol we have a ConfigValidator implementing extra checks for ``composer validate``, on top of the schema validation: https://github.com/composer/composer/blob/master/src/Composer/Util/ConfigValidator.php
We could add a validation of ``config.platform`` there (ensuring that all versions can be parsed)

alcohol:
I know we have that. But to my knowledge it does not check for this.

Seldaek:
This could be added to the ValidatingArrayLoader, by checking that all values of the config > platform key pass a `$this->versionParser->normalize()` call without throwing, much like is done for validating the package's version.

Seldaek:
Fixed by #6281

PR #6281, opened 2017-03-21
Add config>platform checking

swumao:
Complete #6098

Now run `composer validate` will notice:

```
./composer.json is invalid, the following errors/warnings were found:
config.platform.php : invalid value (^7.0): Invalid version string "^7.0"
```


Seldaek:
Thanks :)

PR #6284, opened 2017-03-21
Improve memory usage a little by not creating new Link objects

Toflar:
I was able to shave off only a litte memory usage (about 2% on average) by this change.
At first I did not consider it to be worth discussing about because the improvements are not significant enough (well, a penny saved is a penny earned but still). However, I still decided to make the changes public because at the end of the day this is what open source is all about. Maybe someone can improve the impact even more based on my idea 😄 

So the idea behind it is that most `Rule` instances get a `$reasonData` assigned which mostly is an instance of `Link` or `CompletePackage`. While debugging a little, I noticed that even though a big amount of `Link` instances are actually of equal content, they are not the same (they do not share the same object hash). They are different objects because every `Package` creates its `Link` objects (`requires`,  `conflicts` etc.) when it is created. So the `Link` objects are not shared across memory.
Hence, I thought by making sure to not create them but instead reference to an already created one if it is the same should decrease memory usage. This seems to be true but I was expecting an even bigger impact. Now somebody can tell me maybe why the impact is not bigger or make it bigger based on my idea 😋

### DIFF
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -28,6 +28,7 @@ class ArrayLoader implements LoaderInterface
 {
     protected $versionParser;
     protected $loadOptions;
+    protected $linksCache = array();
 
     public function __construct(SemverVersionParser $parser = null, $loadOptions = false)
     {
@@ -241,7 +242,13 @@ class ArrayLoader implements LoaderInterface
                 $parsedConstraint = $this->versionParser->parseConstraints($constraint);
             }
 
-            $res[strtolower($target)] = new Link($source, $target, $parsedConstraint, $description, $constraint);
+            $key = $source . $target . $parsedConstraint . $description . $constraint;
+
+            if (isset($this->linksCache[$key])) {
+                $res[strtolower($target)] = $this->linksCache[$key];
+            } else {
+                $res[strtolower($target)] = $this->linksCache[$key] = new Link($source, $target, $parsedConstraint, $description, $constraint);
+            }
         }
 
         return $res;


staabm:
Could you provide a blackfire diff or similar?

Toflar:
Composer just creates so many objects that - I guess - it's hard to profile for blackfire.io 😄  So maybe you should just test it locally for your own projects and see for yourself for real results. I just checked the blackfire.io results and they are basically nonsense because the metrics differ on so many levels that only running multiple profiling sessions and taking the average would result in useful data (which is an unreasonable demand as one profile takes about 3.5 minutes for Composer ;-)).

src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -241,7 +242,13 @@ public function parseLinks($source, $sourceVersion, $description, $links)
                 $parsedConstraint = $this->versionParser->parseConstraints($constraint);
             }
 
-            $res[strtolower($target)] = new Link($source, $target, $parsedConstraint, $description, $constraint);
+            $key = $source . $target . $parsedConstraint . $description . $constraint;
+
+            if (isset($this->linksCache[$key])) {
```
ronfroy: it can be refactored to not assign `$res[strtolower($target)] = $this->linksCache[$key]` and use `else`

ronfroy:
```
if (!isset($this->linksCache[$key])) {
    $this->linksCache[$key] = new Link($source, $target, $parsedConstraint, $description, $constraint);
}

$res[strtolower($target)] = $this->linksCache[$key];
```

src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -28,6 +28,7 @@ class ArrayLoader implements LoaderInterface
 {
     protected $versionParser;
     protected $loadOptions;
+    protected $linksCache = array();
```
stof: I would make it private

src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -241,7 +242,13 @@ public function parseLinks($source, $sourceVersion, $description, $links)
                 $parsedConstraint = $this->versionParser->parseConstraints($constraint);
             }
 
-            $res[strtolower($target)] = new Link($source, $target, $parsedConstraint, $description, $constraint);
+            $key = $source . $target . $parsedConstraint . $description . $constraint;
```
stof: you need to use a delimiter between them. Otherwise, it is possible to create conflicts in the key with crafted things (`foo/bar` requiring `b/c` and `foo/ba` requiring `rb/c` for instance)

staabm:
I measured this change on a project of mine using
`blackfire run php ../tools/composer/bin/composer update --dry-run`.

before this patch it consumed 207-208MB, after this change its about 206-207MB.

I guess the change is not worth it.
(or maybe on a different workload)

naderman:
What is the CPU overhead created by doing this? Is the additional computation required worth the memory decrease?

naderman:
I suppose it's just a string concat and some extra hash lookups, so hopefully not really note-worthy

staabm:
please see https://blackfire.io/profiles/compare/8c956193-a16f-4744-9e72-eb9981c64a24/graph

the diff in the profile is mainly because of different libs beeing downloaded (as updated libraries arrived at packagist.org in between the profile runs)

searching within the profile for ArrayLoader does not show a difference.

stof:
I suggest you to hack some logging in this class to evaluate the hit/miss ratio of this cache.

Seldaek:
A similar approach to this PR was actually done in https://github.com/composer/composer/pull/7904 (https://github.com/composer/composer/blob/2.0/src/Composer/Package/Loader/ArrayLoader.php#L68 and https://github.com/composer/composer/blob/2.0/src/Composer/Package/Loader/ArrayLoader.php#L274) so closing this.

PR #6296, opened 2017-03-24
Add a test with KWStyle

RyDroid:


Seldaek:
Thanks but no thanks

RyDroid:
Why?

Seldaek:
First of all, I might ask you the same question.. why should we? You didn't bother explaining anything. I'm not familiar with KWStyle and had to spend time figuring out what it even is.

And then why? Because I don't see the point, this does not solve a pain point, it just adds stuff because you think we should have it.

RyDroid:
- It checks that there is a new line at the end of lines (so that using `cat` is convenient).
- There is no reason to have more than 2 empty lines.
- It is generally a good idea to detect files with too much lines to split them.

KWStyle is a free/libre command line tool for code quality. It is mostly designed for C/C++, but it has some useful checks for any type of source code.

PR #6297, opened 2017-03-24
Adding a test for encoding (ASCII or UTF-8)

RyDroid:


Seldaek:
Not sure why we even need this? I don't remember ever having an issue with file encoding in this project, and overall this hasn't happened in maybe 10years.. UTF-8 is pretty much standard at this point.

RyDroid:
It is the standard, but some people on Windows are still not using it. So this check warns them on merge request.

Seldaek:
I work on windows and I find your prejudice unnecessary. As I said in the other PR, this solves a problem that does not exist..

PR #6298, opened 2017-03-24
Add a specific message for stupid users that put useless spaces in license

RyDroid:
I was one of this stupid users, so I would like to help future ones.


src/Composer/Util/ConfigValidator.php
```diff
@@ -5,6 +5,7 @@
  *
  * (c) Nils Adermann <naderman@naderman.de>
  *     Jordi Boggiano <j.boggiano@seld.be>
+ *     Nicola Spanti <dev@nicola-spanti.info>
```
staabm: not your software ;)

RyDroid: Of course. Is it just a joke or do you want to have copyright for you ?

staabm: Nope, but you shouldnt add your name their

RyDroid: Ok, I removed it.

Issue #6311, opened 2017-03-31
Version of root package not detected


IchHabRecht:
My `composer.json`:

https://github.com/sebastianbergmann/phpunit/blob/master/composer.json

When I run this command:

```
composer update --no-interaction --no-ansi --no-progress --no-suggest --optimize-autoloader --prefer-stable -vvv 
```

See https://ci.appveyor.com/project/helhum/phpunit/build/1.0.7/job/l2rbao080bvoo7a7#L43

I get the following output:

```
Your requirements could not be resolved to an installable set of packages.
  Problem 1
    - phpunit/phpunit-mock-objects 4.0.1 conflicts with phpunit/phpunit[**No version set (parsed as 1.0.0)**].
    - phpunit/phpunit-mock-objects 4.0.0 requires sebastian/exporter ^2.0 -> satisfiable by sebastian/exporter[2.0.0, 2.0.x-dev] but these conflict with your requirements or minimum-stability.
    - Installation request for phpunit/phpunit No version set (parsed as 1.0.0) -> satisfiable by phpunit/phpunit[No version set (parsed as 1.0.0)].
    - Installation request for phpunit/phpunit-mock-objects ^4.0 -> satisfiable by phpunit/phpunit-mock-objects[4.0.0, 4.0.1].
```

And I expected this to happen:
Version of phpunit/phpunit should be 6.1.x-dev as defined in its branch alias

curry684:
The behavior is on purpose for a bunch of good reasons. The root package is technically 'versionless', but set to `1.0.0` so it practically fulfills every requirement that could be thrown against it, and circular dependencies are explicitly guarded. Not sure we've ever seen a package that via a dependency conflicts with itself though  😆 

A conflict with the root package should just always be explicitly ignored, as it should indeed be 'versionless'.

helhum:
@curry684 See discussion here: #5942 and the fix here e54c7478ee1207e60b0a345391a80b3511886429

To clarify the case:

We want to test PHPUnit dev-master on [Appveyor](https://ci.appveyor.com/project/helhum/phpunit/build/1.0.7/job/l2rbao080bvoo7a7#L40)

As you can see, the output of `git branch --no-color --no-abbrev -v` is
`* (HEAD detached at ef7c6aa68) ef7c6aa684e284970cae15e2f045d32678dc2036 Add Appveyor build file`

in contrast to

[...]

PR #6312, opened 2017-03-31
Fix Git branch parsing for detached HEAD on a commit

IchHabRecht:
Current versions of Git output the commit hash as detached HEAD instead
of FETCH_HEAD. The VersionGuesser should be able to handle commit hashes
as well as FETCH_HEAD to detect the correct branch of a commit.

Resolves: #6311

### DIFF
--- a/tests/Composer/Test/Package/Version/VersionGuesserTest.php
+++ b/tests/Composer/Test/Package/Version/VersionGuesserTest.php
@@ -158,16 +158,15 @@ class VersionGuesserTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals("dev-$commitHash", $versionData['version']);
     }
 
-    public function testDetachedHeadBecomesDevHashGit2()
+    public function testDetachedFetchHeadBecomesDevHashGit2()
     {
-        $commitHash = '03a15d220da53c52eddd5f32ffca64a7b3801bea';
+        $commitHash = 'da53c52eddd5f32ffca64a7b3801bea';
 
         $executor = $this->getMockBuilder('\\Composer\\Util\\ProcessExecutor')
-            ->setMethods(array('execute'))
+            ->setMethods(['execute'])
             ->disableArgumentCloning()
             ->disableOriginalConstructor()
-            ->getMock()
-        ;
+            ->getMock();
 
         $self = $this;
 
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Package/Version/VersionGuesserTest.php
```diff
@@ -176,7 +176,7 @@ public function testDetachedHeadBecomesDevHashGit2()
             ->method('execute')
             ->willReturnCallback(function ($command, &$output) use ($self, $commitHash) {
                 $self->assertEquals('git branch --no-color --no-abbrev -v', $command);
-                $output = "* (HEAD detached at FETCH_HEAD) $commitHash Commit message\n";
+                $output = "* (HEAD detached at " . substr($commitHash, 0, 9) . ") $commitHash Commit message\n";
```
staabm: I guess we need to support "both formats" - which essentially means you need 2 tests. one for the old format and one for newer one.

IchHabRecht: I'm not sure about this as the regular expression fetches both cases (FETCH_HEAD is text as the commit hash is, but not the other way around). I don't mind to add a second test here, but want to prevent unnecessary overhead.

helhum: I agree with @staabm here. Better keep the old test to make sure the old case is still working.

> as the regular expression fetches both cases

That might be true currently, but tests are there to also ensure this for the future.

IchHabRecht: Done

tests/Composer/Test/Package/Version/VersionGuesserTest.php
```diff
@@ -158,16 +158,15 @@ public function testDetachedHeadBecomesDevHash()
         $this->assertEquals("dev-$commitHash", $versionData['version']);
     }
 
-    public function testDetachedHeadBecomesDevHashGit2()
+    public function testDetachedFetchHeadBecomesDevHashGit2()
```
stof: I would remove the ``2`` as there is no conflict anymore

alcohol:
And how would we know what this commit references?

IchHabRecht:
@alcohol Sorry, I don't understand your question. Have you seen the link to the issue in the pull request description?

IchHabRecht:
Of course I have tested this patch with the issue locally and even added a test to show it solves the problem. I don't need any phar file, if I do an installation with the checkout of composer on my local system.

alcohol:
What I mean to say is, you merely loosened the constraint that enables the `isDetached` flow to trigger a call to `versionFromGitTags` which runs `git describe --exact-match --tags`. Does the latter result in positive match for your use-case?

tests/Composer/Test/Package/Version/VersionGuesserTest.php
```diff
@@ -158,16 +158,15 @@ public function testDetachedHeadBecomesDevHash()
...
     {
-        $commitHash = '03a15d220da53c52eddd5f32ffca64a7b3801bea';
+        $commitHash = 'da53c52eddd5f32ffca64a7b3801bea';
```
alcohol: Please use realistic output. Git hashes are `40` characters long, not `31`.

tests/Composer/Test/Package/Version/VersionGuesserTest.php
```diff
@@ -179,8 +178,37 @@ public function testDetachedHeadBecomesDevHashGit2()
...
+        $config->merge(array('repositories' => array('packagist' => false)));
+        $guesser = new VersionGuesser($config, $executor, new VersionParser());
+        $versionData = $guesser->guessVersion(array(), 'dummy/path');
+
+        $this->assertEquals("dev-$commitHash", $versionData['version']);
+    }
+
+    public function testDetachedCommitHeadBecomesDevHashGit2()
+    {
+        $commitHash = 'da53c52eddd5f32ffca64a7b3801bea';
```
alcohol: See previous comment about realistic hash length.

helhum:
@alcohol See the build here: https://ci.appveyor.com/project/helhum/phpunit/build/1.0.7/job/l2rbao080bvoo7a7#L32

`git branch --no-color --no-abbrev -v`

results in:

```
* (HEAD detached at ef7c6aa68) ef7c6aa684e284970cae15e2f045d32678dc2036 Add Appveyor build file
  master ef7c6aa684e284970cae15e2f045d32678dc2036 Add Appveyor build file
```

As you can clearly see, ef7c6aa68 in fact references the master branch, but composer does not detect this. In fact the code tries to detect it with the changes applied here: e54c7478ee1207e60b0a345391a80b3511886429

This code works if the current detatched head also is `FETCH_HEAD`

```
* (HEAD detached at FETCH_HEAD) ef7c6aa684e284970cae15e2f045d32678dc2036 Add Appveyor build file
  master ef7c6aa684e284970cae15e2f045d32678dc2036 Add Appveyor build file
```

I see no reason why, we should handle one case, but ignore the other. Both detatched head states clearly point to the master branch.

alcohol:
> Both detatched head states clearly point to the master branch.

As I explained in the issue linked, this is not "clear" at all. But the change is acceptable as it is unlikely to break anything.

Seldaek:
Patch looks fine to me, there are just many ways git shows this depending on versions and the way the detached state was reached I think. Anyway if the commit hash matches then indeed we should take the most meaningful version, as we already do in other cases.

If you can update the tests according to the comments it looks good to go for me.

Seldaek:
Thanks

helhum:
Thanks for the swift review and merge!

PR #6320, opened 2017-04-03
Cleanup: make use of variable

greydnls:


src/Composer/Command/RequireCommand.php
```diff
@@ -74,7 +74,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $io = $this->getIO();
 
         $newlyCreated = !file_exists($file);
-        if (!file_exists($file) && !file_put_contents($file, "{\n}\n")) {
+        if (!$newlyCreated && !file_put_contents($file, "{\n}\n")) {
```
Seldaek: This one is a good example of a seemingly obvious fix that can bite one in the ass if one doesn't review carefully ;)

$newlyCreated is `!file_exists($file)` already so the `!` should be removed here if I'm not mistaken.

alcohol: Correct.

greydnls: Yeah, my bad there. Fixed it.

Seldaek:
Heh no worries, thanks for the patch

PR #6326, opened 2017-04-05
[Doc] Explained a bit more about the possible contents of the "name" property

javiereguiluz:
In the JSON schema it says that this property is of type "string" ... but maybe we can explain this a bit more and recommend to use good package names.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -35,7 +35,7 @@ separated by `/`. Examples:
 * igorw/event-source
 
 The name can contain any character, including white spaces, and it's case
-sensitive (`foor/bar` is considered a different package than `Foo/Bar`). In order
+insensitive (`foor/bar` and `Foo/Bar` are considered the same package). In order
 to simplify its installation, it's recommended to define a short and lowercase
 name that doesn't include non-alphanumeric characters or white spaces.
 
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -35,7 +35,7 @@ separated by `/`. Examples:
 * igorw/event-source
 
 The name can contain any character, including white spaces, and it's case
-insensitive (`foor/bar` and `Foo/Bar` are considered the same package). In order
+insensitive (`foo/bar` and `Foo/Bar` are considered the same package). In order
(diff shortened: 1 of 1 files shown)


doc/04-schema.md
```diff
@@ -29,13 +29,16 @@ The config of dependencies is ignored. This makes the `config` field
...
-separated by `/`.
-
-Examples:
+separated by `/`. Examples:
 
 * monolog/monolog
 * igorw/event-source
 
+The name can contain any character, including white spaces, and it's case
+sensitive (`foor/bar` is considered a different package than `Foo/Bar`). In order
```
Seldaek: That is incorrect :) Package names are case-insensitive. We just enforce lowercase on packagist for uniformity's sake.

javiereguiluz: Fixed! Thanks.

doc/04-schema.md
```diff
@@ -29,13 +29,16 @@ The config of dependencies is ignored. This makes the `config` field
...
-separated by `/`.
-
-Examples:
+separated by `/`. Examples:
 
 * monolog/monolog
 * igorw/event-source
 
+The name can contain any character, including white spaces, and it's case
+insensitive (`foor/bar` and `Foo/Bar` are considered the same package). In order
```
stloyd: `foor/bar` is not same as `For/Bar` (`foo/bar` in fact) ;)

Seldaek:
Thanks

doc/04-schema.md
```diff
@@ -29,13 +29,16 @@ The config of dependencies is ignored. This makes the `config` field
...
-Examples:
+separated by `/`. Examples:
 
 * monolog/monolog
 * igorw/event-source
 
+The name can contain any character, including white spaces, and it's case
+insensitive (`foo/bar` and `Foo/Bar` are considered the same package). In order
+to simplify its installation, it's recommended to define a short and lowercase
+name that doesn't include non-alphanumeric characters or white spaces.
```
stof: shared packages have some extra requirements (due to the fact that the name will be used in a filesystem path). And Packagist enforces the strict regex.

PR #6329, opened 2017-04-06
Aligns the verb form of commands' description

issei-m:
Right now, the verb form in descriptions of commands is confused base form with third-person singular form. So I want to aligh this to the latter because Symfony's general commands are so.


Seldaek:
Thanks

PR #6330, opened 2017-04-06
Fix php notice

enumag:
Of course autoloading of an empty string should not actually happen (fixed that in https://github.com/twigphp/Twig/pull/2438) but if it does happen it should not throw a php notice.

```
Notice: Uninitialized string offset 0
```

The notice is thrown few lines below beacause of `$first = $class[0];`.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -369,6 +369,10 @@ class ClassLoader
 
     private function findFileWithExtension($class, $ext)
     {
+        if ($class === '') {
+            return false;
+        }
+
         // PSR-4 lookup
         $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR) . $ext;
 


src/Composer/Autoload/ClassLoader.php
```diff
@@ -369,6 +369,10 @@ public function findFile($class)
 
     private function findFileWithExtension($class, $ext)
     {
+        if (! $class) {
```
curry684: Please use typesafe comparisons, this may or may not (depending on PHP config) also reject class names "0" and "false". Also PSR mandates that the '!' operator has no space behind it. You should use `"" === $class` in this case.

Seldaek:
IMO it's fine as is, as it's clearly only caused by bug in the calling code. This is no valid use case, and it adds a check that is otherwise not needed, to all autoload calls for all users of composer, forever. That's a lot of wasted cpu cycles :)

Issue #5989, opened 2016-12-22
Stricter answer checking for init


kelunik:
`composer init` currently doesn't enforce `yes|y|no|n` as an answer to `bool` questions like the one below. Somehow I wanted to directly search now multiple times, but that results in a `no` instead of asking again with a hint of valid answers.

**Current**

```
Would you like to define your dependencies (require) interactively [yes]? async-interop
Would you like to define your dev dependencies (require-dev) interactively [yes]? ^C
```

**Expected**

```
Would you like to define your dependencies (require) interactively [yes]? async-interop
Please answer yes, y, no, or n.
Would you like to define your dependencies (require) interactively [yes]?
```

tomzx:
This appears to be fixed @ v1.5.5.

@Seldaek Can you please close this?

Seldaek:
True that, thanks :)

PR #6337, opened 2017-04-07
Fix for https://github.com/composer/composer/issues/5989

theotonge:
This adds makes the init command enforce yes/no answers for boolean questions as requested in issue #5989

### DIFF
--- a/src/Composer/Question/StrictConfirmationQuestion.php
+++ b/src/Composer/Question/StrictConfirmationQuestion.php
@@ -1,0 +1,94 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Question;
+
+use Symfony\Component\Console\Exception\InvalidArgumentException;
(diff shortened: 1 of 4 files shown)


src/Composer/Question/StrictConfirmationQuestion.php
```diff
@@ -0,0 +1,93 @@
...
+    private $trueAnswerRegex;
+    private $falseAnswerRegex;
+
+    /**
+     * Constructor.s
+     *
+     * @param string $question        The question to ask to the user
+     * @param bool   $default         The default answer to return, true or false
+     * @param string $trueAnswerRegex A regex to match the "yes" answer
+     */
```
amanmahajan26: missing param for $falseAnswerRegex

theotonge: @amanmahajan26 Thank you, I have pushed a corrected version with this fixed

Seldaek:
Thanks for the PR. However I think this change/fix should be done inside ConsoleIO::askConfirmation so that all askConfirmation calls become more strict, instead of just changing those in init command, which also means the init command doesn't use the IOInterface consistently anymore..

Seldaek:
Also if you can rebase this on the 1.4 branch and change the pull request to target 1.4 that would be nice as it's a bugfix it can be applied to the next 1.4.x :)

theotonge:
I have updated accordingly - but forgot to rerun tests :( Will fix failed tests and update again

Seldaek:
Thanks :)

Issue #6334, opened 2017-04-07
Annoying and inappropriate donation-suggestion


dereuromark:
Is there a policy on the suggest field, and how not to abuse it for the wrong purposes?
Especially when it is yellow-flash "in your face" all the time?

<img width="521" alt="screenshot" src="https://cloud.githubusercontent.com/assets/39854/24780904/9cb2ed42-1b3b-11e7-8442-c3dff3c6fdad.png">

Imagine everyone would do this in every color, or even CAPITAL LETTERS and RED.
Just wondering - I am fairly certain the intention of this suggest field is to give helpful and incentive suggestions with technical relevance on this specific subject of dependency.

I think people should be keeping donation suggestions to their readme/repo, and shouldn't force it upon people when pulling those things as dependency-dependencies ("3rd party").

Seldaek:
Yeah I tend to agree this is kinda over the top.. On the other hand, you're also welcome to voice your opinion by stopping to use that software if they annoy you, so I would say this is not really Composer's problem, and not something we can reasonably police I think.

One thing we could do though is strip any color codes from it to avoid such stylistic abuse at least, but policing the content is not possible nor reasonable I think.

dereuromark:
I think the stripping part sounds like a reasonable resolution for the time being. I hope the maintainers are responsible enough to remove that noise on their own.
Thanks!

dereuromark:
@Seldaek No I cant stop pulling it in. I dont use it or want to use it but a library i use pulls this in as dependency. So there is only so much one can do. :)

ondrejmirtes:
I'm really sorry about this. PHPStan itself doesn't show any message of this kind (although I accept pledges), but the dependency of PHPStan, parts of Nette framework, are the cause of this obtrusive message during installation. I'd wish that I'd be able to get rid of Nette as dependency but it won't be possible, not at least in the near future.

dg:
Btw, is it actually helpful to show suggestions of dependency-dependencies?

For example, cakephp/cakephp suggests to install lib-ICU to use Text::slug. When cakephp/cakephp will be dependency of dependency, my project code should not use class Text, so this suggestion is useless.

dereuromark:
@dg Yes, it depends on how those things are "optional" - as you can see it is optional use lib-ICU but once you want to use the slug part of the framework you need it. It is then directly used by the framework via dependency code. So at least the dependency of a dependency is needed to be shown to make sense.

dg:
[...]

PR #6340, opened 2017-04-08
Remove ANSI control characters from suggested packages output

derrabus:
With this PR, common control characters are removed from the output of suggest fields, with the exception of new line characters. This fixes the problem described in #6334.

bildschirmfoto 2017-04-08 um 14 02 02

### DIFF
--- a/src/Composer/Installer/SuggestedPackagesReporter.php
+++ b/src/Composer/Installer/SuggestedPackagesReporter.php
@@ -15,6 +15,7 @@ namespace Composer\Installer;
 use Composer\IO\IOInterface;
 use Composer\Package\PackageInterface;
 use Composer\Repository\RepositoryInterface;
+use Symfony\Component\Console\Formatter\OutputFormatter;
 
 /**
  * Add suggested packages from different places to output them in the end.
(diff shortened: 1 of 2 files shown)


naderman:
Still allowing newlines was my suggestion, not sure how strict we want this to be @Seldaek

derrabus:
Given the fact that new lines are still allowed and the string is not truncated, people might want to switch to ASCII art instead. 😉

Seldaek:
Think I'd remove newlines too tbh, I don't see a valid use case for that in a suggest line which should really be short.

Seldaek:
Also do you mind targetting the 1.4 branch with this PR?

derrabus:
Okay, I will open an new PR that includes new line characters and targets the 1.4 branch.

Seldaek:
Thanks but FYI you can just force-push into this branch and edit the PR to target 1.4 :)

derrabus:
Oh, I didn't realize that changing the target is possible now. I'll do that. :-)

derrabus:
All right, now all control characters are filtered. I decided to give the LF character a special treatment. LF is replaced by a space (which is a bit more forgiving on the output), all other control characters are simply thrown away.

stof:
you should also apply the escaping of symfony/console styles, otherwise they could just switch to using `<bg=green>`. The component has a method to apply escaping.

derrabus:
Good point. I've added escaping for Symfony Console markup.

src/Composer/Installer/SuggestedPackagesReporter.php
```diff
@@ -116,11 +117,35 @@ public function output(RepositoryInterface $installedRepo = null)
...
     }
+
+    /**
+     * @param string $string
+     * @return string
+     */
+    private function escapeOutput($string)
+    {
+        return OutputFormatter::escape(
+            $this->removeControlCharacters($string)
```
stof: no need to make this multiline

Seldaek:
Thanks

Issue #5938, opened 2016-12-03
Warn about identical names with different capitalization on zip archive extract failure (Not a directory)


naderman:
When unzip fails with "checkdir error: [path] exists but is not a directory" or if ZipArchive->extractTo fails with `ErrorException ZipArchive::extractTo(): Not a directory` we should display a message warning that the archive may contain a directory and a file with the same name but different capitalization which fails to extract on case insensitive filesystems.

ajiyakin:
@naderman is there anyone work on this?
I'm going to work on this then :smiley:

naderman:
@ajiyakin Please do! :-)

ajiyakin:
Hi @naderman, I've been do a pull request for this issue. Please tell me if there's something that I might be miss understood or missing something :smiley:

naderman:
This PR fixes the issue for ZipArchive, but the description actually also mentioned the same issue when using unzip. Can you make another PR for that too?

ajiyakin:
OK, I'll take a look then :+1:

ajiyakin:
Hi @naderman , sorry for my late pull request.
I already add warning for unzip for this issue. I also set warning message equals to the [master branch for ZipArchive](https://github.com/composer/composer/blob/master/src/Composer/Downloader/ZipDownloader.php#L160) one.

nrayann:
I am getting the same error when I run `composer update` on my Laravel Project.

The process is killed at `symfony/console` updating/installing.

Bash output:

	- Installing symfony/console (v4.2.7): Loading from cache

	Failed to execute unzip -qq  '/mnt/c/Users/rayann.souza/Sites/execut-api/vendor/symfony/console/d5e2f7700964cd8960261b5750040990' -d '/mnt/c/Users/rayann.souza/Sites/execut-api/vendor/composer/76f787b7'

	Killed

	The archive may contain identical file names with different capitalization (which fails on case insensitive filesystems)
	Unzip with unzip command failed, falling back to ZipArchive class

![image](https://user-images.githubusercontent.com/5538985/64639869-272dee00-d3df-11e9-86d6-0d9d4ae5a132.png)

### Steps To Reproduce:

- Ubuntu 18.04 with WSL on Windows 10 Pro
- intel core i7 7th Gen, 16gb RAM
- PHP 7.2.19
- Laravel 5.8

Anyone could help me?

Thank you in advance for any help :D

kolumnin:
Dr.K.Muraleedharan

[...]

PR #6341, opened 2017-04-08
Add warning for identical names with different capitalization when zip archive extract failure

ajiyakin:
This pull request add a warning about identical names with different capitalization on zip archive whenever extract failure with `\ErrorException` (Not a directory)

Issue #5938


Seldaek:
@ajiyakin thanks, looks fine I think and I'll try to merge it soonish but would like to confirm it works here first.

ajiyakin:
Thanks @Seldaek, just tell me if there's something that might need an improvement :+1:

Seldaek:
Thanks, merged in 8858b27ced8eeea6dd10b52c76dab2d4539e5e4d

PR #6345, opened 2017-04-10
Changes the query string parameter `ref` to `sha` to revert #6198

### LINKED ISSUE #6344: GitLab dist URL is wrong for GitLab API v3
I reported this issue in #6198 and provided a merge request to fix it. It seems my initial testing was wrong and the data set i used to verify the fix was wrong as well. The documentation on GitLab's side is not as clarifying as one might hope.

This basically means the 'fix' for the `dist-url` should be reverted. It was changed from `sha` to `ref` and should be reverted to `sha`. This fixes downloading of zipball archives from GitLab API `v3`.

I do apologize for any inconvenience this may have caused.

sjadema:
Updated the unit tests to reflect this change

Fixes #6344

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -175,7 +175,7 @@ class GitLabDriver extends VcsDriver
      */
     public function getDist($identifier)
     {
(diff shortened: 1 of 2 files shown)


Seldaek:
Fixed in be201924cd9f3f22c21399047d8740ba55025d83 actually sorry I missed this

PR #6350, opened 2017-04-12
Remember that class exists to prevent search it again

lchuhin:
We using the cache based on `classMap` array

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -362,6 +362,9 @@ class ClassLoader
         if (false === $file) {
             // Remember that this class does not exist.
             $this->missingClasses[$class] = true;
+        } else {
+            // Remember that this class exist
+            $this->classMap[$class] = $file;
         }
 
         return $file;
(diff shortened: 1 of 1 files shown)


stof:
Altering the ClassMap breaks the OPCache-based optimizations, by forcing to copy the array out of shared memory to mutate it.
So a big -1 here.

Btw, what is the use case to access the same file several times ? PHP never calls the autoloader twice for the same class, because once a class is loaded, it already knows about it and does not need to load it again

lchuhin:
Hello, @stof 
We are saving ClassMap in cache on production environment to avoid the searching classes locations in future script executions, but not in same PHP process

lchuhin:
If altering the classMap breaks the OPCache-based optimizations, can we use another class member for that like this?
```
private $foundClasses = array();

public function getFoundClass()
{
    return $this->foundClasses;
}
```

Seldaek:
To be honest this seems like something you can do pretty easily by decorating the classloader and keeping track of things yourself if it's really so important.. Or you could define another classmap definition in your autoload config and make sure that all classes are found? Either way this isn't a widespread enough use case to justify more clutter in the autoloader IMO.

stof:
Or just use the Composer optimized autoloader so that most classes are in the class map at generation time.

PR #6353, opened 2017-04-12
add support for GitLab subgroups

### LINKED ISSUE #6349: Gitlab subgroup
In my `composer.json`

```json
{
    "type": "gitlab",
    "url": "http://192.168.1.76/dnl/bundle/translation-bundle"
}
```

triggers an API call to http://192.168.1.76/dnl/api/v3/projects/bundle%2Ftranslation-bundle

It should be http://192.168.1.76/api/v3/projects/dnl%2Fbundle%2Ftranslation-bundle

MatTheCat:
Fix #6349 

Now that GitLab supports subgroups, repositories namespace can contain an arbitrary number of slashes which means we cannot get the gitlab domain through the regex, so now we have to parse the URL as well.

I know there is duplicated logic but I don't really know how to improve this.

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -66,7 +66,7 @@ class GitLabDriver extends VcsDriver
      */
     private $isPrivate = true;
 
-    const URL_REGEX = '#^(?:(?P<scheme>https?)://(?P<domain>.+?)/|git@(?P<domain2>[^:]+):)(?P<owner>[^/]+)/(?P<repo>[^/]+?)(?:\.git|/)?$#';
+    const URL_REGEX = '#^(?:(?P<scheme>https?)://(?P<host>[^/]+)/|git@(?P<githost>[^:]+):)(?P<owner>.+)/(?P<repo>[^/]+)(?:\.git|/)?$#';
 
     /**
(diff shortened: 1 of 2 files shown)


delamart:
I didn't check your code in detail but do you support multiple sub groups ?
group/sub1/sub2/project ?

MatTheCat:
@delamart yes

alcohol:
The implementation seems overly complicated to me to be honest. I'll see if I can come up with something more elegant.

alcohol:
Ah I see now why it is so complex. The addition of groups makes it much harder to detect whether or not it is a self-hosted instance that is located in a subdirectory, since these two cases sort of overlap. I'm not sure either if it can be done any easier or more elegant than your current approach, but I'll fiddle around with it for a bit.

MatTheCat:
Yep I initially thought I just had to update `URL_REGEX` but then tests failed because of `mycompany.com/gitlab` “domain”.

My main concern is logic duplication between `initialize` and `supports` methods.

alcohol:
Can you give me your opinion on #6373 in comparison?

MatTheCat:
I would have renamed `path` `namespace` otherwise it seems good to me.

MatTheCat:
Closed in favor of #6373

PR #6355, opened 2017-04-12
Update 05-repositories.md

OussamaElgoumri:
fix typo

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -585,7 +585,7 @@ For instance, if you have the following directory structure in your repository:
   \_ composer.json
 ```
 
-Then, to add the package `my/package` as a dependency, in your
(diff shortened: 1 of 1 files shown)


Seldaek:
That's not a typo, that's the package name.

OussamaElgoumri:
the directory structure you use in the example

 ```
 - apps
 \_ my-app
   \_ composer.json
 - packages
 \_ my-package
    \_ composer.json
  ```

-Then, to add the package `my/package` as a dependency, in your

where is the directory `my` and the subdirectory `package` in directory listing above?

Seldaek:
`my/package` would be the package name defined inside my-package/composer.json. And it is made accessible by defining the path repository pointing to packages/my-package/.

PR #6357, opened 2017-04-14
Made GitLabDriver extendable

anhskohbo:
Use `protected` instead of `private` of properties.

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -28,29 +28,29 @@ use Composer\Util\GitLab;
  */
 class GitLabDriver extends VcsDriver
 {
-    private $scheme;
-    private $owner;
-    private $repository;
+    protected $scheme;
+    protected $owner;
+    protected $repository;
 
     /**
      * @var array Project data returned by GitLab API
      */
-    private $project;
+    protected $project;
 
     /**
      * @var array Keeps commits returned by GitLab API
      */
-    private $commits = array();
+    protected $commits = array();
 
     /**
      * @var array List of tag => reference
      */
-    private $tags;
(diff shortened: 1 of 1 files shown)


stof:
Switching to protected means that Composer then has to provide backward compatibility for all these properties.
And this means that it becomes almost impossible for Composer itself to change the way the driver works in the future.

so please describe your use case for extending this class. Hurting maintenance on composer without a clear use case for this is not something I would like to see happening.

anhskohbo:
Humm, I'm try write a private app for my company allow sync tags and code (without composer.json file) in git-remote server (Gitlab, Github-privte, Bitbucket-private) to our server.

Composer already parse `owner`, `repository-name` when pass a URL, I need to get that, but it's not public properties. So I create `GitLabDriver`, `GitHubDriver` extends base class and pass them to `Composer\Repository\VcsRepository` class.  

Problem when `GitHubDriver` use `protected` for property just get that easy, but GitLab doesn't, can't get private property in a child-class.

BWT, I think change to `protected` is not break anything.

stof:
@anhskohbo changing to protected forbids us to make changes to the internals of the class in future versions of Composer. So it does not break composer. It forbids it to improve in the future (or at least makes it way harder)

anhskohbo:
Oh I known now.

I still leave this pull-request open in here. Someone can be review this and close/merged.

avant1:
@anhskohbo in PHP you can read and change private object properties using reflection or [Closure::bind](http://php.net/manual/en/closure.bind.php) method.

So you probably can implement required functionality without changes to Composer.

Seldaek:
I think we already have quite a bit of public getters, and maybe could have more to make more stuff accessible, but switching all properties to be protected instead of private is not the right solution sorry.

Issue #1147, opened 2012-09-25
Solver fails to resolve dependency.


christianjul:
Given the setup up below "typo3/party" is installed before "flow3/composer-installer" even though the latter is required by the former. 

--COMPOSER--

``` json
{
    "name": "christianjul/barebone-flow3",
    "description" : "Barebone FLOW3 project",
    "license": "GPL-3.0+",
    "authors": [
        {
            "name": "Christian Jul Jensen",
            "email": "julle@typo3.org"
        }
    ],
    "repositories": [
        {
            "type": "composer",
            "url": "https://raw.github.com/christianjul/barebone-flow3/master/t3org"
        }
    ],
    "config": {
            "vendor-dir": "Packages/Vendor",
            "bin-dir": "bin"
    },
    "require": {
        "typo3/flow3": "master"
    },
    "minimum-stability": "dev",
    "scripts": {
               "post-update-cmd": "TYPO3\\FLOW3\\Composer\\InstallerScripts::postUpdateAndInstall",
               "post-install-cmd": "TYPO3\\FLOW3\\Composer\\InstallerScripts::postUpdateAndInstall"
    }
}
```

--REPOSITORY--

``` json
{
    "packages": {
        "flow3/composer-installers": {
            "dev-master": {
                "name": "flow3/composer-installers",
                "version": "dev-master",
                "version_normalized": "9999999-dev",
                "source": {
                    "type": "git",
                    "url": "git://github.com/christianjul/installers.git",
                    "reference": "f49f777bd1e6be2445dd60c070a731736734e1d2"
                },
                "replace": {
                    "shama/baton": "*"
                },
                "require-dev": {
                    "composer/composer": "1.0.*"
                },
                "time": "2012-09-25 14:48:23",
                "type": "composer-installer",
                "extra": {
                    "class": "Composer\\Installers\\Installer"
                },
                "autoload": {
                    "psr-0": {
                        "Composer\\Installers\\": "src/"
                    }
                },
                "license": [
                    "MIT"
                ],
                "authors": [
                    {
                        "name": "Kyle Robinson Young",
                        "email": "kyle@dontkry.com",
                        "homepage": "https://github.com/shama"
                    }
                ],
                "description": "A multi-framework Composer library installer",
                "homepage": "http://composer.github.com/installers/",
                "keywords": [
                    "installer",
                    "cakephp",
                    "laravel",
                    "kohana",
                    "li3",
                    "lithium",
                    "fuelphp",
                    "codeigniter",
[...]

PR #6371, opened 2017-04-22
Move all plugins and their dependencies to the front

balbuf:
Currently plugins are only moved to the front to be installed first when they don't have dependencies beyond platform requirements and `composer-plugin-api`. This causes an issue for our plugin [`oomphinc/composer-installers-extender`](https://github.com/oomphinc/composer-installers-extender) which is built atop `composer/composer-installers` and thus requires it. Users are seeing the same issue as https://github.com/composer/composer/issues/1147, see: https://github.com/oomphinc/composer-installers-extender/issues/6.

What this does is instead of skipping plugins that have additional dependencies, it moves the plugin and those dependencies to the front as well. This works on the assumption that install/update operations for a package's dependencies come before the package that required it. I'm not sure if that is a completely safe assumption to make, but it seemed to be the case in my testing.

I'm not sure if there were other reasons why plugins that had additional dependencies were skipped over from being moved in the first place; are there any other implications here that I am neglecting? As well, do you think this approach guaranteed to move plugins and all their dependencies properly or do you think more robust logic is necessary?

Please let me know your thoughts!

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -744,8 +744,10 @@ class Installer
      */
     private function movePluginsToFront(array $operations)
     {
-        $installerOps = array();
-        $installerRequires = array();
+        $pluginsNoDeps = array();
+        $pluginsWithDeps = array();
+        $pluginRequires = array();
+
         foreach (array_reverse($operations, true) as $idx => $op) {
             if ($op instanceof InstallOperation) {
                 $package = $op->getPackage();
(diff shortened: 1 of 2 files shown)


phenaproxima:
Please, +1 for this.

balbuf:
In terms of my wondering whether a package that depends on another is consistently installed after its dependency, it seems that is entirely the point of how the installation order is determined. At least based on what's been said in the past by some reputable sources:

> The only constraint used when constructing a transaction is that 
packages which are required by another package are installed before the 
package that requires them. 

https://groups.google.com/d/msg/composer-dev/a14g7NiELv0/RFCDYvjJlLcJ

> the rule to sort the packages for installations is that dependencies of a package are always installed before the package itself (except if you have a circular dependency in your packages of course as this would be impossible)

https://github.com/composer/composer/issues/2972#issuecomment-42497953

So if that is indeed true, then when starting from the end of the transaction list and working backwards, you should always first encounter packages that are not depended on by anything but the root package itself, followed by any of their direct dependencies, which in turn are always followed by their direct dependencies and so on. In other words, the algorithm this PR uses should be sound, barring any anomalies I am neglecting.

src/Composer/Installer.php
```diff
@@ -754,19 +755,15 @@ private function movePluginsToFront(array $operations)
...
-                if (!count($requires)) {
-                    $installerOps[] = $op;
-                    unset($operations[$idx]);
-                }
+            if ($package->getType() === 'composer-plugin' 
+                || $package->getType() === 'composer-installer'
+                || in_array($package->getName(), $installerRequires)
+            ) {
+                // capture the requirements for this package so those packages will be moved up as well
+                $installerRequires = array_merge($installerRequires, array_keys($package->getRequires()));
```
naderman: this needs to take provider names into account, not just requires

tests/Composer/Test/Fixtures/installer/plugins-are-installed-first.test
```diff
@@ -25,7 +25,7 @@ Composer installers are installed first if they have no meaningful requirements
 install
 --EXPECT--
 Installing inst (1.0.0)
-Installing inst-with-req (1.0.0)
-Installing pkg (1.0.0)
 Installing pkg2 (1.0.0)
 Installing inst-with-req2 (1.0.0)
+Installing inst-with-req (1.0.0)
```
naderman: this plugin is now not loaded for pkg2 installation anymore, so we should try and still move any plugin without dependencies to the very front to allow them to be loaded for every non-plugin dependency

balbuf: @naderman just to confirm, what you are saying is to move all plugins with no dependencies to the very front, followed by any plugins with their dependencies, followed by all other dependencies in their previously determined order (i.e. left where they are)?

src/Composer/Installer.php
balbuf: @naderman does that mean we need to consider "replaces" as well?

balbuf: @naderman is this what you were getting at? see: https://github.com/composer/composer/pull/6371/commits/4cda7e0a44d608f5bb6071003790046594535549

balbuf:
@naderman I've made changes based on what I understood your requests to be - let me know what you think!

src/Composer/Installer.php
naderman: @balbuf yeah that's what I had in mind.

tests/Composer/Test/Fixtures/installer/plugins-are-installed-first.test
naderman: Exactly :-)

balbuf:
@naderman @Seldaek thanks for merging this in! So that I can provide an update on https://github.com/oomphinc/composer-installers-extender/issues/6, when do you think this will makes its way into a release?

Seldaek:
@balbuf took a while sorry but 1.5 is out with this change

balbuf:
Thanks, @Seldaek !

PR #6377, opened 2017-04-25
Auto-skip interactive require when set by option.

CupOfTea696:
Currently when running `composer init --require=symfony/console`, you will still get the yes/no question to define your dependencies interactively. If you answer no to this question, the `composer.json` will be initialised without the requirement passed to the command initially. I feel like this is not desired behaviour.

This PR skips the 'Would you like to define your dependencies (require) interactively?' and 'Would you like to define your dev dependencies (require-dev) interactively?' questions if the `--require` or `--require-dev` option(s) were set already.


Seldaek:
Sounds reasonable, thanks

PR #6378, opened 2017-04-26
Revert "Fixed an issue when a phar file is used in "files" option in composer.json"

chuangbo:
This reverts commit 41e91f3064cee51dd52f45831dc48748ce37a095.

The commit 41e91f3 in current codebase generates absolute path in
`autoload_static.php` for phar file.

Example: https://github.com/chuangbo/composer-phar-test

Also according to http://php.net/manual/en/phar.using.intro.php, the `phar://`
prefix is not needed.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -534,10 +534,6 @@ EOF;
             }
         }
 
-        if (preg_match('/\.phar$/', $path)) {
-            $baseDir = "'phar://' . " . $baseDir;
-        }
-
         return $baseDir . (($path !== false) ? var_export($path, true) : "");
     }
 


chuangbo:
@mnishihan @Seldaek

alcohol:
Does the original issue this "fix" addressed no longer apply?

Seldaek:
Closing as this code has been fine for years so unless you provide more details here I don't see the point of changing it.

chuangbo:
Much appreciate the reply, but this code is not fine to me actually, more details below.

## What I see

If there is a `phar` file included in `composer.json` like this,

```json
{
    ...
    "require": {
        "laravel/framework": "5.4.*"
    },
    "autoload": {
        "files": [
            "example.phar"
        ]
    }
}
```

and `composer dump` generates `vendor/composer/autoload_static.php` like this

```php
class ComposerStaticInit4265aabe1ca78d4ab173c18dbbc2e921
{
    public static $files = array (
        'f0906e6318348a765ffb6eb24e0d0938' => __DIR__ . '/..' . '/laravel/framework/src/Illuminate/Foundation/helpers.php',
        '58571171fd5812e6e447dce228f52f4d' => __DIR__ . '/..' . '/laravel/framework/src/Illuminate/Support/helpers.php',
        'b53d7494e4da437b0b9ee7ed2cc352b7' => 'phar:///Users/user/project/example.phar',
    );
    ...
}
```

## What I expected

```php
class ComposerStaticInit4265aabe1ca78d4ab173c18dbbc2e921
{
    public static $files = array (
        'f0906e6318348a765ffb6eb24e0d0938' => __DIR__ . '/..' . '/laravel/framework/src/Illuminate/Foundation/helpers.php',
        '58571171fd5812e6e447dce228f52f4d' => __DIR__ . '/..' . '/laravel/framework/src/Illuminate/Support/helpers.php',
        'b53d7494e4da437b0b9ee7ed2cc352b7' => __DIR__ . '/../..' . '/example.phar',
    );
    ...
}
```

## Issue

The phar file pointed to an absolute path (different than others), also this stop us from moving the project to other folders (e.g. using docker).

## The reason I found

When generating `vendor/composer/autoload_static.php`, the function `Composer\Autoload\AutoloadGenerator@getStaticFile` creates absolute paths first,

```php
$value = "array(
  ...
  'f0906e6318348a765ffb6eb24e0d0938' => '/Users/user/project/vendor/laravel/framework/src/Illuminate/Foundation/helpers.php',
  '58571171fd5812e6e447dce228f52f4d' => '/Users/user/project/vendor/laravel/framework/src/Illuminate/Support/helpers.php',
  'b53d7494e4da437b0b9ee7ed2cc352b7' => 'phar:///Users/user/project/example.phar',
)";
```

And then replace `$absoluteAppBaseDirCode` with `$appBaseDirCode` in https://github.com/composer/composer/blob/a6627d55b7f5530d2474ca232e0cb0f8852e7bcf/src/Composer/Autoload/AutoloadGenerator.php#L789

```php
$absoluteAppBaseDirCode =  " => '/Users/user/project/";
$appBaseDirCode =  " => __DIR__ . '/../..' . '/";
$value = str_replace($absoluteAppBaseDirCode, $appBaseDirCode, $value);
// Failed
```

Unfortunately, it fails to replace because the phar file path ` => 'phar:///Users/user/project/` does not match ` => '/Users/user/project/`.

## What I tried to fix

1. Replace value one more time with `phar://` prefix

```php
$absoluteAppBaseDirPharCode =  " => 'phar:///Users/user/project/";
$appBaseDirPharCode =  " => 'phar://' . __DIR__ . '/../..' . '/";
$value = str_replace($absoluteAppBaseDirPharCode, $appBaseDirPharCode, $value);
```

2. Remove `phar://` prefix at the beginning

https://github.com/composer/composer/blob/a6627d55b7f5530d2474ca232e0cb0f8852e7bcf/src/Composer/Autoload/AutoloadGenerator.php#L537-L539

```diff
-        if (preg_match('/\.phar$/', $path)) {
-            $baseDir = "'phar://' . " . $baseDir;
-        }
```

The code above included from 41e91f3 (Jan 16 2013).

The commit does not mention any reasons, neither the pull request. As I have no knowledge of composer history, I could not find when this issue is introduced.

Overall, I personal choose the second way to fix, because according to http://php.net/manual/en/phar.using.intro.php, the `phar://` prefix is not necessary. Also, it is simpler.

Thanks for your time again!

Seldaek:
Alright I finally found a second to try this out here with and without patch, looks fine to me indeed so merged in 1.4 and master. Thanks.

PR #6381, opened 2017-04-26
Forward GitLab driver options to remote filesystem

deeky666:
Fix for: https://github.com/composer/composer/commit/876f14341814d427e87dd358ed91ccc855678a5c#commitcomment-21901159

When using GitLab driver in a non-public network (e.g. company) and the GitLab server instance is using SSL with a self-signed certificate, it is not possible to retrieve repository information because SSL certificate verification errors (encountered in the context of Satis).

This could be solved by explicitly disabling SSL verification check via `options` but unfortunately composer isn't forwarding the options to the remote filesystem.

This patch fixed the issue for VCS based drivers by always forwarding `options` if available. The same has been done for the `composer` repository type as seen in the commit above.

### DIFF
--- a/src/Composer/Repository/Vcs/VcsDriver.php
+++ b/src/Composer/Repository/Vcs/VcsDriver.php
@@ -160,7 +160,9 @@ abstract class VcsDriver implements VcsDriverInterface
      */
     protected function getContents($url)
     {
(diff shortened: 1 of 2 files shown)


alcohol:
(used the wrong option)

Seldaek:
Thanks

PR #6397, opened 2017-05-02
Do not use realpath() with exclude-from-classmap

leofeyer:
Using `realpath()` when resolving the `exclude-from-classmap` paths will also resolve symlinks to their real path. However, they will not match the `$blacklist` regexp after that anymore and thus they will not be applied.

```
project/
    vendor/
        foo/
            bar/  <-- this is a symlink to ../../../../forks/bar/
                src/
                    exclude/

forks/
    bar/
        src/
            exclude/
```

The relevant `composer.json` part is:

```json
"exclude-from-classmap": [
    "src/exclude/"
]
```

Now run `composer dump-autoload`.

**Current situation**

* `$resolvedPath`: `/var/www/forks/bar/src/exclude`
* `$blacklist`: `{(/var/www/project/vendor/foo/bar/src/exclude)}`
* No match

**With the patch**

* `$resolvedPath`: `/var/www/project/vendor/foo/bar/src/exclude`
* `$blacklist`: `{(/var/www/project/vendor/foo/bar/src/exclude)}`
* Match

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,8 @@
         "symfony/filesystem": "^2.7 || ^3.0",
         "seld/phar-utils": "^1.0",
         "seld/cli-prompt": "^1.0",
-        "psr/log": "^1.0"
+        "psr/log": "^1.0",
+        "webmozart/path-util": "^2.3"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.5 || ^5.0.5",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "696d16729ae31dafdd42432da4d74aae",
+    "content-hash": "fe586132ed353b896e7f45e45e0c346e",
     "packages": [
         {
             "name": "composer/ca-bundle",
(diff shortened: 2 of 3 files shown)


alcohol:
You managed to break some tests. Surely you must have noticed this?

leofeyer:
I have. And I am going to adjust the tests once we have agreed that this PR is correct and will be merged.

Seldaek:
Seems reasonable although potentially dangerous too (I can imagine side effects happening) but I'm fine merging it tentatively into master once the test suite is fixed.

leofeyer:
I have updated the PR.

The general problem is that everyone uses `realpath()` to canonicalize paths although `realpath()` follows symlinks. I guess that is why @webmozart has created [`webmozart/path-util`](https://github.com/webmozart/path-util), which does the same without following symlinks.

@Seldaek I really don't know if you should merge this PR. 😄 It fixes the `exclude-from-classmap` issue for sure, however there might be more places in the code where `realpath()` should be replaced with `Path::canonicalize()`.

leofeyer:
Oh, and I am not responsible for the tests failing under HHVM. 😄

alcohol:
Interestingly, if I use the trusty distro, the 5.3 job fails. I'll investigate to see if we can find a work-around. Otherwise we might have to drop hhvm from our Travis matrix.

Seldaek:
How about just calling realpath() on the resolved path when checking it against the blacklist? That might be an easier fix.

leofeyer:
> How about just calling realpath() on the resolved path when checking it against the blacklist? That might be an easier fix.

What exactly do you mean?

Seldaek:
I mean at https://github.com/leofeyer/composer/blob/2dca1d475c916840d64b4920d75c8b5c7f5312b0/src/Composer/Autoload/ClassMapGenerator.php#L93 we could do `strtr(realpath($filePath), '\\', '/'))` instead to check if the item's real path (with symlink resolved) is in the blacklists (with symlink resolved).

leofeyer:
But the point of the change is **not** to resolve any symlinks. 😄

leofeyer:
Changing the [ClassMapGenerator code](https://github.com/leofeyer/composer/blob/2dca1d475c916840d64b4920d75c8b5c7f5312b0/src/Composer/Autoload/ClassMapGenerator.php#L93) to `strtr(realpath($filePath), '\\', '/'))` does not solve the issue. It even breaks the blacklist, because the paths are not canonicalized anymore:

```
{(/var/tmp/composer\-test\-6269592ed007af20f/working\-dir/\./\.\.//classmap/excluded)
```

Seldaek:
Ping @leofeyer ?

leofeyer:
So here is how to reproduce the issue with public packages.

```bash
composer create-project contao/news-bundle --no-dev

# Look at the end of the autoload_classmap.php file
# It shoud be about 450 lines long and end with 'lessc' => …
vi + news-bundle/vendor/composer/autoload_classmap.php

# Now move the core-bundle outside the installation dir and create a symlink
mv news-bundle/vendor/contao/core-bundle .
cd news-bundle/vendor/contao
ln -s ../../../core-bundle/ .
cd ../../../

# Recreate the autoload files
composer dump-autoload

# Now look at the end of the autoload_classmap.php file again
# It is now about 680 lines long and ends with 'uploadable' => …
# because the blacklist paths have been resolved by realpath()
vi + news-bundle/vendor/composer/autoload_classmap.php
```

Seldaek:
@leofeyer I pushed the realpath() fix I mentioned above because after testing it it seems to fix the issue for me.. I'd appreciate if you can `composer self-update --snapshot` and confirm it also fixes it for you.

leofeyer:
It sure does. Thank you very much.

Seldaek:
Great :)

svenmuennich:
Unfortunately bf940fb21b0dcc830e8c8430a3cc9d7f3ac607c0 breaks the comparison of symlinked files completely. We noticed this while developing in [shopware/shopware](https://github.com/shopware/shopware). There the directory [`engine/Shopware/Plugins/Local/` is excluded from the autoloader](https://github.com/shopware/shopware/blob/8efbe6a850a9ca72701efd24d56aaa10408f9192/composer.json#L102).

If I now symlink a directory `Foo` into `engine/Shopware/Plugins/Local/Frontend/`, I expect that all files contained in the symlinked directory `engine/Shopware/Plugins/Local/Frontend/Foo/` are excluded from the class map. However, they are not, because due to calling `realpath()` on the paths of the symlinked files they don't match the excluded path anymore.

The difference to @leofeyer example/use case is, that his rule in `exclude-from-classmap` is not absolute to the project root. According to the docs, paths defined in `exclude-from-classmap` must be relative to the root though:

> The paths are absolute from the package root directory (i.e. composer.json location), and support * to match anything but a slash, and ** to match anything. ** is implicitly added to the end of the paths.

So to me it looks like this has never been an issue that needed fixing, but rather @leofeyer's issue was caused by an invalid exclude rule.

@leofeyer Does changing your exclude path to `**/src/exclude/` fix the problem you originally had, without having to apply any changes to composer?

Seldaek:
@svenmuennich the way I understand it the difference is @leofeyer had symlinked an entire package at its root and you are symlinking a directory into a subdirectory of the package. So his was failing because we compared absolute realpath of the package to linked path of the file, now we compare realpath against realpath and that works. In your case we compared realpath of package rule to the linked path of the file below it anyway so that was fine and now we compare realpath of the package to the realpath of the file but that is still now moving the file outside the package root so the rule doesn't match.

Anyway your use case seems more broken to me than leo's as you are doing symlinks by hand while leo's case is possible to achieve using the path repository meaning composer can create such a situation but not yours. Therefore I don't really feel like reverting is a good option.

The question is whether we can fix it somehow. I can imagine it would be fixed if we compared both realpath or original path to the blacklist, but that'll slow things down a bit sadly.

Seldaek:
In any case, if you really care about this issue please open a new one referencing this as I can't even reopen this as it's a PR. Then we can decide there how to best proceed.

fixpunkt:
> Anyway your use case seems more broken to me than leo's as you are doing symlinks by hand while leo's case is possible to achieve using the path repository meaning composer can create such a situation but not yours. Therefore I don't really feel like reverting is a good option.

There's nothing broken about this, it's just a somewhat exotic use case. We depend on this in order to develop plugins for https://github.com/shopware/shopware. We usually have one working copy of each plugin and symlink that into multiple installations of shopware in order to test it with different versions.

alcohol:
@fixpunkt sounds like a path repository is a perfect use-case there?

fixpunkt:
> @fixpunkt sounds like a path repository is a perfect use-case there? I do not quite see the need for manual symlinking.

We build third-party runtime plugins for https://github.com/shopware/shopware. If I understand that suggestion correctly, that would mean we'd have to patch our plugins into https://github.com/shopware/shopware's `composer.json` as dependencies? 

TBH, having to maintain a fork of every shopware minor version which has our plugins added as dependencies sounds a lot worse than manual symlinking. Also, since excludes over symlinked directories don't work correctly anymore (except when they're excludes defined inside a symlinked package's `composer.json`) since https://github.com/composer/composer/commit/bf940fb21b0dcc830e8c8430a3cc9d7f3ac607c0B,
 this wouldn't even solve our problem of not being able to exclude Shopware's plugin directory `engine/Plugins` from classmap generation.

svenmuennich:
For reference in case anyone runs into this issue: 

#7251 added the original blacklist matching again, in addition to the `realpath` matching added in bf940fb21b0dcc830e8c8430a3cc9d7f3ac607c0.

So both @leofeyer's and my use cases work now.

PR #6404, opened 2017-05-05
relative path is allowed for `mirror.dist-url`

RobberPhex:
when `mirror.dist-url` is `/file/%package%/%reference%.%type%`. it should be parsed in current domain.

such as `http://example.com/file/%package%/%reference%.%type%`.


alcohol:
I *think* this change makes sense, but I am not 100% certain. What do you think @Seldaek ?

Seldaek:
Thanks, merged in 1.4 and master

PR #6418, opened 2017-05-15
Store json file using redis

luoxiaojun1992:


Seldaek:
Sorry but I don't really see the added value here.. Most people don't have redis and increasing memory usage instead of putting stuff on the filesystem doesn't really achieve anything.

PR #6428, opened 2017-05-18
Fix grammar and match form of similar questions

njbooher:


Seldaek:
Fine by me to update the title but please don't rename the file as that breaks links.

njbooher:
Updated.

Seldaek:
Thanks

PR #6429, opened 2017-05-19
Register github repository type

hason:


Seldaek:
Ah crap the 5.3 tests broke, just saw it as I clicked merge.. do you mind sending a fix? :)

hason:
Fixed in #6430

PR #6432, opened 2017-05-20
Update GitLab personal access token URL and the token config command

misenhower:
This change fixes two issues in the GitLab login error message:
1. The personal access tokens page is located at https://gitlab.com/profile/personal_access_tokens.
2. The generated personal access token should be stored in the config as `gitlab-token` instead of `gitlab-oauth`.

The second issue in particular confused me for a while earlier today 😃 

It also might be better to recommend that the user runs `composer global config` instead of applying it to the local config. Let me know if you want me to commit another change to add that.


Seldaek:
Thanks, merged in 1.4 & master.

Issue #6434, opened 2017-05-22
[1.4.2] test failure with recent dependencies


remicollet:
Very probably not detected in travis because of composer.lock

With Symfony 2.8, and json-schema 4 everything is ok.

With **Symfony 3.2** (allowed by "symfony/console": "^2.7 || ^3.0")
```
There was 1 error:

1) Composer\Test\IO\ConsoleIOTest::testAskAndValidate
TypeError: Argument 1 passed to Symfony\Component\Console\Question\Question::setValidator() must be callable, string given, called in /dev/shm/BUILDROOT/composer-1.4.2-1.fc25.remi.x86_64/usr/share/php/Composer/IO/ConsoleIO.php on line 263

/usr/share/php/Symfony3/Component/Console/Question/Question.php:167
/dev/shm/BUILDROOT/composer-1.4.2-1.fc25.remi.x86_64/usr/share/php/Composer/IO/ConsoleIO.php:263
/dev/shm/BUILD/composer-489e09ee6c3ba431fbeeef9147afdaeb6f91b647/tests/Composer/Test/IO/ConsoleIOTest.php:222
```

With **json-schema 5.2.1** (allowed by "justinrainbow/json-schema": "^3.0 || ^4.0 || ^5.0")

```
There were 3 failures:

1) Composer\Test\Json\ComposerSchemaTest::testRequiredProperties
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
         'property' => 'name'
         'message' => 'The property name is required'
         'constraint' => 'required'
+        'context' => 1
     )
     1 => Array (
         'property' => 'description'
         'message' => 'The property description is required'
         'constraint' => 'required'
+        'context' => 1
     )
 )

/dev/shm/BUILD/composer-489e09ee6c3ba431fbeeef9147afdaeb6f91b647/tests/Composer/Test/Json/ComposerSchemaTest.php:28

2) Composer\Test\Json\ComposerSchemaTest::testRequireTypes
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
         'property' => 'require.a'
         'message' => 'Array value found, but a stri...quired'
         'constraint' => 'type'
+        'context' => 1
     )
 )

/dev/shm/BUILD/composer-489e09ee6c3ba431fbeeef9147afdaeb6f91b647/tests/Composer/Test/Json/ComposerSchemaTest.php:52

[...]

PR #6435, opened 2017-05-22
Issue deps

remicollet:
See #6434


remicollet:
Notice, a quick search in composer/src show all call to askAndValidate properly use a callable

Seldaek:
Thanks :)

PR #6458, opened 2017-05-30
Warn identical names w/ different capitalization for unzip

ajiyakin:
Warn about identical names with different capitalization on unzip archive
extract failure (Not a directory)

Issue #5938


Seldaek:
Thanks

Issue #6362, opened 2017-04-19
Clear cache takes forever


barbazul:
I recently run into some issues with packagist.org cache not clearing a deleted repo (which was dutifully fixed by Jordi). However, while experimenting myself I found that composer clear-cache took about 30 minutes on each run, even if I ran it and immediately ran it again (on an emptied cache).

So, for the first time ever I started to look under the hood in Composer core.

First thing I did was locating the cache directory, it had the 3 corresponding directories (files, repo and vcs) and inside each of them was a huge folder tree (thousands of nested folders) and not a single file.

The hypothesis was pretty obvious: Clear cache is recursively iterating this massive tree of folders and deleting only the files within. The 30 minutes were spent on recursing directories and not so much in deleting content.

So I went to the code and found the inner workings of this **bug**.

`Composer\Command\ClearCacheCommand::execute()` runs `Composer\Cache::gc()` on each first-level cache folder which in turn calls iterates a `Symfony\Component\Finder\Finder` instance with a trivial date filter of "every file created before right now". So, essentially, every file in the tree -> Hypothesis confirmed!

So for a final validation, I manually deleted the content of the 3 folders (which took about 3 seconds) and ran composer clear-cache once more. It took 0 seconds this time.

I have yet to figure out how to build composer myself to start working on a fix and I honestly don't think I will find the time to do so in the near future, but if I had to guess, I'd say that `Composer\Cache::getFinder()` should drop the call to `Symfony\Component\Finder\Finder::files()` to avoid filtering out directories. Or alternatively, handle the complete wiping of cache as a special case and not just a "forced expiration of everything" and either use a different Finder instance or iterate and delete the folders right after iterating and deleting the files (potentially taking longer when the cache is really big).

Final consideration: some garbage collection mechanism  to delete emptied cache folders over time could be a better (although more complex) approach

stof:
dropping the `files()` call would break the feature. The last modification date of a directory is not updated when you *update* a file in it. It is updated only when adding or removing files. So a directory might be detected as being older than the threshold, while it still has cached files in it.
We cannot implement `gc` this way.

barbazul:
Yup. That makes sense.

So as mentioned before, you would need to not use gc(0, 0) and instead
create a method "clear" which would use a different finder instance.

Seldaek:
[...]

PR #6460, opened 2017-06-01
Cache: clear cache using removeDirectory

bmax:
fixes: #6362

cc/ @barbazul @Seldaek @stof 

Hello! This is my first PR. As directed in issue #6362 I introduced a separate function for clearing the cache and just used `removeDirectory` directly instead of the `gc` command. 
 
_*Note*_: If `cache-dir` is the root directory of `cache-files-dir`, `cache-repo-dir` and `cache-vcs-dir` than it tries to remove something that is not there -- possible to optimize more?

Let me know if any questions or concerns. thanks.

### DIFF
--- a/src/Composer/Cache.php
+++ b/src/Composer/Cache.php
@@ -175,6 +175,15 @@ class Cache
         return false;
     }
 
+    public function clear()
+    {
+        if ($this->enabled) {
+            return $this->filesystem->removeDirectory($this->root);
+        }
+
+        return false;
+    }
+
     public function gc($ttl, $maxSize)
     {
         if ($this->enabled) {
--- a/src/Composer/Command/ClearCacheCommand.php
+++ b/src/Composer/Command/ClearCacheCommand.php
@@ -42,10 +42,10 @@ EOT
         $io = $this->getIO();
 
         $cachePaths = array(
-            'cache-dir' => $config->get('cache-dir'),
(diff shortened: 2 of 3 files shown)


alcohol:
> Note: If cache-dir is the root directory of cache-files-dir, cache-repo-dir and cache-vcs-dir than it tries to remove something that is not there -- possible to optimize more?

Not sure what you mean by "something that is not there"?

bmax:
@alcohol It tries to remove a directory that doesn't exist because it's already been removed. For example cache-dir = /cache and cache-repo-dir = /cache/repo than when it iterates through those directories it removes /cache first and gets a does not exist for the /cache/repo directory.

alcohol:
Ah, right. Well I think that is a non-issue. But I'd have to glance over the code again to make sure. Will do so later today.

alcohol:
Maybe make `cache-dir` the last entry of the array which is iterated rather than the first? That way we first empty the specific cache directories, and after that the general (possibly root) cache directory.

bmax:
> Maybe make cache-dir the last entry of the array which is iterated rather than the first in the ClearCacheCommand? That way we first empty the specific cache directories, and after that the general (possibly root) cache directory. This should solve any issues mentioned above.

Done @alcohol  cc\ @Seldaek

Seldaek:
Thanks, looks good to me.

barbazul:
Very nice! 👏

PR #6461, opened 2017-06-01
Lock file package time inconsistency

Tezd:
Dev packages which are installed from source  will produce time in `Y:m:d H:i:s` format according to lines 373-375, 435 in file `composer/src/Composer/Package/Locker.php`.
Other packages will have time formatted by `ArrayDumper` class and will have DATE_RFC3339 format as per line 86.

### DIFF
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -432,6 +432,6 @@ class Locker
             }
         }
 
-        return $datetime ? $datetime->format('Y-m-d H:i:s') : null;
+        return $datetime ? $datetime->format(DATE_RFC3339) : null;
     }
 }
--- a/src/Composer/Repository/Vcs/VcsDriver.php
+++ b/src/Composer/Repository/Vcs/VcsDriver.php
@@ -116,7 +116,7 @@ abstract class VcsDriver implements VcsDriverInterface
         $composer = JsonFile::parseJson($composerFileContent, $identifier . ':composer.json');
 
         if (empty($composer['time']) && $changeDate = $this->getChangeDate($identifier)) {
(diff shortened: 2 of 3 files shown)


stof:
👍

alcohol:
There is also https://github.com/composer/composer/blob/9aadbe27283bda0ed0f149fa4304c245f517c369/src/Composer/Repository/Vcs/VcsDriver.php#L119 which I think could also benefit from the same modification.

The other references I could find are probably fine as-is:
* https://github.com/composer/composer/blob/9aadbe27283bda0ed0f149fa4304c245f517c369/src/Composer/Compiler.php#L181
* https://github.com/composer/composer/blob/9aadbe27283bda0ed0f149fa4304c245f517c369/src/Composer/Cache.php#L184

@stof do you agree?

Tezd:
If you will find that https://github.com/composer/composer/blob/9aadbe27283bda0ed0f149fa4304c245f517c369/src/Composer/Repository/Vcs/VcsDriver.php#L119 requires change too. I will add it as a part of this PR.

stof:
@Tezd the VCSDriver indeed needs to be updated, to ensure consistency.

@alcohol the compiler should not be updated. The date here is a human-readable one, not a machine-readable one, so we don't care about the timezone IMO, and having `T` as separator between date and time is less readable than a space.
For the cache, I'm not sure it is needed to change it (this formatting is about configuring the Finder object, it does not impact the lock file at all)

Tezd:
@alcohol @stof Changed date time format to RFC_3399 in VcsDriver as well.

alcohol:
Thanks :-)

stof:
@Seldaek could this fix be shipped in a release any time soon ?

PR #6465, opened 2017-06-03
fixed:class_exists

rustphp:
here is an warning: failed to open stream: Too many open files in /data/www/ldsk/wz/vendor/composer/ClassLoader.php on line 444

if class has exist,not include file.


curry684:
How would this fix anything? `loadClass` is not invoked by definition on a class that is already loaded higher up in the chain.

alcohol:
You clearly misunderstood the error. Your proposed solution will not solve it :-)

rustphp:
yes,i have a mistake.
it's not composer's error!

PR #6475, opened 2017-06-11
Add support for feature-branches setting

NamelessCoder:
A negative list of non-feature-branches names
is already supported - this patch adds a list of
branches names which *will* be considered as
feature branches.

Allows changing the currently hardcoded set of
expected feature branch names, from:

* master|trunk|default|develop

To any set of names or patterns that you desire.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -893,4 +893,47 @@ Then "composer show -s" will give you `versions : * dev-latest-testing`.
 
 Optional.
 
+### feature-branches
+
+If your feature branch names are other than the default names:
+
+* master
+* trunk
+* default
+* develop
+
+You can use this setting to declare the list of branch names which should be considered
+feature branches in your package. For example, if you do not use `develop` but rather use
+a name like `development` and have no `trunk` or `default` branches, you can change this
+setting to:
+
+
+```json
+{
+    "feature-branches": ["master", "development"]
+}
+```
+
+And Composer will consider those two branches (plus any numeric branches) as your only
+feature branches. This allows Composer to recognise when you create a new branch based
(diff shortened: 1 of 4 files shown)


Seldaek:
Looks fine and lightweight enough patch that I guess it's ok to merge but I do wonder why people insist on not using standard names for things.. it just makes everything harder. Anyway thanks for the complete patch with docs and tests :)

NamelessCoder:
> but I do wonder why people insist on not using standard names for things

Thanks a lot for merging! I completely agree with this and I'd not recommend using nonstandard names - in my own case those branch names were chosen I don't know how long before composer days and having the option makes it completely painless to support those, whereas renaming the default branch would be a bit of a hassle. So I thank you a lot for agreeing to merge this.

Re: tests and docs you're very welcome :)

Seldaek:
@NamelessCoder soo I actually didn't realize when reviewing this initially, but while preparing the changelog for 1.5 the naming struck me as odd.. and indeed it was kinda written backwards. master/develop/.. are *not* feature branches. Feature branches are called like that for those temporary branches that you branch out to work on a new feature, then they get merged in master/develop/.. when done. So what you actually want to use is the existing non-feature-branches option, which was kinda broken though for your use case. I fixed it and removed feature-branches so there is no more confusion or duplication in https://github.com/composer/composer/commit/b07be842a033494419e42bf80680b5378bb7f525

NamelessCoder:
@Seldaek without having tested, that approach should also make it possible to solve my use case. The only thing it doesn't do is allow us to modify the list of default non-feature-branches (can only add names), but I can live with that - it's not a blocker for my use case. Thanks for the fix!

PR #6477, opened 2017-06-13
Example of disabling packagist.org globally

joking777:
Obviously this is not a common use case, but it is good for isolated networks. As far as I could tell, this was not documented, but I figured it out by looking through the code.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -655,4 +655,10 @@ You can disable the default Packagist.org repository by adding this to your
 }
 ```
 
+You can disable Packagist.org globally by using the global config flag:
+
+```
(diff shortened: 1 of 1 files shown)


doc/05-repositories.md
```diff
@@ -655,4 +655,10 @@ You can disable the default Packagist.org repository by adding this to your
 }
 ```
 
+You can disable Packagist.org globally by using the global config flag:
+
+```
+composer confg -g repo.packagist false
```
mbrodala: There's a typo here: `config`

Seldaek: Woops, thanks

joking777:
Thanks!

Issue #6480, opened 2017-06-14
Mechanism to disable the creation of .htaccess files


WilHall:
This is #5816.

Currently, composer uses the following block of code in `Composer\Factory` to ensure that the home, cache, and data directories have a `.htaccess` file present:
```
// Protect directory against web access. Since HOME could be
// the www-data's user home and be web-accessible it is a
// potential security risk
$dirs = array($config->get('home'), $config->get('cache-dir'), $config->get('data-dir'));
foreach ($dirs as $dir) {
    if (!file_exists($dir . '/.htaccess')) {
        if (!is_dir($dir)) {
            Silencer::call('mkdir', $dir, 0777, true);
        }
        Silencer::call('file_put_contents', $dir . '/.htaccess', 'Deny from all');
    }
}
```

In the majority of cases, this is not mitigating any security risk. Not all installations are under an Apache document root, and not all systems even run an Apache-based web server. Additionally, it doesn't feel like a package manager's responsibility to ensure that web server configuration files exist.

This ticket is a feature request to add a composer config which would make it possible to disable this feature.

For example, if I know that I do not run Apache on my system, I might like to set `COMPOSER_HTACCESS_PROTECT` to `false` so that these files are not created.

As mentioned on #5816, this _may be_ protecting some people, and I think it's good to air on the safe side for a default. Perhaps this setting should stay on by default, which wouldn't affect people who may be relying on it (knowingly or unknowingly) for security. But for users who do not need it, this would make it possible to disable.

alcohol:
Making it configurable could be a nice compromise vs. totally removing it (which I am still in favor of). Feel free to submit a pull request.

staabm:
Side-note: Apache since 2.4 does not read .htaccess files per default anymore.

PR #6486, opened 2017-06-15
Add htaccess-protect option for disabling the creation of .htaccess files

WilHall:
Fixes Issue #6480.

This pull request adds a `htaccess-protect` option to composer, which makes it possible to disable the creation of `.htaccess` files in the Composer home, data, and cache directories.

The option defaults to `true` in order to preserve the previous behavior for users who do nothing, in case anyone was relying on it (knowingly or unknowingly).

This option can also be overridden by the environment variable `COMPOSER_HTACCESS_PROTECT`.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -900,7 +900,7 @@ repository options.
 
 ### COMPOSER_HTACCESS_PROTECT
 
-Defaults to `true`. If set to `false`, Composer will not create `.htaccess` files in the
+Defaults to `1`. If set to `0`, Composer will not create `.htaccess` files in the
 composer home, cache, and data directories.
 
 &larr; [Libraries](02-libraries.md)  |  [Schema](04-schema.md) &rarr;
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -900,6 +900,7 @@ repository options.
 
 ### COMPOSER_HTACCESS_PROTECT
 
-Defaults to `true`. If set to `false`, Composer will not create `.htaccess` files in the composer home, cache, and data directories.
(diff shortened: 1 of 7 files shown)


doc/06-config.md
```diff
@@ -265,4 +265,10 @@ Example:
 }
 ```
 
+## htaccess-protect
+
+Defaults to `true`. If set to `false`, Composer will not create `.htaccess` files in the composer home, cache, and data directories.
+
+Previously, Composer unconditionally created these files to mitigate the potential for someone to expose these directories under their Apache document root. The default value of this option preserves the previous behavior.
```
stof: `Previously` does not make sense in the documentation. The doc is not a changelog. It documents the current state

doc/03-cli.md
```diff
@@ -898,4 +898,8 @@ If set to 1, this env changes the default path repository strategy to `mirror` i
 of `symlink`. As it is the default strategy being set it can still be overwritten by
 repository options.
 
+### COMPOSER_HTACCESS_PROTECT
+
+Defaults to `true`. If set to `false`, Composer will not create `.htaccess` files in the composer home, cache, and data directories.
```
alcohol: Please wordwrap like the rest of the document.

doc/06-config.md
```diff
@@ -265,4 +265,8 @@ Example:
 }
 ```
 
+## htaccess-protect
+
+Defaults to `true`. If set to `false`, Composer will not create `.htaccess` files in the composer home, cache, and data directories.
```
alcohol: Same as previous comment, wordwrap please :-)

alcohol:
Looks good to me.

Seldaek:
Thanks

PR #6500, opened 2017-06-22
changes related to new trusty images on travis

alcohol:


.travis.yml
```diff
@@ -30,15 +34,24 @@ matrix:
     - php: nightly
 
 before_script:
-    - rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini
-    - flags="--prefer-dist"
-    - composer install $flags
-    - bin/composer install $flags
-    - git config --global user.name travis-ci
-    - git config --global user.email travis@example.com
+    - |
```
stof: this change is not good. Usign a single YAML string instead of separate commands means that Travis will not check each actual commands for failures but only the last one

.travis.yml
```diff
@@ -30,15 +34,24 @@ matrix:
...
+
+      # install dependencies using composer from source
+      bin/composer install $flags
+
+      # make sure git does not complain about user/email not being set
+      git config --global user.name travis-ci
+      git config --global user.email travis@example.com
+
+script:
+    - |
```
stof: why switching to this format ?

alcohol: Easier to add some comments in between.

.travis.yml
```diff
@@ -30,15 +34,22 @@ matrix:
...
 
 before_script:
-    - rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini
-    - flags="--prefer-dist"
-    - composer install $flags
-    - bin/composer install $flags
-    - git config --global user.name travis-ci
-    - git config --global user.email travis@example.com
+    - # disable xdebug if available
+      phpenv config-rm xdebug.ini || echo "xdebug not available"
```
stof: I would keep the `-` on the line defining the value rather than on the line with the comment. It is more readable IMO

alcohol: I honestly have no clue what yaml allows and what it does not allow. Can I put the comments on an empty non `-` prefixed line above the `-` lines?

stof: yes

stof: you can have empty lines anywhere without any impact (except inside a multiline string block of course, where it impacts the string). And comments are ignored, making this back an empty line (the only thing to remember about comments is that the `#` must either be at the beginning of the line, or preceded by some whitespace char)

alcohol:
Still not sure why the tests involving cache directories apparently were not working properly on the precise images/environments. Suspect it is some kind of permission issue. You can still see that on the precise image we use for 5.3, the (functional test) composer install does not use an intermediate cache directories which is the default behavior if a cache directory is available/writable.

Seldaek:
Can we get this on 1.4 branch instead? Makes more sense I think if it improves the build, unless it's not applicable anymore to that file I'm not sure if they diverged.

alcohol:
Sure, shouldn't be too big of a problem.

stof:
@alcohol is there a way to re-enable xdebug before running the XdebugHandlerTest ? Otherwise, they will always be skipped on Travis.

alcohol:
@stof Good question, I am not sure. I can investigate that.

alcohol:
@stof actually, the tests seem to run (because they use the `XdebugHandlerMock` possibly?). I don't think they actually depend on xdebug being enabled. Correct me if I am wrong @johnstevenson ?

stof:
hmm indeed. They are only skipped on 5.3. I missed that

alcohol:
Some tests end up being killed on hhvm now though. Le sigh. More debugging required.

johnstevenson:
@alcohol Correct. The XdebugHandlerTests (on > 5.4 only) use the mock to check certain functionality, independently of xdebug being enabled.

PR #6501, opened 2017-06-22
Fix hg command to retrieve file content

Neurones67:
getFileContent generated command like hg cat -r hg cat -r CHANGESET FILE that caused errors "Skipping missing sub" while fetching composer.json files in repositories containing subrepositories.

This fixes that error.


Seldaek:
Thanks!

PR #6517, opened 2017-06-30
1.4

theseddd:


alcohol:
I think @Seldaek usually does this manually at a time he finds most convenient. Is there any specific reason you need 1.4 merged into master now?

PR #6524, opened 2017-07-03
Fix case insensitivity of ‘require’ command

svenmuennich:
When currently executing the `require` command for a package that is already listed in `require(-dev)`, one must use the exact same, case matching package name as written in `composer.json`. That is, if one changes the case of a character in the package name, the `require` command will add a new entry to `require(-dev)`, instead of updating the existing one:

```json
// Initial composer.json
{
  "require": {
    "foo/bar": "^1"
  }
}

// composer.json after running `composer require fOo/BaR:^2`
{
  "require": {
    "foo/bar": "^1",
    "fOo/BaR": "^2"
  }
}
```

This commit fixes the described behaviour to make it consistent with other commands like `update` that are already case insensitive:

```json
// composer.json after running `composer require fOo/BaR:^2` with this PR
{
  "require": {
    "foo/bar": "^2"
  }
}
```

### DIFF
--- a/src/Composer/Json/JsonManipulator.php
+++ b/src/Composer/Json/JsonManipulator.php
@@ -69,11 +69,15 @@ class JsonManipulator
 
         $links = $matches['value'];
 
-        if (isset($decoded[$type][$package])) {
+        // try to find existing link
+        $packageRegex = str_replace('/', '\\\\?/', preg_quote($package));
+        $regex = '{'.self::$DEFINES.'"(?P<package>'.$packageRegex.')"(\s*:\s*)(?&string)}ix';
+        if ($this->pregMatch($regex, $links, $packageMatches)) {
             // update existing link
-            $packageRegex = str_replace('/', '\\\\?/', preg_quote($package));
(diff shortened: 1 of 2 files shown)


svenmuennich:
I'm a little bit confused by the error of the CI tests:

```
1) Composer\Test\AllFunctionalTest::testIntegration with data set #0 (Symfony\Component\Finder\SplFileInfo Object (...))
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
 'Installing seld/jsonlint (1.0.0)
-  - Installing seld/jsonlint (1.0.0): Cloning 3b4bc2a96f
+  - Installing seld/jsonlint (1.0.0): Cloning 3b4bc2a96f from cache
```

It doesn't look like something this PR breaks, especially because I cannot reproduce this locally.

alcohol:
@svenmuennich that specific issue is solved in 1.4 but not yet merged into master. Also, since this is a bugfix, I think you should base your branch on 1.4 and not master (and assign 1.4 as merge target).

svenmuennich:
Thanks for the hint @alcohol, I updated everything accordingly.

Seldaek:
Thanks!

PR #6541, opened 2017-07-10
Update 04-schema.md

aicma:
Added line to try and clarify, that any dev/non-minimum-stability Requirement needs to be added to root composer.json as well.


Seldaek:
Thanks

PR #6545, opened 2017-07-12
Create bot.php

aprakob:
$response = $bot->replyText('<reply token>', 'hello!');
This procedure sends a message to the destination that is associated with <reply token>.


aprakob:
$textMessageBuilder = new \LINE\LINEBot\MessageBuilder\TextMessageBuilder('hello');
$response = $bot->replyMessage('<reply token>', $textMessageBuilder);
if ($response->isSucceeded()) {
    echo 'Succeeded!';
    return;
}

// Failed
echo $response->getHTTPStatus . ' ' . $response->getRawBody();

weitzman:
This PR makes no sense. Suggest deleting it.

PR #6548, opened 2017-07-12
Added $_SERVER['USER'] as possible value for <vendor>

ivanpepelko:
Added $_SERVER['USER'] as possible value for <vendor> name in init command.


staabm:
Why/when do you need it?

ivanpepelko:
Since I installed composer over package manager, owner of composer executable is `root`.
I have to retype full vendor/package name on every init.

Seldaek:
Thanks

PR #6568, opened 2017-07-26
mute first cd in win proxy stub for sh shell, fixes #6567

### LINKED ISSUE #6567: binary stubs for windows leak "cd" output into $dir
on windows 10 with git bash using the binary proxy stubs (sh shell) caused me issues as it reported file not found (as the proxy is not working)

the actual file to resolve to (or more correct: it's directory from which it was going to be invoked) - which looked correct - was somewhat prefixed with another directory and a space.

when looking into the issue it became apparent to me that the line which resolved the current directory:

    dir=$(d=\${0%[/\\\\]*}; cd "\$d"; cd $binDir && pwd)
                            ^^^^^^^^

was putting that directory in from the first `cd` command in front of the output of `pwd`.

ktomk:
in the sh script, the first cd command did output the directory it cd'ed
into did output the path itself.

fix for me is to redirect it's standard output to /dev/null

using git bash on a windows 10 box.

w/o this fix, the workaround is to invoke the .bat file which is created
as well (but this is not so handy as a non windows user for me).

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -194,7 +194,7 @@ class BinaryInstaller
         $proxyCode = <<<PROXY
 #!/usr/bin/env sh
 
-dir=$(d=\${0%[/\\\\]*}; cd "\$d"; cd $binDir && pwd)
+dir=$(d=\${0%[/\\\\]*}; cd "\$d" > /dev/null; cd $binDir && pwd)
 
 # See if we are running in Cygwin by checking for cygpath program
 if command -v 'cygpath' >/dev/null 2>&1; then


ktomk:
FYI: I didn't test the fix by re-doing the install with a patched installed, but I took it from the created proxy file I edited manually.

ktomk:
I must admit that I have no clue at all why that HHVM test fails (as reported by Travis). IMHO it should not, perhaps it's missing some documentation wheat it means if that test fails.

Refs:

- https://travis-ci.org/composer/composer/jobs/257935789#L754

- https://github.com/composer/composer/blob/14c77130e8f717ceb154720e1abefafe99124cb9/tests/Composer/Test/DefaultConfigTest.php#L17-L27

Seldaek:
I kinda wonder why `cd` echoes for you. I have never seen that, except when doing `cd -`. Do you have CDPATH defined? If so what is it set to?

I guess it's harmless to merge though, but trying to understand the problem first..

Seldaek:
Oh I just saw #6567  which explains in more details. Merging then..

ktomk:
Thanks and yes, it's easy to oversee, the behaviour however is documented, see [STDOUT section of the cd utility](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cd.html#tag_20_14_10) - that should be the normative reference for this (from *The Open Group Base Specifications Issue 7* /  *IEEE Std 1003.1-2008, 2016 Edition*).

PR #6578, opened 2017-07-30
Allow to specific custom version for path repositories

vtsykun:
### Purpose

As developer I want to ignore guess version for path repository and specific it in main composer.json. For example I develop new feature for composer installable package (local branch name "feature/example") and want to test it on my project. But I can not install this package without modification main composer.json or composer.json in this package.
These changes allow to ignore guess version.

**Second example** 
I want to install for testing "symfony/symfony": "dev-master", but some package requires a specific symfony version.
```

  Problem 1
    - incenteev/composer-parameter-handler v2.1.2 requires symfony/yaml ~2.3|~3.0 -> satisfiable by symfony/symfony[v3.3.5], s
```
But I use this approach, I can circumvent these requirements and install master from path
selection_073

### Usage
```json
    "repositories": [
        {
            "type": "path",
            "url": "package/*",
            "options": {
                "versions": {"symfony/symfony": "3.3.6"}
            }
        }
    ]
```

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -637,7 +637,14 @@
                 "options": {
                     "type": "object",
                     "properties": {
-                        "symlink": { "type": ["boolean", "null"] }
+                        "symlink": { "type": ["boolean", "null"] },
+                        "versions": {
+                            "type": "object",
+                            "description": "This is a maps of package name (keys) and version.",
+                            "additionalProperties": {
+                                "type": "string"
+                            }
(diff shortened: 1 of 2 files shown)


Seldaek:
The way to do this IMO is to use [aliases](https://getcomposer.org/doc/articles/aliases.md) so you can require `"symfony/symfony": "dev-master as 3.3.1000"` for example and keep the path repo untouched.

PR #6582, opened 2017-08-02
Improve BitBucket driver usage documentation

jasonquick:
Clarified and expanded the documentation for private repositories using the BitBucket driver. Emphasized the use of the COMPOSER_AUTH variable to encourage people not to save their credentials in composer.json. Pointed out that git endpoints don't seem to work.


stof:
auth info should be stored in the separate auth.json file, not the composer.json, to allow keeping it outside the VCS

jasonquick:
Right, I've clarified that point.

jasonquick:
I don't see how the HHVM failure is anything I did, it looks like the build took a really long time and timed out. Can we re-run the build?

Seldaek:
Thanks

PR #6595, opened 2017-08-08
Fix 1.5.0 ancestor dir composer.json search not using COMPOSER env var

dzuelke:


src/Composer/Command/ConfigCommand.php
```diff
@@ -146,7 +146,7 @@ protected function initialize(InputInterface $input, OutputInterface $output)
         // passed in a file to use
         $configFile = $input->getOption('global')
             ? ($this->config->get('home') . '/config.json')
-            : ($input->getOption('file') ?: trim(getenv('COMPOSER')) ?: 'composer.json');
+            : ($input->getOption('file') ?: Factory::getComposerFile());
 
         // Create global composer.json if this was invoked using `composer global config`
         if ($configFile === 'composer.json' && !file_exists($configFile) && realpath(getcwd()) === realpath($this->config->get('home'))) {
```
Seldaek: This first comparison will fail here as Factory returns `./composer.json` by default. Maybe should check for both..

src/Composer/Console/Application.php
```diff
@@ -127,13 +127,13 @@ public function doRun(InputInterface $input, OutputInterface $output)
...
         // prompt user for dir change if no composer.json is present in current dir
-        if ($io->isInteractive() && !in_array($commandName, array('init', 'about', 'help', 'diagnose', 'self-update', 'global'), true) && !file_exists('./composer.json')) {
+        if ($io->isInteractive() && !in_array($commandName, array('init', 'about', 'help', 'diagnose', 'self-update', 'global'), true) && !file_exists(Factory::getComposerFile())) {
             $dir = dirname(getcwd());
             $home = realpath(getenv('HOME') ?: getenv('USERPROFILE') ?: '/');
 
             // abort when we reach the home dir or top of the filesystem
             while (dirname($dir) !== $dir && $dir !== $home) {
-                if (file_exists($dir.'/composer.json')) {
+                if (file_exists($dir.'/'.Factory::getComposerFile())) {
```
Seldaek: This will break horribly if you pass a full path to a composer.json in COMPOSER and not just a different filename, but hopefully if you do that the path exists and this isn't triggered anyway..

src/Composer/Command/ConfigCommand.php
dzuelke: I, eh, don't understand :)

dzuelke: Oh wait. Your comment was on another line. Got it.

dzuelke: To be honest I don't quite understand that line anyway. For `composer global`, we want to ignore the `COMPOSER` env var, right? That `if` seems broken.

Seldaek: It just initializes an empty composer.json if you are in the global directory (not always as you can run `composer config -l` from anywhere for example and it shouldn't blow up.. 

And well there are two operation modes here, one is `composer global config` (change the global composer.json), the other is `composer config -g` (change the config.json that's in the home dir). Anyway I can fix it :p

PR #6608, opened 2017-08-12
Teach FossilDriver to work with local fossil repository files

kthbit:
Fossil driver was unable to work with urls pointing to a my-repo.fossil files. It only allowed urls to point to local checkout directories. This meant that user had to create checkout directories for all repos referenced in composer.json, even if origin repo fossil file is accessible via local filesystem. This patch attempts to fix this.

The old functionality is still there. Users can still point the fossil url to an open checkout directory. But now, users can also specify local (or network, UNC) file paths in the url.

The logic is as follows:
1. if the url points to a local directory, assume it's a local checkout directory
2. otherwise, the url has to be something that fossil can clone from


Seldaek:
Thanks, looks fine to me but I have zero experience with fossil so I'll just have to take your word for it. 

/cc @bohwaz who was the original FossilDriver author maybe you wanna have a quick look?

bohwaz:
Thanks for the patch @kthbit I'll take a look this week if I have time to confirm that it works as intended, but looks ok so far.

bohwaz:
After testing seems to be working as intended, thanks :)

kthbit:
@bohwaz Thanks for talking the time to look at this.

PR #6613, opened 2017-08-16
Note for aliases

decadence:
When you run Composer without args it prints all commands.
However documentation says nothing about some of them.

And note for short names.


Seldaek:
Thanks

PR #6632, opened 2017-08-25
Document how to define multiple versions in a package repository

### LINKED ISSUE #6627: Document how to define multiple versions in a package repository
The documentation for package repositories at https://getcomposer.org/doc/05-repositories.md#package-2 should include an explanation for how to define multiple versions using an array of package definitions. This works as follows:

```
{
    "repositories": [
        {
            "type": "package",
            "package": [
                {
                    "name": "foo/bar",
                    "version": "1.0.0",
                    ...
                },
                {
                    "name": "foo/bar",
                    "version": "2.0.0",
                    ...
                }
            ]
        }
    ]
}
```

Deamon:
This PR is to Fix #6627.

I'm open to any suggestions on this. 

Do I need to provide a full working example ? or leaving @naderman example as is is fine ?

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -518,6 +518,30 @@ Typically you would leave the source part off, as you don't really need it.
 >   reference you will have to delete the package to force an update, and will
 >   have to deal with an unstable lock file.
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Think that's fine, thanks :)

PR #6637, opened 2017-08-29
Delete last coma if deleting last element before end of file

### LINKED ISSUE #6636: Composer config --unset homepage breaks composer.json syntax
On a regular `composer.json` file, run:

`composer config homepage https://www.myurl.com/`

The `homepage` variable will be added at the **bottom** of the `composer.json` file:
```
          [...]
           }
        }
    },
    "homepage": "https://www.myurl.com/"
}
```
Whereas any other variable I set (`description`, `license`, etc) will be set at the **top** of the file.

Now, if I try to unset the `homepage` variable:
`composer config --unset homepage`

The resulting `composer.json` file will be:
```
           [...]
          }
        }
    },
    }
```

As you can see, it removed the line, but not the extra comma and curly braces.

Now, if I try to set any other variable:
`composer config description 'A Description' `

it will result in:
```
[Seld\JsonLint\ParsingException]
  "./composer.json" does not contain valid JSON
  Parse error on line 102:
  ...       }    },    }
  --------------------^
  Expected: 'STRING' - It appears you have an extra trailing comma
```

I've been browsing around opened/closed issues and found nothing related, so I apologise in advance if there's another issue already opened/solved about this and didn't see it, and greatly appreciate it if you can point me to that issue number.

I'm using composer `1.5.1`, by the way.

Deamon:
This PR aims to fix #6636.

I worked on the deletion of the last element before ending `}` that was breaking the file by leaving the coma.

There must be an other way to fix this, so just let me know if you prefer an other way.

### DIFF
--- a/src/Composer/Json/JsonManipulator.php
+++ b/src/Composer/Json/JsonManipulator.php
@@ -430,6 +430,11 @@ class JsonManipulator
                 return false;
             }
 
+            //check if we do not left a coma alone on previous line if it was the last line
+            if(preg_match('#,\s*$#', $matches['start']) && preg_match('#^\}$#', $matches['end'])) {
+                $matches['start'] = rtrim(preg_replace('#,(\s*)$#', '$1', $matches['start']), $this->indent);
+            }
+
             $this->contents = $matches['start'] . $matches['end'];
             if (preg_match('#^\{\s*\}\s*$#', $this->contents)) {
                 $this->contents = "{\n}";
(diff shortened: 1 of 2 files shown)


Seldaek:
Can you please add a test to the JsonManipulatorTest cases? This really has to be bullet proof tested because it's so easy to mess up a regex without realizing.

Seldaek:
Thanks, merged in 1.5 & master!

PR #6644, opened 2017-08-31
Introduces a new method `copy()`.

codekandis:
Some packages, e. g. `tm/tooly-composer-script`, are using the composer classes to e. g. create symlinks or perform other file operations. While there's only a `copyThenRemove()` method this commit introduces a new `copy()` method. `copy()` behaves the same as the copy part of `copyThenRemove()` did with one exception: it returns `true` on success and `false` on failure. Copying a directory may lead to a `false`, while the whole directory or some of its files couldn't been copied. To ensure backwards compatibility `copyThenRemove()` calls `copy()` now. This commit also adds the necessary tests.

### DIFF
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -247,27 +247,43 @@ class Filesystem
      */
     public function copyThenRemove($source, $target)
     {
+        $this->copy($source, $target);
         if (!is_dir($source)) {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

codekandis:
It's pleasure to contribute for this very awesome tool. I mean ... who does NOT rely on it? So thank you!

Issue #6426, opened 2017-05-18
Look “up” for composer.json if one doesn’t appear in the current folder


rachellawson:
When I run this command:

```
cd some/sub/folder
composer require xxx/yyy
```

And I expected the dependency to be added to the project’s composer dependencies.

But, I actually got a new, empty, composer.json file in the current folder.

It seems to me that we could have composer look to see if there is a composer.json file in the current folder and, if not, “look upwards” to see if one can be found. If it is, and the user confirms, it should add the dependency there.

Essentially, this should be the behaviour for all composer commands except `composer init`, I think

bojanz:
Even if we don't want to look up, we should at least stop execution, cause the result will be unusable.

alcohol:
This is the expected behaviour though. `composer require` will create a `composer.json` from scratch for you if there is not one currently in your working directory. Altering this behaviour would break backwards compatibility.

bojanz:
But doesn't Composer operate under the assumption that each project has a root composer.json?
Below that level running composer require does nothing cause the downloaded files never reach the autoloader. And there's already composer init for initializing a new composer.json.
I'm not convinced there's any convenience gained by sacrificing this much UX.

alcohol:
Composer does not, and has never, searched through a directory tree for the nearest `composer.json`. The current working directory is the only directory that matters. One can tell composer explicitly to use a given working directory through the `-d|--working-dir` flag.

rachellawson:
Yes, I'm suggesting it could change that behaviour, to avoid errors by humans. 

The is no sense in a sub directory of a project to create a new composer.json file for any operation other than `composer init`, surely? It might seem like a convenience but, in practice, it just becomes an inconvenience. 

At the very least we could ask for a confirmation if the file does not exist at that location?

bojanz:
@alcohol 
Yes, and it's fine not to search the directory tree.
What's not fine is accepting a "composer require" in a directory that produces no results to the user or the system.
I would either expect a hard error "hey, you're in the wrong directory" or at least a warning "looks like you're not in your project root, are you sure you want to continue".

jvasseur:
@bojanz but it does produce a result, it creates an empty `composer.json` in the current directory and add the dependency to it.

alcohol:
> The is no sense in a sub directory of a project (sp)

[...]

PR #6651, opened 2017-09-03
Prevent automatic parent directory composer.json discovery if working-dir parameter is set

Ayesh:
#6426 - _Look “up” for composer.json if one doesn’t appear in the current folder_ is a massive UX improvement. However, I believe that Composer should listen to, if set, the `--working-dir` option and not try to discover `composer.json` files in parent directories. 

One use case is at bamarni/composer-bin-plugin#30. This plugin lets you have namespaces and have a separate `composer.json` file in each namespace. With Composer 1.5, users are now prompted if they want to use the upper directory's `composer.json` file (which they should not, in this plugins use case). I'm trying to change this plugin so the plugin can set the working directory and force composer to create the `compsoer.json` file at the same directory.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -127,7 +127,7 @@ class Application extends BaseApplication
         }
 
         // prompt user for dir change if no composer.json is present in current dir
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks! Merged in 1.5 and master as eb078b432fe8d77d2c8f169dd02c5d9923343e2b

PR #6653, opened 2017-09-04
fix parallel test exit scenario shell syntax

glensc:
previous code introduced in 4d8b371908 used subshell capture which would be executed, but intended is just to set exit code to 1. probably never noticed because empty command is not executed by current shell.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,4 +55,4 @@ before_script:
 
 script:
     # run test suite directories in parallel using GNU parallel
-    - ls -d tests/Composer/Test/* | parallel --gnu --keep-order 'echo "Running {} tests"; ./vendor/bin/phpunit -c tests/complete.phpunit.xml --colors=always {} || (echo -e "\e[41mFAILED\e[0m {}" && $(exit 1));'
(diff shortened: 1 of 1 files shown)


Seldaek:
Not sure what this fixes exactly because when tests fail we definitely get a failed exit code so this seems to work?

glensc:
it fixes wrong shell code. read the explanation, it also explains why it appears to work for you.

Seldaek:
I read the explanation, but I don't understand it, which is why I ask.. 

If we get a proper failure when tests fail, yet if as you say `$(exit 1)` is not executed at all, then we could just remove it entirely?

glensc:
i'm not saying `exit 1` is not executed, i'm saying you capture output of sub-shell which runs `exit 1` and try to execute it, but as `exit 1` outputs nothing, then `$()` will be executed, which does nothing. but it's wrong, wrong intention.

run yourself:
```
sh -c '$()' 
```

Seldaek:
Ok I see what you mean now, thanks for explaining again. Makes sense :)

PR #6662, opened 2017-09-06
Fixed error, when composer.phar could be broken

fkryvyts:
Fixed error, when composer.phar was broken, if it was compiled inside folder, that has /composer in path. For example, if project repository will be located in /composer directory on linux, then str_replace will produce wrong result for such files, as /vendor/composer/autoload_real.php, because it will replace all occurrences of "/vendor" inside path, and not only the first one.

### DIFF
--- a/src/Composer/Compiler.php
+++ b/src/Composer/Compiler.php
@@ -164,10 +164,24 @@ class Compiler
         $util->save($pharFile, \Phar::SHA1);
     }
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Nice catch, thanks. Merged in 1.5 and master as dc509b236fad2604d12b251c69884a9f3b3f4173

PR #6667, opened 2017-09-08
Pass a VCS filtered iterator to mirror

GawainLynch:
Use case: Avoiding build cruft being deployed when using path repositories.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Downloader;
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, never noticed this as it usually symlinks but it's indeed much better. Just had to fix a windows regression (8b61315b1). Merged in 1.5 and master as b859bbcdcf3bf4ef168235fc63b67f49c7a21356

GawainLynch:
Yeah, we have a user (on Windows interestingly enough) that was getting very slow mirrors due to very large `node_modules` directories that didn't need to be deployed. 

Thanks for the regression fix too, that would no doubt have bitten us :wink:

PR #6671, opened 2017-09-09
Trying to find package with similar name when package could not be found

lulco:
I'm trying to solve problem with typo in package name.

I would like to see some "Did you mean" message in console with suggested existing package name.

```
composer require lulco/fenix

[InvalidArgumentException]
Could not find package lulco/fenix. Did you mean: lulco/phoenix ?
```

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -13,20 +13,22 @@
 namespace Composer\Command;
 
 use Composer\DependencyResolver\Pool;
-use Composer\Json\JsonFile;
 use Composer\Factory;
-use Composer\Repository\RepositoryFactory;
+use Composer\Json\JsonFile;
 use Composer\Package\BasePackage;
+use Composer\Package\PackageInterface;
 use Composer\Package\Version\VersionParser;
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks for the PR, however iterating over all available packages is just not efficient enough and in the long term might have to be deprecated entirely, so I'd rather not add new functionality that depends on it. I'll have to close this sorry.

lulco:
Hi, I know that iterating over all packages is not efficient enough, but I want to say, that it is called only if no candidate was found and printing "Did you mean" message could be big time saver for developers who have to find correct package name on github or packagist etc.

Seldaek:
I see that but my problem is more that the API might go away entirely in the future.. Perhaps a better way would be to use the search API of the RepositoryInterface to search for the package name - then print out the first 5 results or so? It's a much faster API request and yields good results IMO in this case: https://packagist.org/search/?q=lulco%2Ffenix

lulco:
OK thanks for hint, I will try to implement it with search API

lulco:
@Seldaek I've created new PR #6678 using search API of RepositoryInterface

PR #6672, opened 2017-09-11
Allow custom script descriptions.

thom8:
Allows you to add custom descriptions for custom scripts which is useful for providing some basic inline documentation.

```json
    "extra": {
        "scripts-description": {
            "test": "Run all tests!"
        }
    }
```

Each script can have it's own description set in `extra.scripts-description` and it will still show the default if nothing is set.

image


Seldaek:
Thanks!

PR #6674, opened 2017-09-11
Fixes #6661: Allow a given package and its dependencies (including siblings) to be updated.

### LINKED ISSUE #6661: Allow a given package and its dependencies (including siblings) to be updated.
This feature request is intended to address a very common difficulty with updating composer packages. It is inspired by a large number of support requests.

###  Summary

It is not possible to update a single package and its dependencies if one of those dependencies is also defined in the root composer.json file.

### Scenario

My composer.json file defines the following:

```json
{
    "require": {
        "acme/package1": "^1.0.0",
        "acme/package2": "^1.0.0",
        "unrelated/package1": "^1.0.0"
     }
}
```

Currently installed versions are:

* acme/package1: 1.0.0
* acme/package2: 1.0.0

I would like to update acme/package1 to 1.0.1,  which happens to rely on acme/package2 1.0.1.

I'd like to perform this update using a single command that updates only acme/package1 and its dependencies. Unfortunately, this is not possible with `composer update acme/package1` or `composer update acme/package1 --with-dependencies`. The definition of `"acme/package2": "^1.0.0"` in my composer.json prevents the update from executing, despite the fact that its version constraint is compatible with both new versions of the new packages.

The upshot here is that one needs to either manually traverse the dependency tree and figure out which packages should be manually updated, or else execute a wholesale `composer update`,  which can have unintended consequences by updating more packages than are necessary.

###  Suggested solutions

1. Modify the `--with-dependencies` option such that it is able to update packages that are defined in the root composer.json file.
1. Introduce a new option such as `--with-dependencies-and-siblings` (or something shorter) that would provide this capability.

grasmash:


Seldaek:
Thanks!

PR #6675, opened 2017-09-11
SCA with Php Inspections (EA Extended)

kalessil:
Minor code tweaks:
- alias functions usage fixes
- is_null -> null ===
- one-time use variables
- array_push ->  ` []= `
- more minor stuff

PS:`break; // or `continue 2` ?` - I'm not following the intention there


kalessil:
I need to find out why tests are failing...

PR #6677, opened 2017-09-11
SCA with PHP Inspections (EA Extended)

kalessil:
Minor code tweaks:
- alias functions usage fixes
- is_null -> null ===
- one-time use variables
- array_push -> []=
- more minor stuff

PS: `break; // orcontinue 2?` - I'm not following the intention there

### DIFF
--- a/src/Composer/DependencyResolver/RuleSetGenerator.php
+++ b/src/Composer/DependencyResolver/RuleSetGenerator.php
@@ -283,7 +283,7 @@ class RuleSetGenerator
             switch ($job['cmd']) {
                 case 'install':
                     if (!$job['fixed'] && $ignorePlatformReqs && preg_match(PlatformRepository::PLATFORM_PACKAGE_REGEX, $job['packageName'])) {
-                        continue;
+                        break; // or `continue 2` ?
                     }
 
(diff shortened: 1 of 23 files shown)


src/Composer/DependencyResolver/RuleSetGenerator.php
```diff
@@ -283,7 +283,7 @@ protected function addRulesForJobs($ignorePlatformReqs)
             switch ($job['cmd']) {
                 case 'install':
                     if (!$job['fixed'] && $ignorePlatformReqs && preg_match(PlatformRepository::PLATFORM_PACKAGE_REGEX, $job['packageName'])) {
-                        continue;
+                        break; // or `continue 2` ?
```
Seldaek: Ah interesting, I never realized `continue` applied to switch statements. Then again in this case, since the loop only contains the switch, breaking out of the switch or continuing to the end of the switch or continuing to the end of the loop all have the same effect. I guess `continue 2` was the intent though, but `break` is fine too.

Seldaek:
Thanks

kalessil:
Oh, that was fast, thank you Jordi.

PR #6678, opened 2017-09-11
Search similar packages when package could not be found

lulco:


landsman:
This is can be very helpful with my problem about installing packages like:
`composer require nette/tracy` ... when right package is: `tracy/tracy`.

Seldaek:
Thanks!

PR #6682, opened 2017-09-12
workaround for gitlab installation with relative url

itscaro:
Use the domain declared in gitlab-domains as originalUrl when downloading archives


Seldaek:

PR #6683, opened 2017-09-13
fix gitlab repo config in composer.lock

RobberPhex:
1. When `attemptCloneFallback` use `generatePublicUrl` to get public_url, we cannot forget some gitlab instance doesn't support https. **So we get `scheme` from config.**
2. When generate `composer.lock` file, we should use repo url *what user wrote* at `composer.json` , not we got from gitlab api.

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -351,7 +351,7 @@ class GitLabDriver extends VcsDriver
 
     protected function generatePublicUrl()
     {
-        return 'https://' . $this->originUrl . '/'.$this->namespace.'/'.$this->repository.'.git';
+        return $this->scheme . '://' . $this->originUrl . '/'.$this->namespace.'/'.$this->repository.'.git';
     }
 
     protected function setupGitDriver($url)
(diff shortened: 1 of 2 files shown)


RobberPhex:
btw, I don't know `getRepositoryUrl`'s usage?
1. There is no one use `getRepositoryUrl`
2. this method is only on `GitHubDriver` and `GitLabDriver`
3. `VcsDriver` and `VcsDriverInterface` haven't method `getRepositoryUrl`.

so I think we can delete this method.

*or, add this method to Interface?*

---

https://github.com/composer/composer/blob/e12e0335a4d467878093fc76e3dea592ff3e1ce3/src/Composer/Repository/Vcs/GitLabDriver.php#L163-L166

RobberPhex:
rebased master

RobberPhex:
@Seldaek I think this is a bug, Could you please confirm this?

Seldaek:
The scheme fix I agree is good and I will apply it to the next 1.5 patch release. The rest I am not so sure.. The current state of things is that if you access private repos you need ssh access, and I think that's a good thing as configuring token access is quite a pain. We have gone back and forth on this and given the users the choice is an option of course but it'll also maybe break things for people unexpectedly which is not great, plus it would need to be clearly documented as right now is doens't matter which kind of gitlab URL you use you always get the same result which is easier IMO. I'll leave this open for now in case anyone else has inputs but I am not so keen on the change overall.

Issue #4410, opened 2015-09-12
Some words about installed.json


olvlvl:
Hi,

I can't find information about the `installed.json` file on [getcomposer.com](http://getcomposer.com/), it's a shame because the file looks very interesting. Are packages ordered by weight in this one? That would be great because I wouldn't have to [_cheat_](https://github.com/ICanBoogie/ICanBoogie/blob/3.0/lib/Autoconfig/FakeAutoloadGenerator.php) in order to get packages and that would save me [a lot of work](https://github.com/ICanBoogie/ICanBoogie/blob/3.0/lib/Autoconfig/Hooks.php#L26) during `autoload-dump` event.

Can you comment on this feature and maybe add it to the doc?

Best,

Olivier

Seldaek:
I don't think you can or should really rely on this file's ordering, although they might be after an initial install, I am not sure if that remains true after updates, and we certainly offer no guarantees of that right now.

olvlvl:
Well that's too bad because guarantying that the `installed.json` file is created from ordered packages would allow third parties to parse the file without relying on a composer event and resorting to sorcery: https://github.com/ICanBoogie/ICanBoogie/blob/3.0/lib/Autoconfig/Hooks.php#L26

We could make sure that packages are ordered in `installed.json` and that it stays that way.

Seldaek:
Yes that would be an option, if you'd like to look into it, should just be a matter of ordering things in the FilesystemRepository::write() I believe, and extracting that sorting logic out of the AutoloadGenerator into a re-usable standalone class as well probably.

olvlvl:
I'm already investigating :)

reedy:
+1 to keeping it in order. It's annoying for repos where you commit the resultant repo for it to remove the whole block from installed.json and append it to the bottom. It's just extra noise in the diff

The lock file keeps things in order, and updates the relevant rows, which is much easier to review

olvlvl:
@reedy The thing is, `installed.json` (like the `vendor` directory) shouldn't be commited, only `composer.lock`.

reedy:
Yet, it seems many people do commit the vendor repo for varying reasons. See also #4717

https://getcomposer.org/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md

"The general recommendation is no."


[...]

PR #6696, opened 2017-09-20
Sort vendor/composer/installed.json deterministically

legoktm:
Just like composer.lock, sort installed.json in order of the package
names. This makes it easier to review diffs when this file is changed.

See discussion on #4410.

### DIFF
--- a/src/Composer/Repository/FilesystemRepository.php
+++ b/src/Composer/Repository/FilesystemRepository.php
@@ -83,6 +83,10 @@ class FilesystemRepository extends WritableArrayRepository
             $data[] = $dumper->dump($package);
         }
 
+        usort($data, function($a, $b) {
+            return strcmp($a['name'], $b['name']);
+        });
+
         $this->file->write($data);
     }
 }


olvlvl:
I'd rather have packages sorted by weight so the `installed.json` file is actually useful. I can sort packages alphabetically myself, that's super easy to do, sorting them by weight that's another story, I need composer to do that.

reedy:
How would you define weight?

olvlvl:
The weight is already computed by composer to determine the order in which packages should be installed.

stof:
@olvlvl the weight is not computed for the whole list of installed packages, but for operations being performed. If you already have some installed packages before, they won't be included in the computation of weight (as the have no operation).
And the code maintaining the list of installed packages does not have access to this weight anyway.
So there cannot be such sorting.

The sorting by name would only be useful in this file to ensure stability for people committing the vendor folder (and so looking at diffs for this file).

olvlvl:
Right now I'm calling `sortPackageMap()` myself on `post-autoload-dump`. I wish I could just read sorted packages from `installed.json`.

https://github.com/ICanBoogie/ICanBoogie/blob/4.0/lib/Autoconfig/FakeAutoloadGenerator.php

frob:
This will make my life much easier.

Seldaek:
If we sort the whole list by weight it should also be deterministic tbh. Some things might flip of course but it wouldn't be as random as it is currently. That might satisfy both sides?

legoktm:
I think that should work out fine :)

frob:
I actually think weight would be ideal, but I am on board with either. 

Weight looked like more effort and I wouldn't want to let perfect be the enemy of good when we have a working PR.

stof:
@Seldaek weight is for operations, not for packages. Sorting this file would then require a dedicated computation of the weight taking all packages into account (and it would allow random changes in the diff when adding a new package with a dependency to an existing package)

Seldaek:
Yeah I guess the simplicity and reliability of this name-based PR makes it better than doing the weight based one. I think it can be merged in for 1.6

olvlvl:
Could we have an option to determine how the `installed.json` file is created, with a `sort` option in `composer.json` maybe? Because it looks like a project preference.

Seldaek:
Yet another option for such a minor edge case use case doesn't sound that great to me tbh :/

olvlvl:
Alright :)

danepowell:
I think this introduced at least one regression, for packages using Composer Installers. Generally, it could cause problems with any Composer plugin.

Prior to this, packages were written to `installed.json` in the same order they were installed, meaning that (for instance) if a plugin depended on Installers, you could always count on Installers being installed first, because `composer update` _and_ any other composer command would load plugins in the same order.

After this PR, plugins are still activated in the expected order during `composer update`, but subsequent composer command load plugins alphabetically. This means that a plugin with a name that comes alphabetically before "composer" will now load before Installers does, causing warnings due to missing classes. See https://github.com/composer/composer/issues/8065

frob:
Sounds more like it revealed a side-effect rather than introduced a regression.

PR #6707, opened 2017-09-26
Teach fossil to use HTTP Basic and Fossil built-in authentication

kthbit:
Composer was unable to clone/pull from a Fossil repo that was protected by either HTTP Basic or Fossil  authentication. This patch attempts to fix that. Both (or either) authentications are stored in auth.json. HTTP Basic auth is specified as before, per domain. Fossil credentials are specified per remote repository, in case pulling from multiple remote repositories with different credentials for each.

### DIFF
--- a/src/Composer/Downloader/FossilDownloader.php
+++ b/src/Composer/Downloader/FossilDownloader.php
@@ -131,15 +131,17 @@ class FossilDownloader extends VcsDownloader
         $parsed = parse_url($url);
 
         // Only for http(s) URLs.
-        if (!isset($parsed['scheme']) || ($parsed['scheme'] != 'http' && $parsed['scheme'] != 'https') || empty($parsed['host']))
+        if (!isset($parsed['scheme']) || ($parsed['scheme'] != 'http' && $parsed['scheme'] != 'https') || empty($parsed['host'])) {
             return '';
+        }
 
         $httpBasicConfig = $this->config->get('http-basic');
         if (isset($httpBasicConfig[$parsed['host']])) {
             $creds = $httpBasicConfig[$parsed['host']];
             // We don't allow empty username or password. Both must be defined.
-            if (!empty($creds['username']) && !empty($creds['password']))
+            if (!empty($creds['username']) && !empty($creds['password'])) {
                 return sprintf('-B %s', ProcessExecutor::escape($creds['username'] .':'. $creds['password']));
+            }
         }
 
         return '';
@@ -150,7 +152,7 @@ class FossilDownloader extends VcsDownloader
      *
      *   1. HTTP Basic authentication, if accessing the HTTP URL is so protected by the web server.
      *   2. Fossil authentication, if accessing the remote repo is not allowed for anonymous users.
-     * 
+     *
      * Fossil considers the username and password specified in the HTTP URL to be the credentials for Fossil authentication,
      * and provides a separate parameter for HTTP Basic authentication.
      *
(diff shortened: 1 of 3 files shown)


src/Composer/Downloader/FossilDownloader.php
```diff
@@ -113,4 +121,108 @@ protected function hasMetadataRepository($path)
...
+
+        // Only for http(s) URLs.
+        if (!isset($parsed['scheme']) || ($parsed['scheme'] != 'http' && $parsed['scheme'] != 'https') || empty($parsed['host']))
+            return '';
+
+        $httpBasicConfig = $this->config->get('http-basic');
+        if (isset($httpBasicConfig[$parsed['host']])) {
+            $creds = $httpBasicConfig[$parsed['host']];
+            // We don't allow empty username or password. Both must be defined.
+            if (!empty($creds['username']) && !empty($creds['password']))
```
stof: please add the curly braces. Not having them makes it much easier to introduce a mistake later (not realizing that the second line is not inside the if even though you intended it to be)

src/Composer/Repository/Vcs/FossilDriver.php
```diff
@@ -79,24 +79,28 @@ protected function updateLocalRepo()
...
-        if (is_file($this->repoFile) && is_dir($this->checkoutDir) && 0 === $this->process->execute('fossil info', $output, $this->checkoutDir)) {
-            if (0 !== $this->process->execute('fossil pull', $output, $this->checkoutDir)) {
+        if ($this->isLocalCheckout()) {
+            $cmd = $this->getPullCommand();
+            if (0 !== $this->process->execute($cmd, $output, $this->checkoutDir)) {
                 $this->io->writeError('<error>Failed to update '.$this->url.', package information from this repository may be outdated ('.$this->process->getErrorOutput().')</error>');
             }
-        } else {
+        }
+        else {
```
stof: please don't change the coding standards

src/Composer/Downloader/FossilDownloader.php
stof: this applies to the whole PR btw. I won't comment on all of them.

src/Composer/Downloader/FossilDownloader.php
```diff
@@ -113,4 +121,117 @@ protected function hasMetadataRepository($path)
...
+     *
+     * It is expected that credentials specific for each repository would be located in auth.json in the following structure:
+     *
+     * {
+     *   "fossil-auth": {
+     *     "host.com/path/to/repo": {
+     *       "username": "my-user",
+     *       "password": "my-pass"
+     *     }
+     *   }
```
Seldaek: Is it not possible to simply reuse the existing [http-basic](https://getcomposer.org/doc/06-config.md#http-basic) key for this and do it only based on hostname? Especially as you seem to have both http basic + fossil level auth, calling the http basic one fossil is quite confusing.

src/Composer/Repository/Vcs/FossilDriver.php
```diff
@@ -90,20 +91,159 @@ protected function updateLocalRepo()
...
+                $parsed['host'],
+                isset($parsed['port']) ? ':'.$parsed['port'] : '',
+                $parsed['path'],
+                isset($parsed['query']) ? '?'.$parsed['query'] : '',
+                isset($parsed['fragment']) ? '#'.$parsed['fragment'] : ''
+            );
+        }
+
+        return $url;
+    }
```
Seldaek: getHttpBasicAuth and getUrlWithAuth seem to be pretty much duplicated from the FossilDriver.. It'd be good to instead have a Util class to deduplicate the code.

Seldaek:
Closing for now, happy to receive an updated PR with the changes above.

PR #6714, opened 2017-09-29
add gitlab-dist-format option

hboomsma:
Add `gitlab-dist-format` configuration option for the `GitLabDriver`. It defaults to `zip` as is the current default. You can set it to `zip`, `tar.gz`, `tar` and `tar.bz2`.

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -73,6 +73,10 @@ A list of domain names and private tokens. For example using `{"gitlab.com":
 "privatetoken"}` as the value of this option will use `privatetoken` to access
 private repositories on gitlab.
 
+## gitlab-dist-format
+Type of archive to download from gitlab. Set to `zip`, `tar`, `tar.gz` or `tar.bz2`
+Defaults to `zip`.
+
 ## disable-tls
 
 Defaults to `false`. If set to true all HTTPS URLs will be tried with HTTP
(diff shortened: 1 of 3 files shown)


RobberPhex:
Why Need `tar.gz`, `tar`?

At `packagist.org`, almost every package provide zip dist file. So I prefer just download zip package from gitlab.

hboomsma:
@RobberPhex `tar.gz` and `tar` preserve the executable flag for binaries. Although you can list your script in the `bin` section of your `composer.json`, having it copied to the `bin` folder is not always what one would want.

RobberPhex:
Dist file is designed for distribute source. the package code shouldn't depend executable flag.

hboomsma:
@RobberPhex exactly, but if I clone the source from a git repository, the executable flags are preserved. And when the dist is used with `zip` format they are not. This means the behaviour of dist and source differs.

Seldaek:
IMO adding an option here is unnecessary complication. If tarballs make more sense than zip files, then we probably should switch by default to tarballs, for gitlab as well as github.

The main problem right now is the TarDownloader is kinda dumb, and only uses PharData which will require zlib to be present to work on tar.gz files. We have a GzipDownloader that does the right thing mostly with regard to using gzip from CLI if available, etc. But that one does not handle tar.gz, it'll only turn tar.gz files into tar files. So we'd need the TarDownloader to handle tar.gz files through the GzipDownloader first maybe by extending it. Something like that.

The problem is doing this means we change our requirements from system unzip OR zlib extension to system gzip OR zlib extension, so that might break things for some people out there.. Hence why I targetted it for 2.0.

hboomsma:
@Seldaek switching to `.tar.gz` looks good to me. I'm happy to wait for the next major. Thank you all for your time.

Seldaek:
Closing as it's replaced by https://github.com/composer/composer/issues/7997

Issue #6716, opened 2017-10-02
Composer should not send authorization header when following redirect


maartenba:
While there is no definite answer in the HTTP RFC's, it seems to be best practice not to forward authorization header across redirects. Currently, Composer does this, though:

Consider `auth.json` (and a repository in `composer.json` that lives at `hostname`):

```json
{
    "http-basic": {
        "hostname": {
            "username": "username",
            "password": "password"
        }
    }
}
```

What Composer does today:
* User requests package from `hostname/foo/1.0.0.zip`
* `hostname` redirects using `302` to location `otherhost/foo/1.0.0.zip`
* `otherhost` gets authorization header value

This should not happen, as we're potentially sending credentials to `otherhost` without this being intended. Instead, Composer should strip the header when redirecting away from the origin domain. Or in short: if there's a redirect from `foo.com` to `bar.com`, the header should not be forwarded. If the redirect is from `foo.com` to `content.foo.com`, the header may be forwarded.

Evidence in other places:
* https://lists.apple.com/archives/webkitsdk-dev/2011/Mar/msg00004.html
* https://github.com/request/request/issues/450
* https://stackoverflow.com/questions/17092259/should-authorization-be-kept-when-redirection-is-handled
* http://www.telerik.com/forums/authorization-header-and-redirects

A potential solution would be:
* If credentials for the origin are set, make sure `$options['http']['follow_location'] = 0;` and we're doing a userland redirect (so we can intercept the redirect).
* When handling the userland redirect, strip off the auth header if the original host != redirected host.

maartenba:
A potential solution would be:
* If credentials for the host are set, make sure `$options['http']['follow_location'] = 0;` and we're doing a userland redirect.
* When handling the userland redirect, strip off the auth header if the original host != redirected host.

If this is of interest, I'm up for doing a PR on it. Thoughts?

maartenba:
A PR fixing this is available here: https://github.com/composer/composer/pull/6717

PR #6717, opened 2017-10-02
Composer should not send authorization header when following redirect

maartenba:
Fixes the issue described in #6716

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -230,20 +230,20 @@ class RemoteFilesystem
         unset($tempAdditionalOptions);
         $userlandFollow = isset($options['http']['follow_location']) && !$options['http']['follow_location'];
 
-    if ($isRedirect && $this->io->hasAuthentication($this->originUrl)) {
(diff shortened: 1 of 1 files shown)


dulfe:
+1
Completely fixed the issue I had with authorization headers... thank you!

Seldaek:
Merged as https://github.com/composer/composer/commit/128e424c90b9aad4ed7a86dd43b182aaddfd60d1 and then I realized we actually can do it much simpler by using the correct host to check for credentials at every request instead of re-using the originUrl .. https://github.com/composer/composer/commit/743153e862b7b82bd71f09be9775f460ca8d6ac0

Hopefully this won't break anything though, it's kinda sensitive code.

maartenba:
@Seldaek do you happen to have the .phar somewhere?

alcohol:
`composer self-update --snapshot`

dulfe:
@Seldaek, it worked great. (I used @alcohol command to download the snapshot)

maartenba:
Verified it as well, works just fine. Thanks @Seldaek !

Seldaek:
Alright great thanks

PR #6727, opened 2017-10-07
style: order use statements alphabetically

DMNSteve:
Orders the use statements alphabetically in the source files for neatness.


Slamdunk:
I suggest to also enable [ordered_imports](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.0/src/Fixer/Import/OrderedImportsFixer.php) to .php_cs file.

DMNSteve:
Good thinking! added this.

Seldaek:
Thanks but we don't accept CS fix pull requests.

Slamdunk:
May I ask you why?
Would you accept a .php_cs.dist modification, and then you apply the fix manually after merge?

Seldaek:
Generally because I find it wasteful to have to deal with PRs which are "useless" functionally speaking. As for ordering use statements specifically I really don't see the point so no I'm not really keen on doing this.

DMNSteve:
That’s a shame because I think Composer is a great project and would be keen to contribute, but as a newcomer, this feels quite off-putting.

Slamdunk:
I agree with @DMNSteve 

For sure coding-style PRs take time that seem a waste of time: a lot of file changed, without any new functionality or bug fixed.

But code calisthenics is an essential part to make an OSS easy to read and contribute to.

Moreover there are tests that ensure that nothing is broken; I don't get why this would be considered a waste of time.

I strongly suggest to insert coding-standard check in the CI process.

PR #6735, opened 2017-10-11
Correctly extract username/password from URL

gregatcode:
Corrected regexp. Otherwise it was impossible to use urls with credentials **and** @ character somewhere in the path url segment - it was capturing the whole part of the url (down to the last @ char) as a password.

Not sure if i need to add unit test for this small edge case.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -207,7 +207,7 @@ class RemoteFilesystem
         $this->redirects = 1; // The first request counts.
 
         // capture username/password from URL if there is one
-        if (preg_match('{^https?://(.+):(.+)@([^/]+)}i', $fileUrl, $match)) {
+        if (preg_match('{^https?://([^:]+):([^@]+)@([^/]+)}i', $fileUrl, $match)) {
             $this->io->setAuthentication($originUrl, urldecode($match[1]), urldecode($match[2]));
         }
 


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -207,7 +207,7 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
         $this->redirects = 1; // The first request counts.
 
         // capture username/password from URL if there is one
-        if (preg_match('{^https?://(.+):(.+)@([^/]+)}i', $fileUrl, $match)) {
+        if (preg_match('{^https?://([^:]+):([^@]+)@([^/]+)}i', $fileUrl, $match)) {
```
stof: it must also forbid `/` in both of them (otherwise a URL with a `:` and a `@` in the query string could be matched as authenticated)

stof:
and yes, you should add a unit test. Such edge cases are precisely cases for which a unit test is the only way to avoid breaking them again in the future (it took years to identify this bug, and it could take years again to catch a regression)

src/Composer/Util/RemoteFilesystem.php
gregatcode: Exactly. Probably, parse_url() would be more feasible here.

Seldaek:
Thanks, merged in 1.5 and master

PR #6736, opened 2017-10-11
Return non-zero exit code with --strict and single package

mihaiplasoianu:
Currently, the --strict option is not taken into consideration when a package name is specified. This would be helpful for scripts that only need to check that a single package is up to date.


Seldaek:

PR #6740, opened 2017-10-12
also support NO_PROXY

croensch:
since WINDOWS can only have one case and that is usually upperase

I rely on the all-uppercase settings to get packages from the internet but also get some from our Satis server in the local network. You can have lowercase environment variables on Windows but not both. You can only flip them by unsetting one case before setting the case again (assuming satis.local in global config.json):

```
SET no_proxy=
SET NO_PROXY=satis.local
composer update
SET NO_PROXY=
SET no_proxy=satis.local
composer update
```


Seldaek:
Thanks!

PR #6743, opened 2017-10-14
Composer init default values

zachflower:
This pull request adds the ability to define default `vendor`, `author`, and `license` values when performing a `composer init` (as requested in #5357). These values are defined via a new global config file found at `$COMPOSER_HOME/defaults.json`:

```javascript
{
  "defaults": {
    "vendor": "zachflower",
    "author": {
      "name": "Zachary Flower",
      "email": "zach@zacharyflower.com"
    },
    "license": "MIT"
  }
}
```

The default configs defined in this file take precedence over all other interpreted config values (`.gitconfig`, currently logged in user, etc). The choice of creating a new config file (rather than utilizing the global `config.json` file) is guided by the fact that these default values are used to _create_ a new `composer.json` file, rather than drawn from an existing one.

## Relevant Issues
- `Default composer init settings` - #5357

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -154,11 +154,12 @@ EOT
         $git = $this->getGitConfig();
         $io = $this->getIO();
         $formatter = $this->getHelperSet()->get('formatter');
-
+        $config = Factory::createConfig($io);
+        $defaults = $config->get('defaults');
+        
         // initialize repos if configured
         $repositories = $input->getOption('repository');
         if ($repositories) {
-            $config = Factory::createConfig($io);
             $repos = array(new PlatformRepository);
             foreach ($repositories as $repo) {
                 $repos[] = RepositoryFactory::fromString($io, $config, $repo);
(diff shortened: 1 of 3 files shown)


naderman:
This feature seems a bit too much to me, see https://github.com/composer/composer/issues/5357#issuecomment-343140941

zachflower:
Implementation simplified as per @curry684's suggestion in https://github.com/composer/composer/issues/5357#issuecomment-343169568. If this feels like an acceptable approach, I will update the documentation as well.

zachflower:
To clarify (since the PR description is no longer accurate), the new config file was removed in lieu of 4 new environment variables: `COMPOSER_DEFAULT_VENDOR`, `COMPOSER_DEFAULT_AUTHOR`, `COMPOSER_DEFAULT_EMAIL`, and `COMPOSER_DEFAULT_LICENSE`. Each of which, when set, will update the suggested default when running `composer init`.

For example, here is current behavior:

screenshot 2017-11-13 14 02 33

And then with `COMPOSER_DEFAULT_VENDOR` environment variable set:

screenshot 2017-11-13 14 03 29

Seldaek:
Thanks

PR #6746, opened 2017-10-15
add php version and binary to the dignose command

renanbr:
For those that manage multiple PHP binaries (and multiple Composer binaries), sometimes we aren't sure which PHP nor Composer is running... specially when running it behind a tool, such an IDE or CI server.

This commit expose some informations that can help us to find and fix errors.


Seldaek:
Good idea yes, thanks

PR #6751, opened 2017-10-17
adds 2 steps install script for windows users

am05mhz:
just add a little convenience for windows users


Seldaek:
Why not do it in one step and if it's valid just include or eval the installer? The point of having a script is to automate it so it kinda should be a single step IMO.

PR #6779, opened 2017-11-02
Fix bug setting COMPOSER_ORIGINAL_INIS

johnstevenson:
This variable stores the loaded ini file and any additional scanned ini
files, separated by a path-separator. The loaded ini file should always
be present, even if it is an empty value. Unfortunately I removed any
empty value to parse the ini files, then used the truncated list to set
the variable.

This bug surfaced on docker php images. These do not have a specific
php.ini but store all their settings in the location(s) configured at
build time using --with-config-file-scan-dir.

This bug fixes https://github.com/sensiolabs/SensioDistributionBundle/issues/322

### DIFF
--- a/src/Composer/XdebugHandler.php
+++ b/src/Composer/XdebugHandler.php
@@ -139,11 +139,6 @@ class XdebugHandler
         $iniPaths = IniHelper::getAll();
         $additional = count($iniPaths) > 1;
 
-        if (empty($iniPaths[0])) {
-            // There is no loaded ini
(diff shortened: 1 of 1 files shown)


Seldaek:

johnstevenson:
https://github.com/composer/composer/pull/6780 should also be included, mainly for the change to parsing the environment variable.

Seldaek:
Yup I'll get to that some time next week I guess :)

PR #6780, opened 2017-11-02
Report multiple inis in php.ini specific messages

johnstevenson:
Where systems use multiple ini files it is perhaps more useful to
suggest running `php --ini` to see their locations, rather than showing
the loaded php.ini (if one exists).


Seldaek:
Thanks, merged in 1.5 and master, should also be ported to the xdebug-handler repo though if it hasn't yet.

PR #6781, opened 2017-11-03
Update 03-cli.md

cafferata:
Package `fabpot/php-cs-fixer` is, so I replaced it with `friendsofphp/php-cs-fixer`.


Seldaek:
Thanks

PR #6783, opened 2017-11-04
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

gabriel-caruso:
I use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

I just need to bump PHPUnit version to [`^4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06) and [`^5.7`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.7.md#570---2016-12-02), that support this namespace.

### DIFF
--- a/tests/Composer/Test/Console/HtmlOutputFormatterTest.php
+++ b/tests/Composer/Test/Console/HtmlOutputFormatterTest.php
@@ -10,12 +10,13 @@
  * file that was distributed with this source code.
  */
 
-namespace Composer;
+namespace Composer\Test\Console;
 
 use Composer\Console\HtmlOutputFormatter;
+use PHPUnit\Framework\TestCase;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
 
-class HtmlOutputFormatterTest extends \PHPUnit_Framework_TestCase
+class HtmlOutputFormatterTest extends TestCase
 {
     public function testFormatting()
     {
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "psr/log": "^1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.5 || ^5.0.5",
+        "phpunit/phpunit": "^4.8.35 || ^5.7",
         "phpunit/phpunit-mock-objects": "^2.3 || ^3.0"
     },
     "config": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "88357fd4222a8102552a0d3345696e2e",
+    "content-hash": "d3c8dbadf8d41e2c7933e274b2fe1327",
     "packages": [
         {
             "name": "composer/ca-bundle",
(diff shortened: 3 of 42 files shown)


stof:
FYI, this PR has nothing to do with respecting PSR-1 (the rule in PSR-1 is for classes you define, not for classes you use from external libraries, as you don't have the hand on those).

anyway, migrating to the namespaced classes for PHPUnit is still a good idea to prepare the compat with PHPUnit 6.
But this requires changing the composer.json requirement. The existing one `^4.5 || ^5.0.5` covers many versions which *don't* contain the namespaced classes. You need to update it to `^4.8.35 || ^5.7`

gabriel-caruso:
@stof Oh, it has nothing to do with PSR-1? My bad, I guess I understood the API wrong so :disappointed: 

Thanks for pointing the `composer.json`. I'll bump the versions and test then :smile:

tests/Composer/Test/Console/HtmlOutputFormatterTest.php
```diff
@@ -13,9 +13,10 @@
 namespace Composer;
 
 use Composer\Console\HtmlOutputFormatter;
+use PHPUnit\Framework\TestCase as BaseTestCase;
```
gmsantos: Why do you need an alias for TestCase here?

gabriel-caruso: For some reason, without it, conflicts. I did try without but did not succeed.

stof: this is because the namespace of this test is wrong. It does not match the file structure.
And we indeed already have a `Composer\TestCase` class, which is the one conflicting here.

A better solution is to fix the test namespace, to use `Composer\Test\Console`

gabriel-caruso: @stof I've changed the `namespace` as remove the alias

gabriel-caruso:
@stof I've made some changes in this PR with your comments:

- Removed PSR-1 reference;
- Change `HtmlOutputFormatterTest` namespace;

There's one remain: PHPUnit versions. Should I update `composer.json` with [`^4.8.36`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4836---2017-06-21)? PHPUnit 5 already support these `namespaces`. Also, should I run composer to affect `composer.lock`?

MarioBlazek:
@gabriel-caruso `^4.8.35`

stof:
> PHPUnit 5 already support these namespaces.

That's wrong. *Some* PHPUnit 5 versions support them, but not all of them (it was added in 5.4 IIRC, with improvements in 5.7).
I gave the appropriate requirement in my previous comment: `^4.8.35 || ^5.7`

> Also, should I run composer to affect composer.lock?

yes, but only `composer update --lock`. The existing locked version already satisfies the updated requirement, so your PR should not contain updates of dependencies while not needed for the changes (updating the dependencies shipped in the phar is not required for this PR, and is much more risky than the current state of the PR). You just need to regenerate the lock file so that composer considers the lock file as uptodate with requirements.

Seldaek:
Looks good to me now I think, or are you still planning on updating anything?

gabriel-caruso:
@Seldaek No, I was just fix some conflicts :sweat_smile:

Seldaek:
Alright, thanks :)

PR #6785, opened 2017-11-04
Filter macOS' .DS_Store file.

### LINKED ISSUE #6784: Move files fails after package extraction
It seems that sometimes Composer moves the wrong files/directory when I run `composer install`.

I'm using
- PHP 7.0.25 CLI,
- under Debian 9.2
- in a Docker container. So it runs under root.
- Composer version 1.5.2 2017-09-11 16:59:25

It doesn't matter if I use `unzip` or the zip PHP extention.

After I run `composer install` it fails, because I want to install Symfony 3.3.9. It fails because the Symfony source is under the wrong path: `vendor/symfony/symfony/symfony-symfony-a9d2f68`

The content of `vendor/symfony/symfony/symfony-symfony-a9d2f68/` should be unter `vendor/symfony/symfony/`, without `symfony-symfony-a9d2f68/`.

This behavior appears only when `vendor/` does not exist at all. When only `vendor/symfony/symfony/` does not exist and I run `composer install` it will be extracted to `vendor/symfony/symfony/` instead to `vendor/symfony/symfony/symfony-symfony-a9d2f68/`.

I debugged the newest version of Composer using `COMPOSER_ALLOW_XDEBUG=1 php ./composer/bin/composer install -vv`. I guess this is because in some cases [this condition](https://github.com/composer/composer/blob/bfed974ae969635e622c4844e5e69526d8459baf/src/Composer/Downloader/ArchiveDownloader.php#L55) fails.

---

**Edit**: This issue is because of Mac OS X. When you have the Finder open in `vendor/` (or sub-folder), it creates a `.DS_Store` file inside of `vendor/composer/b1712560/symfony-symfony-a9d2f68/` and this let the above mentioned if-condition `if (1 === count($contentDir) && is_dir(reset($contentDir))) {` fail.

TheFox:
Closes #6784.

### DIFF
--- a/src/Composer/Downloader/ArchiveDownloader.php
+++ b/src/Composer/Downloader/ArchiveDownloader.php
@@ -148,11 +148,7 @@ abstract class ArchiveDownloader extends FileDownloader
         $finder = Finder::create()
             ->ignoreVCS(false)
             ->ignoreDotFiles(false)
-            ->filter(function ($file) {
-                /** @var \SplFileInfo $file */
-                // Filter macOS' .DS_Store file.
-                return $file->getFilename() != '.DS_Store';
-            })
+            ->notName('.DS_Store')
             ->depth(0)
             ->in($dir);
 
--- a/src/Composer/Downloader/ArchiveDownloader.php
+++ b/src/Composer/Downloader/ArchiveDownloader.php
@@ -51,6 +51,12 @@ abstract class ArchiveDownloader extends FileDownloader
 
                 $contentDir = $this->getFolderContent($temporaryDir);
 
+                // Filter macOS' .DS_Store file.
+                $contentDir = array_filter($contentDir, function ($file) {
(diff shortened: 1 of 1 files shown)


src/Composer/Downloader/ArchiveDownloader.php
```diff
@@ -148,6 +148,11 @@ private function getFolderContent($dir)
         $finder = Finder::create()
             ->ignoreVCS(false)
             ->ignoreDotFiles(false)
+            ->filter(function ($file) {
+                /** @var \SplFileInfo $file */
+                // Filter macOS' .DS_Store file.
+                return $file->getFilename() != '.DS_Store';
```
stof: would be better to use `->notName('.DS_Store')` instead, which can be applied in a more efficient way in the Finder than the callback-based filtering (because the Finder has no knowledge about what the callback does)

PR #6788, opened 2017-11-06
Symfony 4

GawainLynch:
The final run to release just started yesterday and by the look of internal discussion, we're likely to use Symfony 4 for our next (in-progress) major … and dependencies :wink: … @Seldaek I know you weren't keen on a PR to handle this until release, but given the changes — albeit not huge — I have no doubt there are going to be updates required/requested.

I guess the contentious one is the copy of ` ProcessUtils::escapeArgument()` … I might not have thought this though enough, and welcome ideas, but `ProcessExecutor::escape()` is "used everywhere", and a private method seemed the least intrusive way of providing support for SF 2.7, 3.x & 4.x.

Any of the merge button, close button, or requests for updates options welcomed :+1:


src/Composer/IO/ConsoleIO.php
```diff
@@ -281,7 +281,7 @@ public function askAndHideAnswer($question)
      */
     public function select($question, $choices, $default, $attempts = false, $errorMessage = 'Value "%s" is invalid', $multiselect = false)
     {
-        if ($this->isInteractive()) {
+        if ($this->isInteractive() && class_exists('Symfony\Component\Console\Helper\DialogHelper')) {
```
stof: This is not a good behavior. You should instead migrate this code to the non-deprecated API using the QuestionHelper and the ChoiceQuestion.

GawainLynch: Yeah, I was in two minds on this (as other bits), but it is only used internally in `ConsoleIOTest` & `NullIOTest` so went for the simple option figuring it was deprecated, just not tagged as such. Should I take it as non-deprecated?

GawainLynch: Done!

stof: well, deprecating an API means it still has to work.

But if it is used only in tests, we could indeed mark it as deprecated. But this should be a separate discussion.

src/Composer/IO/ConsoleIO.php
```diff
@@ -281,8 +282,15 @@ public function askAndHideAnswer($question)
...
     public function select($question, $choices, $default, $attempts = false, $errorMessage = 'Value "%s" is invalid', $multiselect = false)
     {
+        /** @var \Symfony\Component\Console\Helper\QuestionHelper $helper */
+        $helper = $this->helperSet->get('question');
+        $question = new ChoiceQuestion($question, $choices, $default);
+        $question->setMaxAttempts($attempts ?: null); // IOInterface requires false, and Question requires null or int
+        $question->setErrorMessage($errorMessage);
+        $question->setMultiselect($multiselect);
+
         if ($this->isInteractive()) {
```
stof: QuestionHelper is handling the interactive check already. No need to do it here anymore (other methods using the QuestionHelper stopped using it already).

DialogHelper was not doing this check internally.

GawainLynch: Thanks, mate, I really appreciate the heads-up! 

Updated.

Seldaek:
Looks good I think, thanks!

GawainLynch:
Thanks mate!

kejwmen:
@Seldaek I guess this couldn't be considered as patch for 1.5? ;) Or could we get 1.6 sooner?

Looks like repos introducing script handlers need this to be compatible with symfony 4: https://github.com/egeloen/IvoryCKEditorBundle/pull/330

https://travis-ci.org/egeloen/IvoryCKEditorBundle/jobs/310685667#L490

Seldaek:
Yeahhh.. I'll try and get 1.6 done soonish but no promises.

PR #6801, opened 2017-11-11
CS: fix style

keradus:
Brought to you by [PHP CS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer)


Seldaek:
Thanks but we don't accept CS fix PRs as I don't want to encourage people to do this everywhere. I run it myself every now and then.

keradus:
Check them on CI and you won't have to do it manually anytime.
here, there was also little config change

PR #6804, opened 2017-11-13
Reduce the level of structural statements to make it look more comfortable. - Composer\XdebugHandsler:check

medz:


src/Composer/XdebugHandler.php
```diff
@@ -63,12 +63,11 @@ public function check()
...
         if ($this->needsRestart($args[0])) {
-            if ($this->prepareRestart()) {
-                $command = $this->getCommand();
-                $this->restart($command);
+            if (! $this->prepareRestart()) {
+                return ;
             }
 
-            return;
+            $this->restart($this->getCommand());
```
stof: this does not have the return anymore at this end of this `if ($this->needsRestart($args[0])) {`, so it is changing the code

stof:
I don't see how this reduces the complexity of the method. The cyclomatic complexity is exactly the same, as you still have the same number of conditions (replacing a `if/else` by a ternary operator does not reduce complexity at all)

medz:
@stof I am phrased `phpmd` principle of correction, the visual will be better, their logic is the same.

medz:
@stof And the conditions are not established when the direct return than if the implementation of the conditions set up after the operation seems more beautiful.

Seldaek:
Not so sure whether this is an improvement, but anyway this code is being migrated to the composer/xdebug-handler repo so if you still want to do this please send the PR again there.

PR #6811, opened 2017-11-15
Deploy PHAR to GitHub releases

dunglas:
Add the PHAR file to GitHub releases. Amon other things, it will allow to use the GitHub API to download current and old versions.

Before merging this PR, a maintainer will have to add a `GITHUB_TOKEN` secret environment variable using the Travis UI. This GitHub token must have the `repo` scope.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,8 @@ before_script:
 script:
     # run test suite directories in parallel using GNU parallel
     - ls -d tests/Composer/Test/* | parallel --gnu --keep-order 'echo "Running {} tests"; ./vendor/bin/phpunit -c tests/complete.phpunit.xml --colors=always {} || (echo -e "\e[41mFAILED\e[0m {}" && exit 1);'
+
+before_deploy:
     - php -d phar.readonly=0 bin/compile
 
 deploy:
@@ -67,4 +69,4 @@ deploy:
   on:
     tags: true
     repo: composer/composer
-    php:  '5.3'
+    php:  '7.1'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -57,3 +57,14 @@ before_script:
 script:
     # run test suite directories in parallel using GNU parallel
     - ls -d tests/Composer/Test/* | parallel --gnu --keep-order 'echo "Running {} tests"; ./vendor/bin/phpunit -c tests/complete.phpunit.xml --colors=always {} || (echo -e "\e[41mFAILED\e[0m {}" && exit 1);'
+    - php -d phar.readonly=0 bin/compile
```
Seldaek: Can we only run this if php version is 7.1 for example? It'll run a lot faster there, and that way we don't build the phar for every build.. It's safe to pick 7.1 below at line 70 as well. It builds the same phar regardless of version.

stof: We could even go on step further, and run this only for builds on a tag (as only tag builds will deploy there)

Seldaek: Yup definitely, if that's easy to do that'd be preferable. travis being free doesn't mean we have to abuse their servers :)

Jean85: To do that it's sufficient to move that line into the deploy script...

stof: Composer *always* force to install PHP 5.3 deps when running `composer update`.

And Travis should not even run `update` but use the committed lock file with `install` instead, so there is even less issues.

Regarding skipping the PHAR compilation, we can do it like this:

```yaml
before_deploy:
    - php -d phar.readonly=0 bin/compile
```

The `before_deploy` step is running only when Travis determines that a deploy will be done (avoiding any need to duplicate conditions)

Seldaek:
Thanks, we'll see if all goes fine on the next release.

PR #6812, opened 2017-11-15
Add 'remove' to example global commands

saulwilcox-jisc:
Common and useful enough to be explicitly stated in docs


Seldaek:
Fair enough :)

PR #6815, opened 2017-11-16
Add a note that a user could have forgotten to add a custom repo if package not found

naderman:


src/Composer/DependencyResolver/SolverProblemsException.php
```diff
@@ -43,7 +43,7 @@ protected function createMessage()
         }
 
         if (strpos($text, 'could not be found') || strpos($text, 'no matching package found')) {
-            $text .= "\nPotential causes:\n - A typo in the package name\n - The package is not available in a stable-enough version according to your minimum-stability setting\n   see <https://getcomposer.org/doc/04-schema.md#minimum-stability> for more details.\n\nRead <https://getcomposer.org/doc/articles/troubleshooting.md> for further common problems.";
+            $text .= "\nPotential causes:\n - A typo in the package name\n - The package is not available in a stable-enough version according to your minimum-stability setting\n   see <https://getcomposer.org/doc/04-schema.md#minimum-stability> for more details.\n - It's a private package and you forgot to add a custom repository to find it\n\nRead <https://getcomposer.org/doc/articles/troubleshooting.md> for further common problems.";
```
staabm: Should the new mentioned case also link to the docs?

naderman: There is one link for all trouble shooting there in any of these cases, not sure what you mean?

staabm: As for the exact case the error mentions is described on https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md it could be worth linking to it

naderman: I think just one link there to trouble shooting is fine.

PR #6828, opened 2017-11-21
Improve findFileWithExtension PSR-4 code

LC43:
the `foreach` over the `prefixDirsPsr4[$search]` won't affect `$length`. 

It's used for the offset of `$logicalPathPsr4`.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -379,8 +379,8 @@ class ClassLoader
                 $subPath = substr($subPath, 0, $lastPos);
                 $search = $subPath.'\\';
                 if (isset($this->prefixDirsPsr4[$search])) {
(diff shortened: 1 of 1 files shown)


LC43:
reading https://github.com/composer/composer/pull/6224#issuecomment-284420562 comment about the loop (mostly) having one iteration, and confirmed locally, this is not really an optimization.

LC43:
the $subPath is not really needed, and $search can be used instead of 3 different instanciations.
```php
while (false !== $lastPos = strrpos($class, '\\')) {
    $search = substr($class, 0, $lastPos) .'\\';
```

is it worth it?

oliverklee:
Independent of whether this really improves performance, it will not decrease performance and - most importantly - it will make it more clear to the reader what is loop-specific and what isn't.

drscre:
Well, then it makes sense to move `substr($logicalPathPsr4, $length)` out the loop as well

Issue #5196, opened 2016-04-15
json do not match its signature issue 


omerbsh:
With the following composer.json:
(none)

``` json
{
    ...
}
```

When I run this command:

```
composer command -vvv (please include -vvv!)
```

I get this output:

```
  [Composer\Repository\RepositorySecurityException]                                                                                                   
  The contents of http://packagist.org/p/provider-2014%24778fe81238370a6a10514fa2191d8c49e3b0df47ad7c25361bda5e7c0f48797c.json do not match its signature. This should indicate a man-in-the-middle attack. Try running composer again and report this if you think it is a mistake.                    

```

And I expected this to happen:
install cakephp package.

curry684:
Please edit the template a bit more to supply an actual problem case.

omerbsh:
When I try to install by the command "php composer.phar create-project --prefer-dist laravel/laravel public" i get the attached error message.

I wonder if the problem is with the CRT on my computer or on the servers?

alcohol:
Are you behind a proxy or any kind of firewall software?

If the error consistently occurs, one of the above is probably mangling your connections.

omerbsh:
Hey,

no use of proxy, i changed machine and i still get the same error.

[Composer\Repository\RepositorySecurityException]                                                                                                   
  The contents of http://packagist.org/p/provider-2014%24778fe81238370a6a10514fa2191d8c49e3b0df47ad7c25361bda5e7c0f48797c.json do not match its sign  
  ature. This should indicate a man-in-the-middle attack. Try running composer again and report this if you think it is a mistake.

alcohol:
Different machine, but same network?

omerbsh:
Yes, its virtual machines on my computer (Debian).

alcohol:
Sorry, not really sure what could be wrong here. But it is definitely a networking issue, nothing we can solve for you.

Kendysond:
add this to your composer
```
"repositories": {
    "packagist": { "url": "https://packagist.org", "type": "composer" }
 }
```

smilezv:
"repositories": {
    "packagist": { "url": "https://packagist.org", "type": "composer" }
 }
This code above fix the signature issue. 

[Composer\Repository\RepositorySecurityException]
The contents of http://packagist.org/p/provider-2014%24778fe81238370a6a10514fa2191d8c49e3b0df47ad7c25361bda5e7c0f48797c.json do not match its sign
[...]

PR #6839, opened 2017-11-24
Reword exception text to evoke somewhat less panic

curry684:
Refs https://github.com/composer/composer/issues/5196#issuecomment-346754244

Signature mismatch could indicate MitM, or just a CDN issue which is rather more likely.


Seldaek:
Thanks, merged in 1.5 & master.

PR #6842, opened 2017-11-24
Flex support: New event to catch errors from the installer

### LINKED ISSUE #6821: Installing a bundle which lets the scripts fail dependency is reverted afterward
As we discussed in the #SymfonyConHackday2017 the cache:clear fails for `composer require symfony/security-bundle` for example. 

If you do that the `composer.json` file will be reverted after the failing script (unless the dependency is installed in the vendor folder) and the dependency is gone. This is quite confusing because actually, I think with Symfony flex there will be things like that.

weaverryan:
Hi guys!

This is an issue to hopefully resolve a problem we're having with Symfony Flex: symfony/flex#142. It also fixes #6821.

After requiring a new package, it's possible that one of the post-update-cmd scripts will fail... which is actually not a big deal. BUT, this causes Composer to revert the `composer.json` file.

By adding this new event, we should be able to (in Flex):

A) Hook into this event and change the return code. This will prevent the `composer.json` file from being rolled back (https://github.com/composer/composer/blob/bfed974ae969635e622c4844e5e69526d8459baf/src/Composer/Command/RequireCommand.php#L183). We would do this if the Exception is a `ScriptExecutionException` instance.

B) We should also be able to conditionally print a different message, instead of the error.

Thanks!

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -28,6 +28,7 @@ use Composer\Downloader\DownloadManager;
 use Composer\EventDispatcher\EventDispatcher;
 use Composer\Installer\InstallationManager;
 use Composer\Installer\InstallerEvents;
+use Composer\Installer\InstallerErrorEvent;
 use Composer\Installer\NoopInstaller;
 use Composer\Installer\SuggestedPackagesReporter;
 use Composer\IO\IOInterface;
@@ -171,6 +172,28 @@ class Installer
      * @return int        0 on success or a positive error code on failure
      */
     public function run()
+    {
+        try {
+            $e = null;
+            return $this->doRun();
+        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
+        }
+
+        if (null !== $e) {
+            $event = new InstallerErrorEvent(InstallerEvents::INSTALLER_ERROR, $e, $this->io);
+            $this->eventDispatcher->dispatch($event->getName(), $event);
+
+            // if the status code is set to non-zero, do not fail
+            if (0 === $event->getStatusCode()) {
+                return $event->getStatusCode();
+            }
+
+            throw $e;
+        }
+    }
+
+    private function doRun()
     {
         // Disable GC to save CPU cycles, as the dependency solver can create hundreds of thousands
         // of PHP objects, the GC can spend quite some time walking the tree of references looking
--- a/src/Composer/Installer/InstallerErrorEvent.php
+++ b/src/Composer/Installer/InstallerErrorEvent.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Installer;
+
+use Composer\EventDispatcher\Event;
+use Composer\IO\IOInterface;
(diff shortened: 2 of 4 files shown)


naderman:
The only issue I see with this is how this will interact with other plugins? I guess as long as a plugin only ever handles very specific exceptions only that plugin can throw it'll be fine. But I also see potential for this accidentally hiding errors in other plugins? Not sure if there is a way to make that less likely?

fabpot:
👍

src/Composer/Installer.php
```diff
@@ -171,6 +172,23 @@ public function __construct(IOInterface $io, Config $config, RootPackageInterfac
      * @return int        0 on success or a positive error code on failure
      */
     public function run()
+    {
+        try {
+            return $this->doRun();
+        } catch (\Exception $e) {
```
nicolas-grekas: should be also trigger the event on `Error`

weaverryan:
@nicolas-grekas I've updated the PR to catch `\Error` and `\Exception` (but obviously keeping 5.3 support).

@naderman You're correct that this is possible... but it's the nature of events that handle exceptions: the listeners will indeed need to be a bit responsible.

This is ready to go! Tests are passing. I've also already opened a PR against Flex - symfony/flex#216 - you can see how this event allows us to properly handle the failure nicely.

I realize I opened this PR really late - we should have done this weeks ago - my fault :/. But given that Symfony 4 will be released Thu, is it possible to have Composer 1.6 by then? This is a real user-gotcha with Flex.

lsmith77:
@naderman I guess plugins could also be configured to subscribe only to specific Exceptions .. potentially only Exceptions that are in the same namespace as the plugin?

lsmith77:
I guess this automatic limiting of Exception in the namespace of the plugin doesn’t make sense for @weaverryan use case ..

weaverryan:
There *is* a precedent for this type of behavior in HttpKernel with Symfony's `kernel.exception`.

An alternative would be to write the code from my plugin that uses this feature (https://github.com/symfony/flex/pull/216) into Composer. But it *is* a mild behavior change (post-*-cmd script failures would now *not* revert `composer.json`). That's why I used this approach. But I'm +1 for whichever is an easier merge for you guys :)

naderman:
Another thing to consider here is if you make throwing an exception and handling it a regular part of your plugin's control flow you end up preventing other plugins from running in these cases. With it throwing an exception that's not caught that's pretty straight forward and expected, but I'm not sure this encourages bad behaviour?

I think this is quite different from Symfony kernel as that's an application where in the end you intend to fully control all parts anyway, I'm not sure the same can be said for Composer plugins.

I think we should maybe figure out a way for plugins to only be able to handle specific exceptions that they themselves threw and in a way that other plugins then continue to run. But thinking about this, it's not clear to me why you don't just wrap whatever post install command is being executed that should be allowed to error in a try/catch yourself?

Either way we're not going to release a 1.6 so quickly, but after some discussion @Seldaek did commit https://github.com/composer/composer/commit/b4df2c95173f08d887ad0c4066afbb3ca17a33f7 earlier as a workaround for now.

Seldaek:
FWIW, the commit @naderman mentioned is available in the latest snapshot so you can try to make sure it works and fixes the revert on require when cache clear fails. It should still print the exception but without reverting. I'll try and tag a patch release tomorrow so it makes it to stable too.

In the meantime closing this because it's not really a feature we want to support.

weaverryan:
@Seldaek I was already doing it! And yes, it works! A patch release would be amazing. Thanks guys - I really didn't care how it was solved, I appreciate your commit.

Issue #6845, opened 2017-11-27
PackageInterface::getPrettyVersion() returns same value as getVersion() when in version-branch


Jean85:
This was originally reported in https://github.com/Jean85/pretty-package-versions/issues/4.

Basically, in https://github.com/Ocramius/PackageVersions/pull/51 (let's ping @Ocramius) we started using `PackageInterface::getPrettyVersion()` to get the pretty version of the root package, without normalization, as for the dependecies (that we get from the lock).
This doesn't seem to work when the root package is in a version-branch, like `3.0`, because `getPrettyVersion()` and `getVersion()` are returning the same value; but `getPrettyVersion()` should return the non-normalized version instead!

The issue can be reproduced like this:
```bash
git init
cat > composer.json <<EOF
{
    "name": "test/test",
    "require": {
        "ocramius/package-versions": "dev-master#ad8a245decad4897cc6b432743913dad0d69753c as 1.1.3"
    }
}
EOF

cat > main.php  <<EOF
<?php

require __DIR__ . '/vendor/autoload.php';
echo \PackageVersions\Versions::getVersion('test/test'), PHP_EOL;
EOF

git checkout -b 3.0
git add .
git commit -mTest
composer install
php main.php

# 3.0.9999999.9999999-dev@412b22b
# instead of
# 3.0-dev@412b22b
```

Ocramius:
@Jean85 can you please report an issue also on https://github.com/Ocramius/PackageVersions? This is a release blocker on my side.

Jean85:
I've track down the issue to this piece of code: https://github.com/composer/composer/blob/bfed974ae969635e622c4844e5e69526d8459baf/src/Composer/Package/Loader/RootPackageLoader.php#L70-L101

It loads the root package using `$versionData['version']`; but the `VersionGuesser` uses `version` and `pretty_version` as array keys, and instead the `RootPackageLoader` parent constructor does this:

https://github.com/composer/composer/blob/bfed974ae969635e622c4844e5e69526d8459baf/src/Composer/Package/Loader/ArrayLoader.php#L56

IMHO it should at least use the `pretty_version` piece and fall back if it's not there.

Jean85:
I'm sorry @Seldaek but yuor fix doesn't seem to work.. I've opened #6866 to try and fix it.

PR #6847, opened 2017-11-27
Fix root package pretty version

Jean85:
This solves issue #6845.

`RootPackageLoader` was not passing `pretty_version` from the `VersionGuesser` to its parent constructor; PR is conveniently splitted into 2 separated commits, the second one is just a small refactoring to avoid unuseful intermediate variables.

### DIFF
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -47,18 +47,13 @@ class ArrayLoader implements LoaderInterface
             throw new \UnexpectedValueException('Package '.$config['name'].' has no version defined.');
         }
 
-        $prettyVersion = $config['version'];
-        if (isset($config['pretty_version'])) {
-            $prettyVersion = $config['pretty_version'];
-        }
-
         // handle already normalized versions
         if (isset($config['version_normalized'])) {
             $version = $config['version_normalized'];
         } else {
             $version = $this->versionParser->normalize($config['version']);
         }
-        $package = new $class($config['name'], $version, $prettyVersion);
+        $package = new $class($config['name'], $version, $config['version']);
         $package->setType(isset($config['type']) ? strtolower($config['type']) : 'library');
 
         if (isset($config['target-dir'])) {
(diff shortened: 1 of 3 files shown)


tests/Composer/Test/Package/Loader/RootPackageLoaderTest.php
```diff
@@ -91,6 +92,26 @@ public function testNoVersionIsVisibleInPrettyVersion()
         $this->assertEquals("No version set (parsed as 1.0.0)", $package->getPrettyVersion());
     }
 
+    public function testPrettyVersionForRootPackageInVersionBranch()
+    {
+        // see #6845
+        $manager = $this->prophesize('\\Composer\\Repository\\RepositoryManager');
```
alcohol: Why double backslashes here but not on the next line?

Jean85: Don't know, I think I copy pasted it from a previous post! It seem to work anyway. I'll uniform it.

src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -47,13 +47,18 @@ public function load(array $config, $class = 'Composer\Package\CompletePackage')
             throw new \UnexpectedValueException('Package '.$config['name'].' has no version defined.');
         }
 
+        $prettyVersion = $config['version'];
+        if (isset($config['pretty_version'])) {
```
stof: composer.json is not meant to contain a `pretty_version` field, so this code looks suspicious in the ArrayLoader itself

Jean85: Pretty, in this context stands as "not normalized", as indicated in `PackageInterface::getPrettyVersion()`. The name of the array key comes also from the version guesser.

tests/Composer/Test/Package/Loader/RootPackageLoaderTest.php
alcohol: Thanks :-)

Seldaek:
Thanks, I simplified the patch a little to avoid this pretty_version addition, seems to work well still :)

Jean85:
Thank you for the fast merge!

PR #6853, opened 2017-11-28
Update scripts.md

alegout:
Custom descriptions only available in nightly build

### DIFF
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -274,6 +274,8 @@ JSON array of commands.
 
 ## Custom descriptions.
 
+> Only available in the nightly build for the moment.
+
 You can set custom script descriptions with the following extra in your composer.json:
 
  ```json


stof:
Well, this is the documentation in the master branch of Composer.

@Seldaek given that there is quite some time between the release of each minor version, I think it would be important for the online documentation to have the doc for the stable channel (or at least for the 1.5 branch to get doc changes deployed fast) rather than for master

alegout:
You're right 👍

Seldaek:
Thanks but yeah as mentioned the docs are always a bit optimistic. I like it that way as you can always use it via `composer self-update --snapshot` which I'd argue most people should do on their dev boxes so that we get early feedback before releases if anything breaks.

stof:
@Seldaek maybe we should display a banner on top of the documentation pages explaining that they describe features available in snapshot, and stable releases may not have them all ?

Seldaek:
I think this adds more confusion than the odd case where something is new and not released. It's not like we add that many major features..

alegout:
It's frustrating when you try to make it work... lost of time 😕

frederickjh:
Has **scripts-descriptions** made it to the current composer release of v1.8.2, as it still seems to not be working in the help and `composer validate` throws the error (red and white): "**The property scripts-descriptions is not defined and the definition does not allow additional properties**" It has been over a year since this was closed.

Issue #6685, opened 2017-09-13
Experiment: External Dependency Solver


lorenzleutgeb:
Hi!

I just ran a quick experiment and would like to get your opinion.

We're seeing quite long dependency resolution times with Composer. As dependency resolution is hard, that's not something I want to complain about. But I wanted to see how Composer's solver compares to other SAT solvers. So what I did was [a quick hack](https://github.com/composer/composer/compare/master...lorenzleutgeb:extsat) so that Composer's solver writes the formula it is solving to a file in [DIMACS format](https://logic.pdmi.ras.ru/~basolver/dimacs.html).

With this I could get a feel for the size of Composer SAT formula:

```bash
$ head -n1 composer.cnf
p cnf 11253 390122
$ ls -sh composer.cnf 
6.5M composer.cnf
```

It took Composer around 278.56 seconds to find a satisfying assignment for that formula. Now I could try and compare it against some other solver. [Plingeling](http://fmv.jku.at/lingeling/) happily solves the formula within just 0.44 seconds. That's a factor of 633.

I also tried unsatisfiable requirements, and sure enough, Plingeling reported unsatisfiability.

So it seems that there is quite some gap between the performance of Composer and Plingeling. This is what sparks my interest. Transforming a satisfying assignment back into Composer's preferred format should not be that big of a deal. What is tricky when working with an external solver is resolution of conflicts. It is not straight forward to export the state of an external solver, import it into Composer and then resolve conflicts based on that ... I'm not saying it's not possible, but I just don't know the nitty gritty details of how Composers solver does that internally.

So my proposal is as follows:
 * Only use an external solver when the size of the SAT formula (in terms of literals and clauses) exceeds some threshold. For easy formulas it's not worth the fuss.
 * Allow Composer to detect external SAT solvers (some binary from a whitelist in `$PATH`). If there is one, start it.
 * Now the external solver could either run in parallel with the Composer solver, or before the internal one. The problem is that the satisfiability of the requirements is unclear (my guess is that requirements are much more often satisfiable than unsatisfiable), so there should be some simple heuristic either running the solvers in parallel or opting for the external one first, falling back to internal in case of conflicts.
   * Assume that the requirements are satisfiable. Then it would probably make sense to just run the external solver and read the solution.
   * Assume that the requirements are not satisfiable. Then dependency resolution must be performend by the Composer solver, so starting an external one is pretty much useless.

[...]

PR #6864, opened 2017-12-03
Proposed event on SolverProblemsException with power to trigger re-run.

mbaynton:
So this is definitely not ready to actually merge yet, but @naderman suggested discussing PRs early, so here goes. Oh, in case there would otherwise be confusion, this is unrelated to my work on `replace`.

The problem this helps solve is enabling Drupal installations/updates for nontechie users without exceeding the resource limitations of a webserver environment; it's a huge pain point for Drupal right now that it isn't possible to do Composer-enabled installs/updates in a browser. The way it helps solve it is by allowing a plugin to initially restrict the `Pool` to a repository containing only one version of each required package, all of which are pre-selected to be compatible with each other; this should make the sat problem and number of modeled packages/decisions trivial/very fast in execution time/low in memory. For sites that only need things available from this special repository, a complete Composer run ought to work fine within the confines of a webserver environment. When the site does also have requirements from the wider world, the first solver run will fail, but it will fail fast. This PR would allow a plugin to then introduce packagist etc to the Pool and rerun the Solver. (I believe @thom8 has talked with some Composer maintainers about the possibility of running the solver on a remote "solver as a service" host; I've got a vague idea of also swapping out the standard solver for the remote one long-term when a full-scale dependency resolution is required.)

Something like this have a chance of going in?

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -467,11 +467,18 @@ class Installer
         $this->processDevPackages($localRepo, $pool, $policy, $repositories, $installedRepo, $lockedRepository, 'force-links');
 
         // solve dependencies
-        $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
-        $solver = new Solver($policy, $pool, $installedRepo, $this->io);
-        try {
-            $operations = $solver->solve($request, $this->ignorePlatformReqs);
-        } catch (SolverProblemsException $e) {
+        for ($runSolver = true, $solverRuns = 0; $runSolver && $solverRuns < 5; $solverRuns++) {
+            $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $pool, $installedRepo, $request);
+            $solver = new Solver($policy, $pool, $installedRepo, $this->io);
+            try {
+                $operations = $solver->solve($request, $this->ignorePlatformReqs);
+                $e = null;
+                $runSolver = false;
+            } catch (SolverProblemsException $e) {
+                $runSolver = $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::DEPENDENCIES_SOLVING_PROBLEM, $this->devMode, $policy, $pool, $installedRepo, $request);
+            }
+        }
+        if ($e) {
             $this->io->writeError('<error>Your requirements could not be resolved to an installable set of packages.</error>', true, IOInterface::QUIET);
             $this->io->writeError($e->getMessage());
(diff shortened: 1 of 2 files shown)


naderman:
We did indeed talk about this issue. My proposal was to give the Solver a clearer replaceable interface in Composer v2 which would allow easy replacement with an external service, or other solvers as is being worked on in https://github.com/composer/composer/issues/6685

The issue I see with adding this currently as-is is that we'll be stuck having to support this even in Composer v2 and future versions then. Any bug in this potential plugin would lead to rather weird hard to reproduce behaviour which I fear will complicate bug reports for us a lot. How/where did you intend to restrict/modify the pool in your plugin?

mbaynton:
@naderman my rough outline of a plan for how to proceed with modifying the Pool via a plugin was to target the `RuleSetGenerator` class for extension and injection into the Solver. This class has the access and opportunity to tinker with the `Pool` right before solving starts (through `RuleSetGenerator::getRulesFor()`) and I also would have been able to accomplish a similar effect to PR#6843 in an override of `RuleSetGenerator::whitelistFromPackage()`.

I have some measure of reluctance to approach my objectives by participating in their proper development into Composer v2 as opposed to hacking them into Composer v1 plugins, but only because my objective is ostensibly to solve Drupal problems and my approach involving the maintenance of a repository paralleling hundreds of packages from packages.drupal.org and packagist is already more work than one person can probably do. On the other hand, I appreciate your position that adding functionality in a hacky way that must then be retained in v2 is nonsensical for the Composer project, and I think it's highly likely other non-solver related PRs would be necessary as well to fully incorporate Composer-powered installs and updates into an in-browser UI -- this is basically Drupal's ultimate need.

Your response to both my PRs so far suggests more openness to tackling these changes needed by an improved Drupal updater as part of Composer v2 refactoring...how far does that go? In other words, does the Composer project see Drupal as a sufficiently compelling use-case to substantially inform the direction of Composer v2 development if I / others from Drupal did focus in on working on Composer v2?

Seldaek:
@mbaynton what's your status with this? Any updates? Because work on 2.0 on our side is maturing so we'd need to see how important/urgent making the Solver replaceable is.

mbaynton:
Thanks for the heads up @Seldaek. I've started kicking the tires on Composer 2.0 + Drupal 8.8, which is the first Drupal release to have official project templates for `create-project`ing new websites.  Let's start with https://github.com/composer/installers/pull/448 perhaps; Drupal's one of the many projects that needs `composer/installers` to work and not generate a solver error on `composer-plugin-api` version.

Seldaek:
It's still a bit early for marking plugins compatible IMO but we're getting there slowly.

Seldaek:
I think/hope with composer 2 this will not be needed anymore. Especially if we manage to improve things further in 2.1 with https://github.com/composer/composer/pull/9261 - so closing this.

PR #6849, opened 2017-11-27
Refactoring Tests

gabriel-caruso:
I've refactored some tests with [`assertFileExists`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFileExists) and `assertFileNotExists`, [`assertFalse`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFalse), [`assertTrue`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertTrue), [`assertCount`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertCount)  and [`assertEmpty`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEmpty) methods.

PR #6865, opened 2017-12-03
Refactoring tests

carusogabriel:
Continuing to refactored tests, like #6849, using:
- `assertNotEmpty` instead of `empty` method;
- `assertFalse` instead of strict comparisons with `false` keyword;
- `assertSame` instead of strict comparisons;
- `assertContains` instead of `in_array` method;
- `assertStringEqualsFile` when comparing `string` and files.


alcohol:
Thanks, though some changes might have to be re-applied in [`composer/xdebug-handler`](/composer/xdebug-handler).

carusogabriel:
@alcohol Gonna work on that!

PR #6866, opened 2017-12-04
Fix VersionGuesser pretty_version handling

Jean85:
This relates to #6845, since the merged fix doesn't seem to work, see https://github.com/phpstan/phpstan/pull/649

I'm still working on this, for now I've added a regression test that expects a non-normalized version as the `pretty_version` returned by the `VersionGuesser`.

### DIFF
--- a/src/Composer/Package/Version/VersionGuesser.php
+++ b/src/Composer/Package/Version/VersionGuesser.php
@@ -88,8 +88,7 @@ class VersionGuesser
 
     private function postprocess(array $versionData)
     {
-        // make sure that e.g. dev-1.5 gets converted to 1.5.x-dev
-        if ('dev-' !== substr($versionData['version'], 0, 4)) {
+        if ('-dev' === substr($versionData['version'], -4)) {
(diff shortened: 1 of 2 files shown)


Jean85:
I'm not privy to the whole version guessing, but I based the fix (287e0e1) on the tests... Now it should be good.

Seldaek:
Well.. I guess I fixed it for branches, and I am fairly confident you broke that to fix it for tags :) I think both replacements should be in really.

Jean85:
@Seldaek are you sure? Tests are green and the `testNumericBranchesShowNicely` test case seems to cover the branch case:

https://github.com/Jean85/composer/blob/287e0e1c22bc2749f3ea422ce1b0d2a01f390fda/tests/Composer/Test/Package/Version/VersionGuesserTest.php#L486-L487

If I'm missing a case, can you point it to me, so I can add a regression test?

Seldaek:
Right sorry I misread the patch.

Merged in 1.5 and master!

Jean85:
Thanks! ;)

PR #6872, opened 2017-12-05
Clarified language on branch naming for forks

danepowell:
The current language here implies that the branch name on your remote repo should be named according to a certain pattern (`dev-*`), which isn't accurate. You can name the branch whatever you want, but in composer.json you should _reference_ it as `dev-*`.


Seldaek:
Thanks

PR #6875, opened 2017-12-06
output used strategy

OskarStark:
I was a little bit confused, that the output is the same, regarding `-o` or `-a` option.


alcohol:
While I agree more clarity could be beneficial to users here, the currently chosing phrasing seems a bit odd to me. I do not really see it as a strategy. We generate a classmap in either case, it is just that in one case, we consider the classmap to be completely authoritative.

OskarStark:
So what is your opinion @alcohol ? Closing this PR? change the output?

alcohol:
No need to close it, I would personally just prefer different phrasing. But this is merely my opinion.

My suggestion would be to just use the following two messages:

```
```
```
Generating optimized autoload files (authoritative)
```

OskarStark:
> My suggestion would be to just use the following two messages:

sounds good 👍

Seldaek:
Thanks

Issue #6894, opened 2017-12-12
When using a private project on GitLab, using HTTP, apiUrl gets null


renoirb:
When having private projects, hosted on your own GitLab installation instance.
And when GitLab install is **only** over HTTPs on a different port number.
And when GitLab install's SSH access (e.g. `git@gitlab.mycompany.com...`) **IS NOT Working**

When we try to pull dependencies using a GitLab repository, it fails because the `Composer\Repository\Vcs\GitLabDriver::URL_REGEX` does not take a port number into account.

I'd like to see this use-case supported and I opened #6895 to help about it.

![An actual snapshot of a run from the project](https://user-images.githubusercontent.com/296940/33905426-9c2ae3ea-df4c-11e7-9068-e91abc1a214b.PNG)

**NOTICE** that the installation API at `ssh_url_to_repo` looks like this:

```
["ssh_url_to_repo"]=>
      string(73) "ssh://git@gitlab.mycompany.com:50043:20022/mygroup/myproject.git"
```

My `composer.json`:

```json
{
    "repositories": [
      {
            "type": "gitlab",
            "url": "https://gitlab.mycompany.com:50043/mygroup/myproject.git"
        }
    ],
    "config": {
        "gitlab-domains": ["gitlab.mycompany.com"]
    }
}
```

Output of `composer diagnose`:

```
composer diagnose

Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: WARNING
The zlib extension is not loaded, this can slow down Composer a lot.
If possible, enable it or recompile php with --with-zlib

Your command-line PHP is using multiple ini files. Run `php --ini` to show them.
If you can not modify the ini file, you can also run `php -d option=value` to modify ini values on the fly. You can use -d multiple times.

Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: FAIL
Missing pubkey for tags verification
Missing pubkey for dev verification
Run composer self-update --update-keys to set them up
Checking composer version: OK



composer -V

Composer version 1.5.5 2017-12-01 14:42:57
```

When I run this command:

```
// e.g. CI_PRIVATE_TOKEN="foobar"
composer config gitlab-token.gitlab.mycompany.com ${CI_PRIVATE_TOKEN}
composer update -v
```

I get the following output:

```
[...]

PR #6895, opened 2017-12-12
Support GitLab repositories w/ port over HTTPs

renoirb:
When we have self-hosted GitLab, over HTTPs, on a different port.

See composer/composer#6894

To test, use the `@group gitlabHttpPort` tag

```console
vendor/bin/phpunit --group gitlabHttpPort
```


Seldaek:
Merged in 1.5 and master, thanks

PR #6896, opened 2017-12-12
Add PHPUnit 6 support

carusogabriel:
I've added `PHPUnit 6` support to `Composer`. Changes:
- Define mocks with fluent interface;
- `setExpectedException` isn't available in version `6`, so I substitute to `@expectedException` and `@expectedExceptionMessage` notations. But, there are some tests that these expectations are variables, so, I created a method layer in `Composer\TestCase` and extend those TestCases that need this method.

One question: why not extend all TestCases from `Composer\TestCase`? I guess is easier for maintenance, isn't?

### DIFF
--- a/tests/Composer/Test/Command/InitCommandTest.php
+++ b/tests/Composer/Test/Command/InitCommandTest.php
@@ -77,17 +77,21 @@ class InitCommandTest extends TestCase
         $this->assertEquals('john@example.com', $author['email']);
     }
 
+    /**
+     * @expectedException InvalidArgumentException
+     */
     public function testParseEmptyAuthorString()
     {
         $command = new InitCommand;
-        $this->setExpectedException('InvalidArgumentException');
         $command->parseAuthorString('');
     }
 
+    /**
+     * @expectedException InvalidArgumentException
+     */
     public function testParseAuthorStringWithInvalidEmail()
     {
         $command = new InitCommand;
-        $this->setExpectedException('InvalidArgumentException');
         $command->parseAuthorString('John Smith <john>');
     }
 }
--- a/tests/Composer/TestCase.php
+++ b/tests/Composer/TestCase.php
@@ -105,4 +105,28 @@ abstract class TestCase extends BaseTestCase
             $this->markTestSkipped($executableName . ' is not found or not executable.');
         }
     }
+
+    /**
+     * Method to keep PHPUnit 4.x compatibility.
+     *
+     * @param mixed  $exceptionName
+     * @param string $exceptionMessage
+     * @param int    $exceptionCode
+     */
+    public function setExpectedException($exception, $exceptionMessage = '', $exceptionCode = null)
+    {
+        if (method_exists($this, 'expectException')) {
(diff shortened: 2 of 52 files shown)


tests/Composer/Test/Command/InitCommandTest.php
```diff
@@ -77,17 +77,21 @@ public function testParseValidAlias2AuthorString()
         $this->assertEquals('john@example.com', $author['email']);
     }
 
+    /**
+     * @expectedException InvalidArgumentException
```
hkdobrev: Why not use `$this->expectException()` and `$this->expectExceptionMessage()` methods which I think are quite better API than the annotations?

carusogabriel: Method or annotations, that's the question :sweat_smile: 
Maybe we should ask @Seldaek about it.

stof: `expectException` does not exist in PHPUnit 4.8, which must remain supported to be able to test Composer on PHP 5.3 to 5.5 (PHPUnit 5.x requires PHP 5.6+).
So using the methods would require a forward compatibility layer.

carusogabriel: Actually, these methods aren’t avaiable in `PHPUnit 4.8`, so annotations is the better way to keep compatibility!

tests/Composer/TestCase.php
```diff
@@ -105,4 +105,28 @@ protected function skipIfNotExecutable($executableName)
             $this->markTestSkipped($executableName . ' is not found or not executable.');
         }
     }
+
+    /**
+     * Method to keep PHPUnit 4.x compatibility.
```
stof: I would rather build a forward compatibility layer (allowing to use the new API all the time) than a backward compatibility layer, if possible.

tests/Composer/Test/Command/InitCommandTest.php
stof: The advantage of using a method is that they define that the exception is expected only after the method got called (so any exception happening earlier in the test initialization would be reported as an error). The annotation is equivalent to calling the method at the very beginning of the test.

tests/Composer/TestCase.php
carusogabriel: What do you have in mind? I couldn't figure out a solution since there's no way to set an `expectedExceptionMessage` in `PHPUnit 4` without calling `setExpectedException`.

Seldaek:
I think this can wait until 2.0 or 3.0, whenever we will drop older PHP versions. We can keep using older PHPUnit until then as far as I can tell this causes no issue. We can then skip the whole compatibility layer once we drop older PHP versions altogether.

carusogabriel:
So, can I simplify this `PR` in some specific point, or no, let’s just wait? 😁

Seldaek:
I think let's rather wait because any work done now might anyway end up 
conflicting or not be up to date with PHPUnit at the point where we 
actually migrate.

PR #6903, opened 2017-12-14
Clean elses

carusogabriel:
Small one: cleaned up the `else`s when we have already returned something :put_litter_in_its_place:


curry684:
Doesn't `return ($this->watch1 == $literal) ? $this->watch2 : $this->watch1;` compile even more efficiently?

carusogabriel:
Didn't think on that, gonna search :thinking:

curry684:
From what I can find it even used to be slower in many cases due to an issue in the PHP engine (copy-on-write issues). While this has been fixed in the PHP7 engine they are now technically identical, so readability is preferred. No change required.

PR #6904, opened 2017-12-14
clarify optionality of package in require command

brettsantore:
This is a behavior I was unaware of and I didn't feel it was clearly identified in the docs


Seldaek:
Thanks

PR #6906, opened 2017-12-15
Corrects troubleshooting.md documentation link by removing space

sonjaq:
There was an extra space lurking that prevented a link from rendering. The space is now removed.


Seldaek:
Thanks

PR #6911, opened 2017-12-17
Fix shell proxy to allow user-defined cygdrive

johnstevenson:
Recently fixed in Composer-Setup: https://github.com/composer/windows-setup/blob/master/CHANGELOG.md#v480

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -194,21 +194,14 @@ class BinaryInstaller
         $proxyCode = <<<PROXY
 #!/usr/bin/env sh
 
-dir=$(d=\${0%[/\\\\]*}; cd "\$d" > /dev/null; cd $binDir && pwd)
+dir=\$(cd "\${0%[/\\\\]*}" > /dev/null; cd $binDir && pwd)
 
-# See if we are running in Cygwin by checking for cygpath program
-if command -v 'cygpath' >/dev/null 2>&1; then
-	# Cygwin paths start with /cygdrive/ which will break windows PHP,
-	# so we need to translate the dir path to windows format. However
-	# we could be using cygwin PHP which does not require this, so we
-	# test if the path to PHP starts with /cygdrive/ rather than /usr/bin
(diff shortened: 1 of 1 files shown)


src/Composer/Installer/BinaryInstaller.php
```diff
@@ -194,21 +194,14 @@ protected function generateUnixyProxyCode($bin, $link)
...
-		dir=$(cygpath -m "\$dir");
-	fi
+dir=\$(cd "\${0%[/\\\\]*}" > /dev/null; cd $binDir && pwd)
+
+if [ -d /proc/cygdrive ] && [[ \$(which php) == \$(readlink -n /proc/cygdrive)/* ]]; then
+   # We are in Cgywin using Windows php, so the path must be translated
+   dir=\$(cygpath -m "\$dir");
 fi
 
-dir=$(echo \$dir | sed 's/ /\ /g')
```
Seldaek: I'm kinda worried about this removal here.. are you sure this doesn't have any ill effect if $dir has spaces in it? I see it's quoted below so should be fine but I wonder why we had this in then..

johnstevenson: We had it because the code was copied from Composer-Setup, which itself was based on a script from an old Git for Windows version, which later removed it (and now doesn't use this script at all because it handles things in `.exe` wrappers). It was always quoted.

I finally got brave enough to remove it from Composer-Setup in March 2016 (https://github.com/composer/windows-setup/commit/cb217b96c82ae90e9e7c50e4f8b1d7c345dccfd9#diff-1e307925c805eb0d5d524703adb85788) even though I knew it _should_ work!

It originally looked like this:
```
PHAR="$(echo $DIRECTORY | sed 's/ /\ /g')/composer.phar"
php "${PHAR}" $*
```
then was improved to this:
```
dir=$(echo $dir | sed 's/ /\ /g')
php "${dir}/composer.phar"
```

Seldaek:
Ok then I guess let's give it a shot.. Will merge in 1.6 only though for safety :)

johnstevenson:
Ha ha, you're as brave as me!

Issue #6892, opened 2017-12-11
Require command should add dependency with letter case from it's name in composer.json


kubawerlos:
Having empty `composer.json` whether I run `composer require squizlabs/PHP_CodeSniffer` or composer `composer require squizlabs/PhP_CoDeSNiFFeR` I should end up with:
```json
{
    "require": {
        "squizlabs/php_codesniffer": "^3.1"
    }
}
```
as `squizlabs/php_codesniffer` is the [name of the package](https://github.com/squizlabs/PHP_CodeSniffer/blob/3.1.1/composer.json#L2).

Seldaek:
Fixed by #6912

PR #6912, opened 2017-12-18
Add feature for issue #6892

swumao:
Add feature for issue #6892

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -711,10 +711,10 @@ EOT
             ));
         }
 
-        return [
+        return array(
             $package->getPrettyName(),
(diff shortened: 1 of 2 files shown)


Seldaek:
This is not really fixing it to use the actual case from the package though, it just lowercases it. Packagist.org enforces lowercased names but if you access/publish the package differently it can very well have a different name.

swumao:
Should I replace the package name with the Packagist.org's package name ?

Seldaek:
Yes it should use the name from the Package instance (getPrettyName) we get back but I am not sure how easy it is to do that to be honest.

Seldaek:
Great thanks for the update

kubawerlos:
Thanks @swumao

PR #6913, opened 2017-12-18
Move "scripts-description" to toplevel

### LINKED ISSUE #6881: Move "scripts-description" to toplevel
With #6672 a new config option `scripts-description` was added to provide description for custom commands.

Since this was implemented as `extra` config option every change here requires a update of the `composer.lock` file. This is not really useful since changes to script descriptions don't affect package states or similar at all.

Thus I propose to move `scripts-description` toplevel which also allows to put them next to `scripts`.

mbrodala:
Fixes #6881

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -379,7 +379,7 @@
         },
         "scripts": {
             "type": ["object"],
-            "description": "Scripts listeners that will be executed before/after some events.",
+            "description": "Script listeners that will be executed before/after some events.",
             "properties": {
                 "pre-install-cmd": {
                     "type": ["array", "string"],
@@ -449,7 +449,7 @@
         },
         "scripts-descriptions": {
             "type": ["object"],
-            "description": "Descriptions for scripts listeners, shown in console help.",
+            "description": "Descriptions for custom commands, shown in console help.",
             "additionalProperties": {
                 "type": "string"
             }
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -447,6 +447,10 @@
                 }
             }
         },
+        "scripts-description": {
+            "type": ["object"],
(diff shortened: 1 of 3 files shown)


res/composer-schema.json
```diff
@@ -447,6 +447,10 @@
                 }
             }
         },
+        "scripts-description": {
+            "type": ["object"],
```
stof: would be great to describe it further, by using `additionalProperties` to mark items in it as strings.

res/composer-schema.json
```diff
@@ -447,6 +447,10 @@
                 }
             }
         },
+        "scripts-description": {
```
stof: naming it `script-descriptions` would be more correct, as the plural goes to the last noun in English.

mbrodala: I agree but the base option is called `scripts` so it makes sense to name it after that to keep them closely aligned.

res/composer-schema.json
mbrodala: I'll take care of this.

res/composer-schema.json
```diff
@@ -447,6 +447,13 @@
                 }
             }
         },
+        "scripts-descriptions": {
+            "type": ["object"],
+            "description": "Descriptions for scripts listeners, shown in console help.",
```
stof: should be `for script listeners` to be valid English (noun used as qualifiers are always singular)

res/composer-schema.json
stof: well, I would still vote for the valid plural form (`scripts` is the valid plural form for `script` too btw)

res/composer-schema.json
mbrodala: Fixed for `scripts` now, changed completely for `scripts-descriptions`.

Seldaek:
Thanks

PR #6919, opened 2017-12-19
Combine consecutive issets

carusogabriel:
`isset` allow us to pass multiple values to check.

Also enabled in `.php_cs`.


Seldaek:
Thanks but I rather not have this, I find it very confusing in situations where it's negated for example, `if (!isset($foo, $bar, $baz))`. A lot of people don't know what this means, and I doubt the performance difference is worth the readability troubles.

PR #6920, opened 2017-12-19
Remove build platform from appveyor.yml

johnstevenson:
The platform config value is used for MSBuild, rather than providing an OS platform (which is always x64)


johnstevenson:
The Appveyor cache is 1GB, so plenty of space for PHP, composer.phar and the Composer cache directory.

Seldaek:
Not sure how to enable the cache? From what I read it seems to be enabled by default and you just need to configure which dirs are cached?

johnstevenson:
I was only reading their docs (linked above). I think it is enabled for PRs, but it is hard to tell because I keep invalidating it by changing `appveyor.yml` with these PRS!

> Note: By default, saving cache is disabled in Pull Request builds. Use Save build cache in Pull Requests checkbox on the General tab of project settings if you need to save cache during PR builds. This setting is not exposed in YAML to prevent unauthorized cache modifications.

Seldaek:
Yup I saw that checkbox, but I figured it's best to leave it off so that PRs don't update the cache which will reduce the chances it becomes full? Anyway I'll merge this then if it helps once master built?

Seldaek:
It actually doesn't seem to work.. I built master twice and the second is still downloading everything. I'm not sure what PHP_CACHE_DIR is supposed to be but maybe I can add this to the cache in the appveyor settings?

Seldaek:
Or am I supposed to define that env var myself? :)

johnstevenson:
Not all all. PHP_CACHE_DIR is where php and composer.phar and shims end up, plus the Composer cache directory.

Seldaek:
Oh it does say this at the end of the build:

```
Updating build cache...
Cache 'C:\tools\php' - Updated
Build success
```

Trying a new build..

johnstevenson:
Seems to have kicked in now on https://ci.appveyor.com/project/Seldaek/composer/build/2443

johnstevenson:
It took me a few builds too get it working on one of my repos.

Seldaek:
Yup it worked that way, cool. It shaves off ~20seconds of build time. That just means we're stuck on PHP 7.2.0 forever though, not sure how bad that is.

johnstevenson:
When chocolatey update to the next release, we just update `PHP_CHOCO_VERSION`. Not ideal but probably the safest way at the moment.

Seldaek:
Ah I see the cache depends on the yaml file hash? I didn't realize that.

johnstevenson:
There are other ways to do this. We could download a zipped PHP release, but it would still be a manual update process. What PHP versions would you like to test on?

johnstevenson:
I think I misunderstood the disabling for PRs, because the cache is still used.

Seldaek:
Oh no it's fine to be only testing on one version on windows IMO, I just worried that we'd never be able to update it without deleting the cache by hand, but if we can update the version number and it invalidates the cache it's fine.

As for the PR cache thing, it just means that PRs do not *update* the cache, but they still use it. That's good I think.

johnstevenson:
Yup, that seems to be what happens and that's good.

For clarity, the env vars `PHP_CHOCO_VERSION`, `PHP_CACHE_DIR` and `PHP` are local to the script.

PR #6921, opened 2017-12-19
Add no_useless_else in phpcs

carusogabriel:
Should we add `no_useless_else` based on #6903?


Seldaek:
Sure, that's generally a good thing IMO, thanks.

PR #6926, opened 2017-12-20
Fix the platform requirement check command

### LINKED ISSUE #6925: check-platform-reqs does not work fine
My `composer.json`:

```json
{
    "require": {
        "php": "7.1.*",
        "ext-amqp": "*",
        "ext-apcu": "*",
        "ext-ftp": "*",
        "ext-intl": "*",
        "ext-mbstring": "*",
        "ext-openssl": "*",
        "ext-pcntl": "*"
    }
}
```

Plus a lot of other non-platform requirements of course

When I run this command:

```
php7.2 $(which composer) check-platform-reqs
```

I get the following output:

```
php            7.2.0       success
```

Telling me that my PHP version is wrong (as PHP 7.2.0 does not match the `7.1.*` constraint)

stof:
The command must validate that the current package matches the constraint, not that any newer package matches it.

Closes #6925

### DIFF
--- a/src/Composer/Command/CheckPlatformReqsCommand.php
+++ b/src/Composer/Command/CheckPlatformReqsCommand.php
@@ -82,7 +82,7 @@ EOT
                     $version = $currentPlatformPackageMap[$require]->getVersion();
 
                     foreach ($links as $link) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Oh yeah nice catch thanks

PR #6942, opened 2017-12-27
Fix: Remove unused parameter and field

localheinz:
This PR

* [x] removes an unused parameter and field from `Locker`

💁‍♂️ The instance of `RepositoryManager` passed in into its constructor is never used.


nicolas-grekas:
That's a BC break for plugins.

Seldaek:
Yup that's why 2.0 ;)

localheinz:
Thank you, @Seldaek!

PR #6945, opened 2017-12-27
Fix: Remove key when value is null

localheinz:
This PR

* [x] asserts that `JsonManipulator::removeMainKey()` removes a key when its value is `null`
* [x] uses `array_key_exists()` instead of `isset()` to determine if data has a key


localheinz:
Thank you, @seldaek!

PR #6947, opened 2017-12-28
Fix: Add 'sort-packages' to composer-schema.json

localheinz:
This PR

* [x] adds a definition for `sort-packages` to `composer-schema.json`

Follows #4716.


res/composer-schema.json
```diff
@@ -278,6 +278,10 @@
                 "htaccess-protect": {
                     "type": "boolean",
                     "description": "Defaults to true. If set to false, Composer will not create .htaccess files in the composer home, cache, and data directories."
+                },
+                "sort-packages": {
```
localheinz: Since the definitions do not appear to be sorted here, I just added it at the end.

localheinz:
Thank you, @seldaek!

PR #6949, opened 2017-12-29
[docs] Fix links

mhujer:
Some of the links were either not using https or linking to a redirecting page, so I fixed them.


Seldaek:
Thanks!

PR #6956, opened 2017-12-31
Fix for issue 6602 - warning when running composer as another user

wsandin:
Raising a warning if the current user doesn't match the initial ownership of the $home directory. 
Not sure how this works on let's say Windows, might have to add some extra checks for that.

Would love some feedback on how I can improve this PR in order to merge into master.

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -93,7 +93,9 @@ EOT
         $cacheDir = $config->get('cache-dir');
         $rollbackDir = $config->get('data-dir');
         $home = $config->get('home');
+        $homeOwner = posix_getpwuid(fileowner($home));
         $localFilename = realpath($_SERVER['argv'][0]) ?: $_SERVER['argv'][0];
+        $composeUser = posix_getpwuid(posix_geteuid());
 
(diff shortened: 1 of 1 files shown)


src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -93,7 +93,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $cacheDir = $config->get('cache-dir');
         $rollbackDir = $config->get('data-dir');
         $home = $config->get('home');
+        $homeOwner = posix_getpwuid(fileowner($home));
         $localFilename = realpath($_SERVER['argv'][0]) ?: $_SERVER['argv'][0];
+        $composeUser = posix_getpwuid(posix_geteuid());
```
Seldaek: This check should only be done if posix_getpwuid / posix_geteuid are defined and callable, as the posix functions are typically disabled or not available in some environments and we don't want to fail the update in those cases.

wsandin: Thanks @Seldaek. Will have a look into how I can improve this.

Seldaek:
Thanks

wsandin:
Thanks, would love to contribute more to composer in the future.

PR #6958, opened 2017-12-31
possibility to remove wget as dependency

ktomk:
use php to download the signature to standard output instead of wget.

just stumbled over the installer script and used it with some docker image that had no wget and then wondered why to use wget while the other downloads in that file are done via php.

happy new years eve btw


OpenUAS:
The sooner honored the better...

Seldaek:
The problem is php might be misconfigured and hence not doing https secure, I don't think that's as likely or even possible with the wget variant. I'd rather keep it as it is and let people figure it out if they don't want to have wget installed.

PR #6961, opened 2018-01-02
Allow bin key to receive string

carusogabriel:
Something that I always see in packages, including `Composer`, is an `array` wrapper in a single bin file:
```json
"bin": ["bin/composer"],
```

So, the way we do with our `licenses` key, why not support `array` and `string` in the `bin` key? :nerd_face:

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -368,8 +368,8 @@
             "description": "If set to true, stable packages will be preferred to dev packages when possible, even if the minimum-stability allows unstable packages."
         },
         "bin": {
-            "type": ["array"],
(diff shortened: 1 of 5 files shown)


Seldaek:
Going to merge this because why not but a word of caution to anyone looking to use this change: It will break your package for existing/older Composer versions, so I would not advise anyone use this for the next few months at least to give people time to upgrade Composer.

carusogabriel:
@Seldaek Sure. I'll track `Travis CI` and when they got the version compatibility with this feature, I propose to change our `composer.json` file to encourage everyone to use it. Thanks for accepted it :pray:

PR #6964, opened 2018-01-03
[Refactoring] Use array_filter and array_map

carusogabriel:


src/Composer/Repository/ComposerRepository.php
```diff
@@ -207,16 +207,13 @@ public function search($query, $mode = 0, $type = null)
...
-                    $results[] = array('name' => $name);
-                }
-            }
-
-            return $results;
+            return array_map(function ($name) {
+                return compact('name');
+            }, array_filter($this->getProviderNames(), function ($name) use ($regex) {
+                return preg_match($regex, $name);
+            }));
```
Seldaek: Sorry but changes like these I really don't find an improvement.. Sure it's 2 less lines but they're much more packed lines, use weird stuff like compact instead of a simple array literal. It's smarter code maybe but it's not more readable.

Seldaek:
I don't really see the point of this overall. It's not adding any value it just changes code for the sake of changing it.

PR #6973, opened 2018-01-04
Fix BC of generated static map

nicolas-grekas:
This needs an emergency release, as it breaks using legacy class-loader, like the one bundled into phpunit.
ping @Seldaek

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -785,10 +785,7 @@ HEADER;
 
         // BC handling when converting to a new ClassLoader
         if (isset($maps['prefixLengthsPsr4'])) {
-            $maps['firstCharsPsr4'] = array_map(function () {
-                return true;
-            }, $maps['prefixLengthsPsr4']);
-            unset($maps['prefixLengthsPsr4']);
+            $maps['firstCharsPsr4'] = array_map('is_array', $maps['prefixLengthsPsr4']);
         }
 
         foreach ($maps as $prop => $value) {
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -44,6 +44,7 @@ class ClassLoader
 {
     // PSR-4
     private $firstCharsPsr4 = array();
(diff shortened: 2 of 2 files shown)


Seldaek:
Any details on what/how it breaks? Just trying to understand here..

nicolas-grekas:
How: see https://travis-ci.org/symfony/symfony/jobs/324996450
Why: missing `prefixLengthsPsr4` map in newly dumped static maps breaks the ClassLoader provided by phpunit (old one)

Seldaek:
I see.. but I am not sure if the fix will really fix it then. What if you dump a completely new loader into that dir (so it won't add the `prefixLengthsPsr4` at all) and then phpunit will break?

nicolas-grekas:
What about reverting 3be9591930688dc8be146095a321cdb280098e2b then?
With static arrays in shared memory, the gain it non existent anyway.

Seldaek:
Yeah I think that might be the only way forward. It sucks that we basically can't make any change to this code though with the static array. I'll work on a partial revert as we can still keep some of the changes.

nicolas-grekas:
(revert pushed)

nicolas-grekas:
>  I'll work on a partial revert as we can still keep some of the changes

already done here :)

Seldaek:
Yup I see, sorry have 5 things going on atm :/ Looks good to me except I don't know if it'll break people that upgraded to 1.6.0 and already dumped a new autoloader.. Maybe we need reverse-BC code?

nicolas-grekas:
Here we are.

Seldaek:
Thanks.. Let's hope no further adverse effect.

nicolas-grekas:
I beg for a new tag please :)

Seldaek:
Yes yes sorry everything is on fire right now, on it ASAP

Seldaek:
Also why are you the only one affected here? Anything special symfony is doing? I just wonder why travis builds haven't caught fire everywhere.

nicolas-grekas:
I think it's because of a cached ClassLoader that were compiled when symfony/phpunit-bridge compiled its variant of phpunit. There must be some unusual autoloader loading logic there.

Seldaek:
Ok 1.6.1 is out

Issue #6937, opened 2017-12-23
Redundant prefixLengthsPsr4 lookup in ClassLoader::findFileWithExtension()


drscre:
I am looking at the following piece of code from [ClassLoader::findFileWithExtension](
https://github.com/composer/composer/blob/master/src/Composer/Autoload/ClassLoader.php#L375):

```php
if (isset($this->prefixLengthsPsr4[$first])) {
    $subPath = $class;
    while (false !== $lastPos = strrpos($subPath, '\\')) {
        $subPath = substr($subPath, 0, $lastPos);
        $search = $subPath.'\\';
        if (isset($this->prefixDirsPsr4[$search])) {
            $length = $this->prefixLengthsPsr4[$first][$search]; // === $lastPos + 1 === strlen($search) ?
            foreach ($this->prefixDirsPsr4[$search] as $dir) {
                if (file_exists($file = $dir . DIRECTORY_SEPARATOR . substr($logicalPathPsr4, $length))) {
                    return $file;
                }
            }
        }
    }
}
```
And it seems to me that `$length = $this->prefixLengthsPsr4[$first][$search];` is the same as the length of `$search` which in turn is equal to `$lastPos + 1`.

As far as i understand, `$length` is the length  of namespace prefix we want to substitute for directory.
So, probably to avoid `strlen($search)` function call (is it that slow?), the length of namespace prefix is looked up in pre-computed prefix length map `$this->prefixLengthsPsr4`:

`$length = $this->prefixLengthsPsr4[$first][$search];`

But it turns out even this lookup is redundant, because we already know the prefix length after calling `strrpos`:

`$length = $lastPos + 1;`

P.S.
`substr($logicalPathPsr4, $length)` can also be taken out of `foreach ($this->prefixDirsPsr4[$search] as $dir)` loop

Seldaek:
Good catch, this allows to simplify the code a little.

PR #6976, opened 2018-01-04
Fix generated static map...

nicolas-grekas:
Follow up of #6937 - No urgency on Symfony's side: I cleared the Travis cache :)

I completely missed this: the generated static map relies on the actual properties of the ClassLoader at compile time. Thus, the `prefixLengthsPsr4` property must be populated if we want the static map to be correcty built (for compat with legacy ClassLoader implementations).

As a bonus (hum...), the static initializer now only considers existing properties. The legacy one created public dynamic properties. This should improve BC/FC in the future.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -43,8 +43,7 @@ namespace Composer\Autoload;
 class ClassLoader
 {
     // PSR-4
-    private $firstCharsPsr4 = array();
-    private $prefixLengthsPsr4 = array(); // For BC with legacy static maps
+    private $prefixLengthsPsr4 = array();
     private $prefixDirsPsr4 = array();
     private $fallbackDirsPsr4 = array();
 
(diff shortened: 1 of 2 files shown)


src/Composer/Autoload/ClassLoader.php
```diff
@@ -175,6 +175,7 @@ public function addPsr4($prefix, $paths, $prepend = false)
                 throw new \InvalidArgumentException("A non-empty PSR-4 prefix must end with a namespace separator.");
             }
             $this->firstCharsPsr4[$prefix[0]] = true;
+            $this->prefixLengthsPsr4[$prefix[0]][$prefix] = \strlen($prefix);
```
Seldaek: Now this means we keep both maps forever though, with the dual isset() too.. not sure if this is a great outcome :) Could the static loader not check if firstCharsPsr4 is there then to ignore the other one entirely?

nicolas-grekas: actually I think this is the way to provide a migration path to kill `prefixLengthsPsr4`:
once v1.6.2 is legacy, everyone will have both properties working, so that we could then drop it.
the generated static map must contain boths maps now... (or we can just drop firstCharsPsr4 and not care about a 4hours old BC issue)

Seldaek: Yes maybe just dropping it now is the best approach. Curious what #6977 is about though before doing any other release.

nicolas-grekas:
PR now totally reverting 3be959193

Seldaek:
Thanks, I'll wait until tomorrow and then release.

PR #6982, opened 2018-01-06
Fixes from PHPStan level 0

carusogabriel:
Some, not all, fixes found by running [`PHPStan`](https://github.com/phpstan/phpstan) at level 0:

File | Problem
---- | ----
src/Composer/DependencyResolver/Rule.php:56 | Access to an undefined property Composer\DependencyResolver\Rule::$job.
src/Composer/Factory.php:549 | Class Composer\Installer\MetapackageInstaller does not have a constructor and must be instantiated without any parameters.
src/Composer/Command/DiagnoseCommand.php:38 | Property Composer\Command\DiagnoseCommand::$rfs has unknown class Composer\Command\RemoteFileSystem as its type.
src/Composer/Command/InitCommand.php:282 | Anonymous function has an unused use $self.
src/Composer/Repository/RepositoryInterface.php:73 | Return typehint of method Composer\Repository\RepositoryInterface::search() has invalid type array.
src/Composer/Package/Archiver/BaseExcludeFilter.php:78 | Parameter $lineParser of method Composer\Package\Archiver\BaseExcludeFilter::parseLines() has invalid typehint type Composer\Package\Archiver\callback.
src/Composer/Installer/PluginInstaller.php: 37 | Constructor of class Composer\Installer\PluginInstaller has an unused parameter $type.
src/Composer/Json/JsonManipulator.php:232 | Anonymous function has an unused use $name.

### DIFF
--- a/src/Composer/Package/Archiver/BaseExcludeFilter.php
+++ b/src/Composer/Package/Archiver/BaseExcludeFilter.php
@@ -71,7 +71,7 @@ abstract class BaseExcludeFilter
      * Processes a file containing exclude rules of different formats per line
      *
      * @param array    $lines      A set of lines to be parsed
-     * @param callback $lineParser The parser to be used on each line
+     * @param callable $lineParser The parser to be used on each line
      *
      * @return array Exclude patterns to be used in filter()
      */
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -34,7 +34,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 class DiagnoseCommand extends BaseCommand
 {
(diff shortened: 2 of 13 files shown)


src/Composer/Package/Archiver/BaseExcludeFilter.php
```diff
@@ -71,7 +71,7 @@ public function filter($relativePath, $exclude)
      * Processes a file containing exclude rules of different formats per line
      *
      * @param array    $lines      A set of lines to be parsed
-     * @param callback $lineParser The parser to be used on each line
+     * @param \Closure $lineParser The parser to be used on each line
```
jvasseur: This should be `callable` instead of `\Closure`

Seldaek:
Thanks

PR #6984, opened 2018-01-07
Added instruction to disable xdebug cli

girol:
Hello guys,

I've found this solution [here](https://chrisloftus.github.io/tooling/2016/05/05/how-to-disable-xdebug-php7-cli-ubuntu/), that helps when setting up the composer environment.

Worked fine for me and is a simple solution for solving the xdebug performance issue globally, without having to mess up with aliases.

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -153,6 +153,13 @@ This issue can also happen on cPanel instances, when the shell fork bomb protect
 To improve performance when the xdebug extension is enabled, Composer automatically restarts PHP without it.
 You can override this behavior by using an environment variable: `COMPOSER_ALLOW_XDEBUG=1`.
 
+You can also disable xdebug, only in cli, using the following command (on Linux boxes):
+
+```bash
+# As root:
+phpdismod -s cli xdebug
+```
+
 Composer will always show a warning if xdebug is being used, but you can override this with an environment variable:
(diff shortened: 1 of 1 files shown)


Seldaek:
This is not needed anymore as we disable xdebug anyway. That blog post you linked to is outdated IMO. I don't really feel like documenting how to disable xdebug on all possible environments as it's not really a step people need to take.

girol:
Hello @Seldaek I didn't know about that. Thanks for your time and response!

I wrote that because I'm starting to use composer and got the same error and worked for me. Even with the auto-disable feature, I've noticed a significant impact using composer after the procedure.

Anyway, thank you!

Seldaek:
@girol hmm that's odd that you would see an impact still. What kind of impact are we talking about? Have you tried running with -vvv to see what's going on? Maybe you can share a full composer install run or something with -vvv so we can have a look.

@johnstevenson this brought up a thought here, should the restart really happen for everything? IMO xdebug only has a bad impact on the solver, the rest it's pretty negligible, so maybe we should disable it only on install/update/require/remove as those are CPU heavy commands. The other commands can probably run fine with it, and they'd benefit from not having to waste time restarting php. Wdyt?

johnstevenson:
@Seldaek Sure, we could conditionally invoke the xdebug-handler later on if it proves more performant.

girol:
@Seldaek I will test again and try to replicate my results. I've noticed some slowness when downloading packages and processing the deployment of packages, but I did *not* measure anything.

PR #6985, opened 2018-01-07
Add RemoteFilesystem::getRemoteContents() extension point

nicolas-grekas:
Using this new extension point, I'm able to use a shared cURL connection and speed up downloads by a significant margin (depends on round-trip latency).

This is not as efficient as prefetching as done by prestissimo, but on the other side, this plays well with the existing life cycle.


nicolas-grekas:
I just added `$originUrl` because it is useful to deal with authentication issues at this stage.

PR #6995, opened 2018-01-09
Use $strict param in functions that have it

carusogabriel:
Some functions, e.g. `in_array`, have a third param to enable strict comparison. With PHP-CS-Fixer `strict_param` rule, we can fix all of them :shipit:


Seldaek:
I don't really want to merge this blindly as this changes the semantics of the in_array calls, it has to be checked carefully, and I don't quite see what this brings.

PR #7024, opened 2018-01-19
A different set of slashes

Golodnyi:
On the Windows platform this code will form a path with a different set of slashes


Seldaek:
This doesn't seem to fix anything.. `/` works just fine on Windows. If you explain why you are doing this maybe I can reopen but as such I have to close.

Issue #7025, opened 2018-01-19
Composer update is deleting pathed Windows symlinked local repos again, ref #4955


narration-sd:
Looks exactly like #4955 --- do  you  have a regression?  

Very disconcerting, only saved because just had pushed the github repo to an origin.

My `composer.json`:

{
  "name": "craftcms/craft",
  "description": "Craft CMS",
  "keywords": [
    "craft",
    "cms",
    "craftcms",
    "project"
  ],
  "license": "MIT",
  "homepage": "https://craftcms.com/",
  "type": "project",
  "support": {
    "email": "support@craftcms.com",
    "issues": "https://github.com/craftcms/cms/issues",
    "forum": "https://craftcms.stackexchange.com/",
    "source": "https://github.com/craftcms/cms",
    "docs": "https://craftcms.com/docs",
    "rss": "https://craftcms.com/changelog.rss"
  },
  "minimum-stability": "beta",
  "repositories": [
    {
      "type": "path",
      "url": "plugin-dev/live-vue",
      "symlink": true
    }
  ],
  "require": {
    "php": ">=7.0.0",
    "craftcms/cms": "dev-develop as 3.0.0-RC99",
    "vlucas/phpdotenv": "^2.4.0",
    "roave/security-advisories": "dev-master",
    "craftcms/element-api": "^2.5",
    "craftcms/redactor": "^1.0.0-beta.1",
    "laravel/homestead": "^7.0.0",
    "narrationsd/live-vue": "^0.8.0"
  },
  "autoload": {
    "psr-4": {
      "modules\\": "modules/"
    }
  },
  "config": {
    "optimize-autoloader": true
  },
  "scripts": {
    "post-root-package-install": [
      "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
    ],
    "post-create-project-cmd": [
      "@composer dump-autoload -o",
      "@php craft setup/welcome"
    ]
  },
  "require-dev": {
  }
}

Output of `composer diagnose`:

C:\vagrant\gnu\seventh-page_html>composer diagnose
Checking composer.json: WARNING
require.roave/security-advisories : unbound version constraints (dev-master) should be avoided
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
[...]

PR #7032, opened 2018-01-23
provide fix for issue #7025

narration-sd:
This small change means consistent values will be returned for lstat modes on repositories 'symlinked' by composer require.  In truth these are linked by junctions when on Windows. 

With the prior statcacheclear according to file only, lstat() will return arbitrary values from other files or folders. With this general clear, php-windows will return a consistent zero, which although incorrect, suffices for Composer to determine it is indeed a junction, and use its correct handling.

The issues with php-windows lstat are known to the Composer team, and the serious problem from them in #7025 shows up at least on a fast multicore Windows machine, running php7.1-windows.  This patch appears to relieve this problem consistently.

### DIFF
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -665,7 +665,7 @@ class Filesystem
          * Stat cache should be cleared before to avoid accidentally reading wrong information from previous installs.
          */
         clearstatcache(true, $junction);
-        // clearstatcache(false);
+        clearstatcache(false);
         $stat = lstat($junction);
 
         return !($stat['mode'] & 0xC000);
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -664,7 +664,8 @@ class Filesystem
          *
          * Stat cache should be cleared before to avoid accidentally reading wrong information from previous installs.
          */
-        clearstatcache(false);
+        clearstatcache(true, $junction);
+        // clearstatcache(false);
         $stat = lstat($junction);
 
         return !($stat['mode'] & 0xC000);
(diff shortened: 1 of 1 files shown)


src/Composer/Util/Filesystem.php
```diff
@@ -663,8 +663,11 @@ public function isJunction($junction)
          * #define	_S_IFREG	0x8000
          *
          * Stat cache should be cleared before to avoid accidentally reading wrong information from previous installs.
+         
+         * n.b. now clearing the entire stat cache, for its troubles in php7.1-windows, appearing and
+         * diagnosed in composer github issue #7025 -- @narrationsd 22 Jan 2018
```
curry684: Comments in open source are not "personal", if someone wants to know why someone did what they'll grab the blame/commit logs. I don't think the comment change is needed at all, the previous one still describes the resulting situation fine.

Seldaek:
@curry684 do you concur that having *both* clearstatcache calls is necessary? Sounds a bit overkill but then again harmless.

narration-sd:
Thanks, @curry684 , @Seldaek .  Jordi, the reason for having the two is that only with both does this pass both my tests and yours under CI.  It is completely crazy, yes.  But we are dealing with flawed underlying code. Take care, Clive.

Seldaek:
Merged in 1.6/master as fd61a21bffb191a00980c64266ced43a820376e7 - thanks!

curry684:
@Seldaek my verdict here is also 'better safe than sorry' with both safeguards. It's harmless and costs no performance, and given the consequences if it goes wrong I'm all for doing both.

The code will be retired anyway if we move to Win10 real symlinks.

narration-sd:
err, the above test guarded with an isdir() check, of course; didn't quite say that...

Issue #6642, opened 2017-08-31
Gitlab driver expects visibility from API, but it does not exist if repository is public and client is not authendicated


mikk150:
My `composer.json`:

```json
{
    "minimum-stability": "dev",
    "require": {
        "yiisoft/yii2-queue": "^2.2.0",
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "https://gitlab.com/yii2-queue/yii2-queue"
        },
    ]
}

```

Output of `composer diagnose`:

```
# composer diagnose
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
```

When I run this command:

```
composer update
```

I get the following output:

```
                               
  [ErrorException]             
  Undefined index: visibility  
                               

```

And I expected this to happen:

should update my dependencies

gitlab does not tell if the project is public of private, if user is not authendicated... prolly because it seeing it while not being authendicated means it HAS to be public

mlncn:
Very very excited to have this fixed!  This seems to be the last step in finally being able to have GitLab projects managed by composer.

Exadra37:
Same error here when running `composer install` from docker.

#### Composer File

```json
{
    "name": "exadra37-docker-compose/laravel5-demo",
    "description": "A Laravel 5 demo to show how to use exadra37-docker-compose/laravel5.",
    "keywords": [
        "Exadra37",
        "laravel 5",
        "docker",
        "docker compose",
        "laravel 5 docker compose demo"
    ],
    "homepage": "https://gitlab.com/exadra37-docker-compose/laravel5-demo",
    "license": "GPL-3.0",
    "authors": [
        {
            "name": "Paulo Silva",
            "email": "exadra37ingmailpointcom",
            "homepage": "https://exadra37.com",
            "role": "Software Developer"
        }
    ],
    "repositories": [
        {
            "type": "vcs",
[...]

PR #7065, opened 2018-01-31
Correct a tiny typo

katzien:
`project b's` should be `project a's` in the example given

### DIFF
--- a/doc/articles/vendor-binaries.md
+++ b/doc/articles/vendor-binaries.md
@@ -66,7 +66,7 @@ Say project `my-vendor/project-b` has requirements setup like this:
 ```
 
 Running `composer install` for this `composer.json` will look at
-all of project-b's dependencies and install them to `vendor/bin`.
+all of project-a's dependencies and install them to `vendor/bin`.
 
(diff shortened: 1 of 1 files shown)


doc/articles/vendor-binaries.md
```diff
@@ -66,7 +66,7 @@ Say project `my-vendor/project-b` has requirements setup like this:
 ```
 
 Running `composer install` for this `composer.json` will look at
-all of project-b's dependencies and install them to `vendor/bin`.
+all of project-a's dependencies and install them to `vendor/bin`.
```
Seldaek: Oh that was real confusing. It should be `a` but it should also be binaries and not dependencies as far as I can tell.

katzien: A very good point! Changed that as well :)

Seldaek:
Thanks

andronuxx:
`marged`  
`fix usage and reslove all error  @files`
@termux app
#6642 
@katzien 
 ##@katzien heroku

andronuxx:
imvs0

PR #7070, opened 2018-02-01
Fix RemoteFilesystem::getRemoteContents() on-failure behavior

nicolas-grekas:
Instead of "Undefined variable $http_response_header" when network is off.


nicolas-grekas:
This one is easy to merge @Seldaek @naderman

PR #7075, opened 2018-02-02
Optimize

huangzhhui:
1. use strict syntax
2. optimize comments
3. improves maintainability
4. optimize some condition cost

### DIFF
--- a/bin/composer
+++ b/bin/composer
@@ -42,7 +42,7 @@ if (function_exists('ini_set')) {
 
     $memoryLimit = trim(ini_get('memory_limit'));
     // Increase memory_limit if it is lower than 1.5GB
-    if ($memoryLimit != -1 && $memoryInBytes($memoryLimit) < 1024 * 1024 * 1536) {
+    if ($memoryLimit !== -1 && $memoryInBytes($memoryLimit) < 1024 * 1024 * 1536) {
         @ini_set('memory_limit', '1536M');
     }
     // Set user defined memory limit
--- a/bin/composer
+++ b/bin/composer
@@ -42,7 +42,7 @@ if (function_exists('ini_set')) {
 
     $memoryLimit = trim(ini_get('memory_limit'));
     // Increase memory_limit if it is lower than 1.5GB
-    if ($memoryLimit !== -1 && $memoryInBytes($memoryLimit) < 1024 * 1024 * 1536) {
+    if ($memoryLimit !== '-1' && $memoryInBytes($memoryLimit) < 1024 * 1024 * 1536) {
         @ini_set('memory_limit', '1536M');
     }
     // Set user defined memory limit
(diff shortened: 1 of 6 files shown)


bin/composer
```diff
@@ -42,7 +42,7 @@ if (function_exists('ini_set')) {
 
     $memoryLimit = trim(ini_get('memory_limit'));
     // Increase memory_limit if it is lower than 1.5GB
-    if ($memoryLimit != -1 && $memoryInBytes($memoryLimit) < 1024 * 1024 * 1536) {
+    if ($memoryLimit !== -1 && $memoryInBytes($memoryLimit) < 1024 * 1024 * 1536) {
```
stof: that's broken. `$memoryLimit` is a string here

src/Composer/Cache.php
```diff
@@ -51,24 +52,34 @@ public function __construct(IOInterface $io, $cacheDir, $whitelist = 'a-z0-9.',
         }
 
         if (
+            !is_writable($this->root) ||
             (!is_dir($this->root) && !Silencer::call('mkdir', $this->root, 0777, true))
-            || !is_writable($this->root)
```
stof: this does not try to check for writable status after the directory was created anymore, so this might  change the logic now. Are you sure this change is 100% safe ?

src/Composer/Cache.php
```diff
@@ -81,6 +92,12 @@ public function read($file)
         return false;
     }
 
+    /**
+     * @param $file
+     * @param $contents
```
stof: putting phpdoc without putting types means this phpdoc is invalid.

src/Composer/Factory.php
```diff
@@ -148,7 +150,12 @@ protected static function getDataDir($home)
 
     /**
      * @param  IOInterface|null $io
-     * @return Config
+     * @param null              $cwd
+     * @return \Composer\Config
+     * @throws \Exception
+     * @throws \InvalidArgumentException
+     * @throws \RuntimeException
+     * @throws \UnexpectedValueException
```
stof: Adding all these `@throws` without any explanation about what case could trigger them makes this phpdoc totally useless for handling recovery (and if you cannot recover, you don't care about which kind of exception will be thrown)

src/Composer/Cache.php
stof: this comment applies to almost all phpdoc you added

src/Composer/Factory.php
```diff
@@ -148,7 +150,12 @@ protected static function getDataDir($home)
 
     /**
      * @param  IOInterface|null $io
-     * @return Config
+     * @param null              $cwd
```
stof: wrong type. It is not always `null`

src/Composer/Cache.php
huangzhhui: you are right, i revert it now

bin/composer
huangzhhui: then i will change -1 to string type

Issue #7086, opened 2018-02-06
Composer tries/fails to remove PHP extensions when one is listed in "provide"


paragonie-scott:
My `composer.json`:

https://github.com/paragonie/sodium_compat/blob/a511834959a072000b7c328e8f02b957d4fd8430/composer.json

Output of `composer diagnose`:

```Cannot create cache directory /home/scott/.composer/cache/repo/https---packagist.org/, or directory is not writable. Proceeding without cache
Cannot create cache directory /home/scott/.composer/cache/files/, or directory is not writable. Proceeding without cache
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 1.6.3
PHP version: 7.2.1-1+ubuntu17.10.1+deb.sury.org+1
PHP binary path: /usr/bin/php7.2
```

When I run this command:

```
composer update -vv
```

I get the following output:

```
Cannot create cache directory /home/scott/.composer/cache/repo/https---packagist.org/, or directory is not writable. Proceeding without cache
Cannot create cache directory /home/scott/.composer/cache/files/, or directory is not writable. Proceeding without cache
Loading composer repositories with package information
Updating dependencies (including require-dev)
Dependency resolution completed in 0.075 seconds
Analyzed 1654 packages to resolve dependencies
Analyzed 83727 rules to resolve dependencies
Dependency resolution completed in 0.000 seconds
Package operations: 0 installs, 0 updates, 1 removal
Removals: ext-sodium

                                                
  [InvalidArgumentException]                    
  Package is not installed: ext-sodium-7.2.1.0  
                                                

Exception trace:
 () at phar:///usr/bin/composer/src/Composer/Installer/LibraryInstaller.php:123
 Composer\Installer\LibraryInstaller->uninstall() at phar:///usr/bin/composer/src/Composer/Installer/InstallationManager.php:211
[...]

PR #7088, opened 2018-02-07
Fix for #7086

paragonie-scott:
#7086


naderman:
This isn't a solution, it won't uninstall the package, but it will resolve other dependencies as if it could delete the extension, so the fix needs to be made elsewhere.

PR #7098, opened 2018-02-09
Fix permissions when using the PearBinaryInstaller

PatrickRose:
The PearBinaryInstaller is old and not used by many people, so the
world writable permissions weren't caught


Seldaek:
Thanks, merged in 1.6/master as b90987fde

Issue #7085, opened 2018-02-06
Wording: Downgrading instead of Updating


flowl:
This is not a bug or a major deal, it's more a "wish" concerning the following wording:

When downgrading, composer says `Updating`:

    - Updating phing/phing (2.16.0 => 2.14.0): Downloading (100%)

It should be 

    - Downgrading phing/phing (2.16.0 => 2.14.0): Downloading (100%)

When the new version is smaller than the version before.

Seldaek:
Fixed in #7119

PR #7119, opened 2018-02-15
Wording:Use Downgrading instead of Updating

austris-argalis:
Add logic for choosing correct word for the action performed on a package (update/downgrade). 
Related to #7085.

Should work with both packages that use tags and those which does not

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -214,7 +214,8 @@ class FileDownloader implements DownloaderInterface
         $from = $initial->getPrettyVersion();
         $to = $target->getPrettyVersion();
 
-        $this->io->writeError("  - Updating <info>" . $name . "</info> (<comment>" . $from . "</comment> => <comment>" . $to . "</comment>): ", false);
+        $actionName = version_compare($from, $to, '<') ? 'Updating' : 'Downgrading';
+        $this->io->writeError("  - " . $actionName . " <info>" . $name . "</info> (<comment>" . $from . "</comment> => <comment>" . $to . "</comment>): ", false);
 
         $this->remove($initial, $path, false);
         $this->download($target, $path, false);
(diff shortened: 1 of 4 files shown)


gmsantos:
I always wanted to PR this fix +1

src/Composer/Downloader/FileDownloader.php
```diff
@@ -214,7 +214,8 @@ public function update(PackageInterface $initial, PackageInterface $target, $pat
         $from = $initial->getPrettyVersion();
         $to = $target->getPrettyVersion();
 
-        $this->io->writeError("  - Updating <info>" . $name . "</info> (<comment>" . $from . "</comment> => <comment>" . $to . "</comment>): ", false);
+        $actionName = version_compare($from, $to, '<') ? 'Updating' : 'Downgrading';
```
stof: you should compare the normalized versions rather than the pretty versions.

And how does this deal with going from `v2.5.1` to `dev-my-feature` ? there are not comparable.

src/Composer/Downloader/VcsDownloader.php
```diff
@@ -242,6 +243,23 @@ protected function cleanChanges(PackageInterface $package, $path, $update)
...
+     *
+     * @param  PackageInterface $initial
+     * @param  PackageInterface $target
+     * @param  string           $operation
+     * @return bool
+     */
+    protected function packageCompare(PackageInterface $initial, PackageInterface $target, $operation = '<')
+    {
+        if ($initial->getPrettyVersion() == $target->getPrettyVersion()) {
+            return false;
```
stof: this looks weird to me, as `=`, `>=` and `<=` are supported operations in `version_compare`

Seldaek:
Thanks, I tweaked things a bit so that going from or to a dev-* version is always considered an update, because we really can't assert any other way.

PR #7120, opened 2018-02-16
 Update 00-intro.md

alexanderilyin:
1. Added Index
2. Added Homebrew


Seldaek:
We don't support homebrew as it breaks too often, and the index is generated on the fly on getcomposer.org so rather not break all the headings/anchors.

PR #7122, opened 2018-02-19
Added space to concatenation

davidyell:
Fix a missed space for string concatenation

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -377,7 +377,7 @@ class ClassLoader
             $subPath = $class;
             while (false !== $lastPos = strrpos($subPath, '\\')) {
                 $subPath = substr($subPath, 0, $lastPos);
-                $search = $subPath.'\\';
+                $search = $subPath . '\\';
                 if (isset($this->prefixDirsPsr4[$search])) {
                     $pathEnd = DIRECTORY_SEPARATOR . substr($logicalPathPsr4, $lastPos + 1);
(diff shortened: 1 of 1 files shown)


localheinz:
At the moment it's not quite clear whether the code base should consistently use 

* one
* no

space for string concatenation.

What do you think about adding a rule to [`.php_cs`](https://github.com/composer/composer/blob/master/.php_cs) to enable the `concat_space` fixer:

```diff
diff --git a/.php_cs b/.php_cs
index ff43b358..6433e02a 100644
--- a/.php_cs
+++ b/.php_cs
@@ -27,6 +27,7 @@ return PhpCsFixer\Config::create()
         'binary_operator_spaces' => true,
         'blank_line_before_return' => true,
         'cast_spaces' => true,
+        'concat_space' => array('spacing' => 'none'),
         'header_comment' => array('header' => $header),
         'include' => true,
         'array_syntax' => array('syntax' => 'long'),
```

and then run

```
$ composer global require friendsofphp/php-cs-fixer
$ php-cs-fixer fix --diff --verbose
```

?

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.10.4#usage:

>**concat_space** [`@Symfony`]
>
>Concatenation should be spaced according configuration.
>
>Configuration options:
>
>* `spacing` (`'none'`, `'one'`): spacing to apply around concatenation operator; defaults to `'none'`

davidyell:
I just assumed it would be PSR2.

PR #7140, opened 2018-02-27
Update 05-repositories.md - Fix typo

ErwanTouba:
typo

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -575,7 +575,7 @@ lightweight, static file-based version of packagist.
 
 You give it a `composer.json` containing repositories, typically VCS and
 package repository definitions. It will fetch all the packages that are
-`require`d and dump a `packages.json` that is your `composer` repository.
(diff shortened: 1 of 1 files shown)


doc/05-repositories.md
```diff
@@ -575,7 +575,7 @@ lightweight, static file-based version of packagist.
 
 You give it a `composer.json` containing repositories, typically VCS and
 package repository definitions. It will fetch all the packages that are
-`require`d and dump a `packages.json` that is your `composer` repository.
+`required` and dump a `packages.json` that is your `composer` repository.
```
localheinz: I don't believe this is a type, but refers to running

```
$ composer require
```

That is, either we keep it as it is, or modify as such

```diff
-`require`d and dump a `packages.json` that is your `composer` repository.
+required and dump a `packages.json` that is your `composer` repository.
```

What do you think?

Seldaek:
Indeed not a typo but thanks :)

PR #7148, opened 2018-03-01
Remove the word "just" from documentation

adactio:
Although the word "just" is intended to convey simplicity, it can have the unintended side effect of make the reader feel disempowered:

http://bradfrost.com/blog/post/just/


Seldaek:
Replaced by #7206 which is more exhaustive but thanks

PR #7172, opened 2018-03-07
Fix: Add type field to schema for inline-package

localheinz:
This PR

* [x] adds a `type` field to the definition of an `inline-package`

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -683,6 +683,7 @@
             }
         },
         "inline-package": {
+            "type": "object",
             "required": ["name", "version"],
             "properties": {
                 "name": {


res/composer-schema.json
```diff
@@ -683,6 +683,7 @@
             }
         },
         "inline-package": {
+            "type": "object",
```
localheinz: While it is apparently not necessary to provide a `type`, we clearly describe an object here.

Admittedly, I'm only proposing this change here as otherwise I would need to guess the type in [`localheinz/json-normalizer`](https://github.com/localheinz/json-normalizer/blob/0.5.2/src/SchemaNormalizer.php#L263-L274) (used for [`localheinz/composer-normalize`](https://github.com/localheinz/composer-normalize)). 

Also see https://github.com/json-schema/json-schema/issues/172.

staabm:
hmm I am curious what are those `inline-package` are about..
these are not mentioned on https://getcomposer.org/doc/04-schema.md

@localheinz it seems you know what they are supposed todo... would you be willing to describe them in a few sentences in the docs?

localheinz:
@staabm 

Here is an example of an `inline-package`: https://github.com/justinrainbow/json-schema/blob/dd084db298119352cc854e668e52bff68a408d8a/composer.json#L28-L36.

Came across this issue when attempting to run

```
$ composer normalize
```

on this specific `composer.json` and found it didn't actually fix it. 

You are right, the documentation doesn't actually explain the differences between the different package schemas. However, there's actually an example of an `inline-package` at

*  https://github.com/composer/composer/blob/master/doc/05-repositories.md#package-1

or 

* https://getcomposer.org/doc/05-repositories.md#package-2

What do you think, does that suffice?

staabm:
Thanks for those examples. 

I know see that there is no `inline-package` key but its just a symbolic name when using the `package` structure within the `repositories`.
This wasnt clear for me in the first place but I got that after looking at your links.

Thats fine for me

localheinz:
Thank you, @Seldaek!

PR #7174, opened 2018-03-08
Use external XdebugHandler library

johnstevenson:
This is a direct replacement for the existing functionality (I haven't wired in the `-vvv` debug output at this time).

I've kept the `Composer\XdebugHandler` class but added a deprecated notice. I have also kept the `Composer\Factory::createOutput()` method - which was introduced just for XdebugHandler. But I don't know if these decisions are correct?


src/Composer/XdebugHandler.php
```diff
@@ -12,290 +12,11 @@
...
-use Symfony\Component\Console\Output\OutputInterface;
+trigger_error('The ' . __NAMESPACE__ . '\XdebugHandler class is deprecated, use Composer\XdebugHandler\XdebugHandler instead.', E_USER_DEPRECATED);
 
 /**
- * @author John Stevenson <john-stevenson@blueyonder.co.uk>
+ * @deprecated use Composer\XdebugHandler\XdebugHandler instead
  */
 class XdebugHandler
 {
-    const ENV_ALLOW = 'COMPOSER_ALLOW_XDEBUG';
```
felixfbecker: If you remove all the code inside this class, then this would be a breaking change? And if it is, you could just remove the class altogether...
I don't know how the public interface of the composer package is defined. Do you use `@internal`/`@public` annotations on classes?

johnstevenson: Yup. I simply followed what has been done with other classes that have been extracted to an external library, by adding the deprecation error. And I've questioned this action in the main comment because I have no idea how this is meant to work either!

alcohol: No, you forgot to extend the replacement. But if the replacement is not BC, then this will require a new release regardless. @Seldaek can probably explain this better than me though.

alcohol:
Meant to comment, not approve..

src/Composer/XdebugHandler.php
johnstevenson: @alcohol Oops - thanks for that

felixfbecker: What are the API differences?

src/Composer/XdebugHandler.php
```diff
@@ -12,290 +12,20 @@
 
 namespace Composer;
 
-use Composer\Util\IniHelper;
 use Symfony\Component\Console\Output\OutputInterface;
 
+trigger_error('The ' . __NAMESPACE__ . '\XdebugHandler class is deprecated, use Composer\XdebugHandler\XdebugHandler instead,', E_USER_DEPRECATED);
```
felixfbecker: maybe mention that the user needs to install a different composer package?
E.g. _...from the composer/xdebug-handler package_

src/Composer/XdebugHandler.php
```diff
@@ -12,290 +12,20 @@
...
 /**
- * @author John Stevenson <john-stevenson@blueyonder.co.uk>
+ * @deprecated use Composer\XdebugHandler\XdebugHandler instead
  */
-class XdebugHandler
+class XdebugHandler extends XdebugHandler\XdebugHandler
 {
     const ENV_ALLOW = 'COMPOSER_ALLOW_XDEBUG';
     const ENV_VERSION = 'COMPOSER_XDEBUG_VERSION';
-    const RESTART_ID = 'internal';
```
felixfbecker: Since `const`s are public this should be kept?

src/Composer/XdebugHandler.php
johnstevenson: Maybe, but I've just copied what has been done for other packages to keep it consistent.

src/Composer/XdebugHandler.php
johnstevenson: `RESTART_ID` is already defined

src/Composer/XdebugHandler.php
stof: well, as it is a dependency of composer/composer, it is already installed when you get this version of composer/composer.

theofidry:
@johnstevenson did I miss something or the following statements (in `XdebugHandler::writeTmpIni()`) have been completely removed?

```php
$content .= 'allow_url_fopen='.ini_get('allow_url_fopen').PHP_EOL;
$content .= 'disable_functions="'.ini_get('disable_functions').'"'.PHP_EOL;
$content .= 'memory_limit='.ini_get('memory_limit').PHP_EOL;
```

johnstevenson:
@theofidry Yes. These were only added in Composer to cover some common `-d setting=value` uses cases.

They are not needed now that loaded ini settings are merged in.

PR #7184, opened 2018-03-12
Use string instead of array when loading bin files

carusogabriel:
Using https://github.com/composer/composer/pull/6961

Edit: Should we wait more time to start use it, or is there something else to do?


Seldaek:
I'd rather not do this as it breaks old composer versions and brings no new functionality.

PR #7185, opened 2018-03-13
Load autoload files for global package

joelwurtz:
Actually only a class loader is created, as an example functions defined in global package are never loaded.

However i'm not sure if this is the right way to do that.

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -177,6 +177,14 @@ class PluginManager
         $classLoader = $generator->createLoader($map);
         $classLoader->register();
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Also not sure what the consequences of this would be to be honest. I rather stay away from files autoloading in plugins because it doesn't cope well with multiple loads and might result in unexpected behavior.

joelwurtz:
Problem is that a lot of packages now provides functions that can only be autoloaded with a file, and so they are not available in plugins :/

Seldaek:
Closing this as really not something we want to allow I think. Plugins do have some restrictions and IMO we can live with that.

PR #7095, opened 2018-02-08
LocalChanges for ArchiveFiles

hectorprats:
Added LocalChanges detector for ArchiveFiles
Passed php-cs-fixer

PR #7191, opened 2018-03-14
LocalChanges for ArchiveFiles

hectorprats:
Added LocalChanges detector for ArchiveFiles, it is necesary for the command "composer status" for when you need to check if you have changed any dependency. 
Currently only checks Git, so the status says that they have not changed packages that have done so. With this we complete that process of truth with zip packages.

This PR comes from https://github.com/composer/composer/pull/7095 without CS fixes

### DIFF
--- a/src/Composer/Command/StatusCommand.php
+++ b/src/Composer/Command/StatusCommand.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Command;
 
+use Composer\Downloader\DownloaderInterface;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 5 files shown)


hectorprats:
No way...

Seldaek:
Thanks, merged and tweaked a bit in b1a78b60fe552bc6e7df09cb1c0154797fea6f64 - I am not sure if I like the idea though as it slows down the command quite dramatically but we'll see how it goes once people play with it.

hectorprats:
thanks

PR #7197, opened 2018-03-18
Choose correct url on saving credentials after redirects

ytorbyk:
Problem with storing credentials after redirects

What Composer does today:
- User requests package from hostname/foo/1.0.0.zip
- `hostname` does authorization 
- `hostname` redirects to location otherhost/foo/1.0.0.zip
- `other-host` download the package
- tries to save credentials for `other-host`

Problem is that there are no credentials for the `other-host` but are for `hostname`.

Solution:
Check if there exists credentials for the  `other-host` and if yes then store them.
Otherwise try to store credentials for `hostname`.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -504,7 +504,8 @@ class RemoteFilesystem
 
             if ($this->storeAuth && $this->config) {
                 $authHelper = new AuthHelper($this->io, $this->config);
-                $authHelper->storeAuth($this->originUrl, $this->storeAuth);
+                $originUrl = $this->io->hasAuthentication($this->originUrl) ? $this->originUrl : $originUrl;
+                $authHelper->storeAuth($originUrl, $this->storeAuth);
                 $this->storeAuth = false;
             }
 
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -504,8 +504,8 @@ class RemoteFilesystem
 
             if ($this->storeAuth && $this->config) {
                 $authHelper = new AuthHelper($this->io, $this->config);
-                $originUrl = $this->io->hasAuthentication($this->originUrl) ? $this->originUrl : $originUrl;
-                $authHelper->storeAuth($originUrl, $this->storeAuth);
+                $authUrl = $this->io->hasAuthentication($this->originUrl) ? $this->originUrl : $originUrl;
+                $authHelper->storeAuth($authUrl, $this->storeAuth);
                 $this->storeAuth = false;
             }
 


Seldaek:
I don't think the way it's done here is the correct approach as it kinda relies on luck and this exact setup for things to work. I think we should make sure we always store the credentials for the right domain after asking for them I guess, but that would require a much deeper refactoring of this code. Targetting for 2.0

ytorbyk:
I see, but now it doesn't work with at all. The fix could be applied in addition to this https://github.com/composer/composer/issues/6716 issue  with fix https://github.com/composer/composer/pull/6717.
Would be awesome to have this soon.

Seldaek:
As a workaround you can also set the credentials for the correct domains before running composer and then it will work fine.

ytorbyk:
Yes, I understand it. But we provide private composer for clients and the issue causes inconvenience for them.

duffner:
@Seldaek, there is a business problem here:  If you want to sell extensions with private authentication this fix works very well.  With the world of Amazon and other cloud hosting solutions, entering the correct "domain" is often unreadable, unbranded, and not easily recognizable or not even possible.  Would you want to sell extensions with an Amazon s3 domain as your code?  

This solution has no risk but is a BLOCKER for all paid extension developers who want privately branded URLs.    Please advise.

Seldaek:
@duffner that sounds like an unrelated issue though, if your private/branded domain has authentication setup and then redirects to an S3 authenticated URL it should be fine. This was fixed in 1.5.3 as "Fixed Authorization header being forwarded across domains after a redirect". Will you be at Imagine next week by any chance?

duffner:
@Seldaek, I will be at Imagine and yourself?

Seldaek:
@duffner yes I will be there and @naderman too, we can meet up and discuss that extension shop if you have a second as we are currently wrapping up a new feature for package vendors on https://packagist.com

ytorbyk:
@Seldaek Yes, you are right "Fixed Authorization header being forwarded across domains after a redirect" fix allows as to authorization, but after that use prompted to save entered credentials. Currently composer tries to save them for the domain in which it was redirected, sense it doesn't have them it saves next
```
"<s3 domain>": {
            "username": "null",
            "password": "null
        }
```
The fix allows to save credentials and it backward compatible. It doesn't break previous logic.
It checks if credentials exist for domain on which was redirected and only if doesn't it tries to save for the original domain.
I agree that it is not perfect, but still if fixes the case and doesn't break anything.

Seldaek:
Closed by https://github.com/composer/composer/pull/7904 which does process redirects better (the $origin is determined at every hop based on the hostname, and not anymore persisting across requests) and should not suffer from this issue anymore.

PR #7228, opened 2018-03-30
Fix usage of svn user-provided credentials; fix #7114

### LINKED ISSUE #7114: SVN credentials error
When running a composer install or update using external svn repos with basic authentication, the system user running the command is sent rather than the one prompted for svn creds.

No auth.json is being used since I am on server used by other devs, so creds are needed on every command run.

The issue is non-existent on composer 1.6.2, only on 1.6.3

ctrlaltca:
Before commit https://github.com/composer/composer/commit/9bc83d698e48dff2cd5559134c9c0877ca6977a4 , the svn command was recreated at every auth retry, in order to catch the new credentials provided by the used and add them to the command.
That commit excluded the `$command = $this->getCommand($svnCommand, $url, $path);` from the loop, so even if the user provides the svn credentials interactively, the old command missing the credentials will always be used. This causes issue #7114.
This PR moved the command regeneration back inside the retry loop, and also fixes up the `executeLocal()` method.

### DIFF
--- a/src/Composer/Util/Svn.php
+++ b/src/Composer/Util/Svn.php
@@ -102,9 +102,7 @@ class Svn
         // Ensure we are allowed to use this URL by config
         $this->config->prohibitUrlByConfig($url, $this->io);
 
-        $svnCommand = $this->getCommand($command, $url, $path);
-
-        return $this->executeWithAuthRetry($svnCommand, $cwd, $path, $verbose);
+        return $this->executeWithAuthRetry($command, $cwd, $url, $path, $verbose);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Seldaek:
@staabm can you double check this? It seems to me like 9bc83d698e48dff2cd5559134c9c0877ca6977a4 *added* credentials to executeLocal and this PR removes that.. so that might introduce yet another regression.

ctrlaltca:
Yep, that comment is a bit bad worded. Its original meaning was meant to be: "while a remove svn command needs both a remote url and a local path, a local svn command only needs a local path, that's why an empty string is passed as the remote url"

PR #7253, opened 2018-04-12
add readme to support fields in the schema docs

tflori:
This change is related to https://github.com/composer/packagist/pull/750

It should also be added to the schema.json but I don't know where it is. Is it stored in this repo or generated? And where is the definition of it?

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -237,6 +237,7 @@ Support information includes the following:
 * **source:** URL to browse or download the sources.
 * **docs:** URL to the documentation.
 * **rss:** URL to the RSS feed.
+* **readme:** Relative path to the readme document.
 
 An example:
 


Seldaek:
Thanks, the schema is at https://github.com/composer/composer/blob/1.6/res/composer-schema.json#L461 if you'd like to send a PR (please target the 1.6 branch)

tflori:
good morning,

That was wrong. sorry, my fault. I thought the support nodes where under root but they have a own node support. Readme is under root node. Should I commit both to 1.6 or fix the documentation in master and only the schema in 1.6?

Seldaek:
Oh yeah I totally missed that :) Please send a PR against 1.6, then I merge that in master.

Seldaek:
Ah but the docs are not present in 1.6 so you can't fix that in a 1.6 PR.. then please two PRs

PR #7284, opened 2018-04-24
Improve grammar in the programmatic Composer installation page

Calinou:
This improves grammar and uses official casing for UNIX and GitHub.


Seldaek:
Thanks

PR #7289, opened 2018-04-27
get classname from tokens

RobberPhex:
get classname from tokens, not via regex.

For a project with autoload:
```json
"classmap": [
    "classes/"
]
```
and file `classes/CLS.php`:
```php
<?
/** Class Guzzle
 */


class CLS
{
}
```

composer `1.6.4` will generated error result:
```shell
$ cat vendor/composer/autoload_classmap.php  | grep CLS.php
    'CLS' => $baseDir . '/classes/CLS.php',
    'Guzzle' => $baseDir . '/classes/CLS.php',
```


RobberPhex:
If this is merged, users should keep same php run time between composer and server.

For example, composer in php7 cannot recoginze `enum`, so generated `autoload_classmap.php` without enum info.
if server use hhvm, cannot load enum via classmap.

RobberPhex:
Looks like error is caused by phpunit, <https://travis-ci.org/composer/composer/jobs/371943844>.

stof:
using tokens was done in the past, and this was changed because some packages had files so huge that getting all tokens for the file would reach the memory limit.

stof:
fixing this bug could be done by stripping comments instead, as done already for stripping nowdoc and heredoc

src/Composer/Autoload/ClassMapGenerator.php
```diff
@@ -172,6 +185,10 @@ private static function findClasses($path)
                 return array();
             }
         }
+        // open tag must be '<?php' or '<?hh'
+        if (substr($contents, 0, 5) !== '<?php' && substr($contents, 0, 5) !== '<?hh') {
```
stof: wrong length for the HH case

RobberPhex:
Firstly, convert short tag to (normal) opening tag.
Then, comments can be stripped via php_strip_whitespace.

PR #7303, opened 2018-05-03
Force "C" locale to prevent issue with Turkish "I"

nicolas-grekas:
Fixes https://github.com/symfony/flex/issues/362
Forces case insensitive operations and lowercasing to consider `I` <=> `i`, ignoring the dot-less Turkish "i".


alcohol:
Would `LC_COLLATE` and `LC_CTYPE` not be enough?

Seldaek:
Heh, hadn't seen one of these Turkish locale bugs in ages..

PR #7316, opened 2018-05-08
Filter dev-dependencies from "dump-autoload --no-dev" , fixes #4343

### LINKED ISSUE #4343: dump-autoload --no-dev should not dump require-dev packages and dependents
Issue spawned from #4252 - This can be used for the implementation https://github.com/composer/composer/blob/master/src/Composer/Installer.php#L285-L304 but it should first be made a bit more reusable (probably still needs to be on the Installer itself though as it depends on so much of its internal methods, and refactoring that is another task than this issue).

ph-fritsche:
Strips entries from packageMap that are not included by recursion of Package->getRequires()

Testcases have been adjusted to prevent false positives

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -388,6 +388,7 @@ EOF;
     public function parseAutoloads(array $packageMap, PackageInterface $mainPackage)
     {
         $mainPackageMap = array_shift($packageMap);
+        $packageMap = $this->filterPackageMap($packageMap, $mainPackage);
         $sortedPackageMap = $this->sortPackageMap($packageMap);
         $sortedPackageMap[] = $mainPackageMap;
         array_unshift($packageMap, $mainPackageMap);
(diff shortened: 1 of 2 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -899,6 +900,48 @@ protected function getFileIdentifier(PackageInterface $package, $path)
...
+        $packages = array();
+        $include = array();
+
+        foreach ($packageMap as $item) {
+            $package = $item[0];
+            $name = $package->getName();
+            $packages[$name] = $package;
+        }
+
+        $add = function (PackageInterface $package) use (&$add, $mainPackage, $packages, &$include) {
```
alcohol: Why an inline closure? Why not just a private method on this class?

ph-fritsche: Sticking to the style already used in the sorting function below.

And since it - at least inside this class - probably won't be reused, i liked keeping the callback inline.
Guess it could make sense to implement the recursion on the PackageInterface, but - since it is my first work on the project - i didn't want to tamper with more code than necessary.

alcohol: Fair enough

Seldaek:
Thanks

SpacePossum:
thanks @ph-fritsche , much appreciated :+1:  : D

PR #7331, opened 2018-05-10
Places, where was used `composer.json` string directly coded is now obtained by `\Composer\Factory::getComposerFile()`, all places with more complex code changed to work properly.

tomFlidr:


Seldaek:
I am not sure I understand the point of doing this.. And in some cases it seems downright harmful as we don't want to allow people to change which file is looked up in VcsRepositories for example.

src/Composer/Repository/Vcs/VcsDriver.php
```diff
@@ -107,13 +107,13 @@ public function getComposerInformation($identifier)
...
     {
-        $composerFileContent = $this->getFileContent('composer.json', $identifier);
+        $composerFileContent = $this->getFileContent(ltrim(\Composer\Factory::getComposerFile(),'/.'), $identifier);
 
         if (!$composerFileContent) {
             return null;
         }
 
-        $composer = JsonFile::parseJson($composerFileContent, $identifier . ':composer.json');
+        $composer = JsonFile::parseJson($composerFileContent, $identifier . ':' . ltrim(\Composer\Factory::getComposerFile(),'/.'));
```
stof: This is really bad. Changing the name of the composer.json used locally must absolutely not impact the name used for repositories (the whole ecosystem will not have duplicated their config with your custom name)

alcohol:
Please, motivate a PR with either an issue or at least a description. This title tells us nothing. The changes proposed also do more harm than good.

Issue #7337, opened 2018-05-12
hiddeninput.exe not included in composer.phar


helhum:
If I understand the [phar compiler code](https://github.com/composer/composer/blob/master/src/Composer/Compiler.php#L83-L90) correctly, it will not include the `hiddeninput.exe` file provided by symfony/console

This is leading to https://github.com/helhum/TYPO3-Distribution/issues/17 when doing `composer create-project helhum/typo3-distribution`

It requires a composer plugin, which asks for db credentials using ` Composer\IO\ConsoleIO` in post-autoload-dump script event.

Any reason the exe is not included? If so, I will change my code to not ask for hidden input on Windows.

alcohol:
Probably not intentional, but I cannot say for certain. @Seldaek do you know if this is a bug?

Seldaek:
The reason is we ship our own https://github.com/composer/composer/blob/master/src/Composer/Compiler.php#L108 - I guess we need to include the one from symfony/console as well if it is present. I was not aware they added it there as well.

stof:
yeah, it was added in the console component a long time ago.

Composer may even drop its own instance now. the min version of the console component is high enough to ensure that symfony/console will be able to ask hidden questions.

helhum:
Thanks for the explanations. I proposed a pull request to use the API symfony/console already provides (and adding the binary found in this package)

PR #7349, opened 2018-05-14
Use symfony/console for hidden questions

helhum:
Instead of using a custom package, use symfony/console
functionality, which is already a requirement to composer

Composer command: `composer remove seld/cli-prompt`

Fixes: #7337


Seldaek:
Thanks

PR #7357, opened 2018-05-18
Improve SAT resolving developer debug information

Toflar:
Currently, when you debug using `-vvv`, with rather complex dependencies, your CLI output is stuck at `Resolving dependencies through SAT` for 20+ seconds and you don't know if Composer is still doing anything or it crashed, so I tried to improve this.

Before it just prompts after 22 seconds:

bildschirmfoto 2018-05-18 um 11 03 05

After:

ezgif com-video-to-gif

### DIFF
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -223,9 +223,10 @@ class Solver
         /* make decisions based on job/update assertions */
         $this->makeAssertionRuleDecisions();
 
-        $this->io->writeError('Resolving dependencies through SAT', true, IOInterface::DEBUG);
+        $this->io->writeError('Resolving dependencies through SAT', false, IOInterface::DEBUG);
         $before = microtime(true);
         $this->runSat(true);
+        $this->io->writeError('', true, IOInterface::DEBUG);
         $this->io->writeError(sprintf('Dependency resolution completed in %.3f seconds', microtime(true) - $before), true, IOInterface::VERBOSE);
 
         // decide to remove everything that's installed and undecided
@@ -762,6 +763,7 @@ class Solver
 
             for ($i = 0, $n = 0; $n < $rulesCount; $i++, $n++) {
                 if ($i == $rulesCount) {
+                    $this->io->writeError('.', false, IOInterface::DEBUG);
                     $i = 0;
                 }
 
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -223,7 +223,7 @@ class Solver
         /* make decisions based on job/update assertions */
         $this->makeAssertionRuleDecisions();
 
-        $this->io->writeError('Resolving dependencies through SAT', false, IOInterface::DEBUG);
+        $this->io->writeError('Resolving dependencies through SAT', true, IOInterface::DEBUG);
         $before = microtime(true);
         $this->runSat(true);
         $this->io->writeError('', true, IOInterface::DEBUG);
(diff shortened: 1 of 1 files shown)


src/Composer/DependencyResolver/Solver.php
```diff
@@ -762,6 +763,7 @@ private function runSat($disableRules = true)
 
             for ($i = 0, $n = 0; $n < $rulesCount; $i++, $n++) {
                 if ($i == $rulesCount) {
+                    $this->io->writeError('.', false, IOInterface::DEBUG);
```
staabm: I would love a single space before all those dots get printed

curry684:
When I added the "Resolving dependencies through SAT" log line I did that specifically because it isolates a single line of code that tends to run for a long time. I'm not sure the dots really help, I mean it's SAT so it cannot hang indefinitely by definition. I tried to find some more reliable way of displaying progress but didn't find one at the time (again, because SAT pretty much by definition does not know when it will find a solution).

Toflar:
I've placed it exactly at the place where a re-run happens. This happens like 20 times for my ruleset. So if you have a smaller one, likely there will be less re-runs and if you have more, it'll likely re-run more often. So it's a pretty accurate progress indicator although it will print dots faster towards the end because the number of comparisons needed gets smaller.
Yes, you cannot indicate anything like a percentage because you don't know how long it'll take in advance. And it doesn't have to be reliable at all. It's just an indicator that it's still running (and remember, it's only on debug level, so it won't show unless you pass `-vvv`).

alcohol:
I feel like the dots do not represent anything useful and merely clutter the output. If you're at the debug level already, at least output something debug worthy.

Toflar:
Like so?

```
Something's changed, looking at all rules again (pass #1)
Something's changed, looking at all rules again (pass #2)
Something's changed, looking at all rules again (pass #3)
[...]
Something's changed, looking at all rules again (pass #20)
Finished resolving dependencies through SAT
```

alcohol:
Sure, though you could overwrite the previous output, rather than outputting a new line again and again.

Toflar:
Here we go, now it looks like so:

ezgif com-optimize

src/Composer/DependencyResolver/Solver.php
```diff
@@ -759,10 +760,19 @@ private function runSat($disableRules = true)
             }
 
             $rulesCount = count($this->rules);
+            $pass = 1;
 
+            $this->io->writeError('Looking at all rules.', true, IOInterface::DEBUG);
             for ($i = 0, $n = 0; $n < $rulesCount; $i++, $n++) {
                 if ($i == $rulesCount) {
+                    if (1 === $pass) {
+                        $this->io->writeError("Something's changed, looking at all rules again (pass #$pass)", false, IOInterface::DEBUG);
```
staabm: IMO it shoud just be „Looking at all rules (pass #$pass)“.

No need for this more verbose (and non obvious) „something changed“ thing.

Toflar: I just took the comment there as an idea so. But yeah, whatever. Not gonna change this now, it can be changed to whatever the maintainers prefer.

curry684:
Just a thought - why the hell are we only going to do this on `-vvv`? It would appear to me the ultimate extra line of output *always* to have a `Resolving dependencies...` message which occasionally adds a dot, and will likely be just about as long the next time you run it, making it a semi-reliable indicator on the project level for everyone.

alcohol:
I'm sorry, but can you rephrase that? I'm not really sure what you are saying or asking right now.

curry684:
We've had tons of issues past years, complaining about Composer hanging indefinitely. This PR adds a progress indicator showing it's not dead or hanging. Why are we hiding it in `-vvv` instead of always in plain sight?

alcohol:
Fair point, I'll also leave that up to @Seldaek to decide though :d

Seldaek:
@naderman I think this loop would become redundant with Composer 2.0 refactorings as all providers would load first right? Or is this mistaken? Because if that's the case then I'd rather not merge as it'll become useless.

Toflar:
I don't think it's related. The output is during the recomputation in case any rule was marked as false or true. So it's SAT internal and not related to any previous stage.

Seldaek:
Alrighty then, thanks :)

Toflar:
Cheers, see you later :)

Issue #7383, opened 2018-06-01
Missing execution bits with ZipArchive fallback


mindplay-dk:
Composer quietly generates a non-working install when zip-archives get installed via the `ZipArchive` fallback on systems without `unzip`.

It took myself and one other developer a few hours to track down this issue, and it's very difficult to figure out where or how to document this issue for other developers.

This issue has been reported [dozens of times](https://github.com/composer/composer/issues?q=is%3Aissue+unzip+permission+is%3Aclosed), so it doesn't appear to be uncommon.

Composer ought to at least issue a warning if it has to fall back on `ZipArchive`, and @curry684 added the warning in #4966, which was merged in February'16 and gave an appropriate message:

```diff
         if (true !== $zipArchive->extractTo($path)) {
-            throw new \RuntimeException("There was an error extracting the ZIP file. Corrupt file?");	+            $this->io->writeError("<warn>As there is no 'unzip' command installed zip files are being unpacked using the PHP zip extension.</warn>");
+            $this->io->writeError("<warn>This may cause invalid reports of corrupted archives. Installing 'unzip' may remediate them.</warn>");
+
+            throw new \RuntimeException("There was an error extracting the ZIP file, it is either corrupted or using an invalid format");
         }	         }
```

It looks like the message was removed in a [later commit](https://github.com/composer/composer/commit/08d48cb61a87e82dc53f0865e4877c561f5c420f#diff-634696635d493cc5c70e0f42ee3fe55a) with the description "Prefer ZipArchive over unzip on Windows", which sounds like an unrelated change, so perhaps by accident?

The current version doesn't produce any warning or error and instead quietly produces an installation without the execution bit applied.

I'm running this version of `composer`:

    Running 1.6.5 (2018-05-04 11:44:59) with PHP 7.2.5-1+ubuntu16.04.1+deb.sury.org+1 on Linux / 4.4.0-17134-Microsoft

stof:
@mindplay-dk the error message was removed to avoid displaying it all the time on windows where ZipArchive is the preferred way, not the fallback way.

A much better way would have been to wrap the message in a condition to display it only on non-windows platforms.

mindplay-dk:
@stof like so?

PR #7387, opened 2018-06-04
add back the warning about missing unzip

mindplay-dk:
display an error-message on non-Windows OS if unzip is unavailable, per #7383

### DIFF
--- a/src/Composer/Downloader/ZipDownloader.php
+++ b/src/Composer/Downloader/ZipDownloader.php
@@ -56,6 +56,14 @@ class ZipDownloader extends ArchiveDownloader
             self::$hasZipArchive = class_exists('ZipArchive');
         }
 
+        if (!self::$hasZipArchive && !self::$hasSystemUnzip) {
+            // php.ini path is added to the error message to help users find the correct file
+            $iniMessage = IniHelper::getMessage();
+            $error = "The zip extension and unzip command are both missing, skipping.\n" . $iniMessage;
+
+            throw new \RuntimeException($error);
+        }
+
         if (null === self::$isWindows) {
             self::$isWindows = Platform::isWindows();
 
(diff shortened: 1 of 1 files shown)


JKetelaar:
Great change, I was wondering the same, but you already made the PR!

But what if there's a situation with the following conditions:
- Linux operating system
- `unzip` command not installed
- PHP `zip` extension not installed

Now you'll first see your warning:
```
As there is no 'unzip' command installed zip files are being unpacked using the PHP zip extension.
This may cause invalid reports of corrupted archives. Installing 'unzip' may remediate them.
```
But then the error from not having the PHP extension will also occur:
```
The zip extension and unzip command are both missing, skipping.
```

So in the end you'll have Composer first saying "_zip files are being unpacked using the PHP zip extension_" and then it will throw an error that it's not installed.

Wouldn't it be better to make the code like the following:
```
        if (null === self::$isWindows) {
            self::$isWindows = Platform::isWindows();
        }

        // Check if any ZIP module is installed
        if (!self::$hasZipArchive && !self::$hasSystemUnzip) {
            // php.ini path is added to the error message to help users find the correct file
            $iniMessage = IniHelper::getMessage();
            $error = "The zip extension and unzip command are both missing, skipping.\n" . $iniMessage;
            throw new \RuntimeException($error);

        // Warning that the OS unzip module isn't installed
        }elseif (!self::$isWindows && !self::$hasSystemUnzip) {
            $this->io->writeError("<warn>As there is no 'unzip' command installed zip files are being unpacked using the PHP zip extension.</warn>");
            $this->io->writeError("<warn>This may cause invalid reports of corrupted archives. Installing 'unzip' may remediate them.</warn>");
        }
```

mindplay-dk:
Instead of duplicating the conditions, I just swapped the tests around - if it throws because nothing is available, it won't reach the OS-specific test.

Seldaek:
Thanks

Seldaek:
Merged in 1.6 & master

PR #7400, opened 2018-06-10
chinese documentation

awei922:
Chinese document


alcohol:
Sorry I did not mean to close this. Can you get some other readers proficient in Chinese to review this? I could use Google translate but that would probably give very inaccurate results.

awei922:
OK，all right

PR #7416, opened 2018-06-19
Support json format for 'show tree'

### LINKED ISSUE #7244: [Feature request] `composer show --tree --format json`
Running `composer show --tree --format=json` currently prints the same tree as `composer show --tree`. This is much harder to parse in a script than the JSON format presented by `composer show --format=json`.

unglud:
I would like to resolve #7244.
I separated generation of tree from writing a tree to stdout

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -178,9 +178,6 @@ EOT
 
         // show single package or single version
         if (($packageFilter && false === strpos($packageFilter, '*')) || !empty($package)) {
-            if ('json' === $format) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks :)

unglud:
Thank you, man!

XedinUnknown:
Thanks a million! This will make the management of many granular packages much easier. For example, I can easily understand which dependencies of my dependency I need to release first, and in what order.

For those who wonder, this change is in version `1.7.0-RC`, which means it will be most likely available in `1.7.0`.

PR #7419, opened 2018-06-21
Update example for auth.json in BitBucket config

nathanpurcell:
In the current context (the paragraph above relating to auth.json), the sample given would be incorrect as it references a `config` node which is for [root only](https://getcomposer.org/doc/04-schema.md#config).


alcohol:
You are correct. There barely is any documention regarding auth.json at all it seems :-(

PR #7421, opened 2018-06-21
[WIP] make optimized autoloader respect PSR standards

### LINKED ISSUE #7352: Autoload optimizer violates PSR-4 and changes autoloading algorithm
When composer optimizes loading of files under PSR-4 root, it does not check correspondence between directory paths and namespaces. So if developer makes a mistake and puts php file in wrong directory, the application behaves differently when optimized and when not optimized.

I am not completely sure if we have to implement this. At first, this issue may be classified as "undefined behavior" and it isn't Composer's fault. Also it would be a breaking change for projects relying on this bug.  
But It's very productive when my tools help me to find errors on early stage, not hide them.

Composer version: 1.6.5

My directory layout:
```
src
|---BadDirectory
|   |---Example.php
composer.json
test.php
```

composer.json:
```json
{
    "autoload": {
        "psr-4": {
            "and800\\Example\\": "src/"
        }
    }
}

```

Example.php:
```
<?php
namespace and800\Example;
class Example {}
```

test.php
```
<?php
include __DIR__ . '/vendor/autoload.php';
new and800\Example\Example();
echo "autoload succeed";
```

When I run this command:

```
$ composer install && php test.php
$ composer install -o && php test.php
```

I get the following output:

```
PHP Fatal error:  Uncaught Error: Class 'and800\Example\Example' not found in /tmp/ctest/test.php:3
autoload succeed
```

```
PHP Fatal error:  Uncaught Error: Class 'and800\Example\Example' not found in /tmp/ctest/test.php:3
PHP Fatal error:  Uncaught Error: Class 'and800\Example\Example' not found in /tmp/ctest/test.php:3
```

and800:
fix #7352.
It isn't ready, needs work

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -255,7 +255,7 @@ EOF;
                             continue;
                         }
 
-                        $namespaceFilter = $namespace === '' ? null : $namespace;
+                        $namespaceFilter = array('name' => $namespace, 'type' => $group['type']);
                         $classMap = $this->addClassMapCode($filesystem, $basePath, $vendorPath, $dir, $blacklist, $namespaceFilter, $classMap);
                     }
                 }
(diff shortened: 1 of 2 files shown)


Seldaek:
@and800 what work still needs to be done here?

and800:
@Seldaek seems that the functionality is complete, only tests missing

kfb-ts:
@and800 do you think you'll find time to add tests soon or would you like someone to help you with that / take over?
@Seldaek are you happy with the implementation?

and800:
As I had deleted my fork, this PR cannot be modified now. So I've reopened it. ⬆️

PR #7424, opened 2018-06-24
Fix: Remove VersionEye badges

localheinz:
This PR

* [x] removes the VersionEye badges

Follows #4019.

💁‍♂️ It's been shut down a while ago. For reference, also see https://blog.versioneye.com/2017/10/19/versioneye-sunset-process/.


localheinz:
Thank you, @alcohol!

alcohol:
Thank you too :-)

PR #7432, opened 2018-06-29
Fix for composer.json file

c0ntax:
The `composer config repo.packagist false` command will add the following to you composer.json file
```json
{
    "repositories": [
        {
            "packagist": false
        }
    ]
}
```

So I'm pretty sure that the json example given should be `packagist` instead of `packagist.org`

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -700,7 +700,7 @@ You can disable the default Packagist.org repository by adding this to your
 {
     "repositories": [
         {
-            "packagist.org": false
+            "packagist": false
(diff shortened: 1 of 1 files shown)


stof:
Well, the canonical name in composer for that repo is `packagist.org` since a while (probably since the introduction of private packagist running on packagist.com, to make things clearer). The `packagist` name is still supported for backward compatibility, but I think the doc should stay as it is.

alcohol:
It's better to use packagist.org explicitly since the introduction of packagist.com

PR #7441, opened 2018-07-05
Improvements

carusogabriel:
Together with PHPStorm and https://github.com/kalessil/phpinspectionsea, here are some improvements that we can make to Composer :smile:

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -251,7 +251,7 @@ class Installer
                 continue;
             }
 
-            $replacement = (is_string($package->getReplacementPackage()))
+            $replacement = is_string($package->getReplacementPackage())
                 ? 'Use ' . $package->getReplacementPackage() . ' instead'
                 : 'No replacement was suggested';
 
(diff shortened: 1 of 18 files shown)


src/Composer/Installer.php
```diff
@@ -1033,11 +1033,8 @@ private function processDevPackages($localRepo, $pool, $policy, $repositories, $
                         $package->setReplaces($newPackage->getReplaces());
                     }
 
-                    if ($task === 'force-updates' && $newPackage && (
-                        (($newPackage->getSourceReference() && $newPackage->getSourceReference() !== $package->getSourceReference())
-                            || ($newPackage->getDistReference() && $newPackage->getDistReference() !== $package->getDistReference())
-                        )
-                    )) {
+                    if ($task === 'force-updates' && $newPackage && ($newPackage->getSourceReference() && $newPackage->getSourceReference() !== $package->getSourceReference())
```
stof: Are you sure about this change ? I don't think it is right regarding the operator precedence

carusogabriel:
@alcohol Done. I've also squashed the commits. Should be RTM, thank you :smile:

alcohol:
Thanks

src/Composer/Installer.php
stof: my concern is that without parenthesis, I suspect the initial conditions only apply to the left-side of the `||` (which your var-dumps don't test, as they make them always `true`)

Seldaek: Fixed in https://github.com/composer/composer/commit/d5a9d86ee4e934d06073521e630556d4f337629c for readability + clarity, no matter if it worked or not.

Issue #7452, opened 2018-07-09
[Optimization idea] Register conflict rules only based on required packages


stof:
Currently, conflict rules are registered in the solver for all packages matching the constraint (for `replaces` and `conflicts` in the metadata), even when the conflicting package is never required.
In https://blackfire.io/profiles/04234c24-b8c4-43b0-b23a-0a836b8c766e/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes&selected=Composer%5CDependencyResolver%5CRuleSet%3A%3Aadd&callname=Composer%5CDependencyResolver%5CRuleSet%3A%3Aadd (shared in #7449), the `RuleSet::add` method is called 194446 and `Rules2Literal::getHash` is called 193248 times. `Rules2Literal` is used only to represent these conflict rules (preventing 2 packages to be installed at the same time) and `getHash` is called only once when adding them in the set. So this tells us that 99.4% of rules are conflict rules in this case. Reducing the number of these rules might then be a good news for the solver.

These rules can never make composer select a package to install. They are only about rejecting them. And it is common for packages to conflict with packages you won't need in the end (if they were hard dependencies, they would be registered using `require` for the supported versions, not `conflicts` for the unsupported ones). Thus, a conflict constraint will only grow its number of rules over time (as more releases are made) and you cannot really narrow it down (it would not have been a conflict in the first place otherwise).
Thus, this also requires loading metadata for these never used packages (because the metadata are used by the conflict rules)

My idea would be to generate the RuleSet in 2 passes:

- build rules allowing to bring in packages (`require` ones, and the root dev requirements), while registering the list of packages considered in additive way by this process
- process the conflicts/replaces rules of all involved packages, but matching them on the pool of packages created in the first phase instead of matching them on the repository pool.

This way, any conflict constraint would only add rules for packages which can be selected by a positive rule (much less than currently), and we would not load metadata at all for packages never required (as the pool used by the second pass would not be loading more package metadata).

What do you think about this idea ?

stof:
[...]

PR #7454, opened 2018-07-09
 Refactor the handling of conflict rules in the solver 

stof:
This is an implementation of my idea from #7452.

My profiling shows a very small increase in the timing when doing `composer update --dry-run`. I haven't yet run it on a project where the conflict rules might have loaded a bunch of metadata for non-required packages (where we could see an actual win, as the loading of metadata is 3x longer than the SAT solving in my cases).

While it does not improve performance much, it actually fixes https://github.com/composer/composer/issues/3754 without performance regression (the behavior change reported in this issue is the side-effect of an optimization implemented in 1.0-alpha8)


staabm:
I tested this change against the same workload I used to generate the blackfire profile of https://github.com/composer/composer/issues/7452

the resulting profiles dont look "better" then before, no performance improvement.
(just to let you know)

stof:
@staabm yeah, looks like this does not improve performance much. But it fixes the bug introduced by a previous optimization, without performance regression, which is why I still submitted it.

stof:
And there is still one case for which the "conflict" rules are still created in the first pass. It is for the obsolete providers. But I'm not sure this case can be moved safely. And lots of conflict rules seems to be coming from this place.
I will need to discuss with @naderman to see whether this part can be optimized or no.

alcohol:
Could you resolve the conflicts?

stof:
@alcohol done

alcohol:
Not sure if this should target the 1.7 branch or master btw. Is it an improvement purely? Or also a bugfix?

naderman:
@stof thanks, will take some time to really review this for any unintended side effects.

stof:
It fixes a bug introduced in February 2015 when optimizing composer. This implementation manages to keep the same level of performance, without sacrificing correctness (which was the case for the past few years).
The intended side effect is that some existing broken setup related to `replace` rule will now fail properly (like the one reported in the test).

I originally wanted to improve performance with that, but it does not improve it compared to the incorrect implementation (it does compared to the old non-optimized but correct implementation, of course, as it preserves the optimization). So for now, it is more a bugfix than a perf improvement.
Reworking the case of "obsolete providers" might unlock some improvements, but I don't understand that part enough to touch it (as it can both create require rules or conflict rules). I don't know what these "obsolete providers" actually are and the commit introducing this was not very explanatory.

alcohol:
If it is indeed a bugfix, I think it would be better to target the 1.7 branch, cause that makes it easier for @Seldaek to merge it back into master as well.

alcohol:
@stof can you rebase this on top of 1.7 ?

stof:
@alcohol done

naderman:
This looks good to me! Didn't test the performance, but I think functionality wise this makes a lot of sense. As a nice side effect this should also remove unnecessary repository lookups for conflicting packages which are not required anywhere if using a private composer repo. :-)

Seldaek:
And as I mentioned to @naderman a while back.. now that this is in a release, it's breaking things. See https://github.com/composer/composer/issues/7764 - not much we can do except help people out of the mess I guess.

barryvdh:
I ran into this issue also. After searching it does indeed seem that I have a conflict in my requirements and the `replace` entry in a library. The only thing which is a bit confusing is that it doesn't state WHY it cannot resolve those packages:

```
$ composer update

  Problem 1
    - don't install twbs/bootstrap v3.3.0|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.1|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.2|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.4|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.5|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.6|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.7|don't install magento/magento2-base 2.2.6
    - magento/product-community-edition 2.2.6 requires magento/magento2-base 2.2.6 -> satisfiable by magento/magento2-base[2.2.6].
    - Installation request for magento/product-community-edition 2.2.6 -> satisfiable by magento/product-community-edition[2.2.6].
    - Installation request for twbs/bootstrap ^3.3 -> satisfiable by twbs/bootstrap[v3.3.0, v3.3.1, v3.3.2, v3.3.4, v3.3.5, v3.3.6, v3.3.7].
```

Not sure how hard it would be, but stating that `magento/magento2-base 2.2.6` replaces ` twbs/bootstrap 3.1.0` would be helpful in this case.

naderman:
Hey @barryvdh I think that's a great suggestion for improving that error message, can you just paste the same example and your suggestion in https://github.com/composer/composer/pull/7454 for 2.0?

barryvdh:
You mean #7779 I guess? Posted it there. Thanks!

naderman:
Oops, sorry not sure what I did there, yes thanks :-)

PR #7464, opened 2018-07-11
Add `composer i` shorthand for `composer install`

rellect:
In npm I really like this shorthand `npm i`
So thought it may also be convenient in composer

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -115,7 +115,7 @@ resolution.
   requirements and force the installation even if the local machine does not
   fulfill these. See also the [`platform`](06-config.md#platform) config option.
 
-## update
+## update / u
 
 In order to get the latest versions of the dependencies and to update the
 `composer.lock` file, you should use the `update` command. This command is also
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -65,7 +65,7 @@ php composer.phar init
   to a `composer` repository or a JSON string which similar to what the
   [repositories](04-schema.md#repositories) key accepts.
 
-## install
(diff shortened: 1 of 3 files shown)


apfelbox:
:+1:

it currently conflicts, as both `install` and `init` start with the same two letters. But I would argue that `install` is used way more often than `init` so we should optimize the common case.

Same thing please for `composer u` to link to `update` (Symfony Flex adds `unpack` which conflicts with `update`).

Seldaek:
In principle I'm not against this, but I wish it could be done without adding those aliases in the list of commands. We already have a few duplicates in there like dump-autoload/dumpautoload, update/upgrade, clear-cache/clearcache. I'm happy to help people but it's slowly making the help output garbage.. 

Can anyone research if it isn't possible to avoid listing aliases in the list command? Or maybe we do the aliasing in bin/composer before starting the application?

stof:
@Seldaek the Symfony ListCommand does not display aliases anymore in the list of commands (it displays the available aliases in the description of the listed commands itself). But the issue here is that the composer phar ships with symfony/console 2.8 (to make the phar compatible with PHP 5.3) and so does not benefit from changes done in 3.x

Seldaek:
Ok thanks @stof for the update. In that case I'm fine merging this for now and the aliases will vanish in time.

Seldaek:
@rellect would you mind adding `u` for the UpdateCommand too though? For completeness it probably makes sense.

rellect:
@Seldaek Sure, done.

Seldaek:
Thanks

PR #7470, opened 2018-07-14
[minor] SCA

kalessil:
A couple of micro-optimizations and tweaks around hardening security

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -543,16 +543,17 @@ class Installer
 
         foreach ($operations as $operation) {
             // collect suggestions
-            if ('install' === $operation->getJobType()) {
+            $jobType = $operation->getJobType();
+            if ('install' === $jobType) {
                 $this->suggestedPackagesReporter->addSuggestionsFromPackage($operation->getPackage());
             }
 
             // updating, force dev packages' references if they're in root package refs
             if ($this->update) {
                 $package = null;
-                if ('update' === $operation->getJobType()) {
+                if ('update' === $jobType) {
                     $package = $operation->getTargetPackage();
-                } elseif ('install' === $operation->getJobType()) {
+                } elseif ('install' === $jobType) {
                     $package = $operation->getPackage();
                 }
                 if ($package && $package->isDev()) {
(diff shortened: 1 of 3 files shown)


alcohol:
Not sure about the `uniqid` changes. Directory name length for caching can already be an issue sometimes on windows hosts.

kalessil:
Oh, the uniqid changes would add extra characters. Reverted.

src/Composer/Installer.php
```diff
@@ -561,20 +562,24 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
...
                 }
-                if ('update' === $operation->getJobType()
-                    && $operation->getTargetPackage()->isDev()
-                    && $operation->getTargetPackage()->getVersion() === $operation->getInitialPackage()->getVersion()
-                    && (!$operation->getTargetPackage()->getSourceReference() || $operation->getTargetPackage()->getSourceReference() === $operation->getInitialPackage()->getSourceReference())
-                    && (!$operation->getTargetPackage()->getDistReference() || $operation->getTargetPackage()->getDistReference() === $operation->getInitialPackage()->getDistReference())
-                ) {
-                    $this->io->writeError('  - Skipping update of '. $operation->getTargetPackage()->getPrettyName().' to the same reference-locked version', true, IOInterface::DEBUG);
-                    $this->io->writeError('', true, IOInterface::DEBUG);
+                if ('update' === $jobType) {
```
alcohol: I really wonder if this could not be simplified or made mode readable in some manner or form.. this nested `if` collection makes me cringe. Definitely not blaming you, it already was quite horrible. But since you are touching upon this code anyway..

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -75,9 +75,9 @@ public function getPackageFilename(PackageInterface $package)
         $nameParts = array(preg_replace('#[^a-z0-9-_]#i', '-', $package->getName()));
 
         if (preg_match('{^[a-f0-9]{40}$}', $package->getDistReference())) {
-            $nameParts = array_merge($nameParts, array($package->getDistReference(), $package->getDistType()));
```
alcohol: opcode wise this doesn't seem to differ at all, unless I am mistaken?

It is more readable though, so I'm OK with it.

src/Composer/Installer.php
kalessil: I'll check if the control flow can be simplified a little bit more ;). Since I don't have much know-how about the project, I'll stick to obvious simplifications.

src/Composer/Package/Archiver/ArchiveManager.php
kalessil: We are reducing number of memory allocations here. It was: construct array of 2 elements, merge arrays(CPU/memory intensive operation), now it pushes 2 items directly into the original array.

alcohol: I looked at the opcodes, and internally they are no different AFAIK? But like I said, it is more readable so fine with me.

src/Composer/Installer.php
stof: the thing is, I'm not sure this actually makes the code simpler to read (and in terms of cyclomatic complexity, it is exactly the same, as you still have all the same conditions being checked)

kalessil: Original intention is to reduce repetitive methods calls, but the code block is indeed needs simplification. I'll keep in mind CCM, it would be great to reduce it as well.

Seldaek:
Thanks

Issue #7463, opened 2018-07-11
Authentication with auth.json ignored with Mercurial


Elendev:
My `composer.json`:

```json
{
    "repositories": [
        {
            "type": "hg",
            "url": "https://my.private.repository.com/my-package"
        }
    ],
}
```
My `auth.json`:

```json
{
    "http-basic": {
      "my.private.repository.com": {
        "username": "myusername",
        "password": "secret"
      }
    }
}
```

Output of `composer diagnose`:

```
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking HTTP proxy: OK
Checking HTTP proxy support for request_fulluri: OK
Checking HTTPS proxy support for request_fulluri: OK
Checking github.com rate limit: OK
Checking disk free space: OK
```

When I run this command:

```
composer update
```

I get the following output:

```
[RuntimeException]                                                                                                    
  Failed to clone https://my.private.repository.com/my-package, could not read packages from it  
                                                                                                                        
  abort: http authorization required for https://my.private.repository.com/my-package      
```

And I expected this to happen:

I expect Composer to be able to use a Mercurial repository behind a http-basic authentication mechanism with the `auth.json` file.

If I use this:
```json
{
    "repositories": [
        {
            "type": "hg",
            "url": "https://myusername:secret@my.private.repository.com/my-package"
        }
    ],
}
```
I'm able to use my private package, but as soon as I'm using an `auth.json` file, the credentials are not used.
The `hg` command executed is the following: `hg clone --noupdate 'https://my.private.repository.com/my-package' '/root/.composer/cache/vcs/https---my-private-repository-com-my-package/'`.

alcohol:
It would help if you could provide us with a reproducible scenario/case. I do not have any hosted http-basic auth protected HG repositories, nor do I have the time to create one. So there is nothing I can do to debug this presently.

Elendev:
@alcohol You are right, as soon as I have time, I will try to create a publicly available password-protected mercurial repository to be able to easily reproduce the issue.
In the meantime, I'll try to solve it through a PR since I have this configuration available in my environment.

alcohol:
:+1:

Seldaek:
Closed by https://github.com/composer/composer/pull/7475

PR #7475, opened 2018-07-17
Add support for authentication with mercurial repositories.

Elendev:
This PR aim to resolve the issue #7463.

I've added a `Composer\Util\Hg` class similar to the `Composer\Util\Git` class. The purpose of this class is to add the credentials (username and password) at the beginning of the mercurial's URL to allow authentication with the `auth.json` file.
An URL like `https://my-domain.com/my-repo` will be transformed to `https://my-username:my-password@my-domain.com/my-repo` if authentication is required.
The `Composer\Util\Hg` class works in the same way as the `Composer\Util\Git`, providing a `runCommand` method that handle the potential URL modification.

### DIFF
--- a/src/Composer/Util/Hg.php
+++ b/src/Composer/Util/Hg.php
@@ -56,7 +56,7 @@ class Hg
         // Try with the authentication informations available
         if (preg_match('{^(https?)://((.+)(?:\:(.+))?@)?([^/]+)(/.*)?}mi', $url, $match) && $this->io->hasAuthentication($match[5])) {
             $auth = $this->io->getAuthentication($match[5]);
-            $authenticatedUrl = $match[1] . '://' . rawurldecode($auth['username']) . ':' . rawurlencode($auth['password']) . '@' . $match[5] . (!empty($match[6])? $match[6]: null);
+            $authenticatedUrl = $match[1] . '://' . rawurlencode($auth['username']) . ':' . rawurlencode($auth['password']) . '@' . $match[5] . (!empty($match[6])? $match[6]: null);
 
             $command = call_user_func($commandCallable, $authenticatedUrl);
 
--- a/src/Composer/Util/Hg.php
+++ b/src/Composer/Util/Hg.php
@@ -1,0 +1,91 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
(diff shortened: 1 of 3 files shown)


src/Composer/Util/Hg.php
```diff
@@ -0,0 +1,91 @@
...
+        $command = call_user_func($commandCallable, $url);
+
+        if (0 === $this->process->execute($command, $ignoredOutput, $cwd)){
+            return;
+        }
+
+        // Try with the authentication informations available
+        if (preg_match('{^(https?)://((.+)(?:\:(.+))?@)?([^/]+)(/.*)?}mi', $url, $match) && $this->io->hasAuthentication($match[5])) {
+            $auth = $this->io->getAuthentication($match[5]);
+            $authenticatedUrl = $match[1] . '://' . rawurldecode($auth['username']) . ':' . rawurlencode($auth['password']) . '@' . $match[5] . (!empty($match[6])? $match[6]: null);
```
Seldaek: It seems odd that you're doing rawurl**decode** on the username. Should that not also be an encode?

src/Composer/Util/Hg.php
```diff
@@ -0,0 +1,91 @@
...
+        }, $message);
+    }
+
+    private function throwException($message, $url)
+    {
+        if (0 !== $this->process->execute('hg --version', $ignoredOutput)) {
+            throw new \RuntimeException(self::sanitizeUrl('Failed to clone ' . $url . ', hg was not found, check that it is installed and in your PATH env.' . "\n\n" . $this->process->getErrorOutput()));
+        }
+
+        throw new \RuntimeException(self::sanitizeUrl($message));
```
stof: this is not including the process output anymore, which means you don't see the hg error anymore (which can be very useful to understand why it failed and fix it)

src/Composer/Util/Hg.php
Elendev: You are totally right, nice catch, thank you ! An autocomplete mistake :+1:

src/Composer/Util/Hg.php
Elendev: You are right. The git version is not so verbose, but as soon as the feature was here before, I'll keep it here

Seldaek:
Thanks

PR #7501, opened 2018-07-29
Do not dump source and dist for metapackages

kassner:
I stumbled upon composer/satis#439, did some tests and it looks like nothing breaks when removing `source` and `dist`. Anything else that this could go wrong?


alcohol:
Not sure to be honest.

stof:
They are indeed not used, as the metapackage does not get downloaded. So this should be fine.

alcohol:
Thanks

PR #7511, opened 2018-08-02
Don't unescape surrogate pairs in fallback JSON formatter

kamazee:
The proposed solution to having broken UTF-8 out of the fallback formatter is to skip surrogate pairs rather than trying to handle them correctly. If we were to implement "The Proper" solution, which is basically to port JSON unescaping to PHP and put into formatter, it would require at least looking ahead/behind (because a pair of 2 escape sequences in 0xD800..0xDFFF matters) and handling errors in case of unpaired surrogates; that seems to be an overkill for workaround for PHP 5.3 which is dead anyways.

### DIFF
--- a/src/Composer/Json/JsonFormatter.php
+++ b/src/Composer/Json/JsonFormatter.php
@@ -69,6 +69,13 @@ class JsonFormatter
                         $l = strlen($match[1]);
 
                         if ($l % 2) {
+                            $code = hexdec($match[2]);
+                            // 0xD800..0xDFFF denotes UTF-16 surrogate pair which won't be unescaped
+                            // see https://github.com/composer/composer/issues/7510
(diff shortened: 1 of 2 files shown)


kamazee:
Note: test fails everywhere but the method under test is really only called for PHP < 5.4.

kamazee:
@Seldaek it's ready for review. Besides the unit test, I checked if it worked for the case I had with installing nette/tester on PHP 5.3: it did 🎉

Seldaek:
Thanks!

Seldaek:
Merged in 1.7/master as https://github.com/composer/composer/commit/5a22a4f1f31a44c41a518764dc9e49b113431400 and https://github.com/composer/composer/commit/e1a6bd5ff1bf2b73cd7a6b2fd029641d4a505147

PR #7521, opened 2018-08-05
Add interactive option to install dependencies after running init command

shalvah:
When the `init` command is run in interactive mode, the dev will be asked if they wish to install dependencies straightaway (only if they specified any dependencies). It's a minor improvement that makes life a tad easier. 😀

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -147,7 +147,7 @@ EOT
             }
         }
 
-        $question = 'Would you like to install dependencies now? [<comment>yes</comment>]? ';
+        $question = 'Would you like to install dependencies now [<comment>yes</comment>]? ';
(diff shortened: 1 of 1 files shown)


alcohol:
If installing failed, wouldn't it be useful to let the user know why?

shalvah:
@alcohol Not really. It's not really the responsibility of the `init` command. It's done its major job. Installing dependencies is just a nice add-on. If something is wrong, the user can then run `install` to know the exact problem. Plus, I feel it's important for `init` to always end on a positive note. 🙂

sbuzonas:
What about supporting install configuration that isn't easily changed like `--prefer-source` or `--prefer-dist`

davidjeddy:
I could see something like `composer init --install-deps` flagging that the deps should be installed during the process. Each dep the dev. adds is installed, output displayed, then back to defining the next package.

Seldaek:
Thanks

PR #7534, opened 2018-08-08
Use target directory instead of CWD for hg pull with auth.json

dnlk17:
There are 2 missing parameters in the second try of the process-execution. (in case of authentication / use of auth.json)
If a package is updated, the `composer update` results in a `abort: repository is unrelated` because it is executed in the current working directory (CWD) instead of the install-dir.

I've added these 2 parameters to the `process->execute()` call.

### DIFF
--- a/src/Composer/Util/Hg.php
+++ b/src/Composer/Util/Hg.php
@@ -60,7 +60,7 @@ class Hg
 
             $command = call_user_func($commandCallable, $authenticatedUrl);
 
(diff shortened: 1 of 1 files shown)


Seldaek:
@alcohol please don't merge bugfixes to master 'cause I then have to cherrypick them in 1.7 and it makes an even bigger mess out of the git history

alcohol:
Will keep that in mind.

PR #7549, opened 2018-08-14
Update AutoloadGenerator.php

voyula:
performance optimized


stof:
That change is wrong, as `autoload.php` is precisely meant to be require-able as many time as you want (to get access to the autoloader instance).

PR #7553, opened 2018-08-14
Performance Improved

voyula:
fixed unnecessary 'dirname's


stof:
Composer supports PHP 5.x, so that's not an option

PR #7555, opened 2018-08-15
Revert composer.json changes when require process stops

pierredup:
When you run `require some/package` and the process stops, the `composer.json` file doesn't get reverted. This PR adds a signal handler that captures when a process stops unexpectedly (E.G pressing ^C) and reverts the `composer.json` file


Seldaek:
Thanks

Issue #7559, opened 2018-08-18
Reconsider recommendation to not commit vendor


devsnacks:
The [general recommendation](https://getcomposer.org/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md) is to not commit the dependencies in the vendor directory. 

**My proposal is to withdraw this general recommendation and instead, give a more comprehensive explanation about the different use cases and pros and cons of each approach.**

The reasons given are very valid in a **distribution** context. 
Use case: I want to share a library others can use.

In a **deployment** context there may be several disadvantages in not committing the vendor directory:

**Lack of traceability**:
Changes in dependencies are not easily traceable anymore.
Failure analysis is made more difficult.

**Lack of reproducibility**
The additional build step to pull the dependencies increases the fragility of the build pipeline.
 
**Decreasing build performance**
Installing all dependencies on every build can have high impact on build performance.

Especially in enterprise context, the mentioned problems are critical.

Do you basically agree to
a) provide more information about the deployment use case and the mentioned downsides?
b) remove the general recommendation to not commit vendor?

localheinz:
@devsnacks 

How about opening a PR with your proposal instead of an issue requesting someone else do the work?

Personally, I would never recommend to commit the `vendor` directory, though. All of the issues you mentioned can be solved differently:

- committing `composer.lock`
- id not committing `composer.lock`, proper dependency management (hint: semantic versioning)
- caching composer’s cache in build and deploy systems

Seldaek:
a) Maybe some reasonable amount of additional docs
b) Definitely not. The general recommendation is to use the composer.lock and gitignore the vendor. That's applicable to most people and use cases. There ways to mitigate all the problems you mention that don't cause the huge mess of having all dependencies' code tracked in your git repo.

devsnacks:
@localheinz 

thank you for your comment. 

It´s not about requesting someone else to just change the documentation.
My intent was so start a discussion on the topic, as this is about a general advice/best practice
which i guess needs more conversation in the first place to find an agreement on it.
Of course i can provide a PR, if there is basically consensus on it.

> commiting `composer.lock`

We already do this. Nonetheless, this does not solve the mentioned problems:
It has no impact on _traceability_, as you are still not able to have a single diff showing
all changes of the commits in a specific release.

Also it has just minimal impact on _reproducibility_, as it just ensures that the pulled versions are consistent, which i would consider as an absolute minimum level of reproducibility.

[...]

PR #7563, opened 2018-08-19
Added use cases for committing vendor

seb-h-k:
Regarding #7559 i´ve added some use cases where committing vendor would be preferable.

### DIFF
--- a/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
+++ b/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
@@ -16,6 +16,16 @@ problems:
   submodules. This is problematic because they are not real submodules, and you
   will run into issues.
 
+Committing the dependencies might be an option if you:
+
+- have very high demands on reproducibility. Committing vendor makes you
+  independent of upstream changes (e.g. renames, deletes, history overwrites)
+  and makes you´re build process resilient against outages of upstream servers.
+- require a single diff with all changes, including upstream changes.
+  This could make failure analysis and traceability easier.
+- want to reduce your build times
+  (and other options like caching do not work well in your environment)
+
 If you really feel like you must do this, you have a few options:
 
 1. Limit yourself to installing tagged releases (no dev versions), so that you


doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
```diff
@@ -16,6 +16,16 @@ problems:
   submodules. This is problematic because they are not real submodules, and you
   will run into issues.
 
+Committing the dependencies might be an option if you:
+
+- have very high demands on reproducibility. Committing vendor makes you
+  independent of upstream changes (e.g. renames, deletes, history overwrites)
+  and makes you´re build process resilient against outages of upstream servers.
+- require a single diff with all changes, including upstream changes.
```
stof: the drawback being of course that any `composer update` leads to a huge diff

doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
```diff
@@ -16,6 +16,16 @@ problems:
   submodules. This is problematic because they are not real submodules, and you
   will run into issues.
 
+Committing the dependencies might be an option if you:
+
+- have very high demands on reproducibility. Committing vendor makes you
+  independent of upstream changes (e.g. renames, deletes, history overwrites)
```
stof: the lock file locks a specific commit. So history overwrites are not an issue regarding the reproducibility of `composer install` (unless the upstream does not allow accessing the code of the old commit at all anymore after that)

devsnacks: You´re right, history overwrite is not quite suitable in the context of reproducibility then. It´s still true for renaming/deletes of the respective cvs-sources? Which also means you can´t reliably rebuild a specific commit in history?

doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
devsnacks: Depends on the number of dependencies and update frequency. In terms of QA and traceability i don´t see a difference between application changes and library changes. Being able to quickly analyse the complete changeset can be more important than having a neat - but incomplete - diff.

doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
stof: well, for renaming, it all depends whether the old URL still works. For instance, when renaming a repo on Github, the old URL redirects to the new one (until you recreate a new repo reusing the old name), so composer can still download the code. If you rename without redirections, that's actually a case of delete. And for these, composer cannot do anything indeed (if the code is not available for download anymore, downloading will fail)

doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
```diff
@@ -16,6 +16,16 @@ problems:
...
+  and makes you´re build process resilient against outages of upstream servers.
```
Seldaek: s/you're/your/

Seldaek:
Would be nice to add https://github.com/composer/composer/issues/7559#issuecomment-417465798 as well for completeness..

doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
naderman: This also entirely depends on what the source of your packages is. If you simply fork all your dependencies or use a Composer repository of your own, you are safe from renames, deletes, history overwrites or anything else without commiting your vendor folder. So I think this advise is only correct under specific circumstances.

doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
naderman: You could make your build process resilient against outages by having your own Composer repository or hosting all packages yourself, without commiting the vendor directory. I think it needs to be really clear that all these things can be achieved without commit your vendor directory.

doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md
```diff
@@ -16,6 +16,16 @@ problems:
...
+- require a single diff with all changes, including upstream changes.
+  This could make failure analysis and traceability easier.
+- want to reduce your build times
+  (and other options like caching do not work well in your environment)
```
naderman: This may actually increase build times if the third party dependencies you include in your vendor directory slow down VCS operations (e.g. containing large binary files in history), while composer cache be more easily cached with standard means in most CI systems

naderman:
Considering that there are other solutions for each of these listed advantages, I'm not sure if we wouldn't be better of just leaving the current phrasing "If you really feel like you must do this, " because it's not like we can really in all completeless evaluate each developers preferences in one article.

devsnacks:
@naderman 
Agree. I see that most of you have a clear opinion on this and i agree that there are many advantages in the recommended way of doing it as well. My main point was to open up a discussion on it and there was great contribution and maybe this discussion could help other people in finding the right way in their context. Thank you all for your contributions!

PR #7565, opened 2018-08-20
add removePackage() to RepositoryInterface

alcohol:
This seems to be implemented by all repositories, but not part of any interface. Since both the `Installer` and `CompositeRepository` depend on it, it should probably be a part of the interface.


alcohol:
Technically, yeah. But in practice, I doubt it.

stof:
and as the code relies on the methods, the implementations not having this method are probably already broken anyway.

alcohol:
Yeah that was my point too. I cannot think of any useful implementations that do not implement this method already.

Seldaek:
Thanks

Issue #7560, opened 2018-08-18
[tool/feature] update to latest available version


strarsis:
Is it possible to let composer bump all versions in `composer.json` to their latest versions?
After bumping to latest version, tests are run to check if the packages are still compatible.
Having a feature/tool for automatically bumping all versions in `composer.json` (not in the lockfile) would be helpful.

alcohol:
I have personally been looking into this lately a bit myself, but haven't really invested the time yet to actually submit a PR.

strarsis:
@alcohol: This feature would be very helpful!

strarsis:
@alcohol: Is this still planned to be implemented in near future?
It is very useful to bulk-update WordPress plugins managed by composer for example.

alcohol:
No explicit plans, on my personal to-do, but really low on time to be honest. :-(

Seldaek:
Duplicate of https://github.com/composer/composer/issues/7273

PR #7567, opened 2018-08-20
extract command logic to helper ( prep for #7560 )

alcohol:
This work is a start towards #7560 

- [x] extract logic
- [ ] write test scenarios

### DIFF
--- a/src/Composer/Util/Exception/PackageHelper/NoMatchForPhpVersionException.php
+++ b/src/Composer/Util/Exception/PackageHelper/NoMatchForPhpVersionException.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Composer\Util\Exception\PackageHelper;
+
+class NoMatchForPhpVersionException extends \InvalidArgumentException
+{
+    public static function given($name, $phpVersion)
+    {
+        return new self(sprintf(
+            'Could not find package %s in any version matching your PHP version (%s)',
+            $name,
+            $phpVersion
+        ));
+    }
+}
--- a/src/Composer/Util/PackageHelper.php
+++ b/src/Composer/Util/PackageHelper.php
@@ -1,0 +1,156 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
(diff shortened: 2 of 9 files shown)


src/Composer/Util/Exception/PackageHelper/NoMatchForPhpVersionException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Composer\Util\Exception\PackageHelper;
+
+class NoMatchForPhpVersionException extends \InvalidArgumentException
+{
+    public static function given($name, $phpVersion)
```
Seldaek: Not sure I see the value of those given() methods vs using a constructor (which would let you pass a $previousException too for example)

src/Composer/Util/PackageHelper.php
```diff
@@ -37,6 +37,8 @@ public function findSimilar($package, $limit = 5)
             return array();
         }
 
+        var_dump($results);
```
staabm: Debug leftover

alcohol: Yeah, it is not finished yet. Still got some scenarios I can't quite put my finger on.

src/Composer/Util/Exception/PackageHelper/NoMatchForPhpVersionException.php
alcohol: Wow I kind of forgot about this PR. I can refactor the exceptions to use constructors. But I'm not sure if this whole PR is still applicable with all the work you are putting into 2.0? If it is, then just let me know and I'll update accordingly with the comments left so far :-)

Seldaek: I thiiink it'd be still useful but not 100% sure, probably would be best to rebase and retarget it for 2.0 branch though.

alcohol: Alright, will pick it up again in the near future.

Seldaek:
Closing as it's got quite outdated at this point. If there is still an interest feel free to reopen, I don't think it's necessarily bad to refactor commands, but I think maybe working on integration tests for them first would make more sense so we can ensure we avoid regressions when doing such big refactorings.

PR #7570, opened 2018-08-20
Failing test case for Composer 1.7 autoloader regression for replacing packages

naderman:
If you have a package in your composer.lock which is not directly "require"d anywhere, but replaces a package name which is required by some of your packages, the autoloader will no longer contain code for this package and all its dependencies.

The case where this particular bug can be found is with existing yii2 lock files. These seem to sometimes contain the "yii2-dev" package which replaces "yii2", even though it wasn't directly required anywhere in the lock file. See this example: https://github.com/heroku/heroku-buildpack-php/issues/294

The solution can be to fix the introduced non-dev autoloading exclusion mechanism (https://github.com/composer/composer/commit/eedbd218f52c526ec41c6c623137c3781eb4d928#diff-6f3360cad3227d2e21169d372b9ad4f5 which I think may be at fault here). This will get us back to previous behaviour.

However the real question I have is, how can this happen in the first place. Composer is not supposed to install any packages which are not required by name somewhere.

### DIFF
--- a/tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
+++ b/tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
@@ -419,6 +419,39 @@ class AutoloadGeneratorTest extends TestCase
         $this->assertFileExists($this->vendorDir.'/composer/autoload_classmap.php', "ClassMap file needs to be generated, even if empty.");
     }
 
+    public function testNonDevAutoloadExclusionWithRecursionReplace()
+    {
+        $package = new Package('a', '1.0', '1.0');
+        $package->setRequires(array(
+            new Link('a', 'a/a'),
+        ));
+
+        $packages = array();
+        $packages[] = $a = new Package('a/a', '1.0', '1.0');
+        $packages[] = $b = new Package('b/b', '1.0', '1.0');
+        $a->setAutoload(array('psr-0' => array('A' => 'src/', 'A\\B' => 'lib/')));
+        $a->setRequires(array(
+            new Link('a/a', 'c/c'),
+        ));
+        $b->setAutoload(array('psr-0' => array('B\\Sub\\Name' => 'src/')));
+        $b->setReplaces(array(
+            new Link('b/b', 'c/c'),
+        ));
+
+        $this->repository->expects($this->once())
+            ->method('getCanonicalPackages')
+            ->will($this->returnValue($packages));
+
+        $this->fs->ensureDirectoryExists($this->vendorDir.'/composer');
+        $this->fs->ensureDirectoryExists($this->vendorDir.'/a/a/src');
+        $this->fs->ensureDirectoryExists($this->vendorDir.'/a/a/lib');
+        $this->fs->ensureDirectoryExists($this->vendorDir.'/b/b/src');
+
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', false, '_5');
+        $this->assertAutoloadFiles('vendors', $this->vendorDir.'/composer');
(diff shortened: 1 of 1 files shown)


naderman:
I'm leaning toward saying that this test should actually fail, and that we need to figure out & fix how this situation ever occured in the lock file here: https://github.com/heroku/heroku-buildpack-php/issues/294

lock file & composer.json: https://gist.github.com/boboldehampsink/379ffff4c3f5d6db4d674af17dac3cfc /

naderman:
So I'm leaning toward just closing this, since I think the cause is really just a broken lock file which gets fixed if one runs an update on it. However the core question here to me remain (1) how does one generate a lock file with a file which is not required directly (2) should we be detecting these and warning people?

I've seen this independently from 2 people in one day today, and both cases they just couldn't work out what the problem was, so at least some improved error handling for this case is probably a good idea.

dzuelke:
A start (for debugging purposes) would be to write the Composer version to the lock file, together maybe with other metadata for info (+/- the command args?); future releases could then read the lock file and throw an error on `composer install`, instructing users to re-generate it.

boboldehampsink:
I have run `composer update` multiple times but that didn't fix it before. Removing the lockfile and the vendor folder and running `composer install` did it for me.

Seldaek:
@naderman I can confirm what @boboldehampsink says, running an update from the broken state doesn't fix anything. If you have a dangling package installed it'd be removed but if it replaces something that is required then I guess it is seen as needed even though it's not required itself. This is the same issue as https://github.com/composer/composer/issues/7573 as I understand it.

What is odd is I can not reproduce it in an isolated test though, so I don't think we understand the whole thing yet..

In the test, `dependency` is equivalent to `yiisoft/yii2` and `extra-replacer` is `yiisoft/yii2-dev`.

```
--TEST--
Update from an installed state with an extraneous replacer package in vendor dir
--COMPOSER--
{
    "repositories": [
        {
            "type": "package",
            "package": [
                { "name": "whitelisted", "version": "1.1.0", "requires": { "dependency": "1.0.0" } },
                { "name": "extra-replacer", "version": "1.0.0" },
                { "name": "dependency", "version": "1.0.0" }
            ]
        }
    ],
    "require": {
        "whitelisted": "1.*"
    }
}
--LOCK--
{
    "packages": [
        { "name": "whitelisted", "version": "1.1.0" },
        { "name": "extra-replacer", "version": "1.0.0", "replaces": { "dependency": "*" } }
    ],
    "packages-dev": null,
    "aliases": [],
    "minimum-stability": "dev",
    "stability-flags": [],
    "prefer-stable": false
}
--INSTALLED--
[
    { "name": "whitelisted", "version": "1.1.0" },
    { "name": "extra-replacer", "version": "1.0.0", "replaces": { "dependency": "*" } }
]
--RUN--
update
--EXPECT--
Uninstalling extra-replacer (1.0.0)
Installing dependency (1.0.0)
```

It fails with:

```
--- Expected
+++ Actual
@@ @@
 Uninstalling extra-replacer (1.0.0)
-Installing dependency (1.0.0)
```

So it removes the replacer (yii2-dev) but does not install dependency (yii2).. While with the real example code it fails to do anything at all.

naderman:
I'm curious on how far https://github.com/composer/composer/pull/7454 affects this change.

@boboldehampsink do you still have the vendor for around, so you could try if a composer update with the latest composer master fixes your lock file without having to delete the vendor directory?

boboldehampsink:
@naderman no sorry, to fix it I have removed it already.

Seldaek:
Merged in as 386382503 in 1.8/master now that https://github.com/composer/composer/pull/7900 fixed this issue.

PR #7574, opened 2018-08-21
Remove github API request retries as it is not needed anymore

seferov:
Returning null inside the catch doesn't have a meaning since it makes not found retry thing useless. Either not found retry or return null should be removed.

### DIFF
--- a/src/Composer/Repository/Vcs/GitHubDriver.php
+++ b/src/Composer/Repository/Vcs/GitHubDriver.php
@@ -183,30 +183,13 @@ class GitHubDriver extends VcsDriver
             return $this->gitDriver->getFileContent($file, $identifier);
         }
 
-        $notFoundRetries = 2;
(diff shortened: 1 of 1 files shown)


stof:
I support removing the not-found-tries instead (there is a TODO for that in the code), as it is broken since https://github.com/composer/composer/pull/5862 was merged, while the retry was added in 2013 (in https://github.com/composer/composer/commit/5267bafa2cb28e564dbca8d1bfaa64de2a120827) because the github API was unreliable at that time (randomly failing with 404)

alcohol:
@Seldaek should something like this also target the 1.7 branch, or does this belong in master only?

Seldaek:
Could go in 1.7 I guess but it's mostly a code cleanup so it's not a big deal either way IMO

Seldaek:
Thanks

Issue #7546, opened 2018-08-13
Composer fail to update dependencies which are using `dev-master`.


shulard:
My `composer.json`:

```json
{
    "type": "project",
    "license": "proprietary",
    "require": {
        "php": "^7.2.0",
        "ext-iconv": "*",
        "enqueue/messenger-adapter": "^0.1.0",
        "enqueue/redis": "^0.8.23",
        "league/flysystem": "^1.0",
        "league/fractal": "^0.17.0",
        "nelmio/alice": "^3.4",
        "php-http/httplug-pack": "^1.1",
        "predis/predis": "^1",
        "sensio/framework-extra-bundle": "^5.2@dev",
        "symfony/asset": "^4.1",
        "symfony/console": "^4.1",
        "symfony/debug-pack": "^1.0",
        "symfony/dotenv": "^4.1",
        "symfony/expression-language": "^4.1",
        "symfony/flex": "^1.0",
        "symfony/form": "^4.1",
        "symfony/framework-bundle": "^4.1",
        "symfony/messenger": "^4.1",
        "symfony/orm-pack": "^1.0",
        "symfony/security-bundle": "^4.1",
        "symfony/serializer-pack": "^1.0",
        "symfony/translation": "^4.1",
        "symfony/twig-bundle": "^4.1",
        "symfony/validator": "^4.1",
        "symfony/web-link": "^4.1",
        "symfony/webpack-encore-pack": "^1.0",
        "symfony/yaml": "^4.1"
    },
    "require-dev": {
        "phpspec/phpspec": "^5",
        "roave/security-advisories": "dev-master",
        "squizlabs/php_codesniffer": "^3.2",
        "symfony/browser-kit": "^4.1",
        "symfony/phpunit-bridge": "^4.1",
        "symfony/process": "^4.1",
        "symfony/property-access": "^4.1",
        "symfony/web-server-bundle": "^4.1"
    },
    "config": {
        "bin-dir": "bin",
        "preferred-install": {
            "*": "dist"
        },
        "sort-packages": true
    },
    "autoload": {
        "psr-4": {
            "App\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "App\\Tests\\": "tests/"
        }
    },
    "replace": {
        "symfony/polyfill-iconv": "*",
        "symfony/polyfill-php71": "*",
        "symfony/polyfill-php70": "*",
        "symfony/polyfill-php56": "*"
    },
    "scripts": {
        "auto-scripts": {
[...]

PR #7575, opened 2018-08-21
Metapackage: Don't rely on sourceReference property to avoid invalid comparison.

shulard:
Since the metapackage don't store `sourceReference` anymore in the `installed.json` file
we can't use it in the `getFullPrettyVersion` to compare version.

That particular method is used in the outdated comparison algorithm (see #7546 for more details).

### DIFF
--- a/src/Composer/Package/BasePackage.php
+++ b/src/Composer/Package/BasePackage.php
@@ -212,7 +212,10 @@ abstract class BasePackage implements PackageInterface
      */
     public function getFullPrettyVersion($truncate = true)
     {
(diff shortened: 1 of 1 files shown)


sbuzonas:
This may have undesirable side-effects. See [comment](https://github.com/composer/composer/issues/7546#issuecomment-414727309) on #7546 

We need a way to know when it is outdated. IMO, current behavior of always treating it as stale is safer than always treating it as up to date.

Seldaek:
Closing as per https://github.com/composer/composer/issues/7546#issuecomment-414941999

Issue #7577, opened 2018-08-21
Add extra.symfony.require into composer core.


rubenrua:
The Symfony project has too many tags into packagist. This excess of tags produces big memory and CPU usage.

See original patch from @nicolas-grekas into symfony/flex: https://github.com/symfony/flex/pull/378

I am coding a generic composer plugin with the Flex patch for generic projects with Symfony dependencies.

A *fast* test with PHP 5.6 (sorry)

| | Internal big project | Sylius/Sylius-Standard | laravel/laravel
| ----- | ----- | ---- | --- |
| Before | 2820.6MB / 171.19s | 2961.1MB / 197.20s | 617.1MB / 28.76s
| After | 430.1MB / 30.75s | 416.0MB / 22.04s | 314.7MB / 18.07s

After several iterations, Symfony Flex is going to add a new extra parameter `'extra.symfony.require'` (see https://github.com/symfony/flex/pull/409 and https://github.com/symfony/skeleton/pull/93)

IMHO adding this extra parameter to the composer core could be a good idea. Symfony components are very used in PHP. I created this issue to discuss about this idea.

rubenrua:
Another *fast* test with PHP 7.2 

See https://github.com/rubenrua/composer/tree/clear-tags

```
$ php -v
PHP 7.2.1-1ubuntu2 (cli) (built: Jan 31 2018 18:36:35) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.2.1-1ubuntu2, Copyright (c) 1999-2017, by Zend Technologies
$ php -d memory_limit=-1 ../rubenrua/bin/composer update --profile --ignore-platform-reqs --dry-run
```

| | Internal big project | [Sylius/Sylius-Standard](https://github.com/Sylius/Sylius-Standard) | [laravel/laravel](https://github.com/laravel/laravel)
| ----- | ----- | ---- | --- |
| Before | Memory: 337.9MB (peak: 1582.09MB), time: 31.84s|  Memory: 384.84MB (peak: 1670.44MB), time: 28.11s | Memory: 265.09MB (peak: 417.44MB), time: 6.57s
| After | Memory: 183.05MB (peak: 286.56MB), time: 11.04s | Memory: 218.76MB (peak: 251.73MB), time: 5.02s| Memory: 210.17MB (peak: 236.37MB), time: 4.38s

[...]

PR #7579, opened 2018-08-23
[WIP] #7577 Add extra.symfony.require into composer core

rubenrua:
WIP. Fix Add extra.symfony.require into composer core. #7577

Only for CI.

### DIFF
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -299,6 +299,12 @@ class Factory
             $localConfig = $file->read();
         }
 
+        if (isset($localConfig['extra']['symfony']['require'])){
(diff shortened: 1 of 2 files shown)


src/Composer/Factory.php
```diff
@@ -299,6 +299,12 @@ public function createComposer(IOInterface $io, $localConfig = null, $disablePlu
             $localConfig = $file->read();
         }
 
+        if (isset($localConfig['extra']['symfony']['require'])){
```
staabm: imo any solution which explicitly benefits only a single framework cannot be merged into composer itself.

PR #7597, opened 2018-08-29
Updated ArchiveManager class to skip realpath on urls for the targetDir

rieves:
For the issue #7598

### DIFF
--- a/src/Composer/Package/Archiver/ArchiveManager.php
+++ b/src/Composer/Package/Archiver/ArchiveManager.php
@@ -133,13 +133,7 @@ class ArchiveManager
 
         // Archive filename
         $filesystem->ensureDirectoryExists($targetDir);
-
-        if(filter_var($targetDir, FILTER_VALIDATE_URL)) {
-            $targetBase = $targetDir;
-        } else {
-            $targetBase = realpath($targetDir);
-        }
-
+        $targetBase = filter_var($targetDir, FILTER_VALIDATE_URL) ? $targetDir : realpath($targetDir);
         $target = $targetBase.'/'.$packageName.'.'.$format;
         $filesystem->ensureDirectoryExists(dirname($target));
 
--- a/src/Composer/Package/Archiver/ArchiveManager.php
+++ b/src/Composer/Package/Archiver/ArchiveManager.php
@@ -133,7 +133,14 @@ class ArchiveManager
 
         // Archive filename
         $filesystem->ensureDirectoryExists($targetDir);
-        $target = realpath($targetDir).'/'.$packageName.'.'.$format;
+
+        if(filter_var($targetDir, FILTER_VALIDATE_URL)) {
+            $targetBase = $targetDir;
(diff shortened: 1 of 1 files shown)


src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -133,7 +133,14 @@ public function archive(PackageInterface $package, $format, $targetDir, $fileNam
 
         // Archive filename
         $filesystem->ensureDirectoryExists($targetDir);
-        $target = realpath($targetDir).'/'.$packageName.'.'.$format;
+
+        if(filter_var($targetDir, FILTER_VALIDATE_URL)) {
```
ronfroy: in my opinion a ternary is more readable

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -133,7 +133,8 @@ public function archive(PackageInterface $package, $format, $targetDir, $fileNam
 
         // Archive filename
         $filesystem->ensureDirectoryExists($targetDir);
```
sbuzonas: This line no longer makes sense if `$targetDir` isn't always a directory any longer.

sbuzonas: Perhaps nest it in a block with the url check instead of the ternary.

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -133,7 +133,8 @@ public function archive(PackageInterface $package, $format, $targetDir, $fileNam
...
-        $target = realpath($targetDir).'/'.$packageName.'.'.$format;
+        $targetBase = filter_var($targetDir, FILTER_VALIDATE_URL) ? $targetDir : realpath($targetDir);
+        $target = $targetBase.'/'.$packageName.'.'.$format;
         $filesystem->ensureDirectoryExists(dirname($target));
```
sbuzonas: Same with this line.

src/Composer/Package/Archiver/ArchiveManager.php
sbuzonas: Second thought, the stream wrapper should end up raising an appropriate exception if the stream context does not support `streamWrapper::mkdir`

sbuzonas:
Although the change appears simple, the added complexity is not. Can it be tested?

* How will a normal path vs `file://` uri behave?
* What is the expected result when a given context does not support `mkdir`?
* Does it work with complex streams like `php://filter/write=convert.base64-encode/resource=php://fd/3`? (This is also a use case where the surrounding `mkdir` calls makes zero sense)

rieves:
Ok, I'll look those points. Thank you.

Seldaek:
Closing as this opens up too many weird potential cases IMO as @sbuzonas mentioned - the archives can still be pushed to s3 after the fact if really needed..

PR #7601, opened 2018-08-31
Update xdebug-handler, fixes #7596

### LINKED ISSUE #7596: composer install results in NoSslException
Hi,

I update composer from 1.6.5 to composer 1.7.2 and the result is that composer doesn't work any longer, see diagnose output.
With version 1.6.5 anything is ok.
I use a bash shell (Windows 10, git-bash 2.14.1)

```
  [Composer\Exception\NoSslException]
  The openssl extension is required for SSL/TLS protection but is not available. If you can not enable the openss
  l extension, you can disable this error, at your own risk, by setting the 'disable-tls' option to true.

diagnose
```

openssl is installed and within composer 1.6.5 it works as expected.

```
php -m | grep openssl
openssl
```
I found out that we use the environment variable $PHP_INI_SCAN_DIR which shows to an alternative php.ini. In the composer shell script I see the line with php "${dir}/composer.phar" "$@" and when I add after php the parameter '-c $PHP_INI_SCAN_DIR' anything works again as expected.
I don't know why php in your script do not support such variable any longer. May be a bug.

Kind regards
Lars

johnstevenson:


johnstevenson:
Sorry, forgot that bit!

alcohol:
Thanks!

Seldaek:
Thanks

Issue #7314, opened 2018-05-07
How to exclude `require-dev` from `check-platform-reqs` command


eidng8:
My `composer.json`:

```json
{
  "require": {
    "php": ">=7.1",
    "ext-apcu": "*",
    "ext-imap": "*",
    "ext-intl": "*",
    "ext-json": "*",
    "ext-mbstring": "*",
    "ext-openssl": "*",
    "ext-PDO": "*",
    "ext-zip": "*",
    "alcohol/iso4217": "^3.1",
    "guzzlehttp/guzzle": "^6.3",
    "laravel/framework": "5.6.*",
    "laravel/tinker": "~1.0",
    "league/iso3166": "^2.1",
    "spatie/laravel-permission": "^2.11",
    "aws/aws-sdk-php": "^3.43",
    "webklex/laravel-imap": "^1.0"
  },
  "require-dev": {
    "ext-xdebug": "*",
    "barryvdh/laravel-debugbar": "^3.1",
    "barryvdh/laravel-ide-helper": "^2.4",
    "doctrine/dbal": "^2.5",
    "fzaninotto/faker": "~1.7",
    "justinrainbow/json-schema": "^5.2",
    "laravel/dusk": "~3.0",
    "mockery/mockery": "^1.0",
    "phpunit/phpunit": "~7.0",
    "squizlabs/php_codesniffer": "^3.2"
  },
  "autoload": {
    "files": [
      "app/helpers.php"
    ],
    "classmap": [
      "database"
    ],
    "psr-4": {
      "SomeApp\\": "app/"
    },
    "exclude-from-classmap": [
      "/tests/"
    ]
  },
  "autoload-dev": {
    "psr-4": {
      "Tests\\": "tests/"
    }
  },
  "scripts": {
    "post-root-package-install": [
      "php -r \"file_exists('.env') || copy('.env.example', '.env');\""
    ],
    "post-create-project-cmd": [
      "php artisan key:generate"
    ],
    "post-install-cmd": [
      "Illuminate\\Foundation\\ComposerScripts::postInstall"
    ],
    "post-update-cmd": [
      "Illuminate\\Foundation\\ComposerScripts::postUpdate",
      "php artisan ide-helper:generate",
      "php artisan ide-helper:meta"
    ]
  },
  "config": {
    "preferred-install": "dist",
    "sort-packages": true,
    "optimize-autoloader": true,
    "secure-http": false
  }
}
```

Output of `composer diagnose`:

```
Checking composer.json: OK
[...]

PR #7605, opened 2018-09-02
Add --no-dev option to check-platform-reqs command (#7314)

lamp-of-god:
This pull request covers https://github.com/composer/composer/issues/7314 issue.

It adds "--no-dev" option to "check-platform-reqs" command.

To do this it changes "check-platform-reqs" logic. Previous logic considered `installed.json` file to check platform reqs. Current logic considers `composer.lock` to perform such check.

Logic has been changed, because currently there is no easy way to check whether requirement is `required-dev` or not with `installed.json` file. On other hand, it's really easy to do with `composer.lock`.

`installed.json` file should be related to `composer.lock`, so there is no big deal at that point.

### Why it's ok to use `composer.lock`

On dev machines, both of files (`installed.json` and `composer.lock`) exist.

On prod (which may not contain `composer.lock`) such commands should not be executed (like you state [here](https://github.com/composer/composer/issues/7314#issuecomment-386985135) ) until this pull request, so it is not a big breaking change. 

Even more: `check-platform-reqs` command currently requires `composer.json` to exists and be valid. This means, that command can rely on dev-files, and `composer.lock` is just one of them, so its totally ok to use it.

### DIFF
--- a/src/Composer/Command/CheckPlatformReqsCommand.php
+++ b/src/Composer/Command/CheckPlatformReqsCommand.php
@@ -17,6 +17,7 @@ use Composer\Package\PackageInterface;
 use Composer\Semver\Constraint\Constraint;
 use Symfony\Component\Console\Helper\Table;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Composer\Repository\PlatformRepository;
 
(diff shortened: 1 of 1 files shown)


lamp-of-god:
Please, note that Travis CI checks are passed, but just one of them failed due to timeout. Seems that there was problem with CI server, not with code.

Seldaek:
Thanks

PR #7623, opened 2018-09-08
Clean up things

carusogabriel:
Each commit with its message should describe what was done :)


alcohol:
I'm not sure how I feel about the 'inline useless variables' commit. The benefit of those assignments before returning is that it makes the flow easier to debug in interactive debug sessions with xdebug.

Issue #7630, opened 2018-09-12
Composer v2: Pool/Solver/Repo/Installer Tasks


naderman:
- [x] Separate RepositorySet from Pool to allow for clear separation of package loading from dep resolution https://github.com/composer/composer/pull/7625 https://github.com/composer/composer/pull/7629
  - [x] Do not modify package pool during dep resolution
- Pool Builder
  - [x] use version ranges as part of loading mechanism, to potentially allow on-the-fly discarding of JSON data if we add a streaming parser https://github.com/composer/composer/pull/7631
    - [X] For nested dependencies, attempt to determine a union constraint which loads all potential required versions, but not all versions of packages which are deps of deps of the root package or deeper, needs some mechanism to notice that the constraint is getting too complex and it's not worth it, but could be helpful if at least there are simple ranges to be combined https://github.com/composer/composer/pull/7644
    - [ ] improve estimated constraints to replicate performance of dropped old versions in flex
  - [x] don't load package data and generate rules multiple times for identical packages https://github.com/composer/composer/pull/7830
  - [x] fully remove information on installed packages
  - [x] Tests for Pool Builder mechanism https://github.com/composer/composer/commit/921a97457dffbb2e969b54abb13deaa64238c95b
  - [x] do not load unnecessary package data for fixed packages (e.g. partial updates) - see issue https://github.com/composer/composer/issues/3672 - fixed by https://github.com/composer/composer/pull/8531 and https://github.com/composer/composer/pull/8535
- Solver
  - [x] do not involve installed packages, removal operations for all installed packages which are not in the new lock file should be automatic
  - ~~create clear interface that allows replacement (e.g. by sat solver utility on system) related to #6685~~
  - [x] reduce conflict rules - see issue https://github.com/composer/composer/issues/7452
- Error reporting
  - [x] https://github.com/composer/composer/issues/7779
- Installer
  - [x] clean up messy update whitelisting mechanism related https://github.com/composer/composer/issues/8355 and https://github.com/composer/composer/issues/5438 - PR https://github.com/composer/composer/pull/8717
  - [x] update Request class to represent a root composer.json + instructions more directly
- Composer Repository
  - [x] Load named packages directly instead of provider files
  - [x] Load metadata for multiple packages in parallel
  - ~Fully parallelize metadata downloads across multiple repositories~ (cancelled until at least 2.1)

### Memory Reduction
[...]

PR #7631, opened 2018-09-12
Only load package versions which fit the root composer.json constraints

naderman:
Part of https://github.com/composer/composer/issues/7630

### DIFF
--- a/tests/Composer/Test/Fixtures/installer/solver-problems.test
+++ b/tests/Composer/Test/Fixtures/installer/solver-problems.test
@@ -42,14 +42,16 @@ Updating dependencies (including require-dev)
 Your requirements could not be resolved to an installable set of packages.
 
   Problem 1
-    - The requested package unstable/package 2.* exists as unstable/package[1.0.0] but these are rejected by your constraint.
+    - The requested package unstable/package could not be found in any version, there may be a typo in the package name.
   Problem 2
     - The requested package bogus could not be found in any version, there may be a typo in the package name.
   Problem 3
-    - The requested package stable-requiree-excluded (installed at 1.0.0, required as 1.0.1) is satisfiable by stable-requiree-excluded[1.0.0] but these conflict with your requirements or minimum-stability.
+    - The requested package stable-requiree-excluded 1.0.1 exists as stable-requiree-excluded[1.0.0] but these are rejected by your constraint.
   Problem 4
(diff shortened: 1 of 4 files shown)


tests/Composer/Test/Fixtures/installer/solver-problems.test
```diff
@@ -42,14 +42,16 @@ Updating dependencies (including require-dev)
 Your requirements could not be resolved to an installable set of packages.
 
   Problem 1
-    - The requested package unstable/package 2.* exists as unstable/package[1.0.0] but these are rejected by your constraint.
+    - The requested package unstable/package could not be found in any version, there may be a typo in the package name.
```
stof: That change looks wrong to me.

Optimizing the resolution algorithm is fine. But in case of failure, we should still look at *all* available packages to decide whether the package could not be found at all, or only in incompatible versions.

src/Composer/Plugin/PluginInterface.php
```diff
@@ -27,7 +27,7 @@
      *
      * @var string
      */
-    const PLUGIN_API_VERSION = '2.0.0';
+    const PLUGIN_API_VERSION = '1.2.0';
```
staabm: Typo?

tests/Composer/Test/Fixtures/installer/solver-problems.test
stof: note that we could extend the Repository class to provide a fast check for package existence. A ComposerRepository using providers (like Packagist does) does not need to load package metadata to know whether it knows the package or no (as providers are split by package names, the list of providers is enough to know whether the package name exists)

src/Composer/Plugin/PluginInterface.php
naderman: Well accident. Had to revert this to test on some real projects ;-) thanks

tests/Composer/Test/Fixtures/installer/solver-problems.test
naderman: I already noted this in https://github.com/composer/composer/issues/7630 - the error reporting will need to be changed just as you described, but that will be a separate PR.

Part of this refactoring will actually be to remove the providers files from packagist, so that part is not true anymore ;-)

stof: well, we should still have a way to check for package existence on Packagist, even if it is different, no ?

naderman: Yes absolutely, but we'll have to build that first, so I'm going to do the error reporting task at a later point, and just leave these messages here as they are for now.

PR #7633, opened 2018-09-12
Remove pool whitelist

naderman:
Part of https://github.com/composer/composer/issues/7630


stof:
shouldn't we have code added in PoolBuilder too ?

naderman:
@stof no the pool builder already does this, it was just redundant at this point

Issue #7624, opened 2018-09-08
"composer validate" exits with code 1 when composer.json is valid


sebastianbergmann:
```
$ git clone https://github.com/sebastianbergmann/phpunit.git
$ cd phpunit
```
```
$ composer install
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 26 installs, 0 updates, 0 removals
  - Installing theseer/tokenizer (1.1.0): Loading from cache
  - Installing sebastian/version (2.0.1): Loading from cache
  - Installing sebastian/environment (3.1.0): Loading from cache
  - Installing sebastian/code-unit-reverse-lookup (1.0.1): Loading from cache
  - Installing phpunit/php-token-stream (3.0.0): Loading from cache
  - Installing phpunit/php-text-template (1.2.1): Loading from cache
  - Installing phpunit/php-file-iterator (2.0.1): Loading from cache
  - Installing phpunit/php-code-coverage (6.0.7): Loading from cache
  - Installing phpunit/php-timer (2.0.0): Loading from cache
  - Installing sebastian/global-state (2.0.0): Loading from cache
  - Installing sebastian/recursion-context (3.0.0): Loading from cache
  - Installing sebastian/object-reflector (1.1.1): Loading from cache
  - Installing sebastian/object-enumerator (3.0.3): Loading from cache
  - Installing sebastian/resource-operations (1.0.0): Loading from cache
  - Installing myclabs/deep-copy (1.8.1): Loading from cache
  - Installing phar-io/version (2.0.1): Loading from cache
  - Installing phar-io/manifest (1.0.3): Loading from cache
  - Installing doctrine/instantiator (1.1.0): Loading from cache
  - Installing sebastian/exporter (3.1.0): Loading from cache
  - Installing sebastian/diff (3.0.1): Loading from cache
  - Installing sebastian/comparator (3.0.2): Loading from cache
  - Installing webmozart/assert (1.3.0): Loading from cache
  - Installing phpdocumentor/reflection-common (1.0.1): Loading from cache
  - Installing phpdocumentor/type-resolver (0.4.0): Loading from cache
  - Installing phpdocumentor/reflection-docblock (4.3.0): Loading from cache
  - Installing phpspec/prophecy (1.8.0): Loading from cache
sebastian/global-state suggests installing ext-uopz (*)
Writing lock file
Generating optimized autoload files
```
```
$ cp composer.json composer.json.bak
$ composer require --no-update phpunit/php-invoker:^2.0
[...]

PR #7637, opened 2018-09-13
Update ValidateCommand.php

yassine-ah:
Skip publish and lock check even in strict check mode when the user want so.
Example : `composer validate --no-check-lock --strict composer.json`.
Issue : #7624


Seldaek:
Thanks!

PR #7642, opened 2018-09-13
Improve path repository UX

mleczakm:
When running "composer outdated" command and using local (path) repositories - result has always return code higher than 0 and local repository listed. This PR makes it also displays found new version, but I'm not sure if this is solution to this problem, since the result of "composer outdated" is still wrong. i think we should dropped version checking for path repositories, because:
1, when they are symlinks - they are always updated by default
2. when not - they are copied on `install` command anyway

### DIFF
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -245,6 +245,7 @@ class Pool implements \Countable
 
                 case self::MATCH_NAME:
                     $nameMatch = true;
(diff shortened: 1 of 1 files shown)


Seldaek:
Sorry but I am not so comfortable merging this fix as it touches the pool and I don't think it should. I don't see path repo packages in my outdated list here so I am not sure what you mean exactly, please report this as an issue with a reproduction case and we can see if it can be improved/fixed then.

PR #7644, opened 2018-09-14
PoolBuilder:  prune impossible versions before pool creation

naderman:
This still loads all package metadata by name but it collects all constraints for all names as it loads data. It then tries to prune all versions away which do not match a single constraint used for that package name, before the pool is created and before the solver creates rules.

Results:
Composer: 8MB peak saved, 14% rule reduction, 12% package object count reduction
Sylius: 22MB peak saved, 2.4% rule reduction, 14% package object count reduction

Applying this pruning to aliased packages seems unnecessarily complex for hardly any benefit that I could measure, so I may want to remove that part from this PR.

Alternatively we could still try to prune versions as part of data loading already. The only downside is that it's possible that a versions already pruned for a package may be needed if we later encounter a package with a different constraint for the same name that's wider than the previous constraints.

### DIFF
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -58,12 +58,12 @@ class Pool implements \Countable
 
     public function setPackages(array $packages, array $priorities = array())
     {
-        $this->priorities = $priorities;
-        $this->packages = $packages;
-
         $id = 1;
 
-        foreach ($this->packages as $package) {
+        foreach ($packages as $i => $package) {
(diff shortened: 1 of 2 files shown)


Letsgetitons:
Help me

naderman:
Pull Request 7644 - is that you speaking?

naderman:
Looked at this some more, while the impact of the alias map is minimal, it may help when installing various branch aliased versions together. Further it's helpful for additional work in the pool builder to have a map of all aliases available, so going to merge as-is.

PR #7648, opened 2018-09-16
Allow plugin commands to be run from child folders like core commands

NickWilde1990:
Core commands can be run from child directories and will look for the first composer.json in the parent folders. However, plugin commands get kicked with a `Symfony\Component\Console\Exception\CommandNotFoundException` before that/before plugins are actually known.
This changes it so that that does work.


Seldaek:
Thanks

PR #7649, opened 2018-09-17
Use a case insenstive method to check that SHA384 is a supported openssl algorithm

aragon999:
This pull request ports the fix from the installer script, see here https://github.com/composer/getcomposer.org/pull/128, to the composer self-update command, where the same bug is present.


alcohol:
@Seldaek see also https://github.com/composer/getcomposer.org/pull/128

Seldaek:
@aragon999 can you please rebase this on the 1.7 branch?

aragon999:
@Seldaek done :-)

alcohol:
Are you sure? It does not look rebased properly. Is your 1.7 branch up to date?

aragon999:
@alcohol sorry that was my mistake, should be fixed now

Seldaek:
Thanks

PR #7673, opened 2018-09-26
Fix the doc markup

stof:
In `*nix`, the star needs to be part of the content, not of the markdown markup, and so it requires escaping.


stof:
Note that the Github markdown renderer does not show any difference here, because it already considers that this is not italic. But Parsedown considers that case as italic.

stof:
another option would be to change the parser used by the composer website to use a Commonmark parser.

Seldaek:
Thanks

PR #7680, opened 2018-09-30
Substituted wget commands with curl to support more platforms includi…

b23prodtm:
…ng Darwin


alcohol:
Might want to include `-L` and `-s`.

staabm:
I also stumbled over systems in the past which had curl but not wget.

not sure this change is worthwhile.

alcohol:
Yeah. It's often one or the other (although most systems I own/use have both). Maybe we should just provide easy copy & paste instructions for both.

Seldaek:
Thanks but nope, Ubuntu server for example comes with wget but no curl AFAIR. We can't please everyone either way, so rather keep it the way it is.

PR #7687, opened 2018-10-02
Fixes from PHPStan

carusogabriel:
This PR aims to fix some issues founded by https://github.com/phpstan/phpstan, such as: undefined variables and properties, autoload issues, excess of variables, phpDocs, and much more.

I've only got to level 3, as I don't know the interesting of the maintainers in static analysis. Let me know if I should keep the work up to level 7, as well as slipt the commit and others, joining similar changes, with a description to help debug in the future :)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -771,8 +771,8 @@ class ComposerStaticInit$suffix
         $vendorPathCode = ' => ' . $filesystem->findShortestPathCode(realpath($targetDir), $vendorPath, true, true) . " . '/";
         $appBaseDirCode = ' => ' . $filesystem->findShortestPathCode(realpath($targetDir), $basePath, true, true) . " . '/";
 
-        $absoluteVendorPathCode = ' => ' . substr(var_export(rtrim($vendorDir, '\\/') . '/', true), 0, -1);
-        $absoluteAppBaseDirCode = ' => ' . substr(var_export(rtrim($baseDir, '\\/') . '/', true), 0, -1);
+        $absoluteVendorPathCode = ' => ' . substr(var_export(rtrim($vendorPath, '\\/') . '/', true), 0, -1);
```
carusogabriel: Looks like these changes broke the Windows build, but, `$vendorDir` isn't a defined variable, so we might be facing some false-positives?

tests/Composer/Test/TestCase.php
```diff
@@ -10,7 +10,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Composer;
+namespace Composer\Test;
```
carusogabriel: Couldn't figure out this error: https://travis-ci.org/composer/composer/jobs/435914320#L751

alcohol: The problem is that this file should not be in this namespace.

It is hardcoded in the bootstrap for the testsuite: https://github.com/composer/composer/blob/master/tests/bootstrap.php

The exact reasons behind this are actually not known to me. I am sure there is some history behind it.

src/Composer/Autoload/AutoloadGenerator.php
alcohol: How can you know it is not defined? Several files were `require`ed in this method before these lines. I would not touch these lines. The `AutoloadGenerator` is one of the most critical parts of Composer. The code is quite well tested and definitely working as intended.

src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -264,7 +264,7 @@ protected function getPhpExecCommand()
         $finder = new PhpExecutableFinder();
         $phpPath = $finder->find();
         if (!$phpPath) {
-            throw new \RuntimeException('Failed to locate PHP binary to execute '.$scriptName);
+            throw new \RuntimeException('Failed to locate PHP binary to execute '.$phpPath);
```
alcohol: This is a good catch.

tests/Composer/Test/TestCase.php
alcohol: Our PSR-4 definition for `Composer\Test` expects all files to be in `tests/Composer/Test`, so I really think you should undo this change. Because I am not sure if we actually want to move this file into said directory.

tests/Composer/Test/InstallerTest.php
```diff
@@ -151,7 +151,7 @@ public function provideInstaller()
     public function testIntegration($file, $message, $condition, $composerConfig, $lock, $installed, $run, $expectLock, $expectOutput, $expect, $expectResult)
     {
         if ($condition) {
-            eval('$res = '.$condition.';');
+            $res = $condition;
```
alcohol: I'm not sure about this change...

tests/Composer/Test/Config/JsonConfigSourceTest.php
```diff
@@ -10,11 +10,11 @@
  * file that was distributed with this source code.
  */
 
-namespace Composer\Test\Json;
+namespace Composer\Test\Config;
```
alcohol: Huh, weird. Surprised this did not cause any issues.

src/Composer/Autoload/AutoloadGenerator.php
alcohol: Those variables are defined in each of the above `require`ed files (they'll have the same values in all of them).

tests/Composer/Test/Plugin/PluginInstallerTest.php
```diff
@@ -10,7 +10,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Composer\Test\Installer;
+namespace Composer\Test\Plugin;
```
alcohol: Another good catch.

tests/Composer/Test/Question/StrictConfirmationQuestionTest.php
```diff
@@ -10,7 +10,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Composer\Question\Test;
+namespace Composer\Test\Question;
```
alcohol: I really wonder how all these mistaken namespaces never caused any issues.

alcohol:
:+1: 
- The docblock fixes are generally welcome. 
- The removal of redundant arguments being passed to methods (probably legacy artifacts) is also welcome.
- Some small fixes (wrong variable name. wrong namespace name etc).

:man_shrugging: 
- Not sure about the moving of and changing the namespace of the `TestCase`.
- Also not sure about removal of the `eval()` call. I think this actually alters behaviour.

:-1: 
- I would definitely revert the changes to the `AutoloadGenerator`.

alcohol:
> slipt the commit in others, joining similar changes, with a description to help debug in the future

Do you mean squash / rebase? I would definitely be in favor of joining similar changes in the same commits if it is not too much work for you. Makes it easier to cherry pick or revert them in the future if something proves to not work out as expected.

carusogabriel:
> Not sure about the moving of and changing the namespace of the TestCase.

What do you suggest?

> Also not sure about removal of the eval() call. I think this actually alters behaviour.

I don't think so, but we can listen from @Seldaek, as the commit is from him.

> I would definitely revert the changes to the AutoloadGenerator.

Gotcha, gonna revert them.

> Do you mean squash / rebase? I would definitely be in favor of joining similar changes in the same commits if it is not too much work for you

I'll do it and continue the work up to level 7, separating into different commits that will help us in the future :)

alcohol:
> > Not sure about the moving of and changing the namespace of the TestCase.

> What do you suggest?

I am not sure if it was done intentionally or not. Maybe @Seldaek can also answer this question.

tests/Composer/Test/Config/JsonConfigSourceTest.php
sstok: Tests are executed as files, their namespace does not matter unless the class is used elsewhere 👍

tests/Composer/Test/InstallerTest.php
sstok: This indeed changes the outcome. Technically this could be written as `eval("\$res = $condition;");`, and it shows the condition is executed. Also this is part of the test-suite, so no major issues here.

tests/Composer/Test/Question/StrictConfirmationQuestionTest.php
curry684: PHPunit just doesn't give a damn - it scans the test folders for `*Test.php` files and then loads whatever is in there. You could put them all in namespace `a/b/c/d/e` for all it cares, or none at all. Essentially namespaces only matter if you rely on PSR autoloading, and PHPunit cannot because it supports all projects, hence why you need to use `vendor/autoload.php` as the bootstrap code manually.

tests/Composer/Test/TestCase.php
stof: this file is *alreadys in said directory. but all places extending this class must be updated to use the new namespace

tests/Composer/Test/InstallerTest.php
stof: this change is totally wrong. `$condition` is a PHP source code, not a PHP value to be used in `$res`

stof:
I think TestCase being out of the Test namespace was a mistake (its location corresponds to being in that namespace)

tests/Composer/Test/Question/StrictConfirmationQuestionTest.php
alcohol: Yeah, still sloppy though.

alcohol:
I took the liberty of rebasing your branch and splitting your work into individual commits, while also dropping the undesired/unwanted changes. I hope you don't mind.

alcohol:
Oh, it fails in Travis because it expects tests inside `tests/Composer/Test/TestCase.php`, but this file only contains an abstract class.

carusogabriel:
@alcohol I don't mind, but I really want to fix all warnings from PHPStan, especially those ones from Composer's methods documentation. That will help us in the future migrate to PHP 7 by automating it via CS :)

alcohol:
Sure, but I do suggest we do not let this PR grow too far. Perhaps split the levels over individual PRs. Makes it easier to review them also and the impact of their changes also more easy to manage.

carusogabriel:
@alcohol Gotcha. So, feel free to merge this one, and for the next ones, I'll open one PR per level, and also group similar changes in different commits. Thanks for the review and interesting in these changes :blush:

alcohol:
@carusogabriel I would suggest waiting for some feedback from @Seldaek, but if you really already want to get started on the next level, I suggest you base a new branch from this one and open another PR for that. Will make it easier to rebase if this gets accepted and merged.

src/Composer/DependencyResolver/RuleWatchGraph.php
```diff
@@ -127,9 +127,9 @@ public function propagateLiteral($decidedLiteral, $level, $decisions)
      *
      * The rule node's watched literals are updated accordingly.
      *
-     * @param $fromLiteral mixed A literal the node used to watch
-     * @param $toLiteral mixed A literal the node should watch now
-     * @param $node mixed The rule node to be moved
+     * @param mixed $fromLiteral A literal the node used to watch
+     * @param mixed $toLiteral A literal the node should watch now
```
stof: literals of the solver are integers

src/Composer/DependencyResolver/RuleWatchGraph.php
```diff
@@ -127,9 +127,9 @@ public function propagateLiteral($decidedLiteral, $level, $decisions)
...
+     * @param mixed $node The rule node to be moved
```
stof: this is a `RuleWatchNode`

src/Composer/Util/NoProxyPattern.php
```diff
@@ -35,7 +35,7 @@ public function __construct($pattern)
      *
      * @param string $url
      *
-     * @return true if the URL matches one of the rules.
+     * @return bool if the URL matches one of the rules.
```
stof: should read like `@return bool true if the URL matches one of the rules.`

src/Composer/Util/TlsHelper.php
```diff
@@ -164,7 +164,7 @@ public static function isOpensslParseSafe()
      *
      * @param string $certName CN/SAN
      *
-     * @return callable|null
+     * @return callable|void
```
stof: `callable|void` does not make sense in phpdoc

carusogabriel: @stof Look into the method, there're void returns :)

Seldaek:
Thanks. I merged as these were mostly minor fixes and no brainers. Looking at the higher level fixes of phpstan though it seems to contain a lot of nonsense and things that are technically correct but also do not matter in practice and aren't really fixable. I doubt it brings much value to go through all of them in details.

Seldaek:
The other aspect is that these kind of changes tend to cause conflicts.. and as we have some work going on in the 2.0 branch I'd rather avoid any further such PRs until 2.0 comes out. It makes merging things much easier.

PR #7689, opened 2018-10-02
Apply yoda style conditions

omarkdev:
This pull request has the intention of applying Yoda Style, since in some files (the main ones) this style of code is very common, so to maintain the project's fidelity and prevent certain errors, was applied the Yoda Style in codes that did not have

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -53,6 +53,7 @@ return PhpCsFixer\Config::create()
         'standardize_not_equals' => true,
         'ternary_operator_spaces' => true,
         'trailing_comma_in_multiline_array' => true,
+        'yoda_style' => true,
     ))
     ->setFinder($finder)
 ;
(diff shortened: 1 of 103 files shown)


alcohol:
Yoda style is a bit controversial and more a matter of opinion/preference and not really significant in any other way. I do agree though that it would be preferable to maintain a single consistent style throughout the codebase. And as @carusogabriel pointed out, it would be best to do this as much as possible through the use of the ruleset for [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer).

omarkdev:
Ok, following @carusogabriel comment, I also added `yoda_style` in the `.php_cs` file and also I performed `php-cs-fixer` to resolve the other conditions of yoda style that I had not found.

Seldaek:
Thanks but I rather not apply such change that will cause tons of conflicts for arguably almost no benefit.

PR #7699, opened 2018-10-09
Warning about the UNIX permissions lost if unzip command is not installed

Deuchnord:
This pull request updates the warning about the `unzip` command not available to put in the spotlight, the fact that UNIX permissions will not be kept. This is especially important for the packages such as [Symfony Panther](https://github.com/symfony/panther) or [Laravel Dusk](https://github.com/laravel/dusk), which provide some binary files that should have the executable permission.

This issue is due to PHP's ZIP extension not handling UNIX permissions, which are a an extension to Microsoft's standard by the open-source software's community.

As using PHP's Unzip extension may lead to some strange issues, I'd like to propose to ask the user if they really want to continue that way. What do you think?


Seldaek:
Thanks, merged in 1.7 & master as https://github.com/composer/composer/commit/a51563300c77e339300d2fbeff796aea95f3d0e1

PR #7709, opened 2018-10-12
Don't call Symfony ProcessUtils::escapeArgument

mlocati:
We currently check if `Symfony\Component\Process\ProcessUtils` has the `escapeArgument` method in order to call it (it has been removed in Symfony 4).

BTW `ProcessUtils::escapeArgument` throws a `E_USER_DEPRECATED` warning for Symfony 3.3+.

So, what about avoid calling it?

### DIFF
--- a/src/Composer/Util/ProcessExecutor.php
+++ b/src/Composer/Util/ProcessExecutor.php
@@ -131,15 +131,11 @@ class ProcessExecutor
      */
     public static function escape($argument)
     {
-        if (method_exists('Symfony\Component\Process\ProcessUtils', 'escapeArgument')) {
-            return ProcessUtils::escapeArgument($argument);
-        }
-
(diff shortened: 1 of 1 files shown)


staabm:
composer also supports older symfony/process versions, therefore needs code which works with all of them

https://github.com/composer/composer/blob/master/composer.json#L37

mlocati:
> composer also supports older symfony/process versions, therefore needs code which works with all of them

And the copy of `escapeArgument` currently present in Composer `ProcessExecutor` doesn't work for older symfony/process versions?

staabm:
ohh sorry, I misread your patch. LGTM

Seldaek:
Thanks, merged in 1.7 & master as https://github.com/composer/composer/commit/41458c75184bde3805d622503b7b5973ef3f1a8b

Issue #7696, opened 2018-10-07
`composer show` does not respect --path option if called for single package


axel-rutz:
As a shell scripter, i want to extract the path of a package.

When i try this with `composer show --path`,

* `composer show --path foo/*` gives me what i want (name and path)
* `composer show --path foo/bar` does **not** give me what i want (name and path), but seems to ignore the --path option

## Example

```
$ composer show -P twig/*
twig/twig /home/merlin/PhpstormProjects/d8.5-dev/vendor/twig/twig

$ composer show -P symfony/*
symfony/browser-kit             /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/browser-kit
symfony/class-loader            /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/class-loader
symfony/console                 /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/console
symfony/css-selector            /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/css-selector
symfony/debug                   /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/debug
symfony/dependency-injection    /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/dependency-injection
symfony/dom-crawler             /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/dom-crawler
symfony/event-dispatcher        /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/event-dispatcher
symfony/finder                  /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/finder
symfony/http-foundation         /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/http-foundation
symfony/http-kernel             /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/http-kernel
symfony/phpunit-bridge          /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/phpunit-bridge
symfony/polyfill-iconv          /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/polyfill-iconv
symfony/polyfill-mbstring       /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/polyfill-mbstring
symfony/polyfill-php70          /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/polyfill-php70
symfony/process                 /home/merlin/PhpstormProjects/d8.5-dev/vendor/symfony/process
[...]

PR #7698, opened 2018-10-08
Composer Show to respect --path flag

vallode:
Attempts to fix #7696 

When the --path option exists for a single package call it simply returns the name and path of the package
I also thought it'd be nice to include the path when returning the full package meta.

PR #7711, opened 2018-10-13
composer/composer#7696: composer show does not respect --path option

dmanners:
`composer show` does not respect `--path` option if called for single package

 - update the handling of the single package composer show command to include path info
 - validate we have the path set,
 - show and get the path information,
 - move up validation of `$composer` and path option to before our changes so as not to cause issues if `$composer` is not set

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -183,6 +183,11 @@ EOT
 
         $packageFilter = $input->getArgument('package');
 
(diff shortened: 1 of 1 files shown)


dmanners:
I have updated the output to show the path information last if the option is passed to the command.

Happy to update anything if needed.

dmanners:
Woops seems there is already a PR that should cover this :( never mind

https://github.com/composer/composer/pull/7698

Issue #7384, opened 2018-06-01
New feature - Slack and/or Discord support links


Truemedia:
So I see the support section of the composer schema has an option for an irc channel, but this is quite old school now and so many organisations use slack or discord as a more modern alternative.

If you try putting a URL for either of those in the IRC field it causes a composer error, so ideally would be nice to have new field to put these URL's in.

I use satis and been able to directly search for a package and click the support links it fetches back from composer is extremely useful for me.

Seldaek:
We could add a "chat" node I guess, calling it slack sounds a bit restrictive, and I'd rather not add gitter and co as well. For discord I'd put it in the `forum` URL.

Truemedia:
Yeh something agnostic would be ideal, and forum does make sense if linking to a discord server for a dedicated community rather than a dedicated channel.

ryanaslett:
discourse (https://www.discourse.org/) should go to the forums. Discord (https://discordapp.com/) is a slack clone, of which theres also rocketchat, matrix, and many others, so having a "chat" heading is a good idea.

Seldaek:
Ah thanks for the clarification @ryanaslett, I indeed got those mixed up.

Seldaek:
Oh this was fixed by https://github.com/composer/composer/pull/7714 ages ago.

PR #7714, opened 2018-10-14
composer/composer#7384: add chat to support options

dmanners:
- update schema documents to note that chat is allowed,
 - validate that chat must be a string and a url similar to forum options

Is there anything else that should be changed? As far as I can see most chat options like slack and discord take a url to register users.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -243,7 +243,7 @@ Support information includes the following:
 * **source:** URL to browse or download the sources.
 * **docs:** URL to the documentation.
 * **rss:** URL to the RSS feed.
-* **chat:** URL to chat server.
+* **chat:** URL to the chat channel.
 
 An example:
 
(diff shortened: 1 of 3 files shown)


doc/04-schema.md
```diff
@@ -243,6 +243,7 @@ Support information includes the following:
 * **source:** URL to browse or download the sources.
 * **docs:** URL to the documentation.
 * **rss:** URL to the RSS feed.
+* **chat:** URL to chat server.
```
sanmai:
```suggestion
* **chat:** URL to a chat channel.
```

dmanners:
Thanks for the review @sanmai I have updated this based on your feedback.

er1z:
Why to keep both irc and chat nodes?

dmanners:
Good point @er1z I never thought about it but I guess in this case irc and chat would overlap. I would be happy to replace one with the other but I am not sure if we would need to keep irc for backwards compatibility, or if during the validate we should flag that irc is deprecated for the chat.

What is the suggested approach here?

Seldaek:
Can you also please update res/composer-schema.json See https://github.com/composer/composer/pull/5125/files for an example

Seldaek:
Thanks, did the fix in https://github.com/composer/composer/commit/b89daf5322711ac154d5583ce4443588e43c1716

PR #7715, opened 2018-10-14
composer/composer#7159: make the remove command do a lookup

### LINKED ISSUE #7159: [Feature] Remove all packages with same vendor
I dont know if this is already implemented but it would be handy to remove all packages with same vendor with one command instead of typing all package names manually, specially if you have a lot.

For example i have 3 packages all under the same vendor "foo":
- foo/foo
- foo/bar
- foo/baz

If i want to remove them all, i could do it like this: **composer remove foo/***
This would then remove the 3 packages listed above.

dmanners:
Make the remove command do a lookup for package name

 - if you have multiple vendor modules installed you should be able to do composer remove vendor/* to remove all
 - update remove and also remove from alternative type to also do a preg_grep for what the user has inputed

Fixes #7159

### DIFF
--- a/src/Composer/Command/RemoveCommand.php
+++ b/src/Composer/Command/RemoveCommand.php
@@ -94,10 +94,23 @@ EOT
             if (isset($composer[$type][$package])) {
                 $json->removeLink($type, $composer[$type][$package]);
             } elseif (isset($composer[$altType][$package])) {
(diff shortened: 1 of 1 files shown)


dmanners:
I am thinking it might be best to refactor this a bit since we are now doing some duplication of the removal and asking the user if they want to remove from the alternative type.

Happy to make these changes if there is feedback here.

Seldaek:
Thanks, I did the fix in https://github.com/composer/composer/commit/66d84f60c6e6c869ba739db239f23366c66cb144

PR #7719, opened 2018-10-15
Report "same as actual" version if override package matches actual

ComiR:
When running `composer show -p` and having defined overrides via config.platform, the corresponding lines currently will show as
```
php                 7.2.9    Package overridden via config.platform (actual: 7.2.9)
```
This isn't very practical if you want to determine discrepancies.
Therefor I propose changing it to the following if the version matches:
```
php                 7.2.9    Package overridden via config.platform (same as actual)
```
If the versions don't match, the same format as before is used:
```
lib-curl            7.47.0   Package overridden via config.platform (actual: 7.52.1)
```


Seldaek:
Thanks, merged in 1.7 & master as https://github.com/composer/composer/commit/3c543b275279b27d3e723ffe4363e0aecaf07857

PR #7745, opened 2018-10-24
Fix: undefined index name in VcsRepository

glaubinix:
Requiring a VcsRepository that has tags where the composer.json is missing the name attribute in certain tags results in PHP notices.

This is the error message if I force throwing exceptions on notices
`Notice: Undefined index: name {"exception":"[object] (ErrorException(code: 0): Notice: Undefined index: name at /path/to/vendor/composer/composer/src/Composer/Repository/VcsRepository.php:201)

### DIFF
--- a/src/Composer/Repository/VcsRepository.php
+++ b/src/Composer/Repository/VcsRepository.php
@@ -188,7 +188,8 @@ class VcsRepository extends ArrayRepository implements ConfigurableRepositoryInt
                     continue;
                 }
 
-                if ($existingPackage = $this->findPackage($data['name'], $data['version_normalized'])) {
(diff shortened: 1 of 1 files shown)


src/Composer/Repository/VcsRepository.php
```diff
@@ -188,6 +188,7 @@ protected function initialize()
                     continue;
                 }
 
+                $tagPackageName = isset($data['name']) ? $data['name'] : $this->packageName;
                 if ($existingPackage = $this->findPackage($data['name'], $data['version_normalized'])) {
```
stof: this is still using `$data['name']` here. You variable looks unused

glaubinix: sorry, I never pushed the fixed commit earlier. This should be fixed now.

PR #7756, opened 2018-10-29
VcsRepository: fix undefined index notice in preProcess

glaubinix:
Requiring a VcsRepository that doesn't have a composer.json in the rootIdentifier branch and that has tags and/or branches where the composer.json is missing the name attribute results in PHP notices.

`$this->packageName` is `null` if there is no composer.json in the rootIdentifier branch.

`Notice: Undefined index: name {"exception":"[object] (ErrorException(code: 0): Notice: Undefined index: name at /path/to/vendor/composer/composer/src/Composer/Repository/VcsRepository.php:298`

### DIFF
--- a/src/Composer/Repository/VcsRepository.php
+++ b/src/Composer/Repository/VcsRepository.php
@@ -295,7 +295,8 @@ class VcsRepository extends ArrayRepository implements ConfigurableRepositoryInt
     protected function preProcess(VcsDriverInterface $driver, array $data, $identifier)
     {
         // keep the name of the main identifier for all packages
-        $data['name'] = $this->packageName ?: $data['name'];
+        $dataPackageName = isset($data['name']) ? $data['name'] : null;
+        $data['name'] = $this->packageName ?: $dataPackageName;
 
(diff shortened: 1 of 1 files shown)


src/Composer/Repository/VcsRepository.php
```diff
@@ -295,7 +295,8 @@ protected function initialize()
     protected function preProcess(VcsDriverInterface $driver, array $data, $identifier)
     {
         // keep the name of the main identifier for all packages
-        $data['name'] = $this->packageName ?: $data['name'];
+        $dataPackageName = isset($data['name']) ? $data['name'] : null;
+        $data['name'] = $this->packageName ?: $dataPackageName;
```
sbuzonas: Does it make sense to update package name here? Hard to guess which name is canonical without root metadata.

glaubinix: Right now, if no name is defined composer skips the version. Setting the root metadata package name here would change this behaviour for versions that are being processed after that. I don't think this is a good idea or something that should be introduced in a bug fix pr.

PR #7760, opened 2018-10-31
Fix ini_get() for boolean values

deguif:
Currently setting `false` or `off`, ... value to configure some PHP ini directives will make this evaluated to true as this is equal to a non empty string.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -279,7 +279,7 @@ class ClassLoader
      */
     public function setApcuPrefix($apcuPrefix)
     {
-        $this->apcuPrefix = function_exists('apcu_fetch') && ini_get('apc.enabled') ? $apcuPrefix : null;
+        $this->apcuPrefix = function_exists('apcu_fetch') && filter_var(ini_get('apc.enabled'), FILTER_VALIDATE_BOOLEAN) ? $apcuPrefix : null;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


jvasseur:
This shouldn't be needed, `ini_get` already returns a normalized value for boolean configs ("" for false and "1" for true).

deguif:
I encountered this issue when using environment variables to configure PHP ini directives:
```
#php.ini
[opcache]
opcache.enable_cli = ${PHP_INI_OPCACHE_ENABLE_CLI}
```

When using `false` or `off` string values in `${PHP_INI_OPCACHE_ENABLE_CLI}` instead of using `0`, this was not working as expected, as ini_get() will return the real value `false` in my case and comparing the string value `'false'` when using `ini_get()`will return true instead of returning false as expected, as this considered by PHP as valid value for boolean for ini file, see http://php.net/manual/fr/function.parse-ini-file.php

See this code: https://3v4l.org/dRsTO that may explain better

Seldaek:
IMO this is a misunderstanding on your part. The ini parser parses the file first, then if it sees `${ENV_VAR}` in a value it resolves that env var at runtime to whatever the value of the env var is. That means you can't use ini-style values, but need to use proper runtime values in your env var. Set the env var to `0` or empty string to get a false.

deguif:
@Seldaek what leads to this issue is that PHP evaluates rightly the environment variable and disables APCu which brings the issue.

Checking '`if (ini_get('apc.enable_cli'))`' will evaluate to true when it was set to `false` by the environment variable, but in the meantime PHP has interpreted the environment variable at runtime and has disabled APCu.

See the issue linked by @nicolas-grekas as I provided an example there.

Seldaek:
I see.. Sounds like something that should be reported to php as well though, seems like a bug in ini_get if they do parse it correctly internally.

Seldaek:
Thanks

PR #7763, opened 2018-11-01
Add source code comment block

medz:


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -58,12 +58,26 @@ class AutoloadGenerator
      */
     private $runScripts = false;
 
+    /**
```
staabm: This phpdoc for construct does not provide any value. Its repeating whats already present via the syntax

staabm: Actually this applies to kost comments in this PR

medz:
I am not the maintainer of Composer, so the description when adding these PHPDOCs is not very accurate, but it will make the automatic prompts in the code editor more friendly.

alcohol:
The English in some of these is rather.... incorrect.

alcohol:
While I appreciate the effort, I think it would be in your own benefit if you merely specified the variable names and their types and omit potentially difficult descriptions. You can't go wrong that way :-)

PR #7766, opened 2018-11-01
Dispatch "post-package-update" event after writing lock

### LINKED ISSUE #7765: Dispatch "post-package-update" event after updating composer.lock write
At the moment the `post-package-update` event is dispatched **before** updating the `composer.lock` file:

https://github.com/composer/composer/blob/ff59bbd/src/Composer/Installer.php#L613-L620

I would like to suggest to dispatch the event **after** updating the `composer.lock` file instead. This allows event listeners to assume a clean and complete package state.

Background: I would like to rewrite an existing Shell helper script which automatically performs all pending package updates one by one and does a Git commit after each single update. This allows for easy bisect and better revert if necessary. The users then only need to perform a `composer up` and rest assured that there's a Git commit for every single update performed. However, if the `composer.lock` was not written at the time my `post-package-update` listener is called, I cannot commit the package update.

mbrodala:
Fixes #7765

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -610,13 +610,13 @@ class Installer
                 }
             }
 
+            if ($this->executeOperations || $this->writeLock) {
+                $localRepo->write();
+            }
+
             $event = 'Composer\Installer\PackageEvents::POST_PACKAGE_'.strtoupper($jobType);
             if (defined($event) && $this->runScripts) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Seems reasonable to me, but obviously it's hard to imagine all possible use cases here.. I guess we try it on the snapshots for a while see if anyone complains.

mbrodala:
Thanks a lot.

PR #7778, opened 2018-11-09
Fixed Typo: Wrong folder name for Unix Systems

zf2timo:
The documentation says, the cache folder for unix systems is `$home/cache` but the correct path is `$home/.cache`.


Seldaek:
$home in this variable is the composer home, so ~/.composer/cache, which is correct. ~/.cache is only for XDG distros which are putting it in XDG_CACHE_DIR which is ~/.cache.

PR #7798, opened 2018-11-19
Add support for running composer with phpdbg

mlocati:
In order to run PHPUnit, it's very handy to add a section like this in composer.json:
```json
{
    "require-dev": {
        "phpunit/phpunit": "whatever"
    },
    "scripts": {
        "test": "phpunit"
    }
}
```

so that we can invoke PHPUnit simply by running

```sh
composer run-script test
```

This is currently working great.

We have one problem though: when we want to collect code coverage (`composer run-script test -- --coverage-clover=coverage.xml`), we can use xdebug or phpdbg.

Using xdebug works fine.

BTW using phpdbg doesn't work, since `PhpExecutableFinder::find()` returns `'/usr/bin/phpdbg -rrq'` (note the extra argument), and composer calls `ProcessExecutor::escape()` on it, so we'll try to execute the command `/usr/bin/phpdbg -qrr`, which does not exist (what exists is `/usr/bin/phpdbg`).

So, let's just escape `/usr/bin/phpdbg` (the result of `PhpExecutableFinder::find(false)`), and add optional php executable arguments ( `PhpExecutableFinder::findArguments()`) in a separate way.

### DIFF
--- a/bin/composer
+++ b/bin/composer
@@ -1,7 +1,7 @@
 #!/usr/bin/env php
 <?php
 
-if (PHP_SAPI !== 'cli') {
+if (PHP_SAPI !== 'cli' && PHP_SAPI !== 'phpdbg') {
     echo 'Warning: Composer should be invoked via the CLI version of PHP, not the '.PHP_SAPI.' SAPI'.PHP_EOL;
 }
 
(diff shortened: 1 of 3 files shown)


Seldaek:
Thanks

PR #7803, opened 2018-11-24
Coding style fixes

TomasVotruba:
This is just basic code cleanup:

- remove unused imports
- sort imports a-z
- remove unused variables
- remove unused ()

Let me know if you'd like to put there into coding standard setup, so they can be check by CI.


localheinz:
@TomasVotruba 

Maybe enable a few more fixers in [`.php_cs`](https://github.com/composer/composer/blob/master/.php_cs)?

TomasVotruba:
I'll update the ruleset with what's possible.

I can't find the command that checks coding stadnard. I looked to `.travis.yml` and `composer.json`. Where is it?

localheinz:
@TomasVotruba 

I think it’s not integrated yet, and only run occasionally. I’m 👍 for an integration into CI.

TomasVotruba:
@localheinz It's like having tests without phpunit run in Travis CI :D

TomasVotruba:
Why closing the ruleset update then?

ntzm:
I think it's worth mentioning this in a contributing document then

Seldaek:
Yes I was thinking of adding this to CONTRIBUTING.md as that would hopefully save some people from wasting their time on this.

shalvah:
@seldaek what do you think of a bot that would automatically make this explanation to any new coding style PRs and then close them? I can do that.

TomasVotruba:
What's so too far on `ordered_imports`? It's easier to read the code, when there is some order.

> Also, I have had to explain this 15 times every time some new person gets excited about fixing CS in every project out there, and they always wonder why because from their point of view it's the right thing to do. So please don't take it too personally, but it is exhausting from my perspective.

No troubles. It's just confusing for me to see `.php_cs` in this project with kind of contradictory attitude. If there was none, I'd not have to waste our both times.

> Yes I was thinking of adding this to CONTRIBUTING.md as that would hopefully save some people from wasting their time on this.

I don't expect people to look to `CONTIRBUTING.md` to find out if coding standard is used only partially.

TomasVotruba:
I think you missunderstood me, I don't ask about adding these code style changes.
I asked about **the `php_cs` ruleset**.

I don't mind dropping the changes and I understand that :+1:

Seldaek:
I understood you, butI don't feel like debating in depth the value of every single one of PHP CS Fixer's million options. I feel like what we have is enough, if you disagree I am sorry but that's how it is.

@shalvah thanks but I don't think that will be necessary, we don't get that many of them.

Anyway locking this now and going back to work..

PR #7806, opened 2018-11-27
Fix support for imagemagick <3.3

meyerbaptiste:
This PR is an alternative to https://github.com/composer/composer/commit/489e0d4b124905a26310c9026d83eb82eb0e57ce (which introduced another bug).
See #7762:

> This regex is not correct and breaks Composer with PHP Imagick < `3.3.0`:
> * Before `3.3.0`:
> ```C
> php_info_print_table_row(2, "ImageMagick version", MagickGetVersion(&version_number));
> ```
> * Since `3.3.0`:
> ```C
> php_info_print_table_row(2, "Imagick using ImageMagick library version", MagickGetVersion(&version_number));
> ```

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -158,15 +158,10 @@ class PlatformRepository extends ArrayRepository
                     break;
 
                 case 'imagick':
-                    $reflector = new \ReflectionExtension('imagick');
-
-                    ob_start();
-                    $reflector->info();
-                    $output = ob_get_clean();
-
-                    preg_match('/^(Imagick using ImageMagick library version|ImageMagick version) => ImageMagick ([\d.]+)-(\d+)/m', $output, $matches);
+                    $imagick = new \Imagick();
+                    $imageMagickVersion = $imagick->getVersion();
+                    preg_match('/^ImageMagick ([\d.]+)-(\d+)/', $imageMagickVersion['versionString'], $matches);
                     $prettyVersion = "{$matches[1]}.{$matches[2]}";
-
                     break;
 
                 case 'libxml':


Seldaek:
Sorry about the messed up fix, that's what I get for rushing things.. Anyway thanks this looks much better.

src/Composer/Repository/PlatformRepository.php
```diff
@@ -158,15 +158,10 @@ protected function initialize()
...
                 case 'imagick':
-                    $reflector = new \ReflectionExtension('imagick');
-
-                    ob_start();
-                    $reflector->info();
-                    $output = ob_get_clean();
-
-                    preg_match('/^(Imagick using ImageMagick library version|ImageMagick version) => ImageMagick ([\d.]+)-(\d+)/m', $output, $matches);
+                    $imagick = new \Imagick();
+                    $imageMagickVersion = $imagick->getVersion();
```
meyerbaptiste: The documentation says that this method is static but it has not always been the case: https://github.com/mkoppanen/imagick/commit/6958c0160d6a7158cbbf7b11ec0779178748561c#diff-372a08f4bac0eb82b7f5fe3f8cf2975dR2230

I my case:
>   [ErrorException]                                                                                                   
>  Non-static method Imagick::getversion() should not be called statically, assuming $this from incompatible context 

Fortunately both `ZEND_ACC_ALLOW_STATIC` and `ZEND_ACC_PUBLIC` are valid now.

meyerbaptiste:
No worries, thanks @Seldaek! ❤️

glensc:
thanks for finding [Imagick::getVersion](http://php.net/manual/en/imagick.getversion.php), in [original PR#7762](https://github.com/composer/composer/pull/7762) i only checked reflection method, and it was not returning value.

Issue #7795, opened 2018-11-17
Suggestion, ask confirmation when is run as admin


PeopleInside:
Hi, sometimes i forget composer should be run as user and not admin.
Instead also to advise to run as user can you add a confirmation message that ask to the user if are sure to run as admin?

Actually seems if is lunched as admin just alert and lunch without give the possibility to stop.
I will imagine this:

- run as admin -> composer show the alert that should be run as user and ask if you want continue ( y / n ).

This can be a security feature that allow you to say n (no) than switch to user and run as user.

peter279k:
@PeopleInside, do you want to have the `prompt` to confirm that user run [composer](url) as super user?

According to the original approach, this will only output the warning message to notify the user run `composer` as super user.

I also agree this feature because this will let users do the confirmation again.

PeopleInside:
@peter279k exact. I run as root for error and was not able to stop... alert is not enough.
If the system was asking if continue as root i would choose not and choose the normal user :)

Seldaek:
Fixed by #7819

PR #7819, opened 2018-12-04
Ask confirmation when is run as admin

yassine-ah:
Following #7795 suggestion, 
I added a confirmation question before continue running Composer as root/super user.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -210,7 +210,7 @@ class Application extends BaseApplication
                         
                         if ($io->isInteractive()) {
                             if (!$io->askConfirmation('<info>Continue as root/super user</info> [<comment>yes</comment>]? ', true)) {
-                                return 0;
+                                return 1;
                             }
                         }
                     }
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -210,7 +210,7 @@ class Application extends BaseApplication
                         
                         if ($io->isInteractive()) {
                             if (!$io->askConfirmation('<info>Continue as root/super user</info> [<comment>yes</comment>]? ', true)) {
-                                exit(0);
+                                return 0;
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


src/Composer/Console/Application.php
```diff
@@ -207,6 +207,12 @@ function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknow
                 if (function_exists('posix_getuid') && posix_getuid() === 0) {
                     if ($commandName !== 'self-update' && $commandName !== 'selfupdate') {
                         $io->writeError('<warning>Do not run Composer as root/super user! See https://getcomposer.org/root for details</warning>');
+                        
+                        if ($io->isInteractive()) {
+                            if (!$io->askConfirmation('<info>Continue as root/super user</info> [<comment>yes</comment>]? ', true)) {
+                                exit(0);
```
stof: you should return here, rather than killing the process, to let the Application run cleanly (and for testing purposes, it is possible to ask it to never exit the process for instance, but such code breaks it and would exit the whole testsuite if it was reached by such a test)

yassine-ah: You are absolutely right, my mistake, I didn't thought about that.

src/Composer/Console/Application.php
```diff
@@ -210,7 +210,7 @@ function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknow
                         
                         if ($io->isInteractive()) {
                             if (!$io->askConfirmation('<info>Continue as root/super user</info> [<comment>yes</comment>]? ', true)) {
-                                exit(0);
+                                return 0;
```
stof: I think it should actually return `1`, as the command could not run successfully.

yassine-ah: About that, I am not very sure, 1 is when an error occured and it is not our case, no?.

stof: well, from the view of the caller script, which expected a `composer install` to run, it is.

alcohol: I think returning `1` here would make sense yeah, even though the user might have intentionally aborted the operation it is still a failure state.

Seldaek:
Thanks

PR #7830, opened 2018-12-11
Pool Builder: don't add identical packages (same name, version, stability, require, provide, replace, conflict)

naderman:
For now this improves performance with present lock file and breaks tests which check if packages get updated to remote package. After further installer refactoring this will lead to correct updating behaviour (lock packages will always be lowest priority) and improve updates with lock file present.

Current improvement on a Sylius project with vendor directory present: 0.5% reduction in rules (= 15k rules), 185 package objects less analyzed (~= number of installed packages), 8MB memory reduced (=0.4%).

For now this breaks tests verifying installed packages get updated to newest metadata on remote end. After refactoring the installer to always pick remote packages as highest priority we can drop any special code to update installed packages and this patch will take care of ensuring the right package data is used.

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -32,6 +32,7 @@ class PoolBuilder
     private $filterRequires;
     private $rootAliases;
 
+    private $identityMap = array();
     private $aliasMap = array();
     private $nameConstraints = array();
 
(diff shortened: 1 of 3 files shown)


src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -133,11 +134,12 @@ public function buildPool(array $repositories, array $rootAliases, Request $requ
             }
         }
 
-        $this->pool->setPackages($this->packages, $this->priorities);
-
         unset($this->aliasMap);
         unset($this->loadedNames);
         unset($this->nameConstraints);
+        unset($this->identityMap);
```
stof: you should reset it to an empty array rather than unseting the property itself (which requires PHP to remember that this instance does not have the property belonging to the class definition anymore)

stof: the same is true for other properties too of course

naderman:
No longer necessary with other optimizations:

- the lock repo is only used if package cannot be found remotely
- repo priority is now dealt with first, so identical packages would only matter on the same repo, which is not really something that happens.

PR #7833, opened 2018-12-13
Allows the JsonFile class to be used for other json files than composer.json

Duregar:
Normally, there should be no side effects.
Unit tests work identically.
The goal is to allow the JsonFile class to be more easily used by any composer plugin.


Seldaek:
Thanks

PR #7835, opened 2018-12-13
Fix htaccess-protect problems

bugreportuser:
htaccess-protect doesn't work in global config because the option is checked before the config is read.

This also fixes the parsing for htaccess-protect. It's parsed as a string but should be parsed as a bool.


Seldaek:
Thanks, if you could rebase on the 1.8 branch though even better, but otherwise I'll do this myself when merging.

Seldaek:
Thanks!

PR #7851, opened 2018-12-21
MB to MiB

michaelKaefer:
I did not study computer science, so correct me if I'm wrong. But I think you are calculating mebibyte (MiB) not megabyte (MB). Megabyte would be:
... round($valueInByte / 1000 / 1000, 2).'MB ...
Or is there some specific standard you follow? According to https://en.wikipedia.org/wiki/Binary_prefix both calculations (yours and mine) are correct in a way but I find yours to be not completely clear.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -262,7 +262,7 @@ class Application extends BaseApplication
             }
 
             if (isset($startTime)) {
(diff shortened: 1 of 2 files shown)


michaelKaefer:
https://en.wikipedia.org/wiki/Megabyte

https://cseducators.stackexchange.com/questions/4425/should-i-teach-that-1-kb-1024-bytes-or-1000-bytes?newreg=d344a0dd70fc414ea42d01b1241f34a0

Seldaek:
Thanks, merged in 1.8/master as https://github.com/composer/composer/commit/4765a8f21b38b7ee65968569a00fa96d00f9d2f9

PR #7853, opened 2018-12-24
Fix: Keep environment variables sorted by name

localheinz:
This PR

* [x] keeps the list of environment variables on https://getcomposer.org/doc/03-cli.md sorted

💁‍♂️ Probably not so nice to review, but unless there are reasons for the environment variables *not* to be sorted I am unaware of, I think there are no reasons not to sort them.


localheinz:
Thank you, @Seldaek!

PR #7856, opened 2018-12-26
Use Ubuntu 16.04

carusogabriel:
Ubuntu 16.04 is available for usage: https://docs.travis-ci.com/user/reference/xenial/#using-xenial


carusogabriel:
Not valid as Xenial dist only has the >=7.1 PHP versions 😕

PR #7858, opened 2018-12-26
Updating lock file hash

chr-hertel:
while working on #7857 i found this little fellow.
might be caused by https://github.com/composer/composer/commit/4301c19ce3b2468ad841ca9f5f35ca5558fb005f#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780


Seldaek:
Thanks

PR #7861, opened 2019-01-02
Enhancement: Add .editorconfig

localheinz:
This PR

* [x] adds an `.editorconfig`

💁‍♂️ For reference, see https://editorconfig.org.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.yml]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+
+[*]
+charset = utf-8
+indent_size = 4
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
Seldaek: Why not? I don't see where this would make sense.

localheinz: Adjusted!

localheinz:
Thank you, @Seldaek!

PR #7863, opened 2019-01-02
Enhancement: Reference phpunit.xsd as installed with composer

localheinz:
This PR

* [x] references `phpunit.xsd` as installed with `composer`
* [x] removes a non-existent attribute from `phpunit.xml.dist`


localheinz:
Thank you, @Seldaek!

PR #7867, opened 2019-01-03
fix(application): use precise helper set

fancyweb:
Currently, the Composer console Application relies on the default helper set of the Symfony Console Component. However, only one is used internally (the Question one).
I think it would be better to require it directly like this.
It is also better for the `BuffferIO` class (cf my next PR https://github.com/composer/composer/pull/7868)


Seldaek:
Thanks

PR #7869, opened 2019-01-03
feat(buffer-io): add the possibility to set user inputs for interactive questions

fancyweb:
This feature makes it possible to easily use the `BufferIO` class to test a Composer command that have interactions.
I did it in a non BC way but I would to work on a new version of this class for the v2.
Inspired by the Symfony `CommandTester` class.
Needs https://github.com/composer/composer/pull/7868.

### DIFF
--- a/src/Composer/IO/BufferIO.php
+++ b/src/Composer/IO/BufferIO.php
@@ -14,6 +14,7 @@ namespace Composer\IO;
 
 use Symfony\Component\Console\Output\StreamOutput;
 use Symfony\Component\Console\Formatter\OutputFormatterInterface;
+use Symfony\Component\Console\Input\StreamableInputInterface;
 use Symfony\Component\Console\Input\StringInput;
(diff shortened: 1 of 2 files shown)


Seldaek:
Hey, not sure what you mean that this is not BC? Is it because it requires Symfony 3.2? That's fine by me as it's an optional feature for people integrating Composer.. If there is something that actually breaks BC and that I miss let me know :)

fancyweb:
What I mean is that my proposed implementation should be backward compatible (since it's a new method). I did not put an extra argument in the constructor in case people are already extending this. But then it's not really logical to specify some things in the constructor and others in another method. 

Also I think that the helper set should be shared somewhere with the one used in the Application so they stay consistent.

Seldaek:
Ok so I will merge this in master.. Then if you want to submit a cleanup PR to the 2.0 branches feel free.

PR #7880, opened 2019-01-07
-no-cache argument for "composer install" and "composer update"

### LINKED ISSUE #6650: Argument to Disable Cache
I would like to revive #1896, because there is one important use-case that was not mentioned there: building of Docker images. The cache is just bloating the size of the images, and not required at all in production. Currently one has to clean the cache manually, it would be much better to simply have a `--no-cache` argument, and be done with it.

Dzhuneyt:
Fixes #6650 and https://github.com/saltstack/salt/issues/45093

However, I was only able to test it under Unix. If someone can check the solution under Windows - that'd be great. Tests run green.

### DIFF
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -15,6 +15,7 @@ namespace Composer\Command;
 use Composer\Installer;
 use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
+use Composer\Util\Platform;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, see https://github.com/composer/composer/commit/02ceb74151eb027cf74c17059dfb80a1af1a088d tho for some review/changes.

mnapoli:
This is awesome!

However it creates this exception when there is a dependency using a `git` URL:

> GitDriver requires a usable cache directory, and it looks like you set it to be disabled

Is there any way that the cache is still used when it's absolutely necessary? That would avoid the exception while still storing as little as possible on disk.

Seldaek:
If you only care about disk space I think it'd be easiest to run it normally and then wipe the cache dir after your build is done.

mnapoli:
🤔  that's a good idea. I run Composer (and php-cs-fixer for https://prettyci.com/) on AWS Lambda, where I only have limited space in the `/tmp` folder. That would work, unless there's more of 500mb of composer cache, but I don't think that happens often! Thanks!

Seldaek:
500mb sounds like a lot for a single install or update cycle.

PR #7886, opened 2019-01-10
add compile script call to composer.json

kaznovac:
expose call for compiling `composer.phar` in composer json, along with description


Seldaek:
Thanks

Issue #7890, opened 2019-01-13
Wrong CWD with PHP ZTS during create-project


Sweetchuck:
My `composer.json`:
The whole project is in a private Git repository, and uses a package which also hosted in a private repository. (not the private repository the problem)
```json
{
  "name": "my/template-01",
  "repositories": {
    "foo/bar": {
      "type": "git",
      "url": "git@example.com:foo/bar.git"
    }
  },
  "require": {
    "foo/bar": "dev-something"
  }
}
```

Output of `composer diagnose`:

```
PHP version: 7.2.12 with ZTS !
```

When I run this command:

```
cd /a/b
composer create-project my/template-01 /a/b/instance-01
```

I get the following output:

```
  - Installing foo/bar (dev-something abcdefg): Cloning abcdefg from cache
                                                                                                                                                  
  [RuntimeException]                                                                                                                              
  Failed to execute git checkout 'abcdefg_LONG_HASH' -- && git reset --hard 'abcdefg_LONG_HASH' --  
                                                                                                                                                  
  fatal: not a git repository (or any parent up to mount point /)                                                                              
  Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set). 
```

And I expected this to happen:
Gather all the dependencies without any problem :-)

I have PHP 7.2.12 with NTS and ZTS as well.
This problem only occurs when I use the one with ZTS.

The `composer create-project` differs from the `composer install` command, because in the terminal the `$PWD` is different than the new project directory will be. Most likely the new project directory not even exists at the moment when you run the `composer create-project` command.

In the `\Composer\Downloader\GitDownloader::updateToCommit()` the `$path` contains a path which is relative from the PHP `getcwd()`, but the PHP `getcwd()` is different from the current directory used in the `$this->process->execute()`.

When I add the following lines:
```php
var_dump('getcwd() = ' . getcwd());
var_dump("\$path = '$path'");
var_dump('`pwd` = ' . trim(`pwd`));

$this->process->execute('pwd', $output);
var_dump('$this->process->execute("pwd") = ' . trim($output));

$this->process->execute('ls -la', $output, $path);
var_dump('$this->process->execute("ls -la") = ' . trim($output));
```
The output is:
```
"getcwd() = /a/b/instance-01"
"$path = 'vendor/foo/bar'"
"`pwd` = /a/b/instance-01"
"$this->process->execute("pwd") = /a/b/instance-01"
"$this->process->execute("ls -la") = total 0
[...]

PR #7891, opened 2019-01-13
Issue #7890 - GitDwnloader CWD

Sweetchuck:
Issue #7890


alcohol:
If the Process class contains a bug, it should be solved on their end. What you propose now looks like a work-around. This is not a desirable solution.

PR #7896, opened 2019-01-17
Added -dry-run option to "composer remove" command

### LINKED ISSUE #7793: add a --dry-run option to the 'remove' command
The install and update commands have a --dry-run option, but the remove command does not.

Dzhuneyt:
Fixes #7793

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -234,6 +234,7 @@ uninstalled.
 
 ### Options
 * **--dev:** Remove packages from `require-dev`.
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks but this skips the modifications to composer.json which means the update that follows won't actually update anything and therefore it isn't going to show the same output as if you run it without --dry-run. I don't think this is an appropriate fix, and I don't really think it's going to be possible to do a clean fix for this right now. Maybe in 2.0 we can do this more cleanly after some of the refactorings that are ongoing.

PR #7897, opened 2019-01-17
Archive: cleanup temp dir on download error

glaubinix:
Composer will not cleanup the temporary directory created in case the archive manager throws an exception while downloading a version from a git repository e.g.
```
Failed to execute git checkout '18313f04cea0e078412a028c5361bd4e' -- && git reset --hard '18313f04cea0e078412a028c5361bd4e' --

fatal: invalid reference: 18313f04cea0e078412a028c5361bd4e
```

This starts consuming quite a lot of disk space if this happens while trying to install a 200MB repo like wordpress several times.

### DIFF
--- a/src/Composer/Package/Archiver/ArchiveManager.php
+++ b/src/Composer/Package/Archiver/ArchiveManager.php
@@ -147,8 +147,13 @@ class ArchiveManager
             $sourcePath = sys_get_temp_dir().'/composer_archive'.uniqid();
             $filesystem->ensureDirectoryExists($sourcePath);
 
-            // Download sources
-            $this->downloadManager->download($package, $sourcePath);
+            try {
+                // Download sources
+                $this->downloadManager->download($package, $sourcePath);
+            } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -147,8 +147,13 @@ public function archive(PackageInterface $package, $format, $targetDir, $fileNam
             $sourcePath = sys_get_temp_dir().'/composer_archive'.uniqid();
             $filesystem->ensureDirectoryExists($sourcePath);
 
-            // Download sources
-            $this->downloadManager->download($package, $sourcePath);
+            try {
+                // Download sources
+                $this->downloadManager->download($package, $sourcePath);
+            } catch (\RuntimeException $e) {
```
stof: I suggest catching any `\Exception` here

glaubinix: done

Seldaek:
Thanks

Issue #7817, opened 2018-12-03
Dependencies not mapped correctly


tbmatuka:
My `composer.json`:

```json
{
    "require": {
        "symfony/security-bundle": "^4.2"
    },
    "require-dev": {
        "symfony/security": "^4.2"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 1.8.0
PHP version: 7.2.9
PHP binary path: /usr/bin/php
```

When I run this command with no other files or folders in the dir:
```
composer update --no-suggest
```

I get this output:
```
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 16 installs, 0 updates, 0 removals
  - Installing symfony/contracts (v1.0.1): Loading from cache
  - Installing symfony/polyfill-ctype (v1.10.0): Loading from cache
  - Installing symfony/inflector (v4.2.0): Loading from cache
  - Installing symfony/property-access (v4.2.0): Loading from cache
  - Installing symfony/polyfill-mbstring (v1.10.0): Loading from cache
  - Installing symfony/http-foundation (v4.2.0): Loading from cache
  - Installing symfony/event-dispatcher (v4.2.0): Loading from cache
  - Installing psr/log (1.1.0): Loading from cache
  - Installing symfony/debug (v4.2.0): Loading from cache
  - Installing symfony/http-kernel (v4.2.0): Loading from cache
  - Installing symfony/security (v4.2.0): Loading from cache
  - Installing psr/container (1.0.0): Loading from cache
  - Installing symfony/dependency-injection (v4.2.0): Loading from cache
  - Installing symfony/filesystem (v4.2.0): Loading from cache
  - Installing symfony/config (v4.2.0): Loading from cache
  - Installing symfony/security-bundle (v4.2.0): Loading from cache
[...]

PR #7900, opened 2019-01-17
Fix: Keep replaced packages for autoload dumping with --no-dev

Great-Antique:
Fixes: #7817

### DIFF
--- a/tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
+++ b/tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
@@ -14,6 +14,7 @@ namespace Composer\Test\Autoload;
 
 use Composer\Autoload\AutoloadGenerator;
 use Composer\Package\Link;
+use Composer\Semver\Constraint\Constraint;
 use Composer\Util\Filesystem;
 use Composer\Package\AliasPackage;
 use Composer\Package\Package;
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
```diff
@@ -419,6 +420,36 @@ public function testNonDevAutoloadExclusionWithRecursion()
...
+            ->will($this->returnValue($packages));
+
+        $this->fs->ensureDirectoryExists($this->vendorDir.'/b/b/src/C');
+        file_put_contents($this->vendorDir.'/b/b/src/C/C.php', '<?php namespace B\\C; class C {}');
+
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', false, '_5');
+
+        include $this->vendorDir.'/autoload.php';
+
+        $this->assertTrue(\class_exists('B\\C\\C'));
```
stof: rather than creating a class and registering a fake autoloader supporting it (which will be global to the whole testsuite starting from that point), you would rather assert that `$this->vendorDir.'/composer/autoload_classmap.php'` contains the class (with an optimized dump).

Great-Antique: @stof changed

Seldaek:
Thanks!

Issue #7903, opened 2019-01-18
Extract download method out of install in VcsDownloaders


Seldaek:
This will ensure all network ops are done before any filesystem is touched. Also need to make sure that COMPOSER_DISABLE_NETWORK is respected where possible.

Also try to fix #7929

PR #7904, opened 2019-01-18
Parallel downloads 

Seldaek:
- Adds parallel downloads when ext-curl is present, including HTTP/2 and all niceties, much faster. (Closes https://github.com/composer/composer/pull/1628 https://github.com/composer/composer/pull/5293 https://github.com/composer/composer/pull/7197)
- Makes sure that all network operations are complete before the filesystem is touched, which should mean in theory we never end up with a borked vendor dir due to a failed download. Still need to follow up with https://github.com/composer/composer/issues/7903 tho. And also there is a dependency on another ongoing refactoring so packages are still done one by one now, but this will be fixed.
- Introduces new metadata format which is more compact on the wire and also reduces memory usage slightly thanks to interned strings (still in development, DO NOT RELY ON THIS UNTIL 2.0 stable) - further improvement might come with https://github.com/composer/composer/issues/6415 which I still have to look at but it sounds sensible to maximize cache hits. e.g. https://repo.packagist.org/p2/symfony/symfony.json is 23KB gzipped vs 37KB before (uncompressed size went from 1.3MB to 243KB)
- JSON package listings (provider-*) have been dropped as well which means more requests are done to verify cache integrity but most of them return a 304 Not Modified very quickly so hopefully this will be a net positive.
- Some memory optimization in the way packages are loaded, might save a few percent on large projects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "symfony/filesystem": "^2.7 || ^3.0 || ^4.0",
         "symfony/finder": "^2.7 || ^3.0 || ^4.0",
         "symfony/process": "^2.7 || ^3.0 || ^4.0",
-        "react/promise": "^1.2"
+        "react/promise": "^1.2 || ^2.7"
     },
     "conflict": {
         "symfony/console": "2.8.38"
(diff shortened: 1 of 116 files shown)


bozerkins:
This also fixes #704

PR #7906, opened 2019-01-19
Cleanup of old compression algo

Seldaek:
Follow-up to https://github.com/composer/composer/pull/7904 - needs all packages to have been re-dumped though before it can be merged.

### DIFF
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -62,39 +62,16 @@ class ArrayLoader implements LoaderInterface
 
     public function loadPackages(array $versions, $class)
     {
-        static $uniqKeys = array('version', 'version_normalized', 'source', 'dist', 'time');
-
         $packages = array();
         $linkCache = array();
 
         foreach ($versions as $version) {
-            if (isset($version['versions'])) {
-                $baseVersion = $version;
-                foreach ($uniqKeys as $key) {
-                    unset($baseVersion[$key.'s']);
-                }
+            $package = $this->createObject($version, $class);
 
-                foreach ($version['versions'] as $index => $dummy) {
-                    $unpackedVersion = $baseVersion;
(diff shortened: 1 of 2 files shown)


stof:
@Seldaek idea to reduce the file size for packagist files: skip the dumping of `require-dev` constraint. Composer only cares about the dev requirements of the root package, and the metadata of the root package are available locally, not from Packagist.

Packagist already strips the `support` info of the package, which are actually more useful for composer (as `composer show` is then unable to display the support info)

stof:
this PR should be rebased so that the diff does not include #7904

stof:
@Seldaek idea for the new minification format: instead of dealing only with top-level keys, we could replicate the same algo on the second level for keys defining constraint (`require`, `conflict`, etc...). this way, adding a new requirement or changing a min version for a requirement does not need to include all other requirements again.

naderman:
I would actually prefer we didn't strip any info, neither require-dev nor support, so they do remain available for inspection, I think this kind of mechanism should save enough space already. I do agree that it may be worth to investigate doing this on a second level as well.

Seldaek:
@stof I tried applying the same diff algo to require/... - it doesn't help much, after gzip is done with the file it's hardly noticeable so I think it's best to leave it as is to avoid even more complexity and chance of bugs.

stof:
well, if it does not make difference after gzip, this is fine then. That was just an idea.

Seldaek:
Yup it was worth a shot

staabm:
Regarding content-encoding: would it be worth to try brotli or similar more modern compression ?

Seldaek:
I'd love to but nginx makes it a PITA so it's still on the backburner. Anyway this can be added later on without BC break so not critical for 2.0

staabm:
Agree - its not critical for 2.0.
Side note: maybe packagist could use a cdn like cloudflare or similar and let this kind of feature be handled by the cdn?

PR #7909, opened 2019-01-20
Add alias of run-script command

xuanquynh:
The `run` command looks like more pretty than `run-script` but it's still meaningful. Let me add an alias for the "run-script" command.

```bash
composer run post-root-package-install # instead of "composer run-script post-root-package-install"
composer run test # instead of "composer run-script test"
```


Seldaek:
Thanks

PR #7912, opened 2019-01-22
Feature/no api config

kathyoungaligent:
Allowing for the no-api flag to be used as a config value for all Github repos, not just on a repo by repo basis.

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -236,11 +236,11 @@ This is used if you use the `gitlab` repository type.
 
 ## use-github-api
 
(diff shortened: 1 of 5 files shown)


alcohol:
Could you also update `doc/06-config.md` and the following file?

https://github.com/composer/composer/blob/0fca816d8f0a12775cfa567ffea793ff72eb1311/res/composer-schema.json#L97-L291

kathyoungaligent:
@alcohol  - I will update the docs files shortly, thanks for pointing that out.

@Seldaek  - thank you for considering this feature and yes, you're right, it should be more self descriptive.  I'll rename the config.

mrda:
Thanks for proposing this @kathyoungaligent - this will help a lot.

Seldaek:
Thanks!

PR #7925, opened 2019-01-28
Fix: Remove empty node

localheinz:
This PR

* [x] adjusts `JsonConfigSource::removeLink()` to remove an empty node


localheinz:
Thank you, @Seldaek!

PR #7932, opened 2019-01-29
Installer: Initialize $nameMatchesRequiredPackage

hco:
We had a case where we got `Undefined variable: nameMatchesRequiredPackage` — I think it should be initialized with false, right?

I am currently not able to create a reproduction case - if you need one before merging, please let me know and I will do my best. But just from looking at the code, I think it can happen that the variable is undefined.


Seldaek:
That makes sense but please rebase on 1.8 branch for bugfixes

hco:
Sorry, will do.

hco:
So that change did make stuff worse, sorry. Give me a few minutes :-)

stof:
you should not only change the target branch of the PR. You also need to rebase your feature branch so that it is based on 1.8 rather than master and to force-push the branch to your github fork.

hco:
Think I did it now :-)

Issue #7776, opened 2018-11-06
SHA384 is not supported by your openssl extension


jreason:
By the looks of things, this has been reported before and fixed.
However this now seems to be an issue again.

Container build is failing

``ARG COMPOSER_CHECKSUM=55d6ead61b29c7bdee5cccfb50076874187bd9f21f65d8991d46ec5cc90518f447387fb9f76ebae1fbbacf329e583e30``
``ARG COMPOSER_URL=https://raw.githubusercontent.com/composer/getcomposer.org/ba0141a67b9bd1733409b71c28973f7901db201d/web/installer``

``Step 14 : RUN curl -o installer "$COMPOSER_URL" && echo "$COMPOSER_CHECKSUM *installer" | shasum -c -a 384 && php installer --install-dir=/usr/local/bin --filename=composer && rm -rf /var/lib/apt/lists/*
 ---> Running in 48c59be3fef8
·[91m % Total % Received % Xferd Average Speed Time Time Time Current
 Dload Upload Total Spent Left Speed
 0 0 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0·[0m·[91m
100 298k 100 298k 0 0 474k 0 --:--:-- --:--:-- --:--:-- 475k
·[0minstaller: OK``
``All settings correct for using Composer``
``SHA384 is not supported by your openssl extension``
``The command '/bin/sh -c curl -o installer "$COMPOSER_URL" && echo "$COMPOSER_CHECKSUM *installer" | shasum -c -a 384 && php installer --install-dir=/usr/local/bin --filename=composer && rm -rf /var/lib/apt/lists/*' returned a non-zero code: 1``

Seldaek:
https://raw.githubusercontent.com/composer/getcomposer.org/ba0141a67b9bd1733409b71c28973f7901db201d/web/installer has an old commit hash in the URL and is not the latest installer. Please use the latest code before reporting bugs.. Especially as you mention you've seen it's been fixed.

PR #7934, opened 2019-01-29
Update installer commit to fix SHA384 bug

IkeLutra:
Installer that is listed in the documentation is affected by #7776.

The newest copy is no longer affected but have updated the link in the docs to prevent people trying to install a broken version.


Seldaek:
Thanks

PR #7936, opened 2019-01-30
Separate Install & Update code, no longer use vendor dir as input to solver

naderman:
This splits the current installer into two clear separate operations: update & install.

That means if you run composer update you will first get a process modifying the lock file with corresponding output, then the install step will run which you can trigger by simply running composer install as well which will synchronize the state of the vendor directory with the lock file.

As a consequence a fair amount of output has been adjusted. Solver inputs have been simplified to only use the lock file as input in partial updates. Otherwise we fully rely on remote package information. This makes sure that we always get current metadata, updates do not depende on local system state and we have a smaller amount of package versions to analyze in the solver.

There are a number of TODO items added in this PR which will have to be separately addressed before the first 2.0 release.


stof:
> no longer use vendor dir as input to solver

the lock file is still used as input though, right ?

naderman:
Yes

stof:
@naderman any progress on this ?

naderman:
@stof working on it ... I'm not going to hide the results when I have them so not sure why you ask

naderman:
Based on a conversation with @Seldaek we had decided that we should change the behavior of URL updating for packages which are not updated to remain unmodified in the lock file. I believe presently this breaks the functionality of composer update mirrors, so we will need a special mode for this feature to replace all metadata without actually updating.

naderman:
Should probably consider writing a new SolverTest case from scratch since this modified version probably has some pretty nonsensible tests now that installed packages aren't input anymore and that we don't really mark updates or removes, but really just fix packages, ask for them to be installed or expect them to be updated.

naderman:
Considering that I don't want to hold up further work on the 2.0 branch and that the branch is considered unstable anyway I'll merge this PR now and address all the TODOs I added in this PR in separate PRs. Additionally code cleanup any missing tests will also happen on the 2.0 branch then.

PR #7941, opened 2019-01-31
Fix mode bitmask when detecting a Windows junction

johnstevenson:
This implements the POSIX `S_ISDIR` macro to determine if the `st_mode` value (returned by the stat functions) is a directory.

Unfortunately it doesn't always fix failing junction tests on PHP 7.3 (and some versions of PHP 7.2), but at least we've got one source of error out of the picture.

@curry684 Could do with your blessing here! Or at least a quick look.

### DIFF
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -648,6 +648,10 @@ class Filesystem
     /**
      * Returns whether the target directory is a Windows NTFS Junction.
      *
+     * We test if the path is a directory and not an ordinary link, then check
+     * that the mode value returned from lstat (which gives the status of the
+     * link itself) is not a directory.
+     *
      * @param  string $junction Path to check.
      * @return bool
      */
@@ -659,22 +663,14 @@ class Filesystem
         if (!is_dir($junction) || is_link($junction)) {
             return false;
         }
-        /**
-         * According to MSDN at https://msdn.microsoft.com/en-us/library/14h5k7ff.aspx we can detect a junction now
-         * using the 'mode' value from stat: "The _S_IFDIR bit is set if path specifies a directory; the _S_IFREG bit
-         * is set if path specifies an ordinary file or a device." We have just tested for a directory above, so if
-         * we have a directory that isn't one according to lstat(...) we must have a junction.
-         *
-         * #define	_S_IFDIR	0x4000
-         * #define	_S_IFREG	0x8000
-         *
-         * Stat cache should be cleared before to avoid accidentally reading wrong information from previous installs.
-         */
+
+        // Important to clear cache first
         clearstatcache(true, $junction);
         clearstatcache(false);
         $stat = lstat($junction);
 
-        return !($stat['mode'] & 0xC000);
+        // S_IFDIR is 0x4000, S_IFMT is the 0xF000 bitmask
+        return $stat ? 0x4000 !== ($stat['mode'] & 0xF000) : false;
     }
 
     /**
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -652,6 +652,11 @@ class Filesystem
      * that the mode value returned from lstat (which gives the status of the
      * link itself) is not a directory.
      *
+     * This relies on the fact that PHP does not set this value because there is
+     * no universal file type flag for a junction or a mount point. However a
+     * bug in PHP can cause a random value to be returned and this could result
+     * in a junction not being detected: https://bugs.php.net/bug.php?id=77552
+     *
      * @param  string $junction Path to check.
      * @return bool
      */
@@ -664,9 +669,8 @@ class Filesystem
             return false;
         }
 
-        // Important to clear cache first
+        // Important to clear all caches first
         clearstatcache(true, $junction);
-        clearstatcache(false);
         $stat = lstat($junction);
 
         // S_IFDIR is 0x4000, S_IFMT is the 0xF000 bitmask


johnstevenson:
I might be a bit closer to nailing the erratic behaviour down: it could be caused by an uninitialized structure in the Php stat code which contains the status information returned by `lstat`. This code doesn't touch the st_mode member (because a junction is not a known thing in the Posix world) so we get back whatever the system put in the structure when it was allocated.

This is normally 0 on the first run (for non-debug builds), but could be anything for subsequent allocations. This would certainly explain why I have found it so hard to reproduce: everything seems fine just calling `isJunction` from a simple script, but not always when running in a more complex environment (like testing). It could also explain the problems found in https://github.com/composer/composer/issues/4955

I've reported this here: https://bugs.php.net/bug.php?id=77552

Seldaek:
Thanks for digging into this. What's the confidence level here about merging this in the next 1.8 patch release?

johnstevenson:
@Seldaek The confidence level is higher than what we currently have. 

The `isJunction` method checks that the path being tested is a directory (`is_dir`) and not a symlink (`!is_link`) and that the `mode` value returned from `lstat` does **not** signify a directory. This works because PHP's implementation of lstat correctly identifies a directory and sets the mode value, whereas it doesn't set the mode for a junction.

Now if the path is a junction and the mode is a random value caused by the PHP bug, then isJunction will only fail if the random value & 0xF000 equals 0x4000. I don't know what the probability is of this, but it probably doesn't matter because although it will now be considered a directory, the command to delete it (via `cmd rmdir /S /Q path`) will only delete the junction.

However, it would matter on older versions of Windows, where the target directory will be removed as well. Researching this, it apparently used to happen on Windows XP, but was fixed in Windows 7, but I don't know about Vista - and this information isn't definitive.

I've fixed the PHP bug https://github.com/php/php-src/pull/3784 so it should soon get merged into PHP 7.2 and 7.3, ready for their next monthly release cycle. We could may be end up with a PHP `is_junction` function, but that might depend on me having a stab at it.

curry684:
Sorry I didn't see this before, missed the highlight. Will schedule to have a look asap when it's not 1:38am  😉

curry684:
Ok, couldn't resist as it piqued my interest hehe. Awesome work on all the debugging this issue must have taken!

The discussion at https://github.com/php/php-src/pull/3784 seems plausible to me as a potential cause for the issues occasionally experienced. I never got to look *that* far into it, and given that it doesn't occur on all versions of Windows I never had a realistic chance of reproduction. The fact that I stopped developing on Windows natively some 2 years ago didn't help.

This PR would indeed slightly decrease the chance for err, as the previous check 'naively' expected `lstat` to behave reliably with its bitmask, so I agree with merging the fix.

I've also quickly redone some of my research on the subject from back then - regrettably it would seem all the alternatives (running command line tooling etc) still require elevation, effectively disqualifying them.

My only peeve is that my huge comment with the MSDN links and quotes are being deleted. I put them there on purpose knowing that the solution was fragile despite all documentation and my own testing indicating it was reliable. Given that we are aware the solution will still remain fragile on older versions of Windows and PHP I'd prefer to leave it there and actually expand it with information about this PR and the PHP bug report.

Seldaek:
Yup, keeping all this info in the code docs sounds like a good thing indeed.

johnstevenson:
Thanks @curry684 the debugging was driving me mad at one point! Funnily enough, your first method (checking if the mode value equals zero) would have worked perfectly if it hadn't been for the PHP bug.

I moved the gist of the MSDN stuff into the doc comment because the link doesn't mention anything about junctions - it just explains Microsoft's various stat functions, which are not used by PHP anyway.

I will improve the doc comment and add the bug reference.

There is another way to check for a junction, which is to run `cmd dir /AL`, which only returns reparse points (which include symlinks) or just `cmd dir` and parse out anything that says "junction". However, I don't know which versions of Windows this works on (XP doesn't have the `/AL` switch), so it is really no better.

Seldaek:
@johnstevenson I know it's sunday so no stress if you can't but if you have time to wrap up the doc comment changes I'd be happy to merge this and get a 1.8.4 out :)

johnstevenson:
@Seldaek You're right, I should be doing other things. But I'll push the fix shortly.

curry684:
> Funnily enough, your first method (checking if the mode value equals zero) would have worked perfectly if it hadn't been for the PHP bug.

Yep, actually took me a lot of time to get to that solution in the first place hehe. MSDN's got tons of information but the closer you get to rare stuff like junctions the more it dries up. The reason I wrote the huge comment, including the quotes and the MSDN link, was that the code otherwise simply wouldn't make sense to others later on - impulsive logic says the check should always fail following the `is_link`/`is_dir` calls above it. Hence my preference to keep it all there, your change is more resilient to 
 the PHP bug but the underlying logic is still "surprise magic" from PHP's platform-agnostic view of things. It effectively states `is_dir` must be lying.

johnstevenson:
And I thought I'd done a reasonably good job of explaining it:

     * We test if the path is a directory and not an ordinary link, then check
     * that the mode value returned from lstat (which gives the status of the
     * link itself) is not a directory.
     *
     * This relies on the fact that PHP does not set this value because there is
     * no universal file type flag for a junction or a mount point. However a
     * bug in PHP can cause a random value to be returned and this could result
     * in a junction not being detected: https://bugs.php.net/bug.php?id=77552

curry684:
Yeah sorry see my last edit right before your reply, I spotted the amending commit late.

Got my blessing as requested 👍

Seldaek:
Thanks both :)

PR #7947, opened 2019-02-04
Check if file being included is a valid file

melmups:
Check if the file being included is not a directory and is indeed a valid file


alcohol:
Some background perhaps? A test maybe to indicate why this is necessary? Because the way this method is actually used, we already do a check before calling it.

Seldaek:
I don't see the point of this.. unless you have a misconfigured autoload configuration in your package, but then you should rather fix that than make the code silently ignore misconfigurations.

PR #7949, opened 2019-02-04
Make unixy proxy code POSIX compatible

johnstevenson:
The `[[` keyword is not compatible with all shells that `/bin/sh` can represent. This non-portablity is a problem when calling Composer, or stuff installed by Composer, from Ubuntu on WSL. 

See: https://github.com/composer/windows-setup/pull/90

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -196,9 +196,13 @@ class BinaryInstaller
 
 dir=\$(cd "\${0%[/\\\\]*}" > /dev/null; cd $binDir && pwd)
 
-if [ -d /proc/cygdrive ] && [[ \$(which php) == \$(readlink -n /proc/cygdrive)/* ]]; then
-   # We are in Cgywin using Windows php, so the path must be translated
-   dir=\$(cygpath -m "\$dir");
+if [ -d /proc/cygdrive ]; then
+    case \$(which php) in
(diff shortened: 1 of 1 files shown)


alcohol:
I think that is only because on Ubuntu by default they use `Dash` instead of `Bash` for `sh`.

Maybe it would be better to just change the shebang to `#!/usr/bin/env bash`.

`[[` is very useful and `bash` is available on virtually every *nix platform, even Ubuntu on WSL.

https://mywiki.wooledge.org/BashGuide/TestsAndConditionals

johnstevenson:
Yeah, I wasn't sure which way was best. I originally used the `[[` keyword based on the link you mention, and at the time WSL didn't operate across the system boundaries. So for simplicity I reckon the most portable option is best.

johnstevenson:
Ah, I've just realized that the shebang should actually be `#!/bin/sh` (which is what we use in windows-setup), since `#!/usr/bin/env sh` results in spinning up the `env` process and a Path lookup. I'll fix this.

alcohol:
Actually using `/usr/bin/env` is preferred since `/bin/bash` is not always guaranteed.

Seldaek:
Thanks - please rebase on 1.8 though if you get a chance, otherwise will do when merging.

johnstevenson:
@Seldaek  I think that worked.

Seldaek:
Thanks

PR #7952, opened 2019-02-04
Fix inconsistent casing

JakeConnors376W:


alcohol:
Yeah there are a lot more inconsistencies. While I personally would appreciate some consistency, I do think it would help to establish an agreed upon form up front.

Seldaek:
There seems to be more non-capitalized ones as capitalized ones.. so maybe let's just standardize on that. It doesn't matter a whole lot in the end.

PR #7961, opened 2019-02-07
Soften hard exit after revert of composer file

mitelg:
Introduced with https://github.com/composer/composer/pull/7555 the composer file gets reverted in an own method in an error case. The problem is the former hard exit in this method.
Supposing `$install->run()` returns an error code, the require command is not terminated correctly.
This PR makes sure, that the hard exit only happens if `pcntl_signal()` is called.


Seldaek:
Merged in 1.8 and master as eee98018f, thanks!

PR #7969, opened 2019-02-11
Remove junctions with PHP rather than system rmdir

johnstevenson:
PHP will happily remove junctions using its `rmdir` function (tested on versions back to 5.2.17). This saves invoking system `rmdir` through cmd.exe.

Note this is just a tweak (it is not crucial and is separate from the previous buggy junction issue).


Seldaek:
Wondering if this is worth risking or not.. But I guess it's fairly safe as rmdir shouldn't recursively remove anything even if it saw it as a real directory with its content.

PR #7970, opened 2019-02-11
Implementation of PECL Extension Installer

torinaki:
## Introduction
There were multiples PRs which were tried to address PECL extension installations:
1. https://github.com/composer/composer/pull/498/files
1. https://github.com/composer/composer/pull/2898/files
1. https://github.com/composer/composer/pull/3897/files

Now the situation is slightly different. The PEAR which includes PECL command code contains multiple PHP 8 compatibility issues:
1. https://github.com/pear/pear-core/pull/82
1. https://github.com/pear/pear-core/pull/85
1. https://github.com/pear/Console_Getopt/pull/3
1. ... and others

There is also an initiative to disable PEAR by default:
1. https://externals.io/message/103977
1. https://github.com/php/php-src/pull/3781

IMO the effort to support PECL in the long term will be more significant then reuse Composer code base with existing functionality. So I'm trying to to find a solution to replace PECL command with a same or similar convenient way for PHP developers to compile and install extensions.

## Scope
*Note!* Scope is topic of discussion, and it might be changed. It also depends on the required effort to implement the Extension Installer. So feature list might be cut to deliver a base implementation of Extension Installer faster.
1. Reuse [PEAR repository REST API client](https://github.com/composer/composer/tree/master/src/Composer/Repository/Pear) to communicate with PECL repositories.
1. Implement possibility to download PECL extensions source using existing Composer functionality
1. Implement extension compilation using `phpize` and `make` commands
1. Implement installation of a compiled extension
1. Download existing binaries and install extensions for Windows

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -608,6 +608,7 @@
                 { "$ref": "#/definitions/path-repository" },
                 { "$ref": "#/definitions/artifact-repository" },
                 { "$ref": "#/definitions/pear-repository" },
+                { "$ref": "#/definitions/pecl-repository" },
                 { "$ref": "#/definitions/package-repository" }
             ]
         },
@@ -673,6 +674,15 @@
             "required": ["type", "url"],
             "properties": {
                 "type": { "type": "string", "enum": ["pear"] },
+                "url": { "type": "string" },
+                "vendor-alias": { "type": "string" }
+            }
+        },
+        "pecl-repository": {
+            "type": "object",
+            "required": ["type", "url"],
+            "properties": {
+                "type": { "type": "string", "enum": ["pecl"] },
                 "url": { "type": "string" },
                 "vendor-alias": { "type": "string" }
             }
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -547,6 +547,7 @@ class Factory
         $im->addInstaller(new Installer\PearInstaller($io, $composer, 'pear-library'));
         $im->addInstaller(new Installer\PluginInstaller($io, $composer));
         $im->addInstaller(new Installer\MetapackageInstaller($io));
+        $im->addInstaller(new Installer\ExtensionInstaller($io, $composer));
     }
 
     /**
--- a/src/Composer/Installer/ExtensionInstaller.php
+++ b/src/Composer/Installer/ExtensionInstaller.php
@@ -1,0 +1,85 @@
+<?php
+
+
+namespace Composer\Installer;
+
+use Composer\Composer;
+use Composer\IO\IOInterface;
+use Composer\Repository\InstalledRepositoryInterface;
+use Composer\Package\PackageInterface;
+use Composer\Util\Filesystem;
+use Composer\Util\Silencer;
+use Composer\Util\Platform;
+use InvalidArgumentException;
+
+class ExtensionInstaller implements InstallerInterface
+{
+    protected $composer;
+    protected $downloadManager;
+    protected $io;
+    protected $pickle = 'pickle';
+    protected $process;
+    protected $cacheDir;
+
+    public function __construct(IOInterface $io, Composer $composer)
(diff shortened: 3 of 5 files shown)


Seldaek:
If you want to look into this, please rebase on the 2.0 branch, as this will definitely not make it into 1.x.

The other considerations are that windows support is indeed a must have, and then the main question is.. does this make any sense at all considering that extensions can only really be installed/updated but not removed (as another project on the same system might still depend on them). It introduces a very special concept to Composer but if it can be done smoothly maybe it's worth it.

nikic:
I think some kind of extension installation functionality would be very valuable to have in composer, so that ordinary users (i.e. everyone minus extension maintainers) do not need to use PEAR.

However, as the library and extension installation mechanisms differ a lot, I think it might best (or at least to start with) limit this functionality to global installations. I.e. support `composer global require pecl/apcu`, but don't try to auto-install extensions from composer.json (or maybe only suggest that the possibility of doing a global installation exists, if an extension is missing).

Seldaek:
Yeah... that was kinda the point/goal of https://github.com/FriendsOfPHP/pickle tho. I'd rather have the efforts focused there tbh, because if we want to support all various linux distros, package managers, windows and various strains of OSX PHP-installers, it gets real messy real quick. And I don't really feel like maintaining all of this within the Composer project.

The ideal situation would have a pickle binary that works well, that users can use, and that Composer can suggest using in case extensions are missing, or maybe we can even execute it directly if available in $PATH.. 

I don't know what the state of the code is in Pickle, but I suspect it's at least more modern as pecl/pear, and probably worth reusing bits of it rather than starting fresh. /cc @pierrejoye @weltling @jubianchi @hywan

pierrejoye:
good morning,

Not much is needed to make it work smoothly with composer.

Back then I provided a PR but it was not optimal. If I could know clearly
where to hook it, I would be happy to finalize a possible integration (no
bundle or such).

pierrejoye:
Also looking at the config, it is missing a lot of things, I wished it would be so easy :). I would relatively strongly suggest to see what is done in pickle and what features are needed. Ideally we could work on creating a PR for composer. However I would like to have clear details about how it should be done or how composer likes to have it. :)

nikic:
@pierrejoye Can you provide some information of what the current state of pickle is? Personally I don't think composer integration is even the important part here, what we need first and foremost is a reasonably modern replacement for `pecl install`. Is pickle ready for that? If yes, why is not widely used right now? If no, what is still missing?

Seldaek:
I agree that composer integration should come later, once a given tool is proven suitable for the task.

pierrejoye:
@nikic yes it is. It supports pecl.php.net, github or technically composer's packagist if necessary. Source compilations on linux and windows (did not try mac but should work as it is gcc+autoconf), with configure options or binary install (windows, from our builds)

@Seldaek and I discussed about using packagist for extensions, I did a pickleweb concept but ideally we should use packagist and only add what is needed for extensions (basically more meta).

pierrejoye:
https://github.com/FriendsOfPHP/pickleweb for the web part of it

Hywan:
We can double down on Pickle if you want to integrate it. That will increase the adoption greatly.

torinaki:
@Seldaek, @nikic, @pierrejoye  according to your discussion I came up with a conclusion that the easiest way is to actualize the Pickle repository. 

Probably it is essential to make Pickle work before PHP 8 will be released and test on currently supported PHP versions starting from 7.1.

I think that in the context of FFI, Preloading, and JIT RFCs we are going to direction when PHP might have something similar to Python's Wheels packages where we can have PHP and C extension code in a single repo (or at least will have a dependency on both). Obvious it would not happen soon if even happen at all.

P.S. I create PR ahead of working and complete implementation to raise the discussion to avoid wasting time on changes that would not be accepted by the community.

PR #7971, opened 2019-02-11
Only use junctions if they can be safely removed

johnstevenson:
I'm probably jumping the gun a bit here since the PHP bug fixes haven't been released yet (they are merged now), but otherwise I'll have forgotten all about this in a month.


Seldaek:
Makes sense, thanks. It should still use Junctions for now on older PHPs as long as you're on windows 7+ if I understand correctly, so strictly speaking no change except upcoming PHP versions won't check the windows version anymore?

johnstevenson:
The only change now is that we won't allow junctions if they can only be removed via php (ie no proc_open).

curry684:
Awesome  👍

PR #7977, opened 2019-02-12
Explain Windows junctions in 05-repositories.md

johnstevenson:
Adds an explanation of junctions and their edge-case limitations.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -666,6 +666,10 @@ Instead of default fallback strategy you can force to use symlink with
 mirroring can be useful when deploying or generating package from a
 monolithic repository.
 
+> **Note:** On Windows, directory symlinks are implemented using NTFS junctions
+> because they can be created by non-admin users. Mirroring will always be used
+> on versions below Windows 7 or if `proc_open` has been disabled.
+
 ```json
 {
     "repositories": [


Seldaek:
You can't let go anymore I see ;) Thanks

johnstevenson:
@Seldaek  Yes, I do see that I have got rather obsessed with this! And what's more I have just noticed that this should probably have said:

> **Note:** On Windows, directory symlinks are implemented using NTFS junctions
> because they can be created by non-admin users. Mirroring will always be used
> on versions below Windows 7, or if `proc_open` has been disabled on PHP versions
> less than 7.2.16 or 7.3.3 due to a PHP bug.

Perhaps you can save me by adding this yourself!

Seldaek:
Will do ;)

Seldaek:
So I rewrote that a bit to make it more readable:

> **Note:** On Windows, directory symlinks are implemented using NTFS junctions
> because they can be created by non-admin users. Due to a bug, on PHP versions
> less than 7.2.16 or 7.3.3 mirroring will always be used on versions below 
> Windows 7, or if `proc_open` has been disabled.

But then I realized.. without proc_open, we should never try junctions I think? Because it'll fail to create them as well.

johnstevenson:
Yeah, very good point. We don't seem to check for that, so I need to change the new `safeJunctions` method to handle it. Do you agree?

Seldaek:
Yeah I think that'd do.. If you wanna change the docs too I ended up aborting that change.

PR #7980, opened 2019-02-12
Improve safe junction logic

johnstevenson:
Please check this logic. To recap:

- we need `proc_open` to call mklink to create a junction.
- we need `proc_open` available so the system rmdir is used on a package removal (rather than the fallback `Filesystem::removeDirectoryPhp`) in case a junction is not recognized due to the PHP bug.
- we need to be running on Windows 7 or up, so the system rmdir doesn't delete target content if given a junction.
- the PHP versions with the bug fix use the `Visual C++ Redistributable for Visual Studio 2017` which require a minimum of Windows 7 Sp1.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -180,24 +180,21 @@ class PathDownloader extends FileDownloader implements VcsCapableDownloaderInter
     }
 
     /**
-     * Returns true if junctions can be safely used on Windows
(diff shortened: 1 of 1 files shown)


staabm:
does this PR handle the case where `proc_open` is disabled via php-ini `disable_functions` directive?

johnstevenson:
@staabm Yes (or if php was built without it!)

PR #7988, opened 2019-02-13
Allow setting value for gitlab_domains via COMPOSER_AUTH

zaporylie:
For self-hosted gitlab there's no point in allowing overriding `gitlab-token` via COMPOSER_AUTH without also specifying `gitlab-domains`.

Please, add `gitlab-domains` to the config whitelist.

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -131,7 +131,7 @@ class Config
         // override defaults with given config
         if (!empty($config['config']) && is_array($config['config'])) {
             foreach ($config['config'] as $key => $val) {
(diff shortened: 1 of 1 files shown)


Seldaek:
In that case, we should also allow github-domains for consistency.. But the point of the env var is to have credentials as you don't want to commit that in the repo usually. Setting the gitlab-domains in the config in the repo's composer.json sounds fine tho, so I am not sure why this change is needed at all?

zaporylie:
Makes sense, hence closing the PR.

Issue #7984, opened 2019-02-13
Add Util class to read composer.json from ZIP file


aschempp:
It would be great if we could extract the private methods from `ArtifactRepository` to read the `composer.json` from ZIP files.

https://github.com/composer/composer/blob/master/src/Composer/Repository/ArtifactRepository.php#L83-L166

I currently have to duplicate these lines in two projects where I already require `composer/composer`. Would you be ok with extracting these two? I'm open to create a pull request for it.

Seldaek:
I guess you could move them to a Composer\Util\Zip, at least locateFile. The other one kinda seems very specific to ArtifactRepo, but I guess this bit https://github.com/composer/composer/blob/master/src/Composer/Repository/ArtifactRepository.php#L128-L150 could be extracted to a somewhat generic thing to read any file?

aschempp:
closed in favor of #7994

PR #7994, opened 2019-02-16
Extract the ZIP utility functions from ArtifactRepository

aschempp:
see https://github.com/composer/composer/issues/7984

I would also add some unit tests if the general approach looks good to you?


src/Composer/Util/Zip.php
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ */
+class Zip
+{
+    /**
+     * Finds the path to a file inside a ZIP archive.
+     *
+     * @param $pathToZip
```
staabm: Missing type

src/Composer/Util/Zip.php
```diff
@@ -0,0 +1,96 @@
...
+            return null;
+        }
+
+        if (0 == $zip->numFiles) {
+            $zip->close();
+
+            return null;
+        }
+
+        $foundFileIndex = static::locateFile($zip, $filename);
```
stof: private methods must be referenced with `self`, not with `static`. Otherwise, this is broken for child class (where it would try to call the method on the child class instead)

src/Composer/Util/Zip.php
```diff
@@ -0,0 +1,96 @@
...
+                $directoryName = dirname($stat['name']);
+                if ($directoryName == '.') {
+                    //if composer.json is in root directory
+                    //it has to be the one to use.
+                    return $i;
+                }
+
+                if (strpos($directoryName, '\\') !== false ||
+                    strpos($directoryName, '/') !== false) {
+                    //composer.json files below first directory are rejected
```
stof: The fact that only top-level files or files in a top-level directory can be found using this utility must be documented, as this utility can now be used in other places which might not expect that restriction.

src/Composer/Util/Zip.php
aschempp: Thanks! Fixed in 9de07bed1b3f4b96e0da90f2caab525073cf5048

src/Composer/Util/Zip.php
aschempp: Thanks! Fixed in 05d6b2178542857131011e3e46a8165414abc8aa

src/Composer/Util/Zip.php
aschempp: Very true! @Seldaek (or anyone else) can you explain why the method does a "file name length" compare? Looks like something ZIP-specific, but I couldn't find anything in the PHP documentation.

Seldaek: I think the problem is that when looping through all files in the zip file, it doesn't necessary follow a sensible order as we would think of it (top level directory listing let's say). It lists all files in the archive, so we skip those in subdirectories because we are only interested in the top level one in this case.

The length thing I believe was another way to get to the top level file as `foo/bar/composer.json` would be longer than the top level `composer.json`. The name length thing below **might** be unnecessary with the optimization to skip sub-directories above.

aschempp: I was able to find the according PR: https://github.com/composer/composer/pull/2188
So apparently the problem is that `ArtifactRepository` needs to find a `composer.json` in the root or the first subdirectory. I don't think this is magically gonna work out with other use cases.
I would therefore suggest to rename the method to `locateComposerJson` and only actually support that use case (for now)?

Seldaek: Sounds reasonable.

aschempp: Changed in 0d0cb53f314eb7422f16b479e2efd145c730362f then

aschempp:
Another thing I noticed, the function returns a `zip://…` url which requires the `zip` stream wrapper to be loaded. This was always the case, so I guess that should generally not be a problem. We could however get rid of that dependency by using `ZipArchive::getStream` and/or simply return the file content from that method. Or does anyone see a point in returning a path from that method?

Seldaek:
Sounds fine to me if you can simplify things a bit that way. But I guess if the ext is present the stream wrapper will be as well so it probably doesn't matter a lot in practice.

Seldaek:
What's the status here? I think it was ready to go but it's still marked WIP so not sure :)

aschempp:
Sorry, you're right. I just checked AppVeyor but the failing test looks unrelated to this PR.

Seldaek:
Thanks

PR #7996, opened 2019-02-18
Fix type issues

muglug:
After seeing the work done in https://github.com/composer/composer/pull/7883, I ran Psalm on the repo.

The only obvious bug it found was mentioned in https://github.com/composer/composer/commit/902cb290e792fdc3a4ff6a7af23ce7fea54b7acc#r32358863, but running it highlighted some redundant code and invalid docblocks.

I can create a separate PR to add Psalm to the build, if you'd like - it would look like this: https://github.com/muglug/composer/pull/2/commits/d14d23949152ea940457e48b7bb17b59a5c9047b.

### DIFF
--- a/src/Composer/Command/BaseCommand.php
+++ b/src/Composer/Command/BaseCommand.php
@@ -27,6 +27,8 @@ use Symfony\Component\Console\Command\Command;
 /**
  * Base class for Composer commands
  *
+ * @method Application getApplication()
+ *
  * @author Ryan Weaver <ryan@knplabs.com>
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
@@ -46,7 +48,7 @@ abstract class BaseCommand extends Command
      * @param  bool              $required
      * @param  bool|null         $disablePlugins
      * @throws \RuntimeException
-     * @return Composer
+     * @return ?Composer
      */
     public function getComposer($required = true, $disablePlugins = null)
     {
@@ -173,7 +175,7 @@ abstract class BaseCommand extends Command
 
         if ($input->getOption('prefer-source') || $input->getOption('prefer-dist') || ($keepVcsRequiresPreferSource && $input->hasOption('keep-vcs') && $input->getOption('keep-vcs'))) {
             $preferSource = $input->getOption('prefer-source') || ($keepVcsRequiresPreferSource && $input->hasOption('keep-vcs') && $input->getOption('keep-vcs'));
-            $preferDist = $input->getOption('prefer-dist');
+            $preferDist = (bool) $input->getOption('prefer-dist');
         }
 
         return array($preferSource, $preferDist);
--- a/src/Composer/Command/BaseCommand.php
+++ b/src/Composer/Command/BaseCommand.php
@@ -48,7 +48,7 @@ abstract class BaseCommand extends Command
      * @param  bool              $required
      * @param  bool|null         $disablePlugins
      * @throws \RuntimeException
-     * @return ?Composer
+     * @return Composer|null
      */
     public function getComposer($required = true, $disablePlugins = null)
     {
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -226,7 +226,7 @@ EOT
         }
 
         $settingKey = $input->getArgument('setting-key');
-        if (!$settingKey) {
+        if (!$settingKey || !is_string($settingKey)) {
             return 0;
         }
 
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -602,20 +602,6 @@ EOT
                         $text .= "Install either of them or recompile php without --disable-iconv";
                         break;
 
-                    case 'unicode':
-                        $text = PHP_EOL."The detect_unicode setting must be disabled.".PHP_EOL;
-                        $text .= "Add the following to the end of your `php.ini`:".PHP_EOL;
-                        $text .= "    detect_unicode = Off";
-                        $displayIniMessage = true;
-                        break;
-
-                    case 'suhosin':
-                        $text = PHP_EOL."The suhosin.executor.include.whitelist setting is incorrect.".PHP_EOL;
-                        $text .= "Add the following to the end of your `php.ini` or suhosin.ini (Example path [for Debian]: /etc/php5/cli/conf.d/suhosin.ini):".PHP_EOL;
-                        $text .= "    suhosin.executor.include.whitelist = phar ".$current;
-                        $displayIniMessage = true;
-                        break;
-
                     case 'php':
                         $text = PHP_EOL."Your PHP ({$current}) is too old, you must upgrade to PHP 5.3.2 or higher.";
                         break;
@@ -713,7 +699,7 @@ EOT
     /**
      * Check if allow_url_fopen is ON
      *
-     * @return bool|string
+     * @return true|string
      */
     private function checkConnectivity()
     {
--- a/src/Composer/DependencyResolver/GenericRule.php
+++ b/src/Composer/DependencyResolver/GenericRule.php
@@ -23,10 +23,10 @@ class GenericRule extends Rule
     protected $literals;
 
     /**
-     * @param array                 $literals
-     * @param int                   $reason     A RULE_* constant describing the reason for generating this rule
-     * @param Link|PackageInterface $reasonData
-     * @param array                 $job        The job this rule was created from
+     * @param array                      $literals
+     * @param int|null                   $reason     A RULE_* constant describing the reason for generating this rule
+     * @param Link|PackageInterface|null $reasonData
+     * @param array                      $job        The job this rule was created from
      */
     public function __construct(array $literals, $reason, $reasonData, $job = null)
     {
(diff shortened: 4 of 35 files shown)


Seldaek:
I am not sure what to do here.. Skimming through changes I spotted a few which are incorrect, which makes me think I'd have to go through all of them in depth before merging. There are definitely some correct ones too though so closing outright isn't good either.

Why are some nullables switched to `?foo` while others are `foo|null`? I believe the latter is preferred as some tooling doesn't support `?foo` in phpdoc.

Adding Psalm to the build I'd rather not as it'll probably be quite redundant with PHPStan if we increase the check level there. It's also one more error-exclusion list to maintain.

src/Composer/Command/DiagnoseCommand.php
```diff
@@ -602,20 +602,6 @@ private function checkPlatform()
                         $text .= "Install either of them or recompile php without --disable-iconv";
                         break;
 
-                    case 'unicode':
```
staabm: this changes seem unrelated?

muglug: Oh it’s not - if you look at the code above this condition can never be met.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -82,7 +82,7 @@ public function __construct(IOInterface $io, Config $config, array $options = ar
      * @param bool   $progress  Display the progression
      * @param array  $options   Additional context options
      *
-     * @return bool true
+     * @return bool|string true
```
Seldaek: This change is invalid. get() will return bool when $fileName is passed in.

src/Composer/Util/Loop.php
```diff
@@ -29,6 +29,7 @@ public function __construct(HttpDownloader $httpDownloader)
 
     public function wait(array $promises)
     {
+        /** @var ?\Exception */
```
Seldaek: Still some `?` nullable here :)

src/Composer/Util/Bitbucket.php
```diff
@@ -202,6 +202,7 @@ public function requestToken($originUrl, $consumerKey, $consumerSecret)
 
         $this->storeInAuthConfig($originUrl, $consumerKey, $consumerSecret);
 
+        /** @var array $this->token */
```
Seldaek: This seems kinda useless to me. The tool might need it, but that seems like a bug in the tool then. The property has a type declaration already. Please remove.

src/Composer/IO/IOInterface.php
```diff
@@ -160,7 +160,7 @@ public function askAndHideAnswer($question);
      * @param bool        $multiselect  Select more than one value separated by comma
      *
      * @throws \InvalidArgumentException
-     * @return int|string|array          The selected value or values (the key of the choices array)
+     * @return int|string|array|bool    The selected value or values (the key of the choices array)
```
Seldaek: This probably sohuld be `mixed` really at this point ;)

src/Composer/Util/Bitbucket.php
muglug: yeah, I agree. It was added to reset the tool's knowledge of `$this->token`, which already had a !empty check above. I'll think about how this could be dealt with better.

src/Composer/Downloader/DownloadManager.php
```diff
@@ -294,9 +294,7 @@ public function update(PackageInterface $initial, PackageInterface $target, $tar
 
         // if downloader type changed, or update failed and user asks for reinstall,
         // we wipe the dir and do a new install instead of updating it
-        if ($initialDownloader) {
-            $initialDownloader->remove($initial, $targetDir);
-        }
+        $initialDownloader->remove($initial, $targetDir);
```
Seldaek: This introduces a bug. $initialDownloader might be null as getDownloaderForPackage can return null. I am not sure why this change was done.

muglug: It does not, as far as I can see.

Have a look at the code above:

```
        if (!$initialDownloader && !$downloader) {
            return;
        }
        // if we have a downloader present before, but not after, the package became a metapackage and its files should be removed
        if (!$downloader) {
            $initialDownloader->remove($initial, $targetDir);
            return;
        }
```

After that, it's impossible for `$initialDownloader` to be null. In fact it's assumed to be non-null when used in `$this->getDownloaderType($initialDownloader)`.

This is one of the areas where Psalm is better than PHPStan - it includes some (very basic) SAT solving to help it understand the above logic.

src/Composer/DependencyResolver/GenericRule.php
```diff
@@ -23,10 +23,10 @@ class GenericRule extends Rule
     protected $literals;
 
     /**
-     * @param array                 $literals
-     * @param int                   $reason     A RULE_* constant describing the reason for generating this rule
-     * @param Link|PackageInterface $reasonData
-     * @param array                 $job        The job this rule was created from
+     * @param array                      $literals
+     * @param int|null                   $reason     A RULE_* constant describing the reason for generating this rule
+     * @param Link|PackageInterface|null $reasonData
```
Seldaek: Also not sure why those two were marked nullable. Rather undo unless there's a good explanation.

src/Composer/DependencyResolver/Solver.php
```diff
@@ -513,7 +514,7 @@ protected function analyze($level, Rule $rule)
             );
         }
 
-        $newRule = new GenericRule($learnedLiterals, Rule::RULE_LEARNED, $why);
+        $newRule = new GenericRule($learnedLiterals, Rule::RULE_LEARNED, null);
```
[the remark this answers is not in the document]
muglug: The third arg of `GenericRule::__construct` expects `Composer\Package\PackageInterface|Composer\Package\Link`, but `$why` is an `int`.

src/Composer/DependencyResolver/GenericRule.php
muglug: You pass `null` in as the third argument in a number of places (happy to provide examples) so I thought it should also be in the signature.

src/Composer/Command/ConfigCommand.php
```diff
@@ -226,7 +226,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         $settingKey = $input->getArgument('setting-key');
-        if (!$settingKey) {
+        if (!$settingKey || !is_string($settingKey)) {
```
Seldaek: This adds no value AFAICT.

src/Composer/Command/BaseCommand.php
```diff
@@ -27,6 +27,8 @@
 /**
  * Base class for Composer commands
  *
+ * @method Application getApplication()
```
Seldaek: This shouldn't be needed, it's defined on the parent class.

src/Composer/DependencyResolver/GenericRule.php
Seldaek: Ok in that case fair enough leave it.

src/Composer/DependencyResolver/Solver.php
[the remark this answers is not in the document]
Seldaek: Right, but in Rule.php:240 it appears to use it as an int, so IMO the doctype should be expanded rather than removing data here.

Generally speaking when following static analysis output I'd tend to suspect the docs of being out of date rather than the code. Fixing the code can lead to bugs, fixing docs can only lead in confusion at worst.

src/Composer/Command/ConfigCommand.php
muglug: Couldn't the output of `$input->getArgument()` be something other than a string? It's passed into `preg_match`, which expects one.

src/Composer/DependencyResolver/Solver.php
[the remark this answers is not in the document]
muglug: Oh sorry! Have fixed.

src/Composer/Command/BaseCommand.php
muglug: This tells Psalm (and PHPStan, and PHPStorm and other type checkers) that the value returned by `getApplication` is a subclass of the return type declared in the parent.

src/Composer/Command/ConfigCommand.php
Seldaek: As far I know in this configuration it only returns string or null, which would be falsey anyway. But obviously getArgument can be sometimes an array if the arg is configured as such, so the docs on that have to allow for more return types than are applicable in a given case. You can leave as is I guess it's harmless either way.

src/Composer/Command/BaseCommand.php
Seldaek: Ahh I see, makes sense thanks.

Seldaek:
Alright, thanks for the updates. I think we might look at Psalm again in the future, maybe with v3 once we can drop old PHP versions and add much more type info inline it'll make it easier to adopt multiple static analyzers.

muglug:
You're welcome! Let me know if you have any questions - I understand the pain of having two competing configs. Some great projects use both, [some great projects use none](https://github.com/sebastianbergmann/phpunit).

YMMV, but there's not much to differentiate Psalm and PHPStan when the type coverage is relatively low (81% for the `src` directory). There only real difference that'll matter to you if/when you attempt to "level up" is configuration - Psalm allows you to pick and choose what issues you care about (and in which files) because we need that level of granularity at Vimeo, whereas PHPStan is a more one-size-fits-all sort of deal.

Seldaek:
Alright thanks for the overview :)

PR #7999, opened 2019-02-21
Store dev mode in installed.json, fixes #3008

### LINKED ISSUE #3008: Save whether --dev or --no-dev is installed
Hi,

I would like to request to save whether the install/update command is run with --dev or --no-dev option.

Background: I would like to add tooling support for this on the composer-eclipse-plugin. To add the respective (autoload or autoload-dev) into the build path, so contents are available for content assist.

Thank you very much

Seldaek:
This introduces a BC break in installed.json as the packages are now nested in a "packages" key, but it also brings it in line with the composer.lock format which isn't bad, and opens up for further improvements later. Any thoughts @gossi as you requested the feature?

Fixed #3008

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ script:
   - ls -d tests/Composer/Test/* | grep -v TestCase.php | parallel --gnu --keep-order 'echo "Running {} tests"; ./vendor/bin/phpunit -c tests/complete.phpunit.xml --colors=always {} || (echo -e "\e[41mFAILED\e[0m {}" && exit 1);'
   # Run PHPStan
   - if [[ $PHPSTAN == "1" ]]; then
-      composer require --dev phpstan/phpstan-shim:^0.11 --ignore-platform-reqs &&
+      bin/composer require --dev phpstan/phpstan-shim:^0.11 --ignore-platform-reqs &&
       vendor/bin/phpstan.phar analyse src tests --configuration=phpstan/config.neon --autoload-file=phpstan/autoload.php;
     fi
 
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -616,7 +616,7 @@ class Installer
             }
 
             if ($this->executeOperations || $this->writeLock) {
-                $localRepo->write();
+                $localRepo->write($this->devMode);
             }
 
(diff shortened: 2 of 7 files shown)


gossi:
Uuh, wow. Thanks for getting this in. I think I've requested this a couple of years ago, very glad about it's implementation. However, I'm mostly doing frontend the last three years I might not be a good candidate for giving proper feedback. However, I remember it was a good thing for tools/IDEs build around composer.

Seldaek:
Hehe ok ;) I guess I'll go ahead with the change then, hopefully it's still useful to someone out there.

Seldaek:
Ping @psliwa as this might impact or benefit the PHPStorm integration. Any comment? Also would you benefit from #2174 or is it rather all pointless and I can close both issues?

gossi:
ping @zulus if this will have an impact on eclipse PDT.

zulus:
This will broke PDT validation (build) path and namespace calculation for now, but we can easy prepare patch for upcoming version (Eclipse 2019.03).

Seldaek:
It'd be great if you can prepare support for this.. Obviously the old format should still be supported as well, but essentially if you have an object with a `packages` key instead of an array as top level, then take list of packages from there instead of top level array.

bobemoe:
Been wanting this feature for a while now so thanks for adding it :) Funny it was added some years ago and I'm still seeing the old behaviour.

How do I go about enabling this? Or store the dev/no-dev option? I feel I'm missing something obvious :/

Seldaek:
Sorry I am not sure what you are asking here. If you are using Composer 2+ you should have the dev field in the installed.json, e.g.

```
$ cat vendor/composer/installed.json  | grep '"dev"'
    "dev": true,
```

bobemoe:
Even so your response was still somewhat helpful. I'd not realised the installed.json was in vendor/composer so thanks for the full path.

Now I can see that file I can confirm composer is saving the last used mode to the "dev" variable, so this seems to be functioning correctly.

I was expecting the next run of `composer install` to then use the mode that is set in installed.json but I now think this is not the expected behaviour.

Seldaek:
No indeed, it is only persisted there so tooling knows if dev dependencies are installed or not.

PR #8001, opened 2019-02-25
WIP: gitlab-protocol to get https URLs from GitLab

bradjones1:
Refs https://github.com/composer/satis/issues/419 et. al.

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -247,7 +247,7 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
                 if ($io->isDebug()) {
                     $io->writeError('    Failed downloading '.$package->getName().': ['.get_class($e).'] '.$e->getCode().': '.$e->getMessage());
                     $io->writeError('    Trying the next URL for '.$package->getName());
-                } elseif (count($urls)) {
+                } else {
                     $io->writeError('    Failed downloading '.$package->getName().', trying the next URL ('.$e->getCode().': '.$e->getMessage().')');
                 }
 


src/Composer/Util/Git.php
```diff
@@ -153,28 +153,36 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
...
-
-                    if (!$gitLabUtil->authorizeOAuth($match[2]) && $this->io->isInteractive()) {
-                        $gitLabUtil->authorizeOAuthInteractively($match[1], $match[2], $message);
-                    }
+            } elseif (preg_match('{^(?:https?|git)://' . self::getGitLabDomainsRegex($this->config) . '/(.*)}', $url, $match)) {
+                $gitLabProtocols = $this->config->get('gitlab-protocols');
+                if (!is_array($gitLabProtocols)) {
+                    throw new \RuntimeException('Config value "gitlab-protocols" must be an array, got ' . gettype($protocols));
                 }
+                if (in_array('https', $gitLabProtocols)) {
```
Seldaek: Not entirely sure what this is doing here.. it checks if https is in the protocols ok, but then below it uses `$match[1]` which from the regex above could be `git`. So why are we checking for https if the protocol is git, and why is gitlab-protocols defaulting to ssh and not https + ssh?

bradjones1: @Seldaek  Thanks for the review. This is probably a poor PR on my part, however the goal is to be able to have the gitlab driver to return https source URLs, per the convo over at https://github.com/composer/satis/issues/419#issuecomment-466875285 . This might not be the right avenue, though I'm curious what would be for that problem space.

PR #8013, opened 2019-03-01
Return non zero exit code on deprecation

### LINKED ISSUE #8004: Validate strict should have a none zero exit code when deprecations are thrown
My `composer.json`:

```json
{
    "require": {
          "beberlei/DoctrineExtensions": "^1.0"
     }
}
```

```
Deprecation warning: require.beberlei/DoctrineExtensions is invalid, it should not contain uppercase characters. Please use beberlei/doctrineextensions instead. Make sure you fix this as Composer 2.0 will error.
Checking composer.json: OK
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys:
OK
Checking composer version: OK
Composer version: 1.8.4
PHP version: 7.1.25
PHP binary path: /usr/local/Cellar/php@7.1/7.1.25/bin/php
```

When I run this command:

```bash
composer validate --strict
```

I get the following output:

```bash
echo $? # returns 0 
```

In strict validation I would think that the exit code of --strict should not be 0 when a deprecation is thrown.

PReimers:
As reported in #8004 the validate command should return a non zero exit code when using `--strict` mode.

Fixes #8004
#OpenSourceFriday

### DIFF
--- a/src/Composer/Package/Loader/ValidatingArrayLoader.php
+++ b/src/Composer/Package/Loader/ValidatingArrayLoader.php
@@ -49,6 +49,10 @@ class ValidatingArrayLoader implements LoaderInterface
         $this->warnings = array();
         $this->config = $config;
 
+        if ($err = self::hasPackageNamingError($config['name'])) {
+            $this->warnings[] = 'Deprecation warning: Your package name '.$err.' Make sure you fix this as Composer 2.0 will error.';
+        }
+
         if ($this->strictName) {
             $this->validateRegex('name', '[A-Za-z0-9][A-Za-z0-9_.-]*/[A-Za-z0-9][A-Za-z0-9_.-]*', true);
         } else {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, this is a good start but one more detail, we should also check `$this->config['name']` against hasPackageNamingError (See top of the function where some checks are already done for the name). Probably the existing logic shouldn't be touched but merely added to as a second layer as we don't want to mess with the strict naming rules until 2.0.

PReimers:
I wasn't sure about the name.
Will find a solution and update the PR ;)

Seldaek:
Yes we also check it in RootPackageLoader IIRC, to warn package authors as well as users, which is kinda the most important part to fix the problem at the source :)

PReimers:
@Seldaek 
Currently the name is validated agains: `[A-Za-z0-9][A-Za-z0-9_.-]*/[A-Za-z0-9][A-Za-z0-9_.-]*`

Should it be validated against `[a-z0-9][a-z0-9_.-]*/[a-z0-9][a-z0-9_.-]*` or rather agains `self::hasPackageNamingError($this->config['name'])`?

Seldaek:
@PReimers the current validation should remain untouched to avoid breaking changes, and it should validate against self::hasPackageNamingError($this->config['name']) additionally, triggering a warning if it fails.

PReimers:
@Seldaek 
I've added the check for `self::hasPackageNamingError($this->config['name'])` and some unit tests to cover them ;)

Seldaek:
Thanks!

Issue #7981, opened 2019-02-12
composer cli config command does not support arrays


alphp:
Composer version: 1.8.4
PHP version: 7.3.2

My `composer.json`:

```json
{
    "name": "me/project",
    "type": "project",
    "config": {
        "vendor-dir": "vendors"
    }
}
```

When I run this command:

```
composer config extra.installer-types ["cakephp-plugin"]
composer config extra.installer-paths.app/Plugin/{$name} ["type: cakephp-plugin"]
```

Output `composer.json`:

```json
{
    "name": "me/project",
    "type": "project",
    "config": {
        "vendor-dir": "vendors"
    },
    "extra": {
        "installer-types": "[cakephp-plugin]",
        "installer-paths": {
            "app/Plugin/{$name}": "[type:cakephp-plugin]"
        }
    }
}
```

Expected `composer.json`:

```json
{
    "name": "me/project",
    "type": "project",
    "config": {
        "vendor-dir": "vendors"
    },
    "extra": {
        "installer-types": ["cakephp-plugin"],
        "installer-paths": {
            "app/Plugin/{$name}": ["type:cakephp-plugin"]
        }
    }
}
```

Seldaek:
`omposer config extra.installer-types.0 "cakephp-plugin"` works for the first one, but the second is too deep and is not supported. Complex operations have to be done manually in the json.. the config command is a convenience thing but meant to be a full blown json editor.

XedinUnknown:
I too would like very much to see this work, and incidentally for the same config keys. The reason is that `composer config` is useful in configuring different kinds of environment, e.g. staging VS dev. In this case, I am developing a WP plugin, and I need to install `wordpress-plugin` packages _alongside mine_ (i.e. `../{$name}/`) in a dev environment, but into the `vendor` folder in a staging environment.

Is there a particular problem with supporting this? Perhaps, it could be marked as a "nice to have" feature, and re-considered in the future?

Seldaek:
Most tools editing json use some sort of XPath-style json query language to address deep keys as the structure can get very complex. I don't really want to add support for this.. and also it'd be hard to add it in a BC way at this point anyway. So we support up to two level deep as a convenience and if that's not enough you're on your own. There probably are other CLI tools you could use to modify the json in a programmatic way if you really need it in a CI setting or similar.

XedinUnknown:
Good point re. other JSON-editing CLI tools, thanks!

alphp:
My method:

```bash
[...]

PR #8017, opened 2019-03-03
Arbitrary deep on config extra properties (CLI)

alphp:
https://github.com/composer/composer/issues/7981

### DIFF
--- a/src/Composer/Json/JsonManipulator.php
+++ b/src/Composer/Json/JsonManipulator.php
@@ -176,7 +176,7 @@ class JsonManipulator
 
             foreach ($nodes as $node) {
                 if (!array_key_exists($node, $pointer)) {
-                    $pointer[$node] = [];
+                    $pointer[$node] = array();
                 }
 
                 $pointer = &$pointer[$node];
--- a/src/Composer/Json/JsonManipulator.php
+++ b/src/Composer/Json/JsonManipulator.php
@@ -168,7 +168,25 @@ class JsonManipulator
     public function addProperty($name, $value)
     {
         if (substr($name, 0, 6) === 'extra.') {
-            return $this->addSubNode('extra', substr($name, 6), $value);
+            $decoded = JsonFile::parseJson($this->contents);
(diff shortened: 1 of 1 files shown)


alphp:
About checks:
AppVeyor: Composer\Test\Util\FilesystemTest::testJunctions => This PR is not the cause of the error
Travis: Fails on PHP-5.3. Why are they still testing EOL versions? http://php.net/supported-versions.php

Seldaek:
This will decode/encode the entire file though, which might lose formatting and create more diffs than needed. I am not sure if we rather have that than not have the functionality at all.

XedinUnknown:
☝️ Precisely the concern I have in mind when editing JSON.

alphp:
I think the best way to maintain the composer.json format is to just rewrite it.

I have seen several times in which the program parse the configuration files and format them according to their style once validated.

For me it is a simple way to maintain a consistent format.

Whenever I have used json in my projects I leave PHP the parsing task json => array / array => json. If I need the json to be friendly with people, I use JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES and that's it.

XedinUnknown:
IMO `composer.json`, while machine readable, is very often read and modified by humans, so it's important to keep unmodified parts. This is especially important since `composer.json` is most of the times committed to VCS. For example, if I want lists to be one item per line, re-encoding JSON may introduce maintainability problems. Although, this is probably not a concern if the editing happens in a CI/CD environment.

alphp:
I would have liked to be able to use recursion with the addSubNode function, but I did not see it possible.

I did not want to make major modifications to addSubNode, and what I chose is to change the operation mode of addProperty when adding an "extra" property.

In this sense I would have liked to be able to limit the modifications of the json to the minimum possible, but I did not see the way to do it.

Seldaek:
This is solved in composer2 using the new --json arg to config https://getcomposer.org/doc/03-cli.md#modifying-extra-values

Issue #8005, opened 2019-02-26
open the composer website via composer cli


staabm:
it would be great if I could invoke the composer website/documentation from the cli tool.

e.g. could `composer docs depends` or similar, open the webbrowser on https://getcomposer.org/doc/03-cli.md#depends-why-

another idea would be a command which just opens https://getcomposer.org/doc/03-cli.md and the user can then manually search inside this website (so the cli tool does not depend too much on the website).

I think this would be a usefull feature, because the website contains way more information then the cli tool itself (which only describes the parameters and more or less expects the user knows what the actual commands are used for)

I will leave this here to gather some input/feedback. In case we can agree on how this will/should look like I can finally provide a PR.

staabm:
another idea could be printing a clickable link inside the commands help-text.

![grafik](https://user-images.githubusercontent.com/120441/53405828-fcd39080-39b8-11e9-94f4-cb728b7a2003.png)

symfony just added support for that: https://symfony.com/blog/new-in-symfony-4-3-console-hyperlinks

stof:
The composer.phar uses Symfony 2.8 components, due to the need to support PHP 5.3. So a Symfony 4.3 new feature won't be helpful here.

but another idea is to improve the command help to include more info there directly.

alcohol:
Though opening can be tricky, e.g. on Mac you would invoke `open` but on Linux generally it is `xdg-open` and I haven't got a clue what to use when it comes to Windows.

staabm:
Yeah I see your issues. Maybe just printing the url and let the user the copy&paste stuff into his browser would be enough as long as the console needs to stay on such an old version.

alcohol:
I think 2.x still aims to support the same PHP version as 1.x, and it is a transitional major release?

PReimers:
Since PHP < 7.1 isn't maintained anymore I would suggest bumping the min. PHP Version to 7.1 for the 2.x release.
[PHP Supported Versions](https://secure.php.net/supported-versions.php)

Then it would be possible to use the new symfony console feature.

r408512:
Since https://getcomposer.org/doc/03-cli.md#depends-why- is md file and it's in the source https://raw.githubusercontent.com/composer/composer/master/doc/03-cli.md

what about using file_get_contents() and parsing? 

proof of concept:

``` php test.php depends ```

test.php
```
if (php_sapi_name() != "cli") 
{ 
	echo "This file can only be run from the command line.".PHP_EOL; 
	exit;
}

[...]

PR #8018, opened 2019-03-04
Add documentation command to get documentation for a command 

r408512:
#8005

### DIFF
--- a/src/Composer/Command/DocumentationCommand.php
+++ b/src/Composer/Command/DocumentationCommand.php
@@ -45,17 +45,14 @@ EOT
         $docs = [];
         $aliases = [];
 
-        $lines = explode("\n", file_get_contents($this->documentationUri) );
+        $lines = explode("\n", file_get_contents($this->documentationUri));
 
-        for ($i=0; $i < sizeof($lines); $i++) 
-        { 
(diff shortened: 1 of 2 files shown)


src/Composer/Command/DocumentationCommand.php
```diff
@@ -0,0 +1,95 @@
...
+        $this
+            ->setName('documentation')
+            ->setAliases(array('docs'))
+            ->setDescription('Get documentation for composer command.')
+            ->setDefinition(array(
+                new InputArgument('name', InputArgument::REQUIRED, 'Command name to get documentation for.'),
+            ))
+            ->setHelp(
+                <<<EOT
+<info>php composer.phar about</info>
```
staabm:
```suggestion
<info>php composer.phar documentation</info>
```

Seldaek:
Thanks for the PR. I think I'd rather have a link to 03-cli#commandname in every command's `setHelp()` - to allow people to read more there..

Something like:

`Read more at https://...` at the end of the text?

Otherwise the info is spread out between --help and this new docs command, that sounds suboptimal.

staabm:
should be superceeded by https://github.com/composer/composer/pull/8019

PR #8027, opened 2019-03-06
Raise phpstan level to 1

adaamz:
Hi,
how are you satisfied with phpstan so far? :-)

I prepared some fixes to raise phpstan to level 1. This is really small step forward, but smaller steps are better for stability.

I'm not sure how you handle error state in Composer, so I can change it after review.

ps. In current phpstan stable version there is small bug which is ignored and commented.
pps. I used ignoreErrors with path filter, that means from now you have to run phpstan from root of repository (because %currentWorkingDirectory% should point there).


phpstan/config.neon
```diff
@@ -12,18 +12,23 @@ parameters:
...
         # imagick is not installed
         - '~^Instantiated class Imagick not found\.$~'
+        # xsl could not be installed
+        #-
+        #    message: '~^Constant LIBXSLT_DOTTED_VERSION not found\.$~'
+        #    path: '%currentWorkingDirectory%/src/Composer/Repository/PlatformRepository.php'
+        # windows specific constants
+        -
+            message: '~^Constant PHP_WINDOWS_VERSION_MAJOR not found\.$~'
+            path: '%currentWorkingDirectory%/src/Composer/Downloader/PathDownloader.php'
```
staabm: Wouldnt this also work without the restriction on having it to run from the root folder?

```suggestion
            path: '*/src/Composer/Downloader/PathDownloader.php'
```

phpstan/config.neon
```diff
@@ -12,18 +12,23 @@ parameters:
...
-
         # ion cube is not installed
         - '~^Function ioncube_loader_\w+ not found\.$~'
         # rar is not installed
         - '~^Call to static method open\(\) on an unknown class RarArchive\.$~'
         # imagick is not installed
         - '~^Instantiated class Imagick not found\.$~'
+        # xsl could not be installed
+        #-
+        #    message: '~^Constant LIBXSLT_DOTTED_VERSION not found\.$~'
```
staabm: what about this commented case?

CZechBoY: Yes, I added it because I do not have xsl enabled, but in travis it is enabled.
I will probably remove it completly as it only confuses other developers.

CZechBoY: From what directory would you like to run analysis and why?

phpstan/config.neon
CZechBoY: From what directory would you like to run analysis and why?

staabm: usually I do analysis only from the project root.

in my projects I use the `*` instead of `%currentWorkingDirectory%` because its easier to read and less tool specific magic

CZechBoY: I think it is more readable when you run analysis from other working directory than magic wildcard star.

stof: I don't understand what you mean here. Your version is not more readable when running from other working directory, as *it does not work when doing it* (you say we *have to run it from the root*).

src/Composer/Command/ShowCommand.php
```diff
@@ -541,6 +541,7 @@ protected function getPackage(RepositoryInterface $installedRepo, RepositoryInte
         $repositorySet->addRepository($repos);
 
         $matchedPackage = null;
+        $package = null;
```
stof: this won't help much. If the loop is empty, you will have `null` as `$package`, which won't work either (and will be reported again by phpstan at a higher level)

src/Composer/DependencyResolver/DefaultPolicy.php
```diff
@@ -68,6 +68,10 @@ public function selectPreferredPackages(Pool $pool, array $installedMap, array $
             });
         }
 
+        if (!isset($policy)) {
```
stof: Move `$policy = $this;` outside the `foreach` loop instead of adding this exception which should never happen

src/Composer/DependencyResolver/Solver.php
```diff
@@ -479,6 +479,10 @@ protected function analyze($level, Rule $rule)
                     }
                 }
 
+                if (!isset($literal)) {
```
stof: this looks like a bug in phpstan. A `while (true)` will *always* enter the loop.

src/Composer/Package/Loader/JsonLoader.php
```diff
@@ -38,6 +38,12 @@ public function load($json)
             $config = JsonFile::parseJson(file_get_contents($json), $json);
         } elseif (is_string($json)) {
             $config = JsonFile::parseJson($json);
+        } else {
+            throw new \InvalidArgumentException(sprintf(
+                "JsonLoader: Unknown \$json parameter %s (\"%s\"). Please report at https://github.com/composer/composer/issues/new.",
+                gettype($json),
+                $json
```
stof: passing `$json` here would break if it is not a scalar (and we already know it is not a string)

src/Composer/Package/Package.php
```diff
@@ -599,6 +599,12 @@ protected function getUrls($url, $mirrors, $ref, $type, $urlType)
                     $mirrorUrl = ComposerMirror::processGitUrl($mirror['url'], $this->name, $url, $type);
                 } elseif ($urlType === 'source' && $type === 'hg') {
                     $mirrorUrl = ComposerMirror::processHgUrl($mirror['url'], $this->name, $url, $type);
+                } else {
+                    throw new \InvalidArgumentException(sprintf(
+                        "Package: Unknown \$urlType %s (\"%s\"). Please report at https://github.com/composer/composer/issues/new.",
```
stof: reporting it to composer is useless. It should be reported to the author of the package (who wrote an invalid file)

src/Composer/Util/ConfigValidator.php
```diff
@@ -184,7 +184,7 @@ public function validate($file, $arrayLoaderValidationFlags = ValidatingArrayLoa
             $errors = array_merge($errors, $e->getErrors());
         }
 
-        $warnings = array_merge($warnings, $loader->getWarnings());
+        $warnings = array_merge($warnings, isset($loader) ? $loader->getWarnings() : array());
```
stof: move the instantiation outside the `try/catch` instead to fix this (a InvalidPackageException will never be throw in the constructor anyway, always in `load`)

src/Composer/Command/ShowCommand.php
CZechBoY: @stof Do you have better solution?

src/Composer/DependencyResolver/Solver.php
CZechBoY: Yes, there is `while(true)`, but also there is `if` which breaks setting `$literal`.
Also I'm confused which literal you want to use - there is `$literal` in `foreach` as item and next is in `$literal = $decision[Decisions::DECISION_LITERAL]`.

CZechBoY:
@stof Thanks for review and hint to fix issues!

src/Composer/Command/ShowCommand.php
Seldaek: Could we return early before the foreach if $matches is empty? That'd probably make more sense. `return array(null, array());` I think that'd work.

src/Composer/DependencyResolver/Solver.php
Seldaek: I'd say it should be $literal from the while loop above, not the one from further up. IMO we should  `unset($literal)` before the `while (true)` and then the isset/throw isn't needed I'd say as the only way to go past the while loop is either $literal is set or the SolverBugException is thrown and we exit the whole function anyway. Happy to have @naderman double check this though.

src/Composer/Package/Package.php
```diff
@@ -599,6 +599,13 @@ protected function getUrls($url, $mirrors, $ref, $type, $urlType)
                     $mirrorUrl = ComposerMirror::processGitUrl($mirror['url'], $this->name, $url, $type);
                 } elseif ($urlType === 'source' && $type === 'hg') {
                     $mirrorUrl = ComposerMirror::processHgUrl($mirror['url'], $this->name, $url, $type);
+                } else {
+                    throw new \InvalidArgumentException(sprintf(
+                        "Package: Unknown \$urlType \"%s\" and \$type \"%s\". Please report to maintainer of package \"%s\".",
+                        $urlType,
+                        $type,
+                        $this->name
+                    ));
```
Seldaek: This should rather be a `continue;` statement, this shouldn't be a failure case AFAICT.

src/Composer/Util/Git.php
```diff
@@ -79,7 +79,7 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
                     return;
                 }
                 $messages[] = '- ' . $protoUrl . "\n" . preg_replace('#^#m', '  ', $this->process->getErrorOutput());
-                if ($initialClone) {
+                if ($initialClone && isset($origCwd)) {
```
Seldaek: This is also a phpstan error.. there is no way $origCwd is unset if $initialClone is true. That's the kind of weird stuff that makes me reluctant to increase the check level tbh.. when you have to do gymnastics in the code to please our computer overlords it's not good.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -287,6 +287,7 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
             }
             $errorMessage .= preg_replace('{^file_get_contents\(.*?\): }', '', $msg);
         });
+        $http_response_header = array();
```
Seldaek: This is also an error IMO as we always check if it's been set below. Shouldn't need to be pre-set.

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -553,6 +554,8 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
      */
     protected function getRemoteContents($originUrl, $fileUrl, $context, array &$responseHeaders = null)
     {
+        $result = false;
```
Seldaek: Not such a big deal this one but technically also an error, as $result is either set or we will throw before using it.

tests/Composer/Test/InstallerTest.php
```diff
@@ -257,7 +257,7 @@ public function testIntegration($file, $message, $condition, $composerConfig, $l
 
         $output = str_replace("\r", '', $io->getOutput());
         $this->assertEquals($expectResult, $result, $output . stream_get_contents($appOutput));
-        if ($expectLock) {
+        if ($expectLock && isset($actualLock)) {
```
Seldaek: Same style of error again, if $expectLock is true $actualLock is always set above.

src/Composer/Installer.php
```diff
@@ -972,6 +972,8 @@ private function processDevPackages($localRepo, Pool $pool, $policy, $repositori
 
         if ($this->update && $this->updateWhitelist) {
             $currentPackages = $this->getCurrentPackages($installedRepo);
+        } else {
+            $currentPackages = array();
```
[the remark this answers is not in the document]
stof: but phpstan may not be sure that these properties have not been mutated in the meantime.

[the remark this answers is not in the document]
Seldaek: Ok for this one that's a fair point, but there are similar ones below that use local variables.

tests/Composer/Test/InstallerTest.php
stof: For local variables, I think phpstan should be smart enough to detect that they haven't been mutated (as it can only be mutated locally and phpstan can inspect that scope for mutations), and so both `if` evaluate to the same condition (and so the variable is always defined if we enter the if). So I would report it to phpstan.

src/Composer/Util/Git.php
ondrejmirtes: Yeah, PHPStan (and no other language that I know of) has a concept of "this variable if exists if this other thing is of some value", so I usually replace them with `isset`s.

stof: Well, if PHPStan reports usage of undefined variables, it should do its best to detect whether they can actually be undefined.

Here, we define the variable inside a `if ($initialClone) {` and we use it inside another `if ($initialClone) {` (same condition) without having mutated the variable used in the condition. That's something that could be detected through static analysis.
https://github.com/composer/composer/pull/8027#discussion_r266794393 is indeed more tricky, as ensuring no mutation would be much more complex (as methods called in the meantime could mutate them)

ondrejmirtes: I'll try my best to solve these simple cases.

ondrejmirtes: (PHPStan will then automatically be able to tell that this `isset` is not necessary.)

Seldaek: Thanks @ondrejmirtes - I realize it's not the greatest code when you have lots of branches like this, but it'd be nice if it could cope with it :)

src/Composer/Util/RemoteFilesystem.php
CZechBoY: I don't get where the variable comes from. Can anybody explain so we can maybe improve phpstan? Thx

naderman: @CZechBoY See http://php.net/manual/en/reserved.variables.httpresponseheader.php - it's a local variable magically populated with data by the http stream wrapper. Pretty awful for static analysis, but that's PHP ;-)

src/Composer/DependencyResolver/Solver.php
naderman: Correct @Seldaek , an unset before the while would be fine.

src/Composer/Util/RemoteFilesystem.php
ondrejmirtes: PHPStan should understand it when calling `fopen` or `file_get_contents`. Please open an issue if it doesn't in some case.

Seldaek: In this case / in this scope it is set by the getRemoteContents call at line 292, as it is passed by reference and assigned a value in that function. I guess that's maybe the case that PHPStan doesn't support yet?

src/Composer/Util/RemoteFilesystem.php
adaamz: I created issue in phpstan for this https://github.com/phpstan/phpstan/issues/2295

adaamz: btw why is there used try+catch? i think file_get_contents does not throw any exception/throwable, but i'm thinking about some custom wrappers...

dereuromark:
This could probably be revised now, since phpstan got a lot smarter in the meantime.

ondrejmirtes:
Yeah, but the need for `isset($actualLock)` still remains.

tests/Composer/Test/InstallerTest.php
dereuromark: `if ($expectLock && isset($actualLock)) {` is actually a code smell IMO.
They together dont make sense.

The current code should work fine.

dereuromark:
Looking at the code, that part should be reverted. It is wrong, and that false positive is either silenced or maybe already fixed in a modern version of phpstan.

ondrejmirtes:
What I'm saying that nothing changed in PHPStan to understand such code:

```php
if ($foo) {
    $bar = true;
}

if ($foo) {
    echo $bar; // PHPStan will still complain about possibly-undefined variable
}
```

https://phpstan.org/r/7d2631fe-cf8c-416c-a96d-b58c31b86ed5

src/Composer/Util/RemoteFilesystem.php
stof: because composer sets will throw in some cases in stream callbacks IIRC.

src/Composer/Command/ShowCommand.php
```diff
@@ -544,6 +544,11 @@ protected function getPackage(RepositoryInterface $installedRepo, RepositoryInte
         $matchedPackage = null;
         $versions = array();
         $matches = $repositorySet->findPackages($name, $constraint);
+
+        if ($matches === array()) {
+        	return array(null, array());
+        }
+
```
Seldaek, line 551:
```suggestion
```

src/Composer/DependencyResolver/Solver.php
```diff
@@ -479,6 +479,10 @@ protected function analyze($level, Rule $rule)
...
+                    throw new \InvalidArgumentException("Solver: \$literal is not defined. Please report at https://github.com/composer/composer/issues/new.");
+                }
+
```
Seldaek, line 485:
```suggestion
```

src/Composer/Installer.php
```diff
@@ -974,6 +974,8 @@ private function processDevPackages($localRepo, Pool $pool, $policy, $repositori
...
```
Seldaek, line 978:
```suggestion
```

Seldaek:
I'll rebase this and merge it, thanks

PR #8032, opened 2019-03-11
using emptyDirectory instead of remove for clearing the cache

kitingChris:
Using composer with docker it can be useful to put the cache in its own data volume.
Doing this will result in problems clearing the cache because composer relies on removing the cache-root directory.
Therefore I want to suggest using emptyDirectory instead of removeDirectory here.


Seldaek:
Thanks sounds reasonable

PR #8042, opened 2019-03-15
Fix for UNC Windows paths

Novicaine:
Made isAbsolutePath recognize Windows UNC-style absolute paths starting with \\


Seldaek:
Thanks

PR #8045, opened 2019-03-19
Correct description grammar

xuanquynh:
I am not an English native speaker but I think the `ensure` verb is used for the `Composer` subject. So, it have to be `ensures`.

The current sentence is too long. Let's separate it into 2 smaller pieces for readable reason.


Seldaek:
Technically it was correct but I agree splitting the sentence can help

PR #8052, opened 2019-03-21
Extract job packageName & constraint to variables

ravage84:
Not sure, though, this is considered CS...


Seldaek:
Thanks, merged as d2ab4f66f (fixed the indents to be 4space instead of tabs)

PR #8053, opened 2019-03-22
Generate same package mutual exclusion ("spme") rules on-demand to reduce memory.

mbaynton:
The vast majority of the total objects Composer previously generated express the
idea that package A 1.0 conflicts with package A 1.1, and 1.2, and ...
Since the number of these rules expands as n*n for the number of available
versions of each package, in many real-world evaluations these rule objects may
dominate Composer's total memory utilization.

Instead of doing that, this modification to the solver only instantiates those
rules as needed while propagating decisions. Intuitively, when we decide A 1.0,
all other packages providing A can be decided against. That's essentially what
this does by newing up conflict unit rules for the other providers of A into
RuleWatchGraph::propagateLiteral() through the RuleWatchChain iterator that
propagateLiteral() pumps from. The decision is made and the SPME rule has served
its purpose; its refcount drops to zero and the memory is reclaimed.

A side-effect of not generating the full compliment of rules upfront is that
watch chains that would only contain SPME rules are not added to the RuleWatchGraph
at all, since the watch graph is built based on the contents of the RuleSet.
This is compensated for by creating that too on demand in propagateLiteral() when
necessary.

In practice, using `--profile`, this saves 50% of peak memory utilization (~600MB down from
~1200MB) when evaluating https://github.com/drupal-composer/drupal-project, for example.

The RuleWatchChain iterator is a bit finnicky and certainly calls for unit testing, but I'd like to
gauge the maintainer's temperature on the change at this stage and...iterate 🤦‍♂️ on the PR from here.

### DIFF
--- a/src/Composer/DependencyResolver/RuleSetGenerator.php
+++ b/src/Composer/DependencyResolver/RuleSetGenerator.php
@@ -225,9 +225,6 @@ class RuleSetGenerator
 
                 if (($package instanceof AliasPackage) && $package->getAliasOf() === $provider) {
                     $this->addRule(RuleSet::TYPE_PACKAGE, $this->createRequireRule($package, array($provider), Rule::RULE_PACKAGE_ALIAS, $package));
-                } elseif (!$this->obsoleteImpossibleForAlias($package, $provider)) {
-                    $reason = ($packageName == $provider->getName()) ? Rule::RULE_PACKAGE_SAME_NAME : Rule::RULE_PACKAGE_IMPLICIT_OBSOLETES;
-                    $this->addRule(RuleSet::TYPE_PACKAGE, $this->createRule2Literals($package, $provider, $reason, $package));
                 }
             }
         }
@@ -276,7 +273,7 @@ class RuleSetGenerator
         }
     }
 
-    protected function obsoleteImpossibleForAlias($package, $provider)
+    public static function obsoleteImpossibleForAlias($package, $provider)
     {
         $packageIsAlias = $package instanceof AliasPackage;
         $providerIsAlias = $provider instanceof AliasPackage;
--- a/src/Composer/DependencyResolver/RuleWatchChain.php
+++ b/src/Composer/DependencyResolver/RuleWatchChain.php
@@ -12,17 +12,134 @@
 
 namespace Composer\DependencyResolver;
 
+use Composer\Package\AliasPackage;
+
 /**
- * An extension of SplDoublyLinkedList with seek and removal of current element
+ * Provides an iterator over rule watches for a given literal.
  *
- * SplDoublyLinkedList only allows deleting a particular offset and has no
- * method to set the internal iterator to a particular offset.
+ * The rules expressing a conflict between two different versions of packages
+ * that provide the same name are generated only as needed during iteration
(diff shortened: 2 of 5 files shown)


naderman:
This is already implemented on this branch here: https://github.com/naderman/composer/tree/multi-conflict-rule - Will ship as part of Composer2.

naderman:
I'll review this in detail and see how it compares to my solution once we finish some more of the 2.0 work.

Either way, thank you for your patch, the problem you found is definitely causing a lot of unnecessary memory use.

mbaynton:
Wow, yes that is quite similar in that it avoids instantiating the same rule objects, although we came up with different ways to work around it. And written just 3 months ago too, crazy.

My ultimate goal is to explore running Composer in very memory constrained environments, both by further reducing memory utilization and/or by progressively solving in small chunks with intermediate results that can be written out to disk. Anything else out there being done along those lines that I should know about?

naderman:
@mbaynton We're currently trying to get ahead with Composer v2.0 which includes some changes which should make it easier to improve memory consumption and performance of the solver. So I'd recommend any further work you do in this direction should happen on the 2.0 branch. I think the biggest missing change for 2.0 right now is in progress here: https://github.com/composer/composer/pull/7936

A major change that is already in 2.0 was the separation of repositoryset and pool with a new poolbuilder step in between: https://github.com/composer/composer/pull/7633 This now allows us to reduce the packages inspected at all: https://github.com/composer/composer/pull/7644 or https://github.com/composer/composer/pull/7830 and hopefully more similar changes

I don't really trust our test suite enough, as seen in https://github.com/composer/composer/pull/7946 to make this change on a regular 1.x release. So I think our focus right now should be to move ahead with 2.0 and make more changes like this one in there.

As I promised I will compare your approach and mine in detail once I finish that other PR and a few smaller changes.

mbaynton:
Thanks for the roadmap. I am indeed continuing to work on reducing memory consumption so I'll take the 2.0 branch & PRs as my new starting point.

naderman:
Hey @mbaynton - it's been quite a while but we finally got the separation of update and install merged into 2.0 so we're getting into a more final phase of 2.0 development and I'd like to get some of these types of changes merged as well now. I did run some benchmarks and it seems like the dynamic computation done in this PR is significantly slower on complex projects - see https://github.com/composer/composer/pull/8424#issuecomment-553103057 for my sylius example - so I'm going to close this PR now and focus on my variation with the new rule object type. Would definitely appreciate to hear if you ever got around to any more changes or any other ideas!

mbaynton:
Oh wow, I'm glad you encountered that slowdown @naderman, as your result of nearly a minute slower is pretty bad. I honestly didn't see anything like that back when I did this, and I think I tried the latest Sylius at the time, but I can imagine the computation slowdown might amplify in certain situations. Thanks for investigating it further.

Nothing further at this time, with luck maybe my employer will sponsor me looking at these things further someday.

naderman:
@mbaynton No worries, still a pretty good job for having spotted the problem in the first place and having built a working solution to the memory issue ;-) I think the spl doubly linked list implemented in C is simply quite a bit faster than having to use an iterator built in PHP. The number of calls to it on a large complex project can be in the millions which does add up even if it's just a small time difference.

PR #8073, opened 2019-03-31
Add instructions for wget | php for always latest web/installer

KeyWeeUsr:
GitHub allows fetching files in a more intuitive / user-friendlier way than copy&pasting hash into url (which is long and error prone anyway). Alternatively `githubusercontent.com` domain could be used though no one remembers that one until redirected to that after pressing `Raw` button IMO :smile:

### DIFF
--- a/doc/faqs/how-to-install-composer-programmatically.md
+++ b/doc/faqs/how-to-install-composer-programmatically.md
@@ -40,3 +40,12 @@ wget https://raw.githubusercontent.com/composer/getcomposer.org/76a7060ccb93902c
 
 You may replace the commit hash by whatever the last commit hash is on
 https://github.com/composer/getcomposer.org/commits/master
+
+In case you would like to get the latest version of the web installer (from
+the `master` branch), just replace the original `blob` part of a standard
+GitHub URL after opening a file and it'll return its raw content you can pipe
+to `php`:
+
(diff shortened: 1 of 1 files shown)


alcohol:
This seems redundant considering the script example above it already does exactly that (download and verify the latest installer).

KeyWeeUsr:
@alcohol The script yes, however that means you need to copy&paste the script into your repo/project which makes no sense if you can just wget the file from repo and github even provides you the url for that. See the part below the script, but with the explicit hash in the url.

alcohol:
You can just point out that you can simply replace the commit hash with `master` if you want the latest version. But this is not really recommended for programmatically installing Composer, which is why we do not advertise it.

stof:
Fetching from latest master rather than from a specific commit hash means that you would need to verify the integrity of the installer again (and then, you're back to the initial case). You don't get the security coming from the fact that existing commits cannot be altered without changing the hash if you don't specify the hash.
So I'm -1 for that change.

alcohol:
Yeah, I'm also -1 on this change. It might seem convenient, but from a security perspective this is really not advisable. So we should not advertise it to our users, because that might give them the wrong impression.

Seldaek:
Closing as per the comments above.

Issue #7779, opened 2018-11-09
Improve Error Reporting


naderman:
- [x] https://github.com/composer/composer/issues/8525
- [x] https://github.com/composer/composer/issues/7254
- [x] load full list of versions for packages in error messages, so users are not confused by errors about "no versions found" or versions they expected not showing up
- [x] Make sure packages in lock file conflicting with the root requirements (because lock is outdated) report errors properly, see test removed in https://github.com/composer/composer/pull/8531/commits/6f44350c0108c4310fe5e85cfac70fce46a33aa0
- [x] Try fixing test updated in  https://github.com/composer/composer/pull/8533/commits/ebe910c3a55b2767549350de114f3c63f243d5c3 - or at least make use of unacceptable fixed packages from the Pool to describe why something is missing
- [x] https://github.com/composer/composer/issues/6513
- [x] https://github.com/composer/composer/issues/2811
- [x] Any message of the type "- don't install symfony/symfony v3.4.17|don't install symfony/asset v4.1.6" should instead say "- install at most one of these: symfony/symfony v3.4.17, symfony/asset v4.1.6"
- [x] mention if a package would exist in a lower prio repo but it was shadowed by a higher prio one See https://github.com/composer/composer/pull/8528/files#diff-360e82e2b4be04c97674f5c1019b0970R2

Labeled "Good first issue" because some of the individual tasks may be suitable, take a look at the SolverProblemsException.

barryvdh:
Linked from #7254:
I ran into this issue also. After searching it does indeed seem that I have a conflict in my requirements and the `replace` entry in a library. The only thing which is a bit confusing is that it doesn't state WHY it cannot resolve those packages:

```
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - don't install twbs/bootstrap v3.3.0|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.1|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.2|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.4|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.5|don't install magento/magento2-base 2.2.6
    - don't install twbs/bootstrap v3.3.6|don't install magento/magento2-base 2.2.6
[...]

PR #8083, opened 2019-04-04
Attempt to improve an error report

sakshamsaxena:
Tried my hand to address one of the items listed at #7779. I believe further/future minor simplification can be carried out in the newly added `simplifyRuleText` as well. Please let me know how this looks.


naderman:
I'm not sure I'm a fan of processing the generated text after the fact? Might be better to generate the message we want to begin with?

Seldaek:
Fixed by https://github.com/composer/composer/pull/8558

PR #8085, opened 2019-04-06
Fixes #8065: Sort plugins deterministically before loading.

### LINKED ISSUE #8065: Race condition in how Composer plugins are activated leads to classmap errors

danepowell:
A second shot at fixing #8065, and follow-up to #8070

Should be pretty self-explanatory. `$packages` was previously sorted non-deterministically depending on the context in which Composer was run (sometimes alphabetically, sometimes according to dependency weight). This ensures that plugins are always loaded according to dependency weight.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -28,7 +28,6 @@ use Composer\Script\ScriptEvents;
  */
 class AutoloadGenerator
 {
-
     /**
      * @var EventDispatcher
      */
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -17,6 +17,7 @@ use Composer\EventDispatcher\EventDispatcher;
 use Composer\Installer\InstallationManager;
 use Composer\IO\IOInterface;
 use Composer\Package\AliasPackage;
+use Composer\Package\Link;
 use Composer\Package\PackageInterface;
 use Composer\Repository\InstalledRepositoryInterface;
 use Composer\Util\Filesystem;
@@ -956,27 +957,18 @@ INITIALIZER;
      *
      * Packages of equal weight retain the original order
      *
-     * @param  array $packageMap
+     * @param  array $packages
      * @return array
      */
-    public function sortPackageMap(array $packageMap)
-    {
-        $packages = array();
-        $paths = array();
+    public function sortPackages(array $packages) {
         $usageList = array();
 
-        foreach ($packageMap as $item) {
-            list($package, $path) = $item;
-            $name = $package->getName();
-            $packages[$name] = $package;
-            $paths[$name] = $path;
-
-            foreach (array_merge($package->getRequires(), $package->getDevRequires()) as $link) {
+        foreach ($packages as $package) { /** @var PackageInterface $package */
+            foreach (array_merge($package->getRequires(), $package->getDevRequires()) as $link) { /** @var Link $link */
                 $target = $link->getTarget();
-                $usageList[$target][] = $name;
+                $usageList[$target][] = $package->getName();
             }
         }
-
         $computing = array();
         $computed = array();
         $computeImportance = function ($name) use (&$computeImportance, &$computing, &$computed, $usageList) {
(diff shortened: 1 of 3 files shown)


danepowell:
Not quite sure why automated tests are failing, but before I spend time debugging that I'd appreciate a sign-off on the general approach.

danepowell:
You're right that I don't think this solves the problem for every edge case. If you have a composer.json that depends on a composer plugin that starts with letters "a" or "b" (let's call it bar) as well as composer installers, and bar doesn't declare a dependency on composer installers, this _might_ still occur. As a matter of practice, though, this seems unlikely, because with this patch composer installers is _always_ weighted such that it's installed first, even without an explicit dependency.

Even if it it doesn't fix every possible permutation of this bug, I think there's still a lot of value in this patch simply by loading plugins in a consistent and deterministic order.

danepowell:
Ready for review again.

danepowell:
@Seldaek would you mind taking another look? This has been a long and ugly problem for downstream projects with no other viable workaround, I'd love to get this (or any other solution) merged soon.

danepowell:
I opened https://github.com/composer/composer/pull/8108 as a simpler alternative to this if you think that would be easier to review.

Seldaek:
@danepowell Looks good to me now there is just one thing I am not sure is the array_reverse of the sort.. why is it? 

It sounds to me like both autoload and plugin registration should sort equally.. i.e. things with no dependencies should be loaded first, then things with dependencies should follow. The array_reverse being present in only one of them makes me think one of them is broken but maybe I missed something.

danepowell:
Thanks for reviewing. Actually the autoloader uses array_reverse too, and this just matches that: https://github.com/composer/composer/blob/8288d2c456287d4ec2d746db1d285d6bd82d52ec/src/Composer/Autoload/AutoloadGenerator.php#L403

Basically, packages right now are sorted such that if package foo depends on bar, foo comes _before_ bar in the array. But for autoloading and plugin loading to work correctly, that needs to be reversed so dependencies come first.

Let me know if I'm missing something.

danepowell:
@Seldaek please let me know if there's anything else you need to get this merged

dgpalmer:
When can we expect an official release for this?

Thanks,
Donovan

malikkotob:
Thanks for taking this on @danepowell and @Seldaek! This issue has been causing confusion and grief for awhile, and it's exciting to see it so close to the finish line. @Seldaek, any chance we can get this merged before next week?

briantully:
Great to see that we're making progress on this frustrating and all-too-common issue. @Seldaek Are there any blockers preventing this from getting merged and released?

Seldaek:
The blocker is as usual time, and wasting more of it with regular emails does not help. I know full well this needs merging, it's scheduled in the 1.8 milestone, it's not getting forgotten.

malikkotob:
Glad to hear it. Thank you for your efforts, @Seldaek!

Seldaek:
Ok, let's hope this doesn't have any unexpected adverse effects :) Thanks @danepowell for the PR and patience.

This will be available in snapshot builds shortly, I will try and finalize a 1.8 patch release early next week.

danepowell:
Awesome, thanks so much for your commitment and time. I'm pretty confident this is safe, unless someone has been writing plugins that _rely_ on things being loaded alphabetically. In which case, I think they only have themselves to blame :smile:

yyss8:
@danepowell I'm seeing the issue mentioned in acquia/blt#3148 again after updating composer to v1.9.0.
 Downgrading composer back to 1.8.6 seems to be a solution to remove those warnings at this time.

image

danepowell:
Replied to the above comment in https://github.com/acquia/blt/pull/3701

Issue #3672, opened 2015-01-23
Optimization suggestion for partial updates


stof:
Currently, a partial update loads every possible package versions for all packages in the composer.json. However, most of them are locked to the version available in the lock file because of the exact match rule added during a partial update.
This could be taken into account when fetching providers for a dependency (similar to what is done based on root package requirements currently). We could even avoid loading the package metadata from Packagist entirely in this case.
I think this could make partial updates much faster than they are currently (where the are equivalent to a full update for the resolution part).

stof:
@naderman you are probably the best one to deal with this.

fduch:
:+1:

naderman:
Yeah this is actually already on my todo list, Jordi tried loading smaller sets of packages from packagist a while ago and ran into some problems, but I'm going to try that again, as we no longer need all provider versions. So reducing what we need to load based on the lock file would be a next step to that.

francoispluchino:
@naderman Is that this optimization can load only the package version that is the latest compatible version?

##### Example:

**The package `my/library` with versions:**

- 2.0.x-dev (dev-master)
- 1.1.0
- 1.0.5
- 1.0.4
- 1.0.3
- 1.0.2
- 1.0.1
- 1.0.0

**The composer.json:**
```json
{
    "require": {
        "my/library": "1.0.*"
    }
}
```

Is that the `1.0.5` version will only be charged? Or is that the Composer will load all compatible packages (`1.0.5`, `1.0.4`, `1.0.3`, `1.0.2`, `1.0.1`, `1.0.0`)?

Currently, this is the behavior for the Install. If this is not the case, it would be really interesting to have this optimization.

Regarding the reduction in the number of loading using the file lock, I have already done this optimization in my plugin, but there is a problem when a dependency change to the down version (ex. `1.0.1` to `1.0.0`).

Do you have a technical solution (in the solver?) to avoid this problem?

stof:
@francoispluchino no, because knowing that 1.0.5 should be selected depends on all other dependencies in the graph too.

This optimization suggestion is only about partial updates. If your lock files is locking the library to 1.0.4 and the package is not whitelisted in the package update, Composer is not allowed to select anything else than 1.0.4. So it does not actually need to look for other versions.

Btw, note that for ``composer install``, Composer only reads the lock file already. So there is nothing to optimize in the dependency resolution process regarding the number of packages in this case (we are already limiting it to the needed packages only)

francoispluchino:
[...]

PR #8090, opened 2019-04-09
[v2.0]Depreciate use of Request for Pool Builder and track true ~dev packages for loadAsyncPackages

sakshamsaxena:
Hello! Apologies if I went too ambitiously ahead of myself with this PR.

Summary of changes :
* `buildPool` takes `array` instead of `Request` to obtain the installation requirements.
* Generation of such an array is scoped only within `/src/Composer/Installer.php::doInstall()` 
* Removed all superfluous calls to Packagist with each package's `~dev` versions using the above mentioned array's moderated generation.

This [gist](https://gist.github.com/sakshamsaxena/bf0a7896b9065855fee05ae01acd54b0) illustrates the difference between this version and base 2.0. 

This PR also serves as a precursor to #3672 (correct me if I'm wrong).

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -104,11 +104,6 @@ class Installer
      */
     protected $autoloadGenerator;
 
-    /**
-     * @var array
-     */
-    protected $rootRequirements;
-
     protected $preferSource = false;
     protected $preferDist = false;
     protected $optimizeAutoloader = false;
@@ -392,8 +387,11 @@ class Installer
             $repositorySet->addRepository($lockedRepository);
         }
 
-        // creating requirements request
+        // Creating requirements request
         $request = $this->createRequest($this->package, $platformRepo);
+
+        // An exclusive list of install requests to be used exclusively for Pool Builer
+        $installRequests = array();
 
         if ($this->update) {
             // remove unstable packages from the localRepo if they don't match the current stability settings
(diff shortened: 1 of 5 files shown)


sakshamsaxena:
@naderman Would be great if you could take a look :)

Seldaek:
Hey, sorry for the delay on this, been away lately. But on the surface/first look it seems good. At least the goal is definitely correct, how to best achieve it I am not sure and don't have much time to dive in the PR in depth today unfortunately.

naderman:
This will definitely conflict a lot with my other 2.0 changes, so should update & review only after that has been merged.

src/Composer/Installer.php
```diff
@@ -438,12 +453,13 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
...
                                 $constraint->setPrettyString($description . ' ' . $curPackage->getPrettyVersion() . $requiredAt . ')');
                                 $request->install($curPackage->getName(), $constraint);
+                                $installRequests[$curPackage->getName()] = $constraint;
                             }
                             break;
                         }
                     }
                 }
-            }
+            }
```
stof: should be reverted

src/Composer/Installer.php
```diff
@@ -407,12 +410,24 @@ protected function doInstall($localRepo, $installedRepo, $platformRepo, $aliases
...
+            $links = array_merge($packageRequirements, $packageDevRequirements);
 
             foreach ($links as $link) {
                 $request->install($link->getTarget(), $link->getConstraint());
             }
 
+            foreach($packageRequirements as $link) {
+                $installRequests[$link->getTarget()] = $link->getConstraint();
+            }
+            foreach($packageDevRequirements as $link) {
```
stof: I think this is wrong. `~dev` is not about packages from dev requirement, but packages allowing the usage of the `dev` stability.

stof: the proper algorithm is to add loading for `~dev` based on the configured min stability, and on the min stability overrides coming from root requirements.

stof:
Also, this has no relation to #3672 at all.

stof:
@Seldaek @naderman shouldn't this be closed entirely as invalid ? This whole PR seems to be based on a wrong understanding of `~dev` AFAICT and other changes are conflicting with the 2.x refactoring anyway.

Seldaek:
Yeah keeping open for now as reminder that we need to make sure `~dev` files are not loaded if the appropriate min-stability isn't allowed for the given package, but otherwise I guess this PR does not make a ton of sense anymore given the current 2.0 state.

PR #8088, opened 2019-04-09
Add a helper to disable process timeouts

beryllium:
This came about because Sculpin's blog skeleton was recently changed to use Composer to manage long running processes like `yarn encore --watch` and `sculpin generate --watch`. Shortly thereafter, it was observed that this triggered the 5 minute default Composer timeout.

It seemed unwise to set the entire Composer configuration to unlimited timeouts, and I felt that this narrower approach was potentially worthy of being included in Composer itself.

The helper can be included in custom script definitions by calling
`"Composer\\Config::disableProcessTimeout"`.

Example:

```json
{
  "scripts": {
    "watch": [
      "Composer\\Config::disableProcessTimeout",
      "vendor/bin/sculpin generate --watch --server"
    ]
  }
}
```

PR #8093, opened 2019-04-10
Add documentation for Composer\\Config::disableProcessTimeout

beryllium:
This change modifies the `06-config.md` section and the `scripts` article to provide information about the new `Composer\\Config::disableProcessTimeout` helper added in #8088.

It also adds some extra information about process timeouts in general, and is careful to preface the disable-timeout helper documentation with a note that Composer isn't exactly intended for that use case.


alcohol:
Thanks

PR #8096, opened 2019-04-11
fix typo

viniciosbarros:
- fix markdown typo in reserved php word


alcohol:
I think this was actually done intentionally to refer to it as the key that is used in the `satis.json` configuration file.

Seldaek:
What @alcohol said is correct

PR #8097, opened 2019-04-12
Handle absolute phar:// paths in autoload_static.php

ShiraNai7:
Composer is currently able to autoload classes from phars, but the generated paths in `autoload_static.php` are absolute and non-portable.

This PR extends the replacements done in `AutoloadGenerator` to include absolute paths starting with `phar://`.

Current `autoload_static.php` entry code (example from running the unit tests):

```php
'Sit\\' =>
array (
    0 => 'phar:///private/var/folders/3f/_931d8d7fi7_nb58fis9y7inx0ce_qm9/d/composer-test-bb0071bbbe798d079/composer-test-autoload/a/a/dir/sit.phar/src',
),
```

New entry code after changes in this PR:

```php
'Sit\\' =>
array (
    0 => 'phar://' . __DIR__ . '/..' . '/a/a/dir/sit.phar/src',
),
```

The regular expression that detects phar paths has been changed from `/\.phar.+$/` to a simple `strpos()` check (equivalent to `/\.phar.*$/`) to allow autoloading classes from phar roots (e.g. a class located at`foo.phar/Bar.php`).


tests/Composer/Test/Autoload/Fixtures/autoload_phar_static.php
```diff
@@ -0,0 +1,87 @@
...
+            'Bar' => 
+            array (
+                0 => 'phar://' . __DIR__ . '/../..' . '/dir/bar.phar/src',
+            ),
+        ),
+    );
+
+    public static function getInitializer(ClassLoader $loader)
+    {
+        return \Closure::bind(function () use ($loader) {
```
alcohol: Not entirely sure about this, since we technically still support PHP 5.3 and this became available in 5.4 AFAIK.

ShiraNai7: This code is generated by `AutoloadGenerator` and wasn't introduced in this PR (see fd2f51cea8e5f1ef978cd8f90b87b69dc1778976). `autoload_static.php` is loaded only on PHP >= 5.6.

alcohol: 🤦‍♂️ D'oh, my bad.

Seldaek:
Looks good, thanks!

PR #8100, opened 2019-04-14
[WIP] Command to interactively generate a minimal reproduce case for solver bugs

naderman:
Pretty rough, helps as-is, but before we actually ship this to anyone needs to be cleaned up a lot. The idea is that this first moves all data into composer.json directly to remove all reliance on network. It then attempts to reduce the data step by step and keeps checking if the bug in question can still be reproduced. For now it does that by asking interactively, so you can run a composer command and manually check. Ideally this would actually take an executable string as an argument which would exit 0 or non-zero depending on whether the bug is there to automate the process.

### DIFF
--- a/src/Composer/Command/GenerateTestCommand.php
+++ b/src/Composer/Command/GenerateTestCommand.php
@@ -1,0 +1,393 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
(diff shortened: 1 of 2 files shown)


alcohol:
So a similar approach to git-bisect sort of?

naderman:
Yes that's the idea in a way :-)

Seldaek:
Closing this as I don't think we'll ever clean this up :D Worst case it can still be found here..

PR #8102, opened 2019-04-15
Let curl handle proxy and cipher list itself

nicolas-grekas:
Lesson learned from https://github.com/symfony/flex/pull/486 and https://github.com/symfony/flex/pull/489


Seldaek:
Thanks!

PR #8108, opened 2019-04-24
Fixes #8065: Load composer installers first.

### LINKED ISSUE #8065: Race condition in how Composer plugins are activated leads to classmap errors

danepowell:
Fixes #8065. This is an alternative approach to #8085.

This is somewhat simpler but also a less general fix, in that it just bumps composer/installers (the most common cause of this problem) to the top of the list when loading plugins.

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -253,7 +253,15 @@ class PluginManager
      */
     private function loadRepository(RepositoryInterface $repo)
     {
-        foreach ($repo->getPackages() as $package) { /** @var PackageInterface $package */
+        $packages = $repo->getPackages();
+        foreach ($packages as $i => $package) {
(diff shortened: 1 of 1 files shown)


danepowell:
Not sure what the failure is there, I don't think it's specific to this PR.

Seldaek:
Closed by #8085

PR #8111, opened 2019-04-25
Allow to set process executor to be set from outside the VcsRepository

Toflar:
I'm trying to add my own git repo to Composer in some plugin and I need to pass env variables to the `ProcessExecutor`. Specifically, I need to be able to set the `GIT_SSH_COMMAND` env variable to make sure a specifiy SSH key is taken instead of the ones configured on the OS. It's a very specific use case so there's no need to make that configurable from the outside somehow but atm it's also pretty cumbersome to do as I have to extend the `ProcessExecutor` and the `VcsRepository` just to pass a simple environment variable to the process that's being executed.

### DIFF
--- a/src/Composer/Repository/VcsRepository.php
+++ b/src/Composer/Repository/VcsRepository.php
@@ -22,6 +22,7 @@ use Composer\Package\Loader\LoaderInterface;
 use Composer\EventDispatcher\EventDispatcher;
 use Composer\IO\IOInterface;
 use Composer\Config;
+use Composer\Util\ProcessExecutor;
 
 /**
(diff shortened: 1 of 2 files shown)


Toflar:
Failing tests unrelated imho.

Seldaek:
Passing null to the env arg makes the child process inherit all env vars AFAIK, so you should be able to simply do `putenv('GIT_SSH_COMMAND=xxx');` before calling the code. Pretty sure we do this in a few places too.

Toflar:
Indeed. Closing then, thx for the feedback Jordi :)

PR #8115, opened 2019-04-26
Fix URL resolution for non-HTTP(S) Composer repositories

pfofi:
Hello,

I had trouble using a local Composer repository created by [Satis](https://packagist.org/packages/composer/satis) with the [`"providers": true`](https://getcomposer.org/doc/articles/handling-private-packages-with-satis.md#other-options) option.

# Example
My `~/.composer/config.json` looks like this:
```json
{
    "repositories": {
        "myRepository": {
            "type": "composer",
            "url": "file:///path/to/repository"
        }
    }
}
```

And the `/path/to/repository/packages.json` (shortened):
```json
{
    "packages": [],
    "providers-url": "/path/to/repository/p/%package%$%hash%.json",
    "providers": {
        ...
    }
}
```

Trying to install a package from that repository leads to a `RepositorySecurityException`: https://gist.github.com/pfofi/f145951cc7b8ff95e403de51082edcc1

# Fix
The problem seems to be in [`ComposerRepository::canonicalizeUrl()`](https://github.com/composer/composer/blob/dd40d74bf6041ddcefa1ebe46fb9117d031b1ccd/src/Composer/Repository/ComposerRepository.php#L562) which only handles HTTP(S) URLs. This PR should fix that.

Thank you.

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -562,7 +562,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
     protected function canonicalizeUrl($url)
     {
         if ('/' === $url[0]) {
-            return preg_replace('{(https?://[^/]+).*}i', '$1' . $url, $this->url);
+            return preg_replace('{([^:]+://[^/]*).*}', '$1' . $url, $this->url);
         }
 
         return $url;
--- a/tests/Composer/Test/Repository/ComposerRepositoryTest.php
+++ b/tests/Composer/Test/Repository/ComposerRepositoryTest.php
@@ -204,4 +204,47 @@ class ComposerRepositoryTest extends TestCase
             $repository->search('foo', RepositoryInterface::SEARCH_FULLTEXT, 'library')
         );
     }
+
+    /**
+     * @dataProvider canonicalizeUrlProvider
+     *
+     * @param string $expected
+     * @param string $url
+     * @param string $repositoryUrl
+     */
+    public function testCanonicalizeUrl($expected, $url, $repositoryUrl)
+    {
+        $repository = new ComposerRepository(
+            array('url' => $repositoryUrl),
+            new NullIO(),
+            FactoryMock::createConfig()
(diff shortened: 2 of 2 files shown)


pfofi:
(Added -vvvv output)

src/Composer/Repository/ComposerRepository.php
```diff
@@ -562,7 +562,7 @@ protected function loadRootServerFile()
     protected function canonicalizeUrl($url)
     {
         if ('/' === $url[0]) {
-            return preg_replace('{(https?://[^/]+).*}i', '$1' . $url, $this->url);
+            return preg_replace('{([^:]+://[^/]*).*}', '$1' . $url, $this->url);
```
alcohol: Can you elaborate as to why you removed the case-insensitive flag?

pfofi: That's just a clean-up. The flag should not be needed for the new pattern.

alcohol:
If it is a bugfix, it should target the 1.8 branch. Can you rebase on top of that?

pfofi:
"Rebased" to 1.8.

(This is new to me. I hope I got it right.)

alcohol:
Looks like it either did not go right, or your 1.8 branch was not up to date locally. Can you update your branch?

pfofi:
Should be mergeable now according to [Compare](https://github.com/composer/composer/compare/1.8...pfofi:fix-urlCanon).

alcohol:
Yep, it all looks good. :-)

src/Composer/Repository/ComposerRepository.php
pfofi: I just realized there is another issue that leads to the same error when `$url` contains pattern references like *$0* or *$1* etc.
I'm closing this PR until I have that solved.

alcohol: Not sure what you mean? Maybe updating the test to highlight this would have helped make it more clear. Either way, it is your PR, so alright :-)

PR #8121, opened 2019-04-30
Add documentation how to run plugins manually

gregorhyneck:
By trial, I figured out that also plugins can be run manually. Would be nice if documentation would show that as well.


Seldaek:
Thanks

PR #8123, opened 2019-04-30
Fix: Bitbucket getChangeDate throws exception for branches containing a slash

glaubinix:
Calling the getChangeDate method in the BitbucketDriver with an identifier containing a slash will result in a 404 exception.

Note: this is only an issue while using the VcsRepositories directly outside of a regular composer install or update. For instance to check if a vcs repository contains a composer file.


alcohol:
Thanks

PR #8124, opened 2019-05-01
Display branches and tags if verbose is specified

### LINKED ISSUE #8113: Move reading of custom repository composer.json messages behind OutputInterface::VERBOSITY_VERBOSE
As pointed out in #8112, while iterating tags and branches of custom repositories, Composer prints out a message for each encountered. These individual messages are not behind any level of verbosity currently and are thus always printed. Because this can become quite lengthy in logs that do not support rewriting output, it would perhaps make sense to move them behind at least the first level of verbosity (`OutputInterface::VERBOSITY_VERBOSE`).

samlitowitz:
Resolves #8113

### DIFF
--- a/src/Composer/Repository/VcsRepository.php
+++ b/src/Composer/Repository/VcsRepository.php
@@ -30,7 +30,8 @@ class VcsRepository extends ArrayRepository implements ConfigurableRepositoryInt
 {
     protected $url;
     protected $packageName;
-    protected $verbose;
+    protected $isVerbose;
+    protected $isVeryVerbose;
     protected $io;
     protected $config;
     protected $versionParser;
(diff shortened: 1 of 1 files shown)


alcohol:
I think the `getCachedPackageVersion` method should also be adjusted?

alcohol:
Aside from that, looks good to me so far. :+1:

samlitowitz:
@alcohol , `getCachedPackageVersion` has been updated.

alcohol:
Nice work, thanks :-)
I'm not entirely sure if this should target master or 1.8. I think master is fine since it is not a bugfix.

Riszin:
#`68ac50ed-1576-4f56-a8e6-eb0542eb6320`

PR #8130, opened 2019-05-07
Cannot find file by class name with duplicated namespace separators

### LINKED ISSUE #8129: PSR-0 Autoloading issue - duplicated namespace separator
My `composer.json`:

```json
{
    "name": "webimpress/composer-autoload-issue",
    "require": {},
    "autoload": {
        "psr-0": {
            "": "src/"
        }
    }
}
```

I've prepared repository to demonstrate that issue: https://github.com/webimpress/composer-autoload-issue

So basically the file is included when autoloader is called with class name where namespace separator are duplicated, for example: `MyNamespace\\MyClass` instead of `MyNamespace\MyClass`.
However autoloading chain is not broken at this point.

For more details please see example in related repository.

The problem is not present with PSR-4 autoloading.

**EDIT:** There is Travis build: https://travis-ci.com/webimpress/composer-autoload-issue/builds/110903433

webimpress:
Fixes #8129

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -372,6 +372,10 @@ class ClassLoader
         // PSR-4 lookup
         $logicalPathPsr4 = strtr($class, '\\', DIRECTORY_SEPARATOR) . $ext;
 
+        if (strpos($logicalPathPsr4, DIRECTORY_SEPARATOR . DIRECTORY_SEPARATOR) !== false) {
+            return false;
+        }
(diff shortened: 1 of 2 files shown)


stof:
this is adding overhead to autoloading calls to avoid weird side-effects for people passing an invalid class name to the autoloader (not the name of a non-existent class, but a name which cannot correspond to a class). I'm not sure this is worth it.

alcohol:
Closing because if "fixes" a non-existant issue.

PR #8136, opened 2019-05-10
Fix URL resolution for Composer repositories

pfofi:
Continuation of #8115 which i messed up with a force push.

Changes since then:
- fix the pattern reference issue by ditching `preg_replace()` for `preg_match()`
- add relevant tests (remove an irrelevant one)

Sorry for the inconvenience :blush:

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -562,7 +562,11 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
     protected function canonicalizeUrl($url)
     {
         if ('/' === $url[0]) {
-            return preg_replace('{(https?://[^/]+).*}i', '$1' . $url, $this->url);
+            if (preg_match('{[^:]+://[^/]*}', $this->url, $matches)) {
+                return $matches[0] . $url;
+            }
+
+            return $this->url;
         }
 
         return $url;
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -562,7 +562,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
     protected function canonicalizeUrl($url)
     {
         if ('/' === $url[0]) {
-            if (preg_match('{^[^:]+://[^/]*}', $this->url, $matches)) {
+            if (preg_match('{^[^:]++://[^/]*+}', $this->url, $matches)) {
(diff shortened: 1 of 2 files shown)


src/Composer/Repository/ComposerRepository.php
```diff
@@ -562,7 +562,11 @@ protected function loadRootServerFile()
     protected function canonicalizeUrl($url)
     {
         if ('/' === $url[0]) {
-            return preg_replace('{(https?://[^/]+).*}i', '$1' . $url, $this->url);
+            if (preg_match('{[^:]+://[^/]*}', $this->url, $matches)) {
```
staabm: what do you think of

```suggestion
            if (preg_match('{^[^:]+://[^/]+$}', $this->url, $matches)) {
```

pfofi: I like the `^` anchor, it's stricter and faster. Added :)

The `+` quantifier does not pass the `file://` test. The hostname can be omitted and then the pattern doesn't match anymore. (file URI syntax: [RFC 8089](https://tools.ietf.org/html/rfc8089#section-2))

The `$` anchor also doesn't pass because `$this->url` can contain slashes.

src/Composer/Repository/ComposerRepository.php
```diff
@@ -562,7 +562,11 @@ protected function loadRootServerFile()
     protected function canonicalizeUrl($url)
     {
         if ('/' === $url[0]) {
-            return preg_replace('{(https?://[^/]+).*}i', '$1' . $url, $this->url);
+            if (preg_match('{^[^:]+://[^/]*}', $this->url, $matches)) {
```
stof: should use possessive quantifiers for both IMO (`[^:]++`  and `[^/]*+`) to avoid useless backtracking when not matching.

pfofi: I experimented a bit with this, but have not been able to cause any backtracking intentionally (PCRE 10.32). I don't think a recent PCRE engine will backtrack in this particular pattern. Though I don't know about older versions.

Possessive quantifiers certainly don't hurt performance and no tests are failing, so... committed. Thanks!

stof: well, maybe recent PCRE versions are able to detect automatically that these quantifiers could be possessive without any impact on the matching and turn them on. Such kind of optimization may be available.

Seldaek:
Thanks, looks good

PR #8137, opened 2019-05-10
Replace 'timeout static helper' mention with alternatives

GuilhermeRossato:
The static helper does **not work** and **raises an error** ("Method Composer\Config::disableProcessTimeout is not callable, can not call serve script") (tested on composer 1.7.2 and 1.8.0).

The 3 alternatives added are:
1. Adding to config directive (Previously missing an example)
2. Setting the COMPOSER_PROCESS_TIMEOUT=0
3. Using the `run` command to specify a `--timeout`.

### DIFF
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -224,21 +224,32 @@ to the `phpunit` script.
 Although Composer is not intended to manage long-running processes and other
 such aspects of PHP projects, it can sometimes be handy to disable the process
 timeout on custom commands. This timeout defaults to 300 seconds and can be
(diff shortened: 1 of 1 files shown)


GuilhermeRossato:
By the way, is that failing check caused by Travis not having admin permissions to send a GitHubPullRequest notification?

alcohol:
It is only available in master (snapshot). Not in any releases yet.

alcohol:
You're still free to document the alternatives I guess. But do not remove the documentation for a feature that actually does exist.

PR #8139, opened 2019-05-13
Document the alternatives to disable the default script timeout

GuilhermeRossato:
Mentioning and giving an example of the usage of the 4 options to disable the default script timeout of 300 seconds:
1. Static helper (already exists and kept).
2. Config key "process-timeout".
3. Environment variable "COMPOSER_PROCESS_TIMEOUT".
4. The "--timeout" parameter.


alcohol:
Thanks :+1:

PR #8162, opened 2019-05-29
Add a small delay to the CacheClear command

Nyholm:
Yes, this is insane. But it might make sense. 

There is no real benefit of clearing the local cache. But people (including myself) like to do it because we run in to errors. 
The most common error (for me) is that I have to wait until pagagist systems are updated. So having this "forced wait" will actually "fix" my problem.


alcohol:
No, thank you.

alcohol:
```
sleep 30 && composer clear-cache
```

PR #8168, opened 2019-06-05
Add --update option on create-project

ph-fritsche:
With this option composer will install a package in an existing root package directory as long as the package names in the composer.json are identical.

This way changes made in the source can be applied to an installation in the same way as it was initially created.

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -22,11 +22,13 @@ use Composer\IO\IOInterface;
 use Composer\Package\BasePackage;
 use Composer\DependencyResolver\Pool;
 use Composer\DependencyResolver\Operation\InstallOperation;
+use Composer\DependencyResolver\Operation\UpdateOperation;
 use Composer\Package\Version\VersionSelector;
 use Composer\Package\AliasPackage;
 use Composer\Repository\RepositoryFactory;
 use Composer\Repository\CompositeRepository;
 use Composer\Repository\PlatformRepository;
+use Composer\Repository\InstalledArrayRepository;
 use Composer\Repository\InstalledFilesystemRepository;
 use Composer\Script\ScriptEvents;
 use Composer\Util\Silencer;
@@ -79,6 +81,7 @@ class CreateProjectCommand extends BaseCommand
                 new InputOption('remove-vcs', null, InputOption::VALUE_NONE, 'Whether to force deletion of the vcs folder without prompting.'),
                 new InputOption('no-install', null, InputOption::VALUE_NONE, 'Whether to skip installation of the package dependencies.'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore platform requirements (php & ext- packages).'),
+                new InputOption('update', null, InputOption::VALUE_NONE, 'Overwrite an existing root package.'),
             ))
             ->setHelp(
                 <<<EOT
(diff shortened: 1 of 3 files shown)


alcohol:
I'm trying to understand the actual use or purpose of this...

`create-project <package> [target-dir]` basically is equal to `git clone <package-repo> [target-dir] && cd [target-dir] && composer install`. Afterwards, to update, you could simply run `git pull && composer install|update`. Why would you want to use `create-project` for this?

ph-fritsche:
I think for developing that project it is of no use.

But it's a nice feature that you can install packages via Composer and Packagist if you just want to use them. Sure you don't need a registry like Packagist but it's a nice shortcut to every project keeping a list of source repositories.
What's nice for consuming required packages is also nice for the root package when someone just intends to use and not develop it.

They don't need the complete repository - just a dist version. And pointing someone to a single command with the name of registry and package achieving just that is a nice thing.

Side-effects: Using dist is faster and you can move a source repository without telling every user.

alcohol:
How would you deal with modified files in the root project though? Right now it looks like you just overwrite them.

ph-fritsche:
Yes it does overwrite existing files. There is no information on the previous state as the root package information is not kept.

There is no definite way to determine if a vcs is present. Guessing it might be just overhead, because for someone who intends to alter files the approach using the vcs and refering to the source repository like you mentioned is the approach anyway.
Even when the initial install via `create-project` was from source, Composer explicitly asks if one wants to remove that vcs from the copy.

alcohol:
> There is no information on the previous state as the root package information is not kept.

Except, in many of such cases, people turn the target directory into their own root project (git project). Thus the `.git` directory is already there. Not sure how you would actually perform this update then?

ph-fritsche:
You simply would not. The same as you would not perform automatic updates and manual changes to one of your `vendor/groupA/projectX` folders (you installed from dist) because the update just wipes the folder.

You could use `create-project` as a shortcut to start your own project. But then you have to maintain some sort of control over any change introduced into that folder anyway, so you use your vcs.

But someone not interested in tampering with the files is stuck with that initial version or has to wipe the folder usually including the vendor directory. With the change he can just run `create-project --no-dev --update`.
(You can even use optional config files in that folder, as long as they are not part of the installed package - e.g. included in `.gitignore`)

Previously a workaround would be using a meta-package to require just that one piece of work and run `update` which would the just wipe the folder in `vendor/groupA/that-root-package-you-wanted`. Sure that's nice if you need it, but an ugly default for someone who just wants the package to be there in its current version.

Seldaek:
IMO the better way to do this is to fork the repo you want to keep updating on, so you can merge changes from upstream into your fork. Overwriting everything blindly will result in loss at some point or another depending on the complexity of the changes and what you did to the project. Having the git history and git merge ability makes this a lot easier.

ph-fritsche:
@Seldaek How does that solve the problem of being able to use Composer as a simple installer for a copy of dist, but requiring to wipe that installation with all dependencies for every little patch or to keep vcs and reference source on every such copy?

Yeah the better way if you wanted to make changes to such a dist version is forking the source and using that fork in the installation...
Then you end up with a new forked package that you can not just install and update from dist.

Seldaek:
If you really have a use case for this I'd suggest you write a command as a
plugin which can be required in your project and will do the update in
place thing.. it's not really functionality that belongs in composer itself
IMO

ph-fritsche:
If you do deny this change, I will end up writing a wrapper for Composer just to catch that case. But I will do so very reluctantly.

As mentioned above `create-project` is just a shortcut for those who intend to use it as a starting point for own changes.

But it can in fact serve a distinct purpose for those who just want to install a project.
For them Composer and Satis/Packagist form a very powerful alliance that is of no use if they are stuck with a specific unpatched version.

For everybody else the availability of an extra option using previously "not supported" methods comes at basically no costs.

Side-note: If installing a package not for developing but for using it is out of the scope of Composer, the options to install from dist or to remove the vcs when installing from source are completely nonsensical.

Seldaek:
As I said, writing a plugin would let you add the functionality in a special command of your own writing, so you can call `composer update-project` for example instead of `create-project --update`. That sounds better to me than a wrapper, but up to you.

Seldaek:
And if it helps: https://getcomposer.org/doc/articles/plugins.md#command-provider

Issue #6777, opened 2017-11-02
Feature request: Namespace interal functions and constants


Michal-sk:
Since we want composer to be as fast as possible, maybe we could namespace the internal functions and constants?

Couldn't find any other issue mentioning it. If interested i could make a pull request. There are multiple tools to achieve this:

- https://github.com/nilportugues/php-backslasher
- https://github.com/Roave/FunctionFQNReplacer 
- https://github.com/kelunik/fqn-check
- https://github.com/FriendsOfPHP/PHP-CS-Fixer (native_function_invocation )

alcohol:
@Seldaek do you think this will have any significant impact? I suppose it couldn't make things worse, so we might as well?

Michal-sk:
Thing to keep in mind is that the look of the code might get 'messier'.

Michal-sk:
Is there a general way how you guys benchmark the composer package? I am running https://github.com/FriendsOfPHP/PHP-CS-Fixer/  at this moment and would like to do a benchmark afterwards. I have Blackfire so if there is any script i could run i could get some benchmarks online.

alcohol:
There currently is not a predefined way to benchmark Composer.

We generally do not run php-cs-fixer against our entire codebase because that would invalidate/put a lot of existing pull-requests into a conflict state.

Michal-sk:
i have ran it only with the specific rule which applies the '\'. How do you run benchmarks on the codebase? If there is a simple script i can run for the benchmark i could run that trouhg blackfire and publish the result.

alcohol:
Like I said, I do not know of any way we run benchmarks ( I have never ran one myself ).

Michal-sk:
Ok got it. If anyone has some suggestions is would love to hear them.

Seldaek:
I don't think this makes enough of an impact to be measurable in the Composer context tbh.. And it makes the code way uglier than it needs to be, so I'd rather not.

keradus:
at PHP CS Fixer, we got 5% boost on PHP 7.
what about not following it in codebase, but apply it just before running tests on CI and building phar file ?

staabm:
When constants like PHP_VERSION etc. Opcache can eliminate dead code at compiletime which can have significant gains

alcohol:
@Seldaek thoughts on above latest comments?

Seldaek:
[...]

PR #8178, opened 2019-06-11
Native function and constant invocation CS

Tobion:
The first commit fixes the CS according to the current ruleset. The second/third adds `native_constant_invocation` and `native_function_invocation` fixers which could speed up composer a little. Ref. https://github.com/symfony/symfony/pull/27852

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -30,6 +30,8 @@ return PhpCsFixer\Config::create()
         'header_comment' => array('header' => $header),
         'include' => true,
         'class_attributes_separation' => array('elements' => array('method')),
+        'native_constant_invocation' => true,
+        'native_function_invocation' => array('include' => array('@compiler_optimized')),
         'no_blank_lines_after_class_opening' => true,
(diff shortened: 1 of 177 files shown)


Tobion:
I saw #6777 but don't agree with not enabling it. 

The CS rules are not enforced by composer so it does not add a burden on contributors. And it does not add a burden on maintainers because it's the CS fixing is fully automated.

Even if this only improves composer performance by 0.5% (others reported 5%), please be aware that composer runs everywhere (CI, developers machines, autoload on servers). So saving a little has big impact for everyone.

**One small step for composer, one giant leap for computers.**

Seldaek:
Right now with the 2.0 branch in the air I am definitely not going to merge this.. Way too much conflict potential.

As for saving the planet.. I am doubtful of the impact this has, but if there is an impact, 99% of it can probably be done by tweaking a couple files with hot loops, as opposed to polluting all files with useless backslashes. I don't care if it's automated, humans have to read the code still, and this doesn't do any good on that front.

PR #8180, opened 2019-06-11
fixes #8179

### LINKED ISSUE #8179: HgDriver on php5.3 Fatal error: Using $this when not in object context
We are in the process to upgrade our 5.3 codebase to 7.2, but we've hit a snag when working with composer: it simply errors out on php5.3when trying to initialize the hg driver. This code is using $this in an anonymous function, which is not allowed in php5.3.

Offending code:

```
                $command = function ($url) {
                    return sprintf('hg clone --noupdate %s %s', ProcessExecutor::escape($url), ProcessExecutor::escape($this->repoDir));
                };

                $hgUtils->runCommand($command, $this->url, $this->repoDir);
```

My `composer.json`:

```json
Imagine a link to a hg repo in here
```

```
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Do not run Composer as root/super user! See https://getcomposer.org/root for details
require.doctrine/cache : exact version constraints (1.5.4) should be avoided if the package follows semantic versioning
Checking platform settings: The xdebug extension is loaded, this can slow down Composer a little.
 Disabling it when using Composer is recommended.

Checking git settings: OK
Checking disk free space: OK
Checking pubkeys: 
OK
Checking composer version: OK
Composer version: 1.8.5
PHP version: 5.3.10
```

When I run this command:

```
php composer.phar update

```

I get the following output:

```
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Do not run Composer as root/super user! See https://getcomposer.org/root for details
[11-Jun-2019 10:28:07 UTC] PHP Fatal error:  Using $this when not in object context in phar:///src/idefix/application/composer.phar/src/Composer/Repository/Vcs/HgDriver.php on line 65

Fatal error: Using $this when not in object context in phar:///src/idefix/application/composer.phar/src/Composer/Repository/Vcs/HgDriver.php on line 65
```

It to work. ;) As far as I know Composer works with 5.3.  There is another issue with the same error #251, but it references a different problem.

alcohol:


src/Composer/Repository/Vcs/HgDriver.php
```diff
@@ -61,8 +61,9 @@ public function initialize()
                 // clean up directory and do a fresh clone into it
                 $fs->removeDirectory($this->repoDir);
 
-                $command = function ($url) {
-                    return sprintf('hg clone --noupdate %s %s', ProcessExecutor::escape($url), ProcessExecutor::escape($this->repoDir));
+                $self = $this;
+                $command = function ($url) use ($self) {
+                    return sprintf('hg clone --noupdate %s %s', ProcessExecutor::escape($url), ProcessExecutor::escape($self->repoDir));
```
Seldaek: $repoDir property needs to be made public for this to work, or probably better solution is to assign $repoDir = $this->repoDir and `use` that instead of `$self`.

Seldaek:
Hm, somehow didn't get an email for the force push it seems..

alcohol:
I think you get a notification though when I marked the issue as resolved? :man_shrugging: I figured since that was the only feedback and everything else was in order, it wouldn't hurt to merge it :d

Seldaek:
Nope marking resolved doesn't notify either (by mail at least). Anyway totally fine to merge yes was just surprised I did not see this before.

Issue #8183, opened 2019-06-11
Composer\Script\Event should have access to originating event details


freality:
When `Composer\EventDispatcher\EventDispatcher::doDispatch()` encounters a custom script, it creates a new `Composer\Script\Event` which loses any `InstallerEvent, PackageEvent, etc.` details.

Adding `getOriginatingEvent` and `setOriginatingEvent` to `Composer\Script\Event` would allow handlers access to the originating details without modifying the current constructor.

Example `composer.json`:
```json
"scripts": {
        "pre-install-cmd": [
            "DrupalProject\\composer\\ScriptHandler::checkComposerVersion"
        ],
        "pre-update-cmd": [
            "DrupalProject\\composer\\ScriptHandler::checkComposerVersion"
        ],
        "post-install-cmd": [
            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
        ],
        "post-update-cmd": [
            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
        ],
        "post-package-install": [
            "Drush\\composer\\DrushHandler::getDetails",
            "@details | xargs echo"
        ],
        "pre-package-uninstall": [
            "@details | grep 'packageName' | xargs ./bin/disable-cms-package"
        ],
        "details": "My\\composer\\Handler::getPackageDetails"
    }
```

alcohol:
You mean, that because you call `details` from within `pre-package-uninstall`, that you expect Composer to let `details` know that it is being called from `pre-package-uninstall` ?

freality:
Yes. 

I'd like the ability to chain composer information from the operation to other scripts. From what I see, this is not currently possible.

If my proposed solution is acceptable, I can create the PR.

alcohol:
Eh, I would say no. But check with @Seldaek

Seldaek:
The request sounds somewhat reasonable although the way you are using it with mixing `@details` and shell notation is not going to work.. 

Also not sure what happens with setOriginatingEvent in the case of a longer chain of `@foo` pointing to `@bar` etc..  I guess you can just call getOriginatingEvent until you have nothing left, or probably in doDispatch it should call getOriginatingEvent to make sure it resolves the originating event set on the Script\Event instance to the real one.

freality:
[...]

PR #8186, opened 2019-06-13
allow Composer\Script\Event to reference originating event

freality:
This PR implements the feature discussed in issue #8183.


Seldaek:
Thanks

PR #8200, opened 2019-06-21
Free $solver asap

nicolas-grekas:
The `$solver` variable in the `Installer::doInstall()` method holds a pretty big object.
Keeping this object around is not needed once dependencies are solved.
In my use case, this reduces the peak memory from 1.3GB to 1.0GB.
Not bad :)

Discovered while working on https://github.com/symfony/flex/pull/520


naderman:
Looked at the impact this could have on events and I think it's pretty safe to move this up :-) Thanks for finding this!

PR #8225, opened 2019-07-14
add php 7.4snapshot

andreybolonin:


alcohol:
Should add it to allowed_failures for now.

PR #8238, opened 2019-07-24
Remove unnecessary string concatenations

carusogabriel:


Seldaek:
These are there to avoid automatic tooling replacing them so they stay :)

PR #8244, opened 2019-07-24
Remove arguments that are same as default

carusogabriel:


src/Composer/Installer.php
```diff
@@ -1506,7 +1506,7 @@ public function setOptimizeAutoloader($optimizeAutoloader = false)
         if (!$this->optimizeAutoloader) {
             // Force classMapAuthoritative off when not optimizing the
             // autoloader
-            $this->setClassMapAuthoritative(false);
```
stof: I would keep it here, as this makes the code much easier to understand

localheinz: Perhaps an intention-revealing method name would be even better?

We could even remove the comment above, then.

Seldaek:
Thanks but I rather not do this, code is usually written like that for a reason, be it clarity or future proofing against default value changes, in any case not worth changing IMO.

PR #8245, opened 2019-07-24
Simplify switch with early return

carusogabriel:
As we only have one case and default, we can use an early return to simplify it.


Seldaek:
Right now we have evolutive/extensible code, if you need to add a variable you add a case and done. Your version is arguably simpler but less extensible so closing.

PR #8250, opened 2019-07-29
Add optional tag/branch converters

x000000:
This feature mainly for VCS repositories. Sometimes vendors group releases under common prefix, e.g. "ver/1.0.0" or "rel/2.0". Having converters enables simple regex search-replace to extract version string from the tags.

```json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/Seldaek/monolog",
            "tagConvert": [
                "s/^ver\\/(\\d+\\.\\d+\\.\\d+)$/$1/",
                "s/^ver\\/(\\d+\\.\\d+)\\/(\\d+)$/$1.$2/"
            ],
            "branchConvert": "s/^rel\\/(.+)$/$1/"
        }
    ]
}
```

### DIFF
--- a/src/Composer/Repository/VcsRepository.php
+++ b/src/Composer/Repository/VcsRepository.php
@@ -38,6 +38,8 @@ class VcsRepository extends ArrayRepository implements ConfigurableRepositoryInt
     protected $type;
     protected $loader;
     protected $repoConfig;
+    protected $tagConverters;
+    protected $branchConverters;
     protected $branchErrorOccurred = false;
     private $drivers;
     /** @var VcsDriverInterface */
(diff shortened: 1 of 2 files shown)


Seldaek:
I am not so sure about adding this really, I think this practice should rather be discouraged than worked around, as this doesn't teach anyone anything.. Plus the fact that open source code on packagist.org won't support this means encouraging it as long as people define their own vcs repo is even nastier.

We made it so far without needing this, so I am thinking this isn't a common enough problem to warrant such hacks.

x000000:
It's up to you, but i'm personally working with really old products that won't do changes in the repository anymore. We could fork it and create alternative tags, but doing so in legacy code is even more broken way imo. 
I wish we could use something standard, but sometimes you have to live with what you got.

Seldaek:
But for these you always have the option of using Package type repos https://getcomposer.org/doc/05-repositories.md#package-2 to exactly define the version-branch (i.e. reference) mapping as you need it to be

x000000:
It might be a solution when you have only one or two of such dependencies. But when you're dealing with dozen it becomes quite a hassle - need to manualy resolve related dependencies eliminates one of the key features provided by Composer. As a workaround I could write a script to generate composer.json, but I'd like to have this functionality built-in.

Seldaek:
I can see that you'd like to have it, but the question to me is does it make sense to add it and maintain it forever when it benefits only a tiny portion of the user base/one person.

alcohol:
This feels like a hack not worth maintaining to me :-1:

Seldaek:
Right, kinda my thinking above too.. Closing, sorry @x000000

PR #8253, opened 2019-07-31
Remove require-dev key from packages in lock

jvasseur:
I often find myself reviewing changes in composer.lock to see what was updated. When doing so changes to `require-dev` section of dependencies is just noise since it doesn't influence my project, that's why I propose to remove the key to make it easier to find the relevant information.

### DIFF
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -368,6 +368,9 @@ class Locker
             $spec = $this->dumper->dump($package);
             unset($spec['version_normalized']);
 
+            // Remove require-dev as it is root-only
+            unset($spec['require-dev']);
+
             // always move time to the end of the package definition
             $time = isset($spec['time']) ? $spec['time'] : null;
             unset($spec['time']);


Seldaek:
I'd tend to agree that this doesn't bring much info and just makes the lock files longer than needed.. The question though is how far do we want to take this, scripts and other stuff are also root-only and could be skipped. I think we need to think about it for v2.

jvasseur:
I checked a few `composer.lock` files of projects (I couldn't easily extract the list from the code) and the only root-only key that is dumped in the lock file.

Seldaek:
After thinking about this a bit more, I think we rather keep the whole info in there, to ensure the entire package information is present for installed packages, which allows tooling to be built and to display any sort of info that's needed. We don't really want to make a decision on what tooling should be allowed to be built by filtering data.

jvasseur:
But then why is `require-dev` is dumped in the while `autoload-dev` isn't (same as all the other root-only keys) ?

I think we should be consistent here and since root only key are supposed to have no influence on the parent packages removing all of them look like the good choice.

Seldaek:
We don't exclude anything at this level, autoload-dev is missing because packagist.org isn't returning it but that's arguably an oversight..

kubawerlos:
@jvasseur I had a similar problem you are describing and ended up with creating [this](https://github.com/kubawerlos/composer-smaller-lock) super simple plugin. Take a look, I'd appreciate a feedback.

Issue #7875, opened 2019-01-04
Enforce valid package names in name/require/require-dev/provide/replace/conflict/suggest


naderman:
After the warning introduced via https://github.com/composer/composer/issues/7874 - we should error on all composer commands (update, install, require, ...) if any package name in composer.json is invalid (name/require/require-dev/provide/replace/conflict/suggest) - we should probably not error on upper case letters as they are too widely spread but keep the warning for this case.

Seldaek:
Should consider #7975 before moving ahead here.

PR #8262, opened 2019-08-05
Validate schema name, type and version

aschempp:
I'm trying to validate an artifact package and was wondering why there is no validation on the schema properties like name, type and version. 

The patterns now match the documentation at https://getcomposer.org/doc/04-schema.md, but I'm not sure this is absolutely correct. For example, as far as I know, the version can have four digits (`1.0.0.0`) in the constraint, or at least that's what `normalizeVersion` usually gives me. That might also be the patch version, so `1.0.0-p1` would be correct.

Feel free to tell me if this might cause more harm than good 😉 

/cc @Toflar

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -7,8 +7,7 @@
     "properties": {
         "name": {
             "type": "string",
-            "description": "Package name, including 'vendor-name/' prefix.",
-            "pattern": "^.+/.+$"
+            "description": "Package name, including 'vendor-name/' prefix."
         },
         "type": {
             "description": "Package type, either 'library' for common packages, 'composer-plugin' for plugins, 'metapackage' for empty packages, or a custom type ([a-z0-9-]+) defined by whatever project this package applies to.",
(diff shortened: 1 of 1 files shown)


aschempp:
Maybe the version should be matched like in https://github.com/composer/semver/blob/master/src/VersionParser.php#L129 ?

Seldaek:
The version probably needs to allow one more `.\d+` and https://github.com/composer/semver/blob/master/src/VersionParser.php#L136 should be added too.. Not sure it's worth validating really as the version will get validated anyway when loading the package. Seems like getting it all in one regex will get messy and potentially introduce bugs..

Type validation 👍 

Name validation would have to be targetted for 2.0 together with https://github.com/composer/composer/issues/7875

aschempp:
You mean the name should not be validated because it is supported not to contain a slash/vendor prefix?

Seldaek:
Right now yes it can have no slash or many slashes, only packagist.org enforces it. But we are deprecating this in Composer too and 2.0 should disallow it.

aschempp:
I have removed the name validation now. The version matching is now much simpler, as it can be 1-4 digits separated by dots, followed by dash and any characters. So `1.0.0.0-foobar` would be valid, but `1.0.0beta` would not.

Seldaek:
1.0.0+foo is also valid per semver with +foo being build metadata.. That's what I meant with not worth validating. I am worried I'll forget something and then it breaks things.

aschempp:
Ok, so should we remove all validation or at least validate for starting with numbers?

Seldaek:
Sure the starting part sounds good. I guess dropping the `(-.+)?$` part makes it safe enough to use.

aschempp:
Sorry for the delay, I have updated the version matching now.

Seldaek:
Merged as 8cbc59579 - thanks!

Seldaek:
@aschempp damn how did I miss this ;) https://github.com/composer/composer/commit/b6673612e0d27424a2e35d500776e89a10e4f806

PR #8279, opened 2019-08-16
GitHub: don't display access token in debug log on redirect

glaubinix:
If you rename a private repository on GitHub which has been added as VCS repository the redirect url will contain the access token as GET parameter which will then be displayed in the log output.

```
Downloading https://api.github.com/repos/pcom-test-group/test-repo [] 
 []
Following redirect (2) https://api.github.com/repositories/82816341?access_token=my-github-token []
Downloading https://api.github.com/repositories/82816341?access_token=my-github-token []
```

Note: this is not an issue for GitLab because GitLab keeps the url the same even if you rename a project.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -301,7 +301,7 @@ class RemoteFilesystem
 
         $actualContextOptions = stream_context_get_options($ctx);
         $usingProxy = !empty($actualContextOptions['http']['proxy']) ? ' using proxy ' . $actualContextOptions['http']['proxy'] : '';
-        $this->io->writeError((substr($origFileUrl, 0, 4) === 'http' ? 'Downloading ' : 'Reading ') . $origFileUrl . $usingProxy, true, IOInterface::DEBUG);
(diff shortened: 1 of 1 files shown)


alcohol:
You are actually overwriting the variable contents now though, rather than merely filtering it out of the output. Is this intentional?

https://github.com/composer/composer/blob/91ff562f7e0bea6643fea6ea1b7941bf136d444a/src/Composer/Util/RemoteFilesystem.php#L278 explicitly adds the access token. Related possibly?

glaubinix:
Correct, there are two places where the redirect url is displayed in the log output. Once right where I strip the access token from the url and once in the `get` method.

Therefore I strip the access_token parameter from the url and pass it without it to to get method where it is added again (what you linked in your post). The access token needs to be there as GET parameter or else we won't be able to access private GitHub repositories.

As an alternative I can also strip the access token every time we display the url in the output.

stof:
I think we should strip it on display.

glaubinix:
Ok, changed to strip on display.

Seldaek:
Thanks

PR #8295, opened 2019-08-27
Optimize performance by filtering dependencies early

ausi:
Inspired by @Toflar’s #8275 this pull request tries to skip loading for packages that conflict with the root requirements. But after every [loading loop iteration](https://github.com/composer/composer/blob/4dabc17ec1e55d42d9611271d1728dcf9f79d56a/src/Composer/DependencyResolver/PoolBuilder.php#L65) in the `PoolBuilder` it extends the root requirements with new information.

This can be done because if all versions of a root dependency require the same package, this package can become a root dependency itself with a union of all version constraints. For example:

1. The root *composer.json* requires package `A` in version `‌^2.0`
2. The matching versions `2.0` and `2.1` get loaded, all lower versions get skipped.
3. `A 2.0` itself requires package `B` in `‌^3.0` and `A 2.1` requires `B` in `^3.1` therefore we can now add a root requirement for `B ^3.0‌`
4. Now when package `B` gets loaded all versions not matching `^3.0‌` get skipped directly.

I tested the version of my pull request and the results look very promising I think:

|                                      | 1.9.0                | 2.0-dev<br> 2019-08-02   | this pull request  |
|--------------------------------------|---------------------:|---------------------:|-------------------:|
| `"symfony/framework-bundle": "^3.4"` |  7.14 s <br> 265.14 MB |  3.82 s <br>  123.84 MB | 2.78 s <br>  67.03 MB |
| `"contao/managed-edition": "^4.8"`   | 33.10 s <br> 1618.04 MB | 87.53 s <br> 2802.62 MB | 8.02 s <br> 126.56 MB |
| `"contao/core-bundle": "^4.8"`       | 33.25 s <br> 1142.62 MB | 68.31 s <br> 2569.67 MB | 7.16 s <br> 77.73 MB |
| `"contao/core-bundle": "*"`          | 39.55 s <br> 1615.29 MB | 83.93 s <br> 2804.13 MB | 78.83 s <br> 2002.55 MB |

As you can see in the last example: if the requirement is very unrestricted the benefits aren’t that great, but I think in many real world use cases this change could make a huge impact.

The concept is pretty much the same as #8275 but without the need for manually adding root requires or conflicts.

The state of this pull request is just a first draft, if you like the concept I would continue working on it:

- [ ] Support `replaces` and `provide`.
- [ ] Support aliases.
- [ ] Add optimization to reduce the large number of constraints in the `MultiConstraint` objects.
- [ ] Update `loadPackages` method signature.
- [ ] Fix tests.

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -48,7 +48,6 @@ class PoolBuilder
 
     public function buildPool(array $repositories, array $rootAliases, Request $request)
     {
-        $pool = new Pool($this->filterRequires);
         $this->rootAliases = $rootAliases;
 
         // TODO do we really want the request here? kind of want a root requirements thingy instead
@@ -85,7 +84,7 @@ class PoolBuilder
                     $packages = $repository->returnPackages($loadIds[$key]);
                 } else {
                     // TODO should we really pass the callable into here?
-                    $packages = $repository->loadPackages($loadNames, $this->isPackageAcceptableCallable);
+                    $packages = $repository->loadPackages($loadNames, $this->isPackageAcceptableCallable, $this->filterRequires);
                 }
 
                 foreach ($packages as $package) {
@@ -94,6 +93,8 @@ class PoolBuilder
                     }
                 }
             }
+
+            $this->extendFilterRequires();
 
             $loadNames = $newLoadNames;
         }
@@ -133,6 +134,7 @@ class PoolBuilder
             }
         }
 
+        $pool = new Pool($this->filterRequires);
         $pool->setPackages($this->packages, $this->priorities);
 
         unset($this->aliasMap);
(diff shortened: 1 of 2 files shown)


src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -142,6 +144,60 @@ public function buildPool(array $repositories, array $rootAliases, Request $requ
...
+     */
+    private function extendFilterRequires()
+    {
+        $addRequires = array();
+
+        foreach ($this->filterRequires as $name => $constraint) {
+            $count = 0;
+            $subRequireConstraints = array();
+            foreach($this->packages as $package) {
+                if ($package->getName() === $name && $constraint->matches(new Constraint('==', $package->getVersion()))) {
```
stof: how does this deal with `provide` and `replace` (i.e. packages having multiple names) ?

ausi: The pull request is not yet finished. ☺️ 
Support for `provide` and `replace` is on my ToDo list in the issue description.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -142,6 +144,60 @@ public function buildPool(array $repositories, array $rootAliases, Request $requ
...
+            if (isset($this->filterRequires[$name])) {
+                $constraints[] = $this->filterRequires[$name];
+            }
+            $this->filterRequires[$name] = \count($constraints) > 1 ? new MultiConstraint($constraints, true) : $constraints[0];
+        }
+    }
+
+    private function optimizeConstraints(array $constraints)
+    {
+        // TODO: constraints can possibly be optimized even more, maybe this should go into the MultiConstraint class?
```
Toflar: I think it should go to `MutliConstraint`, yes. However, we have to make sure that the pretty version is not changed as this could cause confusion when information is shown to the user. :)

ausi: These constraints aren’t shown to users so the pretty string shouldn’t be an issue here.

Toflar:
I wonder why v2.0 is so much slower than 1.9 without your PR in some cases 🤔

Seldaek:
Definitely looks like a good way to go to me, but tbh I'd wait before investing more time in this until #7936 gets completed as it's already huge and risks conflicting with a lot of things.

Seldaek:
As for why 2.0 uses so much more memory and time than 1.x, that is indeed odd and should be investigated eventually but we also aren't there yet I think..

ausi:
Thanks for the feedback. I’ll leave this PR as it is then for now.

I subscribed to #7936 so that I can continue working once it got merged. Otherwise feel free to ping me once you think 2.0 is ready for such pull requests ☺️

andrerom:
Whats the state of this after #8424 was merged? Not applicable or still relevant?

ausi:
I will have to take another look at this, rebase the PR and compare the performance.

Toflar:
I already did :) It's still very relevant but the current 2.0 branch still needs some updates. Optimizations should go to the new `PoolBuilder`. Let's work on this together, ping me on Slack :)

Seldaek:
Happy new year ;) Any progress on a rebase?

ausi:
It’s on my list 🙃  
But it will take some days/weeks to find me some free time...

stof:
should this be closed now that https://github.com/composer/composer/pull/8850 has been merged ?

Seldaek:
Yup I don't think this makes a ton of sense anymore.

Issue #8309, opened 2019-09-02
VCS repositories with the format git@github.com:/my/repo behave differently to git@github.com:my/repo


greysteil:
Specifically, the former will fall back to using HTTPS to fetch the repository if git authentication fails, while the latter won't. 

We're seeing this on Dependabot, where we have git configured to use credentials when using HTTPs, but don't have any SSH tokens. URLs of the format `git@github.com:my/repo` work fine with this setup, but those using `git@github.com:/my/repo` (note the `:/`) don't.

stof:
this is because the case with `:/` does not match the regex at https://github.com/composer/composer/blob/f0d565bb6d354331cadf12712b4f6e8ebbaf51a6/src/Composer/Repository/Vcs/GitHubDriver.php#L50 (or rather at https://github.com/composer/composer/blob/f0d565bb6d354331cadf12712b4f6e8ebbaf51a6/src/Composer/Repository/Vcs/GitHubDriver.php#L270 but this is the same regex) and so won't be detected as a github repository (with the support for github-specific features like this fallback) but will use the raw git support instead.

PR #8310, opened 2019-09-02
Allow for leading slash in path part of GitHub URLs

greysteil:
Fixes https://github.com/composer/composer/issues/8309.

Credit to @stof for pointing me to the regexes that needed tweaking.

### DIFF
--- a/src/Composer/Repository/Vcs/GitHubDriver.php
+++ b/src/Composer/Repository/Vcs/GitHubDriver.php
@@ -47,7 +47,7 @@ class GitHubDriver extends VcsDriver
      */
     public function initialize()
     {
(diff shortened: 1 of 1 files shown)


alcohol:
Maybe add a test to https://github.com/composer/composer/blob/master/tests/Composer/Test/Repository/Vcs/GitHubDriverTest.php perhaps for the `supports` method? Not sure how much value there would be in that, but it would not hurt.

greysteil:
Awesome. I'm afraid my PHP skills are considerable worse than my regex skills - I've got no idea how to `phpunit`... 😬

Seldaek:
Thanks

PR #8313, opened 2019-09-04
Provide `--fixed` option for the `require` command, fixes #8303

### LINKED ISSUE #8303: Add flag to require latest but fixed version: `composer require acme/foo --fixed`
Usually, I prefer to keep fixed/specific versions of dependencies in my projects (applications):

```json
{
    "require": {
        "acme/foo": "1.0.42"
    }
}
```
There is always chance that some minor update of dependency may break compatibility even if it must not. I explicitly change version whenever I need to update specific dependency.

Every time I want to add a new dependency, I run `composer require acme/foo` and after that I edit `composer.json` to remove `^` from the version:
```diff
-        "acme/foo": "^1.0.42"
+        "acme/foo": "1.0.42"
```

I think it makes sense to consider adding option that would force composer to save `composer.json` with specific version, e.g. `composer require acme/foo --fixed`.

Some similar opinions/issues:
https://blog.martinhujer.cz/17-tips-for-using-composer-efficiently/#tip-%233%3A-use-specific-dependencies%27-versions-for-applications
https://stackoverflow.com/questions/41200946/update-composer-dependencies-to-latest-despite-static-version-numbers

What do you think?

unkind:
See #8303.

Unfortunately, I didn't find tests for `require` command.

```bash
$ php bin/composer require symfony/http-client-contracts --ignore-platform-reqs 
Using version ^1.1 for symfony/http-client-contracts
```

... with `--fixed`:
```bash
$ php bin/composer require symfony/http-client-contracts --ignore-platform-reqs --fixed
Using version v1.1.6 for symfony/http-client-contracts
```

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -400,7 +400,7 @@ EOT
         return $this->repos;
     }
 
-    protected function determineRequirements(InputInterface $input, OutputInterface $output, $requires = array(), $phpVersion = null, $preferredStability = 'stable', $checkProvidedVersions = true)
+    final protected function determineRequirements(InputInterface $input, OutputInterface $output, $requires = array(), $phpVersion = null, $preferredStability = 'stable', $checkProvidedVersions = true, $fixed = false)
     {
         if ($requires) {
             $requires = $this->normalizeRequirements($requires);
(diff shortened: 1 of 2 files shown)


unkind:
I just realized it should be disabled for the `library` package type to prevent misuses.

unkind:
Added assertion against library project types (`"--fixed" option should not be used for "library" project types.`), it's ready for review.

Seldaek:
Good call in excluding library types, but I wonder if this should also apply when the type is empty as then it defaults to library.

stof:
hmm, most custom package types are also meant to be used as dependencies, and so are closer to the `library` type than the `project` type. So should we really blacklist `library`, or whitelist `project` instead ?

Seldaek:
yeah that might make more sense indeed.. not sure if there is a case for using this on another package type, I am kinda sure someone will complain about the restriction, but project-only --fixed makes sense to me.

unkind:
I didn't write anything after last push, but it's done I think.

Seldaek:
Thanks!

PR #8331, opened 2019-09-17
8330 Correct issue where permission umask of files were not set when …

jimmy-ho:
…archiving in zip format with ZipArchiver

### DIFF
--- a/src/Composer/Package/Archiver/ZipArchiver.php
+++ b/src/Composer/Package/Archiver/ZipArchiver.php
@@ -45,6 +45,18 @@ class ZipArchiver implements ArchiverInterface
                 } else {
                     $zip->addFile($filepath, $localname);
                 }
+
+                /**
+                 * ZipArchive::setExternalAttributesName is available from >= PHP 5.6
+                 */
+                if (PHP_VERSION_ID >= 50600) {
+                    $perms = fileperms($filepath);
+
+                    /**
+                     * Ensure to preserve the permission umasks for the filepath in the archive.
+                     */
+                    $zip->setExternalAttributesName($localname, ZipArchive::OPSYS_UNIX, $perms << 16);
+                }
             }
             if ($zip->close()) {
                 return $target;


src/Composer/Package/Archiver/ZipArchiver.php
```diff
@@ -44,6 +44,18 @@ public function archive($sources, $target, $format, array $excludes = array(), $
                     $zip->addEmptyDir($localname);
                 } else {
                     $zip->addFile($filepath, $localname);
+
+                    /**
+                     * ZipArchive::setExternalAttributesName is available from >= PHP 5.6
+                     */
+                    if (defined('PHP_VERSION_ID') && PHP_VERSION_ID >= 50600) {
```
stof: `defined('PHP_VERSION_ID')` will always be true (it was added in some PHP 5.2 version)

jimmy-ho: Right, I saw similar checks so added it for safety

jimmy-ho:
> should we also set the permissions when adding an empty dir ?

Guess there should not be any harm to

jimmy-ho:
@stof code updated to always be applied on directory and files, and removed the defined() check if the minimum requirement is PHP 5.3 for composer.

stof:
Have you check the support of empty dirs here ? According to https://stackoverflow.com/questions/36287554/php-ziparchive-file-permissions you might need to handle dirs with a specific change.

jimmy-ho:
> Have you check the support of empty dirs here ? According to https://stackoverflow.com/questions/36287554/php-ziparchive-file-permissions you might need to handle dirs with a specific change.

@stof I have not had this use-case to test unfortunately, the immediate issue has been the executable flag for files being dropped... this has caused binaries which should be executable to stop working.

Happy to explore if there are examples to really validate that use-case - perhaps a new issue to address it further.

stof:
Well, try to run the archiver on a folder containing an empty dir and see whether the empty dir preserves the permission when unzipping.
I don't think we have existing issues about permissions of empty directories specifically, given that the existing archiver does not support permissions at all. But adding support for permissions should ideally be done in a working way

alcohol:
Do we need to take windows hosts into account here, or can we safely run this at all times?

a-camacho:
Any news from this ? We are also experiencing the problem where composer is installing packages and using CHMOD 666 instead of default value of server CHMOD 644. When I unzip the .zip or .tar file myself it works, but not with Composer.

alcohol:
@a-camacho this issue is unrelated to yours. You are referring to Composer extracting files from archive(s). This issue is related to archives generated by Composer.

stof:
@alcohol he's referring to extracting files from a Satis repo according to https://github.com/composer/satis/issues/577, and so extracting files generated by the code being fixed here.

Seldaek:
Thanks

PR #8332, opened 2019-09-18
Suggest running 'update --lock' when .lock file is out of date

svenluijten:
The current warning suggests running a blanket `composer.phar update`, but I believe this to be bad practice. Instead, we'll suggest the user to run `composer.phar update --lock`, to resolve the dependencies from the `composer.json` file and update the `composer.lock` accordingly.


Seldaek:
update --lock will not update any dependencies, and will most likely fail if you added any new dependency or changed requirements in a way that is not compatible with what you previously had. So IMO running a blanket update is indeed a better way, or at least a partial update targetting what you changed if you are not ready to update all the things.

svenluijten:
👍 Fair, thanks for your explanation Jordi!

PR #8334, opened 2019-09-20
Fix incorrect strip of non-php blocks in the file

zakonnic:
Example of current behaviour: https://regex101.com/r/A7jsTv/1

The selected block will be replaced with ?><?, giving incorrect results. Therefore some classes could not be found. Of course, this class is malformed, but even if it was written in correct way, the output could become corrupted after preg_replace, if there were several non-php blocks.

Lazy quantifier fixes that.


Seldaek:
Looks good to me @zakonnic - any reason this is still marked WIP?

zakonnic:
@Seldaek I just wanted to figure out why this test in AppVeyor failed, but in the past days there was no time for this. It seems not connected with my changes.

> 1) Composer\Test\Util\FilesystemTest::testJunctions
> Failed asserting that false is true.
> C:\projects\composer\tests\Composer\Test\Util\FilesystemTest.php:303

Seldaek:
Ok merging then, the appveyor build is quite unreliable unfortunately.

PR #8335, opened 2019-09-21
Extend Default composer init settings #5357

CJDennis:
- Allow author to be optional if both COMPOSER_DEFAULT_AUTHOR and COMPOSER_DEFAULT_EMAIL are set to 0 (extends pull request #6743 from issue #5357)
- Add COMPOSER_DEFAULT_MINIMUM_STABILITY
- Add COMPOSER_DEFAULT_TYPE


Seldaek:
Looks good to me although I don't quite get the use case for all these defaults..

Seldaek:
Closing, unless the rationale/benefit can be explained..

PR #8349, opened 2019-10-01
Allow more finegrained control over what caches to clear

svenluijten:
_Still a WIP, but I'm happy to receive feedback, let me know what you think about this and how you'd want this to work!_

The current idea is to add a `vendor` (and/or `package`) argument to the `clear-cache` command, so the usage will look like this:

```bash
# Clears all the caches from the `illuminate/*` namespace.
$ php composer.phar clear-cache illuminate

# Clears the `monolog/monolog` package from the cache.
$ php composer.phar clear-cache monolog/monolog
```

This should be a non-breaking change, seeing as the `vendor`/`package` argument will be optional.

## Todo
- [ ] Settle on an API that just _works_
- [ ] Write tests
- [ ] Make it work
- [ ] Update the documentation
  - [ ] In the code ("help" section)
  - [ ] In [the `doc/` folder](https://github.com/composer/composer/tree/master/doc)

### DIFF
--- a/src/Composer/Command/ClearCacheCommand.php
+++ b/src/Composer/Command/ClearCacheCommand.php
@@ -16,6 +16,7 @@ use Composer\Cache;
 use Composer\Factory;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputArgument;
 
 /**
  * @author David Neilsen <petah.p@gmail.com>
@@ -28,10 +29,16 @@ class ClearCacheCommand extends BaseCommand
             ->setName('clear-cache')
             ->setAliases(array('clearcache'))
             ->setDescription('Clears composer\'s internal package cache.')
+            ->setDefinition(array(
+                new InputArgument('vendor', InputArgument::OPTIONAL, 'Optional "vendor" to clear the cache for.'),
+            ))
             ->setHelp(
                 <<<EOT
 The <info>clear-cache</info> deletes all cached packages from composer's
 cache directory.
+
+If you specify a vendor to clear the cache for, composer will only delete
+the cached packages published under that namespace.
 
 Read more at https://getcomposer.org/doc/03-cli.md#clear-cache-clearcache-
 EOT


Seldaek:
What's the use case though?

drzraf:
Excellent idea!

Short version:
- `~/.cache/composer` is huge, but per-package cache is small
- `cache clear` sometimes solves inconsistent cache states that lead to confusing errors. The ability to **selectively** clear per-package cache allows users to **quickly** confirm (or invalidate) a possible source for a bug they would experience.

My personal use-case is the following: I deploy various feature-branches of a PHP project in different environments (database, filesystem). `Docker containers` are used for virtualization and `composer` for PHP dependencies' deployment.

It scales quite well but only because I share composer cache across various composer runs. But this  has also been **the** weak point of my setup from the beginning.
Not only parallel containers runs (= parallel runs of `composer require` of different branch of the same project) resulted in (hard-to-track) race-conditions and inconsistent cache states, but even when running the (dockerized) composer commands (almost) sequentially was the cache state still frequently problematic.

I took the habit of clearing the 10 MB of cache (related to my PHP project) in order to workaround this issue while still keeping the other >600MB living inside `~/.composer/cache`.

But there are two paths involved:
- `.composer/cache/vcs/git-gitlab.com-xxxx-foo-bar.git/refs`
- `.composer/cache/repo/gitlab.com/xxx/foo/bar`

Having a way to clear the cache without doing manual package-name interpolation externally would greatly improve the situation.

drzraf:
Why closing?
With cache being several hundreds MB (to redownload), having granularity in cache-clearing isn't total non-sense.

Seldaek:
Because I don't see the point at all. If you don't want to clear it don't clear it. Micro-managing your cache is a waste of time IMO in 2020 with disk space being so cheap.

drzraf:
I don't know how painful it would be programming-wise, but bandwidth isn't that _cheap_: In the [OECD](https://www.statista.com/statistics/604623/share-of-fibre-connections-in-broadband-oecd/) (only), less than a ⅓ of the population is using optic fiber which, for them, would make downloading hundreds of MB actually "cheap".

Issue #8316, opened 2019-09-05
Authentication debug output


glaubinix:
When I run this command:

```
composer require -vvv example-vendor/example-package
```

I get the following output:

```
Reading ./composer.json
Loading config file /Users/user/.composer/config.json
Loading config file /Users/user/.composer/auth.json
Loading auth config from COMPOSER_AUTH
Loading config file ./composer.json
Checked CA file /private/etc/ssl/cert.pem: valid
Executing command (/Users/user/code): git branch --no-color --no-abbrev -v
Reading /Users/user/.composer/composer.json
Loading config file /Users/user/.composer/config.json
Loading config file /Users/user/.composer/auth.json
Loading auth config from COMPOSER_AUTH
Loading config file /Users/user/.composer/composer.json
Loading config file /Users/user/.composer/auth.json
Reading /Users/user/.composer/auth.json
Reading /Users/user/code/vendor/composer/installed.json
Running 1.8.6 (2019-06-11 15:03:05) with PHP 7.2.21 on Darwin / 18.5.0
Downloading https://packagist-repository.example.com/packages.json
```

Based on the output it is not 100% obvious which credentials are being used for the composer command. In the output above the credentials in `COMPOSER_AUTH` will be used. Even though the `~/.composer/auth.json` appears before and after in the output.

To easier debug issues related to authentication it would be helpful if composer could output something like:

```
Running 1.8.6 (2019-06-11 15:03:05) with PHP 7.2.21 on Darwin / 18.5.0
Using http basic credentials with username: "http-username"
Downloading https://packagist-repository.example.com/packages.json
```

Ideally even a few more information but I believe this is currently not possible:
```
Running 1.8.6 (2019-06-11 15:03:05) with PHP 7.2.21 on Darwin / 18.5.0
Using http basic credentials with username: "http-username" from COMPOSER_AUTH
Downloading https://packagist-repository.example.com/packages.json
```

PR #8359, opened 2019-10-04
Debug: display used authentication for http calls

glaubinix:
Implements https://github.com/composer/composer/issues/8316

This currently output the used authentication for every http call even if the same origin gets called multiple times. If preferred I can also limit the output to once per origin.


Seldaek:
I think once per origin would be a lot cleaner if you can update it'd be good.

Seldaek:
Thanks

PR #8360, opened 2019-10-07
Command::execute() should always return an integer

derrabus:
In Symfony 4.4, a deprecation error is triggered if `Command::execute()` does not return an integer, see symfony/symfony#33747. This PR adds the necessary `return` statements.


Seldaek:
Thanks

PR #8361, opened 2019-10-07
Small refactor, fix failing test #hacktoberfest

proggga:


Seldaek:
The fix for clearstatcache makes sense I think /cc @curry684 

The other changes removing $dummy vars I don't see the point. Extracting all array keys to do the foreach is kind of a waste, so why fix something that doesn't need fixing?

proggga:
Clean code: remove useless variables, and remove strangw var delete

proggga:
Created new mr only with cache fix
#8366

PR #8366, opened 2019-10-08
Added clear cache for windows, fix tests

proggga:


curry684:
If memory serves me right (wrote this 3 years ago...) this change is irrelevant as `is_dir` and `is_link` will not 'lie' as their state is tightly coupled with PHP's native filesystem functions which properly update the caches. The `clearstatcache` is so important for the `lstat` because it has to detect the junctions which are created by invoking the external CLI commands, thus bypassing the cache and potentially making it stale.

So, I think the change is irrelevant. Is this fixing an actual issue you are having?

/cc @Seldaek

proggga:
Im trying to fix bug in appveiour, looks like maybe it can help

proggga:
From php doc: is_dir . and is_link can be cached
https://www.php.net/manual/en/function.clearstatcache.php

> Affected functions include stat(), lstat(), file_exists(), is_writable(), is_readable(), is_executable(), is_file(), is_dir(), is_link(), filectime(), fileatime(), filemtime(), fileinode(), filegroup(), fileowner(), filesize(), filetype(), and fileperms().

curry684:
Yes but they don't check the bits that can 'lie', the mask we apply on the `lstat` result can.

alcohol:
Considering there is no drawback to moving it before the methods of which it could possibly affect the outcome, I don't see why this change is irrelevant? Worst case scenario, nothing gained, nothing lost. The bug solving potential is enough IMHO.

alcohol:
@proggga bugfixes should target the latest release branch though (1.9 in this case). Can you adjust and rebase?

proggga:
Adjusted and rebased

curry684:
I'm fine with the change, I was just explaining that the bug solving potential should be close to zero if PHP is working as documented. That in itself is always a gamble so indeed fine to change.

proggga:
Sorry for asking, but will you merge it or no?

PR #8367, opened 2019-10-08
Add messages to junction tests to see failures

johnstevenson:
I'm trying to track down the seemingly random failures on Appveyor. This first step is to see if there is a single failing condition (which I doubt).

### DIFF
--- a/tests/Composer/Test/Util/FilesystemTest.php
+++ b/tests/Composer/Test/Util/FilesystemTest.php
@@ -300,16 +300,16 @@ class FilesystemTest extends TestCase
 
         // Create and detect junction
         $fs->junction($target, $junction);
-        $this->assertTrue($fs->isJunction($junction));
-        $this->assertFalse($fs->isJunction($target));
-        $this->assertTrue($fs->isJunction($target . '/../../junction'));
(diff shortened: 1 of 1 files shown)


johnstevenson:
Ha ha, perfect!

Despite the appveyor test on my fork succeeding (https://ci.appveyor.com/project/johnstevenson/composer/builds/27966702), the one for this PR failed (https://ci.appveyor.com/project/Seldaek/composer/builds/27966930)

More confusingly, it failed trying to remove the junction stating that:
```
... is not a junction and thus cannot be removed as one
```
having successfully tested that it is a junction 3 times.

Seldaek:
See also #8366

johnstevenson:
Yeah, I saw that after I'd opened the PR. I don't know if moving the clearstatcache call before the not-a-directory-or-is-a-link check would make much difference.

Would this PR be better on master or 1.9?

Seldaek:
1.9 for bugfixes. The other PR seems to help so far but it's hard to say as
it's somewhat random.

curry684:
The test would fail if the underlying filesystem is not NTFS. Unheard in the real world but possible...

Issue #8364, opened 2019-10-08
Composer show is not displaying the homepage


stof:
When using `composer show some/package`, a bunch of metadata about this package is displayed. But the homepage is not part of it. This looks quite unexpected to me.

mathiasdonoso:
Hello, I have created a PR for this issue [here](https://github.com/composer/composer/pull/8375).

PR #8371, opened 2019-10-10
 #8364 Add Homepage informations in command composer show

romdum:
Asked in issue #8364. 
I added homepage url between license and source.


alcohol:
Doesn't have to be wrapped in a `<comment></comment>` I think

alcohol:
This one was first, but I would personally prefer the solution in #8375

Seldaek:
Fixed by #8375 but thanks for the PR anyway!

Issue #8280, opened 2019-08-16
interactive dependencies definition


hubroot:
Search for a package: testrepo
Found 15 packages matching test repo
[0] test/repo
...

Enter package # to add, or the complete package name if it is not listed: 0

Search for a package: testrepo
Found 15 packages matching test repo
[0] test/repo
...

i think the scenario shouldn't show the same results as test/repo has been chosen before.

alcohol:
So you want to filter out current requirements?

Seldaek:
Fixed by https://github.com/composer/composer/pull/8372

PR #8372, opened 2019-10-12
Hide installed packages from suggestions when package is not found

xy2z:
This solves #8280

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -436,11 +436,40 @@ EOT
         }
 
         $versionParser = new VersionParser();
+
+        $composer = $this->getComposer(false);
+        $installedRepo = ($composer) ? $composer->getRepositoryManager()->getLocalRepository() : null;
         $io = $this->getIO();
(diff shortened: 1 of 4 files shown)


src/Composer/Command/InitCommand.php
```diff
@@ -792,7 +792,13 @@ private function findSimilar($package)
         }
         $similarPackages = array();
 
+        $installedRepo = $this->getComposer()->getRepositoryManager()->getLocalRepository();
+
         foreach ($results as $result) {
+            if ($installedRepo->hasPackageName($result['name'])) {
```
Seldaek: Pleasse rather use `->findPackage($result['name'], '*')` to avoid adding a new method to all repositories.

xy2z: Fixed

xy2z:
I'll look into it 👍

xy2z:
@Seldaek This should be fixed now

Seldaek:
Thanks, simplified the code a little so it only loops once instead of on every package search, but otherwise patch was pretty good

PR #8377, opened 2019-10-16
Fix misc phpdoc and strpos arg order nits for suggestions

TysonAndre:
https://www.php.net/strpos has the signature
`strpos ( string $haystack , mixed $needle [, int $offset = 0 ] ) : int`
(The needle is usually the constant)

`strpos('/', $suggestion)` would only be false for `'/'` and `''`.

So the existing check would just not suggest **anything** that was
already installed (from pecl, built-in, or composer).

The intent seems to be to not suggest non-vendored php packages
that were already installed. (b20cc22ebb1d9f8ed467f0ea353391be7442ad3d)


Seldaek:
Thanks

Issue #8354, opened 2019-10-02
Config option to disable lockfile generation?


aboks:
I know that opinions differ on whether the `composer.lock` file should be versioned for libraries. I don't really want to pick a side here; fact is that there's quite a lot of people who omit the lock file from version control. For those projects composer still generates a lock file, so that you end up with a local lock file that may be outdated compared to the`composer.json` versioned in the repository. This means that after pulling you either need to throw away your lock file before running `composer install` or run `composer update` instead. This seems less than ideal, especially as the workflow now differs between projects with and without a versioned lock file.

How would you feel about adding a `no-lock` config option to disable lockfile generation? By putting that into the `config` section of `composer.json` one can prevent a lockfile from being generated at all, so that they can still run `composer install` safely after pulling from VCS.

I'm willing to try and give the implementation a shot, just wanted to check first if this is something that you would accept as a PR at all.

Seldaek:
The use case makes sense, but might conflict with #6822 - or at least should be taken into consideration there too.

stof:
Well, if you don't generate a lock file, any install you run is actually an update. So I would still be in favor of #6822. We could make explicit that when you don't generate a lock file, you can only ever update dependencies, not install them.

naderman:
Essentially this would be an option to delete the lock file on a composer update run after the automatic install following the update has completed.

aboks:
Yes, essentially that would be it. Although (if I understand correctly), composer does not actually write the lockfile do disk between the 'update' and the 'install' phase, right? 

In relation to #6822, the way I see it is that this option would currently prevent generating a lockfile on both `install` and `update`. When #6822 is implemented it isn't possible to run `install` without a lockfile anymore. That means the option then only applies to `update`. The absence of a lockfile (even after running `update`) then nicely signals that the package is meant to be used with `update` only, always using the latest versions of dependencies.

My6UoT9:
I got composer 9.1 and I have set lock to false, as described: https://getcomposer.org/doc/06-config.md#lock
But it generates a lock file every time, does not matter if update or install.

How to fix, I am pretty sure it did work for a while.

trickeyone:
I was kinda confused about the config option, too, per what @My6UoT9 said. Are the docs just out-of-date or is this a bug? The docs say that setting:
```json
{"config": {"lock": false}
```
should disable the lock file generation, but it is still generated even if you set it to `false`.

Seldaek:
[...]

PR #8384, opened 2019-10-20
Config option to disable lockfile generation

aboks:
This implements the feature proposed in #8354 to make workflows without a versioned lockfile less error-prone.


Seldaek:
Thanks

PR #8385, opened 2019-10-20
Add `brew install composer` for macOS install

christiangenco:
In googling around for how to install composer (because I was trying to install [tailwindo](https://github.com/awssat/tailwindo)) I first found this official installation guide.

These installation instructions looked *way* to complicated for what I'm used to for installing a package manager on macOS. Usually I'd just `brew install npm`. Sure enough, I found [a guide to installing composer on macOS that recommended just using `brew install composer`](https://pilsniak.com/install-composer-mac-os) which seemed to do the trick.

This pull request adds a separate section for macOS installations that recommends just using `brew install composer`.


alcohol:
We do not support brew so there is no need to document this.

PR #8395, opened 2019-10-29
Don't show root warning for docker containers

vsychov:
When composer ran inside docker container, it's show "Do not run Composer as root/super user! See https://getcomposer.org/root for details" warning. 

For official composer docker, env "COMPOSER_ALLOW_SUPERUSER" used for remove this warning, but not very useful to use it for all docker images, who use composer inside.


Seldaek:
Thanks, merged as 99e23d526 in 1.9/master

PR #8397, opened 2019-10-29
make optimized autoloader respect PSR standards

### LINKED ISSUE #7352: Autoload optimizer violates PSR-4 and changes autoloading algorithm

and800:
fix #7352

This is a revived pull request #7421. I accidentally deleted my fork repo, so that PR became readonly. This one is ready for review now.

Quick recap: optimized dump-autoload adds every found class to the classmap, even if it is not under psr0/psr4 rules. My proposition is to filter the class list extracted from every file and fire a warning for any class with misaligned namespace and location. If there is a proper class in the file, improper classes from that file are skipped silently. If a class has even different namespace, it is also skipped with no warnings(that's the current behavior, I kept it).

I've also written a simple test for it.


Seldaek:
Thanks, looking good

crearc:
I run optimized autoloader in all environments. The change happened 20 minutes before I was deploying so hadn't even ran it between that time locally. My local version wasn't updated that far ahead either, self-update only updates to stable versions. But yeah what you're saying is correct. 

I would have caught it in staging since I do canary deployments, but I deployed staging hours before the 1.10-dev rollout. Fun times!

and800:
ok it was rather stupid stuff from me, maybe it shouldn't have been so hardcore.

> I highly suggest you switch this to either 1) display warning but do not skip the import or 2) throw an error and bail. Ignoring an import and not bailing out can lead to bad things, per my example.

Option (1) seems like the best solution for now. Option (2) has an issue that it would throw the warnings even if a class was never meant to be autoloaded.

So we may stick to throwing the warnings for several releases, and decide its destiny later. Formally it isn't a breaking api change, no need to wait for the next major release, but let's give some time to users.

ameenross:
@and800 Honestly, I disagree. You fixed a bug. Some people were relying on said bug. Most people will notice once they update their composer to the fixed version in DEV. Even if they don't, the chance of them being bitten by it still isn't large, as they would need to run an old version of composer up until production.

@crearc Basically did the opposite of what's desirable: using an older version of composer in dev and a brand spanking new version in prod (unstable master version, no less!).

Again, you fixed a bug. Let's not be backwards compatible with bugs, that should be left to Microsoft.

Seldaek:
https://github.com/composer/composer/commit/f6b8643dcdbad43d0373ec03089634daed32b324 changes it to only warn for now, given this can break apps it's probably best to give a little bit of a warning period.

Issue #8414, opened 2019-11-06
GitHub for repository type git doesn't use setup authentication for https://github.com/acme/private-repository


glaubinix:
Hello,

If I add a private GitHub repository as type git with a url like `https://github.com/acme-org/my-private-repository` then composer will fail if my ssh key doesn't have access to the GitHub repository, even if I have a GitHub token defined in my auth.json. Access with the token works when using `https://MY_GITHUB_TOKEN@github.com/acme-org/my-private-repository` but the composer GitDriver is not trying to access it.

My `composer.json`:

```json
{
    "repositories": [
        {"type": "git", "url":  "https://github.com/acme-org/my-private-repository"}
    ],
    "require": {
        "acme-org/my-private-repository": "dev-master"
    }
}

```

My `auth.json`:
```json
{
    "github-oauth": {
        "github.com": "MY_GITHUB_TOKEN"
    }
}

```

When I run this command:

```
composer install/update
```

I get the following output:

```
composer install
Loading composer repositories with package information


  [RuntimeException]
  Failed to clone https://github.com/acme-org/my-private-re
  pository via https, ssh protocols, aborting.

  - https://github.com/acme-org/my-private-repository
    Cloning into bare repository '/Path/to/.compose
  r/cache/vcs/https---github.com-acme-org-my-private-reposi
  tory'...
    remote: Invalid username or password.
    fatal: Authentication failed for 'https://github.com/ac
  me-org/my-private-repository/'

  - git@github.com:acme-org/my-private-repository
    Cloning into bare repository '/Path/to/.compose
  r/cache/vcs/https---github.com-acme-org-my-private-reposi
  tory'...
    git@github.com: Permission denied (publickey).
    fatal: Could not read from remote repository.

    Please make sure you have the correct access rights
    and the repository exists.
```

And I expected this to happen:

Composer install/update should use the GitHub authentication as defined in the auth.json for git if defined.

stof:
can you try with the latest snapshot version (install it with `composer self-update --snapshot`; you can revert to another release channel later using that command again with the right flag) ? @Seldaek made some fixes recently regarding the github/gitlab/bitbucket auth.

glaubinix:
sorry, forgot to add that information. This happens with `Composer version 1.10-dev (1.10-dev+7e929503671fc94de8ba16d665936e63f46d5a56) 2019-11-06 10:27:40` but also latest stable. From what I can see in the git log the GitDriver was last edited 9 months ago.

[...]

PR #8418, opened 2019-11-06
Git: fix authentication handling for private GitHub repositories

glaubinix:
Addresses: https://github.com/composer/composer/issues/8414

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -85,7 +85,9 @@ class Git
             }
 
             // failed to checkout, first check git accessibility
-            $this->throwException('Failed to clone ' . $url . ' via ' . implode(', ', $protocols) . ' protocols, aborting.' . "\n\n" . implode("\n", $messages), $url);
+            if (!$this->io->hasAuthentication($match[1]) && !$this->io->isInteractive()) {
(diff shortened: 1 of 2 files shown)


src/Composer/Util/Git.php
```diff
@@ -96,21 +98,21 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
         $auth = null;
         if ($bypassSshForGitHub || 0 !== $this->process->execute($command, $ignoredOutput, $cwd)) {
             // private github repository without ssh key access, try https with auth
-            if (preg_match('{^git@' . self::getGitHubDomainsRegex($this->config) . ':(.+?)\.git$}i', $url, $match)
+            if (preg_match('{^(git)@' . self::getGitHubDomainsRegex($this->config) . ':(.+?)\.git$}i', $url, $match)
                 || preg_match('{^(https?)://' . self::getGitHubDomainsRegex($this->config) . '/(.*)}', $url, $match)
```
glaubinix: without the additional brackets git and https urls have different matching groups

Seldaek: Probably should rather drop the brackets on this one as match[1] is never used below?

glaubinix: Adjusted

PR #8420, opened 2019-11-08
Update ClassLoader.php

AlexRill:
Sorry, i don't speak english, watch examples.

Example 1:
Autoload file:

spl_autoload_register(
    function ($classname) {
        include __DIR__ . '/' . str_replace('\\', '/', $classname) . '.php';
    }
);

application file:

use App;
require 'autoload.php';
$app = new App;

This example don't worked.

Example 2:
Autoload file:

spl_autoload_register(
    function ($classname) {
        require __DIR__ . '/' . str_replace('\\', '/', $classname) . '.php';
    }
);

application file:

use App;
require 'autoload.php';
$app = new App;

This example worked!

include vs require. Require - worked, include - don't worked.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -441,5 +441,5 @@ class ClassLoader
  */
 function includeFile($file)
 {
-    include $file;
+    require $file;
 }


Seldaek:
Sorry but without more explanation as to why this is needed, I don't think we can accept this change.

Use google translate or something if needed, but we need more infos. The examples make no sense to me.

PR #8421, opened 2019-11-08
matching 'require' to repo URL

cliffordp:


doc/05-repositories.md
```diff
@@ -216,7 +216,7 @@ Example assuming you patched monolog to fix a bug in the `bugfix` branch:
     "repositories": [
         {
             "type": "vcs",
-            "url": "https://github.com/igorw/monolog"
+            "url": "https://github.com/monolog/monolog"
```
Camwyn: This url is dead,
did you mean https://github.com/Seldaek/monolog ?

xuanquynh:
This is just an example of documentation. Nothing need to be changed.

cliffordp:
> This is just an example of documentation. Nothing need to be changed.

@xuanquynh the documentation is incorrect, as it does not match, regardless of whether it's `monolog/monolog` or `foo/bar`. I was simply matching this line to the `"require"` one below it.

naderman:
You're deleting the very point of this documentation example: You can use a repository which does not match the package name and which is not the original package's repository. By applying this patch you would actually make it so the example no longer reflects what the section is trying to explain.

cliffordp:
Ah, I'll have to take another look. Thanks for responding.

PR #8424, opened 2019-11-12
New Multi Conflict Rule for transitive conflicts, to reduce memory use by rules

naderman:
A large part of rule memory is used up with rules representing transitive conflicts, or mutual exclusion. This PR implements a new type of rule representing a combination of all these individual rules into a single object, greatly reducing memory usage and consequently also improving performance.

This PR is an alternative approach to https://github.com/composer/composer/pull/8053 - the upside is that its approach is more generic than just different versions of one package name. At the same time the alternative approach in #8053 may reduce memory more effectively by getting rid of parts of the watch graph entirely. So will need to do some benchmarking and explore whether these two approaches can maybe even be combined.

Sylius example before (on 1.9 branch):
```
[492.1MiB/75.38s] Memory usage: 492.1MiB (peak: 2943.1MiB), time: 75.38s
```

After (on 2.0 branch combined with this PR):
```
[167.0MiB/34.28s] Memory usage: 166.97MiB (peak: 1552.14MiB), time: 34.28s
```

### DIFF
--- a/src/Composer/DependencyResolver/MultiConflictRule.php
+++ b/src/Composer/DependencyResolver/MultiConflictRule.php
@@ -1,0 +1,99 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\DependencyResolver;
+
+use Composer\Package\PackageInterface;
+use Composer\Package\Link;
+
+/**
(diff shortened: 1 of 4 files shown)


naderman:
Alternative #8053 merged into 2.0 branch on the same sample:
```
[167.0MiB/107.44s] Memory usage: 166.96MiB (peak: 1627.68MiB), time: 107.44s
```

While the memory reduction on the latter seems close to this PR, the slowdown resulting from the new computation seems to be so significant that it's probably not worth exploring the dynamic nature of that change any further.

naderman:
Just to verify I reran the tests with the new 2.0 option COMPOSER_DISABLE_NETWORK=1 and the performance was hardly impacted, so definitely sticking with the solution in this PR.

PR #8431, opened 2019-11-16
Hide commit signature for git log.

cyrilverloop:
Hi,

When you have `git config --global log.showSignature true`. The commit signature is displayed for each `git log` command.

If you require a project with a signed commit (eg: "foo/project": "dev-master"), you are unable to install or update.

Adding `-c log.showSignature=false` to the command seems to solve the problem.


Seldaek:
Thanks, merged in 1.9 and master as 0bd3f2769

PR #8432, opened 2019-11-16
Allow Symfony 5

kylekatarnls:


kylekatarnls:
The `continuous-integration/appveyor/pr — AppVeyor build failed` error seems not related to this change as it downloads symfony 2.* version.

Seldaek:
Thanks

peter-gribanov:
When can i expect a release with this feature?

Seldaek:
When it's done. But most likely soon.

TomasVotruba:
This block my Symfony 5.0 build :/

brianfreytag:
@Seldaek Do you have an update on this? I'm still trying to get my 5.0 upgrade going, but this is holding us up.

kylekatarnls:
🙏 Please a minor version with this PR.

Seldaek:
1.10.0-RC is out now.. Please note that you could also have required `"composer/composer": "^1.10@dev"` to use pre-release code without pestering us about a release..

Anyway now you can use `"composer/composer": "^1.10@RC"` which is rather future proof, or if you are using composer as a CLI util `composer self-update --preview`

kylekatarnls:
Thanks @Seldaek but sorry it's not the same, when you deliver a library using composer/composer and wanting to be compatible with symfony 5, requiring `"^1.10@dev"` won't make your library compatible with Symfony 5 for users that stay on `"minimum-statbility": "stable"` you can only choose `@dev` if you're on the end of the chain. And it requires to undertake the instability risk.

So I would say, providing dev/alpha/beta/RC channels is nice and I thank your for it. But it can't exactly substitute stable releases. Until a stable release, composer is still not compatible with Symfony 5.

Of course releasing an RC before the stable and waiting for feedback on it is the right move.

So thanks!

Seldaek:
Yup fair enough for libraries this is trickier, although it would let you fix your build at least.. and let users who really want also require composer/composer `@dev` or `@RC` to allow it to install. But anyway I hope there are not too many issues with the RC and we can move on to a final release soon.

barryvdh:
Don't want to rush/pressure you, but do you happen to have an ETA for the release of 1.10? Since it's also not compatible with Laravel 7. If not anytime soon, I'll just workaround using a different Classmap generator.

Seldaek:
Early next week 90% sure :)

Seldaek:
1.10.0 is out now

barryvdh:
Thank you very much!

PR #8442, opened 2019-11-22
Fix typos

james-becker:
Not too much explanation required. The use of a slash was confusing to me.


xuanquynh:
https://getcomposer.org/doc/01-basic-usage.md#package-names

That is the name convention. A package name contains vendor name and project name, they are separated by a slash.

Seldaek:
Yup.. thanks but this isn't quite an improvement.

adaamz:
mb better would be `me/package` or `vendor/package`?

PR #8444, opened 2019-11-23
Removed the filters from the pool

Toflar:
After discussions with @naderman at #SymfonyHackday, here's the removal of the `$filterRequires` on the `Pool`. I've also renamed it to `$rootRequires` because that's what it's really is. It was only ever used for `Problem` as well as `Rule`, so it had no influence on the resolving process but it was there to display information only. This information, however, should be provided by the new `RepositorySet` and not the `Pool` itself.

### DIFF
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -32,20 +32,16 @@ class Pool implements \Countable
     const MATCH = 1;
     const MATCH_PROVIDE = 2;
     const MATCH_REPLACE = 3;
-    const MATCH_FILTERED = 4;
 
-    protected $providerRepos = array();
     protected $packages = array();
     protected $packageByName = array();
     protected $packageByExactName = array();
     protected $priorities = array();
     protected $versionParser;
     protected $providerCache = array();
-    protected $filterRequires;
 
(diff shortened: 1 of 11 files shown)


naderman:
@Toflar phpstan failures: https://travis-ci.org/composer/composer/jobs/615977317?utm_medium=notification&utm_source=github_status - can you fix those?

Toflar:
Done, had to rebase to 2.0 too. The Windows build failure looks unrelated.

Seldaek:
I think it looks good to merge but it means we need to fix Problem/Rule to be able to query the RepositorySet right? Just checking I got things up to date too.

Toflar:
Yes, we do. I don't know if Nils wants to just inject that into the problem/rule right away or if there's anything else in mind so when sitting at the hackathon he just told me to do this here first :)

Seldaek:
Ok thanks, merging then, will open an issue to track the TODOs as they are quite important

naderman:
It looks to me like $rootRequires in this version no longer serves any purpose at all and can be removed. It's unnecessarily computed and passed around but never used anywhere?

If we want to filter further in the pool builder we can access the root requirements there directly, no need for any of the code that's calculating it.

PR #8453, opened 2019-11-28
Add funding field to composer.json and composer fund command

naderman:
Packages can specify a list of funding options each with a type and URL. The type is used to specify the kind of funding or the platform through which funding is possible.

Thanks to the symfony/thanks implementation which uses GitHub funding files https://github.com/symfony/thanks/pull/73/files we decided to add a general funding feature to Composer. This PR is based on the implementation in npm: https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md and https://github.com/npm/cli/pull/273 which has its roots in various previous projects/ideas/experiments: https://github.com/npm/rfcs/blob/latest/accepted/0017-add-funding-support.md#prior-art

Still missing, now that the field is supported and loaded and saved everywhere:

- [x] composer fund command which outputs a list of funding URLs for all the packages you have installed, grouped by vendor/package/URL
- [x] adding an output line to composer update/install which informs you how many of your installed packages have funding information and to run composer fund to get information on how to fund those packages
- [x] support on packagist.org https://github.com/composer/packagist/pull/1052
- [x] support for loading .github/FUNDING.yml information via GraphQL API in GitHubDriver if no funding information is defined in composer.json

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -285,6 +285,7 @@ An example:
             "type": "other",
             "url": "https://www.doctrine-project.org/sponsorship.html"
         }
+    ]
 }
 ```
 
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -258,6 +258,38 @@ An example:
 
 Optional.
 
+### funding
+
+A list of URLs to provide funding to the package authors for maintenance and
+development of new functionality.
+
+Each entry consists of the following
+
+* **type:** The type of funding or the platform through which funding can be provided, e.g. patreon, opencollective, tidelift or github.
+* **url:** URL to a website with details and a way to fund the package.
+
+An example:
+
+```json
+{
+    "funding": [
+        {
+            "type": "patreon",
+            "url": "https://www.patreon.com/phpdoctrine"
+        },
+        {
+            "type": "tidelift",
+            "url": "https://tidelift.com/subscription/pkg/packagist-doctrine_doctrine-bundle"
+        },
+        {
+            "type": "other",
+            "url": "https://www.doctrine-project.org/sponsorship.html"
+        }
+}
+```
+
+Optional.
+
 ### Package links
 
 All of the following take an object which maps package names to
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -522,6 +522,24 @@
                 }
             }
         },
+        "funding": {
+            "type": "array",
+            "description": "A list of options to fund the development and maintenance of the package.",
+            "items": {
+                "type": "object",
+                "properties": {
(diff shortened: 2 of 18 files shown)


res/composer-schema.json
```diff
@@ -518,6 +518,24 @@
                 }
             }
         },
+        "funding": {
+            "type": "array",
+            "description": "A list of options to fund the development and maintenance of the package.",
+            "items": {
+                "type": "object",
+                "properties": {
```
stof: I would add `"required": ["type", "url"]` here, to ensure that each item in this array has these 2 properties set. IMO, it does not make sense to allow an item without type or without url.

naderman: @stof Reasoning was as follows: We can technically infer the type from most URLs e.g. github for github sponsors urls, tidelift for a tidelift package url, patreon for a patreon user url, but we may not always be able to, so documentation says you should define it, but maybe we change our mind in a future composer version?

I don't want to make not having a URL a failure right now since that would mean if we ever add support for other ways of specifying funding than a URL in the future, old composer versions will error on those new packages.

I'm a bit unclear on where we actually have to use loose validation to be reasonably compatible with future changes, the schema, the validating array loader, the regular array loader - cc @Seldaek

doc/04-schema.md
```diff
@@ -258,6 +258,38 @@ An example:
...
+        },
+        {
+            "type": "tidelift",
+            "url": "https://tidelift.com/subscription/pkg/packagist-doctrine_doctrine-bundle"
+        },
+        {
+            "type": "other",
+            "url": "https://www.doctrine-project.org/sponsorship.html"
+        }
+}
```
localheinz:
```suggestion
    ]
}
```

res/composer-schema.json
Seldaek: ConfigValidator is the best place IMO as that only impacts the `validate` command. ValidatingArrayLoader is used in some contexts to load package data while validating it, so that'll break stuff like packagist updates.. same for schema it'll potentially break more places.

alcohol:
Do we want to add an optional flag similar to `--no-suggests`, e.g. `--no-funding` in this case?

naderman:
@alcohol since this only prints those two lines with how many packages there are / how to get info, rather than listing all the URLs, I'm not sure we even need that?

alcohol:
I'm definitely not saying we need this. But I can imagine some people will ask for this. Merely asking if we want to anticipate this preemptively or simply wait and see how we deal with it if and then.

stof:
I suggest waiting, especially given this is only about a 2 lines in the output, not about a list which could get very long.

PR #8458, opened 2019-11-29
Rewrite NoProxyPattern to include IPv6

johnstevenson:
This includes two breaking changes:
- the hostname is not resolved in the case of an IP address.
- a hostname with a trailing period (FQDN) is not matched.

This brings the basic implementation in line with curl behaviour, with 
the addition of full IP address and range matching (curl does not
differentiate between IP addresses host names).

The NO_PROXY environment variable can be set to either a comma-separated
list of host names that should not use a proxy, or single asterisk `*`
to match all hosts.

- Port numbers can be included by prefixing the port with a colon `:`
- IP addresses can be used, but must be enclosed in square brackets
`[...]` if they include a port number
- IP address ranges can specified in CIDR notation, separating the IP
address and prefix-length with a forward slash `/`

I've put this on master due to the breaking changes, but perhaps it would be better on 2.0? The diff is a mess because I have rewritten the class, so I could redo it as a new class (but this would add another breaking change).

**Removing host resolution**
This seems to me like additional overhead. Golang (which powers Docker) considered doing this but came to the conclusion that DNS resolution wasn't being done by anyone else back in 2012: https://github.com/golang/go/issues/2158#issuecomment-66058160

**Using curl as a benchmark**
There is no _official_ NO_PROXY format. curl apparently based their version on Mosaic and Lynx browser behaviour (https://www.w3.org/Daemon/User/Proxies/ProxyClients.html) and it seems that this has subsequently provided the basis for languages like

- Python (https://bugs.python.org/issue29142) and 
- Golang (https://github.com/golang/go/commit/82ae6434b3a95beb64f708078af5a525088d6ccc).

### DIFF
--- a/src/Composer/Util/NoProxyPattern.php
+++ b/src/Composer/Util/NoProxyPattern.php
@@ -12,34 +12,76 @@
 
 namespace Composer\Util;
 
+use stdClass;
+
 /**
- * Tests URLs against no_proxy patterns.
+ * Tests URLs against NO_PROXY patterns
  */
 class NoProxyPattern
 {
     /**
      * @var string[]
      */
+    protected $hostNames = array();
+
+    /**
+     * @var object[]
+     */
     protected $rules = array();
 
     /**
-     * @param string $pattern no_proxy pattern
+     * @var bool
+     */
+    protected $noproxy;
+
+    /**
+     * @param string $pattern NO_PROXY pattern
      */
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks for doing the research and porting the curl stuff over. I think it's fine to target 1.10 and see how it goes. The BC breaks sound reasonable.

PR #8459, opened 2019-11-30
[help wanted] travis: php 7.4

adaamz:


adaamz:
hm, that phpunit php requirement is strange...

Does anybody know how to solve that unbound `phpunit` dependency on `php >= 5.3.3`?

I have only idea with some hidden magic like run command composer require phpunit/phpunit with some specific version for php ^7.4 env.

adaamz:
closed in favor of #8487

PR #8460, opened 2019-12-02
Added new Alias For Clear cache + Documentation

TheGarious:
Same as command update with up and install with i. 

I think it's a good idea to use cc for clear-cache, like npm command.


TheGarious:
Added a new alias into documentation

PR #8469, opened 2019-12-06
Update 01-basic-usage.md - add dump-autoload

abdounikarim:
Hello,

I add the `php composer.phar dump-autoload` command explicity because i didn't find it on the doc like this.

Cheers 😉


Seldaek:
Thanks

PR #8471, opened 2019-12-08
updated phpstan to 0.12

adaamz:


src/Composer/Installer/InstallationManager.php
```diff
@@ -245,7 +245,7 @@ public function execute(RepositoryInterface $repo, array $operations, $devMode =
                 if (defined($event) && $runScripts && $dispatcher) {
                     $dispatcher->dispatchPackageEvent(constant($event), $devMode, $repo, $operations, $operation);
                 }
-            }, function ($e) use ($jobType, $installer, $package, $initialPackage, $loop, $io) {
+            }, function ($e) use ($jobType, $installer, $package, $initialPackage, $loop) {
                 $this->io->writeError('    <error>' . ucfirst($jobType) .' of '.$package->getPrettyName().' failed</error>');
```
Seldaek: Actually instead of removing the var use, this line should read `$io->` instead of `$this->io` as the latter won't work on php 5.3.

src/Composer/Installer.php
```diff
@@ -166,7 +166,7 @@ class Installer
      * @param EventDispatcher      $eventDispatcher
      * @param AutoloadGenerator    $autoloadGenerator
      */
-    public function __construct(IOInterface $io, Config $config, RootPackageInterface $package, DownloadManager $downloadManager, RepositoryManager $repositoryManager, Locker $locker, InstallationManager $installationManager, EventDispatcher $eventDispatcher, AutoloadGenerator $autoloadGenerator)
+    final public function __construct(IOInterface $io, Config $config, RootPackageInterface $package, DownloadManager $downloadManager, RepositoryManager $repositoryManager, Locker $locker, InstallationManager $installationManager, EventDispatcher $eventDispatcher, AutoloadGenerator $autoloadGenerator)
```
Seldaek: Why the `final` here and in Factory? This seems quite unrelated to the phpstan upgrade..

adaamz: There is new rule that if you use `new static`, the constructor must be marked as final (or class itself).

src/Composer/Installer/InstallationManager.php
adaamz: So we found bug!

src/Composer/Installer.php
localheinz: Rather than marking the method as `final`, what do you think about ignoring the error for now?

phpstan/config.level1.neon
```diff
@@ -0,0 +1,9 @@
+parameters:
```
localheinz: This file is currently not used.

Can we remove it?

src/Composer/Factory.php
```diff
@@ -48,6 +48,10 @@
  */
 class Factory
 {
+    final public function __construct()
```
localheinz: Why?

src/Composer/Installer.php
Seldaek: Yeah to be honest I don't feel like this rule makes sense.. I get the point that an extending class might override the constructor *and* break the signature, but it also might do it and not break things, so this feels a little overly strict to me, especially for level1 stuff..

phpstan/config.level1.neon
adaamz: i just moved ignore errors from used config there, so we can in future use it easily

src/Composer/Installer.php
adaamz: tbh it is level 0 and i dont like this rule too, should be in some strict rules or higher level

phpstan/config.level1.neon
localheinz: What do you think about using `phpstan`'s baseline feature instead?

For reference, see https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff.

adaamz: It is good if you want to write better/type-safe code. If you don't it will only bother you.

adaamz:
I added custom rule which checks for using $this inside closures, pls check it out.

phpstan/config.level1.neon
Seldaek: It's an interesting feature IMO if it allows  raising the level faster, but that's something I'd rather investigate later as we get closer to 2.0 release to avoid too many conflicts from 1.x merges. For now let's focus on getting this PR done as is I'd say.

.travis.yml
```diff
@@ -64,8 +64,9 @@ script:
   - ls -d tests/Composer/Test/* | grep -v TestCase.php | parallel --gnu --keep-order 'echo "Running {} tests"; ./vendor/bin/phpunit -c tests/complete.phpunit.xml --colors=always {} || (echo -e "\e[41mFAILED\e[0m {}" && exit 1);'
   # Run PHPStan
   - if [[ $PHPSTAN == "1" ]]; then
-      bin/composer require --dev phpstan/phpstan-shim:^0.11 --ignore-platform-reqs &&
-      vendor/bin/phpstan.phar analyse src tests --configuration=phpstan/config.neon --autoload-file=phpstan/autoload.php;
+      composer config platform.php ${TRAVIS_PHP_VERSION} &&
```
ondrejmirtes: What is this stuff for?

phpstan/config.neon
```diff
@@ -1,22 +1,15 @@
...
+        - autoload.php
     level: 0
+    paths:
+        - ../src
+        - ../tests
     excludes_analyse:
-        - 'tests/Composer/Test/Fixtures'
-        - 'tests/Composer/Test/Autoload/Fixtures'
-        - 'tests/Composer/Test/Plugin/Fixtures'
+       - '*/tests/Composer/Test/Fixtures/*'
```
ondrejmirtes: Instead of `*/tests/...`, we should use correct relative path . `../tests/Composer/Test/Fixtures/*`.

phpstan/config.level1.neon
ondrejmirtes: I'd also remove this unused file. It will most likely get obsolete before it gets used. And we an easily generate it with the basline feature.

composer.json
```diff
@@ -66,7 +66,9 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Composer\\Test\\": "tests/Composer/Test"
+            "Composer\\Test\\": "tests/Composer/Test",
+            "Composer\\PHPStanRules\\": "phpstan/rules/src",
```
ondrejmirtes: The `phpstan` directory should be added to `.gitattributes` as `export-ignore`.

phpstan/Rules/src/AnonymousFunctionWithThisRule.php
```diff
@@ -0,0 +1,45 @@
...
+        if (!\is_string($node->name) || $node->name !== 'this') {
+            return [];
+        }
+
+        if ($scope->isInClosureBind()) {
+            return [];
+        }
+
+        if (!$scope->isInClass()) {
+            return [\PHPStan\Rules\RuleErrorBuilder::message('Using $this outside a class.')->build()];
```
ondrejmirtes: This is already reported by PHPStan itself, see: https://phpstan.org/r/00b1a3d5-a5c6-4a2a-a587-f52eee336adb

It can be removed from here.

phpstan/Rules/tests/AnonymousFunctionWithThisRuleTest.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    protected function getRule(): \PHPStan\Rules\Rule
+    {
+        return new AnonymousFunctionWithThisRule();
+    }
+
+    public function testWithThis(): void
+    {
+        $this->analyse([__DIR__ . '/method-with-this.php'], [
+            ['rip', 13],
```
ondrejmirtes: This isn't the right error message to assert.

.travis.yml
adaamz: Solved with existing env deps high, it sets composer platform php version in before script.

.gitattributes
```diff
@@ -15,3 +15,4 @@
 .travis.yml export-ignore
 appveyor.yml export-ignore
 phpunit.xml.dist export-ignore
+phpstan export-ignore
```
localheinz: Is this a file? A directory?

In case it is the latter, what do you think about adding a trailing slash to indicate it is?

adaamz:
@Seldaek yeah i know, im in the end with that CI fail, trying everything i found on google.

composer.json
```diff
@@ -66,8 +66,13 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Composer\\Test\\": "tests/Composer/Test"
-        }
+            "Composer\\Test\\": "tests/Composer/Test",
+            "Composer\\PHPStanRules\\": "phpstan/rules/src",
+            "Composer\\PHPStanRulesTests\\": "phpstan/rules/tests"
```
Seldaek: This `phpstan/rules/tests` here and the src one above does not match the case of the files in git which are in a capitalized Rules folder. That might be the cause of the travis failure.

adaamz: ah, many thanks!
i thought about case mismatch (im on windows), but looks like i didnt check it properly

.gitattributes
adaamz: added slashes

Seldaek:
Thanks!

PR #8475, opened 2019-12-10
Enhancement: Keep packages sorted in composer.json

localheinz:
This PR

* [x] keeps packages sorted in`composer.json`

💁‍♂ For reference, I ran

```
$ bin/composer config sort-packages true
```

followed by

```
$ bin/composer require psr/log:^1.0
```

to force sorting of packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,13 +29,13 @@
         "composer/xdebug-handler": "^1.1",
         "justinrainbow/json-schema": "^3.0 || ^4.0 || ^5.0",
         "psr/log": "^1.0",
+        "react/promise": "^1.2 || ^2.7",
         "seld/jsonlint": "^1.4",
         "seld/phar-utils": "^1.0",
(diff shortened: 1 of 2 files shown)


stof:
@Seldaek there are changes in the lock file here related to author and support metadata. Could these be caused by a different behavior when using Composer 1 vs Composer 2-dev to create it ?

localheinz:
@stof 

I think that might be the case, yes. I have wondered myself because the changes were rather massive. I can remove b5228f24a8db72715d26df8b2001892f29ff260d, though.

Seldaek:
Sorry but I am personally not a fan of this package sorting thing..

And yes author/support are present in v2 metadata, while support is missing from v1 metadata, and author was initially omitted from v2 but is now included but not all packages have had a metadata dump yet since that change I assume.

localheinz:
No problem, @Seldaek!

PR #8487, opened 2019-12-19
[WIP] Test on PHP 7.4 Stable

GrahamCampbell:
Test on stable PHP 7.4, and use latest distributions available on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: php
 
-dist: trusty
-
 git:
   depth: 5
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Looks like we'll have to upgrade PHPUnit to test on PHP 7.4 properly.

johnstevenson:
@GrahamCampbell There would be quite a lot of work to make it compatibile with lower PHP/PHPUnit versions.

I did this for https://github.com/composer/xdebug-handler/commit/6624d7ee1ce21572af8558e7420f8ae954a902fb which is a tiny codebase and had to handle:

- replacing setupBeforeClass/ tearDownAfterClass with different annotated methods.
- replacing deprecated asserts (like assertContains).
- changing `@expectedException` annotations to use either `setExpectedException` or `expectException` depending on the PHPUnit version.

Perhaps symfony/phpunit-bridge would be a better solution for larger codebases?

adaamz:
@johnstevenson That is not problem. Problem is that composer is installing phpunit 4.8 for php 7.4, because phpunit has no upper bound restriction.

johnstevenson:
@adaamz Yes, but to get xdebug-handler to work with lastest php (inc nightly) and the corresponding versions of phpunit, I had to make those changes (phpunit/phpunit": "^4.8.35 || ^5.7 || 6.5 - 8).

adaamz:
@johnstevenson That is already there, or not?

GrahamCampbell:
If someone wants to take over here, and finish this, that would be good. I don't have the free time needed to finish this at the moment.

Seldaek:
Oh sorry forgot about this PR, but this was done on the 2.0 branch already so closing https://github.com/composer/composer/blob/2.0/.travis.yml

PR #8496, opened 2019-12-23
Update 00-intro.md

dotku:
add homebrew as install alternative


Seldaek:
We do not support installing via homebrew as it led to too much confusion in the past.

PR #8503, opened 2019-12-29
Fix nits on typos

TysonAndre:
https://english.stackexchange.com/questions/117552/why-does-information-not-have-a-plural-form


Seldaek:
Thanks

PR #8504, opened 2019-12-30
Fixed highlighting required. d was excluded.

dennybrandes:


Seldaek:
This is by design as the field is called require

PR #8508, opened 2020-01-03
validate: suggest composer update --lock

glensc:
Suggest more safer `composer update --lock` command when lock file is not up to date.

### DIFF
--- a/src/Composer/Command/ValidateCommand.php
+++ b/src/Composer/Command/ValidateCommand.php
@@ -93,7 +93,7 @@ EOT
         $composer = Factory::create($io, $file, $input->hasParameterOption('--no-plugins'));
         $locker = $composer->getLocker();
         if ($locker->isLocked() && !$locker->isFresh()) {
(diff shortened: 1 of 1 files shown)


glensc:
...or suggest both commands?

svenluijten:
Dupes #8332

glensc:
I find that people who don't know how composer work, want to bump only one dependency, edit `composer.json`, then found that project uses lock file, notice `composer.lock` is not updated, end up bumping all dependencies because composer tells them to run `composer update`, and then they don't really look `composer.lock` diff, commit whole lock file, and in the review process, it's typically ignored, because GitHub/GitLab by default hides such large diffs and reivewer sees only bumped dep on `composer.json`.

Seldaek:
If it's an education problem then I hope my commit helps educating people in the right direction. `update --lock ` is IMO not the solution to this.

glensc:
Indeed, if the lock file is out of date because someone used to edit package.json to bump a dependency, proper would for them to run `composer update <package>`.

PR #8510, opened 2020-01-04
VcsRepositories: mark archived repositories as abandoned

glaubinix:
This implements: https://github.com/composer/packagist/issues/1055

Note:
* Bitbucket does not have an archive functionality
* GitLab only provides the properties if the project data is accessed with authentication


Seldaek:
Thanks

PR #8511, opened 2020-01-05
Rename .php_cs to .php_cs.dist

voyula:


glensc:
Good pull requests explain the motivation behind the change.

Seldaek:
Don't really see the point, we don't want people modifying this config, or even running it.. it's a maintainer tool as far as I am concerned.

Issue #8512, opened 2020-01-06
Allow $schema in composer.json (and composer.lock)


glen-84:
**Edit:** Correction, it is _not_ supported by PhpStorm.

---

I thought that this was [VS Code-specific](https://code.visualstudio.com/docs/languages/json#_mapping-in-the-json), but it seems to work in PhpStorm as well.

![image](https://user-images.githubusercontent.com/261509/71811229-d2dd0700-307c-11ea-85c2-cab83a8c2098.png)

![image](https://user-images.githubusercontent.com/261509/71811246-da041500-307c-11ea-8dbb-86dd7e46eebd.png)

localheinz:
Sounds good!

Was just thinking about this the other day as well!

mfn:
Nice!

You can already choose at the bottom, but of course it's auto-included, that would be even better:
![image](https://user-images.githubusercontent.com/87493/71814008-804f1b00-307b-11ea-9476-6d12e6b80dcd.png)

paul-gene:
In PHPStorm 2019.2, it shows a warning that `$schema` isn't allowed.

glen-84:
@paul-gene That's expected, and what this issue is about. See #8513, which will fix this.

paul-gene:
@glen-84 but in your screenshot, it's not being underlined as erroneous?

glen-84:
@paul-gene It's highlighted as a warning in my screenshot.

stof:
I'm quite sure this is VS Code specific (as they document themselves, with a gotcha that it might make validation fail). And https://github.com/json-schema-org/json-schema-spec/issues/828 makes it clear that this is *not* what is expected by the JSON-Schema spec.

AFAIK, PHPStorm does not read this `$schema`. But it probably already has the mapping for composer.json built-in (either it is built-in, or it is part of the composer.json plugin).

localheinz:
@glen-84 @mfn 

If you are interested in support for specifying the underlying JSON schema in a JSON file, please chime in at https://github.com/json-schema-org/json-schema-spec/issues/828.

stof:
btw, if you install https://marketplace.visualstudio.com/items?itemName=ikappas.composer VS Code will know automatically that composer.json files are using the composer json schema (this extension registers this mapping in the config)

Relequestual:
> If you are interested in support for specifying the underlying JSON schema in a JSON file, please chime in at json-schema-org/json-schema-spec#828.

[...]

PR #8513, opened 2020-01-06
Enhancement: Add $schema property to composer-schema.json

localheinz:
This PR

* [x] adds `$schema` property to `composer-schema.json`

Related to #8512.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -5,6 +5,10 @@
     "additionalProperties": false,
     "required": [ "name", "description" ],
     "properties": {
+        "$schema": {
+            "type": "string",
+            "description": "URL to the underlying JSON schema, can be specified to assist with auto-completion."
+        },
         "name": {
             "type": "string",
             "description": "Package name, including 'vendor-name/' prefix."


res/composer-schema.json
```diff
@@ -5,6 +5,10 @@
     "additionalProperties": false,
     "required": [ "name", "description" ],
     "properties": {
+        "$schema": {
+            "type": "string",
+            "description": "URL to the underlying JSON schema, can be specified to assist with auto-completion."
```
localheinz: Should the description include a note that the specified schema URL will **not** be used for validation (because it isn't)?

stof:
This looks weird to me. According to https://json-schema.org/understanding-json-schema/reference/schema.html, `$schema` is meant to identify that a JSON fragment is a schema (and reference the version of json-schema being used).
A composer.json file is not a json schema file.

localheinz:
@stof 

Yes, perhaps this is not the appropriate way.

Also see https://github.com/json-schema-org/json-schema-spec/issues/828.

stof:
the issue you link tells me that this feature request is indeed wrong.

localheinz:
@stof 

> the issue you link tells me that this feature request is indeed wrong.

No, the feature request is not wrong, there currently isn't a standard way that supports the request.

Issue #6301, opened 2017-03-26
Set preferred-install for specific packages from command line


jonathanjfshaw:
It doesn't seem possible to set  a global config preferred-install for specific packages from the command line.

I'm trying to setup containers for contributing to Drupal, using a third-party composer.json from drupal-composer/drupal-project. I'd like all drupal packages to be downloaded prefer-source, but all upstream dependencies to be prefer-dist.

However, from the cli this sort of thing does not seem possible:
`composer config --global preferred-install '{"drupal/*":"source"}'`

alcohol:
The `composer config` command is somewhat of a glued together bundle of spaghetti. It could very well be the case that this particular config attribute has not been added yet to the list of cases that the command can handle.

From what I can tell by just quickly looking at the source code, it only accepts values on a global level, not on a vendor/package level: https://github.com/composer/composer/blob/master/src/Composer/Command/ConfigCommand.php#L300-L307

Pull requests are welcome.

smoqadam:
Hi

I'm a new contributer and want to help. I want to grab this one but have no idea what to do? 

If this issue is not good for start please introduce me something else

lf-jeremy:
It is worth noting that in the meantime, this can be accomplished via config as an alternative option that might be acceptable in some use cases:

https://getcomposer.org/doc/06-config.md#preferred-install

That support was added by https://github.com/composer/composer/pull/3790 should anyone want to see related work for implementing this for cli.

It seems a cli option will involve the nuances of how explicit cli options may interact with config-based options that are inherited.

alcohol:
@smoqadam it depends on your familiarity with php I suppose.

Basically, the config command can be used to modify the config section of the `composer.json` file from the command line interface.

This is the command in question: [ConfigCommand](https://github.com/composer/composer/blob/master/src/Composer/Command/ConfigCommand.php)

If you read through the `execute()` method you should be able to get a gist of how it handles the input it receives. The `preferred-install` config section is currently handled here: https://github.com/composer/composer/blob/3c173702b5e45ba8180f991744ef947e863ece28/src/Composer/Command/ConfigCommand.php#L305-L312

Perhaps it would suffice to simply treat it as a multi-value setting, similar to these: https://github.com/composer/composer/blob/3c173702b5e45ba8180f991744ef947e863ece28/src/Composer/Command/ConfigCommand.php#L412-L455

adrianosferreira:
[...]

PR #8517, opened 2020-01-07
Implemented way to use preferred-install for defining granular preferences through CLI

adrianosferreira:
Reported in https://github.com/composer/composer/issues/6301

Currently, preferred-install accepts the hash of patterns as the value in the composer.json. I've followed the same approach as used in extra and platform for letting the user define install preferences through CLI in the format:

```
composer config preferred-install my-organization/stable-package.dist
composer config preferred-install my-other-organization/stable-package.source
composer config preferred-install my-other-organization/stable-package.auto
```

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -306,14 +306,6 @@ EOT
             'process-timeout' => array('is_numeric', 'intval'),
             'use-include-path' => array($booleanValidator, $booleanNormalizer),
             'use-github-api' => array($booleanValidator, $booleanNormalizer),
-            'preferred-install' => array(
-                function ($val) {
-                    return in_array($val, array('auto', 'source', 'dist'), true);
-                },
-                function ($val) {
-                    return $val;
-                },
-            ),
             'store-auths' => array(
                 function ($val) {
                     return in_array($val, array('true', 'false', 'prompt'), true);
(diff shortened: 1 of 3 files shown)


Seldaek:
Thanks, refactored quite a bit while merging as a lot of it was actually not necessary, see the merge commit https://github.com/composer/composer/commit/ee3fd0f74544406dbcdae4ea72a1baef3b62c049 + https://github.com/composer/composer/commit/f5dab18b7d392ce17c17aa0c29f0a2559aca3390 for details.

adrianosferreira:
@Seldaek hold on. You can probably remove the block https://github.com/composer/composer/pull/8517/files#diff-2a42cc8f017c002beea510989bbf163eR483 too as you removed the definition of `$uniqueOrDotNestedArray` and confirmed we don't need this.

Seldaek:
That's done in the second commit

adrianosferreira:
@Seldaek never mind, it looks like you did it already :)
Do you have sure that your current version will accept the format `composer config preferred-install auto`? Because I've added this basically for that:

```
try {
		    	$this->handleSingleValue($settingKey, $uniqueOrDotNestedArray[$settingKey], $values, 'addConfigSetting');
		    }
```

I try to handle as a single value, respecting the valid ones and then try to handle as package.

Seldaek:
Yes it does handle it as it did previously since I undid the deletion in https://github.com/composer/composer/pull/8517/files#diff-2a42cc8f017c002beea510989bbf163eL309-L316

adrianosferreira:
@Seldaek makes sense!

Issue #5460, opened 2016-06-22
CLI suggest cmd


amansilla:
I think would be nice to have a command e.g. `composer suggest` for adding a new package to the `suggest` section. So if the user type `composer suggest monolog/monolog "Allows more advanced logging of the application flow"` the following content would be added to the `composer.json`.

```
{
    "suggest": {
        "monolog/monolog": "Allows more advanced logging of the application flow"
    }
}
```

What do you think about this?

curry684:
If this is implemented it should be part of existing `config` command.

amansilla:
You are right, using the `config` command there is no need for a new `suggest` command. Just need to update `config` to be able to set the `suggest` value of the `composer.json`.

Taluu:
Not sure this has anything to do with configuration. A suggested package is * not * a suggestion, it is, well... a suggestion.

curry684:
You should be aware that typing `composer suggest` today already produces a meaningful result.

sbuzonas:
Wouldn't it just be as simple as adding `suggest` to the properties array? https://github.com/composer/composer/blob/master/src/Composer/Command/ConfigCommand.php#L234

adrianosferreira:
Maybe what I've added here could be improved for this: https://github.com/composer/composer/pull/8517/files#diff-2a42cc8f017c002beea510989bbf163eR506

Currently, it allows you to add a key pair inside of config > your_setting_key `composer config preferred-install my-organization/stable-package.dist`. It can be just a matter of not enforcing the `config.` like I did there.

adrianosferreira:
Proposed a PR in https://github.com/composer/composer/pull/8520

Seldaek:
Fixed by #8520 usable as `composer config suggest.package/foo reason for suggesting`

PR #8520, opened 2020-01-10
Implemented way to add packages into suggest through CLI

adrianosferreira:
Issue https://github.com/composer/composer/issues/5460


Seldaek:
Thanks, added support for --unset and for suggests with multiple words in https://github.com/composer/composer/commit/e8426d2c013562c187f2c2e1a26ac6fdf2d7ee45#diff-2a42cc8f017c002beea510989bbf163eR620 but otherwise was all good

PR #8524, opened 2020-01-11
WIP ProcessExecutor - do not format output from executed processes (failing test)

ondrejmirtes:


ondrejmirtes:
Hi, I've encountered the problem that Composer does not leave the output from `composer run` scripts as-is, but tries to format it according to Symfony Console rules, so it changes characters around `<` and `>`, like `<error>foo</error>`. This is a problem in practice because the output is usually some well-defined format like JSON with its own formatted rules.

I found this problem thanks to https://github.com/phpstan/phpstan/issues/2843.

This PR is just a failing test because I'm not sure where the fix should be applied and what are the implications.

ondrejmirtes:
A tip - when I found out about this in PHPStan itself, I renamed my output abstractions to force the user to decide if they want to format the output or not: https://github.com/phpstan/phpstan-src/blob/master/src/Command/Output.php#L8-L12

PR #8528, opened 2020-01-15
Avoid loading packages found in a higher prio repo entirely from lower prio repos

### LINKED ISSUE #5076: When using private repository, Composer silently prefers the Packagist one if the private is not installable
Let's say there's a package `foo/foo`. It's public, but there's a private version too.

I add a dependency to the dependent (root) project as `"foo/foo": "dev-master"`.

I add the private repository to composer.json's `repositories` key and that version's master branch has different requirements for dependencies than the public one.

If the root project has other requirements that are satisfied only by the public version on Packagist, Composer silently installs the public version even if there's a repositories section with the private version repository.

I think that adding the repository to `repositories` section clearly states my intent to use that package version instead of the publicly available one and that Composer should throw an error in this case.

Seldaek:
@naderman Question: the way I understand things, this makes the concept of priorities in PoolBuilder/Pool/DefaultPolicy entirely irrelevant. If that's correct I can also delete all that. 

Fixes #5076

### DIFF
--- a/src/Composer/DependencyResolver/DefaultPolicy.php
+++ b/src/Composer/DependencyResolver/DefaultPolicy.php
@@ -69,7 +69,6 @@ class DefaultPolicy implements PolicyInterface
         }
 
         foreach ($packages as &$sortedLiterals) {
-            $sortedLiterals = $this->pruneToHighestPriority($pool, $sortedLiterals);
             $sortedLiterals = $this->pruneToBestVersion($pool, $sortedLiterals);
             $sortedLiterals = $this->pruneRemoteAliases($pool, $sortedLiterals);
         }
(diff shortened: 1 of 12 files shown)


naderman:
@Seldaek I wonder if we should add some more options to the Composer repository definition to give people a way to work around problems. Specifically I think we could add properties "exclude" and "only" using an array of patterns:

```json
[{
    "type": "composer",
    "url": "https://foobar.com",
    "exclude": [
        "foo/bar",
        "baz/*"
    ]
},{
    "type": "composer",
    "url": "https://acme.com",
    "only": [
        "acme/*",
        "acme-forks/xyz"
    ]
}]
```
This would make sure that if you currently rely on Composer picking a version from a lower priority repository because none of the versions in the higher prio match at all, you have a way to manually replicate this behavior. Additionally it makes it a bit more predictable what third party repositories can do, since you can restrict them to particular vendor names only.

Issue #2764, opened 2014-02-26
Should composer play nice with make?


mtibben:
I'm working on a codebase using make. The makefile looks like

``` makefile
vendor: composer.lock
    composer install

test: vendor
    ./bin/phpunit
```

This works well, because make knows that in order to run phpunit, it should build the vendor directory by running `composer install`. Once it has been run once, make knows that the timestamp of vendor is greater than composer.lock, so doesn't need to run it again if I `make test`. And when composer.lock does get changed, make knows to rerun `composer install`.

The problem occurs when composer.lock changes with a trivial change (e.g. the description might be changed) which means that composer doesn't need to do anything when `composer install` is run.

Now make thinks that vendor is out of date, and will run `composer install` each time I need to `make test`.

I got around this problem by adding `touch vendor` to the vendor target. This works around the problem, but it seems to me that this is something composer should be doing when `composer install` is run.

PR #8529, opened 2020-01-15
Change only modified files for docker caching

vitalyzhakov:
Hello!
We use CI process, which includes a  `composer install` command.
Every `composer install` execution modifies the dates for the vendor folder and some files.

It blocks docker layer caching system and increase CI JOB time.

This PR allows composer install and docker cache system to run along.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -285,24 +285,24 @@ EOF;
             }
         }
 
-        file_put_contents($targetDir.'/autoload_namespaces.php', $namespacesFile);
-        file_put_contents($targetDir.'/autoload_psr4.php', $psr4File);
-        file_put_contents($targetDir.'/autoload_classmap.php', $classmapFile);
+        $this->filePutContentsIfModified($targetDir.'/autoload_namespaces.php', $namespacesFile);
+        $this->filePutContentsIfModified($targetDir.'/autoload_psr4.php', $psr4File);
+        $this->filePutContentsIfModified($targetDir.'/autoload_classmap.php', $classmapFile);
         $includePathFilePath = $targetDir.'/include_paths.php';
         if ($includePathFileContents = $this->getIncludePathsFile($packageMap, $filesystem, $basePath, $vendorPath, $vendorPathCode52, $appBaseDirCode)) {
-            file_put_contents($includePathFilePath, $includePathFileContents);
+            $this->filePutContentsIfModified($includePathFilePath, $includePathFileContents);
         } elseif (file_exists($includePathFilePath)) {
             unlink($includePathFilePath);
         }
         $includeFilesFilePath = $targetDir.'/autoload_files.php';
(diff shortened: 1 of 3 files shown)


src/Composer/Installer.php
```diff
@@ -311,13 +311,6 @@ public function run()
             foreach ($localRepo->getPackages() as $package) {
                 $this->installationManager->ensureBinariesPresence($package);
             }
-
-            $vendorDir = $this->config->get('vendor-dir');
-            if (is_dir($vendorDir)) {
-                // suppress errors as this fails sometimes on OSX for no apparent reason
-                // see https://github.com/composer/composer/issues/4070#issuecomment-129792748
-                @touch($vendorDir);
-            }
```
Seldaek: This was done to fix https://github.com/composer/composer/issues/2764 - so removing it will break that use case.

vitalyzhakov: If I correctly understand #2764, @mtibben changed composer.lock file via `composer update` command.
Ok, I fix this situation using `touch` only if `composer update` operation.

Seldaek: I don't know how you understood this, the word `update` isn't anywhere on that issue.. The point was that after someone updated, if you fetch changes, you run install, and the vendor dir gets modified. What we could do is only update the timestamp if anything changed in the directory / if any operations were executed.

vitalyzhakov: bq. after someone updated, if you fetch changes, you run install
After some requirements updated, we must use `composer update` command, is it right?

Seldaek: If you changed the requirements you have to run update, but if you fetch an updated lock file from somewhere which updated what was installed, you have to run install to sync up your vendor dir with that state.

vitalyzhakov: Thank your for fast answer!
How about use `touch` only after operations?

`Nothing to install or update` message means, that nothing changed, we can inverse it.

Seldaek: Yup that looks better.

vitalyzhakov:
@Seldaek , hello!
Do you have any comments on this PR?

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -1005,11 +1014,42 @@ protected function sortPackageMap(array $packageMap)
      */
     protected function safeCopy($source, $target)
     {
-        $source = fopen($source, 'r');
-        $target = fopen($target, 'w+');
+        if (!file_exists($target) OR !file_exists($source) OR !$this->filesAreEqual($source, $target)) {
```
Seldaek: This should be || not OR.

vitalyzhakov: Ok, fixed

vitalyzhakov:
@Seldaek , hello!
Do you have any comments on this PR?

Or can your merge it to master branch?

Seldaek:
I merge when I have time thank you please refrain from asking every day it just wastes more of my time.

Seldaek:
Thanks!

PR #8531, opened 2020-01-15
Pool builder cleanups

Seldaek:
Avoids loading fixed packages from remote repos entirely, which is good, but it breaks a test as the problem reporting doesn't work anymore since the package is missing.

Fixes https://github.com/composer/composer/issues/3672

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -52,6 +52,7 @@ class PoolBuilder
         // TODO do we really want the request here? kind of want a root requirements thingy instead
         $loadNames = array();
         foreach ($request->getFixedPackages() as $package) {
+            $this->nameConstraints[$package->getName()] = null;
             $this->loadedNames[$package->getName()] = true;
             unset($loadNames[$package->getName()]);
             $loadNames += $this->loadPackage($request, $package);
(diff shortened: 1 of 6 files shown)


naderman:
Looks great except for the test failing. I guess the problem is just that 1.0.1 doesn't get loaded at all anymore and so the error message disappears? I guess correct the test for now and make sure we document this for improvements to the error handler?

I'm just a bit confused where the 1.0.0 gets filtered out currently in a way that means it doesn't get shown as part of the problem exception? The root requires should not be applied to it?

stof:
Well, problem reporting should not rely on the pool to know which packages are available in the repositories, as any pool optimization would affect the reporting (which is what happens here, but it is already an issue in 1.x when we filter the pool based on root requirements).
Problem reporting wanting to know which (excluded) versions are available should look at repositories directly.

naderman:
@stof yes there is already a ticket for this, problem reporting will be switched to the repo set anyway. The question here though is why is this particular version which is supposedly fixed from the lock file in a partial update even filtered out from the pool? I don't think it should be.

stof:
Well, this PR is precisely about introducing this filtering for partial updates, as suggested in my issue (but naming the PR `cleanups` when it actually implements a new optimization is quite misleading)

naderman:
@stof I don't think you understood me at all. There is a problem with a particular test case where this PR results in a locked package getting filtered as a non-matching root requirement and disappears from an error message. However it is entirely unclear to me where/why this gets filtered.

The PR is a cleanup because the logic for handling fixed packages in the pool builder was already there. It was simply missing some code to skip loading the packages afterwards, the main concept was already added before.

PR #8533, opened 2020-01-16
Remove callback and pass stabilities all the way instead

Seldaek:
This allows optimizing the loading of ~dev files, and cleans up a few things

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -380,12 +380,10 @@ class Installer
         }
 
         // if the updateWhitelist is enabled, packages not in it are also fixed
-        // to the version specified in the lock, except if their stability is not
-        // acceptable anymore, to make sure that they get updated/downgraded to
-        // a working version
+        // to the version specified in the lock
         if ($this->updateWhitelist && $lockedRepository) {
             foreach ($lockedRepository->getPackages() as $lockedPackage) {
-                if (!$this->isUpdateable($lockedPackage) && $repositorySet->isPackageAcceptable($lockedPackage->getNames(), $lockedPackage->getStability())) {
+                if (!$this->isUpdateable($lockedPackage)) {
                     // TODO add reason for fix?
                     $request->fixPackage($lockedPackage);
                 }
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -364,19 +364,6 @@ class Installer
 
         $request = $this->createRequest($this->fixedRootPackage, $platformRepo, $lockedRepository);
 
-        if ($lockedRepository) {
-            // TODO do we really always need this? Maybe only to skip fix() in updateWhitelist case cause these packages get removed on full update automatically?
-            foreach ($lockedRepository->getPackages() as $lockedPackage) {
-                if (!$repositorySet->isPackageAcceptable($lockedPackage->getNames(), $lockedPackage->getStability())) {
-                    $constraint = new Constraint('=', $lockedPackage->getVersion());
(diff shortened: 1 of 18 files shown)


src/Composer/Installer.php
```diff
@@ -362,19 +362,6 @@ protected function doUpdate(RepositoryInterface $localRepo, $doInstall)
...
-            foreach ($lockedRepository->getPackages() as $lockedPackage) {
-                if (!$repositorySet->isPackageAcceptable($lockedPackage->getNames(), $lockedPackage->getStability())) {
-                    $constraint = new Constraint('=', $lockedPackage->getVersion());
-                    $constraint->setPrettyString('(stability not acceptable)');
-
-                    // if we can get rid of this remove() here, we can generally get rid of remove support in the request
-                    $request->remove($lockedPackage->getName(), $constraint);
-                }
-            }
-        }
```
Seldaek: This seemed entirely unnecessary or at least I could not find a case where it's needed..

src/Composer/Installer.php
```diff
@@ -391,10 +378,11 @@ protected function doUpdate(RepositoryInterface $localRepo, $doInstall)
...
         // if the updateWhitelist is enabled, packages not in it are also fixed
-        // to the version specified in the lock
-        if ($this->updateWhitelist) {
+        // to the version specified in the lock, except if their stability is not
+        // acceptable anymore, to make sure that they get updated/downgraded to
+        // a working version
+        if ($this->updateWhitelist && $lockedRepository) {
             foreach ($lockedRepository->getPackages() as $lockedPackage) {
-                // TODO should this really be checking acceptability here?
                 if (!$this->isUpdateable($lockedPackage) && $repositorySet->isPackageAcceptable($lockedPackage->getNames(), $lockedPackage->getStability())) {
```
Seldaek: Ideally the isPackageAcceptable would be removed here, so that non-updateable packages are fixed, and then we'd filter the fixed packages according to stability-acceptability in the pool builder, and these packages would then be missing/uninstallable if the stability rules changed. However I tried this and it didn't work it failed saying that a fixed package was not added to the pool.

So right now the way it works is that if you change the stability rules and then do a partial update, any locked package which doesn't match the stability rules anymore will be updated as if it was in the update whitelist. It's not terrible and most likely not that common a case, but just wanted to clarify it here for the record. If you see a way to fix the above that'd be better though I think.

src/Composer/Installer.php
naderman: As per the comment here, can you then get rid of the remove() function in the request altogether?

src/Composer/Installer.php
Seldaek: Totally agree. I had another look, if I change [this line](https://github.com/composer/composer/blob/2.0/src/Composer/DependencyResolver/RuleSetGenerator.php#L294) to `continue` instead, then I can do the change as I wanted to in PoolBuilder and skip loading the fixed packages entirely if they are not matching the stability requirements. Do you see any problems doing this change in RuleSetGenerator?

This would require merging #8531 first tho, and will require tweaking the Problem class to detect this case when a package is missing entirely, but it seems to work otherwise.

naderman: I would prefer if a package which did not meet stability requirements would simply not get fixed in the first place, and we keep fixed packages to mean they will always be part of the result set? I did have a bug before that made me add this check and I think it's easy to mess this up again when doing more optimizations in the pool builder, so I'd rather avoid fixing packages altogether if they are not meant to be used?

PR #8535, opened 2020-01-17
Optimize loading of deps from fixed packages

Seldaek:


src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -69,22 +69,39 @@ public function buildPool(array $repositories, Request $request)
...
-                    // also see the solver-problems.test test case
-                    $constraint = array_key_exists($job['packageName'], $loadNames) ? null : $job['constraint'];
-                    $loadNames[$job['packageName']] = $constraint;
-                    $this->nameConstraints[$job['packageName']] = $constraint ? new MultiConstraint(array($constraint), false) : null;
+
+                    $loadNames[$job['packageName']] = null;
+                    if ($job['constraint']) {
+                        if (!array_key_exists($job['packageName'], $this->nameConstraints)) {
+                            $this->nameConstraints[$job['packageName']] = new MultiConstraint(array($job['constraint']), false);
+                        } elseif ($this->nameConstraints[$job['packageName']]) {
```
naderman: You define all the nameConstraints as null in the fixed packages loop above, so this will never happen?

Besides it seems pointless to even define values for nameConstraints if we already picked a fixed package for this name and will never consider any other packages by that name? So really we should just skip any install request for a package that's already fixed here?

naderman: I guess this could happen if there are multiple install requests for the same package name, but I think that's something the request should probably reject? Or is there a scenario where we would allow requiring the same package in require and require-dev but with different version numbers?

Seldaek: Yeah this should never happen, I just copied the stuff from below in addPackage.. Which would make sense to have as a helper method anyway I guess.

But note that this code is skipped entirely for fixed packages, as the `continue 2` above triggers for those.

naderman:
Closing in favour of rebased version https://github.com/composer/composer/pull/8557

Issue #5522, opened 2016-07-17
When installing vendor binaries the path does not get updated


rohdef:
Running in a clean directory (that is there's no vendor folder or at least no vendor/bin) apart from tests:

With the following composer.json:

``` json
{
  "require-dev": {
    "phpunit/phpunit": "5.*.*"
  },
  "scripts": {
    "test": [
      "@composer install",
      "phpunit -c phpunit.xml"
    ]
  }
}
```

When I run this command:

```
composer test -vvv
```

I get this output:

```
php composer.phar test -vvv
Reading ./composer.json
Loading config file ./composer.json
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/var/www/html/bookstudenter): git branch --no-color --no-abbrev -v
Failed to initialize global composer: Composer could not find the config file: /home/rohdef/.composer/composer.json
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
Running 685ff8699ba378cc815849fb7daa8d0d2d25c609 (2016-07-12 16:19:15) with PHP 7.0.8-0ubuntu0.16.04.1 on Linux / 4.4.0-30-generic
> test: @pre-test
> pre-test: @composer install
Executing command (CWD): /usr/bin/php7.0  /var/www/html/bookstudenter/composer.phar install
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
  - Installing mikey179/vfsstream (v1.6.3)
    Loading from cache

  - Installing webmozart/assert (1.0.2)
    Loading from cache

  - Installing phpdocumentor/reflection-common (1.0)
    Loading from cache

  - Installing phpdocumentor/type-resolver (0.2)
    Loading from cache

  - Installing phpdocumentor/reflection-docblock (3.1.0)
    Loading from cache

  - Installing phpunit/php-token-stream (1.4.8)
    Loading from cache

  - Installing symfony/yaml (v3.1.2)
    Loading from cache

  - Installing sebastian/version (2.0.0)
    Loading from cache

  - Installing sebastian/resource-operations (1.0.0)
    Loading from cache

  - Installing sebastian/recursion-context (1.0.2)
    Loading from cache

  - Installing sebastian/object-enumerator (1.0.0)
    Loading from cache

  - Installing sebastian/global-state (1.1.1)
    Loading from cache

  - Installing sebastian/exporter (1.2.2)
    Loading from cache

  - Installing sebastian/environment (1.3.7)
[...]

PR #8538, opened 2020-01-20
Append the bin dir on each listener iteration

adrianosferreira:
Issue https://github.com/composer/composer/issues/5522

The "composer install" can create the vendor/dir folders and be used as a script item on composer.json. Having another script running after it that relies on vendor/bir binaries (such as phpunit) will cause it to not find the binary. This fix addresses the issue by trying to append the path on each script iteration.

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -145,27 +145,15 @@ class EventDispatcher
      */
     protected function doDispatch(Event $event)
     {
-        $pathStr = 'PATH';
-        if (!isset($_SERVER[$pathStr]) && isset($_SERVER['Path'])) {
-            $pathStr = 'Path';
-        }
-
-        // add the bin dir to the PATH to make local binaries of deps usable in scripts
-        $binDir = $this->composer->getConfig()->get('bin-dir');
-        if (is_dir($binDir)) {
-            $binDir = realpath($binDir);
-            if (isset($_SERVER[$pathStr]) && !preg_match('{(^|'.PATH_SEPARATOR.')'.preg_quote($binDir).'($|'.PATH_SEPARATOR.')}', $_SERVER[$pathStr])) {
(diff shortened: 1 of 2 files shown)


adrianosferreira:
Not sure why the test is not passing on AppVeyor though, it is seems to be something with the filesystem functions that I've used. Any ideas?

src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -145,27 +145,15 @@ public function dispatchInstallerEvent($eventName, $devMode, PolicyInterface $po
...
-        }
-
         $listeners = $this->getListeners($event);
 
         $this->pushEvent($event);
 
         $return = 0;
         foreach ($listeners as $callable) {
+
+            $this->append_bin_dir_on_path();
```
Seldaek: Please follow coding style of the files you edit, this should be `ensureBinDirIsInPath()` or something

src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -545,4 +533,22 @@ protected function popEvent()
     {
         return array_pop($this->eventStack);
     }
+
+    private function append_bin_dir_on_path()
+    {
+        $pathStr = 'PATH';
+        if (!isset($_SERVER[$pathStr]) && isset($_SERVER['Path'])) {
+            $pathStr = 'Path';
+        }
```
Seldaek: This is the reason AppVeyor fails I think. On windows the PATH env var is Path (sometimes anyway) so you need to check for this too in the test.

adrianosferreira:
Thanks @Seldaek, I've fixed and pushed, can you double-check? There are still some fails, unrelated to my changes AFAIK.

Just for curiosity, why did you choose AppVeyor over TravisCI? Any special reason?

Seldaek:
We use both, AppVeyor for windows builds only, but it's fairly unreliable.. 

Anyway thanks for the fix, looks good now I think.

PR #8545, opened 2020-01-21
Add support for "composer show --format=json <package>" #8544

### LINKED ISSUE #8544: Support "composer show <package> --format=json"
When I run `composer show <package> --format=json` I get the same output as `composer show <package>`, it would be nice to have the output as JSON to easily get package versions with for example jq.

**Use Case**
`composer global config repositories.repo.magento.com composer https://repo.magento.com`
`composer global show -a --format=json magento/magento2-base | jq '.versions | join(",")'`

Serializator:
Resolves #8544, luckily most of the front work was already done in #7416

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -238,18 +238,12 @@ EOT
 
                     return $exitCode;
                 }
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

PR #8546, opened 2020-01-22
Implemented php version check in autoload.php

Toflar:
Integrates https://github.com/composer/semver/pull/75 in Composer.
Successfully tested on one of my projects. Ran `composer update` with PHP 7.4.1 and then tried to run the app with 7.3.13 which now doesn't work anymore (in my case, because the version of `ocramius/proxy-manager` in my `composer.lock` requires `7.4.*`) :)


naderman:
Should we do this just for PHP or also all extensions?

Toflar:
From my experience, most devs tend to only require `"ext-foobar": "*"` so it would most likely end up in unbound comparisons anyway. And we'd have quite some files to include 😄

naderman:
@Toflar Well I'd compile the check into a single file anyway, no need for multiple files?

naderman:
But yeah for most extensions I guess we just check presence, then again any unbounded constraint can just be skipped for code generation anyway?

Toflar:
The code generation currently just generates `return true;` for unbound constraints. Of course compiling into one file would be possible but not really the way the semver PR is currently built.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -552,21 +555,58 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
...
 
         return <<<AUTOLOAD
 <?php
 
 // autoload.php @generated by Composer
 
-require_once $vendorPathToTargetDirCode;
+\$phpVersionCheck = require_once $phpVersionCheckPath;
+if (!\$phpVersionCheck) {
+    die('The PHP version you are running does not match the combination in your composer.lock. You likely ran "composer update" with a different PHP version than the one you are running right now!');
```
naderman: I think the error message should display current version and the constraint used for the check? So probably move the check into the required file? Or just compile the full check into this file, so we don't need to add another file?

naderman: Further I think this message is quite unclear. What does "does not match the combination" mean.

I think it should clearly state that there is at least one dependency in composer.lock requiring a higher or lower version.

Toflar: That's all not really possible with the current semver PR...

Seldaek: Regarding wording I'd maybe suggest `'The PHP version you are running does not match the one required by your project or dependencies. You likely ran "composer update" with a different PHP version than the one you are running right now! Run composer install with '.PHP_VERSION.' to get more details.'`

I have a bigger worry though.. that suddenly Composer turns from a passive dev time tool to an active runtime "white page of death" generator, if anything goes bad with the semver PR or just if someone had been running fine with the wrong version (which I could imagine as some projects require higher PHP than they really need). This can wipe out a production site due to a composer self-update.. I wonder if we should start with a warning of some sort that gives the code a chance to continue? And/or maybe define a constant or something that frameworks could act upon, as they can much better decide how to react at runtime given env etc.

cc @nicolas-grekas wdyt of the above, and did you have any bad feedback from the flex experiment so far?

nicolas-grekas: @crell expressed some surprise in https://twitter.com/Crell/status/1219667356114477057 but I think this PR does a better job.

But so far so good, we've got no negative feedback. No prod "broken".

On the other side, here is a recent example of someone that felt into the trap:
https://github.com/symfony/symfony/pull/33497#discussion_r371698580

I think there will be as many "broken prod" as there are now. The difference will be that people will know immediately what to do, wheras now they wrongly complain to upstream projets.

We should do it IMHO.

Seldaek: Maybe something else to add to the die message, and if possible doing it conditionally, would be that if config.platform.php is defined we would change the message to add `'Run composer update again after adjusting your config.platform.php to be '.PHP_VERSION.'.'` or something along those lines instead of the Run composer install with php x, because if you have platform defined it won't actually do anything for you.

naderman: @Toflar I'd say then let's working on making it possible in the semver PR? ;-) I did realize this would require some additional changes there when i mentioned it. I think it should be possible to do all of what I suggested without too much extra work there?

naderman: And indeed mentioning config platform version if defined makes sense too.

Crell: Having been bitten by the Symfony version of this, I'm -1 overall.  At least putting it in composer.* rather than symfony.* would make it easier to find.  If it's going to go forward, extremely clear documentation and error messages are a must, including "if your dev and prod are different versions, you should use config.platform.php to pin to the production version of PHP you're using."

That said, working around a similar feature in Bundler has been a pain, and this kind of version locking is going to increase my workload maintaining templates for a few dozen projects.  At least now I'll know what's going on, but it's still going to add more work for me.

stof: we could add a composer config setting to opt-out of this check if you know what you are doing

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -552,21 +555,58 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
...
+            $phpVersionCheckPath = $vendorPathToTargetDirCode . " . '/php_version_check.php'";
         }
 
         return <<<AUTOLOAD
 <?php
 
 // autoload.php @generated by Composer
 
-require_once $vendorPathToTargetDirCode;
+\$phpVersionCheck = require_once $phpVersionCheckPath;
```
stof: you should not use `require_once` but `require` if you care about the value returned by the file.

stof: also, I suggest moving this check to the `autoload_real.php` file, so that it runs only once rather than each time `autoload.php` is required to return the loader instance.

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: Kinda against a new config setting if we can avoid it. The setting to use is the platform config IMO, then you'll most likely avoid this kind of issue.

Toflar: I've updated the SemVer PR to support multiple packages and thus checks. While implementing the changes in this PR, I noticed a new problem. So it's pretty easy to find out which packages are platform packages (the regex) but finding out the right version is not easy. For the `php` package it is `PHP_VERSION` for example. For other extensions it's `ReflectionExtension` and again for others, we have `PHP_INT_SIZE` (64-bit or not), `PHP_ZTS`, `curl_version()` and so on and so forth... @naderman any idea on how to deal with that? Also, I'm not sure if it's a good idea to use `ReflectionExtension` on every request (even though this might actually be optimized by OPcode maybe...).
I haven't moved the check to `autoload_real.php` yet.

composer.json
```diff
@@ -24,7 +24,7 @@
     "require": {
         "php": "^5.3.2 || ^7.0",
         "composer/ca-bundle": "^1.0",
-        "composer/semver": "^1.0",
+        "composer/semver": "dev-feature/constraint-bounds",
```
staabm: This needs to be fixed before merge

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: @Toflar had a chat with @naderman and we agreed to the following:

- php should check for PHP_VERSION
- extensions should only check for presence, i.e. extension_loaded, more is not worth it given most of the versions are garbage and it might be hard to get to
- ignore zts, php-64bit and whatever else

That'll be a good starting point if we can get this shipped :)

src/Composer/Autoload/AutoloadGenerator.php
Toflar: I moved it, so I'm marking this discussion as resolved.

src/Composer/Autoload/AutoloadGenerator.php
Toflar: Here we go. Updated this PR to the latest semver changes. I'll mark this conversation as resolved because it's all different now anyway :)

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -570,6 +572,69 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
...
+        $requiredExtensions = var_export($requiredExtensions, true);
+
+        return <<<PLATFORM_CHECK
+<?php
+
+// platform_check.php @generated by Composer
+
+\$issues = array();
+
+\$lowerBound = '$lowestPhpVersion';
```
naderman, line 613: This is generated code, the goal is to be optimized, instead of assigning variables for later use, just use the actual value in all places where the variable is used?

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -570,6 +572,69 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
...
+    \$issues[] = sprintf('Your Composer dependencies require PHP in at least version %s and not higher than %s. You are running %s.',
+        \$lowerBound,
+        \$upperBound,
+        PHP_VERSION
+    );
+}
+
+\$missingExtensions = array_diff($requiredExtensions, array_map('strtolower', get_loaded_extensions()));
+
+if (0 !== count(\$missingExtensions)) {
```
naderman, line 627: there a reason for this check as opposed to a simple `!\$missingExtensions`?

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -570,6 +572,69 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
...
+
+// platform_check.php @generated by Composer
+
+\$issues = array();
+
+\$lowerBound = '$lowestPhpVersion';
+\$upperBound = '$highestPhpVersion';
+\$phpVersion = PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION . '.' . PHP_RELEASE_VERSION . '0.0';
+
+if (version_compare(\$phpVersion, \$lowerBound, '<') || version_compare(\$phpVersion, \$upperBound, '>=')) {
```
naderman, line 617: How do you know it's "<" or ">=", shouldn't this depend on whether the bounds are inclusive or not?

src/Composer/Autoload/AutoloadGenerator.php
Toflar: Matter of taste I guess :) Changed in 27130d165185c6b4d5eb0dc5120b6f19fb3b2b54.

src/Composer/Autoload/AutoloadGenerator.php
Toflar: Right, fixed in 27130d165185c6b4d5eb0dc5120b6f19fb3b2b54.

src/Composer/Autoload/AutoloadGenerator.php
Toflar: Type safety, I'm just used to it and it's my CS in all of my projects. Want me to change to `!\$missingExtensions`?

tests/Composer/Test/Autoload/Fixtures/platform_check.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+// platform_check.php @generated by Composer
+
+$issues = array();
+$phpVersion = PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION . '.' . PHP_RELEASE_VERSION . '0.0';
```
Seldaek, line 6: There's an zero too many here after PHP_RELEASE_VERSION (before the dot)

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -570,6 +572,64 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
...
+
+        return <<<PLATFORM_CHECK
+<?php
+
+// platform_check.php @generated by Composer
+
+\$issues = array();
+\$phpVersion = PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION . '.' . PHP_RELEASE_VERSION . '0.0';
+
+if (!version_compare(\$phpVersion, '$lowestPhpVersion', '$lowestOperator') && !version_compare(\$phpVersion, '$highestPhpVersion', '$highestOperator')) {
```
Seldaek, line 616: If you modify lowestPhpVersion and highestPhpVersion to look like PHP_VERSION_ID format, then we could dump this if as a much simpler `if (PHP_VERSION_ID $operator $lowestPhpVersion && ...) {}`. That'd be much more likely to be optimized away by opcache IMO as comparing two constant ints is very easy to figure out.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -618,6 +678,8 @@ public static function getLoader()
             return self::\$loader;
         }
 
+        require_once __DIR__ . '/platform_check.php';
```
Seldaek, line 681: Why require_once and not simple require? AFAIK the latter is a tiny bit faster, and anyway this code won't execute more than once so seems it'd be better.

Toflar: Done in 60c6ffbb5ddd9c59ecf855d78e9efb56a3683064. I've copied from a bit further down. Seems like `require` and `require_once` is mixed here :)

Seldaek: Yup I noticed this too. @nicolas-grekas any clue why autoload_static is required with require_once vs require? Accidental or was there a reason? See https://github.com/composer/composer/commit/fd2f51cea8e5f1ef978cd8f90b87b69dc1778976#diff-33037407f7f293f37df2258567b56598R584

tests/Composer/Test/Autoload/Fixtures/platform_check.php
Toflar: Lol, I had that locally but never pushed. Anyway, marking as resolved because we switched to `PHP_VERSION_ID` now.

src/Composer/Autoload/AutoloadGenerator.php
Toflar: Done in 064af60c27123e20670a12c48203c226dc2616e6. The comparison might look a bit strange because it's inversed but it then allows for a more concise error message. I prefer this :)

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -570,6 +572,79 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
...
+            $chunks = array_map('intval', explode('.', $version));
+            return $chunks[0] * 10000 + $chunks[1] * 100 + $chunks[2];
+        };
+
+        $lowestOperator = $lowestPhpVersion->isInclusive() ? '>=' : '>';
+        $highestOperator = $highestPhpVersion->isInclusive() ? '<=' : '<';
+        $lowestPhpVersionId = $formatToPhpVersionId($lowestPhpVersion);
+        $highestPhpVersionId = $formatToPhpVersionId($highestPhpVersion);
+        $lowestPhpVersion = $lowestPhpVersion->getVersion();
+        $highestPhpVersion = $highestPhpVersion->getVersion();
```
Seldaek, line 621: Maybe strip the last `.0` from these two? It makes the error message look unnecessarily strange as PHP versions never have 4 components.

Toflar: Done in 4e36e16b967b3da9edf1c9e087189d0e7cb4009e. Let me add some more tests.

GromNaN:
Thank you for this useful feature. It is a breaking change that should be noted to [`UPGRADE-2.0.md`](https://github.com/composer/composer/blob/master/UPGRADE-2.0.md)

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: none I can recall of:)

stof: I'm quite sure this comes from the fact that the PR was putting it in `autoload.php` initially.

Seldaek:
Thanks! 

@GromNaN that's right it should be mentioned, I'll update UPGRADE/CHANGELOG sometime this week again with latest changes.

it-can:
Getting this error now


```
composer --version
PHP Warning:  require(phar:///usr/local/bin/composer/vendor/composer/platform_check.php): failed to open stream: phar error: "vendor/composer/platform_check.php" is not a file in phar "/usr/local/bin/composer" in phar:///usr/local/bin/composer/vendor/composer/autoload_real.php on line 25

Warning: require(phar:///usr/local/bin/composer/vendor/composer/platform_check.php): failed to open stream: phar error: "vendor/composer/platform_check.php" is not a file in phar "/usr/local/bin/composer" in phar:///usr/local/bin/composer/vendor/composer/autoload_real.php on line 25
PHP Fatal error:  require(): Failed opening required 'phar:///usr/local/bin/composer/vendor/composer/platform_check.php' (include_path='.:/usr/local/Cellar/php/7.4.5_1/share/php/pear') in phar:///usr/local/bin/composer/vendor/composer/autoload_real.php on line 25

Fatal error: require(): Failed opening required 'phar:///usr/local/bin/composer/vendor/composer/platform_check.php' (include_path='.:/usr/local/Cellar/php/7.4.5_1/share/php/pear') in phar:///usr/local/bin/composer/vendor/composer/autoload_real.php on line 25
```

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: Ok in that case I'll switch it to require and hopefully nothing breaks.

curzio-della-santa:
I didn't analyzed the code in detail, but I'm pretty sure this change does not takes care of the provided PHP-extensions, as is in the case of using [alcaeus/mongo-php-adapter](https://detailnet.atlassian.net/browse/DENNER/fixforversion/29071)

We had a production issue on deploy, because we perform an autoloader optimization downloading the composer.phar from https://getcomposer.org/composer.phar  that currently returns the 2.x dev branch (issue is already open https://github.com/composer/composer/issues/8831 )

Seldaek:
@curzio-della-santa that's absolutely correct, I am coincidentally just working on handling this :) Will commit a fix soon.

PR #8557, opened 2020-01-29
Optimize loading of deps from fixed packages

naderman:
Replaces https://github.com/composer/composer/pull/8535

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -53,14 +53,21 @@ class PoolBuilder
 
     public function buildPool(array $repositories, Request $request)
     {
-        $pool = new Pool();
-
-        // TODO do we really want the request here? kind of want a root requirements thingy instead
         $loadNames = array();
         foreach ($request->getFixedPackages() as $package) {
+            // TODO do we need to add this to nameConstraints at all?
             $this->nameConstraints[$package->getName()] = null;
             $this->loadedNames[$package->getName()] = true;
-            unset($loadNames[$package->getName()]);
+
+            // replace means conflict, so if a fixed package replaces a name, no need to load that one, packages would conflict anyways
+            foreach ($package->getReplaces() as $link) {
+                $this->nameConstraints[$package->getName()] = null;
+                $this->loadedNames[$link->getTarget()] = true;
+            }
+
+            // TODO in how far can we do the above for conflicts? It's more tricky cause conflicts can be limited to
+            // specific versions while replace is a conflict with all versions of the name
+
             if (
                 $package->getRepository() instanceof RootPackageRepository
                 || $package->getRepository() instanceof PlatformRepository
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Fixtures/installer/solver-problems.test
```diff
@@ -61,8 +61,10 @@ Your requirements could not be resolved to an installable set of packages.
   Problem 2
     - The requested package bogus/pkg could not be found in any version, there may be a typo in the package name.
   Problem 3
-    - The requested package stable-requiree-excluded/pkg could not be found in any version, there may be a typo in the package name.
+    - The requested package dependency/pkg could not be found in any version, there may be a typo in the package name.
```
stof: this problem looks wrong to me. There are 2 versions in the repository. Could we update the problem report to check repositories themselves for package name existence rather than the optimized repository set or the pool builder ? The fact that we haven't loaded metadata for the package due to optimization does not mean that this package does not exist in any version.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -50,14 +50,21 @@ public function __construct(array $acceptableStabilities, array $stabilityFlags,
 
     public function buildPool(array $repositories, Request $request)
     {
-        $pool = new Pool();
-
-        // TODO do we really want the request here? kind of want a root requirements thingy instead
         $loadNames = array();
         foreach ($request->getFixedPackages() as $package) {
+            // TODO do we need to add this to nameConstraints at all?
```
Seldaek: I'd say yes as the fact it's null avoid any merging of constraints below in loadPackage (line 211)

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -70,17 +77,23 @@ public function buildPool(array $repositories, Request $request)
...
             if (isset($this->loadedNames[$packageName])) {
                 continue;
             }
-            // TODO currently lock above is always NULL if we adjust that, this needs to merge constraints
-            // TODO does it really make sense that we can have install requests for the same package that is actively locked with non-matching constraints?
-            // also see the solver-problems.test test case
-            $constraint = array_key_exists($packageName, $loadNames) ? null : $constraint;
+
             $loadNames[$packageName] = $constraint;
+            // TODO do we even need to set these in name constraints or can we skip them entirely?
```
Seldaek: Same comment as above.

tests/Composer/Test/Fixtures/installer/solver-problems.test
Seldaek: This is being taken care of somewhere else, will send a PR later or tomorrow when I'm done.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -53,14 +53,21 @@ public function __construct(array $acceptableStabilities, array $stabilityFlags,
...
-            unset($loadNames[$package->getName()]);
+
+            // replace means conflict, so if a fixed package replaces a name, no need to load that one, packages would conflict anyways
+            foreach ($package->getReplaces() as $link) {
+                $this->nameConstraints[$package->getName()] = null;
+                $this->loadedNames[$link->getTarget()] = true;
+            }
+
+            // TODO in how far can we do the above for conflicts? It's more tricky cause conflicts can be limited to
+            // specific versions while replace is a conflict with all versions of the name
```
Seldaek: I guess we could add a reverse constraint to the conflict constraints to `nameConstraints[conflictTarget]`, but not sure how to "add a reverse constraint" exactly ;)

naderman: I think I'd like to leave this comment for a potential future optimization, but merge the PR as is.

PR #8558, opened 2020-01-29
Improve error reporting of solver issues

### LINKED ISSUE #6513: Imrove message when package was not found
In combination with satis I tried to create a project. I've got the message `Could not find package myCompany/myPackage with stability stable. `. Since I've also just created a stable version 1.0.0 I was wondering why composer did not find the stable version. Took me some time to find out that it did not find the package at all. (I think my mistake was that I thought composer would evaluate an existing composer.json in the directory which has a repository declaration for the the satis server - specifying it as a command parameter is the correct approch - not sure). 

So you might improve the message to something like: 
* Package not found at all: `Could not find package myCompany/myPackage in any version.` or `Could not find package myCompany/myPackage  in any version at repository http://some.satis.url .`
* Package found without needed stability: `Found package myCompany/myPackage, but not with stability stable.` 

Or something like that.

### LINKED ISSUE #7254: Improve error reporting for conflicts/replaces
This:

```
{
    "require": {
        "symfony/symfony": "^3.4"
    },
    "replace": {
        "symfony/asset": "*"
    }
}
```
Yields:
```

  Problem 1
    - don't install symfony/symfony v3.4.0|remove __root__ No version set (parsed as 1.0.0)
    - don't install symfony/symfony v3.4.1|remove __root__ No version set (parsed as 1.0.0)
    - don't install symfony/symfony v3.4.2|remove __root__ No version set (parsed as 1.0.0)
    - don't install symfony/symfony v3.4.3|remove __root__ No version set (parsed as 1.0.0)
    - don't install symfony/symfony v3.4.4|remove __root__ No version set (parsed as 1.0.0)
    - Installation request for __root__ No version set (parsed as 1.0.0) -> satisfiable by __root__[No version set (parsed as 1.0.0)].
    - Installation request for symfony/symfony ^3.4 -> satisfiable by symfony/symfony[v3.4.0, v3.4.1, v3.4.2, v3.4.3, v3.4.4].
```

Because both the root package and symfony/symfony replace symfony/asset. So they can not be installed together. It would be good if this was clearer in the output.

### LINKED ISSUE #8525: Report problems properly using the RepositorySet from the Pool or Problem/Rule
See TODOs added by #8444

Seldaek:
Refs #7779 (incl fixes #8525, fixes #6513 and fixes #7254)

Things are already a lot cleaner now. Probably not perfect but big improvements already.

### DIFF
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -188,11 +188,52 @@ abstract class Rule
                 return $text;
 
             case self::RULE_PACKAGE_OBSOLETES:
+                if (count($literals) === 2 && $literals[0] < 0 && $literals[1] < 0) {
+                    $package1 = $pool->literalToPackage($literals[0]);
+                    $package2 = $pool->literalToPackage($literals[1]);
+                    $conflictingNames = array_values(array_intersect($package1->getNames(), $package2->getNames()));
+                    $provideClash = count($conflictingNames) > 1 ? '['.implode(', ', $conflictingNames).']' : $conflictingNames[0];
+
+                    if ($conflictingNames && isset($installedMap[$package1->id]) && !isset($installedMap[$package2->id])) {
+                        // swap vars so the if below passes
+                        $tmp = $package2;
+                        $package2 = $package1;
+                        $package1 = $tmp;
+                    }
+                    if ($conflictingNames && !isset($installedMap[$package1->id]) && isset($installedMap[$package2->id])) {
+                        return $package1->getPrettyString().' can not be installed as that would require removing '.$package2->getPrettyString().'. They both provide '.$provideClash.' and can thus not coexist.';
+                    }
+                    if (!isset($installedMap[$package1->id]) && !isset($installedMap[$package2->id])) {
+                        if ($conflictingNames) {
+                            return 'Only one of these can be installed: '.$package1->getPrettyString().', '.$package2->getPrettyString().'. They both provide '.$provideClash.' and can thus not coexist.';
+                        }
+
+                        return 'Only one of these can be installed: '.$package1->getPrettyString().', '.$package2->getPrettyString().'.';
+                    }
+                }
+
                 return $ruleText;
             case self::RULE_INSTALLED_PACKAGE_OBSOLETES:
                 return $ruleText;
             case self::RULE_PACKAGE_SAME_NAME:
-                return 'Same name, can only install one of: ' . $this->formatPackagesUnique($pool, $literals) . '.';
+                $conflictingNames = null;
+                $allNames = array();
+                foreach ($literals as $literal) {
+                    $package = $pool->literalToPackage($literal);
+                    if ($conflictingNames === null) {
+                        $conflictingNames = $package->getNames();
+                    } else {
+                        $conflictingNames = array_values(array_intersect($conflictingNames, $package->getNames()));
+                    }
+                    $allNames = array_unique(array_merge($allNames, $package->getNames()));
+                }
+                $provideClash = count($conflictingNames) > 1 ? '['.implode(', ', $conflictingNames).']' : $conflictingNames[0];
+
+                if ($conflictingNames && count($allNames) > 1) {
+                    return 'Only one of these can be installed: ' . $this->formatPackagesUnique($pool, $literals) . '. They all provide '.$provideClash.' and can thus not coexist.';
+                }
+
+                return 'Only one of these can be installed: ' . $this->formatPackagesUnique($pool, $literals) . '.';
             case self::RULE_PACKAGE_IMPLICIT_OBSOLETES:
                 return $ruleText;
             case self::RULE_LEARNED:
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -162,7 +162,7 @@ abstract class Rule
                 $package1 = $pool->literalToPackage($literals[0]);
                 $package2 = $pool->literalToPackage($literals[1]);
 
-                return $package1->getPrettyString().' conflicts with '.$this->formatPackagesUnique($pool, array($package2)).'.';
+                return $package2->getPrettyString().' conflicts with '.$package1->getPrettyString().'.';
 
             case self::RULE_PACKAGE_REQUIRES:
                 $sourceLiteral = array_shift($literals);
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -196,18 +196,18 @@ abstract class Rule
             case self::RULE_PACKAGE_IMPLICIT_OBSOLETES:
                 return $ruleText;
             case self::RULE_LEARNED:
-                // TODO not sure this is a good idea, most of these rules should be listed in the problem anyway
-                $learnedString = '(learned rule, ';
                 if (isset($learnedPool[$this->reasonData])) {
+                    $learnedString = ', learned rules:'."\n        - ";
+                    $reasons = array();
                     foreach ($learnedPool[$this->reasonData] as $learnedRule) {
-                        $learnedString .= $learnedRule->getPrettyString($repositorySet, $request, $installedMap, $learnedPool);
+                        $reasons[] = $learnedRule->getPrettyString($repositorySet, $request, $installedMap, $learnedPool);
                     }
+                    $learnedString .= implode("\n        - ", array_unique($reasons));
                 } else {
-                    $learnedString .= 'reasoning unavailable';
+                    $learnedString = ' (reasoning unavailable)';
                 }
-                $learnedString .= ')';
 
-                return 'Conclusion: '.$ruleText.' '.$learnedString;
+                return 'Conclusion: '.$ruleText.$learnedString;
             case self::RULE_PACKAGE_ALIAS:
                 return $ruleText;
             default:
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -15,6 +15,7 @@ namespace Composer\DependencyResolver;
 use Composer\Package\CompletePackage;
 use Composer\Package\Link;
 use Composer\Package\PackageInterface;
+use Composer\Repository\RepositorySet;
 
 /**
  * @author Nils Adermann <naderman@naderman.de>
@@ -122,8 +123,9 @@ abstract class Rule
 
     abstract public function isAssertion();
 
-    public function getPrettyString(Pool $pool, array $installedMap = array(), array $learnedPool = array())
+    public function getPrettyString(RepositorySet $repositorySet, Request $request, array $installedMap = array(), array $learnedPool = array())
     {
+        $pool = $repositorySet->getPool();
         $literals = $this->getLiterals();
 
         $ruleText = '';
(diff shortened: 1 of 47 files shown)


tests/Composer/Test/Fixtures/installer/solver-problems.test
```diff
@@ -57,14 +107,44 @@ Updating dependencies
...
+    - Root composer.json requires package/found7 2.* -> satisfiable by package/found7[2.0.0].
+    - package/found7 2.0.0 requires php-64bit 1.0.1 -> your php-64bit version (%s) does not satisfy that requirement.
+  Problem 14
     - Root composer.json requires requirer/pkg 1.* -> satisfiable by requirer/pkg[1.0.0].
-    - requirer/pkg 1.0.0 requires dependency/pkg 1.0.0 -> no matching package found.
+    - requirer/pkg 1.0.0 requires dependency/pkg 1.0.0 -> found dependency/pkg[1.0.0] but it conflicts with your root composer.json require (2.*).
+  Problem 15
+    - requirer/pkg 1.0.0 requires dependency/pkg 1.0.0 -> found dependency/pkg[1.0.0] but it conflicts with your root composer.json require (2.*).
+    - package/found5 2.0.0 requires requirer/pkg 1.* -> satisfiable by requirer/pkg[1.0.0].
+    - Root composer.json requires package/found5 2.* -> satisfiable by package/found5[2.0.0].
```
Seldaek: @naderman 2.0/master also have similar issues to this, so would be glad if you can investigate why it doesn't mean the root require on `dependency/pkg 2.*` in this Problem.

tests/Composer/Test/Fixtures/installer/conflict-between-dependents.test
```diff
@@ -0,0 +1,38 @@
...
+2
+
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
+    - Root composer.json requires conflicter/pkg 1.0.0 -> satisfiable by conflicter/pkg[1.0.0].
+    - victim/pkg 1.0.0 conflicts with conflicter/pkg[1.0.0].
```
Seldaek: @naderman this here I find is quite strange too.. it sounds like the conflictee is conflicting, but I'd say the package declaring the conflict should be the one "conflicting". Do you see this differently? It seems like the Rule's handling of RULE_PACKAGE_CONFLICT causes this, but I can't tell if the literals order is gonna be consistently in the wrong order or if it's just bad luck here.

tests/Composer/Test/Fixtures/installer/solver-problems.test
Seldaek: Added another test for this which might make it easier to work on in isolation https://github.com/composer/composer/pull/8558/commits/504072c9028d384a581fac639f09b8be34d821e5

src/Composer/Repository/ArtifactRepository.php
```diff
@@ -43,6 +43,11 @@ public function __construct(array $repoConfig, IOInterface $io)
         $this->repoConfig = $repoConfig;
     }
 
+    public function getRepoName()
+    {
+        return 'platform repo ('.$this->lookup.')';
```
stof: shouldn't it be `artifact repo` instead ?

src/Composer/Repository/ComposerRepository.php
```diff
@@ -122,6 +124,11 @@ public function __construct(array $repoConfig, IOInterface $io, Config $config,
         $this->loop = new Loop($this->httpDownloader);
     }
 
+    public function getRepoName()
+    {
+        return 'composer repo ('.$this->url.')';
```
stof: are we sure that we never have credentials in the URL here ?

src/Composer/Repository/CompositeRepository.php
```diff
@@ -39,6 +39,11 @@ public function __construct(array $repositories)
         }
     }
 
+    public function getRepoName()
+    {
+        return 'composite repo ('.count($this->repositories).' repos)';
```
stof: should we include the repo names of inner repositories instead ?

src/Composer/Repository/PathRepository.php
```diff
@@ -111,6 +111,11 @@ public function __construct(array $repoConfig, IOInterface $io, Config $config)
         parent::__construct();
     }
 
+    public function getRepoName()
+    {
+        return 'path repo ('.$this->repoConfig['url'].')';
```
stof: same here about credentials

src/Composer/Repository/VcsRepository.php
```diff
@@ -79,6 +79,17 @@ public function __construct(array $repoConfig, IOInterface $io, Config $config,
...
 
+    public function getRepoName()
+    {
+        $driverClass = get_class($this->getDriver());
+        $driverType = array_search($driverClass, $this->drivers);
+        if (!$driverType) {
+            $driverType = $driverClass;
+        }
+
+        return 'vcs repo ('.$driverType.' '.$this->url.')';
```
stof: same here about the credentials in the URL

tests/Composer/Test/Fixtures/installer/provider-conflicts.test
```diff
@@ -0,0 +1,49 @@
...
+2
+
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
+    - __root__ is present at version 1.2.3 and cannot be modified by Composer
+    - provider/pkg 1.0.0 can not be installed as that would require removing __root__ 1.2.3. They both provide [root-provided/transitive-provided, root-replaced/transitive-provided, root-provided/transitive-replaced, root-replaced/transitive-replaced] and can thus not coexist.
```
stof: this message looks confusing to me. `provide` does not imply that you can only have one of them. You can install as many providers as you want. `replace` implies a conflict (that's the only difference between `replace` and `provide` AFAIK)

src/Composer/Repository/ComposerRepository.php
naderman: I wonder if we should simplify this further, I'm not sure how useful the info "composer repo" is for the average dev? Then again I guess it's helpful to see in output when we get bug reports. Maybe flip this around? `$this->url.' (composer repo)`?

src/Composer/Repository/CompositeRepository.php
naderman: I'd say so

src/Composer/Repository/PackageRepository.php
```diff
@@ -58,4 +58,9 @@ protected function initialize()
             $this->addPackage($package);
         }
     }
+
+    public function getRepoName()
+    {
+        return preg_replace('{^array }', 'package ', parent::getRepoName());
```
naderman: Any idea how to make this a bit clearer? List the names of the packages maybe? It's not usually many?

src/Composer/Repository/ComposerRepository.php
Seldaek: Not sure no, will run through some sanitizer method good point, and fix the other ones

src/Composer/Repository/CompositeRepository.php
Seldaek: Yup why not I guess

src/Composer/DependencyResolver/Rule.php
```diff
@@ -178,85 +180,75 @@ public function getPrettyString(Pool $pool, array $installedMap = array(), array
...
+                    }
+                    $allNames = array_unique(array_merge($allNames, $package->getNames()));
+                }
+                $provideClash = count($conflictingNames) > 1 ? '['.implode(', ', $conflictingNames).']' : $conflictingNames[0];
+
+                if ($conflictingNames && count($allNames) > 1) {
+                    return 'Only one of these can be installed: ' . $this->formatPackagesUnique($pool, $literals) . '. They all provide '.$provideClash.' and can thus not coexist.';
+                }
+
+                return 'Only one of these can be installed: ' . $this->formatPackagesUnique($pool, $literals) . '.';
```
naderman: Why did you remove the same name information here? I intentionally added this recently to explain why only one of them can be installed?

tests/Composer/Test/Fixtures/installer/provider-conflicts.test
Seldaek: Yes I guess this should use replace will fix

src/Composer/Repository/PackageRepository.php
Seldaek: IMO this is just gonna add cruft in the output. In practice you either have one package repo per package in which case you know which one it is by the package name, or you have a single one with all packages and then you also know which one it is as it's the only one.

src/Composer/DependencyResolver/Solver.php
```diff
@@ -65,11 +67,12 @@ class Solver
      * @param Pool                $pool
      * @param IOInterface         $io
      */
-    public function __construct(PolicyInterface $policy, Pool $pool, IOInterface $io)
+    public function __construct(PolicyInterface $policy, Pool $pool, IOInterface $io, RepositorySet $repositorySet)
     {
         $this->io = $io;
         $this->policy = $policy;
         $this->pool = $pool;
+        $this->repositorySet = $repositorySet;
```
naderman: Can we remove this from the pool? I would prefer if it continued to only take the pool as input. We'd just need to pass it into the problem for display?

src/Composer/DependencyResolver/Rule.php
Seldaek: I am not sure how the previous message gave more information? "Same name" is fairly vague at least.. I'm happy to improve it tho if you can explain more.

naderman:
I think VersionSelector, used in e.g. create project needs to be adjusted to use the new flags for the repository set?

src/Composer/DependencyResolver/Rule.php
naderman: Well the reason only one of those packages can be installed is that they have the same name, and you can only install one package of a given name? Not sure how else to state that?

naderman: Maybe "You can only install one version of a package, so only one of these can be installed: ..."?

Seldaek:
AFAIK I updated all usages. VersionSelector just calls with a name and constraint

src/Composer/Repository/PackageRepository.php
naderman: Hm true, I guess this is always used in combination with a package name.

tests/Composer/Test/Fixtures/installer/provider-conflicts.test
naderman: And why do the provided packages get listed here? Afterall only replacers result in conflicts?

src/Composer/DependencyResolver/Rule.php
Seldaek: What about the case covered by `provider-conflicts3.test` tho? I guess this should also mention replace rather than provide, because in that case they don't have the same name, but the replace makes it so that they do.

Anyway for this case without replace I can roll back.

tests/Composer/Test/Fixtures/installer/conflict-between-dependents.test
naderman: The solver really sees a conflict as unidirectional, so I guess it's just a matter of what order we stick them into the rule in. In the RuleSetGenerator it looks like we always have the "issuer" of the conflict first, so conflicter/pkg in this case should be the first literal, so probably safe to flip the output around.

src/Composer/DependencyResolver/Solver.php
Seldaek: Done in https://github.com/composer/composer/pull/8558/commits/c41df325d837b898da4aaea14a9c9a09e1db2bb2#diff-56a374bab973be760e61f08887a83c4f I hope that's what you were thinking about

tests/Composer/Test/Fixtures/installer/solver-problems.test
naderman: Turns out the issue here is that dependency/pkg 1.0.0 gets filtered in the pool due to root require, so the error is just that when it looks for 1.* it can't find it, the fact that 2.0.0 is there and conflicts with the removed 1.0.0 is not known to the solver and doesn't become part of the problem.

src/Composer/DependencyResolver/Rule.php
Seldaek: Fixed in https://github.com/composer/composer/pull/8558/commits/189d5adab0dd504d1c4eea2f6dde7f2d9c0302b2

tests/Composer/Test/Fixtures/installer/provider-conflicts.test
Seldaek: Fixed in https://github.com/composer/composer/pull/8558/commits/189d5adab0dd504d1c4eea2f6dde7f2d9c0302b2

PR #8561, opened 2020-01-30
Allow providers which are selected to be installed in place of existing packages which do not satisfy requirements

### LINKED ISSUE #6753: Provide does not satisfy package requirement if package with same name exists
I believe this should be allowed. It works if you swap replace/provide.

```
{
	"repositories": [
		{
			"packagist": false
		},
		{
			"type": "package",
			"package": [
				{
					"name": "foo/polyfill",
					"provide": {
						"foo/standard": "1.0.0"
					},
					"replace": {
					},
					"version": "1.0.0"
				},
				{
					"name": "foo/standard",
					"require": {
						"foo/does-not-exist": "1.0.0"
					},
					"version": "1.0.0"
				}
			],
			"type": "package"
		}
	],
	"require": {
		"foo/standard": "1.0.0",
		"foo/polyfill": "1.0.0"
	}
}
```

Seldaek:
Fixes #6753

I think it's fine really as per the 3 new tests, as they only get installed if selected, this isn't risky.

### DIFF
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -27,7 +27,6 @@ use Composer\Package\PackageInterface;
  */
 class Pool implements \Countable
 {
-    const MATCH_NAME = -1;
     const MATCH_NONE = 0;
     const MATCH = 1;
     const MATCH_PROVIDE = 2;
@@ -117,27 +116,15 @@ class Pool implements \Countable
             $candidates = $this->packageByName[$name];
         }
 
-        $matches = $provideMatches = array();
-        $nameMatch = false;
+        $matches = array();
 
         foreach ($candidates as $candidate) {
             switch ($this->match($candidate, $name, $constraint)) {
                 case self::MATCH_NONE:
                     break;
 
-                case self::MATCH_NAME:
-                    $nameMatch = true;
-                    break;
-
                 case self::MATCH:
-                    $nameMatch = true;
-                    $matches[] = $candidate;
-                    break;
-
                 case self::MATCH_PROVIDE:
-                    $provideMatches[] = $candidate;
-                    break;
-
                 case self::MATCH_REPLACE:
                     $matches[] = $candidate;
                     break;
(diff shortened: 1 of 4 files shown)


src/Composer/DependencyResolver/Pool.php
```diff
@@ -147,11 +147,6 @@ private function computeWhatProvides($name, $constraint, $mustMatchName = false)
             }
         }
 
-        // if a package with the required name exists, we ignore providers
-        if ($nameMatch) {
```
stof: Computing `$nameMatch` is now useless, as it is never read. so it should be removed.

tests/Composer/Test/Fixtures/installer/provider-packages-can-be-installed-if-selected.test
```diff
@@ -0,0 +1,39 @@
...
+    "repositories": [
+        {
+            "type": "package",
+            "package": [
+                {
+                    "name": "foo/polyfill",
+                    "provide": {
+                        "foo/standard": "1.0.0"
+                    },
+                    "replace": {
```
naderman: can be dropped?

tests/Composer/Test/Fixtures/installer/provider-packages-can-be-installed-if-selected.test
```diff
@@ -0,0 +1,39 @@
...
+                },
+                {
+                    "name": "foo/standard",
+                    "require": {
+                        "foo/does-not-exist": "1.0.0"
+                    },
+                    "version": "1.0.0"
+                }
+            ],
+            "type": "package"
```
naderman: duplicate of type: package at the top?

tests/Composer/Test/Fixtures/installer/provider-packages-can-be-installed-together-with-provided-if-both-installable.json
```diff
@@ -0,0 +1,37 @@
...
+                    "replace": {
+                    },
+                    "version": "1.0.0"
+                },
+                {
+                    "name": "foo/standard",
+                    "version": "1.0.0"
+                }
+            ],
+            "type": "package"
```
naderman: same as above test: duplicate type package and unnecessary replace

PR #8565, opened 2020-01-30
Create project: add option to add the repository to the composer.json

glaubinix:
This PR adds an option to the create-project command to also add the repository option to the composer.json of the project.

It prevents doing so if the project already contains a composer.lock file because this will lead to an outdated composer.lock file.

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -69,6 +69,7 @@ class CreateProjectCommand extends BaseCommand
                 new InputOption('prefer-dist', null, InputOption::VALUE_NONE, 'Forces installation from package dist even for dev versions.'),
                 new InputOption('repository', null, InputOption::VALUE_REQUIRED, 'Pick a different repository (as url or json config) to look for the package.'),
                 new InputOption('repository-url', null, InputOption::VALUE_REQUIRED, 'DEPRECATED: Use --repository instead.'),
+                new InputOption('add-repository', null, InputOption::VALUE_NONE, 'Add the repository option to the composer.json.'),
                 new InputOption('dev', null, InputOption::VALUE_NONE, 'Enables installation of require-dev packages (enabled by default, only present for BC).'),
                 new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables installation of require-dev packages.'),
                 new InputOption('no-custom-installers', null, InputOption::VALUE_NONE, 'DEPRECATED: Use no-plugins instead.'),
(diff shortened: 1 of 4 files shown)


src/Composer/Command/CreateProjectCommand.php
```diff
@@ -162,6 +164,24 @@ public function installProject(IOInterface $io, Config $config, InputInterface $
...
+
+        // add the repository to the composer.json and use it for the install run later
+        if ($repository !== null && $addRepository) {
+            if ($composer->getLocker()->isLocked()) {
+                $io->writeError('<error>Adding a repository when creating a project that provides a composer.lock file is not supported</error>');
+
+                return false;
+            }
+
+            $repoConfig = RepositoryFactory::configFromString($io, $config, $repository, true);
```
Seldaek: $config should probably be $composer->getConfig()?

src/Composer/Command/CreateProjectCommand.php
```diff
@@ -69,6 +69,7 @@ protected function configure()
                 new InputOption('prefer-dist', null, InputOption::VALUE_NONE, 'Forces installation from package dist even for dev versions.'),
                 new InputOption('repository', null, InputOption::VALUE_REQUIRED, 'Pick a different repository (as url or json config) to look for the package.'),
                 new InputOption('repository-url', null, InputOption::VALUE_REQUIRED, 'DEPRECATED: Use --repository instead.'),
+                new InputOption('add-repository', null, InputOption::VALUE_NONE, 'Add the repository option to the composer.json.'),
```
Seldaek: Can you add this to doc/03-cli.md too?

tests/Composer/Test/Repository/RepositoryFactoryTest.php
```diff
@@ -46,4 +46,22 @@ public function testManagerWithAllRepositoryTypes()
...
+        $this->assertSame($expected, RepositoryFactory::generateRepositoryName($index, $config, $existingRepos));
+    }
+
+    public function generateRepositoryNameProvider()
+    {
+        return array(
+            array(0, array(), array(), 0),
+            array(0, array(), array(array()), '02'),
+            array(0, array('url' => 'https://example.org'), array(), 'example.org'),
+            array(0, array('url' => 'https://example.org'), array('example.org' => array()), 'example.org2'),
```
Seldaek: Might be good to add tests where $index is a string, as technically the repositories can be a json object  with named keys for each repo instead of numerically-keyed array

glaubinix:
Updated

Seldaek:
Thanks

PR #8566, opened 2020-01-30
Restore installer events and add a PRE_POOL_CREATE hook for plugins

### LINKED ISSUE #8348: Plugin hook to filter the pool before running the solver
Now that Composer 2 will separate loading packages in the pool from running the solver, it would be great if plugins could hook between these 2 steps.

One of the features of Symfony Flex is to force restricting the available range for Symfony components. Currently, it does that by replacing the repository implementation for composer repositories, and filtering out the metadata. That's quite invasive in Composer, and can be done only by one plugin at a time (as it relies on extending the ComposerRepository class). A similar feature could be achieved in Composer 2 by filtering the pool after it has been built, to remove unwanted Symfony versions before the solver runs.

Note that I suggested a bunch of pool optimizations (#8287 for instance). To keep these optimization safe, the hook allowing to add or remove packages from the pool should run before them rather than after them.

Seldaek:
Fixes #8348 - possibly. 

@stof I am not sure what you meant there with: 

> To keep these optimization safe, the hook allowing to add or remove packages from the pool should run before them rather than after them.

I would tend to say the opposite, that our optimizations should run first then the plugins can filter what is left. That's what I implemented here but it's definitely up for discussion where the hook should happen.

Also looping in @nicolas-grekas

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -382,7 +382,6 @@ class Installer
         }
 
         $pool = $repositorySet->createPool($request, $this->eventDispatcher);
-        $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $repositorySet, $pool, $request, $policy);
 
         // solve dependencies
         $solver = new Solver($policy, $pool, $this->io, $repositorySet);
@@ -399,9 +398,6 @@ class Installer
 
             return max(1, $e->getCode());
         }
-
-        // TODO should we warn people / error if plugins in vendor folder do not match contents of lock file before update?
-        $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::POST_DEPENDENCIES_SOLVING, $this->devMode, $repositorySet, $pool, $request, $policy, $lockTransaction);
 
         $this->io->writeError("Analyzed ".count($pool)." packages to resolve dependencies", true, IOInterface::VERBOSE);
         $this->io->writeError("Analyzed ".$ruleSetSize." rules to resolve dependencies", true, IOInterface::VERBOSE);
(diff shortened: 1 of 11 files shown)


Seldaek:
Ping @stof @nicolas-grekas @naderman can I get anyone to look at this? :) Especially when it comes to the new `pre-pool-create` event and re my comment in OP.

nicolas-grekas:
The behavior you describe makes sense to me: composer prepares the list of packages, then plugins can alter it if needed. I see how this would allow flex to prune older Symfony tags.
LGTM in theory :)

src/Composer/Installer.php
```diff
@@ -381,10 +381,8 @@ protected function doUpdate(RepositoryInterface $localRepo, $doInstall)
             }
         }
 
-        // TODO reenable events
-        //$this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $policy, $repositorySet, $installedRepo, $request);
-
-        $pool = $repositorySet->createPool($request);
+        $pool = $repositorySet->createPool($request, $this->eventDispatcher);
+        $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_DEPENDENCIES_SOLVING, $this->devMode, $repositorySet, $pool, $request, $policy);
```
naderman: Shouldn't devMode only be relevant for doInstall?

naderman:
Should we either separate or at least through paramters make it clearer what context the PRE/POST DEPENDENCIES_SOLVING events get triggered? (doUpdate/extractDevPackages/doInstall)?

naderman:
New pool builder event looks good to me

Seldaek:
Yes it probably would make sense to get more context info.. On the other hand I wonder what the value of these pre/post-dependencies-solving events really is.. Maybe a pre-operations-exec or something would be better and it would only happen on doInstall ?

Seldaek:
I feel like before they were used for two cases: filtering/injecting packages (better done with pool builder event now) and pre-fetching packages to be installed (would be possible with the pre-operations-exec I propose). Maybe I am missing something tho.

naderman:
Indeed, I'm actually not sure who uses those at all or for what? The pool builder event should replace any use case I can imagine?

stof:
@Seldaek my sentence was referring to suggested optimizations (not implemented yet), which would rely on knowing the full set of available packages passed to the solver. If listeners are allowed to add more packages in the pool, the suggested optimizations are unsafe if they run before that (as they would run without considering the new packages).

Seldaek:
@stof ok I see. I was thinking more of the flex use case of filtering packages even further. Let's say we keep it as is for now and if there is a sane use case for something else we can add another hook.

PR #8567, opened 2020-01-31
Remove obsolete rules

naderman:
Started out by implementing that a package providing a name should not conflict with a package replacing it. This PR does correct this behavior but it expanded in scope due to the way this was accomplished:

The only implicit conflict (not using the conflict keyword) we have now results from packages using the same name, either by literally having the same name and being different versions, or they replace the same name, so:
    
- removed all types of obsolete rules
- simplified rule generation significantly, since there are now only explicit conflicts or same name conflicts which use a single multi conflict rule per name
- removed match types from Pool as these are no longer used
- matching by name only is no longer available in whatsProvide because mustMatchName was unused
- fixed some language in error handling
    
The effect of this is a significant reduction in rule objects (mostly because of the consistent use of multi conflict rules #8424 for all conflicts now) and consequently memory use and performance.

For example for a contao project:

before
```
[59.8MiB/2.54s] Generating rules
[2041.8MiB/27.25s] Resolving dependencies through SAT
...
[2048.1MiB/36.74s] Dependency resolution completed in 9.492 seconds
[130.6MiB/37.84s] Analyzed 16073 packages to resolve dependencies
[130.6MiB/37.84s] Analyzed 4114091 rules to resolve dependencies
[130.6MiB/37.84s] Lock file operations: 162 installs, 0 updates, 0 removals
...
[77.8MiB/37.88s] Memory usage: 77.79MiB (peak: 2048.55MiB), time: 37.88s
```

After:
```
[58.8MiB/2.59s] Generating rules
[1028.8MiB/15.63s] Resolving dependencies through SAT
....
[1034.2MiB/20.30s] Dependency resolution completed in 4.669 seconds
[96.1MiB/20.67s] Analyzed 16073 packages to resolve dependencies
[96.1MiB/20.67s] Analyzed 1747077 rules to resolve dependencies
[96.2MiB/20.67s] Lock file operations: 162 installs, 0 updates, 0 removals
...
[45.8MiB/23.38s] Memory usage: 45.77MiB (peak: 1034.6MiB), time: 20.71s
```

So close to 50% in memory reduction and runtime reduction.

### DIFF
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -27,11 +27,6 @@ use Composer\Package\PackageInterface;
  */
 class Pool implements \Countable
 {
-    const MATCH_NONE = 0;
-    const MATCH = 1;
-    const MATCH_PROVIDE = 2;
-    const MATCH_REPLACE = 3;
-
     protected $packages = array();
     protected $packageByName = array();
     protected $versionParser;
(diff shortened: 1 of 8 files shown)


Seldaek:
Looks good overall for memory usage but CPU time suffers in some cases. @naderman will investigate.

naderman:
Performance issue seemed resolved with other changes on 2.0 branch.

PR #8581, opened 2020-02-05
use Symfony PHPUnit Bridge

IonBazan:
This PR makes use of Symfony PHPUnit Bridge to test against multiple versions of PHPUnit.
`setExpectedException()` has been added to `\Composer\Test\TestCase` to provide compatibility with all versions. Please make sure to always extend this base test class.

- [x] Use Symfony PHPUnit Bridge
- [x] Run tests on PHP 7.4
- [x] Fix PHP 7.4 test issues (`Trying to access array offset on value of type null`)
- [x] Fix risky tests
- [x] Fix Windows build (using php `echo` instead of system one)

Note that tests are no longer run in parallel but it doesn't really affect performance.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,14 +25,14 @@ matrix:
     - php: 7.1
     - php: 7.2
     - php: 7.3
-    - php: 7.3
-      env: deps=high
     - php: nightly
-    - php: 7.4snapshot
+    - php: 7.4
+      env:
+        - deps=high
+        - SYMFONY_PHPUNIT_VERSION=7.5
   fast_finish: true
   allow_failures:
     - php: nightly
-    - php: 7.4snapshot
 
 before_install:
   # disable Xdebug if available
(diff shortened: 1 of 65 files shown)


Seldaek:
Looks good overall, thanks for all the updates and fixes. Just one thing, the 8.0 build fails to run phpunit entirely (which is fine for now IMO), but that failure is not even detected, which makes me a little worried that a catastrophic failure would not be detected as a build failure..

See https://travis-ci.org/composer/composer/jobs/647209828?utm_medium=notification&utm_source=github_status

If you can fix that it'd be great. Seems like it might be an issue in simple-phpunit script though that it returns 0 even though it failed to install or execute phpunit.

Seldaek:
Also please avoid further force-pushes as I'd rather not have to go through the whole changeset once more ;)

IonBazan:
Thank you. Looks like the PHP8 issue is caused by using `passthru` without checking the status by `simple-phpunit`. This has been fixed in versions: 3.4, 4.4 and 5.0 so I'm locking it at `^3.4`.

Seldaek:
Great thanks

PR #8590, opened 2020-02-07
Wrap version constaints that contain * with quotes

bocharsky-bw:
Some shells like ZSH require wrapping * with quotes, otherwise, it shows an error like:

> zsh: no matches found: 2.2.*

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -190,7 +190,7 @@ If you do not want to choose requirements interactively, you can pass them
 to the command.
 
 ```sh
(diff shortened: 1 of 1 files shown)


Seldaek:
Do double quotes work on zsh too? I guess so? If so I'd rather use double as I think single quotes cause issues on windows cmd

bocharsky-bw:
Yes, double quotes works as well. Updated to double quotes, thanks!

Seldaek:
Thanks

PR #8593, opened 2020-02-11
rename signature to checksum

sualko:
Sha-384 is a hash algorithm which can be used as checksum, but it's no digital signature.

### DIFF
--- a/doc/faqs/how-to-install-composer-programmatically.md
+++ b/doc/faqs/how-to-install-composer-programmatically.md
@@ -1,7 +1,7 @@
 # How do I install Composer programmatically?
 
 As noted on the download page, the installer script contains a
-signature which changes when the installer code changes and as such
+checksum which changes when the installer code changes and as such
 it should not be relied upon in the long term.
 
 An alternative is to use this script which only works with UNIX utilities:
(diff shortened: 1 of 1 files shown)


alcohol:
To quote a random Google source

> Digital signatures are a form of checksum, in that they provide data integrity, but they require asymmetric encryption to also provide authenticity. Digital signatures are away to attach an identity to the checksum. This implies a level of trust between you and the 3rd party, such as Debian with our example above. If you have met with the 3rd party, or dealt with them enough to establish some level of trust, then you can install the 3rd party's public key into your system. Then, when they provide data attached with a digital signature, you can verify that the data did in fact come from the 3rd party, and no other source.

So I guess in a semantic sense this change is correct.

sualko:
On a bigger scale I would also rename https://composer.github.io/installer.sig to https://composer.github.io/installer.checksum, but this would have to be prepared to not break the self update. I'm also wondering what the purpose of the public keys on https://composer.github.io/pubkeys.html is, because I see no signature for any tags on github.

In terms of security I would recommend to sign every release and to relay on that signature instead of a checksum, because if the server is compromised it's also easy to change the hash even if you publish it on two sites. A signature key should only be present on one personal computer and therefore easier to protect.

I'm wondering that there was never a discussion before, or didn't I find it in the issue tracker?

alcohol:
Those keys are used by the self-update process if I am not mistaken:

https://github.com/composer/composer/blob/78b8c365cd879ce29016884360d4e61350f0d176/src/Composer/Command/SelfUpdateCommand.php#L170

Seldaek:
Right, the releases are signed and signatures are verified by self-update, the installer is indeed just a checksum.

PR #8594, opened 2020-02-11
Distinguish between updates and downgrades in dry-run

IonBazan:
Currently, running `composer install --dry-run` produces different output than `composer install` when some of packages will be downgraded:

```
$ composer u --dry-run                           
Package operations: 0 installs, 1 update, 0 removals
  - Updating composer/ca-bundle (1.2.6) to composer/ca-bundle (1.0.0)

$ composer u
Package operations: 0 installs, 1 update, 0 removals
  - Downgrading composer/ca-bundle (1.2.6 => 1.0.0): Loading from cache
Writing lock file
Generating autoload files
```

I usually use `--dry-run` switch to see which operations will be executed on a feature branch I am working on and `Upgrading` may be misleading for `1.2.6 -> 1.0.0` transition.

This logic is repeated in `VcsDownloader` and `FileDownloader` so it might be worth to move it somewhere else in the future.

### DIFF
--- a/src/Composer/DependencyResolver/Operation/UpdateOperation.php
+++ b/src/Composer/DependencyResolver/Operation/UpdateOperation.php
@@ -13,6 +13,7 @@
 namespace Composer\DependencyResolver\Operation;
 
 use Composer\Package\PackageInterface;
+use Composer\Package\Version\VersionParser;
 
 /**
  * Solver update operation.
@@ -74,7 +75,9 @@ class UpdateOperation extends SolverOperation
      */
     public function __toString()
     {
-        return 'Updating '.$this->initialPackage->getPrettyName().' ('.$this->formatVersion($this->initialPackage).') to '.
+        $actionName = VersionParser::isUpgrade($this->initialPackage->getVersion(), $this->targetPackage->getVersion()) ? 'Updating' : 'Downgrading';
+
+        return $actionName.' '.$this->initialPackage->getPrettyName().' ('.$this->formatVersion($this->initialPackage).') to '.
             $this->targetPackage->getPrettyName(). ' ('.$this->formatVersion($this->targetPackage).')';
     }
 }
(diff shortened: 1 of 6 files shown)


alcohol:
I'm not sure how `update` can be interpreted semantically.

I suppose one could say that 'updating' dependencies can lead to them either being upgraded or downgraded (or added/removed). But should we then also make a stronger distinction there, rather than only focus on the downgrading aspect here?

Seldaek:
I think this is fine, it's just bringing dry-run in line with the normal runs. Not sure what you mean by stronger distinction @alcohol ?

Seldaek:
Gonna merge anyway, thanks @IonBazan

alcohol:
I was just wondering if we're aiming at semantic wording changes, if the following is a desirable change as well:

Old:

```
--EXPECT--
Downgrading b/unstable (1.1.0-alpha) to b/unstable (1.0.0)
Updating a/old (0.9.0) to a/old (1.0.0)
Downgrading c/uptodate (2.0.0) to c/uptodate (1.0.0)
Installing d/removed (1.0.0)
Installing e/newreq (1.0.0)
```

New:

```
--EXPECT--
Downgrading b/unstable (1.1.0-alpha) to b/unstable (1.0.0)
Upgrading a/old (0.9.0) to a/old (1.0.0)
Downgrading c/uptodate (2.0.0) to c/uptodate (1.0.0)
Installing d/removed (1.0.0)
Installing e/newreq (1.0.0)
```

Since update can be interpreted as both upgrade and downgrade.

Seldaek:
Ah I see, yeah might be worth a look for 2.0..

alcohol:
It's definitely not important, but it seems in-line with some other recent changes. :man_shrugging:

Seldaek:
Done in https://github.com/composer/composer/commit/9609729de7ce39e206161e6512cf69de496d0107

IonBazan:
@Seldaek isn’t the “non-dry” run missing from that?

Seldaek:
Yes that's right, some stuff uses the Operation formatting, but not all, fixed now. Thanks

PR #8562, opened 2020-01-30
Uses config data from Composer object whenever possible on ArchiveCommand

adrianosferreira:
It was previously blindly getting a new instance from the factory thus ignoring what is on Composer object config data.

PR #8595, opened 2020-02-11
Fix risky test

IonBazan:
#8562 introduced a risky test `ArchiveCommandTest::testUsesConfigFromFactoryWhenComposerIsNotDefined()` which did not make sure that the method is really called and contained some unused test code. 
PHPUnit config has been updated to prevent such tests to pass but it will only fail on v5.7 and up because PHPUnit 4.8 does not support `failOnRisky` directive.


Seldaek:
cc @adrianosferreira

Seldaek:
Thanks for the updates

adrianosferreira:
@IonBazan thanks! I will pay more attention next time.

PR #8597, opened 2020-02-12
Git: log error for sync mirror with existing local copy

glaubinix:
In case sync mirror fails if there already is a local copy available this will now output the reason why the sync fails:

Before
```
Failed to update git@github.com:glaubinix/test-repo-rename.git, package information from this repository may be outdated
```

After
```
Sync mirror failed: Failed to execute git remote set-url origin 'git@github.com:glaubinix/test-repo-rename.git' && git remote update --prune origin && git remote set-url origin 'git@github.com:glaubinix/test-repo-rename.git'

git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
error: Could not fetch origin

Failed to update git@github.com:glaubinix/test-repo-rename.git, package information from this repository may be outdated
```

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -97,6 +97,7 @@ class Git
 
         $auth = null;
         if ($bypassSshForGitHub || 0 !== $this->process->execute($command, $ignoredOutput, $cwd)) {
+            $errorMsg = $this->process->getErrorOutput();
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

PR #8612, opened 2020-02-16
Delete variable

souweb22:
Removed variables that I think are not used. I'm sorry if I made any mistakes.


naderman:
Indeed, these were left over from some old code that has since been deleted.

PR #8618, opened 2020-02-18
allow php 8 for nightly testing

alcohol:


alcohol:
Well rats, it seems `vendor/bin/simple-phpunit` ignore the `php.platform` configuration setting.

Seldaek:
Soo closing here? I think it's too early to look at php8 builds really..

alcohol:
I wanted to test them for `satis` but since it depends on `composer/composer` and that still requires 7.x only :-(

composer.json
```diff
@@ -22,7 +22,7 @@
         }
     ],
     "require": {
-        "php": "^5.3.2 || ^7.0",
+        "php": "^5.3.2 || ^7.0 || ^8.0",
```
GrahamCampbell: 👎 for adding this yet. People should test on PHP 8 by telling composer to ignore platform deps. It's too early in the game to say that any package is compatible with PHP 8.

GrahamCampbell:
> I wanted to test them for satis but since it depends on composer/composer and that still requires 7.x only :-(

You can simply run composer with the ignore platform reqs flag?

PR #8639, opened 2020-02-22
Rename how-to-install-composer-programmatically.md to on-how-to-insta…

loqlaq:
…ll-composer-programmatically.md


Seldaek:
Ehh not sure why, but I think no thanks. Also this will break links.

Issue #8648, opened 2020-02-25
run composer commands from within a plugin


staabm:
I am in the process of building a custom plugin. in this plugin I am listening for `ScriptEvents::PRE_AUTOLOAD_DUMP` and want to execute another composer command.

I need to copy a lot of code arround (from composer itself) to be able to do it.

my current code looks like:

```php
class AutoloadPlugin implements PluginInterface, EventSubscriberInterface
{
    public function activate(Composer $composer, IOInterface $io)
    {
        // Nothing to do here, as all features are provided through event listeners
    }

    /**
     * {@inheritdoc}
     */
    public static function getSubscribedEvents(): array
    {
        return [ScriptEvents::PRE_AUTOLOAD_DUMP => 'buildAutoloadClassmap'];
    }

    public static function buildAutoloadClassmap(Event $event)
    {
        /*
        $extra = json_decode('{
            "rocket-autoload": {
                "admin": {
                    "classmap": [
                        "lib/admin/123"
                    ]
                },
                "shop": {
                    "classmap": [
                        "lib/shop/123"
                    ]
                }
            }
        }', true);
        */
        /**
         * @var array<string, array<string, array<string, string>>>
         */
        $extra = $event->getComposer()->getPackage()->getExtra();
        $pluginExtra = $extra['rocket-autoload'] ?? [];

        if (getenv('ROCKET_AUTOLOAD')) {
            if (getenv('COMPOSER_VENDOR_DIR')) {
                // We need the root package so we can add our classmaps to its loader.
                $package = $event
                    ->getComposer()
                    ->getPackage();

                $autoload = $package->getAutoload();

                if (empty($autoload['classmap'])) {
                    $autoload['classmap'] = [];
                }

// do my autoloading maginc

                $package->setAutoload($autoload);
            }
        } else {
            $io = $event->getIO();
// copied from Composer EventDispatcher
            $cmd = self::getPhpExecCommand() .' '. ProcessExecutor::escape(getenv('COMPOSER_BINARY')) . ' dump-autoload';
            putenv("ROCKET_AUTOLOAD=1");

            // dump a new autoloader for-each app
            foreach ($pluginExtra as $appName => $additionalClassMap) {
                putenv("COMPOSER_VENDOR_DIR=vendor-$appName");

                $process = new ProcessExecutor($io);
                $process->execute($cmd);
            }
        }
    }

// copied from Composer EventDispatcher
    static protected function getPhpExecCommand()
    {
[...]

PR #8649, opened 2020-02-25
Added `autoloader-dir` parameter to `dump-autoload` command

clxmstaab:
I need several autoloaders because I am in a classical monorepo setup which contains several applications. the whole monorepo contains just a single composer.json.

I want composer to build a separate autoloader per application. this is something I will implent via a composer plugin (paths per app will be configured via composer.json `extra` attribute).
therefore I need to define a different class-loader name to make sure composer will not override all the files over and over again.

the use-case of this change was discussed in detail in https://github.com/composer/composer/issues/8648

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -29,6 +29,8 @@ use Composer\Util\PackageSorter;
  */
 class AutoloadGenerator
 {
+    public const DEFAULT_AUTOLOADER_NAME = 'composer';
+
     /**
      * @var EventDispatcher
      */
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -29,7 +29,7 @@ use Composer\Util\PackageSorter;
  */
 class AutoloadGenerator
 {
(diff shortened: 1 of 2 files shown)


Seldaek:
I'm kinda reluctant to add public-facing API to commands for edge case scenarios in plugins.. Adding the option to the AutoloadGenerator itself, why not I guess although I'd rather do without.

Why not call the autoload generator directly from your plugin, passing it a custom `$targetDir` fitting your needs?

The code from dump-autoload command isn't all that much to duplicate in your plugin, especially as you won't need half of it.

staabm:
after trying with this change in my local fork I came to the conclusion that I need other means for the use case.

I will try with nette/robot-loader which has a better api for my use-case to generate a loader.

Seldaek:
Yeah one thing to keep in mind is that PHP supports having multiple autoloaders.. so you could very easily use the composer one from all apps to get dependencies loaded, and then a secondary one per app with just that app's classmap.

staabm:
> Why not call the autoload generator directly from your plugin, passing it a custom `$targetDir` fitting your needs?

I dont want to leave this question open:

I cannot do this, because the vendor/autoload.php will be overriden on each invocation (so only the last one will win)

Issue #8650, opened 2020-02-25
"Ambiguous class resolution" message contains "$baseDir" variable


clxmstaab:
when composer builds a classmap and finds ambigious classes it reports them as a warning:

https://github.com/composer/composer/blob/f154d5c09d69a6338d6a955bf51cc264d0448b3d/src/Composer/Autoload/AutoloadGenerator.php#L329-L344

the warning is as smart, as a variable `vendorDir` is resolved, so we get a absolute path in https://github.com/composer/composer/blob/master/src/Composer/Autoload/AutoloadGenerator.php#L338

the path returned by `getPathCode()` contains in some cases a `$baseDir` variable instead of `$vendorDir`.

https://github.com/composer/composer/blob/f154d5c09d69a6338d6a955bf51cc264d0448b3d/src/Composer/Autoload/AutoloadGenerator.php#L551-L555

this variable does not get resolved and appears in the end-user message, see e.g.

```
mstaab@COMPLEX-67 MINGW64 /c/dvl/Zend/DefaultWorkspace13/kunzmann (production)
$ php ../composer/bin/composer dump-autoload  --autoloader-dir="autoload-shop"
Generating optimized autoload files

Warning: Ambiguous class resolution, "Mail" was found in "$baseDir . '/app/shop/models/Mail.php" and "C:/dvl/Zend/DefaultWorkspace13/kunzmann/app/shared/models/Mail.php", the first will be used.
```

PR #8651, opened 2020-02-25
resolve `$baseDir` in ambigious class resolution error

clxmstaab:
closes https://github.com/composer/composer/issues/8650

before this change
```
Warning: Ambiguous class resolution, "Composer\Util\Zip",
 was found in both "$baseDir . '/src/Composer/Util/Zip.php" and "C:/dvl/Zend/DefaultWorkspace13/composer/src2/Composer/Util/Zip.php", the first will be used.
```

after this change
```
Warning: Ambiguous class resolution, "Composer\Util\Zip",
 was found in both "C:/dvl/Zend/DefaultWorkspace13/composer/src/Composer/Util/Zip.php" and "C:/dvl/Zend/DefaultWorkspace13/composer/src2/Composer/Util/Zip.php", the first will be used.
```

`$baseDir` is getting resolved within the paths of the error message.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -326,7 +326,6 @@ EOF;
         return 0;
     }
 
-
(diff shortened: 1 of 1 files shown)


Seldaek:
Probably makes more sense to do this in one go with #8652 to avoid conflicts, so closing here.

Issue #8654, opened 2020-02-25
Add PluginEvents::POST_FILE_DOWNLOAD event


heddn:
We're working on trying to build an update mechanism for Drupal that supports composer. One of the features we have currently with the MVP (without composer) is/was to do some security hash checking that what we downloaded matches what we expected. In looking to extend that into the composer world, it would be nice to have a POST download event. So we can do what we want and validate what we downloaded is what we expected.

heddn:
Questions came up in Drupal slack about this approach. I'm going to pose them and answer them so we have the thoughts captured in a persistent location.


> @greg-1-anderson
> I'm not sure if this new event is going to be substantially different than the existing post-package-update event

Response:
I don't see these as the same thing at all. I believe that post package-{update|install} has already moved the files into the final destination. Post file download only fires in FileDownloader, meaning for tagged/packaged archives. Which is exactly what we want. We don't want to trigger it for source or git cloned things. Because git already has security built-in for those.

And prior to moving into the final destination, I want to have a chance to review and throw an exception if there's a MiM attack. Where I triggered the event, it hasn't moved them from temp yet.

bruno-ds:
My question is probably stupid, but still, I don't see why:
If you use a "**trusted**" + "**https**" endpoint, there is no room for a man in the middle attack, no?

heddn:
@bruno-ds for Drupal, we've opted to not assume that every hosting site, etc has a fully functional SSL setup. Yes, much of that is in the past, but to _assume_ that SSL is always available isn't true. Additionally, the "et cetera" part of the various attacks would also include things like verifying that the packages themselves haven't been modified on the remote server by someone who obtains FTP access and starts hacking away at the zip files. For Drupal, there's a 90 key rotation ceremony and regeneration of all the zip archives. We don't want to have any opportunity for anything odd happening when we deliver these artifacts.

PR #8655, opened 2020-02-25
Provides a post download event

heddn:
Start the conversation on resolving #8654

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -21,6 +21,7 @@ use Composer\Package\Comparer\Comparer;
 use Composer\Package\PackageInterface;
 use Composer\Package\Version\VersionParser;
 use Composer\Plugin\PluginEvents;
+use Composer\Plugin\PostFileDownloadEvent;
 use Composer\Plugin\PreFileDownloadEvent;
 use Composer\EventDispatcher\EventDispatcher;
 use Composer\Util\Filesystem;
@@ -140,7 +141,7 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
                     ->then($accept, $reject);
             }
 
-            return $result->then(function ($result) use ($fileName, $checksum, $url) {
+            return $result->then(function ($result) use ($fileName, $checksum, $url, $eventDispatcher) {
                 // in case of retry, the first call's Promise chain finally calls this twice at the end,
                 // once with $result being the returned $fileName from $accept, and then once for every
                 // failed request with a null result, which can be skipped.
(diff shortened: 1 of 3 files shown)


heddn:
I think (hope) the use case for Drupal wanting to hash the zip downloads using its security mechanism makes sense. Linking in #5935, which proposed this some time ago, but didn't provide a strong use case.

Seldaek:
Sounds reasonable to me, the question is whether this should be more specific than just file, and include the package info etc.. PRE_FILE_DOWNLOAD is triggered in other places like ComposerRepository, but I think the POST_* case really only makes sense for the package downloaders.

heddn:
I looked all the places that did a PRE_FILE_DOWNLOAD and (to me) it only made sense in package downloading. We could always add it in more places later if we decide it is needed.

heddn:
Rebased PR so it now has no conflicts. Any chance on this still landing in 2.0?

src/Composer/Downloader/FileDownloader.php
```diff
@@ -157,6 +158,11 @@ public function download(PackageInterface $package, $path, PackageInterface $pre
                     throw new \UnexpectedValueException('The checksum verification of the file failed (downloaded from '.$url['base'].')');
                 }
 
+                if ($eventDispatcher) {
+                    $postFileDownloadEvent = new PostFileDownloadEvent(PluginEvents::POST_FILE_DOWNLOAD, $fileName, $checksum, $url);
```
Seldaek, line 162: Sorry looking at the implementation in a little more detail.. I think $url here should just pass `$url['processed']` as that is what is being downloaded, and otherwise we are leaking implementation details here into the plugin API etc it just sounds like maintenance hell.

PostFileDownloadEvent docs should be updated to reflect this too.

src/Composer/Plugin/PluginEvents.php
```diff
@@ -49,6 +49,16 @@ class PluginEvents
      */
     const PRE_FILE_DOWNLOAD = 'pre-file-download';
 
+    /**
+     * The POST_FILE_DOWNLOAD event occurs after downloading a file
```
Seldaek, line 53:
```suggestion
     * The POST_FILE_DOWNLOAD event occurs after downloading a package dist file
```

As PRE_FILE_DOWNLOAD is actually triggered in other circumstances the distinction is important IMO

src/Composer/Plugin/PostFileDownloadEvent.php
```diff
@@ -0,0 +1,87 @@
...
+class PostFileDownloadEvent extends Event
+{
+
+    /**
+     * @var string
+     */
+    private $fileName;
+
+    /**
+     * @var string
```
Seldaek, line 32: This is technically string|null, might be good to clarify everywhere.

Seldaek:
Needs a few more changes but yes no worries this will absolutely make it in 2.0.

heddn:
Working on addressing the latest feedback.

heddn:
Feedback addressed.

Seldaek:
Thanks for the quick update.

PR #8662, opened 2020-03-01
Enhancement: Run jobs on GitHub Actions

localheinz:
This PR

* [x] switches the build process from Travis CI to GitHub Actions

💁‍♂ For reference, see

- https://github.com/localheinz/composer/pull/1
- https://github.com/localheinz/composer/pull/1/checks

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,77 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+name: "Continuous Integration"
+
+on:
+  - "pull_request"
+
+env:
+  COMPOSER_FLAGS: "--ansi --no-interaction --no-progress --no-suggest --prefer-dist --optimize-autoloader"
+  SYMFONY_PHPUNIT_VERSION: ""
+
+jobs:
+  tests:
+    name: "Tests"
+
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "5.3"
+          - "5.4"
+          - "5.5"
+          - "5.6"
+          - "7.0"
+          - "7.1"
+          - "7.2"
+          - "7.3"
+          - "7.4"
+
+        dependencies:
+          - "locked"
+          - "highest"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          ini-values: "memory_limit=-1"
+          php-version: "${{ matrix.php-version }}"
+
+      - name: "Determine composer cache directory"
+        id: "determine-composer-cache-directory"
+        run: "echo \"::set-output name=directory::$(composer config cache-dir)\""
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "${{ steps.determine-composer-cache-directory.outputs.directory }}"
+          key: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}-${{ hashFiles('**/composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}"
+
+      - name: "Install highest dependencies from composer.json using composer binary provided by system"
+        if: "matrix.dependencies == 'highest'"
+        run: "composer update ${{ env.COMPOSER_FLAGS }}"
+
+      - name: "Install dependencies from composer.lock using composer binary provided by system"
+        if: "matrix.dependencies == 'locked'"
+        run: "composer install ${{ env.COMPOSER_FLAGS }}"
+
+      - name: "Install dependencies from composer.lock using composer binary from source"
+        if: "matrix.dependencies == 'locked'"
+        run: "bin/composer install ${{ env.COMPOSER_FLAGS }}"
+
+      - name: "Prepare git environment"
+        run: "git config --global user.name composer && git config --global user.email composer@example.com"
+
+      - name: "Set SYMFONY_PHPUNIT_VERSION environment variable"
+        if: "matrix.php-version == '7.4'"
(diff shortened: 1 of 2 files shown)


localheinz:
As far as I can tell, this works fine.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: "Prepare git environment"
+        run: "git config --global user.name composer && git config --global user.email composer@example.com"
+
+      - name: "Set SYMFONY_PHPUNIT_VERSION environment variable"
+        if: "matrix.php-version == '7.4'"
+        run: "echo \"::set-env name=SYMFONY_PHPUNIT_VERSION::7.5\""
+
+      - name: "Run tests"
+        run: "vendor/bin/simple-phpunit"
```
staabm: what do you think about activating a phpunit problem matcher?
https://github.com/shivammathur/setup-php#problem-matchers

stof: Given that problem matchers are limited to adding 10 annotations per step, with a limit of 50 per job, I find them a bit overrated actually.

staabm: still its better then looking up errors in the logfiles (and saves time)

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "${{ steps.determine-composer-cache-directory.outputs.directory }}"
+          key: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}-${{ hashFiles('**/composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}"
+
+      - name: "Install highest dependencies from composer.json"
+        if: "matrix.php-version == '7.4'"
```
stof: shouldn't we still keep a job running on PHP 7.4 with our locked dependencies ?

.travis.yml
```diff
@@ -16,20 +16,7 @@ addons:
...
-    - php: 5.3
-      dist: precise
-    - php: 5.4
-    - php: 5.5
-    - php: 5.6
-    - php: 7.0
-    - php: 7.1
-    - php: 7.2
-    - php: 7.3
     - php: nightly
```
stof: keeping a single job, which is in `allow_failures` looks weird to me.

.travis.yml
```diff
@@ -16,20 +16,7 @@ addons:
...
 matrix:
   include:
-    - php: 5.3
-      dist: precise
-    - php: 5.4
-    - php: 5.5
-    - php: 5.6
-    - php: 7.0
-    - php: 7.1
-    - php: 7.2
```
stof: This PHP 7.2 job is the one uploading the composer.phar to the github release on tags. So this is now lost as the `deploy` section will never match anymore on Travis and you don't have the equivalent on github actions.

.github/workflows/continuous-integration.yml
stof: but as soon as you have more than 10 errors, you must go in the log files (and there is nothing telling you that there might have additional errors in the log file)

.travis.yml
shivammathur: @localheinz @stof we can have another workflow to deploy, similar to [this](https://github.com/staabm/annotate-pull-request-from-checkstyle/blob/master/.github/workflows/continuous-deployment.yml) for composer

.github/workflows/continuous-integration.yml
localheinz: We currently aren’t running such a job on Travis.

.travis.yml
localheinz: We can't run a job in GitHub Actions and allow it to fail at the some time.

We could use [`continue-on-error`](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error), but I don't think that running jobs that are allowed to fail makes sense at all.

.travis.yml
localheinz: Releases can be created with [`actions/create-release`](https://github.com/actions/create-release). Release assets can be uploaded using [`actions/upload-release-asset`](https://github.com/actions/upload-release-asset).

.github/workflows/continuous-integration.yml
Seldaek: That would probably make sense to have though if you don't mind adding it here.

Seldaek:
This seems to trigger two builds for every PR (one for push and one for pull_request), isn't this a little wasteful? Not sure what can be done about it..

stof:
@Seldaek when the push in the main repo, that's also what Travis does (building the branch, and building the PR which runs on the merge commit of that branch with master).
Both Travis and Github Actions allow to restrict push builds to pushes to some specific branches. But that is not used on Travis currently either.

Seldaek:
Ok no problem about that then. Adding a regular 7.4 build would be nice still.

Seldaek:
Merged in 1.10/master, last commit shows nothing from github tho, and https://github.com/composer/composer/actions also empty.. Sorry I'm probably dumb here but I can't find out how to enable this, only how to create a new file?!

Also #8773 should be backported here if possible, as a single build in the matrix.. not sure if we can avoid doing locked/latest builds for every php version too, that seems a little extreme.

staabm:
@Seldaek actions are activated in the repo config?

grafik

Seldaek:
Yeah I checked that and it's on.

Seldaek:
Ah it seems it builds on https://github.com/composer/composer/pull/8803 so maybe it's only enabled for PRs now?

staabm:
> Ah it seems it builds on #8803

yep

https://github.com/composer/composer/blob/1eaa8cd55fbcc6e5fbee10820456da3700f46593/.github/workflows/continuous-integration.yml#L6

fix incoming.

Seldaek:
Migrated all the travis/appveyor  stuff we still had on github actions now. Thanks for getting the ball rolling @localheinz

localheinz:
@Seldaek 

You are welcome!

Feel free to ping me any time when I can help!

Issue #8551, opened 2020-01-25
Allow to install/update only the composer.lock without installation


mvorisek:
This is a feature request. For dependency analysises like this: https://gist.github.com/mvorisek/91ffee6254aed784daff3f1a0994d11a it is more than needed to be able to save the `composer.lock` file only without installing the actual files.

Of course it can be done from PHP, but this should be available via the composer CLI for easy use.

Seldaek:
That'd be https://github.com/composer/composer/blob/7e6fe585bc8a1d63e15431ae234114e8fa20ac2a/src/Composer/Installer.php#L241-L244 (on the 2.0 branch) if anyone wants to tackle it.

PR #8669, opened 2020-03-04
Introduce --only-create-lock option for the Installer

PirxDanford:
I liked the idea requested in https://github.com/composer/composer/issues/8551, maybe it can be done like this.

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -218,6 +218,17 @@ class Installer
             $this->mockLocalRepositories($this->repositoryManager);
         }
 
+        if ($this->onlyCreateLock) {
+            $this->runScripts = false;
+            $this->executeOperations = true;
+            $this->writeLock = true;
+            $this->dumpAutoloader = false;
+            $this->mockLocalRepositories($this->repositoryManager);
+            if ($this->locker->isLocked()) {
+                throw new \RuntimeException("Can only create a lock file using --only-create-lock if it not already exists");
+            }
+        }
+
         if ($this->runScripts) {
             $devMode = (int) $this->devMode;
             putenv("COMPOSER_DEV_MODE=$devMode");
@@ -239,8 +250,8 @@ class Installer
 
         try {
             if ($this->update) {
-                // TODO introduce option to set doInstall to false (update lock file without vendor install)
-                $res = $this->doUpdate($localRepo, true);
+                // we do not want to install if we only want the lock
+                $res = $this->doUpdate($localRepo, !$this->onlyCreateLock);
             } else {
                 $res = $this->doInstall($localRepo);
             }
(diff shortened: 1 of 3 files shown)


src/Composer/Installer.php
```diff
@@ -218,6 +218,17 @@ public function run()
...
         }
 
+        if ($this->onlyCreateLock) {
+            $this->runScripts = false;
+            $this->executeOperations = true;
+            $this->writeLock = true;
+            $this->dumpAutoloader = false;
+            $this->mockLocalRepositories($this->repositoryManager);
+            if ($this->locker->isLocked()) {
+                throw new \RuntimeException("Can only create a lock file using --only-create-lock if it not already exists");
```
stof: why not allowing to run a `composer update --only-create-lock` ?

PirxDanford: Because we would have a composer.lock file which would not match the installed state.

stof: and that's fine. It is also what happens when you do a `git pull` and someone else committed an update.

PirxDanford: Hmm, so if this behaviour would be okay then:
- Rename the option to `--just-touch-lock`
- Remove the exception if file exists and allow update of the existing lock in install
- Implement the option also for update and require
?

Seldaek:
I'm thinking the name here might make more sense as `--no-install`, as essentially you are asking for an update (which is doing lock file update + install of lock file) but without the install step. Thoughts @naderman ?

naderman:
@Seldaek indeed I agree "--no-install" similar to "--no-update" on composer require makes sense.

Seldaek:
Also note that it should be added on the `update` command. Possibly I'd add it to `install` command as well but if it is used there I think we should error with a message saying update should be used to update the lock file.

PirxDanford:
Hi, so if I understand correctly you suggest to use `composer install --no-install` and `composer update --no-update`? If yes then I disagree about this, because we are actually doing something, if only with the lockfile. Also just negating the requested action in an option feels somewhat counterintuitive.

@Seldaek and @naderman 
How about a compromise? I suggest these options in the order of my preference:
1. `composer install --only-install-lock` and `composer update --only-update-lock`
1. Alternatively `composer install --only-lockfile` and `composer update --only-lockfile`
1. or even `composer install --no-full-install` and `composer update --no-full-update`

Seldaek:
I suggest `composer update --no-install`, `composer install --no-install` should error saying to use `composer update --no-install`. There is no `--no-update` anywhere, you just use the install command if you don't want an update.

update --no-install is like your suggested --only-update-lock, but it fits better with the naming of our other options IMO..

killua99:
It will support `composer require --no-install` as well?

Seldaek:
@killua99 there is already a --no-update on require to stop the update operation (which includes an install) from happening. But I guess we could add a --no-install to require and remove commands too for completeness.

killua99:
@Seldaek then we probably should update the documentation about the `--no-update` command? and add an alias as `--no-install` for semantics reason?

> --no-update: Disables the automatic update of the dependencies.

Add the text you did mention "which includes an install"

Seldaek:
Update always includes an install in composer, so it's kinda implicit, but yes maybe doesn't hurt to spell it out..

src/Composer/Installer.php
dereuromark: I agree that the same for update would be very handy.

dereuromark: Sry, I just saw that below `--no-install` in that command. All good then.

Issue #8642, opened 2020-02-24
More explicit support for authorization/bearer header as authentication method


ethanclevenger91:
GitLab repositories support `Authorization: Bearer xxx` headers as a method for authentication via `auth.json`. If a repository domain is listed under `gitlab-domains` in the config and `auth.json` contains an `http-basic` entry for that domain with the username `oauth2`, the password will be sent as the token.

GitLab isn't the only platform in the world using API tokens to authenticate. Right now you can add any functioning Composer repository with token-based authentication, GitLab-backed or not, to the `gitlab-domains` config and then configure a token as described and you'll authenticate just fine. Seems it would make sense to generalize this somehow. Perhaps dedicated token auth support in `auth.json` that isn't tied to the `gitlab-domains` config?

My global `config.json`:

```json
{
    "config": {
    	"gitlab-domains": [
    		"ethan.packagepeak.test"
    	]
    }
}
```

That domain is not a GitLab instance. It's a Laravel app using Airlock, FWIW.

My global `auth.json`:

```json
{
	"http-basic":
	{
		"ethan.packagepeak.test": {
			"username": "valid-api-token",
			"password": "oauth2"
		}
	}
}
```

And this works, but seems you could add proper support for Bearer token authentication rather than tie it just to GitLab. So no config required, and an `auth.json` like:

```json
{
	"bearer":
	{
		"ethan.packagepeak.test": "valid-api-token"
	}
}
```

Pretty sure this can currently be accomplished per-project when defining the repository with something like:

```json
{
    "repositories": {
        "package-peak": {
            "type": "composer",
            "url": "http://ethan.packagepeak.test",
            "options": {
                "http": {
                    "header": [
                        "Authorization: Bearer valid-api-token"
                    ]
                }
            }
        }
    }
}
```

But for whatever reason I couldn't get that method to work at all. Ultimately I'd rather not be committing API tokens anyway.

Output of `composer diagnose`:

```
Checking composer.json: WARNING
require.roots/wordpress : unbound version constraints (>=5.2.2) should be avoided
... // several more of unbound version warnings
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
[...]

PR #8671, opened 2020-03-04
Bearer support.

ethanclevenger91:
No documentation updates or additional tests yet. Would appreciate direction on tests.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -835,6 +835,8 @@ class RemoteFilesystem
                     $headers[] = 'Authorization: Bearer ' . $auth['password'];
                     $authenticationDisplayMessage = 'Using Bitbucket OAuth token authentication';
                 }
+            } elseif ($auth['password'] === 'bearer' ) {
+                $headers[] = 'Authorization: Bearer '.$auth['username'];
             } else {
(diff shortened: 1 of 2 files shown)


ethanclevenger91:
Oh, addresses #8642

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -835,6 +835,8 @@ protected function getOptionsForUrl($originUrl, $additionalOptions)
                     $headers[] = 'Authorization: Bearer ' . $auth['password'];
                     $authenticationDisplayMessage = 'Using Bitbucket OAuth token authentication';
                 }
+            } elseif ($auth['password'] === 'bearer' ) {
```
Seldaek: I think this should be above, possibly the first if() of the whole block, because otherwise if you somehow want to configure a GitHub/GitLab/BitBucket domain directly using bearer auth you won't be able to reach this block.

Seldaek: Fixed this myself as I'd like to include this in the 1.10 release and can't wait

Seldaek:
Added some missing stuff in https://github.com/composer/composer/commit/f964b8301837a49ee1a4cf77dbfb38d0c2c2a340 so it's ready to be released.

Issue #8645, opened 2020-02-24
composer check-platform-reqs fails because ext-mongo is missing, even if alcaeus/mongo-php-adapter provides ext-mongo


floverdevel:
My `composer.json`:

```json
{
    "name": "worker",
    "description": "Service Worker",
    "type": "project",
    "license": "proprietary",
    "minimum-stability": "stable",
    "config": {
        "optimize-autoloader": true
    },
    "require": {
        "php": "7.3.*",
        "ext-curl": "*",
        "ext-json": "*",
        "ext-xml": "*",
        "ext-mbstring": "*",
        "ext-apcu": "*",
        "alcaeus/mongo-php-adapter": "^1.1",
        "doctrine/mongodb": "^1.6",
        "doctrine/mongodb-odm": "^1.2"
    },
    "require-dev": {
        "phake/phake": "^3",
        "phpunit/phpunit": "^8",
        "squizlabs/php_codesniffer": "^3"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 1.9.3
PHP version: 7.3.12
PHP binary path: /usr/local/Cellar/php@7.3/7.3.12/bin/php
```

When I run this command:

```
composer check-platform-reqs
```

I get the following output:

```
composer-plugin-api  1.1.0                                                   success
ext-apcu             5.1.18                                                  success
ext-ctype            7.3.12                                                  success
ext-curl             7.3.12                                                  success
ext-dom              20031129                                                success
ext-filter           7.3.12                                                  success
ext-hash             7.3.12                                                  success
ext-iconv            7.3.12                                                  success
ext-json             1.7.0                                                   success
ext-libxml           7.3.12                                                  success
ext-mbstring         7.3.12                                                  success
ext-mongo            n/a       doctrine/mongodb requires ext-mongo (^1.6.7)  missin
ext-mongodb          1.7.2                                                   success
[...]

PR #8676, opened 2020-03-07
check-platform-reqs looks into packages that provides requirements

floverdevel:
Fix for #8645 
This is my first PR for this repo, I think i followed the guidelines correctly.
Sorry if missed something :)

### DIFF
--- a/src/Composer/Command/CheckPlatformReqsCommand.php
+++ b/src/Composer/Command/CheckPlatformReqsCommand.php
@@ -69,6 +69,12 @@ EOT
 
         ksort($requires);
 
+        $lock = $this->getComposer()->getLocker()->getLockData();
+        $providers = array_filter($lock['packages'], function ($package) {
+            return isset($package['provide']);
+        });
+
+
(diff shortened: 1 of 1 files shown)


Seldaek:
Hey, thanks a lot for the PR. Looking at it though I realized this could use some larger refactor for 2.0, which led to rewriting most of it anyway in https://github.com/composer/composer/commit/2445ef5e1f00d9b1c23c0a6982690fca72260ca2 - I did add you as co-author tho because really this PR was mostly fine in its own, it just led to a little bit more duplication than ideal.

floverdevel:
Hey i'm sorry i'm a bit late on this.
But realized that, If I understand correctly, the fix 2445ef5 is only for the 2.* version of composer.
Is there any chance to have a fix for the 1.* version?

Seldaek:
Yeah that is correct, and sorry but no I am focusing all time and effort on getting 2.0 out the door so not keen on adding non-critical fixes to 1.x anymore. You can already use the 2.0 pre-releases though, it is quite stable and much faster.

PR #8682, opened 2020-03-11
GitLab: properly handle token which has Guest only access

glaubinix:
This causes `Notice: Undefined index: default_branch at GitLabDriver.php:22` and never tries to fallback to the git driver.

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -427,6 +427,30 @@ class GitLabDriver extends VcsDriver
             if ($fetchingRepoData) {
                 $json = JsonFile::parseJson($res, $url);
 
+                // Accessing the API with a token with Guest (10) access will return
+                // more data than unauthenticated access but no default_branch data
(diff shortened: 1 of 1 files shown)


src/Composer/Repository/Vcs/GitLabDriver.php
```diff
@@ -427,6 +427,30 @@ protected function getContents($url, $fetchingRepoData = false)
...
+                        if ($permission && $permission['access_level'] > 10) {
+                            $moreThanGuestAccess = true;
+                        }
+                    }
+
+                    if (!$moreThanGuestAccess) {
+                        $this->io->writeError('<warning>GitLab token with Guest only access detected</warning>');
+                        $this->setupGitDriver($this->generateSshUrl());
+
+                        return $res;
```
Seldaek: Should it not instead do:

```suggestion

                        return $this->attemptCloneFallback(); 
```

?

glaubinix: Yes, that makes more sense.

Issue #8570, opened 2020-01-31
Issue with handling of require-dev in lock


Seldaek:
With this:

```json
{
    "repositories": [{
        "type": "package",
        "package": [
            {"name": "provider/requirer", "version": "1.0.0", "type": "metapackage", "require": {"b/b": "1.0.0"}},
            {"name": "b/b", "version": "1.0.0", "type": "metapackage", "provide": {"provided/pkg": "1.0.0"}}
        ]
    }],
    "require": {
        "provided/pkg": "1.0.0"
    },
    "require-dev": {
        "provider/requirer": "1.0.0"
    }
}
```

The first `composer update` succeeds, but the first problem manifests itself: the "packages-dev" in lock file is empty. `provided/requirer` should be in `packages-dev` though IMO, but the package `b/b` which provides `provided/pkg` should be in `packages`. This probably fails to be picked up correctly due to the provide indirection.

----

The second `composer update` however fails during the `extractDevPackages` resolution with:

> Root composer.json requires provided/pkg, it could not be found in any version, there may be a typo in the package name. 

$resultRepo contains `provider/requirer` and `b/b`, but the request only contains `provided/pkg`  which means that the provider is not selected explicitly and thus not loaded at all.

PR #8684, opened 2020-03-11
Handle dev extraction exit codes instead of completing broken lock with errors

naderman:
Addresses https://github.com/composer/composer/issues/8570

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -422,11 +422,13 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
     public function getProviders($packageName)
     {
         if (!$this->providersApiUrl) {
+            // TODO should this return the info based on getPackages in other cases?
             return array();
         }
 
         $result = $this->httpDownloader->get(str_replace('%package%', $packageName, $this->providersApiUrl), $this->options)->decodeJson();
 
+        // TODO filter packageName out here?
         return $result['providers'];
     }
 
--- a/src/Composer/Repository/RepositorySet.php
+++ b/src/Composer/Repository/RepositorySet.php
@@ -163,17 +163,31 @@ class RepositorySet
         return $candidates;
     }
 
-    public function getProviders($packageName)
+    public function getProvidersAndReplacers($packageName)
     {
+        $providers = array();
         foreach ($this->repositories as $repository) {
             if ($repository instanceof ComposerRepository) {
-                if ($providers = $repository->getProviders($packageName)) {
-                    return $providers;
+                if ($repoProviders = $repository->getProviders($packageName)) {
+                    $providers = array_merge($providers, $repoProviders);
+                }
+            } else {
+                foreach ($repository->getPackages() as $candidate) {
+                    foreach (array_merge($candidate->getProvides(), $candidate->getReplaces()) as $link) {
+                        if ($packageName === $link->getTarget()) {
+                            $providers[] = array(
+                                'name' => $candidate->getName(),
+                                'description' => $candidate->getDescription(),
+                                'type' => $candidate->getType(),
+                            );
+                            continue 2;
+                        }
+                    }
                 }
             }
         }
 
-        return array();
+        return $providers;
     }
 
     public function isPackageAcceptable($names, $stability)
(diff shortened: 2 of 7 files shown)


src/Composer/Repository/RepositorySet.php
```diff
@@ -163,17 +163,31 @@ public function findPackages($name, ConstraintInterface $constraint = null, $fla
...
 
-    public function getProviders($packageName)
+    public function getProvidersAndReplacers($packageName)
     {
+        $providers = array();
         foreach ($this->repositories as $repository) {
             if ($repository instanceof ComposerRepository) {
-                if ($providers = $repository->getProviders($packageName)) {
-                    return $providers;
+                if ($repoProviders = $repository->getProviders($packageName)) {
```
Seldaek: Might be good to rename this method too? And should it exclude the package having name==$packageName? Because right now the API does return it. https://packagist.org/providers/monolog/monolog.json

stof:
wouldn't this cause issues when a library has a requirement on a virtual package (`php-http/client-implementation` for instance) because it does not care about the implementation being used ? The way to work on them for the contributors is to have a dev requirement providing such an implementation.

src/Composer/Repository/RepositorySet.php
naderman: @Seldaek yeah I think that both makes sense.

naderman:
@stof Interesting, that's a use case I had not considered. I did not think this was a widely spread problem in the first place, and was just trying to correct the error message to be clearer, not change behavior.

In general we have the rule that only directly required packages will be installed. So if you want to pick an implementation you'd need to require it. If we want to support this use case, we need to change pool builder behavior during dev dependency extraction entirely. @Seldaek ?

naderman:
@stof So I now added a way to construct the pool with all packages for dev extraction (all packages being only the solution to the first solver run, so not many packages anyway). This will allow your use case to continue to work, since all providers will now be used automatically if available because of a require-dev requirement.

That said, there are a few questions left in this PR regarding the getProviders calls for error messages.

stof:
Well, that use case exists in all packages depending on a virtual package. that's the only way to make virtual packages useful (it makes no sense to put a specific implementation in your `require` section alongside the virtual package).

naderman:
@stof Yes I did understand that this matters ;-)

stof:
Also, I suggest adding a test (maybe even adding it in 1.x) covering that use case.

naderman:
@stof there is a test added in this PR, I just haven't updated it, so right now it's actually failing ;-)

src/Composer/Repository/ComposerRepository.php
```diff
@@ -422,11 +422,13 @@ public function search($query, $mode = 0, $type = null)
     public function getProviders($packageName)
     {
         if (!$this->providersApiUrl) {
+            // TODO should this return the info based on getPackages in other cases?
             return array();
         }
 
         $result = $this->httpDownloader->get(str_replace('%package%', $packageName, $this->providersApiUrl), $this->options)->decodeJson();
 
+        // TODO filter packageName out here?
```
Seldaek: I'd say probably yes.

src/Composer/Repository/ComposerRepository.php
```diff
@@ -422,11 +422,13 @@ public function search($query, $mode = 0, $type = null)
     public function getProviders($packageName)
     {
         if (!$this->providersApiUrl) {
+            // TODO should this return the info based on getPackages in other cases?
```
Seldaek: There are many other cases to consider there, but yeah I'll try to complete the implementation once this is merged.

src/Composer/Repository/RepositorySet.php
```diff
@@ -186,10 +200,8 @@ public function isPackageAcceptable($names, $stability)
      *
      * @return Pool
      */
-    public function createPool(Request $request, EventDispatcher $eventDispatcher = null)
+    public function createPool(Request $request, EventDispatcher $eventDispatcher = null, $allPackages = false)
```
Seldaek: A new method would probably be cleaner than a new arg here, as it anyway is a completely different code path.

tests/Composer/Test/Fixtures/installer/provider-dev-require-can-satisfy-require.test
```diff
@@ -0,0 +1,52 @@
+--TEST--
+Test that an appropriate error is thrown if a requirement is only satisfied by a package provided by a dependency of a dev requirement.
```
Seldaek: That's outdated

PR #8686, opened 2020-03-12
Fix PHP 8 deprecated warnings

matbech:
The PR fixes this PHP 8 deprecated warning:
PHP Deprecated:  Required parameter $ambiguousClasses follows optional parameter $blacklist in phar://.../composer.phar/src/Composer/Autoload/AutoloadGenerator.php on line 339


Seldaek:
Thanks

PR #8690, opened 2020-03-13
Error reporting improvements

Seldaek:
Fixes the dev-master alias issue as well as the other lock-related problem we found yesterday.

### DIFF
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -15,6 +15,7 @@ namespace Composer\DependencyResolver;
 use Composer\Package\CompletePackage;
 use Composer\Package\Link;
 use Composer\Package\PackageInterface;
+use Composer\Package\AliasPackage;
 use Composer\Repository\RepositorySet;
 
 /**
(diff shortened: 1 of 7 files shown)


src/Composer/DependencyResolver/Rule.php
```diff
@@ -188,8 +189,8 @@ public function getPrettyString(RepositorySet $repositorySet, Request $request,
 
             case self::RULE_PACKAGE_OBSOLETES:
```
naderman, line 190: this makes a change in obsoletes which one master no longer exists, double check if the same call needs to now be added somewhere in the RULE_PACKAGE_SAME_NAME section

src/Composer/DependencyResolver/Rule.php
```diff
@@ -279,4 +280,13 @@ private function getReplacedNames(PackageInterface $package)
 
         return $names;
     }
+
+    private function deduplicateMasterAlias(PackageInterface $package)
+    {
+        if ($package instanceof AliasPackage && $package->getPrettyVersion() === '9999999-dev') {
```
staabm, line 286: should this magic `'9999999-dev'` value be hidden behind a constant? this could also improve semantics

Seldaek: Yes sure could be if you want to PR.

staabm: here we go https://github.com/composer/composer/pull/8742

PR #8693, opened 2020-03-13
Add --dry-run to composer require

### LINKED ISSUE #8692: composer require --dry-run
It would be useful for me if `composer require` had a `--dry-run` option. I'd like to check if a package can be added to the project without conflicts but without actually adding it.

enumag:
Closes #8692


Seldaek:
This is not a complete implementation. It does a dry run of the update, but still leaves the composer.json file modified at the end.

enumag:
@Seldaek I see. Is this better?

Seldaek:
Yes that looks much better, I guess safest would be to modify the root package in memory to avoid touching the file system entirely.. but not sure if it's really worth it if your approach works.

enumag:
@Seldaek Is there an easy way to do that? I don't understand the code that well. Personally I don't really care, reverting the file afterwards is good enough for me.

Seldaek:
It's probably not worth changing this tbh, let's go with this for now. It'll have to go in 2.0 though as I don't plan any further 1.x minor release.

enumag:
I see. Is there any ETA for 2.0?

Seldaek:
I wish but no it's not clear.. Trying hard to get it done though atm

PR #8697, opened 2020-03-17
`*` is not a valid version, changed to `1`

strayobject:
Minor documentation change.

### DIFF
--- a/doc/01-basic-usage.md
+++ b/doc/01-basic-usage.md
@@ -198,7 +198,7 @@ includes PHP itself, PHP extensions and some system libraries.
 
 * `ext-<name>` allows you to require PHP extensions (includes core
   extensions). Versioning can be quite inconsistent here, so it's often
-  a good idea to set the constraint to `*`.  An example of an extension
+  a good idea to set the constraint to `1`.  An example of an extension
   package name is `ext-gd`.
 
 * `lib-<name>` allows constraints to be made on versions of libraries used by


strayobject:
I might be wrong about this, if so a proper example would be nice.

For me, this does not work
```
"platform": {
  "ext-mbstring": "*"
}
```

And this does:
```
"platform": {
  "ext-mbstring": "1"
}
```

Seldaek:
As a requirement you'd need to use `*` to match any possible version, but when declaring which version the extension has in platform config then you need a version and not a constraint, so indeed there `1` or something would make more sense, but that's not really what that piece of doc had in mind. Maybe should be a footnote if anything?

dereuromark:
I think the `*` should be kept as general and agnostic advice over all extensions.
The main focus of these is to check existence of the extension, the exact version is often a detail less relevant here.

stof:
When requiring an extension, setting the constraint to `1` is definitely a bad idea, as that would allow only 1.0.0 of the extension (and so will almost always fail).

strayobject:
I have figured out that my change suggestion is in the wrong section.
I wrote the whole essay about how we are leading newcomers on the wrong path, while my use case and change pertains to
https://github.com/strayobject/composer/blob/patch-1/doc/06-config.md#platform

I wonder if it is just me who sees a mention of `platform` in the basic usage and haphazardly jumps to conclusion that this is the docs for the config section :man_facepalming: 

If it's not just me then maybe we could add a link to the config's `platform` section in the docs, otherwise, sorry I've wasted your time.

PR #8702, opened 2020-03-19
Fixed up tests

GrahamCampbell:


tests/Composer/Test/Repository/PathRepositoryTest.php
```diff
@@ -85,7 +85,7 @@ public function testLoadPackageFromFileSystemWithWildcard()
         $packages = $repository->getPackages();
         $names = array();
 
-        $this->assertEquals(2, $repository->count() >= 2);
+        $this->assertGreaterThanOrEqual(2, $repository->count());
```
GrahamCampbell: This is the important change. Because this was passing because `2` is the same as `true = $repository->count() >= 2`, weakly...

GrahamCampbell: NB This mistake was introduced in https://github.com/composer/composer/commit/7a05286f5527ec11bc23fc23a6b91c6a8d4cfcef.

Seldaek: Ah yeah I meant to switch to assertTrue, but this is better anyway thanks.

Issue #5438, opened 2016-06-15
Update whitelisting does not correctly whitelist dependencies of new packages to be installed


naderman:
composer.json.1:

``` json
{
    "repositories": [
        {
            "packagist": false
        },
        {
            "type": "package",
            "package": {
                "name": "a/a",
                "version": "1.0.0",
                "type": "metapackage",
                "require": {
                    "a/b": "*"
                }
            }
        },
        {
            "type": "package",
            "package": {
                "name": "a/b",
                "version": "1.0.0",
                "type": "metapackage"
            }
        },
        {
            "type": "package",
            "package": {
                "name": "a/c",
                "version": "1.0.0",
                "type": "metapackage",
                "require": {
                    "a/b": "2.0.0"
                }
            }
        }
    ],
    "require": {
        "a/a": "1.0.0"
    }
}
```

composer.json.2:

``` json
{
    "repositories": [
        {
            "packagist": false
        },
        {
            "type": "package",
            "package": {
                "name": "a/a",
                "version": "1.0.0",
                "type": "metapackage",
                "require": {
                    "a/b": "*"
                }
            }
        },
        {
            "type": "package",
            "package": {
                "name": "a/b",
                "version": "1.0.0",
                "type": "metapackage"
            }
        },
        {
            "type": "package",
            "package": {
                "name": "a/b",
                "version": "2.0.0",
                "type": "metapackage"
            }
        },
        {
            "type": "package",
            "package": {
                "name": "a/c",
                "version": "1.0.0",
                "type": "metapackage",
                "require": {
                    "a/b": "2.0.0"
                }
            }
        }
    ],
    "require": {
        "a/a": "1.0.0",
        "a/c": "1.0.0"
    }
}
```

When I run this command:

```
COMPOSER=composer.json.1 composer update
COMPOSER=composer.json.2 composer update a/c --with-dependencies --dry-run
```

I get this output:

```
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Can only install one of: a/b[2.0.0, 1.0.0].
    - Can only install one of: a/b[2.0.0, 1.0.0].
    - a/c 1.0.0 requires a/b 2.0.0 -> satisfiable by a/b[2.0.0].
    - Installation request for a/c 1.0.0 -> satisfiable by a/c[1.0.0].
    - Installation request for a/b (installed at 1.0.0) -> satisfiable by a/b[1.0.0].

```

[...]

Issue #8355, opened 2019-10-02
Locked indirect dependency breaks installing additional packages


andig:
No `composer.json`- start with empty directory. Output of `composer diagnose`:

```
❯ composer diagnose
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 1.9.0
PHP version: 7.3.10
PHP binary path: /usr/local/Cellar/php/7.3.10/bin/php
```

When I run this command:

```
❯ composer require symfony/symfony php-pm/httpkernel-adapter
```

I get the following output:

```
Using version ^4.3 for symfony/symfony
Using version ^2.0 for php-pm/httpkernel-adapter
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 40 installs, 0 updates, 0 removals
  - Installing psr/link (1.0.0): Downloading (100%)         
  - Installing fig/link-util (1.0.0): Downloading (100%)         
  - Installing ralouphie/getallheaders (3.0.3): Loading from cache
  - Installing psr/http-message (1.0.1): Loading from cache
  - Installing guzzlehttp/psr7 (1.6.1): Loading from cache
  - Installing symfony/polyfill-ctype (v1.12.0): Loading from cache
  - Installing symfony/polyfill-mbstring (v1.12.0): Loading from cache
  - Installing twig/twig (v2.11.3): Downloading (100%)         
  - Installing symfony/polyfill-php73 (v1.12.0): Loading from cache
  - Installing symfony/polyfill-php72 (v1.12.0): Loading from cache
  - Installing symfony/polyfill-intl-idn (v1.12.0): Loading from cache
  - Installing symfony/symfony (v4.3.4): Downloading (100%)         
  - Installing symfony/polyfill-intl-icu (v1.12.0): Downloading (100%)         
[...]

PR #8717, opened 2020-03-27
Move processing of partial update argument list into the pool builder

naderman:
This removes a lot of now unnecessary code from the installer. If you use either of the options to update with transitive dependencies, then packages get marked for update in the pool builder while it loads package data and finds dependencies.

This is implemented as a reversal of the first decision to "fix" the package at the top of the pool vuilder. This should not incur too much of a performance penalty as the amount of packages this should impact should always be lower than the amount of packages fixed. At worst this should undo the performance benefit we gain from marking all these packages as fixed and not loading remote data for them.

As a consequence we can now automatically mark packages for update which are dependencies of either updated versions which are not locked yet, or of new packages, e.g. during composer require.

Replace is now also treated as if a transitive dependency was found, this allows requiring a new package to automatically remove one that it replaces.

Fixes https://github.com/composer/composer/issues/5438 and https://github.com/composer/composer/issues/8355


src/Composer/Command/RequireCommand.php
```diff
@@ -248,6 +249,13 @@ private function doUpdate(InputInterface $input, OutputInterface $output, IOInte
         $authoritative = $input->getOption('classmap-authoritative') || $composer->getConfig()->get('classmap-authoritative');
         $apcu = $input->getOption('apcu-autoloader') || $composer->getConfig()->get('apcu-autoloader');
 
+        $updateAllowTransitiveDependencies = Request::UPDATE_ONLY_LISTED;
+        if ($input->getOption('update-with-all-dependencies')) {
+            $updateAllowTransitiveDependencies = Request::UPDATE_TRANSITIVE_ROOT_DEPENDENCIES;
+        } elseif ($input->getOption('update-with-dependencies')) {
+            $updateAllowTransitiveDependencies = Request::UPDATE_TRANSITIVE_DEPENDENCIES;
```
Seldaek: Just thinking of naming here, TRANSITIVE_ROOT_DEPENDENCIES should ideally be called TRANSITIVE_DEPENDENCIES, and TRANSITIVE_DEPENDENCIES could then be.. TRANSITIVE_NON_FIXED_DEPENDENCIES or smth? Not sure about names, but what is sure is this should be documented clearly on the const definitions on Request, because the naming alone isn't clear enough I find.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -217,6 +239,16 @@ private function loadPackage(Request $request, PackageInterface $package)
...
                 $loadNames[$require] = null;
+            // if this is a partial update with transitive dependencies we need to unfix the package we now know is a
+            // dependency of another package which we are trying to update, and then attempt to load it again
+            } elseif ($propagateUpdate && $request->getUpdateAllowTransitiveDependencies() && isset($this->skippedLoad[$require])) {
+                if ($request->getUpdateAllowTransitiveRootDependencies() || !$this->isRootRequire($request, $require)) {
+                    $this->unfixPackage($request, $require);
+                    $loadNames[$require] = null;
+                } elseif (!$request->getUpdateAllowTransitiveRootDependencies() && $this->isRootRequire($request, $require) && !isset($this->updateAllowWarned[$require]) && $this->io) {
+                    $this->updateAllowWarned[$require] = true;
+                    $this->io->writeError('<warning>Dependency "'.$require.'" is also a root requirement. Package has not been listed as an update argument, so keeping locked at old version. Use --with-all-dependencies to include root dependencies.</warning>');
```
Seldaek: $this->io is nullable so all usages should probably be wrapped in `if`s.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -177,7 +199,7 @@ public function buildPool(array $repositories, Request $request)
         return $pool;
     }
 
-    private function loadPackage(Request $request, PackageInterface $package)
+    private function loadPackage(Request $request, PackageInterface $package, $propagateUpdate = true)
     {
         $index = count($this->packages);
```
Seldaek: I think this breaks due to the unset() in unfixPackage below? It'll start potentially reusing indices.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -233,7 +265,97 @@ private function loadPackage(Request $request, PackageInterface $package)
...
+    }
+
+    /**
+     * Reverts the decision to use a fixed package from lock file if a partial update with transitive dependencies
+     * found that this package actually needs to be updated
+     */
+    private function unfixPackage(Request $request, $name)
+    {
+        // remove locked package by this name which was already initialized
+        foreach ($this->packages as $i => $loadedPackage) {
```
Seldaek: Wondering if it wouldn't be more efficient to find the package with $name in the lockedRepo first, then doing an in_array() call to see if it's been loaded? Otherwise it's getName() on every package which might be a lot of them. Probably not that important though as partial updates shouldn't have sooo many loaded. Just thinking out loud :p

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -36,24 +37,45 @@ class PoolBuilder
...
+            $this->updateAllowList = $request->getUpdateAllowList();
+            $this->warnAboutNonMatchingUpdateAllowList($request);
+
+            foreach ($request->getLockedRepository()->getPackages() as $lockedPackage) {
+                if (!$this->isUpdateAllowed($lockedPackage)) {
+                    $request->fixPackage($lockedPackage);
+                    // remember which packages we skipped loading remote content for in this partial update
+                    $this->skippedLoad[$lockedPackage->getName()] = true;
+                    foreach ($lockedPackage->getReplaces() as $link) {
+                        $this->skippedLoad[$link->getTarget()] = true;
```
Seldaek: Should unfixPackage also unset the skipped packages due to replace? Not entirely sure right now what the implications are.

src/Composer/DependencyResolver/PoolBuilder.php
Seldaek: Another point here, is that --with-all-dependencies on require cmd is --update-with-all-dependencies, maybe we should alias that so the recommendation always works.

naderman: Check for nullable is already in the elseif, too confusing? And agree on standardizing/aliasing the option

src/Composer/DependencyResolver/PoolBuilder.php
naderman: Yes it should, otherwise a transitive dep which is fulfilled by a replaced package would not get updated if you do --with-dependencies, may even want to add it for provide as well?

src/Composer/DependencyResolver/PoolBuilder.php
naderman: Indeed, thanks, had a feeling there was something still of with aliases, will need to come up with a test for this.

src/Composer/DependencyResolver/PoolBuilder.php
Seldaek: Not too confusing, just missed it all alone at the end there :)

src/Composer/Command/RequireCommand.php
naderman: UPDATE_TRANSITIVE_DEPENDENCIES UPDATE_TRANSITIVE_DEPENDENCIES_NO_ROOT_REQUIRE

would probably be clearest, but indeed, comments would help.

src/Composer/DependencyResolver/PoolBuilder.php
Seldaek: Yeah I guess provide makes sense too.

src/Composer/DependencyResolver/PoolBuilder.php
staabm: > this->io is nullable so all usages should probably be wrapped in `if`

Couldnt we use `Composer\IO\NullIO` instead of spreading `if`s?

naderman: I suppose, this is mostly about tests which may use this class without an io instance

src/Composer/DependencyResolver/PoolBuilder.php
naderman: Unfortunately this isn't directly possible due to root aliases, but I suppose I could assume that they always immediately follow in the packages array and look for them there, not sure that's really worth it?

naderman: Actually nevermind, since we have the alias map, I can fix this too.

src/Composer/DependencyResolver/PoolBuilder.php
naderman: Funnily enough this actually wasn't a bug. The index is not used for assignment to the packages array, so that was still fine. The only place it's used is for the alias look up table, so you can go from base package to all the aliases for that package which in turn reference the indexes of that alias package. That is only used to delete packages which do not match nameConstraints which are currently reset to null for all unfixed package names, so there was no way this could lead to a bug. So I failed to write a test case that shows this ...

src/Composer/DependencyResolver/PoolBuilder.php
naderman: Realized the implementation of this was broken. Added a test now which checks this works right. This doesn't cover basic replace, that's dealt with further down, if we find a package which gets loaded for an update which replaces another package which was originally skipped, we unfix it too. So that already allows you to install a new package which replaces an existing transitive dep.

However there is another "mutual replace" case. If I have a package foo/foo replacing mutual/mutual in the lock file from a former root require and i removed it from lock file and added bar/bar which also replaces mutual/mutual, and then run a partial update for bar/bar with dependencies, this should remove foo/foo rather than keeping it locked and preventing installation of bar/bar. The same is not true for provide since there would not be a conflict, so it's cleaner to install both if no update of the package was requested.

naderman:
Also fixed the lock file generation to only lock composer.json aliases which are actually in use in the lock file. This prevents a partial update from adding aliases to the lock file for packages which were not updated and did not previously have the alias. Currently the alias wouldn't have been used in resolution, it would have only gotten written to the lock file, resulting in the next partial update using the alias as input, which could potentially have lead to hard to debug issues.

PR #8718, opened 2020-03-29
Fix: Return early

localheinz:
This PR

* [x] returns early

### DIFF
--- a/src/Composer/Package/Version/VersionGuesser.php
+++ b/src/Composer/Package/Version/VersionGuesser.php
@@ -63,26 +63,28 @@ class VersionGuesser
      */
     public function guessVersion(array $packageConfig, $path)
     {
-        if (function_exists('proc_open')) {
-            $versionData = $this->guessGitVersion($packageConfig, $path);
-            if (null !== $versionData && null !== $versionData['version']) {
-                return $this->postprocess($versionData);
-            }
+        if (!function_exists('proc_open')) {
+            return;
+        }
 
-            $versionData = $this->guessHgVersion($packageConfig, $path);
-            if (null !== $versionData && null !== $versionData['version']) {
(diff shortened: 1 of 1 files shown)


src/Composer/Package/Version/VersionGuesser.php
```diff
@@ -63,26 +63,28 @@ public function __construct(Config $config, ProcessExecutor $process, SemverVers
      */
     public function guessVersion(array $packageConfig, $path)
     {
-        if (function_exists('proc_open')) {
-            $versionData = $this->guessGitVersion($packageConfig, $path);
-            if (null !== $versionData && null !== $versionData['version']) {
-                return $this->postprocess($versionData);
-            }
+        if (!function_exists('proc_open')) {
+            return;
```
localheinz: Should we explicitly return `null` here?

```suggestion
            return null;
```

src/Composer/Package/Version/VersionGuesser.php
```diff
@@ -63,26 +63,28 @@ public function __construct(Config $config, ProcessExecutor $process, SemverVers
...
-            }
+        $versionData = $this->guessFossilVersion($packageConfig, $path);
+        if (null !== $versionData && null !== $versionData['version']) {
+            return $this->postprocess($versionData);
+        }
+
+        $versionData = $this->guessSvnVersion($packageConfig, $path);
+        if (null !== $versionData && null !== $versionData['version']) {
+            return $this->postprocess($versionData);
         }
```
localheinz: Should we explicitly return `null` here?

```suggestion
        }
        
        return null;
```

Seldaek:
Thanks

localheinz:
Thank you, @Seldaek!

Issue #8320, opened 2019-09-06
Unable to override composer/installers due to plugin loading order.


roygoldman:
My `composer.json`:

```json
{
    "name": "example/drupal-project",
    "description": "Composer installers error demo",
    "type": "project",
    "minimum-stability": "dev",
    "prefer-stable": true,
    "license": "GPL-2.0-only",
    "repositories": [
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    ],
    "require": {
        "roygoldman/drupal-project-installers": "^8.0",
        "roygoldman/composer-installers-discovery": "^1.0",
        "roygoldman/composer-where": "^1.0",
        "drupal/core": "^8.7",
        "drupal/webform": "^5.0",
        "composer/installers": "^1.2.0"
    }
}
```

Output of `composer diagnose`:
```
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (1.8.0 => 1.9.0)
Composer version: 1.8.0
PHP version: 7.3.7
PHP binary path: /usr/bin/php7.3
```

When I run this command:

```
composer install && composer where drupal/webform
```

I get the following output:

```
drupal/webform(type:drupal-module managed by Composer\Installers\Installer) is installed at /home/reg98/drupal_sites/wifi/modules/webform
```
Due to the sorting of composer packages during plugin loading, the composer plugins get loaded in reverse alphabetical order. As a result, the `roygoldman/drupal-project-installers` which leverages `roygoldman/composer-installers-discovery` to determine the installation path of the drupal modules, so that they don't have to be tracked in each project's composer.json breaks. This is the result of the installer being part of  `roygoldman/composer-installers-discovery`  which gets loaded prior to `composer/installers`, at which point the generic installer takes priority over the custom installer. The sorting in question was added in relation to #8065. A similar race condition is what is happening here. 

[...]

PR #8723, opened 2020-03-31
Fix package sorting

helhum:
PackageSorter weighs the importance of a package
by counting how many times it is required by other packages.
This works by calculating the weight for each package name.

However currently the package index of the package array
is currently passed the weigh function, which basically
disables package sorting.

The reason for that is, that a package repository previously
returned the package list as associative array with  package name as keys,
but currently just as an array with integer keys.

Therefore we must extract the package name from the package
before passing it to the weigh function.

Related to #2642, #8065, 9d8db4e14858537e440f1a9523b6ea952013c184

### DIFF
--- a/src/Composer/Util/PackageSorter.php
+++ b/src/Composer/Util/PackageSorter.php
@@ -55,9 +55,9 @@ class PackageSorter
 
         $weightList = array();
 
-        foreach ($packages as $name => $package) {
-            $weight = $computeImportance($name);
-            $weightList[$name] = $weight;
+        foreach ($packages as $index => $package) {
+            $weight = $computeImportance($package->getName());
+            $weightList[$index] = $weight;
         }
 
         $stable_sort = function (&$array) {
@@ -84,8 +84,8 @@ class PackageSorter
 
         $sortedPackages = array();
 
-        foreach (array_keys($weightList) as $name) {
-            $sortedPackages[] = $packages[$name];
+        foreach (array_keys($weightList) as $index) {
+            $sortedPackages[] = $packages[$index];
         }
         return $sortedPackages;
     }
--- a/tests/Composer/Test/Util/PackageSorterTest.php
+++ b/tests/Composer/Test/Util/PackageSorterTest.php
@@ -1,0 +1,129 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Test\Util;
+
+use Composer\Package\Link;
+use Composer\Package\Package;
+use Composer\Test\TestCase;
+use Composer\Util\PackageSorter;
+
+class PackageSorterTest extends TestCase
+{
+    public function testSortingDoesNothingWithNoDependencies()
+    {
+        $packages[] = $this->createPackage('foo/bar1', array());
+        $packages[] = $this->createPackage('foo/bar2', array());
+        $packages[] = $this->createPackage('foo/bar3', array());
+        $packages[] = $this->createPackage('foo/bar4', array());
+
+        $sortedPackages = PackageSorter::sortPackages($packages);
+
+        self::assertSame($packages, $sortedPackages);
+    }
+
+    public function sortingOrdersDependenciesHigherThanPackageDataProvider()
+    {
+        return array(
+            'one package is dep' => array(
+                array(
+                    $this->createPackage('foo/bar1', array('foo/bar4')),
+                    $this->createPackage('foo/bar2', array('foo/bar4')),
+                    $this->createPackage('foo/bar3', array('foo/bar4')),
+                    $this->createPackage('foo/bar4', array()),
+                ),
+                array(
+                    'foo/bar4',
+                    'foo/bar1',
+                    'foo/bar2',
+                    'foo/bar3',
+                ),
+            ),
+            'one package has more deps' => array(
+                array(
+                    $this->createPackage('foo/bar1', array('foo/bar2')),
+                    $this->createPackage('foo/bar2', array('foo/bar4')),
+                    $this->createPackage('foo/bar3', array('foo/bar4')),
+                    $this->createPackage('foo/bar4', array()),
+                ),
+                array(
+                    'foo/bar4',
+                    'foo/bar2',
+                    'foo/bar1',
+                    'foo/bar3',
+                ),
+            ),
+            'package is required by many, but requires one other' => array(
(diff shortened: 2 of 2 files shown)


helhum:
The bug was introduced in [this commit](266a41e0464c8ccf23949e3794189dc4ccf3caba) during a refactoring of the PluginManager. `$packages = $repo->getPackages();` returns an array of packages, not an associative array with package name as keys. But the `$generator->sortPackages()` method required exactly that.

Seldaek:
Could you have a look what the impact is on https://github.com/composer/composer/issues/8320 ?

Also it'd be great to have more tests which show the fix. I feel like every six months someone has a problem with the way things are sorted, messes around with the sort function, then everyone else is unhappy.

Seldaek:
Also cc @danepowell and @roygoldman please have a look? :)

danepowell:
The regression makes sense to me as described, although I'm a little surprised this wasn't discovered previously. But _maybe_ it would lead to the kind of stochastic behavior that explains why the current sorting works for some people (i.e. me :smile: ) and not others (i.e. #8320)?

Anyway, I absolutely agree we need an automated unit test for PackageSorter that clearly describes the expected sort behavior. Specifically, it needs to demonstrate that if a package such as `acquia/blt` depends on another package such as `composer/installers`, `composer/installers` _always_ gets installed prior to `acquia/blt`.

While this seems to be the case currently, I worry about this comment:

> PackageSorter weighs the importance of a package by counting how many times it is required by other packages

If that's true, it implies that if `acquia/blt` depends on `composer/installers` but `acquia/blt` is required by a ton of other packages for whatever reason, `acquia/blt` might be loaded first. That would be really bad. So this should also be part of the test case (i.e. ensuring that if `acquia/blt` depends on `composer/installers`, `composer/installers` gets loaded prior to `acquia/blt` even if 1000 other packages depend on `acquia/blt`)

helhum:
> Anyway, I absolutely agree we need an automated unit test for PackageSorter

I agree as well.

> That would be really bad. So this should also be part of the test case: i.e. ensuring that if acquia/blt depends on composer/installers, composer/installers gets loaded prior to acquia/blt even if 1000 other packages depend on acquia/blt

If this really is the expectation, then the dependency sorting must be done differently, not by counting the references.

> I feel like every six months someone has a problem with the way things are sorted, messes around with the sort function, then everyone else is unhappy.

Yeah. When I was researching the history of this bug, I indeed found a ton of changes in this area. To help out a bit more here (write tests), I need to dive deeper into the use cases people needed to have covered.

danepowell:
Yeah, I think the dependency sorter needs to change to ensure that packages with dependencies get loaded only after those dependencies are loaded. Otherwise I don't know how we prevent issues like #8085 

The algorithm to do this wouldn't be so hard to implement, I don't think. Let me know if you think this is a good idea and/or want help. For most users, it's possible that packages would more or less be sorted the same way, since I expect the number of dependencies roughly correlates with the actual dependency tree.

I know this PR is more strictly a bug fix so I don't necessarily want to hold it up, but I do want to ensure it's not going to cause a regression like #8320

Seldaek:
Yeah I'd rather hold it up and get a correct solution into 2.0 if it comes to that, than switching the order around again in a point release.

Another thing to consider with weighing, is that there can be circular dependencies, in that case it can be tricky to determine an order, and a general "which is most required" can be a good result perhaps.

This is also used to sort autoload rules IIRC, which has slightly different effect than with installs where mostly plugins matter. For autoloading it's more about sorting "files" autoload rules. Might be outdated info tho I haven't checked.. Just brain dumping here in case that helps.

helhum:
> Another thing to consider with weighing, is that there can be circular dependencies, in that case it can be tricky to determine an order, and a general "which is most required" can be a good result perhaps.

Very good point. If we're going to change this for Composer 2.0 only anyway, how about introducing new directives for plugin execution order and autoload build order (or maybe only one of them)? I'm not sure though if this helps to clear up things or increases complexity even more.

roygoldman:
To clarify a few bits of my understanding, my issue in #8320 is focused more on the plugin loading order, and reordering the packages likely wouldn't solve those issues. Before the change in #2644, I was able to effect the plugin loading order by adding my project earlier in the root dependency list. I don't think fixing the sorting will cause a significant effect on #8320, but I'll try and test it out today.

> That would be really bad. So this should also be part of the test case: i.e. ensuring that if acquia/blt depends on composer/installers, composer/installers gets loaded prior to acquia/blt even if 1000 other packages depend on acquia/blt

Regarding sort order, [based on the code](https://github.com/composer/composer/blob/d4150cafc46ccea244ec6fc4ea7519a0afe85aee/src/Composer/Util/PackageSorter.php#L30), `composer/installers` should always get loaded before `acquia/blt`, which is determined by subtracting the `acquia/blt` weight from `composer/installers'`. For example if blt has a weight of 1000, then installer's should be `1 - 1000` or `-999`. Items are then sorted in lowest to highest order. This should in theory, however do to the fact that packages are sorted starting from the root, if `composer/installers` is only required by `acquia/blt` then installer's is only reached by way of blt, who's weight will be evaluated as `0` resulting in `composer/installers` being weight 1, which only by convince is lower then blt's 1000. This algorithm should result in packages which are "required" by the most packages being loaded earlier then packages which only required once. This should also enforce that any package which is a dependency should get loaded earlier then any dependents, so I'm unsure there is an issue with the algorithm.

I think the issue here is that the referenced [refactoring of the package sorter](https://github.com/composer/composer/commit/266a41e0464c8ccf23949e3794189dc4ccf3caba) broke the sorting algorithm, right? The proposed changes in this pr should bring us back to a correctly sorted package list.

Seldaek:
@roygoldman I think you might be right, and as per https://github.com/composer/composer/pull/8723#issuecomment-606606957 all I wish for really is to have tests proving this so we can make sure it's correct and especially doesn't get broken in the future.

helhum:
@Seldaek Side note regarding the Travis status from the build not being reported back to this PR: A logout and re-login on Travis helped me to resolve this issue. Maybe this helps here as well.

I'll come up with some unit test for the PackageSorter then

danepowell:
@roygoldman thanks for clarifying, I wasn't familiar with that detail of the sorting algorithm. I'm satisfied that if we just get unit tests on this then a more extensive rework isn't necessary for now.

helhum:
I added some tests now (happy to include more cases if someone can come up with more scenarios I haven't thought of).

The tests show, that PackageSorter is broken without this change and fixed with it.

They also demonstrate that @roygoldman is right and even if package A has many depdendencies and B has not, but A depdends on B, that B will be sorted to be before A.

Seldaek:
Thanks :) Still targetting 2.0 with this fix tho to avoid more back and forth on v1 which at this point is slowly entering hibernation..

helhum:
Cool! Thanks for the merge!

Issue #7665, opened 2018-09-24
Composer hangs in "Something's changed, looking at all rules again" state


kelunik:
My [`composer.json`](https://github.com/prooph/event-store-client/blob/deee1b5f6336d9df7bae5bc0084ca8815ca96974/composer.json):

```json
{
  "name": "prooph/event-store-client",
  "type": "library",
  "license": "BSD-3-Clause",
  "authors": [
    {
      "name": "Sascha-Oliver Prolic",
      "email": "saschaprolic@googlemail.com"
    }
  ],
  "keywords": [
    "EventStore",
    "EventSourcing",
    "DDD",
    "prooph",
    "Amp",
    "Async"
  ],
  "prefer-stable": true,
  "require": {
    "php": "^7.2",
    "ext-json": "*",
    "amphp/amp": "^2.0.7",
    "amphp/artax": "^3.0.14",
    "amphp/byte-stream": "^1.3.1",
    "amphp/cache": "^1.2.0",
    "amphp/dns": "^0.9.13",
    "amphp/file": "^0.3.1",
    "amphp/log": "^1.0.0",
    "amphp/parallel": "^0.2.5",
    "amphp/process": "^0.3.3",
    "amphp/socket": "^0.10.9",
    "amphp/uri": "^0.1.3",
    "ramsey/uuid": "^3.7.3"
  },
  "require-dev": {
    "malukenho/docheader": "^0.1.4",
    "phpspec/prophecy": "^1.7",
    "phpunit/phpunit": "^7.3.3",
    "prooph/php-cs-fixer-config": "^0.2.1",
    "satooshi/php-coveralls": "^1.0"
  },
  "autoload": {
    "psr-4": {
      "Prooph\\EventStoreClient\\": "src/"
    }
  },
  "autoload-dev": {
    "psr-4": {
      "ProophTest\\EventStoreClient\\": "tests/"
    }
  },
  "scripts": {
    "check": [
      "@cs",
      "@test"
    ],
    "cs": "php-cs-fixer fix -v --diff --dry-run",
    "cs-fix": "php-cs-fixer fix -v --diff",
    "test": "phpunit"
  },
  "config": {
    "sort-packages": true
  }
}
```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
[...]

PR #8733, opened 2020-04-02
Test for issue resulting in overly long learnt rules and solver bug exception

naderman:
Addresses https://github.com/composer/composer/issues/7665


naderman:
completed the test case and rebased onto https://github.com/composer/composer/pull/9312 which fixes this problem too.

Seldaek:
Think you forgot to push or forgot --force here. This is not rebased.

naderman:
@Seldaek indeed forgot to --force, test is there now. Caveat: it's really slow, not sure that's acceptable?

Seldaek:
What are we talking about with really slow? If really too slow we could split it into its own test perhaps with `@group slow` but it'd require refactoring the test runner a bit I guess.

Does the test really bring anything tho that's the other question..

Also it's not passing right now.

PR #8742, opened 2020-04-07
extracted `VersionParser::DEV_MASTER_ALIAS`

clxmstaab:
as suggested in https://github.com/composer/composer/pull/8690/files#r404786378

### DIFF
--- a/src/Composer/Package/Version/VersionParser.php
+++ b/src/Composer/Package/Version/VersionParser.php
@@ -18,7 +18,7 @@ use Composer\Semver\Semver;
 
 class VersionParser extends SemverVersionParser
 {
-    public const DEV_MASTER_ALIAS = 'VersionParser::DEV_MASTER_ALIAS';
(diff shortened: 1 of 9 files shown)


src/Composer/Package/Version/VersionParser.php
```diff
@@ -18,6 +18,8 @@
 
 class VersionParser extends SemverVersionParser
 {
+    public const DEV_MASTER_ALIAS = 'VersionParser::DEV_MASTER_ALIAS';
```
Seldaek, line 21: I think I'd rather have this have a value of 9999999-dev still, as this might be relevant in some cases for sorting purposes and possibly other stuff I can't think of.

clxmstaab: sorry, this was a search/replace leftover :-)

Seldaek:
Thanks

PR #8744, opened 2020-04-07
use more precise phpstan/psam return-types

clxmstaab:
used `@psalm-return` in places where types are more advanced and my IDE (phpstorm) would not understand the type.

using regular `@return` in places where "regular IDEs" support the notation

### DIFF
--- a/src/Composer/Package/PackageInterface.php
+++ b/src/Composer/Package/PackageInterface.php
@@ -269,7 +269,8 @@ interface PackageInterface
      * Returns a set of package names and reasons why they are useful in
      * combination with this package.
      *
-     * @return array<string, string> An array of package suggestions with descriptions
+     * @return array An array of package suggestions with descriptions
+     * @return array<string, string>
      */
     public function getSuggests();
 
--- a/src/Composer/Package/PackageInterface.php
+++ b/src/Composer/Package/PackageInterface.php
@@ -159,7 +159,7 @@ interface PackageInterface
     /**
      * Returns the urls of the distribution archive of this version, including mirrors
      *
-     * @return array
+     * @return string[]
      */
     public function getDistUrls();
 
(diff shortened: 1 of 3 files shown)


clxmstaab:
didn't  expect this will go green on the first run :-)

src/Composer/Repository/RepositoryInterface.php
```diff
@@ -81,6 +83,7 @@ public function loadPackages(array $packageNameMap, array $acceptableStabilities
      * @param string $type  The type of package to search for. Defaults to all types of packages
      *
      * @return array[] an array of array('name' => '...', 'description' => '...')
+     * @psalm-return array{name: string, description: string}
```
clxmstaab, line 86: hmm I guess this is wrong..

clxmstaab:
seems good to go.. more in a separate PR

src/Composer/Package/PackageInterface.php
```diff
@@ -282,6 +283,7 @@ public function getSuggests();
      * directories for autoloading using the type specified.
      *
      * @return array Mapping of autoloading rules
+     * @psalm-return array<string, array<string, string>>
```
stof, line 286: that's not actually true. `psr-0` and `psr-4` keys are indeed of type `array<string, string>`. But for `classmap` and `files`, the type is `list<string>`

staabm: I see. Do you have a propsal which type to use instead?

stof: well, the return array is a struct array. We know the keys in it (even though they are all optional). So I think the `array{}` syntax should be used, to specify the type for each key.

staabm, line 286: So you mean something along ..

```suggestion
     * @psalm-return array{?psr-0: array<string, string>, ?psr-4: array<string, string>, ?classmap: list<string>, ?files: listr<string>}
```

Seldaek:
👍  thanks

Issue #8454, opened 2019-11-29
github will replace and discontinue authenticating using query parameters


yb199478:
I recently read the oauth documentation on github and found that the mention of deprecating authentication to the GitHub API using query parameters:

- https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters

Will this affect composer in the future?

https://github.com/composer/composer/blob/a0690525b52f841748176e1c43dbe900f07e72da/src/Composer/Util/RemoteFilesystem.php#L281

alcohol:
A quick glance over the blogpost and our code suggests that this might impact some of our code perhaps. But I was also under the impression we already used headers to authenticate in some scenarios or flows. This does warrant further inspection I think. Thank you for letting us know!

stof:
The blog post also mention deprecating the API for creating personal access tokens. I know for sure that composer used it in the past, but I don't remember whether that's still the case today.

yb199478:
The API for creating the personal token, I remember that it provided a uri to display, not directly generated through the api:

https://github.com/composer/composer/blob/e8626505718981bf67fd7262f65a611d4000f6c6/src/Composer/Util/GitHub.php#L89

stof:
OK, so it means we already removed our code generating a token.

Rarst:
FYI GitHub have started proactively notifying about this, I just got the following email:

> On February 4th, 2020 at 07:59 (UTC) your personal access token ( ... ) using Composer/1.9.1 (Windows NT; 10.0; PHP 7.4.1) was used as part of a query parameter to access an endpoint through the GitHub API:
> ...
> Please use the Authorization HTTP header instead, as using the `access_token` query parameter is deprecated.

Since the fix doesn't seem to be in stable release yet, this might go out a lot. :)

Seldaek:
Yup, saw this too. Will definitely get to get 1.10 out next week with this fix.

glensc:
They'll keep sending that email out every 3 days:

> Depending on your API usage, we'll be sending you this email reminder once every 3 days for each token and User-Agent used in API calls made on your behalf.
Just one URL that was accessed with a token and User-Agent combination will be listed in the email reminder, not all.

Seldaek:
1.9.3 is now out with the fix. `composer self-update` party.

taisph:
I'm still getting the GitHub notification with composer 1.9.3. :confused: 

```
[...]

PR #8747, opened 2020-04-08
Making a fallback for broken auth on github

matthiaz:
TLDR: We shouldn't be adding authentication on anything but the api.github.com url

Following the discussion about github deprecating auth via query string https://github.com/composer/composer/issues/8454
(Original blog post here https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters )

A fix was created see: https://github.com/composer/composer/commit/4b6c25d4bc33d49097320e29e6e5705b12e9d6ef

The problem is that this broke hardcoded URLs to tag archives. This works:

```
% curl -sv https://github.com/components/codemirror/archive/5.48.2.zip |& grep HTTP
> GET /components/codemirror/archive/5.48.2.zip HTTP/1.1
< HTTP/1.1 302 Found
```

Add a scopeless token, and it doesn't redirect anymore:

```
% curl -sv -H "Authorization: token <TOKENHERE>" https://github.com/components/codemirror/archive/5.48.2.zip |& grep HTTP
> GET /components/codemirror/archive/5.48.2.zip HTTP/1.1
< HTTP/1.1 404 Not Found
```

Why is this a problem?

In the PHP world, it is really not uncommon to see composer.lock files that have hardcoded URLs like this. There are a number of them hosted publicly on GitHub. [Here is a random example](https://github.com/eGovPDX/portlandor/blob/master/composer.lock):

```
        {
            "name": "codemirror/codemirror",
            "version": "5.48.2",
            "dist": {
                "type": "zip",
                "url": "https://github.com/components/codemirror/archive/5.48.2.zip"
            },
            "require": {
                "composer/installers": "~1.0"
            },
            "type": "drupal-library",
            "extra": {
                "installer-name": "codemirror"
            }
        },
```

This used to work, now it doesn't.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -424,6 +424,13 @@ class RemoteFilesystem
             }
         }
 
+        // check for github 404 when downloading .zip over non-codeload url
+        if($statusCode === 404 && preg_match('/https:\/\/github.com\/([^\/]+)\/([^\/]+)\/archive\/([^\/]+)\.([^\/]+)/',$fileUrl,$matches)) {
+            list($match,$user,$repo,$version,$extension) = $matches;
+            $fileUrl = "https://codeload.github.com/$user/$repo/$extension/$version";
+            return $this->get($this->originUrl, $fileUrl, $additionalOptions, $this->fileName, $this->progress);
+        }
+        
         // handle 3xx redirects, 304 Not Modified is excluded
         $hasFollowedRedirect = false;
         if ($statusCode >= 300 && $statusCode <= 399 && $statusCode !== 304 && $this->redirects < $this->maxRedirects) {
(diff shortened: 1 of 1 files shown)


damz:
Alternatively, we could only add the authentication token when the domain is literally `https://api.github.com`. Adding authentication on `https://github.com` itself never worked, and was only harmless because it was just adding a query parameter.

matthiaz:
@damz Thanks for the pointer, I've changed the PR to reflect this.

damz:
:+1: That sounds right to me. I think we could even restrict this to only `https`, because sending authentication credentials over unencrypted HTTP sounds like a bad idea to begin with.

Seldaek:
Thanks, I think this makes a lot of sense, funny how things can be wrong yet working for ages sometimes.

PR #8750, opened 2020-04-09
Allow duplicate dashes in package names

### LINKED ISSUE #8749: Package naming in 2.0 and the asset ecosystem
Hi, I wanted to make you aware of an impact of the package naming restriction in 2.0, which is going to affect users of https://asset-packagist.org/ and [composer-asset-plugin](https://github.com/fxpio/composer-asset-plugin/issues/333). This is not about whether one should use these (I know your opinion on this @Seldaek :smiley:) but the fact that upgrading to Composer 2.0 becomes really hard (currently impossible) for people on this ecosystem.

### The issue

Composer 2.0 will restrict the use of `-` (dashes). When converting npm package names which include a vendor name (`vendor/package`) we can not keep the `/` because having multiple `/` in a package name has obvious issues (as discussed in #7975). Currently the `/` is converted to `--` (see [docs](https://github.com/fxpio/composer-asset-plugin/blob/fe3943dfddbb599d77002734a7077dc5d98cfe4f/Resources/doc/index.md#work-with-the-npm-scopes)). Double dashes however are not valid anymore in Composer 2.0.

### How to solve this issue?

There are two ways to solve this:

1. The easiest solution: allow multiple dashes in package names.
2. Come up with a different replacement when converting package names in `composer-asset-plugin`. 
  Given that we find a replacement that matches the naming restrictions and does not cause other problems (which is already quite hard), this also means a lot of upgrade work as people need to rename all their package requirements.
  We also we need a breaking change in the repository of asset-packagist as well as a new major version of composer-asset-plugin, which is incompatible with the current version, meaning a composer.json file will work either with version 1 or 2 but never with both. The fact that the plugin needs to be installed globally will make it very hard to work on two projects with different state. 

Allowing multiple dashes in package names is not a huge problem for composer and does not introduce any overhead as far as I see. Compared to the amount of work needed to make the composer-asset-plugin/asset-packagist ecosystem compatible with Composer 2.0.

Thank you!

Btw, docs should be adjusted:

https://github.com/composer/composer/blob/master/doc/04-schema.md#name

(I can help with that once this issue is solved)

cebe:
Just to have this request complete, I'd like to show how small the change is to make everyone happy :)

fixes #8749

I'll add a test if you agree to merge it.

### DIFF
--- a/src/Composer/Package/Loader/ValidatingArrayLoader.php
+++ b/src/Composer/Package/Loader/ValidatingArrayLoader.php
@@ -374,8 +374,8 @@ class ValidatingArrayLoader implements LoaderInterface
             return;
         }
 
-        if (!preg_match('{^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9]([_.-]?[a-z0-9]+)*$}iD', $name)) {
-            return $name.' is invalid, it should have a vendor name, a forward slash, and a package name. The vendor and package name can be words separated by -, . or _. The complete name should match "[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9]([_.-]?[a-z0-9]+)*".';
+        if (!preg_match('{^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9]([_.-]{0,2}[a-z0-9]+)*$}iD', $name)) {
+            return $name.' is invalid, it should have a vendor name, a forward slash, and a package name. The vendor and package name can be words separated by -, . or _. The complete name should match "[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9]([_.-]{0,2}[a-z0-9]+)*".';
         }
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Fine by me with the revised regex now, if you can add that test that'd be great :)

stof:
should this change be done in 1.x instead to have the right regex for the warning there ?

Seldaek:
Yes that'd probably make sense.. I was wondering about that earlier.. @cebe have you guys been running with tons of warnings for the last few months without saying anything?!

cebe:
> have you guys been running with tons of warnings for the last few months without saying anything?!

Well, https://github.com/hiqdev/asset-packagist/issues/103 was created on 16 Sep 2019, I just got aware of this last week. It does not happen for all packages, only those with vendor prefix, so if you only use e.g. jquery all works fine. With other packages the warning is showing up.

I'll make another branch to make the change in 1.10.

While writing the test for master, I found that even in 2.0 it is a warning, are you going to change that later? Maybe turn warnings into errors now to ensure hard failures for people testing out 2.0?

cebe:
> should this change be done in 1.x instead to have the right regex for the warning there ?

here you go: https://github.com/composer/composer/pull/8767

cebe:
@Seldaek how are you going to handle the warning -> error change? should I adjust this PR or are you merging 1.10 into master eventually and change warning to error later?

Issue #8751, opened 2020-04-09
broken symfony/console method call


etarel:
Since `symfony/console` removed `TableStyle::setVerticalBorderChar()` method in favor of `TableStyle::setVerticalBorderChars()` in 5.0.0(see [CHANGELOG](https://github.com/symfony/console/blob/master/CHANGELOG.md#500) here)

When I run this command:

```
composer why package
```

I get the following error:

```
PHP Fatal error:  Uncaught Error: Call to undefined method Symfony\Component\Console\Helper\TableStyle::setVerticalBorderChar() in vendor/composer/composer/src/Composer/Command/BaseDependencyCommand.php:187
```

PR #8752, opened 2020-04-09
compatible with symfony/console 5.0.0 getVerticalBorderChars() method

etarel:
#8751


Seldaek:
Thanks, merged this in 1.10 and master

PR #8754, opened 2020-04-09
added phpdocs

clxmstaab:


src/Composer/Downloader/GitDownloader.php
```diff
@@ -27,9 +27,21 @@
  */
 class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
 {
+    /**
+     * @var array<string, bool>
+     */
     private $hasStashedChanges = array();
+    /**
+     * @var array<string, bool>
```
Seldaek, line 35: Is this syntax supported by phpstorm & co? Or does one need to use psalm-var + `@var bool[]`?

clxmstaab: arghs didn't think of the IDE.. you are right... will fix

clxmstaab:
@Seldaek fixed per feedback

Seldaek:
Thanks, it is a PITA tbh and it'd be much nicer if there was common agreed upon syntax for this :/

staabm:
yeah absolutely. hopefully the phpstorm team reads its bugtracker and reacts based on the issue which have hundreds of +1s on this topic

PR #8755, opened 2020-04-09
Add options to configure repository priorities

Seldaek:


doc/05-repositories.md
```diff
@@ -41,14 +41,20 @@ be preferred.
...
+By default only the Packagist.org repository is registered in Composer. You can
 add more repositories to your project by declaring them in `composer.json`.
 
 Repositories are only available to the root package and the repositories
 defined in your dependencies will not be loaded. Read the
 [FAQ entry](faqs/why-can't-composer-load-repositories-recursively.md) if you
 want to learn why.
 
+When resolving dependencies, packages are looked up from repositories from
+top to bottom, and by default as soon as a package is found in one Composer
```
naderman, line 53: missing comma
```suggestion
top to bottom, and by default as soon as a package is found in one, Composer
```

doc/articles/repository-priorities.md
```diff
@@ -0,0 +1,94 @@
+<!--
+    tagline: Configure which packages are found in which repositories
+-->
+
+# Repository priorities
+
+## Canonical repositories
+
+When Composer resolves dependencies it will look up a given package in the
```
naderman, line 9:
```suggestion
When Composer resolves dependencies, it will look up a given package in the
```

doc/articles/repository-priorities.md
```diff
@@ -0,0 +1,94 @@
...
+topmost repository. If that repository does not contain the package, it
+goes on to the next one, until one repository contains it and the process ends.
+
+Canonical repositories are better for a few reasons:
+
+- Performance wise, it is more efficient to stop looking for a package once it
+has been found somewhere. It also avoids loading duplicate packages in case
+the same package is present in several of your repositories.
+- Security wise, it is safer to treat them canonically as it means that your most
+important repositories will return the packages you expect them to always. Let's
```
naderman, line 19: This isn't quite right, since you have no guarantees over what packages are returned by top repo, just that they don't get loaded from elsewhere if they get returned.

```suggestion
- Security wise, it is safer to treat them canonically as it means that packages you
expect to come from your most important repositories will never be loaded from 
another repository instad. Let's
```

doc/articles/repository-priorities.md
```diff
@@ -0,0 +1,94 @@
...
+say you have a private repository which is not canonical, and you require your
+private package `foo/bar ^2.0` for example. Now if someone publishes
+`foo/bar 2.999` to packagist.org, suddenly Composer will pick that package as it
+has a higher version than your latest release (say 2.4.3), and you end up install
```
naderman, line 23:
```suggestion
has a higher version than your latest release (say 2.4.3), and you end up installing
```

doc/articles/repository-priorities.md
```diff
@@ -0,0 +1,94 @@
...
+            "canonical": false
+        }
+    ]
+}
+```
+
+## Filtering packages
+
+You can also filter packages which a repository will be able to load, either by
+selecting which you want, or by excluding those you do not want.
```
naderman, line 61:
```suggestion
selecting which ones you want, or by excluding those you do not want.
```

doc/articles/repository-priorities.md
```diff
@@ -0,0 +1,94 @@
...
+
+For example here we want to pick only the `foo/bar` and all the packages from
```
naderman, line 63:
```suggestion
For example here we want to pick only the package `foo/bar` and all the packages from
```

doc/articles/repository-priorities.md
```diff
@@ -0,0 +1,94 @@
...
+        {
+            "type": "composer",
+            "url": "https://example.org",
+            "exclude": ["toy/package"]
+        }
+    ]
+}
+```
+
+Both `only` and `exclude` should be array of package names, which can also
```
naderman, line 93:
```suggestion
Both `only` and `exclude` should be arrays of package names, which can also
```

src/Composer/Repository/FilterRepository.php
```diff
@@ -0,0 +1,193 @@
...
+     */
+    public function loadPackages(array $packageMap, array $acceptableStabilities, array $stabilityFlags)
+    {
+        foreach ($packageMap as $name => $constraint) {
+            if (!$this->isAllowed($name)) {
+                unset($packageMap[$name]);
+            }
+        }
+
+        $result = $this->repo->loadPackages($packageMap, $acceptableStabilities, $stabilityFlags);
```
naderman, line 119: Actually should this be skipped entirely if the package map is empty? Avoids loading repos that are irrelevant?

Seldaek: Yes true will fix

src/Composer/Repository/FilterRepository.php
```diff
@@ -0,0 +1,193 @@
...
+
+        return $result;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function search($query, $mode = 0, $type = null)
+    {
+        return $this->repo->search($query, $mode, $type);
```
stof, line 132: shouldn't this filter the search result too ?

src/Composer/Repository/FilterRepository.php
```diff
@@ -0,0 +1,193 @@
...
+    {
+        return $this->repo->removePackage($package);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function count()
+    {
+        return $this->repo->count();
```
stof, line 178: this looks wrong to me, as it does not reflect the filtering.

Seldaek: You're right of course, for both functions I decided to skip because it doesn't matters too much in most conditions and it was additional work. I'll happily take a PR for it though.

src/Composer/Repository/FilterRepository.php
naderman: I think this one we should really do @Seldaek cause I think it may lead to some confusion/misunderstandings.

Seldaek: Done in https://github.com/composer/composer/commit/6bed9d8f13c2ef467ba543902a30e4d11e339451

PR #8767, opened 2020-04-10
Allow duplicate dashes in package names (1.10)

### LINKED ISSUE #8749: Package naming in 2.0 and the asset ecosystem

cebe:
This is #8750 but for the 1.10 branch and including tests.

fixes #8749

### DIFF
--- a/src/Composer/Package/Loader/ValidatingArrayLoader.php
+++ b/src/Composer/Package/Loader/ValidatingArrayLoader.php
@@ -374,8 +374,8 @@ class ValidatingArrayLoader implements LoaderInterface
             return;
         }
 
-        if (!preg_match('{^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9]([_.-]?[a-z0-9]+)*$}iD', $name)) {
-            return $name.' is invalid, it should have a vendor name, a forward slash, and a package name. The vendor and package name can be words separated by -, . or _. The complete name should match "[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9]([_.-]?[a-z0-9]+)*".';
+        if (!preg_match('{^[a-z0-9](?:[_.-]?[a-z0-9]+)*/[a-z0-9](?:(?:[_.]?|-{0,2})[a-z0-9]+)*$}iD', $name)) {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

schmunk42:
@Seldaek 
What's about the 1.x branch and packagist? 
Would this have to be ported to fix the warnings they are issuing?

Seldaek:
This is the port to 1.x?!

schmunk42:
Ah sorry, I missed that - so it will also make it to packagist.

I am getting errors for this repo https://packagist.org/packages/dmstr/yii2-vue-shufflejs-widget

Seldaek:
@schmunk42 fixed now

PR #8770, opened 2020-04-12
Add missing pre-status-cmd event in doc

villfa:
This line was removed (by mistake I guess) with this commit: https://github.com/composer/composer/commit/4d1416050dcf3493d2fb2d411dc1490d47fd3001#diff-d34ee8b7fff0f721fe2321f9c22e34f0


Seldaek:
Thanks

Issue #8759, opened 2020-04-09
Arrays are not allowed in class constants 


Drewzar:
Composer version 1.10.3 2020-04-09 16:26:35

My `composer.json`:

```json
{
    "require": {
        "guzzlehttp/guzzle": "^6.0",
        "smarty/smarty": "3.1.12",
        "musicglue/sidekiq-job-pusher": "dev-master",
        "symfony/yaml": ">=2.0,<3.0",
        "league/flysystem": "1.0.4",
        "league/flysystem-aws-s3-v3": "1.0.9",
        "league/flysystem-memory": "^1.0",
        "league/glide": "dev-master",
        "aws/aws-sdk-php": "^3.4",
        "codeguy/upload": "dev-master",
        "cloudinary/cloudinary_php": "dev-master",
        "drewm/mailchimp-api": "^2.2",
        "firebase/php-jwt": "^4.0",
        "paypal/rest-api-sdk-php": "*"
    },
    "require-dev": {
        "phpunit/phpunit": "^4.8"
    },
    "minimum-stability": "dev",
    "config": {
        "bin-dir": "bin/"
    },
    "autoload": {
        "psr-0": {
            "Liftopia": "lib/"
        }
    }
}
```

Output of `composer diagnose`:

```
Cannot create cache directory /home/me/.composer/cache/repo/https---repo.packagist.org/, or directory is not writable. Proceeding without cache
Cannot create cache directory /home/me/.composer/cache/files/, or directory is not writable. Proceeding without cache
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
require.smarty/smarty : exact version constraints (3.1.12) should be avoided if the package follows semantic versioning
require.musicglue/sidekiq-job-pusher : unbound version constraints (dev-master) should be avoided
require.league/flysystem : exact version constraints (1.0.4) should be avoided if the package follows semantic versioning
require.league/flysystem-aws-s3-v3 : exact version constraints (1.0.9) should be avoided if the package follows semantic versioning
require.league/glide : unbound version constraints (dev-master) should be avoided
require.codeguy/upload : unbound version constraints (dev-master) should be avoided
require.cloudinary/cloudinary_php : unbound version constraints (dev-master) should be avoided
require.paypal/rest-api-sdk-php : unbound version constraints (*) should be avoided
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
[...]

PR #8773, opened 2020-04-12
Check syntax of PHP files in src

villfa:
This check detects syntax errors like this one: #8759

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,6 +71,7 @@ before_script:
   - git config --global user.email travis@example.com
 
 script:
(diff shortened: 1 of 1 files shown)


Seldaek:
IMO it should be made to run only on the 5.3 build to avoid slowing down all builds, but also the question is do we add this here or should we switch to github actions first https://github.com/composer/composer/pull/8662 - that's still kinda open so maybe best wait a bit here.

Seldaek:
Ok just going to merge this already, thanks

PR #8774, opened 2020-04-12
Update HTTP URLs to their HTTPS if they already redirect automatically

Ayesh:
A low-effort PR 🙊. 

Some of the old URLs used in composer still use plain HTTP. I have scanned through, and updated them where appropriate and after checking if they automatically redirect to HTTPS. It does _not_ include HTTP URLs that refer to namespaces, schema, etc, nor change examples (that use `example.com` as the domain name). 

Thank you.

## Summary of changes

*The following domains automatically redirect HTTPS*

 - php.net
 - php-fig.org
 - getcomposer.org
 - git-scm.com
 - naderman.de
 - seld.be
 - tools.ietf.org

*URL below support both HTTP and HTTPS, and are updated to their HTTPS variants*
 - json-schema.org

*URL below redirects to a new URL, but with same contents*
  - `http://contributor-covenant.org/version/1/4/` -> `https://www.contributor-covenant.org/version/1/4/code-of-conduct/` (301 to HTTPs -> 301 to `www` -> 301 to new URL).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,7 +2,7 @@
 *       text=auto eol=lf
 
 # These files are always considered text and should use LF.
-# See core.whitespace @ http://git-scm.com/docs/git-config for whitespace flags.
+# See core.whitespace @ https://git-scm.com/docs/git-config for whitespace flags.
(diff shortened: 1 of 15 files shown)


Seldaek:
Thanks

PR #8775, opened 2020-04-13
PHP 8 fixes for functional optional parameters before required parameters

Ayesh:
Related to #8686

I linted the current master branch, with PHP master builds, and found two more deprecation notices:

1. `Deprecated: Required parameter $name follows optional parameter $constraint in src\Composer\Repository\ComposerRepository.php on line 745`

2.  `Deprecated: Required parameter $operation follows optional parameter $operations in src\Composer\Installer\PackageEvent.php on line 73`
…ository\ComposerRepository::isVersionAcceptable() required arguments used after optional, which is deprecated in PHP 8.0

Optional parameters with a type declared, and a default value of `null` is excepted from this deprecation. See https://php.watch/versions/8.0/deprecate-required-param-after-optional. This is the case in `ComposerRepository::isVersionAcceptable`, which still has two optional parameters as first two parameters, but this will not raise a deprecation notice.

### DIFF
--- a/src/Composer/Installer/PackageEvent.php
+++ b/src/Composer/Installer/PackageEvent.php
@@ -70,7 +70,7 @@ class PackageEvent extends Event
      * @param OperationInterface[] $operations
      * @param OperationInterface   $operation
      */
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #8777, opened 2020-04-13
Update 04-schema.md to reflect package naming in 2.0

cebe:
issue #8749


doc/04-schema.md
```diff
@@ -34,12 +34,13 @@ separated by `/`. Examples:
...
-to simplify its installation, it's recommended to define a short and lowercase
-name that doesn't include non-alphanumeric characters or white spaces.
+The name must consist of words separated by `-`, `.` or `_`.
+The complete name should match `^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$`.
 
-Required for published packages (libraries).
+The `name` property is required for published packages (libraries).
+
+> **Note:** Before Composer version 2.0, a name could contain any character, including white spaces, and it was case
+> insensitive (`foo/bar` and `Foo/Bar` are considered the same package).
```
cebe, line 43: note sure about "case insensitive" here. This might still be the case but it ist not relevant if everything is lowercase now?

Seldaek: Right it still is technically case insensitive but this has no effect in practice. I'd maybe remove that part of the sentence tho (all after the last comma) to avoid inducing people in error.

Seldaek:
Thanks

Issue #8353, opened 2019-10-02
Possible to set deep "extra" config values via the command-line interface? (workarounds included)


TravisCarden:
tl;dr: I need to set (>3 levels) deep "extra" config values from the command-line. Is it possible to do this with Composer, or must one resort to external tools? Could the 3 levels depth limit be increased?

The documentation for [modifying extra values via the command-line interface](https://getcomposer.org/doc/03-cli.md#modifying-extra-values) states, "dots indicate array nesting, a max depth of 3 levels is allowed though". Unfortunately, I frequently need to set deeper values--for [cweagans/composer-patches](https://github.com/cweagans/composer-patches) config, for example, which looks like this:

```json
{
  "extra": {
    "patches": {
      "vendor/project": {
        "Patch title": "http://example.com/url/to/patch.patch"
      }
    }
  }
}
```

Ideally, I would create that like this:

```bash
composer config "extra.patches.vendor/project.Patch title" http://example.com/url/to/patch.patch
```

But of course that results in this instead:

```json
{
  "extra": {
    "patches": {
      "vendor/project.Patch title": "http://example.com/url/to/patch.patch"
    },
  }
}
```

Is it currently possible to accomplish this some other way via the command-line interface? If not, could it be _made_ possible--perhaps by deepening the maximum depth limit or making it configurable via argument (e.g., `composer config --max-depth=5`)?

## Workarounds for weary travelers

- For the Composer patches use case specifically, [using an external patch file](https://github.com/cweagans/composer-patches#using-an-external-patch-file) should probably be the cleanest solution, though I have not personally succeeded in making it work, e.g.:

    ```bash
    composer config extra.patches-file local/path/to/your/composer.patches.json
    ```

- Another good command-line solution (which is installed by default in Travis CI build environments) is [jq](https://stedolan.github.io/jq/), e.g.:

    ```bash
    jq '.extra.patches = {"vendor/project": {"Patch title": "http://example.com/url/to/patch.patch"}}' composer.json > temp.json && mv temp.json composer.json
    ```

- Within a PHP script, it can be accomplished like this:

    ```php
    $json_file = new \Composer\Json\JsonFile('composer.json');
    $data = $json_file->read();
    $data['extra']['patches']['vendor/project']['Patch title'] = 'http://example.com/url/to/patch.patch';
    $json_file->write($data);
    ```

After any of these you'll need to force Composer to update the lock file and (in the case of Composer patches, specifically) redownload and patch the affected project:

```bash
[...]

PR #8779, opened 2020-04-13
Added --json and --merge options to ConfigCommand

effulgentsia:
This is a PR for https://github.com/composer/composer/issues/8353#issuecomment-599117797.

The goal is to enable using composer config for setting deeper than the current limit of 3 by allowing the value to be specified as a JSON array or object. For example:

`composer config --json extra.patches.vendor/project '{"Patch title": "http://example.com/url/to/patch.patch"}'`

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -73,8 +73,8 @@ class ConfigCommand extends BaseCommand
                 new InputOption('list', 'l', InputOption::VALUE_NONE, 'List configuration settings'),
                 new InputOption('file', 'f', InputOption::VALUE_REQUIRED, 'If you want to choose a different composer.json or config.json'),
                 new InputOption('absolute', null, InputOption::VALUE_NONE, 'Returns absolute paths when fetching *-dir config values instead of relative'),
-                new InputOption('json', 'j', InputOption::VALUE_NONE, 'JSON decode the setting value'),
-                new InputOption('merge', 'm', InputOption::VALUE_NONE, 'Merge the setting value with the current value'),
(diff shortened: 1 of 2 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -73,6 +73,8 @@ protected function configure()
                 new InputOption('list', 'l', InputOption::VALUE_NONE, 'List configuration settings'),
                 new InputOption('file', 'f', InputOption::VALUE_REQUIRED, 'If you want to choose a different composer.json or config.json'),
                 new InputOption('absolute', null, InputOption::VALUE_NONE, 'Returns absolute paths when fetching *-dir config values instead of relative'),
+                new InputOption('json', 'j', InputOption::VALUE_NONE, 'JSON decode the setting value'),
+                new InputOption('merge', 'm', InputOption::VALUE_NONE, 'Merge the setting value with the current value'),
```
Seldaek, line 77: These should definitely mention they only work for `extra.*` config keys.

src/Composer/Command/ConfigCommand.php
```diff
@@ -621,7 +623,21 @@ function ($vals) {
...
+                $bits = explode('.', $settingKey);
+                foreach ($bits as $bit) {
+                  $current_value = isset($current_value[$bit]) ? $current_value[$bit] : NULL;
+                }
+                if (is_array($current_value)) {
+                  $value = array_merge($current_value, $value);
+                }
+              }
+            }
+            $this->configSource->addProperty($settingKey, $value);
```
Seldaek, line 640: If you can switch the above to 4 space indents and use camelCase var names I think that's all I have to complain about implementation wise. Looks fine to me as far as I can tell from looking at the code, I'd have to quickly try it still though before merging.

Seldaek:
Thanks!

PR #8782, opened 2020-04-14
added phpstan types to `CompletePackageInterface`

clxmstaab:


src/Composer/Package/CompletePackageInterface.php
```diff
@@ -69,13 +73,15 @@ public function getHomepage();
      * Each item can contain name/homepage/email keys
      *
      * @return array
+     * @psalm-return array<array{?name: string, ?homepage: string, ?email: string}>
```
stof, line 76: There is also an optional `role`

src/Composer/Package/CompletePackageInterface.php
```diff
@@ -32,20 +33,23 @@ public function getScripts();
      * {"<type>": {<config key/values>}}
      *
      * @return array Repositories
+     * @psalm-return array<string, array{type: string, url: string, allow_ssl_downgrade: bool}>
```
stof, line 36: Not all repository types have a `url`. That's more complex than that. Each type has its own fields (see the json schema)

clxmstaab: thx for the hint. changed the url to be optional.

tbh: I dont think whether its usefull to repeat the complex types which are defined in the schema in the docblock

clxmstaab:
@Seldaek wdyt about upping the phpstan level of composer to e.g. 3 or 4 so we can see more meaningfull errors and commit a [phpstan baseline](https://phpstan.org/blog/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard) to get a green build?

src/Composer/Package/CompletePackageInterface.php
```diff
@@ -85,6 +91,7 @@ public function getSupport();
      * Each item will contain type and url keys
      *
      * @return array
```
Seldaek, line 93: Could be made `array[]` too I suppose.. it's not much info but it's something :)

src/Composer/Package/CompletePackageInterface.php
```diff
@@ -32,20 +33,23 @@ public function getScripts();
...
      * @return array The package licenses
+     * @psalm-return string[]
      */
     public function getLicense();
 
     /**
      * Returns an array of keywords relating to the package
      *
      * @return array
+     * @psalm-return string[]
```
Seldaek, line 52: This too could be simply string[] on the `@return`, no need for psalm annotations

clxmstaab:
adjusted per Jordi's feedback

Seldaek:
Thanks

Seldaek:
As for the baseline yeah you can give it a go if it doesn't end up being a 2MB file maybe it's worth it :) I kinda imagine it'll be worse here than on composer/semver tho.

PR #8792, opened 2020-04-15
Remove the pear repository type from the JSON schema

stof:
This was missed in the removal.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -640,7 +640,6 @@
                 { "$ref": "#/definitions/vcs-repository" },
                 { "$ref": "#/definitions/path-repository" },
                 { "$ref": "#/definitions/artifact-repository" },
-                { "$ref": "#/definitions/pear-repository" },
                 { "$ref": "#/definitions/package-repository" }
             ]
         },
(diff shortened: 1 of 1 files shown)


stof:
hmm, actually, it looks like the config is still meant to accept it, but throwing an exception.

Seldaek:
Yeah I thought it's better to provide a decent error, but I think we can also remove it from the schema..

cebe:
When removing it from the schema, the error message is quite cryptic now:

```
$ ../bin/composer install

                                                                                                                                                                    
   - repositories[0] : Must contain no more than 1 properties                                                                                                       
   - repositories[0].type : String value found, but a boolean is required                                                                                           
   - repositories[0].type : Does not have a value in the enumeration [false]                                                                                        
   - repositories[0].url : String value found, but a boolean is required                                                                                            
   - repositories[0].url : Does not have a value in the enumeration [false]                                                                                         
   - repositories[0].package : The property package is required                                                                                                     
   - repositories[0].type : Does not have a value in the enumeration ["package"]                                                                                    
   - repositories[0].type : Does not have a value in the enumeration ["composer"]                                                                                   
   - repositories[0].type : Does not have a value in the enumeration ["vcs","github","git","gitlab","git-bitbucket","hg","hg-bitbucket","fossil","perforce","svn"]  
   - repositories[0].type : Does not have a value in the enumeration ["path"]                                                                                       
   - repositories[0].type : Does not have a value in the enumeration ["artifact"]                                                                                   
   - repositories[0] : Failed to match exactly one schema                                                                                                           
                                                                                                                                                                    

```

Seldaek:
Ok in that case not worth the cleanup IMO. Let's keep it for now.

PR #8793, opened 2020-04-15
Add package naming pattern to the composer.json JSON schema

cebe:
as discussed in #8749

```
$ ../bin/composer validate

                                                                                                              
   - name : Does not match the regex pattern ^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$  
                                                                                                              

```

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -7,7 +7,8 @@
     "properties": {
         "name": {
             "type": "string",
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #8795, opened 2020-04-15
Add PHP CS Fixer to CI

kubawerlos:
Why there is a config for PHP CS Fixer and it's not a part of CI?

Should I apply the changes to the codebase?

Edit: these changes from the failing job: https://travis-ci.org/github/composer/composer/jobs/675378416

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,11 @@ matrix:
       env:
         - deps=high
         - SYMFONY_PHPUNIT_VERSION=7.5
+    # PHP CS Fixer checks
+    - php: 7.4
+      env:
+        - deps=high
+        - PHP_CS_FIXER=1
     # PHPStan checks
     - php: 7.4
       env:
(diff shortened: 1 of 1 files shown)


kubawerlos:
Ping @Seldaek for an opinion.

Seldaek:
CS application is a job for maintainers IMO, so not interested in CI but thanks.

keradus:
@Seldaek , PHP CS Fixer is not applying "Coding Style" only, but "Coding Standards" - eg, ensuring project is using strict construction (like strict param for `in_array`), use new framework constructions (Eg for PHPUnit) instead of legacy etc. With that in mind, it's hard to understand why to run one tool like that (PHPStan), but not another.

Seldaek:
@keradus for me phpstan is mostly about finding bugs, while php-cs-fixer is mostly about correcting code style. Of course they overlap a little bit, although I wish every tool would stay focused on one thing.

But the bottom line is I want bugs to be found early and before merging if possible. Code style I don't care if it lags a little bit, so I don't want to add an extra hurdle for contributors. When contributing to some project and you get slapped in the face by some tool you may not know telling you you did it wrong, you can say "it's easy just run x" but as a newcomer this can be a pain, you may not be familiar with how to run the tool, etc. It just adds friction to open source contribution IMO. You don't have to agree with me.

keradus:
thanks for your answer @Seldaek , appreciating
I can somehow agree with your reasoning about "Style" part, and it's hard to run Fixer only about "Standards" but skip all rules for "Style". I understand you had to find a golden point for new contributors how much to poke them about it.

PR #8799, opened 2020-04-16
Mono/1.10.5

MWDDevelopment:
If you try to run a mono-repository in Git where, within a single repository, you have multiple projects existing as their own unrelated branches, composer has some bugs related to branches and tags connected to a project. It mistakenly mixes tags and branches improperly to the wrong projects. This code change creates a new repository type called "gitmono" that does not suffer these ill effects. These changes also allow for namespacing tags within the mono-repository (refs/tags/<project name>/<schematic version>).

### DIFF
--- a/src/Composer/Composer.php
+++ b/src/Composer/Composer.php
@@ -45,14 +45,14 @@ class Composer
      * const SOURCE_VERSION = '';
      *
      * source (git clone):
-     * const VERSION = '@package_version@';
-     * const BRANCH_ALIAS_VERSION = '@package_branch_alias_version@';
-     * const RELEASE_DATE = '@release_date@';
+     * const VERSION = '1.10.5';
+     * const BRANCH_ALIAS_VERSION = '';
+     * const RELEASE_DATE = '2020-04-10 11:44:22';
      * const SOURCE_VERSION = '1.8-dev+source';
      */
-    const VERSION = '@package_version@';
-    const BRANCH_ALIAS_VERSION = '@package_branch_alias_version@';
-    const RELEASE_DATE = '@release_date@';
+    const VERSION = '1.10.5';
+    const BRANCH_ALIAS_VERSION = '';
+    const RELEASE_DATE = '2020-04-10 11:44:22';
     const SOURCE_VERSION = '1.10-dev+source';
 
     public static function getVersion()
--- a/src/Composer/Repository/RepositoryFactory.php
+++ b/src/Composer/Repository/RepositoryFactory.php
@@ -122,6 +122,7 @@ class RepositoryFactory
         $rm->setRepositoryClass('package', 'Composer\Repository\PackageRepository');
         $rm->setRepositoryClass('pear', 'Composer\Repository\PearRepository');
         $rm->setRepositoryClass('git', 'Composer\Repository\VcsRepository');
+        $rm->setRepositoryClass('gitmono', 'Composer\Repository\VcsRepository');
         $rm->setRepositoryClass('git-bitbucket', 'Composer\Repository\VcsRepository');
         $rm->setRepositoryClass('github', 'Composer\Repository\VcsRepository');
         $rm->setRepositoryClass('gitlab', 'Composer\Repository\VcsRepository');
--- a/src/Composer/Repository/Vcs/GitMonoDriver.php
+++ b/src/Composer/Repository/Vcs/GitMonoDriver.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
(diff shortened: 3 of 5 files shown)


Seldaek:
Sorry but I don't think we want to support this kind of messy practice in Composer. If the branches share nothing why are you storing all this stuff in the same repo in the first place? 

Anyway, you can probably achieve this by writing a plugin which you would first install globally with `composer global require your/plugin` and that would add the new repository driver to the repository manager at runtime.

mdavi1982:
I understand your concern. In response to your question about why we store
unrelated branches in the same repo, here is my explanation.

In the Fortune 500 corporation I work for's development house, we build our
software using modularized development with composer as the composition
system. In other words, we write individual composer-ized packaged at
functional levels of libraries, applets, and applications, track their
individual versioning, and re-use them repeatedly .

With the number of developers and packages that we have, without a
mono-repository approach, there is too much overhead to be functional.
Which is why a lot of large corporations have been moving towards
mono-repositoried in recent years. With traditional repository
implementations, every time a new library is written, that means a new file
location. Every new file location at my corporation would mean the
following:

- Setting up automated backups and redunencies for the new location.
- Reconfiguring our source documentation system to auto-generate
documentation for the new location.
- Hooking in our automated building and testing environments to the new
location.
- Every developer in the company having to adjust their composer configs to
include the new location as an available source.
- Numerous other environmental adjustments.
...

This also increases available package transparency since all of our
packages show up when we use the "composer show --available <company
identifier>/*" command.

So we have a single master repository and namespace the branches and tags.
For example, among the countless projects in our master repository we have
the following refs:

refs/heads/configuration-manager/master
refs/tags/configuration-manager/1.0.0
refs/tags/configuration-manager/1.0.1
refs/heads/configuration-driver/master
refs/tags/configuration-driver/1.0
refs/heads/configuration-driver-ini/master
refs/tags/configuration-driver-ini/1.0.0
refs/tags/configuration-driver-ini/1.0.1
refs/heads/configuration-driver-json/master
refs/tags/configuration-driver-json/1.0.0
refs/heads/configuration-driver-xml/master
refs/tags/configuration-driver-xml/1.0.0
refs/heads/authentication-manager/master
refs/tags/authentication-manager/1.0.0
refs/tags/authentication-manager/1.1.0
refs/heads/authentication-password/master
refs/tags/authentication-password/1.0.0
refs/heads/authentication-saml/master
refs/tags/authentication-saml/1.0.0
...

Our development staff operate their development environments in a
distributed architecture: Fetching from our master mono-repository; Doing
their development; Pushing and pulling from each other when collaboration
needs require it; Submiting pull request back to our master repository.

Our Release Managment group manages the dissemination of completed software
package versions from the master mono-repository.

Just like placing packages into packagist for composer requires that the
package contain a composer.json file, all of our branches and their tags
contain their own composer file.

The git driver included in the current composer package makes assumptions
that are not valid in this type of scenario, which cause the following
issues.

- The driver searches other branches and guesses at a composer.json file if
one can't be found in the current worktree.
- The driver assumes that all tags correspond to all branches in that
repository.
- The driver is incapable of handling namespaced tags. (Which in my
opinion, is beneficial whether you're using a mono-repository or not.)

This driver corrects those issues without affecting the architecture in
place for the currently implemented git driver.

I wrote this variation and driver for the use of my corporation, but my
corporation does not have any claim to it since I developed it, and
maintain it in my own time from my person development environment at home.
The corporation has been using it for years now - with me rebasing it in my
off time regularly to bring in updates.

Since the Fortune 500 corporation I work for is not the only one using
mono-repositories, I just thought this might be a useful addition.

I completely understand if you disagree.

Seldaek:
Thanks for the detailed write-up.

However it is unfortunate to see that yet another fortune 500 corp is using our software for free and not content with that also demands that one of their own employees spend evening time rebasing branches because surely open source contributions could not happen during business hours.

It is also unfortunate that they are wasting your evenings due to a failure to recognize that multiple directories can be backed up at once, and that automation has to be setup manually instead of being itself automated.

Anyway I still think this doesn't belong in composer itself sorry, but I can offer to write that plug-in for you so you can stop using a custom composer build. I would happily do it for a small (for a fortune 500 corporation) fee.

As an alternative, I would suggest you look at setting up GitLab or GitHub and combine that with a [Private Packagist](https://packagist.com) instance to get a composer repository syncing up all your packages from it. All this would let you use multiple repos, can be setup on-premises, and would help fund the open source software you rely on as a bonus.

naderman:
Can't help myself but add that you seem to have a misunderstanding of what a monorepo is. The monorepos in use at many large but these days also smaller organizations specifically keep all code in a single tree versioned together and not in separate branches, to allow for attomic commits across all projects, see https://en.wikipedia.org/wiki/Monorepo and the referenced sources. What you describe is some odd technical way of storing multiple repositories in a single git repository by using disconnected graphs in a single repository.

PR #8802, opened 2020-04-17
Add "no-check-version" option to ValidateCommand

kubawerlos:
Field `version` can be a useful source of truth for packages that are not published on Packagist.

### DIFF
--- a/src/Composer/Util/ConfigValidator.php
+++ b/src/Composer/Util/ConfigValidator.php
@@ -40,10 +40,11 @@ class ConfigValidator
      *
      * @param string $file                       The path to the file
      * @param int    $arrayLoaderValidationFlags Flags for ArrayLoader validation
+     * @param bool   $checkVersion               Whether or not check if version field is present
      *
      * @return array a triple containing the errors, publishable errors, and warnings
      */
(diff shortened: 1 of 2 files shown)


src/Composer/Util/ConfigValidator.php
```diff
@@ -40,10 +40,11 @@ public function __construct(IOInterface $io)
      *
      * @param string $file                       The path to the file
      * @param int    $arrayLoaderValidationFlags Flags for ArrayLoader validation
+     * @param bool   $checkVersion               Whether or not check if version field is present
      *
      * @return array a triple containing the errors, publishable errors, and warnings
      */
-    public function validate($file, $arrayLoaderValidationFlags = ValidatingArrayLoader::CHECK_ALL)
+    public function validate($file, $arrayLoaderValidationFlags = ValidatingArrayLoader::CHECK_ALL, $checkVersion = true)
```
Seldaek, line 47: I'd rather have a $flags argument taking constants like the $arrayLoaderValidationFlags but for flags of the ConfigValidator. There would only be one flag now, but it's a more future proof API then.

kubawerlos: @Seldaek like this?

Seldaek: Yup, thanks!

PR #8803, opened 2020-04-19
Add a Composer\Versions class which is available in all projects at runtime to query installed packages/versions

Seldaek:
Open questions:

- Naming? Does `Composer\Versions` make sense? Are all public methods there sensibly named and understandable?
- Is `Versions::reload()` to allow projects like PHPUnit to integrate this without conflicting with the project they run in? Sounds potentially tricky but I couldn't think of a better solution right now. If someone has one please do tell.
- Is this going to be ok performance wise or can opcache caching of the array in `installed.php` be improved by dumping this in another way somehow? Would it be better to also include the array directly in the `Composer\Versions` class definition? /cc @nicolas-grekas

### DIFF
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -1,0 +1,186 @@
+<?php
+
+namespace Composer;
+
+use Composer\Semver\VersionParser;
+
+/**
+ * This class is copied in every Composer installed project and available to all
+ *
+ * To require it's presence, you can require `composer-runtime-api ^2.0`
+ */
+class InstalledVersions
+{
+    private static $installed;
+
+    /**
+     * Returns a list of all package names which are present, either by being installed, replaced or provided
+     *
+     * @return string[]
+     * @psalm-return list<string>
+     */
+    public static function getInstalledPackages()
+    {
+        return array_keys(self::$installed['versions']);
+    }
+
+    /**
+     * Checks whether the given package is installed
+     *
+     * This also returns true if the package name is provided or replaced by another package
+     *
+     * @param  string $packageName
+     * @return bool
+     */
+    public static function isInstalled($packageName)
+    {
+        return isset(self::$installed['versions'][$packageName]);
+    }
+
+    /**
+     * Checks whether the given package satisfies a version constraint
+     *
+     * e.g. If you want to know whether version 2.3+ of package foo/bar is installed, you would call:
+     *
+     *   Composer\InstalledVersions::satisfies(new VersionParser, 'foo/bar', '^2.3')
+     *
+     * @param VersionParser $parser      Install composer/semver to have access to this class and functionality
+     * @param string        $packageName
+     * @param ?string       $constraint  A version constraint to check for, if you pass one you have to make sure composer/semver is required by your package
+     *
+     * @return bool
+     */
+    public static function satisfies(VersionParser $parser, $packageName, $constraint)
+    {
+        $constraint = $parser->parseConstraints($constraint);
+        $provided = $parser->parseConstraints(self::getVersionRanges($packageName));
+
+        return $provided->matches($constraint);
+    }
+
+    /**
+     * Returns a version constraint representing all the range(s) which are installed for a given package
+     *
+     * It is easier to use this via isInstalled() with the $constraint argument if you need to check
+     * whether a given version of a package is installed, and not just whether it exists
+     *
+     * @param  string $packageName
+     * @return string Version constraint usable with composer/semver
+     */
+    public static function getVersionRanges($packageName)
+    {
+        if (!isset(self::$installed['versions'][$packageName])) {
+            throw new \OutOfBoundsException('Package "' . $packageName . '" is not installed');
+        }
+
+        $ranges = array();
+        if (isset(self::$installed['versions'][$packageName]['pretty_version'])) {
+            $ranges[] = self::$installed['versions'][$packageName]['pretty_version'];
+        }
+        if (array_key_exists('aliases', self::$installed['versions'][$packageName])) {
+            $ranges = array_merge($ranges, self::$installed['versions'][$packageName]['aliases']);
+        }
+        if (array_key_exists('replaced', self::$installed['versions'][$packageName])) {
+            $ranges = array_merge($ranges, self::$installed['versions'][$packageName]['replaced']);
+        }
+        if (array_key_exists('provided', self::$installed['versions'][$packageName])) {
+            $ranges = array_merge($ranges, self::$installed['versions'][$packageName]['provided']);
+        }
+
+        return implode(' || ', $ranges);
+    }
+
+    /**
+     * @param  string      $packageName
+     * @return string|null If the package is being replaced or provided but is not really installed, null will be returned as version, use satisfies or getVersionRanges if you need to know if a given version is present
+     */
+    public static function getVersion($packageName)
+    {
+        if (!isset(self::$installed['versions'][$packageName])) {
+            throw new \OutOfBoundsException('Package "' . $packageName . '" is not installed');
+        }
+
+        if (!isset(self::$installed['versions'][$packageName]['version'])) {
+            return null;
+        }
+
+        return self::$installed['versions'][$packageName]['version'];
+    }
+
+    /**
+     * @param  string      $packageName
+     * @return string|null If the package is being replaced or provided but is not really installed, null will be returned as version, use satisfies or getVersionRanges if you need to know if a given version is present
+     */
+    public static function getPrettyVersion($packageName)
+    {
+        if (!isset(self::$installed['versions'][$packageName])) {
+            throw new \OutOfBoundsException('Package "' . $packageName . '" is not installed');
+        }
+
+        if (!isset(self::$installed['versions'][$packageName]['pretty_version'])) {
+            return null;
+        }
+
+        return self::$installed['versions'][$packageName]['pretty_version'];
+    }
+
+    /**
+     * @param  string      $packageName
+     * @return string|null If the package is being replaced or provided but is not really installed, null will be returned as reference
+     */
+    public static function getReference($packageName)
+    {
+        if (!isset(self::$installed['versions'][$packageName])) {
+            throw new \OutOfBoundsException('Package "' . $packageName . '" is not installed');
+        }
+
+        if (!isset(self::$installed['versions'][$packageName]['reference'])) {
+            return null;
+        }
+
+        return self::$installed['versions'][$packageName]['reference'];
+    }
+
+    /**
+     * @return array
+     * @psalm-return array{name: string, version: string, reference: string, pretty_version: string, aliases: string[]}
+     */
+    public static function getRootPackage()
+    {
+        return self::$installed['root'];
+    }
+
+    /**
+     * Returns the raw installed.php data for custom implementations
+     *
+     * @return array[]
+     * @psalm-return array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[]}, versions: list<string, array{pretty_version: ?string, version: ?string, aliases: ?string[], reference: ?string, replaced: ?string[], provided: ?string[]}>}
+     */
+    public static function getRawData()
+    {
+        return self::$installed;
+    }
+
+    /**
+     * Lets you reload the static array from another file
+     *
+     * This is only useful for complex integrations in which a project needs to use
+     * this class but then also needs to execute another project's autoloader in process,
+     * and wants to ensure both projects have access to their version of installed.php.
+     *
+     * A typical case would be PHPUnit, where it would need to make sure it reads all
+     * the data it needs from this class, then call reload() with
+     * `require $CWD/vendor/composer/installed.php` (or similar) as input to make sure
+     * the project in which it runs can then also use this class safely, without
+     * interference between PHPUnit's dependencies and the project's dependencies.
+     *
+     * @param  array[] $data A vendor/composer/installed.php data set
+     * @return void
+     *
+     * @psalm-param array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[]}, versions: list<string, array{pretty_version: ?string, version: ?string, aliases: ?string[], reference: ?string, replaced: ?string[], provided: ?string[]}>} $data
+     */
+    public static function reload($data)
+    {
+        self::$installed = $data;
+    }
+}
--- a/src/Composer/Repository/FilesystemRepository.php
+++ b/src/Composer/Repository/FilesystemRepository.php
@@ -14,6 +14,8 @@ namespace Composer\Repository;
 
 use Composer\Json\JsonFile;
 use Composer\Package\Loader\ArrayLoader;
+use Composer\Package\RootPackageInterface;
+use Composer\Package\AliasPackage;
 use Composer\Package\Dumper\ArrayDumper;
 use Composer\Installer\InstallationManager;
 use Composer\Util\Filesystem;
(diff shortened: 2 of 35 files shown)


nicolas-grekas:
Would it make sense to add a method on the class loader instead? Could that replace the reload method, by having the instance in the autoloading stack report this info, for the scope it knows about?

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -105,5 +116,70 @@ public function write($devMode, InstallationManager $installationManager)
...
+            // add real installed packages
+            foreach ($packages as $package) {
+                if ($package instanceof AliasPackage) {
+                    continue;
+                }
+                $versions['versions'][$package->getName()] = array(
+                    'pretty_version' => $package->getPrettyVersion(),
+                    'version' => $package->getVersion(),
+                    'aliases' => array(),
+                    'reference' => $package->getSourceReference() ?: $package->getDistReference() ?: null,
```
aschempp, line 138: Should this check how the package was installed instead of picking source first if available (`$package->getInstallationSource()`)?

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+
+        return self::$installed['root'];
+    }
+
+    /**
+     * Returns the raw installed.php data for custom implementations
+     *
+     * @return array[]
+     */
+    public static function getAll()
```
aschempp, line 162: `Versions::getAll()` sounds like _getting all versions_ to me. Maybe it should rather be `getRawData()`?

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+    public static function reload($file)
+    {
+        self::$installed = require $file;
+    }
+
+    private static function initialize()
+    {
+        if (!self::$installed) {
+            if (basename(__DIR__) === 'Composer') {
+                throw new \LogicException('This class should only be used once dumped into vendor/composer/Versions.php');
```
aschempp, line 194: How can this ever not be true? Also, the class does not seem to be dumped (contrary to `ocramius/package-versions`)? Are we intentionally not checking if the file exists, but have `require` fail with a ambiguous message?

src/Composer/Repository/FilesystemRepository.php
Seldaek: Yup that probably makes sense will fix.

src/Composer/Versions.php
Seldaek: The class gets dumped into vendor/composer/Versions.php, so once it is there this is not true anymore and the class is usable. If it gets used directly from src/Composer/Versions.php then it'll fail to initialize so I wanted to make sure that doesn't happen.

staabm:
//cc @Ocramius as author of https://github.com/Ocramius/PackageVersions might bring in some useful input

nicolas-grekas:
> Not sure how the class loader is related to installed packages?

I think the autoloader is a good entry point yes. It has the appropriate scope. It could just be one method to get access to a Versions object.

> And/or would it dump the installed.php again on dump-autoload?

As best fits.

naderman:
`Composer\Versions` is nice and simple, but not sure if something like `Composer\InstalledVersions` or so would make it a bit clearer that this isn't a semver library but the way to access installed package info?

naderman:
I don't think this should be tied to the autoloader but remain separate, technically this can be used without the autoloader, and there is no reason to regenerate/re-dump this, when dumping the autoloader, it's rather directly tied to installed.json. I do wonder if there should be some way to work out if packages are dev requirements or if $devMode in any other way should influence the generated file?

Ocramius:
Would `Composer\Versions` have a meta-package that we can rely upon to have stable API guarantees? Something like `composer-plugin-api` comes in mind.

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+use Composer\Semver\VersionParser;
+
+class Versions
+{
+    private static $installed;
+
+    /**
+     * Returns a list of all package names which are present, either by being installed, replaced or provided
+     *
+     * @return string[]
```
Ocramius, line 14: What's the index structure of this?

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+    public static function getInstalledPackages()
+    {
+        self::initialize();
+
+        return array_keys(self::$installed['versions']);
+    }
+
+    /**
+     * @param  string  $packageName
+     * @param  ?string $constraint A version constraint to check for, if you pass one you have to make sure composer/semver is required by your package
```
Ocramius, line 25: Would split to a different method, such as `matchesConstraint`

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
+<?php
+
+namespace Composer;
+
+use Composer\Semver\VersionParser;
+
+class Versions
```
Ocramius, line 7: Maybe `Installed`? I didn't put too much thinking into naming on this, but since this is a mirror of `installed.json`, it makes sense to call it that way

src/Composer/Versions.php
naderman: It does get all installed versions? So can either be fixed by renaming the class to InstalledVersions or by calling it getAllInstalled?

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+        if (!isset(self::$installed['versions'][$packageName])) {
+            return false;
+        }
+
+        if ($constraint === null) {
+            return true;
+        }
+
+        if (!class_exists('Composer\Semver\VersionParser')) {
+            throw new \RuntimeException('The use of the $constraint argument requires that you install composer/semver');
```
Ocramius, line 41: Uhhhhhhh this part is really ugly. Possibly best to add `VersionParser` as a parameter to a new method here, so that it is up to the caller to ensure that the symbol exists (checked also via https://github.com/maglnet/ComposerRequireChecker)

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    /**
+     * Returns a version constraint representing all the range(s) which are installed for a given package
+     *
+     * It is easier to use this via isInstalled() with the $constraint argument if you need to check
+     * whether a given version of a package is installed, and not just whether it exists
+     *
+     * @param  string $packageName
+     * @return string
```
Ocramius, line 62: Is there a better representation than `string` here? Also: what would be the use-case here?

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+        if (!isset(self::$installed['versions'][$packageName]['version'])) {
+            return null;
+        }
+
+        return self::$installed['versions'][$packageName]['version'];
+    }
+
+    /**
+     * @param  string      $packageName
+     * @return string|null If the package is being replaced or provided but is not really installed, null will be returned as version, use getVersionRanges or isInstalled with a version argument if you need to know if a given version is present
```
Ocramius, line 110: Probably best to throw instead, and leave existence checks to `isInstalled()`

The code below already throws an `OutOfBoundsException` , so there's a confusing `string|null|throw` dichotomy

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+        if (array_key_exists('provided', self::$installed['versions'][$packageName])) {
+            $ranges = array_merge($ranges, self::$installed['versions'][$packageName]['provided']);
+        }
+
+        return implode(' || ', $ranges);
+    }
+
+    /**
+     * @param  string      $packageName
+     * @return string|null If the package is being replaced or provided but is not really installed, null will be returned as version, use getVersionRanges or isInstalled with a version argument if you need to know if a given version is present
```
Ocramius, line 91: Probably best to throw instead, and leave existence checks to `isInstalled()`

The code below already throws an `OutOfBoundsException` , so there's a confusing `string|null|throw` dichotomy

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+        if (!isset(self::$installed['versions'][$packageName]['pretty_version'])) {
+            return null;
+        }
+
+        return self::$installed['versions'][$packageName]['pretty_version'];
+    }
+
+    /**
+     * @param  string      $packageName
+     * @return string|null If the package is being replaced or provided but is not really installed, null will be returned as reference
```
Ocramius, line 129: Probably best to throw instead, and leave existence checks to `isInstalled()`

The code below already throws an `OutOfBoundsException` , so there's a confusing `string|null|throw` dichotomy

src/Composer/Versions.php
```diff
@@ -0,0 +1,200 @@
...
+    {
+        self::initialize();
+
+        return self::$installed['root'];
+    }
+
+    /**
+     * Returns the raw installed.php data for custom implementations
+     *
+     * @return array[]
```
Ocramius, line 160: This structure requires refinement: `array[]` is not really useful to consumers.

Maybe best to not expose it at all, if the structure cannot be guaranteed.

naderman:
Should probably have a separate composer-runtime-api to version all our generated code to be used in applications at runtime.

src/Composer/Versions.php
Ocramius: If this class is intended to be abstract, it should be `abstract` and have a `private final function __construct()`

src/Composer/Versions.php
Ocramius: Would it ever be autoloaded before being dumped? If not, moving it to something like `.stub` would prevent a PSR autoloader from accidentally picking it up.

src/Composer/Versions.php
naderman, line 91: :+1: on the null solution here

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -105,5 +116,70 @@ public function write($devMode, InstallationManager $installationManager)
         });
 
         $this->file->write($data);
+
+        if ($this->dumpVersions) {
+            $versions = array('versions' => array());
+            $packages = $this->getPackages();
+            $packages[] = $rootPackage = $this->rootPackage;
+            while ($rootPackage instanceof AliasPackage) {
+                $rootPackage = $rootPackage->getAliasOf();
```
naderman, line 125: I always get confused by this, but I think you need to watch out not to apply the aliases from composer.json, but only the root aliases actually in the lock file?

nicolas-grekas:
> I don't think this should be tied to the autoloader

I understand that. I was mainly looking for a way to make the API able to report when several autoloaders are stacked (aka several `vendors/`, like when phpunit runs).

Should the list of versions start empty, then as each autoloader is activated, the list is populated with what is available?

src/Composer/Versions.php
Seldaek: It technically returns ['root' => ..., 'versions' => ...] with all the data in there which is somewhat a mess, so I think getRawData is probably more appropriate.

src/Composer/Repository/FilesystemRepository.php
Seldaek: This only affects the root package, which alias' are actually always read from composer.json's extra. Other alias packages below are only those loaded from lock file.

src/Composer/Versions.php
Seldaek: `string[]` kinda implies a numerically indexed array no? I don't see much point in adding a psalm-return for `list<int, string>` which adds no information.

Ocramius: `string[]` is `array<string|int, string>`

Ocramius: Also, `array<int, string>` allows gaps, while `list<string>` doesn't (although only understood by psalm right now)

src/Composer/Versions.php
Seldaek: It is used by the test suite, so if renamed to .stub it would need to be required by the test class.. And technically it is usable as is, you just have to call `reload()` on it if you are not using it from a dumped state. I am thinking though that reload should just take an array instead of filename. Will fix that at least.

src/Composer/Repository/FilesystemRepository.php
naderman: I don't believe that to be true. Root aliases go into the alias section of the lock file I thought? We had a conversation about a lack of tests for this specifically the other day if you remember? I'm pretty sure that root aliases shouldn't just get updated on a composer install without modifying the lock file, since they have an impact on dependency resolution?

Seldaek: I meant the alias (as in branch-alias) of the root package, not the root composer.json-defined aliases. Those are indeed stored in lock file, and I am not sure how they persist here, I have to check if these get passed correctly into the installed repo or not.

Seldaek:
Ok done updating things here, thanks everyone for your feedback!

There is now a `composer-runtime-api` platform package one can depend on, I renamed to Composer\InstalledVersions and renamed some methods as well as documented a little more to hopefully make it as self-explanatory as possible.

Regarding loading @nicolas-grekas @aschempp it is not part of the autoloader now (except for the part adding it to the classmap so it's autoloadable). It gets the array dumped in the class itself in vendor/composer/InstalledVersions.php, and IMO it'll be the responsibility of either PHPUnit or projects to make sure they call `InstalledVersions::reload(require 'vendor/composer/installed.php)` if they need it to get the project data into it in case the PHPUnit one gets autoloaded (which will depend on autoload config, how things are bootstrapped and whether PHPUnit uses the class at all..). Note that PHPUnit here is just an example, any such CI tool is likely to be affected. I think we'll have to explore this later with real use cases.

If anyone has more feedback let me know.

src/Composer/InstalledVersions.php
```diff
@@ -0,0 +1,186 @@
...
+    /**
+     * Lets you reload the static array from another file
+     *
+     * This is only useful for complex integrations in which a project needs to use
+     * this class but then also needs to execute another project's autoloader in process,
+     * and wants to ensure both projects have access to their version of installed.php.
+     *
+     * A typical case would be PHPUnit, where it would need to make sure it reads all
+     * the data it needs from this class, then call reload() with
+     * $CWD/vendor/composer/installed.php as path (or similar) to make sure the project
```
naderman, line 173: the function doesn't take a path anymore, so maybe mention that you need to load it with `require $path` return value?

Ayesh:
I think this should only happen if the root or dependencies require `composer-runtime-api ^2`, right? I'm testing now, and the classmap seems to get `Composer\InstalledVersions` and regardless.

Seldaek:
@Ayesh yes it is always available. There are other features composer-runtime-api 2.0 brings in as per https://github.com/composer/composer/blob/master/doc/07-runtime.md so I don't think it's that meaningful to turn things on when this is required. One extra class in the classmap is fairly harmless.

Ayesh:
Thank you @Seldaek for the explanation.

I was wondering how this would work when there are multiple Composer autoloaders loaded at run-time (WordPress plugins seem to do it). I will do some tests and see. Thank you.

Seldaek:
In that case the first one to load the class wins, but that's why we have the `reload()` function there so you can kinda control this. In WP with n autoloaders this might turn into a mess where every plugin needs to call reload() before using the class to ensure they get the right data though. But that's a special use case, requiring a special solution.

emodric:
This is only usable when using Composer 2? If that is true, will it be backported to Comoposer 1?

Seldaek:
Nope it won't be backported. But hopefully Composer 2 is stable soon.

emodric:
That's too bad. Can't really rely on my clients or random people on internet to use Composer 2 to make this work. Requiring Composer 2 is also not a solution, since I can't know with what environments they work.

glensc:
Luckily 2.0 has the same PHP version requirement as 1.0:
- https://github.com/composer/composer/blob/a074819a5193e6cee802947eb9ad873a2b5fe63e/composer.json#L25

well, at least now. and this is very good, otherwise, 2.0 adoption could be poor.


which brings me to a question, can project or library indicate in their composer.json that they don't work with composer 1.0?

something like this into composer.json?
- require composer-runtime-api ^2.0

Seldaek:
@emodric as @glensc pointed out yes you can require `composer-runtime-api ^2.0` to make sure that that package will be installed by Composer 2, or at least not installed/selected if the user runs Composer 1. Then you can guarantee the new InstalledVersions class is present.

emodric:
Yes, I understand how it works. However, We don't want to limit the potential user base for our product on account of using a feature as trivial as showing the user currently installed version.

I guess we will still use `ocramius/package-versions` until there are some tangible metrics on Composer 2.0 adoption.

Seldaek:
That's completely fine. I would not advise migrating just yet.. I would however recommend using `composer/package-versions-deprecated` instead which forks and replaces `ocramius/package-versions`. It offers support for both Composer 1 and 2 so it gives your users a chance to migrate smoothly, and then whenever not needed anymore you can drop the dependency.

emodric:
We actually depend on `jean85/pretty-package-versions`, and it looks like it acts as a compatibility layer, requiring either `ocramius/package-versions`, the deprecated variant, or Composer v2, depending on version installed.

PR #8807, opened 2020-04-20
backport the legacy event api that allowed the substitue of rfs

elernonelma:
as found at https://getcomposer.org/apidoc/1.10.1/Composer/Plugin/PreFileDownloadEvent.html#method_setRemoteFilesystem

### DIFF
--- a/src/Composer/Plugin/PreFileDownloadEvent.php
+++ b/src/Composer/Plugin/PreFileDownloadEvent.php
@@ -55,6 +55,16 @@ class PreFileDownloadEvent extends Event
     }
 
     /**
+     * Sets the http downloader
+     *
+     * @param HttpDownloader $httpDownloader
+     */
+    public function setHttpDownloader(HttpDownloader $httpDownloader)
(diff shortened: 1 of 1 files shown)


Seldaek:
This is not going to be possible for v2 anymore, as the http downloader is doing async stuff we must have a unique instance so replacing it is not feasible.

We could perhaps add a setProcessedUrl so the URL can be modified and make sure that event dispatchers read the URL back before executing the request though?

That said.. it sounds like maybe building smth you can plug into HTTP_PROXY env var would have been easier than the mad hacks you had to pull to support flex and co there :)

elernonelma:
Right, I agree - I'm closing this and will try to provide something that way.

Thank you @Seldaek for your time and have a nice day!

Seldaek:
@elernonelma just in case, https://github.com/composer/composer/pull/8975 implements the setProcessedUrl I suggested above.

elernonelma:
Ok thanks!

Issue #8809, opened 2020-04-20
Expand context in PostFileDownloadEvent


heddn:
In https://github.com/composer/composer/issues/8654 we added a post file download event. I'm working to dog food that new event and noticed we are missing some glaring context details. Namely the package name. Working to review the interfaces and see if that is all we are missing.

PR #8810, opened 2020-04-20
#8809: expand context for post file download event

heddn:


Seldaek:
Thanks makes sense :)

heddn:
And for posterity, here's a PoC that leverages all of this: https://github.com/heddn/php-signify-composer-integration/pull/1

PR #8812, opened 2020-04-21
Optimize extension checks

nicolas-grekas:
Small optimization of the code generated after #8546
`extension_loaded()` can be optimized out by opcache so better use it.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -591,13 +591,13 @@ EOF;
                 }
 
                 if (preg_match('{^ext-(.+)$}iD', $link->getTarget(), $match)) {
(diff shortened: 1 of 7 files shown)


nicolas-grekas:
Also, I'm wondering: shouldn't all extension checks be skipped when ignore-platform-reqs is used? I use it quite often precisely to ignore missing extensions. The extra checks for extensions would break this workflow, isn't it?

stof:
To me, the whole platform check should be skipped when ignoring platform reqs (extensions but also the PHP version)

naderman:
Agree with skipping it entirely on ignore-platform-reqs

Seldaek:
I'll impl the skipping on --ignore-platform-reqs as I am anyway working on this. Thanks for the optimisation @nicolas-grekas !

PR #8821, opened 2020-04-22
Fix: Install ext-intl

localheinz:
This PR

* [x] installs `ext-intl` on GitHub Actions

Related to https://github.com/shivammathur/setup-php/issues/215.

PR #8822, opened 2020-04-22
Setup intl and xml extensions in GA workflow.

shivammathur:


Seldaek:
Note that https://github.com/composer/composer/pull/8821 is also doing this and on the correct branch so I'll close this PR. 

Thanks for sending a PR though, but IMO the issue I reported is still an issue because the default set of extension seems to differ per version and that's a little unexpected.

shivammathur:
> Note that #8821 is also doing this and on the correct branch so I'll close this PR.
> 
> Thanks for sending a PR though, but IMO the issue I reported is still an issue because the default set of extension seems to differ per version and that's a little unexpected.

Yes, I just noticed.

PR #8823, opened 2020-04-22
scripts: indicate how to see scripts-descriptions

glensc:
I've heard about the field but was not aware of how to see the descriptions.


glensc:
Thanks for improving it even more! :)

PR #8836, opened 2020-04-25
close #8784, Added --locked option to composer show command. Displaying locked packages with --all option

### LINKED ISSUE #8784: composer show --locked option
The composer show command should have a --locked option to view what is locked rather than what is currently locall installed.

Should also look at how this may interact or may need to interact with other options.

kbacia7:
Hey, In this pull request I added `--locked` option to `composer show`. I also added displaying locked packages when using `composer show --all`

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -66,6 +66,7 @@ class ShowCommand extends BaseCommand
                 new InputArgument('package', InputArgument::OPTIONAL, 'Package to inspect. Or a name including a wildcard (*) to filter lists of packages instead.'),
                 new InputArgument('version', InputArgument::OPTIONAL, 'Version or version constraint to inspect'),
                 new InputOption('all', null, InputOption::VALUE_NONE, 'List all packages'),
+                new InputOption('locked', null, InputOption::VALUE_NONE, 'List all locked packages'),
                 new InputOption('installed', 'i', InputOption::VALUE_NONE, 'List installed packages only (enabled by default, only present for BC).'),
                 new InputOption('platform', 'p', InputOption::VALUE_NONE, 'List platform packages only'),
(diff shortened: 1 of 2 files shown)


src/Composer/Command/ShowCommand.php
```diff
@@ -168,13 +170,27 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            } else {
+                $installedRepo = new InstalledRepository(array($localRepo, $platformRepo));
+            }
             $repos = new CompositeRepository(array_merge(array($installedRepo), $composer->getRepositoryManager()->getRepositories()));
         } elseif ($input->getOption('all')) {
             $defaultRepos = RepositoryFactory::defaultRepos($io);
             $io->writeError('No composer.json found in the current directory, showing available packages from ' . implode(', ', array_keys($defaultRepos)));
             $installedRepo = new InstalledRepository(array($platformRepo));
             $repos = new CompositeRepository(array_merge(array($installedRepo), $defaultRepos));
+        } elseif ($input->getOption('locked') && $composer) {
```
Seldaek, line 186: It should not check for $composer here IMO but rather below do `if (!$composer || !$composer->getLocker()->isLocked())`. Otherwise --locked will just silently be ignored which isn't great.

kbacia7: I'm not sure that I'm I understood correctly but I changed this.  Thanks for review

Seldaek: The point is that with the code the way you had it before, if I ran `composer show --locked` without composer.json it would just behave as if --locked was not passed, and show installed packages, which would be very confusing.

In reality though the getComposer() call at https://github.com/composer/composer/pull/8836/files#diff-b0c94b96dc86be178b130cd557412587R195 would have made it fail hard anyway as there was no composer.json, but I think this way is more explicit and not just working the intended way by chance.

Seldaek:
Thanks!

Issue #8833, opened 2020-04-24
PHP Fatal error:  Can't inherit abstract function Composer\DependencyResolver\Operation\OperationInterface::show() (previously declared abstract in Composer\DependencyResolver\Operation\SolverOperation) in phar:///var/www/html/composer.phar/src/Composer/DependencyResolver/Operation/SolverOperation.php on line 23 when run install


JulianaApolo:
My `composer.json`:

```json
{
    "name": "transfolha/wms",
    "description": "WMS do Transfolha",
    "type": "project",
    "license": "proprietary",
    "repositories": [
        { "type": "git", "url": "git@gitlab.corp.folha.com.br:folha/spiffy-framework.git" },
        { "type": "git", "url": "git@gitlab.corp.folha.com.br:folha/spiffy-boilerplate.git" },
        { "type": "git", "url": "git@gitlab.corp.folha.com.br:folha/spiffy-permission.git" }
    ],
    "autoload": {
        "classmap": [
            "src/common/classes/",
            "src/admin/wms.corp.transfolha.com.br/scripts/",
            "src/api/v1/scripts/"
        ]
    },
    "require": {
        "folha/spiffy-framework": "~1",
        "folha/spiffy-boilerplate": "~2",
        "folha/spiffy-permission": "~1",
        "fzaninotto/faker": "^1.7",
        "wobeto/php-barcode-generator": "^1.1"
    },
    "require-dev": {
        "phpunit/phpunit": "4.8.19",
        "squizlabs/php_codesniffer": "2.*"
    },
    "scripts": {
        "phpcbf": "vendor/bin/phpcbf --standard=PSR1,PSR2 --ignore=*/templates/* --extensions=php src"
    }
}
```

Output of `composer diagnose`:

```
Your version of PHP, 5.3.3, is affected by CVE-2013-6420 and cannot safely perform certificate validation, we strongly suggest you upgrade.
The "folha/composer-plugin-build" plugin was skipped because it requires a Plugin API version ("^1.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.
The "hirak/prestissimo" plugin was skipped because it requires a Plugin API version ("^1.0.0") that does not match your Composer installation ("2.0.0"). You may need to run composer update with the "--no-plugins" option.
[...]

PR #8837, opened 2020-04-25
Provide default implementation for abstract method in interface for PHP < 5.3.9

naderman:
Fixes https://github.com/composer/composer/issues/8833


stof:
Another option would be to remove the abstract method entirely (relying on the abstract method coming from the interface). This way, even PHP would still enforce overriding it.

naderman:
@stof Ah I didn't realize it was possible to implement an interface in an abstract class without specifically marking the unimplemented methods abstract.

naderman:
Done in https://github.com/composer/composer/commit/56edd53046fd697d32b2fd2fbaf45af5d7951671

PR #8843, opened 2020-04-28
Add (default) base name of archive into composer.json

grv87:
`--file` argument for `archive` command, introduced in #4397 and #4479, is great, but:

* Written by hands, it's just tedious
* Used from CI, command becomes non-reusable. You can't have a single command to build any package

I propose to add `name` property into `archive` section of `composer.json` and use it for base name for archive, if `--file` argument was not provided.
*Base* means that Composer still concatenates it with version (and this is not configurable).

I haven't write any tests yet. They are probably required, but I'm not sure on the test strategy.
Let's discuss.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -883,6 +883,21 @@ A set of options for creating package archives.
 
 The following options are supported:
 
+* **name:** Allows configuring base name for archive.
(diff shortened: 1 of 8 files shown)


Seldaek:
Looks good to me, thanks. Will be only available in 2.0 though.

PR #8845, opened 2020-04-29
Use compiled constraint in Composer

jderusse:
This is draft of integrating compiled constraint in Composer:

https://github.com/composer/semver/pull/86.

I Probably need help to find the right way to inject the `Compiledatcher` with the right cache dir configured.

Note: I did not replace ALL  occurrences of `->matches()` but just the one that are the most often called.

### DIFF
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -14,6 +14,7 @@ namespace Composer\DependencyResolver;
 
 use Composer\Package\AliasPackage;
 use Composer\Package\Version\VersionParser;
+use Composer\Semver\CompilingMatcher;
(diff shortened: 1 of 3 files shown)


Seldaek:
@jderusse can you rebase this on master without all the commits from https://github.com/composer/composer/pull/8850? Your commit seems fine to me as a starting point at least.

composer.json
```diff
@@ -24,7 +24,7 @@
     "require": {
         "php": "^5.3.2 || ^7.0",
         "composer/ca-bundle": "^1.0",
-        "composer/semver": "^2.0",
+        "composer/semver": "^2.1.0",
```
GrahamCampbell, line 27: just `^2.1` is fine

jderusse:
closed in favor of #8926

PR #8850, opened 2020-04-29
Filter dependent packages early

Toflar:
This is a follow up of the work I did in https://github.com/composer/composer/pull/8275 and @ausi tried to extend even more in https://github.com/composer/composer/pull/8295.
With the new `PoolBuilder` things have changed quite a lot since those attempts so here's another one.

What happens currently is that early package filtering happens only for root dependencies. So the dependent packages are not filtered at all.
Let's take an example to illustrate: Let's take the package `api-platform/api-pack` in version `^1.2`.

Because `api-platform/api-pack` is a root requirement, all the package versions `<1.2.0.0` are not considered. Their information is downloaded from Packagist but it doesn't make it into the `Pool` of rules that are considered.
However, all the packages that `api-platform/api-pack` requires (which are quite a lot) are not filtered. For these packages, **all** versions are loaded into the dependency resolving process.

This PR tries to improve that by remembering the last constraint of a certain package that was loaded and if things do not match, extend that and load it again. Example:

1. `api-platform/api-pack` requires `doctrine/orm` with `^2.4.5`
2. with this PR, only the versions of `doctrine/orm` that match `^2.4.5` are loaded!
3. if any other package was to require `doctrine/orm` in e.g. `^2.6`, nothing happens because `^2.4.5` also matches `^2.6`.
4. if any other package was to require `doctrine/orm` in e.g. `^2.3`, we turn the original constraint from `^2.4.5` into `^2.3 || ^2.4.5` and mark the package as not loaded so eventually, the missing packages will make it into the resolving process as well :)

Thanks to my previous work on the `MultiConstraint::create()` in `composer/semver` and @jderusse's latest work, we can optimize these constraints quite a bit along the way which should result in even better numbers :)

This optimization should improve all kinds of combinations but it will **massively** improve projects that depend on meta packages such as the `api-platform/api-pack` example.
Here are some tests:

For `"api-platform/api-pack": "^1.2"`:

|  | 1.10.5 | master@56edd5304 | this PR |
|------|----------:|-----------------:|--------:|
| RAM | 2054.08MB | 882.87MB | 97.61MB |
| Time | 64.48s | 27.17s | 3.65s |

For `"contao/manager-bundle": "^4.9"`:

|  | 1.10.5 | master@56edd5304 | this PR |
|------|---------:|-----------------:|---------:|
| RAM | 1999.2MB | 1170.1MB | 198.74MB |
| Time | 137.71s | 66.57s | 10.18s |

I'm not going to comment on the numbers, they should speak for themselves :)

I did not encounter any issues except for one project where it ended in a non-resolvable set of dependencies. But for that project, dependencies are so complex, it's very hard to debug. I guess there's just a use case I've not considered yet. My hope is that maybe someone else will have the same issue with a simpler dependency layout so debugging might become easier ;)


src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -92,13 +91,11 @@ public function buildPool(array $repositories, Request $request)
 
         $loadNames = array();
         foreach ($request->getFixedPackages() as $package) {
-            $this->nameConstraints[$package->getName()] = null;
-            $this->loadedNames[$package->getName()] = true;
+            $this->loadedNames[$package->getName()] = new Constraint('==', $package->getVersion());
```
stof, line 94: Be careful about that. If we have fixed packages, we should never expand the constraint and mark them as unloaded, otherwise we remove the optimization of partial updates. So you probably need to register an EmptyConstraint here (matching all) rather than an exact-match constraint.

src/Composer/DependencyResolver/Request.php
```diff
@@ -55,7 +56,7 @@ public function __construct(LockArrayRepository $lockedRepository = null)
     public function requireName($packageName, ConstraintInterface $constraint = null)
     {
         $packageName = strtolower($packageName);
-        $this->requires[$packageName] = $constraint;
+        $this->requires[$packageName] = $constraint ? $constraint : new EmptyConstraint();
```
stof, line 59: use `?:`

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -259,8 +246,10 @@ private function loadPackage(Request $request, PackageInterface $package, $propa
         $loadNames = array();
         foreach ($package->getRequires() as $link) {
             $require = $link->getTarget();
+            $linkConstraint = $link->getConstraint();
+
             if (!isset($this->loadedNames[$require])) {
-                $loadNames[$require] = null;
+                $loadNames[$require] = $linkConstraint;
```
stof, line 252: there is still a case seeting `$loadNames[$require]` to `null` a few lines below.

src/Composer/DependencyResolver/PoolBuilder.php
naderman, line 94: Since these packages are fixed, shouldn't this actually assign an EmptyConstraint, so this never attempts to load a different version later on, that's kind of the point here.

When you do benchmarks, try also benchmarking "composer update mirrors" and "composer update <single-package>", to see if your changes negatively impact partial updates.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -92,13 +91,11 @@ public function buildPool(array $repositories, Request $request)
...
         foreach ($request->getFixedPackages() as $package) {
-            $this->nameConstraints[$package->getName()] = null;
-            $this->loadedNames[$package->getName()] = true;
+            $this->loadedNames[$package->getName()] = new Constraint('==', $package->getVersion());
 
             // replace means conflict, so if a fixed package replaces a name, no need to load that one, packages would conflict anyways
             foreach ($package->getReplaces() as $link) {
-                $this->nameConstraints[$package->getName()] = null;
-                $this->loadedNames[$link->getTarget()] = true;
+                $this->loadedNames[$link->getTarget()] = $link->getConstraint();
```
naderman, line 98: I think this should also be EmptyConstraint, since we really do not need to load any versions at all because they would all conflict, even if only specific versions are actually getting replaced, you can't install them at the same time as another version.

naderman:
This is looking like a great start. I think this really needs to come with a suite of tests for the pool builder to check which packages are loaded giving a certain set of inputs. Any bug that leads to unnecessary loading too many packages will be undetectable elsewhere since the solver will just pick the right versions anyway. So really the only way to make sure this doesn't break in the future is to add tests for which things should or should not be in the pool in a given scenario.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -122,7 +119,6 @@ public function buildPool(array $repositories, Request $request)
             }
 
             $loadNames[$packageName] = $constraint;
-            $this->nameConstraints[$packageName] = $constraint && !($constraint instanceof EmptyConstraint) ? array($constraint) : null;
         }
 
         // clean up loadNames for anything we manually marked loaded above
```
naderman, line 124: Does this clean up now actually lead to bugs because loadNames contents may actually differ from the constraint in `$this->loadedNames` here? Need to come up with a test case that validates this.

src/Composer/DependencyResolver/PoolBuilder.php
naderman: Added a test case which fails because of that now.

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: I found this `$loadNames` thing to be very confusing so I tried to cleanup a bit in https://github.com/composer/composer/pull/8850/commits/4ed9cbdedb6c7663c882abf0e1ce2bf306b49efc

naderman: I think the idea was to avoid the word "package" in that context, cause we're really loading versions of packages by that name or something like that ;-)

src/Composer/Package/Link.php
```diff
@@ -59,7 +60,7 @@ public function __construct($source, $target, ConstraintInterface $constraint =
     {
         $this->source = strtolower($source);
         $this->target = strtolower($target);
-        $this->constraint = $constraint;
+        $this->constraint = $constraint ? $constraint : new EmptyConstraint();
```
naderman, line 63: I'm still not convinced this is even the right place to do this, shouldn't this just never receive null then, if we really need to make this change?

src/Composer/DependencyResolver/Request.php
naderman: Why did you resolve this?

tests/Composer/Test/DependencyResolver/SolverTest.php
```diff
@@ -652,9 +652,9 @@ public function testConflictResultEmpty()
 
             $msg = "\n";
             $msg .= "  Problem 1\n";
-            $msg .= "    - Root composer.json requires a -> satisfiable by A[1.0].\n";
+            $msg .= "    - Root composer.json requires a [] -> satisfiable by A[1.0].\n";
```
naderman, line 655: Please undo this change? Why would this display a set of empty brackets?

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/package-versions-are-not-loaded-if-not-required-simple.test
```diff
@@ -0,0 +1,26 @@
+--TEST--
+Test irrelevant package versions are not loaded
```
naderman, line 2: I would probably just delete this test, since this is fully covered by package-versions-are-not-loaded-if-not-required-recursive.test as well, and more test do slow down the suite ;-)

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/package-versions-are-not-loaded-if-not-required-expansion.test
```diff
@@ -0,0 +1,34 @@
...
+]
+
+--EXPECT--
+[
+    "root/req-1.0.0.0",
+    "dep/dep-2.3.4.0",
+    "dep/dep-2.3.5.0",
+    "dep/dep2-2.3.4.0",
+    "dep/dep-2.3.4.0",
+    "dep/dep-2.3.5.0",
```
naderman, line 32: Why is this package in that exact version in here twice? That seems like a bug?

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -235,9 +237,11 @@ private function loadPackagesMarkedForLoading(Request $request, $repositories)
         }
 
         // Make sure we empty the packagesToLoad here as it would result
-        // in an endless loop with non-existent packages for example
-        // TODO: fixme, this should only happen if it's not a new package
-        // $this->packagesToLoad = array();
+        // in an endless loop if the array remains unchanged.
+        // This could happen with non-existent packages for example.
+        if ($preLoad == $this->packagesToLoad) {
```
naderman, line 242: This is kind of why I think reusing the same variable across loops is bad? That's why previously there was a $newLoadNames, that seemed a lot cleaner?

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -43,7 +44,7 @@ class PoolBuilder
     private $io;
 
     private $aliasMap = array();
-    private $nameConstraints = array();
+    private $packagesToLoad = array();
```
naderman, line 47: I think this should still be called something with "name" in it, rather than just packages. Additionally the name of this variable should be somewhat symmetrical to `$loadedNames` which it ends up being afterwards?

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -210,14 +172,78 @@ public function buildPool(array $repositories, Request $request)
...
+        }
+
+        // No need to load this package with this constraint because it was
+        // already loaded in one that matches
+        if ($this->loadedNames[$name]->matches($constraint)) {
+            return;
+        }
+
+        // Maybe it was already marked before but not loaded yet. In that case
+        // we have to extend the constraint (we don't check if the match because
```
Tobion, line 197: typo: if they match

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -210,14 +172,78 @@ public function buildPool(array $repositories, Request $request)
...
+        unset($this->loadedNames[$name]);
+    }
+
+    private function loadPackagesMarkedForLoading(Request $request, $repositories)
+    {
+        foreach ($this->packagesToLoad as $name => $constraint) {
+            $this->loadedNames[$name] = $constraint;
+        }
+
+        $preLoad = $this->packagesToLoad;
```
Seldaek, line 216: I think here you should set `$this->packagesToLoad = array();`, then within the foreach below you'd only use $preLoad, which I guess I'd call $packageBatch or something. Only unset from that, check if it's empty, read from it, etc. Then as you do that and loadPackages the packagesToLoad property gets filled again, once foreach is done it exits the function and re-enters it due to the while loop at line 127.

That would also let you get rid of the check at line 242.

And as a bonus, it means you avoid having bugs whereas while loading packages from the 2nd repo, loadPackage queues up a reload of name $foo inside $this->packagesToLoad, which was e.g. already loaded from first repo, then suddenly your foreach reaches the 3rd repo and as it reads from $this->packagesToLoad, it tries to load $foo from that 3rd repo without trying it again from the top, and you end up loading the wrong thing from the wrong place.

src/Composer/Package/Link.php
Seldaek: Yup would make this non-nullable and fix in ArrayLoader

tests/Composer/Test/DependencyResolver/SolverTest.php
Seldaek: This is due to the adding of EmptyConstraints instead of null to links IMO, EmptyConstraints without prettyConstraint will stringify to `[]`. Maybe we should change this to `*`?

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/package-versions-are-not-loaded-if-not-required-expansion.test
Toflar: Fixed the test in 1c9a93ea0db868f8da8f41542f499d867a5685f6.

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/package-versions-are-not-loaded-if-not-required-simple.test
Toflar: Agree, removed in 8dcb1eb2c4e457381b5c047102c07b91f037971c

tests/Composer/Test/DependencyResolver/SolverTest.php
Toflar: Done in 1bd525d52583f5e436dff4ee7b1b77414bb733f4.

src/Composer/Package/Link.php
Toflar: And in a few other places ;) Done in 09955a2494335fc88f5ad6294c9f6ac39b4772d2.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -43,7 +44,7 @@ class PoolBuilder
     private $io;
 
     private $aliasMap = array();
-    private $nameConstraints = array();
+    private $packagesToLoad = array();
     private $loadedNames = array();
```
staabm, line 48: Just to make sure we dont mix up types

```suggestion
    /** @psalm-var array<string, Constraint>
    private $loadedNames = array();
```

src/Composer/DependencyResolver/PoolBuilder.php
staabm, line 47:
```suggestion
    /** @psalm-var array<string, ConstraintInterface>
    private $packagesToLoad = array();
```

Toflar:
This PR is on hold because it cannot be finished until https://github.com/composer/semver/issues/94 is resolved.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -238,17 +201,91 @@ public function buildPool(array $repositories, Request $request)
...
+        // a subset of the constraint with which we have already loaded packages
+        if ($constraint->isSubsetOf($this->loadedNames[$name])) {
+            return;
+        }
+
+        // Maybe it was already marked before but not loaded yet. In that case
+        // we have to extend the constraint (we don't check if they match because
+        // MultiConstraint::create() will optimize anyway
+        if (isset($this->packagesToLoad[$name])) {
+            $constraint = MultiConstraint::create(array($this->packagesToLoad[$name], $constraint), false);
```
Seldaek, line 242: I wonder if it wouldn't be more efficient here to be able to quickly deduplicate constraint by doing the equivalent of what `$nameConstraints[$name] = MultiConstraint::create(array_values(array_unique($constraints)), false);` used to do in the code you deleted.

i.e. here only append the new constraint to an array of constraints for $name, possibly only appending if it doesn't exist in the array yet.

Then above to call isSubsetOf you'd MultiConstraint::create a temporary one with the current constraints. That way you avoid creating a nested multiconstraint monstrosity by recreating them into each other in a loop. 

If this doesn't make any sense to you I can also try it myself sometime, not sure if I'm making my point clearly..

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: Done in 35fe5b01b7197d38c01d15cd7e9f761e1651faa9.

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: Yeah I guess it's easier if you just try with code directly rather than describing what to code in English :D

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: "names" is just too broad. I like being explicit but I agree that we're not loading packages but package names.
`packageNamesScheduledForLoading` is a bit long. Proposals?

naderman: Well the more I think about it, the more this does specify packages now that it has a set of constraints too? So how about `$packagesToLoad` and `$loadedPackages` then? Or is your problem that you find "toLoad" and "loaded" hard to distinguish?

Toflar: Renamed in 0c974e55a77dba3d9f9cb4aa3d0b324694cf0cf7.

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: That's done differently now, so I'm resolving this.

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: It's now only merged if it's not a subset. Not sure if that resolves your remark?

src/Composer/DependencyResolver/PoolBuilder.php
stof: well, a package is not name+constraint but name+version. This still specify names to load.

naderman: It specifies a set of packages to load using a name and a constraint, hence plural ;-)

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -238,17 +206,85 @@ public function buildPool(array $repositories, Request $request)
             $this->unacceptableFixedPackages = $prePoolCreateEvent->getUnacceptableFixedPackages();
         }
 
+        // Filter duplicate packages
```
naderman, line 209: Why are there duplicates in the first place? Isn't this an indicator stuff is going wrong elsewhere?

Toflar: I'm not quite sure. Packages just get added to `$this->packages` in `loadPackage()` right away. I think we had duplicates before as well and the `$this->nameConstraints` took care of removing them. I've added a todo so we can look at this later in 6917387f4f13a91509614f772e554768b6b921c7.

naderman: I'm not sure what it did before. But I have a feeling that your filter mechanism would filter out duplicates if they come from multiple repos, and I'm not sure the package that stays would be the one from the higher prio (requires that you actually define an additional repo with canonical: false)

naderman:
Reminder for something to look into later: We take some information from the latest version of a package, e.g. abandoned flag. Need to check this still works if we now only partially load versions per package.

stof:
@naderman shouldn't Packagist mark all versions as abandoned instead ?

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -216,19 +218,28 @@ public function buildPool(array $repositories, Request $request)
...
     {
         // Maybe it was already marked before but not loaded yet. In that case
         // we have to extend the constraint (we don't check if they match because
         // MultiConstraint::create() will optimize anyway)
-        if (isset($this->packagesToLoad[$name]) && !Intervals::isSubsetOf($constraint, $this->packagesToLoad[$name])) {
+        if (isset($this->packagesToLoad[$name])) {
+            // Already marked for loading and this does not expand the constraint to be loaded, nothing to do
+            if (Intervals::isSubsetOf($constraint, $this->packagesToLoad[$name])) {
+                return;
+            }
```
Seldaek, line 261: This was the main bug left in your implementation @Toflar - if you required for example `1.*` then `1.0.*` of the same package, it would first time set packagesToLoad to `1.*` then second time as `1.0.*` was a subset it skipped this block and went straight to line 233 (in old code, or 244 in new) below, overwriting the constraint with a narrower one.

Seldaek:
Tests are fixed now. Last commit also fixed a couple issues due to extending the loaded packages beyond the root constraints, which makes no sense as the root package requirements always have to be fulfilled. 

So https://github.com/composer/composer/pull/8850/commits/2fa58ccf963ca5f4f0e059d0be238d8f4226e519 now forces it to load the root constraint at most (for unfixed packages, it'll load the least possible, up to a max of the root constraint, for non fixed ones it'll just load the whole root constraint at the beginning).

Last thing I'd like to do is try to optimize the loading of ComposerRepository to avoid it reloading the same packages over and over. Won't have time for this today though. Not sure if more PoolBuilder tests are needed or if we're happy with the current set? @naderman?

@Toflar can you make it run on those projects where you got errors at some point to see if it behaves normally now?

naderman:
@Seldaek there were definitely still tests missing, did you consider all the situations regarding replacers matching the root requirements when making the change you described (haven't looked at code changes yet)

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: Ah, right - stupid me 🤦‍♂️

Toflar:
> @Toflar can you make it run on those projects where you got errors at some point to see if it behaves normally now?

Yep, works normally now! All three runs return the same output (I mean, would update to the same package combination). Here are the profile results on a pretty huge project (248 packages):

| Branch                  | Results                                                  |
|-------------------------|----------------------------------------------------------|
| 1.10                    | Memory usage: 562.19MiB (peak: 2486.33MiB), time: 69.29s |
| master@9d5051f          | Memory usage: 91.17MiB (peak: 1626.48MiB), time: 44.13s  |
| filter-packages@2fa58cc | Memory usage: 63.3MiB (peak: 1033.55MiB), time: 28.87s    |

🎉

Seldaek:
Ok all tests are back to green, rebased on latest master, ready to merge IMO..

It would be great if we can get more people to test this on various projects though to make sure it does not regress in any major way. From our limited testing it seems to have benefits in some cases but not all (which is fine).

## You can help test this PR like this:

- `cd` into a project you want to try out with
- Download the phar file for this branch: `wget https://seld.be/composer-test.phar`
- Run `composer self-update --snapshot` to get the latest 2.0 dev build
- Run `composer update --dry-run` to prime the cache
- Run `COMPOSER_DISABLE_NETWORK=1 composer update -v --dry-run --profile` to run a profile without network jitter, with current master branch
- Run `COMPOSER_DISABLE_NETWORK=1 php composer-test.phar update -v --dry-run --profile` to run another profile with the optimizations in this branch
- Make sure the results are the same 👀 e.g. the `Lock file operations: 3 installs, 61 updates, 0 removals` lines in both runs should have similar numbers.
- Report results here 📝 We're mostly interested in the last lines of each profile run, e.g. `Memory usage: 90.53MiB (peak: 1213.05MiB), time: 19.38s`, but if you spot a discrepancy in what gets updated please do tell as that would be bad.
- `Unsubscribe` from notifications on the top right of this page if you don't want to get spammed with everyone else's results 😃

If you're interested in more details, you can look at the `Analyzed x packages/rules` lines which show up somewhere on top to see how many packages and inter-package dependency rules were saved by this PR.

Thanks! 🚀

Bacteries:
First : 
> Memory usage: 49.57MiB (peak: 804.99MiB), time: 14.81s

Second : 
> Memory usage: 71.03MiB (peak: 804.9MiB), time: 14.68s

And for each it was the same changes : _Package operations: 3 installs, 31 updates, 5 removals_

rivimey:
Just tried against a drupal 8 setup in need of updates. The composer.json includes "cwegans/composer-patches" which fails to resolve for composer 2.

Running:
    COMPOSER_DISABLE_NETWORK=1 composer update -v --dry-run --profile
uses my 1.0-stable composer:

`[10.2MiB/0.01s] Loading composer repositories with package information
[10.7MiB/0.58s] Updating dependencies (including require-dev)
[545.1MiB/17.64s] Dependency resolution completed in 0.450 seconds
[363.0MiB/17.81s] Analyzed 14313 packages to resolve dependencies
[363.0MiB/17.81s] Analyzed 317223 rules to resolve dependencies
[364.9MiB/17.86s] Dependency resolution completed in 0.001 seconds
[360.4MiB/17.86s] 48 packages you are using are looking for funding.
[360.4MiB/17.86s] Use the composer fund command to find out more!
[360.4MiB/17.87s] Memory usage: 360.43MiB (peak: 545.36MiB), time: 17.87s
`

Then running with composer-test.phar:

`[12.2MiB/0.01s] Loading composer repositories with package information
[12.4MiB/0.01s] Updating dependencies
[12.6MiB/0.22s] https://packages.drupal.org/8 could not be fully loaded (Network disabled, request canceled: https://packages.drupal.org/8/packages.json), package information was loaded from the local cache and may be out of date
[20.9MiB/0.47s] https://repo.packagist.org could not be fully loaded (Network disabled, request canceled: https://repo.packagist.org/packages.json), package information was loaded from the local cache and may be out of date
[468.9MiB/4.46s] Dependency resolution completed in 0.001 seconds
[469.0MiB/4.46s] Your requirements could not be resolved to an installable set of packages.
[469.0MiB/4.46s]
  Problem 1
    - drupal-composer/drupal-scaffold[2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.6.0, 2.6.1] require composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - Root composer.json requires drupal-composer/drupal-scaffold ^2.5 -> satisfiable by drupal-composer/drupal-scaffold[2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.5.4, 2.6.0, 2.6.1].
`

1ma:
```
composer 2.0-dev (2.0-dev+c3028c02d95bd68a0a1bcc035d7d75561c916d82) 2020-07-17 11:18:02
Package operations: 2 installs, 18 updates, 0 removals
Memory usage: 36.62MiB (peak: 283.48MiB), time: 7.81s

composer-test.phar (commit 071350286fb196722728511aecd00e1defa826c7)
Package operations: 2 installs, 18 updates, 0 removals (same)
Memory usage: 37.95MiB (peak: 283.52MiB), time: 7.76s
```

navarr:
snapshot: 
```
Package operations: 1 install, 10 updates, 76 removals
Memory usage: 57.89MiB (peak: 631.16MiB), time: 17.33s
```

test:
```
Package operations: 1 install, 10 updates, 76 removals
Memory usage: 81.69MiB (peak: 598.12MiB), time: 17.75s
```

Use case: Magento installation with replacements and path repository usage

deresh:
snapshot

`Package operations: 5 installs, 74 updates, 1 removal`
`[54.6MiB/8.14s] Memory usage: 54.6MiB (peak: 888.81MiB), time: 8.14s`

test

`Package operations: 5 installs, 74 updates, 1 removal`
`[55.9MiB/7.86s] Memory usage: 55.87MiB (peak: 847.81MiB), time: 7.86s`

case: custom symfony 5 app

Levivb:
Use case: werkenbijdefensie.nl requiring ±2200MB with composer v1.10

snapshot:
```
[124.1MiB/7.81s] Lock file operations: 2 installs, 18 updates, 0 removals
[59.4MiB/7.88s] Memory usage: 59.37MiB (peak: 361.31MiB), time: 7.88s
```

test:
```
[99.2MiB/7.49s] Lock file operations: 2 installs, 18 updates, 0 removals
[74.1MiB/7.54s] Memory usage: 74.08MiB (peak: 336.22MiB), time: 7.54s
```

Composer v1.10.9
```
[707.2MiB/97.99s] Package operations: 2 installs, 32 updates, 0 removals
[698.2MiB/98.00s] Memory usage: 698.15MiB (peak: 2181.61MiB), time: 98s
```

----------------------------------------
----------------------------------------

And another project with loads of packages:
snapshot:
```
[96.0MiB/2.91s] Lock file operations: 259 installs, 0 updates, 0 removals
[44.9MiB/2.93s] Memory usage: 44.88MiB (peak: 260.76MiB), time: 2.93s
```

test
```
[77.9MiB/2.83s] Lock file operations: 259 installs, 0 updates, 0 removals
[57.6MiB/2.89s] Memory usage: 57.58MiB (peak: 242.46MiB), time: 2.89s
```

HDVinnie:
FIrst: 
`Lock file operations: 0 installs, 1 update, 5 removals`
`Memory usage: 30.32MiB (peak: 383.13MiB), time: 6.21s`

Second:
`Lock file operations: 0 installs, 1 update, 5 removals`
`Memory usage: 31.17MiB (peak: 357.29MiB), time: 6.19s`

src/Composer/DependencyResolver/Request.php
```diff
@@ -55,7 +56,7 @@ public function __construct(LockArrayRepository $lockedRepository = null)
     public function requireName($packageName, ConstraintInterface $constraint = null)
     {
         $packageName = strtolower($packageName);
-        $this->requires[$packageName] = $constraint;
+        $this->requires[$packageName] = $constraint ? $constraint : new MatchAllConstraint();
```
stof, line 59: you should use `?:` here

GrahamCampbell: Even better, don't do that, and compare against null explicitly in the boolean expression.

ryanaslett:
Drupal 9.1.x + drush + commerce + webform:
Snapshot: 
[75.8MiB/4.09s] Package operations: 3 installs, 17 updates, 0 removals
[45.5MiB/4.10s] Memory usage: 45.45MiB (peak: 415.35MiB), time: 4.1s

composer-test.phar
[67.7MiB/3.98s] Package operations: 3 installs, 17 updates, 0 removals
[53.8MiB/3.99s] Memory usage: 53.81MiB (peak: 407.1MiB), time: 3.99s

composer 1.16.9
[476.4MiB/29.19s] Package operations: 3 installs, 17 updates, 0 removals
[472.1MiB/29.20s] Memory usage: 472.12MiB (peak: 1356.02MiB), time: 29.2s

This is _very_ exciting. We're sprinting today at drupalconGlobal2020 on supporting composer 2 for packages.drupal.org.

fritzmg:
Contao 4.9.4, all core bundles plus some extensions **and** `symfony/flex` plugin (restricting to `symfony/symfony: >=4.4`):

First run (current snapshot), average of 4 runs:

```
Memory usage: 24.84MiB (peak: 95.51MiB), time: 3.52s
``` 

Second run (this branch), average of 4 runs:

```
Memory usage: 66.13MiB (peak: 93.76MiB), time: 2.55s
```

---

Contao 4.9.4, all core bundles plus some extensions, **no** `symfony/flex` plugin:

First run (current snapshot):

```
Memory usage: 88.49MiB (peak: 1555.31MiB), time: 76.05s
``` 

Second run (this branch):

```
Memory usage: 97.14MiB (peak: 948.36MiB), time: 30.3s
```

Toflar:
We‘re not expecting huge gains for „regular“ projects here. This PR should benefit projects that require a little amount of packages in their root composer.json which themselves require the lion‘s share of packages (aka transitive deps). So I‘m expecting noticeable gains for people requiring e.g. Symfony „packs“ etc.
For all the other projects the goal is to not have a negative impact which seems to work out so far😎💪

corphi:
For a legacy Symfony bundle with rather loose restrictions on dependencies:

Composer Version  |  Package operations  |  Profile
-------|------|-----
1.10.9  |  17 installs, 75 updates, 3 removals  |  Memory usage: 725.73MiB (peak: 5241.15MiB), time: 116.75s
2.0.0-alpha2  |  17 installs, 76 updates, 3 removals  |  Memory usage: 61.71MiB (peak: 859.12MiB), time: 15.92s
2.0.0-snapshot 750a92b4b7aecda0e5b2f9b963f1cb1421900675  |  17 installs, 76 updates, 3 removals  |  Memory usage: 61.64MiB (peak: 858.99MiB), time: 15.58s
`composer-test.phar`  | 17 installs, 76 updates, 3 removals  |  Memory usage: 82.62MiB (peak: 858.44MiB), time: 15.9s

The difference in package updates stems from a dependency on composer.

alexander-schranz:
Tested this against the [sulu-demo](https://github.com/sulu/sulu-demo/):

First:

```bas
Memory usage: 42.78MiB (peak: 523.56MiB), time: 11.3s
```

Second:

```bash
Memory usage: 70.67MiB (peak: 327.99MiB), time: 5.59s
```

Here the completed diff: https://www.diffchecker.com/BjslchcM

stof:
@Seldaek looking at the output of these runs of @alexander-schranz (and on summaries of other reports before that), I'm wondering why the memory does not go down any more (well, a lot less) after the package upgrades. Some parts of the memory might be referenced longer than before.

The improvements in peak memory usage look good though.

typhonius:
Tested this against one of my Symfony applications: https://github.com/typhonius/acquia_cli 

```
# Composer v1.10.9
# Composer version 1.10.9 2020-07-16 12:57:00
[284.2MiB/16.75s] Package operations: 96 installs, 0 updates, 0 removals
[282.3MiB/7.18s] Memory usage: 282.34MiB (peak: 508.58MiB), time: 7.18s

# Snapshot
# Composer version 2.0-dev (2.0-dev+750a92b4b7aecda0e5b2f9b963f1cb1421900675) 2020-07-17 15:29:00
[38.0MiB/0.88s] Package operations: 96 installs, 0 updates, 0 removals
[16.9MiB/0.89s] Memory usage: 16.87MiB (peak: 72.31MiB), time: 0.89s

# composer-test.php
# Composer version 2.0-dev (2.0-dev+071350286fb196722728511aecd00e1defa826c7) 2020-07-17 14:16:58
[27.1MiB/0.79s] Package operations: 96 installs, 0 updates, 0 removals
[16.8MiB/0.81s] Memory usage: 16.77MiB (peak: 61.57MiB), time: 0.81s
```

gitressa:
Drupal updates:

- drupal/admin_toolbar (2.2.0 => 2.3.0)
- drupal/core (8.9.1 => 8.9.2)
- drush/drush (10.3.0 => 10.3.1)

```
# Composer v1.10.9
[1219.2MiB/12.53s] Dependency resolution completed in 1.591 seconds
[402.0MiB/12.93s] Analyzed 16005 packages to resolve dependencies
[402.0MiB/12.93s] Analyzed 1742304 rules to resolve dependencies
[402.0MiB/12.93s] Package operations: 0 installs, 6 updates, 0 removals
[398.6MiB/12.93s] Memory usage: 398.62MiB (peak: 1219.78MiB), time: 12.93s

# Snapshot
[456.9MiB/4.06s] Dependency resolution completed in 0.384 seconds
[85.0MiB/4.20s] Analyzed 13246 packages to resolve dependencies
[85.0MiB/4.20s] Analyzed 705611 rules to resolve dependencies
[86.1MiB/4.21s] Package operations: 0 installs, 6 updates, 0 removals
[42.6MiB/4.22s] Memory usage: 42.59MiB (peak: 457.31MiB), time: 4.22s

# composer-test.phar
[436.8MiB/3.90s] Dependency resolution completed in 0.366 seconds
[65.0MiB/4.05s] Analyzed 6837 packages to resolve dependencies
[65.0MiB/4.05s] Analyzed 705611 rules to resolve dependencies
[66.1MiB/4.05s] Package operations: 0 installs, 6 updates, 0 removals
[41.7MiB/4.06s] Memory usage: 41.66MiB (peak: 437.29MiB), time: 4.06s

```

gitressa:
Drupal 8 installation with Drush 10 and 28 contrib modules:

```
# Composer v1.10.9
[475.1MiB/30.59s] Memory usage: 475.15MiB (peak: 1799.94MiB), time: 30.59s

# Snapshot
[44.1MiB/5.39s] Memory usage: 44.1MiB (peak: 572.63MiB), time: 5.39s

# composer-test.phar
[45.9MiB/5.57s] Memory usage: 45.86MiB (peak: 561.9MiB), time: 5.57s
```

heddn:
I got the following error while attempting to run `$ composer update --dry-run`:
```
$ composer update --dry-run
No composer.json in current directory, do you want to use the one at /var/www/html? [Y,n]? 
Updating dependencies

  Problem 1
    - Root composer.json requires drupal/html_title dev-8.1.x, it is satisfiable by drupal/html_title[dev-8.1.x] from package repo (defining 1 package) but drupal/html_title[dev-1.x, 1.0.0-rc1, 1.0.0-rc2, 1.0.0, 1.x-dev (alias of dev-1.x)] from composer repo (https://packages.drupal.org/8) has higher repository priority. The packages with higher priority do not match your constraint and are therefore not installable. See https://getcomposer.org/repoprio for details and assistance.

You are using a snapshot build of Composer 2, which may be the cause of the problem. Run `composer self-update --stable` and then try again. In case it solves the problem, please report an issue mentioning Composer 2
```
I have a repository setup for that project and I do not get this error without composer v2. And to be clear, this was with simply HEAD on composer v2. Not the test phar.

```
        {
            "type": "package",
            "package": {
                "name": "drupal/html_title",
                "version": "dev-8.1.x",
                "type": "drupal-module",
                "source": {
                    "url": "https://github.com/markcarver/drupal-module-html_title.git",
                    "type": "git",
                    "reference": "patch-1"
                }
            }
        },
```

heddn:
Composer 2 HEAD:
`Memory usage: 46.78MiB (peak: 609.27MiB), time: 8.8s`

composer-test.phar:
`Memory usage: 51.58MiB (peak: 601.15MiB), time: 8.8s`

stof:
@heddn the order of repositories is important in the config. And with the introduction of canonical repositories (which is the default behavior in Composer 2), the order is even more important. I think you you need to swap the order of your package repository and your `https://packages.drupal.org/8` repo. Btw, the error message contains a link with the explanation of that feature.

heddn:
We have some very interesting numbers over in https://www.drupal.org/project/drupal/issues/3161163#comment-13771033. Seems to indicate that these updates here help a lot.

andrerom:
eZ Platform 2.5.6 _(older tag)_, with:
- PHP 7.2
- `--no-plugins` to not use Flex
- `--ignore-platform-reqs` to not blow up on `ocramius/package-versions`

```
1.10.9:
Package operations: 7 installs, 93 updates, 1 removal
Memory usage: 540.03MiB (peak: 2882.82MiB), time: 43.7s

2.0 snapshot:
Lock file operations: 9 installs, 91 updates, 3 removals
Memory usage: 53.32MiB (peak: 987.67MiB), time: 14.96s

2.0#8850 composer-test.phar:
Lock file operations: 9 installs, 91 updates, 3 removals
Memory usage: 96.07MiB (peak: 974.82MiB), time: 14.98s
```

gitressa:
As @heddn just mentioned, testing an update with 30 popular Drupal 9 modules individually, both `composer self-update --preview` and `--snapshot` were much more memory demanding and slower (167 MB or 456 MB), compared to `composer-test.phar`.

Also, 11 of the modules were extra slow, for some reason, whereas all 30 modules updated with much lower memory usage of 96 MB using `composer-test.phar`. The process:

```
$ composer create-project drupal/recommended-project:9.0.1 drupal901
$ cd drupal901
```

**FAST module, drupal/captcha**
```
$ composer require drupal/captcha
$ COMPOSER_DISABLE_NETWORK=1 composer update drupal/core-recommended --with-all-dependencies -v --dry-run --profile
[91.4MiB/1.67s] Analyzed 17704 packages to resolve dependencies
[91.4MiB/1.67s] Analyzed 112582 rules to resolve dependencies
[32.4MiB/1.71s] Memory usage: 32.36MiB (peak: 165.47MiB), time: 1.71s
$ COMPOSER_DISABLE_NETWORK=1 php composer-test.phar update drupal/core-recommended --with-all-dependencies -v --dry-run --profile
[41.6MiB/1.25s] Analyzed 3624 packages to resolve dependencies
[41.6MiB/1.25s] Analyzed 70519 rules to resolve dependencies
[30.5MiB/1.27s] Memory usage: 30.5MiB (peak: 96.8MiB), time: 1.27s

```
**SLOW module, drupal/block_class**
```
$ composer require drupal/block_class
$ COMPOSER_DISABLE_NETWORK=1 composer update drupal/core-recommended --with-all-dependencies -v --dry-run --profile
[104.8MiB/3.60s] Analyzed 17704 packages to resolve dependencies
[104.8MiB/3.60s] Analyzed 602126 rules to resolve dependencies
[44.2MiB/3.62s] Memory usage: 44.23MiB (peak: 456.36MiB), time: 3.62s
$ COMPOSER_DISABLE_NETWORK=1 php composer-test.phar update drupal/core-recommended --with-all-dependencies -v --dry-run --profile
[41.7MiB/1.24s] Analyzed 3624 packages to resolve dependencies
[41.7MiB/1.24s] Analyzed 70519 rules to resolve dependencies
[30.1MiB/1.25s] Memory usage: 30.08MiB (peak: 96.8MiB), time: 1.25s

```
### Results for `composer self-update --preview` and `--snapshot`
MB | Package
------------ | -------------
 164  | drupal/google_analytics
 165  | drupal/admin_toolbar
 165  | drupal/captcha
 165  | drupal/colorbox
 165  | drupal/entity
 165  | drupal/imce
 165  | drupal/redirect
 165  | drupal/search_api
 165  | drupal/token
 165  | drupal/views_bulk_operations
 165  | drupal/xmlsitemap
 166  | drupal/devel
 167  | drupal/backup_migrate
 167  | drupal/ctools
 167  | drupal/entity_reference_revisions
 167  | drupal/mailsystem
 167  | drupal/paragraphs
 167  | drupal/pathauto
 167  | drupal/smtp
 456  | drupal/better_exposed_filters
 456  | drupal/block_class
 456  | drupal/bootstrap
 456  | drupal/features
 456  | drupal/honeypot
 456  | drupal/inline_entity_form
 456  | drupal/menu_block
 456  | drupal/metatag
 456  | drupal/panels
 456  | drupal/recaptcha
 456  | drupal/views_slideshow

### Results for `composer-test.phar`
MB | Package
------------ | -------------
 96 | drupal/admin_toolbar
 96 | drupal/backup_migrate
 96 | drupal/block_class
 96 | drupal/bootstrap
 96 | drupal/captcha
 96 | drupal/colorbox
 96 | drupal/ctools
 96 | drupal/entity
 96 | drupal/entity_reference_revisions
 96 | drupal/features
 96 | drupal/google_analytics
 96 | drupal/honeypot
 96 | drupal/imce
 96 | drupal/inline_entity_form
 96 | drupal/mailsystem
 96 | drupal/menu_block
 96 | drupal/metatag
 96 | drupal/panels
 96 | drupal/paragraphs
 96 | drupal/recaptcha
 96 | drupal/redirect
 96 | drupal/search_api
 96 | drupal/smtp
 96 | drupal/token
 96 | drupal/views_bulk_operations
 96 | drupal/views_slideshow
 96 | drupal/xmlsitemap
 97 | drupal/better_exposed_filters
 97 | drupal/devel
 97 | drupal/pathauto

simonberger:
Both results are from checked out master and branch + merged master as I faced a bug which had been fixed in the meantime.
I also could not use COMPOSER_DISABLE_NETWORK=1 because our artifactory repository is still called.

composer 2 head (c0eb9834fef009316b0146763f9440ddd844169d)
```
Analyzed 18461 packages to resolve dependencies
Analyzed 2391210 rules to resolve dependencies

Memory usage: 110.75MiB (peak 1618.77MiB), time: 147.91s
```

composer pull/8850 + merged head
```
Analyzed 17314 packages to resolve dependencies
Analyzed 2177711 rules to resolve dependencies

Memory usage: 226.97MiB (peak 1611.46MiB), time: 180.25s
```

This is a Symfony project with many direct and indirect dependencies.
Runtime unfortunately increased. I hoped for a similar or even better impact in the other direction.

Side note:
Composer 2 itself improved memory usage and runtime two to three times compared to 1.10.

Seldaek:
> @Seldaek looking at the output of these runs of @alexander-schranz (and on summaries of other reports before that), I'm wondering why the memory does not go down any more (well, a lot less) after the package upgrades. Some parts of the memory might be referenced longer than before.

https://github.com/composer/composer/pull/8850/commits/00e268cdbfd1dde8c311fca3f22d49e51daeaffe attempts to fix that.. at least it fixes some of it, but I could not reproduce the amount of extra memory @alexander-schranz reported so I'm not sure what's going on there.

simonberger:
I had a closer look at the verbose lock of this test vs snapshot why the performance is decreased in our case.
I noted an increased amount of lines in general resulting from more and partly multiple requests to packages. I would not call it a problem in the implementation at this point. The Artifactory repositories we use might hurt this test (more), but I can confirm the bigger numbers in multiple runs.

Seldaek:
@simonberger We haven't optimized the usage for non-v2-compatible repos there. That's probably why it slows down for you.

simonberger:
I can confirm that after removing the artifactory layer and registering all private packages as repositories the timings are almost matching.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -238,6 +238,8 @@ public function buildPool(array $repositories, Request $request)
         $this->skippedLoad = array();
         $this->indexCounter = 0;
 
+        Intervals::clear();
```
GrahamCampbell, line 241: Does this still need to be run if the above code throws an exception?

effulgentsia:
A bit more info for the Drupal use-case detailed by @gitressa [above](#issuecomment-667117313).

The difference between the fast (low'ish peak memory) and slow (high peak memory) modules appears to be their (direct or indirect) version constraints on `drupal/core`.

I can replicate the same high peak memory situation without any of those modules, and instead with just a [local path package](https://getcomposer.org/doc/05-repositories.md#path) containing just the following composer.json:
```
{
  "name": "my/composer-8850-test",
  "require": {
    "drupal/core": "^8 || ^9"
  }
}
```

drupal/core has many dependencies, and allowing "^8" ends up allowing many different version possibilities of them, which I suspect is what contributes the bulk of the 17k packages and 600k rules that need to be analyzed.

Even if I include `"drupal/core": "^9"` in my root composer.json, that doesn't help Composer 2 alpha3 avoid all of the unnecessary checking of the rules brought in by the "^8" of a dependent such as above.

However, this PR does appear to successfully optimize those away (and not consume corresponding memory), which is awesome!

stof:
> However, this PR does appear to successfully optimize those away (and not consume corresponding memory), which is awesome!

this is exactly what this Pr is meant to optimize

simonberger:
I used conflict filtering on a slightly reduced set of packages:

With conflict filtering:
```
Memory usage: 92.85MiB (peak: 1080.98MiB), time: 13.13s
Analyzed 14873 packages to resolve dependencies
Analyzed 1951039 rules to resolve dependencies
```

Without conflict filtering
```
Memory usage: 125.91MiB (peak: 1436.66MiB), time: 14.76s
Analyzed 16621 packages to resolve dependencies
Analyzed 2231769 rules to resolve dependencies
```

I don't know how common questionable requirement usage like this are
```
"require": {
     "codeception/codeception": "*@dev"
},
"conflict": {
     "codeception/codeception": "<4.0"
}
```

which turned in `>=4.0.0.0-dev`. Beside of 12 occurrences of this I had 4 more different conflicts with 48 occurrences. 40 had a conflict out of the range of the requirement (just took time without changes) and the other 8 filtered single versions.

The code I used to generate a new constraint from a requirement and a conflict worked on this project but is only a prototype.

Seldaek:
@simonberger thinking a bit more about this conflict support, I don't think it's worth doing as indeed the pattern is not very common (and rather silly), and we don't have an easy way to implement this with current composer/semver functionality.

I also took a look at the artifactory slowdown and as far as I can tell it should be more or less as fast as before. I am not sure what is slowing it down.. Maybe you can have a closer look as I don't have the ability to check this. But the bottom line is we want to support primarily repositories serving v2-metadata format as it is more efficient, so I don't know if it's worth spending too much time there.

I merged master into this branch btw so it's up to date with all the fixes done there.

simonberger:
@Seldaek The problem regarding artifactory is less the v1 API. It does not use any way to provide package metadata (https://www.jfrog.com/jira/browse/RTFACT-20967).
We use two artifactory repositories. One as a proxy to packagist which obviously needs this packages.json strategy to receive new packages but the other (higher) priority repository has our internal packages which currently respond 404 to all other package requests. The artifactory proxy repository has those packages then.
But as you wrote that should be the same for master and this branch. I can have a close look at it the next days but it is kinda hard to debug. I just noticed some more requests from this branch yet. Maybe you get an idea from those setup details.

To the conflict handling, yeah I am curious if it is generally worth it too and it is definitely not easy to do. I added a `public static function diff(ConstraintInterface $from, ConstraintInterface $to)` method to the Intervals class which returns a new constraint but it currently is in a very naiv state leaving a lot out (stability, branches...).
From what I saw conflict handling could be improved also for the root package and probably it would be more worth to try to drop conflicting packages over all packages. But the scope of this PR is already big enough. I learned a lot how composer works the last days. Maybe I'll start to think about it again later (or after composer 2 release). Hopefully at this point those big slowdown packages are fixed on the side of the author.

simonberger:
After some log analyzing and debugging the problem of the slowdown was luckily quiet comprehensible.

If we change a constraint of an already loaded package it obviously will be reloaded. A package in a repository with lazy loading (or maybe even in other cases) will be checked for changes after cache reading. This reverses the gain. At least in my scenario where only a few versions could be dropped at the end.

I tested a patch which fixes the problem.
I don't see a reason how this could break anything but there might be a better solution.

```
Index: src/Composer/Repository/ComposerRepository.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/Composer/Repository/ComposerRepository.php	(revision 9a04ecefbf84a776d9897fe6efdf9142dd56c80f)
+++ src/Composer/Repository/ComposerRepository.php	(date 1598572617187)
@@ -571,11 +571,15 @@
                     if ($contents = $this->cache->read($cacheKey)) {
                         $contents = json_decode($contents, true);
                         if (isset($contents['last-modified'])) {
-                            $response = $this->fetchFileIfLastModified($url, $cacheKey, $contents['last-modified']);
-                            if (true === $response) {
-                                $packages = $contents;
-                            } elseif ($response) {
-                                $packages = $response;
+                            if (isset($alreadyLoaded[$name])) {
+                                $packages = $contents;
+                            } else {
+                                $response = $this->fetchFileIfLastModified($url, $cacheKey, $contents['last-modified']);
+                                if (true === $response) {
+                                    $packages = $contents;
+                                } elseif ($response) {
+                                    $packages = $response;
+                                }
                             }
                         }
                     }
```

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -241,19 +229,111 @@ public function buildPool(array $repositories, Request $request)
...
+        // we make sure that we load at most the intervals covered by the root constraint.
+        $rootRequires = $request->getRequires();
+        if (isset($rootRequires[$name]) && !Intervals::isSubsetOf($constraint, $rootRequires[$name])) {
+            $constraint = $rootRequires[$name];
+        }
+
+        // Maybe it was already marked before but not loaded yet. In that case
+        // we have to extend the constraint (we don't check if they match because
+        // MultiConstraint::create() will optimize anyway)
+        if (isset($this->packagesToLoad[$name])) {
```
naderman, line 271: It uses isSubsetOf below, is the comment outdated?

Should we check if the constraint is actually identical to the one already loaded to avoid calling create/compact in that case or is that too uncommon to matter?

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -241,19 +229,111 @@ public function buildPool(array $repositories, Request $request)
...
+        // Not yet loaded or already marked for a reload, override the existing constraint
+        // (either it's a new one to load, or it has already been extended above)
+        if (!isset($this->loadedPackages[$name])) {
+            $this->packagesToLoad[$name] = $constraint;
+            return;
+        }
+
+        // No need to load this package with this constraint because it is
+        // a subset of the constraint with which we have already loaded packages
+        if (Intervals::isSubsetOf($constraint, $this->loadedPackages[$name])) {
```
naderman, line 290: Can't this be moved up to before the create/compact to skip merging constraints if the new constraint requested now is already contained in what is loaded? It seems like this wouldn't change anything and just skip some unnecessary work in some cases?

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: That's actually what this comment is supposed to say. "we don't check if they match" === "we don't check if they are exactly the same" :) I guess we could do a string comparison on both constraints but not sure if worth it.

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: Create/Compact before does not compact what's already loaded. It compacts what's going to be loaded :)
Line 271 is `packagesToLoad`. So this case handles multiple consecutive `markPackageNameForLoading()` calls for the same package. I think it's okay the way it is now. Chances that the check on `loadedPackages` has to do something are a lot higher than on `packagesToLoad`.

simonberger: Theoretically packagesToLoad could be unset here. But a situation where packagesToLoad was already set and is (still) and subset of the loadedPackages should never happen, right?

src/Composer/DependencyResolver/PoolBuilder.php
naderman: Well matching constraints kind of has a different meaning, so I'll update this ;-)

src/Composer/DependencyResolver/PoolBuilder.php
naderman: I moved this section into the other if to make it clear how this works. You're right that what i had in mind would never happen as loadedPackages automatically gets unset whenever packagesToLoad gets set.

Toflar:
🎉

simonberger:
That was quick at the end. No one was up to review https://github.com/composer/composer/pull/8850#issuecomment-682257801 ?

naderman:
@simonberger Ah sorry, overlooked that when I checked if any issues were left open. Can you maybe just submit that as a new PR against master?

simonberger:
Yes I'll add one.

PR #8867, opened 2020-05-04
Feat/gitlab deploy token

KminekMatej:
Added ability to work with GitLab deploy tokens: https://docs.gitlab.com/ee/user/project/deploy_tokens/

Deploy tokens can be specified two ways:
1) GIT CONFIG:
git config --add gitlab.deploytoken.user USERNAME && git config --add gitlab.deploytoken.token TOKEN
2) Auth.json:
"gitlab-token": {
    "gitlab.com": {"username": "USERNAME", "token": "TOKEN"}
}

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -184,9 +184,9 @@ class Git
                 if ($this->io->hasAuthentication($match[2])) {
                     $auth = $this->io->getAuthentication($match[2]);
                     if ($auth['password'] === 'private-token' || $auth['password'] === 'oauth2' || $auth['password'] === 'gitlab-ci-token') {
-                        $authUrl = $match[1] . '://' . rawurlencode($auth['password']) . ':' . rawurlencode($auth['username']) . '@' . $match[2] . '/' . $match[3] . '.git'; // swap username and password
+                        $authUrl = $match[1] . '://' . rawurlencode($auth['password']) . ':' . rawurlencode($auth['username']) . '@' . $match[2] . '/' . $match[3]; // swap username and password
                     } else {
-                        $authUrl = $match[1] . '://' . rawurlencode($auth['username']) . ':' . rawurlencode($auth['password']) . '@' . $match[2] . '/' . $match[3] . '.git';
+                        $authUrl = $match[1] . '://' . rawurlencode($auth['username']) . ':' . rawurlencode($auth['password']) . '@' . $match[2] . '/' . $match[3];
                     }
 
                     $command = call_user_func($commandCallable, $authUrl);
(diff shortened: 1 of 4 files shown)


src/Composer/Util/Git.php
```diff
@@ -184,9 +184,9 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
                 if ($this->io->hasAuthentication($match[2])) {
                     $auth = $this->io->getAuthentication($match[2]);
                     if ($auth['password'] === 'private-token' || $auth['password'] === 'oauth2' || $auth['password'] === 'gitlab-ci-token') {
-                        $authUrl = $match[1] . '://' . rawurlencode($auth['password']) . ':' . rawurlencode($auth['username']) . '@' . $match[2] . '/' . $match[3]; // swap username and password
+                        $authUrl = $match[1] . '://' . rawurlencode($auth['password']) . ':' . rawurlencode($auth['username']) . '@' . $match[2] . '/' . $match[3] . '.git'; // swap username and password
                     } else {
-                        $authUrl = $match[1] . '://' . rawurlencode($auth['username']) . ':' . rawurlencode($auth['password']) . '@' . $match[2] . '/' . $match[3];
+                        $authUrl = $match[1] . '://' . rawurlencode($auth['username']) . ':' . rawurlencode($auth['password']) . '@' . $match[2] . '/' . $match[3] . '.git';
```
Seldaek, line 189: I think these changes should be reverted now that the .git suffix is fixed?

src/Composer/Util/GitLab.php
```diff
@@ -168,9 +179,9 @@ private function createToken($scheme, $originUrl)
         );
 
         $token = $this->httpDownloader->get($scheme.'://'.$apiUrl.'/oauth/token', $options)->decodeJson();
-
-        $this->io->writeError('Token successfully created');
-
+        
+        $this->io->writeError('Token successfully created');	
+
```
Seldaek, line 184: Please don't add trailing whitespace, you can see it highlighted here in dark green, best is to set editor to automatically trim trailing whitespace on save.

Seldaek:
This new stuff would also need to be added to the docs at https://github.com/composer/composer/blob/master/doc/06-config.md#gitlab-token

src/Composer/Util/Git.php
KminekMatej: Yea, sorry, my bad. Removed in next commit

src/Composer/Util/GitLab.php
KminekMatej: Unfortunately, Netbeans doesnt have this ability. Removed manually.

KminekMatej:
> This new stuff would also need to be added to the docs at https://github.com/composer/composer/blob/master/doc/06-config.md#gitlab-token

I pushed here another commit with documentation. However I found that there is no mention about using git config attributes `bitbucket.accesstoken` , `github.accesstoken` and `gitlab.accesstoken` + in my commit, there are additional possible configs `gitlab.deploytoken.user` and `gitlab.deploytoken.token` . Shouldnt it be mentioned in docs somewhere as well?

Seldaek:
Regarding the git config stuff, if you want to add it for all of the supported ones that would be a nice addition in a separate PR.

Otherwise here it looks good, thanks!

PR #8868, opened 2020-05-04
Add .git extensions into Gitlab git requests

KminekMatej:


src/Composer/Util/Git.php
```diff
@@ -184,9 +184,9 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
                 if ($this->io->hasAuthentication($match[2])) {
                     $auth = $this->io->getAuthentication($match[2]);
                     if ($auth['password'] === 'private-token' || $auth['password'] === 'oauth2' || $auth['password'] === 'gitlab-ci-token') {
-                        $authUrl = $match[1] . '://' . rawurlencode($auth['password']) . ':' . rawurlencode($auth['username']) . '@' . $match[2] . '/' . $match[3]; // swap username and password
+                        $authUrl = $match[1] . '://' . rawurlencode($auth['password']) . ':' . rawurlencode($auth['username']) . '@' . $match[2] . '/' . $match[3] . '.git'; // swap username and password
```
Seldaek, line 187: So it seems to me according to the code above that `.git` might be present if the line 169 matched, but is excluded if line 168 matched. So maybe a more accurate fix (to avoid duplicating .git) is to move the \.git into the match group at line 168?

Seldaek: i.e. what is broken is the handling of `git@gitlab.com:..` URLs, but I believe if you provide it as `https://gitlab.com/foo/bar.git` then it will use the correct URL.

KminekMatej: Yea, that seems right, that slipped my mind. Changing the regex on line 169 to 
`'{^(https?)://' . self::getGitLabDomainsRegex($this->config) . '/(.*)\.git$}'` 
should be enough. Im gonna push another commit with that

PR #8869, opened 2020-05-04
Adding a test to verify replace version overlap with unbound constraints and single versions

naderman:
This will require a semver library update to be fixed.


stof:
should this be fixed only in master or in 1.10 ?

naderman:
@stof In principle it's a bug in 1.10 too. However it's a pretty rare scenario which is why I think we haven't noticed this until now. As it requires a somewhat strong change in the semver lib, I think I would prefer to keep the "buggy" behavior in semver 1.x and composer 1.x, and only fix it in 2.x semver and composer. I'm not entirely sure that's the right way to go however?

Seldaek:
Should be fine now so merging

PR #8871, opened 2020-05-04
Make the VersionSelector take Composer API/Runtime versions into account

Seldaek:
Fixes getsentry/sentry-php#1008


naderman:
So I think if anything we should do this for the entire platform repo? Then the statement is simply: composer outdated will only list potential updates to any versions which are compatible with your platform (php, exts, composer version, etc). Maybe there is value in providing an option for whether this is wanted or not?

That said, is this going to potentially hinder upgrades to Composer v1 because updated versions using Composer v2 will not show up?

This branch targets master, but wouldn't this be needed on 1.x to be useful currently?

Seldaek:
Merging as is for 1.10, will work on a full-platform-repo solution for 2.0

Issue #8834, opened 2020-04-24
Option to disable deprecation notice about psr4


Yivan:
Hello,

Last composer version send a lot of `Deprecation Notice` saying `does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0`.
Actually it reports vendor ones and we cannot change them.

It mess the cli output, even more in verbose mode where we can see only this and all other message are lost in a yellow fog of deprecated notice ; )

Purpose of this issue is having the ability to disable this message.
Setting php to error_reporting 0 don't work. It should respect it (or at least E_USER_DEPRECATED).

Thanks!

Seldaek:
These are only warnings.. Just make sure you update your dependencies to latest version which might have a fix already, or if not then report it to them so it gets fixed. Things need to get fixed to avoid problems in the future, ignoring warnings is not a solution.

That said if it's that disturbing in development, you can also avoid dumping an optimized autoloader, because that's the only point where the warning is emitted, and optimized autoloader on dev machines does not bring all that much.

Yivan:
@Seldaek Thanks for your fast return on this.
Yes I understand totally that you prefer to not provide this option to ensure better ugrading to composer 2.0 (which i wait too because faster! : ) ).

But sadly we cannot upgrade dependency like you suggest as many projects are enterprise production project. So this way of upgrading dependencies is not applicable.
The way to disable optimize is tricky for us because we use Pimcore as development plateform and post-script composer are running optimizing automatically. So it needs some refactoring, changing our CI, etc. And telling all devs of the company to no more use optimize, etc... you anderstand just for a warning message is not a good solution either.

I imagine every people have different use case, so I was really thinking such an option to disable warning would be helpful. For now we have to live with it as no solution can work easily in our case.

The more easy solution is to downgrade composer, i imagine.

I hope if some more people ask for this feature it will be reopened!

Thanks again,
Best regards.

PR #8901, opened 2020-05-13
feat: add option to prevent error_reporting() overriding runtime configuration

gsdevme:
This adds the ability to disable the override within the error_handler to meet what you might expect that your php.ini will be able to correctly set the error_reporting.

Somewhat fixes https://github.com/composer/composer/issues/8834 but perhaps in a more acceptable manner (I hope). 

While I agree these are very useful and I will want to be upgrading and fixing them if you are using `composer` as part of a deployment process its not very useful to have a chunk of spam messages within your logs and filling up your CI output. 

I agree in a development environment they are very important, but if you have actively configured the environment to ignore deprecations (rightly due to the Production environment) I don't fully understand the benefit. Also assuming this means scripts executing within Composer also don't get the correct runtime configuration. 

Im not 100% sure on the best way to contribute to this project so hopefully a PR to master I can also get it merged into the 1.x branch? 

Also sorry for the naming, if anyone has anything better happy to change it. 

Thanks

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -24,6 +24,7 @@ The following options are available with every command:
 * **--no-plugins:** Disables plugins.
 * **--no-cache:** Disables the use of the cache directory. Same as setting the COMPOSER_CACHE_DIR
   env var to /dev/null (or NUL on Windows).
+* **--no-runtime-error-reporting-override:** Prevent the error handler changing the error_reporting configuration from the one set by the runtime.
 * **--working-dir (-d):** If specified, use the given directory as working directory.
 * **--profile:** Display timing and memory usage information
 * **--ansi:** Force ANSI output.
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -127,7 +127,14 @@ class Application extends BaseApplication
         $io = $this->io = new ConsoleIO($input, $output, new HelperSet(array(
             new QuestionHelper(),
         )));
-        ErrorHandler::register($io);
+
+        $overrideRuntimeErrorReporting = true;
+
+        if ($input->hasParameterOption('--no-runtime-error-reporting-override')) {
+            $overrideRuntimeErrorReporting = false;
+        }
+
+        ErrorHandler::register($io, $overrideRuntimeErrorReporting);
 
         if ($input->hasParameterOption('--no-cache')) {
             $io->writeError('Disabling cache usage', true, IOInterface::DEBUG);
@@ -230,7 +237,7 @@ class Application extends BaseApplication
                 if (function_exists('posix_getuid') && posix_getuid() === 0) {
                     if ($commandName !== 'self-update' && $commandName !== 'selfupdate') {
                         $io->writeError('<warning>Do not run Composer as root/super user! See https://getcomposer.org/root for details</warning>');
-                        
+
                         if ($io->isInteractive()) {
                             if (!$io->askConfirmation('<info>Continue as root/super user</info> [<comment>yes</comment>]? ', true)) {
                                 return 1;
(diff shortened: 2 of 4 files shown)


Seldaek:
I'd really rather avoid a new flag for such a thing. We could perhaps avoid adding E_DEPRECATED / E_USER_DEPRECATED to error_reporting unless it is in already.. Possibly only avoid doing it when --no-dev is present so typical prod `composer install --no-dev` would not see warnings. Or could also do that if `getenv('CI')` is present.. 

But overall I don't really like to hide these warnings though, because you might be hiding them now to get rid of autoloading deprecations, and then forget to remove it once the warnings are actually fixed and next time we add a warning you miss it entirely until things break. I think it's much more productive to expend energy fixing the warnings than working around them.

gsdevme:
> I think it's much more productive to expend energy fixing the warnings than working around them.

Yeah I understand the need for them, just practically can't drop everything and resolve instantly as I'm sure with most development houses they have to be resourced and planned in. Even more tricky when you get some from 3rd party code you can't control directly. 

The important thing to get them on the radar and yeah they are splattered all over the dev/staging/test/prod radar  

I don't mind changing the PR to either option, having an explicit "I know I'm disabling this error handling" I thought would fit the need for composer to flag these though? something like `--no-dev` or  `getenv('CI')` is implicit and you might not be aware. 

> Also assuming this means scripts executing within Composer also don't get the correct runtime configuration.

I also still not sure around this though? if we are changing the runtime would this really be expected that any "scripts" would also be subject to this? Another reason why I thought an explicit "I know" might be better.

Seldaek:
Closing this as I still kinda disagree with the intent here, stuffing warning under the rug is the main reason there are so many warnings out there.

DaanBiesterbos:
The warnings from vendor packages are driving me borderline insane. I will keep using composer 1 until the whole company is ready to upgrade to Symfony 5. Or perhaps wait for a fork. 

I do agree that warnings should be visible (by default). 
I also think it's MY development environment and since there are many packages I cannot change. Who cares that the tests in some vendor package are not autoloaded. Those should not be in the dist either. Maybe add a warning for that as well? 
Although I do not disagree with your point of view in general I feel like you are taking it too far. I am perfectly able to type "--verbose"  or "--quiet" or what so ever. I don' t need to be spammed with a ton of warnings from packages I have no control over on a daily basis.

Seldaek:
@DaanBiesterbos composer 1 also warns though.. So I am not sure what you are raving about there. On your development environment you also don't need an optimized autoloader IMO, so the warnings should not happen there?

DaanBiesterbos:
Hmm. Well thanks for pointing out that downgrading won't make a difference lol xD I assumed that the warnings were caused by the upgrade but you are right, the optimized autoloader is now enabled in composer.json so that seems to be the problem. Guess I will to see if we can revert it.

gsdevme:
You could downgrade to a specific version 1 that didn't include this verbose output though while you don't have time or the capacity to resolve the issues @DaanBiesterbos 

I'm not sure what version of 1.x introduced these warnings.

Seldaek:
IMO optimized autoloading should be part of the build/deploy script, adding `--optimized-autoloader` or whatever flags you need to `install` command. Local install runs don't need to be slowed down by this, so I'd remove it from the composer.json.

stof:
Thus, the optimized autoloader hurts DX. If you remove a class (which can happen easily during refactoring), the optimized autoloader will still expect it to exist until you re-dump it, and this will have weird effects. You really don't want to re-dump the autoloader each time you change something in your code in a dev setup (otherwise, you could as well use `classmap` in the composer config instead of convention-based discovery of classes)

gsdevme:
Sorry yeah my original PR and issue was for deployment not for development, just giving a solution to the exact problems @DaanBiesterbos is referencing.

gsdevme:
I can't think of anything else that yells warnings at you without the ability to take a conscious decision to accept them until the time is right to action them, not to mention alter the terms of the environment which you've asked to run in. 

Nobody has even suggested we wouldn't want to fix the warnings or putting them "under the rug" its about solving problems at the right time and that isn't as part of a deployment script you might run 50 times per day every day.

bretticus:
Wow, composer still won't let us suppress these? How do I avoid the nag to update to Composer 2 LOL

Seldaek:
@bretticus by upgrading to Composer 2..

PR #8908, opened 2020-05-15
Add commit constraint to documentation

nathansalter:
I wanted to add a reliable constraint on a branch to a specific version for an external library which did not provide an up-to-date release for the code I needed to deploy. I couldn't find any description of this in the composer docs so I've provisionally added this here.


stof:
This is **not** a reliable way, as metadata (dependencies for instance) will still be taken from `dev-develop`. This feature is an installer-level *hack* changing the installed version after the solver ran.

My experience is tnat *all* projects using this have suffered from this choice a few weeks/months later (when the branch changes its metadata).

nathansalter:
Ok, so a more reliable way would be to fork the repo and create a tag on the commit that you require? I guess as commit hashes aren't persistent (with history rewrites) that makes a lot of sense. I'll close this PR.

stof:
it is not about commit hashes not being persistent. It is about Packagist not storing metadata for each commit, but only for branches and tags.

PR #8912, opened 2020-05-17
Use the verbose option to determine whether to output "Changed current directory to"

Serializator:
**Problem / Context**
I was writing a Node.js application which used `composer global ...` to add a repository to the global configuration (`composer global config ...`) and get the versions of a package (`composer global show --all ...`).

I used the exit code of the process as well as the output to stderr to determine whether the command executed successfully, this didn't seem to work. After debugging some stuff I realised the "Changed current directory to" message was written to stderr instead of stdout causing my check to fail.

"Changed current directory to" should (in my opinion) be written to stdout as it's not an error. But because of the nature of some commands such as `composer global show --all ...` this message shouldn't always be written to stdout because then stdout wouldn't be able to be parsed anymore when e.g. `--format=json` is used.

**Solution**
I decided to solve this by only writing the "Changed current directory to" message to stdout if `--verbose` (or any of its aliases) are passed to the `composer global` command.

### DIFF
--- a/src/Composer/Command/GlobalCommand.php
+++ b/src/Composer/Command/GlobalCommand.php
@@ -16,6 +16,7 @@ use Composer\Factory;
 use Composer\Util\Filesystem;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\StringInput;
 use Symfony\Component\Console\Output\OutputInterface;
 
@@ -31,6 +32,7 @@ class GlobalCommand extends BaseCommand
             ->setDescription('Allows running commands in the global composer dir ($COMPOSER_HOME).')
             ->setDefinition(array(
                 new InputArgument('command-name', InputArgument::REQUIRED, ''),
+                new InputOption('verbose', 'v|vv|vvv', InputOption::VALUE_NONE, ''),
                 new InputArgument('args', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, ''),
             ))
             ->setHelp(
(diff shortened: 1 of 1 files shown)


Serializator:
I created this pull request as a draft because I'd like some feedback before I actually make this a pull request with the intention of it being merged.

- I don't know why stderr was used in the first place
- I rewrote a sizeable chunk of code to be able to use the `InputInterface`, I tested it thoroughly but I might not be aware of some of the edge cases (if there are any). I derived the logic I wrote from the original for loop.

stof:
Writing to `stderr` does not indicate a failure. It is about writing any diagnostic message meant to be read by a human but not being part of the output piped to another program (btw, I never found a bash reference calling the second pipe `stderr`).
Identifying failures is what the exit code is about.

Serializator:
Hi @stof,

I can agree that `stderr` shouldn't be used to indicate a failure, which helps me a lot because that means indeed I shouldn't be checking `stderr` but only the exit code (and use the `stderr` for any diagnostic messages if the exit code indicates a failure). Thanks!

In that case there is the question what should and what shouldn't be written to `stderr`, because as you say it should be used for diagnostic messages. Arguably the "Changed current directory to" can be seen as a diagnostic message might something fail. At the same time, what is verbosity then for (verbose messages are mostly used as a diagnostic "tool")? 

I think diagnostic messages written to `stderr` should be in direct correlation to the failure that caused the program to exit. What is your opinion?

Kind regards,
Julian

Seldaek:
The stderr vs stdout distinction is really quite fuzzy. For commands where the output is unlikely to be manipulated it makes almost no difference tbh. For commands where the output is likely piped in some other program for further processing then stdout should only contain the relevant stuff. Verbosity doesn't necessarily mean debug info/stderr IMO. Verbose information could be supplementing the regular info and belong on stdout, or it might be debug info which belongs rather on stderr.

In any case, the exit code is what you want to look at. If composer fails and still returns 0 we'll most likely consider that a bug and happily fix. This PR however is in its current state misguided IMO so I will close.

PR #8921, opened 2020-05-21
fix deprecations in PHP 8

azjezz:
Method `ReflectionParameter::getClass()` is deprecated in PHP 8, `ReflectionParameter::getType` is only available in PHP 7.0, `ReflectionType::__toString()` is deprecated in PHP 7.4, and `ReflectionNamedType` is only avilable in PHP 7.4+

image

I'm also working on a fix for the json-scheme package :)

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -326,13 +326,22 @@ class EventDispatcher
             return $event;
         }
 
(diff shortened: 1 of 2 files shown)


azjezz:
deprecations already fixed in json-schema:master, they need to be backported into 5.x https://github.com/justinrainbow/json-schema/pull/631

Seldaek:
Thanks

PR #8926, opened 2020-05-26
Compile & compact constraints

Seldaek:
Alternative to https://github.com/composer/composer/pull/8845 which does compaction on top of compilation. Doesn't seem to be a great success in my minor tests, and I even found some weird oddities in the solver result so something is definitely off although I couldn't figure out what yet.

@jderusse if you want to compare/profile, I rebased your PR too so both are using semver 3.0.0.

### DIFF
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -14,6 +14,7 @@ namespace Composer\DependencyResolver;
 
 use Composer\Package\AliasPackage;
 use Composer\Package\Version\VersionParser;
+use Composer\Semver\CompilingMatcher;
 use Composer\Semver\Constraint\ConstraintInterface;
 use Composer\Semver\Constraint\Constraint;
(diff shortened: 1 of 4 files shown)


jderusse:
From my test (on magento)

snapshot:  [13.2s](https://blackfire.io/profiles/74d4a167-21d2-45b1-be0a-9fce5c196f37/graph)
compiling+compacting: [11.3s](https://blackfire.io/profiles/91df325d-438f-404f-91d8-6942d77e24f7/graph)
compiling only: [10.5s](https://blackfire.io/profiles/efda6de3-f9e7-4392-b401-d480995f8be5/graph)

note: `Intervals::compactConstraint` method is called 12,241 times but in only 76 cases it generates a constraint different from the original one

Seldaek:
Ok so it's not worth it for the places I put it on it seems. Still might come in handy for #8850, we'll see. But IMO we could merge the compilation at least, that seems fairly risk free and definitely improves perf.

PR #8928, opened 2020-05-27
Added composer wrapper command

benbor:
## What for?

Nearly every README for a non-ancient PHP project begins with `composer install`. Some of them explain how to install composer itself, some of them don't and rely on developer's experience, web search engine of developer's choice, or implicitly on a globally installed composer.
Well, there's nothing wrong with that, really, but I think we can do better.

Here is a couple of assumptions:
* composer is meant to be up to date;
* composer is born to be installed locally (to be upgraded easily at will, for example).

So, if there was a script that checked if composer was there, installed it if it wasn't, upgraded if outdated, and finally proxied parameters as is to it as if we referred to a composer itself, all the issues outlined above would have been solved...

## How To

1. Go to a directory within your project where "binaries" (well, usually just CLI entry points) reside
2a. If you already have globally installed composer run `composer wrap` command. The `composer` file will be added to current working directory.
2b .... Or just copy file content from github sources directly into local `composer` file:
```
curl https://raw.githubusercontent.com/benbor/composer/composer-wrapper/src/Composer/Wrapper/ComposerWrapper.php -o composer
```
\* after merge this PR, the script should be placed in more suitable place, aka  https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md
3. Add execution permissions to `composer` file: `chmod +x composer`
4. Add `composer` file to version control
5. Add `composer.phar` in that directory into version control ignore file

Done, your colleagues should worry never again about installing composer and keeping it up to date, and neither should ops team care how it appears on a build server!

**Just use the `composer` script instead of actual composer, and it will make sure you have an actual composer installed and up to date!**

`./composer install`, `./composer update`, `./composer dump-autoload`, and other commands you used to should just work as if you had installed composer yourself.

## Configuration

All the configuration is done through [*environment variables*](https://www.digitalocean.com/community/tutorials/how-to-read-and-set-environmental-and-shell-variables-on-a-linux-vps) in order to not interact with composer's CLI parameters. The following environment variables are supported:

* `COMPOSER_UPDATE_FREQ`. Time between checks for updates (defaults to 7 days). This is a [relative time specifier](http://php.net/manual/en/datetime.formats.relative.php) that is fed to [`DateTime::modify`](http://php.net/manual/en/datetime.modify.php). It's chosen because it can be perfectly readable by someone who knows no PHP and doesn't want to (e.g. ops people), and it's recommended to keep it that way, e.g. "5 days", "2 weeks", "1 month".
* `COMPOSER_DIR`. Directory where composer.phar will be searched for and copied to. Defaults to the directory where the script is located (`__DIR__`); note: *not* the current working directory! Sometimes it's useful to keep real composer.phar a couple of levels higher up the directory where wrapper is placed, for example, on a CI server: it would help avoid downloading composer afresh for every build.  

## Similar solutions

* graddlew -  wrapper for Graddle https://docs.gradle.org/current/userguide/gradle_wrapper.html
* mvnw - maven wrapper https://www.baeldung.com/maven-wrapper

## Credits
Original idea https://github.com/kamazee/composer-wrapper

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,7 @@
         "symfony/console": "2.8.38"
     },
     "require-dev": {
+        "mikey179/vfsstream": "^1.6.5",
         "phpunit/phpunit": "^4.8.35 || ^5.7",
         "phpunit/phpunit-mock-objects": "^2.3 || ^3.0"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "0367be765bb2ea718da11dbb9b3ed793",
+    "content-hash": "8c36732f764481b1d5247956e7f1e239",
     "packages": [
         {
             "name": "composer/ca-bundle",
@@ -872,6 +872,52 @@
                 "instantiate"
             ],
             "time": "2015-06-14T21:17:01+00:00"
+        },
+        {
+            "name": "mikey179/vfsstream",
+            "version": "v1.6.6",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/bovigo/vfsStream.git",
+                "reference": "095238a0711c974ae5b4ebf4c4534a23f3f6c99d"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/bovigo/vfsStream/zipball/095238a0711c974ae5b4ebf4c4534a23f3f6c99d",
+                "reference": "095238a0711c974ae5b4ebf4c4534a23f3f6c99d",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=5.3.0"
+            },
+            "require-dev": {
+                "phpunit/phpunit": "~4.5"
+            },
+            "type": "library",
+            "extra": {
+                "branch-alias": {
+                    "dev-master": "1.6.x-dev"
+                }
+            },
+            "autoload": {
+                "psr-0": {
+                    "org\\bovigo\\vfs\\": "src/main/php"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "BSD-3-Clause"
+            ],
+            "authors": [
+                {
+                    "name": "Frank Kleine",
+                    "homepage": "http://frankkleine.de/",
+                    "role": "Developer"
+                }
+            ],
+            "description": "Virtual file system to mock the real file system in unit tests.",
+            "homepage": "http://vfs.bovigo.org/",
+            "time": "2019-04-08T13:54:32+00:00"
         },
         {
             "name": "phpdocumentor/reflection-docblock",
--- a/src/Composer/Command/WrapCommand.php
+++ b/src/Composer/Command/WrapCommand.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Command;
+
+use Composer\Util\Filesystem;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ */
+class WrapCommand extends BaseCommand
+{
+    protected function configure()
+    {
+        $this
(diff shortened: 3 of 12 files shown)


alcohol:
I'm having a hard time understanding why this should be a part of Composer itself?

I understand the idea, but I feel like this only appeals to a subset of users. And some of those might even solve it in a different manner (for example, I alias `composer` to a Docker command on all my machines).

benbor:
@alcohol  Thank you  for the comment! Let me clarify why this should be a part of Composer:

1. This link is official  https://getcomposer.org/doc/00-intro.md#locally composer recommendation for installing composer locally. This pr proposes to replace only that installation path. 
     1. If a developer has not global-installed copy of composer, he\she goes to the exactly the same page (as before https://getcomposer.org/download/) and copy-paste a content of composer wrapper (Please, pay attention, composer wrapper is following all rules from https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md) 
     1. If a developer has global-installed copy of composer, he\she just runs one command to get exactly the same result: `composer wrap`
     1. @alcohol  you are absolutely right. The proposed way of using composer won't help for you case. But, pay attention, there are no BC changes - you don't have to use this way if you don't want :) 
2.  The size of composer.phar on today is 1.9M. For php-like project it is not a small size (please, compare with composer wrapper - 8.9K). For CI\CD projects, which  follows "Clean installation" procedure, it is extra effort to checkout 1.9M for each build. Wrapper allows to configure cache 
3.  Also, the composer wrapper will replace custom implementations from https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md

So,  I see, the wrapper will be very useful for composer projects

kamazee:
Also, one of my goals was to make composer installation procedure the same and effortless for both local and CI environments.

@benbor thanks for doing what I procrastinated to do :)

kamazee:
> Wouldn't this rather fragment the composer versions across multiple projects?

Well, not really: everyone will get fresh composer when it's called. By default, it will download a phar per project, though, but it can be easily changed with an environment variable.

> With composer or even in CI builds, it's possible to run composer self-update to make sure the composer file in the Docker container or anything else is up to date

Here are 2 points why I did it and ended up using this in all of my projects in favor of just having phar or global installation:
1. Saving freshness checks: I don't really need composer to check whether it's at the latest version a hundred times a day. Once per week per build is OK. While it can be scheduled somewhere to run self-update, the wrapper does just that without requiring any privileges or any scheduler.
2. It allows just using `./composer` anywhere without worrying about whether it has been installed globally by a someone with superpower or not. It just works. More than that, it makes it work the same way in any environment, whether it's a local developer's machine or a CI server. It's very convenient to just use `./composer` without explaining to CI server maintainer or a new developer how to install it. It spares me from writing more lines in README and everyone from reading those lines.

> Is it? Composer's backwards compatibility (Composer v2 even supports PHP 5.3), and its behavior with global configuration options, global GitHub tokens, and the behavior with it looks up parent directories for a composer.json file suggests me otherwise.

This is true; however, I don't see why it suggests otherwise. I agree that "born to be installed locally" is a bold statement but the it works without being globally installed exceptionally well for fact (by the way, having global -- or, rather, user-wide -- configuration with tokens or package cache is not the same as having it installed globally; using wrapper plays nicely with user-wide token and package cache).

> I think `grunt-cli` resembles the idea of this PR

It doesn't. `grunt-cli` just allows to type `grunt` instead of `node_modules/.bin/grunt`. It neither installs nor updates grunt.
[`gradlew`](https://docs.gradle.org/current/userguide/gradle_wrapper.html) and [`mvnw`](https://github.com/takari/maven-wrapper) do, though: they are indeed very similar. If I happened to use them before I came up with composer-wrapper, I should have said that composer-wrapper is inspired by them :) Still, I used them; I found out that `gradlew` is rather popular among those who use Gradle and those 2 are definitely more mature than composer-wrapper because they do more (e.g. allow to pin a specific version or download from a specified location).

Seldaek:
I am also rather skeptical about this tbh, I find it's much more convenient to have a globally installed composer, which I do on my machine, on all servers I manage, and most CIs also come with Composer out of the box anyway.

IMO it would be possible to publish this as a composer plugin which adds the `wrap` command, if that helps people bootstrap the wrapper, they could then require your plugin in the project, and then run composer wrap.

kamazee:
By the way, there's another interesting case: wrapper can be conveniently located in a homesick/homeshick castle along with dotfiles (they will also add it to the `PATH`): it makes it global(-ish: user-wide, in fact) and always fresh with no effort to install and self-update: this is what I also do. If an environment variable `COMPOSER_DIR` is set, then all the project-wide wrappers will also use the global composer without downloading another one. It's very convenient to use `./composer` without even caring that is has to be installed or kept fresh!

The idea with a plugin is worth exploring IMO: the approach is gradual this way; if it gets traction, including it into core might be revisited.

Seldaek:
Closing in favor of a plugin for people who really want this.

PR #8937, opened 2020-06-02
Optimize rule creation for alias packages

Seldaek:
From a couple tests, this can speed up the actual dependency resolution CPU time quite a bit (got ~50% reduction) but mainly when used with `minimum-stability: dev`, as that leads to having way more alias packages.

Would be good to have more people test this to check it does not have any bad side effects.


Seldaek:
Merging so this can easily be tested with --snapshot builds now. If anyone has a good project with min-stability dev to test with :)

Issue #3508, opened 2014-12-04
Add more tests covering conflict rules


naderman:
After https://github.com/composer/composer/pull/3482 there were reports of a few missing conflict rules for conflicts. So we need to create proper test cases for these situations to find those edge cases, and work out a version of #3482 that does not cause these bugs.

https://github.com/composer/composer/pull/3493
https://github.com/composer/composer/issues/3507

naderman:
This should be covered by installer test cases added in 2.0

PR #8938, opened 2020-06-02
Remove some dead code from RuleSetGenerator

Seldaek:
Found this while looking at https://github.com/composer/composer/issues/3508 and the original change in https://github.com/composer/composer/pull/3482 - IMO #3508 can be closed now as addConflictRules does a single pass on all packages, there shouldn't be duplicate conflicts created anymore.


naderman:
I left #3508 open mostly to actually try and work out test cases for what broke back then which may stil help with avoiding multi conflict rule issues too, but yes the overall problem there was solved, so if you prefer, close it and make a new ticket just for the test cases with links to what happened back then.

Issue #8946, opened 2020-06-03
2.0: issue with ext-zend-opcache


mpesari:
My `composer.json`:

```json
{
  "require": {
    "ext-zend-opcache": "*",
    "vimeo/psalm": "^3.11"
  }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: FAIL
Missing pubkey for tags verification
Missing pubkey for dev verification
Run composer self-update --update-keys to set them up
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (2.0.0-alpha1 => 1.10.7)
Composer version: 2.0.0-alpha1
PHP version: 7.4.3
PHP binary path: /usr/bin/php7.4
OpenSSL version: OpenSSL 1.1.1f  31 Mar 2020
```

Output of `composer show --platform | grep -i opcache`:

```
ext-Zend-OPcache     7.4.3    The Zend OPcache PHP extension
```

When I run this command:

```
$ vendor/bin/psalm
```

I get the following output:

```
Composer detected issues in your platform:

Your Composer dependencies require the following PHP extensions to be installed: zend-opcache
```

And I expected this to happen:

No problems, it works with 1.10.6. `composer install` does not trigger it.

Related (old) issue I found: https://github.com/composer/composer/issues/2489

stof:
Could it be related to the fact that the extension name is not `zend-opcache` but `Zend-OPcache` ?

Ayesh:
It turns out `extension_loaded` is case insensitive, so `extension_loaded('zend opcache')` also returns true.

Seldaek:
Fixed by https://github.com/composer/composer/pull/8950

PR #8950, opened 2020-06-04
Platform Check: Add a special case for `zend-opcache`

Ayesh:
Ref #8946

The platform-check feature maps `ext-X` to `extension_loaded('X')` calls. While most of the extensions can be tested this way, the `zend-opcache` extension requires `zend opcache` to be probed instead of the `zend-opcache` name.

This commit adds a special case for `zend-opcache` to use the correct name in `extension_loaded()` calls in generated `platform_check.php` files.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -640,6 +640,10 @@ EOF;
                         }
                     }
 
+                    if ($match[1] === 'zend-opcache') {
+                        $match[1] = 'zend opcache';
+                    }
+
                     $extension = var_export($match[1], true);
                     if ($match[1] === 'pcntl' || $match[1] === 'readline') {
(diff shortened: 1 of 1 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -643,6 +643,8 @@ protected function getPlatformCheck($packageMap, array $ignorePlatformReqs)
                     $extension = var_export($match[1], true);
```
Seldaek, line 643: I think it'd be shorter and more DRY to have one line above here `if ($match[1] === 'zend-opcache') { $match[1] = 'zend opcache'; }`, then the whole extension_loaded line doesn't need duplicating.

Ayesh: I think you are right, it certainly is more cleaner. Force-pushing that change. Thank you.

Seldaek:
Thanks

PR #8956, opened 2020-06-06
Removing $originUrl parameter from RemoteFileSystem::getRemoteContents()

mchekin:
Removing unused $originUrl parameter from RemoteFileSystem::getRemoteContents()

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -16,7 +16,6 @@ use Composer\Config;
 use Composer\IO\IOInterface;
 use Composer\Downloader\TransportException;
 use Composer\CaBundle\CaBundle;
-use Composer\Util\HttpDownloader;
 use Composer\Util\Http\Response;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -530,13 +529,14 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
...
-     * @param resource $context   The stream context
+     * @param string   $fileUrl         The file URL
+     * @param resource $context         The stream context
+     * @param array    $responseHeaders The HTTP response headers
      *
      * @return string|false The response contents or false on failure
+     *
      */
-    protected function getRemoteContents($originUrl, $fileUrl, $context, array &$responseHeaders = null)
+    protected function getRemoteContents($fileUrl, $context, array &$responseHeaders)
```
staabm, line 539: This is a BC break and might introduce bufs in classes which subclass RemoteFilesystem

mchekin: There are no sub-classes in the repository for this class.

Should we never change protected method signature, even when it contains parameters which are not used inside the body of the method?

Seldaek: Yeah but the point of this method is to be an extension point.. So probably rather not touch it. That said, in 2.0 it's probably less likely to be extended, or really shouldn't be, but IMO not really worth fixing as this code is hopefully going to be deprecated entirely sooner or later.

PR #8962, opened 2020-06-09
Revert "Added docs for glob pattern support in classmaps"

Qdigital:


Seldaek:
If you don't explain why.. then nope.

PR #8973, opened 2020-06-12
Add new summary format for licenses

jdrieghe:
This was inspired by https://www.npmjs.com/package/license-checker which has a `--summary` option to have a list of used licenses and the number of dependencies using each license.

Here's how it looks when executed:
image


Seldaek:
Thanks!

Issue #2960, opened 2014-05-02
Internal Error [ERR_RESULT], exit code 1


srph:
I am using Windows 7 along with xampp 1.8.3 vc11.

For some reasons, I am unable to complete the installation of composer. After selecting the php.exe (xampp 1.8.3 vc11) then going to the next part of the installation, I receive this error.

`Internal Error [ERR_RESULT], exit code 1`

PR #8974, opened 2020-06-13
PHPStan fixes: `autoload_files`, and `ignoreErrors`

Ayesh:
Includes a few improvements in PHPStan CI runs. 
 - Require PHPStan `^0.12.26` (currently `^0.12`) because this version contains major improvements how files are autoloaded. 
 - Remove `autoload_files` from PHPStan `config.neon` because PHPStan no longer needs to autoload files for it to analyze files. 
 - Add `ignoreErrors` pattern for the use of `sapi_windows_set_ctrl_handler` function. The ignore pattern is specific about files, where we already have `function_exists()` calls to make sure this function is available. 
 - Add `ignoreErrors` pattern for `ZipArchive::LIBZIP_VERSION` usage, which is checked before hand. This rule is specific to the single usage we have in Composer. 

With this PR, we bring down PHPStan report from **6 errors and 1 warning** to **1 error**.

The only remaining warning is about the usage of `ZipArchive::OPSYS_UNIX`, which I'm not 100% sure if we should ignore because there doesn't appear to be `defined()` call to safe-guard it.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -89,7 +89,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          extensions: "intl"
+          extensions: "intl, zip"
           ini-values: "memory_limit=-1, phar.readonly=0"
           php-version: "${{ matrix.php-version }}"
 
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -52,5 +52,5 @@ jobs:
 
       - name: Run PHPStan
         run: |
-          bin/composer require --dev phpstan/phpstan:^0.12 phpunit/phpunit:^7.5 --with-all-dependencies
+          bin/composer require --dev phpstan/phpstan:^0.12.26 phpunit/phpunit:^7.5 --with-all-dependencies
           vendor/bin/phpstan analyse --configuration=phpstan/config.neon || vendor/bin/phpstan analyse --configuration=phpstan/config.neon --error-format=checkstyle | cs2pr
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -31,7 +31,7 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          extensions: "intl"
+          extensions: "intl, zip"
           ini-values: "memory_limit=-1"
           php-version: "${{ matrix.php-version }}"
           tools: "cs2pr"
--- a/phpstan/config.neon
+++ b/phpstan/config.neon
@@ -26,15 +26,8 @@ parameters:
         # BC with older PHPUnit
         - '~^Call to an undefined static method PHPUnit\\Framework\\TestCase::setExpectedException\(\)\.$~'
 
-        - # Windows SAPI functions, already checked before using them.
-            message: '~^Function sapi_windows_set_ctrl_handler not found.$~'
-            paths:
-                - ../src/Composer/Command/CreateProjectCommand.php
-                - ../src/Composer/Installer/InstallationManager.php
-
-        - # ZipArchive::LIBZIP_VERSION Class constant is already checked before use.
-            message: '~^Access to undefined constant ZipArchive::LIBZIP_VERSION.$~'
-            path: ../src/Composer/Repository/PlatformRepository.php
+        # ZipArchive::* Class constants are already checked before use.
+        - '~^Access to undefined constant ZipArchive::~'
 
     paths:
         - ../src
--- a/phpstan/config.neon
+++ b/phpstan/config.neon
@@ -32,6 +32,10 @@ parameters:
                 - ../src/Composer/Command/CreateProjectCommand.php
                 - ../src/Composer/Installer/InstallationManager.php
 
+        - # ZipArchive::LIBZIP_VERSION Class constant is already checked before use.
+            message: '~^Access to undefined constant ZipArchive::LIBZIP_VERSION.$~'
+            path: ../src/Composer/Repository/PlatformRepository.php
+
     paths:
         - ../src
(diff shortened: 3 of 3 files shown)


.github/workflows/phpstan.yml
```diff
@@ -52,5 +52,5 @@ jobs:
 
       - name: Run PHPStan
         run: |
-          bin/composer require --dev phpstan/phpstan:^0.12 phpunit/phpunit:^7.5 --with-all-dependencies
+          bin/composer require --dev phpstan/phpstan:^0.12.26 phpunit/phpunit:^7.5 --with-all-dependencies
```
staabm, line 55: Since phpstan updates tend to introduce new errors and therefore break the built, I would suggest using a separate composer.json file for the workflows and let them be managed by dependabot

Ayesh: PHPStan builds are currently run in separate GitHub Actions workflow, and it doesn't break the main unit test builds. I too was a bit surprised to see PHPStan making these changes in a  _seemingly_ minor release. PHPStan is the only package installed ad-hoc like this. I'd wait for a decision maker to decide the way forward if we need to use more static analyzer tools against composer.

`^0.12.26` is used because that is where `autoload_files` is deprecated, and contains major changes in its internal reflection APIs. Effective final package installed by this command remain the current `0.12.28` before and after the PR.

Seldaek:
Regarding the ZipArchiver error for the missing constant, it's a bug in phpstan when the zip ext is not loaded according to #2960 - so maybe we need to make sure zip is included in the list of php extensions in the phpstan workflow. At least you can try that see if it fixes it.

phpstan/config.neon
```diff
@@ -27,6 +25,17 @@ parameters:
...
         # BC with older PHPUnit
         - '~^Call to an undefined static method PHPUnit\\Framework\\TestCase::setExpectedException\(\)\.$~'
+
+        - # Windows SAPI functions, already checked before using them.
+            message: '~^Function sapi_windows_set_ctrl_handler not found.$~'
+            paths:
+                - ../src/Composer/Command/CreateProjectCommand.php
+                - ../src/Composer/Installer/InstallationManager.php
+
+        - # ZipArchive::LIBZIP_VERSION Class constant is already checked before use.
```
Seldaek, line 35: This is tracked by https://github.com/phpstan/phpstan/issues/2971 as it's really a bug in phpstan that it doesn't see we are checking for the constant before using it. So hopefully one day it can be removed..

.github/workflows/phpstan.yml
Seldaek: I think it's fine as is for now, it's a bit annoying when new stuff starts breaking out of the blue but also hopefully not that common an issue..

staabm:
>  src/Composer/Package/Archiver/ZipArchiver.php#L58
Access to undefined constant ZipArchive::OPSYS_UNIX.

reported upstream in  src/Composer/Package/Archiver/ZipArchiver.php#L58

Access to undefined constant ZipArchive::OPSYS_UNIX.

staabm:
> 
> 
> > src/Composer/Package/Archiver/ZipArchiver.php#L58
> > Access to undefined constant ZipArchive::OPSYS_UNIX.
> 
> reported upstream in [phpstan/phpstan#3480](https://github.com/phpstan/phpstan/issues/3480)

phpstubs fix incoming: https://github.com/JetBrains/phpstorm-stubs/pull/837

.github/workflows/continuous-integration.yml
```diff
@@ -86,10 +86,10 @@ jobs:
           fi
 
       - name: "Install PHP"
-        uses: "shivammathur/setup-php@v2"
+        uses: "shivammathur/setup-php@2.3.0"
```
staabm, line 89: `@v2` is a "magic" version which automatically is re-tagged on each release.
so, this change itself is not necessary.

you can let dependabot manage the github actions versions, if you like

Seldaek: Ah thanks fixed

.github/workflows/phpstan.yml
ondrejmirtes: > it's a bit annoying when new stuff starts breaking out of the blue

These aren't intentional BC breaks, just bugs. The 0.12.26 release was a bit more ambitious with a lot of internal changes (it started relying on phpstorm-stubs in some cases), but I tested it before the release on a few projects and it worked okay. I also fixed quickly most of the serious problems in the first 7 days after this release.

I'll keep more in mind in the future that people using open-source value stability over innovation more - nothing wrong with that, but my default mindset is "move fast and break things".

The current 0.12.29 version should be much better than what I shipped in 0.12.26.

ondrejmirtes: Literally any bugfix in PHPStan can use CI builds to fail because of its nature - when I fix a type inference bug, it might realize there's an error in someone's codebase thanks to that. But people usually welcome that the tool is improving and they get new errors discovered in minor and patch versions.

If you don't want this, you should simply pin to a specific version and upgrade when you want to.

Also, I don't see a reason why `phpstan/phpstan` couldn't be in `require-dev` in `composer.json` - it doesn't have to be installed dynamically. That way you could take advantage of dependabot functionality as already suggested here.

Seldaek:
Yeah don't get me wrong I'm not complaining about the fixes and
improvements, they're always welcome :)

I was just trying to offer perspective on the various use cases, not
expecting you to change anything.

.github/workflows/phpstan.yml
Ayesh: I think `phpstan/phpstan` is installed dynamically because Composer still needs to support PHP 5.3, and if we were to add `phpstan/phpstan`, it would fail because PHPStan requires PHP `^7.1`.

ondrejmirtes: Makes sense!

PR #8975, opened 2020-06-14
Composer 2: Allow plugins to override the URL before triggering the download

ffraenz:
I am the maintainer of [private-composer-installer](https://github.com/ffraenz/private-composer-installer) offering users to inject environment values into the dist URLs of their repositories without leaking sensitive information into `composer.lock`. In the next major version ([pull request](https://github.com/ffraenz/private-composer-installer/pull/25)) I'd like to add support for Composer 2.

Currently the plugin replaces the remote file system to alter the dist URL before the download gets triggered. As explained in the [upgrade guide](https://github.com/composer/composer/blob/master/UPGRADE-2.0.md) this will no longer work in Composer 2.

As already suggested in https://github.com/composer/composer/pull/8807#issuecomment-617864883 I'd like to implement a way for plugins to alter the URL before the actual download gets triggered using the existing `PRE_FILE_DOWNLOAD` event. I'd like to suggest adding the method `setProcessedUrl` to the event object where the URL gets fed back to the caller before triggering the actual download.

I could not find existing test cases for events so I was unsure about how to implement unit tests for this change.

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -125,7 +125,7 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
             $url = reset($urls);
 
             if ($eventDispatcher) {
-                $preFileDownloadEvent = new PreFileDownloadEvent(PluginEvents::PRE_FILE_DOWNLOAD, $httpDownloader, $url['processed']);
+                $preFileDownloadEvent = new PreFileDownloadEvent(PluginEvents::PRE_FILE_DOWNLOAD, $httpDownloader, $url['processed'], 'package', $package);
                 $eventDispatcher->dispatch($preFileDownloadEvent->getName(), $preFileDownloadEvent);
                 $url['processed'] = $preFileDownloadEvent->getProcessedUrl();
             }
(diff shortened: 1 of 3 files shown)


src/Composer/Plugin/PreFileDownloadEvent.php
```diff
@@ -55,12 +55,22 @@ public function getHttpDownloader()
...
      */
     public function getProcessedUrl()
     {
         return $this->processedUrl;
     }
+
+    /**
+     * Sets the processed URL that will be downloaded
+     *
+     * @return string
```
Seldaek, line 70: Should be a param annotation not return

src/Composer/Downloader/FileDownloader.php
```diff
@@ -127,6 +127,7 @@ public function download(PackageInterface $package, $path, PackageInterface $pre
             if ($eventDispatcher) {
                 $preFileDownloadEvent = new PreFileDownloadEvent(PluginEvents::PRE_FILE_DOWNLOAD, $httpDownloader, $url['processed']);
```
Seldaek, line 128: I think it would be good to add some additional context data here like a $type (could be just a string, here 'package' and below 'metadata') and a mixed $details property which could contain here the package object. That might allow plugins to make more informed decisions about what the URL is. If you can add that into PreFileDownloadEvent that'd be great.

Seldaek:
Ignore the PHPStan failures about ZipArchive btw, I'll take care of those on the master branch.

src/Composer/Plugin/PreFileDownloadEvent.php
staabm, line 70:
```suggestion
     * @param string $processedUrl
```

src/Composer/Downloader/FileDownloader.php
ffraenz: Indeed, that will result in more supported use cases using this hook. I've added two new properties and connected them up like you suggested.

Seldaek:
Thanks!

ffraenz:
Awesome, Thank you!

PR #8980, opened 2020-06-15
Discard changes is incomplete if Git project has additional folders

stucki:
Problem:
A package which is installed using Git is not cleaned up completely during `composer install`.

How to reproduce:
1. Install a package as a Git repository (including `.git` folder)
1. Make some modifications to the package (incl. adding new files and folders)
1. Update the package + confirm discard notice. Composer warns about the modified files and asks if these changes should be discarded.
1. The result is that all files that were modified have been reverted (`git reset --hard`), but any files which are not part of the Git repository are still there.

Expected result:
Git project folders should be clean when choosing "discard changes".

Solution:
Run `git clean -df` when attempting to discard local changes.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -451,7 +451,7 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
     protected function discardChanges($path)
     {
         $path = $this->normalizePath($path);
-        if (0 !== $this->process->execute('git reset --hard', $output, $path)) {
+        if (0 !== $this->process->execute('git clean -df && git reset --hard', $output, $path)) {
             throw new \RuntimeException("Could not reset changes\n\n:".$this->process->getErrorOutput());
         }
 


Seldaek:
Sounds good, we already do the same with --include-untracked for the stash option. Thanks

staabm:
>  src/Composer/Repository/PlatformRepository.php#L234
Access to undefined constant ZipArchive::LIBZIP_VERSION.

created a upstream issue: https://github.com/JetBrains/phpstorm-stubs/pull/839

Seldaek:
@staabm ?! :)

staabm:
in the "File changed"-Tab of this PR is a phpstan error reported, which is a false positive:

grafik

the PR linked above tries to fix this error upstream

Seldaek:
Ok yeah but it's been ignored as I fixed this earlier, and this PR is merged.. so thanks I guess but you don't need to bother repeating this on every PR based on an outdated master :)

stucki:
@Seldaek What is the workflow to get this change into 1.10? Should I create a new PR for it?

Seldaek:
You're in luck as I am anyway busy with 1.10 stuff, so I will quickly backport this as it's a small one, but generally speaking 1.10 is maintenance only and 2.0 is the target for all new things.

stucki:
@Seldaek Many thanks!

PR #8990, opened 2020-06-17
Add major version install to programmatically install script example

gisostallenberg:
Add info on how to install a specific major version of composer as added in https://github.com/composer/getcomposer.org/pull/156.
This example script should be updated once the new major is released.
```diff
--- doc/faqs/how-to-install-composer-programmatically.md
+++ doc/faqs/how-to-install-composer-programmatically.md
@@ -20,7 +20,7 @@ then
     exit 1
 fi

-php composer-setup.php --quiet --${COMPOSER_MAJOR_VERSION:-1}
+php composer-setup.php --quiet --${COMPOSER_MAJOR_VERSION:-2}
 RESULT=$?
 rm composer-setup.php
 exit $RESULT
```

### DIFF
--- a/doc/faqs/how-to-install-composer-programmatically.md
+++ b/doc/faqs/how-to-install-composer-programmatically.md
@@ -20,7 +20,7 @@ then
     exit 1
 fi
 
-php composer-setup.php --quiet
+php composer-setup.php --quiet --${COMPOSER_MAJOR_VERSION:-1}
 RESULT=$?
 rm composer-setup.php
 exit $RESULT


GrahamCampbell:
I think this PR was sent to the wrong branch. Should target 1.10?

gisostallenberg:
I rebased it.

Seldaek:
Thanks but I think I'd rather not have this in the docs like that, as it's gonna be outdated soonish anyway, and anyway wanting to specify a --flag can just do so, no need for a new env var etc.

PR #8993, opened 2020-06-19
GitDriver: use authentication for supports check

glaubinix:
The GitDriver wasn't using authentication when trying to determine whether a URL is a git repository. The `GitDriver::supports` methods therefore returned always false for private git repositories which for instance look like regular URLs.

### DIFF
--- a/src/Composer/Repository/Vcs/GitDriver.php
+++ b/src/Composer/Repository/Vcs/GitDriver.php
@@ -219,8 +219,17 @@ class GitDriver extends VcsDriver
             return false;
         }
 
-        $process = new ProcessExecutor($io);
+        $gitUtil = new GitUtil($io, $config, new ProcessExecutor($io), new Filesystem());
+        GitUtil::cleanEnv();
 
-        return $process->execute('git ls-remote --heads ' . ProcessExecutor::escape($url), $output) === 0;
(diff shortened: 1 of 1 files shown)


src/Composer/Repository/Vcs/GitDriver.php
```diff
@@ -219,8 +219,17 @@ public static function supports(IOInterface $io, Config $config, $url, $deep = f
...
         }
 
-        $process = new ProcessExecutor($io);
+        $gitUtil = new GitUtil($io, $config, new ProcessExecutor($io), new Filesystem());
+        GitUtil::cleanEnv();
+
+        try {
+            $gitUtil->runCommand(function ($url) {
+                return 'git ls-remote --heads ' . ProcessExecutor::escape($url);
+            }, $url, null);
```
Seldaek, line 228: Probably should be `sys_get_temp_dir()` instead of `null` to avoid running this accidentally inside a git repo, or add a new method on git util to run a command without `git remote -v` in the directory before, like runRemoteCommand for stuff that doesn't need or want a local git repo?

glaubinix: 👍 changed it to use `sys_get_temp_dir()`

Issue #7051, opened 2018-01-29
hangs at `Resolving dependencies through SAT`


kimkit:
My `composer.json`:

```json
{
    "require": {
        "laravel/lumen-framework": "5.2.*",
        "illuminate/session": "5.2.*",
        "illuminate/cookie": "5.2.*",
        "baryshev/tree-route": "2.0.*",
        "qiniu/php-sdk": "7.0.*",
        "endroid/qrcode": "1.7.*",
        "geoip2/geoip2": "^2.5",
        "hashids/hashids": "^2.0",
        "elasticsearch/elasticsearch": "^5.2",
        "fguillot/json-rpc": "1.2.*",
        "nikic/php-parser": "^2.0",
        "symfony/yaml": "^3.0",
        "michelf/php-markdown": "^1.7",
        "guzzlehttp/guzzle": "^6.2",
        "phpunit/phpunit": "^5.7",
        "friendsofphp/php-cs-fixer": "^2.10"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 1.6.2
PHP version: 7.1.11
PHP binary path: /usr/local/Cellar/php71/7.1.11_22/bin/php
```

When I run this command:

```
composer update -vvv
```

I get the following output:

```
Reading ./composer.json
Loading config file /Users/kimkit/.composer/config.json
Loading config file /Users/kimkit/.composer/auth.json
Loading config file ./composer.json
Checked CA file /usr/local/etc/openssl/cert.pem: valid
Executing command (/Users/kimkit/workspace/composer_hangs): git branch --no-color --no-abbrev -v
Executing command (/Users/kimkit/workspace/composer_hangs): git describe --exact-match --tags
[...]

PR #8999, opened 2020-06-19
Fix solver bug exception 7051

naderman:
First attempt at fixing https://github.com/composer/composer/issues/7051

This now appears to lead to the same problem description as composer v1 for the test case provided in https://github.com/composer/composer/issues/7051#issuecomment-637745282

However this definitely needs a bit more work, e.g. I don't think the increment of `$num` by exactly 1 necessarily always works. There may also be necessary changes to `$l1num` depending on the level the multi conflict rule is found at. There may be other problems with this workaround.

### DIFF
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -471,8 +471,20 @@ class Solver
                 if ($rule instanceof MultiConflictRule) {
                     // there is only ever exactly one positive decision in a multiconflict rule
                     foreach ($rule->getLiterals() as $literal) {
-                        if (!isset($seen[abs($literal)]) && !$this->decisions->satisfy($literal)) {
-                            $num++;
+                        if (!isset($seen[abs($literal)]) && $this->decisions->satisfy(-$literal)) {
+                            $l = $this->decisions->decisionLevel($literal);
+                            if (1 === $l) {
+                                $l1num++;
+                            } elseif ($level === $l) {
+                                $num++;
+                            } else {
+                                // not level1 or conflict level, add to new rule
+                                $learnedLiterals[] = $literal;
+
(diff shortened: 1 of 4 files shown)


Seldaek:
Might be good to include the test case from https://github.com/composer/composer/issues/7051#issuecomment-637745282 too as a .test file?

Also I believe https://github.com/composer/composer/pull/8733 can be closed then, or is it really a different PR?

naderman:
For the record this is separate from 8733, and as per title still WIP, removed the wrong test case now, adding the other one and then working on the rest of the todos here.

naderman:
I think this should be good to go now @Seldaek 

Conflict error output for complex cases like this definitely needs some work but I think that's a separate issue.

src/Composer/DependencyResolver/Solver.php
```diff
@@ -459,6 +465,35 @@ protected function analyze($level, Rule $rule)
                     $l1num++;
                     $l1retry = true;
                 }
+
+                $decision = $this->decisions->atOffset($decisionId);
+                $rule = $decision[Decisions::DECISION_REASON];
```
Seldaek, line 470: Just to be completely sure, it's by design you overwrite $rule here? I guess it doesn't matter as it gets reassigned after the foreach anyway, and it is otherwise not used within the foreach during further occurrences. But I'm wondering if it wouldn't be safer to have another variable name here.

naderman: This is intentional. `$rule` essentially tracks the current rule that is being analyzed. Usually the `$l1retry` while loop doesn't move to a new rule, so the assignment within the foreach on line 465 is enough to ensure the rule gets overwritten, but you could really at any point throughout this function reset `$rule` to whatever the current `$decisionId` references and it'd work correctly.

PR #9008, opened 2020-06-22
Replace cascading str_replace() calls with strtr()

Ayesh:
Replaces cascading `str_replace()` calls with a single `strtr()` call that makes it more readable and provides a micro performance improvement.
Some instances of cascading `str_replace()` calls were not refactored due to the potential recurring replacement patterns.

Thank you.


Seldaek:
Thanks

PR #9010, opened 2020-06-23
Sort licenses summary

jdrieghe:
### Fixed
- Used licenses in the summary overview are now correctly sorted, showing the most used license first
- Output of `composer help licenses` now shows the summary format option just like the regular docs.

---
I recently added the feature in https://github.com/composer/composer/pull/8973 and while playing around with the latest snapshot version of `2.x` I found these two small improvements.

### DIFF
--- a/src/Composer/Command/LicensesCommand.php
+++ b/src/Composer/Command/LicensesCommand.php
@@ -34,7 +34,7 @@ class LicensesCommand extends BaseCommand
             ->setName('licenses')
             ->setDescription('Shows information about licenses of dependencies.')
             ->setDefinition(array(
-                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the output: text or json', 'text'),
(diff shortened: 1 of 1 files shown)


src/Composer/Command/LicensesCommand.php
```diff
@@ -108,23 +108,26 @@ protected function execute(InputInterface $input, OutputInterface $output)
                     'name' => $root->getPrettyName(),
                     'version' => $root->getFullPrettyVersion(),
                     'license' => $root->getLicense(),
-                    'dependencies' => $dependencies,
+                    'dependencies' => $usedLicenses,
```
apfelbox, line 111: This seems to be a wrong rename here in the `case 'json'` branch?

jdrieghe: Oh wow, good spot. Looks like the automated refactoring tools were a bit too eager. Fixing.

Seldaek:
Thanks

PR #9042, opened 2020-07-10
GitLab: clarify interactive auth prompt

glaubinix:
This PR is trying to improve the OAuth token prompt which appears if a user tries to install a VCS repository from GitLab without having any auth setup:

```
Installing acme-company/gitlab-test (dev-master jfdjhfdjhd): Cloning failed, enter your GitLab credentials to access private repos
A token will be created and stored in "/path/to/composer/project/auth.json", your password will never be stored
To revoke access to this token you can visit gitlab.com/profile/applications
Username: my-gitlab-user
Password:
Bad credentials.
You can also manually create a personal token at https://gitlab.com/profile/personal_access_tokens
```

Changes:
* The prompt now shows a message if the authentication failed because of 2FA which prevents OAuth tokens creation with the password grant
* GitLab has now quite a few [scopes](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#limiting-scopes-of-a-personal-access-token) and from the prompt it is not clear which scope one needs to enable.

Note: the automatically created OAuth token doesn't provide a scope which makes GitLab grant the "api" scope. The "read_api" scope was only introduced with GitLab 12.10 and several self hosted GitLab versions might still be running on older versions but in the long run it might probably be a good idea to adjust this as Composer only ever fetches data and never modifies it.

### DIFF
--- a/src/Composer/Util/GitLab.php
+++ b/src/Composer/Util/GitLab.php
@@ -104,7 +104,7 @@ class GitLab
         }
 
         $this->io->writeError(sprintf('A token will be created and stored in "%s", your password will never be stored', $this->config->getAuthConfigSource()->getName()));
-        $this->io->writeError('To revoke access to this token you can visit '.$originUrl.'/profile/applications');
+        $this->io->writeError('To revoke access to this token you can visit '.$scheme.'://'.$originUrl.'/profile/applications');
 
         $attemptCounter = 0;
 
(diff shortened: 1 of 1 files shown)


Ayesh:
Not related to this PR, PHP 8 lowest-ignore test fails because [match expression](https://php.watch/versions/8.0/match-expression) is now merged to PHP master.

Seldaek:
Probably should be targetting the 1.10 branch

Seldaek:
Thanks

PR #9054, opened 2020-07-16
Fix tests on PHP 8.0

GrahamCampbell:
We need PHPUnit 9.3@dev (and its @dev stability dependencies), which include PHP 8.0 support, including https://github.com/sebastianbergmann/phpunit/pull/4374 which fixes the issue with Match which is currently crashing composer's test suite on actions.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -95,17 +95,7 @@ jobs:
           extensions: "intl, zip"
           ini-values: "memory_limit=-1, phar.readonly=0"
           php-version: "${{ matrix.php-version }}"
-
-      - name: "Determine composer cache directory"
-        id: "determine-composer-cache-directory"
-        run: "echo \"::set-output name=directory::$(composer config cache-dir)\""
-
-      - name: "Cache dependencies installed with composer"
-        uses: "actions/cache@v2"
-        with:
-          path: "${{ steps.determine-composer-cache-directory.outputs.directory }}"
-          key: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}-${{ hashFiles('**/composer.lock') }}"
-          restore-keys: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}"
+          tools: composer:v2
 
       - name: "Handle lowest dependencies update"
         if: "contains(matrix.dependencies, 'lowest')"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -38,31 +38,31 @@ jobs:
         os: [ubuntu-latest]
         experimental: [false]
         include:
-          - php-version: 5.3
+          - php-version: "5.3"
             dependencies: highest
             os: ubuntu-latest
             experimental: false
-          - php-version: 5.3
+          - php-version: "5.3"
             dependencies: lowest
             os: ubuntu-latest
             experimental: false
-          - php-version: 7.4
+          - php-version: "7.4"
             dependencies: highest
             os: ubuntu-latest
             experimental: false
-          - php-version: 7.4
+          - php-version: "7.4"
             os: windows-latest
             dependencies: locked
             experimental: false
-          - php-version: 7.4
+          - php-version: "7.4"
             os: macos-latest
             dependencies: locked
             experimental: false
-          - php-version: 8.0
+          - php-version: "8.0"
             dependencies: lowest-ignore
             os: ubuntu-latest
             experimental: true
-          - php-version: 8.0
+          - php-version: "8.0"
             dependencies: highest-ignore
             os: ubuntu-latest
             experimental: true
@@ -83,7 +83,7 @@ jobs:
           elif [ "${{ matrix.php-version }}" = "7.1" ]; then
             echo "::set-env name=SYMFONY_PHPUNIT_VERSION::7.5";
           elif [ "${{ matrix.php-version }}" = "8.0" ]; then
-            echo "::set-env name=SYMFONY_PHPUNIT_VERSION::9.2";
+            echo "::set-env name=SYMFONY_PHPUNIT_VERSION::9.3";
           else
             echo "::set-env name=SYMFONY_PHPUNIT_VERSION::8.3";
           fi
@@ -95,14 +95,6 @@ jobs:
           extensions: "intl, zip"
           ini-values: "memory_limit=-1, phar.readonly=0"
           php-version: "${{ matrix.php-version }}"
-
-      - name: "Update to latest Composer snapshot"
-        if: "!startsWith(matrix.os, 'windows')"
-        run: "sudo -i composer self-update --snapshot"
-
-      - name: "Update to latest Composer snapshot on Windows"
-        if: "startsWith(matrix.os, 'windows')"
-        run: "composer self-update --snapshot"
 
       - name: "Determine composer cache directory"
         id: "determine-composer-cache-directory"
@@ -136,12 +128,10 @@ jobs:
 
       - name: "Require latest PHPUnitBridge for PHP 8"
         if: "matrix.php-version == '8.0'"
-        run: |
-          composer require --no-update --dev symfony/phpunit-bridge:^5.1
-          composer config -g platform-check false
+        run: composer require --no-update --dev symfony/phpunit-bridge:5.1.*@dev
 
       - name: "Update Symfony's PHPUnitBridge to latest available for the current PHP always as it is not really a dependency of the project"
-        run: "composer update ${{ env.COMPOSER_FLAGS }} symfony/phpunit-bridge"
+        run: "composer update ${{ env.COMPOSER_FLAGS }} symfony/phpunit-bridge --with-all-dependencies"
 
       - name: "Run install again using composer binary from source"
         run: "bin/composer install ${{ env.COMPOSER_FLAGS }}"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -118,10 +118,10 @@ jobs:
 
       - name: "Require latest PHPUnitBridge for PHP 8"
         if: "matrix.php-version == '8.0'"
-        run: composer require --no-update --dev symfony/phpunit-bridge:5.1.*@dev
+        run: composer require --no-update --dev symfony/phpunit-bridge:^5.1
 
       - name: "Update Symfony's PHPUnitBridge to latest available for the current PHP always as it is not really a dependency of the project"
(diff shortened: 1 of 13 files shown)


GrahamCampbell:
gahhhhhhh

image

GrahamCampbell:
Why is simple-phpunit ignoreing the environment variable...

.github/workflows/continuous-integration.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           elif [ "${{ matrix.php-version }}" = "7.1" ]; then
             echo "::set-env name=SYMFONY_PHPUNIT_VERSION::7.5";
           elif [ "${{ matrix.php-version }}" = "8.0" ]; then
```
Seldaek, line 84: This needs to be `"8"` as somehow that's the version 8.0 gets set to.

composer.json
```diff
@@ -86,5 +86,7 @@
     "support": {
         "issues": "https://github.com/composer/composer/issues",
         "irc": "irc://irc.freenode.org/composer"
-    }
+    },
+    "minimum-stability": "dev",
+    "prefer-stable": true
```
Seldaek, line 91: This shouldn't be in composer.json IMO, should be set via `composer config` above at line 139 or so

.github/workflows/continuous-integration.yml
GrahamCampbell: Oh, it's because you didn't quote it, so it gets read as a float, then cast to a string for comparisons.

composer.json
GrahamCampbell: > The minimum-stability property can not be set in the global config.json file. Use `composer global config` to apply changes to the global composer.json

Seldaek: Ah yeah you need this to be in the phpunit-bridge composer.json, AFAIK there is no way to do this.. /cc @nicolas-grekas

GrahamCampbell: I raised a PR which explicitly shows this hole: https://github.com/symfony/symfony/pull/37595.

GrahamCampbell: Hmm, there should be a way to get this to be triggered... https://github.com/symfony/phpunit-bridge/blob/v5.1.2/bin/simple-phpunit.php#L191

GrahamCampbell: Notice the `-s dev` on that line

GrahamCampbell: I don't think their code works properly... They intended line 168 to get the dev version when there was no tag yet, but in fact composer says there is no version...

GrahamCampbell: I mean, the script there flat out just doesn't work for dev versions, and never did. I don't understand why they have code in there that must have never been tested.

GrahamCampbell: I think I will just temporarily replace the script for the purposes of this PR, then we can remove the hack once 9.3.0 is tagged.

Seldaek: `composer show -a phpunit/phpunit 9.3.*` shows this for me:

```
name     : phpunit/phpunit
descrip. : The PHP Unit Testing framework.
keywords : testing, phpunit, xunit
versions : dev-master, 9.3.x-dev
```

The problem is the check at line 190 IMO, as in this case there are two versions (master + the alias), and it expects exactly 1 for the dev switch.

GrahamCampbell: they don't call it with the `.*`?

GrahamCampbell: ah, wait

GrahamCampbell: well, there's 2 versions there, not one

GrahamCampbell: Which version of composer do you have?

Seldaek: https://github.com/symfony/phpunit-bridge/blob/v5.1.2/bin/simple-phpunit.php#L168 sure has a `.*` following $PHPUNIT_VERSION

GrahamCampbell: composer 2.0alpha2 shows me nothing when I run that command

GrahamCampbell: Composer 1 also shows me nothing?

image

Seldaek: This is not a chat, breath once before hitting send :p

I am on latest master but get the same with alpha2, seems like composer 2 is breaking their script actually. With 1.10 I get only `versions : 9.3.x-dev`.

Seldaek: You're quoting it so you pass the whole thing to the package name argument, remove the quotes and it'll work

GrahamCampbell: Ah, yes (and my quotes were in the wrong place, which I why I was seeing nothing).

GrahamCampbell: I guess their logic should be changed to look for non-dev versions, rather than to count the versions. Counting was wrong anyway, because there could be one actual tag but no dev branches.

GrahamCampbell: I think we can simply use composer v1 to execute simple-phpunit then, and we should be sorted, for now.

.github/workflows/continuous-integration.yml
```diff
@@ -137,10 +129,11 @@ jobs:
         if: "matrix.php-version == '8.0'"
         run: |
           composer require --no-update --dev symfony/phpunit-bridge:^5.1
-          composer config -g platform-check false
```
GrahamCampbell, line 140: We actually don't need this anymore anyway, since composer is no longer strict about upper bounds.

GrahamCampbell:
Ok, got the tests to run now. Looks like there are some issues (mostly due to deprecated usage of phpunit causing real issues in 9.3.x or php 8.0 changing exception message text, from what I can see at a quick glance).

Seldaek:
Yeah given the amount of deprecation warnings we had on phpunit 8, I was pretty sure it wouldn't run on 9.. I doubt this can be fixed without adding a huge amount of overrides for assertions and a huge mess.. So I am not sure what's best tbh, might be better to ignore php8 for now, and hopefully we can then in composer 2.1 target php8 more easily once we drop older php releases. I don't think it's realistic to support 5.3 all the way up given phpunit and phpunit-bridge restrictions.

GrahamCampbell:
The PHP ecosystem already relies on composer 2 working on PHP 8.0. I think we should try and update the test suite to work for both versions. I can try and work on it today... but we should also ask for help from others. If they wanna use composer on PHP 8.0, then they gotta put in their share. ;)

Seldaek:
Well Composer works fine on php8, the test suite doesn't now because of phpunit and co.. If someone puts in the work and it's not creating an unreasonable mess in the test suite I'll happily take it in, but otherwise IMO it can wait, it runs well enough for now, and php8 isn't out yet.

GrahamCampbell:
Another approach would be to patch phpunit 8 to fix the match thing. ;)

Seldaek:
That'd be much easier (for us) but I have a feeling @sebastianbergmann may not see it the same?

GrahamCampbell:
Yeh, I think so, and he has good reasons. For one, that might not be the only PHP 8.0 compatibility issue we run into, so the amount of stuff we might want backporting is potentially unbounded (it is certainly not the only PHP 8.0 compatibly issue within phpunit and its dependencies - there were a lot of other changes in particular around reflection of types).

sebastianbergmann:
> That'd be much easier (for us) but I have a feeling @sebastianbergmann may not see it the same?

I am sorry, but PHPUnit 8.5 and PHPUnit 9.2 will not be patched to support PHP 8.

GrahamCampbell:
The tests are running now, at least, and there do seem to be some genuine issues (but also some issues due to using old phpunit methods).

.github/workflows/continuous-integration.yml
```diff
@@ -135,12 +117,10 @@ jobs:
 
       - name: "Require latest PHPUnitBridge for PHP 8"
         if: "matrix.php-version == '8.0'"
-        run: |
-          composer require --no-update --dev symfony/phpunit-bridge:^5.1
-          composer config -g platform-check false
+        run: composer require --no-update --dev symfony/phpunit-bridge:5.1.*@dev
```
Seldaek, line 120: Why `@dev`, is that still needed at this point?

.github/workflows/continuous-integration.yml
```diff
@@ -135,12 +117,10 @@ jobs:
...
 
       - name: "Update Symfony's PHPUnitBridge to latest available for the current PHP always as it is not really a dependency of the project"
-        run: "composer update ${{ env.COMPOSER_FLAGS }} symfony/phpunit-bridge"
+        run: "composer update ${{ env.COMPOSER_FLAGS }} symfony/phpunit-bridge --with-all-dependencies"
```
Seldaek, line 123: Why the new flag? It'd allow updating symfony components we rely on, I am not sure that's a good idea as it breaks the validity of locked dependencies builds.

.github/workflows/continuous-integration.yml
```diff
@@ -94,25 +94,7 @@ jobs:
...
-      - name: "Determine composer cache directory"
-        id: "determine-composer-cache-directory"
-        run: "echo \"::set-output name=directory::$(composer config cache-dir)\""
-
-      - name: "Cache dependencies installed with composer"
-        uses: "actions/cache@v1"
-        with:
-          path: "${{ steps.determine-composer-cache-directory.outputs.directory }}"
-          key: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}-${{ hashFiles('**/composer.lock') }}"
-          restore-keys: "php-${{ matrix.php-version }}-symfony-php-unit-version-${{ env.SYMFONY_PHPUNIT_VERSION }}"
```
Seldaek, line 115: Why remove the cache of composer cache?

GrahamCampbell: Because using the cache is slower than letting composer do parallel downloads.

.github/workflows/continuous-integration.yml
GrahamCampbell: That was needed because of old dependencies in the composer.lock, at the time.

.github/workflows/continuous-integration.yml
GrahamCampbell: Not needed any more.

Seldaek:
Ok thanks for the feedback. Looks better now, will see what I can do about test failures later this week.

.github/workflows/continuous-integration.yml
```diff
@@ -117,10 +117,10 @@ jobs:
 
       - name: "Require latest PHPUnitBridge for PHP 8"
         if: "matrix.php-version == '8.0'"
-        run: composer require --no-update --dev symfony/phpunit-bridge:5.1.*@dev
+        run: composer require --no-update --dev symfony/phpunit-bridge:^5.1
```
GrahamCampbell, line 121: Not sure we even need this now. Can we just always use `^5.1`?

Seldaek: Nope that doesn't support 5.3 so we need an older version in composer.json.

Seldaek:
Ok we are down to one weird syntax error bug which I assume is phpunit-bridge based, and one actual failure due to php8 sorting order in TransactionTest.. so I'll merge this for now it's already improving things a lot. Thanks

GrahamCampbell:
Thanks for finishing up this PR. 🍻

PR #9056, opened 2020-07-16
Create project event dispatcher

logicalor:
This is a patch to allow plugins which hook into the [pre-file-download plugin event](https://getcomposer.org/doc/articles/scripts.md#plugin-events) to interact with the create-project command. I created the patch as a result of an issue I found while using @naderman 's [composer-aws plugin](https://github.com/naderman/composer-aws) in combination with composer's create-project command ([issue here](https://github.com/naderman/composer-aws/issues/19)).


Seldaek:
This should be rebased on the master branch, I'm not sure why it contains my 1.10.9 release commit..

Seldaek:
Sorry should be rebased on 1.10 rather, and the PR should be edited to target 1.10 too.

logicalor:
> Sorry should be rebased on 1.10 rather, and the PR should be edited to target 1.10 too.

Thanks, updated.

logicalor:
Going to resubmit.

PR #9057, opened 2020-07-16
Supply event dispatcher in Create Project Command

logicalor:
Apologies for the previous (bungled) PR.

This is a patch to allow plugins which hook into the [pre-file-download plugin event](https://getcomposer.org/doc/articles/scripts.md#plugin-events) to interact with the create-project command. I created the patch as a result of an issue I found while using @naderman 's [composer-aws plugin](https://github.com/naderman/composer-aws) in combination with composer's create-project command ([issue here](https://github.com/naderman/composer-aws/issues/19)).

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -39,6 +39,7 @@ use Composer\Json\JsonFile;
 use Composer\Config\JsonConfigSource;
 use Composer\Util\Filesystem;
 use Composer\Package\Version\VersionParser;
+use Composer\EventDispatcher\EventDispatcher;
 
 /**
  * Install a package as new project into new directory.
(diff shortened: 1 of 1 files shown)


Seldaek:
No worries :) Looks good I think but I'll try to review more in depth tomorrow.

src/Composer/Command/CreateProjectCommand.php
```diff
@@ -286,8 +287,12 @@ protected function installRootPackage(IOInterface $io, Config $config, $packageN
             $config->merge(array('config' => array('secure-http' => false)));
         }
 
+        $composer = Factory::create($io, $config->all(), $disablePlugins);
```
logicalor, line 290: I'm wondering if this can be done better.  I originally had lines 285-290 including this instantiation of $composer on line 157, before the call to installRootPackage, and removed the re-instantiation of $composer on line 182.  However, loading it with the $config->all() caused problems further down (lines 168 on) so I moved it to a separate instantiation here.

Seldaek: It needs to be reinstantiated there to read the composer.json from the root package which just got installed by this method, so I think it's fine as you did it really. It's a bit messy  here but it's a really special case as composer isn't very well made to run without composer.json.

Seldaek:
Found a couple simplifications made possible by the instantiation of that complete Composer instance you added, see https://github.com/composer/composer/commit/eac03e16e7213e2ddc7e73818e91a0f21e571bda - but otherwise I think it looks good. Thanks!

logicalor:
Excellent - happy to help!

PR #9058, opened 2020-07-16
Clean up Zip Util to be more strict about what is a valid package archive

### LINKED ISSUE #8931: Where to place the composer.json file for Artifact?
Looking at the Artifact docs, it mention that the composer.json file need to be at the root directory. But while looking at the `Composer\Util\Zip::getComposerJson($pathToZip)` it's working differently: it searches in the root directory for a composer.json but if it doesn't find it there it keeps searching until it finds the file with the shortest path.

Is that the expected behaviour? Or how should that work?

Seldaek:
Fixes #8931


naderman:
@wissem looks like Jordi already implemented a similar solution for Composer now. Can you check if this really matches what we have in Private Packagist and if this should be unified further? Does the test suite here miss any of the private packagist examples?

wissem:
that looks like it covers most of what we were trying to implement in private packagist. What's missing here is that in Private Packagist we throw an  exception in case there are multiple composer.json files found in top directories and no composer.json in the root directory.

naderman:
@wissem can you provide a test case which covers that and fails and send it as a PR against this PR?

Seldaek:
I think this is already covered here, as a file with something like:
```
foo/composer.json
bar/composer.json
```
Or even just:

```
foo/composer.json
bar/empty
```

Both of these will return null as in no valid package. The only things valid are `./composer.json` or a single dir on top level which contains directly a composer.json.

Seldaek:
The main difference perhaps is this returns null without details as to why the file is ignored/invalid, no exception with info.

naderman:
Ah I see, that would indeed be useful to have so we could replace our own implementation with simply using the Composer class?

src/Composer/Util/Zip.php
```diff
@@ -71,37 +71,41 @@ public static function getComposerJson($pathToZip)
      */
     private static function locateFile(\ZipArchive $zip, $filename)
     {
-        $indexOfShortestMatch = false;
-        $lengthOfShortestMatch = -1;
+        // return root composer.json if it is there and is a file
+        if (false !== ($index = $zip->locateName($filename)) && $zip->getFromIndex($index) !== false) {
```
GrahamCampbell, line 75: yoda naming on one side, but not the other ;)

Seldaek: Yet more symmetrical, it is. The beauty in it, you should see.

GrahamCampbell: :P

PR #9059, opened 2020-07-16
Remove cygwin php handling in UAC elevation

johnstevenson:
When called from a native shell and using cygwin PHP, cygpath translates
`/tmp` to `User\AppData\Local\Temp`, rather than `/cygdrive/.../tmp`.
This change does not affect using windows PHP from a Cygwin shell.

## Testing
As before, using this compiled composer.phar: [composer-1.10-dev+8ddbae3.zip](https://github.com/composer/composer/files/4933043/composer-1.10-dev%2B8ddbae3.zip) (incidentally, it is still not possible to compile on Windows as seld/phar-utils needs to be at 1.1.1)

On Windows, create a test folder in C:\Program Files and copy this composer.phar here:

```cmd
# From an Administrator cmd prompt
mkdir "C:\Program Files\test"
copy /B <path to>\composer.phar "C:\Program Files\test\composer.phar"
```
To test:

```cmd
# From a User cmd prompt
cd "C:\Program Files\test"
php composer.phar self-update

# bonus test, now that 1.10.9 is a thing
php composer.phar self-update --rollback
```

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -380,8 +380,8 @@ TAGSPUBKEY
             // see if we can run this operation as an Admin on Windows
             if (!is_writable(dirname($localFilename))
                 && $io->isInteractive()
-                && $this->isWindowsNonAdminUser($isCygwin)) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, I updated deps now in 1.10 branch sorry I forgot about that.

PR #9062, opened 2020-07-20
Fix path translation when MSYS_NO_PATHCONV=1

luka8088:
Under windows git bash the following produces an unexpected result:

```bash
$ MSYS_NO_PATHCONV=1 vendor/bin/some-package-script
Could not open input file: /c/Users/me/project/vendor/some-package-script/some-package-script
```

It is due to `MSYS_NO_PATHCONV` disabling automatic path translations.
This patches detects such cases and forces the path translation so that the example command works as expected.

Reference:
https://github.com/git-for-windows/msys2-runtime/pull/11

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -205,6 +205,13 @@ if [ -d /proc/cygdrive ]; then
     esac
 fi
 
+# @see https://github.com/git-for-windows/msys2-runtime/pull/11
+# @see https://github.com/msys2/MSYS2-packages/blob/f31cce2ca6e1ceba/bash/0006-bash-4.3-add-pwd-W-option.patch#L25
+if [[ "\$OSTYPE" == "msys" ]] && [[ ! -z \${MSYS_NO_PATHCONV+x} ]]; then
+    # Setting MSYS_NO_PATHCONV disables automatic unix to windows path translations by msys - translate them here.
+    dir="$(cd "\$dir" && pwd -W)";
+fi
+
 "\${dir}/$binFile" "\$@"
 
 PROXY;


Seldaek:
Thanks but please rebase/re-target this PR to the 1.10 branch so it can be included as a bugfix in the next patch release.

johnstevenson:
> Under windows git bash the following produces an unexpected result:

Far from being unexpected, this is actually expected NOT to work, so I'm not entirely sure this is Composer's responsibility.

https://github.com/git-for-windows/git/issues/407#issuecomment-141548288

As I understand it, using `MSYS_NO_PATHCONV` is to solve a specific problem and it should only be set temporarily.

Seldaek:
Yup I agree this does make it sound like patching this is rather unnecessary.

luka8088:
> Thanks but please rebase/re-target this PR to the 1.10 branch so it can be included as a bugfix in the next patch release.

Rebased

luka8088:
> 
> 
> > Under windows git bash the following produces an unexpected result:
> 
> Far from being unexpected, this is actually expected NOT to work, so I'm not entirely sure this is Composer's responsibility.
> 
> [git-for-windows/git#407 (comment)](https://github.com/git-for-windows/git/issues/407#issuecomment-141548288)
> 
> As I understand it, using `MSYS_NO_PATHCONV` is to solve a specific problem and it should only be set temporarily.

Interesting, I didn't know they had that attitude towards it. I have it set as a default all the time because as a practical matter addressing the issues that it causes is much less work than using it selectively.

I do understand if you prefer not to merge it though.

johnstevenson:
> I do understand if you prefer not to merge it though.

While its certainly not my call, it is interesting that Git for Windows doesn't even try to "fix" things for its own known `gitk` invocation (as per the thread I linked to).

There are other ways to get around the path conversion: http://www.mingw.org/wiki/Posix_path_conversion

Also from https://github.com/git-for-windows/build-extra/blob/master/ReleaseNotes.md 

> Alternatively, you can double the first slash to avoid POSIX-to-Windows path conversion, e.g. "//usr/bin/bash.exe".

Although it all looks very cumbersome!

Seldaek:
Yeah, overall this seems like it's not really worth a patch.

Besides, MSYS/cygwin are really best replaced by WSL IMO at this point, which offers a `wslpath` for cross-boundary scripting when absolutely needed.

luka8088:
> 
> 
> > I do understand if you prefer not to merge it though.
> 
> While its certainly not my call, it is interesting that Git for Windows doesn't even try to "fix" things for its own known `gitk` invocation (as per the thread I linked to).
> 
> There are other ways to get around the path conversion: http://www.mingw.org/wiki/Posix_path_conversion
> 
> Also from https://github.com/git-for-windows/build-extra/blob/master/ReleaseNotes.md
> 
> > Alternatively, you can double the first slash to avoid POSIX-to-Windows path conversion, e.g. "//usr/bin/bash.exe".
> 
> Although it all looks very cumbersome!

Yeah, I read already through that but the thing is I have lots of cross-platform bash scripts, thanks anyways though!

luka8088:
> 
> 
> Yeah, overall this seems like it's not really worth a patch.
> 
> Besides, MSYS/cygwin are really best replaced by WSL IMO at this point, which offers a `wslpath` for cross-boundary scripting when absolutely needed.

Yeah, maybe it's time that I look into migrating onto WSL, thanks!

PR #9064, opened 2020-07-21
Use CompilingMatcher in DefaultPolicy

Toflar:
While fiddling around I found that we're not using the new `CompilingMatcher` in the `DefaultPolicy`.
I think we should but the failing unit tests indicate that there might be an issue in `composer/semver`.

### DIFF
--- a/src/Composer/DependencyResolver/DefaultPolicy.php
+++ b/src/Composer/DependencyResolver/DefaultPolicy.php
@@ -15,6 +15,7 @@ namespace Composer\DependencyResolver;
 use Composer\Package\PackageInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\BasePackage;
+use Composer\Semver\CompilingMatcher;
 use Composer\Semver\Constraint\Constraint;
 
 /**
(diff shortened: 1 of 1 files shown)


src/Composer/DependencyResolver/DefaultPolicy.php
```diff
@@ -39,9 +40,8 @@ public function versionCompare(PackageInterface $a, PackageInterface $b, $operat
         }
 
         $constraint = new Constraint($operator, $b->getVersion());
-        $version = new Constraint('==', $a->getVersion());
 
-        return $constraint->matchSpecific($version, true);
```
Seldaek, line 44: This used matchSpecific with `true` and not match  (which calls matchSpecific with false). That means dev-xx versions will always return false instead of comparing/sorting correctly. AFAIK there is no way to do this with the compiling matcher. I assume that's the reason for the test failure.

So unless it really makes a perf difference to switch to the compiling matcher here, I'd say we just close the PR.

Toflar: I see. Nah it doesn't on current `master`. Closing.

PR #9065, opened 2020-07-21
Fix: Reference

localheinz:
This PR

* [x] fixes a reference from a deprecated method to a replacement method

Follows #8957.


localheinz:
Thank you, @Seldaek!

PR #9066, opened 2020-07-22
`composer require vendor/package` does not generate correct constraint for packages with four-digit version

rob006:
If last version of required package has 4 digits, Composer does not use `^` as prefix for constraint and uses version name instead. So for version `0.1.2.3` constraint is `0.1.2.3` instead of `^0.1.2.3`.

**Expected:**

```
$ composer require rob006/flarum-lang-polish
Using version ^0.3.12.5 for rob006/flarum-lang-polish
```

**Actual:**

```
$ composer require rob006/flarum-lang-polish
Using version v0.3.12.5 for rob006/flarum-lang-polish
```

### DIFF
--- a/src/Composer/Package/Version/VersionSelector.php
+++ b/src/Composer/Package/Version/VersionSelector.php
@@ -150,16 +150,18 @@ class VersionSelector
         $semanticVersionParts = explode('.', $version);
 
         // check to see if we have a semver-looking version
-        if (count($semanticVersionParts) == 4 && preg_match('{^0\D?}', $semanticVersionParts[3])) {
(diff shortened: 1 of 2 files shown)


Seldaek:
This is by design. If you have 4 version parts you are not following semver, and ^ is specifically for semver.. Bump your version to 1.0.0 and stop pretending you're in 0.x IMO if you want to follow semver.

rob006:
> This is by design. If you have 4 version parts you are not following semver, and ^ is specifically for semver.

That is a bit inconsistent. `1` or `1.2` are not semver either, but they will result constraint prefixed by `^`.

PR #9069, opened 2020-07-23
Add global bin directory to user path on Windows

johnstevenson:
On Windows this is `C:/Users/<user>/AppData/Roaming/Composer/vendor/bin`. Although the Windows installer (Composer-Setup.exe) sets this value for the current user, in an admin install the admin account might not be the end-user account (and other users will not have this value set). Obviously this value is not set for manual installs.

An additional message to the directory change is shown to the user when the value is set:
```
$ composer global config
Changed current directory to C:/Users/<user>/AppData/Roaming/Composer
Added "C:/Users/<user>/AppData/Roaming/Composer/vendor/bin" to your PATH. Open a new terminal to use it.
```

#### Questions:
1. Is this the correct place for such an action?
2. Should I check for the existence of the global bin-dir first (or is it guaranteed to exist)?

#### Hackery
In order to correctly escape the new path value for `setx`, I've used `Composer\XdebugHandler\Process`, which is marked as an `@internal` class (although it is very unlikely to change).

This is needed because `Composer\Util\ProcessExecutor::escapeArgument` is not robust enough to preserve path values containing `%....%` which is important when messing with the registry. I could fix it, but perhaps this should be in another PR.

### DIFF
--- a/src/Composer/Command/GlobalCommand.php
+++ b/src/Composer/Command/GlobalCommand.php
@@ -12,8 +12,12 @@
 
 namespace Composer\Command;
 
+use Composer\Config;
 use Composer\Factory;
+use Composer\IO\IOInterface;
 use Composer\Util\Filesystem;
+use Composer\Util\Platform;
+use Composer\XdebugHandler\Process;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\StringInput;
@@ -40,6 +44,8 @@ within the global context of COMPOSER_HOME.
 
 You can use this to install CLI utilities globally, all you need
 is to add the COMPOSER_HOME/vendor/bin dir to your PATH env var.
+
+This is already done for you on Windows, when a global command is run.
 
 COMPOSER_HOME is c:\Users\<user>\AppData\Roaming\Composer on Windows
 and /home/<user>/.composer on unix systems.
@@ -76,6 +82,7 @@ EOT
         }
 
         // change to global dir
+        $io = $this->getIO();
         $config = Factory::createConfig();
         $home = $config->get('home');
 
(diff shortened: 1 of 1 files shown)


stof:
I don't see why running `composer global config should magically alter the PATH on windows-only.

If the issue is the fact that the windows installer can install composer in a shared way but alters the path only for the admin user when running as admin, I suggest reporting an issue to https://github.com/composer/windows-setup so that this behavior changes (it is possible to define Path entries applying to all users on Windows)

johnstevenson:
> I don't see why running `composer global config should magically alter the PATH on windows-only.

I did say that I wasn't sure if this was the right place, if that is what you mean. Composer "magically" creates the local user folders on both platforms anyway, so why shouldn't it update the user path on Windows as well? What exactly is your objection?

> it is possible to define Path entries applying to all users on Windows

Which doesn't help those who have installed manually. However, I wasn't aware that this was the case, so please explain how this can be done.

stof:
> Which doesn't help those who have installed manually. However, I wasn't aware that this was the case, so please explain how this can be done.

Well, if you install it manually, you also need to update your Path manually. and that matches the behavior on other operating systems too.

johnstevenson:
So you don't like it because it isn't replicated across all OSs - fair enough.

> it is possible to define Path entries applying to all users on Windows

I presume you know how this can be reliably done. I don't, so it would be really helpful for Composer-Setup.exe if you could explain it to me - thanks.

Seldaek:
To be honest I wouldn't bother with this. People can do it manually if they care (this is the same currently on all platforms, it's not like the global dir is added to PATH on linux or OSX..). Long term we probably anyway want to go away from `composer global` and towards a require-tool or similar, as per https://github.com/composer/composer/issues/5390#issuecomment-463625428

johnstevenson:
@Seldaek Okay, no worries if that is the way things are going. I'll probably keep the Composer-Setup.exe functionality for the time being.

PR #9074, opened 2020-07-25
Replace Compiler::stripWhitespace with native implementation

Chi-teck:
This reduces the size of composer.phar file about 2% and makes compiler a bit more simple.

### DIFF
--- a/src/Composer/Compiler.php
+++ b/src/Composer/Compiler.php
@@ -191,10 +191,9 @@ class Compiler
     private function addFile($phar, $file, $strip = true)
     {
         $path = $this->getRelativeFilePath($file);
-        $content = file_get_contents($file);
-        if ($strip) {
-            $content = $this->stripWhitespace($content);
-        } elseif ('LICENSE' === basename($file)) {
+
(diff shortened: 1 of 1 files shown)


src/Composer/Compiler.php
```diff
@@ -215,40 +214,6 @@ private function addComposerBin($phar)
         $phar->addFromString('bin/composer', $content);
     }
 
-    /**
-     * Removes whitespace from a PHP source string while preserving line numbers.
```
Chi-teck, line 219: I do not know what was the reason to preserve line numbers. With this PR newlines will be removed.

Seldaek:
As you pointed out, this removes newlines, which is unacceptable to us as it means the bug reports we get with stack traces are impossible to match to real source. We'd need to store sourcemaps which would be even more complex than keeping the newlines.

src/Composer/Compiler.php
stof: the reason is that 100% of bug reports we receive are providing stack traces generated by the phar. Preserving line numbers makes them actionable.

Chi-teck: We could leave a note about that reason in the compiler.

PR #9085, opened 2020-07-29
Allow specifying a version requirement for CLDR

lstrojny:
libicu ships with a version of CLDR (http://cldr.unicode.org/), which is basically the data version. It would be great if one could specify the CLDR version specifically using composer.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -154,9 +154,11 @@ class PlatformRepository extends ArrayRepository
                     break;
 
                 case 'intl':
-                    # Add a seperate version for the CLDR library version
-                    $cldrVersion = \ResourceBundle::create('root', 'ICUDATA-curr', false)->get('Version');
-                    $this->addLibrary('cldr', 'The unicode CLDR project', $cldrVersion);
+                    if (class_exists('ResourceBundle', false)) {
+                        # Add a seperate version for the CLDR library version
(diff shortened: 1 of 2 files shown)


src/Composer/Repository/PlatformRepository.php
```diff
@@ -154,7 +154,12 @@ protected function initialize()
                     break;
 
                 case 'intl':
-                    $name = 'ICU';
+                    # Add a seperate version for the CLDR library version
+                    $cldrVersion = \ResourceBundle::create('root', 'ICUDATA-curr', false)->get('Version');
+                    $this->addLibrary('cldr', 'The unicode CLDR project', $cldrVersion);
+
+                    $name = 'icu';
+                    $description = 'The ICU unicode and globalization support library';
```
Seldaek, line 162: We need to support 5.3 and 5.4 which don't have the class present:

```suggestion
                    // Add a seperate version for the CLDR library version
                    if (class_exists('ResourceBundle')) {
                        $icuBundle = \ResourceBundle::create('root', 'ICUDATA-curr', false);
                        $this->addLibrary('cldr', 'The unicode CLDR project', $icuBundle->get('Version'));
                    }

                    $name = 'icu';
                    $description = 'The ICU unicode and globalization support library';
```

lstrojny: Good point, I've misread the docs on that. Pushed a fix to make that conditional.

Seldaek:
Thanks

PR #9093, opened 2020-07-30
Expand library version checking capabilities

lstrojny:
Extends composer’s library version checking capability by a bunch of other libraries that a) are bundled or at least somewhat common and b) I happen to have available.

### Structure
As two PHP extension can depend on different versions of the same library, the naming convention that was followed at the moment won’t work. Case in point: php-src depends on zlib but ext/curl which comes with php-src might depend on a different version of zlib.
The structure is now that PECL extension that depend on a library prefix the name of the library with the name of the extension. So `lib-curl-zlib` is the zlib version of curl and `lib-zlib` means the php-src zlib version.

### New versionable libraries
* `lib-amqp-librabbitmq`
* `lib-amqp-protocol`
* `lib-bz2`
* `lib-curl-libssh2`
* `lib-curl-openssl`
* `lib-curl-openssl-fips` (provides `lib-curl-openssl`)
* `lib-curl-gnutls` (replaces `lib-curl-openssl`)
* `lib-curl-nss` (replaces `lib-curl-openssl`)
* `lib-curl-zlib`
* `lib-date-timelib`
* `lib-date-zoneinfo`
* `lib-timezonedb-zoneinfo` (replaces `lib-date-zoneinfo`)
* `lib-fileinfo-libmagic`
* `lib-gd`
* `lib-gd-libjpeg`
* `lib-gd-libpng`
* `lib-gd-freetype`
* `lib-gd-libxpm`
* `lib-gmp`
* `lib-iconv`
* `lib-icu-cldr`
* `lib-icu-unicode`
* `lib-icu-zoneinfo`
* `lib-libxml` (provides `lib-simplexml-libxml`, `lib-dom-libxml`, `lib-xml-libxml`, `lib-xmlreader-libxml`, `lib-xmlwriter-libxml`)
* `lib-libsodium`
* `lib-ldap-openldap`
* `lib-mbstring-libmbfl`
* `lib-mbstring-oniguruma`
* `lib-memcached-libmemcached`
* `lib-mongodb-libmongoc`
* `lib-mongodb-libbson`
* `lib-openssl`
* `lib-openssl-fips` (provides `lib-openssl`)
* `lib-pcre-unicode`
* `lib-pgsql-libpq`
* `lib-pdo_pgsql-libpq`
* `lib-mysqlnd-mysqlnd`
* `lib-pdo_mysql-mysqlnd`
* `lib-sqlite3-sqlite`
* `lib-pdo_sqlite-sqlite`
* `lib-ssh2-libssh2`
* `lib-libxslt` (replaces `lib-xsl`)
* `lib-libxslt-libxml`
* `lib-yaml-libyaml`
* `lib-zlib`

### Changes
 * `lib-imagick` is now `lib-imagick-imagemagick` but `lib-imagick` stays for backwards compatibility. ~Would have loved to set `lib-imagick` as abandoned with the relevant successor package but this seems to not be reported anywhere. Do I miss something?~ This now uses `Package::setReplaces()` to offer both names.
 * `lib-ICU` is now `lib-icu` internally which won’t make a difference for users
 * `lib-libxslt` is the proper name for `lib-xsl`, which is kept for backwards compatibility
 * `lib-libzip` replaces `lib-zip`, which is kept for backwards compatibility
 * OpenSSL versions are now handled even for the `0.9.8zg` case which is converted to ~`0.9.8.31`~ (this had bugs, off by one of course, and is actually `0.9.8.33`)
 * `lib-uuid` was erroneously exposing the extension version not the library version. `lib-uuid` is now replaced by `ext-uuid`

### Other considerations
 * Testing: Implemented a Runtime object to make testing quite simple. The implementation is trivial and doesn’t require testing in itself but the contract is super valuable for testing and mocking
 * Performance should be good enough as a) code will only be executed for existing extensions and b) what is happening is relatively cheap


stof:
>     * Would have loved to set `lib-imagick` as abandoned with the relevant successor package but this seems to not be reported anywhere. Do I miss something?

you don't miss anything. Abandoned packages are reported when they are listed as installed packages. Here, the package is a platform package so it won't ever appear in those lists.
Thus, the reporting is based on installed packages, not on defined constraints, so this would not help you.

For such deprecation to work, I think that you would need to add support for deprecated platform package names in `composer validate` when validating constraints.

But in any case, they would have to be kept for a very long time due to the vast amount of existing package releases using these names, that would become impossible to install if the old name is removed.

src/Composer/Repository/PlatformRepository.php
```diff
@@ -147,109 +239,196 @@ protected function initialize()
...
                 case 'uuid':
-                    $prettyVersion = phpversion('uuid');
+                    $this->addLibrary($name, phpversion('uuid'));
                     break;
 
                 case 'xsl':
-                    $prettyVersion = LIBXSLT_DOTTED_VERSION;
+                    $this->addLibrary('libxslt', LIBXSLT_DOTTED_VERSION);
+                    // Alias for backwards compatibility
+                    $this->addLibrary('xsl', LIBXSLT_DOTTED_VERSION);
```
stof, line 409: should it be added as a separate package or as an alternate name for the same package using `provide` ?

lstrojny: Good idea, using replaces now

lstrojny:
Now the test coverage should be pretty good plus added a huge test suite for converting OpenSSL versions to something sensible.

Ready for review now

lstrojny:
I've set up this little data donation script to crowdsource more test data: https://usrportage.de/composer-data-donation/

src/Composer/Repository/PlatformRepository.php
```diff
@@ -337,6 +546,11 @@ private function addExtension($name, $prettyVersion)
         $packageName = $this->buildPackageName($name);
         $ext = new CompletePackage($packageName, $version, $prettyVersion);
         $ext->setDescription('The '.$name.' PHP extension'.$extraDescription);
+
+        if ($name === 'uuid') {
+            $ext->setReplaces(array(new Link('ext-uuid', 'lib-uuid', new Constraint('=', $version))));
```
stof, line 551: is `ext-uuid` actually using the libuuid version as its version number ? This looks wrong to me.

lstrojny: No, PECL uuid exposes it’s own version (as it should)
 * https://github.com/php/pecl-networking-uuid/blob/master/php_uuid.h#L39
 * https://github.com/php/pecl-networking-uuid/blob/master/uuid.c#L67

stof: then this change looks wrong to me. Why not adding a library package ?

lstrojny: Let’s see, we either misunderstand each other or what I intend to do is different from what I am doing. 

Previously composer was including `lib-uuid` with the version of the uuid extension. The intent is now to get rid of `lib-uuid` because it’s not really a thing. On the off chance that somebody might be requiring `lib-uuid`, `ext-uuid` provides now `lib-uuid` for backwards compatibility. Just to make sure, this particular change is in `addExtension()` while all other changes are around *libraries*.

lstrojny:
@Seldaek this should be good to go now

Seldaek:
Thanks for the huge amount of research here! Fixed a couple nitpicks and co, see linked commits above.

ondrejmirtes:
Seems like this PR introduced a bug: https://github.com/composer/composer/issues/9128

PR #9101, opened 2020-08-03
Fix the example link for the list endpoint

stof:
The autolinking from GFM does not include the `*` char in the link. Using an explicit link in the markdown fixes it.


Seldaek:
Ah thanks

PR #9105, opened 2020-08-06
Add support for TAR in Artifact packages

wissem:
Followup on https://github.com/composer/composer/issues/8931 to add support for tar artifacts.
I tried to reproduce the same behaviour for zip files and added the same test cases.


src/Composer/Repository/ArtifactRepository.php
```diff
@@ -89,8 +90,22 @@ private function scanDirectory($path)
...
     {
         $json = null;
+        $fileType = null;
+        $fileExtension = pathinfo($file->getPathname(), PATHINFO_EXTENSION);
+        if (in_array($fileExtension, array('gz', 'tar'), true)) {
+            $fileType = 'tar';
+        } else if ($fileExtension === 'zip') {
+            $fileType = 'zip';
+        } else {
+            throw new \RuntimeException('Files with "'.$fileExtension.'" extensions aren\'t supported. Only ZIP and TAR archives are supported.');
```
staabm, line 100:
```suggestion
            throw new \RuntimeException('Files with "'.$fileExtension.'" extensions aren\'t supported. Only ZIP and TAR/TAR.GZ archives are supported.');
```

src/Composer/Repository/ArtifactRepository.php
```diff
@@ -66,7 +67,7 @@ private function scanDirectory($path)
 
         $directory = new \RecursiveDirectoryIterator($path, \RecursiveDirectoryIterator::FOLLOW_SYMLINKS);
         $iterator = new \RecursiveIteratorIterator($directory);
-        $regex = new \RegexIterator($iterator, '/^.+\.(zip|phar)$/i');
+        $regex = new \RegexIterator($iterator, '/^.+\.(zip|phar|tar|gz)$/i');
```
Seldaek, line 70: Maybe we should also support `tgz`?

src/Composer/Util/Tar.php
```diff
@@ -0,0 +1,84 @@
...
+     * @param \PharData PharData
+     * @param \PharData $folder
+     * @param int      $searchLevels
+     * @param string      $path
+     *
+     * @throws \RuntimeException
+     *
+     * @return string
+     */
+    private static function extractComposerJsonFromFolder(\PharData $phar, \PharData $folder, $searchLevels, $path = '')
```
Seldaek, line 46: It seems like $phar is entirely unused, so probably should remove it.

src/Composer/Util/Tar.php
```diff
@@ -0,0 +1,84 @@
...
+
+        $composerJsons = array();
+        foreach ($directories as $dir) {
+            $pathName = $dir->getPathname();
+            $composerJsons[] = self::extractComposerJsonFromFolder($phar, new \PharData($pathName), $searchLevels - 1, $path . $dir->getBasename() . '/');
+        }
+
+        if (count($composerJsons) > 1) {
+            throw new \RuntimeException('Multiple composer json were found in the archive file. Make sure to have a single composer.json at the root directory.');
+        }
```
Seldaek, line 76: IMO this should be simplified a lot. The recursion is not needed and adds unnecessary complexity, and a few things like line 52/53 also don't seem correct or needed.

I think the approach for Util\Zip to check for the root composer.json first, and if not look for top level folders, and if you only have one check for folder/composer.json, would be more than enough and shorter/clearer.

You can use `isset($phar['composer.json'])` to check for the file presence, then collect $directories, and if you only have one check `isset($phar[$directory.'/composer.json'])` then.

tests/Composer/Test/Util/TarTest.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        Tar::getComposerJson(__DIR__.'/Fixtures/Tar/multiple.tar.gz');
+    }
+
+    public function testReturnsComposerJsonFromFirstSubfolder()
+    {
+        $result = Tar::getComposerJson(__DIR__.'/Fixtures/Tar/single-sub.tar.gz');
+
+        $this->assertEquals("{\n    \"name\": \"foo/bar\"\n}\n", $result);
+    }
```
Seldaek, line 80: Is this the same test as `testReturnsComposerJsonInFirstFolder` ? It sounds like it but I can't see what's in the archives here so not 100% sure.

tests/Composer/Test/Util/TarTest.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @expectedException \RuntimeException
+     */
+    public function testThrowsExceptionIfMultipleComposerInSubFoldersWereFound()
+    {
+        Tar::getComposerJson(__DIR__.'/Fixtures/Tar/multiple_subfolders.tar.gz');
+    }
```
Seldaek, line 88: This also looks like a duplicate of testMultipleTopLevelDirsIsInvalid

tests/Composer/Test/Util/TarTest.php
wissem: those are indeed similar (just different folder names), same for the test below. I've removed them.

src/Composer/Util/Tar.php
wissem: ok, I've tried to use similar approach to Util\Zip, I think that approach is indeed shorter and clearer.

src/Composer/Repository/ArtifactRepository.php
```diff
@@ -89,8 +90,22 @@ private function scanDirectory($path)
...
     {
         $json = null;
+        $fileType = null;
+        $fileExtension = pathinfo($file->getPathname(), PATHINFO_EXTENSION);
+        if (in_array($fileExtension, array('gz', 'tar', 'tgz'), true)) {
+            $fileType = 'tar';
+        } else if ($fileExtension === 'zip') {
+            $fileType = 'zip';
+        } else {
+            throw new \RuntimeException('Files with "'.$fileExtension.'" extensions aren\'t supported. Only ZIP and TAR/TAR.GZ archives are supported.');
```
Seldaek, line 100: I guess this should say TAR/TAR.GZ/TGZ now 😅

src/Composer/Util/Tar.php
```diff
@@ -0,0 +1,70 @@
...
+        foreach ($phar as $folderFile) {
+            $name = $folderFile->getBasename();
+
+            if ($folderFile->isDir()) {
+                $topLevelPaths[$name] = true;
+                if (\count($topLevelPaths) > 1) {
+                    throw new \RuntimeException('Archive has more than one top level directories, and no composer.json was found on the top level, so it\'s an invalid archive. Top level paths found were: '.implode(',', array_keys($topLevelPaths)));
+                }
+
+                continue;
```
Seldaek, line 59: This is unnecessary as there is nothing following in the loop anyway, but otherwise yeah IMO much easier to understand what is going on here :) 👍

wissem: indeed :)

Seldaek:
Thanks!

PR #9115, opened 2020-08-10
Docs: Move note about not committing lock file to correct section.

PrinsFrank:
Move note about when it is not necessary to commit the lockfile from the "Updating dependencies to their latest version" section to the "Commit your composer.lock file to version control" section as that placements makes better sense.


Seldaek:
Thanks

PR #9116, opened 2020-08-10
Update PathDownloader.php

ryanaslett:
If a path repository points at a directory that is managed by composer installers, the path that gets set ends up being relative, and this check fails to see that the source is already present, and therefore removes it.

Since ->install is already using realpath around the $path argument, remove should as well.

For an example repository that demonstrates this bug See: https://github.com/ryanaslett/pathrepotestcase


Seldaek:
Thanks

Issue #9109, opened 2020-08-08
Detect Github Action runs and format errors/warnings in Github Actions format


staabm:
When composer is invoked from within a Github Action It should output error messages in the Github Action native error format.

See https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message

This would make error messages appear in the Pull Request itself and people would not need to dig into the Github Actions logs to find out what happend/what the error is

The Github Action run can be detected by presence of a Env var, see https://github.com/OndraM/ci-detector/blob/main/src/Ci/GitHubActions.php

Seldaek:
Should be doable by hooking into Application::doRun, at the end of it we handle exceptions, could output them as errors and then proceed to rethrow to get the usual output as well.

staabm:
I thought about subclassing ConsoleIO and emitting github formatted markup based on the message-formatting tags used in the error messages

Seldaek:
That sounds way overkill IMO, and possibly wrong too, as we sometimes emit warnings and possibly error level stuff which should not really trigger errors on github. Uncaught exceptions tho are definitely a problem we can report. Might just need a special case for solving errors I can't remember if those are output somehow or just thrown.

PR #9120, opened 2020-08-12
emit github action formatted error messages

clxmstaab:
need some more testing on my end..

closes https://github.com/composer/composer/issues/9109


src/Composer/Console/Application.php
```diff
@@ -191,6 +192,10 @@ public function doRun(InputInterface $input, OutputInterface $output)
                 }
             } catch (NoSslException $e) {
                 // suppress these as they are not relevant at this point
+            } catch (ParsingException $e) {
```
clxmstaab, line 195: added this case, to catch errors while `composer validate` and emit them github formatted

src/Composer/Console/Application.php
```diff
@@ -307,12 +312,21 @@ function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknow
...
             restore_error_handler();
             throw $e;
         }
     }
 
+    private function emitGithubActionError(\Exception $e) {
+        if (!empty($_ENV['GITHUB_ACTIONS'])) {
+            // newlines need to be encoded
+            // see https://github.com/actions/starter-workflows/issues/68#issuecomment-581479448
+            echo "::error ::". str_replace("\n", '%0A', $e->getMessage());
```
clxmstaab, line 326: would the stacktrace be helpful?

clxmstaab:
just did a few tests and it seem to work like expected.

see https://github.com/clxmstaab/composer/actions/runs/206753940

solver errors are not yet converted to github action errors though, since I was not able to find a good spot where to put the catch/formatting lines for the Installer.

Seldaek:
I think for Installer, every catch block catching a SolverProblemsException maybe should call `$this->outputSolverError($message, $details[, $message2])`, with message/message2 being output with QUIET, and going to github output as well, and $details I am not sure if it should go to github or not, as it can get quite long.

src/Composer/Console/Application.php
```diff
@@ -191,6 +192,19 @@ public function doRun(InputInterface $input, OutputInterface $output)
                 }
             } catch (NoSslException $e) {
                 // suppress these as they are not relevant at this point
+            } catch (ParsingException $e) {
+                $details = $e->getDetails();
+
+                $file = realpath(Factory::getComposerFile());
```
stof, line 198: what if a different json file was parsed ?

clxmstaab: I had the feeling that the above call to `getPluginCommands()` also uses `Factory::getComposerFile()` so this should be in sync.

clxmstaab:
> I think for Installer, every catch block catching a SolverProblemsException maybe should call `$this->outputSolverError($message, $details[, $message2])`, with message/message2 being output with QUIET, and going to github output as well, and $details I am not sure if it should go to github or not, as it can get quite long.

I tried your suggestion. having a generic method with abstract variables like $message, $message2, $details did not read well for me.

therefore I just added the necessary lines inside the try/catch blocks directly, which felt more readable for me.

here you can see a example build, showing a solver error:
https://github.com/clxmstaab/composer/actions/runs/206974603

src/Composer/Console/Application.php
stof: ah, I missed this was a catch around the loading of plugin commands and not against the command execution. So this is indeed less likely to parse more JSON files.

src/Composer/Console/GithubActionError.php
```diff
@@ -0,0 +1,26 @@
...
+    public static function emit($message, $file = null, $line = null) {
+        if (getenv('GITHUB_ACTIONS')) {
+            // newlines need to be encoded
+            // see https://github.com/actions/starter-workflows/issues/68#issuecomment-581479448
+            $message = str_replace("\n", '%0A', $message);
+
+            if ($file && $line) {
+                echo "::error file=". $file .",line=". $line ."::". $message;
+            } elseif ($file) {
+                echo "::error file=". $file ."::". $message;
```
stof, line 20: should we actually use echo or should we use the IO abstraction to write to the output ? This would help cases wanted to trap the output (tests for instance...)

Seldaek: Yeah I'd rather use $io->write, should be equivalent as far as I can tell, but it is indeed better for testing.

clxmstaab: adjusted per feedback

clxmstaab: with the change to IO a lot of tests now fail because of different output/expectations.

could you suggest a way to solve this?

stof: well, force the integration tests to run in non-action mode, even when the runner uses github actions

clxmstaab: this seems to work https://github.com/composer/composer/pull/9120/commits/c538b6e7eed22d1d149586ecb760e4d90642f53e

clxmstaab:
test with the current PR contents still error like expected:

https://github.com/clxmstaab/composer/actions/runs/207201487

Seldaek:
Thanks

PR #9124, opened 2020-08-14
Fix openssl_free_key deprecation notice in PHP 8

johnstevenson:
PHP 8 deprecates `openssl_free_key` (actually `openssl_pkey_free` which it aliases) and automatically destroys the key instance when it goes out of scope.

As per https://github.com/composer/windows-setup/issues/122

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -271,7 +271,12 @@ TAGSPUBKEY
             $signature = json_decode($signature, true);
             $signature = base64_decode($signature['sha384']);
             $verified = 1 === openssl_verify(file_get_contents($tempFilename), $signature, $pubkeyid, $algo);
-            openssl_free_key($pubkeyid);
+
+            // PHP 8 automatically frees the key instance and deprecates the function
+            if (PHP_VERSION_ID < 80000) {
+                openssl_free_key($pubkeyid);
+            }
+
             if (!$verified) {
                 throw new \RuntimeException('The phar signature did not match the file you downloaded, this means your public keys are outdated or that the phar file is corrupt/has been modified');
             }


src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -271,7 +271,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $signature = json_decode($signature, true);
             $signature = base64_decode($signature['sha384']);
             $verified = 1 === openssl_verify(file_get_contents($tempFilename), $signature, $pubkeyid, $algo);
-            openssl_free_key($pubkeyid);
+
+            // PHP 8 automatically frees the key instance and deprecates the function
+            if (PHP_VERSION_ID < 80000) {
```
stof, line 276: I suggest using `\PHP_VERSION_ID` which resolves the version check at compile time.

johnstevenson: Thanks, but I am simply following the predominant style in the repo. If this is how we are _meant_ to do it from now on, then I will obviously comply.

GrahamCampbell: 👍 for `\PHP_VERSION_ID` too. Not all global constants are treated equally. Some can be optimised by opcache, which includes constant propagation and removing (some) dead paths.

stof: technically, all internal constants are resolved at compile-time when they are fully-qualified. But then, the benefits on removing dead paths is more on how these constants are used.
the `@Symfony:risky` ruleset of php-cs-fixer turns prefixing on for 3 constants only, which are commonly used in conditions: `PHP_VERSION_ID`, `PHP_SAPI` and `DIRECTORY_SEPARATOR`

PR #9130, opened 2020-08-17
Downloader: add a max_file_size option to prevent too big files to be downloaded

glaubinix:
This pull request adds an option `max_file_size` which can be passed either to the RemoteFileSystem or the CurlDownloader which will prevent files larger than the limit to be downloaded and throw an exception in case this happens.

Being able to set such a limit is useful for applications that use Composer programmatically to prevent memory limit fatal error.


GrahamCampbell:
When do you see a fatal error. Composer shouldn't ever have the whole file in memory at once?

glaubinix:
The `RemoteFileSystem` does load the entire file into memory via `file_get_contents` https://github.com/composer/composer/blob/master/src/Composer/Util/RemoteFilesystem.php#L547. Same goes for the HttpDownloader with the CurlDownloader if you call `get` or `add` on the HttpDownloader which sets `copyTo` to false which then sets `filename` to false in the CurlDownloader resulting in `stream_get_contents($job['bodyHandle'])` to be called which loads the stream as string as well https://github.com/composer/composer/blob/master/src/Composer/Util/Http/CurlDownloader.php#L293-L295

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -532,23 +538,34 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
...
-     * @param resource $context   The stream context
+     * @param string   $originUrl   The origin URL
+     * @param string   $fileUrl     The file URL
+     * @param resource $context     The stream context
+     * @param int      $maxFileSize The maximum allowed file size
      *
      * @return string|false The response contents or false on failure
      */
-    protected function getRemoteContents($originUrl, $fileUrl, $context, array &$responseHeaders = null)
+    protected function getRemoteContents($originUrl, $fileUrl, $context, array &$responseHeaders = null, $maxFileSize = null)
```
nicolas-grekas, line 548: this change is a BC break affecting symfony/flex
@Seldaek would you accept a PR replacing this explicit argument by a call to `func_get_arg()`?

nicolas-grekas: mmm, this is on master, which means v2, this BC breaks allowed.
I'm going to patch flex instead.

Seldaek: Yeah sorry, didn't realize that.. Either or is fine by me tbh. It's a minor feature but also people really should have curl installed ;)

nicolas-grekas: No worries, we'll fix this in https://github.com/symfony/flex/pull/672

Issue #9128, opened 2020-08-17
Fatal Error: "inet_pton" not found on PHP 8.0 + Windows


ondrejmirtes:
I'm getting this fatal error when running `composer update` on PHP 8 nightly + Windows:

```
2020-08-17T06:06:07.1198057Z Loading composer repositories with package information
2020-08-17T06:06:07.1217380Z PHP Fatal error:  Uncaught TypeError: call_user_func_array(): Argument #1 ($function) must be a valid callback, class "inet_pton" not found in phar://C:/tools/php/composer/src/Composer/Platform/Runtime.php:44
2020-08-17T06:06:07.1219569Z Stack trace:
2020-08-17T06:06:07.1220021Z #0 phar://C:/tools/php/composer/src/Composer/Platform/Runtime.php(44): call_user_func_array()
2020-08-17T06:06:07.1241154Z 
2020-08-17T06:06:07.1241611Z #1 [internal function]: Composer\Platform\Runtime->invoke()
2020-08-17T06:06:07.1242402Z Fatal error: Uncaught TypeError: call_user_func_array(): Argument #1 ($function) must be a valid callback, class "inet_pton" not found in phar://C:/tools/php/composer/src/Composer/Platform/Runtime.php:44
2020-08-17T06:06:07.1244029Z #2 phar://C:/tools/php/composer/src/Composer/Util/Silencer.php(68): call_user_func_array()
2020-08-17T06:06:07.1244751Z Stack trace:
2020-08-17T06:06:07.1245437Z #3 phar://C:/tools/php/composer/src/Composer/Repository/PlatformRepository.php(128): Composer\Util\Silencer::call()
2020-08-17T06:06:07.1246750Z #0 phar://C:/tools/php/composer/src/Composer/Platform/Runtime.php(44): call_user_func_array()
2020-08-17T06:06:07.1247160Z #4 phar://C:/tools/php/composer/src/Composer/Repository/ArrayRepository.php(265): Composer\Repository\PlatformRepository->initialize()
2020-08-17T06:06:07.1247473Z #1 [internal function]: Composer\Platform\Runtime->invoke()
2020-08-17T06:06:07.1247756Z #5 phar://C:/tools/php/composer/src/Composer/Installer.php(824): Composer\Repository\ArrayRepository->getPackages()
2020-08-17T06:06:07.1248111Z #2 phar://C:/tools/php/composer/src/Composer/Util/Silencer.php(68): call_user_func_array()
[...]

PR #9133, opened 2020-08-18
Fix regression when inet_pton() does not exist

lstrojny:
varargs double indirection was incorrect, this fixes it + ads a regression test.

Adresses #9128


Seldaek:
Thanks

ondrejmirtes:
Thank you!

PR #9146, opened 2020-08-24
RemoteFilesystem: avoid warning when setting max file size

glaubinix:
Follow up for https://github.com/composer/composer/pull/9130 prevents PHP warning `stream_context_create(): options should have the form ["wrappername"]["optionname"] = $value`


Seldaek:
Thanks

PR #9148, opened 2020-08-24
Put the github token authentication back

matthiaz:
This was originally created here: https://github.com/composer/composer/pull/8747
But for some reason, this got removed in a merge 8b934a415f9b3787efe9a0201defaa011f8b68f9 to 1.10, not sure if this was on purpose.
re-adding, because this is breaking again.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -230,6 +230,14 @@ class RemoteFilesystem
 
         $origFileUrl = $fileUrl;
 
+        if (isset($options['github-token'])) {
+            // only add the access_token if it is actually a github URL (in case we were redirected to S3)
+            if (preg_match('{^https://api\.github\.com/}', $fileUrl)) {
+                $options['http']['header'][] = 'Authorization: token '.$options['github-token'];
+            }
+            unset($options['github-token']);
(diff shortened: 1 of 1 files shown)


staabm:
Any chance to get this covered by a unit test, so it won‘t get accidentally removed again?

Seldaek:
In master/2.0 it is done in AuthHelper, that's why it's not in RemoteFilesystem anymore. RemoteFilesystem is not meant to be used directly anymore, you should use HttpDownloader which abstracts over CurlDownloader or RemoteFilesystem/streams. See https://github.com/composer/composer/blob/master/src/Composer/Util/AuthHelper.php#L202-L207

The github-token option is also gone FWIW.

I am not sure how you got to this problem, but probably it's mis-use in a plugin or using composer as a library.

matthiaz:
Thanks for pointing that out @Seldaek I missed that. 

github-token is still very much alive https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token

I'll dig deeper into this issue. But for now, this ticket can indeed be closed.

Seldaek:
I mean passing the `github-token` to the composer downloader is gone. You should configure the auth for the right domain on the IO instance.

PR #9154, opened 2020-08-26
Util/Zip: fix strpos args order

quasilyte:
`strpos()` first argument is a haystack, not a needle.

`strpos('x', $s)` is identical to `$s === 'x'` which is probably not what we want here.

### DIFF
--- a/src/Composer/Util/Zip.php
+++ b/src/Composer/Util/Zip.php
@@ -92,7 +92,7 @@ class Zip
             }
 
             // handle archives which do not have a TOC record for the directory itself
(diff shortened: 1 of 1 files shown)


quasilyte:
I'm not sure I understand the CI failure.

staabm:
inspired by your change I searched the whole composer codebase.
it seems the only offending line. 

I found the same problem in the json-schema lib though, and opened a new https://github.com/justinrainbow/json-schema/pull/641

quasilyte:
And your pr inspired me to check against incorrect return value checks.

Seldaek:
Thanks :)

PR #9156, opened 2020-08-27
AuthHelper: Allow fall-through GitLab-specific HTTP headers for auth

Ayesh:
Previously, `AuthHelper` consumed the authentication credentials for GitLab domains and added access tokens as GitLab-specific headers.
[Composer repositories now supported in GitLab](https://php.watch/articles/composer-gitlab-repositories) require standard Authorization headers with a personal access to function, which failed to work due to out GitLab-specific headers.

With this commit, AuthHelper checks if the password is an access token, and falls through to HTTP basic authentication even if the domain name is a GitLab domain name.


Seldaek:
Thanks

Ayesh:
Thank you 🙏

PR #9158, opened 2020-08-27
Fix symlink check

infabo:
Given you have a symlink in one of your path repositories and mirroring-strategy enabled.

```bash
lrwxrwxrwx 1 inf inf    8 27. Aug 17:41 Create.html -> New.html
-rw-r--r-- 1 inf inf  750 27. Aug 17:24 Edit.html
-rw-r--r-- 1 inf inf    0 27. Aug 17:24 List.html
-rw-r--r-- 1 inf inf 5064 27. Aug 17:24 New.html
-rw-r--r-- 1 inf inf  716 27. Aug 17:24 Show.html
```

`$file->getLinkTarget()` just returns a relative path in this example. It does not return an absolute normalized path.
`$sources` is always a normalized path.

`strpos` can never be `0`.

Use `$file->getRealPath()` to fix the strpos-check.

### DIFF
--- a/src/Composer/Package/Archiver/ArchivableFilesFinder.php
+++ b/src/Composer/Package/Archiver/ArchivableFilesFinder.php
@@ -43,7 +43,7 @@ class ArchivableFilesFinder extends \FilterIterator
     {
         $fs = new Filesystem();
 
-        $sources = $fs->normalizePath($sources);
+        $sources = $fs->normalizePath(realpath($sources));
 
         if ($ignoreFilters) {
             $filters = array();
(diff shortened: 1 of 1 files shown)


naderman:
Currently this works for symlinks to absolute paths and is broken for relative targets, as you say.

Technically I suppose we'd have to run realpath on the `$sources` variable too, as that too may contain a link in which case this wouldn't really help by itself?

That said overall this definitely makes sense.

infabo:
> Technically I suppose we'd have to run realpath on the $sources variable too, as that too may contain a link in which case this wouldn't really help by itself?

Agreed.

naderman:
@infabo Do you want to take care of that in this PR as well then?

naderman (approved):

Seldaek:
Thanks @infabo

PR #9161, opened 2020-08-28
Update config section to note required scope for GitLab tokens

bradjones1:
I intuitively thought the requirement was `read_repository`, but this needs API access. Document as such.


Seldaek:
Thanks

PR #9165, opened 2020-08-29
Performance improvements

bpesch:
Hey, this is some stuff I came across while debugging something.

- Removed unnecessary method calls
- Sorted conditions by cost
- Improved code readability

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -474,7 +474,7 @@ EOT
             ),
         );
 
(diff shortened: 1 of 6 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -605,7 +605,7 @@ function ($vals) {
                 $value = strtolower($values[0]);
                 if (true === $booleanValidator($value)) {
                     if (false === $booleanNormalizer($value)) {
-                        $this->configSource->addRepository($matches[1], false);
+                        $this->configSource->addRepository($matches[1], []);
```
Ayesh, line 608: This causes PHP 5.3 to fail because short array syntax is PHP 5.4+.

Seldaek:
Thanks

Issue #8287, opened 2019-08-20
[Idea] Reducing versions entering the SAT solver by identifying useless ones


stof:
This is an idea to investigate for Composer 2. It is somewhat related to https://github.com/composer/composer/issues/7643 (it might solve the same cases, but I haven't checked whether they are strictly equivalent).

The idea here is to reduce the impact of the growing number of releases per package over time, thanks to the fact that most of these releases don't change the requirements (especially for patch releases).
This would run as an optimization pass at the end of building the pool (after having collected all necessary packages), before entering the solver. This optimization would work like that:

- for each package name, collect all sets of constraints coming from other packages (we need to take both `require` and `conflict` of each package into account to build this set, and also `require-dev` in case of the root package. The impact of replacement rules is still to determined here)
- for each package name, reduce the available versions: if 2 versions have the same constraints (considering both `require` and `conflict`) and they match the same incoming constraints from the sets collected above, keep only the preferred version among them ("preferred" as defined by the PolicyInterface).

Note that this might not drop some of the older versions, because they would be below some of the min requirements of some constraints but not of others, so the impact of this optimization should be checked more in detail (it might help more for patch versions than for minor versions).
To help for that, a more powerful check could be added: for 2 versions, if the preferred versions among them matches a superset of incoming requirements and a subset of incoming conflicts and its requirements are a subset of requirements and its conflicts are a superset of conflicts, keep only the preferred version (i.e. keep the preferred version if we are sure that it will always be selectable when the other one is)

mxr576:
Related #7051

stof:
Note that there is another condition to keep a package: if the version comes from a repository at lower priority than another version, it cannot hide it (but it may still be kept itself if its own version is higher so the opposite hiding not being satisfied either), due to the fact that repo priorities are taken into account when selecting the version.

Toflar:
I've actually had this idea a while ago too but never investigated it more. I now did and the issue here is the following.
Imagine `package/name` in versions

* `1.0.0`
* `1.0.1`
* `1.0.2`
* `1.1.0`
* `2.0.0`

It's perfectly possible for them to have exactly the same definitions of `require`, `provide`, `conflict` and `replace` (which is what makes up the same definition imho).

This then is a problem because the optimization step would remove all of them except `2.0.0` (when not `--prefer-lowest`) but that's an issue because other dependencies may require `package/name` in e.g. `^1.1`.

Toflar:
[...]

PR #9171, opened 2020-09-01
[POC] Poolbuilder optimizer

Toflar:
Now that https://github.com/composer/composer/pull/8850 got merged, let's try to take the pool builder to yet another level and optimizing things even more :) 

To better understand what the idea in this PR is, we should maybe go back to see what Composer 1.10 (or pre #8850) did and what it does now.

For this, it's maybe easiest to use and example and an illustration of our dependencies.
So let's assume we have a root requirement for `packageA` in `^2.0` which then requires `packageC` and `packageB` whereas `packageB` requires `packageC` as well but with a different constraint:

dep

As you can see, dependencies are of course circular because packages can reference the same packages already found but it kind of helps to visualize them as a "dependency tree". You start on the very top and at some point you're done.

Here's what Composer 1.10 (or pre #8850) loaded into the pool:

* All versions of `packageA` fulfilling the constraint `^2.0`
* All versions of `packageB`
* All versions of `packageC`

As you can see, there was no optimization in terms of which package versions are even referenced anywhere in the dependency tree. Except for the root requirement, just **all versions of all packages** ever published were loaded into the pool.

Here's what Composer loads now (post #8850):

* All versions of `packageA` fulfilling the constraint `^2.0`
* All versions of `packageB` fulfilling the constraint `^1.1`
* All versions of `packageC` fulfilling the constraint `^2.0 || ^3.0 || ^3.0` (which gets simplified to `^2.0 || ^3.0`)

We've now reduced the amount of package versions loaded to **any package version** referenced **anywhere** in the dependency tree. So we don't blindly load all package versions ever published anymore but we still load quite a lot of them.

Here's what this PR tries to do:

* All versions of `packageA` fulfilling the constraint `^2.0`
* All versions of `packageB` fulfilling the constraint `^1.1`
* All versions of `packageC` fulfilling the constraint `^3.0`

How?

If you look at how a dependency tree is structured, you'll notice a rather simple rule:
The higher up in the dependency tree a package was required, the more likely it was controlled/requested by the user.
The highest level (level 0) being your root composer.json which is of course 100% controlled by the user and
contains 100% of what the user requests. Then, on level 1, we have all the direct dependencies of the root
dependencies, making those ones **extremely likely** to be requested by the user as well.
In other words, the higher up the dependency tree the requirer of a package was found, the higher the probability
that this constraint can be used as a hard requirement and does not need to be expanded on.

If you look at our example dependency tree, you can see that `packageC` was first found on level 1 (`packageA` is a root requirement = level 0). It was then required again but this time it was on level 2.

So why don't we do this during the loading process already?

I thought about it but it doesn't work. If we don't load all the packages referenced anywhere in the dependency tree, we might miss out on `replace` statements thus ending in an incorrect result.
We have to load all of the versions referenced anywhere in the tree first. Only then we can filter after we've properly handeld the replace statements.

This is a POC illustrating the concept. I didn't work on any tests or anything else but I'd love to read the feedback. Maybe someone wants to run some tests etc.

I've ran it on one of my bigger projects and it was able to filter out `25%` of the ~16,000 package versions found.
Current master: `[104.8MiB/31.22s] Memory usage: 104.76MiB (peak: 1130.78MiB), time: 31.22s`
This PR: `[87.8MiB/21.57s] Memory usage: 87.76MiB (peak: 647.36MiB), time: 21.57s`

Note that even if the optimizer only filters out 5%, it's still going to be beneficial as every additional package version found increases the number of rules to create exponentially :)

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -606,6 +606,7 @@ class PoolBuilder
         $total = \count($this->packages);
 
         $this->unlearnLoadPackageOperationTracesFromReplacedPackages();
+        $this->evenLoadPackageOperationTraces();
         $this->applyOptimizations();
 
         $filtered = $total - \count($this->packages);
@@ -721,6 +722,46 @@ class PoolBuilder
                 foreach ($operations as $i => $operation) {
                     if ($operation->getSource() === $replacedPackage->getName() && $operation->getSourceVersion() === $replacedPackage->getVersion()) {
                         unset($this->loadPackageOperationsTrace[$targetPackage][$i]);
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Let's say packageA in version 1.0.0 required packageB in ^2.0
+     * and in later versions that dependency was removed. We are not
+     * allowed to use ^2.0 as a hard constraint then so we need
+     * to "even out" these ones by allowing any version of that
+     * package found on this level.
+     */
+    private function evenLoadPackageOperationTraces()
+    {
+        $requiredPackageNamesOverAllSourceVersions = array();
+        $requiredPackageNamesPerSourceVersion = array();
+
+        // Collect all packages required in any version and per version
+        foreach ($this->loadPackageOperationsTrace as $targetPackage => $operations) {
+            foreach ($operations as $operation) {
+                $requiredPackageNamesOverAllSourceVersions[$operation->getLevelFoundOn()][$operation->getSource()][$operation->getTarget()] = true;
+                $requiredPackageNamesPerSourceVersion[$operation->getLevelFoundOn()][$operation->getSource()][$operation->getSourceVersion()][] = $operation->getTarget();
+            }
+        }
+
+        // Even out
+        foreach ($requiredPackageNamesPerSourceVersion as $levelFoundOn => $sources) {
+            foreach ($sources as $source => $info) {
+                foreach ($info as $sourceVersion => $targetPackages) {
+                    if ($packagesToEvenOut = array_diff(array_keys($requiredPackageNamesOverAllSourceVersions[$levelFoundOn][$source]), $targetPackages)) {
+                        foreach ($packagesToEvenOut as $packageName) {
+                            $this->traceLoadPackageOperation($packageName, new LoadPackageOperation(
+                                $source,
+                                $sourceVersion,
+                                $packageName,
+                                new MatchAllConstraint(),
+                                $levelFoundOn
+                            ));
+                        }
                     }
                 }
             }
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -216,6 +216,8 @@ class PoolBuilder
             }
         }
 
+        $this->runOptimizer();
+
         if ($this->eventDispatcher) {
             $prePoolCreateEvent = new PrePoolCreateEvent(
                 PluginEvents::PRE_POOL_CREATE,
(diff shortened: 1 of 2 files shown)


Toflar:
Two more things:

1. Packages referenced on the same tree level in different constraints need to be requested with a disjunctive constraint. Unfortunately this will result in more packages loaded into the pool but if we didn't do this, we'd actually already start dependency resolving which we cannot do at this stage :) 

2. This PR will change the way you write your `composer.json`. Right now there's no difference whether you require `^2.1` or `^2.2` of a package if you're personally okay with both of them. However, the narrower you specify the dependencies now, the better the optimizer can work. If `2.1.0` of the package allowed for e.g. `symfony/filesystem` `^3.4 || ^4.4` and `2.2.0` changed this to `^4.4` only, maybe you can get rid of everything not matching `^4.4`. Depending on the other requirements of course :)

simonberger:
Some results I got are really promising !

I found a problem though and I hope it is a not very fundamental one.

First the minimal setup to reproduce it
```
"require": {
    "symfony/security": "2.7.51|2.8.52"
},
"require-dev": {
    "roave/security-advisories": "dev-master"
}
```

The problem is in short that lower package versions which aren't picked at the end currently still can further constrain package versions with their own requirements.

**What is happening in the example?**
2.7.51 requires paragonie/random_compat in version `~1.0` while 2.8.52 (not directly) allows `~1.0|~2.0|~9.99` which would be correct to choose from.
`roave/security-advisories` is good to let it break at the end. But it can just be left out. The problem is that 9.99.99 of paragonie/random_compat is not picked.

Toflar:
Ah nice find! It's not a fundamental one and I was able to fix it in 0ff6579be6fda897ba451fc6ae212f1a5bf99dba, I guess. However, it obviously reduces possible optimizations again. But your example still resulted in a 12% reduction of package versions :) 

> Some results I got are really promising !

Can you elaborate on that maybe? :)

simonberger:
That change fixed it. Really correct timings are currently not possible. I would need to get rid of networking. It is not or not much faster. But it dramatically reduces memory usage which is very good.

Master
```
Memory usage: 209.09MiB (peak: 1628.97MiB), time: 119.25s
```

This PR
```
Pool builder found a total of 17,483 package versions referenced in your dependency tree. The optimizer was able to filter out 5,352 (31%) of them early!
Memory usage: 177.2MiB (peak: 934.52MiB), time: 111.31s
```

Both are the fastest out of 3 runs. Time of runs varied by ~10-20sec. Mostly they were closer together.

Toflar:
You can disalbe the network by using `COMPOSER_DISABLE_NETWORK`, see https://github.com/composer/composer/blob/a2ab6f2b543a706be12b719fe3c9a5201e8c01fe/doc/03-cli.md#composer_disable_network.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -170,12 +174,15 @@ public function buildPool(array $repositories, Request $request)
...
+            $loadOperation = new LoadPackageOperation('root', 'root', $packageName, $constraint, 0);
+
             $this->maxExtendedReqs[$packageName] = true;
+            $this->packagesToLoad[$packageName] = $loadOperation;
+            $this->traceLoadPackageOperation($packageName, $loadOperation);
         }
 
         // clean up packagesToLoad for anything we manually marked loaded above
-        foreach ($this->packagesToLoad as $name => $constraint) {
+        foreach (array_keys($this->packagesToLoad) as $name) {
```
stof, line 185: this is a bad idea. It iterates twice on the array (once to build the list of keys and once for the foreach)

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -317,20 +344,24 @@ private function loadPackagesMarkedForLoading(Request $request, $repositories)
...
-            $result = $repository->loadPackages($packageBatch, $this->acceptableStabilities, $this->stabilityFlags, isset($this->loadedPerRepo[$repoIndex]) ? $this->loadedPerRepo[$repoIndex] : array());
 
-            foreach ($result['namesFound'] as $name) {
-                // avoid loading the same package again from other repositories once it has been found
-                unset($packageBatch[$name]);
-            }
-            foreach ($result['packages'] as $package) {
-                $this->loadedPerRepo[$repoIndex][$package->getName()][$package->getVersion()] = $package;
-                $this->loadPackage($request, $package);
+            foreach ($packageBatchesByLevel as $levelFoundOn => $packageBatch) {
```
stof, line 348: this overrides the `$packageBatch` variable defined at line 325 and used in the loop at line 338, which will cause issues.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -317,20 +344,24 @@ private function loadPackagesMarkedForLoading(Request $request, $repositories)
...
+                $nextLevel = $levelFoundOn + 1;
+                $result = $repository->loadPackages($packageBatch, $this->acceptableStabilities, $this->stabilityFlags, isset($this->loadedPerRepo[$repoIndex]) ? $this->loadedPerRepo[$repoIndex] : array());
+
+                foreach ($result['namesFound'] as $name) {
+                    // avoid loading the same package again from other repositories once it has been found
+                    unset($packageBatch[$name]);
```
stof, line 354: this does not achieve the result anymore, as unsetting in `$packageBatch` won't affect the `$packageBatchesByLevel` variable

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -505,5 +572,256 @@ private function removeLoadedPackage(Request $request, PackageInterface $package
...
+        // Optimization 1: Remove all package versions that are not mentioned in any top level constraint
+        $removedPackages = array();
+        $highestDependencyTreeConstraints = $this->extractHighestDependencyTreeConstraints();
+
+        foreach ($this->packages as $i => $package) {
+            if (isset($highestDependencyTreeConstraints[$package->getName()]) && !$highestDependencyTreeConstraints[$package->getName()]->matches(new Constraint('==', $package->getVersion()))) {
+                unset($this->packages[$i]);
+
+                if (!isset($removedPackages[$package->getName()])) {
+                    $removedPackages[$package->getName()] = array();
```
stof, line 638: this is useless. `$removedPackages[$package->getName()][] = $package->getVersion();` will already initialize the array if needed.

stof:
@Toflar have you looked at #8287 where I described a pool optimization aimed at filtering packages, in a way which does *not* impact the result of the solver, thanks to relying on the way Composer chooses between multiple acceptable versions of a package, and to the fact that many releases don't change their requirements (especially for patch releases) ?
Regarding your own optimization, I'm not sure introducing a concept of level is actually safe. The dependency graph is not a tree (it is not even an acyclic graph), so the level is not well defined.

>     1. Packages referenced on the same tree level in different constraints need to be requested with a disjunctive constraint. Unfortunately this will result in more packages remaining in the pool but if we didn't do this, we'd actually already start dependency resolving which we cannot do at this stage :)

AFAIK, this happens even for packages at different levels.

simonberger:
> You can disalbe the network by using `COMPOSER_DISABLE_NETWORK`, see https://github.com/composer/composer/blob/a2ab6f2b543a706be12b719fe3c9a5201e8c01fe/doc/03-cli.md#composer_disable_network.

I tried this env setting before but it throws an error when using (lazy) repositories. This might be not intended though.

Seldaek:
@simonberger you should let it run once without the env var set to prime the cache, then it's fine in theory for profiling. It will show a warning that the request failed due to network being disabled just so you are aware of it, but it doesn't prevent it from working.

simonberger:
@Seldaek I got that but It just does not work for artifactory.
Composer recovers from not being able to read a package.json (because of disabled network). But not from trying to (lazy) load a package from a composer repository.
I'll maybe create an issue for it.

Toflar:
Closing this in favor of https://github.com/composer/composer/pull/9261.

PR #9176, opened 2020-09-02
Composer2 RemoteFileSystem: response decode for TransportException

glaubinix:
The response property of TransportExceptions thrown by the CurlDownloader are always decoded. The RemoteFileSystem however throws TransportExceptions which will contain a gzipped response for gzipped responses.

### DIFF
--- a/src/Composer/Util/RemoteFilesystem.php
+++ b/src/Composer/Util/RemoteFilesystem.php
@@ -299,7 +299,12 @@ class RemoteFilesystem
                 $e = new TransportException('Content-Length mismatch, received '.Platform::strlen($result).' bytes out of the expected '.$contentLength);
                 $e->setHeaders($http_response_header);
                 $e->setStatusCode($this->findStatusCode($http_response_header));
-                $e->setResponse($this->decodeResult($result, $http_response_header));
+                try {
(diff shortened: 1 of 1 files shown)


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -299,7 +299,7 @@ protected function get($originUrl, $fileUrl, $additionalOptions = array(), $file
                 $e = new TransportException('Content-Length mismatch, received '.Platform::strlen($result).' bytes out of the expected '.$contentLength);
                 $e->setHeaders($http_response_header);
                 $e->setStatusCode($this->findStatusCode($http_response_header));
-                $e->setResponse($result);
+                $e->setResponse($this->decodeResult($result, $http_response_header));
```
glaubinix, line 302: I am not 100% sure whether we should attempt to decode for content-length mismatch

Seldaek: Yeah sounds like you rather not decode in this case, or try to decode and if it fails use the original, but not fail hard with a decode error as that just makes things worse :)

Seldaek:
Thanks

glaubinix:
Thank you!

PR #9177, opened 2020-09-03
Re-Fetch cached packages only once in a run

simonberger:
While testing the recently added feature #8850 I noticed slowdowns instead of a speedups from this feature. 
The cause are multiple requests will be done to a repository with lazy loading when a constraint of an already loaded package changes.
Some more details [here](https://github.com/composer/composer/pull/8850#issuecomment-680099238) and the subsequent comment.
With this PR a cached package will be (re-)fetched  maximally once and otherwise just used without looking for potential changes. I don't think that has to be considered in between a run.


simonberger:
I also added two small, not related code improvements

Seldaek:
Thanks, makes sense.

PR #9186, opened 2020-09-07
optimize regex by strpos checks

voku:
ping: https://github.com/phpstan/phpstan-src/pull/318

---
This change is [Reviewable](https://reviewable.io/reviews/composer/composer/9186)

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -611,7 +611,13 @@ EOF;
         foreach ($packageMap as $item) {
             list($package, $installPath) = $item;
             foreach (array_merge($package->getReplaces(), $package->getProvides()) as $link) {
-                if (preg_match('{^ext-(.+)$}iD', $link->getTarget(), $match)) {
+	            $linkTarget = $link->getTarget();
+
+            	if (
+	                strpos($linkTarget, 'ext-') === 0
+	                &&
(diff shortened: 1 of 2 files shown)


Seldaek:
https://github.com/composer/composer/pull/9187 fixed the substr vs strpos which did seem to have a minor impact, but the changes here don't really seem to do much for me in real world benchmark, while adding lots of extra complexity to the code, so I am really not sure if it's worth it. When you do a strpos it has to scan through the whole string, which is also what a preg_match does, sure it has to compile the regex first but that is cached AFAIK and on fairly large class files the strpos or preg_match I guess perform very close to each other when no match is found, while when a match is found you then end up having to scan the file twice.

Seldaek:
Also.. this is scanning 9000 classes in about 2seconds here, so it's already pretty fast IMO.  We made sure the regexes are optimized already.

So yeah, unless there is a real solid argument for adding complexity on one of them, I'd rather not add this to all regexes "just because".

voku:
@Seldaek understand and thanks for the explanation. Is there already something like a performance test or something like that, so that we can compare the results?

Seldaek:
Not really, I run `composer dump-autoload -o --profile` on a few projects to see how much time is spent. Definitely could do more specific profiling if you are interested in digging more in depth, but IMO it's kinda fast enough that it doesn't need a ton of time investment.

PR #9187, opened 2020-09-07
Micro optimization: Changed all substr calls used to compare fragments of text to strpos

simonberger:
Some additional code cleanups in those classes

I tried to only change code smells which shouldn't raise up discussions.


src/Composer/Package/Version/VersionGuesser.php
```diff
@@ -96,7 +94,7 @@ public function guessVersion(array $packageConfig, $path)
 
     private function postprocess(array $versionData)
     {
-        if (!empty($versionData['feature_version']) && $versionData['feature_version'] === $versionData['version'] && $versionData['feature_pretty_version'] === $versionData['feature_pretty_version']) {
+        if (!empty($versionData['feature_version']) && $versionData['feature_version'] === $versionData['version']) {
```
simonberger, line 97: Here it was probably planned to compare `feature_version` to `feature_pretty_version` which currently would be a bug. I just removed the useless check for now.

Seldaek: Good catch, fixed it up.

PR #9188, opened 2020-09-08
Fix self usage in closures results in fatal error on PHP 5.3

simonberger:
Added test to cancel ProcessExecutor::asyncExecute.
This failed for PHP 5.3.


Seldaek:
Ah nice catch thanks. So looking forward to dropping 5.3, if only for closures support :)

simonberger:
Who does not want to get rid of this old array syntax. 👍

PR #9197, opened 2020-09-10
Specify covered names with regexes in 2.x repos

mbaynton:
This supplements the available-packages list so that repositories may
rule themselves out of a given name (and thus not be probed with lazy
load requests) by regex, as well as by exact name match.

The use case is sizeable and varying supplemental Composer repositories
such as packages.drupal.org, which otherwise must either maintain a list
of over 10k package names in their root packages.json or accept lots of
lazy-load requests for unrelated packages that will 404.

With this, we can just do `"available-package-regexes": ["drupal/*"]`, shrinking
the root repo file considerably and keeping its contents static.

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Repository;
 
+use Composer\Package\BasePackage;
 use Composer\Package\Loader\ArrayLoader;
 use Composer\Package\PackageInterface;
 use Composer\Package\AliasPackage;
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks @mbaynton, definitely makes sense. I updated some things mostly coding style preferences in https://github.com/composer/composer/commit/2a82371adf9fb7b1d02c80f1b1636bfea5815ed5 FYI

Seldaek:
2.0RC1 is out with this merged in, so hopefully it's quickly worth it for you to implement :)

gitressa:
Thanks @mbaynton and @Seldaek for writing the new feature, and rolling it out so quickly. I look forward to testing it, when it has been integrated in Drupal. Or can I already try it out? I suppose the steps could be something like this:

1. Update to the latest RC version (2.x) with `composer self-update --preview`
2. Edit composer.json, adding `"available-package-regexes": ["drupal/*"]` somewhere?

mbaynton:
I saw that RC1 tweet hours later. Doesn't help my reputation for procrastination, but hey, if it's in it's in :laughing: 

@gitressa The new attribute has to come from the server side root file, not the local project composer.json. I've tested it on a trivial composer repository but packages.drupal.org isn't a 2.x repository yet. It's on the DA's roadmap and this specific solution was done in collaboration with @ryanaslett so once they've gotten it 2.x-conformant, this trick will be used.

gitressa:
Thanks for detailing the intricacies @mbaynton, I thought the scenario might be something like you outline, and difficult to test locally. So I will be looking forward to it getting integrated into Drupal proper :)

PR #9203, opened 2020-09-11
Do not hard fail on lazy repository when network is disabled and no package cache exists

simonberger:
While testing some performance improvements I noticed disabling network does not work in projects I use lazy repositories in (discussed in https://github.com/composer/composer/pull/9171#issuecomment-690617374).
The problems are appearing if a lazy repository does not have this package and as a result of this no cache item can be found.

I added status code 499 beside 404 to not fail hard in this situation but still log the warning as it happens on regular repositories as well.


Seldaek:
Thanks

PR #9204, opened 2020-09-11
Code improvements

simonberger:
Some code improvements. All commits (should) fix all occurrences of this issue.
My suggestion is to review commit by commit instead of the whole diff.
I tried to leave all changes worthy of discussion out of this merge request.


src/Composer/Util/Git.php
```diff
@@ -79,9 +79,10 @@ public function runCommand($commandCallable, $url, $cwd, $initialClone = false)
                     return;
                 }
                 $messages[] = '- ' . $protoUrl . "\n" . preg_replace('#^#m', '  ', $this->process->getErrorOutput());
-                if ($initialClone && isset($origCwd)) {
-                    $this->filesystem->removeDirectory($origCwd);
-                }
+            }
+
+            if ($initialClone && isset($origCwd)) {
```
simonberger, line 84: This is a strange issue. I could only think it accidentally slid inside the foreach.

Seldaek: No I think it was intended. Just to make sure the whole directory gets wiped clean between clone attempts in case some stuff was left over from the previous attempt. Please revert.

src/Composer/Installer.php
```diff
@@ -996,7 +996,7 @@ public function setOptimizeAutoloader($optimizeAutoloader = false)
         if (!$this->optimizeAutoloader) {
             // Force classMapAuthoritative off when not optimizing the
             // autoloader
-            $this->setClassMapAuthoritative(false);
```
stof, line 999: IMO, the explicit `false` here makes the code more self-explanatory

src/Composer/Platform/HhvmDetector.php
```diff
@@ -35,7 +35,7 @@ public function reset()
     public function getVersion()
     {
         if (null !== self::$hhvmVersion) {
-            return self::$hhvmVersion ?: null;
+            return self::$hhvmVersion;
```
stof, line 38: this is broken, as it won't turn `false` into `null` anymore.

src/Composer/Installer.php
simonberger: Yes that's a discord and I re-add it if there is another vote for.
A function name is badly named or default value shouldn't be there if it is needed to understand the code and should be fixed in the long go then.
Passing a default value is a code smell imho. Even if a very tiny one.

stof: `$this->setClassMapAuthoritative()` making it non authoritative is indeed a code smell to me. but the Installer class is one of the main entry points of composer and that's a public method. So making it a mandatory argument would require triggering a deprecation warning in 1.x first.

Seldaek: Yeah the method probably should have no default value tbh. If anything I'd fix that. But please rather in a separate PR. This has enough stuff in it already, and frankly given all these changes have potential for breakage, it's horrible to review and brings little value overall. There are clear simple cleanups and then there are risky not worth it cleanups IMO. Please try not to get carried away :)

Seldaek: @stof given this would break any bad usage very clearly and very quickly, and that it's super easy to fix, I think it'd be fine without deprecation notice. It's not like there are many people integrating with the Installer class directly, I can only think of one actually, tho I can imagine there are a few more. Also we do not make any BC promises except our best effort to follow semver.

simonberger: Ok I understand the history at least if this is the result of a later change. I'll revert it here. Can be "fixed" somewhere else.

src/Composer/Compiler.php
```diff
@@ -199,9 +199,10 @@ private function addFile($phar, $file, $strip = true)
         }
 
         if ($path === 'src/Composer/Composer.php') {
-            $content = str_replace('@package_version@', $this->version, $content);
-            $content = str_replace('@package_branch_alias_version@', $this->branchAliasVersion, $content);
-            $content = str_replace('@release_date@', $this->versionDate->format('Y-m-d H:i:s'), $content);
+            $content = str_replace(
```
Ayesh, line 202: Wouldn't a `strtr` call with from=>to pattern make it more readable?

Ayesh:
Pretty nice improvements 👍🏼.

src/Composer/Compiler.php
simonberger: I agree. Changed and used constants from `Composer` class

src/Composer/Compiler.php
```diff
@@ -199,9 +199,14 @@ private function addFile($phar, $file, $strip = true)
...
         if ($path === 'src/Composer/Composer.php') {
-            $content = str_replace('@package_version@', $this->version, $content);
-            $content = str_replace('@package_branch_alias_version@', $this->branchAliasVersion, $content);
-            $content = str_replace('@release_date@', $this->versionDate->format('Y-m-d H:i:s'), $content);
+            $content = strtr(
+                $content,
+                array(
+                    Composer::VERSION => $this->version,
+                    Composer::BRANCH_ALIAS_VERSION => $this->branchAliasVersion,
+                    Composer::RELEASE_DATE => $this->versionDate->format('Y-m-d H:i:s')
```
stof, line 207: Using the constants is not a good idea IMO. These constants are precisely the place where we want to do the replacement. They are not meant to always represent the replaced placeholders. So I suggest going back to using the placeholder strings.

simonberger: Short answer:
I changed it as you suggested.

Long answer:
In my understanding this is always a one time replacement and the value of the const I want to replace should best know the replacement candidate.
Yet I got your point and this MR also shall not include those kind of discussable changes.

src/Composer/Compiler.php
Ayesh: Thank you.

src/Composer/Repository/Vcs/FossilDriver.php
```diff
@@ -165,7 +165,7 @@ public function getFileContent($file, $identifier)
     public function getChangeDate($identifier)
     {
         $this->process->execute('fossil finfo -b -n 1 composer.json', $output, $this->checkoutDir);
-        list($ckout, $date, $message) = explode(' ', trim($output), 3);
+        list(,$date) = explode(' ', trim($output), 2);
```
Seldaek, line 168: This change means $date now contains date + message, as you changed the limit to 2. I will fix but just so you know.. That's exactly why I'm scared of such seemingly harmless PRs doing "cleanups". The value add isn't worth the risk of missing such regression in review IMO.

src/Composer/Downloader/FileDownloader.php
```diff
@@ -338,7 +338,7 @@ public function update(PackageInterface $initial, PackageInterface $target, $pat
         $this->io->writeError("  - " . UpdateOperation::format($initial, $target) . ": ", false);
 
         $promise = $this->remove($initial, $path, false);
-        if (!$promise instanceof PromiseInterface) {
+        if ($promise === null || !$promise instanceof PromiseInterface) {
```
Seldaek, line 356: What's the point of this? null is definitely not going to be an instance of PromiseInterface

Seldaek:
Thanks

PR #9212, opened 2020-09-14
Specify APCu autoloader prefix

### LINKED ISSUE #7049: Reproducable build: deterministic APCU prefix
Each time when `composer install --no-dev --optimize-autoloader --apcu-autoloader` is run, a new random prefix is generated 

https://github.com/composer/composer/blob/be040f8e3111b38c1e49ae5cd6b929dbf5c39045/src/Composer/Autoload/AutoloadGenerator.php#L657

This is problematic if one tries to achieve reproducible builds where the same input leads to the same output. It would be cool if composer would offer an option to override the prefix or make it deterministic based on the `composer.lock` hash. Any other idea is welcome as well.

lstrojny:
Closes #7049 

Introduces new option `--apcu-autoloader-prefix` to `install`, `require`, `update`, `remove` and `--apcu-prefix` in `dump-autoload`.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -53,9 +53,14 @@ class AutoloadGenerator
     private $classMapAuthoritative = false;
 
     /**
-     * @var bool|string
+     * @var bool
      */
     private $apcu = false;
+
+    /**
+     * @var string|null
+     */
+    private $apcuPrefix;
 
     /**
      * @var bool
@@ -92,11 +97,13 @@ class AutoloadGenerator
     /**
      * Whether or not generated autoloader considers APCu caching.
      *
-     * @param bool|string $apcu
+     * @param bool        $apcu
+     * @param string|null $apcuPrefix
      */
-    public function setApcu($apcu)
+    public function setApcu($apcu, $apcuPrefix = null)
     {
-        $this->apcu = is_string($apcu) ? $apcu : (bool) $apcu;
+        $this->apcu = (bool) $apcu;
+        $this->apcuPrefix = $apcuPrefix !== null ? (string) $apcuPrefix : $apcuPrefix;
     }
 
     /**
@@ -858,7 +865,7 @@ CLASSMAPAUTHORITATIVE;
         }
 
         if ($this->apcu) {
-            $apcuPrefix = var_export(is_string($this->apcu) ? $this->apcu : substr(base64_encode(md5(uniqid('', true), true)), 0, -3), true);
+            $apcuPrefix = var_export(($this->apcuPrefix ?: substr(base64_encode(md5(uniqid('', true), true)), 0, -3)), true);
             $file .= <<<APCU
         \$loader->setApcuPrefix($apcuPrefix);
 
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -858,9 +858,9 @@ CLASSMAPAUTHORITATIVE;
         }
 
         if ($this->apcu) {
-            $apcuPrefix = is_string($this->apcu) ? $this->apcu : substr(base64_encode(md5(uniqid('', true), true)), 0, -3);
+            $apcuPrefix = var_export(is_string($this->apcu) ? $this->apcu : substr(base64_encode(md5(uniqid('', true), true)), 0, -3), true);
             $file .= <<<APCU
-        \$loader->setApcuPrefix('$apcuPrefix');
+        \$loader->setApcuPrefix($apcuPrefix);
 
 APCU;
         }
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -53,7 +53,7 @@ class AutoloadGenerator
     private $classMapAuthoritative = false;
 
     /**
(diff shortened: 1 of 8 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -858,7 +858,7 @@ public static function getLoader()
         }
 
         if ($this->apcu) {
-            $apcuPrefix = substr(base64_encode(md5(uniqid('', true), true)), 0, -3);
+            $apcuPrefix = is_string($this->apcu) ? $this->apcu : substr(base64_encode(md5(uniqid('', true), true)), 0, -3);
```
stof, line 861: if we allow user input, we cannot assume that the content is fine to output in a string without any escaping. So we must use `var_export($value, true)` to generate the string representation instead (removing the quotes around the variables when using it in the code just below). Otherwise, we might generate invalid code.

src/Composer/Command/RemoveCommand.php
```diff
@@ -54,7 +54,7 @@ protected function configure()
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore all platform requirements (php & ext- packages).'),
                 new InputOption('optimize-autoloader', 'o', InputOption::VALUE_NONE, 'Optimize autoloader during autoloader dump'),
                 new InputOption('classmap-authoritative', 'a', InputOption::VALUE_NONE, 'Autoload classes from the classmap only. Implicitly enables `--optimize-autoloader`.'),
-                new InputOption('apcu-autoloader', null, InputOption::VALUE_NONE, 'Use APCu to cache found/not-found classes.'),
+                new InputOption('apcu-autoloader', null, InputOption::VALUE_OPTIONAL, 'Use APCu to cache found/not-found classes (passing a string sets the APCu prefix)'),
```
stof, line 57: what happens when running `composer remove --apcu-autoloader some/package` ? Is `some/package` still treated as the argument or does it become the value of the option ?

lstrojny: Uh, yeah, that’s indeed a problem and one that can probably only be solved by upgrading. What would work is a separate parameter with a required value, so `--apc-autoloader-prefix`. @Seldaek what do you think?

Seldaek: I think this is not that important and can maybe wait until 2.1, because 2.0 should really be stabilized and shipped.

stof: @lstrojny what is the min Symfony version in which this works fine ?

src/Composer/Command/DumpAutoloadCommand.php
```diff
@@ -33,7 +33,7 @@ protected function configure()
                 new InputOption('no-scripts', null, InputOption::VALUE_NONE, 'Skips the execution of all scripts defined in composer.json file.'),
                 new InputOption('optimize', 'o', InputOption::VALUE_NONE, 'Optimizes PSR0 and PSR4 packages to be loaded with classmaps too, good for production.'),
                 new InputOption('classmap-authoritative', 'a', InputOption::VALUE_NONE, 'Autoload classes from the classmap only. Implicitly enables `--optimize`.'),
-                new InputOption('apcu', null, InputOption::VALUE_NONE, 'Use APCu to cache found/not-found classes.'),
+                new InputOption('apcu', null, InputOption::VALUE_OPTIONAL, 'Use APCu to cache found/not-found classes.'),
```
staabm, line 36: Description should mention how to use with a string  argument

src/Composer/Command/RemoveCommand.php
lstrojny: @stof we would need at least this change https://github.com/symfony/symfony/commit/80867422acd83d2602187155e7eb978d6c3d0d44, which is earliest available in 3.3.0

But: in `RequireCommand` and `RemoveCommand` we would have both an option with an optional value (`apcu-autoloader`) as well as an argument with an optional value (`packages`). Assume `composer require --apcu-autoloader package/name` and the `apcu-autoloader` this will lead to `package/name` being treated as the value of the `apcu-autoloader` option. `composer require --apcu-autoloader -- package/name` would work as it clarifies what is meant. Doing so would be a BC break and unfortunately the worst one: a command that was previously working might silently change behavior.

In conclusion: we need to introduce a separate option because a newer Symfony version will only fix half of the problem to preserve backwards compatibility.

(My current stance would be that Symfony should probably forbid having options and arguments with optional values in the same command as it will automatically lead to a quite counter-intuitive user experience)

lstrojny:
The current version of this PR introduces a separate option which is the only sensible path forward I see. Question: should passing an APCu autoloader prefix automatically turn on APCu autoloading?

Seldaek:
Ok yeah probably makes sense the way it is. It's not like this flag will be needed by a ton of people anyway.. I'd say passing the prefix should turn on the apcu option too to avoid having to type this twice, because really setting the prefix if you are not using apcu makes no sense.

lstrojny:
@Seldaek done and done

lstrojny:
@Seldaek anything else needed?

Seldaek:
Just free time to review one last time.

Seldaek:
But it's in the right milestone, it won't get forgotten

lstrojny:
Great

Seldaek:
Thanks @lstrojny

Issue #9210, opened 2020-09-14
platform requirements do not have a pretty constraint (composer 2.0-RC1)


mitelg:
Is there a reason why [here (PlatformRepository)](https://github.com/composer/composer/blob/master/src/Composer/Repository/PlatformRepository.php#L589) no pretty constraint is passed to the platform requirements?

We read them out and got an error with composer 2
Works fine with composer 1.x

Seldaek:
Probably no good reason no. Could be fixed. Feel free to send a PR fixing it for the couple cases in this class.

mitelg:
okay cool 👍 I will do that

PR #9214, opened 2020-09-15
Add pretty constraint

mitelg:
Fix for #9210 

the PlatformRepository now sets the pretty constraint to Links.

I also introduced constants for the different Link types

### DIFF
--- a/src/Composer/Package/BasePackage.php
+++ b/src/Composer/Package/BasePackage.php
@@ -22,12 +22,15 @@ use Composer\Repository\PlatformRepository;
  */
 abstract class BasePackage implements PackageInterface
 {
+    /**
+     * @phpstan-var array<string, array{description: string, method: Link::TYPE_*}>
+     */
     public static $supportedLinkTypes = array(
-        'require' => array('description' => 'requires', 'method' => 'requires'),
-        'conflict' => array('description' => 'conflicts', 'method' => 'conflicts'),
-        'provide' => array('description' => 'provides', 'method' => 'provides'),
-        'replace' => array('description' => 'replaces', 'method' => 'replaces'),
-        'require-dev' => array('description' => 'requires (for development)', 'method' => 'devRequires'),
+        'require' => array('description' => 'requires', 'method' => Link::TYPE_REQUIRE),
+        'conflict' => array('description' => 'conflicts', 'method' => Link::TYPE_CONFLICT),
+        'provide' => array('description' => 'provides', 'method' => Link::TYPE_PROVIDE),
+        'replace' => array('description' => 'replaces', 'method' => Link::TYPE_REPLACE),
+        'require-dev' => array('description' => 'requires (for development)', 'method' => Link::TYPE_DEV_REQUIRE),
     );
 
     const STABILITY_STABLE = 0;
--- a/src/Composer/Package/Link.php
+++ b/src/Composer/Package/Link.php
@@ -21,6 +21,26 @@ use Composer\Semver\Constraint\ConstraintInterface;
  */
 class Link
 {
+    const TYPE_REQUIRE = 'requires';
+    const TYPE_DEV_REQUIRE = 'devRequires';
+    const TYPE_PROVIDE = 'provides';
+    const TYPE_CONFLICT = 'conflicts';
+    const TYPE_REPLACE = 'replaces';
+
+    /**
+     * Will be converted into a constant once the min PHP version allows this
+     *
+     * @private
+     * @var string[]
+     */
+    public static $TYPES = array(
+        self::TYPE_REQUIRE,
+        self::TYPE_DEV_REQUIRE,
+        self::TYPE_PROVIDE,
+        self::TYPE_CONFLICT,
+        self::TYPE_REPLACE,
+    );
+
     /**
      * @var string
      */
@@ -38,6 +58,7 @@ class Link
 
     /**
      * @var string
+     * @phpstan-var self::TYPE_* $description
      */
     protected $description;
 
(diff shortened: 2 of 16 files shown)


src/Composer/Repository/PlatformRepository.php
```diff
@@ -585,8 +587,8 @@ private function addLibrary($name, $prettyVersion, $description = null, array $r
         $lib = new CompletePackage('lib-'.$name, $version, $prettyVersion);
         $lib->setDescription($description);
 
-        $links = function ($alias) use ($name, $version) {
-            return new Link('lib-'.$name, 'lib-'.$alias, new Constraint('=', $version));
+        $links = function ($alias) use ($name, $version, $lib) {
+            return new Link('lib-'.$name, 'lib-'.$alias, new Constraint('=', $version), Link::LINK_TYPE_REPLACE, $lib->getPrettyVersion());
```
mitelg, line 591: I am not sure, if the pretty version of the lib is also the pretty constraint. 

I also tried to call `getPrettyString` on the constraint object, but then I get something like `== 1.2.3.4`

Please let me know, what is the desired behaviour :blush:

mitelg:
I guess the failing tests are not related to my changes, as they are also failing on the master branch?

src/Composer/Package/Link.php
```diff
@@ -21,6 +21,20 @@
  */
 class Link
 {
+    const LINK_TYPE_REQUIRE = 'requires';
+    const LINK_TYPE_DEV_REQUIRE = 'devRequires';
+    const LINK_TYPE_PROVIDE = 'provides';
+    const LINK_TYPE_CONFLICT = 'conflicts';
+    const LINK_TYPE_REPLACE = 'replaces';
```
stof, line 28: I suggest removing the `LINK_` prefix, given that they are in a `Link` class already. `Link::TYPE_REQUIRE` is readable enough

mitelg: done

src/Composer/Package/BasePackage.php
```diff
@@ -23,11 +23,11 @@
 abstract class BasePackage implements PackageInterface
 {
     public static $supportedLinkTypes = array(
```
staabm, line 25: I guess this could be typed like

```suggestion
    /**
     * @phpstan-var array<string, array<description: string, method: Link::TYPE_*>>
     */
    public static $supportedLinkTypes = array(
```

mitelg: thanks, I added this

src/Composer/Package/Locker.php
```diff
@@ -213,7 +213,7 @@ public function getPlatformRequirements($withDevReqs = false)
             $requirements = $this->loader->parseLinks(
                 '__root__',
                 '1.0.0',
-                'requires',
+                Link::TYPE_REQUIRE,
```
staabm, line 216: looks like the `parseLinks()` method could also benefit from a more specialized typehint now:
```
/**
 * @phpstan-param Link::TYPE_* $description
 */
```

in https://github.com/mitelg/composer/blob/f957a192cb143ad127eca77c496670689b756fcc/src/Composer/Package/Loader/ArrayLoader.php#L298-L305

------------

same for the `$description` parameter in the `Link` class

https://github.com/mitelg/composer/blob/f957a192cb143ad127eca77c496670689b756fcc/src/Composer/Package/Link.php#L63-L72

mitelg: I add the additional annotation. I hope in the correct way :grin:

mitelg: at lease the PHPStan check is fine with it :+1:

src/Composer/Package/BasePackage.php
```diff
@@ -22,12 +22,15 @@
  */
 abstract class BasePackage implements PackageInterface
 {
+    /**
+     * @phpstan-var array<string, array<description: string, method: Link::TYPE_*>>
```
staabm, line 26: hmm I guess I had a typo in here

```suggestion
     * @phpstan-var array<string, array{description: string, method: Link::TYPE_*}>
```

src/Composer/Package/Link.php
```diff
@@ -38,6 +52,7 @@ class Link
 
     /**
      * @var string
+     * @phpstan-var self::TYPE_* $description
```
staabm, line 55: please also add it for the $desciption parameter in `__construct()`

src/Composer/Package/BasePackage.php
mitelg: yeah, your are right https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

will fix this

mitelg: I'm wondering why PHPStan does not complain about the syntax? :thinking:

src/Composer/Package/Link.php
mitelg: done

src/Composer/Package/BasePackage.php
staabm: > I'm wondering why PHPStan does not complain about the syntax?

it seems phpstan only errors when the level is 2 or higher, see
https://phpstan.org/r/b8878f4d-af86-4d0c-a39b-8fabb4811980

mitelg: oh okay, didn't recognized the low level in the PHPStan config 🙈

src/Composer/Package/Link.php
```diff
@@ -21,6 +21,20 @@
  */
 class Link
 {
+    const TYPE_REQUIRE = 'requires';
+    const TYPE_DEV_REQUIRE = 'devRequires';
+    const TYPE_PROVIDE = 'provides';
+    const TYPE_CONFLICT = 'conflicts';
+    const TYPE_REPLACE = 'replaces';
+
+    public static $types = array(
```
Seldaek, line 30: I'd make this $TYPES and mark it `@private` for now, then we can convert it to a proper const once we drop old php versions, and make it publicly usable.

src/Composer/Repository/PlatformRepository.php
Seldaek: This is fine, tho you could simply reuse $prettyVersion, but it's really the same..

Seldaek:
No don't worry about the php8 failures. Aside from that minor comment about $types, looks good to me 👍

Seldaek:
Thanks!

PR #9220, opened 2020-09-16
Composer 2: Adapt file download cache keys to changed processed URLs

ffraenz:
Following up the introduction of a `setProcessedUrl` method on the `PreFileDownloadEvent` in https://github.com/composer/composer/pull/8975 we found another issue that is blocking us from upgrading plugins to Composer 2. When altering the `processedUrl` for packages the file download `cacheKey` unexpectedly does not change with it, potentially reusing outdated files.

I am therefore suggesting that the cache key gets regenerated when the processed URL gets changed during `PreFileDownloadEvent` dispatching.

For [use cases](https://github.com/ffraenz/private-composer-installer/pull/25#issuecomment-692792485) that introduce temporary processed URLs (e.g. timestamps, session tokens, etc.) further optimization may be possible by allowing plugin authors to optionally provide a custom cache key in `PreFileDownloadEvent`. They may leave time-related params out of the value from which the cache key hash is generated.

The two proposed changes only apply to file downloads of type `package` as the cache keys of `metadata` downloads do not depend on the processed URL.

/cc @mcaskill

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -106,6 +106,11 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
             throw new \InvalidArgumentException('The given package is missing url information');
         }
 
+        $generateCacheKey = function (PackageInterface $package, $key) {
+            $cacheKey = sha1($key);
+            return $package->getName().'/'.$cacheKey.'.'.$package->getDistType();
+        };
+
         $retries = 3;
         $urls = $package->getDistUrls();
         foreach ($urls as $index => $url) {
@@ -117,7 +122,7 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
                 // from different packages, which would potentially allow a given package
                 // in a third party repo to pre-populate the cache for the same package in
                 // packagist for example.
-                'cacheKey' => $this->getCacheKey($package, $processedUrl)
+                'cacheKey' => $generateCacheKey($package, $processedUrl)
             );
         }
 
@@ -134,16 +139,16 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
 
         $accept = null;
         $reject = null;
-        $download = function () use ($io, $output, $httpDownloader, $cache, $eventDispatcher, $package, $fileName, &$urls, &$accept, &$reject) {
+        $download = function () use ($io, $output, $httpDownloader, $cache, $generateCacheKey, $eventDispatcher, $package, $fileName, &$urls, &$accept, &$reject) {
             $url = reset($urls);
 
             if ($eventDispatcher) {
                 $preFileDownloadEvent = new PreFileDownloadEvent(PluginEvents::PRE_FILE_DOWNLOAD, $httpDownloader, $url['processed'], 'package', $package);
                 $eventDispatcher->dispatch($preFileDownloadEvent->getName(), $preFileDownloadEvent);
                 if ($preFileDownloadEvent->getCustomCacheKey() !== null) {
-                    $url['cacheKey'] = $this->getCacheKey($package, $preFileDownloadEvent->getCustomCacheKey());
+                    $url['cacheKey'] = $generateCacheKey($package, $preFileDownloadEvent->getCustomCacheKey());
                 } else if ($preFileDownloadEvent->getProcessedUrl() !== $url['processed']) {
-                    $url['cacheKey'] = $this->getCacheKey($package, $preFileDownloadEvent->getProcessedUrl());
+                    $url['cacheKey'] = $generateCacheKey($package, $preFileDownloadEvent->getProcessedUrl());
                 }
                 $url['processed'] = $preFileDownloadEvent->getProcessedUrl();
             }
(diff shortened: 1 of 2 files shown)


src/Composer/Plugin/PreFileDownloadEvent.php
```diff
@@ -88,6 +93,26 @@ public function setProcessedUrl($processedUrl)
         $this->processedUrl = $processedUrl;
     }
 
+    /**
+     * Retrieves a custom package cache key for this download.
+     *
+     * @return string
```
stof, line 99: `string|null` is the right type.

src/Composer/Plugin/PreFileDownloadEvent.php
```diff
@@ -88,6 +93,26 @@ public function setProcessedUrl($processedUrl)
...
+     */
+    public function getCustomCacheKey()
+    {
+        return $this->customCacheKey;
+    }
+
+    /**
+     * Sets a custom package cache key for this download.
+     *
+     * @param string $customCacheKey New cache key
```
stof, line 109: `string|null` is the right type.

src/Composer/Plugin/PreFileDownloadEvent.php
```diff
@@ -32,6 +32,11 @@ class PreFileDownloadEvent extends Event
      */
     private $processedUrl;
 
+    /**
+     * @var string
```
stof, line 36: `string|null` is the right type.

stof (changes requested):

src/Composer/Plugin/PreFileDownloadEvent.php
ffraenz: Thank you for the heads up. I fixed the PHPDoc types on this property.

src/Composer/Downloader/FileDownloader.php
```diff
@@ -106,14 +106,23 @@ public function download(PackageInterface $package, $path, PackageInterface $pre
             throw new \InvalidArgumentException('The given package is missing url information');
         }
 
+        $generateCacheKey = function (PackageInterface $package, $key) {
```
GrahamCampbell, line 109: `$cacheKeyFactory` or `$cacheKeyGenerator` is probably a better name of that variable.

src/Composer/Downloader/FileDownloader.php
```diff
@@ -106,14 +106,23 @@ public function download(PackageInterface $package, $path, PackageInterface $pre
...
+            $cacheKey = sha1($key);
```
GrahamCampbell, line 110: Need a blank line before the return statement.

GrahamCampbell (changes requested):

mcaskill:
@ffraenz If there's a way to add me as a collaborator on this pull request, it would my pleasure to help with any changes requested by the Composer team.

src/Composer/Downloader/FileDownloader.php
ffraenz: Thank you for bringing this up. I renamed the variable to `$cacheKeyGenerator`.

src/Composer/Downloader/FileDownloader.php
ffraenz: Done!

ffraenz:
@mcaskill I added you to the PR repository, you should be able to push into it.

Seldaek:
Thanks, looks good!

mcaskill:
Hi @ffraenz,

I've been testing out 2.0.0-RC2 and I've noticed we missed how `FileDownloader::download()` defines and uses its callback functions (`$download`, `$accept`, `$reject`). The `$urls` variable is passed by reference to each callback but the manipulations, in `$download`, made to `$url` are done on a copy:

```php
$download = function () use ($io, $output, $httpDownloader, $cache, $cacheKeyGenerator, $eventDispatcher, $package, $fileName, &$urls, &$accept, &$reject) {
    $url = reset($urls);
```

Once `$accept` is called, it still has the original dataset:

```php
$accept = function ($response) use ($cache, $package, $fileName, $self, &$urls) {
    $url = reset($urls);
```

We would need to do the following to ensure the manipulated `$url` is reflected on `$urls`:

```php
$download = function () use ($io, $output, $httpDownloader, $cache, $cacheKeyGenerator, $eventDispatcher, $package, $fileName, &$urls, &$accept, &$reject) {
    $url   = reset($urls);
    $index = key($urls);

    if ($eventDispatcher) {
        // ...
    }

    $urls[$index] = $url;
```

Also, I'm still having issues with extracting the cached package and writing the filtered dist URL to the lock file. I've finally installed the Composer repository to be able to tinker with it while testing out our plugins.

Have you noticed these issues or anything else?

Thanks,

P.S.: @Seldaek Sorry, for the incomplete testing and implementation of this pull request.

Seldaek:
@mcaskill no worries, just glad if you can send a PR but no rush, I don't really think this stuff affects anyone else right now.

mcaskill:
Regarding the filtering of the dist URL for the lock file, I've got that resolved.

Plugin developers that need to manipulate the dist URL will need to use `PRE_POOL_CREATE` in Composer v2 and `PRE_PACKAGE_INSTALL` and `PRE_PACKAGE_UPDATE` in Composer v1.

Note: in using `PRE_POOL_CREATE`, we technically don't need `PreFileDownloadEvent::setCustomCacheKey()` after all (confirmed but unsubstantiated) but its presence does require developers to explicitly assign a custom cache key to prevent the manipulated processed URL from overwriting the original cache key.

On my end, all that's left to resolve is the extraction of the cached package. I still end up with an empty directory where the package should be.

Issue #8870, opened 2020-05-04
composer update packages on remove a package


vnagara:
Update: Next my comment describes how to reproduce it easier:

Cross it out: I tried to reduce composer to minimal version to reproduce bug and unfortunately the bug doesn't reveal. I have to add my lock file to reproduce the bug as it very tricky.

Prerequisite: download composer.lock (https://gist.github.com/vnagara/1b097e6e941e7fae724c6b4619743c33) and do `composer install` first.

My `composer.json`:

```json
{
    "name": "symfony/skeleton",
    "type": "project",
    "license": "MIT",
    "description": "A minimal Symfony project recommended to create bare bones applications",
    "require": {
        "php": "^7.2.5",
        "ext-ctype": "*",
        "ext-iconv": "*",
        "symfony/apache-pack": "^1.0",
        "symfony/console": "*",
        "symfony/dom-crawler": "5.0.*",
        "symfony/dotenv": "*",
        "symfony/flex": "^1.3.1",
        "symfony/form": "5.0.*",
        "symfony/framework-bundle": "*",
        "symfony/security-csrf": "5.0.*",
        "symfony/twig-pack": "^1.0",
        "symfony/var-dumper": "5.0.*",
        "symfony/yaml": "*"
    },
    "require-dev": {
    },
    "config": {
        "preferred-install": {
            "*": "dist"
        },
        "sort-packages": true,
        "platform": {
            "php": "7.2.29"
        }        
    },
    "autoload": {
        "psr-4": {
            "App\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "App\\Tests\\": "tests/"
        }
    },
    "replace": {
        "symfony/intl": "5.*",
        "paragonie/random_compat": "2.*",
        "symfony/polyfill-ctype": "*",
        "symfony/polyfill-iconv": "*",
        "symfony/polyfill-php72": "*",
        "symfony/polyfill-php71": "*",
        "symfony/polyfill-php70": "*",
        "symfony/polyfill-php56": "*"
    },
    "scripts": {
        "auto-scripts": {
            "cache:clear": "symfony-cmd",
            "assets:install %PUBLIC_DIR%": "symfony-cmd"
        },
        "post-install-cmd": [
            "@auto-scripts"
        ],
        "post-update-cmd": [
            "@auto-scripts"
        ]
    },
    "conflict": {
        "symfony/symfony": "*"
    },
    "extra": {
        "symfony": {
            "allow-contrib": false,
[...]

PR #9223, opened 2020-09-17
Partial updates should remove all unused dependencies

naderman:
Instead of marking locked packages as fixed, we change the pool builder to load only the locked version and treat it like a fixed package for loading purposes. But removing the actual request fix, makes the solver treat it as a regular optional dependency. As a consequence locked packages may be removed on a partial update of another package, but they still cannot be updated. This keeps the performance improvements achieved by fixing packages (no loading of alternatives unless explicitly listed for update) but it removes the automatic decision to install locked packages.

Tests fail because this changes some other assumptions, but I believe in a way that we would actually prefer.

For example a locked dependency may now get removed in favour of another package which can satisfy the same dependency when running composer require.

This is made clear by `update-allow-list-require-new-replace.test`: Here the requirement new/pkg is added and a partial update for the new package is run, however it replaces a dependency of another package which previously could not get removed during a partial update, and the update resulted in a conflict. With this PR it now removes the locked dependency and installs the new package without a problem.

Addresses https://github.com/composer/composer/issues/8870

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -81,7 +81,7 @@ class PoolBuilder
     /**
      * @psalm-var list<Package>
      */
-    private $unacceptableFixedPackages = array();
+    private $unacceptableFixedOrLockedPackages = array();
     private $updateAllowList = array();
     private $skippedLoad = array();
 
@@ -90,7 +90,7 @@ class PoolBuilder
      * have already their maximum required range loaded and can not be
      * extended by markPackageNameForLoading
      *
-     * Packages get cleared from this list if they get unfixed as in that case
+     * Packages get cleared from this list if they get unlocked as in that case
      * we need to actually load them
      */
     private $maxExtendedReqs = array();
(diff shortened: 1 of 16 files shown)


naderman:
Cleanup is required: tests need to be updated and we probably want to clarify comments as well as some variable names to distinguish between locked packages and fixed packages in the pool builder now.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -129,7 +129,8 @@ public function buildPool(array $repositories, Request $request)
 
             foreach ($request->getLockedRepository()->getPackages() as $lockedPackage) {
                 if (!$this->isUpdateAllowed($lockedPackage)) {
-                    $request->fixPackage($lockedPackage);
+                    $request->lockPackage($lockedPackage);
+                    $fixedOrLockedPackages[] = $lockedPackage;
```
Seldaek, line 133: This var seems unused

src/Composer/DependencyResolver/Rule.php
```diff
@@ -180,11 +184,22 @@ public function getPrettyString(RepositorySet $repositorySet, Request $request,
                     return 'No package found to satisfy root composer.json require '.$packageName.($constraint ? ' '.$constraint->getPrettyString() : '');
                 }
 
+                $packagesNonAlias = array_values(array_filter($packages, function ($p) {
+                    return !($p instanceof AliasPackage);
+                }));
+                if (count($packagesNonAlias) === 1) {
+                    $package = $packagesNonAlias[0];
+                    if (!($package instanceof AliasPackage) && $request->isLockedPackage($package)) {
```
Seldaek, line 192: I guess the instanceof here is redundant

tests/Composer/Test/Fixtures/installer/root-alias-change-with-circular-dep.test
```diff
@@ -62,6 +62,4 @@ Your lock file does not contain a compatible set of packages. Please run compose
     - b/requirer is locked to version 1.0.0 and an update of this package was not requested.
     - b/requirer 1.0.0 requires root/pkg ^1 -> found root/pkg[2.x-dev] but it does not match the constraint.
 
-Use the option --with-all-dependencies to allow upgrades, downgrades and removals for packages currently locked to specific versions.
-
```
Seldaek, line 66: Not sure why the hint got lost here, but would be nice to keep it IMO?

naderman: This is on an install from lock file, the message above tells you to run composer update. No need to list this hint in this context?

Seldaek: Ah yeah missed that it was install, then for sure makes sense.

PR #9226, opened 2020-09-17
Exclude conflicts early (Performance)

simonberger:
Starting from a discussion and a test in #8850 this excludes all conflicts from not already loaded packages.
Most of the new code comes from a branch in the composer/semver package. Because I am not satisfied with the quality it is not part of a PR yet.

I am especially interested how much this reduces the analyze packages and runtime of your projects.
And surely if it raises any problems.

Here is a [composer-conflicts-test.phar.zip](https://github.com/composer/composer/files/5241611/composer-conflicts-test.phar.zip) file

**How to test:**
- Run `composer self-update --snapshot` to get the latest 2.0 dev build
- Run `composer update --dry-run` to prime the cache
- Run `COMPOSER_DISABLE_NETWORK=1 composer update -v --dry-run --profile` to run a profile without network jitter, with current master branch
- Run `COMPOSER_DISABLE_NETWORK=1 php composer-conflicts-test.phar update -v --dry-run --profile` to run another profile with the optimizations in this branch
- The three interesting lines from the snapshot and test runs are:
`Analyzed 13476 packages to resolve dependencies  Analyzed 1590120 rules to resolve dependencies`  and `Memory usage: 107.02MiB (peak: 955.09MiB), time: 57.85s`

I took parts of @Seldaek [instruction](https://github.com/composer/composer/pull/8850#issuecomment-660145941)

The numbers I got are looking almost too good. I'll post results later.
There is still a lot to do, but I want to early make sure this is worth it and I do not miss something essential.

Todos:

- [ ] Receive some more results. Assure that this change is worth it
- [ ] Merge PR in composer/semver
    - [ ] Add tests
    - [ ] Improve code
- [ ] Add tests here

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,10 +21,16 @@
             "homepage": "https://seld.be"
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "git@github.com:simonberger/semver.git"
+        }
+    ],
     "require": {
         "php": "^5.3.2 || ^7.0 || ^8.0",
         "composer/ca-bundle": "^1.0",
-        "composer/semver": "^3.0",
+        "composer/semver": "dev-exclude-conflict-test",
         "composer/spdx-licenses": "^1.2",
         "composer/xdebug-handler": "^1.1",
         "justinrainbow/json-schema": "^5.2.10",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9b94ece2895724239b36aec399e5321a",
+    "content-hash": "fd06218ced76485750b95e92051c1e9a",
     "packages": [
         {
             "name": "composer/ca-bundle",
@@ -83,16 +83,16 @@
         },
         {
             "name": "composer/semver",
-            "version": "3.2.0",
+            "version": "dev-exclude-conflict-test",
             "source": {
                 "type": "git",
-                "url": "https://github.com/composer/semver.git",
-                "reference": "da7ce661431b17a71271cdf7f5437dc722133123"
+                "url": "https://github.com/simonberger/semver.git",
+                "reference": "e924e4bbd2a27de82f7aa9a4cbe779e1838527a4"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/semver/zipball/da7ce661431b17a71271cdf7f5437dc722133123",
-                "reference": "da7ce661431b17a71271cdf7f5437dc722133123",
+                "url": "https://api.github.com/repos/simonberger/semver/zipball/e924e4bbd2a27de82f7aa9a4cbe779e1838527a4",
+                "reference": "e924e4bbd2a27de82f7aa9a4cbe779e1838527a4",
                 "shasum": ""
             },
             "require": {
@@ -113,7 +113,16 @@
                     "Composer\\Semver\\": "src"
                 }
             },
-            "notification-url": "https://packagist.org/downloads/",
+            "autoload-dev": {
+                "psr-4": {
+                    "Composer\\Semver\\": "tests"
+                }
+            },
+            "scripts": {
+                "test": [
+                    "vendor/bin/simple-phpunit"
+                ]
+            },
             "license": [
                 "MIT"
             ],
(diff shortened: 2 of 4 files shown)


simonberger:
**Results of a real project I also tested #8850 on  (uses Symfony 4, Codeception, Sonata ...

||master@bd2f321 | this PR
-- | -- | --
RAM | 1466.63MiB | 901.76MiB
Time | 14.55s | 10.10s
Analyzed packages | 16717 | 12793
Analyzed Rules | 2331184 | 1535399

**Results of only package `"drupa/core" "^8"`**

|| master@bd2f321 | this PR
-- | -- | --
RAM | 157.39MiB | 141.78MiB
Time |  1.81s | 1.61s
Analyzed packages | 4461 | 4154
Analyzed Rules | 208732 | 178519

**Results of only package `"drupa/core" "^9"`**

|| master@bd2f321 | this PR
-- | -- | --
RAM | 63.33MiB | 26.47MiB
Time |  0.95s | 0.46s
Analyzed packages | 2630 | 1511
Analyzed Rules | **51503** | **4165**

Often the required packages use no conflicts or just unimportant ones. Obviously In those cases there is no improvement but I found no situation this PR performs (notable) worse.

Seldaek:
@simonberger did you close this on purpose?

simonberger:
@Seldaek 
For the moment. Yes.
I started it early to get some feedback if it is worth it or even possible.

I now noted that the base function I created in semver is just too buggy to get test results that helps anyone at this point.
Building it is a nice challenge so I decided to finalize a robust conflict exclusion on package level before I come back to this. No matter if it can be done the way I planned it.

stof:
AFAICT, you tried applying this to *all* conflict rules. But that's broken.
A conflict rule can forbid a package from being in the pool only if you are 100% sure that a package defining that conflict is part of the result of the dependency solving. This means that only these cases can do it:

- the conflict is defined in the root package. That case is easy to match (and might even already be implemented, but I haven't checked)
- a list of packages are **all** conflicting with that package, and you are 100% sure that one of those packages will be part of the result (that could be a package name required by the root package and all versions of that package name and all versions of potential replacers would have that conflict for instance). This case is quite tricky to implement. And anyway, it cannot be implemented during package loading, as the loading can load extra packages later, which might invalidate the `all` part of the condition. See https://github.com/composer/composer/issues/8287 for an issue describing an algorithm to do a pool filtering step before running the solver. It takes care of a bunch of complex rules.

your current patch goes even further, as it even removes rules preventing to install multiple versions of the same package name at the same time (according to failing tests). You cannot remove entirely the concept of conflicts from the solver.

Also, removing all versions early from the pool will impact the explanations of solver failures, as your PR would make it report `this package was not found` instead of reporting that this package is found, but in versions that conflicts with another one. So if this early filtering is done, it requires to solve the case of reporting solver failures (changing the output to report something that don't allow understanding anymore why it fails is a bad move).

simonberger:
@stof Thank you for your comments ! While working on this I came to most problems you listed.

> A conflict rule can forbid a package from being in the pool only if you are 100% sure that a package defining that conflict is part of the result of the dependency solving.

Took me a while to understand how composer works there. I am now aware I can't get it to run in this state. After finishing the constraint exclusion function I'll try some basic conflict filtering. Root conflict filtering should be doable without a negative effect but it won't have the bigger improvement I hoped for and conflict filtering has the theoretically potential for.

> your current patch goes even further, as it even removes rules preventing to install multiple versions of the same package name at the same time (according to failing tests). You cannot remove entirely the concept of conflicts from the solver.

This last change was just a quick test that helped me to find conflicts I didn't filter correctly. I didn't really hope that can make it in the in the final version but thanks for the hint with the same names. I didn't know it has this job as well.

> Also, removing all versions early from the pool will impact the explanations of solver failures

This was already on my notes. In opposite to other problems this shouldn't be a show stopper at the end.

Seldaek:
To be honest, conflicts are so seldom used in the wild IMO that the effect regardless of implementation is unlikely to be huge, hence why I think it's probably not worth the effort.

simonberger:
That's not my experience. It is used a lot in bigger frameworks. I got even better numbers than those in the second post in tests. I know I won't reach those at the end. But it still shows that many major versions of e.g. symfony, monolog... could be dropped and with it many more packages.
On top of that those bad constructs mainly from dev packages.

Seldaek:
Eh well, I'm happy if I'm proven wrong here ;)

stof:
Well, they are used in bigger frameworks, sure. But the extra conditions necessary to make this work without breaking the dependency solving will make it much harder to ensure this provides some benefits, as they are quite hard to satisfy (and they almost imply doing some dependency solving before the solver).
In any case, the way it started in this Pr is clearly not OK (it brings a lot of performance benefits, but by breaking correctness as it does not even keep conflicts between multiple versions of the packages, which must be kept. A correct implementation will probably not be as good as the incorrect one, as it will keep work in the solver).

stof:
not even counting the extra complexity that will be necessary to rebuild the problem explanation when the solver won't know anymore about parts of the problem at the time it provides its explanation (not even sure that's doable without redoing a second non-filtering solving).

PR #9232, opened 2020-09-20
Removed deprecated code from 1.x

simonberger:
Let me know if you want an entry in upgrade notes. For example PackageEvent and CommandEvent are not in.


Seldaek:
Those have been deprecated for ages so I think it's ok really.

PR #9247, opened 2020-09-26
Access error fixed.

selcukcukur:
Since the **"autoload_static.php"** element tries to access, it would be better to use these elements publicly.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -43,16 +43,16 @@ namespace Composer\Autoload;
 class ClassLoader
 {
     // PSR-4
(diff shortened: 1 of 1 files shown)


staabm:
Which access errors does this PR fix? How to reproduce?

selcukcukur:
ClassLoader is being referenced for variables "$prefixLengthsPsr4", "$prefixDirsPsr4", "$prefixesPsr0", "$classMap". Cannot see if it is, because the variable is "private" defined as. This fixes it from showing the error.

In addition :
https://i.hizliresim.com/k4LeMV.png

staabm:
Accessing private properties from within the same file is perfectly valid.

Seems your IDE is wrong

PR #9250, opened 2020-09-29
Document authentication for private packages

PrinsFrank:
Currently, only authentication using http basic is documented. That is not the only authentication method that can be used to authenticate against a private package server. This PR tries to fix that and explain some more generic structure on how authentication works.


doc/05-repositories.md
```diff
@@ -507,7 +507,7 @@ package repository definitions. It will fetch all the packages that are
 `require`d and dump a `packages.json` that is your `composer` repository.
 
 Check [the satis GitHub repository](https://github.com/composer/satis) and
-the [Satis article](articles/handling-private-packages-with-satis.md) for more
```
PrinsFrank, line 510: This file was renamed because it doesn't specifically relate to satis anymore in it's current form.

doc/articles/authentication-for-private-packages.md
```diff
@@ -0,0 +1,226 @@
...
+            "type": "composer",
+            "url": "https://username:password@example.org"
+        }
+    ]
+}
+```
+
+## Custom token authentication
+
+### Manual custom token authentication
```
PrinsFrank, line 154: I couldn't figure out the command line syntax for this. any ideas? Can also be added in the future when the next person contributes.

doc/05-repositories.md
```diff
@@ -507,7 +507,7 @@ package repository definitions. It will fetch all the packages that are
...
+the [Satis article](articles/handling-private-packages.md) for more
```
stof, line 510: the wording `the Satis article` is not correct anymore either.

stof: @Seldaek do we have a way to create doc redirections for the composer website to avoid breaking existing links ?

doc/articles/authentication-for-private-packages.md
```diff
@@ -0,0 +1,226 @@
...
+
+Sometimes automatic authentication is not possible, or you may want to predefine
+authentication credentials.
+
+Credentials can be stored on 3 different places; in an auth.json for the project, a global
+auth.json or in the composer.json itself.
+
+## Authentication in auth.json per project
+
+In this authentication storage method, an 'auth.json' file will be present in the same folder
```
stof, line 35: should be `` `auth.json` `` rather than `'auth.json'`

doc/articles/authentication-for-private-packages.md
```diff
@@ -0,0 +1,226 @@
...
+You can open this file in your favorite editor and fix the error.
+
+## Authentication in composer.json file itself
+
+> **Note:** **This is not recommended** as these credentials are visible
+> to anyone who has access to the composer.json, either when it is shared through
+> a version control system like git or when an attacker gains (read) access to
+> your production server files.
+
+It is also possible to add credentials to a composer.json on a per-project basis in the 'config'
```
stof, line 93: should also use backticks rather than apostrophs here

doc/05-repositories.md
PrinsFrank: @stof I renamed the link, but will wait for an answer about avoiding breaking existing links before resolving this conversation.

PrinsFrank: Should I undo the filename change for now and create a seperate PR for it?

PrinsFrank: Reverted all page renames/deletions so all old links will keep working.

PrinsFrank:
@stof I split off the page rename to #9267 so we can discuss this there individually. Is the current content like this ok?

doc/articles/authentication-for-private-packages.md
```diff
@@ -0,0 +1,226 @@
...
+# Authentication for privately hosted packages
+
+Your [private package server](handling-private-packages-with-satis.md) is probably secured with one
+or more authentication options. In order to allow your project to have access to these
+packages you will have to tell Composer how to authenticate with the server that hosts the
+package(s).
+
+# Authentication principles
+
+Whenever composer encounters a protected composer repository it will try to authenticate
```
Seldaek, line 14: Composer should be capitalized when not refering to the cli command or composer.json

PrinsFrank:
Thanks, will do!

PR #9257, opened 2020-10-01
With-dependencies alias

### LINKED ISSUE #9213: Can we have an alias for --with-dependencies
My `composer.json`:

```json
Not Required
```

```
Not Required
```

When I run this command:

```
composer update Repo/Package --with-dependencies
```

I get the following output:

```
I tend to mistype the word dependencies
```

I expect to have an alias like `-wd` which is much smaller and there are less chances of mistake.

I know its not just me, whose first language is not English. Many other devs including my team members / friends are facing this issue.

flavioheleno:
This PR adds aliases for `--with-dependencies` (`-w`) and `--with-all-dependencies` (`-W`).

Closes #9213.


stof:
Maybe `-d` and `-D` would be better (referring to the word `dependencies` rather than `with`) if they are available?

stof:
OK, answering my own comment: `-d` is already the shortcut for the global `--working-dir` option of Composer...

flavioheleno:
@stof I've followed @Seldaek's suggestion from #9213 :-)

stof:
yeah, he probably remembered from the start that `-d` was already used :smile:

Seldaek:
Thanks

PR #9258, opened 2020-10-02
Fixed description of StatusCommand: It do list all changes in vendor, not only for "source" ones

lyrixx:
I tested it, and even with "dist" packages, the status command is able to find modified vendor 
(And that's amazing, thanks)


Seldaek:
True that was added later :) Thanks

Issue #6621, opened 2017-08-21
GitHub rate limit exceed when token provided


sadortun:
When a Github token is provided, and valid, BUT the rate limit is exceed. Composer ask to create a new token, which is very miscleading.

How to reproduce:
1.  Perform 5000 API requests to Github within one hour
2. Verify that https://api.github.com/rate_limit?access_token=XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX   `    "remaining": 0,`
3.  Run `composer update`

My `composer.json`:

```json
{
    "name": "symfony/symfony",
    "type": "library",
    "description": "The Symfony PHP framework",
    "keywords": ["framework"],
    "homepage": "https://symfony.com",
    "license": "MIT",
    "authors": [
        {
            "name": "Fabien Potencier",
            "email": "fabien@symfony.com"
        },
        {
            "name": "Symfony Community",
            "homepage": "https://symfony.com/contributors"
        }
    ],
    "require": {
        "php": "^7.1.3",
        "ext-xml": "*",
        "doctrine/common": "~2.4",
        "fig/link-util": "^1.0",
        "twig/twig": "~1.34|~2.4",
        "psr/cache": "~1.0",
        "psr/container": "^1.0",
        "psr/link": "^1.0",
        "psr/log": "~1.0",
        "psr/simple-cache": "^1.0",
        "symfony/polyfill-intl-icu": "~1.0",
        "symfony/polyfill-mbstring": "~1.0"
    },
    "replace": {
        "symfony/asset": "self.version",
        "symfony/browser-kit": "self.version",
        "symfony/cache": "self.version",
        "symfony/config": "self.version",
        "symfony/console": "self.version",
        "symfony/css-selector": "self.version",
        "symfony/dependency-injection": "self.version",
        "symfony/debug": "self.version",
        "symfony/debug-bundle": "self.version",
        "symfony/doctrine-bridge": "self.version",
        "symfony/dom-crawler": "self.version",
        "symfony/dotenv": "self.version",
        "symfony/event-dispatcher": "self.version",
        "symfony/expression-language": "self.version",
        "symfony/filesystem": "self.version",
        "symfony/finder": "self.version",
        "symfony/form": "self.version",
        "symfony/framework-bundle": "self.version",
        "symfony/http-foundation": "self.version",
        "symfony/http-kernel": "self.version",
        "symfony/inflector": "self.version",
        "symfony/intl": "self.version",
        "symfony/ldap": "self.version",
[...]

PR #9261, opened 2020-10-02
Implemented PoolOptimizer

Toflar:
This PR optimizes the `Pool` before the rules are created and passed on to the `Solver`. Every package added to the pool results in an exponential increase of the number of rules the `Solver` has to analyze so the best we can do is to try to throw away all the packages that aren't of any interest early on.

I've implemented a new `PoolOptimizer` and set up a test framework to which we can add more optimizations in the future but the current implementation already provides **huge** benefits!

It works by early selecting the correct package based on the provided policy. It's common to have loads of bugfix releases that all have the same dependency definition. If you imagine e.g. `symfony/routing` in its `4.4` LTS branch, all of them hardly ever define new `require`, `replace`, `provide` or `conflict` definitions. Chances that `4.4.0` and `4.4.14` are the same, are really high and so we don't need to keep all of the package versions in between, we can drop the irrelevant ones depending on `--prefer-stable` and `--prefer-lowest`. There's a bit more to it and it's not just that simple but that's the basic idea behind the first optimization in the new `PoolOptimizer`.

And here are the **very promising** results:

Contao Managed Edition (https://github.com/contao/managed-edition/blob/master/composer.json):

|  | 1.10.13 | master@5df1797d2 | this PR |
|------|----------:|-----------------:|--------:|
| RAM | 3351.08MiB | 1518.31MiB | 134.54MiB |
| Time | 84s | 28.95s | 3.25s |

Magento (https://github.com/magento/magento2/blob/2.4-develop/composer.json):

|  | 1.10.13 | master@5df1797d2 | this PR |
|------|----------:|-----------------:|--------:|
| RAM | 961.86MiB | 169.64MiB | 46.39MiB |
| Time | 24.61s | 4.38s | 1.17s |

Laravel 7.x (https://github.com/laravel/laravel/blob/7.x/composer.json):

|  | 1.10.13 | master@5df1797d2 | this PR |
|------|----------:|-----------------:|--------:|
| RAM | 911.97MiB | 152.42MiB | 37.62MiB |
| Time | 8.81s | 1.51s | 0.95s |

Packagist (https://github.com/composer/packagist/blob/master/composer.json)

|  | 1.10.13 | master@5df1797d2 | this PR |
|------|----------:|-----------------:|--------:|
| RAM | 3308.18MiB | 1503.63MiB | 134.73M |
| Time | 51.63s | 14.9s | 4.04s |

Of course, the more complex the dependencies (or just the more dependencies in general) the higher the benefits of this PR.

I was too lazy to calculate the benefits in % but I guess we can summarize it like so: 🚀 

I'm off for a few days now so I might not be replying to questions quickly but would love to see others giving it a try and share some numbers :) 

For the lazy folks: Here's a precompiled phar (c996457a039b2f0a5eee3b6364dfce307c614737):

[composer.phar.zip](https://github.com/composer/composer/files/5484474/composer.phar.zip)

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -353,3 +353,27 @@ See also https://github.com/composer/composer/issues/4180 for more information.
 Composer can unpack zipballs using either a system-provided `unzip` or `7z` (7-Zip) utility, or PHP's
 native `ZipArchive` class. On OSes where ZIP files can contain permissions and symlinks, we recommend
 installing `unzip` or `7z` as these features are not supported by `ZipArchive`.
+
+
+## Disabling the pool optimizer
+
+In Composer, the `Pool` class contains all the packages that are relevant for the dependency
+resolving process. That is what is used to generate all the rules which are then
+passed on to the dependency solver.
+In order to improve performance, Composer tries to optimize this `Pool` by removing useless
+package information early on.
+
+If all goes well, you should never notice any issues with it but in case you run into
+an unexpected result such as an unresolvable set of dependencies or conflicts where you
+think Composer is wrong, you might want to disable the optimizer by using the environment
+variable `COMPOSER_POOL_OPTIMIZER` and run the update again like so:
+
+```bash
+COMPOSER_POOL_OPTIMIZER=0 php composer.phar update
+```
+
+Now double check if the result is still the same. It will take significantly longer and use
+a lot more memory to run the dependency resolving process.
+
+If the result is different, you likely hit a problem in the pool optimizer.
+Please [report this issue](https://github.com/composer/composer/issues) so it can be fixed.
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -36,16 +36,57 @@ class Pool implements \Countable
     protected $providerCache = array();
     /** @var BasePackage[] */
     protected $unacceptableFixedOrLockedPackages;
+    /** @var array<string, array<string, string>> Map of package name => normalized version => pretty version */
+    protected $removedVersions = array();
+    /** @var array<string, array<string, string>> Map of package object hash => removed normalized versions => removed pretty version */
+    protected $removedVersionsByPackage = array();
 
     /**
      * @param BasePackage[] $packages
      * @param BasePackage[] $unacceptableFixedOrLockedPackages
+     * @param array<string, array<string, string>> $removedVersions
+     * @param array<string, array<string, string>> $removedVersionsByPackage
      */
-    public function __construct(array $packages = array(), array $unacceptableFixedOrLockedPackages = array())
+    public function __construct(array $packages = array(), array $unacceptableFixedOrLockedPackages = array(), array $removedVersions = array(), array $removedVersionsByPackage = array())
     {
         $this->versionParser = new VersionParser;
         $this->setPackages($packages);
         $this->unacceptableFixedOrLockedPackages = $unacceptableFixedOrLockedPackages;
+        $this->removedVersions = $removedVersions;
+        $this->removedVersionsByPackage = $removedVersionsByPackage;
+    }
+
+    /**
+     * @param  string $name
+     * @return array<string, string>
+     */
+    public function getRemovedVersions($name, ConstraintInterface $constraint)
+    {
+        if (!isset($this->removedVersions[$name])) {
+            return array();
+        }
+
+        $result = array();
+        foreach ($this->removedVersions[$name] as $version => $prettyVersion) {
+            if ($constraint->matches(new Constraint('==', $version))) {
+                $result[$version] = $prettyVersion;
+            }
+        }
+
+        return $result;
+    }
+
+    /**
+     * @param  string $objectHash
+     * @return array<string, string>
+     */
+    public function getRemovedVersionsByPackage($objectHash)
+    {
+        if (!isset($this->removedVersionsByPackage[$objectHash])) {
+            return array();
+        }
+
+        return $this->removedVersionsByPackage[$objectHash];
     }
 
     /**
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -221,6 +221,14 @@ class Pool implements \Countable
         return \in_array($package, $this->unacceptableFixedOrLockedPackages, true);
     }
 
+    /**
+     * @return BasePackage[]
+     */
+    public function getUnacceptableFixedOrLockedPackages()
+    {
+        return $this->unacceptableFixedOrLockedPackages;
+    }
+
     public function __toString()
     {
         $str = "Pool:\n";
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -62,6 +62,10 @@ class PoolBuilder
      */
     private $eventDispatcher;
     /**
+     * @var PoolOptimizer|null
+     */
+    private $poolOptimizer;
+    /**
      * @var IOInterface
      */
     private $io;
@@ -128,13 +132,14 @@ class PoolBuilder
      * @param string[] $rootReferences an array of package name => source reference
      * @phpstan-param array<string, string> $rootReferences
      */
-    public function __construct(array $acceptableStabilities, array $stabilityFlags, array $rootAliases, array $rootReferences, IOInterface $io, EventDispatcher $eventDispatcher = null)
+    public function __construct(array $acceptableStabilities, array $stabilityFlags, array $rootAliases, array $rootReferences, IOInterface $io, EventDispatcher $eventDispatcher = null, PoolOptimizer $poolOptimizer = null)
     {
         $this->acceptableStabilities = $acceptableStabilities;
         $this->stabilityFlags = $stabilityFlags;
         $this->rootAliases = $rootAliases;
         $this->rootReferences = $rootReferences;
         $this->eventDispatcher = $eventDispatcher;
+        $this->poolOptimizer = $poolOptimizer;
         $this->io = $io;
     }
 
@@ -258,6 +263,8 @@ class PoolBuilder
         $this->maxExtendedReqs = array();
         $this->skippedLoad = array();
         $this->indexCounter = 0;
+
+        $pool = $this->runOptimizer($request, $pool);
 
         Intervals::clear();
 
@@ -571,5 +578,34 @@ class PoolBuilder
             }
             unset($this->aliasMap[spl_object_hash($package)]);
         }
+    }
+
+    /**
+     * @return Pool
+     */
+    private function runOptimizer(Request $request, Pool $pool)
+    {
+        if (null === $this->poolOptimizer) {
+            return $pool;
+        }
+
+        $total = \count($pool->getPackages());
+
+        $pool = $this->poolOptimizer->optimize($request, $pool);
+
+        $filtered = $total - \count($pool->getPackages());
+
+        if (0 === $filtered) {
+            return $pool;
+        }
+
+        $this->io->write(sprintf(
+            '<info>Found %s package versions referenced in your dependency graph. %s (%d%%) were optimized away.</info>',
+            number_format($total),
+            number_format($filtered),
+            round(100/$total*$filtered)
+        ), true, IOInterface::VERY_VERBOSE);
+
+        return $pool;
     }
 }
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -14,6 +14,7 @@ namespace Composer\DependencyResolver;
 
 use Composer\Package\AliasPackage;
 use Composer\Package\BasePackage;
+use Composer\Package\Version\VersionParser;
 use Composer\Semver\CompilingMatcher;
 use Composer\Semver\Constraint\ConstraintInterface;
 use Composer\Semver\Constraint\Constraint;
@@ -57,6 +58,11 @@ class PoolOptimizer
      */
     private $aliasesPerPackage = array();
 
+    /**
+     * @var array<string, array<string, string>>
+     */
+    private $removedVersionsByPackage = array();
+
     public function __construct(PolicyInterface $policy)
     {
         $this->policy = $policy;
@@ -81,6 +87,7 @@ class PoolOptimizer
         $this->conflictConstraintsPerPackage = array();
         $this->packagesToRemove = array();
         $this->aliasesPerPackage = array();
+        $this->removedVersionsByPackage = array();
 
         return $optimizedPool;
     }
@@ -165,13 +172,16 @@ class PoolOptimizer
     private function applyRemovalsToPool(Pool $pool)
     {
         $packages = array();
+        $removedVersions = array();
         foreach ($pool->getPackages() as $package) {
             if (!isset($this->packagesToRemove[$package->id])) {
                 $packages[] = $package;
+            } else {
+                $removedVersions[$package->getName()][$package->getVersion()] = $package->getPrettyVersion();
             }
         }
 
-        $optimizedPool = new Pool($packages, $pool->getUnacceptableFixedOrLockedPackages());
+        $optimizedPool = new Pool($packages, $pool->getUnacceptableFixedOrLockedPackages(), $removedVersions, $this->removedVersionsByPackage);
 
         // Reset package removals
         $this->packagesToRemove = array();
@@ -184,8 +194,8 @@ class PoolOptimizer
      */
     private function optimizeByIdenticalDependencies(Pool $pool)
     {
-        $identicalDefinitionPerPackage = array();
-        $packageIdsToRemove = array();
+        $identicalDefinitionsPerPackage = array();
+        $packageIdenticalDefinitionLookup = array();
 
         foreach ($pool->getPackages() as $package) {
 
@@ -195,7 +205,7 @@ class PoolOptimizer
                 continue;
             }
 
-            $packageIdsToRemove[$package->id] = true;
+            $this->markPackageForRemoval($package->id);
 
             $dependencyHash = $this->calculateDependencyHash($package);
 
@@ -206,7 +216,6 @@ class PoolOptimizer
                 }
 
                 foreach ($this->requireConstraintsPerPackage[$packageName] as $requireConstraint) {
-
                     $groupHashParts = array();
 
                     if (CompilingMatcher::match($requireConstraint, Constraint::OP_EQ, $package->getVersion())) {
@@ -234,32 +243,19 @@ class PoolOptimizer
                         continue;
                     }
 
-                    $identicalDefinitionPerPackage[$packageName][implode('', $groupHashParts)][$dependencyHash][] = $package;
+                    $groupHash = implode('', $groupHashParts);
+                    $identicalDefinitionsPerPackage[$packageName][$groupHash][$dependencyHash][] = $package;
+                    $packageIdenticalDefinitionLookup[$package->id][$packageName] = array('groupHash' => $groupHash, 'dependencyHash' => $dependencyHash);
                 }
             }
         }
 
-        $keepPackage = function (BasePackage $package, $aliasesPerPackage) use (&$packageIdsToRemove, &$keepPackage) {
-            unset($packageIdsToRemove[$package->id]);
-            if ($package instanceof AliasPackage) {
-                // recursing here so aliasesPerPackage for the aliasOf can be checked
-                // and all its aliases marked to be kept as well
-                $keepPackage($package->getAliasOf(), $aliasesPerPackage);
-            }
-            if (isset($aliasesPerPackage[$package->id])) {
-                foreach ($aliasesPerPackage[$package->id] as $aliasPackage) {
-                    unset($packageIdsToRemove[$aliasPackage->id]);
-                }
-            }
-        };
-
-        foreach ($identicalDefinitionPerPackage as $package => $constraintGroups) {
+        foreach ($identicalDefinitionsPerPackage as $constraintGroups) {
             foreach ($constraintGroups as $constraintGroup) {
-                foreach ($constraintGroup as $hash => $packages) {
-
+                foreach ($constraintGroup as $packages) {
                     // Only one package in this constraint group has the same requirements, we're not allowed to remove that package
                     if (1 === \count($packages)) {
-                        $keepPackage($packages[0], $this->aliasesPerPackage);
+                        $this->keepPackage($packages[0], $identicalDefinitionsPerPackage, $packageIdenticalDefinitionLookup);
                         continue;
                     }
 
@@ -272,17 +268,12 @@ class PoolOptimizer
                     }
 
                     foreach ($this->policy->selectPreferredPackages($pool, $literals) as $preferredLiteral) {
-                        $keepPackage($pool->literalToPackage($preferredLiteral), $this->aliasesPerPackage);
+                        $this->keepPackage($pool->literalToPackage($preferredLiteral), $identicalDefinitionsPerPackage, $packageIdenticalDefinitionLookup);
                     }
                 }
             }
         }
 
-        foreach ($packageIdsToRemove as $id => $dummy) {
-            $this->markPackageForRemoval($id);
-        }
-
-        // Apply removals
         return $this->applyRemovalsToPool($pool);
     }
 
@@ -341,5 +332,55 @@ class PoolOptimizer
         }
 
         $this->packagesToRemove[$id] = true;
+    }
+
+    /**
+     * @param array<string, array<string, array<string, list<BasePackage>>>> $identicalDefinitionsPerPackage
+     * @param array<int, array<string, array{groupHash: string, dependencyHash: string}>> $packageIdenticalDefinitionLookup
+     * @return void
+     */
+    private function keepPackage(BasePackage $package, $identicalDefinitionsPerPackage, $packageIdenticalDefinitionLookup)
+    {
+        unset($this->packagesToRemove[$package->id]);
+
+        if ($package instanceof AliasPackage) {
+            // recursing here so aliasesPerPackage for the aliasOf can be checked
+            // and all its aliases marked to be kept as well
+            $this->keepPackage($package->getAliasOf(), $identicalDefinitionsPerPackage, $packageIdenticalDefinitionLookup);
+        }
+
+        // record all the versions of the package group so we can list them later in Problem output
+        foreach ($package->getNames(false) as $name) {
+            if (isset($packageIdenticalDefinitionLookup[$package->id][$name])) {
+                $packageGroupPointers = $packageIdenticalDefinitionLookup[$package->id][$name];
+                $packageGroup = $identicalDefinitionsPerPackage[$name][$packageGroupPointers['groupHash']][$packageGroupPointers['dependencyHash']];
+                foreach ($packageGroup as $pkg) {
+                    if ($pkg instanceof AliasPackage && $pkg->getPrettyVersion() === VersionParser::DEFAULT_BRANCH_ALIAS) {
+                        $pkg = $pkg->getAliasOf();
+                    }
+                    $this->removedVersionsByPackage[spl_object_hash($package)][$pkg->getVersion()] = $pkg->getPrettyVersion();
+                }
+            }
+        }
+
+        if (isset($this->aliasesPerPackage[$package->id])) {
+            foreach ($this->aliasesPerPackage[$package->id] as $aliasPackage) {
+                unset($this->packagesToRemove[$aliasPackage->id]);
+
+                // record all the versions of the package group so we can list them later in Problem output
+                foreach ($aliasPackage->getNames(false) as $name) {
+                    if (isset($packageIdenticalDefinitionLookup[$aliasPackage->id][$name])) {
+                        $packageGroupPointers = $packageIdenticalDefinitionLookup[$aliasPackage->id][$name];
+                        $packageGroup = $identicalDefinitionsPerPackage[$name][$packageGroupPointers['groupHash']][$packageGroupPointers['dependencyHash']];
+                        foreach ($packageGroup as $pkg) {
+                            if ($pkg instanceof AliasPackage && $pkg->getPrettyVersion() === VersionParser::DEFAULT_BRANCH_ALIAS) {
+                                $pkg = $pkg->getAliasOf();
+                            }
+                            $this->removedVersionsByPackage[spl_object_hash($aliasPackage)][$pkg->getVersion()] = $pkg->getPrettyVersion();
+                        }
+                    }
+                }
+            }
+        }
     }
 }
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -1,0 +1,345 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\DependencyResolver;
+
+use Composer\Package\AliasPackage;
+use Composer\Package\BasePackage;
+use Composer\Semver\CompilingMatcher;
+use Composer\Semver\Constraint\ConstraintInterface;
+use Composer\Semver\Constraint\Constraint;
+use Composer\Semver\Constraint\MultiConstraint;
+use Composer\Semver\Intervals;
+
+/**
+ * Optimizes a given pool
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class PoolOptimizer
+{
+    /**
+     * @var PolicyInterface
+     */
+    private $policy;
+
+    /**
+     * @var array<int, true>
+     */
+    private $irremovablePackages = array();
+
+    /**
+     * @var array<string, array<string, ConstraintInterface>>
+     */
+    private $requireConstraintsPerPackage = array();
+
+    /**
+     * @var array<string, array<string, ConstraintInterface>>
+     */
+    private $conflictConstraintsPerPackage = array();
+
+    /**
+     * @var array<int, true>
+     */
+    private $packagesToRemove = array();
+
+    /**
+     * @var array<int, BasePackage[]>
+     */
+    private $aliasesPerPackage = array();
+
+    public function __construct(PolicyInterface $policy)
+    {
+        $this->policy = $policy;
+    }
+
+    /**
+     * @return Pool
+     */
+    public function optimize(Request $request, Pool $pool)
+    {
+        $this->prepare($request, $pool);
+
+        $optimizedPool = $this->optimizeByIdenticalDependencies($pool);
+
+        // No need to run this recursively at the moment
+        // because the current optimizations cannot provide
+        // even more gains when ran again. Might change
+        // in the future with additional optimizations.
+
+        $this->irremovablePackages = array();
+        $this->requireConstraintsPerPackage = array();
+        $this->conflictConstraintsPerPackage = array();
+        $this->packagesToRemove = array();
+        $this->aliasesPerPackage = array();
+
+        return $optimizedPool;
+    }
+
+    /**
+     * @return void
+     */
+    private function prepare(Request $request, Pool $pool)
+    {
+        $irremovablePackageConstraintGroups = array();
+
+        // Mark fixed or locked packages as irremovable
+        foreach ($request->getFixedOrLockedPackages() as $package) {
+            $irremovablePackageConstraintGroups[$package->getName()][] = new Constraint('==', $package->getVersion());
+        }
+
+        // Extract requested package requirements
+        foreach ($request->getRequires() as $require => $constraint) {
+            $constraint = Intervals::compactConstraint($constraint);
+            $this->requireConstraintsPerPackage[$require][(string) $constraint] = $constraint;
+        }
+
+        // First pass over all packages to extract information and mark package constraints irremovable
+        foreach ($pool->getPackages() as $package) {
+            // Extract package requirements
+            foreach ($package->getRequires() as $link) {
+                $constraint = Intervals::compactConstraint($link->getConstraint());
+                $this->requireConstraintsPerPackage[$link->getTarget()][(string) $constraint] = $constraint;
+            }
+            // Extract package conflicts
+            foreach ($package->getConflicts() as $link) {
+                $constraint = Intervals::compactConstraint($link->getConstraint());
+                $this->conflictConstraintsPerPackage[$link->getTarget()][(string) $constraint] = $constraint;
+            }
+
+            // Keep track of alias packages for every package so if either the alias or aliased is kept
+            // we keep the others as they are a unit of packages really
+            if ($package instanceof AliasPackage) {
+                $this->aliasesPerPackage[$package->getAliasOf()->id][] = $package;
+            }
+        }
+
+        $irremovablePackageConstraints = array();
+        foreach ($irremovablePackageConstraintGroups as $packageName => $constraints) {
+            $irremovablePackageConstraints[$packageName] = 1 === \count($constraints) ? $constraints[0] : new MultiConstraint($constraints, false);
+        }
+        unset($irremovablePackageConstraintGroups);
+
+        // Mark the packages as irremovable based on the constraints
+        foreach ($pool->getPackages() as $package) {
+            if (!isset($irremovablePackageConstraints[$package->getName()])) {
+                continue;
+            }
+
+            if (CompilingMatcher::match($irremovablePackageConstraints[$package->getName()], Constraint::OP_EQ, $package->getVersion())) {
+                $this->markPackageIrremovable($package);
+            }
+        }
+    }
+
+    /**
+     * @return void
+     */
+    private function markPackageIrremovable(BasePackage $package)
+    {
+        $this->irremovablePackages[$package->id] = true;
+        if ($package instanceof AliasPackage) {
+            // recursing here so aliasesPerPackage for the aliasOf can be checked
+            // and all its aliases marked as irremovable as well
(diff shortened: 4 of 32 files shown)


stof:
My spec in https://github.com/composer/composer/issues/8287 includes details about how conflict rules impact this optimization. But your implementation ignores them entirely, which might lead to breaking the dependency resolution (imagine that the latest version of a package adds a conflict rule forbidding only the latest release of the other one but allowing the previous patch release for instance)

stof:
regarding the time where the pool optimizer is called, I'm not sure it is the best one compared to the `PRE_POOL_CREATE` event triggered for plugins. If plugins are modifying the pool packages, that might break the invariants of the optimization if it is already done.
I think it would be safer to optimize the pool just before entering the solver, but after plugins have tweaked it.

Toflar:
> My spec in https://github.com/composer/composer/issues/8287 includes details about how conflict rules impact this optimization. But your implementation ignores them entirely, which might lead to breaking the dependency resolution (imagine that the latest version of a package adds a conflict rule forbidding only the latest release of the other one but allowing the previous patch release for instance)

It would get a different dependency hash and not be removed from the pool. They are actually not ignored at all :)

stof:
ah sorry, I missed that part about the hash.

stof:
@Toflar idea to ensure that the PoolOptimizer has no negative effect on the dependency solving: run the solver functional tests with and without the PoolOptimizer:

- all successful tests MUST have the same outcome in both cases
- tests reporting a dependency solving error MUST do it in both cases, but the exact output might change (as that output depends on versions that the solver tried, and it will try less of them with the optimization). Fixture files must be written for the optimized pool, as that's what users will see.

What do you think ?

stof:
Should we have some way (CLI flag, env variable, magic incantation, etc...) to disable the PoolOptimizer when running composer, to help identify cases where a dependency solving error is caused by a bug of the optimization (if solving finds a solution without the PoolOptimizer, there is a bug which must then be reported and fixed in Composer) ?
Of course, users should never *need* to disable it. But we are all human (at least I think so) and bugs happen. Being able to quickly know that the optimizer is breaking things would help.
And any such case would then enhance the testsuite forever if we implement the idea I gave in my previous comment.

hussainweb:
I just wanted to share results of my tests with Drupal. Before each test, I removed the `vendor`, `web/core` (Drupal core) directories, and the `composer.lock` file.

A not-very-complex Drupal 9.1 dev project ([this repo](https://github.com/hussainweb/drupal-php8-composer2))

-|Composer 1.10.3 | Composer 2.0 RC1 | This PR (phar file)
-|----------------|------------------|--------------------
Memory|482.7MiB (peak: 2015.39MiB)|60.74MiB (peak: 699.45MiB)|34.28MiB (peak: 60.42MiB)
Time|136.12s|36.06s|31.43s

[Drupal 9.1 core (dev release)](https://git.drupalcode.org/project/drupal)

-|Composer 1.10.3 | Composer 2.0 RC1 | This PR (phar file)
-|----------------|------------------|--------------------
Memory|419.74MiB (peak: 1005.93MiB)|31.03MiB (peak: 214.35MiB)|28MiB (peak: 38.14MiB)
Time|24.77s|13.06s|10.61s

I can see the time improvement is not significant but memory improvement is. This is awesome work. Thanks! 👏

hussainweb:
I also tried this on a project with Drupal 8; this time running `composer update --profile`. I ran it once before the actual tests so that caches get created, etc. This project had many out-of-date dependencies so there were many packages that got updated. I didn't clear out `vendor` directory for each run but before each run, I restored `composer.lock` to the previous state.

-|Composer 1.10.13 | Composer 2.0 RC1 | This PR (phar file)
-|----------------|------------------|--------------------
Memory|625.66MiB (peak: 2234.86MiB)|133.01MiB (peak: 785.85MiB)|112.02MiB (peak: 158.98MiB)
Time|43.57s|23.27s|21.38s

Again, not a significant differnce in time (this could be because it actually times updating packages on disk) but a significant reduction in peak memory.

wouterj:
A Symfony application using the Symfony Flex plugin (needed because our Symfony packages use `*` as version constraint):

| - | Composer 1.10.13 | Composer 2.0 RC1 | This PR (phar file) |
| --- | --- | --- | --- |
| Memory | 303.22MiB (peak: 859.14MiB) | 32.69MiB (peak: 79.3MiB) | 32.92MiB (peak: 80.23MiB) |
| Time |12.62s | 8.29s | time: 8.45s |

Not much changes, which I think it expected as the Symfony Flex plugin already significantly reduces the versions that are considered.

---

A Neos base distribution - the most complex project on my PC that isn't requiring Symfony Flex:

| - | Composer 1.10.13 | Composer 2.0 RC1 | This PR (phar file) |
| --- | --- | --- | --- |
| Memory | 413.34MiB (peak: 1321.35MiB) | 46.11MiB (peak: 275.17MiB) | 25.32MiB (peak: 98.34MiB) |
| Time | 20.56s | 11.86s | 9.24s |

Similar results as Drupal 8, not a significant step in performance but a significant reduction of peak memory usage.

simonberger:
Good job man ! I tested this two days ago with almost same runtime. Now this PR is a lot faster. No idea why :)
Another thing is the memory usage. Really impressive. Improvements in this area in my opinion not less important than the runtime.

|| master@5df1797d2 | this PR
 -- | -- | --
RAM | 1376.52MiB | 85.65MiB
Time | 12.47s | 4.4s

effulgentsia:
I tested a different Drupal scenario than the two from @hussainweb above, and got less dramatic results, but still a very nice improvement.

My interest was for the use-case of doing a patch-level Drupal core update when all other dependencies are already up to date. This is because we're working on an initiative to add an auto-updater to Drupal, where this scenario will be run on production, including on cheap shared hosting plans with low memory limits.

Setup: begin with a Drupal installation that's on a recently outdated patch-level version (latest Drupal version is 9.0.6). Also include some non-core dependencies (lightning_media brings in a bunch).
```
> composer create-project drupal/recommended-project drupal 9.0.5
> cd drupal
> composer config repositories.assets composer https://asset-packagist.org
> composer require drupal/lightning_media
```

Test: profile an update to the latest Drupal core version. Start with a cold cache, since that's the likely scenario on shared hosting by the time this would be run.
```
> composer clearcache
> composer update drupal/core-recommended --with-all-dependencies --profile
```

Result:
|| Composer 2.0-RC1  | This PR |
|| --- | --- |
|RAM| 101.02MiB (peak: 173.81MiB)  | 102.67MiB (peak: 116.44MiB)  |
|Time| 96.16s  | 86.61s  |

While not an order of magnitude difference, that's still an excellent reduction in peak memory usage, which can make a big difference on some cheap hosting plans.

Toflar:
@stof regarding the test suite: We already have loads of integration tests in the `InstallerTest` which test from the actual `composer update` command to the expected lock and output. Those tests are all green. It took me a while to get them there and you can actually see I had to adjust one of them 😉 I agree that the more tests, the better. But I guess they'll come along once we stumble upon new constellations which haven't been tested yet. We can also isolate tests specifically to the `PoolOptimizer`, that's why I've implemented the foundation for additional tests in this PR and also provided a few basic tests.

> Should we have some way (CLI flag, env variable, magic incantation, etc...) to disable the PoolOptimizer when running composer, to help identify cases where a dependency solving error is caused by a bug of the optimization (if solving finds a solution without the PoolOptimizer, there is a bug which must then be reported and fixed in Composer) ?

Yes, totally. I've implemented and documented the new `--no-optimizer` flag in 96dcbc9fc2cd657e5c43d9550c7df688008c155f 😎 

To everybody: Thanks a lot for the numbers, they look really promising which makes me very happy 😊
Just a quick note regarding the performance tests: This PR does not have any influence on downloading or unpacking etc. but is specifically designed to improve the resolving process. So there's no point in running comparisons by clearing the Composer cache first. You should run `composer update` first, to make sure the package information cache is primed. Then run `COMPOSER_DISABLE_NETWORK=1 composer update --profile` to make sure there are no requests dispatched. This should work without any problems if you primed the cache correctly in the first place.

stof:
> Those tests are all green. It took me a while to get them there and you can actually see I had to adjust one of them wink I agree that the more tests, the better. But I guess they'll come along once we stumble upon new constellations which haven't been tested yet.

The fact that problem reporting is impacted by the pool optimization is precisely the reason why my proposal is that this "no-optimizer` run disables the assertion on the output in case of reported problems (keeping only an assertion that a problem is indeed reported).
Such double run would allow us to guarantee that the optimized pool yields the same outcome than the non optimized pool (and so that the optimization is safe).

Toflar:
The output can't be the same though. In case of an unresolvable set of dependencies, the list of problems is a lot shorter if not all possible problems are listed but just the ones of the optimized pool (which BTW also improves readability a lot :)).

stof:
> The output can't be the same though. In case of an unresolvable set of dependencies, the list of problems is a lot shorter if not all possible problems are listed but just the ones of the optimized pool (which BTW also improves readability a lot :)).

I know that. That's why I'm saying that for problems, the non-optimized run must NOT assert the output, but only the fact that the run also triggers some problems (to make sure that the dependency solver is not successful when disabling optimization).

GrahamCampbell:
I'm not sure a no-optimizer flag is the way to go. What about controlling if it is enabled via an environment variable. Composer already does this to control some other stuff.

GrahamCampbell:
In any case, wouldn't this also need to be added to the install command too, since it still defaults to doing an update when there is no lock file?

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,275 @@
...
+ */
+class PoolOptimizer
+{
+    /**
+     * @var PolicyInterface
+     */
+    private $policy;
+
+    /**
+     * @var Request
```
GrahamCampbell, line 36: can be null?

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,275 @@
...
+     */
+    private $request;
+
+    /**
+     * @var Pool
```
GrahamCampbell, line 41: can be null?

doc/articles/troubleshooting.md
```diff
@@ -320,3 +320,28 @@ Composer can unpack zipballs using either a system-provided `unzip` utility or P
...
+```
+
+and see, if the result is still the same. It will take significantly longer and use
+a lot more memory to run the dependency resolving process.
+
+If the result is different, you likely hit a problem in the Pool Optimizer.
+Please report this [issue](https://github.com/composer/composer/issues) so it can be fixed.
+
+
+
```
GrahamCampbell, line 347: Too many blank lines at end of file.

tests/Composer/Test/DependencyResolver/PoolOptimizerTest.php
```diff
@@ -0,0 +1,167 @@
...
+            $packages[] = $this->loadPackage($packageData);
+        }
+
+        return $packages;
+    }
+
+    private function loadPackage(array $packageData)
+    {
+        $loader = new ArrayLoader();
+        return $loader->load($packageData);
```
GrahamCampbell, line 165: missing newline before return

GrahamCampbell (changes requested):

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -54,6 +54,10 @@ class PoolBuilder
      * @var EventDispatcher
      */
     private $eventDispatcher;
+    /**
+     * @var PoolOptimizer
```
stof, line 58: `PoolOptimizer|null`

andrerom:
For [ezsystems/ezplatform:2.5](https://github.com/ezsystems/ezplatform/blob/2.5/composer.json):
-- | Composer 2.0-RC2 | Phar in this PR
-- | -- | --
RAM | 110.68MiB (peak: 1419.27MiB) | 58.73MiB (peak: 127.75MiB)
Time | 13.89s | 3.55s


For [ezsystems/ezplatform-ee:2.5](https://github.com/ezsystems/ezplatform-ee/blob/2.5/composer.json), contains commercial packages so you won't be able to replicate w/o subscription:
-- | Composer 2.0-RC2 | Phar in this PR
-- | -- | --
RAM | 123.41MiB (peak: 1580.36MiB) | 74.08MiB (peak: 150.51MiB)
Time | 22.28s | 4.67s

Toflar:
For the record: I had a look at why Drupal is so much slower compared to all the other projects and the reason is not Composer at all. The SAT solver is super fast for them too but the way Drupal handles package meta information is very, very slow (loads of redirects and 404s on https://www.drupal.org/files/packages and https://packages.drupal.org). It easily accounts for 80-90% of the total time spent during a `composer update` run. I think that's a bottleneck the Drupal community has to tackle.

andypost:
@Toflar how to reproduce it? I mean how to enable logging of redirects?

Toflar:
> @Toflar how to reproduce it? I mean how to enable logging of redirects?

@andypost You can see what's going on when running the commands using `-vvv`. I'm not involved in the Drupal project so I just did what was described in https://github.com/composer/composer/pull/9261#issuecomment-705068298 :)

> In any case, wouldn't this also need to be added to the install command too, since it still defaults to doing an update when there is no lock file?

@GrahamCampbell I don't think that's required. There are many other flags like e.g. `--prefer-stable` or `--prefer-lowest` which you cannot control. If you want to control that, use `composer update`.

@stof running the integration tests against both cases is implemented in e6000f0c5b56c325aa95038cf11a1302d739896b.

Toflar:
I found that compacting the constraints when calculating the dependency hash doesn't really bring a lot of benefit. It actually hurts performance more. Here's the comparison on a rather bigger project of mine:

With `Intervals::compactConstrant()`:
```
[81.5MiB/2.93s] Pool builder found a total of 16,440 package versions referenced in your dependency tree. The optimizer was able to filter out 12,660 (77%) of them early!
[70.2MiB/2.93s] Generating rules
[155.1MiB/3.70s] Resolving dependencies through SAT
[155.6MiB/3.74s] Looking at all rules.
[155.6MiB/3.87s] [155.7MiB/3.97s] Something's changed, looking at all rules again (pa[156.1MiB/4.07s] Something's chang[156.6MiB/4.14s] Something's changed, looking at all rules again (pass #4)[156.6MiB/4.15s] 
[156.6MiB/4.15s] Dependency resolution completed in 0.447 seconds
```

Without `Intervals::compactConstrant()`:

```
[81.4MiB/2.49s] Pool builder found a total of 16,440 package versions referenced in your dependency tree. The optimizer was able to filter out 12,656 (77%) of them early!
[70.2MiB/2.49s] Generating rules
[155.3MiB/3.25s] Resolving dependencies through SAT
[155.7MiB/3.30s] Looking at all rules.
[155.8MiB/3.43s] [155.8MiB/3.53s] Something's changed, looking at all rules again (pa[156.2MiB/3.63s] Something's chang[156.7MiB/3.70s] Something's changed, looking at all rules again (pass #4)[156.7MiB/3.71s] 
[156.7MiB/3.71s] Dependency resolution completed in 0.460 seconds
```

As you can see. Compacting the constraints only resulted in 4 additional packages that could be filtered out :)
However, it was almost half a second faster now. So this PR has just gotten yet another small performance boost 😄

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,279 @@
...
+                    // TODO: should that go into package itself?
+                    if ($package->getName() === $link->getTarget()) {
+                        continue;
+                    }
+
+                    $this->addIrremovablePackageConstraint($link->getTarget(), $link->getConstraint());
+                }
+            }
+        }
+
```
Seldaek, line 127: What if you do compactConstraints here once on every irremovablePackageConstraints (instead of doing it in addIrremovablePackageConstraint which ends up doing it many times per constraint I guess)?

Just thinking it's worth a try :)

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,279 @@
...
+            foreach ($package->getRequires() as $link) {
+                $constraint = Intervals::compactConstraint($link->getConstraint());
+                $this->requireConstraintsPerPackage[$link->getTarget()][(string)$constraint] = $constraint;
+            }
+
+            // Mark the alias package as well as the aliased package as irremovable (maybe this can be improved?)
+            if ($package instanceof AliasPackage) {
+                $this->addIrremovablePackageConstraint($package->getName(), new Constraint('==', $package->getVersion()));
+                $this->addIrremovablePackageConstraint($package->getAliasOf()->getName(), new Constraint('==', $package->getAliasOf()->getVersion()));
+            }
```
Seldaek, line 108: For reference IMO here both should be added but only if either ends up being added. Always adding all alias packages is kinda wasteful, but it might be good enough as there aren't that many around and it keeps the code simpler.

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,279 @@
...
+                $hash[] = $target . '@' . $constraint;
+            }
+        }
+
+        return $this->dependencyHashes[$package->id] = implode('', $hash);
+    }
+
+    private function markPackageForRemoval(PackageInterface $package)
+    {
+        // We are now allowed to remove packages if they have been marked as irremovable
```
Seldaek, line 270: typo now=>not

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,279 @@
...
+            $hash[] = $key . ':';
+
+            $subhash = array();
+
+            foreach ($links as $link) {
+                // To get the best dependency hash matches we should use Intervals::compactConstraint() here.
+                // However, the majority of projects are going to specify their constraints already pretty
+                // much in the best variant possible. In other words, we'd be wasting time here and it would actually hurt
+                // performance more than the additional few packages that could be filtered out would benefit the process.
+                $subhash[$link->getTarget()] = (string) $link->getConstraint();
```
Seldaek, line 254: You might be better off using getPrettyString than stringifying constraints here (and possibly in other places). Perf wise at least it'll in most cases return an existing string vs constructing one in __toString.

On the other hand, it risks deoptimizing a few packages because they have two different pretty constraints like ~2.0 and ^2.0 would be the same stringified but not the same with pretty string. Feel free to experiment if you think it's worth it.

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,279 @@
...
+        $this->packagesToRemove = array();
+    }
+
+    private function optimizeByIdenticalDependencies()
+    {
+        $identicalDefinitionPerPackage = array();
+        $packageIdsToRemove = array();
+
+        foreach ($this->pool->getPackages() as $i => $package) {
+
```
Seldaek, line 176: If I get this correctly, it'd be IMO clearer and faster to do the check for irremovability here, so you can skip processing that package entirely.

```suggestion
        if (isset($this->irremovablePackages[$package->id])) {
            continue;
        }
```

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,279 @@
...
+        $this->packagesToRemove = array();
+
+        return $this->pool;
+    }
+
+    private function prepare()
+    {
+        // Mark fixed and required packages from the request as irremovable
+        foreach ($this->request->getRequires() as $require => $constraint) {
+            $this->addIrremovablePackageConstraint($require, $constraint);
```
Seldaek, line 89: Not sure why direct requirements are so special that none of them can be removed? This leads to an issue I'll highlight in a sec by pushing a commit with a couple new tests.

Seldaek:
@Toflar See https://github.com/composer/composer/pull/9261/commits/a16e713ed6e8480a51657c176bcc7a2118c1b4f4 for a failing test due to conflicts which are not handled right now. I'm not entirely sure how to best fix it 'cause it's getting late, but I hope you see the problem.

What I mentioned in https://github.com/composer/composer/pull/9261#discussion_r515415614 also makes it worse because at first I had the test and it passed and I thought huh that's weird what did I not get in the implementation.. and then I realized the root requires are never excluded, so my test was falsly passing, and adding one level of indirection in the -nested test version actually surfaced the issue.

Toflar:
Uh, interesting find! I've extracted the installer test into a more specific pool optimizer test case in https://github.com/composer/composer/pull/9261/commits/d39d3426df71dc5110bc708c17882c8ff04ca34d.
Not sure how to tackle it best yet but I'm sure I'll find a solution for that.

src/Composer/DependencyResolver/PoolOptimizer.php
Toflar: It's basically the "good enough" approach here. Same goes for replacers and maybe more. It's safer to just leave packages untouched when in doubt. We can always apply more improvements to the pool optimizer :)

src/Composer/DependencyResolver/PoolOptimizer.php
Toflar: That's correct. Added in 2b004dbc3f10a05067b8a544c6baf9653c51965f.

src/Composer/DependencyResolver/PoolOptimizer.php
Toflar: To be honest, if we wanted to optimize this for best performance, I think the only correct way is to provide the precalculated hash as package meta data via packagist.org or any repository. It can be calculated once, optimized using `Intervals::compactConstraint()` etc. and there'd be no need to calculate it here at all (unless not provided in which case we'd still need to do it at runtime).

src/Composer/DependencyResolver/PoolOptimizer.php
Seldaek: I'm ok with good enough but I don't understand here what makes these requires so special at all. Can't you just skip this entirely and it'll still work all the same?

tests/Composer/Test/InstallerTest.php
```diff
@@ -198,6 +198,17 @@ public function testSlowIntegration($file, $message, $condition, $composerConfig
      */
     public function testIntegration($file, $message, $condition, $composerConfig, $lock, $installed, $run, $expectLock, $expectInstalled, $expectOutput, $expect, $expectResult)
     {
+        // Test both, with and without pool optimizer to ensure the same outcome
+        $this->doTestIntegration(true, $file, $message, $condition, $composerConfig, $lock, $installed, $run, $expectLock, $expectInstalled, $expectOutput, $expect, $expectResult);
+        $this->doTestIntegration(false, $file, $message, $condition, $composerConfig, $lock, $installed, $run, $expectLock, $expectInstalled, $expectOutput, $expect, $expectResult);
```
stof, line 203: I would do that as a separate test using the same data provider instead (`testIntegrationNoOptimizer`) so that we can have separate failure status for both cases rather than having a single test that contains both runs and so reports success only if both worked, not distinguishing which one failed.

tests/Composer/Test/InstallerTest.php
```diff
@@ -198,6 +198,17 @@ public function testSlowIntegration($file, $message, $condition, $composerConfig
...
+    }
+
+    private function doTestIntegration($withPoolOptimizer, $file, $message, $condition, $composerConfig, $lock, $installed, $run, $expectLock, $expectInstalled, $expectOutput, $expect, $expectResult)
+    {
+        if (!$withPoolOptimizer) {
+            $this->setName($this->getName() . ' [--no-optimizer]');
```
stof, line 209: this looks wrong to me. You are changing the name of the test in the middle of the test execution (as you really have only one test), which might lead to weird things in PHPUnit.

mxr576:
@andypost Hi, is there a Drupal issue related to @Toflar's finding about the performance issue possibly caused by Drupal.org's packagist implementation on Drupal projects?
https://github.com/composer/composer/pull/9261#issuecomment-718987976

andypost:
@mxr576 I already pinged infra-team at https://drupal.slack.com/archives/C51GNJG91/p1604002335238700?thread_ts=1604002335.238700&cid=C51GNJG91

src/Composer/DependencyResolver/PoolOptimizer.php
Toflar: Properly filtered in https://github.com/composer/composer/pull/9261/commits/9b9bf255ca89f807f75d6bffbdcf544d5dd11d60 now :)

Toflar:
Conflict issues are resolved and requested packages are now also filtered - yet another performance improvement 🕺

Toflar:
Now replaced packages are also optimized. That's another 650 packages down in one of my projects and thus yet another 40 MB RAM gone:

```
// Before
[80.3MiB/5.64s] Memory usage: 80.31MiB (peak: 174.78MiB), time: 5.64s
// After
[77.4MiB/5.65s] Memory usage: 77.4MiB (peak: 134.73MiB), time: 5.65s
```

leofeyer:
I ran another test with the Contao Managed Edition (https://github.com/contao/managed-edition/blob/4.9/composer.json) and the latest changes in this PR (https://github.com/composer/composer/commit/c996457a039b2f0a5eee3b6364dfce307c614737) today:

| - | Composer 1.10.17 | Composer 2.0.4 | Phar in this PR |
| --- | --- | --- | --- |
| RAM (peak) | 2295.04MiB | 998.08MiB | 94.99MiB |
| Time | 47.43s | 25.32s | 4.59s |

Looks very promising. 👍🏻

Toflar:
Disabling the pool optimizer is now done using either `"pool-optimizer": false` as configuration value or the `COMPOSER_POOL_OPTIMIZER=0` env var.
I've also restructured the integration tests on the installer.

stof:
> Disabling the pool optimizer is now done using either `"pool-optimizer": false` as configuration value

Do we actually want to have a persistent way to disable it ? to me, the possibility to disable it is a troubleshooting tool (to help identify cases where the pool optimizer is buggy and alters the result of the resolution). But there should be no reason to disable it permanently IMO.

Seldaek:
Yup I tend to agree, as we discussed Yanick I think the env var is enough.
Config clutter isn't needed until it's probably needed :)

doc/articles/troubleshooting.md
GrahamCampbell: Still not resolved.

GrahamCampbell (changes requested):

Toflar:
I agree. The reason I did it like that in the first place was because I've searched the source for how Composer handles env vars and the vast majority of them are configuration values which can be either set in the config or as an environment variable. Hence, I just followed that principle. Moreover, I didn't really find a nice place to set that configuration value other than the `Config` itself. I've moved it to the `Installer` class in ef8a8a7d7a3cc72f43f741aa7204c2614f68b2e9 now which is kind of against everything I've learnt from a software architectural POV but it's probably just fine for debugging purposes only :) 

@GrahamCampbell I don't think your reviews where you comment on coding style are really helpful which is why I ignored them. Sorry about that. My opinion is that CS can be fixed anytime using tools and your comments clutter the PR diffs and make it really hard for me to filter for actual suggestions for improvements on the logic. I'm sure you meant no harm and you had the best intentions only which I highly appreciate! I don't know for others but thank you for not commenting on CS at least for my PRs anymore. Any other suggestions are super welcome of course :)

doc/06-config.md
```diff
@@ -321,3 +321,9 @@ check the presence of extension. If set to `false`, Composer will not create and
 require a `platform_check.php` file as part of the autoloader bootstrap.
 
 &larr; [Repositories](05-repositories.md)  |  [Runtime](07-runtime.md) &rarr;
+
+## pool-optimizer
```
GrahamCampbell, line 325: This doc can be deleted now?

GrahamCampbell (changes requested):

doc/06-config.md
Toflar: Oh right! Removed in 16ed5f6d1181d7c7684b9d9a877567f48a40974e :)

stof:
@Toflar @Seldaek what is the state of this work ?

Seldaek:
@stof you can click here to see the current state:

image

Here is another piece of information:

image

And this is why it's not moving any further I guess, we're all slacking too much

image

fgilio:
This looks amazing 🎉 

Number from a Laravel 8 project with a ton of dependencies:
<table>
<tr>
	<td>
	<td>Composer 2.0.4
	<td>Phar in this PR
<tr>
	<td>Peak RAM
	<td>396.02MiB
	<td>97.26MiB
<tr>
	<td>Time
	<td>10.25s
	<td>9.16s
</table>

mxr576:
This is a promising enhancement, it should not be forgotten.

@Toflar can you resolve the above mentioned conflict?

Toflar:
It won't be forgotten, don't worry. I'll be working on this once Jordi and Nils are ready to spend time on 2.1. Top priority must be to ensure stability of 2.0 now. So be patient, folks :)

driskell:
> @andypost Hi, is there a Drupal issue related to @Toflar's finding about the performance issue possibly caused by Drupal.org's packagist implementation on Drupal projects?
> [#9261 (comment)](https://github.com/composer/composer/pull/9261#issuecomment-718987976)

I think the Drupal issue is down to the `drupal/core-recommended` and others having very large `replaces` lists which contain core modules, or modules moved into core, which do not exist in any repository. These all attempt to get loaded even though they do not exist, thus many 404s.

It happens here. I'm just not sure why these are attempting to get loaded into the pool, when it seems to me, nothing depends on them, and they're not installed. So the replaces declaration seem to be irrelevant since it's not actually performing any replacement. If it was performing a replacement of a package that did exist and mattered then that package would have been encountered in a requires somewhere, right? Should this line be just removed?
https://github.com/composer/composer/blob/master/src/Composer/DependencyResolver/PoolBuilder.php#L402

effulgentsia:
> having very large replaces lists
Is this in and of itself the problem, or is there a different contributing problem to this creating 404s / other performance problems? If it's a problem in and of itself, then have best practices evolved since https://getcomposer.org/doc/04-schema.md#replace was written, which references https://github.com/symfony/symfony/blob/5.x/composer.json as another example that does this?

driskell:
> > having very large replaces lists
> 
> Is this in and of itself the problem, or is there a different contributing problem to this creating 404s / other performance problems? If it's a problem in and of itself, then have best practices evolved since https://getcomposer.org/doc/04-schema.md#replace was written, which references https://github.com/symfony/symfony/blob/5.x/composer.json as another example that does this?

I think the large replace lists is fine and whether or not it 404 (because individual package does not exist) or it 200 (because it does exist) I am unsure why they get added to the pool at all (or try to get added in case of 404). If the replaced package is also required by something then yes it should be in the pool but I’m not sure adding them from the replaces is right approach - but I don’t fully understand that region of code.

Toflar:
That doesn't seem to be related to that PR at all though and is something we might optimize/fix/discuss about in the current 2.0.x release already. Would you mind creating a separate issue for that? You may ping me, I'll look into it.

stof:
@Toflar now that the master branch is officially the codebase for 2.1, it might make sense to fix conflicts in this PR.

src/Composer/DependencyResolver/Pool.php
```diff
@@ -190,6 +193,11 @@ public function isUnacceptableFixedOrLockedPackage(PackageInterface $package)
         return \in_array($package, $this->unacceptableFixedOrLockedPackages, true);
     }
 
+    public function getUnacceptableFixedOrLockedPackages()
```
stof, line 196: I suggest adding `@return PackageInterface[]` in the phpdoc

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,308 @@
...
+
+                    foreach ($this->policy->selectPreferredPackages($this->pool, $literals) as $preferredLiteral) {
+                        unset($packageIdsToRemove[$preferredLiteral]);
+                    }
+                }
+            }
+        }
+
+        foreach (array_keys($packageIdsToRemove) as $id) {
+            $this->markPackageForRemoval($this->pool->packageById($id));
```
stof, line 250: Why passing a `Package` to `markPackageForRemoval` when all it needs is the id, and what you have here are ids as well, forcing you to get the corresponding package ?

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,308 @@
...
+        // in the future with additional optimizations.
+
+        $this->dependencyHashes = array();
+        $this->irremovablePackageConstraints = array();
+        $this->irremovablePackages = array();
+        $this->requireConstraintsPerPackage = array();
+        $this->conflictConstraintsPerPackage = array();
+        $this->packagesToRemove = array();
+
+        return $this->pool;
```
stof, line 84: `$this->pool` and `$this->request` should be reset to `null` before returning IMO, to avoid keeping references to them in the PoolOptimizer

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,308 @@
...
+
+    /**
+     * @param Request $request
+     * @param Pool    $pool
+     *
+     * @return Pool
+     */
+    public function optimize(Request $request, Pool $pool)
+    {
+        $this->request = $request;
```
stof, line 65: `$this->request` is only used inside `prepare()`. To simplify the fact of resetting the state properly, maybe the property should be dropped entirely, passing a `$request` argument to `prepare()` instead.

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        return $this->dependencyHashes[$package->id] = implode('', $hash);
+    }
+
+    private function markPackageForRemoval(PackageInterface $package)
+    {
+        // We are not allowed to remove packages if they have been marked as irremovable
+        if (isset($this->irremovablePackages[$package->id])) {
+            return false;
```
stof, line 301: why returning a boolean ? The return value is ignored

tests/Composer/Test/Fixtures/installer-slow/github-issues-7665.test
```diff
@@ -2800,237 +2800,11 @@ Updating dependencies
...
-    - Conclusion: don't install symfony/finder[v4.1.4] | install symfony/finder[v2.8.49] (conflict analysis result)
-    - Conclusion: don't install symfony/finder[v4.1.3] | install symfony/finder[v2.8.49] (conflict analysis result)
-    - Conclusion: don't install symfony/finder v2.8.49 (conflict analysis result)
     - Root composer.json requires behat/behat ~2.5 -> satisfiable by behat/behat[v2.5.5].
-    - behat/behat v2.5.5 requires symfony/finder ~2.0 -> satisfiable by symfony/finder[v2.6.0, ..., v2.8.49].
-    - You can only install one version of a package, so only one of these can be installed: symfony/finder[v2.6.0, ..., v2.8.49, v3.0.0, ..., v3.4.24, v4.0.0, ..., v4.2.5].
-    - friendsofphp/php-cs-fixer v2.12.0 requires symfony/finder ^3.0 || ^4.0 -> satisfiable by symfony/finder[v3.0.0, ..., v3.4.24, v4.0.0, ..., v4.2.5].
-    - Conclusion: don't install symfony/finder v3.0.1 (conflict analysis result)
+    - You can only install one version of a package, so only one of these can be installed: symfony/finder[v2.8.49, v4.2.5].
+    - friendsofphp/php-cs-fixer v2.14.2 requires symfony/finder ^3.0 || ^4.0 -> satisfiable by symfony/finder[v4.2.5].
```
stof, line 2805: maybe the wording of these `satisfiable by` explanations should be tweaked when the optimizer runs, as it only shows versions that were kept after optimizations, but the constraint can be satisfied by much more versions too (or maybe the explanation for those should use the unoptimized pool).

tests/Composer/Test/InstallerTest.php
```diff
@@ -198,6 +198,27 @@ public function testSlowIntegration($file, $message, $condition, $composerConfig
...
+    {
+        // Disable output comparison
+        $expectOutput = null;
+
+        // Disable pool optimizer
+        putenv('COMPOSER_POOL_OPTIMIZER=0');
+
+        $this->doTestIntegration($file, $message, $condition, $composerConfig, $lock, $installed, $run, $expectLock, $expectInstalled, $expectOutput, $expect, $expectResult);
+
+        putenv('COMPOSER_POOL_OPTIMIZER');
```
stof, line 218: this should be done in a `tearDown` method instead, so that it is properly reset even when tests fail (otherwise, it would affect the next test)

mvorisek:
Currently, composer does large number of requests (ref: https://github.com/composer/composer/issues/6621) when packages are beiing updated. Does this PR reduces the number of requests as well?

mvorisek:
Hi @Toflar, the results are impressive, can you please fix the conflict and feedback so it can get merged soon?

Seldaek:
@mvorisek this is not helpful. Work is being tracked and scheduled here, please don't go around nagging people.

simonberger:
@Seldaek The comment from him might have sounded offensive to you but to be fair people not that closely related to this project have no chance to understand how much communication or agreement between you members and Toflar about a feature like this exists outside of this PR (at least this is what I read in some comments). They see just a feature of a contributor laying around for such a long time and no obvious reason. 
A lack of a real (even rough) feature or release scheduling is not helpful for one looking forward having this or that feature as well.
Don't get me wrong I understand the reasons here and other priorities I just tried to give another perspective on this.

theofidry:
Just in case it may help, from #10200 there is a scenario that benefits **greatly** from this improvement (from timeout to 2s):

```json
{
    "require-dev": {
        "alcaeus/mongo-php-adapter": "^1.2.1",
        "doctrine/data-fixtures": "^1.5.1",
        "doctrine/doctrine-bundle": "^2.4.3",
        "doctrine/mongodb-odm": "^2.2.2",
        "doctrine/mongodb-odm-bundle": "^4.3.0",
        "doctrine/orm": "^2.10.1",
        "doctrine/phpcr-bundle": "^2.3.0",
        "doctrine/phpcr-odm": "^1.5.4",
        "jackalope/jackalope-doctrine-dbal": "^1.7.1",
        "ocramius/proxy-manager": "^2.13",
        "psy/psysh": "^0.10.9",
        "symfony/symfony": "^4.4.32 || ^5.3.9",
        "theofidry/composer-inheritance-plugin": "^1.2",
        "theofidry/psysh-bundle": "^4.4",
        "wouterj/eloquent-bundle": "^2.1.1"
    }
}
```

(the difference is relevant only if there is no .lock & no vendor). I do not know if this may worsen other cases though.

dereuromark:
Do you want to try https://github.com/spryker-shop/suite ? :)
This is probably the most demanding composer project in the world, and as such most likely will show the most dramatic improvements here.

tests/Composer/Test/Fixtures/installer-slow/github-issues-7665.test
Seldaek: Yes maybe we need to update this.. It does read a bit weird. I tried addressing this in https://github.com/Toflar/composer/pull/1 which fixes the problem to some extent.

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -0,0 +1,348 @@
...
+                    }
+
+                    if ($package->getReplaces()) {
+                        foreach ($package->getReplaces() as $link) {
+                            // Make sure we do not replace ourselves (if someone made a mistake and tagged it)
+                            // See e.g. https://github.com/BabDev/Pagerfanta/commit/fd00eb74632fecc0265327e9fe0eddc08c72b238#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780
+                            // TODO: should that go into package itself?
+                            if ($package->getName() === $link->getTarget()) {
+                                continue;
+                            }
```
Seldaek, line 223: Will remove this, fixed in https://github.com/composer/composer/commit/b09a39f9c80185ae96643294b9af08a39aaff379

tests/Composer/Test/DependencyResolver/Fixtures/pooloptimizer/aliases.test
```diff
@@ -1,55 +1,99 @@
 --TEST--
-Test aliased packages remain untouched.
+Test aliased and aliasees remain untouched if either is required, but are still optimized away otherwise.
```
stof, line 2: typo in here

Seldaek: If you mean aliasees, it refers to the aliasOfs i.e. aliasee of the alias packages.

Toflar:
@dereuromark here you go 😉 

  | main@68847ed | this PR@d0b8072
-- | -- | --
RAM | 2004.48MiB | 211.59MiB
Time | 17.46s | 4.38s | 6.66s

gitressa:
This looks awesome, about 25% faster and 80% reduction in memory usage for a medium sized Drupal 9 site with 30 modules:

|          | 2.1.14    | 2.2.0-RC1 |
|----------|-----------|-----------|
| Peak RAM | 292.63MiB | 59.68MiB  |
| Time     | 2.09s     | 1.58s     |


Tested with this:
```
$ composer self-update --preview
$ COMPOSER_DISABLE_NETWORK=1 composer update -v -n --dry-run --profile
```

podarok:
Tested with Open Y ( 229 composer packages )

```
MacBook-Pro-Andrii:openy-project podarok$ composer show | wc -l
     229
```
composer cache warmed up

OSX host

```
php -v
PHP 7.4.25 (cli) (built: Oct 23 2021 18:05:05) ( NTS )
```

```
composer --version
Composer version 2.1.11 2021-11-02 12:10:25
[174.5MiB/44.21s] Memory usage: 174.54MiB (peak: 1000.15MiB), time: 44.21s
```

```
composer --version
Composer version 2.2.0 2021-12-22 11:03:27
[143.5MiB/48.11s] Memory usage: 143.49MiB (peak: 222.83MiB), time: 48.11s
```

Memory improvement - 
77.7 usage reduction peak in comparison to composer 2.1.11. 
Average memory - 17.7% usage reduction

Time is a bit higher +8.8% increase

Toflar:
@podarok you should test without the network interfering (see https://github.com/composer/composer/pull/9261#issuecomment-707030778) :)

fgilio:
Just wanted to take a minute to thank you for making composer this fast, we really appreciate it! It has been a game changer

PR #9267, opened 2020-10-05
Rename private packages page to better reflect content

PrinsFrank:
Split of from #9250 so this can be individually discussed.


PrinsFrank:
@Seldaek, is there a way to redirect doc pages on the getcomposer.org site or do we always have to keep old pages to avoid breaking existing links?

Seldaek:
We can definitely do a redirect somehow, but I'd avoid doing this too often :)

PrinsFrank:
What would you recommend? The page is currently also referencing a private packagist server, and I was also planning to write some documentation about the new gitlab composer package registry. Should we split those up into three pages? Or can we link from the current page to the new page and hide the original one?

Seldaek:
IMO feel free to rename the URL in the other PR, I'll make sure I add a fallback redirect before merging it. Closing this here then.

PR #9270, opened 2020-10-07
Version Guesser: Look at branches on remotes origin & upstream to look for version numbers too

naderman:
Addresses https://github.com/composer/packagist/issues/1125 which previously only worked reliably when using a branch alias for git master.

### DIFF
--- a/src/Composer/Package/Version/VersionGuesser.php
+++ b/src/Composer/Package/Version/VersionGuesser.php
@@ -142,7 +142,7 @@ class VersionGuesser
                 }
 
                 if ($branch && !preg_match('{^ *.+/HEAD }', $branch)) {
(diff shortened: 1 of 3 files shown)


stof:
Tests are broken.

And IMO, you should also add some test cases where the the git comment returns some remote branches too.

naderman:
@stof I'm aware and I will add some, just haven't worked out why those ones are failing yet.

stof:
Well, looks like RootPackageLoaderTest relies on the internal implementation of the VersionGuesser

naderman:
@stof please either send a patch to fix the tests or give me a minute to work out what the problem is. I do understand how to read a stacktrace.

naderman:
Alright, test added, existing ones fixed. Think this is good to go.

Seldaek:
Fix sorting in https://github.com/composer/composer/commit/791bbc80a4e9192a59a37d1282ff5032c4e52692 so that it first tries local branches, just in case they're different/more accurate than remote ones. Otherwise yeah looks good to me

PR #9272, opened 2020-10-08
Fix VcsRepositoryTest

nicolas-grekas:
`getUniqueTmpDirectory` already creates the directory, so that this mkdir is always false, preventing the tests from running.


Seldaek:
Thanks :)

PR #9273, opened 2020-10-08
Add support for "extra.branch-version"

nicolas-grekas:
As discussed in https://github.com/composer/packagist/issues/1125 and in https://github.com/symfony/symfony/pull/38460

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -658,6 +658,17 @@ the branch or tag that is currently checked out. Otherwise, the version should
 be explicitly defined in the package's `composer.json` file. If the version
 cannot be resolved by these means, it is assumed to be `dev-master`.
 
+When the version cannot be inferred from the local VCS repository, you should use
+the special `branch-version` entry under `extra` instead of `version`:
+
+```json
+{
+    "extra": {
+        "branch-version": "4.2-dev"
+    }
+}
+```
+
 The local package will be symlinked if possible, in which case the output in
 the console will read `Symlinking from ../../packages/my-package`. If symlinking
 is _not_ possible the package will be copied. In that case, the console will
--- a/src/Composer/Package/Loader/RootPackageLoader.php
+++ b/src/Composer/Package/Loader/RootPackageLoader.php
@@ -81,8 +81,10 @@ class RootPackageLoader extends ArrayLoader
         if (!isset($config['version'])) {
             $commit = null;
 
-            // override with env var if available
-            if (getenv('COMPOSER_ROOT_VERSION')) {
+            if (isset($config['extra']['branch-version'])) {
+                $config['version'] = preg_replace('{(\.x)?(-dev)?$}', '.x-dev', $config['extra']['branch-version']);
+            } elseif (getenv('COMPOSER_ROOT_VERSION')) {
+                // override with env var if available
                 $config['version'] = getenv('COMPOSER_ROOT_VERSION');
             } else {
                 $versionData = $this->versionGuesser->guessVersion($config, $cwd ?: getcwd());
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -165,6 +165,11 @@ class PathRepository extends ArrayRepository implements ConfigurableRepositoryIn
             );
             $package['transport-options'] = $this->options;
 
+            // use the branch-version as the package version if available
+            if (!isset($package['version']) && isset($package['extra']['branch-version'])) {
+                $package['version'] = preg_replace('{(\.x)?(-dev)?$}', '.x-dev', $package['extra']['branch-version']);
+            }
+
             // carry over the root package version if this path repo is in the same git repository as root package
             if (!isset($package['version']) && ($rootVersion = getenv('COMPOSER_ROOT_VERSION'))) {
                 if (
--- a/tests/Composer/Test/Package/Loader/RootPackageLoaderTest.php
+++ b/tests/Composer/Test/Package/Loader/RootPackageLoaderTest.php
@@ -201,4 +201,15 @@ class RootPackageLoaderTest extends TestCase
 
         $this->assertEquals("dev-latest-production", $package->getPrettyVersion());
     }
+
+    public function testLoadExtraBranchVersion()
+    {
+        $package = $this->loadPackage(array(
+            'extra' => array(
+                'branch-version' => '1.2',
+            ),
+        ));
+
+        $this->assertEquals('1.2.x-dev', $package->getPrettyVersion());
+    }
 }
--- a/tests/Composer/Test/Repository/Fixtures/path/with-branch-version/composer.json
+++ b/tests/Composer/Test/Repository/Fixtures/path/with-branch-version/composer.json
@@ -1,0 +1,6 @@
+{
+  "name": "test/path-branch-versioned",
+  "extra": {
+    "branch-version": "1.2"
+  }
+}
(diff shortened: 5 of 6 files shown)


stof:
should this be in `extra` which is a free-form storage that Composer cannot validate (as it has to allow anything there) or should the `dev-version` key be a known key in the composer schema (that can be documented, validated, etc...) ?

naderman:
I think the only issue I have with this is the naming. I'm not sure calling it "dev-version" is really explaining well what it does. Something like "feature-branch-base-version" would be a bit clearer but is maybe a bit longer? Stof's idea of "root-version" was confusing too, as it doesn't necessarily only apply on the root level. Other ideas?

nicolas-grekas:
What about `extra.branch-alias.dev-*` then? It's quite understandable to me.

Alternatively, what about using the existing `version` field? It would require patching `VcsRepository` instead, which is the only place that blocks. We would make it allow a 3.4.5 tag with a 3.4.x-dev version.

naderman:
I suppose technically it is kind of a wildcard branch alias, so I actually prefer that over the current solution. I suppose we could even treat it as such rather than having any kind of special treatment for this particular setting?

I definitely want to avoid using the current version field for this because of what I tried to explain on the symfony issue already. Let's please not give other people the idea that setting a version number and commiting it is a good idea. This is already the causing the number one support issue we get contacted about: People tagging with one version but defining another in composer.json making it impossible for composer to know which is right. I would really prefer we did not add more places in public which suggested to people that they may want to manually set version numbers despite using git tags.

nicolas-grekas:
> I suppose technically it is kind of a wildcard branch alias, so I actually prefer that over the current solution. I suppose we could even treat it as such rather than having any kind of special treatment for this particular setting?

it is still special, because of the precedence it has over COMPOSER_ROOT_VERSION.

naderman:
@nicolas-grekas Should it have precendence over that? Shouldn't the root version env var have precedence and if it isn't set the branch alias gets used? We just need to make sure the path repository actually ignores the root version, which in my eyes never really made any sense?

nicolas-grekas:
For the root package, I think the exact precedence doesn't matter much.
For Path repos it matters of course. Are you suggesting to remove reading COMPOSER_ROOT_REPO in PathRepository? That'd be a BC break I suppose.

nicolas-grekas:
I updated the PR to use `extra.branch-alias.dev-*`.
I kept it with higher precedence than COMPOSER_ROOT_REPO: this env var is only a fallback, in case the composer.json doesn't contain the version info. Previously only the `version` field allowed defining this version. Now `dev-*` is possible.

Seldaek:
To be honest, I am still not convinced this is a good idea.. Let me try to summarize the issue and proposed solutions to make sure we're all on the same page:

## Problem

- A monorepo contains sub-packages which have a different version than the branch would suggest.
- e.g. Branch 5.x still has a sub-package with 2.3.x-dev for example (for symfony/contracts)
- Defining a different branch-alias in the sub-project was possible before (dev-master => 5.x for root json, and dev-master => 2.3.x for sub-project), but is not when the branch names are numeric, as those can not be aliased, so when 5.x is checked out, all packages have 5.x-dev as version.
- COMPOSER_ROOT_VERSION can set/override the version of root package and sub-packages to 5.x-dev, but that's not correct for the sub-package.

## Proposed Solutions

- `root_version` field, I find confusing as it is in this case not even applied as a root package, the sub-package gets loaded via the path repo. If it was the root package, we could use COMPOSER_ROOT_VERSION to set the version and there would be no problem.
- `extra.branch-alias.dev-*` while my initial idea due to a misunderstanding (I thought it was only about branches not being detected correctly), I don't think this is appropriate for this anymore. As implemented, it isn't a fallback branch-alias, it's an override of the branch entirely.
- `extra.dev-version` field is also quite confusing IMO as it has not much to do with development.

## Conclusion

The actual solution needed is a way to define a version for some sub-package so that the branch/VCS version detection does not trigger, because for some reason the versioning of the sub-package is detached from the versioning of the rest of the repo. It lives in its own parallel universe in there.

It sounds to me like the perfect solution here would actually be to set the `version` field to `2.3.x-dev` and then when tagging it'd receive another/fixed version. If we do add some other mechanism, then I'd like for it to be clearly distinguishable and ideally self-explanatory. It should be a vcs override of some sort, which lets a path repo be loaded with another version than the branch of the repo.

I am not sure if it is possible to achieve something like this though.. The current PR's solution means that if I git checkout symfony/symfony 5.1.0 for example, I'd get symfony/contracts 2.3.x-dev still.. Is that correct? Not really. I don't know what the correct version would be, but IMO not a dev one if I'm on a tag. So it should not only override branch detection but also tag detection, and then it really means it should act exactly like the `version` field. So why not use that?

One more gap I have in my understanding is.. how do you currently handle the subtree split versioning @nicolas-grekas? Because the problematic there I assume is very similar, you have a 5.1.0 tag on the main repo, but somehow it doesn't get reflected on the split repo.. How/where do you define the versions for the split?

nicolas-grekas:
> The current PR's solution means that if I git checkout symfony/symfony 5.1.0 for example, I'd get symfony/contracts 2.3.x-dev still.. Is that correct?

You're missing the fact that symfony/symfony doesn't "replace" symfony/contracts. You're never "on a tag" here, so "dev" is accurate. Composer will install the package separately. For dev, we define src/Symfony/Contracts as a "path" repo, so that the CI runs the patched code, when contracts are touched in a PR.

> how do you currently handle the subtree split versioning

for contracts: manually.

> then it really means it should act exactly like the version field. So why not use that?

that's my initial reasoning also. And actually, this works perfectly fine. But `VcsRepository` adds this additional constraint on the "version" field, where it rejects those who don't match a tag exactly. At first, I was looking for patching `VcsRepository` to relax this constraint. But if we go this route, it'd mean ppl won't be able to install symfony from source until they upgrade composer before (right?). So that for BC/FC, a separate field might be a smoother path. Call it dev-version, root_version, `dev-*`, I don't really care, tell me :)

Seldaek:
Well we don't need to patch vcs repo if you make sure to update the version field before tagging symfony/contracts to match what the tag is going to be. Is this an option for you?

You can do a commit updating it in a new tmp branch which then has the tag and gets deleted after, so you don't need two commits to change version and then undo it. I do the same for composer releases every tag is a separate commit "forking off" master.

Otherwise.. if we do another solution indeed the question is what to call it. I wish I knew. `extra.branch-override`?

nicolas-grekas:
> update the version field before tagging symfony/contracts to match what the tag is going to be. Is this an option for you?

This will create a significant maintenance burden and will be error prone I fear.

`extra.branch-version`?

naderman:
Think that name `extra.branch-version` works.

nicolas-grekas:
Great, PR updated, thanks!

doc/04-schema.md
```diff
@@ -75,6 +75,9 @@ to omit it.
 > **Note:** Packagist uses VCS repositories, so the statement above is very
 > much true for Packagist as well. Specifying the version yourself will
 > most likely end up creating problems at some point due to human error.
+> If VCS tag names don't work for you and you need to define the version of the
+> root package in its composer.json file, use the special `branch-version` entry
+> under `extra`.
```
Seldaek, line 80: I think maybe I'd remove this note here.. I don't really understand what it's trying to say, so I guess it'll confuse others even more. And I don't think this edge case feature needs to be so prominently mentioned.

nicolas-grekas: alright, removed.

Seldaek:
Thanks @nicolas-grekas for the fast response and sorry for the back and forth here.. hopefully everyone will be happy with this solution for years to come ;)

nicolas-grekas:
Thanks to both of you for caring! :)

fabpot:
Thank you @naderman and @Seldaek for your help on this one, very much appreciated.

PR #9278, opened 2020-10-13
Plugin helpers for v2

### LINKED ISSUE #9209: Download a zip file in a plugin for Composer 1 & 2
I'm creating a Composer plugin that does 2 things:

- calls an URL that points to a JSON, obtain the result, and decode it
- downloads an archive file (zip, rar, tar) and expands it in a given folder.

Because I'm creating the plugin now, if I target only Composer 1 I'm afraid that soon I'll need to maintain 2 versions, obe code Composer 1 and one for Composer 2, as I don't expect to be able to deprecate the version that targets Composer 1 son, considering I'm releasing it now.

This is why I would like to have **a plugin that works with both Composer 1 and Composer 2**.

The plugin is not public so I can't point to it, but the simplified version of relevant code that works with Composer 1 looks like this:

```php
class MyPlugin implements \Composer\Plugin\PluginInterface {

    private $composer;
    private $io;

    public function activate(\Composer\Composer $composer, \Composer\IO\IOInterface $io)
    {
        $this->composer = $composer;
        $this->io = $io;
    }


    public function getJson(string $url): array
    {
        $config = $this->composer->getConfig();
        $rfs = \Composer\Factory::createRemoteFilesystem($this->io, $config);
        $origin = \Composer\Util\RemoteFilesystem::getOrigin($url);
        $response = $rfs->getContents($origin, $url, false);
        $json = $response ? json_decode($response, true) : null;
        if (!$json || !is_array($json)) {
            throw new \Exception("Failed call to {$url}.");
        }

        return $json;
    }

    
    public function download(string $url, string $targetDir): void
    {
        $type = strtolower((string)pathinfo($url, PATHINFO_EXTENSION));

        $config = $this->composer->getConfig();
        $executor = new \Composer\Util\ProcessExecutor($this->io);
        $rfs = \Composer\Factory::createRemoteFilesystem($this->io, $config);

        $params = [$this->io, $config, null, null, $executor, $rfs];

        switch ($type) {
            case 'rar':
                $downloader = new \Composer\Downloader\RarDownloader(...$params);
                break;
            case 'tar':
                $downloader = new \Composer\Downloader\TarDownloader(...$params);
                break;
            case 'xz':
                $downloader = new \Composer\Downloader\XzDownloader(...$params);
                break;
            case 'zip':
            default:
                $downloader = new \Composer\Downloader\ZipDownloader(...$params);
                break;
        }

        $package = new \Composer\Package\Package('my/download', '1.0.0.0', '1');
        $package->setDistType($type);
        $package->setDistUrl($url);
        $package->setTargetDir($targetDir);

        $downloader->download($package, $targetDir);
    }
}
```

In the upgrade to 2.0 notes, I read that I should use `HttpDownloader` instead of `RemoteFilesystem` (that became `@internal`) so I'm trying to do some conditions that work in both versions.

For the `getJson` method, I'm doing something like this:

```php
    public function __construct()
    {
        $this->ver2 = version_compare(
            \Composer\Plugin\PluginInterface::PLUGIN_API_VERSION,
            '2',
            '>='
        );
    }

    public function getJson(string $url): array
    {
        $config = $this->composer->getConfig();
        $client = $this->ver2
            ? Composer\Factory::createHttpDownloader($this->io, $config)
            : Composer\Factory::createRemoteFilesystem($this->io, $config);

        $response = $this->ver2
            ? $client->get($url)
            : $client->getContents(\Composer\Util\RemoteFilesystem::getOrigin($url), $url, false);

        $json = $response ? json_decode($response, true) : null;
        if (!$json || !is_array($json)) {
            throw new \Exception("Failed call to {$url}.");
        }

        return $json;
    }
```

but this **fail** in Composer 2.

The reason is that `HttpDownloader::get()` that is supposed to be sync calls internally `HttpDownloader::addJob()` by passing as second parameter and hardcoded `true` (see https://github.com/composer/composer/blob/master/src/Composer/Util/HttpDownloader.php#L84) which is the `$sync` paramer of that method (see https://github.com/composer/composer/blob/master/src/Composer/Util/HttpDownloader.php#L132).

Because `HttpDownloader::addJob()` is called by `get` with second parameter `true`, and because the `HttpDownloader` instance I have is not async ([`HttpDownloader::enableAsync()`](https://github.com/composer/composer/blob/master/src/Composer/Util/HttpDownloader.php#L286) that is `@internal` was not called) the result of calling `get()` is that an exception saying _"You must use the HttpDownloader instance which is part of a Composer\Loop instance to be able to run async http requests"_ is thrown here: https://github.com/composer/composer/blob/master/src/Composer/Util/HttpDownloader.php#L145

Even if I obtain a `Loop` instance via `Composer::getLoop()` then there's no way to obtain a `HttpDownloader` from `Loop` because the property that holds it is private and there's no accessor.

Even if I would extend `HttpDownloader` to change the behavior of `get` and call `HttpDownloader::addJob()` with the second param set to `false`, that is also not possible, being `HttpDownloader::addJob()` a private method.

I that that `HttpDownloader::get()` should accept a third param `$sync` (probably defaulted to `true` to don't break existing code) so that it would be possible in such cases execute get synchronously.

So that:

```php
    public function get($url, $options = array())
    {
        list($job, $promise) = $this->addJob(array('url' => $url, 'options' => $options, 'copyTo' => false), true);
        $this->wait($job['id']);

        return $this->getResponse($job['id']);
    }
```

would become:

```php
    public function get($url, $options = array(), $sync = true)
    {
        list($job, $promise) = $this->addJob(
            array('url' => $url, 'options' => $options, 'copyTo' => false),
            (bool)$sync
        );
        $this->wait($job['id']);

        return $this->getResponse($job['id']);
    }
```

This would solve (I think) the _first part_ of my plugin.

For the second part (download an archive and expand), now that the `DownloaderInterface` has completely changed (involving `download`, `prepare`, `install` and `cleanup`), I'm not sure how the line `$downloader->download($package, $targetDir);` that worked in Composer 1 should be changed to work in Composer 2.

If I understood correctly, it should be:

```php
$promise = $downloader->download($package, $targetDir, null, false);
$promise->then(
    function () use ($downloader, $package, $path) {
       $downloader->prepare('install', $package, $path);
       $downloader->install($package, $path, false);
       $downloader->cleanup('install', $package, $path);
   },
   function () use ($downloader, $package, $path) {
       $this->io->writeError("Could not download " . $package->getName());
       $downloader->cleanup('install', $package, $path);
   }
);
```

Is that correct?

Is calling `then` enough to tell the promise should be resolved or do I need to use a `Loop` instance?

Or maybe there's a different approach I could take to make my `MyPlugin::download()` method work in Composer 2?

Seldaek:
Fixes #9209

Open for more ideas if someone has anything to add here.

### DIFF
--- a/src/Composer/Util/Loop.php
+++ b/src/Composer/Util/Loop.php
@@ -21,20 +21,38 @@ use Symfony\Component\Console\Helper\ProgressBar;
  */
 class Loop
 {
+    /** @var HttpDownloader */
     private $httpDownloader;
+    /** @var ProcessExecutor|null */
     private $processExecutor;
+    /** @var Promise[]|null */
     private $currentPromises;
 
-    public function __construct(HttpDownloader $httpDownloader = null, ProcessExecutor $processExecutor = null)
+    public function __construct(HttpDownloader $httpDownloader, ProcessExecutor $processExecutor = null)
     {
         $this->httpDownloader = $httpDownloader;
-        if ($this->httpDownloader) {
-            $this->httpDownloader->enableAsync();
-        }
+        $this->httpDownloader->enableAsync();
+
         $this->processExecutor = $processExecutor;
         if ($this->processExecutor) {
             $this->processExecutor->enableAsync();
         }
+    }
+
+    /**
+     * @return HttpDownloader
+     */
+    public function getHttpDownloader()
+    {
+        return $this->httpDownloader;
+    }
+
+    /**
+     * @return ProcessExecutor|null
+     */
+    public function getProcessExecutor()
+    {
+        return $this->processExecutor;
     }
 
     public function wait(array $promises, ProgressBar $progress = null)
--- a/doc/articles/plugins.md
+++ b/doc/articles/plugins.md
@@ -161,6 +161,14 @@ class AwsPlugin implements PluginInterface, EventSubscriberInterface
         $this->io = $io;
     }
 
+    public function deactivate(Composer $composer, IOInterface $io)
+    {
+    }
+
+    public function uninstall(Composer $composer, IOInterface $io)
+    {
+    }
+
     public static function getSubscribedEvents()
     {
         return array(
@@ -175,9 +183,7 @@ class AwsPlugin implements PluginInterface, EventSubscriberInterface
         $protocol = parse_url($event->getProcessedUrl(), PHP_URL_SCHEME);
 
         if ($protocol === 's3') {
-            $awsClient = new AwsClient($this->io, $this->composer->getConfig());
-            $s3Downloader = new S3Downloader($this->io, $event->getHttpDownloader()->getOptions(), $awsClient);
-            $event->setHttpdownloader($s3Downloader);
+            // ...
         }
     }
 }
@@ -263,7 +269,7 @@ Now the `custom-plugin-command` is available alongside Composer commands.
 
 ## Running plugins manually
 
-Plugins for an event can be run manually by the `run-script` command. This works the same way as 
+Plugins for an event can be run manually by the `run-script` command. This works the same way as
 [running scripts manually](scripts.md#running-scripts-manually).
 
 ## Using Plugins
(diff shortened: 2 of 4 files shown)


naderman (approved):

src/Composer/Util/Loop.php
```diff
@@ -21,22 +21,40 @@
...
     private $currentPromises;
 
-    public function __construct(HttpDownloader $httpDownloader = null, ProcessExecutor $processExecutor = null)
+    public function __construct(HttpDownloader $httpDownloader, ProcessExecutor $processExecutor = null)
     {
         $this->httpDownloader = $httpDownloader;
-        if ($this->httpDownloader) {
-            $this->httpDownloader->enableAsync();
-        }
+        $this->httpDownloader->enableAsync();
```
GrahamCampbell, line 34: I know this is not an issue with this PR, but it seems odd that we mutate the client here.

Seldaek: Yeah it's not ideal, but it lets people create and use a sync client without shooting themselves in the foot.

PR #9280, opened 2020-10-13
Fix parsing "branch-version"

nicolas-grekas:
My bad, the current code turns `1.2-dev` in to `1.2.x-dev.x-dev`.


tests/Composer/Test/Package/Loader/RootPackageLoaderTest.php
```diff
@@ -211,5 +211,13 @@ public function testLoadExtraBranchVersion()
         ));
 
         $this->assertEquals('1.2.x-dev', $package->getPrettyVersion());
+
+        $package = $this->loadPackage(array(
+            'extra' => array(
+                'branch-version' => '1.2-dev',
```
stof, line 217: Rather than doing several tests one after each other in the same test method, I suggest using a data provider.

stof: Btw, I suggest also adding `1.2.x-dev` in the test

nicolas-grekas: Good idea, updated.

PR #9283, opened 2020-10-13
Update authentication methods in documentation

PrinsFrank:
In PR #9250 a first start was made on documenting authentication for private packages. The github and bitbucket authentication methods were not initially included as I was only thinking about composer registry options, but this PR documents those remaining authentication methods (as requested by @Seldaek) and clarifies that the documentation is about authentication in general, not only against composer registries.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -900,9 +900,10 @@ If set to 1, this env allows running Composer when the Xdebug extension is enabl
 ### COMPOSER_AUTH
 
 The `COMPOSER_AUTH` var allows you to set up authentication as an environment variable.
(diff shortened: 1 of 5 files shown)


Seldaek:
Thanks! If you want to continue doing more, generally speaking as a style thing I'd say try to [avoid "here" as a link](http://stephanieleary.com/2015/05/why-click-here-is-a-terrible-link-and-what-to-write-instead/). See why [here](http://stephanieleary.com/2015/05/why-click-here-is-a-terrible-link-and-what-to-write-instead/) ;)

It's a good thing to keep in mind IMO, both from an accessibility perspective for screenreaders but also generally for everyone reading it tends to flow better if not using this good old click here pattern.

PR #9286, opened 2020-10-14
Refactor Operations classes to get rid of strings comparison

bezpiatovs:
This commit is made in order to make possible to compare operation types without hard-coding strings within code base.

The new way of comparison removes dependencies on strings (e.g. `'update'`, `'install`' etc.) during comparison:
`if ($opType !== UninstallOperation::TYPE) {
...
}`

It will also avoid ambiguity of using `instanceOf` comparison for those who is against of comparing strings explicitly.


Seldaek:
Thanks for the PR but I'll stop you before you spend any more time here.. This is IMO solving a non-issue. And it adds BC breaks on the operation interface on top of it, and changes the API for update operation.. it's just not worth it. Adding constants perhaps although it's really fine as is IMO, not like these names are spreading all over the place, but the rest of the changes are not needed nor desired.

bezpiatovs:
> Thanks for the PR but I'll stop you before you spend any more time here.. This is IMO solving a non-issue. And it adds BC breaks on the operation interface on top of it, and changes the API for update operation.. it's just not worth it. Adding constants perhaps although it's really fine as is IMO, not like these names are spreading all over the place, but the rest of the changes are not needed nor desired.

@Seldaek Agree on this.
I will leave just constants changes. On your desire to use it or not, but it was an issue for us today what exact comparison to use within our installer, `instanceOf` or `=== 'update'` instead. And I assume we are not alone with this. Constants will add explicitness in this case IMO.

Seldaek:
Ok yes current PR looks better, thanks for the updates.

What "installer" are you working on if you don't mind me asking? Curious what you'd do that needs such deep integration with operations.

bezpiatovs:
> Ok yes current PR looks better, thanks for the updates.
> 
> What "installer" are you working on if you don't mind me asking? Curious what you'd do that needs such deep integration with operations.

We are using simplified fork of
https://github.com/wikimedia/composer-merge-plugin/blob/master/src/MergePlugin.php#L312

PR #9291, opened 2020-10-15
Fix check for symfony/flex

nicolas-grekas:


Seldaek:
Oh already fixed in https://github.com/composer/composer/commit/2ca2ec3d0ae77bbadc8a4c60180c7770bddc4b32 but thanks :)

PR #9296, opened 2020-10-15
Fix availability of $urls in FileDownloader

mcaskill:
Fixes #9220, as per [comment](https://github.com/composer/composer/pull/9220#issuecomment-709510751):

> The `$urls` variable is passed by reference to each callback (`$download`, `$accept`, `$reject`) but the manipulations, in `$download`, made to `$url` are done on a copy:
> 
> ```php
> $download = function () use ($io, $output, $httpDownloader, $cache, $cacheKeyGenerator, $eventDispatcher, $package, $fileName, &$urls, &$accept, &$reject) {
>     $url = reset($urls);
> ```
> 
> Once `$accept` is called, it still has the original dataset:
> 
> ```php
> $accept = function ($response) use ($cache, $package, $fileName, $self, &$urls) {
>     $url = reset($urls);
> ```

This pull request ensures the manipulated `$url` is reassigned back to `$urls` to be available to `$accept` and `$reject`.

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -142,6 +142,7 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
         $reject = null;
         $download = function () use ($io, $output, $httpDownloader, $cache, $cacheKeyGenerator, $eventDispatcher, $package, $fileName, &$urls, &$accept, &$reject) {
             $url = reset($urls);
+            $index = key($urls);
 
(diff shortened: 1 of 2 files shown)


mcaskill:
If someone could validate that the logic of my unit tests is sound, that would be much appreciated!

Seldaek:
Looks good to me, both fix and tests.. Merging now, feel free to validate things work for you with the latest build `composer self-update --snapshot` :)

PR #9305, opened 2020-10-20
Improve readability of version info message

TerryKern:
Instead of: 

> You are already using composer version 1.10.15 (1 channel).

it shows: 

> You are already using composer version 1.10.15 (1.x channel).

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -176,11 +176,12 @@ EOT
             return 1;
         }
 
+        $channelString = $versionsUtil->getChannel();
+        if (is_numeric($channelString)) {
+            $channelString .= '.x';
+        }
+
         if (Composer::VERSION === $updateVersion) {
-            $channelString = $versionsUtil->getChannel();
-            if (is_numeric($channelString)) {
-                $channelString .= '.x';
-            }
             $io->writeError(
                 sprintf(
                     '<info>You are already using composer version %s (%s channel).</info>',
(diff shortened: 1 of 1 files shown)


src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -177,7 +177,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         if (Composer::VERSION === $updateVersion) {
-            $io->writeError(sprintf('<info>You are already using composer version %s (%s channel).</info>', $updateVersion, $versionsUtil->getChannel()));
+            $io->writeError(sprintf('<info>You are already using composer version %s (%s.x channel).</info>', $updateVersion, $versionsUtil->getChannel()));
```
stof, line 180: wouldn't this cause issues when the channel is `preview` for instance ?

TerryKern: What are the possible contents of ``$versionsUtil->getChannel())``?

TerryKern: Is there an alternative to ``getChannel()`` that produces a more readable output?

stof: The list of channels is at https://github.com/composer/composer/blob/99a5469b5e3d956d3b7e014147186bbd3678276e/src/Composer/SelfUpdate/Versions.php#L24

TerryKern: @stof What do you think of the new solution?

TerryKern:
@Seldaek I retargetted it to 1.10

Seldaek:
Thanks, sorry looking at this again, I'd say it probably makes sense to tweak also line 211 the output for when an update actually happens.

TerryKern:
> Thanks, sorry looking at this again, I'd say it probably makes sense to tweak also line 211 the output for when an update actually happens.

@Seldaek Done :)

Seldaek:
Thanks

TerryKern:
@Seldaek Would you consider marking this as a valid Hacktoberfest PR?
https://hacktoberfest.digitalocean.com/hacktoberfest-update

Seldaek:
I opted in the repo so this should be valid now.

TerryKern:
@Seldaek Thank you :)

Issue #9067, opened 2020-07-23
Composer2: SolverBugException


glaubinix:
My `composer.json`:

```json
{
    "repositories": [
        {
            "type": "package",
            "package": [{
                "name": "test/test",
                "version": "1.0",
                "require": {
                    "illuminate/cache": "^6.5"
                }
            }]
        },
        {
            "type": "package",
            "package": [{
                "name": "test/other",
                "version": "1.0",
                "require": {
                    "vlucas/phpdotenv": "^2.4"
                }
            }]
        }
    ],
    "require": {
        "laravel/framework": "^6.0",
        "test/test": "^1.0",
        "test/other": "^1.0"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.0-dev+d8fa7464334404643237971bdf921120f784916e
PHP version: 7.2.31
PHP binary path: /usr/local/Cellar/php@7.2/7.2.31_1/bin/php
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
```

When I run this command:

```
composer update -vvv
```

I get the following output:

```
Loading composer repositories with package information
Updating dependencies


  [Composer\DependencyResolver\SolverBugException]
  Reached invalid decision id 0 while looking through (-184) for a literal present in the analyzed
   rule (80|81|82|83|84|85|86|87|88|89|90|91|92|93|94|95|96|97|98|99|100|101|102|103|104|105|106|1
  07|108|109|110|111|112|113|114|115|116|117|118|119|120|121|122|123|124|125|126|127|128|129|130|1
  31|132|133|134|135|136|137).
  This exception was most likely caused by a bug in Composer.
  Please report the command you ran, the exact error you received, and your composer.json on https
  ://github.com/composer/composer/issues - thank you!


[...]

Issue #9294, opened 2020-10-15
Solver Issues Summary


Seldaek:
There are a number of open issues with tons of repro cases in each which is getting very messy, here is a summary of known test cases which still fail with RC2.

These seem to work in v1 so look more like regressions / higher prio

- [x] https://gist.github.com/Seldaek/cb330c80c333ea191fd5f256447ffb82 minified test case from #9290 which actually triggers a SolverBugException while v1 resolves to an install, looks like a legit regression
- [x] https://gist.github.com/Seldaek/f69ce1b1628c049a5e4903c1feab41a4 minified test case from #9067 which actually triggers a SolverBugException while v1 resolved to a problem exception, looks like a legit regression
- [x] https://gist.github.com/Seldaek/951e27a30398de2db0cacb85c76aa917 minified test cases extracted from OP in #8903, one case resolves quickly and outputs **lots** of problems, the other gets stuck quickly

----

These I think already failed in v1.. they should still be fixed tho for sure if possible.

- [x] https://gist.github.com/naderman/1399cdee05d6d0027890bd0b3e3c0e7f minified case from https://github.com/composer/composer/issues/7665#issuecomment-483069643, never resolves (pass `#1687` and then frozen for hours) - already has a PR open without fix tho in https://github.com/composer/composer/pull/8733
- ~~https://gist.github.com/Seldaek/81ee6492e1e6f8b07d4f4268ccb8d0a2 minified test case from OP in https://github.com/composer/composer/issues/9067 - it resolves and outputs lots of problems, might have been minified too much, not sure~~
- [x] https://gist.github.com/Seldaek/4e2dbc2cea4b4fd7a8207bb310ec8e34 minified test case from OP in #9012 - it gets stuck quickly unless some versions get deleted then it resolves and outputs lots of problems

stof:
@naderman @Seldaek is this expected to be solved before releasing 2.0 as stable in a few days ?

naderman:
@stof Definitely going to try to at least figure out the regression cases, but not entirely decided yet, really depends on the complexity/time required to work this out.

Seldaek:
Closing this as all related issues are closed 🥳

Issue #9290, opened 2020-10-15
Composer 2 solver error


fabpot:
My `composer.json`:

```json
{
    "require": {
        "php": ">=7.4",
        "mailgun/mailgun-php": "^2.8",
        "symfony/flex": "^1"
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "extra": {
        "symfony": {
            "require": "5.2.*"
        }
    }
}
```

When I run this command:

```
composer up
```

I get the following output:

```
  [Composer\DependencyResolver\SolverBugException]                                                                                                                                  
  Reached invalid decision id 0 while looking through (-275) for a literal present in the analyzed rule (-94|274|275|276|277|278|279|280|281|282|283|284|285|286|287|288|289|328).  
  This exception was most likely caused by a bug in Composer.                                                                                                                       
  Please report the command you ran, the exact error you received, and your composer.json on https://github.com/composer/composer/issues - thank you!
```

/cc @nicolas-grekas

Seldaek:
This is somehow caused by flex as it works fine with --no-plugins, but I am not sure why.

nicolas-grekas:
Note that this happens only with composer 2.
Also, after the first error, if I run `composer i`, I get
`Fixed package symfony/options-resolver 5.1.7.0 was not added to solver pool.`

Seldaek:
Yeah that last one is definitely a flex bug.. fixed packages should not be removed from the pool ever, no matter what the extra.symfony.require key says..

Seldaek:
I think you can use `$prePoolCreateEvent->getRequest()->getFixedOrLockedPackages()` to get the packages which you can not filter out.

nicolas-grekas:
> you can use $prePoolCreateEvent->getRequest()->getFixedOrLockedPackages()

thanks for the hint, it worked, see https://github.com/symfony/flex/pull/704

About the original issue, it happens because of Flex' `PackageFilter::removeLegacyPackages()`, but honestly I don't see anything wrong in the result of this method, it correctly filters unwanted packages.

This looks like a composer issue really.

Seldaek:
yeah it might be a composer bug anyway, but it's hard to repro because it's dependent on the filtering that flex does.. Is it correct you only filter on PRE_POOL_CREATE for Composer 2? Or are you filtering on any other event still?

nicolas-grekas:
> Is it correct you only filter on PRE_POOL_CREATE for Composer 2?

[...]

PR #9309, opened 2020-10-21
MultiConflictRule conflict analysis handling on level 1 decisions

naderman:
Addresses https://github.com/composer/composer/issues/9290 which is also item 1 on https://github.com/composer/composer/issues/9294

After the fix in https://github.com/composer/composer/commit/851050e85c07f68ce0a0290fa53f6ed884751a6b it could now happen that $l1num gets incremented twice or $l1num and $num both got incremented for the same literal, which is not supposed to happen.

### DIFF
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -465,35 +465,35 @@ class Solver
                     // only level 1 marks left
                     $l1num++;
                     $l1retry = true;
-                }
+                } else {
+                    $decision = $this->decisions->atOffset($decisionId);
+                    $rule = $decision[Decisions::DECISION_REASON];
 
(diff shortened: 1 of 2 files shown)


naderman:
Appears to also fix https://github.com/composer/composer/issues/9067

stof (approved):

mvorisek:
please check if fixes https://github.com/composer/composer/issues/9313

CI install commands are defined in https://github.com/atk4/ui/pull/1493/files#diff-cc9b95888c99ec7211a604efe09d0745c289e75b112394022440e8b6873f7952R264-R267

naderman:
@mvorisek You can check it yourself, just clone this repo, composer install the deps and use the bin/composer script in here instead of your system composer installation.

Issue #9012, opened 2020-06-24
[v2] Reached invalid decision id 0 while looking through


williamdes:
My `composer.json`:

```json
{
    "require": {
        "php": ">=7.1.7",
        "astrotomic/laravel-translatable": "^11.7",
        "barryvdh/laravel-snappy": "^0.4.4",
        "brozot/laravel-fcm": "^1.2",
        "daltcore/lara-pdf-merger": "^2.0",
        "doctrine/dbal": "^2.9",
        "eluceo/ical": "^0.16.0",
        "ezyang/htmlpurifier": "^4.9",
        "facebook/graph-sdk": "^5.7",
        "biscolab/laravel-recaptcha": "^4.0",
        "guzzlehttp/guzzle": "~6.3",
        "intervention/image": "^2.4",
        "intervention/validation": "^2.4",
        "jenssegers/agent": "^2.6",
        "jenssegers/date": "^4.0",
        "laravel/framework": "^6.0",
        "laravel/helpers": "^1.1.1",
        "laravel/tinker": "^2.4",
        "laravelcollective/html": "^6.0",
        "league/flysystem-aws-s3-v3": "^1.0",
        "league/fractal": "^0.17.0",
        "maatwebsite/excel": "^3.1",
        "mattketmo/email-checker": "^1.5",
        "omniphx/forrest": "^2.6",
        "pixelpeter/laravel5-isocodes-validation": "^3.0",
        "predis/predis": "^1.1",
        "propaganistas/laravel-phone": "^4.0",
        "ramsey/uuid": "^3.7",
        "sentry/sentry-laravel": "^1.0",
        "symfony/intl": "^4.0",
        "therezor/laravel-transactional-jobs": "^1.0",
        "tymon/jwt-auth": "1.0.0",
        "ext-json": "*"
    },
    "require-dev": {
        "barryvdh/laravel-debugbar": "^3.2",
        "barryvdh/laravel-ide-helper": "^2.6",
        "editorconfig-checker/editorconfig-checker": "^10.0",
        "filp/whoops": "^2.3",
        "fzaninotto/faker": "^1.8",
        "johnkary/phpunit-speedtrap": "^3.2",
        "mockery/mockery": "^1.2",
        "nunomaduro/collision": "^4.0",
        "nunomaduro/larastan": "^0.6",
        "phpunit/phpunit": "^8.0",
        "slevomat/coding-standard": "^6.1",
        "squizlabs/php_codesniffer": "^3.3",
[...]

PR #9312, opened 2020-10-21
Solver: Prevent infinite recursion in analyzeUnsolvableRule

naderman:
In complex scenarios reasons for learned rules can themselves be learned
rules caused by other learned rules which had some of the same
reasons. In this situation iterating over all problem rules requires
keeping track of which rules have previously been analyzed to avoid an
endless loop.

Side effect is that the sorting of problems including learned rules
changes slightly.

Addresses https://github.com/composer/composer/issues/9012 which is item 6 on https://github.com/composer/composer/issues/9294

### DIFF
--- a/src/Composer/DependencyResolver/Solver.php
+++ b/src/Composer/DependencyResolver/Solver.php
@@ -514,16 +514,12 @@ class Solver
         return array($learnedLiterals[0], $ruleLevel, $newRule, $why);
     }
 
-    /**
-     * @param Problem $problem
-     * @param Rule    $conflictRule
-     */
-    private function analyzeUnsolvableRule(Problem $problem, Rule $conflictRule, &$ruleSeen)
+    private function analyzeUnsolvableRule(Problem $problem, Rule $conflictRule, array &$ruleSeen)
     {
-        $ruleSeen[spl_object_hash($conflictRule)] = true;
+        $why = spl_object_hash($conflictRule);
+        $ruleSeen[$why] = true;
 
         if ($conflictRule->getType() == RuleSet::TYPE_LEARNED) {
-            $why = spl_object_hash($conflictRule);
             $learnedWhy = $this->learnedWhy[$why];
             $problemRules = $this->learnedPool[$learnedWhy];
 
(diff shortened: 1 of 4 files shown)


naderman:
Turns out this also seems to fix long standing issue #7665

williamdes (approved):
I am not familiar with the code base but the diff looks okay

src/Composer/DependencyResolver/Solver.php
```diff
@@ -518,15 +518,19 @@ protected function analyze($level, Rule $rule)
      * @param Problem $problem
      * @param Rule    $conflictRule
      */
-    private function analyzeUnsolvableRule(Problem $problem, Rule $conflictRule)
+    private function analyzeUnsolvableRule(Problem $problem, Rule $conflictRule, &$ruleSeen)
```
stof, line 521: phpdoc needs to be updated for the new argument.

stof: also, this might have an `array` typehint

src/Composer/DependencyResolver/Solver.php
```diff
@@ -518,15 +518,19 @@ protected function analyze($level, Rule $rule)
...
     {
+        $ruleSeen[spl_object_hash($conflictRule)] = true;
+
         if ($conflictRule->getType() == RuleSet::TYPE_LEARNED) {
             $why = spl_object_hash($conflictRule);
```
stof, line 526: as you need the `spl_object_hash` all the time for the key in `$ruleSeen`, you might move that variable at the beginning instead of calling the method multiple times

src/Composer/DependencyResolver/Solver.php
naderman: Haha always worried to do something too modern for PHP 5.3 ;-)

naderman: going to delete the phpdoc as it doesn't state anything beyond the signature anyway.

src/Composer/DependencyResolver/Solver.php
naderman: done

Issue #9308, opened 2020-10-21
Composer 2 doesn't fetch package when it is "provided" by some dependency


andrew-demb:
My `composer.json`:

```json
{
    "name": "gh/issue",
    "require": {
        "php-http/socket-client": "^2"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.0.0-RC2
PHP version: 7.4.11
PHP binary path: /usr/bin/php7.4
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
cURL version: 7.64.0 libz 1.2.11 ssl OpenSSL/1.1.1g
zip extension: not loaded
```

When I run this command:

```
composer update
```

I get the following output:

```
Loading composer repositories with package information
Updating dependencies
Lock file operations: 10 installs, 0 updates, 0 removals
  - Locking nyholm/psr7 (1.3.1)
  - Locking php-http/httplug (2.2.0)
  - Locking php-http/message-factory (v1.0.2)
  - Locking php-http/promise (1.1.0)
  - Locking php-http/socket-client (2.0.1)
  - Locking psr/http-factory (1.0.1)
  - Locking psr/http-message (1.0.1)
  - Locking symfony/deprecation-contracts (v2.2.0)
  - Locking symfony/options-resolver (v5.1.7)
  - Locking symfony/polyfill-php80 (v1.18.1)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 10 installs, 0 updates, 0 removals
  - Installing psr/http-message (1.0.1): Extracting archive
  - Installing php-http/message-factory (v1.0.2): Extracting archive
  - Installing php-http/promise (1.1.0): Extracting archive
  - Installing symfony/polyfill-php80 (v1.18.1): Extracting archive
[...]

PR #9317, opened 2020-10-22
Test: Verify require for package provided or replaced by pkg itself is used

naderman:
Asserts the behavior discussed in https://github.com/composer/composer/issues/9308 so it doesn't change in the future.

### DIFF
--- a/tests/Composer/Test/Fixtures/installer/replacer-satisfies-its-own-requirement.test
+++ b/tests/Composer/Test/Fixtures/installer/replacer-satisfies-its-own-requirement.test
@@ -1,0 +1,25 @@
+--TEST--
+Test that a package requiring something it replaces itself, satisfies itself even though a package exists.
+--COMPOSER--
+{
+    "repositories": [
+        {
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Fixtures/installer/replacer-satisfies-its-own-requirement.test
```diff
@@ -0,0 +1,25 @@
...
+            "package": {
+                "name": "replaced/pkg",
+                "version": "1.0.0"
+            }
+        }
+    ],
+    "require": {
+        "replaced/pkg": "1.0.0"
+    },
+    "provide": {
```
glaubinix, line 17: @naderman shouldn't this be `replace` instead of `provide`?

Seldaek: Yup, fixed in https://github.com/composer/composer/commit/447416a357ee6245d90e8784c82a1054676e17d0

Issue #9311, opened 2020-10-21
Add validation rule warning when provides contains a package name which is also required


stof:
See #9308 for cases where a mistake was done by providing the wrong package. All these cases have in common that their broken `provide` package is also part of their requirement.

AFAICT, there is no valid reason to write such code. If your own package provides the implementation of the other one, you don't need to require it. And if the implementation actually comes from the dependency, your own package *does not* provide it.
I suggest that `composer validate` reports a warning or an error when a package in `provides` also appears in `require` or `require-dev`.

`composer validate` could also report that for packages in `replaces` but this is less an issue: as `replaces` implies a conflict rule, making such mistake there would result in unresolvable dependencies.

Toflar:
> `composer validate` could also report that for packages in `replaces` but this is less an issue: as `replaces` implies a conflict rule, making such mistake there would result in unresolvable dependencies.

Actually it could mess things up in the optimizer. Not sure if it would in the current implementation (https://github.com/composer/composer/pull/9261/files#diff-10d97cb4fcfcb628d9464403d19b36edf65fe37fcac088480022d06bde0c981bR115) but it **could** depending on what we're going to add there. Replacing itself should be prevented as well.

naderman:
Could also generally warn if the package exists, as provide is really meant to be used for "virtual packages", so names only, but not real packages, that's what replace is for.

naderman:
@Toflar issue here is replacing or providing something the package requires, think you're more thinking of replacing the package itself, which is definitely not intended behavior and should error.

boan-jfm:
I think the documentation should be updated too: https://github.com/composer/composer/blob/1.10/doc/04-schema.md#provide

>List of other packages that are provided by this package. This is mostly useful for common interfaces. A package could depend on some virtual logger package, any library that implements this logger interface would simply list it in provide.

It is not clear which convention to use - seems if you provide an implementation of an interface, you should add '-implementation' as suffix to the virtual package name. But how should you know that, based on the documentation?

https://devedge.wordpress.com/2014/09/27/composer-and-virtual-packages/

stof:
@boan-jfm the real need is to have a different name between the package shipping the interface and the virtual package representing implementations. then, the naming does not matter as long as you keep 2 names. The `-implementation` suffix is a common convention, but you could go crazy and do something else for your own package interfaces if you want.

boan-jfm:
[...]

PR #9318, opened 2020-10-22
Validate: Warn about providing or replacing packages you require

naderman:
Fixes https://github.com/composer/composer/issues/9311


stof:
I think this should be merged in 1.x, so that people running `composer validate` will get this warning even if they are still using Composer 1 (as other devs using their packages might already use Composer 2 and be affected)

stof (approved):

PR #9322, opened 2020-10-22
Setting permissions prevents a warning from swapon

TechnologyClassroom:
Warning text:
```
/sbin/swapon /var/swap.1
swapon: /var/swap.1: insecure permissions 0644, 0600 suggested.
```

Adding `/bin/chmod 0600 /var/swap.1` prevents the warning for me on Trisquel 9 (Ubuntu 18.04).


Seldaek:
Thanks

PR #9323, opened 2020-10-22
Fix support of provided packages in the why-not command

### LINKED ISSUE #8415: The why-not command reports wrong blockages when using a * replacement
My `composer.json`:

```json
{
    "require": {
        "symfony/symfony": "3.4.33"
    },
    "replace": {
        "paragonie/random_compat": "*",
        "symfony/polyfill-apcu": "*",
        "symfony/polyfill-ctype": "*",
        "symfony/polyfill-intl-icu": "*",
        "symfony/polyfill-mbstring": "*",
        "symfony/polyfill-php56": "*",
        "symfony/polyfill-php70": "*",
        "symfony/polyfill-php71": "*",
        "symfony/polyfill-php72": "*"
    }
}
```

```
require.symfony/symfony : exact version constraints (3.4.33) should be avoided if the package follows semantic versioning
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys: 
OK
Checking composer version: OK
Composer version: 1.9.1
PHP version: 7.3.11
PHP binary path: /usr/bin/php7.3
```

When I run this command (assuming I already ran a `composer install` before to have a lock file and vendor folder):

```
composer why-not symfony/symfony 4.3.2
```

I get the following output:

```
__root__         -       requires          symfony/symfony (3.4.33)                                                    
symfony/symfony  v4.3.2  requires          symfony/polyfill-ctype (~1.8)                                               
__root__         -       does not require  symfony/polyfill-ctype (but No version set (parsed as 1.0.0) is installed)  
symfony/symfony  v4.3.2  requires          symfony/polyfill-php72 (~1.5)                                               
__root__         -       does not require  symfony/polyfill-php72 (but No version set (parsed as 1.0.0) is installed) 
```

I should see only the `__root__         -       requires          symfony/symfony (3.4.33)  ` line.
`symfony/polyfill-ctype` and `symfony/polyfill-php72` requirements in Symfony are not blocking the upgrade, as my root package is replacing `*`, not `self.version`.

I guess that any replacement involving a range of versions (rather than an exact match) would also be affected, not only `*`, in case the range could intersect with the requirement.

stof:
The version provided by such a rule is defined by the constraint of the rule, not by the package version.

Closes #8415

### DIFF
--- a/src/Composer/Repository/BaseRepository.php
+++ b/src/Composer/Repository/BaseRepository.php
@@ -146,6 +146,16 @@ abstract class BaseRepository implements RepositoryInterface
                         }
 
                         $version = new Constraint('=', $pkg->getVersion());
+
+                        if ($link->getTarget() !== $pkg->getName()) {
+                            foreach (array_merge($pkg->getReplaces(), $pkg->getProvides()) as $prov) {
+                                if ($link->getTarget() === $prov->getTarget()) {
+                                    $version = $prov->getConstraint();
+                                    break;
+                                }
+                            }
+                        }
+
                         if (!$link->getConstraint()->matches($version)) {
                             // if we have a root package (we should but can not guarantee..) we show
                             // the root requires as well to perhaps allow to find an issue there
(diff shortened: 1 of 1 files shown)


Seldaek:
I couldn't repro the original issue anymore so not so sure if it fixes it, but the fix looks reasonable so merging, thanks

stof:
I was still able to reproduce this (with `twig/twig` as the package depending on replaced polyfills rather than `symfony/symfony`)

Issue #9314, opened 2020-10-21
Composer2 and proxies


johnstevenson:
I've been testing proxy support in Composer2 and discovered a few issues. I'll PR something to address these, with suggested solutions outlined below. Comments, criticisms and ideas very welcome.

# 1. curl
Composer2 does not specifically set proxy options for curl, but relies on curl picking up proxy environment variables. Curl uses a `[scheme]_proxy` system to determine if a request should be proxied which leads to inconsistencies when the user has only one variable set, for example:

 * http_proxy: curl will not use a proxy for https requests (Composer streams will)
 * CGI_HTTP_PROXY: curl does not recognize this so will not use a proxy for any requests (Composer streams will)

There are also differences between `no_proxy` IP address handling: curl string-matches the values, while Composer does binary comparison, which allows handling CIDR ranges (similar to golang).

Additionally, in the case of a secure-proxy (see below) curl requires specific certificate options to be set.

### _Solution_
1. Determine if a proxy should be used from the environment (as currently) then set `CURLOPT_PROXY` to either:
    * the proxy url - curl will use this
    * an empty string - curl won't use a proxy or read the environment

2. For secure-proxies, set `CURLOPT_PROXY_CAINFO` or `CURLOPT_PROXY_CAPATH `

Note that curl does not understand `HTTP(S)_PROXY_REQUEST_FULLURI` environment variables, but they are irrelevant as curl uses the full uri where appropriate.

# 2. Secure proxy support
A secure HTTP proxy is one that only accepts encrypted TLS connections, compared to a basic HTTP proxy which accepts unencrypted connections. Both types allow a client to make http and https requests to an origin-server.

The table shows current PHP built-in proxy handling capabilities:

|             | basic-proxy >http | basic-proxy >https | secure-proxy >http | secure-proxy >https |
|-------------|-------------------|--------------------|--------------------|---------------------|
| php streams | Yes               | Yes                | Yes                | No                  | 
| php curl    | Yes               | Yes                | *PHP 7.3 ?          | *PHP 7.3 ?          |

* curl needs the `CURL_VERSION_HTTPS_PROXY` feature (added in curl 7.52.0, Dec 2016). This constant was added in PHP 7.3 which allows checking for this feature.

Although rare, secure-proxies do exist. For _secure-proxy >https_, PHP will error with : `SSL/TLS already set-up for this stream` (see issue #6294) because it only uses a single connection - and this is already TLS enabled from connecting to the proxy. 

### _Solution_
[...]

PR #9324, opened 2020-10-22
Refactor proxy handling for Composer2

johnstevenson:
First attempt at fixing the issues highlighted in https://github.com/composer/composer/issues/9314

* ProxyHelper is a utility class that process the environment variables. I've split this out from StreamContextFactory so it can be more easily used by the installer.
* ProxyManager uses the ProxyHelper to provide the context options for streams, and the proxy url for curl. It's a singleton class but I couldn't find any other way of doing it.

In the rare cases of a secure proxy, PHP streams will fail in the case of an https request (in a similar way that a missing openssl extension does). The curl handling also takes a request based approach - however curl cannot handle http requests either so it could lead to a situation where using streams would have worked.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -41,6 +41,9 @@ class CurlDownloader
     private $authHelper;
     private $selectTimeout = 5.0;
     private $maxRedirects = 20;
+    /** @var ProxyManager */
+    private $proxyManager;
+    private $supportsSecureProxy;
     protected $multiErrors = array(
         CURLM_BAD_HANDLE      => array('CURLM_BAD_HANDLE', 'The passed-in handle is not a valid CURLM handle.'),
         CURLM_BAD_EASY_HANDLE => array('CURLM_BAD_EASY_HANDLE', "An easy handle was not good/valid. It could mean that it isn't an easy handle at all, or possibly that the handle already is in used by this or another multi handle."),
@@ -92,6 +95,11 @@ class CurlDownloader
         }
 
         $this->authHelper = new AuthHelper($io, $config);
+        $this->proxyManager = ProxyManager::getInstance();
+
+        $version = curl_version();
+        $features = $version['features'];
+        $this->supportsSecureProxy = defined('CURL_VERSION_HTTPS_PROXY') && ($features & CURL_VERSION_HTTPS_PROXY);
     }
 
     /**
@@ -176,7 +184,8 @@ class CurlDownloader
         }
 
         $options['http']['header'] = $this->authHelper->addAuthenticationHeader($options['http']['header'], $origin, $url);
-        $options = StreamContextFactory::initOptions($url, $options);
+        // Merge in headers - we don't get any proxy values
+        $options = StreamContextFactory::initOptions($url, $options, true);
 
         foreach (self::$options as $type => $curlOptions) {
             foreach ($curlOptions as $name => $curlOption) {
@@ -187,6 +196,25 @@ class CurlDownloader
                         curl_setopt($curlHandle, $curlOption, $options[$type][$name]);
                     }
                 }
+            }
+        }
+
+        // Always set CURLOPT_PROXY to enable/disable proxy handling
+        // Any proxy authorization is included in the proxy url
+        $proxy = $this->proxyManager->getProxyForRequest($url);
+        curl_setopt($curlHandle, CURLOPT_PROXY, $proxy->getUrl());
+
+        // Curl needs certificate locations for secure proxies.
+        // CURLOPT_PROXY_SSL_VERIFY_PEER/HOST are enabled by default
+        if ($proxy->isSecure()) {
+            if (!$this->supportsSecureProxy) {
+                throw new TransportException('Connecting to a secure proxy using curl is not supported on PHP versions below 7.3.0.');
+            }
+            if (!empty($options['ssl']['cafile'])) {
+                curl_setopt($curlHandle, CURLOPT_PROXY_CAINFO, $options['ssl']['cafile']);
+            }
+            if (!empty($options['ssl']['capath'])) {
+                curl_setopt($curlHandle, CURLOPT_PROXY_CAPATH, $options['ssl']['capath']);
             }
         }
 
@@ -206,7 +234,7 @@ class CurlDownloader
             'reject' => $reject,
         );
 
-        $usingProxy = !empty($options['http']['proxy']) ? ' using proxy ' . $options['http']['proxy'] : '';
+        $usingProxy = $proxy->getLastProxy(' using proxy (%s)');
         $ifModified = false !== stripos(implode(',', $options['http']['header']), 'if-modified-since:') ? ' if modified' : '';
         if ($attributes['redirects'] === 0) {
             $this->io->writeError('Downloading ' . Url::sanitize($url) . $usingProxy . $ifModified, true, IOInterface::DEBUG);
--- a/src/Composer/Util/StreamContextFactory.php
+++ b/src/Composer/Util/StreamContextFactory.php
@@ -15,6 +15,7 @@ namespace Composer\Util;
 use Composer\Composer;
 use Composer\CaBundle\CaBundle;
 use Composer\Downloader\TransportException;
+use Composer\Util\Http\ProxyManager;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -57,10 +58,11 @@ final class StreamContextFactory
     /**
      * @param string $url
      * @param array $options
+     * @param bool $forCurl
      * @psalm-return array{http:{header: string[], proxy?: string, request_fulluri: bool}, ssl: array}
      * @return array formatted as a stream context array
      */
-    public static function initOptions($url, array $options)
+    public static function initOptions($url, array $options, $forCurl = false)
     {
         // Make sure the headers are in an array form
         if (!isset($options['http']['header'])) {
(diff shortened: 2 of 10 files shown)


src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -190,6 +200,26 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
             }
         }
 
+        // Always set CURLOPT_PROXY to enable/disable proxy handling
+        // Any proxy authorization is included in the proxy url
+        $proxyUrl = $this->proxyManager->getCurloptProxy($url, $secureProxy);
```
Seldaek, line 205: Could this use `list($proxyUrl, $isSecureProxy) = ...` kind of pattern instead of references? References are really not helping readability, it might make sense when you wrote the code but here I just have to guess at what is going on..

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -206,8 +236,8 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
             'reject' => $reject,
         );
 
-        $usingProxy = !empty($options['http']['proxy']) ? ' using proxy ' . $options['http']['proxy'] : '';
-        $ifModified = false !== stripos(implode(',', $options['http']['header']), 'if-modified-since:') ? ' if modified' : '';
+        $usingProxy = $this->proxyManager->getLastProxy(' using proxy (%s)');
+        $ifModified = false !== strpos(strtolower(implode(',', $options['http']['header'])), 'if-modified-since:') ? ' if modified' : '';
```
Seldaek, line 240: Why the change from stripos? This doesn't seems like an improvement?

src/Composer/Util/ProxyManager.php
```diff
@@ -0,0 +1,224 @@
...
+        $info = array();
+
+        if ($httpProxy) {
+            $this->setData($httpProxy, 'http', $info);
+        }
+        if ($httpsProxy) {
+            $this->setData($httpsProxy, 'https', $info);
+        }
+        if ($this->hasProxy) {
+            $this->info = implode(', ', $info);
```
Seldaek, line 143: Generally here too this is super hard to read.. I have no idea what's in $info, I look up and see it's again some by ref argument.. Why isn't setData returning the $info data? Why is it called $info if it represents the safe proxy url? `$safeProxyUrl = $this->setData(...)` would be clearer I'd think.

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -206,8 +236,8 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
             'reject' => $reject,
         );
 
-        $usingProxy = !empty($options['http']['proxy']) ? ' using proxy ' . $options['http']['proxy'] : '';
-        $ifModified = false !== stripos(implode(',', $options['http']['header']), 'if-modified-since:') ? ' if modified' : '';
+        $usingProxy = $this->proxyManager->getLastProxy(' using proxy (%s)');
```
Seldaek, line 239: I was fine with a singleton but now it looks like it's a stateful one too, which makes things really hard to follow and bug prone IMO.

How about converting ProxyManager to something like a ProxyConfig class, which an instance of is created for every request here, it can have a static var for the constant config/state stuff so it doesn't need to recompute it all on every request, but IMO it'd make things much easier, would allow it to have simple getters for values which are sure to be linked to the current URL being processed, no state changes, no by-reference args.

src/Composer/Util/ProxyHelper.php
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        if (isset($parts['user'])) {
+            $parts['user'] = '***';
+        }
+        if (isset($parts['pass'])) {
+            $parts['pass'] = '***';
+        }
+
+        return self::formatParsedUrl($parts, $includeAuth = true);
```
Seldaek, line 80: I think this whole method could be replaced by `Composer\Util\Url::sanitize($proxyUrl)`

Seldaek (changes requested):

Seldaek:
Also please move those new Proxy* classes into Composer\Util\Http namespace I think it'd make more sense as they are internal.

src/Composer/Util/ProxyManager.php
johnstevenson: Yeah, it's rubbish . I had to remind myself what it is (which is a scheme=safeUrl pair). I'm not sure it's even needed now.

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Hmm, I must have screwed up a rebase merge confilct (I hadn't realised this line had change)

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Yes, I agree.

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -190,6 +200,26 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
...
+        // Always set CURLOPT_PROXY to enable/disable proxy handling
+        // Any proxy authorization is included in the proxy url
+        $proxyUrl = $this->proxyManager->getCurloptProxy($url, $secureProxy);
+        curl_setopt($curlHandle, CURLOPT_PROXY, $proxyUrl);
+
+        // Curl needs certificate locations for secure proxies.
+        // CURLOPT_PROXY_SSL_VERIFY_PEER/HOST are enabled by default
+        if ($proxyUrl && $secureProxy) {
+            if (!$this->supportsSecureProxy) {
+                $message = 'Connecting to a secure proxy using curl is not supported on PHP versions below 7.3.0.';
```
stof, line 212: why using a variable for the message instead of inlining it ?

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -176,7 +185,8 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
         }
 
         $options['http']['header'] = $this->authHelper->addAuthenticationHeader($options['http']['header'], $origin, $url);
-        $options = StreamContextFactory::initOptions($url, $options);
+        // Merge in headers, $forCurl ensures that we don't get any proxy values
+        $options = StreamContextFactory::initOptions($url, $options, $forCurl = true);
```
stof, line 189: why defining a `$forCurl` local variable here ? That's useless as you don't use it.

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Yes that makes more sense: perhaps the ProxyManager could become a ProxyFactory, which stores the state and creates a ProxyConfig (or something) class per request. I originally designed it to be session based, with the HttpDownloader determining if curl is compatible - and using streams if it isn't.

A properly implemented request based approach will make things cleaner - with the proviso that folks using a secure-proxy need PHP 7.3+ and curl (and won't be able to make http requests any more).

src/Composer/Util/ProxyHelper.php
johnstevenson: Will do.

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Agree, there is no need for that (the message was a bit more complicated at one point involving curl versions, until I discovered that even PHP 5.6 supports the version but not the feature)

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: To make a boolean flag more descriptive and the comment easier to write/understand, but I can remove it if you want?

stof: But then, phpstan (and IDEs) would complain about an unused variable (phpstan does not complain right now because composer runs it at a very low level and this rule is at a higher level).
Regarding the behavior of `$forCurl == true`, the proper place to document its effect would be in the phpdoc of `StreamContextFactory::initOptions` IMO, not on the usage

src/Composer/Util/ProxyHelper.php
johnstevenson: Oops, that last reply was intended to be to "Also please move those new Proxy* classes into Composer\Util\Http namespace I think it'd make more sense as they are internal."

I had hoped that it would do the job, but it doesn't work without a scheme. Incidentally, the ProxyHelper version masks the user but Url::sanitize does not - so is it safe (good practice) to show the user part ?

Seldaek: Well generally speaking showing the user can help people spot misconfiguration I'd think. Feel free to tweak the sanitize method so it handles URLs without scheme tho. I guess turning `://` into `(://|^)` in the regex and then returning $m[1] instead of :// should do.

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Thanks for explaining, I hadn't realised the phpstan bit.

johnstevenson: Resolved in Code review:1 fixes

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Resolved in Code review:1 fixes

src/Composer/Util/ProxyHelper.php
johnstevenson: Resolved in Code review:1 fixes

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Resolved in Code review:1 fixes

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Resolved in Refactor ProxyManager

src/Composer/Util/Http/CurlDownloader.php
johnstevenson: Resolved in Refactor ProxyManager

src/Composer/Util/ProxyManager.php
johnstevenson: Resolved in Refactor ProxyManager

johnstevenson:
@Seldaek Okay, we're almost there. I need to add tests to the `RequestProxy` class (terrible name) but the core data is already covered so these can wait.

I'm going to spend a short time running it through my secure-proxy implementation, then I'll rebase it all. Shout now if you want anything changing or spot something that is glaringly awful or just plain wrong.

johnstevenson:
Squashed and rebased. It appears to work well running it through a proxy.

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -176,7 +184,8 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
         }
 
         $options['http']['header'] = $this->authHelper->addAuthenticationHeader($options['http']['header'], $origin, $url);
-        $options = StreamContextFactory::initOptions($url, $options);
+        // Merge in headers - we don't get any proxy values
```
Seldaek, line 187: Not sure I understand what this means.. We don't get any proxy values? And what merging of headers?

src/Composer/Util/StreamContextFactory.php
```diff
@@ -70,76 +72,26 @@ public static function initOptions($url, array $options)
...
-                    break;
+        // Add stream proxy options if there is a proxy
+        if (!$forCurl) {
+            $proxy = ProxyManager::getInstance()->getProxyForRequest($url);
+            if ($proxy->isSecure()) {
+                if (!extension_loaded('openssl')) {
+                    throw new TransportException('You must enable the openssl extension to use a proxy over https.');
+                }
+                if (0 === strpos($url, 'https://')) {
+                    throw new TransportException('PHP does not support https requests to a secure proxy.');
```
Seldaek, line 83: Maybe suggest enabling the curl extension, and specify that PHP **streams** do not support this? Generally speaking the more verbose and helpful the errors are the better, if they allow people to fully understand or at least fix the problem without reaching out to us.

Seldaek (approved):
Looks fine overall so merging, still left a couple comments and fixed a couple things myself in 62eff8e979a5adeffc2b5490f581d5ea5bbf008e

PR #9326, opened 2020-10-23
Docs: Remove unnecessary uses of simple/easy

naderman:
They only serve to make anyone reading the docs who doesn't find something as simple or easy as stated feel bad about themselves, they don't add anything valuable to the docs in these cases.


doc/faqs/why-are-unbound-version-constraints-a-bad-idea.md
```diff
@@ -16,6 +16,6 @@ For example instead of using `>=3.4` you should use `~3.4` which allows all
 versions up to `3.999` but does not include `4.0` and above. The `^` operator
 works very well with libraries following [semantic versioning](https://semver.org).
 
-**Note:** As a package maintainer, you can make the life of your users easier
+**Note:** As a package maintainer, you can make the help your users
```
guilliamxavier, line 19: fixed in #9347

PR #9330, opened 2020-10-24
Escape special characters in Markdown files

fredden:
Some characters in Markdown have special meaning and should be escaped if used in text.

<details><summary>Example</summary>

This screen-shot was taken from https://getcomposer.org/changelog/2.0.0-RC1 at time of authorship. Note the second bullet has some unintentionally-emphasised text.

Screenshot_2020-10-24_11-53-45

</details>


Seldaek:
IMO the better way would be to quote those as `ext-*` for example.

Seldaek:
Thanks

PR #9334, opened 2020-10-24
Improve proxy error messages for streams

johnstevenson:
I got my logic wrong, so this needs fixing. The exception messages have also been improved (hopefully).

Previously, this threw a RuntimeException but I changed it to a TransportException to match what I was using for secure-proxy >https not being supported natively.

However, there is also a NoSslException, which looks like it is only specifically used when the application first runs. So what should I be using for no openssl situations?

Also, I could add tests for this, but it seems a bit pointless unless the tests explicity run without openssl. What should I do.

### DIFF
--- a/src/Composer/Util/StreamContextFactory.php
+++ b/src/Composer/Util/StreamContextFactory.php
@@ -58,7 +58,7 @@ final class StreamContextFactory
     /**
      * @param string $url
      * @param array $options
-     * @param bool $forCurl
+     * @param bool $forCurl When true, will not add proxy values as these are handled separately
      * @psalm-return array{http:{header: string[], proxy?: string, request_fulluri: bool}, ssl: array}
      * @return array formatted as a stream context array
      */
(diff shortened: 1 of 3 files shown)


src/Composer/Util/StreamContextFactory.php
```diff
@@ -75,13 +75,17 @@ public static function initOptions($url, array $options, $forCurl = false)
...
-                    throw new TransportException('You must enable the openssl extension to use a proxy over https.');
+                    throw new TransportException('You must enable the openssl extension to use a secure proxy.');
                 }
-                if (0 === strpos($url, 'https://')) {
-                    throw new TransportException('PHP does not support https requests to a secure proxy.');
+                if ($isHttpsRequest) {
+                    throw new TransportException('You must enable the curl extension to make https requests through a secure proxy.');
                 }
+            } elseif ($isHttpsRequest && !extension_loaded('openssl')) {
+                throw new TransportException('You must enable the openssl extension to make https requests through a proxy.');
```
Seldaek, line 88: This code will actually be reached even if no proxy is defined unless I'm mistaken, so maybe it should check before mentioning the proxy in the exception message. I guess https even without proxy will fail later on anyway if openssl is missing.

Seldaek:
Regarding exception classes, IMO it's fine if all the HttpDownloader throws is TransportException, kinda more consistent that way.

src/Composer/Util/StreamContextFactory.php
johnstevenson: Ah you're right (and I'm still getting my logic wrong!). I'll change this to only be effective with a proxy (as per the original). Okay?

Seldaek: Sure that sounds fine to me :)

johnstevenson:
>  it's fine if all the HttpDownloader throws is TransportException

Is it worth annotating that the NoSslException exception is only used for the start up then (if that is the intended case)?

Seldaek:
Sure if you want to add that I never argue with more docs :)

Seldaek:
Thanks

PR #9336, opened 2020-10-24
Correctly avoid partial updates in require commands

helhum:
The installer does not support partial updates, when
no lock file is present. The check in require command
to prevent a partial update however does not check whether
a lock file is present, but whether a lock file should be written,
which most likely always is true.

This check is now changed to check if a lock file is actually present.

### DIFF
--- a/src/Composer/Command/RequireCommand.php
+++ b/src/Composer/Command/RequireCommand.php
@@ -314,7 +314,7 @@ EOT
 
         // if no lock is present, or the file is brand new, we do not do a
         // partial update as this is not supported by the Installer
-        if (!$this->firstRequire && $composer->getConfig()->get('lock')) {
+        if (!$this->firstRequire && $composer->getLocker()->isLocked()) {
             $install->setUpdateAllowList(array_keys($requirements));
         }
 


helhum:
Steps to reproduce:

1. Have a composer.json with at least one package and no composer.lock
1. Require any other (existing) package e.g. `composer req composer/composer`

Expected result: composer.json changed, composer.lock created
Actual result: Error message: `Cannot update only a partial set of packages without a lock file present`

naderman:
This should probably also get a test case, but merging, so at least snapshots have this fixed.

helhum:
Thanks! Will look into command tests...

naderman:
@helhum I mean thanks if you want to do that, but it's not exactly easy to do that currently, so maybe just something to leave for another day ;-)

helhum:
@naderman any hints where I should put the test? Installer integration tests seem to currently only support install and update, and in functionals I didn't find a way to provide a composer.json as a starting point. Thanks

helhum:
> @helhum I mean thanks if you want to do that, but it's not exactly easy to do that currently, so maybe just something to leave for another day ;-)

Ah, haven't seen your comment. Invested some time trying to implement such test and decided to indeed leave it for another day :)

tomasnorre:
Thanks for fixing this. I can confirm that it solved my issue. 
https://stackoverflow.com/questions/64525816/problem-with-travisci-version-matrix-after-updating-to-composer-2-0

PR #9339, opened 2020-10-25
Fix plugin install order for plugins modifying downloads

### LINKED ISSUE #9333: Prioritizing local Composer plugins
Hi Composer,

> ℹ️ As per @Seldaek's recommendation, this continues the discussion started in ffraenz/private-composer-installer#25.

With the introduction of Composer 2.0, a number of specialty Composer plugins are having trouble supporting the new architecture.

These Composer plugins offer a way to install private packages that do not support Composer and, more specifically, do not use standard authentication methods. Instead, they rely on URIs with sensitive credentials.

<details><summary>Example 1.0</summary>

```json
{
  "type": "package",
  "package": {
    "name": "package-name/package-name",
    "version": "1.0.0",
    "dist": {
      "type": "zip",
      "url": "https://example.com/package-name.zip?key={%PACKAGE_KEY}&version={%VERSION}"
    },
    "require": {
      "ffraenz/private-composer-installer": "^4.0"
    }
  }
}
```
</details>

Most Composer plugins catering to these private packages outsource the sensitive parts of these URIs into environment variables or a `.env` file typically ignored by version control.

For some packages, the actual URI to the archive is hidden behind APIs (like [Easy Digital Downloads](https://docs.easydigitaldownloads.com/article/384-software-licensing-api)). The download URI is often limited to the latest version of the package _and_ often expirable.

During a download operation, the plugin replaces the placeholders in the URI with the sensitive values or fetches the URI from the intermediate API. The lock file will often store a variant of this URI that contains only the replaced version number in order to maintain a consistent cache key.

<details><summary>Example 2.0</summary>

```json
"dist": {
  "type": "zip",
  "url": "https://example.com/package-name.zip?key={%PACKAGE_KEY}&version=1.0.0"
}
```
</details>

In Composer v1, this has worked rather well because of its nonconcurrent architecture where plugins would be installed first to be available to mutate the private packages' processed URIs.

With Composer v2, the operations are executed in batches, plugins first then the rest. But the downloads are executed all at once before the installs are executed. Composer will obviously fail when it hits this special placeholder-filled URI since the Composer plugin is often not available yet.

---

Currently, the possible solutions available are:

1. Install the Composer plugin globally but this isn't the most practical solution for a group of users on a project, they need to be aware of what global plugin to install (a crude example, I know).
2. Build a private Composer repository but not everyone can build and maintain one and this essentially kills these plugins.

@Seldaek has proposed to tweak the Composer install process to allow plugins to mark themselves as having to be installed ahead of time.

Being able to prioritize the availability of local Composer plugins might also be beneficial beyond URI replacements, obviously.

Thanks

---

Concerned plugins:

- [ffraenz/private-composer-installer](https://github.com/ffraenz/private-composer-installer)
- [junaidbhura/composer-wp-pro-plugins](https://github.com/junaidbhura/composer-wp-pro-plugins)
- [philippbaschke/acf-pro-installer](https://github.com/PhilippBaschke/acf-pro-installer)
- [pivvenit/acf-pro-installer](https://github.com/pivvenit/acf-pro-installer)

Parties involved: @ffraenz, @justinrainbow @Qrious

Seldaek:
Fixes #9333

Still needs documentation.

### DIFF
--- a/src/Composer/DependencyResolver/Transaction.php
+++ b/src/Composer/DependencyResolver/Transaction.php
@@ -248,9 +248,9 @@ class Transaction
      */
     private function movePluginsToFront(array $operations)
     {
-        $dlModyingPluginsNoDeps = array();
-        $dlModyingPluginsWithDeps = array();
-        $dlModyingPluginRequires = array();
+        $dlModifyingPluginsNoDeps = array();
+        $dlModifyingPluginsWithDeps = array();
+        $dlModifyingPluginRequires = array();
         $pluginsNoDeps = array();
         $pluginsWithDeps = array();
         $pluginRequires = array();
@@ -267,7 +267,7 @@ class Transaction
             $isDownloadsModifyingPlugin = $package->getType() === 'composer-plugin' && ($extra = $package->getExtra()) && isset($extra['plugin-modifies-downloads']) && $extra['plugin-modifies-downloads'] === true;
 
             // is this a downloads modifying plugin or a dependency of one?
-            if ($isDownloadsModifyingPlugin || count(array_intersect($package->getNames(), $dlModyingPluginRequires))) {
+            if ($isDownloadsModifyingPlugin || count(array_intersect($package->getNames(), $dlModifyingPluginRequires))) {
                 // get the package's requires, but filter out any platform requirements
                 $requires = array_filter(array_keys($package->getRequires()), function ($req) {
                     return !PlatformRepository::isPlatformPackage($req);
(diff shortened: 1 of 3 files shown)


src/Composer/DependencyResolver/Transaction.php
```diff
@@ -248,6 +248,9 @@ protected function getProvidersInResult(Link $link)
      */
     private function movePluginsToFront(array $operations)
     {
+        $dlModyingPluginsNoDeps = array();
```
naderman, line 251: Typo: Modifying

naderman:
Looks good to me except for the variable name typo.

mcaskill (approved):
This looks good and works well in both junaidbhura/composer-wp-pro-plugins#24 and ffraenz/private-composer-installer#25 (barring a fix for the cache key).

The typos in `Transaction.php` are the only thing I've noticed.

src/Composer/Installer/InstallationManager.php
```diff
@@ -397,15 +428,20 @@ private function executeBatch(RepositoryInterface $repo, array $operations, arra
...
-            $this->loop->wait($promises, $progress);
-            if ($progress) {
-                $progress->clear();
-            }
+        if (count($promises)) {
+            $this->waitOnPromises($promises);
+        }
+    }
+
+    private function waitOnPromises(array $promises)
```
simonberger, line 436: I know the code did not change but in my opinion this is better readable. Partly redundant code vs additional checks ...

```
if ($this->outputProgress && $this->io instanceof ConsoleIO && !$this->io->isDebug() && count($promises) > 1) {
   $progress = $this->io->getProgressBar();
   $this->loop->wait($promises, $progress);
   $progress->clear();
} else {
    $this->loop->wait($promises);
}
```

Seldaek: Yeah, opinions.. :)

PR #9341, opened 2020-10-25
Update authentication-for-private-packages.md

schramm-michael:
Typo on line 278
changed cosumer... to consumer-key


Seldaek:
Thanks

PR #9345, opened 2020-10-26
Fix error message

fabpot:


src/Composer/Command/InitCommand.php
```diff
@@ -741,17 +741,17 @@ private function findBestVersionAndNameForPackage(InputInterface $input, $name,
...
+                    $requiredVersion ? ' at version '.$requiredVersion : ''
                 ));
             }
             // Check whether the required version was the problem
             if ($requiredVersion && $versionSelector->findBestCandidate($name, null, $preferredStability, $ignorePlatformReqs)) {
                 throw new \InvalidArgumentException(sprintf(
                     'Could not find package %s in a version matching %s',
                     $name,
-                    $requiredVersion
+                    $requiredVersion ?: '*'
```
stof, line 754: isn't this always true due to the `if ($requiredVersion)` ?

naderman: Indeed, reverted with https://github.com/composer/composer/commit/289972a177108aaf0be7dbb000e8b47c54cdac7c

Issue #7736, opened 2018-10-19
composer self-update/selfupdate command is missing


crisps-for-breakfast:
Output of `composer diagnose`:

```
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Composer version: 1.7.2
PHP version: 7.1.23
PHP binary path: /usr/bin/php
```

When I run this command:

```
composer self-update
```

I get the following output:

```
 [Symfony\Component\Console\Exception\CommandNotFoundException]  
  Command "self-update" is not defined.
```

And I expected composer to update itself. Running `composer list` also does not show the self-update/selfupdate commands. All other commands appear to be working normally.

I tried deleting my `/home/{username}/.local/bin/composer` executable and reinstalling using the php commands listed at https://getcomposer.org/download/

I get the same problem

I have another server with the exact same version of composer in the exact same location which has self-update/update commands listed and working as expected

xabbuh:
It looks like you also have a version of Composer installed using your distribution's package manager which is then executed when typing `composer`. `which composer` will probably give you a bit more information.

crisps-for-breakfast:
Thanks that was very helpful! 

You were indeed correct, I had to use `yum remove composer`, repeat the installation from  https://getcomposer.org/download/, and then restart my ssh session to get the commands back.

`which composer` was pointing at `/usr/bin/composer` instead

It'd be nice if `composer diagnose` or `composer --version` differentiated between the two different types

xabbuh:
I think that's something you need to file as a feature request to the maintainers of the RPM package. It's managed by other maintainers than Composer itself.

crisps-for-breakfast:
Thanks, that makes sense

PR #9351, opened 2020-10-26
Document availability of self-update command

lucaswerkmeister:
The self-update command is only available when running as a PHAR (see commit 761ad6d171); notably, it is not defined in the Ubuntu package.

Screenshot from 2020-10-26 13-04-07

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -607,6 +607,9 @@ you may have to run the command with `root` privileges
 sudo -H composer self-update
 ```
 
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
> notably, it is not defined in the Ubuntu package.

Sounds like Yum-based distributions (Fedora? RHEL?) have the same problem, see #7736.

Seldaek:
Yup, and as such I'd not recommend installing Composer using OS package managers as it then gets quickly outdated.

PR #9353, opened 2020-10-26
Fix PHPDoc return type for BaseCommand::getComposer

kubawerlos:
The `null` will never be returned, if the Composer instance cannot be created then exception will be thrown.

Because of type `Composer|null` PHPStan return an error:

> Cannot call method getConfig() on Composer\Composer|null.

_I imagine in the future (after dropping PHP 5 support) wrong PHPDoc type can result in the wrong return type._


Seldaek:
Only if $required is true do we throw, so null can be returned.

PR #9360, opened 2020-10-27
PoolBuilder: Ensure versions matching locked constraints get loaded

naderman:
This fixes an issue experienced by user as an unexplainable conflict:

```
Updating dependencies

  Problem 1
    - root/req1 is locked to version 1.0.0 and an update of this package was not requested.
    - root/req1 1.0.0 requires dep/pkg1 1.* -> found dep/pkg1[1.0.0] but these were not loaded, likely because it conflicts with another require.
```

The explanation "because it conflicts with another require" is false in this case. The package simply didn't get loaded into the pool because of a bug.

Fixes https://github.com/composer/composer/issues/9489

### DIFF
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -271,7 +271,7 @@ class Problem
                 return array("- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose).' in lock file but not in remote repositories, make sure you avoid updating this package to keep the one from lock file.');
             }
 
-            return array("- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose).' but '.(self::hasMultipleNames($packages) ? 'these conflict' : 'it conflicts').' with another require.');
+            return array("- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose).' but these were not loaded, likely because '.(self::hasMultipleNames($packages) ? 'they conflict' : 'it conflicts').' with another require.');
         }
 
         // check if the package is found when bypassing stability checks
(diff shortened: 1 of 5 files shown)


src/Composer/DependencyResolver/Problem.php
```diff
@@ -271,7 +271,7 @@ public static function getMissingPackageReason(RepositorySet $repositorySet, Req
                 return array("- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose).' in lock file but not in remote repositories, make sure you avoid updating this package to keep the one from lock file.');
             }
 
-            return array("- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose).' but '.(self::hasMultipleNames($packages) ? 'these conflict' : 'it conflicts').' with another require.');
+            return array("- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose).' but these were not loaded, likely because '.(self::hasMultipleNames($packages) ? 'they conflict' : 'it conflicts').' with another require.');
```
stof, line 274: should this say `it was not loaded` in singular ?

naderman: depends on singular/plural, so will have to adjust, yes.

Toflar:
PR for a fix: https://github.com/naderman/composer/pull/2

naderman:
Made the test cases a bit broader to point out an issue in @Toflar 's first attempt at a fix and made one myself now. However I'm not sure yet if it also needs some adjustments for replace/provide or if that doesn't matter here. Will give that some more thought tomorrow.

stof:
@naderman what is left here ?

naderman:
@stof my last commit fixes some more bugs that I need test cases for. Additionally I think there are further issues with the pool builder and partial updates involving replace/provide which I need to come up with test cases and fixes for.

naderman:
Turns out that last commit cannot be verified by an additional test as it doesn't fix a bug, as markPackagesForLoading caught the incorrect call by checking loadedPackages on fixed packages. Either way, it reduces some function calls in the refactored versions and still passes all existing tests.

naderman:
Okay so I split off additional failing tests related to unlocking of providers and replacers to https://github.com/composer/composer/pull/9538 and this here should fix the main issue reported now and we can do a release with that. The other PR is a bit of an endless rabbit hole it seems, so can do that later.

PR #9374, opened 2020-10-29
Remove fulluri checks and tidy diagnose output

johnstevenson:
There is now no need for `checkHttpProxyFullUriRequestParam()` and `checkHttpsProxyFullUriRequestParam()` because we don't use the corresponding environment variables any more.

I've removed the lines below since I can't see how the condition can exist, because `Factory::createHttpDownloader()` will have already thrown a terminating NoSslException in this situation. Likewise it also throws a TransportError if a CA file cannot be found. 

https://github.com/composer/composer/blob/ec5f02018643ccd8a1d7f8ae41f3c6d8e01dab78/src/Composer/Command/DiagnoseCommand.php#L229-L231

I've reordered the tls warning, otherwise it "hides" an exception by reporting it as a WARNING rather than a FAIL.


Seldaek:
Thanks

PR #9378, opened 2020-10-29
Fix running tests with non standard default branch

studioromeo:
When the global git config has `init.defaultbranch` set to something other than master the tests fail. This is because the tests assume that the default branch is called master but that may not be the case.

```
There were 3 errors:

1) Composer\Test\Package\Archiver\ArchiveManagerTest::testArchiveTar
RuntimeException: Failed to execute git checkout 'master' -- && git reset --hard 'master' --

fatal: invalid reference: master
```

This PR fixes the tests by ensuring that when running git init we ensure the default branch is called master regardless of how git is configured.

### DIFF
--- a/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
+++ b/tests/Composer/Test/Package/Archiver/ArchiveManagerTest.php
@@ -115,7 +115,7 @@ class ArchiveManagerTest extends ArchiverTest
         chdir($this->testDir);
 
         $output = null;
-        $result = $this->process->execute('git init -q', $output, $this->testDir);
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

Seldaek:
Hm, this breaks for me on my local env because -b does not exist on git init there. Any clue since when that is available? Might need to conditionally add it.

studioromeo:
@Seldaek oh crumbs sorry! I'm not sure exactly but I'll raise another pr shortly and do `git checkout -b master` instead

Seldaek:
Thanks, no biggie :)

PR #9380, opened 2020-10-30
Ignore '__MACOSX' folders while looking for a composer.json in zip artifacts

wissem:
The built-in utility for creating ZIP files in Mac OSX creates a folder named "__MACOSX" which contains a replication of the zipped directory. This causes the utility class that looks for a composer.json to fail since it becomes ambiguous to find a composer.json.


wissem:
Closing because it's a duplicate of https://github.com/composer/composer/pull/9379

Issue #9381, opened 2020-10-30
handshake failure with self-signed cert on composer 2.0.3


martyrs:
I just upgraded to composer 2.0.3 and my self signed cert is not working anymore.
Rolling back to 1.10.5 solves the issue without any changes.

My `composer.json`:

```json
  "repositories": [
    {
      "type": "composer",
      "url": "https://repo.xyz.com",
      "options": {
        "ssl": {
          "verify_peer": false,
          "allow_self_signed": true,
          "local_cert": "/etc/openssl/project/certs/composer.pem"
        }
      }
    }
  ],

```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.0.3
PHP version: 7.4.11
PHP binary path: /usr/bin/php7.4
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
cURL version: 7.58.0 libz 1.2.11 ssl OpenSSL/1.1.1g
zip extension: OK

```

When I run this command:

```
composer install -vvv
```

I get the following output:

```
Reading ./composer.json
Loading config file /home/tamer/.config/composer/auth.json
Loading config file ./composer.json
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/home/tamer/project/tlcomposerv2): git branch -a --no-color --no-abbrev -v
Failed to initialize global composer: Composer could not find the config file: /home/tamer/.config/composer/composer.json
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
Running 2.0.3 (2020-10-28 15:50:55) with PHP 7.4.11 on Linux / 5.4.0-52-generic
No lock file found. Updating dependencies instead of installing from lock file. Use composer update over composer install if you do not have a lock file.
Loading composer repositories with package information
Updating dependencies
Downloading https://repo.xzy.com/packages.json
[...]

PR #9384, opened 2020-10-30
Update CurlDownloader enable support for client ssl-certificates

rtm-ctrlz:
`сomposer-1` used for http requests `RemoteFilesystem` with context, with passed directly from `options` section of repository and there was not problems to pass client ssl-certificates to `stream-context`.

Upgrading to `composer-2` (with `ext-curl` installed) causes HTTP-requests to fail - `CurlDownloader` doesn't use client ssl-certificates.

Sample `config.json`:
```json
{
    "repositories": {
        "private-repo": {
            "type": "composer",
            "url": "https://private.repo.example.com",
            "options": {
                "ssl": {
                    "local_cert": "/users/user/.composer/certs.d/pravate-repo-user.cert",
                    "local_pk": "/users/user/.composer/certs.d/pravate-repo-user.key"
                }
            }
        }
    }
}
```

There is no need (for me) to store `passphrase` at `config.json`, but I think - let's make it possible to use it by anyone who needs this.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -63,6 +63,9 @@ class CurlDownloader
             'capath' => CURLOPT_CAPATH,
             'verify_peer' => CURLOPT_SSL_VERIFYPEER,
             'verify_peer_name' => CURLOPT_SSL_VERIFYHOST,
+            'local_cert' => CURLOPT_SSLCERT,
+            'local_pk' => CURLOPT_SSLKEY,
+            'passphrase' => CURLOPT_SSLKEYPASSWD,
         ),
     );
 


Seldaek:
Thanks, have you actually verified that the code works for you?

johnstevenson:
Do we need to set `CURLOPT_SSLCERTTYPE` to `'PEM'` as well, because it is not the default on macs? 

https://curl.haxx.se/libcurl/c/CURLOPT_SSLCERT.html
https://curl.haxx.se/libcurl/c/CURLOPT_SSLCERTTYPE.html

Seldaek:
Where do you see this @johnstevenson ? Both https://www.php.net/manual/en/function.curl-setopt.php and the link you posted show PEM as default.

rtm-ctrlz:
> Thanks, have you actually verified that the code works for you?

Yes, I made couple tests without any changes in `config.json` (it's almost like example above):
 - disabled `ext-curl` to test `RemoteFilesystem` and rest of updated `composer` in my main setup - it works as expected - some kind of regression check
 - patched `CurlDownloader` in my main setup - it works as expected - PR check
 - patched `CurlDownloader` in test setup (self-signed certificates + password encrypted private key) - it works as expected - PR check 2

> Do we need to set `CURLOPT_SSLCERTTYPE` to `'PEM'` as well, because it is not the default on macs?

I'm not sure, PR code works on my mac (Catalina) without setting this option.

johnstevenson:
> Supported formats are "PEM" and "DER", except with Secure Transport. OpenSSL (versions 0.9.3 and later) and Secure Transport (on iOS 5 or later, or OS X 10.7 or later) also support "P12" for PKCS#12-encoded files. 

See also: https://rapidportal.intercede.com/docs/RapID/testing/usingcurl/ which seems to suggest that PHP could be built with either openssl or Apple's SecureTransport.

`php -i | grep "SSL Version"` 
SSL Version => SecureTransport or SSL Version => OpenSSL/1.0.2l

johnstevenson:
Additionally, concerning `CURLOPT_SSLKEY`, from https://curl.haxx.se/libcurl/c/CURLOPT_SSLKEY.html

> ... The default format is "PEM" and can be changed with CURLOPT_SSLKEYTYPE.
(iOS and Mac OS X only) This option is ignored if curl was built against Secure Transport. Secure Transport expects the private key to be already present in the keychain or PKCS#12 file containing the certificate. 

@Seldaek Perhaps you could incorporate a SecureTransport check in the new `canUseCurl()` method?

Seldaek:
@johnstevenson "also supports" doesn't mean the default is different there? I think it's fine as is, we don't let you change the default so you have to use PEM but that sounds like an ok limitation for now. 

As for SecureTransport not sure what you're suggesting, I'm assuming this is an implementation detail and that it's abstracted away by curl options?

Seldaek:
@johnstevenson happy to keep the discussion going if needed here, but already merging this so I can move forward with stuff :)

@rtm-ctrlz thanks

johnstevenson:
@Seldaek I've got my head aroud it now, so:

The libcurl docs imply that you need to set `CURLOPT_SSLKEY` in conjunction with `CURLOPT_SSLCERT` which expect pem as the default, unless you're using libcurl with SecureTransport in which case curl expects `CURLOPT_SSLCERT` in P12 format and will ignore any `CURLOPT_SSLKEY` value.

> As for SecureTransport not sure what you're suggesting, I'm assuming this is an implementation detail and that it's abstracted away by curl options?

For options that specify a file, PHP generally just sets whatever you give it. Inconsistently, it applies an an openbasedir check in some cases.

Apologies for the noise before - I certainly would not wish to hold back the wheels of progress!

rtm-ctrlz:
Just checked patched (with this PR) with setting only `local_cert` (stream option mapping for `CURLOPT_SSLCERT`) with a file that contains both certificate and private key and it worked without any notice/warning/etc.

```bash
# macOS Catalina 10.15.7
$ uname -a
Darwin Archon.local 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64

# php from brew
$ php -v
PHP 7.4.10 (cli) (built: Sep  3 2020 18:21:42) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Xdebug v2.9.7, Copyright (c) 2002-2020, by Derick Rethans
    with Zend OPcache v7.4.10, Copyright (c), by Zend Technologies
$ php -i | grep 'SSL Version'
SSL Version => OpenSSL/1.1.1g

$ php -i | grep -A1 'cURL support'
cURL support => enabled
cURL Information => 7.72.0
```

rtm-ctrlz:
I posted a comment with a successful test of using only `CURLOPT_SSLCERT` at #9381  .

PR #9386, opened 2020-10-30
Add article on resolving merge conflicts

PrinsFrank:


PrinsFrank:
@lpeabody I'm not familiar with it, so I don't really know what to write here. What change in the article would you suggest?

lpeabody:
It's a tool that excels at handling merge conflicts generated by composer.json and composer.lock. It, for the most part, resolves the majority of conflicts that get generated on its own. The only time I've seen it leave conflicts behind is when two different versions of a package have been updated relative to the base branch (e.g. you're merging two branches where both versions of the same package were updated, and each update is a different version).

I would just mention the tool as an aid for assisting with resolving conflicts from both files. It doesn't get enough love considering how much of a quality of life change I've experienced since I've started using it.

doc/articles/resolving-merge-conflicts.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+This method is easiest when at least one of the braches doesn't have any changes to the `require` and `require-dev` section.
+
+## Validating your merged files
+
+Before comitting, make sure the resulting `composer.json` and `composer.lock` files are valid. If you followed method A, you should never encounter an anvalid file, but it is good to still make sure. If you did merge the composer.json manually using method B you should always run this command as that method is more error prone.
+
+```sh
+composer validate
```
naderman, line 38: This is not sufficient to check correctness, as it doesn't verify the set of packages in the lock file is actually installable after a manual text based merge. Would have to do an actual install (--dry-run)

doc/articles/resolving-merge-conflicts.md
```diff
@@ -0,0 +1,45 @@
...
+
+An example where we have two branches:
+1. Package 'A' has been added
+2. Package 'B' has been removed and package 'C' is added.
+
+To resolve the conflict when we merge these two branches:
+- We choose the branch that has the most changes, and accept the `composer.json` and `composer.lock` files from that branch. In this case, we choose the composer files from branch 2.
+- We reapply the changes from the other branch (branch 1). In this case we have to run ```composer require package/A``` again.
+
+## B. Manually merging the `composer.json` and updating the `composer.lock` file
```
naderman, line 21: I'm not sure we should mention or advocate for this option at all? The correct way to merge is to reapply changes from the branch you are merging?

doc/articles/resolving-merge-conflicts.md
```diff
@@ -0,0 +1,45 @@
+<!--
+    tagline: On gracefully resolving conflicts while merging
+-->
+
+# Resolving merge conflicts
+
+When working as a team on the same Composer project, you'll eventually run into a scenario where multiple people added, updated or removed something in the `composer.json` and `composer.lock` files in multiple branches. When those branches are eventually merged together, you'll get merge conflicts. Resolving these merge conflicts is not as straight forward as on other files, especially not regarding the `composer.lock` file.
```
naderman, line 7: I think this article needs to explain, why text based merging of lock files is not possible.

E.g. if you have one branch which added A 1.0.0 requiring B ^1.0 and the lock file contains B 1.0.0 and another branch which added C 1.0.0 which conflicts with B <1.2 then text merge would lead to A 1.0.0, B 1.0.0 and C 1.0.0, which is actually invalid because C's conflict wasn't considered, and would require an upgrade of B.

naderman (changes requested):

doc/articles/resolving-merge-conflicts.md
PrinsFrank: I wouldn't use this method myself but wanted to be as thorough as possible documenting all the options. But agreed, less is more here and better to learn the correct way immediately.

doc/articles/resolving-merge-conflicts.md
PrinsFrank: Cool, I didn't know that!

doc/articles/resolving-merge-conflicts.md
```diff
@@ -0,0 +1,44 @@
...
+```sh
+composer validate
+composer install [--dry-run]
+```
+
+## Important considerations
+
+Keep in mind that whenever merge conflicts occur on the lock file, the information about the exact version new packages were locked on for one of the branches gets lost. When package A in branch 1 is constrained as `^1.2.0` and locked as `1.2.0`, it might get updated when branch 2 is used as baseline and a new `composer require package/A:^1.2.0` is executed, as that will use the most recent version that the constraint allows when possible. There might be a version 1.3.0 for that package available by now, which will now be used instead.
+
+Choosing the correct [version constraints](../articles/versions.md) and making sure the packages adhere to [semantic versioning](https://semver.org/) when using [netx significant release operators](https://getcomposer.org/doc/articles/versions.md#next-significant-release-operators) should make sure that merging branches doesn't break anything by accidentally updating a dependency.
```
stof, line 44: typo: `next` vs `netx`

stof, line 44: The doc URL should be relative: `versions.md#next-significant-release-operators`

naderman:
@lpeabody based on just having read the README for that tool it actually improperly resolves situations regarding newly introduced conflicts in unrelated dependencies in separate branches, so I would definitely not mention it in official documentation, even though it may work fine in a lot of cases.

lpeabody:
@naderman thanks for the insight. Is there any chance you could take several minutes to post your analysis in an issue for that project?

naderman:
@lpeabody see my comment here https://github.com/composer/composer/pull/9386#discussion_r515902794 and the updated version of this article why text based merges and that's still what the other tool does, even if it understands JSON will not result in valid lock files. With Composer2 I suppose at least a composer update --lock will fail if the lock file isn't resolvable, so I suppose the tool isn't so bad in that at least it will hopefully error if the merge didn't work cleanly.

PrinsFrank:
This are really valuable discussions, thanks! I removed the section about the merge driver!

PrinsFrank:
@Seldaek @naderman The status of this PR is still "changes requested", but as far as I see I processed all requested changes. Am I missing something or am I misunderstanding the way github handles PRs?

Seldaek:
It's just remaining like that until someone else reviews, sorry haven't had time for this in a while but I'll try and get it merged this week.

Seldaek:
Thanks @PrinsFrank, this turned out to be a solid read :)

Issue #9403, opened 2020-11-03
Composer archive removes parts of paths similar to the working directory


tanelpp:
Example:
* the working directory of my project is `/application`
* I have the following file in my project: `templates/application/index.html.twig` 
(So the absolute path is `/application/templates/application/index.html.twig`

When I run this command:
```
composer archive --format=zip
```

Then the ZIP archive contains a file named `templatesindex.html.twig`
 instead of the expected `templates/application/index.html.twig`.

PR #9404, opened 2020-11-03
Remove CWD from only the beginning of a path in ZipArchiver

tanelpp:
Fixes https://github.com/composer/composer/issues/9403


Seldaek:
Thanks

PR #9410, opened 2020-11-03
Fix HTTP status when Platform check failed

jakubboucek:
Composer 2 is checking platform requirements in runtime. It can cause shut down production in a lot of cases.

Here is big trouble the Error message is printed to output and replaced the Web Application. And although Composer is produced an Error message, it's introduced as a serious response of Web application ([status `200 OK`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)).

It's Fail because the robots aren't correctly notified the response is Error and accepts the response as a valid response. It can cause mass burning of Website SEO reputation because the error message is indexed Search engine's robots and replaces the main content.

**Trust me its't very serious problem**, try to [**search "Composer detected issues in your platform"**](https://www.google.com/search?q=%22Composer+detected+issues+in+your+platform%22) for proof.

This PR is update Composer to send [status ~~`503 Service Unavailable`~~](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) [`500 Internal Server Error`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) when the platform check fails.

## FAQ

**App should never be deployed to production with Dev deps**

You right. But Composer 2 is extremely widespread among developers, regardless of their seniority. A lot of them are juniors. The problem is the platform check is in runtime with the most default usage of Composer, it's no requires any additional configuration in `composer.json`, no specific switch in `install`/`require` command. 

Considering the extent of possible damage is necessary to respect most general cases and a variety of uses.

**Check can be disabled by `--no-platform-reqs`, and similar ways…**

You right. But as I wrote above, platform check is Composer's default behavior and it's should be gently to prevent cause damage.

**Developer should ever check production environment before deploy**

You right. But s*it happens. The App can interfere with Platform requirements for a lot of reasons:

- Web hosting can silently change the server environment. Especially its problem on Shares web hosting.
- Developer can do a seemingly innocent update (`vendor-name/package-name 1.2.3 -> 1.2.4`) and does not expect any problem (Composer is not notified about changing platform-reqs during update).
- Developers just do mistakes - we are just humans.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -734,7 +734,8 @@ EXT_CHECKS;
 
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
-if (\$issues) {
+if (\$issues && !headers_sent()) {
+    header('HTTP/1.1 500 Internal Server Error');
     echo 'Composer detected issues in your platform:' . "\\n\\n" . implode("\\n", \$issues) . "\\n\\n";
     exit(104);
 }
--- a/tests/Composer/Test/Autoload/Fixtures/platform/no_extensions_required.php
+++ b/tests/Composer/Test/Autoload/Fixtures/platform/no_extensions_required.php
@@ -8,7 +8,8 @@ if (!(PHP_VERSION_ID >= 70200)) {
     $issues[] = 'Your Composer dependencies require a PHP version ">= 7.2.0". You are running ' . PHP_VERSION  .  '.';
 }
 
-if ($issues) {
+if ($issues && !headers_sent()) {
+    header('HTTP/1.1 500 Internal Server Error');
     echo 'Composer detected issues in your platform:' . "\n\n" . implode("\n", $issues) . "\n\n";
     exit(104);
 }
--- a/tests/Composer/Test/Autoload/Fixtures/platform/no_php_required.php
+++ b/tests/Composer/Test/Autoload/Fixtures/platform/no_php_required.php
@@ -12,7 +12,8 @@ if ($missingExtensions) {
     $issues[] = 'Your Composer dependencies require the following PHP extensions to be installed: ' . implode(', ', $missingExtensions);
 }
 
-if ($issues) {
+if ($issues && !headers_sent()) {
+    header('HTTP/1.1 500 Internal Server Error');
     echo 'Composer detected issues in your platform:' . "\n\n" . implode("\n", $issues) . "\n\n";
     exit(104);
 }
--- a/tests/Composer/Test/Autoload/Fixtures/platform/no_php_upper_bound.php
+++ b/tests/Composer/Test/Autoload/Fixtures/platform/no_php_upper_bound.php
@@ -8,7 +8,8 @@ if (!(PHP_VERSION_ID >= 70200)) {
     $issues[] = 'Your Composer dependencies require a PHP version ">= 7.2.0". You are running ' . PHP_VERSION  .  '.';
 }
 
-if ($issues) {
+if ($issues && !headers_sent()) {
+    header('HTTP/1.1 500 Internal Server Error');
     echo 'Composer detected issues in your platform:' . "\n\n" . implode("\n", $issues) . "\n\n";
     exit(104);
 }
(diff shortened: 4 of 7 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -735,6 +735,7 @@ protected function getPlatformCheck($packageMap, array $ignorePlatformReqs)
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
 if (\$issues) {
+    http_response_code(503);
```
stof, line 738: this won't work on PHP 5.3 (which is still supported by Composer 2.0), so this should at least be guarded by a version check

stof (changes requested):

src/Composer/Autoload/AutoloadGenerator.php
jakubboucek: Hm, you right. 

Please help me to choose better way:
- `header("HTTP/1.1 503 Service Unavailable");` or
- `header(':', true, 503);` or
- `header('X-PHP-Response-Code: 404', true, 404);`

PHP 5.3 is now too far away for me, I must [consult it with Google](https://stackoverflow.com/questions/3258634/php-how-to-send-http-response-code), thanks for understand.

jakubboucek: Changed to `header("HTTP/1.1 503 Service Unavailable");`

ondrejmirtes:
I agree something should be done about this. Maybe the check should run only in CLI SAPI?

jakubboucek:
@ondrejmirtes:
> Maybe the check should run only in CLI SAPI?

Checks on runtime in non-CLI SAPI is not completely bad idea. 

Just it must be implemented correctly and must be required by developer, instead of as default behavior in default setup.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -735,6 +735,7 @@ protected function getPlatformCheck($packageMap, array $ignorePlatformReqs)
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
 if (\$issues) {
+    header('HTTP/1.1 503 Service Unavailable');
```
Seldaek, line 738: Needs to be wrapped in `if (!headers_sent()) { ... }` for safety to avoid triggering errors.

naderman:
This needs to return a 500 Internal Server Error. The server is not unavailable in this case and it's not a temporary error that will potentially be fixed upon retry.

src/Composer/Autoload/AutoloadGenerator.php
jakubboucek: You right. Updated.

The `!headers_sent()` condition is wrapping whole error reporting, not only `header()`, because it prevent to corrupting application response in half.

jakubboucek:
> @naderman This needs to return a 500 Internal Server Error. The server is not unavailable in this case and it's not a temporary error that will potentially be fixed upon retry and it is not server maintenance.

Good point, fixed.

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: I don't think that's good, because now it means we do not check/abort at all if headers were sent.

dg: What about `echo '<!-- Composer detected issues in your platform` when SAPI is not CLI?

Seldaek: Please don't side-track this, this PR is only about the header. We have other improvements tracked in https://github.com/composer/composer/issues/9412

jakubboucek:
I suggest close this PR and replace it with #9414.

src/Composer/Autoload/AutoloadGenerator.php
jakubboucek: @Seldaek What happen when Composer break rendering page in half? It's produce broken response and provide it as classic response – same result for SEPR and produce huge damage to SEO. 

I thinking and do some tests on few environments with result: it's all wrong, follow #9414.

dkulyk:
When header_send error will not show.
```php
if (\$issues && !headers_sent()) {
    header('HTTP/1.1 500 Internal Server Error');
    echo 'Composer detected issues in your platform:' . "\\n\\n" . implode("\\n", \$issues) . "\\n\\n";
    exit(104);
}
```
Maybe replace to
```php
if (\$issues) {
    headers_sent() || header('HTTP/1.1 500 Internal Server Error');
    echo 'Composer detected issues in your platform:' . "\\n\\n" . implode("\\n", \$issues) . "\\n\\n";
    exit(104);
}
```

Seldaek:
@dkulyk see https://github.com/composer/composer/pull/9414 - this is marked merged because it was included in the other PR, but the code hsa been replaced.

Issue #9412, opened 2020-11-04
Changes to default platform check behavior


Seldaek:
It's quite clear the more people move to Composer 2 that this is causing problems in many places. We've had many issues reported, and tweets and whatnot, plus as https://github.com/composer/composer/issues/9411 shows a few sites are actually down still due to this.

I see a few ways to improve this while still retaining some value to the feature:

- [x] **Skip extension checks by default** by changing the platform-check option to be `false` (off), `php-only` (new default), or `true` (full check of exts+php, current default). The reason here is it seems like in most cases it's failing due to some ext which is optionally required, and everything would work fine still without it. Not exactly warranting a white page.
- [x] **Skip dev dependencies** by persisting to installed repo which packages are dev requires and which are not. The reason here is I suspect in some cases people deploy with dev dependencies included and that can change the required platform packages quite substantially in some cases.
- [x] **Improve output** explaining a bit more what is going on, how this can be fixed, etc.
- [x] **Improve http responses** by making sure if we're in a web sapi we set the response status code to 500 before exiting. Also adding  a link to the platform-check docs. In progress in https://github.com/composer/composer/pull/9410
- [ ] **Improve docs** to include details from https://github.com/composer/composer/issues/9377#issuecomment-719019204
- [x] **Follow error reporting standards more closely** On CLI throw a regular error and let PHP decide how to log/output it, on other sapis try to display some hint as to what caused the error without exposing specific info like php version (ranges), throw a PHP error to make sure logging/error reporting settings are respected. In progress in https://github.com/composer/composer/pull/9414

Frameworks who wish to force this on their users could still include the platform-check option set to `true` in their project templates.

JohnMaguire:
>  Skip dev dependencies by persisting to installed repo which packages are dev requires and which are not. The reason here is I suspect in some cases people deploy with dev dependencies included and that can change the required platform packages quite substantially in some cases.

Just wanted to chime in and say thanks for this change! Our development workflow involves creating a "production" build container (without dev dependencies), and then overlaying our development repository over it. The development repository contains dev dependencies since we need to run tests, but the production image does not contain the required PHP extensions for these, since they're not used inside the Docker container itself.

This caused an issue where we suddenly started seeing errors in our development environment which broke the application, but not our production environment. The behavior was present in 2.0.4, but not 2.0.7 and was a bit tricky to track down the reason. I believe this is it, and the new behavior is much welcomed!

PR #9414, opened 2020-11-04
Platform check: Trigger native PHP error

jakubboucek:
Following up #9412 and #9410 I found current error handling with mechanism `header(); echo …; die(1+);` is not satisfied for success debugging on most implementations.

General monitoring tools prefers to monitor native error log instead to parsing response body (tested on [Google App Engine](https://cloud.google.com/appengine) and [Google Run](https://cloud.google.com/run)).

Here are two reasons:

- native error handling in any level (regardless to app-level or sys-level),
- **security** – printing too detailed internal error message in runtime is very sensitive, it can provide implementation details for attackers,
- indepentent to SAPI (cli, cgi, etc.), environment, and others, because it's produce native error state in app runtime lifecycle.

**This PR is replacement to #9410 and it's conflicts with them.** Only one can be merged, I prefer this PR.

## FAQ
**This PR is cause to show empty page without body.**

That's right. I understand it's unpleasantly. This is standard way to propagate raw errors. User can change this behavior by setting php runtime like:
```ini
ini_set('display_errors', 1);
ini_set('display_startup_errors', 1);
error_reporting(E_ALL);
```
or by [custom error handling](https://www.php.net/manual/en/function.set-error-handler.php).

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -734,7 +734,8 @@ EXT_CHECKS;
 
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
-if (\$issues) {
+if (\$issues && !headers_sent()) {
+    header('HTTP/1.1 500 Internal Server Error');
     echo 'Composer detected issues in your platform:' . "\\n\\n" . implode("\\n", \$issues) . "\\n\\n";
     exit(104);
 }
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -734,10 +734,11 @@ EXT_CHECKS;
 
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
-if (\$issues && !headers_sent()) {
-    header('HTTP/1.1 500 Internal Server Error');
-    echo 'Composer detected issues in your platform:' . "\\n\\n" . implode("\\n", \$issues) . "\\n\\n";
-    exit(104);
+if (\$issues) {
+    trigger_error(
+        'Composer detected issues in your platform: ' . implode(', ', \$issues),
+        E_USER_ERROR
+    );
 }
 
 PLATFORM_CHECK;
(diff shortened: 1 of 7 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -735,8 +735,10 @@ protected function getPlatformCheck($packageMap, array $ignorePlatformReqs)
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
 if (\$issues) {
-    echo 'Composer detected issues in your platform:' . "\\n\\n" . implode("\\n", \$issues) . "\\n\\n";
-    exit(104);
+    trigger_error(
+        'Composer detected issues in your platform: ' . implode(', ', \$issues),
+        E_USER_ERROR
+    );
```
stof, line 741: This won't exit anymore though

jakubboucek: Are you sure? I tested it first, see here: https://3v4l.org/KD3ho and in all versions `E_USER_ERROR` cause to exit.

In addition I originally append `die(104);` after code, but code inspection notify me:
```
tests/Composer/Test/Autoload/Fixtures/platform/no_php_required.php:20: Unreachable statement 
Inspection info: Reports the statements that cannot be normally reached
```

Is here case where running does not exits?

naderman: @stof it will by default exit with a fatal error. Depends on whether you have some custom exception handler loaded before this gets called which may decide to do sth different, which is really by design as we shouldn't circumvent standard php mechanisms.

naderman:
@jakubboucek Thanks for the PR I definitely prefer this over what is currently happening, we should be using standard PHP error handling mechanisms, as I already documented in the last point on the checklist here: https://github.com/composer/composer/issues/9412

The main concern now would be that people who e.g. use wordpress and just update some plugin will be confronted with a blank page with no idea what caused it. We shouldn't display internal info, but I suppose a hint like "Composer detected an incompatible PHP version" with a link to our docs may be output anyway if sapi is web and headers haven't been sent yet?

jakubboucek:
@naderman Thanks. I am thinking about it for while time and I changing my opinion: the any variant of fatal error is undesirable for default behavior, until developer expressly require it.

Composer v1 was been very useful utility to compose app. 

Composer v2 has now new face, it's not yet helper, friend, slave for my work. Now is getting new despotic face, it's can violently break my job when anything reqs is missed.

Here should be `E_USER_WARNING` at most. But never break runtime in default setup, until developer change default behavior if he is aware of the consequences of his actions.

Seldaek:
Thanks, merged and added some extra output back in case errors are not visible, see 8c1355f448ec8087bf7ed468208f8d63794de909

PR #9416, opened 2020-11-04
Update `self-update` options

pdelre:
There were a few missing options in the documentation that didn't align with the help from `composer self-update --help`. In particular, the `--1` and `--2` options.


Seldaek:
Thanks

PR #9418, opened 2020-11-04
Fix platform check error message

jakubboucek:
Fixes bug from commit: 8c1355f448ec8087bf7ed468208f8d63794de909

- ~~Prints error to output only when `ini_get('display_errors')` is truly,~~
- Add status `500 Internal server error` to error message (see #9410 discussion; ~~note the `trigger_error()` below doesn't change status, because `echo` is already sent data to output~~).

@Seldaek however, in any case, your solution is better than my previous solution. Thanks for your great work!

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -741,6 +741,9 @@ EXT_CHECKS;
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
 if (\$issues) {
+    if (!headers_sent()) {
+        header('HTTP/1.1 500 Internal Server Error');
+    }
     if (!ini_get('display_errors')) {
         if (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') {
             fwrite(STDERR, 'Composer detected issues in your platform:' . PHP_EOL.PHP_EOL . implode(PHP_EOL, \$issues) . PHP_EOL.PHP_EOL);
--- a/tests/Composer/Test/Autoload/Fixtures/platform/no_extensions_required.php
+++ b/tests/Composer/Test/Autoload/Fixtures/platform/no_extensions_required.php
@@ -9,6 +9,9 @@ if (!(PHP_VERSION_ID >= 70200)) {
 }
 
 if ($issues) {
+    if (!headers_sent()) {
+        header('HTTP/1.1 500 Internal Server Error');
+    }
     if (!ini_get('display_errors')) {
         if (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') {
             fwrite(STDERR, 'Composer detected issues in your platform:' . PHP_EOL.PHP_EOL . implode(PHP_EOL, $issues) . PHP_EOL.PHP_EOL);
--- a/tests/Composer/Test/Autoload/Fixtures/platform/no_php_required.php
+++ b/tests/Composer/Test/Autoload/Fixtures/platform/no_php_required.php
@@ -13,6 +13,9 @@ if ($missingExtensions) {
 }
 
 if ($issues) {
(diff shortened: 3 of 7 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -735,10 +735,11 @@ protected function getPlatformCheck($packageMap, array $ignorePlatformReqs, $che
 \$issues = array();
 ${requiredPhp}${requiredExtensions}
 if (\$issues) {
-    if (!ini_get('display_errors')) {
+    if (ini_get('display_errors')) {
```
Seldaek, line 738: I think you misunderstood here :) The point is to ensure that if we break a site and it has no errors visible, we print the message so it's not a blank page and gives a hint to people. Please revert.

Seldaek (changes requested):

Seldaek:
Makes sense to add the 500 back, totally forgot, thanks.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -735,10 +735,11 @@ protected function getPlatformCheck($packageMap, array $ignorePlatformReqs, $che
...
         if (PHP_SAPI === 'cli' || PHP_SAPI === 'phpdbg') {
             fwrite(STDERR, 'Composer detected issues in your platform:' . PHP_EOL.PHP_EOL . implode(PHP_EOL, \$issues) . PHP_EOL.PHP_EOL);
         } elseif (!headers_sent()) {
+            header('HTTP/1.1 500 Internal Server Error');
```
Seldaek, line 742: actually seems like if display_errors is off, php leaves it 200, so please move this all the way up above line 738 so we always set the status code (unless headers have been sent already..).

src/Composer/Autoload/AutoloadGenerator.php
jakubboucek: @Seldaek That sounds like a very bad idea. 

**Reasons:**

- Developers is turns off this flat **to prevent any error message to output**. Their customers does not care about any internal bugs. It's look seriously non-proffesional.
- Developers often have in their projects [custom error handling](https://www.php.net/manual/en/function.set-error-handler.php) which is render custom error page. This `echo` is bypass clean handling.
- The message is leaking too detailed specifics about app/server configuration, it's very unsecure.

src/Composer/Autoload/AutoloadGenerator.php
jakubboucek: Hm, you right. I was really trusts the `trigger_error()` do it herself, but he didn't :-/

Moved as you said & rebased.

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: Well I think we're gonna have to agree to disagree here :) 

- Now that I changed this to only check php version by default, and to only include non-dev dependencies in the check, the chance that this outputs some "internal bug" which is irrelevant is very unlikely. Stuff will most likely break if the PHP version is too low.
- The project error handler will in most cases not be loaded yet when the platform check runs as it is loaded with the autoloader, which is typically one of the first lines of code running.
- It's not leaking any application info anymore, as it now str_replaces the php version away when outputting on non-CLI SAPIs.

On top of that, if display_errors is on, the error will be shown already, so echoing it again on top is completely useless and duplicate.

jakubboucek: Your arguments are strong, but I'm still convinced the direct printing error message to output in runtime (expecially at non-CLI SAPI) is subject of exceeding the boundary of responsibility that users can fairly expect. :)

Reverted, thanks.

Seldaek: 👍  Thanks, merging then

PR #9422, opened 2020-11-05
Mention no-slash requirement for getInstallPath()

kdambekalns:


kdambekalns:
See https://github.com/neos/composer-plugin/issues/16 for the kind of issue that causes…

Issue #8785, opened 2020-04-14
PHP support strategy


Seldaek:
(Following-up from https://github.com/composer/composer/issues/8759#issuecomment-613642026)

When it comes to supported PHP versions, should we:

- [ ] 😕  Ship multiple phar files with different dependency versions for different min-PHP versions. This has the potential to be messy when people report bugs, and it means we still need to support PHP 5.3+ in Composer so overall not great.
- [ ] 🎉  Skip to 3.0 directly after the 2.0 release and drop PHP versions there. Easier on maintenance for us, but it causes a mess for all plugins which would need to be adjusted again to support 3.0. 
- [x] 👍  Drop older PHP versions in 2.1, but do not make any refactorings of signature to use new language features. That'd be a good compromise maybe to keep BC in signatures for now but be allowed to modernize internally a little bit at least. Old PHP users can still use 2.0.x which would be supported as far as bugfixes are concerned.

You can vote in reactions if you care either way.

stof:
>     * Skip to 3.0 directly after the 2.0 release and drop PHP versions there. Easier on maintenance for us, but it causes a mess for all plugins which would need to be adjusted again to support 3.0.

Well, does it actually require bumping the plugin-api-version ? (avoiding confusion in the ecosystem might be a reason, if composer 3.0 ships with the plugin-api version 2 too)

stof:
Also, do we already have features in place in the composer install and the `self-update` command to account for required PHP version (so that a phar does not self-update itself to a new version not compatible with your PHP runtime) ? If no, this must be implemented first.

Seldaek:
> Well, does it actually require bumping the plugin-api-version ? (avoiding confusion in the ecosystem might be a reason, if composer 3.0 ships with the plugin-api version 2 too)

Yes it does IMO, if we keep everything BC to keep the plugin-api version to 2, there is no need to bump the version to 3.0. That's the 👍 option.

> Also, do we already have features in place in the composer install and the self-update command to account for required PHP version

Yes that's been part of it since the beginning.

cebe:
Which versions are you planning to drop? Would be good to have a 2020 version of <https://seld.be/notes/php-versions-stats-2018-1-edition> to see in which environments composer is used currently.

Seldaek:
[...]

PR #9433, opened 2020-11-06
dirname(__FILE__) no longer needed

mvorisek:
we no longer support PHP ~7.2~ 5.2 in `master`, see https://3v4l.org/sevVD

UPDATE: see discussion below, merge once PHP 5.2 support is dropped completely (for use, not only defined by composer)


stof:
As of now, the master branch still supports PHP 5.3: https://github.com/composer/composer/blob/86f0c100049d05b457c215f51b981457b9a587c7/composer.json#L25

mvorisek:
@stof was type from me in the desc, we support >= 5.3

stof:
@mvorisek Composer 1 **always** supported running only on PHP 5.3+. But the generated `vendor/composer/autoload_*.php` were usable on PHP 5.2+ (except for `autoload_static.php` which targets PHP 5.6+ anyway).
The discussion in #8785 decided to drop that only when dropping older PHP versions too (so in Composer 2.2)

Issue #9424, opened 2020-11-06
Composer 2, no error message produced when using insecure repository URL. Just hangs on "Updating dependencies".


the-toster:
My `composer.json`:

```json
...
  "repositories": [{
    "type": "composer",
    "url": "http://192.168.1.105:8081/"
  }],
...
```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.0.4
PHP version: 7.4.1
PHP binary path: C:\Bitnami\nginxstack-1.12.2-5\php\php.exe
OpenSSL version: OpenSSL 1.1.1d  10 Sep 2019
cURL version: 7.67.0 libz 1.2.11 ssl OpenSSL/1.1.1d
zip extension: OK

C:\Users\Programmer\Desktop\lounge\lounge-eloquent-connector>composer diagnose
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.0.4
PHP version: 7.4.1
PHP binary path: C:\Bitnami\nginxstack-1.12.2-5\php\php.exe
OpenSSL version: OpenSSL 1.1.1d  10 Sep 2019
cURL version: 7.67.0 libz 1.2.11 ssl OpenSSL/1.1.1d
zip extension: OK

```

When I run this command:

```
composer install -vvv
```

I get the following output:

```
...
Loading composer repositories with package information
[...]

PR #9442, opened 2020-11-08
Reject promise within "HttpDownloader" when an exception is thrown

Serializator:
### Description
An infinite loop was caused by promises which weren't either resolved or rejected within the [`Composer\Util\HttpDownloader::startJob`](https://github.com/composer/composer/blob/master/src/Composer/Util/HttpDownloader.php#L243) when an exception was thrown by the [`Composer\Util\Http\CurlDownloader::download`](https://github.com/composer/composer/blob/master/src/Composer/Util/Http/CurlDownloader.php#L112) method.

A promise and reject callback are passed to the aforementioned methods but are only passed through to [`$this->jobs`](https://github.com/composer/composer/blob/master/src/Composer/Util/Http/CurlDownloader.php#L223) but not invoked if there is an exception within [`Composer\Util\Http\CurlDownloader::initDownload`](https://github.com/composer/composer/blob/master/src/Composer/Util/Http/CurlDownloader.php#L126) itself.

In case if exceptions thrown by the method itself there could be two scenarios.

- [An exception thrown when HTTP is used if `secure-http` is enabled](https://github.com/composer/composer/blob/master/src/Composer/Util/Http/CurlDownloader.php#L138)
- [If `$copyTo` is passed and the the file wasn't able to be written](https://github.com/composer/composer/blob/master/src/Composer/Util/Http/CurlDownloader.php#L155)

I decided to catch these exceptions and pass them to the `$reject` callback within `Composer\Util\HttpDownloader` rather than `Composer\Util\Http\CurlDownloader` because `Composer\Util\Http\CurlDownloader::download` returns the job id which was started and I wouldn't know what to return if an exception came up.

### Related Issues
- #9424

### DIFF
--- a/src/Composer/Util/HttpDownloader.php
+++ b/src/Composer/Util/HttpDownloader.php
@@ -268,10 +268,14 @@ class HttpDownloader
             return;
         }
 
-        if ($job['request']['copyTo']) {
-            $job['curl_id'] = $this->curl->download($resolve, $reject, $origin, $url, $options, $job['request']['copyTo']);
-        } else {
-            $job['curl_id'] = $this->curl->download($resolve, $reject, $origin, $url, $options);
+        try {
+            if ($job['request']['copyTo']) {
+                $job['curl_id'] = $this->curl->download($resolve, $reject, $origin, $url, $options, $job['request']['copyTo']);
+            } else {
+                $job['curl_id'] = $this->curl->download($resolve, $reject, $origin, $url, $options);
+            }
+        } catch (\Exception $exception) {
+            $reject($exception);
         }
     }
 


Seldaek:
Thanks, nice catch 👍 

I wonder if the same shouldn't be done for the RemoteFilesystem fallback code (lines 184-197), it probably should try/catch there and call $reject directly? If you can try it out that'd be great. If you make canUseCurl return false it'll use this fallback path.

Serializator:
If I add a throw to `Composer\Util\RemoteFilesystem::getContents` (which is called on line 192 in `Composer\Util\HttpDownloader::addJob`) the output is like below.

Screenshot 2020-11-09 at 17 12 32 

`Hello, World` is the message I passed to the exception. I believe there is already some kind of exception handling covering this scenario.

@Seldaek, what do you think?

Seldaek:
You're right, it doesn't seem necessary. Thanks again

PR #9461, opened 2020-11-11
Reference co-relating code sequence for tty test, refs #9454

### LINKED ISSUE #9454: Composer-v2 run-script output regression
Given a composer-script:

~~~
{
  "scripts": {
    "cat": ["cat vendor/composer/LICENSE"]
  }
}
~~~

When running this script with Composer v1, then following output is given:

~~~
$ VER=1; composer -q self-update --$VER && composer --version
Composer version 1.10.17 2020-10-30 22:31:58
$ composer run-script cat >/dev/null
> cat vendor/composer/LICENSE
~~~
_(the redirection to `/dev/null` effectively hides the cat script output, only the informal message by composer run-script is given on standard error)_

when running this script with Composer v2, then following output is given:

~~~
$ VER=2; composer -q self-update --"$VER" && composer --version
Composer version 2.0.6 2020-11-07 11:21:17
$ composer run-script cat >/dev/null
> cat vendor/composer/LICENSE

Copyright (c) Nils Adermann, Jordi Boggiano

Permission is hereby granted, free of charge, to any person obtaining a copy
[...]
LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
THE SOFTWARE.
~~~
_(shortened, the redirection to `/dev/null` is not in effect [/edit: for the scripts output])_

Expected behaviour:

The redirection is in effect (the behaviour of Composer v1).

Additional Info:

Running on Ubuntu 18.04 LTS in standard bash shell.

Standard error is unchanged.

First noticed it with a php script, then reduced the example to cat for the report.

Also disabled as much as possible in php to run it with the minimal setup possible, in php 7.1/2/3/4 and 8.0

~~~
php -n -dextension=phar.so -dextension=json.so -f "$HOME/bin/composer" -- run-scrip cat >/dev/null
~~~

ktomk:
The change in 380dce9a6b2886f1ff89103ea170092abe8866f5 ~effectively renders the earlier change superfluous.~

~This reverts commit 14962776971e753b31a062783ad758ed0e8dde24.~

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -131,6 +131,7 @@ class Application extends BaseApplication
 
         if (
             getenv('COMPOSER_NO_INTERACTION')
+            /* @see \Composer\Util\ProcessExecutor::executeTty - tty test */
             || (function_exists('stream_isatty') && !stream_isatty(STDOUT))
             || (function_exists('posix_isatty') && !posix_isatty(STDOUT))
         ) {
--- a/src/Composer/Util/ProcessExecutor.php
+++ b/src/Composer/Util/ProcessExecutor.php
@@ -78,6 +78,7 @@ class ProcessExecutor
     public function executeTty($command, $cwd = null)
     {
         if (
+            /* @see \Composer\Console\Application::doRun - tty test */
             (function_exists('stream_isatty') && !stream_isatty(STDOUT))
             || (function_exists('posix_isatty') && !posix_isatty(STDOUT))
         ) {


Seldaek:
I think it might be better to keep the check here in case this gets called somewhere else later or in a plugin without checking for the io interactivity. Seems fairly harmless and safer.

ktomk:
Seeing now, this is a public method. Me is also weak on comments, this is the best idea I have then. Your say.

Seldaek:
Ok yeah sure this is even safer.

PR #9468, opened 2020-11-12
Make sure the root aliases always get loaded for root packages, even if not stored in the lock file

Seldaek:
fixes #9448

PR #9473, opened 2020-11-13
Make sure the root aliases always get installed when a package is updated

### LINKED ISSUE #9448: dev-master as x inline aliasing regressions/issues
My `composer.json`:

```json
"dereuromark/cakephp-ide-helper": "dev-master as 1.8.2",
```
This is also what is already installed (in lock file).

When I run this command:

```
composer require --dev dereuromark/cakephp-ide-helper-extra:dev-master
```

I get the following output:

```
Problem 1
    - Root composer.json requires dereuromark/cakephp-ide-helper-extra dev-master -> satisfiable by dereuromark/cakephp-ide-helper-extra[dev-master].
    - dereuromark/cakephp-ide-helper-extra dev-master requires dereuromark/cakephp-ide-helper ^1.3.0 -> found dereuromark/cakephp-ide-helper[1.3.0, ..., 1.8.2] but it conflicts with your root composer.json require (dev-master as 1.8.2).

```

The installed version already matches the requirements (1.8.2 is inside `^1.3`). So the error here is false positive.
If I manually put the the line in:
```json
"dereuromark/cakephp-ide-helper-extra": "dev-master",
```

It just does
```
Updating dependencies
Lock file operations: 1 install, 0 updates, 0 removals
  - Locking dereuromark/cakephp-ide-helper-extra (dev-master c9d9c46)
Writing lock file
Package operations: 1 install, 0 updates, 0 removals
  - Syncing dereuromark/cakephp-ide-helper-extra (dev-master c9d9c46) into cache
  - Installing dereuromark/cakephp-ide-helper-extra (dev-master c9d9c46): Cloning c9d9c46088 from cache
```
and all works.

So that proves that there is an issue with require calls and inline aliasing (for testing some branch).

Note: the same issue appears also for "composer update -W" and inline aliasing used there, if nothing got changed it still errors here too with "it cannot update".

Seldaek:
fixes #9448

replaces and closes #9468 with a better implementation which makes sure the root aliases get stored in lock file, but newly defined aliases do not make it in there on partial updates, and do not get loaded at all on locked packages

### DIFF
--- a/src/Composer/DependencyResolver/LockTransaction.php
+++ b/src/Composer/DependencyResolver/LockTransaction.php
@@ -133,6 +133,10 @@ class LockTransaction extends Transaction
             }
         }
 
+        usort($usedAliases, function ($a, $b) {
+            return strcmp($a['package'], $b['package']);
+        });
+
         return $usedAliases;
     }
 }
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -55,6 +55,11 @@ class Pool implements \Countable
         }
     }
 
+    public function getPackages()
+    {
+        return $this->packages;
+    }
+
     /**
      * Retrieves the package object for a given package id.
      *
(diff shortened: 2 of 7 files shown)


naderman:
Closed in favour of https://github.com/composer/composer/pull/9475

Issue #9458, opened 2020-11-11
Composer v2 regression in conflict resolution


xificurk:
My `composer.json`:

```json
{
    "name": "xificurk/composer-conflict-bug",
    "type": "project",
    "require": {
        "nette/utils": "^3.0@dev"
    },
    "conflict": {
        "nette/utils": ">=3.2"
    }
}

```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.0.6
PHP version: 7.4.12
PHP binary path: /usr/local/bin/php
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
cURL version: 7.69.1 libz 1.2.11 ssl OpenSSL/1.1.1g
zip extension: OK
```

When I run this command:

```
composer update --dry-run
```

I get the following output:

```
Loading composer repositories with package information
Updating dependencies
Lock file operations: 1 install, 0 updates, 0 removals
  - Locking nette/utils (dev-master 7a71315)
Installing dependencies from lock file (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
  - Installing nette/utils (dev-master 7a71315)
2 package suggestions were added by new dependencies, use `composer suggest` to see details.
```

And I expected this to happen:

Lower version of `nette/utils` should be installed, not `dev-master` which should be excluded by conflict specified in composer.json (Note: master has branch alias 3.2-dev).

The same composer.json and the same update command with composer 1.x correctly resolve the dependency and install `3.1.x-dev` version of `nette/utils`

```
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
  - Installing nette/utils (v3.1.x-dev 3ec0ff0)
```

[...]

PR #9477, opened 2020-11-13
Explicit conflicts should conflict with aliases of packages too

naderman:
Fixes https://github.com/composer/composer/issues/9458

Rule generation for explicit conflict definitions skipped over alias packages so far.


stof (approved):

naderman:
This should be ready now. The new solution makes it so aliases must always be installed with their original package.

Otherwise this lead to all kinds of problems as we'd have to know which aliases in the lock file or installed.json were not actually installed despite them being defined in the JSON metadata. Alternatively we may have needed to remove branch alises from the lock/installed files but then metadata would have differed from remote which would have made it hard to know when to update it on e.g. update mirrors.

The only consequence this has is that you cannot install a branch anymore now if you or one of your deps explicitly conflicts with its alias version definition.

Issue #9450, opened 2020-11-09
Increase phpstan level / introduce a phpstan baseline


staabm:
Hey guys!

I was wondering why composer uses phpstan at level 1.
This leaves behind a lot of opportunities, because phpstan will deliver more [important error at higher levels](https://phpstan.org/user-guide/rule-levels).

From my experience we should aim for at least level 5, which enables type-checking of arguments passed to methods and functions.

To ease the transition to higher levels [phpstan supports a baseline feature](https://phpstan.org/user-guide/baseline), which basically allows us to ignore all errors the codebase currently has.

After everything is setup this effectivly means new code introduced to the codebase benefits from those more usefull checks.
Still you could baseline/ignore errors which we think aren‘t worth to be fixed when developing new code.

Any opinions?

Seldaek:
Yes we could do that.. though I'd also like to fix issues as much as possible (5.3 support and BC restrictions might make fixing some of it hard).

I think at latest 2.2 when we drop older PHP it might be a good time to do this.

Seldaek:
This was done in a while back

PR #9478, opened 2020-11-13
Fix phpstan phpdocs issues

adlacruzes:
Hi,

after reading #9450, I think we can adopt an intermediate step when increasing phpstan level.

Level 2 show ~280 errors and could be hard fixing them without a baseline. Instead of increase full level we can add explicit rules from a higher level and fix errors progressively easier.

This PR fixes phpdocs issues and add some phpstan level 2 rules to config file.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -599,7 +599,7 @@ class PlatformRepository extends ArrayRepository
     /**
      * Check if a package name is a platform package.
      *
-     * @param $name
+     * @param mixed $name
      * @return bool
      */
     public static function isPlatformPackage($name)
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -599,7 +599,7 @@ class PlatformRepository extends ArrayRepository
     /**
      * Check if a package name is a platform package.
      *
-     * @param mixed $name
+     * @param string $name
      * @return bool
      */
     public static function isPlatformPackage($name)
--- a/tests/Composer/Test/Util/AuthHelperTest.php
+++ b/tests/Composer/Test/Util/AuthHelperTest.php
@@ -514,8 +514,8 @@ class AuthHelperTest extends TestCase
     }
 
     /**
-     * @param $origin
-     * @param $auth
+     * @param mixed $origin
+     * @param mixed $auth
      */
     private function expectsAuthentication($origin, $auth)
     {
--- a/tests/Composer/Test/Util/AuthHelperTest.php
+++ b/tests/Composer/Test/Util/AuthHelperTest.php
@@ -514,8 +514,8 @@ class AuthHelperTest extends TestCase
     }
 
     /**
-     * @param mixed $origin
-     * @param mixed $auth
+     * @param string $origin
+     * @param array $auth
      */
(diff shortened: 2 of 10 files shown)


staabm (approved):
Lgtm

src/Composer/Repository/PlatformRepository.php
```diff
@@ -599,7 +599,7 @@ private function addLibrary($name, $prettyVersion, $description = null, array $r
     /**
      * Check if a package name is a platform package.
      *
-     * @param $name
+     * @param mixed $name
```
stof, line 602: `string` actually

tests/Composer/Test/Util/AuthHelperTest.php
```diff
@@ -514,8 +514,8 @@ public function testStoreAuthWithPromptInvalidAnswer()
     }
 
     /**
-     * @param $origin
-     * @param $auth
+     * @param mixed $origin
```
stof, line 517: `string`

tests/Composer/Test/Util/AuthHelperTest.php
```diff
@@ -514,8 +514,8 @@ public function testStoreAuthWithPromptInvalidAnswer()
...
+     * @param mixed $auth
```
stof, line 518: `array`

stof (changes requested):

stof:
> after reading #9450, I think we can adopt an intermediate step when increasing phpstan level.
> 
> Level 2 show ~280 errors and could be hard fixing them without a baseline. Instead of increase full level we can add explicit rules from a higher level and fix errors progressively easier.

I'd rather go to level 2 and introducing the baseline. that would be easier to maintain (no need to maintain a list of rules belonging to level 2) and would bring more value (new code would be checked at level 2).

the answer from @Seldaek on #9450 did not reject the usage of a baseline. This answer was rejecting the fact of adding a baseline without trying to fix as much reported issues as possible before generating it (so that moving to level 2 actually helps increasing the quality of the code).

src/Composer/Repository/PlatformRepository.php
adlacruzes: Done

tests/Composer/Test/Util/AuthHelperTest.php
adlacruzes: Done

tests/Composer/Test/Util/AuthHelperTest.php
adlacruzes: Applied

adlacruzes:
Hi @stof, I rechecked mixed types and applied your correct suggestions.

The idea is to remove these rules once level 2 is reached, but its true we are introducing one more thing to maintain.

Should I limit this pull request to the proposed fixes, so we can merge it?

Thanks!

stof:
> The idea is to remove these rules once level 2 is reached, but its true we are introducing one more thing to maintain.

I'd rather do this by configuring level 2 with a baseline, and then work on emptying the baseline little by little to reach full level 2. this way, new code would be written at level 2 already.

Seldaek:
Going to merge this as it already fixes some things and adds more strictness, it goes in the right direction. But yes full level 2 or more with baseline probably makes more sense in the long run.

adlacruzes:
Thanks everyone!

Issue #601, opened 2012-04-19
local-global repositories (composer link command)


fprochazka:
Hi,
 I'm writting simulatenously few libraries and it's realy pain to commit everything and for every few lines write <code>$ composer update</code>. So it would be nice to have new kind of local repository or something, that wouldn't copy the dependency into project, but rather link to it. So when I change something in one project, it affects the other one immediately and I don't have to commit stuff and run update on composer.

Some command like "link this library, but don't touch it, just autoload it".

Or is there any suggested worflow on this problem?

dantleech:
This would be a nice feature..

minutephp:
Yes, this is a long sought after feature.. right now I'm doing this to get a similar result (see the "Brave new Local Development" part)

http://gos.si/blog/composer-development-with-local-dependencies

rrajkomar:
@minutephp : What you suggest in your post (which I had already read prior to creating issue #4152) is nice but would only ork for the current version you have on your machine (if you switch branches dependecies  content get switched as well).
Using the repositories option in the ~/.composer/config.json file, you can achieve the same result, you would also be able to handle branches and tags without issues (when using git, AFAIK  it would not work with other VCS, i.e. with svn you can't do it because there's no local repository)  but as mentionned in the issue it is not very intuitive.

Maybe a solution would be to have another configuration option that would work the other way around : the global config could override values defined in the project's composer.json

For the moment, since I'm working mainly with git, I can work around the issue with the reposiotires option set only in each developer's config.json (and using the same method on the different servers I use).

ahuszko:
This may help you to improve local package development: http://ahuszko.ghost.io/compact/

staabm:
@Seldaek I think we can close this one, because we have `PathRepository` now?
https://getcomposer.org/doc/05-repositories.md#path

Seldaek:
No I'd still like to investigate a better solution.. path repo has several issues IMO it's a good simple solution but not really the best.

Petah:
+1

Seldaek:
@Petah there is a subscribe button on the right to avoid spamming people with useless +1s.

Petah:
Thanks :+1:

mreschke:
The `path repo` can get tripped up sometimes for odd reasons with versioning...not a bug, just logic I can't figure out.  But for me the `path repo` works amazingly for package development.  @Seldaek what are some of the issues you are seeing?

rudiedirkx:
[...]

PR #9488, opened 2020-11-16
Introduce "links" feature

gmazzap:
This PR introduces **"link" functionality** to Composer, inspired by [npm link](https://docs.npmjs.com/cli/v6/commands/npm-link). See #601.

## New commands
- **`composer link`** command, which exceuted from a package's folder export package's path to a "link repository": a JSON file containing a map of package names to local paths. The "link repository" file is created if not exists yet. It is located by default in Composer home, but the command accept as argument an alternative path.
- **`composer unlink`** command. Specular to `link` command removes what `composer link` adds.

## New command options
- **`--load-links`** flag for `composer update` command. When used any required package that was previously exported with `composer link` will be symlinked in the installation folder from the local path instead of being pulled from the actual repository.
- **`--links-from`** option for `composer update` command. Similar to `--load-links` this allows to load packages from alternative link repositories (it is an array option): it is the companion of using `composer link` with custom path parameter. Can be used in combination with `--load-links` to load links from default link repository _and_ from custom one(s).
- **`--ignore-links`** option for `composer install` command. Instruct Composer to ignore any linked packages information eventually present in the `composer.lock`.

### On update
When `--load-links` and/or `--link-from` are used, an `ArrayRepository` is created containing all packages defined in the "links repository". This repository is pre-pended to the Installer's `RepositorySet` so packages in it will take precedence.
Moreover, packages in the link repository bypass the Solver so that linked local packages requirements are accepted for install even if the requirement does not match the local package version, which is hardcoded to `dev-local`.
Note: every linked package keeps track of the actual requirement for it, and thows an exception if it is required using a non-compatible set of requirements. In other words if `composer update` without links fails for a version conflict in the requirement for a package "foo/bar", then running `composer update --load-links` will also fail even if "foo/bar" is linked and so bypass Solver.

### On install
Considering linked packages are normal packages with a dist type of "path" and no source, there's nothing new or special about how they are locked. So on install they would normally be installed with no issue.
The `--ignore-links` flag make the `ArrayRepository` created for linked packages to be empty, which means that Composer will look for the packages in the remote repository which likely means that installation will fail. This flag can be used to avoid linked packages can be installed "by accident".

## Other changes

- The `composer.lock` has a new "links" section with information about linked packages and the link repository they are loaded from
- The `composer show` command, when using to show a single package, eventually shows information about a package being linked (e.g. the link repository path).
- A `LinkPackage` class is added. It extends `CompletePackage` and it is used to represent packages loaded from links. It carries information for the link repostory path and it is responsible to check a linked package is loaded for a compatible set of requirements.
- A `PackageLinks` class is added to take care all the link-related funcitonality, like loading packages from link repositories and instantiate the `ArrayRepository`.
- A `LinkPackageConstraintException` class is added. It is an exception thown if a linked package is required using a non compatible set of requirements.

## Is it working already?

I tested manually the most common use cases and **it appears so**.

## Is it backward compatible?

Some existing methods have now a new parameter which has a default.

This is done mostly either in constructors (where signature changes are allowed) or in internal classes (that should not be extended by 3rd party).
There are only 2 cases of non-constructor signature changes in classes not marked as internal, those are:
- `PoolBuilder::buildPool()`
- `Locaker::setLockData()`

If plugins are defining classes extending those two classes and overriding those two methods it would break. Not sure if in a project like Composer this prevents to ship the feature without major version bump.

Maybe those two classes or methods are worth to be marked internal? If not, the additional parameter passed to the two methods could be passed to their classes' constructor.

## What is missing

Unit tests. Both updating existing ones and adding new ones.

### DIFF
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -52,6 +52,7 @@ class InstallCommand extends BaseCommand
                 new InputOption('apcu-autoloader-prefix', null, InputOption::VALUE_REQUIRED, 'Use a custom prefix for the APCu autoloader cache. Implicitly enables --apcu-autoloader'),
                 new InputOption('ignore-platform-req', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Ignore a specific platform requirement (php & ext- packages).'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore all platform requirements (php & ext- packages).'),
+                new InputOption('ignore-links', null, InputOption::VALUE_NONE, 'Ignore all package links in composer.lock.'),
                 new InputArgument('packages', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, 'Should not be provided, use composer require instead to add a given package to composer.json.'),
             ))
             ->setHelp(
@@ -122,6 +123,7 @@ EOT
             ->setClassMapAuthoritative($authoritative)
             ->setApcuAutoloader($apcu, $apcuPrefix)
             ->setIgnorePlatformRequirements($ignorePlatformReqs)
+            ->setIgnoreLinks($input->getOption('ignore-links'))
         ;
 
         if ($input->getOption('no-plugins')) {
--- a/src/Composer/Command/LinkCommand.php
+++ b/src/Composer/Command/LinkCommand.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Command;
+
+use Composer\Json\JsonFile;
+use Composer\Util\Filesystem;
+use Composer\Util\Platform;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Giuseppe Mazzapica <giuseppe.mazzapica@gmail.com>
+ */
+class LinkCommand extends BaseCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('link')
+            ->setDescription('Add current package to a "links repository".')
+            ->setDefinition(array(
+                new InputArgument('path', InputArgument::OPTIONAL, 'Target path for the links repository. If not provided defaults to Composer home.'),
+            ))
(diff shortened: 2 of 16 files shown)


gmazzap:
@stof 
>wouldn't it be better for the solver to actually resolve things using the non-linked packages, to ensure that the composer.lock has actually a valid set of dependencies for the --ignore-link case ?

That means that:
- Composer has to load remote repositories, download remote packages, so consuming resources just to fill the lock. Current approach works even without internet connection if the packages you need to install are all linked locally.
- The linked packages must exists in some remote repo. One of the main reason of having this feature is to be able to develop a _new_ library being able to require it in a project before it is publicly available
- Assuming I do that, because at the end I'd need to install the linked package and not the resolved package, the lock file will contain the linked package anyway. Or `install` won't work at all with links. To make it work I would need to edit the installer to ignore locked versions for linked packages and just consider the locked linked version. But that would be probably a more "invasive" change to existing code than bypassing the Solver.

> Available links are specific to the machine on which you ran composer link. The composer.lock file is meant to be committed in your VCS and shared between all devs. This would likely lead to a non-portable lock file that can be used only on the machine which generated it.

Yes, before committing a `composer.lock` you'll need to run a `composer update` (without `--load-links`) and that make sense anyway, as it makes sense, before pushing the lock, to test your code with the packages that are available online and so that other people will install when using your lock.

Sure thing people make mistakes, so it is very possible that a `composer.lock` with locked local paths is pushed, but anyone receiving such lock could run `composer update nothing` and obtain a fresh  `composer.lock` without links and then run `composer install` with that (In current code this does not work but would not be hard to make it work).

stof:
>     2\. run `composer update --load-links` so that lock is created again, including links information, but keeping locked packages from existing lock (if available)

how would you know which package to keep from the existing lock ?

gmazzap:
> how would you know which package to keep from the existing lock?

I would have to obtain the `Locker` instance pre-update and from there obtain a `LockArrayRepository`.

When would be the time to write the new lock, inside `Locker::lockPackages` any package that was installed via links (`Locker` is already aware of `PackageLinks`) and was also part of the pre-update `LockArrayRepository` with a version that is not `dev-local` is a package that was previously installed from a remote repository but was now installed from links. So I could "dump" the package obtained from the `LockArrayRepository` into the new lock data.

gmazzap:
I realized there are issues with this PR, so I'm closing it. I hope to be able to send a better soon.

lkraav:
> I realized there are issues with this PR, so I'm closing it. I hope to be able to send a better soon.

Hi @gmazzap. Did you figure out what "better" might look like over xmas?

gmazzap:
@lkraav Yes, indeed. But had zero time to test, and did not want to send another broken PR.

lkraav:
> Yes, indeed. But had zero time to test, and did not want to send another broken PR.

Is a one-paragraph summary possible of what the new ideas, for those interested?

PR #9508, opened 2020-11-20
InstalledVersions: add more information about packages

VaclavPavek:
Added more information for listing / checks packages in APP
- is_dev
- is_abandoned

### DIFF
--- a/src/Composer/Repository/FilesystemRepository.php
+++ b/src/Composer/Repository/FilesystemRepository.php
@@ -147,6 +147,7 @@ class FilesystemRepository extends WritableArrayRepository
                 }
 
                 $versions['versions'][$package->getName()] = array(
+                    'name' => $package->getName(),
                     'pretty_version' => $package->getPrettyVersion(),
                     'version' => $package->getVersion(),
                     'aliases' => array(),
@@ -154,7 +155,6 @@ class FilesystemRepository extends WritableArrayRepository
                 );
                 if ($package instanceof RootPackageInterface) {
                     $versions['root'] = $versions['versions'][$package->getName()];
-                    $versions['root']['name'] = $package->getName();
                 }
             }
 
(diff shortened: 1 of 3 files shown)


src/Composer/Repository/FilesystemRepository.php
```diff
@@ -147,14 +147,16 @@ public function write($devMode, InstallationManager $installationManager)
                 }
 
                 $versions['versions'][$package->getName()] = array(
+                    'name' => $package->getName(),
                     'pretty_version' => $package->getPrettyVersion(),
                     'version' => $package->getVersion(),
                     'aliases' => array(),
                     'reference' => $reference,
+                    'is_dev' => $package->isDev(),
```
stof, line 155: I don't think this makes sense. This is something derived from the version (note that this is **not** telling you whether it is a dev requirement)

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -147,14 +147,16 @@ public function write($devMode, InstallationManager $installationManager)
                 }
 
                 $versions['versions'][$package->getName()] = array(
+                    'name' => $package->getName(),
```
stof, line 150: why duplicating something which is already the key in the array of versions ?

VaclavPavek: Duplicated for compatibility schema with  [InstalledVersions::getRootPackage](https://github.com/composer/composer/blob/master/src/Composer/InstalledVersions.php#L148) which return package name as array key.

stof: but the difference is that when asking for info about another package, you already know the name, as you passed it as argument

src/Composer/Repository/FilesystemRepository.php
VaclavPavek: Yeah, you're right.

src/Composer/Repository/FilesystemRepository.php
VaclavPavek: ok, rollback

VaclavPavek:
Closing this issue, I will continue to parse `composer.lock` because I need more information about the package:

- important: is dev requirement, is `abandoned`
- less important: `license`, url of `homepage` or `support`

PR #9517, opened 2020-11-23
Clarify difference in create-project related scripts

bradjones1:
I wasn't entirely clear the difference between post-root-package-install and post-create-project-cmd, so this is an effort to clarify (as to the best of my understanding.)


Seldaek:
Yup that's definitely clearer :)

PR #9537, opened 2020-12-02
Fix error with bin config in PHP8

villfa:
The `bin` config is supposed to be an array, but before PHP8 it could also be a string and it worked fine.

From PHP8 the behaviour has changed and now a warning is emitted: `Only the first byte will be assigned to the string offset`.

It's worth mentioning it affects Composer v1 & Composer v2.

### DIFF
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -65,10 +65,13 @@ class ArrayLoader implements LoaderInterface
         }
 
         if (isset($config['bin'])) {
-            foreach ((array) $config['bin'] as $key => $bin) {
+            if (!\is_array($config['bin'])) {
+                $config['bin'] = array($config['bin']);
+            }
+            foreach ($config['bin'] as $key => $bin) {
                 $config['bin'][$key] = ltrim($bin, '/');
             }
-            $package->setBinaries((array) $config['bin']);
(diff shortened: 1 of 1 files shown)


mvorisek:
I tried in sandpit - https://3v4l.org/u67G3 - can you update it to reproduce the issue?

Seldaek:
I don't quite understand this. Please give more info on how to reproduce, because casting a string to an array seems to work exactly the same on PHP 8 as per @mvorisek's 3v4l

villfa:
Here https://3v4l.org/07d5L

I'll write an unit test if you wish.

src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -65,10 +65,13 @@ public function load(array $config, $class = 'Composer\Package\CompletePackage')
         }
 
         if (isset($config['bin'])) {
-            foreach ((array) $config['bin'] as $key => $bin) {
+            if (!\is_array($config['bin'])) {
+                $config['bin'] = array($config['bin']);
+            }
+            foreach ($config['bin'] as $key => $bin) {
```
mvorisek, line 71: add one space before

Seldaek:
Ah I see now thanks, the issue was in the assignment after, not the cast. Will merge this tomorrow, I don't think it's really worth bothering with a unit test but if you like feel free.

PR #9538, opened 2020-12-03
Pool Builder: Correct unlock behavior for provide/replace and root requirements

naderman:


Seldaek:
Closed by https://github.com/composer/composer/pull/10280

PR #9540, opened 2020-12-03
fixes #9539

### LINKED ISSUE #9539: Either this file is not  a zipfile, or it constitutes one disk of a multi-part archive
Cache could be a 0-byte file due ownership issue.

https://github.com/composer/composer/blob/adf7bebf21aa10754ed37ab32c23d96f3a4e2da9/src/Composer/Cache.php#L178-L180

Which is causing unzip errors
```
[vendor/composer/tmp-6613ba18adb01391c4c6aaadf6281523]
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of vendor/composer/tmp-6613ba18adb01391c4c6aaadf6281523 or
        vendor/composer/tmp-6613ba18adb01391c4c6aaadf6281523.zip, and cannot find vendor/composer/tmp-6613ba18adb01391c4c6aaadf6281523.ZIP, period.

    The archive may contain identical file names with different capitalization (which fails on case insensitive filesystems)
    Unzip with unzip command failed, falling back to ZipArchive class
```

Steps to reproduce:

```bash
rm -rf ~/.cache/composer/files/friendsofsymfony/user-bundle/2b9c800269f4ef02df47dd466f354ed54d4a5b27.zip
touch ~/.cache/composer/files/friendsofsymfony/user-bundle/2b9c800269f4ef02df47dd466f354ed54d4a5b27.zip
```
or

```bash
chown root: ~/.cache/composer/files/friendsofsymfony/user-bundle/2b9c800269f4ef02df47dd466f354ed54d4a5b27.zip
```

and

```bash
rm -rf vendor/friendsofsymfony/user-bundle
```

oleg-andreyev:


src/Composer/Cache.php
```diff
@@ -176,7 +188,17 @@ public function copyTo($file, $target)
                 } catch (\ErrorException $e) {
                     // fallback in case the above failed due to incorrect ownership
                     // see https://github.com/composer/composer/issues/4070
-                    Silencer::call('touch', $this->root . $file);
+                    $this->io->error(
+                        sprintf(
+                            'Failed to update utime of %s probably due to incorrect ownership, skipping cache',
+                            $this->root . $file
+                        )
+                    );
```
Seldaek, line 196: This should be reverted. The silenced touch was correct IMO. It only does it if file_exists so it will not create empty files.

src/Composer/Cache.php
```diff
@@ -176,7 +188,17 @@ public function copyTo($file, $target)
...
+                }
+
+                // cached file is empty
+                if (!file_exists($this->root . $file ) || filesize($this->root . $file) === 0) {
```
Seldaek, line 200: The file_exists check makes no sense here, as we check it exists above. Overall this whole check makes no sense IMO, an empty file should not be created in the cache.

src/Composer/Cache.php
```diff
@@ -157,7 +157,19 @@ public function copyFrom($file, $source)
...
+            } catch (\ErrorException $e) {
+                // fallback in case the above failed due to incorrect ownership
+                $this->io->error(
+                    sprintf(
+                        'Failed to save cache %s probably due to incorrect ownership, skipping cache',
+                        $this->root . $file
+                    )
+                );
+
+                return false;
```
Seldaek, line 171: I don't understand why we need this, if the cache permissions are borked and we fail to write to it we should error so the user can fix their setup.

Seldaek (commented):
Overall I am not sure what this fixes, but I don't think the fix is appropriate. Let's move the discussion to the issue first to make sure the problem is correctly identified.

PR #9555, opened 2020-12-11
Change Composer Command Line Interface

novalramdhani:
change the php composer.phar command to composer


Seldaek:
I think lots of people do use it as a phar file not installed globally, so we tend to use that in the docs, but it's also easy to replace this with whichever way you run composer.. I'm not sure what's best tbh but rather not change this arbitrarily on that page.

PR #9569, opened 2020-12-23
Add example for "replace"

mbrodala:


doc/04-schema.md
```diff
@@ -439,6 +439,16 @@ above. You should then typically only replace using `self.version` as a version
 constraint, to make sure the main package only replaces the sub-packages of
 that exact version, and not any other version, which would be incorrect.
 
+Example:
+
+```json
+{
+    "replace": {
+        "acme/foo-legacy": "self.version"
```
GrahamCampbell, line 447: When replacing a legacy package which is no longer receiving releases, you typically would not want to replace with `self.version`, actually. For example, fakerphp/faker replaces 1.9.1 and 1.9.2 with the same versions, only, and future releases do not replace the old package.

This example should be replaced with a subsplit package replacement example.

GrahamCampbell (changes requested):

Seldaek:
Not sure having things people can copy paste without understanding anything will help here, or if it won't just add more confusion. Replace is a dangerous tool, which can be very easily misused, so people should read in depth..

mbrodala:
@Seldaek Maybe my intention wasn't clear: just from reading it is not clear if `replace` is a `list` or a `map` in JSON. That's what I wanted to illustrate with the example. Not sure if these terms are already well-established yet but this should be mentioned.

Seldaek:
Ok, that is technically covered in the first line of https://getcomposer.org/doc/04-schema.md#package-links but it's a long section so I can see how it might be missed :)

mbrodala:
You are right. I must admit that I never read that part. ^_~

Maybe it would help to avoid the term _Lists_ in each option?

Seldaek:
Hopefully that does the job https://github.com/composer/composer/commit/19eef374d012df3d8bb36aa85eb687675b6d5369

mbrodala:
Yes, perfect. Thanks. :-)

PR #9574, opened 2020-12-26
Add allowInstalledRepositories to installer

wachterjohannes:
This change is needed to upgrade a library which embeds composer into a phar application. It is used there to merge the internal already installed dependencies (inside of the phar) with the "external" ones.
The external ones are plugins which are installed via a "composer.json" named "nanbando.json".

For ^1.0 it worked great but with ^2.0 it needs to set the flag `allowInstalledRepositories` to allow a installed repository for updates / installs.

Linked PR:

* Embedded Composer library https://github.com/nanbando/dflydev-embedded-composer/pull/1
* Nanbando (the phar application) https://github.com/nanbando/core/pull/108

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -154,6 +154,11 @@ class Installer
     protected $additionalFixedRepository;
 
     /**
+     * @var bool
+     */
+    protected $allowInstalledRepositories = false;
+
+    /**
      * Constructor
      *
      * @param IOInterface          $io
@@ -179,6 +184,11 @@ class Installer
         $this->autoloadGenerator = $autoloadGenerator;
 
         $this->writeLock = $config->get('lock');
+    }
+
+    public function allowInstalledRepositories($allow = true)
+    {
+        $this->allowInstalledRepositories = $allow;
     }
 
     /**
@@ -769,6 +779,10 @@ class Installer
             $repositorySet->addRepository($this->additionalFixedRepository);
         }
 
+        if ($this->allowInstalledRepositories) {
+            $repositorySet->allowInstalledRepositories();
+        }
+
         return $repositorySet;
     }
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Would appreciate if you can confirm the fix, but as far as I can tell this should work.

wachterjohannes:
@Seldaek thanks for the reply - but there is a problem with my code: it uses a `CompositeRepository` which contains 2 `InstalledFilesystemRepository`'s. This seems not to work. Maybe i have to change that in my code - but the second fix inside this PR is missing for me (the `getPackage` function). would that be possible to add?

Seldaek:
So you now end up with FilterRepo -> CompositeRepo -> InstalledFilesystemRepo[], and that does not work? How so?

As for the getPackage function.. I am not sure I understand why you need this. If you create the Installer class yourself you surely have access to the root package instance? It can also be retrieved from $composer->getPackage() in most contexts.

wachterjohannes:
@Seldaek nur really - your code does not change anything for me - i passed in a `CompositeRepo -> InstalledFilesystemRepo[]` but your change would only do something for `InstalledFilesystemRepo`

but i think i have found another solution: When i use `CompositeRepo -> FilterRepo[] -> InstalledFilesystemRepo` the install/update seems to work as expected. What do you think could this be a valid solution?

Seldaek:
Yeah ok I see the problem with CompositeRepo being unwrapped by RepositorySet::addRepository.. I guess we have to conditionally set allowInstalledRepositories if the repo is installed one or is a composite which contains an installed repo. I'll look at doing that tomorrow.

Seldaek:
Ok see https://github.com/composer/composer/commit/5c35f37f92694cdff563942f3ce3e5143893b853 - feel free to try again.

wachterjohannes:
@Seldaek thanks will do 👍

PR #9580, opened 2020-12-29
feat(Cache): make cache writes more atomic

### LINKED ISSUE #9568: Random failures on CI related to file rename
We are running several builds (15-20) at the same time on the same server in parallel as part of our CI workflow. Some builds are randomly failing with the below-described errors. All builds are running in Docker, every build item spins up an isolated environment from the same images. The source code is mounted from the host, but Composer's cache is container only, it is not shared or mounted.

Not sure what other information can I disclose which could be useful to figure out what is failing exactly and why. Does this ring a bell to anybody?

My `composer.json`:

N/A

```
Checking git settings: OK
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys: 
OK
Checking composer version: OK
Composer version: 2.0.8
PHP version: 7.3.25
PHP binary path: /usr/local/bin/php
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
```

When I run this command: <!-- run it with `-vvv` added to it ideally to get full debug output -->

```
composer install foo/bar
```

I get the following output: <!-- FULL OUTPUT please, not just what you think is relevant -->

```
[ErrorException]                                                                                                                                                                                                              

rename(/home/user/.composer/cache/repo/https---repo.packagist.org/provider-php-http~promise.json.tmp,/home/user/.composer/cache/repo/https---repo.packagist.org/provider-php-http~promise.json): No such file or directory  
```

rtm-ctrlz:
Running parallel writes to cache may lead to race condition, when .tmp file is already renamed.

I'm getting error during tests via GitLab CI:
 - using docker-runners with shared composer cache (by mounting `/var/composer/cache`)
 - 3 parallel jobs (phpstan, phpcs, phpunit)
 - each make own `composer install`
 - usually all jobs pass install-phase without errors
 - sometimes one of jobs fails to install (see #9568)

May be there is other way:
 - to reproduce this race condition (it is necessary?)
 - workaround this problem with locks (ex. flock+timeout) - for me using locks is a proper way, but it would be a bit more complicated rather then just using more unique temporary filenames for every file/install.

Closes #9568

### DIFF
--- a/src/Composer/Cache.php
+++ b/src/Composer/Cache.php
@@ -114,21 +114,21 @@ class Cache
 
             $this->io->writeError('Writing '.$this->root . $file.' into cache', true, IOInterface::DEBUG);
 
+            $tempFileName = $this->root . $file . uniqid('.', true) . '.tmp';
             try {
-                $tempFileName = $this->root . $file . uniqid('.tmp.', true);
                 return file_put_contents($tempFileName, $contents) !== false && rename($tempFileName, $this->root . $file);
             } catch (\ErrorException $e) {
                 $this->io->writeError('<warning>Failed to write into cache: '.$e->getMessage().'</warning>', true, IOInterface::DEBUG);
                 if (preg_match('{^file_put_contents\(\): Only ([0-9]+) of ([0-9]+) bytes written}', $e->getMessage(), $m)) {
                     // Remove partial file.
-                    unlink($this->root . $file);
+                    unlink($tempFileName);
 
                     $message = sprintf(
                         '<warning>Writing %1$s into cache failed after %2$u of %3$u bytes written, only %4$u bytes of free space available</warning>',
-                        $this->root . $file,
+                        $tempFileName,
                         $m[1],
                         $m[2],
-                        @disk_free_space($this->root . dirname($file))
+                        @disk_free_space($this->root . dirname($tempFileName))
                     );
 
                     $this->io->writeError($message);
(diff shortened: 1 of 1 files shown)


mxr576:
Thanks @rtm-ctrlz for the patch and for confirming that the reported issues exist :)

I compiled a phar from the current state of this PR and testing it in our CI.

[composer.phar.txt](https://github.com/composer/composer/files/5754655/composer.phar.txt)

mxr576:
The first build was successful with the code in this PR, but the second failed with this error again. Maybe there is another race condition somewhere or this issue is unrelated?

https://github.com/composer/composer/issues/9568#issuecomment-752629717

mxr576:
Updated my findings in #9568, tl;dr; as of today, it looks to me this PR fixes the random failures caused by race conditions in cache writes.

src/Composer/Cache.php
```diff
@@ -115,7 +115,8 @@ public function write($file, $contents)
             $this->io->writeError('Writing '.$this->root . $file.' into cache', true, IOInterface::DEBUG);
 
             try {
-                return file_put_contents($this->root . $file.'.tmp', $contents) !== false && rename($this->root . $file . '.tmp', $this->root . $file);
+                $tempFileName = $this->root . $file . uniqid('.tmp.', true);
```
stof, line 118: this does not make them have a `.tmp` extension anymore

rtm-ctrlz: Yep, but there is no (at least I couldn't find) any code, that will use `.tmp` extension.

rtm-ctrlz:
@stof have a look, please

- temporary files will have `.tmp` extension
- also `catch` block will use correct filename

Seldaek:
Thanks, looks good to me.

PR #9589, opened 2021-01-05
Add option to configure the env of the Process

bambamboole:
Hey guys,

first of all, thanks for your brilliant work.

This PR adds the option to set custom Environment Variables to the ProcessExecuter.
These will be passed to the Process on creation.

I feature was born out of the need to set a custom GIT_SSH_COMMAND programmatically.

If you need further infos or more tests, just drop a line.

### DIFF
--- a/src/Composer/Util/ProcessExecutor.php
+++ b/src/Composer/Util/ProcessExecutor.php
@@ -40,6 +40,10 @@ class ProcessExecutor
      * @psalm-var array<int, array<string, mixed>>
      */
     private $jobs = array();
+    /**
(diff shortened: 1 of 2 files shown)


bambamboole:
@Seldaek do you have a clue why the test fails on Windows?

Seldaek:
The whole env is forwarded to child processes, so you can use putenv() to set additional env in the child process IMO. What you did here means if setEnv is called it will override the entire env vars, which may cause other issues/unexpected side effects, so I'd rather not do that.

PR #9592, opened 2021-01-07
Add response info for HTTP requests in CurlDownloader

wissem:
This PR adds a way to expose response information for HTTP requests in CurlDownloader. The response information stored is from `curl_getinfo()`.

This makes it possible to log certain data, like for instance the total response time per HTTP request.


Seldaek:
Thanks

PR #9604, opened 2021-01-12
Remove sleeps from curl handling, select it is responsible for waiting

naderman:
The current sleeps mean that large files download slowly as select would return quickly when data has arrived and needs to be processed, but the sleep waits while the buffers are full.

On the flipside we need to ensure that some code that would keep the CPU busy if run too often does not get run every time select returns.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -49,7 +49,6 @@ class CurlDownloader
         CURLM_OUT_OF_MEMORY => array('CURLM_OUT_OF_MEMORY', 'You are doomed.'),
         CURLM_INTERNAL_ERROR => array('CURLM_INTERNAL_ERROR', 'This can only be returned if libcurl bugs. Please report it to us!'),
     );
-    private $lastInfoUpdate = 0;
 
     private static $options = array(
         'http' => array(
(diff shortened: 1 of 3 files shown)


src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -390,30 +392,35 @@ public function tick()
             }
         }
 
-        foreach ($this->jobs as $i => $curlHandle) {
-            if (!isset($this->jobs[$i])) {
-                continue;
-            }
-            $curlHandle = $this->jobs[$i]['curlHandle'];
-            $progress = array_diff_key(curl_getinfo($curlHandle), self::$timeInfo);
+        if (microtime(true) - $this->lastInfoUpdate > 0.1) {
```
naderman, line 395: This bit is still a bit problematic, as we have to somehow make sure that the code below gets run at least once when all requests are complete, don't have a good solution for that yet.

naderman:
I removed the condition which executed the curl getinfo calls only if enough time had passed, which was problematic when the requests end. I did not see any negative effect on CPU usage or runtime when benchmarking, so I think this is good to merge as-is.

src/Composer/Util/Loop.php
```diff
@@ -95,15 +96,20 @@ function ($e) use (&$uncaught) {
                 $activeJobs += $this->processExecutor->countActiveJobs();
             }
 
-            if ($progress) {
+            if ($progress && microtime(true) - $lastUpdate > 0.1) {
+                $lastUpdate = microtime(true);
+                echo "setting progress\n";
```
stof, line 101: should be removed

naderman: That entire code was removed?

PR #9617, opened 2021-01-16
GitDownloader: combine checkout + reset commands into a single process

clxmstaab:
use a single process instead of 3 to improve performance

this saves 0.5-1 second while running `COMPOSER_DISABLE_NETWORK=1 php
 composer/bin/composer install -vvv --profile` on the [rector/rector project](https://github.com/rectorphp/rector)

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -454,13 +454,10 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
 
             $command = sprintf('git checkout %s --', ProcessExecutor::escape($branch));
             $fallbackCommand = sprintf('git checkout '.$force.'-B %s %s --', ProcessExecutor::escape($branch), ProcessExecutor::escape('composer/'.$branch));
-            if (0 === $this->process->execute($command, $output, $path)
-                || 0 === $this->process->execute($fallbackCommand, $output, $path)
-            ) {
-                $command = sprintf('git reset --hard %s --', ProcessExecutor::escape($reference));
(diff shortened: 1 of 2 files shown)


src/Composer/Downloader/GitDownloader.php
```diff
@@ -454,13 +454,10 @@ protected function updateToCommit($path, $reference, $branch, $date)
 
             $command = sprintf('git checkout %s --', ProcessExecutor::escape($branch));
             $fallbackCommand = sprintf('git checkout '.$force.'-B %s %s --', ProcessExecutor::escape($branch), ProcessExecutor::escape('composer/'.$branch));
-            if (0 === $this->process->execute($command, $output, $path)
-                || 0 === $this->process->execute($fallbackCommand, $output, $path)
-            ) {
-                $command = sprintf('git reset --hard %s --', ProcessExecutor::escape($reference));
```
clxmstaab, line 460: put this command into `$resetCommand` variable and put it before the if-construction.
the conditions are now evaluated in bash within a single process

clxmstaab: it seems this requires some test changes. in case you agree with the change I will try to adjust the test expectations.

Seldaek: Hmm yeah this seems fine to me. Most likely not gonna be fun to fix the tests but also shouldn't be that hard.

clxmstaab: test expectations adjusted. should be good to go.

Seldaek:
Thanks

PR #9618, opened 2021-01-16
Filesystem: added removeDirectoryAsync() and use it in ArchiveDownloader

clxmstaab:
this turns half of the `rm -rf ...` executions async and therefore improves performance.

before:
```
COMPOSER_DISABLE_NETWORK=1 php ../../../Users/mstaab/Documents/GitHub/composer/bin/composer install -vvv --profile
...
[15.5MiB/27.04s] Memory usage: 15.5MiB (peak: 17.53MiB), time: 27.04s
```

after:
```
COMPOSER_DISABLE_NETWORK=1 php ../../../Users/mstaab/Documents/GitHub/composer/bin/composer install -vvv --profile
...
[15.5MiB/25.83s] Memory usage: 15.5MiB (peak: 17.53MiB), time: 25.83s
```

### DIFF
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -113,7 +113,7 @@ class Filesystem
         // clear stat cache because external processes aren't tracked by the php stat cache
         clearstatcache();
 
-        if ($result && !file_exists($directory)) {
+        if ($result && !is_dir($directory)) {
             return true;
         }
 
@@ -151,7 +151,7 @@ class Filesystem
             clearstatcache();
 
             if ($process->isSuccessful()) {
-                if (!file_exists($directory)) {
+                if (!is_dir($directory)) {
                     return \React\Promise\resolve(true);
                 }
             }
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -131,7 +131,7 @@ class Filesystem
             $promise = $this->getProcess()->executeAsync($cmd);
 
             $self = $this;
-            return $promise->then(function ($process) use ($directory, $cmd, $self) {
+            return $promise->then(function ($process) use ($directory, $self) {
                 // clear stat cache because external processes aren't tracked by the php stat cache
                 clearstatcache();
 
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Util;
 
+use React\Promise\Promise;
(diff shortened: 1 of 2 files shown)


src/Composer/Util/Filesystem.php
```diff
@@ -126,16 +127,34 @@ public function removeDirectory($directory)
...
+
+            // clear stat cache because external processes aren't tracked by the php stat cache
+            clearstatcache();
+
+            if ($result && !file_exists($directory)) {
+                return true;
+            }
+
+            return $this->removeDirectoryPhp($directory);
+        }
```
clxmstaab, line 157: this duplicates some logic in this helper, because there is not enough context available to await the promise in non-async context, which would allow to re-use some logic between async/sync code paths

staabm:
While working on this async stuff I noticed that my workstation laptop never utilized all available cpu or IO resources. Most of the time its pending between 20-40% utilization, which makes me feel that composer could run faster when running even more things parallel (tested on a 8 threads cpu, with ssd and 16gb ram)

src/Composer/Util/Filesystem.php
```diff
@@ -92,9 +93,9 @@ public function emptyDirectory($dir, $ensureDirectoryExists = true)
      *
      * @param  string            $directory
      * @throws \RuntimeException
-     * @return bool
+     * @return bool|Promise
      */
-    public function removeDirectory($directory)
+    public function removeDirectory($directory, $async = false  )
```
Seldaek, line 98: I'd rather have a removeDirectoryAsync than a $async flag here, it'd be more explicit.

clxmstaab: just separated the sync and async case

src/Composer/Util/Filesystem.php
```diff
@@ -96,6 +97,75 @@ public function emptyDirectory($dir, $ensureDirectoryExists = true)
...
+            return \React\Promise\resolve($self->removeDirectoryPhp($directory));
+        });
+    }
+
+    /**
+     * @param string $directory
+     *
+     * @return bool|null Returns null, when no edge case was hit. Otherwise a bool whether removal was successfull
+     */
+    private function removeEdgeCases($directory) {
```
clxmstaab, line 168: extracted the common path of `removeDirectory` and `removeDirectoryAsync` into this new method.

Seldaek:
Thanks! Added https://github.com/composer/composer/commit/67261e8d4707eacf759782429c97f257d0f0d940 too which was a quick win at this point to speed things up a little further :)

PR #9619, opened 2021-01-17
Do not unnecessarily load replaced packages into the pool

driskell:
I'm seeing slow performance with composer and Drupal. I thought it may be related to the Pool Optimiser PR not having much impact on Drupal installations (see https://github.com/composer/composer/pull/9261#issuecomment-703171395) but it might not be. To be specific, this is when performing updates. It heavily impacts dependabot.

To reproduce. Take the composer.json from https://github.com/drupal/recommended-project/blob/9.2.x/composer.json
Then install `composer install --profile`
Then install a module `composer require drupal/media_entity_twitter:2.5.0 --profile`
Modify composer.json for media_entity_twitter to `^2.0`
Then update that module `composer update -W drupal/media_entity_twitter --profile` (goes to 2.6.0 currently.)

The timings I get are:

    $ composer --version
    Composer version 2.0.7 2020-11-13 17:31:06
    $ composer install --profile
    [31.7MiB/31.07s] Memory usage: 31.66MiB (peak: 228.37MiB), time: 31.07s
    $ composer require drupal/media_entity_twitter:2.5.0 --profile
    [19.1MiB/3.14s] Memory usage: 19.07MiB (peak: 20.42MiB), time: 3.14s
    $ nano composer.json
    $ composer update -W drupal/media_entity_twitter --profile
    [46.9MiB/25.46s] Memory usage: 46.86MiB (peak: 203.91MiB), time: 25.46s

As you can see the update takes a long time. When using a composer version that outputs the network requests, you can see many many 404 like these:

    [23.5MiB/4.08s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-bridge.json
    [23.5MiB/4.19s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-front-matter.json
    [23.5MiB/4.34s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-diff.json
    [23.5MiB/4.35s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-php-storage.json
    [23.6MiB/4.43s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-discovery.json
    [23.6MiB/4.52s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-file-cache.json
    [23.6MiB/4.58s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-file-security.json
    [23.6MiB/4.62s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-filesystem.json
    [23.6MiB/4.83s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-http-foundation.json
    [23.6MiB/4.87s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-graph.json
    [23.6MiB/4.95s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-gettext.json
    [23.6MiB/5.00s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-proxy-builder.json
    [23.7MiB/5.05s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-render.json
    [23.7MiB/5.06s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-uuid.json
    [23.7MiB/5.09s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-plugin.json
    [23.7MiB/5.29s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-utility.json
    [23.7MiB/5.29s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/datetime.json
    [23.7MiB/5.35s] [404] https://packages.drupal.org/files/packages/8/p2/drupal/core-transliteration.json

This goes on for a while.

Essentially the drupal/core-recommended depends on drupal/core package which has a large replaces list. This list contains the name of every core module. Whether or not it should is left for discussion elsewhere (there are reasons to do so.) This list of replaced packages for the most part do not exist - for example drupal/taxonomy is, and has always been, a core module (I think). When you run an update it sees that media_entity_twitter depends on drupal/core and thus unlocks it as a non-root dependency and marks it for load, but it then proceeds to load all the replacements - even though no other package ever requires them.

So it seems somewhat unnecessary to be loading the replaced packages. If there is indeed a require to one of them somewhere, it would make sense. Maybe I'm missing something.

With this PR the 404s disappear and the timings for the update becomes as expected:

    $ composer update -W drupal/media_entity_twitter --profile
    [29.7MiB/5.84s] Memory usage: 29.66MiB (peak: 250.26MiB), time: 5.84s

Savings of 20 seconds.

WDYT?

CC @Toflar after discussion in #9261 (Thanks!)

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -474,7 +474,6 @@ class PoolBuilder
 
                     if ($request->getUpdateAllowTransitiveRootDependencies() || !$skippedRootRequires) {
                         $this->unlockPackage($request, $repositories, $replace);
-                        $this->markPackageNameForLoading($request, $replace, $link->getConstraint());
                     } else {
                         foreach ($skippedRootRequires as $rootRequire) {
                             if (!isset($this->updateAllowWarned[$rootRequire])) {
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/avoid-loading-replaced-packages.test
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/avoid-loading-replaced-packages.test
@@ -1,17 +1,14 @@
 --TEST--
-Fixed packages and replacers get unfixed correctly (refs https://github.com/composer/composer/pull/8942)
+Do not load replaced packages into the pool unless they are required elsewhere
 
 --REQUEST--
 {
     "require": {
-        "root/req1": "*",
-        "root/req3": "*"
+        "root/req1": "*"
     },
     "locked": [
         {"name": "root/req1", "version": "1.0.0", "require": {"replacer/pkg": "1.*"}},
-        {"name": "root/req3", "version": "1.0.0", "require": {"replaced/pkg": "1.*", "dep/dep": "2.*"}},
-        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "*"}},
-        {"name": "dep/dep", "version": "2.3.5"}
+        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "*"}}
     ],
     "allowList": [
         "root/req1"
@@ -28,21 +25,15 @@ Fixed packages and replacers get unfixed correctly (refs https://github.com/comp
     [
         {"name": "root/req1", "version": "1.0.0", "require": {"replacer/pkg": "1.*"}},
         {"name": "root/req1", "version": "1.1.0", "require": {"replacer/pkg": "1.*"}},
-        {"name": "root/req3", "version": "1.0.0", "require": {"replaced/pkg": "1.*"}},
-        {"name": "root/req3", "version": "1.1.0", "require": {"replaced/pkg": "1.*"}},
         {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "*"}},
         {"name": "replacer/pkg", "version": "1.1.0", "replace": {"replaced/pkg": "*"}},
         {"name": "replaced/pkg", "version": "1.2.3"},
-        {"name": "replaced/pkg", "version": "1.2.4"},
-        {"name": "dep/dep", "version": "2.3.5"},
-        {"name": "dep/dep", "version": "2.3.6"}
+        {"name": "replaced/pkg", "version": "1.2.4"}
     ]
 ]
 
 --EXPECT--
 [
-    "root/req3-1.0.0.0 (locked)",
-    "dep/dep-2.3.5.0 (locked)",
     "root/req1-1.0.0.0",
     "root/req1-1.1.0.0",
     "replacer/pkg-1.0.0.0",
@@ -51,8 +42,6 @@ Fixed packages and replacers get unfixed correctly (refs https://github.com/comp
 
 --EXPECT-OPTIMIZED--
 [
-    "root/req3-1.0.0.0 (locked)",
-    "dep/dep-2.3.5.0 (locked)",
     "root/req1-1.1.0.0",
     "replacer/pkg-1.1.0.0"
 ]
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test
@@ -101,7 +101,6 @@ Check that replacers from additional repositories are loaded when doing a partia
     "indirect/replacer-1.0.0.0",
     "replacer/package-1.2.0.0",
     "replacer/package-1.0.0.0",
-    "base/package-1.0.0.0",
     "shared/dep-1.0.0.0",
     "shared/dep-1.2.0.0"
 ]
@@ -112,6 +111,5 @@ Check that replacers from additional repositories are loaded when doing a partia
     "indirect/replacer-1.0.0.0",
     "replacer/package-1.2.0.0",
     "replacer/package-1.0.0.0",
-    "base/package-1.0.0.0",
     "shared/dep-1.2.0.0"
 ]


src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -399,7 +399,6 @@ private function loadPackage(Request $request, PackageInterface $package, $propa
                 if (isset($this->loadedPackages[$replace], $this->skippedLoad[$replace])) {
                     if ($request->getUpdateAllowTransitiveRootDependencies() || !$this->isRootRequire($request, $this->skippedLoad[$replace])) {
                         $this->unlockPackage($request, $replace);
-                        $this->markPackageNameForLoading($request, $replace, $link->getConstraint());
```
stof, line 402: this means that this constraint won't be taken into account anymore if we then have a reason to load that package. Is this correct @naderman ?

driskell: Absolutely need some insight on this. From what I can gather, this just means the REPLACED package isn't loaded. Which should be fine as there should still be a REQUIRE for it somewhere that causes it to load. Because of the skippedLoad check it only does the unlock if it was locked initially. And I think if a REQUIRE was hit, it would unlock it already.

Essentially I'm just trying to prevent composer loading packages that are never referenced anywhere except with a REPLACE. As with `drupal/core-recommended` the long list of replace pretty much don't exist. Ideally composer would only load a package referenced by REPLACE if it was also referenced by a REQUIRE. Since it loads everything in a REQUIRE anyway it means don't need to do anything for REPLACE. (I think!)

stof: @driskell but if later (or ealier) we decide to load the package due to a `require` rule, the constraint of the replace rule is now totally ignored. That's what I want to confirm whether it is safe or no.

naderman: I agree with what this is trying to do, but I do think this may cause bugs, so I'll try and come up with a test case for a scenario where this causes a problem so we can figure out a way to address that.

kadirselcuk (approved):

kadirselcuk (approved):

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers.test
```diff
@@ -46,7 +46,5 @@ Fixed packages and replacers get unfixed correctly (refs https://github.com/comp
     "root/req1-1.0.0.0",
     "root/req1-1.1.0.0",
     "replacer/pkg-1.0.0.0",
-    "replacer/pkg-1.1.0.0",
-    "replaced/pkg-1.2.3.0",
-    "replaced/pkg-1.2.4.0"
```
Seldaek, line 51: In this test for example, the replaced/pkg is required by root/req3, so I am not sure why it does not get loaded, I guess because root/req3 is not in allowList, but I wonder if this is fine or could be a bug in some cases. Anyway will let @naderman think about it some more ;)

driskell: Yes, the replaced package is not in the allow list, as we've already loaded replacer/pkg so the constraints are met in the pool. Loading the replaced package wouldn't serve anything I think.

If you imagine a scenario where "replaced/pkg" does not exist at all - then this avoids a 404 too, which is the goal.

driskell:
Hi @naderman @Seldaek 
Do you have any other questions about the change? Is it worth rebasing time check any new tests?
Would be great to work out what might be able to add confidence to this as I appreciate it’s quite deep in the critical areas.

src/Composer/DependencyResolver/PoolBuilder.php
Seldaek: I just rebased the PR to make sure it still passes given we added tests, and it does.. but I am still unsure whether the case you described above as second case is fine or not. 

I did not manage to create a test which breaks it, but I'm not sure if I missed something.. may need to come back to this with a fresher mind.

Below is a sample test file which I did not add as I don't think it tests anything of value since it didn't break it.. But maybe @naderman or anyone else can spot a mistake that'd turn it into a failure here, or at least confirm the value of the test?

```
--TEST--
Fixed packages and replacers get unfixed correctly even if a replaced package is then required after a loaded replacer unfixed it

--REQUEST--
{
    "require": {
        "root/dep": "*",
        "dep/on-replaced": "*"
    },
    "locked": [
        {"name": "root/dep", "version": "1.1.0", "require": {"replacer/pkg": "1.*"}},
        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "1.0.3"}},
        {"name": "dep/on-replaced", "version": "1.0.0", "require": {"intermediate/pkg": "1.*"}},
        {"name": "intermediate/pkg", "version": "1.0.0", "require": {"replaced/pkg": "1.0.*"}}
    ],
    "allowList": [
        "root/dep",
        "dep/on-replaced"
    ],
    "allowTransitiveDeps": true
}

--FIXED--
[
]

--PACKAGE-REPOS--
[
    [
        {"name": "root/dep", "version": "1.0.0", "require": {"replaced/pkg": "1.0.*"}},
        {"name": "root/dep", "version": "1.1.0", "require": {"replacer/pkg": "1.*"}},
        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "1.0.3"}},
        {"name": "replaced/pkg", "version": "1.0.0"},
        {"name": "replaced/pkg", "version": "1.0.1"},
        {"name": "replaced/pkg", "version": "1.0.3"},
        {"name": "replaced/pkg", "version": "1.0.4"},
        {"name": "replaced/pkg", "version": "1.1.0", "replace": {"foo/bar": "1.*"}},
        {"name": "foo/bar", "version": "1.2.0"},
        {"name": "dep/on-replaced", "version": "1.0.0", "require": {"intermediate/pkg": "1.*"}},
        {"name": "intermediate/pkg", "version": "1.0.0", "require": {"replaced/pkg": "1.0.*"}}
    ]
]

--EXPECT--
[
    "root/dep-1.0.0.0",
    "root/dep-1.1.0.0",
    "dep/on-replaced-1.0.0.0",
    "replacer/pkg-1.0.0.0",
    "replaced/pkg-1.0.0.0",
    "replaced/pkg-1.0.1.0",
    "replaced/pkg-1.0.3.0",
    "replaced/pkg-1.0.4.0",
    "intermediate/pkg-1.0.0.0"
]

--EXPECT-OPTIMIZED--
[
    "root/dep-1.0.0.0",
    "root/dep-1.1.0.0",
    "dep/on-replaced-1.0.0.0",
    "replacer/pkg-1.0.0.0",
    "replaced/pkg-1.0.4.0",
    "intermediate/pkg-1.0.0.0"
]
```

driskell: The more I think the more I think a replacement spec should not affect the loaded packages in the pool at all. So it’s only the unlocking that matters (or doesn’t, but isn’t what’s at question).

If you have: Root require A. A require B. Root require C. No replace. Pool is A, B, C of versions matching require constraints.

Question could be, is there a way we can add one or more replace spec anywhere in A, B or C such that the above pool of A B C packages is insufficient and so we need to load more?

No matter how many replace you add to any of the packages in the above list I cannot see any resulting pool that isn’t a subset. It’s maybe kind of same as for conflict in that it only changes the solution and does not expand the  needed packages for that solution. In a way perhaps a require is the only thing that expands the pool.

naderman:
```
--TEST--
Ensure that a package gets loaded which was previously skipped due to replacement

--REQUEST--
{
    "require": {
        "root/dep": "*",
        "root/no-update": "*"
    },
    "locked": [
        {"name": "root/dep", "version": "1.1.0", "require": {"replacer/pkg": "1.*"}},
        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "1.0.0"}},
        {"name": "root/no-update", "version": "1.0.0", "require": {"replaced/pkg": "1.0.0"}}
    ],
    "allowList": [
        "root/dep"
    ],
    "allowTransitiveDepsNoRootRequire": true
}

--FIXED--
[
]

--PACKAGE-REPOS--
[
    [
        {"name": "root/dep", "version": "1.2.0", "require": {"replacer/pkg": ">=1.1.0"}},
        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "1.0.0"}},
        {"name": "replacer/pkg", "version": "1.1.0"},
        {"name": "replaced/pkg", "version": "1.0.0"},
        {"name": "root/no-update", "version": "1.0.0", "require": {"replaced/pkg": "1.0.0"}}
    ]
]

--EXPECT--
[
    "root/no-update-1.0.0.0 (locked)",
    "root/dep-1.2.0.0",
    "replacer/pkg-1.1.0.0",
    "replaced/pkg-1.0.0.0"
]

--EXPECT-OPTIMIZED--
[
    "root/no-update-1.0.0.0 (locked)",
    "root/dep-1.2.0.0",
    "replacer/pkg-1.1.0.0",
    "replaced/pkg-1.0.0.0"
]
```

This is roughly what I had in mind. The upgrade of replacer/pkg here removes the replace, so that the original name (replaced/pkg) needs to be installed. I thought that is what the mark packages for loading call being removed in this PR was for. But this test even fails on main, so either I'm wrong, or there is actually another bug prohibiting this from working on main?

naderman: And to add to that, the opposite case may be interesting too. Should upgrading a package to a version which now replaces a different locked package not result in that package getting unlocked/loaded and potentially removed? But I suppose the unlock call remains so that may still work as desired?

stof:
`replaces` and `provides` should indeed not mark things for loading if they get unlocked IMO, as in a normal pool without any locking done, those don't trigger loading of metadata AFAIK.

naderman:
@stof see https://github.com/composer/composer/pull/9619#discussion_r760237931 ?

src/Composer/DependencyResolver/PoolBuilder.php
driskell: I took the above test, "Ensure that a package gets loaded which was previously skipped due to replacement", and applied it into a branch and investigated the failure on main, and the below code solved it. Effectively composer isn't loading a replaced package if the replacer is unlocked. When a package is unlocked the code below will check to see if this package is currently locked and replacing - if so then it assumes that replacement might change and so it marks the replacement for loading.

https://github.com/composer/composer/compare/main...driskell:load-replaced-package-if-replacer-dropped
If this is worthy of a PR let me know - I didn't want to raise it without checking here as there's so much context.

Interestingly, after I did the above I then proceeded to retest this change on top of it and it was still fine - and in fact I was able to delete the entire section (469 onwards in this PR) without breaking any tests, and providing the same speed benefit to Drupal updates. So either there's still some missing tests for this code (469 onwards) or the reasons it existed have been duplicated elsewhere so it is not necessary anymore... Not entirely confident in myself on that matter though 😬

Seldaek: I didn't completely follow but I'd say yes please open a PR for the change in `load-replaced-package-if-replacer-dropped`, if we can get that merged first maybe it'll help simplify this PR.

driskell: I opened a PR for that branch with the test and what I think was the fix.

So what's remaining for this PR I guess is to find a test for the below code or see what it's trying to do, as deleting it currently still passes all tests, along with greatly improving the performance for Drupal site module updates due to preventing 404s when trying to load replaced packages that don't exist.

https://github.com/composer/composer/blob/main/src/Composer/DependencyResolver/PoolBuilder.php#L468-L489

driskell:
@stof @naderman Regarding comment https://github.com/composer/composer/pull/9619#discussion_r760237931

I initially investigated this and found it was related to a different area of code: https://github.com/composer/composer/pull/9619#discussion_r776339911

As per comment https://github.com/composer/composer/pull/9619#discussion_r777023213 I raised PR #10410 with that test and also what I think is the fix for the test.

So I'm unsure if there's anything remaining outstanding except a closer scrutiny on the change in this PR, which as per the last comment above could be increased in scope to delete the entire section of code below without failing a single test. So I think the remaining questions are: Is this code that slows down composer needed? Is there a test that will fail without it?

https://github.com/composer/composer/blob/24ce1eddbdb4658e1d106a5d982951342f86f4d3/src/Composer/DependencyResolver/PoolBuilder.php#L468-L489

jeffwidman:
👋 Hi from the Dependabot team.

I was digging into a few reports we've received of timeouts during composer updates, particularly drupal and stumbled across this PR.

Given the response from @driskell above, what else is needed to move this forward? s it just lack of maintainer time? (which as a fellow maintainer of several OSS projects, I realize is _always_ in short supply!)

naderman:
Yeah that's all it is, I've had these on my list for a while and am really sorry I still haven't gotten to it ...

markdorison:
Hi all! We are still battling with the Dependabot timeouts on Drupal projects that we _hope_ this may resolve. As Jeff mentioned, I completely understand your time is in short supply but I wanted to add a _friendly_ bump to this. ❤️

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: If you rebase on `main` you'll notice that you cannot remove this code - it's been covered with some tests in the meantime :)

Toflar:
I did some in-depth analysis of this PR (finally - sorry @driskell) and I'm fairly confident that the approach is actually correct! There's no point in marking the `replace` links as required for download (`markPackageNameForLoading()`). 
Basically, a `replace` target is optional until some other package `require`s it. Because if not, then there's no point in loading something that is replaced anyway. So we can really just remove this one line 👍 
There are only two tests that need adjustments (and actually prove that there are useless packages in the pool).
So once this PR is rebased, it should look like this:

```diff
Index: src/Composer/DependencyResolver/PoolBuilder.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/Composer/DependencyResolver/PoolBuilder.php b/src/Composer/DependencyResolver/PoolBuilder.php
--- a/src/Composer/DependencyResolver/PoolBuilder.php	(revision 11879ea737978fabb8127616e703e571ff71b184)
+++ b/src/Composer/DependencyResolver/PoolBuilder.php	(date 1683036654517)
@@ -483,7 +483,6 @@
 
                     if ($request->getUpdateAllowTransitiveRootDependencies() || !$skippedRootRequires) {
                         $this->unlockPackage($request, $repositories, $replace);
-                        $this->markPackageNameForLoading($request, $replace, $link->getConstraint());
                     } else {
                         foreach ($skippedRootRequires as $rootRequire) {
                             if (!isset($this->updateAllowWarned[$rootRequire])) {
Index: tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test	(revision 11879ea737978fabb8127616e703e571ff71b184)
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test	(date 1683024149715)
@@ -101,7 +101,6 @@
     "indirect/replacer-1.0.0.0",
     "replacer/package-1.2.0.0",
     "replacer/package-1.0.0.0",
-    "base/package-1.0.0.0",
     "shared/dep-1.0.0.0",
     "shared/dep-1.2.0.0"
 ]
@@ -112,6 +111,5 @@
     "indirect/replacer-1.0.0.0",
     "replacer/package-1.2.0.0",
     "replacer/package-1.0.0.0",
-    "base/package-1.0.0.0",
     "shared/dep-1.2.0.0"
 ]
Index: tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers.test
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers.test b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers.test
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers.test	(revision 11879ea737978fabb8127616e703e571ff71b184)
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers.test	(date 1683024343988)
@@ -46,9 +46,7 @@
     "root/req1-1.0.0.0",
     "root/req1-1.1.0.0",
     "replacer/pkg-1.0.0.0",
-    "replacer/pkg-1.1.0.0",
-    "replaced/pkg-1.2.3.0",
-    "replaced/pkg-1.2.4.0"
+    "replacer/pkg-1.1.0.0"
 ]
 
 --EXPECT-OPTIMIZED--
@@ -56,6 +54,5 @@
     "root/req3-1.0.0.0 (locked)",
     "dep/dep-2.3.5.0 (locked)",
     "root/req1-1.1.0.0",
-    "replacer/pkg-1.1.0.0",
-    "replaced/pkg-1.2.4.0"
+    "replacer/pkg-1.1.0.0"
 ]
```

stof:
@Toflar how does this analysis interact with https://github.com/composer/composer/pull/10410 ?

Toflar:
@stof actually, your comment made me think about "why do we need `markPackageNameForLoading()` in https://github.com/composer/composer/pull/10410 but here we want to get rid of it for `replace` packages.
So I've incorporated both PRs in another approach in https://github.com/composer/composer/pull/11449. 

I hope this makes it much more readable because all it is is basically not calling `markPackageNameForLoading()` for `replace` definitions but only load them if any other package `require`s one of them 😊 Should also be easier to understand from the diff - I hope.

src/Composer/DependencyResolver/PoolBuilder.php
naderman: > And to add to that, the opposite case may be interesting too. Should upgrading a package to a version which now replaces a different locked package not result in that package getting unlocked/loaded and potentially removed? But I suppose the unlock call remains so that may still work as desired?

This is already covered in https://github.com/composer/composer/blob/main/tests/Composer/Test/Fixtures/installer/update-allow-list-with-dependencies-require-new-replace.test

naderman: I think there may be another bug relating to multiple replacers replacing different versions of the same package and doing a partial update of it, but that's such an extrem edge case, and definitely already problematic in main, that I don't think it's worthwhile looking into now and further delaying this work here.

naderman:
Same question for @markdorison

markdorison:
I can test this once the latest changes from #11449 make it into Dependabot! @jeffwidman may be able to help with that or at least let us know when it has been included. 🥳

jeffwidman:
Historically we've only used released `composer` versions to keep life simple, they get picked up by :dependabot: :
https://github.com/dependabot/dependabot-core/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+%22composer%2Fcomposer%22+

If you want to test ahead of time, you could theoretically do it locally, but my guess is it'd be painful to wire that in since our whole build wiring assumes a released version, eg https://github.com/dependabot/dependabot-core/blob/36c7ccb07121ae9d574b264c8bc4b27cb4b65bf3/composer/Dockerfile#L3

driskell:
I can confirm this is fixed by the changes in `main`! No more unexpected 404s.

Profile results of the OP from running on `2.5.5` (using Drupal fixed to `9.2.*` as that's a branch with `replaces` still in place):

```
[20.4MiB/7.64s] Memory usage: 20.36MiB (peak: 79.86MiB), time: 7.64s
[20.4MiB/8.06s] Memory usage: 20.36MiB (peak: 79.86MiB), time: 8.06s
[20.4MiB/7.77s] Memory usage: 20.36MiB (peak: 79.86MiB), time: 7.77s
```

Using commit 595559f68d8c01715a2be33a54247898f9fbc34e on `main` containing the changes:

```
[20.3MiB/3.99s] Memory usage: 20.28MiB (peak: 75.13MiB), time: 3.99s
[20.3MiB/3.77s] Memory usage: 20.28MiB (peak: 75.13MiB), time: 3.77s
[20.3MiB/3.69s] Memory usage: 20.28MiB (peak: 75.13MiB), time: 3.69s
```

And when modifying the CurlDownloader to log requests, all the unexpected 404s are gone.

I'll post in the Dependabot ticket some timings for Dependabot too. I tested `9.5.*` too which no longer has `replaces` and it still benefits (and still has 404s on `2.5.5`) and it is faster as well (though the unpatched version is already a couple of seconds faster, so the gains are a little less.)

Toflar:
🥳
by the way: you don't need to modify anything to log requests, it should be visible when running with `-vvv` 😊

driskell:
> 🥳
> by the way: you don't need to modify anything to log requests, it should be visible when running with `-vvv` 😊

Hahaha 😆 Yeh I don't know if I forgot about that or the output was too noisy and I was too lazy to grep but I'll definitely remember that 😊

PR #9632, opened 2021-01-21
Expose COMPOSER_* paths to scripts as env vars

nicolas-grekas:
It has always bugged me as to why this was missing :)

To me, the main benefit of running `composer something` should be that composer would create a nice environment to make running scripts easy, as in eg: easy to spot where the vendor dir is.

This PR exposes all `COMPOSER_*` paths as env vars for this purpose.

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -17,6 +17,7 @@ use Composer\Installer\InstallerEvent;
 use Composer\IO\IOInterface;
 use Composer\Composer;
 use Composer\DependencyResolver\Operation\OperationInterface;
+use Composer\Factory;
 use Composer\Repository\RepositoryInterface;
 use Composer\Script;
 use Composer\Installer\PackageEvent;
@@ -155,6 +156,8 @@ class EventDispatcher
         $listeners = $this->getListeners($event);
 
         $this->pushEvent($event);
+
+        $this->setupEnv();
 
         $return = 0;
         foreach ($listeners as $callable) {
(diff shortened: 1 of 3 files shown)


Seldaek:
Got an example of a use case of a script which would benefit from this? Just wondering..

nicolas-grekas:
Sure, here is one that works in practice, but is quite fragile:
https://github.com/sebastianbergmann/phpunit/blob/5d63ba878d458bcd0e047a5b24806651d21ce5ef/phpunit#L30

Seldaek:
Yeah ok, although this kind of pattern won't be fixed unless the script requires that you run it via `composer exec phpunit`, which I doubt sebastian would want to do.

stof:
`__DIR__ . '/../../autoload.php'` is the only one you need to care about when dealing with the fact that your package is installed as a dependency. There is nothing fragile in it.

`__DIR__ . '/vendor/autoload.php'` is about covering the usage of this script inside PHPUnit itself (using it with PHPUnit as the root repository).
I have no idea what `__DIR__ . '/../vendor/autoload.php'` is about (probably a misunderstanding about how composer installs binaries when it cannot do symlinks according to the commit message)

nicolas-grekas:
> is the only one you need to care about when dealing with the fact that your package is installed as a dependency. There is nothing fragile in it.

It's fragile when `bin-dir` is changed AFAIK.

> this kind of pattern won't be fixed unless the script requires that you run it via composer exec phpunit, which I doubt sebastian would want to do.

Sure, but at least it would make running `composer test` a reality. The benefit of not having to play guesswork (thus enforcing a convention on the whole ecosystem) is worth it IMHO.

stof:
@nicolas-grekas no it is not. These paths are not resolving things relatively to the bin-dir but to the location of the file, which is still inside `vendor/phpunit/phpunit`

nicolas-grekas:
Argh, you're right.

Maybe this is not worth it then.

Seldaek:
Yeah I am thinking this may not be worth it indeed. Let's close for now, we can come back to it if a real use case appears :)

PR #9634, opened 2021-01-21
Remove dependency on wget from install docs

UlrichEckhardt:
Some people have wget, others have cURL, but we all have PHP. ;)

It looks silly, but e.g. the `debian:buster-slim` Docker image has neither `curl` nor `wget` installed by default.

### DIFF
--- a/doc/faqs/how-to-install-composer-programmatically.md
+++ b/doc/faqs/how-to-install-composer-programmatically.md
@@ -9,7 +9,7 @@ An alternative is to use this script which only works with UNIX utilities:
 ```bash
 #!/bin/sh
 
-EXPECTED_CHECKSUM="$(wget -q -O - https://composer.github.io/installer.sig)"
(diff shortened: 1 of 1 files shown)


Seldaek:
The problem is some envs also do not have a properly configured PHP / cacert setup so SSL isn't working properly. So wget kinda ensures this works. It's very hard to have a one liner which works everywhere though, and I'm not sure it's up to us to do this, if you're in a very restricted environment you most likely know what you are doing and will be capable of tweaking this, or getting curl/wget installed. 

So I'd tend to say we should close this, but happy to hear more from others.

UlrichEckhardt:
It's fine either way. I just thought it's simpler this way, because in the next line you're doing almost the same for a different URL.

BTW, while I'm at it, thank you for your work!

Seldaek:
Hm yeah you are right, if it is going to fail with php streams it'd fail anyway at the next line, so this seems safe enough. Thanks!

PR #9635, opened 2021-01-21
Ensure InstalledVersions reports info about all currently registered class loaders

nicolas-grekas:
Fixes issues like https://github.com/getsentry/sentry-symfony/issues/383

(the phpstan failure is a false-positive)

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -838,7 +838,7 @@ PLATFORM_CHECK;
 
         $file .= <<<CLASSLOADER_INIT
         spl_autoload_register(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'), true, $prependAutoloader);
-        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader();
+        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader(\\dirname(\\dirname(__FILE__)));
         spl_autoload_unregister(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'));
 
 
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -12,6 +12,7 @@
 
 namespace Composer;
 
+use Composer\Autoload\ClassLoader;
 use Composer\Semver\VersionParser;
 
 /**
@@ -22,6 +23,8 @@ use Composer\Semver\VersionParser;
 class InstalledVersions
 {
     private static $installed;
+    private static $canGetVendors;
+    private static $installedByVendor = array();
 
     /**
      * Returns a list of all package names which are present, either by being installed, replaced or provided
@@ -31,7 +34,17 @@ class InstalledVersions
      */
     public static function getInstalledPackages()
     {
-        return array_keys(self::$installed['versions']);
+        $packages = array();
+        foreach (self::getInstalled() as $installed) {
+            $packages[] = array_keys($installed['versions']);
+        }
+
+
+        if (1 === \count($packages)) {
+            return $packages[0];
+        }
+
+        return array_keys(array_flip(\call_user_func_array('array_merge', $packages)));
     }
 
     /**
@@ -44,7 +57,13 @@ class InstalledVersions
      */
     public static function isInstalled($packageName)
     {
-        return isset(self::$installed['versions'][$packageName]);
+        foreach (self::getInstalled() as $installed) {
+            if (isset($installed['versions'][$packageName])) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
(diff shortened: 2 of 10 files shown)


nicolas-grekas:
(the phpstan failure is a false-positive)

Jean85:
Thank you Nicolas, fixing this here at the root is surely a catch-all for these bugs!

ondrejmirtes:
FYI I fixed most PHPStan false positives in [0.12.69](https://github.com/phpstan/phpstan/releases/tag/0.12.69). There's one remaining in `InstalledVersions::getInstalled()` because PHPStan doesn't understand the code structure that uses `method_exists` like that. Feel free to put `@phpstan-ignore-line` in there (https://phpstan.org/user-guide/ignoring-errors).

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -838,7 +838,7 @@ public static function getLoader()
 
         $file .= <<<CLASSLOADER_INIT
         spl_autoload_register(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'), true, $prependAutoloader);
-        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader();
+        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader(\\dirname(__DIR__));
```
Seldaek, line 841: To be consistent with https://github.com/composer/composer/pull/9433 - I think we should use `dirname(dirname(__FILE__))` here for until Composer 2.2, regardless of how absurd using PHP 5.2 may seem at this point.

src/Composer/InstalledVersions.php
```diff
@@ -192,5 +229,32 @@ public static function getRawData()
...
+            foreach (ClassLoader::getRegisteredLoaders() as $vendor => $loader) {
+                if (isset(self::$installedByVendor[$vendor])) {
+                    $installed[] = self::$installedByVendor[$vendor];
+                } elseif (is_file($vendor.'/composer/installed.php')) {
+                    $installed[] = self::$installedByVendor[$vendor] = require $vendor.'/composer/installed.php';
+                }
+            }
+        }
+
+        $installed[] = self::$installed;
```
Seldaek, line 256: I'm not entirely sure how good this is to append self::$installed here. It means if someone uses reload() and has newly fixed classloaders, the installed.php data will prevail over whatever was passed to reload(). I can't think of any case where it's an issue, and IMO we could even deprecate reload() as its only purpose was allowing people to fix what this PR fixes. But strictly speaking it's breaking things, so I wonder if we should not prepend self::$installed instead of appending. This would also mean that the InstalledVersions.php should not be dumped with self::$installed pre-populated anymore, which anyway makes sense I guess.

src/Composer/InstalledVersions.php
```diff
@@ -157,7 +192,9 @@ public static function getReference($packageName)
      */
     public static function getRootPackage()
     {
-        return self::$installed['root'];
+        $installed = self::getInstalled();
+
+        return $installed[0]['root'];
```
Seldaek, line 197: This is a tricky one.. How do we determine which root package is the most root package? If for example you use phpunit, which loads phpunit's autoloader and then the project's autoloader, as prepend-autoloader is true by default, we end up with the project's root package being in installed[0] here, which sounds good to me. But if the project for some reason has prepend-autoloader false, or it's a different setup where the project includes another internal autoloader which has prepend true, then things may not be what one expects..

I don't know if this needs fixing or not, nor how to even fix it. Just starting a discussion..

nicolas-grekas: I went through this reasoning also, tricky one :)

That's what made me write the logic that ensures that the order in `ClassLoader::$registeredLoaders` is consistent with the `$prepend` argument.

So, on my side, things are as I expect them to be.

src/Composer/InstalledVersions.php
nicolas-grekas: It would be me, I would deprecate `reload()` (and I would also stop populating the `$installed` property...)

I think the current way is what is the most useful - aka dynamic data first, static one last.

src/Composer/Autoload/AutoloadGenerator.php
nicolas-grekas: updated

nicolas-grekas:
> FYI I fixed most PHPStan false positives in 0.12.69

Thanks @ondrejmirtes, PR updated!

src/Composer/InstalledVersions.php
```diff
@@ -192,5 +229,33 @@ public static function getRawData()
...
+
+        $installed = array();
+
+        if (self::$canGetVendors) {
+            // @phpstan-ignore-next-line
+            foreach (ClassLoader::getRegisteredLoaders() as $vendor => $loader) {
+                if (isset(self::$installedByVendor[$vendor])) {
+                    $installed[] = self::$installedByVendor[$vendor];
+                } elseif (is_file($vendor.'/composer/installed.php')) {
+                    $installed[] = self::$installedByVendor[$vendor] = require $vendor.'/composer/installed.php';
```
naderman, line 252: Can you rename `$vendor` to `$vendorDir` here, was a bit confusing on first read that `$vendor` doesn't refer to an individual package vendor.

src/Composer/InstalledVersions.php
Seldaek: Ok actually we do use/need reload internally too for plugins, so probably can't deprecate it:

https://github.com/composer/composer/blob/e5e9c2cd59d33f4b2a30a53e03fa378380c79f14/src/Composer/Repository/FilesystemRepository.php#L212-L216

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -838,7 +838,7 @@ public static function getLoader()
 
         $file .= <<<CLASSLOADER_INIT
         spl_autoload_register(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'), true, $prependAutoloader);
-        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader();
+        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader(\\dirname(\\dirname(__FILE__)));
```
Seldaek, line 841: Instead of changing ClassLoader at all, it would probably be cleaner to add something like this to autoload_real.php:

```
\Composer\InstalledVersions::add(require './installed.php', $prependAutoloader);
```

Then this new method on InstalledVersions would build up the array of installed versions as autoload.php files get included, and the whole InstalledVersions::getInstalled / ClassLoader::getRegisteredLoaders indirection would not be needed.

Seldaek: Actually maybe `::add($vendorDir, ...)` so we can also allow `remove($vendorDir)` in case someone needs this for some reason.

src/Composer/InstalledVersions.php
nicolas-grekas: sure, renamed!

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: Ok just to sum up some internal discussions, keeping it as is so ClassLoaders deregistering is automatically reflected, and we avoid loading the InstalledVersions class for nothing when not used by the project.

src/Composer/InstalledVersions.php
Seldaek: As for self::$installed yes let's keep it for compat with people not using ClassLoader, as well as to support reload() as a use case like my comment above shows.

Seldaek:
Alright, added https://github.com/composer/composer/commit/51371944e1c95b0fa3bc4ed921a366f553938ef3 for posterity, thanks!

PR #9637, opened 2021-01-21
Handle "versions" option in PathRepository, remove support for "branch-version"

nicolas-grekas:
Allows easily creating versioned package repositories out of collections of directories.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -658,6 +658,26 @@ the branch or tag that is currently checked out. Otherwise, the version should
 be explicitly defined in the package's `composer.json` file. If the version
 cannot be resolved by these means, it is assumed to be `dev-master`.
 
+When the version cannot be inferred from the local VCS repository, or when you
+want to override the version, you can use the `versions` option when declaring
+the repository:
+
+```json
+{
+    "repositories": [
+        {
+            "type": "path",
+            "url": "../../packages/my-package",
+            "options": {
+                "versions": {
+                    "my/package": "4.2-dev"
+                }
+            }
+        }
+    ]
+}
+```
+
 The local package will be symlinked if possible, in which case the output in
 the console will read `Symlinking from ../../packages/my-package`. If symlinking
 is _not_ possible the package will be copied. In that case, the console will
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -658,17 +658,6 @@ the branch or tag that is currently checked out. Otherwise, the version should
 be explicitly defined in the package's `composer.json` file. If the version
 cannot be resolved by these means, it is assumed to be `dev-master`.
 
-When the version cannot be inferred from the local VCS repository, you should use
-the special `branch-version` entry under `extra` instead of `version`:
-
-```json
-{
-    "extra": {
-        "branch-version": "4.2-dev"
-    }
-}
-```
-
 The local package will be symlinked if possible, in which case the output in
 the console will read `Symlinking from ../../packages/my-package`. If symlinking
 is _not_ possible the package will be copied. In that case, the console will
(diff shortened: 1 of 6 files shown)


naderman:
Can you explain the use case a bit better? I don't see why yet another way to specify versions is useful? If this PR was to be considered, can you add some corresponding entry in end user docs that explains what this is for?

nicolas-grekas:
I would use this when referencing local paths that don't use `version` (as I'm never going to use that entry anyway), and when I want tighter control over what `branch-alias` / `branch-version` provide. I would also use this to actually override what `branch-alias` / `branch-version` provide.

In Symfony, this could be a better option than using `branch-version` actually, don't you think?
THen we could remove `branch-version` entirely as it's relatively new and only Symfony uses it :)

Seldaek:
It indeed does seem like [nobody else is using branch-version](https://github.com/search?l=JSON&p=96&q=branch-version&type=Code) so far, so yeah if we remove it I think I'd prefer this solution, because branch-version never seemed like a good solution to me, it seemed conflicting with branch-alias and rather messy.

Right now it's used in two places, PathRepo and RootPackageLoader, so if you rely on the latter use case @nicolas-grekas I guess you'd have to switch back to COMPOSER_ROOT_VERSION env var for the root package, which anyway is clearer IMO.

If that works for you, please update the PR and also re-target it to the 1.10 branch.

nicolas-grekas:
PR updated!

Seldaek:
Thanks

naderman:
Great, like this solution!

baschny:
@Seldaek:

> It indeed does seem like [nobody else is using branch-version](https://github.com/search?l=JSON&p=96&q=branch-version&type=Code) so far, so yeah if we remove it I think I'd prefer this solution, because branch-version never seemed like a good solution to me, it seemed conflicting with branch-alias and rather messy.

How do you come to the conclusion that "nobody else is using" it?? This was the recommended way of handling local packages (it was documented so), and we use it for all our projects. If you just scanned the public available packages in packagist this might be the reason you not finding it in use there!

Now this change was introduced in 2.0.9 (a patch level release!!!) breaking all our projects and CI processes. And you don't even mention this breaking change in the CHANGELOG, which is even more frustrating, as I had to find out my reading the commits and merge request.

Please consider this kind of situation and change by not breaking stuff in patch level releases. After all, there will probably be a place for a "2.1.x" and later composer releases, which could allow us a better planning when updating our projects.

baschny:
Other than that, trying out the new solution is much more limiting, since we usually have this in our main composer.json:

```
    "repositories": [
        {
            "type": "path",
            "url": "packages/*",
```

Thus I have a drop-in place where to put "local packages" each with its own `composer.json` where I could previously state the `branch-version` in order to require them properly. The new functionality does not work with such a wildcard pattern.

Anyway, I will switch to using "version" in the local package files, since this seem to work for our use cases too. But just to let you know how I feel about this kind of change.

Seldaek:
@baschny it was introduced in https://github.com/composer/composer/commit/893fbfcb898f27a8ff70c2fd0939f1b58b83e4f8 3 months before this PR here, so it wasn't live for very long, and was not documented in the changelog either as it was kind of a weird hack for @nicolas-grekas to solve stuff in the symfony build. The mistake was definitely that this was in the docs so it looks like you found it in the meantime, sorry it caused trouble for you. Setting a version makes a lot more sense IMO in private code, it's more explicit and simpler. This branch-version was a hack to be able to set a version without really setting it, because if you do it causes problems when releasing/tagging things for public consumption.

We're generally very careful about BC but here given the short timeline and the fact that it's a niche feature yeah we cut corners to avoid having to live with this legacy thing for ever.

baschny:
@Seldaek thanks for the prompt feedback.

Indeed I just now realize that when we switched our project to use the `branch-version` a couple of months back this was a new feature. I thought it has been there forever (I even wondered how I didn't stumbled over it before).

Switching to a fixed "version" string in the local packages is what we are now doing.

I understand your reasonings and sorry for my harsh words on the matter, I was just frustrated. Nevertheless I appreciate your efforts on the whole composer universe and keep on the good work! 👍

PR #9658, opened 2021-01-29
Minor improvement to ClassLoader::register no-op block

Ayesh:
A recent change in the ClassAutoloader (#9635) added support for specifying a vendorDir, and `\Composer\Autoload\ClassLoader::register` now has a if-elseif-else chain.
The first block has a `// no-op` comment, followed by an `elseif` block. It's more readable to `return;`, and remove the `elseif` for readability.


nicolas-grekas (approved):

Ayesh:
Thank you @nicolas-grekas and @Seldaek 🙏🏼.

PR #9669, opened 2021-02-05
JsonFile: add missing ParsingException throws annotations

adlacruzes:
Hi,

while developing a composer plugin, I missed some throws annotations from `Composer\Json\JsonFile`

This PR adds `@throws ParsingException` to methods:
  - `parseJson`
  - `read`
  - `validateSchema`


adlacruzes:
Thank you, @Seldaek!

PR #9670, opened 2021-02-06
Add support for autoloading Enums

Ayesh:
PHP 8.1 supports Enums, and [Enums follow class-semantics](https://php.watch/versions/8.1/enums#class-semantics-autoload).

Composer's class-map generator currently looks for `class`, `interface`, and `trait` keywords. If Composer is run in PHP 8.1 or later, Composer now additionally looks for `enum` keyword as well. This is similar to how Hack's `enum` support is added.

This PR also adds tests for basic enums, backed enums, namespaced enums, and an enum that implements an interface and extends a class.

### DIFF
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -215,7 +215,7 @@ class ClassMapGenerator
     private static function findClasses($path)
     {
         $extraTypes = PHP_VERSION_ID < 50400 ? '' : '|trait';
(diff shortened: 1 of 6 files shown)


Ayesh:
Current vote count is 30:6 in favor. It's highly likely that this vote will pass, so I'd open the PR for review. There are some pending changes with error messages, but the autoloader is already functional.

Thank you.

Seldaek:
Looks good to me so far

Ayesh:
RFC vote ended with 44:7 in favor. PHP 8.1 will have Enums 🎉🎊.

Seldaek:
Thanks!

PR #9678, opened 2021-02-08
Add --format option to `composer fund` command

ochorocho:
Hey,

currently building a little GUI around composer. I want to show funding information in the GUI.
Therefore i added `--format` option to the `composer fund` command which allows to output
fund details as JSON.

### DIFF
--- a/src/Composer/Command/FundCommand.php
+++ b/src/Composer/Command/FundCommand.php
@@ -12,12 +12,14 @@
 
 namespace Composer\Command;
 
+use Composer\Json\JsonFile;
 use Composer\Package\CompletePackageInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\BasePackage;
 use Composer\Semver\Constraint\MatchAllConstraint;
 use Composer\Repository\CompositeRepository;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
@@ -30,6 +32,9 @@ class FundCommand extends BaseCommand
     {
         $this->setName('fund')
             ->setDescription('Discover how to help fund the maintenance of your dependencies.')
+            ->setDefinition(array(
+                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the output: text or json', 'text')
+            ))
         ;
     }
 
@@ -81,7 +86,14 @@ class FundCommand extends BaseCommand
 
         $io = $this->getIO();
 
-        if ($fundings) {
+        $format = $input->getOption('format');
+        if (!in_array($format, array('text', 'json'))) {
+            $io->writeError(sprintf('Unsupported format "%s". See help for supported formats.', $format));
+
+            return 1;
+        }
+
+        if ($fundings && $format === 'text') {
             $prev = null;
 
             $io->write('The following packages were found in your dependencies which publish funding information:');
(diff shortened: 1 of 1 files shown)


src/Composer/Command/FundCommand.php
```diff
@@ -104,6 +116,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $io->write("");
             $io->write("Please consider following these links and sponsoring the work of package authors!");
             $io->write("Thank you!");
+        } elseif ($fundings && $format === 'json') {
```
stof, line 119: this looks wrong to me. If there is no funding links found, the output should still be in JSON rather than plaintext when requesting json.

src/Composer/Command/FundCommand.php
```diff
@@ -104,6 +116,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            $fundingJson = array();
+            foreach ($fundings as $vendor => $links) {
+                $fundingJson[$vendor] = array();
+                foreach ($links as $url => $packages) {
+                    $fundingJson[$vendor]['packages'] = implode(', ', $packages);
+                    $fundingJson[$vendor]['url'] = $url;
```
stof, line 125: this output is wrong. It will keep only the last link of each vendor as it overrides it each time.

src/Composer/Command/FundCommand.php
```diff
@@ -104,6 +116,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $io->write("");
             $io->write("Please consider following these links and sponsoring the work of package authors!");
             $io->write("Thank you!");
+        } elseif ($fundings && $format === 'json') {
+            $fundingJson = array();
+            foreach ($fundings as $vendor => $links) {
+                $fundingJson[$vendor] = array();
+                foreach ($links as $url => $packages) {
+                    $fundingJson[$vendor]['packages'] = implode(', ', $packages);
```
stof, line 124: in JSON, a list of package as a JSON array is probably a lot better than a string containing a comma-separated list.

stof (changes requested):

src/Composer/Command/FundCommand.php
ochorocho: Makes sense. Now the packages is an array.

src/Composer/Command/FundCommand.php
ochorocho: Thanks, missed this one.

src/Composer/Command/FundCommand.php
ochorocho: True, fixed. :-)

ochorocho:
@stof thanks for reviewing this.

Seldaek (approved):

src/Composer/Command/FundCommand.php
```diff
@@ -104,6 +116,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $io->write("");
             $io->write("Please consider following these links and sponsoring the work of package authors!");
             $io->write("Thank you!");
+        } elseif ($format === 'json') {
+            $fundingJson = array();
+            foreach ($fundings as $vendor => $links) {
+                $fundingJson[$vendor] = array();
+                foreach ($links as $url => $packages) {
+                    $fundingJson[$vendor]['packages'] = $packages;
```
stof, line 124: this is still wrong. The list of packages is not per vendor but per url. Otherwise, you loose all except the ones of the last link.

stof (changes requested):

src/Composer/Command/FundCommand.php
ochorocho: @stof makes sense. Instead of building my own array the existing one is now used. So the issue should be fixed.
May i ask for another review? :-)

stof (approved):

Seldaek:
Thanks!

PR #9682, opened 2021-02-09
Allow querying InstalledVersions::isInstalled() for no-dev packages

nicolas-grekas:
Will help to fix issues like https://github.com/symfony/symfony/issues/40136

### DIFF
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -53,13 +53,14 @@ class InstalledVersions
      * This also returns true if the package name is provided or replaced by another package
      *
      * @param  string $packageName
+     * @param  bool   $includeDevRequirements
      * @return bool
      */
-    public static function isInstalled($packageName)
+    public static function isInstalled($packageName, $includeDevRequirements = true)
     {
         foreach (self::getInstalled() as $installed) {
             if (isset($installed['versions'][$packageName])) {
-                return true;
+                return $includeDevRequirements || empty($installed['versions'][$packageName]['dev-requirement']);
             }
         }
 
@@ -73,10 +74,9 @@ class InstalledVersions
      *
      *   Composer\InstalledVersions::satisfies(new VersionParser, 'foo/bar', '^2.3')
      *
-     * @param VersionParser $parser      Install composer/semver to have access to this class and functionality
-     * @param string        $packageName
-     * @param string|null   $constraint  A version constraint to check for, if you pass one you have to make sure composer/semver is required by your package
-     *
+     * @param  VersionParser $parser      Install composer/semver to have access to this class and functionality
+     * @param  string        $packageName
+     * @param  string|null   $constraint  A version constraint to check for, if you pass one you have to make sure composer/semver is required by your package
      * @return bool
      */
     public static function satisfies(VersionParser $parser, $packageName, $constraint)
--- a/src/Composer/Repository/FilesystemRepository.php
+++ b/src/Composer/Repository/FilesystemRepository.php
@@ -125,7 +125,7 @@ class FilesystemRepository extends WritableArrayRepository
 
 
         if ($this->dumpVersions) {
-            $versions = $this->generateInstalledVersions($installationManager);
+            $versions = $this->generateInstalledVersions($installationManager, $devMode);
 
             $fs->filePutContentsIfModified($repoDir.'/installed.php', '<?php return '.var_export($versions, true).';'."\n");
             $installedVersionsClass = file_get_contents(__DIR__.'/../InstalledVersions.php');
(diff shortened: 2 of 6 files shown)


tests/Composer/Test/InstalledVersionsTest.php
```diff
@@ -49,9 +49,10 @@ public function testGetInstalledPackages()
     /**
      * @dataProvider isInstalledProvider
      */
-    public function testIsInstalled($expected, $name, $constraint = null)
+    public function testIsInstalled($expected, $name, $noDev = false)
     {
-        $this->assertSame($expected, InstalledVersions::isInstalled($name));
+        // @phpstan-ignore-next-line
```
nicolas-grekas, line 54: Sames from `Composer\InstalledVersions::isInstalled() invoked with 2 parameters, 1 required`

Seldaek: That'll fix itself once the code is merged tho, so I'd rather have useless warnings in the PR than have to remove ignore stuff later :)

tests/bootstrap.php
```diff
@@ -17,4 +17,5 @@
 }
 
 require __DIR__.'/../src/bootstrap.php';
+require __DIR__.'/../src/Composer/InstalledVersions.php';
```
Seldaek, line 20: Why this change?

stof: I think this is to enforce that tests run against the code of the source rather than against the code generated in `vendor/` by the phar used to install deps, to be sure that tests use the uptodate code.

nicolas-grekas: @stof is right :)

tests/Composer/Test/InstalledVersionsTest.php
nicolas-grekas: removed

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -151,6 +152,7 @@ public function write($devMode, InstallationManager $installationManager)
                     'version' => $package->getVersion(),
                     'aliases' => array(),
                     'reference' => $reference,
+                    'dev' => isset($devPackages[$package->getName()]),
```
naderman, line 155: Wonder if this should be 'require-dev' or something that makes it clear this isn't about a dev version, e.g. 2.3.x-dev but rather about whether the package was installed because of require-dev?

src/Composer/InstalledVersions.php
```diff
@@ -53,13 +53,14 @@ public static function getInstalledPackages()
      * This also returns true if the package name is provided or replaced by another package
      *
      * @param  string $packageName
+     * @param  bool   $noDev
```
naderman, line 56: Similarly to my other comment, I think a better name would make it clear that this is about dependencies resulting from requireDev rather than dependencies which happen to have a dev version like 2.3.x-dev

src/Composer/Repository/FilesystemRepository.php
nicolas-grekas: The code uses the term "dev-mode", so I adopted it. WDYTN

src/Composer/InstalledVersions.php
nicolas-grekas: I propose `$noDevMode` then :)

naderman: Works for me

src/Composer/Repository/FilesystemRepository.php
Seldaek: Sorry for being late to the party but I wonder.. Wouldn't `$noDevRequirements`/`dev-requirement` be more self-explanatory? require-dev => dev requirements.

naderman: Yeah I think that works even better, as "dev-mode" may be misunderstood as simply whether composer was run with --dev or not. Yay bikeshedding, but let's go with @Seldaek's suggestion.

nicolas-grekas: Here you are updated.

nicolas-grekas:
(rebased)

src/Composer/InstalledVersions.php
```diff
@@ -53,13 +53,14 @@ public static function getInstalledPackages()
      * This also returns true if the package name is provided or replaced by another package
      *
      * @param  string $packageName
+     * @param  bool   $noDevRequirements
      * @return bool
      */
-    public static function isInstalled($packageName)
+    public static function isInstalled($packageName, $noDevRequirements = false)
```
Seldaek, line 59: Can we instead have a `$includeDevRequirements = true` to avoid the double negation?

nicolas-grekas: sure thing, updated.

Seldaek:
Thanks, merging now that master targets 2.1, but be aware this won't be in a release yet for a while :)

Issue #9694, opened 2021-02-11
It's possible to inject malicious packages on non-existant repository URLs


codekandis:
According to this article
https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610
and this german short about it
https://www.heise.de/news/Sicherheitsforscher-bricht-ueber-Open-Source-Repositories-bei-PayPal-Co-ein-5051635.html
I did some research and testing.

Assume I have a non-packagist repository in use called "vendor/non-packagist-repo".

```json
{
	"name": "vendor/some-application",
	"minimum-stability": "stable",
	"prefer-stable": true,
	"require": {
		"vendor/non-packagist-repo": "^0"
	},
	"repositories": {
		"vendor/non-packagist-repo": {
			"type": "vcs",
			"url": "file:///some/path/to/the/repository/vendor/non-packagist-repo.git"
		}
	}
}
```

Even if a repository with the same name exists on packagist composer resolves the given repository URL and pulls that non-packagist repository.

Now assume **there is** a repository on packagist which contains malicious code. Furtheron assume one has a typo in the repository URL which leads to a non-existing repository

```json
{
	"name": "vendor/some-application",
	"minimum-stability": "stable",
	"prefer-stable": true,
	"require": {
		"vendor/non-packagist-repo": "^0"
	},
	"repositories": {
		"vendor/non-packagist-repo": {
			"type": "vcs",
			"url": "file:///some/path/to/the/repository/vendor/non-existing-repo.git"
		}
	}
}
```

composer cannot resolve the repository URL but doesn't reject installing the package. Instead it pulls the malicious packagist repository without any hint or warning.

## I expect this to happen

As soon as I explicitely reference a repository by its name in the `repository` section

```json
"repositories": {
	"vendor/non-packagist-repo": {...}
}
```

I expect composer to validate the referenced repository data and to reject the repository pull. The explicit reference in fact means `I want you to pull this one! Not another one! This one!"

If no explicit reference is given

```json
"repositories": [
	{
		"type": "vcs",
		"url": "file:///some/path/to/the/repository/vendor/non-packagist-repo.git"
	}
]
```

composer is free to lookup on further resources. Because `I want you to look up first at this URL. But you're free to use another source.`

stof:
`vendor/non-packagist-repo` is not a package name in your composer.json `repositories`. It is a repository name. Repositories can contain many packages (for instance the default repository `packagist.org` contains thousands of packages).

Seldaek:
[...]

PR #9695, opened 2021-02-11
Fix processes silently ignoring the CWD when it does not exist

Seldaek:
refs #9694


stof (approved):

Seldaek:
Looks like Symfony 4.0 fixed this in Process. Can't wait for Composer 2.2 :/

PR #9696, opened 2021-02-12
Fix installed versions usage issues when using it in plugins during a Composer update process

Seldaek:


nicolas-grekas (approved):

naderman:
Can you add a test for how this works if you have 2 plugins being updated which are checking the version installed of each other?

I suppose this improves the situation, but the installed package versions in a plugin now always report the final state after installation. If you're in the first plugin being updated you'll see the 2nd plugin as already updated, although it's still loaded in the old version? So if the goal is to be able to rely on installed package info during the plugin invocation of an update, you now have a different but still incorrect state?

Seldaek:
IMO you can't really have two plugins requiring each other. I mean A requiring B sure, then B will update first and will be updated once A updates/activates again. But if it's a circular dep I don't see how we can possibly deal with this.

> I suppose this improves the situation, but the installed package versions in a plugin now always report the final state after installation.

Not really, it always reports the current state after each package's install/update/.. It does not report the final state. See https://github.com/composer/composer/pull/9696/files#diff-e2dde1ff6d8789d0db27dd803004c892694d9dd70a399a9789129def7be82711R6 and all the other !!Versions lines. At first it reports bundled versions from Composer itself, but then as newer versions of the symfony packages get installed they one by one switch to their newer version in InstalledVersions.

So I'm not sure what a new test here would look like.

naderman:
Thanks for the test, that looks like what I was thinking of!

Issue #9648, opened 2021-01-26
Improvements to InstalledVersions


Seldaek:
As per https://github.com/composer/installers/issues/470#issuecomment-740691709 & following comments - we should add a way to retrieve packages by type, and also include a bit more information about each package than only name/version/ref, namely:

- type
- install path

We should also include the root package in all this.

ochorocho:
@Seldaek i'm working on this now. What do you mean by "We should also include the root package in all this."?
Do you want me to include the "Root package name" along with "type" and "install_path" ?

Seldaek:
@ochorocho I mean the root package should also have an install path and all, I don't think it's anything special really, just a reminder to check when reviewing, but I think you can probably ignore this and it'll just work.

aimeos:
@Seldaek When is this feature expected to occur in composer? Is there a planned time frame?

Seldaek:
Hoping to merge @ochorocho 's PR next week, but it'll be shipping in Composer 2.1 for sure, when that ships though I am not sure.

PR #9699, opened 2021-02-15
Add install-path and type to installedVersions.php and installed.php,…

ochorocho:
… add method to get installed packages by type

* Extended installedVersions.php and installed.php with `type` and `install_path`
* Added `getInstalledPackagesByType` to allow retrieval of of installed packages by a given type

@Seldaek your feedbeack is very welcome :-)

https://github.com/composer/composer/issues/9648

### DIFF
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -24,9 +24,14 @@ use Composer\Semver\VersionParser;
  */
 class InstalledVersions
 {
-    private static $installed;
+    private static $installed = array();
     private static $canGetVendors;
     private static $installedByVendor = array();
+
+    /**
+     * Initialize $installed array
+     */
+    public static function initializeInstalled() {}
 
     /**
      * Returns a list of all package names which are present, either by being installed, replaced or provided
@@ -46,6 +51,28 @@ class InstalledVersions
         }
 
         return array_keys(array_flip(\call_user_func_array('array_merge', $packages)));
+    }
+
+    /**
+     * Returns a list of all package names with a specific type e.g. 'library'
+     *
+     * @param  string   $type
+     * @return string[]
+     * @psalm-return list<string>
+     */
+    public static function getInstalledPackagesByType($type)
+    {
+        $packagesByType = array();
+
+        foreach (self::getInstalled() as $installed) {
+            foreach ($installed['versions'] as $name => $package) {
+                if (isset($package['type']) && $package['type'] === $type) {
+                    $packagesByType[] = $name;
+                }
+            }
+        }
+
+        return $packagesByType;
     }
 
     /**
@@ -274,3 +301,5 @@ class InstalledVersions
         return $installed;
     }
 }
+
+InstalledVersions::initializeInstalled();
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -88,7 +88,7 @@ class InstalledVersions
     {
         foreach (self::getInstalled() as $installed) {
             if (isset($installed['versions'][$packageName])) {
-                return $includeDevRequirements || empty($installed['versions'][$packageName]['dev-requirement']);
+                return $includeDevRequirements || empty($installed['versions'][$packageName]['dev_requirement']);
             }
         }
 
@@ -230,7 +230,7 @@ class InstalledVersions
      *
      * @deprecated Use getAllRawData() instead which returns all datasets for all autoloaders present in the process. getRawData only returns the first dataset loaded, which may not be what you expect.
      * @return array[]
-     * @psalm-return array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool}, versions: array<string, array{dev-requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[]}>}
+     * @psalm-return array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool}, versions: array<string, array{dev_requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[]}>}
      */
     public static function getRawData()
     {
@@ -266,7 +266,7 @@ class InstalledVersions
      * @param  array[] $data A vendor/composer/installed.php data set
      * @return void
      *
-     * @psalm-param array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool}, versions: array<string, array{dev-requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[]}>} $data
+     * @psalm-param array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool}, versions: array<string, array{dev_requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[]}>} $data
      */
     public static function reload($data)
     {
@@ -276,7 +276,7 @@ class InstalledVersions
 
     /**
      * @return array[]
-     * @psalm-return list<array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool}, versions: array<string, array{dev-requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[]}>}>
+     * @psalm-return list<array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool}, versions: array<string, array{dev_requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[]}>}>
      */
     private static function getInstalled()
     {
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -259,7 +259,7 @@ class InstalledVersions
      * Returns the raw data of all installed.php which are currently loaded for custom implementations
      *
      * @return array[]
-     * @psalm-return list<array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], install_path: string}, versions: array<string, array{dev_requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[], install_path?: string}>}>
+     * @psalm-return list<array{root: array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool, install_path: string}, versions: array<string, array{dev_requirement: bool, pretty_version?: string, version?: string, aliases?: string[], reference?: string, replaced?: string[], provided?: string[], install_path?: string}>}>
      */
     public static function getAllRawData()
     {
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -215,8 +215,25 @@ class InstalledVersions
     }
 
     /**
+     * @param  string      $packageName
+     * @return string|null If the package is being replaced or provided but is not really installed, null will be returned as install path. Packages of type metapackages also have a null install path.
+     */
+    public static function getInstallPath($packageName)
+    {
+        foreach (self::getInstalled() as $installed) {
+            if (!isset($installed['versions'][$packageName])) {
+                continue;
+            }
+
+            return isset($installed['versions'][$packageName]['install_path']) ? $installed['versions'][$packageName]['install_path'] : null;
+        }
+
+        throw new \OutOfBoundsException('Package "' . $packageName . '" is not installed');
+    }
+
+    /**
      * @return array
-     * @psalm-return array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool}
+     * @psalm-return array{name: string, version: string, reference: string, pretty_version: string, aliases: string[], dev: bool, install_path: string}
      */
     public static function getRootPackage()
     {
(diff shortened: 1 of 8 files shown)


ochorocho:
Just had a look whats going on with phpstan issues.
Get 352 errors, even when running it on `master`

```
 [ERROR] Found 352 errors     
 ```

File: src/Composer/Repository/FilesystemRepository.php [hunk 1]
```diff
@@ -146,9 +146,15 @@ public function write($devMode, InstallationManager $installationManager)
                     $reference = ($package->getSourceReference() ?: $package->getDistReference()) ?: null;
                 }
 
+                $basePath = $fs->normalizePath(realpath(realpath(getcwd())));
+                $path = $package instanceof RootPackageInterface ? $basePath : $installationManager->getInstallPath($package);
+                $installPath = ('' !== $path && null !== $path) ? $fs->findShortestPath($repoDir, $fs->isAbsolutePath($path) ? $path : getcwd() . '/' . $path, true) : null;
```
stof, line 151: returning a relative path to the `installed.json` file looks bad to me: there is no way to actually resolve that to a usable path when actually using the InstalledVersions API as the path to the installed.json file is not exposed anywhere (and when you merged data from multiple vendor folders, the relative path will not always be relative to the same thing)

stof (changes requested):

src/Composer/Repository/FilesystemRepository.php
ochorocho: Thanks for reviewing :-) Now using absolute paths.

ochorocho:
@stof can't wrap my head around this phpstan issue. It says `Call to an undefined static method`. The method is there, and it works locally and in "most" Actions except [windows](https://github.com/composer/composer/pull/9699/checks?check_run_id=1912251047)
Any idea what could cause this?

grafik

https://github.com/composer/composer/pull/9699/checks?check_run_id=1912250647

Seldaek:
@ochorocho sorry for the trouble but could you rebase this on latest master? I merged other things which conflict with your PR.

As for the phpstan error, you can ignore these, it's a quirk of the fact the InstalledVersions file used in CI does not yet contain the new methods until the PR gets merged.

ochorocho:
@Seldaek rebase done :-)

ochorocho:
FYI, all packages defined with `self.version` are missing paths.
Working on a fix right now.

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -171,9 +171,14 @@ private function generateInstalledVersions(InstallationManager $installationMana
...
             }
 
+            $path = $installationManager->getInstallPath($package);
+            $installPath = realpath($path);
+
             $versions['versions'][$package->getName()] = array(
                 'pretty_version' => $package->getPrettyVersion(),
                 'version' => $package->getVersion(),
+                'type' => $package->getType(),
+                'install_path' => $installPath,
```
mvorisek, line 181: path should be relative to root project (or vendor?) directory to allow moving the project to somewhere else (realpath should be therefore NOT used as well)

Seldaek: getInstallPath is usually an absolute path though, so the right way here is to do like it's doing at line 109 to initialize `$pkgArray['install-path']`. Possibly the paths could be kept there in an array per package and passed to `generateInstalledVersions` too so you don't have to redo it here.

Seldaek: Also if `getInstallPath` returned `''` (like it does for packages of type `metapackage`) maybe we should set `install_path` to `null` to make this clear that there is no path/nothing installed. For the root package it should however be set to `''` representing getcwd().

src/Composer/InstalledVersions.php
```diff
@@ -39,14 +39,37 @@ public static function getInstalledPackages()
...
     }
 
+    /**
+     * Returns a list of all package names with a specific type e.g. 'library'
+     *
+     * @param  string   $type
+     * @return string[]
+     * @psalm-return list<string>
+     */
+    public static function getInstalledPackagesByType($type = 'library')
```
Seldaek, line 56: I don't think I'd have a default value here. If you don't know what you are looking for you shouldn't be calling this :)

src/Composer/InstalledVersions.php
```diff
@@ -39,14 +39,37 @@ public static function getInstalledPackages()
...
+        foreach (self::getInstalled() as $installed) {
+            $packages[] = $installed['versions'];
+        }
+
+        $packagesByType = array();
+        foreach ($packages[0] as $key => $package) {
+            if (isset($package['type']) && $package['type'] == $type) {
+                $packagesByType[] = $key;
+            }
+        }
```
Seldaek, line 68: You are ignoring everything but the first array returned by getInstalled here, which is not good in case there are multiple autoloaders loaded.

```suggestion
        $packagesByType = array();

        foreach (self::getInstalled() as $installed) {
            foreach ($installed['versions'] as $name => $package) {
                if (isset($package['type']) && $package['type'] === $type) {
                    $packagesByType[] = $name;
                }
            }
        }
```

Seldaek:
> FYI, all packages defined with self.version are missing paths.

I am not sure what you mean, if you have more details maybe we can figure out a solution or at least identify the root cause.

src/Composer/InstalledVersions.php
ochorocho: Makes sense :-) Removed the default value.

ochorocho:
> > FYI, all packages defined with self.version are missing paths.
> 
> I am not sure what you mean, if you have more details maybe we can figure out a solution or at least identify the root cause.

`install_path` was missing for some packages in `vendor/composer/InstalledVersions.php`.
Checked it again after i applied your suggestion and it seems to work now.

Thank you.

Let me know if you need anything else. :-)

Seldaek:
Thanks for the fixes. The last thing open is https://github.com/composer/composer/pull/9699#pullrequestreview-613122433 which definitely needs fixing but then it looks good to me.

src/Composer/Repository/FilesystemRepository.php
stof: note that not only metapackages will have no install path. Replaced packages will also be in such case (as they are not installed anywhere)

ochorocho: @stof thanks for that insight. So i reckon the problem described [here](https://github.com/composer/composer/pull/9699#issuecomment-800311716) still exists. I will verify that and let you know.

tests/Composer/Test/InstalledVersionsTest.php
```diff
@@ -187,6 +187,8 @@ public function testGetRootPackage()
         $this->assertSame(array(
             'pretty_version' => 'dev-master',
             'version' => 'dev-master',
+            'type' => 'library',
+            'install_path' => false,
```
stof, line 191: `string|null` is a better type than `string|false` IMO

ochorocho: Changed to `null`

ochorocho:
> Thanks for the fixes. The last thing open is [#9699 (review)](https://github.com/composer/composer/pull/9699#pullrequestreview-613122433) which definitely needs fixing but then it looks good to me.

@Seldaek now using an array containing package name and its path and passing it to `generateInstalledVersions`.
The root package path is now `'install_path' => '',` and `metapackage` using `'install_path' => NULL,`

tests/Composer/Test/Repository/Fixtures/installed.php
```diff
@@ -14,6 +14,8 @@
     'root' => array(
         'pretty_version' => 'dev-master',
         'version' => 'dev-master',
+        'type' => 'library',
+        'install_path' => false,
```
stof, line 18: this is wrong

src/Composer/Repository/FilesystemRepository.php
stof: the `InstalledVersions` API should still work with absolute paths. Otherwise, it is unusable as you won't be able to resolve them.

To make things portable, the solution is to dump paths using `__DIR__`, as done in the generated autoload files.

stof: which would require dumping that in a PHP file rather than JSON.
The other option would be to resolve relative paths in the place reading the json file (and so knowing its location)

Seldaek: Hmm yes `__DIR__` indeed would be good. `vendor/composer/installed.php` is indeed a PHP file not JSON so no problem there. It'll definitely make this PR more complex though as var_export can't be used as is anymore, sorry @ochorocho :)

Seldaek: Yup that's the idea. You can check AutoloadGenerator to see examples of doing this.

ochorocho: @Seldaek created a method to dump the version array including `__DIR__`.
Ready for review :-)

tests/Composer/Test/Repository/Fixtures/installed.php
ochorocho: Changed to `'install_path' => '',`.

Thanks you!

src/Composer/InstalledVersions.php
```diff
@@ -22,10 +22,15 @@
...
 {
-    private static $installed;
+    private static $installed = array();
     private static $canGetVendors;
     private static $installedByVendor = array();
 
+    /**
+     * Initialize $installed array
+     */
+    public static function __constructStatic() {}
```
stof, line 32: why `__constructStatic` ? names prefixed by `__` are soft-reserved by PHP for magic methods, and this is not one of them.

src/Composer/InstalledVersions.php
```diff
@@ -258,3 +284,5 @@ private static function getInstalled()
         return $installed;
     }
 }
+
+InstalledVersions::__constructStatic();
```
stof, line 288: why ? this is a no-op

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -125,24 +130,53 @@ public function write($devMode, InstallationManager $installationManager)
...
+                } else {
+                    $lines .= "array(),\n";
+                }
+            }  elseif (is_bool($value)) {
+                $lines .= $value ? 'true' : 'false';
+                $lines .= ",\n";
+            } else {
+                $stringContent = str_replace(array('\\', '\''), array('\\\\', '\\\''), $value);
+                $folder = $key === 'install_path' ? '__DIR__ . DIRECTORY_SEPARATOR . ' : '';
+                $lines .= $folder . '\'' . $stringContent . "',\n";
```
stof, line 168: you should still be using `var_export` to export the string literal, rather than rebuilding it in userland (in an incomplete way)

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -171,9 +205,13 @@ private function generateInstalledVersions(InstallationManager $installationMana
                 $reference = ($package->getSourceReference() ?: $package->getDistReference()) ?: null;
             }
 
+            $installPath = $package instanceof RootPackageInterface ? '' : $installPaths[$package->getName()];
```
stof, line 208: this is wrong. the relative path from the location of the installed.php file to the root package is not the empty string.

stof (commented):
I don't see any tests covering the generation of the `install_path` key

src/Composer/Repository/FilesystemRepository.php
ochorocho: This is the final result in `InstalledVersions.php` so i reckon the path correct:
```php
self::$installed = array(
    'root' => array(
        'pretty_version' => 'dev-master',
        'version' => 'dev-master',
        'type' => 'library',
        'install_path' => __DIR__ . DIRECTORY_SEPARATOR . '',
        'aliases' => array(
            0 => '2.1.x-dev',
        ),
        'reference' => 'a614587ec91031b99ab23b385838f6d9641c847d',
        'dev-requirement' => true,
        'name' => 'composer/composer',
    ),
   ....
```

Seldaek: root should have an install_path of `__DIR__ . '/../../'` for composer, but the result should be dependent on the vendor-dir, so it should really compute the relative path from vendor-dir/composer to getcwd() to find the root install path.

src/Composer/InstalledVersions.php
ochorocho: Makes sense, will rename it.

src/Composer/Repository/FilesystemRepository.php
ochorocho: Well, i need this to be able to write `__DIR__ . DIRECTORY_SEPARATOR . ` to the PHP file and not the resulting path.
Happy to have any suggestions to avoid using my own method.

stof: Your generated line is `__DIR__ . DIRECTORY_SEPARATOR .` followed by a string literal. I'm talking only about the string literal, not about the whole line.

src/Composer/Repository/FilesystemRepository.php
```diff
@@ -205,7 +205,16 @@ private function generateInstalledVersions(InstallationManager $installationMana
                 $reference = ($package->getSourceReference() ?: $package->getDistReference()) ?: null;
             }
 
-            $installPath = $package instanceof RootPackageInterface ? '' : $installPaths[$package->getName()];
+            if($package instanceof RootPackageInterface) {
+                $fs = new Filesystem();
+                $config = $package->getConfig();
+                $vendorDir = !empty($config['vendor-dir']) ? $config['vendor-dir'] : 'vendor';
+                $to = getcwd();
+                $from = $to . DIRECTORY_SEPARATOR . $vendorDir . DIRECTORY_SEPARATOR . 'composer';
```
stof, line 213: no need to do your own processing of the vendor-dir config. The `$from` is the `$repoDir` that we have in `->write()`, i.e. the folder in which we are generating the file. You only need to pass it as an argument in this private method

ochorocho: Thanks for that hint. Changed it.

ochorocho:
@Seldaek tried to fix `testRepositoryWritesInstalledPhp` but failed. :-( 
The path is confusing me `/foo/bar/vendor/woop/woop`.

Can you give me any advice to get this test running again?

ochorocho:
@Seldaek all tests passed now. Branch is up to date with master.

Seldaek:
@ochorocho thanks for this, I rebased it again and squashed commits to clean up a little then reviewed in depth and did some changes (some of it are kinda unrelated to this PR, just other things I noticed needed tweaking before 2.1 for InstalledVersions while I was poking around). Have a look at my individual commits if you're interested in more details.

I think this is mergeable now, but happy if someone else has a second to review again.

ochorocho (commented):
Looks good to me.

Thank you! :-)

PR #9700, opened 2021-02-16
Allow $schema field in composer.json

RedGuy12:
When I add 
```json
"$schema":"https://getcomposer.org/schema.json"
```
to my composer file, it says
```
WARNING: Property $schema is not allowed.
```
This commit fixes that.

haha I got #9700

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -7,7 +7,7 @@
     "properties": {
         "$schema": {
 			"type": "string",
-			"description": "The URL to the schema."
+			"description": "The URL to the schema.",
             "enum": [
                 "https://getcomposer.org/schema.json"
             ]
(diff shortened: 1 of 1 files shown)


stof:
Why adding such property in the composer.json file ?

RedGuy12:
> Why adding such property in the composer.json file ?

So VS code will autocomplete fields for me, alerts me when I made a mistake, etc.

stof:
Requiring a `$schema` property to know that a JSON file correspond to a given schema is **not** the way JSON schema is designed (otherwise, it would have automatically allowed this field). I think this is an issue in VS Code instead.

RedGuy12:
microsoft/vscode#116793

PR #9702, opened 2021-02-16
Check if SODIUM_LIBRARY_VERSION is defined

brandonkelly:
PackageRepository assumes that if the `libsodium`/`sodium` extension is installed, that the `SODIUM_LIBRARY_VERSION` constant will be defined, however that constant wasn’t added until v2 of the Sodium extension.

This PR checks to make sure the constant is defined before attempting to reference it.


Seldaek:
Thanks

Issue #9703, opened 2021-02-16
EventDispatcher uses only the return value of the last one listener


Sweetchuck:
My `composer.json`: v2.0.9

When there are two or more Composer plugin which are subscribe to the `PluginEvents::COMMAND` event, then only the return value of the last one listener will be take into account to determine the exit code.

```php
<?php
class Plugin implements PluginInterface, EventSubscriberInterface, Capable
{
    public static function getSubscribedEvents()
    {
        return [
            PluginEvents::COMMAND => 'onCommandEvent',
        ];
    }

    public function onCommandEvent(CommandEvent $event): bool
    {
        switch ($event->getCommandName()) {
            case 'validate':
                // For testing purposes always returns FALSE.
                return false;
        }

        return true;
    }
}
```

And an other plugin which always returns TRUE for the same event, and this one is the last one listener. And you ran the `composer validate` command, then the exit code is always `0`.

In the `\Composer\EventDispatcher\EventDispatcher::doDispatch` the boolean return value of a listener is converted to int, which represent the exit code.

Excepted behaviour:
If any of the listeners returns FALSE, then the exit code should be 1.

PR #9704, opened 2021-02-16
Issue #9703 - EventDispatcher uses only the return value of the last one listener

Sweetchuck:
Issue #9703 - EventDispatcher uses only the return value of the last one listener


Seldaek:
Thanks, the fix definitely makes sense to me. It'd be great if you could add a test for it though as that's something that would otherwise easily be missed/broken in the future.

PR #9713, opened 2021-02-23
Generate binary proxy in PHP if the target binary is detected as a PHP script

### LINKED ISSUE #9325: Windows, vendor binaries and the proxy script
Currently Composer handles installation of binaries in a special way when run in a Windows environment:

* A `.bat` file is generated automatically to reference the binary
* _A Unix-style proxy file with the same name as the binary is generated automatically (useful for Cygwin or Git Bash)_

https://getcomposer.org/doc/articles/vendor-binaries.md#what-about-windows-and-bat-files-

I am wondering why Composer creates the proxy file at all, instead of creating a relative symlink to the actual bin file, just like on Linux. There does not seem to be any benefit to having the proxy file. Instead it prevents you from using a specific PHP CLI:

* When using the Windows CMD, executing `vendor\bin\the-bin-file` will automatically use the `vendor\bin\the-bin-file.bat` file.
* When using the Git Bash etc., executing `vendor/bin/the-bin-file` will execute said proxy shell script.
* **However**, when using _either_ the Windows CMD _or_ Git Bash etc. executing `php vendor/bin/the-bin-file` or `php vendor\bin\the-bin-file` respectively will not work in either case, because PHP of course cannot interpret a shell script and just outputs its content.

If instead I manually create the relative symlink on Windows to the PHP binary (example with the Contao Managed Edition binary):

everything works as expected, both in the Windows CMD:

as well as in the Git Bash for example:

So - what is the purpose of the proxy shell script? The reason I am wondering about this at all: for Contao we want to switch from using a Composer script handler to directly using a command: https://github.com/contao/contao/pull/2183

However, this currently does not work on Windows, because of the proxy shell script.

Seldaek:
This allows custom PHP processes to be used to call them via the bin proxy path, fixes #9325

e.g. `php7.4 bin/foo` would now work, whereas currently it yields:

```
$ php7.4 vendor/bin/jsonlint

dir=$(cd "${0%[/\\]*}" > /dev/null; cd '../seld/jsonlint/bin' && pwd)

if [ -d /proc/cygdrive ]; then
    case $(which php) in
        $(readlink -n /proc/cygdrive)/*)
            # We are in Cygwin using Windows php, so the path must be translated
            dir=$(cygpath -m "$dir");
            ;;
    esac
fi

"${dir}/jsonlint" "$@"
```

The downside of this is it uses eval and is doing nasty hackery to remove the shebang from the original script, to avoid having that in the output.

On the other hand, this hackery is only used on Windows as on Linux/OSX we simply symlink the original bin file.

Here is a copy of the rendered binary proxy to make it easier to review than the heredoc nonsense:

```php
$ cat vendor/bin/jsonlint
#!/usr/bin/env php
<?php

/**
 * Proxy PHP file generated by Composer
 *
 * This file includes the referenced bin path (../seld/jsonlint/bin/jsonlint) using eval to remove the shebang if present
 *
 * @generated
 */

$binPath = realpath(__DIR__ . "/" . '../seld/jsonlint/bin/jsonlint');
$contents = file_get_contents($binPath);
$contents = preg_replace('{^#!/.+\r?\n<\?(php)?}', '', $contents, 1, $replaced);
if ($replaced) {
    $contents = strtr($contents, array(
        '__FILE__' => var_export($binPath, true),
        '__DIR__' => var_export(dirname($binPath), true),
    ));

    eval($contents);
    exit(0);
}
include $binPath;
```


src/Composer/Installer/BinaryInstaller.php
```diff
@@ -191,6 +191,42 @@ protected function generateUnixyProxyCode($bin, $link)
...
+ * Proxy PHP file generated by Composer
+ *
+ * This file includes the referenced bin path ($binPath) using eval to remove the shebang if present
+ *
+ * @generated
+ */
+
+\$binPath = realpath(__DIR__ . "/" . $binPathExported);
+\$contents = file_get_contents(\$binPath);
+\$contents = preg_replace('{^#!/.+\\r?\\n<\\?(php)?}', '', \$contents, 1, \$replaced);
```
naderman, line 215: I don't think this should be replacing `.+` here which potentiall swallows additional unrelated lines. Maybe `[^\\r\\n]+`

src/Composer/Installer/BinaryInstaller.php
```diff
@@ -191,6 +191,42 @@ protected function generateUnixyProxyCode($bin, $link)
         $binDir = ProcessExecutor::escape(dirname($binPath));
         $binFile = basename($binPath);
 
+        $binContents = file_get_contents($bin);
+        if (preg_match('{^(?:#!(?:/usr)?/bin/env php|#!(?:/usr)?/bin/php|<?php)\r?\n}', $binContents, $match)) {
+            $proxyCode = trim($match[0]);
+            // carry over the existing shebang if present, otherwise add our own
```
stof, line 197: if there was no original shebang, do we actually need to generate code checking it again at runtime ?

Seldaek: I guess not really, but that's most likely a very rare case anyway so not sure it's worth the added complexity

src/Composer/Installer/BinaryInstaller.php
Seldaek: . does not include newlines so it just swallows anything between `#!` and the first newline which sounds fine?

naderman: Heh indeed, always forget that's what /s is for, sorry :-)

But then I'm unsure why it also removes an opening PHP tag but only if there is a shebang?

Seldaek: Because if there is no shebang we can simply include the file and skip the whole eval nonsense. So if no shebang no replacement at all is done and $replaced is 0.

naderman (approved):

oleg-andreyev:
Looks like this PR is causing quite interesting side-effect.
We have a large project where we gradually add psalm, so we have root-level psalm.xml and some modules have their own psalm.xml.

When running psalm on a module, it's loading psalm.xml from a module and changing script root (chroot), trying to find composer.lock and since it's missing doing re-scan and now psalm is able to find **new** php-file ./vendor/bin/psalm before it was a bash wrapper.

And reports
image

Seldaek:
So you're saying the issue is psalm is analysing the vendor/bin/x files, reporting issues? This honestly seems like an issue in your psalm config, vendor probably should be excluded, or at least vendor/bin?

oleg-andreyev:
@Seldaek looks like a combo configs and how psalm detects the current file.

Seldaek:
The question is also how you execute psalm.. If you execute it via Composer, Composer will add the bin-dir to PATH, and so psalm may find a psalm binary in there.. I am not sure exactly what error you're seeing, nor how psalm does bootstrapping, so I don't know if my understanding is correct, but maybe this helps you figure out the root cause.

PR #9714, opened 2021-02-23
Auto-detect dev-mode in autoload-dump and deprecate dump-autoload --no-dev

Seldaek:
Triggered by https://github.com/composer/package-versions-deprecated/issues/23 - I realized this now can be automated. IMO there really should be no reason to dump an autoloader without dev packages, if you installed dev packages. This causes inconsistencies like InstalledVersions having packages which are then not autoloadable, so should be discouraged.

### DIFF
--- a/src/Composer/Command/DumpAutoloadCommand.php
+++ b/src/Composer/Command/DumpAutoloadCommand.php
@@ -35,7 +35,7 @@ class DumpAutoloadCommand extends BaseCommand
                 new InputOption('classmap-authoritative', 'a', InputOption::VALUE_NONE, 'Autoload classes from the classmap only. Implicitly enables `--optimize`.'),
                 new InputOption('apcu', null, InputOption::VALUE_NONE, 'Use APCu to cache found/not-found classes.'),
                 new InputOption('apcu-prefix', null, InputOption::VALUE_REQUIRED, 'Use a custom prefix for the APCu autoloader cache. Implicitly enables --apcu'),
-                new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables autoload-dev rules.'),
+                new InputOption('no-dev', null, InputOption::VALUE_NONE, 'DEPRECATED: Disables autoload-dev rules. Composer now infers this automatically according to the last install or update run.'),
                 new InputOption('ignore-platform-req', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Ignore a specific platform requirement (php & ext- packages).'),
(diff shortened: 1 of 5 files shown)


src/Composer/Command/DumpAutoloadCommand.php
```diff
@@ -77,7 +77,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $ignorePlatformReqs = $input->getOption('ignore-platform-reqs') ?: ($input->getOption('ignore-platform-req') ?: false);
 
         $generator = $composer->getAutoloadGenerator();
-        $generator->setDevMode(!$input->getOption('no-dev'));
+        if ($input->getOption('no-dev')) {
+            $this->getIO()->writeError('<warning>You are using the deprecated option "--no-dev". Composer now guesses the value automatically according to the last install or update run.</warning>');
```
naderman, line 81: "guess" suggests it is sometimes wrong, that's not the case here? Think "infers" would be a better word?

naderman (changes requested):
I would tend to agree with you, but I have a feeling you'll run into some valid use case for --no-dev autoloading with dev deps installed XD

Seldaek:
I mean --no-dev still works so maybe we leave it as is, see if anyone complains about the deprecation.. we can always un-deprecate it :)

Ayesh:
I will bite the bullet :)

Going against the recommendations, I commit the vendor directory to VCS. Before each commit, I quickly run the test suit, run `composer dump-autoload --no-dev`, and then push changes to the server. This works for me well, because the CI server does not need to download anything, but simply dump the autoloader with `require-dev` packages. Helps me to keep the cost down, with each CI run taking under a minute to install. On the other hand, the production server does not even need Composer installed, because the repository is production-read just after `git pull`.

I'm totally supportive of keeping Composer as minimal as possible, and I totally understand and am grateful for the maintenance burden @naderman and @Seldaek put in Composer. However, the `no-dev` option helps me a lot to quickly get rid of the hundreds of `require-dev` items in the autoloader with almost no changes to vendor directory. Unless there is a technical reason, I would like to humbly propose to not deprecate this option. My workflow isn't what Composer project recommends, but I don't think I'm the only one who commits `vendor` directory to VCS. For those who do, the `composer dump-autoload --no-dev` is a quick hanfy command to prime the repository for production. 

Thank you.

Seldaek:
@Ayesh alright alright, https://github.com/composer/composer/commit/387e504437e422d6e369964a3eaa73ec25b0d9c1 undoes the deprecation :)

Ayesh:
Thank you so much for being very considerate and undeprecating this change @Seldaek.

PR #9729, opened 2021-02-25
Remove version argument from why and enforce it for why-not command

Seldaek:
Refs #9693

Strictly speaking a minor break but as it's really the way this was intended to be used and kinda just enforces the expectations instead of loosely allowing misuse, I think it's fine.


naderman:
I also strongly doubt anyone is using this in an automated fashion, so I think this should be fine.

naderman (approved):

PR #9730, opened 2021-02-25
Fix compiler on Windows

johnstevenson:
I guess I'm the only one who runs the compiler on Windows PHP.


stof (approved):

Seldaek:
Yeah since I upgraded to WSL2 I am hardly running stuff natively on Windows these days..

PR #9733, opened 2021-02-26
Reset file access on Windows admin self-update

johnstevenson:
Windows PHP `rename` uses MoveFileEx which preserves the source file
permissions when creating the destination file*. If the user is running
as an Admin and the destination file is in a common (non-user) location,
then the permission for other users will be replaced by the admin user.

This means that other users will no longer be able to run
composer.phar

Additionally, the UAC elevation feature uses the cmd.exe `move` command
which behaves as above, rather than the `copy` command which creates the
destination file with permissions inherited from the parent folder - but
only if inheritance is enabled, which is the default.

This fix runs `icacls` after admin operations to reset the inherited
permissions. This is equivalent to deleting the destination file first,
then creating it as new from the source file.

*PHP's use of MoveFileEx results in copy semantics when the source and
destination files are on different volumes.

## Testing
Download and unzip [composer-2.1-dev+711721a.zip](https://github.com/composer/composer/files/6050457/composer-2.1-dev%2B711721a.zip). Create a `test` folder in `C:\Program Files` and copy this composer.phar here

```cmd
# From an Administrator cmd prompt
mkdir "C:\Program Files\test"
copy /B <path to>\composer.phar "C:\Program Files\test\composer.phar"
```
To test:

```cmd
# From a User cmd prompt
cd "C:\Program Files\test"

# See the current permissions
icacls composer.phar

# The Users permissions will look like (in English):
BUILTIN\Users:(I)(RX)

# Run the update
php composer.phar self-update

# Check that the Users permissions are present and the same
icacls composer.phar
```

_Note:_ The User permissions are inherited (I) from the parent folder with read-execute (RX) access.

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -425,7 +425,14 @@ TAGSPUBKEY
         }
 
         try {
-            rename($newFilename, $localFilename);
+            if (Platform::isWindows()) {
+                // use copy to apply permissions from the destination directory
+                // as rename uses source permissions and may block other users
+                copy($newFilename, $localFilename);
+                @unlink($newFilename);
+            } else {
+                rename($newFilename, $localFilename);
+            }
 
             return true;
         } catch (\Exception $e) {
@@ -534,7 +541,7 @@ TAGSPUBKEY
     /**
      * Invokes a UAC prompt to update composer.phar as an admin
      *
-     * Uses a .vbs script to elevate and run the cmd.exe move command.
+     * Uses a .vbs script to elevate and run the cmd.exe copy command.
      *
      * @param  string $localFilename The composer.phar location
      * @param  string $newFilename   The downloaded or backup phar
@@ -560,13 +567,13 @@ TAGSPUBKEY
 
         $checksum = hash_file('sha256', $newFilename);
 
-        // cmd's internal move is fussy about backslashes
+        // cmd's internal copy is fussy about backslashes
         $source = str_replace('/', '\\', $newFilename);
         $destination = str_replace('/', '\\', $localFilename);
 
         $vbs = <<<EOT
 Set UAC = CreateObject("Shell.Application")
-UAC.ShellExecute "cmd.exe", "/c move /y ""$source"" ""$destination""", "", "runas", 0
+UAC.ShellExecute "cmd.exe", "/c copy /b /y ""$source"" ""$destination""", "", "runas", 0
 Wscript.Sleep(300)
 EOT;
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, also probably should be targetting the 2.0 branch.

src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -427,6 +427,12 @@ protected function setLocalPhar($localFilename, $newFilename, $backupTarget = nu
         try {
             rename($newFilename, $localFilename);
 
+            if ($this->isWindowsAdminUser()) {
+                // Windows uses MoveFileEx which uses the source permissions, so
+                // reset to inherited permissions to keep access for other users
+                exec('icacls.exe "'.$localFilename.'" /reset');
```
stof, line 433: shouldn't this use proper argument escaping to avoid issues ? Composer already has this in its ProcessExecutor.

johnstevenson: Possibly, but for consistency I'd then have to escape the commands for the VBScript as well - and I have no idea how its `ShellExecute` function parses its params (and to be truthful, no desire to find out!).

Another solution would be to use PHP's `copy` function and delete the source file on success, thus removing the need for `icacls`.

Seldaek: Hm yeah copy + delete with a comment explaining why sounds better to me than calling external processes with obscure names :)

johnstevenson: Okay. Just for windows? Or for all platforms?

Incidentally, in the installer script we use `rename` but the source file is created in the destination folder (as `composer-temp.phar`), so similar semantics to `copy`.

stof: or maybe create it in the destination folder, as done in the installer script.

johnstevenson: @stof For rollbacks the source exists in the user's `data-dir` though.

Seldaek: Hmm I'd say just for windows unless we have a reason to do this everywhere? It seems to work fine as is in other places?

johnstevenson:
I've rebased this to make it cleaner. Copy + delete used throughout on Windows. Test phar: [composer-2.0-dev+c8c7bed.zip](https://github.com/composer/composer/files/6070866/composer-2.0-dev%2Bc8c7bed.zip)

src/Composer/Command/SelfUpdateCommand.php
johnstevenson: Implemented in c8c7bed

Seldaek:
Thanks

johnstevenson:
@Seldaek Shouldn't this be in 1.10 as well?

Seldaek:
It could be.. but I personally don't feel it's a common/critical enough problem to warrant patching there, seeing as it's kinda EOL.

johnstevenson:
The other candidate would be fixing the HTTP_PROXY_REQUEST_FULLURI stuff introduced in 2.0. I get that 1.10 is EOL but these are sort of infrastructure bugs, albeit uncommon, rather than operational ones.

Entirely up to you, of course.

Seldaek:
Yeah I just rather spend my time on 2.x unless absolutely necessary. This is more interesting, and the work benefits more people..

PR #9734, opened 2021-02-26
PHPStan Level 3 and interface cleanups

Seldaek:
To be able to raise to level 2 I had to do some cleaning up of PackageInterface and CompletePackageInterface, as well as add some restrictions to which types are passed in to AliasPackage and in the Solver/Pool to force BasePackage instances in a few places.

I don't think this is a huge deal in terms of public-facing BC, we don't really have the assumption that people should implement their own package instances. From a consumer point of view people having PackageInterface type-hinted should not cause any problem.

I'd even go as far as marking all these things `@internal` for clarity, but I am not sure if that will cause issues in other packages if they type-hint an internal interface?

### DIFF
--- a/src/Composer/Package/Link.php
+++ b/src/Composer/Package/Link.php
@@ -26,6 +26,11 @@ class Link
     const TYPE_PROVIDE = 'provides';
     const TYPE_CONFLICT = 'conflicts';
     const TYPE_REPLACE = 'replaces';
+    /**
+     * TODO should be marked private once 5.3 is dropped
+     * @private
+     */
+    const TYPE_UNKNOWN = 'relates to';
 
     /**
      * Will be converted into a constant once the min PHP version allows this
@@ -58,7 +63,7 @@ class Link
 
     /**
      * @var string
-     * @phpstan-var self::TYPE_* $description
+     * @psalm-var self::TYPE_* $description
      */
     protected $description;
 
@@ -73,15 +78,14 @@ class Link
      * @param string              $source
      * @param string              $target
      * @param ConstraintInterface $constraint  Constraint applying to the target of this link
-     * @param string              $description Used to create a descriptive string representation
-     * @phpstan-param self::TYPE_* $description
+     * @param self::TYPE_*        $description Used to create a descriptive string representation
      * @param string|null $prettyConstraint
      */
     public function __construct(
         $source,
         $target,
         ConstraintInterface $constraint,
-        $description = 'relates to',
+        $description = self::TYPE_UNKNOWN,
         $prettyConstraint = null
     ) {
         $this->source = strtolower($source);
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -12,9 +12,15 @@
 
 namespace Composer\Package\Loader;
 
-use Composer\Package;
 use Composer\Package\AliasPackage;
+use Composer\Package\BasePackage;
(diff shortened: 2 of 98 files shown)


src/Composer/Package/Link.php
```diff
@@ -26,6 +26,11 @@ class Link
     const TYPE_PROVIDE = 'provides';
     const TYPE_CONFLICT = 'conflicts';
     const TYPE_REPLACE = 'replaces';
+    /**
+     * TODO should be marked private once 5.3 is dropped
```
stof, line 30: private constants don't only require dropping PHP 5.3

Seldaek:
Bumped PR to PHPStan level 3 🥳

src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -101,8 +133,18 @@ private function createObject(array $config, $class)
         return new $class($config['name'], $version, $config['version']);
     }
 
-    private function configureObject($package, array $config)
+    /**
+     * @template PackageClass of CompletePackageInterface
```
stof, line 137: this template is unused

src/Composer/Plugin/PluginManager.php
```diff
@@ -462,15 +462,18 @@ protected function getCapabilityImplementationClassName(PluginInterface $plugin,
...
-     * @param  string          $capabilityClassName The fully qualified name of the API interface which the plugin may provide
-     *                                              an implementation of.
-     * @param  array           $ctorArgs            Arguments passed to Capability's constructor.
-     *                                              Keeping it an array will allow future values to be passed w\o changing the signature.
-     * @return null|Capability
+     * @param  class-string<CapabilityClass> $capabilityClassName The fully qualified name of the API interface which the plugin may provide
+     *                                                            an implementation of.
+     * @param  array                         $ctorArgs            Arguments passed to Capability's constructor.
+     *                                                            Keeping it an array will allow future values to be passed w\o changing the signature.
+     * @return null|CapabilityClass
```
stof, line 476: given that PhpStorm does not yet support generics with `@template` (I don't know what the state is for other PHP IDEs as I haven't used them in years), it might be better to keep the enhanced type definitions in `@phpstan-return` and `@phpstan-param` annotations instead. this would avoid issues reporting that the method return a non-existent `CapabilityClass` class.

src/Composer/Repository/Vcs/PerforceDriver.php
```diff
@@ -147,7 +147,7 @@ public function hasComposerFile($identifier)
      */
     public function getContents($url)
     {
-        return false;
+        throw new \LogicException('Not implemented/used in PeforceDriver');
```
stof, line 150: typo. `PerforceDriver`.

And maybe `BadMethodCallException` (which extends `LogicException`) could be used here

src/Composer/Util/PackageSorter.php
```diff
@@ -22,15 +22,15 @@ class PackageSorter
      *
      * Packages of equal weight retain the original order
      *
-     * @param  array $packages
+     * @param  PackageInterface[] $packages
      * @return array
```
stof, line 26: `PackageInterface[]` for the return type too

src/Composer/Package/Link.php
Seldaek: But we'll for sure drop up to 7.1 if not 7.2, and 7.1 supports them so yes within the context of this project this statement is true.

PR #9757, opened 2021-03-06
Update GitHub token pattern

Ayesh:
GitHub is updating the format of auth tokens from `a-z0-9` to `A-Za-z0-9` ([notice](https://github.blog/changelog/2021-03-04-authentication-token-format-updates/)).
I'm not sure why `.` is allowed, but I dare not to remove it. In this PR, the token validation regex is updated to allow `A-Za-z0-9` instead of the current all lower-case `a-z` and disallowed `_`.

### DIFF
--- a/src/Composer/IO/BaseIO.php
+++ b/src/Composer/IO/BaseIO.php
@@ -124,7 +124,7 @@ abstract class BaseIO implements IOInterface
         }
 
         foreach ($githubOauth as $domain => $token) {
-            if (!preg_match('{^[.a-z0-9]+$}', $token)) {
+            if (!preg_match('{^[.A-Za-z0-9_]+$}', $token)) {
                 throw new \UnexpectedValueException('Your github oauth token for '.$domain.' contains invalid characters: "'.$token.'"');
             }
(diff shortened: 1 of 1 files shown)


Seldaek:
The dot is because of https://github.com/composer/composer/commit/ad41c324d3c473df1bdaabbac4ea4f08aabf1d73 - anyway thanks for pointing this out! I'll merge this in 1.10 and above and add some docs to explain things too.

Seldaek:
Merged as https://github.com/composer/composer/commit/dc83ba93f3d8a35629f9a387632e8cd373a144d0

denniskoenig:
Even if it is now closed, why is the format checked at all? Is it necessary for something or just a convenience function for the user? By removing it (if possible), one thing less could break when another system is changing something.

Seldaek:
I believe this was added because people entered random stuff in there and were confused it didn't work, so we do some basic sanity checking. I believe the current regex is permissive enough that this should be rather forward compatible now, but yeah only time will tell :)

denniskoenig:
If the check is too permissive, it is maybe not so useful anymore, but probably does no harm also.
Thanks for the fast response, answered my question.

PR #9771, opened 2021-03-11
Only call sapi_windows_set_ctrl_handler() for CLI requests

brandonkelly:
We are seeing the following PHP error when Composer is run from a web request using Windows:

> Error updating Composer requirements: CTRL events trapping is only supported on console

Here’s where it originates in PHP: https://github.com/php/php-src/blob/cb7b49dcf84846a79a19d75d0cdb82a9869c0465/win32/signal.c#L107-L110

We’ve verified that this PR fixes it, by checking if it’s a CLI request in addition to whether the `sapi_windows_set_ctrl_handler()` function exists.

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -405,7 +405,7 @@ EOT
             }
         }
         // handler Ctrl+C for Windows on PHP 7.4+
-        if (function_exists('sapi_windows_set_ctrl_handler')) {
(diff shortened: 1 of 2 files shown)


Seldaek:
I wonder why this is even defined on a web SAPI then 🤦🏻‍♂️ Anyway thanks for the fix, can you please rebase/retarget it to the 2.0 branch though?

brandonkelly:
@Seldaek Doh sorry, done.

Seldaek:
Thanks!

PR #9778, opened 2021-03-15
MaxFileSizeExceededException should reject download job

glaubinix:
MaxFileSizeExceededException didn't trigger the job reject callback and never deleted the temporary file with the `~` at the end.


Seldaek:
Thanks

PR #9782, opened 2021-03-16
Allow optional name and description fields for project type

BoShurik:
According to documentation [name](https://getcomposer.org/doc/04-schema.md#name) and [description](https://getcomposer.org/doc/04-schema.md#description) are optional for non-published packages

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -3,7 +3,26 @@
     "name": "Package",
     "type": "object",
     "additionalProperties": false,
-    "required": [ "name", "description" ],
+    "oneOf": [
+        {
+            "properties": {
+                "type": {
+                    "not": {
+                        "enum": ["project"]
+                    }
+                }
+            },
+            "required": ["name", "description"]
+        },
+        {
+            "properties": {
+                "type": {
+                    "enum": ["project"]
+                }
+            },
+            "required": ["type"]
+        }
+    ],
     "properties": {
         "name": {
             "type": "string",
--- a/tests/Composer/Test/Json/ComposerSchemaTest.php
+++ b/tests/Composer/Test/Json/ComposerSchemaTest.php
@@ -41,18 +41,35 @@ class ComposerSchemaTest extends TestCase
     {
         $json = '{ }';
         $result = $this->check($json);
-        $this->assertContains(array('property' => 'name', 'message' => 'The property name is required', 'constraint' => 'required'), $result);
-        $this->assertContains(array('property' => 'description', 'message' => 'The property description is required', 'constraint' => 'required'), $result);
+        $this->assertContains(array('property' => 'type', 'message' => 'The property type is required', 'constraint' => 'required'), $result);
 
         $json = '{ "name": "vendor/package" }';
         $this->assertEquals(array(
+            array('property' => 'type', 'message' => 'The property type is required', 'constraint' => 'required'),
             array('property' => 'description', 'message' => 'The property description is required', 'constraint' => 'required'),
+            array('property' => '', 'message' => 'Failed to match exactly one schema', 'constraint' => 'oneOf'),
         ), $this->check($json));
 
         $json = '{ "description": "generic description" }';
         $this->assertEquals(array(
+            array('property' => 'type', 'message' => 'The property type is required', 'constraint' => 'required'),
             array('property' => 'name', 'message' => 'The property name is required', 'constraint' => 'required'),
+            array('property' => '', 'message' => 'Failed to match exactly one schema', 'constraint' => 'oneOf'),
         ), $this->check($json));
+
+        $json = '{ "type": "library" }';
+        $this->assertEquals(array(
+            array('property' => 'type', 'message' => 'Does not have a value in the enumeration ["project"]', 'constraint' => 'enum', 'enum' => array('project')),
+            array('property' => 'name', 'message' => 'The property name is required', 'constraint' => 'required'),
+            array('property' => 'description', 'message' => 'The property description is required', 'constraint' => 'required'),
+            array('property' => '', 'message' => 'Failed to match exactly one schema', 'constraint' => 'oneOf'),
+        ), $this->check($json));
+
+        $json = '{ "type": "project" }';
+        $this->assertTrue($this->check($json));
+
+        $json = '{ "name": "vendor/package", "description": "description" }';
+        $this->assertTrue($this->check($json));
     }
 
     public function testOptionalAbandonedProperty()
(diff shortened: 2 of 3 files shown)


stof:
does this account for cases where the type property is omitted ?

BoShurik:
No. Trying to figure out how to do this.

Seldaek:
We already take care of tweaking the schema before validation in `composer validate`, so unless this can be really solid I'd rather keep it as is, defining the name is anyway a good idea IMO as you get more sensible errors than stuff mentioning __root__ which is the default name otherwise, and you can leave the description empty so it's not too dramatic.

BoShurik:
The problem that `symfony/flex` removes those keys automatically and `PhpStorm` validates `composer.json` according to schema. So all `composer.json` files in symfony projects are highlighted as invalid.

Anyway, looks like it's impossible to implement this check in json schema

Seldaek:
Then that maybe should be fixed in symfony/flex. Can you give more details as to when this happens? /cc @nicolas-grekas

BoShurik:
`PhpStorm` gets schema from https://www.schemastore.org/json/
Previously there was a patched version with similar changes: https://github.com/SchemaStore/schemastore/pull/505
Now schema has been updated: https://github.com/SchemaStore/schemastore/pull/1498 and this problem occurs

nicolas-grekas:
Maybe SchemaStore/schemastore#1498 should be reverted. Name and description are optional for top level projects, flex has nothing to put there...

stof:
Well, https://github.com/SchemaStore/schemastore/pull/1498 should not be fully reverted, otherwise it looses the sync of all the newer changes done in composer's schema.
The issue is that the official composer schema needs to be patched to be able to use it to validate composer.json files, making it unusable directly by other tooling than Composer which are not aware of the patching.

There is 2 solutions there:
1. composer should fix its schema so that it does not require patching (what this PR attempts)
2. SchemaStore must implement a process to re-apply the patch each time they synchronise the official composer schema

res/composer-schema.json
```diff
@@ -3,7 +3,25 @@
...
+                }
+            },
+            "required": ["name", "description"]
+        },
+        {
+            "properties": {
+                "type": {
+                    "enum": ["project"]
+                }
+            }
```
stof, line 22: shouldn't this alternative mark the `type` property as required ? Otherwise, a package without an explicit type will be allowed to match both alternatives.

guilliamxavier:
There have been additional patches in https://github.com/SchemaStore/schemastore/commits/master/src/schemas/json/composer.json recently. Currently a diff between https://getcomposer.org/schema.json and https://json.schemastore.org/composer.json gives 2 changes:

- line 3 (here showing 1-4):
```diff
 {
     "$schema": "https://json-schema.org/draft-04/schema#",
-    "name": "Package",
+    "title": "Package",
     "type": "object",
```

- line 45 (here showing 42-46):
```diff
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
-            "pattern": "^v?\\d+(((\\.\\d+)?\\.\\d+)?\\.\\d+)?|^dev-"
+            "pattern": "^v?\\d+(((\\.\\d+)?\\.\\d+)?\\.\\d+)?(-dev|((-patch|-p|-alpha|-a|-beta|-b|-RC)(\\d+)?))?$"
         },
```

And yeah, the "Missing required properties 'description', 'name'" (even with `"type": "project"` and `"license": "proprietary"`) is annoying

guilliamxavier:
For the record: while waiting for things to move here (maybe), I have submitted https://github.com/SchemaStore/schemastore/pull/1617

Seldaek:
@guilliamxavier this isn't accurate IMO as it does not handle cases where the type is not defined, which is perfectly valid. See https://github.com/composer/composer/pull/9782#issuecomment-800243552 

Anyway I am not sure what we should do here.. We could make the name/description optional in the schema, and change the rules at runtime to make them required instead of the opposite (which is what we do now). I think that probably makes the schema the most re-usable while not changing much for composer users. 

Any more thoughts on this?

tests/Composer/Test/Json/ComposerSchemaTest.php
```diff
@@ -41,18 +41,35 @@ public function testRequiredProperties()
     {
         $json = '{ }';
         $result = $this->check($json);
-        $this->assertContains(array('property' => 'name', 'message' => 'The property name is required', 'constraint' => 'required'), $result);
-        $this->assertContains(array('property' => 'description', 'message' => 'The property description is required', 'constraint' => 'required'), $result);
+        $this->assertContains(array('property' => 'type', 'message' => 'The property type is required', 'constraint' => 'required'), $result);
```
guilliamxavier, line 44:
```suggestion
        $this->assertContains(array('property' => 'type', 'message' => 'The property type is required', 'constraint' => 'required'), $result);
        $this->assertContains(array('property' => 'name', 'message' => 'The property name is required', 'constraint' => 'required'), $result);
        $this->assertContains(array('property' => 'description', 'message' => 'The property description is required', 'constraint' => 'required'), $result);
        $this->assertContains(array('property' => '', 'message' => 'Failed to match exactly one schema', 'constraint' => 'oneOf'), $result);
```

guilliamxavier:
@Seldaek: This PR has been updated since that comment, all type cases are correctly handled now, as shown in the tests here (I have just added one last suggestion) and on SchemaStore

Seldaek:
@guilliamxavier the problem still exists that this makes the type property required if you do not set a name/description. Before it was optional.. But at least setting it to project lets you omit the name/description, so I guess this is maybe an ok compromise.

Anyway I merged and tweaked things a little bit in https://github.com/composer/composer/commit/458bd41d8fe8df7e3da7dad8cfd09ea4dfc0d3cd to make it more user friendly:

image

Thanks everyone!

guilliamxavier:
> the problem still exists that this makes the type property required if you do not set a name/description. Before it was optional..

Hmm... If you don't set a `name`/`description`:
- Before, you were getting an error anyway (missing required name/description), independently of `type`'s value (or absence, equivalent to default "library"). 
- Now, for `"type": "project"` you don't get an error, otherwise you get a "composite" error (failed to match exactly one schema) with two "branches": the same error as before (missing required name/description), and:
  * if you set a `type` with any value other than "project": (type value not in enum ["project"]);
  * if you don't set a `type` at all: (missing required type).

Maybe the "problem" is that the error "tree" is "flattened" in the console? 🤔

> Anyway I merged and tweaked things a little bit in https://github.com/composer/composer/commit/458bd41d8fe8df7e3da7dad8cfd09ea4dfc0d3cd to make it more user friendly:

Great!

> Thanks everyone!

Thank you!

Issue #9636, opened 2021-01-21
Add require-tools and 


nicolas-grekas:
This issue is a follow up of https://github.com/composer/composer/issues/5390#issuecomment-463625428

Composer has become the primary installation method of libraries and is so good at it that it is pretty commonplace nowadays to also install tools like static analyzers or testing framework with it.

But this comes at a price: tools are micro-applications with a lot of dependencies, and those dependencies can often be found in your application, or in other tools.

Every time you add one of them to your toolchain, your application becomes more constrained, and very often, for no good reason, because in most cases, you do not reference classes or interfaces from those
tools in your code.

Soon enough, you find yourself in the dependency hell Composer is supposed to protect you from.

I'm dreaming of a new section in our `composer.json` files, called `require-tools`, that you be aimed at solving this issue by letting you install one or several tools inside separate vendor dirs, so that:

1. Tools have zero influence on each other.
2. Tools have zero influence on the dependencies of your project,
   meaning you will always be able to upgrade your project's dependencies,
3. In the case your tool has dependencies in common with your project, the
   tool will use your project's constraints into account.

The proposed section would not resolve the issue with tools that do require a version of a dep that is not compatible with your project. This is by design: as discussed in https://github.com/composer/composer/issues/5390#issuecomment-463625428, a phar could be the way to go for such tools.

A minimal example could look like the following:
```json
"require-tools": {
    "phpstan": { "phpstan/phpstan": "^0.9.2" }
}
```

Some tools like Behat are extensible, and you would be able to install behat and its extensions like this:
```json
"require-tools": {
    "behat": {
        "behat/behat": "^3.4.3",
        "behat/mink": "^1.7.1",
        "behat/mink-browserkit-driver": "^1.3.3",
        "behat/symfony2-extension": "^2.1.4"
    }
}
```

Of course, we could install phpstan next to behat like that:
```
"require-tools": {
    "phpstan": { "phpstan/phpstan": "^0.9.2" },
    "behat": {
        "behat/behat": "^3.4.3",
        "behat/mink": "^1.7.1",
        "behat/mink-browserkit-driver": "^1.3.3",
        "behat/symfony2-extension": "^2.1.4"
    }
}
```

The goal of these sections would be to create scoped vendor directories per entry in the `require-tools` sections. E.g. `vendor/composer/tools/phpstan` could contain all the deps of phpstan, etc.

[...]

PR #9792, opened 2021-03-24
Add AppRepository to enable tools inside apps

nicolas-grekas:
This is a tentative PR towards a better separation of tools' dependencies vs app's dependencies. This is related to #9636 and the discussion that is happening there.

Inside a dummy app, I created a `tools/test/composer.json` with the following content:
```json
{
    "config": {
        "bin-dir": "."
    },
    "require": {
        "app/app": "*",
        "phpunit/phpunit": "^9.5"
    },
    "repositories": [
        {
            "type": "app",
            "url": "../../"
        }
    ]
}
```

Running `composer install` inside this `tools/test/` directory installs PHPUnit and its dependencies in `tools/test/vendor`, with symlinks to the root app for shared dependencies.

Since many apps don't have a `name` in their `composer.json` file, the `app/app` that you see in the `require` section is a convention I added to allow tools to reference the app itself as a dependency when needed (for PHPUnit typically). If a tool doesn't need access to the app's source code, this entry should be just skipped.

By default, packages found in the app's `vendor/` directory are locked: a tool that shares a dependency with an app will always resolve that dependency to the same exact version as the app. This can be changed by making the above `app` repository "non-canonical", as explained in [Composer's doc](https://getcomposer.org/doc/articles/repository-priorities.md#canonical-repositories).

I feel like this PR is a small but useful step towards splitting tools' deps from app's ones. The DX could be improved for sure, but this can be done in future PRs I think.

Before adding tests: is this worth pursuing? I think yes :) What about you?

TL;DR: running phpunit is done by typing `$> ./tools/test/phpunit`


src/Composer/Repository/AppRepository.php
```diff
@@ -0,0 +1,161 @@
...
+ *         "url": "../../relative/path/to/app/",
+ *         "options": {
+ *             "symlink": false
+ *         }
+ *     },
+ * ]
+ * @endcode
+ *
+ * @author Samuel Roze <samuel.roze@gmail.com>
+ * @author Johann Reinke <johann.reinke@gmail.com>
```
OskarStark, line 54: True?

nicolas-grekas: updated :)

glensc:
> TL;DR: running phpunit is done by typing $> ./tools/test/phpunit

shouldn't that be `./vendor/bin/phpunit`?

gmsantos:
It would be nice if composer scripts can find `./tools/test/phpunit` somehow https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands

naderman:
I think adding a separate composer.json file for this which may exist in different directories based on the project is really a step backward from standardized composer.json at the root of the project. If this is to be implemented I think it should be as part of the project's composer.json as discussed in previous issues and the one you linked.

nicolas-grekas:
@naderman I totally agree, as I suggested in the description, this PR is a step towards this target. I feel safer to take this path than to dream about an issue without anyone actually contributing some code :)

mabar:
If I for example installed phpstan and phpunit as a tools, would I be able to analyse phpunit tests with phpstan? From what I understand, phpstan would know nothing about phpunit.

I think phars are better solution for tools. Phar can be published as a separate package, with no dependencies, so it can be still installed via Composer. For example phpstan do that - [phpstan/phpstan-src](https://github.com/phpstan/phpstan-src) is published via [phpstan/phpstan](https://github.com/phpstan/phpstan) phar

nicolas-grekas:
@mabar yes, phpstan would read the `tools/test/vendor/` to learn everything it needs to about phpunit.
About phar's, this has been discussed in the linked issue: they don't work well with extensions/plugins, and they require careful care by authors, which means we cannot expect all tools to have a corresponding phar anyway.

nicolas-grekas:
@gmsantos this works already, what else would you need?
```
    "scripts": {
        "test": [
            "./tools/test/phpunit"
        ]
    },
```

gmsantos:
@nicolas-grekas it would be related with this note from the docs:

> Note: Before executing scripts, Composer's bin-dir is temporarily pushed on top of the PATH environment variable so that binaries of dependencies are directly accessible. In this example no matter if the phpunit binary is actually in vendor/bin/phpunit or bin/phpunit it will be found and executed.

If you just use phpunit, ./tools/test/phpunit would not be found, i.e. 

    "scripts": {
        "test": [
            "phpunit"
        ]
    },

nicolas-grekas:
@glensc @gmsantos now achieved by changing the `bin-dir` in the composer.json file, see updated description.

src/Composer/Repository/AppRepository.php
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        $json = file_get_contents($composerFilePath);
+        $package = JsonFile::parseJson($json, $composerFilePath);
+        $package['name'] = 'app/app';
+        $package['version'] = '99999';
+        $package['dist'] = array(
+            'type' => 'path',
+            'url' => $this->url,
+            'reference' => sha1(serialize($this->options)),
```
glensc, line 124: should the `$this->options` be sorted to ensure a reproducible outcome across systems?

nicolas-grekas: Composer already uses similar hashes internally. There's no need to sort because the structure is stable, since it's loaded from files. When the file changes, changing the reference is fine.

Jean85:
Nice experiment!

How would the autoloaders interact in such a case?

I can see multiple situations in which I would need to load the app and multiple tools at the same time... For example, PHPUnit, phpspec/prophecy-phpunit (and its transient dependency Prophecy) and the app's code. Plus, I would want to run static analysis (and its plugins!) on top of that.

nicolas-grekas:
@Jean85 there is one autoloader per "tools" directory, which can load the app or not depending on the `app/app` in the require above. When multiple tools should share the same autoloader, they should be required in the same `tools/foo/` directory. I suppose there won't be many of them actually, as grouping many tools in one set of shared deps could be fine, as long as individual tools allow it.

Jean85:
@nicolas-grekas and what about transient dependencies? Let's say, both the app and the tool require symfony/console (which is very likely), the first to be loaded wins? Wouldn't that create clashing?

nicolas-grekas:
I suppose you mean transitive dependencies, right? What I wrote in the description: the vendor/ dir of the app will be used as the package repository for shared deps. This means that the tool's console directory will be a symlink to the app's one, by default at least (see PR description, about canonical repo for more).

Jean85:
So, let me rephrase is to see if I get this right: transitive (yep, sorry I meant that) dependencies that are present in both the app and the tool wouldn't get duplicated, but instead the tool will use the same from the app, with a symlink if needed. You approach will reduce the "clash" because transitive dependencies that do NOT overlap with the app will be resolved in a different tree, reducing the "pollution" that the app would get from the normal `require-dev` procedure.

Did I get this right? If this is the case, we still have the (big) issue of common transitive deps, and that it's a tough problem to solve.

nicolas-grekas:
I think you've got this right. Which issue are you talking about? To me they are all solved:
- if a tool needs to be able to autoload the classes of your app, any shared deps *must* be compatible with the app's deps. This PR provides that.
- if a tool doesn't need to autoload the classes of your app, this approach can provide opportunistic sharing of shared deps, but allows any other version of said deps to be installed for the need of a specific tool.

Jean85:
I'm just pondering the full impact of your solution... surely you solved a big chunk of possible conflicts, but the tools still needs to be compatible with your own set of dependencies. Anyway, as said above, this shouldn't be a big issue for applications (where lock is present and target is known) and for libraries like Symfony too (where you should aim to reduce your dependencies). Good! :+1:

mxr576:
I see that this looks like a valid workaround for the problem but is it the right way to get it solved? Aren't we trying to abuse Composer? 

Can we collect and compare how others are dealing with this problem in the industry? (Like how it happened with [Enum PSR](https://github.com/Crell/enum-comparison) recently.)

I am more like with those who asked would not be better using PHAR for development dependencies? Figuring out best practices around that instead of trying to add tricks to a dependency manager.

PHAR is similar to [Snap](https://snapcraft.io/)/[Flatpak](https://flatpak.org/) in my vocabulary and they were born for a reason.

Related: https://tomasvotruba.com/blog/2019/12/02/how-to-box-symfony-app-to-phar-without-killing-yourself/

Jean85:
The limitation of the PHAR approach is that every single maintainer has to apply it to its project; if we can bake in a similar solution inside Composer, maybe with a scoper too, everyone will benefit from that for free.

WinterSilence:
Is it still needed? `Composer\InstalledVersions` contains similar functionality.

PR #9793, opened 2021-03-24
Add dev mode env var for scripts run command

vitman:
I propose to add COMPOSER_DEV_MODE env var for run-scripts command.

It's useful if i want to run some shell(or any non-php) script in hook, so It's impossible to check passed php var inside event.


Seldaek:
Right that makes sense, thanks.

PR #9803, opened 2021-04-01
Optimize the `ClassLoader` class with FQFN prefixes

Ayesh:
Performance of the `\Composer\Autoload\ClassLoader` class is paramount because it is run on all projects that depend on Composer. Every bit of performance with squeeze is worth it.

Because this class is namespaced, PHP does not attempt to inline or otherwise [optimize internal functions and constant](https://php.watch/articles/php-zend-engine-special-inlined-functions). Making them fully-qualified (with a backslash prefix, e.g. `\strlen`) enables the optimizations back. Since Composer needs to work on older PHP versions, it cannot use the more cleaner `use function strlen` pattern, but the `\strlen` still works.

This change applies the backslash to all internal PHP functions and APCu functions, making it possible the engine to inline them where possible. With this, PHP can inline its [special functions](https://php.watch/articles/php-zend-engine-special-inlined-functions#func-list), and replace `INIT_NS_FCALL_BY_NAME` OPCodes with the more performant `INIT_FCALL` OPCodes. In addition to using more performant OPCodes, the [absolute number of OPCodes](https://gist.github.com/Ayesh/adade5efc78c731efc0fd2b180d69fbb) is reduced by ~25 too.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -69,7 +69,7 @@ class ClassLoader
     public function getPrefixes()
     {
         if (!empty($this->prefixesPsr0)) {
-            return call_user_func_array('array_merge', array_values($this->prefixesPsr0));
+            return \call_user_func_array('array_merge', \array_values($this->prefixesPsr0));
         }
 
         return array();
@@ -101,7 +101,7 @@ class ClassLoader
     public function addClassMap(array $classMap)
     {
         if ($this->classMap) {
-            $this->classMap = array_merge($this->classMap, $classMap);
+            $this->classMap = \array_merge($this->classMap, $classMap);
         } else {
             $this->classMap = $classMap;
         }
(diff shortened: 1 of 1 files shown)


Seldaek:
I'm not sure this is really worth the awful looking code..

- findFile is the only hot path, as this is what is used at runtime, everything else used to configure the ClassLoader is not executed when on php 5.6+ as the class gets initialized via autoload_static.php
- if you care about performance, you should use an authoritative classmap really.. in which case the findFile code is reduced to:

        if (isset($this->classMap[$class])) {
            return $this->classMap[$class];
        }
        if ($this->classMapAuthoritative || isset($this->missingClasses[$class])) {
            return false;
        }

And there the backslashing does not help at all. 

So I'd say.. maybe we could keep the backslashes findFile/findFileWithExtension as those will benefit people not using an authoritative classmap slightly, but the rest I would rather not clutter the code for nothing.

Ayesh:
Understood, thanks for the quick response.

I will close this PR for now, because this one is a more YOLO approach to backslashes without filtering the hot paths. It's very difficult to make a meaningful benchmark, with an autoloader, but even so, I don't think we will see a meaningful difference, and I don't think the maintenance burden is worth that tiny fraction either. 

Thank you.

Issue #9740, opened 2021-03-01
On securing Composer updates with The Update Framework (TUF)


TravisCarden:
Several open source PHP projects including Drupal, Joomla, and Typo3 are working on a Composer-based implementation of [The Update Framework (TUF)](https://theupdateframework.io/) for securing software update systems ([php-tuf](https://github.com/php-tuf)). TUF provides protection against various attacks including those involving compromised repositories or signing keys. Before we undertake to implement it, we want to make sure we don't duplicate effort and that whatever we end up doing has the potential to benefit the greatest number of Composer users. Is this something that interests the maintainers here? Would you like to collaborate on provide guidance? What are your thoughts?

trishankatdatadog:
Just so maintainers have a better idea, I think the projects mentioned above are trying to achieve something like [PEP 458](https://www.python.org/dev/peps/pep-0458/) level of security for at least some PHP package repositories (such as for Drupal).

joshuagl:
Because [PEP 458](https://www.python.org/dev/peps/pep-0458/) is long, I wanted to try and provide an elevator pitch.

[PEP 458](https://www.python.org/dev/peps/pep-0458/) is a design for repository signing with TUF which is transparent to both developers (uploading packages) and users (downloading packages). Repository signing ensures that package contents have not been modified (tampered with) from the time the package is added to the repository to when the client downloads and installs it.

One of the nice things about the [PEP 458](https://www.python.org/dev/peps/pep-0458/) design is that it can be extended (at a later time) to add optional developer signing of packages as described in [PEP 480](https://www.python.org/dev/peps/pep-0480/) (note we intend to update this PEP Real Soon Now™️). Developer signing gives the additional security benefits of ensuring the package contents are not modified from the time they are signed by the developer (before upload to the repository), and further provides origin authentication – that the key associated with a user was used to sign the package.

trishankatdatadog:
Thanks for the summary, Joshua!

> Repository signing ensures that package contents have not been modified (tampered with) from the time the package is added to the repository to when the client downloads and installs it.

IIUC, with PEP 458, you don't get that guarantee exactly: you get protection from MitM attacks between the repository and end-users (e.g., from malicious mirrors that are basically MitM). With PEP 480, you get the protection you mentioned above.

stof:
Given that the main composer repository (packagist.org) does not store the code but only the metadata, would TUF work there ?
[...]

PR #9817, opened 2021-04-07
Make ComposerRepository::configurePackageTransportOptions() protected

phenaproxima:
This is spun off from discussion in #9740.

The php-tuf/composer-integration plugin, which aims to help secure Composer operations with The Update Framework, has to do some..._interesting_ things to work properly. Amongst those interesting things is the roundabout use of a specialized package loader in order to set certain transport options on packages. @Seldaek pointed out that it would be easier and cleaner to just override ComposerRepository::configurePackageTransportOptions(), but unfortunately that's a private method. This PR proposes to change it to protected so that it can be extended and overridden by subclasses of ComposerRepository.


phenaproxima:
Whoops, I opened this with a bad branch. Closing this out so I can re-open another one using the 2.0 branch from my fork. Apologies!

PR #9819, opened 2021-04-07
Make ComposerRepository::asyncFetchFile() and ::fetchFileIfLastModified() protected

phenaproxima:
This is spun off from #9740, and follows the same rationale as #9818.

If ComposerRepository::asyncFetchFile() were protected, it would be possible to chain additional callbacks on the promises it returns. This would allow php-tuf/composer-integration to verify the downloaded metadata without having to do intrusive things like overriding the HttpDownloader. This also goes for ComposerRepository::fetchFileIfLastModified(); if it were protected, subclasses could take additional action after the fetch is completed.

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -1166,7 +1166,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
         return $data;
     }
 
-    private function fetchFileIfLastModified($filename, $cacheKey, $lastModifiedTime)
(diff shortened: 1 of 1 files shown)


phenaproxima:
Pinging @Seldaek for review of this one, because the workflow failures appear to be due to some temporary hiccup at Packagist. I could not figure out how to rerun the failed actions (I assume I don't have permission for that).

Seldaek:
IMO the PostFileDownloadEvent we discussed would be more appropriate as it'd hook right after the download is complete, and would give a chance to act before the file is used or cache or anything else happens to it.

phenaproxima:
That seems fair! Closing this out in favor of #9822.

PR #9822, opened 2021-04-08
Fire POST_FILE_DOWNLOAD event for metadata fetched by ComposerRepository

phenaproxima:
This is spun off from discussion in #9740 and #9819.

Right now, it's not possible for plugins (in this case, specifically php-tuf/composer-integration) to react to Composer downloading package metadata. The methods which do that grunt work cannot be overridden because they're private, and no event is dispatched after the metadata is downloaded. @Seldaek suggested modifying ComposerRepository to fire the POST_FILE_DOWNLOAD plugin event after metadata is downloaded, but before it is cached. This PR implements that.

This requires changes to the PostFileDownloadEvent class. Namely, it has be able to accept a type (`package` or `metadata`, similar to what PreFileDownloadEvent already does) and additional context beyond just an instance of `PackageInterface`.

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -192,7 +192,7 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
                 }
 
                 if ($eventDispatcher) {
-                    $postFileDownloadEvent = new PostFileDownloadEvent(PluginEvents::POST_FILE_DOWNLOAD, $fileName, $checksum, $url['processed'], $package);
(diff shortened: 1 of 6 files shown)


Seldaek:
Please check out my changes, see if that works for you? I figured passing the repository to metadata and having an array to allow for future additions in a BC way there if needed would be good to have.

phenaproxima:
I think those changes look great, and will serve our purposes beautifully. I considered adding the repository as context, but for TUF's purposes, the response was more valuable. But I agree that passing both in an array is more flexible and future-proof. Thanks!

Seldaek:
Alright then let's merge this for now, not sure what the timeline is on 2.1 tho but at least you can keep moving forwards based on snapshot builds.

PR #9826, opened 2021-04-09
Update InstallCommand.php

94noni:
Hi, small proposition to warn that running v1 can be upgraded to v2 when using the much used `composer install` command.
Feel free to close if already proposed/or not necessary, thank you,

### DIFF
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -86,9 +86,7 @@ EOT
             $io->writeError('<warning>You are using the deprecated option "dev". Dev packages are installed by default now.</warning>');
         }
 
-        if (version_compare(Composer::RUNTIME_API_VERSION, '2.0.0', '<')) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Yeah we probably need to nudge people a bit stronger at some point, but I think we may still need to wait a bit.

94noni:
@Seldaek thank you for your answer :)

At the end of this month, it will be half a year since release of v2, do you have a roadmap or plan for the v1 official end of life datetime?
I've read https://blog.packagist.com/deprecating-composer-1-support/, back then no delay were announced
thank you :)

Seldaek:
No we do not have a clear timeline, but v1 is still used in bigger numbers than v2, so I don't think it's very appropriate yet to call it dead, especially as some users are still unable to migrate to v2 due to plugin support etc. It's a large ecosystem, 6 months isn't that long.

94noni:
Friendly ping @Seldaek 
I remind this when reading https://github.com/composer/composer/issues/10340 and your tweet :)

Seldaek:
Yeah it's probably time now, thanks

PR #9827, opened 2021-04-09
ComposerRepository::asyncFetchFile() does not pass the downloaded URL to PostFileDownloadEvent

phenaproxima:
#9822 made a small mistake -- in `ComposerRepository::asyncFetchFile()`, we should be constructing `PostFileDownloadEvent` with the full URL of the downloaded metadata. Instead, we're passing it the base URL of the repository (possibly due to the ambiguous naming of the `$url` and `$filename` variables). Let's be sure to pass the actual downloaded URL to `PostFileDownloadEvent`, which is what `fetchFile()` and `fetchFileIfLastModified()` already do.


Seldaek:
Oh totally missed that, indeed somewhat poor variable naming there :D

PR #9829, opened 2021-04-11
"composer init --autoload" - Interactive generates PSR-4 autoloader in composer.json

camya:
Hello,

This PR extends the "composer init" command. It adds the "--autoload" parameter.

**Automatically generates a PSR-4 autoload** entry in your composer.json based on your **package name** and the interactive **configured src** path. The configured path is created automatically. The autoload feature is **optional**.

This is **helpful to kickstart a new project**, because PSR-4 autoload is required most of the time for new projects.

**Example info:**
In the following examples, the name of the new composer package is: `new_projects.acme-extra/package-name`

**Interactive mode**
`composer init`
...
`Add PSR-4 autoload mapping? Map namespace "Acme\Package" to the entered relative path. [src/, n to skip]:`

**Command parameter**
`composer init --autoload src/`

**Command --help**
`-a, --autoload=AUTOLOAD Add PSR-4 autoload mapping. Maps your package's namespace to the provided directory. (Expects a relative path, e.g. src/)`

**Implementation:**
Namespace: `new_projects.acme-extra/package-name` becomes namespace `NewProjectsAcmeExtra\PackageName`.
Path: Relative paths with trailing forward slash, defaults to `src/` (subfolders allowed)

**Create autoload files**
Composer "dump-autoload" runs automatically.

**Generated PSR-4 autoload entry in composer.json**
 ```
"autoload": {
    "psr-4": {
        "NewProjectsAcmeExtra\\PackageName\\": "src/"
    }
}
```

**Onboarding - Output after confirmed generation**
This is a simple help to onboard the developer. It allows to copy and paste the namespace and autoload require.

 ```
# PSR-4 autoload configured:
# Namespace: "NewProjectsAcmeExtra\PackageName"
# Path: "src/"

namespace NewProjectsAcmeExtra\PackageName;
```

**Documentation**

`* **--autoload (-a):** Add a PSR-4 autoload mapping to the composer.json. Automatically maps your package's namespace to the provided directory. (Expects a relative path, e.g. src/)  See also [PSR-4 autoload](04-schema.md#psr-4).`

**Unit tests**
Included

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -65,6 +65,7 @@ php composer.phar init
   the list of requires. Every repository can be either an HTTP URL pointing
   to a `composer` repository or a JSON string which similar to what the
   [repositories](04-schema.md#repositories) key accepts.
+* **--autoloader (-a):** Add PSR-4 autoloader.
 
 ## install / i
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -65,7 +65,7 @@ php composer.phar init
   the list of requires. Every repository can be either an HTTP URL pointing
   to a `composer` repository or a JSON string which similar to what the
   [repositories](04-schema.md#repositories) key accepts.
-* **--autoload (-a):** Add PSR-4 autoload.
+* **--autoload (-a):** Add a PSR-4 autoload mapping to the composer.json. Automatically maps your package's namespace to the provided directory. (Expects a relative path, e.g. src/) See also [PSR-4 autoload](04-schema.md#psr-4).
 
 ## install / i
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -65,7 +65,7 @@ php composer.phar init
   the list of requires. Every repository can be either an HTTP URL pointing
   to a `composer` repository or a JSON string which similar to what the
   [repositories](04-schema.md#repositories) key accepts.
-* **--autoloader (-a):** Add PSR-4 autoloader.
+* **--autoload (-a):** Add PSR-4 autoload.
 
 ## install / i
 
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -24,6 +24,7 @@ use Composer\Repository\CompositeRepository;
 use Composer\Repository\PlatformRepository;
 use Composer\Repository\RepositoryFactory;
 use Composer\Repository\RepositorySet;
+use Composer\Util\Filesystem;
 use Composer\Util\ProcessExecutor;
 use Composer\Semver\Constraint\Constraint;
 use Symfony\Component\Console\Input\ArrayInput;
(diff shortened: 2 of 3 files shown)


src/Composer/Command/InitCommand.php
```diff
@@ -69,6 +70,7 @@ protected function configure()
                 new InputOption('stability', 's', InputOption::VALUE_REQUIRED, 'Minimum stability (empty or one of: '.implode(', ', array_keys(BasePackage::$stabilities)).')'),
                 new InputOption('license', 'l', InputOption::VALUE_REQUIRED, 'License of package'),
                 new InputOption('repository', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Add custom repositories, either by URL or using JSON arrays'),
+                new InputOption('autoload', 'a', InputOption::VALUE_OPTIONAL, 'Add PSR-4 autoload'),
```
stof, line 73: The value should be required. Your code does not currently make a distinction between not passing `--autoload` and passing `--autoload` without value (handling `VALUE_OPTIONAL` options properly was a huge mess in older symfony/console versions).

src/Composer/Command/InitCommand.php
```diff
@@ -143,6 +157,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $file->write($options);
 
+        // --autoload - Create src folder
+        if ($autoloadPath) {
```
stof, line 161: I think it should do it only if `$input->isInteractive()` is true, just like installing dependencies is not done for your if you run it in non-interactive mode

src/Composer/Command/InitCommand.php
camya: Hi stof, nice to meet you. Thank you for the hints.

In my latest commit. ce0cf8e625634149718b67d5d016b6ab1aab9402 I updated the input option to a required value

src/Composer/Command/InitCommand.php
camya: Totally missed that. I now committed a fix for this. f0ae3751714c9eada5cbf5c78e08f69592b815dc 

It now creates the src folder and runs dump-autoload only if it's in interactive mode.

doc/03-cli.md
```diff
@@ -65,6 +65,7 @@ php composer.phar init
   the list of requires. Every repository can be either an HTTP URL pointing
   to a `composer` repository or a JSON string which similar to what the
   [repositories](04-schema.md#repositories) key accepts.
+* **--autoload (-a):** Add PSR-4 autoload.
```
Seldaek, line 68: This here needs to say it expects a path to src/ or something, it's not clear it expects anything from the description, and especially not *what*. Also applies to the InputOption's description below.

Seldaek:
Thanks, looks like a neat addition from a quick glance at the PR. Still have to try it out to see how it behaves live though.

doc/03-cli.md
camya: Hi Jordi, nice to meet you. 

I changed the wording for both messages (see below) but not commited them yet. 

Can you please review them? 

**composer init -h:**
`Add PSR-4 autoload mapping. Maps your package's namespace to the provided directory. (Expects a relative path, e.g. src/)`

**Command interaction:**
`Add PSR-4 autoload mapping? Map namespace "Acme\Package" to the entered relative path. [src/, n to skip]:`

If these are good for you, I will commit them asap.

Seldaek: Yup sounds good to me :)

camya: Nice, just updated the PR with the new wordings.

src/Composer/Command/InitCommand.php
Seldaek: I reverted this change because IMO the dependency update is not done when non-interactive because we can't ask if you want to do it. I don't really see what interactivity should change otherwise.

PR #9831, opened 2021-04-12
Allow PreFileDownloadEvent to carry transport options for metadata

phenaproxima:
As part of #9740 (and our proof of concept being implemented in php-tuf/composer-integration), we need to be able to prevent infinite-data attacks by passing a maximum number of bytes to the HTTP downloader when downloading TUF-validated package metadata. It looks like Composer already supports this using the `max_file_size` transport option (normally pulled from a package's transport options), but sadly that option, if used in a repository's transport options, cannot vary (at least, not easily) for different metadata files. While we could just implement some sort of hard limit, it would be best to specify the actual expected size of each individual download.

One way to solve this, which this PR implements, is to allow the PreFileDownloadEvent to provide transport options.


effulgentsia:
As an aside, while out of scope for this issue, I just want to make folks here aware that the TUF spec also intends to protect against slow retrieval attacks. See https://github.com/theupdateframework/tuf/issues/932 for background, and as part of that, the Python implementation recently abstracted out a RequestsFetcher in https://github.com/theupdateframework/tuf/pull/1250.

I don't know what that means for PHP / Composer: whether there are existing timeout settings that protect against slow retrieval, or if we'll need to add those somehow.

phenaproxima:
Pinging @Seldaek for review; hopefully I'm not breaching etiquette. (Please let me know if I am.)

Seldaek:
Thanks, makes sense I believe. Tweaked things some, would appreciate if you can double check.

- This is not applied to FileDownloader's downloads, but as I added in the docs IMO that is fine.
- Should the options here really be merged in as additional options? I'm thinking maybe it'd make more sense to pass `$this->options` into the event, and then overwrite that with a getTransportOptions once the event is done, that gives more flexibility and power to the event handlers, and I feel like it'd generally be more in line with how we do things in other places (letting you override the URL and such..)

phenaproxima:
>  I'm thinking maybe it'd make more sense to pass $this->options into the event, and then overwrite that with a getTransportOptions once the event is done, that gives more flexibility and power to the event handlers, and I feel like it'd generally be more in line with how we do things in other places (letting you override the URL and such..)

I agree. I'll change this.

> As for the etiquette, it's fine to ping if really needed, but by default it's just creating noise for nothing, unless it's urgent I'll ignore pings like I ignored the initial PR notification, and handle things when I have time :)

Duly noted! Thanks for the clarification.

Seldaek:
Thanks

PR #9832, opened 2021-04-12
Upgrade to xdebug-handler 2

johnstevenson:
This adds support for Xdebug3 modes and changes the default behaviour from always restarting if Xdebug is loaded, to only restarting if Xdebug is active.

Xdebug is considered active if it is loaded, and for Xdebug3, if it is running in a mode other than `xdebug.mode=off`.

### DIFF
--- a/bin/composer
+++ b/bin/composer
@@ -14,7 +14,7 @@ use Composer\XdebugHandler\XdebugHandler;
 error_reporting(-1);
 
 // Restart without Xdebug
-$xdebug = new XdebugHandler('Composer', '--ansi');
+$xdebug = new XdebugHandler('Composer');
 $xdebug->check();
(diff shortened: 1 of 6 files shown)


Seldaek:
Thanks!

Jimbolino:
will this be merged to 2.x branch?

currently we have some packages that cannot be installed in one project because one requires composer/xdebug-handler:^2 and another requires composer/composer:^1.6|^2
so ideally it would be nice if composer/composer works with both ^1.1|^2.0

Seldaek:
I guess I can backport this to 2.0.x, didn't realize this was going to block people.

Seldaek:
@Jimbolino https://github.com/composer/composer/releases/tag/2.0.14 is out now

PR #9840, opened 2021-04-21
Add Installation - Termux (Android)

PuneetGopinath:
Hi composer maintainers!

I just added how to install composer on termux (android) in the docs.

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -158,6 +158,22 @@ C:\Users\username>composer -V
 Composer version 2.0.12 2021-04-01 10:14:59
 ```
 
+## Installation - Termux (Android)
+
+Run
+
+```sh
+pkg install composer
+composer self-update
+```
+
+Then close your current terminal. Test usage with a new terminal:
+
+```sh
+$ composer -V
+Composer version 2.0.12 2021-04-01 10:14:59
+```
+
 ## Using Composer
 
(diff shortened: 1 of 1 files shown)


stof:
Installing composer using packages provided by distributions is not officially supported, and so not documented in the official documentation (for instance, Debian and Ubuntu have a `composer` package providing an outdated version of composer through `apt`, but this is not documented here).

PuneetGopinath:
Ok, Thanks

staabm:
> package providing an outdated version

maybe it would make sense to mention/warn something simliar in the the docs...

PuneetGopinath:
> maybe it would make sense to mention/warn something simliar in the the docs... 

Did you mean to warn/mention the version of composer available in termux?

stof:
> maybe it would make sense to mention/warn something simliar in the the docs...

well, the doc does not even mention that such package _exists_.

staabm:
> Did you mean to warn/mention the version of composer available in termux?

I mean mention in the readme that composer is installable over various ways (e.g. Debian and Ubuntu) which is not considered supported.

PuneetGopinath:
> > Did you mean to warn/mention the version of composer available in termux?
> 
> I mean mention in the readme that composer is installable over various ways (e.g. Debian and Ubuntu package manager) which is not considered supported.

This is a good idea, anyway @stof doesn't accept it, so I will close this pr.

Issue #9841, opened 2021-04-21
vendor/bin/*bat files when using WSL


staabm:
when using composer in WSL on windows, the composer cli does not generate *.bat files under vendor/bin/.

phpstorm on the other hand requires e.g. a phpstan.bat file when configuring the integration within phpstorm (.phar and other files are considered invalid).

would it be possible to either create the .bat files all the time (even on non-windows platform), or e.g. feature detect WSL and generate the *.bat files additionally in this case?

![grafik](https://user-images.githubusercontent.com/120441/115553267-81767c80-a2ad-11eb-8567-05e943a5d58a.png)

![grafik](https://user-images.githubusercontent.com/120441/115553282-850a0380-a2ad-11eb-9e00-3a0f6b71577c.png)

![grafik](https://user-images.githubusercontent.com/120441/115553343-97843d00-a2ad-11eb-9592-88347e2dc3d1.png)

tested on phpstorm 2021.1.1

stof:
is there an easy way to feature-detect WSL ?

Seldaek:
You can set `composer config -g bin-compat full` to always generate .bat files https://getcomposer.org/doc/06-config.md#bin-compat

Seldaek:
That said, I believe it's pretty easy to detect WSL so maybe we can do this automatically. Reopening.

PR #9855, opened 2021-04-29
BinaryInstaller: install full binaries on WSL when bin-compat=auto

clxmstaab:
feature detect WSL and install all binaries in this case.

this is usefull to get the *.bat files on a windows host, which can then be invoked from the regular CLI, or used within IDEs etc.

closes https://github.com/composer/composer/issues/9841

tested on Windows10x64 in WSL, CMD and git-bash-for-windows

### DIFF
--- a/src/Composer/Util/Platform.php
+++ b/src/Composer/Util/Platform.php
@@ -75,6 +75,11 @@ class Platform
     public static function isWindowsSubsystemForLinux() {
         if (null === self::$isWindowsSubsystemForLinux) {
             self::$isWindowsSubsystemForLinux = false;
+            
+            // while WSL will be hosted within windows, WSL itself cannot be windows based itself.
+            if (self::isWindows()) {
+                return self::$isWindowsSubsystemForLinux = false;
+            }
 
             $process = new ProcessExecutor();
             try {
--- a/src/Composer/Util/Platform.php
+++ b/src/Composer/Util/Platform.php
@@ -21,6 +21,8 @@ class Platform
 {
     /** @var ?bool */
     private static $isVirtualBoxGuest = null;
+    /** @var ?bool */
+    private static $isWindowsSubsystemForLinux = null;
 
     /**
      * Parses tildes and environment variables in paths.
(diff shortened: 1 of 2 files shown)


src/Composer/Util/Platform.php
```diff
@@ -67,6 +69,26 @@ public static function getUserDirectory()
...
+    /**
+     * @return bool Whether the host machine is running on the Windows Subsystem for Linux (WSL)
+     */
+    public static function isWindowsSubsystemForLinux() {
+        if (null === self::$isWindowsSubsystemForLinux) {
+            self::$isWindowsSubsystemForLinux = false;
+
+            $process = new ProcessExecutor();
+            try {
+                if (0 === $process->execute('cat /proc/version | grep "Microsoft"', $ignoredOutput)) {
```
clxmstaab, line 81: in https://github.com/composer/composer/issues/9841#issuecomment-828190917 there was another suggestion to use `/etc/wsl.conf` for the detect. this file was not present on my WSL though ¯\_(ツ)_/¯

stof, line 81: should we even attempt running this `cat` and `grep` commands on Windows ? I don't think that will work. That should be checked first to return `false`

clxmstaab: it would throw an exception and this will be handled in the catch block.
per my tests it works - perf wise it might be worth checking the platform beforehand.

the code gets a little bit weired though, since we would check for a non-windows platform in a method called  `isWindowsSubsystemForLinux` 8-)

stof: well, if you test the whole usage rather than this method alone, you just never execute it on Windows (as the only existing usage is short-circuited when `Platform::isWindows()` returns true).


> the code would get a little bit weird though, since we would check for a non-windows platform in a method called `isWindowsSubsystemForLinux` 8-)

and that's fine. WSL is not windows. It is linux (running inside windows)

clxmstaab: > well, if you test the whole usage rather than this method alone, you just never execute it on Windows (as the only existing usage is short-circuited when `Platform::isWindows()` returns true).

I have tested the function in isolation and it worked.

mvorisek: It must not fail composer when `exec` (or similar functions) are disabled, on Windows as well as on linux

Seldaek: Fixed to just do a file_get_contents, no need for a process running `cat` to read a file :)

mvorisek: right :)

Seldaek:
Thanks

PR #9861, opened 2021-05-01
Introduce a cross-platform safe version of is_readable to support UNC / wsl$ paths on Windows

### LINKED ISSUE #9858: UNC path support
When I run this command: 

```
composer i
```

I get the following output: 

```
'\\wsl$\Ubuntu-18.04\home\bilge'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
Composer could not find a composer.json file in C:\Windows
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
```

UNC path is supported.

---
With the increasing popularity of WSL and with respect to Docker support for WSL 2, it may be increasingly common to want to access the WSL container's files from outside the container. UNC paths are used to address the WSL files from outside the container, such as `\\wsl$\Ubuntu-18.04\home\bilge` in this case.

Of course, we could just use Composer from _inside_ the container, but we may have very good reasons for not doing so, such as having our Composer cache and configuration outside the container, plus some commands such as `browse` will simply never work within the container either.

Seldaek:
Fixes #9858

### DIFF
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -559,7 +559,7 @@ class Filesystem
         }
 
         if (is_file($path)) {
-            return false !== Silencer::call('file_get_contents', $path);
+            return false !== Silencer::call('file_get_contents', $path, false, null, 0, 1);
         }
 
         if (is_dir($path)) {
--- a/src/Composer/Util/Filesystem.php
+++ b/src/Composer/Util/Filesystem.php
@@ -543,6 +543,33 @@ class Filesystem
         return preg_replace('{^file://}i', '', $path);
     }
 
+    /**
+     * Cross-platform safe version of is_readable()
+     *
+     * This will also check for readability by reading the file as is_readable can not be trusted on network-mounts
+     * and \\wsl$ paths. See https://github.com/composer/composer/issues/8231 and https://bugs.php.net/bug.php?id=68926
+     *
+     * @param string $path
+     * @return bool
+     */
+    public static function isReadable($path)
+    {
+        if (is_readable($path)) {
+            return true;
+        }
+
+        if (is_file($path)) {
+            return false !== Silencer::call('file_get_contents', $path);
+        }
+
+        if (is_dir($path)) {
+            return false !== Silencer::call('opendir', $path);
+        }
+
+        // assume false otherwise
+        return false;
+    }
+
     protected function directorySize($directory)
     {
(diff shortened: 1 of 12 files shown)


src/Composer/Util/Filesystem.php
```diff
@@ -543,6 +543,33 @@ public static function getPlatformPath($path)
...
+     * @return bool
+     */
+    public static function isReadable($path)
+    {
+        if (is_readable($path)) {
+            return true;
+        }
+
+        if (is_file($path)) {
+            return false !== Silencer::call('file_get_contents', $path);
```
mvorisek, line 562:
```suggestion
            return false !== Silencer::call('file_get_contents', $path, false, null, 0, 1);
```

quite important for huge files

src/Composer/Util/Filesystem.php
```diff
@@ -543,6 +543,33 @@ public static function getPlatformPath($path)
...
+     *
+     * This will also check for readability by reading the file as is_readable can not be trusted on network-mounts
+     * and \\wsl$ paths. See https://github.com/composer/composer/issues/8231 and https://bugs.php.net/bug.php?id=68926
+     *
+     * @param string $path
+     * @return bool
+     */
+    public static function isReadable($path)
+    {
+        if (is_readable($path)) {
```
mvorisek, line 557: I wonder if the later usecases are covered.

src/Composer/Util/Filesystem.php
Seldaek: Good point, thanks :)

src/Composer/Util/Filesystem.php
Seldaek: I'm not sure I understand what you're saying here.

Switching to this isn't needed IMO as if something is a file, it's not a dir.. and vice versa. The only other cases might be links or junctions and other such esoterica, but for those IMO it's fine to just fallback to assume they're not readable. UNC paths are rare enough a use case already, then combining them with junctions + hitting that on one of the few places we are actually using `isReadable` is just super low chance of being a problem. Plus I wouldn't know how to fix these anyway..

mvorisek: My point is that `return true;` will be reached if and only if the usecase is reached, as I think we use line coverage than branch coverage.

Seldaek: We don't do code coverage.

PR #9873, opened 2021-05-10
Add Platform-PHP version to user-agent header to improve targetted-php stats

Seldaek:
Just an idea.. not sure what others think, but this would let us collect info on what the actual php-deploy-target version of the project is, vs using what the developers are running.

For composer this results in something like `User-Agent: Composer/2.0.999-dev+source (Linux; 4.19.104-microsoft-standard; PHP 8.0.3; cURL 7.68.0; Platform-PHP 5.3.9)` - where we see although I'm using php 8, my stats for this should actually count as 5.3.9 because it's what we "deploy" to.

I am not sure if the two versions differ often enough to be important/relevant/interesting though.


naderman:
Sounds reasonable to me, could improve the quality of the PHP version guidance we give developers through your version blog posts. At the same time it doesn't really expose anything more sensitive than what's currently there. Only question is how commonly this really differs from the php version used, but I think that'd be worth finding out.

stof (approved):

reindert-vetter:
This creates a security risk.

reindert-vetter (approved):

PR #9875, opened 2021-05-11
Speedup extracting on Windows

nicolas-grekas:
This is a quick and dirty PRto share an experiment I've been doing on Windows to speedup extracting ZIP files.

### DIFF
--- a/src/Composer/Downloader/ZipDownloader.php
+++ b/src/Composer/Downloader/ZipDownloader.php
@@ -25,9 +25,8 @@ use ZipArchive;
  */
 class ZipDownloader extends ArchiveDownloader
 {
-    protected static $hasSystemUnzip;
     private static $unzipCommands;
-    private static $hasZipArchive = false;
+    private static $hasZipArchive;
     private static $isWindows;
 
     /** @var ZipArchive|null */
@@ -42,10 +41,10 @@ class ZipDownloader extends ArchiveDownloader
             self::$unzipCommands = array();
             $finder = new ExecutableFinder;
             if (Platform::isWindows() && ($cmd = $finder->find('7z', null, array('C:\Program Files\7-Zip')))) {
-                self::$unzipCommands[] = ProcessExecutor::escape($cmd).' x -y %s -o%s';
+                self::$unzipCommands[] = array('7z', ProcessExecutor::escape($cmd).' x -bb0 -y %s -o%s');
             }
             if ($cmd = $finder->find('unzip')) {
-                self::$unzipCommands[] = ProcessExecutor::escape($cmd).' %s -d %s';
+                self::$unzipCommands[] = array('unzip', ProcessExecutor::escape($cmd).' -qq %s -d %s');
             }
         }
 
@@ -56,7 +55,7 @@ class ZipDownloader extends ArchiveDownloader
         if (!self::$hasZipArchive && !self::$unzipCommands) {
             // php.ini path is added to the error message to help users find the correct file
             $iniMessage = IniHelper::getMessage();
-            $error = "The zip extension and unzip command are both missing, skipping.\n" . $iniMessage;
+            $error = "The zip extension and unzip/7z commands are both missing, skipping.\n" . $iniMessage;
 
             throw new \RuntimeException($error);
         }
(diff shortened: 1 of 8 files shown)


src/Composer/Downloader/ZipDownloader.php
```diff
@@ -36,16 +37,22 @@ class ZipDownloader extends ArchiveDownloader
      */
     public function download(PackageInterface $package, $path, PackageInterface $prevPackage = null, $output = true)
     {
-        if (null === self::$hasSystemUnzip) {
+        if (null === self::$unzipCommands) {
+            self::$unzipCommands = array();
             $finder = new ExecutableFinder;
-            self::$hasSystemUnzip = (bool) $finder->find('unzip');
+            if (Platform::isWindows() && ($cmd = $finder->find('7z', null, array('C:\Program Files\7-Zip')))) {
```
nicolas-grekas, line 43: 7z is way faster than unzip on Windows

src/Composer/Util/Filesystem.php
```diff
@@ -116,7 +116,7 @@ public function removeDirectory($directory)
             throw new \RuntimeException('Aborting an attempted deletion of '.$directory.', this was probably not intended, if it is a real use case please report it.');
         }
 
-        if (!\function_exists('proc_open')) {
+        if (1||!\function_exists('proc_open')) {
```
nicolas-grekas, line 119: `removeDirectoryPhp` is way faster than calling `rmdir` on Windows

src/Composer/Downloader/ZipDownloader.php
```diff
@@ -222,7 +233,7 @@ public function extract(PackageInterface $package, $file, $path)
     {
         // Each extract calls its alternative if not available or fails
         if (self::$isWindows) {
-            return $this->extractWithZipArchive($package, $file, $path, false);
+            //return $this->extractWithZipArchive($package, $file, $path, false);
```
nicolas-grekas, line 236: async works on my Windows VM, not sure why it's skipped?

nicolas-grekas (commented):
Installing the deps of Symfony goes from 49s to 14s with this patch.
For thoughts :)

src/Composer/Downloader/ZipDownloader.php
mvorisek: and what is the time using native php?

nicolas-grekas: 49s is the time using native php. It's slow because it cannot be parallelized.

mvorisek: Does `composer` uses `zip` ext, ie. https://www.php.net/manual/en/ziparchive.extractto.php or some native php solution?

If it uses `ZipArchive::extractTo` and it is slow, the performance should be fixed in php-src.

src/Composer/Util/Filesystem.php
mvorisek: Yes, we should not call a process until really needed, Windows process needs usually about 10 ms to start. Probably drop the rmdir/rm completely. (or is rm much better on linux?)

src/Composer/Downloader/ZipDownloader.php
mvorisek: async ?

src/Composer/Util/Filesystem.php
nicolas-grekas: If there's a reason for using rmdir, maybe it could be parallelized btw?

src/Composer/Downloader/ZipDownloader.php
staabm: Maybe the same problem as https://github.com/staabm/php-phardata-benchmark

mvorisek: The php bug 80399 is solely for writing. Did you noticed any perfomrance issues for extracting as well?

src/Composer/Downloader/ZipDownloader.php
nicolas-grekas: Please read the code. Async yes, since several processes are run concurrently on other OSes.

src/Composer/Util/Filesystem.php
Seldaek: Reverted this but fixed the performance issue in https://github.com/composer/composer/pull/9875/commits/28c7d0ea5d4477d3f74236c72d9548fa9760c2f1 - this seems to be enough to remove usage of removeDirectory() under normal circumstances. Plus there was another PR recently adding removeDirectoryAsync for a few use cases, it's just not really worth using everywhere given the added complexity of promises.

Seldaek:
Thanks for getting the ball rolling here @nicolas-grekas, definitely a good point, I think async wasn't enabled on default on windows because in 1.x it was usually not worth it to try the `unzip` code path as it's rarely available, and didn't bring so much benefit as there was no async. Now in 2.x it makes a lot of sense to run it async if possible, and it's a good idea to add 7z, that's easy to install to get a perf boost.

Anyway I cleaned things up, see the commits above.

nicolas-grekas:
Thanks for taking over, LGTM!

PR #9877, opened 2021-05-11
Fix EOL of text files

mvorisek:


stof:
changing json fixture files looks wrong to me

mvorisek:
> 
> 
> changing json fixture files looks wrong to me

CI passes, end EOL there is fine, better to add end EOL so there are less warnings in IDEs.

PR #9878, opened 2021-05-11
Improve CS - native constant/function invocation

mvorisek:
speedup composer


Seldaek:
Thanks but I've already done this in files which matter/are hot code paths. This just serves no purpose except clobbering the code with backslashes, not quite what I would describe as an improvement to CS.

mvorisek:
@Seldaek `@compiler_optimized` functions only, this is to keep everything optimized (I mean checked by CI)

PR #9883, opened 2021-05-15
Make a note about lower casing of package names

BackEndTea:
If i understand correctly, package names are required to be lower case since composer 2.0, so these docs were no longer up to date.


Seldaek:
Thanks

PR #9887, opened 2021-05-17
Use a simpler suggested require version of * to keep things simple for extensions which are versioned like PHP

### LINKED ISSUE #9483: Ext require should be added with "*" by default
Example:

`composer require ext-bcmath` should add `require` like:

```
    "require": {
        "ext-php": "*"
    }
```

as php version is already determined by php constraints. This is also how php extensions constraint is defined in almost every package.

Currently, it is added like:

```
    "require": {
        "ext-php": "^7.4"
    }
```

which only complicates version requirements without any extra value.

Seldaek:
Fixes #9483

/cc @mvorisek

### DIFF
--- a/src/Composer/Package/Version/VersionSelector.php
+++ b/src/Composer/Package/Version/VersionSelector.php
@@ -153,6 +153,16 @@ class VersionSelector
      */
     public function findRecommendedRequireVersion(PackageInterface $package)
     {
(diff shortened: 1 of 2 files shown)


mvorisek (approved):
LGTM

naderman (approved):

PR #9897, opened 2021-05-20
Fix schema minimum-stability pattern

guilliamxavier:


tests/Composer/Test/Json/ComposerSchemaTest.php
```diff
@@ -71,25 +71,23 @@ public function testRequireTypes()
...
-                'property' => 'minimum-stability',
-                'message' => 'Does not match the regex pattern ^dev|alpha|beta|rc|RC|stable$',
-                'constraint' => 'pattern',
-                'pattern' => '^dev|alpha|beta|rc|RC|stable$',
-            ),
-        ), $this->check($json), 'dummy');
+        $this->assertEquals($expectedError, $this->check($json), 'dummy');
+
+        $json = '{ "name": "vendor/package", "description": "generic description", "minimum-stability": "devz" }';
+        $this->assertEquals($expectedError, $this->check($json), 'devz');
```
guilliamxavier, line 90: Added test, was failing before the fix (i.e. "devz" was incorrectly accepted)

res/composer-schema.json
```diff
@@ -397,7 +397,7 @@
         "minimum-stability": {
             "type": ["string"],
             "description": "The minimum stability the packages must have to be install-able. Possible values are: dev, alpha, beta, RC, stable.",
-            "pattern": "^dev|alpha|beta|rc|RC|stable$"
+            "enum": ["dev", "alpha", "beta", "rc", "RC", "stable"]
```
guilliamxavier, line 400: The pattern was missing parentheses `"^(dev|alpha|beta|rc|RC|stable)$"`, but an enum seemed more appropriate

res/composer-schema.json
```diff
@@ -42,7 +42,7 @@
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
-            "pattern": "^v?\\d+(((\\.\\d+)?\\.\\d+)?\\.\\d+)?|^dev-"
+            "pattern": "^v?\\d+(\\.\\d+){0,3}|^dev-"
```
guilliamxavier, line 45: First reordered to `"^v?\\d+(\\.\\d+(\\.\\d+(\\.\\d+)?)?)?|^dev-"` to avoid needless backtracking, then simplified (actually it could be simplified even more to just `"^v?\\d+|^dev-"`, but I guess the pattern can also serve as some kind of documentation)

stof, line 45: This change does not seem related to fixing the min stability pattern

res/composer-schema.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "https://json-schema.org/draft-04/schema#",
-    "name": "Package",
+    "title": "Package",
```
guilliamxavier, line 3: Cf [SchemaStore/schemastore@`02607b1`](https://github.com/SchemaStore/schemastore/pull/1432/commits/02607b1dce97be80d6ec8012863b8bc1f87e5ea8)

.github/CONTRIBUTING.md
```diff
@@ -41,7 +41,7 @@ To achieve this, you need to acquire the Composer source code:
 2. Download the [`composer.phar`](https://getcomposer.org/composer.phar) executable
 3. Run Composer to get the dependencies: `cd composer && php ../composer.phar install`
 
-You can run the test suite by executing `vendor/bin/phpunit` when inside the
+You can run the test suite by executing `vendor/bin/simple-phpunit` when inside the
```
guilliamxavier, line 44: When I wanted to test: "bash: vendor/bin/phpunit: No such file or directory"

guilliamxavier (commented):
Some explanations:

res/composer-schema.json
guilliamxavier: Indeed, but it fixes unnecessary repeated backtracking (for e.g. "1.2" or "1.2.3") that I noticed while reviewing other occurrences of "pattern". There are other small additional fixes. I have added some comments

Seldaek: `\d` doesn't include `.` so that even simpler version would not work IMO, but the current version is a good improvement. 👍🏻

res/composer-schema.json
Seldaek: FYI `()` are not needed for `|` to work, `()` merely creates a new capturing group, but there is no capture needed here.

But enum definitely is clearer :)

Seldaek:
Merged as https://github.com/composer/composer/commit/8d8842eb8cd72807d22dbd98dedb5eaec4e7f6a7 in 2.0/master, thanks

res/composer-schema.json
guilliamxavier: Yes it would work (I have checked it), the same as `^ab*` is equivalent to just `^a` for validation purposes (i.e. not using the matches), but I too prefer it like this 😉

res/composer-schema.json
guilliamxavier: There are needed here (as checked by the new test, which was failing without them, and passing with them) because they have a lower "precedence" than the anchors `^`/`$`, so that was testing "`^dev` or `alpha` or ... or `stable$`" instead of intended "`^dev$` or `^alpha$` or ... or `^stable$`".

But glad you agreed for enum 🙂

Seldaek: Ah right right with the anchors it makes sense, sorry about that :)

res/composer-schema.json
Seldaek: Oh right there is no $ here, it's just checking the prefix.. Clearly I am not paying enough attention :)

stof: A valid version can only have `-`, `+` or the end of the string after those numbers through. So we should be able to actually validate the whole version part.

guilliamxavier: Thanks, please see #9903

res/composer-schema.json
guilliamxavier: Oh my... I just tried a `composer validate` with `"minimum-stability": "rc"` (_lowercase_ "RC", which was in the pattern and I kept for the enum), and (after successful schema validation) that's actually an error:
```
./composer.json is invalid, the following errors/warnings were found:
minimum-stability : invalid value (rc), must be one of stable, RC, beta, alpha, dev
```
So the pattern had two bugs actually, and I only fixed one 🤦‍♂️ I'll submit a second PR soon

PR #9898, opened 2021-05-20
macOS cache dir

el7cosmos:
Other popular CLI apps that put caches in `~/Library/Caches` in macOS:
- [helm](https://helm.sh/)
- [heroku](https://devcenter.heroku.com/articles/heroku-cli)
- [homebrew](https://brew.sh/)
- [pip](https://pypi.org/project/pip/)
- [vercel](https://vercel.com/cli)
- [yarn](https://yarnpkg.com/)

Those apps may also leverage `~/Library/Application Support` or `~/Library/Preferences` which may be a place for `$COMPOSER_HOME`

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -198,9 +198,10 @@ versions. See also [COMPOSER_HOME](03-cli.md#composer-home).
 ## cache-dir
 
 Defaults to `C:\Users\<user>\AppData\Local\Composer` on Windows,
-`$XDG_CACHE_HOME/composer` on unix systems that follow the XDG Base Directory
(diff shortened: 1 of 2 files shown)


Seldaek:
From my (non-macOS-user) understanding, this is a good change and it's good to follow the OS guidelines. 

I am not entirely sure what macOS users would expect though from a CLI app like Composer, so still happy to hear if anyone thinks this is a bad idea.

Seldaek:
Thanks!

jk:
Putting it in ~/Library/Caches has the added benefit, that the dir is excluded from Time Machine Backups by default (the OS integrated backup solution on macOS€

Issue #9842, opened 2021-04-22
Validate command doesn't detect missing packages anymore, neither does 'update nothing'/'update --lock'


PrinsFrank:
Since Composer 2.0.0 (still present in 2.0.12), packages that are present in the ```composer.json``` but are not locked in the ```composer.lock``` or installed in the vendor folder are not correctly detected as an out of sync lock file, or resolved when running ```composer update nothing``` or ```composer update --lock```. In composer 1.10.21 and before, this behaviour was correct.

### Reproduction
1. Create a new folder with this composer.json:
```json
{
    "name": "reproduction/reproduction",
    "description": "reproduction",
    "license": "MIT",
    "require": {
        "php": "^7.4"
    }
}
```
2. Run ```composer update``` to generate an up to date lock file.
3. Add an extra dependency manually in the composer.json file (For example a dependency to ```"composer/semver": "^3.0"```)
> This is not the correct way to add a dependency, but the composer.json and composer.lock might get out of sync in invalid merges, or editors like phpstorm suggest edits to the composer.json without updating the lock file etc.
4. Run ```composer validate```. An error about an out of date lock file is displayed:
```The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update` or `composer update <package name>`.```
5. Run ```composer u nothing``` or ```composer u --lock```
6. Run ```composer validate```. The warning is now gone implicating the lock file is up to date. The lock file is not up to date however as dependencies from the composer.json are not in the lock file or installed into the vendor folder.

### Expected behaviour
1. When a package is present in the ```composer.json``` file but not locked in the ```composer.lock``` file a warning is displayed when running ```composer validate```.
2. When ```composer update nothing``` or ```composer update --lock``` is ran, the package is installed and added to the lock file (How this previously worked in v1.*) or a warning is displayed to not manually add to the ```composer.json``` file with a message to run ```composer require vendor/package-name``` to require the package correctly and fix the discrepancy between the two files.

stof:
@Seldaek this looks like a bug of `composer update --lock`. It should not only change the content-hash if the rest does not match

PrinsFrank:
I've just looked moro thoroughly trough the code for anything that might give this functionality

PrinsFrank:
I've just looked more thoroughly trough the code for anything that might give this functionality and couldn't find where this should be present. Maybe this was lost during the move to composer 2? If anyone can point me to a place I should look at I'm willing to contribute.

[...]

PR #9899, opened 2021-05-20
Detect missing packages that are a requirement or dev-requirement but not locked

PrinsFrank:
Implements #9842 

I tried writing contained code that works in 5.3. I'm not sure if this should live in it's own class and if there is a way to UT this, please let me know if you want anything changed!

### DIFF
--- a/src/Composer/Command/ValidateCommand.php
+++ b/src/Composer/Command/ValidateCommand.php
@@ -100,8 +100,8 @@ EOT
         }
 
         $lockData = $locker->getLockData();
-        $lockPackageNames = array_map(static function($lockPackage) {return $lockPackage['name'];}, $lockData['packages']);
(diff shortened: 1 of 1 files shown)


mvorisek:
Hmm, what about merging `composer.json` rules with `composer.lock` and testing if that rules resolves to the same `composer.lock`?

a) It is quick
b) any difference is either an error in the lock file or `composer.json` represent unresolveable contraints (when merged with lock file) or was extended (and that dependencies are missing in the lock file)

Seldaek:
Thanks @PrinsFrank - the implementation wasn't bad IMO but it missed a few corner cases and I had to implement it quite differently to fix them, see https://github.com/composer/composer/commit/2eaf81c8138e76393e0aa89d345dde70b72d75d7 if you're interested :)

Seldaek:
@mvorisek check my commit, running the solver isn't necessary here.

PR #9347, opened 2020-10-26
Docs: Fix sentence after reformulation

guilliamxavier:

PR #9900, opened 2021-05-21
Update 06-config.md

trevorpan:
From an error I got, this needs to be plural.

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -171,7 +171,7 @@ Therefore if you use this it is recommended, and safer, to also run the
 deployment strategy.
 
 If a dependency requires some extension that you do not have installed locally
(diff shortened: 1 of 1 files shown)


guilliamxavier:
Hi, what version are you using? From what I found:
- Composer 1 only supported plural `--ignore-platform-reqs` (without value)
- Composer 2 still supports plural `--ignore-platform-reqs` (without value) and also a new singular `--ignore-platform-req=ext-foo` (with value, multiple occurrences possible)

The problem (that I had already pointed out in https://github.com/composer/composer/pull/9347#issuecomment-716472193, just after the release of Composer 2.0.0) is that there's only one version of the online docs (for the latest version of Composer) 😕

PR #9902, opened 2021-05-21
Detect and output a better hint for cyclic dependencies

### LINKED ISSUE #9837: DX: Separate error message for cyclic dependencies
I get asked from time to time from less experienced composer users why they have errors like

grafik

the actual problem is, that they have defined a cyclic dependency without realizing it.
would it make sense/feasable to use/print a error/warning message which explicitly states that?

cyclic dependencies lead to a lot of problems, and leaving less experienced people with a error message which they need to interpret and realize that they actually build a circle is not helpful.

maybe its not possible or hard to identify - I have no idea.. just leaving the idea here, to gather some feedback.

Seldaek:
Fixes #9837

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -47,32 +47,39 @@ This is a list of common pitfalls on using Composer, and how to avoid them.
    In this case add the `--with-dependencies` argument **or** add all dependencies which
    need an update to the command.
 
-## Package not found on travis-ci.org
+## Dependencies on the root package
 
-1. Check the ["Package not found"](#package-not-found) item above.
+When your root package depends on a package which ends up depending (directly or
+indirectly) back on the root package itself, issues can occur in two cases:
 
-2. If the package tested is a dependency of one of its dependencies (cyclic
-   dependency), the problem might be that Composer is not able to detect the version
-   of the package properly. If it is a git clone it is generally alright and Composer
-   will detect the version of the current branch, but travis does shallow clones so
-   that process can fail when testing pull requests and feature branches in general.
+1. During development, if you are on a branch like `dev-main` and the branch has no
+   [branch-alias](aliases.md#branch-alias) defined, and the dependency on the root package
+   requires version `^2.0` for example, the `dev-main` version will not satisfy it.
+   The best solution here is to make sure you first define a branch alias.
+
+2. In CI (Continuous Integration) runs, the problem might be that Composer is not able
+   to detect the version of the root package properly. If it is a git clone it is
+   generally alright and Composer will detect the version of the current branch,
+   but some CIs do shallow clones so that process can fail when testing pull requests
+   and feature branches. In these cases the branch alias may then not be recognized.
    The best solution is to define the version you are on via an environment variable
-   called COMPOSER_ROOT_VERSION. You set it to `dev-master` for example to define
-   the root package's version as `dev-master`.
-   Use: `before_script: COMPOSER_ROOT_VERSION=dev-master composer install` to export
-   the variable for the call to composer.
+   called COMPOSER_ROOT_VERSION. You set it to `dev-main` for example to define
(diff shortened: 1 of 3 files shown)


doc/articles/troubleshooting.md
```diff
@@ -47,32 +47,45 @@ This is a list of common pitfalls on using Composer, and how to avoid them.
    In this case add the `--with-dependencies` argument **or** add all dependencies which
    need an update to the command.
 
-## Package not found on travis-ci.org
+## Cyclic / circular dependencies
```
naderman, line 50: Cycling or circular dependencies are a normal thing and don't cause problems. What this describes is rather a Cyclic / circular dependency including the root package, so I think the title too should make it clear this is only about situations involving the root package.

doc/articles/troubleshooting.md
```diff
@@ -47,32 +47,45 @@ This is a list of common pitfalls on using Composer, and how to avoid them.
...
 
-1. Check the ["Package not found"](#package-not-found) item above.
+Cyclic (or circular) dependency occurs when your root package depends on a package
+which ends up depending (directly or indirectly) back on the root package itself.
+
+The best is to avoid this really if possible as it can lead to problems, but if you
+absolutely must have a cyclic dependency read on to figure out solutions.
+
+Cyclic dependencies commonly causes issues in two cases:
```
naderman, line 58: Here to you need to be clear, cyclic dependencies do not typically cause any of these problems, a cyclic dependency on the root package may.

src/Composer/DependencyResolver/Problem.php
```diff
@@ -420,6 +421,16 @@ private static function computeCheckForLowerPrioRepo($isVerbose, $packageName, $
             }
         }
 
+        if ($higherRepoPackages) {
+            $topPackage = reset($higherRepoPackages);
+            if ($topPackage instanceof RootPackageInterface) {
+                return array(
+                    "- Root composer.json requires $packageName".self::constraintToText($constraint).', it is ',
+                    'satisfiable by '.self::getPackageList($nextRepoPackages, $isVerbose).' from '.$nextRepo->getRepoName().' but the root package can not be updated. This is due to a cyclic dependency. See https://getcomposer.org/cyclic-dep for details and assistance.'
```
naderman, line 429: This is confusing, what is due to a cyclic dependency? The root package cannot be updated?

naderman (changes requested):
I actually think this should entirely be rewritten to put less of a focus on cyclic/circular dependencies and instead make it clear that a dependency on something which depends on a specific root package version is the problem.

tests/Composer/Test/Fixtures/installer/circular-dependency-errors.test
```diff
@@ -0,0 +1,45 @@
...
+2
+
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
+    - Root composer.json requires requires/root 1.0.0 -> satisfiable by requires/root[1.0.0].
+    - requires/root 1.0.0 requires root/pkg ^1.0 -> satisfiable by root/pkg[1.0.0] from package repo (defining 2 packages) but the root package can not be updated so it is not installable. See https://getcomposer.org/cyclic-dep for details and assistance.
```
naderman, line 43:
```suggestion
    - requires/root 1.0.0 requires root/pkg ^1.0 -> satisfiable by root/pkg[1.0.0] from package repo (defining 2 packages) but root/pkg is the root package and cannot be modified. See https://getcomposer.org/cyclic-dep for details and assistance.
```

naderman (approved):

Seldaek:
Added https://github.com/composer/composer/commit/0dce0f80f12543304fb0f1f755fb7bf9dbecfc9e to this :)

PR #9904, opened 2021-05-21
Remove lowercase "rc" from schema minimum-stability enum

guilliamxavier:
Refs https://github.com/composer/composer/pull/9897#discussion_r636874795

### DIFF
--- a/tests/Composer/Test/Json/ComposerSchemaTest.php
+++ b/tests/Composer/Test/Json/ComposerSchemaTest.php
@@ -94,9 +94,9 @@ class ComposerSchemaTest extends TestCase
         $expectedError = array(
             array(
                 'property' => 'minimum-stability',
-                'message' => 'Does not have a value in the enumeration ["dev","alpha","beta","rc","RC","stable"]',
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Json/ComposerSchemaTest.php
```diff
@@ -119,7 +119,7 @@ public function testMinimumStabilityValues()
         $this->assertTrue($this->check($json), 'beta');
 
         $json = '{ "name": "vendor/package", "description": "generic description", "minimum-stability": "rc" }';
-        $this->assertTrue($this->check($json), 'rc lowercase');
+        $this->assertEquals($expectedError, $this->check($json), 'rc lowercase');
```
guilliamxavier, line 122: I had to change this test (from OK to error), not sure if `composer validate` was accepting it back in 2015...

Seldaek:
I investigated this and `rc` is actually accepted at runtime, only validate was complaining (wrongly) about it. So the schema is fine 👍🏻

Issue #9908, opened 2021-05-22
Composer does not use headers provided as part of the repository definition for packages


samlitowitz:
My `composer.json`:

```json
{
    "name": "composer/composer",
    "type": "library",
    "description": "Composer helps you declare, manage and install dependencies of PHP projects. It ensures you have the right stack everywhere.",
    "keywords": [
        "package",
        "dependency",
        "autoload"
    ],
    "homepage": "https://getcomposer.org/",
    "license": "MIT",
    "authors": [
        {
            "name": "Nils Adermann",
            "email": "naderman@naderman.de",
            "homepage": "https://www.naderman.de"
        },
        {
            "name": "Jordi Boggiano",
            "email": "j.boggiano@seld.be",
            "homepage": "https://seld.be"
        }
    ],
    "require": {
        "php": "^5.3.2 || ^7.0 || ^8.0",
        "composer/ca-bundle": "^1.0",
        "composer/metadata-minifier": "^1.0",
        "composer/semver": "^3.0",
        "composer/spdx-licenses": "^1.2",
        "composer/xdebug-handler": "^2.0",
        "justinrainbow/json-schema": "^5.2.10",
        "psr/log": "^1.0",
        "seld/jsonlint": "^1.4",
        "seld/phar-utils": "^1.0",
        "symfony/console": "^2.8.52 || ^3.4.35 || ^4.4 || ^5.0 || ^6.0",
        "symfony/filesystem": "^2.8.52 || ^3.4.35 || ^4.4 || ^5.0 || ^6.0",
        "symfony/finder": "^2.8.52 || ^3.4.35 || ^4.4 || ^5.0 || ^6.0",
        "symfony/process": "^2.8.52 || ^3.4.35 || ^4.4 || ^5.0 || ^6.0",
        "react/promise": "^1.2 || ^2.7",
        "sglitowitzsoci/private-test": "dev-v1.0.0"
    },
    "require-dev": {
        "symfony/phpunit-bridge": "^4.2 || ^5.0 || ^6.0",
        "phpspec/prophecy": "^1.10"
    },
    "suggest": {
        "ext-openssl": "Enabling the openssl extension allows you to access https URLs for repositories and packages",
        "ext-zip": "Enabling the zip extension allows you to unzip archives",
        "ext-zlib": "Allow gzip compression of HTTP requests"
    },
    "config": {
        "platform": {
            "php": "5.3.9"
        },
        "platform-check": false
    },
    "extra": {
        "branch-alias": {
[...]

PR #9909, opened 2021-05-22
Initial commit, issue-9908

samlitowitz:
Addresses #9908 

Composer will use headers provided as part of repository package definintions.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -20,6 +20,7 @@ use Composer\Util\StreamContextFactory;
 use Composer\Util\AuthHelper;
 use Composer\Util\Url;
 use Composer\Util\HttpDownloader;
+use Composer\Util\Http\HeaderHelper;
 use React\Promise\Promise;
 
 /**
@@ -29,7 +30,7 @@ use React\Promise\Promise;
  */
 class CurlDownloader
 {
-    private $multiHandle;
+    public $multiHandle;
     private $shareHandle;
     private $jobs = array();
     /** @var IOInterface */
@@ -38,6 +39,8 @@ class CurlDownloader
     private $config;
     /** @var AuthHelper */
     private $authHelper;
+    /** @var HeaderHelper */
+    private $headerHelper;
     private $selectTimeout = 5.0;
     private $maxRedirects = 20;
     /** @var ProxyManager */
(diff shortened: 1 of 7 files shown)


src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -29,7 +30,7 @@
  */
 class CurlDownloader
 {
-    private $multiHandle;
+    public $multiHandle;
```
stof, line 33: why making this public ?

src/Composer/Util/Http/HeaderHelper.php
```diff
@@ -0,0 +1,92 @@
...
+
+        foreach ($repos as $repo) {
+            if ($repo['type'] !== 'package') {
+                continue;
+            }
+            if (!array_key_exists('package', $repo)) {
+                continue;
+            }
+            if (
+                array_key_exists('dist', $repo['package'])
```
stof, line 71: a `package` repository can provide multiple packages

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -184,6 +189,7 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
 
         $options['http']['header'] = $this->authHelper->addAuthenticationHeader($options['http']['header'], $origin, $url);
         $options = StreamContextFactory::initOptions($url, $options, true);
+        $options['http']['header'] = $this->headerHelper->addHeader($options['http']['header'], $origin, $url);
```
stof, line 192: This implementation looks weird to me:

- it works only for the CurlDownloader
- it goes reading the configuration from repositories of type `package` (why only those ?) to alter headers even for URLs which are not related to that repository (well, they have to match).

to me, it looks like this has been implemented in the wrong layer of composer.

src/Composer/Util/Http/HeaderHelper.php
samlitowitz: A complication for sure, probably address this after the discussion where to implement this if it remains relevant.

src/Composer/Util/Http/CurlDownloader.php
samlitowitz: Unintentional.

src/Composer/Util/Http/CurlDownloader.php
samlitowitz: I am unfamiliar with the composer architecture and welcome suggestions. Only applying it to `package` types does appear to be unnecessarily limiting.

PR #9910, opened 2021-05-23
Add Docker information

Pen-y-Fan:
Basic docker information added.  Given Docker is one of the key development environments I think it is worth adding some basic information.

I know this isn't officially supported, so I have added the details on where issues should be posted.

I hope this is agreeable.

### DIFF
--- a/doc/00-intro.md
+++ b/doc/00-intro.md
@@ -160,7 +160,9 @@ Composer version 2.0.12 2021-04-01 10:14:59
 
 ## Docker Image
 
-Composer has a **community supported** image available on [docker hub](https://hub.docker.com/r/composer/composer).
+Composer is published as Docker container in a few places, see the list in the [composer/docker README](https://github.com/composer/docker).
+
+Example usage:
 
 ```sh
 docker pull composer/composer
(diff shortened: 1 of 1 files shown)


doc/00-intro.md
```diff
@@ -158,6 +158,27 @@ C:\Users\username>composer -V
 Composer version 2.0.12 2021-04-01 10:14:59
 ```
 
+## Docker Image
+
+Composer has a **community supported** image available on [docker hub](https://hub.docker.com/_/composer).
```
Seldaek, line 163: IMO we should rather recommending using https://hub.docker.com/r/composer/composer and composer/composer below, as this is built by us and may be fully automated at some point. The https://hub.docker.com/_/composer image relies on someone at docker merging our PR which is sometimes fairly delayed.

stof:
This currently documents 2 ways to use that docker image:
- copying composer into your own image
- running composer directly from a container using the composer image.

The second way (which actually comes first in your doc right now) has a huge drawback: this runs composer against the PHP runtime of the composer image rather than against your own one. And so platform dependencies are not resolved against the right one, which is a very good way to shoot yourselves in the foot.
To use composer reliably to resolve versions, copying it to your own PHP image (using the same PHP version and extensions that what runs your app) is a must-have.

doc/00-intro.md
alcohol: Or document both and explain the difference/benefit of using the vendor image? 

@Pen-y-Fan see https://github.com/composer/docker/blob/master/README.md for differences between the images.

alcohol:
Perhaps better to merely mention the existence and link directly to source repo / documentation for further details. That way, if anything ever changes, it only has to be updated in one place.

doc/00-intro.md
stof: well, the difference documented there is precisely the one given by @Seldaek: the delay is shorter because there are less manual steps (and the remaining manual step is done by the composer team, not by an external team)

doc/00-intro.md
```diff
@@ -158,6 +158,30 @@ C:\Users\username>composer -V
...
+## Docker Image
+
+Composer is published as Docker container in a few places, see the list in the [composer/docker README](https://github.com/composer/docker).
+
+Example usage:
+
+```sh
+docker pull composer/composer
+docker run --rm -it -v "$(pwd):/app" composer/composer -V
+Composer version 2.0.13 2021-04-27 13:11:08
```
alcohol, line 170:
```suggestion
docker run --rm -it -v "$(pwd):/app" composer/composer install
```

Seldaek:
Alright finally wrapped this up, thanks everyone

PR #9911, opened 2021-05-23
Remove funding info from standard install/update output

mvorisek:
The funding info is taking 2 lines, better to keep the output of composer install/update minimal.


Seldaek:
This isn't useless information, it's bringing awareness that stuff you use for free is looking for funding..

PR #9912, opened 2021-05-24
Invert strict/lax schema validation

guilliamxavier:
Refs https://github.com/composer/composer/pull/9782#issuecomment-844983851:

> We could make the name/description optional in the schema, and change the rules at runtime to make them required instead of the opposite (which is what we do now). I think that probably makes the schema the most re-usable while not changing much for composer users.

This PR basically reverts the recently merged #9782 (and the follow-up commit 458bd41d8fe8df7e3da7dad8cfd09ea4dfc0d3cd) and inverts the original logic.

So name/description become always optional for pure JSON schema validation [even for non-project type], and are still always required at runtime for `composer validate` (strict errors, or warnings with `--no-check-publish`) [even for project type].
Similarly, additional properties become allowed in the schema (but are still forbidden for publication).

I think that makes sense, because you may want to have a private (non-published) custom-type package without name/description and/or with custom properties, or on the contrary, publish a project-type package (e.g. https://packagist.org/packages/symfony/skeleton).

This should also avoid the potentially confusing error output depicted in https://github.com/composer/composer/pull/9782#issuecomment-845159854 (and explained in https://github.com/composer/composer/pull/9782#issuecomment-845217123).

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -94,9 +94,7 @@ Out of the box, Composer supports four types:
   CMSs like the [SilverStripe installer](https://github.com/silverstripe/silverstripe-installer)
   or full fledged applications distributed as packages. This can for example
   be used by IDEs to provide listings of projects to initialize when creating
-  a new workspace.  Setting the type to `project` also makes the `name` and
-  `description` fields optional, making it a good choice for closed source
-  projects wishing to use `composer validate`.
+  a new workspace.
 - **metapackage:** An empty package that contains requirements and will trigger
   their installation, but contains no files and will not write anything to the
   filesystem. As such, it does not require a dist or source key to be
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -2,8 +2,6 @@
     "$schema": "https://json-schema.org/draft-04/schema#",
     "title": "Package",
     "type": "object",
-    "additionalProperties": false,
-    "required": [ "name", "description" ],
     "properties": {
         "name": {
             "type": "string",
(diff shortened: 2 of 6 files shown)


stof:
they should not be warnings for `--no-check-publish`. the name and descriptions are only needed for publication

guilliamxavier:
@stof: They have always been warnings with `--no-check-publish` (and errors without) 😕

stof:
> you may want to have a private (non-published) custom-type package without name/description

custom types are useless for a root package (as the type is meant to allow changing the installer to a non-default one, and the root package is not installed as a dependency).
And if your package is meant to be installed as a dependency, you will still need to publish it to some repository (even if it is a private one) and so you will need a name (the description being required for publication is not a technical requirement but a policy one, as the description is only used for displaying it, not for any feature of composer)

guilliamxavier:
@stof: But as the docs say, name and description are required for *published* packages, not "project-type packages", so (on second thoughts) this PR (my interpretation of @Seldaek's suggestion) may be a "better" solution than #9782?

stof:
@guilliamxavier but published does not mean "published to packagist.org". If you publish it to a private repository (to allow using it in another private project), you also need a name (whether you write a non-empty description is up to you then, and depends on how your private repository looks like and how big it is, to know whether descriptions are useful to find the packages back)

guilliamxavier:
Of course you need a name to *publish* it (be it to public Packagist or private Satis or whatever). The point is that it's actually *not* related to the package *type* (notably, there are project-type published packages)

Seldaek:
Right, I think this may be a better approach indeed.. It produces less confusing output too without the `oneOf`.

Seldaek:
Thanks!

PR #9925, opened 2021-05-31
Add support for mirror-style placeholders for the dist url

### LINKED ISSUE #9661: Support placeholders in dist URLs
composer.json dist URLs should support replacement of placeholders like the version number and normalized version, to allow People to reference external dist files without having to modify the composer.json for each release. Useful if you keep a composer.json in your git repository but need a build process for dist files. There is currently already support for this in mirror dist URLs (%package%, %version%, %reference%) so it should work the same for regular dist URLs.

e.g.

```
            "dist": {
                "type": "zip",
                "url": "https://example.com/custom-builds/%package%/%version%/%reference%.%type%",
                "reference": "916356edb5c30fa564d02de94bde51a9c0bb4469",
                "shasum": "",
                "mirrors": [
                    {
                        "url": "https://repo.packagist.com/packagist-conductors/dists/%package%/%version%/%reference%.%type%",
                        "preferred": true
                    }
                ]
            },

Seldaek:
Fixes #9661

### DIFF
--- a/src/Composer/Package/Package.php
+++ b/src/Composer/Package/Package.php
@@ -625,14 +625,14 @@ class Package extends BasePackage
         }
 
         if ($urlType === 'dist' && false !== strpos($url, '%')) {
-            $url = ComposerMirror::processUrl($url, $this->name, $this->version, $ref, $type);
+            $url = ComposerMirror::processUrl($url, $this->name, $this->version, $ref, $type, $this->prettyVersion);
         }
 
         $urls = array($url);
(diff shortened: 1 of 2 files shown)


stof:
should be documented IMO

naderman:
@glaubinix can you take a look at this too?

glaubinix:
Works like expected. What would be useful are additional replacement properties. Especially the prettyVersion might be more useful with dists as it can for instance be used do download GitHub releases.

PR #9932, opened 2021-06-03
PHP 8.1 | Fix "Implicit conversion from non-compatible float" deprecation notice

### LINKED ISSUE #9931: PHP 8.1: Implicit conversion from non-compatible float
**When I run this command** (or any variation thereof): 

```
composer install
```

**I get the following output**:

```
Deprecation Notice: Implicit conversion from non-compatible float 304.11 to int in phar:///usr/local/bin/composer/src/Composer/Platform/Version.php:100
Deprecation Notice: Implicit conversion from non-compatible float 204.49 to int in phar:///usr/local/bin/composer/src/Composer/Platform/Version.php:100
Updating dependencies
```

**And I expected this to happen**:
No deprecation notices to come from Composer itself.

This issue is unrelated to any specific  setup and can be seen in the CI workflow for this package as well: https://github.com/composer/composer/runs/2728227548?check_suite_focus=true#step:7:9

The issue is related to the following change made in PHP 8.1: https://wiki.php.net/rfc/implicit-float-int-deprecate

jrfnl:
As reported in #9931.

Unfortunately I gave not been able to track down the input which is causing the notices to be thrown, which is why I haven't added a test case to the unit tests. Would be great if someone would known the exact input to use for those.

The issue (and fix) can be verified by using the "quick preview" option and choosing `git.master` with this code sample: https://3v4l.org/1fr2g

Fixes #9931

### DIFF
--- a/src/Composer/Platform/Version.php
+++ b/src/Composer/Platform/Version.php
@@ -97,7 +97,7 @@ class Version
         return sprintf(
             '%d.%d.%d',
             $versionId / ($base * $base),
-            ($versionId / $base) % $base,
+            (int) ($versionId / $base) % $base,
             $versionId % $base
         );
     }


Seldaek:
Thanks 👍🏻

jrfnl:
You're welcome Jordi. Happy to be able to contribute ;-)

stof:
using `floor($version / $base) % $base` might be easier to understand than a int cast

jrfnl:
@stof I see you point and had considered that, however, `floor()` returns a float and while the implicit int cast _should_ be fine with the output of `floor()`, it won't always be and would be susceptible to semi-random deprecation notices popping up.

stof:
@jrfnl implicit conversion of int-compatible floats is *not* deprecated. The explicit cast is actually **less safe**, because it will happily perform unsafe conversion loosing decimal parts. See the output of https://3v4l.org/lFCgr/rfc#output for PHP's master branch.

jrfnl:
@stof I'm aware of that, but due to floating point imprecision and the return type of `floor` being float, that means you may still end up with a return value of `7.99999999999999999999999998`, which will cause the implicit float to int deprecation notice, which is what I meant by "semi-random deprecation notices popping up" - I've seen these type of return values reported in the wild for `ceil()`, so they are bound to exist for `floor()` as well.

stof:
>  I've seen these type of return values reported in the wild for `ceil()`, so they are bound to exist for `floor()` as well.

these look like bugs to me. ceil and floor are meant to return the closest integers. and all integers (up to `2^53-1`) can be represented exactly as a float.

PR #9938, opened 2021-06-03
Fix JsonFile when using custom json schema with no "name" and "descri…

adlacruzes:
Hi,

due to changes in #9912 I am getting the next errors when using a custom json schema and validation is set to `STRICT_SCHEMA`:
- "name : The property name is required"
- "description : The property description is required"

As I understood from the discussions, these required properties are needed for validate composer schema in some scenarios,
but they should not affect custom schema files.

This PR:
 - applies required `name` and `description` only when composer schema is used with `STRICT_SCHEMA`
 - adds tests using custom schema

 @guilliamxavier could you check if I am missing something? Thanks!

### DIFF
--- a/src/Composer/Json/JsonFile.php
+++ b/src/Composer/Json/JsonFile.php
@@ -180,6 +180,7 @@ class JsonFile
     {
         $content = file_get_contents($this->path);
         $data = json_decode($content);
+        $requiredSchemaData = array();
 
         if (null === $data && 'null' !== $content) {
             self::validateSyntax($content, $this->path);
@@ -187,6 +188,7 @@ class JsonFile
 
         if (null === $schemaFile) {
             $schemaFile = __DIR__ . self::COMPOSER_SCHEMA_PATH;
+            $requiredSchemaData = array('name', 'description');
         }
 
         // Prepend with file:// only when not using a special schema already (e.g. in the phar)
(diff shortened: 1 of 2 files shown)


Seldaek:
Oh so you are reusing this code to validate another schema file? I think your fix looks good but will check again tomorrow :)

Seldaek:
So previously what was happening in https://github.com/composer/composer/blob/2.0/src/Composer/Json/JsonFile.php#L199-L202

is that did not touch anything for strict schema, and it removed required/additionalProperties checks for lax schema.

Now your change caused it to override required/additionalProperties for strict schema, and do nothing on lax schema, which isn't great. I fixed that so it now does the same as it used to for custom schemas, but keeps the desired behavior on strict schema for the composer schema.

guilliamxavier:
Thanks for pinging me, and sorry for missing the custom schema case :s

Thanks for the fix, I agree with the final version: remove additionalProperties/required constraints in lax validation for both composer and custom schemas (like it was in 2.0), and add them in strict validation only for composer schema :+1:

adlacruzes:
Thanks, @guilliamxavier!

Thanks, @Seldaek! I use the schema validation in a composer plugin and works perfectly.

The code looks cleaner now!

PR #9942, opened 2021-06-04
Fix --no-scripts disabling events for plugins

nicolas-grekas:
Right now, when composer runs with `--no-scripts`, it not only disables "scripts" entries from the composer.json file, but it also prevents plugins from actually doing anything. This PR fixes it.

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -183,6 +183,7 @@ EOT
         }
 
         $composer = Factory::create($io, null, $disablePlugins);
+        $composer->getEventDispatcher()->setRunScripts(!$noScripts);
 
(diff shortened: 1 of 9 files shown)


nicolas-grekas:
This change basically means that `Installer::setRunScripts()` and `AutoloadGenerator::setRunScripts()` are now useless (and the related logic.) I didn't remove them because I don't know how you want to deal with any BC on the topic.

Seldaek:
I think this is fine as is. Keeping setRunScripts in those places is probably fine, because when Installer is running in dry-run mode, I think suppressing the dispatching of events completely is probably a good idea still to avoid having plugins doing weird stuff while we output operations which aren't being executed. This is done automatically when dry-run is enabled, so I deprecated the setter in Installer.

PR #9960, opened 2021-06-09
Add Composer version in about command

noniagriconomie:
Hi,
I found relevant to add this info in the `composer about` output
Even if it is already present in the `composer -V` as well
Feel free to close :)


Seldaek:
Thanks

PR #9979, opened 2021-06-21
GH Actions: set error_reporting to E_ALL

jrfnl:
### GH Actions: set error reporting to E_ALL

Turns out the default setting for `error_reporting` used by the SetupPHP action is `error_reporting=E_ALL & ~E_DEPRECATED & ~E_STRICT` and `display_errors` is set to `Off`.

For the purposes of CI, I'd recommend running with `E_ALL` and `display_errors=On` to ensure **all** PHP notices are shown.

### GH Actions: ensure linting is done against highest/lowest supported PHP version

`latest` in the matrix will always refer to the latest stable PHP release, which would now be PHP 8.0.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -76,7 +76,7 @@ jobs:
         with:
           coverage: "none"
           extensions: "intl, zip"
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #9996, opened 2021-07-12
Allow versions like: "1.0.0-dev" and "4.8.0-p1"

MaXal:
We use the schema in PhpStorm to validate `composer.json` and now the schema is not correct.

See https://youtrack.jetbrains.com/issue/WI-60011 for reference.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -40,7 +40,7 @@
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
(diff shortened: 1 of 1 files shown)


Seldaek:
`^v?\\d+(\\.\\d+){0,3}` just says it must start with 1.0.0 or v1.0.0 format. It doesn't prevent any suffix, so I am not sure why you have issues.

stof:
This PR is breaking support for version like `dev-master`, so it is not just about making it more complex for nothing. It also introduces bugs.

alcohol:
Looks more like a bug on the side of Jetbrains to me. Submitted a question to the open issue on their tracker.

imme-emosol:
label:Support
close issue

PR #10002, opened 2021-07-14
Simplify code

ondrejmirtes:
`if ($cacheKey)` is always true because it's concatenated with literal parts a few lines above :)


ondrejmirtes:
To check that all I did was un-indenting the contents of the if branch, see this view: https://github.com/composer/composer/pull/10002/files?w=1

Seldaek:
Thanks

PR #10016, opened 2021-07-22
Highlight the proprietary license trick

tdutrion:
From the Symfony Dev #french slack channel (symfony-devs.slack.com), people look confused regarding the value to use as license for proprietary projects, even though it's written in the documentation.

Because proprietary software is still a massive part of composer's usage, I think it can be interesting to have it as a note, more readable to people.

Extract from the conversation:
> J’étais sur la bonne page, il me manquait deux lignes de scroll pour voir ça -.- On a tous nos petits moments de faiblesse

Which roughly translates to:
> I was looking at the right page, just about two lines above... We all have our weak moments


Seldaek:
Thanks, makes sense

PR #10018, opened 2021-07-22
In CurlDownloader, use the global timeout (instead of a hardcoded 5-minute timeout)

smokris:
`CurlDownloader` is hardcoded to a 5-minute timeout.  When downloading a large file on a slow internet connection, it hits this timeout, then aborts and restarts the download from the beginning, and `composer install` is unable to make any progress.

This PR modifies `CurlDownloader` to use the global `ProcessExecutor` timeout (`COMPOSER_PROCESS_TIMEOUT`) instead.  (`COMPOSER_PROCESS_TIMEOUT` defaults to 5 minutes, just like the hardcoded value in `CurlDownloader`, so the default behavior remains the same when that environment variable is not set.)  With this PR, when I set `COMPOSER_PROCESS_TIMEOUT` to an appropriate value for my internet connection, `composer install` is able to complete successfully.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -20,6 +20,7 @@ use Composer\Util\StreamContextFactory;
 use Composer\Util\AuthHelper;
 use Composer\Util\Url;
 use Composer\Util\HttpDownloader;
+use Composer\Util\ProcessExecutor;
 use React\Promise\Promise;
 
 /**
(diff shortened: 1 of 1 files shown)


Seldaek:
It's a valid point that this shouldn't be hardcoded, but IMO the process timeout isn't the way to change this. The `default_socket_timeout` ini setting is the closest thing we have to having this configurable in the past, so it now respects this too for curl transfers. I hope this lets you resolve your issue too.

smokris:
Ah, that makes sense.  With your commit, I confirmed that I'm able to increase the timeout by setting `default_socket_timeout=3600` in my `php.ini`, which enables me to download large files here.  Thanks!

Issue #10008, opened 2021-07-20
Deprecation notices with PHP 8.1


sebastianbergmann:
From https://github.com/sebastianbergmann/phpunit/runs/3111696987:

```
Deprecation Notice: Return type of Composer\Repository\ArrayRepository::count()
should either be compatible with Countable::count(): int,
or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
in phar:///home/runner/work/phpunit/phpunit/tools/composer/src/Composer/Repository/ArrayRepository.php:286
```
```
Deprecation Notice: Return type of Composer\Repository\ArrayRepository::count()
should either be compatible with Countable::count(): int,
or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
in phar:///home/runner/work/phpunit/phpunit/tools/composer/src/Composer/Repository/WritableArrayRepository.php:286
```
```
Deprecation Notice: Return type of Composer\Repository\ArrayRepository::count()
should either be compatible with Countable::count(): int,
or the #[ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
in phar:///home/runner/work/phpunit/phpunit/tools/composer/src/Composer/Repository/InstalledFilesystemRepository.php:286 
```

#### Related
* https://github.com/symfony/symfony/issues/42199
* https://github.com/justinrainbow/json-schema/issues/668

Seldaek:
@shivammathur sorry to nag you but as I've been wondering this for a while, I figured this is as good a time as any to ask.. What's the timeline on getting new composer releases to be included by setup-php? Is it dependent on some manual build step you have to do?

I just released 2.1.4 which fixes 8.1 deprecations and it'd be great to get it out to GH actions :)

Seldaek:
@shivammathur Similarly, I have now tried switching to `tools: composer:snapshot` but it seems I get the snapshot from some-time-yesterday and not the latest. It'd be great if snapshot would force it to download the latest always at least and not use the latest container you've built.

stof:
@Seldaek Looking at the setup-php code, it downloads it from https://github.com/shivammathur/composer-cache, which is updated by a cron once a day

shivammathur:
@Seldaek 
[...]

PR #10039, opened 2021-08-05
PHP 8.1: fix more return type deprecation warnings

jrfnl:
Follow up on #10008 and the various commits made for that.


Seldaek:
Thanks

PR #10047, opened 2021-08-11
Update to xdebug-handler 2.0.2

johnstevenson:
Changes:

- Added: support for xdebug_info('mode') in Xdebug 3.1.
- Added: support for Psr\Log versions 2 and 3.
- Fixed: remove ini directives from non-cli HOST/PATH sections.

Note that although Xdebug 3.1 has not yet been released, the new API `xdebug_info('mode')` has been documented: https://xdebug.org/docs/all_functions#xdebug_info

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -313,21 +313,21 @@
         },
         {
             "name": "composer/xdebug-handler",
-            "version": "2.0.1",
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

PR #10050, opened 2021-08-11
ClassMapGenerator: fix two bugs (long heredocs + markers in text)

### LINKED ISSUE #10037: Bug: long heredoc causes missing classes in classmap generation
The following deprecation notice shows during the generation of an optimized classmap on PHP 8.1:
```
Deprecation Notice: preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:251
```

Steps to reproduce:
* Clone the [phpDocumentor](https://github.com/phpDocumentor/phpDocumentor) repository
* Run `composer install --no-progress --prefer-dist --optimize-autoloader` while on PHP 8.1(beta1)

The error can be seen in the "Install Composer dependencies & cache dependencies" step in this CI transcript: https://github.com/jrfnl/phpDocumentor/runs/3241452949?check_suite_focus=true

This is caused by a bug in the regex for which I have a fix lined up. I'm still trying to verify, however, whether there is an existing test which already covers this or not.

This bug has probably been around since PHP 7.3 when PHP switched to PCRE 2.0, which contains stricter regex validation (`preg_replace()` returns `null` when there is an error in the regex).

jrfnl:
### ClassMapGenerator: add tests for "long heredoc" bug

... to proof the existence of the bug and demonstrate the effect.

Note: in the test the backtrack limit is being lowered (and restored back to the default afterwards) to prevent the tests needing ridiculously huge test fixture files.

### ClassMapGenerator: add test for "marker in text" bug

In PHP < 7.3, the heredoc/nowdoc marker was allowed to occur in the text, as long as it did not occur at the very start of the line.

This was also not handled correctly.

Ref: https://www.php.net/manual/en/migration73.incompatible.php#migration73.incompatible.core.heredoc-nowdoc

### ClassMapGenerator: fix the regex

By using a look ahead assertion to match "new line - maybe whitespace - marker", the negative performance impact of the `.*` is significantly mitigated and backtracing will be severely limited.

This fixes the bug as reported in #10037.

The bug was discovered due to a PHP 8.1 "passing null to non-nullable" deprecation notice being thrown, but is not a PHP 8.1 bug.

In actual fact, this issue affected all PHP versions and could lead to incomplete classmaps when the code base contained files with huge heredocs/nowdocs.

The regex change (not completely) incidentally also fixes an issue with markers in a heredoc/nowdoc not being correctly handled. This bug could lead to "classes" being added to the class map which aren't actually classes.

Fixes #10037

@Seldaek Please let me know if this should go to an earlier branch. Happy to rebase.

### DIFF
--- a/tests/Composer/Test/Autoload/ClassMapGeneratorTest.php
+++ b/tests/Composer/Test/Autoload/ClassMapGeneratorTest.php
@@ -244,6 +244,25 @@ class ClassMapGeneratorTest extends TestCase
         $fs->removeDirectory($tempDir);
     }
 
+    public function testCreateMapDoesNotHitRegexBacktraceLimit()
+    {
+        $expected = array(
+            'Foo\\StripNoise'            => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/StripNoise.php',
+            'Foo\\VeryLongHeredoc'       => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongHeredoc.php',
+            'Foo\\ClassAfterLongHereDoc' => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongHeredoc.php',
+            'Foo\\VeryLongPHP73Heredoc'  => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongPHP73Heredoc.php',
+            'Foo\\VeryLongPHP73Nowdoc'   => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongPHP73Nowdoc.php',
+            'Foo\\ClassAfterLongNowDoc'  => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongPHP73Nowdoc.php',
+            'Foo\\VeryLongNowdoc'        => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongNowdoc.php',
+        );
+
+        ini_set('pcre.backtrack_limit', '30000');
+        $result = ClassMapGenerator::createMap(__DIR__ . '/Fixtures/pcrebacktracelimit');
+        ini_restore('pcre.backtrack_limit');
+
+        $this->assertEqualsNormalized($expected, $result);
+    }
+
     protected function assertEqualsNormalized($expected, $actual, $message = '')
     {
         foreach ($expected as $ns => $path) {
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -246,7 +246,7 @@ class ClassMapGenerator
         }
 
         // strip heredocs/nowdocs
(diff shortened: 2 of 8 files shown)


tests/Composer/Test/Autoload/ClassMapGeneratorTest.php
```diff
@@ -244,6 +244,25 @@ public function testDump()
...
+            'Foo\\ClassAfterLongHereDoc' => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongHeredoc.php',
+            'Foo\\VeryLongPHP73Heredoc'  => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongPHP73Heredoc.php',
+            'Foo\\VeryLongPHP73Nowdoc'   => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongPHP73Nowdoc.php',
+            'Foo\\ClassAfterLongNowDoc'  => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongPHP73Nowdoc.php',
+            'Foo\\VeryLongNowdoc'        => realpath(__DIR__) . '/Fixtures/pcrebacktracelimit/VeryLongNowdoc.php',
+        );
+
+        ini_set('pcre.backtrack_limit', '30000');
+        $result = ClassMapGenerator::createMap(__DIR__ . '/Fixtures/pcrebacktracelimit');
+        ini_restore('pcre.backtrack_limit');
```
staabm, line 261: Should this use a try/finally, so in case of exceptions the value gets properly restored?

staabm: Should this retrieve the previous value and re-set it instead of restoring the initial value, which might not be the expected one?

jrfnl: @staabm Good question. I'd done it like this to ensure that, even if the _assertion_ fails, restore will still happen, but yes, in case of exceptions it may not.

I'm not a fan of using `try/catch` in test methods. Moving it to the `tearDown()`  would be more appropriate IMO, or even moving the method to its own test class and then having the ini changes in `setUp()` and `tearDown()`.

What do you think ?

Seldaek: IMO it's fine as is. Low chances of problems, and tests only, so not really worth a debate :)

Seldaek:
Thanks, given the rather exotic aspect of the fix, I don't think it's worth backporting to 1.x, so it's fine targetting 2.1 👍🏻

Seldaek:
Yeah the more exotic bugs are usually the more interesting ones. It's not everyday you discover age-old bugs in widely used software :)

jrfnl:
Yup, when I looked at the history I saw this particular regex hadn't been touched since 2012... 😂  (and was introduced to move away from the Tokenizer, which makes sense to me for all sorts of reasons)

jrfnl:
Just out of curiousity - were there ever any issues reported about classes missing from the classmap where the cause was never previously found, which _might_ just be related to this issue ?
Or about the classmap containing non-existent classes ?

Seldaek:
I can't remember any such issue in the recent years. Early on as we ironed out issues sure but I guess such long heredocs really are that rare.

PR #10052, opened 2021-08-12
Rule::getRequiredPackage(): always return a string

jrfnl:
The return type of the `Rule::getRequiredPackage()` method effectively was `?string` as, if no matching reason was found, the function wasn't returning anything.

In PHP 8.1, this results in 97 notices along the lines of the below from existing tests:
```
Deprecation triggered by Composer\Test\DependencyResolver\SolverTest::testConflictResultEmpty:

Stack trace:
0 [internal function]: Symfony\Bridge\PhpUnit\DeprecationErrorHandler->handleError(8192, '...', '...', 111)
1 src/Composer/DependencyResolver/SolverProblemsException.php(111): strpos(NULL, '...')
2 src/Composer/DependencyResolver/SolverProblemsException.php(44): Composer\DependencyResolver\SolverProblemsException->hasExtensionProblems(Array)
3 tests/Composer/Test/DependencyResolver/SolverTest.php(659): Composer\DependencyResolver\SolverProblemsException->getPrettyString(Object(Composer\Repository\RepositorySet), Object(Composer\DependencyResolver\Request), Object(Composer\DependencyResolver\Pool), false)
....
```

Fixed by returning an empty string if no reason was matched.

👉🏻 I'm actually a little surprised this wasn't being flagged by PHPStan. Couldn't find an issue for it in the baseline either.

### DIFF
--- a/src/Composer/DependencyResolver/Rule.php
+++ b/src/Composer/DependencyResolver/Rule.php
@@ -91,6 +91,8 @@ abstract class Rule
         if ($reason === self::RULE_PACKAGE_REQUIRES) {
             return $this->reasonData->getTarget();
         }
+
+        return '';
     }
 
     public function setType($type)


staabm:
> I'm actually a little surprised this wasn't being flagged by PHPStan. Couldn't find an issue for it in the baseline either.

I guess reason is that the phpstan level is not high enough for this problem to be checked

https://github.com/composer/composer/blob/c65bd832d6529a542ad5f74d0e49d3353f8632c8/phpstan/config.neon#L6

Seldaek:
Fixed the usage rather than the returned value, because IMO if there is no required package null is a better value than an empty string, thanks for pointing it out though.

jrfnl:
Thanks @Seldaek !

Issue #6755, opened 2017-10-18
single pipe in version constraint, undocumented or buggy?


joachim-n:
My `composer.json`:

```json
        "drush/drush": "~8.0|^9.0.0-beta7",

```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
```

When I run this command:

```
composer validate
```

I get the following output:

```
./composer.json is valid
```

And I expected this to happen:

The use of the | (single pipe) symbol is not documented at https://getcomposer.org/doc/articles/versions.md#caret-version-range-. Therefore, either the documentation has an error, or 'composer validate' does.

stof:
Well, a single pipe is the older OR operator, just like AND can be done either by a comma or just by putting spaces.
The doc at https://getcomposer.org/doc/articles/versions.md#version-range has been simplified a while ago to show only the recommended alternative for the OR operator (the one being described in the semver spec).

joachim-n:
> The doc at https://getcomposer.org/doc/articles/versions.md#version-range has been simplified a while ago to show only the recommended alternative for the OR operator (the one being described in the semver spec).

That's a mistake. Docs should still mention deprecated features, while stating that they are deprecated. Otherwise, users who encounter deprecated features in use won't be able to understand them.

alcohol:
@joachim-n opinions vary.

We prefer not to pollute the documentation with details that are no longer relevant.

joachim-n:
It's relevant in that users may encounter it in libraries that still use it in their composer.json files, and then need to understand what it means!

alcohol:
We cannot wait for everyone to update their files/libraries. Some of those never will. So no, it is not relevant. Not anymore. Also, the amount of times people actually run into this is insignificant.

fritzmg:
Does Composer 2.x still support single pipe ORs, or will the packages that still use the single piples have to be updated in order to work with Composer 2.x?

[...]

PR #10057, opened 2021-08-13
Note in docs about how single pipes are treated

nicholsonjf:
The single pipe (the old alternative to logical OR) is treated as a logical OR but is not currently documented. This can be confusing to developers new to composer who encounter a single pipe and look to the docs for an explaination.

See this issue for more detail: https://github.com/composer/composer/issues/6755#issuecomment-898537753


Seldaek:
Thanks!

Issue #10038, opened 2021-08-05
PHP 8.1 compatibility


jrfnl:
While trying to figure out whether there are any tests covering a particular PHP 8.1 issue I'd encountered elsewhere (see #10037), I toggled the value for `SYMFONY_DEPRECATIONS_HELPER` in the `phpunit.xml.dist` file from `weak` to `max[self]=0` and found there are still some 458 deprecation notices to be addressed, with only 11 of those related to underlying dependencies of Composer.

I've already fixed a few in #10036 (fixes 176 notices).

I'm willing to spend some more time on creating fixes for these issues, but wanted to check first if there is already work ongoing for this or not, to prevent interfering with ongoing efforts/double work.

As a side-note: fixing these issues is kind of slow going as I need to go via the CI to get the issue details as I have not been able to get the tests running locally on PHP 8.1 with Symfony PHPUnit Bridge.

It might be an idea to add some additional information to the `CONTRIBUTING` doc to guide people on how to get the tests running locally on PHP 8.0+ ?

---
<details>
  <summary>Summary list of deprecation notices found in the test run (fold out for details)</summary>

```
Unsilenced deprecation notices (458)

  205x: preg_split(): Passing null to parameter #3 ($limit) of type int is deprecated
    157x in InstallerTest::getIntegrationTests from Composer\Test
    18x in PoolBuilderTest::getIntegrationTests from Composer\Test\DependencyResolver
    8x in NoProxyPatternTest::testHostName from Composer\Test\Util
    6x in NoProxyPatternTest::testIpAddress from Composer\Test\Util
    6x in NoProxyPatternTest::testIpRange from Composer\Test\Util
    4x in NoProxyPatternTest::testPort from Composer\Test\Util
    3x in ProxyManagerTest::testGetProxyForRequest from Composer\Test\Util\Http
    1x in InstallerTest::getSlowIntegrationTests from Composer\Test
    1x in StreamContextFactoryTest::testHttpProxyWithNoProxy from Composer\Test\Util
    1x in StreamContextFactoryTest::testHttpProxyWithNoProxyWildcard from Composer\Test\Util

  128x: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated
    95x in InstallerTest::testIntegration from Composer\Test
    25x in VersionSelectorTest::testFindRecommendedRequireVersion from Composer\Test\Package\Version
[...]

PR #10062, opened 2021-08-18
PHPUnit: move environment variable into PHPUnit config file

jrfnl:
... to make it easier on contributors to get the tests running locally with the current setup.

Also includes:
* Adding the typical PHPUnit 8/9 cache file to `.gitignore`.
* Adding information on how to get the tests running locally on PHP 8.0/8.1 to the CONTRIBUTING file.

Related to #10038

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -42,9 +42,17 @@ To achieve this, you need to acquire the Composer source code:
 3. Run Composer to get the dependencies: `cd composer && php ../composer.phar install`
 
 You can run the test suite by executing `vendor/bin/simple-phpunit` when inside the
-composer directory, and run Composer by executing the `bin/composer`. To test
-your modified Composer code against another project, run `php
-/path/to/composer/bin/composer` inside that project's directory.
+composer directory, and run Composer by executing the `bin/composer`.
+
+For running the tests against the most recent PHP versions (PHP 8.0/8.1), you will
+need to run `composer config platform.php --unset && composer update --ignore-platform-reqs`
+before running the `vendor/bin/simple-phpunit` command.
+If that still doesn't work, try running
+`composer require --dev "symfony/phpunit-bridge:~5.2" --update-with-dependencies` and
+then running the `vendor/bin/simple-phpunit` command again.
+
+To test your modified Composer code against another project, run
+`php /path/to/composer/bin/composer` inside that project's directory.
 
 Contributing policy
 -------------------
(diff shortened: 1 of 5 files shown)


.github/CONTRIBUTING.md
```diff
@@ -42,9 +42,17 @@ To achieve this, you need to acquire the Composer source code:
 3. Run Composer to get the dependencies: `cd composer && php ../composer.phar install`
 
 You can run the test suite by executing `vendor/bin/simple-phpunit` when inside the
-composer directory, and run Composer by executing the `bin/composer`. To test
-your modified Composer code against another project, run `php
-/path/to/composer/bin/composer` inside that project's directory.
+composer directory, and run Composer by executing the `bin/composer`.
+
+For running the tests against the most recent PHP versions (PHP 8.0/8.1), you will
+need to run `composerdev config platform.php --unset && composer update --ignore-platform-reqs`
```
staabm, line 48: `composerdev` reads like a typo

jrfnl: Oops.. yes, thanks for catching that. Copy/pasta. Fixed now.

.github/CONTRIBUTING.md
```diff
@@ -42,9 +42,17 @@ To achieve this, you need to acquire the Composer source code:
...
-your modified Composer code against another project, run `php
-/path/to/composer/bin/composer` inside that project's directory.
+composer directory, and run Composer by executing the `bin/composer`.
+
+For running the tests against the most recent PHP versions (PHP 8.0/8.1), you will
+need to run `composer config platform.php --unset && composer update --ignore-platform-reqs`
+before running the `vendor/bin/simple-phpunit` command.
+If that still doesn't work, try running
+`composer require --dev "symfony/phpunit-bridge:~5.2" --update-with-dependencies` and
+then running the `vendor/bin/simple-phpunit` command again.
```
Seldaek, line 52: Actually good point you made here, --ignore-platform-reqs is enough, and it's cleaner as it won't do any changes to composer.json, so I believe this is sufficient:

```suggestion
For running the tests against the most recent PHP versions (PHP 8.0/8.1), you will
need to run `composer update --ignore-platform-reqs`  before running 
the `vendor/bin/simple-phpunit` command.
```

Seldaek:
Thanks!

.github/CONTRIBUTING.md
jrfnl: Yes, of course ;-)
I used the original steps and in that case, the `--ignore-platform-reqs` is only needed for PHP 8.1 as one of the PHPUnit requirements hasn't declared PHP 8.1 compatibility yet, though I suspect that will be solved soon enough.

Issue #10024, opened 2021-07-27
Composer does not respect plugin files autoload requirements


cs278:
Composer does not appear to load the [`files`](https://getcomposer.org/doc/04-schema.md#files) section of the `autoload` definition for plugins. This appears to hold true for the plugin itself and any dependencies, I expect this might be a design decision but I cannot see it documented anywhere. I've stumbled upon this while investigating a problem with the new releases of `symfony/yaml` package using PHP 8 functions breaking plugins on PHP 7.4 which led me to https://github.com/symfony/symfony/issues/42280. I've put together a [test plugin](https://github.com/cs278/composer-include-test-plugin) to demonstrate the problem.

My `composer.json`:

```json
{
    "repositories": [{
            "type": "git",
            "url": "https://github.com/cs278/composer-include-test-plugin"
        }],
    "require": {
        "cs278/composer-include-test-plugin": "@dev"
    }
}

```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
require.cs278/composer-include-test-plugin : unbound version constraints (@dev) should be avoided
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: FAIL
Missing pubkey for tags verification
Missing pubkey for dev verification
Run composer self-update --update-keys to set them up
Checking composer version: OK
Composer version: 2.1.5
PHP version: 7.4.21
PHP binary path: /usr/bin/php7.4
OpenSSL version: OpenSSL 1.1.1  11 Sep 2018
cURL version: 7.58.0 libz 1.2.11 ssl OpenSSL/1.1.1
zip: extension present, unzip present, 7-Zip present (7z)
```

When I run this command: 

```
php -f composer.phar test-plugin -vvv
```

I get the following output: 

```
Reading ./composer.json (/tmp/chris.smith/tmp.DROuBGKkqm/composer.json)
Loading config file ./composer.json (/tmp/chris.smith/tmp.DROuBGKkqm/composer.json)
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/tmp/chris.smith/tmp.DROuBGKkqm): git branch -a --no-color --no-abbrev -v
Executing command (/tmp/chris.smith/tmp.DROuBGKkqm): git describe --exact-match --tags
[...]

PR #10065, opened 2021-08-19
Fix loading files section for plugins

nicolas-grekas:
Trying to help move #10024 forward.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -571,6 +571,12 @@ EOF;
             }
         }
 
+        if (isset($autoloads['files'])) {
+            foreach ($autoloads['files'] as $fileIdentifier => $file) {
+                composerRequire($fileIdentifier, $file);
+            }
+        }
+
         return $loader;
     }
 
@@ -1283,5 +1289,14 @@ INITIALIZER;
         }
 
         return $sortedPackageMap;
+    }
+}
+
+function composerRequire($fileIdentifier, $file)
+{
+    if (empty($GLOBALS['__composer_autoload_files'][$fileIdentifier])) {
+        require $file;
+
+        $GLOBALS['__composer_autoload_files'][$fileIdentifier] = true;
     }
 }


Seldaek:
See https://github.com/composer/composer/issues/10024#issuecomment-888270023 tho :) 

But yeah, while the situation with updates isn't ideal and may lead to broken processes, it's maybe better than the current situation.. I am not sure really.

nicolas-grekas:
I was that, but I miss reasons why to understand this feeling you express in that comment :)

Seldaek:
Well.. there is https://github.com/composer/composer/issues/10024#issuecomment-887466557 showing an example scenario.

Another scenario that could break is like:

- Package A has func.php autoloaded, declaring a bunch of functions.
- It gets autoloaded, then updated in the same process to a new version which renamed to functions.php
- when autoloading it again, functions.php is now included, and include_once won't save you there, and you get a fatal because of redefined function..

Plus the fact that these files were never loaded during composer runs, and now they would be, and that may trigger all sorts of unforeseen effects.

So I'm not 100% against, but it is not without risks, and I'm not sure it's worth it.

jderusse:
> Another scenario that could break is like:
> 
> * Package A has func.php autoloaded, declaring a bunch of functions.
> * It gets autoloaded, then updated in the same process to a new version which renamed to functions.php
> * when autoloading it again, functions.php is now included, and include_once won't save you there, and you get a fatal because of redefined function..

don't we have the same issue with class autoloader? when a plugin use a class that is updated?

Seldaek:
Not really because the class is already in memory and it won't be loaded again. Sure you may have problems if the code changed massively and is not compatible anymore, but that's something we really cannot fix. Plugins should thus ideally be rather self-contained as much as possible.

jderusse:
sounds like the same edge case.. one could changes code of a class, adds methods, changes interface signature...

If it worth it for classes, I believe it is the same for functions/polyfills


> Plus the fact that these files were never loaded during composer runs, and now they would be, and that may trigger all sorts of unforeseen effects.

Indeed, this is the major risk of this change.

nicolas-grekas:
> It gets autoloaded, then updated in the same process to a new version which renamed to functions.php

Maybe plugins shouldn't be allowed to have a "files" section, but their dependencies should?
The linked issue is quite annoying to say the least. It almost kills the concept of polyfills, in the generic sense.
This risk fir updates is just a fatal error that can be worked around by restarting the process (maybe after deleting `vendor/`), isn't it? I feel like this is an acceptable tradeoff.


> Indeed, this is the major risk of this change.

I don't think it is: those files are autoloaded for regular runs without such issues.

Seldaek:
Ok, maybe we can try it out for 2.2, but I think it's too risky to smuggle this as a bugfix.

nicolas-grekas:
Works for me, PR rebased.

helhum:
What worries me more with such change aren't polyfill packages for which the files are included, but the fact that literally anything can be done in these inclusion files, as there is no specification for that. While one could of course include side effects in class files as well, doing so (IIRC) violates the PSR-4 specification.

But we indeed will not find out without trying how much will break. Hopefully we're lucky.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -571,6 +571,12 @@ public function createLoader(array $autoloads, $vendorDir = null)
             }
         }
 
+        if (isset($autoloads['files'])) {
+            foreach ($autoloads['files'] as $fileIdentifier => $file) {
+                composerRequire($fileIdentifier, $file);
```
helhum, line 576: Does this have to be a globally declared function, being declared in a class file? Wouldn't an immediately called closure be scoped enough as well?

hexus: I'd personally suggest against an immediately invoked function if we expect this method to be called multiple times in a single execution, to avoid resource consumption.

That said, I totally agree that global scope shouldn't be polluted here. A `static` method on this class (or some utility class?) could achieve the same result while safely using namespaces.

nicolas-grekas: this function is not global, it's namespaced

helhum: true. what do you see as benefit of a namespaced function over a static class method?

hexus: Whoops, I didn't clock that it was namespaced. In that case, either approach works well!

nicolas-grekas: > what do you see as benefit of a namespaced function over a static class method?

A static function gives access to the class' scope. A namespaced one doesn't.

nicolas-grekas:
> literally anything can be done in these inclusion files

This is true, but this is not specific to plugins. Regular auloaders already load those files. The plugin's autoloader has nothing specific to care about on this topic.

src/Composer/Autoload/AutoloadGenerator.php
helhum: Well, yes. I still don't like to declare other symbols in a class file, but I guess it'll work fine.

helhum:
> > literally anything can be done in these inclusion files
> 
> This is true, but this is not specific to plugins. Regular auloaders already load those files. The plugin's autoloader has nothing specific to care about on this topic.

There is quite a difference between runtime inclusion of the autoload file and building a temporary autoloader on the fly during Composer build time.

One example Jordi gave before is renaming the inclusion file and declaring the same symbols during update. 

In previous Composer versions composer updates failed likely, when a plug-in was updated with incompatible classes. Now the update process is much mir likely to fail for any dependent package changing autoload includes.

Anyway. We'll see hinein goes. Maybe it'll be fine

mvorisek:
I belive this is a bug and `files` autoload should work everywhere. Nice example are php polyfill scripts or other global config/declarations.

So 🚀 from my side. If execution of plugins can be easily detected, please post the condition here so developers who did not want `files` autoload feature in their scripts can use it.

Seldaek:
Thanks!

barryvdh:
Is this causing the issues in #10277 ?
I'm getting various autoload errors on Laravel and Magento2 projects.

barryvdh:
Magento2 uses the file autoloader for module registration, but that seems to rely on having the normal namespaced autoloading available.

Seldaek:
@barryvdh yes most likely, please try with the latest snapshot it was just fixed in https://github.com/composer/composer/pull/10279

barryvdh:
This fixed my issue with Laravel, but not with Magento 2 unfortunately.

```
s 'Magento\Framework\Component\ComponentRegistrar' not found in /var/www/project/app/code/MageB2B/Staff/registration.php:9
Stack trace:
#0 /var/www/project/app/etc/NonComposerComponentRegistration.php(29): require_once()
#1 [internal function]: Magento\NonComposerComponentRegistration\{closure}()
#2 /var/www/project/app/etc/NonComposerComponentRegistration.php(31): array_map()
#3 /var/www/project/app/etc/NonComposerComponentRegistration.php(34): Magento\NonComposerComponentRegistration\{closure}()
#4 phar:///usr/local/bin/composer/src/Composer/Autoload/AutoloadGenerator.php(1419): require('/var/www/project...')
#5 phar:///usr/local/bin/composer/src/Composer/Plugin/PluginManager.php(206): Composer\Autoload\composerRequire()
#6 phar:///usr/local/bin/composer/src/Composer/Plugin/PluginManager.php(425): Composer\Plugin\PluginManager->registerPackage()
#7 phar:///usr/local/bin/composer/src/Composer/Plugin/PluginManager.php(87): Composer\Plugin\PluginManager->loadRepository()
#8 phar:///usr/local/bin in /var/www/project/app/code/MageB2B/Staff/registration.php on line 9
```

Seldaek:
@barryvdh please report an issue with more details how one can reproduce this. I'm assuming it's an issue with a magento package/plugin not requiring all its dependencies though..

PR #10070, opened 2021-08-20
PHPStan Level 5

Seldaek:
Moar type fixes 🥳 

Level 6 is going to be a tough one though, we have 4000 errors (of which 2400 in src/), mostly missing property, parameter and return types. It's going to be a huge undertaking.


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -41,7 +41,7 @@ class AutoloadGenerator
     private $eventDispatcher;
 
     /**
-     * @var IOInterface
+     * @var ?IOInterface
```
williamdes, line 44: I do not understand why you use this format, it's not standard IMO
Do you have links to docs that make it a standard ?

Seldaek: It's the format PHP implemented for typed properties so IMO anything not supporting that has a bug but that is just an opinion :) Why/where is it causing issues for you?

williamdes: Yes but for me this is not a valid syntax according to everything I see on code bases and to https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html

Let me know, maybe I missed something obvious somewhere

Seldaek: Well again the question is do you experience an issue in practice? Because I don't really care for the theory here. Phpdocumentor is one implementation of something reading those phpdoc annotations, but it is not an implementation we use in this project, so I don't think what it supports is relevant here.

Phpstan handles ?foo very well for example, and we do use phpstan in the CI.

williamdes: It will probably output errors using [Doctum (fork of Sami)](https://github.com/code-lts/doctum#readme) because it uses phpdocumentor composer lib

For now there is no real issue, but I wanted to learn about that because this is the second repository with phpunit that I see doing that

Issue #10069, opened 2021-08-20
Failed to extract ZIP


cafferata:
**Problem**

When you upgrade the package [simple-bus/symfony-bridge][1] from `5.0.0` to `5.1.0`, the error message below keeps appearing.

The https://github.com/SimpleBus/symfony-bridge/compare/v5.0.0...v5.1.0 with possible cause https://github.com/SimpleBus/symfony-bridge/commit/24e0207d99241ecf2600be4e72919487c7210ad3?

**Reproducible case**

Create an empty directory, and run both composer require commands.

```bash
composer require "simple-bus/symfony-bridge" "5.0.0"
composer require "simple-bus/symfony-bridge" "5.1.0"
```

My `composer.json`:

```json
{
    "require": {
        "simple-bus/symfony-bridge": "5.0.0"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
require.simple-bus/symfony-bridge : exact version constraints (5.1.0) should be avoided if the package follows semantic versioning
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.1.6
PHP version: 7.4.22
PHP binary path: /usr/local/Cellar/php@7.4/7.4.22/bin/php
OpenSSL version: OpenSSL 1.1.1k  25 Mar 2021
cURL version: 7.78.0 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1k
zip: extension present, unzip present, 7-Zip not available
```

When I run this command: 

```
composer require "simple-bus/symfony-bridge" "5.1.0"
```

I get the following output:

```
./composer.json has been updated
Running composer update simple-bus/symfony-bridge
Loading composer repositories with package information
Updating dependencies
Lock file operations: 0 installs, 1 update, 0 removals
  - Upgrading simple-bus/symfony-bridge (v5.0.0 => v5.1.0)
[...]

PR #10076, opened 2021-08-25
Correctly resolve promises when package type changes

aschempp:
This fixes https://github.com/composer/composer/issues/10069

When the type of a package changes, Composer does an uninstall and install instead of an update. 78d7792eb810319f86421071bf795ce41eca24f2 added support for handling promises of the installer, but because it returns the `uninstall` promise instead of the resulting `install` promise, the `cleanup` operation in https://github.com/composer/composer/blob/master/src/Composer/Installer/InstallationManager.php#L420 runs before `install` is executed.

---

FYI instead of setting `$promise = $promise->then(…)` we could also just `return $promise->then(…)`

### DIFF
--- a/src/Composer/Installer/InstallationManager.php
+++ b/src/Composer/Installer/InstallationManager.php
@@ -511,7 +511,7 @@ class InstallationManager
             }
 
             $installer = $this->getInstaller($targetType);
-            $promise->then(function () use ($installer, $repo, $target) {
+            $promise = $promise->then(function () use ($installer, $repo, $target) {
                 return $installer->install($repo, $target);
             });
         }


Seldaek:
Thanks!

jnaklaas:
Can I test this, I've got the same bug when extracting my latest private packages? Tried to `composer self-update master` but that didn't work.

Toflar:
You can update to the snapchot channel using `composer selfupdate --snapshot`.

jnaklaas:
Okay thanks. Tried that, the issue didn't get resolved though. I also noticed it's not exactly the same bug.

Issue #10045, opened 2021-08-10
Support wildcard in option ignore-platform-req


dahaupt:
When installing or updating packages, it would be useful if all platform requirements could be ignored, except for PHP.

Currently, this is only possible by ignoring each extension individually, e.g.: 
`composer update --ignore-platform-req=ext-json --ignore-platform-req=ext-xml`

or by _faking_ platform requirements in `config.platform`, e.g.:
```
"config": {
    "platform": {
        "ext-json": "7.4",
        "ext-xml": "7.4"
    }
},
```

We have this use case when updating composer projects with Renovate. See details in linked issue(s).

So the subject of this issue could also be:
**Installing Composer dependencies in environments without valid platform requirements while respecting the PHP version requirement**

Related: 
- https://github.com/renovatebot/renovate/issues/2355
- https://stackoverflow.com/questions/65367417/is-there-any-way-to-ignore-platform-req-all-but-keep-one-in-composer-json

dereuromark:
So basically you want

    composer update --ignore-platform-req=ext-*
    
Seems reasonable to me, if this kind of wildcard syntax can be parsed into some PHP comparison checks etc.

dahaupt:
@dereuromark Exactly! 

Basically, we need `--ignore-platform-reqs` without ignoring the PHP version. But supporting wildcards would be useful enough and also helpful for other use cases, I think.

dereuromark:
Maybe also the opposite could be useful then for your case, using `-` prefix to invert?

    composer update --ignore-platform-req=-php

Just spitballing some ideas here for the team to assess.

dahaupt:
@dereuromark My tests have shown that a comma separated list is not yet supported.

Correct me If I am wrong, but running 

```
composer update --ignore-platform-req=!php
``` 

on its own would be useless, since all platform requirements are checked by default.

It would make sense, e.g. by running 
```
composer update --ignore-platform-req=*,!php` 
```
or combined with `--ignore-platform-reqs` as
```
composer update --ignore-platform-reqs --ignore-platform-req=!php
```

stof:
a comma separated list is indeed not supported. To pass multiple values, you need to repeat the `--ignore-platform-req=`

dereuromark:
Sure, this was just spitballing some ideas. Would sure be trivial to add and quite useful over the repeating of such config.
I do that also on some configs to allow a list of params the easy way.

[...]

Issue #9534, opened 2020-11-30
Feature Request: Ignore only upper bound platform reqs


dbrumann:
**Introduce a new flag similar to `--ignore-platform-reqs` that only lifts upper bound requirements, while still observing lower bound requirements.**

I propose a new flag `--ignore-upper-bound-platform-reqs` (alternative suggestions welcome) which works similar to the existing `--ignore-platform-reqs`, but only ignores higher platform versions, not lower ones. I would be happy to work on a PR for it.

Additionally it might make sense for composer to ignore unrestricted platform versions that were unknown at the point of the release. For example [phpunit/phpunit:4.8.18](https://packagist.org/packages/phpunit/phpunit#4.8.18) (requiring `php: >=5.3.3`) should likely not be a viable candidate when running`composer require phpunit/phpunit` on the latest PHP nightly build.

---

*Background:*

As discussed with @naderman and @derrabus recently, lifting upper bound platform requirements might lead to issues. Most likely the package was not tested against the newest, unstable version. This is less of an issue for packages that proactively test against newest snapshots and address issues early, but especially if package maintenance slows down this can later lead to issues.

Putting in the restriction afterwards will cause Composer to still pull in an earlier, still not working, version of the package instead. This is especially detrimental with packages that are updated less frequently and the issue might not be obvious if one does not rely on that package directly.

*Example:*
Package `foo` in 1.0 has the following platform requirement: `php:>=7.4`.

An issue is found with PHP 8.0 that can not be addressed directly and therefore the next (minor or patch) release of `foo` changes the requirement to `php:^7.4` to make sure it no longer proclaims to work with the untested, newer version of PHP.

If I now run `composer require foo` in my project, where I use PHP 8.0, composer will still pick `foo:1.0` even though it is not "actually" supported.

In order to make testing against newer versions easier while still making sure that platform requirements can be set to the actually supported versions a new option similar to `--ignore-platform-reqs` would be helpful.

derrabus:
If I may add, `--ignore-platform-reqs=php` seems to do the trick at first, but that would ignore PHP requirements of packages completely. But I would want to keep the lower boundaries intact: I'd like to have an option that does not become a footgun if I forget to remove it from my build scripts.

Seldaek:
[...]

PR #10079, opened 2021-08-26
Refactor ignore platform reqs checks

herndlm:
Introduces a `PlatformRequirementFilter` with methods that help to decide if a requirement is ignored or not as discussed in #10045 but without changing behaviour.
The idea behind this is to be able to adapt simple wildcard or regex ignores in a follow-up.

I changed as much as needed so that the filter and its logic is used everywhere where the `bool|string[]` config was before, because it felt inconsistent otherwise. But I also tried to change as less as possible which is the reason the argument is nullable in some cases and falls back to the nothing filter then or is defined like that in the constructor.
Initially I only had one filter class, but this became already too complex and mixed regarding types in my opinion, so I split responsibilities into dedicated filter classes, which is of course more code/classes but way simpler to read/understand/test.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -14,7 +14,9 @@ namespace Composer\Autoload;
 
 use Composer\Config;
 use Composer\EventDispatcher\EventDispatcher;
-use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilter;
+use Composer\Filter\PlatformRequirementFilter\IgnoreAllPlatformRequirementFilter;
+use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilterFactory;
+use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilterInterface;
 use Composer\Installer\InstallationManager;
 use Composer\IO\IOInterface;
 use Composer\Package\AliasPackage;
@@ -70,7 +72,7 @@ class AutoloadGenerator
     private $runScripts = false;
 
     /**
-     * @var PlatformRequirementFilter
+     * @var PlatformRequirementFilterInterface
      */
     private $platformRequirementFilter;
 
@@ -79,7 +81,7 @@ class AutoloadGenerator
         $this->eventDispatcher = $eventDispatcher;
         $this->io = $io;
 
-        $this->platformRequirementFilter = PlatformRequirementFilter::ignoreNothing();
+        $this->platformRequirementFilter = PlatformRequirementFilterFactory::ignoreNothing();
     }
 
     /**
@@ -142,13 +144,13 @@ class AutoloadGenerator
     {
         trigger_error('AutoloadGenerator::setIgnorePlatformRequirements is deprecated since Composer 2.2, use setPlatformRequirementFilter instead.', E_USER_DEPRECATED);
 
-        $this->setPlatformRequirementFilter(PlatformRequirementFilter::fromBoolOrList($ignorePlatformReqs));
+        $this->setPlatformRequirementFilter(PlatformRequirementFilterFactory::fromBoolOrList($ignorePlatformReqs));
     }
 
     /**
      * @return void
      */
-    public function setPlatformRequirementFilter(PlatformRequirementFilter $platformRequirementFilter)
+    public function setPlatformRequirementFilter(PlatformRequirementFilterInterface $platformRequirementFilter)
     {
         $this->platformRequirementFilter = $platformRequirementFilter;
     }
@@ -399,7 +401,7 @@ EOF;
             unlink($includeFilesFilePath);
         }
         $filesystem->filePutContentsIfModified($targetDir.'/autoload_static.php', $this->getStaticFile($suffix, $targetDir, $vendorPath, $basePath, $staticPhpVersion));
-        $checkPlatform = $config->get('platform-check') && !$this->platformRequirementFilter->isAllIgnored();
+        $checkPlatform = $config->get('platform-check') && !($this->platformRequirementFilter instanceof IgnoreAllPlatformRequirementFilter);
         $platformCheckContent = null;
         if ($checkPlatform) {
             $platformCheckContent = $this->getPlatformCheck($packageMap, $config->get('platform-check'), $devPackageNames);
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -771,7 +771,7 @@ EOF;
             }
 
             foreach ($package->getRequires() as $link) {
-                if ($this->platformRequirementFilter->isReqIgnored($link->getTarget())) {
+                if ($this->platformRequirementFilter->isIgnored($link->getTarget())) {
                     continue;
                 }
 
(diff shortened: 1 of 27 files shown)


herndlm:
Nevermind, had an idea to improve this, I'll adapt it and open it for review then

herndlm:
Is it safe to do those changes from an API perspective? E.g. could plugins use those classes? And if so - if they extend from them too, can we even change any non-private signatures?

src/Composer/Installer.php
```diff
@@ -1174,24 +1180,12 @@ public function isVerbose()
...
-     *
-     * If this is set to true, all platform requirements are ignored
-     * If this is set to false, no platform requirements are ignored
-     * If this is set to string[], those packages will be ignored
-     *
-     * @param  bool|array $ignorePlatformReqs
+     * @param PlatformRequirementFilter $platformRequirementFilter
      * @return Installer
      */
-    public function setIgnorePlatformRequirements($ignorePlatformReqs)
```
Seldaek, line 1186: I think it'd be good to keep the old method around here too for API compatibility, which would I guess simply call `PlatformRequirementFilter::fromBoolOrList($ignorePlatformReqs)`.

src/Composer/Package/Version/VersionSelector.php
```diff
@@ -53,30 +54,32 @@ public function __construct(RepositorySet $repositorySet, PlatformRepository $pl
...
-     * @param  string                 $preferredStability
-     * @param  bool|array             $ignorePlatformReqs
+     * @param  string                    $packageName
+     * @param  string                    $targetPackageVersion
+     * @param  string                    $preferredStability
+     * @param  PlatformRequirementFilter $platformRequirementFilter
+     * @param  int                       $repoSetFlags
      * @return PackageInterface|false
      */
-    public function findBestCandidate($packageName, $targetPackageVersion = null, $preferredStability = 'stable', $ignorePlatformReqs = false, $repoSetFlags = 0)
```
Seldaek, line 62: I think it'd be good to keep the supporting an array here too for API compatibility, which would I guess simply call `PlatformRequirementFilter::fromBoolOrList($ignorePlatformReqs)` if the arg is not of type `PlatformRequirementFilter`.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -119,24 +121,9 @@ public function setRunScripts($runScripts = true)
...
-    /**
-     * Sets whether platform requirements should be ignored
-     *
-     * If this is set to true, the platform check file will not be generated
-     * If this is set to false, the platform check file will be generated with all requirements
-     * If this is set to string[], those packages will be ignored from the platform check file
-     *
-     * @param array|bool $ignorePlatformReqs
-     */
-    public function setIgnorePlatformRequirements($ignorePlatformReqs)
```
Seldaek, line 131: I think it'd be good to keep the old method around here too for API compatibility, which would I guess simply call `PlatformRequirementFilter::fromBoolOrList($ignorePlatformReqs)`.

Seldaek (changes requested):
IMO it's mostly looking good, thanks for the work!

The few places I commented on it'd be good to keep the old API functional, but the other places are internal-enough that we shouldn't bother. When restoring setters for BC, we could add `trigger_error(E_USER_DEPRECATED...)` to those few places so people notice and migrate away.

Note that for the follow-up of adding wildcard support, using this code is probably a good idea to turn wildcards into regexes, we do this in a few places already: https://github.com/composer/composer/blob/master/src/Composer/Package/BasePackage.php#L251-L256

herndlm:
> IMO it's mostly looking good, thanks for the work!

I'm very glad to hear that, thx!

> The few places I commented on it'd be good to keep the old API functional, but the other places are internal-enough that we shouldn't bother. When restoring setters for BC, we could add `trigger_error(E_USER_DEPRECATED...)` to those few places so people notice and migrate away.

Yes, keeping BC here is luckily as simple as that. I adapted it and added `@deprecated` phpdoc annotations too.

> Note that for the follow-up of adding wildcard support, using this code is probably a good idea to turn wildcards into regexes, we do this in a few places already: https://github.com/composer/composer/blob/master/src/Composer/Package/BasePackage.php#L251-L256

Thx for the hint, I stumbled over that already yesterday and thought of using it (+ adding a few tests). Previously I was looking at [`fnmatch`](https://www.php.net/manual/en/function.fnmatch.php), which is not available on exotic platforms apparently though.

Seldaek (approved):

Seldaek:
I can squash on merge. I'll leave this open until master starts targetting 2.2 though. If you want to do a follow-up PR with the pattern matching you could already open a second one that is based on this branch?

herndlm:
Sorry, rebased with auto-squash after all.. I was afraid this might otherwise mess up the new branch / PR on merge and wanted to start clean there.

Seldaek:
No worries :)

herndlm:
rebased to fix a conflict

helhum (approved):
Nice cleanup imho

imme-emosol:
My comment https://github.com/composer/composer/issues/9534#issuecomment-941293738 might affect this issue.

herndlm:
> My comment [#9534 (comment)](https://github.com/composer/composer/issues/9534#issuecomment-941293738) might affect this issue.

I'm not sure if it does or if I missed something. The intention here was "just" to refactor the platform reqs checks and prepare them to add more functionality like glob like wildcards for which I added another PR. I wasn't even aware of the feature-request regarding upper bounds. Should be independent at least if I'm not mistaken. And would be easier to implement afterwards for sure.

imme-emosol:
> > My comment [#9534 (comment)](https://github.com/composer/composer/issues/9534#issuecomment-941293738) might affect this issue.
> 
> I'm not sure if it does or if I missed something. [...] Should be independent at least if I'm not mistaken. And would be easier to implement afterwards for sure.

You are right in that the code changes made here will probably make things easier anyway indeed.

herndlm:
rebased to fix a confict and added the inheratiance -> composition changes in a fixup as I requested a re-review already anyways

herndlm:
rebased and should be ready again for review

src/Composer/Filter/PlatformRequirementFilter/PlatformRequirementFilterInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Composer\Filter\PlatformRequirementFilter;
+
+interface PlatformRequirementFilterInterface
+{
+    /**
+     * @param string $req
+     * @return bool
+     */
+    public function isReqIgnored($req);
```
Seldaek, line 11: If you don't mind, I think this would be better as simply isIgnored?

```suggestion
    public function isIgnored($req);
```

Seldaek (changes requested):
Looks like a great improvement to me, just the method name nitpick and then happy to finally merge this :)

src/Composer/Filter/PlatformRequirementFilter/PlatformRequirementFilterInterface.php
herndlm: Definitely better, `req` is not needed in this context

Seldaek:
Thanks!

herndlm:
Great thx, I'll rebase and make the wildcard-follow-up feature ready soon then :)

PR #10083, opened 2021-08-29
Add wildcard support to ignore platform req

herndlm:
This is the implementation of what is described in #10045 and the follow-up of #10079.

### DIFF
--- a/src/Composer/Filter/PlatformRequirementFilter/IgnoreListPlatformRequirementFilter.php
+++ b/src/Composer/Filter/PlatformRequirementFilter/IgnoreListPlatformRequirementFilter.php
@@ -2,21 +2,22 @@
 
 namespace Composer\Filter\PlatformRequirementFilter;
 
+use Composer\Package\BasePackage;
 use Composer\Repository\PlatformRepository;
 
 final class IgnoreListPlatformRequirementFilter implements PlatformRequirementFilterInterface
 {
     /**
-     * @var string[]
+     * @var string
      */
-    private $reqList;
+    private $regexp;
 
     /**
      * @param string[] $reqList
      */
     public function __construct(array $reqList)
     {
-        $this->reqList = $reqList;
+        $this->regexp = BasePackage::packageNamesToRegexp($reqList);
     }
 
     /**
@@ -29,6 +30,6 @@ final class IgnoreListPlatformRequirementFilter implements PlatformRequirementFi
             return false;
         }
 
-        return in_array($req, $this->reqList, true);
+        return 1 === preg_match($this->regexp, $req);
     }
 }
--- a/src/Composer/Package/BasePackage.php
+++ b/src/Composer/Package/BasePackage.php
@@ -263,4 +263,23 @@ abstract class BasePackage implements PackageInterface
 
         return sprintf($wrap, $cleanedAllowPattern);
     }
+
+    /**
+     * Build a regexp from package names, expanding * globs as required
+     *
+     * @param string[] $packageNames
+     * @param string $wrap
+     * @return string
+     */
+    public static function packageNamesToRegexp(array $packageNames, $wrap = '{^(?:%s)$}iD')
+    {
+        $packageNames = array_map(
+            function ($packageName) {
+                return BasePackage::packageNameToRegexp($packageName, '%s');
+            },
+            $packageNames
+        );
+
+        return sprintf($wrap, implode('|', $packageNames));
+    }
 }
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -117,7 +117,7 @@ resolution.
   See also the [`platform`](06-config.md#platform) config option.
 * **--ignore-platform-req:** ignore a specific platform requirement(`php`,
   `hhvm`, `lib-*` and `ext-*`) and force the installation even if the local machine
-  does not fulfill it.
+  does not fulfill it. Multiple requirements can be ignored via wildcard.
 
 ## update / u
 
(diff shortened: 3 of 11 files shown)


herndlm:
As #10079 is not merged yet this PR contains that PR's commit too. After it got merged this one needs to be re-targeted.

src/Composer/Filter/PlatformRequirementFilter/IgnoreListPlatformRequirementFilter.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        if (!PlatformRepository::isPlatformPackage($req)) {
+            return false;
+        }
+
+        foreach ($this->reqList as $ignoredReq) {
+            $patternRegexp = BasePackage::packageNameToRegexp($ignoredReq);
+            if (preg_match($patternRegexp, $req)) {
+                return true;
+            }
```
Seldaek, line 37: This could be converted once and combined into one big regex with `|` IMO inside __construct, then here you only have to do a single preg_match per `$req` you need to check.

herndlm: good idea, I adapted it adding a new method in `BasePackage` dealing with a list of packages. `packageNameToRegexp` would not allow `|` because of `preg_quote` which makes sense.

src/Composer/Package/BasePackage.php
```diff
@@ -254,4 +254,23 @@ public static function packageNameToRegexp($allowPattern, $wrap = '{^%s$}i')
...
+     *
+     * @param string[] $packageNames
+     * @param string $wrap
+     * @return string
+     */
+    public static function packageNamesToRegexp($packageNames, $wrap = '{^%s$}i')
+    {
+        $packageNames = array_map(
+            static function ($packageName) {
+                return str_replace('\\*', '.*', preg_quote($packageName));
```
Seldaek, line 269:
```suggestion
                return BasePackage::packageNameToRegexp($packageName, '%s');
```

Then you could also make use of this new function in https://github.com/composer/composer/blob/5b16d61762579fcd86507292789f1c92432d9e8b/src/Composer/Repository/FilterRepository.php#L40-L50 as that essentially does the same :)

herndlm: nice, yes this is a good idea! adapted.

src/Composer/Package/BasePackage.php
```diff
@@ -254,4 +254,23 @@ public static function packageNameToRegexp($allowPattern, $wrap = '{^%s$}i')
...
     }
+
+    /**
+     * Build a regexp from package names, expanding * globs as required
+     *
+     * @param string[] $packageNames
+     * @param string $wrap
+     * @return string
+     */
+    public static function packageNamesToRegexp($packageNames, $wrap = '{^%s$}i')
```
Seldaek, line 265: Last nitpick, this regex makes no sense as it is missing parenthesis it'd only match the ^ with the first package name and $ with the last. Using `{^(?:%s)$}iD` as default here would make more sense.

Looking good to me otherwise, thanks 👍🏻

herndlm: you're right, that would be a problem. got it.

Seldaek (approved):
Thanks, LGTM now, will need a rebase once the other PR is merged but I can do that myself when branching out 2.2 if needed :)

herndlm:
Great, I could not resist cleaning up the fixup commits, I'll leave the rest to you now, except you need me to resolve merge conflicts or so :)

Btw unrelated, but if you need help in increasing the phpstan level or cleanup some more code for phpstan by adding/fixing types and so on, feel free to create issues for that or ping me. I would gladly help out, but will touch too much if you're also working on it

herndlm:
Also: I had to remove the static keyword in the anonymous function in `array_map`. PHP 5.3 was not quite happy with that..

Seldaek:
> Btw unrelated, but if you need help in increasing the phpstan level or cleanup some more code for phpstan by adding/fixing types and so on, feel free to create issues for that or ping me. I would gladly help out, but will not touch too much if you're also working on it

Yeah that would be great, I have been working on typing all properties already, because that gives a good foundation to then type method return values and parameters, which will then force more typing in various places to fix usage warnings surely.

I'm almost done with properties, then maybe I can create an issue with some instructions how to run this and what would be the most valuable, but I'm not entirely sure how much is worth doing right now and how much would be best to wait for us to drop old PHP then we can add proper scalar types instead of adding docblocks now and having to migrate them in a few months. So yeah happy to get help at some point but probably it's too soon :)

herndlm:
rebased to fix a conflict

staabm:
> types instead of adding docblocks now and having to migrate them in a few months

We can automate this step with rector

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -670,7 +675,7 @@ protected function getPathCode(Filesystem $filesystem, $basePath, $vendorPath, $
         return $baseDir . var_export($path, true);
     }
 
-    protected function getPlatformCheck(array $packageMap, array $ignorePlatformReqs, $checkPlatform, array $devPackageNames)
+    protected function getPlatformCheck(array $packageMap, $checkPlatform, array $devPackageNames)
```
staabm, line 678: should this be considered a BC break?

herndlm: This change belongs to #10079 (dependency mess..). I think Jordi said it was fine there, but it's good yo double-check.
Please only review the second commit here though

herndlm:
rebased after the precondition was merged 🎉

Seldaek:
Thanks!

PR #10086, opened 2021-08-31
Docs: Place general troubleshooting steps in a more logical order

iandunn:
* `self-update` should be run first, because it may affect the results of `diagnose`. 
* `clear-cache` should be run before `update`, otherwise `update` would need to be run again after clearing the cache.


Seldaek:
Thanks

PR #10087, opened 2021-09-01
Document archive --format=tar.gz

HenkPoley:
I see that if `extension=bz2` is enabled in php.ini, then archive --format=tar.bz2 works as well.


Seldaek:
Right, feel free to add that to the docs too, and perhaps to the ArchiveCommand argument's description?

PR #10089, opened 2021-09-04
Feature proposal: sorting keys by logical blocks

andrey-helldar:
In continuation of the [discussion](https://github.com/ergebnis/composer-normalize/issues/779), this PR was created.

Only the order of the elements has been changed. Codestyle has remained the same.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -132,27 +132,6 @@
                 }
             }
         },
-        "repositories": {
-            "type": ["object", "array"],
-            "description": "A set of additional repositories where packages can be found.",
-            "additionalProperties": {
-                "anyOf": [
-                    { "$ref": "#/definitions/repository" },
-                    { "type": "boolean", "enum": [false] }
-                ]
-            },
-            "items": {
-                "anyOf": [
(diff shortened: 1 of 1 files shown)


andrey-helldar:
Just attaching the https://github.com/composer/composer/discussions/10073 discussion so as not to search :)

PS: I will make changes in the coming days.

andrey-helldar:
@stof, now I look at the [`dumper code`](https://github.com/composer/composer/blob/main/src/Composer/Package/Dumper/ArrayDumper.php) and understand that there is no point in making changes to it.

If you configure the dumper so that it stores the keys in the order we need, then it will have to be completely rewritten.

On the one hand, this can be done, and on the other, you can use the [`normalizer`](https://github.com/ergebnis/composer-normalize). ¯\_(ツ)_/¯

Seldaek:
OK, seems fine to me, thanks. I agree shifting the whole dumper code isn't really worth it.

andrey-helldar:
Thank you

localheinz:
Thank you, @andrey-helldar!

Issue #8330, opened 2019-09-17
composer archive --format=zip does not preserve permission umask on files


jimmy-ho:
It is noticed when using **composer/satis** that during the archive process to create **dist** of repository tags/branches that permission umask on files that may have an executable flag set is not maintained.

composer/satis simply uses the composer archive functionality where the issue lies.

I have traced this down to the ZipArchiver not using the **setExternalAttributesName** applied on the resultant file in the zip archive.

I have a patch forthcoming to address this issue for review.

Unless someone can explain if this was intentional.

This is related to https://github.com/composer/satis/issues/314

stof:
I don't think this was intentional. Most probably the developer implementing the ZipArchiver was not aware of the `setExternalAttributesName` API (or maybe it was implemented before this API was added in PHP 5.6)

jimmy-ho:
Cool thanks @stof for your feedback, I'll update once I have a PR up for quick review.

stof:
@jimmy-ho make sure to keep the ZipArchiver working on PHP < 5.6 (as the min version of PHP supported by Composer is 5.3), by skipping the `setExternalAttributesName` usage there (it is better to generate the archive as today on old PHP runtimes than to fail to generate it).

jimmy-ho:
Thanks for the callout @stof, will do.

jimmy-ho:
PR for this at: https://github.com/composer/composer/pull/8331

Seldaek:
Fixed by https://github.com/composer/composer/pull/8331

PR #10092, opened 2021-09-08
Add ability check for ZipArchive

SvenRtbg:
ZipArchive::setExternalAttributesName() is only available when libzip 0.11.2 or higher is present. 
This is not the case with at least SUSE SLES 12, so a simple PHP version check will not work here.

As stated in #8330 and fixed in #8331, Composer is adding the file permissions to ZIPs, but the version check is insufficient to detect if the zip extension is powered by a recent enough libzip version. The extension will transparently not provide the called method if libzip is too old, so the `method_exists()` check has to be added.

Should be able to fix the underlying issue in https://github.com/composer/satis/issues/653

### DIFF
--- a/src/Composer/Package/Archiver/ZipArchiver.php
+++ b/src/Composer/Package/Archiver/ZipArchiver.php
@@ -51,6 +51,7 @@ class ZipArchiver implements ArchiverInterface
 
                 /**
                  * ZipArchive::setExternalAttributesName is available from >= PHP 5.6
+                 * setExternalAttributesName() is only available with libzip 0.11.2 or above
                  */
                 if (PHP_VERSION_ID >= 50600 && method_exists($zip, 'setExternalAttributesName')) {
                     $perms = fileperms($filepath);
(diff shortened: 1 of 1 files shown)


src/Composer/Package/Archiver/ZipArchiver.php
```diff
@@ -52,7 +52,7 @@ public function archive($sources, $target, $format, array $excludes = array(), $
                 /**
                  * ZipArchive::setExternalAttributesName is available from >= PHP 5.6
                  */
-                if (PHP_VERSION_ID >= 50600) {
+                if (PHP_VERSION_ID >= 50600 && method_exists($zip, 'setExternalAttributesName')) {
```
staabm, line 55: So we could drop the version check?

SvenRtbg: I have not checked whether the ZipArchive class would react negatively to that attempt - I don't own such an old PHP anymore.

Seldaek:
Thanks

PR #10104, opened 2021-09-15
Changed GitLab's applications URL

vitalybaev:
Currently, Composer shows the following text:
```
To revoke access to this token you can visit https://gitlab.com/profile/applications
```
But this URL returns 404. The correct URL is https://gitlab.com/-/profile/applications

There is another way to get needed credentials in GitLab - [Personal Access Tokens](https://gitlab.com/-/profile/personal_access_tokens) - maybe we should display that URL instead?


Seldaek:
Thanks, makes sense

PR #10107, opened 2021-09-15
Rewrite php file cleaning step to be less regex intensive and support extreme cases better

### LINKED ISSUE #10106: composer 2.1.7 stack overflow/ seg-fault with --optimise-autoloader
My `composer.json`:

```json
{
    "name": "a/test",
    "license": "proprietary",
    "require": {
        "php": "^5.6|^7.4"
    },
    "require-dev": {
        "fzaninotto/faker": "^1.8"
    },
    "type": "project",
    "prefer-stable": true
}
```

```
Checking composer.json: OK
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys: 
OK
Checking composer version: OK
Composer version: 2.1.8
PHP version: 5.6.40
PHP binary path: /usr/bin/php5.6
OpenSSL version: OpenSSL 1.1.1  11 Sep 2018
cURL version: 7.58.0 libz 1.2.11 ssl OpenSSL/1.1.1

```

When I run this command: 

```
composer self-update 2.1.7 # or 2.1.8
composer install --optimize-autoloader -vvv
```

I get the following output:

```
Running 2.1.8 (2021-09-15 13:55:14) with PHP 5.6.40-52+ubuntu18.04.1+deb.sury.org+1 on Linux / 4.14.232-176.381.amzn2.x86_64
Reading ./composer.json (/var/www/erp_dev/composertest/composer.json)
Loading config file ./composer.json (/var/www/erp_dev/composertest/composer.json)
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/var/www/erp_dev/composertest): git branch -a --no-color --no-abbrev -v
Executing command (/var/www/erp_dev/composertest): git rev-list master..test-2021-09
Executing command (/var/www/erp_dev/composertest): git rev-list remotes/origin/master..test-2021-09
Failed to initialize global composer: Composer could not find the config file: /root/.composer/composer.json

Reading /var/www/erp_dev/composertest/vendor/composer/installed.json
Reading ./composer.lock (/var/www/erp_dev/composertest/composer.lock)
Reading ./composer.lock (/var/www/erp_dev/composertest/composer.lock)
Generating rules
Looking at all rules.

Package fzaninotto/faker is abandoned, you should avoid using it. No replacement was suggested.
Segmentation fault (core dumped)
```

So, I know that PHP 5 is now out of support and that Faker is now abandoned, but, providing other people are using PHP 5 (and I suspect there are a few), then this will likely cause an issue.

### Stack Trace

Taken from the core file:
```
#0  0x00007f1da2ef7941 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#1  0x00007f1da2ef99d0 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#2  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#3  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#4  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#5  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#6  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#7  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#8  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#9  0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#10 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#11 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#12 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#13 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#14 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#15 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#16 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#17 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#18 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#19 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#20 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#21 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
#22 0x00007f1da2f05837 in ?? () from /lib/x86_64-linux-gnu/libpcre.so.3
...repeats until...
#16323 0x00007f1da2f087f6 in pcre_exec () from /lib/x86_64-linux-gnu/libpcre.so.3
#16324 0x00005616938d97f8 in php_pcre_replace_impl ()
#16325 0x00005616938da7cb in php_pcre_replace ()
```

Looks like an infinite recursion/ stack overflow issue and I note there were a few `preg_replace` optimisations in the [update from 2.1.6](https://github.com/composer/composer/compare/2.1.6...2.1.7)

### It Works Without `--optimise-autoloader`

```bash
composer install

    Package fzaninotto/faker is abandoned, you should avoid using it. No replacement was suggested.
    Generating autoload files
```

### It Works In 2.1.6:

```bash
composer self-update 2.1.6

    Upgrading to version 2.1.6 (stable channel).

    Use composer self-update --rollback to return to version 2.1.7

composer install --optimize-autoloader

    Package fzaninotto/faker is abandoned, you should avoid using it. No replacement was suggested.
```

It's the `--optimize-autoloader` phase which causes the seg-fault.

This same issue is also present in the latest `2.1.8` version, as well:

```bash
composer self-update

    Upgrading to version 2.1.8 (stable channel).

    Use composer self-update --rollback to return to version 2.1.7

composer install --optimize-autoloader

    Package fzaninotto/faker is abandoned, you should avoid using it. No replacement was suggested.
    Segmentation fault (core dumped)
```

### Issue not present in PHP 7+

I tried the above with PHP 7.4 and no issue exists.

Seldaek:
Fixes #10106

The only downside here as far as I can tell is a slight slowdown, I see a `composer dump-autoload -o --profile` currently taking around 1.3x as long as it did before. The upside is that it's passing the whole test suite while the old code doesn't.

This means it's fine as is IMO, but **if someone wants to poke at it with a profiler to try to extract some last bits of perf, that'd be great**! I already did what I could to remove obvious unnecessary allocation and co but didn't have time to profile.

/cc @jrfnl this last issue was finally the one drop too many and I got rid of the whole thing, happy to get your review here if you have time

### DIFF
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -214,10 +214,7 @@ class ClassMapGenerator
      */
     private static function findClasses($path)
     {
-        $extraTypes = PHP_VERSION_ID < 50400 ? '' : '|trait';
-        if (PHP_VERSION_ID >= 80100 || (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.3', '>='))) {
-            $extraTypes .= '|enum';
-        }
+        $extraTypes = self::getExtraTypes();
 
         // Use @ here instead of Silencer to actively suppress 'unhelpful' output
         // @link https://github.com/composer/composer/pull/4886
@@ -241,57 +238,14 @@ class ClassMapGenerator
         }
 
         // return early if there is no chance of matching anything in this file
-        if (!preg_match('{\b(?:class|interface'.$extraTypes.')\s}i', $contents)) {
+        preg_match_all('{\b(?:class|interface'.$extraTypes.')\s}i', $contents, $matches);
+        if (!$matches) {
             return array();
         }
 
-        // strip heredocs/nowdocs
-        $heredocRegex = '{
-            # opening heredoc/nowdoc delimiter (word-chars)
-            <<<[ \t]*+([\'"]?)(\w++)\\1
-            # needs to be followed by a newline
-            (?:\r\n|\n|\r)
-            # the meat of it, matching line by line until end delimiter
-            (?:
-                # a valid line is optional white-space (possessive match) not followed by the end delimiter, then anything goes for the rest of the line
-                [\t ]*+(?!\\2 \b)[^\r\n]*+
-                # end of line(s)
-                [\r\n]++
-            )*
-            # end delimiter
-            [\t ]*+ \\2 (?=\b)
-        }x';
-
-        // run first assuming the file is valid unicode
-        $contentWithoutHeredoc = preg_replace($heredocRegex.'u', 'null', $contents);
-        if (null === $contentWithoutHeredoc) {
-            // run again without unicode support if the file failed to be parsed
-            $contents = preg_replace($heredocRegex, 'null', $contents);
-        } else {
-            $contents = $contentWithoutHeredoc;
-        }
-        unset($contentWithoutHeredoc);
-
-        // strip strings
-        $contents = preg_replace('{"[^"\\\\]*+(\\\\.[^"\\\\]*+)*+"|\'[^\'\\\\]*+(\\\\.[^\'\\\\]*+)*+\'}s', 'null', $contents);
-        // strip leading non-php code if needed
-        if (strpos($contents, '<?') !== 0) {
-            $contents = preg_replace('{^.+?<\?}s', '<?', $contents, 1, $replacements);
-            if ($replacements === 0) {
-                return array();
-            }
-        }
-        // strip non-php blocks in the file
-        $contents = preg_replace('{\?>(?:[^<]++|<(?!\?))*+<\?}s', '?><?', $contents);
-        // strip trailing non-php code if needed
-        $pos = strrpos($contents, '?>');
-        if (false !== $pos && false === strpos(substr($contents, $pos), '<?')) {
-            $contents = substr($contents, 0, $pos);
-        }
-        // strip comments if short open tags are in the file
-        if (preg_match('{(<\?)(?!(php|hh))}i', $contents)) {
-            $contents = preg_replace('{//.* | /\*(?:[^*]++|\*(?!/))*\*/}x', '', $contents);
-        }
+        $p = new PhpFileCleaner($contents, count($matches[0]));
+        $contents = $p->clean();
+        unset($p);
 
         preg_match_all('{
             (?:
@@ -327,5 +281,19 @@ class ClassMapGenerator
         }
 
         return $classes;
+    }
+
+    private static function getExtraTypes()
+    {
+        static $extraTypes = null;
+        if (null === $extraTypes) {
+            $extraTypes = PHP_VERSION_ID < 50400 ? '' : '|trait';
+            if (PHP_VERSION_ID >= 80100 || (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.3', '>='))) {
+                $extraTypes .= '|enum';
+            }
+            PhpFileCleaner::setTypeConfig(array_merge(array('class', 'interface'), array_filter(explode('|', $extraTypes))));
+        }
+
+        return $extraTypes;
     }
 }
--- a/src/Composer/Autoload/PhpFileCleaner.php
+++ b/src/Composer/Autoload/PhpFileCleaner.php
@@ -1,0 +1,228 @@
+<?php
+
+namespace Composer\Autoload;
+
+/**
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ * @internal
+ */
+class PhpFileCleaner
+{
+    /** @var array<array{name: string, length: int, pattern: string}> */
+    private static $typeConfig;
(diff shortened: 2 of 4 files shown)


Seldaek:
cc @fredemmott I am assuming the `(?<![\$:>])` bit in this is due to HHVM/Hack syntax 

https://github.com/composer/composer/blob/ffee8ca329884a23b5bb0624b83901f89d76547c/src/Composer/Autoload/ClassMapGenerator.php#L298-L299

Is this still needed? I'm not entirely sure how this would behave with the new parsing code but I suspect not so well. Do you use composer autoloading at all with HHVM?

jrfnl:
@Seldaek Happy to take a look, but may be a little before I get to it.

Seldaek:
@mvorisek btw this also addresses the problem you raised in the other PR as it does all the cleaning in one pass it can't mistake a heredoc inside a string or the like.

fredemmott:
I got an email notification here, but not seeing the mention here now

re Hack autoloading, we don't actually use Composer's autoloading, but I think we depend on composer not erroring if it sees .php files that contain syntax errors (especially if they start with `<?hh`, maybe after shebang).

There's currently two common approaches to autoloading in hack:
- use [a separate project](https://github.com/hhvm/hhvm-autoload) that installs a plugin that hooks into `ScriptEvents::POST_AUTOLOAD_DUMP` to trigger
- an autoloader built into HHVM

We hope to entirely move to the built-in autoloader soon and stop using a composer plugin, but usability is a bit too much of a concern at the moment as it requires external services (watchman) to be configured.

Seldaek:
@fredemmott I deleted my comment because I realized what the code I linked to was about and it had nothing to do with HHVM :) Thanks for the clarifications tho, that should be no problem tho with the new parser I imagine.

mvorisek:
Are the backtrack limits really the issue? If they can be fixed, one combined `pcre_replace` will be always faster as compiled than looping manually in php.

Seldaek:
@mvorisek  Well would you consider a segfault an issue? (see #10106) I would.. I doubt it can be fixed, and while this PR is indeed a tiny bit slower, it's not that bad. Feel free to give it a shot though if you like.

mvorisek:
> 
> 
> @mvorisek Well would you consider a segfault an issue? (see #10106) I would.. I doubt it can be fixed, and while this PR is indeed a tiny bit slower, it's not that bad. Feel free to give it a shot though if you like.

I see - segfault with regex introduced in https://github.com/composer/composer/pull/10072 with PHP 5.6 only. Long time ago, I have seen some issues with possessive quantifiers (like `*+`), please test without them, php should be smart enough to optimize things like `\w*+\W*+` automatically. If this does not help, the regex should be tweaked further or fixed like with this PR, but for PHP 5.x only, as it will never be faster.

https://github.com/composer/composer/issues/10106#issuecomment-920089925 is interesting, but I am not sure if high step count always mean a slower performance.

ondrejmirtes:
The approach in master also silently fails on PHP 7.1 and PHP 7.2 so this new approach is needed and welcome.

mvorisek:
> 
> 
> The approach in master also silently fails on PHP 7.1 and PHP 7.2 so this new approach is needed and welcome.

Please link (or open a new) PHP issue for it. PHP guys might know what is causing the problems.

mvorisek:
Also, I wonder how much the test file is compressible as the raw size is ~1 MB which might slow git clones a little bit. Replacing the long texts with some repeated paragraphs might hugely decreate it's compressed size.

ondrejmirtes:
I've talked to @Seldaek about it, it's the same file that's causing the problem in this PR.

mvorisek:
I mean what exactly is causing php regex engine to fail, eg. how to tweak the regex, not the input data.

Seldaek:
> Also, I wonder how much the test file is compressible as the raw size is ~1 MB which might slow git clones a little bit. Replacing the long texts with some repeated paragraphs might hugely decreate it's compressed size.

I removed the file, it didn't really add a ton of value given the new solution does not have a regex, it can't really fail due to the input size.

PR #10110, opened 2021-09-17
tests for solver handling multiple packages with same name and version

dzuelke:
Not a case with regular Packagist packages, but some custom installers with custom repos do this, and the current behavior should not randomly change at some point, as that would cause downstream breakage.

As discussed with / requested by @naderman


dzuelke:
(I can also change these to live in `test/Fixtures/installer/` instead if you prefer that over the code-test approach taken here)

Seldaek:
Looks good thanks!

PR #10118, opened 2021-09-22
Don't revert composer.json when "require" fails because of scripts

nicolas-grekas:
Fix eg https://github.com/symfony/flex/issues/751


Seldaek:
I agree we should fix this as it's annoying, but I wonder if we could target only post-update scripts failures somehow.. Because if some script fails earlier than when the update is complete, we probably still want to revert.

dzuelke:
@nicolas-grekas could you please check that https://github.com/composer/composer/pull/10541 does not break this (and https://github.com/symfony/flex/issues/142) again?

Issue #10085, opened 2021-08-31
Debug output add where authentication was loaded from


glaubinix:
Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (2.1.5 => 2.1.6)
Composer version: 2.1.5
PHP version: 7.4.19
PHP binary path: /usr/local/Cellar/php@7.4/7.4.19_1/bin/php
OpenSSL version: OpenSSL 1.1.1k  25 Mar 2021
cURL version: 7.77.0 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1k
zip: extension present, unzip present, 7-Zip not available
```

When I run this command: 

```
any command with -vvv that does HTTP requests e.g. composer oudated -vvv
```

I get the following output: 

```
Reading ./composer.json (/Users/glaubinix/Projects/example/composer.json)
Loading config file /Users/glaubinix/.composer/config.json
Loading config file /Users/glaubinix/.composer/auth.json
Loading config file ./composer.json (/Users/glaubinix/Projects/example/composer.json)
Loading config file /Users/glaubinix/Projects/example/auth.json
Reading /Users/glaubinix/Projects/example/auth.json
Checked CA file /usr/local/etc/openssl@1.1/cert.pem: valid
Executing command (/Users/glaubinix/Projects/example): git branch -a --no-color --no-abbrev -v
Reading /Users/glaubinix/.composer/composer.json
Loading config file /Users/glaubinix/.composer/config.json
Loading config file /Users/glaubinix/.composer/auth.json
Loading config file /Users/glaubinix/.composer/composer.json (/Users/glaubinix/.composer/composer.json)
Loading config file /Users/glaubinix/.composer/auth.json
Reading /Users/glaubinix/.composer/auth.json
Reading /Users/glaubinix/Projects/example/vendor/composer/installed.json
[...]

PR #10129, opened 2021-09-29
Config: add source option for command to show where a config value is loaded from

glaubinix:
Implements: https://github.com/composer/composer/issues/10085

Can be used either to output a specific config e.g.
```
composer config http-basic.repo.example.org.username --source
glaubinix (/Users/glaubinix/.composer/auth.json)
```

or can be used with listing the entire configuration e.g. 
```
composer config -l --source
[process-timeout] 300 (default)
[use-include-path] false (default)
[preferred-install] dist (default)
[notify-on-install] true (default)
[github-protocols] [https, ssh] (default)
[gitlab-protocol]  (default)
[vendor-dir] vendor (/Users/glaubinix/Projects/example/vendor) (default)
[bin-dir] bin (/Users/glaubinix/Projects/example/bin) (./composer.json)
```

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -77,6 +77,7 @@ class ConfigCommand extends BaseCommand
                 new InputOption('json', 'j', InputOption::VALUE_NONE, 'JSON decode the setting value, to be used with extra.* keys'),
                 new InputOption('merge', 'm', InputOption::VALUE_NONE, 'Merge the setting value with the current value, to be used with extra.* keys in combination with --json'),
                 new InputOption('append', null, InputOption::VALUE_NONE, 'When adding a repository, append it (lowest priority) to the existing ones instead of prepending it (highest priority)'),
+                new InputOption('source', null, InputOption::VALUE_NONE, 'Display where the config value is loaded from'),
                 new InputArgument('setting-key', null, 'Setting key'),
                 new InputArgument('setting-value', InputArgument::IS_ARRAY, 'Setting value'),
             ))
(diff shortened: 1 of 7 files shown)


Seldaek:
@glaubinix can you please rebase this?

glaubinix:
done

src/Composer/Config.php
```diff
@@ -150,16 +164,18 @@ public function getAuthConfigSource()
...
+    public function merge($config, $source = self::SOURCE_UNKNOWN)
     {
         // override defaults with given config
         if (!empty($config['config']) && is_array($config['config'])) {
             foreach ($config['config'] as $key => $val) {
                 if (in_array($key, array('bitbucket-oauth', 'github-oauth', 'gitlab-oauth', 'gitlab-token', 'http-basic', 'bearer')) && isset($this->config[$key])) {
                     $this->config[$key] = array_merge($this->config[$key], $val);
+                    $this->setSourceOfConfigValue($val, $key, $source);
                 } elseif (in_array($key, array('gitlab-domains', 'github-domains')) && isset($this->config[$key])) {
                     $this->config[$key] = array_unique(array_merge($this->config[$key], $val));
```
Seldaek, line 180: Why not setting the source here? 

It also raises the question what to do when things are merged like this.. Would it make sense to list multiple sources where it came from? Is this overkill?

glaubinix: I think I skipped this first because I wasn't sure how to best handle this, and then forgot about it. The fields I am interested in are non-array fields, so no strong preference.  Could start with having the source as "from where was the value last edited" and if anyone needs more info we can add a list of sources that edited the array field?

Seldaek: Yeah that sounds ok for now, it definitely isn't the most useful one to be looking at this one anyway. So please add a simple update and then looks good to me 👍🏻

src/Composer/Command/ConfigCommand.php
```diff
@@ -77,6 +77,7 @@ protected function configure()
                 new InputOption('json', 'j', InputOption::VALUE_NONE, 'JSON decode the setting value, to be used with extra.* keys'),
                 new InputOption('merge', 'm', InputOption::VALUE_NONE, 'Merge the setting value with the current value, to be used with extra.* keys in combination with --json'),
                 new InputOption('append', null, InputOption::VALUE_NONE, 'When adding a repository, append it (lowest priority) to the existing ones instead of prepending it (highest priority)'),
+                new InputOption('source', null, InputOption::VALUE_NONE, 'Display from where the config value is loaded from'),
```
Seldaek, line 80: Ah also please add the new flag to the 03-cli.md docs

Seldaek: Also double from => `Display where the config value is loaded from` :)

src/Composer/Config.php
glaubinix: added

src/Composer/Command/ConfigCommand.php
glaubinix: Added and also added --append which was missing

Seldaek:
Thanks, I referenced it in the troubleshooting page too as it seems that's the main use case :) https://github.com/composer/composer/commit/8542321a31a129398eebb87f819b33c15615fc60

PR #10132, opened 2021-09-30
VcsRepository: do not continue when receiving 429 rate limit exception

glaubinix:
Currently Composer ignores all TransportExceptions (except 401 and 403) while fetching composer.json files to import all available versions and then skips those versions.

This pull request changes this behaviour for rate limit errors with a 429 status code. After a rate limit error response was received all future requests to that resource will most likely receive the same error status code which make it unlikely that Composer is able to finish the command.

Sample Composer output for such a scenario:
```
Reading composer.json of acme/package (2.0.0) 
Downloading https://api.bitbucket.org/2.0/repositories/acme/package/src/hash/composer.json
[429] https://api.bitbucket.org/2.0/repositories/acme/package/src/hash/composer.json
Skipped tag 2.0.0, no composer file was found (429 HTTP status code)
Reading composer.json of acme/package (2.0.1) 
Downloading https://api.bitbucket.org/2.0/repositories/acme/package/src/hash/composer.json
[429] https://api.bitbucket.org/2.0/repositories/acme/package/src/hash/composer.json
Skipped tag 2.0.1, no composer file was found (429 HTTP status code)
Reading composer.json of acme/package (2.0.2) 
Downloading https://api.bitbucket.org/2.0/repositories/acme/package/src/hash/composer.json
[429] https://api.bitbucket.org/2.0/repositories/acme/package/src/hash/composer.json
Skipped tag 2.0.2, no composer file was found (429 HTTP status code)
```

I do wonder whether the behaviour in the VcsRepository should be changed in general and abort for most errors including 5XX errors and curl errors. A VcsRepository that doesn't receive all available versions can lead to an unexpected result of a Composer command e.g. composer update might unexpectedly downgrade a package because the request to fetch the composer.json file of the latest tag returned a 500 response.

### DIFF
--- a/src/Composer/Repository/VcsRepository.php
+++ b/src/Composer/Repository/VcsRepository.php
@@ -282,7 +282,7 @@ class VcsRepository extends ArrayRepository implements ConfigurableRepositoryInt
                     if ($e->getCode() === 404) {
                         $this->emptyReferences[] = $identifier;
                     }
-                    if ($e->getCode() === 401 || $e->getCode() === 403) {
+                    if (in_array($e->getCode(), array(401, 403, 429), true)) {
                         throw $e;
                     }
                 }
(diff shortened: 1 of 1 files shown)


Seldaek:
Yeah I guess the odds of github starting to return 500s in the middle of the update is low enough and it'll break enough other things that the update most likely wouldn't complete anyway, but technically maybe we could include all 5xx. Will merge this already though as the rate limit is definitely a more likely scenario and it makes sense for sure.

Issue #10119, opened 2021-09-24
Package PHP binaries can not include the correct autoload.php, when installed via PathRepository as symlink


helhum:
My `composer.json`:

```json
{
    "repositories": {
        "packages": {
            "type": "path",
            "url": "packages/*"
        }
    },
    "require": {
         "vendor/package-with-bin": "@dev"
    }
}
```

When I run this command:

```
composer update
```

I will get the `vendor/package-with-bin` installed in vendor folder as a symlink, to `packages/` folder.

Now trying to execute the binary (which is linked to `vendor/bin`), the PHP binary usually includes
the autoload.php with code like this:

```php
$files       = [__DIR__ . '/../vendor/autoload.php', __DIR__ . '/../../../autoload.php'];
$loader      = null;

foreach ($files as $file) {
    if (file_exists($file)) {
        $loader = require $file;

        break;
    }
}

```

Now, sine `__DIR__` resolves to `packages/package-with-bin`, the code above fails, as it expects the code to be installed in the vendor folder.

And I expected this to happen:

Currently there is no way for packages with a PHP binary can reliably obtain the path to the correct `autoload.php`, when they are not directly installed (and not symlinked) into the vendor folder.

The only idea I can come up with is to:

1. Always create proxy files for PHP enabled binaries
2. Expose the path to the `autoload.php` in the proxy, which the PHP binary scripts can access

e.g. something like this

```php
#!/usr/bin/env php
<?php

# Proxy file created by Composer

$GLOBALS['__composer_autoload_path'] = __DIR__ . '/../autoload.php';
require __DIR__ . '/../vendor/package-with-bin/php-script';
```

in `vendor/package-with-bin/php-script` it could be checked for `$GLOBALS['__composer_autoload_path']`
and the autoload.php could be included from there like:

```php

$files       = [__DIR__ . '/../vendor/autoload.php', $GLOBALS['__composer_autoload_path'] ?? __DIR__ . '/../../../autoload.php'];
$loader      = null;

foreach ($files as $file) {
    if (file_exists($file)) {
        $loader = require $file;

        break;
    }
}

```

Any thoughts on that?

I'm happy to provide a PR in case this is considered fixing.

Seldaek:
Always creating a proxy isn't as easy as it seems, the current code has to jump through some hoops to get rid of shebangs on top of php files: https://github.com/composer/composer/blob/master/src/Composer/Installer/BinaryInstaller.php#L220-L242

[...]

PR #10137, opened 2021-10-03
Expose path to autoload in a global var for binaries

helhum:
For binaries of packages to be able include the autoload file,
a global variable is now exposed and a proxy file now generated
by default, which exposes the path and includes the original
binary file.

Additionally it is now checked on binary creation whether
the reference binary has a shebang and if not, generates
a much simple proxy code with no magic and eval involved.

Fixes: #10119

* [x] Decide whether we can rely on `vendor-dir` being injected, or what to do when the object is created by other installers not providing it.
* [x] Should symlink creation for binaries remain or be removed?
* [x] Depending on the decision above, maybe add code to migrate from symlinks to proxies, when symlinks are found

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -98,17 +98,14 @@ class BinaryInstaller
                 }
             }
 
-            if ($this->binCompat === "auto") {
-                if (Platform::isWindows() || Platform::isWindowsSubsystemForLinux()) {
-                    $this->installFullBinaries($binPath, $link, $bin, $package);
-                } else {
-                    $this->installUnixyProxyBinaries($binPath, $link);
-                }
-            } elseif ($this->binCompat === "full") {
+            $binCompat = $this->binCompat;
+            if ($binCompat === "auto" && (Platform::isWindows() || Platform::isWindowsSubsystemForLinux())) {
+                $binCompat = 'full';
+            }
+
+            if ($this->binCompat === "full") {
                 $this->installFullBinaries($binPath, $link, $bin, $package);
-            } elseif ($this->binCompat === "symlink") {
-                $this->installUnixyProxyBinaries($binPath, $link);
-            } elseif ($this->binCompat === "proxy") {
+            } else {
                 $this->installUnixyProxyBinaries($binPath, $link);
             }
             Silencer::call('chmod', $binPath, 0777 & ~umask());
@@ -225,6 +222,16 @@ class BinaryInstaller
     {
         $binPath = $this->filesystem->findShortestPath($link, $bin);
         $caller = self::determineBinaryCaller($bin);
+
+        // if the target is a php file, we run the unixy proxy file
+        // to ensure that _composer_autoload_path gets defined, instead
+        // of running the binary directly
+        if ($caller === 'php') {
+            return "@ECHO OFF\r\n".
+                "setlocal DISABLEDELAYEDEXPANSION\r\n".
+                "SET BIN_TARGET=%~dp0/".trim(ProcessExecutor::escape(basename($link, '.bat')), '"\'')."\r\n".
+                "{$caller} \"%BIN_TARGET%\" %*\r\n";
+        }
 
         return "@ECHO OFF\r\n".
             "setlocal DISABLEDELAYEDEXPANSION\r\n".
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -36,19 +36,23 @@ class BinaryInstaller
     protected $io;
     /** @var Filesystem */
     protected $filesystem;
+    /** @var string|null */
+    private $vendorDir;
 
     /**
      * @param IOInterface $io
      * @param string      $binDir
      * @param string      $binCompat
      * @param Filesystem  $filesystem
+     * @param string|null $vendorDir
      */
-    public function __construct(IOInterface $io, $binDir, $binCompat, Filesystem $filesystem = null)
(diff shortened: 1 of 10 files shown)


helhum:
@Seldaek here is my first shot. I will add a few todos, which came to my mind for things to be discussed

src/Composer/Installer/BinaryInstaller.php
```diff
@@ -92,12 +96,14 @@ public function installBinaries(PackageInterface $package, $installPath, $warnOn
                 if (Platform::isWindows() || Platform::isWindowsSubsystemForLinux()) {
                     $this->installFullBinaries($binPath, $link, $bin, $package);
                 } else {
-                    $this->installSymlinkBinaries($binPath, $link);
+                    $this->installUnixyProxyBinaries($binPath, $link);
                 }
             } elseif ($this->binCompat === "full") {
                 $this->installFullBinaries($binPath, $link, $bin, $package);
             } elseif ($this->binCompat === "symlink") {
                 $this->installSymlinkBinaries($binPath, $link);
```
Seldaek, line 104: I think I'd treat `symlink` the same as `proxy`, and get rid of `installSymlinkBinaries` altogether, then at least we get some simplification out of this.

src/Composer/Installer/BinaryInstaller.php
```diff
@@ -237,7 +237,22 @@ protected function generateUnixyProxyCode($bin, $link)
...
+<?php
+
+/**
+ * Proxy PHP file generated by Composer
+ *
+ * This file includes the referenced bin path ($binPath)$evalDescription
+ *
+ * @generated
+ */
+\$GLOBALS['_composer_autoload_path'] = $autoloadPathCode;$evalCode
```
Seldaek, line 254: I wonder if we also want to expose the root dir i.e. `getcwd()` and perhaps simply expose the vendor dir instead of path to autoload.php?

Seldaek (changes requested):
Thanks for the PR.

- This needs a rebase for sure, as the new code is very different (sorry:).
- You should bump the `COMPOSER_RUNTIME_API` to 2.2.0 in `Composer\Composer` as this adds a new project-runtime feature that packages may want to depend on.
- This also raises the question to me of what happens if the vendor-dir changes, and the bin-dir is not inside it, then bin proxies will not get re-generated and they'll suddenly contain incorrect information? We probably should detect/check that and reinstall the proxies in that case? Complexity++ though..

helhum:
Thanks @Seldaek for looking into this!

> This also raises the question to me of what happens if the vendor-dir changes, and the bin-dir is not inside it, then bin proxies will not get re-generated and they'll suddenly contain incorrect information? We probably should detect/check that and reinstall the proxies in that case? Complexity++ though..

Hm, isn't that the same with Symlinks? Since Symlinks are relative they will break as well wenn changing bin folder relative to the vendor folder. I'd always considered this acceptable and removed the old bin folder to force a re-install.

> This needs a rebase for sure, as the new code is very different (sorry:).

Will do, no problem!

> You should bump the COMPOSER_RUNTIME_API to 2.2.0 in Composer\Composer as this adds a new project-runtime feature that packages may want to depend on.

Yeah, good idea, will do.

src/Composer/Installer/BinaryInstaller.php
helhum: Makes sense. I will remove this code then and replace it with proxy code. I'd suggest to also change the config then to "full" and "proxy" and only keep "symlink" in the config for BC reasons.

src/Composer/Installer/BinaryInstaller.php
helhum: Hm, not sure. In general I'd say root-dir and vendor-dir *could* be useful during runtime, but I'd rather expose them in `InstalledVersions`. Root dir is btw indirectly exposed already, as the "install path" of the root package equals the root dir.

vendor-dir is indeed missing and could only be indirectly derived by using `dirname(InstalledVersions::getInstallPath('my/package-in-vendor'), 2)`, but I would not expose any of those paths in a global var in the proxy file of binaries.

Seldaek: Alright, let's leave it as is.

src/Composer/Installer/BinaryInstaller.php
Seldaek: Right symlink should just remain supported but silently switch to behave as proxy

Seldaek:
> Hm, isn't that the same with Symlinks? Since Symlinks are relative they will break as well wenn changing bin folder relative to the vendor folder. I'd always considered this acceptable and removed the old bin folder to force a re-install.

I guess you're right :) If we could live with it until now we can probably leave it as is.

src/Composer/Installer/BinaryInstaller.php
helhum: Done

helhum:
Thanks for the additional fixes @Seldaek! If I can be of any additional help, let me know.

Seldaek:
Heh no worries, it's a case of the "the more you look at it the more problems you find" :D 

Wrote some docs too, and fixed more edge cases. I believe it's ready to go now.

Seldaek (approved):
LGTM now. Happy if anyone else can review though.

helhum:
> Heh no worries, it's a case of the "the more you look at it the more problems you find" :D

Yeah. Although I know Composer pretty well, the things you now fixed weren't obvious to me. I will e.g. try to keep in mind that docs are part of this repo too. 👍

> Wrote some docs too, and fixed more edge cases. I believe it's ready to go now.

 Awesome! Thanks again!

PR #10148, opened 2021-10-11
Add a config option to use shallow git clone

wissem:
Large Bitbucket repositories cannot be downloaded via ZIP download because Bitbucket throws a 403 HTTP error with the message "You can't download this repository because it's too large". The fallback to git clone also fails because the repo is too large and it's downloading all of its history.
Using git shallow clone helps in this case since at least it only downloads the latest commits and it doesn't attempt to download the entire repo history.
By the default the new config is set to false so this would continue to work with git clone by default.


Seldaek:
@wissem can you please rebase this?

wissem:
@Seldaek done

src/Composer/Util/Git.php
```diff
@@ -300,11 +301,25 @@ public function syncMirror($url, $dir)
         // clean up directory and do a fresh clone into it
         $this->filesystem->removeDirectory($dir);
 
-        $commandCallable = function ($url) use ($dir) {
-            return sprintf('git clone --mirror -- %s %s', ProcessExecutor::escape($url), ProcessExecutor::escape($dir));
-        };
+        if ($shallowClone) {
+            $commandCallable = function ($url) use ($dir) {
+                return sprintf('git clone --depth=1 --mirror -- %s %s && git remote set-branches origin "*"', ProcessExecutor::escape($url), ProcessExecutor::escape($dir));
```
Seldaek, line 306: Sorry but I'm not familiar with this here, why is it doing `git remote set-branches origin "*"` in addition to the clone? And why the additional fetch --depth=1 below? Is that to force fetching one commit of every branch?

In any case, this needs cross-platform escaping:

```suggestion
                return sprintf('git clone --depth=1 --mirror -- %s %s && git remote set-branches origin -- %s', ProcessExecutor::escape($url), ProcessExecutor::escape($dir), ProcessExecutor::escape('*'));
```

wissem: `git remote set-branches origin "*"`  is to fetch all the remote branches, without it I only got the default branch and the additional fetch is to fetch one commit for every branch. I'm also not super with how this works but found this https://stackoverflow.com/questions/23708231/git-shallow-clone-clone-depth-misses-remote-branches/27393574#27393574 while searching for how to fetch all branches

Seldaek: OK thanks for confirming, looks good to me then, thanks!

Seldaek:
Actually closing as this is not really workable since it breaks installs when they're done from a shallow clone.

Issue #10147, opened 2021-10-11
symfony recipes : composer recipes vs composer help "recipes"


Jduret:
I'm trying to use your composer Docker image. My project use Symfony and the Symfony/Flex Composer extension.
I use a shortcut to run composer command : 
```sh
#/usr/local/bin/composer
sudo docker run --rm --interactive --tty \
--volume $PWD:/app \
--volume ${COMPOSER_HOME:-$HOME/.composer}:/tmp \
composer "$@"
```
My issue occured when I try to call `composer recipes` under my project. 
> /docker-entrypoint.sh: exec: line 24: recipes: not found

So I open a bash in the container.
```
docker run -it --volume $PWD:/app \
--entrypoint /bin/bash composer
```
And then I ran the command : `composer recipes`
```
 bash-5.1# composer recipes

 Available recipes.

 * doctrine/annotations
  * doctrine/doctrine-bundle (update available)
 * doctrine/doctrine-migrations-bundle (update available)
 * symfony/apache-pack (update available)
 * symfony/console (update available)
 * symfony/flex
 * symfony/framework-bundle (update available)
 * symfony/maker-bundle
 * symfony/monolog-bundle (update available)
 * symfony/phpunit-bridge (update available)
 * symfony/routing (update available)
 * symfony/twig-bundle (update available)
 * symfony/validator (update available)
 * symfony/web-profiler-bundle

Run:
 * composer recipes vendor/package to see details about a recipe.
 * composer recipes:install vendor/package --force -v to update that recipe.
 ```
So the error is not from composer but from the entrypoint. 
In the entrypoint I found this : 
```sh
isCommand() {
  # Retain backwards compatibility with common CI providers,
  # see: https://github.com/composer/docker/issues/107
  if [ "$1" = "sh" ]; then
    return 1
  fi

  composer help "$1" > /dev/null 2>&1
}
```
So I ran this command  : `composer help "recipes"`
```
bash-5.1# composer help "recipes"

  [Symfony\Component\Console\Exception\CommandNotFoundException]
  Command "recipes" is not defined.

help [--xml] [--format FORMAT] [--raw] [--] [<command_name>]
```
Can you explain me why `composer recipes` work and `composer help "recipes"` don't ?

I've already tried the commande `composer help "recipes"` using a local composer and it works

alcohol:
No idea if relevant or related as well, but running `composer list symfony` lists several commands from the symfony namespace, but running `composer list recipes` fails:

```
[...]

PR #10150, opened 2021-10-12
[ISSUE - #10147] Add "help" in the list of command that need plugins

Jduret:
[See issue](https://github.com/composer/composer/issues/10147)
- Add 'help' in the list of commands that need plugin command ($mayNeedPluginCommand)
- Remove `'global' !== $commandName` because `$mayNeedPluginCommand` ensure that `$commandName` is one of `false`, `empty`, `list` or `help` (so never 'global')


Seldaek:
Thanks

PR #10157, opened 2021-10-14
Prevent auto-unlocked path repo packages from also unlocking their transitive deps when -w/-W is used

Seldaek:
Refs https://github.com/composer/composer/discussions/10093

As @stof said: 

> @Seldaek maybe this forced inclusion of path packages should be done after building the whitelist with -w/-W, so that the path packages don't automatically whitelist their dependencies when this option are used (they would still whitelist their dependencies if the path packages are explicitly marked for update in the whitelist in the command)

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -165,9 +165,18 @@ class PoolBuilder
 
             foreach ($request->getLockedRepository()->getPackages() as $lockedPackage) {
                 if (!$this->isUpdateAllowed($lockedPackage)) {
+                    $lockedName = $lockedPackage->getName();
+
+                    // remember which packages we skipped loading remote content for in this partial update
+                    $this->skippedLoad[$lockedPackage->getName()][] = $lockedPackage;
+                    foreach ($lockedPackage->getReplaces() as $link) {
+                        $this->skippedLoad[$link->getTarget()][] = $lockedPackage;
+                    }
+
                     // Path repo packages are never loaded from lock, to force them to always remain in sync
                     // unless symlinking is disabled in which case we probably should rather treat them like
-                    // regular packages
+                    // regular packages. We mark them specially so they can be reloaded fully including update propagation
+                    // if they do get unlocked, but by default they are unlocked without update propagation.
                     if ($lockedPackage->getDistType() === 'path') {
                         $transportOptions = $lockedPackage->getTransportOptions();
                         if (!isset($transportOptions['symlink']) || $transportOptions['symlink'] !== false) {
@@ -177,11 +186,6 @@ class PoolBuilder
                     }
 
                     $request->lockPackage($lockedPackage);
-                    $this->skippedLoad[$lockedPackage->getName()][] = $lockedPackage;
-                    // remember which packages we skipped loading remote content for in this partial update
-                    foreach ($lockedPackage->getReplaces() as $link) {
-                        $this->skippedLoad[$link->getTarget()][] = $lockedPackage;
-                    }
                 }
             }
         }
@@ -204,7 +208,7 @@ class PoolBuilder
                 || $package->getRepository() instanceof PlatformRepository
                 || StabilityFilter::isPackageAcceptable($this->acceptableStabilities, $this->stabilityFlags, $package->getNames(), $package->getStability())
             ) {
-                $this->loadPackage($request, $package, false);
+                $this->loadPackage($request, $repositories, $package, false);
             } else {
                 $this->unacceptableFixedOrLockedPackages[] = $package;
             }
@@ -354,7 +358,7 @@ class PoolBuilder
      * @param RepositoryInterface[] $repositories
      * @return void
      */
-    private function loadPackagesMarkedForLoading(Request $request, $repositories)
+    private function loadPackagesMarkedForLoading(Request $request, array $repositories)
     {
         foreach ($this->packagesToLoad as $name => $constraint) {
             $this->loadedPackages[$name] = $constraint;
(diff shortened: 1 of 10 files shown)


stof:
is there a way to test this ?

Seldaek:
@naderman the last commit adds a full installer test with pretty much the same inputs as the poolbuilder test, which already fails because the new transitive requirements are for 2.x and the locked ones are 1.x, it just wasn't shown as failing in poolbuilder because that doesn't run the solver. I hope this is what you meant.

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixes-path-repo-replacer-with-transitive-deps.test
```diff
@@ -0,0 +1,80 @@
...
+]
+
+--EXPECT--
+[
+   "symlinked/transitive-1.0.0.0",
+   "symlinked/transitive-2.0.2.0",
+   "symlinked/path-pkg-2.0.0.0",
+   "root/update-1.0.4.0",
+   "symlinked/transitive3-1.0.3.0",
+   "symlinked/transitive3-2.0.4.0",
```
Seldaek, line 75: @naderman I don't see why symlinked/transitive3 should be unlocked in this scenario? root/update does not depend on it as far as I can tell..

root/update depends on symlinked/replaced-pkg and symlinked/path-pkg. Those both require symlinked/transitive3-replaced, which is then replaced by symlinked/transitive3. 

As far as I can tell though, we only ever load replacers if they are explicitly required now, so the replacer here does not get loaded at all (not even in its locked form). That may be a PoolBuilder bug, but it seems really unrelated to this PR.

Same for symlinked/transitive4 which depends on it loading symlinked/path-pkg-replace if I follow correctly, but that never happens either.

naderman: Yes sorry I got this wrong and forgot to add an actual requirement for the package from somewhere else that would lead to it actually being required at all.

naderman: @Seldaek ok so fixed in my branch here: https://github.com/naderman/composer/tree/do_not_auto_unlock_transitive_path_deps

That said, unlocking has to propagate to replacers, and it currently does so for other packages.

e.g. if I require a/a and b/b and a/a requires c/c and b/b replaces b/b, then we expect "update --with-all-dependencies a/a" to update b/b because it is a dependency of a/a through replacement.

naderman: I actually thought tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers.test was testing replacement but it seems to test the opposite case. So that's another test case we're missing for the pool builder in general.

Seldaek:
Rebased this on top of #10280, fixed some more bugs relating to my implementation here, and fixed the test (`"symlinked/path-pkg-replace": "*"` was still missing in the root requires for that package to be loaded at all), and now it looks good I believe.

Seldaek:
The diff is a mess tho with the two PRs so let's focus on merging #10280 first :)

Seldaek:
OK the diff is back to a reviewable state. Most of the changes are actually to fix a bug in `removeLoadedPackage` when removing packages which were not locked. The rest is fairly minimal changes.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -155,12 +165,27 @@ public function buildPool(array $repositories, Request $request)
 
             foreach ($request->getLockedRepository()->getPackages() as $lockedPackage) {
                 if (!$this->isUpdateAllowed($lockedPackage)) {
-                    $request->lockPackage($lockedPackage);
-                    $this->skippedLoad[$lockedPackage->getName()][] = $lockedPackage;
+                    $lockedName = $lockedPackage->getName();
```
naderman, line 168: variable is unused now?

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -604,14 +621,22 @@ private function unlockPackage(Request $request, $name)
             }
         }
 
-        unset($this->skippedLoad[$name], $this->loadedPackages[$name], $this->maxExtendedReqs[$name]);
+        if (isset($this->pathRepoUnlocked[$name])) {
+            foreach ($this->packages as $index => $package) {
+                if ($package->getName() === $name) {
+                    $this->removeLoadedPackage($request, $repositories, $package, $index, false);
```
naderman, line 627: It's not clear to me why this shouldn't unlock alias packages if a remove is necessary?

tests/Composer/Test/Fixtures/installer/partial-update-with-symlinked-path-repos.test
```diff
@@ -0,0 +1,91 @@
...
+    "platform": [],
+    "platform-dev": []
+}
+--INSTALLED--
+[
+    { "name": "a/old", "version": "0.9.0" },
+    { "name": "b/unstable", "version": "1.1.0-alpha", "require": {"f/dependency": "1.*"} },
+    { "name": "c/uptodate", "version": "2.0.0" },
+    { "name": "f/dependency", "version": "1.0.0" }
+]
```
naderman, line 75: Is there a reason for this unrelated set of installed packages?

naderman (approved):

naderman:
Tests look correct to me and seem to work. In general fine with this now, just the small notes I left you could still address.

tests/Composer/Test/Fixtures/installer/partial-update-with-symlinked-path-repos.test
Seldaek: Oh no copy paste nonsense, will delete

src/Composer/DependencyResolver/PoolBuilder.php
Seldaek: Right, rebase issue, thanks

src/Composer/DependencyResolver/PoolBuilder.php
Seldaek: Actually it's not clear to me why unlockPackage was ever called in removeLoadedPackage - as it's only ever called from unlockPackage, it makes no sense. The aliases have the same name as the aliased package, and unlockPackage works only based on the name, so unlocking the same name again is useless. Removed this entirely in the last commit.

PR #10160, opened 2021-10-14
Validate composer.lock

villfa:
This new step will detect when the lock file is not updated.


Seldaek:
There's already a validate step in the same file above, but it is not done with --strict, maybe we could simply add that flag there?

stof:
@Seldaek the existing validation runs after a partial update, so it does not actually validate the committed lock file

Seldaek:
Oh yeah I missed that. Sounds good then :)

PR #10166, opened 2021-10-15
Reaching phpstan level 6 in Composer\IO

imme-emosol:


Seldaek:
Thanks! See review comments for adjustments if you plan on doing more of these :)

PR #10176, opened 2021-10-17
Fix validator, commit-ref must warn only for non-dev packages

### LINKED ISSUE #10175: Commit ref in dev dependencies
### Issue

For composer config like:
```
    "require-dev": {
        "behat/behat": "^3.8.2 || dev-master#6f38d11",
    }
```

I get the following error from validator:
```
- The package "behat/behat" is pointing to a commit-ref, this is bad practice and can cause unforeseen issues.
```

I agree with the purpose of the message as dependencies defined by commit-ref are respected only by the root package [1].

However dev sections (like `require-dev`) are also only for the root package.

Taking this two things together shows, that dependencies references in dev sections by commit-ref are completely safe and there are usecases like the one defined above.

[1] https://getcomposer.org/doc/04-schema.md#package-links

### Proposed solution

This is a feature request to:
1. change the message to
  ```
  - The package "behat/behat" is pointing to a commit-ref which is root package feature only and it will be ignored in dependent packages.
  ```
2. and do not show it if used in dev sections as it is safe there

mvorisek:
fixes #10175

### DIFF
--- a/src/Composer/Util/ConfigValidator.php
+++ b/src/Composer/Util/ConfigValidator.php
@@ -158,13 +158,11 @@ class ConfigValidator
         }
 
         // check for commit references
-        $require = isset($manifest['require']) ? $manifest['require'] : array();
-        $requireDev = isset($manifest['require-dev']) ? $manifest['require-dev'] : array();
-        $packages = array_merge($require, $requireDev);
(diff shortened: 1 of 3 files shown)


stof:
This is also something discouraged in root packages.

Issue #10159, opened 2021-10-14
Raising the PHPStan level to 6


Seldaek:
I have laid down the ground work for reaching level 6 by typing all class properties and interface methods throughout the codebase, which should already help ensure that types are correct when they get added/copied to getters and such.

I definitely could use help though in reaching level 6, because it's a ton of work left (1465 errors for `src/`, 400 errors for `tests/`).

If you are interested in helping, here is how to go about it:

# Setup a Composer environment first if you haven't

- `git clone https://github.com/composer/composer`
- `cd composer && wget https://getcomposer.org/composer.phar && php composer.phar install`

# Setup the env for PHPStan-usage

This requires PHP 7.4 (it won't run with php8 currently due to reasons, don't try to make it do that)

- `php7.4 composer.phar phpstan-setup` to get phpstan installed in the project - make sure you have a composer.phar or some global composer install to run this command with, you can't run phpstan-setup with bin/composer as it replaces parts of the vendor dir and it'll break the install

# Running PHPStan

- Pick a single sub-namespace and focus work on one area, otherwise it's too many errors and also we risk having multiple people doing the same work. Ideally you can comment here to claim a namespace if you're gonna work on it.
- Change level from 5 to 6 in `phpstan/config.neon`
- Run it with the sub-path you are interested in e.g. `php7.4 vendor/bin/phpstan analyse --configuration=phpstan/config.neon src/Composer/Package/`

# Keep in mind

- If you are unsure about some type, rather leave it undefined/erroring than use `mixed` or something equally vague, leave it for someone else to fill the gap later, because reviewing the whole code to find inappropriate mixed-usage is going to be near impossible.
- The main focus is src/ IMO. Typing the tests can't hurt, but it might be a mess in places as some of that code is very old and using php 5.3-compatible patterns.

Thanks in advance to any volunteer :)

# Status

Pick one that's not checked, and claim below in comments please. 

- [x] src/Composer https://github.com/composer/composer/pull/10213
- [x] ├── Autoload https://github.com/composer/composer/pull/10179
- [x] ├── Command https://github.com/composer/composer/pull/10214
- [x] ├── Config https://github.com/composer/composer/pull/10182
- [x] ├── Console https://github.com/composer/composer/pull/10183
- [x] ├── DependencyResolver https://github.com/composer/composer/pull/10178
- [x] ├── Downloader https://github.com/composer/composer/pull/10193
- [x] ├── EventDispatcher https://github.com/composer/composer/pull/10192
[...]

PR #10178, opened 2021-10-17
Reaching phpstan level 6 in Composer/DependencyResolver (#10159)

imme-emosol:


Seldaek:
That last PHPStan failure can be ignored with a `// @phpstan-ignore-next-line` in the test file.

There are still lots of class `@param` declarations which you added which add no value as the params have type hints in the php code too. If you could go through the diff and get rid of those that'd be good. Once really done and you're ready to stop force-pushing let me know and I'll happily go over all the changes one last time but then please no more force push otherwise it's impossible to keep track of things. I can squash all commits when merging, that's no concern.

PR #10115, opened 2021-09-20
Mask Git credentials in the error message

SerheyDolgushev:
Error message might contain git credentials. The simplest way to reproduce it:
1. Setup `github-oauth` in `~/.composer/auth.json`
2. Add some public package to your project:
    ```
    composer require contextualcode/public-package
    ```
3. Switch the previously added package Git repository into private, and make sure current `github-oauth` has no access to it.
4. Remove the caches:
    ```
    composer clearcache
    ```
5. Remove the previously checked out sources for `contextualcode/public-package`:
    ```
     rm -Rf vendor/contextualcode/public-package
    ```
6. Run the composer install
```
    composer clearcache
```
7. You will get error message similar to:
    ```
    [RuntimeException]                                                                                                                                                                                       
    Failed to execute git clone --mirror -- 'https://YOUR_TOKEN:x-oauth-basic@github.com/contextualcode/public-package.git' '/Users/sd/Library/Caches/composer/vcs/git-github.com-contextualcode-public-package.git/'                                                                                                                                                                                                        
    
    Cloning into bare repository '/Users/sd/Library/Caches/composer/vcs/git-github.com-contextualcode-public-package.git'...                                                                                     
    remote: Repository not found.                                                                                                                                                                            
    fatal: repository 'https://YOUR_TOKEN:x-oauth-basic@github.com/contextualcode-public-package.git/' not found 
    ```

There is no need to show the real value of `YOUR_TOKEN`. Instead, the masked version should be showed to improve the security.

PR #10186, opened 2021-10-19
Remove scalar parameter type declaration

herndlm:
`string` can only be used in parameters starting from PHP 7.0.

Was added in #10115


stof:
This tells me that this code is untested, otherwise the CI would have failed on PHP 5

PR #10190, opened 2021-10-19
Add types to `Util`

herndlm:
Contribution to #10159

I tried adding stricter `non-empty-string` and `positive-int` types where possible (obvious directly or privately used and indirectly infered).

Also I found out that the prepared `Perforce` port type seems to be wrong, according to usage and https://www.perforce.com/manuals/cmdref/Content/CmdRef/P4PORT.html. It looks like it is actually `string` instead of `int`.

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -330,8 +330,6 @@ class Git
     }
 
     /**
-     * @param ProcessExecutor $process
-     *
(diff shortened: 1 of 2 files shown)


herndlm:
It's ready to review, will not force-push anymore. I just missed one phpdoc format thingy that I wanted to change

PR #10193, opened 2021-10-19
Add types to `Downloader`

herndlm:
Contribution for #10159

There were some inconsistencies between return types of inherited methods of `VcsDownloader`, `GitDownloader` and `SvnDownloader` where one would return `React\Promise\PromiseInterface` and the other `void`. I unified them by adding the missing promises but I'm not sure if that's the right or best solution.


Seldaek:
OK I think leave it how you have it with simply PromiseInterface, that's unlikely to cause problems.

PR #10198, opened 2021-10-20
Add types to `Package/Dumper`

herndlm:
Contribution for #10159

I thought of defining the dump data array shape but this would end up in some mad thing with more than 25 keys, many of them being optional and nested. I could still try to do that though if wanted and it could be shared as a phpstan type. On the other hand, maybe it would make more sense to refactor that to some value object at some point..


Seldaek:
Yup IMO not worth defining all the array structure.. seems overkill and we have the composer-schema.json to validate this in a way.

PR #10238, opened 2021-10-27
phpstan: update tests/Composer/Test/Downloader/* to level 6 standard

megubyte:
Relates #10159.

All done except two errors I can't quite get my head around a clear solution for (without changing tests):
image

It makes sense that they're the current types which they are now, but it seems that some tests pass the unmocked version of `Composer\Config` and the `getDownloader()` instance assigns the mocked version to `$this->config` - and there seems to be no clean way to allow either the unmocked or mocked version to be easily passed because setting `$this->config` to be either version causes the calls to ->expects to fail under the static analysis.

I propose that the tests are either made to take a mocked instance, or they're made to take your regular instance of `\Composer\Config` - what would you recommend I do with this? I'm open to other suggestions on alternatives if required. I didn't want to blindly change tests for a task that involves updating the code with docblocks.


Seldaek:
Thanks!

PR #10253, opened 2021-11-03
Upgrade PHPStan to 1.0

Seldaek:
Still WIP for now, need to fix a few errors which are newly reported in 1.0 (if anyone wants to help feel free to send a PR against my PR branch, I won't have time to wrap this up in the next few days/weeks I think).

I am not sure what's up with all the ClassLoader failures though. I am thinking it may be a loading issue whereas the ClassLoader that is shipped with PHPStan is used and not the one from the Composer repo. Seemed to work fine in 0.12 though. @ondrejmirtes any clue about this? If not I can also just ignore these in baseline, it's not a huge deal as this class is pretty much written in stone.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,7 +81,7 @@
         "test": "simple-phpunit",
         "phpstan-setup": [
             "@composer config platform --unset",
-            "@php composer.phar update",
+            "@composer update",
             "@composer require --dev phpstan/phpstan:^1.0 phpstan/phpstan-phpunit:^1.0 phpunit/phpunit:^7.5.20 --with-all-dependencies",
             "git checkout composer.json composer.lock"
         ],
--- a/composer.json
+++ b/composer.json
@@ -82,7 +82,7 @@
         "phpstan-setup": [
             "@composer config platform --unset",
             "@php composer.phar update",
-            "@composer require --dev phpstan/phpstan:^0.12.93 phpstan/phpstan-phpunit:^0.12.17 phpunit/phpunit:^7.5.20 --with-all-dependencies",
+            "@composer require --dev phpstan/phpstan:^1.0 phpstan/phpstan-phpunit:^1.0 phpunit/phpunit:^7.5.20 --with-all-dependencies",
             "git checkout composer.json composer.lock"
         ],
         "phpstan": "@php vendor/bin/phpstan analyse --configuration=phpstan/config.neon"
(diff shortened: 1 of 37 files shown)


composer.json
```diff
@@ -82,7 +82,7 @@
         "phpstan-setup": [
             "@composer config platform --unset",
             "@php composer.phar update",
```
stof, line 84: @Seldaek shouldn't this use `@composer` rather than `@php composer.phar`, to be compatible with cases where composer is installed globally ?

Seldaek: Yes right that'd be better

ondrejmirtes:
The ClassLoader thing is definitely a weird bug, I'm gonna look into it.

ondrejmirtes:
It should work with PHPStan 1.0.2, please upgrade and test it :) https://github.com/phpstan/phpstan/releases/tag/1.0.2

Seldaek:
Yup that worked, thanks!

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -286,10 +286,10 @@ private function initDownload($resolve, $reject, $origin, $url, $options, $copyT
      */
     public function abortRequest($id)
     {
-        if (isset($this->jobs[$id], $this->jobs[$id]['handle'])) {
+        if (isset($this->jobs[$id], $this->jobs[$id]['curlHandle'])) {
```
Seldaek, line 289: That was actually a bug causing aborts to fail 👍🏻

Seldaek:
@ondrejmirtes sorry to bother again but would you please look at the last failing one? I can't make any sense of it, and can't repro it in a minimalistic test case sadly https://phpstan.org/r/ef8e8b43-5e90-41c8-8453-6ab3c7b011d9 

See https://github.com/composer/composer/runs/4199944625?check_suite_focus=true

herndlm:
@Seldaek I remember that one. Could it be that isset is not needed because it is used in the canceller, which is always called after startJob, where curl_id is always set? Which would narrow the type and then it's not optional any more.

Must be something like that.

Seldaek:
I don't think it's that smart (great if it is tho;)) but could be something about the closures using $job by reference which somehow alerts the type? Not sure.

Seldaek:
OK found a way to repro it, reported https://github.com/phpstan/phpstan/issues/6008 and will add to baseline for now.

PR #10256, opened 2021-11-04
Limit GitHub token scope when authorizing interactively

mpdude:
When the documentation at https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth is accurate, only limited token scopes are needed:

> For public repositories when rate limited, the public_repo scope is required, for private repositories the repo:status scope is needed. 

This can be indicated in the URL to create tokens with limited capabilities.

### Update

It turns out the documentation is not accurate, #10257 updates it.

This PR has been amended to better explain to the user what options they have and to provide different URLs.

### DIFF
--- a/src/Composer/Util/GitHub.php
+++ b/src/Composer/Util/GitHub.php
@@ -90,9 +90,15 @@ class GitHub
         }
         $note .= ' ' . date('Y-m-d Hi');
 
-        $url = 'https://'.$originUrl.'/settings/tokens/new?scopes=repo:status,public_repo&description=' . str_replace('%20', '+', rawurlencode($note));
-        $this->io->writeError(sprintf('Head to %s', $url));
-        $this->io->writeError(sprintf('to retrieve a token. It will be stored in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
+        $url = 'https://'.$originUrl.'/settings/tokens/new?scopes=&description=' . str_replace('%20', '+', rawurlencode($note));
+        $this->io->writeError(sprintf('When working with _public_ GitHub repositories only, head to %s to retrieve a token.', $url));
+        $this->io->writeError('This token will have read-only permission for public information only.');
+
+        $url = 'https://'.$originUrl.'/settings/tokens/new?scopes=repo&description=' . str_replace('%20', '+', rawurlencode($note));
+        $this->io->writeError(sprintf('When you need to access _private_ GitHub repositories as well, go to %s', $url));
+        $this->io->writeError('Note that such tokens have broad read/write permissions on your behalf, even if not needed by Composer.');
+        $this->io->writeError(sprintf('Tokens will be stored in plain text in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
+        $this->io->writeError('For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth.');
 
         $token = trim($this->io->askAndHideAnswer('Token (hidden): '));
 
--- a/src/Composer/Util/GitHub.php
+++ b/src/Composer/Util/GitHub.php
@@ -90,7 +90,7 @@ class GitHub
         }
         $note .= ' ' . date('Y-m-d Hi');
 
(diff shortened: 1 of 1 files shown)


stof:
For public repos, no scopes are actually needed. We don't need to have rights to modify things in the public repo, only to read info about them.
and for private repos, I don't think `repo:status` is the right one. composer does not care about commit statuses in private repos. It care about getting the code.

stof:
This means that the current URL is right. The thing to edit is the doc.

mpdude:
I seem to be unable to reproduce or force the situation to occur where the token is needed… 🧐 when should that happen?

mpdude:
More precisely, for which operations does composer have to call the GitHub API so that it might hit rate limits?

stof:
@mpdude when using a vcs repository (for packages coming from packagist, github is only reached to download the zip, and the archive endpoint has been excluded from the normal rate limit of github precisely to allow package managers to use it more easily. Packagist is the one using the API)

mpdude:
D'oh, I had a `vcs` repo in my tests, but it had `"no-api": true` set as well...

mpdude:
According to https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps, you can create a `(no scope)` token.

> `(no scope)` Grants read-only access to public information (including user profile info, repository info, and gists)

That seems to be sufficient to `composer update` a public repo.

stof:
@mpdude yes, and that's what I did in my case (as I don't have a need for composer to access private repos). But we can only have 1 URL, not 2, and it currently gives the scopes covering private repos too (for private repos, you need the token even when you don't reach the rate limit)

mpdude:
My guess is that `repo:status` and "commit status" means the CI (or similar checks) status for a given commit. I can confirm that this alone is not sufficient for using private `vcs` type repositories.

@stof Why can't we have two URLs?

I mean, we could say something like 

> Head to https://github.com/settings/tokens/new?scopes=repo&... to retrieve a token.
> In case you do not need to access private "vcs" type repositories, you can also generate a token with reduced permissions at https://github.com/settings/tokens/new?scopes=&...

In case you wonder why I am so annoying about such details 😂 – I think issues like the recent https://github.com/advisories/GHSA-pjwm-rvh2-c87w show that sooner or later we all will run into situations where we will execute scripts or code that might easily grab our package manager tokens. The less permissions these tokens have, the better...

stof:
hmm, indeed. We could do something like that

stof:
> My guess is that `repo:status` and "commit status" means the CI (or similar checks) status for a given commit.

yes, that's what it means

mpdude:
I created #10257 to update the documentation and will try to extend this one to better explain to the user what they are doing.

stof (approved):

src/Composer/Util/GitHub.php
```diff
@@ -90,9 +90,15 @@ public function authorizeOAuthInteractively($originUrl, $message = null)
...
+        $this->io->writeError(sprintf('When working with _public_ GitHub repositories only, head to %s to retrieve a token.', $url));
+        $this->io->writeError('This token will have read-only permission for public information only.');
+
         $url = 'https://'.$originUrl.'/settings/tokens/new?scopes=repo&description=' . str_replace('%20', '+', rawurlencode($note));
-        $this->io->writeError(sprintf('Head to %s', $url));
-        $this->io->writeError(sprintf('to retrieve a token. It will be stored in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
+        $this->io->writeError(sprintf('When you need to access _private_ GitHub repositories as well, go to %s', $url));
+        $this->io->writeError('Note that such tokens have broad read/write permissions on your behalf, even if not needed by Composer.');
+        $this->io->writeError(sprintf('Tokens will be stored in plain text in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
+        $this->io->writeError('For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth.');
```
Seldaek, line 101: Removing the trailing dot here to avoid bad links if the terminal link detection is not so smart

```suggestion
        $this->io->writeError('For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth');
```

Seldaek:
Thanks

PR #10257, opened 2021-11-04
Update what GitHub token scope is needed and what it implies

mpdude:
Be more explicit and better explain to users the implications of using GitHub tokens with broad permissions.

Follows up on https://github.com/composer/composer/pull/10256.


stof (approved):

Seldaek:
Thanks!

PR #10259, opened 2021-11-07
Add back abandoned key in repository search results

herndlm:
This also fixes the abandoned info via interactive search in the init command.

Was mistakenly removed in https://github.com/composer/composer/commit/4940009f8377d06b9fb48df0cbe67193e0989830#diff-fcfc43da4ed67efe626640158c4aa49d5a162efd01e8844f0fbbc655fcffbddeL431-R433 in the ComposerRepository and apparently never implemented in the ArrayRepository

Noticed via https://github.com/Seldaek/composer/pull/5 

Test case for the init command: add deps interactively and search for e.g. "faker", which should list "fzaninotto/faker" as being abandoned.

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -482,11 +482,6 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
                     continue;
                 }
 
-                if (isset($result['abandoned']) && '' === $result['abandoned']) {
-                    // The API returns '' in case a package is abandoned but does not have a replacement
-                    $result['abandoned'] = true;
-                }
-
                 $results[] = $result;
             }
 
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -482,8 +482,9 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
                     continue;
                 }
 
-                if (isset($result['abandoned'])) {
-                    $result['abandoned'] = $result['abandoned'] ?: true;
+                if (isset($result['abandoned']) && '' === $result['abandoned']) {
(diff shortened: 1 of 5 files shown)


src/Composer/Repository/ComposerRepository.php
```diff
@@ -478,9 +478,15 @@ public function search($query, $mode = 0, $type = null)
...
             foreach ($search['results'] as $result) {
                 // do not show virtual packages in results as they are not directly useful from a composer perspective
-                if (empty($result['virtual'])) {
-                    $results[] = array('name' => $result['name'], 'description' => $result['description']);
+                if (!empty($result['virtual'])) {
+                    continue;
+                }
+
+                if (isset($result['abandoned'])) {
+                    $result['abandoned'] = $result['abandoned'] ?: true;
```
stof, line 486: this looks wrong, as you will turn `false` into `true`

herndlm: yeah you're right.. the only thing it needs to do here is turn '' into true, so I could explicitly do that I guess. The API seems to be retourning either a string which indicates the replacement, or an empty string which means there is none, or the key is not set which means it's not abandonded

Seldaek: Fixed it in the API as it is really an issue there IMO https://github.com/composer/packagist/commit/7888a6a6e38b9af665a7ee7d039b7652a38f688b 

So could simply forward the abandoned key, and make allow string|bool in the RepositoryInterface's return type, as I guess strictly speaking returning false there would be reasonable too, it's just not done right now.

herndlm: Oh thx, that's better. I removed the "fix" completely

Seldaek:
Thanks

Issue #10188, opened 2021-10-19
Get rid of using getenv


shyim:
The composer source code currently uses `getenv` to get environment variables. When composer is embedded into a project and a `.env` is loaded with example of `symfony/dotenv` the variables will be by default set only to $_SERVER, $_ENV. Composer didn't pick up that variable then.

As `putenv` is not thread safe, it's not a good solution https://github.com/symfony/dotenv/blob/5.3/Dotenv.php#L70

So I would recommend creating a custom helper method to get the environment variable using $_SERVER, $_ENV instead

PR #10262, opened 2021-11-08
Add COMPOSER_NO_DEV environment variable

yakatz:
This works fine in practice, but the tests fail because it doesn't call the initialization function, so the setup isn't done. I don't see how to fix that.

As discussed in https://github.com/composer/composer/discussions/9974

### DIFF
--- a/src/Composer/Command/BaseCommand.php
+++ b/src/Composer/Command/BaseCommand.php
@@ -154,7 +154,7 @@ abstract class BaseCommand extends Command
         }
 
         if (true == $input->hasOption('no-dev')) {
-            if (!$input->getOption('no-dev') && true == getenv('COMPOSER_NO_DEV')) {
+            if (!$input->getOption('no-dev') && true == Platform::getEnv('COMPOSER_NO_DEV')) {
                 $input->setOption('no-dev', true);
             }
         }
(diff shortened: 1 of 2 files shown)


src/Composer/Command/BaseCommand.php
```diff
@@ -153,6 +175,14 @@ protected function initialize(InputInterface $input, OutputInterface $output)
             $input->setOption('no-progress', true);
         }
 
+        if (true == $input->hasOption('no-dev')) {
+            if (true == $input->getOption('no-dev') || true == getenv('COMPOSER_NO_DEV')) {
+                $this->setOptionNoDev(true);
+            } else {
+                $this->setOptionNoDev(false);
+            }
+        }
```
Seldaek, line 184: IMO it'd be a better/less intrusive change if it simply would set a default value for the option if the option is not provided e.g.

```suggestion
        if (true == $input->hasOption('no-dev') && !$input->getOption('no-dev') && Platform::getEnv('COMPOSER_NO_DEV')) {
            $input->setOption('no-dev', true);
        }
```

Then all other changes in every command can be undone.

yakatz: I kept the two `if`s separate because I thought it was easier to understand that way. Is that an issue?

Can you also clarify the condition? I think it would be `($input->getOption('no-dev') || Platform::getEnv('COMPOSER_NO_DEV'))`.

And, does this work if #10188 hasn't been merged yet?

Seldaek: https://github.com/composer/composer/pull/10218 will be merged first if needed, that's no problem.

Seldaek:
Thanks!

PR #10264, opened 2021-11-09
Expanding home diretory in vcs url for private repositories

cyrilverloop:
Hi,

I would like to propose the ability to expand the "~" (home) folder for private repositories.

For example in a composer.json :
```json
{
    "repositories": [
        {
            "type": "git",
            "url": "~/repositories/project"
        }
    ]
}
```

### DIFF
--- a/src/Composer/Repository/RepositoryFactory.php
+++ b/src/Composer/Repository/RepositoryFactory.php
@@ -17,6 +17,7 @@ use Composer\IO\IOInterface;
 use Composer\Config;
 use Composer\EventDispatcher\EventDispatcher;
 use Composer\Util\HttpDownloader;
+use Composer\Util\Platform;
 use Composer\Util\ProcessExecutor;
 use Composer\Json\JsonFile;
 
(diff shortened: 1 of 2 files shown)


Seldaek:
I'm thinking maybe this would make more sense here https://github.com/composer/composer/blob/main/src/Composer/Repository/RepositoryFactory.php#L166

So if `isset($repo['url'])` then you'd call expandPath on it.. That way it would work for all repo types at once.

Seldaek:
Alternatively you could do it in every repo's constructor, but IMO doing it only in the git util here isn't good enough.

cyrilverloop:
Thanks for tips. I do not know much about Composer's internals and i think you are right.
Is this what you had in mind ?

Seldaek:
Yup that looks better to me thanks

Seldaek:
I am actually unsure if this works.. I assume this will then write the expanded path including `/home/$user/` into the lock file, which makes the lock file not very portable to other users then? I assume the point of having ~ expansion was to make it more portable, but not sure.

cyrilverloop:
Oh, i see what you mean. Yes the URL in the lock file has the "\~" replaced.
To keep the lock file portable, it should not replace the "~" in it.
If keeping the "~" in the lock file is not an option, then this feature should be discarded.

stof:
even keeping the `~` in the lock file won't be portable: different users might not have the same thing in their home directory. For things to be portable between machines, you need repositories to be addressable in a way that ensure that they mean the same for all machines

cyrilverloop:
@stof yes, i was aware of it. In my case, private repositories, the path would have been the same for different machines/containers.

Thanks for your feedback, i am closing this PR.

PR #10274, opened 2021-11-10
Add name sorting to ArchivableFilesFinder.

func0der:
This should address the issues discussed in https://github.com/composer/composer/discussions/10271


src/Composer/Package/Archiver/ArchivableFilesFinder.php
```diff
@@ -84,7 +84,8 @@ public function __construct($sources, array $excludes, $ignoreFilters = false)
             ->in($sources)
             ->filter($filter)
             ->ignoreVCS(true)
-            ->ignoreDotFiles(false);
+            ->ignoreDotFiles(false)
+            ->sortByName(true);
```
stof, line 88: the natural sorting is not available in the Symfony version used by the composer phar (it was added in symfony/finder 4.2). So you should rather use the basic sorting by name instead, to keep the archiving consistent between `composer archive` and cases using composer as a library.

func0der:
I have tested this change locally and it leads to zips that are identical across different machines.
Given, that you manually `touch` all files before ziping them.

naderman (approved):

stof:
@naderman would be great that you approve the CI run (and you might want to re-evaluate the level of workflow approval)

naderman:
@stof indeed missed that. @Seldaek want to change the approval requirements?

Seldaek:
Thanks, makes sense.

PR #10280, opened 2021-11-12
Fix unlocking of replacers when a replaced package is unlocked

Seldaek:
Note that the only relevant test is the added one (which failed before by adding way too few packages to the pool).

The other test changes are simply output order changes due to https://github.com/composer/composer/commit/ad4c9671432bad61661ed90d53b627beac16c985

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -535,7 +535,19 @@ class PoolBuilder
             // as long as it was not unfixed yet
             && isset($this->skippedLoad[$this->skippedLoad[$name]])
         ) {
-            $this->unlockPackage($request, $this->skippedLoad[$name]);
+            $replacerName = $this->skippedLoad[$name];
+            $this->unlockPackage($request, $replacerName);
+
+            if ($this->isRootRequire($request, $replacerName)) {
+                $this->markPackageNameForLoading($request, $replacerName, new MatchAllConstraint);
+            } else {
+                foreach ($this->packages as $loadedPackage) {
+                    $requires = $loadedPackage->getRequires();
+                    if (isset($requires[$replacerName])) {
+                        $this->markPackageNameForLoading($request, $replacerName, $requires[$replacerName]->getConstraint());
+                    }
+                }
+            }
         }
 
         unset($this->skippedLoad[$name], $this->loadedPackages[$name], $this->maxExtendedReqs[$name]);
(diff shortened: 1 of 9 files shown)


stof (approved):

Seldaek:
This isn't ready to merge anymore IMO. I noticed it solves part of what https://github.com/composer/composer/pull/9538 was fixing, backported the tests in this branch and will try and solve everything for provide cases as well.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -158,9 +158,8 @@ public function buildPool(array $repositories, Request $request)
                     $request->lockPackage($lockedPackage);
                     $lockedName = $lockedPackage->getName();
                     // remember which packages we skipped loading remote content for in this partial update
-                    $this->skippedLoad[$lockedName] = $lockedName;
-                    foreach ($lockedPackage->getReplaces() as $link) {
-                        $this->skippedLoad[$link->getTarget()] = $lockedName;
+                    foreach ($lockedPackage->getNames() as $name) {
+                        $this->skippedLoad[$name][] = $lockedPackage;
```
naderman, line 162: The way this is changed, it's changing desired behavior. This basically means that unlocking now always propagates to all replacers and providers instead of just replacers. If we want it to work that way it needs a test verifying this behavior so at least it's clear we made this choice and don't change it later.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -529,13 +590,26 @@ private function warnAboutNonMatchingUpdateAllowList(Request $request)
      */
     private function unlockPackage(Request $request, $name)
     {
-        if (
+        foreach ($this->skippedLoad[$name] as $providedBy) {
+            $providedBy = $providedBy->getName();
             // if we unfixed a replaced package name, we also need to unfix the replacer itself
```
naderman, line 595: if i have a/a providing z/z and b/b providing z/z and i unlock a/a, do I want to unlock b/b? I don't think so. This logic here really only works for replace. Not for provide.

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/partial-update-unfixing-with-replacers-providers.test
```diff
@@ -0,0 +1,87 @@
...
+            "name": "indirect/replacer",
+            "version": "1.0.0",
+            "require": {
+                "replacer/package": "^1.0"
+            }
+        }
+    ]
+]
+
+--EXPECT--
```
naderman, line 78: add expect optimized to the tests here too?

Seldaek: Only need to define them if they're different.

naderman: Oh sorry didn't check how you add implemented that, assumed it wouldn't optimize if the expectation wasn't there.

naderman (approved):

PR #10283, opened 2021-11-14
Better hint for missing extensions

shalvah:
I often run into a case where a package I'm using in production requires extensions I don't have locally. For instance, Laravel Horizon requires Linux (pcntl, posix) and I'm on Windows. It's an advanced admin dashboard, though, so I only need it in production, so I force the installation via

```
composer install--ignore-platform-req=ext-pcntl --ignore-platform-req=ext-posix
```

Unfortunately, this means whenever I'm running most Composer commands, such as adding or upgrading an unrelated package, I have to remember to add those arguments (`composer install --ignore-platform-req=ext-pcntl --ignore-platform-req=ext-posix vendor/package`). I often forget and encounter the Composer  error message.

I thought it would be helpful if the error message telling me the missing extensions also tells me how I can easily go ahead, so I've added that. 

```
You can also run `php --ini` inside terminal to see which files are used by PHP in CLI mode.
Alternatively, you can run Composer with `--ignore-platform-req=ext-pcntl--ignore-platform-req=ext-posix`
 to temporarily ignore these required extensions.
```

Now, when I forget, I can just copy and paste the arguments from the error message. No more long typing!


herndlm:
While I understand your use case, I'm a bit sceptical in regard that this might confuse non-experienced users that simply copy/paste this and ignore extensions as a quick fix without being aware of the implications potentially.
I recently added wildcard support, would that improve your work flow? E.g. you can ignore `ext-*`

shalvah:
> it also sounds a bit like those extensions are optional

No, they aren't. They are required for the package to run, but let's just say the package is never run in development.

> I recently added wildcard support, would that improve your work flow? E.g. you can ignore ext-*

Yes, that would definitely help.



> this might confuse non-experienced users that simply copy/paste this and ignore extensions as a quick fix

I get this concern, but I don't know how likely it is. For one, the package installation would still fail on production (at least at first). Plus, I think the error message is clear enough that the extensions are required and you're only temporarily ignoring them. But I'll go with whatever you decide.

herndlm:
Btw the wildcard feature is not released yet, but will be part of 2.2
Also: I'm not a maintainer, I just wanted to add my thoughts, definitely not deciding anything ;)

Seldaek:
Looks reasonable to me, thanks. I think I'd rather use an explicit long list of extensions like this which is accurate than using ext-* and risk pulling in other dependencies with extension requirements which would otherwise not be needed.

gitressa:
This is a great improvement and quite useful. Could it also be included under `composer create`? Currently, if a PHP extension is missing, this is the feedback you get:

`Could not find package drupal-composer/drupal-project with version 7.x-dev in a version installable using your PHP version, PHP extensions and Composer version.`

... where the specific PHP extension would be really helpful.

Seldaek:
@gitressa this is a completely different code path and not as easy to do there IMO, but if you really think it's important at least open a new issue, this here will be forgotten.

gitressa:
Thanks for the fast reply and for clarifying. As a workaround (ignoring missing extensions during creation) the code added here also works:

```
$ composer create-project --ignore-platform-reqs -n drupal-composer/drupal-project:7.x-dev
Creating a "drupal-composer/drupal-project:7.x-dev" project at "./drupal-project"
[...]

$ cd drupal-project
$ composer install -n
[...]
  Problem 1
    - Root composer.json requires PHP extension ext-gd * but it is missing from your system. Install or enable PHP's gd extension.
```

Have a nice day.

PR #10288, opened 2021-11-17
GitDownloader: add option for single use git clone to avoid --dissociate flag when cloning from local copy

glaubinix:
Running `git clone` with the `--dissociate` flag can take several minutes for large git repositories. In certain scenarios, like creating archives, we don't need a fully working copy of the git repository that works independent of the local cache. The archiver removes the clone at the end anyway.

I moved the transport option into a `git` key instead of setting it in the root in case the transport options get passed to `StreamContextFactory::getContext` then this would throw a `Options should have the form ["wrappername"]["optionname"] = $value` error.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -92,8 +92,15 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
 
         if (!empty($this->cachedPackages[$package->getId()][$ref])) {
             $msg = "Cloning ".$this->getShortHash($ref).' from cache';
+
+            $cloneFlags = '--dissociate --reference %cachePath% ';
+            $transportOptions = $package->getTransportOptions();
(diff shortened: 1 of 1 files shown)


naderman (approved):
I think the --reference option hardly has any use in either case, but definitely pointless when cloning a local repo for single use.

stof:
This clearly misses some documentation

glaubinix:
@stof where would you expect such documentation? As far as I can tell, transport options are not documented anywhere.

stof:
That's because until now, transport options were all about `StreamContextFactory`, and so they are documented on php.net in the stream context documentation.

stof:
What happens if the package gets installed from dist rather than source ? Transport options are not separate, so we would probably end up passing this `git` option to `StreamContextFactory`

naderman:
I don't entirely disagree with documenting this somewhere. But putting it in the repo docs is just a way to tell people how to shoot themselves in the foot, as this will have unintended consequences if one doesn't really only use the repo once and then deletes it again. This is really just for programmatic use in other tools, not sure of a good current place to document this.

PR #10289, opened 2021-11-17
Docs: add transport options example and differentiate between cURL and stream options

glaubinix:
* Removed reference to Bitbucket supporting Mercurial
* Added transport options example and reworded the description to differentiate between cURL and stream options

For cURL, shall I list all available options to make it clearer what is possible, e.g. that using `['ssl']['allow_self_signed'] = true` results in streams being used instead of cURL?


Seldaek:
Thanks, I think it's fine as is. allow_self_signed is so seldom used, I don't think it's worth mentioning specifically here.

PR #10293, opened 2021-11-19
Support ignore-platform-reqs in `composer outdated`

### LINKED ISSUE #10291: Support --ignore-platform-reqs in `outdated` command
Allowing `composer outdated` to ignore platform reqs and return absolute latest versions of packages allows developers to calculate better the pressure of shifting to a higher version of the language, for example from 7.4 -> 8.

Currently `outdated` would not inform you of new versions that require platform changes like upgrading PHP to a new version. If you are not looking for the "data that is not there" you may get a false sense of being fully updated.

A couple other tickets also mention this:
- #9631 
- #9602
- #9688

rdohms:
This allows users to also find libraries that require major platform
changes to unlock updates.

Closes #10291.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -485,6 +485,13 @@ php composer.phar show monolog/monolog 1.0.2
 * **--direct (-D):** Restricts the list of packages to your direct dependencies.
 * **--strict:** Return a non-zero exit code when there are outdated packages.
 * **--format (-f):** Lets you pick between text (default) or json output format.
+* **--ignore-platform-reqs:** ignore all platform requirements (`php`, `hhvm`,
+  `lib-*` and `ext-*`) and force the installation even if the local machine does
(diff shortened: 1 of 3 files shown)


herndlm:
Hi, I think most of the other commands support also `ignore-platform-req` which can be specified multiple times. I'd expect that this command supports it too then. What do you think of adding it? See e.g. https://github.com/composer/composer/blob/main/src/Composer/Command/InstallCommand.php#L120

rdohms:
@herndlm I pondered the same and if it was worth it.
Struggling to find the time to do the extra work, but if its a blocker I'll try to book some time.

rdohms:
@herndlm had a boring meeting, managed to squeeze it in. Let me know how this looks.

herndlm:
LGTM and seems to be working fine, thank you for the adaption :)

Seldaek:
Thanks!

PR #10298, opened 2021-11-23
Separate the optimized and raw outputs in the exception

stof:
This makes it easier to compare them


stof:
Btw, maybe we could also remove common lines (or move them to a separate section) to highlight the actual difference

Seldaek:
It's now enabled by default so no need for this anymore but thanks

Issue #10299, opened 2021-11-23
[FEATURE REQUEST] Global configuration to look for composer.json in upper directory


CDRO:
It would be nice to have the possibility to globally define an option for composer to always use the composer.json found in a directory above.

Working mainly on TYPO3 projects, I often encounter myself to be in a folder and wanting to update all packages and I always get the message below:
`No composer.json in current directory, do you want to use the one at ...`

Most of the time this does not bother me, as I understand the reason why this is (and I'm glad for this feature), as I simply type `composer update` and hit enter twice, which does the trick nicely (and so do my colleagues).
But recently I started working on some quality of life improvements in our projects (such as project-wide CGL check for our developments) and thought of handling this on a project basis, adding a simple composer command such as `composer php-cs-fix <path to folder>`. Now, executing this works, per se, as I can simply execute it from anywhere, give the script the absolute path to the folder I want to check and hit enter twice, but working on quality of life also means that I'd only have to hit enter once.

Therefore I'd propose to add a configuration flag that allows you to always answer `yes` to the question about using the upper `composer.json`.

As far as I can see, this *should* not be too complicated but I understand that this might be a very niche use case and therefore might be either completely thrown out and dismissed (which would be fair, my comfort is not your priority and that's ok, OOS projects are hard enough as they are) or not on high priority but I'd be very happy if you could consider this.

I'm not sure if this is the right place for this kind of requests, but I feel more confortable bothering you guys here than on twitter :-)

I'm also willing to give a try on creating a PR for this feature, if I can get a composer-dev setup working locally so I can run the tests before pushing anything to the PR.

Best Regards
Tizian

PR #10307, opened 2021-11-24
[FEATURE] Configure automatic yes for composer.json in upper directory

CDRO:
Resolves: #10299 

I'm far off being done, since I haven't got any tests for this new feature (and don't even know if you'd want to have it), so it'll stay at WIP state for now.

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -33,6 +33,7 @@ class Config
     public static $defaultConfig = array(
         'process-timeout' => 300,
         'use-include-path' => false,
+        'use-parent-dir' => 'prompt',
         'preferred-install' => 'dist',
         'notify-on-install' => true,
         'github-protocols' => array('https', 'ssh', 'git'),
@@ -73,7 +74,6 @@ class Config
         'use-github-api' => true,
         'lock' => true,
         'platform-check' => 'php-only',
-        'use-parent-dir-composer' => 'prompt',
         // valid keys without defaults (auth config stuff):
         // bitbucket-oauth
         // github-oauth
@@ -409,13 +409,6 @@ class Config
                 }
 
                 return $protos;
-
-            case 'use-parent-dir-composer':
-                if($this->config[$key] === null) {
-                    return false;
-                }
-                // This might seem weird, but we want to retur false if the value is anythung _but_ true
-                return $this->config[$key] === true;
 
             default:
                 if (!isset($this->config[$key])) {
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -73,6 +73,7 @@ class Config
         'use-github-api' => true,
         'lock' => true,
         'platform-check' => 'php-only',
+        'use-parent-dir-composer' => 'prompt',
         // valid keys without defaults (auth config stuff):
         // bitbucket-oauth
         // github-oauth
@@ -408,6 +409,13 @@ class Config
                 }
 
                 return $protos;
+
+            case 'use-parent-dir-composer':
+                if($this->config[$key] === null) {
+                    return false;
+                }
+                // This might seem weird, but we want to retur false if the value is anythung _but_ true
+                return $this->config[$key] === true;
 
             default:
                 if (!isset($this->config[$key])) {
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -171,11 +171,8 @@ class Application extends BaseApplication
             }
         }
 
-        $composer = Factory::createGlobal($this->io);
-        $useParentDirIfNoJsonAvailable = $composer->getConfig()->get('use-parent-dir');
-
         // prompt user for dir change if no composer.json is present in current dir
-        if ($io->isInteractive() && $useParentDirIfNoJsonAvailable !== false && !$newWorkDir && !in_array($commandName, array('', 'list', 'init', 'about', 'help', 'diagnose', 'self-update', 'global', 'create-project', 'outdated'), true) && !file_exists(Factory::getComposerFile())) {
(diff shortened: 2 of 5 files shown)


herndlm:
Hi, it looks like I'm the author of the commit, which I'm not :) can you rebase and fix that please?

src/Composer/Console/Application.php
```diff
@@ -171,6 +171,9 @@ public function doRun(InputInterface $input, OutputInterface $output)
             }
         }
 
+        $composer = Factory::createGlobal($this->io);
+        $alwaysUseParentComposerIfNoneAvailable = $composer->getConfig()->get('use-parent-dir-composer') || false;
```
stof, line 175: `|| false` is weird here. Are you using the javascript way of defining default values ?

CDRO:
Yeah, I've seen that, but I don't know why? I've rebased my master branch and suddenly you were the author of my commit ¯\＿(ツ)＿/¯ I'll try and fix this.

src/Composer/Console/Application.php
CDRO: Thanks for your feedback.

This was actually suggested by Php Inspections (EA Extended), instead of using `?: false` (as `??` is not available in PHP 5.3 yet)

stof: if you put a value in the default config, it will return a value AFAICT.

CDRO: From what I can tell, looking at `Config.php`, which actually gets the value and `ConfigCommand.php` which sets the value, there is no way to set a default value for something, so I could drop this if default value setting if I ensure something is set in `SelfUpdateCommand.php`?
image

stof: `Config::$defaultConfig` is where the default config is set, so that `$this->config[$key]` is always set.

stof: and if you want extra safety, add this config key in the section `booleans without env var support` of the switch.

CDRO: Yes and yes, thanks very much!

src/Composer/Config.php
```diff
@@ -73,6 +73,7 @@ class Config
         'use-github-api' => true,
         'lock' => true,
         'platform-check' => 'php-only',
+        'use-parent-dir-composer' => false,
```
Seldaek, line 76:
```suggestion
        'use-parent-dir-composer' => 'prompt',
```

src/Composer/Config.php
```diff
@@ -314,6 +315,7 @@ public function get($key, $flags = 0)
             case 'secure-http':
             case 'use-github-api':
             case 'lock':
+            case 'use-parent-dir-composer':
```
Seldaek, line 318: This should move below into its own section handling true/false/prompt as value.

src/Composer/Console/Application.php
```diff
@@ -171,6 +171,9 @@ public function doRun(InputInterface $input, OutputInterface $output)
             }
         }
 
+        $composer = Factory::createGlobal($this->io);
+        $alwaysUseParentComposerIfNoneAvailable = $composer->getConfig()->get('use-parent-dir-composer');
+
         // prompt user for dir change if no composer.json is present in current dir
         if ($io->isInteractive() && !$newWorkDir && !in_array($commandName, array('', 'list', 'init', 'about', 'help', 'diagnose', 'self-update', 'global', 'create-project', 'outdated'), true) && !file_exists(Factory::getComposerFile())) {
```
Seldaek, line 178:
```suggestion
        if ($io->isInteractive() && $alwaysUseParentComposerIfNoneAvailable !== false && !$newWorkDir && !in_array($commandName, array('', 'list', 'init', 'about', 'help', 'diagnose', 'self-update', 'global', 'create-project', 'outdated'), true) && !file_exists(Factory::getComposerFile())) {
```

src/Composer/Console/Application.php
```diff
@@ -179,7 +182,7 @@ public function doRun(InputInterface $input, OutputInterface $output)
             // abort when we reach the home dir or top of the filesystem
             while (dirname($dir) !== $dir && $dir !== $home) {
                 if (file_exists($dir.'/'.Factory::getComposerFile())) {
-                    if ($io->askConfirmation('<info>No composer.json in current directory, do you want to use the one at '.$dir.'?</info> [<comment>Y,n</comment>]? ')) {
+                    if ($alwaysUseParentComposerIfNoneAvailable || $io->askConfirmation('<info>No composer.json in current directory, do you want to use the one at '.$dir.'?</info> [<comment>Y,n</comment>]? ')) {
```
Seldaek, line 185:
```suggestion
                    if ($alwaysUseParentComposerIfNoneAvailable === true || $io->askConfirmation('<info>No composer.json in current directory, do you want to use the one at '.$dir.'?</info> [<comment>Y,n</comment>]? ')) {
```

Seldaek (changes requested):
Looks fine to me but a few things need to be improved:

- Please allow false/true/'prompt' as values. That way people like me who never actually want this feature can turn it off and just get a new composer.json in a subdir whenever needed without having to confirm.
- Please add the new config value to the doc/06-config.md docs and res/composer-schema.json

src/Composer/Config.php
CDRO: I've looked at this and the only comparable option I've seen to this is 'store-auths' which does not have it's own section, can I assume that the default case will handle this properly?

src/Composer/Console/Application.php
CDRO: I'm not sure about this one, since if you decide to not use this option, you would still probably want to be asked if you want to use the composer.json in the upper directory, setting this option to false will simpy default to telling you that there is no composer.json.

src/Composer/Config.php
Seldaek: Yup it should be fine you're right.

src/Composer/Console/Application.php
Seldaek: Yes, I want it to tell me there is no composer.json, because it then allows me to use `composer require` in a subdir to create a new composer.json without having to say no, as that one won't say there is no composer.json it'll just create one..

Seldaek:
OK, had a play with it and spotted a few more issues but it's all resolved now, thanks!

PR #10308, opened 2021-11-24
Add support for setting platform packages to false to disable them

### LINKED ISSUE #9664: Provide ability to disable platform dependencies
Currently you can fake platform packages (PHP and extensions) so that you can emulate a production environment or define your target platform in the config. Example: `{"php": "7.0.3", "ext-something": "4.0.3"}`. However what you cannot do is defining that a specific PHP extension is _not_ available in the target. This can be problematic if your local environment where you run the `composer update` has a specific PHP extension available and thus composer might resolve different package versions which require these extensions - but then the created `composer.lock` cannot actually be deployed in the target environment, because the PHP extension is missing of course.

This has become apparent recently with the release of [`lcobucci/jwt: 4.1.0`](https://packagist.org/packages/lcobucci/jwt#4.1.0) for example, which requires `ext-sodium` as a hard dependency. `ext-sodium` is particularly troublesome, because PHP needs to be compiled with `--with-sodium` for it to be available. Whether this is done or not depends on the target environment's ecosystem. So if your local development environment has `ext-sodium` enabled, which will result in `lcobucci/jwt` getting installed and locked into the `composer.lock` - and your target environment does not actually have support for `ext-sodium`, you are currently left with the following options:

1. Disable `ext-sodium` locally. While this is easy enough under Windows (since it's just a DLL), it might not be so easy in Linux or MacOS, where you are relying on PHP being compiled by the maintainers of whatever package repository you are obtaining PHP from. In order to have full control, you would have to compile PHP yourself.
2. Enable `ext-sodium` in the target. This might simply be out of your control and again also depends on the environment and where that environment obtains their PHP binaries from. Hosters like Alfahosting and HostEurope have confirmed that their PHP binary will have `ext-sodium` enabled starting with PHP 8.0 for example - even though `ext-sodium` is bundled within the PHP core since PHP 7.2.

For this particular instance it would be useful if you could explicitly exclude PHP extensions for dependency resolving. e.g. via

```json
{
    "config": {
        "platform": {
            "ext-sodium": false
        }
    }
}
```

or something to that effect.

Seldaek:
Fixes #9664

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -704,7 +704,7 @@ EOT
                 return 0;
             }
 
-            $this->configSource->addConfigSetting($settingKey, $values[0]);
+            $this->configSource->addConfigSetting($settingKey, $values[0] === 'false' ?  false : $values[0]);
 
             return 0;
         }
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -191,6 +191,9 @@ you may ignore it instead by passing `--ignore-platform-req=ext-foo` to `update`
 extensions as if you ignore one now and a new package you add a month later also
 requires it, you may introduce issues in production unknowingly.
 
+If you have an extension installed locally but *not* on production, you may want
+to artificially hide it from Composer using `{"ext-foo": false}`.
+
 ## vendor-dir
 
 Defaults to `vendor`. You can install dependencies into a different directory if
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -203,7 +203,7 @@
                     "type": "object",
                     "description": "This is a hash of package name (keys) and version (values) that will be used to mock the platform packages on this machine.",
                     "additionalProperties": {
-                        "type": "string"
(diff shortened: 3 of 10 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -704,7 +704,7 @@ function ($vals) {
                 return 0;
             }
 
-            $this->configSource->addConfigSetting($settingKey, $values[0]);
+            $this->configSource->addConfigSetting($settingKey, $values[0] === 'false' ?  false : $values[0]);
```
stof, line 707: should we turn `true` into a `*` requirement, or should we reject `true` in the ValidateCommand ? the json schema won't reject it so it needs to be validated separately.

Seldaek: You can't use `*` in the platform config as it has to be a valid version. `true` currently works as it gets converted to "1" and parsed as if you set the version "1", which sucks indeed. I'll make sure it fails.

PR #10309, opened 2021-11-24
Archiving fixes/tweaks

### LINKED ISSUE #7358: Composer archive should not read ignore files (gitignore, hgignore, ...)
Composer archive currently reads ignore files and excludes these files/directories from generated zip files. This behaviour is not correct since ignored directories should only be ignored for commiting changes. They may very well exist in the repository anyway and should be included in the resulting zip file, too.

To skip files or directories when generating the zip file other mechanisms ought to be used. For git one should use git attributes to exclude directories from export. Composer should only take these settings into account. We could see what alternatives there are for other VCS systems or whether there should be a way to define such an exclusion list in composer.json.

### LINKED ISSUE #9153: Support for unsetting export-ignore attribute in .gitattributes
The following allows to express "exclude everything but `src` and `LICENSE` and is understood as such by `git archive` and Github:

```
/* export-ignore
/src -export-ignore
/src/** -export-ignore
/LICENSE -export-ignore
```

It does not seem to be handled when using `satis build` though. `statis build` seems to relate on this piece of code from Composer for that purpose:

https://github.com/composer/composer/blob/2a13bb2649151932407aae1a37a356179b1f5199/src/Composer/Package/Archiver/GitExcludeFilter.php#L69-L79

As you can see, lines ending with `-export-ignore` (note the first dash) will not be taken into account, which means this probably will result in an empty archive.

Seldaek:
- Gets rid of `.gitignore`/`.hgignore` handling (fixes #7358)
- Adds support for `-export-ignore` in .gitattributes (fixes #9153)


naderman (approved):

wandersonwhcr:
If I didn't have a script to check the contents of my zip archive files, I'll never realize that Composer is not using `.gitignore` files anymore. My `.gitattributes` file doesn't have `/vendor` directory and archiver created a huge zip archive.

TY.

wandersonwhcr:
If someone needs a script to check zip contents:

https://github.com/wandersonwhcr/romans/blob/main/.github/bin/package-check.sh

Issue #10311, opened 2021-11-25
Composer 2.2 might break compatibility with post update autoload generator


CDRO:
I've tested this with TYPO3 10 and TYPO3 9 and ran into and exception on finalisation of composer update.

I'm aware, that this is probably something that has to be solved on the vendor side rather than composer side, but since the setup for TYPO3 worked in composer 2.1, I'm wondering if snapshot has not introduced an involuntary BC.

The error (with `composer update -vvv`):
```
> pre-autoload-dump: Helhum\Typo3ConsolePlugin\Plugin->listen
Class Cabag\BaloisesessionImport\Command\VipImportCommand located in ./web/typo3conf/ext/baloisesession_import/Classes/Command/OpaccVipImport.php does not comply with psr-4 autoloading standard. Skipping.
Class Cabag\BaloisesessionEvent\Domain\Model\Interview located in ./web/typo3conf/ext/baloisesession_event/Classes/Domain/Model/Inteview.php does not comply with psr-4 autoloading standard. Skipping.
> post-autoload-dump: TYPO3\CMS\Composer\Installer\Plugin->listen
Executing TYPO3 installer scripts
Executing "Helhum\Typo3Console\Composer\InstallerScript\PopulateCommandConfiguration": 
Executing "TYPO3\CMS\Composer\Plugin\Core\InstallerScripts\EntryPoint": 
Executing "TYPO3\CMS\Composer\Plugin\Core\InstallerScripts\EntryPoint": 
Executing "TYPO3\CMS\Core\Composer\CliEntryPoint": 
Executing "TYPO3\CMS\Composer\Plugin\Core\InstallerScripts\EntryPoint": 
> post-autoload-dump: TYPO3\CMS\Composer\Installer\Plugin->listen
> post-autoload-dump: TYPO3\ClassAliasLoader\Plugin->onPostAutoloadDump
> post-autoload-dump: Helhum\Typo3ConsolePlugin\Plugin->listen
> post-autoload-dump: Helhum\Typo3ConsolePlugin\Plugin->listen
> post-autoload-dump: PackageVersions\Installer->dumpVersionsClass
29 packages you are using are looking for funding.
Use the `composer fund` command to find out more!

                                                             
  [RuntimeException]                                         
  Cannot set the alias loader, as it is already registered!  
                                                             

Exception trace:
 () at /Users/st/scripts/projects/project_ci-baloisesession_ch/src/vendor/typo3/class-alias-loader/src/ClassAliasMap.php:72
[...]

PR #10312, opened 2021-11-25
Allow recursively call composerRequire

helhum:
Fixes: #10311
Related: #10065


stof (approved):

Seldaek:
Ah yes that makes total sense, thanks.

PR #10314, opened 2021-11-25
Add allow-plugins config value

### LINKED ISSUE #5659: Global config option to disable plugins and scripts
There's [documentation available to disable plugins and scripts](https://getcomposer.org/doc/faqs/how-to-install-untrusted-packages-safely.md) by passing the appropriate command line options: `--no-plugins --no-scripts`. Unfortunately, there's no (documented) way to make that the default via the global configuration.

Please add a configuration option to disable scripts and / or plugins globally.

Seldaek:
Fixes #5659

Still a few TODOs / open questions

- [x] Currently this warns on every operation once you have a plugin which is not enabled, I am not sure if this gets annoying or not. Are there cases where you want to have plugins installed but not enable them? Or is this only a safety net?
image
- [x] Probably would be nicer if it instead prompted the user if the plugin should be added or not when something doesn't match the allow-plugins rules. Default to no of course.. Then we could modify the config in band and proceed activating the plugin if desired. Otherwise store it in a disallow-plugins list to make explicit those you don't wanna hear about again?
- [x] Docs
- [x] add the new `/allow-plugins` shortlink on getcomposer.org

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -138,42 +138,42 @@
         },
         "require": {
             "type": "object",
-            "description": "This is a hash of package name (keys) and version constraints (values) that are required to run this package.",
+            "description": "This is an object of package name (keys) and version constraints (values) that are required to run this package.",
             "additionalProperties": {
                 "type": "string"
             }
         },
         "require-dev": {
             "type": "object",
-            "description": "This is a hash of package name (keys) and version constraints (values) that this package requires for developing it (testing tools and such).",
+            "description": "This is an object of package name (keys) and version constraints (values) that this package requires for developing it (testing tools and such).",
             "additionalProperties": {
                 "type": "string"
             }
         },
         "replace": {
             "type": "object",
-            "description": "This is a hash of package name (keys) and version constraints (values) that can be replaced by this package.",
+            "description": "This is an object of package name (keys) and version constraints (values) that can be replaced by this package.",
             "additionalProperties": {
                 "type": "string"
             }
         },
         "conflict": {
             "type": "object",
-            "description": "This is a hash of package name (keys) and version constraints (values) that conflict with this package.",
+            "description": "This is an object of package name (keys) and version constraints (values) that conflict with this package.",
             "additionalProperties": {
                 "type": "string"
             }
         },
         "provide": {
             "type": "object",
-            "description": "This is a hash of package name (keys) and version constraints (values) that this package provides in addition to this package's name.",
+            "description": "This is an object of package name (keys) and version constraints (values) that this package provides in addition to this package's name.",
             "additionalProperties": {
                 "type": "string"
             }
         },
         "suggest": {
             "type": "object",
-            "description": "This is a hash of package name (keys) and descriptions (values) that this package suggests work well with it (this will be suggested to the user during installation).",
+            "description": "This is an object of package name (keys) and descriptions (values) that this package suggests work well with it (this will be suggested to the user during installation).",
             "additionalProperties": {
                 "type": "string"
             }
@@ -217,7 +217,7 @@
             "properties": {
                 "psr-0": {
                     "type": "object",
-                    "description": "This is a hash of namespaces (keys) and the directories they can be found into (values, can be arrays of paths) by the autoloader.",
+                    "description": "This is an object of namespaces (keys) and the directories they can be found into (values, can be arrays of paths) by the autoloader.",
                     "additionalProperties": {
                         "type": ["string", "array"],
                         "items": {
@@ -227,7 +227,7 @@
                 },
                 "psr-4": {
                     "type": "object",
-                    "description": "This is a hash of namespaces (keys) and the PSR-4 directories they can map to (values, can be arrays of paths) by the autoloader.",
+                    "description": "This is an object of namespaces (keys) and the PSR-4 directories they can map to (values, can be arrays of paths) by the autoloader.",
                     "additionalProperties": {
                         "type": ["string", "array"],
                         "items": {
@@ -283,7 +283,7 @@
             "properties": {
                 "platform": {
                     "type": "object",
-                    "description": "This is a hash of package name (keys) and version (values) that will be used to mock the platform packages on this machine.",
+                    "description": "This is an object of package name (keys) and version (values) that will be used to mock the platform packages on this machine.",
                     "additionalProperties": {
                         "type": ["string", "boolean"]
                     }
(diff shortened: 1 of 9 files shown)


src/Composer/Command/ConfigCommand.php
```diff
@@ -501,6 +501,18 @@ function ($vals) {
                     return $vals;
                 },
             ),
+            'allow-plugins' => array(
+                function ($vals) {
+                    if (!is_array($vals)) {
```
stof, line 506: what about booleans ?

Seldaek:
Latest version of the PR does prompt the user whether new unknown plugins should be enabled if interactive. When non-interactive and allow-plugins is not set at all, it allows all plugins for BC. When non-interactive and allow-plugins is set but does not include a given package, that package is not allowed.

Now I do wonder if we shouldn't build in some time-based feature switch in there that makes it forbid all plugins if `allow-plugins` is not set in 6months or 1year from now? That way 2.2.0 and above would keep working BC but would warn users, getting people to slowly add that config, and then at some point it'd remove the BC layer and default `allow-plugins` to `false`.

Any thoughts on this last point?

src/Composer/Config.php
```diff
@@ -177,6 +177,11 @@ public function merge($config, $source = self::SOURCE_UNKNOWN)
                 if (in_array($key, array('bitbucket-oauth', 'github-oauth', 'gitlab-oauth', 'gitlab-token', 'http-basic', 'bearer')) && isset($this->config[$key])) {
                     $this->config[$key] = array_merge($this->config[$key], $val);
                     $this->setSourceOfConfigValue($val, $key, $source);
+                } elseif (in_array($key, array('allow-plugins')) && isset($this->config[$key]) && is_array($this->config[$key])) {
+                    // merging $val first to get the local config on top of the global one, then appending the global config,
+                    // then merging local one again to make sure the values from local win over global ones for keys present in both
```
naderman, line 182: Not sure I understand this logic? Why not just do it once, sounds like that should have the same effect?

naderman:
I agree with adding a mechanism to get people to start adding the setting in due time.

src/Composer/Plugin/PluginManager.php
```diff
@@ -69,6 +80,9 @@ public function __construct(IOInterface $io, Composer $composer, Composer $globa
         $this->globalComposer = $globalComposer;
         $this->versionParser = new VersionParser();
         $this->disablePlugins = $disablePlugins;
+
+        $this->allowPluginRules = $this->parseAllowedPlugins($composer->getConfig()->get('allow-plugins'));
+        $this->allowGlobalPluginRules = $this->parseAllowedPlugins($globalComposer ? $globalComposer->getConfig()->get('allow-plugins') : false);
```
naderman, line 85: I wonder how global plugin allow should even work? Merging the setting with local and applying to all plugins executed in this process or should the global setting apply to global plugins and local setting to local plugins?

naderman (changes requested):

src/Composer/Plugin/PluginManager.php
```diff
@@ -596,4 +624,108 @@ public function getPluginCapabilities($capabilityClassName, array $ctorArgs = ar
...
+            }
+        }
+
+        if (!isset($warned[$package])) {
+            if ($this->io->isInteractive()) {
+                $composer = $isGlobalPlugin ? $this->globalComposer : $this->composer;
+
+                $this->io->writeError('<warning>'.$package.($isGlobalPlugin ? ' (installed globally)' : '').' contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins</warning>');
+                while (true) {
+                    switch ($answer = $this->io->ask('<warning>Do you trust "'.$package.'" to execute code and wish to enable the plugin now? [y,n,d,?]</warning> ', '?')) {
```
naderman, line 691: I think this should probably clarify it will add the name to the composer.json for future execution?

res/composer-schema.json
```diff
@@ -288,6 +288,13 @@
                         "type": ["string", "boolean"]
                     }
                 },
+                "allow-plugins": {
+                    "type": ["object", "boolean"],
+                    "description": "This is an array of package names which are allowed to be loaded as plugins, or true to allow all, false to allow none, defaults to true.",
```
Seldaek, line 293:
```suggestion
                    "description": "This is an object of {\"pattern\": true|false} with packages which are allowed to be loaded as plugins, or true to allow all, false to allow none. Defaults to {} which prompts when an unknown plugin is added.",
```

src/Composer/Plugin/PluginManager.php
Seldaek: Right now it does apply global setting to global plugins and local to local plugins. Note that this refers to the `composer  global` installed plugins. The global config.json would just be merged in both local and global composer configs.

I think it's fine as is. If you add a plugin via `composer global require ...` it will prompt you and add it in that composer.json and it works.

src/Composer/Plugin/PluginManager.php
Seldaek: Yeah this is clarified in the help section. I am not sure how to formulate it in a way that's still short enough to avoid a wall of text as prompt here :/

src/Composer/Config.php
Seldaek: The thing is the patterns are looked up in order of declaration, so if you have `composer/*: true` locally but `composer/bla: false` in global config you want to make sure the local `composer/*` takes precedence. But if both define the same name doing `array_merge(local, global)` would overwrite the value of local conf with global one, which is then fixed by doing `array_merge(local, global, local)`

src/Composer/Plugin/PluginManager.php
naderman: I agree that's fine but this may require a bit more documentation? Can also add that when the first confused person shows up ;-)

bmack:
Hey @Seldaek ,

> Now I do wonder if we shouldn't build in some time-based feature switch in there that makes it forbid all plugins if allow-plugins is not set in 6months or 1year from now? That way 2.2.0 and above would keep working BC but would warn users, getting people to slowly add that config, and then at some point it'd remove the BC layer and default allow-plugins to false.

I was just reviewing your PR and was wondering why you put the timing check in the code, instead of using versioning. Otherwise people could stick with old versions (of course they should not do that), instead of faking their system time to make plugins work without configuration anymore :)

Seldaek:
@bmack why would anyone want to do this though? I mean if you really don't care for security you can do `composer config --global allow-plugins true` to just tell it to run anything. **NOT RECOMMENDED** but it's there.

src/Composer/Plugin/PluginManager.php
naderman:
```suggestion
                    switch ($answer = $this->io->ask('<warning>Do you trust "'.$package.'" to execute code and wish to enable it now? (writes allow-plugins to composer.json) [y,n,d,?]</warning> ', '?')) {
```

naderman (approved):

bmack:
@Seldaek well, personally, I'd never do such things (but you know, people come up with stupid ideas to solve their issues). Also: I'd never put this in a global setting at all (my CI wouldn't have the plugins enabled, but I have it in my global config - would lead to inconsistencies), but I see this 100% on a per-project config level, and love it.

Doing this switch not via versioning but by date seems odd. I just wondered why we shouldn't stick to sem-ver here, instead of going with a fixed date "deadline". In any case: Thanks a bunch, very much looking forward to this!

naderman:
@bmack We want this in 2.2 LTS so that this will apply to the majority of all users within a few months. As 2.3 is going to drop support for lots of old PHP versions, dependencies etc., we would have to introduce the requirement for allow-plugins in a 2.2.x release. But that would mean people who don't want the requirement have to stay on an outdated version missing bugfixes on the 2.2 LTS line, which we want to discourage them from doing.

So instead it'll be timed and included in all 2.2 releases, and worst case as Jordi described it can be disabled.

GPHemsley-RELX:
Is there a command that can be run that will do nothing but prompt about updating the `allow-plugins` config?

Seldaek:
`composer show` for example would load the plugins and prompt, without other side effect.

PR #10318, opened 2021-11-28
Add support for ignoring the upper bound of platform requirements using "name+" notation

Seldaek:
e.g. `--ignore-platform-req=php+` ignores the upper bound of the php requirements.

- [x] TODO docs but would like to hear if anyone has objections to the idea/impl first

Fixes https://github.com/composer/composer/issues/9534

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -117,7 +117,10 @@ resolution.
   See also the [`platform`](06-config.md#platform) config option.
 * **--ignore-platform-req:** ignore a specific platform requirement(`php`,
   `hhvm`, `lib-*` and `ext-*`) and force the installation even if the local machine
-  does not fulfill it. Multiple requirements can be ignored via wildcard.
+  does not fulfill it. Multiple requirements can be ignored via wildcard. Appending
+  a `+` makes it only ignore the upper-bound of the requirements. For example, if a package
+  requires `php: ^7`, then the option `--ignore-platform-req=php+` would allow installing on PHP8,
+  but installation on PHP 5.6 would still fail.
 
 ## update / u
 
@@ -202,7 +205,10 @@ php composer.phar update vendor/package:2.0.1 vendor/package2:3.0.*
   See also the [`platform`](06-config.md#platform) config option.
 * **--ignore-platform-req:** ignore a specific platform requirement(`php`,
   `hhvm`, `lib-*` and `ext-*`) and force the installation even if the local machine
-  does not fulfill it. Multiple requirements can be ignored via wildcard.
+  does not fulfill it. Multiple requirements can be ignored via wildcard. Appending
+  a `+` makes it only ignore the upper-bound of the requirements. For example, if a package
+  requires `php: ^7`, then the option `--ignore-platform-req=php+` would allow installing on PHP8,
+  but installation on PHP 5.6 would still fail.
 * **--prefer-stable:** Prefer stable versions of dependencies.
 * **--prefer-lowest:** Prefer lowest versions of dependencies. Useful for testing minimal
   versions of requirements, generally used with `--prefer-stable`.
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -15,6 +15,7 @@ namespace Composer\DependencyResolver;
 use Composer\Package\CompletePackageInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\BasePackage;
+use Composer\Package\Link;
 use Composer\Package\PackageInterface;
 use Composer\Package\RootPackageInterface;
 use Composer\Pcre\Preg;
@@ -242,19 +243,19 @@ class Problem
                 $ext = substr($packageName, 4);
                 $msg = "- Root composer.json requires PHP extension ".$packageName.self::constraintToText($constraint).' but ';
 
-                if (extension_loaded($ext)) {
-                    $version = self::getPlatformPackageVersion($pool, $packageName, phpversion($ext) ?: '0');
-
-                    if (null === $version) {
-                        return array($msg, 'the '.$packageName.' package is disabled by your platform config. Enable it again with "composer config platform.'.$packageName.' --unset".');
+                $version = self::getPlatformPackageVersion($pool, $packageName, phpversion($ext) ?: '0');
+                if (null === $version) {
+                    if (extension_loaded($ext)) {
+                        return array(
+                            $msg,
+                            'the '.$packageName.' package is disabled by your platform config. Enable it again with "composer config platform.'.$packageName.' --unset".',
+                        );
                     }
 
-                    $error = 'it has the wrong version ('.$version.') installed';
-                } else {
-                    $error = 'it is missing from your system';
+                    return array($msg, 'it is missing from your system. Install or enable PHP\'s '.$ext.' extension.');
                 }
 
-                return array($msg, $error.'. Install or enable PHP\'s '.$ext.' extension.');
+                return array($msg, 'it has the wrong version installed ('.$version.').');
             }
 
             // handle linked libs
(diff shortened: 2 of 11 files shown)


derrabus:
Thank you for working on this feature!

herndlm:
Hehe, I had a similar idea. Really glad to see that the ignore platform reqs refactoring was helpful :)

Seldaek:
@herndlm yeah this definitely enabled a much easier implementation for this use case.

doc/03-cli.md
```diff
@@ -117,7 +117,10 @@ resolution.
   See also the [`platform`](06-config.md#platform) config option.
 * **--ignore-platform-req:** ignore a specific platform requirement(`php`,
   `hhvm`, `lib-*` and `ext-*`) and force the installation even if the local machine
-  does not fulfill it. Multiple requirements can be ignored via wildcard.
+  does not fulfill it. Multiple requirements can be ignored via wildcard. Appending
+  a `+` makes it only ignore the upper-bound of the requirements, so that if something
+  requires for example `php: ^7` and you run PHP 8 `--ignore-platform-req=php+` would
+  allow installation, but on PHP 5.6 it would still fail.
```
naderman, line 123:
```suggestion
  does not fulfill it. Multiple requirements can be ignored via wildcard. Appending
  a `+` makes it only ignore the upper-bound of the requirements. For example, if a package
  requires `php: ^7`, then the option `--ignore-platform-req=php+` would allow installing on PHP8,
  but installation on PHP 5.6 would still fail.
```

doc/03-cli.md
```diff
@@ -202,7 +205,10 @@ php composer.phar update vendor/package:2.0.1 vendor/package2:3.0.*
...
```
naderman, line 211:
```suggestion
  does not fulfill it. Multiple requirements can be ignored via wildcard. Appending
  a `+` makes it only ignore the upper-bound of the requirements. For example, if a package
  requires `php: ^7`, then the option `--ignore-platform-req=php+` would allow installing on PHP8,
  but installation on PHP 5.6 would still fail.
```

src/Composer/Filter/PlatformRequirementFilter/IgnoreListPlatformRequirementFilter.php
```diff
@@ -30,6 +50,33 @@ public function isIgnored($req)
...
+        }
+
+        if (1 === preg_match($this->ignoreRegex, $req)) {
+            return new MatchAllConstraint;
+        }
+
+        $intervals = Intervals::get($constraint);
+        $last = end($intervals['numeric']);
+        if ($last && (string) $last->getEnd() !== (string) Interval::untilPositiveInfinity()) {
+            $constraint = new MultiConstraint(array($constraint, new Constraint('>=', $last->getEnd()->getVersion())), false);
```
naderman, line 77: want to optimize/simplify this one here?

tests/Composer/Test/Fixtures/installer/update-ignore-platform-package-requirement-list-upper-bounds.test
```diff
@@ -0,0 +1,46 @@
...
+            ]
+        }
+    ],
+    "require": {
+        "a/a": "1.0.*",
+        "php": "^4.3",
+        "ext-foo-baz": "9.0.0"
+    },
+    "provide": {
+        "ext-foo-bar": "5.0.3"
```
naderman, line 19: can you add one more require/provide here that tests it actually throws the right error if an extension is present with a version number too low?

naderman (changes requested):

src/Composer/Filter/PlatformRequirementFilter/IgnoreListPlatformRequirementFilter.php
Seldaek: Not sure it matters?

tests/Composer/Test/Fixtures/installer/update-ignore-platform-package-requirement-list-upper-bounds.test
Seldaek: Done 👍

src/Composer/Filter/PlatformRequirementFilter/IgnoreListPlatformRequirementFilter.php
naderman: Just the usual question: what does it cost to optimize vs how often it gets checked later, also wonder if this is usually the same constraint in a lot of packages and caching the result instead of working with intervals matters performance wise? But probably all only answerable with benchmarks.

Seldaek: Yeah given this is very special use I don't think it's worth obsessing over the perf aspects too much.

tests/Composer/Test/Fixtures/installer/update-ignore-platform-package-requirement-list-upper-bounds.test
```diff
@@ -36,11 +38,13 @@ Your requirements could not be resolved to an installable set of packages.
 
   Problem 1
     - Root composer.json requires PHP extension ext-foo-baz [== 9.0.0.0 || >= 9.0.0.0] but it is missing from your system. Install or enable PHP's foo-baz extension.
+  Problem 2
+    - Root composer.json requires b/b 1.0.* -> satisfiable by b/b[1.0.1].
+    - b/b 1.0.1 requires ext-foo-bar 10.* -> it is missing from your system. Install or enable PHP's foo-bar extension.
```
naderman, line 43: This message seems wrong now? ext-foo-bar is present, the version is just too low?

Seldaek: Yeah that was kinda unrelated bug in Problem output, added a commit that fixes it.

src/Composer/DependencyResolver/Problem.php
```diff
@@ -241,16 +242,18 @@ public static function getMissingPackageReason(RepositorySet $repositorySet, Req
...
+                            'the '.$packageName.' package is disabled by your platform config. Enable it again with "composer config platform.'.$packageName.' --unset".',
+                        );
+                    } else {
+                        $error = 'it is missing from your system';
                     }
-
-                    $error = 'it has the wrong version ('.$version.') installed';
                 } else {
-                    $error = 'it is missing from your system';
+                    $error = 'it has the wrong version ('.$version.') installed';
```
naderman, line 256:
```suggestion
                    $error = 'it has the wrong version installed ('.$version.')';
```

Hard to read otherwise

tests/Composer/Test/Fixtures/installer/update-ignore-platform-package-requirement-list-upper-bounds.test
```diff
@@ -0,0 +1,50 @@
...
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
+    - Root composer.json requires PHP extension ext-foo-baz [== 9.0.0.0 || >= 9.0.0.0] but it is missing from your system. Install or enable PHP's foo-baz extension.
+  Problem 2
+    - Root composer.json requires b/b 1.0.* -> satisfiable by b/b[1.0.1].
+    - b/b 1.0.1 requires ext-foo-bar 10.* -> it has the wrong version (5.0.3 provided by __root__ 1.0.0+no-version-set) installed. Install or enable PHP's foo-bar extension.
```
naderman, line 43:
```suggestion
    - b/b 1.0.1 requires ext-foo-bar 10.* -> it has the wrong version (5.0.3 provided by __root__ 1.0.0+no-version-set) installed. Install or enable PHP's foo-bar extension.
```

That last part is still a bit confusing but I guess okay as long as the first part is there?

Seldaek: Fixed that as well 👍🏻

doc/03-cli.md
```diff
@@ -117,7 +117,10 @@ resolution.
   See also the [`platform`](06-config.md#platform) config option.
 * **--ignore-platform-req:** ignore a specific platform requirement(`php`,
   `hhvm`, `lib-*` and `ext-*`) and force the installation even if the local machine
-  does not fulfill it. Multiple requirements can be ignored via wildcard.
+  does not fulfill it. Multiple requirements can be ignored via wildcard. Appending
+  a `+` makes it only ignore the upper-bound of the requirements. For example, if a package
+  requires `php: ^7`, then the option `--ignore-platform-req=php+` would allow installing on PHP8,
+  but installation on PHP 5.6 would still fail.
```
derickr, line 123: So it's PHP 5.6, but then PHP8. There is a space missing there ;-)

Seldaek: Will fix ;)

dg:
It seems to unnecessarily print the error message `Cannot use texy/texy's latest version v3.2.0 as it requires php 8.1 - 8.3 which is not satisfied by your platform.`, although it does install the library.

Seldaek:
@dg Yeah that warning is handled somewhere else, please open a new issue with details how to repro.

PR #10319, opened 2021-11-29
Drivers: avoid early initialize of tags and branches

glaubinix:
Initializing `$this->branches` and `$this->tags` before the data is fetched can lead to wrong/inconsistent data

Example:
calling `$this->getBranches()` once while it throw a 5XX error on the first request (including retries) and then catching that exception here https://github.com/composer/composer/blob/012556d/src/Composer/Repository/VcsRepository.php#L214 results in the driver having an empty branches array stored. Calling it further down will then just return an empty array.

In addition to this PR, I am not sure the VcsRepository should just catch any kind of exception here and ignore it. I feel at least 5XX responses, request timeouts, network errors and such should be thrown again https://github.com/composer/composer/blob/012556d/src/Composer/Repository/VcsRepository.php#L214

### DIFF
--- a/src/Composer/Repository/Vcs/GitBitbucketDriver.php
+++ b/src/Composer/Repository/Vcs/GitBitbucketDriver.php
@@ -293,7 +293,7 @@ class GitBitbucketDriver extends VcsDriver
         }
 
         if (null === $this->tags) {
-            $this->tags = array();
+            $tags = array();
             $resource = sprintf(
                 '%s?%s',
                 $this->tagsUrl,
@@ -311,7 +311,7 @@ class GitBitbucketDriver extends VcsDriver
             while ($hasNext) {
                 $tagsData = $this->fetchWithOAuthCredentials($resource)->decodeJson();
                 foreach ($tagsData['values'] as $data) {
-                    $this->tags[$data['name']] = $data['target']['hash'];
+                    $tags[$data['name']] = $data['target']['hash'];
                 }
                 if (empty($tagsData['next'])) {
                     $hasNext = false;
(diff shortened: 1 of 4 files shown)


stof (approved):

src/Composer/Repository/Vcs/GitBitbucketDriver.php
```diff
@@ -311,14 +311,16 @@ public function getTags()
...
+                    $tags[$data['name']] = $data['target']['hash'];
                 }
                 if (empty($tagsData['next'])) {
                     $hasNext = false;
                 } else {
                     $resource = $tagsData['next'];
                 }
             }
+
+            $this->tags = $tags;
```
staabm, line 323: Maybe add a comment, that the assignment to the property happens intentionally only here.. otherwise people not familiar with the problem you are describing might reverse the change from this PR

glaubinix: This is the same already for all other drivers but was missing in the three drivers where I fixed this.

Seldaek:
Regarding the over-greedy catch block, I think we could apply the same logic in the catch as we do for caught tag/branch exceptions i.e. https://github.com/composer/composer/blob/012556d/src/Composer/Repository/VcsRepository.php#L301-L309 except also rethrow 404 and probably rethrowing 5xxs also makes sense yes.

glaubinix:
Unless I am missing something then 404 rethrow does change the current behaviour a bit. You won't be able to use a VCS repository with a default branch without a composer.json anymore. Not sure how often this occurs but it might,

Seldaek:
Fair enough.. I was more thinking about packages entirely gone, but I guess you're right. Anyway 5xx at least would be good I suppose, not sure if we should also apply this to tag/branches instead of silently suppressing them.

Seldaek:
Thanks

PR #10320, opened 2021-11-30
Add completion to commands options and arguments

### LINKED ISSUE #10292: [2.3] Add support for completion when on Symfony 5.4
According to this tweet:
https://twitter.com/seldaek/status/1461670175976169478

Composer 2.3+ will be on [Symfony 5.4 which got support for completion in Console](https://symfony.com/blog/new-in-symfony-5-4-console-autocompletion). See [Symfony 5.4 adding completion to their own commands](https://github.com/symfony/symfony/issues/43594).

The proposal here is to add this feature directly into Composer, allowing custom completion where appropriate, for example
1. `composer require sa<TAB>` could search Packagist for packages with name `sa%`
2. `composer update <TAB>` could complete package names already in `composer.json`

etc.

 The completion script could then be generated and shipped when Composer is distributed as a native package (for example, I'm using a [package prepared by Remi](https://rpms.remirepo.net/rpmphp/zoom.php?rpm=composer)), meaning many users will get DX improvement by default, others can run `composer completion bash > ~/.bash_completion.d/composer.sh` (or similar) once to get the same features.

GromNaN:
Fix #10292

# How to use ?

Install bash and bash_completion on your system.

```console
# Ubuntu
sudo apt install bash-completion
# Mac
brew install bash bash-completion@2
```

Init completion script

```console
bin/composer completion bash > completion.bash
source completion.bash
```

```console
bin/composer <tab,tab>
```

~Code requires `symfony/console >= 5.4` to work and `PHP >= 7.1` to not fail (`void` return type). The dependency updates are only for testing.~

Note: for testing I use the debug command described here https://github.com/symfony/symfony/issues/43594#issuecomment-947535113


composer.json
```diff
@@ -32,15 +31,17 @@
         "psr/log": "^1.0 || ^2.0",
         "seld/jsonlint": "^1.4",
         "seld/phar-utils": "^1.0",
-        "symfony/console": "^2.8.52 || ^3.4.35 || ^4.4 || ^5.0 || ^6.0",
+        "symfony/console": "^5.4",
```
GromNaN, line 34: Only for testing.

composer.json
```diff
@@ -22,7 +22,6 @@
         }
     ],
     "require": {
-        "php": "^5.3.2 || ^7.0 || ^8.0",
```
stof, line 25: Removing the PHP requirement is a no-go. From a dependency resolution PoV, this would support *more* versions than before

Seldaek: It's pretty clear this is a temp commit that'll have to be removed before merge..

src/Composer/Command/DependsCommand.php
```diff
@@ -22,6 +24,14 @@
  */
 class DependsCommand extends BaseDependencyCommand
 {
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        if ($input->mustSuggestArgumentValuesFor(self::ARGUMENT_PACKAGE)) {
+            $rootPackage = $this->getComposer(true, $input->getOption('no-plugins'))->getPackage();
```
Seldaek, line 30: This is sufficient/equivalent:

```suggestion
            $rootPackage = $this->getComposer()->getPackage();
```

src/Composer/Command/ExecCommand.php
```diff
@@ -108,4 +107,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+    /**
+     * @return string[]
+     */
+    private function getBinaries()
+    {
+        $composer = $this->getComposer();
+        $binDir = $composer->getConfig()->get('bin-dir');
+        $bins = glob($binDir . '/*');
+        $bins = array_merge($bins, array_map(function ($e) {
+            return "$e (local)";
```
Seldaek, line 120: I guess the local here may need to be stripped in the suggestions above? To get one you need to define a binary in the root package if you want to test.

composer.json
GromNaN: Indeed, should not have committed these changes. We have to wait for the requirements to be updated in order to finish this PR

src/Composer/Command/OutdatedCommand.php
```diff
@@ -23,6 +25,14 @@
  */
 class OutdatedCommand extends ShowCommand
 {
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        if ($input->mustSuggestArgumentValuesFor('package') || $input->mustSuggestOptionValuesFor('ignore')) {
+            $rootPackage = $this->getComposer(true, $input->getOption('no-plugins'))->getPackage();
```
Seldaek, line 31: Same here 
```suggestion
            $rootPackage = $this->getComposer()->getPackage();
```

src/Composer/Command/OutdatedCommand.php
```diff
@@ -23,6 +25,14 @@
...
+            $suggestions->suggestValues(array_merge(array_keys($rootPackage->getRequires()), array_keys($rootPackage->getDevRequires())));
```
Seldaek, line 32: This probably needs more work, I think it should probably list all packages which are installed.

src/Composer/Command/RemoveCommand.php
```diff
@@ -32,6 +34,14 @@
  */
 class RemoveCommand extends BaseCommand
 {
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        if ($input->mustSuggestArgumentValuesFor('packages')) {
+            $rootPackage = $this->getComposer(true, $input->getOption('no-plugins'))->getPackage();
```
Seldaek, line 40:
```suggestion
            $rootPackage = $this->getComposer()->getPackage();
```

src/Composer/Command/RequireCommand.php
```diff
@@ -57,6 +60,24 @@ class RequireCommand extends InitCommand
     /** @var bool */
     private $dependencyResolutionCompleted = false;
 
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        $value = $input->getCompletionValue();
+        if ($input->mustSuggestArgumentValuesFor('packages') && strlen($value) > 1) {
+            $composer = $this->getComposer(true, $input->getOption('no-plugins'));
```
Seldaek, line 67:
```suggestion
            $composer = $this->getComposer();
```

src/Composer/Command/RequireCommand.php
```diff
@@ -57,6 +60,24 @@ class RequireCommand extends InitCommand
...
+            $repos = new CompositeRepository($composer->getRepositoryManager()->getRepositories());
+
+            $packages = $repos->search('^'.$input->getCompletionValue(), RepositoryInterface::SEARCH_NAME);
+
+            foreach (array_slice($packages, 0, 150) as $package) {
+                $suggestions->suggestValue($package['name']);
+            }
+
+        } elseif ($input->mustSuggestOptionValuesFor('prefer-install')) {
+            $suggestions->suggestValues(array('dist', 'source', 'auto'));
```
Seldaek, line 77: Can we reuse this perhaps? Maybe have some common options autocompleted in BaseCommand?

src/Composer/Command/RunScriptCommand.php
```diff
@@ -44,6 +46,16 @@ class RunScriptCommand extends BaseCommand
         ScriptEvents::POST_AUTOLOAD_DUMP,
     );
 
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        if ($input->mustSuggestArgumentValuesFor('script')) {
+            $scripts = $this->getComposer()->getPackage()->getScripts();
+            foreach ($scripts as $name => $script) {
+                $suggestions->suggestValue($name);
```
GromNaN, line 54: array_keys

src/Composer/Command/ShowCommand.php
```diff
@@ -61,6 +63,14 @@ class ShowCommand extends BaseCommand
     /** @var ?RepositorySet */
     private $repositorySet;
 
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        if ($input->mustSuggestArgumentValuesFor('package') || $input->mustSuggestOptionValuesFor('ignore')) {
+            $rootPackage = $this->getComposer(true, $input->getOption('no-plugins'))->getPackage();
```
Seldaek, line 69:
```suggestion
            $rootPackage = $this->getComposer()->getPackage();
```

src/Composer/Command/ShowCommand.php
```diff
@@ -61,6 +63,14 @@ class ShowCommand extends BaseCommand
...
+            $suggestions->suggestValues(array_merge(array_keys($rootPackage->getRequires()), array_keys($rootPackage->getDevRequires())));
```
Seldaek, line 70: This here should also include more than just root requirements, I'd say at least include installed packages, and perhaps platform packages too. Not sure if we want to go as far as allowing autocomplete from the full available list of packages.

Essentially either of these if blocks:
all 
https://github.com/composer/composer/blob/b93530d72ea5b1f3f6fd44b947b4ff5b8c338704/src/Composer/Command/ShowCommand.php#L208-L212

Or installed + platform
https://github.com/composer/composer/blob/b93530d72ea5b1f3f6fd44b947b4ff5b8c338704/src/Composer/Command/ShowCommand.php#L221-L226

src/Composer/Command/OutdatedCommand.php
GromNaN, line 32: This code is duplicated in several commands and could be improved in the future. To factorise, we can add a protected method to the parent class, or create a `CompletionHelepr` class.

src/Composer/Command/UpdateCommand.php
```diff
@@ -37,6 +39,17 @@
  */
 class UpdateCommand extends BaseCommand
 {
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        if ($input->mustSuggestArgumentValuesFor('packages')) {
+            $rootPackage = $this->getComposer(true, $input->getOption('no-plugins'))->getPackage();
```
Seldaek, line 45:
```suggestion
            $rootPackage = $this->getComposer()->getPackage();
```

src/Composer/Command/UpdateCommand.php
```diff
@@ -37,6 +39,17 @@
...
+            $suggestions->suggestValues(array_merge(array_keys($rootPackage->getRequires()), array_keys($rootPackage->getDevRequires())));
```
Seldaek, line 46: This should definitely list transient dependencies, so anything installed on top of the root requirements (I'd merge root requirements too as this might include things which are updatable but not necessarily installed)

Seldaek (changes requested):
Thanks for getting started on this, looking forward to having it available!

src/Composer/Command/ShowCommand.php
dkarlovi: > Not sure if we want to go as far as allowing autocomplete from the full available list of packages.

Note that the **suggestions are further filtered down based on user input after Console suggest them**. If you have 200 packages, it will not just dump all 200 at you each time, they might be filtered down to 3 based on input.

IMO, we should start with suggesting everything that makes sense, otherwise it will be surprising why isn't it suggesting a thing you know is installed.

Seldaek: Yes but returning the full list of packages from packagist.org would be >300K items, maybe that'll cause issues somewhere.

dkarlovi: Oh, sorry, I misread, I meant you mean to skip transient, not packagist. That one should definitely be filtered out on source, agreed. :+1:

src/Composer/Command/RequireCommand.php
```diff
@@ -57,6 +59,12 @@ class RequireCommand extends InitCommand
     /** @var bool */
     private $dependencyResolutionCompleted = false;
 
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        // @todo add ext-* ?
```
dkarlovi, line 64: IMO no.

src/Composer/Command/BaseCommand.php
```diff
@@ -218,6 +227,83 @@ protected function getPreferredInstallOptions(Config $config, InputInterface $in
         return array($preferSource, $preferDist);
     }
 
+    /**
+     * @return bool
+     */
+    protected function completePreferInstall(CompletionInput $input, CompletionSuggestions $suggestions)
```
dkarlovi, line 233: `: bool`

src/Composer/Command/BaseCommand.php
```diff
@@ -218,6 +227,83 @@ protected function getPreferredInstallOptions(Config $config, InputInterface $in
...
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * @return bool
+     */
+    protected function completeInstalledPackage(CompletionInput $input, CompletionSuggestions $suggestions, $withPlatform = false)
```
dkarlovi, line 247: `: bool`

src/Composer/Command/BaseCommand.php
```diff
@@ -218,6 +227,83 @@ protected function getPreferredInstallOptions(Config $config, InputInterface $in
...
+            return $package->getName();
+        }, $installedRepo->getPackages()));
+
+        return true;
+    }
+
+    /**
+     * @return bool
+     */
+    protected function completeAvailablePackage(CompletionInput $input, CompletionSuggestions $suggestions)
```
dkarlovi, line 285: `: bool`

src/Composer/Command/BaseCommand.php
```diff
@@ -218,6 +227,83 @@ protected function getPreferredInstallOptions(Config $config, InputInterface $in
...
+            !$input->mustSuggestOptionValuesFor('require') &&
+            !$input->mustSuggestOptionValuesFor('require-dev')
+        ) {
+            return false;
+        }
+
+        $composer = $this->getComposer();
+        $repos = new CompositeRepository($composer->getRepositoryManager()->getRepositories());
+
+        $packages = $repos->search('^'.$input->getCompletionValue(), RepositoryInterface::SEARCH_NAME);
```
dkarlovi, line 298: Does this require some sort of escaping since it contains a slash?

dkarlovi (commented):
Amazing work! :rocket:

src/Composer/Command/RequireCommand.php
Seldaek: IMO yes? :) Why not?

dkarlovi: To be honest, it never even occurred to me you could `composer require ext-whatever`, I've never ever used that, I would usually add them manually, but now that I think about it some more, they're the same dependencies.

Do we have a canonical list of possible values?

Seldaek: Possible values no, but one can easily list what's available on the currently-running PHP from a PlatformRepository instance.

src/Composer/Command/BaseCommand.php
Seldaek: slash isn't a special character, but running preg_quote on `$input->getCompletionValue()` would be safer I suppose.

dkarlovi: @Seldaek how does that reflect here, the `search()` takes an already valid regex (including delimiters?) or does it construct one (in which it needs to `preg_quote7 with an already known delimiter).

Seldaek: IMO the current usage is fine, but it should preg_quote the user input. Delimiters are then added in search() implementation, but delimiters are always `{`/`}` which are anyway escaped by preg_quote so it's safe.

dkarlovi (commented):
Hands on experience.

src/Composer/Command/RequireCommand.php
```diff
@@ -57,6 +59,12 @@ class RequireCommand extends InitCommand
...
+        $this->completeAvailablePackage($input, $suggestions);
```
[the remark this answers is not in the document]
dkarlovi: Note: this might be a good candidate for a second PR after this one is merged, we don't need to complete *all* the completions here because it might be a lot of work. This already sets a really nice groundwork.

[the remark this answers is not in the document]
dkarlovi: Another such example (a good "later" candidate) is suggesting values for the `config` command, I guess reading the [schema](https://getcomposer.org/schema.json) might be a way to do it?

src/Composer/Command/BaseCommand.php
```diff
@@ -218,6 +227,83 @@ protected function getPreferredInstallOptions(Config $config, InputInterface $in
...
+
+        foreach (array_slice($packages, 0, 150) as $package) {
```
dkarlovi, line 300: The fact that we're limiting to 150 here has non-obvious consequences when used, see #10325. This part needs some special attention to polish the UX here.

src/Composer/Command/BaseCommand.php
GromNaN: I sticked to PHP 5.3 features, but since the `complete` method has `:void` return type, PHP 7.1+ is required. I'll update my changes to use short-array syntax, arrow functions and return type.

Seldaek: Yeah once this gets merged we'll be requiring PHP 7.2+ so go for it :)

dkarlovi: Yeah, these changes are totally 5.3 incompatible already so no reason not to go all in.

src/Composer/Command/RequireCommand.php
GromNaN: I'm adding the completion on package when the input starts with `ext-`.

Seldaek:
Please rebase on latest main branch :)

GromNaN:
I rebased, fixed some phpstan warnings.

src/Composer/Command/GlobalCommand.php
```diff
@@ -26,6 +29,16 @@
...
 class GlobalCommand extends BaseCommand
 {
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        $application = $this->getApplication();
+        if ($input->mustSuggestArgumentValuesFor('command-name')) {
+            $suggestions->suggestValues(array_filter(array_map(function (Command $command) {
+                return $command->isHidden() ? null : $command->getName();
+            }, $application->all())));
+        }
```
GromNaN, line 39: Completion of subcommand is hard. I have this for beginning; but the global state of the application must be modified.
```php
        if ($application->has($command = $input->getArgument('command-name'))) {
            $input = CompletionInput::fromString(preg_replace(['{\bg(?:l(?:o(?:b(?:a(?:l)?)?)?)?)?\b}', '{|$}'], ['', ''], $input->__toString(), 1), 3);
            $command = $this->getApplication()->find($command);
            $input->bind($command->getDefinition());
            $this->getApplication()->resetComposer();
            $command->complete($input, $suggestions);
        }
```

Seldaek: Not sure what you mean by "the global state of the application must be modified"?

GromNaN: All the thinks that are done in the `run` method.
https://github.com/composer/composer/blob/5e4a8a882dc70398c3c53e7b3aabd16790c9e198/src/Composer/Command/GlobalCommand.php#L90-L116

That said ... that's not so complex.

Seldaek: Ahh I see. Yes that should be easy to extract in a method to avoid duplication?

src/Composer/Command/CreateProjectCommand.php
```diff
@@ -55,6 +57,11 @@
  */
 class CreateProjectCommand extends BaseCommand
 {
+    public function complete(CompletionInput $input, CompletionSuggestions $suggestions): void
+    {
+        $this->completeAvailablePackage($input, $suggestions) || $this->completePreferInstall($input, $suggestions);
```
GromNaN, line 62: I'm not 100% satisfied by this code syntax. It would be more decoupled if completion callback were attached to each argument/option #44948.

src/Composer/Package/Version/VersionGuesser.php
```diff
@@ -72,6 +72,12 @@ public function guessVersion(array $packageConfig, $path)
             return null;
         }
 
+        // bypass version guessing in bash completions as it takes time to create
+        // new processes and the root version is usually not that important
+        if (isset($_SERVER['argv'][1]) && $_SERVER['argv'][1] === '_complete') {
+            return null;
+        }
```
Seldaek, line 79: Should do a similar bypass to get rid of https://github.com/composer/composer/blob/bbf2e94579cb91f8eacb9bd050f4b13aef67a853/src/Composer/Factory.php#L304-L317 

And I guess move this condition above to `Platform::isBashCompletionProcess()` or something so we avoid duplicating it.

GromNaN: `Platform::isInputCompletionProcess()` added.

src/Composer/Command/GlobalCommand.php
Seldaek: See https://github.com/composer/composer/pull/10320/commits/dd2e2d7e26d2698883960ee7dccda705ae89182e - I tried doing it but it doesn't seem to work, I think I missed something, perhaps you can have a look?

src/Composer/Command/CompletionTrait.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * Suggest package names available on all configured repositories or
+     * ext- packages from the ones available on the currently-running PHP
+     */
+    private function suggestAvailablePackageOrExtension(): \Closure
+    {
+        return function (CompletionInput $input) {
+            if (!str_starts_with($input->getCompletionValue(), 'ext-')) {
```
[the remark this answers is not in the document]
Seldaek: I'll try and take care of improving this method and suggestAvailablePackage together with taking a look at performance tomorrow, please leave it alone for now.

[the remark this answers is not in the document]
stof: > Do you have more details about `lib-*`? I've never seen such packages.

Look in the PlatformRepository

> This vendor prefix could be prohibed?

I don't see any reason for that. There is no conflicts between such packages and platform packages (as such packages will always have a `/` and platform packages don't have it).
Forbidding such prefixes (making existing packages unusable) just for lazyness of merging 2 lists in the completion suggestions looks wrong to me.

src/Composer/Factory.php
```diff
@@ -704,6 +704,10 @@ private static function getUserDir(): string
      */
     private static function validateJsonSchema(?IOInterface $io, $fileOrData, int $schema = JsonFile::LAX_SCHEMA, ?string $source = null): void
     {
+        if (!Platform::isInputCompletionProcess()) {
```
stof, line 707: should the `!` be removed here ?

Seldaek: Oops yes thanks.

src/Composer/Command/CompletionTrait.php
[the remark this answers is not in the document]
Seldaek: OK pushed a bunch of fine-tuning and improvements, IMO the package completion is now working well enough that it's usable.

src/Composer/Command/BaseCommand.php
Seldaek: I did improve this in https://github.com/composer/composer/pull/10320/commits/55dc80862ea81748df77153fdf14f861759c68b3 - for vendor names though when going through packagist.org results there are still so many that it's kind of a mess and you'll in most cases have to type enough of what you want to get it, but then it completes packages names within the vendor which is already quite useful IMO.

Seldaek:
OK PR is good to merge for me now except for the glitches in the global command. @GromNaN do you think you have time to look at that soon? Otherwise I'd probably merge as is and it can be done in a follow-up PR.

GromNaN:
> OK PR is good to merge for me now except for the glitches in the global command. @GromNaN do you think you have time to look at that soon? Otherwise I'd probably merge as is and it can be done in a follow-up PR.

Completion on `global` command needs some hooks in the `Composer\Console\Application`. I don't have anything conclusive ATM. Better work on it in a separate PR.

GromNaN:
Next step will be the setup the completion script with composer.phar.
This line may be removed.
https://github.com/composer/composer/blob/575fbfb53fcc2388916d554271c99c8281fea782/src/Composer/Compiler.php#L123

PR #10327, opened 2021-12-02
Install full binaries on Windows by default

helhum:
A bug was introduced in #10137 that leads
to the situation that by default .bat binaries
are not installed on Windows any more.

Check the correct variable to install .bat
files on Windows by default again.

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -103,7 +103,7 @@ class BinaryInstaller
                 $binCompat = 'full';
             }
 
-            if ($this->binCompat === "full") {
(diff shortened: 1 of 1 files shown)


helhum:
This small mistake was introduced here: [`e3ffad5` (#10137)](https://github.com/composer/composer/pull/10137/commits/e3ffad57f49875a6d4bb7255af49688f03b865a0)

stof (approved):

Seldaek:
Ah nice catch thanks

helhum:
> Ah nice catch thanks

Yeah, was wondering why my apveyour Windows builds were failing :)

Thanks for the quick merge!

PR #10335, opened 2021-12-07
Make it explicitly clear that `dev-` shouldn't be in the actual branch name

colans:
According to https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository, "you should prefix your custom branch name with 'dev-'", but doing so doesn't work.  It needs to be stated that this should only happen in `composer.json`, not the actual branch name.

I've been confused by this [along with many others](https://stackoverflow.com/questions/13498519/how-to-require-a-fork-with-composer#comment82782412_13500676).

Please update the docs with the request to save others some trouble.  Thanks!

I chose the `main` branch because there were too many unrelated changed, but feel free to cherry-pick for 2.1 .


Seldaek:
Thanks

PR #10352, opened 2021-12-10
Improve the doc for gitlab-token option

Jean85:
This highlights the security risk of using a `read_api` scope for the GitLab Personal Access Token

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -146,17 +146,26 @@ Further info can also be found [here](articles/authentication-for-private-packag
 
 ## gitlab-token
 
-A list of domain names and private tokens. Private token can be either simple
-string, or array with username and token. For example using `{"gitlab.com":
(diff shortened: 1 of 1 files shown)


Jean85:
I'm converting to draft because it looks like I've found a bug with Composer while using the `CI_JOB_TOKEN` in CI. I'll investigate further and report back.

Jean85:
I discovered a workaround: I can use `CI_JOB_TOKEN` as a valid HTTP basic auth so with:
```
composer config http-basic.gitlab.example.com gitlab-ci-token $(CI_JOB_TOKEN)
```

I can confirm that it works, at least against a repository that has an `Internal` visibility level set.

Should I reword this whole PR? Should this become a feature request?

Seldaek:
Closing here as I am not sure if we should teach people how to use env vars in general right here in the docs.. We already have COMPOSER_AUTH env var btw where you can store a json string with all the auth you need.

PR #10354, opened 2021-12-11
Always allow plugin "composer/package-versions-deprecated"

nicolas-grekas:
I propose to not annoy ppl with allowing this plugin since it's managed by the same team that makes composer.

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -657,6 +657,10 @@ class PluginManager
      */
     private function isPluginAllowed($package, $isGlobalPlugin)
     {
+        if ($package === 'composer/package-versions-deprecated') {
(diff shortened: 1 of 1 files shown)


Seldaek:
I am not so keen on special-casing anything here. A package name is just a name, it could come from anywhere. If you have a repository with higher prio than packagist.org which serves a different package under that name and it gets installed and activated this isn't great. Of course this could still happen even if we prompt, but at least if you don't expect it you may not hit yes in the prompt.

I just overall feel this would be unexpected and bad to auto-enable anything.

stof:
Actually, given that `package-versions-deprecated` is able to use the composer-runtime-api, we could do the opposite: always *disabling* it if not configured explicitly. We really need it to be a plugin only on composer 1.

PR #10363, opened 2021-12-13
$home -> $COMPOSER_HOME (doc: config)

ktomk:
for conclusion of the change please see discussion in PR comments below.

* [x] conclude the way of change (`$COMPOSER_HOME` variant)
* [ ] get review on `$home` for `data-dir` section, too.
* [ ] find the correct branch to PR against.

---
<details>
  <summary>original description (outdated)</summary>

the HOME environment parameter is upper-case in most of all systems.

> HOME
>
> The system will initialise this variable at the time of login to be a pathname of the user's home directory. See <pwd.h>.

(from: https://pubs.opengroup.org/onlinepubs/7908799/xbd/envvar.html)

change case to upper-case (US-ASCII; ref RFC 4790  9.2.1.  ASCII Casemap Collation Description)
</details>

---

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -253,8 +253,8 @@ into this directory.
 
 Defaults to `C:\Users\<user>\AppData\Roaming\Composer` on Windows,
 `$XDG_DATA_HOME/composer` on unix systems that follow the XDG Base Directory
-Specifications, and `$home` on other unix systems. Right now it is only
-used for storing past composer.phar files to be able to rollback to older
+Specifications, and `$COMPOSER_HOME` on other unix systems. Right now it is only
+used for storing past composer.phar files to be able to roll back to older
 versions. See also [COMPOSER_HOME](03-cli.md#composer-home).
 
 ## cache-dir
@@ -262,8 +262,8 @@ versions. See also [COMPOSER_HOME](03-cli.md#composer-home).
 Defaults to `C:\Users\<user>\AppData\Local\Composer` on Windows,
 `/Users/<user>/Library/Caches/composer` on macOS, `$XDG_CACHE_HOME/composer`
 on unix systems that follow the XDG Base Directory Specifications, and
-`$home/cache` on other unix systems. Stores all the caches used by Composer.
-See also [COMPOSER_HOME](03-cli.md#composer-home).
+`$COMPOSER_HOME/cache` on other unix systems. Stores all the caches used by
+Composer. See also [COMPOSER_HOME](03-cli.md#composer-home).
 
 ## cache-files-dir
 


ktomk:
Just adding in case it is noteworthy:

> $XDG_CACHE_HOME defines the base directory relative to which user-specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.

(from: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

Right now for me it is unclear (from level of the documentation) what the consequence within composer is when $XDG_CACHE_HOME is not set or empty:

* Those systems are considered unix systems _**not**_ _"follow[ing] the XDG Base Directory Specifications"_ and then using `$HOME/cache`
* Those systems are considered unix systems _"follow[ing] the XDG Base Directory Specifications"_ and then using `$HOME/.cache`

(mind the dot in front of "cache")

Seldaek:
The reason it is spelled like that is that the default value of the config option **is** `$home/cache` as per https://github.com/composer/composer/blob/main/src/Composer/Config.php#L45

The $home here is not an env var but the Composer `home` config value. 

> Right now for me it is unclear (from level of the documentation) what the consequence within composer is when $XDG_CACHE_HOME is not set or empty

As per this code:

https://github.com/composer/composer/blob/73a12bb21abe5fdae0793b115ece3df7dcb604cb/src/Composer/Factory.php#L138-L142

It'll use composer-home/cache if xdg is not detected, or $HOME/.cache/composer if it is detected but $XDG_CACHE_HOME is not set

ktomk:
Okay, then this first patch is bogus.

I've looked a bit into it based on your feedback. The issue is caused for me by reading `$home` as a shell parameter / environment variable (that it is not).

From what I can see now it is `$COMPOSER_HOME` in shell (which is also defined in the docs and already referenced with the entry).

There exist no such thing as `<composer-home>` as a config setting but with this notation it is also more clear that it's not an environment variable but a configuration setting (a special one, that is configured per `$COMPOSER_HOME` and the system otherwise).

XDG fallback detection looks legit (is_dir(/etc/xdg)), just writing, the detection _was not_ an issue on my end, I just was not aware of how it worked earlier in case no environment variables were set.

My current suggestion then would be either of these (in order of preference):

1. `$COMPOSER_HOME`

   > `$XDG_CACHE_HOME/composer` on unix systems that follow the XDG Base Directory Specifications, and `$COMPOSER_HOME/cache` on other unix systems. Stores all the caches used by Composer.

2. `<composer-home>`

   > `$XDG_CACHE_HOME/composer` on unix systems that follow the XDG Base Directory Specifications, and
`<composer-home>/cache` on other unix systems. Stores all the caches used by Composer.

3. `"{$home} ..."`

   > `$XDG_CACHE_HOME/composer` on unix systems that follow the XDG Base Directory Specifications, and
`"{$home}/cache"` on other unix systems. Stores all the caches used by Composer.

I'm undecided because I only looked once through the code. 1.) is it for me if `$COMPOSER_HOME` is correct. I can't say in context of older composer versions etc., checked it with my current one. 2.) I find an improvement as the ambiguity by the dollar-sign `$` is gone and things like `<user>` already use this bracket convention, so it is clear that this is a place-holder and not a variable name (`$home` requires to read the source first). 3.) is just doing the improvement to show that it can't be an shell/environment variable (win/osx/linux) so it must be PHP code. Does not improve readability that much in my eyes, so I put it last (makes it a similar puzzle then). 

2.) is perhaps less/least(?) ambiguous, so I'll update the PR later to it. Please let me know if 1.) is preferable or you think it's not going to pass an update at all. @Seldaek

Seldaek:
> I've looked a bit into it based on your feedback. The issue is caused for me by reading $home as a shell parameter / environment variable (that it is not).
>
> From what I can see now it is $COMPOSER_HOME in shell (which is also defined in the docs and already referenced with the entry).

What it is is a reference to another config setting, namely the `home` one which indeed will resolve to `$COMPOSER_HOME` if that is set.

The docs are wrong though as the only supported syntax by Composer\Config is `{$home}/cache`, so we should for sure fix that. But using `<composer-home>/cache` for docs purposes is probably better as it is clearer to anyone, even though not 100% accurate from a code perspective.

Overriding the home config from the system-based auto-detection is indeed possible through the env var, but also by setting `home` in your composer.json AFAIK, so showing it like you suggest in 1) is kinda reasonable to me but also strictly speaking not entirely correct. I don't know of a use case for setting `home` in composer.json though.. 

So in conclusion I'd say probably use 1).

ktomk:
allright, updating. there is a similar scenario in the _data-dir_ section of that file (`$home` only this time), I'd also go with `$COMPOSER_HOME` for it with the update.

ktomk:
@Seldaek: If I it right for the 2.2 milestone, it PR against master fine? Alternatively I can backport on top of 2.1 branch or any other, just let me know.

Seldaek:
Looks good like this, 2.2 is fine as the docs are built from the main branch anyway :) Thanks

PR #10373, opened 2021-12-20
Disable files autoloading for scripts to avoid untrusted code execution at runtime

Seldaek:
Plugin dependencies still get files autoloaded, but for scripts we keep the old restrictions which is documented at https://getcomposer.org/doc/articles/scripts.md#defining-scripts

> Callbacks can only autoload classes from psr-0, psr-4 and classmap definitions. If a defined callback relies on functions defined outside of a class, the callback itself is responsible for loading the file containing these functions.

Otherwise defining any PHP-callback script would automatically load all `files` from all packages installed, not giving you a chance to review what's present before they execute code, which goes against the new allow-plugin model.

cc @helhum @nicolas-grekas


naderman (approved):

helhum:
Makes sense, thanks!

PR #10377, opened 2021-12-20
Use web URL for Gitlab support metadata

GromNaN:
Currently, when using Gitlab VCS, the **support.issues** provides an API endpoint. That was added in a PR related to an other topic (#8510).

Before, with this sample project : https://packagist.org/packages/codingpaws/gitlab-feature
```json
    "support": {
        "issues": "https://gitlab.com/api/v4/projects/28361162/issues"
    }
```

This PR provide web UI URL, inspired by the `GitHubDriver`:
```json
    "support": {
        "source": "https://gitlab.com/codingpaws/gitlab-feature/-/tree/1.3",
        "issues": "https://gitlab.com/codingpaws/gitlab-feature/-/issues"
    },
```

### DIFF
--- a/src/Composer/Repository/Vcs/GitLabDriver.php
+++ b/src/Composer/Repository/Vcs/GitLabDriver.php
@@ -163,8 +163,12 @@ class GitLabDriver extends VcsDriver
 
             if ($composer) {
                 // specials for gitlab (this data is only available if authentication is provided)
-                if (!isset($composer['support']['issues']) && isset($this->project['_links']['issues'])) {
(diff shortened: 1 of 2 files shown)


GromNaN:
PHPStan claims that usage of `empty` and `?:` is not allowed. I took example from surrounding code.

Seldaek:
Thanks, I'll fix the phpstan mess, this should become easier in the future as we can rely on `??` instead as of Composer 2.3.

stof:
hmm, maybe some phpstan strict rules need to be disabled when targetting PHP 5.3

PR #10402, opened 2021-12-28
Make COMPOSER_BIN_DIR env or _composer_bin_dir global available to binaries

### LINKED ISSUE #10389: [2.2.0] Scripts from "bin" are executed not in "bin" directory.
Create a composer json that require any package, which contains a bash/sh script in `bin`, which executes another bin file from a `require`.

My `composer.json`:

```json
{
    "require": {
        "ninjify/qa": "^0.12.1"
    },
    "scripts": {
        "codesniffer": "codesniffer"
    },
    "config": {
        "allow-plugins": {
            "dealerdirect/phpcodesniffer-composer-installer": true
        }
    }
}
```

```
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys: FAIL
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (226689b90cbe0a4692a047df19539d4d3e959d79 => 2.2.0)
Composer version: 2.2-dev+226689b90cbe0a4692a047df19539d4d3e959d79
PHP version: 7.4.26
OpenSSL version: OpenSSL 1.1.1f  31 Mar 2020
cURL version: 7.68.0 libz 1.2.11 ssl OpenSSL/1.1.1f
```

When I run this command: 

```
composer codesniffer
```

I get the following output: 

```
Reading ./composer.json (/tmp/comtest/composer.json)
Loading config file ./composer.json (/tmp/comtest/composer.json)
Checked CA file /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a file.
Checked directory /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a directory.
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/tmp/comtest): git branch -a --no-color --no-abbrev -v
Executing command (/tmp/comtest): git describe --exact-match --tags
Executing command (CWD): git --version
Executing command (/tmp/comtest): git log --pretty="%H" -n1 HEAD --no-show-signature
Executing command (/tmp/comtest): hg branch
Executing command (/tmp/comtest): fossil branch list
Executing command (/tmp/comtest): fossil tag list
Executing command (/tmp/comtest): svn info --xml
Failed to initialize global composer: Composer could not find the config file: /home/home/.config/composer/composer.json

Reading /tmp/comtest/vendor/composer/installed.json
Loading plugin Dealerdirect\Composer\Plugin\Installers\PHPCodeSniffer\Plugin (from dealerdirect/phpcodesniffer-composer-installer)
Running 2.2-dev+226689b90cbe0a4692a047df19539d4d3e959d79 (2021-12-22 15:04:34) with PHP 7.4.26 on Linux / 5.4.0-91-generic
> codesniffer: codesniffer
Executing command (CWD): codesniffer
/tmp/comtest/vendor/ninjify/qa/bin/codesniffer: line 31: /tmp/comtest/vendor/ninjify/qa/bin/phpcs: No such file or directory
/tmp/comtest/vendor/ninjify/qa/bin/codesniffer: line 39: /tmp/comtest/vendor/ninjify/qa/bin/phpcs: No such file or directory
Script codesniffer handling the codesniffer event returned with error code 127
```

And I expected this to happen: Command is executed.

Seldaek:
Fixes #10389 - or at least adds a utility for packages with problems to fix it by doing something like this:

```
if [[ -z "$COMPOSER_BIN_DIR" ]]; then
  DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
else
  DIR="$COMPOSER_BIN_DIR"
fi
```

Instead of simply `DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"` to find the Composer bin dir.

The problem is before Composer 2.2, when a binary was symlinked into bin-dir, it got executed as `vendor/bin/foo` which ran the script really stored in `vendor/foo/bar/bin/foo`, but `$BASH_SOURCE[0]` was set to `vendor/bin/foo` still, so that allowed people to discover the bin dir. Composer 2.2 always creates proxies instead of symlinking, so now the proxy runs `vendor/foo/bar/bin/foo` directly and so the resolved dir is `vendor/foo/bar/bin` instead of bin-dir, which breaks some assumptions.

These scripts were kinda broken anyway as they relied on symlinks which was just one way binaries could be installed, and so they were working on some envs but not all. The current 2.2 model is simpler as it's always working the same, but it will require some adjustments.

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -231,12 +231,14 @@ class BinaryInstaller
             return "@ECHO OFF\r\n".
                 "setlocal DISABLEDELAYEDEXPANSION\r\n".
                 "SET BIN_TARGET=%~dp0/".trim(ProcessExecutor::escape(basename($link, '.bat')), '"\'')."\r\n".
+                "SET COMPOSER_BIN_DIR=%~dp0\r\n".
                 "{$caller} \"%BIN_TARGET%\" %*\r\n";
         }
 
         return "@ECHO OFF\r\n".
             "setlocal DISABLEDELAYEDEXPANSION\r\n".
             "SET BIN_TARGET=%~dp0/".trim(ProcessExecutor::escape($binPath), '"\'')."\r\n".
+            "SET COMPOSER_BIN_DIR=%~dp0\r\n".
             "{$caller} \"%BIN_TARGET%\" %*\r\n";
     }
 
@@ -260,14 +262,15 @@ class BinaryInstaller
             // carry over the existing shebang if present, otherwise add our own
             $proxyCode = empty($match[1]) ? '#!/usr/bin/env php' : trim($match[1]);
             $binPathExported = $this->filesystem->findShortestPathCode($link, $bin, false, true);
-            $autoloadPathCode = $streamProxyCode = $streamHint = '';
+            $streamProxyCode = $streamHint = '';
+            $globalsCode = '$GLOBALS[\'_composer_bin_dir\'] = __DIR__;'."\n";
             // Don't expose autoload path when vendor dir was not set in custom installers
             if ($this->vendorDir) {
-                $autoloadPathCode = '$GLOBALS[\'_composer_autoload_path\'] = ' . $this->filesystem->findShortestPathCode($link, $this->vendorDir . '/autoload.php', false, true).";\n";
+                $globalsCode .= '$GLOBALS[\'_composer_autoload_path\'] = ' . $this->filesystem->findShortestPathCode($link, $this->vendorDir . '/autoload.php', false, true).";\n";
             }
             // Add workaround for PHPUnit process isolation on PHPUnit 6+
             if ($this->filesystem->normalizePath($bin) === $this->filesystem->normalizePath($this->vendorDir.'/phpunit/phpunit/phpunit')) {
-                $autoloadPathCode .= '$GLOBALS[\'__PHPUNIT_ISOLATION_EXCLUDE_LIST\'] = $GLOBALS[\'__PHPUNIT_ISOLATION_BLACKLIST\'] = array(realpath('.$binPathExported.'));'."\n";
+                $globalsCode .= '$GLOBALS[\'__PHPUNIT_ISOLATION_EXCLUDE_LIST\'] = $GLOBALS[\'__PHPUNIT_ISOLATION_BLACKLIST\'] = array(realpath('.$binPathExported.'));'."\n";
             }
             if (trim($match[0]) !== '<?php') {
                 $streamHint = ' using a stream wrapper to prevent the shebang from being output on PHP<8'."\n *";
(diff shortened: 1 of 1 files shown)


src/Composer/Installer/BinaryInstaller.php
```diff
@@ -231,12 +231,14 @@ protected function generateWindowsProxyCode($bin, $link)
...
                 "setlocal DISABLEDELAYEDEXPANSION\r\n".
                 "SET BIN_TARGET=%~dp0/".trim(ProcessExecutor::escape(basename($link, '.bat')), '"\'')."\r\n".
+                "SET COMPOSER_BIN_DIR=%~dp0\r\n".
                 "{$caller} \"%BIN_TARGET%\" %*\r\n";
         }
 
         return "@ECHO OFF\r\n".
             "setlocal DISABLEDELAYEDEXPANSION\r\n".
             "SET BIN_TARGET=%~dp0/".trim(ProcessExecutor::escape($binPath), '"\'')."\r\n".
+            "SET COMPOSER_BIN_DIR=%~dp0\r\n".
```
Seldaek, line 241: @johnstevenson do you see any problem with this?

johnstevenson: No, except that `COMPOSER_BIN_DIR` there will include a trailing `\`, if that matters.

I'm curious why this `%~dp0` stuff is needed, when you know the location anyway - or will it move?

Seldaek: I have no idea why it's there anymore :) I saw the trailing slash, but i am assuming this will safely be ignored by most usages.

Seldaek: It's been in since the very beginning https://github.com/composer/composer/pull/140/files#diff-254824db37e271a109993b63f65d2072eaeeaf7b89996c8ccc0f937cb81ed2d0R173-R195 but perhaps it's not needed at all. Perhaps it was needed to create an absolute path to help with cygwin and such crappy envs.

johnstevenson: I think it was just copied from the way Composer-Setup.exe does it. Whatever, it is certainly not a crucial enough issue to be worrying about at the moment, so I'll shut up!

johnstevenson: Actually, in case of spaces in directory names it should probably be:
`"SET COMPOSER_BIN_DIR=\\"%~dp0\\"\r\n".

Seldaek: You sure? Because the one above BIN_TARGET does not have it, it only uses quotes when using the variable, not setting it.

johnstevenson: Sorry, you're correct. I had assumed that because BIN_TARGET was being escaped, this would also need it. But it would appear that even though `SET` is a command, it uses its full command-line and ignores any whitespace.

Note entirely sure why BIN_TARGET is escaped and then has any surrounding quotes removed, but I'm not going to worry about that.

naderman (approved):

PR #10405, opened 2021-12-29
fix: Do not optimise away packages due to a requirement by a locked package that will be uninstalled

### LINKED ISSUE #10394: Composer 2.2: selective update breaks on changed requirement
### Summary

There's a change in behaviour between Composer 2.1 and 2.2, where dependencies which were previously resolved without problem, now no longer get resolved.

Basic premise is a repo which has a committed `composer.lock` file and uses on-the-fly updates in the CI runs via GH Actions.
These updates used to work fine, but are now broken and failing builds.

I've set up a reproduction sample here: https://github.com/jrfnl/composer-2.2-bug-poc

And the changed behaviour and full `-vvv` output can be seen in the GH Actions output: https://github.com/jrfnl/composer-2.2-bug-poc/actions/runs/1616471281

### My `composer.json`:

```json
{
  "name": "jrfnl/composer-2.2-bug-poc",
  "description": "Proof of concept",
  "require": {
    "php": ">=5.6"
  },
  "require-dev": {
    "yoast/wp-test-utils": "^1.0.0"
  },
  "config": {
    "platform": {
      "php": "5.6.40"
    }
  }
}
```

### When I run this command **on PHP 8.0**:

```
composer require --dev phpunit/phpunit:"^7.5" --no-update --ignore-platform-req=php
composer update yoast/wp-test-utils --with-dependencies --ignore-platform-req=php
```

### I get the following output:

```
```
```
Updating dependencies

  Problem 1
    - phpunit/phpunit[7.5.0, ..., 7.5.20] require sebastian/exporter ^3.1 -> found sebastian/exporter[3.1.0, ..., 3.1.4] but these were not loaded, likely because it conflicts with another require.
    - Root composer.json requires phpunit/phpunit ^7.5 -> satisfiable by phpunit/phpunit[7.5.0, ..., 7.5.20].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

### And I expected this to happen:

The update to succeed without problems.

### Other notes:

Additional things I've tried:
* `--with-dependencies` vs `--with-all-dependencies` does not make a difference.
* Specifying the versions in the `update` command didn't help.
* Using `--with ...` in the update command doesn't help.
* Using `--no-install` instead of `--no-update` in the first command doesn't help.

### Work-around which works

... but may not play nice with caching in GH Actions as much:
```
composer remove --dev yoast/wp-test-utils
composer require --dev yoast/wp-test-utils:"^1.0.0" phpunit/phpunit:"^7.5" --update-with-dependencies --ignore-platform-reqs
```

driskell:
Fixes #10394, introduced by #9620

As explained in that issue when I did the optimisation to filter out packages from the pool that would ultimately never be installed due to a locked packages requirements, I hadn't considered that when packages are removed... they may still be in the set of locked packages.

This PR updates the optimisation to ignore any locked package which isn't required by any package in the pool - as that will, as I understand, mean the package is going to be removed and it's requirements will be moot. I have considered the case where a package IS required somewhere in the pool, or even an older version for example, and in those cases the package would have been unlocked when using `-W` anyway. So I believe this is the correct fix.

I tested it against the scenario in #10394 and when running the second command in the following set:

```
composer require --dev phpunit/phpunit:"^7.5" --no-update --ignore-platform-req=php
composer update yoast/wp-test-utils --with-dependencies --ignore-platform-req=php
```

You then get an error about PHPUnit is locked.

When then running the second command with `-W` then before this fix (2.2.1), you get the following:

```
Updating dependencies

  Problem 1
    - phpunit/phpunit[7.5.0, ..., 7.5.20] require sebastian/exporter ^3.1 -> found sebastian/exporter[3.1.0, ..., 3.1.4] but these were not loaded, likely because it conflicts with another require.
    - Root composer.json requires phpunit/phpunit ^7.5 -> satisfiable by phpunit/phpunit[7.5.0, ..., 7.5.20].

```

With this PR merged it is now successful.

I will run the same timing/memory tests I did with the original PR to make sure the update is not making the optimisation useless...

@Seldaek thanks for bringing it to my attention - this will definitely need scrutiny to make sure the approach is right!

### DIFF
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -420,14 +420,15 @@ class PoolOptimizer
         foreach ($request->getLockedPackages() as $package) {
             // If this locked package is no longer required by root or anything in the pool, it may get uninstalled so do not apply its requirements
             // In a case where a requirement WERE to appear in the pool by a package that would not be used, it would've been unlocked and so not filtered still
-            $ignoreUnusedPackage = false;
+            $isUnusedPackage = true;
             foreach ($package->getNames(false) as $packageName) {
-                if (!isset($this->requireConstraintsPerPackage[$packageName])) {
-                    $ignoreUnusedPackage = true;
+                if (isset($this->requireConstraintsPerPackage[$packageName])) {
+                    $isUnusedPackage = false;
+                    break;
                 }
             }
 
-            if ($ignoreUnusedPackage) {
+            if ($isUnusedPackage) {
                 continue;
             }
 
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -418,6 +418,19 @@ class PoolOptimizer
         }
 
         foreach ($request->getLockedPackages() as $package) {
+            // If this locked package is no longer required by root or anything in the pool, it may get uninstalled so do not apply its requirements
+            // In a case where a requirement WERE to appear in the pool by a package that would not be used, it would've been unlocked and so not filtered still
+            $ignoreUnusedPackage = false;
+            foreach ($package->getNames(false) as $packageName) {
+                if (!isset($this->requireConstraintsPerPackage[$packageName])) {
+                    $ignoreUnusedPackage = true;
+                }
+            }
+
+            if ($ignoreUnusedPackage) {
+                continue;
+            }
+
             foreach ($package->getRequires() as $link) {
                 $require = $link->getTarget();
                 if (!isset($packageIndex[$require])) {
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/filter-impossible-packages-locked-replacer.test
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/filter-impossible-packages-locked-replacer.test
@@ -1,0 +1,60 @@
+--TEST--
+Do not load packages into the pool that cannot meet the fixed/locked requirements, when a loose requirement is encountered during update
+(The locked replacer/pkg should restrict dependencies even though it is only referenced by its replaced name)
+
+--REQUEST--
+{
+    "require": {
+        "first/pkg": "*",
+        "second/pkg": "*",
+        "replacer/dep": "*"
+    },
+    "locked": [
+        {"name": "first/pkg", "version": "1.0.0", "require": {"replaced/pkg": "1.0.0"}},
+        {"name": "second/pkg", "version": "1.0.0"},
+        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "1.0.0"}, "require": {"replacer/dep": "1.*"}},
+        {"name": "replacer/dep", "version": "1.0.0"}
+    ],
+    "allowList": [
+        "second/pkg",
+        "replacer/dep"
+    ],
+    "allowTransitiveDeps": true
+}
+
+--FIXED--
+[
+]
+
+--PACKAGE-REPOS--
+[
+    [
+        {"name": "first/pkg", "version": "1.0.0", "require": {"replaced/pkg": "1.0.0"}},
+        {"name": "first/pkg", "version": "1.1.0", "require": {"replaced/pkg": "2.0.0"}},
+        {"name": "second/pkg", "version": "1.0.0"},
+        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "1.0.0"}, "require": {"replacer/dep": "1.*"}},
+        {"name": "replacer/pkg", "version": "1.1.0", "replace": {"replaced/pkg": "2.0.0"}, "require": {"replacer/dep": "1.*"}},
+        {"name": "replacer/pkg", "version": "1.2.0", "replace": {"replaced/pkg": "3.0.0"}, "require": {"replacer/dep": "1.*"}},
+        {"name": "replacer/dep", "version": "1.0.0"},
+        {"name": "replacer/dep", "version": "1.0.1"},
+        {"name": "replacer/dep", "version": "2.0.0"}
+    ]
+]
+
+--EXPECT--
+[
+    "first/pkg-1.0.0.0 (locked)",
+    "replacer/pkg-1.0.0.0 (locked)",
+    "second/pkg-1.0.0.0",
+    "replacer/dep-1.0.0.0",
+    "replacer/dep-1.0.1.0",
+    "replacer/dep-2.0.0.0"
+]
+
+--EXPECT-OPTIMIZED--
+[
+    "first/pkg-1.0.0.0 (locked)",
+    "replacer/pkg-1.0.0.0 (locked)",
+    "second/pkg-1.0.0.0",
+    "replacer/dep-1.0.1.0"
+]
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/filter-impossible-packages-only-required-provides.test
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/filter-impossible-packages-only-required-provides.test
@@ -1,0 +1,58 @@
+--TEST--
+When filtering packages from the pool that cannot meet the fixed/locked requirements, ensure that the requirements for a package that is not required anywhere is not used to filter, as it will be ultimately be removed.
+(Variant where the requirement is on a provided package, the locked third/pkg must not restrict the provider)
+(NOTE: We are not optimising this scenario currently)
+
+--REQUEST--
+{
+    "require": {
+        "first/pkg": "*",
+        "second/pkg": "1.1.0",
+        "provider/pkg": "*"
+    },
+    "locked": [
+        {"name": "first/pkg", "version": "1.0.0", "require": {"second/pkg": "^1.0"}},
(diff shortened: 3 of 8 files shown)


src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -397,6 +397,8 @@ private function optimizeImpossiblePackagesAway(Request $request, Pool $pool)
         }
 
         $packageIndex = array();
+        $requireIndex = array();
+        $rootRequires = $request->getRequires();
```
Seldaek, line 401: You could simply put this into the requireIndex here I think as only keys are relevant. Possibly there's already an array with the info needed here built by the other optimization pass, I'm not sure and am on my phone..

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -415,9 +417,18 @@ private function optimizeImpossiblePackagesAway(Request $request, Pool $pool)
             }
 
             $packageIndex[$package->getName()][$package->id] = $package;
+            foreach ($package->getRequires() as $requiredPackage) {
+                $requireIndex[$requiredPackage->getTarget()] = true;
+            }
         }
 
         foreach ($request->getLockedPackages() as $package) {
+            // If this locked package is no longer required by root or anything in the pool, it may get uninstalled so do not apply it's requirements
```
Seldaek, line 426:
```suggestion
            // If this locked package is no longer required by root or anything in the pool, it may get uninstalled so do not apply its requirements
```

driskell:
Benchmarks.

Using same as #9620, starting with https://github.com/drupal/recommended-project/blob/8.9.x/composer.json

```
$ composer install
...
$ composer require drupal/block_permissions:1.0.0
...
$ nano composer.json # Change block_permissions requirement to ^1.0
...
$ composer -vvv update -W drupal/block_permissions --profile
```

I did above to warm cache and then repeated the require:1.0.0, nano and update to get results below. *I am on a laptop, MacBook Pro though, so there's variances...*

```
[57.5MiB/10.24s] Dependency resolution completed in 0.004 seconds
[56.2MiB/10.24s] Analyzed 847 packages to resolve dependencies
[56.2MiB/10.24s] Analyzed 1693 rules to resolve dependencies
...
[47.8MiB/9.92s] Memory usage: 47.76MiB (peak: 113.61MiB), time: 9.92s
[47.5MiB/9.84s] Memory usage: 47.46MiB (peak: 113.36MiB), time: 9.84s
[47.2MiB/10.35s] Memory usage: 47.22MiB (peak: 113.21MiB), time: 10.35s
[47.6MiB/9.94s] Memory usage: 47.58MiB (peak: 113.46MiB), time: 9.94s
```

Below is 2.2.1 stable without the fix, it looks slower but it's probably pretty much the same nearly and only minuscule change and the variance I put down to me testing it with a laptop.

```
[52.2MiB/9.48s] Dependency resolution completed in 0.003 seconds
[50.8MiB/9.49s] Analyzed 847 packages to resolve dependencies
[50.8MiB/9.49s] Analyzed 1693 rules to resolve dependencies
...
[47.2MiB/9.91s] Memory usage: 47.2MiB (peak: 113.18MiB), time: 9.91s
[47.6MiB/10.81s] Memory usage: 47.58MiB (peak: 113.46MiB), time: 10.81s
[47.2MiB/10.04s] Memory usage: 47.2MiB (peak: 113.18MiB), time: 10.04s
[47.2MiB/11.57s] Memory usage: 47.2MiB (peak: 113.19MiB), time: 11.57s
```

Finally, when running latest main and disabling the optimisation, to ensure it is still beneficial, results show hopeful still as the rule count is massively increased without this. Granted the time saving does seem to be small still but this is a test with a single drupal module only.

```
[73.3MiB/9.75s] Dependency resolution completed in 0.030 seconds
[55.6MiB/9.76s] Analyzed 1781 packages to resolve dependencies
[55.6MiB/9.76s] Analyzed 23527 rules to resolve dependencies
...
[47.5MiB/10.17s] Memory usage: 47.49MiB (peak: 113.36MiB), time: 10.17s
[47.6MiB/11.52s] Memory usage: 47.64MiB (peak: 113.45MiB), time: 11.52s
[47.3MiB/12.05s] Memory usage: 47.29MiB (peak: 113.21MiB), time: 12.05s
[47.3MiB/11.47s] Memory usage: 47.29MiB (peak: 113.21MiB), time: 11.47s
```

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -415,9 +417,18 @@ private function optimizeImpossiblePackagesAway(Request $request, Pool $pool)
             }
 
             $packageIndex[$package->getName()][$package->id] = $package;
+            foreach ($package->getRequires() as $requiredPackage) {
+                $requireIndex[$requiredPackage->getTarget()] = true;
```
Seldaek, line 421: I think this should be moved above all the continue statements to make sure we collect requires for all packages and not only those that may be removed?

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -415,9 +417,18 @@ private function optimizeImpossiblePackagesAway(Request $request, Pool $pool)
...
+            }
         }
 
         foreach ($request->getLockedPackages() as $package) {
+            // If this locked package is no longer required by root or anything in the pool, it may get uninstalled so do not apply it's requirements
+            // In a case where a requirement WERE to appear in the pool by a package that would not be used, it would've been unlocked and so not filtered still
+            if (!isset($requireIndex[$package->getName()]) && !isset($rootRequires[$package->getName()])) {
```
Seldaek, line 428: This should foreach on getNames(false) to make sure replacers are considered as required still, I'm not sure if including providers would be safe here so rather not, unless maybe @naderman thinks otherwise

src/Composer/DependencyResolver/PoolOptimizer.php
Seldaek: Yes, I think you can just look-up into `$this->requireConstraintsPerPackage` if the name exists, that should save some cycles building up this index again here.

tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/filter-impossible-oackages-only-required.test
```diff
@@ -0,0 +1,48 @@
...
+[
+    [
+        {"name": "first/pkg", "version": "1.0.0", "require": {"second/pkg": "*"}},
+        {"name": "second/pkg", "version": "1.0.0", "require": {"third/pkg": "1.0.0"}},
+        {"name": "second/pkg", "version": "2.0.0"},
+        {"name": "third/pkg", "version": "1.0.0", "require": {"fourth/pkg": "1.0.0"}},
+        {"name": "third/pkg", "version": "2.0.0", "require": {"fourth/pkg": "2.0.0"}},
+        {"name": "fourth/pkg", "version": "1.0.0"},
+        {"name": "fourth/pkg", "version": "2.0.0"},
+        {"name": "root/req", "version": "1.0.0", "require": {"second/pkg": "2.0.0"}, "require": {"fourth/pkg": "2.0.0"}}
```
driskell, line 36: This test was mangled - will re-address the entire thing

src/Composer/DependencyResolver/PoolOptimizer.php
driskell: Using the requiredPackagesbyConstraint means this is already generated now :+1:

driskell:
@Seldaek The current set of comments I implemented, and the fixed test is properly reproducing again - no idea how I missed that test issue 😞 I will just double confirm that the scenario in the original issue is resolved still.

driskell:
Yes, this resolved the repository example in the original issue.

Seldaek:
Great, thanks, will review again when I get a chance.

naderman:
Think this may also need tests for similar situations with replace/provide to make sure it works correctly in that case?

driskell:
> Think this may also need tests for similar situations with replace/provide to make sure it works correctly in that case?

I added some tests for both those scenarios. Currently the scenarios for replace/provide are not optimised for so I tried to note that in the test. It at least means we can check that it is behaving as expected currently and if optimisation is added down the line for provide/replace the test can be updated accordingly.

It's worth noting... whilst doing these tests I noticed that "allowTransitiveDeps" doesn't work for providers. If the provider is a root requirement and is not in the "allowList", and so attempts to update a package that requires the provided package will not unlock the provider. Not sure if this is intentional...?

Seldaek:
> It's worth noting... whilst doing these tests I noticed that "allowTransitiveDeps" doesn't work for providers. If the provider is a root requirement and is not in the "allowList", and so attempts to update a package that requires the provided package will not unlock the provider. Not sure if this is intentional...?

This is intentional yes, as many things can provide a package, unlocking these was getting messy and we weren't sure if it would be expected anyway, so we decided to leave it as is.

Thanks for the updates, I'll try and review this before 2.2.3 goes out.

Seldaek:
OK turns out the name checking logic was inverted, so it would have broken if something had a replace which was not required by anything, as it would have found name not set in $this->requireConstraintsPerPackage, and it should only ignore if *none* of the names are referenced, not if any of them is.

Fixed that and added an additional test which was breaking before (although it covers the opposite case, a replacer required by the name it replaces but not by its own name, but it's the same problem really).

Otherwise looks good to me, thanks!

jrfnl:
Thank you all for all your work on this. I have tested the situations in which I ran across the issue and all look to be fixed, so 🎉

PR #10411, opened 2021-12-31
test: Add a test for not unlocking provider automatically

driskell:
In #10405 it was confirmed that a provider package is deliberately not automatically unlocked if a package that requires it is unlocked for update. However, I did not see a test for that case so I added one here.


Claudio-code (approved):

jeffwidman:
Is this still relevant? If so, curious why it wasn't merged given that it only adds a test case?

I looked at https://github.com/composer/composer/pull/10405#issuecomment-1003342452 but at least based on a cursory glance they appear to be testing slightly different things.

PR #10412, opened 2021-12-31
Fix symlink resolution in shell proxy

johnstevenson:
Fixes ```self=$(realpath $0 >/dev/null 2>&1)```, which pipes stdout to /dev/null (and is obviously not the intention).

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -397,9 +397,8 @@ PROXY;
         return <<<PROXY
 #!/usr/bin/env sh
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Haha what an idiot I made there, thanks. I was so focused on trying out the error suppression handling that I forgot to even check it was still working for the happy path :D

johnstevenson:
I'd imagine your focus levels were off the scale these past few days! Wishing you all a great New Year.

Seldaek:
It's been an eventful couple weeks for sure :D Happy new year to you too

PR #10416, opened 2022-01-03
Allow xdebug-handler 3.x, update deps

Seldaek:
cc @johnstevenson see the hack I had to do in bootstrap.php - I am not sure why this is required with v3, or rather I am not sure why it worked before :) 

But it would be nice if [isXdebugActive](https://github.com/composer/xdebug-handler/blob/main/src/XdebugHandler.php#L257) would be working standalone as it is static.

I guess all of the following could be moved into it really, I don't know why this stuff is stored on instance props and not static ones: 

https://github.com/composer/xdebug-handler/blob/ec6aa897c8b6cab05ebaeddade62bb6e4a69ef38/src/XdebugHandler.php#L94-L100

Keeping this PR open as reminder, if it gets fixed in xdebug-handler I will drop the hack on this end.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "composer/metadata-minifier": "^1.0",
         "composer/semver": "^3.0",
         "composer/spdx-licenses": "^1.2",
-        "composer/xdebug-handler": "^2.0",
+        "composer/xdebug-handler": "^2 || ^3.0.1",
         "justinrainbow/json-schema": "^5.2.11",
         "psr/log": "^1.0 || ^2.0",
         "seld/jsonlint": "^1.4",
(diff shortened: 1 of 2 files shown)


johnstevenson:
Okay, this is a bug because calling `isXdebugActive()` before the class has been created results in a type error in version 3 because `null` is returned rather than a bool, so I definitely need to fix this.

> But it would be nice if isXdebugActive would be working standalone as it is static.

Yes, I think it should do this.

>I guess all of the following could be moved into it really, I don't know why this stuff is stored on instance props and not static ones...

Indeed. It just evolved this way, as things often do.

johnstevenson:
@Seldaek Now fixed
- https://github.com/composer/xdebug-handler/releases/tag/2.0.4
- https://github.com/composer/xdebug-handler/releases/tag/3.0.1

Seldaek:
@johnstevenson thanks!

PR #10417, opened 2022-01-03
Remove codeblocks for php 5 compatibilities

Nielsvanpach:
This PR removes a couple of codeblocks that exists just to keep the codebase compatible with deprecated php versions.
Let me know if a different behaviour is needed anywhere instead of removal, I'll be happy to update the PR!

### DIFF
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -293,7 +293,7 @@ class ClassMapGenerator
     {
         static $extraTypes = null;
         if (null === $extraTypes) {
-            $extraTypes = PHP_VERSION_ID < 50400 ? '' : '|trait';
+            $extraTypes = '|trait';
             if (PHP_VERSION_ID >= 80100 || (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.3', '>='))) {
                 $extraTypes .= '|enum';
             }
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -292,13 +292,12 @@ class ClassMapGenerator
     private static function getExtraTypes()
     {
         static $extraTypes = null;
-        if (null === $extraTypes) {
-            $extraTypes = '|trait';
-            if (PHP_VERSION_ID >= 80100 || (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.3', '>='))) {
-                $extraTypes .= '|enum';
-            }
-            PhpFileCleaner::setTypeConfig(array_merge(array('class', 'interface'), array_filter(explode('|', $extraTypes))));
+
+        if (PHP_VERSION_ID >= 80100 || (defined('HHVM_VERSION') && version_compare(HHVM_VERSION, '3.3', '>='))) {
+            $extraTypes .= '|enum';
         }
+
+        PhpFileCleaner::setTypeConfig(array_merge(['trait', 'class', 'interface'], array_filter(explode('|', $extraTypes))));
 
         return $extraTypes;
     }
--- a/src/Composer/Autoload/ClassMapGenerator.php
+++ b/src/Composer/Autoload/ClassMapGenerator.php
@@ -241,7 +241,7 @@ class ClassMapGenerator
         }
 
         // return early if there is no chance of matching anything in this file
-        Preg::matchAll('{\b(?:class|interface'.$extraTypes.')\s}i', $contents, $matches);
+        Preg::matchAll('{\b(?:class|interface|trait'.$extraTypes.')\s}i', $contents, $matches);
         if (!$matches) {
             return array();
         }
@@ -252,7 +252,7 @@ class ClassMapGenerator
 
         Preg::matchAll('{
             (?:
-                 \b(?<![\$:>])(?P<type>class|interface'.$extraTypes.') \s++ (?P<name>[a-zA-Z_\x7f-\xff:][a-zA-Z0-9_\x7f-\xff:\-]*+)
+                 \b(?<![\$:>])(?P<type>class|interface|trait'.$extraTypes.') \s++ (?P<name>[a-zA-Z_\x7f-\xff:][a-zA-Z0-9_\x7f-\xff:\-]*+)
                | \b(?<![\$:>])(?P<ns>namespace) (?P<nsname>\s++[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*+(?:\s*+\\\\\s*+[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*+)*+)? \s*+ [\{;]
             )
         }ix', $contents, $matches);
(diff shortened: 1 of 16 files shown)


src/Composer/Command/DiagnoseCommand.php
```diff
@@ -693,8 +693,8 @@ private function checkPlatform()
                         break;
 
                     case 'php':
-                        $text = "Your PHP ({$current}) is quite old, upgrading to PHP 5.3.4 or higher is recommended.".PHP_EOL;
-                        $text .= " Composer works with 5.3.2+ for most people, but there might be edge case issues.";
+                        $text = "Your PHP ({$current}) is quite old, upgrading to PHP 7.2.5 or higher is recommended.".PHP_EOL;
+                        $text .= " Composer works with 7.2.5+ for most people, but there might be edge case issues.";
```
Tobion, line 697: not sure if the edge case issues are still relevant with php 7.2+

src/Composer/Json/JsonFile.php
```diff
@@ -238,35 +238,12 @@ public function validateSchema($schema = self::STRICT_SCHEMA, $schemaFile = null
...
 
-        $prettyPrint = (bool) ($options & self::JSON_PRETTY_PRINT);
-        $unescapeUnicode = (bool) ($options & self::JSON_UNESCAPED_UNICODE);
-        $unescapeSlashes = (bool) ($options & self::JSON_UNESCAPED_SLASHES);
-
-        if (!$prettyPrint && !$unescapeUnicode && !$unescapeSlashes) {
-            return $json;
-        }
-
-        return JsonFormatter::format($json, $unescapeUnicode, $unescapeSlashes);
```
Tobion, line 269: JsonFormatter class can be removed

src/Composer/Json/JsonFile.php
```diff
@@ -238,35 +238,12 @@ public function validateSchema($schema = self::STRICT_SCHEMA, $schemaFile = null
...
-
-        $json = json_encode($data);
+        $json = json_encode($data, $options);
         if (false === $json) {
             self::throwEncodeError(json_last_error());
         }
 
-        $prettyPrint = (bool) ($options & self::JSON_PRETTY_PRINT);
-        $unescapeUnicode = (bool) ($options & self::JSON_UNESCAPED_UNICODE);
-        $unescapeSlashes = (bool) ($options & self::JSON_UNESCAPED_SLASHES);
```
Tobion, line 263: these constants should be removed

src/Composer/Command/DiagnoseCommand.php
Seldaek: Indeed this whole `case` can be deleted, along with the whole `if` block at line 560 which defines `$warnings['php]`

src/Composer/Autoload/ClassMapGenerator.php
```diff
@@ -293,7 +293,7 @@ private static function getExtraTypes()
     {
         static $extraTypes = null;
         if (null === $extraTypes) {
-            $extraTypes = PHP_VERSION_ID < 50400 ? '' : '|trait';
+            $extraTypes = '|trait';
```
Seldaek, line 296: You could move `trait` into the default array below together with class and interface at line 300.

src/Composer/Json/JsonFile.php
Nielsvanpach: They are both public and still referenced to in the codebase. Will need some more work to clean it up.

src/Composer/Autoload/ClassMapGenerator.php
```diff
@@ -292,14 +292,13 @@ private static function findClasses($path)
     private static function getExtraTypes()
     {
         static $extraTypes = null;
-        if (null === $extraTypes) {
```
Seldaek, line 295: This guard should not have been removed, now this code will execute every time getExtraTypes is called instead of being initialized just once

src/Composer/Autoload/ClassMapGenerator.php
```diff
@@ -292,14 +292,13 @@ private static function findClasses($path)
...
-            $extraTypes = PHP_VERSION_ID < 50400 ? '' : '|trait';
```
Seldaek, line 296: This also still needs to be there, to at least initialize $extraTypes to `''`

src/Composer/Console/Application.php
```diff
@@ -252,8 +252,8 @@ public function doRun(InputInterface $input, OutputInterface $output): int
                 function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknown OS'
             ), true, IOInterface::DEBUG);
 
-            if (PHP_VERSION_ID < 50302) {
-                $io->writeError('<warning>Composer only officially supports PHP 5.3.2 and above, you will most likely encounter problems with your PHP '.PHP_VERSION.', upgrading is strongly recommended.</warning>');
+            if (PHP_VERSION_ID < 70205) {
+                $io->writeError('<warning>Composer only officially supports PHP 7.2.5 and above, you will most likely encounter problems with your PHP '.PHP_VERSION.', upgrading is strongly recommended.</warning>');
```
Seldaek, line 256:
```suggestion
                $io->writeError('<warning>Composer supports PHP 7.2.5 and above, you will most likely encounter problems with your PHP '.PHP_VERSION.'. Upgrading is strongly recommended but you can use Composer 2.2.x LTS as a fallback.</warning>');
```

src/Composer/Json/JsonFormatter.php
```diff
@@ -21,6 +21,8 @@
  *
  * @author Konstantin Kudryashiv <ever.zet@gmail.com>
  * @author Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * @abandoned
```
Seldaek, line 25:
```suggestion
 * @abandoned Use json_encode or JsonFile::encode() with modern JSON_* flags to configure formatting - this class will be removed in 3.0
```

tests/Composer/Test/Json/JsonFileTest.php
```diff
@@ -298,7 +298,7 @@ public function testEscape()
 
     public function testUnicode()
     {
-        if (!function_exists('mb_convert_encoding') && PHP_VERSION_ID < 50400) {
+        if (!function_exists('mb_convert_encoding')) {
             $this->markTestSkipped('Test requires the mbstring extension');
         }
```
Seldaek, line 303: This is the correct transform here, as PHP_VERSION_ID is always going to be > 50400 we never need to skip the test

```suggestion
```

tests/Composer/Test/Json/JsonFileTest.php
```diff
@@ -312,7 +312,7 @@ public function testUnicode()
 
     public function testOnlyUnicode()
     {
-        if (!function_exists('mb_convert_encoding') && PHP_VERSION_ID < 50400) {
+        if (!function_exists('mb_convert_encoding')) {
             $this->markTestSkipped('Test requires the mbstring extension');
         }
```
Seldaek, line 317:
```suggestion
```

Seldaek:
Thanks! Applied a few tweaks to quickly get it mergeable but nice work here overall :)

src/Composer/Json/JsonFormatter.php
```diff
@@ -21,6 +21,8 @@
  *
  * @author Konstantin Kudryashiv <ever.zet@gmail.com>
  * @author Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * @abandoned Use json_encode or JsonFile::encode() with modern JSON_* flags to configure formatting - this class will be removed in 3.0
```
stof, line 25: should be `@deprecated` rather than `@abandoned`, which is the proper name used in phpdoc (the `@abandoned` tag does not exist AFAIK)

Seldaek: Yup noticed too already fixed on my machine

PR #10419, opened 2022-01-03
explicitly add public to constants

Nielsvanpach:
This PR explicitly adds 'public' to all constants in the codebase. Made use of Rector for this.


Seldaek:
Thanks but this sort of carpet-bombing rector I am not a huge fan of. This does not add any value to the code and it only means we don't know anymore which constants are really meant to be public and which just happened to be because it's so by default. I'd rather go through them all myself and make a conscious decision and mark private/protected where appropriate.

PR #10421, opened 2022-01-03
replace all string class names by ::class constants

Nielsvanpach:

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -366,7 +366,7 @@ EOF;
             );
         }
 
-        $classMap['Composer\\InstalledVersions'] = "\$vendorDir . '/composer/InstalledVersions.php',\n";
+        $classMap[\Composer\InstalledVersions::class] = "\$vendorDir . '/composer/InstalledVersions.php',\n";
         ksort($classMap);
         foreach ($classMap as $class => $code) {
             $classmapFile .= '    '.var_export($class, true).' => '.$code;
--- a/src/Composer/Command/BaseCommand.php
+++ b/src/Composer/Command/BaseCommand.php
@@ -282,7 +282,7 @@ abstract class BaseCommand extends Command
      */
     protected function getTerminalWidth()
     {
-        if (class_exists('Symfony\Component\Console\Terminal')) {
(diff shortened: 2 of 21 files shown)


ricardoboss (approved):

Seldaek:
I don't know if I am getting old or what, but I don't see a whole lot of added value here?

This is more characters, and phpstan will already check that the strings are valid class names when passed to things expecting a class-string, so no gain there.

GromNaN:
This helps IDE to link usage and definition of classes. They are not all as good as phpstan and phpstorm (thinking to intelephense).

It should be done using `use` statement for clarity.

Seldaek:
Right, if done correctly with use statements it sounds better, but the simple rector transform here just look messy IMO.

Nielsvanpach:
Agree. I found a way to transform this into use statements with Rector, but it changes all codeblocks with FCQN to use statements. Result: 159 classes are updated.

I've made the changes on a separate branch, I'll open a PR if you are interested to merge this. I had to make a few adjustments to satisfy Phpstan, but overall all checks passed: https://github.com/composer/composer/compare/main...Nielsvanpach:replace-string-class-names?expand=1

GromNaN:
You can reduce changes if you:
- skip root classes (like `\InvalidArgumentException`)
- keep partial namespaces when already in use (`Composer\Command` registration, `Composer\DependencyResolver\Operation`, `Composer\Downloader`)
- keep functions for later (`React\Promise\resolve`)

Nielsvanpach:
Indeed. However, this keeps it consistent within the codebase, so I personally wouldn't mind all the changed files. And afaik the auto import names option is not customisable with Rector.

GromNaN:
I'd say that rector needs to be improved if we want to use it to make this changes.

Seldaek:
> so I personally wouldn't mind all the changed files

I do mind as it changes tons of lines that don't need changing (which means potential for merge conflicts increases), and adding use statements for root classes is often quite superfluous IMO. 

We'll revisit in a few months perhaps once 2.2 is less active, but I can run rector or php-cs-fixer myself if needed really.. I'd rather do it and review the changes as I commit them than reviewing someone else's run which requires much more attention.

Nielsvanpach:
> I'd rather do it and review the changes as I commit them than reviewing someone else's run which requires much more attention.

Understandable!

Issue #9224, opened 2020-09-17
End of php52 supporting


sveneld:
7 years ago such issue was created https://github.com/composer/composer/issues/1730
As a result such code is using now in production
`$vendorPathCode52 = str_replace('__DIR__', 'dirname(__FILE__)', $vendorPathCode);`

Mayby it's time to remove this code and say good bay to php5.2?

Using `dirname(__FILE__)` is also slower 
```
$startTime = microtime(true);
for ($i=1;$i<1000000;$i++){
    $a = dirname(__FILE__);
}
echo microtime(true) - $startTime;

echo PHP_EOL;
$startTime = microtime(true);
for ($i=1;$i<1000000;$i++){
    $a = __DIR__;
}
echo microtime(true) - $startTime;
```

Result
0.95782494544983
0.14195108413696

xabbuh:
see #8785

sveneld:
@xabbuh thanks, but as i can see in discussion minimum version is 5.3. And minimum version in composer.json for composer is also "php": "^5.3.2

stof:
AFAICT, the only places supporting PHP 5.2 are the generated `vendor/autoload_{files,namespaces,classmap}.php` files. As these are not used at all anymore when using the `vendor/autoload.php` file on PHP 5.6+ (it reads `vendor/autoload_static.php` instead), there is no performance impact.

So my suggestion is to keep that legacy feature until Composer 2.1 and to clean this at the same time than other things.

Wirone:
Probably duplicate of #9303 @Seldaek

Issue #9303, opened 2020-10-19
Drop support for older PHP releases


Seldaek:
As per #8785 - for 2.1 or 2.2, bump php requirement to 7.1.3, and bump all dependencies.

Grep for PHP_VERSION_ID to find stuff needing fixing.

Edit May 2021: Probably targetting PHP 7.2.5 now for Composer 2.3 as the timelines have slipped.

GrahamCampbell:
Going to PHP 5.5.9 with Symfony 3.4 in Composer 2.1 might be a good idea, in particular, because PHP 8 support was never implemented on Symfony 2.8. Then maybe go all the way in a later release to PHP 7.1.3 with Symfony 4.4. Then again, I guess you don't want to end up with more than one LTS series of Composer 2.x...

stof:
Given that Symfony 3.4 reaches end of bugfix support at the end of November, I doubt making it a target for the version used in Composer 2.1 is a good idea: it would probably already be in security-only support by the time composer 2.1 gets released.

GrahamCampbell:
Certainly better than Symfony 2.8 though, which is what we currently have.

simonberger:
Raising support to PHP 7.1.3 is a nice step. It gives much room to improve the code with new PHP language and dependency features and isn't that new leave too many behind.

In my opinion sticking to any PHP 5.X version and with it using Symfony 3 instead of 4 just slows down improvements. The long planned support for Composer 2.0 should fill the gap enough.

morrisonlevi:
How long do you plan to support 2.0.x for bug and security fixes? As already noted, PHP 7.1 isn't an LTS version on any known OS that do LTS's: https://github.com/composer/composer/issues/8785#issuecomment-635240404. Using Symfony 5 means using a supported Symfony release longer than Symfony 4.4.

Circling back to my opening question:
  1. There are LTS users on PHP 5.5 and 7.0 and you already plan to support them through Composer 2.0 (for some unknown amount of time)
  2. PHP 7.2 is the next version with widespread LTS support
  3. I don't see any reason PHP 7.1 can't use the same security updates on composer 2.0 meant for the other outdated PHP versions.

IMO, the other option to support PHP 7.0 as it also has LTS support, but since it seems you've already ruled that out it makes sense to solidify on the next major LTS, which isn't PHP 7.1.

Seldaek:
[...]

PR #10428, opened 2022-01-04
Deprecations & more php version cleanups

Seldaek:
- Fixes the last bits of https://github.com/composer/composer/issues/9303 and introduces a few deprecations (JsonFormatter, JsonFile::* constants, TlsHelper)
- Drops PHP <7.2, HHVM and ZendGuard (as it never supported PHP 7+) support from autoload_real, so that autoload_static.php is always used. Other files are kept for BC as I suspect other tooling may rely on them. I do not plan on removing them for now.
- Closes https://github.com/composer/composer/pull/9433 as it is included in this PR, which fixes https://github.com/composer/composer/issues/9224 too


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -981,7 +980,7 @@ public static function getLoader()
 
         $file .= <<<CLASSLOADER_INIT
         spl_autoload_register(array('ComposerAutoloaderInit$suffix', 'loadClassLoader'), true, $prependAutoloader);
-        self::\$loader = \$loader = new \\Composer\\Autoload\\ClassLoader(\\dirname(\\dirname(__FILE__)));
```
Chrico, line 985: Is there any reason why `dirname(dirname(__FILE__))` was replaced by just `dirname(__FILE__)` ? It looks like this is breaking @Seldaek with version 2.3.0 of Composer for `autoload.files` for us. Currently investigating - will add a ticket as soon as i got more information, but curious about that change 🤔

Seldaek: Because `dirname(__FILE__)` is the equivalent of `__DIR__`? So this should not be an issue. Most likely you'ae having issues due to https://github.com/composer/composer/pull/10617 perhaps? Check if you can repro with 2.2.10 as well, and if 2.2.7 fixes it?

Chrico: It looks like that not this is the problem. With Composer 2.2.10 it is working again.

The problem for us is, that we're using our custom Composer Plugin https://github.com/inpsyde/vip-composer-plugin , 
 which replaces the paths in those `autoload_*.php` files. But it seems those files are not used anymore in Composer v2.3.?

---

https://github.com/inpsyde/vip-composer-plugin/issues/7

PR #10430, opened 2022-01-04
Add links to repo to show/outdated commands package lists

### LINKED ISSUE #9367: Suggestion: Display package's "homepage" instead of "description" in `composer show` list
I'm suggesting to replace the last column of the output table of `composer show` and `composer outdated`: Show the packages website (i.e. `homepage`) instead of its `description`

Reasons:
* When I'm looking at that list, most of the time I need more information about the package. So I usually copy the package name into my web browser to search for it. An url which is clickable right away would be quicker.
* The `description` of most packages is non-saying, If I really don't know what this package is, its description certainly won't tell me. To find out, I'd have to, well, visit its website :-)
* While not fully solving https://github.com/composer/composer/issues/8104, it would at least be a step in this direction.

Which URL to display?
If given, use [`homepage`](https://getcomposer.org/doc/04-schema.md#homepage) from `composer.json`. If not, use the package's packagist.org URL.

Seldaek:
fixes #9367

image

If someone thinks of another good use case for this, do tell.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -904,10 +904,19 @@ EOT
             if ($showSource) {
                 $source = ' (' . $this->config->getSourceOfValue($k . $key) . ')';
             }
-            if (is_string($rawVal) && $rawVal != $value) {
-                $io->write('[<comment>' . $k . $key . '</comment>] <info>' . $rawVal . ' (' . $value . ')</info>' . $source, true, IOInterface::QUIET);
+
+            if (null !== $k && 0 === strpos($k, 'repositories')) {
+                $link = 'https://getcomposer.org/doc/05-repositories.md';
             } else {
-                $io->write('[<comment>' . $k . $key . '</comment>] <info>' . $value . '</info>' . $source, true, IOInterface::QUIET);
+                $id = Preg::replace('{\..*$}', '', $k === '' || $k === null ? (string) $key : $k);
+                $id = Preg::replace('{[^a-z0-9]}i', '-', strtolower(trim($id)));
+                $id = Preg::replace('{-+}', '-', $id);
+                $link = 'https://getcomposer.org/doc/06-config.md#' . $id;
+            }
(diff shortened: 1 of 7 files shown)


ThomasLandauer:
The general problem here is that there are two "standards" on how to display links in a console:
1. Old-fashioned: Display the URL verbatim: `https://github.com/symfony/...`, and then rely on the terminal's auto-detection to make it clickable. This is (as far as I know) supported by *every* terminal.
2. Fancy (i.e. HTML-like): Add some special characters to allow displaying a different link text. This is supported by *some* terminals: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

I don't know how the adoption of (2) will progress. But for Konsole (=default in KDE) it doesn't look too good: https://bugs.kde.org/show_bug.cgi?id=379294
And (frankly) I'm also not sure if it's a good idea at all...

What I originally had in mind in https://github.com/composer/composer/issues/9367 was to display the link *instead* of the description:
```
symfony/polyfill-ctype   v1.23.0   https://github.com/symfony/...
```
Or maybe introduce some configuration to let the user decide?

Seldaek:
Yes to me the critical links that need to be seen need to be of the first type as even the dumbest terminal will let you copy paste the url. Here though it's kinda nice to have functionality and I feel like this approach lets us pack more info and functionality in the same amount of text/output which is good.

Letting the user decide is always an option but it also adds complexity across the board (in code, docs, cognitive for users,..) so in this case I don't particularly feel it's justified as one can also easily get the info somewhere else if needed.

staabm:
Looking at it, it might be usefull to have thr version number link to packagist

Seldaek:
@staabm that's unfortunately quite messy to do as we don't have a direct link to packagist available there.

I have now added more links in various other commands though.

Seldaek:
Think I am going to merge as is, sorry @ThomasLandauer if this doesn't quite match what you had in mind, but note that you can use --format=json output to get to the URLs otherwise. Combined with [jq](https://stedolan.github.io/jq/) this should kinda easily let you script this yourself.

PR #10432, opened 2022-01-04
Fix #10366: Improve messaging when GitHub tokens need SSO authorization

### LINKED ISSUE #10366: Improve messaging when GitHub tokens need SSO authorization
Composer could be more helpful if you are part of a GitHub organization but forget to enable SSO on your GitHub token. Ideally Composer would detect that SSO is not enabled on the token, or at least mention in the error message somewhere that lack of SSO could be the cause.

Specifically, if the following are true:
- You are part of a GitHub organization that requires SSO
- You have configured Composer to use a GitHub token
- You've not enabled SSO for that token
- You try to require a package from that organization (_even if the package is public_)

... then you get an error message:
> Could not fetch [package], please review your configured GitHub OAuth token or enter a new one to access private repos

This is especially confusing when trying to download a _public package_. I know the error message says in part "review your configured token", but it's easy to infer that the token is _missing_ rather than misconfigured. And then go through an endless loop of wondering why a token is even required for a public package, creating and setting a new token, and getting the same error message over and over until you realize the problem is missing SSO.

This might sound a bit like an edge case but I promise you nearly everyone in an SSO org has gone through this at least once (and in my case, multiple times 😄 )

danepowell:
Fix #10366

This might be a little rough, I tried to keep it consistent with the rest of the auth helper methods which I found a little hard to parse. But if you're happy with the overall approach and provide feedback, I'll clean it up.

### DIFF
--- a/src/Composer/Util/AuthHelper.php
+++ b/src/Composer/Util/AuthHelper.php
@@ -92,6 +92,22 @@ class AuthHelper
             $message = "\n";
 
             $rateLimited = $gitHubUtil->isRateLimited($headers);
+            $requiresSso = $gitHubUtil->requiresSso($headers);
+
+            if ($requiresSso) {
+                $ssoUrl = $gitHubUtil->getSsoUrl($headers);
+                $message = sprintf(
+                        'GitHub API token requires SSO authorization. Authorize this token at ' . $ssoUrl,
+                        $ssoUrl
+                    ) . "\n";
+                $this->io->writeError($message);
+                if (!$this->io->isInteractive()) {
+                    throw new TransportException('Could not authenticate against ' . $origin, 403);
+                }
+                $this->io->ask('After authorizing your token, confirm that you would like to retry the request');
+                return ['retry' => TRUE, 'storeAuth' => $storeAuth];
+            }
+
             if ($rateLimited) {
                 $rateLimit = $gitHubUtil->getRateLimit($headers);
                 if ($this->io->hasAuthentication($origin)) {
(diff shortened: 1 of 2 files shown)


src/Composer/Util/GitHub.php
```diff
@@ -171,6 +171,27 @@ public function getRateLimit(array $headers)
...
+        foreach ($headers as $header) {
+            $header = trim($header);
+            if (false === strpos($header, 'x-github-sso: required')) {
+                continue;
+            }
+            list(, $url) = explode('=', $header, 2);
+            return $url;
+        }
+
+        return '?';
```
stof, line 192: this should return null, not a `'?'` string

danepowell: That's fine, I was just following the example of [getRateLimit](https://github.com/composer/composer/pull/10432/files#diff-dd5b1f87d789c32cce81100ce282931f6af3c0894d755837c483d7eaf67e33c8R148), which also returns '?'. Both methods seem like poor UX, there needs to be better error handling here and in the auth helper.

Seldaek:
Thanks

src/Composer/Util/AuthHelper.php
```diff
@@ -92,6 +92,23 @@ public function promptAuthIfNeeded($url, $origin, $statusCode, $reason = null, $
             $message = "\n";
 
             $rateLimited = $gitHubUtil->isRateLimited($headers);
+            $requiresSso = $gitHubUtil->requiresSso($headers);
+
+            if ($requiresSso) {
+                $ssoUrl = $gitHubUtil->getSsoUrl($headers);
+                $message = sprintf(
```
levu42, line 99: Here you're using sprintf but you're not really using it, you just concatenate the strings. Is that supposed to be like that? @danepowell @Seldaek

Seldaek: Yeah that looks like a mistake, although it's functional. I'll fix it when I get a chance

PR #10440, opened 2022-01-06
GitLab: detect repository functionality disabled in Driver

glaubinix:
Can otherwise cause `Warning: Undefined array key "default_branch" /src/Composer/Repository/Vcs/GitLabDriver.php:287`


Seldaek:
Thanks

PR #10449, opened 2022-01-10
Use new url for semver testing tool

jdrieghe:
### Changed
- [semver.mwl.be](https://semver.mwl.be) returns a permanent redirect to [semver.madewithlove.com](https://semver.madewithlove.com). We should use the target domain directly.


Seldaek:
Thanks

PR #10454, opened 2022-01-13
Allow psr/log ^3.0

simPod:
Same as in https://github.com/composer/composer/pull/10158

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "composer/spdx-licenses": "^1.2",
         "composer/xdebug-handler": "^2 || ^3.0.1",
         "justinrainbow/json-schema": "^5.2.11",
-        "psr/log": "^1.0 || ^2.0",
+        "psr/log": "^1.0 || ^2.0 || ^3.0",
         "seld/jsonlint": "^1.4",
         "seld/phar-utils": "^1.2",
         "symfony/console": "^5.4.1 || ^6.0",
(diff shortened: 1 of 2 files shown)


Seldaek:
To allow psr/log 3 we'd have to add the PHP 8 type hints IIRC?

simPod:
@Seldaek true, I missed that composer/composer implements that interface as well.

Not sure what is the possible upgrade path here, I guess it is not possible to support (v1+v2) and v3 at the same time as method signatures are incompatible?

stof:
you can support v1+v2+v3 at the same time with 2 conditions:

- your code requires PHP 7.2+ to have the covariance rules
- your code implements the methods with return types (to support v3) but without argument types (to support v1)

The first condition is not met in the composer 2.2 LTS branch.

simPod:
I've changed target to master

stof:
not having changes in the BaseIO class is wrong. It does not actually make it compatible with v3 (see the second condition I listed in my previous comment)

simPod:
@stof yup I did but the return types are already there 🤔 So since it was compatible now I assume it still is. Removing `array` from `array $context` should not be necessary, or?

stof:
the `array` argument type is already there in the v1 signature. So it is indeed fine to have it. the `without argument types` only applies to the new ones being added in psr/log.

> the return types are already there

indeed sorry. I looked at the wrong composer branch.

Seldaek:
Ok let's give this a shot, I forgot it was possible with 7.2 by ignoring the $message type. Thanks!

PR #10455, opened 2022-01-13
Move "packagist.org" to named property

wbars:
This is the only allowed syntax according to the https://getcomposer.org/doc/05-repositories.md#disabling-packagist-org and existence of property name will enable code assistance for tools that are work based on this schema, for example IDE that provide code completion

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -181,10 +181,12 @@
         "repositories": {
             "type": ["object", "array"],
             "description": "A set of additional repositories where packages can be found.",
+            "properties": {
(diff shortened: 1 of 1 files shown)


stof:
This can also be used to disable a repository registered under a different name. `packagist.org` is not the only repo you can disable

wbars:
I see. Misread documentation, thanks.

stof:
`packagist.org` is the only repo name configured in the default composer config. But you might have some other named repos in your global config and disabling them again at the project level.

PR #10458, opened 2022-01-14
Disable "composer/package-versions-deprecated" by default

nicolas-grekas:
Instead of  #10354

Inspired by @stof's comment at https://github.com/composer/composer/pull/10354#issuecomment-1013108766

The goal here is to free the ecosystem from having to reference a legacy plugin in their composer.json files.

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -699,6 +699,10 @@ class PluginManager
             }
         }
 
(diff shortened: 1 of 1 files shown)


mvorisek:
👍 for this change as otherwise `composer/package-versions-deprecated` allow-plugin config is needed and the package is deprecated (basically for composer v1.x only)

Seldaek:
Thanks, this approach makes more sense yes.

PR #10463, opened 2022-01-17
report error if binary is a directory

dbu:
file_exists is true also for directory and symlink. but later in generateUnixyProxyCode we call `file_get_contents` on the binary, which fails with `file_get_contents(): read of 8192 bytes failed with errno=21 Is a directory` if the binary is a directory.

maybe we can just change file_exists to is_file, but that would probably mess with symlinks.

this problem only started to occur with composer 2.2, in 2.1.16, the problem was silently ignored. because the error in 2.2 is a fatal error, we get no context about the composer package that causes the problem, nor the path that is a directory instead of a file.

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -77,6 +77,10 @@ class BinaryInstaller
                 $this->io->writeError('    <warning>Skipped installation of bin '.$bin.' for package '.$package->getName().': file not found in package</warning>');
                 continue;
             }
+            if (is_dir($binPath)) {
+                $this->io->writeError('    <warning>Skipped installation of bin '.$bin.' for package '.$package->getName().': found a directory at that path</warning>');
+                continue;
+            }
             if (!$this->filesystem->isAbsolutePath($binPath)) {
                 // in case a custom installer returned a relative path for the
                 // $package, we can now safely turn it into a absolute path (as we


Seldaek:
Always amazed to see what sort of weird broken stuff was "working" for ages 😄  Have you identified the source package of the problem by now?

dbu:
yep. when running with `-vvv`, it is the last package mentioned to be synced before the crash :-) thats also how i am sure that this was our problem.

my guess would be that composer until 2.2 did not try to read the binary file contents...

dbu:
hm, not sure if this is correct at all. i just realized that in our application, it was on purpose that we used the `bin` instruction to link a whole folder (of bash scripts) into the `vendor/bin` directory. that seems to have worked fine until composer 2.1.x.

could we not try to read the file when its a directory and instead just symlink the whole directory again? not sure if you consider the previous behaviour a bug - seems like one of those "your bugfix is a BC break for me" things.

stof:
> that seems to have worked fine until composer 2.1.x.

only on systems where composer was using a symlink, which has never been the case cross-platform

Seldaek:
Yeah that was another case of abusing the symlink functionality.. and I don't really see a way to fix this one except by restoring the symlinking but I'd really rather not go there. If you have multiple binaries to define why not list them all in composer.json and they are then accessible in vendor/bin/x,y,z..?

dbu:
> only on systems where composer was using a symlink, which has never been the case cross-platform

all our servers are linux systems, so symlinks it was.

> If you have multiple binaries to define why not list them all in composer.json 

in our case: it is a folder for deployment hooks. it has subfolders again, with each subfolder denoting at which point of the deployment the scripts in that folder have to be run.

however, we now set the deployment folder to the actual bundle path in vendor. that is leaking the directory structure in the bundle into the application using it, but as thats an internal bundle anyways, we can totally live with the situation.

if you want ultimate BC, you could fall back to symlinks when the bin is a directory. but i guess not that many people did the thing we did... if we throw a clear error that should be good enough imho.

src/Composer/Installer/BinaryInstaller.php
```diff
@@ -77,6 +77,10 @@ public function installBinaries(PackageInterface $package, $installPath, $warnOn
                 $this->io->writeError('    <warning>Skipped installation of bin '.$bin.' for package '.$package->getName().': file not found in package</warning>');
                 continue;
             }
+            if (is_dir($binPath)) {
```
dbu, line 80: do you have a test whether the binary installer works if the binary target is a symlink?

src/Composer/Installer/BinaryInstaller.php
```diff
@@ -77,6 +77,10 @@ public function installBinaries(PackageInterface $package, $installPath, $warnOn
                 $this->io->writeError('    <warning>Skipped installation of bin '.$bin.' for package '.$package->getName().': file not found in package</warning>');
```
dbu, line 77: should we use is_readable instead of file_exists, as we later try to read the contents of the binary file?

src/Composer/Installer/BinaryInstaller.php
Seldaek: We have very few tests of this code.. and none covering that case no.

src/Composer/Installer/BinaryInstaller.php
Seldaek: Having an unreadable file sounds very unlikely as things were typically just written by Composer itself. Not sure I would care about such a possible edge case here.

Seldaek:
Overall I think I would merge as is, it covers an unusual use case with a clear error. I don't think we need much more?

dbu:
if you don't want to re-support folder symlinks, i think we can merge as is, yes. it gives a better error than the cryptic failure one got before.

Seldaek:
Yeah I'd rather not restore symlink support unless there's a major use case for it, but given you're the first to mention it in a month since release, I think it's probably not the case. It made the whole thing fairly unreliable as it was platform dependent, now the bin proxies look the same on every platform which is much more predictable.

Anyway thanks for the PR!

PR #10470, opened 2022-01-19
ArrayLoader: handle none string values for version/version_normalized

glaubinix:
Asserting that `Package::getVersion/Package::getPrettyVersion` always return a string. Currently this can return bool/int/float/string.

Additionally catching cases where version is not an array/object which currently throws a `TypeError : trim(): Argument #1 ($string) must be of type string, array given`

### DIFF
--- a/src/Composer/Package/Loader/ArrayLoader.php
+++ b/src/Composer/Package/Loader/ArrayLoader.php
@@ -113,12 +113,12 @@ class ArrayLoader implements LoaderInterface
         if (!isset($config['name'])) {
             throw new \UnexpectedValueException('Unknown package has no name defined ('.json_encode($config).').');
         }
-        if (!isset($config['version'])) {
+        if (!isset($config['version']) || !is_scalar($config['version'])) {
             throw new \UnexpectedValueException('Package '.$config['name'].' has no version defined.');
         }
 
         // handle already normalized versions
-        if (isset($config['version_normalized'])) {
+        if (isset($config['version_normalized']) && is_string($config['version_normalized'])) {
             $version = $config['version_normalized'];
 
             // handling of existing repos which need to remain composer v1 compatible, in case the version_normalized contained VersionParser::DEFAULT_BRANCH_ALIAS, we renormalize it
@@ -129,7 +129,7 @@ class ArrayLoader implements LoaderInterface
             $version = $this->versionParser->normalize($config['version']);
         }
 
-        return new $class($config['name'], $version, $config['version']);
+        return new $class($config['name'], $version, (string) $config['version']);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -113,12 +113,12 @@ private function createObject(array $config, $class)
         if (!isset($config['name'])) {
             throw new \UnexpectedValueException('Unknown package has no name defined ('.json_encode($config).').');
         }
-        if (!isset($config['version'])) {
+        if (!isset($config['version']) || !is_scalar($config['version'])) {
```
stof, line 116: I would even use `is_string` here. A valid package metadata contains a strict here, not other scalars (same for `version_normalized`)

tests/Composer/Test/Package/Loader/ArrayLoaderTest.php
```diff
@@ -314,4 +314,15 @@ public function testPluginApiVersionDoesSupportSelfVersion()
         $this->assertArrayHasKey('composer-plugin-api', $links);
         $this->assertSame('6.6.6', $links['composer-plugin-api']->getConstraint()->getPrettyString());
     }
+
+    public function testNoneStringVersion()
+    {
+        $config = array(
+            'name' => 'A',
+            'version' => 1,
```
stof, line 322: This is not a valid package metadata

src/Composer/Package/Loader/ArrayLoader.php
glaubinix: Not sure this is a good idea. There are packages in popular repositories using int values as versions. This would essentially break those packages.

glaubinix: Not on any package on packagist.org but on wpackagist. Hard to say how common this is in other repositories.

Further limiting `version_normalized` could potentially break existing setups. For instance, the composer.json below successfully completes a `composer update` with the latest Composer release:

```
{
    "repositories": [
        {"type": "package", "package": {
            "name": "smarty/smarty",
            "version": "3",
            "version_normalized": 3,
            "dist": {
                "url": "https://www.smarty.net/files/Smarty-3.1.7.zip",
                "type": "zip"
            }
        }}
    ],
    "require": {
        "smarty/smarty": "*"
    }
}
```

glaubinix: Oh, but I think as long as we just overwrite `version_normalized` with a wrong type and not error then this should be ok.

stof: then wpackagist should really be fixed.

Providing a `version_normalized` that is not actually the normalized version as computed by Composer will break things when comparing versions (for inline packages, it is recommended to omit it and let composer compute it, but for composer repositories like packagist.org that have lots of packages, pre-computing the normalized version server-side helps with performance). And a valid normalized version **cannot** be casted as int, as it always have 4 segments.

PR #10475, opened 2022-01-21
Clean up properly if self-update fails

johnstevenson:
If self-update fails because the downloaded file cannot be renamed or copied, it is left wherever it was downloaded to. Because it was always given the same name (`composer-temp.phar`), this didn't seem much of a problem - and could allow the user to manually move it if things got hairy.

However, it is now given a randomized name (3bb78fd1e) which produces multiple files, so this fix deletes the downloaded file on self-update failure.


Seldaek:
Makes sense 👍🏻

PR #10476, opened 2022-01-21
Add phpstan-symfony & some refactorings

Seldaek:
- Added phpstan-symfony to get more type info about command inputs
- Extracted common init/require commands functionality into PackageDiscoveryTrait
- Extracted some helper methods into BaseCommand for better types

### DIFF
--- a/src/Composer/Command/BaseCommand.php
+++ b/src/Composer/Command/BaseCommand.php
@@ -16,11 +16,15 @@ use Composer\Composer;
 use Composer\Config;
 use Composer\Console\Application;
 use Composer\Factory;
+use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilterFactory;
+use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilterInterface;
 use Composer\IO\IOInterface;
 use Composer\IO\NullIO;
+use Composer\Pcre\Preg;
 use Composer\Plugin\PreCommandRunEvent;
 use Composer\Package\Version\VersionParser;
 use Composer\Plugin\PluginEvents;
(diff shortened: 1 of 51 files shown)


src/Composer/Command/BaseCommand.php
```diff
@@ -50,20 +52,32 @@ abstract class BaseCommand extends Command
...
+
+    /**
+     * @param  bool              $required       Should be set to false, or use `requireComposer` instead
+     * @param  bool|null         $disablePlugins If null, reads --no-plugins as default
+     * @param  bool|null         $disableScripts If null, reads --no-scripts as default
      * @throws \RuntimeException
      * @return Composer|null
+     * @deprecated since Composer 2.3.0 use requireComposer or tryComposer depending on whether you have $required set to true or false
      */
     public function getComposer($required = true, $disablePlugins = null, $disableScripts = null)
```
naderman, line 75: just to make it clearer this is deprecated now could replace implementation with try/require Composer calls based on require flag.

src/Composer/Command/RequireCommand.php
```diff
@@ -38,8 +39,16 @@
  * @author Jérémy Romey <jeremy@free-agent.fr>
  * @author Jordi Boggiano <j.boggiano@seld.be>
  */
-class RequireCommand extends InitCommand
+class RequireCommand extends BaseCommand
 {
+    use PackageDiscoveryTrait;
+
+    // Properties for PackageDiscoveryTrait
```
stof, line 46: why not adding them in the trait directly ?

Seldaek: Because I did this late at night and was clearly not thinking straight 😄 Will fix

PR #10485, opened 2022-01-26
Add abandoned flag to show/outdated commands package lists

boesing:
Hey there,

we are using the `composer outdated --format=json` output to create scheduled JSON files we can use to notify us when projects needs some maintenance. I'd love to have some `abandoned` flag in there so we do not have to parse the `warning` string. 

I'm open for additional changes but this flag would definitely help us and maybe others who are consuming this JSON.
Feedback welcome.

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -461,6 +461,7 @@ EOT
                             $packageViewData['path'] = strtok(realpath($composer->getInstallationManager()->getInstallPath($package)), "\r\n");
                         }
 
+                        $packageIsAbandoned = false;
                         if ($latestPackage instanceof CompletePackageInterface && $latestPackage->isAbandoned()) {
                             $replacement = is_string($latestPackage->getReplacementPackage())
                                 ? 'Use ' . $latestPackage->getReplacementPackage() . ' instead'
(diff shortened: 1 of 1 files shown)


src/Composer/Command/ShowCommand.php
```diff
@@ -471,7 +472,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
                                 $replacement
                             );
                             $packageViewData['warning'] = $packageWarning;
+                            $packageIsAbandoned = true;
```
Seldaek, line 475: Usually we export this as `false|true|"replacement/package"` so truthy values are abandoned, and false or unset is non-abandoned.

It just means it should here use this I guess:
```suggestion
                            $packageIsAbandoned = $latestPackage->getReplacementPackage() ?? true;
```

boesing: So in case we want to stay compatible with how it is written to the `composer.json`, right?
Is this a strict requirement to reflect the `composer.json` information or can we add a dedicated column which provides a replacement suggestion instead?

Something like this is way more easy to parse IMHO:

```json
{
    "abandoned": bool,
    "replacementPackage": null|non-empty-string
}
```

But I will implement your suggestion if we want to keep in-sync with the `composer.json` `abandoned` flag ofc.

Seldaek: Yeah it's mostly about consistency as it is represented like that in composer.json and in various packagist.org json APIs

boesing: Got it, thanks for explanation. For consistency, I've changed the code appropriately.

Seldaek:
Thanks

PR #10486, opened 2022-01-26
update to composer/pcre 1.0.1

clxmstaab:


clxmstaab:
any hint on how to handle the remaining CI error?

src/Composer/Package/BasePackage.php
```diff
@@ -255,21 +255,25 @@ public function __clone()
...
+     * @return non-empty-string
      */
     public static function packageNameToRegexp($allowPattern, $wrap = '{^%s$}i')
     {
         $cleanedAllowPattern = str_replace('\\*', '.*', preg_quote($allowPattern));
 
-        return sprintf($wrap, $cleanedAllowPattern);
+        $regex = sprintf($wrap, $cleanedAllowPattern);
+        if ($regex === '') {
+            throw new \RuntimeException('Regex is empty');
```
Seldaek, line 266: Can we not avoid this by also requiring a non-empty-string in $wrap?

src/Composer/Repository/FilterRepository.php
```diff
@@ -205,6 +205,10 @@ private function isAllowed($name)
             return Preg::isMatch($this->only, $name);
         }
 
+        if ($this->exclude === null) {
+            return true;
+        }
```
Seldaek, line 210: Why is this needed though? The checks above should ensure that at this point $this->exclude cannot be falsey (and thus not === null)?

Seldaek (commented):
A few notes, the remaining issue is due to array_merge support being lacking in PHPStan and line 161 in CurlDownloader throwing it off.. So it can be shoved in the baseline for now.

Seldaek:
Also if you can rebase on 2.2 that'd be nice.

src/Composer/Package/BasePackage.php
```diff
@@ -269,7 +269,7 @@ public static function packageNameToRegexp($allowPattern, $wrap = '{^%s$}i')
      *
      * @param string[] $packageNames
      * @param string $wrap
-     * @return string
+     * @return non-empty-string
      */
     public static function packageNamesToRegexp(array $packageNames, $wrap = '{^(?:%s)$}iD')
```
staabm, line 274: I am wondering why we have this arg? its only used within the test-suite... can it be dropped?

src/Composer/Repository/FilterRepository.php
staabm: opened a [phpstan issue](https://github.com/phpstan/phpstan/issues/6486)

clxmstaab:
rebased on 2.2 in https://github.com/composer/composer/pull/10496

Issue #10482, opened 2022-01-25
Do not always load remote package info for path repository packages


sveneld:
In this issue https://github.com/composer/composer/issues/9751 updating of path repository was added, but
when many developers works with same package which is configured as 'path' they get a conflict during merging their branches due to the change of reference for path package in composer.lock. Is it possible to add an config option for path packages for removing such behavior?

Seldaek:
IMO the downsides are worse than the conflicts you may get on the reference. But one question to me is whether we could also always fix the reference to a single value when it comes from a path repo, that would probably remove the conflicts entirely. I need to think some more about this but I think it works.

Seldaek:
On second thought, the problem really is that if you don't change the reference every time you may have an out of date path package if your path repo is copying and not symlinking.

So I am not sure what we can do at least if copying is allowed. Perhaps we could suppress the git reference if symlink:true is set in the repository options.

naderman:
Agreed: default/copying requires the reference, for anyone who wants to suppress the conflicts, we can force symlink:true and then leave the reference empty

Seldaek:
OK abusing symlink:true is a bit of a mess after all IMO so I'd rather add a new option `reference` which could have a value of `auto` (default), `none` (always null, probably what you want here) and `config` (which is the default when no .git info is present, i.e. it hashes the composer.json and repo config, and still provides a useful reference showing when something changed in that package, which I find more useful than no ref at all).

Seldaek:
@sveneld I saw yes, could you update it to match what I described above perhaps with the different modes?

sveneld:
Ok i will do it tomorrow

sveneld:
@Seldaek done

PR #10488, opened 2022-01-27
Add config option for update reference in path repository

sveneld:
Example of composer.json
```
{
  "name": "example/example",
  "require": {
    "pathExample/example": "^1.0"
  },
  "repositories": [
    {
      "type": "path",
      "url": "./pathExample"
    }
  ]  
}
```
Source code in folder pathExample is under git version control and offently updated by developers, and after each update "reference" in composer.lock for this package is update, but it caused a conflict between different branches. So it will be great to have opportunity to configure reference updating like this

```
{
  "name": "example/example",
  "require": {
    "pathExample/example": "^1.0"
  },
  "repositories": [
    {
      "type": "path",
      "url": "./pathExample",
      "reference": "none"
    }
  ]  
}
```

Related to https://github.com/composer/composer/issues/10482

### DIFF
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -56,6 +56,13 @@ use Composer\Util\Git as GitUtil;
  *             "symlink": false
  *         }
  *     },
+ *     {
+ *         "type": "path",
+ *         "url": "../../relative/path/to/package/",
+ *         "options": {
+ *             "reference": "none"
+ *         }
+ *     },
  * ]
  * @endcode
  *
@@ -81,7 +88,7 @@ class PathRepository extends ArrayRepository implements ConfigurableRepositoryIn
 
     /**
      * @var mixed[]
-     * @phpstan-var array{url: string, options?: array{symlink?: bool, relative?: bool, versions?: array<string, string>}}
+     * @phpstan-var array{url: string, options?: array{symlink?: bool, reference?: string, relative?: bool, versions?: array<string, string>}}
      */
     private $repoConfig;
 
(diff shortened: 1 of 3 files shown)


src/Composer/Repository/PathRepository.php
```diff
@@ -171,8 +181,12 @@ protected function initialize()
             $package['dist'] = array(
                 'type' => 'path',
                 'url' => $url,
-                'reference' => sha1($json . serialize($this->options)),
             );
+            if ('none' === $this->options['reference']) {
+                $package['dist']['reference'] = '';
```
Seldaek, line 186: I believe null would be better here
```suggestion
                $package['dist']['reference'] = null;
```

src/Composer/Repository/PathRepository.php
```diff
@@ -118,6 +125,9 @@ public function __construct(array $repoConfig, IOInterface $io, Config $config)
             $filesystem = new Filesystem();
             $this->options['relative'] = !$filesystem->isAbsolutePath($this->url);
         }
+        if (!isset($this->options['reference'])) {
+            $this->options['reference'] = 'auto';
```
Seldaek, line 129: I think this should rather be done below in initialize something like `$reference = $this->options['reference'] ?? 'auto';` and then use that variable below. That would avoid changing the reference for everyone based on the composer version (which is why you had to edit the tests).

src/Composer/Repository/PathRepository.php
sveneld: ++

src/Composer/Repository/PathRepository.php
sveneld: agree

Seldaek:
Thanks. The other bonus thing would be to add this new option to the docs of the path repo in `doc/05-repositories.md` :)

sveneld:
Ok. tomorrow :)

Seldaek:
No rush :)

sveneld:
Done

src/Composer/Repository/PathRepository.php
```diff
@@ -91,14 +98,14 @@ class PathRepository extends ArrayRepository implements ConfigurableRepositoryIn
     private $process;
 
     /**
-     * @var array{symlink?: bool, relative?: bool, versions?: array<string, string>}
+     * @var array{symlink?: bool, reference?: string, relative: bool, versions?: array<string, string>}
```
Seldaek, line 101: This seems to be a regression causing the PHPStan warning below 

```suggestion
     * @var array{symlink?: bool, reference?: string, relative?: bool, versions?: array<string, string>}
```

Seldaek:
Thanks!

PR #10496, opened 2022-01-30
update composer/pcre to 1.0.1

clxmstaab:


clxmstaab:
I don't think the remaining error is related

Seldaek:
Thanks

PR #10500, opened 2022-01-31
Fix usage of spaces for package names in `composer search`/`composer require` commands

realFlowControl:
Hello there :vulcan_salute: 

I found out while using `composer search` that I can not search for two words with a space in between:
```bash
$ composer search ramsey commits

curl error 3 while downloading https://packagist.org/search.json?q=ramsey commits&type=: URL using bad/illegal format or missing URL

search [-N|--only-name] [-t|--type TYPE] [-f|--format FORMAT] [--] <tokens> (<tokens>)...
```
Same is true for using `composer require`:
```bash
$ composer require
Search for a package: ramsey commits

curl error 3 while downloading https://packagist.org/search.json?q=ramsey commits&type=: URL using bad/illegal format or missing URL
```
I do not know for the `composre require` command, but for the `composer search` command such a usage is documented via the `composer help search` command, so I assumed it is the expected behavior for this to work.

With this PR applied those two commands behave as following:
```bash
$ ./bin/composer require
Search for a package: ramsey commits

Found 1 packages matching ramsey commits

   [0] ramsey/conventional-commits

Enter package # to add, or the complete package name if it is not listed:
[...]
```
and
```bash
./bin/composer search ramsey commits
ramsey/conventional-commits A PHP library for creating and validating commit messages according to the Conventional Commits specification. In...
```

What do you think?

Kind regards
Florian

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -540,7 +540,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
         $this->loadRootServerFile(600);
 
         if ($this->searchUrl && $mode === self::SEARCH_FULLTEXT) {
-            $url = str_replace(array('%query%', '%type%'), array($query, $type), $this->searchUrl);
+            $url = str_replace(array('%query%', '%type%'), array(urlencode($query), $type), $this->searchUrl);
 
             $search = $this->httpDownloader->get($url, $this->options)->decodeJson();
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Weird because I cannot reproduce this:

```
$ composer search ramsey commits -vvv
[...]
Downloading https://packagist.org/search.json?q=ramsey commits&type=
[200] https://packagist.org/search.json?q=ramsey commits&type=
ramsey/conventional-commits A PHP library for creating and validating commit messages according to the Conventional Commits specification. Includes a CaptainHook action!
```

That said, it makes absolute sense to urlencode here regardless of space handling, so thanks :)

PR #10516, opened 2022-02-05
Introduce phpstan-type about autoload array type to be DRY

sasezaki:
To be clear, and to be DRY.
I propose refactor autoload ~~mapping array~~ array type with type alias.

NOTES: `exclude-from-classmap` key is defined in composer-schema, but current array type does not have it. I don't know what to do it. 
https://github.com/composer/composer/blob/main/res/composer-schema.json#L670

In additionally, first of all, there are some duplicates in BetterReflection too. Therefore, I do this PR for clarification about autoload mapping array.
https://github.com/Roave/BetterReflection/blob/5.1.x/src/SourceLocator/Type/Composer/Factory/MakeLocatorForInstalledJson.php

### DIFF
--- a/src/Composer/Package/Package.php
+++ b/src/Composer/Package/Package.php
@@ -21,7 +21,8 @@ use Composer\Util\ComposerMirror;
  *
  * @author Nils Adermann <naderman@naderman.de>
  *
- * @phpstan-import-type ComposerAudoload from PackageInterface
+ * @phpstan-import-type AutoloadRules from PackageInterface
+ * @phpstan-import-type DevAutoloadRules from PackageInterface
  */
 class Package extends BasePackage
 {
@@ -81,9 +82,15 @@ class Package extends BasePackage
     protected $devRequires = array();
     /** @var array<string, string> */
     protected $suggests = array();
-    /** @var ComposerAudoload */
+    /**
+     * @var array
+     * @phpstan-var AutoloadRules
+     */
     protected $autoload = array();
-    /** @var ComposerAudoload */
+    /**
+     * @var array
+     * @phpstan-var DevAutoloadRules
+     */
     protected $devAutoload = array();
     /** @var string[] */
     protected $includePaths = array();
@@ -600,7 +607,7 @@ class Package extends BasePackage
      *
      * @return void
      *
-     * @phpstan-param ComposerAudoload $autoload
+     * @phpstan-param AutoloadRules $autoload
      */
     public function setAutoload(array $autoload)
     {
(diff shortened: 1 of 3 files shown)


src/Composer/Package/Package.php
```diff
@@ -20,6 +20,8 @@
  * Core package definitions that are needed to resolve dependencies and install packages
  *
  * @author Nils Adermann <naderman@naderman.de>
+ *
+ * @phpstan-import-type AutoloadMapping from \Composer\Package\PackageInterface
```
Seldaek, line 24: I think this should work too?

```suggestion
 * @phpstan-import-type AutoloadMapping from PackageInterface
```

Seldaek:
You are right exclude-from-classmap is missing, it probably should be added as `list<string>`.

src/Composer/Package/Package.php
sasezaki: Yes, you are right. Thanks for suggestion!
I would commit this "Suggested change"
 
NOTES: Using FQCN is just by my current habit. Because  I often met other coding standard check tool reports "Unused Use" false-positive. :cry:

sasezaki:
@Seldaek 

> You are right exclude-from-classmap is missing, it probably should be added as list<string>.

I'm glad to hear that.
On the other hand, I have two questions.

1. `exclude-from-classmap` is not defined in `autoload-dev` of composer-schema.json, can I add `exclude-from-classmap` into `AutoloadMapping` ?

2. Is `list<non-empty-string>` is more preferable ?

sasezaki:
> Is `list<non-empty-string>` is more preferable ?

Please ignore this. It would require other code changes.

Seldaek:
> 1. `exclude-from-classmap` is not defined in `autoload-dev` of composer-schema.json, can I add `exclude-from-classmap` into `AutoloadMapping` ?

I think it's fine to add as all keys are anyway optional.. Sure if you set the key on setAutoloadDev for example it will be ignored, but that doesn't sound too dramatic to me as almost nobody interacts directly with this at the PHP API level.

Or perhaps rename the current type alias to AutoloadDevMapping and add a new AutoloadMapping as `AutoloadDevMapping&array{exclude-from-classmap?: list<string>}`. That way you get correct types on everything.

sasezaki:
> I think it's fine to add as all keys are anyway optional.. Sure if you set the key on setAutoloadDev for example it will be ignored, but that doesn't sound too dramatic to me as almost nobody interacts directly with this at the PHP API level.

OK, thanks for confirm.  I just did add `exclude-from-classmap` at 409cf89

sasezaki:
As pointed at https://github.com/Roave/BetterReflection/pull/1004, 
`AutoloadMapping` is not good naming. So I would to change `ComposerAutoload`.

Seldaek:
I did rename things again (not a fan of including a Composer prefix, we are in the Composer repo it's quite clear it's composer stuff:) and decided to split it after all to have more correct types. 

Thanks!

PR #10517, opened 2022-02-05
Remove duplicated keys

franmomu:
Spotted looking at the php-cs-fixer configuration, I've targeted `2.2`, but not sure if it should be `main`.


phansys (approved):

Seldaek:
Thanks

PR #10534, opened 2022-02-13
Add `da` alias to dump-autoload command

squiaios:
Add new alias to dump-autoload `da`


Seldaek:
You can also use `composer du` or `composer dump` or whatever, as long as its non-ambiguous it'll execute the command it finds that matches, and if it is ambiguous it'll tell you so it's risk-free to try.

As such I don't think we really need to add short aliases for all the things.

PR #10536, opened 2022-02-15
Fix licenses summary with missing license info

gzumba:
Currently the LicensesCommand fails, when getting the summary (-f  summary) and one of the packages  is missing the  license information.
This fixes the issue by simply using 'none' as the license, as is used in the text output already.


gzumba:
I based this on the wrong branch, so I made a new PR here: https://github.com/composer/composer/pull/10537

PR #10537, opened 2022-02-15
Handle missing license when showing license summary

gzumba:
Currently the LicensesCommand fails, when getting the summary (-f summary) and one of the packages is missing the license information.
This fixes the issue by simply using 'none' as the license, as is used in the text output already.


Seldaek:
Thanks

PR #10542, opened 2022-02-18
Renamed the only file with a quote character to remove it

seanm:
The quote character can be problematic and as this is the only one, just removing it makes the codebase more compatible with other systems that forbid the character as a matter of policy.


Seldaek:
Thanks

PR #10545, opened 2022-02-18
Improve the pool optimizer logging

Toflar:
I would like to give the pool optimizer another shot 😇 
`--profile` is nice but it doesn't help me at the moment as I don't know when the pool optimizer starts working and I think for quick comparisons of improvements to come (I'm pretty sure I'll find something), a separate message like we also have it in the solver would be great.


Seldaek:
Thanks

PR #10572, opened 2022-02-25
Rename doc/faqs/how-to-install-composer-programmatically.md to instal…

muha0039:
…l-composer-programmatically.m


Seldaek:
Why? This breaks links and offers no value otherwise that I can see?

Issue #10558, opened 2022-02-21
prefer-lowest fails installing vendors


emodric:
I have a composer.json file which installs a couple of packages (amongst others) that require `twig/twig` in different versions: either `^3.0` or `^2.14 || ^3.3`. Now while regular `composer update` works fine and latest `twig/twig` is installed, using `--prefer-lowest` just errors out and I can't seem to parse the output to figure out the problem.

In my understanding, Composer should install the lowest possible version that matches `^3.0` or `^2.14 || ^3.3`, which would be `3.3.0`, but it looks like it tries to install `2.14.0` and fails.

My `composer.json`:

```json
{
    "minimum-stability": "dev",
    "prefer-stable": true,
    "require": {
        "netgen/layouts-ibexa": "~1.4.0"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.2.6
PHP version: 8.1.2
PHP binary path: /usr/bin/php8.1
OpenSSL version: OpenSSL 1.1.1l  24 Aug 2021
cURL version: 7.74.0 libz 1.2.11 ssl OpenSSL/1.1.1l
zip: extension present, unzip present, 7-Zip present (7z)
```

When I run this command:

```
composer update --prefer-lowest
```

I get the following output:

```
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - twig/twig v3.0.0 requires php ^7.2.9 -> your php version (8.1.2) does not satisfy that requirement.
    - twig/twig[v3.0.1, ..., v3.0.3] require php ^7.2.5 -> your php version (8.1.2) does not satisfy that requirement.
[...]

PR #10579, opened 2022-02-28
Fix PoolOptimizer should consider disjunctive MultiConstraints

Toflar:
Fixes https://github.com/composer/composer/issues/10558

The problem is that we're currently only generating one "dependency constraint group" for disjunctive `MultiConstraint` instances. This means that for `^2.14 || ^3.3` we only keep one package that matches this constraint. However, this can go bad as e.g. `3.3.0` goes missing in favor of `2.14.0` for `--prefer-lowest` which might cause an unresolvable set of dependencies whereas `3.3.0` would have actually led to a solution.

If a `MultiConstraint` is disjunctive, we need to create individual groups for every sub constraint.

### DIFF
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -490,22 +490,15 @@ class PoolOptimizer
      */
     private function expandDisjunctiveMultiConstraints(ConstraintInterface $constraint)
     {
-        $expanded = array();
         $constraint = Intervals::compactConstraint($constraint);
 
         if ($constraint instanceof MultiConstraint && $constraint->isDisjunctive()) {
-            foreach ($constraint->getConstraints() as $sub) {
-                // No need to call ourselves recursively here because Intervals::compactConstraint() ensures that there
-                // are no nested disjunctive MultiConstraint instances possible
-                $expanded[] = $sub;
-            }
-
-            return $expanded;
+            // No need to call ourselves recursively here because Intervals::compactConstraint() ensures that there
+            // are no nested disjunctive MultiConstraint instances possible
+            return $constraint->getConstraints();
         }
 
         // Regular constraints and conjunctive MultiConstraints
-        $expanded[] = $constraint;
-
-        return $expanded;
+        return array($constraint);
     }
 }
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -495,11 +495,8 @@ class PoolOptimizer
 
         if ($constraint instanceof MultiConstraint && $constraint->isDisjunctive()) {
             foreach ($constraint->getConstraints() as $sub) {
-                // TODO: Do we need to call ourselves recursively? Is it even possible?
-                // Tried to write a testcase in PoolOptimizerTest::testNestedDisjunctiveMultiConstraints() but
-                // couldn't find an example where Intervals::compactConstraint() leaves a nested disjunctive MultiConstraint
-                // untouched.
-               // $expanded = array_merge($expanded, $this->expandDisjunctiveMultiConstraints($sub));
+                // No need to call ourselves recursively here because Intervals::compactConstraint() ensures that there
+                // are no nested disjunctive MultiConstraint instances possible
                 $expanded[] = $sub;
             }
 
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -110,21 +110,18 @@ class PoolOptimizer
 
         // Extract requested package requirements
         foreach ($request->getRequires() as $require => $constraint) {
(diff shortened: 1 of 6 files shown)


Toflar:
@emodric and @Seldaek: I think this is ready to review/test. It's kind of funny nobody noticed this yet :) 
Failing test looks unrelated.

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -453,4 +450,60 @@ private function optimizeImpossiblePackagesAway(Request $request, Pool $pool)
...
+     * @return ConstraintInterface[]
+     */
+    private function expandDisjunctiveMultiConstraints(ConstraintInterface $constraint)
+    {
+        $expanded = array();
+        $constraint = Intervals::compactConstraint($constraint);
+
+        if ($constraint instanceof MultiConstraint && $constraint->isDisjunctive()) {
+            foreach ($constraint->getConstraints() as $sub) {
+                $expanded = array_merge($expanded, $this->expandDisjunctiveMultiConstraints($sub));
```
Seldaek, line 498: Is the recursion here needed vs simply `$expanded[] = $sub;`? I don't think we ever create nested disjunctive constraints but maybe I'm overlooking something.

Seldaek:
Thanks for the fix, it makes sense I think. I suppose the perf regression isn't too bad as this kind of constraints isn't so widespread?

src/Composer/DependencyResolver/PoolOptimizer.php
Toflar: That's one of my questions ;) I don't know if it's even possible to define something like `(^2.0 || ^3.0) || (^4.1 || ^2.1)`?

Toflar:
I could not see any regression, no. I actually tested that :) According to blackfire there are other parts that require far more time which is something I want to investigate but that was already the case before this PR.

emodric:
I cannot comment in implementation, but I tested the fix on example `composer.json` from the issue and on the real life project `composer.json` and install works fine :+1:

src/Composer/DependencyResolver/PoolOptimizer.php
naderman: It technically is possible I think, although very uncommon. But we do currently allow multiconstraints to be arbitrarily nested.

Toflar: I'll try to work on another testcase then. I want to have this covered.

Toflar: Brackets are not allowed in version constraints so I don't know how I could create this. I also tried in an actual project and there was no case 🤔

Toflar: It probably only works when instantiating a `MultiConstraint` manually but I don't think it's possible using a written constraint and the `VersionParser`?

Seldaek: Yeah that's what I meant, I don't think Composer lets you create these.

Toflar: They could be created by a plugin maybe. So I guess I need to write a test where we do that manually?

Toflar: We could also prevent it here: https://github.com/composer/semver/blob/main/src/Constraint/MultiConstraint.php#L48 I think the impact would be very low because it can only be created by code so it won't really be used anywhere.

Seldaek: Yeah I mean we can leave this as is for completeness. Add a test case if you like but if not I will survive, the important case is covered 👍🏻

Toflar: I've tried hard finding an example but I was not able to think of a nested **disjunctive** `MultiConstraint` that is not optimized by `Intervals::compactConstraint()`. I've committed my notes and the test skeleton in https://github.com/composer/composer/pull/10579/commits/160e1f42c9b9de61228a5e5bedec11695cb0d091 nevertheless. Maybe one of you guys can think of an example that would make the recursive call necessary?

naderman: Ah indeed, compact constraint will never have nested disjunction in the output. Disjunction is only used on top level to basically list all the intervals.

Toflar: With that knowledge in mind, we can also drop the hint (and the time-consuming test if you want 😢) 😉 Just let me know what to do :)

naderman: Sorry had missed that compact was called :( I'd remove it then and maybe just leave a comment that because of the compact that can't exist there?

Toflar: No prob, it helped us find out that it's not a use case so that's good :) This PR is now finished then, I guess.

Toflar:
Failing test again unrelated.

Seldaek:
Thanks!

PR #10585, opened 2022-03-02
Improve performance of DefaultPolicy

Toflar:
The `$policy->selectPreferredPackages()` is another bottleneck in the `PoolOptimizer` as we call it for every dependency group which can easily happen for 10 000+ times on medium size projects. Most packages match to a lot of dependency groups so we actually compare the same packages over and over again to sort them within one `selectPreferredPackages()` call. The respective 2 `usort()` calls were actually responsible for up to 25% of the whole `PoolOptimizer::optimize()` process.

With this PR we're now caching the result so the sorting can be way faster (another 0.5 - 1 seconds down on the `PoolOptimizer`).

Because we're using the `PolicyInterface` and `Pool $pool` is an argument, we have to make sure the cache is specific to the `$pool` instance. Thanks to the new PHP 7.2 requirement, we can use the faster `spl_object_id()` for that 😎

### DIFF
--- a/src/Composer/DependencyResolver/DefaultPolicy.php
+++ b/src/Composer/DependencyResolver/DefaultPolicy.php
@@ -27,6 +27,8 @@ class DefaultPolicy implements PolicyInterface
     private $preferStable;
     /** @var bool */
     private $preferLowest;
+    /** @var array<int, array<string, int>> */
+    private $sortingCachePerPool;
 
     /**
      * @param bool $preferStable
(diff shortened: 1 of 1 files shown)


src/Composer/DependencyResolver/DefaultPolicy.php
```diff
@@ -64,10 +66,17 @@ public function versionCompare(PackageInterface $a, PackageInterface $b, string
     public function selectPreferredPackages(Pool $pool, array $literals, string $requiredPackage = null): array
     {
         $packages = $this->groupLiteralsByName($pool, $literals);
+        $poolId = spl_object_id($pool);
 
         foreach ($packages as &$nameLiterals) {
-            usort($nameLiterals, function ($a, $b) use ($pool, $requiredPackage): int {
-                return $this->compareByPriority($pool, $pool->literalToPackage($a), $pool->literalToPackage($b), $requiredPackage, true);
+            usort($nameLiterals, function ($a, $b) use ($pool, $requiredPackage, $poolId): int {
```
Seldaek, line 72: Have you tried caching the whole result like `sort($nameLiterals); $cacheKey = implode(',', $nameLiterals);` then memoize the whole usort() call instead of every individual comparison?

I am not entirely sure if the combinations of literals here will vary so much that it makes it useless or not.. But if not I guess it'd be even more efficient. Gotta try this with various projects though

Toflar: I think a complete match is pretty unlikely to happen. We compare within the dependency groups. So we have a group for e.g. `^4.2` and `^4.3` etc. Having the exact same packages inside those groups sounds very unrealistic to me. But I'll run some tests to see how many hits we'd have.

Toflar: Turns out I was wrong :) There are actually also thousands of hits 🥳 Both caches have very high hit rates now. I had to fix the cache key of the sorting cache too. I forgot to consider `$requiredPackage`.

Seldaek:
Thanks

PR #10591, opened 2022-03-05
GH Actions: version update for various predefined actions

jrfnl:
A number of predefined actions have had major release, which warrant an update to the workflow(s).

These updates don't actually contain any changed functionality, they are mostly just a change of the Node version used by the action itself (from Node 14 to Node 16).

Refs:
* https://github.com/actions/checkout/releases

P.S.: looks like there are some more actions in use which have had new major releases. As those actually _do_ contain changed functionality, I've excluded the others from this PR.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -62,7 +62,7 @@ jobs:
 
     steps:
       - name: "Checkout"
-        uses: "actions/checkout@v2"
+        uses: "actions/checkout@v3"
 
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"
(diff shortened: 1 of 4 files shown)


localheinz (approved):
👍

GrahamCampbell:
Node 12 to 16, actually. Not 14. ;)

GrahamCampbell (approved):

jrfnl:
> Node 12 to 16, actually. Not 14. ;)

Correct. I realized the typo after I'd already pulled the change to 40+ repos. Wasn't going to go back and edit them all 😂

Seldaek:
Thanks

Issue #10578, opened 2022-02-26
Add an additional `/` for symlinks to make them Windows friendly


bilogic:
Hi,

I'm running Ubuntu as a VirtualBox guest on my Windows 10.

In Windows, there are `SYMLINK`, referring to a file, and `SYMLINKD`, referring to a folder.
The difference between the 2 when accessed from a Linux guest is 

1. SYMLINK - `/path/to/folder/`
2. SYMLINKD - `/path/to/folder//`, note the addtional `/` at the end

Today, `composer` running in Linux creates symlinks using method 1, this causes Windows to interpret it as a file and thus, doesn't allow the user to click into it the folder when in Windows. 

My expected output is that `composer` creates the symlinks with an additional `/`, this then makes both Linux and Windows parse the symlinks correctly.

I'm happy to submit a PR if this makes sense. Thank you.

Seldaek:
Please submit a PR because I don't quite see how that'd work.

bilogic:
Thanks, I'm onto it. 

I tried searching through the codes and found `relativeSymlink` here https://github.com/composer/composer/blob/575fbfb53fcc2388916d554271c99c8281fea782/src/Composer/Util/Filesystem.php#L737

However, no code seems to be calling it (found nothing for `>relativeSymlink` and `::relativeSymlink`).

Could you be so kind as to point me to the spot where composer processes the `repositories` section of `composer.json`? 
I'm specifically looking for the part where it performs symlink or mirror.

Thank you!

PR #10592, opened 2022-03-06
In a Windows host/Linux guest filesystem, adding the '/' creates a SY…

bilogic:
See #10578. Thank you.

### DIFF
--- a/src/Composer/Downloader/PathDownloader.php
+++ b/src/Composer/Downloader/PathDownloader.php
@@ -121,9 +121,9 @@ class PathDownloader extends FileDownloader implements VcsCapableDownloaderInter
                         $this->io->writeError(sprintf('Symlinking from %s', $url), false);
                     }
                     if ($transportOptions['relative']) {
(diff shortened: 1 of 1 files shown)


bilogic:
I just thought of a better commit message

`creates SYMLINKD instead of SYMLINK on Windows host/Linux guest filesystems`

bilogic:
I just thought of a better commit message

`creates SYMLINKD instead of SYMLINK on Windows host/Linux guest filesystems`

Seldaek:
Is this specific to VirtualBox? Because I think I tried exactly this patch before asking you to send a PR and I could not see any difference when running it on my WSL2 linux filesystem.

Seldaek:
Merging already, thanks

bilogic:
To answer your question, I have not tested on WSL2, but my setup is VirtualBox and it makes a difference on Windows host and did not have an impact on the Linux guest, so I guess yes, it is specific to VirtualBox.

Thank you.

PR #10611, opened 2022-03-11
Fix: Simplify

localheinz:
This pull request

- [x] simplifies conditions for step

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -76,23 +76,23 @@ jobs:
           tools: composer
 
       - name: "Handle lowest dependencies update"
-        if: "contains(matrix.dependencies, 'lowest')"
+        if: "matrix.dependencies == 'lowest'"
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -76,23 +76,23 @@ jobs:
           tools: composer
 
       - name: "Handle lowest dependencies update"
-        if: "contains(matrix.dependencies, 'lowest')"
```
localheinz, line 79: `matrix.dependencies` only assumes 

- `'ignore'`
- `'locked'`
- `'lowest'`
- `'highest'`

so it suffices to compare with these values, rather than checking if `matrix.dependencies` contains these values.

Seldaek: Yeah We used to have `highest-ignore` when php 8.0 or 8.1 was not supported yet in deps. So thanks but I think it's better to keep as is, as it makes it easier to change the matrix without having to change everything below.

Issue #10599, opened 2022-03-08
Plugin not removed from allow-plugins config on remove


chr-hertel:
When I remove a plugin which was listed in the `allow-plugins` section of `config`, it is not removed from that section.

Example `composer.json`:

```json
{
    "require": {
        "symfony/flex": "^2.1"
    },
    "config": {
        "allow-plugins": {
            "symfony/flex": true
        }
    }
}
```

After succesfully running the remove cmd: `composer rem symfony/flex`:

```json
{
    "config": {
        "allow-plugins": {
            "symfony/flex": true
        }
    }
}
```

Is my expectation correct that it should be remove in the config as well?

Seldaek:
I guess we could automatically do that on plugin uninstall, I just wonder if that will end up annoying for some use cases.

PR #10615, opened 2022-03-13
Remove allow-plugin config on remove of plugin

chr-hertel:
Fixing #10599. No idea if this approach is too lazy


Seldaek:
Yeah this probably makes more sense than doing it whenever a plugin is uninstalled IMO. If you run a remove command it's a clear intent to get rid of the plugin.

Seldaek:
Thanks

PR #10616, opened 2022-03-14
Allow global ignore-platform-req

yakatz:
The particular code I am working with now interacts with several different database systems (mysql, postgres and oracle). My development workstation can't access the oracle database, but I need to always remember to always specify `--ignore-platform-req=ext-oci8`. This patch allows setting that value in the environment instead.


src/Composer/Command/BaseCommand.php
```diff
@@ -216,6 +218,27 @@ protected function initialize(InputInterface $input, OutputInterface $output)
...
+                $io->writeError('<warning>COMPOSER_IGNORE_PLATFORM_REQS is set. You may experience unexpected errors.</warning>');
+            }
+        }
+
+        if (true == $input->hasOption('ignore-platform-req') && (!$input->hasOption('ignore-platform-reqs') || !$input->getOption('ignore-platform-reqs'))) {
+            $raw_ignored_platform_req = Platform::getEnv('COMPOSER_IGNORE_PLATFORM_REQ');
+            if (
+                ("" != $input->getOption('ignore-platform-req'))
+                && "" != $raw_ignored_platform_req
+                && false !== $raw_ignored_platform_req
```
yakatz, line 234: This odd-looking check is required to make `phpstan` happy

yakatz:
The failed tests seem to be stopping on the `StandWithUkraine` message:
image

I don't see why this is an issue with only PHP 7.3, but I don't think it should invalidate this PR.

Seldaek:
Thanks!

PR #10637, opened 2022-03-18
Switch to LazyCommand implementations to improve bootstrap speed

Seldaek:
Makes use of https://symfony.com/blog/new-in-symfony-5-3-lazy-command-description - TBH kind of micro-optimization level I guess as running the configure of most of those commands isn't a whole lot of code that runs, but if it saves a few ms it's a win I guess especially in the context of #10320

If someone with a profiler setup at hand and some free time wants to check the result, please do :)


Seldaek:
Eh well, closing after talking to @nicolas-grekas, seems rather pointless and long term anyway soon deprecated in favor of AsCommand attributes (which we cannot use)

PR #10645, opened 2022-03-21
Use short array syntax

IonBazan:
This PR enables short array and list syntax, replacing all `array(...)` with `[...]`.


Seldaek:
Thanks but as you can see in the php-cs config there was already a TODO with a note, and IMO this is not ready to be done yet. It'll cause conflicts with 2.2 merges and I have better things to do with my time :)

Also please don't send CS-tweaking PRs, I rather do these myself so I don't have to carefully review 5000 lines diffs.

PR #10646, opened 2022-03-21
SvnDriver: buildIdentifier  must be of type int, string given

glaubinix:
```
In SvnDriver.php line 403:
                                                                               
  [TypeError]                                                                  
  Composer\Repository\Vcs\SvnDriver::buildIdentifier(): Argument #2 ($revisio  
  n) must be of type int, string given, called in src/Composer/Reposi  
  tory/Vcs/SvnDriver.php on line 273  
```

### DIFF
--- a/src/Composer/Repository/Vcs/SvnDriver.php
+++ b/src/Composer/Repository/Vcs/SvnDriver.php
@@ -396,11 +396,11 @@ class SvnDriver extends VcsDriver
      * Build the identifier respecting "package-path" config option
      *
      * @param string $baseDir  The path to trunk/branch/tag
-     * @param int    $revision The revision mark to add to identifier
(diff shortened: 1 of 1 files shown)


stof:
There is nothing in the regex ensuring that only integers are matched. So to me, a better fix would be to change the type in buildIdentifier instead (which does not actually care about getting a number anyway, as it only uses it to concatenate it with a string)

glaubinix:
Right, but at the same time SVN revisions are always integers, so it makes sense for that method to only accept integers. Accepting string values would then allow the method to generate invalid build identifiers.

Seldaek:
Yeah I can see both ways are somewhat valid.. Given svn is somewhat on the EOL track, let's merge as is I guess :) Thanks

PR #10647, opened 2022-03-21
ArrayLoader: assert that source/dist reference are string values

glaubinix:
Source and dist reference from third party repositories can contain int values. Example in wpackagist: https://wpackagist.org/p/wpackagist-theme/blankslate%2491eb3d60b45cca3c230a8bbb96e88b06ce3d7ea22a4af26caeaeaaab2f16869d.json


Seldaek:
Thanks

PR #10656, opened 2022-03-23
RemoteFilesystem: fix result length check

glaubinix:
Fixes: `Uncaught Error: Composer\Util\Platform::strlen(): Argument #1 ($str) must be of type string, bool given, called in src/Composer/Util/RemoteFilesystem.php on line 529`


Seldaek:
Thanks

PR #10657, opened 2022-03-24
Bitbucket: handle missing authentication response from API returning 404 status code

glaubinix:
Bitbucket now seems to return a 404 status code with the below body if you access a repository you don't have access to
```
{"type": "error", "error": {"message": "The requested repository either does not exist or you do not have access. If you believe this repository exists and you have access, make sure you're authenticated."}}
```


glaubinix:
Failing test case seems unrelated.

Seldaek:
Thanks

PR #10658, opened 2022-03-24
ValidatingArrayLoader: only validate source/dist properties if they are set

glaubinix:
Resolves:
```
Warning: Undefined array key "url"
Uncaught Error: Composer\Pcre\Preg::isMatch(): Argument #2 ($subject) must be of type string, null given in ValidatingArrayLoader.php on line 367
```


stof (approved):

Seldaek:
Thanks

PR #10660, opened 2022-03-24
ArrayLoader: fix integer index of branch alias

glaubinix:
Fixes `Console Exception: strtolower(): Argument #1 ($string) must be of type string, int given in ArrayLoader.php:440`

For example, happens for package "silverstripe/crontask" from packagist.org which has an `extra` key with value `{"branch-alias":{"1":"1.3.x-dev"}}` in version `1.x-dev`

### DIFF
--- a/tests/Composer/Test/Package/Loader/ArrayLoaderTest.php
+++ b/tests/Composer/Test/Package/Loader/ArrayLoaderTest.php
@@ -347,4 +347,23 @@ class ArrayLoaderTest extends TestCase
         $this->assertSame('2019', $package->getSourceReference());
         $this->assertSame('2019', $package->getDistReference());
     }
+
+    public function testBranchAliasIntegerIndex(): void
+    {
+        $config = array(
+            'name' => 'acme/package',
+            'version' => 'dev-1',
+            'extra' => [
+                'branch-alias' => [
+                    '1' => '1.3-dev',
+                ],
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/Package/Loader/ArrayLoaderTest.php
```diff
@@ -347,4 +347,25 @@ public function testNoneStringSourceDistReference(): void
...
+            ],
+            'dist' => [
+                'type' => 'zip',
+                'url' => 'https://example.org/',
+            ],
+        );
+
+        $package = $this->loader->load($config);
+        $this->assertSame('2019', $package->getSourceReference());
+        $this->assertSame('2019', $package->getDistReference());
```
Seldaek, line 369: The assertion is wrong here, probably should check that we get an alias package with the expected version or something?

Seldaek:
Good find, super nasty these numeric array indices..

tests/Composer/Test/Package/Loader/ArrayLoaderTest.php
glaubinix: Right, copied this and never updated it. I adjusted the test and I am only calling the `getBranchAlias` method. I am not sure, I understand what you mean with the alias package. The branch alias will never match because `1` does not match a dev version.

PR #10663, opened 2022-03-25
ArrayLoader: handle links where target is invalid numeric package name

glaubinix:
Resolves: `Uncaught Console Exception: strtolower(): Argument #1 ($string) must be of type string, int given in ArrayLoader.php:373`

Traces:
```
src/Composer/Package/Loader/ArrayLoader.php:373",
src/Composer/Package/Loader/ArrayLoader.php:68",
src/Composer/Repository/VcsRepository.php:306",
src/Composer/Repository/ArrayRepository.php:311",
```


Seldaek:
Thanks

PR #10665, opened 2022-03-28
SplFileInfo: getRealPath can return false in ArchivableFilesFinder

glaubinix:
Fixes: `TypeError(code: 0): strpos(): Argument #1 ($haystack) must be of type string, bool given at src/Composer/Package/Archiver/ArchivableFilesFinder.php:61`

Happens for instance if files in the bin folder are pointing to the vendor directory and vendors are not installed


Seldaek:
Thanks

Issue #10667, opened 2022-03-30
oAuth Authentication with Private BitBucket Fails


sjregan:
My `composer.json`:

```json
{
	"name": "namespace/project-name",
	"description": "Example",
	"minimum-stability": "dev",
	"config": {
		"autoloader-suffix": "ExampleSuffix",
		"optimize-autoloader": true
	},
	"autoload": {
		"classmap": [
          "project-name.php", "src/"
		],
		"exclude-from-classmap": [
			"src/assets/"
		],
		"psr-4": {
			"ExampleName\\Space\\": "src"
		}
	},
	"require-dev": {
		"automattic/jetpack-autoloader": "2.11.x-dev",
		"example-namespace/project-name": "dev-main"
	},
	"require": {
		"piggly/php-api-client": "dev-develop",
		"example-namespace/other-project-name": "dev-main"
	},
	"repositories": [
		{
			"type": "git",
			"url": "https://bitbucket.org/org-name/repo-name.git"
		},
		{
			"type": "git",
			"url": "https://bitbucket.org/org-name/other-repo-name.git"
		}
	]
}
```

Output of `composer diagnose`:

```
hecking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.2.10
PHP version: 7.3.29
PHP binary path: /Applications/MAMP/bin/php/php7.3.29/bin/php
OpenSSL version: OpenSSL 1.0.2u  20 Dec 2019
cURL version: 7.68.0 libz 1.2.11 ssl OpenSSL/1.0.2u
zip: extension present, unzip present, 7-Zip not available
```

When I run this command: 

```
/Applications/MAMP/bin/php/php7.4.21/bin/php  /usr/local/bin/composer install -vvv
```

I get the following output: 

```
Running 2.2.10 (2022-03-29 21:55:35) with PHP 7.4.21 on Darwin / 21.4.0
Reading ./composer.json (/private/tmp/test/example/composer.json)
[...]

PR #10678, opened 2022-03-30
Bitbucket: when OAuth consumer authentication fails mention that the `This is a private consumer` needs to be enabled

glaubinix:
We could try to detect this by matching either the description or error but who knows if Bitbucket will ever change this and we might easily miss the change. Just for reference the 400 response is `{"error_description": "Cannot use client_credentials with a consumer marked as \"public\". Calls for auto generated consumers should use urn:bitbucket:oauth2:jwt instead.", "error": "invalid_grant"}` 

See https://github.com/composer/composer/issues/10667


Seldaek:
Sounds good enough like this, thanks!

PR #10681, opened 2022-03-31
Correction the function getComposerInformation in SvnDriver.php.

nargil07:
Sometimes the getBaseComposerInformation had an exception so we write in the cache a string value but the funtion want to return null or Array value.
Now we write a empty string value if it happened

Bug in the 2.3
It's the fix about the issues 10680


Seldaek:
Thanks

Seldaek:
I had a closer look and in fact this should have been null all along. Tweaked further in https://github.com/composer/composer/commit/29513c15da7115874ea247fe281ccd48c768b2f5

nargil07:
I guessed that but i had no Time to be sure !
Thanks for all of it !
☺️

PR #10687, opened 2022-04-01
GitDriver: try to fetch default branch from remote

glaubinix:
The initial clone determined what the default branch of the cache git repository was. Changing it on the remote didn't have any impact on the local data. However, cloning it on a different machine would then store a different default branch on that machine. This could lead to different results for the same command on different machines.

Unfortunately, I couldn't find a way to update the default branch on the local cache because checkout and symbolic-ref are both unavailable.

### DIFF
--- a/src/Composer/Repository/Vcs/GitDriver.php
+++ b/src/Composer/Repository/Vcs/GitDriver.php
@@ -97,11 +97,8 @@ class GitDriver extends VcsDriver
             if (!(bool) Platform::getEnv('COMPOSER_DISABLE_NETWORK')) {
                 try {
                     $this->process->execute('git remote show origin', $output, $this->repoDir);
-                    $lines = $this->process->splitLines($output);
-                    foreach ($lines as $line) {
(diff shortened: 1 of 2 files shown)


src/Composer/Repository/Vcs/GitDriver.php
```diff
@@ -93,6 +94,20 @@ public function getRootIdentifier(): string
...
 
+            if (!(bool) Platform::getEnv('COMPOSER_DISABLE_NETWORK')) {
+                try {
+                    $this->process->execute('git remote show origin', $output, $this->repoDir);
+                    $lines = $this->process->splitLines($output);
+                    foreach ($lines as $line) {
+                        if (Preg::match('{^\s*HEAD branch:\s(.+)\s*$}', $line, $matches) > 0) {
+                            return $this->rootIdentifier = $matches[1];
+                        }
+                    }
```
Seldaek, line 105:
```suggestion
                    if (Preg::isMatch('{^\s*HEAD branch:\s(.+)\s*$}m', $output, $matches)) {
                        return $this->rootIdentifier = $matches[1];
                    }
```

PR #10700, opened 2022-04-03
Fix run-script not setting Path correctly on Windows

### LINKED ISSUE #10699: Running scripts from bin-dir no longer works after updating to 2.3 (on Windows 10)
My `composer.json`:

```json
{
	"scripts": {
		"phpstan": "phpstan analyse"
	},
	"require-dev": {
		"phpstan/phpstan": "^1.0",
	}
}
```

```
Checking composer.json: OK
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys:
OK
Checking composer version: OK
Composer version: 2.3.3
PHP version: 8.1.2
PHP binary path: php\php.exe
OpenSSL version: OpenSSL 1.1.1m  14 Dec 2021
cURL version: 7.77.0 libz 1.2.11 ssl OpenSSL/1.1.1m
zip: extension present, unzip not available, 7-Zip present (7z)
```

When I run this command: 

```
$ composer run phpstan -vvv
```

I get the following output: 

```
Running 2.3.3 (2022-04-01 22:15:35) with PHP 8.1.2 on Windows NT / 10.0
Reading ./composer.json (\composer.json)
Loading config file ./composer.json (\composer.json)
Checked CA file D:\WebDevelopment\conf\cacert.pem: valid
Executing command (): git branch -a --no-color --no-abbrev -v
Reading /Composer/composer.json (\Composer\composer.json)
Loading config file /Composer/composer.json (\Composer\composer.json)
Reading /vendor/composer/installed.json (\vendor\composer\installed.json)
Skipped loading "composer/package-versions-deprecated" as it is not in config.allow-plugins
Reading /Composer/vendor/composer/installed.json (\Composer\vendor\composer\installed.json)
> phpstan: phpstan analyse
Executing command (CWD): phpstan analyse
'phpstan' is not recognized as an internal or external command,
operable program or batch file.
```

Same on `composer self-update --snapshot`, downgrading to `--2.2` makes it work again:

```
> phpstan: phpstan analyse
Executing command (CWD): phpstan analyse
Note: Using configuration file phpstan.neon.

 [OK] No errors

 377/377 [============================] 100%
```

As per documentation:
> Before executing scripts, Composer's bin-dir is temporarily pushed on top of the PATH environment variable so that binaries of dependencies are directly accessible. 

Perhaps this no longer happens?

xPaw:
Fixes #10699.

`getenv` is case-insensitive as far as I understand it, which makes it not work correctly.

Not entirely sure why it works on 2.2, because the same code should be there.

It was changed here: https://github.com/composer/composer/commit/bd4d624cc7468623b29e56fa800b1750886b04c7#diff-4ec8bee1b3bdbc03e2097675b4586f3bb92ff70afff4e6ee5e4cc25847816e97L569

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -571,6 +571,10 @@ class EventDispatcher
     private function ensureBinDirIsInPath(): void
     {
         $pathEnv = 'PATH';
+
+        // checking if only Path and not PATH is set then we probably need to update the Path env
+        // on Windows getenv is case-insensitive so we cannot check it via Platform::getEnv and
+        // we need to check in $_SERVER directly
         if (!isset($_SERVER[$pathEnv]) && isset($_SERVER['Path'])) {
             $pathEnv = 'Path';
         }
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -571,7 +571,7 @@ class EventDispatcher
     private function ensureBinDirIsInPath(): void
     {
         $pathEnv = 'PATH';
-        if (false === Platform::getEnv('PATH') && false !== Platform::getEnv('Path')) {
+        if (!isset($_SERVER[$pathEnv]) && isset($_SERVER['Path'])) {
             $pathEnv = 'Path';
         }
 


src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -571,7 +571,7 @@ protected function popEvent(): ?string
     private function ensureBinDirIsInPath(): void
     {
         $pathEnv = 'PATH';
-        if (false === Platform::getEnv('PATH') && false !== Platform::getEnv('Path')) {
```
herndlm, line 574: I'm pretty sure the `Platform::getEnv` helper is used for a good reason and should not be removed :)

xPaw: `Platform::getEnv` will fallback to `getenv` which is not case sensitive, making this check moot.

GrahamCampbell: No, that's still not right. Env variables may exist in $_SERVER only, and not in putenv.

xPaw: Well, this is the code that was used before it was changed to getEnv.

xPaw:
I did a bisect on main branch, and it stops working with this commit: d2ede370be20aedfc74749bc87394a79ed4c1e52 which makes me think `symfony/process` (v2.8.52 => v5.4.0) upgrade is what actually made it stop working (disregarding the fact that PATH/Path capilization is still broken that this PR attemps to fix back).

Seldaek:
Thanks, I believe the fix makes sense. Probably could be checked even further by using `getenv()` without args which returns the whole env then we can do a case sensitive check in that array I assume. If you want to improve further.. 

But I'll merge this now as it is definitely a fix already.

PR #10701, opened 2022-04-04
GitDriver: try to fetch default branch form remote using auth

glaubinix:
Building on top of https://github.com/composer/composer/pull/10687

Just running `git remote show origin` works fine if authentication is set up via public/private key. Fetching the data from a remote that requires username/password authentication requires updating the remote URL first.

This also exposed the command output of `Git:: runCommand` which previously was just ignore because not necessary anywhere.


Seldaek:
Thanks

Issue #10696, opened 2022-04-02
php8.1 selfupdate error when no old files exist


it-can:
I get this error after a fresh install on my mac machine with php8.1, I run this command

```
composer self-update --clean-backups

composer self-update --clean-backups
You are already using the latest available Composer version 2.3.3 (stable channel).

In SelfUpdateCommand.php line 491:

  [TypeError]
  Composer\Command\SelfUpdateCommand::cleanBackups(): Argument #2 ($except) must be of type
   ?string, bool given, called in phar:///usr/local/bin/composer/src/Composer/Command/SelfU
  pdateCommand.php on line 215


Exception trace:
  at phar:///usr/local/bin/composer/src/Composer/Command/SelfUpdateCommand.php:491
 Composer\Command\SelfUpdateCommand->cleanBackups() at phar:///usr/local/bin/composer/src/Composer/Command/SelfUpdateCommand.php:215
 Composer\Command\SelfUpdateCommand->execute() at phar:///usr/local/bin/composer/vendor/symfony/console/Command/Command.php:298
 Symfony\Component\Console\Command\Command->run() at phar:///usr/local/bin/composer/vendor/symfony/console/Application.php:1015
 Symfony\Component\Console\Application->doRunCommand() at phar:///usr/local/bin/composer/vendor/symfony/console/Application.php:299
 Symfony\Component\Console\Application->doRun() at phar:///usr/local/bin/composer/src/Composer/Console/Application.php:334
 Composer\Console\Application->doRun() at phar:///usr/local/bin/composer/vendor/symfony/console/Application.php:171
 Symfony\Component\Console\Application->run() at phar:///usr/local/bin/composer/src/Composer/Console/Application.php:130
 Composer\Console\Application->run() at phar:///usr/local/bin/composer/bin/composer:83
 require() at /usr/local/bin/composer:29
```

PR #10704, opened 2022-04-04
[2.3] fix php8.1 error with self-update --clean-backups #10696

it-can:
Fix for #10696


Seldaek:
Thanks for the PR, however your change changed the semantics of cleanBackups as it always kept one file instead of cleaning all when no $except is passed. I also took a good look at all phpstan errors for this file to ensure we fix all edge cases as it's quite a critical code path, so I ended up fixing this differnetly in https://github.com/composer/composer/commit/106149d102bb6bc93a985b267ffc92b29935071d

PR #10710, opened 2022-04-05
Restore the ability to list the scripts in the 'run-script' command without providing a script

### LINKED ISSUE #10707: 2.3 regression: -l/--list broken for run-script command
```ShellSession
% composer --version
Composer version 2.3.3 2022-04-01 22:15:35
% composer run-script -l

                                             
  Not enough arguments (missing: "script").  
                                             

run-script [--timeout TIMEOUT] [--dev] [--no-dev] [-l|--list] [--] <script> [<args>...]

% composer run-script --list

                                             
  Not enough arguments (missing: "script").  
                                             

run-script [--timeout TIMEOUT] [--dev] [--no-dev] [-l|--list] [--] <script> [<args>...]

```

mbabker:
Fixes #10705
Fixes #10707

After #10476 the ability to list scripts from the `run-script` command without passing a script name (i.e. `composer run-script -l`) was lost because the `script` argument was made required.  This restores the ability to list the scripts without needing to pass something to fill the `script` argument while also failing if not listing scripts and a script name wasn't provided.

### DIFF
--- a/tests/Composer/Test/Command/RunScriptCommandTest.php
+++ b/tests/Composer/Test/Command/RunScriptCommandTest.php
@@ -12,13 +12,35 @@
 
 namespace Composer\Test\Command;
 
+use Composer\Command\RunScriptCommand;
 use Composer\Composer;
 use Composer\Config;
+use Composer\Console\Application;
 use Composer\Script\Event as ScriptEvent;
 use Composer\Test\TestCase;
+use Composer\Util\Filesystem;
+use Symfony\Component\Console\Tester\ApplicationTester;
 
 class RunScriptCommandTest extends TestCase
 {
+    /**
+     * @var string
+     */
+    private $home;
+
+    public function setUp(): void
+    {
+        $this->home = $this->getUniqueTmpDirectory();
+    }
+
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        $fs = new Filesystem();
+        $fs->removeDirectory($this->home);
+    }
+
     /**
      * @dataProvider getDevOptions
      * @param bool $dev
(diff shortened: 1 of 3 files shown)


stof (approved):

tests/Composer/Test/Command/RunScriptCommandTest.php
```diff
@@ -88,6 +88,48 @@ public function testDetectAndPassDevModeToEventAndToDispatching(bool $dev, bool
...
+                'initialize',
+                'requireComposer',
+                'getApplication',
+            ))
+            ->getMock();
+        $command->expects($this->any())->method('requireComposer')->willReturn($composer);
+        $command->expects($this->any())->method('getApplication')->willReturn($app);
+
+        $this->assertSame(0, $command->run($input, $output));
+    }
```
Seldaek, line 131: I really hate to see so much mocking here, this seems way overkill and will make for brittle tests.

Would you mind trying with https://symfony.com/doc/current/console.html#testing-commands ?

Combined with $this->getUniqueTmpDirectory() to get a temp dir, write a composer.json in it with some scripts and pass `-d` flag so composer uses that dir.

That'd be much a more realistic/integrated and thus more valuable test IMO.

mbabker: Test's updated

Seldaek:
Thanks for the test updates, that looks pretty good already! 

I should be able to build up some utilities to shorten the test bootstrap a bit further and then we have something really usable to improve the coverage of all our command code (which sits somewhere close to 0% right now).

And thanks for the fix :)

PR #10718, opened 2022-04-07
Fix discrepancy between documentation and code in package's dist reference configuration

sveneld:
https://github.com/composer/composer/pull/10488 in this pull request there is a mistake. In code used none for nullable reference configuration, in documentation used "null". I add "null" as possible value in code. 'none' config value should be removed in future.


stof:
I would rather fix the doc as the code never used the string `null` anywhere for now.

Seldaek:
I agree with @stof - fixed the docs as any misuse can easily be fixed and it's a fairly new option it's not worth adding BC cruft in the code IMO.

PR #10722, opened 2022-04-11
Add warning if host is accessed via verify_peer or verify_peer_name disabled

glaubinix:
Similar to the warning about accessing a host via insecure protocol, this adds a warning about accessing a host with verify_peer or verify_peer_name disabled.

```
composer update
Warning: Accessing api.github.com with verify_peer disabled.
```


Seldaek:
Thanks

Issue #10715, opened 2022-04-06
"composer validate" aborts with code 1 with "--no-check-lock" and JSON Text parse error on lock-file contents


ktomk:
given

~~~
{
    "config": {
        "lock": false
    }
}
~~~

~~~
composer config lock false
~~~

and then creating a zero-byte `composer.lock` file

~~~
< /dev/null >| composer.lock
~~~

"composer validate --no-check-lock" aborts with exit code 1 choking on the composer.lock file not being valid JSON Text.

Despite the `--no-check-lock`:

~~~
$ composer --no-ansi validate --no-check-lock


  [Seld\JsonLint\ParsingException]
  "./composer.lock" does not contain valid JSON
  Parse error on line 1:

  ^
  Expected one of: 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '['


validate [--no-check-all] [--no-check-lock] [--no-check-publish] [--no-check-version] [-A|--with-dependencies] [--strict] [--] [<file>]

$ echo $?
1
~~~

## Expected Behavior

The `composer.lock` file is not checked as per `--no-check-lock` which should include dropping any errors in JSON Text processing if the file exists.

## Additional Info

Report against

~~~
$ composer --version
Composer version 2.2.8 2022-03-15 12:55:20
~~~

An object with no properties prevents the abortion:

~~~
printf '{}' >| composer.lock
~~~

And interestingly further validation checks on the lock file (?):

~~~
$ composer config lock true # for brevity
$ composer --no-ansi validate --strict
./composer.json is valid
~~~

Most of the validate messages related to the lock file with `lock: false` and `--no-check-lock` appear noisy to me, but this is rather cosmetic.

Seldaek:
Sorry but that's not doable..  when loading a Composer instance the lock file is loaded if present, and if it is broken it fails hard at that point, no matter which command you are running. If you don't want a lock then don't create a null byte file..

ktomk:
Okay, this is a bit what I thought as well (from output it didn't look like it's from the validate command specifically).

And yes, I did create the null byte composer.lock intentionally in explorative testing.

What wondered me in the end thought is that an empty object as lock file prevented any errors to be shown regarding the lock file (e.g. the warning that it is out of date and one should run composer update) with the verify command.

[...]

PR #10723, opened 2022-04-11
Patch "validate --no-check-lock"

ktomk:
if no lock-file is configured, turn lock file validation errors into warnings (implicit --no-check-lock) unless those are explicitly promoted via the new `--check-lock` option.

- `{"config": {"lock": false}}` is an implicit `--no-check-lock` for   composer validate.
- `--check-lock` overrides an (implicit or explicit) `--no-check-lock`, always.

rationale is that if no composer.lock file is configured in composer.json, the validate command should not _fatal_ over it but instead demote to _warning_ as the suggestion given in the warning won't work unless composer.lock is configured again (and there is no error as the lock-file is ignored by composer).

the difference this makes is the one between an exit status of zero or non-zero, between success and failure.

that is, if (and as) composer validate is used in check scripts (e.g. CI, tests, hooks etc.), it won't stumble over old lock-file artifacts, whatever the reasons may be.

(originally reported and fix idea / context in #10715)

### DIFF
--- a/src/Composer/Command/ValidateCommand.php
+++ b/src/Composer/Command/ValidateCommand.php
@@ -45,6 +45,7 @@ class ValidateCommand extends BaseCommand
             ->setDescription('Validates a composer.json and composer.lock.')
             ->setDefinition(array(
                 new InputOption('no-check-all', null, InputOption::VALUE_NONE, 'Do not validate requires for overly strict/loose constraints'),
+                new InputOption('check-lock', null, InputOption::VALUE_NONE, 'Check if lock file is up to date (even config.lock is false, overrides --no-check-lock)'),
                 new InputOption('no-check-lock', null, InputOption::VALUE_NONE, 'Do not check if lock file is up to date'),
                 new InputOption('no-check-publish', null, InputOption::VALUE_NONE, 'Do not check for publish errors'),
                 new InputOption('no-check-version', null, InputOption::VALUE_NONE, 'Do not report a warning if the version field is present'),
@@ -95,6 +96,8 @@ EOT
 
         $lockErrors = array();
         $composer = Factory::create($io, $file, $input->hasParameterOption('--no-plugins'));
+        // config.lock = false ~= implicit --no-check-lock; --check-lock overrides
+        $checkLock = ($checkLock && $composer->getConfig()->get('lock')) || $input->getOption('check-lock');
         $locker = $composer->getLocker();
         if ($locker->isLocked() && !$locker->isFresh()) {
             $lockErrors[] = '- The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update` or `composer update <package name>`.';
--- a/phpstan/baseline.neon
+++ b/phpstan/baseline.neon
@@ -2376,6 +2376,11 @@ parameters:
 			path: ../src/Composer/Command/ValidateCommand.php
 
 		-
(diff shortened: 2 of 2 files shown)


src/Composer/Command/ValidateCommand.php
```diff
@@ -95,6 +96,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $lockErrors = array();
         $composer = Factory::create($io, $file, $input->hasParameterOption('--no-plugins'));
+        // config.lock = false ~= implicit --no-check-lock; --check-lock overrides
+        $checkLock = ($checkLock && $composer->getConfig()->getBool('lock')) || $input->getOption('check-lock');
```
stof, line 100: Please keep using `get()` instead of introducing `getBool`.

There is a work in progress to create a phpstan extension to get a precise type based on the config setting being requested, which will solve this in the future. In the meantime, keep the phpstan error in the baseline, as done elsewhere.

ktomk: thanks, this answers the main question I had with this draft @stof.

can you give me a pointer to the phpstan baseline process in composer?

stof: I would not even put the explicit bool cast here (as we already have a boolean, even though phpstan does not know it yet because #10635 is not done yet)

ktomk: > I would not even put the explicit bool cast here

&

> In the meantime, keep the phpstan error in the baseline, as done elsewhere.

@stof I removed the bool cast as well and could now get more traction with phpstan. however when updating the baseline-file:

~~~
$ vendor/bin/phpstan analyse --configuration=phpstan/config.neon -b phpstan/baseline.neon 
 393/393 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%


                                                                                                                        
 [OK] Baseline generated with 3015 errors.                                                                              
                                                                                                                        

~~~

it is with many more edits (and out of context to the change itself):

~~~
 phpstan/baseline.neon | 133 ++++++--------------------------------------------
 1 file changed, 14 insertions(+), 119 deletions(-)
~~~

Just that it does not get unnoticed - I've kept the out-of-context changes out of the file for this PR.

stof: if you run it on PHP 8.1, there is a specific baseline file in the project.

ktomk: @stof: Yes, but I've set this PR against 2.2 so this may not be the culprit here. Let me know if you see some actions to do or any other feedback.

src/Composer/Command/ValidateCommand.php
```diff
@@ -45,6 +45,7 @@ protected function configure()
             ->setDescription('Validates a composer.json and composer.lock.')
             ->setDefinition(array(
                 new InputOption('no-check-all', null, InputOption::VALUE_NONE, 'Do not validate requires for overly strict/loose constraints'),
+                new InputOption('check-lock', null, InputOption::VALUE_NONE, 'Check if lock file is up to date (even config.lock is false, overrides --no-check-lock)'),
```
Seldaek, line 48: I'd keep this simpler, or leave the flag altogether as I don't really see the point if you disabled the lock file..

```suggestion
                new InputOption('check-lock', null, InputOption::VALUE_NONE, 'Check if lock file is up to date (even when config.lock is false)'),
```

PR #10724, opened 2022-04-12
Update on check out of commits

Livecan:
DOCS Added the info on checking out commits. This is particularly helpful because # is an allowed character in git branch name, but gets interpreted by Composer, which may cause problems.


stof:
The commit pinning is a huge hack added in the early days of composer, with lots of drawbacks. And it is considered legacy by the maintainers. So I would not add this paragraph that recommends it (especially not if the big warning is not repeated there).
See the block on a blue background on https://getcomposer.org/doc/04-schema.md#package-links (scrolling a bit)

Seldaek:
Yup I think I'd rather not highlight this any more than it is.

Livecan:
ok, I understand those reasons. Would you consider somehow somewhere mentioning that # should not be part of the branch name when wanting to use a particular branch in composer?
I now it now, but spent good few hours on finding out why my branch couldn't be used by composer. The confusion was only bigger when composer told me that my branch name (e.g. roman#1928) wasn't among the allowed branches - followed by a list containing the branch with # symbol.

Seldaek:
Yes adding that to the docs sounds reasonable for sure. Or making a warning clearer in the composer output. I am not sure exactly what you mean by `composer told me that my branch name (e.g. roman#1928) wasn't among the allowed branches - followed by a list containing the branch with # symbol`

PR #10725, opened 2022-04-13
Clarify that autoloader-suffix should be a non-empty-string

### LINKED ISSUE #10720: `composer.json`: empty string in `autoloader-suffix`
My `composer.json`:
```json
{
    <...>
    "config": {
        "autoloader-suffix": ""
    }
}
```

```
Checking composer.json: OK
Checking git settings: OK
Checking disk free space: OK
...
Checking composer version: OK
Composer version: 2.3.4
PHP version: 7.4.27
...
```

When I run this command: 

```
composer update -W -o
```

Actual result: `class ComposerAutoloaderInit... {...}`(`autoload_real.php`)

And I expected this to happen: `class ComposerAutoloaderInit {...}`

Seldaek:
Fixes #10720

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -159,12 +159,12 @@ class AutoloadGenerator
     /**
      * @param string $targetDir
      * @param bool $scanPsrPackages
-     * @param string $suffix
+     * @param string|null $suffix
      * @return int
      * @throws \Seld\JsonLint\ParsingException
      * @throws \RuntimeException
      */
-    public function dump(Config $config, InstalledRepositoryInterface $localRepo, RootPackageInterface $rootPackage, InstallationManager $installationManager, $targetDir, $scanPsrPackages = false, $suffix = '')
+    public function dump(Config $config, InstalledRepositoryInterface $localRepo, RootPackageInterface $rootPackage, InstallationManager $installationManager, $targetDir, $scanPsrPackages = false, $suffix = null)
     {
         if ($this->classMapAuthoritative) {
             // Force scanPsrPackages when classmap is authoritative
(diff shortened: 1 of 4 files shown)


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -159,12 +159,12 @@ public function setPlatformRequirementFilter(PlatformRequirementFilterInterface
     /**
      * @param string $targetDir
      * @param bool $scanPsrPackages
-     * @param string $suffix
+     * @param non-empty-string|null $suffix
```
stof, line 162: this should stay `string` if the code checks for the empty string.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -374,16 +374,16 @@ public static function autoload(\$class)
...
                 $content = file_get_contents($vendorPath.'/autoload.php');
                 if (Preg::isMatch('{ComposerAutoloaderInit([^:\s]+)::}', $content, $match)) {
                     $suffix = $match[1];
                 }
             }
 
-            if (!$suffix) {
-                $suffix = $config->get('autoloader-suffix') ?: md5(uniqid('', true));
+            if (null === $suffix) {
+                $suffix = $config->get('autoloader-suffix') ?? md5(uniqid('', true));
```
stof, line 386: this is a no-go in the 2.2 branch

src/Composer/Autoload/AutoloadGenerator.php
Seldaek: Yeah my point was more to have self-documenting code, but you cannot really trust that the caller respects the type in a public API, so you still have to verify IMO. But anyway back to `string`, not so important here.

stof: Another option is to configure phpstan with `treatPhpDocTypesAsCertain: false` so that it does not complain when doing validation of types defined only in phpdoc

Seldaek: Yeah we already have that on main branch so I'll do it on 2.2 too.

Seldaek: Ah well, that didn't fix it.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -159,12 +159,12 @@ public function setPlatformRequirementFilter(PlatformRequirementFilterInterface
...
      * @param string $targetDir
      * @param bool $scanPsrPackages
-     * @param string $suffix
+     * @param string|null $suffix
      * @return int
      * @throws \Seld\JsonLint\ParsingException
      * @throws \RuntimeException
      */
-    public function dump(Config $config, InstalledRepositoryInterface $localRepo, RootPackageInterface $rootPackage, InstallationManager $installationManager, $targetDir, $scanPsrPackages = false, $suffix = '')
+    public function dump(Config $config, InstalledRepositoryInterface $localRepo, RootPackageInterface $rootPackage, InstallationManager $installationManager, $targetDir, $scanPsrPackages = false, $suffix = null)
```
WinterSilence, line 167: it's break BC

PR #10728, opened 2022-04-13
Docs: clarify carret contraint with ^0.0.x constraint

glaubinix:
Make it clearer that `^0.0.3` does not allow installing `0.0.4`

### DIFF
--- a/doc/articles/versions.md
+++ b/doc/articles/versions.md
@@ -181,7 +181,7 @@ The `^` operator behaves very similarly, but it sticks closer to semantic
 versioning, and will always allow non-breaking updates. For example `^1.2.3`
 is equivalent to `>=1.2.3 <2.0.0` as none of the releases until 2.0 should
 break backwards compatibility. For pre-1.0 versions it also acts with safety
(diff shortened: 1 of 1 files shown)


xabbuh:
This does not look correct to me. Where did you find this information?

glaubinix:
Might be easier with an actual example:
```
{
    "repositories": [
        {"type": "package", "package": [
            {"name": "acme/test", "version": "0.0.2"},
            {"name":  "acme/test", "version": "0.0.1.1"},
            {"name": "acme/test", "version": "0.0.1"}
        ]}
    ],
    "require": {
        "acme/test": "^0.0.1"
    }
}
```
And then run `composer update --no-install`
```
composer update --no-install 
Updating dependencies
Lock file operations: 1 install, 0 updates, 0 removals
  - Locking acme/test (0.0.1.1)
Writing lock file
```

stof:
@xabbuh the major version 0 makes the next segment a major version as well. By that definition, it is recursive in case that next segment is itself 0.
Anyway, pre-1.0 versions are full of surprise in semver.

Seldaek:
Thanks

PR #10729, opened 2022-04-13
BitbucketDriver: remove unnecessary HTTP call to fetch the root identifier

glaubinix:
The `mainbranch` information is already available when fetching the repo data. The HTTP call in `getMainBranchData` was calling the same endpoint as `getRepoData` but only fetch the `mainbranch` data.

Windows test failing with a random timeout


Seldaek:
Thanks

PR #10735, opened 2022-04-14
ComposerRepository: add a warning if the packages key doesn't match the package name

glaubinix:
Even though, this is not enforced by Composer. A Composer repository using the `packages` key to return a set of packages in the `packages.json` is expected to use the package name as index based on [documentation](https://getcomposer.org/doc/05-repositories.md#packages). 

Sample packages.json response of repositories not following that standard
```
{
    "packages": {
        "test": {
            "dev-main": {
                "name": "acme/test",
                "version": "dev-main"
            }
        }
    }
}
```

Output with composer update --no-install
```
composer update --no-install
Warning: the packages key 'test' doesn't match the name defined in the package metadata 'acme/test' in repository https://example.org
Updating dependencies
Nothing to modify in lock file
```


Seldaek:
Thanks

PR #10737, opened 2022-04-14
Use `addOption` and `addArgument` for command input definition

GromNaN:
Working on adding bash completion to commands #10320, I found the implementation very verbose. To make things easier, I added to Symfony 6.1 a feature to define suggested values to argument and option definition https://github.com/symfony/symfony/pull/44948.

This preliminary PR uses the functions `Command::addArgument` and `Command::addArgument`; so that I can backport the additional argument to define suggested values for each argument or option.

This change does not affect performances.

### DIFF
--- a/src/Composer/Command/ArchiveCommand.php
+++ b/src/Composer/Command/ArchiveCommand.php
@@ -47,15 +47,13 @@ class ArchiveCommand extends BaseCommand
         $this
             ->setName('archive')
             ->setDescription('Creates an archive of this composer package.')
-            ->setDefinition(array(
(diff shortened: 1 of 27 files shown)


GromNaN:
I'm not sure what is the preferred solution: add argument `$suggestedValues` to `addOption` and `addArgument`, or extend classes `InputOption` and `InputArgument` to add add argument `$suggestedValues` to their constructior (currently done in #10320)

Seldaek:
IMO both approaches are fine, not sure what the implications are, but keeping it in a way that is compatible with newer symfony would be good so we can one day upgrade smoothly.

Not sure why you closed here?

GromNaN:
I implemented the other way in #10320. Less changes, closer from the Symfony code.

PR #10756, opened 2022-04-27
#10738 - Allow config unset for unrecognized keys

### LINKED ISSUE #10738: `composer config xxx --unset` should support any `xxx`
For specific usecases, composer.json can contain keys not defined in the composer.json schema. Such composer.json is valid and accepted by composer.

However `composer config xxx --unset` command supports only the keys defined in the schema, the following error is currently printed:

```
  [InvalidArgumentException]
  Setting require-release does not exist or is not supported by this command
```

This is a feature request to support any key to be unset, as such command is does not need any knowledge about the key from the schema.

Example composer.json file:

```
{
    "name": "atk4/data",
    "description": "Agile Data - Database access abstraction framework",
    "license": "MIT",
    "type": "library",
    "version": "dev-develop",
    "require": {
        "php": ">=7.4 <8.2",
        "atk4/core": "dev-develop"
    },
    "require-release": {
        "php": ">=7.4 <8.2",
        "atk4/core": "~3.2.0"
    }
}
```

when `composer config require-release --unset` is run, I expect the file to change to:

```diff
 {
     "name": "atk4/data",
     "description": "Agile Data - Database access abstraction framework",
     "license": "MIT",
     "type": "library",
     "version": "dev-develop",
     "require": {
         "php": ">=7.4 <8.2",
         "atk4/core": "dev-develop"
-    },
-    "require-release": {
-        "php": ">=7.4 <8.2",
-        "atk4/core": "~3.2.0"
-    }
+    }
 }
```

iionut:
Fix #10738 

Allows unsetting any config key and not only the ones from the recognised schema.

Because key removal is done without matching the key in the recognized composer.json schema, leaving the key removal before other checks would have stopped the execution there without reaching the specific config add (eg. `composer config repo [--unset]` which was right after the removal)

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -634,11 +634,6 @@ EOT
         if ($input->getOption('global') && (isset($uniqueProps[$settingKey]) || isset($multiProps[$settingKey]) || strpos($settingKey, 'extra.') === 0)) {
             throw new \InvalidArgumentException('The ' . $settingKey . ' property can not be set in the global config.json file. Use `composer global config` to apply changes to the global composer.json');
         }
-        if ($input->getOption('unset')) {
-            $this->configSource->removeProperty($settingKey);
-
-            return 0;
-        }
         if (isset($uniqueProps[$settingKey])) {
             $this->handleSingleValue($settingKey, $uniqueProps[$settingKey], $values, 'addProperty');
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Great, thanks! I think I'll just roll it in the next 2.3 as it's such a small patch, even though it's technically a new feature.

PR #10757, opened 2022-05-02
Inline PackageEvents constants

nicolas-grekas:
To make it easier to grep the source for those.


stof (approved):

Seldaek:
Yup good idea, thanks

PR #10769, opened 2022-05-11
Move navigation arrows to the bottom

carlosbuenosvinos:
Probably by merging other PR the arrows were not at the bottom. 

I have checked the rest of the pages and none was having this behavior, so I understand is a mistake. Keep the good work!


Seldaek:
Definitely a mistake, thanks

PR #10771, opened 2022-05-12
CurlDownloader: detect max file size for gzipped responses with missing content-length header

glaubinix:
`size_download` of the `$progress` array  for gzipped responses refers to the gzipped size making it possible to download larger files than specified via the `max_file_size` if no `Content-Length` header is defined in the response.


Seldaek:
> `size_download` of the `$progress` array for gzipped responses refers to the gzipped size making it possible to download larger files than specified via the `max_file_size` if no `Content-Length` header is defined in the response.

can you put that somewhere in the code as a comment? Sounds like if it needs explaining, it is the best place to do it :)

PR #10772, opened 2022-05-12
Remove dead code in poolbuilder

Seldaek:
In https://github.com/composer/composer/pull/8850 and specifically https://github.com/composer/composer/commit/912aecb6661b068cf187fb6fd0359798838d86ad#diff-d6d8d75ac1397a615f08c7a046ddbb11ad1ae869623f120be02ccd3c8e49b583L202 I think this whole loop was made useless.

The way it looks, it puts $package inside $aliasedPackages, then creates a constraint with the package's version, and then checks if that constraint matches the version, so $found will always end up being true as $package's version is always equal to itself, and aliases are never unset/removed.

cc @Toflar

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -234,30 +234,6 @@ class PoolBuilder
             $this->loadPackagesMarkedForLoading($request, $repositories);
         }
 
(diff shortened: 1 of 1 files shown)


stof (approved):

Seldaek:
Closing as #10773 took over the code for greater purposes instead of deleting it.

PR #10773, opened 2022-05-12
Allow using temporary update constraints on all packages (incl non-root)

### LINKED ISSUE #10436: Allow `update --with` for non-root packages
#8860 added the `--with` parameter to `update` to run an update with a specific version constraint on a package.

In #8756, @Seldaek wrote without elaborating:
> This should be only allowed for packages required by the root package for now, we can re-evaluate later if needed.

Following the resolution of the issue, @froemken (not OP) commented that the change did not satisfy his needs because of its restricted application:
https://github.com/composer/composer/issues/8756#issuecomment-861315906

It seems that @louisitvn was also looking for this functionality for non-root packages:
https://github.com/composer/composer/issues/8756#issuecomment-997533294

I, too, am looking for this functionality.

What was the motivation for restricting `update --with` to root packages? Is there a technical reason why it can't apply to any package? If not, can the restriction be removed?

Seldaek:
Fixes #10436

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -21,6 +21,7 @@ use Composer\Pcre\Preg;
 use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
 use Composer\Package\Version\VersionParser;
+use Composer\Semver\Constraint\ConstraintInterface;
 use Composer\Util\HttpDownloader;
 use Composer\Semver\Constraint\MultiConstraint;
 use Composer\Package\Link;
@@ -144,20 +145,13 @@ EOT
             }
         }
 
-        $rootPackage = $composer->getPackage();
-        $rootRequires = $rootPackage->getRequires();
-        $rootDevRequires = $rootPackage->getDevRequires();
+        $parser = new VersionParser;
+        $temporaryConstraints = [];
         foreach ($reqs as $package => $constraint) {
-            if (isset($rootRequires[$package])) {
-                $rootRequires[$package] = $this->appendConstraintToLink($rootRequires[$package], $constraint);
-            } elseif (isset($rootDevRequires[$package])) {
-                $rootDevRequires[$package] = $this->appendConstraintToLink($rootDevRequires[$package], $constraint);
-            } else {
-                throw new \UnexpectedValueException('Only root package requirements can receive temporary constraints and '.$package.' is not one');
-            }
+            $temporaryConstraints[strtolower($package)] = $parser->parseConstraints($constraint);
         }
-        $rootPackage->setRequires($rootRequires);
-        $rootPackage->setDevRequires($rootDevRequires);
+
+        $rootPackage = $composer->getPackage();
         $rootPackage->setReferences(RootPackageLoader::extractReferences($reqs, $rootPackage->getReferences()));
         $rootPackage->setStabilityFlags(RootPackageLoader::extractStabilityFlags($reqs, $rootPackage->getMinimumStability(), $rootPackage->getStabilityFlags()));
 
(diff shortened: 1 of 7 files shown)


src/Composer/Command/UpdateCommand.php
```diff
@@ -144,20 +145,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-            } else {
-                throw new \UnexpectedValueException('Only root package requirements can receive temporary constraints and '.$package.' is not one');
-            }
+            $temporaryConstraints[strtolower($package)] = $parser->parseConstraints($constraint);
         }
-        $rootPackage->setRequires($rootRequires);
-        $rootPackage->setDevRequires($rootDevRequires);
+
+        $rootPackage = $composer->getPackage();
+        $rootPackage->setTemporaryConstraints($temporaryConstraints);
```
stof, line 155: shouldn't the temporary constraints be added in the installer rather than in the root package ?

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -234,24 +240,28 @@ public function buildPool(array $repositories, Request $request): Pool
...
+                foreach ($packageAndAliases as $packageOrAlias) {
                     if (CompilingMatcher::match($constraint, Constraint::OP_EQ, $packageOrAlias->getVersion())) {
                         $found = true;
                     }
                 }
+
                 if (!$found) {
-                    foreach ($aliasedPackages as $index => $packageOrAlias) {
+                    foreach ($packageAndAliases as $index => $packageOrAlias) {
                         unset($this->packages[$index]);
```
stof, line 265: shouldn't the unsetting be done separately for each package depending on whether it matches or no ?

src/Composer/DependencyResolver/Problem.php
```diff
@@ -294,6 +294,16 @@ public static function getMissingPackageReason(RepositorySet $repositorySet, Req
                 }
             }
 
+            $tempReqs = $repositorySet->getTemporaryConstraints();
+            if (isset($tempReqs[$packageName])) {
+                $filtered = array_filter($packages, function ($p) use ($tempReqs, $packageName): bool {
+                    return $tempReqs[$packageName]->matches(new Constraint('==', $p->getVersion()));
+                });
+                if (0 === count($filtered)) {
+                    return array("- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but '.(self::hasMultipleNames($packages) ? 'these conflict' : 'it conflicts').' with your temporary update constraint ('.$packageName.':'.$tempReqs[$packageName]->getPrettyString().').');
```
stof, line 303: temporary constraints are not about the root composer.json, but about the command arguments.

stof (changes requested):

src/Composer/DependencyResolver/PoolBuilder.php
Seldaek: I am not sure.. the package and its aliases are usually connected and have to be all installed or none. So here we keep the whole group if any of the aliased or alias matches the constraint, and if none match the whole thing is removed. I think this is correct.

src/Composer/DependencyResolver/Problem.php
Seldaek: The "root composer.json" part is about the require/constraint which caused the problem, not about the temporary constraint, this is correct as is (but slightly obscure code to be fair).

src/Composer/Command/UpdateCommand.php
Seldaek: Yup that probably makes more sense I'll update.

src/Composer/DependencyResolver/Problem.php
stof: Then, can you add a test showing what happens when a temporary constraint causes a conflict on a transitive dependency ?

Issue #10745, opened 2022-04-20
Composer Deleted all my files on under root folder


tuyenvandovn:
Hello,
I have executed a Composer file which belongs to Madelineproto yesterday. which project store under home/www/animedono.com. 
After executing, All of my websites files are gone even they were not under home/www/animedono.com. Which means the Script step back to home/www/ to Delete all the files.
I tried to remove the scripts part in composer file and execute it again, but the deleting was still happened.
I wonder is it madelineproto script problem or Composer problem. But I suspect Composer more since I removed the suspicious rows on madelineproto composer file..

Server Information: 
Panel: AAPanel.
Composer Version: 2.2.3
Php Version: 8.0
Total of affected websites: more than 20

My `composer.json`:
```

{
    "name": "danog/madelineproto",
    "description": "PHP implementation of telegram's MTProto protocol.",
    "type": "project",
    "license": "AGPL-3.0-only",
    "homepage": "https://docs.madelineproto.xyz",
    "keywords": [
        "telegram",
        "mtproto",
        "protocol",
        "bytes",
        "messenger",
        "client",
        "PHP",
        "video",
        "stickers",
        "audio",
        "files",
        "GB"
    ],
    "conflict": {
        "krakjoe/pthreads-polyfill": "*",
        "ext-pthreads": "*"
    },
    "require": {
        "php-64bit": ">=8.0",
        "danog/primemodule": "^1",
        "erusev/parsedown": "^1.7",
        "symfony/polyfill-mbstring": "*",
        "ext-mbstring": "*",
        "ext-json": "*",
        "ext-xml": "*",
        "ext-dom": "*",
        "ext-filter": "*",
        "ext-hash": "*",
        "ext-zlib": "*",
        "ext-fileinfo": "*",
        "amphp/amp": "^2",
        "amphp/http-client": "^4",
        "amphp/socket": "^1",
        "amphp/dns": "^1",
        "amphp/byte-stream": "^1",
        "amphp/file": "^2",
        "amphp/mysql": "^2",
        "amphp/postgres": "^1.2",
        "danog/dns-over-https": "^0.2",
        "amphp/http-client-cookies": "^1",
        "danog/tg-file-decoder": "^0.1",
        "danog/magicalserializer": "^1.0",
        "league/uri": "^6",
        "danog/ipc": "^0.1",
        "amphp/log": "^1.1",
        "danog/loop": "^0.1.0",
        "phpseclib/phpseclib": "^3",
[...]

PR #10775, opened 2022-05-12
Fix cmd splitting paths on commas

johnstevenson:
Cmd apparently splits paths on commas, which can have bad consequences when calling destructive commands like `rmdir`.

Simply fixed by double-quoting the param (as with white-space).

See https://github.com/composer/composer/issues/10745#issuecomment-1118604588


Seldaek:
Thanks

PR #10785, opened 2022-05-18
GitHub: add support for files >1MB to getFileContent

glaubinix:
See GitHub documentation https://docs.github.com/en/rest/repos/contents#size-limits


Seldaek:
Thanks

PR #10798, opened 2022-05-24
Add audit command to check for security issues

### LINKED ISSUE #10329: Security Advisory Features
For a while now we have had security advisories available through an API on packagist.org: https://packagist.org/apidoc#list-security-advisories. Packagist.org also lists advisories on packages and marks vulnerable versions with a warning sign. However this information is not easily accessible in Composer itself. A common solution to avoid installing vulnerable versions is requiring the [roave/security-advisories](https://github.com/Roave/SecurityAdvisories) package which conflicts with all the packages in https://github.com/FriendsOfPHP/security-advisories. The same database is imported into packagist.org (with additional ones coming, e.g. https://github.com/composer/packagist/pull/1206). There are also various tools available which scan repositories either on the CLI or as a service using the same database and/or others (e.g. https://github.com/symfonycorp/security-checker-action).

As updating vulnerable versions and identifying them is a core task of working with dependencies, Composer should have tools built in to deal with them. In particular:

- `composer audit` command to display a list of vulnerable versions with details on the advisories used in the current composer.lock. Should ideally also tell the user what they need to update to get to safe versions, e.g. suggest a partial update command?
- `composer update` should allow filtering out any vulnerable versions. There are a two main alternatives
  - Keep the default behavior as-is, add a new option, e.g. `--filter-vulnerable`
  - Change the default behavior to filter out any vulnerable version, and add an option to install them anyway, e.g. `--allow-vulnerable`
  
  Handling partial updates will likely require additional options: Should locked versions remain or should they also get updated to a secure release? Should there be warnings printed about vulnerable versions in this case?
- Print warnings about vulnerable installed versions on all updates and installs.

The vulnerabilities should be readable from additional composer repositories as well, so that repositories like drupal's or magento's can publish their own advisories if they wish.

GuySartorelli:
Closes #10329

## Description
Adds an audit command which checks for security vulnerability advisories for installed packages using [the API on packagist.org](https://packagist.org/apidoc#list-security-advisories).

Also adds optional auditing (on by default) to the following commands:
- `create-project`
- `update`
- `install`
- `require`
- `remove`

### Table output (default)
audit-output-table

### Plain output
audit-output-plain

## Things the audit command does not do:
- Confirm the packages in `vendor` dir match what `composer.lock` says they should be
- If composer.lock is missing, check for advisories based on the constraints in the `composer.json` file

## Out of scope, but would be a good future enhancement
- A new `--no-insecure` (or `--allow-insecure`) flag could be added to `update` and `require`. This could use `Auditor` to take versions with advisories out of the version candidate pool, so that you can't accidentally update to or require an insecure version of a package.
- new config in composer.json where you can declare specific advisories which should be ignored by the audit (e.g. the advisory doesn't apply for your project and you don't want audit to keep failing because of it). Would likely need to include the advisory ID in audit output for this to be feasible.

## Things I'm unsure about
- I don't know how to resolve the PHPStan failures - some seem overly strict so I'm hesitant to just blindly follow them.
- Is there a clean way to test the different formats are working as expected in a unit test?

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -103,6 +103,8 @@ resolution.
 * **--no-autoloader:** Skips autoloader generation.
 * **--no-progress:** Removes the progress display that can mess with some
   terminals or scripts which don't handle backspace characters.
+* **--no-audit:** Does not run the audit step after installation is complete.
+* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
 * **--optimize-autoloader (-o):** Convert PSR-0/4 autoloading to classmap to get a faster
   autoloader. This is recommended especially for production, but can take
   a bit of time to run so it is currently not done by default.
@@ -182,6 +184,8 @@ and this feature is only available for your root package dependencies.
 * **--dev:** Install packages listed in `require-dev` (this is the default behavior).
 * **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader generation skips the `autoload-dev` rules.
 * **--no-install:** Does not run the install step after updating the composer.lock file.
+* **--no-audit:** Does not run the audit steps after updating the composer.lock file.
+* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
 * **--lock:** Only updates the lock file hash to suppress warning about the
   lock file being out of date.
 * **--with:** Temporary version constraint to add, e.g. foo/bar:1.0.0 or foo/bar=1.0.0
@@ -253,6 +257,8 @@ If you do not specify a package, Composer will prompt you to search for a packag
   terminals or scripts which don't handle backspace characters.
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
+* **--no-audit:** Does not run the audit steps after updating the composer.lock file.
+* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
 * **--update-no-dev:** Run the dependency update with the `--no-dev` option.
 * **--update-with-dependencies (-w):** Also update dependencies of the newly required packages, except those that are root requirements.
 * **--update-with-all-dependencies (-W):** Also update dependencies of the newly required packages, including those that are root requirements.
@@ -850,6 +856,8 @@ By default the command checks for the packages on packagist.org.
   mode.
 * **--remove-vcs:** Force-remove the VCS metadata without prompting.
 * **--no-install:** Disables installation of the vendors.
+* **--no-audit:** Does not run the audit steps after installation is complete.
+* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
 * **--ignore-platform-reqs:** ignore all platform requirements (`php`, `hhvm`,
   `lib-*` and `ext-*`) and force the installation even if the local machine does
   not fulfill these.
@@ -954,6 +962,22 @@ php composer.phar archive vendor/package 2.0.21 --format=zip
   or zip (default: "tar").
 * **--dir:** Write the archive to this directory (default: ".")
 * **--file:** Write the archive with the given file name.
+
+## audit
+
+This command is used to audit the packages in your composer.lock
+for possible security issues. Currently this only checks for and
+lists security vulnerability advisories according to the
+[packagist api](https://packagist.org/apidoc#list-security-advisories).
+
+```sh
+php composer.phar audit
+```
+
+### Options
+
+* **--no-dev:** Disables auditing of require-dev packages.
+* **--format (-f):** Audit output format. Must be "table" (default) or "plain".
 
 ## help
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -967,7 +967,7 @@ php composer.phar archive vendor/package 2.0.21 --format=zip
 
 ## audit
 
-This command is used to audit the packages in your composer.lock
+This command is used to audit the packages you have installed
 for possible security issues. Currently this only checks for and
 lists security vulnerability advisories according to the
 [Packagist.org api](https://packagist.org/apidoc#list-security-advisories).
@@ -980,6 +980,7 @@ php composer.phar audit
 
 * **--no-dev:** Disables auditing of require-dev packages.
 * **--format (-f):** Audit output format. Must be "table" (default), "plain", or "summary".
+* **--locked:** Audit packages from the lock file, regardless of what is currently in vendor dir.
 
 ## help
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -301,6 +301,8 @@ uninstalled.
   terminals or scripts which don't handle backspace characters.
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
+* **--no-audit:** Does not run the audit steps after installation is complete.
+* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
 * **--update-no-dev:** Run the dependency update with the --no-dev option.
 * **--update-with-dependencies (-w):** Also update dependencies of the removed packages.
   (Deprecated, is now default behavior)
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -104,7 +104,7 @@ resolution.
 * **--no-progress:** Removes the progress display that can mess with some
   terminals or scripts which don't handle backspace characters.
 * **--no-audit:** Does not run the audit step after installation is complete.
-* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
 * **--optimize-autoloader (-o):** Convert PSR-0/4 autoloading to classmap to get a faster
   autoloader. This is recommended especially for production, but can take
   a bit of time to run so it is currently not done by default.
@@ -185,7 +185,7 @@ and this feature is only available for your root package dependencies.
 * **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader generation skips the `autoload-dev` rules.
 * **--no-install:** Does not run the install step after updating the composer.lock file.
 * **--no-audit:** Does not run the audit steps after updating the composer.lock file.
-* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
 * **--lock:** Only updates the lock file hash to suppress warning about the
   lock file being out of date.
 * **--with:** Temporary version constraint to add, e.g. foo/bar:1.0.0 or foo/bar=1.0.0
@@ -258,7 +258,7 @@ If you do not specify a package, Composer will prompt you to search for a packag
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
 * **--no-audit:** Does not run the audit steps after updating the composer.lock file.
-* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
 * **--update-no-dev:** Run the dependency update with the `--no-dev` option.
 * **--update-with-dependencies (-w):** Also update dependencies of the newly required packages, except those that are root requirements.
 * **--update-with-all-dependencies (-W):** Also update dependencies of the newly required packages, including those that are root requirements.
@@ -302,7 +302,7 @@ uninstalled.
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
 * **--no-audit:** Does not run the audit steps after installation is complete.
-* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
 * **--update-no-dev:** Run the dependency update with the --no-dev option.
 * **--update-with-dependencies (-w):** Also update dependencies of the removed packages.
   (Deprecated, is now default behavior)
@@ -859,7 +859,7 @@ By default the command checks for the packages on packagist.org.
 * **--remove-vcs:** Force-remove the VCS metadata without prompting.
 * **--no-install:** Disables installation of the vendors.
 * **--no-audit:** Does not run the audit steps after installation is complete.
-* **--audit-format:** Audit output format. Must be "table" (default) or "plain".
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
 * **--ignore-platform-reqs:** ignore all platform requirements (`php`, `hhvm`,
   `lib-*` and `ext-*`) and force the installation even if the local machine does
   not fulfill these.
@@ -979,7 +979,7 @@ php composer.phar audit
 ### Options
 
 * **--no-dev:** Disables auditing of require-dev packages.
-* **--format (-f):** Audit output format. Must be "table" (default) or "plain".
+* **--format (-f):** Audit output format. Must be "table" (default), "plain", or "summary".
 
 ## help
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -185,7 +185,7 @@ and this feature is only available for your root package dependencies.
 * **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader generation skips the `autoload-dev` rules.
 * **--no-install:** Does not run the install step after updating the composer.lock file.
 * **--no-audit:** Does not run the audit steps after updating the composer.lock file.
-* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
 * **--lock:** Only updates the lock file hash to suppress warning about the
   lock file being out of date.
 * **--with:** Temporary version constraint to add, e.g. foo/bar:1.0.0 or foo/bar=1.0.0
@@ -258,7 +258,7 @@ If you do not specify a package, Composer will prompt you to search for a packag
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
 * **--no-audit:** Does not run the audit steps after updating the composer.lock file.
-* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
 * **--update-no-dev:** Run the dependency update with the `--no-dev` option.
 * **--update-with-dependencies (-w):** Also update dependencies of the newly required packages, except those that are root requirements.
 * **--update-with-all-dependencies (-W):** Also update dependencies of the newly required packages, including those that are root requirements.
@@ -302,7 +302,7 @@ uninstalled.
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
 * **--no-audit:** Does not run the audit steps after installation is complete.
-* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
 * **--update-no-dev:** Run the dependency update with the --no-dev option.
 * **--update-with-dependencies (-w):** Also update dependencies of the removed packages.
   (Deprecated, is now default behavior)
@@ -859,7 +859,7 @@ By default the command checks for the packages on packagist.org.
 * **--remove-vcs:** Force-remove the VCS metadata without prompting.
 * **--no-install:** Disables installation of the vendors.
 * **--no-audit:** Does not run the audit steps after installation is complete.
-* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
 * **--ignore-platform-reqs:** ignore all platform requirements (`php`, `hhvm`,
   `lib-*` and `ext-*`) and force the installation even if the local machine does
   not fulfill these.
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -970,7 +970,7 @@ php composer.phar archive vendor/package 2.0.21 --format=zip
 This command is used to audit the packages in your composer.lock
 for possible security issues. Currently this only checks for and
 lists security vulnerability advisories according to the
-[packagist api](https://packagist.org/apidoc#list-security-advisories).
+[Packagist.org api](https://packagist.org/apidoc#list-security-advisories).
 
 ```sh
 php composer.phar audit
--- a/src/Composer/Command/AuditCommand.php
+++ b/src/Composer/Command/AuditCommand.php
@@ -18,7 +18,7 @@ class AuditCommand extends BaseCommand
             ->setDescription('Checks for security vulnerability advisories for packages in your composer.lock.')
             ->setDefinition(array(
                 new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables auditing of require-dev packages.'),
-                new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Output format. Must be "table" or "plain".', Auditor::FORMAT_TABLE),
+                new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Output format. Must be "table", "plain", or "summary".', Auditor::FORMAT_TABLE),
             ))
             ->setHelp(
                 <<<EOT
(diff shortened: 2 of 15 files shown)


stof:
> I also got the 502 error when trying the query in the browser. This seems to be caused by too many packages in a single request; if I significantly reduce the number of packages in the query, it returns an appropriate response. This is a major blocker for this functionality being implemented, so if someone knows a way around this, please let me know.

Well, one thing you could do is splitting the list of packages into batches (the batch size should be chosen based on the supported size for the Packagist API) and send multiple requests. Thanks to the parallelization support in Composer 2+, you can even send those requests in parallel.

GuySartorelli:
That's a good idea - I couldn't find any information about what the limit actually is but I'll have another look, and failing that I can always do some trial and error to find it. Bonus points for anyone who can point me in the right direction or who already knows what the limit is.

Seldaek:
Yes IMO putting the package list in the POST body will fix this as this is AFAIK not size limited at all, unlike URL length which has limits in various places.

GuySartorelli:
Currently blocked because the packagist API returns a 404 error on POST requests. I've created an issue for it under composer/packagist#1309

GuySartorelli:
The basic functionality is working now - any review on the `AuditCommand` or `Auditor` and related documentation would be useful at this stage.

GuySartorelli:
It looks like there's some PHPStan issues as well - to be honest I don't understand what I need to do to resolve those. A lot of the errors are saying the new command doesn't define some options... but it doesn't need those options.

GuySartorelli:
I've resolved some of the PHPStan issues. The ones that I think are false alarms I've tried to add to the baseline, but it doesn't seem to have taken. There are a couple of PHPStan issues there as well which are unrelated to this PR.

GuySartorelli:
This is now ready to be reviewed.

Seldaek:
Thanks - sorry I've had a rough few weeks here so lagging behind on many things and it may take me a bit longer to get to this, but regarding phpstan if you're still having issues just ignore them that's no biggie for now.

src/Composer/Command/AuditCommand.php
```diff
@@ -0,0 +1,56 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Composer\Factory;
+use Composer\Util\Auditor;
+use Composer\Util\Filesystem;
+use Symfony\Component\Console\Input\InputOption;
+
+class AuditCommand extends BaseCommand
+{
+    protected function configure()
```
herndlm, line 14: I saw other commands use `void` as return type here, that should get rid of one of the phpstan errors

GuySartorelli:
@Seldaek No worries, there's no rush. 

@herndlm interesting! I didn't think I could add a return type there because it wouldn't match the signature of the symfony command class's configure method - but you're right, other commands here have added it. I'll make that change. 
https://github.com/symfony/console/blob/c9646197ef43b0e2ff44af61e7f0571526fd4170/Command/Command.php#L198

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,252 @@
...
+        $url = self::API_URL;
+        if ($updatedSince !== null) {
+            $url .= "?updatedSince=$updatedSince";
+        }
+
+        // Get advisories from API
+        $response = $this->httpDownloader->get($url, $this->createPostOptions($packages));
+        $advisories = $response->decodeJson()['advisories'];
+
+        if (count($packages) > 0 && $filterByVersion) {
```
GuySartorelli, line 111:
```suggestion
        if (count(advisories) > 0 && $filterByVersion) {
```
Oops. I put the wrong variable here. I'll make this change once people have been given a chance to review the or as a whole.

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,252 @@
...
+            foreach ($packageAdvisories as $advisory) {
+                $io->getTable()
+                    ->setHorizontal()
+                    ->setHeaders([
+                        '<fg=red>Package</>',
+                        '<fg=red>CVE</>',
+                        '<fg=red>Title</>',
+                        '<fg=red>URL</>',
+                        '<fg=red>Affected versions</>',
+                        '<fg=red>Reported at</>',
```
Seldaek, line 209: I think I'd remove the red text hre and in the plain output. It doesn't seem to add a lot of value and I'd argue it even highlights the wrong thing as the headings aren't the most important thing here.

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,252 @@
...
+    private function outputAdvisories(IOInterface $io, array $advisories): void
+    {
+        if ($this->format === self::FORMAT_TABLE) {
+            if (!($io instanceof ConsoleIO)) {
+                throw new InvalidArgumentException('Cannot use table format with ' . get_class($io));
+            }
+            $this->outputAvisoriesTable($io, $advisories);
+        } else {
+            $this->outputAdvisoriesPlain($io, $advisories);
+        }
```
Seldaek, line 189: I'd add a `summary` format which only outputs `there are X installed packages with known vulnerabilities, run composer audit to see them`. I suspect we may not want to spam with the full output on every update/install.

src/Composer/Installer.php
```diff
@@ -381,6 +386,16 @@ public function run(): int
             gc_enable();
         }
 
+        if ($this->audit) {
+            $packages = $this->locker->getLockedRepository($this->devMode)->getPackages();
```
Seldaek, line 390: Rather use `$localRepo->getCanonicalPackages()` here I think, so it checks all installed packages, not what's in the lock file, although at this point it should be the same, it's semantically more accurate IMO.

src/Composer/Command/AuditCommand.php
```diff
@@ -0,0 +1,56 @@
...
+        $lockFile = Factory::getLockFile(Factory::getComposerFile());
+        if (!Filesystem::isReadable($lockFile)) {
+            $this->getIO()->writeError('<error>' . $lockFile . ' is not readable.</error>');
+            return 1;
+        }
+
+        $composer = $this->requireComposer($input->getOption('no-plugins'), $input->getOption('no-scripts'));
+        $locker = $composer->getLocker();
+        $packages = $locker->getLockedRepository(!$input->getOption('no-dev'))->getPackages();
+        $httpDownloader = Factory::createHttpDownloader($this->getIO(), $composer->getConfig());
```
Seldaek, line 46:
```suggestion
        $httpDownloader = $composer->getLoop()->getHttpDownloader();
```

src/Composer/Command/AuditCommand.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $lockFile = Factory::getLockFile(Factory::getComposerFile());
+        if (!Filesystem::isReadable($lockFile)) {
+            $this->getIO()->writeError('<error>' . $lockFile . ' is not readable.</error>');
+            return 1;
+        }
+
+        $composer = $this->requireComposer($input->getOption('no-plugins'), $input->getOption('no-scripts'));
+        $locker = $composer->getLocker();
+        $packages = $locker->getLockedRepository(!$input->getOption('no-dev'))->getPackages();
```
Seldaek, line 45: It'd be preferable/more consistent IMO to do something like https://github.com/composer/composer/blob/main/src/Composer/Command/ShowCommand.php#L234-L259 with a --locked flag to optionally read from lock file, but by default it should read from the local (installed) repository.

Seldaek (changes requested):
Thanks, overall it look good I think! Lots of comments but a lot is for minor code tweaks.

src/Composer/Util/Auditor.php
Seldaek: @GuySartorelli please append new commits for new changes, it's much easier to do further reviews. I can squash when merging if it turns into a huge mess.

src/Composer/Command/AuditCommand.php
```diff
@@ -0,0 +1,56 @@
...
+class AuditCommand extends BaseCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('audit')
+            ->setDescription('Checks for security vulnerability advisories for packages in your composer.lock.')
+            ->setDefinition(array(
+                new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables auditing of require-dev packages.'),
+                new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Output format. Must be "table" or "plain".', Auditor::FORMAT_TABLE),
```
GromNaN, line 21: Other commands use `text` instead of `plain`.

You should use `Composer\Console\Input\InputOption` to suggest values in bash completion. See the LicenseCommand if you need an example.

GuySartorelli: "table" and "plain" are both text formats - In `LicencesCommand` I take "text" to be as opposed to "json" i.e. it's not just a layout format difference, it's a difference between plain text and a JSON formatting of data.
I can change "plain" to "text" if there is a push for that, but I think in this situation "plain" probably makes more sense.

I will make the change to aid bash completion though.

src/Composer/Util/Auditor.php
GuySartorelli: Good idea.
1. Should "summary" be the default format for all commands other than `audit` itself?
2. Should the "summary" format also be available from `composer audit` itself? It might be a bit odd to run `composer audit --format=summary` and have it tell you to run `composer audit` - but at the same time, if someone intentionally selects the summary format maybe that is expected?

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,252 @@
...
+            'http' => [
+                'method' => 'POST',
+                'header' => ['Content-type: application/x-www-form-urlencoded'],
+                'timeout' => 3,
+            ],
+        ];
+        if (count($packages) > 0) {
+            $content = ['packages' => []];
+            foreach ($packages as $package) {
+                $content['packages'][] = $package->getName();
```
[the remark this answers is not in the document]
GuySartorelli: That sounds like a good idea - though I don't know what that would look like as far as how that information is passed to the API...
In either case I don't think that's critical to this functionality. If the packagist API gets updated to accept version numbers I'll be happy to update this, but that can always be done as an enhancement later on as well.

src/Composer/Installer.php
```diff
@@ -381,6 +386,16 @@ public function run(): int
...
+            if (count($packages) > 0) {
+                $auditor = new Auditor(Factory::createHttpDownloader($this->io, $this->config), $this->auditFormat);
+                $auditor->audit($this->io, $packages);
+            } else {
+                $this->io->writeError('No packages - skipping audit.');
```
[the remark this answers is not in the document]
GuySartorelli: I added this because users may be expecting some audit message (either "no issues" or "found x issues") - but if there are no packages, there will be no audit message. That may seem like something went wrong rather than an intentional lack of messaging.

I'm happy to remove this if with this context you still feel it's not needed.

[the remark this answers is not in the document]
Seldaek: Ok let's leave it in for now, we'll see with day to day usage if it gets too noisy.

src/Composer/Util/Auditor.php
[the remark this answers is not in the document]
Seldaek: Absolutely, feel free to keep it as is for now. I'll take a look at the packagist side whenever.

src/Composer/Util/Auditor.php
Seldaek: 1. I'd say summary should be default yes, except audit cmd probably using table as default?
2. Yeah not such a big deal :)

src/Composer/Util/Auditor.php
GuySartorelli: Will do

src/Composer/Command/CreateProjectCommand.php
```diff
@@ -88,6 +89,8 @@ protected function configure(): void
                 new InputOption('keep-vcs', null, InputOption::VALUE_NONE, 'Whether to prevent deleting the vcs folder.'),
                 new InputOption('remove-vcs', null, InputOption::VALUE_NONE, 'Whether to force deletion of the vcs folder without prompting.'),
                 new InputOption('no-install', null, InputOption::VALUE_NONE, 'Whether to skip installation of the package dependencies.'),
+                new InputOption('no-audit', null, InputOption::VALUE_NONE, 'Whether to skip auditing of the installed package dependencies.'),
+                new InputOption('audit-format', null, InputOption::VALUE_OPTIONAL, 'Audit output format. Must be "table" or "plain".', Auditor::FORMAT_TABLE),
```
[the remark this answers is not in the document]
GuySartorelli: I'll add a new const array to hold all formats on the `Auditor` class and pass that in here instead - that way any changes to the formats just need to be updated in that one place.

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,252 @@
...
+
+    /**
+     * Set the format that will be used to output audit results.
+     *
+     * @param string $format
+     * @return self
+     */
+    public function setFormat(string $format): self
+    {
+        if (!in_array($format, [self::FORMAT_TABLE, self::FORMAT_PLAIN])) {
```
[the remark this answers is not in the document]
GuySartorelli: I'll make this change, but I don't think in this case it makes a difference.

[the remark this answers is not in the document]
Seldaek: It doesn't, but our phpstan config enforces strict mode to avoid surprises.

src/Composer/Util/Auditor.php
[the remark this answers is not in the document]
glaubinix: I wonder whether instead of or in addition to sending versions, storing the response in the Composer cache and sending a `if-modified since` header wouldn't be a better option? Advisories do not change that often and on CI you will run several composer installs that could reuse the cache file.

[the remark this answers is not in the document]
GuySartorelli: I don't know enough about cache to even begin implementing that - especially since it's not only the advisories that might change but the installed packages as well. I'm happy for someone else to implement that improvement after this PR has been merged, but it is not something I feel confident tackling.

src/Composer/Command/AuditCommand.php
GuySartorelli: Looks that that uses a method [`filterRequiredPackages()`](https://github.com/composer/composer/blob/e9b44b449945dd7ac614c759240bb25f0617a38d/src/Composer/Command/ShowCommand.php#L1418) to filter out the dev dependencies, which is remarkably similar to [`LicensesCommand::filterRequiredPackages()`](https://github.com/composer/composer/blob/e9b44b449945dd7ac614c759240bb25f0617a38d/src/Composer/Command/LicensesCommand.php#L163).

I really don't want to add a _third_ implementation of this method, but I'm not sure where the method should exist so that it can be shared across commands. Do you have any preference? My current thinking is to just put it into `BaseCommand`.

GuySartorelli: I've duplicated it for now, just so that the functionality is available in `AuditCommand` - but it's obviously not ideal having such unDRY code.

GuySartorelli:
Requested changes have been made.

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,268 @@
...
+    private $httpDownloader;
+
+    /** @var self::FORMAT_* */
+    private $format;
+
+    /**
+     * @param HttpDownloader $httpDownloader
+     * @param self::FORMAT_* $format The format that will be used to output audit results.
+     */
+    public function __construct(HttpDownloader $httpDownloader, string $format = self::FORMAT_TABLE)
```
GuySartorelli, line 37: Just giving this another once-over... I wonder if I should remove the format parameter from the constructor? It wouldn't make sense for someone who is using this class for its `getAdvisories()` method to need to think at all about the format of `audit()` output... 🤔

Seldaek: Yes having that param in audit() probably makes more sense :)

GuySartorelli: Sweet. I've made that change.

ktomk:
For the output format could JSON Text be a other variant? I've seen there is a table, so perhaps some very straight forward JSON Array of JSON Objects could be nice to pass along to further (automated) processing.

GuySartorelli:
@ktomk At this stage I'd just be keen to see the feature merged - that sounds like an enhancement that can be added after the fact quite easily if there is real world use for it. I'd be quite happy for someone to create a follow-up PR with that format if there is a need for it but it's not something I want to add to this PR right now.

doc/03-cli.md
```diff
@@ -182,6 +184,8 @@ and this feature is only available for your root package dependencies.
 * **--dev:** Install packages listed in `require-dev` (this is the default behavior).
 * **--no-dev:** Skip installing packages listed in `require-dev`. The autoloader generation skips the `autoload-dev` rules.
 * **--no-install:** Does not run the install step after updating the composer.lock file.
+* **--no-audit:** Does not run the audit steps after updating the composer.lock file.
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
```
Seldaek, line 188:
```suggestion
* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
```

doc/03-cli.md
```diff
@@ -253,6 +257,8 @@ If you do not specify a package, Composer will prompt you to search for a packag
   terminals or scripts which don't handle backspace characters.
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
+* **--no-audit:** Does not run the audit steps after updating the composer.lock file.
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
```
Seldaek, line 261:
```suggestion
* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
```

doc/03-cli.md
```diff
@@ -295,6 +301,8 @@ uninstalled.
   terminals or scripts which don't handle backspace characters.
 * **--no-update:** Disables the automatic update of the dependencies (implies --no-install).
 * **--no-install:** Does not run the install step after updating the composer.lock file.
+* **--no-audit:** Does not run the audit steps after installation is complete.
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
```
Seldaek, line 305:
```suggestion
* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
```

doc/03-cli.md
```diff
@@ -850,6 +858,8 @@ By default the command checks for the packages on packagist.org.
   mode.
 * **--remove-vcs:** Force-remove the VCS metadata without prompting.
 * **--no-install:** Disables installation of the vendors.
+* **--no-audit:** Does not run the audit steps after installation is complete.
+* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default)..
```
Seldaek, line 862:
```suggestion
* **--audit-format:** Audit output format. Must be "table", "plain", or "summary" (default).
```

src/Composer/Installer.php
```diff
@@ -381,6 +387,23 @@ public function run(): int
...
         }
 
+        if ($this->audit) {
+            $packages = $localRepo->getCanonicalPackages();
+            if (count($packages) > 0) {
+                try {
+                    $auditor = new Auditor(Factory::createHttpDownloader($this->io, $this->config));
+                    $auditor->audit($this->io, $packages, $this->auditFormat);
+                } catch (TransportException $e) {
+                    $this->io->error('Error occurred while auditing locked packages.');
```
Seldaek, line 397:
```suggestion
                    $this->io->error('Failed to audit installed packages.');
```

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,242 @@
...
+
+use Composer\IO\ConsoleIO;
+use Composer\IO\IOInterface;
+use Composer\Package\PackageInterface;
+use Composer\Semver\Semver;
+use InvalidArgumentException;
+
+class Auditor
+{
+    public const API_URL = 'https://packagist.org/api/security-advisories/';
```
Seldaek, line 13:
```suggestion
    private const API_URL = 'https://packagist.org/api/security-advisories/';
```

src/Composer/Command/AuditCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        return $auditor->audit($this->getIO(), $packages, $input->getOption('format'), false);
+    }
+
+    /**
+     * @param InputInterface $input
+     * @return PackageInterface[]
+     */
+    private function getPackages(InputInterface $input): array
+    {
+        $composer = $this->requireComposer($input->getOption('no-plugins'), $input->getOption('no-scripts'));
```
Seldaek, line 58: No need to forward those they are used as default value already for conciseness

```suggestion
        $composer = $this->requireComposer();
```

src/Composer/Command/AuditCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        if ($input->getOption('locked')) {
+            if (!$composer->getLocker()->isLocked()) {
+                throw new \UnexpectedValueException('Valid composer.json and composer.lock files is required to run this command with --locked');
```
Seldaek, line 61:
```suggestion
                throw new \UnexpectedValueException('Valid composer.json and composer.lock files are required to run this command with --locked');
```

Seldaek (approved):
Thanks, looking good I think now, a couple typos/nitpicks and then it's mergeable I think.

src/Composer/Command/AuditCommand.php
```diff
@@ -0,0 +1,102 @@
...
+class AuditCommand extends BaseCommand
+{
+    protected function configure(): void
+    {
+        $this
+            ->setName('audit')
+            ->setDescription('Checks for security vulnerability advisories for installed packages.')
+            ->setDefinition(array(
+                new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables auditing of require-dev packages.'),
+                new InputOption('format', 'f', InputOption::VALUE_OPTIONAL, 'Output format. Must be "table", "plain", or "summary".', Auditor::FORMAT_TABLE),
```
Seldaek, line 22:
```suggestion
                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Output format. Must be "table", "plain", or "summary".', Auditor::FORMAT_TABLE),
```

src/Composer/Util/Auditor.php
```diff
@@ -0,0 +1,242 @@
...
+
+namespace Composer\Util;
+
+use Composer\IO\ConsoleIO;
+use Composer\IO\IOInterface;
+use Composer\Package\PackageInterface;
+use Composer\Semver\Semver;
+use InvalidArgumentException;
+
+class Auditor
```
Seldaek, line 11:
```suggestion
/**
 * @internal
 */
class Auditor
```

src/Composer/Command/AuditCommand.php
Seldaek: Yeah this is fine, I made a note to clean this up later :)

src/Composer/Util/Auditor.php
GuySartorelli: I don't see any reason to mark this as internal. Part of the reason I wanted to do this was so that this functionality _can_ be used in other projects. It's also why the `getAdvisories()` method is public and why it includes functionality not used but `AuditCommand`.

Seldaek: Yeah I am a bit overly paranoid about promising BC these days because a v3 is not around the corner..

If the point is reuse I'd argue we should later (once stabilized) publish this as a standalone lib because people depending on the whole of composer/composer is usually a bad idea..

GuySartorelli: I'm okay with that approach.

GuySartorelli:
Requested changes have been made.

GuySartorelli:
Tests fixed

Seldaek:
Thanks! Merged as d93239ddd9e6c18a2f7db3cdf94213dd892fa950 and fixed up some remaining issues in 611b21589 🥳

Seldaek:
The more I think of it the more I think we can't really extract this, as it'll probably need to be integrated deeper into the Repository code than it is now. 

I think the better way would be to publish a packagist.org API client library really, that would be completely separate from Composer, and would not require Package objects or anything to function.

GuySartorelli:
That's fair. I'd be happy to do development work on such a library with as much or as little guidance as you want to provide. While I could implement such a library in some arbitrary repo it makes sense to me that there would be such a repository under the composer organisation which people can just add to their dependencies when and as they need it.

stof:
You mean something like https://packagist.org/packages/knplabs/packagist-api ?

Seldaek:
Ah yes there's that, then maybe worth sending a PR there porting the Auditor querying bits @GuySartorelli if you feel like it?

GuySartorelli:
I'll take a look - I wasn't aware of that package. Still doesn't feel as appropriate as having one in GitHub.com/composer but I won't push for that if the appetite for creating it isnt here 😅

barryvdh:
> For the output format could JSON Text be another variant? I've seen there is a table, so perhaps some very straight forward JSON Array of JSON Objects could be nice to pass along to further (automated) processing.

I created a follow-up PR for the json format here: https://github.com/composer/composer/pull/10965

PR #10810, opened 2022-05-31
docs: Fix inconsist semver operator suggestion

jtojnar:
Using caret over tilde is better since it behaves the same as in npm: https://jubianchi.github.io/semver-check/#/constraint/~3.0
But when this change was introduced in https://github.com/composer/composer/pull/5396, it was not complete.


stof (approved):

Seldaek:
Thanks

PR #10811, opened 2022-05-31
Fix JSON schema regex pattern

### LINKED ISSUE #10805: Possibly invalid regex in composer-schema.json
Since commit https://github.com/composer/composer/commit/44a52e4157eb3a8c529204d6799ae7e0909d4de0, Visual Studio is having trouble validating `composer.json`'s JSON schema. While PHP seems to have no issues compiling the regular expression, javascript sees it as invalid.

Screenshot 2022-05-28 at 07 10 48

https://github.com/composer/composer/blob/44a52e4157eb3a8c529204d6799ae7e0909d4de0/res/composer-schema.json#L6-L10

Compiling the regex with PHP works and seems to validate as expected : 
- https://3v4l.org/kTnon
- https://regex101.com/r/vBTWRa/1

Compiling the regex with JS produces an error: 
- https://jsfiddle.net/3fjr9ny2/
- https://regex101.com/r/LYG1VC/1

Visual Studio Code uses the following schema for validating the `composer.json` file:

```
https://raw.githubusercontent.com/composer/composer/master/res/composer-schema.json
```

**Edit:**

Some [additional research](https://stackoverflow.com/a/25255810) suggests that JS does not support possessive quantifiers, as used in this pattern. This could probably be fixed by using the initial pattern only in the JSON schema.

nhedger:
This PR is an attempt to fix #10805 by removing all possessive quantifiers from the regex pattern.

https://jsfiddle.net/wj345vsa/

I'm sending this to 2.2 since it's the oldest branch the issue applies to.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -6,7 +6,7 @@
         "name": {
             "type": "string",
             "description": "Package name, including 'vendor-name/' prefix.",
-            "pattern": "^[a-z0-9]([_.-]?[a-z0-9]++)*+/[a-z0-9](([_.]|-{1,2})?[a-z0-9]++)*+$"
+            "pattern": "^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*$"
         },
         "description": {
             "type": "string",
--- a/tests/Composer/Test/Json/ComposerSchemaTest.php
+++ b/tests/Composer/Test/Json/ComposerSchemaTest.php
@@ -25,9 +25,9 @@ class ComposerSchemaTest extends TestCase
         $expectedError = array(
             array(
                 'property' => 'name',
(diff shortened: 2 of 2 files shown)


radar3301:
It's actually [this diff](https://github.com/composer/composer/commit/d70b58089960991978058669c2a581afc026af53) that needs modified/reverted.

[res/composer-schema.json](https://github.com/composer/composer/commit/d70b58089960991978058669c2a581afc026af53#diff-f19353690a3cee8d579021b46e47b536cd6925ebaad0918bc343d7c1019a4661)
[src/Composer/Package/Loader/ValidatingArrayLoader.php](https://github.com/composer/composer/commit/d70b58089960991978058669c2a581afc026af53#diff-4bb44830e256528ccfded1222e657f57986634a6afd3175814bb234caa817fe5)
[tests/Composer/Test/Json/ComposerSchemaTest.php](https://github.com/composer/composer/commit/d70b58089960991978058669c2a581afc026af53#diff-95f3425265c27c65fcde07e8ce7baf1bd34ecffa950500aac57990c5717b16b4)

Seldaek:
Thanks!

PR #10815, opened 2022-06-02
Specify optional array keys in `ConfigReturnTypeExtension`

herndlm:
Refs: https://github.com/composer/composer/pull/10635#discussion_r887745061

The value types of the optional props are already unioned with a NullType though. Just wanted to mention this again because I'm not a 100% sure about the schema.

Looks like constant arrays are currently only created for 3 paths and only one of them has 2 optional keys.
- `gitlab-token.additionalProperties`
- `http-basic.additionalProperties`
- `bitbucket-oauth.additionalProperties`: `access-token` and `access-token-expiration` are optional


Seldaek:
I think this is correct, thanks!

Seldaek:
Merged as 3b204cb6f in 2.3

Issue #10796, opened 2022-05-24
[Help Wanted] Increasing test coverage of Command classes


Seldaek:
As you can see on this image, Command classes have very poor coverage currently (16% of lines covered in total):

<img src="https://user-images.githubusercontent.com/183678/169980310-62da227b-44ff-4a45-988f-054970436f51.png" width=200>

**Edit: To see latest guidelines on what to work on see https://github.com/composer/composer/issues/10796#issuecomment-1277340969**

Here some examples of how to write sane integration tests for most commands: 

https://github.com/composer/composer/blob/176d25851d1f99345c652a6ecbc7c3787071218d/tests/Composer/Test/Command/RunScriptCommandTest.php#L92-L113

https://github.com/composer/composer/blob/176d25851d1f99345c652a6ecbc7c3787071218d/tests/Composer/Test/Command/ConfigCommandTest.php#L23-L52

https://github.com/composer/composer/blob/70f2dd6eddd5916f70b3d73dd3170a3d2d3a7681/tests/Composer/Test/Command/BumpCommandTest.php#L28-L47

If you feel like helping, PRs are very much welcome, I don't think we absolutely need 100% coverage but having the most common use cases covered would already be very valuable. Please target the `main` branch if you want to contribute some, and feel free to announce it here if you start some larger chunk of work to avoid duplicating efforts.

theoboldalex:
@Seldaek I'd love to take on some of this work. I have opened a first PR here #10932 and will start to work my way through the commands.

Seldaek:
@theoboldalex great, thanks! Maybe wait until I get a chance to review the first one to make sure you don't run off and do more of them with the wrong assumptions. I'll try to get to it soon.

Also note the listing above is a bit outdated as I tried to work on this here and there when touching commands the last few months. But there is still plenty to be done :)

ralflang:
No promises but I may make the occassional PR every now and then.
Might be a good learning exercise. I have been playing around with the Command, Plugin, Capable ... parts of the codebase last month.

Seldaek:
Update, we're at 36% (up from 16!) already now:

<img width="400" alt="image" src="https://user-images.githubusercontent.com/183678/185733135-3854c0b4-dd49-435f-ac6e-626655a36c82.png">

tomekpryjma:
I'd like to help out with this now and again - is there a way to see the current coverage?

Seldaek:
Now at 41%! Here are the classes still needing the most effort IMO:

[...]

PR #10817, opened 2022-06-03
Fix processing null config value in DiagnoseCommand

### LINKED ISSUE #10814: Error when calling composer diagnose
My `composer.json`:

irrelevant (error shows even without composer.json)

```
   Checking git settings: OK

   In DiagnoseCommand.php line 117:

     [TypeError]
     count(): Argument #1 ($value) must be of type Countable|array, null given


   Exception trace:
     at phar:///usr/bin/composer/src/Composer/Command/DiagnoseCommand.php:117
    count() at phar:///usr/bin/composer/src/Composer/Command/DiagnoseCommand.php:117
    Composer\Command\DiagnoseCommand->execute() at phar:///usr/bin/composer/vendor/symfony/console/Command/Command.php:298
    Symfony\Component\Console\Command\Command->run() at phar:///usr/bin/composer/vendor/symfony/console/Application.php:1024
    Symfony\Component\Console\Application->doRunCommand() at phar:///usr/bin/composer/vendor/symfony/console/Application.php:299
    Symfony\Component\Console\Application->doRun() at phar:///usr/bin/composer/src/Composer/Console/Application.php:335
    Composer\Console\Application->doRun() at phar:///usr/bin/composer/vendor/symfony/console/Application.php:171
    Symfony\Component\Console\Application->run() at phar:///usr/bin/composer/src/Composer/Console/Application.php:130
    Composer\Console\Application->run() at phar:///usr/bin/composer/bin/composer:88
    require() at /usr/bin/composer:29

```

PHP-Version is 8.1.6, errors occurs on x86_64 and arm64.

b1rdex:
Fixes #10814.

2.2 isn't affected, the issue was introduced in https://github.com/composer/composer/commit/bd6403a6bef36fdbefdb287dc7e26286d5e06e5e#diff-608f3426a5aefbff3d9da84c38d75164db806317f92e4a987531915847852380

IMO, these options should default to an empty array.
A possible solution could be to introduce a future BC, as it's done with `allow-plugins`.

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -76,13 +76,12 @@ class Config
         'use-github-api' => true,
         'lock' => true,
         'platform-check' => 'php-only',
-        // valid keys without defaults (auth config stuff):
-        // bitbucket-oauth
-        // github-oauth
-        // gitlab-oauth
-        // gitlab-token
-        // http-basic
-        // bearer
+        'bitbucket-oauth' => array(),
+        'github-oauth' => array(),
+        'gitlab-oauth' => array(),
+        'gitlab-token' => array(),
+        'http-basic' => array(),
+        'bearer' => array(),
     );
 
     /** @var array<string, mixed> */
--- a/src/Composer/IO/BaseIO.php
+++ b/src/Composer/IO/BaseIO.php
@@ -114,12 +114,12 @@ abstract class BaseIO implements IOInterface
      */
     public function loadConfiguration(Config $config)
     {
-        $bitbucketOauth = $config->get('bitbucket-oauth') ?: array();
-        $githubOauth = $config->get('github-oauth') ?: array();
-        $gitlabOauth = $config->get('gitlab-oauth') ?: array();
-        $gitlabToken = $config->get('gitlab-token') ?: array();
-        $httpBasic = $config->get('http-basic') ?: array();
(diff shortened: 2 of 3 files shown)


mvorisek:
`composer diagnose` should have a functional test

Seldaek:
@mvorisek feel free to help, see https://github.com/composer/composer/issues/10796 - just stating the obvious here isn't helping anyone.

Seldaek:
@b1rdex i think it would make more sense to change the default for these keys to an empty array. This shouldn't cause any issues as an empty array is already a possible value. Then we keep the type simpler instead of adding some nullable types.

mvorisek:
> @mvorisek feel free to help, see #10796 - just stating the obvious here isn't helping anyone.

this issue has broken our Docker images CI, so I just pointed out this command is (more) important (than others) to be tested

Seldaek:
Ah yes, because your CI is more important than others. Good to know.

b1rdex:
> @b1rdex i think it would make more sense to change the default for these keys to an empty array. This shouldn't cause any issues as an empty array is already a possible value. Then we keep the type simpler instead of adding some nullable types.

Done.

alcohol (approved):

Seldaek:
Thanks! Merged as a0b789623 in 2.3 branch.

Issue #10820, opened 2022-06-05
PHP-Warning due to redundant stream_wrapper_register-calls in php-proxy-binaries, when using chained inclusions


brandung-sjorek:
Hello,

I have two independently installed projects, let's say `vendor-a/foo-project` and `vendor-b/bar-project`. Both projects are installed independently in directories which do not intersect and both projects have a composer-generated php-proxy-binary installed in their `{$bin-dir}` (`vendor/bin`). The binary from `a-vendor/foo-project` shall `include(…);` the binary from `vendor-b/bar-project`. _(This is weird and it looks a bit esoteric, at least at the first glance. I know this, but I have reasons. 😞)_ Nevertheless, this strategy combined with the lately introduced `phpvfscomposer` stream-wrapper, which was embedded in these proxy-binaries to avoid problems with the shebang of included php-scripts for PHP < 8, results in warnings for composer version ≥ 2.2 in combination with PHP version < 8.

My `composer.json`:

```json
{
    …
   "bin": [
        "foo",
        "bar"
    ],
    …
}
```

Output of `composer diagnose`:

```bash
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.3.6
PHP version: 7.4.29
PHP binary path: /opt/local/bin/php74
OpenSSL version: OpenSSL 1.1.1n  15 Mar 2022
cURL version: 7.83.1 libz 1.2.12 ssl OpenSSL/3.0.3
zip: extension present, unzip present, 7-Zip not available
```

An example: If php-proxy-binary `vendor/bin/foo` includes another php-proxy-binary `vendor/bin/bar`, and when I run this command:

```
❯ ./vendor/bin/foo
```

I get the following output:

```
PHP Warning:  stream_wrapper_register(): Protocol phpvfscomposer:// is already defined. in /…/vendor/bin/foo on line 110
PHP Stack trace:
PHP   1. {main}() /…/vendor/bin/foo:0
[...]

PR #10823, opened 2022-06-05
allow chained proxy-binary php-inclusions - solution (c)

brandung-sjorek:
This PR allows chained proxy-binary php-inclusions by skipping the registration of `phpvfscomposer` stream-wrapper, if a stream-wrapper with the same name ist already registered. For details, please see #10820 …


Seldaek:
Thanks

PR #10829, opened 2022-06-08
Add bump 👊🏻 command to bump requirements to the currently installed version

### LINKED ISSUE #7273: [RFC] New command to update minimum version in json to lock values
Currently I have a bash script (company internal) that syncs the values from the lock to the json. Say I have `^3.0.1` in my json, but I have actually installed `3.6.10`, running this script would sync the value from the lock file, into the json, making it `^3.6.10`. There's no point in composer having to check compatibility with versions lower than what I have installed whenever I update packages.

Would it be an idea to implement such a script to composer? If I'm correct, it should improve performance and memory usage, as the list of possible dependencies is smaller. Currently running an update in my symfony application doesn't even break 600MB usage with ~250 packages installed and all scripts running, where as others are reporting up to 1.5GB on memory usage.

Seldaek:
Fixes #7273

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -322,6 +322,24 @@ uninstalled.
 * **--apcu-autoloader-prefix:** Use a custom prefix for the APCu autoloader cache.
   Implicitly enables `--apcu-autoloader`.
 
+## bump
+
+The `bump` command increases the lower limit of your composer.json requirements
+to the currently installed versions. This helps to ensure your dependencies do not
+accidentally get downgraded due to some other conflict, and can slightly improve
+dependency resolution performance as it limits the amount of package versions
+Composer has to look at.
+
+Running this blindly on libraries is **NOT** recommended as it will narrow down
+your allowed dependencies, which may cause dependency hell for your users.
+Running it with `--dev-only` on libraries may be fine however as dev requirements
+are local to the library and do not affect consumers of the package.
+
+### Options
+
+* **--dev-only:** Only bump requirements in "require-dev".
+* **--no-dev-only:** Only bump requirements in "require".
+
 ## reinstall
 
 The `reinstall` command looks up installed packages by name,
--- a/src/Composer/Command/BumpCommand.php
+++ b/src/Composer/Command/BumpCommand.php
@@ -1,0 +1,227 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Composer.
+ *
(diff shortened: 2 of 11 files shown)


doc/03-cli.md
```diff
@@ -322,6 +322,23 @@ uninstalled.
...
+Composer has to look at.
+
+Running this blindly on libraries is **NOT** recommended as it will narrow down
+your allowed dependencies, which may cause dependency hell for your users.
+Running it with `--dev` on libraries is fine however as dev requirements are
+local to the library and do not affect consumers of the package.
+
+### Options
+
+* **--dev:** Only bump requirements in "require-dev".
```
naderman, line 340: Should this be `--dev-only` so one doesn't expect it to do both like our other --dev options?

doc/03-cli.md
```diff
@@ -322,6 +322,23 @@ uninstalled.
...
+
+The `bump` command increases the lower limit of your composer.json requirements
+to the currently installed versions. This helps to ensure your dependencies do not
+accidentally get downgraded due to some other conflict, and can slightly improve
+dependency resolution performance as it limits the amount of package versions
+Composer has to look at.
+
+Running this blindly on libraries is **NOT** recommended as it will narrow down
+your allowed dependencies, which may cause dependency hell for your users.
+Running it with `--dev` on libraries is fine however as dev requirements are
```
stof, line 335: note that it is not always fine either. Libraries generally support multiple PHP versions, so running `bump --dev` will generally be fine only if the current dependencies were installed for the lowest supported PHP version.

src/Composer/Command/BumpCommand.php
```diff
@@ -0,0 +1,226 @@
...
+        if (false === $contents) {
+            $io->writeError('<error>'.$composerJsonPath.' is not readable.</error>');
+
+            return self::ERROR_GENERIC;
+        }
+        unset($contents);
+
+        // check for writability by writing to the file as is_writable can not be trusted on network-mounts
+        // see https://github.com/composer/composer/issues/8231 and https://bugs.php.net/bug.php?id=68926
+        if (!is_writable($composerJsonPath) && false === Silencer::call('file_put_contents', $composerJsonPath, $composerJson)) {
```
stof, line 104: shouldn't this use `$contents` here rather than `$composerJson` ?

Seldaek: Ah yes, renamed vars around a couple times and this got lost, thanks.

PR #10836, opened 2022-06-10
Fix exported path with space in binary installer proxy

cjreed121:
I noticed an issue with a composer vendor binary. If the current working directory path contains a space, upon running a composer vendor binary the following error is thrown:
`./vendor/bin/sail: 26: export: PART_OF_PATH/vendor/bin: bad variable name`
PART_OF_PATH would be `website` if working directory path was `/home/user/my website` for example

Identified issue and tested fix on Ubuntu 20.04. The package/binary I was using: https://github.com/laravel/sail

This PR just quotes the export value in the shell script to allow for paths with spaces.

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -437,7 +437,7 @@ if [ -d /proc/cygdrive ]; then
     esac
 fi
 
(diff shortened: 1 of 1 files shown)


stof (approved):

Seldaek:
Thanks! This stuff is so easy to miss.. I probably should rename my local clone to `comp oser` or something :D

PR #10841, opened 2022-06-12
`composer bump` - Fix typo in the warning message

Ayesh:
Change warning text 'Alternatively you can use --dev-only to only bump dependencies within "require-dev"' that said `--dev` instead of `--dev-only`.


Seldaek:
Thanks

PR #10845, opened 2022-06-15
Inform about path for UnexpectedValeException

imme-emosol:


imme-emosol:
This for instance happens when using path repository and in the path is
a project with a composer.json that has an empty/wrong value in require's version constraint.

PR #10846, opened 2022-06-15
Ignoring passed arguments for @putenv

maryo:
When running a composer script with arguments, the arguments are currently passed to each executable.
This causes problems when using `@putenv` feature.

For example like this

```
"test:integration-fast": [
    "@putenv POPULATE_DB=0",
    "@test:integration"
],
```

then calling `composer test:integration-fast tests/integration/Database/DatabaseTest.php`

sets ENV `POPULATE_DB=0 'tests/integration/Database/DatabaseTest.php'` instead of `0`.

This change ignores the arguments when using `@putenv`.

OFC, this is a minor BC break so please consider if it's acceptable.

Thanks.

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -272,11 +272,11 @@ class EventDispatcher
                         }
                     }
 
-                    if (strpos($exec, '@putenv ') === 0) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, merged in 2.2 as https://github.com/composer/composer/commit/690ab5166a2a5ef6db7125ca8346970c9832590e

PR #10849, opened 2022-06-16
Final public method for abstract class

MathiasReker:
All public methods of abstract classes should be final. Enforce API encapsulation in an inheritance architecture. If you want to override a method, use the Template method pattern.


Seldaek:
This is madness tbh. Huge BC break, no way this is happening. Vut thanks for telling us what we should do and which patterns we should use, that is really helpful 🙄

PR #10850, opened 2022-06-16
Nullable type declaration for default null value

MathiasReker:
Adds ? before type declarations for parameters with a default null value


Seldaek:

PR #10851, opened 2022-06-16
Ordered interfaces

MathiasReker:
Orders the interfaces in an implements or interface extends clause.


Seldaek:

PR #10852, opened 2022-06-16
Regular callable call

MathiasReker:
Callables must be called without using call_user_func* when possible.


Seldaek:

PR #10856, opened 2022-06-16
Assign null coalescing to coalesce equal

MathiasReker:
Use the null coalescing assignment operator ??= where possible.


Seldaek:
This breaks Composer as we support 7.2 which does not have ??=.

PR #10858, opened 2022-06-16
Octal notation

MathiasReker:
Literal octal must be in 0o notation.


Seldaek:
This is only available in php8.1..

PR #10859, opened 2022-06-16
Ternary to null coalescing

MathiasReker:
Use null coalescing operator ?? where possible.


Seldaek:

PR #10861, opened 2022-06-16
Get class to class keyword

MathiasReker:
Replace get_class calls on object variables with class keyword syntax.


Seldaek:
This breaks Composer as we support 7.2 which does not support ::class on variables. Also this is change for the sake of change IMO, it does not add much value.

PR #10864, opened 2022-06-16
No alias functions

MathiasReker:
Master functions shall be used instead of aliases.


Seldaek:
Thanks, that one slipped in

PR #10865, opened 2022-06-16
Use arrow functions

MathiasReker:
Anonymous functions with one-liner return statement must use arrow functions.


Seldaek:
This breaks Composer as we support 7.2.

PR #10866, opened 2022-06-16
Void return

MathiasReker:
Add void return type to functions with missing or empty return statements, but priority is given to @return annotations.


Seldaek:
Cannot do this for BC reasons

PR #10867, opened 2022-06-16
New with braces

MathiasReker:
All instances created with new keyword must (not) be followed by braces.


Seldaek:
Change for the sake of change, zero added value.

PR #10868, opened 2022-06-16
Combine consecutive issets

MathiasReker:
Using isset($var) && multiple times should be done in one call.


Seldaek:

PR #10869, opened 2022-06-16
Combine consecutive unsets

MathiasReker:
Calling unset on multiple items should be done in one call.


Seldaek:
Closing as CS changes is the responsibility of maintainers. And I don't find this particularly more readable.

PR #10870, opened 2022-06-16
Error suppression

MathiasReker:
Error control operator should be added to deprecation notices and/or removed from other cases.


Seldaek:
The composer error handler is smart enough to not crap out on deprecation warnings, and other error handlers should too. Besides this code isn't really meant to be run in other contexts than Composer so I think it's good enough as it is.

PR #10871, opened 2022-06-16
Explicit indirect variable

MathiasReker:
Add curly braces to indirect variables to make them clear to understand.

### DIFF
--- a/src/Composer/Installer/InstallationManager.php
+++ b/src/Composer/Installer/InstallationManager.php
@@ -414,7 +414,7 @@ class InstallationManager
                 if ($this->io->isDebug()) {
                     $this->io->writeError('  - ' . $operation->show(false));
                 }
-                $this->$opType($repo, $operation);
+                $this->{$opType}($repo, $operation);
 
                 continue;
             }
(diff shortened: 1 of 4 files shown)


localheinz:
@MathiasReker 

While I believe that your suggestions are good, these could be potential modifications to [`.php-cs-fixer.php`](https://github.com/composer/composer/blob/main/.php-cs-fixer.php). However, I believe @Seldaek has rejected these changes in the past.

MathiasReker:
I think it will make sense to come up to date with these changes. Additionally, you may want to consider updating the [.php-cs-fixer.php](https://github.com/composer/composer/blob/main/.php-cs-fixer.php) file. In that case, it will make sense to merge PRs first so that CI does not fail when [.php-cs-fixer.php](https://github.com/composer/composer/blob/main/.php-cs-fixer.php) is updated.

Seldaek:
All these PRs could have been discussed first.. applying code style changes to code is a task of maintainers IMO, not of drive-by-PRs.

And all these PRs could have been done in one PR then, with individual commits perhaps, but not flooding my inbox with 50ish new PRs.

TLDR: this pisses me off, and I'll ignore it until I find more time to look at it again, because right now my gut reaction would be to close it all in one go. Enjoy your weekend.

MathiasReker:
Hello @Seldaek 

I am very sorry that the process has not been correct. It has not been my intention to bother you.

I think composer is a super cool tool so I would see if I couldn't help improve the overall quality.

I agree that there are many PR's. My thought was that it was easy to relate to one thing at a time with a well described PR. In that way, PR's could be selected individually.

Next time I will follow your process. I can follow your point.

MathiasReker:
@Seldaek Point taken. Thank you for your feedback. :-)

PR #10872, opened 2022-06-16
Fopen flags

MathiasReker:
The flags in fopen calls must omit t, and b must be omitted or included consistently.


Seldaek:
As per [the docs](https://www.php.net/manual/en/function.fopen.php) `b` is the default if `t` is not given, which we don't do, so I don't see what this is doing.

PR #10875, opened 2022-06-16
No superfluous elseif

MathiasReker:
Replaces superfluous elseif with if.


Seldaek:

PR #10879, opened 2022-06-16
Standardize increment

MathiasReker:
Increment and decrement operators should be used if possible.


Seldaek:

PR #10881, opened 2022-06-16
Php unit dedicate assert

MathiasReker:
PHPUnit assertions like assertInternalType, assertFileExists, should be used over assertTrue.


Seldaek:
Thanks

Issue #10241, opened 2021-10-29
Introduce an option of `dump-autoload` to fail on PSR uncompliant classes


Kolyunya:
I would like to suggest adding an option to the `dump-autoload` command which would force it to fail when it encounters PSR uncompliant classes. This would be very profitable to add this kind of validation to the CI scripts.

Currently, the command just prints to the stdout notices like the following but the return code is always `0` thus there's no straightforward way to automate this important validation process.
```
Class Foo located in ./Foo.php does not comply with psr-4 autoloading standard. Skipping.
```

Seldaek:
This is really not so trivial to add IMO.. So sounds like not really worth the trouble to me.

danepowell:
Agreed, and I'd like to see this option anywhere that a warning can be generated. For instance, you can get the PSR-4 warnings on `composer install` as well. It's hard to have bulletproof CI when problems like this can't be caught.

Seldaek:
Yeah I got a possible solution in the works - hopefully can wrap this up next week

Seldaek:
See https://github.com/composer/composer/pull/10886

Kolyunya:
@Seldaek thank you so much for your work! Looking forward for a release to test it out!

Seldaek:
You can give it a shot with "composer self-update --snapshot" already

danepowell:
Thanks! We've already implemented this in our CI to catch tests with bad class names that Composer would otherwise fail to load.

PR #10886, opened 2022-06-19
Add --strict-psr flag to dump-autoload to fail the process if psr violations were detected

Seldaek:
fixes https://github.com/composer/composer/issues/10241

Builds up on #10885 so really only the third commit (https://github.com/composer/composer/commit/a41fb25ecf26e4e5fedf42f2f9a2379cc831f052) is relevant here.


mvorisek:
@Seldaek In our CI we use `composer validate --strict` to validate composer file, what about supporting `--strict-psr flag` in `composer validate` command as well?

Seldaek:
IMO you rather do dump with --strict-psr on top of it if you want that, because it requires an optimized dump and right now validate doesn't really do any dumping at all, so not sure it'd fit in there.

PR #10893, opened 2022-06-22
Audit command docs: Remove currently restriction

naderman:
- No reason to state that the audit command "currently" only checks packagist.org. It should stay that way, if we need to add more sources, we can add them to packagist.org.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1004,7 +1004,7 @@ php composer.phar archive vendor/package 2.0.21 --format=zip
 ## audit
 
 This command is used to audit the packages you have installed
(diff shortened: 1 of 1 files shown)


glaubinix:
The API already returns advisories for packages that are not hosted on packagist.org if they are part of the GitHub or FriendsOfPHP advisory database. I think for now the currently is fine, might explain to users why it doesn't work for them at the moment if their setup doesn't allow connections to packagist.org and letting them know that this is something that will/might change in the future

Seldaek:
Merging, we can always update if/when it changes.

PR #10899, opened 2022-06-24
Generate autoload files only when some config or package changed

### LINKED ISSUE #10792: Skip generate autoload if <nothing to install>
Running `composer install` always runs `composer dump-autoload` afterwards, showing `Generating optimized autoload files`, even if the install output is `Nothing to install, update or remove`.

Always re-generating autoload files is an unnecessary step for us, that consumes too much time when handling docker containers. 

But if we skip this during installation using `--no-autoloader` and there was a package installed, we need to manually run `composer dump-autoload` to update the autoload files.

I could not find any workaround yet. So I guess it should be possible to skip this behavior conditionally, if something/nothing was installed.

Seldaek:
Fixes #10792

This is in theory a nice change IMO, minor perf boost when an `install` ends up not changing anything.

In practice however I worry this will miss changes. One problem I can think of is with `path` repositories, a symlink package may have some changes on disk which are not reflected by a reference change in installed.json, and will thus not be detected. This might be fixable by detecting whether any package has a dist type of `path`. 

So I don't know if it's worth going further.. Will there be other edge cases/regressions?

cc @btxtiger

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -162,7 +162,7 @@ class AutoloadGenerator
     /**
      * @param string $targetDir
      * @param bool $scanPsrPackages
-     * @return ClassMap
+     * @return ClassMap|null
      * @throws \Seld\JsonLint\ParsingException
      * @throws \RuntimeException
      */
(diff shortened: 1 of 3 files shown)


stof:
Be careful about not skipping the generation of an optimized autoloader (which can change without any change to the autoloading config of a package).

And when using `classmap` as autoloading type, you need to include all the processed files in the hash (because if one of those files changes the classes it contains, this impact the generated autoloader, even non-optimized)

Seldaek:
Yeah if the main project has classmap autoloading we cannot do this optimization IMO as we'd have to rescan all classes anyway.. It gets too complex.



> Be careful about not skipping the generation of an optimized autoloader (which can change without any change to the autoloading config of a package).

I think it's already taken care of but maybe I'm misunderstanding?

Anyway I'm still unsure if this is a good thing to do or not. Seems like it's easy to miss some edge case which will cause trouble later, and the savings aren't that huge.

stof:
> I think it's already taken care of but maybe I'm misunderstanding?

Changing the optimization config of the generator will indeed invalidate that cache as you hash the config. But if the config is for a generated autoloader and the previous one was as well, you have the same invalidation issue than for cases where the root package has classmap autoloading.

Seldaek:
Yeah I think I'm gonna close this as not worth the risk/trouble.

PR #10904, opened 2022-06-30
Fix TypeError

ktomk:
Fix string type requirement in case of NULL value:

    [TypeError]                                                                                                                                                                  
    Composer\Command\InitCommand::parseAuthorString(): Argument #1 ($author) must be of type string, null given, called in phar:///composer.phar/src/Composer/Command/InitCommand.php on line 345

### DIFF
--- a/src/Composer/Command/InitCommand.php
+++ b/src/Composer/Command/InitCommand.php
@@ -342,7 +342,7 @@ EOT
                     return;
                 }
                 $value = $value ?: $author;
-                $author = $this->parseAuthorString($value);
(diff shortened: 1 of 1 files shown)


ktomk:
The previous line 344 ternary shorthand lets NULL $author pass on empty input value, propagating NULL to $author parameter which throws the TypeError.

Fix is to fall-through by null-coalescing operator to empty string (zero length string).

ktomk:
Fix not necessary in 2.2 / 1.1.

Seldaek:
Thanks

Issue #10907, opened 2022-07-01
Config.php ("allow-plugins") : Expected parameter 1 to be an array, bool given


Striffly:
**My `composer.json`:**

```json
{
  "config": {
    "allow-plugins": true
  }
}
```

**Output of `composer diagnose`:**

```
array_merge(): Expected parameter 1 to be an array, bool given
```

**When I run this command**:

```
composer install
```

**I get the following output:**

```
In Config.php line 192:
                                                                  
  [ErrorException]                                                
  array_merge(): Expected parameter 1 to be an array, bool given  
                                                                  

Exception trace:
  at phar:///usr/local/bin/composer/src/Composer/Config.php:192
 Composer\Util\ErrorHandler::handle() at n/a:n/a
 array_merge() at phar:///usr/local/bin/composer/src/Composer/Config.php:192
 Composer\Config->merge() at phar:///usr/local/bin/composer/src/Composer/Factory.php:326
 Composer\Factory->createComposer() at phar:///usr/local/bin/composer/src/Composer/Factory.php:615
 Composer\Factory::create() at phar:///usr/local/bin/composer/src/Composer/Console/Application.php:470
 Composer\Console\Application->getComposer() at phar:///usr/local/bin/composer/src/Composer/Command/BaseCommand.php:122
 Composer\Command\BaseCommand->tryComposer() at phar:///usr/local/bin/composer/src/Composer/Command/BaseCommand.php:200
 Composer\Command\BaseCommand->initialize() at phar:///usr/local/bin/composer/vendor/symfony/console/Command/Command.php:264
 Symfony\Component\Console\Command\Command->run() at phar:///usr/local/bin/composer/vendor/symfony/console/Application.php:1024
 Symfony\Component\Console\Application->doRunCommand() at phar:///usr/local/bin/composer/vendor/symfony/console/Application.php:299
 Symfony\Component\Console\Application->doRun() at phar:///usr/local/bin/composer/src/Composer/Console/Application.php:335
 Composer\Console\Application->doRun() at phar:///usr/local/bin/composer/vendor/symfony/console/Application.php:171
 Symfony\Component\Console\Application->run() at phar:///usr/local/bin/composer/src/Composer/Console/Application.php:130
 Composer\Console\Application->run() at phar:///usr/local/bin/composer/bin/composer:88
 require() at /usr/local/bin/composer:29

[...]

PR #10909, opened 2022-07-01
Correctly merge boolean flag of allow-plugin config

aschempp:
This attempts to fix https://github.com/composer/composer/issues/10907

- if global config has `allow-plugins: true` and local has `allow-plugins: [some-array]`, only local would be allowed
- if local has `allow-plugins: true` and global has an array, local would win and allow everything

I guess that's how it's supposed to work?

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -34,7 +34,7 @@ class Config
     public static $defaultConfig = array(
         'process-timeout' => 300,
         'use-include-path' => false,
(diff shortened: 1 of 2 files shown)


Seldaek:
Oh crap, that's a nasty null behavior :/ Thanks for the fix!

dzuelke:
Ahhh, gotta love me time bombs in code :p

Thanks for the quick releases, @Seldaek.

PR #10915, opened 2022-07-02
redownload if zip file is error

ym2048:
We found that the services deployed in Europe were wrong when implementing composer install to download dependency from our private nexus. There are two errors, one is that the downloaded code is not a legal zip package, and the other is that the code package checksum is incorrect. We carefully checked the package on nexus, and confirmed that there is no problem with the package. 
Platform infos:
 PHP 7.3
composer 2.2.7

We tried to upgrade to the latest version of composer, but the problem is still there.

Finally, it was solved by re-download the code package when there was an error which is the code of this merge request.

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -208,7 +208,15 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
                 }
 
                 if ($checksum && hash_file('sha1', $fileName) !== $checksum) {
-                    throw new \UnexpectedValueException('The checksum verification of the file failed (downloaded from '.$url['base'].')');
+                    // try redownload with file_get_contents
+                    $content = file_get_contents($url['base']);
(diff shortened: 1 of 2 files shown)


ym2048:
the checksum error is as follow
15d152b7-8a41-4a29-b386-1ab694925d14

ym2048:
the checksum error is as follow
ae15baca-705a-4c67-87f9-52f0976586cc

ym2048:
This change is harmless and helps to increase the stability of the download function, so I think it can be merged into master, or do you have any better suggestions? Looking forward to communicating with you, thank you.

Seldaek:
Sorry but this seems like adding some hacky mess in our code (plus doing file_get_contents download instead of using out HttpDownloader class) instead of making sure the broken implementation of a repository is fixed.. This is a nexus problem IMO not a Composer problem.

We do offer reliable and Composer-compatible private hosting at [Private Packagist](https://packagist.com) btw if that helps.

PR #10921, opened 2022-07-04
Fix deprecation notice

fluffycondor:
Deprecation Notice: trim(): Passing null to parameter #1 ($string) of type string is deprecated in phar:///usr/bin/composer/src/Composer/Util/GitHub.php:103

### DIFF
--- a/src/Composer/Util/GitHub.php
+++ b/src/Composer/Util/GitHub.php
@@ -101,8 +101,7 @@ class GitHub
         $this->io->writeError(sprintf('Tokens will be stored in plain text in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
         $this->io->writeError('For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth');
 
-        $token = $this->io->askAndHideAnswer('Token (hidden): ');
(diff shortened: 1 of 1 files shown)


stof:
This will break if someone enters only spaces, as you will still end up with an empty token by turning it empty after the validation.

src/Composer/Util/GitHub.php
```diff
@@ -101,9 +101,10 @@ public function authorizeOAuthInteractively($originUrl, $message = null)
         $this->io->writeError(sprintf('Tokens will be stored in plain text in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
         $this->io->writeError('For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth');
 
-        $token = trim($this->io->askAndHideAnswer('Token (hidden): '));
+        $token = $this->io->askAndHideAnswer('Token (hidden): ');
+        $token = is_string($token) ? trim($token) : '';
```
Seldaek, line 105: IMO casting is fine here, it's exactly the same effect but reads easier (to me anyway..)

```suggestion
        $token = trim((string) $this->io->askAndHideAnswer('Token (hidden): '));
```

Seldaek:
Thanks

PR #10929, opened 2022-07-07
Pre-commit hook to run composer validate

fengtan:
Is it possible to add a [pre-commit](https://pre-commit.com/) hook to this repo ? [Several projects](https://pre-commit.com/hooks.html) do provide a hook.

Here is a hook that automatically validates `composer.json` and `composer.lock` when one modifies these files and tries to commit those.

How to test this PR:

#### 1. Get an environment with PHP, Composer, Git and Pre-commit

```sh
$ docker run --rm -it php:7.4-cli /bin/sh
$ apt-get update && apt-get install --yes git wget unzip pip vim
$ wget https://raw.githubusercontent.com/composer/getcomposer.org/76a7060ccb93902cd7576b67264ad91c8a2700e2/web/installer -O - -q | php -- --quiet && mv composer.phar /usr/local/bin/composer
$ pip install pre-commit
$ git config --global user.email "foo@example.com"
$ git config --global user.name "Foo Bar"
```

#### 2. Create a new project with Composer (say, Laravel) and use our pre-commit hook

```sh
$ composer create-project laravel/laravel laravel 4.2.* --no-plugins
$ cd laravel/
$ git init
$ cat <<EOT >> .pre-commit-config.yaml
repos:
  # You would use https://github.com/composer/composer after this PR is merged
  - repo: https://github.com/fengtan/composer
    # You would use a tag after this PR is merged
    rev: cf70a2c62a76c2749ce75790b3519c01d2df2fb6
    hooks:
      - id: composer-validate
EOT
$ pre-commit install
```

#### 3. Simulate an invalid `composer.lock`

```
$ echo foo >> composer.lock
```

#### 4. Try to commit -- the hook will run and fail because of the invalid `composer.lock`

```
$ git add -A
$ git commit
[INFO] Initializing environment for https://github.com/fengtan/composer.
composer validate........................................................Failed
- hook id: composer-validate
- exit code: 1

In JsonFile.php line 349:
                                                 
  "./composer.lock" does not contain valid JSON  
  Parse error on line 2151:                      
  ...version": "2.3.0"}foo                       
  --------------------^                          
  Expected one of: 'EOF', '}', ',', ']'          
```

If/when this is merged we may want to request it to be documented at https://pre-commit.com/hooks.html

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -7,3 +7,11 @@
   language: system
   pass_filenames: false
   files: '^composer.(json|lock)$'
+
+- id: composer-validate-strict
+  name: composer validate (strict mode)
+  description: Validates composer.json and composer.lock with --strict mode enabled
+  entry: composer validate --strict
+  language: system
+  pass_filenames: false
+  files: '^composer.(json|lock)$'
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+# See https://pre-commit.com/#new-hooks
+- id: composer-validate
+  name: composer validate
+  description: Validates composer.json and composer.lock
+  entry: bin/composer validate
+  language: script
+  pass_filenames: false
+  files: '^composer.(json|lock)$'
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,8 +1,9 @@
 # See https://pre-commit.com/#new-hooks
+# Requires composer to be installed locally
 - id: composer-validate
   name: composer validate
   description: Validates composer.json and composer.lock
-  entry: bin/composer validate
-  language: script
+  entry: composer validate
+  language: system
   pass_filenames: false
   files: '^composer.(json|lock)$'


Seldaek:
@stloyd @cafferata any negative opinion you'd like to share here?

I don't really have a strong opinion either way, it seems harmless to add this but on the other hand it could also be provided somewhere else on another repo as many other projects seem to do (probably because it's then managed by the community of people who care about pre-commit.com - making it more maintained).

So I'd say perhaps it's best if you publish this yourself @fengtan?

cafferata:
Sure! Ask yourself, is this a Composer functionality? Is this change for the benefit of a Composer user? The answer on both questions will be 'no, is not'. Makes it the life of a [pre-commit](https://pre-commit.com/) user easier? Absolute. But, is the Composer repository/organization the right place for that? No, a community repository is a much better place for that. Since all pre-commits are [linked in](https://pre-commit.com/hooks.html), there is no advantage in adopting/maintaining the Composer pre-commit by yourself.

Assuming there is a good will to help @fengta. You have to ask yourself if you want to bring this maintaining work to your self? There are a lot of plugins/tools that want or can use Composer. Two examples that come to my mind:
1. [GrumPHP - Composer task](https://github.com/phpro/grumphp/commits/master/doc/tasks/composer.md)
2. [Warp - Workflows](https://www.commands.dev/)

Are you going to adopting/maintaining these examples in the Composer repository too? Are you open to that? Or maybe even bigger/more out-of-the-ordinary. Why has there never been a Composer plugin central management set up within the Composer organization? Probably all for the same reason, there is no immediate added value for you/Composer.

Questions you should think about more. What if other pre-commit users want additions? That might mean breaking changes are necessary. Are you going to communicate these from the Composer releases (notes)? Are you going to maintain this in sem version? How do you handle the (extreme?) situation when pre-commit syntax changes from YAML to XML. How do these external changes relate to Composer sem version strategy?

Probably still a lot to think about. My advice would be, let @fengta (_or anyone else from the pre-commit community_) set up a git repository by himself and include it from the [pre-commit list](https://pre-commit.com/hooks.html). The result will be the same, all pre-commit users can apply this hook. Everyone happy right? Without costing you additional future work. A big win for you.

Seldaek:
Yes, I think you put what I said above very well in a longer and more detailed form :)

I'll close this then. Thanks anyway @fengtan - I hope you will publish this somewhere anyway as it seems it could be useful to some.

fengtan:
Thanks for the feedback @Seldaek @cafferata !

Some communities prefer to distribute pre-commit hooks as part of their official repo (e.g. [ansible-lint](https://github.com/ansible/ansible-lint/blob/main/.pre-commit-hooks.yaml), [yamllint](https://github.com/adrienverge/yamllint/blob/master/.pre-commit-hooks.yaml) and [codespell](https://github.com/codespell-project/codespell/blob/master/.pre-commit-hooks.yaml)) whereas others prefer to store them in separate repos.

It sounds like in our case a separate repo would make more sense so I have created https://github.com/fengtan/pre-commit-composer

Cheers !

PR #10932, opened 2022-07-09
test: Create test for CheckPlatformReqsCommand

theoboldalex:
Created a test for the CheckPlatformReqsCommand which increases coverage from 8% to 76%.

### DIFF
--- a/tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
+++ b/tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
@@ -28,7 +28,7 @@ class CheckPlatformReqsCommandTest extends TestCase
         $appTester->assertCommandIsSuccessful();
     }
 
-    protected function flagGenerator(): \Generator
+    public function flagGenerator(): \Generator
     {
         yield 'Disables checking of require-dev packages requirements.' => [
             ['--no-dev' => true]
--- a/tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
+++ b/tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Test\Command;
+
+use Composer\Test\TestCase;
+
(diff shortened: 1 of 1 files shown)


tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
```diff
@@ -0,0 +1,45 @@
...
+        $appTester->run(array_merge(['command' => 'check-platform-reqs'], $flag));
+
+        $appTester->assertCommandIsSuccessful();
+    }
+
+    public function flagGenerator(): \Generator
+    {
+        yield 'Disables checking of require-dev packages requirements.' => [
+            ['--no-dev' => true]
+        ];
```
Seldaek, line 35: This is good to have but would be great to verify the behavior actually works by creating some packages and checking the output matches expected output.

See TestCase::initTempComposer, createInstalledJson  and createComposerLock for this. https://github.com/composer/composer/blob/70f2dd6eddd5916f70b3d73dd3170a3d2d3a7681/tests/Composer/Test/Command/BumpCommandTest.php#L28-L47 also has a real example of using these.

tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+        yield 'Whether to disable plugins.' => [
+            ['--no-plugins' => true]
+        ];
+
+        yield 'Checks requirements only from the lock file, not from installed packages.' => [
+            ['--lock' => true]
+        ];
```
Seldaek, line 43: As above, would be nice to check this actually does something and not just runs without failing.

Seldaek (changes requested):

Seldaek:
Thanks for getting started here. This definitely is better than nothing as it runs the code and would catch any critical error, but it'd be cool if you find time to go further and actually check the output. See this example too for inspiration on checking output https://github.com/composer/composer/blob/70f2dd6eddd5916f70b3d73dd3170a3d2d3a7681/tests/Composer/Test/Command/UpdateCommandTest.php#L31

theoboldalex:
Thanks for the feedback. I'll get to work on the requested changes.

jeffwidman:
Gentle nudge @theoboldalex be great to see this move forward

theoboldalex:
Apologies for the delay on getting back to this. I have actioned the points raised by @Seldaek so this is ready for another review. Thanks

Seldaek (approved):
Looks good now, just a couple style nitpicks

Seldaek:
Thanks!

theoboldalex:
I have actioned the code style fixes as requested.

PR #10933, opened 2022-07-09
Update README.md

cxtruong70:
Minor change. Added a new line between sentence regarding Public and Private packages for slightly increased readability through spatial distinction.


Seldaek:
Thanks

PR #10937, opened 2022-07-12
BUG#AC-5897:Allure reports are not generated for composer builds when…

faizan-shk:
… running from repo

        - Error while running MFTF test magento 2.4.4-p1 and magento-2.4.5 - magento/composer-dependency-version-audit-plugin contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -700,7 +700,7 @@ class PluginManager
 
         // This is a BC mode for lock files created pre-Composer-2.2 where the expectation of
         // an allow-plugins config being present cannot be made.
-        if ($rules === null) {
+        if (empty($rules)) {
             if (!$this->io->isInteractive()) {
                 $this->io->writeError('<warning>For additional security you should declare the allow-plugins config with a list of packages names that are allowed to run code. See https://getcomposer.org/allow-plugins</warning>');
                 $this->io->writeError('<warning>This warning will become an exception once you run composer update!</warning>');


stof:
I don't understand what you are trying to solve here.

faizan-shk:
> I don't understand what you are trying to solve here.

In case when the rule is returned as empty array i.e **array()** the condition becomes invalid and then Exception is thrown, inspite of adding the entry in **allowed-plugins**

faizan-shk:
@Seldaek can you please check this!

In case where rules = [], the throws error,

Screenshot 2022-07-13 at 11 45 26 AM

I already have added in my composer.json the entry for it

Screenshot 2022-07-13 at 4 18 36 PM

Seldaek:
The current behavior is correct IMO, if you set `allow-plugins: []` or simply do not declare it, then it will prompt you if interactive or throw if non-interactive, to ensure that non-allowed plugins are not executed.

faizan-shk:
> 

We did try with allow-plugins option as you can see in above comment, however the issue still persists

stof:
Well, the screenshot is from a custom command that probably executes composer internally. But then, the question is whether the composer.json that you edited is the composer.json for which composer is executed by your command.

faizan-shk:
@Seldaek @stof this issue is resolved yesterday after 2.2.17 release, it was happening specifically for 2.2.16, previous versions worked fine as well.

Closing this PR as it is no longer required!

PR #10956, opened 2022-07-20
Remove -p from --patch-only, --platform took it

joelpittet:
Stumbled upon this with the output
```
❯ composer show -Dlop
The --direct (-D) option is not usable in combination with --all, --platform (-p) or --available (-a)
```


joelpittet:
Just a nuance, it works with outdated so didn't update that instance...

Seldaek:
Ah good catch, and yes indeed the outdated options on show command are kinda not so relevant as you should ideally use them via outdated.

PR #10962, opened 2022-07-22
Improve documentation about the new bash completion feature

mitelg:
Hello :wave: 

I tried the new autocompletion feature and thought I was doing something wrong, until I accidentally discovered the `completion.bash` file in the directory I was working in. Yes, I didn't remember what `>` is doing :sweat_smile: 

To make it more clear, what happens, I slightly adjusted the docs. Hope that helps other people, too :slightly_smiling_face:

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -15,9 +15,10 @@ calls `composer dump-autoload`.
 
 ## Bash Completions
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, makes sense.

PR #10965, opened 2022-07-23
Add json format to Audit command

barryvdh:
Follow up on #10798 and #10898, as suggested in https://github.com/composer/composer/pull/10798#issuecomment-1152782931

This adds a `format=json` option to the command, to allow for. Currently in the same format it already is (because why not?). In the 'advisories' object, so a summary or the list of packages can still be added later if required. So grouped by package:

image

No actual real-life use case for this, except for opening the possibility to process it during a pipeline/action (and perhaps suggest updates or format an advisory yourself)

### DIFF
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -4,6 +4,7 @@ namespace Composer\Advisory;
 
 use Composer\IO\ConsoleIO;
 use Composer\IO\IOInterface;
+use Composer\Json\JsonFile;
 use Composer\Package\PackageInterface;
 use Composer\Repository\RepositorySet;
 use InvalidArgumentException;
@@ -18,11 +19,14 @@ class Auditor
 
     public const FORMAT_PLAIN = 'plain';
 
+    public const FORMAT_JSON = 'json';
+
     public const FORMAT_SUMMARY = 'summary';
 
     public const FORMATS = [
         self::FORMAT_TABLE,
         self::FORMAT_PLAIN,
+        self::FORMAT_JSON,
         self::FORMAT_SUMMARY,
     ];
 
@@ -37,6 +41,11 @@ class Auditor
     public function audit(IOInterface $io, RepositorySet $repoSet, array $packages, string $format, bool $warningOnly = true): int
     {
         $advisories = $repoSet->getMatchingSecurityAdvisories($packages, $format === self::FORMAT_SUMMARY);
+        if (self::FORMAT_JSON === $format) {
+            $io->write(JsonFile::encode(['advisories' => $advisories]));
+            return count($advisories);
+        }
+
         $errorOrWarn = $warningOnly ? 'warning' : 'error';
         if (count($advisories) > 0) {
             [$affectedPackages, $totalAdvisories] = $this->countAdvisories($advisories);
--- a/src/Composer/Command/AuditCommand.php
+++ b/src/Composer/Command/AuditCommand.php
@@ -21,7 +21,7 @@ class AuditCommand extends BaseCommand
             ->setDescription('Checks for security vulnerability advisories for installed packages.')
             ->setDefinition(array(
                 new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables auditing of require-dev packages.'),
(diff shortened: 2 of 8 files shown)


src/Composer/Command/AuditCommand.php
```diff
@@ -21,7 +21,7 @@ protected function configure(): void
             ->setDescription('Checks for security vulnerability advisories for installed packages.')
             ->setDefinition(array(
                 new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables auditing of require-dev packages.'),
-                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Output format. Must be "table", "plain", or "summary".', Auditor::FORMAT_TABLE, Auditor::FORMATS),
+                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Output format. Must be "table", "plain", "json" or "summary".', Auditor::FORMAT_TABLE, Auditor::FORMATS),
```
Seldaek, line 24: Thanks, looks good I think! just one comment can you please update the other Input Option in other commands (look for audit-format) as well as the 04-cli.md docs to make sure the json format is listed everywhere?

Seldaek (approved):

src/Composer/Command/AuditCommand.php
barryvdh: Yes updated!

src/Composer/Advisory/Auditor.php
```diff
@@ -37,6 +41,11 @@ class Auditor
     public function audit(IOInterface $io, RepositorySet $repoSet, array $packages, string $format, bool $warningOnly = true): int
     {
         $advisories = $repoSet->getMatchingSecurityAdvisories($packages, $format === self::FORMAT_SUMMARY);
+        if (self::FORMAT_JSON === $format) {
+            $io->write(JsonFile::encode(['advisories' => $advisories]));
```
GuySartorelli, line 45: Why add the "advisories" key, as opposed to just encoding the array as is?

src/Composer/Advisory/Auditor.php
```diff
@@ -37,6 +41,11 @@ class Auditor
     public function audit(IOInterface $io, RepositorySet $repoSet, array $packages, string $format, bool $warningOnly = true): int
     {
         $advisories = $repoSet->getMatchingSecurityAdvisories($packages, $format === self::FORMAT_SUMMARY);
+        if (self::FORMAT_JSON === $format) {
```
GuySartorelli, line 44: This should me moved into `outputAdvisories()` with the other format outputs.

GuySartorelli (changes requested):

src/Composer/Advisory/Auditor.php
barryvdh: I think that is more clear. And also leaves the option to add any other (meta) data like summary/number of errors/list of packages later.

src/Composer/Advisory/Auditor.php
barryvdh: I originally did that, but I think this was clearer in the end, because the section below also writes the error summary (good or bad) and I don't think you want that  (even though it's in the error output). And that output is only written when there are advisories, I think it's clearer that there is always the advisories array, even when empty.

GuySartorelli: The "found x advisories" could (probably should) also be moved into the output method, and only be done for the nonhjson formats.

barryvdh:
I've updated the docs and audit-format command. Not sure why CI is failing on 7.3 only. It looks a bit random?

Seldaek:
Yeah sorry that's a temporary failure because we're on a dev version of symfony console, ignore it.

src/Composer/Advisory/Auditor.php
barryvdh: Okay, I'll let @Seldaek decide on this. What do you prefer?

Seldaek: I don't have a super strong opinion either way. It's all private code so can be refactored later if really needed. As such I think it's good enough as is.

Seldaek:
Thanks again

PR #10975, opened 2022-07-30
when my gitlab domain is git.example.org,clone fail

Chrisdowson:


Seldaek:
You should also configure the gitlab-domains option, then things should be handled correctly I believe.

Seldaek:
See my comment above.. I don't think anything needs fixing on our side.

PR #10976, opened 2022-07-30
[FEATURE] allow multible `@php` within scripts

xerc:
`@php vendor/bin/typo3cms upgrade:prepare && @php vendor/bin/typo3cms cache:flush`


Seldaek:
Thanks for the PR but I'd rather not do that as you can easily add more scripts by defining the script as an array of strings, and doing a str_replace like that carries the risk of accidental side effects.

PR #10977, opened 2022-07-30
[BUGFIX] respect PHP-CLI `error_reporting` parameter

xerc:
`/opt/php_8.1/bin/php -d error_reporting=8191 ~/composer u` # E_ALL & ~E_DEPRECATED & ~E_USER_DEPRECATED

### DIFF
--- a/bin/composer
+++ b/bin/composer
@@ -18,7 +18,9 @@ use Composer\XdebugHandler\XdebugHandler;
 use Composer\Util\Platform;
 use Composer\Util\ErrorHandler;
 
-error_reporting(-1);
+if(error_reporting() > 0) {
+    error_reporting(-1);
+}
 
 // Restart without Xdebug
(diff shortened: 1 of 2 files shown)


Seldaek:
Sorry nope, we do this on purpose.

xerc:
may introduce via arg./flag

Seldaek:
I don't see a need for it. We report deprecation errors without aborting the process. If you have an actual issue perhaps you should try to describe that in more details.

xerc:
mainly because i have PHP 8.1 and PHP 7.4 runnig and compiled both with https://github.com/composer/composer/pull/10976 but there are a lot of deprecation messages in the [8] version which would be nice to hide if you do not want to patch in this specific feature into MAIN
for the example provided there ; it is relevant not to clear the (static-)cache if the database upgrade fails [sh:`&&`]

PR #10982, opened 2022-08-04
Add SearchCommand tests

IonBazan:
This PR adds functional tests for `SearchCommand` - searching in the local repository only. See #10796

Please take note of additional changes in `RequireCommandTest` , adding `'packagist.org' => false` to avoid HTTP hit in the tests and causing tests to fail due to unexpected `Info from https://repo.packagist.org: #StandWithUkraine` message in the output. 

@Seldaek Is there a better way to disable `packagist.org` in tests to make sure no additional requests are performed? I noticed `$this->getApplicationTester()` uses `Factory` instead of `FactoryMock` but not sure how to improve that.

Coverage changes from 16.98% to 94.34% for this class.

### DIFF
--- a/tests/Composer/Test/Command/SearchCommandTest.php
+++ b/tests/Composer/Test/Command/SearchCommandTest.php
@@ -1,0 +1,129 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Test\Command;
+
+use Composer\Test\TestCase;
+use InvalidArgumentException;
+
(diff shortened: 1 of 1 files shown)


Seldaek:
I think setting packagist.org=>false in TestCase::initTempComposer inside $composerJson would probably be the better fix. I'd probably only do it if the `repositories` key is set though to avoid side effects in tests which are not interested in loading test packages.

Seldaek:
Fixed the packagist.org issue in https://github.com/composer/composer/commit/a6d872191c14a00d907801c284756100d44c3f86  and rebased.

IonBazan:
Perfect, thanks!

Seldaek:
Thank you for the tests :)

IonBazan:
I'll look into remaining commands that need testing, since the `packagist.org` repository is now disabled in tests. Are there any plans to host the coverage report somewhere like CodeCov, etc?

Seldaek:
No plans. I am not a fan of code coverage monitoring as I don't think it brings much value as an obsession/metric. It is quite useful here as a one-off way to find totally untested code though.

But yes please feel free to send more tests

PR #10988, opened 2022-08-08
Added token refresh for GitLab 

### LINKED ISSUE #10970: Gitlab access token refresh
Since Gitlab 15.0 was released I have to enter my username and password almost every time I use composer update.

See here:
https://docs.gitlab.com/ee/integration/oauth_provider.html#access-token-expiration

Please add support for access token refresh.

Thanks

tlueder:
Fixes #10970


Seldaek:
Thanks - I did remove some output and made sure the old token code still gets used in case the gitlab instance does not an `expires_in` key (some very old instance I think may not).

PR #10994, opened 2022-08-15
Correct prohibits/why-not example command

cdubz:
Running the example command `composer prohibits php:8` returns an error, "Not enough arguments (missing: "version")."


stof (approved):

Seldaek:
Thanks

PR #11001, opened 2022-08-17
performance: Do not create a local cache repo for local repos

exteon:
performance: Do not create a local cache repo for local repos


Seldaek:
Thanks

PR #11005, opened 2022-08-17
Fix docs for issue composer/satis#656

ralflang:
In issue composer/satis#656 somebody mentioned the docs still show an example with invalid json.
The PR fixes the docs, not the original problem.


Seldaek:
Thanks

PR #11008, opened 2022-08-18
Do not apply non-array package links in ArrayLoader

zanbaldwin:
Fix type error when a non-array package link is provided in config array.
Encountered when the following `composer.json` was submitted:

```json
{
    ...,
    "replace": "vendor/package"
}
```


Seldaek:
Thanks

PR #11010, opened 2022-08-18
Match default choice to actual default

yakatz:
The 'Y' is capital, so Yes should be the default choice and is what most people would want, but No was the default.


Seldaek:
Makes sense to me.. Any objection @johnstevenson?

johnstevenson:
None at all. The user can always change their mind with the default Cancel action of the UAC dialog.

Seldaek:
Ok, thanks @yakatz

PR #11013, opened 2022-08-19
Adding hint what "Direct dependencies" means

ThomasLandauer:


Seldaek:
Thanks, tweaked wording a bit further but yeah makes sense to explain clearly what these are.

PR #11014, opened 2022-08-20
Tests bootstrap - remove require InstalledVersions

ondrejmirtes:
This causes issues for PHPStan in the latest version. Because PHPStan itself loads its own InstalledVersions from inside the PHAR, this bootstrap line makes it crash.

If we find out that loading this class is indeed necessary for Composer's test suite to run, I'll create a separate PHPStan bootstrap next to this one.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -10,6 +10,7 @@
  * file that was distributed with this source code.
  */
 
+use Composer\InstalledVersions;
 use Composer\Util\Platform;
 
 error_reporting(E_ALL);
@@ -19,7 +20,11 @@ if (function_exists('date_default_timezone_set') && function_exists('date_defaul
 }
 
 require __DIR__.'/../src/bootstrap.php';
-require __DIR__.'/../src/Composer/InstalledVersions.php';
+
+if (!class_exists(InstalledVersions::class, false)) {
+    require __DIR__.'/../src/Composer/InstalledVersions.php';
+}
+
 require __DIR__.'/Composer/Test/TestCase.php';
 
 Platform::putEnv('COMPOSER_TESTS_ARE_RUNNING', '1');


ondrejmirtes:
Looks like the test suite is green even without this line, so I'd appreciate if this could be merged as-is. Thank you!

(The one failure is already present on main.)

Seldaek:
IIRC I did add this because otherwise it got the one loaded from vendor and or from phpstan and I ended up not catching/reporting errors in the actual file. Any chance you can conditionally load yours only if it doesn't exist yet?

ondrejmirtes:
Just changed it, thanks.

Seldaek:
Sorry I meant can't you do that conditional loading in the PHPStan repo? :D I guess you should even delay it until the project bootstrap has been loaded.. But I can see this is maybe too painful to fix just this one use case.

ondrejmirtes:
I'm sorry, I can't, Symfony Console needs the version provided really early before running PHPStan itself: https://github.com/phpstan/phpstan-src/blob/62ac4d9e35dc8b9e9993c0f15ca7abaa8fdf9400/bin/phpstan#L157

Seldaek:
Yeah this sucks, it's not the first time I see this application version need cause problems. You could easily hardcode it in the phar at build time tho.. But anyway maybe I should submit a PR to symfony/console to allow passing a version callback or something lazy..

Seldaek:
Anyway merging this for now as workaround, thanks

ondrejmirtes:
I don't think it's worth to bother symfony/console with this :) We could read the JSON with the version directly, or just read the information from Git, if we come to another blocker because of this.

ondrejmirtes:
Thank you!

Seldaek:
Or overriding getLongVersion (and perhaps getVersion for completeness) in your own Application extending symfony's would also be an option to lazily evaluate this only when it's needed (which is only on --version I guess)

PR #11015, opened 2022-08-20
Lock down include wrappers to avoid abuse from third parties

Seldaek:
It occurred to me that with static lambdas we can achieve the same effect as with those public functions, except we don't leave them open for use by third parties.

As a side-note @mir-hossein pointed out to me that it's sometimes used in attacks to bypass restrictions, I am not entirely sure how this would work but I guess some filtering may allow calls to `includeFile()` but not to include/require? In any case it seems like a good thing to close this potential vector.

I would target 2.5.0 with this because it may cause disruption to people using this function (even tho they really should not..) for "legit" cases, so I'd rather not have this in a patch release.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -1007,9 +1007,16 @@ REGISTER_LOADER;
 
         if ($useIncludeFiles) {
             $file .= <<<INCLUDE_FILES
-        \$includeFiles = \Composer\Autoload\ComposerStaticInit$suffix::\$files;
-        foreach (\$includeFiles as \$fileIdentifier => \$file) {
-            composerRequire$suffix(\$fileIdentifier, \$file);
+        \$filesToLoad = \Composer\Autoload\ComposerStaticInit$suffix::\$files;
+        \$requireFile = static function (\$fileIdentifier, \$file) {
+            if (empty(\$GLOBALS['__composer_autoload_files'][\$fileIdentifier])) {
+                \$GLOBALS['__composer_autoload_files'][\$fileIdentifier] = true;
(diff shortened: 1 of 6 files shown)


mir-hossein:
Thank you @Seldaek .
I will provide more details about the issue by Email.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -42,6 +42,9 @@
  */
 class ClassLoader
 {
+    /** @var \Closure */
```
staabm, line 45:
```suggestion
    /** @var \Closure(string):void */
```

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -1008,8 +1008,15 @@ public static function getLoader()
         if ($useIncludeFiles) {
             $file .= <<<INCLUDE_FILES
         \$includeFiles = \Composer\Autoload\ComposerStaticInit$suffix::\$files;
+        \$includeFile = static function (\$fileIdentifier, \$file) {
```
stof, line 1011: should probably be named `$requireFile` as it does a `require`, not an `include`.

Seldaek: Right, fixed

src/Composer/Autoload/ClassLoader.php
stof: @Seldaek why not applying that suggestion ?

Seldaek: Oh I don't know what happened to this, nor why it's resolved. Maybe a mistake on the GH mobile app. Anyway will apply.

PR #11023, opened 2022-08-26
Config - always respect COMPOSER_DISCARD_CHANGES

ondrejmirtes:
If it was set to `'0'`, the config value would be read instead.

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -376,7 +376,8 @@ class Config
                 return $value;
 
             case 'discard-changes':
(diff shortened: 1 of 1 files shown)


ondrejmirtes:
The failure is because I released phpstan-symfony that understands `$input->getOptions()` 🤦 

Because PHPStan 1.8.x-dev contains more fixes so that InitCommand is understood better, I'll wait for the next PHPStan 1.8.3 release.

Seldaek:
Thanks, merged as 13421f7d66ae8e824e9b821f2aee75772284cbbb

PR #11030, opened 2022-08-28
Fix the heading levels for the doc about composer repositories

stof:
The headings don't make sense before this change:

image


JabriAbdelilah (approved):

Seldaek:
Thanks

PR #11038, opened 2022-08-30
Update gitlab fetch failure message with new path

KorvinSzanto:
`https://gitlab.com/-/profile/personal_access_token` is not correct, the current URL for personal access token settings is `https://gitlab.com/-/profile/personal_access_tokens`. This is the only occurrence of the bad URL, we use the proper URL elsewhere.


Seldaek:
Thanks

PR #11041, opened 2022-08-31
Add download-only mode

### LINKED ISSUE #11035: Download-only mode
I'm looking to download/cache packages only and it doesn't seem supported.
Digging around in the code it seems like if we `return;` here on line 308:
https://github.com/composer/composer/blob/b6028951b8a2254d4896b4dd13bcbe380a3d51bf/src/Composer/Installer/InstallationManager.php#L304-L316

It would download but not install the packages;

```bash
# composer install --download-only
Package operations: 1 install, 0 updates, 0 removals
  - Downloading phpstan/phpstan (1.7.12)
# 
 ```
Another option would be to implement an `DownloadOperation`, and do a `if ($operation instanceof DownloadOperation) continue;` on line 314 above.

Would any of these be an acceptable solution? In that case I'll prepare a PR.

### PR DESCRIPTION
**Pull request by Jellyfrog**
Fixes #11035

### CODE REVIEW

**Comment by Jellyfrog** [comment]
One thing left to consider fixing is this leaves the `vendor/composer/tmp-*` files.

**Comment by Jellyfrog** [comment]
It now cleans the tmp files.  
Another possible solution would be to add a new operation type, then we don't need to pass the `$downloadOnly`-variable.

File: src/Composer/Installer.php [hunk 1]
```diff
@@ -258,6 +260,12 @@ public function run(): int
             $this->mockLocalRepositories($this->repositoryManager);
         }
 
+        if ($this->downloadOnly) {
+            $this->runScripts = false;
+            $this->writeLock = false;
+            $this->dumpAutoloader = false;
```
Seldaek, line 266: I'm not so sure about thse..

Removing autoload dumping sure, makes sense. 

Disabling scripts I think that is a mistake as plugins and whatever might depend on those scripts to do the downloading correctly. 

And writeLock I think I would leave it on, as if you do an update with download only (not possible via CLI, but possible if you do this programmatically) it should still persist the update result to the lock file IMO.

src/Composer/Installer/InstallationManager.php
```diff
@@ -368,6 +371,12 @@ private function executeBatch(InstalledRepositoryInterface $repo, array $operati
                 $package = $operation->getPackage();
                 $initialPackage = null;
             }
+
+            if ($downloadOnly) {
+                $this->download($package);
+                continue;
+            }
```
Seldaek, line 378: I have a gut feeling this isn't the best way to bypass the operations here.. But I gotta try it myself, if I can improve I'll add a commit to your branch.

Seldaek:
Thanks!

PR #11047, opened 2022-09-03
Add "--dry-run" to bump command

kubawerlos:
With this flag it would be possible to run `composer bump --dry-run` as a part of CI - to keep synced `composer.json` and `composer.lock`.

When run on this repo:
```console
$ php ./bin/compile 
$ php composer.phar bump --dry-run
Warning: Bumping dependency constraints is not recommended for libraries as it will narrow down your dependencies and may cause problems for your users.
./composer.json would be updated with:
 - require.composer/ca-bundle: ^1.3.3
 - require.composer/semver: ^3.3.2
 - require.justinrainbow/json-schema: ^5.2.12
 - require.psr/log: ^1.1.4 || ^2.0 || ^3.0
 - require.seld/jsonlint: ^1.9
 - require.seld/phar-utils: ^1.2.1
 - require.symfony/console: ^5.4.12 || ^6.0.11
 - require.symfony/filesystem: ^5.4.12 || ^6.0
 - require.symfony/finder: ^5.4.11 || ^6.0
 - require.symfony/process: ^5.4.11 || ^6.0
 - require.react/promise: ^2.9
 - require.symfony/polyfill-php73: ^1.26
 - require.symfony/polyfill-php80: ^1.26
 - require.seld/signal-handler: ^2.0.1
 - require-dev.symfony/phpunit-bridge: ^6.1.3
 - require-dev.phpstan/phpstan: ^1.8.3
 - require-dev.phpstan/phpstan-phpunit: ^1.1.1
 - require-dev.phpstan/phpstan-strict-rules: ^1.4.3
 - require-dev.phpstan/phpstan-symfony: ^1.2.13
$ echo $?
1
```

I've swapped the order of `if (!$input->getOption('dev-only'))` and `if (!$input->getOption('no-dev-only'))`, so when there are packages in both to update then `require` is shown first.

### DIFF
--- a/src/Composer/Command/BumpCommand.php
+++ b/src/Composer/Command/BumpCommand.php
@@ -196,7 +196,11 @@ EOT
             $lock->write($lockData);
         }
 
-        return $dryRun && $changeCount > 0 ? self::ERROR_GENERIC : 0;
+        if ($dryRun && $changeCount > 0) {
+            return self::ERROR_GENERIC;
+        }
+
+        return 0;
     }
 
     /**
--- a/src/Composer/Command/BumpCommand.php
+++ b/src/Composer/Command/BumpCommand.php
@@ -45,6 +45,7 @@ final class BumpCommand extends BaseCommand
                 new InputArgument('packages', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, 'Optional package name(s) to restrict which packages are bumped.', null, $this->suggestRootRequirement()),
                 new InputOption('dev-only', 'D', InputOption::VALUE_NONE, 'Only bump requirements in "require-dev".'),
                 new InputOption('no-dev-only', 'R', InputOption::VALUE_NONE, 'Only bump requirements in "require".'),
+                new InputOption('dry-run', null, InputOption::VALUE_NONE, 'Outputs the packages to bump, but will not execute anything.'),
             ])
             ->setHelp(
                 <<<EOT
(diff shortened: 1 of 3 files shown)


cafferata:
Nice! I also played with this idea so that my colleagues can guarantee the Composer bump (from CI)

src/Composer/Command/BumpCommand.php
```diff
@@ -45,6 +45,7 @@ protected function configure(): void
                 new InputArgument('packages', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, 'Optional package name(s) to restrict which packages are bumped.', null, $this->suggestRootRequirement()),
                 new InputOption('dev-only', 'D', InputOption::VALUE_NONE, 'Only bump requirements in "require-dev".'),
                 new InputOption('no-dev-only', 'R', InputOption::VALUE_NONE, 'Only bump requirements in "require".'),
+                new InputOption('dry-run', null, InputOption::VALUE_NONE, 'Outputs the packages to bump, but will not execute anything.'),
```
Seldaek, line 48: Please also add the new flag to the 03-cli.md docs file.

src/Composer/Command/BumpCommand.php
```diff
@@ -184,7 +196,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $lock->write($lockData);
         }
 
-        return 0;
+        return $dryRun && $changeCount > 0 ? self::ERROR_GENERIC : 0;
     }
```
Seldaek, line 200: I'd probably prefer to avoid ternary here:

```suggestion
        if ($dryRun && $changeCount > 0) {
            return self::ERROR_GENERIC;
        }

        return 0;
    }
```

Seldaek:
Looks good overall, thanks, just the couple nitpicks :)

src/Composer/Command/BumpCommand.php
kubawerlos: Done

Seldaek:
Thanks

PR #11049, opened 2022-09-05
`require` command: allow `--fixed` option for library dev dependencies

gharlan:
https://github.com/composer/composer/pull/8313

I think it can make sense to use the `--fixed` option for dev dependencies of libraries. We use fixed versions for static analysis tools etc., because in new tool versions the baseline could change etc.

### DIFF
--- a/src/Composer/Command/RequireCommand.php
+++ b/src/Composer/Command/RequireCommand.php
@@ -173,8 +173,8 @@ EOT
             /**
              * @see https://github.com/composer/composer/pull/8313#issuecomment-532637955
              */
-            if ($packageType !== 'project') {
-                $io->writeError('<error>The "--fixed" option is only allowed for packages with a "project" type to prevent possible misuses.</error>');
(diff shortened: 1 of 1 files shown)


src/Composer/Command/RequireCommand.php
```diff
@@ -173,7 +173,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             /**
              * @see https://github.com/composer/composer/pull/8313#issuecomment-532637955
              */
-            if ($packageType !== 'project') {
+            if ($packageType !== 'project' && !$input->getOption('dev')) {
                 $io->writeError('<error>The "--fixed" option is only allowed for packages with a "project" type to prevent possible misuses.</error>');
```
staabm, line 177: The message not longer fits

Nuttapon43:
@gharlan

Seldaek:
Thanks, merged as fc75efd71

PR #11057, opened 2022-09-13
Delete PORTING_INFO

jeffwidman:
I don't think this file is used anymore?

It was used during the initial port, but that was 12 years ago, so I suspect this is no longer relevant.


jeffwidman:
To be clear: I am not 100% sure this change is safe, but I thought the best way to start that discussion was to open a PR. 😀

Seldaek:
I'd tend to agree, but will defer to @naderman here :)

naderman:
I actually still use this every once in a while when running into some obscure bug in the ported code. Sometimes helpful to look at the original C source for reference to see if this is a mistake when it was ported or if the bug was present in the original code too potentially. I don't see the harm in keeping this file, even if it's just for me? Guess could keep it somewhere else, just not sure where I won't eventually lose/forget it? :D

Seldaek:
It's fine then, let's close I'd say.

jeffwidman:
Makes sense @naderman thanks for clarifying.

PR #11058, opened 2022-09-13
Replace deprecated `PSR-2` with `PSR-12`

jeffwidman:
At the top of https://www.php-fig.org/psr/psr-2/, it says:
> Deprecated - As of 2019-08-10 PSR-2 has been marked as deprecated. PSR-12 is now recommended as an alternative.

Looking at the PSR-12 docs, the summary of changes implies that they mostly took PSR-2 and extended it to clarify new language features.


Seldaek:
Thanks

PR #11071, opened 2022-09-19
Unify command names in the docs

kubawerlos:
It would be great to have the CLI docs consistent, as currently some aliases of commands are skipped, some are in parenthesis, and some are with a slash.

Maybe the better format would be to use parenthesis e.g. `## update (u, upgrade)`, but that's up to @Seldaek to decide.

Both TODOs will be done in separate PRs - if this one goes through.


Seldaek:
Thanks, makes sense and having CI for the docs so we don't forget to add anything would be a great step forward 👍🏻 I'm not sure if you can check the description but at least checking for presence of the flags and commands would be awesome.

PR #11074, opened 2022-09-21
Load plugin commands for autocompletion

susannemoog:
To allow autocompletion for commands provided by plugins, the plugin commands should also be loaded in the completion step.

This results in both the completion for the command name itself as well as - if configured - completion for arguments and values for said command.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -237,7 +237,7 @@ class Application extends BaseApplication
                 || in_array($commandName, ['', 'list', 'help'], true)
                 // autocompletion requires plugin commands but if we are running as root without COMPOSER_ALLOW_SUPERUSER
                 // we'd rather not autocomplete plugins than abort autocompletion entirely, so we avoid loading plugins in this case
-                || $commandName === '_complete' && !$isNonAllowedRoot
+                || ($commandName === '_complete' && !$isNonAllowedRoot)
             );
 
         if ($mayNeedPluginCommand && !$this->disablePluginsByDefault && !$this->hasPluginCommands) {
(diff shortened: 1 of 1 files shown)


src/Composer/Console/Application.php
```diff
@@ -234,7 +234,7 @@ public function doRun(InputInterface $input, OutputInterface $output): int
                 // not a composer command, so try loading plugin ones
                 false === $commandName
                 // list command requires plugin commands to show them
-                || in_array($commandName, ['', 'list', 'help'], true)
+                || in_array($commandName, ['', 'list', 'help', '_complete'], true)
```
Seldaek, line 237:
```suggestion
                || in_array($commandName, ['', 'list', 'help'], true)
                // autocompletion requires plugin commands but if we are running as root without COMPOSER_ALLOW_SUPERUSER
                // we'd rather not autocomplete plugins than abort autocompletion entirely, so we avoid loading plugins in this case
                || $commandName === '_complete' && !$isNonAllowedRoot
```

Seldaek: I hope this makes sense?

Your change absolutely does for non-root usages.

src/Composer/Console/Application.php
```diff
@@ -235,6 +235,9 @@ public function doRun(InputInterface $input, OutputInterface $output): int
                 false === $commandName
                 // list command requires plugin commands to show them
                 || in_array($commandName, ['', 'list', 'help'], true)
+                // autocompletion requires plugin commands but if we are running as root without COMPOSER_ALLOW_SUPERUSER
+                // we'd rather not autocomplete plugins than abort autocompletion entirely, so we avoid loading plugins in this case
+                || $commandName === '_complete' && !$isNonAllowedRoot
```
Seldaek, line 240: Sorry forgot parenthesis here for completeness

```suggestion
                || ($commandName === '_complete' && !$isNonAllowedRoot)
```

Seldaek:
Thanks, merged as https://github.com/composer/composer/commit/b2ebe841e48d66512e72ec5dd6d50de8f91d153f in 2.4 branch

PR #11078, opened 2022-09-25
Test for About command

Jimlah:
Added test for about command


Seldaek:
Not sure why you closed, but seems ok to me? Thanks

Jimlah:
I saw the test coverage already covers the about command. That was why I closed it. But since this is dedicated for tge about command it should be fine too.

PR #11079, opened 2022-09-25
Added CheckPlatformReqsCommand Test

Jimlah:

### DIFF
--- a/tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
+++ b/tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
@@ -1,0 +1,151 @@
+<?php
+
+namespace Composer\Test\Command;
+
+use Composer\Test\TestCase;
+use Generator;
+use LogicException;
+
+class CheckPlatformReqsCommandTest extends TestCase
+{
+    /**
+     * @dataProvider provideCheckPlatformReqs
+     * @param array<array<string>> $composerJson
+     * @param array<boolean> $command
+     * @param boolean $lock
+     * @param string $output
+     * @return void
+     */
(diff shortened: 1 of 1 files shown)


tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
```diff
@@ -0,0 +1,151 @@
...
+        yield 'by default without vendor folder' => [
+            'composerJson' => [
+                'require' => [
+                    'first/pkg' => '^2.0',
+                    'second/pkg' => '3.*',
+                ],
+                'require-dev' => [
+                    'dev/pkg' => '~2.0',
+                ],
+            ],
```
Seldaek, line 121: This is lots of duplication and cruft as these are entirely unnecessary to run this test. Would be nice to reduce duplication at least.

tests/Composer/Test/Command/CheckPlatformReqsCommandTest.php
```diff
@@ -0,0 +1,151 @@
...
+        yield 'by default with vendor folder' => [
+            'composerJson' => [
+                'require' => [
+                    'first/pkg' => '^2.0',
+                    'second/pkg' => '3.*',
+                ],
+                'require-dev' => [
+                    'dev/pkg' => '~2.0',
+                ],
+            ],
```
Seldaek, line 55: Lots of duplication here too, and it doesn't seem to provide much value as you only check the first line of output.

It'd be more valuable to actually check behavior i.e. make sure that the installed packages above in testCheckPlatformReqs have some requirements on platform and then check that the output is pass or fail depending if it's a valid requirement or not.

Jimlah: Okay I will work on that.

PR #11092, opened 2022-09-30
Add test and documentation regarding audit command when no packages are required

gfaugere:
Related to #10796

I wanted to add tests to commands, and my thought was to add test to interesting behaviours that were worth validating. As far as I can tell, most code in the commands is related to output formatting or calling other components which implements the behaviours.

I found the audit command, which has a successful response code when no packages are required, and thought that it was a behaviour worth documenting/testing

Let me know what you think!

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1041,6 +1041,8 @@ for possible security issues. It checks for and
 lists security vulnerability advisories according to the
 [Packagist.org api](https://packagist.org/apidoc#list-security-advisories).
 
+_The audit command returns successfully (0) when no packages are required in the project._
+
 ```shell
 php composer.phar audit
 ```
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1041,7 +1041,7 @@ for possible security issues. It checks for and
 lists security vulnerability advisories according to the
 [Packagist.org api](https://packagist.org/apidoc#list-security-advisories).
 
-_The audit command returns successfully (0) when no packages are required in the project._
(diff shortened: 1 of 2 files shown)


doc/03-cli.md
```diff
@@ -1041,6 +1041,8 @@ for possible security issues. It checks for and
 lists security vulnerability advisories according to the
 [Packagist.org api](https://packagist.org/apidoc#list-security-advisories).
 
+_The audit command returns successfully (0) when no packages are required in the project._
+
```
Seldaek, line 1045: Tweaking this to be more generic because the case of no dependencies is not the most interesting as it's rare and anyway why would you audit the project deps if you have no deps :)

```suggestion
The audit command returns the amount of vulnerabilities found. `0` if successful, and up to `255` otherwise.

```

tests/Composer/Test/Command/AuditCommandTest.php
```diff
@@ -0,0 +1,28 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Test\Command;
+
+use Composer\Test\TestCase;
+
+class AuditCommandTest extends TestCase
+{
+    public function testSuccessfullResponseCodeWhenNoPackagesAreRequired(): void
```
Seldaek, line 19: Typo

```suggestion
    public function testSuccessfulResponseCodeWhenNoPackagesAreRequired(): void
```

Seldaek:
Thanks

PR #11094, opened 2022-10-01
#10796 Add test for ExecCommand

LeoVie:
This PR belongs to #10796
This increases the line coverage of ExecCommand to 80%.
At the moment this test only affects the listing of binaries, but not the execution. 

grafik


LeoVie:
Hmm, I don't really get what's the problem with phpstan here.

Seldaek:
The phpstan 8.1 build runs with latest phpstan so sometimes it fails due to newly reported errors, they can be ignored sorry for the noise.

Anyway thanks for the PR!

PR #11096, opened 2022-10-03
Typofixes

nexxai:


nexxai:
I do not believe the PHPStan failures have anything to do with my changes

Seldaek:
Thanks

PR #11105, opened 2022-10-09
Added new `LicensesCommandTest` class

greew:
Added a new LicensesCommandTest which gives 100% coverage on `Composer\Command\LicensesCommand`.


Seldaek:
Thanks

PR #11106, opened 2022-10-09
Added lots of new tests for `InitCommand`

greew:
This commit adds tests for all the different input options.


Seldaek:
Great work, thanks

Issue #11102, opened 2022-10-06
Why does the autoloader `require` instead of `require_once`?


wgevaert:
# Description of the problem:

When using the [wikimedia merge plugin](https://github.com/wikimedia/composer-merge-plugin) I ran into an [issue](https://github.com/wikimedia/composer-merge-plugin/issues/247) related to some files being loaded more than once.

I re-created the issue with a minimal composer.json and dependency below. 

I could "fix" the problem when in `./vendor/composer/autoload_real.php` I replaced the following lines:
```diff
 /**   
  * @param string $fileIdentifier
  * @param string $file
  * @return void
  */
 function composerRequire1e2b6faf2727c3fa0dd77ff8c2b47883($fileIdentifier, $file)
 {
     if (empty($GLOBALS['__composer_autoload_files'][$fileIdentifier])) {
         $GLOBALS['__composer_autoload_files'][$fileIdentifier] = true;
        
-        require $file;
+        require_once $file;
     }
 }
```

(Of course, editing this file is bad practice, and I don't usually do that).

So my questions/issues are: 
1. Is there a use case for this line to have `require` instead of `require_once`?
2. Could that line be changed to `require_once`?
3. Do you have any advise for solving the [original issue](https://github.com/wikimedia/composer-merge-plugin/issues/247)?

---------------
# Information you requested:
My `composer.json`:
```json
{
  "name": "test/test",
  "require": {
    "test/dependency": "@dev"
  },
  "repositories": [{
    "type": "path",
    "url": "./test-dependency"
  }],
  "autoload": {
    "files": [
      "test-dependency/Test.php"
    ]
  }
}
```
My `test-dependency/composer.json`:
```json
{
        "name": "test/dependency",
        "autoload": {
                "files": [
                        "Test.php"
                ]
        }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
require.test/dependency : unbound version constraints (@dev) should be avoided
Checking platform settings: OK
Checking git settings: OK git version 2.30.1
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
[...]

PR #11109, opened 2022-10-11
Add array_unique call to prevent loading the same file multiple times.

wgevaert:
The [composer json example I showed in an issue on this project](
https://github.com/composer/composer/issues/11102#issuecomment-1269867978) runs into a problem where it loads the same file multiple times.

This patch will not generate such an autoload file, by ensuring these files are unique.

# Example setup:
`composer.json`:
```json
{
    "name": "test/test",
    "require": {
        "wikimedia/composer-merge-plugin": "@dev"
    },
    "config": {
        "allow-plugins": {
            "wikimedia/composer-merge-plugin": true,
            "composer/installers": true
        }
    },
    "extra": {
        "merge-plugin": {
            "include": [
                "composer.local.json"
            ],
            "recursive": true,
            "merge-dev": false
        }
    }
}
```
`composer.local.json`:
```json
{
  "require": {
    "mediawiki/user-functions": "^2.7"
  },
  "extra": {
    "merge-plugin": {
      "include": [
        "extensions/*/composer.json"
      ]
    }
  }
}
```

# Problem without this patch:
The file `vendor/composer/autoload_files.php` that is generated looks like
```php
<?php

// autoload_files.php @generated by Composer

$vendorDir = dirname(__DIR__);

return array(
    '7bb4d66942ef8c1728c8f742a0d5639d' => $baseDir . '/extensions/UserFunctions/UserFunctions.php',
    '3661fc4b5b3581d0c7a3aa2f63bb491b' => $baseDir . '/extensions/UserFunctions/UserFunctions.php',
);
```
which results in an error as described in [this issue](
https://github.com/composer/composer/issues/11102).

# Result after this patch:
The file `vendor/composer/autoload_files.php` looks like

```php
<?php

// autoload_files.php @generated by Composer

$vendorDir = dirname(__DIR__);

return array(
    '7bb4d66942ef8c1728c8f742a0d5639d' => $baseDir . '/extensions/UserFunctions/UserFunctions.php',
);
```


physikerwelt:
@Seldaek, what do you think about this PR? It works for me locally, but I am a bit concerned about using a custom composer version in production.

wgevaert:
I believe the consensus on this pull request is that it should not be merged, because of [this reason](https://github.com/composer/composer/issues/11102#issuecomment-1269795826). Instead, you should prevent multiple inclusions of the same file. e.g. by including every package only once.

wgevaert:
> @Seldaek, what do you think about this PR? It works for me locally, but I am a bit concerned about using a custom composer version in production.

I believe they do not want this pull request to be merged, as seen in [this reply](https://github.com/composer/composer/issues/11102#issuecomment-1276085657).

The recommended solution (for now) is NOT to use the following in your `composer.local.json`:
```
{
  "require": {
    "...": "..."
  },
  "extra": {
    "merge-plugin": {
      "include": [
        "extensions/*/composer.json"
      ]
    }
  }
```

But instead, use something like this:
```
{
  "require": {
    "...": "..."
  },
  "extra": {
    "merge-plugin": {
      "include": [
        "extensions/ExtensionOne/composer.json"
        "extensions/ExtensionTwo/composer.json"
        "extensions/ExtensionThree/composer.json"
      ]
    }
  }
```
Where you only include those composer.json files that are not automatically parsed already by composer.

See also [this issue](https://github.com/wikimedia/composer-merge-plugin/issues/247) for the merge plugin and [this comment](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5479#issuecomment-1604045185) on an issue related to this on SemanticMediaWiki.

wgevaert:
> Is there also a patch for the merge-plugin? 

Yes, I made [this pull request](https://github.com/wikimedia/composer-merge-plugin/pull/251) some time ago, which I believe fixes this particular issue.

> In general, I think it is a good idea to ensure that there are no duplicates in the autoload file.

Composer already tries very hard to make sure there are no duplicates, see [this comment](https://github.com/composer/composer/issues/11102#issuecomment-1270058822):
Every file has as the key the value `md5( package/name:path/to/file )`, so the only way a file is included twice is when two different packages attempt to include the same file. This is a very very weird situation: Usually, composer puts every package in the folder `vendor/package/name/`, and thus only `md5(package/name:path/to/file)` should include that file, and not `md5(other/package:path/to/file)`. It is very weird if two packages try to include the same file.

I hope the patch for the merge plugin fixes your issues.

physikerwelt:
> Composer already tries very hard to make sure there are no duplicates, see [this comment](https://github.com/composer/composer/issues/11102#issuecomment-1270058822): Every file has as the key the value `md5( package/name:path/to/file )`, so the only way a file is included twice is when two different packages attempt to include the same file. This is a very very weird situation: Usually, composer puts every package in the folder `vendor/package/name/`, and thus only `md5(package/name:path/to/file)` should include that file, and not `md5(other/package:path/to/file)`. It is very weird if two packages try to include the same file.

I would have said, try harder, which is what this PR does;-) However, either way is fine with me.
> 
> I hope the patch for the merge plugin fixes your issues.

It does. 

So, if I understand you right, this PR can be closed?

wgevaert:
> So, if I understand you right, this PR can be closed?

I still believe either this PR or the merge-plugin pr should be merged, so I'll leave this open until one of them is merged.

I still think this PR adds some extra security for very little performance loss, so I don't think I want to close it just yet, although it seems quite clear that this will not be merged.

physikerwelt:
I think the merge fix looks good and will be merged soon (maybe WMF wants to be conservative and add a feature flag).

The only issue I see with this patch is that it silently fixes problems occured somewhere else. Maybe adding a warning on duplicate values would resolve the concerns by the package maintainers. I hope after summer break we will get some feedback.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -665,7 +665,18 @@ function (string $functionFile) use ($filesystem, $basePath, $vendorPath): strin
             },
             $files
         );
+        $oldFiles = $files;
         $files = array_unique($files);
+        if ( count( $files ) < count( $oldFiles ) ) {
+            $dupedFiles = array_diff_assoc( $oldFiles, $files );
+            $this->io->writeError(
+                '<warning>'
+                . 'Some files are included multiple times, not including these twice in autoloader: '
```
physikerwelt, line 674: There is a mix of "multiple" and "twice" in the message. 

Files have been requested multiple times for inclusion. This might indicate a problem in the configuration. Affected files:

Seldaek:
I'd rather not merge this because this is a misconfiguration which should not happen, and if a plugin decides to break things the fix and deduplication should be done in that plugin IMO.

physikerwelt:
I can understand this way if thinking. Instead of a warning one could also error out. However, I think generating autoload files that can not work should be avoided and discovered as early as possible.

Seldaek:
That depends really, many files which are meant for files autoloading support being included many times without problem by doing conditional defining of functions for example. Also if we error out it would probably prevent a plugin from resolving that situation after the autoload dump is complete.

physikerwelt:
I understand there are situations when no problems occur if one requires the same file twice. I would guess that this is, however, not the majority of packages. However, I am primarily a simple PHP user who found it difficult and time-consuming to trace back the problem through different layers way down to composer itself. If anything can be done to simplify it for future people who have the same problem, I would appreciate it, even if I better understand now that composer does not feel guilty:-) Users appreciate robust products a lot.

stof:
@Seldaek maybe composer could detect that case and display a warning message (not an exception stopping everything). this way, the user would be aware that they have an unexpected state (which might be due to a bad plugin)

tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
```diff
@@ -369,6 +378,31 @@ public function testRootPackageAutoloadingWithTargetDir(): void
...
+        $this->fs->ensureDirectoryExists($this->workingDir.'/src');
+        $this->fs->ensureDirectoryExists($this->workingDir.'/lib');
+
+        file_put_contents($this->workingDir.'/foo.php', '<?php class FilesFoo {}');
+        file_put_contents($this->workingDir.'/bar.php', '<?php class FilesBar {}');
+
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', false, 'FilesWarning');
+        self::assertFileContentEquals(__DIR__.'/Fixtures/autoload_files_duplicates.php', $this->vendorDir.'/composer/autoload_files.php');
+        $expected = '<warning>The following "files" autoload rules are included multiple times, this may cause issues and should be resolved:</warning>'.PHP_EOL.
+            '<warning> - $baseDir . \'/foo.php\'</warning>'.PHP_EOL;
```
stof, line 402: should we really output the PHP code in the warnings ?

Seldaek: I was also wondering about this.. but I think at least it is clear what dir this is, otherwise you'd just see "./foo" and we don't have the reference to the package or anything.. So I find this is maybe clearer.

Seldaek:
Alright merged now as a simple warning, to help catch this early. Thanks everyone

PR #11117, opened 2022-10-13
Improve ArrayLoader resilience against non-string reasons

ralflang:
I have seen this happen in some package with the latest satis build The package was admittedly abusing the reason field with a version 3.0. This would deserialize to a float value which in turn crashes trim(). 

Arguably, we could improve resilience even further by ensuring the value is actually stringable - but I did not want to overdo it.


Seldaek:
`'3.0' === json_decode('"3.0"')` https://3v4l.org/ARkcj

So no, json may not produce non string reasons IMO. Malformed composer.json may contain `"suggest": {"foo": 3.0}` but this is caught by composer's validation:

```
   - suggest.foo : Double value found, but a string is required
```

So it is the responsibility of the repo to provide good data types here IMO.

ralflang:
Understood. I will try to catch this on another level.

PR #11126, opened 2022-10-14
GH Actions: fix use of deprecated `set-output`

jrfnl:
GitHub has deprecated the use of `set-output` (and `set-state`) in favour of new environment files.

This commit updates workflows to use the new methodology.

Refs:
* https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files


Seldaek:
Thanks

jrfnl:
yvw

PR #11130, opened 2022-10-16
Feature: Add support for autocompleting setting-keys in config-command

brandung-sjorek:
This PR implements the support for autocompletion of setting-keys in the config-command and its related functional tests. The implementation takes existing command-line options like `--unset` or `--global` into account, as well as existing exposable package-properties. The result is a context-sensitive auto-completion with individual lists of setting-keys.

Additionally showing the value of `suggest.` sub-keys has been allowed in the config-command, as setting and unsetting them has always been possible, and I did not want to distinguish between showing, setting and unsetting those keys.

In the setting-key flattening-method I used the `array_is_list()` function, which was added in PHP 8.1 recently. For backwards PHP-compatibility the package `symfony/polyfill-php81` has been added and the corresponding test-fixtures have been fixed accordingly.

Fixes: #11131 
Signed-off-by: Stephan Jorek <stephan.jorek@brandung.de>


Seldaek:
Thanks, looking good!

brandung-sjorek:
> Thanks, looking good!

Thank you!

PR #11142, opened 2022-10-21
Add support for a tmp-dir config item

coudenysj:
This can help speed up installation on NFS volumes by fetching, unpacking and cleaning files on a local disk.

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -311,6 +311,7 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
             $this->config->get('vendor-dir').'/'.explode('/', $package->getPrettyName())[0],
             $this->config->get('vendor-dir').'/composer/',
             $this->config->get('vendor-dir'),
+            $this->config->get('tmp-dir'),
         ];
 
         if (isset($this->additionalCleanupPaths[$package->getName()])) {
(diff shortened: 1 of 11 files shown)


src/Composer/Downloader/FileDownloader.php
```diff
@@ -311,6 +311,7 @@ public function cleanup(string $type, PackageInterface $package, string $path, ?
             $this->config->get('vendor-dir').'/'.explode('/', $package->getPrettyName())[0],
             $this->config->get('vendor-dir').'/composer/',
             $this->config->get('vendor-dir'),
+            $this->config->get('tmp-dir'),
```
Seldaek, line 314:
```suggestion
            $this->config->get('tmp-dir'),
            $this->config->get('vendor-dir'),
```

Seldaek:
I'm not so keen on calling this tmp dir because it is actually quite specific to installation and extraction of packages. I mean right now we don't have another temp dir concept but maybe we will need that for other things too in the future, and taking that name now seems wrong.

I don't know what I'd call it tho.. extract-dir, extract-tmp-dir? It's not 100% accurate either but for most intents and purposes this works..

@naderman any better idea?

naderman:
May indeed be confusing if something else then actually uses the system tmp dir and users would expect this setting to override that, so extract-tmp-dir works for me.

Seldaek:
OK then @coudenysj if you don't mind renaming all the tmp-dir into extract-tmp-dir I think it'd be mergeable after 👍🏻

coudenysj:
Thanks @Seldaek and @naderman for the reviews. I'm still testing the performance improvements, because our initial testing with this fork don't show significant improvements :(.

Seldaek:
Ah ok in that case closing would also be a decent option for me :)

I imagine writing the files once is what takes time, writing the zip archive initially probably doesn't take much time, and extracting + moving or moving from temp dir amounts to about the same (each file has to be written once, you just save a move of the whole directory but that's fast)

Seldaek:
Any updates here?

coudenysj:
I'll close it, because we don't see the expected improvements.

PR #11150, opened 2022-10-25
Adds test coverage to the Fund Command

theoboldalex:
This PR adds test coverage to the fund command bringing coverage from 7% to 82%.


Seldaek:
Thanks, I added one test and tweaked the rest to avoid depending on packagist.org, but otherwise it looked good 👍🏻

PR #11151, opened 2022-10-26
Add support for adding Command classes as scripts, 

### LINKED ISSUE #11134: Feature request: integrate custom, complete Symfony Console scripts
It's possible to include [simple CLI scripts into Composer's `scripts`/`scripts-descriptions` list](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands).
It's also possible to [point static methods/functions to run as complex-ish scripts](https://getcomposer.org/doc/articles/scripts.md#defining-scripts).
Lastly, under the hood, Composer uses the (great) Symfony Console library.

Thus, is there any chance it's possible to include complete Symfony Console scripts into the main list of scripts a project uses, without having to duplicate command names and descriptions into `composer.json`?

I see it's possible to include custom commands from plugins, but what about doing that directly from the main project? It's an overshoot to create a whole plugin to include commands to manage a single project (but it still looks better than the current flow of having a separate executable for that lol).

What I'm thinking is something like a `scripts-folder` config entry which will import all classes found under it into the project's composer executable, much like `scripts` does with CLI commands. The end result is being able to seamlessly see our (complex) custom scripts when running `composer list`, including the description and help content configured in a subclass of `\Symfony\Component\Console\Command\Command`.

Example:
I'm writing some i18n scripts to extract/upload/download translations, and that's best done from a CLI tool such as Symfony Console, which allows for processing of arguments, options, and console I/O. In the meantime, we already have a bunch of one-liners as Composer scripts (such as clear DB cache, compile frontend, migrations work) which work great being defined directly in `composer.json`.

Seldaek:
Fixes #11134

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -265,12 +265,16 @@ class EventDispatcher
                         continue;
                     }
                     if (defined('Composer\Script\ScriptEvents::'.str_replace('-', '_', strtoupper($event->getName())))) {
-                        $this->io->writeError('<warning>You cannot bind '.$event->getName().' to a Command class, use your own script name.</warning>', true, IOInterface::QUIET);
+                        $this->io->writeError('<warning>You cannot bind '.$event->getName().' to a Command class, use a non-reserved name</warning>', true, IOInterface::QUIET);
                         continue;
                     }
 
                     $app = new Application();
                     $cmd = new $className($event->getName());
+                    if ($cmd->getName() !== $event->getName()) {
+                        $this->io->writeError('<warning>Class '.$className.' must have '.$event->getName().' as Command name (or skip defining a name altogether)</warning>');
+                        continue;
+                    }
                     $app->add($cmd);
                     try {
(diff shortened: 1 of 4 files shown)


src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -249,6 +254,36 @@ protected function doDispatch(Event $event)
...
+                    if (!is_a($className, Command::class, true)) {
+                        $this->io->writeError('<warning>Class '.$className.' does not extend '.Command::class.', can not call '.$event->getName().' script</warning>', true, IOInterface::QUIET);
+                        continue;
+                    }
+                    if (defined('Composer\Script\ScriptEvents::'.str_replace('-', '_', strtoupper($event->getName())))) {
+                        $this->io->writeError('<warning>You cannot bind '.$event->getName().' to a Command class, use a non-reserved name</warning>', true, IOInterface::QUIET);
+                        continue;
+                    }
+
+                    $app = new Application();
```
stof, line 272: should the application catch exceptions to render them in the output, or should it be disabled to allow the outer code to handle them ?)

Seldaek: Yup, fixed thanks

Issue #8272, opened 2019-08-11
[Composer/Packagist.org] Mark package versions as being maintained


Toflar:
I would like to propose a way to improve two things:

* give users the option to understand what versions of a library are actually supported
* massively reduce the size of the ruleset when resolving dependencies

I found https://github.com/composer/packagist/issues/938 which is similar but it requires action on every single package version.
I think we could introduce an option on packagist.org where package maintainers can (totally optionally) configure which versions are being maintained and which ones are not. I think this can be one single version constraint on package level so managing and maintaining it should be pretty easy.
In case of Symfony for example, that would look like this as of today:

`2.8.* || 3.4.* || 4.2.* || 4.3.*`

One positive effect this would have is that I could visit packagist.org and see immediately which versions are supported and which ones are not. We could even introduce some command that would display what package versions are not supported anymore. 
On the other hand, we could have a new flag to `composer update` which would limit the provider information to only the maintained versions. Running `composer update --maintained-only` would then greatly reduce the number of rules that are added to the ruleset and thus reduce memory usage and speed up the performance massively.
In fact it's similar to what Symfony Flex (`extra.require` section) does but it's not specific to Symfony, doesn't require any knowlege about the maintained versions by the user and works for transitive dependencies as well.
As a user you can simply decide to only include maintained versions into your dependency graph by setting the `--maintained-only` or `-mo` flag.
In case a package maintainer did not specify anything, we'd just fall back to the current behaviour, including all versions of that package.
When you run `composer update -mo` you'll probably not get a resolvable set of dependencies because you're using outdated package versions. In that case, you can still omit the `-mo` flag and have the current behaviour.

Any feedback on this? Would it be hard to implement and why?

/cc @nicolas-grekas @leofeyer

Toflar:
I also think the "maintained package versions only" mode should become default in version 2 and instead of an `--include-non-maintained-versions` flag which would again load half of the Internet, exceptions should be added explicitly to the `composer.json`. Let's say I'm using `symfony/filesystem` in version `4.1.*` which is an outdated version. I then have two options:

1. Update to a maintained version (of course recommended)
2. Add an explicit exception to the `composer.json` like e.g.
```
"maintenance-exceptions": {
    "symfony/filesystem": "4.1.*"
}
```

All these exception packages would be loaded into the resolving process explicitly plus then the regular resolving process with the maintained packages only starts.
[...]

PR #11158, opened 2022-10-27
Also bump match all constraint

Toflar:
Currently `composer bump` ignores `*` constraints.
I think this is inconsistent because the way I understand the bump command (maybe I'm wrong and that's the issue? :D), it's supposed to ensure that it does not downgrade any already existing package which would currently happen for `*` requirements.

### DIFF
--- a/src/Composer/Package/Version/VersionBumper.php
+++ b/src/Composer/Package/Version/VersionBumper.php
@@ -18,6 +18,7 @@ use Composer\Package\Dumper\ArrayDumper;
 use Composer\Pcre\Preg;
 use Composer\Semver\Constraint\Constraint;
 use Composer\Semver\Constraint\ConstraintInterface;
+use Composer\Semver\Constraint\MatchAllConstraint;
 use Composer\Semver\Intervals;
 use Composer\Util\Platform;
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Yeah I am not so sure here. I'd say if you explicitly put `*` in you are kinda on purpose giving the finger to proper constraints and you most likely don't want it bumped? But that's just how I'd see this..

Toflar:
Not using `*` myself. Let me give you some background, maybe we can find a better solution.
Over at Contao, our regular users (so not the developers of packages but the users) are not familiar with the versioning strategy of all the packages. They ususally don't know which exact version constraint they need. What they need is this:

> "Hello Composer, I want the best version of `<x>` that fits the rest of my current dependencies".

Let's say you have Contao 4.9 installed now and there's an extension in version `1.2.*` compatible to it because the developer decides to maintain the `1.2.*` branch. However, there's also a new minor `1.3.*` development branch which is meant for newer Contao versions. If you run `composer req <x>` without version constraint today, you get no valid solution because Composer would try `PackageSorter::getMostCurrentVersion($this->getRepos()->findPackages($name))` which would translate to some `1.3.*` in our example. 
There's only one way to force Composer to take the latest possible which is to try to install any version (`composer req <x>:*`) in order for it to be able to install `1.2.0`. This would be the best option from a UX perspective for us.

It's actually not related to being a regular user or developer or Contao whatsoever. Maybe we should just default to `composer require` without version constraint to use `*` rather than the most current version? 

Whether or not the version you get is maintained, is a whole different question: https://github.com/composer/composer/issues/8272

PR #11159, opened 2022-10-27
Fix output inconsistencies when aliases are being used when running 'composer install'

PrinsFrank:
When running `composer install` in a state where all packages are already installed, the following output will be shown:

```
Generating autoload files
61 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
```

When there is any dependency that relies on an alias, a different output will be shown however:
```diff
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
- Nothing to install, update or remove
+ Package operations: 0 installs, 0 updates, 0 removals
Generating autoload files
61 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
```
In the operations array of the Transaction class, besides the InstallOperation, UpdateOperation and UninstallOperation, there can also be MarkAliasInstalledOperations. If this is the case and no install/update/uninstalls are being done, the different message is show because the entire operations array is checked for emptyness instead of checking if one of the modifying operations have been executed. This PR fixes this inconsistency, so even when there are dependencies aliased the 'Nothing to install, update or remove' message is shown.

This is a trivial change resulting in a slight output changes, but makes it easier to asses the output of `composer install --dry-run` to check if the current installed set of package is correct, for example in [beyondcode/laravel-self-diagnosis](https://github.com/beyondcode/laravel-self-diagnosis/blob/e6d36fea90f8d04bc8729f21a9ef7085c3d1b52a/src/Checks/ComposerWithoutDevDependenciesIsUpToDate.php#L46)

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -751,22 +751,20 @@ class Installer
         $localRepoTransaction = new LocalRepoTransaction($lockedRepository, $localRepo);
         $this->eventDispatcher->dispatchInstallerEvent(InstallerEvents::PRE_OPERATIONS_EXEC, $this->devMode, $this->executeOperations, $localRepoTransaction);
 
-        if (!$localRepoTransaction->getOperations()) {
-            $this->io->writeError('Nothing to install, update or remove');
+        $installs = $updates = $uninstalls = [];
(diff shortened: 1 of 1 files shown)


Seldaek:
Good catch, thanks

PR #11160, opened 2022-10-28
Make the require command guess versions more accurately

Seldaek:
If no version is provided we usually select the latest working version but in some cases this leads to problems if that conflicts with other stuff you have installed.

This PR makes it instead use a `*` constraint temporarily, effectively delegating the job to the solver (except if called with --no-update then we still use the best-effort version selector stuff). Once we have a solution we then replace version constraints again with whatever the locked version is.

This should improve the `require` experience especially for people that do not have very up to date dependencies.


naderman (approved):
Looks great!

Toflar (approved):
Yes! <3

PR #11163, opened 2022-10-31
[changelog] Add 2.2.18

captn3m0:
https://getcomposer.org/changelog/2.2.18 was returning a 404, this should fix it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,6 +186,13 @@
   * Fixed symlink creation in linux VM guest filesystems to be recognized by Windows (#10592)
   * Performance improvement in pool optimization step (#10585)
 
+### [2.2.18] 2022-08-20
+
+  * Fixed `COMPOSER_NO_DEV` so it also works with require and remove's `--update-no-dev` (#10995)
+  * Fixed duplicate missing extension warnings being displayed (#10938)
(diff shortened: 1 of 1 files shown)


usta:
@Seldaek ping ?

captn3m0:
Could this be merged?

Seldaek:
Is it really so urgent?

I am not super happy with the solution as it'd mean having to backport every changelog change from 2.2 branch into main.. I haven't yet had time to figure out a better solution tho, so it's in limbo.

stof:
@Seldaek the issue is that the download page of the website has a changelog link for each version. But the website uses the file from the main branch to read the changelog, so the link fails for 2.2.18 as the changelog only exists in the 2.2 branch.

stof:
The solution might be to merge 2.2 up into newer branches after releasing new 2.2 versions.

captn3m0:
Not really urgent, just wanted to follow up.

stof:
An option might be to make the changelog URL redirect to the GitHub release to see the changelog instead of parsing the changelog file of a given release.

Seldaek:
Yup, that's a good idea @stof - tho I fetched it from API instead of redirecting, but maybe it should redirect if the API call failed due to rate limit just for safety ;)

PR #11166, opened 2022-10-31
Allow the 'composer remove --unused' command to run in non-interactive mode

PrinsFrank:
Currently, the command 'composer remove --unused' doesn't work in non-interactive mode. This prevents it from being run in github actions for example;

```
COMPOSER_NO_INTERACTION=true composer remove --unused
```

```
Error: Not enough arguments (missing: "packages").

                                               
  Not enough arguments (missing: "packages").  
                                               

remove [--dev] [--dry-run] [--no-progress] [--no-update] [--no-install] [--no-audit] [--audit-format AUDIT-FORMAT] [--update-no-dev] [-w|--update-with-dependencies] [-W|--update-with-all-dependencies] [--with-all-dependencies] [--no-update-with-dependencies] [--unused] [--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform-reqs] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] [--] <packages>...

Error: Process completed with exit code 1.
```

There is nothing actually interacting with a user in the command, as far as I can see it is only run in the interact method to circumvent the required validation of the packages argument. Instead of this, we can make the packages argument not required in the InputArgument definition and check if either the 'packages' argument is present or the '--unused' flag is given.

### DIFF
--- a/src/Composer/Command/RemoveCommand.php
+++ b/src/Composer/Command/RemoveCommand.php
@@ -20,6 +20,7 @@ use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginEvents;
 use Composer\Json\JsonFile;
 use Composer\Factory;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
 use Symfony\Component\Console\Input\InputInterface;
 use Composer\Console\Input\InputOption;
 use Composer\Console\Input\InputArgument;
@@ -44,7 +45,7 @@ class RemoveCommand extends BaseCommand
             ->setName('remove')
             ->setDescription('Removes a package from the require or require-dev')
             ->setDefinition([
-                new InputArgument('packages', InputArgument::IS_ARRAY | InputArgument::REQUIRED, 'Packages that should be removed.', null, $this->suggestRootRequirement()),
+                new InputArgument('packages', InputArgument::IS_ARRAY, 'Packages that should be removed.', null, $this->suggestRootRequirement()),
                 new InputOption('dev', null, InputOption::VALUE_NONE, 'Removes a package from the require-dev section.'),
                 new InputOption('dry-run', null, InputOption::VALUE_NONE, 'Outputs the operations but will not execute anything (implicitly enables --verbose).'),
                 new InputOption('no-progress', null, InputOption::VALUE_NONE, 'Do not output download progress.'),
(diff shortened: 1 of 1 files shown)


src/Composer/Command/RemoveCommand.php
```diff
@@ -117,24 +125,15 @@ protected function interact(InputInterface $input, OutputInterface $output)
...
             }
-            $input->setArgument('packages', array_merge($input->getArgument('packages'), $unused));
+            $packages = array_merge($packages, $unused);
 
-            if (count($input->getArgument('packages')) === 0) {
+            if (count($packages) === 0) {
                 $this->getIO()->writeError('<info>No unused packages to remove</info>');
                 $this->setCode(static function (): int {
                     return 0;
                 });
```
Seldaek, line 134: This can now be simplified:

```suggestion
                $this->getIO()->writeError('<info>No unused packages to remove</info>');

                return 0;
```

Seldaek:
Thanks

PrinsFrank:
For anyone ending up on this PR and wanting to use this feature before Composer 2.5 is released, you can fake a TTY to be present in github actions like this:
```shell
- name: Check for superfluous packages in Composer lock file
  shell: script -q -e -c "bash {0}" # Work around for no TTY available. This line can be removed after composer 2.5 has been released
  run: composer remove --unused --dry-run
```

src/Composer/Command/RemoveCommand.php
```diff
@@ -79,10 +80,17 @@ protected function configure()
...
      */
-    protected function interact(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if ($input->getArgument('packages') === [] && !$input->getOption('unused')) {
+            throw new InvalidArgumentException('Not enough arguments (missing: "packages")');
+        }
+
+        $packages = $input->getArgument('packages');
+        $packages = array_map('strtolower', $packages);
```
stof, line 92: This could even be simplified more by moving this logic after that `if ($input->getOption('unused')) {` block

Seldaek: I'm not sure I understand what could be simplified?

PR #11170, opened 2022-11-03
Use binary-only images for installing Composer

Wirone:
Document optimal way of copying Composer binary using `COPY --from` in Dockerfiles by using binary-only image.

See: https://blog.codito.dev/2022/11/composer-binary-only-docker-images/


Seldaek:
Thanks

PR #11171, opened 2022-11-04
Make possible return type in `IOInterface::select()` more explicit

eliashaeussler:
`IOInterface::select()` either returns any of `int`, `string`, `bool` or an array of `string` values. However, since the returned array is always a list, we can safely change the return type annotation to `list<string>`, making it more explicit in terms of static code analysis.

### DIFF
--- a/src/Composer/IO/IOInterface.php
+++ b/src/Composer/IO/IOInterface.php
@@ -191,7 +191,7 @@ interface IOInterface extends LoggerInterface
      * @param bool        $multiselect  Select more than one value separated by comma
      *
      * @throws \InvalidArgumentException
-     * @return int|string|string[]|bool     The selected value or values (the key of the choices array)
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

src/Composer/IO/IOInterface.php
```diff
@@ -191,7 +191,7 @@ public function askAndHideAnswer(string $question);
      * @param bool        $multiselect  Select more than one value separated by comma
      *
      * @throws \InvalidArgumentException
-     * @return int|string|string[]|bool     The selected value or values (the key of the choices array)
+     * @return int|string|list<string>|bool     The selected value or values (the key of the choices array)
```
staabm, line 194: I guess using generics and a conditional return types, this could be narrowed even further

staabm: lets see https://github.com/composer/composer/pull/11219

PR #11174, opened 2022-11-04
Check for unbounded constraints on 'php' & friends

fredden:
See https://github.com/composer/composer/pull/2320 where this feature was first introduced, and https://github.com/composer/composer/commit/e7b82cdd886f305cd5b86fbf52d80e2d9020c9fb where all "platform" constraints are ignored.

We run `composer validate` as part of our integration testing suite. This has helped a lot when developers attempt to introduce unbound version constraints for packages, but the command does not complain when a developer claims to support PHP version 12.4.8 (eg `"require": { "php": ">=7.4", ...`). There's no way that the developer has considered all the breaking changes between version 9..10, let alone all through version 12.4 (since none of these exist yet). I was surprised that `composer validate` was not alerting the developer to this problem.

This pull request adjusts the 'unbound constraint' checks to include most platform packages. We still excludes/ignore `ext-` packages, as these do not typically have sensible version numbers.
I've also increased the test coverage by requiring the 'success' test to report no warnings.


fredden:
The static test failures for PHPStan with PHP v8.1 and the latest composer dependencies are unrelated to the changes made in this pull request. I have opened #11176 to fix those.

Seldaek:
Thanks for the good PR (from a purely code perspective:). I disagree this is a good change though.

Excluding `lib-*`/`ext-*` prefix makes sense, but I'd also exclude `php*` as well still because many projects declare their require like that, and while some disagree it's a good idea I think it's a choice I'd rather leave to package authors as while there are risks there are also benefits to keeping the door open there.

isPlatformPackage also includes `composer*` packages tho and those definitely should not be required with unbounded constraints so fixing that sounds good to me if you're willing to update the PR..

stof:
For PHP packages, it might be covered by a flag to include them or no, to give the choice to projects.

fredden:
The whole point of / motivator for this pull request was to check the `php*` packages for unbound constraints. I can certainly update the pull request when we've worked out what it should contain.

> many projects declare their require like that

This sounds like motivation to merge this pull request as-is, not to change it. ;)

> it's a choice I'd rather leave to package authors

Given this is a warning not an error, I would like to encourage package authors to follow best practices. We're not saying that the package is invalid, but that there is a "better" way of writing the constraints. Is this not the exact same logic for the unbound warning for other packages?

> it might be covered by a flag

I would be okay having a flag to optionally disable this check, but having the check off by default would not likely aid in adoption of / adherence to this recommendation.

src/Composer/Package/Loader/ValidatingArrayLoader.php
```diff
@@ -273,12 +273,13 @@ public function load(array $config, string $class = 'Composer\Package\CompletePa
                             continue;
                         }
 
-                        // check requires for unbound constraints on non-platform packages
+                        // check requires for unbound constraints, but ignore 'ext-*' and `lib-*` platform packages
                         if (
                             ($this->flags & self::CHECK_UNBOUND_CONSTRAINTS)
                             && 'require' === $linkType
+                            && strpos($package, 'ext-') !== 0
+                            && strpos($package, 'lib-') !== 0
```
stof, line 281: Something starting with `ext-` or `lib-` might not be a platform package. There is nothing preventing a non-platform package to have a vendor name starting with such prefix. So the `PlatformRepository::isPlatformPackage($package)` check cannot be entirely removed.

stof (changes requested):

Seldaek:
> Given this is a warning not an error, I would like to encourage package authors to follow best practices. We're not saying that the package is invalid, but that there is a "better" way of writing the constraints. Is this not the exact same logic for the unbound warning for other packages?

The issue is I don't necessarily agree it's a best practice to strictly semver your php requirement, and many others also don't. It causes huge pains when libs get outdated/stale even tho they might well still work fine and in most cases they do.

So no I'd rather not have this and I also don't want a flag for it sorry. People who think semver for php is best will already set their php requirement accordingly and those who don't shouldn't have to be nagged into setting a flag off IMO.

fredden:
> [differing opinions on] best practice to strictly semver your php requirement

My view is that the `php` requirement should be listed with an appropriate constraint (eg, semver), or omitted entirely.

I also don't want to create a lot of extra work, nor nag people to implement a flag to have `composer validate` behave "properly" when there's an uncertainty about what is a "good" constraint for a platform package.

> I'd rather not have this

I think we should probably close this pull request. We can use it as a reference for anyone wanting insight into why platform packages do not report warnings when unbound. I will look into checking for unbound constraints on the `composer*` platform packages separately.

PrinsFrank:
A thought here: Would it be an idea to set the unbound version range AND several bound versions that indicate the actual supported versions? Something like:
```
>=7.4 || ~7.4.0 || ~8.0.0
```

This would make a package installable with all future PHP versions, but indicates that 7.4 and 8.0 are currently supported and tested. Any depending projects can then check for implicit/explicit support for their new php version when doing an upgrade.

stof:
@PrinsFrank this does not indicate anything though. Your constraint is *strictly* equivalent to `>= 7.4` (and composer might even be able to optimize it that way.
If you expect devs to manually go check the exact syntax used for each constraint in the generated lock file, you could as well make them read all readmes or have a tool checking the CI setup of each of those packages. But it won't help you when asking Composer to do an update.

Seldaek:
Ok closing sounds fine to me. Thanks everyone for the civil discussion.

PR #11176, opened 2022-11-04
Fix some complaints from PHPStan

fredden:
When reviewing a recent pull request that I have opened, I noticed that there were some errors detected by PHPStan which are unrelated to that pull request. This pull request updates the baseline to cover these, and also fixes some of the complaints.


Seldaek:
Replaced by https://github.com/composer/composer/pull/11189 but thanks for the PR anyway!

PR #11182, opened 2022-11-11
Changing colors of success messages

ThomasLandauer:
IMO, green should only be used for *success* messages. So everything that's currently green (in the output of `composer outdated`) should be changed to some other (neutral) color (what about blue?) => But I don't know how to do this :-(


Seldaek:
I kinda disagree here, we use the info tag (and hence the green) to make some things stand out, except when they're errors/warnings type of things that need to stand out.

In this case I don't really see the value in switching to info, the headings are info and then we have a status/list of outdated in plain text

ThomasLandauer:
It's especially this message which I find irritating:

> Direct dependencies required in composer.json:  <= green
> Everything up to date <= white

IMO it should be the other way round.

Seldaek:
Yes I can see this particular case may appear backwards, but when you have outdated ones highlighting just the headings makes more sense IMO, and for consistency with other commands it also IMO is more sensible.

PR #11188, opened 2022-11-16
Allow storing credentials globally when 'local' auth.json file exists

PrinsFrank:
When in interactive mode with missing credentials, the credentials are always stored in the local auth.json file if it exists, even though credentials in the auth.json might be shared through git and the user wants to store their credentials in their global auth.json:

```diff
Your GitHub credentials are required to fetch private repository metadata (https://github.com/vendor/package)
When working with _public_ GitHub repositories only, head to https://github.com/settings/tokens/new?scopes=&description=MachineName+TIMESTAMP to retrieve a token.
This token will have read-only permission for public information only.
When you need to access _private_ GitHub repositories as well, go to https://github.com/settings/tokens/new?scopes=repo&description=MachineName+TIMESTAMP
Note that such tokens have broad read/write permissions on your behalf, even if not needed by Composer.
Tokens will be stored in plain text in "~/user/.composer/auth.json" for future use by Composer.
For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth
Token (hidden): 
```

This PR adds the ability to choose where to store this newly acquired token when such a local auth.json file exists before prompting for the token:

```diff
Your GitHub credentials are required to fetch private repository metadata (https://github.com/vendor/package)
When working with _public_ GitHub repositories only, head to https://github.com/settings/tokens/new?scopes=&description=MachineName+TIMESTAMP to retrieve a token.
This token will have read-only permission for public information only.
When you need to access _private_ GitHub repositories as well, go to https://github.com/settings/tokens/new?scopes=repo&description=MachineName+TIMESTAMP
Note that such tokens have broad read/write permissions on your behalf, even if not needed by Composer.
-Tokens will be stored in plain text in "~/user/.composer/auth.json" for future use by Composer.
+Tokens will be stored in plain text in "~/user/.composer/auth.json" OR "~/user/project/auth.json" for future use by Composer.
For additional information, check https://getcomposer.org/doc/articles/authentication-for-private-packages.md#github-oauth
-Token (hidden): 
+ A local auth config source was found, do you want to store the token there? [y]:
```

Also, in the Github and BitBucket authentication, existing keys were incorrectly removed from the composer.json file instead of the auth.json file. Those issues are also fixed in this PR.

### DIFF
--- a/src/Composer/Util/Bitbucket.php
+++ b/src/Composer/Util/Bitbucket.php
@@ -141,10 +141,16 @@ class Bitbucket
             $this->io->writeError($message);
         }
 
+        $localAuthConfig = $this->config->getLocalAuthConfigSource();
         $url = 'https://support.atlassian.com/bitbucket-cloud/docs/use-oauth-on-bitbucket-cloud/';
         $this->io->writeError(sprintf('Follow the instructions on %s', $url));
-        $this->io->writeError(sprintf('to create a consumer. It will be stored in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
+        $this->io->writeError(sprintf('to create a consumer. It will be stored in "%s" for future use by Composer.', ($localAuthConfig !== null ? $localAuthConfig->getName() . ' OR ' : '') . $this->config->getAuthConfigSource()->getName()));
         $this->io->writeError('Ensure you enter a "Callback URL" (http://example.com is fine) or it will not be possible to create an Access Token (this callback url will not be used by composer)');
+
+        $storeInLocalAuthConfig = false;
+        if ($localAuthConfig !== null) {
+            $storeInLocalAuthConfig = $this->io->askConfirmation('A local auth config source was found, do you want to store the token there?', true);
+        }
 
         $consumerKey = trim((string) $this->io->askAndHideAnswer('Consumer Key (hidden): '));
 
@@ -171,7 +177,8 @@ class Bitbucket
         }
 
         // store value in user config
-        $this->storeInAuthConfig($originUrl, $consumerKey, $consumerSecret);
+        $authConfigSource = $storeInLocalAuthConfig && $localAuthConfig !== null ? $localAuthConfig : $this->config->getAuthConfigSource();
+        $this->storeInAuthConfig($authConfigSource, $originUrl, $consumerKey, $consumerSecret);
 
         // Remove conflicting basic auth credentials (if available)
         $this->config->getAuthConfigSource()->removeConfigSetting('http-basic.' . $originUrl);
@@ -195,7 +202,7 @@ class Bitbucket
             return '';
         }
 
-        $this->storeInAuthConfig($originUrl, $consumerKey, $consumerSecret);
+        $this->storeInAuthConfig($this->config->getLocalAuthConfigSource() ?? $this->config->getAuthConfigSource(), $originUrl, $consumerKey, $consumerSecret);
 
         if (!isset($this->token['access_token'])) {
             throw new \LogicException('Failed to initialize token above');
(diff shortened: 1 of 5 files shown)


src/Composer/Util/GitHub.php
```diff
@@ -129,9 +135,10 @@ public function authorizeOAuthInteractively(string $originUrl, ?string $message
             throw $e;
         }
 
-        // store value in user config
-        $this->config->getConfigSource()->removeConfigSetting('github-oauth.'.$originUrl);
```
PrinsFrank, line 133: Previously, the github-oauth setting was removed from the composer.json, but this should be removed from the auth.json when updating credentials, right?

src/Composer/Util/Bitbucket.php
```diff
@@ -207,9 +214,9 @@ public function requestToken(string $originUrl, string $consumerKey, string $con
     /**
      * Store the new/updated credentials to the configuration
      */
-    private function storeInAuthConfig(string $originUrl, string $consumerKey, string $consumerSecret): void
+    private function storeInAuthConfig(Config\ConfigSourceInterface $authConfigSource, string $originUrl, string $consumerKey, string $consumerSecret): void
     {
-        $this->config->getConfigSource()->removeConfigSetting('bitbucket-oauth.'.$originUrl);
```
PrinsFrank, line 212: Previously, the bitbucket-oauth setting was removed from the composer.json, but this should be removed from the auth.json when updating credentials, right? (getConfigSource() vs getAuthConfigSource())

src/Composer/Util/GitHub.php
Seldaek: I don't think so no. The point of the removal was making sure we migrate auth stuff to the (then new) auth.json.

Removing it from auth is otherwise useless as writing to it overwrites the value anyway if it's present.

src/Composer/Util/Bitbucket.php
Seldaek: Same as above

src/Composer/Util/Bitbucket.php
```diff
@@ -141,11 +141,17 @@ public function authorizeOAuthInteractively(string $originUrl, ?string $message
...
+        $localAuthConfig = $this->config->getLocalAuthConfigSource();
         $url = 'https://support.atlassian.com/bitbucket-cloud/docs/use-oauth-on-bitbucket-cloud/';
         $this->io->writeError(sprintf('Follow the instructions on %s', $url));
-        $this->io->writeError(sprintf('to create a consumer. It will be stored in "%s" for future use by Composer.', $this->config->getAuthConfigSource()->getName()));
+        $this->io->writeError(sprintf('to create a consumer. It will be stored in "%s" for future use by Composer.', ($localAuthConfig !== null ? $localAuthConfig->getName() . ' OR ' : '') . $this->config->getAuthConfigSource()->getName()));
         $this->io->writeError('Ensure you enter a "Callback URL" (http://example.com is fine) or it will not be possible to create an Access Token (this callback url will not be used by composer)');
 
+        $storeInLocalAuthConfig = false;
+        if ($localAuthConfig !== null) {
+            $storeInLocalAuthConfig = $this->io->askConfirmation('A local auth config source was found, do you want to store the token there?', false);
```
Seldaek, line 152: Considering it was previously always writing to the local one, I think the default here should be `true` to ensure BC if non-interactive. Same for GitHub/GitLab.

src/Composer/Util/GitHub.php
PrinsFrank: Ah, sorry, I misunderstood.

src/Composer/Util/Bitbucket.php
PrinsFrank: The initial reason I added this PR was because we have a license key for a private satis server distributed in an auth.json file, but when developers get rate limited on the github api for example their oauth key gets added to the auth.json file and may now accidentally be comitted. Adding false as a default makes sense here IMHO, as it has to be a consious descision to add credentials to vcs. 

I do see your point though, as the auth.json file might not be in vcs for all Composer users. I've updated the default to false.

Seldaek:
Thanks!

PR #11154, opened 2022-10-26
Document recovering from invalid merges on composer.lock and composer.json

PrinsFrank:
Hi @Seldaek, It's been a while since my previous PR regarding documenting correctly resolving merge conflicts in #9386 and subsequently detecting missing packages after incorrect merges in #9899, but I now work at a different company and I am once again trying to recover from an incorrect merge without walking through months of git logs or removing the lock file. ;)

When walking through code from #9386 and trying to figure out if there is an easy way to check unused packages in the validate command, I noticed there was already code present for an undocumented feature added in #8786 where the remove command can be run with the '--unused' flag to remove unused packages.

Would you be open to use the code in the 'interact' method in the 'removeCommand' to also display errors when running 'composer validate' with packages that are locked but not required? I can open a PR for that.

Meanwhile, I decided to add some documentation about the --unused flag and recovering from incorrect merges. Do you agree with the content here?

PR #11195, opened 2022-11-23
Check missing required packages when running install and fail when there are any

PrinsFrank:
In #9899 a check was added to the validate command to check if there are any required packages that are not locked. According to the discussion in #11154, the install command should also fail on this state. It doesn't seem there is currently any code present in the Installer that checks this, so I am adding it in this PR. 

There are two things in here that I am not sure about:
- The logic is used in the Validate and Installer, so I moved it to a more generic place. I decided on the locker as the correct place, do you agree?
- The 'validate' command fails on missing locked packages in the require and require-dev section. But the installer can be run with the flag 'no-dev'. In the current implementation there are no errors displayed when running with that flag even when there are packages in de dev section that are not locked. Is that desired behaviour or should I also return errors there?


Seldaek:
Implementation looks reasonable and good to me, thanks!

rarkins:
FYI this change broke all Renovate Bot users, I have created an issue: https://github.com/composer/composer/issues/11232

nickygerritsen:
We have a scenario which breaks with this new check: we have a composer script that modifies versions when we are on a `develop` branch: on `main` we set the version to (for example) `1.0.0` but on develop we want to automatically use `develop` to keep developing. We created a Composer plugin for this.

However, during deploy we run `composer install --no-scripts` because we do not want to run any OTHER scripts (like running migrations via post-install-cmd). However, now the version in the '.json' does not match the .lock anymore.

Are we just doing something that is not supported or should there be a way to allow this?

Seldaek:
@nickygerritsen that sounds like an edge case that we don't necessarily want to cover, but maybe we can help figure out a solution. I'd suggest you open a new issue tho to avoid spamming everyone here, and please add a bit more details as to what you do, I don't yet exactly understand the develop/main/1.0.0 and where they fit in your workflow.

PR #11197, opened 2022-11-24
[Tests] Use static data providers

IonBazan:
Non-static data providers are deprecated since PHPUnit 10 and will be removed in v11.

Most helper methods in `TestCase` like `getCmd()`, `getVersionConstraint()`, `getPackage()`, etc became static too.


Seldaek:
Thanks!

Issue #10123, opened 2021-09-27
HgDownloader - clone fails with custom installation path (reincarnation of #2096)


sb-relaxt-at:
My `composer.json`:

```json
{
    "repositories": [
        {
            "type": "composer",
            "url": "https://internal-hg-satis.example"
        }
    ],
    "require": {
        "php": ">=7.1",
        "internal/silverstripe-module-required": "^2.0@dev"
    },
    "extra": {
            "installer-paths": {
                    "cms/": ["silverstripe/cms"],
                    "framework/": ["silverstripe/framework"],
                    "modules/{$name}/": ["type:silverstripe-module"]
            }
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
…
OK
Checking composer version: OK
Composer version: 2.1.6
PHP version: 7.1.33
PHP binary path: /usr/bin/php
OpenSSL version: LibreSSL 2.6.5
cURL version: 7.54.0 libz 1.2.11 ssl LibreSSL/2.6.5
zip: extension present, unzip present, 7-Zip present (7z)

```

When I run this command: (`--ignore-platform-reqs` due to local enviroment, issue occurs on other machines without this argument as well)
```
composer install -vvv --ignore-platform-reqs
```

I get the following output: (full output on request)

```
Running 2.1.6 (2021-08-19 17:11:08) with PHP 7.1.33 on Darwin / 18.7.0
…
  - Installing internal/silverstripe-module-required (2.5.x-dev 63aa942b33de): Executing command (modules/silverstripe-module-required): hg clone -- 'ssh://xxx@repo.internal-hg-satis.example/source/dev.php.silverstripe.module-required/' 'modules/silverstripe-module-required'
Executing command (/Users/example/Temp/composer/modules/silverstripe-module-required): hg up -- '63aa942b33de'
Failed: [RuntimeException] Failed to execute hg up -- '63aa942b33de'

abort: no repository found in '/Users/example/Temp/composer/modules/silverstripe-module-required' (.hg not found)

    Install of internal/silverstripe-module-required failed
…
Downloading https://packagist.org/downloads/
[201] https://packagist.org/downloads/

                                                                                                                       
[...]

PR #11212, opened 2022-12-06
Bugfix for hg clone failing when using custom installation path

sb-relaxt-at:
This is a proposed bugfix for #10123


sb-relaxt-at:
Thanks for your feedback. This solution has been implemented prior the ongoing conversation and has been tested for some time. I will give the different approach (not passing a cwd) a try and update the PR as soon as I have tested it.

Seldaek:
I don't think a realpath here is a good solution, as if the dir doesn't exist yet it'll return false, which would be definitely a bad outcome.

Issue #11179, opened 2022-11-08
`composer bump` doesn't bump libraries marked with unbound version constraints


ethanclevenger91:
My `composer.json`:

```json
{
    "name": "roots/bedrock",
    "type": "project",
    "license": "MIT",
    "description": "WordPress boilerplate with modern development tools, easier configuration, and an improved folder structure",
    "homepage": "https://roots.io/bedrock/",
    "authors": [
        {
            "name": "Scott Walkinshaw",
            "email": "scott.walkinshaw@gmail.com",
            "homepage": "https://github.com/swalkinshaw"
        },
        {
            "name": "Ben Word",
            "email": "ben@benword.com",
            "homepage": "https://github.com/retlehs"
        }
    ],
    "keywords": [
        "bedrock",
        "composer",
        "roots",
        "wordpress",
        "wp",
        "wp-config"
    ],
    "support": {
        "issues": "https://github.com/roots/bedrock/issues",
        "forum": "https://discourse.roots.io/category/bedrock"
    },
    "config": {
        "preferred-install": "dist",
        "allow-plugins": {
            "composer/installers": true,
            "roots/wordpress-core-installer": true
        }
    },
    "repositories": {
        "sterner-stuff/made-of-sterner-stuff": {
            "type": "vcs",
            "url": "git@github.com:sterner-stuff/made-of-sterner-stuff.git"
        },
		"sterner-stuff/kinsta-mu-plugin": {
			"type": "vcs",
			"url": "git@github.com:sterner-stuff/kinsta-mu-plugin.git"
		},
        "0": {
            "type": "composer",
            "url": "https://wpackagist.org"
        },
        "package-peak": {
            "type": "composer",
            "url": "https://packagepeak.app/packages.json"
        },
        "deliciousbrains": {
            "type": "composer",
            "url": "https://composer.deliciousbrains.com"
        },
        "woocommerce/action-scheduler": {
            "type": "vcs",
            "url": "git@github.com:woocommerce/action-scheduler.git"
        },
        "fresa": {
            "type": "vcs",
            "url": "git@github.com:ethanclevenger91/fresa.git"
        },
        "sterner-stuff/advanced-custom-fields-pro": {
            "type": "vcs",
            "url": "git@github.com:sterner-stuff/advanced-custom-fields-pro.git"
        },
        "rocketgenius/gravityforms": {
            "type": "vcs",
            "url": "git@github.com:sterner-stuff/gravityforms.git"
        }
    },
    "require": {
[...]

PR #11218, opened 2022-12-12
Fix preserving of major.minor.patch version when running `composer bump` and installed patch version is `0`, and add bumping of >=x to >=latest

localheinz:
This pull request

- [x] adds a failing test case to prevent the `VersionBumper` from dropping the `.0` patch version when `major.minor.patch` version is used in version constraint and the installed patch version is `0`

See #11220.


Seldaek:
Thanks for the test, I hijacked the PR to do the fix and also fixed #11179 while I was in the file (and to avoid conflicts with a second PR..) :)

localheinz:
Thank you very much, @Seldaek!

PR #11219, opened 2022-12-12
Conditional return type for `IOInterface::select()`

staabm:
I am not into all the details, so I am not 100% confident the phpdoc matches the implementation.


eliashaeussler:
That's a nice addition to my original patch. Thanks, @staabm!

Seldaek:
Thanks

PR #11222, opened 2022-12-13
Update PHPStan

ondrejmirtes:
I managed to auto-detect phpunit-bridge in PHPStan itself. I actually adapted and used the code from your repository https://github.com/phpstan/phpstan-src/commit/c760b1f7a45c512cd34c91d8e078046dfdc453ba, I hope you appreciate this :)


Seldaek:
Hm I don't quite understand the error, it looks like the phpunit is not being autoloaded at all anymore..

Seldaek:
I think I get it, your version doesn't include the phpunit autoloader right? It simply adds it to your locator for reflection purposes.

Then my tests/bootstrap.php here does a `require __DIR__.'/Composer/Test/TestCase.php';` so it expects phpunit files to be autoloadable at that point, as it was the case before when including via phpstan, and is the case when including via phpunit.

Seldaek:
This was added a long time ago I am not sure why https://github.com/composer/composer/commit/38917c204758f1dd97569af299b3b87b3d0d2c10#diff-96a82592013e5f4b91e682332583b7a4a6ca1ff0431d267c61179b1a21b167da - possibly/probably can be removed now :)

Seldaek:
Ok works now, cool

Issue #11235, opened 2022-12-21
Define default exclude patterns for the archive command


Sweetchuck:
## My `composer.json`:

```json
{
    "name": "me/p1",
    "archive": {
        "exclude": []
    }
}
```

## When I run this command: 

```bash
composer archive
ls -lh ./me-p1-*
echo 'Again'
composer archive
ls -lh ./me-p1-*
```

```
Creating the archive into ".".
Created: ./me-p1-HASH.tar
4MB me-p1-HASH.tar
Again
Creating the archive into ".".
Created: ./me-p1-HASH.tar
8MB me-p1-HASH.tar
```

## And I expected this to happen:

Exclude the destination file (all the variants, zip, tar.bz2 etc...) from the archive.

The second archive contains the previously created archive, that is why its size is doubled.
It is rarely the desired result to have the previously created archive included into the next one.

I know it can be prevented by defining exclude patterns in `composer.json#/archive/exclude`, but I think those exclude patterns should be added by default (or hard-coded).

PR #11239, opened 2022-12-22
Default exclude patterns for the "archive" command

Sweetchuck:
Issue #11235

Adds default exclude patterns for the "archive" command based on the target file name.

### DIFF
--- a/src/Composer/Package/Archiver/ArchiveManager.php
+++ b/src/Composer/Package/Archiver/ArchiveManager.php
@@ -71,6 +71,52 @@ class ArchiveManager
     }
 
     /**
+     * @return array<string, string>
+     */
+    public function getPackageFilenameParts(CompletePackageInterface $package): array
+    {
+        $baseName = $package->getArchiveName();
+        if (null === $baseName) {
+            $baseName = Preg::replace('#[^a-z0-9-_]#i', '-', $package->getName());
+        }
+
+        $parts = [
+            'base' => $baseName,
+        ];
+
+        $distReference = $package->getDistReference();
+        if (null !== $distReference && Preg::isMatch('{^[a-f0-9]{40}$}', $distReference)) {
+            $parts['dist_reference'] = $distReference;
+            $parts['dist_type'] = $package->getDistType();
+        } else {
+            $parts['version'] = $package->getPrettyVersion();
+            $parts['dist_reference'] = $distReference;
+        }
+
+        $sourceReference = $package->getSourceReference();
+        if (null !== $sourceReference) {
+            $parts['source_reference'] = substr(sha1($sourceReference), 0, 6);
+        }
+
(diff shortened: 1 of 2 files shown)


src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -199,4 +232,54 @@ public function archive(CompletePackageInterface $package, string $format, strin
...
         return $target;
     }
+
+    /**
+     * @param string[] $parts
+     * @param string[] $formats
+     *
+     * @return string[]
+     */
+    public function buildExcludePatterns(array $parts, array $formats): array
```
Seldaek, line 242: Can this be made private?

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -199,4 +232,54 @@ public function archive(CompletePackageInterface $package, string $format, strin
...
+            $patterns[] = "$base.$format";
+        }
+
+        return $patterns;
+    }
+
+  /**
+   * @return string[]
+   */
+    public function getSupportedFormats(): array
```
Seldaek, line 260:
```suggestion
    /**
     * @return string[]
     */
    private function getSupportedFormats(): array
```

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -71,36 +71,61 @@ public function setOverwriteFiles(bool $overwriteFiles): self
...
-        $name = implode('-', array_filter($nameParts, static function ($p): bool {
-            return !empty($p);
-        }));
+        return $parts;
+    }
+
+    /**
+     * @param array<string, string> $parts
+     *
+     * @return string
```
Seldaek, line 112: Let's avoid expanding the public method surface here

```suggestion
     * @return string
     * @internal
```

src/Composer/Package/Archiver/ArchiveManager.php
```diff
@@ -71,36 +71,61 @@ public function setOverwriteFiles(bool $overwriteFiles): self
     }
 
     /**
-     * Generate a distinct filename for a particular version of a package.
-     *
-     * @param CompletePackageInterface $package The package to get a name for
-     *
-     * @return string A filename without an extension
+     * @return array<string, string>
```
Seldaek, line 74: Let's avoid expanding the public method surface here

```suggestion
     * @return array<string, string>
     * @internal
```

src/Composer/Package/Archiver/ArchiveManager.php
Seldaek, line 242:
```suggestion
    private function buildExcludePatterns(array $parts, array $formats): array
```

Seldaek:
Thanks!

PR #11257, opened 2023-01-08
Allow v2 metadata target to provide security advisories without releases

### LINKED ISSUE #11256: Allow composer type repository to return only security advisories for use with audit command
Currently it is not possible for a metadata endpoint to provide security-advisories without also providing a list of package download versions.

In ComposerRepository::getSecurityAdvisories() when the security-advisories['metadata'] flag is set and no security-advisories['api-url'] is provided ComposerRepository::startCachedAsyncDownload() will be called to download the v2 format JSON from the path specified in the repository metadata-url parameter.

ComposerRepository::startCachedAsyncDownload() however will not return a response unless the packages key is present with the package name:

If the "packages" field is present but empty in the response when ComposerRepositry::loadAsyncPacakage() is called by ComposerRepositry::getPacakges() the package will become unreachable in lower priority repositories preventing installation with `composer install` (if the package is already present in the lock file it will however remain available.) 

The current alternative is for a repository to only provide security advisories through the api-url target, this however can not be cached as effectively by CDN's due to being a POST with potentially each caller having a different packages list.

Currently required to provide security advisory data:

```
{
  "packages": {
    "example/package": []
  },
  "security-advisories": [
    {
      "package": "example/package",
      "link": "https://www.example.org",
      "affectedVersions": "<=3.1.0",
      "reportedAt": "2022-12-27 04:19:23",
      "title": "Test package vulnerability record",
      "advisoryId": "test"
    }
  ]
}
```

Requested format:
```
{
  "security-advisories": [
    {
      "package": "example/package",
      "link": "https://www.example.org",
      "affectedVersions": "<=3.1.0",
      "reportedAt": "2022-12-27 04:19:23",
      "title": "Test package vulnerability record",
      "advisoryId": "test"
    }
  ]
}
```

cmlara:
Fixes #11256

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -1017,7 +1017,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
             $promises[] = $this->startCachedAsyncDownload($name, $realName)
                 ->then(function (array $spec) use (&$packages, &$namesFound, $realName, $constraint, $acceptableStabilities, $stabilityFlags, $alreadyLoaded): void {
                     [$response, $packagesSource] = $spec;
-                    if (null === $response) {
+                    if (null === $response || !isset($response['packages'][$realName])) {
                         return;
                     }
 
@@ -1092,7 +1092,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
                     $response = $contents;
                 }
 
-                if (!isset($response['packages'][$packageName])) {
+                if (!isset($response['packages'][$packageName]) && !isset($response['security-advisories'])) {
                     return [null, $packagesSource];
                 }
 


src/Composer/Repository/ComposerRepository.php
```diff
@@ -1017,7 +1017,7 @@ private function loadAsyncPackages(array $packageNames, ?array $acceptableStabil
             $promises[] = $this->startCachedAsyncDownload($name, $realName)
                 ->then(function (array $spec) use (&$packages, &$namesFound, $realName, $constraint, $acceptableStabilities, $stabilityFlags, $alreadyLoaded): void {
                     [$response, $packagesSource] = $spec;
-                    if (null === $response) {
+                    if (null === $response || !isset($response['packages'][$realName])) {
```
Seldaek, line 1020: How about just this? As that would clearly indicate that package data is not provided at all here IMO.

```suggestion
                    if (null === $response || !isset($response['packages'])) {
```

cmlara: Previously if $response['packages'][$realName]  wasn't set $response would be set to null by startCachedAsyncDownload()

However now with the change in startCachedAsyncDownload() if a security advisory is present $response will not be null yet $response['packages'] could  be present while  $response['packages'][$realName] is not present.

The remaining code in that method method relies on the assumption that $response['packages'][$realName] has been validated as being set.

It would be a (badly) broken repository for that to happen, however in my opinion it is better that we validate rather than assume the response contains the key we need.

Seldaek: Ok never mind I took a closer look now, thanks for explaining. Looks good to me then, thanks!

PR #11270, opened 2023-01-13
Warn when require ends up auto-selecting a feature branch

### LINKED ISSUE #11264: BC: Composer `2.5.0` introduces new dependency resolution that may cause issues
As of Composer `^2.5.0` (#11160), it is now possible to resolve dependencies in a new way that was not previously possible. Here's an example assuming the following:
- The user is working within a Laravel 9 application.
- The user wants to require a third-party/package that does not currently support Laravel 9.
- The package developer has started a branch (feat/laravel-9-support) to add support for Laravel 9, but the work is not yet complete.

By running the command:
```bash
> composer require third-party/package
```

The user will get:
```bash
Using version dev-feat/laravel-9-support for third-party/package
```

It is crucial to keep in mind that this functionality may result in confusion for the user in specific scenarios, and there is a high likelihood that it will disrupt existing CI processes that take actions — such as deployments, testing results, automatic releases — based on the resolution of those dependencies.

EDIT 1: It's worth noting that in the example given, as well as in all Laravel applications, the following settings are typically set in the composer.json file:

```json
"minimum-stability": "dev",
"prefer-stable": true
```

It is a topic of debate whether these settings should be present in Laravel applications, however, it is important to note that the example provided earlier (the third-party/package) would not have been possible when using Composer `2.4.4`.

EDIT 2: Just pointing out, that even if the `dev-feat/laravel-9-support` gets merged on `master` - the branch used to tag releases - previously composer would never resolve `dev-master` while now, it installs `dev-master` with the following message: `Using version ^1.0@dev for third-party/package` for example.

Seldaek:
fixes #11264

### DIFF
--- a/src/Composer/Command/RequireCommand.php
+++ b/src/Composer/Command/RequireCommand.php
@@ -338,7 +338,7 @@ EOT
         try {
             $result = $this->doUpdate($input, $output, $io, $requirements, $requireKey, $removeKey);
             if ($result === 0 && count($requirementsToGuess) > 0) {
-                $this->updateRequirementsAfterResolution($requirementsToGuess, $requireKey, $removeKey, $sortPackages, $input->getOption('dry-run'));
+                $result = $this->updateRequirementsAfterResolution($requirementsToGuess, $requireKey, $removeKey, $sortPackages, $input->getOption('dry-run'));
             }
 
             return $result;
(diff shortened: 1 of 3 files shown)


stof (approved):

naderman:
Sorry missed that you requested my review, but indeed looks good like we discussed!

PR #11271, opened 2023-01-14
Add security to support options

ramsey:
This support option allows projects to specify a URL to the project's vulnerability disclosure policy (VDP).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -98,6 +98,7 @@
     },
     "support": {
         "issues": "https://github.com/composer/composer/issues",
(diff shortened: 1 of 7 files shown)


Seldaek:
Shame this doesn't seem to be returned as part of the GitHub API so we can't easily detect its presence without additional API calls.. Which is not really worth it IMO.

But the feature looks good anyway for those who care to list their policy explicitly. Thanks!

Seldaek (approved):

stof (approved):

PR #11277, opened 2023-01-20
Test install command

theoboldalex:
re: #10796 

Adds some initial tests for the Install command. I think there is some more work to do here in improving tests on the happy path but this gets us started at least.

Would appreciate some guidance/feedback on other test cases that could improve this.

Thanks

### DIFF
--- a/tests/Composer/Test/Command/InstallCommandTest.php
+++ b/tests/Composer/Test/Command/InstallCommandTest.php
@@ -13,8 +13,73 @@
 namespace Composer\Test\Command;
 
 use Composer\Test\TestCase;
+use Generator;
 
 class InstallCommandTest extends TestCase
 {
+    /**
+     * @dataProvider useCaseProvider
+     * @param array<mixed> $composerJson
+     * @param array<mixed> $command
+     */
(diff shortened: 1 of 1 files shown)


Seldaek:
Yeah this looks good for error cases, but testing the happy path definitely would be good as well. I think you could do it with packages of type metapackage (which then doesn't require source/dist to be present) and adding new packages in the package repository definition so that they can be installed even if they were not yet present in the installed.json. With/without lock file etc.

theoboldalex:
Thanks, I'll get to fixing this up as soon as possible.

Seldaek:
@theoboldalex no worries if you don't have time but are you still interested in tweaking this further? Otherwise I can also merge as is, it's already something for sure :)

theoboldalex:
@Seldaek I haven't had much time to work on this recently unfortunately. Feel free to merge this in as it is. If I can find some time, I will revisit it in future to improve/add more tests.

Seldaek:
Ok thanks

Seldaek:
Added some basic happy paths tests myself in 0d68dabb5 just so I don't forget :)

PR #11283, opened 2023-01-27
Fix self.version requirements failing the lock file integrity check if on a different version

### LINKED ISSUE #11274: Composer fails to resolve self.version strings for Merge Requests
After updating composer to 2.5.1 the Drupal communities testing pipeline began to fail with the following errors:

```
00:00:51.277 > Drupal\Composer\Composer::ensureComposerVersion
00:00:51.277 Installing dependencies from lock file (including require-dev)
00:00:51.277 Verifying lock file contents can be installed on current platform.
00:00:51.277 - Required package "drupal/core" is in the lock file as "10.1.x-dev" but that does not satisfy your constraint "self.version".
00:00:51.277 - Required package "drupal/core-project-message" is in the lock file as "10.1.x-dev" but that does not satisfy your constraint "self.version".
00:00:51.277 - Required package "drupal/core-vendor-hardening" is in the lock file as "10.1.x-dev" but that does not satisfy your constraint "self.version".
00:00:51.277 This usually happens when composer files are incorrectly merged or the composer.json file is manually edited.

```

Pinning to version 2.4.4 resolved the issue temporarily, we are also tracking the issue here: https://www.drupal.org/project/drupalci_environments/issues/3333480

The issue began when updating php for security using these php versions: https://hub.docker.com/u/drupalci
The php version is likely unrelated, we just did not pin composer to 2.4.4 and composer was updated to 2.5.1 when the container was reubuilt.
If a patch file was created for testing instead of a merge request composer commands succeeded. Only merge requests were affected.

My `composer.json`:

```json
attached as a separate comment
```

```
I don't have access to run this.
```

When I run this command: 

```
sudo -u www-data /usr/local/bin/composer install --prefer-dist --no-suggest --no-interaction --no-progress --working-dir /var/www/html
```

I get the following output: 
Full CI run here:
https://dispatcher.drupalci.org/job/drupal_patches/161235/console

```
You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.
08:31:49 > Drupal\Composer\Composer::ensureComposerVersion
08:31:49 Installing dependencies from lock file (including require-dev)
08:31:49 Verifying lock file contents can be installed on current platform.
08:31:49 - Required package "drupal/core" is in the lock file as "10.1.x-dev" but that does not satisfy your constraint "self.version".
08:31:49 - Required package "drupal/core-project-message" is in the lock file as "10.1.x-dev" but that does not satisfy your constraint "self.version".
08:31:49 - Required package "drupal/core-vendor-hardening" is in the lock file as "10.1.x-dev" but that does not satisfy your constraint "self.version".
08:31:49 This usually happens when composer files are incorrectly merged or the composer.json file is manually edited.
08:31:49 Read more about correctly resolving merge conflicts https://getcomposer.org/doc/articles/resolving-merge-conflicts.md
08:31:49 and prefer using the "require" command over editing the composer.json file directly https://getcomposer.org/doc/03-cli.md#require-r
08:31:49 Composer error. Unable to continue.
08:31:49 --- Commands Executed ---
08:31:49 sudo -u www-data /usr/local/bin/composer config -g discard-changes true
08:31:49 sudo -u www-data /usr/local/bin/composer install --prefer-dist --no-suggest --no-interaction --no-progress --working-dir /var/www/html
08:31:49 Return Code: 4
```

Composer to install the dependencies from the merge request that triggered the run.

Seldaek:
Fixes #11274

I don't think it is worth it to check self.version here as the version is very likely to change and break things.

### DIFF
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -518,6 +518,9 @@ class Locker
                 if (PlatformRepository::isPlatformPackage($link->getTarget())) {
                     continue;
                 }
+                if ($link->getPrettyConstraint() === 'self.version') {
+                    continue;
+                }
                 if ($installedRepo->findPackagesWithReplacersAndProviders($link->getTarget(), $link->getConstraint()) === []) {
                     $results = $installedRepo->findPackagesWithReplacersAndProviders($link->getTarget());
                     if ($results !== []) {
--- a/tests/Composer/Test/Fixtures/installer/root-requirements-do-not-affect-locked-versions.test
+++ b/tests/Composer/Test/Fixtures/installer/root-requirements-do-not-affect-locked-versions.test
@@ -2,26 +2,30 @@
 The locked version will not get overwritten by an install but fails on invalid packages
 --COMPOSER--
 {
+    "version": "1.2.3",
     "repositories": [
         {
             "type": "package",
             "package": [
                 { "name": "foo/bar", "version": "1.0.0" },
                 { "name": "foo/baz", "version": "1.0.0" },
-                { "name": "foo/baz", "version": "2.0.0" }
+                { "name": "foo/baz", "version": "2.0.0" },
+                { "name": "foo/self", "version": "1.2.3" }
             ]
         }
     ],
     "require": {
         "foo/bar": "2.0.0",
-        "foo/baz": "2.0.0"
+        "foo/baz": "2.0.0",
+        "foo/self": "self.version"
     }
 }
 --LOCK--
 {
     "packages": [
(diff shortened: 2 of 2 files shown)


naderman:
@Seldaek is there a good reason at all that we store the root package in the lock file as a locked package?

Seldaek:
We don't do it. This is about the new check on install that lock file packages fulfill root requirements. Which if the root version changed they do not anymore in case of self.version requirements.

naderman:
@Seldaek sorry i was misreading this, i thought "foo/self" was the root package, got it now and makes sense!

PR #11285, opened 2023-01-27
Fix autoloading on PHP 5.6

### LINKED ISSUE #11245: Failed to autoload classes by upgrading composer to `2.5.1`
My `composer.json`:

```json
{
    "require": {
        "php": "^7.0.0",
        "aws/aws-sdk-php": "3.224.*",
        "facebook/graph-sdk": "5.6.*",
        "google/protobuf": "v3.7.1",
        "grpc/grpc": "1.19.0",
        "justinrainbow/json-schema": "^5.2",
        "mongodb/mongodb": "1.4.*",
        "phpunit/phpunit": "6.5.*"
    },
    "config": {
        "platform": {
            "php": "7.0"
        }
    },
    "autoload": {
        "psr-0": {
            ... // proprietary packages
        },
        "psr-4": {
            "GPBMetadata\\": "./Protobuf/GPBMetadata/",
            "Protobuf\\": "./Protobuf/Protobuf/",
            ... // proprietary packages
        }
    },
    "minimum-stability": "stable"
}

```

```
Checking git settings: OK git version 2.30.1
Checking disk free space: OK
Checking pubkeys:
OK
Checking composer version: OK
Composer version: 2.5.1
PHP version: 8.2.0
PHP binary path: /opt/homebrew/Cellar/php/8.2.0/bin/php
OpenSSL version: OpenSSL 1.1.1s  1 Nov 2022
cURL version: 7.87.0 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1s
```

When I launched my program, an error thrown with:

```
PHP Fatal error:  Uncaught TypeError: Return value of Composer\Autoload\ClassLoader::initializeIncludeClosure() must be an instance of Composer\Autoload\void, none returned in /var/app/current/_libs/vendor/composer/ClassLoader.php:580
Stack trace:
#0 /var/app/current/_libs/vendor/composer/ClassLoader.php(112): Composer\Autoload\ClassLoader::initializeIncludeClosure()
#1 /var/app/current/_libs/vendor/composer/autoload_real.php(28): Composer\Autoload\ClassLoader->__construct('/var/app/curren...')
#2 /var/app/current/_libs/vendor/autoload.php(25): ComposerAutoloaderInit8cbab0e9160881f88c44458183518071::getLoader()
#3 /var/app/current/ajax.php(24): require_once('/var/app/curren...')
#4 {main}
  thrown in /var/app/current/_libs/vendor/composer/ClassLoader.php on line 580
```

Though I am using PHP 8 on my dev machine, we're actually running codes with PHP 7.0, and the `ClassLoader.php` in newer version seems to have 7.0 incompatible syntax like:

```php
private static function initializeIncludeClosure(): void // void return types are not allowed in PHP 7.0
{
    if (self::$includeFile !== null) {
        return;
    }

    /**
     * Scope isolated include.
     *
     * Prevents access to $this/self from included files.
     *
     * @param  string $file
     * @return void
     */
    self::$includeFile = static function($file) {
        include $file;
    };
}
```

Reference to RFC: https://wiki.php.net/rfc/void_return_type

Seldaek:
Fixes #11245

### DIFF
--- a/.github/workflows/autoloader.yml
+++ b/.github/workflows/autoloader.yml
@@ -1,0 +1,36 @@
+name: "Autoloader"
+
+on:
+  push:
+    paths-ignore:
+      - 'doc/**'
+  pull_request:
+    paths-ignore:
+      - 'doc/**'
+
+permissions:
+  contents: read
+
+jobs:
+  tests:
+    name: "Autoloader"
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v3"
+
+      - name: "Install locked dependencies"
+        run: "cd tests/Composer/Test/Autoload/MinimumVersionSupport; composer config platform --unset && composer install"
+
+      - name: "Install oldest supported PHP version for autoloader"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          extensions: "intl, zip"
+          ini-values: "memory_limit=-1"
+          php-version: "5.6"
+
+      - name: "Check the autoloader can be executed"
+        run: "cd tests/Composer/Test/Autoload/MinimumVersionSupport; php main.php"
--- a/.github/workflows/autoloader.yml
+++ b/.github/workflows/autoloader.yml
@@ -21,8 +21,11 @@ jobs:
       - name: "Checkout"
         uses: "actions/checkout@v3"
 
-      - name: "Install locked dependencies"
-        run: "cd tests/Composer/Test/Autoload/MinimumVersionSupport; composer config platform --unset && composer install"
+      - name: "Install Composer dependencies"
+        run: "composer config platform --unset && composer install"
+
+      - name: "Dump autoloader in the test directory using latest Composer"
+        run: "./bin/composer install -d tests/Composer/Test/Autoload/MinimumVersionSupport"
 
       - name: "Install oldest supported PHP version for autoloader"
         uses: "shivammathur/setup-php@v2"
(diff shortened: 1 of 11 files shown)


.github/workflows/autoloader.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: "Install oldest supported PHP version for autoloader"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          extensions: "intl, zip"
+          ini-values: "memory_limit=-1"
+          php-version: "5.6"
+
+      - name: "Check the autoloader can be executed"
+        run: "cd tests/Composer/Test/Autoload/MinimumVersionSupport; php main.php"
```
stof, line 39: instead of this `cd`, you could use the `working-directory` option

tests/Composer/Test/Autoload/MinimumVersionSupport/main.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+include 'vendor/autoload.php';
+
+Test\Foo::test();
```
stof, line 5: Is it only the ClassLoader that is expected to support PHP 5.6+ or the whole runtime API ? In the second case, this should also try to use the InstalledVersions API.

Seldaek: Yup good points thanks!

benwil1447:
I need PHP mailer.? pls are you on whatsapp?

PR #11292, opened 2023-02-04
Allow comments for root project composer.json

vtsykun:
## Proposal
This PR proposes to allow the use of comments in the root composer.json of project. At the same time for vendor packages comments in `composer.json` are not allowed (as before). Today we have such tools `docker-compose`, for example, where possible fast disable the some configuration via comments in yaml. 

This feature allows the users to quickly change `composer.json` in local development environment for debug purposes:
```
{
    "repositories": [
        {"type": "composer", "url": "https://example.com/mirror/org1/"}
       // {"packagist": false}
    ],
    "require": {
        "php": ">=8.1",
        "ext-redis": "*",
        "firebase/php-jwt": "^6.0",
      //  "babdev/pagerfanta-bundle": "^3.7",
        "babdev/pagerfanta-bundle": "^2.0",
        "cebe/markdown": "^1.1"
    }
}
```

### New composer commands behaviour if JSON is not valid.

| The Command       | behaviour                       |
|-------------------|----------------------------------------------------------|
| composer install  | ignore comments                                          |
| composer update   | ignore comments                                          |
| composer require  | ignore comments and dump composer.json without comments  |
| composer validate | not allow comments                                       |

composer `depends`, `suggests`, `run-script`, `reinstall` etc. ignore comments too

Composer validate example
validate

### Examples comment format

Allowed `// ...` and `/* ... */` 

### Packagist support?

This is major BC break, so the repository implementors MUST **not allow** to use comments for vendors composer.json files. The metadata should be easily extracted from `composer.json` using the standard json lib function.

### Impact on performance.

No impact. Before reading `composer.json` with a new json parser we try to read json file with standard `json_decode` function 

## Backward Incompatible Changes

No backwards incompatible changes. 

## Todo List

- [x] Create PR https://github.com/Seldaek/jsonlint/pull/81  
- [x] Added supporting comments to composer
- [ ] Test cover (blocked by PR 1)
- [ ] Documentation usage notices

### DIFF
--- a/src/Composer/Command/RequireCommand.php
+++ b/src/Composer/Command/RequireCommand.php
@@ -154,7 +154,7 @@ EOT
             file_put_contents($this->file, "{\n}\n");
         }
 
-        $this->json = new JsonFile($this->file);
+        $this->json = new JsonFile($this->file, null, null, JsonFile::ALLOW_COMMENTS);
         $this->lock = Factory::getLockFile($this->file);
         $this->composerBackup = file_get_contents($this->json->getPath());
         $this->lockBackup = file_exists($this->lock) ? file_get_contents($this->lock) : null;
--- a/src/Composer/Factory.php
+++ b/src/Composer/Factory.php
@@ -298,7 +298,7 @@ class Factory
         if (is_string($localConfig)) {
             $composerFile = $localConfig;
 
-            $file = new JsonFile($localConfig, null, $io);
+            $file = new JsonFile($localConfig, null, $io, JsonFile::ALLOW_COMMENTS);
 
             if (!$file->exists()) {
                 if ($localConfig === './composer.json' || $localConfig === 'composer.json') {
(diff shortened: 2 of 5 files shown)


Seldaek:
As we have discussed in the past, this is IMO not great for the ecosystem and we'd rather not allow it as it would mean every tool, IDE etc out there which parses composer.json suddenly will have to learn to parse comments.

We have added explicit support in the schema for a `_comment` top level key in which you can store notes if you really must:

```
{
    "_comment": [
        "hello world"
    ]
}
```

If you need to get rid of stuff temporarily you can really just delete them.. that's why we have version control.

vtsykun:
I agree about that is violation JSON format and not good for the ecosystem. However here I have only assumed that the comments will be only used locally - but not pushed to VCS.

> If you need to get rid of stuff temporarily you can really just delete them.. that's why we have version control.

Anyway the comments are an easily way to disable some functionality. I'll try to do this feature as composer plugin.

Seldaek:
Not pushed to VCS is something we cannot enforce, and I can guarantee you if it is available it will be abused :)

PR #11296, opened 2023-02-04
Update 03-cli.md

iVegas:
Update incorrect alias for `require` command and provide alias info for `remove` command


Seldaek:
Thanks but there are no req or rem aliases if you look at the sources. What happens is if you use `req` or `requ` or `requi` etc it guesses which command you meant and if it is not ambiguous it just calls it. There is however an explicit `r` alias to make `r` call require, otherwise without alias `r` would be ambiguous (require, remove, reinstall, run-script all match).

iVegas:
Oh, that is interesting, it worth mention that in the documentation page. Thanks for explamation.

Seldaek:
Feel free to create a documentation PR :)

PR #11303, opened 2023-02-08
Verify 'versions' key existence

flancer64:
`self::getInstalled()` can return empty array. We should validate key existence to prevent exception like:
```
Exception: Notice: Undefined index: versions in /.../vendor/composer/composer/src/Composer/InstalledVersions.php on line 54 in /.../vendor/magento/framework/App/ErrorHandler.php:61
```

### DIFF
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -53,7 +53,8 @@ class InstalledVersions
     {
         $packages = array();
         foreach (self::getInstalled() as $installed) {
-            $packages[] = array_keys($installed['versions']);
+            if (isset($installed['versions']))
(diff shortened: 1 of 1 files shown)


flancer64:
Error in PHPStan:
```
Error: Offset 'versions' on array{root: array{name: string, pretty_version: string, version: string, reference: string|null, type: string, install_path: string, aliases: array<string>, dev: bool}, versions: array<string, array{pretty_version?: string, version?: string, reference?: string|null, type?: string, install_path?: string, aliases?: array<string>, dev_requirement: bool, replaced?: array<string>, ...}>} in isset() always exists and is not nullable.
```

OK, I'll check array size.

Issue #9937, opened 2021-06-03
vendor/composer/composer/src/Composer/installed.php: failed to open stream 2.1.0


pookmish:
My `composer.json`:

```
{
    "name": "su-hsdo/suhumsci",
    "description": "Project template for Drupal 9 projects with a relocated document root",
    "type": "project",
    "license": "GPL-2.0-or-later",
    "homepage": "https://github.com/SU-HSDO/suhumsci",
    "repositories": [
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        },
        {
            "url": "https://git.drupalcode.org/issue/domain_301_redirect-2954889",
            "type": "vcs"
        },
        {
            "url": "https://git.drupalcode.org/issue/audio_embed_field-3146159",
            "type": "vcs"
        },
        {
            "type": "package",
            "package": {
                "name": "kenwheeler/slick",
                "version": "1.8.0",
                "type": "drupal-library",
                "dist": {
                    "url": "https://github.com/kenwheeler/slick/archive/1.8.0.zip",
                    "type": "zip"
                },
                "require": {
                    "composer/installers": "^1.2.0"
                }
            }
        },
        {
            "type": "package",
            "package": {
                "name": "dinbror/blazy",
                "version": "1.8.2",
                "type": "drupal-library",
                "dist": {
                    "url": "https://github.com/dinbror/blazy/archive/1.8.2.zip",
                    "type": "zip"
                },
                "require": {
                    "composer/installers": "^1.2.0"
                }
            }
        },
        {
            "type": "package",
            "package": {
                "name": "ckeditor-plugin/a11ychecker",
                "version": "1.1.1",
                "type": "drupal-library",
                "dist": {
                    "url": "https://download.ckeditor.com/a11ychecker/releases/a11ychecker_1.1.1.zip",
                    "type": "zip"
                },
                "require": {
                    "composer/installers": "~1.0"
                }
            }
        },
        {
            "type": "package",
            "package": {
                "name": "ckeditor-plugin/balloonpanel",
                "version": "4.14.0",
                "type": "drupal-library",
                "dist": {
                    "url": "https://download.ckeditor.com/balloonpanel/releases/balloonpanel_4.14.0.zip",
                    "type": "zip"
                },
                "require": {
                    "composer/installers": "~1.0"
                }
            }
        },
        {
            "type": "package",
            "package": {
                "name": "ckeditor-plugin/link",
                "version": "4.14.0",
                "type": "drupal-library",
[...]

PR #11304, opened 2023-02-08
Verify size of installed packages array

flancer64:
`self::getInstalled()` can return -empty array- array with empty array. We should validate key existence to prevent exception like:
```
Exception: Notice: Undefined index: versions in /.../vendor/composer/composer/src/Composer/InstalledVersions.php on line 54 in /.../vendor/magento/framework/App/ErrorHandler.php:61
```

### DIFF
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -52,11 +52,8 @@ class InstalledVersions
     public static function getInstalledPackages()
     {
         $packages = array();
-        $installed = self::getInstalled();
-        if ((0 < \count($installed) && (\count(\array_keys($installed[0])) > 0))) {
-            foreach ($installed as $one) {
-                $packages[] = array_keys($one['versions']);
-            }
+        foreach (self::getInstalled() as $installed) {
+            $packages[] = array_keys($installed['versions']);
         }
 
         if (1 === \count($packages)) {
@@ -344,11 +341,11 @@ class InstalledVersions
             // and not from its source location in the composer/composer package, see https://github.com/composer/composer/issues/9937
             if (substr(__DIR__, -8, 1) !== 'C') {
                 self::$installed = require __DIR__ . '/installed.php';
-            } else {
-                self::$installed = array();
             }
+            // don't set as empty array on 'else' to prevent 'Undefined index: versions'
+            // exception in `self::getInstalledPackages()` (see https://github.com/composer/composer/pull/11304)
         }
-        $installed[] = self::$installed;
+        if(null !== self::$installed) $installed[] = self::$installed;
 
         return $installed;
     }
--- a/src/Composer/InstalledVersions.php
+++ b/src/Composer/InstalledVersions.php
@@ -342,10 +342,12 @@ class InstalledVersions
             if (substr(__DIR__, -8, 1) !== 'C') {
                 self::$installed = require __DIR__ . '/installed.php';
             }
-            // don't set as empty array on 'else' to prevent 'Undefined index: versions'
-            // exception in `self::getInstalledPackages()` (see https://github.com/composer/composer/pull/11304)
+            // don't set as empty array on 'else' to prevent 'Undefined index: versions' exception in `self::getInstalledPackages()`
+            // (see https://github.com/composer/composer/pull/11304)
         }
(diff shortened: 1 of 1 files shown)


src/Composer/InstalledVersions.php
```diff
@@ -341,11 +341,11 @@ private static function getInstalled()
...
             if (substr(__DIR__, -8, 1) !== 'C') {
                 self::$installed = require __DIR__ . '/installed.php';
-            } else {
-                self::$installed = array();
             }
+            // don't set as empty array on 'else' to prevent 'Undefined index: versions'
+            // exception in `self::getInstalledPackages()` (see https://github.com/composer/composer/pull/11304)
         }
-        $installed[] = self::$installed;
+        if(null !== self::$installed) $installed[] = self::$installed;
```
stof, line 348: please add the curly braces to respect the PSR-2 coding standards

stof:
Those `@psalm-return` annotations should probably be turned into `@phpstan-return` instead, so that phpstan actually reports failures when the code does not respect them. I'm quite sure it would have reported that bug in that case.
For projects consuming the class, both Psalm and phpstan will read the more detailled types based on annotations using their own prefix or the other prefix. But they won't fail the code analysis based on other tool annotations of your own class.

Seldaek:
Well I am still not sure what this fixes. It seems what @stof said above is still in the PR, I don't see a problem returning an empty array here. Please explain the intent or let's close this.

Seldaek:
No? Empty array will never enter the foreach?

flancer64:
Yes, you are right. I did not explain the problem correctly. It was too long time ago :)

In [this place](https://github.com/composer/composer/blob/f41abfca34fb597ec83503ca84b34861c56ecaf6/src/Composer/InstalledVersions.php#L345) `self::$installed` can be an empty array, and the result will be an array with empty array 
```
private static function getInstalled()
    ...
        if (null === self::$installed) {
            ...
            if (...) {
                ...
            } else {
                self::$installed = array();
            }
        }
        $installed[] = self::$installed;
    ...
}
```

If result of `getInstalledPackages()` will be an array with empty array, then an exception will be thrown in `getInstalledPackages()`.

Seldaek:
Sorry I just still don't get it, here is the code from getInstalledPackages:

```
        $packages = array();
        foreach (self::getInstalled() as $installed) {
            $packages[] = array_keys($installed['versions']);
        }

        if (1 === \count($packages)) {
            return $packages[0];
        }

        return array_keys(array_flip(\call_user_func_array('array_merge', $packages)));
```

Now if we replace getInstalled by an empty array we get:

```
        $packages = array();
        foreach (array() as $installed) {
            $packages[] = array_keys($installed['versions']);
        }

        if (1 === \count($packages)) {
            return $packages[0];
        }

        return array_keys(array_flip(\call_user_func_array('array_merge', $packages)));
```

So foreach of an empty array does nothing, so we can remove the foreach:

```
        $packages = array();
        if (1 === \count($packages)) {
            return $packages[0];
        }

        return array_keys(array_flip(\call_user_func_array('array_merge', $packages)));
```

Then we get an empty $packages, so the if here does not match for sure so we remove it:

```
        $packages = array();
        return array_keys(array_flip(\call_user_func_array('array_merge', $packages)));
```

And now we're left with this which evaluates fine: https://3v4l.org/ss0f9

So I don't understand where you get an exception.

vtsykun:
Probably it's related with #9937 when uses `composer/composer` 1

```php
if (null === self::$installed) {
    // only require the installed.php file if this file is loaded from its dumped location,
    // and not from its source location in the composer/composer package, see https://github.com/composer/composer/issues/9937
    if (substr(__DIR__, -8, 1) !== 'C') {
        self::$installed = require __DIR__ . '/installed.php';
    } else {
        self::$installed = array();
    }
}
$installed[] = self::$installed;
```

If the 1st condition is true and the 2nd  condition = false 

```php
self::$installed = array();
$installed[] = self::$installed;
```

Return result is

```php
$installed = [[]]
```

flancer64:
Thanks, @vtsykun , exactly:

```
foreach ([[]] as $installed) {
    $packages[] = array_keys($installed['versions']);
}
```

Issue #11289, opened 2023-02-02
Package names should be clickable in composer why-not


stof:
When running `composer outdated`, package names are clickable in terminals supporting links.

It would be great if they could also be clickable in the output of `composer why-not` (and probably also `composer why` but I haven't checked the current state there), to easily go to the project page of a package that blocks an update for instance.

agostinofiscale:
I thought this was a useful feature so here we go: https://github.com/composer/composer/pull/11308 :)

PR #11308, opened 2023-02-09
Add source link to composer packages

agostinofiscale:
Hi!

This pull request adds a source/homepage link to packages when the aliases who depends on BaseDependencyCommand gets called. Currently "why" and "why-not".

Fixes https://github.com/composer/composer/issues/11289

I hope this can help,
Agostino.


Seldaek:
Thanks

PR #11310, opened 2023-02-09
Fix strpos deprecation notice

homersimpsons:
I do not exactly know the minimum reproduction, but I had this notice when executing composer. This fix the notice without altering the behaviour.

### DIFF
--- a/src/Composer/DependencyResolver/SolverProblemsException.php
+++ b/src/Composer/DependencyResolver/SolverProblemsException.php
@@ -77,8 +77,11 @@ class SolverProblemsException extends \RuntimeException
     {
         foreach ($reasonSets as $reasonSet) {
             foreach ($reasonSet as $reason) {
-                if (isset($reason["rule"]) && 0 === strpos($reason["rule"]->getRequiredPackage(), 'ext-')) {
-                    return true;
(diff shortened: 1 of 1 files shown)


stof (approved):
this matches the way this is handle in 2.x

However, any reason you are still using Composer 1.x ?

homersimpsons:
> However, any reason you are still using Composer 1.x ?

I do have some plugin incompatibilities that forbid me to use composer 2, furthermore upgrading those projects is not really relevant.

stof:
Given that support for composer 1.x is deprecated in packagist.org, it might still be relevant to upgrade (after all, you decided to upgrade PHP for those projects, otherwise you would not care about the deprecation reported in PHP 8.1+).

Note that I don't know what the policy of @Seldaek and @naderman is regarding patches submitted against the 1.x branch.

Seldaek:
I'll merge this as you did it so why not, but yeah we're not really maintaining 1.10 beyond critical security fixes, and new-PHP support is definitely not on the table, you should rely on v2 for that, or if you're stuck on an old project with old deps then most likely you're better off also using an old PHP.

homersimpsons:
Thank you for the merge :)

I understand there won't be new-PHP support for composer.

PR #11312, opened 2023-02-09
PHPStan: Fixed handling of union types in ConfigReturnTypeExtension

staabm:
see https://phpstan.org/blog/why-is-instanceof-type-wrong-and-getting-deprecated

after this fix the PHPStan extension supports cases like

```php
if (..) {
  $property = 'license';
} else {
  $property = 'description';
}

$value = Config::get($property);
```

which it did not before

### DIFF
--- a/src/Composer/PHPStan/ConfigReturnTypeExtension.php
+++ b/src/Composer/PHPStan/ConfigReturnTypeExtension.php
@@ -30,6 +30,7 @@ use PHPStan\Type\MixedType;
 use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\TypeUtils;
 use PHPStan\Type\UnionType;
 
 final class ConfigReturnTypeExtension implements DynamicMethodReturnTypeExtension
@@ -69,9 +70,18 @@ final class ConfigReturnTypeExtension implements DynamicMethodReturnTypeExtensio
         }
 
         $keyType = $scope->getType($args[0]->value);
-        if ($keyType instanceof ConstantStringType) {
-            if (isset($this->properties[$keyType->getValue()])) {
-                return $this->properties[$keyType->getValue()];
+        // for compat with old phpstan versions, we use a deprecated phpstan method.
+        $strings = TypeUtils::getConstantStrings($keyType);
+        if ($strings !== []) {
+            $types = [];
+            foreach($strings as $string) {
+                if (isset($this->properties[$string->getValue()])) {
+                    $types[] = $this->properties[$string->getValue()];
+                }
+            }
+            
+            if ($types !== []) {
+                return TypeCombinator::union(...$types);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Composer/PHPStan/ConfigReturnTypeExtension.php
```diff
@@ -69,10 +70,16 @@ public function getTypeFromMethodCall(MethodReflection $methodReflection, Method
         }
 
         $keyType = $scope->getType($args[0]->value);
-        if ($keyType instanceof ConstantStringType) {
-            if (isset($this->properties[$keyType->getValue()])) {
-                return $this->properties[$keyType->getValue()];
+        // for compat with old phpstan versions, we use a deprecated phpstan method.
+        $strings = TypeUtils::getConstantStrings($keyType);
```
staabm, line 74: refactored to use the deprecated `TypeUtils::getConstantStrings` method, since composer itself does not specify a min-phpstan version requirement.

src/Composer/PHPStan/ConfigReturnTypeExtension.php
```diff
@@ -69,9 +70,18 @@ public function getTypeFromMethodCall(MethodReflection $methodReflection, Method
...
+        if ($strings !== []) {
+            $types = [];
+            foreach($strings as $string) {
+                if (isset($this->properties[$string->getValue()])) {
```
stof, line 78: Any string not being known here should force skipping the more precise type, falling back to the basic method signature

stof (changes requested):

src/Composer/PHPStan/ConfigReturnTypeExtension.php
```diff
@@ -69,10 +70,16 @@ public function getTypeFromMethodCall(MethodReflection $methodReflection, Method
         }
 
         $keyType = $scope->getType($args[0]->value);
-        if ($keyType instanceof ConstantStringType) {
-            if (isset($this->properties[$keyType->getValue()])) {
-                return $this->properties[$keyType->getValue()];
+        // for compat with old phpstan versions, we use a deprecated phpstan method.
```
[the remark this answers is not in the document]
staabm: good point, fixed

src/Composer/PHPStan/ConfigReturnTypeExtension.php
```diff
@@ -64,18 +65,32 @@ public function getTypeFromMethodCall(MethodReflection $methodReflection, Method
     {
         $args = $methodCall->getArgs();
 
+        $defaultReturn = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
```
staabm, line 68: in case we would have a min phpstan version of 1.5.0 or higher, the method could just return null and the [phpstan-core would take care of determining the default-return type.](https://github.com/phpstan/phpstan-src/commit/a897d3a4a0bc1cc8a385eab34da399a75969a538)

Seldaek: Feel free to assume latest PHPStan and bump the require if you like :)

staabm: this won't work. it depends on the phpstan version used by the project which depends on composer.

the extension is meant for re-use in plugin projects. this means I would either need to declare a non-dev dependency on phpstan or add a conflict. both things I don't like.

Seldaek: Oh right I forgot that aspect, let's leave as is then 👍🏻

PR #11315, opened 2023-02-10
Add extra.plugin-optional to auto-disable plugins in non-interactive mode

nicolas-grekas:
https://github.com/composer/composer/pull/10920 added a protection against inadvertently *not* running a plugin. But not all plugins are mandatory to run, and this check creates needless friction to plugin adoption.

I propose to fix this by allowing composer to auto-disable a plugin in non-interactive mode with a new extra.plugin-optional flag.

### DIFF
--- a/doc/articles/plugins.md
+++ b/doc/articles/plugins.md
@@ -332,6 +332,20 @@ in your composer.json to hint to Composer that the plugin should be activated as
 as possible to prevent any bad side-effects from Composer assuming packages are installed
 in another location than they actually are.
 
+### plugin-optional
+
+Because Composer plugins can be used to perform actions which are necessary for installing
+a working application, like modifying which path files get stored in, skipping required
+plugins unintentionally can result in broken applications. So, in non-interactive mode,
+Composer will fail if a new plugin is not listed in ["allow-plugins"](../06-config.md#allow-plugins)
+to force users to decide if they want to execute the plugin, to avoid silent failures.
+
+As of Composer 2.5.3, you can use the setting `{"extra": {"plugin-optional": true}}` on
+your plugin, to tell Composer that skipping the plugin has no catastrophic consequences,
+and it can safely be disabled in non-interactive mode if it is not yet listed in
+"allow-plugins". The next interactive run of Composer will still prompt users to choose if
+they want to enable or disable the plugin.
+
 ## Plugin Autoloading
 
 Due to plugins being loaded by Composer at runtime, and to ensure that plugins which
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -187,7 +187,7 @@ class PluginManager
             }
         }
 
-        if (!$this->isPluginAllowed($package->getName(), $isGlobalPlugin)) {
+        if (!$this->isPluginAllowed($package->getName(), $isGlobalPlugin, $package->getExtra()['plugin-optional'] ?? false)) {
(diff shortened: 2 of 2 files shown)


doc/articles/plugins.md
```diff
@@ -332,6 +332,15 @@ in your composer.json to hint to Composer that the plugin should be activated as
...
 in another location than they actually are.
 
+### plugin-optional
+
+Because composer plugins can be used to install modules in the right place e.g. for Drupal,
+Wordpress, etc. not running them can result in broken applications. That's why, in non-interactive
+mode, Composer will fail if a plugin is not listed in ["allow-plugins"](../06-config.md#allow-plugins).
+
+As of Composer 2.5.3, you can opt-out from this error by setting `{"extra": {"plugin-optional": true}}`.
+In non-interactive mode, this will allow Composer to skip your plugin if it's not listed in "allow-plugins".
```
naderman, line 342:
```suggestion
Composer plugins can be used to perform actions which are necessary for installing a working application, like modifying which path files get stored in. So skipping required plugins unintentionally can result in broken applications. So, in non-interactive
mode, Composer will fail if a new plugin is not listed in ["allow-plugins"](../06-config.md#allow-plugins) to force users to decide if they want to execute the plugin, but to avoid silent failure.

As of Composer 2.5.3, you can use the setting `{"extra": {"plugin-optional": true}}` on your plugin, to tell Composer that skipping the plugin has no catastrophic consequences, and it can safely be disabled in non-interactive mode if it is not yet listed in "allow-plugins". The next interactive run of Composer will still prompt users to choose if they want to enable or disable the plugin.
```

naderman (changes requested):

doc/articles/plugins.md
nicolas-grekas: updated, thanks!

clxmstaab:
thanks for working on a better user experience for the recent issue <3.

stof:
This looks good to me.

And I suggest this gets backported to the 2.2 LTS as well.

src/Composer/Plugin/PluginManager.php
```diff
@@ -733,6 +733,8 @@ public function isPluginAllowed(string $package, bool $isGlobalPlugin): bool
                         break;
                 }
             }
+        } elseif ($optional) {
+            return false;
```
Seldaek, line 737: Shouldn't there be a warning in this case that an optional plugin was skipped? I'm thinking that would be good especially as this happens when non interactive you kinda want a trail of what happened

nicolas-grekas: There's already one in debug mode, see on the next line of the changes of the attached patch

nicolas-grekas:
> And I suggest this gets backported to the 2.2 LTS as well.

Would work for me. Let me know @naderman @Seldaek if I should rebase.

Seldaek:
Yes for sure I'd backport this but no need to do anything @nicolas-grekas i take care of 2.2

AndreasA:
I know it is a bit late but wouldn't it make more sense to assume a plugin is optional by default and add a "require-plugin" flag instead?

I mean most alrady existing plugins are mostly optional. But it is fine this way as well, as long as it is communicated so it is used correctly by plugin developers.

naderman:
@AndreasA There are a lot of them (basically all custom installers) which are not optional, would argue that's the majority of plugins. And they would break if we made this change now. So no, this definitely is the right way of doing it.

AndreasA:
@naderman OK. Not 100% sure I agree as this mostly impacts non-interactive mode only and using `--no-plugins` would be even worse, but I am fine with your reason.

PR #11316, opened 2023-02-10
Add extra.plugin-optional to auto-disable plugins in non-interactive mode

nicolas-grekas:
Backporting #11315 to 2.2LTS


Seldaek:
Thanks

Issue #11317, opened 2023-02-10
Confusing behaviour of the new plugin-optional flag


mnocon:
My `composer.json`:

```json
{
    "name": "mareknocon/composer-optional-plugin",
    "type": "project",
    "license": "MIT",
    "autoload": {
        "psr-4": {
            "Mareknocon\\ComposerOptionalPlugin\\": "src/"
        }
    },
    "authors": [
        {
            "name": "Marek Nocoń",
            "email": "mnocon@users.noreply.github.com"
        }
    ],
    "require": {
        "php-http/discovery": "1.15.1"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
require.php-http/discovery : exact version constraints (1.15.1) should be avoided if the package follows semantic versioning
Checking platform settings: OK
Checking git settings: OK git version 2.39.1
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.5.3
PHP version: 8.1.15
PHP binary path: /usr/local/Cellar/php@8.1/8.1.15/bin/php
OpenSSL version: OpenSSL 1.1.1s  1 Nov 2022
cURL version: 7.87.0 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1s
zip: extension present, unzip present, 7-Zip not available
```

When I run this command: 

```
composer update --no-interaction
```

I get the following output: 

```
Running 2.5.3 (2023-02-10 13:23:52) with PHP 8.1.15 on Darwin / 22.2.0
Reading ./composer.json (/Users/mareknocon/Desktop/repos/product-catalog/composer-optional-plugin/composer.json)
Loading config file /Users/mareknocon/.composer/config.json
Loading config file /Users/mareknocon/.composer/auth.json
Loading config file ./composer.json (/Users/mareknocon/Desktop/repos/product-catalog/composer-optional-plugin/composer.json)
Checked CA file /usr/local/etc/ca-certificates/cert.pem: valid
[...]

PR #11318, opened 2023-02-10
Added optional plugin check in PluginInstaller

mnocon:
This is the code I used to produce the expected* output for https://github.com/composer/composer/issues/11317 - skipping the optional plugin installation.

*expected from my point of view - not sure if it's correct


Seldaek:
Yeah that definitely looks like an omission :)

PR #11319, opened 2023-02-11
add return type and doc

guanhui07:
php7.2 ，add return type and params doc in main branch


src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -610,9 +612,14 @@ public function createLoader(array $autoloads, ?string $vendorDir = null)
 
     /**
      * @param array<int, array{0: PackageInterface, 1: string}> $packageMap
+     * @param Filesystem $filesystem
+     * @param string $basePath
+     * @param string $vendorPath
```
staabm, line 617: Duplicating types which are already defined natively doesnt add value

Seldaek:
Thanks but this is not desirable. Adding return type is a BC break and we on purpose did not do it on some classes which were known to be extended in the wild.

PR #11331, opened 2023-02-16
ComposerRepository: handle package name with null value as versions

glaubinix:
Prevents:

```
Warning: foreach() argument must be of type array|object, null given at src/Composer/Repository/ComposerRepository.php:1365
```

For repository responses like:
```
{
  "packages": {
    "acme/package": null
  }
}
```


stof:
Please fix your repository instead. That's an invalid response.

glaubinix:
I am aware of that but I don't have control over that Composer repository

src/Composer/Repository/ComposerRepository.php
```diff
@@ -1361,12 +1361,14 @@ private function loadIncludes(array $data): array
 
         if (isset($data['packages'])) {
             foreach ($data['packages'] as $package => $versions) {
-                $packageName = strtolower((string) $package);
-                foreach ($versions as $version => $metadata) {
-                    $packages[] = $metadata;
-                    if (!$this->displayedWarningAboutNonMatchingPackageIndex && $packageName !== strtolower((string) ($metadata['name'] ?? ''))) {
-                        $this->displayedWarningAboutNonMatchingPackageIndex = true;
-                        $this->io->writeError(sprintf("<warning>Warning: the packages key '%s' doesn't match the name defined in the package metadata '%s' in repository %s</warning>", $package, $metadata['name'] ?? '', $this->baseUrl));
+                if (is_array($versions)) {
```
stof, line 1364: Instead of indenting the whole content, I would do a check using `continue` instead. And I would show a warning about the broken repository instead of silently ignoring it:

```php
foreach ($data['packages'] as $package => $versions) {
    if (!\is_array($versions)) {
        $this->io->writeError(sprintf("<warning>Warning: the packages key '%s' doesn't contain a valid value in repository %s</warning>", $package', $this->baseUrl));
        continue;
    }

    // existing code looping over versions goes here
}
```

stof:
> but I don't have control over that Composer repository

I suggest that you report it to the maintainers of that repository then.

glaubinix:
Fixed by the repository maintainer

PR #11350, opened 2023-02-23
Pick higher revision number to guarantee successful file retrieval

AnrDaemon:
`SvnDriver::getTags` and `SvnDriver::getBranches` both fail to compose valid reference to `composer.json` if repository root was moved after their creation. Pick highest available revision from the list to raise chances for success.

### DIFF
--- a/src/Composer/Repository/Vcs/SvnDriver.php
+++ b/src/Composer/Repository/Vcs/SvnDriver.php
@@ -236,14 +236,19 @@ class SvnDriver extends VcsDriver
             if ($this->tagsPath !== false) {
                 $output = $this->execute('svn ls --verbose', $this->baseUrl . '/' . $this->tagsPath);
                 if ($output) {
+                    $lastRev = null;
                     foreach ($this->process->splitLines($output) as $line) {
                         $line = trim($line);
                         if ($line && Preg::isMatch('{^\s*(\S+).*?(\S+)\s*$}', $line, $match)) {
-                            if (isset($match[1], $match[2]) && $match[2] !== './') {
-                                $tags[rtrim($match[2], '/')] = $this->buildIdentifier(
-                                    '/' . $this->tagsPath . '/' . $match[2],
-                                    $match[1]
-                                );
+                            if (isset($match[1], $match[2])) {
+                                if ($match[2] === './') {
+                                    $lastRev = $match[1];
+                                } else {
+                                    $tags[rtrim($match[2], '/')] = $this->buildIdentifier(
+                                        '/' . $this->tagsPath . '/' . $match[2],
+                                        max($lastRev, $match[1])
+                                    );
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


AnrDaemon:
```
$ svn ls -v https://svn.rootdir.org/misc/php-tools/library/branches/
   1034 anrdaemon              сен 14 2019 ./
   1005 anrdaemon              мар 11 2019 1.12/

$ svn log https://svn.rootdir.org/misc/php-tools/library/branches/1.12/ | head -n 10
------------------------------------------------------------------------
r1034 | anrdaemon | 2019-09-14 13:13:49 +0300 (Сб, 14 сен 2019) | 2 lines

= Moved remaining sources to the separate path.

------------------------------------------------------------------------
r1005 | anrdaemon | 2019-03-11 03:31:07 +0300 (Пн, 11 мар 2019) | 2 lines

. Import changelog.

$ svn cat https://svn.rootdir.org/misc/php-tools/library/branches/1.12/composer.json@1005
svn: warning: W160013: '/misc/!svn/rvr/1005/php-tools/library/branches/1.12/composer.json' path not found
svn: E200009: Could not cat all targets because some targets don't exist
svn: E200009: Illegal target for the requested operation

$ svn cat https://svn.rootdir.org/misc/php-tools/library/branches/1.12/composer.json@1034
{
…success
```

AnrDaemon:
I think it deserves a backport to LTS. Say if you need it rebased.

AnrDaemon:
As an afterthought, the revision is useless in the task of retrieving a single file. Thus the patch is really only 3 lines.

```diff
diff --git a/src/Composer/Repository/Vcs/SvnDriver.php b/src/Composer/Repository/Vcs/SvnDriver.php
index f2bfad0f7..371faaf13 100644
--- a/src/Composer/Repository/Vcs/SvnDriver.php
+++ b/src/Composer/Repository/Vcs/SvnDriver.php
@@ -180,15 +180,13 @@ public function getFileContent(string $file, string $identifier): ?string
         Preg::match('{^(.+?)(@\d+)?/$}', $identifier, $match);
         if (!empty($match[2])) {
             $path = $match[1];
-            $rev = $match[2];
         } else {
             $path = $identifier;
-            $rev = '';
         }

         try {
             $resource = $path.$file;
-            $output = $this->execute('svn cat', $this->baseUrl . $resource . $rev);
+            $output = $this->execute('svn cat', $this->baseUrl . $resource);
             if (!trim($output)) {
                 return null;
             }
```

stof:
The revision being useless to retrieve a single file looks like a weird argument to me. the whole point of a VCS system is to track changes of files over revisions. Saying that getting a file at any revision is equivalent implies that you never modify any file after adding them in your VCS.

AnrDaemon:
Except when you meet a VCS that follows different versioning model, than what you are used to in Git.
In Subversion, it is entirely possible to move a tree and be unable to refer to the file by @revision it is apparently attributed to, since it is a revision, in which it was changed, but not necessarily at the place where it is located now.

stof:
@AnrDaemon but ignoring revisions entirely does not seem the right way to fix this issue to me, as it ignores the fact that you wanted an older version of the file.

AnrDaemon:
Where do I want an older version, since at every point in code it refers to current one?

AnrDaemon:
To make it perfectly clear: in Subversion, every directory is a repository in itself. Thus, 'trunk', 'tags', 'branches' - merely a point of view. For VCS, they are all separate directories, with their own revision history.

Seldaek:
I'd rather not merge this as bugfix as I have no way to really try this out.. but let's do it in 2.6 as the change seems reasonable to me. Thanks

AnrDaemon:
I will provide a test case once I get a few free hours. And I'm still leaning to the idea that specifying `@rev` is not relevant to what the code does. I'll check with my copy of svnbook.

Seldaek:
Yes I believe you might be right there, but that's a bigger change that I am not so keen on making in this codepath that few people use as it might lead to regressions we won't notice for a while.

Issue #11288, opened 2023-02-01
Caret operator is ignored by PowerShell under Windows


ravage84:
Running Composer 2.5.1 (2022-12-22 15:33:54) with PHP 7.3.33 on Windows 10, Swiss-German

When using the caret operator in PowerShell under Windows, it gets ignored, regardless of using double quotes. Only using four (!) subsequent caret operators work correctly.

Syntax variant 1:

- ``composer require vendor/package ^2.0.1`` -> ``"2.0.1"`` in composer.json, _incorrect_
- ``composer require vendor/package ^^2.0.1`` -> ``"2.0.1"`` in composer.json, _incorrect_
- ``composer require vendor/package ^^^2.0.1`` -> ``"2.0.1"`` in composer.json, _incorrect_
- ``composer require vendor/package ^^^^2.0.1`` -> ``"^2.0.1"`` in composer.json, **correct**

Syntax variant 2:

- ``composer require vendor/package "^2.0.1"`` -> ``"2.0.1"`` in composer.json, _incorrect_
- ``composer require vendor/package "^^2.0.1"`` -> ``"2.0.1"`` in composer.json, _incorrect_
- ``composer require vendor/package "^^^2.0.1"`` -> ``"2.0.1"`` in composer.json, _incorrect_
- ``composer require vendor/package "^^^^2.0.1"`` -> ``"^2.0.1"`` in composer.json, **correct**

Syntax variant 3:

- ``composer require "vendor/package:^2.0.1"`` ->  "2.0.1"`` in composer.json, _incorrect_
- ``composer require "vendor/package:^^2.0.1"`` ->  "2.0.1"`` in composer.json, _incorrect_
- ``composer require "vendor/package:^^^2.0.1"`` ->  "2.0.1"`` in composer.json, _incorrect_
- ``composer require "vendor/package:^^^^2.0.1"`` ->  "2.0.1"`` in composer.json, **correct**

Even running with very verbose output (``-vvv``), one cannot find any hint about the caret operator being ignored by Composer or that a different version than requested will get installed.

[...]

PR #11351, opened 2023-02-24
Add note about issue with caret operator in PowerShell

ravage84:
Refers to #11288

### DIFF
--- a/doc/articles/versions.md
+++ b/doc/articles/versions.md
@@ -194,6 +194,10 @@ library code.
 
 Example: `^1.2.3`
 
+> **Note:** If you are using the PowerShell on Windows, make sure you enter four
+> subsequent caret operators, e.g. `^^^^1.2.3`, to ensure the caret operator gets
+> passed to Composer correctly.
+
 ## Stability Constraints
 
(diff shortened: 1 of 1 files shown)


doc/articles/versions.md
```diff
@@ -194,6 +194,10 @@ library code.
 
 Example: `^1.2.3`
 
+> **Note:** If you are using the PowerShell on Windows, make sure you enter four
+> subsequent caret operators, e.g. `^^^^1.2.3`, to ensure the caret operator gets
+> passed to Composer correctly.
+
```
Seldaek, line 200:
```suggestion
> **Note:** If you are using PowerShell on Windows, you have to escape
> carets when using them as argument on the CLI for example when using the
> `composer require` command. You have to use four
> subsequent caret operators, e.g. `^^^^1.2.3`, to ensure the caret operator gets
> passed to Composer correctly.

```

Seldaek (commented):
Thanks. Added some clarification tho as it's not clear in the docs above that you're talking about a cli input, these definitely shouldn't be escaped when editing the json file directly for example :)

PR #11363, opened 2023-03-03
Fix markdown: Move headline out of quote

cweiske:
This was caused by
> Add note about issue with caret operator in PowerShell (#11351)

Related: #11288, #11351


cweiske:
See the bug on https://getcomposer.org/doc/articles/versions.md#caret-version-range-

IonBazan (approved):

Seldaek:
Oops thanks

PR #11367, opened 2023-03-07
Add rm alias to composer remove command

dmitryuk:
Inspired from #11163
As you know most `deleting|removing` commands have shortcut `rm`, for example, `rm a.txt`, `npm rm --help`.
So I think it will be great have the same keyboard shortcut here. 
p.s. `composer rem` is not easy to remember

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -298,7 +298,7 @@ If you do not specify a package, Composer will prompt you to search for a packag
 * **--apcu-autoloader-prefix:** Use a custom prefix for the APCu autoloader cache.
   Implicitly enables `--apcu-autoloader`.
 
-## remove
+## remove / rm
 
 The `remove` command removes packages from the `composer.json` file from
 the current directory.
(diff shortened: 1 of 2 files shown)


cafferata:
Would you like to change the anchor documentation link as well? 🙏

```diff
-Read more at https://getcomposer.org/doc/03-cli.md#remove
+Read more at https://getcomposer.org/doc/03-cli.md#remove-rm
```
https://github.com/composer/composer/blob/2.5.4/src/Composer/Command/RemoveCommand.php#L76

dmitryuk:
> Would you like to change the anchor documentation link as well? pray
> 
> ```diff
> -Read more at https://getcomposer.org/doc/03-cli.md#remove
> +Read more at https://getcomposer.org/doc/03-cli.md#remove-rm
> ```
> 
> https://github.com/composer/composer/blob/2.5.4/src/Composer/Command/RemoveCommand.php#L76

Thanks, done

Seldaek (approved):
Looks good, thanks

PR #11376, opened 2023-03-14
Added time of composer.lock generation

Toflar:
Sometimes I would like to know when a `composer.lock` was generated. This would ease debugging as `mtime` is often not accurate (containers and stuff). This could help finding out why a certain constellation of packages resulted as maybe one remembers that a certain package was only published after `<time>` or whatever. 

Just a little helper - anyway, let me know what you think :)

### DIFF
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -365,11 +365,15 @@ class Locker
             return $alias;
         }, $aliases);
 
+        $time = !isset($this->getLockData()['_time']) ?
+            $this->now->format(\DateTimeInterface::ATOM) :
+            ($this->isFresh() ? $this->getLockData()['_time'] : $this->now->format(\DateTimeInterface::ATOM));
+
         $lock = [
             '_readme' => ['This file locks the dependencies of your project to a known state',
                                'Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies',
                                'This file is @gener'.'ated automatically', ],
-            '_time' => $this->now->format(\DateTimeInterface::ATOM),
+            '_time' => $time,
             'content-hash' => $this->contentHash,
             'packages' => null,
             'packages-dev' => null,
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -365,11 +365,17 @@ class Locker
             return $alias;
         }, $aliases);
 
+        try {
+            $time = $this->getLockData()['_time'] ?? $this->now->format(\DateTimeInterface::ATOM);
+        } catch (\Exception $e) {
+            $time = $this->now->format(\DateTimeInterface::ATOM);
+        }
+
         $lock = [
             '_readme' => ['This file locks the dependencies of your project to a known state',
                                'Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies',
                                'This file is @gener'.'ated automatically', ],
-            '_time' => $this->getLockData()['_time'] ?? $this->now->format(\DateTimeInterface::ATOM),
+            '_time' => $time,
             'content-hash' => $this->contentHash,
             'packages' => null,
             'packages-dev' => null,
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -53,6 +53,8 @@ class Locker
     private $lockDataCache = null;
     /** @var bool */
     private $virtualFileWritten = false;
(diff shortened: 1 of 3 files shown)


stof:
To me, this is increasing the risk of non-reproducible output.



> This could help finding out why a certain constellation of packages resulted as maybe one remembers that a certain package was only published after `<time>` or whatever.

Due to partial updates being supported, I don't think this timestamp will be a good way to debug that either (maybe the last update was a partial one and not a full one).

Instead of relying on `mtime`, you could rely on the VCS history of your lock file.

GromNaN:
In order to make it reproductible, this value could be the max of `packages[].time`.

Toflar:
Yes. Which does not answer the question of when the composer.lock was generated. This is what I want to know and which is impossible to tell atm. Hence the PR. If existing data were enough, I wouldn't be submitting a PR.

stof:
I still have the concern about how this will impact reproducible builds, and how you would rely on it without knowing whether the last lock generation was a full update or a `composer update --lock` that locks all packages to their existing version (and so the selected version actually depend on the previous run of an update, for which you won't have the timestamp either)

Toflar:
Your input about `composer update --lock` is valid, imho. I've changed the implementation to only update the `_time` if the `content-hash` changes (`!$this->isFresh()`). So now, the timestamp only gets updated if the file actualy has relevant changes (it does not, see my notes later on - I just left this part here for you to understand my thought process).

I don't see the problem with partial updates, though. Maybe you want to elaborate? It's just an additional information stored to the lock file and otherwise completely ignored by Composer. Just like the `_readme`. It doesn't have any influence on the `content-hash`. If you do a partial update.......

......okay, I tested this and I now see what you mean. I've always thought the `content-hash` was the hash of the `composer.lock` content and it would update if you change the `composer.lock` (= also during a partial update) 🤣 I thought it's there to compare two `composer.lock` files easily. I only now noticed that this is not the case and this is in fact comparing the `composer.json` just to see if they are out of sync. So yeah, a partial update would have updated the `composer.lock` but not `_time` which of course was not my intention. So that's independent from the `content-hash` and `$this->isFresh()` then. I've updated the implementation again in a second commit.

GromNaN:
This feature will produce conflicts systematically.

Toflar:
Sorry, @GromNaN but these comments are pretty useless. You have to elaborate on what exactly won't work. Two `composer.lock` files always generate conflicts if they changed. And if they didn't, you don't have a new `_time` value.

GromNaN:
Mea culpa, I was too quick. 
If the goal is to know when the dependencies were last updated, it should be possible to change it without changing composer.lock, and thus mark the dependencies as up to date with the new state of the available version at this new date.

For this purpose, the `main' branch would receive regular updates of `_time'. In parallel, a feature branch could add a new dependency and update the `_time` with a different value. That would result in a merge conflict. Similarly, if 2 branches make minor updates of distinct dependencies, they will have a conflict on `_time` while there is no conflict otherwise.

Toflar:
No worries. So let me try to understand your issue: So you have two branches with two different `composer.lock` files because you were working on them simultaneously. Hence, there will be a merge conflict. So far so good.
But what I don't understand: What's the difference with this PR now? Your use case will always generate a merge conflict, no? It has been for years. All that would happen is that on top of differences in the `packages` etc. array, you'd also have an updated `_time` value which you need to merge. Which - in my opinion - can even be very helpful because you can see which of the two branches updated their `composer.lock` before/after the other. Or maybe I still don't quite get the problem :)

Seldaek:
It doesn't always generate conflicts no, if you do updates in two branches for examples that update some of the same packages, or even different ones, as long as the content hash doesn't change you wouldn't get a conflict. If you do the same change in two places you would get the same hash but possibly different modification times, so this would definitely increase the change of conflicts.. all for a use case (non-VCS'd project) which is IMO pretty fringe and bad anyway. So I'd tend to say no here.

Toflar:
I personally don't see an issue in having those conflicts (I don't understand the fuzz about conflicts, people treat them as if they were something evil and the world was collapsing on their heads when really all they need to do is just use a good IDE and pick the right lines from left and right) but I sense the general opposition here. So I'm closing the PR.

Seldaek:
> people treat them as if they were something evil and the world was collapsing on their heads

Yes. It's a flow interruption even if you know how to handle them, but the reality is it does confuse lots of people and it can result in bad merges and mistakes happening, so IMO it's best to reduce them, no matter how much I think people should learn git more.

PR #11386, opened 2023-03-20
Composer support string

glaubinix:
Fixes `getComposerInformation` call if `support` is not an array but instead a string for example:

```
{"name": "acme/package", "support": "https://github.com/org/repo"}
```

Error
```
    "exception": {
        "class": "TypeError",
        "message": "Cannot access offset of type string on string",
        "code": 0,
        "file": "src\/Composer\/Repository\/Vcs\/GitHubDriver.php:179",
        "trace": [
            "src\/Composer\/Repository\/VcsRepository.php:248",
```


Seldaek:
Thanks!

Issue #11341, opened 2023-02-18
Tabs in `composer.lock` if `composer.json` indented using tabs


maximal:
I like the way NPM handles lock files indentation: if `package.json` uses tabs then `package-lock.json` also will be indented using tabs, if the former has spaces then the latter will have spaces.

Maybe it would be semantically correct to implement the similar behavior in Composer, so that `composer.lock` was indented the way `composer.json` is.

Also, this issue is linked to #3192

Seldaek:
Yeah I guess we could do that, sounds reasonable although a little work for not much value IMO as the lock file should not really be opened/edited.

maximal:
@Seldaek, at least users using tabs will have much smaller `composer.lock` :-)

I guess, I could do this, I only need to find where the JSON files generation is located.

maximal:
I taught `JsonFile` to preserve indentation when writing files. See #11390.
Maybe this behavior will be enough. Can you check the pull request please?

So, it is sufficient to once reformat `composer.lock` using tabs and then Composer will be preserving tabs in it.

This is not what I originally wanted (in this issue specifically), but at least something to move on. I don’t know yet how to pass `composer.json` indentation to `Locker` instance.

Seldaek:
Closing as IMO https://github.com/composer/composer/pull/11390 is good enough - if you care about this (and not many do I suppose as this is really a machine-readable file mostly) then change your lock file once to tabs by hand.

PR #11390, opened 2023-03-21
Preserve indentation when writing JSON files

maximal:
Refs #11341, doesn’t resolve specifically it though, but has something to move on.

UPD: I’ll fix PHPStan warnings if the PR is OK in general.


Seldaek:
Yeah that seems reasonable to me

src/Composer/Json/JsonFile.php
```diff
@@ -262,15 +268,28 @@ public static function validateJsonSchema(string $source, $data, int $schema, ?s
...
         if (false === $json) {
             self::throwEncodeError(json_last_error());
         }
 
+        if (($options & JSON_PRETTY_PRINT) > 0 && $indent !== self::INDENT_DEFAULT ) {
+            // Pretty printing and not using default indentation
+            return Preg::replaceCallback(
+                '#^ {4,}#m',
+                static function ($match) use ($indent): string {
+                    return str_repeat($indent, (int)(strlen($match[0] ?? '') / 4));
```
stof, line 287: if the number of spaces is not a multiple of 4, we should probably keep the remaining spaces (as those would not be about indentation but about alignment)

stof (commented):
it would be great to add tests covering that behavior

src/Composer/Json/JsonFile.php
maximal: Mathematically speaking, number of tabs must always be equal of indent level, and then any number of spaces for alignment (aka [SmartTabs](https://www.emacswiki.org/emacs/SmartTabs) approach), but I think here we can simplify the process as we are dealing with `json_encode()` result which can have either 4 or 0 spaces for indentation and have no alignment at all.

Seldaek:
Thanks

PR #11398, opened 2023-03-26
Update 01-basic-usage.md

theking2:
A composer file could be empty. The example with monolog is just and example.


Seldaek:
Thanks for the PR but I think I'd rather not add confusion at this point in the docs, this works for most projects and those that really don't require anything can kinda skip it.

PR #11418, opened 2023-04-06
PlatformRepository library support for libpq (from ext-pq) and librdkafka (from ext-rdkafka)

dzuelke:
Self-explanatory :)

Both info sources have been around in the respective extensions for many, many years, see https://github.com/m6w6/ext-pq/commit/6840072b9e4fe5269d0f36a9cb8cd9cde04525af (2015) and https://github.com/arnaud-lb/php-rdkafka/commit/a38b32e43c68dd796595c4daa44c487fc8acd117 (2016).


Seldaek:
Thanks

PR #11442, opened 2023-04-29
Update 01-basic-usage.md

pravintargaryen:
Removed a comma after monolog/monolog


Seldaek:
Thanks but it is fine as is IMO

PR #11449, opened 2023-05-02
Optimize PoolBuilder to not load replaced targets if not required

Toflar:
Another attempt to fix https://github.com/composer/composer/pull/10410 and https://github.com/composer/composer/pull/9619.

The goal here is to not load replaced packages into the pool if no other package actually requires them. We have to treat all `replace` definitions as "optional packages".

/cc @driskell @stof @naderman @Seldaek

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -639,6 +639,8 @@ class PoolBuilder
                     // make sure that any requirements for this package by other locked or fixed packages are now
                     // also loaded, as they were previously ignored because the locked (now unlocked) package already
                     // satisfied their requirements
+                    // and if this package is replacing another that is required by a locked or fixed package, ensure
+                    // that we load that replaced package in case an update to this package removes the replacement
                     foreach ($request->getFixedOrLockedPackages() as $fixedOrLockedPackage) {
                         if ($fixedOrLockedPackage === $lockedPackage) {
                             continue;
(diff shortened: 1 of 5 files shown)


Toflar:
@driskell compared to your PR this should get rid of **most** of the 404 calls but not all of them. Some might still be there because some package actually **does** `require` one of those replaced packages.

Toflar:
Note to self/mergers: This should definitely be `Co-authored-by` @driskell :)

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -649,13 +658,43 @@ private function unlockPackage(Request $request, array $repositories, string $na
                             if (isset($requires[$lockedPackage->getName()])) {
                                 $this->markPackageNameForLoading($request, $lockedPackage->getName(), $requires[$lockedPackage->getName()]->getConstraint());
                             }
+
+                            foreach ($lockedPackage->getReplaces() as $replace) {
+                                if (isset($requires[$replace->getTarget()], $this->skippedLoad[$replace->getTarget()])) {
+                                    $this->unlockPackage($request, $repositories, $replace->getTarget());
+                                    // Mark as optional - if no other package requires it, we don't need to load it
+                                    $this->markPackageNameForOptionalLoading($replace->getTarget());
```
naderman, line 666: Logically this doesn't make any sense to me, why not just call `markPackageNameForLoading`. Clearly here `$fixedorLockedPackage` is already in `$this->packages` and its requires stored in `$requires` contain a reference to `$replace->getTarget()` so `loadOptionalPackages` is guaranteed to later call `markPackageNameForLoading` anyway, so what's the point of putting it in some optional array first?

Toflar: Right, cut short in https://github.com/composer/composer/pull/11449/commits/9ced20fd0df986ed930def71bff76bd7f9fa8157

naderman (approved):
Think this makes sense now!

Seldaek:
Thanks everyone for finally getting this through 👍🏻

PR #11450, opened 2023-05-02
Remove optional package loading to simplify pool builder

naderman:
Instead we just directly check if any currently loaded package requires the package to be loaded optionally.


Toflar:
I did it the other way around to loop over `$this->packages` only once but I guess at this point it's not a big deal.

naderman:
Yeah I think since this is now only for deps of the allow-listed packages and `$this->packages` should mostly contain locked/fixed packages at that point it really shouldn't make a measurable performance difference.

Toflar (approved):

PR #11453, opened 2023-05-05
Add support for env vars and ~ (for HOME) in repo paths for vcs/artifact repos

### LINKED ISSUE #11409: Environment variable not work with local repo
Composer ver.2.5.5, Win10

Environment variables work in Windows when I use local package with "type": "path", but not work with local repo with "type": "vcs".

My `composer.json`:

```json
    "repositories": [
    {
            "type": "vcs",
            "url": "d:/Rabota/PHP/_packages_/alex"
        }
    ],
    "require": {
        "alex/alex": "dev-master"
    }
```
work fine.

But (when MY_PHP_PATH=d:\Rabota\PHP)

```json
    "repositories": [
    {
            "type": "vcs",
            "url": "$MY_PHP_PATH/_packages_/alex"
        }
    ],
    "require": {
        "alex/alex": "dev-master"
    }
```
not work.

```
In VcsRepository.php line 191:

  No driver found to handle VCS repository $MY_PHP_PATH/_packages_/alex
```

Seldaek:
Fixes #11409

### DIFF
--- a/src/Composer/Repository/ArtifactRepository.php
+++ b/src/Composer/Repository/ArtifactRepository.php
@@ -17,6 +17,7 @@ use Composer\Json\JsonFile;
 use Composer\Package\BasePackage;
 use Composer\Package\Loader\ArrayLoader;
 use Composer\Package\Loader\LoaderInterface;
+use Composer\Util\Platform;
 use Composer\Util\Tar;
 use Composer\Util\Zip;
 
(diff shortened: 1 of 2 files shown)


stof:
should this have tests ?

Also, you probably need to update the `ValidRepositoryValidator` of Packagist to reject URLs starting with a `~` before merging this, to make sure that Packagist will not end up expanding `~` to the HOME on the server.

Seldaek:
Ah yes good point I'll make sure this is fine on the packagist end. Tests I don't think it's really worth it here..

PR #11459, opened 2023-05-09
Update 05-repositories.md

JRyven:
The verb phrase is 'internally' is an adverb describing the sort of seeing that Composer is doing. IE: Composer internally sees. The sentence _is_ much clearer when the adverb is moved earlier in the sentence, but this mutation necessitates a comma separating the adverb from the noun.


Seldaek:
Thanks

PR #11460, opened 2023-05-10
feat: allow installation of packages only from trusted vendors 

dunglas:
This patch adds a new feature to allow the installation of packages only from trusted vendors.
If a required package (whether it is a direct or a transient dependency) is not in the "trusted" allow list, the installation will fail.

This option is especially useful to prevent [supply chain attacks](https://dunglas.dev/2018/11/about-the-dependencies-of-symfony/).
It allows detecting when dependencies coming from new, untrusted vendors are installed, for instance, when
a dependency of the root package adds a new transient dependency.

All packages from a specific vendor can be allowed by using the `vendor/*` special syntax.

Example:

```json
{
    "trusted": [
        "api-platform/*",
        "symfony/*",
        "doctrine/*",
        "psr/*",
        "dunglas/doctrine-json-odm"
    ]
}
```

With the previous configuration, Composer will only allow installation of packages provided by the `api-platform`, `symfony`,
`doctrine` and `psr` vendors as well as `dunglas/doctrine-json-odm` (but not other packages from `dunglas`).

We plan to use this new feature in Symfony and API Platform to ensure that our CI fails when a new, unaudited, package pops up because one of our dependencies now requires it.

This is also useful for all other security-sensitive projects when you want to be sure of what is installed and executed.

I first discussed this feature with @Seldaek and @nicolas-grekas during SymfonyCon 2022, and I'll present this feature in-depth on Friday during AFUPday 2023.

### DIFF
--- a/src/Composer/Package/RootPackageInterface.php
+++ b/src/Composer/Package/RootPackageInterface.php
@@ -60,6 +60,34 @@ interface RootPackageInterface extends CompletePackageInterface
     public function getPreferStable(): bool;
 
     /**
+     * Sets the list of trusted packages.
+     *
+     * @param string[] $trusted
+     */
+    public function setTrusted(array $trusted): void;
+
+    /**
+     * Returns the list of trusted packages (a wildcard can be used to trust all packages from a specific vendor).
+     *
+     * @return string[]
+     */
+    public function getTrusted(): array;
+
+    /**
+     * Sets the list of trusted packages in dev.
+     *
+     * @param string[] $devTrusted
+     */
+    public function setDevTrusted(array $devTrusted): void;
+
+    /**
+     * Returns the list of packages trusted in dev (a wildcard can be used to trust all packages from a specific vendor).
+     *
+     * @return string[]
+     */
+    public function getDevTrusted(): array;
+
+    /**
      * Returns the root package's configuration
      *
      * @return mixed[]
--- a/src/Composer/Package/RootPackageInterface.php
+++ b/src/Composer/Package/RootPackageInterface.php
@@ -68,6 +68,7 @@ interface RootPackageInterface extends CompletePackageInterface
 
     /**
      * Returns the list of trusted packages (a wildcard can be used to trust all packages from a specific vendor).
+     * An empty array means that this feature is disabled, all packages can be installed.
      *
      * @return string[]
      */
@@ -82,6 +83,7 @@ interface RootPackageInterface extends CompletePackageInterface
 
     /**
      * Returns the list of packages trusted in dev (a wildcard can be used to trust all packages from a specific vendor).
+     * An empty array means that this feature is disabled, all packages can be installed.
      *
      * @return string[]
      */
--- a/src/Composer/Package/RootPackageInterface.php
+++ b/src/Composer/Package/RootPackageInterface.php
@@ -62,32 +62,32 @@ interface RootPackageInterface extends CompletePackageInterface
     /**
      * Sets the list of trusted packages.
      *
-     * @param string[] $trusted
+     * @param string[]|null $trusted
      */
-    public function setTrusted(array $trusted): void;
+    public function setTrusted(?array $trusted = null): void;
 
     /**
      * Returns the list of trusted packages (a wildcard can be used to trust all packages from a specific vendor).
-     * An empty array means that this feature is disabled, all packages can be installed.
+     * A null value means that this feature is disabled, all packages can be installed.
      *
-     * @return string[]
+     * @return string[]|null
      */
-    public function getTrusted(): array;
+    public function getTrusted(): ?array;
 
     /**
      * Sets the list of trusted packages in dev.
      *
-     * @param string[] $devTrusted
+     * @param string[]|null $devTrusted
      */
-    public function setDevTrusted(array $devTrusted): void;
+    public function setDevTrusted(?array $devTrusted = null): void;
 
     /**
      * Returns the list of packages trusted in dev (a wildcard can be used to trust all packages from a specific vendor).
-     * An empty array means that this feature is disabled, all packages can be installed.
+     * A null value means that this feature is disabled, all packages can be installed.
      *
-     * @return string[]
+     * @return string[]|null
      */
-    public function getDevTrusted(): array;
+    public function getDevTrusted(): ?array;
 
     /**
      * Returns the root package's configuration
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -482,6 +482,41 @@ Example:
 }
 ```
 
+#### trusted <span>([root-only](04-schema.md#root-package))</span>
+
+The list of trusted packages that can be installed.
+If a required package (whether it is a direct or a transient dependency) is not in
+the "trusted" allow list, the installation will fail.
+
+This option is especially useful to prevent [supply chain attacks](https://dunglas.dev/2018/11/about-the-dependencies-of-symfony/).
+It allows detecting when dependencies coming from new, untrusted vendors are installed, for instance, when
+a dependency of the root package adds a new transient dependency.
+
+All packages from a specific vendor can be allowed by using the `vendor/*` special syntax.
+
+Example:
+
+```json
+{
+    "trusted": [
+        "api-platform/*",
+        "symfony/*",
+        "doctrine/*",
+        "psr/*",
+        "dunglas/doctrine-json-odm"
+    ]
+}
+```
+
+With the previous configuration, Composer will only allow installation of packages provided by the `api-platform`, `symfony`,
+`doctrine` and `psr` vendors as well as `dunglas/doctrine-json-odm` (but not other packages from `dunglas`).
+
+#### trusted-dev <span>([root-only](04-schema.md#root-package))</span>
+
+List of trusted packages that can be installed for developing.
+The syntax is the same as for the "trusted" property. Packages and vendors listed in the "trusted" can also always be
+installed for developing.
+
 ### autoload
 
 Autoload mapping for a PHP autoloader.
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -517,6 +517,8 @@ List of trusted packages that can be installed for developing.
 The syntax is the same as for the "trusted" property. Packages and vendors listed in the "trusted" can also always be
 installed for developing.
 
+Setting "trusted-dev" does nothing if "trusted" is not set.
+
 ### autoload
 
 Autoload mapping for a PHP autoloader.
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -666,6 +666,20 @@
             "additionalProperties": {
                 "type": "string"
             }
+        },
+        "trusted": {
+            "type": "array",
+            "items": {
+                "type": "string",
+                "description": "A list of trusted packages (wildcard supported to trust an entire vendor)."
+            }
+        },
+        "trusted-dev": {
+            "type": "array",
+            "items": {
+                "type": "string",
+                "description": "A list of trusted packages in development (wildcard supported to trust an entire vendor)."
+            }
         }
     },
     "definitions": {
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -428,8 +428,8 @@ class Installer
      *
      * @return string[]
      */
-     private function getUntrustedPackages(array $packages, array $patterns): array {
-         if (count($patterns) === 0) {
+     private function getUntrustedPackages(array $packages, ?array $patterns): array {
+         if ($patterns === null) {
              return [];
          }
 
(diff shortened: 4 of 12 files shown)


nicolas-grekas:
Awesome, thanks for starting this!
Let's start with some bideshading :) What about nesting this under `config`?

src/Composer/Package/RootPackageInterface.php
```diff
@@ -59,6 +59,34 @@ public function getReferences(): array;
...
 
+    /**
+     * Sets the list of trusted packages.
+     *
+     * @param string[] $trusted
+     */
+    public function setTrusted(array $trusted): void;
+
+    /**
+     * Returns the list of trusted packages (a wildcard can be used to trust all packages from a specific vendor).
```
stof, line 70: Apparently, the code currently treats an empty array as meaning that anything is trusted. This must clearly be explained in the phpdoc (also, what happens if you configure only `trusted-dev` ?)

vtsykun:
Composer already has packet filtering function

```json
{
    "repositories": [
        { "packagist": false },
        {
            "type": "composer",
            "url": "https://repo.packagist.org",
            "only": [
                "api-platform/*",
                "symfony/*",
                "doctrine/*",
                "psr/*",
                "dunglas/doctrine-json-odm"
            ]
        }
    ]
}
```

src/Composer/Package/RootPackageInterface.php
dunglas: PHPDoc updated. Configuring only `trusted-dev` makes no sense, and this doesn't do anything. I updated the doc to make it explicit.

dunglas:
@vtsykun nice hack! I wasn't aware of this feature. Is it documented?

Even if packet filtering mostly do the job, I still think that the "trusted" property has some benefits over it:

1. the error is more readable:

  With packet filtering:
  
  ```
  
    Problem 1
      - Root composer.json requires dunglas/foo * -> satisfiable by dunglas/foo[1.0.0].
      - dunglas/foo 1.0.0 requires symfony/bar * -> could not be found in any version, there may be a typo in the package name.
  
  Potential causes:
   - A typo in the package name
     see <https://getcomposer.org/doc/04-schema.md#minimum-stability> for more details.
   - It's a private package and you forgot to add a custom repository to find it
  ```
  
  With my "trusted":
  
  ```
  Updating dependencies
  The installation of the following packages have been requested, but are untrusted: "symfony/bar".
  ```

2. It supports filtering both prod and dev packages (filtering dev packages is also useful because we saw some supply chain attacks targeting especially developers' computers)
3. The syntax is more explicit and user-friendly and will probably help the adoption of this practice.

WDYT?

stof:
> 2\. It supports filtering both prod and dev packages (filtering dev packages is also useful because we saw some supply chain attacks targeting especially developers' computers)

Actually, filtering the repository also filters dev packages (what it cannot do is providing separate lists) as all packages have to be found in a repository.

However, relying on repository filters for that has another drawback in addition to the less clear error message: it also becomes very tricky to maintain when your `repositories` setting is not _only_ about forbidding untrusted packages.

nicolas-grekas:
> put it in config depends on whether you intend to allow configuring this system-wide through the global composer config

putting this system-wide makes sense to me, so I'm all for nesting this under options :)

stof:
@zimzat this is a root-only setting. It is for a project to define its own trusted list.

Merging trust lists from dependencies would make no sense as it would mean that a project would not be in control of its trust list (at this point, you are at the same level of protection that with the existing behavior that has no trust list)

> If the attacker has the ability to update a package's `composer.json` then they also have the ability to set their own `trusted` list within that package; how would the system prevent the attacker from trusting themselves?

If the attacker has the possibility to alter the root composer.json of your project, they can probably also modify other project files and so they can already inject malicious code in your project without relying on a malicious composer dependency. That's something to be solved by protecting access to your git repo, not by using a composer feature.

> How would this stop someone who gains control of a package maintainer account from injecting malicious code directly into the package source itself?

This feature is not preventing that. This feature is not meant to prevent _all_ attacks. It adds one tool in the defense.

zimzat:
I don't follow how this trust list helps _in practice_.

Symfony sets it on their packages and they get to feel good that during development they're not vulnerable to supply chain attacks but anyone who uses their packages still are?

Symfony sets it in their framework skeleton, it confirms the skeleton only has the 19(?) packages they list, but the moment anyone wants to install `ramsey/uuid` they have to update it to include `brick/math`, and then when they install `aws/aws-sdk-php` they have to update it again to include `aws/*` and `guzzlehttp/*` and `mtdowling/*` and `psr/*`. Or if it doesn't auto-add those dependencies to their trusted list then any time they do `composer update` it will block _all_ versions until they do. How did that make their experience more secure if they didn't trust the package they are installing in the first place? At some point they will get annoyed and delete the entire list so it stops warning them every time they want to install a new package or update their existing packages. 

Isn't this a verbatim copy-paste of package.lock? Having a "warn when installing new dependent packages" toggle (similar to the "run scripts?" prompt) is just as secure and less friction on the user.

I'm sorry if it sounds like I'm being antagonistic; I'm really baffled about the benefit vs cost of this specific implementation and don't know how else to phrase it.

dunglas:
@zimzat even if it can be useful for all kind of projects, this feature is mainly designed for maintainers of popular FOSS libraries, frameworks and projects.

For instance, `symfony/html-sanitizer` depends on `masterminds/html5`, a package we trust but that we don't control.

If, at some point, Masterminds adds a new dependency to their HTML5 package coming from a third party vendor (and this package may require other dependencies in cascade), thanks to this new feature the Symfony CI will fail, and we will notice these new vendors we have to trust.

We'll have the opportunity to decide if we trust these new packages, or if we prefer to remove our dependency on `masterminds/html5` (or discuss with the maintainers of Masterminds HTML5 to see if they are OK to remove the new dependencies we don't trust).

Yes, during a (most likely short) window, end users will download the new dependencies anyway. But at least we'll be able to notice the change and react accordingly.

This is even better for end users projects shipping a `composer.lock` file (ex: Magento). Thanks to this feature they will detect the Magento team will be able to detect that one of their dependencies introduced a new dependency. Before tagging a new release of Magento, they will be able to decide if it's OK or not. And it's not, no end user will get this new dependency.

vtsykun:
I created an alternative to this PR which also protects against a recent attack on packagist.org https://blog.packagist.com/packagist-org-maintainer-account-takeover/.

stof:
it also becomes cumbersome if you start using Private packagist, as zip archives will then be downloaded from a different source for the same packages thanks to mirrored archives.

vtsykun:
> it also becomes cumbersome if you start using Private packagist, as zip archives will then be downloaded from a different source for the same packages thanks to mirrored archives.

you need to add only the url of your organization `https://packagist.com/orgname/` to allow downloading from the mirror. 

```
    "config": {
        "allowed-urls": [
            "https://packagist.com/orgname/",
```


> The developer experience is less good for the use case of trusting packages (you have to dig into the internals of the packages you install to find out the URLs of their repositories)

This can be done by adding a new command or updating the command `composer info` 

Actually, this is analog of the Content Security Policy for Composer

zimzat:
@dunglas Thank you for taking the time to explain the specific usage in mind.

While I still have some reservations about this implementation, mainly around the usage of the word "trusted" and the lack of any actual cryptographic backing to that word, it does provide the mechanism you're looking for to create an allowed list of indirect vendors or packages that is separate from the existing lock file's version-specific constraint. I just hope this doesn't create a false sense of security since it doesn't provide any actual security.

The interactive mode to review and allow the transitive dependencies can be done in a separate pull request, though IMHO it should be designed with that in mind now (to avoid data structure lock-in); when I mentioned this feature to my boss they were eager to start using it immediately so I would expect it to get used widely and quickly. If users get burnt by the friction without seeing a benefit they may disable it and then never revisit it once it's easier to use.

For projects that commit their lock file, the `apt` style "This will install X, and also install Y, Z" or "This will upgrade X, and also install Y, Z" + "Do you want to continue [Y/n]?" (see: https://unix.stackexchange.com/a/70652) provides just as much awareness and security but with almost no friction for the average project. That would still be nice to see too.

Thanks!

C-Duv:
In the end, for various reasons (trusts, internal policies, security …) one would like to avoid installing packages for production and/or development.
Globally or by repositories.

It's either:
* I never want package/vendor *x*
* I don't wan't package/vendor *x* unless it's used by package *y* (or a package from vendor *z*)

The logic could be extended to licensing.

To be dev-friendly the rules should be clear, at a single place and must tell why package is forbidden or allowed (so that the `composer` command can print why at package is blocked).

Something like:

```json
{
    …
    "filtering":
        "only": [
            "vendor/*",
            "vendor2/foo",
            "license": [
                "MIT",
                "Apache-2.0"
            ]
        ],
        "never": {
            "vendor3/*": "trust",
            "vendor4/bar": "security",
            "vendor5/qux": "update-policy"
            "license": [
                "proprietary",
                "WTFPL"
            ]
        },
        "allowed": {
            "vendor/*": "trust-transitive",
            "vendor2/foo": trust",
        },
    ],
}
```

I guess theses rules should not determine how Composer determine which packages matches the `require` rules for installation but only applies a final GO/NO-GO message to the developper.

dunglas:
@zimzat what do you think about "allowed" instead of "trusted" as suggested by @C-Duv?

Do the maintainers have any preference regarding all these options?

naderman:
Here are some problems I see with this proposal:

## May communicate a false sense of security

These settings give the idea that vendor names are somehow "owned" by someone, and nobody else can publish packages under that vendor name. But that's only true on packagist.org, any additional Composer repository you add to composer.json may define any packages in any vendor name without checking who owns these vendor names on packagist.org.

When adding additional repositories you need to be really sure you trust them. But the proposed feature may give a wrong idea of safety in this situation as users may think ...... For context, this situation is what the repository filters like "only" were mainly built for: If I want to add a third party repository to install a package, e.g. a private plugin for my CMS, then I may want to restrict the operator of that repository to packages with their specific vendor prefix, so they don't decide to just ship their forked copy of e.g. symfony/foo to me.

## Does not protect projects using frameworks applying these measures

As this is supposed to apply to the root project only, users of a framework like Symfony would have to manage this list of dependencies themselves. Symfony having this list only provides limited protection, as functionality such as "replace" may modify what dependencies of symfony are actually pulled in an overall project which has other requirements.

## Does it really make sense from a trust perspective?

If you trust a dependency enough to install and use their code without manually verifying all changes - and you could theoretically do so for dependencies using lock file diffs, too - then why wouldn't you trust them to define dependencies? I'm not sure I quite see the threat model this protects from? If you are trying to protect from maintainers adding bad third party dependencies, should you trust the packages these maintainers publish in the first place?

## Too cumbersome to manage

As a root only feature (and I really don't see how one could reasonably implement this in any other way), it means that you need to have a detailed understanding of all your dependencies' dependencies. You cannot use this feature and just require a symfony framework package and trust symfony to make these decisions about their dependencies for you.

zimzat:
@dunglas I agree that "allowed" is better; the full proposal of also having "never" and "only" under the concept of "filtering" sounds great but also opens the door to a lot of edge cases so I'd want to have a closer look at how other systems implement those concepts to avoid architecture pitfalls. Being able to block a dependency while still installing the main package sounds useful in theory but also sounds like it would generate a lot of pain if used casually or occurs without explicit warning.

You could make this a composer plugin or stand-alone phar which projects could manually install. That wouldn't have the same effect of blocking installation prior to the first use but it would provide a mechanism to use in CI to make the same check and alert for Symfony/et.al. that is the intended scope for this feature.

nicolas-grekas:
@naderman I think most/all of your arguments make sense for final apps, but they don't apply for libraries: from a package author perspective, having an allow-list of vendors makes sense and would be useful to better follow the package principles. As a library author, I want to know when one of my dependencies adds a new vendor. That would give me an opportunity to audit this new vendor and have a conversation if I see any issues with e.g. some maintenance policies. Basically, this would help answer "yes" with more confidence to your question "should you trust the packages these maintainers publish in the first place?".

naderman:
@nicolas-grekas And why do you then not simply commit a lock file and check if any new dependencies showed up when you do an update on your library? Is this tool here really needed? Part of the issue is that you're always just looking at one resolution, you don't really know if some combination of versions of dependencies could somehow still add the new dependency, but your particular set of installed dependency versions doesn't show it. The lock file suffers from that issue as much as this proposed solution here.

To be honest I think what you are describing would probably be best built as a separate tool that analyses your composer.json and then all versions that are compatible for a list of required names/versions so you can be sure that none of them end up introducing a new dependency for your users, which you as a library author with a single  install/update may not even notice.

stof:
Committing the lock file in Symfony is a pain when the CI runs with different PHP versions (which might resolve different versions od deps) and also trying to test against the lowest and the highest dependencies. It also would not test against the version that our users would actually use most of the time unless we have an automated way to keep that lock file uptodate.

naderman:
@stof I'm not sure what we're arguing here, this doesn't change that the proposal here doesn't accomplish what it claims to accomplish, so in my eyes is misleading and would give users of this features a false sense of security. It only reliably works for cases that have a single lock file, and there you don't really need this cause you have the lock file?

nicolas-grekas:
For the record, this command generates a file that lists all vendors, with a `+` in front of transitive ones:
```sh
diff -u50 <(composer show -D | cut -d/ -f1 | uniq) <(composer show | cut -d/ -f1 | uniq) | tail -n +4
```

For example on Symfony, the output is:
```
 amphp
 async-aws
+brick
 cache
+daverandom
 doctrine
 dragonmantank
 egulias
 friendsofphp
 guzzlehttp
+kelunik
+laminas
+lcobucci
 league
+lorenzo
 masterminds
 monolog
 nyholm
+paragonie
 pda
 php-http
 phpdocumentor
 phpstan
 predis
 psr
+revolt
+spomky-labs
 symfony
+tijsverkoyen
 twig
 web-token
+webmozart
```

Seldaek:
Finally caught up with the discussion here, and I have to say I agree with what @naderman [wrote above](https://github.com/composer/composer/pull/11460#issuecomment-1547507135).

Thanks for the high quality PR with tests and docs though @dunglas - it always pains me to have to decline such PRs, but yet here we are. If someone really wants this feature it could easily be built as a third party utility, but I'd rather not ship something we are not convinced of.

PR #11462, opened 2023-05-12
The package URLs change protection

vtsykun:
This PR adds a new feature to allow the download the code only from trusted sources.

It gives protection from attacks, like change url or supply chain attacks

1) If packagist.org maintainer account takeover https://blog.packagist.com/packagist-org-maintainer-account-takeover/
2) If the packagist.org secret was leaked
3) Dependency injection attack #11460 

Example usage

```json
    "config": {
        "allowed-urls": [
            "https://api.github.com/repos/dragonmantank/cron-expression/",
            "https://api.github.com/repos/php-fig/",
            "https://api.github.com/repos/symfony/" // the all symfony packages
        ]
    },
```
Installation will fail if untrusted source url of package is found. it is supported for dist/source/mirrors downloads

```
  The downloading of the following url have been requested, but is untrusted: "https://api.github.com/repos/webmozarts/assert/zipball/11cb2199493b2f8a3b53e7f19068fc6aac760991"
```

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -582,6 +582,13 @@
                 "platform-check": {
                     "type": ["boolean", "string"],
                     "description": "Defaults to \"php-only\" which checks only the PHP version. Setting to true will also check the presence of required PHP extensions. If set to false, Composer will not create and require a platform_check.php file as part of the autoloader bootstrap."
+                },
+                "allowed-urls": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "description": "A list of URL prefixes, from where Composer can download the source code."
+                    }
                 }
             }
         },
--- a/src/Composer/Downloader/DownloadManager.php
+++ b/src/Composer/Downloader/DownloadManager.php
@@ -38,6 +38,8 @@ class DownloadManager
     private $filesystem;
     /** @var array<string, DownloaderInterface> */
     private $downloaders = [];
+    /** @var array<string>|null */
+    private $allowedUrls = null;
 
     /**
      * Initializes download manager.
@@ -125,6 +127,30 @@ class DownloadManager
     }
 
     /**
+     * Set list of trusted url prefixes.
+     *
+     * @param array<string>|null $allowedUrls array of trusted url patterns
+     *
+     * @return DownloadManager
+     */
+    public function setAllowedUrls(array $allowedUrls = null): self
+    {
+        $this->allowedUrls = $allowedUrls;
+
+        return $this;
+    }
+
+    /**
+     * Return list of trusted url prefixes.
+     *
+     * @return array<string>|null
+     */
+    public function getAllowedUrls(): ?array
+    {
+        return $this->allowedUrls;
+    }
+
+    /**
      * Returns downloader for already installed package.
      *
      * @param  PackageInterface          $package package instance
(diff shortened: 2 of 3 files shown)


stof:
Installing from sources bypasses any restriction you configure currently...

res/composer-schema.json
```diff
@@ -582,6 +582,13 @@
                 "platform-check": {
                     "type": ["boolean", "string"],
                     "description": "Defaults to \"php-only\" which checks only the PHP version. Setting to true will also check the presence of required PHP extensions. If set to false, Composer will not create and require a platform_check.php file as part of the autoloader bootstrap."
+                },
+                "allowed-urls": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "description": "A list of URL prefixes, from where Composer can download the source code."
```
dunglas, line 590: Instead or in addition to prefixes, it could be nice to support the [URL pattern syntax](https://wicg.github.io/urlpattern/) for more flexibility.

vtsykun:
> Installing from sources bypasses any restriction you configure currently...

this also works for source downloads. ` --prefer-source`
```
../composer/bin/composer install --prefer-source
                                                                                                                                 
  The downloading of the following url have been requested, but is untrusted: "https://github.com/php-fig/event-dispatcher.git"
```

dunglas:
Wouldn't a signature-based solution similar to [what NPM introduced recently](https://github.blog/2023-04-19-introducing-npm-package-provenance/) be a better and more user-friendly solution to mitigate this kind of attack?

stof:
Given that Packagist does not host the code archives, we cannot have the equivalent feature of cryptographically signing a proof of where those archives where generated to upload them to Packagist.

dunglas:
@stof couldn't we do something like that:

1. Store private keys as secrets accessible by the CI system
2. Store the corresponding public keys on Packagist
3. Allow importing the public keys in the local `composer.json` file (to prevent a Pakcagist maintainer account takeover)
4. Compute the signature in the CI system when a release is tagged, and include the signature in the API call creating the  version on Packagist

If I'm not missing something, this should work even if sources are not stored in Packagist.

res/composer-schema.json
vtsykun: I agree, but maybe more better to use wildcard asterisk, like in CSP rules

stof:
@dunglas there is no API call creating versions on Packagist. Packagist versions are managed automatically based on available branches and tags in the source VCS repository (the closest you could find to an API call is the usage of github webhooks to notify packagist when a push is made in the repository, to trigger a re-processing of that package in Packagist).

Spomky:
> Wouldn't a signature-based solution similar to [what NPM introduced recently](https://github.blog/2023-04-19-introducing-npm-package-provenance/) be a better and more user-friendly solution to mitigate this kind of attack?

Hi,

Actually, this is what I had in mind at first glance.
As git supports signed commits and tags, I imagined a way where the list of trusted public key thumbprints is available and last commit or tag signature is verified when installed by composer.
One issue though: mono-repo/many-repo split using gitsplit loose the signatures (see [here](https://github.com/jderusse/docker-gitsplit/issues/29) or [here](https://github.com/splitsh/lite/issues/52)).

Seldaek:
Thanks but I'd rather not do this to keep things simpler, and as we now restrict url editing on packagist for packages with high download counts it should prevent any such attacks with a major impact.

PR #11470, opened 2023-05-19
 Ensure stripos() receives a string in Platform::isWindowsSubsystemForLinux()

prikkprikkprikk:
`composer install` was failing on our server while installing codesniffer:

```
  - Installing squizlabs/php_codesniffer (3.7.2): Extracting archive
    Install of squizlabs/php_codesniffer failed

In Platform.php line 151:

  [TypeError]
  stripos(): Argument #1 ($haystack) must be of type string, bool given
```

Platform::isWindowsSubsystemForLinux() checks the contents of the file **/proc/version** on the system, but if this file is not found (like on our server, which uses a customized Linux distro, which I guess is the problem), file_get_contents() returns false and causes stripos() to throw a TypeError due to the introduction of strict_types in composer v2.3.0.

This PR fixes the issue by checking if the return value from file_get_contents() is false before using stripos() on it.

I have tested that `composer install` still works on my local environment (where this wasn't an issue) and that the issue has been fixed on our server.

Also, I took the liberty to refactor the rest of the logic a little bit so that it just uses one return statement after setting the `self::$isWindowsSubsystemForLinux` property.

### DIFF
--- a/src/Composer/Util/Platform.php
+++ b/src/Composer/Util/Platform.php
@@ -148,7 +148,7 @@ class Platform
             if (
                 !ini_get('open_basedir')
                 && is_readable('/proc/version')
-                && false !== stripos(Silencer::call('file_get_contents', '/proc/version'), 'microsoft')
+                && false !== stripos((string)Silencer::call('file_get_contents', '/proc/version'), 'microsoft')
                 && !file_exists('/.dockerenv') // docker running inside WSL should not be seen as WSL
             ) {
                 return self::$isWindowsSubsystemForLinux = true;


fredden:
If I'm understanding correctly, on your server the call to `is_readable('/proc/version')` returns `true`, but `file_get_contents('/proc/version')` returns `false`? That's an unexpected situation. Perhaps the cause of that mismatch needs to be investigated and resolved, rather than making changes to this project.

I wonder if a simpler fix here would be to cast the value from `file_get_contents()` to always be a string.

prikkprikkprikk:
> If I'm understanding correctly, on your server the call to `is_readable('/proc/version')` returns `true`, but `file_get_contents('/proc/version')` returns `false`? That's an unexpected situation. Perhaps the cause of that mismatch needs to be investigated and resolved, rather than making changes to this project.

Good point! I will contact our server provider about this. It might be a matter of misconfiguration of their custom distro.

> I wonder if a simpler fix here would be to cast the value from `file_get_contents()` to always be a string.

Agreed, a much simpler fix!

Even though it's unexpected behavior, it has been proved (by my case) that it might happen.

I have made changes to this PR accordingly.

fredden (approved):

Seldaek:
Thanks, merged in 2.5 as https://github.com/composer/composer/commit/9885d23e2a0aeee3079063059ab71e73fa7f5962

PR #11476, opened 2023-05-23
Tests for empty or absent repository filters

dzuelke:
The behavior in `FilterRepository` is currently correct, but not explicitly tested.

Also add a test that ensures both filters can't be there simultaneously.


Seldaek:
Thanks

PR #11478, opened 2023-05-24
Add FAQ entry for requiring specific commit or better stable version

jpmschuler:
I wondered how to require a specific commit but allow newer stable versions and from my tests it seems it is simple and already working, but I couldn't find any concrete explanation why. The reason though is simple: best match is defined as fulfilling the constraints (version and stability) and then being newest.

However I wasn't able to predict this behaviour at first glance. Thus, I thought an additional FAQ entry might help.

### DIFF
--- a/doc/faqs/how-to-require-commit-but-allow-newer-stable-versions.md
+++ b/doc/faqs/how-to-require-commit-but-allow-newer-stable-versions.md
@@ -1,0 +1,13 @@
+# How to require a specific commit currently, but allow newer stable versions later, too?
+
+Sometimes a newer commit might be availble in the repository with a fix you need, without a new stable version being released. 
+
+So lets say there is a tagged release of `v12.4.4` and you are currently requiring `^12.4.1`, which installs `v12.4.4`.
+But a later commit `fceda69` on the main branch was added, that fixes your problem.
+You can require that commit via e.g. `dev-main#fceda69` however such a constraint would prevent future updates to stable versions.
+
+So instead you could have a require of `^12.4.5 || dev-main#fceda69` so requiring (currently not available) next stable patch version or the specific commit. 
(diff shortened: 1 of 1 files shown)


doc/faqs/how-to-require-commit-but-allow-newer-stable-versions.md
```diff
@@ -0,0 +1,13 @@
...
+
+Sometimes a newer commit might be availble in the repository with a fix you need, without a new stable version being released. 
+
+So lets say there is a tagged release of `v12.4.4` and you are currently requiring `^12.4.1`, which installs `v12.4.4`.
+But a later commit `fceda69` on the main branch was added, that fixes your problem.
+You can require that commit via e.g. `dev-main#fceda69` however such a constraint would prevent future updates to stable versions.
+
+So instead you could have a require of `^12.4.5 || dev-main#fceda69` so requiring (currently not available) next stable patch version or the specific commit. 
+As long as no new version is released, still your specific commit will be the best match. As soon as a `v12.4.5` or a `v12.5.0` is released 
+(on the same branch) the algorithm will detect it is allowed and newer as your specified commit, thus switch to that newer stable version instead.
```
Seldaek, line 11: This is not true. By default dev-main will be preferred as the solver has no clue about the git references and where they sit in the history. If you have `prefer-stable:true` in composer.json tho then yes 12.4.5 would be preferred if available as it is more stable than dev-main, but there is no "detect is it newer than your specified commit" bit.

The better (or at least more concise) constraint IMO would be `^12.4.5@dev`, assuming that the project correctly has a branch-alias on dev-main that aliases it to 12.4.x-dev or 12.x-dev, then Composer will use dev-main or tag, but again this depends on prefer-stable.

jpmschuler:
Closed because too complex and wrong assumptions

PR #11484, opened 2023-05-30
Update entries in `.gitattributes`

drupol:
Hello,

I was working on a Composer plugin when I noticed that many files were in `vendor/composer/composer` and not required for running Composer.

I therefore updated the `.gitattributes` and updated it.

Before the PR, running `git archive --format zip --output zipfile-after.zip main` would produce an archive of 829k.
After the PR, the archive is down to 658k.

If think that given the amount of times composer is installed, this small change makes a difference.

Let me know if it's ok for you.


.gitattributes
```diff
@@ -9,11 +9,19 @@
...
+/phpstan/ export-ignore
+/tests/ export-ignore
+/.editorconfig export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.php-cs-fixer.php export-ignore
+/CHANGELOG.md export-ignore
+/CODE_OF_CONDUCT.md export-ignore
+/composer.lock export-ignore
+/LICENSE export-ignore
```
stof, line 23: Removing the license is a no-go. This would violate the license

stof (changes requested):

Seldaek:
Yeah, I wish people wouldn't require it as much as they do, but anyway here we are so thanks I guess this makes sense.

PR #11485, opened 2023-05-30
Update `SyncHelper` - Remove 3 phpstan issues

drupol:
This PR:

- [x] Get rid of 3 PHPStan issues.

### DIFF
--- a/phpstan/baseline.neon
+++ b/phpstan/baseline.neon
@@ -4934,21 +4934,6 @@ parameters:
 			path: ../src/Composer/Util/Svn.php
 
 		-
-			message: "#^Only booleans are allowed in a ternary operator condition, Composer\\\\Package\\\\PackageInterface\\|null given\\.$#"
-			count: 1
-			path: ../src/Composer/Util/SyncHelper.php
-
-		-
-			message: "#^Only booleans are allowed in an if condition, React\\\\Promise\\\\PromiseInterface\\|null given\\.$#"
-			count: 1
-			path: ../src/Composer/Util/SyncHelper.php
-
-		-
-			message: "#^Parameter \\#2 \\$target of method Composer\\\\Downloader\\\\DownloaderInterface\\:\\:update\\(\\) expects Composer\\\\Package\\\\PackageInterface, Composer\\\\Package\\\\PackageInterface\\|null given\\.$#"
-			count: 1
-			path: ../src/Composer/Util/SyncHelper.php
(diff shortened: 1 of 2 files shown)


phpstan/baseline.neon
```diff
@@ -5610,4 +5595,3 @@ parameters:
 			message: "#^Only booleans are allowed in a ternary operator condition, array\\<string\\> given\\.$#"
 			count: 1
 			path: ../tests/Composer/Test/Util/TlsHelperTest.php
-
```
stof, line 5613: this should not be removed as phpstan generates the baseline with this empty line in it.

drupol: Oops! I will fix this as soon as Github works again on my laptop :)

drupol: I just double checked and there's an empty line already.

image

stof: This is not the empty line on which I commented

drupol: I just noticed, fix is incoming.

drupol: Fixed.

stof (approved):

Seldaek:
Thanks

drupol:
@Seldaek I think we could improve this helper further.

What do you think about replacing the `DownloaderInterface` with `DownloadManager` in [the function signature](https://github.com/drupol/composer/blob/7528c86e58884f5676d5851d52cd94c67372d609/src/Composer/Util/SyncHelper.php#L32).

If we do that, we would benefit of the very useful logic in `DownloadManager::download` where the source of the package is automatically detected.

The only drawback here is that instead of relying on an interface, we rely on a concrete implementation... but nothing prevent us to create an interface for the `DownloadManager`.

WDYT ?

Seldaek:
I don't need an interface there, but this is also a breaking change. If we can accept both for BC then I guess why not.

stof:
The BC layer is easy: instead of replacing the interface with DownloadManager, change the argument to accept `DownloadManager|DownloaderInterface`.

drupol:
Yes, but this is not supported in PHP 7.2, this is why I did it in a different way: https://github.com/composer/composer/pull/11496

stof:
Well, you can put the union type in phpdoc (with a check inside the method if you want to keep a TypeError on invalid type) instead of having 2 methods in the API.

PR #11486, opened 2023-05-30
Handle better nullable file parameter

### LINKED ISSUE #11483: Handle better nullable file reference in \Composer\Json\JsonFile::validateSyntax()
When `\Composer\Json\JsonFile::validateSyntax()` called without a second parameter (e.g, via `\Composer\Json\JsonFile::parseJson('foo')`) it generates a cryptic exception message like

> "" does not contain valid JSON
Parse error on line 1:
warning: log for 'HE
^

Notice how the message start `"" does not contain valid JSON`.

https://github.com/composer/composer/blob/2.5.7/src/Composer/Json/JsonFile.php#L339-L347

mxr576:
Closes #11483


Seldaek:
Thanks

PR #11493, opened 2023-06-05
Force CreateProjectCommand to use the installed composer.json

aschempp:
In my local environment ([Contao Manager](https://github.com/contao/contao-manager)), I always have the `COMPOSER` environment variable set to the root of my project directory. When using the `create-project` command with an new, empty directory target (because my project dir is not empty), Composer correctly downloads the files into the new directory, but then fails when calling [`Factory::create`](https://github.com/composer/composer/blob/main/src/Composer/Command/CreateProjectCommand.php#L205) because that uses the `COMPOSER` environment variable instead of the newly downloaded files.

I'm not sure what implications this could have on a running command, but I would assume that the variable is gone as soon as the `create-project` command is finished?

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -469,6 +469,7 @@ EOT
         $io->writeError('<info>Created project in ' . $directory . '</info>');
         chdir($directory);
 
+        Platform::putEnv('COMPOSER', $directory.'/composer.json');
         Platform::putEnv('COMPOSER_ROOT_VERSION', $package->getPrettyVersion());
 
         // once the root project is fully initialized, we do not need to wipe everything on user abort anymore even if it happens during deps install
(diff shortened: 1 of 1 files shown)


src/Composer/Command/CreateProjectCommand.php
```diff
@@ -469,6 +469,7 @@ protected function installRootPackage(IOInterface $io, Config $config, string $p
         $io->writeError('<info>Created project in ' . $directory . '</info>');
         chdir($directory);
 
+        Platform::putEnv('COMPOSER', $directory.'/composer.json');
```
Seldaek, line 472: Seems reasonable to me but I'd wrap this in a `if (file_exists($directory.'/composer.json'))` to make sure nobody relies on setting COMPOSER env to load some other file when using create-project. Just making sure we remain as BC as possible..

Seldaek: Nope, I mean if the file exists then we override the COMPOSER env to make sure that file is used. If the file is not there then leave COMPOSER alone as it might by some chance point to the correct file (which we cannot know)

aschempp: changed in 2d79c0df6c03c391bf3a867282d49274dc697fd8

Seldaek:
Thanks :)

PR #11496, opened 2023-06-06
refactor: update `SyncHelper::downloadAndInstallPackageSync`

drupol:
This PR:

- [x] is a proposal, originating from [a custom plugin](https://github.com/drupol/composer-local-repo-plugin/) I made
- [x] widen a single parameter of `SyncHelper::downloadAndInstallPackageSync`

This simple change would prevent duplicating code related to the download logic in `DownloadManager::download` where the source of the package is automatically detected (_this would also prevent me from copying the complete function [in my plugin](https://github.com/drupol/composer-local-repo-plugin/blob/main/src/Service/RepoBuilder.php#L24) and modify its signature only_).

WDYT ?

### DIFF
--- a/src/Composer/Plugin/PluginInterface.php
+++ b/src/Composer/Plugin/PluginInterface.php
@@ -32,7 +32,7 @@ interface PluginInterface
      *
      * @var string
      */
-    public const PLUGIN_API_VERSION = '2.3.0';
+    public const PLUGIN_API_VERSION = '2.6.0';
 
     /**
      * Apply plugin modifications to Composer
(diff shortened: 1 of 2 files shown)


Seldaek:
I'm wondering if we wouldn't be better off just widening the argument instead of adding one method just for that. You can do an `assert($dl instanceof DownloaderInterface || $dl instanceof DownloadManager);` to enforce it and then have it in the phpdoc. IMO that'd be enough and reduce duplication here.

drupol:
Good idea, proceeding...

drupol:
> Could we make `\Composer\Downloader\DownloadManager` implement `\Composer\Downloader\DownloaderInterface`? Or create another interface that only requires the methods used here (`download`, `prepare`, `update`, `install`, `cleanup`)?

I also think about that... I don't know yet if this is a good idea or not. In terms of design pattern, is this a good idea to do that?

Seldaek:
It's fine as is IMO, this is such a narrow use case not worth adding an interface and complicating things.

PR #11497, opened 2023-06-07
Add IOMock and get rid of all withConsecutive calls in tests

Seldaek:
Refs https://github.com/sebastianbergmann/phpunit/issues/4564 - this gets us ready for an upgrade eventually, and adding an IOMock is anyway a good thing to be able to assert input/output more easily there.

### DIFF
--- a/tests/Composer/Test/Mock/IOMock.php
+++ b/tests/Composer/Test/Mock/IOMock.php
@@ -1,0 +1,196 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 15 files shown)


tests/Composer/Test/Mock/IOMock.php
```diff
@@ -0,0 +1,196 @@
...
+     * @param list<array{text: string, verbosity?: IOInterface::*, regex?: true}|array{ask: string, reply: string}|array{auth: array{string, string, string|null}}> $expectations
+     * @param bool                                                                                                   $strict         set to true if you want to provide *all* expected messages, and not just a subset you are interested in testing
+     */
+    public function expects(array $expectations, bool $strict = false): void
+    {
+        $this->expectations = $expectations;
+        $inputs = [];
+        foreach ($expectations as $expect) {
+            if (isset($expect['ask'], $expect['reply'])) {
+                if (!is_string($expect['reply'])) {
```
stof, line 69: this won't be triggered for `null` values as the surrounding `isset` will be false for them.

Seldaek: Good point, fixed in 29134d5a1

PR #11500, opened 2023-06-08
Update type declarations on ClassLoader

### LINKED ISSUE #11482: Wrong return type on Classloader::getPrefixes
I haven't PRed because I'm not sure what the type is in all cases but `string[]` appears to be incorrect, we get a multidimensional array back

https://github.com/composer/composer/blob/52f6f74b7c342f7b90be9c1a87e183092e8ab452/src/Composer/Autoload/ClassLoader.php#L116

Seldaek:
Fixes #11482

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -45,35 +45,34 @@ class ClassLoader
     /** @var \Closure(string):void */
     private static $includeFile;
 
-    /** @var ?string */
+    /** @var string|null */
     private $vendorDir;
 
     // PSR-4
     /**
-     * @var array[]
-     * @psalm-var array<string, array<string, int>>
+     * @var array<string, array<string, int>>
      */
     private $prefixLengthsPsr4 = array();
     /**
-     * @var array[]
-     * @psalm-var array<string, array<int, string>>
+     * @var array<string, array<int, string>>
      */
     private $prefixDirsPsr4 = array();
     /**
-     * @var array[]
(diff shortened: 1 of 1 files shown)


src/Composer/Autoload/ClassLoader.php
```diff
@@ -45,66 +45,63 @@ class ClassLoader
...
 
     // PSR-0
     /**
-     * @var array[]
-     * @psalm-var array<string, array<string, string[]>>
+     * List of PSR-0 prefixes
+     *
+     * Structured as array('F (first letter)' => array('Foo\Bar (full prefix)' => array('path', 'path2')))
+     *
+     * @var array<string, array<string, string[]>>
```
stof, line 71: should this use `array<string, array<string, list<string>>>` to be even more precise ?

src/Composer/Autoload/ClassLoader.php
```diff
@@ -125,44 +122,39 @@ public function getPrefixes()
     }
 
     /**
-     * @return array[]
-     * @psalm-return array<string, array<int, string>>
+     * @return array<string, array<int, string>>
```
stof, line 125: `array<string, list<string>>` would be more precise

Seldaek: Yup, thanks 45368a5bd

Seldaek: Sorry 3f385d466fef9bdd355a6ec01ec7c1875bd95f92

PR #11501, opened 2023-06-09
Add access token authentication for private Bitbucket repositories

BernhardK91:
This PR adds access token authentication for Bitbucket repositories.

## Background

Composer already supports Bitbucket authentication with OAuth. Bitbucket's OAuth configuration has some down sites. Because the OAuth token is configured at workspace levels, it applies to all repositories in that workspace.

To have repository-based tokens, Bitbucket provides so-called "Access tokens", that can be found in the repository settings:

image

As a workaround you could set access token as OAuth token for the user "x-token-auth", but then only one is possible in the whole project.

## Configuration

This PR adds a new parameter group to auth json. A Bitbucket repository that uses access token authentication can be configured as followed in the `composer.json` file:

```
[...]
    "repositories": [
        {
            "type": "bitbucket",
            "url": "https://bitbucket.org/{workspace-name}/{repository-name}.git",
        }
    ],
[...]
```

And the access token will be stored in the `auth.json` file by defining for which repository it applies:

```
{
    "bitbucket-token": {
        "{workspace-name}/{repository-name}" : "{repository access token}"
    }
}
```

## Implementation

A few comments on the implementation. The authentication on the API with the bearer token is straightforward. Unfortunately, this does not affect the git authentication and that had to be updated to use the token in the git command as described in the [Bitbucket documentation](https://support.atlassian.com/bitbucket-cloud/docs/using-access-tokens#Include-the-Repository-Access-Token-in-the-URL). 

The authentication for the zip file that composer wants to download (dist) is not supported with access token. To avoid ugly warnings for the users, I added a check that returns null for the dist data. This way, composer will not try to download the dist and directly falls back to the git clone approach.

## Conclusion

This is my first PR for composer. So please let me know if I missed anything or if you have any concerns about this change. Thank you!

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -198,7 +198,7 @@ Read more [here](articles/authentication-for-private-packages.md#bitbucket-oauth
 
 ## bitbucket-token
 
-A list of repository names with the particular (access token)[https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/].
+A list of repository names with the particular [access token](https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/).
 For example using `{"composer/composer": "example-access-token"}`.
 Read more [here](articles/authentication-for-private-packages.md#bitbucket-token).
 
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -196,6 +196,12 @@ A list of domain names and consumers. For example using `{"bitbucket.org":
 {"consumer-key": "myKey", "consumer-secret": "mySecret"}}`.
 Read more [here](articles/authentication-for-private-packages.md#bitbucket-oauth).
 
+## bitbucket-token
+
+A list of repository names with the particular (access token)[https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/].
+For example using `{"composer/composer": "example-access-token"}`.
+Read more [here](articles/authentication-for-private-packages.md#bitbucket-token).
+
 ## cafile
 
 Location of Certificate Authority file on local filesystem. In PHP 5.6+ you
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -199,7 +199,7 @@ Read more [here](articles/authentication-for-private-packages.md#bitbucket-oauth
 ## bitbucket-token
 
 A list of repository names with the particular [access token](https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/).
-For example using `{"composer/composer": "example-access-token"}`.
+For example using `{"workspace-name/repository-name": "repository-access-token"}`.
 Read more [here](articles/authentication-for-private-packages.md#bitbucket-token).
 
 ## cafile
--- a/doc/articles/authentication-for-private-packages.md
+++ b/doc/articles/authentication-for-private-packages.md
@@ -364,7 +364,7 @@ php composer.phar config [--global] --editor --auth
 
 ## bitbucket-token
 
-The BitBucket driver is also able to access your private repositories via the BitBucket REST APIs, using (Bitbucket Repository Access Tokens)[https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/].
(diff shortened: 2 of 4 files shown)


cafferata:
Something I could think of: Completing the documentation with this feature. See`bitbucket-oauth`:
1. https://getcomposer.org/doc/06-config.md#bitbucket-oauth
1. https://getcomposer.org/doc/articles/authentication-for-private-packages.md#bitbucket-oauth

BernhardK91:
@cafferata Thank you for your response and mentioning updating the documentation. While going through that, I also figured out a better way of adding the access token that is more in line with the other authentication options. 
In addition, I found a couple of more issues of the original PR and fixed them as well. 

Looking forward to more feedback. :-)

doc/articles/authentication-for-private-packages.md
```diff
@@ -24,6 +24,7 @@ for credentials and save them (or a token if Composer is able to retrieve one).
 |[gitlab-token](#gitlab-token)|yes|
 |[github-oauth](#github-oauth)|yes|
 |[bitbucket-oauth](#bitbucket-oauth)|yes|
+|[bitbucket-oauth](#bitbucket-oauth)|no|
```
cafferata, line 27: @BernhardK91 I guess you mean `bitbucket-token`?

```suggestion
|[bitbucket-token](#bitbucket-token)|no|
```

cafferata (changes requested):

cafferata:
Your welcome! I'll check it later 👍

doc/articles/authentication-for-private-packages.md
BernhardK91: Ouh yeah. I missed that. Thank you!

doc/articles/authentication-for-private-packages.md
```diff
@@ -360,3 +361,25 @@ php composer.phar config [--global] --editor --auth
     }
 }
 ```
+
+## bitbucket-token
+
+The BitBucket driver is also able to access your private repositories via the BitBucket REST APIs, using (Bitbucket Repository Access Tokens)[https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/].
```
cafferata, line 367: (Multiple) small typo in de `Bitbucket` name.

```suggestion
The Bitbucket driver is also able to access your private repositories via the Bitbucket REST APIs, using (Bitbucket Repository Access Tokens)[https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/].
```

cafferata, line 367: Here too I suspect that the link is desired to be clickable. For this, the Markdown syntax must be reversed.

```suggestion
The BitBucket driver is also able to access your private repositories via the BitBucket REST APIs, using [Bitbucket Repository Access Tokens](https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/).
```

doc/06-config.md
```diff
@@ -196,6 +196,12 @@ A list of domain names and consumers. For example using `{"bitbucket.org":
 {"consumer-key": "myKey", "consumer-secret": "mySecret"}}`.
 Read more [here](articles/authentication-for-private-packages.md#bitbucket-oauth).
 
+## bitbucket-token
+
+A list of repository names with the particular (access token)[https://support.atlassian.com/bitbucket-cloud/docs/create-a-repository-access-token/].
+For example using `{"composer/composer": "example-access-token"}`.
```
cafferata, line 202: Would it be good to use [the same](https://github.com/composer/composer/pull/11501/files#diff-2efd65948b461045cd73c395a83e71eeca38405bd2a65803933dc69bb5b6e3f1R382) example here?

""

```suggestion
For example using `{"workspace-name/repository-name": "repository-access-token"}`.
```

cafferata (changes requested):
I think after reading the documentation it's a clear story. Just minor feedback points. I leave the technical code review to @Seldaek.

BernhardK91:
@cafferata Thanks again for your suggestions. I agree with all and committed as suggested.

Seldaek:
Per @glaubinix's review this is not really workable without much more invasive changes so given there are already existing solutions via SSH keys + org wide oauth I would probably tend to say we rather not add this.

Issue #6929, opened 2017-12-21
RFC: Automatically resolve conflicts in composer.lock


hkdobrev:
When merging/rebasing branches it's not that rare that you may encounter merge conflicts in `composer.lock`. Most of the time the conflict would be in the hash or in a timestamp or in some cases you'd need to choose both blocks on the same lines.

However, it's quite rare you'd actually need to manually resolve something in `composer.lock` in some custom way.

The usual workflow:

1. You have a conflict in `composer.lock`
1. You run your `git-mergetool` and chose one of the hashes - doesn't matter which one.
1. Run `composer update --lock`
1. Run `git add composer.lock`
1. Run `git rebase --continue` or `git commit` (if you're merging).

How about automatically fix the merge conflicts when running `composer install` or `composer update --lock` if possible?

This was already implemented in Yarn:
- Pull request: https://github.com/yarnpkg/yarn/pull/3544
- Announcement: https://code.facebook.com/posts/274518539716230
- Screenshot: https://user-images.githubusercontent.com/13352/27958995-8104ab7e-631d-11e7-8050-a36cbb48e75d.png

What do you think?

adanilowski:
Maybe not very clean but - separate folder like '.composer.lock' having files separate per library or just remove content-hash. In bigger projects with large teams, this becomes a real pain.

hkdobrev:
@adanilowski I think this is quite a drastic change compared to how the lock file currently works and is a breaking change. My suggestion is around the current implementation and it is similar to how other package managers work. I really appreciate having all dependencies locked into a single file and the hash of dependencies for the warnings it produces.

Seldaek:
The content-hash serves a real purpose and removing it is not a solution, it might resolve some merge conflicts but would introduce other issues.

Adding a merge functionality that checks the two lock files and sees if they are compatible and still match the current requirements though sounds like something that's feasible and definitely should be looked.

stof:
This will be harder than in Yarn though, because our composer.lock file is a JSON file, but a     merge conflict makes it unparseable, and so would require to build a special parser for conflicting files (not impossible though).

hkdobrev:
I suppose we need to detect just the following:
- is there a merge conflict
- is the merge conflict just in the hash

If it is, we just need to apply a merge strategy to choose one of the sides without parsing the JSON. Then run `composer update --lock` and add the updated file to the index. (I'm assuming Git in all this)

ostrolucky:
[...]

PR #11517, opened 2023-06-19
[feature] Automatically recover from trivial merge conflicts in `composer.lock`

fredden:
Related to: https://github.com/composer/composer/issues/6929

This is a very naive approach to try to gracefully handle trivial merge conflicts in `composer.lock` files. This does nothing for conflicts in `composer.json` files, and does not handle "complex" merge conflicts where more than just the `content-hash` has conflicted.

In the scenario where only the `content-hash` has conflicted, and when using the `git` version-control system\*, the change in this pull request is enough to resolve this situation gracefully. A user running `composer install` in this case will get the expected packages installed (most of the time), and a suitable notice saying that the content-hash is not up to date. A user running `composer update --lock` in this case will have a "fixed" lock file produced (most of the time).

\* I've not tested other version control systems. This approach could be adapted to handle other version control systems's merge conflict markers.

This pull request includes tests and an update to the documentation.

## Workflow before this pull request

1. `git merge ...` results in a conflict in `composer.lock`.
1. `git diff composer.lock` shows that only the content-hash lines are conflicting.
1. Note: running `composer update --lock` here throws an exception because the `composer.lock` file no longer contains valid JSON.
1. Open `composer.lock` in an editor.
   1. Delete lines 6, 7, 8, 10. This makes the file valid JSON again.
   1. (Optional) Replace `content-hash` value with an invalid hash (eg `x`). This isn't strictly necessary, but makes it clear that Composer did something in the following steps.
   1. Save the file, close editor.
1. Note: `git diff composer.lock` now shows no merge conflict markers are present.
1. Run `composer update --lock` to generate a new `content-hash` and validate that the packages are installable together.
1. `git diff composer.lock` shows that the content-hash line has been updated to be different from both sides of the merge.
1. `git add composer.lock`
1. `git commit -v` as normal, or another `git` command to continue the process (eg, `git rebase --continue`)

## Workflow after this pull request

1. `git merge ...` results in a conflict in `composer.lock`.
1. `git diff composer.lock` shows that only the content-hash lines are conflicting.
1. ~Note: running `composer update --lock` here throws an exception because the `composer.lock` file no longer contains valid JSON.~
1. ~Open `composer.lock` in an editor.~
   1. ~Delete lines 6, 7, 8, 10. This makes the file valid JSON again.~
   1. ~(Optional) Replace `content-hash` value with an invalid hash (eg `x`). This isn't strictly necessary, but makes it clear that Composer did something in the following steps.~
   1. ~Save the file, close editor.~
1. ~Note: `git diff composer.lock` now shows no merge conflict markers are present.~
1. Run `composer update --lock` to generate a new `content-hash` and validate that the packages are installable together.
1. `git diff composer.lock` shows that the content-hash line has been updated to be different from both sides of the merge.
1. `git add composer.lock`
1. `git commit -v` as normal, or another `git` command to continue the process (eg, `git rebase --continue`)

### DIFF
--- a/src/Composer/Json/JsonFile.php
+++ b/src/Composer/Json/JsonFile.php
@@ -343,12 +343,12 @@ class JsonFile
         }
         $data = json_decode($json, true);
         if (null === $data && JSON_ERROR_NONE !== json_last_error()) {
-            $lines = explode(PHP_EOL, $json);
+            $lines = explode("\n", $json);
             if (isset($lines[9]) && strpos($lines[9], '"content-hash": "') !== false && strpos($lines[7], '"content-hash": "') !== false) {
                 // We may have found a git merge conflict in composer.lock. Remove the offending lines and try again.
                 unset($lines[6], $lines[7], $lines[9], $lines[10]);
                 $lines[8] = '    "content-hash": "VCS merge conflict detected. Please run `composer update --lock`.",';
-                return self::parseJson(implode(PHP_EOL, $lines), $file);
+                return self::parseJson(implode("\n", $lines), $file);
             }
 
             self::validateSyntax($json, $file);
--- a/src/Composer/Json/JsonFile.php
+++ b/src/Composer/Json/JsonFile.php
@@ -343,6 +343,14 @@ class JsonFile
         }
         $data = json_decode($json, true);
         if (null === $data && JSON_ERROR_NONE !== json_last_error()) {
+            $lines = explode(PHP_EOL, $json);
+            if (isset($lines[9]) && strpos($lines[9], '"content-hash": "') !== false && strpos($lines[7], '"content-hash": "') !== false) {
+                // We may have found a git merge conflict in composer.lock. Remove the offending lines and try again.
+                unset($lines[6], $lines[7], $lines[9], $lines[10]);
+                $lines[8] = '    "content-hash": "VCS merge conflict detected. Please run `composer update --lock`.",';
+                return self::parseJson(implode(PHP_EOL, $lines), $file);
+            }
+
             self::validateSyntax($json, $file);
         }
 
--- a/src/Composer/Json/JsonFile.php
+++ b/src/Composer/Json/JsonFile.php
@@ -343,12 +343,20 @@ class JsonFile
         }
         $data = json_decode($json, true);
         if (null === $data && JSON_ERROR_NONE !== json_last_error()) {
-            $lines = explode("\n", $json);
-            if (isset($lines[9]) && strpos($lines[9], '"content-hash": "') !== false && strpos($lines[7], '"content-hash": "') !== false) {
-                // We may have found a git merge conflict in composer.lock. Remove the offending lines and try again.
-                unset($lines[6], $lines[7], $lines[9], $lines[10]);
-                $lines[8] = '    "content-hash": "VCS merge conflict detected. Please run `composer update --lock`.",';
-                return self::parseJson(implode("\n", $lines), $file);
+            if ($file !== null && str_ends_with($file, '.lock') && str_contains($json, '"content-hash"')) {
+                $replaced = Preg::replace(
+                    '{\r?\n<<<<<<< [^\r\n]+\r?\n\s+"content-hash":.*?\r?\n(?:\|{7} [^\r\n]+\r?\n\s+"content-hash":.*?\r?\n)?=======\r?\n\s+"content-hash":.*?\r?\n>>>>>>> [^\r\n]+(\r?\n)}',
+                    '    "content-hash": "VCS merge conflict detected. Please run `composer update --lock`.",$1',
+                    $json,
+                    1,
+                    $count
+                );
+                if ($count === 1) {
+                    $data = json_decode($replaced, true);
+                    if ($data !== null) {
+                        return $data;
+                    }
+                }
             }
 
             self::validateSyntax($json, $file);
--- a/src/Composer/Json/JsonFile.php
+++ b/src/Composer/Json/JsonFile.php
@@ -343,9 +343,10 @@ class JsonFile
         }
         $data = json_decode($json, true);
         if (null === $data && JSON_ERROR_NONE !== json_last_error()) {
+            // attempt resolving simple conflicts in lock files so that one can run `composer update --lock` and get a valid lock file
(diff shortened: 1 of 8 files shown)


src/Composer/Json/JsonFile.php
```diff
@@ -318,6 +318,14 @@ public static function parseJson(?string $json, ?string $file = null)
         }
         $data = json_decode($json, true);
         if (null === $data && JSON_ERROR_NONE !== json_last_error()) {
+            $lines = explode("\n", $json);
+            if (isset($lines[9]) && strpos($lines[9], '"content-hash": "') !== false && strpos($lines[7], '"content-hash": "') !== false) {
+                // We may have found a git merge conflict in composer.lock. Remove the offending lines and try again.
+                unset($lines[6], $lines[7], $lines[9], $lines[10]);
+                $lines[8] = '    "content-hash": "VCS merge conflict detected. Please run `composer update --lock`.",';
```
Seldaek, line 325: I'm not sure here.. is this better than just picking one? IMO if anything we should rather parse the composer.json to get the current hash and make sure we update the hash in composer.lock, but that will not guarantee that the merge was done correctly in terms of dependencies, and you may still be in an inconsistent state.

Plus you have the problem that git will still mark the file conflicted until you add it again, so all in all I am not sure if this helps more than it causes/hides potential problems.

I'd recommend using https://github.com/balbuf/composer-git-merge-driver if these conflicts are a problem for you.

fredden: > I'm not sure here.. is this better than just picking one? IMO if anything we should rather parse the composer.json to get the current hash and make sure we update the hash in composer.lock, but that will not guarantee that the merge was done correctly in terms of dependencies, and you may still be in an inconsistent state.

This string is intentionally not a valid hash. I don't think it will ever be written to disk, but is instead only used in-memory within Composer. Picking either of the two hashes would give the same result (ie, the hash isn't valid), but be somewhat misleading. When running `composer install` any invalid hash will trigger the "outdated" warning. When running `composer update`, the correct hash will be calculated and written out.

> Plus you have the problem that git will still mark the file conflicted until you add it again, so all in all I am not sure if this helps more than it causes/hides potential problems.

Yes, git will pause with conflicts and require these to be marked as resolved. Nothing's changing in git here. The changes here are an attempt to make the "resolve conflict" process smoother, not bypass it.

> I'd recommend using https://github.com/balbuf/composer-git-merge-driver if these conflicts are a problem for you.

Using a git merge driver requires specific set-up, which isn't always compatible with how/where composer runs. Yes, that merge driver looks very useful, but doesn't work when the system PHP isn't compatible with the application being manipulated.

For example, many of the projects that I work on require specific (read: older) versions of PHP with various extensions available (to match production). I use docker to get the correct version of PHP etc, and run Composer in those containers. Running `composer update --lock` on my host system results in a failure due to the mismatch of PHP versions. I run `git` on my host system.
Using a git merge driver would require me to configure a different driver for each project so that it could run in the correct environment. This means that the set-up for each would need to be in each individual project. I'm not the only person working on these projects, so anyone else trying to do the same would have conflicting configuration. Anyone without the git merge driver installed may not be able to work on the project trivially. Some of this could be mitigated by using git within the container/environment where PHP works, but this then prevents users from using their much-loved graphical tools for working with git.

I've updated the pull request description to demonstrate the steps in my workflow which this pull request would remove.

Seldaek: Ok thanks, I guess I'll have to try this out to see a bit clearly more how it behaves.

nguyenpham00:
#MarkDown

Seldaek:
Alright, rebased and added support for extended conflict markers with a `||||||| parent of ad845c646f (Commit message here)` line in them, and overall made this a bit more resilient by not using exact line numbers.

cc @TimWolla 

I'd say it's good to merge now but will leave this for a bit in case anyone wants to review.

src/Composer/Json/JsonFile.php
```diff
@@ -343,6 +343,22 @@ public static function parseJson(?string $json, ?string $file = null)
         }
         $data = json_decode($json, true);
         if (null === $data && JSON_ERROR_NONE !== json_last_error()) {
+            if ($file !== null && str_ends_with($file, '.lock') && str_contains($json, '"content-hash"')) {
+                $replaced = Preg::replace(
+                    '{\r?\n<<<<<<< [^\r\n]+\r?\n\s+"content-hash":.*?\r?\n(?:\|{7} [^\r\n]+\r?\n\s+"content-hash":.*?\r?\n)?=======\r?\n\s+"content-hash":.*?\r?\n>>>>>>> [^\r\n]+(\r?\n)}',
```
TimWolla: I believe the `.*?` after the content-hash might match newlines. I would make it a little more specific. Something like this perhaps: `"content-hash":\s*"[0-9a-f]+"\s*\r?\n`

Seldaek: It shouldn't given it's not greedy, but yes could be a bit more strict for safety, will do

Seldaek:
Thanks @fredden and sorry this took so long 😅

staabm:
Awesome. 👏

naderman:
Guess no way for me to retroactively mark the approval on the review request here that i sent over slack, so just removing me so this isn't still awaiting my review :facepalm:

alexander-schranz:
Awesome feature, thx @fredden 👍

hostep:
A colleague of mine just accidentally put a unresolved merge conflict in `composer.lock` around the `content-hash` line in our git repo and ran a deploy, which for the most part ran surprisingly fine, but at the end when running our application's binary, we saw a weird crash that was a bit hard to debug, but it turned out to be a syntax error in the file `vendor/autoload.php` where the very last line of that file looked like this:

```php
return ComposerAutoloaderInitVCS merge conflict detected. Please run `composer update --lock`.::getLoader();
```

Which isn't really valid php code.

I'm just putting this comment here on this PR, since it's the only recent change in composer sourcecode that I remembered that sounds related, I haven't verified if this change is the one causing it or not, I don't have too much time to dig in deeper at the moment, but if asked I can try to spend a bit longer investigating later this week and making a proper bug report.

stof:
@hostep please open an issue instead of commenting on a merged PR. It is much more likely to be noticed and followed

PR #11521, opened 2023-06-23
Fix problem output of exact constraints to state more clearly that only those versions are matched

### LINKED ISSUE #11479: Clarify that exact constraints are exact in problem output
`{"require":{"php":"8.2"}}` for ex will output:

```

  Problem 1
    - Root composer.json requires php 8.2 but your php version (8.2.3) does not satisfy that requirement.
```

The 8.2 bit there isn't very clear, maybe if we detect a short pinned version like that we could add warning/suggestion, and generally add a `=` in front of all pinned versions like `requires php =8.2` for clarity? Assuming this would bring clarity 😄

Seldaek:
Fixes #11479

### DIFF
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -557,6 +557,19 @@ class Problem
      */
     protected static function constraintToText(?ConstraintInterface $constraint = null): string
     {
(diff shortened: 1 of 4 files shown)


naderman (approved):
Looks good, technically `=1.0.0` will match `1.0` too, but that's probably not something you see in reality much.

Seldaek:
Yeah I think it's really only been confusing people in the other direction where they assume 1.0 == 1.0.*

PR #11522, opened 2023-06-24
Test status command

theoboldalex:
This gets us started on testing `StatusCommand`. I have also been trying to test some of the other paths through this command but have as yet been unsuccessful when installing a package and actually trying to modify its source to help us hit some of the uncovered paths. Any insider knowledge on how I could achieve this would be appreciated.


Seldaek:
I think you'd need to embed a zip archive in the test suite to have a dummy package installed, instead of adding it as a metapackage as this will not do any on-disk install. Then it might be possible to modify files and get output. Or alternatively use a git repo and install a package from source that should also work, you can use composer/class-map-generator or similar small package to get a quick clone and do a source install.

theoboldalex:
Just an update on this. I haven't forgotten about this PR, just having a little trouble installing a package from source within my test. Will update the PR as soon as I figure this out.

Seldaek:
Ok let me know if I can help, feel free to commit some partially working stuff and I can take a look if you're stuck.

theoboldalex:
I have updated the PR with some debug and comments. I think it is likely that I have misunderstood something fundamental about how packages are installed from source. Any guidance you could provide would be hugely appreciated.

tests/Composer/Test/Command/StatusCommandTest.php
```diff
@@ -0,0 +1,68 @@
...
+        $package->setSourceUrl('https://github.com/composer/class-map-generator.git');
+        $package->setSourceReference('953cc4ea32e0c31f2185549c7d216d7921f03da9');
+
+        $this->createComposerLock([$package], []);
+        $this->createInstalledJson([$package], []);
+
+        // installed.json LGTM
+        var_dump(file_get_contents(getcwd() . '/vendor/composer/installed.json'));
+        // Expected some source here instead of an empty dir. Do I actually need to run a `composer install` here?
+        var_dump(scandir(getcwd() . '/vendor/composer/class-map-generator/'));
```
Seldaek, line 53: Right here you would need to skip createInstalledJson as that make it look like the package is installed but it's really just faking things for speed of testing. Instead execute a proper composer install (I think you could do that via `$appTester->run(['command' => 'install']);`) so the package gets installed from source.

theoboldalex: Yes, that was it thank you! I had tried running an install through the appTester but was calling install _after_ `createInstalledJson`. Tests passing now. I'll spend some time covering off a couple more paths through the StatusCommand and post for review again once I'm done. Thank you 🙏

Seldaek:
Cool, looking good now, thanks!

PR #11523, opened 2023-06-24
Fix outdated package name regex pattern in documentation

mteu:
The regular expression used to validate the package name in the `composer.json` as specified in https://getcomposer.org/schema.json differs from the example given in the documentation.

This PR seeks to update the documentation against the current pattern:

## schema.json
```json
"name": {
    "type": "string",
    "description": "Package name, including 'vendor-name/' prefix.",
    "pattern": "^[a-z0-9]([_.-]?[a-z0-9]+)*\/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*$"
},
```

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -35,7 +35,7 @@ separated by `/`. Examples:
 * igorw/event-source
 
 The name must be lowercase and consist of words separated by `-`, `.` or `_`.
-The complete name should match `^[a-z0-9]([_.-]?[a-z0-9]+)*\/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*$`.
+The complete name should match `^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*$`.
 
 The `name` property is required for published packages (libraries).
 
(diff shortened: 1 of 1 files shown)


doc/04-schema.md
```diff
@@ -35,7 +35,7 @@ separated by `/`. Examples:
 * igorw/event-source
 
 The name must be lowercase and consist of words separated by `-`, `.` or `_`.
-The complete name should match `^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$`.
+The complete name should match `^[a-z0-9]([_.-]?[a-z0-9]+)*\/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*$`.
```
stof, line 38: the `\` should be removed here. As this regex is given without delimiters, there is no reason to escape the `/`

mteu: Thanks for your feedback. You're of course right. I've unescaped the delimiter in [`59c82f3`](https://github.com/composer/composer/pull/11523/commits/59c82f3c00fb1730b99b4294a5573b6db5353a6b).

Seldaek:
Thanks

PR #11524, opened 2023-06-24
Adds .vscode dir to gitignore file

theoboldalex:
Just adding .vscode dir to the gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ phpunit.xml
 .vagrant
 Vagrantfile
 .idea
+.vscode
 .php-cs-fixer.cache


fredden:
@theoboldalex perhaps this would be better suited in your personal global git ignore file?
https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

However, `.idea/` is already listed specifically here.

theoboldalex:
@fredden I somewhat agree although given vscode's prevalence it kind of makes sense to have this at the repo level rather than global level (example being for some repos it makes sense to include the .vscode dir where there is some funky custom debug launch file etc).

I'm not a vscode user myself but have seen this dir getting committed so frequently and requiring an extra git step to remove, that I think it is worth having in the .gitignore.

I guess it is down to the reviewer/maintainer whether this gets merged but from my perspective this change doesn't impact users of other editor/IDEs but provides benefit for a large proportion of potential contributors to the project.

Seldaek:
I also think people should have this in their global gitignore but that opinion unfortunately does not match reality so if it helps someone we might as well have it.

Issue #11519, opened 2023-06-22
Allow relative local directory for repositories of type composer


helhum:
For a custom mono repo approach, I'm currently generating `packages.json` file in a local directory.
I would love to be able to specify the path to this repository relative to the root composer.json similar to repositories of type path:

```json
{
    "repositories": [
        {
            "type": "composer",
            "url": "./path/to/local-dir/"
        }
    ]
}
```

Composer repositories though, only allow URLs. I could specify a `file://` URL, but then the specified path needs to be an absolute file path:

```json
{
    "repositories": [
        {
            "type": "composer",
            "url": "file:///absolute/path/to/local-dir/"
        }
    ]
}
```

Is there anything conceptually ore security wise that is a blocker to allow local relative file system paths for repositories of type Composer?

If not, I'm happy to create a PR to allow it.

Seldaek:
Ah weird I thought this supported relative paths, but indeed it seems like it doesn't. I think it should be fine to allow it..

helhum:
Cool, thanks. Then I'll try to provide a PR for that

helhum:
Added support for local file paths in a very simple way, which I assumed has the lowest impact. Let me know in the PR if this is fine and whether I should add some docs or tests for that.

PR #11526, opened 2023-06-26
Feature: Allow local directory paths in repository of type composer

helhum:
Fixes: #11519


dzuelke:
FWIW, this already works if you use `file://` explicitly.

helhum:
> FWIW, this already works if you use `file://` explicitly.

Yes, but not with relative paths. And I can't specify absolute file paths in a composer.json file

dzuelke:
> > FWIW, this already works if you use `file://` explicitly.
> 
> Yes, but not with relative paths. And I can't specify absolute file paths in a composer.json file

Sorry, was still editing that comment ;)

Seldaek:
Thanks

helhum:
> Thanks

Thank you!

Issue #11046, opened 2022-09-02
2.4 regression: composer show --platform requires a composer.json


dzuelke:
```
~ % cd $(mktemp -d)
tmp.cwO1AqLu % php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"

tmp.cwO1AqLu % php composer-setup.php --filename=composer-2.3.10 --version=2.3.10 
All settings correct for using Composer
Downloading...
Composer (version 2.3.10) successfully installed to: …/tmp.cwO1AqLu/composer-2.3.10
Use it: php composer-2.3.10

tmp.cwO1AqLu % php composer-setup.php --filename=composer-2.4.1 --version=2.4.1
All settings correct for using Composer
Downloading...
Composer (version 2.4.1) successfully installed to: …/tmp.cwO1AqLu/composer-2.4.1
Use it: php composer-2.4.1
```

Older versions let me list platform packages; this is useful in all kinds of scripts etc ("are the right extensions there", for instance):

```
tmp.cwO1AqLu % ./composer-2.3.10 show --platform
composer               2.3.10   Composer package
…
php                    8.1.5    The PHP interpreter
php-64bit              8.1.5    The PHP interpreter, 64bit
php-ipv6               8.1.5    The PHP interpreter, with IPv6 support
```

Since 2.4., this errors:

```
tmp.cwO1AqLu % ./composer-2.4.1 show --platform
Composer could not find a composer.json file in /private/var/folders/n9/nlkmp1dn0x9616hsy3bb2z4m0000gq/T/tmp.cwO1AqLu
To initialize a project, please create a composer.json file. See https://getcomposer.org/basic-usage
```

PR #11533, opened 2023-07-04
Fix 'composer show --platform <package>' erroring without composer.json

dzuelke:
Sort of related to #11046 (although this is not a regression, but didn't work before, either)

This now skips output of the `path` field entirely (it previously was always empty)... fine IMO, since this could happen previously (e.g. if `realpath()` failed in case of JSON output), and the path has always been `null` anyway for platform packages.

Also think this should maybe get backported to 2.2? Would you like me to rebase on top of that branch instead of `main`? Or should I target 2.5?


Seldaek:
Thanks.

PR #11536, opened 2023-07-04
Fix trailing whitespace in 'composer show -N'

dzuelke:
The name column was always padded to maximum width, even if no other columns were printed.

This makes it difficult to use the output e.g. in pipelines.

Fixed for all possible columns, and with tests for two cases (regular show and show outdated).

Again, lmk if I should rebase against 2.2 or 2.5.


Seldaek:
Thanks, fixed the phpstan issues

PR #11543, opened 2023-07-07
curl: Retry download if error 55 is encountered

jeliebig:
With the latest curl release I sometimes encounter the following error:

> curl error 55 while downloading [PACKAGE_JSON_URL]: Failed sending data to the peer

This seems to happen because of a curl issue with HTTP/2: https://github.com/curl/curl/issues/10591

In order to work around this issue I tried two things and noticed they both worked equally as well:
1. Retry the download job and downgrade to HTTP/1.1
2. Retry the download job, but don't change the HTTP version

This PR adds a new check for for curl error 55 and restarts the download job if it was encountered, so composer doesn't error and finish the update/install command as expected.
I added a separate if-statement for that, since this can be removed again, once the curl issue is resolved.

Let me know what you think about this! This is my first PR to composer, so I'm not sure if this counts as a "bug fix", but I'm targeting 2.5, since that's where I encountered the issue the first time.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -357,6 +357,13 @@ class CurlDownloader
                         continue;
                     }
 
+                    // TODO: Remove this as soon as https://github.com/curl/curl/issues/10591 is resolved
+                    if ($errno === 55 /* CURLE_SEND_ERROR */) {
+                        $this->io->writeError('Retrying ('.($job['attributes']['retries'] + 1).') ' . Url::sanitize($job['url']) . ' due to curl error '. $errno, true, IOInterface::DEBUG);
+                        $this->restartJobWithDelay($job, $job['url'], ['retries' => $job['attributes']['retries'] + 1]);
+                        continue;
+                    }
+
                     if ($errno === 28 /* CURLE_OPERATION_TIMEDOUT */ && PHP_VERSION_ID >= 70300 && $progress['namelookup_time'] === 0.0 && !$timeoutWarning) {
                         $timeoutWarning = true;
                         $this->io->writeError('<warning>A connection timeout was encountered. If you intend to run Composer without connecting to the internet, run the command again prefixed with COMPOSER_DISABLE_NETWORK=1 to make Composer run in offline mode.</warning>');


fsmeier:
So the fix on curl side was released in curl 8.2.0 but it seems that this did not fix the issue :( 

```bash
╰─$ curl --version
curl 8.2.0 (x86_64-alpine-linux-musl) libcurl/8.2.0 OpenSSL/3.0.9 zlib/1.2.13 brotli/1.0.9 nghttp2/1.51.0
Release-Date: 2023-07-19
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp ws wss
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL threadsafe TLS-SRP UnixSockets
╭─/data | Store: DE | Env: docker.dev | Debug: (.) | Testing: (.)
╰─$ composer update --lock
Info from https://repo.packagist.org: #StandWithUkraine
https://repo.packagist.org could not be fully loaded (curl error 55 while downloading https://repo.packagist.org/p2/symfony/event-dispatcher-contracts.json: Failed sending data to the peer), package information was loaded from the local cache and may be out of date

In CurlDownloader.php line 377:
                                                                                                                                    
  curl error 55 while downloading https://repo.packagist.org/p2/symfony/event-dispatcher-contracts.json: Failed sending data to th  
  e peer                                                                                                                            
                                                                                                                                    

update [--with WITH] [--prefer-source] [--prefer-dist] [--prefer-install PREFER-INSTALL] [--dry-run] [--dev] [--no-dev] [--lock] [--no-install] [--no-autoloader] [--no-suggest] [--no-progress] [-w|--with-dependencies] [-W|--with-all-dependencies] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] [--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [-i|--interactive] [--root-reqs] [--] [<packages>...]
```

Seldaek:
Thanks, merging so you can at least try with the latest composer snapshot.. Happy if you can send a follow-up to remove it if it ever gets resolved upstream though.

PR #11547, opened 2023-07-10
Tests for base dependency command

rad8329:
This PR adds test cases for `why` and `why-not` as mentioned in #10796

- It adds 20 test cases (now from 198 to 218)
- It increases the test coverage for `BaseDependencyCommand` from 0% to 97%
- It adds full coverage to the `DependsCommand` command (`why` command)
- It adds full coverage to the `ProhibitsCommand` command (`why-not` command)
- It adds poor coverage to `CompletionTrait`, maybe it could be addressed in other PR

Thanks in advantage for your feedback.

### DIFF
--- a/tests/Composer/Test/Command/BaseDependencyCommandTest.php
+++ b/tests/Composer/Test/Command/BaseDependencyCommandTest.php
@@ -13,6 +13,7 @@
 
 namespace Composer\Test\Command;
 
+use Composer\Semver\Constraint\Constraint;
 use Composer\Semver\Constraint\MatchAllConstraint;
 use Symfony\Component\Console\Command\Command;
 use UnexpectedValueException;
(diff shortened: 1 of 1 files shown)


rad8329:
The CI failing step is totally unrelated to my provided changes, plus, when I run `phpstan` locally, everything seems to be fine
image

tests/Composer/Test/Command/BaseDependencyCommandTest.php
```diff
@@ -0,0 +1,436 @@
...
+    ): void {
+        $this->expectException(RuntimeException::class);
+        $this->expectExceptionMessage($expectedExceptionMessage);
+
+        $appTester = $this->getApplicationTester();
+        $this->assertEquals(Command::FAILURE, $appTester->run(['command' => $command] + $parameters));
+    }
+
+    /**
+     * @return Generator<string, array<string|string[]>> [string $command, array $parameters, array $expectedExceptionMessage]
```
stof, line 54: this should be described as `@return Generator<string, array{string, array<string, string>, string}>` instead, so that phpstan can know about the expected array shape

tests/Composer/Test/Command/BaseDependencyCommandTest.php
```diff
@@ -0,0 +1,436 @@
...
+     * Test that SUT will throw an exception when there were not provided some parameters
+     *
+     * @covers       \Composer\Command\BaseDependencyCommand
+     * @covers       \Composer\Command\DependsCommand
+     * @covers       \Composer\Command\ProhibitsCommand
+     *
+     * @dataProvider noParametersCaseProvider
+     *
+     * @param string $command
+     * @param array<mixed> $parameters
```
stof, line 37: `array<string, string>` actually

rad8329:
Thanks @Seldaek  @stof , let me know if you think it should cover other cases.
Sorry for the delay, I was handling some family issues.

Seldaek:
Thanks! And no worries at all, I'm even more delayed than you :D

PR #11552, opened 2023-07-11
GitHubDriver: better handle empty composer.json file

glaubinix:
Having a GitHub repository with an empty composer.json file in the default branch currently fails with `\RuntimeException('Could not retrieve composer.json for main')`. This PR changes the behavior to match all other VCS drivers, where an empty composer.json file is treated the same as if there were no composer.json file.

Example GitHub response for empty files: https://api.github.com/repos/pcom-test-group/empty-composer-json/contents/composer.json?ref=main


Seldaek:
Thanks

PR #11557, opened 2023-07-18
Update BinaryInstaller.php

fr3nch13:
https://github.com/composer/composer/blob/68b7a07187a62435e514324052d5a649b6662efb/src/Composer/Installer/BinaryInstaller.php#L404C4-L404C4

Change `"\${dir}/$binFile" "\$@"`
To `sh "\${dir}/$binFile" "\$@"`

The file being referenced by `"\${dir}/$binFile" "\$@"` isn't always chmod +x so in situation like running in gitlab-ci workflows, it throwing a Permission Denied error.

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -401,7 +401,7 @@ if [ -n "\$bashSource" ]; then
     fi
 fi
 
-"\${dir}/$binFile" "\$@"
+sh "\${dir}/$binFile" "\$@"
 
 PROXY;
     }


Seldaek:
Thanks

fr3nch13:
This may also need to be changed for the 2.2 branch.

Seldaek:
Yes true, tagged it as such, thanks

Seldaek:
I had to revert this due to #11612 - I didn't realize this would take over the default shell but indeed this seems like a bad idea, and as this is just a hack to fix a misconfiguration I think it'd be best that you properly set the execute bit on your executable files instead.

Issue #11555, opened 2023-07-17
Unable to execute phar in vendor/bin


Wulfheart:
My `composer.json`:

```json
{
    "name": "aw/tmp",
    "require": {
        "wulfheart/phpunit-shim": "^10.2"
    }
}

```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK git version 2.34.1
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.5.8
PHP version: 8.2.6
PHP binary path: /usr/bin/php8.2
OpenSSL version: OpenSSL 3.0.2 15 Mar 2022
cURL version: 7.81.0 libz 1.2.11 ssl OpenSSL/3.0.2
zip: extension present, unzip present, 7-Zip not available

```

When I run this command: `composer update -vvv`

I get the following output:

```
Running 2.5.8 (2023-06-09 17:13:21) with PHP 8.2.6 on Linux / 6.2.6-76060206-generic
Reading ./composer.json (/home/aw/Code/priv/tmp/composer.json)
Loading config file /home/aw/.config/composer/config.json
Loading config file /home/aw/.config/composer/auth.json
Loading config file ./composer.json (/home/aw/Code/priv/tmp/composer.json)
Checked CA file /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a file.
Checked directory /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a directory.
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/home/aw/Code/priv/tmp): 'git' 'branch' '-a' '--no-color' '--no-abbrev' '-v'
Executing command (/home/aw/Code/priv/tmp): git describe --exact-match --tags
Executing command (CWD): git --version
[...]

PR #11562, opened 2023-07-21
Set $_SERVER['SCRIPT_NAME'] within proxy command

fredden:
Fixes: https://github.com/composer/composer/issues/11555
Fixes: https://github.com/sebastianbergmann/phpunit/issues/5447

This produces something like the following. This specific example has been taken from the test suite.

```php
if (__FILE__ === realpath($_SERVER['SCRIPT_NAME'])) {
    $_SERVER['SCRIPT_NAME'] = realpath(__DIR__ . '/..'.'/vendor/foo/bar/binary');
}
```

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -359,6 +359,11 @@ namespace Composer;
 
 $globalsCode
 $streamProxyCode
+
+if (__FILE__ === realpath(\$_SERVER['SCRIPT_NAME'])) {
+    \$_SERVER['SCRIPT_NAME'] = realpath($binPathExported);
+}
+
 return include $binPathExported;
 
 PROXY;
--- a/tests/Composer/Test/Installer/BinaryInstallerTest.php
+++ b/tests/Composer/Test/Installer/BinaryInstallerTest.php
@@ -112,6 +112,18 @@ EOL
 echo 'success '.$_SERVER['argv'][1];
 EOL
             ],
+            'php file which validates $_SERVER["SCRIPT_NAME"]' => [<<<'EOL'
+<?php
+if (__FILE__ === realpath($_SERVER['SCRIPT_NAME'])) {
+    echo 'success '.$_SERVER['argv'][1];
+} else {
+    fwrite(STDERR, "Test failure: __FILE__ does not match \$_SERVER['SCRIPT_NAME']\n");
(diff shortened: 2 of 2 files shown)


src/Composer/Installer/BinaryInstaller.php
```diff
@@ -359,6 +359,11 @@ public function url_stat(\$path, \$flags)
 
 $globalsCode
 $streamProxyCode
+
+if (__FILE__ === realpath(\$_SERVER['SCRIPT_NAME'])) {
+    \$_SERVER['SCRIPT_NAME'] = $binPathExported;
```
Seldaek, line 364: We should do a realpath on $binPathExported I think otherwise you end up with a relative path which would not be the full path, but maybe that's fine too I guess SCRIPT_NAME isn't guaranteed to be canonical?

fredden: I wondered about this while writing the code for this. From what I can tell, SCRIPT_NAME holds the same path that was used on the command line.

Screenshot_2023-07-23_20-53-10

The [documentation for `$_SERVER`](https://php.net/_SERVER) says that SCRIPT_NAME contains a "path" whereas `__FILE__` contains a "full path". I don't know if difference in terms there is intentional or significant.

> Contains the current script's path. This is useful for pages which need to point to themselves. The `__FILE__` constant contains the full path and filename of the current (i.e. included) file.

I've updated the pull request to use `realpath()` as requested.

Seldaek:
Thanks

Wulfheart:
@Seldaek out of curiosity: I guess there are plans to make a new release where this change is also included. Any ETA for that release? Thank you in advance.

Seldaek:
@Wulfheart yes, I'll try asap, life is a bit hectic at the moment tho, but definitely my next target is a 2.6.0 release :)

Wulfheart:
@Seldaek thanks for your response. Glad that it is on the table and not in the far future. 
Take care of yourself!

Seldaek:
I had to revert this due to https://github.com/composer/composer/issues/11617 - it was a hacky attempt at fixing it and well it backfired as these things often do.

So I think the best is to just call the phar directly as per https://github.com/sebastianbergmann/phpunit/issues/5447#issuecomment-1704059984 - or someone sends a PR detecting the Composer bin proxy and also allowing that for execution of the phar.  This is detectable by checking for presence of the global `_composer_bin_dir` variable.

PR #11565, opened 2023-07-26
Update Silverstripe casing

n8-dev:
Just updating the casing here; as Silverstripe have changed their casing.

You can see more here
https://www.silverstripe.com/about-us/our-brand/


Seldaek:
Thanks

PR #11581, opened 2023-08-15
Add tests for `DumpAutoloadCommand`

herndlm:
I might have found a tiny bug in `DumpAutoloadCommand`, but before fixing it I wanted to add tests. I'm not well accustomed to Composer or Command tests in general, so I'm not a 100% sure about the changes here.

Kind of open things:
- Is it a good idea to use the default application / composer.json for this? (e.g. could be faster with a temp composer.json I guess)
- ~~Is it a good idea to hardcode the dumped amount of classes? (e.g. dependency/code changs will break this test) Any better ideas of ensuring that the full output is correct without this, e.g. regex?~~ changed to regex since it was already failing here..
- I don't think I can cover actual PSR violations with `--strict-psr` with such a test, can I? I think I'd need to have a temp composer.json AND temp code? https://github.com/composer/composer/blob/95dca79fc2e18c3a4e33f207c1fcaa7d5b559400/src/Composer/Command/DumpAutoloadCommand.php#L111 is the uncovered line

### DIFF
--- a/tests/Composer/Test/Command/DumpAutoloadCommandTest.php
+++ b/tests/Composer/Test/Command/DumpAutoloadCommandTest.php
@@ -1,0 +1,77 @@
+<?php declare(strict_types=1);
+
+namespace Composer\Test\Command;
+
+use Composer\Test\TestCase;
+use InvalidArgumentException;
+
+class DumpAutoloadCommandTest extends TestCase
+{
+    public function testDumpAutoload(): void
+    {
+        $appTester = $this->getApplicationTester();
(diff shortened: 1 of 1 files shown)


Seldaek:
Right you would need to create a composer dir with `$this->initTempComposer()` (see other usages), this can have the autoload config, and then you create a few classes which conflict on disk (initTempCompsoer will `chdir` into a new temp dir, so just write into it using relative paths).

Otherwise current tests look good to me, so will merge this already as is and let you open a follow up if you like.

PR #11508, opened 2023-06-12
Ensure phpstan/rules.neon is exported

fredden:
Fixes https://github.com/composer/composer/issues/11507

PR #11586, opened 2023-08-16
Fix #11507: phpstan/rules.neon missing from archives

### LINKED ISSUE #11507: Regression: phpstan/rules.neon missing in distributed release.
In #10813 the phpstan/rules.neon file was included in the distribution to allow re-use in composer plugins, this was regressed while merging #11484 which did not keep the exclusion.

danepowell:
This fixes #11507 (for real this time 😄 )

#11508 was intended to fix #11507. However, I don't think it worked. If you exclude a directory via .gitattributes, you apparently can't later unexclude a file inside that directory.

The correct fix is to exclude the _contents_ of the directory rather than the directory itself. You can confirm this works locally by checking out this feature branch, running `git archive issue-11507 -o /tmp/composer.tar`, then inspecting the resulting archive to confirm phpstan/rules.neon (and _only_ phpstan/rules.neon) exists.

I was bit by this in https://github.com/acquia/cli/pull/1577 when upgrading to a Composer 2.x-dev release which included the new .gitattributes. Composer's rules.neon suddenly disappeared, producing errors like this:

> Config file /home/runner/work/cli/cli/vendor/composer/composer/phpstan/rules.neon does not exist or isn't readable

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,7 +11,7 @@
 # Exclude non-essential files from dist
 /.github/ export-ignore
 /doc export-ignore
-/phpstan/ export-ignore
+/phpstan/* export-ignore
 /tests/ export-ignore
 /.editorconfig export-ignore
 /.gitattributes export-ignore


stof (approved):

Seldaek:
Thanks!

PR #11590, opened 2023-08-20
allow bumping for patch level version constraints (fixes #11579)

verfriemelt-dot-org:
hi,

i think this is quite a straight forward thing and i think its very benefical to support bumping patch level constraints.
this will enable the mentioned constraint from #11579 to be updated

```json
{
  "require-dev": {
    "phpstan/phpstan": "~1.10.0"
  }
}
```

what do you think?

PR #11594, opened 2023-08-24
update generated baseline for php8.1

verfriemelt-dot-org:
to fix current pipelinefails


verfriemelt-dot-org:
had the wrong base here, set it correctly to `2.5` to fix the issues with #11590

Seldaek:
Thanks but sorry it's my bad the 2.5 branch is closed, it should be deleted but I haven't gotten to do a 2.6 release as quickly as I expected as life got in the way

PR #11597, opened 2023-08-27
COMPOSER_DISABLE_NETWORK aware `diagnose` checks

ktomk:
Make `diagnose` checks aware of COMPOSER_DISABLE_NETWORK (true) and skip Composer network operations that would otherwise spill stack traces into diagnostic messages and taint the result as error while the check itself is not applicable/useful within the environment.

`COMPOSER_DISABLE_NETWORK` was released with [2.0.0-alpha1] and introduced in fc03ab9bb (Add COMPOSER_DISABLE_NETWORK env var for debugging, 2019-01-14).

The previous behaviour was to exit with a status of two (2), denoting an error.

The new behaviour is to exit with a status of zero (0), showing the successful skipping of diagnostics that can only be run when Composer network is enabled - not disabled.

## NOTE: The "prime" Value

It is irrelevant for diagnose checks, as all diagnostic checks that spilled were with the HTTP Downloader and the check is aligned (both "1" or "prime" values disable):

    (bool) Platform::getEnv('COMPOSER_DISABLE_NETWORK')

## NOTE: Not Affected

 * The `allow_url_fopen` diagnostic check, platform related
 * The `disable-tls` setting related HTTP Downloader creation warning

[2.0.0-alpha1]: <https://getcomposer.org/changelog/2.0.0-alpha1> "released 2020-06-03"

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -251,7 +251,7 @@ EOT
      */
     private function checkHttp(string $proto, Config $config)
     {
-        $result = $this->checkConnectivity();
+        $result = $this->checkConnectivityAndComposerNetworkHttpEnablement();
         if ($result !== true) {
             return $result;
         }
@@ -290,7 +290,7 @@ EOT
      */
     private function checkHttpProxy()
     {
-        $result = $this->checkConnectivity();
+        $result = $this->checkConnectivityAndComposerNetworkHttpEnablement();
         if ($result !== true) {
             return $result;
         }
@@ -318,7 +318,7 @@ EOT
      */
     private function checkGithubOauth(string $domain, string $token)
     {
-        $result = $this->checkConnectivity();
+        $result = $this->checkConnectivityAndComposerNetworkHttpEnablement();
         if ($result !== true) {
             return $result;
         }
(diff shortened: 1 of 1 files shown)


src/Composer/Command/DiagnoseCommand.php
```diff
@@ -738,4 +738,36 @@ private function checkConnectivity()
...
+
+    /**
+     * Check if Composer network is enabled for HTTP/S
+     *
+     * @return string|true
+     */
+    private function checkComposerNetworkHttpEnablement()
+    {
+        if ((bool) Platform::getEnv('COMPOSER_DISABLE_NETWORK')) {
+            return '<info>SKIP</> <comment>Composer network is disabled by COMPOSER_DISABLE_NETWORK=1.</>';
```
Seldaek, line 768:
```suggestion
            return '<info>SKIP</> <comment>Network is disabled by COMPOSER_DISABLE_NETWORK.</>';
```

ktomk: Less is more, consistency a beauty, will update, thanks for quick review @Seldaek.

ktomk: Please find the requested changes applied and check if the changes for 736 are to your liking.

Seldaek:
Thanks

ktomk:
Thank you!

Issue #10340, opened 2021-12-08
Composer 2.2 LTS, 2.3 PHP Version Support, 1.x Deprecation


Seldaek:
## 2.2 LTS and the way forward

The upcoming release of **2.2 is going to be an LTS** (Long Term Support) version. We will provide bugfixes for critical bugs and security issues until at least end of 2023, and will reassess then according to the usage we see.

The reason we are doing this is that after over 10 years stuck with a PHP 5.3 requirement, **Composer 2.3 will increase the required PHP version to `>=7.2.5`** and thus stop supporting PHP `5.3.2 - 7.2.4`.

According to our usage metrics this should leave a bit less than 10% of users stuck on the Composer 2.2 LTS until they can upgrade their PHP version. With PHP 7.1 [being EOL for about 2 years](https://www.php.net/supported-versions.php) already it is anyway about time everyone using 7.1 or below would upgrade.

## 1.x Deprecation

The more worrying state of things is that about 1/3rd of composer installs are still run with Composer 1.x. We are committed to keep the v1 repository format working in its [currently reduced capacity](https://blog.packagist.com/deprecating-composer-1-support/) until at least end of 2022, but afterwards we will probably turn it into a read-only snapshot that will not receive updates anymore.

If you are still using Composer 1 I would really urge you to upgrade to Composer 2. If something is blocking you from doing so please do get in touch if you think we can help.

PR #11599, opened 2023-08-28
PHP 8.3 | Tests: fix deprecation notices

jrfnl:
_Re: branch selection: as Composer 2.2 is marked as an LTS release, by rights, this PR should be pulled to the 2.2 branch, but the 2.2 branch isn't being tested against PHP 8.2, which gives me the impression that the "LTS" is only about Composer native functionality, not about PHP version support._

_Might be a good idea to clarify this somewhere ?_

---

### PHP 8.3 | ZipDownloaderTest: fix deprecation notice

Calling `ReflectionProperty::setValue()` with only one argument (to set a static property) is deprecated.
Passing `null` as the first (`$object`) parameter will work cross-version.

As the `ZipDownloaderTest::setPrivateProperty()` method has a `null` default value for the `$obj` parameter anyway, this means the if/else toggle can be removed.

Ref: https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures#reflectionpropertysetvalue

### PHP 8.3 | InstalledVersionsTest: fix deprecation notice

Calling `ReflectionProperty::setValue()` with only one argument (to set a static property) is deprecated.
Passing `null` as the first (`$object`) parameter will work cross-version.

Ref: https://wiki.php.net/rfc/deprecate_functions_with_overloaded_signatures#reflectionpropertysetvalue

### DIFF
--- a/tests/Composer/Test/Downloader/ZipDownloaderTest.php
+++ b/tests/Composer/Test/Downloader/ZipDownloaderTest.php
@@ -65,11 +65,7 @@ class ZipDownloaderTest extends TestCase
         $reflectionClass = new \ReflectionClass('Composer\Downloader\ZipDownloader');
         $reflectedProperty = $reflectionClass->getProperty($name);
         $reflectedProperty->setAccessible(true);
-        if ($obj === null) {
-            $reflectedProperty->setValue($value);
-        } else {
-            $reflectedProperty->setValue($obj, $value);
-        }
+        $reflectedProperty->setValue($obj, $value);
     }
 
     public function testErrorMessages(): void
--- a/tests/Composer/Test/InstalledVersionsTest.php
+++ b/tests/Composer/Test/InstalledVersionsTest.php
@@ -33,14 +33,14 @@ class InstalledVersionsTest extends TestCase
         $prop = new \ReflectionProperty('Composer\Autoload\ClassLoader', 'registeredLoaders');
         $prop->setAccessible(true);
         self::$previousRegisteredLoaders = $prop->getValue();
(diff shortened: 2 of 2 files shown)


Seldaek:
> Re: branch selection: as Composer 2.2 is marked as an LTS release, by rights, this PR should be pulled to the 2.2 branch, but the 2.2 branch isn't being tested against PHP 8.2, which gives me the impression that the "LTS" is only about Composer native functionality, not about PHP version support.

The README (as well as the issue milestone) say: `PHP versions 5.3.2 - 8.1 are still supported via the LTS releases of Composer (2.2.x)`. I think it's clear that it ends at 8.1, but for sure you gotta find this.. Anyway this is fine as is on main branch thanks :)

jrfnl:
> > Re: branch selection: as Composer 2.2 is marked as an LTS release, by rights, this PR should be pulled to the 2.2 branch, but the 2.2 branch isn't being tested against PHP 8.2, which gives me the impression that the "LTS" is only about Composer native functionality, not about PHP version support.
> 
> The README (as well as the issue milestone) say: `PHP versions 5.3.2 - 8.1 are still supported via the LTS releases of Composer (2.2.x)`. I think it's clear that it ends at 8.1, but for sure you gotta find this.. Anyway this is fine as is on main branch thanks :)

Ah, I see and yes, that should be fine. When I searched for info on "Composer 2.2 LTS", [this (pinned) issue](https://github.com/composer/composer/issues/10340) comes up as the first result and it has no mention about the max supported PHP version, which is what caused the confusion.

Seldaek:
Ah good point, I added a note there

PR #11601, opened 2023-08-28
GH Actions: run against PHP 8.3

jrfnl:
_Re: branch selection: as Composer 2.2 is marked as an LTS release, by rights, this PR should be pulled to the 2.2 branch, but the 2.2 branch isn't even being tested against PHP 8.2, which gives me the impression that the "LTS" is only about Composer native functionality, not about PHP version support._

_Might be a good idea to clarify this somewhere ?_

---

### GH Actions: run against PHP 8.3

What with PHP 8.3 being close to the first RC, I'd like to suggest enabling runs against PHP 8.3 for the linting and test runs.

* Linting passes on PHP 8.3, so I propose to not allow new failures to be introduced there.
* The test runs, however, do not pass against PHP 8.3, so I'm marking those as `experimental` for now to allow for fixing the issue(s).

As for the compatibility issues (based on the test runs):
* PR #11599 fixes all known deprecation notices.
* There is, however, one test failure, which I'm not exactly sure how to fix, so I'm leaving this for the maintainers to decide upon.
    Details:
    Prior to PHP 8.3, `ReflectionMethod` could set a `private` method on a parent class to accessible. This is no longer possible in PHP 8.3 since php/php-src#9470 and breaks the `Composer\Test\Repository\ComposerRepositoryTest::testWhatProvides` test.
    Also see: https://3v4l.org/8YcIk/rfc#vgit.master

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -63,6 +63,18 @@ jobs:
             dependencies: highest-ignore
             os: ubuntu-latest
             experimental: false
+          - php-version: "8.3"
+            dependencies: locked
+            os: ubuntu-latest
+            experimental: true
+          - php-version: "8.3"
+            dependencies: lowest-ignore
+            os: ubuntu-latest
+            experimental: true
+          - php-version: "8.3"
+            dependencies: highest-ignore
+            os: ubuntu-latest
+            experimental: true
 
     steps:
       - name: "Checkout"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -31,6 +31,7 @@ jobs:
           - "8.0"
           - "8.1"
           - "8.2"
+          - "8.3"
         dependencies: [locked]
         os: [ubuntu-latest]
         experimental: [false]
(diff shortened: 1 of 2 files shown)


stof:
The test should be changed to do `new ReflectionMethod(ComposerRepository::class, 'whatProvides')` instead of `new ReflectionMethod($repo, 'whatProvides')` (as `$repo` is a a mock so its class is a child class of ComposerRepository)

jrfnl:
> The test should be changed to do `new ReflectionMethod(ComposerRepository::class, 'whatProvides')` instead of `new ReflectionMethod($repo, 'whatProvides')` (as `$repo` is a a mock so its class is a child class of ComposerRepository)

Good point, my brain clearly wasn't working. I've opened PR #11602 to fix this.

jrfnl:
Once PR #11599 and #11602 have been merged, I could rebase this PR and remove the `experimental: true` (or explicitly set it to `false`), if so desired.

Seldaek:
Happy to get a rebase here, and yes experimental false sounds ok if it builds, now that 8.3 is in RC I guess it shouldn't have surprises anymore.

.github/workflows/continuous-integration.yml
```diff
@@ -63,6 +63,18 @@ jobs:
...
+            os: ubuntu-latest
+            experimental: true
+          - php-version: "8.3"
+            dependencies: lowest-ignore
+            os: ubuntu-latest
+            experimental: true
+          - php-version: "8.3"
+            dependencies: highest-ignore
+            os: ubuntu-latest
+            experimental: true
```
Seldaek, line 77: Sorry these in fact should replace the 8.2 above at line 58 and 62, instead of adding new builds. And then at line 33 it needs to be added as well to get the regular locked build. That keeps the amount of builds to a reasonable-ish level. Thanks

.github/workflows/lint.yml
```diff
@@ -22,6 +22,7 @@ jobs:
         php-version:
           - "7.2"
           - "latest"
+          - "8.3"
```
Seldaek, line 25: This isn't needed IMO, as latest will cover 8.3 once it's stable.

.github/workflows/continuous-integration.yml
stof: To reduce the needs to update, should the `lowest-ignore` and `highest-ignore` jobs use `latest` as PHP version ?

jrfnl:
@Seldaek Rebased and updated.

Seldaek:
Thanks

PR #11602, opened 2023-08-28
PHP 8.3 | ComposerRepositoryTest: fix test failure

jrfnl:
Prior to PHP 8.3, `ReflectionMethod` could set a private method on a parent class to accessible. This is no longer possible in PHP 8.3 since php/php-src#9470 and breaks the `Composer\Test\Repository\ComposerRepositoryTest::testWhatProvides` test.
Also see: https://3v4l.org/8YcIk/rfc#vgit.master

Fixed now.

### DIFF
--- a/tests/Composer/Test/Repository/ComposerRepositoryTest.php
+++ b/tests/Composer/Test/Repository/ComposerRepositoryTest.php
@@ -160,7 +160,7 @@ class ComposerRepositoryTest extends TestCase
                 ],
             ]));
 
-        $reflMethod = new \ReflectionMethod($repo, 'whatProvides');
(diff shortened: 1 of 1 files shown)


stof (approved):

Seldaek:
Thanks

jrfnl:
FYI: the bug fix in PHP 8.3 has been reverted for now: https://github.com/php/php-src/issues/9470#issuecomment-1702536651
I think this fix can stay in though as it is a more correct way of approaching a private method in a parent class than the original code.

Seldaek:
Yeah it's fine as is :)

PR #11605, opened 2023-08-30
Change audit.ignore behavior before 2.6.0

### LINKED ISSUE #11604: Before 2.6.0: Replace ability to ignore vulnerabilities reported by Composer audit with a bypass option instead
Composer 2.6.0 is planning to introduce [a new ignore feature](https://github.com/composer/composer/blob/879524d8e3fc04a0bd9a1b6f29ea49f1280741c0/doc/06-config.md?plain=1#L108-L120) for security advisories reported by `composer audit`. The original request was raised for this feature in #11298.

Before the implementation was merged the issue got little traction but after it was merged, it got controversial feedback. As you can see from the thread, personally, I am on the side that a feature like this should not exist.

Today, @janmashat came up with a "you can eat the cake but also keep it" type of idea, a different one that I was suggested before with `composer audit-changes`:

> Rather than an "audit.ignored config setting to ignore security advisories" perhaps we could agree on this compromise: "audit.pass config setting to exit(0) for certain security advisories" (but still list them in the console output).
>
> That way, if somebody is reviewing the "audit log" then they won't have to wonder whether some entries are missing.

(Source: https://github.com/composer/composer/issues/11298#issuecomment-1699177603)

mxr576:
Closes #11604


src/Composer/Advisory/Auditor.php
```diff
@@ -50,28 +50,30 @@ class Auditor
...
-            $plurality = $totalAdvisories === 1 ? 'y' : 'ies';
-            $pkgPlurality = $affectedPackages === 1 ? '' : 's';
+        if (count($allAdvisories) > 0) {
+            [$affectedPackagesCount, $totalAdvisoryCount] = $this->countAdvisories($allAdvisories);
+            $plurality = $totalAdvisoryCount === 1 ? 'y' : 'ies';
+            $pkgPlurality = $affectedPackagesCount === 1 ? '' : 's';
             $punctuation = $format === 'summary' ? '.' : ':';
-            $io->writeError("<$errorOrWarn>Found $totalAdvisories security vulnerability advisor{$plurality} affecting $affectedPackages package{$pkgPlurality}{$punctuation}</$errorOrWarn>");
-            $this->outputAdvisories($io, $advisories, $format);
+            $io->writeError("<$errorOrWarn>Found $totalAdvisoryCount security vulnerability advisor{$plurality} affecting $affectedPackagesCount package{$pkgPlurality}{$punctuation}</$errorOrWarn>");
```
mxr576, line 71: Do we want to change/extend this wording to better explain that "X number of vulnerabilities found, Y amount were silenced so X-Y caused the command failure"?

Seldaek: Yes definitely would need to clarify wording IMO.

res/composer-schema.json
```diff
@@ -329,9 +329,9 @@
                     "type": "object",
                     "description": "Security audit configuration options",
                     "properties": {
-                        "ignored": {
+                        "pass": {
```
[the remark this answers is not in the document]
Seldaek: Not a fan of pass, but I will need to sleep over this one I think :)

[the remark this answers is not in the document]
mxr576: Me neither, maybe ignore remain as config key with different meaning.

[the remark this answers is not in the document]
mxr576: My current winner is `suppressed` and I am also thinking about turning the suppressed list to an associative array where the key would be the cve id and the value would be a mandatory human readonly description about why a vulnerability report was suppressed on the given project.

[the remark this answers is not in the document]
Seldaek: I think I'd prefer `ignore` (not `ignored` as it was, that's my bad, `ignore` is more consistent with other directives in composer.json). But I agree having a reason in there makes sense for self documenting it. That's a good improvement.

src/Composer/Advisory/Auditor.php
```diff
@@ -44,34 +44,50 @@ class Auditor
...
-            $advisories = $this->filterIgnoredAdvisories($advisories, $ignoredIds);
+          if (!\array_is_list($ignoredIds)) {
+              $ignoredIds = array_keys($ignoredIds);
+          }
+          $advisories = $this->filterIgnoredAdvisories($advisories, $ignoredIds);
         }
 
         if (self::FORMAT_JSON === $format) {
-            $io->write(JsonFile::encode(['advisories' => $advisories]));
+            $io->write(JsonFile::encode(['advisories' => $allAdvisories]));
```
mxr576, line 64: How about introducing a new JSON output format that also contains the ignored vulnerabilities?

```json
{
  "advisories": {},
  "ignored-advisories": {}
}
```

For enabling the new format in the 2.x branch we could leverage a logic like ` if (Platform::getEnv('COMPOSER_AUDIT_FORMAT_VERSION') === '2.6.0') {}` or simiar.

src/Composer/Advisory/Auditor.php
```diff
@@ -44,34 +44,50 @@ class Auditor
...
+            }
+
+            $ignoredAdvisories = $this->diffAdvisories($allAdvisories, $advisories);
+            [$ignoredAffectedPackagesCount, $ignoredTotalAdvisoryCount] = $this->countAdvisories($ignoredAdvisories);
+            if ($ignoredAffectedPackagesCount > 0) {
+                $plurality = $ignoredTotalAdvisoryCount === 1 ? 'y' : 'ies';
+                $pkgPlurality = $ignoredAffectedPackagesCount === 1 ? '' : 's';
+                $punctuation = $format === 'summary' ? '.' : ':';
+                $io->writeError("<info>$ignoredTotalAdvisoryCount ignored security vulnerability advisor{$plurality} affecting $ignoredAffectedPackagesCount package{$pkgPlurality}{$punctuation}</info>");
+                $this->outputAdvisories($io, $ignoredAdvisories, $format);
```
mxr576, line 87: Since we may have the reason for ignoring a given security advisory, I'd consider adding that to the output (also to the JSON above). What do you think?

res/composer-schema.json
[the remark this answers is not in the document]
mxr576: Switched back to `ignore` after checked the meaning of "suppress" vs "ignore".

src/Composer/Advisory/Auditor.php
mxr576: Hm, maybe we do not even need a feature flag since we would only extend the object with a new property... :thinking:

Seldaek: Right, that sounds good to me, that's why I never do json objects with top level data directly. Always have top level keys so you can extend in a BC way :)

Seldaek: And I think I would not include the ignored ones in the `advisories` key in the json. I know they're all listed in one on the text output but for json it probably makes more sense to split them up.

src/Composer/Advisory/Auditor.php
Seldaek: Yes makes sense to me. It probably needs to be highlighted like `<warning>Ignored: $reason</>`

And for the table output add an Ignore column with `Yes`/empty and no reason to keep it small perhaps, because otherwise the table width will blow up

src/Composer/Advisory/Auditor.php
```diff
@@ -44,34 +44,50 @@ class Auditor
...
     {
-        $advisories = $repoSet->getMatchingSecurityAdvisories($packages, $format === self::FORMAT_SUMMARY);
+        $allAdvisories = $advisories = $repoSet->getMatchingSecurityAdvisories($packages, $format === self::FORMAT_SUMMARY);
 
         if (\count($ignoredIds) > 0) {
-            $advisories = $this->filterIgnoredAdvisories($advisories, $ignoredIds);
+          if (!\array_is_list($ignoredIds)) {
+              $ignoredIds = array_keys($ignoredIds);
+          }
+          $advisories = $this->filterIgnoredAdvisories($advisories, $ignoredIds);
```
Seldaek, line 60: Indent seems off

```suggestion
            if (!\array_is_list($ignoredIds)) {
                $ignoredIds = array_keys($ignoredIds);
            }
            $advisories = $this->filterIgnoredAdvisories($advisories, $ignoredIds);
```

src/Composer/Advisory/Auditor.php
mxr576: > Yes makes sense to me. It probably needs to be highlighted like <warning>Ignored: $reason</>

Just to be clear, do you mean display all ignore reasons as a console message unless the output format is JSON?

Speaking of JSON, partially because the JSON output leverages that advisories are JSON serializable, should I just introduce a new `class IgnoredSecurityAdvisory extends SecurityAdvisory{}` that can also hold the reason?

src/Composer/Advisory/Auditor.php
mxr576: Added.

src/Composer/Advisory/Auditor.php
Seldaek: Yeah I was also wondering how to best achieve this. For sure for JSON it would be easier with a subclass.. 

For the plain text output I meant it probably should go here https://github.com/composer/composer/blob/main/src/Composer/Advisory/Auditor.php#L207 as a new line with a <warning> so it is highlighted as being ignored.

src/Composer/Advisory/Auditor.php
```diff
@@ -44,34 +44,59 @@ class Auditor
...
+        $allAdvisories = $advisories = $repoSet->getMatchingSecurityAdvisories($packages, $format === self::FORMAT_SUMMARY);
+        $ignoredAdvisories = [];
 
         if (\count($ignoredIds) > 0) {
-            $advisories = $this->filterIgnoredAdvisories($advisories, $ignoredIds);
+          if (!\array_is_list($ignoredIds)) {
+              $ignoredIds = array_keys($ignoredIds);
+          }
+          $advisories = $this->filterIgnoredAdvisories($advisories, $ignoredIds);
+          $ignoredAdvisories = $this->diffAdvisories($allAdvisories, $advisories);
```
Seldaek, line 62: Instead of adding a new method here I think I'd change filterIgnoredAdvisories to return `[$advisories, $ignoredAdvisories]` so you get it done in one go. Returning both there shouldn't be adding much complexity.

mxr576: Done: https://github.com/composer/composer/compare/99551b7f3f2478d1e5982af48f4b181a6c0edb8a..e570c5a1b9eaf67076c2336651fc6c4adac35576

This probably also helps with casting the ignored advisories to a new object as a next step

src/Composer/Advisory/Auditor.php
mxr576: It seems to me that wrapping text in tables works OOTB, so is there any need for workarounds like...?

```
                $ignoreReason = '';
                if ($advisory instanceof IgnoredSecurityAdvisory && $advisory->ignoreReason !== NULL) {
                    if (strlen($advisory->ignoreReason) > 80) {
                        $ignoreReason = substr($advisory->ignoreReason, 0, 77) . '...';
                    }
                    else {
                        $ignoreReason = $advisory->ignoreReason;
                    }
                }
```

image

src/Composer/Advisory/IgnoredPartialSecurityAdvisory.php
```diff
@@ -0,0 +1,39 @@
...
+        $this->ignoreReason = $ignoreReason;
+    }
+
+    /**
+     * @return mixed
+     */
+    #[\ReturnTypeWillChange]
+    public function jsonSerialize()
+    {
+        $data = parent::jsonSerialize();
```
mxr576, line 32: I do not like traits with parent method calls so I rather copy-pasted these couple of lines.

src/Composer/Advisory/SecurityAdvisory.php
```diff
@@ -42,7 +42,7 @@ class SecurityAdvisory extends PartialSecurityAdvisory
     public $reportedAt;
 
     /**
-     * @var array<array{name: string, remoteId: string}>
+     * @var non-empty-array<array{name: string, remoteId: string}>
```
mxr576, line 45: PHPStan told me after that this was misaligned with the constructor.

src/Composer/Advisory/Auditor.php
mxr576: https://github.com/composer/composer/compare/e570c5a1b9eaf67076c2336651fc6c4adac35576..708db5e9a4a47135753f938e87b3ba5646c4e1d0

src/Composer/Advisory/IgnoredSecurityAdvisory.php
```diff
@@ -0,0 +1,41 @@
...
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Advisory;
+
+
+class IgnoredSecurityAdvisory extends SecurityAdvisory {
```
Seldaek, line 16: Could be marked final and `@internal` too I think.

src/Composer/Advisory/Auditor.php
```diff
@@ -81,36 +97,79 @@ public function audit(IOInterface $io, RepositorySet $repoSet, array $packages,
...
-                return true;
-            });
-            if (\count($advisories[$package]) === 0) {
-                unset($advisories[$package]);
+                if ($isFiltered) {
+                    $filtered[$package][] = $advisory;
+                }
+                else {
+                    if ($advisory instanceof SecurityAdvisory) {
+                        $advisory = new IgnoredSecurityAdvisory(
```
stof, line 149: should we create an object that extends SecurityAdvisory in which you copy all the metadata or should we rather have a `IgnoredSecurityAdvisory` that would take a `SecurityAdvisory|PartialSecurityAdvisory` as argument and implements its `jsonSerialize` method by adding an ignore_reason key in the array returned by the delegated call ?

src/Composer/Advisory/IgnoredPartialSecurityAdvisory.php
mxr576: Sounds reasonable, although the typehint on `\Composer\Advisory\Auditor::processAdvisories() ` (used to call `filterIgnoredAdvisories()`) said that PartialSecurityAdvisorys are potential value.

src/Composer/Advisory/Auditor.php
```diff
@@ -44,34 +44,50 @@ class Auditor
      * @param PackageInterface[] $packages
      * @param self::FORMAT_* $format The format that will be used to output audit results.
      * @param bool $warningOnly If true, outputs a warning. If false, outputs an error.
-     * @param string[] $ignoredIds Ignored advisory IDs, remote IDs or CVE IDs
+     * @param array<string>|array<string,string> $ignoreList List of advisory IDs, remote IDs or CVE IDs that reported but not listed as vulnerabilities.
```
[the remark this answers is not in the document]
mxr576: You are absolutely right, so rolling back to `string[]`.

src/Composer/Advisory/Auditor.php
mxr576: These were the options that I also considered, the solution in the PR just felt better, even if it is "ugly" a bit.

src/Composer/Advisory/IgnoredPartialSecurityAdvisory.php
mxr576: How about this?


https://github.com/composer/composer/compare/f7aff76e2d1dec71f0dfd392f5dc1b6398dbca8f..89eb824f52e19bf39ac87aef382783fcee0b8690

mxr576: Got rid of the IgnoredPartialSecurityAdvisory

src/Composer/Advisory/Auditor.php
fredden: Perhaps `IgnoredSecurityAdvisory` could be a proxy class or have a factory method which takes in a `SecurityAdvisory` object. That'd make the code here much cleaner.

`$advisory = IgnoredSecurityAdvisory::createFromSecurityAdvisory($advisory);` or `$advisory = new IgnoredSecurityAdvisory($advisory);`

src/Composer/Advisory/Auditor.php
```diff
@@ -81,36 +97,78 @@ public function audit(IOInterface $io, RepositorySet $repoSet, array $packages,
...
+                            $advisory->packageName,
+                            $advisory->advisoryId,
+                            $advisory->affectedVersions,
+                            $advisory->title,
+                            $advisory->sources,
+                            $advisory->reportedAt,
+                            $advisory->cve,
+                            $advisory->link
+                        );
+                        if ($ignoreReason !== NULL) {
```
fredden, line 161: I thought that the reason was a required field.

src/Composer/Advisory/Auditor.php
mxr576: `IgnoredSecurityAdvisory::createFromSecurityAdvisory($advisory`  I would have gone with this approach by default, but since everything in public on these objects (as public properties) this approach just moves copying object data to somewhere else. Would that have a real added value? In this case, should I make the constructor of `IgnoredSecurityAdvisory` private so it could be only constructed from a SecurityAdvisory from now?

src/Composer/Advisory/Auditor.php
mxr576: Actually I decided to give more freedom to developers... so it cannot be mandatory.
https://github.com/composer/composer/pull/11605/files#diff-89992649ff6652d737d4dac645fbbac6ceeb112ab29010d0eaba3b42c658dc7eR112-R132

res/composer-schema.json
```diff
@@ -329,12 +329,23 @@
...
-                            "type": "array",
-                            "description": "A set of advisory ids, remote ids or CVE ids that should be ignored and not reported as part of an audit.",
-                            "items": {
-                                "type": "string"
-                            }
+                        "ignore": {
+                            "anyOf": [
+                                {
+                                    "type": "object",
+                                    "description": "A list of advisory ids, remote ids or CVE ids (keys) and the explanations (values) for why they're being ignored. The listed items are reported but they do not make the audit command fail.",
```
naderman, line 336:
```suggestion
                                    "description": "A list of advisory ids, remote ids or CVE ids (keys) and the explanations (values) for why they're being ignored. The listed items are reported but let the audit command pass.",
```

naderman (changes requested):

Seldaek:
@mxr576 please don't change anything anymore, I'm editing a couple things here and then merging :) thanks a lot for the quick work here!

Seldaek:
I think it looks good now but I perhaps broke some tests.. anyway bedtime, will wrap up tomorrow.

mxr576:
> @mxr576 please don't change anything anymore, I'm editing a couple things here and then merging :) thanks a lot for the quick work here!

Roger! 🙂 Time well spent! 🍻

naderman (approved):

Seldaek:
Ok looking good, thanks again.

PR #11606, opened 2023-08-30
Add test case for `DumpAutoloadCommand` with failing --strict-psr check

herndlm:
Follow-up for missing test case mentioned in https://github.com/composer/composer/pull/11581


Seldaek:
Thanks

PR #11607, opened 2023-08-30
Allow --strict-psr in `DumpAutoloadCommand` also with --classmap-authoritative

herndlm:
The "bug" I mentioned in https://github.com/composer/composer/pull/11581

Since `--classmap-authoritative` automatically enables level 1 optimisations (https://getcomposer.org/doc/articles/autoloader-optimization.md#what-does-it-do--2) `--strict-psr` should be useable too without having to pass `--optimise` again.


stof (approved):

Seldaek:
Thanks

Issue #11580, opened 2023-08-15
Add the ability to detect PSR errors without changing files


TravisCarden:
**tl;dr:** I want to be able to detect PSR mapping autoloader errors without side effects, i.e., without actually writing autoload files.

- Background: https://github.com/phpro/grumphp/issues/1088

I recently discovered that `composer validate` will pass even if there are PSR-4 problems--which makes sense if you consider that PSR errors aren't technically located in `composer.json`, but it still surprised me.

I found `composer dump-autoload --strict-psr`, which does exactly what I want except that it has side-effects, i.e., it actually changes files on the filesystem. Could an option be added to skip that part? Something like a `--dry-run` option perhaps?

```
$ composer dump-autoload --strict-psr --dry-run
Class Example\Good located in ./src/Bad.php does not comply with psr-4 autoloading standard. Skipping.
Would have generated optimized autoload files containing 1 classes
```

Seldaek:
Yeah sounds possible to add a dry run, the process in AutoloadGenerator is kinda clear cut with generating everything first then writing to disk at the end so I think it shouldn't be too difficult.

TravisCarden:
^^ Something along these lines, perhaps, @Seldaek?

PR #11608, opened 2023-08-30
Add `dumpautoload --dry-run` option

TravisCarden:
Fixes https://github.com/composer/composer/issues/11580.

Note: Because this indents a large block of code without actually changing it, I recommend reviewing the changes with whitespace hidden to reduce noise ([like so](https://github.com/composer/composer/pull/11608/files?w=1)).


Seldaek:
Thanks

PR #11362, opened 2023-03-03
Exit with non-zero code if post-install audit fails

IonBazan:
Fixes #11357 by adding a new `ERROR_AUDIT_FAILED = 5` exit code used when running `composer install --audit`.

PR #11616, opened 2023-09-01
Fix exit code 5 on composer require/create-project command

vtsykun:
After #11362, CI builds may fail with status code 5 if the composer command require/create-project is used.
This PR fix this BC break and the command will fail if the user has explicitly specified the `--audit` flag

```
composer install --audit
```

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -170,6 +170,8 @@ class Installer
     protected $executeOperations = true;
     /** @var bool */
     protected $audit = true;
+    /** @var bool */
+    protected $errorOnAudit = false;
     /** @var Auditor::FORMAT_* */
     protected $auditFormat = Auditor::FORMAT_SUMMARY;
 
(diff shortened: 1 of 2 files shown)


src/Composer/Installer.php
```diff
@@ -402,7 +404,7 @@ public function run(): int
                         $repoSet->addRepository($repo);
                     }
 
-                    return $auditor->audit($this->io, $repoSet, $packages, $this->auditFormat, true, $this->config->get('audit')['ignore'] ?? []) > 0 ? self::ERROR_AUDIT_FAILED : 0;
+                    return $auditor->audit($this->io, $repoSet, $packages, $this->auditFormat, true, $this->config->get('audit')['ignore'] ?? []) > 0 && true === $this->errorOnAudit ? self::ERROR_AUDIT_FAILED : 0;
```
stof, line 407: `true ===` is useless as `$this->errorOnAudit` is already a boolean

Seldaek:
Thanks, probably makes sense to do this only on request

Issue #5447, opened 2016-06-20
Change "outdated" color semantic


mbrodala:
The `outdated` command currently outputs updates in red; if the update is breaking according to semver it is displayed in yellow.

Since red matches the "error" state and yellow the "warning" state, one would expect the less breaking change to be yellow and the definitely breaking change to be red. Thus I think these two colors should be swapped.

What do you think?

alcohol:
I think that would better match expectations indeed.

Seldaek:
I rather keep "mind compatibility" with npm when possible :)

c33s:
basically i understand the decision to match npm but sticking with something semantically wrong only because other system stick to a wrong semantic is maybe not the best choice.

sidenote:
thank you for this awesome project. composer is really a rocksolid application. thank you for all your work.

ThomasLandauer:
I just suggested to add a "legend" below the output to clarify the meaning of the colors: https://github.com/composer/composer/issues/9716

PR #11617, opened 2023-09-03
Revert "Set $_SERVER['SCRIPT_NAME'] within proxy command"

### LINKED ISSUE #11615: Longer backtrace when installed using 2.6.0
with composer 2.6.0/2.6.1:

CI output with composer 2.6.1: https://github.com/atk4/core/actions/runs/6053504555/job/16429145566#step:8:154

repro: https://github.com/atk4/core/commit/3a60dd943aaeb955d9bd5c86bfdb583cc93bbe98 - one phpunit test is failing

---

with composer 2.5.8:

CI output with composer 2.5.8: https://github.com/atk4/core/actions/runs/6053551255/job/16429280994#step:8:139

repro: https://github.com/atk4/core/commit/d54a022841067014d48015660459139e7c53a8ef - phpunit tests are passing

---

the 1st repro is different from the 2nd only in the composer version

maybe this change is wanted/needed, but please confirm, I find lower backtrace depth better

mvorisek:
reverts #11562

fixes #11615

why the PR should be reverted?

- `$_SERVER['SCRIPT_NAME']` is expected to be the entry file, this is a php-src convenction and it should be safe to assume it no matter how the script is started
- the entry file path must match the first path in a backtrace, otherwise shortening backtraces might not work (as in #11615)
- original problem #5447 mentioned https://github.com/sebastianbergmann/phpunit/blob/15a89f123d/build/templates/binary-phar-autoload.php.in#L58-L62 - so the fix was done to fix this - incorrect/not always valid assumption - and the proper fix should be to fix sebastianbergmann/phpunit instead (and possibly with some added global/API in composer like "entry target file" in https://github.com/mvorisek/composer/blob/6200985c2a/src/Composer/Installer/BinaryInstaller.php#L224)

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -359,11 +359,6 @@ namespace Composer;
 
 $globalsCode
 $streamProxyCode
-
-if (__FILE__ === realpath(\$_SERVER['SCRIPT_NAME'])) {
-    \$_SERVER['SCRIPT_NAME'] = realpath($binPathExported);
-}
-
 return include $binPathExported;
 
 PROXY;
--- a/tests/Composer/Test/Installer/BinaryInstallerTest.php
+++ b/tests/Composer/Test/Installer/BinaryInstallerTest.php
@@ -112,18 +112,6 @@ EOL
 echo 'success '.$_SERVER['argv'][1];
 EOL
             ],
-            'php file which validates $_SERVER["SCRIPT_NAME"]' => [<<<'EOL'
-<?php
-if (__FILE__ === realpath($_SERVER['SCRIPT_NAME'])) {
-    echo 'success '.$_SERVER['argv'][1];
-} else {
-    fwrite(STDERR, "Test failure: __FILE__ does not match \$_SERVER['SCRIPT_NAME']\n");
-    fwrite(STDERR, "\t__FILE__:\t" . __FILE__ . "\n");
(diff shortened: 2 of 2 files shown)


Seldaek:
Yeah I would still like to understand and look at the code producing the backtrace which causes your failure. Is there any chance you have a link to that?

mvorisek:
here is a repo: https://github.com/atk4/core with CI/test failure starting with composer 2.6.0

the actual single usage should not matter much, as more projects can rely on native php-src properties

Seldaek:
It matters to me to understand what kind of code causes the failure.. I know the repo is there but I didn't figure out from a quick look the source of the problem, as I am not familiar with it at all. I imagine it'd be easier for you to find out, so it'd be nice if you help.

mvorisek:
Sure and thank you.

in phpunit printer the stack trace is redacted by
https://github.com/sebastianbergmann/phpunit/blob/899db92716/src/Framework/ExceptionWrapper.php#L53C58-L53C58
which calls
https://github.com/sebastianbergmann/phpunit/blob/899db92716/src/Util/Filter.php#L64
which redacts the stack trace lines in
https://github.com/sebastianbergmann/phpunit/blob/899db92716/src/Util/Filter.php#L90
and https://github.com/sebastianbergmann/phpunit/blob/899db92716/src/Util/Filter.php#L98

also `realpath()`  is slow (requires many IO syscalls) and can even fail (silently, `realpath` returns `false`) if the `$binPathExported` is not on physical disk - repro https://3v4l.org/V9GMD (see last line, wrapper is copied example from https://www.php.net/manual/en/stream.streamwrapper.example-1.php)

Seldaek:
Ok yeah I see, thanks.

Issue #11540, opened 2023-07-06
Difference between COMPOSER_MEMORY_LIMIT=X and memory_limit=X


jontjs:
I suspect this is expected behaviour, but I think it might be worth pointing out in the documentation (https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors).

When running `php -d memory_limit=X /usr/local/bin/composer install` I get a memory limit error when it runs Symfony's `cache:clear` under `auto-scripts`. I understand this is because composer does not pass its PHP options to the the `cache:clear` external process, which is fair enough.

However, when running `COMPOSER_MEMORY_LIMIT=X composer install` I do not run into the memory limit error. I guess this is because the environment variable exists for all processes triggered by the command, which is also fair enough.

I think it's worth mentioning this distinction in the documentation at https://getcomposer.org/doc/articles/troubleshooting.md#memory-limit-errors.

Seldaek:
You are correct that the env var propagates while the flag does not, but I'm not sure why symfony's cache:clear would respect Composer's memory limit env var. That seems weird.

Seldaek:
We definitely could mention in the docs that this will only fix Composer specific memory issues though, and child processes are on their own and should be fixed where relevant.

abreel:
@Seldaek Has this been handled? Can I look into this?

Seldaek:
If you like to PR the docs as per my last comment above feel free yes.

abreel:
okay, @Seldaek 

I will look into that and revert

abreel:
@Seldaek, made pr

https://github.com/composer/composer/pull/11624

PR #11624, opened 2023-09-08
updated memory limit distinction for memory limit errors in docs

abreel:
Updated docs to fix the updates on the memory limit errors mentioned in the issue below, that was approved by @Seldaek 

https://github.com/composer/composer/issues/11540#issue-1791143502


Seldaek:
Thanks

Issue #11626, opened 2023-09-08
Regression in dependency resolution since 2.6.0


T2L:
Unable to update `drupal/paragraphs` package from version 1.15.0 to 1.16.0 (latest atm) with Composer 2.6. See the attached lock file to reproduce.

See [composer.zip](https://github.com/composer/composer/files/12559217/composer.zip)

My `composer.json`:

```json
{
    "type": "project",
    "require": {
        "drupal/classy": "^1.0",
        "drupal/core-recommended": "^9.2",
        "drupal/paragraphs": "^1.15"
    },
    "repositories": {
        "_drupal": {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
        
    },
    "minimum-stability": "dev",
    "prefer-stable": true
}
```

Output of `composer diagnose`:

```
❯ composer diagnose
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK git version 2.42.0
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.6.2
PHP version: 8.1.23
PHP binary path: /usr/local/Cellar/php@8.1/8.1.23/bin/php
OpenSSL version: OpenSSL 3.1.2 1 Aug 2023
cURL version: 8.2.1 libz 1.2.11 ssl (SecureTransport) OpenSSL/3.1.2
zip: extension present, unzip present, 7-Zip not available
```

When I run this command:

```
composer update drupal/paragraphs --with-all-dependencies --no-interaction --prefer-dist --prefer-stable -vv
```

I get the following output: 

```
Loading composer repositories with package information
Pool optimizer completed in 2.586 seconds
Found 12,346 package versions referenced in your dependency graph. 11,376 (92%) were optimized away.
Updating dependencies
Dependency resolution completed in 0.002 seconds
Your requirements could not be resolved to an installable set of packages.

  Problem 1
[...]

PR #11629, opened 2023-09-11
Fixed base package replacers being removed incorrectly

Toflar:
Fixes https://github.com/composer/composer/issues/11626

If the replaced package is part of the require request itself, we must also load it. We actually incorrectly adjusted the test in https://github.com/composer/composer/pull/11449.

I hope the changes are correct, requesting a review by @naderman :)

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -669,9 +669,8 @@ class PoolBuilder
 
     private function markPackageNameForLoadingIfRequired(Request $request, string $name): void
     {
-        if (isset($request->getRequires()[$name])) {
+        if ($this->isRootRequire($request, $name)) {
             $this->markPackageNameForLoading($request, $name, $request->getRequires()[$name]);
-            return;
         }
 
         foreach ($this->packages as $package) {
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -669,6 +669,11 @@ class PoolBuilder
 
     private function markPackageNameForLoadingIfRequired(Request $request, string $name): void
     {
+        if (isset($request->getRequires()[$name])) {
+            $this->markPackageNameForLoading($request, $name, $request->getRequires()[$name]);
+            return;
+        }
+
         foreach ($this->packages as $package) {
             foreach ($package->getRequires() as $link) {
                 if ($name === $link->getTarget()) {
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/load-replaced-root-package-if-replacer-dropped.test
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/load-replaced-root-package-if-replacer-dropped.test
@@ -1,0 +1,49 @@
+--TEST--
+Ensure that a root package gets loaded which is replaced by old versions of another requirement
+
+--REQUEST--
+{
+    "require": {
+        "root/dep": "*",
+        "replaced/pkg": "1.0.0"
+    },
+    "locked": [
+        {"name": "root/dep", "version": "1.1.0", "require": {"replacer/pkg": "1.*"}},
+        {"name": "replacer/pkg", "version": "1.1.0"},
+        {"name": "replaced/pkg", "version": "1.0.0"}
+    ],
+    "allowList": [
+        "root/dep"
+    ],
+    "allowTransitiveDeps": true
+}
+
+--FIXED--
+[
+]
+
+--PACKAGE-REPOS--
+[
+    [
+        {"name": "root/dep", "version": "1.2.0", "require": {"replacer/pkg": "1.*"}},
+        {"name": "replacer/pkg", "version": "1.0.0", "replace": {"replaced/pkg": "1.0.0"}},
+        {"name": "replacer/pkg", "version": "1.1.0"},
+        {"name": "replaced/pkg", "version": "1.0.0"}
+    ]
+]
+
+--EXPECT--
+[
+    "replaced/pkg-1.0.0.0",
+    "replacer/pkg-1.0.0.0",
+    "replacer/pkg-1.1.0.0",
+    "root/dep-1.2.0.0"
+]
+
+--EXPECT-OPTIMIZED--
+[
+    "replaced/pkg-1.0.0.0",
+    "replacer/pkg-1.0.0.0",
+    "replacer/pkg-1.1.0.0",
+    "root/dep-1.2.0.0"
+]
--- a/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test
+++ b/tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/multi-repo-replace-partial-update-all.test
@@ -97,6 +97,7 @@ Check that replacers from additional repositories are loaded when doing a partia
 
 --EXPECT--
 [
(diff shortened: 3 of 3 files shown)


src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -669,6 +669,11 @@ private function unlockPackage(Request $request, array $repositories, string $na
 
     private function markPackageNameForLoadingIfRequired(Request $request, string $name): void
     {
+        if (isset($request->getRequires()[$name])) {
```
naderman, line 672: Shouldn't `$this->packages` below always contain the root package?

naderman: I guess we don't actually stick the root package in there at all, so yes of course need to handle the root requires here too ...

naderman (approved):

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: It actually is part of the `$this->packages`. Maybe This whole method should handle the replaces as well? Like so?

```php
    private function markPackageNameForLoadingIfRequired(Request $request, string $name): void
    {
        foreach ($this->packages as $package) {
            foreach ($package->getRequires() as $link) {
                if ($name === $link->getTarget()) {
                    $this->markPackageNameForLoading($request, $link->getTarget(), $link->getConstraint());
                }
            }
            foreach ($package->getReplaces() as $link) {
                if ($name === $link->getTarget()) {
                    $this->markPackageNameForLoading($request, $link->getTarget(), $link->getConstraint());
                }
            }
        }
    }
```

Toflar: The problem in https://github.com/composer/composer/issues/11626 is that the `composer.json` requires a package that is replaced :)

naderman: Huh? That makes no sense at all? We definitely do not want to load packages just cause something else is replacing them? The idea of the function is to load in case there is a requirement for the package, not something else is replacing it?

naderman: Okay I'll take a step back maybe and think about the actual bug and rethink this all. Not sure I can really do that today though XD

naderman: @Toflar Maybe you can leave some more context on that other ticket? The sample composer.json doesn't have any replace at all?

Toflar: Yeah, same 😅  The problem is: The `composer.json` requires `drupal/classy`. This package is not required by any package in `$this->packages` though, so it's not loaded. Hence, we have a dependency resolution problem.
But in `$this->packages` we have packages that `replace` `drupal/classy` - those are the ones that should be loaded I guess 🤔

Toflar: I will mark this discussion as resolved, as it's irrelevant and thus noise. I'll reply again.

Toflar:
I've investigated this problem more in-depth and I don't think this is the right solution. In fact, I think the bug is not even related to https://github.com/composer/composer/pull/11449 at all. I'm assuming that 11449 just by coincidence raised another issue to the surface. I'm trying to explain where I think the actual bug resides:

If you run the test case from https://github.com/composer/composer/issues/11626 and diff the logic between 2.5.8 and current main, for both versions, the bug is actually that `drupal/classy` is removed from `$this->packages` although it shouldn't. This is when it gets removed:

Bildschirmfoto 2023-09-11 um 17 01 36

In 2.5.8, however, this was not an issue because the package was re-added to the pool again in a later step. This does not happen in `main` anymore due to https://github.com/composer/composer/pull/11449.

So I think the actual bug here is the `removeLoadedPackage()` call on `drupal/classy`.

Seldaek:
I had a short night so take this with a grain of salt but after looking at this my impression is that the unlockPackage call is correct here, and it correctly removes the package which was loaded from the lock file, but then the `$this->markPackageNameForLoadingIfRequired($request, $replace);` call fails to load it as it is only required by the root package and nothing else.. So IMO the fix in this PR makes sense as it fixes that problem.

The root package not being checked there is probably simply an omission, and it seems reasonable to me that it would trigger a package load once the package is unlocked.

But anyway I got another fix which also works and is maybe more general as it handles it at the place where other unlocked packages get loaded, I'm not sure which is best to be honest.

See https://github.com/composer/composer/pull/11634

Toflar:
I see, I'd vote to leaving it in the `markPackageNameForLoadingIfRequired()` as this is where it logically belongs. Updated the PR here to reuse `isRootRequire()`.

Seldaek:
Yeah both are really fine by me and more or less equivalent I think. @naderman any also-sleep-deprived counter opinion? 😆

naderman:
Finally understood why this fix works on the original bug report. drupal/core removed the replace for drupal/classy in some 9.x version. The update of drupal/paragraphs with all deps means however that some older drupal/core version compatible with the selected versions (^9.2) get loaded which still have the replace. This in turn results in drupal/classy getting unlocked as it is now also part of "-W/with-all-dependencies" for the update. As the actually chosen newer drupal version does not replace drupal/classy, the package itself needs to be loaded. That doesn't currently happen because loading of root requirement packages on replace unlocking is missing. So this is right.

Going to still add a test specifically for this scenario outside of the multi repo replace unlocking one that had to be updated here by coincidence.

naderman:
```
There was 1 failure:

1) Composer\Test\DependencyResolver\PoolBuilderTest::testPoolBuilder with data set "load-replaced-root-package-if-replacer-dropped.test" ('load-replaced-root-package-if...d.test', 'Ensure that a root package ge...rement', array('replaced/pkg-1.0.0.0', 'replacer/pkg-1.0.0.0', 'replacer/pkg-1.1.0.0', 'root/dep-1.2.0.0'), array('replaced/pkg-1.0.0.0', 'replacer/pkg-1.0.0.0', 'replacer/pkg-1.1.0.0', 'root/dep-1.2.0.0'), array(), array(array('*', '1.0.0'), array(array('root/dep', '1.1.0', array('1.*')), array('replacer/pkg', '1.1.0'), array('replaced/pkg', '1.0.0')), array('root/dep'), true), array(array(array('root/dep', '1.2.0', array('1.*')), array('replacer/pkg', '1.0.0', array('1.0.0')), array('replacer/pkg', '1.1.0'), array('replaced/pkg', '1.0.0'))), array())
Unoptimized pool does not match expected package set
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
-    0 => 'replaced/pkg-1.0.0.0'
-    1 => 'replacer/pkg-1.0.0.0'
-    2 => 'replacer/pkg-1.1.0.0'
-    3 => 'root/dep-1.2.0.0'
+    0 => 'replacer/pkg-1.0.0.0'
+    1 => 'replacer/pkg-1.1.0.0'
+    2 => 'root/dep-1.2.0.0'
 )

/home/naderman/projects/composer/composer/tests/Composer/Test/DependencyResolver/PoolBuilderTest.php:142
```

Worked out a test reproducing this which fails on main. Added here.

Toflar:
Makes sense! Thanks for the test - funny we had one already (`tests/Composer/Test/DependencyResolver/Fixtures/poolbuilder/load-replaced-package-if-replacer-dropped.test`)  - just didn't cover root requirements being replaced :) Glad we have another one now! 🥳

naderman (approved):

T2L:
@Toflar @naderman @Seldaek thanks for such a quick turnaround! You are awesome

Issue #10099, opened 2021-09-13
Failed to extract package with warning:  stripped absolute path spec from 


jnaklaas:
When requiring my latest composer packages (built with satis, see [this issue](https://github.com/composer/satis/issues/652)), extracting fails. The package is unzipped, but all files are inside a nested subdirectory, instead of in the root of the package directory:
`vendor/mypackage/var/folders/21/gnrfrmtx7z72qv4j9cc7psvc0000gn/T/composer_archive613f6e6d68385/`

So the package is not properly installed and not available in the project. When I require an older version of the package, it unzips and installs as expected: `composer require myvendor/mypackage:1.1.6`.

My `composer.json`:

```json
{
  "name": "test/test",
  "type": "project",
  "repositories": [
    {
      "type": "composer",
      "url": "https://packages.myvendor.tld"
    }
  ],
  "require": {
    "php": ">=7.1",
    "myvendor/mypackage": "^1.1"
  }
}
```

Output of `composer diagnose`:

```
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.1-dev+4bcd860b6574fd71553019296965613cbaf95161
PHP version: 7.4.10
PHP binary path: /usr/local/Cellar/php/7.4.10/bin/php
OpenSSL version: OpenSSL 1.1.1g  21 Apr 2020
cURL version: 7.74.0 libz 1.2.11 ssl (SecureTransport) Ope
```

When I run this command: 

```sh
composer install
# or
composer require myvendor/mypackage
```

I get the following output: 

```
No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Loading composer repositories with package information
Updating dependencies
Lock file operations: 2 installs, 0 updates, 0 removals
  - Locking jjgrainger/posttypes (2.1)
  - Locking myvendor/mypackage (1.1.10)
Writing lock file
Installing dependencies from lock file (including require-dev)
[...]

PR #11636, opened 2023-09-12
Get realpath for `ZipArchive`

jasonmccreary:
I came across a few open issues, notably #10099, and the fix commented by @Seldaek resolved the issue. However, it has yet to be PR'd since Satis is now replaced by [Private Packagist](https://packagist.com/).

Nonetheless, here is the PR to resolve the issue where the archive paths are incorrect due to `ZipArchive::archive` not using the `realpath()`. Notably on MacOS where `sys_get_temp_dir()` returns a symlinked path. This change is inline with `PharArchive::archive` as it uses `realpath()` [already](https://github.com/composer/composer/blob/main/src/Composer/Package/Archiver/PharArchiver.php#L41).

There are other instances in the codebase which use `sys_get_temp_dir()`. However, since they are only used for storage, and not pathing information, I don't believe they require changes.

### DIFF
--- a/src/Composer/Package/Archiver/ZipArchiver.php
+++ b/src/Composer/Package/Archiver/ZipArchiver.php
@@ -31,7 +31,7 @@ class ZipArchiver implements ArchiverInterface
     public function archive(string $sources, string $target, string $format, array $excludes = [], bool $ignoreFilters = false): string
     {
         $fs = new Filesystem();
-        $sources = $fs->normalizePath($sources);
+        $sources = $fs->normalizePath(realpath($sources));
 
         $zip = new ZipArchive();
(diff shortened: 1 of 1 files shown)


jasonmccreary:
Not sure the appropriate fix here for PHPStan as checking for `false` requires some kind of action, e.g. throwing an exception or using `$sources` as is.

stof:
I suggest using `realpath($sources) ?: $sources`. If the resolution of the realpath fails, we should keep using the original one.

jasonmccreary:
Haha, looks like PHPStan doesn't like short ternary.

Seldaek:
Thanks

PR #11639, opened 2023-09-13
Add audit.abandoned warnings for abandoned packages

### LINKED ISSUE #11623: Audit should (have an option to) fail on abandoned packages
While implementing some security workflows for a client I noticed that `composer audit` does not in any way care about abandoned packages. Purists would argue that abandoned packages are by definition insecure.

I would argue that `composer audit` should by default list abandoned packages in use and fail on their presence, or at the very least support this via a flag.

My actual underlying problem is trying to fail a CI workflow when any security issues are found, and right now I don't see any way to do so for abandoned packages. `composer outdated` supports the `--strict` flag, but then doesn't support any kind of filtering on abandoned packages. As argued in this issue, I think `audit` is a better place for this anyway.

Seldaek:
Fixes #11623

### DIFF
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -15,8 +15,10 @@ namespace Composer\Advisory;
 use Composer\IO\ConsoleIO;
 use Composer\IO\IOInterface;
 use Composer\Json\JsonFile;
+use Composer\Package\CompletePackageInterface;
 use Composer\Package\PackageInterface;
 use Composer\Repository\RepositorySet;
+use Composer\Util\PackageInfo;
 use InvalidArgumentException;
 use Symfony\Component\Console\Formatter\OutputFormatter;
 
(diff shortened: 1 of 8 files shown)


src/Composer/Advisory/Auditor.php
```diff
@@ -40,17 +42,26 @@ class Auditor
...
      * @param string[] $ignoreList List of advisory IDs, remote IDs or CVE IDs that reported but not listed as vulnerabilities.
+     * @param self::ABANDONED_* $abandoned
      *
      * @return int Amount of packages with vulnerabilities found
      * @throws InvalidArgumentException If no packages are passed in
      */
-    public function audit(IOInterface $io, RepositorySet $repoSet, array $packages, string $format, bool $warningOnly = true, array $ignoreList = []): int
+    public function audit(IOInterface $io, RepositorySet $repoSet, array $packages, string $format, bool $warningOnly = true, array $ignoreList = [], string $abandoned = self::ABANDONED_REPORT): int
     {
+        if ($abandoned === 'default' && $format !== self::FORMAT_SUMMARY) {
```
stof, line 61: `default` is not one of the values supported in the method signature for static analysis.

curry684: I would define an extra constant `ABANDONED_DEFAULT = 'default` for now and switch it to `ABANDONED_DEFAULT = self::ABANDONED_FAIL` in 2.7.

src/Composer/Config.php
```diff
@@ -37,7 +38,7 @@ class Config
         'allow-plugins' => [],
         'use-parent-dir' => 'prompt',
         'preferred-install' => 'dist',
-        'audit' => ['ignore' => []],
+        'audit' => ['ignore' => [], 'abandoned' => 'default'], // TODO in 2.7 switch to ABANDONED_FAIL
```
curry684, line 41: Referencing `Auditor::ABANDONED_DEFAULT` here automatically future-proofs the code.

curry684:
Awesome quick work man 👍

src/Composer/Advisory/Auditor.php
Seldaek: The point is default as a value will be gone in 2.7 so i rather keep it low profile. IMO it's fine as is

src/Composer/Advisory/Auditor.php
```diff
@@ -40,17 +42,26 @@ class Auditor
...
+            $io->writeError('<warning>The new audit.abandoned setting (currently defaulting to "report" will default to "fail" in Composer 2.7, make sure to set it to "report" or "ignore" explicitly by then if you do not want this.</warning>');
```
naderman, line 62: Should this warning really always be output or only if an actual warning is output? Seems a bit much to force everyone to set this to a value to make the warning go away?

naderman (approved):

src/Composer/Advisory/Auditor.php
Seldaek: Well if you run audit in CI you can also just ignore this warning.. my thinking was more that if it's hidden then you won't notice it unless you have abandoned packages. Then people will come complaining when 2.7 hits and they have an abandoned package in six months..

curry684: Is this really an issue? I highly doubt `audit` is frequently used outside CI context, so it's not really a 'daily annoyance' or anything, and those that do consciously run `audit` will be the kind of user happy with the verbose explanation.

PR #11663, opened 2023-09-26
feat: improve Composer's output reproducibility

drupol:
This PR:

- [x] Improve reproducibility of Composer's output (_the `vendor` directory_). Currently, only the file content is guaranteed to be reproducible, the file time metadata are not (yet?) reproducible.
- [x] Should fix https://github.com/composer/composer/issues/7049

Reproduce this at home, inside the `composer` source directory:

image

### DIFF
--- a/doc/01-basic-usage.md
+++ b/doc/01-basic-usage.md
@@ -152,6 +152,13 @@ a Composer `install` to make sure the vendor directory is up in sync with your
 php composer.phar install
 ```
 
+Composer enables reproducible builds by default. This means that running the
+same command multiple times will produce a `vendor/` directory containing files
+that are identical (*except their timestamps*), including the autoloader files.
+It is especially beneficial for environments that require strict
+verification processes, as well as for Linux distributions aiming to package PHP
+applications in a secure and predictable manner.
+
 ## Updating dependencies to their latest versions
 
 As mentioned above, the `composer.lock` file prevents you from automatically getting
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -33,6 +33,7 @@ use Composer\Util\Platform;
 use Composer\Script\ScriptEvents;
 use Composer\Util\PackageSorter;
 use Composer\Json\JsonFile;
+use Composer\Package\Locker;
 
 /**
  * @author Igor Wiedler <igor@wiedler.ch>
@@ -172,7 +173,7 @@ class AutoloadGenerator
      * @throws \Seld\JsonLint\ParsingException
      * @throws \RuntimeException
      */
-    public function dump(Config $config, InstalledRepositoryInterface $localRepo, RootPackageInterface $rootPackage, InstallationManager $installationManager, string $targetDir, bool $scanPsrPackages = false, ?string $suffix = null)
+    public function dump(Config $config, InstalledRepositoryInterface $localRepo, RootPackageInterface $rootPackage, InstallationManager $installationManager, Locker $locker, string $targetDir, bool $scanPsrPackages = false, ?string $suffix = null)
     {
         if ($this->classMapAuthoritative) {
             // Force scanPsrPackages when classmap is authoritative
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -406,9 +406,8 @@ EOF;
                 }
             }
 
-            // generate one if we still haven't got a suffix
             if (null === $suffix) {
-                $suffix = md5(uniqid('', true));
+                $suffix = $locker->isLocked() ? $locker->getLockData()['content-hash'] : md5(uniqid('', true));
             }
         }
 
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -349,7 +349,17 @@ class Installer
             $this->autoloadGenerator->setApcu($this->apcuAutoloader, $this->apcuAutoloaderPrefix);
             $this->autoloadGenerator->setRunScripts($this->runScripts);
             $this->autoloadGenerator->setPlatformRequirementFilter($this->platformRequirementFilter);
-            $this->autoloadGenerator->dump($this->config, $localRepo, $this->package, $this->installationManager, 'composer', $this->optimizeAutoloader);
+            $this
+                ->autoloadGenerator
+                ->dump(
+                    $this->config,
+                    $localRepo,
+                    $this->package,
+                    $this->installationManager,
+                    $this->locker,
+                    'composer',
+                    $this->optimizeAutoloader
+                );
         }
 
         if ($this->install && $this->executeOperations) {
--- a/tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
+++ b/tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
@@ -15,7 +15,6 @@ namespace Composer\Test\Autoload;
 use Composer\Autoload\AutoloadGenerator;
 use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilterFactory;
 use Composer\IO\BufferIO;
-use Composer\IO\IOInterface;
 use Composer\Package\CompletePackage;
 use Composer\Package\Link;
 use Composer\Package\Version\VersionParser;
@@ -31,6 +30,7 @@ use Composer\Repository\InstalledRepositoryInterface;
 use Composer\Installer\InstallationManager;
 use Composer\Config;
 use Composer\EventDispatcher\EventDispatcher;
+use Composer\Package\Locker;
 use Composer\Util\Platform;
 use PHPUnit\Framework\MockObject\MockObject;
 
@@ -80,6 +80,11 @@ class AutoloadGeneratorTest extends TestCase
      * @var BufferIO
      */
     private $io;
+
+    /**
+     * @var Locker
+     */
+    private $locker;
 
     /**
      * @var EventDispatcher&MockObject
@@ -159,6 +164,10 @@ class AutoloadGeneratorTest extends TestCase
             ->getMock();
 
         $this->generator = new AutoloadGenerator($this->eventDispatcher, $this->io);
+
+        $this->locker = $this->getMockBuilder('Composer\Package\Locker')
+            ->disableOriginalConstructor()
+            ->getMock();
     }
 
     protected function tearDown(): void
@@ -208,7 +217,7 @@ class AutoloadGeneratorTest extends TestCase
         $this->fs->ensureDirectoryExists($this->workingDir.'/composersrc');
         file_put_contents($this->workingDir.'/composersrc/foo.php', '<?php class ClassMapFoo {}');
 
-        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', true, '_1');
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, $this->locker, 'composer', true, '_1');
 
         // Assert that autoload_namespaces.php was correctly generated.
         $this->assertAutoloadFiles('main', $this->vendorDir.'/composer');
@@ -248,7 +257,7 @@ class AutoloadGeneratorTest extends TestCase
 
         // generate autoload files with the dev mode set to true
         $this->generator->setDevMode(true);
-        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', true, '_1');
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, $this->locker, 'composer', true, '_1');
 
         // check standard autoload
         $this->assertAutoloadFiles('main5', $this->vendorDir.'/composer');
@@ -281,7 +290,7 @@ class AutoloadGeneratorTest extends TestCase
         $this->fs->ensureDirectoryExists($this->workingDir.'/devfiles');
         file_put_contents($this->workingDir.'/devfiles/foo.php', '<?php function foo() { echo "foo"; }');
 
-        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', true, '_1');
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, $this->locker, 'composer', true, '_1');
 
         // check standard autoload
         $this->assertAutoloadFiles('main4', $this->vendorDir.'/composer');
@@ -316,7 +325,7 @@ class AutoloadGeneratorTest extends TestCase
         $this->fs->ensureDirectoryExists($this->vendorDir.'/composersrc');
         file_put_contents($this->vendorDir.'/composersrc/foo.php', '<?php class ClassMapFoo {}');
 
-        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', true, '_2');
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, $this->locker, 'composer', true, '_2');
         $this->assertAutoloadFiles('main3', $this->vendorDir.'/composer');
         $this->assertAutoloadFiles('psr4_3', $this->vendorDir.'/composer', 'psr4');
         $this->assertAutoloadFiles('classmap3', $this->vendorDir.'/composer', 'classmap');
@@ -345,7 +354,7 @@ class AutoloadGeneratorTest extends TestCase
 
         $this->fs->ensureDirectoryExists($this->workingDir.'/composersrc');
         file_put_contents($this->workingDir.'/composersrc/foo.php', '<?php class ClassMapFoo {}');
-        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', false, '_3');
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, $this->locker, 'composer', false, '_3');
         $this->assertAutoloadFiles('main2', $this->vendorDir.'/composer');
         $this->assertAutoloadFiles('psr4_2', $this->vendorDir.'/composer', 'psr4');
         $this->assertAutoloadFiles('classmap2', $this->vendorDir.'/composer', 'classmap');
@@ -374,7 +383,7 @@ class AutoloadGeneratorTest extends TestCase
         file_put_contents($this->workingDir.'/foo.php', '<?php class FilesFoo {}');
         file_put_contents($this->workingDir.'/bar.php', '<?php class FilesBar {}');
 
-        $this->generator->dump($this->config, $this->repository, $package, $this->im, 'composer', false, 'TargetDir');
+        $this->generator->dump($this->config, $this->repository, $package, $this->im, $this->locker, 'composer', false, 'TargetDir');
         $this->assertFileContentEquals(__DIR__.'/Fixtures/autoload_target_dir.php', $this->vendorDir.'/autoload.php');
         $this->assertFileContentEquals(__DIR__.'/Fixtures/autoload_real_target_dir.php', $this->vendorDir.'/composer/autoload_real.php');
         $this->assertFileContentEquals(__DIR__.'/Fixtures/autoload_static_target_dir.php', $this->vendorDir.'/composer/autoload_static.php');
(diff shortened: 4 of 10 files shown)


Seldaek:
Sorry but I am really confused as to what this tries to achieve? :) It seems to `touch()` all vendor files with a `0` timestamp. What's the point? And why would you need this flag vs doing it manually with `find vendor -type f -exec touch {} 0` or smth if it's a special need it sounds ok to me? 

Also not sure how it relates to #7049 at all.. But then again there are so many posts talking about different things in that thread, maybe I missed your point.

drupol:
Hi @Seldaek,

I apologize for any confusion. This PR is part of an initiative to achieve reproducible builds with `composer`. The aim is to ensure that `composer` can create a `vendor/` directory that is bit-for-bit identical across multiple runs of the same command. 

By zeroing out the timestamps of all `vendor/` files, we eliminate one variable that could make builds non-deterministic. Integrating this feature into `composer` itself would streamline the process and make it more accessible for users focused on reproducibility (_debian, NixOS, Arch, etc etc_).

This is totally related to #7049 from @lstrojny since the goal of that PR was precisely what I'm doing here. Ensuring reproducible `vendor/` directories aligns with this goal.

For more context on why reproducible builds are important, I'd refer you to this page: https://reproducible-builds.org/docs/buy-in/

I've added a screenshot demonstrating the default non-deterministic behavior of Composer. Running `composer install` twice resulted in two different directory hashes. However, when using the new `--deterministic` flag, the directory hashes were identical in multiple runs.

I believe this feature would be a significant step in making PHP applications more secure and predictable through Composer. Additionally, this would greatly benefit Linux distributions that aim to package PHP applications in a secure and predictable manner.

Seldaek:
Yeah ok, I have already fixed a ton of issues with reproducibility so you don't need to convince me, but this one here seems like the nuclear option to me. 

We could perhaps by default copy the filemtime of the composer.lock to the files Composer creates in vendor? That would be more sensible IMO than 0, and the other files in vendor should not need modifying as these have filemtime controlled by their own package source, and zeroing these feels really wrong to me.

On the other hand, we also touch the vendor dir itself after dependencies get modified (https://github.com/composer/composer/blob/main/src/Composer/Installer.php#L801), but this probably could be changed to match the lock file's mtime? I think it would still achieve the original goal of https://github.com/composer/composer/issues/2764

drupol:
Yes, this is also an option, as long as this is predictable, it's fine.

I'll modify the PR accordingly right now and report back.

Seldaek:
Ok so could we do this **without** --deterministic flag? I'd really rather have these things just work out of the box, and avoid special behavior where possible.

Also could the `touch` be done directly in the AutoloadGenerator instead of twice in Installer & DumpAutoloadCommand?

Regarding the autoload suffix, I think we could always use the lock file hash if there is one, that's probably unique enough. If there is no lock file available then just pass null and have it default to generating one or reusing existing one. You anyway won't have anything deterministic without lock file so it shouldn't matter in that case.

Seldaek:
> I'd really rather have these things just work out of the box, and avoid special behavior where possible.

This is assuming you don't need to carpet bomb the whole vendor dir with `touch`, because this is IMO unacceptable behavior to modify vendor files.

drupol:
I definitely agree that this should be the default behavior.

I reworked the PR from the beginning and this is now the default behavior:

image

I still need to fix the tests and add new tests.

src/Composer/Installer.php
```diff
@@ -418,6 +431,16 @@ public function run(): int
             }
         }
 
+        if ($this->locker->isLocked()) {
```
stof, line 434: this should depend on `$this->executeOperations` IMO. A `--dry-run` should not do it.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -456,6 +456,13 @@ public static function autoload(\$class)
             ]);
         }
 
+        if ($locker->isLocked()) {
+            $mtime = $locker->getMtime();
+            foreach ((array) glob(sprintf('%s/%s', $targetDir, 'autoload_*.php')) as $file) {
```
Seldaek, line 461: This won't cover autoload.php AFAICT, and installed.json/installed.php/InstalledVersions.php possibly also need addressing somehow?

Seldaek: Also vendor/autoload.php

src/Composer/Installer.php
```diff
@@ -418,6 +432,16 @@ public function run(): int
...
 
+        if ($this->executeOperations && $this->locker->isLocked()) {
+            $mtime = $this->locker->getMtime();
+            $rii = new RecursiveIteratorIterator(new RecursiveDirectoryIterator($this->config->get('vendor-dir')));
+
+            /** @var SplFileInfo $file */
+            foreach ($rii as $file) {
+                @touch($file->getRealPath(), $mtime, $mtime);
+            }
+        }
```
Seldaek, line 443: IMO this should just change the `touch` at line 801 to use the mtime from lock file. It should only touch the vendorDir and not its contents. This is probably the reason why you didn't notice the other files which weren't taken care of, as this one still touches everything.

src/Composer/Autoload/AutoloadGenerator.php
drupol: Thanks, this is now done.

src/Composer/Installer.php
drupol: I understand, we should only "touch" (_pun intented!_) the files that composer creates, nothing else.

When removing that part, this is obviously not reproducable.

When analyzing the differences with `diffoscope`, I get this list:

<details>
❯ diffoscope example-project/vendor.left example-project/vendor.right
--- example-project/vendor.left
+++ example-project/vendor.right
├── stat {}
│ @@ -1,7 +1,7 @@
│  
│    Size: 4096         Blocks: 8          IO Block: 4096   directory
│  Device: 259,2        Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│  
│ -Modify: 2023-09-27 12:45:30.576449945 +0000
│ +Modify: 2023-09-27 12:46:43.204202660 +0000
│   --- example-project/vendor.left/bin
├── +++ example-project/vendor.right/bin
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:43.004203341 +0000
│ │ -Modify: 2023-09-27 12:45:30.424450462 +0000
│ │   --- example-project/vendor.left/bin/jsonlint
│ ├── +++ example-project/vendor.right/bin/jsonlint
│ │ ├── stat {}
│ │ │ @@ -1,7 +1,7 @@
│ │ │  
│ │ │    Size: 3336             Blocks: 8          IO Block: 4096   regular file
│ │ │  Device: 259,2    Access: (0755/-rwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │ │  
│ │ │ +Modify: 2023-09-27 12:46:42.880203762 +0000
│ │ │ -Modify: 2023-09-27 12:45:30.308450859 +0000
│ │   --- example-project/vendor.left/bin/phpstan
│ ├── +++ example-project/vendor.right/bin/phpstan
│ │ ├── stat {}
│ │ │ @@ -1,7 +1,7 @@
│ │ │  
│ │ │    Size: 3327             Blocks: 8          IO Block: 4096   regular file
│ │ │  Device: 259,2    Access: (0755/-rwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │ │  
│ │ │ +Modify: 2023-09-27 12:46:42.988203396 +0000
│ │ │ -Modify: 2023-09-27 12:45:30.412450504 +0000
│ │   --- example-project/vendor.left/bin/phpstan.phar
│ ├── +++ example-project/vendor.right/bin/phpstan.phar
│ │ ├── stat {}
│ │ │ @@ -1,7 +1,7 @@
│ │ │  
│ │ │    Size: 3342             Blocks: 8          IO Block: 4096   regular file
│ │ │  Device: 259,2    Access: (0755/-rwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │ │  
│ │ │ +Modify: 2023-09-27 12:46:43.004203341 +0000
│ │ │ -Modify: 2023-09-27 12:45:30.424450462 +0000
│ │   --- example-project/vendor.left/bin/simple-phpunit
│ ├── +++ example-project/vendor.right/bin/simple-phpunit
│ │ ├── stat {}
│ │ │ @@ -1,7 +1,7 @@
│ │ │  
│ │ │    Size: 3381             Blocks: 8          IO Block: 4096   regular file
│ │ │  Device: 259,2    Access: (0755/-rwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │ │  
│ │ │ +Modify: 2023-09-27 12:46:42.988203396 +0000
│ │ │ -Modify: 2023-09-27 12:45:30.404450532 +0000
│ │   --- example-project/vendor.left/bin/validate-json
│ ├── +++ example-project/vendor.right/bin/validate-json
│ │ ├── stat {}
│ │ │ @@ -1,7 +1,7 @@
│ │ │  
│ │ │    Size: 3387             Blocks: 8          IO Block: 4096   regular file
│ │ │  Device: 259,2    Access: (0755/-rwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │ │  
│ │ │ +Modify: 2023-09-27 12:46:42.876203777 +0000
│ │ │ -Modify: 2023-09-27 12:45:30.300450886 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.424450462 +0000
│ │ +Modify: 2023-09-27 12:46:43.004203341 +0000
│   --- example-project/vendor.left/composer
├── +++ example-project/vendor.right/composer
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:43.204202660 +0000
│ │ -Modify: 2023-09-27 12:45:30.576449945 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.576449945 +0000
│ │ +Modify: 2023-09-27 12:46:43.204202660 +0000
│   --- example-project/vendor.left/justinrainbow
├── +++ example-project/vendor.right/justinrainbow
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:42.840203898 +0000
│ │ -Modify: 2023-09-27 12:45:30.268450995 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.268450995 +0000
│ │ +Modify: 2023-09-27 12:46:42.840203898 +0000
│   --- example-project/vendor.left/phpstan
├── +++ example-project/vendor.right/phpstan
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:42.980203423 +0000
│ │ -Modify: 2023-09-27 12:45:30.408450517 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.408450517 +0000
│ │ +Modify: 2023-09-27 12:46:42.980203423 +0000
│   --- example-project/vendor.left/psr
├── +++ example-project/vendor.right/psr
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:42.920203626 +0000
│ │ -Modify: 2023-09-27 12:45:30.348450722 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.348450722 +0000
│ │ +Modify: 2023-09-27 12:46:42.920203626 +0000
│   --- example-project/vendor.left/react
├── +++ example-project/vendor.right/react
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:42.856203844 +0000
│ │ -Modify: 2023-09-27 12:45:30.284450940 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.284450940 +0000
│ │ +Modify: 2023-09-27 12:46:42.856203844 +0000
│   --- example-project/vendor.left/seld
├── +++ example-project/vendor.right/seld
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:42.904203681 +0000
│ │ -Modify: 2023-09-27 12:45:30.332450777 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.332450777 +0000
│ │ +Modify: 2023-09-27 12:46:42.904203681 +0000
│   --- example-project/vendor.left/symfony
├── +++ example-project/vendor.right/symfony
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ +Modify: 2023-09-27 12:46:42.972203450 +0000
│ │ -Modify: 2023-09-27 12:45:30.392450571 +0000
│ ├── stat {}
│ │ @@ -1,7 +1,7 @@
│ │  
│ │    Size: 4096       Blocks: 8          IO Block: 4096   directory
│ │  Device: 259,2      Access: (0755/drwxr-xr-x)  Uid: ( 1000/  devlin)   Gid: ( 1000/  devlin)
│ │  
│ │ -Modify: 2023-09-27 12:45:30.392450571 +0000
│ │ +Modify: 2023-09-27 12:46:42.972203450 +0000
❯
</details>

I'm going to work on this and see when these files and directories are created... and do the changes accordingly. Help is more than welcome :)

Seldaek: Yeah ok, of course every directory created for a package will have this same problem.. We're pretty good at binary reproducibility in terms of contents, but filemtime are such a pain to handle :/

And binary proxies too makes sense, just not sure how to handle these.. touching them every time to match lock file is possible but sounds messy if the files didn't change. I don't know if filemtime reproducibility is really something to strive for, what's the use case exactly? Is nix not happy with bytes on disk being the same?

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -456,6 +456,20 @@ public static function autoload(\$class)
             ]);
         }
 
+        if ($locker->isLocked()) {
+            $atime = $locker->getAtime();
+            $mtime = $locker->getMtime();
+            $generatedFiles = (array) array_merge(
+                (array) glob(sprintf('%s/*', $targetDir)),
```
stof, line 463: this should rather be done in places using `filePutContentsIfModified` and `safeCopy` instead of doing a glob here. This glob can still find extra files (for instance when you depend on a package from the `composer` namespace)

drupol: Thanks Stof... However, I think I will remove that commit and postpone it until the Composer is updated so we can have access to the `Locker` in Installers classes. (see my comment https://github.com/composer/composer/pull/11663#discussion_r1338610261)

src/Composer/Installer.php
drupol: Fair enough. This part is also doable through a plugin: https://github.com/loophp/composer-strip-nondeterminism (_quickly did that just for fun_).

tests/Composer/Test/Autoload/AutoloadGeneratorTest.php
```diff
@@ -177,6 +177,9 @@ protected function tearDown(): void
 
     public function testRootPackageAutoloading(): void
     {
+        $locker = $this->getMockBuilder('Composer\Package\Locker')
+            ->disableOriginalConstructor()
+            ->getMock();
```
Seldaek, line 182: It would be nicer to move this to setUp and then just use `$this->locker` in every test instead of duplicating these lines in every test.

doc/01-basic-usage.md
```diff
@@ -152,6 +152,15 @@ a Composer `install` to make sure the vendor directory is up in sync with your
 php composer.phar install
 ```
 
+Composer enables reproducible builds by default. This means that running the
+same command multiple times will produce a `vendor/` directory containing files
+that are identical (*except their timestamps*), including the autoloader files.
+This feature is now enabled out-of-the-box to enhance the security and
+predictability of your PHP applications by ensuring that the Composer output is
+deterministic. It is especially beneficial for environments that require strict
```
Seldaek, line 160: Let's keep it short, this is the basic usage page, and the majority of people don't care all that much about reproducibility ;)

```suggestion
It is especially beneficial for environments that require strict
```

Seldaek:
LGTM aside from the two minor comments, thanks for the quick turnaround here. Given the scope change I now think it's safe to move this to 2.6 too.

doc/01-basic-usage.md
drupol: > and the majority of people don't care all that much about reproducibility

image

Seldaek: Is 2024 the year of nix on the desktop? 😉

Seldaek:
Thanks

doc/01-basic-usage.md
drupol: Oooh you're teasing me, :D that's mean !

On the topic of Nix, I'd love to dive deeper into this conversation with you. How about we meet up for a chat over beer or tea at https://phpconference.com/web-development/leveraging-nix-php-ecosystem/ (if you plan to come) ?

As for the concept of reproducibility (_r13y_), it's important to note that its roots extend beyond just Nix. While Nix has certainly popularized the concept, it was originally pioneered by the Debian community as far back as the early 2000s. What Nix has done is make this idea of reproducibility more accessible to a wider audience.

Seldaek: I will not be at IPC, but I'll be at SymfonyCon which is in Brussels early December.

drupol: I will be there at both, hope to finally meet you in person!

drupol:
@Seldaek Do you think it would be possible to edit the Github changelog and add a line about this, so I can spread the word about it ?

Seldaek:
Sure, added to https://github.com/composer/composer/releases/tag/2.6.4 sorry :)

PR #11664, opened 2023-09-27
PR Request For Updating README.md

Mitesh2020:
Added : 
Benefits of using Composer, How to use Composer, Conclusion, How Composer can make you a more professional PHP developer.


Seldaek:
Thanks but I'd rather keep the info in the docs & getcomposer.org instead of duplicating things in the readme, which then is yet another place we need to keep up to date.

PR #11677, opened 2023-10-05
10796 Increase coverage of ShowCommand

greew:
I've added more tests for the `Composer\Command\ShowCommand` command to increase the test coverage.

Please add suggestions for changes :)


Seldaek:
Looks great thanks, maybe there is still more that could be done but that looks already pretty comprehensive to me at first glance.

PR #11678, opened 2023-10-05
Added tests

jakobvibe:
Added some test for InitCommand as requested in https://github.com/composer/composer/issues/10796

Simple testing, but added a test approach for testing protected functions through dummyclass extension


Seldaek:
I have nothing against the cleanup with the dummy class, but did I miss something or did you get rid of a bunch of tests?

Issue #11573, opened 2023-08-07
Error when trying to install laravel 


mrzM7md:
In CurlDownloader.php line 365:

  curl error 28 while downloading https://repo.packagist.org/packages.json:    
  SSL connection timeout 

my php 8, last version of composer, used xampp

abreel:
@Seldaek Can I look into this?

fredden:
@mrzM7md please can you supply the information requested in the issue template. This is the content that you deleted when creating this issue. For your reference, you can find a copy here: [https://github.com/composer/composer/blob/main/.github/ISSUE_TEMPLATE.md](https://github.com/composer/composer/blob/75ae504b2c8a7ee8533c69788c3bbdbf3740e610/.github/ISSUE_TEMPLATE.md?plain=1)

I imagine that the output of `composer diagnose` may be helpful here.

PR #11681, opened 2023-10-06
Update composer.json for solving  the issue #11573

tusharv01:
I have added the repositories new section and made few changes in the config section.


Seldaek:
Thanks for the PR but I have no idea what this is trying to fix. This repository definition is invalid and definitely does not belong there, and proxy definitions also do not belong here..

Issue #11683, opened 2023-10-10
Feature: display `libpq` version with `composer show -p` command


llaville:
Hello, 

This is not a bug, but will help to debug depending on current platform.
Recently, I've begun to start to implement PHP 8.3 support on my project https://github.com/llaville/php-compatinfo-db and when I update the `pgsql` extension reference, I've noticed that new `PGSQL_ERRORS_SQLSTATE`  constant added since PHP 8.3.0alpha1 (see https://github.com/php/php-src/commit/f31d253849716b585943abec4b7d38e89036f78b#diff-c5961ea04ab805b0059964fac68d6e159095a4b56aaaded0b1b941b8768c0f52) depends on libpq > 11.0

BTW, Composer 2.6 did not yet display version of libpq extension with the `composer show -p` command.

```
Running 2.6.5 (2023-10-06 10:11:52) with PHP 8.3.0RC3 on Linux / 5.15.90.1-microsoft-standard-WSL2
Reading ./composer.json (/shared/backups/bartlett/php-compatinfo-db/composer.json)
Loading config file ./composer.json (/shared/backups/bartlett/php-compatinfo-db/composer.json)
Checked CA file /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a file.
Checked directory /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a directory.
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/shared/backups/bartlett/php-compatinfo-db): 'git' 'branch' '-a' '--no-color' '--no-abbrev' '-v'
Failed to initialize global composer: Composer could not find the config file: /home/devilbox/.composer/composer.json

Reading /shared/backups/bartlett/php-compatinfo-db/vendor/composer/installed.json
Loading plugin cweagans\Composer\Patches (from cweagans/composer-patches)
composer                2.6.5     Composer package
composer-plugin-api     2.6.0     The Composer Plugin API
composer-runtime-api    2.2.2     The Composer Runtime API
ext-amqp                2.1.0     The amqp PHP extension
ext-apcu                5.1.22    The apcu PHP extension
ext-ast                 1.1.0     The ast PHP extension
ext-bcmath              8.3.0RC3  The bcmath PHP extension
ext-bz2                 8.3.0RC3  The bz2 PHP extension
ext-calendar            8.3.0RC3  The calendar PHP extension
ext-ctype               8.3.0RC3  The ctype PHP extension
[...]

PR #11684, opened 2023-10-11
Use global constant if available for libpq version

fredden:
As suggested in https://github.com/composer/composer/issues/11683#issuecomment-1757311564, this pull request adds a performance optimisation for detecting the `libpq` version.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -449,6 +449,12 @@ class PlatformRepository extends ArrayRepository
                     break;
 
                 case 'pgsql':
+                    if ($this->runtime->hasConstant('PGSQL_LIBPQ_VERSION')) {
+                        $this->addLibrary('pgsql-libpq', $this->runtime->getConstant('PGSQL_LIBPQ_VERSION'), 'libpq for pgsql');
+                        break;
+                    }
+                // intentional fall-through to next case...
+
                 case 'pdo_pgsql':
                     $info = $this->runtime->getExtensionInfo($name);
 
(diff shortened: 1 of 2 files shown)


src/Composer/Repository/PlatformRepository.php
```diff
@@ -450,10 +450,14 @@ protected function initialize(): void
 
                 case 'pgsql':
                 case 'pdo_pgsql':
-                    $info = $this->runtime->getExtensionInfo($name);
+                    if (defined('PGSQL_LIBPQ_VERSION')) {
+                        $this->addLibrary($name.'-libpq', PGSQL_LIBPQ_VERSION, 'libpq for '.$name);
```
stof, line 454: What if pdo_pgsql is compiled against a different libpq version than the pgsql extension ? This constant is part of the pgsql extension.

src/Composer/Repository/PlatformRepository.php
```diff
@@ -450,10 +450,14 @@ protected function initialize(): void
 
                 case 'pgsql':
                 case 'pdo_pgsql':
-                    $info = $this->runtime->getExtensionInfo($name);
+                    if (defined('PGSQL_LIBPQ_VERSION')) {
```
stof, line 453: this should use `$this->runtime->hasConstant` and `$this->runtime->getConstant` to allow testing this code.

src/Composer/Repository/PlatformRepository.php
fredden: Excellent point; I'll move this so it only affects the 'pgsql' extension and not 'pdo_pgsql'.

Seldaek:
Thanks

PR #11689, opened 2023-10-16
Fix Git Driver to use supported Git VCS driver URL

### LINKED ISSUE #11409: Environment variable not work with local repo

ktomk:
Otherwise the URL may not be supported since 3bb191a46 (Add support for env vars and ~ (for HOME) in repo paths for vcs and artifact repositories, fixes #11409 (#11453), 2023-05-07)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-main": "2.6-dev"
+            "dev-main": "2.7-dev"
         },
         "phpstan": {
             "includes": [
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "4bceaf933dcf6bc05808134e78d21496",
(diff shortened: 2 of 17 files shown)


Seldaek:
Thanks, the fix makes sense but I do wonder if it wouldn't be better to do this in the constructor directly? To avoid this invalid URL causing problems somewhere else later?

ktomk:
Yes indeed, moved it into the constructor @Seldaek

Seldaek:
Merged in 2.6 as 03085c818 - thanks

Issue #11690, opened 2023-10-17
The 'get another view on the problem' message shouldn't suggest impossible commands


joachim-n:
My `composer.json`:

```json
{
    "name": "drupal/recommended-project",
    "description": "Project template for Drupal 9 projects with a relocated document root",
    "type": "project",
    "license": "GPL-2.0-or-later",
    "homepage": "https://www.drupal.org/project/drupal",
    "support": {
        "docs": "https://www.drupal.org/docs/user_guide/en/index.html",
        "chat": "https://www.drupal.org/node/314178"
    },
    "repositories": [
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    ],
    "require": {
        "composer/installers": "^1.9",
        "dealerdirect/phpcodesniffer-composer-installer": "*",
        "drupal/admin_toolbar": "^3.2",
        "drupal/coder": "^8.3",
        "drupal/computed_field_plugin": "^1.3",
        "drupal/core-composer-scaffold": "^9.4",
        "drupal/core-dev": "^9.4",
        "drupal/core-recommended": "^9.5",
        "drupal/ctools": "^4.0",
        "drupal/devel": "^5.0",
        "drupal/devel_entity_updates": "^4.1",
        "drupal/filelog": "^2.1",
        "drupal/flag": "^4.0@beta",
        "drupal/module_builder": "^3.24",
        "drupal/token": "^1.11",
        "drupal/ultimate_cron": "^2.0@alpha",
        "drush/drush": "^11.2",
        "edemeijer/serialize-debugger": "^0.1.4",
        "kint-php/kint": "^4.2",
        "phpspec/prophecy-phpunit": "^2"
    },
    "conflict": {
        "drupal/drupal": "*"
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "config": {
        "allow-plugins": {
            "composer/installers": true,
            "drupal/core-composer-scaffold": true,
            "dealerdirect/phpcodesniffer-composer-installer": true
        },
        "sort-packages": true
    },
    "extra": {
        "drupal-scaffold": {
            "locations": {
                "web-root": "web/"
            }
        },
        "installer-paths": {
            "web/core": [
                "type:drupal-core"
            ],
            "web/libraries/{$name}": [
                "type:drupal-library"
            ],
            "web/modules/contrib/{$name}": [
                "type:drupal-module"
            ],
            "web/profiles/contrib/{$name}": [
                "type:drupal-profile"
            ],
[...]

PR #11691, opened 2023-10-17
Suggest running 'require' not 'update' if package is a root requirement

fredden:
This was inspired by and might fix https://github.com/composer/composer/issues/11690


Seldaek:
Thanks

PR #11692, opened 2023-10-17
Display error instead of throwing exception when unable to update with temporary constraint

fredden:
This was inspired by and might fix https://github.com/composer/composer/issues/11690


Seldaek:
Thanks

jasonbertoldo:
myqr_1704107881568

PR #11694, opened 2023-10-23
Bump wildcard constraint

fredden:
I recently ran `composer bump` on a project which used a wildcard constraint (`*`) for a particular package.

The help text for the `bump` command reads:

> Increases the lower limit of your composer.json requirements to the currently installed versions

However the lower-limit for the wildcard constraint (`*`) was not changed. This pull request adds this functionality.

While I do not believe that unbound version constraints should be used generally, I recognise that others have different opinions and that there are valid use-cases for such constraints.


Seldaek:
Hm interesting.. I kinda hate it but also it makes sense strictly speaking. So thanks? ;)

PR #11724, opened 2023-11-21
Adds a test for UpdateCommand

theoboldalex:
Adds a single test case for the `UpdateCommand`.

This is a start for covering the parts of the commands which need to be tested in interactive mode.


Seldaek:
Thanks

PR #11726, opened 2023-11-23
GH Actions: update the CI workflow for the release of PHP 8.3

jrfnl:
Suggested change for the CI workflow what with the PHP 8.3 release expected later today.

PHP 8.4 has been added as experimental.
And the "high" PHP builds for the variations have been updated to use PHP 8.3.


Seldaek:
Thanks

PR #11734, opened 2023-11-28
Exposing GitLab's project metadata

madbob:
This permits to directly access project's metadata from the GitLab driver, exactly as it already happens with the GitHub's one.

The main purpose of this change is to be then leveraged in Packagist to grab addictional informations about stars, forks, open issues and more, to be displayed in the web interface or be used for internals (e.g. "popularity" evaluation based on project's stars).

cfr. https://github.com/composer/packagist/blob/main/src/Package/Updater.php#L564


Seldaek:
Thanks

PR #11762, opened 2023-12-19
Add release date to show command, and also release date for latest package in --latest/outdated mode

### LINKED ISSUE #11737: Add the time a package was released in `composer show` command
It would be helpful to output the release time next to the package in `composer show` (or  only when running with `--outdated` option) command to see how old the current packages are. Having this info condensated in the package list would help to draw conclusions whether some packages need upgrade.

I would use the relative time like `1 month ago`(or make it configurable via composer.json `extra` key?) - it better affects humans. I would create a new option to sort desc/asc (from oldest to latest and vice versa) based on release time.

We could also add this info to the output of `composer show <package>`.

We already have info about release time in composer.lock.

Seldaek:
Fixes #11737

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -556,6 +556,7 @@ php composer.phar show monolog/monolog 1.0.2
 * **--major-only (-M):** Use with --latest or --outdated. Only shows packages that have major SemVer-compatible updates.
 * **--minor-only (-m):** Use with --latest or --outdated. Only shows packages that have minor SemVer-compatible updates.
 * **--patch-only:** Use with --latest or --outdated. Only shows packages that have patch-level SemVer-compatible updates.
+* **--sort-by-age (-A):** Sort by and display current release age. Use with the --latest or --outdated option.
 * **--direct (-D):** Restricts the list of packages to your direct dependencies.
 * **--strict:** Return a non-zero exit code when there are outdated packages.
 * **--format (-f):** Lets you pick between text (default) or json output format.
@@ -589,6 +590,7 @@ The color coding is as such:
 * **--major-only (-M):** Only shows packages that have major SemVer-compatible updates.
 * **--minor-only (-m):** Only shows packages that have minor SemVer-compatible updates.
 * **--patch-only (-p):** Only shows packages that have patch-level SemVer-compatible updates.
+* **--sort-by-age (-A):** Sort by and display current release age.
 * **--format (-f):** Lets you pick between text (default) or json output format.
 * **--no-dev:** Do not show outdated dev dependencies.
 * **--locked:** Shows updates for packages from the lock file, regardless of what is currently in vendor dir.
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -556,7 +556,7 @@ php composer.phar show monolog/monolog 1.0.2
 * **--major-only (-M):** Use with --latest or --outdated. Only shows packages that have major SemVer-compatible updates.
 * **--minor-only (-m):** Use with --latest or --outdated. Only shows packages that have minor SemVer-compatible updates.
 * **--patch-only:** Use with --latest or --outdated. Only shows packages that have patch-level SemVer-compatible updates.
-* **--sort-by-age (-A):** Sort by and display current release age. Use with the --latest or --outdated option.
+* **--sort-by-age (-A):** Displays the installed version's age, and sorts packages oldest first. Use with the --latest or --outdated option.
 * **--direct (-D):** Restricts the list of packages to your direct dependencies.
 * **--strict:** Return a non-zero exit code when there are outdated packages.
 * **--format (-f):** Lets you pick between text (default) or json output format.
@@ -590,7 +590,7 @@ The color coding is as such:
 * **--major-only (-M):** Only shows packages that have major SemVer-compatible updates.
 * **--minor-only (-m):** Only shows packages that have minor SemVer-compatible updates.
 * **--patch-only (-p):** Only shows packages that have patch-level SemVer-compatible updates.
-* **--sort-by-age (-A):** Sort by and display current release age.
+* **--sort-by-age (-A):** Displays the installed version's age, and sorts packages oldest first.
 * **--format (-f):** Lets you pick between text (default) or json output format.
 * **--no-dev:** Do not show outdated dev dependencies.
 * **--locked:** Shows updates for packages from the lock file, regardless of what is currently in vendor dir.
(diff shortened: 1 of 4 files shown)


src/Composer/Command/ShowCommand.php
```diff
@@ -1447,4 +1470,26 @@ private function getRepositorySet(Composer $composer): RepositorySet
...
+
+        if ($diff->days < 14) {
+            return 'last week';
+        }
+
+        if ($diff->days < 30) {
+            return floor($diff->days / 7) . ' weeks ago';
+        }
+
+        return $releaseDate->format('Y-m-d');
```
Seldaek, line 1493: After a month I revert to just a date here. I don't know if X months ago would bring much more value or not? Any opinions?

Seldaek: Also note that I am showing the time since the release of the new package, not the old one. Were you expecting to see the old one? I am not sure what value either of these infos gives you in the outdated list view tbh.

Seldaek: Ok I added a --sort-by-age / -A flag now, so it shows the installed package's age only when the flag is used, and it also sorts then. That removes the confusion about what date is even shown as it is documented and opt-in.

doc/03-cli.md
```diff
@@ -589,6 +590,7 @@ The color coding is as such:
 * **--major-only (-M):** Only shows packages that have major SemVer-compatible updates.
 * **--minor-only (-m):** Only shows packages that have minor SemVer-compatible updates.
 * **--patch-only (-p):** Only shows packages that have patch-level SemVer-compatible updates.
+* **--sort-by-age (-A):** Sort by and display current release age.
```
naderman, line 593: Same as above, should specify direction and "display installed version release age"

src/Composer/Command/OutdatedCommand.php
```diff
@@ -40,6 +40,7 @@ protected function configure(): void
                 new InputOption('major-only', 'M', InputOption::VALUE_NONE, 'Show only packages that have major SemVer-compatible updates.'),
                 new InputOption('minor-only', 'm', InputOption::VALUE_NONE, 'Show only packages that have minor SemVer-compatible updates.'),
                 new InputOption('patch-only', 'p', InputOption::VALUE_NONE, 'Show only packages that have patch SemVer-compatible updates.'),
+                new InputOption('sort-by-age', 'A', InputOption::VALUE_NONE, 'Sort by and display current release age.'),
```
naderman, line 43: Same as above, should specify direction and "display installed version release age"

src/Composer/Command/ShowCommand.php
```diff
@@ -98,6 +98,7 @@ protected function configure()
                 new InputOption('major-only', 'M', InputOption::VALUE_NONE, 'Show only packages that have major SemVer-compatible updates. Use with the --latest or --outdated option.'),
                 new InputOption('minor-only', 'm', InputOption::VALUE_NONE, 'Show only packages that have minor SemVer-compatible updates. Use with the --latest or --outdated option.'),
                 new InputOption('patch-only', null, InputOption::VALUE_NONE, 'Show only packages that have patch SemVer-compatible updates. Use with the --latest or --outdated option.'),
+                new InputOption('sort-by-age', 'A', InputOption::VALUE_NONE, 'Sort by and display current release age. Use with the --latest or --outdated option.'),
```
naderman, line 101: Same as above, should specify direction and "display installed version release age"

src/Composer/Command/ShowCommand.php
```diff
@@ -469,9 +470,20 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
 
                 $hasOutdatedPackages = false;
 
+                if ($input->getOption('sort-by-age')) {
+                    usort($packages[$type], function ($a, $b) {
+                        if (is_object($a) && is_object($b)) {
+                            return $a->getReleaseDate() <=> $b->getReleaseDate();
+                        }
+
+                        return 0;
```
naderman, line 483: Huh, what else can go in here than objects?

src/Composer/Command/ShowCommand.php
```diff
@@ -1447,4 +1489,30 @@ private function getRepositorySet(Composer $composer): RepositorySet
...
+        if ($diff->days < 7) {
+            return 'this week';
+        }
+
+        if ($diff->days < 14) {
+            return 'last week';
+        }
+
+        if ($diff->m < 1 && $diff->days < 31) {
+            return floor($diff->days / 7) . ' weeks ago';
```
naderman, line 1509: Why floor instead of round?

naderman (changes requested):

src/Composer/Command/ShowCommand.php
Seldaek: Because 2 weeks and 6 days isn't 3 weeks?

naderman: Closer to 3 weeks than it is to 2 weeks, nature of rounding? :-P

src/Composer/Command/ShowCommand.php
Seldaek: When you list everything we only get strings for package names from packagist.org - so it's not always objects, but in relevant cases where --latest would be yes they should always be.

naderman: Ah okay, I see.

HeenaBansal20:
@Seldaek ,  I was trying the show command like below , but the output didn't show the released date of the requested package. Do I need to add some other arg to get the release date

`composer show --available --format=json --ignore-platform-reqs --latest --sort-by-age  yiisoft/yii2 `

Seldaek:
This doesn't work with --available I'm afraid, only for installed packages.

HeenaBansal20:
> This doesn't work with --available I'm afraid, only for installed packages.

hmm.. What exactly this argument `--available ` is for , removing --available from command also didn't show up the release dates.


> --available (-a): List available packages only.

HeenaBansal20:
> > This doesn't work with --available I'm afraid, only for installed packages.
> 
> hmm.. What exactly this argument `--available ` is for , removing --available from command also didn't show up the release dates.
> 
> > --available (-a): List available packages only.

Basically , I want to get the release date for latest version of this package. Is there any way from show command I can get the release date of the packages.

Seldaek:
For an installed package I do see it:

```
$ composer show --format=json --ignore-platform-reqs --latest --sort-by-age  composer/semver | grep release
    "released": "2024-09-19T14:15:21+00:00",
```

In my case composer/semver is installed. --available makes it check all packages on packagist and not just installed, so it runs slower and lists multiple versions so there is no release date I think

PR #11766, opened 2023-12-20
Re-organized PoolOptimizer - reverting #9620 to fix #9393

### LINKED ISSUE #9393: Requirements error says 'but it conflicts with another require' without saying what the problem actually is
My `composer.json`:

```json
{
    "name": "drupal/git-clone-project",
    "description": "TODO",
    "type": "project",
    "homepage": "https://www.drupal.org/project/drupal",
    "support": {
        "docs": "https://www.drupal.org/docs/user_guide/en/index.html",
        "chat": "https://www.drupal.org/node/314178"
    },
    "repositories": [
        {
            "type": "path",
            "url": "repos/drupal",
            "options": {
                "symlink": true
            }
        },
        {
            "type": "path",
            "url": "repos/drupal/core"
        },
        {
            "type": "path",
            "url": "repos/drupal/composer/Plugin/ProjectMessage"
        },
        {
            "type": "path",
            "url": "repos/drupal/composer/Plugin/VendorHardening"
        },
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    ],
    "require": {
        "composer/installers": "^1.9",
        "drupal/core-composer-scaffold": "^8.9",
        "drupal/drupal": "*",
        "drush/drush": "^10.3",
        "symfony/var-dumper": "^5.1"
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "config": {
        "sort-packages": true
    },
    "extra": {
        "drupal-scaffold": {
            "locations": {
                "web-root": "web/"
            }
        },
        "installer-paths": {
            "web/core": [
                "type:drupal-core"
            ],
            "web/libraries/{$name}": [
                "type:drupal-library"
            ],
            "web/modules/contrib/{$name}": [
                "type:drupal-module"
            ],
            "web/profiles/contrib/{$name}": [
                "type:drupal-profile"
            ],
            "web/themes/contrib/{$name}": [
                "type:drupal-theme"
            ],
            "drush/Commands/contrib/{$name}": [
                "type:drupal-drush"
            ],
            "web/modules/custom/{$name}": [
                "type:drupal-custom-module"
            ],
            "web/themes/custom/{$name}": [
                "type:drupal-custom-theme"
            ]
        }
    }
}
```

Additionally need to do the following in the project folder before installing with Composer:

```
mkdir repos
cd repos
git clone --branch 9.2.x https://git.drupalcode.org/project/drupal.git
```

```
require.drupal/drupal : unbound version constraints (*) should be avoided
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys:
OK
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (2.0.3 => 2.0.4)
Composer version: 2.0.3
PHP version: 7.4.2
PHP binary path: /Applications/MAMP/bin/php/php7.4.2/bin/php
OpenSSL version: OpenSSL 1.0.2o  27 Mar 2018
cURL version: 7.65.1 libz 1.2.8 ssl OpenSSL/1.0.2o
zip extension: OK
```

When I run this command:

```
composer require -W drupal/core-dev ^9.2
```

I get the following output:

```
Running composer update drupal/core-dev --with-all-dependencies
Updating dependencies

  Problem 1
    - Root composer.json requires drupal/drupal *, found drupal/drupal[9.2.x-dev] but it conflicts with another require.

```

I need to be told what the 'another require' actually is, otherwise I can't understand or fix the problem.

Toflar:
Imho this should fix https://github.com/composer/composer/issues/9393 by reverting https://github.com/composer/composer/pull/9620.

The problem of https://github.com/composer/composer/pull/9620 is that it marks a package for removal although this has been previously marked to be kept. This causes the bug described in #9393. While debugging, I noticed that understanding the logic in the `PoolOptimizer` is not as easy because it's designed to 

* first mark all packages for removal
* then decide which ones of those should be kept

This was fine because it was the way it was initially designed when I first contributed the optimizer in https://github.com/composer/composer/pull/9261 which only consisted of `optimizeByIdenticalDependencies()`.

However, https://github.com/composer/composer/issues/9393 introduced a second optimization step (`optimizeImpossiblePackagesAway()`) in which additional packages are marked for removal even if the first step marked them as to be kept - causing the mentioned bug.

So what I did in this PR is invert the logic:

* mark all packages that need to be kept
* then remove all the remaining ones

@driskell - maybe you could try to re-implement your logic of https://github.com/composer/composer/pull/9620 on top of this PR? Basically instead of `markPackageForRemoval()` they should not even be marked as required to keep.

### DIFF
--- a/src/Composer/DependencyResolver/PoolOptimizer.php
+++ b/src/Composer/DependencyResolver/PoolOptimizer.php
@@ -49,11 +49,6 @@ class PoolOptimizer
     private $conflictConstraintsPerPackage = [];
 
     /**
-     * @var array<int, true>
-     */
-    private $packagesToRemove = [];
-
-    /**
      * @var array<int, BasePackage[]>
      */
     private $aliasesPerPackage = [];
@@ -74,8 +69,6 @@ class PoolOptimizer
 
         $this->optimizeByIdenticalDependencies($request, $pool);
 
-        $this->optimizeImpossiblePackagesAway($request, $pool);
-
         $optimizedPool = $this->applyRemovalsToPool($pool);
 
         // No need to run this recursively at the moment
@@ -86,7 +79,6 @@ class PoolOptimizer
         $this->irremovablePackages = [];
         $this->requireConstraintsPerPackage = [];
         $this->conflictConstraintsPerPackage = [];
-        $this->packagesToRemove = [];
         $this->aliasesPerPackage = [];
         $this->removedVersionsByPackage = [];
 
@@ -145,6 +137,11 @@ class PoolOptimizer
 
     private function markPackageIrremovable(BasePackage $package): void
     {
+        // Already marked to keep
+        if (isset($this->irremovablePackages[$package->id])) {
+            return;
+        }
+
         $this->irremovablePackages[$package->id] = true;
         if ($package instanceof AliasPackage) {
             // recursing here so aliasesPerPackage for the aliasOf can be checked
@@ -166,7 +163,7 @@ class PoolOptimizer
         $packages = [];
         $removedVersions = [];
         foreach ($pool->getPackages() as $package) {
-            if (!isset($this->packagesToRemove[$package->id])) {
+            if (isset($this->irremovablePackages[$package->id])) {
                 $packages[] = $package;
             } else {
                 $removedVersions[$package->getName()][$package->getVersion()] = $package->getPrettyVersion();
@@ -190,8 +187,6 @@ class PoolOptimizer
             if (isset($this->irremovablePackages[$package->id])) {
                 continue;
             }
-
-            $this->markPackageForRemoval($package->id);
 
             $dependencyHash = $this->calculateDependencyHash($package);
 
@@ -240,7 +235,7 @@ class PoolOptimizer
                 foreach ($constraintGroup as $packages) {
                     // Only one package in this constraint group has the same requirements, we're not allowed to remove that package
                     if (1 === \count($packages)) {
-                        $this->keepPackage($packages[0], $identicalDefinitionsPerPackage, $packageIdenticalDefinitionLookup);
+                        $this->keepPackageWithinDependencyGroups($packages[0], $identicalDefinitionsPerPackage, $packageIdenticalDefinitionLookup);
                         continue;
                     }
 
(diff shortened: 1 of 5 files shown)


Toflar:
We certainly also need to add a reproducer test for https://github.com/composer/composer/issues/9393 - I'll try to work on that.

src/Composer/DependencyResolver/PoolOptimizer.php
```diff
@@ -364,70 +342,6 @@ private function keepPackage(BasePackage $package, array $identicalDefinitionsPe
...
-                continue;
-            }
-
-            $packageIndex[$package->getName()][$package->id] = $package;
-        }
-
-        foreach ($request->getLockedPackages() as $package) {
-            // If this locked package is no longer required by root or anything in the pool, it may get uninstalled so do not apply its requirements
-            // In a case where a requirement WERE to appear in the pool by a package that would not be used, it would've been unlocked and so not filtered still
-            $isUnusedPackage = true;
```
driskell, line 402: @Toflar It's a weird one, maybe I missed something but I feel it was considered because of this area. Potentially is this code not working?

In #9393 it looks like it would need this code - to not apply the constraints from a package if it's no longer required.  Perhaps when it got moved to the optimisation step and started using `requireConstraintsPerPackage` it changed something.

I can see if I can find some time soon to relook at this in this location. But I think it's this specific 7 lines of code that is failing us 🤔

driskell: Probably worth dropping it out for now. I think `requireConstraintsPerPackage` is including requirements of each package added to the pool so if multiple versions got added of something and one of those versions references a package that would still be removed it likely then optimises away. One for a separate PR with a better test.

Is it possible to generate a failing test for this for future attempts?

I think at a guess it needs a package in the pool with two versions, one of those that would be installed, the other that references a require on a package that is going to be removed. Then that package that is going to be removed has a constraint that can't be met.

Toflar: > Is it possible to generate a failing test for this for future attempts?

Yeah, I think I know exactly why this is happening. I just need some time to come up with a reproducer test but I do think we can also keep the existing locked packages optimizer, I just have to sit down and work on it 😇

Toflar: I've described the issue here: https://github.com/composer/composer/issues/9393#issuecomment-1866616992 - relying on `requireConstraintsPerPackage` is exactly the problem because that does never contain the packages that **will be required** but more than that. Hence, removing packages based on that can always fail.

Toflar:
See https://github.com/composer/composer/pull/11772 for the test.

PR #11769, opened 2023-12-21
Add Global Command tests

franckranaivo:
Add Test for Global Command as it is required in https://github.com/composer/composer/issues/10796

This PR is still a work in progress as I try to acquire internal knowledge.

If anyone sees what is nice to test for the global command ;)  , I'm open to suggestions.

### DIFF
--- a/tests/Composer/Test/Command/GlobalCommandTest.php
+++ b/tests/Composer/Test/Command/GlobalCommandTest.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Composer.
+ *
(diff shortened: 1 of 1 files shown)


Seldaek:
I think we'd need to set up a global composer dir, you can do this by creating a temp dir in the tests and then setting `Platform::putEnv('COMPOSER_HOME', the new temp dir)` so it gets used as global dir then the global dir could be populated with some dependencies or whatever to make sure that we can use `composer global show` for example and it returns the stuff installed in the global dir.

As I wrote on the issue tho this is one of the most involved commands to test as it does some odd things internally.

Issue #11765, opened 2023-12-19
Option to disable fund message


shyim:
I would love to have an option like `--no-fund` like on [NPM](https://docs.npmjs.com/cli/v7/commands/npm-install#fund) to disable that message and the data aggregation.

I just like to have my terminal output clean 😅

alcohol:
> the data aggregation

Can you elaborate?

Seldaek:
Fixed by https://github.com/composer/composer/pull/11779

PR #11774, opened 2023-12-23
Add configuration option to disable fund message

samlitowitz:
Add a configuration option to disable the fund message addressing #11765

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -77,7 +77,7 @@ class Config
         'htaccess-protect' => true,
         'use-github-api' => true,
         'lock' => true,
-        'no-fund' => false,
+        'fund' => true,
         'platform-check' => 'php-only',
         'bitbucket-oauth' => [],
         'github-oauth' => [],
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -77,6 +77,7 @@ class Config
         'htaccess-protect' => true,
         'use-github-api' => true,
         'lock' => true,
+        'no-fund' => false,
         'platform-check' => 'php-only',
         'bitbucket-oauth' => [],
         'github-oauth' => [],
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -375,22 +375,24 @@ class Installer
             }
         }
 
-        $fundingCount = 0;
-        foreach ($localRepo->getPackages() as $package) {
-            if ($package instanceof CompletePackageInterface && !$package instanceof AliasPackage && $package->getFunding()) {
-                $fundingCount++;
+        if ($this->fund) {
+            $fundingCount = 0;
+            foreach ($localRepo->getPackages() as $package) {
+                if ($package instanceof CompletePackageInterface && !$package instanceof AliasPackage && $package->getFunding()) {
+                    $fundingCount++;
+                }
             }
-        }
-        if ($this->fund && $fundingCount > 0) {
-            $this->io->writeError([
-                sprintf(
-                    "<info>%d package%s you are using %s looking for funding.</info>",
-                    $fundingCount,
-                    1 === $fundingCount ? '' : 's',
-                    1 === $fundingCount ? 'is' : 'are'
-                ),
-                '<info>Use the `composer fund` command to find out more!</info>',
-            ]);
+            if ($fundingCount > 0) {
+                $this->io->writeError([
+                    sprintf(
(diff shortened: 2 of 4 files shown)


src/Composer/Config.php
```diff
@@ -77,6 +77,7 @@ class Config
         'htaccess-protect' => true,
         'use-github-api' => true,
         'lock' => true,
+        'no-fund' => false,
```
staabm, line 80: should the config flag be named `fund` to prevent double negation?

samlitowitz: Excellent suggestion, thank you!

src/Composer/Installer.php
```diff
@@ -378,7 +381,7 @@ public function run(): int
                 $fundingCount++;
             }
         }
-        if ($fundingCount > 0) {
+        if ($this->fund && $fundingCount > 0) {
```
stloyd, line 384: Maybe skip whole counting if it's disabled?

samlitowitz: I think this would make it easier to understand the code even though it pushes code to the right. Thank you!

samlitowitz:
@Seldaek for visibility because you commented on the originating issue #11765

tests/Composer/Test/Fixtures/installer/install-funding-notice-no-fund.test
```diff
@@ -0,0 +1,61 @@
...
+                }
+            ]
+        }
+    ],
+    "require": {
+        "a/a": "1.0.0"
+    },
+    "config": {
+        "fund": false
+    }
```
fredden, line 42: @Seldaek said that this should be an environment variable, not a configuration option. Can you help us understand why you have decided to make this a configuration option anyway?

Ref: https://github.com/composer/composer/issues/11765#issuecomment-1864686605
> there could be an env var to turn this off if anything, I don't think flags or config option makes sense

samlitowitz: @fredden 
I think the ability to "set it and forget it" across multiple environments provides an ease of use which is valuable to those who would use this option and a configuration option seems like the best method to do so.

I think the ability to temporarily or individually set this is also very helpful and an environmental variable would solve this. I do think that adding support for the environmental variable should be treated as a different "feature" done under a separate PR.

fredden:
Thinking about this a little more, I suggest that a simple boolean isn't the right approach here. What are your thoughts on a list of packages to ignore the funding requests/messages for instead? This way when a new dependency is added then it's an explicit choice to not fund that new dependency, rather than an implicit "we do not fund anything" choice. This feels more in keeping with the (assumed) intent: to increase funding of open-source software.

samlitowitz:
> Thinking about this a little more, I suggest that a simple boolean isn't the right approach here. What are your thoughts on a list of packages to ignore the funding requests/messages for instead? This way when a new dependency is added then it's an explicit choice to not fund that new dependency, rather than an implicit "we do not fund anything" choice. This feels more in keeping with the (assumed) intent: to increase funding of open-source software.

To me this seems like another useful feature which gives library/projects a more granular option in addition to the all or none model. I think it would be great as an addition to this current work. My concern with using it as a replacement for the `fund` option is that it would not be inline with the original ask and would likely be so cumbersome to use for that case as to be unusable.


Having the `fund` option default to true means a library/project would have to make an explicit choice to hide funding messages. I believe that the majority of libraries and projects will retain the default value to support open source work.

tests/Composer/Test/Fixtures/installer/install-funding-notice-no-fund.test
samlitowitz: > With an environment variable, a user can "set and forget" a system-level environment variable on their machine. This then has a consistent experience for that user across all projects with which they interact.

Absolutely, I think an environmental variable would be very useful. To my mind it is a related but separate piece of work to be done in a different PR. I believe it probably should be done even if this change is not accepted and definitely be done if this change is accepted.

samlitowitz: @fredden As an update to this conversation, I have made another PR (#11779 ) which adds support for an environmental variable to show/hide the funding message.

Seldaek:
Thanks for the PR but IMO the env var in https://github.com/composer/composer/pull/11779 is more than enough. And having the option to ignore by name etc seems way overkill to me.

Issue #11744, opened 2023-12-04
composer self update (missing hyphen) ought to give a sensible hint or error message.


HenkPoley:
My `composer.json`:

```text
Irrelevant, happens without composer.json file.
```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK git version 2.43.0
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 2.6.5
PHP version: 8.0.27 - Package overridden via config.platform, actual: 8.0.30
PHP binary path: C:\laragon\bin\php\php-8.0.30-Win32-vs16-x64\php.exe
OpenSSL version: OpenSSL 1.1.1t  7 Feb 2023
cURL version: 7.76.1 libz 1.2.12 ssl OpenSSL/1.1.1t
zip: extension present, unzip not available, 7-Zip present (7z)
```

When I run this command: 

```
C:\laragon\bin\php\php-8.0.30-Win32-vs16-x64\php.exe .\composer.phar self update
```

I get the following output: 

```
Upgrading to version update (stable channel).

In SelfUpdateCommand.php line 245:
                                        
  Version "update" could not be found.  
                                        

In CurlDownloader.php line 630:
                                                                                                              
  The "https://getcomposer.org/download/update/composer.phar.sig" file could not be downloaded (HTTP/2 404 )  
                                                                                                              

self-update [-r|--rollback] [--clean-backups] [--no-progress] [--update-keys] [--stable] [--preview] [--snapshot] [--1] [--2] [--2.2] [--set-channel-only] [--] [<version>]
```

And I expected this to happen:

Some kind message "Did you mean self-update? [Y/n]".

HenkPoley:
Durr..

`php composer.phar self update`

versus

`php composer.phar self-update`

Maybe add something helpful for this PEBCAK ? 🤣

Seldaek:
[...]

PR #11775, opened 2023-12-24
[11744] handle missing hyphen when attempting to run self-update…

lykbl:
Fixes [#11744](https://github.com/composer/composer/issues/11744) 

Gracefully handle `composer self update` typo. Prompt to continue as `composer self-update`, otherwise throw the `CommandNotFoundException`.

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -23,6 +23,7 @@ use Composer\SelfUpdate\Versions;
 use Composer\IO\IOInterface;
 use Composer\Downloader\FilesystemException;
 use Composer\Downloader\TransportException;
+use Symfony\Component\Console\Exception\CommandNotFoundException;
 use Symfony\Component\Console\Input\InputInterface;
 use Composer\Console\Input\InputOption;
 use Composer\Console\Input\InputArgument;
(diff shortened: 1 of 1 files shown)


src/Composer/Command/SelfUpdateCommand.php
```diff
@@ -156,6 +157,15 @@ class_exists('Composer\Downloader\FilesystemException');
             return $this->rollback($output, $rollbackDir, $localFilename);
         }
 
+        if ($input->getArgument('command') === 'self' && $input->getArgument('version') === 'update') {
+            $question = '<info>"update"</info> is not a valid version. Did you mean "composer self-update" [<comment>Y/n</comment>]? ';
+            if ($io->isInteractive() && $io->askConfirmation($question, false)) {
+                $input->setArgument('version', null);
+            } else {
+                throw new CommandNotFoundException(sprintf('The command "%s" does not exist.', $input->getArgument('command')));
+            }
```
Seldaek, line 166: I think we could just always fix this silently, it's pretty obvious what the reason is, and there's no point in asking for confirmation nor in failing hard if non-interactive.

```suggestion
            $input->setArgument('version', null);
```

lykbl: Sure, this one's updated

Seldaek:
Cool, thanks!

PR #11783, opened 2024-01-03
Adds a test for invalid arg combo

theoboldalex:
re: #10796 

Adds a single test to cover an invalid argument combination on the config command.


Seldaek:
Thanks

Issue #11771, opened 2023-12-22
`composer audit` checks alias instead of actual package version


jfastnacht:
My `composer.json`:

```json
{
    "name": "jfastnacht/composer-audit-alias-test",
    "require": {
        "guzzlehttp/guzzle": "7.7.0 as 6.0.1"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
require.guzzlehttp/guzzle : exact version constraints (7.7.0 as 6.0.1) should be avoided if the package follows semantic versioning
Checking platform settings: OK
Checking git settings: OK git version 2.30.2
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: FAIL
Missing pubkey for tags verification
Missing pubkey for dev verification
Run composer self-update --update-keys to set them up
Checking composer version: OK
Composer version: 2.6.6
PHP version: 8.0.30
PHP binary path: /usr/bin/php8.0
OpenSSL version: OpenSSL 1.1.1n  15 Mar 2022
cURL version: 7.74.0 libz 1.2.11 ssl OpenSSL/1.1.1w
zip: extension present, unzip present, 7-Zip not available
```

When I run this command: 

```
composer audit -vvv --locked
```

I get the following output: 

```
Running 2.6.6 (2023-12-08 18:32:26) with PHP 8.0.30 on Linux / 5.15.133.1-microsoft-standard-WSL2
Reading ./composer.json (/var/www/html/composer.json)
Loading config file /home/jfastnacht/.composer/auth.json
Loading config file ./composer.json (/var/www/html/composer.json)
gitlab.<domain>.<tld> is not in the configured gitlab-domains, adding it implicitly as authentication is configured for this domain
Checked CA file /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a file.
Checked directory /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a directory.
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/var/www/html): 'git' 'branch' '-a' '--no-color' '--no-abbrev' '-v'
Executing command (/var/www/html): git describe --exact-match --tags
Executing command (CWD): git --version
Executing command (/var/www/html): git log --pretty="%H" -n1 HEAD --no-show-signature
[...]

PR #11789, opened 2024-01-10
test: Covers audit of pkg with no sec advisories

theoboldalex:
Adds a test case for the happy path of the Audit Command where a given package has no security advisories.


Seldaek:
Please rebase on 2.6 branch if you add fixes for #11771 - if it's just new tests I guess it doesn't matter as much

Seldaek:
Thanks

PR #11794, opened 2024-01-12
Introduce COMPOSER_AUDIT_ABANDONED env var

### LINKED ISSUE #11792: Introduce an environment variable for setting abandoned package reporting mode in composer audit
#11639 introduced an abandoned package reporting mode without providing an environment variable for customization. The default reporting mode is set to change to "fail" in [2.7.0](https://github.com/composer/composer/issues/11736). Rather than requiring code adjustments in every project to revert to the reporting mode via [configuration change in composer.json](https://getcomposer.org/doc/06-config.md#abandoned), it would be favorable to allow configuration through an environment variable in CI environments.

Another use case for implementing this environment variable is the ability to dynamically switch between reporting modes for specific CI jobs as needed (like use `fail` on PR review and `report`ing on nightly scanners that measures collected technical debt).

mxr576:
Closes #11792

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1237,6 +1237,11 @@ similar use case), and need to support proxies, please provide the `CGI_HTTP_PRO
 environment variable instead. See [httpoxy.org](https://httpoxy.org/) for further
 details.
 
+### COMPOSER_AUDIT_ABANDONED
+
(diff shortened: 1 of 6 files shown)


Seldaek:
Thanks, I'll add some more thoughts on the issue

PR #11795, opened 2024-01-12
Add support for php extension packages

Seldaek:
Basic infrastructure to let pie reuse Composer code for PHP extension packages.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -294,13 +294,31 @@
             "properties": {
                 "priority": {
                     "type": "integer",
-                    "description": "The installation \"priority\". Defaults to 20, and is used when generating an ini file to enable the extension."
+                    "description": "This is used to add a prefix to the INI file, e.g. `90-xdebug.ini` which affects the loading order. The priority is a number in the range 10-99 inclusive, with 10 being the highest priority (i.e. will be processed first), and 99 being the lowest priority (i.e. will be processed last). There are two digits so that the files sort correctly on any platform, whether the sorting is natural or not.",
+                    "minimum": 10,
+                    "maximum": 99,
+                    "example": 80,
+                    "default": 80
                 },
-                "config": {
-                    "type": "object",
-                    "description": "A list of configure options that can be passed to `./configure`, and whether whey would be passed by default to configure in that stage (either true, or false). These must match the names in the config.m4 file.",
-                    "additionalProperties": {
-                        "type": ["boolean"]
+                "configure-options": {
+                    "type": "array",
+                    "description": "These configure options make up the flags that can be passed to ./configure when installing the extension.",
+                    "items": {
+                        "type": "object",
+                        "required": ["name"],
+                        "properties": {
+                            "name": {
+                                "type": "string",
+                                "description": "The name of the flag, this would typically be prefixed with `--`, for example, the value 'the-flag' would be passed as `./configure --the-flag`.",
+                                "example": "without-xdebug-compression",
+                                "pattern": "^[a-zA-Z0-9][a-zA-Z0-9-_]*$"
+                            },
+                            "description": {
+                                "type": "string",
+                                "description": "The description of what the flag does or means.",
+                                "example": "Disable compression through zlib"
+                            }
+                        }
                     }
                 }
             }
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -6,7 +6,7 @@
         "name": {
             "type": "string",
             "description": "Package name, including 'vendor-name/' prefix.",
-            "pattern": "^[a-z0-9]([_.-]?[a-z0-9]+)*\/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*$"
+            "pattern": "^([a-z0-9]([_.-]?[a-z0-9]+)*\/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*|ext-[a-z0-9]([_.-]?[a-z0-9]+)*)$"
         },
         "description": {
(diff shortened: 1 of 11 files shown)


res/composer-schema.json
```diff
@@ -288,6 +288,23 @@
...
+            "type": "object",
+            "description": "Settings for PHP extension packages.",
+            "properties": {
+                "priority": {
+                    "type": "integer",
+                    "description": "The installation \"priority\". Defaults to 20, and is used when generating an ini file to enable the extension."
+                },
+                "config": {
+                    "type": "object",
+                    "description": "A list of configure options that can be passed to `./configure`, and whether whey would be passed by default to configure in that stage (either true, or false). These must match the names in the config.m4 file.",
```
stof, line 301:
```suggestion
                    "description": "A list of configure options that can be passed to `./configure`, and whether they would be passed by default to configure in that stage (either true, or false). These must match the names in the config.m4 file.",
```

src/Composer/Package/Loader/ValidatingArrayLoader.php
```diff
@@ -449,7 +451,11 @@ public static function hasPackageNamingError(string $name, bool $isLink = false)
             return null;
         }
 
-        if (!Preg::isMatch('{^[a-z0-9](?:[_.-]?[a-z0-9]++)*+/[a-z0-9](?:(?:[_.]|-{1,2})?[a-z0-9]++)*+$}iD', $name)) {
+        if (str_starts_with($name, 'ext-')) {
```
stof, line 454: should this ensure that `ext-*` packages have a type corresponding to `php-ext` or `php-ext-zend` ?

src/Composer/Package/Loader/ValidatingArrayLoader.php
```diff
@@ -245,6 +245,8 @@ public function load(array $config, string $class = 'Composer\Package\CompletePa
             }
         }
 
+        $this->validateArray('php-ext');
```
stof, line 248: should we validate that this is set only for packages with the extension types ?

stof:
Also, should the ValidatingArrayLoader enforce that an extension package only defines links (requirements and conflicts) on platform packages ? It would not make any sense for an extension to depend on a userland package AFAICT.

src/Composer/Package/Loader/ValidatingArrayLoader.php
stof, line 454: this will prevent vendor names starting with `ext-`, which are valid today AFAICT. This condition should probably also do `&& !str_contains($name, '/')`

res/composer-schema.json
```diff
@@ -288,6 +288,23 @@
...
+                    "additionalProperties": {
+                        "type": ["boolean"]
```
derickr, line 303: This is not supposed to be a boolean, as many will have string variants. The `false` from my doc was meant to signal "don't set / add these by default" when calling `./configure`.

stof: So should this be `"type": ["boolean", "string"]` ?

Seldaek: As this is still unclear and being worked on, I'm postponing this PR to 2.8

res/composer-schema.json
```diff
@@ -288,6 +288,23 @@
                 }
             }
         },
+        "php-ext": {
```
asgrim, line 291: We have a proposed JSON schema for this, which I think is more up to date here:

https://github.com/ThePHPF/pie-design/blob/c30488fd067240532eddfca54300b0a3ed20bee6/composer-json-php-ext-schema.json#L8-L42

I think this would eliminate the questions about `additionalProperties` below

res/composer-schema.json
```diff
@@ -6,7 +6,7 @@
         "name": {
             "type": "string",
             "description": "Package name, including 'vendor-name/' prefix.",
-            "pattern": "^[a-z0-9]([_.-]?[a-z0-9]+)*\/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*$"
+            "pattern": "^([a-z0-9]([_.-]?[a-z0-9]+)*\/[a-z0-9](([_.]|-{1,2})?[a-z0-9]+)*|ext-[a-z0-9]([_.-]?[a-z0-9]+)*)$"
```
asgrim, line 9: Questions have been raised about the naming here; so we should agree first on the best approach on https://github.com/ThePHPF/pie-design/issues/8 before merging this IMO

res/composer-schema.json
Seldaek: Thanks, I updated it. Please let me know if the schema changes on your end in the future.

asgrim: It will change very soon actually; I need to put some ZTS/NTS compatibility settings in there; will ping you on the MR when I open it!

res/composer-schema.json
Seldaek: I have rolled back the ext- package naming changes for now, so IMO this would be mergeable as is as a first step. What do you think?

asgrim: > IMO this would be mergeable as is as a first step. What do you think?

Question: if someone adds a package type `php-ext`/`php-ext-zend` to their project's `composer.json`, for example say we have `xdebug/xdebug` eventually, what will Composer do? will it download the source code into their `vendor` directory?

Seldaek: Yes.. Not sure if we should prevent this or leave it as an exercise to the reader. I think I could pretty easily just ignore these packages in the solver and make them invisible to Composer.

Seldaek: Ok done

PR #11798, opened 2024-01-17
Add failing test and fix attempt

Seldaek:
Refs https://github.com/sebastianbergmann/phpunit/pull/5661

Providers which are not required by anything should not be installable. However the only way to fix this is to make every package require back everything that requires it, to avoid dangling providers being selected.

This may use up too much memory and CPU to be worth fixing, but we'll try here.. This first attempt works for the new test but it does break other things:

- removal of unused packages, I assume because they now have back-requires keeping them selected
- detection of dev require vs require, also probably due to the back requires making it all appear to be non-dev requires


src/Composer/DependencyResolver/RuleSetGenerator.php
```diff
@@ -199,12 +200,20 @@ protected function addRulesForPackage(BasePackage $package, PlatformRequirementF
                 }
 
                 $possibleRequires = $this->pool->whatProvides($link->getTarget(), $constraint);
+                foreach ($possibleRequires as $index => $require) {
+                    $requirers = [];
+                    foreach ($this->pool->getPackages() as $pkg) {
+                        if (isset($pkg->getRequires()[$require->getName()])) {
```
naderman, line 206: you need to loop over all `$require->getNames()` here to make this work for providers of different names?

Issue #11811, opened 2024-01-24
auth token links on separate line?


delboy1978uk:
Feature request!

When interactively authorising on  GitHub, it would be great if the generate token links were on their own separate line, in order to make copy pasting easier 😀

I have a PR for you which I'll open in a minute!

Seldaek:
Fixed by https://github.com/composer/composer/pull/11812

PR #11812, opened 2024-01-24
issue #11811 auth token links on separate lines

delboy1978uk:
links now display on their own line


delboy1978uk:
@stloyd I guess we should do this with the other classes too that implement `authorizeOAuthInteractively()`?
https://github.com/search?q=repo%3Acomposer%2Fcomposer+authorizeOAuthInteractively&type=code

Seldaek:
Thanks

PR #11818, opened 2024-01-28
Add post-outdated-cmd and pre-outdated-cmd events

alexander-schranz:
I use a root composer.json inside my monorepository and forward commands to run in all subpackages. Beside install and update, I would like todo that also for the `outdated` command here:  https://github.com/schranz-search/schranz-search/blob/6e84eb2692ea257397d9ebe7b67fce1753cdba98/composer.json#L15

So I thought I could add new events to achieve this.

Example Usage:

```json
{
    "type": "project",
    "license": "MIT",
    "require-dev": {
        "schranz/mono": "^2.0.1"
    },
    "scripts": {
        "post-outdated-cmd": "@php vendor/bin/mono run composer outdated"
    }
}
```

Let me know what you think about this new events?

### DIFF
--- a/src/Composer/Command/OutdatedCommand.php
+++ b/src/Composer/Command/OutdatedCommand.php
@@ -12,6 +12,9 @@
 
 namespace Composer\Command;
 
+use Composer\Plugin\CommandEvent;
+use Composer\Plugin\PluginEvents;
+use Composer\Script\ScriptEvents;
 use Symfony\Component\Console\Input\InputInterface;
 use Composer\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\ArrayInput;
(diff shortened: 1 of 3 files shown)


src/Composer/Command/OutdatedCommand.php
```diff
@@ -110,7 +113,22 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $input = new ArrayInput($args);
 
-        return $this->getApplication()->run($input, $output);
+        $composer = $this->tryComposer();
+
+        if ($composer !== null) {
+            $commandEvent = new CommandEvent(PluginEvents::COMMAND, 'outdated', $input, $output);
+            $eventDispatcher = $composer->getEventDispatcher();
+            $eventDispatcher->dispatch($commandEvent->getName(), $commandEvent);
```
alexander-schranz, line 121: I orientated by the ArchiveCommand not sure if this additional dispatch is expected or the dispatchScript is enough to be added.

Seldaek:
Why not use just the PluginEvents::COMMAND event which fires for every command run?

```
$ COMPOSER_DEBUG_EVENTS=1 c outdated
Dispatching init event
Dispatching pre-command-run (outdated) event
Dispatching pre-command-run (show) event
Dispatching command (show) event
```

pre-command-run or command event should do no?

alexander-schranz:
@Seldaek wanted that it works in all cases the same and is transparent in the composer.json what is happening. As I already use the following post commands already:

```
        "post-install-cmd": "@php vendor/bin/mono run composer install",
        "post-update-cmd": "@php vendor/bin/mono run composer update",
```

Seldaek:
OK but then it sounds to me like we'd need a more generic solution like the pre-command-run, a post-command-run that fires always after the end of the command?

alexander-schranz:
If you have an idea how to implement a general solution I'm also open for that :) The only thing is that we may then not have constants for the pre and post commands if somebody uses them?

Seldaek:
Well have a look where pre-command-run is dispatched in BaseCommand, we could dispatch a post-command-run there as well IMO before returning?

Seldaek:
Is this still needed or should we close?

alexander-schranz:
@Seldaek more lack of time, did not yet continue on it. But I can keep track of the issue on my side https://github.com/alexander-schranz/mono/issues/12. And bring in future a merge request up when I have time again. So we can currently close this and if I have time I will resubmit a merge request.

PR #11825, opened 2024-01-31
chore(doc): add `_comment` documentation inside `composer.json` schema

94noni:
I've asked a discussion on symfony slack and got this info, found it not documented so my I propose to add it

as per https://github.com/composer/composer/blob/d3aeb1357f8cb00474f0b640279d192301b3906f/CHANGELOG.md?plain=1#L1388


Seldaek:
Thanks

PR #11830, opened 2024-02-06
ValidatingArrayLoader: fix link validation with missing name

glaubinix:
Fixes `Undefined array key "name"`

Wasn't sure if there is going to be another 2.6 release but commit can be applied on main too.


Seldaek:
Thanks

PR #11842, opened 2024-02-09
Output more warnings about plugins being disabled to hint that it may cause problems

### LINKED ISSUE #11839: Composer 2.7.0 usage with root user leads to plugins not being loaded
### Edit from @Seldaek: Here is an explanation and solution:

The problem stems from the fact that when running as root, for safety we now disable plugins. If running interactively Composer prompts you whether you want to enable them, but when non-interactive we just disable them and output a warning.

If you want to allow plugins to run you should set the `COMPOSER_ALLOW_SUPERUSER=1` environment variable. This will suppress the warning and let Composer run as usual even when running as root.

However, you should IMO only do this if you can ensure you run trusted code, or if you are in a disposable environment/containerized CI that will get reset at the end of the run anyway. If you run composer with sudo in production machines however IMO you are likely doing something wrong and you should re-evaluate these processes instead of simply suppressing the errors.

See also https://getcomposer.org/doc/faqs/how-to-install-untrusted-packages-safely.md

----
----

#### Original issue text:

Our running platform is AWS ElasticBeanstalk.
The deployment process is failing when using [composer 2.7.0](https://github.com/composer/composer/releases/tag/2.7.0) whereas success when using 2.6.6.

> sh: symfony-cmd: command not found
> Script symfony-cmd handling the auto-scripts event returned with error code 127

The problem is that AWS ElasticBeanstalk run `composer` as `root` (as I can see from logs).
composer 2.7.0 is fixing [CVE-2024-24821](https://github.com/advisories/GHSA-7c6p-848j-wh5h).. 

To solve the problem, I force the version of composer to version 2.6.6 : `composer self-upgrade 2.6.6`.

I opened this issue to help other people to find a solution.

Seldaek:
Fixes #11839

This adds 3 warnings:

One during installation if plugins are disabled:

image

One additional line before a command fails to be found if plugins are disabled:

image

And one generic `Plugins have been disabled automatically as you are running as root, this may be the cause of the following exception. See also https://getcomposer.org/root` line if plugins are disabled and you're running as root + non-interactive, also just before a thrown exception is displayed.

### DIFF
--- a/src/Composer/Installer/InstallationManager.php
+++ b/src/Composer/Installer/InstallationManager.php
@@ -94,8 +94,8 @@ class InstallationManager
     /**
      * Disables plugins.
      *
-     * We prevent any plugins from being instantiated by simply
-     * deactivating the installer for them. This ensure that no third-party
+     * We prevent any plugins from being instantiated by
+     * disabling the PluginManager. This ensures that no third-party
      * code is ever executed.
      */
     public function disablePlugins(): void
@@ -105,7 +105,7 @@ class InstallationManager
                 continue;
             }
 
-            unset($this->installers[$i]);
+            $installer->disablePlugins();
         }
     }
 
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -220,7 +220,7 @@ class Application extends BaseApplication
 
         // Clobber sudo credentials if COMPOSER_ALLOW_SUPERUSER is not set before loading plugins
         if ($needsSudoCheck) {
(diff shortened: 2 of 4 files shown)


src/Composer/Installer/InstallationManager.php
```diff
@@ -105,7 +105,7 @@ public function disablePlugins(): void
                 continue;
```
naderman, line 105: technically the comment on this function doesn't fit anymore now that you call a function on the installers to disable.

PR #11840, opened 2024-02-08
Repositories docs reference

theoboldalex:
Adds a small section to the repositories page of the docs based on [a comment](https://github.com/composer/composer/issues/11735#issuecomment-1830251947) from @stof on #11735

PR #11844, opened 2024-02-09
Makes note appear in a note section.

theoboldalex:
PR #11840 added docs.

The note section was missing angle brackets at the start of the line so the note section did not appear in the coloured block.


Seldaek:
Ah yeah missed that :D

theoboldalex:
Apologies, I should have spotted that before submitting the first PR :D

Issue #11787, opened 2024-01-09
`update --lock` should not update packages


obriat:
I ran `composer update --lock` in order to fix the "Warning: The lock file is not up to date with the latest changes in composer.json..." after removing a `cweagans/composer-patches`patch in the `composer.json`.
I was expecting to only get a hash change in the `composer.lock` as said in the documentation.
But there were also minor changes :

- some `authors` updates
- a `timestamp` and `version` update in a drupal module  ("type": "drupal-module"  drupal/rest_entity_display:dev-2.0.x > extra  > drupal > datestamp + version)
- an `abandoned` addition in php-http/guzzle6-adapter:v2.0.2 
- funding addition, source > url & dist > url updates in  squizlabs/php_codesniffer:3.7.2

If it's a normal behavior, maybe the doc (https://getcomposer.org/doc/03-cli.md#update-u-upgrade) should be updated. Adding a notice such as  "expect also other minor changes on packages metadata) ?

Output of `composer diagnose`:

```
composer diagnose
Checking composer.json: WARNING
require.cloudflare/sdk : exact version constraints (1.1.6) should be avoided if the package follows semantic versioning
require.drupal/base_field_display : exact version constraints (1.0.x-dev@dev) should be avoided if the package follows semantic versioning
require.drupal/core-composer-scaffold : exact version constraints (9.5.11) should be avoided if the package follows semantic versioning
require.drupal/core-project-message : exact version constraints (9.5.11) should be avoided if the package follows semantic versioning
require.drupal/core-recommended : exact version constraints (9.5.11) should be avoided if the package follows semantic versioning
require.drupal/leaflet_mapbox : exact version constraints (1.0-alpha1) should be avoided if the package follows semantic versioning
require.drupal/leaflet_maptiler : exact version constraints (1.0-beta10) should be avoided if the package follows semantic versioning
require.drupal/office_hours : exact version constraints (1.11) should be avoided if the package follows semantic versioning
require.drupal/openapi_ui : exact version constraints (1.0-rc4) should be avoided if the package follows semantic versioning
require.drupal/openapi_ui_swagger : exact version constraints (1.0-rc4) should be avoided if the package follows semantic versioning
require.drupal/read_time : exact version constraints (2.x-dev@dev) should be avoided if the package follows semantic versioning
require.drupal/rest_entity_display : exact version constraints (2.0.x-dev@dev) should be avoided if the package follows semantic versioning
[...]

PR #11850, opened 2024-02-13
Issue in `composer.lock` generation ? 

drupol:
Hello,

Today I encountered an unexpected behaviour while attempting to upgrade from version 2.6.6 to 2.7.1. After making a thorough bisect, I pinpointed the introduction of the weird behaviour to this specific commit: [042a8c2](https://github.com/composer/composer/commit/042a8c212801aeac42b7a41b42cd1185ae28123a) (related to https://github.com/composer/composer/issues/11787).

Originally, this issue has been found while Composer is used in conjunction with a custom plugin: [nix-community/composer-local-repo-plugin](https://github.com/nix-community/composer-local-repo-plugin). Note, this issue cannot be reproduced (_pun intended!_) when using Composer < 2.7. 

<details>

<summary>Reproduce the issue with the Composer plugin</summary>

What makes this issue complicated is that the core of the problem involves a custom Composer plugin ([nix-community/composer-local-repo-plugin](https://github.com/nix-community/composer-local-repo-plugin)), which is essential for illustrating the encountered issue. This plugin plays a critical role in achieving total reproducibility for PHP/Composer-based projects within the Nix ecosystem (_introduced in https://github.com/NixOS/nixpkgs/pull/248184_).

The plugin works by generating a local Composer repository (https://getcomposer.org/doc/05-repositories.md#composer) from a project's `composer.json` and `composer.lock` files, including the `packages.json` file, then, update each package's `dist` information in `composer.lock` file to point to the local repository instead. This repository then serves as a `composer` repository type within `composer.json`. Our objective is to use this plugin to create a stable version of the files that will be of the `vendor` directory at the end (_using the Composer cache is not a valid option_). This stability is crucial for Nix, as it relies on predictable inputs (_in this case, the `composer` repository_) to calculate its hashes accurately. The development and implementation of this plugin were motivated by the necessity for predictable outputs to ensure the reproducibility of PHP/Composer based builds... and it was working super fine so far (_despite that I wish I could get rid of the plugin one day, ideas are welcome!_). The issue emerged following the aforementioned [commit](https://github.com/composer/composer/commit/042a8c212801aeac42b7a41b42cd1185ae28123a), affecting the plugin's operation and, consequently, the reproducibility of our PHP/Composer-based projects. This situation underscores the importance of that small plugin in the Nix ecosystem and the need for a resolution that maintains the integrity of our requirements. Find more details on how to install and use the plugin in the [README](https://github.com/nix-community/composer-local-repo-plugin?tab=readme-ov-file#usage) file of the project.

No Nix is needed to reproduce the issue locally, you can reproduce it:
1. By installing Composer >= 2.7.0
2. By installing the plugin: `composer global require nix-community/composer-local-repo-plugin`
3. By cloning the following [gist](https://gist.github.com/ab25340ea5fa46e3c2f49c1d9ed5eb71.git): `git clone https://gist.github.com/ab25340ea5fa46e3c2f49c1d9ed5eb71.git`
4. By running the shell script [`run.sh`](https://gist.github.com/drupol/ab25340ea5fa46e3c2f49c1d9ed5eb71#file-run-sh) from that gist

</details>

<details>

<summary>Reproduce the issue without the Composer plugin</summary>

Since many commands are required to reproduce the issue, I've set up a temporary repository at https://github.com/drupol/composer-issue-11850 containing the minimum required file to reproduce the issue. Follow the next steps to quickly reproduce it locally: 

1. Clone the repo: `git clone https://github.com/drupol/composer-issue-11850`
2. In that new directory, run the script: `./run.sh`

</details>

To summarize, this PR contains only 2 commits:
* The first commit fix (_or get rid?_) of the issue by adding a simple extra condition in the complex logic that generates the `composer.lock` file. I'm still a bit unsure of the proposed fix, this is the reason why I ask you to carefully review this. The fix add a simple checks on `$package->getDistType()` but maybe we should do further checks, perhaps we should also compare `$package->getDistUrl()` and `$package->getDistSha1Checksum()` and if any of those properties have changed. I had a first version implementing that, but I changed it until I have some feedback from you. 
* The second commit refactor that logic by just using early conditions to reduce the code indentation and give a better understanding on what is going on in the logic workflow. It is a no-op commit and totally opinionated, I can get rid of it if you don't like it.

Please, let me know what you think of that, as I'm willing to fix the issue in Nix and also contribute to making Composer better.

Thanks.

### DIFF
--- a/src/Composer/DependencyResolver/LockTransaction.php
+++ b/src/Composer/DependencyResolver/LockTransaction.php
@@ -15,7 +15,6 @@ namespace Composer\DependencyResolver;
 use Composer\Package\AliasPackage;
 use Composer\Package\BasePackage;
 use Composer\Package\Package;
-use Composer\Package\PackageInterface;
 use Composer\Pcre\Preg;
 
 /**
@@ -111,9 +110,9 @@ class LockTransaction extends Transaction
 
             $packages[] = $package;
 
-            // if we're just updating mirrors we need to reset references to the same as currently "present" packages' references to keep the lock file as-is
+            // if we're just updating mirrors we need to reset everything to the same as currently "present" packages' references to keep the lock file as-is
             if ($updateMirrors === true && !array_key_exists(spl_object_hash($package), $this->presentMap)) {
-                $this->resetDistSourceRefs($package);
+                $packages[count($packages) - 1] = $this->updateMirrorAndUrls($package);
             }
         }
 
@@ -121,11 +120,11 @@ class LockTransaction extends Transaction
     }
 
     /**
-     * Resets dist and source references to what they were in presentMap
+     * Try to return the original package from presentMap with updated URLs/mirrors
      *
-     * This does not reset references if the currently present package didn't have any, or if the type of VCS has changed
+     * If the type of source/dist changed, then we do not update those and keep them as they were
      */
-    private function resetDistSourceRefs(PackageInterface $package): void
+    private function updateMirrorAndUrls(BasePackage $package): BasePackage
     {
         foreach ($this->presentMap as $presentPackage) {
             if ($package->getName() !== $presentPackage->getName()) {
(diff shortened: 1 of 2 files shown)


drupol:
I just created a repository that shows the issue through a script, find it here: https://github.com/drupol/composer-issue-11850

Seldaek:
I'm not sure I grasp why the dist type changes, but I can see if it changed then yeah maybe we should not update it.. but then I'd still update the source reference I think. So probably need to make only the dist updates conditional on having the same dist type.

If the dist type changed we should probably also just call setSourceReference instead of setSourceDistReferences to ensure we leave the dist reference alone.

drupol:
Nice, glad this has been recognized as a bug.

Now, do you think we should do the following?

```php
$packageDist = sprintf(
    '%s%s%s',
    $package->getDistSha1Checksum(),
    $package->getDistType(),
    $package->getDistUrl()
);
$presentPackageDist = sprintf(
    '%s%s%s',
    $presentPackage->getDistSha1Checksum(),
    $presentPackage->getDistType(),
    $presentPackage->getDistUrl()
);

if ($presentPackage->getSourceReference() && $presentPackage->getSourceType() === $package->getSourceType() && $presentPackageDist === $packageDist) {
  // ... rest of the code
}
```

drupol:
(just rebased since you fixed the CI)

Seldaek:
> Now, do you think we should do the following?

no. Because if the dist reference changes, the url will likely be different, and the sha1 as well, so this does not mean anything.

drupol:
OK fair enough. Is there left to do in here?

Seldaek:
@naderman can you please review the approach here? Most important is the last commit. It switches the mirror update from: "take newest package and reapply the old references to it" to "keep old package but apply newest dist/source urls and mirrors to it". I think it's a more meaningful way to do this, and it should still achieve the desired goal.. But well it carries some risk of unintended consequences for sure.

drupol:
@Seldaek To avoid overriding what you did in this branch, I've created another branch. It contains a fix that makes this one works again, find it here: https://github.com/drupol/composer/pull/1

Seldaek:
Ok merging this as it fixes other issues and does not break nix more than it already was broken ;)

PR #11858, opened 2024-02-22
Add a warning message when Composer is not able to guess the root package version

drupol:
Hello,

I would like to propose this PR to help the user experience with Composer and such kind of issue: https://github.com/vimeo/psalm/issues/10721

Let me know if you would be OK with it.

### DIFF
--- a/doc/articles/troubleshooting.md
+++ b/doc/articles/troubleshooting.md
@@ -77,39 +77,37 @@ indirectly) back on the root package itself, issues can occur in two cases:
    the variable only for the call to composer, or you can define it globally in the
    CI env vars.
 
-## Root package version
+## Root package version detection
 
-Composer relies on accurately knowing the version of the root package to resolve
-dependencies effectively. It first attempts to identify the version of the root
-package using a hierarchical approach that prioritizes automatic detection but
-allows for manual overrides:
+Composer relies on knowing the version of the root package to resolve
+dependencies effectively. The version of the root package is determined
+using a hierarchical approach:
 
-1. **Version Control System (VCS) Inspection**: Composer attempts to guess the
-   version by interfacing with the version control system of the project. For
-   instance, in projects versioned with Git, Composer executes specific Git
-   commands to deduce the project's current version based on tags, branches, and
-   commit history.
+1. **composer.json Version Field**: Firstly, Composer looks for a `version`
+   field in the project's root `composer.json` file. If present, this field
+   specifies the version of the root package directly. This is generally not
+   recommended as it needs to be constantly updated, but it is an option.
 
-2. **Environment Variable**: Composer checks for the `COMPOSER_ROOT_VERSION`
+2. **Environment Variable**: Composer then checks for the `COMPOSER_ROOT_VERSION`
    environment variable. This variable can be explicitly set by the user to
    define the version of the root package, providing a straightforward way to
    inform Composer of the exact version, especially in CI/CD environments or
    when the VCS method is not applicable.
 
-3. **composer.json Version Field**: Lastly, Composer looks for a `version`
-   field in the project's root `composer.json` file. If present, this field
-   specifies the version of the root package directly.
+3. **Version Control System (VCS) Inspection**: Composer then attempts to guess
+   the version by interfacing with the version control system of the project. For
+   instance, in projects versioned with Git, Composer executes specific Git
+   commands to deduce the project's current version based on tags, branches, and
+   commit history. If a `.git` directory is missing or the history is incomplete
+   because CI is using a shallow clone for example, this detection may fail to find
+   the correct version.
 
-It's important to ensure that at least one of these methods successfully
-communicates the root package version to Composer. In scenarios where the
-version cannot be determined due to factors like a missing `.git` directory or
-the absence of explicit version information, Composer uses a fallback logic.
-This logic is designed to offer developers maximum flexibility and convenience
-across a wide range of development and deployment environments. In the absence
-of a determinable version, Composer defaults to assigning a version of `1.0.0`.
-Note that this default version might potentially lead to dependency resolution
-issues, especially when the root package depends on a package which ends up
-depending (directly or indirectly) back on the root package itself.
+4. **Fallback**: If all else fails, Composer uses `1.0.0` as default version.
+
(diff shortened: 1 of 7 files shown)


src/Composer/Package/Loader/RootPackageLoader.php
```diff
@@ -93,6 +99,15 @@ public function load(array $config, string $class = 'Composer\Package\RootPackag
             }
 
             if (!isset($config['version'])) {
+                if ($this->io !== null) {
```
Seldaek, line 102: If no name is set, circular dependencies are impossible so warning about this is probably irrelevant - and as a nice side effect it'll fix most tests I think.

```suggestion
                if ($this->io !== null && $config['name'] !== '__root__') {
```

Seldaek:
I'll need to review again the wording later, but I think it would be nice to have a shorter message with a shortlink to the docs pointing to https://getcomposer.org/doc/articles/troubleshooting.md#dependencies-on-the-root-package or a new entry there with some more details about auto-detection not working. I'm not a fan of the 3-lines warning for something which may not affect people at all.

drupol:
Fair enough, I don't like it either, but it was hard to make a one liner out of it. I'll fix it tomorrow morning.

src/Composer/Package/Loader/RootPackageLoader.php
```diff
@@ -93,6 +99,14 @@ public function load(array $config, string $class = 'Composer\Package\RootPackag
             }
 
             if (!isset($config['version'])) {
+                if ($this->io !== null && $config['name'] !== '__root__') {
+                    $this->io->warning(
+                        sprintf(
+                            "Composer couldn't find the root package version for %s, defaulting to `1.0.0`. See https://getcomposer.org/doc/articles/troubleshooting.md#dependencies-on-the-root-package for more details.",
```
Seldaek, line 105: Let's use a shortcut here, I'll make sure it works later.

```suggestion
                            "Composer couldn't find the root package version for %s, defaulting to `1.0.0`. See https://getcomposer.org/root-version",
```

src/Composer/Package/Loader/RootPackageLoader.php
```diff
@@ -93,6 +99,14 @@ public function load(array $config, string $class = 'Composer\Package\RootPackag
             }
 
             if (!isset($config['version'])) {
+                if ($this->io !== null && $config['name'] !== '__root__') {
+                    $this->io->warning(
+                        sprintf(
+                            "Composer couldn't find the root package version for %s, defaulting to `1.0.0`. See https://getcomposer.org/root-version",
```
Seldaek, line 105: Sorry you were too quick I was editing :D 

```suggestion
                            "Composer could not detect the root package (%s) version, defaulting to "1.0.0". See https://getcomposer.org/root-version",
```

doc/articles/troubleshooting.md
```diff
@@ -71,11 +71,13 @@ indirectly) back on the root package itself, issues can occur in two cases:
...
    The best solution is to define the version you are on via an environment variable
-   called COMPOSER_ROOT_VERSION. You set it to `dev-main` for example to define
+   called `COMPOSER_ROOT_VERSION`. You set it to `dev-main` for example to define
    the root package's version as `dev-main`.
    Use for example: `COMPOSER_ROOT_VERSION=dev-main composer install` to export
    the variable only for the call to composer, or you can define it globally in the
-   CI env vars.
+   CI env vars. Alternatively, adding a `version` field directly in the
+   `composer.json` file specifying the current version can also resolve the
+   issue.
```
Seldaek, line 80: We could add a section below specific to the issue of the version not being able to be guessed, which may link to the above as possible solution but also maybe contain more infos about the cause like an incomplete git history (sparse checkouts) or completely missing git data.

src/Composer/Package/Loader/RootPackageLoader.php
drupol: Going to fix the typo

doc/articles/troubleshooting.md
drupol: Good idea ! Will do it.

src/Composer/Package/Loader/RootPackageLoader.php
```diff
@@ -93,6 +99,14 @@ public function load(array $config, string $class = 'Composer\Package\RootPackag
             }
 
             if (!isset($config['version'])) {
+                if ($this->io !== null && $config['name'] !== '__root__') {
+                    $this->io->warning(
+                        sprintf(
+                            "Composer could not detect the root package (%s) version, defaulting to "1.0.0". See https://getcomposer.org/root-version",
```
Seldaek, line 105: Ah oops quotes.. Anyway with this message I'm happy I think, if you like to update the broken tests (although you could also just set a version in those tests to avoid the warning being output, it's probably cleaner).

```suggestion
                            'Composer could not detect the root package (%s) version, defaulting to "1.0.0". See https://getcomposer.org/root-version',
```

gitressa:
Many Drupal users are seeing the warning `Composer could not detect the root package (drupal/recommended-project) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version` but it's not clear what the problem is and how to fix it ...
[Root cause of 'could not detect the root package'?](https://www.drupal.org/forum/support/composer/2024-03-13/root-cause-of-could-not-detect-the-root-package)

Is there a command which can help locate the cause of this message?

drupol:
Just wait for the next point release, it has been fixed in https://github.com/composer/composer/pull/11885

gitressa:
Wonderful, thanks @drupol for a fast reply and clarifying.

drupol:
You're very welcome!

begrafx:
When can we expect this new point release? Looks like I'm running version 2.7.2. a `composer self-update` says I'm using the latest available version, but I'm still getting the `Composer could not detect the root package (drupal/recommended-project) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version` message. So I'm presuming the update on this is still forthcoming?

Seldaek:
Yes use --snapshot for now, I'll try my best to get a release out next week tho

Seldaek:
[2.7.3](https://github.com/composer/composer/releases/tag/2.7.3) is available now.

PR #11866, opened 2024-02-28
Include PHP information when showing Composer version verbosely

miken32:
I spend a lot of time on Stack Overflow, and confusion about which version of PHP being used is a big driver of questions. This PR adds that information when calling `composer --version --verbose`. e.g.:
```none
Composer version 2.7.999-dev+source @release_date@
PHP version 8.3.2 (/usr/local/bin/php83)
```
I would prefer if it were shown by default, but I figured modifying any existing behaviour would reduce the chance of acceptance.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -608,12 +608,18 @@ class Application extends BaseApplication
             $branchAliasString = sprintf(' (%s)', Composer::BRANCH_ALIAS_VERSION);
         }
 
+        $phpVersionString = '';
+        if ($this->getIO()->isVerbose()) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Yeah this is typically better shown via `composer diagnose` IMO as it includes more details and checks.

I'm unsure about showing it in --version output by default as I'm worried it messes with people trying to parse the version string, but I think I see a way to do this without risk of side-effects. I'll update your PR.

Seldaek:
Alright I think this should do :) Thanks!

stof:
you should be checking the exit code to detect failures, not the output to the reporting stream. Logging this to stdout would break tools that process the output of the version.

Issue #11863, opened 2024-02-26
--ignore does not ignore package


NiklasBr:
https://github.com/composer/composer/blob/d7cdb28b28d8a9ff22a238fe85b903f01d3a16ce/doc/03-cli.md?plain=1#L600

The ability to use `--ignore` was a very welcome one since the server hosting `foo/somepackage` is not always available over the network and the package has no limiting constraints to deal with.

However passing the above option like `compose outdated --ignore=foo/somepackage` will still attempts to connect to the server hosting `foo/somepackage` to download info on it with _Composer version 2.7.1 2024-02-09 15:26:28_.

Result is that Composer fails to report anything as outdated.

stof:
Composer repositories are not mapping 1:1 to package names (and even in case they contain a single package name, the package name is not known before loading it). So there is no way to transform an ignore rule on a package into an ignore rule on the loading of repository metadata.

NiklasBr:
We are mapping this package 1:1 however using the `repositories` section and `only`:

```json
  "repositories": {
    "satis": {
      "type": "composer",
      "url": "https://satis.example.com/",
      "only": [
        "foo/somepackage"
      ]
    }
  }
```

Seldaek:
Yeah sorry but that's not something we can reasonably fix.

NiklasBr:
Sorry for a stupid follow-up question, in that case I don't understand the documentation. I thought the argument would cause the packager to be ignored, but it is not?

Seldaek:
It is ignored from the output, but the metadata is fetched for all packages from all repos first. But on second thought this could actually be optimized away earlier instead of just on display, which would indeed fix the issue for you I believe. 

Around ShowCommand line 480 we need to check `!Preg::isMatch($ignoredPackagesRegex, $package->getPrettyName())` before fetching the latestPackage. Feel free to give it a shot and send a PR, otherwise I'll get to it eventually.

NiklasBr:
Thanks a lot! I will see what I can do. Meanwhile I'm making a docs clarification PR, hope that's okay.

NiklasBr:
@Seldaek what if instead of changing the behaviour of this `--ignore` option I try to introduce another one called `--ignore-package-reqs` which work similarly to `--ignore-platform-reqs` or would that be very complex?

Seldaek:
Please wait until monday I'll try to get a fix for the initial issue at least as i understood it.. Then if that doesn't work we can discuss further but i think you're over complicating it right now

[...]

PR #11868, opened 2024-03-01
Clarify --ignore option

NiklasBr:
Just something which came out of #11863. The motivation is that `--ignore-platform-req` actually ignores things when it comes to version calculations, so the other `--ignore` option should better specify that it is just the returned output which is adjusted.


Seldaek:
This should be irrelevant now but thanks anyway :)

Issue #11874, opened 2024-03-06
Mercurial repository type doesn't allow for SSH URLs


gaxweb:
We would like to use https://www.codebasehq.com/ to deploy our private Mercurial repositories via SSH and I've run into an issue here. If I use the `hg` repository type in the `composer.json`, I get the error:

```
In Hg.php line 90:
  Failed to clone ssh://hg@codebasehq.com/gaxweb/projectname/repository.hg,
  The given URL (…) does not match the required format (http(s)://(username:password@)example.com/path-to-repository)  
```   

HG URLs can start with "ssh" in general, so why is the `hg` type checking, if it starts with `http`? I can connect just fine, if I use the `vcs` type in the `composer.json` (but I have to enter the ssh key password 3 times) after all. So it's weird that it works one way but not the other.

Is this intentional or just an oversight? I could try my hand at a pull request. I think the problem lies here:

https://github.com/composer/composer/blob/66acb84c12d1d2b3cce7b2967f8e25522f502e60/src/Composer/Util/Hg.php#L61

stof:
Mercurial repositories are much less used in Composer than Git repositories. So it would not surprise me if the support is incomplete (I suspect that support for Mercurial was mostly tested with Bitbucket in the past, and they supported the `https` scheme so this was not an issue).

gaxweb:
CodebaseHQ does support them, too, but the way repository user permissions are implemented means it makes more sense to use SSH for private repositories.

It's not a pressing issue because the `vcs` type works, as I said. But it looks like an omission to me, so I created this issue.

stof:
Feel free to create a PR fixing the issue.

Seldaek:
Fixed by #11878

PR #11878, opened 2024-03-08
Allow for SSH URLs when using hg repository type

gaxweb:
This fixes ticket #11874

It was declared a feature, so I've chosen the main branch. I also applied a few minor fixups to docblocks for example. I hope that's ok, otherwise let me know and I'll remove them.

For the fix itself, I basically just adjusted the regex to allow "ssh" and to grab the password more reliably, even though the latter isn't even used. Worked fine with the repo I'm using, if the SSH key is set up in both the repo and the workstation you're issuing the pull from.

### DIFF
--- a/src/Composer/Util/Hg.php
+++ b/src/Composer/Util/Hg.php
@@ -58,16 +58,19 @@ class Hg
         }
 
         // Try with the authentication information available
-        if (Preg::isMatch('{^(ssh|https?)://(([^:@]+):?([^:@]+)?@)?([^/]+)(/.*)?}mi', $url, $matches)
-            && $this->io->hasAuthentication((string) $matches[5]))
-        {
-            list($fullURL, $protocol, $credentials, $user, $password, $domain, $path) = $matches;
-            if ($protocol === 'ssh') {
-                if($user !== '') $user = rawurlencode((string)$user) . '@';
-                $authenticatedUrl = $protocol.'://'.$user.$domain.$path;
+        if (
+            Preg::isMatch('{^(?P<proto>ssh|https?)://(?:(?P<user>[^:@]+)(?::(?P<pass>[^:@]+))?@)?(?P<host>[^/]+)(?P<path>/.*)?}mi', $url, $matches)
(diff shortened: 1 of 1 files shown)


src/Composer/Util/Hg.php
```diff
@@ -93,6 +102,7 @@ private function throwException($message, string $url): void
     /**
      * Retrieves the current hg version.
      *
+     * @param ProcessExecutor $process
```
stof, line 105: this is useless as it brings no additional information compared to the native signature.

src/Composer/Util/Hg.php
```diff
@@ -70,21 +77,23 @@ public function runCommand(callable $commandCallable, string $url, ?string $cwd)
...
-        $this->throwException('Failed to clone ' . $url . ', ' . "\n\n" . $error, $url);
+        $this->throwException("Failed to clone $url, \n\n" . $error, $url);
     }
 
     /**
      * @param non-empty-string $message
-     *
-     * @return never
+     * @param string $url
+     * @return never-returns
```
stof, line 89: this should be reverted. `never` is the native name for this type. there is no reason to use a phpstan-only alias of this type name.

src/Composer/Util/Hg.php
```diff
@@ -70,29 +77,29 @@ public function runCommand(callable $commandCallable, string $url, ?string $cwd)
...
+                'Failed to clone ' . $url . ', hg was not found, check that it is installed and in your PATH env.' . "\n\n" . $this->process->getErrorOutput()
+            ));
         }
 
         throw new \RuntimeException(Url::sanitize($message));
     }
 
     /**
      * Retrieves the current hg version.
-     *
```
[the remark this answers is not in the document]
fredden: The official documentation says there should be a blank line after the summary and before the tags. https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#5-the-phpdoc-format

Seldaek:
Ok thanks for the PR. I did some more changes to the regex, hopefully you can check with latest composer snapshot that this is still working well for you.

PR #11880, opened 2024-03-10
Update scripts.md

8ig8:
Updates the 'Setting environment variables' example to call Composer using `@composer` in order to automatically resolve to whatever composer.phar is currently being used as detailed in the previous section on this same page.


Seldaek:
Thanks

PR #11885, opened 2024-03-12
Only show warning about default version when not "project" type

Toflar:
https://github.com/composer/composer/pull/11858 is too broad imho because for a lot of project skeletons that you'd use `create-project` for, this will now show a warning. However, if you are of `type` `project`, it's extremely unlikely  ou have circular deps on yourself then. The warning only confuses people in this case.


fritzmg (approved):

aldilop (approved):

Seldaek:
Thanks

PR #11888, opened 2024-03-15
[PHP 8.4] Fix for implicit nullability deprecation

Ayesh:
Fixes a issue that emits a deprecation notice on PHP 8.4.

See:
 - [RFC](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types)
 - [PHP 8.4: Implicitly nullable parameter declarations deprecated](https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated)

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -298,7 +298,7 @@ class Git
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

PR #11889, opened 2024-03-15
[status command]check if package exists before compare

eltharin:
If tag has been deleted from remote, download function will not create $targetDir . '_compare' folder and that make an Error : chdir(): No such file or directory (errno 2).
I suggest this update for prevent error and print to user that the reference disapear.

### DIFF
--- a/src/Composer/Downloader/FileDownloader.php
+++ b/src/Composer/Downloader/FileDownloader.php
@@ -500,12 +500,16 @@ class FileDownloader implements DownloaderInterface, ChangeReportInterface
             $this->install($package, $targetDir.'_compare', false);
             $this->process->wait();
 
-            $comparer = new Comparer();
-            $comparer->setSource($targetDir.'_compare');
-            $comparer->setUpdate($targetDir);
-            $comparer->doCompare();
-            $output = $comparer->getChangedAsString(true);
-            $this->filesystem->removeDirectory($targetDir.'_compare');
+            if (is_dir($targetDir.'_compare')) {
+                $comparer = new Comparer();
+                $comparer->setSource($targetDir.'_compare');
+                $comparer->setUpdate($targetDir);
+                $comparer->doCompare();
+                $output = $comparer->getChangedAsString(true);
+                $this->filesystem->removeDirectory($targetDir.'_compare');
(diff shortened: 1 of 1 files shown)


src/Composer/Downloader/FileDownloader.php
```diff
@@ -500,12 +500,16 @@ public function getLocalChanges(PackageInterface $package, string $path): ?strin
...
-            $this->filesystem->removeDirectory($targetDir.'_compare');
+            if (is_dir($targetDir.'_compare')) {
+                $comparer = new Comparer();
+                $comparer->setSource($targetDir.'_compare');
+                $comparer->setUpdate($targetDir);
+                $comparer->doCompare();
+                $output = $comparer->getChangedAsString(true);
+                $this->filesystem->removeDirectory($targetDir.'_compare');
+            } else {
+                $output = $package . ' can\'t be downloaded';
```
Seldaek, line 511:
```suggestion
                $output = $package . ' could not be downloaded, aborting.';
```

eltharin:
I think it's better too, but I test your changes, and I ever had an error with my example, it try to download zipball witch does'nt exists (all branche's ids are changes or repo does'nt exists and I get an exception 
``[Composer\Downloader\TransportException (404)]
  The "https://gitserver.local/repos/eltharin/jsetable/zipball/ce53cf1de8bb53a4d45665f5d44096ec87baeb66" file
  could not be downloaded (HTTP/1.1 404 Not Found)``

That break the status and I have no other informations about other packages.

Seldaek:
Ok yeah sure it fails hard, maybe we should return the error nicer to let the others proceed. I'll take a look.

Seldaek:
Alright fixed that, feel free to try again.

eltharin:
Awesome, I test and it’s good,

Thanks

De : Jordi Boggiano ***@***.***>
Envoyé : jeudi 21 mars 2024 11:18
À : composer/composer ***@***.***>
Cc : eltharin ***@***.***>; Author ***@***.***>
Objet : Re: [composer/composer] [status command]check if package exists before compare (PR #11889)


Alright fixed that, feel free to try again.

Issue #11891, opened 2024-03-15
Surface remote IDs for advisories without a CVE so they can be ignored


bradjones1:
Running `composer audit` may find issues which have no related CVE (for whatever reason, e.g. it was not requested or the package is no longer supported at your installed version, so it doesn't get a CVE.)

In some cases you may wish to ignore these (not looking to debate the merits of ignoring some findings - that bikeshed has already been painted) but since there is no CVE to match, you can't.

This is not too dissimilar to other toolchains (e.g., [this wrapper around `yarn audit`](https://github.com/djfdyuruiry/improved-yarn-audit?tab=readme-ov-file#excluding-advisories)) but I think this is a potential area of improvement.

The current format for the `config.audit.ignore` entry doesn't seem to support any namespacing for whole packages, just vulnerability IDs. In this case, the package is from Drupal, which doesn't always pull a CVE.

Is this something that we'd want to support a solution for in Composer? I can and will raise an issue at Drupal to include "something" unique in the CVE field, since this is basically freetext and there are other authorities such as `GHSA` already. But it's entirely possible others will have a similar requirement and can't get the source to make this non-`null`.

bradjones1:
Issue for issuing a namespaced CVE equivalent from Drupal: https://www.drupal.org/project/infrastructure/issues/3428660

bradjones1:
Update - turns out you can filter out by remote ID, but that information is only available over the API and isn't surfaced in the default table render of advisories. I think that would be where to fix this.

https://github.com/composer/composer/pull/11556/files#diff-0663da0b66c02ac641cf3dbe00262e15b105fb4fb1a3bf2696530895ea7c85adR82

PR #11892, opened 2024-03-16
Surface the advisory ID when CVE not present.

bradjones1:
Fixes https://github.com/composer/composer/issues/11891


Seldaek:
Thanks

PR #11895, opened 2024-03-19
HttpDownloader: add option to prevent access to private network

glaubinix:
Adds an option to prevent access to the private network, similar to how Symfony does it in the NoPrivateNetworkHttpClient. 
See https://github.com/symfony/symfony/blob/a9133af83ef26f0edaceca38aa885bb4f4d25a19/src/Symfony/Component/HttpClient/NoPrivateNetworkHttpClient.php#L29


src/Composer/Util/RemoteFilesystem.php
```diff
@@ -249,6 +249,10 @@ protected function get(string $originUrl, string $fileUrl, array $additionalOpti
 
         $origFileUrl = $fileUrl;
 
+        if (isset($options['prevent_ip_callable'])) {
```
Seldaek, line 252:
```suggestion
        if (isset($options['prevent_ip_access_callable'])) {
```

src/Composer/Util/RemoteFilesystem.php
```diff
@@ -249,6 +249,10 @@ protected function get(string $originUrl, string $fileUrl, array $additionalOpti
...
+            throw new \RuntimeException("RemoteFilesystem doesn't support the 'prevent_ip_callable' config.");
```
Seldaek, line 253:
```suggestion
            throw new \RuntimeException("RemoteFilesystem doesn't support the 'prevent_ip_access_callable' config.");
```

Issue #11876, opened 2024-03-07
`http_proxy` is used for `https://` URLs


cs278:
After configuring the `http_proxy` environment variable to proxy plain HTTP requests Composer starts failing because it uses the proxy server for encrypted HTTP requests. Whilst I appreciate there is no defined standard, I'm pretty sure Composer is deviating from the convention here but I could be wrong.

Output of `composer diagnose`:

```
$ http_proxy=http://10.0.17.44:80 php -f bin/composer diagnose
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK git version 2.43.2
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: FAIL
[Composer\Downloader\TransportException] curl error 56 while downloading https://repo.packagist.org/packages.json: Received HTTP code 400 from proxy after CONNECT
Checking HTTP proxy: FAIL
[Composer\Downloader\TransportException] curl error 56 while downloading https://repo.packagist.org/packages.json: Received HTTP code 400 from proxy after CONNECT
Checking github.com oauth access: FAIL
[Composer\Downloader\TransportException] curl error 56 while downloading https://api.github.com/: Received HTTP code 400 from proxy after CONNECT
Checking disk free space: OK
Checking Composer and its dependencies for vulnerabilities: WARNING
Failed performing audit: curl error 56 while downloading https://packagist.org/api/security-advisories/: Received HTTP code 400 from proxy after CONNECT
Composer version: 2.7.999-dev+source
PHP version: 7.2.5 - Package overridden via config.platform, actual: 7.2.34
PHP binary path: /usr/bin/php7.2
OpenSSL version: OpenSSL 3.0.2 15 Mar 2022
cURL version: 7.81.0 libz 1.2.11 ssl OpenSSL/3.0.2
zip: extension present, unzip present, 7-Zip present (7z)
```

curl/wget both seem to operate as I expect:

```
$ http_proxy=http://10.0.17.44:80 curl -I http://repo.packagist.org/packages.json
HTTP/1.1 200 OK
Server: nginx/1.18.0 (Ubuntu)
Date: Thu, 07 Mar 2024 18:42:39 GMT
Content-Type: application/json
Connection: keep-alive
Last-Modified: Thu, 07 Mar 2024 18:33:07 GMT
Vary: Accept-Encoding
ETag: W/"65ea0863-482"

$ http_proxy=http://10.0.17.44:80 curl -I https://repo.packagist.org/packages.json
[...]

PR #11915, opened 2024-04-03
Refactor proxy handling to require https_proxy

johnstevenson:
Composer has always allowed a single http_proxy (or CGI_HTTP_PROXY) environment variable to be used for both HTTP and HTTPS requests. But many other tools and libraries require scheme-specific values.

The landscape is already complicated by the use of and need for upper and lower case values, so to bring matters inline with current practice https_proxy is now required for HTTPS requests.

The new proxy handler incorporates a transition mechanism, which allows http_proxy to be used for all requests when https_proxy is not set and provides a `needsTransitionWarning` method for the main application.

Moving to scheme-specific environment variables means that a user may set a single proxy for either HTTP or HTTPS requests. To accomodate this situation during the transition period, an https_proxy value can be set to an empty string which will prevent http_proxy being used for HTTPS requests.

**Summary**
- Everything works as before
- If a user has a single http_proxy and it is used for HTTPS requests, a warning message is shown after any other output:

```txt
Composer now requires separate proxy environment variables for HTTP and HTTPS requests.
Please set `https_proxy` in addition to your existing proxy environment variables.
```
**Implementation**
The refactoring was such that it was easier to replace the old `ProxyManager` class with a new `ProxyHandler`. In part this was because the original implementation was very rushed to meet the Composer 2 deadline and some unused artifacts from a last minute refactoring of that remained.

There are two small differences:

1. Previously a CGI_HTTP_PROXY value overwrote any http_proxy value when in a CLI environment. Now CGI_HTTP_PROXY (and also cgi_http_proxy which is something Ruby now uses) is only examined if no http_proxy was found. I cannot see this making a difference.

2. The previous `getProxyForRequest` method obtained the url scheme from `parse_url($requestUrl, PHP_URL_SCHEME) ?: 'http'`, which had the unfortunate effect of turning a `false` value into http (like when Windows file urls are used in a couple of tests). The latest version uses `(string) parse_url($requestUrl, PHP_URL_SCHEME)`. My concern is that valid request urls without a scheme are now not converted to http and will bypass any proxy. It is likely there are such urls?

The warning message is only shown if a command makes an HTTPS request, but it could easily be changed to show after every command.

This is a follow up to discussions on #11876


johnstevenson:
Apologies, for the forced-push, but I've updated this to remove some bits that I thought might be needed for the Diagnose command, which it turns out were not. I will provide a separate PR for that later, to bring it up to date with these changes.

I've also fixed some code that was changed in `CurlDownloader::initDownload()` as part of some PHPStan improvements in 8600894003. This changed setting CURLOPT_PROXY with an empty proxy url. However, this is needed because it prevents curl from reading any proxy environment variables (which we have already handled).

The CI tests are now failing on PHP 8.3 and below because StatusCommandTest::testLocallyModifiedPackages() fetches https://www.smarty.net/files/Smarty-3.1.7.zip and that domain name expired on 2024-04-05 is now parked.

Seldaek:
Oh boy.. I thought it was going to be a reasonably small fix 😅  Thanks, but it will take me some time to review.

johnstevenson:
> I thought it was going to be a reasonably small fix

I did try multiple ways to keep the noise down in the diff, but to no avail.

Perhaps we could do it in the same way we reworked the installer script years ago, by breaking it down into a series of PRs that add functionality by stages?

As a first step, I could provide a new PR that simply adds the new ProxyHandler and ProxyItem classes without plumbing them in, which would make reviewing the functionality much easier.

I don't mind doing this, because I already have a some improvements on this PR yet to push, and in retrospect I think it would make your life easier.

>  it will take me some time to review.

There's no great rush here to add this change, but please let me know if you would prefer the above approach.

Seldaek:
No i think it's fine, i can review those new files in isolation from the other changes anyway. Feel free to add whatever is needed here. I won't look until monday earliest

johnstevenson:
5325ed9 changes how userinfo is set for curl by stripping it out of the proxy url and adding it with the `CURLOPT_PROXYUSERPWD` option. This is done because  some curl versions will show userinfo in the error message when given a malformed url.

It also takes the choice of the curl proxy options out of CurlDownloader and lets RequestProxy provide them as an array.

src/Composer/Util/Http/ProxyHandler.php
```diff
@@ -0,0 +1,212 @@
...
+namespace Composer\Util\Http;
+
+use Composer\Downloader\TransportException;
+use Composer\Util\NoProxyPattern;
+
+/**
+ * @internal
+ * @author John Stevenson <john-stevenson@blueyonder.co.uk>
+ */
+class ProxyHandler
```
Seldaek, line 22: I understand this is a full rewrite here but can we maybe keep the ProxyManager name? Or what's the reasoning for changing it (besides the fact I'm sure it was easier for you while rewriting side by side?).

src/Composer/Util/Http/ProxyHandler.php
```diff
@@ -0,0 +1,212 @@
...
+    /** @var ?string */
+    private $error = null;
+    /** @var ?ProxyItem */
+    private $httpProxy = null;
+    /** @var ?ProxyItem */
+    private $httpsProxy = null;
+    /** @var ?NoProxyPattern */
+    private $noProxyHandler = null;
+
+    /** @var ?ProxyHandler */
```
Seldaek, line 33:
```suggestion
    /** @var ?self */
```

src/Composer/Console/Application.php
```diff
@@ -398,6 +399,12 @@ function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknow
                 $io->writeError('<info>Memory usage: '.round(memory_get_usage() / 1024 / 1024, 2).'MiB (peak: '.round(memory_get_peak_usage() / 1024 / 1024, 2).'MiB), time: '.round(microtime(true) - $startTime, 2).'s');
             }
 
+            if (ProxyHandler::getInstance()->needsTransitionWarning()) {
+                $io->writeError('');
+                $io->writeError('<warning>Composer now requires separate proxy environment variables for HTTP and HTTPS requests.</warning>');
+                $io->writeError('<warning>Please set `https_proxy` in addition to your existing proxy environment variables.</warning>');
```
Seldaek, line 405:
```suggestion
                $io->writeError('<warning>Please set `https_proxy` in addition to your existing proxy environment variables.</warning>');
                $io->writeError('<warning>This fallback (and warning) will be removed in Composer 2.8.0.</warning>');
```

src/Composer/Util/Http/ProxyHandler.php
```diff
@@ -0,0 +1,212 @@
...
+     * Searches $_SERVER for case-sensitive values
+     *
+     * @return array{0: string|null, 1: string} value, name
+     */
+    private function getProxyEnv(string $envName): array
+    {
+        $names = [strtolower($envName), strtoupper($envName)];
+
+        foreach ($names as $name) {
+            if (is_string($_SERVER[$name] ?? null)) {
```
Seldaek, line 186: Please use Platform::getEnv here, which will return false if the var is unset

src/Composer/Util/Http/ProxyHandler.php
```diff
@@ -0,0 +1,212 @@
...
+                $info = 'excluded by no_proxy';
+            } else {
+                $url = $proxy->getProxyUrl();
+                $auth = $proxy->getCurlAuth();
+                $options = $proxy->getContextOptions($scheme);
+                $info = $proxy->getSafeUrl();
+            }
+        }
+
+        return new RequestProxy($url, $auth, $options, $info);
```
Seldaek, line 104: IMO this would read better with a few factory methods hiding all the copying around of stuff:

```suggestion
        $proxy = $this->getProxyForScheme($scheme);

        if ($proxy === null) {
            return RequestProxy::none();
        }

        if ($this->noProxy($requestUrl)) {
            return RequestProxy::noProxy();
        }

        return RequestProxy::fromProxyItem($scheme);
```

src/Composer/Util/Http/ProxyItem.php
```diff
@@ -0,0 +1,151 @@
...
+    /** @var string */
+    private $optionsProxy;
+    /** @var ?non-empty-string */
+    private $optionsAuth;
+
+    public function __construct(string $proxy, string $envName)
+    {
+        if (strpbrk($proxy, "\r\n\t") !== false || !$this->checkData($proxy)) {
+            throw new \RuntimeException(sprintf('unsupported `%s` syntax', $envName));
+        }
```
Seldaek, line 38: Maybe move this check to ProxyHandler::getProxyEnv, then it can simply return $env and no name, and this class doesn't need $envName either.

Seldaek: Ah no sorry you kinda need that as checkData can return false too. But I'd still rather move this whole check in checkData, or just move checkData body to be the constructor body, so you can throw proper exception message withw hat is unsupported instead of just a generic throw when checkData returns false.

Seldaek (changes requested):
Ok done reviewing, beyond the few small code changes which IMO would improve readability (I struggled in a few places, but eventually grasped things, it's just code style mostly I think..) I think it looks good to me, thanks!

src/Composer/Util/Http/ProxyHandler.php
johnstevenson: Yes we can keep the name. The reasoning was exactly as you suggest from a rewriting perspective, but also to avoid a messy ProxyManager diff from a reviewing perspective. Of course, this hasn't really worked given the overall changes!

src/Composer/Util/Http/ProxyHandler.php
johnstevenson: Yes, will fix.

src/Composer/Console/Application.php
johnstevenson: > it just behaves as it did before if only http_proxy is set, correct?

Yes, that is correct. I'll add the additional warning.

Incidentally, I intend to submit a PR for a new `how-to-use-composer-behind-a-proxy` FAQ, which will detail the requirements and replace the info from the environment variables section (with a suitable reference). This could perhaps be added to the warning message later, if it is accepted.

src/Composer/Util/Http/ProxyHandler.php
johnstevenson: We have always used $_SERVER because it provides case-sensitive matching, which is crucial, while `getenv` (used by Platform::getEnv, although only as a fallback) does not. Additionally getenv would need to use the `local` param in order not to evaluate SAPI specific environment variables, though this wouldn't fix the case-insensitivity.

I'd prefer to keep this as it is, in house (as it were).

src/Composer/Util/Http/ProxyHandler.php
johnstevenson: Neat. I really like this and will implement.

src/Composer/Util/Http/ProxyHandler.php
Seldaek: Alright then leave it as is.

src/Composer/Console/Application.php
Seldaek: Sounds good to have better docs on this 👍🏻

Seldaek (approved):
Looks good to me now, thanks! Anything else coming or good to merge for you as well?

johnstevenson:
Yes, it's good to merge for me. I've got a Diagnose update and a docs update but I was going to PR them separately.

Seldaek:
Nah it's fine as is I think for the warning. And yes separate PRs later for the rest sounds good, also would appreciate a PR to clean up the transitional stuff for 2.8 but it's not urgent.

I'll merge this then, thanks again

Issue #11917, opened 2024-04-04
[Bitbucket] [AccessTokenAuth] File name too long


p24-max:
<!--
The information we ask for below is IMPORTANT. It will help us help you faster.
Please FOLLOW THE ISSUE TEMPLATE unless you have a good reason not to, and help avoid wasting everyone's time.

Please also make sure you are using the latest Composer version. You can get the latest with `composer self-update`
or even try the latest dev build with `composer self-update --snapshot` to see if the bug was perhaps already fixed but not released yet.
-->

My `composer.json`:

(secret-key is redacted with .)

```
[...]
  "repositories": {
        "filter-cms" : {
            "type": "vcs",
            "url": "https://x-token-auth:ATCT.........................................................................................................................................................................................8FC@bitbucket.org/the-namespace/module-filter-cms.git"
        }
  }
[...]
```

Output of `composer diagnose`:

Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK git version 2.39.3
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking Composer version: OK
Checking Composer and its dependencies for vulnerabilities: OK
Composer version: 2.7.2
PHP version: 8.1.26
PHP binary path: /opt/homebrew/Cellar/php@8.1/8.1.26/bin/php
OpenSSL version: OpenSSL 3.2.0 23 Nov 2023
cURL version: 8.5.0 libz 1.2.12 ssl (SecureTransport) OpenSSL/3.2.0
zip: extension present, unzip present, 7-Zip not available


When I run this command: <!-- run it with `-vvv` added to it ideally to get full debug output -->

php composer.phar require the-namespace/module-filter-cms:dev-master

I get the following output: <!-- FULL OUTPUT please, not just what you think is relevant -->

```
  [RuntimeException]                                                                                                                                                                                                                          
[...]

Issue #7313, opened 2018-05-05
package name too long


pelallemant:
I have a composer.json which requires a custom git project which has the following composer.json .

The gitlab group and the name of the package is a bit long, (about 200 characters).

My `composer.json` of the required package:

```json
{
  "name": "mySuperLongProjectNameWhich-IsSuperSuperLongABCDEFGHIJK"
}
```
When I run this command:

```
composer update
```

I get the following output:

```
 [RuntimeException]                                                                                                                                                                                                                                                                                               
  Failed to execute git clone --mirror 'git@gitlab.mydomain.com:thisisagitlabgroupwhichislong/this-is-a-project-of-100-characters-length.git' '/home/pierre-emmanuel/.composer/cache/vcs/git-gitlab.mydomain.com-this-is-a-project-of-100-characters-length.git/'                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                   
  Cloning into bare repository '/home/pierre-emmanuel/.composer/cache/vcs/git-gitlab.mydomain-coding.com-thisisagitlabgroupwhichislong-this-is-a-project-of-100-characters-length.git'...                                                                                 
  /home/pierre-emmanuel/.composer/cache/vcs/git-gitlab.mydomain.com-thisisagitlabgroupwhichislong-this-is-a-project-of-100-characters-length.git: File name too long               
```

And I expected this to happen:

The composer update isn't expected to crash, but the git command crash .
The problem seems to come from git.

In case we can't increase the limit for git, could it be possible to have a registry.json in the .composer/cache to store the correspondances between packages cloned and in the cache, and generate a hash instead of the fullname of the git repo, to reduce the length of the cloned git.

pelallemant:
Linux filename limit is 255 characters, but when the /home folders are encrypted it is reduced to 143...
I assume it's a good practice to encrypt the home folder to avoid being stolen the data with a livecd for example.

Since in the .composer/cache/vcs/ the repo name is prefixed with `git-$gitlabserver.com-$group-$projectName`` and reach 160 characters, I'm < 255 but > 143 so it won't work.

https://askubuntu.com/questions/728465/how-can-i-increase-ubuntus-143-byte-file-name-limit-for-encrypted-directories for explanations about the limitation.

The names of my projects are a bit long, they are formated with different components "$component1-$component2-$component3" and are not supposed to be individually renamed, that fix would help me alot.

alcohol:
[...]

PR #11918, opened 2024-04-04
#11917 fix: Possibility to handle long URLs

p24-max:
Fixes https://github.com/composer/composer/issues/11917

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -67,7 +67,7 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
 
         GitUtil::cleanEnv();
 
-        $cachePath = $this->config->get('cache-vcs-dir').'/'.Preg::replace('{[^a-z0-9.]}i', '-', $url).'/';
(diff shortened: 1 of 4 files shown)


Seldaek:
I'd rather have some additional processing that removes the username/password from the URL before normalizing it, but keeps existing URLs working as they do now. It's much more debuggable than having md5s of URLs in the cache, plus doing this change would blast everyone's cache which is also a bit of a waste.

p24-max:
@Seldaek yes I do understand, the thing is if we just remove the credentials it would fix my problem for sure. But it will not fix other long URLs which are too long for whatever reason, e.g. https://github.com/composer/composer/issues/7313 

From my point of view - whenever possible - it should be avoided to pass external data to file-system operations. You may encounter different issues on different systems (encoding, limitations) and it tends to security flaws. In this case, a git process is spawned containing the sanitised URL as an argument. If sanitisation is not working properly, it may allow to execute arbitrary code. (I'm not saying that this is possible at the moment).

> plus doing this change would blast everyone's cache which is also a bit of a waste.

True.

Seldaek:
Yeah I don't really see super long name as a valid use case. This showed up once. I see the problem with credentials as that is much more likely to be long, but that really should not be part of the repository URL as a cache identifier.

PR #11932, opened 2024-04-17
Improve proxy reporting in Diagnose command

johnstevenson:
This makes the diagnose command compatible with recent proxy handling changes.

The `checkHttpProxy` method is now called with http and https payloads (or http only if `disable-tls` is true) and the proxy server type (HTTP or HTTPS) is shown, together with the safe proxy url.

```
...
Checking HTTP proxy with http: OK http://localhost:6180
Checking HTTP proxy with https: OK http://localhost:6180
...
```

I'm not entirely sure if we should show the proxy url quite so starkly, although it is of course always visible in the debug output with the -vvv option.

If `no_proxy` contains packagist.org, then this is marked as SKIPped (because there is nothing we can do about it anyway).

```
...
Checking HTTP proxy with http: SKIP Because repo.packagist.org is excluded by no_proxy
Checking HTTP proxy with https: SKIP Because repo.packagist.org is excluded by no_proxy
...
```


Seldaek:
Thanks

PR #11933, opened 2024-04-17
Add FAQ about using a proxy

johnstevenson:
Provides a separate FAQ about using Composer behind a proxy, replacing the information from the environment variables section.

Please improve and polish where appropriate.

I've included a section at the end that documents the recent proxy requirement changes and explains the behaviour during the transition period. If you want to keep this, it could be linked to in the warning message.

Alternatively, this document could be linked to if a `TransportException` message starts with 'Unable to use a proxy'.

### DIFF
--- a/doc/faqs/how-to-use-composer-behind-a-proxy.md
+++ b/doc/faqs/how-to-use-composer-behind-a-proxy.md
@@ -1,0 +1,118 @@
+# How to use Composer behind a proxy
+
+Composer, like many other tools, uses environment variables to control the use of a proxy server and
+supports:
+
+- `http_proxy` - the proxy to use for HTTP requests
+- `https_proxy` - the proxy to use for HTTPS requests
+- `CGI_HTTP_PROXY` - the proxy to use for HTTP requests in a non-CLI context
+- `no_proxy` - domains that do not require a proxy
+
+These named variables are a convention, rather than an official standard, and their evolution and
+usage across different operating systems and tools is complex. Composer prefers the use of lowercase
+names, but accepts uppercase names where appropriate.
+
+## Usage
+
+Composer requires specific environment variables for HTTP and HTTPS requests. For example:
+
+```
+http_proxy=http://proxy.com:80
+https_proxy=http://proxy.com:80
+```
+
+Uppercase names can also be used, but HTTP_PROXY will only be accepted in a CLI context.
+
+### Non-CLI usage
+
+Composer does not look for `http_Proxy` or `HTTP_PROXY` in a non-CLI context. If you are running it
+this way (i.e. integration into a CMS or similar use case) you must use `CGI_HTTP_PROXY` for HTTP
+requests:
+
+```
+CGI_HTTP_PROXY=http://proxy.com:80
+https_proxy=http://proxy.com:80
+
+# cgi_http_proxy can also be used
+```
+
+> **Note:** CGI_HTTP_PROXY was introduced by Perl in 2001 to prevent request header manipulation and
+was popularized in 2016 when this vulnerability was widely reported: https://httpoxy.org
+
+## Syntax
+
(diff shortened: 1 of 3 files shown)


doc/faqs/how-to-use-composer-behind-a-proxy.md
```diff
@@ -0,0 +1,118 @@
...
+## Usage
+
+Composer requires specific environment variables for HTTP and HTTPS requests. For example:
+
+```
+http_proxy=http://proxy.com:80
+https_proxy=http://proxy.com:80
+```
+
+Uppercase names can also be used, but HTTP_PROXY will only be accepted in a CLI context.
```
Seldaek, line 24: As far as I can tell even the lowercase `http_proxy` is not accepted in non-CLI SAPIs. So this is kinda misleading I think

doc/faqs/how-to-use-composer-behind-a-proxy.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Non-CLI usage
+
+Composer does not look for `http_Proxy` or `HTTP_PROXY` in a non-CLI context. If you are running it
```
Seldaek, line 28:
```suggestion
Composer does not look for `http_proxy` or `HTTP_PROXY` in a non-CLI context. If you are running it
```

doc/faqs/how-to-use-composer-behind-a-proxy.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+> **Note:** The user name and password components must be percent-encoded individually and then
+combined with the colon separator. The user name cannot contain a colon (even if percent-encoded),
+because the proxy will split the components on the first colon it finds.
+
+## HTTPS proxy servers
+
+Composer supports HTTPS proxy servers, where HTTPS is the scheme used to connect to the proxy, but
+only from PHP 7.3 with cUrl version 7.52.0 and above.
```
Seldaek, line 73:
```suggestion
only from PHP 7.3 with curl version 7.52.0 and above.
```

doc/faqs/how-to-use-composer-behind-a-proxy.md
johnstevenson: Oops!

Seldaek:
Alright, thanks for the thorough documentation! If you can just address the last open comment somehow, aside from that it looks mergeable to me 👍🏻

doc/faqs/how-to-use-composer-behind-a-proxy.md
Seldaek: Nope, and curl's own website says curl, so I'd say let's stick to curl :D

johnstevenson:
Just spotted that the HTTPS proxy section is not entirely correct, because an HTTPS proxy can be used to make http requests with PHP streams.

doc/faqs/how-to-use-composer-behind-a-proxy.md
Seldaek: I think it's ok to leave as is, because http_proxy is anyway fairly useless as http:// urls are nearly dead. Plus before your refactoring it already skipped http_proxy/HTTP_PROXY if not on CLI SAPI. So nothing changed, I'll try and tweak the docs.

Seldaek:
> Just spotted that the HTTPS proxy section is not entirely correct, because an HTTPS proxy can be used to make http requests with PHP streams.

I'd say that does not really matter. It's <1% usage..

cgrisar:
Hi,
Since today it seems I now have a   Call to undefined method Composer\Util\Http\ProxyManager::needsTransitionWarning()  error.

Seems odd that this is just about when you commited the change

Seldaek:
@cgrisar please upgrade to the latest Composer if you still see this error.

PR #11938, opened 2024-04-19
Remove proxy transition fallback in Composer 2.8.0

johnstevenson:
This is the final stage in changing Composer's proxy requirements to scheme-specific environment variables.

It removes the fallback and warning introduced in Composer 2.7.3


Seldaek (approved):

Seldaek:
Thanks

Issue #11906, opened 2024-03-25
[RFC] Automatically bump requirements when updating packages, with type=project or with an option in the configuration


lyrixx:
Hello,

1️⃣ On **projects** (as opposed on libraries ; `"type": "project"`), I think we could run automatically `composer bump` after each `composer update`

2️⃣ If this is too intrusive, could we at least have a new configuration flag in the `composer.json` to configure this behavior 

```json
    "config": { "bump-after-update": true }
```

3️⃣ We could even have something in the middle : 2️⃣ + with `true` as default value. So if someone doesn't want to to bump, they could turn it out

fredden:
Why should this be a built-in feature of Composer rather than a case of running `composer update && composer bump` ?

lyrixx:
More or less the same arguments each times but:

1. Some bots does not bump by default, so it would fix it
2. Some colleagues does not bump some times, so it would fix them
3. To ensure it's always bumped
4. To avoid typos
5. To automate things

PR #11942, opened 2024-04-22
Add option to run bump after update

carlos-granados:
- Adds a new `bump-after-update` config parameter
- Adds a new `bump-after-update` option to the update command

If the `bump-after-update` config parameter is set or the `bump-after-update` option is passed to the `update` command, it will run the `bump` command after finishing the update. This command will only be run if the update finishes successfully

If the config parameter or the option are set to `dev` or `no-dev` then only the corresponding dependencies will be bumped.

The `dry-run` option passed to the `update` command will also be used for the `bump` command

If any packages are listed in the `update` command, this same package list will be used by the `bump` command.

Includes updates to the documentation and new tests

Fixes https://github.com/composer/composer/issues/11906


tests/Composer/Test/Command/UpdateCommandTest.php
```diff
@@ -22,10 +22,14 @@ class UpdateCommandTest extends TestCase
      * @param array<mixed> $composerJson
      * @param array<mixed> $command
      */
-    public function testUpdate(array $composerJson, array $command, string $expected): void
+    public function testUpdate(array $composerJson, array $command, string $expected, bool $createLock = false): void
     {
         $this->initTempComposer($composerJson);
 
+        if ($createLock) {
```
carlos-granados, line 29: The bump command needs that a lock file exist, so when testing the `bump-after-update` version we need to create one

doc/06-config.md
```diff
@@ -476,4 +476,9 @@ throw, but you can set this config option to `["example.org"]` to allow using sv
 URLs on that hostname. This is a better/safer alternative to disabling `secure-http`
 altogether.
 
+## bump-after-update
+
+Defaults to false. If set to true, Composer will run the Bump command after running
```
Seldaek, line 481: I think this could be a string|bool that also accepts "no-dev" and "dev" as values to configure the `*-only` flags.

Seldaek (changes requested):
Thanks for the PR, after a quick review just one comment but I'll need to take a closer look when I'm back at my desk :)

doc/06-config.md
carlos-granados: This suggestion sounds good. I'll wait until you come back and are able to take a deeper look. If you still feel this is the way to go, I'll modify the PR

Seldaek: Yes please :) Definitely feel like this would be an improvement.

src/Composer/Command/BumpCommand.php
```diff
@@ -72,9 +73,29 @@ protected function configure(): void
...
+     * @param string[] $packagesFilter
+     * @throws \Seld\JsonLint\ParsingException
+     */
+    public function doBump(
+        IOInterface $io,
+        bool $devOnly,
+        bool $noDevOnly,
+        bool $dryRun,
+        array $packagesFilter,
+        bool $calledFromUpdate = false
```
Seldaek, line 95: Seems unused and unnecessary?

src/Composer/Command/UpdateCommand.php
```diff
@@ -248,7 +251,27 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $install->disablePlugins();
         }
 
-        return $install->run();
+        $result = $install->run();
+        $bumpAfterUpdate = $input->getOption('bump-after-update') || $composer->getConfig()->get('bump-after-update');
+
+        if ($bumpAfterUpdate) {
+            if ($result === 0) {
+                $io->writeError('<info>Running Bump after Update.</info>');
```
Seldaek, line 259:
```suggestion
                $io->writeError('<info>Bumping dependencies</info>');
```

src/Composer/Command/UpdateCommand.php
```diff
@@ -248,7 +251,27 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                $result = $bumpCommand->doBump(
+                    $io,
+                    $input->getOption('bump-dev-only'),
+                    $input->getOption('bump-no-dev-only'),
+                    $input->getOption('dry-run'),
+                    $input->getArgument('packages'),
+                    true
+                );
+            } else {
+                $io->writeError('<warning>Not running Bump after Update because the update command did not finish successfully.</warning>');
```
Seldaek, line 271: I don't think this is needed.

```suggestion
```

cafferata:
Hi @Seldaek, is there a reason why this GitHub pull request was not receive a (re)review the past four months? Has this perhaps been lost sight of? No rush, I saw other GitHub pull requests being reviewed/merged. So I was wondering how this works within your available time. Of course an answer of no time/priority is also completely valid. ☺️

Seldaek:
It's just targetting 2.8 and I was focused on shipping patch releases for 2.7 so far.. kinda buffering 2.8 PRs until I'm ready to work on those because once I branch it out then I need to merge 2.7 into 2.8 which is more work for me.

lyrixx:
Oh, thanks for this PR. I had spare time, and decided to try to fix my issue. And I totally miss your PR :/ 
For the record, I had another approach, were I factorized the code in a new service: https://github.com/lyrixx/composer/commit/df8a25bf8f2b215545a79f1f07352bd35fe91246

doc/03-cli.md
```diff
@@ -234,6 +234,7 @@ php composer.phar update vendor/package:2.0.1 vendor/package2:3.0.*
   changes to transitive dependencies. Can also be set via the COMPOSER_MINIMAL_CHANGES=1 env var.
 * **--interactive:** Interactive interface with autocompletion to select the packages to update.
 * **--root-reqs:** Restricts the update to your first degree dependencies.
+* **--bump-after-update:** Runs Bump after performing the update. Set to `dev` or `no-dev` to only bump those dependencies.
```
Seldaek, line 237:
```suggestion
* **--bump-after-update:** Runs `bump` after performing the update. Set to `dev` or `no-dev` to only bump those dependencies.
```

doc/06-config.md
```diff
@@ -476,4 +476,9 @@ throw, but you can set this config option to `["example.org"]` to allow using sv
 URLs on that hostname. This is a better/safer alternative to disabling `secure-http`
 altogether.
 
+## bump-after-update
+
+Defaults to `false` and can be any of `true`, `false`, `"dev"` or `"no-dev"`. If set to true, Composer will run the Bump command after running
+the Update command. If set to `"dev"` or `"no-dev"` then only the corresponding dependencies will be bumped.
```
Seldaek, line 482:
```suggestion
Defaults to `false` and can be any of `true`, `false`, `"dev"` or `"no-dev"`. If set to true, Composer will run the `bump` command after running
the `update` command. If set to `"dev"` or `"no-dev"` then only the corresponding dependencies will be bumped.
```

res/composer-schema.json
```diff
@@ -656,6 +656,10 @@
                 "platform-check": {
                     "type": ["boolean", "string"],
                     "description": "Defaults to \"php-only\" which checks only the PHP version. Setting to true will also check the presence of required PHP extensions. If set to false, Composer will not create and require a platform_check.php file as part of the autoloader bootstrap."
+                },
+                "bump-after-update": {
+                    "type": ["string", "boolean"],
+                    "description": "Defaults to false and can be any of true, false, \"dev\"` or \"no-dev\"`. If set to true, Composer will run the Bump command after running the Update command. If set to \"dev\" or \"no-dev\" then only the corresponding dependencies will be bumped."
```
Seldaek, line 662:
```suggestion
                    "description": "Defaults to false and can be any of true, false, \"dev\"` or \"no-dev\"`. If set to true, Composer will run the bump command after running the update command. If set to \"dev\" or \"no-dev\" then only the corresponding dependencies will be bumped."
```

src/Composer/Command/BumpCommand.php
```diff
@@ -112,27 +132,26 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
 
-        if ($composer->getPackage()->getType() !== 'project' && !$input->getOption('dev-only')) {
+        if ($composer->getPackage()->getType() !== 'project' && !$devOnly) {
             $io->writeError('<warning>Warning: Bumping dependency constraints is not recommended for libraries as it will narrow down your dependencies and may cause problems for your users.</warning>');
 
             $contents = $composerJson->read();
             if (!isset($contents['type'])) {
                 $io->writeError('<warning>If your package is not a library, you can explicitly specify the "type" by using "composer config type project".</warning>');
-                $io->writeError('<warning>Alternatively you can use --dev-only to only bump dependencies within "require-dev".</warning>');
+                $io->writeError('<warning>Alternatively you can use the dev-only option to only bump dependencies within "require-dev".</warning>');
```
Seldaek, line 141:
```suggestion
                $io->writeError('<warning>Alternatively you can use --dev-only to only bump dependencies within "require-dev".</warning>');
```

src/Composer/Command/UpdateCommand.php
```diff
@@ -78,6 +78,7 @@ protected function configure()
                 new InputOption('minimal-changes', 'm', InputOption::VALUE_NONE, 'During a partial update with -w/-W, only perform absolutely necessary changes to transitive dependencies (can also be set via the COMPOSER_MINIMAL_CHANGES=1 env var).'),
                 new InputOption('interactive', 'i', InputOption::VALUE_NONE, 'Interactive interface with autocompletion to select the packages to update.'),
                 new InputOption('root-reqs', null, InputOption::VALUE_NONE, 'Restricts the update to your first degree dependencies.'),
+                new InputOption('bump-after-update', null, InputOption::VALUE_NONE, 'Runs Bump after performing the update.'),
```
Seldaek, line 81:
```suggestion
                new InputOption('bump-after-update', null, InputOption::VALUE_OPTIONAL, 'Runs bump after performing the update.', false, ['dev', 'no-dev', 'all']),
```

src/Composer/Command/UpdateCommand.php
```diff
@@ -248,7 +249,27 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                $result = $bumpCommand->doBump(
+                    $io,
+                    $input->getOption('bump-after-update') === 'dev'
+                        || $composer->getConfig()->get('bump-after-update') === 'dev',
+                    $input->getOption('bump-after-update') === 'no-dev'
+                        || $composer->getConfig()->get('bump-after-update') === 'no-dev',
+                    $input->getOption('dry-run'),
+                    $input->getArgument('packages')
+                );
+            }
```
Seldaek, line 270: Doing `$input->getOption('bump-after-update') === 'dev' || $composer->getConfig()->get('bump-after-update') === 'dev',` here isn't very safe as you might end up with only dev AND only non-dev if you have the config set to no-dev and the option set to --bump-after-update=dev then both will be true.

```suggestion
            $bumpAfterUpdate = $input->getOption('bump-after-update');
            if (false === $bumpAfterUpdate) {
                $bumpAfterUpdate = $composer->getConfig()->get('bump-after-update');
            }

            if (false !== $bumpAfterUpdate) {
                $io->writeError('<info>Bumping dependencies</info>');
                $bumpCommand = new BumpCommand();
                $bumpCommand->setComposer($composer);
                $result = $bumpCommand->doBump(
                    $io,
                    $bumpAfterUpdate === 'dev',
                    $bumpAfterUpdate === 'no-dev',
                    $input->getOption('dry-run'),
                    $input->getArgument('packages')
                );
            }
```

Seldaek (approved):
Mostly found some minor wording style issues, and one minor edge case in the code, overall pretty solid PR thanks! And sorry for the delay :)

tests/Composer/Test/Command/UpdateCommandTest.php
```diff
@@ -126,6 +130,62 @@ public static function provideUpdates(): \Generator
...
+  - Locking dep/pkg (1.0.2)
+  - Locking root/req (1.0.0)
+Installing dependencies from lock file (including require-dev)
+Package operations: 2 installs, 0 updates, 0 removals
+  - Installing dep/pkg (1.0.2)
+  - Installing root/req (1.0.0)
+Bumping dependencies
+<warning>Warning: Bumping dependency constraints is not recommended for libraries as it will narrow down your dependencies and may cause problems for your users.</warning>
+<warning>If your package is not a library, you can explicitly specify the "type" by using "composer config type project".</warning>
+<warning>Alternatively you can use the dev-only option to only bump dependencies within "require-dev".</warning>
```
Seldaek, line 150:
```suggestion
<warning>Alternatively you can use --dev-only to only bump dependencies within "require-dev".</warning>
```

res/composer-schema.json
```diff
@@ -663,6 +663,10 @@
                     "type": ["boolean", "string"],
                     "description": "Defaults to \"php-only\" which checks only the PHP version. Setting to true will also check the presence of required PHP extensions. If set to false, Composer will not create and require a platform_check.php file as part of the autoloader bootstrap."
                 },
+                "bump-after-update": {
+                    "type": ["string", "boolean"],
```
stof, line 667: this should use `enum` to restrict the supported strings

tests/Composer/Test/Command/UpdateCommandTest.php
```diff
@@ -126,6 +130,62 @@ public static function provideUpdates(): \Generator
...
+  - Locking dep/pkg (1.0.2)
+  - Locking root/req (1.0.0)
+Installing dependencies from lock file (including require-dev)
+Package operations: 2 installs, 0 updates, 0 removals
+  - Installing dep/pkg (1.0.2)
+  - Installing root/req (1.0.0)
+Bumping dependencies
+<warning>Warning: Bumping dependency constraints is not recommended for libraries as it will narrow down your dependencies and may cause problems for your users.</warning>
+<warning>If your package is not a library, you can explicitly specify the "type" by using "composer config type project".</warning>
+<warning>Alternatively you can use --dev-only to only bump dependencies within "require-dev".</warning>
```
stof, line 150: this `--dev-only` option is confusing when running this as part of `composer update` instead of `composer bump`

src/Composer/Command/UpdateCommand.php
```diff
@@ -85,6 +85,7 @@ protected function configure()
                 new InputOption('minimal-changes', 'm', InputOption::VALUE_NONE, 'During a partial update with -w/-W, only perform absolutely necessary changes to transitive dependencies (can also be set via the COMPOSER_MINIMAL_CHANGES=1 env var).'),
                 new InputOption('interactive', 'i', InputOption::VALUE_NONE, 'Interactive interface with autocompletion to select the packages to update.'),
                 new InputOption('root-reqs', null, InputOption::VALUE_NONE, 'Restricts the update to your first degree dependencies.'),
+                new InputOption('bump-after-update', null, InputOption::VALUE_OPTIONAL, 'Runs bump after performing the update.', false, ['dev', 'no-dev', 'all']),
```
stof, line 88: using `InputOption::VALUE_OPTIONAL` is confusing here, because it makes `composer update --bump-after-update` valid during argument validation, but this would still make it use the default value (i.e. `false`). This should probably be using `VALUE_REQUIRED` instead.

Seldaek: This is actually not an issue as with VALUE_OPTIONAL when you just do --flag the value is `null`, and not the default value (`false` here).

PR #11946, opened 2024-04-22
Don't show root warning for Podman containers

maximilian-walter:
Podman creates the file `/run/.containerenv` (or `/var/run/.containerenv`) instead of /.dockerenv, so I added these two paths to the existing checks.

> Additionally, a container environment file is created in each container to indicate to programs they are running in a container. This file is located at `/run/.containerenv` (or `/var/run/.containerenv` for FreeBSD containers).

https://docs.podman.io/en/latest/markdown/podman-run.1.html

Follow up of #8395


Seldaek:
Nice thanks

Issue #11940, opened 2024-04-20
Second run of the latest composer install results in error: Call to undefined method Composer\Util\Http\ProxyManager::needsTransitionWarning()


lubikx:
My `composer.json`:

```json
{
    "name": "laravel/laravel",
    "type": "project",
    "description": "The Laravel Framework.",
    "keywords": [
        "framework",
        "laravel"
    ],
    "license": "MIT",
    "require": {
        "arcanedev/log-viewer": "^8.0",
        "doctrine/dbal": "^2.12.1",
        "fideloper/proxy": "^4.4.1",
        "guzzlehttp/guzzle": "^7.0.1",
        "inspector-apm/inspector-laravel": "^4.7",
        "laravel/framework": "^8.0",
        "laravel/tinker": "^3.0",
        "laravel/ui": "^3.0",
        "laravelcollective/html": "^6.2.0",
        "maatwebsite/excel": "^3.1",
        "sentry/sentry-laravel": "*",
        "yoeunes/notify": "^1.0.6"
    },
    "require-dev": {
        "barryvdh/laravel-debugbar": "^3.5",
        "barryvdh/laravel-ide-helper": "^2.12",
        "beyondcode/laravel-dump-server": "^1.6",
        "filp/whoops": "^2.9",
        "fzaninotto/faker": "*",
        "mockery/mockery": "^1.4",
        "nunomaduro/collision": "^5.0",
        "nunomaduro/larastan": "^0.6.2",
        "phpstan/phpstan": "^0.12",
        "phpunit/phpunit": "^9.0"
    },
    "config": {
        "optimize-autoloader": true,
        "preferred-install": "dist",
        "sort-packages": true
    },
    "extra": {
        "laravel": {
            "dont-discover": []
        }
    },
    "autoload": {
        "psr-4": {
            "App\\": "app/"
        },
        "classmap": [
            "database/seeds",
            "database/factories"
        ]
    },
    "autoload-dev": {
        "psr-4": {
            "Tests\\": "tests/"
        }
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "scripts": {
        "post-autoload-dump": [
            "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
            "@php artisan package:discover --ansi"
        ],
        "post-root-package-install": [
            "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
        ],
        "post-create-project-cmd": [
            "@php artisan key:generate --ansi"
        ]
[...]

PR #11948, opened 2024-04-23
Check autoload functions before callback

barryvdh:
Check the autoload functions before and after. Unset the ones added during callback. Refs https://github.com/composer/composer/issues/11940


barryvdh:
The return signature is changed in PHP8.0; https://www.php.net/manual/en/function.spl-autoload-functions.php

> 8.0.0	Return value was updated to always be an [array](https://www.php.net/manual/en/language.types.array.php); previously this function returned [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) if the autoload queue wasn't activated.

In our case we are pretty sure that it is set, but not sure what the best way is to make phpstan happy here for 7.2 and 8.1

PR #11950, opened 2024-04-24
Updated array shape of php-ext options

asgrim:
Updating the `php-ext` section shape in `composer.json` to support the newer properties we added ThePHPF/pie-design#12 and ThePHPF/pie-design#10


Seldaek:
Thanks

PR #11951, opened 2024-04-25
Add `uninstall` as alias to `composer remove`.

BusterNeece:
I've looked through the repository and realized I've never actually seen this proposed either in issue or pull request form.

Some package managers use `uninstall` as the convention for removing/uninstalling dependencies, and as a result I've found myself in the habit of typing `composer uninstall (packagename)` instead of `composer remove (packagename)`.

Composer doesn't currently handle this all that gracefully...it suggests "Did you mean install/reinstall?" without suggesting the actual proper `remove` command.

I think, perhaps to save folks like me the trouble, Composer can follow in the footsteps of package managers like NPM and make `remove` and `uninstall` refer to the same command, so there's no retyping involved if you slip up and forget which is which.

### DIFF
--- a/src/Composer/Command/RemoveCommand.php
+++ b/src/Composer/Command/RemoveCommand.php
@@ -43,7 +43,7 @@ class RemoveCommand extends BaseCommand
     {
         $this
             ->setName('remove')
-            ->setAliases(['rm'])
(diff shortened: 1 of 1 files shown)


Seldaek:
Sure why not, thanks

PR #11956, opened 2024-04-29
Composer update: filter packages with security advisories from pool

glaubinix:
Please note, this is still a very rough draft missing tests, cleanup and taking a few shortcuts. However, feedback is already appreciated.

### Idea

Composer already has an audit functionality that reports any package used by a project with security advisories as part of an install/update/audit command. This PR takes this one step further and filters any packages with security advisories from the pool of available packages during a `composer update`command before the pool get optimized.

The functionality uses cached metadata files wherever possible. However, it is possible that it will trigger a call to the security advisories API endpoint on packagist.org if multiple Composer repositories are defined and not all of them are looked up on packagist.org.

This could potentially replace the need for projects to install a special package defining conflicts with packages that have security advisories.

### Sample composer.json
```
{
    "name": "acme/project",
    "version": "1.0",
    "repositories": [
    ],
    "require": {
        "doctrine/cache": "<=1.3.0,>=1.0.0"
    }
}
```

### How this currently looks in Composer with verbose output
```
Running 2.7.999-dev+source (@release_date@) with PHP
Reading ./composer.json (/tmp/composer.json)
Loading config file ~/.composer/config.json
Loading config file ~/.composer/auth.json
Loading config file ./composer.json (/tmp/composer.json)
Checked CA file /opt/homebrew/etc/ca-certificates/cert.pem: valid
Reading ~/.composer/composer.json
Loading config file ~/.composer/config.json
Loading config file ~/.composer/auth.json
Loading config file ~/.composer/composer.json (~/.composer/composer.json)
Loading config file ~/.composer/auth.json
Reading ~/.composer/auth.json
Reading ./composer.lock (/tmp/composer.lock)
Reading /tmp/vendor/composer/installed.json
Reading ~/.composer/vendor/composer/installed.json
Reading ~/Library/Caches/composer/repo/https---repo.packagist.org/packages.json from cache
Downloading https://repo.packagist.org/packages.json if modified
[200] https://repo.packagist.org/packages.json
Writing ~/Library/Caches/composer/repo/https---repo.packagist.org/packages.json into cache
Downloading https://repo.packagist.org/p2/doctrine/cache.json
[200] https://repo.packagist.org/p2/doctrine/cache.json
Writing ~/composer-cache/repo/https---repo.packagist.org/provider-doctrine~cache.json into cache
Built pool.
Running security advisory pool filter.
Reading ~/composer-cache/repo/https---repo.packagist.org/provider-doctrine~cache.json from cache
Security advisory pool filter completed in 0.001 seconds
Found 105 package versions referenced in your dependency graph. 1 (1%) were filtered away.
Running pool optimizer.
Updating dependencies
Generating rules
Looking at all rules.

Reading ~/composer-cache/repo/https---repo.packagist.org/provider-doctrine~cache.json from cache

Problem 1
  Problem 1
    - Root composer.json requires doctrine/cache <=1.3.0,>=1.0.0, found doctrine/cache[v1.0, v1.1, v1.2.0, v1.3.0] but these were not loaded, because they have security advisories.

```

### Questions
* [ ] Should this step be optional
* [ ] Should this also consider the audit abandoned config and filter those packages too

### DIFF
--- a/src/Composer/Advisory/AuditConfig.php
+++ b/src/Composer/Advisory/AuditConfig.php
@@ -1,0 +1,48 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\Advisory;
+
+use Composer\Config;
+
+class AuditConfig
+{
+    /**
+     * @var array<string>|array<string,string> List of advisory IDs, remote IDs or CVE IDs that reported but not listed as vulnerabilities.
+     */
+    public $ignoreList;
+
+    /**
+     * @var Auditor::ABANDONED_*
+     */
+    public $abandoned;
+
+    /**
+     * @param array<string>|array<string,string> $ignoreList
+     * @param Auditor::ABANDONED_* $abandoned
+    */
+    public function __construct(array $ignoreList, string $abandoned)
+    {
+        $this->ignoreList = $ignoreList;
+        $this->abandoned = $abandoned;
+    }
+
+    public static function fromConfig(Config $config): self
+    {
+        $auditConfig = $config->get('audit');
+
+        return new self(
+            $auditConfig['ignore'] ?? [],
+                $auditConfig['abandoned'] ?? Auditor::ABANDONED_FAIL
+        );
+    }
+}
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -141,7 +141,7 @@ class Auditor
      * @param array<string>|array<string,string> $ignoreList List of advisory IDs, remote IDs or CVE IDs that reported but not listed as vulnerabilities.
      * @phpstan-return array{advisories: array<string, array<PartialSecurityAdvisory|SecurityAdvisory>>, ignoredAdvisories: array<string, array<PartialSecurityAdvisory|SecurityAdvisory>>}
      */
-    private function processAdvisories(array $allAdvisories, array $ignoreList): array
+    public function processAdvisories(array $allAdvisories, array $ignoreList): array
     {
         if ($ignoreList === []) {
             return ['advisories' => $allAdvisories, 'ignoredAdvisories' => []];
--- a/src/Composer/Command/AuditCommand.php
+++ b/src/Composer/Command/AuditCommand.php
@@ -12,6 +12,7 @@
 
 namespace Composer\Command;
 
+use Composer\Advisory\AuditConfig;
 use Composer\Composer;
 use Composer\Repository\RepositorySet;
 use Composer\Repository\RepositoryUtils;
@@ -63,9 +64,9 @@ EOT
             $repoSet->addRepository($repo);
         }
 
-        $auditConfig = $composer->getConfig()->get('audit');
+        $auditConfig = AuditConfig::fromConfig($composer->getConfig());
 
-        return min(255, $auditor->audit($this->getIO(), $repoSet, $packages, $this->getAuditFormat($input, 'format'), false, $auditConfig['ignore'] ?? [], $auditConfig['abandoned'] ?? Auditor::ABANDONED_FAIL));
+        return min(255, $auditor->audit($this->getIO(), $repoSet, $packages, $this->getAuditFormat($input, 'format'), false, $auditConfig->ignoreList, $auditConfig->abandoned));
     }
 
     /**
--- a/src/Composer/DependencyResolver/Pool.php
+++ b/src/Composer/DependencyResolver/Pool.php
@@ -40,20 +40,24 @@ class Pool implements \Countable
     protected $removedVersions = [];
     /** @var array<string, array<string, string>> Map of package object hash => removed normalized versions => removed pretty version */
     protected $removedVersionsByPackage = [];
+    /** @var array<string, array<string, string>> Map of package name => normalized version => pretty version */
+    private $securityRemovedVersions = [];
 
     /**
      * @param BasePackage[] $packages
      * @param BasePackage[] $unacceptableFixedOrLockedPackages
      * @param array<string, array<string, string>> $removedVersions
      * @param array<string, array<string, string>> $removedVersionsByPackage
+     * @param array<string, array<string, string>> $securityRemoveVersions
      */
-    public function __construct(array $packages = [], array $unacceptableFixedOrLockedPackages = [], array $removedVersions = [], array $removedVersionsByPackage = [])
+    public function __construct(array $packages = [], array $unacceptableFixedOrLockedPackages = [], array $removedVersions = [], array $removedVersionsByPackage = [], array $securityRemoveVersions = [])
     {
         $this->versionParser = new VersionParser;
         $this->setPackages($packages);
         $this->unacceptableFixedOrLockedPackages = $unacceptableFixedOrLockedPackages;
         $this->removedVersions = $removedVersions;
         $this->removedVersionsByPackage = $removedVersionsByPackage;
+        $this->securityRemovedVersions = $securityRemoveVersions;
     }
 
     /**
@@ -85,6 +89,25 @@ class Pool implements \Countable
         }
 
         return $this->removedVersionsByPackage[$objectHash];
+    }
+
+    public function isSecurityRemovedPackageVersion(string $packageName, ?ConstraintInterface $constraint): bool
+    {
+        foreach ($this->securityRemovedVersions[$packageName] ?? [] as $version => $prettyVersion) {
+            if ($constraint !== null && $constraint->matches(new Constraint('==', $version))) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @return array<string, array<string, string>>
+     */
+    public function getAllSecurityRemovedPackageVersions(): array
+    {
+        return $this->securityRemovedVersions;
     }
 
     /**
--- a/src/Composer/DependencyResolver/SecurityAdvisoryPoolFilter.php
+++ b/src/Composer/DependencyResolver/SecurityAdvisoryPoolFilter.php
@@ -1,0 +1,100 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\DependencyResolver;
+
+use Composer\Advisory\AuditConfig;
+use Composer\Advisory\Auditor;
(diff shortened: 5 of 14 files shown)


Seldaek:
Cool thanks, not sure either how to answer your questions yet. But I'd say probably needs to be optional, and probably abandoned filtering should also be an option at that level? Maybe like `"audit": {"block-insecure": bool, "block-abandoned": bool}`? 

Does it work on install too btw if an insecure package already made its way into the lock file?

As install runs through the pool builder etc as well, I am thinking it might just work, but it'd be good to test.. And I'm not sure if that should be additionally opt-in behavior as it might break pipelines whenever an advisory is released.

glaubinix:
Having this as separate config options sounds like a good idea! Applied this already. Blocking on install doesn't work yet as the repository list during the composer install command currently only contains the platform/lock/local repositories and none of the once configured in the repositories section in the composer.json file. So the repository set will always return 0 advisories. Still need to figure out how to pass those through.

staabm:
I wonder what the benefit of this PR is over already working things like https://github.com/Roave/SecurityAdvisories

stof:
An install from lock should not filter out vulnerable packages IMO as it would totally break the usage of that lock file when new advisories are released.

And this should definitely stay opt-in (blocking all vulnerable package could harm projects using a package which has a vulnerability for a specific case that does not impact the project by forcing them to update that package before being able to do anything else)

glaubinix:
The main benefits are:
* support for additional sources like GitHub advisories, Drupal out of the box without having to require additional packages as long as the advisories are exposed via a Composer repository
* option to ignore certain advisories that you either don't care about or don't affect your code base

Yes, failing composer installs should definitely be a separate config option and all options should be opt-in by default.

glaubinix:
Adjusted the PR to incorporate the feedback. New audit config options have been added. Removed the part that filtered out versions from the install command. Kept the feature on by default for update commands but added more info on why the update fails and instructions on how to ignore the advisories.

src/Composer/Advisory/Auditor.php
```diff
@@ -186,7 +186,7 @@ private function filterAbandonedPackages(array $packages, array $ignoreAbandoned
      * @param array<string> $ignoredSeverities List of ignored severity levels
      * @phpstan-return array{advisories: array<string, array<PartialSecurityAdvisory|SecurityAdvisory>>, ignoredAdvisories: array<string, array<PartialSecurityAdvisory|SecurityAdvisory>>}
      */
-    private function processAdvisories(array $allAdvisories, array $ignoreList, array $ignoredSeverities): array
+    public function processAdvisories(array $allAdvisories, array $ignoreList, array $ignoredSeverities): array
```
glaubinix, line 189: Not too happy about these two but not sure how else to do this without duplicating the logic in the two private methods

src/Composer/Advisory/AuditConfig.php
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * @readonly
+ * @internal
+ */
+class AuditConfig
+{
+    /**
+     * @var array<string>|array<string,string> List of advisory IDs, remote IDs or CVE IDs that reported but not listed as vulnerabilities.
+     */
+    public $ignoreList;
```
staabm: Can we get native types on the properties?

doc/06-config.md
```diff
@@ -171,6 +171,65 @@ Since Composer 2.8, the option can be overridden via the
 [`--abandoned`](03-cli.md#audit) command line option, which overrides both the
 config value and the environment variable.
 
+### ignore-severity
+
+A list of severity levels that let the audit command pass even if there are security advisories
```
Seldaek:
```suggestion
Defaults to `[]`. A list of severity levels that let the audit command pass even if there are security advisories
```

doc/06-config.md
```diff
@@ -171,6 +171,65 @@ Since Composer 2.8, the option can be overridden via the
...
+        "audit": {
+            "ignore-severity": ["low"]
+        }
+    }
+}
+```
+
+### ignore-unreachable
+
+Should unreachable repositories be ignored during a `composer audit`. This can be helpful if you are running the command
```
Seldaek:
```suggestion
Defaults to `false`. Should unreachable repositories be ignored during a `composer audit`. This can be helpful if you are running the command
```

src/Composer/Advisory/AuditConfig.php
```diff
@@ -0,0 +1,89 @@
...
+        $auditConfig = $config->get('audit');
+
+        return new self(
+            $auditConfig['ignore'] ?? [],
+                $auditConfig['abandoned'] ?? Auditor::ABANDONED_FAIL,
+                (bool) ($auditConfig['block-insecure'] ?? true),
+                (bool) ($auditConfig['block-abandoned'] ?? false),
+                $auditConfig['ignore-severity'] ?? [],
+                (bool) ($auditConfig['ignore-unreachable'] ?? false),
+                $auditConfig['ignore-abandoned'] ?? []
```
Seldaek:
```suggestion
            $auditConfig['abandoned'] ?? Auditor::ABANDONED_FAIL,
            (bool) ($auditConfig['block-insecure'] ?? true),
            (bool) ($auditConfig['block-abandoned'] ?? false),
            $auditConfig['ignore-severity'] ?? [],
            (bool) ($auditConfig['ignore-unreachable'] ?? false),
            $auditConfig['ignore-abandoned'] ?? []
```

src/Composer/Advisory/Auditor.php
Seldaek: Seems acceptable to me considering the whole class is `@internal`

src/Composer/Command/AuditCommand.php
```diff
@@ -68,29 +69,29 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         if ($abandoned !== null && !in_array($abandoned, Auditor::ABANDONEDS, true)) {
             throw new \InvalidArgumentException('--audit must be one of '.implode(', ', Auditor::ABANDONEDS).'.');
         }
 
-        $abandoned = $abandoned ?? $auditConfig['abandoned'] ?? Auditor::ABANDONED_FAIL;
+        $abandoned = $abandoned ?? $auditConfig->abandoned;
 
-        $ignoreSeverities = $input->getOption('ignore-severity') ?? [];
-        $ignoreUnreachable = $input->getOption('ignore-unreachable');
+        $ignoreSeverities = (bool) $input->getOption('ignore-severity') ? $input->getOption('ignore-severity') : $auditConfig->ignoreSeverity;
```
Seldaek: Should this not just be cast to array and always be merged with the stuff from the config? Not sure if passing an additional severity to ignore here should override what's in the config?

src/Composer/DependencyResolver/Pool.php
```diff
@@ -40,20 +42,28 @@ class Pool implements \Countable
     protected $removedVersions = [];
     /** @var array<string, array<string, string>> Map of package object hash => removed normalized versions => removed pretty version */
     protected $removedVersionsByPackage = [];
+    /** @var array<string, array<string, array<SecurityAdvisory|PartialSecurityAdvisory>>> Map of package name => normalized version => pretty version */
```
Seldaek:
```suggestion
    /** @var array<string, array<string, array<SecurityAdvisory|PartialSecurityAdvisory>>> Map of package name => normalized version => security advisories */
```

src/Composer/DependencyResolver/SecurityAdvisoryPoolFilter.php
```diff
@@ -0,0 +1,117 @@
...
+use Composer\Advisory\SecurityAdvisory;
+use Composer\Package\CompletePackage;
+use Composer\Package\PackageInterface;
+use Composer\Package\RootPackageInterface;
+use Composer\Repository\PlatformRepository;
+use Composer\Repository\RepositoryInterface;
+use Composer\Repository\RepositorySet;
+use Composer\Semver\Constraint\Constraint;
+
+class SecurityAdvisoryPoolFilter
```
Seldaek, line 30:
```suggestion
/**
 * @internal
 */
class SecurityAdvisoryPoolFilter
```

src/Composer/DependencyResolver/Pool.php
```diff
@@ -87,6 +97,60 @@ public function getRemovedVersionsByPackage(string $objectHash): array
...
+        return $this->securityRemovedVersions;
+    }
+
+    /**
+     * @return array<string, array<string, string>>
+     */
+    public function getAllAbandonedRemovedPackageVersions(): array
+    {
+        return $this->abandonedRemovedVersions;
+    }
```
Seldaek, line 168:
```suggestion
    /**
     * @return array<string, array<string, array<SecurityAdvisory|PartialSecurityAdvisory>>>
     */
    public function getAllSecurityRemovedPackageVersions(): array
    {
        return $this->securityRemovedVersions;
    }

    /**
     * @return array<string, array<string, string>>
     */
    public function getAllAbandonedRemovedPackageVersions(): array
    {
        return $this->abandonedRemovedVersions;
    }
```

I wonder if we should also add the same for `$removedVersions` / `$removedVersionsByPackage`, to make the various passes interchangeable (just curious if running security filter after pool optimizer would perhaps be faster?)

src/Composer/DependencyResolver/SecurityAdvisoryPoolFilter.php
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        return new Pool($packages, $pool->getUnacceptableFixedOrLockedPackages(), [], [], $securityRemovedVersions, $abandonedRemovedVersions);
+    }
+
+    /**
+     * @param array<string, array<PartialSecurityAdvisory|SecurityAdvisory>> $advisoryMap
+     * @return list<PartialSecurityAdvisory|SecurityAdvisory>
+     */
+    private function getPackageMatchAdvisories(PackageInterface $package, array $advisoryMap): array
```
Seldaek:
```suggestion
    private function getMatchingAdvisories(PackageInterface $package, array $advisoryMap): array
```

tests/Composer/Test/Fixtures/installer/update-security-advisory-matching-indirect-dependency.test
```diff
@@ -0,0 +1,74 @@
+--TEST--
+Security advisory matching indirect dependency preventing a successful update.
```
Seldaek, line 2: It'd be great to have a test showing that install still works even if the lock file has packages with advisories.

src/Composer/DependencyResolver/Pool.php
glaubinix: Making them interchangeable sounds like a great idea. Added that. I could find a noticeable difference in the order. Running `composer update -vvv` on the Composer repo a few times resulted in:

```
Pool optimizer completed in 0.029 seconds
Security advisory pool filter completed in 0.010 seconds
```
and
```
Security advisory pool filter completed in 0.012 seconds
Pool optimizer completed in 0.028 seconds
```

I do think that projects using additional repositories where the optimizer can remove dependencies from the result set when the metadata endpoint doesn't return advisory information removes the need to query packagist.org for advisories for those packages. This potentially results in overall better performance. So swapped them around.

src/Composer/Command/AuditCommand.php
glaubinix: Merge sounds good. Changed.

Seldaek:
Thanks!

PR #11968, opened 2024-05-08
Re-use precalculated information

Toflar:
No need for thousands of `strpos()` calls as we already have this information on the package itself. Shaves off about 0.1 seconds on the `PoolOptimizer` step in my case.

### DIFF
--- a/src/Composer/DependencyResolver/DefaultPolicy.php
+++ b/src/Composer/DependencyResolver/DefaultPolicy.php
@@ -57,7 +57,7 @@ class DefaultPolicy implements PolicyInterface
         }
 
         // dev versions need to be compared as branches via matchSpecific's special treatment, the rest can be optimized with compiling matcher
-        if ($a->isDev() || $b->isDev()) {
+        if (($a->isDev() && str_starts_with($a->getVersion(), 'dev-')) || ($b->isDev() && str_starts_with($b->getVersion(), 'dev-'))) {
             $constraint = new Constraint($operator, $b->getVersion());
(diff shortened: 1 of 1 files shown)


src/Composer/DependencyResolver/DefaultPolicy.php
```diff
@@ -57,7 +57,7 @@ public function versionCompare(PackageInterface $a, PackageInterface $b, string
         }
 
         // dev versions need to be compared as branches via matchSpecific's special treatment, the rest can be optimized with compiling matcher
-        if (strpos($a->getVersion(), 'dev-') === 0 || strpos($b->getVersion(), 'dev-') === 0) {
+        if ($a->isDev() || $b->isDev()) {
```
Seldaek, line 60: isDev uses VersionParser::parseStability which also matches -dev suffix. So I don't think this is strictly equivalent, and may cause issues when comparing 1.999.999-dev to 2.999.999.x-dev for example. It probably won't match incorrectly, but it removes the CompilingMatcher optimization, so you could argue it's making it a bit slower too 🤷🏻 

Can you benchmark this?

```suggestion
        if (($a->isDev() && strpos($a->getVersion(), 'dev-') === 0) || ($b->isDev() && strpos($b->getVersion(), 'dev-') === 0)) {
```

Toflar:
About the same results. Guess in my example project it just doesn't make much of a difference. This is not gonna make any big difference anyway but if we can avoid the additional function calls, I think we should 😊

Seldaek:
Ok sure makes sense, thanks

PR #11969, opened 2024-05-09
Fix composer error when git config safe.bareRepository is set to explicit

krzysztof-ciszewski:
Fixes composer#11855

### DIFF
--- a/tests/Composer/Test/ApplicationTest.php
+++ b/tests/Composer/Test/ApplicationTest.php
@@ -72,6 +72,6 @@ class ApplicationTest extends TestCase
         $output = new BufferedOutput();
         $application->doRun(new ArrayInput(['command' => 'self-update']), $output);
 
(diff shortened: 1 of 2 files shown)


tests/Composer/Test/ApplicationTest.php
```diff
@@ -72,6 +72,6 @@ public function testDevWarningSuppressedForSelfUpdate(): void
         $output = new BufferedOutput();
         $application->doRun(new ArrayInput(['command' => 'self-update']), $output);
 
-        $this->assertSame('', $output->fetch());
+        $this->assertSame("<warning>Composer could not detect the root package (composer/composer) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version</warning>\n", $output->fetch());
```
krzysztof-ciszewski, line 75: Is this fine? Technically it's true right?

krzysztof-ciszewski:
@Seldaek could you please mark this PR as ready to review? I think php 8.4 tests are allowed to fail right?

Seldaek:
Thanks. I am not a huge fan of doing this git specific handling in the ProcessExecutor class tbh, but I can also see it does simplify stuff a bit.. and it's all private so can still extract it without BC issues later if needed.

PR #11972, opened 2024-05-11
Close style tags to avoid colour bleed

fredden:
When I ran `composer require vendor/package:version` in order to get a specific version of a package, I noticed that the colour of the "normal" text was different from when I ran `composer require vendor/package`. I have tracked this down to an unterminated Symfony style tag used in the warning message about specifying a "too strict" restraint. I then scanned all of `src/` and found one other instance of this. This pull request fixes both instances.

Screenshot_2024-05-11_20-01-54


Seldaek:
Nice thanks

PR #11978, opened 2024-05-13
PHP 8.4 nullable type deprecation

sam-bee:
To avoid deprecation notices in PHP 8.4, use explicitly nullable types.

Similarly to the [pull request in the composer/pcre project](https://github.com/composer/pcre/pull/23), we get a deprecation notice for using `null` as a default value for non-nullable types in PHP 8.4.


Seldaek:
Those are in fixtures I don't think this is a real problem.

PR #11988, opened 2024-05-21
Add uopz warning from installer code

johnstevenson:
Syncing from https://github.com/composer/getcomposer.org/blob/main/web/installer

We seemed to have missed this one.


Seldaek:
Oh, thanks

PR #11990, opened 2024-05-27
Improve interactive package updates

jobvink:
An improvement on updating packages with the `--interactive` flag.

The last time I updated some dependencies, I had difficulty remembering which packages to update when running the update command. Adding a list of packages and their current and latest versions should help developers select the right package to update more easily.

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -31,6 +31,7 @@ use Composer\Repository\RepositorySet;
 use Composer\Semver\Intervals;
 use Composer\Util\HttpDownloader;
 use Composer\Advisory\Auditor;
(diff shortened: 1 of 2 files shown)


Seldaek:
I tweaked the code quite a bit to follow composer idiosyncrasies, but please have a look at the new UX to see if it still works as you'd expect, as I never use this interactive mode I am not the best judge I think.

jobvink:
> I tweaked the code quite a bit to follow composer idiosyncrasies, but please have a look at the new UX to see if it still works as you'd expect, as I never use this interactive mode I am not the best judge I think.

Yes, it still works as expected. The most important thing for me was the package versions.

Seldaek:
Alright cool, thanks

PR #11992, opened 2024-05-28
BlockedIPs: reject download jobs like other exceptions

glaubinix:
Just adjusting the blocked IP handling to match `MaxFileSizeExceededException`.


Seldaek:
Thanks

PR #11995, opened 2024-05-29
Make use of new PHP 8.4.0 function to replace implicit $http_response_header var

Seldaek:
Requires a new PHPStan release with https://github.com/phpstan/phpstan-src/pull/3221

### DIFF
--- a/phpstan/baseline.neon
+++ b/phpstan/baseline.neon
@@ -2046,11 +2046,6 @@ parameters:
 			path: ../src/Composer/Installer.php
 
 		-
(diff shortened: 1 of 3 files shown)


Seldaek:
@ondrejmirtes I see even with the release out now it still doesn't find these functions.. I guess the PHP_VERSION_ID check isn't enough to know that within the if the functions are available? If the target php version is 7.2 it assumes that everywhere right? If so I guess I'll baseline them for now.

ondrejmirtes:
I'm not sure about the reason, I usually work on new PHP version after festure freeze. Feel free to open a bug report about this 😊

PR #12000, opened 2024-05-31
Fix UX when a non-required plugin is still present in vendor dir

Seldaek:
Composer now skips it and does not prompt if it is not allowed to run, fixes #11944

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -20,11 +20,13 @@ use Composer\Package\BasePackage;
 use Composer\Package\CompletePackage;
 use Composer\Package\Locker;
 use Composer\Package\Package;
+use Composer\Package\RootPackageInterface;
 use Composer\Package\Version\VersionParser;
 use Composer\PartialComposer;
 use Composer\Pcre\Preg;
 use Composer\Repository\RepositoryInterface;
 use Composer\Repository\InstalledRepository;
+use Composer\Repository\RepositoryUtils;
 use Composer\Repository\RootPackageRepository;
 use Composer\Package\PackageInterface;
 use Composer\Package\Link;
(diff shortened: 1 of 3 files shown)


src/Composer/Plugin/PluginManager.php
```diff
@@ -462,10 +466,28 @@ private function loadRepository(RepositoryInterface $repo, bool $isGlobalRepo):
...
+            if (!in_array($package->getType(), ['composer-plugin', 'composer-installer'],true)) {
+                continue;
+            }
+
+            if (
+                !$isGlobalRepo
+                && !in_array($package, $requiredPackages, true)
+                && !$this->isPluginAllowed($package->getName(), false, true, false)
+            ) {
+                $this->io->writeError('<warning>The "'.$package->getName().'" plugin was not loaded as it is not in the allow-plugins rules and does not seem required by the root package anymore.</warning>');
```
naderman, line 487:
```suggestion
                $this->io->writeError('<warning>The "'.$package->getName().'" plugin was not loaded as it is not listed in allow-plugins and is not required by the root package anymore.</warning>');
```

I don't think we call it "allow-plugins rules" anywhere else? And otherwise let's make a definitive statement, either it is required or it's not, no such thing as "seeming required".

naderman (approved):

src/Composer/Plugin/PluginManager.php
```diff
@@ -462,10 +466,28 @@ private function loadRepository(RepositoryInterface $repo, bool $isGlobalRepo):
...
+        if (!$isGlobalRepo) {
+            $requiredPackages = RepositoryUtils::filterRequiredPackages($packages, $rootPackage, true);
+        }
+
         foreach ($sortedPackages as $package) {
             if (!($package instanceof CompletePackage)) {
                 continue;
             }
+
+            if (!in_array($package->getType(), ['composer-plugin', 'composer-installer'],true)) {
```
naderman, line 478:
```suggestion
            if (!in_array($package->getType(), ['composer-plugin', 'composer-installer'], true)) {
```
and here's a spacing fix ;-P

Issue #12002, opened 2024-05-31
PSR-0 namespaced classname resolving bug


ideus-biz:
Hi!
As to composer docs and the code&comments inside \Composer\Autoload\ClassLoader we can see that composer does support PSR-0 class naming with namespaces, but in fact such classes are not resolved.

Let's assume a vendor's library has a class:
`\MyHouse\Room\Shelf\Fruit_Apple`
located in vendor's folder `./src/Shelf/Fruit/Apple.php`

Vendor's `composer.json` contains:

```json
  "autoload": {
    "psr-0": {
      "MyHouse\\Room\\": "src/"
    },
  },
```
In such way composer searches the file in
`./vendor/...../src/MyHouse/Room/Shelf/Fruit/Apple.php`
but has to look up in
`./vendor/...../src/Shelf/Fruit/Apple.php`
due to composer.json psr-0 autoload directive.

The next code of \Composer\Autoload\ClassLoader::findFileWithExtension method is not sufficient and in fact does not assume autoload configuration:

```
if (isset($this->prefixesPsr0[$first])) {
            foreach ($this->prefixesPsr0[$first] as $prefix => $dirs) {
                if (0 === strpos($class, $prefix)) {
                    foreach ($dirs as $dir) {
                        if (file_exists($file = $dir . DIRECTORY_SEPARATOR . $logicalPathPsr0)) {
                            return $file;
                        }
                    }
                }
            }
        }
```

And it must be replaced with a similar code as above this one which seaches psr-4 class file taking in an account 'namespace' => 'directory' mapping, for example:

```
		if (isset($this->prefixesPsr0[$first])) {
			$subPath = $class;
			while (false !== $lastPos = strrpos($subPath, '\\')) {
				$subPath = substr($subPath, 0, $lastPos);
				$search = $subPath . '\\';
				if (isset($this->prefixesPsr0[$first][$search])) {
					$pathEnd = DIRECTORY_SEPARATOR . substr($logicalPathPsr0, $lastPos + 1);
					foreach ($this->prefixesPsr0[$first][$search] as $dir) {
						if (file_exists($file = $dir . $pathEnd)) {
							return $file;
						}
					}
				}
			}
		}
```
Which resolves file location correctly.

Please look up the issue and correct the code to respect PSR-0 namespaced classes, because composer states it support such class naming.
As well as our software needs such support.

Thank you.

Seldaek:
> In such way composer searches the file in
[...]

PR #12003, opened 2024-05-31
PSR-0 namespaced classname resolving bug

ideus-biz:
Suggested code to fix the issue #12002 (https://github.com/composer/composer/issues/12002)

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,4 +90,4 @@ Acknowledgments
 ---------------
 
 - This project's Solver started out as a PHP port of openSUSE's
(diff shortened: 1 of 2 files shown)


cafferata (changes requested):

stof:
this PR does not fix a bug. It **creates** a bug by making the `psr-0` config implement a mix between the PSR-0 and PSR-4 specs instead of implementing PSR-0.

Seldaek:
Closing per https://github.com/composer/composer/issues/12002#issuecomment-2163068066

PR #12016, opened 2024-06-18
Update OutdatedCommand.php: Adding `out` alias

ThomasLandauer:

### DIFF
--- a/src/Composer/Command/OutdatedCommand.php
+++ b/src/Composer/Command/OutdatedCommand.php
@@ -29,7 +29,8 @@ class OutdatedCommand extends BaseCommand
     {
         $this
             ->setName('outdated')
-            ->setDescription('[out] Shows a list of installed packages that have updates available, including their latest version')
+            ->setAliases(['out'])
+            ->setDescription('Shows a list of installed packages that have updates available, including their latest version')
             ->setDefinition([
                 new InputArgument('package', InputArgument::OPTIONAL, 'Package to inspect. Or a name including a wildcard (*) to filter lists of packages instead.', null, $this->suggestInstalledPackage(false)),
                 new InputOption('outdated', 'o', InputOption::VALUE_NONE, 'Show only packages that are outdated (this is the default, but present here for compat with `show`'),
--- a/src/Composer/Command/OutdatedCommand.php
+++ b/src/Composer/Command/OutdatedCommand.php
@@ -29,7 +29,7 @@ class OutdatedCommand extends BaseCommand
     {
         $this
             ->setName('outdated')
-            ->setDescription('Shows a list of installed packages that have updates available, including their latest version')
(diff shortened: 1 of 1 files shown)


xabbuh:
Explicitly defining the alias is technically not required due to how the Symfony Console component works when only parts of a command name are given.

ThomasLandauer:
I didn't really look into this, just noticed that `[out]` is kinda missing in the output of `composer`:

```
Available commands:
  about                    Shows a short information about Composer
  archive                  Creates an archive of this composer package
  audit                    Checks for security vulnerability advisories for installed packages
  auto-scripts             Runs the auto-scripts script as defined in composer.json
  browse                   [home] Opens the package's repository URL or homepage in your browser
  bump                     Increases the lower limit of your composer.json requirements to the currently installed versions
  check-platform-reqs      Check that platform requirements are satisfied
  clear-cache              [clearcache|cc] Clears composer's internal package cache
  completion               Dump the shell completion script
  config                   Sets config options
  create-project           Creates new project from a package into given directory
  depends                  [why] Shows which packages cause the given package to be installed
  diagnose                 Diagnoses the system to identify common errors
  dump-autoload            [dumpautoload] Dumps the autoloader
  exec                     Executes a vendored binary/script
  fund                     Discover how to help fund the maintenance of your dependencies
  global                   Allows running commands in the global composer dir ($COMPOSER_HOME)
  help                     Display help for a command
  init                     Creates a basic composer.json file in current directory
  install                  [i] Installs the project dependencies from the composer.lock file if present, or falls back on the composer.json
  licenses                 Shows information about licenses of dependencies
  list                     List commands
  outdated                 Shows a list of installed packages that have updates available, including their latest version
  prohibits                [why-not] Shows which packages prevent the given package from being installed
  reinstall                Uninstalls and reinstalls the given package names
  remove                   [rm|uninstall] Removes a package from the require or require-dev
  require                  [r] Adds required packages to your composer.json and installs them
  run-script               [run] Runs the scripts defined in composer.json
  search                   Searches for packages
  self-update              [selfupdate] Updates composer.phar to the latest version
  show                     [info] Shows information about packages
  status                   Shows a list of locally modified packages
  suggests                 Shows package suggestions
  update                   [u|upgrade] Updates your dependencies to the latest version according to composer.json, and updates the composer.lock file
  validate                 Validates a composer.json and composer.lock
```

stof:
Only 3 commands have explici shortcut aliases to ensure that those work to reference those commands even if the shortcut feature of symfony/console would not find an non-ambiguous command name (see `composer i` that would be ambiguous between `install` and `init`, same for `r` and `u` for `require` and `update`).
Other aliases are all about providing alternate names for some commands.

The help will not display every possible way to shorten a command name as this would be unreadable.

Seldaek:
Indeed, not really needed IMO once you know this trick.

PR #12019, opened 2024-06-24
fix(Locker): don't store transport-options.ssl within the lock-file

rtm-ctrlz:
## Description
Don't store `transport-options.ssl` within the lock-file otherwise the installation or updates will (in most cases) fail for repositories that use client TLS certificates, because during installation `transport-options` from `composer.lock` will be used instead of repository settings (certificates) from composer configuration file.

<hr />

## Example

Minimal configuration for case "private registry with client-certificates" (`~/.composer/config.json`)
```json
{
    "repositories": {
        "private.registry": {
            "type": "composer",
            "url": "https://composer.registry.example.com",
            "options": {
                "ssl": {
                    "local_cert": "/home/user/.composer/certs.d/private.cert",
                    "local_key": "/home/user/.composer/certs.d/private.key"
                }
            }
        }
    }
}
```

Dummy project (`~/project/composer.json`)
```json
{
    "name": "exmaple/dummy",
    "type": "project",
    "require": {
        "private-repo/package": "^1.0"
    }
}
```

Dummy project lock (`~/project/compose.lock`) (with current composer)
```json
{
    "content-hash": "...",
    "packages": [
        {
            "name": "private-repo/package",
            "version": "1.0.0",
            "transport-options": {
                "ssl": {
                    "local_cert": "/home/user/.composer/certs.d/private.cert",
                    "local_key": "/home/user/.composer/certs.d/private.key"
                }
            }
        }
    ]
}
```

In this case installation/update will fail if the environment performing such task will have different configuration (`~/.composer/config.json`), because path to certificate will be different and that will lead to cURL error message ("wrong certificate").

Some uses cases when this will could happen:
 - storing `composer.log` within VCS and 2+ developers with different composer configurations
 - CI installations

<hr/>

## Fixed lock-file

With this change lock-file (`~/project/compose.lock`) will look like this:
```json
{
    "content-hash": "...",
    "packages": [
        {
            "name": "private-repo/package",
            "version": "1.0.0"
        }
    ]
}
```
So composer will use TLS transport-options from configuration and installation will succeed.

## Is it breaking change?

I suppose this change as non-breaking, because it may will break in only one case - composer.lock file is somehow in sync with tls certificates in all places calling `install`/`update` and at the same time composer configuration still contains repository configuration, but is out of sync with certificates path; for me this approach is looking quite strange for me because it is "forming install environment around lock-file".

## Is it tested? - Yes

 - local tests with different TLS-configurations and other transport options
 -  within my company CI (patched version)

### DIFF
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -431,12 +431,11 @@ class Locker
 
             $spec = $this->dumper->dump($package);
             unset($spec['version_normalized']);
-            // remove `transport-options.ssl` from lock file to prevent storing registry configuration in the lock file
-            if (isset($spec['transport-options'])) {
-                if (isset($spec['transport-options']['ssl'])) {
-                    unset($spec['transport-options']['ssl']);
-                }
-                if (empty($spec['transport-options'])) {
+            // remove `transport-options.ssl` from lock file to prevent storing 
+            // local-filesystem repo config paths in the lock file as that makes it less portable
+            if (isset($spec['transport-options']['ssl'])) {
+                unset($spec['transport-options']['ssl']);
+                if (\count($spec['transport-options']) === 0) {
                     unset($spec['transport-options']);
                 }
             }
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -431,7 +431,7 @@ class Locker
 
             $spec = $this->dumper->dump($package);
             unset($spec['version_normalized']);
-            // remove transport-options.ssl from lock file
+            // remove `transport-options.ssl` from lock file to prevent storing registry configuration in the lock file
             if (isset($spec['transport-options'])) {
                 if (isset($spec['transport-options']['ssl'])) {
                     unset($spec['transport-options']['ssl']);
--- a/src/Composer/Package/Locker.php
+++ b/src/Composer/Package/Locker.php
@@ -431,6 +431,15 @@ class Locker
 
             $spec = $this->dumper->dump($package);
             unset($spec['version_normalized']);
+            // remove transport-options.ssl from lock file
+            if (isset($spec['transport-options'])) {
+                if (isset($spec['transport-options']['ssl'])) {
+                    unset($spec['transport-options']['ssl']);
+                }
+                if (empty($spec['transport-options'])) {
+                    unset($spec['transport-options']);
+                }
+            }
 
             // always move time to the end of the package definition
             $time = $spec['time'] ?? null;
--- a/phpstan/baseline.neon
+++ b/phpstan/baseline.neon
@@ -2707,7 +2707,7 @@ parameters:
 
 		-
 			message: "#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#"
-			count: 5
+			count: 6
 			path: ../src/Composer/Package/Locker.php
 
 		-
--- a/phpstan/baseline.neon
+++ b/phpstan/baseline.neon
@@ -2712,7 +2712,7 @@ parameters:
 
 		-
 			message: "#^Construct empty\\(\\) is not allowed\\. Use more strict comparison\\.$#"
-			count: 6
+			count: 5
 			path: ../src/Composer/Package/Locker.php
 
 		-


src/Composer/Package/Locker.php
```diff
@@ -431,6 +431,15 @@ private function lockPackages(array $packages): array
 
             $spec = $this->dumper->dump($package);
             unset($spec['version_normalized']);
+            // remove transport-options.ssl from lock file
```
staabm, line 434: The comment states whats obvious from the code but does not explain 'why'

rtm-ctrlz: Initially there was a detailed comment but It took several lines to have the answer.

But I agree that the current comment is meaningless, so I see some options:
1. remove this comment and let other developers try to guess the reason
2. put the full description there, so comment will have the answer
3. add link to this PR so the answer could be found

src/Composer/Package/Locker.php
```diff
@@ -431,6 +431,14 @@ private function lockPackages(array $packages): array
 
             $spec = $this->dumper->dump($package);
             unset($spec['version_normalized']);
+            // remove `transport-options.ssl` from lock file to prevent storing 
+            // local-filesystem repo config paths in the lock file as that makes it less portable
+            if (isset($spec['transport-options']['ssl'])) {
```
Seldaek, line 436: Note: you don't need to check with isset if transport-options is set first, then ssl, you can do both at once, `isset($random->deep['garbage']['not']->existing);` works fine :) https://3v4l.org/UAaJW

Seldaek:
Thanks yes that probably makes sense.

naderman:
It's entirely unclear to me how this is supposed to work. Once a package is in the lock file, you can't tell which repository it originally came from during a composer install from lock, so you can't tell what TLS options to apply.

All the settings necessary to download and install a package must be in the lock file, which of course includes TLS options. If you need to update TLS options in the lock file, this should be possible by running "composer update --lock" without modifying any versions.

Think this needs to be reverted. I don't understand the premise in this bug either. What's wrong with having the lock file point at those files? It's a very unusual set up to define repositories in a config file rather than the composer.json can you explain the use case there a bit more and what the problem even is?

Seldaek:
@rtm-ctrlz A few more questions:

- Do you just run `composer update` all the time? Or do you actually use the lock file?
- Did this patch make `composer install` break for you as the lock file is missing the cert auth info now?
- If you only run update, maybe you'd be better off without a lock file, by setting `composer config lock false` you can [disable lock file creation](https://getcomposer.org/doc/06-config.md#lock).

Please note if we do not get any feedback here we will have to revert this from the next Composer release.

rtm-ctrlz:
@Seldaek
 - > Do you just run `composer update` all the time? Or do you actually use the lock file?

   we actully use lock file
 - > Did this patch make `composer install` break for you as the lock file is missing the cert auth info now?

   no, because path to certificate is provided by `~/.composer/config.json`
 - > If you only run update, maybe you'd be better off without a lock file, by setting `composer config lock false` you can disable lock file creation.

   we need lock file, so we can't disable it

The goal of this change is not to store path of private certificates within `composer.lock`.

Our setup looks like this:
 - private package registry with TLS personal certificate authorisation
 - several developers, each has own location of the certificate, here is an example of `~/.composer/config.json`:
   ```
   "repositories": {
       "registry.out.company.tld": {
           "type": "composer",
           "url": "https://registry.out.company.tld",
           "options": {
               "ssl": {
                   "local_cert": "/home/user/.config/certs.d/composer.cert"
               }
           }
       }
   }
   ```
   we use `~/.composer/config.json` to specify the registry to keep it in one place, because we have number of private packages and in this case we need to specify it only one time, not for "any package that uses any of private packages", and it is more maintainable;
   storing registry configuration within `composer.json` will force to "hardcode" path to private certificate for all (developers and even CI)
 - code and `composer.lock` are stored with git
 - applications are "build" within CI, so we need to use `composer.lock`

Without this change we have several issues:
 - unnesesery chnages in `composer.lock` made by any developer that need to install (even not update) the dependencies after using `composer update --lock` and having a need to merge these changes over and over again
 - CI job "build application" doesn't have access to git for private packages, but has own private certificate and proper `~/.composer/config.json`, but `composer update --lock` fails in this case because it works like this:
   - tries to get package from registry (using certificate path from `composer.lock`) - in our case it will fail because path from lock points to "path on developers machine" and path from `~/.composer/config.json` isn't used
   - tries to get package from git - in our case it will fail because there is no access to git for this job


My point is that there no need to store tls-options with `composer.lock` while they are specified with `~/.composer/config.json`.

stof:
Does downloading dist archives from your custom repository requires to provide the certificate or only reading the metadata ?

rtm-ctrlz:
@stof both, in a nut shell there is NGinx with certificate verification serving static files generated by [composer/satis](https://github.com/composer/satis)

stof:
@rtm-ctrlz I suggest you to confirm whether this is actually true when performing an installation from lock file (without a cache of the downloaded archive). With this patch applied, there is no way for composer to associate the locked package to a TLS certificate it would add to send (as it is not associated to your custom repository in `~/.composer/config.json`)

rtm-ctrlz:
@stof having specified `local_cert` within lock file will force compose to use specified file as certificate for requests to registry:
  - registry requires client certificate
  - `composer install` (curl) can't find certificate specified by lock (because different environment ex. "developer vs CI")
  - composer failing to fetch package

I'm not sure on how can I confirm this behaviour.

stof:
@rtm-ctrlz your setup of having the repository specified outside the project is quite specific though. Wanting to have a reusable lock file based on *different* repository definitions looks weird.

And this PR still does not result in Composer passing the TLS certificate from the global config but passing no TLS certificate.

> I'm not sure on how can I confirm this behaviour.

Try downloading the archive with curl without providing a certificate, maybe.

rtm-ctrlz:
@stof Indeed there is a lack of our configuration and private registry worked without requiring certificate, thanks for pointing.

With our private registry got fixed (now it requires client certificate, as intended) I'm facing that certificate is used by composer only for getting metadata, but not the dist-files, so clean install (no `composer.lock`, no `vendor` directory, empty composer cache) fails.
I made a test with pre-PR version:
 - empty package (`compose init`) with single dependency from private registry
 - clean install works, path to certificate is stored in lock file
 - (no `vendor`, no composer cache, but having lock file with path to certificate) after changing certificate path (and updating `repositories`) running `composer update --lock` or `composer install` with both pre-PR and current version are failing to install dependency

rtm-ctrlz:
> setup of having the repository specified outside the project is quite specific though. 

I'd prefer to see it at "global composer configuration, for developer/CI" - most our work is done with private packages and I can't see the reason to "specify repository in composer.json for every package and in case of change update them all" instead of "setup workspace (composer) with repository configuration in one place and use for every package we develop".

> Wanting to have a reusable lock file based on _different_ repository definitions looks weird.

Indeed, we are trying to achieve (without success for now) configuration when composer is using path to certificate from composer's configuration, but not from lock file, because in our case (as I described above) using certificate from lock will lead to failures of installation (ex. two developers stores their certificates within home directory, but have different usernames, so the path to certificate will also be different).

rtm-ctrlz:
> This still expect that all developers have an equivalent repository defined in their global config instead of making the project self-contained.

Agree, but having less developers than packages makes updating developers configs easer than updating composer.json for every package.

I'll have a look towards implementing such authentication type for `auth.json`.

Seldaek:
Ok in the meantime I will revert this as the patch is clearly broken.

PR #12033, opened 2024-07-03
Remove redundant boolean type casts

seriquynh:
These parameters are declared as boolean and composer also uses `declare(strict_types=1);`. If we pass non-boolean values, it'll raise `TypeError`. So, we needn't to use type casts for them.


fredden:
`strict_types` isn't at play here. (That only matters for things that the file *calls*, not the values that *come in*.)

It's the parameter types which make these explicit casts redundant.

Seldaek:
Thanks

PR #12039, opened 2024-07-09
feat: Upgrade to json-schema v6

DannyvdSluijs:
This draft PR updates the `justinrainbow/json-schema` to `6.0.0-beta` as some early smoke testing.


Seldaek:
Oh.. new dependencies in v6, I hadn't noticed that until now. Not a fan of that, but so it goes. On the other hand Composer can probably stick to v5 for now as we don't really need anything else feature wise..

PR #12042, opened 2024-07-10
Fix addressability of branches containing # characters

### LINKED ISSUE #12029: Hard to understand error message when requiring a branch containing a hashtag
My `composer.json`:

```json
{
    "name": "jan/composer-require-test",
    "description": "",
    "type": "library",
    "repositories": [
        {
            "type": "path",
            "url": "/tmp/composer-version-test"
        }
    ],
    "require": {
        "jan/composer-version-test":"dev-foo/bar/baz/hello#123-lorem-ipsum"
    }
}
```

```
Composer could not detect the root package (jan/composer-require-test) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version
The package "jan/composer-version-test" is pointing to a commit-ref, this is bad practice and can cause unforeseen issues.
Checking git settings: OK git version 2.39.3
Checking disk free space: OK
Checking pubkeys:
OK
Checking Composer version: OK
Composer version: 2.7.7
PHP version: 8.3.8
PHP binary path: /opt/homebrew/Cellar/php/8.3.8/bin/php
OpenSSL version: OpenSSL 3.3.1 4 Jun 2024
curl version: 8.8.0 libz 1.2.12 ssl (SecureTransport) OpenSSL/3.3.1
zip: extension present, unzip present, 7-Zip present (7z)
```

When I run this command:

```
composer install -vvv
```

I get the following output: 

```
Running 2.7.7 (2024-06-10 22:11:12) with PHP 8.3.8 on Darwin / 23.5.0
Reading ./composer.json (/private/tmp/composer-require-test/composer.json)
Loading config file ./composer.json (/private/tmp/composer-require-test/composer.json)
Checked CA file /opt/homebrew/etc/ca-certificates/cert.pem: valid
Executing command (/private/tmp/composer-require-test): 'git' 'branch' '-a' '--no-color' '--no-abbrev' '-v'
Executing command (/private/tmp/composer-require-test): git describe --exact-match --tags
Executing command (CWD): git --version
Executing command (/private/tmp/composer-require-test): git log --pretty="%H" -n1 HEAD --no-show-signature
Executing command (/private/tmp/composer-require-test): hg branch
Executing command (/private/tmp/composer-require-test): fossil branch list
Executing command (/private/tmp/composer-require-test): fossil tag list
Executing command (/private/tmp/composer-require-test): svn info --xml
Composer could not detect the root package (jan/composer-require-test) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version
Failed to initialize global composer: Composer could not find the config file: /Users/jan/.composer/composer.json

No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
Executing command (/private/tmp/composer-version-test/): git log -n1 --pretty=%H --no-show-signature
Built pool.
Running pool optimizer.
Updating dependencies
Generating rules
Looking at all rules.

Reading /Users/jan/Library/Caches/composer/repo/https---repo.packagist.org/packages.json from cache
Downloading https://repo.packagist.org/packages.json if modified
[304] https://repo.packagist.org/packages.json
Downloading https://repo.packagist.org/p2/jan/composer-version-test.json
Downloading https://repo.packagist.org/p2/jan/composer-version-test~dev.json
[404] https://repo.packagist.org/p2/jan/composer-version-test.json
[404] https://repo.packagist.org/p2/jan/composer-version-test~dev.json

  Problem 1
    - Root composer.json requires jan/composer-version-test dev-foo/bar/baz/hello#123-lorem-ipsum, found jan/composer-version-test[dev-foo/bar/baz/hello#123-lorem-ipsum] but it does not match the constraint.
```

For a user, that is unfamiliar with the `require` syntax, it seems like the package is found in the required version but it is unclear why it does not match the constraint.

This happened to us today and we were wondering for about an hour why the package cannot be required.

The problem message brought us to the wrong conclusions.  
We saw that `jan/composer-version-test` was to be correctly required in version `dev-foo/bar/baz/hello#123-lorem-ipsum`. The version is a feature branch, that is named after the GitLab issue reference number (`<path/to/project>#<issue-id>`).  
The following statement 

> found jan/composer-version-test[dev-foo/bar/baz/hello#123-lorem-ipsum]

made us question our sanity. Why would Composer not use the required branch, if it is able to find it in the repository and happily prints it in the problem message?

After about an hour it dawned on me that the `#` in the branch name - and thus in the required version - makes composer look for the commit with the hash `123-lorem-ipsum` inside the branch `dev-foo/bar/baz/hello`.

A more explanatory message would've saved us a lot of time.  
Maybe something like:

```diff
-Root composer.json requires jan/composer-version-test dev-foo/bar/baz/hello#123-lorem-ipsum, found jan/composer-version-test[dev-foo/bar/baz/hello#123-lorem-ipsum] but it does not match the constraint.
+Root composer.json requires jan/composer-version-test dev-foo/bar/baz/hello at 123-lorem-ipsum, found jan/composer-version-test[dev-foo/bar/baz/hello#123-lorem-ipsum] but it does not match the constraint.
```

Looking back at the part of the problem message that says 

> it does not match the constraint

and the output of `composer validate` that says

> The package "jan/composer-version-test" is pointing to a commit-ref, this is bad practice and can cause unforeseen issues.

we should have noticed earlier, that we are doing something wrong.

I'd just wanted to let you know about this "problem" that caused some confusion for two developers that have been using composer for over five years.

And sorry for the long description, we really don't want to steal your time.
Keep up the awesome work on Composer, we love using it!

Seldaek:
Fixes #12029

This makes sure that VCS branches containing `#` signs get it replaced by a `+`. That makes them addressable (right now they are not at all).

And also it adds proper error reporting suggesting a solution if someone requires `dev-foo#bar` meaning to require that branch and not `dev-foo` at ref `bar`.

### DIFF
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -25,6 +25,7 @@ use Composer\Semver\Constraint\Constraint;
 use Composer\Semver\Constraint\ConstraintInterface;
 use Composer\Package\Version\VersionParser;
 use Composer\Repository\PlatformRepository;
+use Composer\Semver\Constraint\MultiConstraint;
 
 /**
  * Represents a problem detected while solving dependencies
@@ -259,6 +260,17 @@ class Problem
                     return ["- ", $package->getPrettyName().' is fixed to '.$package->getPrettyVersion().' (lock file version) by a partial update but that version is rejected by your minimum-stability. Make sure you list it as an argument for the update command.'];
                 }
                 break;
+            }
+        }
+
+        if ($constraint instanceof Constraint && $constraint->getOperator() === Constraint::STR_OP_EQ && Preg::isMatch('{^dev-.*#.*}', $constraint->getPrettyString())) {
+            $newConstraint = Preg::replace('{ +as +([^,\s|]+)$}', '', $constraint->getPrettyString());
+            $packages = $repositorySet->findPackages($packageName, new MultiConstraint([
+                new Constraint(Constraint::STR_OP_EQ, $newConstraint),
+                new Constraint(Constraint::STR_OP_EQ, str_replace('#', '+', $newConstraint))
+            ], false));
+            if (\count($packages) > 0) {
+                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but '.(self::hasMultipleNames($packages) ? 'these' : 'it').' had a # in the branch name which was replaced by +. Make sure to require it as "'.str_replace('#', '+', $constraint->getPrettyString()).'".'];
             }
         }
 
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -270,7 +270,7 @@ class Problem
                 new Constraint(Constraint::STR_OP_EQ, str_replace('#', '+', $newConstraint))
             ], false));
             if (\count($packages) > 0) {
-                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but '.(self::hasMultipleNames($packages) ? 'these' : 'it').' had a # in the branch name which was replaced by +. Make sure to require it as "'.str_replace('#', '+', $constraint->getPrettyString()).'".'];
+                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).'. The # character in branch names is replaced by a + character. Make sure to require it as "'.str_replace('#', '+', $constraint->getPrettyString()).'".'];
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Composer/DependencyResolver/Problem.php
```diff
@@ -262,6 +263,17 @@ public static function getMissingPackageReason(RepositorySet $repositorySet, Req
...
         }
 
+        if ($constraint instanceof Constraint && $constraint->getOperator() === Constraint::STR_OP_EQ && Preg::isMatch('{^dev-.*#.*}', $constraint->getPrettyString())) {
+            $newConstraint = Preg::replace('{ +as +([^,\s|]+)$}', '', $constraint->getPrettyString());
+            $packages = $repositorySet->findPackages($packageName, new MultiConstraint([
+                new Constraint(Constraint::STR_OP_EQ, $newConstraint),
+                new Constraint(Constraint::STR_OP_EQ, str_replace('#', '+', $newConstraint))
+            ], false));
+            if (\count($packages) > 0) {
+                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but '.(self::hasMultipleNames($packages) ? 'these' : 'it').' had a # in the branch name which was replaced by +. Make sure to require it as "'.str_replace('#', '+', $constraint->getPrettyString()).'".'];
```
naderman, line 273:
```suggestion
                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but '.(self::hasMultipleNames($packages) ? 'these' : 'it').'. The # character in branch names is replaced by a + character. Make sure to require "'.str_replace('#', '+', $constraint->getPrettyString()).'".'];
```

It's not really "but" cause the found thing lists the branch with a +

naderman (approved):

PR #12045, opened 2024-07-13
Update phpstan and add composer/pcre extensions

Seldaek:
Requires these to be fixed for all remaining errors to go away:

- [x] https://github.com/phpstan/phpstan/issues/11331
- [x] https://github.com/phpstan/phpstan/issues/11332 
- [x] https://github.com/phpstan/phpstan/issues/11323

Last blocker is to have a composer/pcre release and a phpstan release.

### DIFF
--- a/src/Composer/Downloader/GitDownloader.php
+++ b/src/Composer/Downloader/GitDownloader.php
@@ -255,7 +255,7 @@ class GitDownloader extends VcsDownloader implements DvcsDownloaderInterface
         }
 
         $headRef = $match[1];
-        if (!Preg::isMatchAllStrictGroups('{^'.$headRef.' refs/heads/(.+)$}mi', $refs, $matches)) {
+        if (!Preg::isMatchAllStrictGroups('{^'.preg_quote($headRef).' refs/heads/(.+)$}mi', $refs, $matches)) {
             // not on a branch, we are either on a not-modified tag or some sort of detached head, so skip this
             return null;
         }
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "symfony/finder": "^5.4 || ^6.0 || ^7",
         "symfony/process": "^5.4 || ^6.0 || ^7",
         "react/promise": "^2.8 || ^3",
-        "composer/pcre": "^2.1 || ^3.1",
+        "composer/pcre": "^2.2@dev",
         "symfony/polyfill-php73": "^1.24",
         "symfony/polyfill-php80": "^1.24",
         "symfony/polyfill-php81": "^1.24",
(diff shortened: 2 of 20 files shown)


Seldaek:
cc @staabm it's all coming together here :)

src/Composer/Downloader/GitDownloader.php
```diff
@@ -255,7 +255,7 @@ public function getUnpushedChanges(PackageInterface $package, string $path): ?st
         }
 
         $headRef = $match[1];
-        if (!Preg::isMatchAllStrictGroups('{^'.$headRef.' refs/heads/(.+)$}mi', $refs, $matches)) {
+        if (!Preg::isMatchAllStrictGroups('{^'.preg_quote($headRef).' refs/heads/(.+)$}mi', $refs, $matches)) {
```
Seldaek, line 258: @staabm sorry could you try to look why this happens?

I get this back here as pattern type https://github.com/composer/pcre/blob/2.x/src/PHPStan/UnsafeStrictGroupsCallRule.php#L78

```
43-object(PHPStan\Type\IntersectionType)#168717 (2) {
44-  ["types":"PHPStan\Type\IntersectionType":private]=>
45-  array(2) {
46-    [0]=>
47-    object(PHPStan\Type\Accessory\AccessoryNonFalsyStringType)#168719 (0) {
48-    }
49-    [1]=>
50-    object(PHPStan\Type\StringType)#168715 (0) {
51-    }
52-  }
53-  ["sortedTypes":"PHPStan\Type\IntersectionType":private]=>
54-  bool(false)
55-}
```

Not sure wtf this means..  but it then yields a null $matchedType. And I don't get it because doing this seems to work:

```
	if (preg_match('{^'.preg_quote($headRef).' refs/heads/(.+)$}mi', $s, $matches)) {
		assertType('array{string, non-empty-string}', $matches);
	}
```

staabm: for `preg_quote`  support the extension needs to use the newer `matchExpr` api instead of `matchType` (in all ArrayShapeMachter calling classes)

e.g https://github.com/composer/pcre/blob/bd3aec33e8d3fb8103b66678ef8b9ccac2f04178/src/PHPStan/UnsafeStrictGroupsCallRule.php#L80C50-L80C59

staabm: IIRC the new aapi is not yet in a phpstan release

Seldaek: Thanks, fixed now! Green build :) Indeed phpstan release is the last blocker left 🥳

staabm:
love it!

PR #12046, opened 2024-07-13
Fix deprecation notice on PHP 8.4

ADmad:
PHP 8.4 requires nullable arguments to be explicitly declared as such.


Ayesh:
I've been getting this notice, and can confirm that this PR indeed looks good to me as well.

Seldaek:
Thanks, I also adjusted our CI so we catch these faster in the future https://github.com/composer/composer/pull/12058

PR #12065, opened 2024-07-30
Add Update Interactive tests

mzhubail:
re: #10796

Added two cases for update command interactive mode.

### DIFF
--- a/tests/Composer/Test/Command/UpdateCommandTest.php
+++ b/tests/Composer/Test/Command/UpdateCommandTest.php
@@ -35,7 +35,7 @@ class UpdateCommandTest extends TestCase
         self::assertStringMatchesFormat(trim($expected), trim($appTester->getDisplay(true)));
     }
 
-    public static function provideUpdates(): \Generator
+    public static function provideUpdates(): Generator
     {
         $rootDepAndTransitiveDep = [
             'repositories' => [
@@ -142,7 +142,6 @@ OUTPUT
     /**
      * @dataProvider provideInteractiveUpdates
      * @param array<mixed> $packageNames
-     * @param string $expected
      */
     public function testInteractiveTmp(array $packageNames, string $expected): void
     {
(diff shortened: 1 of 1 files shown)


mzhubail:
Regarding the failing windows test, It seems like CRLF is causing the issue, any pointers on fixing it?

Regarding the 8.4 test, I don't know what to do.

Seldaek:
Sorry but I'm off atm and can't help much but 8.4 dont worry about it that's not your code, and windows same it's usually easy to fix.

mzhubail:
Good to hear.

Just wanted to see that things are good overall, before I start working on other command.

tests/Composer/Test/Command/UpdateCommandTest.php
```diff
@@ -12,7 +12,10 @@
 
 namespace Composer\Test\Command;
 
+use Composer\Package\Link;
+use Composer\Semver\Constraint\MatchAllConstraint;
 use Composer\Test\TestCase;
+use Generator;
```
stof, line 18: AFAIK, composer follows the Symfony coding standard of not import classes of the global namespace with use statements. You should respect that rule.

tests/Composer/Test/Command/UpdateCommandTest.php
```diff
@@ -135,4 +138,97 @@ public function testInteractiveModeThrowsIfNoPackageEntered(): void
...
+        $appTester->setInputs(array_merge(array_keys($packageNames), ['', 'yes']));
+        $appTester->run([
+            'command' => 'update', '--interactive' => true,
+            '--no-audit' => true,
+            '--dry-run' => true,
+        ]);
+
+        self::assertStringEndsWith(
+            trim($expected),
+            trim($appTester->getDisplay())
```
stof, line 204: to solve the tests on Windows, use `$appTester->getDisplay(true)` to make it normalize line breaks in the output (as done in other tests)

PR #12066, opened 2024-07-31
 Add `InitCommand.php` Interactive tests 

mzhubail:
This is to expand test coverage as explained in #10796.

### Why `generateInteractiveInput`

Since `$appTester->inputs` takes in all the inputs as an array, all inputs have
to be provided and all must be in correct order.  I thought it would be
appropriate to create a helper method that would do that.

- You don't have to provide all the data for each test case
  (`generateInteractiveInput` will pull the defaults for you).
- Less of the code would have to be changed in case the prompting process
  changes for `init` command, whether the prompts be reorderd or new prompts
  added.

------

You describe inputs by mapping the input name to an array of inputs.

```php
[
    'package_name' => ['vendor/my-package'],
    'minimum-stability' => 'stable',
]

[
    'package_name' => ['vendor', ''],
    'minimum-stability' => 'stable',
]
```

The reason I used array of inputs is to allow testing for cases where you'd have
to provide more than one input (think adding requirements or testing invalid
package name.

### DIFF
--- a/tests/Composer/Test/Command/InitCommandTest.php
+++ b/tests/Composer/Test/Command/InitCommandTest.php
@@ -748,8 +748,9 @@ class InitCommandTest extends TestCase
         foreach (array_keys($default_inputs) as $key) {
             $flat_inputs = array_merge(
                 $flat_inputs,
-                $inputs[$key] ?? $default_inputs[$key],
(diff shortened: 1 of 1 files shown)


mzhubail:
I'll be using a generator to have much more cases.  But that'll do for today.

PR #12067, opened 2024-08-01
Add missing semicolon to code example in documentation

viktor-kup:
Adds missing semicolon to a code example, which cannot be executed otherwise.


viktor-kup:
First time ever contributing, sorry if I made any mistake in the process.

Seldaek:
Thanks

PR #12072, opened 2024-08-04
Add "require command with conflicting keys" test

mzhubail:
re: #10796

Added four cases for conflicting keys

- with package in require or in require-dev
- with interactive or non interactive

For the interactive mode I'm testing for 'yes' input only.  I can test for 'no' too, but I didn't find an obvious way to do it without making the code too unreadable.

For the way I wrote the command part, I had to structure the code in that weird way to get it to work.  See commit message of d6479b8a for more.


Seldaek:
Thanks

Issue #12078, opened 2024-08-06
dev-main found instead of dev-master


iranimij:
In 2021 there was an issue https://github.com/composer/composer/issues/10360 and the fix for this issue made an exact opposite issue for us, we have `dev-master` and now composer just found `dev-main`

My `composer.json`:

```json
{
    "name": "foobar/project",
    "type": "project",
    "require": {
        "foobar/package": "dev-master"
    },
    "repositories": {
        "foobar/package": {
            "type": "path",
            "url": "/path/to/package/"
        }
    }
}
```

/path/to/package/composer.json
```json
{
    "name": "foobar/package",
    "type": "library"
}
```

Output of `composer diagnose`:

```
$ composer update
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires foobar/package dev-main, found foobar/package[dev-master] but it does not match the constraint.
```

When I run this command: 

```
composer update foobar/package
```

I get the following output: 

```
$ composer update
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires foobar/package dev-master, found foobar/package[dev-main] but it does not match the constraint.
```

And I expected this to happen:
I expect it to work fine without any errors

there was a fix [here](https://github.com/composer/composer/pull/10372/commits/714d68c0e58aa91ef595743fd43313305958fe0b) to have aliases for branches I do not know why it is removed [here](https://github.com/composer/composer/commit/2b99d069b22557cf87f3caacfb50251e3cf6c2ee) : 
and when we are removing a PR changes I expect to revert everything, in the previous PR it is changed to `dev-main` and when we removed the changes we didn't move it back to `dev-master`!

Seldaek:
The default version if no other version can be guessed is dev-main, so why not require dev-main in your main composer.json? Or you can set a version explicitly in path/to/package/composer.json

PR #12079, opened 2024-08-06
Revert code

iranimij:
(https://github.com/composer/composer/issues/12078)
There was a fix [here](https://github.com/composer/composer/pull/10372/commits/714d68c0e58aa91ef595743fd43313305958fe0b) to have aliases for branches and it is reverted [here](https://github.com/composer/composer/commit/2b99d069b22557cf87f3caacfb50251e3cf6c2ee) but we didn't change the default branch name

### DIFF
--- a/src/Composer/Repository/PathRepository.php
+++ b/src/Composer/Repository/PathRepository.php
@@ -218,7 +218,7 @@ class PathRepository extends ArrayRepository implements ConfigurableRepositoryIn
 
                     $package['version'] = $versionData['pretty_version'];
                 } else {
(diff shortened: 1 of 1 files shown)


stof:
This change will break all projects expecting the default version of a PathRepository to be named `dev-main`. So this is a no-go.

stof:
Changing it now would break all projects that rely on the Composer 2.2+ behavior (i.e. a behavior released in December 2021).
To be, this PR is likely to impact much more projects than it would fix (because many projects relying on the default will already have done the switch).

PR #12086, opened 2024-08-19
Add a way to control which scripts get args and where

### LINKED ISSUE #10734: Feature Request: Add possibility to define which script should get forwarded arguments
My `composer.json`:

```json
    "scripts": {
        "phpstan": [
            "@php bin/websiteconsole cache:warmup --env=dev",
            "@php vendor/bin/phpstan analyze"
        ],
    }
```

<details>

<summary>Output of `composer diagnose`:</summary>

```
Checking composer.json: OK
Checking git settings: OK
Checking disk free space: OK
Checking pubkeys:
OK
Checking composer version: OK
Composer version: 2.3.5
PHP version: 8.1.2 - Package overridden via config.platform, actual: 8.1.4
PHP binary path: /opt/homebrew/Cellar/php/8.1.4/bin/php
OpenSSL version: OpenSSL 1.1.1m  14 Dec 2021
cURL version: 7.82.0 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1n
```

</details>

When I run this command: 

```bash
composer phpstan -- src/Kernel.php
```

I get the following output: 

```bash
> @php bin/websiteconsole cache:warmup --env=dev 'src/Kernel.php'


  No arguments expected for "cache:warmup" command, got "src/Kernel.php".


cache:warmup [--no-optional-warmers]

Script @php bin/websiteconsole cache:warmup --env=dev handling the phpstan event returned with error code 1
```

The error totally make sense. It would be nice to have possibility to define which subcommand should get the arguments. E.g.:

```json
    "scripts": {
        "phpstan": [
            "@php bin/websiteconsole cache:warmup --env=dev",
            "@php vendor/bin/phpstan analyze {args}"
        ],
    }
```

Or something like:

```json
    "scripts": {
        "phpstan": {
            "beforeScripts": [
                "@php bin/websiteconsole cache:warmup --env=dev",
            ],
            "run": [
                "@php vendor/bin/phpstan analyze"
            ]
        },
    }
```

Also there are cases which make sense there that multple entries get the given argument e.g.:

```bash
composer fix -- src/Kernel.php
```

```json
    "scripts": {
        "fix": {
            "@rector",
            "@php-cs-fix"
        },
        "rector": "@php vendor/bin/rector process",
        "php-cs-fix": "@php vendor/bin/php-cs-fixer fix",
    }
```

What do you think about `{args}` or something similar maybe `{composer_args}` (avoid conflicts)  to make this possible?

Seldaek:
fixes #10734

This is so far fairly experimental/MVP.

Using the example from https://github.com/composer/composer/issues/10734#issuecomment-1179164480 as a base, I got this test case:

```
{
    "scripts": {
        "command-name": [
            "@composer install --no-dev @no_additional_args",
            "@php runner.php command-name @additional_args",
            "@php runner.php @additional_args command-name"
        ]
    }
}
```

Then running that script outputs:

```
 c run command-name -vv -- foo bar
> command-name: @composer install --no-dev
Installing dependencies from lock file
[...]
> command-name: @php runner.php command-name 'foo' 'bar'
> command-name: @php runner.php 'foo' 'bar' command-name
```

### DIFF
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -410,6 +410,38 @@ JSON array of commands.
 You can also call a shell/bash script, which will have the path to
 the PHP executable available in it as a `PHP_BINARY` env var.
 
+## Controlling additional arguments
+
+When running scripts like `composer script-name arg arg2` or `composer script-name -- --option`,
+Composer will by default append `arg`, `arg2` and `--option` to the script's command.
+
+If you do not want these args in a given command, you can put `@no_additional_args`
+anywhere in it, that will remove the default behavior and that flag will be removed
+as well before running the command.
+
+If you want the args to be added somewhere else than at the very end, then you can put
+`@additional_args` to be able to choose exactly where they go.
+
+For example running `composer run-commands ARG` with the below config:
+
+```json
+{
+    "scripts": {
+        "run-commands": [
+            "echo hello @no_additional_args",
+            "command-with-args @additional_args && do-something-without-args --here"
+        ]
+    }
+}
+```
+
+Would end up executing these commands:
+
+```
+echo hello
+command-with-args ARG && do-something-without-args --here
+```
+
 ## Setting environment variables
 
 To set an environment variable in a cross-platform way, you can use `@putenv`:
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -204,7 +204,7 @@ class EventDispatcher
 
                 $additionalArgs = $event->getArguments();
                 if (is_string($callable) && str_contains($callable, '@no_additional_args')) {
-                    $callable = str_replace('@no_additional_args', '', $callable);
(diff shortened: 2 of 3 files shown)


naderman:
I think the magical behavior of not sending args if at least one command in the script requires args explicitly is going to be too confusing. It'll be odd that you remove that one command and suddenly all commands receive args. It's also inconsistent then with how this would work for single command scripts, and I'd rather keep BC here. So I think it's fine to require explicit @no_additional_args rather than magically removing them if something else explicitly asks for them.

kaznovac:
amazing, thanks @Seldaek 

just my two cents - this can be expanded with shell-like syntax for arguments (I find `_additional` a bit unnecessary)

```json
{
    "scripts": {
        "command-name": [
            "@composer install --no-dev @no_args",
            "@php runner.php command-name @args[1]",
            "@php runner.php @args[1] @args[2] command-name #first two args",
            "@php runner.php @args[*] command-name #all args",
            "@php runner.php @args[0] #the name of the command"
        ]
    }
}
```

PR #12088, opened 2024-08-21
Clean up md5/sha1 usages, upgrade algos where possible

Seldaek:


src/Composer/DependencyResolver/GenericRule.php
```diff
@@ -46,7 +46,7 @@ public function getLiterals(): array
      */
     public function getHash()
     {
-        $data = unpack('ihash', md5(implode(',', $this->literals), true));
+        $data = unpack('ihash', hash(PHP_VERSION_ID > 80100 ? 'xxh3' : 'sha1', implode(',', $this->literals), true));
```
stof, line 49: this should use a qualified name for the constant instead of relying on the global namespace fallback, to allow PHP to resolve it at compile time.

Seldaek: Ah yes thanks

Issue #12070, opened 2024-08-02
Add composer audit --ignore-abandoned or --abandoned=ignore


nicolas-grekas:
That'd make it easier to integrate `composer audit` into CIs (the env var is not always as convenient nor explicit depending on how they're setup on CIs)

mzhubail:
I think this can be added with the precedence of "command line option" > "environment variable" > "config".

I can pick this up, but I'll wait first for the project maintainers assessment.

Seldaek:
Yup sounds good to me. If you want to PR please do :)

Seldaek:
Fixed by https://github.com/composer/composer/pull/12091

PR #12091, opened 2024-08-22
Add `--abandoned` option

mzhubail:
re: #12070

### DIFF
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -47,6 +47,7 @@ class Auditor
     public const ABANDONED_REPORT = 'report';
     public const ABANDONED_FAIL = 'fail';
 
+    /** @internal */
     public const ABANDONEDS = [
         self::ABANDONED_IGNORE,
         self::ABANDONED_REPORT,
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -47,6 +47,12 @@ class Auditor
     public const ABANDONED_REPORT = 'report';
     public const ABANDONED_FAIL = 'fail';
 
+    public const ABANDONEDS = [
+        self::ABANDONED_IGNORE,
+        self::ABANDONED_REPORT,
+        self::ABANDONED_FAIL,
+    ];
+
     /**
      * @param PackageInterface[] $packages
      * @param self::FORMAT_* $format The format that will be used to output audit results.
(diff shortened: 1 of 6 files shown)


src/Composer/Advisory/Auditor.php
```diff
@@ -47,6 +47,12 @@ class Auditor
     public const ABANDONED_REPORT = 'report';
     public const ABANDONED_FAIL = 'fail';
 
+    public const ABANDONEDS = [
+        self::ABANDONED_IGNORE,
+        self::ABANDONED_REPORT,
+        self::ABANDONED_FAIL,
+    ];
+
```
mzhubail, line 55: I'm not sure what to name this.  I tried to use `ABANDONED_VALUES` but that would cause issues with the type `self::ABANDONED_*`.

mzhubail:
- For now I'm not sure if this option should be added to all commands that have auditing (similar to how `--audit-format` is implemented.  I'd like input on that.  If it won't be included in all commands, I will just move the body of `getAuditAbandoned` in `AuditCommand`.

- In `AuditCammand`, I removed null coalescing from `$auditConfig['abandoned']` because I thought it'd be cleaner to let `getAuditAbandoned` handle the default values.

    One way to tackle it is to modify `Composer/Config.php` to make the abandoned key always present, even if `null` in `array{ignore?: mixed, abandoned?: 'fail'|'ignore'|'report'|null}.`

src/Composer/Advisory/Auditor.php
Seldaek: Yeah naming here is hard.. But no biggie. Can you just mark it `@internal` so we can refactor this to an enum at some point?

And I'd also make use of it in Config.php line 443

Seldaek:
> * For now I'm not sure if this option should be added to all commands that have auditing (similar to how `--audit-format` is implemented.  I'd like input on that.  If it won't be included in all commands, I will just move the body of `getAuditAbandoned` in `AuditCommand`.

I think I would just add it to AuditCommand for now to avoid polluting everything with this option which I doubt will see widesrpead use.

> * In `AuditCammand`, I removed null coalescing from `$auditConfig['abandoned']` because I thought it'd be cleaner to let `getAuditAbandoned` handle the default values.
>
>   One way to tackle it is to modify `Composer/Config.php` to make the abandoned key always present, even if `null` in `array{ignore?: mixed, abandoned?: 'fail'|'ignore'|'report'|null}.`

You can't change that, you'd need to change ConfigReturnTypeExtension to somehow return a non-nullable abandoned key, because Config already ensures that abandoned is always set and not null, it's just that phpstan is not aware of it..

src/Composer/Advisory/Auditor.php
mzhubail: Noted

Seldaek (approved):
Looks good to me now thanks!

Issue #11017, opened 2022-08-23
Misleading error message regarding conflicts


devantoine:
My `composer.json`:

```json
{
    "name": "sylius/sylius-standard",
    "description": "Starting point for projects powered by Sylius eCommerce.",
    "keywords": ["symfony", "symfony2", "sylius", "distribution", "ecommerce", "cart", "shopping"],
    "type": "project",
    "license": "MIT",
    "homepage": "http://sylius.org",
    "authors": [
        {
            "name": "Paweł Jędrzejewski",
            "homepage": "http://pjedrzejewski.com"
        },
        {
            "name": "Sylius project",
            "homepage": "http://sylius.org"
        },
        {
            "name": "Community contributions",
            "homepage": "http://github.com/Sylius/Sylius/contributors"
        }
    ],
    "require": {
        "php": "^7.4",
        "algolia/search-bundle": "^3.1",
        "doctrine/data-fixtures": "^1.3",
        "friendsofsylius/sylius-import-export-plugin": "^0.15.0",
        "friendsofsymfony/ckeditor-bundle": "^2.0",
        "giggsey/libphonenumber-for-php": "^8.12",
        "hautelook/alice-bundle": "^2.3",
        "huttopia/console-bundle": "^1.1",
        "incenteev/composer-parameter-handler": "^2.1",
        "laminas/laminas-zendframework-bridge": "^1.5",
        "mangopay/php-sdk-v2": "^2.8",
        "nelmio/api-doc-bundle": "^3.4",
        "nelmio/cors-bundle": "^1.5",
        "oro/doctrine-extensions": "^2.0",
        "portphp/csv": "^1.1",
        "portphp/spreadsheet": "^1.0.1",
        "quotatis/common-php-lib": "^1.1.1",
        "quotatis/algolia-php-lib": "0.1.0",
        "quotatis/authentication-php-lib": "^1.0",
        "quotatis/communication-php-lib": "^1.0",
        "quotatis/i18n-php-lib": "^1.0",
        "quotatis/logger-php-lib": "^1.0.2",
        "quotatis/mailer-php-lib": "^1.2",
        "quotatis/rabbitmq-php-lib": "1.1.1",
        "sensio/framework-extra-bundle": "^5.3",
        "sylius-labs/sensio-distribution-bundle": "^6.0",
        "sylius/shop-api-plugin": "^1.0",
        "sylius/sylius": "1.8.12",
        "symfony/dotenv": "^5.4",
[...]

PR #12111, opened 2024-09-17
Sort problem rules

Seldaek:
Related to https://github.com/composer/composer/issues/11017#issuecomment-2324343575 but it doesn't really solve anything, just cosmetic ordering of rules. I think it does make things a little bit more readable as it groups stuff together, and makes you read them in an order which hopefully helps.. But of course this is to some extent arbitrary and sometimes probably won't help much or even make things worse.


naderman (approved):
As per chat, I think maybe some cases of deep nesting of requires this will make the requires listed in somewhat random (alphabetical) order, but the improvements from putting root statements at top, conclusions low should really help a lot. And then even if the rest isn't in logical order, having it alphabetical makes it easy to look up each rule, and there are more frequently many rules all impacing the situation listed in somewhat random order now anyway, so this should improve the output in most cases.

PR #12113, opened 2024-09-17
Add suggestions of provider packages for ext- and lib- packages

### LINKED ISSUE #11669: Suggest polyfill packages providing an extension when it is missing.
> Disclaimer: I'm using `mbstring` as an example here, but this applies to any extension.

When a package requires `ext-mbstring` and you don't have it, the output currently looks like that:

```

  Problem 1
    - Root composer.json requires PHP extension ext-mbstring * but it is missing from your system. Install or enable PHP's mbstring extension.

To enable extensions, verify that they are enabled in your .ini files:
    - /etc/php/8.2/cli/php.ini
You can also run `php --ini` in a terminal to see which files are used by PHP in CLI mode.
Alternatively, you can run Composer with `--ignore-platform-req=ext-mbstring` to temporarily ignore these required extensions.
```

Installing the mbstring extension is indeed the best solution in term of performance. However, there is another alternative if you cannot install the extension: installing a polyfill package that provides the API implemented in PHP. For instance, `symfony/polyfill-mbstring` provides it for `ext-mbstring`.

The providers-api-url of Packagist works fine for those extension packages. For instance, the list of packages providing `ext-mbstring` is at https://packagist.org/providers/ext-mbstring.json
So it would be great if Composer could load it and display an alternative solution based on installing a polyfill package.

Seldaek:
Fixes #11669

Example output:

```
Updating dependencies

  Problem 1
    - Root composer.json requires PHP extension ext-mbstring * but the ext-mbstring package is disabled by your platform config. Enable it again with "composer config platform.ext-mbstring --unset".

      Alternatively you can require one of these packages that provide the extension (or parts of it):
      - symfony/polyfill-mbstring Symfony polyfill for the Mbstring extension
      - protoqol/quo-php PHP - Quo package.
      - lucifer/domain 域名和url解析器
      - masteryoda101/database Symfony polyfill for the Mbstring extension
      - symfony70/polyfill-mbstring PHP 7.0 support for: Symfony polyfill for the Mbstring extension


To enable extensions, verify that they are enabled in your .ini files:
    - /etc/php/8.3/cli/php.ini
    - /etc/php/8.3/cli/conf.d/10-mysqlnd.ini
    - /etc/php/8.3/cli/conf.d/10-opcache.ini
    - /etc/php/8.3/cli/conf.d/10-pdo.ini
    - /etc/php/8.3/cli/conf.d/15-xml.ini
    - /etc/php/8.3/cli/conf.d/20-apcu.ini
    - /etc/php/8.3/cli/conf.d/20-bcmath.ini
    - /etc/php/8.3/cli/conf.d/20-bz2.ini
    - /etc/php/8.3/cli/conf.d/20-calendar.ini
```

### DIFF
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -278,14 +278,19 @@ class Problem
 
                 $version = self::getPlatformPackageVersion($pool, $packageName, phpversion($ext) ?: '0');
                 if (null === $version) {
+                    $providersStr = self::getProvidersList($repositorySet, $packageName, 5);
+                    if ($providersStr !== null) {
+                        $providersStr = "\n\n      Alternatively you can require one of these packages that provide the extension (or parts of it):\n$providersStr";
+                    }
+
                     if (extension_loaded($ext)) {
                         return [
                             $msg,
-                            'the '.$packageName.' package is disabled by your platform config. Enable it again with "composer config platform.'.$packageName.' --unset".',
+                            'the '.$packageName.' package is disabled by your platform config. Enable it again with "composer config platform.'.$packageName.' --unset".' . $providersStr,
                         ];
                     }
 
-                    return [$msg, 'it is missing from your system. Install or enable PHP\'s '.$ext.' extension.'];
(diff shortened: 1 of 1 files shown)


naderman (approved):
No test? ;-P

PR #12114, opened 2024-09-18
Add ability to reinstall packages by type

### LINKED ISSUE #11364: [FR] Add a `--type` option to the `reinstall` command
It would be nice if it were possible for the `reinstall` command to filter the searched packages by package type, via a new `--type` option.

```sh
composer reinstall "*" --type craft-plugin
```

That way, `reinstall` could be used to get custom package installers to do whatever custom stuff they need to do across all of the packages relevant to them (in Craft CMS’s case, updating `craftcms/plugins.php`) without any extra overhead.

Seldaek:
Fixes #11364

### DIFF
--- a/src/Composer/Command/ReinstallCommand.php
+++ b/src/Composer/Command/ReinstallCommand.php
@@ -51,7 +51,8 @@ class ReinstallCommand extends BaseCommand
                 new InputOption('apcu-autoloader-prefix', null, InputOption::VALUE_REQUIRED, 'Use a custom prefix for the APCu autoloader cache. Implicitly enables --apcu-autoloader'),
                 new InputOption('ignore-platform-req', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Ignore a specific platform requirement (php & ext- packages).'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore all platform requirements (php & ext- packages).'),
-                new InputArgument('packages', InputArgument::IS_ARRAY | InputArgument::REQUIRED, 'List of package names to reinstall, can include a wildcard (*) to match any substring.', null, $this->suggestInstalledPackage(false)),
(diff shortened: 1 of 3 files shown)


src/Composer/Command/ReinstallCommand.php
```diff
@@ -77,23 +78,38 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $localRepo = $composer->getRepositoryManager()->getLocalRepository();
         $packagesToReinstall = [];
         $packageNamesToReinstall = [];
-        foreach ($input->getArgument('packages') as $pattern) {
-            $patternRegexp = BasePackage::packageNameToRegexp($pattern);
-            $matched = false;
+        if (\count($input->getOption('type')) > 0) {
+            if (\count($input->getArgument('packages')) > 0) {
+                throw new \InvalidArgumentException('You cannot specify package names and filter by type at the same time.');
```
naderman, line 83: why not?

naderman (approved):

src/Composer/Command/ReinstallCommand.php
stof: what would be the effect ? Applying the type filter on the list of packages being provided or making a union of both filter results ?

naderman: That's the same thing?

naderman: I would expect the filters to both be applied, if I want a union i need to run the command twice with different arguments.

naderman: Practical example why I thought this may matter was just something like "composer update --type=mycms-plugin --packages=foovendor/*" if I want to update all the plugins from that vendor but not all their packages, but probably a pretty niche use case so whatever in the end ;-)

Seldaek: Yeah mostly I thought throwing is better than undefined behavior.. As i cannot be bothered to implement this edge case.

naderman: Sure, then throwing is fine :+1:

PR #12116, opened 2024-09-18
PHP 8.4 | Remove use of `E_STRICT`

jrfnl:
The `E_STRICT` constant is deprecated as of PHP 8.4 and will be removed in PHP 9.0 (commit finally went in today).

The error level hasn't been in use since PHP 8.0 anyway and was only barely still used in PHP 7.x, so removing the exclusion from the `error_reporting()` setting in these script shouldn't really make any difference in practice.

Ref:
* https://wiki.php.net/rfc/deprecations_php_8_4#remove_e_strict_error_level_and_deprecate_e_strict_constant

### DIFF
--- a/src/Composer/Util/ErrorHandler.php
+++ b/src/Composer/Util/ErrorHandler.php
@@ -78,7 +78,7 @@ class ErrorHandler
     public static function register(?IOInterface $io = null): void
     {
         set_error_handler([__CLASS__, 'handle']);
-        error_reporting(E_ALL | E_STRICT);
+        error_reporting(E_ALL);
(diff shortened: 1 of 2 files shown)


Seldaek:
Yay, thanks

sebastianbergmann:
@Seldaek Do you have an ETA for Composer 2.8? Thanks!

Seldaek:
I'll try this week and if not I guess i roll this in a 2.7 patch release

Seldaek:
It's out https://github.com/composer/composer/releases/tag/2.8.0

PR #12121, opened 2024-09-19
Fix type of $learnedLiterals to avoid issues

Seldaek:
It's probably the 3rd time I look at that code because of some PHPStan error caused by this weird array with `[null|int, ...int]` which cannot be expressed as a type.. So I finally decided to fix it. I'm not sure if you have a better name than $mainLiteral @naderman ?


naderman (approved):

src/Composer/DependencyResolver/Solver.php
```diff
@@ -423,16 +424,14 @@ protected function analyze(int $level, Rule $rule): array
...
 
                     if (!$l1num) {
                         break 2;
                     }
 
-                    foreach ($learnedLiterals as $i => $learnedLiteral) {
-                        if ($i !== 0) {
-                            unset($seen[abs($learnedLiteral)]);
-                        }
+                    foreach ($otherLearnedLiterals as $learnedLiteral) {
```
stof, line 433: you need to rename the loop variable so that it does not override the main learnedLiteral variable.

Seldaek:
Ok some scope creep later.. I reviewed and fixed all but 5 phpstan errors in the entire DependencyResolver namespace :)

stof (approved):

src/Composer/DependencyResolver/Solver.php
```diff
@@ -738,7 +733,7 @@ private function runSat(): void
                     }
                 }
 
-                if ($lastLiteral) {
+                if ($lastLiteral !== null && $lastLevel !== null) {
```
naderman, line 736: safer to assert lastLevel !== null inside the block than adding it to the condition and potentially skipping the code here in case of a bug?

naderman (approved):

PR #12122, opened 2024-09-19
Add --patch-only flag to update command to restrict updates to patch versions and make them safer

### LINKED ISSUE #11446: Enhance update command to only install patch level updates
### Discussed in https://github.com/composer/composer/discussions/11441

<div type='discussions-op-text'>

<sup>Originally posted by **MajPay** April 28, 2023</sup>
Hi!

I was wondering if there are any plans to enhance the update command with a --patch-level-only (example name) flag.

Why would i want something like that?

I know i could use "~1.0.0" to make sure only the patch-level will be updated but i don't want to exclude minor updates from the regular update command. I would need the --patch-level-only flag for automation (CI pipelines). What i am looking for is a way to automatically install updates on a daily basis without risking the app to fail.

minor and major updates often introduce breaking changes that require migrations. Minor updates should not do that but sometimes they do, while patch-level updates won't change any API in 99.9% of the time.</div>

Seldaek:
Fixes #11446

cc @MajPay as you requested the feature

Still needs tests to be written but otherwise implementation is ready to review.

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -28,6 +28,7 @@ use Composer\Repository\CompositeRepository;
 use Composer\Repository\PlatformRepository;
 use Composer\Repository\RepositoryInterface;
 use Composer\Repository\RepositorySet;
+use Composer\Semver\Constraint\MultiConstraint;
 use Composer\Semver\Intervals;
 use Composer\Util\HttpDownloader;
 use Composer\Advisory\Auditor;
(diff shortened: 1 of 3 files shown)


src/Composer/Command/UpdateCommand.php
```diff
@@ -175,6 +177,26 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                    continue;
+                }
+                $constraint = $parser->parseConstraints('~'.$match[1]);
+                if (isset($temporaryConstraints[$package->getName()])) {
+                    $temporaryConstraints[$package->getName()] = MultiConstraint::create([$temporaryConstraints[$package->getName()], $constraint], true);
+                } else {
+                    $temporaryConstraints[$package->getName()] = $constraint;
+                }
+            }
+        }
```
naderman, line 198: Sorry I've looked at this so little, the root requires constraints still get applied, right? So this doesn't allow you to update to a version in conflict with the composer.json constraint?

Should probably have a test for that situation? e.g. composer.json says ">=1.0.1,<=1.0.3" and that shouldn't update to 1.0.4.

Seldaek: Yeah yeah this just applies an additional constraint, much like --with

naderman (approved):

src/Composer/Command/UpdateCommand.php
naderman: Thanks, was thrown off by the logic for checking root requires for --with above, but that's just cause in that case you kind of want it to error if the thing you pick doesn't fit, vs here where it should just stick to what you already have in that case rather than error.

PR #12123, opened 2024-09-20
JSON schema updates

### LINKED ISSUE #7823: JSON Schema for composer.lock file
Hi,

there appears to be no JSON schema for composer.lock structure.
This is a bit problematic for tools building on top of the lockfile, be it for reading, manipulation, maintaining bi-directional compatibility or different use cases.

Would it be possible to add one?

Thanks.

Seldaek:
- Added composer-lock-schema
- Updated composer-repository-schema with new properties
- Added lock file schema validation in `diagnose`

Fixes #7823

### DIFF
--- a/res/composer-lock-schema.json
+++ b/res/composer-lock-schema.json
@@ -1,0 +1,154 @@
+{
+    "$schema": "https://json-schema.org/draft-04/schema#",
+    "title": "Composer Lock File",
+    "type": "object",
+    "required": [ "content-hash", "packages", "packages-dev" ],
+    "additionalProperties": true,
+    "properties": {
+        "_readme": {
+            "type": "array",
+            "items": {
+                "type": "string"
+            },
+            "description": "Informational text for humans reading the file"
+        },
+        "content-hash": {
+            "type": "string",
+            "description": "Hash of all relevant properties of the composer.json that was used to create this lock file."
+        },
+        "packages": {
+            "type": "array",
+            "description": "An array of packages that are required.",
+            "$ref": "#/definitions/versions"
+        },
+        "packages-dev": {
+            "type": "array",
+            "description": "An array of packages that are required in require-dev.",
+            "$ref": "#/definitions/versions"
+        },
+        "aliases": {
+            "type": "array",
+            "description": "Inline aliases defined in the root package.",
+            "items": {
+                "type": "object",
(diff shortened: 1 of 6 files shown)


res/composer-lock-schema.json
```diff
@@ -0,0 +1,154 @@
...
+                "type": { "type": "string" },
+                "version": { "type": "string" },
+                "version_normalized": {
+                    "type": "string",
+                    "description": "Normalized version, optional but can save computational time on client side."
+                },
+                "autoload": { "type": "object" },
+                "require": { "type": "object" },
+                "replace": { "type": "object" },
+                "conflict": { "type": "object" },
```
stof, line 121: should those define the `additionalProperties` as being strings ?

res/composer-lock-schema.json
```diff
@@ -0,0 +1,154 @@
...
+        "package-base": {
+            "properties": {
+                "name": { "type": "string" },
+                "type": { "type": "string" },
+                "version": { "type": "string" },
+                "version_normalized": {
+                    "type": "string",
+                    "description": "Normalized version, optional but can save computational time on client side."
+                },
+                "autoload": { "type": "object" },
```
stof, line 118: should this define the more precise type ?

res/composer-lock-schema.json
Seldaek: Yeah if we want to be more specific we could perhaps refer to the composer-schema.json for the entire packages contents. That'd be better but I just took this from the repo schema for now which is definitely not as complete.

On the other hand refs across files makes it a bit harder to use standalone so I am thinking it's probably good enough as is.. Ultimately the main point is to validate the lock file specific parts. Each package is then just a composer.json.

naderman: Think if we're adding this schema to make it easier to work with this file we should either make it accurate and list everything out here, or use the reference if you can't be bothered to copy stuff / maintain it in two places, but just waving it off with type object here seems like only doing half the task? ;-P

boesing: Maybe create a version which refers to the file and the dereference with a tool to the expected location in a build? That would be best of both worlds.

Seldaek: Ok I've added a ref.. It seems to work well enough so let's keep it simple for now :)

stof (approved):

PR #12136, opened 2024-09-30
Fix deprecated ParametersAcceptorSelector::selectSingle

jwpage:
This fixes a deprecation warning being reported by PHPStan in the `ConfigReturnTypeExtension`, as part of the bleeding edge rules:

```
  Line   Composer/PHPStan/ConfigReturnTypeExtension.php
 ------ ------------------------------------------------------------------------------------------------------------------------------
  68     Call to deprecated method selectSingle() of class PHPStan\Reflection\ParametersAcceptorSelector:
         See https://github.com/phpstan/phpstan-src/blob/2.0.x/UPGRADING.md#removed-deprecated-parametersacceptorselectorselectsingle
```

After this fix PHPStan analysis passes again.

### DIFF
--- a/src/Composer/PHPStan/ConfigReturnTypeExtension.php
+++ b/src/Composer/PHPStan/ConfigReturnTypeExtension.php
@@ -65,7 +65,11 @@ final class ConfigReturnTypeExtension implements DynamicMethodReturnTypeExtensio
     {
         $args = $methodCall->getArgs();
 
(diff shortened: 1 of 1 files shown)


staabm (commented):
The extension can be simplified by returning `null` in the default-return cases. 

That way it gets even simpler

Seldaek:
Is this what you meant @staabm ?

staabm:
Yep 👍

Seldaek:
Ok thanks to you both

PR #12146, opened 2024-10-03
Fix create-project to reuse the target folders permissions for files created

### LINKED ISSUE #9039: composer create-project doesn't correspond to the attributes set on parent directory

```
Checking git settings: OK
Checking disk free space: OK
Checking composer version: OK
Composer version: 1.10.8
PHP version: 7.4.3
OpenSSL version: OpenSSL 1.1.1f  31 Mar 2020
```

Let's create a new project ...

```
mkdir test
chgrp www-data test
chmod g+s test
composer create-project --prefer-dist laravel/laravel test
```

... and have a look at the contents of `test`:

```
$ ll test
total 328
drwxr-sr-x 12 marveloo www-data   4096 Jul  8 19:26 ./
drwxr-xr-x  8 marveloo marveloo   4096 Jul  8 19:25 ../
-rw-r--r--  1 marveloo marveloo    220 May 19 00:50 .editorconfig
-rw-r--r--  1 marveloo www-data    829 Jul  8 19:26 .env
-rw-r--r--  1 marveloo marveloo    778 May 19 00:50 .env.example
-rw-r--r--  1 marveloo marveloo    111 May 19 00:50 .gitattributes
-rw-r--r--  1 marveloo marveloo    163 May 19 00:50 .gitignore
-rw-r--r--  1 marveloo marveloo    174 May 19 00:50 .styleci.yml
-rw-r--r--  1 marveloo marveloo   4465 May 19 00:50 README.md
drwxr-xr-x  6 marveloo marveloo   4096 May 19 00:50 app/
-rwxr-xr-x  1 marveloo marveloo   1686 May 19 00:50 artisan*
drwxr-xr-x  3 marveloo marveloo   4096 May 19 00:50 bootstrap/
-rw-r--r--  1 marveloo marveloo   1586 May 19 00:50 composer.json
-rw-r--r--  1 marveloo www-data 227890 Jul  8 19:26 composer.lock
drwxr-xr-x  2 marveloo marveloo   4096 May 19 00:50 config/
drwxr-xr-x  5 marveloo marveloo   4096 May 19 00:50 database/
-rw-r--r--  1 marveloo marveloo   1034 May 19 00:50 package.json
-rw-r--r--  1 marveloo marveloo   1197 May 19 00:50 phpunit.xml
drwxr-xr-x  2 marveloo marveloo   4096 May 19 00:50 public/
drwxr-xr-x  6 marveloo marveloo   4096 May 19 00:50 resources/
drwxr-xr-x  2 marveloo marveloo   4096 May 19 00:50 routes/
-rw-r--r--  1 marveloo marveloo    563 May 19 00:50 server.php
drwxr-xr-x  5 marveloo marveloo   4096 May 19 00:50 storage/
drwxr-xr-x  4 marveloo marveloo   4096 May 19 00:50 tests/
drwxr-sr-x 44 marveloo www-data   4096 Jul  8 19:26 vendor/
-rw-r--r--  1 marveloo marveloo    538 May 19 00:50 webpack.mix.js
```

As we can see, only the `vendor` directory and two other files reflect the group and other attributes set on the parent directory, whilst others don't.

I can understand that probably under the hood most of the files ar copied or moved from some cache or temporary directory into `test` directory and this is probably the reason why these files preserve their attributes which were set there. However this is undesired behavior from the developer's perspective because all these files are new files for us, therefore we expect all these files to be copied into a new project with `--no-preserve=all` (or similar) option always on.

Seldaek:
Fixes #9039

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -98,7 +98,7 @@ class Config
 
     /** @var array<string, mixed> */
     private $config;
-    /** @var ?string */
+    /** @var ?non-empty-string */
     private $baseDir;
     /** @var array<int|string, mixed> */
     private $repositories;
@@ -137,6 +137,18 @@ class Config
         foreach ($this->repositories as $configKey => $configValue) {
             $this->setSourceOfConfigValue($configValue, 'repositories.' . $configKey, self::SOURCE_DEFAULT);
         }
+    }
+
+    /**
+     * Changing this can break path resolution for relative config paths so do not call this without knowing what you are doing
+     *
+     * The $baseDir should be an absolute path and without trailing slash
+     *
+     * @param non-empty-string $baseDir
+     */
+    public function setBaseDir(string $baseDir): void
+    {
+        $this->baseDir = $baseDir;
     }
 
     public function setConfigSource(ConfigSourceInterface $source): void
@@ -546,7 +558,7 @@ class Config
             return $path;
         }
 
-        return $this->baseDir ? $this->baseDir . '/' . $path : $path;
+        return $this->baseDir !== null ? $this->baseDir . '/' . $path : $path;
     }
 
     /**
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -144,9 +144,9 @@ class Config
      *
      * The $baseDir should be an absolute path and without trailing slash
      *
-     * @param non-empty-string $baseDir
+     * @param non-empty-string|null $baseDir
      */
-    public function setBaseDir(string $baseDir): void
+    public function setBaseDir(?string $baseDir): void
     {
         $this->baseDir = $baseDir;
     }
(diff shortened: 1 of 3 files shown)


src/Composer/Config.php
```diff
@@ -139,6 +139,18 @@ public function __construct(bool $useEnvironment = true, ?string $baseDir = null
...
     }
 
+    /**
+     * Changing this can break path resolution for relative config paths so do not call this without knowing what you are doing
+     *
+     * The $baseDir should be an absolute path and without trailing slash
+     *
+     * @param non-empty-string $baseDir
+     */
+    public function setBaseDir(string $baseDir): void
```
stof, line 149: shouldn't the setter allow setting it back to `null` ?

Seldaek: This really should never be set to null aside from the global config, but I fixed it just in case..

PR #12147, opened 2024-10-03
Ensure installed.php & InstalledVersions are in opcache when preloading is used

### LINKED ISSUE #11018: Disable installed.php file generation
Is there a way to disable this file from being generated? We currently don't have any use for it, nor do we want or need packages in our app to know what's installed and their versions. The `installed.php` file is absolutely huge on our app, adding unnecessary memory usage and resources to each request.

I checked the docs at: https://getcomposer.org/doc/07-runtime.md#installed-versions - however it seems this feature is non opt-in.

Would it be possible to add something to composer.json to disable it, something like:

```php
"installed-file": false
```

And maybe a command line option?

```bash
composer install --no-installed-file
```

Seldaek:
Fixes #11018 hopefully.

I am not sure about this patch tbh.. It'd be good to have feedback from people using preloading whether this looks like it is desired or not.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -1026,6 +1026,16 @@ PLATFORM_CHECK;
 
 CLASSLOADER_INIT;
 
+        $file .= <<<PRELOAD_INSTALLED_VERSIONS
+        // include InstalledVersions class & installed.php data to ensure it is in opcache when preloading
+        if (file_exists(__DIR__ . '/InstalledVersions.php') && !class_exists('Composer\\\\InstalledVersions', false)) {
+            include __DIR__ . '/InstalledVersions.php';
+            include __DIR__ . '/installed.php';
+        }
+
+
+PRELOAD_INSTALLED_VERSIONS;
+
         if ($useIncludePath) {
             $file .= <<<'INCLUDE_PATH'
         $includePaths = require __DIR__ . '/include_paths.php';
--- a/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
+++ b/tests/Composer/Test/Autoload/Fixtures/autoload_real_files_by_dependency.php
@@ -26,6 +26,12 @@ class ComposerAutoloaderInitFilesAutoloadOrder
         self::$loader = $loader = new \Composer\Autoload\ClassLoader(\dirname(__DIR__));
         spl_autoload_unregister(array('ComposerAutoloaderInitFilesAutoloadOrder', 'loadClassLoader'));
 
+        // include InstalledVersions class & installed.php data to ensure it is in opcache when preloading
+        if (file_exists(__DIR__ . '/InstalledVersions.php') && !class_exists('Composer\\InstalledVersions', false)) {
+            include __DIR__ . '/InstalledVersions.php';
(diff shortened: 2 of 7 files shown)


stof:
wouldn't this include it in on all requests even if InstalledVersions is not actually used, making things potentially worse ?

Seldaek:
Yes and no. I mean with opcache active and the file in cache, including this should be very near to a noop, but it is indeed somewhat wasteful to include it if it's never going to be used by the application.

That's why I am a bit unsure. On one hand it should be only people using preloading that should include these files manually.. But on the other hand it's sort of unreasonable to expect people to know about this, so we could do it for them.

I'm ok to close this as well tbh.. Really unsure which is the best solution.

stof:
The installed.php file does not contain any class definition, so there is nothing to preload anyway (having the file in cache is not the same than preloading). However, your change will make it being loaded (hopefully from cache) for each request.

And for InstalledVersions, I would rather let projects decide whether they preload it or no, like any other classes.

garygreen:
This PR has a condition of "when preloading is used" - how does composer know when it's being used? Is there a way currently in composer to generate an optimized preload file, possibly like the optimized dumpautoload so something like "composer --optimize-preloader" and it generates a file you can include in your preloader php file?

I think that would be a better option than having installed.php included in every single request. From what I'm seeing - uptake of php preloading is actually quite low (likely because it's tricky to configure, and pretty much impossible in shared environments) so worry if it's included by default for every single request it would impact memory usage.

So I'd prefer a way to generate an optimized preloader file, but ideally my preference would be to totally opt out of installed version information being generated at all `composer install --no-version-file` or something. Those could be two separate things: 1) optimized preloader file, 2) opt-out of installed version file.

stof:
@garygreen composer does not have such a preloader generator, because it misses context about what should be preloaded. The Symfony framework has benchmarked the effect of preloading everything (which composer could do) vs preloading only classes that are in the hot path of requests (which requires project-specific knowledge that composer does not have, but that Symfony can have for a Symfony-based project) and found out that preloading too much was decreasing performance compared to preloading only used code (it was still faster than preloading nothing though).

However, as I said previously, there is nothing to preload in the installed.php file as preloading is about class definitions.

Seldaek:
Yeah i think this isn't the right fix so I'll close.

Seldaek:
As for skipping installed versions generation.. This would break any code relying on it so I don't see this as a viable option. If you don't want to use it don't include/use it and make sure none of your deps use it I guess.

garygreen:
> However, as I said previously, there is nothing to preload in the installed.php file as preloading is about class definitions.

I'm curious to learn more about that. So if preloading focuses on classes and functions, presumably that means all of the below files are essentially incompatible with preloading (because they return basic arrays and not classes?)

- autoload_classmap.php (this one is particularly large)
- autoload_files.php
- autoload_namespaces.php
- autoload_psr4.php

stof:
@garygreen those files are never loaded by default by composer (they are mostly generated for compatibility in case other tools use them), as the generated autoloader uses `autoload_static.php`. And that file is carefully designed to never require copying the array out of the opcache shared memory.
Note that this has nothing to do with the preloading feature of opcache. It is still all about the *caching* feature of opcache (i.e. the main opcache feature).

PR #12148, opened 2024-10-03
Fix --strict-ambiguous to really report all issues

### LINKED ISSUE #12140: `--strict-ambiguous` not working for some duplicates within project
I really like this [new option](https://github.com/composer/composer/pull/12119) in 2.8, but I think it should be much more strict. 

Documentation states:

```
--strict-ambiguous: Return a failed exit code (2) if the same class is found in multiple files.
```

But that is sadly **not true if the duplicate is present within my project** and not in vendor. Currently, I'm using [name-collission-detector](https://github.com/shipmonk-rnd/name-collision-detector) to find duplicates to ensure no ambiguity, but I think it would much be better to have such check within composer itself.

--------

Here is reproduction repository: https://github.com/janedbal/composer-class-ambiguity-test

Seldaek:
Here the second bit @janedbal - thanks for the good repro case that actually made me find a bug in non-strict mode even.

fixes #12140

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     "require": {
         "php": "^7.2.5 || ^8.0",
         "composer/ca-bundle": "^1.5",
-        "composer/class-map-generator": "^1.3.3",
+        "composer/class-map-generator": "^1.4.0",
         "composer/metadata-minifier": "^1.0",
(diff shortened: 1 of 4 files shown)


janedbal:
Thank you!

PR #12149, opened 2024-10-03
Clean up lock update code to remove redundancy and avoid errors

Seldaek:


src/Composer/Package/Locker.php
```diff
@@ -418,6 +415,60 @@ public function setLockData(array $packages, ?array $devPackages, array $platfor
...
+    /**
+     * Updates the lock file's hash in-place from a given composer.json's JsonFile
+     *
+     * This does not reload or require any packages, and retains the filemtime of the lock file.
+     *
+     * Use this only to update the lock file hash after updating a composer.json in ways that are guaranteed NOT to impact the dependency resolution.
+     *
+     * This is a risky method, use carefully.
+     *
+     * @param (callable(array<mixed>): array<mixed>)|null $dataProcessor Receives the lock data and can process it before it gets written to disk
```
stof, line 427: The type used for the lock data elsewhere in that class is `array<string, mixed>`

Seldaek: Yeah true, fixed in a67ce984e

PR #12159, opened 2024-10-15
Update authentication-for-private-packages.md

axi:
Fix Command line inline http-basic command.

Command was:
` composer config repositories composer.unique-name https://username:password@repo.example.org`

When I tried the command with Composer 2.7.6, I received the following error message:

> Setting repositories does not exist or is not supported by this command

image

The following command works
`composer config repositories.unique-name composer https://username:password@repo.example.org`


stof (approved):

Seldaek:
Thanks

PR #12152, opened 2024-10-08
do not stumble over missing descriptions

xabbuh:
fixes #12151

PR #12162, opened 2024-10-17
don't return a single space for packages without descriptions

xabbuh:
see https://github.com/composer/composer/pull/12152#discussion_r1800980110

### DIFF
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -649,7 +649,7 @@ class Problem
         $providers = $repositorySet->getProviders($packageName);
         if (\count($providers) > 0) {
             $providersStr = implode(array_map(static function ($p): string {
-                $description = $p['description'] !== '' ? ' '.substr($p['description'] ?? '', 0, 100) : '';
+                $description = $p['description'] !== '' && $p['description'] !== null ? ' '.substr($p['description'], 0, 100) : '';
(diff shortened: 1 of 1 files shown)


src/Composer/DependencyResolver/Problem.php
```diff
@@ -649,7 +649,7 @@ private static function getProvidersList(RepositorySet $repositorySet, string $p
         $providers = $repositorySet->getProviders($packageName);
         if (\count($providers) > 0) {
             $providersStr = implode(array_map(static function ($p): string {
-                $description = $p['description'] !== '' ? ' '.substr($p['description'] ?? '', 0, 100) : '';
+                $description = $p['description'] !== '' && $p['description'] !== null ? ' '.substr($p['description'] ?? '', 0, 100) : '';
```
fredden, line 652: Is this version more clear to the intent? We don't need the `??` operator inside the `substr()` call any more.
```suggestion
                $description = (($p['description'] ?? '') !== '') ? ' '.substr($p['description'], 0, 100) : '';
```

stof: I think the condition used by @xabbuh is clearer when reading the code.
But we can indeed remove the `?? ''` in the `substr` call.

fredden (approved):

Seldaek:
Thanks

PR #12180, opened 2024-10-28
On Windows, don't rely on the OS to find executables

nicolas-grekas:
Windows always looks up in the current directory when searching for a binary.
This is risky practice, so this PR tries to prevent this by using ExecutableFinder instead (which looks up using the PATH only).

While at it, I replaced all string command lines by array commands. This removed the need for explicit argument escaping.
Some remain, but most are removed.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "composer/ca-bundle",
-            "version": "1.5.2",
+            "version": "1.5.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/ca-bundle.git",
-                "reference": "48a792895a2b7a6ee65dd5442c299d7b835b6137"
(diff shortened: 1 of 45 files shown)


nicolas-grekas (commented):
Grepping for `ProcessExecutor::escape` is the best way to see what remains of string command-lines after this PR:
- `ZipDownloader`: could be turned into array commands
- `Util\Svn`: this one is rooted for string command-lines (`execute(string $command, ...)`)
- `EventDispatcher`: this class runs scripts only if allowed to so its safe
- `Perforce`: `executeCommand` is document as `@param non-empty-string $command`
- `SvnDownloader`: because it uses `Util\Svn`

These classes are nonetheless hardened thanks to the added logic in ProcessExecutor.

nicolas-grekas:
Symfony "generic" PR at https://github.com/symfony/symfony/pull/58710

Seldaek:
Thanks!

Issue #12175, opened 2024-10-27
allow path repository failing


shyim:
I have optional path repositories and when they don't exist I get:

```
0.078 In PathRepository.php line 163:
0.078                                                                                
0.078   The `url` supplied for the path (custom/plugins/*) repository does not exis  
0.078   t                                                                            
0.078       
```

What I would suggest having a `optional` flag, so when this is set the error is not thrown and it's fine.

```
{
            "type": "path",
            "url": "custom/plugins/*",
            "options": {
                "symlink": true,
                "optional": true
            }
        }
```

It's straightforward to integrate, looking for feedback before I make the PR :)

JoshuaBehrens:
Such a good thing for project templates. Always annoying to not have them at the start but end up setting these up anyways.

Seldaek:
Yeah that sounds reasonable to me if you like to send a PR.

PR #12186, opened 2024-11-06
feat: allow optional path repository

shyim:
allows a path repository to be optional non failing when the path is not there.

https://github.com/composer/composer/issues/12175

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -930,7 +930,8 @@
                 "options": {
                     "type": "object",
                     "properties": {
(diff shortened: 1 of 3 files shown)


stof:
how does this behave regarding the lock file ? What happens if someone that has this path performs an update when someone without the path runs an install (using the lock file) ?

Seldaek:
The repos are not looked at when installing from lock, so that would fail if the path contained in the lock file is not present on the filesystem.

Seldaek:
Closing because I don't think this can work.

PR #12188, opened 2024-11-07
Added backward compatibility for react/promise

glo71317:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "symfony/filesystem": "^5.4.35 || ^6.3.12 || ^7.0.3",
         "symfony/finder": "^5.4.35 || ^6.3.12 || ^7.0.3",
         "symfony/process": "^5.4.35 || ^6.3.12 || ^7.0.3",
-        "react/promise": "^3.2",
+        "react/promise": "^2.8 || ^3.2",
         "composer/pcre": "^2.2 || ^3.2",
         "symfony/polyfill-php73": "^1.24",
         "symfony/polyfill-php80": "^1.24",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "7c2f8d4d04b8146b2fcd203cb81fa8d5",
+    "content-hash": "b7cf35e9197a5667509471751a8990d9",
     "packages": [
         {
             "name": "composer/ca-bundle",


glo71317:
@naderman @Seldaek I would like to understand your point of view on this PR changes, As in current version 2.8.x we have remove `react/promise:^2.x` and now keeping only `react/promise:3.x` due to this several other packages are not backward compatible.
If we listing both supported versions as earlier `react/promise: ^2.8 || ^3` means that interoperability with other packages is maintained; Composer will pick the "best" version, which is typically the highest mutually-supported version. If we use ^3.0 only here, then we're forcing others to upgrade their packages in order for their coding standard to be used in conjunction with this product.

Can you share your point of view on this?

Seldaek:
I'm not sure about this one. react/promise 3.x is not a very difficult migration, and requiring composer/composer is generally frowned upon, so why should we support 2.x still? What's the exact use case you're solving?

glo71317:
> I'm not sure about this one. react/promise 3.x is not a very difficult migration, and requiring composer/composer is generally frowned upon, so why should we support 2.x still? What's the exact use case you're solving?

you may right react/promise 3.x is not a very difficult migration but we want to use composer:2.8 in adobe commerce but some other packages are not allowing to update the latest version of composer e.g. package `ezimuel/ringphp` requires react/promise:2.x and `opensearch-project/opensearch-php package depends on ezimuel/ringphp:1.2.2`.

As a result, these interdependent packages create a conflict, blocking the upgrade to the latest Composer version.

Seldaek:
And have you reached out to ezimuel to see if he can bump his promise requirement to allow v3? Because that'd be a better fix imo (moving forward)

glo71317:
> And have you reached out to ezimuel to see if he can bump his promise requirement to allow v3? Because that'd be a better fix imo (moving forward)

@Seldaek Yes, i tried to reach him but did not get any update https://github.com/ezimuel/ringphp/pull/12

glo71317:
Hi @Seldaek As this PR https://github.com/ezimuel/ringphp/pull/12 is closed from maintainer side.

Are we planning to address the backward compatibility issue regarding react/promise:2.x? It may be beneficial to consider removing support for react/promise:2.x in a future major version of Composer. This would ensure that Composer 2.8.x (and later versions) can be used without introducing breaking changes for users who rely on the current version of react/promise.

Looking forward to feedback and suggestions on this approach.

Seldaek:
Well if ringphp is deprecated and abandoned, I'd then go talk to opensearch-project/opensearch-php  about this because they should not use it anymore.

But yes I'll keep this open in the meantime and take a look whether I can reintroduce the promise v2 requirement or not, I don't quite recall why I removed it.

Seldaek:
Ok it seems to run fine with 2.x, so I guess we can do this change.

Seldaek:
I decided to drop react/promise 2.x again, as the situation upstream seems to have been resolved I hope you can now upgrade the needed dependencies.

Issue #12193, opened 2024-11-11
Support for safe.bareRepository explicit Git config is broken


Magentron:
My `composer.json`:

```json
{
    "name": "Test",
    "type": "project",
    "require": {
        "php": "^8.1",
        "magentron/openid-connect-php": "^0.9.11"
    },
	"repositories": [
		{
			"url": "https://github.com/Magentron/openid-connect-php.git",
			"type": "git",
			"no-api": true
		}
	],
    "config": {
        "optimize-autoloader": true,
        "preferred-install": "dist",
        "sort-packages": true
    },
    "minimum-stability": "dev",
    "prefer-stable": true
}
```

Output of `composer diagnose`:

```
Checking platform settings: OK
Checking git settings: OK git version 2.43.0
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK does not expire
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking Composer version: OK
Checking Composer and its dependencies for vulnerabilities: FAIL
Audit found some issues:
Found 1 security vulnerability advisory affecting 1 package:
+-------------------+----------------------------------------------------------------------------------+
| Package           | symfony/process                                                                  |
| Severity          | high                                                                             |
| CVE               | CVE-2024-51736                                                                   |
| Title             | CVE-2024-51736: Command execution hijack on Windows with Process class           |
| URL               | https://symfony.com/cve-2024-51736                                               |
| Affected versions | >=2.0.0,<3.0.0|>=3.0.0,<4.0.0|>=4.0.0,<5.0.0|>=5.0.0,<5.1.0|>=5.1.0,<5.2.0|>=5.2 |
|                   | .0,<5.3.0|>=5.3.0,<5.4.0|>=5.4.0,<5.4.46|>=6.0.0,<6.1.0|>=6.1.0,<6.2.0|>=6.2.0,< |
|                   | 6.3.0|>=6.3.0,<6.4.0|>=6.4.0,<6.4.14|>=7.0.0,<7.1.0|>=7.1.0,<7.1.7               |
[...]

PR #12194, opened 2024-11-11
fix: refactor bare repository explicit logic

Magentron:
- Do not check git command
- Use current working directory's .git for composer compile

Fixes: #12193

### DIFF
--- a/src/Composer/Util/ProcessExecutor.php
+++ b/src/Composer/Util/ProcessExecutor.php
@@ -42,13 +42,6 @@ class ProcessExecutor
         'setlocal', 'shift', 'start', 'time', 'title', 'type', 'ver', 'vol',
     ];
 
-    private const GIT_CMDS_NEED_GIT_DIR = [
-        ['show'],
-        ['log'],
-        ['branch'],
-        ['remote', 'set-url']
-    ];
-
     /** @var int */
     protected static $timeout = 300;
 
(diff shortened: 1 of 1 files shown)


src/Composer/Util/ProcessExecutor.php
```diff
@@ -556,22 +541,43 @@ private static function escapeArgument($argument): string
...
+            return null;
+        }
+
+        // if composer is being compiled from source, we need to use the git dir from the source checkout
+        $srcComposerDirEnd = DIRECTORY_SEPARATOR . 'src' . DIRECTORY_SEPARATOR . 'Composer';
+        $isComposerSource = $cwd !== null && str_ends_with($cwd, $srcComposerDirEnd);
+
+        if ($isComposerSource) {
+            return getcwd() . '/.git';
+        }
```
Seldaek, line 562: I have no idea what this does or what the comment means sorry could you please explain?

Seldaek: Ok that I fixed in 580f0006d - that's better than having random hacks in this class IMO. I'm not sure what the rest of this PR is doing exactly or why we need it to fix #12193.

Magentron: Yes, you are right, that is a better solution. That also fixes my issue #12193 👍

PR #12203, opened 2024-11-17
Use a bitmask to produce deterministic exit codes for the "audit" command

### LINKED ISSUE #12201: `audit` command should produce deterministic exit codes
## Problem

The exit code from the `audit` command should be deterministic. This way, the exit code can be used to quickly expose the origin of the reported problems (CVEs, abandoned packages, etc) without requiring to read the output for a quick insight.
Additionally, it allows CI environments to handle its job statuses based on the command exit code.

Currently, this is not possible because the command uses the number of packages offending the defined checks to produce the exit code:

https://github.com/composer/composer/blob/9fb833f97e59b6af886fd5d5c3f0a81012a61bf5/src/Composer/Advisory/Auditor.php#L133
https://github.com/composer/composer/blob/9fb833f97e59b6af886fd5d5c3f0a81012a61bf5/src/Composer/Command/AuditCommand.php#L79-L88

In other words, these scenarios produce the same exit code:

1.
    ```
    1 abandoned packages;
    1 package affected by a CVE.
    ```

2.
    ```
    0 abandoned packages;
    2 packages affected by a CVE.
    ```

3.
    ```
    2 abandoned packages;
    0 packages affected by a CVE.
    ```
## Proposal

In order to allow the purposed behavior while keeping the current return type for `Auditor::audit()`, I guess we could use a bitmask there. Something like this:

```php
$bitHasVulnerablePackages = 1;
$bitHasAbandonedPackages = 2;

$result = 0;

if (0 < $affectedPackagesCount) {
    $result |= $bitHasVulnerablePackages;
}

if (0 < $abandonedCount) {
    $result |= $bitHasAbandonedPackages;
}

// Check the results:
$hasVulnerablePackages = 0 < ($result & $bitHasVulnerablePackages);
$hasAbandonedPackages = 0 < ($result & $bitHasAbandonedPackages);
```

Then, if no additional treatment is done at `AuditCommand`, the exit codes would be mapped as:

| Exit code | Has vulnerable packages? | Has abandoned packages?
|---|---|---
| 0 | No | No
| 1 | Yes | No
| 2 | No | Yes
| 3 | Yes | Yes

phansys:
Fixes #12201.

### DIFF
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -54,9 +54,9 @@ class Auditor
     ];
 
     /** Values to determine the audit result. */
-    public const BIT_OK = 0;
-    public const BIT_VULNERABLE = 1;
-    public const BIT_ABANDONED = 2;
+    public const STATUS_OK = 0;
+    public const STATUS_VULNERABLE = 1;
+    public const STATUS_ABANDONED = 2;
 
     /**
      * @param PackageInterface[] $packages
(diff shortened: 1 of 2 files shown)


phansys:
Thank you!

Regarding the `STATUS_` prefix in the constants, please, be aware that I decided not to use the "status" term in order to avoid confusions, as these values are not always what the user receives.
My reasoning was that they are the basis of the final result, but not an exact representation of it.

Seldaek:
Yeah fair enough but BIT_ sounds like leaking an implementation detail in the name to me.. I guess both have some flaws :D But I don't really expect many will rely on these constants as this is usually called via CLI, so no biggie either way.

PR #12206, opened 2024-11-20
Add build-path to php-ext config options for PIE

asgrim:
Adds `build-path` option for :pie: PIE - refs ThePHPF/pie-design#28

Also made the definition for the `php-ext` shape re-usable to avoid copy/pasta errors in future

### DIFF
--- a/src/Composer/Package/Package.php
+++ b/src/Composer/Package/Package.php
@@ -99,7 +99,10 @@ class Package extends BasePackage
     protected $isDefaultBranch = false;
     /** @var mixed[] */
     protected $transportOptions = [];
-    /** @var PhpExtConfig */
+    /**
+     * @var array|null
+     * @phpstan-var PhpExtConfig|null
+     */
     protected $phpExt = null;
 
     /**
@@ -596,7 +599,9 @@ class Package extends BasePackage
     /**
      * Sets the settings for php extension packages
      *
-     * @param PhpExtConfig $phpExt
+     * @param array|null $phpExt
+     *
+     * @phpstan-param PhpExtConfig|null $phpExt
      */
     public function setPhpExt(?array $phpExt): void
     {
--- a/src/Composer/Package/Package.php
+++ b/src/Composer/Package/Package.php
@@ -23,6 +23,7 @@ use Composer\Util\ComposerMirror;
  *
  * @phpstan-import-type AutoloadRules from PackageInterface
  * @phpstan-import-type DevAutoloadRules from PackageInterface
+ * @phpstan-import-type PhpExtConfig from PackageInterface
  */
 class Package extends BasePackage
 {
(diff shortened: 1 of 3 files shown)


src/Composer/Package/Package.php
```diff
@@ -593,9 +594,9 @@ public function getIncludePaths(): array
     }
 
     /**
-     * Sets the list of paths added to PHP's include path.
+     * Sets the settings for php extension packages
      *
-     * @param array{extension-name?: string, priority?: int, support-zts?: bool, support-nts?: bool, configure-options?: list<array{name: string, description?: string}>}|null $phpExt List of directories.
+     * @param PhpExtConfig $phpExt
```
stof, line 599: This should use `@phpstan-param` instead of `@param` when using a phpstan type alias, to avoid issues with tools not supporting the phpstan type alias which would read `@param PhpExtConfig $phpExt` as a PhpExtConfig class

src/Composer/Package/PackageInterface.php
```diff
@@ -326,7 +327,7 @@ public function getIncludePaths(): array;
     /**
      * Returns the settings for php extension packages
      *
-     * @return array{extension-name?: string, priority?: int, support-zts?: bool, support-nts?: bool, configure-options?: list<array{name: string, description?: string}>}|null
+     * @return PhpExtConfig
```
stof, line 330: should be `@phpstan-return`

src/Composer/Package/PackageInterface.php
```diff
@@ -23,6 +23,7 @@
  *
  * @phpstan-type AutoloadRules    array{psr-0?: array<string, string|string[]>, psr-4?: array<string, string|string[]>, classmap?: list<string>, files?: list<string>, exclude-from-classmap?: list<string>}
  * @phpstan-type DevAutoloadRules array{psr-0?: array<string, string|string[]>, psr-4?: array<string, string|string[]>, classmap?: list<string>, files?: list<string>}
+ * @phpstan-type PhpExtConfig     array{extension-name?: string, priority?: int, support-zts?: bool, support-nts?: bool, build-path?: string|null, configure-options?: list<array{name: string, description?: string}>}|null
```
stof, line 26: Personally, I would rather define `PhpExtConfig` as being the array shape and then using `PhpExtConfig|null` for places where it is nullable. But the final decision is for @Seldaek and @naderman

src/Composer/Package/Package.php
asgrim: Thanks, yeah I forget there might be "other tools" :+1:

src/Composer/Package/PackageInterface.php
asgrim: Seems reasonable - updated, thank you

Seldaek:
Thanks

PR #12207, opened 2024-11-22
disable multiplexing for some versions of curl

Lctrs:
I'm behind a corporate proxy and was hitting a `Curl 2 (...) [CONN-1-0] send: no filter connected` error when trying to download some packages.

Some google research led me to https://github.com/rust-lang/cargo/issues/12202 and its fix https://github.com/rust-lang/cargo/pull/12234.

This PR backports this fix to composer.

> In certain versions of libcurl when proxy is in use with HTTP/2
multiplexing, connections will continue stacking up. This was fixed in libcurl 8.0.0 in curl/curl@821f6e2

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -34,6 +34,14 @@ use Symfony\Component\HttpFoundation\IpUtils;
  */
 class CurlDownloader
 {
+    /**
+     * Known libcurl's broken versions when proxy is in use with HTTP/2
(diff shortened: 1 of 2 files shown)


Lctrs:
A more readable version of the backported code can be found here : https://github.com/rust-lang/cargo/blob/88edf01fe39d6d15c8e15dbf4ea478661b7b85e3/src/cargo/util/context/mod.rs#L2983-L3009.

Seldaek:
Thanks!

PR #12211, opened 2024-11-26
Update wording of process-timeout description

beryllium:
Cleans up the description of process-timeout to better separate the config setting from the static helper for script commands.


Seldaek:
Thanks

Issue #12214, opened 2024-11-26
Error reporting and displaying in bin/composer is too aggressive, breaks 2.2 LTS with PHP 8.4


dzuelke:
Running, for example, the LTS Composer 2.2 using PHP 8.4 breaks all sort of things due to some errors being duplicated to stdout when doing, say, a simple `composer config vendor-dir`:

```

Deprecated: Composer\Util\ErrorHandler::register(): Implicitly marking parameter $io as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/local/bin/composer/src/Composer/Util/ErrorHandler.php on line 89
vendor
```

This is in addition to many errors on stderr:

```
PHP Deprecated:  Composer\Util\ErrorHandler::register(): Implicitly marking parameter $io as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/local/bin/composer/src/Composer/Util/ErrorHandler.php on line 89
Deprecation Notice: Composer\Command\InitCommand::determineRequirements(): Implicitly marking parameter $platformRepo as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/local/bin/composer/src/Composer/Command/InitCommand.php:525
Deprecation Notice: Composer\Command\InitCommand::findBestVersionAndNameForPackage(): Implicitly marking parameter $platformRepo as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/local/bin/composer/src/Composer/Command/InitCommand.php:898
Deprecation Notice: Composer\Command\InitCommand::getPlatformExceptionDetails(): Implicitly marking parameter $platformRepo as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/local/bin/composer/src/Composer/Command/InitCommand.php:1022
Deprecation Notice: Composer\Command\CreateProjectCommand::installProject(): Implicitly marking parameter $platformRequirementFilter as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/local/bin/composer/src/Composer/Command/CreateProjectCommand.php:192
Deprecation Notice: Composer\Command\CreateProjectCommand::installRootPackage(): Implicitly marking parameter $repositories as nullable is deprecated, the explicit nullable type must be used instead in phar:///usr/local/bin/composer/src/Composer/Command/CreateProjectCommand.php:361
[...]

PR #12217, opened 2024-11-28
Suppress E_DEPRECATED for PHP 8.4 compatibility

dzuelke:
The `E_STRICT` constant is deprecated in PHP 8.4, and explicit nullable types are also advised with a deprecation notice.

As there are circumstances where people still have to run the 2.2 LTS with newer versions of PHP, we're now suppressing `E_DEPRECATED` (especially because some of the deprecation warnings may end up on stdout due to `display_errors` getting set to `1` in `bin/composer`).

Also see #12214

Targeting 2.2 as discussed in https://github.com/composer/composer/issues/12214#issuecomment-2504292057

### DIFF
--- a/bin/composer
+++ b/bin/composer
@@ -18,7 +18,9 @@ use Composer\XdebugHandler\XdebugHandler;
 use Composer\Util\Platform;
 use Composer\Util\ErrorHandler;
 
-error_reporting(-1);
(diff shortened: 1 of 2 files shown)


dzuelke:
(PHPStan [always fails for the 2.2 branch](https://github.com/composer/composer/actions/workflows/phpstan.yml?query=branch%3A2.2), so this is ready to merge I think, @Seldaek)

Seldaek:
Thanks

Seldaek:
2.2.25 is out with the fix

PR #12226, opened 2024-12-07
Fix bug when plugin defines multiple PluginInterface classes

justinbeaty:
I have a composer plugin that defines multiple `PluginInterface` classes in `composer.json` as mentioned in the [docs](https://getcomposer.org/doc/articles/plugins.md#plugin-package).

However, there is a bug when upgrading between versions of the plugin when there are multiple classes defined.

For example, assume this is the `composer.json` in version 1.0.0.

```json
{
    "name": "my/plugin",
    "version": "1.0.0",
    "autoload": {
        "psr-4": {
            "MyNamespace\\": "src/"
        }
    },
    "extra": {
        "class": [
            "MyNamespace\\FirstPlugin",
            "MyNamespace\\SecondPlugin",
            "MyNamespace\\ThirdPlugin"
        ]
    }
}
```

And then in 2.0.0, we consolidate the functionality into one single class:

```json
{
    "name": "my/plugin",
    "version": "2.0.0",
    "autoload": {
        "psr-4": {
            "MyNamespace\\": "src/"
        }
    },
    "extra": {
        "class": [
            "MyNamespace\\ConsolidatedPlugin"
        ]
    }
}
```

When we run `composer update`, the 1.0.0 package is activated before it's removed, but only a reference to `ThirdPlugin` will be kept in `$this->registeredPlugins[$package->getName()]`, and so only `ThirdPlugin` is deactivated.

Thus `FirstPlugin` and `SecondPlugin` are still called, which will cause an error if those classes aren't self-contained and depend on functionality removed in 2.0.0.

Hopefully the explanation is clear, but I can create a reproduction demo tomorrow if you'd like. I also haven't run unit tests or phpstan yet, but will wait for the workflows.

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -51,9 +51,9 @@ class PluginManager
     protected $disablePlugins = false;
 
     /** @var array<PluginInterface> */
-    protected $plugins = array();
-    /** @var array<string, PluginInterface|InstallerInterface> */
-    protected $registeredPlugins = array();
+    protected $plugins = [];
+    /** @var array<string, array<PluginInterface|InstallerInterface>> */
+    protected $registeredPlugins = [];
 
     /**
      * @var array<non-empty-string, bool>|null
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, I guess nobody else ever used the ability to define multiple classes? :)

Seldaek:
See also https://github.com/composer/composer/commit/5cb97335880c9e604e3a470960159174c4585d6c for main branch

PR #12231, opened 2024-12-10
Adding explicit message to `why-not` if package is already installed

### LINKED ISSUE #12227: Suggestion: `composer why-not`: Better output if package is already installed
If I'm doing `composer why-not` on a package version that I already have, the output is misleading:

> $ composer why-not symfony/asset 7.2.0
There is no installed package depending on "symfony/asset" in versions not matching 7.2.0
Not finding what you were looking for? Try calling `composer require "symfony/asset:7.2.0" --dry-run` to get another view on the problem.

I'm suggesting to replace this with something like:
> symfony/asset 7.2.0 is already installed! To find out why, run `composer why symfony/asset`.

ThomasLandauer:
Closes #12227

I just did what you told me in https://github.com/composer/composer/issues/12227#issuecomment-2531797705 - don't really have an overview ;-)

`return` is still missing, right? Should I just copy this from further down?:
```php
$return = $inverted ? 1 : 0;
```
Or move it upwards?

### DIFF
--- a/src/Composer/Command/BaseDependencyCommand.php
+++ b/src/Composer/Command/BaseDependencyCommand.php
@@ -126,6 +126,8 @@ abstract class BaseDependencyCommand extends BaseCommand
                 $extraNotice = ' (version provided by config.platform)';
             }
             $this->getIO()->writeError('<info>Package "'.$needle.' '.$textConstraint.'" found in version "'.$matchedPackage->getPrettyVersion().'"'.$extraNotice.'.</info>');
+        } elseif ($inverted) {
+            $this->getIO()->write('<comment>Package "'.$needle.'" '.$matchedPackage->getPrettyVersion().' is already installed! To find out why, run `composer why '.$needle.'`</comment>');
         }
 
         // Include replaced packages for inverted lookups as they are then the actual starting point to consider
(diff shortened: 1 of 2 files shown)


Seldaek:
Well I think I'd say `return 0;` always because it's kinda successful as in there is no reason why-not that package.

Seldaek:
As for this:

> I just did what you told me in [#12227 (comment)](https://github.com/composer/composer/issues/12227#issuecomment-2531797705) - don't really have an overview ;-)

Ok but did you try to run the code tho? Does it work and make sense to you? :) Also would need to try the why command to make sure it doesn't bork.

Seldaek:
We do but it's in the .github dir https://github.com/composer/composer/blob/main/.github/CONTRIBUTING.md

ThomasLandauer:
Wouldn't it be better to move `CONTRIBUTING.md` to the top directory?

The code looks good -here's what I did:
```bash
$ git clone -b patch-1 https://github.com/ThomasLandauer/composer
$ cd composer && php ../composer.phar install
$ bin/composer why-not composer/ca-bundle 1.5.3
# Font in orange color:
Package "composer/ca-bundle" 1.5.3 is already installed! To find out why, run `composer why composer/ca-bundle`
$ bin/composer why composer/ca-bundle
composer/composer 2.8.x-dev requires composer/ca-bundle (^1.5)
```

Seldaek:
Alright thanks

PR #12244, opened 2024-12-17
Support schemeless funding URLs

andrewnicols:
We encountered this issue when parsing the FUNDING.yml on our repo - https://github.com/moodle/moodle.

The [GitHub documentation][1] for FUNDING.yml specifically notes that if a custom URL is in the Array format, and includes `:` then it must be quoted.

From this we can infer that the custom URL does not have to contain a `:` at all.

The example for a Custom URL also gives an example of an unquoted URL without any `:` character:

```
custom: ["https://www.paypal.me/octocat", octocat.com]
```

However if a repository uses a URL in this format it is currently rejected because it does not specify a scheme. Furthermore the `parse_url` method treats the `octocat.com` example as a path and not a host. The same is true for URLs such as
`octocat.com/funding/example.html`.

This patch adds an additional allowance for the URL filter to capture the case where a URL has no scheme or host, but does have a path.

[1]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository#about-funding-files

Note: I've also got a branch for the `main` branch at https://github.com/andrewnicols/composer/tree/schemelessUrlMain

### DIFF
--- a/src/Composer/Package/Loader/ValidatingArrayLoader.php
+++ b/src/Composer/Package/Loader/ValidatingArrayLoader.php
@@ -646,6 +646,14 @@ class ValidatingArrayLoader implements LoaderInterface
         }
 
         $bits = parse_url($value);
+        if ($bits === false) {
+            return false;
+        }
+
+        if (array_key_exists('path', $bits) && !array_key_exists('scheme', $bits) && !array_key_exists('host', $bits)) {
+            return true;
+        }
+
         if (empty($bits['scheme']) || empty($bits['host'])) {
             return false;
         }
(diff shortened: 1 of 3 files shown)


medabkari:
@andrewnicols Duplicate of #12245

andrewnicols:
@medabkari:  As noted in both PRs 12244 is against 2.2; 12245 is against main branch.

Seldaek:
Closing per https://github.com/composer/composer/pull/12245#issuecomment-2580366226

PR #12245, opened 2024-12-17
Support schemeless funding URLs

andrewnicols:
Note: This is a copy of #12244 against the `main` branch.

The [GitHub documentation][1] for FUNDING.yml specifically notes that if a custom URL is in the Array format, and includes `:` then it must be quoted.

From this we can infer that the custom URL does not have to contain a `:` at all.

The example for a Custom URL also gives an example of an unquoted URL without any `:` character:

```
custom: ["https://www.paypal.me/octocat", octocat.com]
```

However if a repository uses a URL in this format it is currently rejected because it does not specify a scheme. Furthermore the `parse_url` method treats the `octocat.com` example as a path and not a host. The same is true for URLs such as
`octocat.com/funding/example.html`.

This patch adds an additional allowance for the URL filter to capture the case where a URL has no scheme or host, but does have a path.

[1]: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository#about-funding-files

### DIFF
--- a/src/Composer/Package/Loader/ValidatingArrayLoader.php
+++ b/src/Composer/Package/Loader/ValidatingArrayLoader.php
@@ -623,6 +623,14 @@ class ValidatingArrayLoader implements LoaderInterface
         }
 
         $bits = parse_url($value);
+        if ($bits === false) {
+            return false;
+        }
+
+        if (array_key_exists('path', $bits) && !array_key_exists('scheme', $bits) && !array_key_exists('host', $bits)) {
+            return true;
+        }
+
         if (empty($bits['scheme']) || empty($bits['host'])) {
             return false;
         }
--- a/tests/Composer/Test/Package/Loader/ValidatingArrayLoaderTest.php
+++ b/tests/Composer/Test/Package/Loader/ValidatingArrayLoaderTest.php
@@ -84,6 +84,14 @@ class ValidatingArrayLoaderTest extends TestCase
                         [
                             'url' => 'https://example.org/fund',
                         ],
+                        [
+                            'type' => 'custom',
+                            'url' => 'example.com',
+                        ],
+                        [
+                            'type' => 'custom',
+                            'url' => 'example.com/fund#fundus',
+                        ],
                     ],
                     'require' => [
                         'a/b' => '1.*',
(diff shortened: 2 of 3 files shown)


andrewnicols:
Thanks @stof, That makes sense. I suspect there could be parts of the Packagist site which are impacted by this then.

I've also raised https://github.com/github/docs/issues/35690 for GitHub to improve the documentation on valid formats.

andrewnicols:
Okay, so looking at this another way, given the custom URL seems to be the only URL impacted by this, and the format is not accepted by composer.json, I wonder if the correct fix is in src/Composer/Repository/Vcs/GitHubDriver.php in the `getFundingInfo()` method and we should do something like:

```php
                case 'custom':
                    $bits = parse_url($item['url']);
                    if ($bits === false) {
                        unset($result[$key]);
                        break;
                    }

                    if (array_key_exists('path', $bits) && !array_key_exists('scheme', $bits) && !array_key_exists('host', $bits))   {
                        $result[$key]['url'] = 'https://' . $item['url'];
                    }
                    break;
```

I haven't seen where the information from the vcs driver is passed to the ValidatingArrayLoader yet to know whether this would fix the issue though.

stof:
this would assume that omitting `https://` is the only "invalid" thing being done. While this is true in your case, it might not be true in general.

andrewnicols:
I agree, though I'm not sure what else to do here until GitHub provide a bit more information on the valid formats.

stof:
What you can do in the meantime is to fix this in the moodle repo, by using an absolute URL. There is *no way* in which `moodle.org` will be parsed into a URL representing a link to `https://moodle.org`. Either absolute URLs are mandatory (in which case it is simply invalid config in your repo) or relative URLs are supported and it needs to be resolved against some base URL (but which one?), which has no way to produce `https://moodle.org` as output.

Seldaek:
I think I'd rather close this and just merge #12247 - or you do normalize these URLs in GitHubDriver if you want that'd also be acceptable to me. But adding support for this in general for all URLs in the validating array loader sounds like it'll bite us in the ass somewhere.

andrewnicols:
Does it still break the build on “invalid” value? I think I raised a
separate issue for that

Seldaek:
No I think it should simply ignore them

PR #12247, opened 2024-12-18
Discard unsupported FUNDING.yml URL values

andrewnicols:
This patch is intended as an interim solution for #12245 but some form of this should probably be present anyway even after that issue is sorted.

At the moment if a GitHub repo uses a FUNDING.yml in the format documented by GitHub then that entire branch will be skipped from import into Packagist beacuse some of the documented formats are not supported by Packagist/Composer.

As a worst case I would expect the unsupported data to be removed. This isn't critical data - it's optional metadata which is not even present in the composer.json file and which is correct per the GitHub documentation.

On our side I'm trying to get the GitHub FUNDING.yml updated but many of the branches this issue exists on are archived branches which we do not normally modify.

At the moment the Packagist management view shows the following warning:

> This package is in a broken state and will not update anymore. Some branches contain invalid data and until you fix them the entire package is frozen. Click "Update" below to see details.

This suggests that if _any_ of our branches contain 'invalid' data that the entire package will be rejected.

### DIFF
--- a/src/Composer/Repository/Vcs/GitHubDriver.php
+++ b/src/Composer/Repository/Vcs/GitHubDriver.php
@@ -287,6 +287,19 @@ class GitHubDriver extends VcsDriver
                 case 'buy_me_a_coffee':
                     $result[$key]['url'] = 'https://www.buymeacoffee.com/' . basename($item['url']);
                     break;
+                case 'custom':
+                    $bits = parse_url($item['url']);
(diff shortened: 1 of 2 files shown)


Seldaek:
Ok I added support for very simple scheme-less URLs as that seems like it cannot backfire there. Thanks for the the fix and tests! 👍🏻

PR #12248, opened 2024-12-19
Allow redirect responses to output warnings/infos

glaubinix:
Redirect responses can have a content too. This PR allows redirect responses to return warning(s)/info(s) which then appears as output during `composer install`


Seldaek:
Thanks

PR #12250, opened 2024-12-20
Update installer script URL to include openssl_free_key() deprecation fix

elazar:
If the installer script linked from [this page](https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md) is run using PHP 8, it generates the following deprecation notice.

```
Deprecated: Function openssl_free_key() is deprecated since 8.0, as OpenSSLAsymmetricKey objects are freed automatically in Standard input code on line 982
```

This issue was [addressed in the installer script](https://github.com/composer/getcomposer.org/pull/159), but the documentation was not updated to point to a version of it that includes the fix.


Seldaek:
Thanks

PR #12257, opened 2024-12-31
Fix failing Packagist updates / GitHubDriver::getFundingInfo(): add support for thanks.dev and polar.sh

jrfnl:
### GitHubDriver::getFundingInfo(): order the cases

This re-orders the cases in the `switch` to follow the same order as the GitHub documentation (largely alphabetic) for easier comparisons between the two lists.

Refs:
* https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository

### GitHubDriver::getFundingInfo(): add support for thanks.dev and polar.sh

GitHub looks to have added a dedicated syntax for the thanks.dev funding platform when added to a `funding.yml` file.
However, it looks like Composer does not (yet) support this syntax as can be seen from failed Packagist updates of the dev branches of the [PHP_CodeSniffer](https://packagist.org/packages/squizlabs/php_codesniffer#dev-master) and [PHPCompatibility](https://packagist.org/packages/phpcompatibility/php-compatibility) packages.

image

The polar.sh funding platform also appears to be newly supported by GH and missing from the list.

This PR fixes both.

Refs:
* https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository

---

Note: while I'd love to add some tests for this, it looks like there are no tests covering this functionality at all yet and it would require some hacks to mock the various API calls to reach the code, while this is a relatively straight forward code change.

As it looks like (open) PR #12247 already adds a test method for this code, I'd like to suggest either of the following:
* Merge #12247 first (if ready) and then I could add a few additional test cases to the data provider.
* Or alternatively, I could cherrypick the test @andrewnicols created from their commit and add a data provider to safeguard the currently supported platforms.

Let me know what you'd prefer.

Also feel free to let me know if this should be pulled to an older branch. Happy to rebase.

### DIFF
--- a/src/Composer/Repository/Vcs/GitHubDriver.php
+++ b/src/Composer/Repository/Vcs/GitHubDriver.php
@@ -257,35 +257,35 @@ class GitHubDriver extends VcsDriver
 
         foreach ($result as $key => $item) {
             switch ($item['type']) {
-                case 'tidelift':
-                    $result[$key]['url'] = 'https://tidelift.com/funding/github/' . $item['url'];
+                case 'community_bridge':
+                    $result[$key]['url'] = 'https://funding.communitybridge.org/projects/' . basename($item['url']);
                     break;
                 case 'github':
                     $result[$key]['url'] = 'https://github.com/' . basename($item['url']);
                     break;
-                case 'patreon':
-                    $result[$key]['url'] = 'https://www.patreon.com/' . basename($item['url']);
-                    break;
-                case 'otechie':
-                    $result[$key]['url'] = 'https://otechie.com/' . basename($item['url']);
-                    break;
-                case 'open_collective':
-                    $result[$key]['url'] = 'https://opencollective.com/' . basename($item['url']);
-                    break;
-                case 'liberapay':
-                    $result[$key]['url'] = 'https://liberapay.com/' . basename($item['url']);
+                case 'issuehunt':
+                    $result[$key]['url'] = 'https://issuehunt.io/r/' . $item['url'];
                     break;
                 case 'ko_fi':
                     $result[$key]['url'] = 'https://ko-fi.com/' . basename($item['url']);
                     break;
-                case 'issuehunt':
-                    $result[$key]['url'] = 'https://issuehunt.io/r/' . $item['url'];
+                case 'liberapay':
+                    $result[$key]['url'] = 'https://liberapay.com/' . basename($item['url']);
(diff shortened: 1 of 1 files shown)


jrfnl:
Hmm... turns out this issue is causing bigger problems than I expected, as dependents using the `dev-master` branch of projects using `thanks_dev` in their `funding.yml` can no longer update and get shown a `"Error: Your requirements could not be resolved to an installable set of packages."` message.

This message is very confusing for end-users as it doesn't point out the _real_ problem (an issue on Packagist with updating the `dev-master` branch of the dependency).

The message is also problematic as the issue is not unresolvable by end-users, but they have no way of knowing that and may have spend hours already trying to debug this type of installation problem.

I think it would be a good idea to rethink the messaging to the end-user for this situation too....

Case in point: https://github.com/sirbrillig/phpcs-variable-analysis/issues/348

Seldaek:
I think the confusing errors will be resolved by #12247 - thanks for the fixes here!

Seldaek:
> As it looks like (open) PR #12247 already adds a test method for this code, I'd like to suggest either of the following:
> 
> * Merge [Discard unsupported FUNDING.yml URL values #12247](https://github.com/composer/composer/pull/12247) first (if ready) and then I could add a few additional test cases to the data provider.
> * Or alternatively, I could cherrypick the test @andrewnicols created from their commit and add a data provider to safeguard the currently supported platforms.

Sorry just reading this now 🤦🏻‍♂️ anyway these cases are super simple so I'm ok without much testing but the other PR is now merged so if you wanna add more cases feel free to PR 🙏🏻

jrfnl:
@Seldaek Happy to help. I do have question though - as this is mostly a problem on the Packagist side of things: should I wait for a new release of Composer before re-applying the change in the `funding.yml` file(s) or can I safely do this already now this has been merged ?
I hope the latter, as otherwise it would still be problematic for users using older Composer versions, which would mean updating the `funding.yml` file would still be blocked for quite a while, so thought I'd better check.

Seldaek:
Nope, I just deployed to packagist with the latest composer commit so it should work now.

jrfnl:
@Seldaek Thanks for confirming! I've opened PR #12271 now to add some tests (and found a bug in this PR via the tests :grinning:).

PR #12090, opened 2024-08-22
Improve reproducibility of autoloader files - add missing `sort`

drupol:
While updating PHPDocumentor in Nix ([#336145](https://github.com/NixOS/nixpkgs/pull/336145)), we encountered instability with the new version of the Composer builder introduced in [#308059](https://github.com/NixOS/nixpkgs/pull/308059). In the context of Nix, "unstable" means that the `vendor` directory produced by Composer was not consistently reproducible.

image

To reproduce the issue locally:

```
$ git clone https://github.com/phpdocumentor/phpdocumentor
$ cd phpdocumentor
$ git checkout v3.6.0
$ rm -rf vendor0 vendor1
$ composer install
$ mv vendor vendor0
$ rm -rf ~/.cache/composer
$ composer install
$ mv vendor vendor1
$ diffoscope --exclude-directory-metadata recursive vendor0 vendor1
--- vendor0
+++ vendor1
│   --- vendor0/composer
├── +++ vendor1/composer
│ │   --- vendor0/composer/autoload_psr4.php
│ ├── +++ vendor1/composer/autoload_psr4.php
│ │ @@ -4,15 +4,15 @@
│ │  
│ │  $vendorDir = dirname(__DIR__);
│ │  $baseDir = dirname($vendorDir);
│ │  
│ │  return array(
│ │      'phpDocumentor\\Reflection\\' => array($vendorDir . '/phpdocumentor/reflection-common/src', $vendorDir . '/phpdocumentor/type-resolver/src', $vendorDir . '/phpdocumentor/reflection-docblock/src'),
│ │      'phpDocumentor\\JsonPath\\' => array($baseDir . '/incubator/json-path/tests/unit', $vendorDir . '/phpdocumentor/json-path/src'),
│ │ +    'phpDocumentor\\Guides\\' => array($vendorDir . '/phpdocumentor/guides/src', $vendorDir . '/phpdocumentor/guides-graphs/src', $vendorDir . '/phpdocumentor/guides-restructured-text/src', $vendorDir . '/phpdocumentor/guides-markdown/src'),
│ │ -    'phpDocumentor\\Guides\\' => array($vendorDir . '/phpdocumentor/guides-graphs/src', $vendorDir . '/phpdocumentor/guides/src', $vendorDir . '/phpdocumentor/guides-restructured-text/src', $vendorDir . '/phpdocumentor/guides-markdown/src'),
[...]

PR #12263, opened 2025-01-09
Fix unstable order of psr-0 and psr-4 rules

Seldaek:
Fixes #12090

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -578,12 +578,17 @@ EOF;
         }
         $sortedPackageMap = $this->sortPackageMap($packageMap);
         $sortedPackageMap[] = $rootPackageMap;
-        array_unshift($packageMap, $rootPackageMap);
+        $reverseSortedMap = array_reverse($sortedPackageMap);
 
-        $psr0 = $this->parseAutoloadsType($packageMap, 'psr-0', $rootPackage);
-        $psr4 = $this->parseAutoloadsType($packageMap, 'psr-4', $rootPackage);
-        $classmap = $this->parseAutoloadsType(array_reverse($sortedPackageMap), 'classmap', $rootPackage);
+        // reverse-sorted means root first, then dependents, then their dependents, etc.
+        // which makes sense to allow root to override classmap or psr-0/4 entries with higher precedence rules
+        $psr0 = $this->parseAutoloadsType($reverseSortedMap, 'psr-0', $rootPackage);
+        $psr4 = $this->parseAutoloadsType($reverseSortedMap, 'psr-4', $rootPackage);
+        $classmap = $this->parseAutoloadsType($reverseSortedMap, 'classmap', $rootPackage);
+
+        // sorted (i.e. dependents first) for files to ensure that dependencies are loaded/available once a file is included
         $files = $this->parseAutoloadsType($sortedPackageMap, 'files', $rootPackage);
+        // using sorted here but it does not really matter as all are excluded equally
         $exclude = $this->parseAutoloadsType($sortedPackageMap, 'exclude-from-classmap', $rootPackage);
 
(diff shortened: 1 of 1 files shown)


drupol:
Looking forward for this, thank you <3 !

drupol:
Added as a patch in Nix until 2.8.5 is released https://github.com/NixOS/nixpkgs/pull/372696/commits/097aafd75d76c22eeb6ed5c43c770e50e19c8e5c

Problem solved for us :) Thank you again for this great patch!

drupol:
Heya, 

Apologies for commenting again on this issue, but it seems the problem still persists. 😞  

I've tested it on my machine and observed the same issue:  

Screenshot showing the issue  

Additionally, some friends tested the script and encountered identical results. You can find their feedback here:  
- https://github.com/NixOS/nixpkgs/pull/372987#issuecomment-2585915469
- https://gist.github.com/Ma27/21a597f842bed3cdf5035470171828a9

### Steps to Reproduce  

We used the following reproducer to verify the issue: https://gist.github.com/drupol/002efef91422afa37ca7b05aa48f5e0c

#### Required Dependencies  

To run the reproducer, you will need:  
- Composer (latest `dev-main` branch)  
- Git
- Diffoscope

> **Note:** Using Nix is optional for reproducing this issue. However, if you're using Nix, you can quickly set up a shell environment with all the dependencies using this command:  
> ```bash  
> nix shell github:NixOS/nixpkgs/nixpkgs-unstable#phpPackages.composer nixpkgs#gitMinimal nixpkgs#diffoscopeMinimal  
> ```  

Looking forward to your thoughts on this. Let me know if additional details or logs would help!

Seldaek:
I cannot repro this with latest composer snapshot.. I see your version is 2.8.4 so I am thinking that's the issue, you just don't have the fix

drupol:
Hello Jordi,

I will try again today at the end of the day, but all I can tell so far is that the fix has been applied in 2.8.4 (https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/php-packages/composer/default.nix).
I guess this is the only fix needed to fix that issue?

Thanks!

stof:
@drupol this PR is **not** part of the 2.8.4 release of Composer.

drupol:
> @drupol this PR is **not** part of the 2.8.4 release of Composer.

Sorry I should have been clearer. The release of Composer **in Nix** which is 2.8.4 has this patch, as you can see in here: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/php-packages/composer/default.nix

We decided to include this patch in 2.8.4 because this was critical for us to ensure reproducible PHP packages.

Once 2.8.5 will be tagged, we'll update it in Nix and lines 39 to 47 will be removed.

drupol:
I took a bit of time on my lunchtime to give it another fresh look and the issue was on my side, the patch was not propagated to the vendor directory, apologies for the noise around this. This PR fixes the issue: https://github.com/NixOS/nixpkgs/pull/375306

The reproducer's output is now void, therefore this is totally correct. 

```
❯ nix shell nixpkgs#diffoscopeMinimal github:drupol/nixpkgs/push-nvstlwpxlprr#phpPackages.composer
❯ ./repro.sh 
Composer version 2.8.4 2024-12-11 11:57:47
PHP version 8.3.15 (/nix/store/r2xq8q0ngikflpnjpbai4pycqkdycikl-php-with-extensions-8.3.15/bin/php)
Run the "diagnose" command to get more detailed diagnostics output.
/tmp/foobar/phpdocumentor /tmp/foobar
HEAD is now at 31725634c Fix version display
/tmp/foobar ❯
```

drupol:
> @drupol this PR is **not** part of the 2.8.4 release of Composer.

2.8.5 will soon be committed to `nixpkgs` and the patch has been removed: https://github.com/NixOS/nixpkgs/pull/375639

PR #12271, opened 2025-01-10
GitHubDriverTest::testFundingFormat(): expand the tests + fix bug

jrfnl:
Glad that I added some tests as this meant I found a bug in the PR I pulled previously (#12257).

The `thanks_dev` key expects a username in the format `u/gh/USERNAME`, but the call to `basename()` was stripping the `u/gh/` part off.

If the use of `basename()` is preferred here, the alternative would be to add `u/gh/` to the default URL prefix for thanks.dev. Let me know if you want me to change that.


Seldaek:
Thanks

jrfnl:
You're very welcome ;-)

PR #12289, opened 2025-01-27
Introduce COMPOSER_{,ALL_}WITH_DEPENDENCIES env var

syssi:
I would like to control `-w` and `-W` using env vars.


Seldaek (approved):
Thanks, looks good

PR #12290, opened 2025-01-29
Scripts: add ENV variable to skip script execution for certain event names

glaubinix:
This allows running `COMPOSER_SKIP_SCRIPTS="post-install-cmd" && composer install` to for instance skip a configured `post-install-cmd`


Seldaek:
Thanks :)

PR #12305, opened 2025-02-08
fix: RootPackageLoader.php constructor creates unusable fallback instance of VersionGuesser

das-peter:
Following commit enabled async execution https://github.com/composer/composer/commit/b0665981c2e80578b6692da8604edcda7b2093ab in the VersionGuesser:

```php
$promises[] = $this->process->executeAsync($cmdLine, $path)->then(function (Process $process) use (&$length, &$version, &$prettyVersion, $candidateVersion, &$promises): void {
```

However, the VersionGuesser fallback instance created in [RootPackageLoader constructor](https://github.com/composer/composer/blob/e58aad45b51c6b29b9f47800a182d7e86e654201/src/Composer/Package/Loader/RootPackageLoader.php#L63) still passes a ProcessExecutor instance _wihtout_ async enabled:

```php
$this->versionGuesser = $versionGuesser ?: new VersionGuesser($config, new ProcessExecutor($io), $this->versionParser, $io);
```

I suppose the most backwards compatible way to fix this is to adjust the constructor to actually enable async execution in the passed in ProcessExecutor instance similar to what the [RootPackageLoaderTest.php](https://github.com/composer/composer/blob/e58aad45b51c6b29b9f47800a182d7e86e654201/tests/Composer/Test/Package/Loader/RootPackageLoaderTest.php#L41).

### DIFF
--- a/src/Composer/Package/Loader/RootPackageLoader.php
+++ b/src/Composer/Package/Loader/RootPackageLoader.php
@@ -60,10 +60,10 @@ class RootPackageLoader extends ArrayLoader
 
         $this->manager = $manager;
         $this->config = $config;
-        if (!$versionGuesser) {
-          $processExecutor = new ProcessExecutor($io);
-          $processExecutor->enableAsync();
-          $versionGuesser = new VersionGuesser($config, $processExecutor, $this->versionParser);
+        if (null === $versionGuesser) {
+            $processExecutor = new ProcessExecutor($io);
+            $processExecutor->enableAsync();
+            $versionGuesser = new VersionGuesser($config, $processExecutor, $this->versionParser);
         }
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, so this is only a problem if you create a RootPackageLoader yourself not passing anything right? You didn't run into this while using Composer?

das-peter:
@Seldaek Thank you very much for getting back and working on this! And you are correct, this is _not_ a problem that is hit by composer itself but rather tools using it / its "API".

I was dealing with the following package: https://github.com/Vardot/varbase-updater/blob/df9a172c4bcfb4b1a772790f697a3601906527ea/src/Plugin/VarbaseUpdater.php#L288

Issue #12267, opened 2025-01-10
Add CodebergDriver for VcsRepository


Seldaek:
Having proper support for codeberg.org URLs would be helpful as some people are migrating their repos to it and the experience is suddenly degraded.

ssddanbrown:
Thanks @Seldaek for being open to add Codeberg. In a similar manner to GitLab, Codeberg is effectively a hosted instance of [Forgejo](https://forgejo.org/) (which is a hard fork of Gitea). 

Should the driver be aimed at the underlying technical platform (Forgejo) with the main service (Codeberg) configured as a default, so it could be used for other Forgejo instance too? Or would you prefer to keep things focused on just the service (Codeberg)?

Seldaek:
I guess it's probably fine to have forgejo support and do it like github-domains / gitlab-domains with codeberg configured by default as a forgejo-domain.

Seldaek:
Fixed by https://github.com/composer/composer/pull/12307

ssddanbrown:
Thanks you very much @glaubinix and @Seldaek for adding this (and your continued work on Composer) ❤️

PR #12307, opened 2025-02-10
Add driver to support forgejo/codeberge repositories

glaubinix:
Implements: https://github.com/composer/composer/issues/12267

Includes:
* basic driver
* config options to define domains and tokens
* minimalistic error handling in case there is no access token configured

Note: I had to make a change to the `VcsDriver` to address a PHPStan issue. This now surfaced more PHPStan issues. Happy to address, ignore or baseline them.

Sample composer.json
```
{
    "name": "acme/project",
    "version": "1.0",
    "repositories": [
        {"type": "vcs", "url":  "https://codeberg.org/glaubinix-acme/acme"}
    ],
    "require": {
        "glaubinix-acme/acme": "1.0"
    }
}
```

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -246,6 +246,22 @@ private repositories which will later be cloned in GitLab CI jobs with a
 using HTTP basic auth. By default, Composer will generate a git-over-SSH
 URL for private repositories and HTTP(S) only for public.
 
+## forgejo-domains
+
+Defaults to `["codeberg.org"]`. A list of domains of Forgejo servers.
+This is used if you use the `forgejo` repository type.
+
+## forgejo-token
+
+A list of domain names and username/access-tokens to authenticate against them. For
+example using `{"codeberg.org": {"username": "forgejo-user", "token": "access-token"}}` as the
+value of this option will let Composer authenticate against codeberg.org.
+Please note: If the package is not hosted at
+codeberg.org the domain names must be also specified with the
+[`forgejo-domains`](06-config.md#forgejo-domains) option.
+Further info can also be found [here](articles/authentication-for-private-packages.md#forgejo-token)
+
+
 ## disable-tls
 
 Defaults to `false`. If set to true all HTTPS URLs will be tried with HTTP
--- a/doc/articles/authentication-for-private-packages.md
+++ b/doc/articles/authentication-for-private-packages.md
@@ -371,10 +371,10 @@ To create a new access token, go to your [applications section on Forgejo](https
 
 When creating a Forgejo access token, make sure it has the `read:repository` scope.
 
-### Command line gitlab-token
+### Command line forgejo-token
 
 ```shell
-php composer.phar config [--global] forgejo-token.gitlab.example.org username access-token
+php composer.phar config [--global] forgejo-token.forgejo.example.org username access-token
 ```
 
 In the above command, the config key `forgejo-token.forgejo.example.org` consists of two parts:
--- a/doc/articles/authentication-for-private-packages.md
+++ b/doc/articles/authentication-for-private-packages.md
@@ -360,3 +360,41 @@ php composer.phar config [--global] --editor --auth
     }
 }
 ```
+
+## forgejo-token
+
+> **Note:** For the forge authentication to work on private Forgejo instances, the
+> [`forgejo-domains`](../06-config.md#forgejo-domains) section should also contain the URL.
+
+To create a new access token, go to your [applications section on Forgejo](https://codeberg.org/user/settings/applications)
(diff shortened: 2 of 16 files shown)


doc/06-config.md
```diff
@@ -246,6 +246,22 @@ private repositories which will later be cloned in GitLab CI jobs with a
 using HTTP basic auth. By default, Composer will generate a git-over-SSH
 URL for private repositories and HTTP(S) only for public.
 
+## forgejo-domains
+
+Defaults to `["codeberg.org"]`. A list of domains of Forgejo servers.
+This is used if you use the `forgejo` repository type.
+
+## forgejo-token
```
stof, line 254: those new config settings must be added in the JSON schema

glaubinix: Thanks! Added

res/composer-schema.json
```diff
@@ -648,6 +666,13 @@
                         "type": "string"
                     }
                 },
+                "forge-domain": {
+                    "type": "array",
+                    "description": "Defaults to `[\"codeberg.org\"]`. A list of domains of Forgejo servers. This is used if you use the `forgejo` repository type.",
```
stof, line 671: This is not used if you use the `forgejo` type (which forces using the forgejo driver). It is used when using the `vcs` type with a URL corresponding to a forgejo instance, so that the forgejo driver is used.

res/composer-schema.json
```diff
@@ -648,6 +666,13 @@
                         "type": "string"
                     }
                 },
+                "forge-domain": {
```
stof, line 669: should be `forgejo-domains` as we can configure multiple ones, as done for `github-domains` and `gitlab-domains`

res/composer-schema.json
stof: I would suggest keeping the default at the end of the description, for consistency with github and gitlab domains.

glaubinix: done

src/Composer/Repository/VcsRepository.php
```diff
@@ -83,6 +83,7 @@ public function __construct(array $repoConfig, IOInterface $io, Config $config,
             'gitlab' => 'Composer\Repository\Vcs\GitLabDriver',
             'bitbucket' => 'Composer\Repository\Vcs\GitBitbucketDriver',
             'git-bitbucket' => 'Composer\Repository\Vcs\GitBitbucketDriver',
+            'forgejo' => 'Composer\Repository\Vcs\ForgejoDriver',
```
stof, line 86: this new VCS driver needs to be added in the composer JSON schema, in the definition of VCS repositories

doc/articles/authentication-for-private-packages.md
```diff
@@ -360,3 +360,41 @@ php composer.phar config [--global] --editor --auth
     }
 }
 ```
+
+## forgejo-token
+
+> **Note:** For the forge authentication to work on private Forgejo instances, the
+> [`forgejo-domains`](../06-config.md#forgejo-domains) section should also contain the URL.
```
stof: it should contain the domain, not the URL.

src/Composer/Repository/VcsRepository.php
glaubinix: thanks!

src/Composer/Util/ForgejoUrl.php
```diff
@@ -0,0 +1,77 @@
...
+
+        return new self(
+            $match[3],
+            $match[4],
+            $originUrl,
+            sprintf('https://%s/repos/%s/%s', $apiBase, $match[3], $match[4])
+        );
+    }
+
+    public function generateSshUr(): string
```
Seldaek:
```suggestion
    public function generateSshUrl(): string
```

Seldaek:
LGTM, just would be nice to rebase it and fix that typo then if build is green let's merge!

src/Composer/Repository/Vcs/VcsDriver.php
```diff
@@ -75,6 +75,7 @@ final public function __construct(array $repoConfig, IOInterface $io, Config $co
 
     /**
      * Returns whether or not the given $identifier should be cached or not.
+     * @phpstan-assert-if-true !null $this->cache
```
glaubinix, line 78: @Seldaek I needed this change to prevent another PHPStan issue. Which then force me to add the string cast after all the `json_encode` calls in the driver when writing to the cache.

Seldaek: Ok, fixed to avoid hiding errors in https://github.com/composer/composer/pull/12307/commits/1287062a84a82c33c746773a1596fbc0f23e342e

PR #12320, opened 2025-02-18
Fix fatal error with classes with double backslashes

Fan2Shrek:
Hello, 

The following code
```php
require_once dirname(__DIR__).'/vendor/autoload.php';

var_dump(class_exists('App\\\\Kernel'));
var_dump(class_exists('App\\\\Kernel'));
```

Throws an `Fatal error: Cannot redeclare class App\Kernel`

After briefly looking into Composer's source code, I noticed that the `$file` variable is equal to `/vendor/composer/../../src//Kernel.php`.
The file is included, but the `App\\Kernel` class remains unknown.
When trying to load the class (with the second class_exists), the same file is encountered again, leading to its re-inclusion and the "cannot redeclare class" error.

We have several options to fix this:
1. `include_once` to ensure the file is never re-included
2. Add a verification around line `506` to see if `$file` contains double backslashes (`str_contains($file, '\\\\')`)
3. Throw an exception when loading a class containing double backslashes (`str_contains($class, '\\\\')`)

To me, the first option seems the best. Let me know if you think otherwise!

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -573,7 +573,7 @@ class ClassLoader
          * @return void
          */
         self::$includeFile = \Closure::bind(static function($file) {
-            include $file;
+            include_once $file;
         }, null, null);
     }
 }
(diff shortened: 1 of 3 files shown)


Seldaek:
Thanks for the PR but I cannot accept this unfortunately. You can report this as a bug to PHP if you want, maybe class_exists() could normalize these before sending the string to the autoloader.. But I don't want to fix this and slow down the autoloader for these broken use cases (this has been reported before as a similar problem with `"\Class"` with the backslash prefix.., but these are just misuses, if using a proper classname the code works)

PR #12324, opened 2025-02-22
Fix issues and improve Output from ZipArchiverTest

pscheit:
Hey 👋 

while working on a bug, I found some issues in the ZipArchiverTest and figured would be good to send those changes first.

To reproduce: Create a testcase that fails. Run a few times. Note that the zip archive written is not unlinked producing a bigger and bigger growing zipfile in $target file.

### DIFF
--- a/tests/Composer/Test/Package/Archiver/ZipArchiverTest.php
+++ b/tests/Composer/Test/Package/Archiver/ZipArchiverTest.php
@@ -18,14 +18,33 @@ use Composer\Package\Archiver\ZipArchiver;
 
 class ZipArchiverTest extends ArchiverTestCase
 {
+    /** @var list<string> */
+    private $filesToCleanup = [];
+
+    public function testSimpleFiles(): void
+    {
+        $files = [
+            'file.txt' => null,
+            'foo/bar/baz' => null,
+            'x/baz' => null,
+            'x/includeme' => null,
+        ];
+
+        if (!Platform::isWindows()) {
+            $files['zfoo' . Platform::getCwd() . '/file.txt'] = null;
+        }
+
+        $this->assertZipArchive($files);
+    }
+
     /**
      * @dataProvider provideGitignoreExcludeNegationTestCases
      */
     public function testGitignoreExcludeNegation(string $include): void
     {
-        $this->testZipArchive([
-            'docs/README.md' => '# The doc',
+        $this->assertZipArchive([
             '.gitignore' => "/*\n.*\n!.git*\n$include",
+            'docs/README.md' => '# The doc',
         ]);
     }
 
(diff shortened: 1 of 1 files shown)


tests/Composer/Test/Package/Archiver/ZipArchiverTest.php
```diff
@@ -18,14 +18,33 @@
 
 class ZipArchiverTest extends ArchiverTestCase
 {
+    /** @var list<string> */
+    private $filesToCleanup = [];
+
+    public function testSimpleFiles(): void
```
pscheit, line 24: imho a bit nicer to extract the default testcase into its own testcase

tests/Composer/Test/Package/Archiver/ZipArchiverTest.php
```diff
@@ -18,14 +18,33 @@
...
     /**
      * @dataProvider provideGitignoreExcludeNegationTestCases
      */
     public function testGitignoreExcludeNegation(string $include): void
     {
-        $this->testZipArchive([
-            'docs/README.md' => '# The doc',
+        $this->assertZipArchive([
             '.gitignore' => "/*\n.*\n!.git*\n$include",
+            'docs/README.md' => '# The doc',
```
pscheit, line 47: the ArchiveFilesFinder does sort by name, this wasnt tested. (maybe on purpose)

tests/Composer/Test/Package/Archiver/ZipArchiverTest.php
```diff
@@ -70,12 +77,20 @@ public function testZipArchive(array $files = []): void
...
+
+        $zipContents = [];
+        for ($i = 0; $i < $zip->numFiles; $i++) {
+            $path = $zip->getNameIndex($i);
+            static::assertIsString($path);
+            $zipContents[$path] = $zip->getFromName($path);
         }
         $zip->close();
 
-        unlink($target);
```
pscheit, line 78: this unlink didnt work as expected if you have a row of failing TestCases. The existing $target zip file would haven been opened and have more files appended.

tests/Composer/Test/Package/Archiver/ZipArchiverTest.php
```diff
@@ -70,12 +77,20 @@ public function testZipArchive(array $files = []): void
...
+        static::assertSame(
```
pscheit, line 89: comparing both arrays makes it much easier to get 

- which testcase I am an
- order of files
- which file is missing or too much

Before the assertion wouldnt cover that the zip archiver would ADD a file that is not expected to be there (e.g. if ignoring would not work)

pscheit:
Cannot re-run that phpstan 7.2 with a coredump 🙈

PR #12327, opened 2025-02-23
ZipArchiver: treat backslashes in folder names on Unix

pscheit:
Problem: On Linux there can be directories that contain backslashes in the name.
Repository with a folder with backslashes in it: https://github.com/psc-acme/repo19   
(you cannot test this on windows)

The showed the errors:
```
Warning: fileperms(): stat failed for /tmp/composer_archive558c5d2d06/folder/with/backslashes/README.md 
Warning: ZipArchive::addFile(): No such file or directory
```
when using the ZipArchiver to create a dist file from such a git repository.

Since all backslashes were forwarded the file wasnt found anymore.  
The test reproduces that.

### DIFF
--- a/src/Composer/Package/Archiver/ZipArchiver.php
+++ b/src/Composer/Package/Archiver/ZipArchiver.php
@@ -12,8 +12,9 @@
 
 namespace Composer\Package\Archiver;
 
-use ZipArchive;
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #12332, opened 2025-03-04
Clarify version property pattern in schema.json

Izumi-kun:
### Motivation
The primary reason for these changes is to attempt a workaround for an annoying bug in JetBrains' IDE.

### Background
I create private packages that are distributed via archives (using the `artifact` repository type). As a result, I need to specify the `version` of the packages directly in the `composer.json` file.

### What's wrong
PhpStorm considers the version `1.0.0-dev` as invalid, displaying the error: "String violates the pattern: '^v?\d+(\.\d+){0,3}|^dev-'".
In my opinion, the reason for this behavior is an incorrect implementation of the pattern matcher:

https://github.com/JetBrains/intellij-community/blob/33099feb7000ed4f798ba3f34fd28c08518c96cf/json/src/com/jetbrains/jsonSchema/impl/light/legacy/JsonSchemaObjectReadingUtils.java#L355-L386

1. It uses the `matches()` method (which attempts to match the entire region against the pattern) instead of `find()`.
2. It uses a suspicious `adaptSchemaPattern()` method, which "adapts" the original pattern and is incompatible with alternation (such as **`^v?\d+(\.\d+){0,3}|^dev-`**).

### Solution

1. According to the [Composer documentation](https://getcomposer.org/doc/04-schema.md#version), valid suffixes should be added to the pattern.
2. Following the [JSON Schema reference](https://json-schema.org/understanding-json-schema/reference/string#regexp), the regular expression is surrounded by `^...$`.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -28,7 +28,7 @@
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
-            "pattern": "^v?\\d+(\\.\\d+){0,3}|^dev-"
+            "pattern": "^v?\\d+(\\.\\d+){0,3}(-(dev|(patch|p|alpha|a|beta|b|RC)\\d*))?$|^dev-.*$"
         },
         "default-branch": {
             "type": ["boolean"],
--- a/tests/Composer/Test/Json/ComposerSchemaTest.php
+++ b/tests/Composer/Test/Json/ComposerSchemaTest.php
@@ -38,6 +38,46 @@ class ComposerSchemaTest extends TestCase
         self::assertEquals($expectedError, $this->check($json));
     }
 
+    public function versionProvider(): array
+    {
+        return [
+            ['1.0.0', true],
+            ['1.0.2', true],
+            ['1.1.0', true],
+            ['1.0.0-dev', true],
+            ['1.0.0-alpha3', true],
+            ['1.0.0-beta232', true],
+            ['1.0.0-RC', true],
+            ['v2.0.4-p', true],
+            ['dev-master', true],
+            ['0.2.5.4', true],
+
+            ['invalid', false],
+            ['1.0b', false],
+            ['1.0.0-', false],
+        ];
+    }
+
+    /**
+     * @dataProvider versionProvider
+     */
+    public function testVersionPattern(string $version, bool $isValid): void
+    {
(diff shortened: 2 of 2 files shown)


Seldaek:
Thanks yeah I was trying to keep this short for simplicity but if it causes problems it's worth expanding it 👍🏻

changeschung (approved):
Braaavooo 😂💥

Seldaek:
@changeschung the only way to never break anything is to stop moving forward. Shit happens, no reason to make fun of people.

Izumi-kun:
Oops... It looks like the version format is actually more flexible than what’s in the docs. Ideally, we should update them to avoid confusion — though nailing down the exact wording might be a little tricky.

changeschung:
And when did I? Where did I? Dude?

changeschung:
R u on drugs?

Seldaek:
Yeah I don't really want to document all the horrible formats people use to avoid promoting them, but we do have to support them for BC reasons :)

dbpolito:
This PR broke repositories with type `path (https://getcomposer.org/doc/05-repositories.md#path) on versions `2.8.8` and `2.8.7`, i had to downgrade to `2.8.6`.

Seldaek:
Yeah that is a good improvement tho, it caught a misconfiguration.

shaesen:
Hello, Since installation of Composer 2.8.8 I get this error when running composer diagnose: 

> Checking composer.lock: FAIL
> packages[121].version : Does not match the regex pattern  ^v?\d+(?:[.-]\d+){0,3}[._-]?(?:(?:stable|beta|b|RC|rc|alpha|a|patch|pl|p)(?:(?:[.-]?\d+)*+)?)?(?:[.-]?dev|\.x-dev)?(?:\+.*)?$|^dev-.*$

How can I get rid of this FAIL ? I tried looking for 'packages[121].version' string and substring in the composer .lock file but did not find anything matching. Where is the fault and how can I fix it ?

xabbuh:
Maybe that's related to #12376 not being released yet. Does the issue persist when updating Composer with `composer self-update --snapshot`?

shaesen:
I cannot self-update composer, it is managed by the provider ... I'll have to wait for them to update composer :(

The issue is not occuring on any of my local sites ...

Seldaek:
Ok https://github.com/composer/composer/releases/tag/2.8.9 is now out sorry for the delay.

PR #12348, opened 2025-03-24
Upgrade json-schema to v6

### LINKED ISSUE #12345: Update for justinrainbow/json-schema
Currently composer requires justinrainbow/json-schema ^5.3
justinrainbow/json-schema is currently on version 6.3.1.

Are there any plans to update the version of this package so it can use the latest version?

Thanks in advance.

Seldaek:
Fixes #12345 

cc @DannyvdSluijs there are still some issues I couldn't figure out, maybe you have a clue?

In JsonFile::validateJsonSchema it passes a validation request like this:

```
class stdClass#7961 (4) {
  public $$ref =>
  string(76) "file:///var/www/composer/src/Composer/Json/../../../res/composer-schema.json"
  public $additionalProperties =>
  bool(false)
  public $required =>
  array(2) {
    [0] =>
    string(4) "name"
    [1] =>
    string(11) "description"
  }
}
```

That used to work well, combining the schema file with some additional constraints about additional props and required props.. But not these seem to be ignored.

The other thing that seems to have broken from v5 to v6 is that we cannot pass arrays in anymore to be validated, it needs a `(object)` cast to work.. That seems harmless enough but was surprising as it's not mentioned anywhere.


DannyvdSluijs:
I've spent some time looking into the issue and did find some things:
> - Ignore $ref siblings & abort on infinite-loop references ([#437](https://github.com/jsonrainbow/json-schema/pull/437))
Correct in version 5 there was a misinterpretation of what the `$ref` keyword should do. This was fixed in https://github.com/jsonrainbow/json-schema/pull/437. I'll look into this to find a solution, first idea is to use the `allOf` keyword to extend a `$ref` with some additional information.

>The other thing that seems to have broken from v5 to v6 is that we cannot pass arrays in anymore to be validated, it needs a (object) cast to work.. That seems harmless enough but was surprising as it's not mentioned anywhere.
This is a difficult one apparently array support had found its way into the codebase. Object are more natural to JSON as the concept of an associative array only exists in PHP and not in JSON. This has side effects when casting object to array and vice-versa. It wasn't mentioned and therefor we are trying to fix (if possible) any issues with array data in version 6. But I'm also inclined to deprecate array data/schemas in version 6, removing it in version 7. So objects should be the way to go.

I'm a bit busy at the moment so I'm afraid this might take so time to continue investigation.

Seldaek:
No rush at all, just wanted to get the ball rolling, and thanks for the context!

src/Composer/Json/JsonFile.php
```diff
@@ -256,11 +256,12 @@ public static function validateJsonSchema(string $source, $data, int $schema, ?s
         }
 
         $validator = new Validator();
-        $validator->check($data, $schemaData);
+        $data = (object) $data;
```
stof, line 259: this won't be enough, as it is not recursive for nested JSON objects. I think you need to use `json_decode(json_encode($data))` to decode it into `stdClass` instead of associative arrays.

Seldaek: Yeah I was also suspicious of this but it seemed to fix all issues so I'm not sure.. Anyway yes probably safer to encode/decode like you suggest.

src/Composer/Json/JsonFile.php
```diff
@@ -243,24 +243,24 @@ public static function validateJsonSchema(string $source, $data, int $schema, ?s
...
 
-        $schemaData = (object) ['$ref' => $schemaFile];
+        $schemaData = (object) ['$ref' => $schemaFile, '$schema' => "https://json-schema.org/draft-04/schema#"];
 
-        if ($schema === self::LAX_SCHEMA) {
-            $schemaData->additionalProperties = true;
-            $schemaData->required = [];
-        } elseif ($schema === self::STRICT_SCHEMA && $isComposerSchemaFile) {
+        if ($schema === self::STRICT_SCHEMA && $isComposerSchemaFile) {
+            $schemaData = json_decode((string) file_get_contents($schemaFile));
```
Seldaek, line 249: So getting rid of the $ref and instead loading the whole file in memory + adding the props below seems to work well.. Does anything speak against that do you think @DannyvdSluijs ?

stof:
The Symfony CI is broken when running on a 32bits PHP runtime with this change: https://github.com/symfony/symfony/actions/runs/14262540647/job/39977141591#step:5:23. I reported it at https://github.com/jsonrainbow/json-schema/issues/816

PR #12349, opened 2025-03-24
Add support for --minimal-changes with full updates

Seldaek:
Refs #12341

### DIFF
--- a/tests/Composer/Test/Fixtures/installer/full-update-minimal-changes.test
+++ b/tests/Composer/Test/Fixtures/installer/full-update-minimal-changes.test
@@ -15,9 +15,11 @@ Updating all dependencies only updates what really must change when a minimal up
                 { "name": "root/dep", "version": "1.0.0", "require": { "dependency/pkg": "1.*", "dependency/pkg2": "1.*" } },
                 { "name": "root/dep2", "version": "1.1.0" },
                 { "name": "root/dep2", "version": "1.0.0" },
+                { "name": "dependency/pkg", "version": "2.1.0" },
                 { "name": "dependency/pkg", "version": "2.0.0" },
                 { "name": "dependency/pkg", "version": "1.1.0" },
                 { "name": "dependency/pkg", "version": "1.0.0" },
+                { "name": "dependency/pkg2", "version": "2.1.0" },
                 { "name": "dependency/pkg2", "version": "2.0.0" },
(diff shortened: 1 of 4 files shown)


naderman:
Looks good, just wonder if there should be an error/warning if --minimal-changes is used on a full update and the composer.json is unmodified, which will never result in any modifications?

Seldaek:
Well.. you ask for minimal updates you get minimal updates :p

tests/Composer/Test/Fixtures/installer/full-update-minimal-changes.test
```diff
@@ -0,0 +1,60 @@
...
+            "type": "package",
+            "package": [
+                { "name": "root/dep", "version": "2.0.0", "require": { "dependency/pkg": "2.*", "dependency/pkg2": "1.*" } },
+                { "name": "root/dep", "version": "1.0.0", "require": { "dependency/pkg": "1.*", "dependency/pkg2": "1.*" } },
+                { "name": "root/dep2", "version": "1.1.0" },
+                { "name": "root/dep2", "version": "1.0.0" },
+                { "name": "dependency/pkg", "version": "2.0.0" },
+                { "name": "dependency/pkg", "version": "1.1.0" },
+                { "name": "dependency/pkg", "version": "1.0.0" },
+                { "name": "dependency/pkg2", "version": "2.0.0" },
```
stof, line 21: I suggest having both 2.0.0 and 2.1.0 (which would both match the requirement), to cover that `--minimal-changes` will still update to 2.1.0 and not to 2.0.0 (i.e. it is not the `--minimal-deviation` that was suggested in the discussion)

Seldaek: Done in 12221a3 (I think that's what you meant anyway)

stof: yes

dkarlovi:
IIUC this update: if you didn't update `composer.json` by say requesting some version bump, it will not update anything? And if you did update say 1.0 to 1.1, it will only bump stuff required for that change (which might be nothing)?

In that case I do agree with @naderman about unmodified `composer.json` producing an error, it will help with UX since it makes clear to users nothing will happen unless they make something happen.

Seldaek:
Yes, I'll see if we can easily add a warning or not.

Issue #11635, opened 2023-09-12
[RFC] Optional dependencies


derrabus:
## Motivation

A common problem for packages is dealing with dependencies that may or may not be present.

Take for example Symfony's Form component `symfony/form`. The main purpose of that component is handling HTML forms: Mapping structured data to a rendered HTML form and mapping an HTML form back to a PHP object. A second component, Validator (`symfony/validator`) can validate arbitrary objects against a set of configured constraints.

If you combine both components in compatible versions, you get form validation. ✌🏻 However, making the validator a mandatory dependency of the form component would mean that the validator gets installed even if a project just wants the form handling, e.g. because they use custom validation logic.

The current way to handle such a dependency would be to add the validator to the `require-dev` section of the form component, so it gets installed for development/testing, and add a conflict rule to prevent the installation of incompatible validator versions in downstream projects. The version constraint of the conflict rule would be pretty much the inverse of the versions constraint listed in `require-dev`.

Sounds complicated? Well, it is, and keeping both constraints in sync is a pain.

## Proposal

I'd like to propose a third type of dependecy: optional dependencies, listed under `require-optional` with the same structure as `require` and `require-dev`. Listing a package in `require-optional` tells Composer that if the given package is going to be installed, it should allow the given version range only. For the root project, `require-optional` is merged with `require-dev`, which means an optional dependency is automatically a development dependency.

### Example

```JSON
{
    "require-optional": {
        "acme/foo": "^2.5 || ^3.1"
    }
}
```

Composer would treat this configuration as equivalent to the following

```JSON
{
    "require-dev": {
        "acme/foo": "^2.5 || ^3.1"
    },
    "conflict": {
        "acme/foo": "< 2.5 || >= 3, < 3.1 || >= 4"
    }
}
```

## Implementation

This feature is syntactic sugar. Because `require-optional` entries can be translated to conflict rule, my hope is that the resolver itself does not have to be changed.

## Backwards compatibility

If this feature is accepted, older Composer versions would not interpret a composer.json file correctly that makes use of `require-optional`. However, when consuming such a package via Packagist, the Packagist server could act as a BC proxy by merging all `require-optional` entries into `require-dev` and `conflict`.

Seldaek:
After discussing this a bit I think something like this would be better:

```
{
    "compatible": {
        "acme/foo": "^2.5 || ^3.1"
    }
}
```

[...]

PR #12354, opened 2025-03-27
[RFC / POC] Feature system : allow libraries to declare grouped optional dependencies 

joelwurtz:
Ref #11635 

## Motivation

Some packages propose optional dependencies to add optional features to their library.

This is actually mainly implemented by : 

 * Having suggest in your composer.json (but from what i remember this will be deprecated ? maybe i'm wrong there)
 * Having the dependencies in "require-dev"
 * Adding a conflict rule if necessary to avoid bad versions of the needed library 
 * Adding documentation to tell user to "require" this dependency (or maybe multiple one) if he needs a specific feature
 * Add checks in your code to determine if library is present (either by looking if a class exists or by using the `InstalledVersions` file of composer to see if the dependency is installed)
 
 Also when feature may require multiple dependencies some library author prefer to propose a intermediate package that have the necessary requires in order to offer a better DX for end user (no need to require multiple packages, better handling of versions, ...)
 
 When doing this, library author often prefers to have a monolith repository and splitting directory into multiple git repository which allows to create PR that can update both the main library and the optional dependency without too much hassle
 
 Let's be honest, the current way is complicated and add a lot of maintenance for library authors, this PR aims to add a new way in composer to handle this use case.
 
 ## Goals
 
   * Allowing a library author to declare features on its package by telling : this feature require those packages in those versions
   * Allow a library user to declare which feature he want to use
   * Allowing a library author to better detect if a feature is enabled: a dependency may be installed but maybe it was done by another library and not because the user want this specific feature
   
 ## Declaring features
 
 In order for the author of a library to declare features he will now have to do add this lines in its composer.json file : 
 
 ```json
 "features": {
     "my-awesome-feature": {
         "description": "An optional description describing the feature",
         "require": {
              "vendor/packageA": "1.0",
              "vendor/packageB": "^2.0",
              ...
         }
     }
 }
```

## Using features

In order for a user to use a feature of a library he will now have to add this lines in its composer.json file : 

 ```json
 "require": {
     "vendor/library": "^1.0",
 },
 "require-features": {
     "vendor/library": ["my-awesome-feature"]
 }
```

## Detecting features

When the library want to detect if a feature is enabled it will now be able to do this in PHP : 

```php
<?php

use Composer\InstalledVersions;

// method exists check to avoid bc break old composer version
if (method_exists(InstalledVersions::class, 'hasFeature') && InstalledVersions::hasFeature("vendor/library", "my-awesome-feature")) {
   // handling my awesome feature
}
```

## Testing features

New `--self-feature` option provided to the `install` and `update` commands in order for library authors to test features of their package.

By default (no option specified), no self feature is installed. This allow having conflicting features (by example having a feature for symfony7 and another one for symfony8).

Installing a feature should be done by using the update command : 

```
composer update --self-feature my-awesome-feature
```

It is also possible to use the install command, but this will only works in those 2 scenarios : 

 * no lock file present
 * the previous installation have the same sets of "self-features" that the one passed to the command
 
It is possible to use multiple features:

```bash
composer update --self-feature foo --self-feature bar
```

This allow library authors to install and test their packages with any possible combination of features : 

 * testing without features
 * testing with only one of the features
 * testing with a specific set of features
 * etc ....
 
## Errors
 
 In order to provider better information for end user some rules where added to the SAT resolution : 
 
 * Require a inexistant feature : It will throw an error where it tells user that this feature does not exist on the wanted library
 * Require a feature on a library that is not present : It will create a new error where it tells the user that the package is not present in its requirements
 
## None goals
 
To keep this first version simple, some changes are not considered in this PR : 
 
  * Having a feature that require another feature
  * Having a feature that provides other packages (would be nice, but can be done in another PR)
  * A default feature set and avoiding requiring default features
  * ~~SAT resolution depending on feature (if i require a feature that is only available in a specific version it would be nice to restrict the version or tell user why its current set of dependencies is not possible). There is some rules for feature and SAT but right now the implementation is "stupid"~~ (This has been done finally)
  * Simpler syntax for feature requirements, it would be nice to be able to declare feature requirement directly in the `require` section like the following : 
  
  ```
  "require": {
      "vendor/package": { "version": "^1.0", "features": ["my-awesome-feature"] },
  }
  ```
  
  However this syntax would bring a too big BC break (older versions of composer would not be able to understand this).


src/Composer/DependencyResolver/Rule.php
```diff
@@ -37,6 +37,8 @@ abstract class Rule
     public const RULE_LEARNED = 12; // int (rule id)
     public const RULE_PACKAGE_ALIAS = 13; // BasePackage
     public const RULE_PACKAGE_INVERSE_ALIAS = 14; // BasePackage
+    public const RULE_REQUIRE_FEATURE = 15; // array{packageName: string, feature: string}
+    public const RULE_PROVIDE_FEATURE = 16; // array{packageName: string, feature: string}
```
joelwurtz, line 41: Is this the correct number to use ? it seems to be a bitfield some i'm not sure about those number and since phpstan throw errors on this i may have miss something, would take any hint on how to make this correct

joelwurtz: Did not see the baseline, should be ok

Seldaek:
oof, thanks but I am afraid that's gonna take me a while to digest :D

joelwurtz:
> oof, thanks but I am afraid that's gonna take me a while to digest :D

There is no rush, it's a first implementation that should work, but since it's my first time really digging into composer i may have miss some details so take your times

Seldaek:
Yeah I haven't looked yet at the implementation but this absolutely should be update-time only, you select which features you want when you update dependencies, these are then included in the requirements in the poolbuilder, and then at install time you just have a list of things to install like usual in the lock file, features have no impact at all at install time (except that probably the lock file needs a list of features to be able to dump this in InstalledVersions style fashion if we want that.. Although I'd argue class_exists() checks should maybe be preferred at runtime to see if the dependency you need is present.

joelwurtz:
I think my wording was wrong, it's not dependencies that are resolved with all features, it is only dependencies that are required by features of the **root package** to ensure that any combination of features for the root package is valid (since it's the worst case). It is the same thing that is done with `require-dev` packages.

Dependencies provided by a package feature in require section are not resolved / fetched unless the user want it.

It will certainly prevent some possibilities (like i have a feature that allow to work with older packages) but this has been done to have something working as a first implementation, and avoid some complexity, he could always be achieved latter.

joelwurtz:
> (except that probably the lock file needs a list of features to be able to dump this in InstalledVersions style fashion if we want that.. Although I'd argue class_exists() checks should maybe be preferred at runtime to see if the dependency you need is present.

It's already done in this PR, i mainly have save the `require-feature` and `feature` field in the packages of the composer.lock so it's able to dump the `InstalledVersions` without update and there is a new method `InstalledVersions::hasFeature('vendor/A', 'saas-provider')` that allow to do those checks at runtime.

I think it's important to have that and not rely on `class_exists`, like in my example the feature may tell the user that some data will be fetched for an external site, so it require some dependencies to do that, however those libraries may be already present because they were required by another package but the user don't want to use this feature, with `class_exists` it is impossible to know the correct intent.

joelwurtz:
Hey there,

Is there any news concerning looking at this ? I really believe this would be a great feature for a lot of libraries and maintainers and simplify a lot some processes.

From my point of view this PR is ready for a first version, but i'm available if you think it need more (or even less).

Seldaek:
I'm currently trying to wrap up 2.9, then I think I probably should look at this.. Sorry it's been so long. It's too big to sneak into 2.9 on a short schedule tho especially as it'll require packagist adjustments too, so I rather take time and let people play with snapshots for a while.

joelwurtz:
No worries, if this is accepted i would gladly do the PR on packagist to make the changes, if needed.

janopae:
I like your proposal, as it would have helped me already in these kinds of situations. But I worry a bit about how dependency analysis tooling like `composer-require-checker` would handle this.

Does anyone have an idea whether it's realistic to equip `composer-require-checker` with a detection for `if(InstalledVersions::hasFeature("vendor/library", "my-awesome-feature"))` blocks?

<details>
<summary>Alternative idea that might solve some of the issue in a composer-require-checker-compatible way</summary>

In theory, you could always solve the issue this PR resolves with bridge packages, but:

1. bridge packages are a maintainance burden, as they require a seperate repository
2. they usually require the downstream project not just to declare the dependency, but also to add some setup logic to include the code

If I'm correct that 1. and 2. are the problems this solves, then I think 1. could be solved by making it easier to have multiple packages per Git repository.

So you'd have your code grouped by feature, and all features are separate packages, but live in the same repo.

Like
- my-awesome-lib
  - composer.json
  - awesome-feature-1
    - composer.json
  - awesome-feature-2
    - composer.json

This would make it much easier to analyse depdendencies statically, as the dependencies are clearly grouped by directory.

The downside is that this provides no solution to problem 2. Also, the grouping by directory might still be disruptive if the optional dependency is only used by like 3 lines of code.

So maybe it is indeed better to implement your solution.
</details>

joelwurtz:
I really don't know how `composer-require-checker` works, but i don't think it's a blocker ? If there are missing things there can always be new API methods to handle their problems.

We could have, by example, a `InstalledVersions::getFeatures("vendor/library")` method to get the list of installed features for a specific library.

For the record, this solution is **heavily inspired** by how rust/cargo does this, which works really well, there is more to this in cargo (like having default features) that may be useful here, but this can be done later (it's already big enough).

res/composer-schema.json
```diff
@@ -1006,6 +1006,37 @@
...
+            "type": "object",
+            "description": "This is an object of package name (keys) and associated features (values) that are required.",
+            "additionalProperties": {
+                "type": "array",
+                "items": {
+                    "type": "string"
+                }
+            }
+        },
+        "feature": {
```
Seldaek: I think I'd call this `features` but let's wait for @naderman feedback too.

res/composer-schema.json
```diff
@@ -1006,6 +1006,37 @@
             "additionalProperties": {
                 "type": "array"
             }
+        },
+        "require-feature" : {
```
Seldaek: Probably would call this `require-features` as other plural things are pluralized (I know require isn't but that's more a directive, and here require is the directive too but features should be plural IMO). @naderman wdyt?

Seldaek:
@joelwurtz I'm finally looking into this, will post most comments about code review later but already got some conceptual thing about the [discussion above](https://github.com/composer/composer/pull/12354#issuecomment-2778641778)

> I think my wording was wrong, it's not dependencies that are resolved with all features, it is only dependencies that are required by features of the **root package** to ensure that any combination of features for the root package is valid (since it's the worst case). It is the same thing that is done with `require-dev` packages.

Why would we require that all features can be installed together? I'd argue they very well might conflict with each other in ways like if you have support in a bundle for symfony5 or symfony6 you could put those two as features, adding requires on the framework in each version, but both cannot be installed together. 

For consumers of a package, they'd only require sensible features, and if they require too many then it'd fail to resolve.

For the package's author, i.e. the root package features, I'd argue we should never include them unless specifically enabled in `composer update --features a,b,c` (or allow putting in composer.json `require-features: { self: [foo, bar] }` so that one can set the default features that should be enabled when updating the root package, but those would not be active by default when installing the package as a dependency (that's a bit confusing so maybe we should put this somewhere else than `require-features`).

Doing this would also remove the requirement to have --feature/--no-feature flags on `install` command IMO, which would be cleaner.

Does this not make more sense?

src/Composer/Package/AliasPackage.php
```diff
@@ -395,4 +410,20 @@ public function setSourceDistReferences(string $reference): void
     {
         $this->aliasOf->setSourceDistReferences($reference);
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function getFeaturesRequires(): array
```
Seldaek: Would rename to:

```suggestion
    public function getFeatureRequires(): array
```

src/Composer/Package/AliasPackage.php
```diff
@@ -48,6 +50,10 @@ class AliasPackage extends BasePackage
     protected $provides;
     /** @var Link[] */
     protected $replaces;
+    /** @var array<string, string[]> */
+    protected $featuresRequires = [];
```
Seldaek:
```suggestion
    protected $featureRequires = [];
```

src/Composer/Package/RootPackageInterface.php
```diff
@@ -170,4 +171,18 @@ public function setSuggests(array $suggests): void;
      * @param mixed[] $extra
      */
     public function setExtra(array $extra): void;
+
+    /**
+     * Set features required
+     *
+     * @param array<string, string[]> $featuresRequires A list of features required per package
+     */
+    public function setFeaturesRequires(array $featuresRequires): void;
```
Seldaek:
```suggestion
    public function setFeatureRequires(array $featureRequires): void;
```

src/Composer/Package/Link.php
```diff
@@ -40,14 +41,15 @@ class Link
...
-     * @phpstan-var array<self::TYPE_REQUIRE|self::TYPE_DEV_REQUIRE|self::TYPE_PROVIDE|self::TYPE_CONFLICT|self::TYPE_REPLACE>
+     * @phpstan-var array<self::TYPE_REQUIRE|self::TYPE_DEV_REQUIRE|self::TYPE_PROVIDE|self::TYPE_CONFLICT|self::TYPE_REPLACE|self::TYPE_FEATURE>
      */
     public static $TYPES = [
         self::TYPE_REQUIRE,
         self::TYPE_DEV_REQUIRE,
         self::TYPE_PROVIDE,
         self::TYPE_CONFLICT,
         self::TYPE_REPLACE,
+        self::TYPE_FEATURE,
```
Seldaek: I think I would remove this new type from the array.. It's not really a package link. It's a way to add more requires, and it has a different format so all places using Link::$TYPES would need adjusting/special casing for the feature case anyway so might as well handle that manually below the loop rather than risk breaking third parties relying on this array?
```suggestion
```

src/Composer/Package/Link.php
```diff
@@ -40,14 +41,15 @@ class Link
      *
      * @internal
      * @var string[]
-     * @phpstan-var array<self::TYPE_REQUIRE|self::TYPE_DEV_REQUIRE|self::TYPE_PROVIDE|self::TYPE_CONFLICT|self::TYPE_REPLACE>
+     * @phpstan-var array<self::TYPE_REQUIRE|self::TYPE_DEV_REQUIRE|self::TYPE_PROVIDE|self::TYPE_CONFLICT|self::TYPE_REPLACE|self::TYPE_FEATURE>
```
Seldaek:
```suggestion
     * @phpstan-var array<self::TYPE_REQUIRE|self::TYPE_DEV_REQUIRE|self::TYPE_PROVIDE|self::TYPE_CONFLICT|self::TYPE_REPLACE>
```

src/Composer/Package/AliasPackage.php
```diff
@@ -67,8 +73,17 @@ public function __construct(BasePackage $aliasOf, string $version, string $prett
...
         foreach (Link::$TYPES as $type) {
-            $links = $aliasOf->{'get' . ucfirst($type)}();
-            $this->{$type} = $this->replaceSelfVersionDependencies($links, $type);
+            if ($type === Link::TYPE_FEATURE) {
+                $aliasFeatures = $aliasOf->getFeatures();
+
+                foreach ($this->features as $name => $feature) {
+                    $links = $aliasFeatures[$name]['require'] ?? [];
+                    $this->features[$name]['require'] = $this->replaceSelfVersionDependencies($links, Link::TYPE_FEATURE);
+                }
```
Seldaek: `$this->features` is empty at this stage so this cannot work, need to loop over the aliased package's features:

```suggestion
                foreach ($aliasFeatures as $name => $feature) {
                    $this->features[$name]['require'] = $this->replaceSelfVersionDependencies($aliasFeatures[$name]['require'], Link::TYPE_FEATURE);
                }
```

tests/Composer/Test/Fixtures/installer/partial-update-always-updates-symlinked-path-repos.test
```diff
@@ -38,6 +38,8 @@ Partial updates always update path repos which are symlinked, but not those whic
         { "name": "c/uptodate", "version": "2.0.0" }
     ],
     "packages-dev": [],
+    "packages-feature": [],
+    "packages-feature": [],
```
Seldaek:
```suggestion
    "packages-feature": [],
```

tests/Composer/Test/Fixtures/installer/feature-required-unknown-feature.test
```diff
@@ -0,0 +1,35 @@
...
+2
+
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
+    - Root composer.json requires a/a 1.0.0 -> satisfiable by a/a[1.0.0].
+    - Root composer.json requires feature "unknown" for a/a.
```
Seldaek: I think this should be clearer somehow.. cc @naderman 

```suggestion
    - Root composer.json requires feature "unknown" for a/a -> could not be found.
```

tests/Composer/Test/Fixtures/installer/feature-no-package.test
```diff
@@ -0,0 +1,47 @@
...
+--EXPECT-EXIT-CODE--
+2
+
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
+    - Root composer.json requires b/b, it could not be found in any version, there may be a typo in the package name.
```
Seldaek, line 37: Would be nice if we would mention `b/b is required by feature b of __root__` here.. These we can improve later if needed but just saying as I read the tests so we have a note of it.

tests/Composer/Test/Fixtures/installer/feature-duplicated-lock.test
```diff
@@ -0,0 +1,73 @@
...
+            {
+                "name": "b/b", "version": "1.0.0",
+                "type": "library"
+            },
+            {
+                "name": "c/c", "version": "1.0.0",
+                "type": "library"
+            }
+       ]
+    },
```
Seldaek: Need to think about this some more.. This seems very redundant at first sight and I wish it was just a list of features by package or something. But maybe it is needed..

Also curious why feature names are not namespaced by package name? Is this not gonna cause problems if two packages use the same feature name? At least it seems like it'd cause a less-than-clear lock file for humans inspecting it, even if no bugs.

src/Composer/InstalledVersions.php
```diff
@@ -252,6 +252,27 @@ public static function getInstallPath($packageName)
...
+     * @return bool
+     */
+    public static function hasFeature($packageName, $feature)
+    {
+        foreach (self::getInstalled() as $installed) {
+            if (!isset($installed['versions'][$packageName])) {
+                continue;
+            }
+
+            return in_array($feature, $installed['versions'][$packageName], true);
```
Seldaek: Claude review flagged this.. and I think it's correct :)

```suggestion
            if (!isset($installed['versions'][$packageName]['features'])) {
                continue;
            }

            return in_array($feature, $installed['versions'][$packageName]['features'], true);
```

src/Composer/Package/Locker.php
```diff
@@ -157,9 +157,10 @@ public function isFresh(): bool
      * Searches and returns an array of locked packages, retrieved from registered repositories.
```
Seldaek, line 159: getContentHash() needs to be adjusted above to include `features` and `feature-requires` in $relevantKeys.

src/Composer/Installer/InstallationManager.php
```diff
@@ -177,8 +177,9 @@ public function ensureBinariesPresence(PackageInterface $package): void
      * @param bool                         $devMode      whether the install is being run in dev mode
      * @param bool                         $runScripts   whether to dispatch script events
      * @param bool                         $downloadOnly whether to only download packages
+     * @param string[]|null                $restrictedRootFeatures A list of root package features that are allowed to be installed (or null if all are allowed)
      */
-    public function execute(InstalledRepositoryInterface $repo, array $operations, bool $devMode = true, bool $runScripts = true, bool $downloadOnly = false): void
+    public function execute(InstalledRepositoryInterface $repo, array $operations, bool $devMode = true, bool $runScripts = true, bool $downloadOnly = false, ?array $restrictedRootFeatures = null): void
```
Seldaek: This new arg should be forwarded to repo->write() at the end of the method

Seldaek (changes requested):
First batch of review comments.. More coming I think but it's a lot already.

Thanks overall it looks solid we just need to polish things and then I hope this can be included :)

res/composer-schema.json
joelwurtz: I'm 👍 on this, in fact it was already `features` in my first commit, but i replaced it to be consistent with `require` and some other directives.

tests/Composer/Test/Fixtures/installer/feature-duplicated-lock.test
joelwurtz: Those `packages-features` only concern root features of the library (not required one). It's mainly to know which packages to install when using `composer install --features foo` or `composer install --no-features` so there is no need for namespace by vendor / package.

Package required by features of other library are in the `packages` directives.

joelwurtz:
Thanks for the review.

> Why would we require that all features can be installed together? I'd argue they very well might conflict with each other in ways like if you have support in a bundle for symfony5 or symfony6 you could put those two as features, adding requires on the framework in each version, but both cannot be installed together.

I agree with you, some people may want to use the feature system to support different versions of a package so this will not be possible with the current implementation.

However i'm not sure if those "default" root features should be declared in the root composer.json. 

I try to imagine how a library developer would interact with this system, and i think it's important to have some command lines for those root features because of the CI.

But maybe the command line can edit the `composer.json` file and update packages ? (not a fan of this solution because when working on the dev env, it will imply a git diff, which is cumbersome)

So i'm fine to not have this constraint (it would certainly simplify some code also) but i think we need a good DX to easily switch between root features for dev and CI environment

What about having this : 

```
composer update --self-features foo --self-features bar
```

By default no self / root features are installed, this option is only available in the update command.

stof (commented):
I think this misses tests were a package changes the dependencies required by one of its features, to ensure that the pool optimizer does not break things (keeping the old version available if needed for instance)

joelwurtz:
Thanks,

It certainly misses some test, will work on the review and test next week, if you think it misses more tests i will gladly add them, it may be missing some part of composer that could interact with this (like the alias system, i'm not very familiar with this).

tests/Composer/Test/Fixtures/installer/feature-no-package.test
joelwurtz: Will try to see what could be done for that, but at some point i drop the information that a package is required because of a specific feature (this allow to avoid copying too many data).

naderman:
Have not yet gotten around to reviewing this unfortunately. But concerns I had top of mind apart from what stof mentioned were: Plugin loading order and autoloader ordering. Also I wonder if this level of change to the semantics of lock file and installed.json despite being compatible on a syntax/structure level may require a major version change. But don't want to make a definitive statement on that until I had a chance to fully review this. But maybe that's something you can already think about too.

joelwurtz:
I fixed most of the review comments if you want to take another look.

About the `package-features` in the `lock` file and installing all features, the last commit change this system, it will now work this way : 

 * By default no self feature is required, so adding a feature does not change the lock file
 * `--no-features` and `--feature` are no longer present
 * There is a new option `--self-feature` that allow to install/update a feature only for the feature of the root package (features of dependencies need to be in composer.json with the `require-features` key)
 * When a update is done on lock file with `--self-feature` option, they are registered on the `self-features` key of the lock file
 * When an install is done with `--self-feature` option
   * if no lock file: an update is done
   * if lock file present: those features are compared with the one in the lock file, if they are not the same an error is thrown telling the user that he need to do a `composer update` instead
 * There is no more the `package-features` key in the lock file
 * This allow a package to have feature that may conflict with each other
 
 I think it's better this way

tests/Composer/Test/Fixtures/installer/feature-duplicated-lock.test
joelwurtz: This is no longer present/needed since the change of allowing feature that conflict with each others, root features packages are in the "packages" section depending on the last `composer update` command

tests/Composer/Test/Fixtures/installer/feature-no-package.test
joelwurtz: I introduce a different "require" type rule for features which allow to print something like this : 

```
Problem 1
    - Root composer.json requires a/a ^1.0 -> satisfiable by a/a[1.0.0].
    - a/a 1.0.0 requires b/b 1.0.0 (using feature "b") -> found b/b[1.0.0] but it conflicts with your root composer.json require (^2.0).
```

Is that ok ?

joelwurtz:
> But concerns I had top of mind apart from what stof mentioned were: Plugin loading order and autoloader ordering

Thanks it was indeed breaking for autoloading, fixed it and updated test for plugin to check with feature, seems to works fine.

joelwurtz:
I just updated and rebased the PR with some changes and fixes : 

* All BC break removed (there is no new method in interfaces, they only exists on base class and instanceof check has been added to be sure that those methods exists)
* Removed noise on installed.json / composer.lock the self features and others keys are not writtent if they are empty, this reduce the update noise
* Add warnings when some features where silently ignored (because they don't exist due to a typo or something else, if wanted we can maybe convert them to real errors not sure which one is best)
* Add a documentation on how this new system works and what it should be used for (i have done an article, i think it is the best way to do that from what i have seen)
* `composer show` displays `features` and `requires features` (text and JSON).
* `composer require <pkg> --feature <name>` writes `require-features`
  (`vendor/pkg:feature` when requiring several packages at once).
* `composer remove` drops the matching `require-features` entry.
* `composer install` inherits `self-features` from the lock file, so CI does not
  have to repeat the flags.

PR #12356, opened 2025-03-28
feat: add COMPOSER_MAX_PARALLEL_PROCESSES 

ictbeheer:
This PR introduces the environment variable COMPOSER_MAX_PARALLEL_PROCESSES to increase or decrease the number parallel processes initiated by the ProcessExecutor.

In our case we want to decrease the number as 10 simultaneous unzip processes puts a severe strain on our server. (Especially when there are multiple composer installs being performed at the same time.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1274,7 +1274,7 @@ in performance gains.
 ### COMPOSER_MAX_PARALLEL_PROCESSES
 
 Set to an an integer to configure how many processes can be executed in parallel.
-This defaults to 10 and must be between 1 an 50.
+This defaults to 10 and must be between 1 and 50.
 
 ### COMPOSER_IPRESOLVE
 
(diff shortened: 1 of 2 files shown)


doc/03-cli.md
```diff
@@ -1271,6 +1271,11 @@ defaults to 12 and must be between 1 and 50. If your proxy has issues with
 concurrency maybe you want to lower this. Increasing it should generally not result
 in performance gains.
 
+### COMPOSER_MAX_PARALLEL_PROCESSES
+
+Set to an an integer to configure how many processes can be executed in parallel.
+This defaults to 10 and must be between 1 an 50.
```
Seldaek, line 1277:
```suggestion
This defaults to 10 and must be between 1 and 50.
```

src/Composer/Util/ProcessExecutor.php
```diff
@@ -78,6 +78,9 @@ class ProcessExecutor
     public function __construct(?IOInterface $io = null)
     {
         $this->io = $io;
+        if (is_numeric($maxJobs = Platform::getEnv('COMPOSER_MAX_PARALLEL_PROCESSES'))) {
+            $this->maxJobs = max(1, min(50, (int) $maxJobs));
+        }
```
Seldaek, line 83:
```suggestion
        $this->resetMaxJobs();
```

Seldaek:
Thanks

PR #12363, opened 2025-04-03
CurlDownloader - Add support for HTTP/3

Ayesh:
PHP 8.4 should be able to support HTTP/3 if the Curl extension is compiled with an OpenSSL branch with HTTP/3 support.

GitHub, GitLab, and packagist.org do not seem to support HTTP/3 yet, but we can still safely enable HTTP/3 if it's available to future-proof it.

This also updates the `diagnose` command to show the HTTP versions supported.

image

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -614,12 +614,20 @@ EOT
 
             $version = curl_version();
             $hasZstd = isset($version['features']) && defined('CURL_VERSION_ZSTD') && 0 !== ($version['features'] & CURL_VERSION_ZSTD);
+            $httpVersions = '1.0, 1.1';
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, looks good

danshumaker:
Possibly related: https://stackoverflow.com/questions/79550550/composer-update-and-composer-create-project-are-timing-out-getting-the-https-r

Ayesh:
@danshumaker this will only _add_ support for HTTP/3, but not change if HTTP/3 is not supported by Curl.

Right now, none of the mainstream distros ship a ext-curl build with HTTP/3 enabled. So about that SO question, this PR won't help, unfortunately.

danshumaker:
Thank you @Ayesh .

Issue #12365, opened 2025-04-04
Composer 2.8.7 is broken on 32bits builds of PHP


stof:
The root cause is the upgrade of json-schema to v6 because of https://github.com/jsonrainbow/json-schema/issues/816.

`shivammatur/setup-php` supports a way of using a 32bits build of PHP in CI on Ubuntu. See https://github.com/shivammathur/setup-php#multi-arch-setup and https://github.com/shivammathur/setup-php/issues/302#issuecomment-705528321

It could make sense to add a job running on 32bits to ensure Composer stays compatible with such runtime.

rosasurfer:
Happens on creating/validating `~/.config/composer/auth.json`

![Image](https://github.com/user-attachments/assets/79bae42e-3ede-4058-a400-2b23a8387bb7)

PR #12366, opened 2025-04-04
Add a CI job running with a 32bits build of PHP

stof:
Refs https://github.com/composer/composer/issues/12365

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -131,40 +131,6 @@ jobs:
         if: "matrix.php-version == '7.3'"
         run: "vendor/bin/simple-phpunit --configuration tests/complete.phpunit.xml"
 
-  tests_x86:
-    name: "CI on PHP x86"
-
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


Seldaek:
I'm a little unsure if this is justified, that'll spend a lot of cpu cycles for a one-off issue with 32bit which is really quite rare these days..

stof:
Maybe this can be a separate workflow running in a cron (maybe once a week).

I think it still make sense to be able to discover issues with 32bits compat in composer before it reaches released versions, given how composer is a fundamental tool in the ecosystem.

Seldaek:
Yeah, cron wouldn't have caught this though as I released too quick after merge.. So rather run this only on main, it'll limit the noise/cpu waste while still keeping the safeguard.

stof:
Idea: run it weekly and anytime the composer.lock changes

Seldaek:
I think running on merge whenever something reaches main is ok, if we update dependencies that'll run it anyway..

Seldaek:
Merging now to make sure it fails before I update json-schema

PR #12367, opened 2025-04-04
Fix support for various version schemes in the composer schema

### LINKED ISSUE #12364: Composer 2.8.7 breaks "rc" releases because of updated regex pattern
My `composer.json`:

```json
{
  "name": "opl/opl_about_profile",
  "description": "OPL About site.",
  "type": "project",
  "homepage": "https://www.drupal.org/project/drupal",
  ...
  "autoload": {
    "files": [
      "load.environment.php"
    ]
  },
  "version": "1.0.0-rc.8"
}

```

When I run this command: 

```
composer install
```

I get the following output:

```
In Factory.php line 317:
                                                                               
   - version : Does not match the regex pattern ^v?\d+(\.\d+){0,3}(-(dev|(pat  
  ch|p|alpha|a|beta|b|RC)\d*))?$|^dev-.*$       

```

Run the install without error.

The regex doesn't match on the version string anymore with "rc.1" and seems to require it to be "RC.1" instead now? It worked fine in 2.8.6 but in 2.8.7 it's now enforcing some new semver rule?

Seldaek:
Fixes #12364

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -28,7 +28,7 @@
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
-            "pattern": "^v?\\d+(\\.\\d+){0,3}(-(dev|(patch|p|alpha|a|beta|b|RC)\\d*))?$|^dev-.*$"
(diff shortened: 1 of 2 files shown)


lipingchen666:
Hi @Seldaek ! Will you guys be releasing a patch today including this fix?

Seldaek:
Yes that's my plan just have to fix some other issues first

PR #12368, opened 2025-04-04
Fix the git config for the php32bit workflow

stof:
Apparently, the container image provided by shivammatur for their 32bits setup does not configure the git safe directory by default, unlike the default runner environment.


stof:
a few tests should also be skipped for 32bits PHP because their assertions regarding the `php-64bit` platform package are based on the fact that the runtime itself matches that package.

Seldaek:
I have reached out to shivammatur about the remaining failures which are all related to weird file permission issues.

Seldaek:
I think we can close this now, works fine on main branch

PR #12371, opened 2025-04-08
Don't run bump after update with `--lock`

HypeMC:
I don't think it makes sense to run the `bump` command when `--lock` is used, since that option should only update the hash in the lock file.


Seldaek:
Makes sense thanks

loco993 (approved):

PR #12372, opened 2025-04-09
Add support to authorize with Custom Headers in auth.json or COMPOSER_AUTH env

### LINKED ISSUE #10165: Authorisation with Custom Headers in auth.json
My `composer.json`:

```json
{
    "config": {},
    "repositories": [{
        "type": "composer",
        "url": "https://secure.repo.url",
        "options":  {
            "http": {
                "header": [
                    "cf-access-token: <token>"
                ]
            }
        }
    }]
}

```

From the documentation here: https://getcomposer.org/doc/articles/authentication-for-private-packages.md#custom-token-authentication

It seems that there is no way to use this method of authentication securely. There is no way to define custom headers and tokens in `auth.json`. Additionally, repositories do not "merge" with the global composer file: so you cannot just define the secure repo in the project `composer.json` and then the authentication settings in the global one.

The only solution I have found so far is to not define the repository at all at the project level that does not seem ideal.

Composer configuration also does not support environment variables.

I noticed there is also a `bearer` auth method but only works if the `Authorization` header is required.

It seems like there should be a way of specifying secure/secret headers in `auth.json`.

stankolubomir:
Fixes #10165 

Thanks @pulzarraider for support.

### DIFF
--- a/src/Composer/Config.php
+++ b/src/Composer/Config.php
@@ -84,6 +84,7 @@ class Config
         'gitlab-token' => [],
         'http-basic' => [],
         'bearer' => [],
+        'custom-headers' => [],
         'bump-after-update' => false,
         'allow-missing-requirements' => false,
     ];
--- a/doc/articles/authentication-for-private-packages.md
+++ b/doc/articles/authentication-for-private-packages.md
@@ -209,13 +209,13 @@ Use custom HTTP headers for authentication with private repositories that requir
 ### Command line custom-headers
 
 ```shell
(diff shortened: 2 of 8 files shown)


Seldaek (approved):
Looks good to me, thanks

src/Composer/Config.php
```diff
@@ -84,6 +84,7 @@ class Config
         'gitlab-token' => [],
         'http-basic' => [],
         'bearer' => [],
+        'custom-headers' => [],
```
stof, line 87: This `custom-headers` config needs to be added in the JSON schema

stankolubomir: `custom-headers` was added in the JSON schema.

PR #12376, opened 2025-04-10
ensure that version suffixes are case-insensitive

### LINKED ISSUE #12374: Case sensitivity for versions in compose.lock since version 2.8.8
Today I upgraded from composer 2.8.6 -> 2.8.8. 

I ran a `composer diagnose` on my project and noticed an error which I have not seen before:
```
Checking composer.lock: FAIL
packages[137].version : Does not match the regex pattern ^v?\d+(?:[.-]\d+){0,3}[._-]?(?:(?:stable|beta|b|RC|rc|alpha|a|patch|pl|p)(?:(?:[.-]?\d+)*+)?)?(?:[.-]?dev|\.x-dev)?(?:\+.*)?$|^dev-.*$
```

This did not happen on composer 2.8.6.

After further investigation, I came across this package 137 which had the version defined as `0.0.9-BETA`.
Manually changing this to `0.0.9-beta` seems to fix the issue.

**Question**
Should this '-beta' tag be case sensitive?
There doesn't seem to be any mention of this in the documentation: https://getcomposer.org/doc/04-schema.md#version

Running `composer diagnose` on version 2.8.6:
Image

Running `composer diagnose` on version 2.8.8:
Image

xabbuh:
fixes #12374

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -28,7 +28,7 @@
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
-            "pattern": "^v?\\d+(?:[.-]\\d+){0,3}[._-]?(?:(?:stable|beta|b|RC|rc|alpha|a|patch|pl|p)(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?dev|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
+            "pattern": "^[vV]?\\d+(?:[.-]\\d+){0,3}[._-]?(?:(?:[sS][tT][aA][bB][lL][eE]|[bB][eE][tT][aA]|[bB]|[rR][cC]|[aA][lL][pP][hH][aA]|[aA]|[pP][aA][tT][cC][hH]|[pP][lL]|[pP])(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?[dD][eE][vV]|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
         },
         "default-branch": {
             "type": ["boolean"],
--- a/tests/Composer/Test/Json/ComposerSchemaTest.php
+++ b/tests/Composer/Test/Json/ComposerSchemaTest.php
@@ -49,8 +49,18 @@ class ComposerSchemaTest extends TestCase
             ['1.0.2', true],
             ['1.1.0', true],
             ['1.0.0-dev', true],
+            ['1.0.0-Alpha', true],
+            ['1.0.0-ALPHA', true],
+            ['1.0.0-alphA', true],
             ['1.0.0-alpha3', true],
+            ['1.0.0-Alpha3', true],
+            ['1.0.0-ALPHA3', true],
+            ['1.0.0-Beta', true],
+            ['1.0.0-BETA', true],
+            ['1.0.0-betA', true],
             ['1.0.0-beta232', true],
+            ['1.0.0-Beta232', true],
+            ['1.0.0-BETA232', true],
             ['10.4.13beta.2', true],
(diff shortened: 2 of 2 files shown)


xabbuh:
Those suffixes were treated as being correct before #12332.

res/composer-schema.json
```diff
@@ -28,7 +28,7 @@
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
-            "pattern": "^v?\\d+(?:[.-]\\d+){0,3}[._-]?(?:(?:stable|beta|b|RC|rc|alpha|a|patch|pl|p)(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?dev|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
+            "pattern": "^v?\\d+(?:[.-]\\d+){0,3}[._-]?(?:(?:[sS][tT][aA][bB][lL][eE]|[bB][eE][tT][aA]|b|RC|rc|[aA][lL][pP][hH][aA]|a|patch|pl|p)(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?dev|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
```
Seldaek, line 31: I wonder if we should do this simpler but less correct like this:


```suggestion
            "pattern": "^v?\\d+(?:[.-]\\d+){0,3}[._-]?(?:[a-zA-Z]{1,5}?(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?dev|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
```

Or rather make it fully correct and also apply your change to a/p/pl/patch/b variants. It's ugly as hell but more correct.

res/composer-schema.json
```diff
@@ -28,7 +28,7 @@
         "version": {
             "type": "string",
             "description": "Package version, see https://getcomposer.org/doc/04-schema.md#version for more info on valid schemes.",
-            "pattern": "^v?\\d+(?:[.-]\\d+){0,3}[._-]?(?:(?:stable|beta|b|RC|rc|alpha|a|patch|pl|p)(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?dev|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
+            "pattern": "^v?\\d+(?:[.-]\\d+){0,3}[._-]?(?:(?:[sS][tT][aA][bB][lL][eE]|[bB][eE][tT][aA]|b|[rR][cC]|[aA][lL][pP][hH][aA]|[aA]|[pP][aA][tT][cC][hH]|[pP][lL]|[pP])(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?dev|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
```
Seldaek, line 31: Let's uglify some more 😆 

```suggestion
            "pattern": "^[vV]?\\d+(?:[.-]\\d+){0,3}[._-]?(?:(?:[sS][tT][aA][bB][lL][eE]|[bB][eE][tT][aA]|[bB]|[rR][cC]|[aA][lL][pP][hH][aA]|[aA]|[pP][aA][tT][cC][hH]|[pP][lL]|[pP])(?:(?:[.-]?\\d+)*+)?)?(?:[.-]?[dD][eE][vV]|\\.x-dev)?(?:\\+.*)?$|^dev-.*$"
```

The last ones with dev-/.x-dev I think I'd leave as is as those are typically done by composer and thus more normalized..

stof: should we make the whole pattern case insensitive to make it less ugly ?

stof: answering it myself: this is not supported by json-schema.

Seldaek:
Thanks!

PR #12377, opened 2025-04-11
Warn if the same package is declared at "require" and "conflict" definitions

phansys:
The "require" definition should be updated to avoid the conflicting versions instead of having the constraint split between 2 definitions.


Seldaek:
I think I agree with this in principle.. But I am really worried what the impact will be on packagist. I wonder how many packages do this wrong and will suddenly get validation errors. And technically this isn't something that is really wrong, it's merely slightly confusing when reading the requirements.

stof:
This could be investigated in Packagist:

```sql
SELECT count(r.version_id)
FROM require_link r
JOIN conflict_link c ON c.version_id = r.version_id AND c.package_name = r.package_name
```

Seldaek:
Yeah good point.. and that's too many I'd say. 26000 conflicts in 1148 packages found, including 3800 packages that have problems in dev versions (those would start erroring with validation errors).

https://packagist.org/packages/ac/web-services-bundle
https://packagist.org/packages/psi/news4ward#2.3.2
https://packagist.org/packages/superdesk/web-publisher#2.7.x-dev (knplabs/knp-time-bundle conflict is invalid..) 

These are just a few examples.

So I'd say maybe we rather close this.. Not sure it's worth the hassle to everyone otherwise for a cosmetic "correctness" fix.

stof:
`conflict` and `require-dev` defined together is not an error at all. It is a common case when defining optional dependencies (as you generally want to install those optional dependencies in dev, while `conflict` restricts the compatible versions for users of the package)

Seldaek:
Sure, but this PR is only adding warnings for `require` + `conflict`, not require-dev. 

What I meant by `including 3800 packages that have problems in dev versions` in case that's what caused your response, is that the require+conflict are present in composer.json of branches, where packagist.org errors for validation warnings. In tagged releases we allow warnings to be silently bypassed as those cannot be fixed anymore.

stof:
ah, I misunderstood what you meant with this sentence, indeed.

phansys:
Thanks for the feedback, I was not aware about the implications at Packagist.

Given your concerns, is there something I can do to opt-out this warning in the described scenario? I'd like to identify these cases in my CI pipelines.

Seldaek:
I had a closer look and it seems if we gate this behind a new CHECK_* flag (see self::CHECK_STRICT_CONSTRAINTS in the code), then it would be included in the validate command by default (as that uses CHECK_ALL), but NOT on packagist (as that does not pass any flag). So that's a good way to add extra-strict validations IMO.

naderman:
I'm not sure I even agree with the premise here. I think it may be more accurate to require foo/bar ^2.0 and conflict with 2.4.7 if that contained some weird bug making it incompatible, rather than changing the foo/bar ^2.0 constraint to foo/bar ">= 2.0.0, <= 2.4.6, ^2.4.8"

Seldaek:
> I'm not sure I even agree with the premise here. I think it may be more accurate to require foo/bar ^2.0 and conflict with 2.4.7 if that contained some weird bug making it incompatible, rather than changing the foo/bar ^2.0 constraint to foo/bar ">= 2.0.0, <= 2.4.6 || ^2.4.8"

Yup I tend to agree.. this seems too edge-casey. If anything what we could warn on (and what I've seen while exploring this today) is when the conflict matches versions that aren't in the range of the require, because then it's just dead code. But even for that I'd not warn on packagist, just in strict mode when running composer validate.

phansys:
I just updated the PR by adding the `CHECK_DUPLICATE_CONSTRAINTS` flag.

Please, note that my use case is intended to keep the constraint for a single dependency as concise as posible, regardless the boundaries they may have if provided in multiple definitions (`require` and `conflict`).

If the check of the "version intersection" for these constraints must be also be considered in this PR, I guess I can give it a try; but I need this extra check to be defined in a different validation flag, as my use case must ignore this detail.

Please, let me know what you think.

phansys:
Just for clarification about this example:

>I think it may be more accurate to require foo/bar ^2.0 and conflict with 2.4.7 if that contained some weird bug making it incompatible, rather than changing the foo/bar ^2.0 constraint to foo/bar ">= 2.0.0, <= 2.4.6 || ^2.4.8"

IMO, the constraint `>= 2.0.0, <= 2.4.6 || ^2.4.8` ([ref.](https://semver.madewithlove.com/?constraint=%3E%3D+2.0.0,+%3C%3D+2.4.6+%7C%7C+%5E2.4.8&package=symfony%2Fsymfony)) is not the more concise or practical way to define the expected resolution.

I'd use `^2.0 !=2.4.7` ([ref.](https://semver.madewithlove.com/?constraint=%5E2.0+!%3D2.4.7&package=symfony%2Fsymfony)) instead.

Seldaek:
Ok I think in the current state I'm ok with the PR, given it won't warn on packagist.org, and users of `validate` can use `--no-check-all` to skip the various constraint checks. @naderman wdyt?

naderman:
I still wouldn't merge this. I just don't see the problem with combining requires and conflicts for the same package. I'm not sure what problem we are fixing here.

Seldaek:
Alright then closing I think. Thanks for the discussion everyone.

Issue #9918, opened 2021-05-27
Unable to add repository without name using config command


grappler:
My `composer.json`:

```json
{
  "name": "wearerequired/example",
  "repositories": [],
  "require": {}
}
```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (2.0.13 => 2.0.14)
Composer version: 2.0.13
PHP version: 7.3.24
PHP binary path: /usr/bin/php
OpenSSL version: LibreSSL 2.8.3
cURL version: 7.64.1 libz 1.2.11 ssl (SecureTransport) LibreSSL/2.8.3
zip: extension not loaded, unzip present
```

When I run this command: 

```
composer config repositories vcs git@github.com:wearerequired/user-roles-adjustments.git -vvv
```

I get the following output: 

```
Reading ./composer.json (~/Development/example/composer.json)
Loading config file ~/.composer/auth.json
Loading config file ./composer.json (~/Development/example/composer.json)
Checked CA file /private/etc/ssl/cert.pem: valid
Executing command (~/Development/example): git branch -a --no-color --no-abbrev -v
Reading ~/.composer/composer.json
Loading config file ~/.composer/auth.json
Loading config file ~/.composer/composer.json (~/.composer/composer.json)
Loading config file ~/.composer/auth.json
Reading ~/.composer/auth.json
Reading ~/.composer/vendor/composer/installed.json
Loading plugin Dealerdirect\Composer\Plugin\Installers\PHPCodeSniffer\Plugin (from dealerdirect/phpcodesniffer-composer-installer, installed globally)
Running 2.0.13 (2021-04-27 13:11:08) with PHP 7.3.24-(to be removed in future macOS) on Darwin / 20.4.0
Loading config file ~/.composer/auth.json
Reading ./composer.json (~/Development/example/composer.json)


  [InvalidArgumentException]
[...]

PR #12388, opened 2025-04-21
Change config command to allow empty name when adding new repositories

JoshuaBehrens:
When working with third-party package repositories I learned to just execute a command like

```shell
$ composer config repositories.shopware-packages '{"type": "composer", "url": "https://packages.shopware.com"}'
```

and have access to new packages. I once was confused, that everyone is writing repositories as lists and after that one has a diff like that

```diff
 {
-    "repositories": [
+    "repositories": {
-        {
+        "shopware-packages": {
+            "type": "composer",
+            "url": "https://packages.shopware.com"
+        },
+        "0": {
             "type": "path",
             "url": "custom/static-plugins/*"
         }
-    ]
+    }
 }
```

Back then I was not really questioning it. Just different styles of writing it. Now as of today I am aware of more PHP internals like arrays have three indices. Therefore I better understand warnings like these in the docs

> Using JSON object notation is also possible. However, JSON key/value pairs are to be considered unordered so consistent behaviour cannot be guaranteed.

Now coming back to this topic after some years I decided to tackle this and later noticed there is an open discussion in #9918 and the comments in https://stackoverflow.com/a/32814046 . I tried to follow up on the expected results of the discussions and have this pull request as a suggestion.

With this change you are now able to run 

```shell
$ composer config repositories composer https://packages.shopware.com
```

and get this diff instead

```diff
 {
     "repositories": [
+        {
+            "type": "composer",
+            "url": "https://packages.shopware.com"
+        },
         {
             "type": "path",
             "url": "custom/static-plugins/*"
         }
     ]
 }
```

I do not see this a breaking change as the command line would not allow the command to work like this. Therefore this likely can just be a feature.

---

I hope this pull request meets requirements you expect for pull requests. I tried to update everything one might need:

* documentation
* tests
* open issues
* a reason why the change is needed
* changelog entry
* atomic commits

If you need something else just have it part of the review.

---

Sidenote: at first I tried to implement the logic for the list item to subnode but noticed, that there are so many more logic branches and changes to the regex that it became less easy to read. I noticed that the regex solution must have a reason to exist so I rather not have it a breaking change and started to introduce a new method. I did not try to make a solution that can convert list to assoc and assoc to list. At this point I just expected the fallback logic to make things work right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+  * Added support to `config repositories` command to add repositories without a name (#9918)
+
 ### [2.8.8] 2025-04-04
 
   * Fixed json schema issues with version validation (#12367)
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -862,10 +862,16 @@ In addition to modifying the config section, the `config` command also supports 
 changes to the repositories section by using it the following way:
 
 ```shell
-php composer.phar config repositories.foo vcs https://github.com/foo/bar
+php composer.phar config repositories vcs https://github.com/foo/bar
 ```
 
 If your repository requires more configuration options, you can instead pass its JSON representation :
+
+```shell
+php composer.phar config repositories '{"type": "vcs", "url": "http://svn.example.org/my-project/", "trunk-path": "master"}'
+```
+
+If you want to name the repository you can add a name to it:
 
 ```shell
 php composer.phar config repositories.foo '{"type": "vcs", "url": "http://svn.example.org/my-project/", "trunk-path": "master"}'
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -123,11 +123,16 @@ To edit the global config.json file:
 
 To add a repository:
 
+    <comment>%command.full_name% repositories vcs https://bar.com</comment>
+
+To add a repository by name:
+
     <comment>%command.full_name% repositories.foo vcs https://bar.com</comment>
 
 To remove a repository (repo is a short alias for repositories):
 
     <comment>%command.full_name% --unset repo.foo</comment>
+    <comment>%command.full_name% --unset repo.0</comment>
 
 To disable packagist:
 
@@ -731,8 +736,14 @@ EOT
         }
 
         // handle repositories
-        if (Preg::isMatchStrictGroups('/^repos?(?:itories)?\.(.+)/', $settingKey, $matches)) {
+        if (Preg::isMatchStrictGroups('/^repos?(?:itories)?($|\..+)/', $settingKey, $matches)) {
+            $matches[1] = ltrim($matches[1], '.');
+
             if ($input->getOption('unset')) {
+                if ($matches[1] === '') {
+                    throw new \RuntimeException('You must pass the index or name of the repository to remove. Example: php composer.phar config --unset repositories.foo');
+                }
+
                 $this->configSource->removeRepository($matches[1]);
 
                 return 0;
--- a/src/Composer/Config/JsonConfigSource.php
+++ b/src/Composer/Config/JsonConfigSource.php
@@ -75,10 +75,18 @@ class JsonConfigSource implements ConfigSourceInterface
                 }
             }
 
-            if ($append) {
-                $config['repositories'][$repo] = $repoConfig;
+            if ($repo === '') {
+                if ($append) {
+                    $config['repositories'][] = $repoConfig;
+                } else {
+                    array_unshift($config['repositories'], $repoConfig);
+                }
             } else {
-                $config['repositories'] = [$repo => $repoConfig] + $config['repositories'];
+                if ($append) {
+                    $config['repositories'][$repo] = $repoConfig;
+                } else {
+                    $config['repositories'] = [$repo => $repoConfig] + $config['repositories'];
+                }
             }
         }, $name, $config, $append);
     }
--- a/src/Composer/Json/JsonManipulator.php
+++ b/src/Composer/Json/JsonManipulator.php
@@ -159,7 +159,11 @@ class JsonManipulator
      */
     public function addRepository(string $name, $config, bool $append = true): bool
     {
-        return $this->addSubNode('repositories', $name, $config, $append);
+        if ($name === '') {
+            return $this->addListItem('repositories', $name, $config, $append);
+        } else {
+            return $this->addSubNode('repositories', $name, $config, $append);
+        }
     }
 
     public function removeRepository(string $name): bool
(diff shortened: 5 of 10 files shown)


Seldaek:
One more note, I think the main downside of this is that it doesn't let you unset them without knowing which index it is present at anymore.. But I guess if that is important to people they can still use named repos then.

Seldaek:
Feel free to rebase anyway to get the latest build fixes from main, BUT the changelog I usually do myself at release time so I'd say you can remove it completely that's fine by me.

dzuelke:
@Seldaek I have been thinking about this whole `repositories` stuff again, and I am wondering if a cleaner solution overall would be to:

1. add an `id`/`name`/`nickname` field to all repository types (the schema already allows extra fields, so it'd be fully backwards compatible) to refer to repository by name for programmatic manipulation;
1. have a separate `composer repositories` command for manipulating repos, which
   - allows to have a good `help` screen without overloading the one for `config`;
   - could allow arguments for the various other fields of a repo (so no need to pass in JSON anymore);
   - could even interactively allow (re-)ordering;
   - could let you do things like `composer repositories add --after otherrepoid …`
   - gets rid of a bunch of special casing inside config handling logic;
1. deprecate the object notation, because
   - it requires hashmap implementations with stable keys to handle `composer.json` in other languages;
   - the conversion to numeric keys is confusing;
   - no sorting hapens even though numeric keys might make you think that is the case;
   - two ways of skinning the same animal;
1. maybe even auto-convert any existing object notations to array notations when re-serializing `composer.json` (still readable by older versions of Composer, as noted earlier)

Seldaek:
@dzuelke yeah I think that all sounds great, it would mirror a bit the git remote command I guess (+ ordering..). Definitely would be more sensible than adding more hacks on top of the existing. But it also sounds like quite a bit of work..

JoshuaBehrens:
I admit that I don't really like how `git remote` behaves in some cases but I see the gains of the feedback. Let me cook on that :)

JoshuaBehrens:
@dzuelke @Seldaek I rewrote it to integrate a new command. I hopefully added enough test coverage (which is right now mostly happy-path). I also had to adjust the json schema and force "old" structures to the new one. I tried to only make it happen, when adding new repositories so it does not annoy too much. I made sure the name attribute for repositories is now always on top. This should make sure, they are sorted correctly so @localheinz does not have to sort it afterwards. I am very glad, that there already was a list-item-approach to disable packagist. Docs where AI powered, so if you want it to have more detailed or rewritten, I am happy to do that.

I am not sure about the php8.5 test. I do not see it directly why my changes would've bricked it. The pull request is now ready to review.

Seldaek:
Thanks @JoshuaBehrens again, this is great work. I did tweak some things in the few commits above but I think now we're in a pretty good spot with this!

Only two areas of improvement left that I see:

- I'd like to tweak the formatting from:

```json
    "repositories": [{
        "name": "foo",
        "type": "path",
        "url": "./"
    },{
        "packagist.org": false
    }]
```

to:

```json
    "repositories": [
        {
            "name": "foo",
            "type": "path",
            "url": "./"
        }, {
            "packagist.org": false
        }
    ]
```
I think it would be more readable and more obvious that there is an array of objects?

- It would be great to support marking `canonical` true/false and adding `only`/`exclude` rules too https://getcomposer.org/doc/articles/repository-priorities.md#making-repositories-non-canonical but this is really nice to have and not essential, can still be achieved with the json config for now.

JoshuaBehrens:
@Seldaek Thank you for the guidance and review. Happy to contribute :)

I can add canonical, only, exclude as options. I mean it is not that much different to the URL.

I understand and also like the syntax with the `}, {` but I doubt, that the PHP json_encode will print out anything like this. So I can have a look whether I can get RegEx to work with that, but eventually every fallback json_encode will break it. Doesn't it? Or shall I just `preg_replace('/},\\s*{/g', '}, {', $content)`-ish for everything?

Seldaek:
@JoshuaBehrens if you want to do canonical/only/exclude feel free put please work on a new PR based on this one, as I'd rather get this one merged already. The newline stuff I am trying to look at it for a bit and if that doesn't work I'll skip, it's not *that* important.

Edit: OK not worth the hassle I think :D

JoshuaBehrens:
Thank you for merging :)

dzuelke:
Very nice work @JoshuaBehrens and @Seldaek!

Should the tests for the old object key syntax not maybe remain? They all got changed to the new `name` key variant, but that means there is now no guard against regressions in the handling of that.

Might also be useful to have tests for the "evil" case of old object key syntax but `name` in the value objects, maybe even with mismatching names, like so:

```json
{
  "repositories": {
    "foo": {
      "name": "bar",
      "type": "composer",
      "url": "https://packagist.org/"
    }
  }
}
```

Seldaek:
That sounds sensible.. @JoshuaBehrens if you have time for this be my guest 😇

Kocal:
Thank you for working on this! 🙏🏻

JoshuaBehrens:
@Kocal you are welcome :)

drzraf:
After this version reach my CI I start getting tons of warning output from composer 2.9:

```
[...]
 - repositories.gravityforms/gravityforms : Matched a schema which it should not
 - repositories.gravityforms/gravityforms : Failed to match all schemas
 - repositories.gravityforms/gravityforms : Object value found, but a boolean is required
 - repositories.gravityforms/gravityforms : Does not have a value in the enumeration [false]
 - repositories.gravityforms/gravityforms : Failed to match at least one schema
 - repositories.dearflip/dearflip : Matched a schema which it should not
 - repositories.dearflip/dearflip : Failed to match all schemas
 - repositories.dearflip/dearflip : Object value found, but a boolean is required
 - repositories.dearflip/dearflip : Does not have a value in the enumeration [false]
 - repositories.dearflip/dearflip : Failed to match at least one schema
 [...]
 ```

I believe the below config is not that uncommon and the json error message are not as clear as they should (if you don't already know what & where to look for).

`$ jq '{"repositories":.repositories}' .config/composer/config.json`
```json
{
  "repositories": {
    "wpengine/advanced-custom-fields-pro": {
      "type": "composer",
      "url": "https://connect.advancedcustomfields.com"
    },
    "freemius": {
      "type": "composer",
      "url": "https://composer.freemius.com/packages.json?authorization=Basic+...."
    },
    "dearflip/dearflip": {
      "name": "dearflip/dearflip",
      "type": "package",
      "package": {
        "name": "dearflip/dearflip",
        "version": "2.3.30",
        "type": "wordpress-plugin",
        "dist": {
          "url": "https://foo.gitlab.io/paid-plugins/tarballs/dflip-2.3.30.zip",
          "type": "zip"
        }
      }
    },
    "gravityforms/gravityforms": {
      "name": "gravityforms/gravityforms",
      "type": "package",
      "package": {
        "name": "gravityforms/gravityforms",
        "version": "2.8.16",
        "type": "wordpress-plugin",
        "dist": {
          "type": "zip",
          "url": "https://www.gravityhelp.com/wp-content/plugins/gravitymanager/api.php?op=get_plugin&slug=gravityforms&key=...."
        },
        "require": {
          "composer/installers": "^1.8|^2.1",
          "gotoandplay/gravityforms-composer-installer": "^2.3"
        }
      }
    },
    "wpackagist": {
      "type": "composer",
      "url": "https://wpackagist.org"
    }
  }
}
```

dzuelke:
@drzraf the issue there is the `name` attribute inside the repository definition. It's either that or the object syntax.

Seldaek:
Yeah you ideall convert repos to an array, remove the keys and move them to name properties then you can use the new repo command with it all

Issue #12369, opened 2025-04-07
Intermittent Invalid zip file with compression ratio >99% (possible zip bomb)


deleugpn:
I'm getting the following error from time to time: 

https://github.com/composer/composer/blob/main/src/Composer/Downloader/ZipDownloader.php#L232

As we can [see here](https://github.com/composer/composer/blob/main/src/Composer/Downloader/ZipDownloader.php#L226), it relies on random checks which would explain why its intermittent and not easily reproducible.

I want to share specifically this part of the output error logs:

```
#19 4.676   - Installing vlucas/phpdotenv (v5.6.1): Extracting archive
#19 4.701    0/108 [>---------------------------]   0%
#19 4.852   50/108 [============>---------------]  46%
#19 4.974   80/108 [====================>-------]  74%
#19 5.075  100/108 [=========================>--]  92%
#19 5.121  108/108 [============================] 100%
#19 5.464 In ZipDownloader.php line 232:
#19 5.464

#19 5.464   Invalid zip file with compression ratio >99% (possible zip bomb)
#19 5.464                                                                    

#19 5.464                                                                    
#19 5.464 install [--prefer-source] [--prefer-dist] [--prefer-install PREFER-INSTALL] [--dry-run] [--download-only] [--dev] [--no-suggest] [--no-dev]
[--no-autoloader] [--no-progress] [--no-install] [--audit] [--audit-format AUDIT-FORMAT] 
[-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] 
[--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] 
[--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform -reqs] [--] [<packages>...]
  #19 5.464
  #19 ERROR: process "/bin/sh -c composer install --optimize-autoloader --no-interaction --no-dev --no-scripts" did not complete successfully: exit code  
  : 1
  ------
   > [dependencies 7/7] RUN composer install --optimize-autoloader --no-interaction --no-dev --no-scripts:                                                
  4.285   - Installing aws/aws-sdk-php (3.337.3): Extracting archive
  4.647   - Installing league/flysystem-aws-s3-v3 (3.29.0): Extracting archive

  5.464 In ZipDownloader.php line 232:
  5.464
  5.464   Invalid zip file with compression ratio >99% (possible zip bomb)     
  5.464
  5.464
[...]

PR #12399, opened 2025-05-07
Improve error message on zip bomb

deleugpn:
improvement for diagnosing issues like https://github.com/composer/composer/issues/12369

----

After installing composer from source and patching up the ZipDownloader as instructed, I managed to track down the issue to https://github.com/cebe/markdown which is a dependency of [Yii2](https://github.com/yiisoft/yii2/blob/master/composer.json#L77C10-L77C23), more specifically caused by [this file](https://github.com/cebe/markdown/blob/master/tests/markdown-data/dos.html) which would be best if it could be ignored since its a test file.

Since the library in question has last been released in 2018, I'm not holding my hopes up that it will be fixed, so we will be looking at alternatives. 

In any case, this proposed error message would have made it 1000x easier to figure out what went wrong and why.

----
Edit: reported issues on PHPStan and PHP 8.3 seems unrelated?

### DIFF
--- a/src/Composer/Downloader/ZipDownloader.php
+++ b/src/Composer/Downloader/ZipDownloader.php
@@ -229,7 +229,7 @@ class ZipDownloader extends ArchiveDownloader
                         }
                         $totalSize += $stat['size'];
                         if ($stat['size'] > $stat['comp_size'] * 200) {
-                            throw new \RuntimeException('Invalid zip file for "'.$package->getName().'" with compression ratio >99% (possible zip bomb)');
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, change is definitely good, and thanks for figuring out which package.. This let me identify a problematic file like https://github.com/cebe/markdown/blob/master/tests/markdown-data/dos2.md (makes sense that this compresses well 😆 ) and now I got a confirmed fix for this issue too in https://github.com/composer/composer/pull/12409

PR #12403, opened 2025-05-10
Allow custom scripts to be defined directly by its `Command`'s props/methods

igorsantos07:
> Builds on top of #11151; originally mentioned at https://github.com/composer/composer/issues/11134#issuecomment-2816499996

Long story short, this allows `Command` subclasses to have complete definitions inside them, instead of splitting the command and description between `composer.json` and the class contents.

I included this because, otherwise, someone may try to use `setName()` or `setDescription()` and get confused - while `setHelp()` can still only be defined inside the class.
It's also useful when defining a family of script classes, for writing dynamic descriptions, or pairing the description with the long help contents.

-----------

I'm just not sure about the change in `Command\RunScriptCommand` - was there a specific reason for it to filter only for `ScriptAliasCommand`, or could it be simply because there were no other subclasses of `Command` at that time?

Also: I guess we should now rename `$dummy` at `Console\Application`? In fact, when the `name` is defined inside the Command, the dummy value ends up being the array key :joy:

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -20,6 +20,7 @@ use Composer\Util\Silencer;
 use LogicException;
 use RuntimeException;
 use Symfony\Component\Console\Application as BaseApplication;
+use Symfony\Component\Console\Command\Command as SymfonyCommand;
 use Symfony\Component\Console\Exception\CommandNotFoundException;
 use Symfony\Component\Console\Exception\ExceptionInterface;
 use Symfony\Component\Console\Helper\HelperSet;
@@ -380,6 +381,15 @@ class Application extends BaseApplication
 
                                 $aliases = $composer['scripts-aliases'][$script] ?? [];
 
+                                //if the command points to a valid Command class, import its details directly
+                                if (is_string($dummy) && class_exists($dummy)) {
+                                    $cmd = new $dummy;
+                                    if ($cmd instanceof SymfonyCommand) {
+                                        $this->add($cmd);
+                                        continue;
+                                    }
+                                }
+
                                 $this->add(new Command\ScriptAliasCommand($script, $description, $aliases));
                             }
                         }
@@ -599,7 +609,7 @@ class Application extends BaseApplication
 
     /**
      * Initializes all the composer commands.
-     * @return \Symfony\Component\Console\Command\Command[]
+     * @return SymfonyCommand[]
      */
     protected function getDefaultCommands(): array
     {
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -385,6 +385,12 @@ class Application extends BaseApplication
                                 if (is_string($dummy) && class_exists($dummy)) {
                                     $cmd = new $dummy;
                                     if ($cmd instanceof SymfonyCommand) {
+                                        if ($cmd->getName() === '' || $cmd->getName() === null) {
+                                            $cmd->setName($script);
+                                        }
+                                        if ($cmd->getDescription() === '') {
+                                            $cmd->setDescription($description);
+                                        }
                                         $this->add($cmd);
                                         continue;
                                     }
--- a/doc/articles/scripts.md
+++ b/doc/articles/scripts.md
@@ -202,7 +202,7 @@ If you would like to run the scripts for an event manually, the syntax is:
 php composer.phar run-script [--dev] [--no-dev] script
 ```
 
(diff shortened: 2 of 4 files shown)


igorsantos07:
PHPStan is failing for a file I didn't edit, so I'm not fixing it with fear of conflicts :sweat_smile:

Seldaek:
You can rebase if you want to fix the build. I'll need some more time to review the rest sorry.

igorsantos07:
Not a trivial conflict, it changed the very place I edited :joy: 

But I've rebased, and it's all fine now - just tested these changes against our internal project (I've actually been actively using my branch as a submodule, while I develop all the new commands we'll have).

igorsantos07:
Not trying to rush things, but I was wondering if this is still somewhere on the pile of stuff to review, or if it got forgotten :sweat_smile:

Seldaek:
Not forgotten, just been off a lot over the summer and catching up slowly now.

Seldaek:
CI is failing tho and it seems related to these changes, if you have a sec to take a look

src/Composer/Console/Application.php
```diff
@@ -381,9 +382,28 @@ function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknow
 
                                 $aliases = $composer['scripts-aliases'][$script] ?? [];
 
-                                $scriptAlias = new Command\ScriptAliasCommand($script, $description, $aliases);
+                                //if the command points to a valid Command class, import its details directly
+                                if (is_string($dummy) && class_exists($dummy)) {
```
Seldaek: Should we also check here if str_ends_with 'Command' before even trying to autoload it?

src/Composer/Console/Application.php
```diff
@@ -381,9 +382,28 @@ function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknow
...
+                                    $cmd = new $dummy;
+                                    if ($cmd instanceof SymfonyCommand) {
```
Seldaek: Maybe check first with https://www.php.net/is_a before instantiating it

src/Composer/Console/Application.php
igorsantos07: I guess it makes sense :thinking:

src/Composer/Console/Application.php
igorsantos07: IA autocompleted me to [`is_subclass_of()`](https://www.php.net/is_subclass_of). Thoughts? I can't see a situation where someone would have an actual instance of `Command`, instead of a subclass of it.

Also, it feels this would be redundant together with `str_ends_with($dummy, 'Command')`?

igorsantos07:
Well... I have no clue how the current code works with Symfony\Console v5, but on v7 (the failed tests), the logic I wrote seems to be broken :clown_face: 

The main idea is: write a placeholder name for the `scripts` key in `composer.json`, and point that to the class with all the definitions; then, it would extract the _actual_ command name, aliases and descriptions from such class.
However, [`RunScriptCommand::getScripts()`](https://github.com/igorsantos07/composer/blob/41c99776b125a062e0b979d7a8e1e1eac42bae26/src/Composer/Command/RunScriptCommand.php#L165-L185) can't `Application::find()` anything with the names from `composer.json` because, surprise surprise, they got registered with the names defined inside their classes :smiling_face_with_tear: 

The whole point with that is to avoid repetition, and concentrate all command definitions inside the class itself; it doesn't make much sense, from a user POV, having to define the command name in `composer.json` and _everything else_ inside the class...

But I kinda give up trying to dig through this. At least, aliases and descriptions are still accepted :grimacing: 

------------

~Also, I wasn't able to _properly_ test things with the `highest` dependencies on my machine because, for some reason, autoload isn't picking up the classes written on the fly from `RunScriptCommandTest.php`. I made sure the code and tests work by forcing a pretty `@include('MyCommandWithDefinitions.php');`, for the lack of better ideas on how to fix autoload on Composer tests~ :sweat_smile:

igorsantos07:
Force-pushes above were trying to solve PHPStan complaints. Once it was all green, I merged `main` ~and it looks like PHPStan crashed on the 7.2 run :man_shrugging:~

Seldaek:
Thanks, and sorry this took so long :)

igorsantos07:
Awesome! Many thanks :D

My first decent contribution to a big OSS project :tada:

PR #12406, opened 2025-05-11
feat(auth.json): support for authorization with client TLS-certificates

rtm-ctrlz:
### Description

With this change it will be possible to specify client-side TLS-certificate per origin.
It could be used in cases when a repository requires TLS-certificate to make an HTTPS request, ex. private package registry.
Technically without this PR it's possible to use client certificates for registries (by using `repositories` section of `composer.json`), but in that case path to certificate/key will we stored within the `composer.lock`, but when it is stored in VCS and each developer/CI has it's own certificate it stops working, for details see #12019.

Changes made by the way:
 - [#b423afe](https://github.com/composer/composer/commit/b423afee6c16778721de092afd29e177d2569cea) `AuthHelper::addAuthenticationHeader` transfomed to `AuthHelper::addAuthenticationOptions` to allow AuthHelper configure request options not only headers - to make this pr possible
  - [#47e298c](https://github.com/composer/composer/commit/47e298cb20c62291c3417bd56fbf4d235bd4b9ba) minor change to `AuthHelper::addAuthenticationOptions`, that helps simplify unit-test a bit - just noticed that during test implementation.

### Examples (auth.json)

Minimal (certificate and key are in one file, the key is unencrypted):

```json
{
    "client-certificate": {
        "exmaple.org": {
          "local_cert": "/path/to/cert.key.pem"
        }
    }
}
```

<details>
<summary>Certificate and key are in separated files file, the key is unencrypted</summary>

```json
{
    "client-certificate": {
        "exmaple.org": {
          "local_cert": "/path/to/cert.pem",
          "local_pk": "/path/to/key.pem"
        }
    }
}
```

</details>

<details>
<summary>Certificate and key are in separated files file, the key is encrypted</summary>

```json
{
    "client-certificate": {
        "exmaple.org": {
          "local_cert": "/path/to/cert.pem",
          "local_pk": "/path/to/key.pem",
          "passphrase": "$uper$ecret"
        }
    }
}
```

</details>

### Is it breaking change?

A bit, there could be plugins that relies on `AuthHelper::addAuthenticationHeader` method. (to be resolved in near future)

But there could be issues when `auth.json` has options multiple configurations for one domain (`http-basic + client-cerntifcate`), but the same issue will appear when there is configuration `http-basic + bearer`; in this case there will be a warning in log saying `You should avoid overwriting already defined auth settings`.
Workaround for case `http-basic (or any other) + client-cerntifcate` could be achieved with using prefix for origin, but it will cause to add a lot of mocks for checks of specific authentication types, like mocks that were removed in [#47e298c](https://github.com/composer/composer/commit/47e298cb20c62291c3417bd56fbf4d235bd4b9ba).

### Is it tested?

Yes, PHPUnit + PHPStan.
Other tests: yes - in our internal registry (Satis) with access by client certificates.

### Could this PR be better?

Sure:
- extending [authentication](https://github.com/composer/composer/blob/main/src/Composer/IO/BaseIO.php#L59) beyond `array{username: string|null, password: string|null}`, that "forces" this PR to use `json_encode/json_decode` for configuration (to store within `username`) and using `password` to keep track authentication type, will make code a bit clearer and could help on implementing other authentication method
- distinguish authentications not just by domain/origin will make it possible to use configuration `http-basic + client-cerntifcate` (I don't know who will need that, but it's not the point) without conflicts.
- (beyond this PR) it could be more efficient to add support for an option `transport-options` in `auth.json` (like is done in lock file), but in this case the file will be not about authentication, but "private transport configurations" (global or per-domain).


src/Composer/Util/AuthHelper.php
```diff
@@ -227,26 +227,27 @@ public function promptAuthIfNeeded(string $url, string $origin, int $statusCode,
...
+    public function addAuthenticationOptions(array $options, string $origin, string $url): array
     {
+        $headers = &$options['http']['header'];
         if ($this->io->hasAuthentication($origin)) {
             $authenticationDisplayMessage = null;
             $auth = $this->io->getAuthentication($origin);
             if ($auth['password'] === 'bearer') {
                 $headers[] = 'Authorization: Bearer '.$auth['username'];
-            } elseif ('github.com' === $origin && 'x-oauth-basic' === $auth['password']) {
+            } elseif ('x-oauth-basic' === $auth['password'] && 'github.com' === $origin) {
```
Seldaek, line 242: Why re-order code here and below for bitbucket? IMO it's easier to read seeing github first, you directly know what this block is about

src/Composer/Util/AuthHelper.php
```diff
@@ -227,26 +227,27 @@ public function promptAuthIfNeeded(string $url, string $origin, int $statusCode,
     }
 
     /**
-     * @param string[] $headers
+     * @param mixed[] $options
      *
-     * @return string[] updated headers array
+     * @return mixed[] updated headers array
      */
-    public function addAuthenticationHeader(array $headers, string $origin, string $url): array
```
Seldaek, line 234: We probably need to keep the name here for BC as this is very possibly used by plugins..

I'd say places calling this should probably then call the old and new method, so keep old name here, and add a new method addAuthenticationOptions. Or possibly even better that one could call addAuthenticationHeader so we can deprecate it and not call it internally? But at least it'd still exist if someone relies on it..

src/Composer/Util/AuthHelper.php
rtm-ctrlz: First of all main reason for re-order if this check `in_array($origin, $this->config->get('gitlab-domains'), true)` that calls config-get - having it on the first place forces unit-tests (not related to gitlab) to have mock for this call; as the result "authorization type" moved at first place.
Other IFs doesn't have such problem but I decided to re-order them as well to keep the save order for all checks - "type then origin".

src/Composer/Util/AuthHelper.php
rtm-ctrlz: Rapid search for `addAuthenticationHeader` (excluding composer itself and forks) showed at least one project that is using `AuthHelper::addAuthenticationHeader` - [php/pie](https://github.com/php/pie).

So, definitely I agree with you.
I'll rewrite it this way:
 - keep old method (with the same name and logic) in place, so nothing will be broken for sure; and it will be possible to deprecate it in future;
 - implement `addAuthenticationOptions` as new method, that right now will use `addAuthenticationHeader` internally (to avoid code duplication);
 - internal call will use `addAuthenticationOptions`.

Seldaek: Ok but please tag it with `@deprecated` at least already, and add a trigger_error with E_USER_DEPRECATED so that we don't forget to do this later. Ping @asgrim as the only known user of this, make sure to update pie at some point when upgrading to Composer 2.9 then.

asgrim: Thanks for the ping folks, I've noted that in a ticket to handle when we have updated to Composer 2.9 :+1: php/pie#244

rtm-ctrlz: @Seldaek , have a look, please.
PR is updated and now:
 - `addAuthenticationHeader` still present, but usage will trigger deprecation warning;
 - `addAuthenticationHeader` under the hood is using `addAuthenticationOptions`; there is some overhead on nesting then, but for compatibility I'd prefer to have some overhear rather then duplicate code;
 - there is a test to check that `addAuthenticationHeader` triggers deprecation warning; I'm not sure if we need that kind of test;
  - there is a test to check that `addAuthenticationHeader` is working with HTTP-basic (with help of `basicHttpAuthenticationProvider`), I think it should be enough.

src/Composer/Util/AuthHelper.php
rtm-ctrlz: @Seldaek updated PR to have minimal changes (`$this->config->get('gitlab-domains')

rtm-ctrlz:
@Seldaek I updated PR:

- `AuthHelper::addAuthenticationHeader` is present and deprecated, under the hood `AuthHelper::addAuthenticationOptions` is used, so it would be pretty easy to remove it in future;
- removed unnecessary changes `AuthHelper::addAuthenticationOptions` for authentication type checks;
- PR now contains also updates for docs and schema.

PR #12410, opened 2025-05-13
Always allow running self-update, but with warning

### LINKED ISSUE #12395: Composer installed as a project package has no 'self-update' command and should explain why instead of showing an error
I have a project which has composer/composer installed as a package. `composer why -t composer/composer` shows:

```
composer/composer 2.8.5 Composer helps you declare, manage and install dependencies of PHP projects. It ensures you have the right stack everywhere.
└──drupal/core-dev 10.4.2 (requires composer/composer ^2.8.1)
   └──drupal/recommended-project (requires drupal/core-dev ^10)
```

I've also added `vendor/bin` to my PATH, so that when I'm in the root of a project I can execute commands such as phpunit and phpcbf etc.

This means that when I do `composer` in the root of a project, instead of /usr/local/bin/composer, the vendor/bin/composer is run. 

This seems to do the same as the system composer, but it does not support self-update, and `composer self-update` says:

>  Command "self-update" is not defined.

This is really confusing. It would be better in this situation if the package-installed Composer explained why the command is not available.

In the meantime, I've rearranged my PATH so that /usr/local/bin comes earlier than vendor/bin.

Seldaek:
This will fail with a clear explanation why it is not available when not running from a phar

Fixes #12395

cc @joachim-n

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -78,6 +78,20 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
+        if (strpos(__FILE__, 'phar:') !== 0) {
+            if (str_contains(strtr(__DIR__, '\\', '/'), 'vendor/composer/composer')) {
+                $projDir = dirname(__DIR__, 6);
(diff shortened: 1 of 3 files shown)


joachim-n:
Thanks!

stof (approved):

PR #12418, opened 2025-05-23
Add warning when --minimal-changes is used without package and a fresh lock file

Seldaek:
Refs #12349

cc @dkarlovi

Sample:

image

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -531,6 +531,10 @@ class Installer
 
         if (!$lockTransaction->getOperations()) {
             $this->io->writeError('Nothing to modify in lock file');
+
+            if ($this->minimalUpdate && $this->updateAllowList === null && $this->locker->isFresh()) {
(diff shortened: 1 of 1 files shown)


src/Composer/Installer.php
```diff
@@ -531,6 +531,10 @@ protected function doUpdate(InstalledRepositoryInterface $localRepo, bool $doIns
 
         if (!$lockTransaction->getOperations()) {
             $this->io->writeError('Nothing to modify in lock file');
+
+            if ($this->minimalUpdate && $this->updateAllowList === null && $this->locker->isFresh()) {
+                $this->io->writeError('<warning>Updates using --minimal-changes that do not change any requirement typically will not yield any dependency changes.</warning>');
```
naderman, line 536:
```suggestion
                $this->io->writeError('<warning>The --minimal-changes option should be used with package arguments or after modifying composer.json requirements, otherwise it will likely not yield any dependency changes.</warning>');
```

naderman: Think that's a bit clearer without being too much longer?

PR #12425, opened 2025-05-27
feat: use str_contains comparations and import namespaces

shakaran:
- Use import namespaces
- Improve strpos comparations with in_array
- use str_contains


stof:
AFAICT, composer currently follows the Symfony coding standard which does not import classes from the global namespace.

PR #12437, opened 2025-06-08
Fix git prompt breaking on some systems

### LINKED ISSUE #12435: HTTP Basic auth git vs composer
I have a project that pulls packages from a gitlab installation on a custom domain. For this domain "http-basic" credentials are provided via auth.json. 
Now the problem is on some systems (and we haven't been able to exactly figure out what triggers this) these credentials will not even be used due to the following chain of events (as far as we debugged):

- composers Git Util sets `GIT_ASKPASS=echo`
- composers Git Util runs `git clone --mirror -- <URL> <DIR>` without credentials
- this results on the "broken systems" in two requests to the git server by git, as confirmed by `GIT_CURL_VERBOSE`.
The first one results in a 401 as expected, but it seems on said "broken systems" the GIT_ASKPASS=echo will feed the git prompt for user/pw back to itself so that it will do a second request with literally "Username for BASE URL TO REPO" as the username, which obviously fails and returns an `not valid: is this a git repository?`. At this point composer stops the process.

Isolating the git clone and running it separately I can confirm the behavior as well. What works is instead of `GIT_ASKPASS=echo` to set `GIT_TERMINAL_PROMPT=0` which also seems the more correct choice for composer?

If I use the composer source to do this and replace the respective environment as above, the code correctly identifies a failed authentication as the result is then a `fatal: could not read Username`, after which it will use the credentials provided by auth.json, thus installing the package.

Now I am not sure if that would be a welcome change, as the GIT_ASKPASS "hack" has been around for a long time. Unfortunately setting both still results in the original error as that is not treated as terminal prompt.

I am happy to create a PR but would like to see what the opinions are on this. I have found no way to bypass or circumvent the faulty behavior from outside as running `GIT_TERMINAL_PROMPT=0 composer install` has no effect due to the ASKPASS.

This is with composer 2.8.9 and git 2.49.0 (but also confirmed on 2.39.0)

Thanks!

Seldaek:
fixes #12435

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -438,7 +438,7 @@ class Git
     public static function getNoShowSignatureFlag(ProcessExecutor $process): string
     {
         $gitVersion = self::getVersion($process);
-        if ($gitVersion && version_compare($gitVersion, '2.10.0-rc0', '>=')) {
+        if ($gitVersion !== null && version_compare($gitVersion, '2.10.0-rc0', '>=')) {
             return ' --no-show-signature';
         }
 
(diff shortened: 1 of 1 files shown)


kitsunet:
Sorry didn't come around earlier but can confirm this works, much appreciated! :)

kdambekalns:
Any plans on the next release being tagged?

Seldaek:
@kdambekalns it's finally out sorry for the delay https://github.com/composer/composer/releases/tag/2.8.10

PR #12439, opened 2025-06-09
fix Composer schema tests

xabbuh:
following #12438


Seldaek:
Eh sorry just did this myself earlier but thanks for the fix! It was hidden by other CI failures (thanks bitbucket..)

PR #12445, opened 2025-06-12
Add forward compatibility layer for symfony/console >7.4

chalasr:
Following https://github.com/symfony/symfony/pull/60394. Please tell me if this should rather target main

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -309,7 +309,12 @@ class EventDispatcher
                     }
                     $app->setAutoExit(false);
                     $cmd = new $className($event->getName());
-                    $app->add($cmd);
+                    if (method_exists($app, 'addCommand')) {
+                        $app->addCommand($cmd);
+                    } else {
+                        // Compatibility layer for symfony/console <7.4
+                        $app->add($cmd);
+                    }
                     $app->setDefaultCommand((string) $cmd->getName(), true);
                     try {
                         $args = implode(' ', array_map(static function ($arg) { return ProcessExecutor::escape($arg); }, $additionalArgs));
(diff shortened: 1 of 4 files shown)


chalasr:
failing test looks flaky. unrelated

Seldaek:
Thanks

src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -309,7 +309,12 @@ protected function doDispatch(Event $event)
                     }
                     $app->setAutoExit(false);
                     $cmd = new $className($event->getName());
-                    $app->add($cmd);
+                    if (method_exists($app, 'addCommand')) {
+                        $app->addCommand($cmd);
+                    } else {
+                        // Compatibility layer for symfony/console <7.4
```
alexisLefebvre, line 315: Suggestion: move the comment before the `if` block, since the whole block is actually the compatibility layer.

tests/Composer/Test/InstallerTest.php
```diff
@@ -372,7 +372,7 @@ private function doTestIntegration(string $file, string $message, ?string $condi
 
             return $installer->run();
         });
-        $application->add($install);
+        method_exists($application, 'addCommand') ? $application->addCommand($install) : $application->add($install);
```
alexisLefebvre, line 375: This line doesn't have a comment like the other ones.

tests/Composer/Test/InstallerTest.php
```diff
@@ -420,7 +420,7 @@ private function doTestIntegration(string $file, string $message, ?string $condi
 
             return $installer->run();
         });
-        $application->add($update);
+        method_exists($application, 'addCommand') ? $application->addCommand($update) : $application->add($update);
```
alexisLefebvre, line 423: This line doesn't have a comment like the other ones.

Issue #8587, opened 2020-02-06
running a command thru composer scripts is waaaay slower then natively invoked


staabm:
My `composer.json`:

```json
{
  ...
    "scripts": {
        "phpstan": [
            "phpstan analyse -c phpstan.neon.dist"
        ]
    }
}

```

Output of `composer diagnose`:

```
$ composer diagnose
A script named bin would override a Composer command and has been skipped
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
require.complex/rocket : unbound version constraints (dev-CLX6_2019R03|dev-CLX6_2019R04|dev-trunk) should be avoided
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 1.9.3
PHP version: 7.0 - Package overridden via config.platform (actual: 7.2.9)
PHP binary path: C:\php-7.2\php.exe

```

When I run this command:

```
composer phpstan
```

I get a phpstan scan taking minutes and finally running into a **300 secs** timeout.

when I run `phpstan analyse -c phpstan.neon.dist` directly, I get the results in **3-4 seconds**.

alcohol:
Can't reproduce this :man_shrugging:

Seldaek:
Also not seeing this here.. Wonder what would cause it except some timeout thingy because of a script waiting for input.. BTW could be improved in 2.0 branch as that allows scripts to run in TTY mode.. Maybe worth trying it out to see if you can repro there?

staabm:
hmm I can reproduce this issue only when phpstan needs to scan "enough" files. it happens in a project on file 319 of 501 or at 296 of 477. it does not happen on the same project when scanning less files

alcohol:
I would suggest you try to use something like blackfire to get more insight into the call-graph. Without more information/details, it would just be wild guessing :-(

staabm:
I somehow have the feeling that the antivirus is getting into my way. will investigate further

[...]

PR #12456, opened 2025-06-23
Improve performance of script handlers

Seldaek:
Only create autoloaders when needed, and avoid creating duplicate autoloaders.

In pathological cases like declaring a post-package-install script as `"Foo::bar"`, it would classmap-scan the whole project on every package being installed, causing 200ms of delay at every package. Trying this on a repo here this PR improves a fresh install time from 38 down to 9seconds, which is quite dramatic. I suppose other cases where large classmaps or listener-heavy plugins are present might also see significant speedups.

Targetting 2.9 though because it does carry some risk of regression in case I missed a corner.

Refs #8587

### DIFF
--- a/src/Composer/EventDispatcher/EventDispatcher.php
+++ b/src/Composer/EventDispatcher/EventDispatcher.php
@@ -18,6 +18,7 @@ use Composer\IO\BufferIO;
 use Composer\IO\ConsoleIO;
 use Composer\IO\IOInterface;
 use Composer\Composer;
+use Composer\Package\PackageInterface;
 use Composer\PartialComposer;
 use Composer\Pcre\Preg;
 use Composer\Plugin\CommandEvent;
@@ -69,6 +70,8 @@ class EventDispatcher
     private $eventStack;
     /** @var list<string> */
     private $skipScripts;
+    /** @var string|null */
+    private $previousHash = null;
 
     /**
      * Constructor.
@@ -217,6 +220,15 @@ class EventDispatcher
                 }
                 $formattedEventNameWithArgs = $event->getName() . ($additionalArgs !== [] ? ' (' . implode(', ', $additionalArgs) . ')' : '');
                 if (!is_string($callable)) {
+                    // try setting up autoloading if we do not have a known class as target
+                    // or no valid callable (although in theory this case should not be salvageable by autoloading, it does not really hurt to try)
+                    if (
+                        (is_array($callable) && is_string($callable[0]) && !class_exists($callable[0], false))
+                        || !is_callable($callable)
+                    ) {
+                        $this->makeAutoloader($event);
+                    }
+
                     if (!is_callable($callable)) {
                         $className = is_object($callable[0]) ? get_class($callable[0]) : $callable[0];
 
@@ -271,6 +283,11 @@ class EventDispatcher
                     $className = substr($callable, 0, strpos($callable, '::'));
                     $methodName = substr($callable, strpos($callable, '::') + 2);
 
+                    // try to autoload it if the class is not known yet
+                    if (!class_exists($className, false)) {
+                        $this->makeAutoloader($event);
+                    }
+
                     if (!class_exists($className)) {
                         $this->io->writeError('<warning>Class '.$className.' is not autoloadable, can not call '.$event->getName().' script</warning>', true, IOInterface::QUIET);
                         continue;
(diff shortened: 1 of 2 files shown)


stof:
For the case in https://github.com/composer/composer/issues/8587#issuecomment-2994578569, I think another improvement could be done: `composer run-script` (either used explicitly or through a script alias) should never need to make an autoloader, as it is not running as part of the composer installation that can change the installed packages.

Seldaek:
> `composer run-script` (either used explicitly or through a script alias) should never need to make an autoloader, as it is not running as part of the composer installation that can change the installed packages.

yeah I thought about this but it is most often used for binaries which are anyway now skipping the autoloader completely, so I don't think it is worth detecting this special case (especially as it is maybe not so easy, we lack that context inside EventDispatcher).. As creating a single autoloader is often relatively cheap, the main problem is when it keeps recreating them.

The other issue is that we do not actually load the project's autoloader, so this would fail if we do not create an autoloader inside run-script.

src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -714,4 +718,37 @@ private function getCallbackIdentifier($cb): string
...
+            $hash .= $event->isDevMode() ? '/dev' : '';
+        }
+        $hash = hash('sha256', $hash);
+
+        if ($this->previousHash === $hash) {
+            return;
+        }
+
+        $this->previousHash = $hash;
+        $t = microtime(true);
```
sanmai: Looks unused :thinking:

Seldaek: Ah yeah good catch thanks

eiriksm:
I am about 98% sure this broke our (outdated) scripthandler, which had in it a require to a class being installed with composer

Like so:

```php
use DrupalFinder\DrupalFinder;
use Symfony\Component\Filesystem\Filesystem;

// ...

$fs = new Filesystem();
$drupalFinder = new DrupalFinder();
```

after CI got 2.9.0:

```
Error: Class 'DrupalFinder\DrupalFinder' not found

In ScriptHandler.php line 20:
                                               
Error: r]                                      
  Class 'DrupalFinder\DrupalFinder' not found  
                                               

Exception trace:
  at /home/runner/work/bkf/bkf/scripts/composer/ScriptHandler.php:20
 DrupalProject\composer\ScriptHandler::createRequiredFiles() at phar:///usr/local/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php:527
```

driskell:
@eiriksm I raised in #12602 with a full tiny reproduction scenario so hopefully the (amazing) composer team has everything they need to fix this :)

driskell:
A workaround for Drupal is to remove the `checkComposerVersion` pre-install and pre-update scripts - this will prevent the wrong autoloader generating and fix the post-install and post-update scripts. Of course once composer is updated with a fix likely you'll want to put the pre scripts back in.

src/Composer/EventDispatcher/EventDispatcher.php
```diff
@@ -714,4 +718,36 @@ private function getCallbackIdentifier($cb): string
...
+        $packages = $this->composer->getRepositoryManager()->getLocalRepository()->getCanonicalPackages();
+        $generator = $this->composer->getAutoloadGenerator();
+        $hash = implode(',', array_map(function (PackageInterface $p) {
+            return $p->getName().'/'.$p->getVersion();
+        }, $packages));
+        if ($event instanceof ScriptEvent) {
+            $generator->setDevMode($event->isDevMode());
+            $hash .= $event->isDevMode() ? '/dev' : '';
+        }
+        $hash = hash('sha256', $hash);
```
driskell, line 740: Is it worth just adding in the event type here? As in, if it's `pre` or `post` - as inherently the package list is highly likely to be different between them. This would ensure the post hook generates a new autoloader.

I think issue is the package list for local repository is technically the same before an install as it is after - the only difference is that the vendor folder exists after - which from an autoloader perspective I guess is completely different? Perhaps the above is the best way to fix the issue? Or if there's perhaps a flag somewhere to say if install has been completed or not?

eiriksm:
Awesome.

Also for the record. My comment reads a bit, eheh, short, when I tried to jot this down before running out the door. Sorry about that. I also have the opinion that the composer team is amazing, just to state that :heart:

eiriksm:
...and can I just also point out that projects that are using that scripthandler usually do that because they have inherited the setup from the good old drupal-composer project. Now we have official ways of doing it, and that probably does not break in this way (at least I think it would not do that)

In my case, the script is almost completely useless, so it's a good occasion to clean them up a bit :nerd_face:

Seldaek:
> My comment reads a bit, eheh, short,

No worries, just wish more people tested snapshots or RCs as that'd reduce the need for such stressful moments, but aside from that all good :D

PR #12457, opened 2025-06-24
Consistency with json and command

Jeroeny:
The suggested command `composer config -g repo.packagist false` results in `"packagist": false`. Whereas the shown JSON shows `packagist.org`.  Since Composer suggests to use the CLI as much as possible instead of manual edits, I assume that that is the desired format. Though both may work? I haven't tested it. It would make sense to have one format to specify this.


Seldaek:
The packagist.org format is actually preferred so I fixed it the other way, but yes both work the same.

PR #12459, opened 2025-06-25
Trim trailing spaces in autoload files

vrana:
PhpStorm trims trailing spaces (which I have specified in `.editorconfig`) when I do the Rename class refactoring.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -1201,6 +1201,7 @@ HEADER;
                 ]
             );
             $value = ltrim(Preg::replace('/^ */m', '    $0$0', $value));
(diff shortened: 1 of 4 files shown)


Seldaek:
Thanks

Seldaek:
@drupol this PR will most likely affect nix build hashes when 2.9.0 is released, just a heads up..

drupol:
This is super nice to let us know, thank you, I really appreciate it !!!

In the meantime, @ma27 has posted [a message on the Nix forums](https://discourse.nixos.org/t/reinforments-for-php-maintenance-needed/71794) to try to find the resources to help finding a more reliable solution.

In case you have ideas on how we could approach a reliable solution, we are all ears! I could even dedicate some time implementing the solution.

Have a great day!

PR #12103, opened 2024-09-08
Better app password support for bitbucket

eiriksm:
This fixes #12102 

It also fixes what is kind of a bad UX if you have stored actual credentials for bitbucket (for example an app password) and composer will then try to first request an access token for you using them, and in the process show some errors:

```
./composer/bin/composer update
Composer is operating slower than normal because you have Xdebug enabled. See https://getcomposer.org/xdebug
Composer could not detect the root package (my-vendor/my-package) version, defaulting to '1.0.0'. See https://getcomposer.org/root-version
Loading composer repositories with package information
Invalid OAuth consumer provided.
This can have three reasons:
1. You are authenticating with a bitbucket username/password combination
2. You are using an OAuth consumer, but didn't configure a (dummy) callback url
3. You are using an OAuth consumer, but didn't configure it as private consumer
Invalid OAuth consumer provided.
This can have three reasons:
1. You are authenticating with a bitbucket username/password combination
2. You are using an OAuth consumer, but didn't configure a (dummy) callback url
3. You are using an OAuth consumer, but didn't configure it as private consumer
Updating dependencies                                                          
Nothing to modify in lock file
Installing dependencies from lock file (including require-dev)
Package operations: 2 installs, 0 updates, 0 removals
  - Syncing eiriksm/test-dep-bitbucket (dev-main 9cf2753) into cache
  - Downloading eiriksm/test-dep-github (dev-main bd6bc34)
  - Installing eiriksm/test-dep-bitbucket (dev-main 9cf2753): Cloning 9cf275369a from cache
  - Installing eiriksm/test-dep-github (dev-main bd6bc34): Extracting archive
Generating autoload files
No security vulnerability advisories found.
```

There are actually colors and everything. Looks super scary and wrong. And 2 times even. But it's actually just warnings, and the code path that works, is executed later. Attached also an image of this scary warning

Screenshot from 2024-09-08 21-04-45

So yeah this PR fixes both of those issues. I only created an actual issue for the first of them. I could go ahead and create one for the other one if appropriate and needed for bookkeeping?

PR #12462, opened 2025-06-27
Fix Bitbucket oauth failure

ConnorLock:
Change match index from 1 to 2 to account for recently changed regex expressions for Bitbucket URLs

Based on changes from PR [#12103](https://github.com/composer/composer/pull/12103). Bitbucket URL regex was changed, but following reference to $match for $bitbucketUtil->authorizeOAuthInteractively were not updated accordingly.

### DIFF
--- a/src/Composer/Util/Git.php
+++ b/src/Composer/Util/Git.php
@@ -228,7 +228,7 @@ class Git
                     $message = 'Enter your Bitbucket credentials to access private repos';
 
                     if (!$bitbucketUtil->authorizeOAuth($domain) && $this->io->isInteractive()) {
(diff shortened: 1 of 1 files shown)


stof (approved):

Seldaek:
Makes sense, thanks for digging into it!

ConnorLock:
Thanks @Seldaek , will this make its way into main as well, or is that another pull request.

Just asking as this will be in all future versions.

Seldaek:
Yes for sure, I'll merge up to main later, and try to get a 2.8 patch release out this week some time

PR #12466, opened 2025-07-02
chore: Increase minimum version of composer/pcre

mitelg:
Hey there :wave: 

If I execute `composer update --prefer-lowest` in our project, I have a problem while checking the code with PHPStan. It will install `composer/pcre` in version `2.2.0` and the `%featureToggles.narrowPregMatches%` config in `extension.neon` causes the problem. This file was changed with `2.3.0` and updating to at least this version fixes the problem on our end. 
As we are not directly depending on the `composer/pcre` package, I made the change here.

Thanks for having a look at this and I am looking for feedback 🙂 👍

Best regards

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "symfony/finder": "^5.4.35 || ^6.3.12 || ^7.0.3",
         "symfony/process": "^5.4.35 || ^6.3.12 || ^7.0.3",
         "react/promise": "^2.11 || ^3.2",
-        "composer/pcre": "^2.2 || ^3.2",
+        "composer/pcre": "^2.3 || ^3.3",
         "symfony/polyfill-php73": "^1.24",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -38,7 +38,7 @@
         "symfony/finder": "^5.4.35 || ^6.3.12 || ^7.0.3",
         "symfony/process": "^5.4.35 || ^6.3.12 || ^7.0.3",
         "react/promise": "^2.11 || ^3.2",
-        "composer/pcre": "^2.2 || ^3.2",
+        "composer/pcre": "^2.3 || ^3.2",
```
Seldaek: This should be ^3.3 too
```suggestion
        "composer/pcre": "^2.3 || ^3.3",
```

mitelg:
thanks for merging @Seldaek :+1:

Issue #12467, opened 2025-07-03
Composer Audit non packagist repositories


sschueller:
**Is your feature request related to a problem? Please describe.**
We are using composer audit in production as some projects are not updated frequent enough to do this locally. The issue we run into is with private and internal repositories, we get authentication failures as we do not have composer authentication setup in production. The output produced is also not in the format requested like for example JSON.

**Describe the solution you'd like**
We would like to be able to define a package exclusion list or alternatively skip packages that can not be authenticated. In case of a failure it would also be nice to get the output in the requested format so that automated systems can still convey the information.

**Describe alternatives you've considered**
We have considered altering the composer files before running the check but that may bring up other issues.

**Additional context**
Output we get when authentication fails.

```
docker exec ******** composer audit --locked --format=json

In AuthHelper.php line 205:

  The 'https://packages.***********packages.json' URL could not be accessed (
  HTTP 403): HTTP/1.1 403 Forbidden


audit [--no-dev] [-f|--format FORMAT] [--locked] [--abandoned ABANDONED] [--ignore-severity IGNORE-SEVERITY]
```

Would such a feature be accepted if we made a pull request for it?

Seldaek:
I think maybe a flag --ignore-unreachable or something might work to ignore issues and just audit what can be audited, but it may hide other issues so I don't want to make this default.

As for outputting errors in json format.. I guess what happens currently is errors go to stderr and stdout is blank? If so that sounds like ok behavior to me, and better than having errors output in json format.. But maybe it'd be possible to still write some failure status in json to stdout as well if json format is requested.

And yes if you agree with the above PR welcome for sure.

Seldaek:
Fixed by https://github.com/composer/composer/pull/12470

PR #12470, opened 2025-07-08
feat: Added --ignore-unreachable flag to audit command for private/unreachable repositories

sschueller:
As discussed in https://github.com/composer/composer/issues/12467 this adds the option `--ignore-unreachable` which will let the audit command run even if repositories are unreachable. They will be listed as unreachable in the report. 

No change has been made to the overall error handling.


sschueller:
I'm not sure why phpstan failed, it ran fine on my side: https://github.com/i-web/composer/actions/runs/16144638585/job/45560116487

stof:
This implementation is ignoring all repositories when one of them is unreachable instead of ignoring only the unreachable one.
The try/catch needs to be done inside the RepositorySet to handle each repository separately.

sschueller:
@stof Thank you for catching this. I have corrected it and tested it with several of my repos now (not only the ones where the protected package is the last in the list...)

Seldaek:
Thanks!

PR #12474, opened 2025-07-10
doc: update link to CI to lead to the main branch

alexislefebvre:
The image already specify the `main` branch. Adding it to the link means that it won't show the results of CIs from other branches, PRs, etc.

PR #12477, opened 2025-07-14
chore: fixed docs wrong workflows README.md

ddevsr:
Just fixed.


Seldaek:
Fixed in #12474 already but thanks

PR #12480, opened 2025-07-17
Fix PSR-4 warnings when using exclude-from-classmap with symlinked directories

### LINKED ISSUE #12478: Absolute symlinks cause warnings for PSR-4 scanning
```
$ composer diagnose
Composer version: 2.8.10
PHP version: 8.4.8
```

Create a new project with:

```
{
    "name": "hello/hello",
    "autoload": {
        "psr-4": {
            "MyTools\\": "tools/"
        },
        "exclude-from-classmap": [
            "**/vendor/"
        ]
    }
}
```

Init another project inside the tools directory:

```
(mkdir -p tools; cd tools; composer init --name=hello/tools --require=phpunit/phpunit:^9 -n; composer install)
```

Dump the optimized autoloader:

```
$ composer dump-autoload -o
Generated optimized autoload files containing 1 classes
```

So far so good, the vendor exclusion works as expected.

Now, move the tools directory somewhere else and create an absolute symlink to it:

```
mv tools/ /tmp/
ln -s /tmp/tools 
```

Dump the autoloader again:

```
composer dump-autoload -o
```
And observe an avalance of warnings:

```
Class PHPUnit\Framework\Exception located in ./tools/vendor/phpunit/phpunit/src/Framework/Exception/Exception.php does not comply with psr-4 autoloading standard (rule: MyTools\ => ./tools). Skipping.
```

Surprisingly, there is no such issue with relative symlinks:

```
mv tools tools2
ln -s tools2 tools
composer dump-autoload -o
```

sanmai:
Fixes #12478 in 2.8

### Description

This PR fixes an issue where Composer generates incorrect PSR-4 compliance warnings when using `exclude-from-classmap` patterns with directories that are symlinks.

### The Problem

When a PSR-4 autoload path is a symlink (either relative or absolute) and the project uses `exclude-from-classmap` patterns (e.g., `**/vendor/`), Composer incorrectly reports that classes in the excluded directories don't comply with PSR-4 standards.

Example warning:
```
Class Doctrine\Instantiator\InstantiatorInterface located in ./tools/vendor/doctrine/instantiator/src/Doctrine/Instantiator/InstantiatorInterface.php does not comply with psr-4 autoloading standard (rule: MyTools\ => ./tools). Skipping.
```

### Root Cause

The issue occurs because:
1. The `buildExclusionRegex` method was only matching exclude patterns against the real path of directories
2. When scanning symlinked directories, the patterns needed to match against both the symlink path and the resolved path

### The Solution

The fix updates `buildExclusionRegex` to also check exclude patterns against the normalized (non-realpath) version of the directory being scanned. This ensures that exclude patterns work correctly whether the directory is accessed via its symlink path or its real path.

### Testing

Added a new test case `testAbsoluteSymlinkWithPsr4DoesNotGenerateWarnings` that:
1. Creates a PSR-4 autoload rule pointing to an absolute symlink
2. Uses `exclude-from-classmap` to exclude vendor directories
3. Verifies that no PSR-4 compliance warnings are generated

All existing tests continue to pass.

### Reproducing the Issue

Before this fix:
```bash
# Create a test directory with a symlink (relative or absolute)
mkdir actual-tools
ln -s actual-tools tools
# OR with absolute path:
# ln -s /tmp/tools-composer-test tools

# Add to composer.json:
{
    "autoload": {
        "psr-4": {
            "MyTools\\": "tools/"
        },
        "exclude-from-classmap": ["**/vendor/"]
    }
}

# Run composer dump-autoload -o
# Results in PSR-4 warnings for files in tools/vendor/
```

After this fix: No warnings are generated for excluded paths.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -494,11 +494,17 @@ EOF;
         if (file_exists($dir)) {
             // transform $dir in the same way that exclude-from-classmap patterns are transformed so we can match them against each other
             $dirMatch = preg_quote(strtr(realpath($dir), '\\', '/'));
+            // also match against the non-realpath version for symlinks
+            $fs = new Filesystem();
+            $absDir = $fs->isAbsolutePath($dir) ? $dir : realpath(Platform::getCwd()).'/'.$dir;
+            $dirMatchNormalized = preg_quote(strtr($fs->normalizePath($absDir), '\\', '/'));
+
             foreach ($excluded as $index => $pattern) {
                 // extract the constant string prefix of the pattern here, until we reach a non-escaped regex special character
                 $pattern = Preg::replace('{^(([^.+*?\[^\]$(){}=!<>|:\\\\#-]+|\\\\[.+*?\[^\]$(){}=!<>|:#-])*).*}', '$1', $pattern);
                 // if the pattern is not a subset or superset of $dir, it is unrelated and we skip it
-                if (0 !== strpos($pattern, $dirMatch) && 0 !== strpos($dirMatch, $pattern)) {
+                if (0 !== strpos($pattern, $dirMatch) && 0 !== strpos($dirMatch, $pattern) &&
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks!

PR #12485, opened 2025-07-23
diagnose: correctly show curl features

tbvdm:
If curl has no brotli support, `composer diagnose` should print `missing` for brotli, but instead it prints nothing:

```
curl version: 8.14.1 libz 1.3.1.1-motley brotli  zstd missing ssl LibreSSL/4.1.0
```

In the array returned by `curl_version()`, the `brotli_version` array element is set to to the empty string if brotli support is missing, not to `null`. The same goes for `libz_version` and `ssl_version`.

### DIFF
--- a/src/Composer/Command/DiagnoseCommand.php
+++ b/src/Composer/Command/DiagnoseCommand.php
@@ -613,13 +613,16 @@ EOT
             }
 
             $version = curl_version();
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks! I didn't have a system without support available so didn't notice :)

PR #12487, opened 2025-07-30
ProcessExecutor: mask GitHub fine grained access tokens similar to URL

glaubinix:
Similar to https://github.com/composer/composer/commit/90673e4f66fed79ec4d9ccf0247333b01dc3b444 wondering if the regex should be extracted into a const or similar to easier keep it in sync between the two components.

### DIFF
--- a/src/Composer/Util/GitHub.php
+++ b/src/Composer/Util/GitHub.php
@@ -23,6 +23,8 @@ use Composer\Pcre\Preg;
  */
 class GitHub
 {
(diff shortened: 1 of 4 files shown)


Seldaek:
Yes feel free to extract in a const on the Util\GitHub class perhaps.. And please rebase on 2.8 as it should be in there I'd say.

glaubinix:
Done and done.

Seldaek:
Thanks

PR #12493, opened 2025-08-10
PHP 8.5 | Prevent deprecation notices for Reflection*::setAccessible()

jrfnl:
Since PHP 8.1, calling the `Reflection*::setAccessible()` methods is no longer necessary as reflected properties/methods/etc will always be accessible. However, the method calls are still needed for PHP < 8.1.

As of PHP 8.5, calling the `Reflection*::setAccessible()` methods is now formally deprecated and will yield a deprecation notice, which will fail test runs. As of PHP 9.0, the `setAccessible()` method(s) will be removed.

With the latter in mind, this commit prevents the deprecation notice by making the calls to `setAccessible()` conditional.

Silencing the deprecation would mean, this would need to be "fixed" again come PHP 9.0, while the current solution should be stable, including for PHP 9.0.

Ref: https://wiki.php.net/rfc/deprecations_php_8_5#extreflection_deprecations


jrfnl:
P.S.: let me know if you would like me to rebase this against 2.8.

Seldaek:
Yes please rebase on 2.8, but feel free to ignore phpstan errors I'll add these to the baseline myself (I got a script here due to our dual baseline mess).

src/Composer/Console/Application.php
```diff
@@ -444,7 +444,7 @@ function_exists('php_uname') ? php_uname('s') . ' / ' . php_uname('r') : 'Unknow
             // as http error codes are all beyond the 255 range of permitted exit codes
             if ($e instanceof TransportException) {
                 $reflProp = new \ReflectionProperty($e, 'code');
-                $reflProp->setAccessible(true);
+                (\PHP_VERSION_ID < 80100) && $reflProp->setAccessible(true);
```
stof: to avoid the phpstan error about boolean in `&&`, use a `if` statement instead of using the boolean operator as a control flow operator (or use the `and` control flow operator, but this is not used in the Composer codebase AFAICT).

Seldaek: I'd rather keep `&&` and put it in the baseline for now, that doesn't hurt, and it is code that will disappear eventually.

stof: then use `and` which is the control flow operator (and won't report that error in phpstan)

jrfnl:
> Yes please rebase on 2.8, but feel free to ignore phpstan errors I'll add these to the baseline myself (I got a script here due to our dual baseline mess).

Done (and changed the "base" for the PR).

src/Composer/Console/Application.php
jrfnl: I've actioned this in a separate commit as I suspect it may be flagged for code style (use of `and` often is).
@Seldaek If you want to keep the change, please squash-merge the commits as these really shouldn't be separate.

jrfnl: > then use `and` which is the control flow operator (and won't report that error in phpstan)

@stof Hmm ... looks like that's not true.... https://github.com/composer/composer/actions/runs/17103745891/job/48507066776

@Seldaek Let me know f you'd like me to remove the commit again.

jrfnl:
FYI: I've double-checked with a code-search that there were no calls to `->setAccessible()` in the 2.8 branch which aren't in the `main` branch (and which may have been missed in the rebase).

Seldaek:
Looks like `and` didn't really help but whatever :) Thanks!

PR #12499, opened 2025-08-16
Fix suggestions from PhpStorm Inspector

weshooper:
Opened against `main` since not a bug, just tidying.

### DIFF
--- a/src/Composer/Installer/BinaryInstaller.php
+++ b/src/Composer/Installer/BinaryInstaller.php
@@ -254,15 +254,14 @@ if (PHP_VERSION_ID < 80000) {
         {
             private \$handle;
             private \$position;
-            private \$realpath;
 
             public function stream_open(\$path, \$mode, \$options, &\$opened_path)
             {
                 // get rid of phpvfscomposer:// prefix for __FILE__ & __DIR__ resolution
                 \$opened_path = substr(\$path, 17);
-                \$this->realpath = realpath(\$opened_path) ?: \$opened_path;
-                \$opened_path = $phpunitHack1\$this->realpath;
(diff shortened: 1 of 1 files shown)


src/Composer/Installer/BinaryInstaller.php
```diff
@@ -293,7 +292,7 @@ public function stream_close()
 
             public function stream_lock(\$operation)
             {
-                return \$operation ? flock(\$this->handle, \$operation) : true;
```
Seldaek: Sorry but that is NOT a readability improvement IMO..

src/Composer/Installer/BinaryInstaller.php
```diff
@@ -254,15 +254,14 @@ final class BinProxyWrapper
...
             public function stream_open(\$path, \$mode, \$options, &\$opened_path)
             {
                 // get rid of phpvfscomposer:// prefix for __FILE__ & __DIR__ resolution
                 \$opened_path = substr(\$path, 17);
-                \$this->realpath = realpath(\$opened_path) ?: \$opened_path;
-                \$opened_path = $phpunitHack1\$this->realpath;
-                \$this->handle = fopen(\$this->realpath, \$mode);
+                \$realpath = realpath(\$opened_path) ?: \$opened_path;
+                \$opened_path = $phpunitHack1\$realpath;
+                \$this->handle = fopen(\$realpath, \$mode);
```
Seldaek, line 264: This part of the change seems ok

src/Composer/Installer/BinaryInstaller.php
weshooper: @Seldaek I've reverted that one :D

Seldaek:
Alright thanks :)

PR #12505, opened 2025-08-19
fix: deprecation of curl_close in CurlDownloader.php

jorgsowa:
The function `curl_close()` doesn't have any effect since PHP 8.0: https://www.php.net/manual/en/function.curl-close.php

The deprecation of the function is implemented in PHP 8.5.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -299,7 +299,9 @@ class CurlDownloader
         if (isset($this->jobs[$id], $this->jobs[$id]['curlHandle'])) {
             $job = $this->jobs[$id];
             curl_multi_remove_handle($this->multiHandle, $job['curlHandle']);
-            curl_close($job['curlHandle']);
+            if(\PHP_VERSION_ID < 80000) {
+                curl_close($job['curlHandle']);
+            }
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks for the PR, could you please rebase it on the 2.8 branch tho so we can ship it in the next patch release (which will happen quicker)?

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -305,7 +305,9 @@ public function abortRequest(int $id): void
         if (isset($this->jobs[$id], $this->jobs[$id]['curlHandle'])) {
             $job = $this->jobs[$id];
             curl_multi_remove_handle($this->multiHandle, $job['curlHandle']);
-            curl_close($job['curlHandle']);
+            if (PHP_VERSION_ID < 80000) {
```
stof:
```suggestion
            if (\PHP_VERSION_ID < 80000) {
```

Using a qualified constant there allows OPCache to resolve the condition at compile time, eliminating it entirely.

jorgsowa: Thank you for suggestion. Changed

Seldaek:
Thanks!

PR #12506, opened 2025-08-20
Streamline IO confirmation questions to use lowercase 'y'.

thibsy:
Not sure if this should be reported as a bug, I therefore opened this against `main`. Let me know if I should update.

I was mislead by some of the IO confirmation questions which were indicating an uppercase `Y` as valid answer. I checked, and the default regex used in `Composer\Question\StrictConfirmationQuestion` does in fact not allow an uppercase `Y`. I therefore propose to adjust the remaining questions which do not already indicate a lowercase `y`.

I searched for usages of `Composer\IO\ConsoleIO::askConfirmation()` and these were the remaining questions which did not already use `yes` or `y`.

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -278,7 +278,7 @@ EOT
             && (
                 $input->getOption('remove-vcs')
                 || !$io->isInteractive()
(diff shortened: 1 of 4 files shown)


Seldaek:
The reason for the capital Y was to indicate which is the default if you just press enter IMO.. But I guess it is always yes so maybe it's best to merge this, especially if Y is actually invalid.

Seldaek:
Thanks!

PR #12508, opened 2025-08-21
Add Symfony 8 support

IonBazan:
Allows installing with Symfony 8 components (dev) and updates commands accordingly.

Breaking change:
`BaseCommand::initialize()` added return type: `void`

### DIFF
--- a/src/Composer/Command/BaseCommand.php
+++ b/src/Composer/Command/BaseCommand.php
@@ -218,7 +218,7 @@ abstract class BaseCommand extends Command
      *
      * @return void
      */
-    protected function initialize(InputInterface $input, OutputInterface $output)
+    protected function initialize(InputInterface $input, OutputInterface $output): void
     {
         // initialize a plugin-enabled Composer instance, either local or global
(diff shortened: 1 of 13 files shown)


Seldaek:
Thanks, I hope this won't cause issues for plugins or other third parties extending our commands.. but hopefully they've added return types by now.

src/Composer/Command/BaseCommand.php
```diff
@@ -218,7 +218,7 @@ public function complete(CompletionInput $input, CompletionSuggestions $suggesti
      *
      * @return void
```
GromNaN, line 219: You can now remove this duplicate annotation.

IonBazan:
I had to use some magic to get composer-diff to work: https://github.com/IonBazan/composer-diff/commit/cf17bbfab9f5b8b0abc4dec9b220c3f631540124 but this shouldn't be the case for most plugins as they most likely support PHP 7.2+

src/Composer/Command/BaseCommand.php
Seldaek: Fixed in #12509

PR #12509, opened 2025-08-21
Fix CS

Seldaek:


src/Composer/Command/BaseCommand.php
GromNaN, line 1: You should rebase the branch, this file is not up-to-date with https://github.com/composer/composer/commit/52f0df0eebf8ac9d11101f3f6f4f929e6519afb0

Seldaek: Yeah 🙈 forgot to pull latest changes.. Fixed

PR #12512, opened 2025-08-26
fixes a regression introduced via 304107d breaking bumping versions containing a zero

verfriemelt-dot-org:
the preg match was not updated to support major versions containing a zero :) like ~2025.1.x this should do the trick now. checkout the testcase for a proper example


Seldaek:
Thanks!

PR #12516, opened 2025-08-28
Reduce memory usage in PoolBuilder

Toflar:
Not something you would notice when running a regular `composer update` or so because there are other - much more important - factors that influence memory usage. However, if you use e.g. the `ComposerRepository` pretty much isolated like I do in https://github.com/terminal42/composer-lock-validator, the fact that we're keeping all provider data in memory for all the packages provided is quite noticeable - the more packages, the more memory.
This is because the promises cause `$contents` to never be cleaned up by GC as it is used until the promise is finally resolved.

Loading packages in batches instead of all at once prevents such potential memory allocations quite nicely:

Bildschirmfoto 2025-09-18 um 11 50 43

Bildschirmfoto 2025-09-18 um 11 47 24

So `json_decode()` still allocates `83 MB` in my case in total but because it does it in junks now, we reduce peak memory usage.

### DIFF
--- a/src/Composer/DependencyResolver/PoolBuilder.php
+++ b/src/Composer/DependencyResolver/PoolBuilder.php
@@ -38,6 +38,8 @@ use Composer\Semver\Intervals;
  */
 class PoolBuilder
 {
+    private const LOAD_BATCH_SIZE = 50;
+
     /**
      * @var int[]
      * @phpstan-var array<key-of<BasePackage::STABILITIES>, BasePackage::STABILITY_*>
@@ -420,7 +422,7 @@ class PoolBuilder
         }
 
         // Load packages in chunks of 50 to prevent memory usage build-up due to caches of all sorts
-        $packageBatches = array_chunk($this->packagesToLoad, 50, true);
+        $packageBatches = array_chunk($this->packagesToLoad, self::LOAD_BATCH_SIZE, true);
         $this->packagesToLoad = [];
 
         foreach ($repositories as $repoIndex => $repository) {
@@ -451,7 +453,7 @@ class PoolBuilder
                 }
             }
 
-            $packageBatches = array_chunk(array_merge(...$packageBatches), 50, true);
+            $packageBatches = array_chunk(array_merge(...$packageBatches), self::LOAD_BATCH_SIZE, true);
         }
     }
 


Toflar:
Right. It's already sort of not very efficient now, though. It reads the entire cache file just for the `last-modified` property. Could be done a lot more efficiently by storing meta data either in a file alongside the actual cache file - or using some JSON streaming stuff which as I wrote, might be hard to find re-usable libraries with our current minimum supported PHP version 😊

Seldaek:
So I tried storing cache files as `<?php return [...]` and include them instead of file_get_contents+json_decode.. But for some reason this is much slower.. It does save a bit of memory but this is not worth it really.

I don't fully grasp why this is.. But anyway this isn't helping so I'll investigate your patch closer.

Seldaek:
Interestingly I don't see a memory improvement, just an almost 2x slowdown with your patch.

I tried just reading the last-modified flag by doing a fopen+fseek to the end, and reading the last 80 chars to extract it, but that doesn't improve anything either..  Not sure what is going on today :D

Toflar:
Different approach: Simply chunk the loading of packages in the `PoolBuilder`. This also has the advantage that it covers all sorts of memory/cache build up cases within all kinds of repositories, not just the `ComposerRepository` specifically.
All it does is instead of loading e.g. 250 packages at once, we now chunk into batches of 50 packages giving the GC the opportunity to jump in and free memory.
This fixes my issue (from `100 MB` down to `47 MB`) and I couldn't see any negative effects on a general `composer update` - performance is not affected as far as I could tell. 

Diff best viewed with hiding whitespace - it's basically just one more loop.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -419,32 +419,37 @@ private function loadPackagesMarkedForLoading(Request $request, array $repositor
...
-                // avoid loading the same package again from other repositories once it has been found
-                unset($packageBatch[$name]);
-            }
-            foreach ($result['packages'] as $package) {
-                $this->loadedPerRepo[$repoIndex][$package->getName()][$package->getVersion()] = $package;
+                foreach ($result['namesFound'] as $name) {
+                    // avoid loading the same package again from other repositories once it has been found
+                    foreach ($packageBatches as &$v)  {
+                        unset($v[$name]);
+                    }
```
Seldaek: This could be a simpler unset if you just keep track of $batchIndex above right? Because the package cannot be in multiple batches as we split up one single array into multiple arrays.

```suggestion
                    unset($packageBatches[$batchIndex][$name]);
```

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -419,32 +419,37 @@ private function loadPackagesMarkedForLoading(Request $request, array $repositor
...
-            // never need to load anything else from them
-            if ($repository instanceof PlatformRepository || $repository === $request->getLockedRepository()) {
-                continue;
-            }
-            $result = $repository->loadPackages($packageBatch, $this->acceptableStabilities, $this->stabilityFlags, $this->loadedPerRepo[$repoIndex] ?? []);
+                // these repos have their packages fixed or locked if they need to be loaded so we
+                // never need to load anything else from them
+                if ($repository instanceof PlatformRepository || $repository === $request->getLockedRepository()) {
+                    continue;
+                }
```
Seldaek: This could be moved above the packageBatches foreach as it really relates only to the repository.

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: 🤦 of course, it was late already. Updated :)

src/Composer/DependencyResolver/PoolBuilder.php
Toflar: Right, updated too.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -419,32 +419,37 @@ private function loadPackagesMarkedForLoading(Request $request, array $repositor
...
         $this->packagesToLoad = [];
 
         foreach ($repositories as $repoIndex => $repository) {
-            if (0 === \count($packageBatch)) {
-                break;
-            }
+            foreach ($packageBatches as $packageBatch) {
+                if (0 === \count($packageBatch)) {
+                    break;
+                }
```
Seldaek: This needs to be moved below the packageBatches foreach probably.. or at least it needs to be revisited, because the point was to abort the repository loop when we're done with all packages.

I guess ideally you'd re-batch all packages that are left into a new set of 50-packages batches after every packageBatches foreach, so that every repo gets queried with 50 packages at once consistently.

```suggestion
                if (0 === \count($packageBatch)) {
                    break;
                }
```

Toflar: I can't wrap my head around a solution for that for now.

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -419,33 +419,39 @@ private function loadPackagesMarkedForLoading(Request $request, array $repositor
             $this->loadedPackages[$name] = $constraint;
         }
 
-        $packageBatch = $this->packagesToLoad;
+        // Load packages in chunks of 50 to prevent memory usage build-up due to caches of all sorts
+        $packageBatches = array_chunk($this->packagesToLoad, 50, true);
```
Toflar: @Seldaek should probably make `50` a variable or constant just so that we can prevent errors where one would update this line but forget to adjust line 454?

Seldaek:
Thanks

PR #12537, opened 2025-09-10
Fix `--global` flag path resolution for bin-dir config

kayw-geek:
## Summary

  Fixes incorrect path resolution when using composer global config to get bin-dir.

## How to Reproduce

  Current directory: `/Users/xx/Projects/dotfiles`
  Try to get global bin-dir `composer config -g bin-dir --absolute`

  Expected:  `~/.composer/vendor/bin`
  Actual: `/Users/xx/Projects/dotfiles/vendor/bin`

  ## Solution
  Set the config's baseDir to the home directory when --global flag is used, ensuring absolute paths are calculated correctly for global configuration.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -183,6 +183,11 @@ EOT
         $io = $this->getIO();
         $this->config = Factory::createConfig($io);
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, makes sense, merged as 8c02fa940 in 2.8 branch

PR #12543, opened 2025-09-13
Use sudo.exe with self-update when available

johnstevenson:
On Windows the self-update command requires admin priviledges (if installed by ComposerSetup) and currently uses VBScript to invoke the UAC prompt.

However, VBScript deprecation was announced in 2023 and is slated to be disabled by default in 2027 (but available as an optional Windows feature). It may possibly be removed forever at some time after that.

Sudo for Windows is a recent solution to do the same thing but is currently only available behind a Developer setting.

This PR will use sudo if it is available, falling back to using VBScript. If this is not available I would imagine that the OS will output a message explaining why, as it does if you call sudo without enabling it.

https://learn.microsoft.com/en-us/windows/advanced-settings/sudo/
https://techcommunity.microsoft.com/blog/windows-itpro-blog/vbscript-deprecation-timelines-and-next-steps/4148301

### DIFF
--- a/src/Composer/Command/SelfUpdateCommand.php
+++ b/src/Composer/Command/SelfUpdateCommand.php
@@ -602,7 +602,7 @@ TAGSPUBKEY
     /**
      * Invokes a UAC prompt to update composer.phar as an admin
      *
-     * Uses a .vbs script to elevate and run the cmd.exe copy command.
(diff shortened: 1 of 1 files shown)


fadaile:
serialization magic method has been deprecated. Implement __unserialize() instead (or in addition, if support for old PHP versions is necessary)

Seldaek:
Thanks!

PR #12545, opened 2025-09-18
Add update-with-minimal-changes config option

### LINKED ISSUE #12355: JSON configuration equivalent of `COMPOSER_MINIMAL_CHANGES` env var

For some legacy projects I tend to set `COMPOSER_MINIMAL_CHANGES=1` in my environment, to minimise the risk when updating certain packages, or installing new ones.

I can't currently see a way to make this a rule for all project users, other than via documentation.

A JSON configuration key that makes all commands behave as if `--minimal changes` flag is provided.

IonBazan:
Resolves #12355

Adds `update-with-minimal-changes` config option which can be set on both global and project level, allowing to minimize impact of `composer update`.

~~The test is exactly the same as `full-update-minimal-changes` - simply added config entry there.~~

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -505,7 +505,7 @@ Defaults to `false`. Ignores error during `install` if there are any missing
 requirements - the lock file is not up to date with the latest changes in
 `composer.json`.
 
-## minimal-changes
+## update-with-minimal-changes
 
 Defaults to `false`. If set to true, Composer will only perform absolutely necessary
 changes to transitive dependencies during update.
--- a/src/Composer/Command/RemoveCommand.php
+++ b/src/Composer/Command/RemoveCommand.php
@@ -258,7 +258,7 @@ EOT
         $authoritative = $input->getOption('classmap-authoritative') || $composer->getConfig()->get('classmap-authoritative');
         $apcuPrefix = $input->getOption('apcu-autoloader-prefix');
         $apcu = $apcuPrefix !== null || $input->getOption('apcu-autoloader') || $composer->getConfig()->get('apcu-autoloader');
-        $minimalChanges = $input->getOption('minimal-changes') || $composer->getConfig()->get('minimal-changes');
+        $minimalChanges = $input->getOption('minimal-changes') || $composer->getConfig()->get('update-with-minimal-changes');
 
         $updateAllowTransitiveDependencies = Request::UPDATE_LISTED_WITH_TRANSITIVE_DEPS_NO_ROOT_REQUIRE;
         $flags = '';
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -748,6 +748,10 @@
                 "allow-missing-requirements": {
                     "type": ["boolean"],
                     "description": "Defaults to false. If set to true, Composer will allow install when lock file is not up to date with the latest changes in composer.json."
+                },
+                "update-with-minimal-changes": {
+                    "type": ["boolean"],
+                    "description": "Defaults to false. If set to true, Composer will only perform absolutely necessary changes to transitive dependencies during update."
                 }
             }
         },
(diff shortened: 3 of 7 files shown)


doc/06-config.md
```diff
@@ -505,4 +505,10 @@ Defaults to `false`. Ignores error during `install` if there are any missing
 requirements - the lock file is not up to date with the latest changes in
 `composer.json`.
 
+## minimal-changes
```
IonBazan: Perhaps good to keep this in alphabetical order in the future. I'll keep the scope of my changes minimal (sic!) here though.

tests/Composer/Test/Fixtures/installer/full-update-minimal-changes-config.test
```diff
@@ -0,0 +1,65 @@
...
+                { "name": "dependency/pkg2", "version": "1.0.0" }
+            ]
+        }
+    ],
+    "require": {
+        "root/dep": "2.*",
+        "root/dep2": "1.*"
+    },
+    "config": {
+        "minimal-changes": true
```
IonBazan: I'm on the fence when it comes to naming here - let me know if `update-with-minimal-changes` would be better here.

src/Composer/Command/RemoveCommand.php
```diff
@@ -258,6 +258,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $authoritative = $input->getOption('classmap-authoritative') || $composer->getConfig()->get('classmap-authoritative');
         $apcuPrefix = $input->getOption('apcu-autoloader-prefix');
         $apcu = $apcuPrefix !== null || $input->getOption('apcu-autoloader') || $composer->getConfig()->get('apcu-autoloader');
+        $minimalChanges = $input->getOption('minimal-changes') || $composer->getConfig()->get('minimal-changes');
```
IonBazan: I noticed above changes are repeated in `update`, `remove` and `require`. I can move these to a common class/trait as a follow up PR.

Seldaek: I think it's a case where repeating is clearer than abstracting.

Seldaek: Adding this to the composer-schema.json should resolve the phpstan issue, as it will become aware of the expected datatype.

tests/Composer/Test/Fixtures/installer/full-update-minimal-changes-config.test
Seldaek: Yeah I tend to agree the naming is maybe not clear enough. I don't expect this to be set very often so maybe rather keep it long and explicit.

IonBazan:
Thanks for the review @Seldaek! I've updated the config key name and json schema. Unfortunately, I had to remove the test as it was not working - I realized it doesn't actually test the command but rather the underlying installer so the config was not applied there. I think a bigger change would be needed to test such scenario.

We could add an `InstallerFactory` which would create an instance of `Installer` from `InputInterface` and `Config`, resolving the parameters accordingly. Moving `Installer` instantiation would improve testability and reduce repetitions pointed out above and improve readability.

Seldaek:
Yep to support this in tests you'd need to update https://github.com/composer/composer/blob/main/tests/Composer/Test/InstallerTest.php#L420 too.

IonBazan:
`InstallerTest` creates their own simplified version of `update` command and mocks `Config` too. We could add support to read config from `.test` file and apply it to the mock but that still won't verify actual command behaviour.

Let me try to pick it up as an improvement in a separate PR, if that's ok 🙏🏻

Seldaek:
yeah please don't mix this in this PR for sure :D Not sure if much can be done about this.. I don't really mind it's not using the real command, because these .test files are about testing the Installer class and the solver internals in integration, the command class is irrelevant here.

Seldaek:
Thanks

Issue #11243, opened 2022-12-26
Set default config for namespace and author


Schrank:
I just created a composer.json file via `composer init` and was wondering how I can set default values for the package namespace  and the author. I read the documentation, but couldn't find anything about it. Obviously it is read somewhere, where? 😅  I would like to change it.

I see a ton of options here, but is anything of this set via global config? If yes I'm happy to add documentation, if not I'm happy to add a PR for it :-)
https://github.com/composer/composer/blob/main/src/Composer/Command/InitCommand.php#L272-L292

Same for the author:
https://github.com/composer/composer/blob/main/src/Composer/Command/InitCommand.php#L321-L337

So if I'm correct, it should be documented here:

https://getcomposer.org/doc/03-cli.md#environment-variables

and should work like the other env vars, correct? Then I would provide a PR with a couple of lines description and add `COMPOSER_DEFAULT_VENDOR` and `COMPOSER_DEFAULT_AUTHOR `.

```
-> % composer init

  Welcome to the Composer config generator

This command will guide you through creating your composer.json config.

Package name (<vendor>/<name>) [fabian/shopify-tools]: winkelwagen/shopify-tools
Description []:
Author [Fabian Blechschmidt <blechschmidt@fabian-blechschmidt.de>, n to skip]:
```

Seldaek:
Yes a PR for the docs would be great. I'd add these to 03-cli and ref them in the `init` docs as it is really specific to that command.

IonBazan:
How about having a global config which can store pretty much any default content for `composer.json`, including `require` and `require-dev` too. This would serve as a template for `init` runs.

```json
{
    "config": {
        "defaults": {
            "namespace": "MyApp\\",
            "vendor": "my-vendor",
            "authors": [
                {
                    "name": "Your Name",
                    "email": "test@example.com"
                }
            ],
            "license": "MIT",
            "require-dev": {
                "phpunit/phpunit": "^9.5"
            }
        }
    }
}

```

Seldaek:
Yeah I guess this would be ok, tho maybe would call this `init-defaults`. But do people really create so many projects that they need a template for their composer.json? :s

PR #12546, opened 2025-09-19
Allow non-interactive init without input

### LINKED ISSUE #12189: non interactive composer init without questions
With NPM you can do:

```
npm init -y
```

to skip all questions and just have a package.json.

This does not work with composer (there is obviously no `-y`)

```
❯ composer init -n

In InitCommand.php line 153:

  You have to run this command in interactive mode, or specify at least some data using --name, --require, etc.


init [--name NAME] [--description DESCRIPTION] [--author AUTHOR] [--type TYPE] [--homepage HOMEPAGE] [--require REQUIRE] [--require-dev REQUIRE-DEV] [-s|--stability STABILITY] [-l|--license LICENSE] [--repository REPOSITORY] [-a|--autoload AUTOLOAD]
```

When running it interactively it suggest a composer name, can't we just use that for non interactive

IonBazan:
Fixes #12189

Adds non-interactive no-argument `init` command support and refactor tests. `--name` and `--author` arguments default to previously suggested values from interactive mode (current working directory and git author).

It also makes it easier to implement #11243 in the future if needed.

Test has been refactored to keep it DRY - using data providers where possible.

### DIFF
--- a/tests/Composer/Test/Command/InitCommandTest.php
+++ b/tests/Composer/Test/Command/InitCommandTest.php
@@ -15,6 +15,7 @@ namespace Composer\Test\Command;
 use Composer\Command\InitCommand;
 use Composer\Json\JsonFile;
 use Composer\Test\TestCase;
+use InvalidArgumentException;
 
 class InitCommandTest extends TestCase
 {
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

PR #12553, opened 2025-09-24
Update recommendations on GitHub token usage, pushing users towards fine-grained tokens

mpdude:


mpdude:
Test failures seem to have other causes

mpdude:
Maybe we can additionally clarify that the _private_ repo use case only applies when Composer needs to load data for such repositories that are configured as `vcs` repos?

When private repos are tracked through e. g. Satis (or probably Private Packagist?) and cloned through SSH URLs, a public-only (no-scope) token should be enough, not?

Seldaek:
> Maybe we can additionally clarify that the _private_ repo use case only applies when Composer needs to load data for such repositories that are configured as `vcs` repos?
> 
> When private repos are tracked through e. g. Satis (or probably Private Packagist?) and cloned through SSH URLs, a public-only (no-scope) token should be enough, not?

Yes, if you use Private Packagist and just use zip/dists, you should not need any GitHub token at all locally.. For source installs then yeah ssh should be enough. If you want to edit feel free but otherwise I am happy to merge as is. Just let me know :)

mpdude:
So I could re-phrase this, more clearly stating that it is only relevant which kind of repositories is referenced in `vcs` entries for `repositories`?

Seldaek:
Hmm.. yes it mainly is relevant if you declare inline `vcs` repositories, or otherwise download heavily from github.

mpdude:
I have tried to re-phrase and clarify that. 

Please let me know what you think.

Seldaek:
I think it's pretty good as is, thanks!

mpdude:
Thank you

PR #12555, opened 2025-09-26
Mention "incompatible" in "conflict" docs

cweiske:
So that searching for "incompat" yields results on that page.

https://getcomposer.org/doc/04-schema.md#conflict


Seldaek:
Yeah also it's good to define conflict without using the word conflict 😅  Thanks

PR #12556, opened 2025-09-27
Bump all updated packages when running `composer update` with given packages

kubawerlos:
Packages `acme/foo` and `acme/bar` are in `composer.json`.

Currently running `composer update acme/bar --bump-after-update --with-all-dependencies` updates both packages, but only bumps `acme/bar`.


src/Composer/Command/UpdateCommand.php
```diff
@@ -295,7 +295,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                     $bumpAfterUpdate === 'dev',
                     $bumpAfterUpdate === 'no-dev',
                     $input->getOption('dry-run'),
-                    $input->getArgument('packages')
+                    []
```
Seldaek, line 298: That will bump EVERYTHING, not just updated packages.. Which I think is kinda surprising behavior if you are explicitly only updating some packages.

I however don't really see an easy way to get the list of updated packages, but you could perhaps get the list of dependents from the allowlisted packages, that should be doable to read from the $composer->getRepository(), and only do that if $updateAllowTransitiveDependencies allows it. That'll still be a bunch of code tho to do correctly.. Not sure if worth it.

Seldaek: Ok so had some more thoughts about this and I think the only clean way to do it would be to store $lockTransaction on a property at Installer.php:508 - then add a `getLockTransaction(): ?LockTransaction` getter so that you can fetch it from the UpdateCommand, and in there you can look at operations and make sure to add all packages that have been updated to the array here instead of passing an empty array. That shouldn't be too much code and will work accurately.

kubawerlos:
@Seldaek do you think this is not a good idea anymore, or have you closed it to not have it hanging? I planned to implement it "soon".

Seldaek:
Yeah feel free to reopen when you have time, I thought you lost interest and closed to clean up sorry

Issue #12568, opened 2025-10-09
[Audit] Allow ignoring of specific abandoned packages


tgauges:
**Is your feature request related to a problem? Please describe.**

I'd like my CI to fail when there are unreviewed abandoned packages.
I'd also like to review those packages and ignore the abandoned status of some of those packages.

**Describe the solution you'd like**

There is already an option to ignore IDs or CVEs during audit: <https://getcomposer.org/doc/06-config.md#ignore>
I'd like the same for abandoned packages, e.g.

```json
{
    "config": {
        "audit": {
            "ignore-abandoned": {
                "doctrine/annotations": "There is no security risk and there is a solution in the works already: <https://www.drupal.org/project/drupal/issues/3550917>"
            }
        }
    }
}
```

**Describe alternatives you've considered**

Currently, it is possible to ignore or only warn for all abandoned packages: <https://getcomposer.org/doc/06-config.md#abandoned>

But if this is set, I won't notice newly abandoned packages.

Seldaek:
Fixed by https://github.com/composer/composer/pull/12572

PR #12572, opened 2025-10-15
Audit: add option to ignore individually abandoned packages

glaubinix:
Implements https://github.com/composer/composer/issues/12568

Appears this was causing issue in Drupal because of `doctrine/annotations` being abandoned and people running `composer audit` as part of their builds with `"abandoned": "fail"` which now fails without an easy path to upgrade

### DIFF
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -170,14 +170,13 @@ class Auditor
             $ignoredPackageNames = $ignoreAbandoned;
         }
 
-        if (\count($ignoreAbandoned) === 0) {
-            return $packages;
+        $filter = null;
+        if (\count($ignoreAbandoned) !== 0) {
+            $filter = BasePackage::packageNamesToRegexp($ignoredPackageNames);
         }
 
-        $filter = BasePackage::packageNamesToRegexp($ignoredPackageNames);
-
         return array_filter($packages, static function (PackageInterface $pkg) use ($filter): bool {
-            return !Preg::isMatch($filter, $pkg->getName());
+            return $pkg instanceof CompletePackageInterface && $pkg->isAbandoned() && ($filter === null || !Preg::isMatch($filter, $pkg->getName()));
         });
     }
 


cruno91:
We just started adding `composer audit` as a check that will fail pipelines and prevent merges to find out we can't add abandoned packages as exceptions. Is there a way we can apply this patch via the `composer` CLI in our pipeline to allow for this temporarily?

glaubinix:
@cruno91 no there isn't.

res/composer-schema.json
```diff
@@ -447,6 +447,24 @@
...
                             "enum": ["ignore", "report", "fail"],
                             "description": "Whether abandoned packages should be ignored, reported as problems or cause an audit failure."
+                        },
+                        "ignore-abandoned": {
+                            "anyOf": [
+                                {
+                                    "type": "object",
+                                    "description": "A list of abandoned package names that are reported but let the audit command pass.",
+                                    "additionalProperties": {
+                                        "type": ["string", "string"]
```
Seldaek: I don't think you can define the key type as object keys are always string, so this is probably enough?

```suggestion
                                        "type": "string"
```

res/composer-schema.json
```diff
@@ -447,6 +447,24 @@
                         "abandoned": {
                             "enum": ["ignore", "report", "fail"],
                             "description": "Whether abandoned packages should be ignored, reported as problems or cause an audit failure."
+                        },
+                        "ignore-abandoned": {
+                            "anyOf": [
+                                {
+                                    "type": "object",
+                                    "description": "A list of abandoned package names that are reported but let the audit command pass.",
```
```suggestion
                                    "description": "A list of abandoned package names (as key) mapping to a reason explaining why they are ignored.",
```

src/Composer/Advisory/Auditor.php
```diff
@@ -156,12 +159,21 @@ public function audit(IOInterface $io, RepositorySet $repoSet, array $packages,
...
+            $ignoredPackageNames = array_keys($ignoreAbandoned);
+        } else {
+            $ignoredPackageNames = $ignoreAbandoned;
+        }
+
+        $filter = \count($ignoreAbandoned) > 0 ? BasePackage::packageNamesToRegexp($ignoredPackageNames) : null;
+
+        return array_filter($packages, static function (PackageInterface $pkg) use ($filter): bool {
+            return $pkg instanceof CompletePackageInterface && $pkg->isAbandoned() && ($filter === null || !Preg::isMatch($filter, $pkg->getName()));
         });
```
Seldaek: I don't think we really need to check for isAbandoned here? They really ought to be if they're passed in

```suggestion
        if (\count($ignoreAbandoned) === 0) {
            return $packages;
        }

        $filter = BasePackage::packageNamesToRegexp($ignoredPackageNames);

        return array_filter($packages, static function (PackageInterface $pkg) use ($filter): bool {
            return !Preg::isMatch($filter, $pkg->getName());
        });
```

glaubinix: Makes sense

glaubinix: Actually no, the abandoned check in there is what finds all the abandoned packages in the lock file. So this definitely needs to be there.

Seldaek: Ah sorry I totally ignored the fact that code was there before :D

Seldaek:
Thanks

PhilETaylor:
When running snapshot now Im getting this error - is this related?

```
In Auditor.php line 75:

  [TypeError]
  Composer\Advisory\Auditor::audit(): Argument #9 ($ignoreUnreachable) must be of type bool, array given, called in phar:///usr/local/bin/composer/src/Composer/Installer.php on line 436
```

Composer version 2.9-dev+028e057787298cc0881ecb706406802e4c1c7791 (2.9-dev) 2025-10-28 10:23:57
PHP version 8.4.14 (/opt/homebrew/Cellar/php/8.4.14/bin/php)

glaubinix:
Yes, see https://github.com/composer/composer/pull/12579

PhilETaylor:
Awesome, thanks - thought it best to mention it sooner rather than later :)

PR #12574, opened 2025-10-18
tests: add a test to evaluate the output stability

drupol:


Seldaek:
Closing per https://github.com/composer/composer/pull/12524#issuecomment-3452214794 - let's keep discussing over there if needed.

PR #12578, opened 2025-10-25
Add missing documentation for schema properties

TravisCarden:
There are [a couple of properties in `res/composer-schema.json`](https://github.com/composer/composer/blob/2.8.12/res/composer-schema.json#L1170-L1208) that are missing from the documentation at https://getcomposer.org/doc/04-schema.md. This PR adds them.

I have tried to match the structure and formatting of existing documentation. Obviously, let me know if you'd like changes.

Also, I noticed that a few other fields for which I could not find corresponding documentation--[`security-advisories`](https://github.com/composer/composer/blob/2.8.12/res/composer-repository-schema.json#L35-L51), [`metadata-changes-url`](https://github.com/composer/composer/blob/2.8.12/res/composer-repository-schema.json#L52-L55), [`warnings`](https://github.com/composer/composer/blob/2.8.12/res/composer-repository-schema.json#L72-L88), and [`infos`](https://github.com/composer/composer/blob/2.8.12/res/composer-repository-schema.json#L89-L105). I didn't attempt to add them in this PR because I wanted to get your input first. I'm happy to add them according to your direction, if you like.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -335,6 +335,37 @@ Example:
 
 Optional.
 
+### dist
+
+The distribution archive for the package.
+
+Properties:
+
+* **type:** The type of distribution archive (e.g. `zip`, `tar`, `phar`).
+* **url:** The URL to download the distribution archive.
+* **reference:** The specific commit, tag, or revision that this distribution represents.
+* **shasum:** The SHA-1 checksum of the distribution archive for verification.
+* **mirrors:** An array of alternative download URLs.
+
+Example:
+
+```json
+{
+    "dist": {
+        "type": "zip",
+        "url": "https://api.github.com/repos/Seldaek/monolog/zipball/10d85740180ecba7896c87e06a166e0c95a0e3b6",
+        "reference": "10d85740180ecba7896c87e06a166e0c95a0e3b6",
+        "shasum": "a1b8d00bb5bc3b0308e99a95e01bdbbefbe052b4",
+        "mirrors": [
+            "https://api.github.com/repos/alternative-one/monolog/zipball/10d85740180ecba7896c87e06a166e0c95a0e3b6",
+            "https://api.github.com/repos/alternative-two/monolog/zipball/10d85740180ecba7896c87e06a166e0c95a0e3b6"
+        ]
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


Seldaek:
Yeah source/dist are mostly for internal use (or inline packages, where we do have some explanation of this I believe) and as such not described there to avoid confusion.

All the other fields you mentioned are part of repository metadata and not package metadata / composer.json, so that's something else entirely and should not be documented on that page.

So I'm not sure if we should merge this or not tbh.. @naderman any opinion here?

naderman:
I think I still worry this will get used too much / too widely if features like this in the docs. It's a practice we generally discourage that can be useful in some unique edge cases. At the same time I also dislike that it's not documented at all, it does have some security related implications so maybe better to document that this is in fact possible. So if we add it, should have much more of a warning that we strongly discourage using these, they are meant for internal composer use and they are only documented for that purpose? But we all know people don't read warnings :D

stof:
`dist` might still be usable for projects that have custom hosting for which composer does not have a dedicated driver and so won't be able to automatically discover the dist endpoint for that package when using the basic GitDriver.
That's why we support `%reference%` placeholders in the dist URL of a package to automatically replace the reference (and also `%version%`). And that's also why we automatically copy the source reference as dist reference when we have a dist config without reference and we have a source reference (as GiftDriver will be able to determine the source reference).

For packages hosted on GitHub (or other hosting providers that are supported natively), specifying `dist` is indeed useless as it is guessed automatically.

Seldaek:
Yeah I am also kinda thinking this will just highlight a feature more people will end up misusing.. So probably rather close. If anything this could be documented on a new page documenting the composer repository format in more details. I don't think we have that documented anywhere atm except in the 2.0 upgrade notes partially. If you feel like tackling this @TravisCarden feel free to open a new PR with this content moved there + the other fields you mentioned, but tbh I am not sure it's worth anyone's time given how few people end up implementing repositories (they usually end up looking at packagist.org as a reference impl, or talk to us for the most complex cases)

TravisCarden:
Cool, @Seldaek. If additional documentation pages aren't important (or desirable), I won't spend my time on them. Thanks!

PR #12580, opened 2025-10-28
docs: clarify binary deps

pereorga:
This addresses points 1 and 2 in https://github.com/composer/composer/discussions/12278.

I've listed `unzip` before `7z`, since `unzip` is checked first on most platforms. It's also lighter, so it should be preferable.

I've simplified the ext-zlib description: I imagine "Enables gzip compression/decompression for HTTP requests" would be technically accurate, but I think it's too verbose and redundant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ should be automatically selected.
 
 The need for these binary dependencies may vary depending on individual use cases. For most users,
 only 2 dependencies are essential for Composer: `unzip` (or `7z`/`7zz`), and `git`. If the
-[ext-zip](https://www.php.net/manual/en/zip.installation.php) extension is available, only `git`
+[`ext-zip`](https://www.php.net/manual/en/zip.installation.php) extension is available, only `git`
 is needed, but this is not recommended.
 
 Authors


README.md
```diff
@@ -52,8 +52,8 @@ should be automatically selected.
 
 #### Binary dependencies
 
-- `7z` (or `7zz`)
-- `unzip` (if `7z` is missing)
+- `unzip`
+- `7z` (or `7zz`), if `unzip` is missing
```
stof: Should we list those as `unzip, 7z or 7zz` instead, by grouping together the libraries related to support Zip archives ?

README.md
```diff
@@ -64,9 +64,10 @@ should be automatically selected.
 - Perforce (`p4`)
 - Subversion (`svn`)
 
-It's important to note that the need for these binary dependencies may vary
-depending on individual use cases. However, for most users, only 2 dependencies
-are essential for Composer: `7z` (or `7zz` or `unzip`), and `git`.
+The need for these binary dependencies may vary depending on individual use cases. For most users,
+only 2 dependencies are essential for Composer: `unzip` (or `7z`/`7zz`), and `git`. If the
+[ext-zip](https://www.php.net/manual/en/zip.installation.php) extension is available, only `git`
```
stof:
```suggestion
[`ext-zip`](https://www.php.net/manual/en/zip.installation.php) extension is available, only `git`
```

README.md
pereorga: Done

Seldaek:
Thanks

PR #12582, opened 2025-10-29
Add support for cache-busting

soatok:
See https://github.com/composer/packagist/issues/1606

Sometimes Packagist is too aggressive wit hits caching and we need to bypass it in a pinch. This adds a `--cache-bust` flag for this purpose.

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -63,6 +63,8 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
     private $io;
     /** @var HttpDownloader */
     private $httpDownloader;
+    /** @var Config */
+    private $config;
     /** @var Loop */
     private $loop;
     /** @var Cache */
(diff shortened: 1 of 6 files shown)


soatok:
Hmm. I've never tried to modify the Composer codebase before. I'm not sure what PHPStan is complaining about exactly?

src/Composer/Repository/ComposerRepository.php
```diff
@@ -806,6 +809,18 @@ private function getProviderNames(): array
         return [];
     }
 
+    protected function addCacheBust(string $url): string
+    {
+        if ($this->config->has('cache-bust')) {
+            $bust = $this->config->get('cache-bust');
+            if (!empty($bust) && strpos($url, 'cachebust') === false) {
```
gabrielrbarbosa, line 816: phpstan is complaining about empty usage here, change it or run maybe you can run
`vendor/bin/phpstan analyse --generate-baseline`

Seldaek:
Thanks but I don't think this is needed, or productive really.. We encounter problems sometimes, but bypassing caches and adding load on our servers does not achieve anything for the broader community.

PR #12585, opened 2025-10-30
Add COMPOSER_PREFER_DEV_OVER_PRERELEASE env var

nicolas-grekas:
After a brainstorming session with @Seldaek, this is #12581 again, but behind an env var.

From the added doc:

###  COMPOSER_PREFER_DEV_OVER_PRERELEASE

If set to `1`, when resolving dependencies with both `--prefer-stable` and
`--prefer-lowest` enabled, dev versions are treated as more stable than
alpha/beta/RC versions in cases where no stable release exists. This is useful
to test lowest versions while still preferring branches that may contain
critical fixes over pre-release tags.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1324,6 +1324,14 @@ If set to `1`, it is the equivalent of passing the `--prefer-stable` option to
 If set to `1`, it is the equivalent of passing the `--prefer-lowest` option to
 `update` or `require`.
 
+### COMPOSER_PREFER_DEV_OVER_PRERELEASE
+
+If set to `1`, when resolving dependencies with both `--prefer-stable` and
+`--prefer-lowest` enabled, dev versions are treated as more stable than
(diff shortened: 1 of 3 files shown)


doc/03-cli.md
```diff
@@ -1324,6 +1324,14 @@ If set to `1`, it is the equivalent of passing the `--prefer-stable` option to
 If set to `1`, it is the equivalent of passing the `--prefer-lowest` option to
 `update` or `require`.
 
+### COMPOSER_PREFER_DEV_OVER_UNSTABLE
```
Seldaek: Sorry but I'm thinking prefer dev over *prerelease* might be less confusing. Because dev is unstable too so it sounds weird rn. Wdyt?

nicolas-grekas: so be it; PR updated

Seldaek:
Thanks

nicolas-grekas:
Cool! Thanks to the both of you!

PR #12588, opened 2025-11-01
Only allow name attribute for list representation of repositories

JoshuaBehrens:
Follow up to #12388 . It ensures, that the new and unreleased property name can only be used for the list representation to prevent ambiguous configurations.

grafik

@Seldaek @dzuelke I do not know if I am thinking to simple about it, but when the schema does not even allow it, it cannot go wrong anywhere. Neither IDE nor composer would work with it. And as it is about a new property, that is not yet released, we can introduce new rules around it.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -194,7 +194,7 @@
             "description": "A set of additional repositories where packages can be found.",
             "additionalProperties": {
                 "anyOf": [
-                    { "$ref": "#/definitions/repository" },
+                    { "$ref": "#/definitions/anonymous-repository" },
                     { "type": "boolean", "enum": [false] }
                 ]
             },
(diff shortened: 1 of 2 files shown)


dzuelke:
I think that's reasonable, @JoshuaBehrens, but I'll let @Seldaek confirm :) Thank you so much for the quick follow-up!

One thing, though: if I understand all the test changes from #12388 correctly, you changed a bunch of the example fixtures that use the object/assoc `repositories` notation to the new name notation, and I think it would be a good idea to bring those back, since AFAICT the object notation is currently untested for the regular "load a config" command.

There are a bunch of tests for the object/assoc notation for `JsonManipulator` etc. already, I think it's really just `JsonConfigSourceTest.php‎` that is currently lacking this little bit of coverage by having one or two cases.

But the PR diffs are fairly large and I have lost my train of thought a few times trying to figure out if I am just missing the fact that it's already there, so please forgive me if it's an oversight on my part.

Seldaek:
thanks - this is totally reasonable yes. @JoshuaBehrens any thoughts about the tests?

JoshuaBehrens:
@Seldaek I am still looking into what I have to replicate in the tests and if thre is something broken. I can already tell, that config and repo command behave similar so if this is something, that we do not want to, I have to re-add some code from before as well

---

Thank you for merging :)

Seldaek:
I think @dzuelke's point was just that we should ensure we don't break BC by keeping the old tests in place, but if it's too much work it may not be worth it.. Just want to keep things smooth if possible.

Issue #11204, opened 2022-11-29
Support multiple Gitlab Tokens for the same instance


AdrienPoupa:
I am using the Composer repository of a private GitLab instance. I setup a deploy token for group A, so that all projects under A can be accessed, and it works fine. I setup my composer.json and auth.json as described in the GitLab documentation: https://docs.gitlab.com/ee/user/packages/composer_repository/#working-with-deploy-tokens

composer.json
```json
...
  "config": {
    "gitlab-domains": ["gitlab.mydomain.ca"],
  },
...
  "repositories": {
    "A_ID": {
      "type": "composer",
      "url": "https://gitlab.turbulent.ca/api/v4/group/A_ID/-/packages/composer/",
      "only": ["a/*"]
    },
...
```

auth.json
```json
{
    "gitlab-token": {
        "gitlab.mydomain.ca": {
            "username": "deploy-token-group-A",
            "token": "redacted"
        }
    }
}
```

Now, I want to access all packages from group B. I created a new deploy token for group B, and I did this:

composer.json
```json
...
  "config": {
    "gitlab-domains": ["gitlab.mydomain.ca"],
  },
...
  "repositories": {
    "A_ID": {
      "type": "composer",
      "url": "https://gitlab.mydomain.ca/api/v4/group/A_ID/-/packages/composer/",
      "only": ["a/*"]
    },
    "B_ID": {
      "type": "composer",
      "url": "https://gitlab.mydomain.ca/api/v4/group/B_ID/-/packages/composer/",
      "only": ["b/*"]
    },
...
```

But it seems I cannot specify multiple deploy tokens **for the same GitLab instance**.

Ideally I would be able to do something like this, but it seems Composer will only read one token.

auth.json
```json
{
    "gitlab-token": {
        "gitlab.mydomain.ca": [
          {
            "username": "deploy-token-group-A",
            "token": "redacted"
          },
          {
            "username": "deploy-token-group-B",
            "token": "redacted"
          }
        ]
    }
}
```

I guess we could either
- try several tokens as shown above
- maybe specify the group ID in auth.json?

```json
{
    "gitlab-token": {
        "gitlab.mydomain.ca/A_ID": {
            "username": "deploy-token-group-A",
            "token": "redacted"
        },
       "gitlab.mydomain.ca/B_ID": {
            "username": "deploy-token-group-B",
            "token": "redacted"
        }
    }
}
```

Seldaek:
[...]

PR #12589, opened 2025-11-02
Featture: #11746: Allow environment variables to be used in json files

### LINKED ISSUE #11746: Allow environment variables to be used in auth.json
If I need to rotate my GitHub token or Private Packagist/Artifactory authentication, I would like to do it in one place. NPM supports using environment variables in the `.npmrc` file but I do not see that Composer also supports it in auth.json files. I propose that we add environment variable support so that I could do the following:

```json
{
    "github-oauth": {
        "github.com": "${GITHUB_TOKEN}"
    },
    "http-basic": {
        "packages.mycompany.com": {
            "username": "first.last@mycompany.com",
            "password": "${PACKAGIST_MY_COMPANY_PASSWORD}"
        }
    }
}
```

This would allow me to re-use the same variables across Composer and NPM since we use Artifactory for both at the current project I work at, and I only have to update them in once place (my `.zshenv` file).

Matthiasklaasse:
- ✨ Added JsonEnvParser so JsonFile can resolve ${VAR} placeholders from the project .env.
- 🔄 Preserved backward compatibility: missing env values trigger a warning but return the literal placeholder.
- 🌍 Every .json file (processed by the JsonFile class) is affected and can now take enviroment variables (even 3th party composer.json's).

Fixes #11746

### DIFF
--- a/src/Composer/Json/JsonEnvParser.php
+++ b/src/Composer/Json/JsonEnvParser.php
@@ -66,6 +66,15 @@ class JsonEnvParser
 	 */
 	private function resolvePlaceholder(string $name, string $placeholder, ?string $file = null): string
 	{
+		if (self::$dotEnvPath === null) {
+			if ($this->io instanceof IOInterface && self::$unreadableEnvWarned === false) {
+				$this->io->warning('Dotenv file path is not set.');
+				self::$unreadableEnvWarned = true;
+			}
+
+			return $placeholder;
+		}
+
 		if (!is_readable(self::$dotEnvPath)) {
 			if ($this->io instanceof IOInterface && self::$unreadableEnvWarned === false) {
 				$this->io->warning('Dotenv file ' . self::$dotEnvPath . ' is not readable.');


Matthiasklaasse:
The original ticket asked for environment variables only in auth.json but implementing that would be more work then implementing it for all .json files what do you (moderators) think of this?

bin/composer
```diff
@@ -108,6 +109,12 @@ Platform::putEnv('COMPOSER_BINARY', realpath($_SERVER['argv'][0]));
 
 ErrorHandler::register();
 
+$rootPath = Platform::getCwd();
+if (file_exists($rootPath . '/.env')) {
+    $dotenv = new Dotenv();
+    $dotenv->load(__DIR__.'/.env');
```
Seldaek:
```suggestion
    $dotenv->load($rootPath . '/.env');
```

composer.json
```diff
@@ -43,7 +43,8 @@
         "symfony/polyfill-php73": "^1.24",
         "symfony/polyfill-php80": "^1.24",
         "symfony/polyfill-php81": "^1.24",
-        "seld/signal-handler": "^2.0"
+        "seld/signal-handler": "^2.0",
+        "symfony/dotenv": "^5.4"
```
Seldaek:
```suggestion
        "symfony/dotenv": "^5.4 || ^6.4.25 || ^7.1.10 || ^8.0"
```

src/Composer/Json/JsonFile.php
```diff
@@ -147,20 +159,20 @@ public function write(array $hash, int $options = JSON_UNESCAPED_SLASHES | JSON_
...
             if (file_exists($dir)) {
                 throw new \UnexpectedValueException(
-                    realpath($dir).' exists and is not a directory.'
+                    realpath($dir) . ' exists and is not a directory.'
                 );
             }
             if (!@mkdir($dir, 0777, true)) {
                 throw new \UnexpectedValueException(
-                    $dir.' does not exist and could not be created.'
+                    $dir . ' does not exist and could not be created.'
```
Seldaek, line 160: It'd be great to avoid doing random CS changes together with your PR.. that distracts from the real content.

src/Composer/Json/JsonEnvParser.php
```diff
@@ -0,0 +1,88 @@
...
+
+			if (is_string($value)) {
+				$result[$key] = $this->replacePlaceholders($value, $file);
+				continue;
+			}
+
+			$result[$key] = $value;
+		}
+
+		return $result;
```
Seldaek: I think here it's actually justifiable to use references to avoid recreating the whole array structure for nothing..

```suggestion
		foreach ($data as $key => &$value) {
			if (is_array($value)) {
				$value = $this->apply($value, $file);
				continue;
			}

			if (is_string($value)) {
				$value = $this->replacePlaceholders($value, $file);
				continue;
			}
		}

		return $data;
```

src/Composer/Json/JsonFile.php
```diff
@@ -346,6 +362,13 @@ public static function parseJson(?string $json, ?string $file = null)
             self::validateSyntax($json, $file);
         }
 
+        if (!is_array($data)) {
+            return $data;
+        }
+
+
+        $data = self::$envParser->apply($data, $file);
+
```
Seldaek: I think this should be optional and not applied to all json files being parsed (including those coming from remote network right now). IMO the best would be that it can be turned on where it makes sense (auth.json, maybe composer.json, I don't think anything else needs this?) by passing a JsonEnvParser into the constructor.. and if none is passed then no env parsing happens. I don't think `parseJson` needs this ever?

Seldaek: Sorry thinking about this some more.. I think we really should only apply this to the auth.json - which could be done in Factory.php as that's the only place where we really read this.

And it should not use symfony/dotenv actually I think this should be reserved to real env vars.. Because if you're gonna dump stuff on disk then you might as well have it in the auth.json directly, so I'd rather save us some complexity here and not risk loading things we aren't supposed to from `.env` files.

Matthiasklaasse:
I’ve updated the code with the functionality you requested. Let me know if this is what you had in mind.

src/Composer/Json/JsonEnvParser.php
```diff
@@ -0,0 +1,149 @@
...
+
+	/**
+	 * @return string Resolves a single placeholder to its environment value part of funtion replacePlaceholders.
+	 */
+	private function resolvePlaceholder(string $name, string $placeholder, ?string $file = null): string
+	{
+		if (self::$dotEnvPath === null) {
+			if ($this->io instanceof IOInterface && self::$unreadableEnvWarned === false) {
+				$this->io->warning('Dotenv file path is not set.');
+				self::$unreadableEnvWarned = true;
```
Seldaek, line 72: I don't see why that should be a warning. And IMO if you want to add .env support it should probably be reading from `.env.local` => `.env` as fallback, but then reimplementing all of dotenv libs makes no sense.

Why support dotenv at all? Why not use real env vars?

naderman:
- Agree that only auth.json replacements should happen. Especially not composer.json, which could lead people to do things like requiring an environment defined package name or version, which would make composer.json way too dynamic and ambiguous especially for any kind of tooling operating on it.

- Like Jordi I'm not sure I see the value in supporting .env files here, then you may as well just write an auth.json file to disk? To agree to adding that, I would need to see an actually clear use-case that seems relevant enough to affect a lot of users positively.

src/Composer/Json/JsonFile.php
mbrodala: Notice that it would make a lot of sense to also have this for `composer.json` given that one currently cannot have multiple authentication details for a single domain.

For example you could have multiple `vcs` repositories on `gitlab.com`, each with its own auth token. Currently you need to literally put the tokens in the `composer.json` to achieve this, which is obviously not an option. The Composer `auth.json` only supports one token per domain: https://github.com/composer/composer/issues/11204

If Composer would resolve environment variables in the `composer.json` (and thus `repositories`), the repositories could be left in `composer.json` but the tokens extracted to environment variables.

---

Just saw what @naderman wrote on this and I get what he means. But we should still consider allowing environment variables at least for `repositories` which can contain embedded auth details.

Issue #12592, opened 2025-11-05
Add a platform requirement list


SiebeVE:
**Is your feature request related to a problem? Please describe.**
Composer json/lock contains a lot of information about what the platform should be. If we would be able to fetch this data according to the json/lock files we would be able to integrate tooling that setup an dev/prod environment.

**Describe the solution you'd like**
A new command like the check-platform-reqs, but that doesn't check it with the current platform. Because we can use it to create another environment we aren't interested in the one running on it.

**Describe alternatives you've considered**
Using the check-platforms-reqs won't give me the php constraints of what it needs to be, only the constraint it might conflict with.

PR #12593, opened 2025-11-05
#12592: Added initial POC for the list platform reqs command

SiebeVE:
This MR can be a start for the adding of the list of the platform reqs command https://github.com/composer/composer/issues/12592.
This can be improved because it uses some shared code between the check-platform-reqs.

First wanted to have some feedback on the feasability of the proposal

### DIFF
--- a/src/Composer/Command/ListPlatformReqsCommand.php
+++ b/src/Composer/Command/ListPlatformReqsCommand.php
@@ -1,0 +1,100 @@
+<?php declare(strict_types=1);
+
+namespace Composer\Command;
+
+use Composer\Package\Link;
+use Symfony\Component\Console\Input\InputInterface;
+use Composer\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Composer\Repository\PlatformRepository;
+use Composer\Repository\RootPackageRepository;
+use Composer\Repository\InstalledRepository;
+use Composer\Json\JsonFile;
+
+class ListPlatformReqsCommand extends BaseCommand
+{
+    protected function configure(): void
+    {
+        $this->setName('list-platform-reqs')
+            ->setDescription('List all platform requirements (PHP and extensions) for installed packages')
+            ->setDefinition([
+                new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables listing of require-dev packages requirements.'),
+                new InputOption('lock', null, InputOption::VALUE_NONE, 'Lists requirements only from the lock file, not from installed packages.'),
+                new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the output: text or json', 'text', ['json', 'text']),
+            ])
+            ->setHelp(
+                <<<EOT
+Lists all platform requirements (PHP and extensions) for the installed packages, without checking if they are satisfied.
+
+<info>php composer.phar list-platform-reqs</info>
+EOT
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $composer = $this->requireComposer();
+
+        $requires = [];
+        $removePackages = [];
+        if ($input->getOption('lock')) {
+            $installedRepo = $composer->getLocker()->getLockedRepository(! $input->getOption('no-dev'));
+        } else {
+            $installedRepo = $composer->getRepositoryManager()->getLocalRepository();
+            if (! $installedRepo->getPackages()) {
+                $installedRepo = $composer->getLocker()->getLockedRepository(! $input->getOption('no-dev'));
+            } else {
+                if ($input->getOption('no-dev')) {
+                    $removePackages = $installedRepo->getDevPackageNames();
+                }
+            }
+        }
+        if (! $input->getOption('no-dev')) {
+            foreach ($composer->getPackage()->getDevRequires() as $require => $link) {
+                $requires[$require] = [$link];
+            }
+        }
+
+        $installedRepo = new InstalledRepository([$installedRepo, new RootPackageRepository(clone $composer->getPackage())]);
+        foreach ($installedRepo->getPackages() as $package) {
+            if (in_array($package->getName(), $removePackages, true)) {
+                continue;
+            }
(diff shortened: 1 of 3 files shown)


Seldaek:
cc @dzuelke any thoughts about this? As I know you've done a lot in that area, and it's not clear to me what is most sensible.

Seldaek:
@SiebeVE one comment I'd have is I think I probably would rather have this as a `--list=[format]` option on the check-platform-reqs command, to refrain from adding a command for something which is a very niche use case.

dzuelke:
This will not work the way you think because of the precedence of the `|` operator you are using to combine requirements (which really should be `||` instead since `|` is deprecated).

Let's say you have two packages with the following requirements for `php`:
- package 1 requires `7.*`
- package 2 requires `7.4.* || ^8.0`

But the resulting `7.* || 7.4.* || ^8.0` would now permit 8.0 or later, which package 1 is incompatible with.

However, if you think about it, you want **all** the requirements to be fulfilled, so that all packages can be installed successfully later. But you also cannot use logical ANDs (via ` ` or `,`) to combine them, because that operator has a higher precedence, so the resulting `7.*, 7.4.* || ^8.0` would mean `(7.*, 7.4.*) || ^8.0`, which again would allow 8.0 or later, but what you want is `7.*, (7.4.* || ^8.0)`, which you cannot express because Composer does not allow parentheses in requirement specifications.

There are also other issues. For instance, you may have requirements for several `php` versions, and let's assume that these ultimately all allow `8.*`. You also have a requirement for `ext-newrelic` in any version. Now PHP 8.5 is released, but the New Relic team hasn't updated the extension for that series yet - you only have extension builds for PHP 8.4 and earlier.

Also don't forget `provide` and `replace`. Someone is going to have a requirement for e.g. `ext-mcrypt`, which no longer exists, but there are polyfills for it that `provide` it, so you have to take that into account, as well.

In order to really know what to install, you have to resolve all the version requirements for all the packages using the same solver logic rules that Composer uses during userland installs, and you have to do that against a repository of packages that represent all the platform packages you can install so you have the necessary metadata (PHP, bundled extensions and whether or not they are built shared, third-party extensions, and technically even libraries).

The metadata for the platform packages is really crucial because you need the info I mentioned earlier (e.g. what PHP versions an extension is compatible with), but that would basically be entirely up to the user, so Composer cannot reasonably provide any kind of hook or plugin method for helping with this (because user A will want to build their own PHP packages, user B will want to use OS distro builds, etc).

I solved this problem a long time ago for the [Heroku PHP Buildpack](http://github.com/heroku/heroku-buildpack-php/) by translating all `composer.lock` platform requirements into a dependency graph of userland packages that represent the platform packages, and a repository of platform packages, that then get installed using a custom Composer plugin (so it basically uses PHP and Composer to install PHP and Composer and extensions).

You could always use the newer [Heroku PHP Cloud Native Buildpack](http://github.com/heroku/buildpacks-php/), it figures all this stuff out and builds an OCI image you can run wherever you like ;)

SiebeVE:
Thanks for the thorough explenation!

I started this MR indeed because I want to build something simular to the buildpacks, but without the buildpacks. We have issues there that makes it anyoning for us to use.
(the php in the buildpacks (/layers/heroku_php/platform/bin/php -i) is built with the `--with-config-file-scan-dir=/app/.heroku/php/etc/php/conf.d` and that `/app` directory doesn't exists. (https://github.com/heroku/heroku-buildpack-php/blob/main/support/build/php#L101), so any cli workers don't automatically have the correct extensions. But because I don't know any rust/go myself, I'm not really able to assist in that and improve it)

I see the php list is indeed wrongfully concated, and then should an array so the user can check them itself.

The extensions part is a rabit hole I didn't expect :o, but intereseting stuff!

I'm gonna close this one and the issue, especially for the extensions part. That is indeed way more complexer and don't think the complexity outways the niche usecase for it.

Thanks for the feedback!

dzuelke:
@SiebeVE the INI scan dir gets overridden with the right value when it is installed in the buildpack by the installer plugin (plus another couple of post-install "relocation" operations, such as fixing `php-config` to return the correct paths etc), so it should work fine for a CLI worker - the values for `PHPRC` and `PHP_INI_SCAN_DIR` are correct at runtime:

```ShellSession
dzuelke@localhost:~$ docker run --rm -ti --entrypoint launcher php-cnb-hello-world2 php --ini
Configuration File (php.ini) Path: /app/.heroku/php/etc/php
Loaded Configuration File:         /layers/heroku_php/platform/etc/php/php.ini
Scan for additional .ini files in: /layers/heroku_php/platform/etc/php/conf.d
Additional .ini files parsed:      /layers/heroku_php/platform/etc/php/conf.d/000-heroku.ini,
/layers/heroku_php/platform/etc/php/conf.d/010-ext-zend_opcache.ini,
/layers/heroku_php/platform/etc/php/conf.d/100-ext-gmp.ini,
/layers/heroku_php/platform/etc/php/conf.d/110-ext-imap.ini,
/layers/heroku_php/platform/etc/php/conf.d/120-ext-intl.ini,
/layers/heroku_php/platform/etc/php/conf.d/130-ext-ldap.ini,
/layers/heroku_php/platform/etc/php/conf.d/140-ext-raphf.ini,
/layers/heroku_php/platform/etc/php/conf.d/150-ext-pq.ini,
/layers/heroku_php/platform/etc/php/conf.d/160-ext-redis.ini,
/layers/heroku_php/platform/etc/php/conf.d/170-ext-sqlite3.ini,
/layers/heroku_php/platform/etc/php/conf.d/180-ext-mbstring.ini
```


If you're trying to build something similar to the buildpacks but without the buildpacks, then maybe a custom PHP build is the better choice? ;)

We're getting a little off-topic over here though. Maybe file an issue [on the PHP CNB repo](https://github.com/heroku/buildpacks-php) with a reproduce case, or start a discussion to figure out how we can improve the buildpack experience to better fit your use case. The idea is to support local dev use, as well as CI and production - on  Heroku or elsewhere (that's the point of OCI images and the entire `pack`/CNB ecosystem).

Issue #12602, opened 2025-11-13
Composer 2.9 breaks post install scripts if you have pre install scripts


driskell:
There is an issue with PR https://github.com/composer/composer/pull/12456 that is released in 2.9

It looks like when preinstall scripts run, it generates the autoloader, but it remembers the hash. Then when it does a post install it uses the autoloader from BEFORE the install, so you get class not found errors.

In my case it breaks every one of our GitHub actions for Drupal installs as there is a preinstall and then a postinstall and it can't find `DrupalFinder\DrupalFinder` so you get `Class "DrupalFinder\DrupalFinder" not found`

I've managed to reproduce it here:

My `composer.json`:

```json
{
    "name": "project/test",
    "version": "1.0.0",
    "autoload": {
        "classmap": [
            "TestScript.php"
        ],
        "psr-4": {
            "Project\\Test\\": "src/"
        }
    },
    "scripts": {
        "pre-install-cmd": [
            "TestProject\\composer\\TestScript::preinstall"
        ],
        "post-install-cmd": [
            "TestProject\\composer\\TestScript::postinstall"
        ]
    },
    "authors": [
        {
            "name": "Jason Woods",
            "email": "devel@jasonwoods.me.uk"
        }
    ],
    "require": {
        "symfony/filesystem": "^7.3",
        "webflo/drupal-finder": "^1.3"
    }
}
```

`TestScript.php`:

```php
<?php

namespace TestProject\composer;

use DrupalFinder\DrupalFinder;
use Symfony\Component\Filesystem\Filesystem;

class TestScript {
  public static function preinstall() {
    echo 'Preinstall', PHP_EOL;
  }
  public static function postinstall() {
    $fs = new Filesystem();
    $finder = new DrupalFinder();
    echo 'Postinstall', PHP_EOL;
  }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
The version field is present, it is recommended to leave it out if the package is published on Packagist.
Checking composer.lock: OK
Checking platform settings: OK
Checking git settings: OK git version 2.51.0
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK does not expire
Checking disk free space: OK
Checking pubkeys:
[...]

PR #12606, opened 2025-11-13
Fix script autoloading regressions

Seldaek:
Addresses #12602


driskell:
Thanks @Seldaek ! Nice fix.

I can confirm it fixes my issue

PR #12609, opened 2025-11-13
Add config command support for missing audit settings

Seldaek:
Looks like we missed this somehow :D


Seldaek:
Related PR with some CLI / env flags to be able to disable blocking BTW, for anyone that ends up here  it may be relevant https://github.com/composer/composer/pull/12617

Issue #12612, opened 2025-11-14
Security config and option cleanup for audit&blocking


naderman:
Due to a few missing options / bug reports and confusing docs right now we need to clean up docs, config and command options and env vars that influence security audits and security blocking.

# Config
Ideally we would have a different key than "config.audit", but since this predates blocking, we'll stick with it. The intro text to this docs section should make clear that the options in this section configure two separate processes:
- auditing which is performed at the end of update operations automatically, or triggered explicitly by the compoesr audit command
- blocking which is performed prior to update commands (or any commands triggering an update like require) which entirely removes metadata from dependency resolution for insecure packages

We need to ensure that all settings clearly document which of these two processes they apply to.

## Settings

- *ignore* - applies to both audit & blocking, list of ids, or id/reason, and we'll change this to allow setting `audit|block|all` for each entry to change what it applies to, default is `always`
    ```
        audit:
            ignore:
                cve-123: { apply: audit|block|all, reason: ... }
    ```
- *abandoned* - should apply only to auditing, blocking behavior separately defined by block-abandoned, needs to just be documented
- *ignore-abandoned* - should apply to both audit & blocking, should also get the new format to allow picking if ignored for audit|block|all
- *ignore-severity* - applies to both audit & blocking, should allow `["low": { "apply": "audit|block|all"}]`
- *ignore-unreachable* - applies only to audit command
- *block-insecure* - true/false - applies only to blocking
- *block-abandoned* - true/false - applies only to blocking

# CLI
We'll attempt to remain consistent with current options. So there should be one flag to disable the audit process and one flag to disable the blocking process. Both flags include their handling of abandoned packages that can be configured or overwritten through an env var.

## Update command options
```
--no-audit
--no-security-blocking
```

## Environment variables
```
COMPOSER_NO_AUDIT=0|1
COMPOSER_AUDIT_ABANDONED=ignore|report|fail
COMPOSER_NO_SECURITY_BLOCKING=0|1
COMPOSER_SECURITY_BLOCKING_ABANDONED=0|1
```

wu-edward:
The new command options/env variables will address our use case in our CI build.

PR #12617, opened 2025-11-14
New flags to disable/configure blocking

Seldaek:
Refs #12612 - first part (see individual commits). Will do the `apply: ...` stuff in a follow-up PR as that is likely to grow a bit more complex.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1352,11 +1352,11 @@ Accepts a comma-seperated list of event names, e.g. `post-install-cmd` for which
 
 ### COMPOSER_NO_AUDIT
 
-If set to `1`, it is the equivalent of passing the `--no-audit` option to `require`, `update`, `remove` or `create-project` command.
+If set to `1`, it is the equivalent of passing the `--no-audit` option to a `require`, `update`, `remove` or `create-project` command.
 
 ### COMPOSER_NO_SECURITY_BLOCKING
 
-If set to `1`, it is the equivalent of passing the `--no-security-blocking` option to `require`, `update`, `remove`, or `create-project` command. This allows installing packages with security advisories or that are abandoned.
+If set to `1`, it is the equivalent of passing the `--no-security-blocking` option to a `require`, `update`, `remove`, or `create-project` command. This allows installing packages with security advisories or that are abandoned.
 
 ### COMPOSER_SECURITY_BLOCKING_ABANDONED
 
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -155,7 +155,7 @@ Defaults to `report` in Composer 2.6, and defaults to `fail` from Composer 2.7 o
 - `report` means abandoned packages are reported as an error but do not cause the command to exit with a non-zero code.
 - `fail` means abandoned packages will cause audits to fail with a non-zero code.
 
-Note that this only applies to auditing, and not to the blocking of insecure
+Note, that this only applies to auditing, and not to the blocking of insecure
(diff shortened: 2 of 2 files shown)


doc/03-cli.md
```diff
@@ -1350,6 +1354,14 @@ Accepts a comma-seperated list of event names, e.g. `post-install-cmd` for which
 
 If set to `1`, it is the equivalent of passing the `--no-audit` option to `require`, `update`, `remove` or `create-project` command.
 
+### COMPOSER_NO_SECURITY_BLOCKING
+
+If set to `1`, it is the equivalent of passing the `--no-security-blocking` option to `require`, `update`, `remove`, or `create-project` command. This allows installing packages with security advisories or that are abandoned.
```
naderman:
```suggestion
If set to `1`, it is the equivalent of passing the `--no-security-blocking` option to a `require`, `update`, `remove`, or `create-project` command. This allows installing packages with security advisories or that are abandoned.
```

doc/03-cli.md
```diff
@@ -1350,6 +1354,14 @@ Accepts a comma-seperated list of event names, e.g. `post-install-cmd` for which
 
 If set to `1`, it is the equivalent of passing the `--no-audit` option to `require`, `update`, `remove` or `create-project` command.
```
naderman:
```suggestion
If set to `1`, it is the equivalent of passing the `--no-audit` option to a `require`, `update`, `remove` or `create-project` command.
```

doc/06-config.md
```diff
@@ -155,6 +155,10 @@ Defaults to `report` in Composer 2.6, and defaults to `fail` from Composer 2.7 o
 - `report` means abandoned packages are reported as an error but do not cause the command to exit with a non-zero code.
 - `fail` means abandoned packages will cause audits to fail with a non-zero code.
 
+Note that this only applies to auditing, and not to the blocking of insecure
```
naderman:
```suggestion
Note, that this only applies to auditing, and not to the blocking of insecure
```

doc/03-cli.md
stasadev: Is there a reason not to have this flag for the `install` command?
It doesn't work, if you don't have a `composer.lock` file.

In DDEV, we split `composer create-project` in several steps https://docs.ddev.com/en/stable/users/usage/developer-tools/#ddev-and-composer (needed for internal reasons).

If I run this, it doesn't let me to install:

```bash
composer create-project "typo3/cms-base-distribution:^11" . --no-install
COMPOSER_NO_SECURITY_BLOCKING=1 composer install
```

But at the same time I can do this:

```bash
composer create-project "typo3/cms-base-distribution:^11" . --no-install
COMPOSER_NO_SECURITY_BLOCKING=1 composer update
```

naderman: If you run the install command without a lock file you are performing an update, so call the update command if you don't have a lock file. The install command should not be used without a lock file as is evident from its output too, hopefully.

But since people insist on doing this, I guess it would only be consistent to pass along these flags to any update triggered by the install command when it finds the lock file it requires is missing.

naderman: The create-project command does this internally and executes an update if there is no lock file present: https://github.com/composer/composer/blob/main/src/Composer/Command/CreateProjectCommand.php#L253

Seldaek: Yeah I think I'd rather not add it so that it remains very clear that install itself does NOT do any filtering of vulnerable dependencies, it installs a lock file as is and that's it.

PR #12618, opened 2025-11-14
Add a way to configure block and audit specific ignores

### LINKED ISSUE #12612: Security config and option cleanup for audit&blocking

Seldaek:
Fixes #12612

Also @naderman please review docs carefully (just a reminder:)

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -430,7 +430,7 @@
                             "anyOf": [
                                 {
                                     "type": "object",
-                                    "description": "A list of advisory ids, remote ids or CVE ids (keys) with either explanations (string values) or detailed configuration (object values with 'apply' and 'reason' fields) for why they're ignored.",
+                                    "description": "A list of advisory ids, remote ids, CVE ids or package names (keys) with either explanations (string values) or detailed configuration (object values with 'apply' and 'reason' fields) for why they're ignored.",
                                     "additionalProperties": {
                                         "anyOf": [
                                             {
@@ -461,7 +461,7 @@
                                 },
                                 {
                                     "type": "array",
-                                    "description": "A set of advisory ids, remote ids or CVE ids that are reported but let the audit command pass. Applies to audit reports and version blocking.",
+                                    "description": "A set of advisory ids, remote ids, CVE ids or package names that are reported but let the audit command pass. Applies to audit reports and version blocking.",
                                     "items": {
                                         "type": "string"
                                     }


res/composer-schema.json
```diff
@@ -430,9 +430,33 @@
...
+                                    "description": "A list of advisory ids, remote ids or CVE ids (keys) with either explanations (string values) or detailed configuration (object values with 'apply' and 'reason' fields) for why they're being ignored.",
                                     "additionalProperties": {
-                                        "type": ["string", "string"]
+                                        "anyOf": [
+                                            {
+                                                "type": "string",
+                                                "description": "Explanation for ignoring this advisory (applies to both audit and blocking)"
+                                            },
+                                            {
+                                                "type": "object",
```
stof, line 441: shouldn't this object define some properties as required ?

naderman: I think neither of the two properties is required. You can use the object purely to define a reason, if you want to use the default value for application, or to define the application mode and skip the reason. In the end there's no harm in leaving it empty either. Just means default application, empty reason?

PR #12622, opened 2025-11-17
feat: allow ignoring cve of packages, fixes #12616

### LINKED ISSUE #12616: Ignoring security issues of specific packages
We offer a security composer package which fixes security issues of our own dependencies for people who don't want to update completely yet. (they maybe have to fix breaking change whatever)

Therefore when that security package is installed, those CVE's are solved. We can ignore right now a list of CVE, I would like to ignore entire composer packages for this specific need. The CVE list can grow and is not maintainable for this purpose.

```json
{
"config": {
 "audit": {
    "ignore-packages": [ "shopware/core" ]
  }
}
}
```

Exclude entire Composer Packages from this checking

- Adding all CVE -> which is an hassle for users
- Telling Users to disable this, but that would disable it for all packages which we don't want to recommend users too
- Creating a Composer Plugin (we would like to contribute to composer itself instead of building something on our own)

**Additional context**

I am open to develop this, I just want feedback suggestion before I implement it and then it gets maybe rejected because unwanted 😅

shyim:
Allows ignoring all CVE of specific packages

### DIFF
--- a/src/Composer/Advisory/AuditConfig.php
+++ b/src/Composer/Advisory/AuditConfig.php
@@ -31,7 +31,7 @@ class AuditConfig
     public $auditFormat;
 
     /**
-     * @var array<string>|array<string,string> List of advisory IDs, remote IDs or CVE IDs that reported but not listed as vulnerabilities.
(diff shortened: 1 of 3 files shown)


Seldaek:
Looks good thanks

PR #12626, opened 2025-11-19
Fix partial updates failing when a locked package has security advisories

### LINKED ISSUE #12620: Package security blocking should not affect locked packages in partial updates
Per https://github.com/composer/composer/issues/12607#issuecomment-3533647652

If we do a partial update, a locked package with vulnerabilities should not be forced to update, it should just remain pinned.

Seldaek:
Fixes #12620


src/Composer/DependencyResolver/SecurityAdvisoryPoolFilter.php
```diff
@@ -55,7 +55,7 @@ public function filter(Pool $pool, array $repositories): Pool
 
             $packagesForAdvisories = [];
             foreach ($pool->getPackages() as $package) {
-                if (!$package instanceof RootPackageInterface && !PlatformRepository::isPlatformPackage($package->getName())) {
+                if (!$package instanceof RootPackageInterface && !PlatformRepository::isPlatformPackage($package->getName()) && !$request->isLockedPackage($package)) {
```
naderman: I'm not sure this is right. Now any locked package may remain? Shouldn't the packages marked for an update and if dependencies included, its dependencies be excluded here?

naderman: Will push a test for what I mean I guess

Seldaek: When the PoolBuilder unlocks a package for update it also unlocks it in the Request, so my assumption was that this will just work, but feel free to add a test :)

naderman: Yup sorry you are right, just getting mixed up with the ordering of filter and pool builder. Added a test and it passes.

PR #12647, opened 2025-11-29
Add --strict-psr-autoloader option to install/update commands

Loyen:
I noticed that `composer install` do not support `--strict-psr`, meaning that if you want to make use of it you would have to run `install` followed by `dump-autoload` with `--strict-psr`. 

Since both commands support the `optimize autoloader` option (`composer install --optimise-autoloader` or `composer dump-autoload --optimise`) it feels reasonable that both also support `--strict-psr`.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -121,7 +121,7 @@ resolution.
   a bit of time to run so it is currently not done by default.
 * **--classmap-authoritative (-a):** Autoload classes from the classmap only.
   Implicitly enables `--optimize-autoloader`.
-* **--strict-psr-autoloader:** Return a failed exit code (1) if PSR-4 or PSR-0 mapping errors
+* **--strict-psr-autoloader:** Return a failed exit code (6) if PSR-4 or PSR-0 mapping errors
   are present in the current project (dependencies excluded). Requires `--optimize-autoloader` to work.
 * **--apcu-autoloader:** Use APCu to cache found/not-found classes.
 * **--apcu-autoloader-prefix:** Use a custom prefix for the APCu autoloader cache.
@@ -215,7 +215,7 @@ php composer.phar update vendor/package:2.0.1 vendor/package2:3.0.*
   a bit of time to run, so it is currently not done by default.
 * **--classmap-authoritative (-a):** Autoload classes from the classmap only.
   Implicitly enables `--optimize-autoloader`.
-* **--strict-psr-autoloader:** Return a failed exit code (1) if PSR-4 or PSR-0 mapping errors
+* **--strict-psr-autoloader:** Return a failed exit code (6) if PSR-4 or PSR-0 mapping errors
   are present in the current project (dependencies excluded). Requires `--optimize-autoloader` to work.
 * **--apcu-autoloader:** Use APCu to cache found/not-found classes.
 * **--apcu-autoloader-prefix:** Use a custom prefix for the APCu autoloader cache.
--- a/src/Composer/Command/InstallCommand.php
+++ b/src/Composer/Command/InstallCommand.php
@@ -55,7 +55,7 @@ class InstallCommand extends BaseCommand
                 new InputOption('verbose', 'v|vv|vvv', InputOption::VALUE_NONE, 'Shows more details including new commits pulled in when updating packages.'),
                 new InputOption('optimize-autoloader', 'o', InputOption::VALUE_NONE, 'Optimize autoloader during autoloader dump'),
                 new InputOption('classmap-authoritative', 'a', InputOption::VALUE_NONE, 'Autoload classes from the classmap only. Implicitly enables `--optimize-autoloader`.'),
-                new InputOption('strict-psr-autoloader', null, InputOption::VALUE_NONE, 'Return a failed status code (1) if PSR-4 or PSR-0 mapping errors are present. Requires --optimize-autoloader to work.'),
+                new InputOption('strict-psr-autoloader', null, InputOption::VALUE_NONE, 'Return a failed status code (6) if PSR-4 or PSR-0 mapping errors are present. Requires --optimize-autoloader to work.'),
                 new InputOption('apcu-autoloader', null, InputOption::VALUE_NONE, 'Use APCu to cache found/not-found classes.'),
                 new InputOption('apcu-autoloader-prefix', null, InputOption::VALUE_REQUIRED, 'Use a custom prefix for the APCu autoloader cache. Implicitly enables --apcu-autoloader'),
                 new InputOption('ignore-platform-req', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Ignore a specific platform requirement (php & ext- packages).'),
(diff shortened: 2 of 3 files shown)


Loyen:
I had a look regarding tests for this, they would look pretty similar to what currently exists for `dump-autoload` so I have not included any here for it yet as I am unsure if this is something that is wanted. 

At least when it comes to failures as [DumpAutoloadCommandTest::testFailsUsingStrictPsrIfClassMapViolationsAreFound()](https://github.com/composer/composer/blob/2f70d210bbb285662f8163b8b811c15abf6ea0ba/tests/Composer/Test/Command/DumpAutoloadCommandTest.php#L60-L76) is pretty much the same with a different command. 

Looking at the [DumpAutoloadCommandTest ::testUsingOptimizeAndStrictPsr()](https://github.com/composer/composer/blob/2f70d210bbb285662f8163b8b811c15abf6ea0ba/tests/Composer/Test/Command/DumpAutoloadCommandTest.php#L50-L58) we wouldn't be able to assert by the same things as `composer install --optimize-autoloader` do not output how many classes were part of the optimized autoload files. Perhaps that is fine to leave out as an assertion though and I should add these two tests to `InstallCommandTest` as well.

This can definitely be adjusted so that they are both expected to output similar results if we want that.

src/Composer/Command/InstallCommand.php
```diff
@@ -59,6 +59,7 @@ protected function configure(): void
                 new InputOption('apcu-autoloader-prefix', null, InputOption::VALUE_REQUIRED, 'Use a custom prefix for the APCu autoloader cache. Implicitly enables --apcu-autoloader'),
                 new InputOption('ignore-platform-req', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Ignore a specific platform requirement (php & ext- packages).'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore all platform requirements (php & ext- packages).'),
+                new InputOption('strict-psr', null, InputOption::VALUE_NONE, 'Return a failed status code (1) if PSR-4 or PSR-0 mapping errors are present. Requires --optimize to work.'),
```
stof: In the install and update commands, all options related to configuring the autoloader contain `autoloader` in the name to avoid ambiguity.

stof: The description of this option is wrong. There is no `--optimize` option in the install command. It is named `--optimize-autoloader`

stof: I also suggest putting it near the other autoloader options, so that they are next to each other in the command help.

Loyen: Cheers, must have missed it.

Loyen: Good point. I can see the other options related to autoloader have a `-autoloader` suffix. Will have a look later to adjust it.

Loyen: This should now have been adjusted.

Seldaek:
Seems reasonable to me, and I'd concur in saying tests are not needed, it'll just not test much of value.

src/Composer/Installer.php
```diff
@@ -373,6 +375,10 @@ public function run(): int
                     null,
                     $this->locker
                 );
+
+            if ($this->strictPsrAutoloader && count($classMap->getPsrViolations()) > 0) {
+                return 1;
```
Seldaek: It'd be good to add a new error constant on top of the class for this, with a new value like `public const ERROR_PSR_AUTOLOAD_VIOLATION = 6;` ?

Loyen: Agreed, adds a bit of consistency and clarity. Pushed in commit 627553d.

src/Composer/Command/UpdateCommand.php
```diff
@@ -73,6 +73,7 @@ protected function configure(): void
                 new InputOption('verbose', 'v|vv|vvv', InputOption::VALUE_NONE, 'Shows more details including new commits pulled in when updating packages.'),
                 new InputOption('optimize-autoloader', 'o', InputOption::VALUE_NONE, 'Optimize autoloader during autoloader dump.'),
                 new InputOption('classmap-authoritative', 'a', InputOption::VALUE_NONE, 'Autoload classes from the classmap only. Implicitly enables `--optimize-autoloader`.'),
+                new InputOption('strict-psr-autoloader', null, InputOption::VALUE_NONE, 'Return a failed status code (1) if PSR-4 or PSR-0 mapping errors are present. Requires --optimize-autoloader to work.'),
```
Seldaek:
```suggestion
                new InputOption('strict-psr-autoloader', null, InputOption::VALUE_NONE, 'Return a failed status code (6) if PSR-4 or PSR-0 mapping errors are present. Requires --optimize-autoloader to work.'),
```

src/Composer/Command/InstallCommand.php
```diff
@@ -55,6 +55,7 @@ protected function configure(): void
                 new InputOption('verbose', 'v|vv|vvv', InputOption::VALUE_NONE, 'Shows more details including new commits pulled in when updating packages.'),
                 new InputOption('optimize-autoloader', 'o', InputOption::VALUE_NONE, 'Optimize autoloader during autoloader dump'),
                 new InputOption('classmap-authoritative', 'a', InputOption::VALUE_NONE, 'Autoload classes from the classmap only. Implicitly enables `--optimize-autoloader`.'),
+                new InputOption('strict-psr-autoloader', null, InputOption::VALUE_NONE, 'Return a failed status code (1) if PSR-4 or PSR-0 mapping errors are present. Requires --optimize-autoloader to work.'),
```
Seldaek:
```suggestion
                new InputOption('strict-psr-autoloader', null, InputOption::VALUE_NONE, 'Return a failed status code (6) if PSR-4 or PSR-0 mapping errors are present. Requires --optimize-autoloader to work.'),
```

doc/03-cli.md
```diff
@@ -121,6 +121,8 @@ resolution.
   a bit of time to run so it is currently not done by default.
 * **--classmap-authoritative (-a):** Autoload classes from the classmap only.
   Implicitly enables `--optimize-autoloader`.
+* **--strict-psr-autoloader:** Return a failed exit code (1) if PSR-4 or PSR-0 mapping errors
```
```suggestion
* **--strict-psr-autoloader:** Return a failed exit code (6) if PSR-4 or PSR-0 mapping errors
```

doc/03-cli.md
```diff
@@ -213,6 +215,8 @@ php composer.phar update vendor/package:2.0.1 vendor/package2:3.0.*
   a bit of time to run, so it is currently not done by default.
 * **--classmap-authoritative (-a):** Autoload classes from the classmap only.
   Implicitly enables `--optimize-autoloader`.
+* **--strict-psr-autoloader:** Return a failed exit code (1) if PSR-4 or PSR-0 mapping errors
```
```suggestion
* **--strict-psr-autoloader:** Return a failed exit code (6) if PSR-4 or PSR-0 mapping errors
```

Seldaek:
Thanks

Issue #11125, opened 2022-10-14
Temporary constraints does not work when using wildcard


T2L:
This is follow-up for #11124. Basically, I was trying to fix the  #11124 by using  temporary constraints, but that didn't not work either

My `composer.json`:

```json
{
    "require": {
        "drupal/core-recommended": "^9.4",
        "drush/drush": "^11.2"
    },
    "repositories": {
        "_drupal": {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "config": {
        "preferred-install": "dist",
        "allow-plugins": {
            "dealerdirect/phpcodesniffer-composer-installer": true
        },
        "platform": {
            "php": "8.0"
        }
    },
    "require-dev": {
        "drupal/core-dev": "^9.4"
    }
}
```

Output of `composer diagnose`:

```
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK git version 2.38.0
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: You are not running the latest snapshot version, run `composer self-update` to update (68da418f6e229ef6ade29c9c44d0eb05b996f218 => 9ac186bf46b0a7ae015fbcd725c088be4905f2f6)
Composer version: 2.5-dev+68da418f6e229ef6ade29c9c44d0eb05b996f218
PHP version: 8.0 - Package overridden via config.platform, actual: 8.0.24
PHP binary path: /usr/local/Cellar/php@8.0/8.0.24/bin/php
OpenSSL version: OpenSSL 1.1.1q  5 Jul 2022
cURL version: 7.85.0 libz 1.2.11 ssl (SecureTransport) OpenSSL/1.1.1q
zip: extension present, unzip present, 7-Zip not available
```

When I run this command:

```
composer u 'drupal/core*:9.4.8' -W -vv
```

I get the following output:

```
Loading composer repositories with package information
[...]

PR #12658, opened 2025-12-06
add support for temporary constraint wildcards in UpdateCommand

verfriemelt-dot-org:
This will allow multiple packages to selected for update using a temporary constraint. Useful for CI/CD type workloads testing multiple supported versions for a given package like this

```shell
$ composer update --with "phpunit/*:12.4.*"
```

Updates using wildcards will only affect root requirements!

see https://github.com/composer/composer/issues/11125


verfriemelt-dot-org:
rebased onto main

Seldaek:
Thanks!

PR #12660, opened 2025-12-07
[bugfix] fix crash bumping version with composer update and disabled lockfile

verfriemelt-dot-org:
fixes a crash when using `--bump-after-update` on a project with disabled lockfile
```shell
$ cd $(mktemp -d) 
$ composer require symfony/dotenv "^6.0.0"
$ rm composer.lock
$ composer config lock false
$ composer update --bump-after-update
Updating dependencies
Generating autoload files
6 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
No security vulnerability advisories found.
Bumping dependencies

In JsonFile.php line 398:

  "/dev/null" does not contain valid JSON
  Parse error on line 1:

  ^

```

### DIFF
--- a/src/Composer/Command/BumpCommand.php
+++ b/src/Composer/Command/BumpCommand.php
@@ -121,7 +121,10 @@ EOT
         unset($contents);
 
         $composer = $this->requireComposer();
-        if ($composer->getLocker()->isLocked()) {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks!

PR #12661, opened 2025-12-07
fix composer compile with custom gitconfig

verfriemelt-dot-org:
i happen to have a git configuration:

```
[log]
    showSignature = true
```

this breaks the command to obtain the datetime for the `composer compile` build, which uses `git log -n1 --pretty=%ci` and returning something like the following output

```shell
$  git log -n1 --pretty=%ci
Good "git" signature for easteregg@verfriemelt.org with RSA key SHA256:kwWUDttgZRoJgAE8i9tHFLqCkUWEnj+S5sct1Q9dHBs
2025-12-07 11:48:18 +0100
```

to avoid custom configuration to interfere with the build process, we should ignore that leveraging env vars to disable user-specific configuration: https://git-scm.com/docs/git-config#Documentation/git-config.txt-GITCONFIGGLOBAL

to achieve that, i added `ProcessExecutor::setEnvironment(array $environment): void;` to implement that. let me know if thats something you are intrested in, then i can flesh out the testing supporting that.

### DIFF
--- a/src/Composer/Compiler.php
+++ b/src/Composer/Compiler.php
@@ -49,6 +49,10 @@ class Compiler
         }
 
         $process = new ProcessExecutor();
+        $process->setEnvironment([
+            'GIT_CONFIG_GLOBAL' => '/dev/null',
+            'GIT_CONFIG_SYSTEM' => '/dev/null',
+        ]);
 
         if (0 !== $process->execute(['git', 'log', '--pretty=%H', '-n1', 'HEAD'], $output, dirname(__DIR__, 2))) {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks for the analysis and PR. After a quick google session tho I've found that rev-list doesn't have this problem, so I'd rather do that than overwriting env vars and will thus close this PR in favor of https://github.com/composer/composer/pull/12666

verfriemelt-dot-org:
fair point, thinking about it, reaching for [gits plumbing](https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain) would in fact be the better solution

Issue #12635, opened 2025-11-24
Link to Security Information Page when CVE found


shyim:
**Is your feature request related to a problem? Please describe.**

Right now as an user you get an message like this when you want to install an package with CVE:

```
    - Root composer.json requires shopware/core 6.5.8.8 (exact version match: 6.5.8.8), found shopware/core[v6.5.8.8] but these were not loaded, because they are affected by security advisories. To ignore the advisories, add ("PKSA-w3qy-s9h7-2hqr", "PKSA-b824-t6kf-bqqz", "PKSA-6wp3-462p-vyty", "PKSA-h5dj-jyqc-4fjr", "PKSA-kypv-cx5n-qkc8", "PKSA-v415-g75g-bqsy", "PKSA-8vfm-96b7-t9nt", "PKSA-m54b-2v2z-x1bs", "PKSA-frt7-rv6d-9v53", "PKSA-dbxn-psgm-2qmr", "PKSA-k472-zz4q-rd5r", "PKSA-kt1g-n1g2-hzb4", "PKSA-wp2c-7yp8-5fvs", "PKSA-4spx-rq41-wk8h", "PKSA-6stq-czfs-1nvv") to the audit "ignore" config. To turn the feature off entirely, you can set "block-insecure" to false in your "audit" config.
```

We would like to give users more information what they can do as an vendor.
**Describe the solution you'd like**

Allow linking to an Security Page in the vulnerable package:

```json
{
  "support": {
      "security-info": "foo.com"
   }
}
```

(the key "security" is assigned for "vulnerability disclosure policy https://getcomposer.org/doc/04-schema.md#support)

And that page would be shown to the user too in the message maybe like:

```
To turn the feature off entirely, you can set "block-insecure" to false in your "audit" config.
For more information about security vulnerabilities for this package visit: foo.com
```

**Describe alternatives you've considered**
Writing a Composer Plugin to show help information

**Additional context**
We have an additional composer package, which patches CVE. We would like to inform people about this

77553322AHST:
@shyim 
@davidolrik 
@Pawka 
@zeke 
@holger 
@ZAMAN-COMPANY-COM
@77553322AHST
[BOT](https://core.telegram.org/tmr8_bot)

Seldaek:
@shyim I don't get the whole thing about patching CVEs with a package.. why not tag patch releases with the fix so people can simply update? This seems very convoluted.

Seldaek:
[...]

PR #12665, opened 2025-12-08
Add links to advisories problem output

### LINKED ISSUE #12636: Include or point users to `composer audit` information when a security vulnerability blocks a package

The new built-in security blocking is a fantastic addition, but the information and advice given could be more helpful.

For example:

> - Root composer.json requires twig/twig ^2.16.1, found twig/twig[v2.16.1] but these were not loaded, because they are affected by security advisories. To ignore the advisories, add ("PKSA-yhcn-xrg3-68b1", "PKSA-2wrf-1xmk-1pky") to the audit "ignore" config. To turn the feature off entirely, you can set "block-insecure" to false in your "audit" config.

* The only identifier listed are the "PKSA" identifiers, which are currently very hard to search for (they are not yet listed as aliases in other databases, or indexed by generic search engines)
* The advice only talks about ignoring the advisories, which users should do only _after_ understanding the issue

The information given by `composer audit` is much more helpful, with a CVE reference and a description URL. Perhaps this information could be automatically displayed for vulnerabilities which block an update?

Alternatively, a command could be added, similar to `composer why-not`, which showed vulnerabilities for a particular package, or looked up information based on their PKSA identifier.

The advice in the "Problem" statement could then be reworded to advise running this command **before** considering adding them to the ignore list.

Seldaek:
Refs #12635

Fixes #12636

### DIFF
--- a/src/Composer/DependencyResolver/Problem.php
+++ b/src/Composer/DependencyResolver/Problem.php
@@ -12,6 +12,8 @@
 
 namespace Composer\DependencyResolver;
 
+use Composer\Advisory\PartialSecurityAdvisory;
+use Composer\Advisory\SecurityAdvisory;
 use Composer\Package\CompletePackageInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\BasePackage;
@@ -26,6 +28,7 @@ use Composer\Semver\Constraint\ConstraintInterface;
 use Composer\Package\Version\VersionParser;
 use Composer\Repository\PlatformRepository;
 use Composer\Semver\Constraint\MultiConstraint;
+use Symfony\Component\Console\Formatter\OutputFormatter;
 
 /**
  * Represents a problem detected while solving dependencies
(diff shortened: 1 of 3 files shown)


src/Composer/DependencyResolver/Pool.php
```diff
@@ -132,7 +132,7 @@ public function getSecurityAdvisoryIdentifiersForPackageVersion(string $packageN
         foreach ($this->securityRemovedVersions[$packageName] ?? [] as $version => $packageWithSecurityAdvisories) {
             if ($constraint !== null && $constraint->matches(new Constraint('==', $version))) {
                 return array_map(static function ($advisory) {
```
stof, line 134: The array_map can be removed entirely now as it does not change anything anymore.

And the return type of the method must be updated (and probably also changing the name, as it does not return advisory identifiers anymore)

src/Composer/DependencyResolver/Problem.php
```diff
@@ -388,7 +391,27 @@ public static function getMissingPackageReason(RepositorySet $repositorySet, Req
...
-                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but these were not loaded, because they are affected by security advisories. To ignore the advisories, add ("' . implode('", "', $pool->getSecurityAdvisoryIdentifiersForPackageVersion($packageName, $constraint)). '") to the audit "ignore" config. To turn the feature off entirely, you can set "block-insecure" to false in your "audit" config.'];
+                $advisories = $repositorySet->getMatchingSecurityAdvisories($packages, false, true);
+                if (isset($advisories['advisories'][$packageName]) && \count($advisories['advisories'][$packageName]) > 0) {
+                    $advisoriesList = array_map(static function (SecurityAdvisory $advisory): string {
+                        if ($advisory->link !== null && $advisory->link !== '') {
+                            return '<href='.OutputFormatter::escape($advisory->link).'>'.$advisory->advisoryId.'</>';
+                        }
+
+                        return $advisory->id;
```
stof: should be `advisoryId`, not `id`

src/Composer/DependencyResolver/Pool.php
Seldaek: Yeah, looking at too many things at once and forgot what I was up to there 🙈

PR #12666, opened 2025-12-08
Use git rev-list instead of log to avoid issues with log.showSignature being enabled

Seldaek:
Fixes #12661

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -4,9 +4,9 @@
 $cwd = getcwd();
 assert(is_string($cwd));
 chdir(__DIR__.'/../');
-$ts = rtrim(exec('git log -n1 --pretty=%ct HEAD'));
+$ts = rtrim(exec('git rev-list --no-commit-header -n1 --format=%ct HEAD'));
 if (!is_numeric($ts)) {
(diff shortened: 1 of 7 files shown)


Seldaek:
Converted all other usages, and found one where we had `-c log.showSignature=false` as a hack as well..

mpchadwick:
This change is breaking composer install on our CI system Scrutinizer-CI. I did some debugging and it looks like their VMs have git 2.24.0, which doesn't support `--no-commit-header`.

The composer install command crashes with the following output

```
Skipped branch zzz1, Failed to parse time string (@) at position 0 (@): Unexpected character


Skipped branch zzz2 Failed to parse time string (@) at position 0 (@): Unexpected character
.
.
.
In VcsRepository.php line 423:
                                                                               
  No valid composer.json was found in any branch or tag of https://github.com  
  /org/repo.git, could not lo  
  ad a package from it.                                                        
                      
```

I've logged an issue about this here: https://github.com/composer/composer/issues/12701

enricobattocchi:
Same as @mpchadwick here ([discussion](https://github.com/composer/composer/discussions/12700)). I think it's worth mentioning that it's a bit tricky for us to work around this issue since we are stuck to using a Debian bullseye docker image where Git is 2.30.2 and no upgrade is available.

PR #12667, opened 2025-12-08
fix(AuthHelper:client-certificate): misplaced `username` and `password`

rtm-ctrlz:
There is typo with applying transport options for Client-Certificate authorization option introduced at #12406.


Seldaek:
Oops :D Thanks

Issue #12633, opened 2025-11-22
Implement dependency cooldowns in composer


phront:
**Is your feature request related to a problem? Please describe.**
 I have recently read blog post [We should all be using dependency cooldowns](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns), but not found the feature in composer. It would be nice to have dependency cooldown option in composer

**Describe the solution you'd like**
_Dependency cooldowns are a free, easy, and incredibly effective way to mitigate the large majority of open source supply chain attacks. More individual projects should apply cooldowns (via tools like Dependabot and Renovate) to their dependencies, and packaging ecosystems should invest in first-class support for cooldowns directly in their package managers._

freshleafmedia:
Related https://github.com/composer/composer/issues/12552

Seldaek:
I have also read that post a few days ago and it is definitely more well written. It got me to think that maybe we can do something here, but we need to consider all options carefully.

Namely:

- if a package has a known vulnerability it should at least allow an update to latest version vs remaining stuck on an outdated version.
- we need a way to specify this delay in amount of hours/days IMO
- add `minimum-release-age` (following pnpm's naming) to config settings
- this filtering should not (and can technically not) apply to dev versions.
- should it be enabled by default with a low 6-12h value? It would make it have a greater impact in terms of security, but would also probably disturb a lot of existing workflows so it may not be worth it.

8ctopus:
I agree with all your points, and would like to comment on this one:

> should it be enabled by default with a low 6-12h value? It would make it have a greater impact in terms of security, but would also probably disturb a lot of existing workflows so it may not be worth it.

I would not do it in release 2.10 and based on the feedback, consider changing the default in a later version.

TimWolla:
Some thoughts from my side regarding the possible concerns from that other issue, specifically:

> what is a patch release fixes a regression that was introduced a few days before, you then risk updating to that broken version just because you are ignoring the latest.

What might make sense would be treating the “minimum-release-age” as the age of the oldest version that has been released more recently than the currently pinned version, but then always updating to the latest version (or a newer version with a shorter minimum-release-age). To give an example:

- My application is currently pinned to version 1.2.3 released 2025-11-01.
- Version 1.3.0 is released 2025-12-01.
- Version 1.3.1 fixing a bad regression is released 2025-12-03.
- I configure a minimum-release-age of 7.

[...]

PR #12692, opened 2025-12-27
Add cooldown policy config to filter newly released packages

crocodele:
Introduces a configurable cooldown: a waiting period before a newly published package version becomes installable, to reduce exposure to supply-chain attacks where a compromised maintainer pushes a malicious release (many such versions are caught within hours/days of publication).

### Configuration

Lives under `config.policy.cooldown`:

```json
{
    "config": {
        "policy": {
            "cooldown": {
                "age": "7 days",
                "block": true,
                "audit": "report",
                "ignore": {
                    "vendor/pkg": "reason / pinned for now"
                }
            }
        }
    }
}
```

- **`age`**: cooldown duration. Integer seconds, or `"<n> <unit>"` where unit is `second`/`minute`/`hour`/`day`/`week` (singular or plural). `null`/`0` disables. Relative phrases like `"tomorrow"` are rejected rather than silently misinterpreted.
- **`block`** (default `true`): withhold too-new versions during `update`/`require`.
- **`audit`** (`ignore`|`report`|`fail`, default `ignore`): how `composer audit` treats within-cooldown installed/locked packages.
- **`ignore`**: package/constraint patterns that bypass the cooldown.
- Env overrides: `COMPOSER_POLICY_COOLDOWN_AGE`, `COMPOSER_POLICY_COOLDOWN_BLOCK` (and the global `COMPOSER_POLICY` switch).
- Settable via `composer config policy.cooldown.age "7 days"` etc.

### Trustworthy timestamp (`published-time`)

The cooldown measures against a new repository-owned field, **`published-time`**, which the package author cannot influence – falling back to the existing author-controlled `time` only when a repository does not provide it. When the fallback is used, the cooldown message says so explicitly. This closes the "attacker back-dates `time` to skip the cooldown" hole. `published-time` is added to the repository JSON schema and is persisted in `composer.lock`/`installed.json` so `composer audit` can use the authoritative date offline. (Server-side population on Packagist is already in place.)

### Behaviour

- **Security-fix bypass**: a version released after a matching security advisory (and not affected by it), within 2x the cooldown window, bypasses the cooldown so fixes aren't withheld. Only the oldest such fix per constraint part bypasses. Works even when advisory *blocking* is disabled.
- **Never filtered**: root package, platform packages, locked/installed packages, dev versions, and versions with no usable date (conservative).
- **Audit integration**: `composer audit` can report or fail on within-cooldown packages.
- **Clear errors**: when a required version is withheld, the message states how long until it clears and whether the authoritative or fallback timestamp was used, and points at `policy.cooldown.ignore` / `policy.cooldown.block`.

### Backwards compatibility

Opt-in: no effect until an `age` is set. Repositories and older clients that don't know `published-time` are unaffected (unknown field ignored / graceful fallback to `time`).

Related to #12633.

*Developed with assistance from Claude Code (Opus 4.x).*


src/Composer/Config.php
```diff
@@ -91,6 +91,7 @@ class Config
         'client-certificate' => [],
         'forgejo-domains' => ['codeberg.org'],
         'forgejo-token' => [],
+        'minimum-release-age' => ['minimum-age' => null, 'exceptions' => []],
```
Seldaek: It'd be great to add support in ConfigCommand as well for setting the age and exceptions

Seldaek:
Thanks a bunch, looks quite solid at first sight but I'll have to review more in depth for sure.

src/Composer/Config.php
crocodele: Added in https://github.com/composer/composer/pull/12692/commits/5e094e3cbc0048836b1175bba698777e0758e1c1. I added test cases only for happy paths.

Seldaek:
So thinking some more about this feature, I realized that an attacked could right now easily fake the release date to put it in the past, as the release is under control of the package authors..

Thus I think if we want to do this, we should have a new date like publish_date or something that is purely owned by the package repositories and cannot be set by the packages. And the minimum-release-age would apply to that date (if available). Just writing this down for future reference.. not really expecting you to act on this right now.

stof:
@Seldaek isn't the release date the date of the git commit for VCS-based repositories ?

src/Composer/DependencyResolver/ReleaseAgePoolFilter.php
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+        $constraintParts = $this->getConstraintParts($name, $request->getRequires());
+        $packageConstraint = new Constraint('==', $package->getVersion());
+
+        // Check all matching constraint parts - package bypasses if it's the oldest for ANY part
+        foreach ($constraintParts as $index => $part) {
+            if ($part->matches($packageConstraint)) {
+                $key = $name . ':' . $index;
+                if (isset($oldestSecurityFixDate[$key]) && $releaseDate === $oldestSecurityFixDate[$key]) {
```
Seldaek: It'd be safer here to avoid relying on having the same object instance I think

```suggestion
                if (isset($oldestSecurityFixDate[$key]) && $releaseDate->getTimestamp() === $oldestSecurityFixDate[$key]->getTimestamp()) {
```

src/Composer/DependencyResolver/ReleaseAgePoolFilter.php
```diff
@@ -0,0 +1,334 @@
...
+     * Format the time until a package version will be available
+     */
+    private function formatTimeUntilAvailable(DateTimeInterface $releaseDate): string
+    {
+        $availableAt = (new \DateTimeImmutable($releaseDate->format(\DateTimeInterface::ATOM)))
+            ->modify("+{$this->config->minimumReleaseAge} seconds");
+        $diff = $this->now->diff($availableAt);
+
+        $parts = [];
+        if ($diff->d > 0) {
```
Seldaek: You need to support $diff->m and $diff->y too here for completeness to avoid showing incorrect diffs.

Seldaek: Or maybe rather use ->days which is the full amount of days diff.

src/Composer/DependencyResolver/ReleaseAgeConfig.php
```diff
@@ -0,0 +1,178 @@
...
+            $value = (int) $duration;
+            if ($value < 0) {
+                throw new \RuntimeException("Invalid minimum-release-age: duration cannot be negative ({$duration}).");
+            }
+
+            return $value;
+        }
+
+        // Parse strings like "7 days", "1 week", "24 hours"
+        $timestamp = strtotime($duration, 0);
```
Seldaek: Maybe we should validate expected formats with a regex here to avoid any random config?

src/Composer/DependencyResolver/Problem.php
```diff
@@ -390,31 +390,40 @@ public static function getMissingPackageReason(RepositorySet $repositorySet, Req
...
+            $hasSecurityRemoved = \count($securityPackages) > 0;
+            $hasReleaseAgeRemoved = \count($releaseAgePackages) > 0;
+
+            // Handle combined case: both security and release-age filters removed versions
+            if ($hasSecurityRemoved && $hasReleaseAgeRemoved) {
+                $advisoriesList = self::getAdvisoriesListForPackages($securityPackages, $repositorySet, $pool, $packageName);
+                $releaseAgeInfo = $pool->getReleaseAgeInfoForPackageVersion($packageName, $constraint);
+                $timeInfo = $releaseAgeInfo !== null ? ' (available in '.$releaseAgeInfo['availableIn'].')' : '';
+
+                // Don't pass pool to getPackageList to avoid adding all removed versions - we only want versions specific to each filter
```
Seldaek: This means versions are being hidden though, right? It'd be great to have some .test files to test all the problem output here.

Seldaek:
> @Seldaek isn't the release date the date of the git commit for VCS-based repositories ?

Yes but that is completely under the control of the committer.

And on top of that if you set the 'time' property in the composer.json you can just set whatever date and that will take precedence over the VCS inferred date.

Inscure:
The recent "axios" incident is a perfect example of why we need this feature asap.

A native quarantine period (24–72h) would provide a vital proactive buffer, shielding PHP projects from similar supply chain attacks. It allows the community to flag malicious releases before they are automatically pulled into production via automated updates.

Implementing this now would be a massive win for PHP security and ecosystem stability.

arabcoders:
> And on top of that if you set the 'time' property in the composer.json you can just set whatever date and that will take precedence over the VCS inferred date.

In my opinion this need to be addressed, as this work around rely solely on the time, if attacker is able to change the release date to say month or something then this mitigation is useless,

Recent attacks started to appear in composer registry [for ref](https://socket.dev/blog/mini-shai-hulud-packagist-malicious-intercom-php-package-compromise)

I think this option should be deprecated and removed, or at least should be turned off and rely on date from  the registry first and fallback to vcs when this option turned on

CybotTM:
> Security fixes automatically bypass cooldown when released after a security advisory (within 2x cooldown window)

Security advisories usually mention the fixed versions, only those fixed versions may bypass cooldown - but this should be configurable.

jdsdev:
If security advisories were highlighted during an update, that could also give users the option of researching the advisory, and deciding whether to bypass the cooldown manually.

Recently there was a compromise of the `intercom/intercom-php` package, so this is no longer theoretical.

crocodele:
> So thinking some more about this feature, I realized that an attacked could right now easily fake the release date to put it in the past, as the release is under control of the package authors..
> 
> Thus I think if we want to do this, we should have a new date like publish_date or something that is purely owned by the package repositories and cannot be set by the packages. And the minimum-release-age would apply to that date (if available). Just writing this down for future reference.. not really expecting you to act on this right now.

@Seldaek, do you think we can pursue this publish timestamp that is owned by the package repositories, starting with Packagist? Or what would be the best way forward? I suppose the whole `minimum-release-age` feature is useful only when the release date for package versions can't be faked – otherwise it brings a false sense of security.

Seldaek:
I don't think it's such a hard problem and I'll get on it soon but this had to take a backseat for higher prio things (also security related.. Zero chill at the moment with all the attacks and AI driven security audits)

crocodele:
> I don't think it's such a hard problem and I'll get on it soon but this had to take a backseat for higher prio things (also security related.. Zero chill at the moment with all the attacks and AI driven security audits)

Thanks! If there's something I can do to help, let me know. And once there's a clear path, I'm more than happy to clean up this PR and make the required changes here.

MarkusJLechner:
Came here by because of laravel-lang supply chain attack (https://snyk.io/de/blog/laravel-lang-supply-chain-advisory/). We tried hacking together our own fix, but this really needs to be in composer itself.

looking at the thread, it seems like any custom approach would be easily bypassed by the "backdating the release date" trick that's already been mentioned.

please prioritize merging this, it solves a real problem that we currently have no clean way to fix. thanks for the great work!

Edit: k, https://github.com/composer/composer/issues/12633#issuecomment-4537150729
Came along that comment, would leave this comment here for future readers that +1 should be avoided as this a a change of something bigger

Seldaek:
@crocodele https://github.com/composer/packagist/pull/1765 is now live so **IF** metadata has `published-time` it should be preferred over `time` as the release date. This field needs to be added to composer-schema and parsed by ArrayLoader and added to PackageInterface too.

However a few other things since the 2.10 release need to be changed here:

- I've thought about the name and IMO cooldown might be a better name, as it is much shorter and generally how people refer to this as dependency cooldown. Not sure if we need to rename or not, but worth considering.
- This should be configured within the new framework of [dependency policies](https://blog.packagist.com/composer-2-10-release/#dependency-policy-configuration) i.e. `config: policy: cooldown: { ... }`. We need to review the whole config format in depth and make sure names are unified and follow the existing policy stuff tho, I will try and do that next week.
- This [cooldown announcement](https://blog.rubygems.org/2026/06/03/cooldown-let-new-gems-be-vetted.html) from rubygems made us realize we probably should support cooldown in the `outdated` command as well, but maybe that should be kept for a follow-up PR as this one is already big enough.

src/Composer/DependencyResolver/ReleaseAgePoolFilter.php
crocodele: Addressed in one of the recent commits.

src/Composer/DependencyResolver/ReleaseAgePoolFilter.php
crocodele: Using `->days` now, in `CooldownPolicyConfig.php`.

src/Composer/DependencyResolver/ReleaseAgeConfig.php
crocodele: Ultimately I ended up doing exactly this, i.e. limiting the value to strings like `"1 week"`, `"3 days"`, `"60 hours"` etc., or an integer value which is interpreted as number of seconds. Ambiguity around valid `strtotime()` strings like `"last week"` and `"next week"` might be problematic.

crocodele:
> @crocodele [composer/packagist#1765](https://github.com/composer/packagist/pull/1765) is now live so **IF** metadata has `published-time` it should be preferred over `time` as the release date. This field needs to be added to composer-schema and parsed by ArrayLoader and added to PackageInterface too.

Using `published-time` now when available, falling back to the less-trustworthy `time`.

> However a few other things since the 2.10 release need to be changed here:
> 
> * I've thought about the name and IMO cooldown might be a better name, as it is much shorter and generally how people refer to this as dependency cooldown. Not sure if we need to rename or not, but worth considering.

I went with `minimum-release-age` originally to align with other (non-PHP) tools such as Renovate, npm, and pnpm. Renamed everything from `minimum-release-age` to `cooldown` now.

> * This should be configured within the new framework of [dependency policies](https://blog.packagist.com/composer-2-10-release/#dependency-policy-configuration) i.e. `config: policy: cooldown: { ... }`. We need to review the whole config format in depth and make sure names are unified and follow the existing policy stuff tho, I will try and do that next week.

Changed the config format.

> * This [cooldown announcement](https://blog.rubygems.org/2026/06/03/cooldown-let-new-gems-be-vetted.html) from rubygems made us realize we probably should support cooldown in the `outdated` command as well, but maybe that should be kept for a follow-up PR as this one is already big enough.

Yeah, this is almost 3k lines already, so I'll leave this out for now. :sweat_smile:

crocodele:
@Seldaek, in case you missed it, friendly note that I've updated this cooldown policy PR with the `published-time` changes, among other things. If you want, I can try to split it into smaller, more manageable PRs, and then maybe stack the cooldown support in `outdated` on top of that as well.

Seldaek:
Did not miss it.. just got caught up in summer break and other life stuff. Switched main branch to target 2.11 tho today so will get back to moving this forward now.

src/Composer/DependencyResolver/CooldownPoolFilter.php
```diff
@@ -0,0 +1,338 @@
...
+        );
+    }
+
+    /**
+     * Resolve a package's effective date once per filter() run and reuse it across the
+     * skip/age checks and the security-fix passes.
+     */
+    private function effectiveDate(PackageInterface $package): ?DateTimeInterface
+    {
+        $key = spl_object_hash($package);
```
Seldaek, line 151: Should be spl_object_id - we recently had to fix this as spl_object_hash is deprecated in PHP 8.6.
```suggestion
        $key = spl_object_id($package);
```

src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -250,6 +250,19 @@ private function configureObject(PackageInterface $package, array $config): Base
...
         }
 
+        // Server-set publication timestamp, owned by the repository and not
+        // overridable by the package author (unlike `time`). Preferred by the
+        // cooldown policy when present
+        if (isset($config['published-time']) && '' !== $config['published-time']) {
+            $publishedTime = Preg::isMatch('/^\d++$/D', $config['published-time']) ? '@'.$config['published-time'] : $config['published-time'];
+
+            try {
+                $date = new \DateTime($publishedTime, new \DateTimeZone('UTC'));
```
Seldaek, line 260: Let's not add new baggage here :)

```suggestion
                $date = new \DateTimeImmutable($publishedTime, new \DateTimeZone('UTC'));
```

src/Composer/Package/AliasPackage.php
```diff
@@ -361,6 +361,11 @@ public function getReleaseDate(): ?\DateTimeInterface
         return $this->aliasOf->getReleaseDate();
     }
 
+    public function getPublishedDate(): ?\DateTimeInterface
```
Seldaek, line 364:
```suggestion
    public function getPublishedDate(): ?\DateTimeImmutable
```

src/Composer/Package/Package.php
```diff
@@ -57,6 +57,8 @@ class Package extends BasePackage
     protected $prettyVersion;
     /** @var ?\DateTimeInterface */
     protected $releaseDate;
+    /** @var ?\DateTimeInterface */
```
Seldaek, line 60:
```suggestion
    /** @var ?\DateTimeImmutable */
```

src/Composer/Package/Package.php
```diff
@@ -395,6 +397,19 @@ public function getReleaseDate(): ?\DateTimeInterface
         return $this->releaseDate;
     }
 
+    public function setPublishedDate(?\DateTimeInterface $publishedDate): void
```
Seldaek, line 400:
```suggestion
    public function setPublishedDate(?\DateTimeImmutable $publishedDate): void
```

src/Composer/Package/Package.php
```diff
@@ -395,6 +397,19 @@ public function getReleaseDate(): ?\DateTimeInterface
...
+    {
+        $this->publishedDate = $publishedDate;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getPublishedDate(): ?\DateTimeInterface
```
Seldaek, line 408:
```suggestion
    public function getPublishedDate(): ?\DateTimeImmutable
```

src/Composer/Package/PackageInterface.php
```diff
@@ -228,6 +228,15 @@ public function getFullPrettyVersion(bool $truncate = true, int $displayMode = s
...
     public function getReleaseDate(): ?\DateTimeInterface;
 
+    /**
+     * Returns the publication date of the package as set by the repository.
+     *
+     * Unlike getReleaseDate() (which comes from the author-controlled `time`
+     * field), this is a server-set timestamp the package author cannot
+     * influence, or null if the repository did not provide one.
+     */
+    public function getPublishedDate(): ?\DateTimeInterface;
```
Seldaek, line 238:
```suggestion
    public function getPublishedDate(): ?\DateTimeImmutable;
```

src/Composer/DependencyResolver/CooldownPoolFilter.php
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /**
+     * Check if a package version is a security fix that should bypass the cooldown requirement
+     *
+     * A version is considered a security fix if:
+     * 1. The package has known security advisories
+     * 2. This version is NOT affected by any of those advisories
+     * 3. For full SecurityAdvisory: released after advisory and within 2x the cooldown age after advisory
+     * 4. For PartialSecurityAdvisory: released within 2x the cooldown age from now (fallback when reportedAt unavailable)
```
Seldaek, line 271: updates against packagist.org will normally only return PartialSecurityAdvisory, so I wanna clearly understand what is going on here..

- What's the point of this 2x cooldown age?
- Why released after the advisory? It is IMO very common that you'd release a fix and then publish the advisories, so that sounds like a bad rule.
- Why not allow any version if all the older-than-cooldown-period releases are all affected by advisories?
- Why allow this bypass at all? If someone has access to tag releases, they can also possibly publish advisories to ensure cooldown gets bypassed, so aren't we nerfing the feature?

I think I'd tend to conclude on dropping this bypass entirely, and making sure if all cooldown-available versions  are affected by some advisory that we clearly explain to the user how they can bypass it temporarily to update that one package.

Seldaek:
Some critical issues flagged by Claude review (B3/B4 I have not had time yet to review in full so not sure about accuracy but they seem very plausible so I figured I share with you already so if you look at this PR again before I get to do more research you can move forward at least..)

## B2. `composer require` picks a constraint the solver then rejects

**Severity: high (usability).** `src/Composer/Command/PackageDiscoveryTrait.php:303-304` builds a bare `RepositorySet` + `VersionSelector`, entirely outside the pool/filter pipeline:

```php
$repoSet = $this->getRepositorySet($input);
$versionSelector = new VersionSelector($repoSet, $platformRepo);
```

Same in `src/Composer/Command/CreateProjectCommand.php:446`.

So with `policy.cooldown.age = "7 days"`:

1. `composer require vendor/pkg`, where 2.1.0 was published an hour ago → `^2.1` is written into `composer.json`;
2. the subsequent solve applies the cooldown, `^2.1` becomes unsatisfiable, and the command hard-fails with "has not yet cleared the cooldown".

This is not an edge case — it fires for *any* package that has published a new minor/major inside the cooldown window, which for actively released packages is most of the time. Unlike the advisory filter (where the newest version is normally the *safe* one), the cooldown deliberately targets the newest version, so the interaction is systematic rather than incidental.

**Fix:** thread `CooldownPolicyConfig` into `findBestVersionAndNameForPackage()` and have `VersionSelector` skip within-cooldown candidates (with the same root/dev/ignore exclusions), so the recommended constraint is one the solver can satisfy. At an absolute minimum, detect the case and emit an actionable hint rather than a bare solver failure.

## B3. "Already locked packages are not affected" is false for a full `composer update`

**Severity: high.** `doc/06-config.md` (added) and the filter's own comment both claim:

> - **Already locked packages** — packages in your `composer.lock` are not affected

`Request::lockPackage()` is called from exactly one place — `src/Composer/DependencyResolver/PoolBuilder.php:242` — and it sits inside:

```php
if (\count($request->getUpdateAllowList()) > 0) {
```

i.e. **partial updates only**. On a plain `composer update`, `isLockedPackage()` is false for everything, while the lock repository is still added to the pool by `Installer.php`.

Consequence: if the lock holds a version that is still inside the cooldown — installed via a partial update, before the policy was enabled, or with `COMPOSER_POLICY_COOLDOWN_AGE=0` — the next full `composer update` withholds it and the solver **silently downgrades** the package, then re-upgrades it days later. Version churn, plus a documented promise that this cannot happen.

**Fix:** skip any package whose name+version is present in `$request->getLockedRepository()`, not just `isLockedPackage()`:

```php
$lockedRepo = $request->getLockedRepository();
// ...
|| ($lockedRepo !== null && count($lockedRepo->findPackages($package->getName(), new Constraint('==', $package->getVersion()))) > 0)
```

If the downgrade is intentional, the docs need to say so and warn that the current lock entry can go backwards.

## B4. Inline dev aliases are withheld while their target is kept

**Severity: high.** `CooldownPoolFilter.php` (diff, ~lines 103–107):

```php
if ($package instanceof RootPackageInterface
    || PlatformRepository::isPlatformPackage($package->getName())
    || $request->isLockedPackage($package)
    || $package->isDev()
```

`AliasPackage::isDev()` reflects the **alias** version, not the aliased one (`src/Composer/Package/AliasPackage.php:66-67`):

```php
$this->stability = VersionParser::parseStability($version);
$this->dev = $this->stability === 'dev';
```

Meanwhile the PR's new `AliasPackage::getPublishedDate()` (and the existing `getReleaseDate()`) forward to the aliased package. So for a common inline alias such as `"vendor/pkg": "dev-main as 1.0.0"`:

- the `dev-main` package is skipped (`isDev() === true`) and stays in the pool;
- the `AliasPackage` at version `1.0.0.0` is **not** skipped (`isDev() === false`), and its effective date is the branch's last-commit date — which for an actively developed branch is almost always inside the cooldown;
- so the alias is removed while its target remains. The pool is left inconsistent, and anything requiring `^1.0` fails with "has not yet cleared the cooldown" — for a dev branch, contradicting the documented "dev versions are never filtered".

`RepositorySet::getMatchingSecurityAdvisories()` already handles this class of problem explicitly (`$package instanceof AliasPackage && $package->isRootPackageAlias()`); the cooldown filter has no alias handling at all.

**Fix:** evaluate the skip chain against the aliased package so alias and target decisions stay consistent:

```php
$target = $package instanceof AliasPackage ? $package->getAliasOf() : $package;
if ($target instanceof RootPackageInterface
    || PlatformRepository::isPlatformPackage($target->getName())
    || $request->isLockedPackage($target)
    || $target->isDev()
```

src/Composer/DependencyResolver/PoolBuilder.php
```diff
@@ -881,4 +886,38 @@ private function runSecurityAdvisoryFilter(Pool $pool, array $repositories, Requ
...
+
+        $filtered = $total - \count($pool->getPackages());
+
+        if (0 === $filtered) {
+            return $pool;
+        }
+
+        $this->io->write(sprintf('Cooldown pool filter completed in %.3f seconds', microtime(true) - $before), true, IOInterface::VERY_VERBOSE);
+        $this->io->write(sprintf(
+            '<info>%d package version(s) withheld by the cooldown policy.</info>',
```
Seldaek, line 917: Maybe we should move this line to default verbosity and hint that -vv shows the full list? Just to avoid it withholding versions without users realizing..

tests/Composer/Test/DependencyResolver/CooldownPoolFilterTest.php
```diff
@@ -0,0 +1,878 @@
...
+        $filteredPool = $filter->filter($pool, new Request());
+
+        // Package should be filtered
+        $this->assertEmpty($filteredPool->getPackages());
+
+        // Both the main name and the provided name should be tracked
+        $this->assertTrue($filteredPool->isCooldownRemovedPackageVersion('vendor/pkg', new Constraint('==', '2.0.0.0')));
+        // Note: getNames(false) returns the package's own names, not provides
+        // The provides are tracked separately, so we only check the main name here
+    }
```
Seldaek, line 877: In addition to this, it'd be great to add a few end to end `.test` files under `tests/Composer/Test/Fixtures/installer/` - see in there for examples. These tend to be way easier to review and they cover the whole stack from repo/pool/solver to problem output which tends to be very valuable.

src/Composer/DependencyResolver/CooldownPoolFilter.php
Seldaek: For the record, this finding from Claude review also makes me think it'd just be simpler to drop this bypass feature entirely..

## M3. The bypass groups by root requires only, so transitive deps collapse into one group

`CooldownPoolFilter::getConstraintParts()` (diff, ~line 212):

```php
private function getConstraintParts(string $packageName, array $requires): array
{
    if (!isset($requires[$packageName])) {
        return [new MatchAllConstraint()];
    }
```

`$request->getRequires()` only holds root/temporary requires. For a transitive dependency there is never an entry, so *every* branch of that package collapses into a single group keyed `name:0`, and only the single globally-oldest fix bypasses.

A transitive dependency with separate advisories on `3.x` and `4.x` can therefore end up with the `4.x` fix still withheld while the `3.x` fix bypasses — potentially unsolvable, which is exactly the deadlock the tests set out to prevent. `testOldestSecurityFixPerDisjunctiveConstraintPartBypassesCooldown` only covers the *root* disjunctive case.

Either derive the constraint parts from the package's actual incoming links, or document the limitation explicitly.

doc/03-cli.md
```diff
@@ -1454,6 +1454,20 @@ DEPRECATED, use [COMPOSER_POLICY_ABANDONED_BLOCK](#composer-policy-abandoned-blo
...
+enables the cooldown policy. Package versions newer than the specified age will be
+withheld during `composer update`/`require`. Set to `0` to disable. This overrides only
+the `age` value from [policy.cooldown](06-config.md#cooldown) - any `ignore` rules
+configured in composer.json are still respected. Security fixes (versions released after
+a security advisory that fix the vulnerability) automatically bypass the cooldown.
+
+### COMPOSER_POLICY_COOLDOWN_BLOCK
+
+If set to `1`, forces cooldown blocking on; if set to `0`, disables it. Overrides the
+[policy.cooldown.block](06-config.md#block-2) config option.
```
Seldaek, line 1469: Kinda messy/brittle to deep link to these as they're very redundant I guess, but anyway let's keep it for now and at least fix it to the correct cooldown block (https://github.com/crocodele/composer/blob/ff0a0b67d1e02aa06b1939d70e39109c92870ee5/doc/06-config.md#block-3)

```suggestion
[policy.cooldown.block](06-config.md#block-3) config option.
```

src/Composer/Policy/CooldownPolicyConfig.php
```diff
@@ -0,0 +1,299 @@
...
+        );
+    }
+
+    /**
+     * @param array<string, mixed> $policyConfig
+     */
+    public static function fromRawConfig(array $policyConfig, VersionParser $parser): self
+    {
+        $cooldownConfig = $policyConfig['cooldown'] ?? [];
+        if ($cooldownConfig === false) {
```
Seldaek, line 194: Maybe we should make sure to check `Platform::getEnv('COMPOSER_POLICY_COOLDOWN_AGE')` here (or equivalent check upstream somewhere..), the point is that if cooldown is not enabled but you try to enable it using this env var, it will silently not do anything right now, and I'd argue if we can warn the user about that fact that'd be better. Or even throw 🤷🏻‍♂️

src/Composer/Advisory/Auditor.php
```diff
@@ -199,9 +217,39 @@ public function audit(IOInterface $io, RepositorySet $repoSet, PolicyConfig $pol
             }
         }
 
+        if (count($cooldownPackages) > 0) {
+            $plurality = count($cooldownPackages) === 1 ? '' : 's';
+            $punctuation = $format === self::FORMAT_SUMMARY ? '.' : ':';
+            $style = $cooldownCount > 0 ? 'error' : 'warning';
+
+            $io->writeError(sprintf('<%s>Found %d package%s within the cooldown configured in "policy.cooldown"%s</%s>', $style, count($cooldownPackages), $plurality, $punctuation, $style));
```
Seldaek, line 225: Please switch to write/stdout here as we did this change in the Auditor class a while back

```suggestion
            $io->write(sprintf('<%s>Found %d package%s within the cooldown configured in "policy.cooldown"%s</%s>', $style, count($cooldownPackages), $plurality, $punctuation, $style));
```

src/Composer/Advisory/Auditor.php
```diff
@@ -199,9 +217,39 @@ public function audit(IOInterface $io, RepositorySet $repoSet, PolicyConfig $pol
...
         return $auditResult;
     }
 
+    /**
+     * @param array<string, array{prettyVersion: string, releaseDate: string, availableIn: string, source: string}> $cooldownPackages
+     */
+    private function outputCooldownPackages(IOInterface $io, array $cooldownPackages): void
+    {
+        foreach ($cooldownPackages as $name => $info) {
+            $io->writeError(sprintf(
```
Seldaek, line 240:
```suggestion
            $io->write(sprintf(
```

src/Composer/Advisory/Auditor.php
```diff
@@ -199,9 +217,39 @@ public function audit(IOInterface $io, RepositorySet $repoSet, PolicyConfig $pol
...
+                '  - %s (%s) published %s, available in %s',
+                $name,
+                $info['prettyVersion'],
+                $info['releaseDate'],
+                $info['availableIn']
+            ));
+            if ($info['source'] === 'time') {
+                $io->writeError('    The cooldown used the package-supplied release date because the repository did not provide an authoritative publication timestamp.');
```
Seldaek, line 248:
```suggestion
                $io->write('    The cooldown used the package-supplied release date because the repository did not provide an authoritative publication timestamp.');
```

Seldaek: BTW this warning is really only true for packages coming from composer repositories.. If they come from vcs or other, then the trust model is different and even published-time should not be trusted. Of course usually you would typically use a vcs repo against a repo you kinda trust, or for dev stuff which are anyway bypassing cooldowns, so this is more of a theoretical fact check and I'm not sure we should do anything about it.

res/composer-schema.json
```diff
@@ -692,6 +692,39 @@
...
+                                    "type": "object",
+                                    "description": "Configuration for the cooldown policy, which withholds package versions published more recently than the configured age during update/require.",
+                                    "properties": {
+                                        "age": {
+                                            "description": "Cooldown duration as a string in the form \"<number> <unit>\" where unit is second, minute, hour, day or week (e.g. \"7 days\", \"24 hours\", \"30 minutes\", \"1 week\"), an integer number of seconds, or null to disable. Relative phrases like \"tomorrow\" are not accepted. Defaults to null (no cooldown).",
+                                            "anyOf": [
+                                                { "type": "string" },
+                                                { "type": "integer" },
+                                                { "type": "null" }
+                                            ]
```
Seldaek, line 710: Maybe we can restrict this a bit more?

```suggestion
                                            ],
                                            "pattern": "^\d+\s*(second|minute|hour|day|week)s?$|^\d+$"
```

tests/Composer/Test/DependencyResolver/CooldownPoolFilterTest.php
```diff
@@ -0,0 +1,878 @@
...
+        $oldEnoughPackage = new Package('vendor/pkg', '1.0.0.0', '1.0.0');
+        $oldEnoughPackage->setReleaseDate(new DateTimeImmutable('2026-01-13 12:00:00')); // 2 days old
+
+        $pool = new Pool([$oldEnoughPackage]);
+        $filteredPool = $filter->filter($pool, new Request());
+
+        $this->assertSame([$oldEnoughPackage], $filteredPool->getPackages());
+    }
+
+    public function testExactlyAtCutoffIsFiltered(): void
```
Seldaek, line 223: Test name seems to not match the actual test?

```suggestion
    public function testExactlyAtCutoffIsNotFiltered(): void
```

tests/Composer/Test/DependencyResolver/CooldownPoolFilterTest.php
```diff
@@ -0,0 +1,878 @@
...
+    public function testPackageWithMultipleNamesTrackedCorrectly(): void
+    {
+        $config = new CooldownPolicyConfig(true, ListPolicyConfig::AUDIT_IGNORE, [], 7 * 24 * 3600); // 7 days
+        $now = new DateTimeImmutable('2026-01-15 12:00:00');
+        $filter = new CooldownPoolFilter($config, $now);
+
+        // Create a package with provides (will have multiple names)
+        $newPackage = new Package('vendor/pkg', '2.0.0.0', '2.0.0');
+        $newPackage->setReleaseDate(new DateTimeImmutable('2026-01-14 12:00:00'));
+        $newPackage->setProvides([
```
Seldaek, line 857: Maybe using replace here would yield a more interesting test as replaced names are returned by getNames always, but you gotta check what happens..

```suggestion
        $newPackage->setReplaces([
```

xavierleune:
hi @crocodele thanks a lot for this pull request ! This will be a great addition to composer and will help improve the whole ecosystem 🙏 
Do you think you'll have time to integrate the feedbacks from @Seldaek ? I know this pull request is open since a while now and it may be hard to follow up after so long. If you've moved to something else, please let us know so we (like someone from the community, I'm not in composer team) may try to move your pr forward.
Thanks a lot for the time you've already invested in this !

Seldaek:
I'll try to take the changes over myself next week if @crocodele doesn't have time until then, because indeed it'd be good to move forward with it.

crocodele:
@xavierleune :

> Do you think you'll have time to integrate the feedbacks from @Seldaek ?

It's been busy times lately, but I'll try to address the comments over the next couple of days.

PR #12698, opened 2025-12-30
Add `source-fallback` config option to disable/enable fallback on download failure

### LINKED ISSUE #4591: Add a `--force-dist` option to avoid cloning

kayw-geek:
### Background

By default, Composer silently falls back to cloning from source when dist downloads fail. While convenient, this behavior can be problematic in production/CI environments where:

  - Unpredictable build artifacts: Source installations include `.git` directories and other VCS files, causing deployments to fail or behave unexpectedly
  - Silent failures: Users have no way to detect when fallback occurs, making issues hard to diagnose

  ### Usage

  #### Config file:
```json
  {
      "config": {
          "source-fallback": false
      }
  }
```

  #### Command line:
  `composer install --no-source-fallback`
  `composer update --source-fallback`

  #### Environment variable:
  `COMPOSER_SOURCE_FALLBACK=0 composer install`

  ### Behavior

  - true (default): Current behavior - fallback to alternative source on failure
  - false: Fail immediately if preferred source is unavailable

  When disabled and a download fails, Composer will display a clear error message explaining that fallback is disabled, rather than silently trying alternatives.

  ### Changes

  - Added source-fallback config option (defaults to true)
  - Added --[no-]source-fallback flag to install, update, require, reinstall, and create-project commands
  - Added environment variable support (COMPOSER_SOURCE_FALLBACK)
  - Added tests for config and download manager behavior

  Fixes #4591

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -104,6 +104,9 @@ resolution.
   To get the legacy behavior where Composer use `source` automatically for dev
   versions of packages, use `--prefer-install=auto`. See also [config.preferred-install](06-config.md#preferred-install).
   Passing this flag will override the config value.
+* **--source-fallback / --no-source-fallback:** Override the [config.source-fallback](06-config.md#source-fallback)
+  setting. When disabled, Composer will not fall back to an alternative download source
+  (e.g., from dist to source or vice versa) if the preferred one fails. Also see [COMPOSER_SOURCE_FALLBACK](#composer-source-fallback).
 * **--dry-run:** If you want to run through an installation without actually
   installing a package, you can use `--dry-run`. This will simulate the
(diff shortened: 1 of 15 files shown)


Seldaek:
Thanks!

Seldaek:
Just as an FYI we have decided to take further steps in this direction https://github.com/composer/composer/pull/12885

PR #12699, opened 2025-12-30
Disable HTTP/3 when HTTP proxy is used

Elnadrion:
HTTP/3 is not supported over an HTTP proxy. This change adds a check to detect when a proxy will be used for a request and falls back to HTTP/2 instead of attempting to use HTTP/3, preventing connection errors.

image


Seldaek:
HTTP/3 support for proxies seems possible, so ideally we shouldn't have to do this, but I guess it's too early so probably rather keep this off for now..

Issue #12703, opened 2026-01-02
composer show --no-interaction fails with preg_replace() UTF-8 encoding error after upgrading to Composer 2.9.3


runphp:
After upgrading Composer to version 2.9.3 (mainline patched version for CVE-2025-67746), running composer show --no-interaction results in a preg_replace() error related to malformed UTF-8 characters. The error occurs when executing a regular expression pattern designed to match ANSI control sequences, which appears to be part of the security patch addressing ANSI sequence injection vulnerabilities.

### Error Details
```sh
In PcreException.php line 29:
                                                                                                                  
  preg_replace(): failed executing "{\x1B\[[\x30-\x3F]*[\x20-\x2F]*[\x40-\x7E]|\x1B\].*?(?:\x1B\\|\x07)|\x1B.|[-	
                                                                                                                       
  
(?!                                                                                                          
  )}u": Malformed UTF-8 characters, possibly incorrectly encoded      
```

runphp:
The composer show --no-interaction command can generate invalid UTF-8 sequences when processing package descriptions containing multi-byte characters (like Chinese characters), causing subsequent regex operations with the /u modifier to fail.
https://github.com/composer/composer/blob/ffd56dc36571769295ed4fbf1861be838b4620e3/src/Composer/Command/ShowCommand.php#L751
 
The substr() function operates on bytes, not characters. When it truncates multi-byte UTF-8 characters (like Chinese characters which are 3 bytes each), it can cut in the middle of a character sequence, creating incomplete UTF-8 byte sequences.
For example, if a Chinese character has the byte sequence E4BDA0 (你) and substr cuts after the first two bytes E4BD, it creates an invalid UTF-8 sequence that cannot be processed by regex operations with the /u modifier.

Impact: This causes errors like:

preg_replace(): failed executing "{\x1B\[[\x30-\x3F]*[\x20-\x2F]*[\x40-\x7E]|\x1B\].*?(?:\x1B\\|\x07)|\x1B.|[-\x00-\x1F\x7F-\x9F]}u": Malformed UTF-8 characters, possibly incorrectly encoded

Solution: Replace substr() with mb_substr() to ensure character-level truncation:

if (mb_strlen($description, 'UTF-8') > $remaining) {
    $description = mb_substr($description, 0, $remaining - 3, 'UTF-8') . '...';
}

This ensures that multi-byte characters are not split in the middle, preserving valid UTF-8 sequences for subsequent processing.

PR #12704, opened 2026-01-02
Replace substr with mb_substr for UTF-8 support

runphp:
For bug https://github.com/composer/composer/issues/12703


Seldaek:
Thanks!

Issue #12721, opened 2026-01-16
Local package installation quality of life


innocenzi:
**Is your feature request related to a problem? Please describe.**

When working on PHP packages, it's often useful to have it symlinked to an existing application, for testing/playground purposes.

Doing so by manually adding a `path` repository to `composer.json` is tedious.

**Describe the solution you'd like**

I would like the following CLI command to work:
```sh
composer require /path/to/local/package
```

This would automatically add a `repositories` entry with the `type` set to `path` and the correct `url`, as well as add the package to the list of `require` (or adapt the version constraint if it's already there).

PR #12723, opened 2026-01-18
composer require now automatically adds local repos

discollizard:
Good evening. Fixes https://github.com/composer/composer/issues/12721

### DIFF
--- a/src/Composer/Command/PackageDiscoveryTrait.php
+++ b/src/Composer/Command/PackageDiscoveryTrait.php
@@ -29,6 +29,7 @@ use Composer\Repository\ConfigurableRepositoryInterface;
 use Composer\Repository\PathRepository;
 use Composer\Repository\PlatformRepository;
 use Composer\Repository\RepositoryFactory;
+use Composer\Repository\RepositoryInterface;
 use Composer\Repository\RepositorySet;
 use Composer\Semver\Constraint\Constraint;
 use Composer\Util\Filesystem;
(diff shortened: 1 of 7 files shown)


src/Composer/Command/PackageDiscoveryTrait.php
```diff
@@ -106,22 +110,65 @@ final protected function determineRequirements(InputInterface $input, OutputInte
...
-                        ));
+                foreach($repos as $repo){
+                    if (!$repo instanceof ConfigurableRepositoryInterface) {
+                        continue;
+                    }
+                    $repoConfig = $repo->getRepoConfig();
+                    $repoURLs[] = $repoConfig['url'];
+                }
+
+                $isLocalPackage = $requirement['name'][0] === '/'
```
stof: this won't detect absolute paths on Windows.

Seldaek: You can use Filesystem::isAbsolutePath for that https://github.com/composer/composer/blob/main/src/Composer/Util/Filesystem.php#L567

src/Composer/Command/PackageDiscoveryTrait.php
```diff
@@ -106,22 +110,66 @@ final protected function determineRequirements(InputInterface $input, OutputInte
...
+
+                    if(!$alreadyExists){
+                        $configSource = new JsonConfigSource(new JsonFile('composer.json'));
+
+                        $configSource->addRepository($requirement['name'], [
+                            'type' => 'path',
+                            'url' => $treatedPath,
+                            'options' => [
+                                'symlink' => true
+                            ]
```
Seldaek: This probably should be left to the default value (i.e. delete these lines)

src/Composer/Command/PackageDiscoveryTrait.php
```diff
@@ -106,22 +110,66 @@ final protected function determineRequirements(InputInterface $input, OutputInte
...
+                $fs = new Filesystem();
+                $isLocalPackage = $fs->isAbsolutePath($requirement['name'])
+                || $requirement['name'][0] === '.';
+
+                $alreadyExists = in_array($requirement['name'], $repoURLs, true);
+                $treatedPath = str_replace("\\", '/', $requirement['name']);
+
+                if($isLocalPackage){
+                    $jsonFileInstance = new JsonFile($treatedPath.'/composer.json');
+                    $requirement['name'] = $jsonFileInstance->read()['name'];
```
Seldaek: Should first check if $jsonFile->exists() and error clearly if not.

Same for the name key it might not be present and we should check first

src/Composer/Command/PackageDiscoveryTrait.php
```diff
@@ -106,22 +110,66 @@ final protected function determineRequirements(InputInterface $input, OutputInte
...
+
+                $fs = new Filesystem();
+                $isLocalPackage = $fs->isAbsolutePath($requirement['name'])
+                || $requirement['name'][0] === '.';
+
+                $alreadyExists = in_array($requirement['name'], $repoURLs, true);
+                $treatedPath = str_replace("\\", '/', $requirement['name']);
+
+                if($isLocalPackage){
+                    $jsonFileInstance = new JsonFile($treatedPath.'/composer.json');
```
Seldaek: I'd use treatedPath here, and check already if it exists, then the nested if below at line 140 can be removed

```suggestion
                $treatedPath = str_replace("\\", '/', $requirement['name']);

                if ($isLocalPackage && !in_array($treatedPath, $repoURLs, true)) {
                    $jsonFileInstance = new JsonFile($treatedPath.'/composer.json');
```

discollizard: but if i require the same package twice, it will exit out like it doesn't recognize the package, even though it's already in the require object.

image


 from what I've gathered:
- it needs to be checked once to see if the package is local, 
- if so, it gets assign the proper name, so it doesn't import the same package twice nor crashes
- then, if the treatedPath is not present at the current repoURLs, it gets added.
- otherwise nothing happens

correct me if I'm wrong 🙏 

(screenshot from running require on a project that already has the package in the require object. it updates successfully with the nested logic)

image

src/Composer/Command/PackageDiscoveryTrait.php
discollizard: Done ✅

src/Composer/Command/PackageDiscoveryTrait.php
discollizard: Done ✅

src/Composer/Command/PackageDiscoveryTrait.php
```diff
@@ -106,22 +110,73 @@ final protected function determineRequirements(InputInterface $input, OutputInte
...
-                            'Using version <info>%s</info> for <info>%s</info>',
-                            $requirement['version'],
-                            $requirement['name']
-                        ));
+                foreach($repos as $repo){
+                    if (!$repo instanceof ConfigurableRepositoryInterface) {
+                        continue;
+                    }
+                    $repoConfig = $repo->getRepoConfig();
+                    $repoURLs[] = $repoConfig['url'];
```
Seldaek: I'd also normalize there to make sure we find/avoid duplicates even if the composer.json contains a path with backslashes 

```suggestion
                    $repoURLs[] = str_replace("\\", '/', $repoConfig['url']);
```

src/Composer/Command/PackageDiscoveryTrait.php
```diff
@@ -106,22 +110,73 @@ final protected function determineRequirements(InputInterface $input, OutputInte
...
+                        $configSource->addRepository($requirement['name'], [
+                            'type' => 'path',
+                            'url' => $treatedPath
+                        ], true);
+                    }
+                }
+
+                if (!isset($requirement['version'])) {
+                    if($isLocalPackage){
+                        $requirement['version'] = 'dev-main';
```
Seldaek: This probably should be `*` no? Otherwise if the path repo has a version set this won't work.

src/Composer/Command/PackageDiscoveryTrait.php
```diff
@@ -106,22 +110,73 @@ final protected function determineRequirements(InputInterface $input, OutputInte
...
+                    if (!$jsonFileInstance->exists()) {
+                        throw new \Exception($treatedPath . " was not found");
+                    }
+
+                    $jsonConfigFromPackage = $jsonFileInstance->read();
+                    if (!isset($jsonConfigFromPackage['name'])) {
+                        throw new \Exception("Package name inside composer.json at " . $treatedPath . " is not set");
+                    }
+
+                    $requirement['name'] = $jsonConfigFromPackage['name'];
```
Seldaek:
```suggestion
                    $requirement['name'] = $jsonConfigFromPackage['name'];
                    if (isset($jsonConfigFromPackage['version'])) {
                        $requirement['version'] = $jsonConfigFromPackage['version'];
                    }
```

src/Composer/Command/PackageDiscoveryTrait.php
stof: I would say it should probably use the existing logic that is either guessing the version after resolving dependencies or finding a best version.

stof:
I'm not sure this should be added directly to `require` without any dedicated flag to opt-in. Package names are easily confused with relative paths.

Seldaek:
> I'm not sure this should be added directly to `require` without any dedicated flag to opt-in. Package names are easily confused with relative paths.

Having a real package name that also has a valid local path sounds way less likely, but it anyway forces that the path starts with `.` (which is not allowed for packagist.org package names) or is an absolute path, so I think we're safe.

BrianHenryIE:
This is nice. I tried something similar but TBH I don't use it a whole lot: [composer-prefer-local](https://github.com/BrianHenryIE/composer-prefer-local)

> When a package is available in a sibling or sub-directory, use that instead of Packagist or other configured repos.

PR #12727, opened 2026-01-26
Add support for PIE list of download-url-methods

asgrim:
New upcoming feature in PIE needs `php-ext.download-url-method` to support a list of strings. For backwards compatibility (and simplicity, really), a string is still supported


asgrim:
Thank you! :)

PR #12733, opened 2026-02-01
Bump only actually updated packages when running `composer update --bump-after-update`

kubawerlos:
Following the suggestion from https://github.com/composer/composer/pull/12556#discussion_r2407088799

I had not possibily to reopen the other one, because branch `2.8` no longer exists (at least I think it is the reason).

### DIFF
--- a/src/Composer/Command/UpdateCommand.php
+++ b/src/Composer/Command/UpdateCommand.php
@@ -298,17 +298,19 @@ EOT
                     }
                 }
 
-                $io->writeError('<info>Bumping dependencies</info>');
-                $bumpCommand = new BumpCommand();
-                $bumpCommand->setComposer($composer);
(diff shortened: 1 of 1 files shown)


src/Composer/Command/UpdateCommand.php
```diff
@@ -292,7 +306,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                     $bumpAfterUpdate === 'dev',
                     $bumpAfterUpdate === 'no-dev',
                     $input->getOption('dry-run'),
-                    $input->getArgument('packages'),
+                    $updatedPackages,
                     '--bump-after-update=dev'
                 );
```
Seldaek: This whole thing probably should be wrapped in a `if (\count($updatedPackages) > 0) { ... }` to avoid running bump for nothing?

PR #12735, opened 2026-02-03
Parse HTML in extension info when not on CLI

Toflar:
Non-issue when running Composer on CLI like everybody does 🤣 

As you know, we built a GUI on top of Composer with our Contao Manager. We noticed that when instantiating `new PlatformRepository()` in the web, most of the `lib-*` packages were missing.

The reason for it being that 

```php
$reflector = new \ReflectionExtension($extension);
reflector->info();
```

outputs different information depending on the SAPI. If you are running on `cli`, it will be just text. In `fpm` for example, it will be HTML.
Here are two examples using `pdo_sqlite`:

### CLI output

```
pdo_sqlite

PDO Driver for SQLite 3.x => enabled
SQLite Library => 3.40.1
```

### fpm output

```html
<h2><a name="module_pdo_sqlite" href="#module_pdo_sqlite">pdo_sqlite</a></h2>
<table>
<tr><td class="e">PDO Driver for SQLite 3.x </td><td class="v">enabled </td></tr>
<tr><td class="e">SQLite Library </td><td class="v">3.40.1 </td></tr>
</table>
```

Hence, the regular expressions do not match at all anymore and half of the `lib-*` packages are missing.
A simple `strip_tags()` call would fix this for probably most of the cases.

### DIFF
--- a/src/Composer/Platform/Runtime.php
+++ b/src/Composer/Platform/Runtime.php
@@ -112,6 +112,9 @@ class Runtime
         return self::parseHtmlExtensionInfo($info);
     }
 
+    /**
+     * @internal Only public for unit tests
+     */
     public static function parseHtmlExtensionInfo(string $html): string
     {
         $result = [];


Toflar:
Hmm, as pointed out by @leofeyer, this won't do it because the regular expressions very often match for `=>`.
Any other idea how to fix this @Seldaek?

stof:
Using `html_entity_decode` (after stripping tags) might work.

stof:
ah no, the issue is not that the `>` is encoded, but that the format is totally different in HTML.

Toflar:
Yes, it's not totally random. The HTML format is [actually hard-coded](https://github.com/php/php-src/blob/4188c3ee2c0c3fe8d117e11b8963ac5c955f243e/ext/standard/info.c#L133-L163) so we could address this.
But yeah, not sure if this is a problem Composer has to worry about...

Toflar:
I've created a little helper function which ensures compatibility for at least our issue. There might be others but it should actually do a decent job for most of the php info logic because the format is hard-coded in the php source.
And for all the users using Composer on CLI, everything remains untouched anyway.

Not sure about the coding style (aka phpstan) but I'll leave that to Jordi 😊

src/Composer/Platform/Runtime.php
```diff
@@ -101,6 +103,34 @@ public function getExtensionInfo(string $extension): string
...
+        $info = (string) ob_get_clean();
+
+        if ('cli' === PHP_SAPI) {
+            return $info;
+        }
+
+        return self::parseHtmlExtensionInfo($info);
+    }
+
+    public static function parseHtmlExtensionInfo(string $html): string
```
stof, line 118: this should not be public IMO. There is no reason to expose that to plugins, having to maintain BC on it.

Toflar: How would you unit test it then?

Toflar: I've added `@internal`, should be enough.

stof: You could use Reflection to call the private method directly for that case

Toflar: Yeah, we could. We'll see.

Seldaek:
Thanks

PR #12737, opened 2026-02-08
Relay GitHub API error messages to the user on auth failures

### LINKED ISSUE #12711: GitHub access token message should give more accurate information on why the existing token failed

I had trouble installing a package, getting a message from Composer like

> ```
> Could not fetch https://api.github.com/repos/orgname/repository/zipball/somesha, please review your configured GitHub OAuth token or enter a new one to access private repos
> You need to provide a GitHub access token.
> [etc]
> ```

But the repo is public! And accessing the provided URL directly, without any access token, gave no problem.

Eventually I managed to find the code in `src/Composer/Util/Http/CurlDownloader.php` where I could var_dump the actual response from GitHub, which told me

> ```
> `orgname` forbids access via a personal access token (classic). Please use a GitHub App, OAuth App, or a personal access token with fine-grained permissions.
> ```

Sigh. Fine. So I switched to a fine-grained token.

Now today, same Composer error message about another public repo from a different organization. I had to find again the place in `src/Composer/Util/Http/CurlDownloader.php` to dump the error message, which this time is

> ```
> The 'otherorgname' organization forbids access via a fine-grained personal access tokens if the token's lifetime is greater than 366 days. Please adjust your token's lifetime at the following URL: ...
> ```

Double sigh.

Composer's error message should relay these sorts of errors to the user, instead of incorrectly complaining about private repos. This will help the user debug such problems.

* I don't think I need to access any private repos via Composer at this time ([good thing](https://github.com/composer/composer/issues/12089#issuecomment-2796544335)). Let's see if I run into GitHub API rate limits while not using a token.
* Various orgs aren't likely to change these annoying and pointless policies affecting access to public repos. They're probably more concerned with access to private repos or from apps with more privileges, and GitHub doesn't seem to allow for separate configuration by public/private or by permission level.
* GitHub seems unlikely to do anything about it either, at least not from only me asking.

**Additional context**

https://docs.github.com/en/organizations/managing-programmatic-access-to-your-organization/setting-a-personal-access-token-policy-for-your-organization

On the plus side, at least the "require admin approval of all fine-grained access tokens" setting doesn't seem to apply to the public data Composer uses.

> Fine-grained personal access tokens will still be able to read public resources within the organization without approval.

QDenka:
## Summary

When a GitHub API request fails with 401/403, Composer now extracts and displays the actual error message from GitHub's JSON response body instead of showing a generic message about reviewing OAuth tokens or accessing private repos.

## Problem

As described in #12711, when GitHub returns specific error messages like:
- `organization forbids access via a personal access token (classic)`
- `organization forbids access via fine-grained personal access tokens if the token's lifetime is greater than 366 days`

Composer would only show a generic message suggesting to review your GitHub OAuth token or create one to access private repos. This made it very difficult to diagnose the actual issue.

## Solution

Added an optional `$responseBody` parameter to `AuthHelper::promptAuthIfNeeded()`. For GitHub domains, when a 401/403 response is received (and it's not a rate limit or SSO issue), the method now attempts to decode the JSON response body and extract the `message` field from GitHub's API response.

The extracted message is then displayed to the user, providing accurate information about why the request failed.

## Changes

- `src/Composer/Util/AuthHelper.php`: Added `$responseBody` parameter and GitHub API error message extraction logic
- `src/Composer/Util/Http/CurlDownloader.php`: Pass response body to `promptAuthIfNeeded()`

Fixes #12711

### DIFF
--- a/src/Composer/Util/AuthHelper.php
+++ b/src/Composer/Util/AuthHelper.php
@@ -81,7 +81,7 @@ class AuthHelper
      *                                retried, if storeAuth is true then on a successful retry the authentication should be persisted to auth.json
      * @phpstan-return array{retry: bool, storeAuth: 'prompt'|bool}
      */
-    public function promptAuthIfNeeded(string $url, string $origin, int $statusCode, ?string $reason = null, array $headers = [], int $retryCount = 0): array
+    public function promptAuthIfNeeded(string $url, string $origin, int $statusCode, ?string $reason = null, array $headers = [], int $retryCount = 0, ?string $responseBody = null): array
     {
         $storeAuth = false;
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

Issue #12370, opened 2025-04-08
composer create-project allow requiring additional packages


shyim:
**Is your feature request related to a problem? Please describe.**
Right now you can create a new project using `composer create-project xxx` to install an additional package, you need to cd into that folder and run composer req

**Describe the solution you'd like**
It would be nice to provide in documentation a one-liner to setup a project + additional packages like

```
# local setup
composer create-project shopware/production

# docker setup
composer create-project shopware/production --require 'shopware/docker-local:*'
```

It would just add that package to the root composer.json before it runs `composer install`

**Describe alternatives you've considered**
curl + bash script into docs as a one liner, composer create with an flag would be more trustworhy

**Additional context**

Seldaek:
This probably could work similar to the --add-repository flag, which deletes the composer.lock file  if it is given, thus forcing an update. Anything more  complex than that would require running a full require command within the scope of create-project, which to me sounds like too much, and then it should rather be written as multiple commands like you would do now.

discollizard:
Good evening. I intend to try and implement this feature.

Seldaek:
Fixed by #12738

PR #12738, opened 2026-02-13
Add new --require param to create-project to add new packages to the project as it gets installed

discollizard:
This PR concerns the following issue: https://github.com/composer/composer/issues/12370

### DIFF
--- a/src/Composer/Command/CreateProjectCommand.php
+++ b/src/Composer/Command/CreateProjectCommand.php
@@ -177,6 +177,7 @@ EOT
 
     /**
      * @param string|array<string>|null $repositories
+     * @param array<string> $requiredPackages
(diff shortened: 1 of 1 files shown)


src/Composer/Command/CreateProjectCommand.php
```diff
@@ -158,6 +160,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             $input->getArgument('directory'),
             $input->getArgument('version'),
             $input->getOption('stability'),
+            $input->getOption('require'),
```
Seldaek: Please add this new arg to the end to avoid any BC breaks

Seldaek:
Thanks!

PR #12743, opened 2026-02-19
Improving the error message when package version can't be parsed

mamazu:
## The problem

In the current version of composer if you sepecify an invalid version string you get the following error message:
```text
Invalid version string "AA"
```

This is not very helpful since it doesn't contain the name of the package which contains the faulty package definition.

## The fix

```text
Failed to normalize version for package "acme/package". Invalid version string "AA"
```

I don't know if the place I put the error handling in is the correct place. So please feel free to guide me to where to best put it.

### DIFF
--- a/tests/Composer/Test/Package/Loader/ArrayLoaderTest.php
+++ b/tests/Composer/Test/Package/Loader/ArrayLoaderTest.php
@@ -510,7 +510,7 @@ class ArrayLoaderTest extends TestCase
         ];
 
         $this->expectException(\UnexpectedValueException::class);
-        $this->expectExceptionMessage('Failed to normalize version for package "acme/package". Invalid version string "AA"');
+        $this->expectExceptionMessage('Failed to normalize version for package "acme/package": Invalid version string "AA"');
         $this->loader->load($config);
     }
 }


src/Composer/Package/Loader/ArrayLoader.php
```diff
@@ -118,16 +118,20 @@ private function createObject(array $config, string $class): CompletePackage
...
+                if ($version === VersionParser::DEFAULT_BRANCH_ALIAS) {
+                    $version = $this->versionParser->normalize($config['version']);
+                }
+            } else {
+                    $version = $this->versionParser->normalize($config['version']);
             }
-        } else {
-            $version = $this->versionParser->normalize($config['version']);
+        } catch (\UnexpectedValueException $e) {
+            throw new \UnexpectedValueException('Failed to normalize version for package "'.$config['name'].'". '.$e->getMessage());
```
stof: I suggest attaching `$e` as previous exception, so that the original stacktrace is not lost. It can help debugging things when working on composer.

mamazu: Yes, that's a very good idea. Did that.

Seldaek:
Can you show me a way to reproduce this "invalid package version"? Where was the faulty version located?

mamazu:
That's a good question, the faulty version wasn't in the main composer file. 
I think it was located in one of the dependency that when composer tried to parse the file, it crashed with this error message.

Seldaek:
@mamazu I would like to understand fully before merging this, because we explicitly try to include the package name in other ways in a few places, and so I'd like to know where this happened.

mamazu:
Of course, I've managed to reproduce it when using `composer validate` on this composer file

```json
{
    "name": "mamazu/main",
    "version": "1.0.0",
    "autoload": {
        "psr-4": {
            "Mamazu\\Mamazu\\": "src/"
        }
    },
    "authors": [],
    "require": {
        "phpunit/phpunit": "AAAA"
    }
}
```

Then the error message should now also show the package name that has the broken version.

Seldaek:
Ok I see. In this case the version field is for the root package which is why it doesn't show a name, but I guess it doesn't hurt to show it as well.

mamazu:
I just noticed that parsing the version of the package gives a different error message and a JsonValidation Exception, do you know why this isn't happening when the invalid version is in the require section?

Seldaek:
Because a version is something we can validate with a regex in the schema, but the constraints are much broader and thus too risky to try and validate it in the json schema

Seldaek:
Thanks

PR #12764, opened 2026-03-14
Enforce allow-plugins in non-interactive mode for pre-2.2 lock file

jorgsowa:
As per the comment https://github.com/composer/composer/pull/10920#issuecomment-1174764857 this was intended to be changed in July 2022.

The silent wildcard allowall `($rules = ['{}' => true])` in `PluginManager::isPluginAllowed()` is replaced with a `\RuntimeException` that tells the user exactly what to do:

>   Your composer.lock was generated before the allow-plugins security feature
>   was introduced, and your composer.json does not define allow-plugins.
>   Run "composer update --lock" locally and commit the updated composer.lock,
>   then add an explicit allow-plugins section to composer.json.
>   See https://getcomposer.org/allow-plugins

### DIFF
--- a/src/Composer/Plugin/PluginManager.php
+++ b/src/Composer/Plugin/PluginManager.php
@@ -718,13 +718,11 @@ class PluginManager
         // an allow-plugins config being present cannot be made.
         if ($rules === null) {
             if (!$this->io->isInteractive()) {
-                $this->io->writeError('<warning>For additional security you should declare the allow-plugins config with a list of packages names that are allowed to run code. See https://getcomposer.org/allow-plugins</warning>');
-                $this->io->writeError('<warning>This warning will become an exception once you run composer update!</warning>');
-
-                $rules = ['{}' => true];
-
-                // if no config is defined we allow all plugins for BC
-                return true;
+                throw new \RuntimeException(
+                    'Your composer.lock was generated before the allow-plugins security feature was introduced and your composer.json does not define allow-plugins. '
+                    . 'Run "composer update --lock" locally and commit the updated composer.lock, then add an explicit allow-plugins section to composer.json. '
+                    . 'See https://getcomposer.org/allow-plugins'
+                );
             }
 
             // keep going and prompt the user


Seldaek:
I'm not so sure about this. In a way it hopefully shouldn't impact many people, but that also makes it less relevant.. So we just risk breaking someone's legacy stuff, and the possible benefit is forcing a few more people to upgrade instead of running any plugins coming their way (which on a legacy project with stale dependencies is unlikely to bring in new plugins really).

So I'd say maybe we rather leave it as is.. But pinging @naderman to see if he has another opinion.

naderman:
I can't come up with an actual attack scenario this legacy support would enable. At the same time K think it is at least extremely unintuitive that the allow plugins mechanism can be circumvented with a lock file not containing the setting at all. 

If we consider allow-plugins a necessary security mechanism, it shouldn't be possible to disable it without intent. However if you can control the lock file to attack someone, then you could just as well add something to allow plugins that wasn't meant to be there. So this clearly falls outside the threat model we're working with.

So find this a tricky situation. This change will break running composer on historic lock files. It does not address a security concern within our threat model. This change makes Composer behave as I would expect it to, considering what the allow plugins setting promises.

So I think overall I'd lean to agree with Jordi. The only real alternative to me would be to add another option to install to allow the use of such a legacy lock file, but I think that'd be used so infrequently and just complicate the CLI interface that I wouldn't do that either .

Seldaek:
Yeah I find the attack vector slim but hadn't considered that angle, so I'd perhaps rather merge the PR to close that gap, at the risk of annoying someone running install on a very old project.

Seldaek:
Alright let's merge this to remove some potential for bad surprises.. Hopefully it won't cause too much pain for people having very old lock files still around.

Quezler:
Well it is, ci for an old project (laravel 5.3 / php 7.4) started failing and due to it's age it only still kinda works in the production environment since trying to update it on modern dev machines just causes the different php/composer versions to yell at eachother so now i have a fun time figuring out how to get past this in a way that will not break it.  😂

Seldaek:
You need to add `"config": {"allow-plugins": {"plugin/name": true|false}}` with the name of every plugin to your composer.json and it should proceed without problem AFAIK.

Quezler:
Too bad the error does not list which plugin(s) triggered it, but a grep on the lock file suggests it is (just) `composer-plugin-api`

Oh i guess it does, but not until that section is present at all, gotta add `kylekatarnls/update-helper` too it seems.

Alright the final tally, this causes ci to work again without editing the lock file (weird, since that error was so specific), no idea what either of these plugins do but i guess they had not been running since 2022 from what i could gather?

```diff
    "config": {
-        "preferred-install": "dist"
+        "preferred-install": "dist",
+         "allow-plugins": {
+            "composer-plugin-api": true,
+            "kylekatarnls/update-helper": true
+        }
    }
```

Seldaek:
Yeah they were allowed to run with a simple warning before. And now we just block it if you don't specify anything. So setting them to `true` is the correct BC way. `composer-plugin-api` does not need to be listed there tho that doesn't do anything.

Quezler:
Sweet, thanks for helping me resolve this without the need to update the composer.lock, i hope some others running into this issue will find their way here and be equally as helped with it. 🙏

PR #12766, opened 2026-03-16
Dependency Policies (originally Filter lists): to block/audit package versions affected by malware

glaubinix:
This pull request introduces a new filter lists feature. Similar to security advisories, filters can be used to remove package versions from the pool during a `composer update` and get reported as part of a `composer audit`. This integrates with [malware flagging in packagist.org](https://github.com/composer/packagist/pull/1681) which uses the malware feed provided by Aikido. The feature is on by default.

### Example composer.json to try out the feature

This uses hardcoded package and filter definition, this is not how you would configure any project in practice!

```json
{
    "name": "acme/project",
    "version": "1.0.0",
    "require": {
        "acme/library": "*"
    },
    "config": {
        "filter": true
    },
    "repositories": [
        {
            "type": "package",
            "package": [
                {
                    "name": "acme/library",
                    "version": "1.0.0",
                    "type": "metapackage"
                }
            ],
            "filter": {
                "acme": [
                    {
                        "package": "acme/library",
                        "constraint": "*",
                        "url": "https://example.org/malware/acme/library",
                        "reason": "This package is malware"
                    }
                ]
            }
        }
    ]
}
```

and run:
```
composer update
```

### Concepts

* **Sources**: a data source where list information is fetched from. By default, this includes all ComposerRepositories. Additional sources can be configured via filter config.
* **Lists**: a set of filters e.g. aikido-malware. A ComposerRepository can return multiple lists, additionally configured sources can return a single list

### Basic config

Users can disable filtering by setting the config value to `false`
```json
{
  "config": {
    "filter": false
  }
}
```

See changes to 06-config.md for full config options.

### CI helpers

The feature can be turned on/off via an environment variable

```bash
COMPOSER_FILTER=[0|1]
```

### Information for Composer repositories

A Composer repository can advertize that it supports filtering by setting filter.metadata to true in the packages.json response. Additionally, it is required to advertize which lists are available and which lists are default lists. `default-lists` are list used when filtering is enabled without any configuration. Users can then use any additional lists from the `lists` array if they want to

```json
{
  "filter": {
    "metadata": true,
    "lists": ["acme"],
    "default-lists": ["acme"]
}
```

Metadata returned via the `metadata-url` can contain a new "filter" key with lists of entries matching the requested package. An example can be found here: https://repo.packagist.org/p2/techghoshal/my-library.json

```json
{
  "filter": {
    "list-one": [{
      "constraint": "1.0 || 1.1 || 1.2", 
      "url": "https://packagist.org/packages/acne/package/filters" // optional
      "reason": "Information why the package versions are filtered can be provided here",
      "id": "optional identifier, currently unused but might be used later to ignore individual reports"
    }],
    "list-two": []
  }
}
```

See changes to 05-repositories.md for full config options.

### Information for third-party source providers

Additional sources can be configured
```bash
composer config filter.sources.your-name url https://example.org
```
or by editing the composer.json

```json
{
  "filter": {
    "sources": {"name": {
      "type": "url",
      "url": "https://example.org/php/filters",
    }}
  }
}
```

Composer will make a POST request to the configured URL, sending a list of PURLs (package URLs). 

```json
{
  "packages": ["pkg://composer/acme/package", "pkg://composer/acme/library"]
}
```

Respond with a list of filters matching those packages.
```json
{
  "filter": [{
      "package": "acme/package",
      "constraint": "1.0 || 1.1 || 1.2", 
      "url": "https://packagist.org/packages/acme/package/filters" // optional
      "reason": "Information why the package versions are filtered can be provided here",
      "id": "optional identifier, currently unused but might be used later to ignore individual reports"
    }]
}
```


src/Composer/Advisory/AuditConfig.php
```diff
@@ -35,6 +35,11 @@ class AuditConfig
      */
     public $auditAbandoned;
 
+    /**
+     * @var Auditor::FILTERED_*
+     */
+    public $auditFiltered;
```
glaubinix, line 41: Should this go here or into the filter config?

doc/03-cli.md
```diff
@@ -1125,13 +1125,17 @@ vulnerability advisories using the [Packagist.org api](https://packagist.org/api
...
 the findings:
 
 * `0` No issues;
 * `1` Vulnerable packages;
 * `2` Abandoned packages;
 * `3` Vulnerable and abandoned packages.
+* `4` Filtered packages.
+* `5` Vulnerable and filtered packages.
+* `6` Abandoned and filtered packages.
+* `7` Vulneable, abandoned, and filtered packages.
```
```suggestion
* `7` Vulnerable, abandoned, and filtered packages.
```

src/Composer/DependencyResolver/Problem.php
```diff
@@ -417,6 +417,12 @@ public static function getMissingPackageReason(RepositorySet $repositorySet, Req
                 return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but these were not loaded, because they are affected by security advisories ("' . implode('", "', $advisoriesList). '"). Go to https://packagist.org/security-advisories/ to find advisory details. To ignore the advisories, add them to the audit "ignore" config. To turn the feature off entirely, you can set "block-insecure" to false in your "audit" config.'];
             }
 
+            if ($pool->isFilterListRemovedPackageVersion($packageName, $constraint)) {
+                $filters = $pool->getFilterListEntryForPackageVersion($packageName, $constraint);
+
+                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but these were not loaded, because they were ' . implode(', ', $filters). '. To ignore filters for this package, add the package to the "filter.don-filter-packages" config. To turn the feature off entirely, you can set "filter" to false.'];
```
Seldaek:
```suggestion
                return ["- Root composer.json requires $packageName".self::constraintToText($constraint) . ', ', 'found '.self::getPackageList($packages, $isVerbose, $pool, $constraint).' but these were not loaded, because they were ' . implode(', ', $filters). '. To ignore filters for this package, add the package to the "filter.dont-filter-packages" config. To turn the feature off entirely, you can set "filter" to false.'];
```

tests/Composer/Test/Fixtures/installer/update-filter-entry-matching-direct-dependency.test
```diff
@@ -0,0 +1,50 @@
...
+--EXPECT-EXIT-CODE--
+2
+
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
+    - Root composer.json requires acme/library 1.0.0 (exact version match: 1.0.0 or 1.0.0.0), found acme/library[1.0.0] but these were not loaded, because they were flagged as malware by test-list (see https://example.org/malware/acme/library). To ignore filters for this package, add the package to the "filter.don-filter-packages" config. To turn the feature off entirely, you can set "filter" to false.
```
Seldaek:
```suggestion
    - Root composer.json requires acme/library 1.0.0 (exact version match: 1.0.0 or 1.0.0.0), found acme/library[1.0.0] but these were not loaded, because they were flagged as malware by test-list (see https://example.org/malware/acme/library). To ignore filters for this package, add the package to the "filter.dont-filter-packages" config. To turn the feature off entirely, you can set "filter" to false.
```

src/Composer/FilterList/FitlerListProvider/FilterListProviderSet.php
```diff
@@ -0,0 +1,138 @@
...
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\FilterList\FitlerListProvider;
```
Seldaek:
```suggestion
namespace Composer\FilterList\FilterListProvider;
```

Filename (well the directory) needs a rename too.

src/Composer/FilterList/FitlerListProvider/UrlSourceFilterListProvider.php
```diff
@@ -0,0 +1,84 @@
...
+ * This file is part of Composer.
+ *
+ * (c) Nils Adermann <naderman@naderman.de>
+ *     Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Composer\FilterList\FitlerListProvider;
```
Seldaek:
```suggestion
namespace Composer\FilterList\FilterListProvider;
```

src/Composer/Repository/ComposerRepository.php
```diff
@@ -733,6 +736,62 @@ static function ($data) use ($name, $create) {
...
+            if (PlatformRepository::isPlatformPackage($name) || '__root__' === $name) {
+                continue;
+            }
+
+            $promises[] = $this->startCachedAsyncDownload($name, $name)
+                ->then(static function (array $spec) use (&$filter, $name, $parser): void {
+                    [$response] = $spec;
+
+                    if (isset($response['filter']) && is_array($response['filter'])) {
+                        foreach ($response['filter'] as $listName => $data) {
```
Seldaek: Should this not foreach on $data? Like..
```suggestion
                        foreach ($response['filter'] as $listName => $listEntries) {
                            foreach ($listEntries as $data) {
                             // current foreach body here...
```

src/Composer/Advisory/Auditor.php
```diff
@@ -72,7 +88,7 @@ class Auditor
      * @return int-mask<self::STATUS_*> A bitmask of STATUS_* constants or 0 on success
      * @throws InvalidArgumentException If no packages are passed in
      */
-    public function audit(IOInterface $io, RepositorySet $repoSet, array $packages, string $format, bool $warningOnly = true, array $ignoreList = [], string $abandoned = self::ABANDONED_FAIL, array $ignoredSeverities = [], bool $ignoreUnreachable = false, array $ignoreAbandoned = []): int
+    public function audit(IOInterface $io, RepositorySet $repoSet, array $packages, string $format, bool $warningOnly = true, array $ignoreList = [], string $abandoned = self::ABANDONED_FAIL, array $ignoredSeverities = [], bool $ignoreUnreachable = false, array $ignoreAbandoned = [], string $filtered = self::FILTERED_FAIL, ?FilterListProviderSet $filterListProviderSet = null, ?FilterListConfig $filterListConfig = null): int
```
Seldaek, line 92: Can we add a doctype for the $filtered type with `self::FILTERED_*` ?

doc/06-config.md
```diff
@@ -362,6 +384,170 @@ version blocking is not disabled by setting [`block-insecure`](#block-insecure)
...
+            "exclude-lists": ["list-name-to-skip"]
+        }
+    }
+}
+```
+
+### sources
+
+Additional sources to fetch filter list data from. Keys are source names, values are objects with `type` and `url`.
+Currently only `type: "url"` is supported.
```
Seldaek: We probably should say that this URL will receive your full list of installed dependencies.

src/Composer/Advisory/AuditConfig.php
Seldaek: I think here is ok.

doc/06-config.md
glaubinix: Good point! Added. Nils also mentioned that it would be great to have some kind of meta issue for filters where people (users, source providers, Composer repositories) can look up how to implement/use this.

doc/03-cli.md
```diff
@@ -1362,6 +1369,11 @@ If set to `1`, it is the equivalent of passing the `--no-audit` option to a `req
 Set to `ignore`, `report` or `fail` to override the [audit.abandoned](06-config.md#abandoned)
 config option.
 
+### COMPOSER_FILTER
+
+Set to `0` to disable filter lists, or `1` to enable them. Setting this to `1` will use the filter configuration in the composer.json.
```
```suggestion
Set to `0` to disable filter lists. Setting this to `1` will use the filter configuration in the composer.json.
```

Not sure I'd mention enabling, because it's not enabling anything that isn't enabled? Or is setting 1 if config is false really enabling? Because that's not fully clear from the text I guess

doc/06-config.md
```diff
@@ -362,6 +384,171 @@ version blocking is not disabled by setting [`block-insecure`](#block-insecure)
...
+
+A list of packages to exempt from filtering. Supports three formats:
+
+#### Simple format (exempt entirely):
+
+```json
+{
+    "config": {
+        "filter": {
+            "dont-filter-packages": ["vendor/package", "acme/*"]
```
Seldaek: Wonder if we should call this ignore-packages? `filter` doesn't really need to be in the name as it is nested under `filter`. And the `don't` without `'` reads kinda weird.

doc/03-cli.md
glaubinix: Having no filter config means the feature is off. Running it with `COMPOSER_FILTER=1` will then enable the feature.

doc/06-config.md
glaubinix: I had this first but @naderman mentioned that calling it `ignore-packages` is unclear too in the context of filters, saying it could mean filter out or ignore it from the filter list.

Seldaek: I mean the filter feature filters out things, so if you ignore packages from the filtering you keep them, sounds ok to me. dont-filter-packages reads very strange somehow.

doc/06-config.md
```diff
@@ -362,6 +384,171 @@ version blocking is not disabled by setting [`block-insecure`](#block-insecure)
...
+        }
+    }
+}
+```
+
+### sources
+
+Additional sources to fetch filter list data from. Keys are source names, values are objects with `type` and `url`.
+Currently only `type: "url"` is supported. The configured URL will receive your full list of installed dependencies as
+part of any `composer audit` and the full list of considered packages for a `composer update`.
```
Seldaek: Could maybe mention that by default list data is fetched from the composer repositories configured (if they provide any data)? Otherwise it's kinda unclear that you get anything if you don't specify sources.

src/Composer/Command/AuditCommand.php
```diff
@@ -76,11 +81,20 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+        }
+
         $abandoned = $abandoned ?? $auditConfig->auditAbandoned;
+        $filtered = $filtered ?? $auditConfig->auditFiltered;
 
         $ignoreSeverities = array_merge(array_fill_keys($input->getOption('ignore-severity'), null), $auditConfig->ignoreSeverityForAudit);
         $ignoreUnreachable = $input->getOption('ignore-unreachable') || $auditConfig->ignoreUnreachable;
 
+        $filterListConfig = FilterListConfig::fromConfig($composer->getConfig(), new VersionParser());
+        $filterListProviderSet = $filterListConfig !== null ? FilterListProviderSet::create($filterListConfig, $composer->getRepositoryManager()->getRepositories(), Factory::createHttpDownloader($this->getIO(), $composer->getConfig())) : null;
```
Seldaek: You should replace all the `Factory::createHttpDownloader` calls with `$composer->getLoop()->getHttpDownloader()`

src/Composer/Command/ConfigCommand.php
```diff
@@ -840,6 +841,50 @@ static function ($vals) {
...
+
+            $value = $values;
+            if ($input->getOption('json')) {
+                $value = JsonFile::parseJson($values[0]);
+                if (!is_array($value)) {
+                    throw new \RuntimeException('Expected an array or object for '.$settingKey);
+                }
+            }
+
+            // Allow composer config filter.sources.name url https://example.org
```
Seldaek, line 860: That's neat but probably should be in the docs as an example near the filter.sources docs otherwise not sure anyone figures it out :)

doc/06-config.md
naderman: Happy to think of an alternative, but "ignore-packages" in the filter config sounds to me like a hardcoded list of packages to ignore and filter out, not the opposite, which it actually is here.

naderman:
>  Additionally, it is recommended to advertize which lists and categories are available, making it possible for Composer to skip fetching data.

I think we should simply require this and error if that is not done?

naderman:
PR description:

> Information for Composer repositories

This section doesn't explain where that info goes, I guess packages.json? Should make that clear.

doc/03-cli.md
naderman:
```suggestion
Set to `0` to disable filters on updates and audit, or `1` to enable them. Setting this to `1` will use the filter configuration in the composer.json.
```

There are more filters than just lists now.

doc/03-cli.md
```diff
@@ -1362,6 +1369,11 @@ If set to `1`, it is the equivalent of passing the `--no-audit` option to a `req
...
+If you want to override the config value, use `composer config filter 1` instead.
```
naderman, line 1375: 1 or true?

doc/06-config.md
```diff
@@ -330,6 +330,28 @@ in other places than the `compoder audit` command.
 }
 ```
 
+### filtered
+
+Defaults to `fail`. Defines whether and how audit reports should report filtered packages. There are three possible values:
+
+- `ignore` means audit reports do not consider filtered packages at all.
+- `report` means filtered packages are reported but do not cause the composer audit command return a non-zero exit code.
```
naderman:
```suggestion
- `report` means filtered packages are reported but do not cause the composer audit command to return a non-zero exit code.
```

doc/06-config.md
```diff
@@ -330,6 +330,28 @@ in other places than the `compoder audit` command.
...
+- `fail` means filtered packages will cause the audit command to fail with a non-zero exit code.
+
+Note that this only applies to audit reports, not to version blocking.
```
naderman:
```suggestion
Note that this only applies to audit reports, not to version blocking on updates.
```

But is filtering even the same as version blocking? Or do we use version blocking only for the security vulnerabilities?

I guess in the end the question becomes: Why aren't vulnerabilities just another filter list? But that's probably for a future major release ;-)

naderman (changes requested):
didn't get far with review yet, but submitting anyway for now ;-)

stof:
As this adds new fields in the packages.json file of a composer repository, this PR should also update the [composer-repository-schema.json file](https://github.com/composer/composer/blob/main/res/composer-repository-schema.json) describing the schema of this file.

glaubinix:
@stof thanks, yes. Will add this now that we agreed on what should be in there.

doc/03-cli.md
glaubinix: doesn't matter. either works

doc/06-config.md
glaubinix: yes, I think this would be a great task for Composer 3. Unify all the filter logic

doc/05-repositories.md
```diff
@@ -326,6 +326,126 @@ can be configured.
...
+    "packages": {
+        "vendor/package": [{ ... }]
+    },
+    "filter": {
+        "aikido-malware": [
+            {
+                "constraint": ">=1.0.0,<1.2.0",
+                "url": "https://example.org/advisories/123",
+                "reason": "Remote code execution vulnerability",
+                "id": "GHSA-xxxx-xxxx-xxxx"
```
naderman: Maybe replace this with some other id that isn't a security advisory as this is explicitly not about advisories but malware?

naderman: Actually the whole example is all about vulnerabilities which is misleading.

doc/05-repositories.md
```diff
@@ -326,6 +326,126 @@ can be configured.
...
+            }
+        ]
+    }
+}
+```
+
+In a composer.json file, the `filter` key controls how filter lists from a repository are used for audit reports and version
```
naderman:
```suggestion
In a composer.json file, the `filter` key in a repository definition controls how filter lists from the respective repository are used for audit reports and version
```

doc/06-config.md
naderman: Could go with "exempt-packages"? exempt from filtering isn't as ambiguous as ignoring?

doc/06-config.md
```diff
@@ -394,6 +384,96 @@ version blocking is not disabled by setting [`block-insecure`](#block-insecure)
...
+```json
+{
+    "config": {
+        "filter": {
+            "dont-filter-packages": [
+                {
+                    "package": "vendor/package",
+                    "constraint": "^2.0",
+                    "reason": "Assessed and accepted risk",
+                    "apply": "audit"
```
naderman, line 444: This needs an explanation, right now it's entirely unclear what it means to not filter a particular package only with some version constraint and limited apply. Does that mean in this example here 3.0 would still get filtered, and even 2.0 would get blocked but not reported on audit? Or since we're talking about exempting a package, does it mean it will be reported on audit, but it will not get filtered during version blocking?

doc/06-config.md
```diff
@@ -394,6 +384,96 @@ version blocking is not disabled by setting [`block-insecure`](#block-insecure)
...
+            ]
+        }
+    }
+}
+```
+
+### sources
+
+By default, list data is fetched from the configured Composer repositories that provide filter list data. You can configure
+additional sources to fetch filter list data from. Keys are source names, values are objects with `type` and `url`.
```
naderman:
```suggestion
additional sources to fetch filter list data from. Keys are source names, values are objects with a key `type` and type specific configuration.
```

doc/06-config.md
```diff
@@ -394,6 +384,96 @@ version blocking is not disabled by setting [`block-insecure`](#block-insecure)
...
+Currently only `type: "url"` is supported. The configured URL will receive your full list of installed dependencies as
```
naderman:
```suggestion
Currently only `type: "url"` is supported which requires an additional `"url"` key. The configured URL will receive your full list of installed dependencies as
```

naderman: Think this needs to clarify if the list of packages sent includes information on specific versions used on audit, i.e. does this expose which versions your project uses, or only which packages?

naderman (changes requested):

doc/05-repositories.md
glaubinix: Updated, this now uses the `PKFE-` prefix for Packagist Filter Entry which was proposed in the Packagist PR https://github.com/composer/packagist/pull/1700

doc/06-config.md
glaubinix: Added an explanation

doc/06-config.md
glaubinix: clarified

doc/06-config.md
glaubinix: renamed to unfiltered-packages

doc/05-repositories.md
```diff
@@ -326,6 +326,126 @@ can be configured.
...
+    "repositories": [
+        {
+            "type": "composer",
+            "url": "https://example.org",
+            "filter": {
+                "lists": [
+                    "defaults",
+                    "!untrusted-list",
+                    {"name": "audit-only-list", "apply": "audit"},
+                    {"name": "block-only-list", "apply": "block", "reason": "corporate policy"}
```
Seldaek: To avoid confusion with apply in unfiltered-packages and in security-advisories ignore lists where the apply means "the command where the ignore is applied", I think we should rename this because in this case it really means "the command where the list is active" 

```suggestion
                    {"name": "audit-only-list", "only": "audit"},
                    {"name": "block-only-list", "only": "block", "reason": "corporate policy"}
```

doc/06-config.md
```diff
@@ -394,6 +384,99 @@ version blocking is not disabled by setting [`block-insecure`](#block-insecure)
 }
 ```
 
+## filter
+
+Filter list configuration options. Controls how filter lists are used for audit reports and version blocking.
+Can be set to `true` to enable with defaults, `false` to fully disable, or configured with an object.
```
Seldaek: I think this is always on now so it probably should say so.

```suggestion
Defaults to `true`. Filter list configuration options. Controls how filter lists are used for audit reports and version blocking.
Can be set to `true` to enable with defaults, `false` to fully disable, or configured with an object.
```

src/Composer/FilterList/FilterListConfig.php
```diff
@@ -0,0 +1,107 @@
...
+                if (is_array($source) && isset($source['type']) && $source['type'] === 'url') {
+                    $sources[] = new UrlSource($sourceName, $source['url']);
+                }
+            }
+        }
+
+        return new self(
+            $unfilteredPackages,
+            $sources,
+            (bool) ($config->get('ignore-unreachable') ?? false)
```
Seldaek: This should read from $filterConfig and not from $config root
```suggestion
            (bool) (is_array($filterConfig) ? ($filterConfig['ignore-unreachable'] ?? false) : false)
```

src/Composer/DependencyResolver/FilterListPoolFilter.php
```diff
@@ -0,0 +1,97 @@
...
+        $this->filterListAuditor = $filterListAuditor;
+        $this->httpDownloader = $httpDownloader;
+    }
+
+    /**
+     * @param array<RepositoryInterface> $repositories
+     */
+    public function filter(Pool $pool, array $repositories, Request $request): Pool
+    {
+        if ($this->filterListConfig->getOperationConfig('block') === null) {
```
Seldaek: This is never true as getOperationConfig returns a FilterListConfig instance. Weird that phpstan didn't catch this.

src/Composer/FilterList/FilterListAuditor.php
```diff
@@ -0,0 +1,85 @@
...
+        $filterConfig = $filterListConfig->getOperationConfig($operation);
+        $unFilteredPackageMap = $filterConfig->getUnfilteredPackageMap();
+        foreach ($package->getNames(false) as $packageName) {
+            if (!isset($filterListMap[$packageName])) {
+                continue;
+            }
+
+            $packageConstraint = new Constraint(Constraint::STR_OP_EQ, $package->getVersion());
+            foreach ($filterListMap[$packageName] as $entries) {
+                if (isset($unFilteredPackageMap[$packageName]) && $unFilteredPackageMap[$packageName]->constraint->matches($packageConstraint)) {
```
Seldaek: The isset() here is not quite enough if the unfilteredPackageMap contains `foo/*` with a wildcard.. Should probably loop over the map here and check if the name is a match with BasePackage::packageNameToRegexp()

Could also do `BasePackage::packageNamesToRegexp(array_keys($unFilteredPackageMap))` once above at line 63 then here you'd be able to check in one go if any unfiltered-packages rule apply or not, and if not skip the loop over the whole map.

tests/Composer/Test/Fixtures/installer/update-filter-entry-unfiltered-package-versions.test
```diff
@@ -0,0 +1,60 @@
...
+--COMPOSER--
+{
+    "name": "acme/project",
+    "version": "1.0.0",
+    "require": {
+        "acme/library": "^1.0.0"
+    },
+    "config": {
+        "filter": {
+            "unfiltered-packages": [{"package": "acme/library", "version": "1.1.0"}]
```
Seldaek: Not sure why this test passes tho.. I guess it just ignores all versions if no constraint is given?

```suggestion
            "unfiltered-packages": [{"package": "acme/library", "constraint": "1.1.0"}]
```

tests/Composer/Test/FilterList/FilterListConfigTest.php
```diff
@@ -0,0 +1,70 @@
...
+        $listConfig = $filterConfig->getOperationConfig('block');
+        $this->assertCount(1, $listConfig->unfilteredPackages);
+        $this->assertCount(1, $listConfig->sources);
+    }
+
+    public function testParseConfigDoesntApply(): void
+    {
+        $config = new Config();
+        $config->merge(['config' => ['filter' => [
+            'unfiltered-packages' => [['name' => 'foo/bar', 'apply' => 'audit'],
```
Seldaek: FilterListConfig or UnfilteredPackage should probably error if it cannot even find a package name in there? That'd have caught the misconfiguration

```suggestion
            'unfiltered-packages' => [['package' => 'foo/bar', 'apply' => 'audit'],
```

src/Composer/FilterList/Source/UrlSource.php
```diff
@@ -0,0 +1,34 @@
...
+    public $name;
+    /** @var string */
+    public $url;
+
+    public function __construct(
+        string $name,
+        string $url
+    ) {
+        $this->name = $name;
+        $this->url = $url;
```
Seldaek, line 32: Can we require HTTPS-schemed URLs here?

src/Composer/Config.php
```diff
@@ -234,6 +234,19 @@ public function merge(array $config, string $source = self::SOURCE_UNKNOWN): voi
                     $this->config[$key] = array_merge($this->config['audit'], $val);
                     $this->setSourceOfConfigValue($val, $key, $source);
                     $this->config['audit']['ignore'] = array_merge($currentIgnores, $val['ignore'] ?? []);
+                } elseif ('filter' === $key) {
+                    $unfilteredPackages = $this->config['filter']['unfiltered-packages'] ?? [];
+                    $sources = $this->config['filter']['sources'] ?? [];
+
+                    if (\is_bool($val)) {
+                        $this->config[$key] = $val;
```
Seldaek: Not sure if $val is true whether we should overwrite all previously collected filter configs.. I'd say probably not?

If it is false then sure disable at project level, but if true in project the global config filter rules should probably still apply.

src/Composer/Installer.php
```diff
@@ -455,8 +440,10 @@ public function run(): int
                     foreach ($this->repositoryManager->getRepositories() as $repo) {
                         $repoSet->addRepository($repo);
                     }
+                    $filterListConfig = FilterListConfig::fromConfig($this->config, new VersionParser());
```
Seldaek: This is created here and below at line 1140, maybe we should deduplicate and create the config only once?

src/Composer/FilterList/FilterListConfig.php
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * @return array<string, UnfilteredPackage>
+     */
+    public function getUnfilteredPackageMap(): array
+    {
+        $map = [];
+        foreach ($this->unfilteredPackages as $unfilteredPackage) {
+            $map[$unfilteredPackage->packageName] = $unfilteredPackage;
+        }
```
Seldaek: This risks overwriting rules silently if you declare rules for the same package twice.. Not sure if that's a problem but I'd say having the map keyed by package name is not that useful anyway because of wildcards we anyway need to loop all items to check a name against it.

res/composer-schema.json
```diff
@@ -1122,7 +1160,49 @@
...
                 "allow_ssl_downgrade": { "type": "boolean" },
-                "force-lazy-providers": { "type": "boolean" }
+                "force-lazy-providers": { "type": "boolean" },
+                "filter": {
+                    "type": ["boolean", "object"],
+                    "description": "Filter list configuration for this repository. Set to false to disable filter lists from this repository entirely, or configure with an object.",
+                    "properties": {
+                        "lists": {
+                            "type": "array",
+                            "description": "Filter lists to use from this repository. Use 'defaults' to include all default lists advertised by the repository, prefix a name with '!' to exclude it, or provide objects for detailled configuration.",
```
```suggestion
                            "description": "Filter lists to use from this repository. Use 'defaults' to include all default lists advertised by the repository, prefix a name with '!' to exclude it, or provide objects for detailed configuration.",
```

res/composer-repository-schema.json
```diff
@@ -49,6 +49,31 @@
                 }
             }
         },
+        "filter": {
+            "type": "object",
+            "description": "Filter list configuration advertised by the repository.",
+            "required": ["metadata", "lists"],
```
Seldaek: Should match the docs, IMO ok to make it all required..

```suggestion
            "required": ["metadata", "lists", "default-lists"],
```

doc/05-repositories.md
glaubinix: Makes sense, changed

tests/Composer/Test/FilterList/FilterListConfigTest.php
glaubinix: Throwing exceptions now if the config is wrong

src/Composer/FilterList/Source/UrlSource.php
glaubinix: Makes sense. This requires https now

src/Composer/FilterList/UnfilteredPackage.php
```diff
@@ -0,0 +1,77 @@
...
+        if (\is_string($config)) {
+            return new self($config, new MatchAllConstraint());
+        }
+
+        if (\is_array($config) && \count($config) === 1 && !isset($config['package']) && !isset($config['constraint'])) {
+            return new self(\key($config), $parser->parseConstraints((string) \array_pop($config)));
+        }
+
+        if (!isset($config['package'], $config['constraint'])) {
+            throw new \RuntimeException('Invalid unfiltered package config. "package" and "constraint" are requried.');
```
Seldaek:
```suggestion
            throw new \RuntimeException('Invalid unfiltered package config. "package" and "constraint" are required.');
```

PR #12779, opened 2026-03-25
Add the 'Moto' (more than one class per file) autoloading algo

pmjones:
I was unsure of the proper order to use. I settled on: when PSR-0 and 4 are specified together, Moto comes after; when only PSR-4 is specified, Moto comes before. Let me know if some other order is preferable.

- Moto algorithm: https://github.com/motophp/autoload
- Reference implementation: https://github.com/motophp/autoload/blob/0.x/src/Loader.php


pmjones:
Looks like I have some cleanup work do, will get on it.

stof:
Is this "Moto" algorithm specified somewhere ? If yes, where is this ? The PR description would deserve a link.

pmjones:
> Is this "Moto" algorithm specified somewhere ? If yes, where is this ? The PR description would deserve a link.

Yes it would, of course -- added to the description:

- Moto algorithm: https://github.com/motophp/autoload
- Reference implementation: https://github.com/motophp/autoload/blob/0.x/src/Loader.php

pmjones:
I see that the PHPStan 7.2 check fails, but it looks like a transient failure. @stof can you (or someone else) retry that check?

Seldaek:
Yeah ignore the 7.2 failures if it crashes, unfortunately haven't found a way to fix this but hopefully soon resolved by dropping old php versions.

Anyway thanks for the PR, will need some time to review

pmjones:
> will need some time to review

Understood, and thanks; pick apart the algo itself too if you like, never hurts to have another pair of eyes on that kind of thing.

src/Composer/Autoload/AutoloadGenerator.php
```diff
@@ -349,14 +373,14 @@ public static function autoload(\$class)
...
-                        // if the vendor dir is contained within a psr-0/psr-4 dir being scanned we exclude it
+                        // if the vendor dir is contained within a psr-0/psr-4/moto dir being scanned we exclude it
                         if (str_contains($vendorPath, $dir.'/')) {
                             $exclusionRegex = $this->buildExclusionRegex($dir, array_merge($excluded, [$vendorPath.'/']));
                         } else {
                             $exclusionRegex = $this->buildExclusionRegex($dir, $excluded);
                         }
 
-                        $classMapGenerator->scanPaths($dir, $exclusionRegex, $group['type'], $namespace);
+                        $classMapGenerator->scanPaths($dir, $exclusionRegex, $group['type'] === 'moto' ? 'psr-4' : $group['type'], $namespace);
```
stof: I don't think the class map generator should scan it in PSR-4 mode. It would report other classes as incompatible with PSR-4 and exclude them from the optimized class map (and even fail in case of defining a strict autoloader). Instead, the class map generator needs to get support for Moto as well.

src/Composer/Autoload/ClassLoader.php
```diff
@@ -491,10 +598,80 @@ public static function getRegisteredLoaders()
      */
     private function findFileWithExtension($class, $ext)
     {
+        $first = $class[0];
+
+        // Moto lookup
```
stof, line 629: As PSR-4 is clearly the most common algorithm in the ecosystem, should we keep PSR-4 as the first one being checked in the autoloader as it is more likely to be the one finding the class ?

stof (commented):
Tests for the autoload generator are missing coverage for the moto algorithm in the static autoload file AFAICT.

src/Composer/Autoload/ClassLoader.php
pmjones: I am totally fine with PSR-4 being first; I was not sure which order would be best. Does the Composer team feel the best order is PSR-4, then PSR-0, then Moto -- or PSR-0, PSR-4, then Moto -- or what?

pmjones:
@stof I believe all the above comments have been addressed, plus one oversight. Review again as convenient, let me know what I missed.

pmjones:
Looks like the API Surface Check failed, but I think that's a Github hiccup -- let me know if it's the PR code instead.

PR #12782, opened 2026-03-29
Add `--only-dev` option to `show` and `outdated` commands

### LINKED ISSUE #12747: Add `--dev-only` to `composer show` and `composer outdated`
Hi!
Composer has `--no-dev` in commonly used commands like `outdated` and `show` to exclude dependencies that are in `require-dev`, but there is no symmetric way to target only dependencies from `require-dev`.

This makes a very common workflow awkward: sometimes we want to do maintenance on tooling (phpstan/phpunit/fixer/etc.) without touching production dependencies, but we cannot even quickly see what's in `require-dev` because there is no `composer show --dev-only` / `composer outdated --dev-only`. That way, we have to use brittle workarounds (parse `composer.json`, diff outputs, or pass large package lists manually).

**Requested feature:**
add `--dev-only` to `composer show` and `composer outdated`, with semantics "operate only on `require-dev` (and their dependency graph)".

---

**Note on `update`:**
initially this issue also suggested adding `--dev-only` to `update`, but I've removed that part after learning that `update --no-dev` doesn't do what we might think it does (thanks to @stof reply).
That said, it would still be nice to have a separate flag to update only root packages from `require-dev` (as if we were running `composer update phpunit/phpunit phpstan/phpstan ...`).

Jay-Way:
Fixes #12747  
 

- Added new `--only-dev` option, working with `show` and `outdated` commands, showing only `require-dev` packages in output
- Made sure `--no-dev` and `--only-dev` cannot be combined
- Add test cases in `ShowCommandTest` with the new behaviour (Covers `show` and `outdated` commands)

### DIFF
--- a/src/Composer/Command/OutdatedCommand.php
+++ b/src/Composer/Command/OutdatedCommand.php
@@ -44,6 +44,7 @@ class OutdatedCommand extends BaseCommand
                 new InputOption('format', 'f', InputOption::VALUE_REQUIRED, 'Format of the output: text or json', 'text', ['json', 'text']),
                 new InputOption('ignore', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Ignore specified package(s). Can contain wildcards (*). Use it if you don\'t want to be informed about new versions of some packages.', null, $this->suggestInstalledPackage(false)),
                 new InputOption('no-dev', null, InputOption::VALUE_NONE, 'Disables search in require-dev packages.'),
+                new InputOption('only-dev', null, InputOption::VALUE_NONE, 'Restricts the list of packages to your require-dev packages.'),
                 new InputOption('ignore-platform-req', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Ignore a specific platform requirement (php & ext- packages). Use with the --outdated option'),
                 new InputOption('ignore-platform-reqs', null, InputOption::VALUE_NONE, 'Ignore all platform requirements (php & ext- packages). Use with the --outdated option'),
             ])
@@ -109,6 +110,9 @@ EOT
         }
         if ($input->getOption('no-dev')) {
             $args['--no-dev'] = true;
+        }
+        if ($input->getOption('only-dev')) {
+            $args['--only-dev'] = true;
         }
         if ($input->getOption('sort-by-age')) {
             $args['--sort-by-age'] = true;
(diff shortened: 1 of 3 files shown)


naderman:
I think this makes sense generally. But I think this is still very ambiguous and will need at least some clarification or changes. In particular I think this needs to make it clearer what happens to packages that are dependencies of both dev dependencies and prod dependencies (I'll call non-dev prod here for clarity). Especially if something I explicitly list as a require in my dev dependencies (maybe with a small version constraint) is also a dependency of a prod dependency it would not get analyzed at all here. That's at the very least counterintuitive. I run --only-dev but I don't even get info on my explicitly listed dev dependencies?

I think at the very least this will need some more extensive test coverage over the expected behavior for scenarios with shared dependencies or with explicit dev dependencies that are transitive dependencies of prod dependencies. So when the first complaints/bugs come in, at least we've got this well spec'd out and documented and don't keep changing the behavior back and forth without a clear goal.

That all said, what is the use case exactly? Would you exepect composer outdated --only-dev to list all your dev dependencies and their deps, or just the ones that are not also prod depdencies? I'm really not sure what the right answer to that one is.

Jay-Way:
@naderman Good questions, 

I think the use case is laid out quite well by the issue creator here https://github.com/composer/composer/issues/12747#issue-3968136793

But I can't estimate if it is really useful for the broader userbase

Regarding the shared dependencies - I added these in the test cases, but you're right I do not explicitly check for them in the output. I will add an `self::assertStringNotContainsString('vendor/shared', $output)` where it makes sense - the current behaviour will **not** show dependencies both in `require` and `require-dev`, if we test this explicitly it will become more clear, thanks for the suggestion.

I can add more test cases if you have more ideas for different combinations

Jay-Way:
I think the PHPStan 7.2 job failed due to runner error, maybe someone can retry it

naderman:
We still need a decision on:

require: vendor/prod (which requires vendor/dev)
require-dev: vendor/dev

Should that really hide vendor/dev from the `outdated --only-dev` output? I suppose, but that needs to have a test.

Jay-Way:
Oh so this is about transitive dependencies?

The answer is probably yes, it should be hidden (a package that is needed in production, even transitively, isn't purely dev)

Edit: Oops, misclicked on close PR

Jay-Way:
I added two test cases: https://github.com/composer/composer/pull/12782/commits/18f8d61b6aea904e56bd37005be61f6680f42836

naderman:
Thanks @Jay-Way, that is the scenario I was thinking of. May feel a bit counter intuitive, but feel better we at least documented this behavior in a test case when the first bug report comes in ;-)

@Seldaek this generally as a feature seems alright to me. Didn't really look into whether the implementation makes sense / couldn't be more efficient than that sequence of filter required packages calls. Let me know if you need me to look at that too.

PR #12787, opened 2026-03-31
Add public API surface change detection for PRs

Seldaek:
This is really just an experiment, need to see how it works in practice if it's useful or annoying, but the point is to make it a more conscious decision to add public, non-internal API surface.

Feedback welcome.


sebastianbergmann:
Interesting! Depending on how your experiment goes, do you plan to make this (re)usable for other projects? Thanks!

Issue #9920, opened 2021-05-27
Add possibility to specify the path of an auth.json


sippsolutions:
It would be nice to have the option to specify the location of the `auth.json`.

In our case we have a repository with a `./composer.json` that provides tools to build and deploy the application and a `src/composer.json` that installs the dependencies of the actual software which has its code source in `src/`.

We need to install modules that require authentication in both `./` and `src/` and therefor it would be awesome to have the possibility to use the `src/auth.json` in the root `./composer.json` as well.

Maybe it could be useful for some users to specify multiple paths. We should still respect the user's `{home}/auth.json` and the `auth.json` at the same path as the currently evaluated `composer.json`.

A config example could be:
```json
{
    "config": {
        "auth-files": [
            "src/auth.json",
            "../auth.json",
            "vendor/my/package/auth.json"
        ]
    }
}
```

Seldaek:
IMO auth.json should really only ever be in the home dir, as you shouldn't share/commit credentials in the repo.. So I don't see a huge need for this capability.

sippsolutions:
@Seldaek It's common practice to add an `auth.json` for Enterprise software (such as Magento, where you have different auth keys for different projects) in the same directory as the `composer.json` just to identify as authorized to download these packages (which an build and deployment framework needs to be authorized as well), which shouldn't be an issue as those repositories are privately hosted anyways. As it's currently possible to use the `auth.json` in the same directory as the `composer.json`, I'd say this feature request is justified - it'd be possible to define other locations as well, not only inside the repository.

tpetry:
When using docker build using a different `auth.json` is also very important as the file could be a mounted docker secret which will not be part of the image.

computator:
Using this with docker is my usecase as well

Seldaek:
Do you have a chance to pass env vars easily? Because you could read the user auth.json into COMPOSER_AUTH env var and make that available in docker, then it should override the docker auth.json - https://getcomposer.org/doc/03-cli.md#composer-auth

DaanBiesterbos:
I would argue that your proposal, would be an invitation for bad practices, and possible security issues. I'll try to avoid an in depth argument against hardcoding composer configs per project or even per package.  But to make my point I guess it would be fair to give some examples to consider. Just one recent example I think of... I am currently working for a client that recently migrated from bitbucket to gitlab on premise. If they'd hardcoded auth configs in a couple of thousand projects. This migration would have been so much harder, and arguably not worth it. 

[...]

PR #12795, opened 2026-04-07
Feature/composer auth file env

maiorano84:
See: #9920

This is a small feature enhancement to cover for a use-case in which users might want to set an environment variable to alter the default path of the global `auth.json` configuration.

This is primarily useful in containerized environments in which an `auth.json` file might need to be mounted as a secret. Specifically in Docker containers, the `*_FILE` suffix is a common pattern for indicating the location of a particular piece of sensitive data as a file rather than directly as an environment variable or build arg (which can be viewed in plain text with tools like `docker inspect`).

When `COMPOSER_AUTH_FILE` is present and non-empty, the default path of `$config->get('home').'/auth.json'` is replaced with the value of this variable and the contents of the file in that location will be loaded and validated as normal.

I didn't see any relevant unit tests for the existing `COMPOSER_AUTH` or `createConfig` methods (outside of possibly the FactoryMock, which does not seem to be Environment-aware) so I'm not sure if that's something we would like to add into any tests to account for this new behavior.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -1230,6 +1230,14 @@ github-oauth, bitbucket-oauth, ... objects as needed](articles/authentication-fo
 and following the
 [spec from the config](06-config.md).
 
+### COMPOSER_AUTH_FILE
+
+The `COMPOSER_AUTH_FILE` var allows you to set up authentication under an alternative file path.
+The contents of the variable should be an absolute path to a JSON formatted file containing
+[http-basic, github-oauth, bitbucket-oauth, ... objects as needed](articles/authentication-for-private-packages.md),
+and following the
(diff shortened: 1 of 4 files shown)


maiorano84:
It looks like there's a PHPStan issue affecting a few PRs (including my own). The specific rule being triggered seems to have a bunch of ignore instances throughout the `baseline-8.4.neon` file, so I've opted to add another one to cover for this specific instance.

Seldaek:
Fixed PHPStan issue in main branch so you can rebase. The rest I'll review later sorry focusing on shipping 2.10 atm :)

maiorano84:
@Seldaek 

All good man. I've synced up, so that should be ready when you have a moment to circle back.

PR #12798, opened 2026-04-12
Preserve file mtimes in ZIP archives

### LINKED ISSUE #11342: `ArchiveManager::archive()` doesn't appear to preserve file timestamps
I've reported an issue in composer/satis (https://github.com/composer/satis/issues/715), but as far as I can tell the issue originates here. Composer's [ArchiveManager::archive()](https://github.com/composer/composer/blob/ecb69eaa4aec3c2bd866d17ee7a46e684dc16f71/src/Composer/Package/Archiver/ArchiveManager.php#L146) doesn't seem to preserve the file modification timestamps from the source when creating an archive.

In Satis, when a package is re-archived, the unzipped files have a modified time of `1 January 1980 at 00:00`. Its important (for caching purposes at least) that the file modified timestamps match the timestamp of the tag/release.

ShipItAndPray:
## Summary
- preserve source file modification times when Composer creates ZIP archives
- keep the change guarded by `ZipArchive::setMtimeName` so older environments still work
- add a regression test that verifies the archived mtime

Fixes #11342

## Validation
- git diff --check

## Notes
- local PHP is not installed in this environment, so I could not run the Composer PHPUnit suite here

### DIFF
--- a/src/Composer/Package/Archiver/ZipArchiver.php
+++ b/src/Composer/Package/Archiver/ZipArchiver.php
@@ -69,6 +69,16 @@ class ZipArchiver implements ArchiverInterface
                      */
                     $zip->setExternalAttributesName($relativePath, ZipArchive::OPSYS_UNIX, $perms << 16);
                 }
+
+                /**
+                 * setMtimeName() is only available with libzip 1.0.0 or above.
+                 */
+                if (method_exists($zip, 'setMtimeName')) {
+                    $mtime = filemtime($filepath);
+                    if (false !== $mtime) {
+                        $zip->setMtimeName($relativePath, $mtime);
+                    }
+                }
             }
             if ($zip->close()) {
                 if (!file_exists($target)) {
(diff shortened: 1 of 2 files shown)


Seldaek:
Claude review flagged two issues which are very much worth discussing:

##  Reproducible-build regression

>  This silently changes archive byte output for every user. Anyone relying on byte-stable ZIPs (content-addressable storage, integrity checks  against cached SHAs, distribution mirrors) will see hashes change on every regeneration unless the source mtimes are themselves stable. For  git-derived sources, mtimes are checkout times — effectively random — so re-archiving the same commit on two machines now produces different  bytes. Worth confirming this is acceptable or considering an opt-out flag. (For context, git archive deliberately uses the commit time, not file  mtimes, for exactly this reason.)

IMO we should have a way to optionally pass a timestamp when calling archive() and if that is set then it would override the timestamp for all files, much like git archive does, then we could mimick that behavior a bit closer when archiving a given release you really want to archive with the date of the release IMO and not last modification time on disk which can be fairly jittery.

##  Directories not handled.
 
> setMtimeName is only called inside the else branch (the file path), so addEmptyDir() entries still get the epoch default. Likely fine in practice (statName on directories is rarely consulted), but worth noting for consistency. If kept files-only, the docblock could say so.

IMO would be great to set them as well.

PR #12799, opened 2026-04-12
Avoid Packagist guidance for non-Packagist advisories

ShipItAndPray:
## Summary
- keep Packagist advisory links for PKSA advisories
- show a generic advisory hint when the blocking advisory comes from a non-Packagist source
- add a regression fixture for a Wordfence-style advisory

Fixes #12769

## Validation
- git diff --check

## Note
- This PR was AI-assisted, per the repository contribution guidance.

## Blocker
- PHP is not installed in this environment, so the Composer installer fixture could not be executed locally.

### DIFF
--- a/src/Composer/Package/Archiver/ZipArchiver.php
+++ b/src/Composer/Package/Archiver/ZipArchiver.php
@@ -69,6 +69,16 @@ class ZipArchiver implements ArchiverInterface
                      */
                     $zip->setExternalAttributesName($relativePath, ZipArchive::OPSYS_UNIX, $perms << 16);
                 }
+
+                /**
+                 * setMtimeName() is only available with libzip 1.0.0 or above.
+                 */
+                if (method_exists($zip, 'setMtimeName')) {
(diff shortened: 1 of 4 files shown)


src/Composer/Package/Archiver/ZipArchiver.php
```diff
@@ -69,6 +69,16 @@ public function archive(string $sources, string $target, string $format, array $
                      */
                     $zip->setExternalAttributesName($relativePath, ZipArchive::OPSYS_UNIX, $perms << 16);
                 }
+
+                /**
+                 * setMtimeName() is only available with libzip 1.0.0 or above.
+                 */
+                if (method_exists($zip, 'setMtimeName')) {
```
Seldaek: Looks like this PR includes the changes from #12798

Seldaek: Fixed it myself, rebased only the second commit

Seldaek:
Thanks

PR #12804, opened 2026-04-16
Introduce a unified Policy config that covers advisories, abandoned, and malware

glaubinix:
Introducing a new `policy` config that covers what is currently known as `audit` and `filter` while keeping support for the existing `audit` config.

#### Changes in this PR
* drops filter config and uses new policy config instead
* unifies how `ignore` works across advisories, abandoned and malware
* `ignore-unreachable` can now be configured separately for audit|update|install
* drops `default-lists` as part of the Composer repository schema
* drops `unfiltered-packages` as part of the repository config

#### Pending items that will be added via follow up PR
* New ENV variables https://github.com/composer/composer/pull/12806
* `ignore-source` for the malware list https://github.com/composer/composer/pull/12807
* option to automatically migrate the old audit config to the new policy config (likely only happening in 2.11)
* support for the policy config in the ConfigCommand https://github.com/composer/composer/pull/12809
* implement block on install for malware https://github.com/composer/composer/pull/12826
* get rid of the existing AuditConfig object https://github.com/composer/composer/pull/12831
* adjusted composer repository schema that includes api-url/summary-url for malware
  * summary-url: https://github.com/composer/composer/pull/12833
  * api-url: 

#### The new config format

Take a look at the changes to the composer-schema.json to see all possible values.

```json
{
    "config": {
        "policy": {
            "advisories": {
                "block": true,
                "audit": "fail",
                "ignore": {
                    "vendor/package": "all advisories for this package ignored"
                },
                "ignore-id": {
                    "CVE-2024-1234": "not reachable in our app",
                    "GHSA-xxxx-yyyy": {"on-block": false, "reason": "mitigated by WAF"}
                },
                "ignore-severity": {
                    "low": "we accept low-severity risk"
                }
            },
            "malware": {
                "block": true,
                "block-scope": "all",
                "audit": "fail",
                "ignore": {
                    "vendor/false-positive": "verified safe, flagged by mistake"
                },
                "ignore-source": ["untrusted-repo-name"]
            },
            "abandoned": {
                "block": false,
                "audit": "fail",
                "ignore": {
                    "vendor/old-pkg": "replacement not ready yet",
                    "vendor/legacy-*": {"on-audit": false, "reason": "allow install but keep reporting"},
                    "vendor/pinned": {"constraint": "^2.0", "reason": "only v2 still in use"}
                }
            },
            "company-policy": {
                "sources": [
                    {"type": "url", "url": "https://example.com/policy-filter.json"},
                    {"type": "url", "url": "https://security-team.example.com/extra-rules.json"}
                ],
                "block": true,
                "audit": "fail",
                "ignore": {
                    "vendor/internal-fork": "maintained internally"
                }
            },
            "ignore-unreachable": ["install", "update"]
        }
    }
}
```


doc/06-config.md
```diff
@@ -145,15 +145,429 @@ COMPOSER_SOURCE_FALLBACK=0 composer install
...
+            }
+        }
+    }
+}
+```
+
+##### With scoping:
+
+`on-block: false` means the advisory no longer blocks updates but is still reported in audit.
+`on-audit: false` means the advisory still blocks updates but is no longer reported in audit.
```
Seldaek, line 245: I think this is backwards. If you ignore by default you ignore the advisory from everything (block and audit). If you set on-block:false, then you DO NOT ignore when blocking, so you only ignore in audits.

on-block/on-audit are essentially always set but they default to true, so you can skip some ignoring if you want by setting these to false.

doc/06-config.md
```diff
@@ -145,15 +145,429 @@ COMPOSER_SOURCE_FALLBACK=0 composer install
...
+
+```json
+{
+    "config": {
+        "policy": {
+            "advisories": {
+                "ignore-id": {
+                    "CVE-1234": {"on-block": false, "reason": "Patch applied, still want blocking."},
+                    "GHSA-xx":  {"on-audit": false, "reason": "False positive, still report."}
```
Seldaek, line 254: These comments are correct tho, only the description above is reversed.

doc/06-config.md
```diff
@@ -145,15 +145,429 @@ COMPOSER_SOURCE_FALLBACK=0 composer install
...
+            }
+        }
+    }
+}
+```
+
+##### With scoping:
+
+`on-block: false` ignores only for audit (the package is not blocked during updates).
+`on-audit: false` ignores only for blocking (the package is still reported in audit).
```
Seldaek:
```suggestion
`on-block: false` ignores only for audit (the package is still blocked during updates as on-block ignoring is disabled).
`on-audit: false` ignores only for blocking (the package is still reported in audit).
```

doc/06-config.md
```diff
@@ -145,15 +145,429 @@ COMPOSER_SOURCE_FALLBACK=0 composer install
...
+
+```json
+{
+    "config": {
+        "policy": {
+            "<list>": {
+                "ignore": {
+                    "vendor/package": {"on-block": false, "reason": "Workaround applied; keep reporting."}
```
Seldaek:
```suggestion
                    "vendor/package": {"on-audit": false, "reason": "Workaround applied; keep reporting."}
```

src/Composer/Command/CreateProjectCommand.php
```diff
@@ -93,6 +93,7 @@ protected function configure(): void
                 new InputOption('no-audit', null, InputOption::VALUE_NONE, 'Whether to skip auditing of the installed package dependencies (can also be set via the COMPOSER_NO_AUDIT=1 env var).'),
                 new InputOption('audit-format', null, InputOption::VALUE_REQUIRED, 'Audit output format. Must be "table", "plain", "json" or "summary".', Auditor::FORMAT_SUMMARY, Auditor::FORMATS),
                 new InputOption('no-security-blocking', null, InputOption::VALUE_NONE, 'Allows installing packages with security advisories or that are abandoned (can also be set via the COMPOSER_NO_SECURITY_BLOCKING=1 env var).'),
+                new InputOption('no-blocking', null, InputOption::VALUE_NONE, 'Disables all policy blocking during this command. Alias for --no-security-blocking (can also be set via the COMPOSER_NO_SECURITY_BLOCKING=1 env var).'),
```
Seldaek: Should this not be the other way around? Deprecate the old flag and make this new one the canonical one instead of syaing it's an alias I mean.

Also this is not present in the docs as far as I've seen.

I'd probably also add a COMPOSER_NO_BLOCKING env var for consistency.

src/Composer/Config.php
```diff
@@ -502,22 +509,21 @@ public function get(string $key, int $flags = 0)
...
 
                 return $result;
-            case 'filter':
-                $filterConfig = $this->config[$key];
-                $filterEnv = $this->getComposerEnv('COMPOSER_FILTER');
-                if (false !== $filterEnv) {
-                    if (!in_array($filterEnv, ['0', '1'], true)) {
+            case 'policy':
+                $policyConfig = $this->config[$key];
+                $policyEnv = $this->getComposerEnv('COMPOSER_POLICY');
```
Seldaek, line 527: Looks like this name should be updated in the docs 03-cli.md

src/Composer/Policy/PolicyConfig.php
```diff
@@ -0,0 +1,248 @@
...
+    public const FUTURE_RESERVED_PREFIXES = [
+        'ignore',
+    ];
+
+    public const FUTURE_RESERVED_NAMES = [
+        'package', 'packages',
+        'license', 'licence', 'licenses', 'licences',
+        'support', 'maintenance', 'security',
+        'minimum-release-age',
+    ];
```
Seldaek, line 75: All these seem unused. I think they probably should be documented, and IIRC we said the repo should explicitly ignore these and not load them so new versions can start reading them if needed, without causing problems with older clients if we start pushing out new lists.

glaubinix: Added this to the Composer schema, the docs. Filtering out lists coming from the repo and displaying an error when having this in the composer.json

src/Composer/Config.php
glaubinix: Updated

src/Composer/Command/CreateProjectCommand.php
glaubinix: Added COMPOSER_NO_BLOCKING and changed the alias

Seldaek:
Thanks!

PR #12805, opened 2026-04-16
Improve return type annotation for `Application::getComposer()`

eliashaeussler:
We can utilize PHPStan's conditional return types [^1] here in order to improve usage with static code analyzers.

[^1]: https://phpstan.org/writing-php-code/phpdoc-types#conditional-return-types

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -581,7 +581,7 @@ class Application extends BaseApplication
     /**
      * @throws JsonValidationException
      * @throws \InvalidArgumentException
-     * @return ?Composer If $required is true then the return value is guaranteed
+     * @return ($required is true ? Composer : Composer|null) If $required is true then the return value is guaranteed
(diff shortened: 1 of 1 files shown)


src/Composer/Console/Application.php
```diff
@@ -581,7 +581,7 @@ private function hintCommonErrors(\Throwable $exception, OutputInterface $output
     /**
      * @throws JsonValidationException
      * @throws \InvalidArgumentException
-     * @return ?Composer If $required is true then the return value is guaranteed
+     * @return ($required is true ? Composer : Composer|null) If $required is true then the return value is guaranteed
```
Seldaek:
```suggestion
     * @return ?Composer If $required is true then the return value is guaranteed
     * @phpstan-return ($required is true ? Composer : Composer|null)
```

Seldaek:
Makes sense thanks

PR #12806, opened 2026-04-17
FilterList: introduce new env vars and make sure all env vars are covered by tests

glaubinix:
Follow up PR for https://github.com/composer/composer/pull/12804 to add missing env vars to disable blocking for advisories and malware individually. Also adds tests to cover handling of all env vars.


doc/03-cli.md
```diff
@@ -1392,27 +1392,42 @@ If set to `1`, it is the equivalent of passing the `--no-audit` option to a `req
...
+
+If set to `1`, enables blocking of packages with security advisories during dependency resolution (equivalent to setting `policy.advisories.block` to `true`). If set to `0`, disables blocking. Has no effect when `policy.advisories` is set to `false` in composer.json — the env var only flips an active list, it does not re-enable a list the user explicitly disabled.
+
+### COMPOSER_POLICY_MALWARE_BLOCK
+
+If set to `1`, enables blocking of packages flagged as malware during dependency resolution (equivalent to setting `policy.malware.block` to `true`). If set to `0`, disables blocking. Has no effect when `policy.malware` is set to `false` in composer.json.
+
+### COMPOSER_POLICY_ABANDONED_BLOCK
+
+If set to `1`, enables blocking of abandoned packages during dependency resolution (equivalent to setting `policy.abandoned.block` to `true`). If set to `0`, disables blocking. Has no effect when `policy.abandoned` is set to `false` in composer.json.
```
Seldaek: I'm wondering why mention that if set to 1 it enables something, if it technically does not enable anything if composer.json has it set to false? Does this only enable if the composer.json is set to default value (which blocks advisories and malware but not abandoned)? If so setting =1 only works for abandoned, and for the others it never does anything useful?

This seems a bit misleading as is. I'm not sure what usage patterns might be, but I'd say if composer.json has it set to false but you set the env to 1 it should probably enable that list using default settings (suppressing the `false`) and ensure blocking is enabled (only relevant for abandoned, but still).

src/Composer/Util/Platform.php
```diff
@@ -86,6 +86,30 @@ public static function getEnv(string $name)
...
+     * @param non-empty-string $name
+     */
+    public static function getBoolEnv(string $name, bool $default = false): bool
+    {
+        $value = self::getEnv($name);
+        if (false === $value) {
+            return $default;
+        }
+
+        if (!in_array($value, ['0', '1'], true)) {
```
Seldaek: I wonder if we should allow true/on/1 | false/off/0 here to make it a bit more gracious. On the other hand we only ever document 0/1 I think so people should really be using those already.. Just thinking out loud :)

src/Composer/Util/Platform.php
```diff
@@ -86,6 +86,30 @@ public static function getEnv(string $name)
...
+     *
+     * Returns true for '1', false for '0', and $default when the variable is unset.
+     * Throws \RuntimeException for any other value (including the empty string).
+     *
+     * @param non-empty-string $name
+     */
+    public static function getBoolEnv(string $name, bool $default = false): bool
+    {
+        $value = self::getEnv($name);
+        if (false === $value) {
```
Seldaek: Kinda curious here if you explicitly set `SOME_ENV=` to null a var that is already present in your env, it probably should treat it as default and not throw?

```suggestion
        if (false === $value || '' === $value) {
```

src/Composer/Config.php
```diff
@@ -513,25 +513,22 @@ public function get(string $key, int $flags = 0)
...
+                // to rewrite the raw array shape.
                 $policyEnv = $this->getComposerEnv('COMPOSER_POLICY');
                 if (false !== $policyEnv) {
-                    if (!in_array($policyEnv, ['0', '1'], true)) {
-                        throw new \RuntimeException(
-                            "Invalid value for COMPOSER_POLICY: {$policyEnv}. Expected 0 or 1."
-                        );
-                    }
-                    if ((int) $policyEnv === 0) {
+                    if (false === Platform::getBoolEnv('COMPOSER_POLICY')) {
```
Seldaek: This makes me think that getBoolEnv probably should return `?bool` with $default=null for unset/`''`. You can still set the default to true|false for cases where this is not relevant and you want a real bool back.

tests/Composer/Test/Util/PlatformTest.php
```diff
@@ -36,4 +42,60 @@ public function testIsWindows(): void
         self::assertEquals(('\\' === DIRECTORY_SEPARATOR), Platform::isWindows());
         self::assertEquals(defined('PHP_WINDOWS_VERSION_MAJOR'), Platform::isWindows());
     }
+
+    /**
+     * @return iterable<array{0: bool}>
+     */
+    public function defaultProvider(): iterable
```
Seldaek:
```suggestion
    public static function defaultProvider(): iterable
```

doc/03-cli.md
glaubinix: Makes sense. I reverted the change. This now always applies the ENV. I takes the disabled state of the config e.g. `{"audit": "ignore", "block": false}` and converts it in to `{"audit": "ignore", "block": true}`

src/Composer/Policy/PolicyConfig.php
```diff
@@ -185,24 +185,49 @@ public static function fromConfig(Config $config): self
...
+
+        // COMPOSER_POLICY_ABANDONED_BLOCK is the canonical name following the
+        // COMPOSER_POLICY_<LIST>_BLOCK pattern; COMPOSER_SECURITY_BLOCKING_ABANDONED
+        // is the legacy alias and only applies when the canonical var is unset.
+        $canonicalAbandonedBlockOverride = Platform::getBoolEnv('COMPOSER_POLICY_ABANDONED_BLOCK');
+        $legacyAbandonedBlockOverride = Platform::getBoolEnv('COMPOSER_SECURITY_BLOCKING_ABANDONED');
+        if ($canonicalAbandonedBlockOverride !== null || $legacyAbandonedBlockOverride !== null) {
             $abandoned = new AbandonedPolicyConfig(
-                (bool) (int) $blockAbandonedEnv,
+                $canonicalAbandonedBlockOverride !== null ? $canonicalAbandonedBlockOverride : $legacyAbandonedBlockOverride,
```
Seldaek: I think this is equivalent no?

```suggestion
                $canonicalAbandonedBlockOverride ?? $legacyAbandonedBlockOverride,
```

Seldaek: Or even this:

```suggestion
        $abandonedBlockOverride = Platform::getBoolEnv('COMPOSER_POLICY_ABANDONED_BLOCK') ?? Platform::getBoolEnv('COMPOSER_SECURITY_BLOCKING_ABANDONED');
        if ($abandonedBlockOverride !== null) {
$abandoned = new AbandonedPolicyConfig(
                $abandonedBlockOverride,
```

src/Composer/Config.php
Seldaek: Somehow cannot do a code suggestion here but think the outer if and getComposerEnv call can be removed entirely?

src/Composer/Config.php
```diff
@@ -547,9 +547,9 @@ public function get(string $key, int $flags = 0)
...
                 // to rewrite the raw array shape.
-                $policyEnv = $this->getComposerEnv('COMPOSER_POLICY');
-                if (false !== $policyEnv) {
-                    if (false === Platform::getBoolEnv('COMPOSER_POLICY')) {
+                $policyEnv = Platform::getBoolEnv('COMPOSER_POLICY');
+                if (null !== $policyEnv) {
+                    if (false === $policyEnv) {
                         $policyConfig = false;
                     }
                 }
```
Seldaek: Really this is all that's needed here. The outer if does nothing.

```
if (false === Platform::getBoolEnv('COMPOSER_POLICY')) {
    $policyConfig = false;
}
```

PR #12807, opened 2026-04-17
FilterList: add ignore-source for malware list

glaubinix:
Follow up PR for https://github.com/composer/composer/pull/12804 to add handling of `ignore-source` for the malware policy list.


src/Composer/FilterList/FilterListAuditor.php
```diff
@@ -122,4 +127,40 @@ private function isPackageIgnored(string $packageName, ConstraintInterface $pack
...
+            foreach ($entries[MalwarePolicyConfig::NAME] as $malwareEntry) {
+                if (!in_array($malwareEntry->source, $ignoreSource, true)) {
+                    $packageEntries[] = $malwareEntry;
+                }
+            }
+
+            if (count($packageEntries) > 0) {
+                $filterListMap[$packageName][MalwarePolicyConfig::NAME] = $packageEntries;
+            } else {
+                unset($filterListMap[$packageName][MalwarePolicyConfig::NAME]);
```
Seldaek, line 160: Wonder if we should clean up here.. not sure if this has any side effect to have empty packages left over but seems cleaner.

```suggestion
                unset($filterListMap[$packageName][MalwarePolicyConfig::NAME]);
                if (count($filterListMap[$packageName]) === 0) {
                    unset($filterListMap[$packageName]);
                }
```

PR #12809, opened 2026-04-17
FilterLists: add support for the policy config to the ConfigCommand

glaubinix:
Follow up PR for https://github.com/composer/composer/pull/12804 to add support for the policy config to the `ConfigCommand` and add a new `PolicyCommand` to add sources easier


Seldaek:
One test failing but aside from that LGTM now 👍🏻

PR #12817, opened 2026-04-27
Add note on installing runtime dependencies

jnoordsij:
Added note about manually installing runtime dependencies.

As a follow-up to #12580 and the discussion at https://github.com/composer/composer/discussions/12278, this PR proposes to add some explicit note about runtime dependencies to the Docker usage documentation.


Seldaek:
Thanks

Issue #11273, opened 2023-01-17
Feature Request: Control in which order composer autoloads includes files


snake-py:
Hello,

I am maintaining multiple packages which publish via ./src/helper.php some global helper methods. Therefore I have in the composer.json inside the packages defined:

```json
    "autoload": {
        "files": [
            "./src/helpers.php"
        ],
    } 
```
However, inside my Laravel projects, I want to overwrite some of these helpers, therefore I would like that app/helper.php is loaded before vendor/mypackage/src/helper.php

in the laravel project, I have defined:
```json
    "autoload": {
        "files": [
            "./app/helpers.php"
        ],
```
I think currently there is no way to tell composer that it should first load the project base helper file.

A concrete example I am trying to override would be this. In my package for testing (and as a fallback to avoid crashes) purposes I have defined:

```php
if (!function_exists('getEnvironment')) {
    function getEnvironment(): string
    {
        return config('wd.logging.environment', 'DEFAULT_ENV');
    }
}
```
and in my project I need to overwrite it with this function (this function should be used in production):
```php
if (!function_exists('getEnvironment')) {
    function getEnvironment(): string
    {
        return Auth::user()
            ? getSetting(Auth::user()->client->settings, Setting::KEY_SHORT_CODE)
            : null;
    }
}
```

stof:
Composer will require files from a package after the files from all its dependencies (and so the root package is the last one). This ensures that a package file has access to the functions defined in its dependencies.

snake-py:
@stof I understand this and this behavior is solid. However, I would love that I can preload global files to overwrite files from dependencies. I found a composer plug in which does this already, but I thought this should be also be included into the core concept.  
maybe something like:

```json
    "autoload": {
        "perload": [
            "app/helpers.php"
        ]
    },
``` 

this is the plugin I started using to have this behavior https://github.com/funkjedi/composer-include-files

Seldaek:
If you want to override in the project you can include your own files before you include `vendor/autoload.php`, that will ensure they are always loaded first. We are not going to support more fine grained use cases than this because they are IMO bad practices. Perhaps we will one day get native function autoloading that will allow improving the situation..

snake-py:
@Seldaek Could you elaborate on why this is bad practice? It seems that this is how Laravel wants it or do I misunderstand it?

Seldaek:
[...]

PR #12820, opened 2026-04-27
Prepend autoloader files

djmetzle:
Add a new configuration option to prepend autoload files to the files list.

This allows making sure that root project files are autoloaded _before_ third party files.

As a specific example, this allows overriding global functions added by third party libraries, like illuminate. See:
https://github.com/illuminate/collections/blob/8955cc9996739381c2865ee355093dabae121cc9/helpers.php#L227-L236 

Ref: https://github.com/composer/composer/issues/11273


Seldaek:
If you want to override you should just include your files before including `vendor/autoload.php`. I think that's better/cleaner than a new option here because it gives you more control.

djmetzle:
@Seldaek 
> If you want to override you should just include your files before including vendor/autoload.php. I think that's better/cleaner than a new option here because it gives you more control.

That's fair! And I appreciate the input.

---

I suppose the spirit of this change would be to introduce flexibility.

Currently the AutoloadGenerator sorts and then reverses the package map:
https://github.com/composer/composer/blob/340dda1df38ff306d20db3c3b109ace278c69f12/src/Composer/Autoload/AutoloadGenerator.php#L596-L600

Packages are sorted in dependency order, and otherwise _alphabetically_:
https://github.com/composer/composer/blob/340dda1df38ff306d20db3c3b109ace278c69f12/src/Composer/Util/PackageSorter.php#L66-L68

I found that behavior to be very unintuitive; I would expect that the root package would have its files included before the `files` specifications from dependencies.

---

The existence of [the plugin](https://github.com/funkjedi/composer-include-files) shows there is an appetite for that behavior.

Specifically in our case, using the plugin is problematic, as there is a brief window between main autoload being written, and the plugin _rewriting it_ to include the extra `files` entries.

It could be nice to be able to enable behavior that someone might expect to be the default behavior?

---

Admittedly we could in fact manually `require` in the global `files` entries we need first, before including the autoloader. That is a viable solution.

But it is a minor maintenance burden to need to explicitly work around the autoloader.

Seldaek:
Yeah, we had to pick something, and *usually* you want your dependencies loaded before dependents, so that's what we do, and there is another way (include first) + that plugin apparently to tweak this for the few odd cases that do not fit this pattern. IMO this is good enough and not something I really want to spend time/effort supporting.

djmetzle:
All good! Thanks @Seldaek 

Low risk/effort as an experiment. Thanks for the feedback. We'll inline the require.

PR #12821, opened 2026-04-28
GitHub Actions: use zizmor to harden existing workflows

glaubinix:
This introduces [zizmor](https://github.com/zizmorcore/zizmor) a static analyzer for GitHub Actions to harden existing GitHub Action workflows and adds a new workflow to analyze any future changes to those workflows.

I didn't introduce any new concurrency limitations. Happy to add that if wanted but don't think this is necessary.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,6 +1,6 @@
 name: "Continuous Integration"
 
-on:
+on: # zizmor: ignore[concurrency-limits]
   push:
     paths-ignore:
       - 'doc/**'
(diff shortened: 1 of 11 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -1,6 +1,6 @@
 name: "Continuous Integration"
 
-on:
+on: # zizmor: ignore[concurrency-limits]
```
Seldaek: Here too fixing those would be nice IMO

Seldaek:
Thanks!

PR #12826, opened 2026-05-06
FilterList: block malware packages during composer install

glaubinix:
Follow up PR for https://github.com/composer/composer/pull/12804 to add block malware packages during `composer install`

### DIFF
--- a/src/Composer/Installer.php
+++ b/src/Composer/Installer.php
@@ -38,6 +38,7 @@ use Composer\Filter\PlatformRequirementFilter\PlatformRequirementFilterInterface
 use Composer\FilterList\FilterListAuditor;
 use Composer\FilterList\FilterListProvider\FilterListProviderSet;
 use Composer\Installer\InstallationManager;
+use Composer\Policy\ListPolicyConfig;
 use Composer\Policy\PolicyConfig;
 use Composer\Installer\InstallerEvents;
 use Composer\Installer\SuggestedPackagesReporter;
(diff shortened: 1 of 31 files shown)


src/Composer/Installer.php
```diff
@@ -1166,15 +1167,22 @@ private function createSecurityAuditPoolFilter(): ?SecurityAdvisoryPoolFilter
...
         $policyConfig = $this->getPolicyConfig();
 
         if (!$policyConfig->enabled || $this->updateMirrors) {
             return null;
         }
 
-        return new FilterListPoolFilter($policyConfig, new FilterListAuditor(), $this->repositoryManager->getHttpDownloader());
+        if ($blockScope === ListPolicyConfig::BLOCK_SCOPE_INSTALL && !$policyConfig->malware->shouldBlock($blockScope)) {
+            return null;
+        }
```
Seldaek: Claude review flagged this which seems valid:

Hardcodes the assumption that malware is the only list that ever blocks on install. That's true today (it's the only one overriding supportsInstallBlockScope()), but the rest of this PR is explicitly built around the abstraction so any future list could opt in. The more abstraction‑faithful guard would be:

```suggestion
        if ($blockScope === ListPolicyConfig::BLOCK_SCOPE_INSTALL && \count($policyConfig->getActiveBlockFilterLists($blockScope)) ===
     0) {
            return null;
        }
```

Equivalent for malware today; survives a future list opting in. Worth a one‑liner change.

Seldaek:
One thing I find a bit odd here, but I am not sure if it's really a problem or just a mismatch of what I expected to see in the PR:

It seems you used fixed packages from the request, and not locked packages. In `install` runs, what happens is we mark all locked packages as fixLockedPackage in the request:

https://github.com/composer/composer/blob/e8ff2daa603a0b30be74068e8300b5829534779f/src/Composer/Installer.php#L785-L787

So technically if you filter malware against fixed packages, it will work. BUT it would be semantically more correct to check against locked packages IMO. One problem I see is that Request::isLockedPackage checks both locked and fixedlocked but getLockedPackages was never updated to reutrn fixedlocked ones as well. So that's maybe why it broke and required you to use getFixedPackages.

The only real "problem" of the current code is that it works on fixed packages which it should not IMO semantically AND that it allows platform packages to be marked as malware and made uninstallable which would be messy but also unlikely to happen obviously if you only use trusted filter list sources.

PR #12831, opened 2026-05-07
FilterLists: adjust auditor and audit config to adapt for new filter lists

glaubinix:
Follow up PR for https://github.com/composer/composer/pull/12804 to adjust the audit command for the new policy format. This addresses two issues:
* having both an audit config and a policy config easily lead to them being out of sync
* with the audit config it was impossible to have different audit outcomes for the different lists


Seldaek:
Thanks!

PR #12833, opened 2026-05-08
FilterLists: introduce a summary-url ComposerRepositories can implement to skip fetching all metadata files for all packages

glaubinix:
Follow up PR for https://github.com/composer/composer/pull/12804. Addresses the issue where before during a `composer install` run, Composer would try to fetch metadata information for all packages from packagist.org. This now first fetches a summary-url and only fetches the metadata for packages appearing in the summary URL. At the same time this skips the summary-url if we have already fetched metadata e.g. during a `composer update`.

This needs a corresponding PR on the packagist.org side but otherwise this PR is ready.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -340,7 +340,8 @@ describes what lists are available.
         "lists": {
             "malware": { "enabled": true },
             "typosquatting": { "enabled": true }
-        }
+        },
+        "summary-url": "/p2/filter-summary.json"
     }
 }
 ```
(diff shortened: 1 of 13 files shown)


doc/05-repositories.md
```diff
@@ -338,7 +338,7 @@ describes what lists are available.
     "filter": {
         "metadata": true,
         "lists": ["aikido-malware", "other-malware"],
```
Seldaek:
```suggestion
        "lists": ["malware", "custom-filters"],
```

src/Composer/Repository/ComposerRepository.php
```diff
@@ -760,6 +760,32 @@ public function getFilter(array $packageConstraintMap): array
             }
         }
 
+        // skip the summary fetch if we already pulled package metadata in this run; per-package
+        // calls below will short-circuit on the cache, so the summary would be wasted work.
+        if ($this->filterConfig !== null && $this->filterConfig->summaryUrl !== null && $this->freshMetadataUrls === []) {
```
Seldaek, line 815: I think it's *probably* fine to leave it as is, but just wanted to say I think this may not do what you think it does in some cases. Like a require command will probably fetch some metadata file which will mean freshMetadataUrls isn't empty, and then do a partial update + install which will then end up fetching more files just for the filter instead of using the summary file in that case. Would need to check tho.

glaubinix: Yes, aware of that. The alternative would be to manually check first if all packages we need are already in `freshMetadataUrls` or `packagesNotFoundCache` and then call the summary URL if one or more packages are still missing. Happy to further modify this.

Seldaek: Nah I don't think it's really worth it right now.

PR #12834, opened 2026-05-08
FilterList: assert that malware blocking still applies during partial updates

glaubinix:
Follow up PR to https://github.com/composer/composer/pull/12826

During a partial update the solver isn't run again during the install step. Therefore the pool filter needs to apply install blocking during the update step already.


Seldaek:
One test fails but it looks to me like the behavior is correct and test expectation outdated.

glaubinix:
Thank you! Updated the test that was indeed still from an earlier implementation that produced a different error output.

Seldaek:
Great, thanks

PR #12836, opened 2026-05-10
FilterLists: COMPOSER_POLICY=1 should enable filtering if the config is disabled

glaubinix:
Follow up PR for https://github.com/composer/composer/pull/12804. So far we only handled COMPOSER_POLICY=0. This now also covers COMPOSER_POLICY=1.


Seldaek:
Thanks

PR #12838, opened 2026-05-11
FilterLists: adjust parsing of lists information in packages.json

glaubinix:
This now handles packages.json responses like
```
{
    "filter": {
        "metadata": true,
        "lists": {
            "malware": true
        }
    }
}
```

 See https://github.com/composer/packagist/pull/1721

### DIFF
--- a/res/composer-repository-schema.json
+++ b/res/composer-repository-schema.json
@@ -59,10 +59,10 @@
                     "description": "Whether metadata files contain filter list data."
                 },
                 "lists": {
-                    "type": "array",
-                    "description": "Names of all filter lists provided by this repository.",
-                    "items": {
-                        "type": "string"
+                    "type": "object",
+                    "description": "Names of all filter lists provided by this repository, mapped to true. The boolean value is a presence marker only; the object shape exists for forward compatibility so future per-list metadata can be added without another wire-format break.",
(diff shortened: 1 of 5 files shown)


res/composer-repository-schema.json
```diff
@@ -59,10 +59,10 @@
                     "description": "Whether metadata files contain filter list data."
                 },
                 "lists": {
-                    "type": "array",
-                    "description": "Names of all filter lists provided by this repository.",
-                    "items": {
-                        "type": "string"
+                    "type": "object",
+                    "description": "Names of all filter lists provided by this repository, mapped to true. The boolean value is a presence marker only; the object shape exists for forward compatibility so future per-list metadata can be added without another wire-format break.",
```
stof, line 63: changing values from a boolean `true` to another kind of values will be a wire-format break as well. Changing the type of an existing value is a break, while adding new keys in an object is not.

glaubinix: This should be fine though if we update the schema now to define that the value can be boolean or object, right?

Seldaek: Might be safer to have it as {enabled : bool} for now. This can be a tad verbose it's just for repo maintainers to see/write really.

PR #12839, opened 2026-05-11
FilterLists: add support for api-url to fetch filter list entries

glaubinix:
In case the summary-url file is getting too big, we want to offer an alternative. Composer repositories can implement an api-url which accepts a list of PURLs and lists as a POST request and should then return a list of matching filter entries.

### DIFF
--- a/doc/05-repositories.md
+++ b/doc/05-repositories.md
@@ -374,6 +374,44 @@ describes what lists are available.
   Composer revalidates the cached summary via `If-Modified-Since` on every install/update/audit, using the
   same on-disk repository cache as package metadata.
 
+- **`api-url`** (optional, string) — A URL (absolute or root-relative) that accepts a POST request
+  with the relevant package PURLs and configured list names, and replies with the matching filter
+  entries directly. When set, Composer uses `api-url` instead of `summary-url` and per-package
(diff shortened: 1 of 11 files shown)


doc/05-repositories.md
```diff
@@ -374,6 +374,44 @@ describes what lists are available.
   Composer revalidates the cached summary via `If-Modified-Since` on every install/update/audit, using the
   same on-disk repository cache as package metadata.
 
+- **`api-url`** (optional, string) — A URL (absolute or root-relative) that accepts a POST request
+  with the relevant package PURLs and configured list names, and replies with the matching filter
+  entries directly. When set, Composer uses `api-url` instead of `summary-url` and per-package
+  metadata for filter purposes; this is useful when the summary would be too large to serve in
+  full. If both `summary-url` and `api-url` are advertised, `api-url` takes precedence and
+  `summary-url` is ignored.
```
Seldaek:
```suggestion
  `summary-url` is ignored. Implementors should be aware that large amounts (a few hundreds
  would be normal) of package names can be submitted by Composer. You can decide to cap
  that list and return only the first 1000 or whatever you decide but make sure to limit this
  somehow.
```

Seldaek:
Thanks

PR #12842, opened 2026-05-12
Add option to block downloads based on URL before initializing

jdworschak:
Adds a `prevent_url_access_callable` option to `HttpDownloader`. The callable is invoked synchronously at the start of `CurlDownloader::initDownload()` before `curl_init()`. If it returns true, the request is rejected with a `TransportException` and no connection is opened.

This complements the existing `prevent_ip_access_callable` (#11895), which can only fire after curl has resolved DNS and begun the TCP connection. At that point the request has already been started and is then aborted. The new option also works for redirects as they go through `restartJob()` → `initDownload()`, so each new URL is vetted before its own request begins.

`prevent_ip_access_callable` is intentionally left in place for potential BC reasons, but also because it can still protect from DNS rebinding after an initial lookup has been made which the new option wouldn't be able to cover alone.

`RemoteFilesystem` rejects the option, mirroring the existing IP-filter handling.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -181,6 +181,14 @@ class CurlDownloader
             $this->config->prohibitUrlByConfig($url, $this->io, $options);
         }
 
+        if (
+            isset($options['prevent_url_access_callable']) &&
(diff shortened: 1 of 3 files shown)


Seldaek:
Looking good thanks!

PR #12845, opened 2026-05-12
FilterList: fix reason + URL output once per matching version

glaubinix:
See change in test fixture


Seldaek:
Nice cleanup thanks

PR #12846, opened 2026-05-12
FilterLists: add warning in case security advisories or filter information cannot be fetched

glaubinix:
This will generate an output like below in case one or more repository/source cannot be accessed.

```
composer install
Filter list data could not be fetched from some sources (ignored per policy.ignore-unreachable); matches may be incomplete:
  - The "https://repo.packagist.com/filter/lists/summary.json" file could not be downloaded: HTTP/1.1 502 Bad
  Gateway
Generating autoload files
```


Seldaek:
Thanks

PR #12852, opened 2026-05-13
fix: prevent logging the github oauth token in plaintext

damienwebdev:


DavidLambauer:
The token change announcement for reference

https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/

Seldaek:
Closed by https://github.com/composer/composer/pull/12853 - it is now available as https://github.com/composer/composer/releases/tag/2.9.8

PR #12853, opened 2026-05-13
Fix regexp for new GitHub installation tokens

pscheit:


src/Composer/IO/BaseIO.php
```diff
@@ -139,9 +139,10 @@ public function loadConfiguration(Config $config)
 
             // allowed chars for GH tokens are from https://github.blog/changelog/2021-03-04-authentication-token-format-updates/
```
GromNaN, line 140: This comment is outdated. The regex provided in the blog post is the one that was fixed by this PR.

Seldaek: Should anyway be resolved by https://github.com/composer/composer/pull/12856

PR #12856, opened 2026-05-13
Relax token validation on input, and hide more things on output

Seldaek:

### DIFF
--- a/src/Composer/IO/BaseIO.php
+++ b/src/Composer/IO/BaseIO.php
@@ -128,11 +128,6 @@ abstract class BaseIO implements IOInterface
         }
 
         foreach ($githubOauth as $domain => $token) {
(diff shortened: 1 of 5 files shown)


defunctl:
Heya @Seldaek 

Any updates on this? 

We are seeing 1 in 4 action runs failing with `Error:  Your github oauth token for github.com contains invalid characters: "***"`, so sometimes GitHub is generating a new token format that composer doesn't like.

stof:
@defunctl make sure you are using the latest composer version (see https://blog.packagist.com/composer-2-9-8-and-2-2-28-fix-github-actions-token-disclosure-in-error-messages/)

The change in this PR removes more validation, but it is not about supporting the new github format that it being rolled out gradually these days. It is about avoiding issues for *future* format changes.

defunctl:
@stof We are, actually I realized this is a problem with [Strauss](https://github.com/BrianHenryIE/strauss), I'll make an issue there :)

PR #12860, opened 2026-05-14
perf: early return instead of full array_filter run

jorgsowa:
Changes `array_filter()` on whole array to early return for first value.


Seldaek:
This one could use array_find

Seldaek:
Thanks

PR #12861, opened 2026-05-14
perf: early returns in for loops in Problem.php

jorgsowa:
Early returns in for loops in Problem.php instead of full iterations of `array_filter()`.


Seldaek:
I guess using array_any would be best here. We have polyfills for 8.4 so should be fine.

jorgsowa:
> We have polyfills for 8.4 so should be fine.

I wasn't aware of it. I will make changes to my two PRs then.

Seldaek:
Thanks

PR #12864, opened 2026-05-16
Expose logic for using alternative domain in AuthHelper

eiriksm:
This PR moves the two lines of logic that ends up placing auth on an alternate domain.

The reason for this, is if you wanted to retroactively determine if `IO` contains authentication that would end up being used by composer, you would have to duplicate the logic. Something like

```
if (!$this->io->hasAuthentication($urlHost) && in_array($urlHost, ['api.github.com', 'api.bitbucket.org'], true)) {
    $urlHost = str_replace('api.', '', $urlHost);
}
```

which is short and sweet, but just a copy paste from a moment in time. If this logic was ever changed, one would not notice and would have to manually track this duplication. Which is why I instead added a static helper, so you can pass your `IO` object and get the same result as composer would have.

### DIFF
--- a/src/Composer/Util/AuthHelper.php
+++ b/src/Composer/Util/AuthHelper.php
@@ -273,8 +273,6 @@ class AuthHelper
         $authOrigin = self::findAuthOrigin($this->io, $origin);
         if ($authOrigin !== null) {
             $origin = $authOrigin;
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #12865, opened 2026-05-18
FilterLists: improve fallback handling for config.audit.* keys

glaubinix:
This updates/adds:
* section in the docs that explains how we handle configs where both policy and audit is set
* tests to verify the behaviour
* JSON schema to deprecate audit properties

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -580,20 +580,22 @@ dependencies, ensuring they cannot be installed.
 The legacy `config.audit` keys are only read as a fallback when the corresponding
 [`config.policy`](#policy) block is **absent**. The fallback is all-or-nothing per built-in list:
 
-- If [`config.policy.advisories`](#advisories) is set (to any value, including `false`), every
-  `audit.*` key that maps to advisories — [`block-insecure`](#block-insecure),
-  [`ignore`](#ignore), [`ignore-severity`](#ignore-severity) — is **ignored entirely**.
-  Only `policy.advisories.*` is read. Mix-and-matching, e.g. setting `policy.advisories.block`
-  while expecting `audit.ignore-severity` to still apply, is not supported — migrate all
-  advisories-related settings together.
-- If [`config.policy.abandoned`](#abandoned) is set (to any value, including `false`), every
-  `audit.*` key that maps to abandoned — [`block-abandoned`](#block-abandoned),
-  [`abandoned`](#abandoned), [`ignore-abandoned`](#ignore-abandoned) — is **ignored
-  entirely**. Only `policy.abandoned.*` is read.
+- If [`config.policy.advisories`](#advisories) is set (to any value, including `false`),
+  every advisories-related `audit.*` key ([`audit.block-insecure`](#block-insecure), [`audit.ignore`](#ignore-3),
+  [`audit.ignore-severity`](#ignore-severity-1)) is **ignored entirely** — only
+  [`policy.advisories.block`](#block), [`policy.advisories.ignore`](#ignore), and
+  [`policy.advisories.ignore-severity`](#ignore-severity) are read. Mix-and-matching,
+  e.g. setting `policy.advisories.block` while expecting `audit.ignore-severity` to still
+  apply, is not supported — migrate all advisories-related settings together.
+- If [`config.policy.abandoned`](#abandoned) is set (to any value, including `false`),
+  every abandoned-related `audit.*` key ([`audit.block-abandoned`](#block-abandoned), [`audit.abandoned`](#abandoned-1),
+  [`audit.ignore-abandoned`](#ignore-abandoned)) is **ignored entirely** — only
+  [`policy.abandoned.block`](#block-1), [`policy.abandoned.audit`](#audit-1), and
+  [`policy.abandoned.ignore`](#ignore-1) are read.
 - The two built-in lists are independent: configuring `policy.advisories` while leaving the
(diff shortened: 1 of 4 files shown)


doc/06-config.md
```diff
@@ -571,6 +571,26 @@ and short format versions are automatically reported at the end of update or req
...
+  while expecting `audit.ignore-severity` to still apply, is not supported — migrate all
+  advisories-related settings together.
+- If [`config.policy.abandoned`](#abandoned) is set (to any value, including `false`), every
+  `audit.*` key that maps to abandoned — [`block-abandoned`](#block-abandoned),
+  [`abandoned`](#abandoned), [`ignore-abandoned`](#ignore-abandoned) — is **ignored
+  entirely**. Only `policy.abandoned.*` is read.
+- The two built-in lists are independent: configuring `policy.advisories` while leaving the
+  abandoned settings under `audit.*` is allowed and vice versa.
+- [`audit.ignore-unreachable`](#ignore-unreachable) is superseded by
+  [`policy.ignore-unreachable`](#ignore-unreachable) as soon as the latter is set.
```
Seldaek: Applies to the above block:

  1. Anchor link resolution in the new docs section (medium). GitHub disambiguates duplicate headings by appending -1, -2, … in file order. The new
   section enumerates audit.* keys but the links land on policy.* headings because those come first in the file:

  - [ignore](#ignore) resolves to #### ignore at line 262 (policy.advisories.ignore), not ### ignore at line 574 (audit.ignore).
  - [ignore-severity](#ignore-severity) resolves to #### ignore-severity at line 267 (under policy.advisories), not the audit one at line 735.
  - [abandoned](#abandoned) resolves to ### abandoned at line 302 (policy.abandoned), not the audit one at line 642.
  - Most importantly, [audit.ignore-unreachable](#ignore-unreachable) resolves to line 419 (policy.ignore-unreachable) — the exact opposite of the
  link's label. The audit version is #ignore-unreachable-1.

  The bullet that says "audit.ignore-unreachable is superseded by policy.ignore-unreachable" with both pointing at the same anchor is the most
  visible bug. Worth either (a) fixing the audit-side links to use -1 suffixes, or (b) inverting the framing so the labels match the resolved
  anchors (e.g., link the "use X instead" target instead of the deprecated name).

doc/06-config.md
```diff
@@ -571,6 +571,26 @@ and short format versions are automatically reported at the end of update or req
 discards package versions identified as insecure or abandoned, depending on configuration, before resolving
 dependencies, ensuring they cannot be installed.
 
+### How `config.audit` interacts with `config.policy`
+
+The legacy `config.audit` keys are only read as a fallback when the corresponding
+[`config.policy`](#policy) block is **absent**. The fallback is all-or-nothing per built-in list:
```
Seldaek, line 581: Maybe we should cross-link this and mention is as well from the Policy config section?

res/composer-schema.json
```diff
@@ -442,9 +442,12 @@
                 },
                 "audit": {
                     "type": "object",
-                    "description": "Security audit and version blocking configuration options",
+                    "deprecated": true,
+                    "description": "Deprecated: use 'config.policy' instead. All 'audit.*' keys still work as a fallback, but the fallback is all-or-nothing per built-in list — once 'policy.advisories' is set every 'audit.*' advisories key is ignored, and once 'policy.abandoned' is set every 'audit.*' abandoned key is ignored.",
```
```suggestion
                    "description": "Deprecated: use 'config.policy' instead. All 'audit.*' keys still work as a fallback for now. See https://getcomposer.org/doc/06-config.md#policy",
```

Or ideally we'd deep link to the 
```
How `config.audit` interacts with `config.policy`
``` 
section here maybe?

res/composer-schema.json
```diff
@@ -530,19 +536,24 @@
...
-                            "description": "Whether insecure versions should be blocked during a composer update/require command or not.",
+                            "deprecated": true,
+                            "description": "Deprecated: use 'config.policy.advisories.block' instead. Whether insecure versions should be blocked during a composer update/require command or not.",
                             "default": true
                         },
                         "block-abandoned": {
                             "type": "boolean",
-                            "description": "Whether abandoned packages should be blocked during a composer update/require command or not. Applies only if blocking of insecure versions is enabled.",
+                            "deprecated": true,
+                            "description": "Deprecated: use 'config.policy.abandoned.block' instead. Whether abandoned packages should be blocked during a composer update/require command or not.",
```
Seldaek: Other mentions of deprecations in this file currently all caps it (but don't have deprecated: true set consistently, feel free to fix that :)

```suggestion
                            "description": "DEPRECATED: use 'config.policy.abandoned.block' instead. Whether abandoned packages should be blocked during a composer update/require command or not.",
```

doc/06-config.md
glaubinix: oh, just noticed that website handles this different than GitHub. On GitHub this is ignore-unreachable-1 and on the website it's ignore-unreachable-2

doc/06-config.md
glaubinix: done

res/composer-schema.json
glaubinix: I think now that cross-link this should be clearer already

doc/06-config.md
glaubinix: Adjusting the website in https://github.com/composer/getcomposer.org/pull/260

Seldaek:
Thanks

PR #12868, opened 2026-05-19
Problem: surface hidden constraint chain in CDCL conflict output

### LINKED ISSUE #12862: reasons for install failure aren't clear
My `composer.json`:

```json
It's too complicated - I have drupal/core installed as a path repo from a git clone. Sorry.
```

(Also, I'm about to change the packages I maintain so it satisfies the constraint, so in about 5 minutes this won't be reproducible anyway!)

When I try to require `drupal-code-builder/drupal-code-builder` ^4, I get this error:

```

  Problem 1
    - drupal/admin_toolbar is locked to version 3.6.3 and an update of this package was not requested.
    - Root composer.json requires drupal/module_builder ^4 -> satisfiable by drupal/module_builder[4.0.0, ..., 4.2.2].
    - drupal/admin_toolbar 3.6.3 requires drupal/core ^9.5 || ^10 || ^11 || dev-main || dev-main -> satisfiable by drupal/core[dev-main].
    - drupal/core 12.x-dev requires symfony/finder ^8.0 -> satisfiable by symfony/finder[v8.0.0-BETA1, ..., 8.1.x-dev].
    - Conclusion: install symfony/finder v7.4.8 (conflict analysis result)
    - drupal/core 12.x-dev is an alias of drupal/core dev-main and must be installed with it.
    - You can only install one version of a package, so only one of these can be installed: symfony/finder[v7.0.0-BETA1, ..., 7.4.x-dev, v8.0.0-BETA1, ..., 8.1.x-dev].
```

which fails to state the ACTUAL problem which is that drupal/core `main` requires symfony/finder ^8, while drupal-code-builder/drupal-code-builder requires ^7 (and earlier).

Doing `composer why-not drupal-code-builder/drupal-code-builder` also doesn't give the full picture:

````
drupal/module_builder                     3.0.0-beta11 requires         drupal-code-builder/drupal-code-builder (^3.2.0)
drupal-code-builder/drupal-code-builder   4.6.4        requires         symfony/finder (^4.0 || ^5 || ^6 || ^7)
joachim-n/drupal-core-development-project dev-main     does not require symfony/finder (but v8.0.8 is installed)
Not finding what you were looking for? Try calling `composer update "drupal-code-builder/drupal-code-builder:^4" --dry-run` to get another view on the problem.
````

### PR DESCRIPTION
**Pull request by Seldaek**
Fixes #12862 (maybe/allegedly.. kinda hard to be sure without clear repro case, and this is 100% claude work)

When the SAT solver's conflict-driven clause learning fires, each learned rule prints as "Conclusion: <decision> (conflict analysis result)" without any explanation of *why* that conclusion holds. The rules walked during analyze() to derive the learned clause are recorded in $learnedPool but have never been surfaced - the existing formatter has a @TODO sketching how to expand them but it's disabled because the original attempt produced too much output and risked infinite recursion.

In practice this leaves users staring at output like:

    - drupal/core dev-main requires symfony/finder ^8.0 -> ...
    - Conclusion: install symfony/finder v7.4.8 (conflict analysis result)
    - You can only install one version of a package, so only one of these can be installed: symfony/finder[^7 versions, ^8 versions].

...with no way to know which package transitively required symfony/finder ^7. That rule was walked by the solver during conflict analysis but lives only in $learnedPool[reasonData], never in $problem->reasons.

This change expands learned rules at the Problem level (not the Rule level the original @TODO contemplated) so the existing dedup + version-range template machinery in formatDeduplicatedRules absorbs primitives that were already visible. The user-visible delta is exactly the previously- hidden chain.

The expansion is a bounded BFS:

- Walks $learnedPool[reasonId] following nested learned rules.
- Collects only non-LEARNED ("primitive") rules.
- Cycle-safe via a reasonId visited-set (object-hash dedup would over-prune: Solver::analyze() appends the same Rule object across iterations).
- Capped at 50 surfaced primitives per problem. Breadth, not depth, is the real risk - $learnedPool[id] can be wide.

Problem-level injection (vs the originally-commented Rule-level approach) matters because formatDeduplicatedRules's array_unique on message strings and its requires/conflicts template collapse only work when all rules flow through one pass. Rule-level expansion would emit a nested block that bypasses both, re-introducing duplicates and breaking the [v1, ..., vN] range collapse.

The "(conflict analysis result)" suffix is preserved deliberately - it names the learned literal (the decision the solver was forced into), which the surrounding primitives don't. With the primitives now visible in the same problem block, the cryptic phrase becomes a signpost rather than the sole source of information.

Two fixtures change:

- github-issues-12862.test (new): purpose-built repro mirroring the issue's locked-transitive vs new-transitive shape. Multi-version new/pkg + new/dep with distinct shared/lib constraints plus a wide shared/lib version space forces CDCL. Without this fix the output hides "new/dep 1.1.0 requires shared/lib ~1.5" - the very chain link the user needs to understand why "Conclusion: don't install new/dep 1.1.0" was reached. With this fix that line appears.

- github-issues-7051.test: incidental real-world demonstration. Before: only "friendsofphp/php-cs-fixer v2.10.4 requires symfony/console ..." was visible; the same rule for v2.10.5 was hidden in $learnedPool. After: both versions collapse via the existing template into "friendsofphp/php-cs-fixer[v2.10.4, ..., v2.10.5] require ...".

Out of scope: the pool-optimizer fallback "found X but these were not loaded, likely because it conflicts with another require" message lives in a different code path that has no $learnedPool reference at the call site; improving it is a separate problem.

### DIFF
--- a/tests/Composer/Test/Fixtures/installer/github-issues-12862.test
+++ b/tests/Composer/Test/Fixtures/installer/github-issues-12862.test
@@ -1,0 +1,95 @@
+--TEST--
+GitHub issue #12862: when CDCL conflict analysis fires due to a locked package's transitive constraint conflicting with a new require's branches, the transitive rules walked during analysis that are NOT in $problem->reasons must still be surfaced so users can see the full chain that justified each "Conclusion: ... (conflict analysis result)" line.
+--COMPOSER--
+{
+    "repositories": [
+        {
+            "type": "package",
+            "package": [
+                { "name": "locked/pkg", "version": "1.0.0", "require": { "locked/dep": "^1" } },
+                { "name": "locked/dep", "version": "1.0.0", "require": { "shared/lib": "^2" } },
+
+                { "name": "new/pkg",   "version": "1.0.0", "require": { "new/dep": "^1" } },
+                { "name": "new/pkg",   "version": "1.1.0", "require": { "new/dep": "^1" } },
+
+                { "name": "new/dep",   "version": "1.0.0", "require": { "shared/lib": "^1.0" } },
+                { "name": "new/dep",   "version": "1.1.0", "require": { "shared/lib": "~1.5" } },
+
+                { "name": "shared/lib", "version": "1.0.0" },
+                { "name": "shared/lib", "version": "1.0.1" },
+                { "name": "shared/lib", "version": "1.1.0" },
+                { "name": "shared/lib", "version": "1.2.0" },
+                { "name": "shared/lib", "version": "1.5.0" },
+                { "name": "shared/lib", "version": "1.5.1" },
+                { "name": "shared/lib", "version": "1.6.0" },
+                { "name": "shared/lib", "version": "2.0.0" },
+                { "name": "shared/lib", "version": "2.0.1" },
+                { "name": "shared/lib", "version": "2.1.0" }
+            ]
+        }
+    ],
+    "require": {
+        "locked/pkg": "^1",
+        "new/pkg":    "^1"
+    }
+}
+--INSTALLED--
+[
+    { "name": "locked/pkg", "version": "1.0.0", "require": { "locked/dep": "^1" } },
+    { "name": "locked/dep", "version": "1.0.0", "require": { "shared/lib": "^2" } },
+    { "name": "shared/lib", "version": "2.0.0" }
+]
+--LOCK--
+{
+    "packages": [
+        { "name": "locked/pkg", "version": "1.0.0", "require": { "locked/dep": "^1" } },
+        { "name": "locked/dep", "version": "1.0.0", "require": { "shared/lib": "^2" } },
+        { "name": "shared/lib", "version": "2.0.0" }
+    ],
+    "packages-dev": [],
+    "aliases": [],
+    "minimum-stability": "stable",
+    "stability-flags": {},
+    "prefer-stable": false,
+    "prefer-lowest": false,
+    "platform": {},
+    "platform-dev": {}
+}
+--RUN--
+update new/pkg --with-dependencies
+--EXPECT-EXIT-CODE--
+2
+--EXPECT-OUTPUT--
+Loading composer repositories with package information
+Updating dependencies
+Your requirements could not be resolved to an installable set of packages.
+
+  Problem 1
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: tests/Composer/Test/Fixtures/installer/github-issues-12862.test [hunk 1]
```diff
@@ -0,0 +1,95 @@
...
+
+  Problem 1
+    - locked/pkg is locked to version 1.0.0 and an update of this package was not requested.
+    - Root composer.json requires new/pkg ^1 -> satisfiable by new/pkg[1.0.0, 1.1.0].
+    - locked/dep 1.0.0 requires shared/lib ^2 -> satisfiable by shared/lib[2.0.0, 2.0.1, 2.1.0].
+    - locked/pkg 1.0.0 requires locked/dep ^1 -> satisfiable by locked/dep[1.0.0].
+    - new/dep 1.0.0 requires shared/lib ^1.0 -> satisfiable by shared/lib[1.0.0, ..., 1.6.0].
+    - new/dep 1.1.0 requires shared/lib ~1.5 -> satisfiable by shared/lib[1.5.0, 1.5.1, 1.6.0].
+    - new/pkg[1.0.0, ..., 1.1.0] require new/dep ^1 -> satisfiable by new/dep[1.0.0, 1.1.0].
+    - Conclusion: don't install shared/lib 2.1.0 (conflict analysis result)
```
naderman: this is missing output formatting to include full non-optimized version lists

naderman:
Think this would still be unusable for real packagist.org packages with hundreds of versions.

stof:
`github-issues-12862.test` shows that the output is not more helpful when the optimizer is enabled. And the optimizer is enabled by default, so this is the output that matters most.

Seldaek:
@naderman I pushed it to expand and it tried to deduplicate the mess which kinda works, but in that huge test the outcome is still confusing..

Lots of:

```
    - new/dep[1.2.5, ..., 1.9.9] require shared/lib ^1.5 -> found shared/lib[1.5.0, ..., 1.5.9] but these were not loaded, likely because it conflicts with another require.
```

But nowhere is it mentioned why shared/lib is pinned/not loaded.

So I guess we close this ? 🤷🏻

PR #12871, opened 2026-05-20
Fix confusing error when an inline alias is not quoted

### LINKED ISSUE #12863: 'require as' needs validation at the command level
When I do:

> composer require drupal-code-builder/drupal-code-builder dev-main as 4.6.x-dev

I get:

```
In RootPackageLoader.php line 179:

  require.as is invalid, it should have a vendor name, a forward slash, and a package name. The vendor and package name can be words separa
  ted by -, . or _. The complete name should match "^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$".
```

This looks like it's an error based on the contents of the updated composer.json file. There should be validation at the command level to say this is wrong, as it was really confusing.

I figured out the problem was it needs to be quoted:

> composer require "drupal-code-builder/drupal-code-builder dev-main as 4.6.x-dev"

eyupcanakman:
Passing an inline alias without quotes, like `composer require pkg dev-main as 4.6.x-dev`, leaves `as` as a separate argument and the command fails with a confusing error instead of explaining the problem. Composer now rejects it and shows the quoted form.

Fixes #12863


Seldaek:
Great, thanks! :)

PR #12881, opened 2026-05-26
Use 'dependency policy' terminology in docs and user-facing output, drop audit --filtered and simplify audit exit code to 0/1

naderman:
Renames the user-facing concept previously called "policy list" / "policy lists" to "dependency policy" / "dependency policies" across the documentation and all user-visible strings (command help, error messages, schema descriptions). Also drops "filter list" / "filter list source" from user-facing output.

## Terminology

- Concept noun (prose, schema descriptions, error messages): **dependency policy / dependency policies**.
- Short form where brevity matters (CLI argument labels, JSON placeholders, repeated within-paragraph mentions, config-key paths like `config.policy.x`): **policy**.
- Data fetched from a source URL: bare **sources** where the surrounding context already names a dependency policy; **dependency policy sources** in standalone messages.

A single entry under `config.policy.<name>` is a dependency policy — block/audit/ignore configuration that applies to a set of package versions. Custom dependency policies supply their package versions via one or more `sources`. The old "policy list" wording conflated the policy with the package data.

Internal class/method names (`ListPolicyConfig`, `suggestListNames`, `assertCustomListName`) are intentionally untouched — this is a docs/UX-only change.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -120,7 +120,7 @@ resolution.
 * **--audit:** Run an audit after installation is complete.
 * **--audit-format:** Audit output format. Must be "table", "plain", "json", or "summary" (default).
 * **--no-security-blocking:** DEPRECATED, use `--no-blocking` instead. Allows installing packages with security advisories or that are abandoned. Also see [COMPOSER_NO_SECURITY_BLOCKING](#composer-no-security-blocking).
-* **--no-blocking:** Disables all policy blocking during this command. Also see [COMPOSER_NO_BLOCKING](#composer-no-blocking).
+* **--no-blocking:** Disables all policy based dependency blocking during this command. Also see [COMPOSER_NO_BLOCKING](#composer-no-blocking).
 * **--optimize-autoloader (-o):** Convert PSR-0/4 autoloading to classmap to get a faster
   autoloader. This is recommended especially for production, but can take
   a bit of time to run so it is currently not done by default.
@@ -209,7 +209,7 @@ php composer.phar update vendor/package:2.0.1 vendor/package2:3.0.*
 * **--no-audit:** Does not run the audit steps after updating the composer.lock file. Also see [COMPOSER_NO_AUDIT](#composer-no-audit).
 * **--audit-format:** Audit output format. Must be "table", "plain", "json", or "summary" (default).
 * **--no-security-blocking:** DEPRECATED, use `--no-blocking` instead. Allows installing packages with security advisories or that are abandoned. Also see [COMPOSER_NO_SECURITY_BLOCKING](#composer-no-security-blocking).
-* **--no-blocking:** Disables all policy blocking during this command. Also see [COMPOSER_NO_BLOCKING](#composer-no-blocking).
+* **--no-blocking:** Disables all policy based dependency blocking during this command. Also see [COMPOSER_NO_BLOCKING](#composer-no-blocking).
 * **--lock:** Overwrites the lock file hash to suppress warning about the lock file being out of
   date without updating package versions. Package metadata like mirrors and URLs are updated if
   they changed.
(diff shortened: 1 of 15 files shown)


doc/06-config.md
```diff
@@ -265,8 +266,8 @@ and scoping (`on-block`/`on-audit`) to limit where the ignore applies.
 
 #### ignore
 
-A list of package names to ignore for security advisories. Supports wildcards and optional version
-constraints. See the [ignore format](#ignore-format) for all supported syntax variants.
+A list of package names to ignoire for security advisory handling. Supports wildcards and optional
+version constraints. See the [ignore format](#ignore-format) for all supported syntax variants.
```
glaubinix:
```suggestion
A list of package names to ignore for security advisory handling. Supports wildcards and optional
version constraints. See the [ignore format](#ignore-format) for all supported syntax variants.
```

src/Composer/Command/PolicyCommand.php
```diff
@@ -36,27 +36,27 @@ protected function configure(): void
     {
         $this
             ->setName('policy')
-            ->setDescription('Manages policy lists and their sources')
+            ->setDescription('Manages custom dependency policies and their sources')
```
glaubinix, line 39: Not user facing but should also update the examples in the doc block

glaubinix (commented):
There are a few more:
* `Rule::getPrettyString` says `was removed by a policy filter list (e.g. malware) and cannot be installed.
* `PoolBuilder::runFilterListFilter` says `Filter list pool filter completed` (probably ok as it refers to the implementation) but then `Found %s package versions referenced in your dependency graph. %s (%d%%) were filtered away by filter lists.`
* `Auditor::outputFilteredPackages` says `$entry->packageName . ' is on filter list "' . $entry->listName . '"'`
* `AuditCommand::configure` the filtered definition says `Overrides the per-list audit setting for malware and every custom filter list.`

doc/03-cli.md
```diff
@@ -1204,10 +1204,10 @@ php composer.phar audit
...
 * **--ignore-severity:** Ignore advisories of a certain severity level. Can be passed one or more
   time to ignore multiple severities.
-* **--filtered:** Behavior on packages matched by `malware` and custom filter
-  lists. Must be "ignore", "report", or "fail". Overrides the per-list `audit`
-  setting (`config.policy.malware.audit` and every custom list's `audit`) for
-  the duration of this command.
+* **--filtered:** Behavior on packages matched by the `malware` dependency policy and custom
+  dependency policies. Must be "ignore", "report", or "fail". Overrides the per-policy `audit`
+  setting (`config.policy.malware.audit` and every custom policy's `audit`) for the duration
+  of this command.
```
glaubinix: This needs to be removed now

naderman: This makes me wonder something else here though. This section talks about "filtered" packages. What are "filtered" packages? Any package versions matching a policy other than abandoned/advisories? or any package versions matching any policy?

glaubinix: filtered packages comes from the first implementation of the feature. Filter lists that filter packages. Think this needs to be reworked then too. Wonder how to best do this? Do we split filtered packages into malware packages and packages matching a custom dependency policy and then every new built-in policy we add gets it's own status code?

naderman: probably have to stop at some point and just return "Other" 😂

doc/03-cli.md
```diff
@@ -1173,23 +1173,17 @@ php composer.phar archive vendor/package 2.0.21 --format=zip
 
 ## audit
 
-This command is used to audit the packages you have installed for potential security issues. It checks for and lists security
-vulnerability advisories using the [Packagist.org api](https://packagist.org/apidoc#list-security-advisories) by default
-or other repositories if specified in the `repositories` section of `composer.json`.
-The command also detects abandoned packages.
+This command is used to audit the packages you have installed against defined dependency policies,
+such as security advisories. It checks for and lists security vulnerability advisories using the
+[Packagist.org api](https://packagist.org/apidoc#list-security-advisories) by default or other
```
```suggestion
[Packagist.org API](https://packagist.org/apidoc#list-security-advisories) by default or other
```

glaubinix (approved):

Seldaek:
Thanks!

PR #12882, opened 2026-05-27
Sanitize urls everywhere in output

Seldaek:
Additional hardening for safety of output /cc @PrinsFrank thanks for getting the ball rolling :)


PrinsFrank (approved):
I intentionally skipped some of these because they should never contain credentials. But good call to sanitize them anyway, better safe than sorry!

PR #12885, opened 2026-05-27
Disable source-fallback by default, mark it deprecated and remove env/flags to override

Seldaek:
This new config option was added in https://github.com/composer/composer/pull/12698 (which is not yet released), and while discussing security related things lately we have come to the conclusion that falling back from dist to source or vice versa can be problematic in some cases and we'd rather disable that.

So the source-fallback option is now disabled by default in 2.10, and set to be removed in 2.11 (always off). 

While motivated by security concerns, it is a behavior change that may affect users tho so we want to keep the option in 2.10 as an easier way to transition/upgrade, and to give everyone a chance to voice concerns in case this is a behavior people really rely on.

### DIFF
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -770,7 +770,7 @@
                 },
                 "source-fallback": {
                     "type": "boolean",
-                    "description": "If true (default), Composer will fall back to a different installation source (e.g., from dist to source or vice versa) when a download fails. Set to false to disable this behavior."
+                    "description": "DEPRECATED, will be removed in Composer 2.11. Defaults to false. If true, Composer will fall back to a different installation source (e.g., from dist to source or vice versa) when a download fails. Automatic source fallback has security implications, please open an issue at https://github.com/composer/composer/issues if you need this kept around."
                 },
                 "github-protocols": {
                     "type": "array",
--- a/res/composer-schema.json
+++ b/res/composer-schema.json
@@ -770,7 +770,8 @@
                 },
                 "source-fallback": {
                     "type": "boolean",
-                    "description": "DEPRECATED, will be removed in Composer 2.11. Defaults to false. If true, Composer will fall back to a different installation source (e.g., from dist to source or vice versa) when a download fails. Automatic source fallback has security implications, please open an issue at https://github.com/composer/composer/issues if you need this kept around."
+                    "description": "DEPRECATED, will be removed in Composer 2.11. Defaults to false. If true, Composer will fall back to a different installation source (e.g., from dist to source or vice versa) when a download fails. Automatic source fallback has security implications, please open an issue at https://github.com/composer/composer/issues if you need this kept around.",
+                    "deprecated": true
                 },
                 "github-protocols": {
                     "type": "array",
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -104,9 +104,6 @@ resolution.
   To get the legacy behavior where Composer use `source` automatically for dev
   versions of packages, use `--prefer-install=auto`. See also [config.preferred-install](06-config.md#preferred-install).
   Passing this flag will override the config value.
-* **--source-fallback / --no-source-fallback:** Override the [config.source-fallback](06-config.md#source-fallback)
-  setting. When disabled, Composer will not fall back to an alternative download source
-  (e.g., from dist to source or vice versa) if the preferred one fails. Also see [COMPOSER_SOURCE_FALLBACK](#composer-source-fallback).
 * **--dry-run:** If you want to run through an installation without actually
   installing a package, you can use `--dry-run`. This will simulate the
   installation and show you what would happen.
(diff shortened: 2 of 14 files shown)


res/composer-schema.json
```diff
@@ -770,7 +770,7 @@
                 },
                 "source-fallback": {
                     "type": "boolean",
-                    "description": "If true (default), Composer will fall back to a different installation source (e.g., from dist to source or vice versa) when a download fails. Set to false to disable this behavior."
+                    "description": "DEPRECATED, will be removed in Composer 2.11. Defaults to false. If true, Composer will fall back to a different installation source (e.g., from dist to source or vice versa) when a download fails. Automatic source fallback has security implications, please open an issue at https://github.com/composer/composer/issues if you need this kept around."
```
glaubinix: Can you add `"deprecated": true`?

glaubinix (approved):
Looks good!

res/composer-schema.json
Seldaek: Ah yeah good point thanks

cs278:
Is the PR that should be mentioned in the blog post?

> Implementation in PRs [#12698](https://github.com/composer/composer/pull/12698) and [second PR still missing]

https://blog.packagist.com/composer-2-10-release/#deprecating-source-fallback

naderman:
Haha thanks, like 5 people read that and nobody noticed.

edalzell:
We often pull from a branch on a staging environment to run tests, how would we do that going forward?

Seldaek:
Like you used to do? If there is a specific issue you'll have to be more specific :) 

Maybe read up on https://github.com/composer/composer/issues/12889 first tho it has a few pointers already but I'm really not sure which solution is for you because your problem description is too vague

edalzell:
Yup saw that discussion after, got it working by removing `--prefer-dist` from the composer command. Thanks for all you do on this!

Seldaek:
Alright, are you also on GitLab or what's the environment? Could you share the exact error message you got?

I am thinking we may be able to detect the error combined with the presence of --prefer-dist and output a clearer hint in this case, but the more info I have to pinpoint it the more detailed output we can give (without risking to mislead people in a completely different scenario).

edalzell:
Nope this is on Laravel Forge (https://forge.laravel.com). They used to have that as part of their composer command defaults. So I had copied that command around on all my sites lol. Easy to remove. A full error message, if it helps:

```
Failed to download transformstudios/prime from dist: The "https://api.github.com/repos/transformstudios/prime/zipball/55e9ad09083e30777053a49f2eae9091b0a2256d" file could not be downloaded (HTTP/2 404 ):
{"message":"Not Found","documentation_url":"https://docs.github.com/rest/repos/contents#download-a-repository-archive-zip","status":"404"}
    Source fallback is disabled. Not trying alternative sources.
```

Note this is a private repo

VincentLanglet:
The removal of this option is pretty annoying for situation like ones where github has some issues with dist dependencies (cf issue https://github.com/composer/composer/issues/12958 and https://github.com/composer/composer/issues/12956 at the moment)

It's useful for dev environment/ci to be able to fallback to source when needed.

Priori to the 2.10 version I feel like the default behavior was to fallback https://github.com/composer/composer/pull/12698/changes#diff-b99c0f08ed0d8f48cb9926a9f015a5cc1fadc6a5c4111c106d788c64264bcb30L200 and this current PR is 
- changing the default behavior
- removing the option

While I think it's ok to change this behavior for security reason, it could still be allowed to override this with an option.

Wirone:
In our case this change broke smoke tests of our internal CLI tool. Yes, relying on this fallback was wrong since the introduction of smoke tests, but it was working fine since then and provided value for us. After Composer update in our custom PHP base image, which led to smoke test failures, I had to spend some time on investigation what actually happened, and I had to improve smoke tests workflow (actually improve build process of one system involved in smoke tests + smoke tests workflow itself). This actually improved the setup and made it predictable, and simply better. _But_ IMHO it is _not_ a security concern important enough to introduce breaking change in minor version, especially considering complexity of enterprise ecosystems (custom images, multi-repo workflows) and possible automations (Dependabot/Renovate) that can lead to broken pipelines without human involvement, in systems deeper in dependency tree. In our case it was Composer update in custom generic PHP image → update of base image version in PHP service boilerplate → failure in smoke test of CLI tool that creates actual service from boilerplate and verifies that it works correctly, so it's actually pretty hard to track such thing down as such change happens over time via multiple PRs in different repos.

PS. I guess it's in some way my fault, as I was the one who approved Renovate's PR that upgraded Composer, merged and released it - if I would have checked release notes properly, we probably could have avoided this issue, but to be fair I am not 100% sure I would have connected the dots there and would have figured out that smoke test 3 repos further is going to stop working... You just don't expect BC-breaks in minor version, so most of the time such Dependabot/Renovate maintenance is just a chore-like formality and you can't spend significant time on each update.

tabacitu:
Opened an issue here for what i think is a legitimate use case that warrants this option: https://github.com/composer/composer/issues/13001

PR #12903, opened 2026-05-29
Fix shell escaping in config --editor

trxvorr:
This fixes config --editor by escaping the file path before passing it to system(), preventing shell metacharacters in the path from being interpreted as commands.

### DIFF
--- a/src/Composer/Command/ConfigCommand.php
+++ b/src/Composer/Command/ConfigCommand.php
@@ -196,11 +196,11 @@ EOT
                     }
                 }
             } else {
-                $editor = escapeshellcmd($editor);
+                $editor = escapeshellarg($editor);
             }
 
             $file = $input->getOption('auth') ? $this->authConfigFile->getPath() : $this->configFile->getPath();
-            system($editor . ' ' . $file . (Platform::isWindows() ? '' : ' > `tty`'));
+            system($editor . ' ' . escapeshellarg($file) . (Platform::isWindows() ? '' : ' > `tty`'));
 
             return 0;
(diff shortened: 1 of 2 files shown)


src/Composer/Command/ConfigCommand.php
jack-worman: fix this while we are here?
```suggestion
                $editor = escapeshellarg($editor);
```

Seldaek: Sorry why is this a fix?

jack-worman: My understand is:
`escapeshellarg()` is for each individual argument
`escapeshellcmd()` is for a whole command

so the current code is incorrect because `$editor` is a single argument

See the following warning from https://www.php.net/manual/en/function.escapeshellcmd.php
> Warning
>
> escapeshellcmd() should be used on the whole command string, and it still allows the attacker to pass arbitrary number of arguments. For escaping a single argument [escapeshellarg()](https://www.php.net/manual/en/function.escapeshellarg.php) should be used instead.

jack-worman: This line is also incorrect: https://github.com/composer/composer/blob/7b3147f083b120846d16c3dc371e61268e0c129e/tests/Composer/Test/AllFunctionalTest.php#L124

Seldaek: Yeah you're right sorry, hadn't looked at this in a while.

Seldaek:
Thanks for the fix!

PR #12904, opened 2026-05-30
Fix audit output to use stdout for summary messages

### LINKED ISSUE #12901: Audit output is partially written to STDERR
Piping `composer audit` output into a pager or to a file is useless as the lines indicating what is ignored vs what is a failure is being written to STDERR.

My `composer.json`: N/A

```
Checking pubkeys: FAIL
Checking Composer version: OK
Composer version: 2.10.0
PHP version: 7.4.0 - Package overridden via config.platform, actual: 7.4.33
OpenSSL version: OpenSSL 3.5.5 27 Jan 2026
curl version: 8.14.1 libz 1.3.1 brotli brotli/1.1.0 zstd missing ssl OpenSSL/3.5.6 HTTP 1.0, 1.1, 2
zip: extension present, unzip present, 7-Zip present (7z)
Checking composer.lock: OK
Checking git settings: OK git version 2.47.3
Checking github.com oauth access: OK expires on 2026-08-31 23:00:00 UTC
Checking disk free space: OK
```

When I run this command: 

```
composer audit --locked > audit.txt
```

I get the following output: 

```
Found 16 ignored security vulnerability advisories affecting 3 packages:
Found 25 security vulnerability advisories affecting 3 packages:
```

(No output) the labels for the tables should be written to STDOUT.

```
```

Dhandapanikeerthana:
Fix audit summary output to use stdout instead of stderr for CLI redirection compatibility.

## Problem
composer audit output was using writeError() for summary messages, causing issues when redirecting output to files or pipes.

## Fix
Changed summary-level output to use write() instead of writeError().

## Impact
- Improves CLI redirection support
- Makes `composer audit > file.txt` work correctly

Fixes #12901

### DIFF
--- a/src/Composer/Advisory/Auditor.php
+++ b/src/Composer/Advisory/Auditor.php
@@ -193,7 +193,7 @@ class Auditor
             $punctuation = $format === self::FORMAT_SUMMARY ? '.' : ':';
             $style = $filteredCount > 0 ? 'error' : 'warning';
 
-            $io->writeError(sprintf('<%s>Found %d package%s matching filters%s</%s>', $style, count($filteredPackages), $plurality, $punctuation, $style));
+            $io->write(sprintf('<%s>Found %d package%s matching filters%s</%s>', $style, count($filteredPackages), $plurality, $punctuation, $style));
             if ($format !== self::FORMAT_SUMMARY) {
                 $this->outputFilteredPackages($io, $filteredPackages, $format);
             }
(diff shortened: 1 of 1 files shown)


Seldaek:
Seems like there are further writeError calls below in the file that should also be modified don't you think?

The two calls about $unreachableRepos I would keep on writeError, but everything else probably belongs on stdout here.

Seldaek:
Thanks

PR #12910, opened 2026-06-02
Update CVE link

liviuconcioiu:
`https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2026-48808` redirects to `https://www.cve.org/CVERecord?id=CVE-2026-48808`


Seldaek:
Good catch, thanks

PR #12912, opened 2026-06-02
Check for local changes when switching a package from source to dist

### LINKED ISSUE #11567: Composer install/update removes `.git` folder in vendor libraries
My `composer.json`:

```json
{
    "name": "internal-project",
    "type": "project",
    "description": "Fancy project",
    "autoload": {
        "psr-4": {
            "App\\": "src/"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "App\\Tests\\": "tests/"
        }
    },
    "require": {
        "php": "^8.2",
        "ext-ctype": "*",
        "ext-curl": "*",
        "ext-dom": "*",
        "ext-fileinfo": "*",
        "ext-iconv": "*",
        "ext-intl": "*",
        "ext-json": "*",
        "ext-simplexml": "*",
        "company/internal-library": "^2",
        "php-http/message-factory": "^1.1",
        "symfony/flex": "^1"
    },
    "require-dev": {
        "mi-schi/phpmd-extension": "^4.3",
        "phpmd/phpmd": "^2.6",
        "phpstan/phpstan": "^1.6",
        "phpunit/phpunit": "^9.5",
        "squizlabs/php_codesniffer": "^3.3",
        "symfony/browser-kit": "^6",
        "symfony/css-selector": "^6",
        "symfony/debug-bundle": "^6",
        "symfony/maker-bundle": "^1.0",
        "symfony/monolog-bundle": "^3.0",
        "symfony/phpunit-bridge": "^6",
        "symfony/stopwatch": "^6",
        "symfony/web-profiler-bundle": "^6"
    },
    "scripts": {
        "auto-scripts": {
            "cache:clear": "symfony-cmd",
            "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd"
        },
        "post-install-cmd": [
            "@auto-scripts"
        ],
        "post-update-cmd": [
            "@auto-scripts"
        ]
    },
    "config": {
        "bin-dir": "bin",
        "secure-http": false,
        "sort-packages": true,
        "allow-plugins": {
            "composer/package-versions-deprecated": true,
            "symfony/flex": true,
            "php-http/discovery": true,
            "symfony/runtime": true
        }
    },
    "extra": {
        "symfony-app-dir": "app",
        "symfony-web-dir": "web",
        "symfony-var-dir": "var",
        "symfony-bin-dir": "bin",
        "symfony-tests-dir": "tests",
        "symfony-assets-install": "relative",
        "incenteev-parameters": {
            "file": "app/config/parameters.yml"
        }
    },
    "repositories": [
        {
            "type": "vcs",
            "url": "https://internal-gitlab/internal-library.git"
        }
    ]
}

```

```
Checking git settings: OK git version 2.39.2
Checking disk free space: OK
Checking pubkeys: FAIL
Checking composer version: OK
Composer version: 2.5.8
PHP version: 8.2.8
PHP binary path: /usr/local/bin/php
OpenSSL version: OpenSSL 3.0.9 30 May 2023
cURL version: 7.88.1 libz 1.2.13 ssl OpenSSL/3.0.9
```

When I run this command: 

```
composer update
```

or 

```
composer install
```

I get the following output: 

```
No special output, but when the package is replaced by another version the .git folder is gone.
```

```
Composer should not remove the .git folder unprompted, because it also deletes all the local commits and branches.
It would be better if composer wouldn't do anything in this case and would say that the update is not possible.
```

eyupcanakman:
A `composer update` that switches a package from a source install to a dist install wiped the whole directory, including its `.git` and any local commits, because the removal skipped the local-changes check that a same-source update runs. The source downloader's uninstall guard now runs first on a type change, so the switch aborts on uncommitted or unpushed work instead of deleting it.

Fixes #11567

### DIFF
--- a/phpstan/baseline.neon
+++ b/phpstan/baseline.neon
@@ -1036,6 +1036,11 @@ parameters:
 			path: ../src/Composer/Downloader/DownloadManager.php
 
 		-
+			message: "#^Cannot call method prepare\\(\\) on Composer\\\\Downloader\\\\DownloaderInterface\\|null\\.$#"
+			count: 1
+			path: ../src/Composer/Downloader/DownloadManager.php
+
+		-
 			message: "#^Cannot call method remove\\(\\) on Composer\\\\Downloader\\\\DownloaderInterface\\|null\\.$#"
 			count: 1
 			path: ../src/Composer/Downloader/DownloadManager.php
--- a/src/Composer/Downloader/DownloadManager.php
+++ b/src/Composer/Downloader/DownloadManager.php
@@ -349,7 +349,17 @@ class DownloadManager
 
         // if downloader type changed, or update failed and user asks for reinstall,
         // we wipe the dir and do a new install instead of updating it
-        $promise = $initialDownloader->remove($initial, $targetDir);
+        $promise = \React\Promise\resolve(null);
+        if ($initialType !== $targetType) {
+            // on a type change the existing source install is about to be wiped, so
+            // run its uninstall guard first to avoid silently dropping local changes
+            // in a modified VCS checkout
+            $promise = $initialDownloader->prepare('uninstall', $initial, $targetDir);
+        }
+
+        $promise = $promise->then(function ($res) use ($initialDownloader, $initial, $targetDir): PromiseInterface {
+            return $initialDownloader->remove($initial, $targetDir);
+        });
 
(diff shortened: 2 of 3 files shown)


Seldaek:
Cool, thanks!

Issue #10353, opened 2021-12-10
Composer gets stuck in a loop when getting 404 from GitLab private repo due to invalid credentials


Jean85:
While checking for #10352, I caught a strange bug that creates a loop during `composer install` with a private GitLab project as a dependency and credentials with insufficient permissions (in my case I was trying to use GitLab CI's `$CI_JOB_TOKEN`).
Data is partially redacted and reduced since it's from a proprietary project.

My `composer.json`:

```json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://gitlab.example.com/foo/bar.git"
        }
    ],
    "require": {
        "foo/bar": "^1.0"
    },
    "config": {
        "preferred-install": {
            "*": "dist"
        },
        "gitlab-domains": [
            "gitlab.example.com"
        ]
    }
}
```

Reproduced using Composer 2.1.8, 2.1.14 (latest stable), 2.2.0-RC1 (latest preview)

When I run this command:
`composer config gitlab-token.gitlab.facile.it gitlab-ci-token $(CI_JOB_TOKEN)`
(as suggested by https://stackoverflow.com/a/50163888) and followed by
`composer install -vvv`

I get the following output:

```
Running 2.2.0-RC1 (2021-12-08 15:17:19) with PHP 7.4.23 on Linux / 5.4.0-74-generic
Reading ./composer.json (/var/www/foo/test/composer.json)
Loading config file ./composer.json (/var/www/foo/test/composer.json)
Loading config file /var/www/foo/test/auth.json
Reading /var/www/foo/test/auth.json
Checked CA file /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a file.
Checked directory /etc/pki/tls/certs/ca-bundle.crt does not exist or it is not a directory.
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/var/www/foo/test): git branch -a --no-color --no-abbrev -v
Executing command (/var/www/foo/test): git rev-list remotes/origin/master..98968a35fd7b371828b2ada3705b50003683846f
Executing command (/var/www/foo/test): git rev-list remotes/origin/develop..98968a35fd7b371828b2ada3705b50003683846f
Executing command (/var/www/foo/test): git describe --exact-match --tags
Failed to initialize global composer: Composer could not find the config file: /home/foo/.composer/composer.json
Reading ./composer.lock (/var/www/foo/test/composer.lock)
Installing dependencies from lock file (including require-dev)
[...]

Issue #12889, opened 2026-05-28
Source fallback removal - unable to use dist (firewall blocks zip files)


laurinkeithdavis:
As indicated in the blog post, I'm opening this issue to express serious concern about the removal of the source option  completely. I understand there are risks, but our corporate firewall blocks zip file downloads, so we have to use source as the primary mechanism - otherwise, we will not be able to use composer at all.

https://github.com/composer/composer/pull/12698
https://github.com/composer/composer/pull/12885

Let me know what additional information would be helpful.

Seldaek:
I think my main question is.. why are you not using [`preferred-install`](https://getcomposer.org/doc/06-config.md#preferred-install) to force source installs if dists are not an option for you?

If the firewall blocks everything, I guess you'd need `composer config --global preferred-install.* source` to configure it machine wide, or without `--global` to just configure it in a single project (but your CI may be able to use zips for example, so maybe you don't want to configure it at the project level but rather per machine).

darthsteven:
I had a private repo definition that looks like this:

```
{
  "type": "vcs",
  "url": "https://github.com/vendor/package"
},
```

So super boring, and with:

```
"config": {
  "preferred-install": {
    "vendor/package": "source"
  }
}
```

Composer was still saying:

> Source fallback is disabled. Not trying alternative sources.

I had to add:

```
"config": {
  "source-fallback": true,
  "preferred-install": {
    "vendor/package": "source"
  }
}
```

To get it working.

This was with composer 2.10.0.

laurinkeithdavis:
@Seldaek I apologize. All my brain saw is "removal of source as an option" and I panicked. Yes, we do use this:

```
	"config"      : {
		"preferred-install": {
			"*": "source"
		}
	},
```

So, we should be good? The comment by @darthsteven is concerning.

cjones-vultr:
Woah boy, this broke our build this morning because we had `composer self-update` in the pipeline.  I know that was a bad idea but here we are.

I want the fallback behavior to stay.  In our repo, we have several packages that pull from Gitlab.  They are defined like:
```
{
        "packages": {
                "repo/package1": "v1.0.0"
        },
        "repositories": [
                {
                        "type": "vcs",
                        "url": "git@gitlab.com:repo/package1",
                        "no-api": true
                },
        ]
}
```

In CI, we want them to fallback to source because the pipeline runs with the permissions of the person running it who might have permissions to clone the repo but not download the zip.

Seldaek:
[...]

PR #12916, opened 2026-06-03
Fix handling of GitLab CI job token auth

### LINKED ISSUE #12984: GitLab CI_JOB_TOKEN + internal package does not work - but works with `source-fallback`
## TL;DR
when using a internal/private GitLab repository as a dependency, `CI_JOB_TOKEN` does not currently work as a mean of authentication inside a GitLab CI pipeline; `source-fallback`makes it work anyway, hiding the issue. 

Having `source-fallback`go away before fixing this would break this process, forcing us to ditch a very secure approach (by either pinning to Composer 2.10, or by ditching `CI_JOB_TOKEN`).

### Why `CI_JOB_TOKEN`?
`CI_JOB_TOKEN` is a very secure mean of authentication inside a GitLab CI pipeline, because:
 - it's ephemeral: the token is valid only until the job ends
 - it doesn't require manual setup or rotation
 - it's not bound to a specific user: it impersonates the user that triggered the job execution
 - it has additional security limitations:
     - is authorized to do only certain operations
     - scope can be further reduced in terms of which repositories and which actions

See docs: https://docs.gitlab.com/ci/jobs/ci_job_token/

### Reproduction
Setup:
 - A GitLab PHP project, which depends on a GitLab internal/private PHP dependency (`my/private-package`)
 - `composer install` run in GitLab CI for the project
 - usage of `CI_JOB_TOKEN` in CI to authenticate against GitLab itself 
```json
{
    "require": {
        "my/private-package": "^1.0"
    },
    "config": {
        "gitlab-domains": [
            "gitlab.example.com"
        ],
    }
}
```

Pipeline execution:
```
composer config http-basic.gitlab.example.com -- gitlab-ci-token "$CI_JOB_TOKEN"
composer install
```

Before 2.10, or with `source-fallback` set to `true`:
```
Failed to download my/private-package from dist: Could not authenticate against gitlab.facile.it
Now trying to download from source
 - Syncing my/private-package (v1.0.0)
```
...and it works.

With `source-fallback` not set in 2.10 or set to `false` :
```
Failed to download my/private-package from dist: Could not authenticate against gitlab.example.com
Source fallback is disabled. Not trying alternative sources.


In AuthHelper.php line 169:

Could not authenticate against gitlab.example.com
```

## What should be done
I've already opened #12916 to fix Composer's behavior when handling `CI_JOB_TOKEN`, because it was sent as a `PRIVATE-TOKEN` instead of a correct `JOB-TOKEN` header. This is unfortunately not enough, because while testing my PR I discovered that the token doesn't have access to the needed API by design. I've opened an issue on GitLab to fix this too: https://gitlab.com/gitlab-org/gitlab/-/work_items/602168

Jean85:
[EDIT] Fixes #12984

For a long time, I had issues with [GitLab CI job token authentication](https://docs.gitlab.com/ci/jobs/ci_job_token/\#job-token-access) when leveraging them to install packages from internal repositories. I even reported [a previous issue](https://github.com/composer/composer/issues/10353) that grazed this one, but for years, when running a `composer install` in CI using job tokens, I always got:

```
Failed to download my-custom/internal-package from dist: Could not authenticate against gitlab.example.com
Now trying to download from source
```

Basically, the CI job token never worked via GitLab APIs, and the build always worked thanks to the fallback to source mode. [Since 2.10 is sunsetting this behavior](https://github.com/composer/composer/pull/12885), it became urgent to fix this long-standing problem.

Looking at the docs: https://docs.gitlab.com/ci/jobs/ci_job_token/\#job-token-access
The implementation seems straightforward, so with this PR I attempted to implement it. The logic in `AuthHelper` is not elegant (with the cascade of `if`s), but at least now it should work.

~I'll attempt to incorporate this branch in my CI build to test this out live.~ See https://github.com/composer/composer/pull/12916#issuecomment-5678181747, tested successfully against GitLab 19.3.2

### DIFF
--- a/src/Composer/Util/AuthHelper.php
+++ b/src/Composer/Util/AuthHelper.php
@@ -303,7 +303,13 @@ class AuthHelper
                     $authenticationDisplayMessage = 'Using GitHub token authentication';
                 }
             } elseif (
-                in_array($auth['password'], ['oauth2', 'private-token', 'gitlab-ci-token'], true)
+                $auth['username'] === 'gitlab-ci-token'
+                && in_array($origin, $this->config->get('gitlab-domains'), true)
+            ) {
+                $headers[] = 'JOB-TOKEN: '.$auth['password'];
+                $authenticationDisplayMessage = 'Using GitLab CI job token authentication';
+            } elseif (
+                in_array($auth['password'], ['oauth2', 'private-token'], true)
                 && in_array($origin, $this->config->get('gitlab-domains'), true)
             ) {
                 if ($auth['password'] === 'oauth2') {
--- a/tests/Composer/Test/Util/AuthHelperTest.php
+++ b/tests/Composer/Test/Util/AuthHelperTest.php
@@ -163,18 +163,7 @@ class AuthHelperTest extends TestCase
         self::assertSame($certificateConfiguration, $options['ssl']);
     }
 
-    public static function gitlabPrivateTokenProvider(): array
-    {
-        return [
-          ['private-token'],
-          ['gitlab-ci-token'],
-        ];
-    }
-
-    /**
-     * @dataProvider gitlabPrivateTokenProvider
-     */
-    public function testAddAuthenticationHeaderWithGitlabPrivateToken(string $password): void
+    public function testAddAuthenticationHeaderWithGitlabPrivateToken(): void
     {
         $headers = [
             'Accept-Encoding: gzip',
@@ -185,7 +174,7 @@ class AuthHelperTest extends TestCase
         $url = 'https://api.gitlab.com/';
         $auth = [
             'username' => 'my_username',
-            'password' => $password,
+            'password' => 'private-token',
         ];
 
         $this->expectsAuthentication($origin, $auth);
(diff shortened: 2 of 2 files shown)


Jean85:
Live test failed :( I'll iterate

Seldaek:
Good luck 😅

Jean85:
Bad news... This cannot work, because `CI_JOB_TOKEN` does not have access to the `repository` API by default, see https://docs.gitlab.com/ci/jobs/ci_job_token/#job-token-access

I verified this with a plain `curl` invocation, and I got the standard `{"message":"404 Project Not Found"}` response that GitLab issues when hitting a secured endpoint.

I then tried to use the granular permissions for the token in the package repository, setting "Read" permissions for the `repository` API, but I still get the 404 response.

This seems nonsensical to me, since the same credentials can be used to clone the repo via Git. I'll try to dig trough the GitLab issue tracker to see if I can find anything that could help us, but outside of that I'm out of options. To overcome this, I would need to go around all the repositories and force the `prefer-source` on the impacted packages, and that doesn't scale well...

Seldaek:
Not sure what your exact problem is, but I posted this earlier which may help 🤷🏻 https://github.com/composer/composer/issues/12889#issuecomment-4620128597

Jean85:
The underlying problem is that I would have to re-do `composer update --prefer-source` for all cases where I have an internal package for each project ad `$dayJob`, which surely doesn't scale very well...

I raised an issue to GitLab hoping to change something about the permissions: https://gitlab.com/gitlab-org/gitlab/-/work_items/602168

expedientesxero-source (approved):

Jean85:
I've just rebased this and, as reported in #12984, I've tested it again with a freshly updated 19.3.2 self-hosted GitLab instance, that now included my contribution. Now the API allows the `GET` with `CI_JOB_TOJEN`, but this PR is needed to correctly authenticate against it. 

Using Composer stable (2.10.3):
```
#29 3.247     Failed to download example/client from dist: Could not authenticate against gitlab.example.com
#29 3.247     Now trying to download from source
#29 3.248   - Syncing example/client (1.15.0) into cache
```

Using this PR:
```
#32 0.917   - Downloading facile/blaine-client (1.15.0)
# ...
#32 3.517 Using GitLab CI job token authentication
# ...
#32 3.783 Writing /home/example/.cache/composer/files/example/client/68457c57ff417f582df39b4ad8c842dfb6231705.zip into cache from /var/www/example/vendor/composer/tmp-5cd50364b1c3f1ef00aa0ff23f62af56.zip
# ...
#32 7.289   - Installing example/client (1.15.0): Extracting archive
```

stof (approved):

Jean85:
@Seldaek it would be great if this is released as a 2.10.x patch, so there's no version that is not able to download internal dependencies (if your GitLab is up to date). Do you think it's feasible?

Seldaek:
Just to make sure.. what happens if you use a new Composer client with an outdated GitLab instance?

stof:
@Seldaek you would get the behavior described in https://github.com/composer/composer/pull/12916#issuecomment-4622547749: the download fails (which also fails if the auth does not work at all, as done by using the existing composer client)

Jean85:
Exactly. 

Not sure if you'd get a 404 (no repo existence exposure) or a 401 (exposure but denied), but the download would still fail as before, and it would possibly trigger the source fallback, still as before.

Seldaek:
Rebased on 2.10 and did some fixes from a Claude review, please check the last commit :)

Jean85:
Commit looks good!

PR #12918, opened 2026-06-03
Verify backup signature on self-update --rollback

Seldaek:
`composer self-update --rollback` restored the newest `*-old.phar` backup from `data-dir` and installed it over `composer.phar` **without verifying its signature** — `rollback()` only ran `validatePhar()` (a structural Phar parse), unlike the normal update path which verifies against the published signature.

On a system where `data-dir`/`COMPOSER_HOME` is writable by other (less-privileged) users, someone could plant a malicious `*-old.phar`; a privileged user running `self-update --rollback` would then install it over the composer binary, leading to code execution as that user. The preconditions make this low severity, but it's cheap to close.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -809,6 +809,16 @@ sudo -H composer self-update
 If Composer was not installed as a PHAR, this command is not available.
 (This is sometimes the case when Composer was installed by an operating system package manager.)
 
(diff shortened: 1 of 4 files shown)


PrinsFrank (approved):
Nice! I like the very thorough fix!

PR #12923, opened 2026-06-05
ComposerRepository: output warning for additional requests

glaubinix:
The warning output is missing for:
* search
* security advisories
* filter list entries
* list


Seldaek:
Thanks

Issue #12922, opened 2026-06-05
Globally (homedir) installed composer fails to update itself


HenkPoley:
_I already have a working proof of concept fix under Windows. This is just to get a ticket in. And I would need to test under macOS/Linux as well._

There is a long-standing issue with non-phar composer completely breaking it's own installation when updating `composer/composer` and related packages. Composer basically uninstalls its own .php files and then stumbles when it still needs those classes. Repairing it afterwards with `php ~/composer.phar require composer/composer` gets a bit long in the tooth after a while.

Example of breakage:

```shell
# Install an old version of the composer command into `~\AppData\Roaming\Composer\vendor\bin\composer.bat`
php ~/composer.phar require composer/composer:2.10.0
# NB make sure that is in your local $PATH
# Use the global install itself to upgrade to a newer version:
composer global require composer/composer:2.10.1 -vvv
```

My `~\AppData\Roaming\Composer\composer.json`:

```json
{
  "require": {
    "laravel/installer": "^5.25",
    "composer": "*",
    "rector/rector": "^2.4",
    "phpstan/phpstan": "^2.1",
    "composer/composer": "^2.10"
  },
  "config": {
    "allow-plugins": {
      "composer/package-versions-deprecated": true
    }
  }
}
```

Output of `composer diagnose` inside my homedir (no `composer.json` in that directory):

```
Composer version: 2.10.1
Checking Composer and its dependencies for vulnerabilities: WARNING
Could not find Composer's installed.json, this must be a non-standard Composer installation.
PHP version: 8.2.31
PHP binary path: C:\tools\php82\php.exe
OpenSSL version: OpenSSL 3.0.20 7 Apr 2026
curl version: 8.20.0 libz 1.3.2 brotli brotli/1.2.0 zstd supported ssl OpenSSL/3.0.20 HTTP 1.0, 1.1, 2
zip: extension present, unzip not available, 7-Zip present (7z)
Checking platform settings: OK
Checking git settings: OK git version 2.54.0
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK does not expire
Checking disk free space: OK
```

When I run this command:

```
composer global update
```

I get the following output: 

```
Changed current directory to C:/Users/henk/AppData/Roaming/Composer
Running 2.10.0 (2026-05-28 11:22:08) with PHP 8.2.31 on Windows NT / 10.0
[...]

PR #12924, opened 2026-06-05
Proof-of-Concept for issue #12922 with composer global update

HenkPoley:
Fixes issue #12922

The fix is roughly: detect we're running from the global composer install location, find all packages required by `composer/composer` itself, check if they are being updated, `require_once` all the PHP files of them, so the temporarily deleted class files are not lazy-loaded during install of the new composer version.

Checked that this works under Windows 11 and Linux.

This is LLM assisted code. Codex with GPT 5.5 on Low / High.

I'm not quite sure if this is a properly minimised change.

### DIFF
--- a/src/Composer/Console/Application.php
+++ b/src/Composer/Console/Application.php
@@ -304,8 +304,10 @@ class Application extends BaseApplication
                     $line = $details['line'];
                 }
 
-                $ghe = new GithubActionError($this->io);
-                $ghe->emit($e->getMessage(), $file, $line);
+                if (class_exists(GithubActionError::class)) {
+                    $ghe = new GithubActionError($this->io);
+                    $ghe->emit($e->getMessage(), $file, $line);
+                }
 
                 throw $e;
             }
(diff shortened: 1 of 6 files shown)


HenkPoley:
One place that may encounter this:

```
php ~/composer.phar global require symfony/finder:8.0.8
# Remove symfony/finder from your ~/.composer/composer.json (macOS location)
composer global update
```

During update of `symfony/finder`, its old class files will be deleted, but during install of the updated classes Composer tries to use `Symfony\Component\Finder\Finder`, which crashes Composer.

HenkPoley:
Previously the '[PHPStan / PHPStan (7.2, false) (pull_request)](https://github.com/composer/composer/actions/runs/27014846509/job/79727133499?pr=12924)' check ran just fine my branch. Which should be the same code.

https://github.com/HenkPoley/composer/actions/runs/27007865324/job/79703898296

Seldaek:
Not sure if this is worth it given the amount of code and hackery this adds. Composer's main deployment method is a phar file, so this seems a bit overkill.

PR #12925, opened 2026-06-06
Avoid writing backspaces to non-decorated output

### LINKED ISSUE #12911: VcsRepository emits backspaces to non-TTY output when clearing status line; --no-progress ineffective
# Composer: backspace (`^H`) characters in redirected output

**Observed with:** Composer 2.10.0

## Summary

When Composer stderr/stdout is redirected to a file (not a TTY), the log often contains long runs of `^H` (ASCII `0x08`, backspace). These are terminal cursor-control bytes that are meant to erase the previous line on an interactive terminal; in a file they appear literally and corrupt commit messages or logs.

`--no-progress` does **not** fix this.

## Steps to reproduce

1. Use a project with at least one **VCS repository** in `composer.json`, for example:

   ```json
    {
        "require": {
            "psr/log": "dev-master"
        },
        "repositories": {
            "psr/log": {
                "type": "github",
                "url": "https://github.com/php-fig/log"
            }
        }
    }
   ```

2. Run update with output redirected:

   ```bash
   composer update --no-ansi --no-progress > out.txt 2>&1
   ```

3. Inspect `out.txt` (or search for backspace):

   ```bash
   cat -v out.txt | grep '\^H'
   ```

## Example output

```
^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H                                                      ^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^HUpdating dependencies
```

## Expected behavior

Plain text suitable for logs and file redirection: no backspace or other TTY cursor-control sequences unless the output stream is an interactive terminal.

## Actual behavior

`^H` bytes are written to the redirected stream.

### 1. `ConsoleIO::doOverwrite()` always emits backspaces

`src/Composer/IO/ConsoleIO.php` (approx. lines 223-248)

`overwrite()` / `overwriteError()` call `doOverwrite()`, which erases the previous message by writing `str_repeat("\x08", $size)` before the new text. There is no check for decorated output or non-TTY streams.

Relevant code:

```php
$size = strlen(strip_tags($stderr ? $this->lastMessageErr : $this->lastMessage));
$this->doWrite(str_repeat("\x08", $size), false, $stderr, $verbosity);
```

This behavior is covered by unit tests (`tests/Composer/Test/IO/ConsoleIOTest.php`) for TTY use, but the same path runs when stderr is redirected.

### 2. `VcsRepository` clears a status line at normal verbosity

`src/Composer/Repository/VcsRepository.php`

After loading tags and branches, when **not** very verbose, Composer calls:

```php
if (!$isVeryVerbose) {
    $this->io->overwriteError('', false);
}
```

(approx. lines 318-320 and 418-420)

That runs even at **normal** verbosity (without `-v`), where the tag/branch loop did not use `overwriteError()` for progress. The `clear` still backspaces over `lastMessageErr`, which is often the preceding status line from `Installer.php`:

```php
$this->io->writeError('<info>Loading composer repositories with package information</info>');
```

So each VCS repository initialized during `Loading composer repositories` can emit a burst of backspaces equal to the length of that status line.

### 3. `--no-progress` does not apply here

`--no-progress` only sets `InstallationManager::setOutputProgress()` (download progress bar during package install). It does not disable:

- `VcsRepository` - `overwriteError('', false)` cleanup
- `ConsoleIO::doOverwrite()` backspaces in general

`RemoteFilesystem` also uses `overwriteError()` when `$progress === true`, but the common curl-based `HttpDownloader` path passes `progress: false` to the `RemoteFilesystem` fallback.

## Suggested fixes (for maintainers)

1. **VcsRepository:** Only call `overwriteError('', false)` when `$isVerbose` (i.e. when overwrite lines were actually shown during tag/branch import).
2. **ConsoleIO:** In `doOverwrite()`, if the target output is not decorated (or not a TTY), skip backspaces and use a normal `writeln` instead.

## References

- https://github.com/composer/composer/blob/main/src/Composer/IO/ConsoleIO.php#L223-L248
- https://github.com/composer/composer/blob/main/src/Composer/Repository/VcsRepository.php#L318-L320
- https://github.com/composer/composer/blob/main/src/Composer/Repository/VcsRepository.php#L418-L420
- https://github.com/composer/composer/blob/main/src/Composer/Installer.php#L509

## Local workaround

Pipe Composer output through `tr -d '\010'` to drop only backspace characters before appending to a file

eyupcanakman:
Composer writes backspaces to clear the previous line even when the output is not a terminal. With output redirected to a file these end up in the log as `^H` bytes. `--no-progress` does not suppress them. `doOverwrite()` now writes a plain line when the output is not decorated.

Fixes #12911

### DIFF
--- a/src/Composer/IO/ConsoleIO.php
+++ b/src/Composer/IO/ConsoleIO.php
@@ -225,6 +225,22 @@ class ConsoleIO extends BaseIO
         // messages can be an array, let's convert it to string anyway
         $messages = implode($newline ? "\n" : '', (array) $messages);
 
+        $decorated = $stderr ? $this->getErrorOutput()->isDecorated() : $this->output->isDecorated();
+
+        // backspaces corrupt non-decorated output, so write a plain line instead
+        if (!$decorated) {
+            if ($messages !== '') {
+                $this->doWrite($messages, true, $stderr, $verbosity);
+            }
+            if ($stderr) {
+                $this->lastMessageErr = $messages;
+            } else {
+                $this->lastMessage = $messages;
+            }
+
+            return;
+        }
+
         // since overwrite is supposed to overwrite last message...
         if (!isset($size)) {
             // removing possible formatting of lastMessage with strip_tags
--- a/tests/Composer/Test/IO/ConsoleIOTest.php
+++ b/tests/Composer/Test/IO/ConsoleIOTest.php
@@ -109,6 +109,7 @@ class ConsoleIOTest extends TestCase
         $outputMock->expects($this->any())
             ->method('getVerbosity')
             ->willReturn(OutputInterface::VERBOSITY_NORMAL);
+        $outputMock->method('isDecorated')->willReturn(true);
         $outputMock->expects($this->atLeast(7))
             ->method('write')
             ->willReturnCallback(static function (...$args) {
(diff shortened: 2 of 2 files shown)


Seldaek:
Thanks!

PR #12933, opened 2026-06-12
Don't hide a suggestion provided by the suggesting package itself

### LINKED ISSUE #12819: Don't consider suggestion fulfilled if provided by the same package
When using a Symfony extension polyfill, e.g. https://github.com/symfony/polyfill-uuid, these packages are setup to list the corresponding `ext-*` both in `provides` (as it polyfills the extension and thus ensures you can use it) as well as in `suggest` (given using the actual extension is likely to be more performant and come with less overhead). However, when running `composer suggest` in any project requiring such a polyfill (either directly or nested), the suggestion will never show up as it is considered available.

If an extension is listed in both `suggest` and `provide` in the same `composer.json` file for a dependency, still show it on running `composer suggest` elsewhere (unless already installed 'natively').

Given that I think extension polyfills are the only case where it makes sense to even have this combination of values, I think there is no downsides to this approach. However, given the very minor pros and potential large technical overhead of such a specific behavior, it is questionable whether or not it is wortwhile pursuing this at all.

n/a

**Additional context**
n/a

eyupcanakman:
A package that lists the same name in both `provide` and `suggest`, like an extension polyfill that provides and suggests `ext-uuid`, never showed its suggestion. The package's own provide marked the name as installed, so `composer suggests` filtered it out. A suggestion is now hidden only when a different installed package provides the target.

Fixes #12819

### DIFF
--- a/src/Composer/Installer/SuggestedPackagesReporter.php
+++ b/src/Composer/Installer/SuggestedPackagesReporter.php
@@ -183,10 +183,10 @@ class SuggestedPackagesReporter
         $installedNames = [];
         if (null !== $installedRepo && !empty($suggestedPackages)) {
             foreach ($installedRepo->getPackages() as $package) {
-                $installedNames = array_merge(
-                    $installedNames,
-                    $package->getNames()
-                );
+                $packageName = $package->getName();
+                foreach ($package->getNames() as $name) {
+                    $installedNames[$name][] = $packageName;
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

src/Composer/Installer/SuggestedPackagesReporter.php
```diff
@@ -200,7 +200,18 @@ private function getFilteredSuggestions(?InstalledRepository $installedRepo = nu
 
         $suggestions = [];
         foreach ($suggestedPackages as $suggestion) {
-            if (in_array($suggestion['target'], $installedNames) || (\count($sourceFilter) > 0 && !in_array($suggestion['source'], $sourceFilter))) {
+            // Only treat a suggestion as fulfilled when a package other than the
+            // one making it provides the target, so a package that both provides
+            // and suggests a name (e.g. an extension polyfill) does not hide its
+            // own suggestion.
+            $providedByOthers = isset($installedNames[$suggestion['target']]) && \count(array_filter(
```
stof, line 207: this `count(array_filter())` should be replaced by `array_find`

Seldaek: Ok but please send such comments as a PR/quick edit, or ideally before things are merged ;)

PR #12937, opened 2026-06-16
Use array_any() in the suggestion provider check

eyupcanakman:
Follow-up to #12933. The provider check only needs to know whether another package provides the target, so this uses `array_any()` instead of `count(array_filter()) > 0`, matching the existing usage in `Problem.php` and `Auditor.php`. `array_any()` over `array_find()` since the matched element is not used. Suggested in review at https://github.com/composer/composer/pull/12933#discussion_r3420366889


Seldaek:
Great, thanks

PR #12947, opened 2026-06-25
Only follow http(s) redirects in HTTP transports

Seldaek:
Restricts HTTP redirect following to the `http` and `https` schemes.

The stream-based `RemoteFilesystem` transport previously followed a `Location:` header to any URL scheme. This aligns it with the curl transport, which already limits protocols to HTTP/HTTPS via `CURLOPT_PROTOCOLS`, and adds the same explicit check on the curl side so the rule lives in one shared place (`Url::isAllowedRedirect()`).

Added coverage for the new `Url::isAllowedRedirect()` helper and for rejection of disallowed redirect targets in `RemoteFilesystem`.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -562,6 +562,10 @@ class CurlDownloader
         }
 
         if (!empty($targetUrl)) {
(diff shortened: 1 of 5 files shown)


iliaal:
LGTM

PR #12948, opened 2026-06-25
Only follow http(s) redirects in HTTP transports (2.2)

Seldaek:
Restricts HTTP redirect following to the `http` and `https` schemes (2.2 LTS backport of #12947).

The stream-based `RemoteFilesystem` transport previously followed a `Location:` header to any URL scheme. This aligns it with the curl transport, which already limits protocols to HTTP/HTTPS via `CURLOPT_PROTOCOLS`, and adds the same explicit check on the curl side so the rule lives in one shared place (`Url::isAllowedRedirect()`).

Added coverage for the new `Url::isAllowedRedirect()` helper and for rejection of disallowed redirect targets in `RemoteFilesystem`.

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -513,6 +513,10 @@ class CurlDownloader
         }
 
         if (!empty($targetUrl)) {
+            if (!Url::isAllowedRedirect($targetUrl)) {
+                throw new TransportException('Could not follow the redirect to "'.Url::sanitize($targetUrl).'" because only http and https redirects are supported.');
+            }
+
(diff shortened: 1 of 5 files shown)


iliaal:
LGTM

src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -513,6 +513,10 @@ private function handleRedirect(array $job, Response $response)
         }
 
         if (!empty($targetUrl)) {
+            if (!Url::isAllowedRedirect($targetUrl)) {
```
staabm, line 516: maybe there should also be some logic to prevent a https -> http downgrade

Seldaek: We already disallow http entirely by default so i think that's ok

Issue #12958, opened 2026-06-29
Automatically retry failed Packagist requests: covering HTTP/2 400


jpmschuler:
Can Composer automatically retry refused connections a few times before exiting? (Similar to #10161 - but adding HTTP/2 400)

```
Package operations: 132 installs, 0 updates, 0 removals
  - Downloading symfony/event-dispatcher (v8.1.1)
  - Downloading symfony/var-exporter (v8.1.1)
  - Downloading symfony/cache (v8.1.1)
    Failed to download symfony/cache from dist: The 
    "https://codeload.github.com/symfony/cache/legacy.zip/c14decc1b0755b1e8ab6babeef56e1880348e817" 
    file could not be downloaded (HTTP/2 400 )
    Source fallback is disabled. Not trying alternative sources.
In CurlDownloader.php line 674:
                                                                               
  The "https://codeload.github.com/symfony/cache/legacy.zip/c14decc1b0755b1e8  
  ab6babeef56e1880348e817" file could not be downloaded (HTTP/2 400 )          
```

GitHub seems to have a lot of hiccups recently where HTTP/2 400 are spawned (reuse of a prior working HTTP/2 connection failed) - as there are factors which could worsen the problem (e.g. a firewall with deep package inspection), still this is worth a retry like the other errors, especially as the necessary code is available.

Else bigger composer package downloads might fail, just because one package wasn't retried (Example from TYPO3 nightly jobs: https://git.typo3.org/typo3/CI/cms/-/jobs/5707898#L89).

pawelabrams:
Well, the issue is also with the code they used to mark those failures. In the case of HTTP 400, the client should not perform the exact same request, should they?

jpmschuler:
I understand your point, but I'd say 400 is similar to 408, 425, 429 in this scenario.

So the assumption is: our code is fine and our call is fine; if it still is Bad Request, then it must be some server side issue, which is worth a retry, else 500 wouldn't make sense to retry either IMHO.

But yes, if we believe in a holy RFC and perfection, you are absolutely right.

bram-pkg:
Potentially related to #12956

sindscream:
We also face with this issue  composer 2.10

rrigby:
Also facing this issue currently. I was unable to deploy. I've had to add my own retry wrapping composer install to get deployments going.

```
&& for i in 1 2 3 4 5; do \
         composer install \

alexbowers:
[...]

PR #12962, opened 2026-06-30
Retry downloads failing with HTTP 400 on a fresh connection

dimasfeb19:
Since last week a lot of CI pipelines started failing with errors like:

```
The "https://codeload.github.com/.../legacy.zip/..." file could not be downloaded (HTTP/2 400 )
```

The failing package is different on every run, the same URL downloads fine with a plain `curl`, and the response is a bare `400: Invalid request` with no rate-limit or `Retry-After` headers. It only happens with parallel downloads, which points at a connection being re-used: the first request on a freshly opened connection succeeds, a later request on the same kept-alive/HTTP-2 connection gets the 400.

This makes two changes to `CurlDownloader`:

- A retried request now opens a fresh connection (`CURLOPT_FRESH_CONNECT`) instead of possibly re-using the pooled connection that just failed. This follows the same idea as the existing "retry forcing IPv4" handling, where a retry should change something to be worth doing.
- `400` is added to the list of retryable status codes.

Both are needed: just adding `400` to the list does not help on its own, because re-using the same poisoned connection returns `400` again on every retry until the attempts run out. Opening a new connection on retry is what actually recovers the download.

If you'd rather not retry `400` by default, since it is normally a client error that should not be repeated, I can instead put it behind an opt-in flag (e.g. `COMPOSER_RETRY_HTTP_400`) and leave the default behaviour unchanged. The `CURLOPT_FRESH_CONNECT` part stands on its own either way. Happy to adjust to whichever you prefer.

Refs #12958

### DIFF
--- a/src/Composer/Util/Http/CurlDownloader.php
+++ b/src/Composer/Util/Http/CurlDownloader.php
@@ -231,6 +231,11 @@ class CurlDownloader
             curl_setopt($curlHandle, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V6);
         }
 
+        if ($attributes['retries'] > 0) {
+            // when retrying, do not re-use a kept-alive connection from the pool as it may be the cause of the failure
+            curl_setopt($curlHandle, CURLOPT_FRESH_CONNECT, true);
+        }
+
         if (function_exists('curl_share_init')) {
             curl_setopt($curlHandle, CURLOPT_SHARE, $this->shareHandle);
         }
@@ -462,7 +467,8 @@ class CurlDownloader
                 if ($statusCode >= 400 && $statusCode <= 599) {
                     if (
                         (!isset($job['options']['http']['method']) || $job['options']['http']['method'] === 'GET')
-                        && in_array($statusCode, [423, 425, 500, 502, 503, 504, 507, 510], true)
+                        // 400 is included as some CDNs (e.g. codeload.github.com) intermittently return it on reused connections, see #12958
+                        && in_array($statusCode, [400, 423, 425, 500, 502, 503, 504, 507, 510], true)
                         && $job['attributes']['retries'] < $this->maxRetries
                     ) {
                         $this->io->writeError('Retrying ('.($job['attributes']['retries'] + 1).') ' . Url::sanitize($job['url']) . ' due to status code '. $statusCode, true, IOInterface::DEBUG);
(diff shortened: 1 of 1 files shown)


src/Composer/Util/Http/CurlDownloader.php
```diff
@@ -421,7 +426,8 @@ public function tick()
                 if ($statusCode >= 400 && $statusCode <= 599) {
                     if (
                         (!isset($job['options']['http']['method']) || $job['options']['http']['method'] === 'GET')
-                        && in_array($statusCode, array(423, 425, 500, 502, 503, 504, 507, 510), true)
+                        // 400 is included as some CDNs (e.g. codeload.github.com) intermittently return it on reused connections, see #12958
+                        && in_array($statusCode, array(400, 423, 425, 500, 502, 503, 504, 507, 510), true)
```
Seldaek: Maybe we can do it for codeload URLs only for now? Because retrying 400s in general sounds very wrong.

Seldaek:
Do you still experience 400s? They seem to have resolved the rate limiting they had in place on the API yesterday. I am not sure if it is related to the 400s on dist downloads but the timing seems to coincide.

src/Composer/Util/Http/CurlDownloader.php
dimasfeb19: Good call. Scoped it so 400 is only retried when the host is codeload.github.com; everything else keeps the previous behaviour, and the fresh-connection-on-retry stays general.

dimasfeb19:
> Do you still experience 400s? They seem to have resolved the rate limiting they had in place on the API yesterday. I am not sure if it is related to the 400s on dist downloads but the timing seems to coincide.

No, not since yesterday afternoon, which lines up with the rate limiting being lifted on GitHub's side. Scoped the 400 retry to codeload.github.com per your note so it stays narrow as cheap hardening if codeload does this again. Happy to drop it entirely if you'd rather not carry it now.

Seldaek:
Ok thanks, I'll happily merge this in 2.10.2 but I don't think this really belong in the 2.2 line, if you don't mind to rebase that'd be easier but otherwise I'll do it tomorrow.

dimasfeb19:
Done, rebased onto `main` (was on the 2.2 line). The only thing I had to touch was resolving against the new `ipResolve` block. Both the FRESH_CONNECT-on-retry and the codeload-scoped 400 retry carried over unchanged, and CI is green.

Seldaek:
Thanks

Issue #12964, opened 2026-06-30
expand docs on limitations of using explicit references


joachim-n:
The docs at https://getcomposer.org/doc/04-schema.md#package-links say that using a specific commit hash has limitations.

It would be useful if this had more specific information.

For example, you can't even do `composer update --lock` reliably when using a commit hash, as it takes the requirements of that dependency from the HEAD of the branch (AFAICT!) even though, naively, you'd think that `composer update --lock` just does 'keep everything the same'.

Some of the text from this comment could be moved into the docs too: https://github.com/composer/composer/issues/6366#issuecomment-295761425

stof:
Here is the existing note present in the doc (emphasis is mine):

> This feature has severe technical limitations, as **the composer.json metadata will still be read from the branch name you specify before the hash**. You should therefore only use this as a temporary solution during development to remediate transient issues, until you can switch to tagged releases. The Composer team does not actively support this feature and will not accept bug reports related to it.

joachim-n:
Indeed, but for users less familiar with the internals of composer, it would help to be more explicit about what is affected.

E.g. I was surprised that 'update --lock' was affected, as to my mind that just 'puts things as they are specified in the lock file'.

glaubinix:
Maybe the confusion comes from what `--lock` does? As per help text `Overwrites the lock file hash to suppress warning about the lock file being out of date without updating package versions. Package metadata like mirrors and URLs are updated if they changed.`

Seldaek:
Feel free to submit a PR with improvements if you have ideas.

Issue #6366, opened 2017-04-20
commit hash dependency does not respect requirements at that hash


simonbrahan:
My `composer.json`:

```json
{
  "require": {
    "openid/php-openid": "dev-master#ee669c6a9d4d95b58ecd9b6945627276807694fb"
  }
}

```

Output of `composer diagnose`:

```
Your version of PHP, 5.3.3, is affected by CVE-2013-6420 and cannot safely perform certificate validation, we strongly suggest you upgrade.
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
The package "openid/php-openid" is pointing to a commit-ref, this is bad practice and can cause unforeseen issues.
require.openid/php-openid : unbound version constraints (dev-master#ee669c6a9d4d95b58ecd9b6945627276807694fb) should be avoided
Checking platform settings: WARNING
Your PHP (5.3.3) is quite old, upgrading to PHP 5.3.4 or higher is recommended.
 Composer works with 5.3.2+ for most people, but there might be edge case issues.
The apc.enable_cli setting is incorrect.
Add the following to the end of your `php.ini`:
  apc.enable_cli = Off
The xdebug extension is loaded, this can slow down Composer a little.
 Disabling it when using Composer is recommended.


The php.ini used by your command-line PHP is: /etc/php.ini
If you can not modify the ini file, you can also run `php -d option=value` to modify ini values on the fly. You can use -d multiple times.

Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: FAIL
Missing pubkey for tags verification
Missing pubkey for dev verification
Run composer self-update --update-keys to set them up
Checking composer version: OK
```

When I run this command:

```
composer install
```

I get the following output:

```
Your version of PHP, 5.3.3, is affected by CVE-2013-6420 and cannot safely perform certificate validation, we strongly suggest you upgrade.
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
[...]

PR #12965, opened 2026-06-30
Clarify the limitations of pinning an explicit commit reference

dimasfeb19:
Follow-up to #12964.

The existing note already says the `composer.json` metadata is read from the branch name, but a couple of people in that thread (and back in #6366) found it hard to tell what that means in practice. This expands the note to be concrete about it:

- the package's own `composer.json` (its requires, autoload, etc.) comes from the branch HEAD, not the commit you pinned, so the resolved dependencies can differ from what's declared at that commit
- the override is only reliable for source installs; a dist install can't honor an arbitrary commit unless the source can build an archive for it (e.g. GitHub)

I left the `composer update --lock` point out, since per @glaubinix's reply that reads more like a question about what `--lock` does than about this feature. Happy to add a line on it if you'd rather cover it here.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -378,12 +378,23 @@ Example:
 }
 ```
 
-> **Note:** This feature has severe technical limitations, as the
-> composer.json metadata will still be read from the branch name you specify
-> before the hash. You should therefore only use this as a temporary solution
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

PR #12968, opened 2026-07-02
Fix forgejo driver to handle empty repos better

Seldaek:
Handles empty repos like https://codeberg.org/upside/option more gracefully (this breaks updates of https://packagist.org/packages/upside/option currently)

Somehow this behavior on branches seems buggy and it returns null instead of empty array.

- https://codeberg.org/api/v1/repos/upside/option/contents/composer.json?ref=main => returns `[]`
- https://codeberg.org/api/v1/repos/upside/option/tags?per_page=100 => returns `[]`
- https://codeberg.org/api/v1/repos/upside/option/branches?per_page=100 => returns `null`

### DIFF
--- a/src/Composer/Repository/Vcs/ForgejoDriver.php
+++ b/src/Composer/Repository/Vcs/ForgejoDriver.php
@@ -58,6 +58,10 @@ class ForgejoDriver extends VcsDriver
         $resource = $this->forgejoUrl->apiUrl.'/contents/' . $file . '?ref='.urlencode($identifier);
         $resource = $this->getContents($resource)->decodeJson();
 
+        if ($resource === []) {
+            return '[]';
+        }
+
         // The Forgejo contents API only returns files up to 1MB as base64 encoded files
         // larger files either need be fetched with a raw accept header or by using the git blob endpoint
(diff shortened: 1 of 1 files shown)


glaubinix (approved):
This indeed looks buggy in forgejo

Seldaek:
Alrighty, reported here too https://codeberg.org/forgejo/forgejo/issues/13295

src/Composer/Repository/Vcs/ForgejoDriver.php
```diff
@@ -58,6 +58,10 @@ public function getFileContent(string $file, string $identifier): ?string
         $resource = $this->forgejoUrl->apiUrl.'/contents/' . $file . '?ref='.urlencode($identifier);
         $resource = $this->getContents($resource)->decodeJson();
 
+        if ($resource === []) {
+            return '[]';
```
stof, line 62: why `[]` for file content ? This looks weird to me. If the file does not exist in the repo, wouldn't `null` be a better return value ?

Seldaek: Both have more or less the same effect, but I am tbh not entirely sure which is the best one here and didn't feel like digging in to figure it out as it is not so relevant as long as the package is not loadable.

PR #12983, opened 2026-07-15
Docs: add missing request/response example for custom policies

glaubinix:
Just noticed that the docs are missing the request/response examples for how to implement custom policies as outlined in https://github.com/composer/composer/issues/12786

### DIFF
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -482,7 +482,10 @@ A `url` source is queried the same way as a repository's [`api-url`](05-reposito
 Composer sends a POST request with the relevant package PURLs and the custom dependency policy name,
 and expects the matching filter entries back. The request is not cached client-side because each
 request body is different. Implementors should be aware that large amounts (a few hundred would be
-normal) of package names can be submitted.
+normal) of package names can be submitted. The submitted PURLs are the full set of candidate package
+names gathered *before* dependency resolution, so they identify packages by name only (no version
+constraints yet), and not every submitted package will necessarily be selected by the resolver
+afterwards.
 
 The endpoint receives a JSON body of the form:
 
--- a/doc/06-config.md
+++ b/doc/06-config.md
@@ -496,7 +496,15 @@ The endpoint receives a JSON body of the form:
 }
 ```
 
-and must return JSON of the form:
+The request body reuses the wire format of a Composer repository's [`api-url`](05-repositories.md#filter).
+There, a single endpoint can serve several named filter lists (for example `malware` and
+`typosquatting`), so `lists` is an array naming which of them Composer wants, and the response is a
+`filter` object keyed by list name. A custom dependency policy has no such multiplexing: its `url`
+source exists only to serve that one policy. Composer therefore always sends the policy name as the
+sole element of `lists` (so the array carries exactly one value here), and the endpoint should treat
+any list name it receives as referring to that policy.
+
+The endpoint must return JSON of the form:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


doc/06-config.md
```diff
@@ -478,6 +478,42 @@ versions, supplied by one or more sources (advertised by package repositories or
...
+            "url": "https://example.org/filters/123",
+            "reason": "Assessed and rejected.",
+            "id": "PKFE-xxxx-xxxx-xxxx"
+        }
+    ]
+}
+```
+
+Unlike a repository's `api-url` response (where `filter` is an object keyed by list name), a custom
+dependency policy `url` source is bound to a single list, so `filter` is a flat array of entries. The
```
naderman: Got to say I still find it confusing we're talking of lists here at all. what does it mean that "a source is bound to a list"?

doc/06-config.md
```diff
@@ -478,6 +478,42 @@ versions, supplied by one or more sources (advertised by package repositories or
...
+and expects the matching filter entries back. The request is not cached client-side because each
+request body is different. Implementors should be aware that large amounts (a few hundred would be
+normal) of package names can be submitted.
+
+The endpoint receives a JSON body of the form:
+
+```json
+{
+    "packages": ["pkg://composer/vendor/package", "pkg://composer/other/package"],
+    "lists": ["my-policy"]
```
naderman, line 495: Please document: Why is this called lists? and why is it an array when it has only one value?

doc/06-config.md
```diff
@@ -478,6 +478,42 @@ versions, supplied by one or more sources (advertised by package repositories or
 Source URLs must use `https://`. `http://` and other schemes are rejected both at
 schema validation time (`composer validate`) and at config load time.
 
+A `url` source is queried the same way as a repository's [`api-url`](05-repositories.md#filter):
+Composer sends a POST request with the relevant package PURLs and the custom dependency policy name,
+and expects the matching filter entries back. The request is not cached client-side because each
+request body is different. Implementors should be aware that large amounts (a few hundred would be
+normal) of package names can be submitted.
```
naderman: This should really also explain what we discussed: List of package names before dependency resolution, not every one of the packages will necessarily be chosen by the dependency resolution afterwards. Not limited to specific versions yet, just full package names.

doc/06-config.md
glaubinix: Clarified

doc/06-config.md
glaubinix: Clarified. The alternative is to not send the lists key at all for custom sources. For now this is still there because eventually, if the same source URL is defined for multiple policies, we might want to bundle them together into a single request instead of triggering one request per policy. However this isn't implemented yet to keep complexity low in Composer and I wasn't sure if this is something anyone would even use.

doc/06-config.md
glaubinix: Added

doc/06-config.md
naderman: I'm not sure that's even a good idea design wise, since you then depend on users getting the policy names right in each composer.json file?

naderman: But also fine with keeping this and the new paragraph below basically tells people to just ignore this now, which is fine.

naderman (approved):

doc/06-config.md
glaubinix: Right, though in the end I would assume that most people will just copy & paste the setup from a docs page. Happy to keep everything as is, as this keeps the implementation in Composer simple.

naderman (approved):

PR #12993, opened 2026-07-21
Policies: show which source blocked a package version as part of the filter list pool filter

glaubinix:
We never really display from which source a filter entry is from during the `composer install|update` output, making it hard to use the `ignore-source` config. See integration test adjustments for adjusted output.


Seldaek:
Thanks

Issue #13002, opened 2026-07-24
Improve error message to log failed URL when package metadata fails to download.


sensespidey:
**Is your feature request related to a problem? Please describe.**

In the context of working on the [PHP TUF Composer integration](https://github.com/php-tuf/composer-integration) plugin for Drupal's TUF implementation, we sometimes run into problems where a target package is malformed such that the TUF validation (specifically the filesize check) fails, and outputs an error message like this:

```
https://packages.drupal.org/8 could not be fully loaded (Maximum allowed download size reached. Downloaded 9894 of allowed 9894 bytes), package information was loaded from the local cache and
 may be out of date
```

This message does not let us identify which metadata package failed to download correctly in order to troubleshoot.

**Describe the solution you'd like**

In the case of a failure like this, we'd like to see the URL of the failed download included in the error message that is output.

For example, it appears it would be straightforward to add the `$filename` variable to the error output in [src/Composer/Repository/ComposerRepository::fetchFile()](https://github.com/composer/composer/blob/main/src/Composer/Repository/ComposerRepository.php#L1733) (and the 2 similar methods `asyncFetchFile` and `fetchFileIfLastModified` which have the same `$this->io->writeError` call with the "could not be fully loaded" error message.

**Describe alternatives you've considered**

We've proposed a patch to the php-tuf/composer-integration library which somewhat mitigates the problem by logging the URL to be downloaded in its `pre-file-download` hook, but this is only a partial measure, since we can't detect whether there has been an error and simply have to print out ALL the packages we attempt to download.

**Additional context**

This issue occurs somewhat intermittently, but we are currently seeing this error when we tuf-enable a Drupal project and attempt to `composer require drupal/lti_tool_provider`.

* https://github.com/php-tuf/composer-integration/issues/128 (a proximate report of this issue)
* https://git.drupalcode.org/project/infrastructure/-/work_items/3579174 (details of how to tuf-enable a Drupal composer project and produce this type of output)

Seldaek:
Can you send a PR? It might be easier to discuss the change :)

The detailed URLs are available when running with `-vvv` though just FYI.

sensespidey:
> Can you send a PR? It might be easier to discuss the change :)

[...]

PR #13006, opened 2026-07-29
feat: #13002 Improve logging of failed download URLs by adding filename to message

sensespidey:
This patch simply adds the filename to log messages when a package component fails to download, in order to aid troubleshooting of such failures. See #13002 for more context :)

### DIFF
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -1775,7 +1775,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
                     }
 
                     // TODO use scarier wording once we know for sure it doesn't do false positives anymore
-                    throw new RepositorySecurityException('The contents of '.$filename.' do not match its signature. This could indicate a man-in-the-middle attack or e.g. antivirus software corrupting files. Try running composer again and report this if you think it is a mistake.');
+                    throw new RepositorySecurityException('The contents of '.Url::sanitize($filename).' do not match its signature. This could indicate a man-in-the-middle attack or e.g. antivirus software corrupting files. Try running composer again and report this if you think it is a mistake.');
                 }
 
                 if ($this->eventDispatcher) {
--- a/src/Composer/Repository/ComposerRepository.php
+++ b/src/Composer/Repository/ComposerRepository.php
@@ -1815,7 +1815,7 @@ class ComposerRepository extends ArrayRepository implements ConfigurableReposito
 
                 if ($cacheKey && ($contents = $this->cache->read($cacheKey))) {
                     if (!$this->degradedMode) {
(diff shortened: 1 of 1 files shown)


src/Composer/Repository/ComposerRepository.php
```diff
@@ -1815,7 +1815,7 @@ protected function fetchFile(string $filename, ?string $cacheKey = null, ?string
 
                 if ($cacheKey && ($contents = $this->cache->read($cacheKey))) {
                     if (!$this->degradedMode) {
-                        $this->io->writeError('<warning>'.Url::sanitize($this->url).' could not be fully loaded ('.$e->getMessage().'), package information was loaded from the local cache and may be out of date</warning>');
+                        $this->io->writeError('<warning>'.Url::sanitize($this->url).' could not be fully loaded ('.$e->getMessage().'), package information was loaded from the local cache and may be out of date ('. $filename .')</warning>');
```
Seldaek: I think it'd be good to have the URL closer to the error, and we definitely need to sanitize credentials here, something like:
```suggestion
                        $this->io->writeError('<warning>'.Url::sanitize($this->url).' could not be fully loaded ('.Url::sanitize($filename).': '.$e->getMessage().'), package information was loaded from the local cache and may be out of date</warning>');
```

Would need to be applied below too.

sensespidey: Thanks for the feedback! I'm happy to add in the sanitize call there if you think it makes sense. I left it out in this case as it appeared that `$filename` was already (effectively) sanitized earlier in the function, and is elsewhere used "naked" like this. Nonetheless, better to be safe :)

Seldaek: If you mean the `throw new RepositorySecurityException` above, then feel free to add Url::sanitize there too. Not sure where you've seen it being already sanitized earlier in the function, but I don't think so. If the repo URL contains inline credentials (terrible idea but shit happens..) then it would be incluedd in $filename as far as I can tell.

sensespidey: On a closer reading, I believe you are correct! I had read the function through logically without enough semantics (or tracing called functions carefully) and thought I'd seen it assigned to a "known good" value, rather than sanitized explicitly.

I've pushed another small commit to add `Url::sanitize()` to the RepositorySecurityException instance you note earlier in the method, as you note it probably also makes sense there :)

Seldaek:
Alright let's get this in, so you can try it out now with `composer self-update --snapshot` and hopefully confirm it behaves well for the issue you had.

Seldaek:
On the other hand, this is now very verbose and repetitive:

> https://repo.packagist.org could not be fully loaded (https://repo.packagist.org/p2/phpstan/phpstan.json: The "https://repo.packagist.org/p2/phpstan/phpstan.json" file could not be downloaded (HTTP/2 502 )), package information was loaded from the local cache and may be out of date

So maybe the other patch proposal that was adding the filename in the download size exception would be more approprite :)

Seldaek:
Opened https://github.com/composer/composer/pull/13041 to do an alternative version of the fix that's less verbose then

PR #13014, opened 2026-08-01
Fix suggest filtering for non-installed packages (#12992)

### LINKED ISSUE #12992: Err when providing non-installed package name to `suggest`
The `composer suggest` command allows providing a package name to use as filter, e.g. `composer suggest composer/composer` will only show suggestions by the composer package. However, when one provides an expression of a non-installed package, the output of the command will be empty, which is not distinguishable from a response that the package does not have (fulfilled) suggests.

For example:
- `composer suggest composer/composer` will often result in no output, given you are running
  - if ran on a repository not having `composer/composer` in its dependency tree
-  `composer suggest composer/composr` will always result in no output, not notifying the user of their typo
- `composer suggest composer/*` will always result in no output, not notifying the user of the fact that this syntax is not supported for this command
- `composer suggest all` will always result in no output, not notifying the user of the fact they probably intended to use `composer suggest --all`

Whenever the package provided as filter is not found in the installed packages, err and exit with code non-zero:
```
> composer suggest composer/composr
composer/composr is not installed
```

Alternatively, the output could be altered to "always" display results of some kind, e.g. 
```
> composer suggest composer/composer
no suggestions found for composer/composer
```

This would improve for some of the cases above, but probably less so than an explicit error.

**Additional context**
-

ousamabenyounes:
Fixes #12992.

## Summary

- fail cleanly when package filters cannot be used instead of throwing an uncaught `InvalidArgumentException`
- write each validation error to Composer's error stream and return `Command::FAILURE`
- preserve input casing in errors while matching installed package names case-insensitively
- report every missing filter and distinguish locked dev packages excluded by `--no-dev`

## Test verification (RED -> GREEN)

RED on the old throw path with the updated invalid-filter tests:

```text
5 InvalidArgumentException errors in SuggestsCommandTest invalid-filter cases
```

GREEN on the patched branch:

```text
SuggestsCommandTest: OK (37 tests, 79 assertions)
Full PHPUnit: OK, but incomplete, skipped, or risky tests! Tests: 3336, Assertions: 7443, Skipped: 8
PHPStan: [OK] No errors
PHP diff coverage: PASS 27/27 changed production lines covered
```

The remaining deprecation notice is pre-existing in this checkout: `AuthHelper::addAuthenticationHeader is deprecated since Composer 2.9 use addAuthenticationOptions instead.`

### DIFF
--- a/src/Composer/Command/SuggestsCommand.php
+++ b/src/Composer/Command/SuggestsCommand.php
@@ -69,9 +69,19 @@ EOT
         $installedRepo = new InstalledRepository($installedRepos);
         $reporter = new SuggestedPackagesReporter($this->getIO());
 
-        $filter = $input->getArgument('packages');
+        $filter = array_map('strtolower', $input->getArgument('packages'));
         $packages = $installedRepo->getPackages();
         $packages[] = $composer->getPackage();
+        foreach ($filter as $package) {
+            if ($installedRepo->findPackages($package) !== []) {
+                continue;
+            }
+
+            $this->getIO()->writeError('<error>Package "'.$package.'" is not installed</error>');
+
+            return self::FAILURE;
+        }
+
         foreach ($packages as $package) {
             if (!empty($filter) && !in_array($package->getName(), $filter)) {
                 continue;
--- a/src/Composer/Command/SuggestsCommand.php
+++ b/src/Composer/Command/SuggestsCommand.php
@@ -67,23 +67,47 @@ EOT
         }
 
         $installedRepo = new InstalledRepository($installedRepos);
-        $reporter = new SuggestedPackagesReporter($this->getIO());
+        $filter = $input->getArgument('packages');
+        $normalizedFilter = array_map('strtolower', $filter);
+        $validationRepo = $installedRepo;
+        if ($locker->isLocked() && $input->getOption('no-dev')) {
+            $validationRepo = new InstalledRepository([$locker->getLockedRepository(true)]);
+        }
 
-        $filter = array_map('strtolower', $input->getArgument('packages'));
-        $packages = $installedRepo->getPackages();
-        $packages[] = $composer->getPackage();
-        foreach ($filter as $package) {
+        $missingPackages = [];
+        $excludedPackages = [];
+        foreach ($normalizedFilter as $index => $package) {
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
Addressed the requested filter-validation refinements in `7db53100e`: display casing is preserved while matching case-insensitively, all missing filters are reported together, and locked dev packages filtered by `--no-dev` receive an accurate hint.

Validation: 5 focused cases RED on the previous PR HEAD, then 37 tests / 74 assertions GREEN. `run-ci.sh` final is iso-baseline, with 26/26 changed executable lines covered.

src/Composer/Command/SuggestsCommand.php
```diff
@@ -67,13 +67,47 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                ? 'Package "'.$missingPackages[0].'" is not installed.'
+                : 'Packages "'.implode('", "', $missingPackages).'" are not installed.';
+        }
+        if ($excludedPackages !== []) {
+            $errors[] = count($excludedPackages) === 1
+                ? 'Package "'.$excludedPackages[0].'" is excluded by --no-dev.'
+                : 'Packages "'.implode('", "', $excludedPackages).'" are excluded by --no-dev.';
+        }
+        if ($errors !== []) {
+            throw new \InvalidArgumentException(implode(' ', $errors));
```
Seldaek: I think it'd be better to do this for readability:
```suggestion
            foreach ($errors as $error) {
                $output->writeln('<error>'.$error.'</error>');
            }
            return 1;
```

ousamabenyounes:
Thanks for the review. Addressed the validation branch in `c396bdb43bb1871a9404b34d9292edff5810a5af`: errors are now emitted one per line and the command returns failure, while keeping diagnostics on stderr through Composer IO.

Validation:
- RED: old throw path failed the updated invalid-filter tests with 5 `InvalidArgumentException` errors.
- GREEN: `SuggestsCommandTest` passed in Docker (37 tests, 79 assertions).
- Full local replay passed in Docker: 3336 tests, 7443 assertions, 8 existing skips; PHPStan no errors; PHP diff coverage 27/27 changed production lines.

PR #13025, opened 2026-08-10
Fix overwriting a list repository addressed by numeric index in config command

### LINKED ISSUE #12623: composer config repositories.X now works differently if X is numeric
After updating to Composer 2.9 I noticed a discrepancy between its behaviour before and after, see details below.

(Here I do `composer create-project`)

Now I need to replace stuff in repositories section of my composer.json that looks like this:

```json
{
    "name": "magento/project-enterprise-edition",
    "description": "eCommerce Platform for Growth (Enterprise Edition)",
    "type": "project",
    "license": [
        "proprietary"
    ],
    "config": {
        "preferred-install": "dist",
        "sort-packages": true
    },
    "version": "2.3.7",
    "require": {
        "magento/product-enterprise-edition": "2.3.7",
        "magento/composer-root-update-plugin": "~1.1"
    },
    "autoload": {
        "exclude-from-classmap": [
            "**/dev/**",
            "**/update/**",
            "**/Test/**"
        ],
        "files": [
            "app/etc/NonComposerComponentRegistration.php"
        ],
        "psr-0": {
            "": [
                "app/code/",
                "generated/code/"
            ]
        },
        "psr-4": {
            "Magento\\": "app/code/Magento/",
            "Magento\\Framework\\": "lib/internal/Magento/Framework/",
            "Magento\\Setup\\": "setup/src/Magento/Setup/",
            "Zend\\Mvc\\Controller\\": "setup/src/Zend/Mvc/Controller/"
        }
    },
    "require-dev": {
        "allure-framework/allure-phpunit": "~1.2.0",
        "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
        "friendsofphp/php-cs-fixer": "~2.16.0",
        "lusitanian/oauth": "~0.8.10",
        "magento/magento-coding-standard": "*",
        "magento/magento2-functional-testing-framework": "^2.7",
        "pdepend/pdepend": "~2.7.1",
        "phpcompatibility/php-compatibility": "^9.3",
        "phpmd/phpmd": "^2.8.0",
        "phpstan/phpstan": ">=0.12.3 <=0.12.23",
        "phpunit/phpunit": "^9",
        "sebastian/phpcpd": "~5.0.0",
        "squizlabs/php_codesniffer": "~3.5.4"
    },
    "conflict": {
        "gene/bluefoot": "*"
    },
    "autoload-dev": {
        "psr-4": {
            "Magento\\PhpStan\\": "dev/tests/static/framework/Magento/PhpStan/",
            "Magento\\Sniffs\\": "dev/tests/static/framework/Magento/Sniffs/",
            "Magento\\TestFramework\\Inspection\\": "dev/tests/static/framework/Magento/TestFramework/Inspection/",
            "Magento\\TestFramework\\Utility\\": "dev/tests/static/framework/Magento/TestFramework/Utility/",
            "Magento\\Tools\\": "dev/tools/Magento/Tools/",
            "Magento\\Tools\\Sanity\\": "dev/build/publication/sanity/Magento/Tools/Sanity/"
        }
    },
    "prefer-stable": true,
    "repositories": [
        {
            "type": "composer",
            "url": "https://repo.magento.com/"
        }
    ],
    "extra": {
        "magento-force": "override"
    }
}
```

In order to replace repo.magento.com with my private packagist repository I have been doing the following set of commands:

```shell
composer config repositories.0 path "../path/to/my/local/repository"
composer config repositories.1 "https://repo.packagist.com/vendor"
composer config repositories.packagist.org false
```

All of this to get repositories section to look like this:

```json
{
    "repositories": {
        "packagist.org": false,
        "1": {
            "type": "composer",
            "url": "https://repo.packagist.com/vendor"
        },
        "0": {
            "type": "path",
            "url": "../"
        }
    }
}
```

After the first command I expect the following output:

```json
{
    "repositories": [
        {
            "type": "path",
            "url": "../"
        }
    ]
}
```

But with the new Composer 2.9 I get the following instead:

```json
{
    "repositories": [
        {
            "type": "path",
            "url": "../"
        },
        {
            "type": "composer",
            "url": "https://repo.magento.com/"
        }
    ]
}
```

As you can see, instead of replacing the first section of repositories, it now prepends it with the new value, which breaks my build scripts.

The workaround for the behaviour (and something I should have done in my scripts in the first place) is below, but I still think it is worth to report this as a regression in new release.

```shell
composer config --unset repositories
composer config repositories.packagist.org false
composer config repositories.local path "../"
composer config repositories.vendor "https://repo.packagist.com/vendor"
```

ousamabenyounes:
Fixes #12623

### Problem

`composer config repositories.N <type> <url>`, where `N` is a numeric index into a `repositories` **list**, must overwrite the existing repository at position `N`. Since the `repo` command refactor it no longer does: the old entry is kept and a duplicate is added instead.

```console
$ cat composer.json
{
    "repositories": [
        { "type": "path", "url": "../" },
        { "type": "composer", "url": "https://other.test" }
    ]
}

$ composer config repositories.0 composer https://replaced.test
# before this patch, index 0 is NOT replaced — a duplicate is prepended:
{
    "repositories": [
        { "type": "composer", "url": "https://replaced.test" },
        { "type": "path", "url": "../" },
        { "type": "composer", "url": "https://other.test" }
    ]
}
```

### Root cause

`JsonManipulator::doRemoveRepository()` (called by `addRepository()`/`insertRepository()`/`removeRepository()` before writing the new entry) matched repositories only by their `name` property. A numeric key like `0` never matches a `name`, so nothing is removed and the new entry is simply added. Reading by numeric index already works (`config repositories.0`); only writing regressed.

### Fix

Match a numeric name against the list **position** so the entry at that index is removed before the new one is written back. The branch is guarded with `!$isAssoc && is_numeric($name)`, so named repositories and object-format `repositories` are unaffected.

### Test verification (RED → GREEN)

Two tests were added (unit + functional). Both fail before the production change and pass after.

**RED** — with only the tests applied on top of the current base (production fix reverted):

```
1) Composer\Test\Json\JsonManipulatorTest::testAddRepositoryByNumericIndexOverwritesListEntry
Failed asserting that two strings are equal.
--- Expected  (index 0 replaced)
+++ Actual    (old entry kept, duplicate added)

FAILURES! Tests: 1, Assertions: 2, Failures: 1.

2) Composer\Test\Command\ConfigCommandTest::testConfigUpdates with data set
   "overwrite a repository addressed by numeric index in a list"
Failed asserting that two arrays are identical.
FAILURES! Tests: 75, Assertions: 150, Failures: 1.
```

**GREEN** — with the fix:

```
Composer\Test\Json\JsonManipulatorTest
OK (1 test, 2 assertions)

Composer\Test\Command\ConfigCommandTest
OK (95 tests, 194 assertions)
```

Full related suites pass (`JsonManipulatorTest` 125/125, `ConfigCommandTest` 95/95, `RepositoryCommandTest` 20/20, `JsonConfigSourceTest` 45/45) and PHPStan level 8 is clean.

### DIFF
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -901,16 +901,17 @@ If your repository requires more configuration options, you can instead pass its
 php composer.phar config repositories.foo '{"type": "vcs", "url": "http://svn.example.org/my-project/", "trunk-path": "master"}'
 ```
 
-Repositories which have no `name` can be addressed by their position in the list instead, which is
-what `composer repo list` shows in brackets. The repository at that position is then replaced in
-place rather than a new one being added:
+A numeric key addresses a repository by its position in the `repositories` array of the file being
+modified, replacing the entry at that position rather than adding a new one:
 
 ```shell
 php composer.phar config repositories.0 vcs https://github.com/foo/bar
 ```
 
-Positions shift as repositories are added and removed, so giving each repository a `name` and using
-that is more robust.
+This is discouraged and warns, because positions shift as repositories are added or removed, and
+they do not match the numbers `composer repo list` displays once the global config defines
+repositories of its own. Addressing a repository which has a `name` by its position is an error —
+use the name. Give each repository a `name` and you never have to think about positions.
 
 ### Modifying Extra Values
 
@@ -948,8 +949,9 @@ repo [options] enable packagist.org
 repo [options] disable packagist.org
 ```
 
-`[repo-name]` is the repository's `name`. Repositories which do not have one can be addressed by
-their position in the list instead, as shown in brackets by `repo list`.
+`[repo-name]` is the repository's `name`. A repository without one can be addressed by its position
+in the `repositories` array of the file being modified, but this is discouraged and warns — see
+[Modifying Repositories](#modifying-repositories).
 
 ### Options
 
@@ -957,7 +959,7 @@ their position in the list instead, as shown in brackets by `repo list`.
 - **--file (-f):** to modify a specific file instead of composer.json.
 - **--append:** to add a repository with lower priority (by default repositories are prepended and have thus higher priority than existing ones).
 - **--before [name]:** to insert the new repository before an existing repository named `[name]`.
-- **--after [name]:** to insert the new repository after an existing repository named `[name]`. The `[name]` must match an existing repository name or position.
+- **--after [name]:** to insert the new repository after an existing repository named `[name]`. The `[name]` must match an existing repository name, or its position.
 
 ### Examples
 
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -913,6 +913,10 @@ they do not match the numbers `composer repo list` displays once the global conf
 repositories of its own. Addressing a repository which has a `name` by its position is an error —
 use the name. Give each repository a `name` and you never have to think about positions.
 
+For the same reason a repository `name` must not be numeric: a number always addresses a position,
+so a repository called `1` could not be told apart from the second entry in the list. `composer
+validate` warns about any that are already there.
+
 ### Modifying Extra Values
 
 In addition to modifying the config section, the `config` command also supports making
--- a/doc/03-cli.md
+++ b/doc/03-cli.md
@@ -901,6 +901,17 @@ If your repository requires more configuration options, you can instead pass its
 php composer.phar config repositories.foo '{"type": "vcs", "url": "http://svn.example.org/my-project/", "trunk-path": "master"}'
 ```
 
+Repositories which have no `name` can be addressed by their position in the list instead, which is
+what `composer repo list` shows in brackets. The repository at that position is then replaced in
+place rather than a new one being added:
+
+```shell
+php composer.phar config repositories.0 vcs https://github.com/foo/bar
+```
+
+Positions shift as repositories are added and removed, so giving each repository a `name` and using
+that is more robust.
+
 ### Modifying Extra Values
 
 In addition to modifying the config section, the `config` command also supports making
@@ -937,13 +948,16 @@ repo [options] enable packagist.org
 repo [options] disable packagist.org
 ```
 
+`[repo-name]` is the repository's `name`. Repositories which do not have one can be addressed by
+their position in the list instead, as shown in brackets by `repo list`.
+
 ### Options
 
 - **--global (-g):** to modify the global `$COMPOSER_HOME/config.json`.
 - **--file (-f):** to modify a specific file instead of composer.json.
 - **--append:** to add a repository with lower priority (by default repositories are prepended and have thus higher priority than existing ones).
 - **--before [name]:** to insert the new repository before an existing repository named `[name]`.
-- **--after [name]:** to insert the new repository after an existing repository named `[name]`. The `[name]` must match an existing repository name.
+- **--after [name]:** to insert the new repository after an existing repository named `[name]`. The `[name]` must match an existing repository name or position.
 
 ### Examples
 
--- a/src/Composer/Command/BaseConfigCommand.php
+++ b/src/Composer/Command/BaseConfigCommand.php
@@ -81,6 +81,45 @@ abstract class BaseConfigCommand extends BaseCommand
     }
 
     /**
+     * Repositories are meant to be addressed by name. A numeric key falls back to the position of
+     * the entry in the repositories list, which is unreliable: positions shift as repositories are
+     * added or removed, and they do not line up with what `composer repo list` displays once the
+     * global config defines repositories of its own. So warn about it, and refuse it outright when
+     * it would silently act on a repository which does have a name.
+     */
+    protected function validateRepositoryKey(string $name): void
+    {
+        if (!ctype_digit($name) || !$this->configFile->exists()) {
+            return;
+        }
+
+        $repositories = $this->configFile->read()['repositories'] ?? [];
+
+        // in the object format a numeric key is a name rather than a position
+        if (!is_array($repositories) || !array_is_list($repositories)) {
+            return;
+        }
+
+        foreach ($repositories as $repository) {
+            if (is_array($repository) && ($repository['name'] ?? null) === $name) {
+                return;
+            }
+        }
+
+        $target = $repositories[(int) $name] ?? null;
+
+        if (is_array($target) && isset($target['name']) && is_string($target['name'])) {
+            throw new \RuntimeException(sprintf(
+                'The repository at position %s is named "%s", address it by that name instead.',
+                $name,
+                $target['name']
+            ));
+        }
+
+        $this->getIO()->writeError('<warning>Addressing a repository by its position ("'.$name.'") is unreliable as positions shift when repositories are added or removed, and do not match what "composer repo list" shows when the global config defines repositories. Give the repository a "name" and use that instead.</warning>');
+    }
+
+    /**
      * Get the local composer.json, global config.json, or the file passed by the user
      */
     protected function getComposerConfigFile(InputInterface $input, Config $config): string
--- a/src/Composer/Command/BaseConfigCommand.php
+++ b/src/Composer/Command/BaseConfigCommand.php
@@ -15,6 +15,7 @@ namespace Composer\Command;
 use Composer\Config;
 use Composer\Config\JsonConfigSource;
 use Composer\Json\JsonFile;
+use Composer\Json\JsonManipulator;
 use Composer\Factory;
 use Composer\Util\Platform;
 use Composer\Util\Silencer;
@@ -100,10 +101,9 @@ abstract class BaseConfigCommand extends BaseCommand
             return;
         }
 
-        foreach ($repositories as $repository) {
-            if (is_array($repository) && ($repository['name'] ?? null) === $name) {
-                return;
-            }
+        // a repository actually carrying this name wins over the position
+        if (null !== JsonManipulator::findRepositoryKey($repositories, $name, false)) {
+            return;
         }
 
         $target = $repositories[(int) $name] ?? null;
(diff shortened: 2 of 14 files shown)


src/Composer/Json/JsonManipulator.php
```diff
@@ -281,31 +280,14 @@ public function insertRepository(string $name, $config, string $referenceName, i
             return false;
         }
 
-        $indexToInsert = null;
```
glaubinix, line 284: There is a behaviour change here which looks wrong

Have the composer.json
```json
{
    "name": "acme/package",
    "repositories": [
        {
            "name": "a",
            "type": "path",
            "url": "../a"
        },
        {
            "name": "b",
            "type": "path",
            "url": "../b"
        },
        {
            "name": "c",
            "type": "composer",
            "url": "https://c.test"
        }
    ]
}
```

Run command
```bash
bin/composer repo add 0 vcs https://moved.test --before 1
```

On main this fails with
```
In JsonConfigSource.php line 172:

  The referenced repository "1" does not exist.
```

In this branch this rewrites the composer.json to
```json
{
    "name": "acme/package",
    "repositories": [
        {
            "name": "b",
            "type": "path",
            "url": "../b"
        },
        {
            "type": "vcs",
            "url": "https://moved.test"
        },
        {
            "name": "c",
            "type": "composer",
            "url": "https://c.test"
        }
    ]
}
```

Which I am not sure is correct

glaubinix: I believe this is broken both in `JsonConfigSource::insertRepository ` and `JsonManipulator::insertRepository`

src/Composer/Json/JsonManipulator.php
```diff
@@ -159,20 +159,32 @@ private function sortPackages(array &$packages = []): void
      */
     public function addRepository(string $name, $config, bool $append = true): bool
```
glaubinix, line 160: This test case here changes the behaviour

```php
public function testAddRepositoryByNumericKeyInObjectFormatKeepsItsPosition(): void
    {
        $manipulator = new JsonManipulator('{
    "repositories": {
        "foo": {
            "type": "composer",
            "url": "https://foo.test"
        },
        "1": {
            "type": "path",
            "url": "../one"
        },
        "bar": {
            "type": "composer",
            "url": "https://bar.test"
        }
    }
}');

        self::assertTrue($manipulator->addRepository('1', ['type' => 'composer', 'url' => 'https://replaced.test'], false));
        self::assertSame(
            ['https://replaced.test', 'https://foo.test', '../one', 'https://bar.test'],
            array_column(JsonFile::parseJson($manipulator->getContents())['repositories'], 'url')
        );
    }
```

In your branch this overwrites an existing entry and afterwards the repositories section only has three entries. No matter whether append is set to true/false

doc/03-cli.md
```diff
@@ -901,6 +901,17 @@ If your repository requires more configuration options, you can instead pass its
 php composer.phar config repositories.foo '{"type": "vcs", "url": "http://svn.example.org/my-project/", "trunk-path": "master"}'
 ```
 
+Repositories which have no `name` can be addressed by their position in the list instead, which is
+what `composer repo list` shows in brackets. The repository at that position is then replaced in
+place rather than a new one being added:
```
glaubinix: This is only true if there are no Composer repositories configured in the global composer.json. If there are for instance two entries in the global composer.json then these have index 0, 1 in the brackets and the entries in the local composer.json start with index 2

tests/Composer/Test/Config/JsonConfigSourceTest.php
```diff
@@ -91,6 +91,67 @@ public function testRemoveRepository(): void
...
+                },
+                [['type' => 'path', 'url' => '../b'], ['name' => 'foo', 'type' => 'vcs', 'url' => 'https://new.test']],
+            ],
+            'add by position replaces in place' => [
+                static function (JsonConfigSource $source): void {
+                    $source->addRepository('1', ['type' => 'path', 'url' => '../new']);
+                },
+                [['name' => 'foo', 'type' => 'vcs', 'url' => 'https://foo.test'], ['type' => 'path', 'url' => '../new']],
+            ],
+        ];
```
glaubinix, line 158: When I add the test case below
```
'set-url by position' => [
                static function (JsonConfigSource $source): void {
                    $source->setRepositoryUrl('1', 'https://updated.test');
                },
                [['name' => 'foo', 'type' => 'vcs', 'url' => 'https://foo.test'], ['type' => 'path', 'url' => 'https://updated.test']],
            ],
```
Then I run into
```
Composer\Pcre\PcreException : preg_match(): failed executing "{(?(DEFINE)
       (?<number>    -? (?= [1-9]|0(?!\d) ) \d++ (?:\.\d++)? (?:[eE] [+-]?+ \d++)? )
       (?<boolean>   true | false | null )
       (?<string>    " (?:[^"\\]*+ | \\ ["\\bfnrt\/] | \\ u [0-9A-Fa-f]{4} )* " )
       (?<array>     \[  (?:  (?&json) \s*+ (?: , (?&json) \s*+ )*+  )?+  \s*+ \] )
       (?<pair>      \s*+ (?&string) \s*+ : (?&json) \s*+ )
       (?<object>    \{  (?:  (?&pair)  (?: , (?&pair)  )*+  )?+  \s*+ \} )
       (?<json>      \s*+ (?: (?&number) | (?&boolean) | (?&string) | (?&array) | (?&object) ) )
    )^(?P<start>\s*\{\s*(?:(?&string)\s*:\s*(?&json)\s*,\s*)*?"repositories"\s*:\s*\[\s*((?&json)\s*+,\s*+){1})(?P<repository>(?&object))(?P<end>.*)}sx": Backtrack limit exhausted
```

glaubinix: I think the `PcreException` was there before but somehow the code changes here make it possible/easier to reach the code path?

src/Composer/Json/JsonManipulator.php
Seldaek: Good catch, that was wrong. `insertRepository()` was resolving the *new* repository's name to a position of its own, so `0` deleted the entry at position 0. Since `--before`/`--after` already state the position, the name argument there only ever names the new repository — it is now resolved by name only (`findRepositoryKey($name, false)` / `findRepositoryIndex($repos, $name, false)`), so nothing gets removed.

Your exact reproduction is now a test, and with named repositories it errors instead — see the reply on the docs thread for the rule.

Seldaek: Correct, it was broken in both. Fixed in `JsonManipulator::insertRepository()` and in the `JsonConfigSource::insertRepository()` fallback, which now takes the same `$allowIndex` flag.

While adding your `set-url by position` case I found the `JsonConfigSource::setRepositoryUrl()` fallback had the same class of bug — it compared `$name === $index`, string against int, so it never matched anything in a list. That one is fixed too.

doc/03-cli.md
Seldaek: You are right, and it is worse than that — the numbers are not offset, they are reordered. `Config::merge()` reverses, re-keys on collision, then reverses back, so with two repositories in the global `config.json` and three locally, `repo list` prints:

```
[4] path ../l0
[3] path ../l1
[2] path ../l2
[0] composer https://g1.test
[1] composer https://g2.test
```

while `config repositories.0` still targets `../l0`, the first entry of the local file.

That killed the justification for the feature, so the rule is now:

- a repository is addressed by its **name**
- a numeric key falls back to the position **in the file being modified**, and **warns** that this is unreliable
- if the entry at that position has a `name`, it is **refused** with `The repository at position 1 is named "b", address it by that name instead.`

The check lives in one place (`BaseConfigCommand::validateRepositoryKey()`) for both the `config` and `repo` commands, and applies to whichever argument does the addressing — which for `--before`/`--after` is the ordering option, not the new name. The docs no longer claim the number is what `repo list` shows.

One inconsistency I deliberately left: `repo get-url N` still resolves against the merged config, so it can disagree with `repo set-url N`. That is pre-existing on the read side and we kept this change to writes only.

tests/Composer/Test/Config/JsonConfigSourceTest.php
Seldaek: Reproduced — thanks, that was a real gap. `setRepositoryUrl()` was the only one of the six regex methods without a `PREG_BACKTRACK_LIMIT_ERROR` guard, so it threw instead of degrading to the whole-file fallback. It now catches it like the others.

The fallback it degrades to was also broken (`$name === $index`, string vs int, never matched in a list), so your case would have silently done nothing even with the guard. Both are fixed and your `set-url by position` case is in `provideRepositoryFallback()`.

Seldaek: Your reading is right: pre-existing, but this branch made it reachable. On `main` `setRepositoryUrl('1', ...)` never resolved a numeric name at all, so it returned early and the regex was never built. Now that it resolves, the unguarded `Preg::isMatch()` is hit — which is exactly what your test surfaced.

src/Composer/Json/JsonManipulator.php
Seldaek: This one is intentional, and I would like to keep it. In the object format a numeric key *is* a name, so `"1"` should be replaced rather than duplicated — the reason it was not is the same int-vs-string bug as elsewhere: casting a `stdClass` to array yields **int** keys, so `$repositoryIndex === $name` never matched. `#12623` is the list-format symptom of it; this is the object-format one, and the reporter's own target JSON in that issue uses `"0"`/`"1"` keys.

Your case is now a test (`testAddRepositoryByNumericKeyInObjectFormatReplacesTheNamedEntry`), asserting the three-entry result. Note it does not warn or throw: this resolves by name, not by position, so the new rule does not apply to it.

src/Composer/Command/BaseConfigCommand.php
```diff
@@ -80,6 +80,45 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
...
+
+        if (is_array($target) && isset($target['name']) && is_string($target['name'])) {
+            throw new \RuntimeException(sprintf(
+                'The repository at position %s is named "%s", address it by that name instead.',
+                $name,
+                $target['name']
+            ));
+        }
+
+        $this->getIO()->writeError('<warning>Addressing a repository by its position ("'.$name.'") is unreliable as positions shift when repositories are added or removed, and do not match what "composer repo list" shows when the global config defines repositories. Give the repository a "name" and use that instead.</warning>');
```
glaubinix, line 119: Nice! Only downside is that we now have three methods (this one, findRepositoryIndex, findRepositoryKey) which all do more or less the same thing

glaubinix (approved):

src/Composer/Command/BaseConfigCommand.php
Seldaek: https://github.com/composer/composer/pull/13025/commits/5f4852fa7ee550d0f0dee0edbaf4e34f122f0005 unifies that, ish.. It kinda replaced one with another one, I am not even sure anymore :D

PR #13028, opened 2026-08-11
Replace deprecated spl_object_hash with spl_object_id

### LINKED ISSUE #13027: PHP 8.6 deprecation notice
Deprecation Notice: Function spl_object_hash() is deprecated since 8.6, consider using spl_object_id() instead in phar://C:/tools/php/composer/src/Composer/EventDispatcher/EventDispatcher.php:713

I see this on Windows-builds only

lazerg:
PHP 8.6 deprecates spl_object_hash() in favour of spl_object_id(), so the solver and repository code paths that key arrays by object identity now emit a notice on every call. This swaps all of them over and updates the affected array key types in the docblocks.

The only signature change is Pool::getRemovedVersionsByPackage(), which now takes an int since there is no hash left to pass it.

Fixes #13027

### DIFF
--- a/src/Composer/DependencyResolver/LockTransaction.php
+++ b/src/Composer/DependencyResolver/LockTransaction.php
@@ -27,9 +27,9 @@ class LockTransaction extends Transaction
     /**
      * packages in current lock file, platform repo or otherwise present
      *
-     * Indexed by spl_object_hash
+     * Indexed by spl_object_id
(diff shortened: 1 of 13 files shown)


Seldaek:
Lucky for us spl_object_id was added in 7.2 which is the minimum version we support :)

Thanks for the PR!

Issue #12989, opened 2026-07-20
Introduce an equivalent of npm audit fix


glaubinix:
**Is your feature request related to a problem? Please describe.**
Theses days more and more security advisories are dropped every day making it quite cumbersome to stay on top of them and keeping your lock file up to date. Especially when maintaining multiple Composer projects. The process involves regularly checking dependencies, running `composer audit` or some automated version of that and then manually updating any vulnerable packages.

**Describe the solution you'd like**
It would be great to have an option similar to `npm audit fix [--force-all]` which performs an audit and then updates all vulnerable packages to a secure version.

**Describe alternatives you've considered**
Running `composer audit` and manually copy and pasting all affected packages to run a `composer update` is rather cumbersome.

**Additional context**

PR #13029, opened 2026-08-12
Add vulnerable-only dependency updates

endroid:
Since composer now knows what packages are vulnerable, it can provide a flag to only update vulnerable packages. This way processes / jobs that actively fix vulnerabilities can run safer without introducing new issues or 0-day exploits from other packages.


src/Composer/Command/UpdateCommand.php
```diff
@@ -140,6 +141,35 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         }
 
         $packages = $input->getArgument('packages');
+
+        if ($input->getOption('vulnerable-only')) {
```
IonBazan, line 145: Can we also add `audit fix` command as an alternative for those coming from npm?

endroid: That would be a nice alternative

IonBazan: Also mentioned in #12989

PR #13030, opened 2026-08-12
Add minimum package age option

endroid:
Add a global --min-age option that filters dependency candidates by release date while preserving locked and platform packages. This protects against 0-day exploits and when used in conjunction with the --only-vulnerable flag (as a separate call) it can fix vulnerabilities and update packages in general while allowing for some time for vulnerabilities to be detected in packages before installing.


glaubinix:
Looks like a duplicate of https://github.com/composer/composer/pull/12692

endroid:
You're right, thank you for pointing that out. I didn't see that one and there is clearly overlap. I personally prefer smaller features / configuration that can work in conjunction, but that's a personal preference and if that one is preferred over this approach just continue on that. This one is just focused on one simple feature.

PR #13031, opened 2026-08-12
Return early from ClassLoader::findFile() for an empty class name

spawnia:
PHP can ask a registered autoloader for an **empty class name**. `ClassLoader::findFileWithExtension()` then does `$first = $class[0];` on `''`, which emits `Warning: Uninitialized string offset 0`. In applications that turn warnings into exceptions (Laravel's `HandleExceptions`, for instance) that aborts the request.

Reproduce on any composer-autoloaded project, PHP 8.3.31:

```php
<?php
is_callable('\::someMethod');
// Warning: Uninitialized string offset 0 in vendor/composer/ClassLoader.php on line 427
```

`is_callable('\::someMethod')` strips the leading `\` and asks the autoloader for class `''`. Traced with an `spl_autoload_register` spy:

```
-- is_callable('::a')
-- is_callable('\::a')
autoload requested: ''
-- is_callable('\Foo::a')
autoload requested: 'Foo'
-- is_callable('Foo::a')
autoload requested: 'Foo'
```

The `'::a'` vs `'\::a'` asymmetry looks like a php-src bug and I am filing it there too, but `ClassLoader` should not warn on any input it is handed.

This bit us for real: Laravel's `Factory::expandAttributes()` evaluates `is_callable($attribute)` on every string attribute, and a randomly generated password starting with `\::` (roughly 1 in a million with Faker's `password()`) made unrelated CI tests error out.

## The change

Return `false` immediately for an empty class name. Placing the guard in `findFile()` rather than `findFileWithExtension()` covers the classmap, APCu and PSR-0/PSR-4 lookups in one place, and skips polluting `missingClasses`/APCu with an entry that can never resolve.

## No test

I wrote one and then removed it: `ClassLoaderTest` instantiates `Composer\Autoload\ClassLoader`, but under `vendor/bin/simple-phpunit` that class name resolves to the bridge's own `vendor/bin/.phpunit/phpunit-*/vendor/composer/ClassLoader.php`, not `src/`. A test asserting the new behaviour therefore fails even with this patch applied (verified locally after `bin/composer install`). Happy to add one if you know a way around that.

### DIFF
--- a/src/Composer/Autoload/ClassLoader.php
+++ b/src/Composer/Autoload/ClassLoader.php
@@ -441,6 +441,11 @@ class ClassLoader
      */
     public function findFile($class)
     {
+        // PHP can ask for an empty class name, e.g. via is_callable('\::method')
+        if ('' === $class) {
+            return false;
+        }
+
         // class map lookup
         if (isset($this->classMap[$class])) {
             return $this->classMap[$class];


spawnia:
Can't explain the PHPStan failure in https://github.com/composer/composer/actions/runs/31610438293/job/94160094852#step:9:1 - might be flakyness. Can you retry?

spawnia:
I reproduced the `PHPStan (7.2, false)` job locally under conditions identical to CI, and it passes:

| Condition | CI | Local |
| --- | --- | --- |
| PHP | 7.2 via `setup-php`, extensions `intl, zip`, `memory_limit=-1` | 7.2.34 (`php:7.2-cli`), `intl` + `zip` compiled in, `memory_limit=-1` |
| Dependencies | `composer config platform --unset && composer install --ansi --no-interaction --prefer-dist` | same |
| PHPStan | 1.12.33 (from `composer.lock`) | 1.12.33 |
| Pre-step | `vendor/bin/simple-phpunit --filter NO_TEST_JUST_AUTOLOAD_THANKS` | same |
| Command | `composer phpstan` | same |
| Files analysed | 506 | 506 |
| Result | `Child process error (exit code 135): Bus error (core dumped) while running parallel worker` | `[OK] No errors` |

A bus error at 100% progress is a runner-level crash, not an analysis error — and the diff is five added lines in `ClassLoader::findFile()`, which cannot segfault a PHPStan worker. So this looks like runner flakiness.

The branch is already up to date with `main`, so there is nothing to rebase and no way for me to trigger a fresh run from my side. Could you rerun the failed job? Marking the PR ready for review.

Seldaek:
This is always a bug in the calling code IMO, so I don't want to account for it with an additional check in a code path that is executed to load every single class in almost all PHP projects, that's just wasteful. Callers should ensure they do not use empty class names as that is never a valid class.

Issue #12615, opened 2025-11-14
"curl error 3" on Windows


phenaproxima:
I don't have a specific command to share here, but I use Composer to build the [Drupal CMS Launcher](https://github.com/drupal/cms-launcher) automatically on GitHub Actions. The workflow runs for Linux, macOS, and Windows, which involves running several Composer commands at one point.

Everything was working just fine on Composer 2.8.x, but as soon as the system switched to 2.9.x, the Windows workflow started failing. Here's a build log (linked to the offending area): https://github.com/drupal/cms-launcher/actions/runs/19337936548/job/55318143930#step:12:44

Specifically, it seems to try to download this URL: `https://repo.packagist.org/p2/lib-curl-schannel%0Azlib version => 1.3.1%0Alibssh version => libssh2.json`

Yes, including the spaces and unencoded `=>` and everything, as far as I can tell. This only happens on the Windows builds. 2.9.1 is also affected.

Help?

Seldaek:
Could you run this in your CI so we see what that outputs?

    php -r "(new ReflectionExtension('curl'))->info();"

I think this has a version info that is not matching what we expect, and it breaks..

Amorok:
I'm having the same issue on mac osx 11.7.

Composer 2.8 worked, composer 2.9 fails with this curl info given above, and rolling back to 2.8 works.

anything required to help diagnose?

Seldaek:
As i wrote above run this and share the output.. 

    php -r "(new ReflectionExtension('curl'))->info();"

Amorok:
WORKING 2.8.12


curl

cURL support => enabled
cURL Information => 8.1.2
Age => 10
Features
AsynchDNS => Yes
CharConv => No
Debug => No
GSS-Negotiate => No
IDN => No
IPv6 => Yes
krb4 => No
Largefile => Yes
libz => Yes
NTLM => Yes
NTLMWB => Yes
SPNEGO => Yes
SSL => Yes
SSPI => No
TLS-SRP => No
HTTP2 => Yes
GSSAPI => Yes
KERBEROS5 => Yes
UNIX_SOCKETS => Yes
PSL => No
HTTPS_PROXY => Yes
MULTI_SSL => Yes
BROTLI => No
ALTSVC => Yes
HTTP3 => No
UNICODE => No
ZSTD => No
HSTS => Yes
GSASL => No
[...]

PR #13046, opened 2026-08-25
Prevent curl SSL version parsing across lines

emmadesilva:
## Summary

Composer’s curl SSL parser can consume subsequent `php --ri curl` lines when the SSL backend reports an unversioned value such as `Schannel`. This produces a malformed multiline `lib-curl-schannel...` platform package and breaks dependency resolution on Windows.

This patch restricts the SSL library and version captures to a single line. An unversioned Schannel entry is ignored, while `ext-curl`, `lib-curl`, zlib, and libssh2 remain available normally.

- See #12615.

## Testing

- Added a regression fixture for:
  - `SSL Version => Schannel`
  - `ZLib Version => 1.3.2`
  - `libSSH Version => libssh2/1.11.1`
- Verified the old parser reproduces the malformed multiline package name.
- Focused platform tests pass: 60 tests, 793 assertions.
- Full test suite passes: 3,266 tests, 7,312 assertions.

### DIFF
--- a/src/Composer/Repository/PlatformRepository.php
+++ b/src/Composer/Repository/PlatformRepository.php
@@ -234,7 +234,7 @@ class PlatformRepository extends ArrayRepository
                     $info = $this->runtime->getExtensionInfo($name);
 
                     // SSL Version => OpenSSL/1.0.1t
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #13048, opened 2026-08-26
Do not use a conflicted content-hash as the autoloader suffix

### LINKED ISSUE #13047: Merge conflict marker in composer.lock causes autoload.php file to contain invalid php syntax after composer install
Created as follow up from https://github.com/composer/composer/pull/11517#issuecomment-5411457372

My `composer.json`:

```json
{
    "name": "vendor/test",
    "require": {}
}
```

My `composer.lock`:

```json
{
    "_readme": [
        "This file locks the dependencies of your project to a known state",
        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
        "This file is @generated automatically"
    ],
<<<<<<< HEAD
    "content-hash": "2e7be2ea4bdb6deb3ff2826621f02c46",
||||||| 80ba9167
    "content-hash": "57f98f4c651b894d654eb3d465a648b1",
=======
    "content-hash": "e4b192e093a0a4647d7b0ef25f8d5e2d",
>>>>>>> something
    "packages": [],
    "packages-dev": [],
    "aliases": [],
    "stability-flags": {},
    "prefer-stable": false,
    "prefer-lowest": false,
    "platform": {},
    "platform-dev": {},
    "plugin-api-version": "2.9.0"
}
```

```
...
Composer version: 2.10.2
PHP version: 8.5.9
...
Checking composer.lock:
In JsonFile.php line 398:

  "./composer.lock" does not contain valid JSON
  Parse error on line 6:
  ...tomatically"    ],<<<<<<< HEAD    "co
  --------------------^
  Expected: 'STRING' - It appears you have an extra trailing comma
```

When I run this command: `rm -Rf vendor/ && composer install`, it runs fine without errors and when I look at the resulting file `vendor/autoload.php`, I notice a broken php line at the bottom:
```
return ComposerAutoloaderInitVCS merge conflict detected. Please run `composer update --lock`.::getLoader();
```

Running `composer install` should throw an error if the `composer.lock` is invalid

Amoifr:
Fixes #13047

When a lock file carries an unresolved merge conflict around `content-hash`, `JsonFile::parseJson()` replaces the conflicted block with a human readable message so the file still parses and the "please run `composer update --lock`" hint can be shown. `AutoloadGenerator::dump()` then used that value as the autoloader class suffix, producing:

```php
return ComposerAutoloaderInitVCS merge conflict detected. Please run `composer update --lock`.::getLoader();
```

which is not valid PHP, so the application crashed at runtime while everything else in the deploy looked fine.

The suffix now falls back to the random one (the branch that already exists for the unlocked case) when the lock's `content-hash` is not a hash, leaving the placeholder and its message untouched. Base branch is `main`: 2.2 has neither the placeholder nor the content-hash suffix.

The new test in `DumpAutoloadCommandTest` fails on the current code with the message inside `vendor/autoload.php`.

### DIFF
--- a/src/Composer/Autoload/AutoloadGenerator.php
+++ b/src/Composer/Autoload/AutoloadGenerator.php
@@ -426,7 +426,10 @@ EOF;
             }
 
             if (null === $suffix) {
-                $suffix = $locker !== null && $locker->isLocked() ? $locker->getLockData()['content-hash'] : bin2hex(random_bytes(16));
+                // a lock file with an unresolved merge conflict has its content-hash replaced by a
+                // human readable message, which would end up inside the autoloader class names
+                $contentHash = $locker !== null && $locker->isLocked() ? $locker->getLockData()['content-hash'] : null;
+                $suffix = is_string($contentHash) && Preg::isMatch('{^[a-f0-9]+$}', $contentHash) ? $contentHash : bin2hex(random_bytes(16));
             }
         }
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #12975, opened 2026-07-04
Add installed version to sub-dependencies in show --tree JSON output

lazerg:
In `composer show --tree --format=json`, sub-dependency nodes report the `version` constraint from the parent's require rather than the installed version, so tooling has to parse `composer.lock` separately to get resolved versions. This adds an `installed` key alongside the existing constraint on each tree node, resolved from the installed repository, leaving the current `version` field untouched.

Fixes #12955

PR #13049, opened 2026-08-27
Add installed version to sub-dependencies in show --tree JSON output

### LINKED ISSUE #12955: `show --tree --format=json`: include installed version for sub-dependencies
In [git-pkgs/resolve](https://github.com/git-pkgs/resolve) we build dependency graphs by shelling out to each ecosystem's package manager and parsing the output. For Composer we'd like to use `composer show --tree --format=json`, but below the top level the `version` field is the constraint from the parent's `require` block rather than the installed version:

```json
{
  "name": "guzzlehttp/guzzle",
  "version": "7.12.3",
  "requires": [
    { "name": "guzzlehttp/promises", "version": "^2.5", "requires": [ ... ] }
  ]
}
```

`guzzlehttp/promises` is installed at `2.5.0` but the tree reports `^2.5`. `composer show --format=json` without `--tree` does give installed versions for every package, but it's flat, so we currently parse `composer.lock` directly to get both the graph and resolved versions in one pass.

Would you accept a PR adding an `installed` field to each tree node alongside the existing constraint? That keeps the current field's meaning intact and lets tooling get the full picture from one command.

```sh
composer init -n --name demo/demo
composer require guzzlehttp/guzzle
composer show --tree --format=json | jq '.installed[0].requires[] | select(.name=="guzzlehttp/promises").version'
# "^2.5"
composer show --format=json | jq '.installed[] | select(.name=="guzzlehttp/promises").version'
# "2.5.0"
```

Composer 2.10.1, PHP 8.4.

Seldaek:
In `composer show --tree --format=json`, sub-dependency nodes report the `version` constraint from the parent's require rather than the installed version, so tooling has to parse `composer.lock` separately to get resolved versions. This adds an `installed` key alongside the existing constraint on each tree node, resolved from the installed repository (honoring `replace`/`provide`, and unwrapping aliases so the real installed version is reported), leaving the current `version` field untouched.

This is a re-open of #12975 by @lazerg, which got auto-closed when the source fork was deleted and cannot be reopened. Commits are cherry-picked unchanged, so authorship is preserved.

Fixes #12955

### DIFF
--- a/src/Composer/Command/ShowCommand.php
+++ b/src/Composer/Command/ShowCommand.php
@@ -1309,6 +1309,11 @@ EOT
                 'version' => $require->getPrettyConstraint(),
             ];
 
+            $installedPackage = $installedRepo->findPackage($requireName, '*');
+            if (null !== $installedPackage) {
+                $treeChildDesc['installed'] = $installedPackage->getPrettyVersion();
+            }
+
             $deepChildren = $this->addTree($requireName, $require, $installedRepo, $remoteRepos, $packagesInTree);
 
             if ($deepChildren) {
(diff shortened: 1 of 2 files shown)


Seldaek:
cc @andrew

andrew (approved):

andrew:
Thanks!

Seldaek:
Alright `composer self-update --snapshot` will get you the new output.

PR #13051, opened 2026-08-27
Add batching to the security advisories API queries for projects with very large dep count

Seldaek:
I've noticed we hit the max_input_vars limit (1000) quite often on packagist.org, unsure which routes cause this or whether it's just silly security scanners, but it gave me the idea that maybe this should not do gigantic requests. That seems like an improvement anyway.

Malware requests are not included as those post a single json document it's not suffering from this limitation.


glaubinix:
~~@Seldaek malware has two APIs. One of them accepts a list of package names. Would apply this there too.~~ Not the same issue there

glaubinix (approved):

Issue #13052, opened 2026-08-27
In a script, an empty string should always be a no-op


fisharebest:
My `composer.json`:

```json
{
  "name": "test/composer",
  "description": "empty script should be a no-op",
  "bin": [
    "hello",
    "world"
  ],
  "scripts": {
    "test": [
      "echo BEFORE",
      "",
      "echo AFTER"
    ]
  }
}
```

Output of `composer diagnose`:

```
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking Composer version: OK
Composer version: 2.10.3
Checking Composer and its dependencies for vulnerabilities: OK
PHP version: 8.5.8
PHP binary path: /opt/homebrew/Cellar/php/8.5.8/bin/php
OpenSSL version: OpenSSL 3.6.3 9 Jun 2026
curl version: 8.21.0 libz 1.2.12 brotli brotli/1.2.0 zstd supported ssl OpenSSL/3.6.3 HTTP 1.0, 1.1, 2, 3
zip: extension present, unzip present, 7-Zip not available
Active plugins: 
Checking composer.json: WARNING
No license specified, it is recommended to do so. For closed-source software you may use "proprietary" as license.
Checking platform settings: OK
Checking git settings: OK git version 2.50.1
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com oauth access: OK does not expire
Checking disk free space: OK
```

When I run this command: 

```
composer test
```

I get the following output: 

```
Reading ./composer.json (/Users/gr4376/Projects/composer-bug/composer.json)
Loading config file /Users/gr4376/.composer/config.json
Loading config file /Users/gr4376/.composer/auth.json
Loading config file ./composer.json (/Users/gr4376/Projects/composer-bug/composer.json)
Checked CA file /opt/homebrew/etc/ca-certificates/cert.pem: valid
Executing command (CWD): 'git' '--version'
Executing command (/Users/gr4376/Projects/composer-bug): 'git' 'branch' '-a' '--no-color' '--no-abbrev' '-v'
Executing command (/Users/gr4376/Projects/composer-bug): 'git' 'describe' '--exact-match' '--tags'
[...]

PR #13053, opened 2026-08-27
Fix: #13052 - an empty command in a script should always be a NO-OP

fisharebest:
In the EventDispatcher, an empty string will always match this REGEX.

https://github.com/composer/composer/blob/53e6ddca87c033e2db254c8773124d1d7afed08d/src/Composer/EventDispatcher/EventDispatcher.php#L370

Therefore an empty command will match the first registered binary.

The fix is to add a guard to prevent empty commands getting any special processing.


fisharebest:
Fixes #13052

(adding comment, as this isn't being picked up from the title).

Seldaek:
Thanks

PR #13058, opened 2026-09-03
Document the six composer.json fields missing from the schema chapter

### LINKED ISSUE #12484: Docs: 04-schema.md missing some fields
The documentation says https://github.com/composer/composer/blob/d2a4543db8adc6b3d4b2fb652c953db629328986/doc/04-schema.md?plain=1#L3

But comparing the list with the schema on https://getcomposer.org/schema.json (supported configuration fields extracted with `jq` with the filter `.properties | keys`, see https://play.jqlang.org/s/LDPX2x9E7DnX-JC), the following fields appear to be missing from the documentation

1. `default-branch`
2. `dist`
3. `php-ext`
4. `scripts-aliases` [documented in articles/scripts.md, so maybe just add a quick section pointing there, like for `scripts`?]
5. `scripts-descriptions` [also in articles/scripts.md]
6. `source`

Amoifr:
Fixes #12484

The chapter opens with "This chapter will explain all of the fields available in `composer.json`", which was not quite true: six of the 39 properties in `res/composer-schema.json` had no section. This adds one for each, so the claim holds again.

`source` and `dist` say what the fields contain, that `source` also requires a `reference` while `dist` does not, and that you normally do not write them yourself, with a pointer to the `package` repository section of the repositories chapter where you do. That is the "minor use cases" @Seldaek mentioned.

`scripts-descriptions` and `scripts-aliases` are marked root-only like `scripts`, since `Application` reads them from `Factory::getComposerFile()`, and point at the two existing sections of `articles/scripts.md`.

`php-ext` follows what @stof explained in the issue: it belongs to PIE, and the section links to PIE's extension maintainers doc rather than repeating it here.

`default-branch` says plainly not to set it, and why setting it would not help anyway.

Three claims I checked rather than inferred:

- `composer validate` **errors** on `php-ext` outside a `php-ext` / `php-ext-zend` package, and **warns** on a description or an alias given for a script that does not exist. Ran on a fixture with all three problems, got exactly one entry under "General errors" and two under "General warnings".
- Composer really does discard `default-branch` from the file. I built a git repo with two branches, both declaring `"default-branch": true`, and loaded it through `VcsRepository`:

```
9999999-dev            isDefaultBranch=true
dev-trunk              isDefaultBranch=true
dev-side               isDefaultBranch=false
```

  `trunk` is the repository's default branch, `side` is not, and what the file said made no difference.

All six anchors used (`#type`, `#source`, `#root-package`, `05-repositories.md#package-1`, and the two in `articles/scripts.md`) resolve to a real heading, and the PIE link returns a 200 (the URL given in the issue is a 404 today, php/pie's default branch is `1.5.x`, so this one goes through `HEAD`).

Two placement choices are yours to overrule. `source`, `dist`, `php-ext` and the two `scripts-*` sections sit where the schema puts them. `default-branch` I put next to `non-feature-branches` at the end, since both are about branches and it keeps an internal-only field out of the early part of the page, rather than after `version` where the schema has it.

Note that `doc/**` is in `paths-ignore` for every workflow, so no CI runs on this.

### DIFF
--- a/doc/04-schema.md
+++ b/doc/04-schema.md
@@ -306,6 +306,56 @@ An example:
 
 Optional.
 
+### source
+
+Where the package sources live, used when installing the package from source.
+
+`type`, `url` and `reference` are required, `mirrors` is optional.
+
+```json
+{
+    "source": {
+        "type": "git",
+        "url": "https://github.com/composer/composer.git",
+        "reference": "2.8.0"
+    }
+}
+```
+
+Where `type` is `git`, the reference is a commit id, branch or tag name. Where
+it is `svn`, it is the reference that gets appended to the URL when running
+`svn co`.
+
+You normally do not write this field yourself. The repository a package comes
+from provides it: Packagist and other `composer` repositories fill it in from
+the VCS they read the package from, and a VCS repository fills it in from the
+branch or tag it is reading. The place you do write it by hand is the
+definition of a [`package` repository](05-repositories.md#package-1).
+
+Optional.
+
+### dist
+
+Where the package archive lives, used when installing the package from dist.
+
+`type` and `url` are required, `reference`, `shasum` and `mirrors` are
+optional.
+
+```json
+{
+    "dist": {
+        "type": "zip",
+        "url": "https://www.smarty.net/files/Smarty-3.1.7.zip"
+    }
+}
+```
+
+The same applies as for [source](04-schema.md#source): this is normally
+provided by the repository the package comes from, and written by hand only in
+a [`package` repository](05-repositories.md#package-1).
+
+Optional.
+
 ### Package links
 
 All of the following take an object which maps package names to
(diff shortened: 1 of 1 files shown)


doc/04-schema.md
```diff
@@ -306,6 +306,56 @@ An example:
...
+
+You normally do not write this field yourself. The repository a package comes
+from provides it: Packagist and other `composer` repositories fill it in from
+the VCS they read the package from, and a VCS repository fills it in from the
+branch or tag it is reading. The place you do write it by hand is the
+definition of a [`package` repository](05-repositories.md#package-1).
+
+Optional.
+
+### dist
```
stof, line 337: This should probably mention the placeholders available in the URL, which are useful to define the URL in a way that can apply to all available versions (useful when specifying that from a composer.json file rather than being guessed by a driver. Drivers generally provide the full URL directly for each version)

Amoifr: That makes complete sense, thanks. It is exactly the case where you write the field by hand, so it is the case the section should serve. Added in edd82ff.

The section now lists the five placeholders (`%package%`, `%version%`, `%prettyVersion%`, `%reference%`, `%type%`) with an example serving every version from one URL.

Two things I checked in `Package::getUrls()` and `ComposerMirror::processUrl()` rather than assume, since both would surprise someone writing the URL by hand:

- expansion happens for `dist` only, never for `source`, which matches your point that drivers hand over the full URL per version;
- `%version%` is MD5 hashed when the version contains a `/`, and `%reference%` unless it is already hexadecimal, so a branch or tag name used as the reference lands in the URL hashed rather than verbatim.

Happy to trim that last paragraph if you would rather keep the section short.

doc/04-schema.md
```diff
@@ -306,6 +306,86 @@ An example:
...
+
+The `url` may contain placeholders, which is what lets one URL serve every
+version of a package instead of one entry per version. They are expanded when
+the package is downloaded:
+
+| Placeholder | Replaced by |
+| --- | --- |
+| `%package%` | the package name, e.g. `acme/lib` |
+| `%version%` | the normalized version, e.g. `1.2.3.0` |
+| `%prettyVersion%` | the version as written, e.g. `1.2.3` |
```
Seldaek:
```suggestion
| `%prettyVersion%` | the version defined in the tag or composer.json, e.g. `v1.2.3` |
```

Seldaek:
Thanks

Issue #8231, opened 2019-07-17
composer.json is not readable.


moxuanyuan:
When I run this command:

```
composer require erusev/parsedown
```

I get the following output:

```
./composer.json is not readable.
```

And I expected this to happen:
Windows 10 adds DSM web folder as net drive through SMB. Login accounts have  read and write permissions to web folder, and then use composer require installation package through Windows CMD, and then prompt `./composer.json is not readable.`

Oddly, composer. Json is composer command automatically created, content is `{} `

using Google to check and found that someone had the same problem mentioned, <https://serverfault.com/questions/867669/composer-json-is-not-readable-when-running-on-synology-nas>

 someone replied to a solution, add the everyone read-write Permission to composer. json.

I tried, and it worked.But what I don't understand is that json is already created by the composer command, and it has content in it, so why can't we just add everyone read and write permissions?Also, only composer. json file is required, folder does not

alcohol:
I do not do any development on windows machine, so am unable to reproduce this myself. We'll need to find a developer familiar with the source code and access to a windows system to debug this further.

alcohol:
Looks related https://bugs.php.net/bug.php?id=68926

alcohol:
Looks like `is_readable` and `is_writable` might always return `false` when dealing with network drives/shares. Work around as suggested on the issue would be to check for this condition (networked drive/share) and determine through other means if a file is readable or writable.

Seldaek:
@moxuanyuan it'd be good if you can check if the fix work (you can try pre-release builds using `composer self-update --snapshot` and then try the composer require again)

0ddlyoko:
This error has been reported [here](https://bugs.php.net/bug.php?id=75384).

alcohol:
@Seldaek not sure if this fixes it because the `is_readable` call would still fail. You only moved/resolved the `is_writable` call.

Seldaek:
That php bug you linked says is_readable seems to work fine so I assumed it was ok..

alcohol:
No it applies to both. His original issue also shows that it already failed on the `is_readable` step, not the one after that (`is_writable`).

Seldaek:
I see, then I guess the is_readable should be fixed too..

alcohol:
Yeah it's a bit of a PITA. Especially since it is such an edge case scenario.

moxuanyuan:
[...]

PR #13069, opened 2026-09-11
Replace generated files instead of rewriting them in place

nicolas-grekas:
`file_put_contents()` truncates and rewrites the **existing inode**. Every path hardlinked to that inode — other worktrees, `cp -al` copies, `rsync --link-dest` trees, hardlink-based CI caches — therefore sees the new content, and a concurrent reader can observe a truncated or half-written file.

Reproducer, on `main`:

```sh
mkdir proj && cd proj
echo '{"name":"test/proj","autoload":{"files":["a.php"]}}' > composer.json
touch a.php b.php && composer dump-autoload
cd .. && cp -al proj proj2                     # hardlinked copy, e.g. a second worktree
cd proj
echo '{"name":"test/proj","autoload":{"files":["a.php","b.php"]}}' > composer.json
composer dump-autoload
grep -c b.php ../proj2/vendor/composer/autoload_files.php   # 1 — proj2 was modified too
```

## What this does

Adds `Filesystem::safeFilePutContents()`, which writes to a temp file next to the target and renames it onto it. The target gets a *new* inode, so hardlinked siblings keep theirs, and the path always points at a complete file. Same idea as what `Cache::write()` has been doing since #9580.

It is used for every file Composer generates or rewrites:

- `autoload_*.php`, `autoload.php`, `autoload_real.php`, `platform_check.php`, `include_paths.php`, `installed.php`, `InstalledVersions.php` — through `filePutContentsIfModified()`
- `ClassLoader.php` and `LICENSE` — through `safeCopy()`, which had the same problem via `fopen($target, 'w+')`
- `composer.json`, `composer.lock`, `installed.json`, `auth.json`, `config.json`
- the `vendor/bin` proxies

## Existing behavior that is deliberately kept

- **Permissions and ownership** are carried over onto the new inode (`chown`/`chgrp` only when running as root, where it can succeed).
- **A read-only file still fails to be written.** `rename()` only needs a writable *directory*, so without a guard Composer would start silently overwriting a `chmod 444` composer.json. An `is_writable()` check routes those to the direct write, which fails as before — and a false negative from `is_writable()` on a network mount (#8231) simply means today's behavior.
- **Symlinks are written through**, not replaced by a regular file.
- **Unmodified files are still not touched at all**, so `filePutContentsIfModified()` keeps doing what it was added for in #8529 (Docker layer caching). A test asserts the inode is unchanged in that case.
- **Every failure falls back to the direct write**, which reports the error exactly as it used to: restricted directory, out of disk space, or `rename()` failing with "Access denied" on Windows. That last one is why this deliberately does not go through `Filesystem::rename()` — its fallbacks copy the file over the target, which is the very thing being avoided here.
- The temp file is created `0600` before the content goes into it, so `auth.json` contents are never briefly world-readable.

## Tests

7 new tests in `FilesystemTest` cover the hardlink case, permission preservation, writing through symlinks, no leftover temp files, and the unmodified-file case. Full suite passes, PHPStan is clean.


nicolas-grekas:
This solves using composer in worktrees BTW, that's how I hit the problem.

PR #13070, opened 2026-09-13
CI: Speedup PHPStan analysis

staabm:
setup result cache persistence in CI per documentation in https://staabm.github.io/2023/10/21/phpstan-result-cache-gotchas.html#github-actions 

- Add `-vv` option, to make PHPStan output information whether result-cache is used (or why not)
- Setup result-cache caching between CI runs

----

before PR: phpstan analysis ~60seconds

after PR (warm cache): phpstan analysis ~2-5seconds

----

when updating to PHPStan 2.x the result-cache will be even more useful, as it needs to be invalidated less often.


staabm:
@Seldaek here another one ;)

Seldaek:
Thanks

PR #13073, opened 2026-09-16
Send AI agent info as part of user agent

eiriksm:
This PR adds another field to the user agent, using a similar pattern as the "CI" flag. This could be useful for analyzing traffic to a repo, for example

The detection is based on (almost copied from) https://github.com/laravel/agent-detector

We could of course pull that in as a dependency, but feels safer and still approachable and controllable to simply maintain on the side. Also, this is what phpstan does :sweat_smile: 

I was wondering whether I should add the new user agent field on the end, as a new field, or let the command still be the last one. Trying to speculate how people have implemented parsers of this user agent, for me it's 50/50 if they added logic for parsing the command being the last part, or a specific number. Happy to swap that out

I also went for a boolean switch here. Either agent or not. The library mentioned above also tries to identify the different known ones, which we could pass. Not sure that is useful or necessary here


src/Composer/Util/Platform.php
```diff
@@ -21,6 +21,38 @@
  */
 class Platform
 {
+    /**
+     * Env vars which indicate that Composer is being run by an AI coding agent
+     *
+     * Copied from https://github.com/laravel/agent-detector. Thanks!
+     *
+     * @internal
+     * @var non-empty-list<non-empty-string>
```
stof: `@var` is useless on a constant. phpstan analyses it using the exact value anyway as it cannot change

eiriksm: Cool, thanks! 👌

src/Composer/Util/StreamContextFactory.php
```diff
@@ -125,14 +125,15 @@ public static function initOptions(string $url, array $options, bool $forCurl =
...
-                'User-Agent: Composer/%s (%s; %s; %s; %s%s%s%s)',
+                'User-Agent: Composer/%s (%s; %s; %s; %s%s%s%s%s)',
                 Composer::getVersion(),
                 function_exists('php_uname') ? php_uname('s') : 'Unknown',
                 function_exists('php_uname') ? php_uname('r') : 'Unknown',
                 $phpVersion,
                 $httpVersion,
                 $platformPhpVersion ? '; Platform-PHP '.$platformPhpVersion : '',
                 Platform::getEnv('CI') ? '; CI' : '',
+                Platform::isCodingAgent() ? '; agent' : '',
```
Seldaek, line 136: Would maybe be nice to cache this call as it is checking quite a bit of env vars and filesystem.. all of it being static tho, it'll make testing more difficult I can see that.

eiriksm: Thanks, updated to change to that (squashed the branch so no direct diff, sorry)

PR #13081, opened 2026-09-22
Url: preserve SSH user in stripCredentials()

### LINKED ISSUE #13078: ssh/git user stripped from repo address on private satis+forge deps
Hello,

I opened a [discussion](https://github.com/composer/composer/discussions/13074) and it seems this behavior may not be intended.

we have a forge for some of our internal packages, and we use a private satis instance to distribute/reference theses packages

in some of our project we configured theses packages using "dev-" or "-dev" version to get a git-branch we can work on from the main project, with a composer.json like:

```
    "require": {
        "internal/package-A": "dev-feature/branch",
```

but, then when we `composer install` the git repository address get stripped from the forge user:

```
❯ git remote show -n origin
* remote origin
  Fetch URL: ssh://git.int....
  Push  URL: ssh://git.int....
```

same thing happen with the "composer" git-remote

before version 2.10.3 the result included our forge ssh-user like this (the "gogs@"):

```
❯ git remote show -n origin
* remote origin
  Fetch URL: ssh://gogs@git.internal....
  Push  URL: ssh://gogs@git.internal....
```

Trying to do a git-bisect point me to this [commit](https://github.com/composer/composer/commit/ada00550c3cdec54fe49565ef2874d08744c90ae) and so this related [PR](github.com/composer/composer/pull/13044)

In the [discussion](https://github.com/composer/composer/discussions/13074) @Yuno15-bb did some testing with the stripCredentials method and found out the behavior change from the code.

let me know if you need more information about this issue :)

thanks a lot for the great work on composer! ;)

webdevsamran:
Fixes #13078

### Problem
Following commit `ada00550c` in PR #13044, `Git::runCommands()` replaces `%sanitizedUrl%` using `Url::stripCredentials($url)`. `Url::stripCredentials()` stripped all userinfo matching `{://[^/\s?#]+@}`.

While stripping the entire userinfo is intended for HTTP(S) where a bare token in the user slot (`https://token@host/repo.git`) is a secret credential, in SSH URLs (`ssh://git@forge.example.com/...`, `git+ssh://gogs@...`), `git` or `gogs` is the remote SSH username rather than a secret credential. Stripping the SSH username causes Git remote commands to fall back to the local OS user, triggering authentication failures or password prompts on private Satis / Forge dependencies.

### Solution
In `Url::stripCredentials()`:
- Keep the user segment when the scheme is not `http(s)` and no password is present in the userinfo (`ssh://git@...` -> `ssh://git@...`).
- When a password is present, strip it along with the user info as before (`ssh://user:pass@...` -> `ssh://...`).
- Continue stripping bare tokens and credentials on HTTP(S) URLs (`https://ghp_xxx@...` -> `https://...`).

### Verification
- Added test coverage in `tests/Composer/Test/Util/UrlTest.php` covering `ssh://`, `git+ssh://` with and without passwords.
- Existing tests in `UrlTest` (124 tests, 124 assertions) and `GitTest` (20 tests, 38 assertions) all pass.
- PHPStan analysis passed with 0 errors.


webdevsamran:
Hi @naderman! Just re-triggered CI with an empty commit. The previous `PHPStan (7.2, false)` run failed with `Bus error (core dumped)` (exit code 135 = SIGBUS) — a transient CI runner OOM crash rather than a code error. All other checks (PHP Lint, Autoloader, API Surface Check, Continuous Integration) are green.

